PR #2360, opened 2023-02-24
db: wait for compactions in scan step

bananabrick:
If we're waiting on a condition variable, then we must check that the condition is met after reacquiring the mutex.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -3207,7 +3207,7 @@ func (d *DB) scanObsoleteFiles(list []string) {
 	d.opts.DisableAutomaticCompactions = true
 
 	// Wait for any ongoing compaction to complete before continuing.
-	if d.mu.compact.compactingCount > 0 || d.mu.compact.flushing {
+	for d.mu.compact.compactingCount > 0 || d.mu.compact.flushing {
 		d.mu.compact.cond.Wait()
 	}
 


bananabrick:
This will most likely fix: https://github.com/cockroachdb/pebble/issues/2325, but it's also a bug in general.

RaduBerinde (commented):
LGTM  but it seems odd that we don't do this cleanup before starting compactions - and we end up blocking `Open` here to wait on them to finish. What would be the problem with disabling automatic compactions until after the cleanup step?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2360#-:-NP4x9Hh1YWP-F-3y0OI:b-mwogua)* status: 0 of 1 files reviewed, all discussions resolved

bananabrick (commented):
We might be able to make that work by moving around the order of operations of WAL replay, format major version upgrade, clean up.

The clean up currently happens after we perform the compactions required to update the format major version in `ratchetFormatMajorVersionLocked`. I think this format major version is written to a new Options file, which makes the previous Options file obsolete.

If we want we can do the clean up before, but we'll end up with an additional Options file which is obsolete, but not deleted. We could just delete the Options file later though.

We might also have to move up the clean up before the WAL replay. WAL replay can cause flushes, which in turn can schedule compactions. But if we move the clean up before the WAL replay, then we'll end up with obsolete WAL files in the directory, which aren't deleted.

> What would be the problem with disabling automatic compactions until after the cleanup step?

If we assume the current order of operations, then the `ratchetFormatMajorVersionLocked` step won't work because it synchronously performs compactions.

Currently, in the worst case, we will have to wait for 3 concurrent compactions to finish before we proceed with the `scanObsoleteFiles` step. Each compaction won't take longer than a few seconds even in the worst case, so we're not blocking for long.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2360#-:-NP5-hbT3wDXOcUZldk_:brvirtn)* status: 0 of 1 files reviewed, all discussions resolved

RaduBerinde (commented):
The compactions that are run synchronously (like that from the WAL replay) don't matter, they would be completed by the time we clean up. I assumed that in the issue it was an automatic compaction that was still running, but maybe it's something that's from `ratchet` that's not synchronous? Yeah, it should be fine waiting for those.

Thanks for entertaining my questions, this still LGTM

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2360#-:-NP57k8G7r6_ri32i-6J:b-ovshoj)* status: 0 of 1 files reviewed, all discussions resolved

bananabrick (commented):
I believe that once a compaction is scheduled, we could schedule more compactions, since `db.compact` calls `db.maybeScheduleCompaction`, so it's not synchronous. The ratchet step is synchronous in the sense that it may wait for the necessary compactions to complete.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2360#-:-NP67Ckz753mzuinVYuw:b9f36h)* status: 0 of 1 files reviewed, all discussions resolved

RaduBerinde:
Then maybe we should start out with DisableAutomaticCompactions and enable that at the end?

bananabrick (commented):
> Then maybe we should start out with DisableAutomaticCompactions and enable that at the end?

We need that enabled to do any compactions at all, as we do in the ratchet step.

I don't think waiting for a few compactions before the `scanObsoleteFiles` is something we need to optimize right now.
- We already wait for IO because we wait for flushes during WAL replay, and compactions during the ratchet step.
- There's an upper bound on the number of compactions we'll have to wait for.

We should probably merge this to fix the metamorphic test failure.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2360#-:-NPJ109C-3fiT5PUImOW:b-vmgjqz)* status: 0 of 1 files reviewed, all discussions resolved

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2360#-:-NPJ8Fu5415WQbBWJkSw:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved

PR #2362, opened 2023-02-26
objstorage: allow attaching of shared objects

RaduBerinde:
This commit introduces the concept of a `SharedObjectBacking`. This is an opaque encoding of the metadata necessary to access a shared object (specifically creator ID and creator file num). The backing can be extracted for a shared object on an instance and then used to attach the object to another instance. This will be used for rebalancing.

### DIFF
--- a/objstorage/provider.go
+++ b/objstorage/provider.go
@@ -190,8 +190,8 @@ type ObjectMetadata struct {
 type CreatorID = sharedobjcat.CreatorID
 
 // IsShared returns true if the object is on shared storage.
-func (m *ObjectMetadata) IsShared() bool {
-	return m.Shared.CreatorID.IsSet()
+func (meta *ObjectMetadata) IsShared() bool {
+	return meta.Shared.CreatorID.IsSet()
 }
 
 // Open creates the Provider.
--- a/objstorage/provider_test.go
+++ b/objstorage/provider_test.go
@@ -6,6 +6,7 @@ package objstorage
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/cockroachdb/datadriven"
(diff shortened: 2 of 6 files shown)


itsbilal (approved):
LGTM, thanks for doing this so quick!

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2362#-:-NPJL80M3NtYphFCN12y:b-814g8k)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`objstorage/shared_backing.go` line 65 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2362#-NPEylDH60u72DEAP8ya-r1-65:-NPJFdGdCIpYGHvqVzgb:b-6usjut) ([raw file](https://github.com/cockroachdb/pebble/blob/9e159d182f3a5afe0ecdeb190d2d8ace9a5bc2df/objstorage/shared_backing.go#L65)):*

Might be a good idea to panic on unknown tags? That's what we usually do for things concerning on-disk formats.
</blockquote></details>

Hm, indeed what I have here is broken in that we are not skipping the "payload".

The idea was to allow more "ignorable" tags in the future. For example, maybe we start encoding the file type just to double check it's what's expected, but the new code treats it as optional. Or maybe we start using an "extended" creator ID but we also populate the old tag for backward compatibility (when the creator ID fits).

The manifest code also works like this when it comes to "custom tags", except that some tags are ignored and some cause errors (depending on `customTagNonSafeIgnoreMask`).

Updated to differentiate between ignorable and non-ignorable tags and to require encoding the length for future ignorable tags.

PR #2367, opened 2023-02-28
db: don't recycle tiny log files

bananabrick:
Flushable ingests can produce tiny log files and we shouldn't try and recycle these as there's a limit on the number of log files which are recycled and we want the limit to be used for larger log files.

### DIFF
--- a/checkpoint_test.go
+++ b/checkpoint_test.go
@@ -33,6 +33,7 @@ func TestCheckpoint(t *testing.T) {
 		FormatMajorVersion:    FormatNewest,
 		L0CompactionThreshold: 10,
 	}
+	opts.private.minLogSizeRecycler = 1
 
 	datadriven.RunTest(t, "testdata/checkpoint", func(t *testing.T, td *datadriven.TestData) string {
 		switch td.Cmd {
--- a/cleaner_test.go
+++ b/cleaner_test.go
@@ -96,6 +96,7 @@ func TestCleaner(t *testing.T) {
 				FS:     fs,
 				WALDir: dir + "_wal",
 			}).WithFSDefaults()
+			opts.private.minLogSizeRecycler = 1
 
 			for i := 1; i < len(td.CmdArgs); i++ {
 				switch td.CmdArgs[i].String() {
--- a/event_listener_test.go
+++ b/event_listener_test.go
@@ -47,6 +47,7 @@ func TestEventListener(t *testing.T) {
 			// can make these tests flaky. The TableStatsLoaded event is
 			// tested separately in TestTableStats.
 			opts.private.disableTableStats = true
+			opts.private.minLogSizeRecycler = 1
 			var err error
 			d, err = Open("db", opts)
 			if err != nil {
--- a/ingest_test.go
+++ b/ingest_test.go
@@ -438,6 +438,7 @@ func TestOverlappingIngestedSSTs(t *testing.T) {
 		// delete-only compactions triggered by ingesting range tombstones.
 		opts.DisableAutomaticCompactions = true
 		opts.Experimental.IngestSSTablesAsFlushable = true
+		opts.private.minLogSizeRecycler = 1
 
 		var err error
 		d, err = Open(dir, opts)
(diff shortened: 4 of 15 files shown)


bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2367)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`options.go` line 996 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2367#-NPKM7iCAhp90XDeldNZ:-NPKM7iDBYCeYQ1iSP3E:b-v1799b) ([raw file](https://github.com/cockroachdb/pebble/blob/4dfc43485e834aa937c793494840f697d5e10a04/options.go#L996)):*
> ```Go
> 		// By default use a 1MB value for log recycling since this will always
> 		// be greater than the log size produced by flushable ingestions.
> 		o.private.minLogSizeRecycler = 1 << 20
> ```

1MB should be lower than the WAL sizes which we want to recycle since the memtable sizes will be greater than this.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2367#-:-NPL8gQQ5ROwqwYFB0R3:bnfp4nl)* status: 0 of 15 files reviewed, 2 unresolved discussions (waiting on @bananabrick and @jbowens)

___
*[`open.go` line 147 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2367#-NPL8SMlC3uxPjNp76bC:-NPL8SMlC3uxPjNp76bD:b98uap5) ([raw file](https://github.com/cockroachdb/pebble/blob/4dfc43485e834aa937c793494840f697d5e10a04/open.go#L147)):*
> ```Go
> 		dataDir:             dataDir,
> 		walDir:              walDir,
> 		logRecycler: logRecycler{
> ```

[nit] should move to a `makeLogRecycler` function
___
*[`options.go` line 996 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2367#-NPKM7iCAhp90XDeldNZ:-NPL88V6EX91AtZTZ6ge:b-aan0lr) ([raw file](https://github.com/cockroachdb/pebble/blob/4dfc43485e834aa937c793494840f697d5e10a04/options.go#L996)):*

1MB should be lower than the WAL sizes which we want to recycle since the memtable sizes will be greater than this.
</blockquote></details>

[nit] `1024 * 1024` is more readable than `1 << 20`

jbowens (changes requested):
Has this been observed to be a problem? Log recycling should ensure that our working set of logs have a larger physical size, even if one individual reuse writes very little.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2367#-:-NPMoMKF5JNThNUkz0vT:bxudis)* status: 0 of 15 files reviewed, 2 unresolved discussions (waiting on @bananabrick and @RaduBerinde)

bananabrick (commented):
No, it was a suggestion made in the flushable ingestion pr.

>  Log recycler should ensure that our working set of logs have a larger physical size, even if one individual reuse writes very little.

I guess if a new tiny log is added to the recycler and later used as a log for memtable which will grow large, then the log will also grow large, and eventually, we'll only be recycling large log files.

That makes sense to me.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2367#-:-NPOHq4JEtNCHk318o6_:bu5o0qd)* status: 0 of 15 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

jbowens (commented):
> I guess if a new tiny log is added to the recycler and later used as a log for memtable which will grow large, then the log will also grow large, and eventually, we'll only be recycling large log files.

Yeah, I think that's right.

I think this behavior already exists, because a memtable-overlapping ingest will force an early flush. The log corresponding to the force-flushed memtable can be arbitrarily short. We also will force early flushes within a configurable time window for range deletions and range keys, which can produce shorter logical WALs. I suspect all of this is not a problem, because the WALs being recycled should grow to be the approximately the size of the largest logical WALs.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2367#-:-NPSs6A46F849BVIkfo0:b-oh9b75)* status: 0 of 15 files reviewed, 2 unresolved discussions (waiting on @bananabrick and @RaduBerinde)

bananabrick:
ok closing.

PR #2368, opened 2023-02-28
objstorage: better Writable API

RaduBerinde:
#### db: add missing error check in TestIngestLoad

We weren't checking the error returned from `Close`. It turns out that
some testcases were hitting the error but passing anyway; updating
the directives to set the necessary table version.

#### objstorage: better Writable API

Improve the writable API to more closely match its actual usage.
Simplify `Write` to not return a length, and replace `Sync` and
`Close` with `Finish` and `Abort`.

`vfs.File` can no longer be used as a `Writable`; we provide access to
the objstorage file implementation for tests.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -430,15 +430,14 @@ type compactionWritable struct {
 }
 
 // Write is part of the objstorage.Writable interface.
-func (c *compactionWritable) Write(p []byte) (n int, err error) {
-	n, err = c.Writable.Write(p)
-	if err != nil {
-		return n, err
+func (c *compactionWritable) Write(p []byte) error {
+	if err := c.Writable.Write(p); err != nil {
+		return err
 	}
 
-	*c.written += int64(n)
-	c.versions.incrementCompactionBytes(int64(n))
-	return n, err
+	*c.written += int64(len(p))
+	c.versions.incrementCompactionBytes(int64(len(p)))
+	return nil
 }
 
 type compactionKind int
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -2920,7 +2920,7 @@ func TestCompactionErrorCleanup(t *testing.T) {
 		f, err := mem.Create("ext")
 		require.NoError(t, err)
 
-		w := sstable.NewWriter(f, sstable.WriterOptions{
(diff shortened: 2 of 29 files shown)


sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, 28 of 28 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2368#-:-NPU_kzn8C8emN6R7ZGn:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`objstorage/vfs_writable.go` line 34 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2368#-NPU62R-E-75cRkk3DuF:-NPU62R0BWu0xeDRWkS1:b88zf2) ([raw file](https://github.com/cockroachdb/pebble/blob/08f7340e0c1ff6a50ebf0a69c9dec4ce2ebd346f/objstorage/vfs_writable.go#L34)):*
> ```Go
> // Write is part of the objstorage.Writable interface.
> func (w *fileBufferedWritable) Write(p []byte) error {
> 	_, err := w.bw.Write(p)
> ```

can you add
// Ignoring the length written since bufio.Writer.Write is guaranteed to return an error if the length written is < len(p).

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2368#-:-NPV5xWp3iHamdFdxMCA:b-814g8k)* status: 1 of 29 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @sumeerbhola)

___
*[`objstorage/vfs_writable.go` line 34 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2368#-NPU62R-E-75cRkk3DuF:-NPV5yB02oDPTBWs5PCC:b-fznogt) ([raw file](https://github.com/cockroachdb/pebble/blob/08f7340e0c1ff6a50ebf0a69c9dec4ce2ebd346f/objstorage/vfs_writable.go#L34)):*

can you add
// Ignoring the length written since bufio.Writer.Write is guaranteed to return an error if the length written is < len(p).
</blockquote></details>

Good idea, done.

PR #2377, opened 2023-03-01
db: improve documentation of tableNewIters

RaduBerinde:
It's much easier to navigate code when using interfaces instead of passing around `func`s (e.g. you can easily find all implementations).

Change `tableNewIters` to a new `tableIterFactory` interface. We also improve the documentation and clean up the main implementation (adding some missing error path cleanup).

### DIFF
--- a/level_iter.go
+++ b/level_iter.go
@@ -17,8 +17,15 @@ import (
 )
 
 // tableNewIters creates a new point and range-del iterator for the given file
-// number. If bytesIterated is specified, it is incremented as the given file is
-// iterated through.
+// number.
+//
+// On success, the internalIterator is not-nil and must be closed; the
+// FragmentIterator can be nil.
+// TODO(radu): always return a non-nil FragmentIterator.
+//
+// On error, the iterators are nil.
+//
+// The only (non-test) implementation of tableNewIters is tableCacheContainer.newIters().
 type tableNewIters func(
 	ctx context.Context,
 	file *manifest.FileMetadata,


bananabrick (commented):
Is there more than one implementation of this? At least outside of test code there should only be one implementation for this. I like the improved documentation, though.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2377#-:-NPTkRA01IzRytpefCW6:b-ergl4z)* status: 0 of 21 files reviewed, all discussions resolved (waiting on @jbowens)

RaduBerinde (commented):
No, there isn't - but that wasn't trivial to determine before..

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2377#-:-NPV4_TO6Ow7ctjvVMi6:bfshwcu)* status: 0 of 21 files reviewed, all discussions resolved (waiting on @jbowens)

jbowens (commented):
The funcs are often used throughout Pebble to avoid the cost of dynamic dispatch. The `Comparer` and `EventListener` are good examples. These would make sense as interfaces, but they're called so frequently that even dynamic dispatch will affect benchmarks. I'm not sure how much that argument holds for `tableNewIters`, which is called much less frequently than `Comparer`'s funcs at least. Mind checking some of the iteration microbenchmarks?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2377#-:-NPY-b8g0b-f3A4B8wRS:bh5w47s)* status: 0 of 21 files reviewed, all discussions resolved

RaduBerinde (commented):
Will do.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2377#-:-NPYPzxNErHJaAib-vIH:bhy9ot6)* status: 0 of 21 files reviewed, all discussions resolved

RaduBerinde:
> Mind checking some of the iteration microbenchmarks?

Any tips on what benchmarks I should focus on? I started running everything that contains `Iter` but it's taking a long time.

RaduBerinde:
I see some regressions in some benchmarks (and some small improvements in others!), but they are not directly explicable by the change. For example `BenchmarkIteratorSeqSeekPrefixGENotFound/skip=1/two-level=false/bloom=false/with-tombstone=true` goes from ~210ns to ~240ns ( ~17%) but `newIters` is not being called more than a couple of times during setup.

I did notice on other occasions that micro benchmarks with very small hot paths are sensitive to arbitrary changes in code layout (or at least that's the only explanation I could come up with).

Anyway, I am fine keeping it a function since it's not worth going through all the benchmark results. I will update the change to keep just the doc improvements.

jbowens (approved):
:lgtm:

> Reviewed 21 of 21 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2377#-:-NPcJfWI0MvrESVMLXCe:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`level_iter.go` line 24 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2377#-NPcIbDm410jqmAufVqb:-NPcIbDm410jqmAufVqc:b22lcxz) ([raw file](https://github.com/cockroachdb/pebble/blob/ca5bb009d92137374893c81ef5ad80d125afab51/level_iter.go#L24)):*
> ```Go
> // On success, the internalIterator is not-nil and must be closed; the
> // FragmentIterator can be nil.
> // TODO(radu): always return a non-nil FragmentIterator.
> ```

nit: I'm nervous about returning a non-nil FragmentIterator when there are no rangedels—it'll require dynamic dispatch in the common case.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2377)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`level_iter.go` line 24 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2377#-NPcIbDm410jqmAufVqb:-NPdMfdR4gLoefWeG98W:b-s3bzmw) ([raw file](https://github.com/cockroachdb/pebble/blob/ca5bb009d92137374893c81ef5ad80d125afab51/level_iter.go#L24)):*

nit: I'm nervous about returning a non-nil FragmentIterator when there are no rangedels—it'll require dynamic dispatch in the common case.
</blockquote></details>

I was thinking that we could have a `EmptyFragmentIterator` and the convention would be that you use that singleton when you know it's empty. Callers can compare the pointer against that singleton for the fast path. I don't see why it would be more expensive than a check against `nil`. The advantage would be that you don't *have* to check anything if it's not a hot path.

PR #2380, opened 2023-03-05
sstable: fix error path in writer.Close

RaduBerinde:
`Close()` records any error encountered and exits early if called again. One of the return paths was not setting `w.err`, possibly causing the panic in github.com/cockroachdb/cockroach/issues/97350.

This change fixes this and improves the structure of the function - instead of each exit path having to set `w.err`, we set it in the deferred function.

### DIFF
--- a/sstable/writer.go
+++ b/sstable/writer.go
@@ -1751,13 +1751,18 @@ func (w *Writer) Close() (err error) {
 			w.writable.Abort()
 			w.writable = nil
 		}
+		// Record any error in the writer (so we can exit early if Close is called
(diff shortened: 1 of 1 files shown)


jbowens (approved):
Nice

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2380#-:-NPrGM0R65uHO5z7Nq3F:bmu9b66)* status: 0 of 1 files reviewed, all discussions resolved

RaduBerinde:
TFTR!

PR #2382, opened 2023-03-07
objstorage: implement basic refcounting

RaduBerinde:
This commit implements the following ref tracking scheme:
 - in order to use an object, a provider must create a ref marker
   object `<object>.ref.<creator-id>.<local-filenum>`
 - when creating an object, the marker object is created after the
   backing object is created.
 - when removing an object from a provider, we remove the marker
   object and then check if there are other markers remaining; if not, we
   delete the backing object.

When attaching an object, the "target" provider creates its ref object
then checks that the "source" provider's ref still exists.

Ref tracking can be enabled on a per-object basis. This will be useful
with performance experiments and could also be necessary for the
progressive restore usecase.


sumeerbhola (changes requested):
looks good

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2382#-:-NQ2VYf001rQxWpUZxNH:b-47fafm)* status: all files reviewed, 2 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`objstorage/shared.go` line 145 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NQ2RHZx64AlsIkLZizb:-NQ2RHZx64AlsIkLZizc:b-q3zz01) ([raw file](https://github.com/cockroachdb/pebble/blob/88ebb3fa8ce8b864d70eaf32fed7ef2d81f4802d/objstorage/shared.go#L145)):*
> ```Go
> 	)
> }
> ```

worth having a code comment stating the ref path, <object>.ref.<referencing-creator-id>
___
*[`objstorage/shared_backing.go` line 125 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NQ2TdZD0so6CzZ_Cptf:-NQ2TdZD0so6CzZ_Cptg:b-v78pl7) ([raw file](https://github.com/cockroachdb/pebble/blob/88ebb3fa8ce8b864d70eaf32fed7ef2d81f4802d/objstorage/shared_backing.go#L125)):*
> ```Go
> 	// Create the reference marker objects.
> 	// TODO(radu): check that the actual objects exist so we don't error out later
> 	// TODO(radu): parallelize this.
> ```

Should we scope the disabling of ref tracking to only apply to such imported files? Since after the restore we will continue creating shared files that should be ref counted. Actually, it seems we should specify this as a parameter in `AttachSharedObjects`, since the cluster will have been creating shared files before the restore. Which means we would need to remember whether we are ref counting or not on a per file level in the `Catalog`,

I think this attach/import needs to handle the race if the exporter decides to give up on the export and concurrently tries to unref and delete. I think one scheme that works is for the exporter to specify its creator-id in `SharedObjectBacking`, if the file is being ref-counted, so the importer/attacher can create its own ref, then check that the exporter ref still exists, and only then declare the attach successful.

RaduBerinde:
*[`objstorage/shared_backing.go` line 125 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NQ2TdZD0so6CzZ_Cptf:-NQ2acLdAj0y4ai4ymHi:b1xw9wu) ([raw file](https://github.com/cockroachdb/pebble/blob/88ebb3fa8ce8b864d70eaf32fed7ef2d81f4802d/objstorage/shared_backing.go#L125)):*

Should we scope the disabling of ref tracking to only apply to such imported files? Since after the restore we will continue creating shared files that should be ref counted. Actually, it seems we should specify this as a parameter in `AttachSharedObjects`, since the cluster will have been creating shared files before the restore. Which means we would need to remember whether we are ref counting or not on a per file level in the `Catalog`,

I think this attach/import needs to handle the race if the exporter decides to give up on the export and concurrently tries to unref and delete. I think one scheme that works is for the exporter to specify its creator-id in `SharedObjectBacking`, if the file is being ref-counted, so the importer/attacher can create its own ref, then check that the exporter ref still exists, and only then declare the attach successful.
</blockquote></details>

My thinking was that the restored cluster would never create new shared objects, just local. But I like the flexibility of specifying this per-object, I will give it some thought.

Even for simple rebalancing, there is a similar race between unref and attach. I will think about an API to allow control over when an unref can happen once we extract a backing.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2382)* status: all files reviewed, 2 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`objstorage/shared_backing.go` line 125 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NQ2TdZD0so6CzZ_Cptf:-NQAzojDCjMcxe_tz_P4:b115xyi) ([raw file](https://github.com/cockroachdb/pebble/blob/88ebb3fa8ce8b864d70eaf32fed7ef2d81f4802d/objstorage/shared_backing.go#L125)):*
> Even for simple rebalancing, there is a similar race between unref and attach. I will think about an API to allow control over when an unref can happen once we extract a backing.

Yep, I was thinking of the rebalancing case in the comment about the race. I don't see a way to completely control this on the exporter side -- eventually there has to be a timeout. Maybe the code can be nicely structured that the timeout causes a `exportHandle.unref` on the exporter, but there is still a race with the importer.

itsbilal (approved):
This is :lgtm: to me, though I realize that the per-object `DisableRefCounting` optionality isn't in yet. I'm good with that coming in a future PR, or you're welcome to make that change in this PR and re-request review. Whatever works for you.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2382#-:-NRA-cm9-eygTaiRiOI5:b6oqgt3)* status: all files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`objstorage/shared.go` line 203 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NR9zW9D0jUzHU9Tz_za:-NR9zW9D0jUzHU9Tz_zb:b-j5uxk2) ([raw file](https://github.com/cockroachdb/pebble/blob/88ebb3fa8ce8b864d70eaf32fed7ef2d81f4802d/objstorage/shared.go#L203)):*
> ```Go
> 	}
> 	if !p.st.Shared.DisableRefTracking {
> 		// First, verify we have a reference on this object.
> ```

Do we need this? I can see this being a little unnecessary, if it turns a lot of reads into two sequential s3 requests, up from one. I'm fine with putting this behind an `invariants.Enabled` flag in the name of performance.

RaduBerinde (commented):
TFTRs!

I pushed a new iteration. I implemented:
 - keeping track of cleanup method (ref tracking enabled or not) per object
 - "shared backing handle" API which prevents object unref until handle Close(). This is incomplete - we're missing a mechanism to perform the delayed unref after the handle is closed (currently we just leak the object).
 - checking "originator" reference on attach

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2382#-:-NRdmQlI56d_T5mqiN1b:b-yb76e7)* status: 0 of 24 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`objstorage/shared.go` line 145 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NQ2RHZx64AlsIkLZizb:-NRKUND59lOlTk0QvRtz:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/88ebb3fa8ce8b864d70eaf32fed7ef2d81f4802d/objstorage/shared.go#L145)):*

worth having a code comment stating the ref path, <object>.ref.<referencing-creator-id>
</blockquote></details>

Done.
___
*[`objstorage/shared.go` line 203 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NR9zW9D0jUzHU9Tz_za:-NRKUODU2nMOGiJzUz-h:b-23dtn2) ([raw file](https://github.com/cockroachdb/pebble/blob/88ebb3fa8ce8b864d70eaf32fed7ef2d81f4802d/objstorage/shared.go#L203)):*

Do we need this? I can see this being a little unnecessary, if it turns a lot of reads into two sequential s3 requests, up from one. I'm fine with putting this behind an `invariants.Enabled` flag in the name of performance.
</blockquote></details>

Done. Had to go through some hoops so that the provider test output doesn't depend on `invariants.Enabled`.

sumeerbhola (changes requested):
> Reviewed 8 of 24 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2382)* status: 8 of 24 files reviewed, 7 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`objstorage/objstorage.go` line 158 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NRjAVDY407UaPPKazGJ:-NRjAVDY407UaPPKazGK:bd9c7c9) ([raw file](https://github.com/cockroachdb/pebble/blob/86423be3ff0c884b1ff43879ab694595e6f88992/objstorage/objstorage.go#L158)):*
> ```Go
> 
> 	// SharedCleanupMethod is used for the object when it is created on shared storage.
> 	// The default (zero) value is SharedNoCleanup.
> ```

In production we will always use `SharedRefTracking` when creating, yes?
___
*[`objstorage/objstorageprovider/provider.go` line 376 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NRjBqvw439pVC2CXZd8:-NRjBqvw439pVC2CXZd9:b3v432u) ([raw file](https://github.com/cockroachdb/pebble/blob/86423be3ff0c884b1ff43879ab694595e6f88992/objstorage/objstorageprovider/provider.go#L376)):*
> ```Go
> 		delete(p.mu.protectedObjects, fileNum)
> 		// TODO(radu): keep track of whether the provider tried to unref the object
> 		// and unref it now.
> ```

I was expecting `knownObjects` to have a map value that is a `<ObjectMetadata, refCount>` pair. Usually the refcount will be be 1, but such protection would increment it. Then we have a simple way to decide when to do `sharedUnref`.
___
*[`objstorage/objstorageprovider/shared.go` line 164 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NRjErSq-cV9LKLeAIEX:-NRjErSq-cV9LKLeAIEY:b-gws0m) ([raw file](https://github.com/cockroachdb/pebble/blob/86423be3ff0c884b1ff43879ab694595e6f88992/objstorage/objstorageprovider/shared.go#L164)):*
> ```Go
> 	return fmt.Sprintf(
> 		"%s-%s.ref.%s",
> 		meta.Shared.CreatorID, base.MakeFilename(meta.FileType, meta.Shared.CreatorFileNum), refID,
> ```

A Pebble instance may attach the same shared file multiple times, and it will use different local filenums for it. At the Pebble manifest level the refcounting will for these local filenums, and `provider.Remove` will be called when that local filenum is no longer needed. Don't we need that local filenum (`meta.FileNum`) in this ref path so we only remove that ref? We could also share that ref by doing more in-memory refcounting but I worry about race conditions and the likelihood of having multiple refs is low anyway (so we could ignore the overhead).
___
*[`objstorage/objstorageprovider/shared_backing.go` line 179 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NRjD_vS6NHtdGjn4us2:-NRjD_vS6NHtdGjn4us3:b-olefp2) ([raw file](https://github.com/cockroachdb/pebble/blob/86423be3ff0c884b1ff43879ab694595e6f88992/objstorage/objstorageprovider/shared_backing.go#L179)):*
> ```Go
> 		}
> 		if err := p.sharedCreateRef(meta); err != nil {
> 			// TODO(radu): clean up already created references.
> ```

references created by this provider?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2382)* status: 8 of 24 files reviewed, 7 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`objstorage/objstorageprovider/provider.go` line 376 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NRjBqvw439pVC2CXZd8:-NRjm11p8PStGWG-loYA:b-mxh3ss) ([raw file](https://github.com/cockroachdb/pebble/blob/86423be3ff0c884b1ff43879ab694595e6f88992/objstorage/objstorageprovider/provider.go#L376)):*

I was expecting `knownObjects` to have a map value that is a `<ObjectMetadata, refCount>` pair. Usually the refcount will be be 1, but such protection would increment it. Then we have a simple way to decide when to do `sharedUnref`.
</blockquote></details>

Currently when we remove the object, we remove it from the `knownObjects` map, even if the object is still protected. This is right IMO, the protection should only affect the underlying ref tracking and not be visible in the external API. So we already have an easy way of knowing when to unref - we can look for the object in the `knownObjects` map. With your proposed solution, we'd also need to keep a flag of whether to "hide" the object from regular lookups.
___
*[`objstorage/objstorageprovider/shared.go` line 164 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NRjErSq-cV9LKLeAIEX:-NRjkoCu9_n4mkGPUKrZ:b-wx4lv0) ([raw file](https://github.com/cockroachdb/pebble/blob/86423be3ff0c884b1ff43879ab694595e6f88992/objstorage/objstorageprovider/shared.go#L164)):*

A Pebble instance may attach the same shared file multiple times, and it will use different local filenums for it. At the Pebble manifest level the refcounting will for these local filenums, and `provider.Remove` will be called when that local filenum is no longer needed. Don't we need that local filenum (`meta.FileNum`) in this ref path so we only remove that ref? We could also share that ref by doing more in-memory refcounting but I worry about race conditions and the likelihood of having multiple refs is low anyway (so we could ignore the overhead).
</blockquote></details>

Hm, good point, I haven't thought of supporting attaching the same shared object multiple times. I'll give this some thought.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2382)* status: 8 of 24 files reviewed, 7 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`objstorage/objstorageprovider/provider.go` line 376 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NRjBqvw439pVC2CXZd8:-NRmiV6oAaQQU7QLlfEe:b-1nhv4e) ([raw file](https://github.com/cockroachdb/pebble/blob/86423be3ff0c884b1ff43879ab694595e6f88992/objstorage/objstorageprovider/provider.go#L376)):*
> This is right IMO, the protection should only affect the underlying ref tracking and not be visible in the external API. With your proposed solution, we'd also need to keep a flag of whether to "hide" the object from regular lookups.

This makes sense

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2382)* status: 8 of 24 files reviewed, 7 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`objstorage/objstorage.go` line 158 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NRjAVDY407UaPPKazGJ:-NRnlaEM0LV6PHr-HcMd:b-9o7umy) ([raw file](https://github.com/cockroachdb/pebble/blob/86423be3ff0c884b1ff43879ab694595e6f88992/objstorage/objstorage.go#L158)):*

In production we will always use `SharedRefTracking` when creating, yes?
</blockquote></details>

Good point, swapped the default.
___
*[`objstorage/objstorageprovider/provider.go` line 376 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NRjBqvw439pVC2CXZd8:-NRnloY8781tDBsXs1Wl:by960dk) ([raw file](https://github.com/cockroachdb/pebble/blob/86423be3ff0c884b1ff43879ab694595e6f88992/objstorage/objstorageprovider/provider.go#L376)):*

> This is right IMO, the protection should only affect the underlying ref tracking and not be visible in the external API. With your proposed solution, we'd also need to keep a flag of whether to "hide" the object from regular lookups.

This makes sense
</blockquote></details>

Updated the comment.
___
*[`objstorage/objstorageprovider/shared.go` line 164 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NRjErSq-cV9LKLeAIEX:-NRnmJlA5CWfiqTw4EXI:b-dsq0sr) ([raw file](https://github.com/cockroachdb/pebble/blob/86423be3ff0c884b1ff43879ab694595e6f88992/objstorage/objstorageprovider/shared.go#L164)):*

Hm, good point, I haven't thought of supporting attaching the same shared object multiple times. I'll give this some thought.
</blockquote></details>

Done, I used the append local filenum approach.
___
*[`objstorage/objstorageprovider/shared_backing.go` line 179 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2382#-NRjD_vS6NHtdGjn4us2:-NRnm6IlBMV8Jy3jHWNH:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/86423be3ff0c884b1ff43879ab694595e6f88992/objstorage/objstorageprovider/shared_backing.go#L179)):*

references created by this provider?
</blockquote></details>

Done.

sumeerbhola (approved):
:lgtm: 

> Reviewed 12 of 24 files at r2, 10 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2382#-:-NRtYTm2AH7PKiq6i_zW:bg7igkv)* status: all files reviewed, 7 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

PR #2384, opened 2023-03-08
manifest: refactor BulkVersionEdit for consistency

bananabrick:
Makes bulk version edit accumulation consistent with respect
to files being added and removed from a level; a file added
and deleted from the same level in the same bulk version edit
is no longer tracked in BulkVersionEdit.Added/Deleted.

We also explicitly write down the invariants which aid in proving
the correctness of the new approach.

We also separate out `BulkVersionEdit.Apply` into
`BulkVersionEdit.Apply` and `AccumulateAndApplySingleVE`. It's
easier to reason about accumulation/apply/zombie tables if we
have stricter invariants about when these functions are called.

### DIFF
--- a/internal/keyspan/level_iter_test.go
+++ b/internal/keyspan/level_iter_test.go
@@ -303,8 +303,12 @@ func TestLevelIterEquivalence(t *testing.T) {
 			}
 			// Add all the fileMetadatas to L6.
 			b := &manifest.BulkVersionEdit{}
-			b.Added[6] = metas
-			v, _, err := b.Apply(nil, base.DefaultComparer.Compare, base.DefaultFormatter, 0, 0)
+			amap := make(map[base.FileNum]*manifest.FileMetadata)
+			for i := range metas {
+				amap[metas[i].FileNum] = metas[i]
+			}
+			b.Added[6] = amap
+			v, err := b.Apply(nil, base.DefaultComparer.Compare, base.DefaultFormatter, 0, 0, nil)
 			require.NoError(t, err)
 			levelIter.Init(
 				SpanIterOptions{}, base.DefaultComparer.Compare, tableNewIters,
@@ -436,8 +440,12 @@ func TestLevelIter(t *testing.T) {
 					return NewIter(base.DefaultComparer.Compare, spans), nil
 				}
 				b := &manifest.BulkVersionEdit{}
-				b.Added[6] = metas
-				v, _, err := b.Apply(nil, base.DefaultComparer.Compare, base.DefaultFormatter, 0, 0)
+				amap := make(map[base.FileNum]*manifest.FileMetadata)
+				for i := range metas {
+					amap[metas[i].FileNum] = metas[i]
+				}
+				b.Added[6] = amap
+				v, err := b.Apply(nil, base.DefaultComparer.Compare, base.DefaultFormatter, 0, 0, nil)
 				require.NoError(t, err)
 				iter = NewLevelIter(
 					SpanIterOptions{}, base.DefaultComparer.Compare,
--- a/internal/manifest/l0_sublevels_test.go
+++ b/internal/manifest/l0_sublevels_test.go
@@ -50,7 +50,7 @@ func readManifest(filename string) (*Version, error) {
 		if err := bve.Accumulate(&ve); err != nil {
 			return nil, err
 		}
-		if v, _, err = bve.Apply(v, base.DefaultComparer.Compare, base.DefaultFormatter, 10<<20, 32000); err != nil {
+		if v, err = bve.Apply(v, base.DefaultComparer.Compare, base.DefaultFormatter, 10<<20, 32000, nil); err != nil {
 			return nil, err
 		}
 	}
--- a/internal/manifest/testdata/version_edit_apply
+++ b/internal/manifest/testdata/version_edit_apply
@@ -166,3 +166,26 @@ edit
    2
 ----
 pebble: file deleted L0.000002 before it was inserted
+
+apply
+ L0
+  1:[a#1,SET-b#2,SET]
+edit
+ delete
+  L0
+   1
+ add
+  L2
+  1:[a#1,SET-b#2,SET]
+  4:[c#3,SET-d#4,SET]
+  5:[s#3,SET-z#4,SET]
+edit
+  delete
+    L2
+     1
+    L2
+     4
+----
+2:
+  000005:[s#3,SET-z#4,SET]
+zombies []
--- a/internal/manifest/version_edit.go
+++ b/internal/manifest/version_edit.go
@@ -9,6 +9,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"io"
+	"sort"
 	"sync/atomic"
 
 	"github.com/cockroachdb/errors"
@@ -536,8 +537,23 @@ func (e versionEditEncoder) writeUvarint(u uint64) {
 
 // BulkVersionEdit summarizes the files added and deleted from a set of version
 // edits.
+//
+// INVARIANTS:
+// No file can be added to a level more than once. This is true globally, and
+// also true for all of the calls to Accumulate for a single bulk version edit.
+//
+// No file can be removed from a level more than once. This is true globally,
+// and also true for all of the calls to Accumulate for a single bulk version
+// edit.
+//
+// A file must not be added and removed from a given level in the same version
+// edit.
+//
+// A file that is being removed from a level must have been added to that level
+// before (in a prior version edit). Note that a given file can be deleted from
+// a level and added to another level in a single version edit
 type BulkVersionEdit struct {
-	Added   [NumLevels][]*FileMetadata
+	Added   [NumLevels]map[base.FileNum]*FileMetadata
 	Deleted [NumLevels]map[base.FileNum]*FileMetadata
 
 	// AddedByFileNum maps file number to file metadata for all added files
@@ -558,6 +574,19 @@ type BulkVersionEdit struct {
 
 // Accumulate adds the file addition and deletions in the specified version
 // edit to the bulk edit's internal state.
+//
+// INVARIANTS:
+// If a file is added to a given level in a call to Accumulate and then removed
+// from that level in a subsequent call, the file will not be present in the
+// resulting BulkVersionEdit.Deleted for that level.
+//
+// After accumulation of version edits, the bulk version edit may have
+// information about a file which has been deleted from a level, but it may
+// not have information about the same file added to the same level. The add
+// could've occurred as part of a previous bulk version edit. In this case,
+// the deleted file must be present in BulkVersionEdit.Deleted, at the end
+// of the accumulation, because we need to decrease the refcount of the
+// deleted file in Apply.
 func (b *BulkVersionEdit) Accumulate(ve *VersionEdit) error {
 	for df, m := range ve.DeletedFiles {
 		dmap := b.Deleted[df.Level]
@@ -579,7 +608,12 @@ func (b *BulkVersionEdit) Accumulate(ve *VersionEdit) error {
 		if m.MarkedForCompaction {
 			b.MarkedForCompactionCountDiff--
 		}
-		dmap[df.FileNum] = m
+		if _, ok := b.Added[df.Level][df.FileNum]; !ok {
+			dmap[df.FileNum] = m
+		} else {
+			// Present in b.Added for the same level.
+			delete(b.Added[df.Level], df.FileNum)
+		}
 	}
 
 	for _, nf := range ve.NewFiles {
@@ -590,7 +624,10 @@ func (b *BulkVersionEdit) Accumulate(ve *VersionEdit) error {
 				return base.CorruptionErrorf("pebble: file deleted L%d.%s before it was inserted", nf.Level, nf.Meta.FileNum)
 			}
 		}
-		b.Added[nf.Level] = append(b.Added[nf.Level], nf.Meta)
+		if b.Added[nf.Level] == nil {
+			b.Added[nf.Level] = make(map[base.FileNum]*FileMetadata)
+		}
+		b.Added[nf.Level][nf.Meta.FileNum] = nf.Meta
 		if b.AddedByFileNum != nil {
 			b.AddedByFileNum[nf.Meta.FileNum] = nf.Meta
 		}
@@ -601,30 +638,59 @@ func (b *BulkVersionEdit) Accumulate(ve *VersionEdit) error {
 	return nil
 }
 
-// Apply applies the delta b to the current version to produce a new
-// version. The new version is consistent with respect to the comparer cmp.
+// AccumulateAndApplySingleVE should be called if a single version edit is to be
+// applied to the provided curr Version and if the caller needs to update the
+// versionSet.zombieTables map. This function exists separately from
+// BulkVersionEdit.Apply because it is easier to reason about properties
+// regarding BulkVersionedit.Accumulate/Apply and zombie table generation, if we
+// know that exactly one version edit is being accumulated.
 //
-// curr may be nil, which is equivalent to a pointer to a zero version.
-//
-// On success, a map of zombie files containing the file numbers and sizes of
-// deleted files is returned. These files are considered zombies because they
-// are no longer referenced by the returned Version, but cannot be deleted from
-// disk as they are still in use by the incoming Version.
-func (b *BulkVersionEdit) Apply(
+// Invariant: Any file being deleted through ve must belong to the curr Version.
+// We can't have a delete for some arbitrary file which does not exist in curr.
+func AccumulateAndApplySingleVE(
+	ve *VersionEdit,
 	curr *Version,
 	cmp Compare,
 	formatKey base.FormatKey,
 	flushSplitBytes int64,
 	readCompactionRate int64,
 ) (_ *Version, zombies map[base.FileNum]uint64, _ error) {
-	addZombie := func(fileNum base.FileNum, size uint64) {
-		if zombies == nil {
-			zombies = make(map[base.FileNum]uint64)
-		}
-		zombies[fileNum] = size
+	var b BulkVersionEdit
+	err := b.Accumulate(ve)
+	if err != nil {
+		return nil, nil, err
 	}
-	// The remove zombie function is used to handle tables that are moved from
-	// one level to another during a version edit (i.e. a "move" compaction).
+	zombies = make(map[base.FileNum]uint64)
+	v, err := b.Apply(curr, cmp, formatKey, flushSplitBytes, readCompactionRate, zombies)
+	if err != nil {
+		return nil, nil, err
+	}
+	return v, zombies, nil
+}
+
+// Apply applies the delta b to the current version to produce a new
+// version. The new version is consistent with respect to the comparer cmp.
+//
+// curr may be nil, which is equivalent to a pointer to a zero version.
+//
+// On success, if a non-nil zombies map is provided to Apply, the map is updated
+// with file numbers and files sizes of deleted files. These files are
+// considered zombies because they are no longer referenced by the returned
+// Version, but cannot be deleted from disk as they are still in use by the
+// incoming Version.
+func (b *BulkVersionEdit) Apply(
+	curr *Version,
+	cmp Compare,
+	formatKey base.FormatKey,
+	flushSplitBytes int64,
+	readCompactionRate int64,
+	zombies map[base.FileNum]uint64,
+) (*Version, error) {
+	addZombie := func(fileNum base.FileNum, size uint64) {
+		if zombies != nil {
+			zombies[fileNum] = size
+		}
+	}
 	removeZombie := func(fileNum base.FileNum) {
 		if zombies != nil {
 			delete(zombies, fileNum)
@@ -639,7 +705,7 @@ func (b *BulkVersionEdit) Apply(
 	}
 	v.Stats.MarkedForCompaction += b.MarkedForCompactionCountDiff
 	if v.Stats.MarkedForCompaction < 0 {
-		return nil, nil, base.CorruptionErrorf("pebble: version marked for compaction count negative")
+		return nil, base.CorruptionErrorf("pebble: version marked for compaction count negative")
 	}
 
 	for level := range v.Levels {
@@ -660,7 +726,7 @@ func (b *BulkVersionEdit) Apply(
 				// Initialize L0Sublevels.
 				if curr == nil || curr.L0Sublevels == nil {
 					if err := v.InitL0Sublevels(cmp, formatKey, flushSplitBytes); err != nil {
-						return nil, nil, errors.Wrap(err, "pebble: internal error")
+						return nil, errors.Wrap(err, "pebble: internal error")
 					}
 				} else {
 					v.L0Sublevels = curr.L0Sublevels
@@ -673,18 +739,20 @@ func (b *BulkVersionEdit) Apply(
 		// Some edits on this level.
 		lm := &v.Levels[level]
 		lmRange := &v.RangeKeyLevels[level]
-		addedFiles := b.Added[level]
-		deletedMap := b.Deleted[level]
-		if n := v.Levels[level].Len() + len(addedFiles); n == 0 {
-			return nil, nil, base.CorruptionErrorf(
+
+		addedFilesMap := b.Added[level]
+		deletedFilesMap := b.Deleted[level]
+		if n := v.Levels[level].Len() + len(addedFilesMap); n == 0 {
+			return nil, base.CorruptionErrorf(
 				"pebble: internal error: No current or added files but have deleted files: %d",
-				errors.Safe(len(deletedMap)))
+				errors.Safe(len(deletedFilesMap)))
 		}
 
-		// NB: addedFiles may be empty and it also is not necessarily
-		// internally consistent: it does not reflect deletions in deletedMap.
+		// NB: addedFilesMap may be empty. If a file is present in addedFilesMap
+		// for a level, it won't be present in deletedFilesMap for the same
+		// level.
 
-		for _, f := range deletedMap {
+		for _, f := range deletedFilesMap {
 			addZombie(f.FileNum, f.Size)
 			if obsolete := v.Levels[level].tree.Delete(f); obsolete {
 				// Deleting a file from the B-Tree may decrement its
@@ -692,7 +760,7 @@ func (b *BulkVersionEdit) Apply(
 				// previous level's B-Tree, this should never result in a
 				// file's reference count dropping to zero.
 				err := errors.Errorf("pebble: internal error: file L%d.%s obsolete during B-Tree removal", level, f.FileNum)
-				return nil, nil, err
+				return nil, err
 			}
 			if f.HasRangeKeys {
 				if obsolete := v.RangeKeyLevels[level].tree.Delete(f); obsolete {
@@ -701,19 +769,24 @@ func (b *BulkVersionEdit) Apply(
 					// previous level's B-Tree, this should never result in a
 					// file's reference count dropping to zero.
 					err := errors.Errorf("pebble: internal error: file L%d.%s obsolete during range-key B-Tree removal", level, f.FileNum)
-					return nil, nil, err
+					return nil, err
 				}
 			}
 		}
 
+		addedFiles := make([]*FileMetadata, 0, len(addedFilesMap))
+		for _, f := range addedFilesMap {
+			addedFiles = append(addedFiles, f)
+		}
+		// Sort addedFiles by file number. This isn't necessary, but tests which
+		// replay invalid manifests check the error output, and the error output
+		// depends on the order in which files are added to the btree.
+		sort.Slice(addedFiles, func(i, j int) bool {
+			return addedFiles[i].FileNum < addedFiles[j].FileNum
+		})
+
 		var sm, la *FileMetadata
 		for _, f := range addedFiles {
-			if _, ok := deletedMap[f.FileNum]; ok {
-				// Already called addZombie on this file in the preceding
-				// loop, so we don't need to do it here.
-				continue
-			}
-
 			// NB: allowedSeeks is used for read triggered compactions. It is set using
 			// Options.Experimental.ReadCompactionRate which defaults to 32KB.
 			var allowedSeeks int64
@@ -728,12 +801,12 @@ func (b *BulkVersionEdit) Apply(
 
 			err := lm.tree.Insert(f)
 			if err != nil {
-				return nil, nil, errors.Wrap(err, "pebble")
+				return nil, errors.Wrap(err, "pebble")
 			}
 			if f.HasRangeKeys {
 				err = lmRange.tree.Insert(f)
 				if err != nil {
-					return nil, nil, errors.Wrap(err, "pebble")
+					return nil, errors.Wrap(err, "pebble")
 				}
 			}
 			removeZombie(f.FileNum)
@@ -748,27 +821,26 @@ func (b *BulkVersionEdit) Apply(
 		}
 
 		if level == 0 {
-			if curr != nil && curr.L0Sublevels != nil && len(deletedMap) == 0 {
+			if curr != nil && curr.L0Sublevels != nil && len(deletedFilesMap) == 0 {
 				// Flushes and ingestions that do not delete any L0 files do not require
 				// a regeneration of L0Sublevels from scratch. We can instead generate
 				// it incrementally.
 				var err error
 				// AddL0Files requires addedFiles to be sorted in seqnum order.
-				addedFiles = append([]*FileMetadata(nil), addedFiles...)
 				SortBySeqNum(addedFiles)
 				v.L0Sublevels, err = curr.L0Sublevels.AddL0Files(addedFiles, flushSplitBytes, &v.Levels[0])
 				if errors.Is(err, errInvalidL0SublevelsOpt) {
 					err = v.InitL0Sublevels(cmp, formatKey, flushSplitBytes)
 				}
 				if err != nil {
-					return nil, nil, errors.Wrap(err, "pebble: internal error")
+					return nil, errors.Wrap(err, "pebble: internal error")
 				}
 				v.L0SublevelFiles = v.L0Sublevels.Levels
 			} else if err := v.InitL0Sublevels(cmp, formatKey, flushSplitBytes); err != nil {
-				return nil, nil, errors.Wrap(err, "pebble: internal error")
+				return nil, errors.Wrap(err, "pebble: internal error")
 			}
 			if err := CheckOrdering(cmp, formatKey, Level(0), v.Levels[level].Iter()); err != nil {
-				return nil, nil, errors.Wrap(err, "pebble: internal error")
+				return nil, errors.Wrap(err, "pebble: internal error")
 			}
 			continue
 		}
@@ -789,9 +861,9 @@ func (b *BulkVersionEdit) Apply(
 				}
 			})
 			if err := CheckOrdering(cmp, formatKey, Level(level), check.Iter()); err != nil {
-				return nil, nil, errors.Wrap(err, "pebble: internal error")
+				return nil, errors.Wrap(err, "pebble: internal error")
 			}
 		}
 	}
-	return v, zombies, nil
+	return v, nil
 }
--- a/internal/manifest/version_edit_test.go
+++ b/internal/manifest/version_edit_test.go
@@ -315,6 +315,7 @@ func TestVersionEditApply(t *testing.T) {
 		return m, nil
 	}
 
+	// TODO(bananabrick): Improve the parsing logic in this test.
 	datadriven.RunTest(t, "testdata/version_edit_apply",
 		func(t *testing.T, d *datadriven.TestData) string {
 			switch d.Cmd {
@@ -323,7 +324,7 @@ func TestVersionEditApply(t *testing.T) {
 				// avoid repeating it, and make it the inverse of
 				// Version.DebugString().
 				var v *Version
-				ve := &VersionEdit{}
+				var veList []*VersionEdit
 				isVersion := true
 				isDelete := true
 				var level int
@@ -334,6 +335,7 @@ func TestVersionEditApply(t *testing.T) {
 					switch data {
 					case "edit":
 						isVersion = false
+						veList = append(veList, &VersionEdit{})
 					case "delete":
 						isVersion = false
 						isDelete = true
@@ -346,6 +348,10 @@ func TestVersionEditApply(t *testing.T) {
 							return err.Error()
 						}
 					default:
+						var ve *VersionEdit
+						if len(veList) > 0 {
+							ve = veList[len(veList)-1]
+						}
 						if isVersion || !isDelete {
 							meta, err := parseMeta(data)
 							if err != nil {
@@ -386,21 +392,28 @@ func TestVersionEditApply(t *testing.T) {
 
 				bve := BulkVersionEdit{}
 				bve.AddedByFileNum = make(map[base.FileNum]*FileMetadata)
-				if err := bve.Accumulate(ve); err != nil {
-					return err.Error()
+				for _, ve := range veList {
+					if err := bve.Accumulate(ve); err != nil {
+						return err.Error()
+					}
 				}
-				newv, zombies, err := bve.Apply(v, base.DefaultComparer.Compare, base.DefaultFormatter, 10<<20, 32000)
+				zombies := make(map[base.FileNum]uint64)
+				newv, err := bve.Apply(v, base.DefaultComparer.Compare, base.DefaultFormatter, 10<<20, 32000, zombies)
 				if err != nil {
 					return err.Error()
 				}
 
 				zombieFileNums := make([]base.FileNum, 0, len(zombies))
-				for fileNum := range zombies {
-					zombieFileNums = append(zombieFileNums, fileNum)
+				if len(veList) == 1 {
+					// Only care about zombies if a single version edit was
+					// being applied.
+					for fileNum := range zombies {
+						zombieFileNums = append(zombieFileNums, fileNum)
+					}
+					sort.Slice(zombieFileNums, func(i, j int) bool {
+						return zombieFileNums[i] < zombieFileNums[j]
+					})
 				}
-				sort.Slice(zombieFileNums, func(i, j int) bool {
-					return zombieFileNums[i] < zombieFileNums[j]
-				})
 
 				return fmt.Sprintf("%szombies %d\n", newv, zombieFileNums)
 
--- a/replay/replay.go
+++ b/replay/replay.go
@@ -665,11 +665,12 @@ func (r *Runner) prepareWorkloadSteps(ctx context.Context) error {
 	}
 	currentVersion := func() (*manifest.Version, error) {
 		var err error
-		v, _, err = bve.Apply(v,
+		v, err = bve.Apply(v,
 			r.Opts.Comparer.Compare,
 			r.Opts.Comparer.FormatKey,
 			r.Opts.FlushSplitBytes,
-			r.Opts.Experimental.ReadCompactionRate)
+			r.Opts.Experimental.ReadCompactionRate,
+			nil /* zombies */)
 		bve = manifest.BulkVersionEdit{AddedByFileNum: bve.AddedByFileNum}
 		return v, err
 	}
--- a/tool/db.go
+++ b/tool/db.go
@@ -483,7 +483,7 @@ func (d *dbT) runProperties(cmd *cobra.Command, args []string) {
 				d.fmtValue.setForComparer(ve.ComparerName, d.comparers)
 			}
 		}
-		v, _, err := bve.Apply(nil /* version */, cmp.Compare, d.fmtKey.fn, d.opts.FlushSplitBytes, d.opts.Experimental.ReadCompactionRate)
+		v, err := bve.Apply(nil /* version */, cmp.Compare, d.fmtKey.fn, d.opts.FlushSplitBytes, d.opts.Experimental.ReadCompactionRate, nil /* zombies */)
 		if err != nil {
 			return err
 		}
--- a/tool/manifest.go
+++ b/tool/manifest.go
@@ -234,7 +234,7 @@ func (m *manifestT) runDump(cmd *cobra.Command, args []string) {
 			}
 
 			if cmp != nil {
-				v, _, err := bve.Apply(nil /* version */, cmp.Compare, m.fmtKey.fn, 0, m.opts.Experimental.ReadCompactionRate)
+				v, err := bve.Apply(nil /* version */, cmp.Compare, m.fmtKey.fn, 0, m.opts.Experimental.ReadCompactionRate, nil /* zombies */)
 				if err != nil {
 					fmt.Fprintf(stdout, "%s\n", err)
 					return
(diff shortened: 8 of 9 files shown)


bananabrick:
Not sure why tests don't fail locally.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`internal/manifest/version_edit.go` line 541 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NPzD99bDsuVTVmUPr7l:-NPzD99bDsuVTVmUPr7m:b-uqrr01) ([raw file](https://github.com/cockroachdb/pebble/blob/387a441b17f446b0af0636b87f7e8e4d484c4e7a/internal/manifest/version_edit.go#L541)):*
> ```Go
> //
> // INVARIANTS:
> // - No file can be added to a level more than once. This is true globally, and
> ```

These assume that intra L0 move compactions aren't possible, which I believe is the case.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`internal/manifest/version_edit.go` line 541 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NPzD99bDsuVTVmUPr7l:-NPzE4yZ3Lhb5g9qtwL1:bpbplxv) ([raw file](https://github.com/cockroachdb/pebble/blob/387a441b17f446b0af0636b87f7e8e4d484c4e7a/internal/manifest/version_edit.go#L541)):*

These assume that intra L0 move compactions aren't possible, which I believe is the case.
</blockquote></details>

I don't know if that was a purposeful choice, but I think intra L0 compactions are only picked if more than one file is part of the compaction, and move compactions are only picked if there is exactly 1 file in the start level of the compaction.

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @bananabrick)

___
*[`internal/manifest/version_edit.go` line 541 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NPzD99bDsuVTVmUPr7l:-NQ0sBlW68KA_yygZYBj:b-6ybbiv) ([raw file](https://github.com/cockroachdb/pebble/blob/387a441b17f446b0af0636b87f7e8e4d484c4e7a/internal/manifest/version_edit.go#L541)):*
> These assume that intra L0 move compactions aren't possible, which I believe is the case.

Correct. We use `minCompactionDepth=4` for intra-L0 compactions. But probably worth checking here and crashing if it is violated.

> I don't know if that was a purposeful choice, but I think intra L0 compactions are only picked if more than one file is part of the compaction

If the compaction depth is 1 then it isn't going to reduce the read-amp in L0, so it was purposeful.
___
*[`internal/manifest/version_edit.go` line 556 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ0uxHo1DQgGtwaMJtv:-NQ0uxHo1DQgGtwaMJtw:b35fdq5) ([raw file](https://github.com/cockroachdb/pebble/blob/d52333ce1a349d2bc308c2c1e32a726adbef415e/internal/manifest/version_edit.go#L556)):*
> ```Go
> type BulkVersionEdit struct {
> 	Added   [NumLevels]map[base.FileNum]*FileMetadata
> 	Deleted [NumLevels]map[base.FileNum]*FileMetadata
> ```

Does `Deleted` need to be a map. See my comment below.
___
*[`internal/manifest/version_edit.go` line 582 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ0tjJpBMk6ACBLz9k2:-NQ0tjJpBMk6ACBLz9k3:b-84cisn) ([raw file](https://github.com/cockroachdb/pebble/blob/d52333ce1a349d2bc308c2c1e32a726adbef415e/internal/manifest/version_edit.go#L582)):*
> ```Go
> // VersionEdit.DeletedFiles, but not present in VersionEdit.AddedFiles. Note
> // that move compactions must move the files from one level to another
> // atomically through a single version edit.
> ```

Do we need to track the zombies? We don't have the complete list here since we don't know which Version this BVE will be applied to. It seems we have enough information to do this calculation in Apply where we have the `Version`.

Copying my comment from the other PR:

-----

We could switch to
```
Added [NumLevels]map[base.FileNum]*FileMetadata
Deleted [NumLevels][]*FileMetadata
```
and in BulkVersionEdit.Accumulate remove the files in Added when they are deleted. And only add to Deleted if there was no corresponding file in Added. So all the de-duping is already done, and the remaining thing in Apply is to add files that were not in the Version or to remove files that are in the Version. This also allows us to iterate over the added files first and don't have to reason in a complicated manner about the refcount.

----

We simply don't add to `Deleted[j]` when the `VersionEdit` has both added and deleted from level j. So since there is no-deduping or removal needed from `Deleted[j]`, it is fine for it to be a slice. 

A file in `Deleted[j]` must have been added in a `VersionEdit` that is not part of this BVE. The file may then have been move compacted down to level k, and we will cancel out add and delete pairs in every level, so this file is either only present in one `Deleted[j]` or is also present in one `Added[k]` where k > j. So BVE.Apply can do the zombie computation using a local `zombies map[base.FileNum]uint64`.

Am I missing something?

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`internal/manifest/version_edit.go` line 556 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ0uxHo1DQgGtwaMJtv:-NQ1EKUsD-cXrLjfJXrY:b-zcj781) ([raw file](https://github.com/cockroachdb/pebble/blob/d52333ce1a349d2bc308c2c1e32a726adbef415e/internal/manifest/version_edit.go#L556)):*

Does `Deleted` need to be a map. See my comment below.
</blockquote></details>

Not necessarily, but it's used for some invariant checking in Accumulate while we're adding files.

```
// A new file should not have been deleted in this or a preceding
// VersionEdit at the same level (though files can move across levels).
if dmap := b.Deleted[nf.Level]; dmap != nil {
    if _, ok := dmap[nf.Meta.FileNum]; ok {
        return base.CorruptionErrorf("pebble: file deleted L%d.%s before it was inserted", nf.Level, nf.Meta.FileNum)
    }
}
```

I believe this also checks for intra L0 move compactions inadvertently.
___
*[`internal/manifest/version_edit.go` line 582 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ0tjJpBMk6ACBLz9k2:-NQ1F1cRAvZ9u0Z4RJmf:bibrm5z) ([raw file](https://github.com/cockroachdb/pebble/blob/d52333ce1a349d2bc308c2c1e32a726adbef415e/internal/manifest/version_edit.go#L582)):*
> Do we need to track the zombies? We don't have the complete list here since we don't know which Version this BVE will be applied to.

I believe we do(explained below). Why isn't the list complete? bve application always happens to the latest version, or a new empty version. If a bve contains a deleted entry, but no added entry for a file, then the file will be deleted from the latest version, and it's a zombie. If there's no deleted entry for the file, then it won't be a zombie.

>  BulkVersionEdit.Accumulate remove the files in Added when they are deleted. And only add to Deleted if there was no corresponding file in Added.

This is what we're doing now.

> We simply don't add to Deleted[j] when the VersionEdit has both added and deleted from level j. So since there is no-deduping or removal needed from Deleted[j], it is fine for it to be a slice.

It's used for invariant checking as mentioned in the comment above.

> A file in Deleted[j] must have been added in a VersionEdit that is not part of this BVE. The file may then have been move compacted down to level k, and we will cancel out add and delete pairs in every level, so this file is either only present in one Deleted[j] or is also present in one Added[k] where k > j. So BVE.Apply can do the zombie computation using a local zombies map[base.FileNum]uint64.

Let's say a file has an added entry for level 0 in one ve, and then a deleted entry for level 0 in another ve. Let's say pebble is closed before this file is deleted from fisk. Let's say we're replaying the manifest. In the accumulate step, we won't add this file to `bve.Deleted[0]`. But this file is a zombie and it must be present in the zombie list until it is deleted from disk. The apply step won't have any information on this file.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`internal/manifest/version_edit.go` line 582 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ0tjJpBMk6ACBLz9k2:-NQ26tC7C5yfWRBnVM7v:bbuh957) ([raw file](https://github.com/cockroachdb/pebble/blob/d52333ce1a349d2bc308c2c1e32a726adbef415e/internal/manifest/version_edit.go#L582)):*

> Do we need to track the zombies? We don't have the complete list here since we don't know which Version this BVE will be applied to.

I believe we do(explained below). Why isn't the list complete? bve application always happens to the latest version, or a new empty version. If a bve contains a deleted entry, but no added entry for a file, then the file will be deleted from the latest version, and it's a zombie. If there's no deleted entry for the file, then it won't be a zombie.

>  BulkVersionEdit.Accumulate remove the files in Added when they are deleted. And only add to Deleted if there was no corresponding file in Added.

This is what we're doing now.

> We simply don't add to Deleted[j] when the VersionEdit has both added and deleted from level j. So since there is no-deduping or removal needed from Deleted[j], it is fine for it to be a slice.

It's used for invariant checking as mentioned in the comment above.

> A file in Deleted[j] must have been added in a VersionEdit that is not part of this BVE. The file may then have been move compacted down to level k, and we will cancel out add and delete pairs in every level, so this file is either only present in one Deleted[j] or is also present in one Added[k] where k > j. So BVE.Apply can do the zombie computation using a local zombies map[base.FileNum]uint64.

Let's say a file has an added entry for level 0 in one ve, and then a deleted entry for level 0 in another ve. Let's say pebble is closed before this file is deleted from fisk. Let's say we're replaying the manifest. In the accumulate step, we won't add this file to `bve.Deleted[0]`. But this file is a zombie and it must be present in the zombie list until it is deleted from disk. The apply step won't have any information on this file.
</blockquote></details>

I thought about what should go into the zombie tables map a bit more:

1. https://github.com/cockroachdb/pebble/blob/master/version_set.go#L750. A file must be present in the zombie tables map before it is added to the obsolete tables list through `addObsoleteLocked`.

2. Files are also added to the obsolete tables list without checking if the file is also present in the zombie tables list in `scanObsoleteFiles` during open.

3. https://github.com/cockroachdb/pebble/blob/master/compaction.go#L3372. A file is removed from the zombie tables map once the file is deleted from disk, but it doesn't necessarily have to be present in the zombie tables list because it is safe to call `delete` on a key on a map even if the key isn't present in the map.

In general, the zombie tables map is supposed to represent files which are no longer required by the latest version. But even if a file is required by no version, and it is obsolete, we only remove the file from the zombie table map once the file is deleted from disk.

So the question is whether a file `f` which was both added and deleted in the same bulk version edit to the same level `l`, must be present in the zombie tables map. I'm going to assume that this file isn't moved.

Currently, we will add such a file to the zombie list. Since it wasn't moved, we know that it isn't required by the latest version. So, technically it is a zombie. But then we need to make sure that this file will actually be removed from the zombie tables list. I think during regular Pebble operation, where we accumulate a bve and a file is newly made a zombie, then the file will be present on disk, and will eventually be deleted and removed from the zombie map.

But I think there is a case where a file can be present in the zombie map, and will never be deleted from the zombie map. Let's say the file `f`, described above, is added to the zombie map, and then deleted from disk, and then removed from the zombie map. But the file `'f` can still exist in the manifest. Let's say Pebble is now restarted. During replay, we will add file `f` back to the zombie map. But file f will no longer be present on disk, so it will never be deleted, and hence, we will never remove file `f` from the zombie map.

So, we should add file `f` to the zombie map iff we know that it hasn't been deleted yet. We could also remove files which don't exist on disk from the zombie map during the `scanObsoleteFiles` step.

What do you think? I tried to do a proof, but I ended up rambling. Lmk if I need to clarify anything.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`internal/manifest/version_edit.go` line 582 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ0tjJpBMk6ACBLz9k2:-NQ2DAlF6GjN1lKRa3vx:b2jnamz) ([raw file](https://github.com/cockroachdb/pebble/blob/d52333ce1a349d2bc308c2c1e32a726adbef415e/internal/manifest/version_edit.go#L582)):*

I thought about what should go into the zombie tables map a bit more:

1. https://github.com/cockroachdb/pebble/blob/master/version_set.go#L750. A file must be present in the zombie tables map before it is added to the obsolete tables list through `addObsoleteLocked`.

2. Files are also added to the obsolete tables list without checking if the file is also present in the zombie tables list in `scanObsoleteFiles` during open.

3. https://github.com/cockroachdb/pebble/blob/master/compaction.go#L3372. A file is removed from the zombie tables map once the file is deleted from disk, but it doesn't necessarily have to be present in the zombie tables list because it is safe to call `delete` on a key on a map even if the key isn't present in the map.

In general, the zombie tables map is supposed to represent files which are no longer required by the latest version. But even if a file is required by no version, and it is obsolete, we only remove the file from the zombie table map once the file is deleted from disk.

So the question is whether a file `f` which was both added and deleted in the same bulk version edit to the same level `l`, must be present in the zombie tables map. I'm going to assume that this file isn't moved.

Currently, we will add such a file to the zombie list. Since it wasn't moved, we know that it isn't required by the latest version. So, technically it is a zombie. But then we need to make sure that this file will actually be removed from the zombie tables list. I think during regular Pebble operation, where we accumulate a bve and a file is newly made a zombie, then the file will be present on disk, and will eventually be deleted and removed from the zombie map.

But I think there is a case where a file can be present in the zombie map, and will never be deleted from the zombie map. Let's say the file `f`, described above, is added to the zombie map, and then deleted from disk, and then removed from the zombie map. But the file `'f` can still exist in the manifest. Let's say Pebble is now restarted. During replay, we will add file `f` back to the zombie map. But file f will no longer be present on disk, so it will never be deleted, and hence, we will never remove file `f` from the zombie map.

So, we should add file `f` to the zombie map iff we know that it hasn't been deleted yet. We could also remove files which don't exist on disk from the zombie map during the `scanObsoleteFiles` step.

What do you think? I tried to do a proof, but I ended up rambling. Lmk if I need to clarify anything.
</blockquote></details>

Looking at the problem from a higher level. During regular pebble operations, where we accumulate a single `ve` in `logAndApply`, we will never have a file being added and removed from the same level in the same bulk version edit.

During manifest replay, we could have files which are added and removed from the same level in the same bulk version edit. But such files are obsolete, and we shouldn't count them as zombies because they'll be deleted in the `scanObsoleteFiles` step during `Open` anyway.

If we go with the above, then we can move zombie table computation to `Apply`.

RaduBerinde (commented):
We're missing a version edit test that checks what the BVE looks like after add/remove from the same level.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384#-:-NQ2JqhR0lq6lxxV1ngZ:bj7st4)* status: 0 of 2 files reviewed, 8 unresolved discussions (waiting on @bananabrick and @sumeerbhola)

___
*[`internal/manifest/version_edit.go` line 551 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ2F7wk981r_3ciqOWp:-NQ2F7wk981r_3ciqOWq:bz31kzb) ([raw file](https://github.com/cockroachdb/pebble/blob/22ff4ab9b730455a721484f3723122bcda549241/internal/manifest/version_edit.go#L551)):*
> ```Go
> // edit.
> //
> // A file must be added to a level before(in a prior version edit) it is removed
> ```

[nit] I'd rephrase as "A file that is being removed from a level must have been added to that level before (in a prior version edit)."

And "Note that a given file can be deleted from a level and added to another level in a single version edit" (is this what you meant?)
___
*[`internal/manifest/version_edit.go` line 557 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ2FbSy93VG8HH-AMiu:-NQ2FbSy93VG8HH-AMiv:b-eack2o) ([raw file](https://github.com/cockroachdb/pebble/blob/22ff4ab9b730455a721484f3723122bcda549241/internal/manifest/version_edit.go#L557)):*
> ```Go
> 	Added   [NumLevels]map[base.FileNum]*FileMetadata
> 	Deleted [NumLevels]map[base.FileNum]*FileMetadata
> 	Zombies map[base.FileNum]uint64
> ```

[nit] needs a comment explaining what a zombie is exactly (and what the `uint64` is)
___
*[`internal/manifest/version_edit.go` line 585 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ2Fs4O5L6VT_TTVHRT:-NQ2Fs4O5L6VT_TTVHRU:b1rim17) ([raw file](https://github.com/cockroachdb/pebble/blob/22ff4ab9b730455a721484f3723122bcda549241/internal/manifest/version_edit.go#L585)):*
> ```Go
> //
> // If a file is added and deleted from a given level across multiple calls to
> // accumulate, the file will not be present be present in
> ```

[nit] `If a file is added to a given level in a call to Accumulate and then removed from that level in a subsequent call, the file will not be present in the resulting BulkVersionEdit`
___
*[`internal/manifest/version_edit.go` line 631 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ2ITsm5rZZFp-apREK:-NQ2ITsm5rZZFp-apREL:b-sze5z3) ([raw file](https://github.com/cockroachdb/pebble/blob/22ff4ab9b730455a721484f3723122bcda549241/internal/manifest/version_edit.go#L631)):*
> ```Go
> 			b.MarkedForCompactionCountDiff--
> 		}
> 		if b.Added[df.Level] == nil || b.Added[df.Level][df.FileNum] == nil {
> ```

[nit] no need for the first `nil` check (it's ok to read from a nil map). And the usual pattern is `if _, exists := b.Added[df.Level][df.FileNum]; exists`
___
*[`internal/manifest/version_edit.go` line 648 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ2JJJy7VeEWlfLdH8P:-NQ2JJJy7VeEWlfLdH8Q:b9bua7j) ([raw file](https://github.com/cockroachdb/pebble/blob/22ff4ab9b730455a721484f3723122bcda549241/internal/manifest/version_edit.go#L648)):*
> ```Go
> 			}
> 		}
> 		amap := b.Added[nf.Level]
> ```

This variable isn't very useful, can just do `if b.Added[nf.Level] == nil { b.Added[nf.Level] = make() }`
___
*[`internal/manifest/version_edit.go` line 796 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ2GXVs9z86PyYdgLsf:-NQ2GXVs9z86PyYdgLsg:bv6zzjx) ([raw file](https://github.com/cockroachdb/pebble/blob/22ff4ab9b730455a721484f3723122bcda549241/internal/manifest/version_edit.go#L796)):*
> ```Go
> 				var err error
> 				// AddL0Files requires addedFiles to be sorted in seqnum order.
> 				var l0Files []*FileMetadata
> ```

[nit] why not `l0Files := append([]*FileMetadata(nil), addedFiles...)`? This would allocate it just once. Or do `l0Files := make([]*FileMetadata, 0, len(addedFiles))` if you prefer it the other way

sumeerbhola (changes requested):
+1 to testing the new logic

rest looks good

> Reviewed 1 of 2 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384#-:-NQ2K9We-wDJXHHb7miK:b-rf7oze)* status: all files reviewed, 9 unresolved discussions (waiting on @bananabrick)

___
*[`internal/manifest/version_edit.go` line 556 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ0uxHo1DQgGtwaMJtv:-NQ29CDv0Vw1xpB3sepI:bq0ncok) ([raw file](https://github.com/cockroachdb/pebble/blob/d52333ce1a349d2bc308c2c1e32a726adbef415e/internal/manifest/version_edit.go#L556)):*

Not necessarily, but it's used for some invariant checking in Accumulate while we're adding files.

```
// A new file should not have been deleted in this or a preceding
// VersionEdit at the same level (though files can move across levels).
if dmap := b.Deleted[nf.Level]; dmap != nil {
    if _, ok := dmap[nf.Meta.FileNum]; ok {
        return base.CorruptionErrorf("pebble: file deleted L%d.%s before it was inserted", nf.Level, nf.Meta.FileNum)
    }
}
```

I believe this also checks for intra L0 move compactions inadvertently.
</blockquote></details>

sounds good
___
*[`internal/manifest/version_edit.go` line 557 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ29RlaESYqgCsnrK3c:-NQ29RlaESYqgCsnrK3d:b-vtycnz) ([raw file](https://github.com/cockroachdb/pebble/blob/22ff4ab9b730455a721484f3723122bcda549241/internal/manifest/version_edit.go#L557)):*
> ```Go
> 	Added   [NumLevels]map[base.FileNum]*FileMetadata
> 	Deleted [NumLevels]map[base.FileNum]*FileMetadata
> 	Zombies map[base.FileNum]uint64
> ```

// Zombies tracks files that are deleted across the version edits accumulated in the BulkVersionEdit.
// It includes files that were both added and deleted across these version edits -- this is necessary since BulkVersionEdit is also used at Pebble reopen, where knowing these zombie files allows us to delete them from disk.

bananabrick (commented):
We can disregard what I said regarding zombie tables because it turns out that we don't update the zombie table map during manifest replay. That makes sense. Any zombie file will be deleted during `Open` anyway, so there's no point to updating the map. It's a bit confusing that we always compute these zombie tables when `BulkVersionEdit.Apply` is called, but only use it from `versionSet.logAndApply`.

> We're missing a version edit test that checks what the BVE looks like after add/remove from the same level.

Done. The test didn't have support for applying multiple version edits to a single bve. I added that. Test is getting quite messy and could use a refactor in the future.

I also separated out `bve.Apply` into `bve.Apply` and `AccumulateAndApplySingleVE`. I think it's easier to reason about the properties of bulk version edit accumulation/application and zombie table generation, if we have stricter invariants about when these functions can be called. This separation will also help with the virtual sstables code.

Since zombie tables are only required by the `AccumulateAndApplySingleVE` step, we now only set the zombie table map in that function. We can also more precisely reason about why the file is a zombie.

PTAL. The code might have changed quite a bit since the previous review.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384#-:-NPzKiTm7aI4McjUOBPX:b-yymq2b)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/manifest/version_edit.go` line 556 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ0uxHo1DQgGtwaMJtv:-NQ6kHDu9Hzgk-JSoqgW:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/d52333ce1a349d2bc308c2c1e32a726adbef415e/internal/manifest/version_edit.go#L556)):*

sounds good
</blockquote></details>

Done.
___
*[`internal/manifest/version_edit.go` line 551 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ2F7wk981r_3ciqOWp:-NQ6kcarDXMIL-6aFoUX:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/22ff4ab9b730455a721484f3723122bcda549241/internal/manifest/version_edit.go#L551)):*

[nit] I'd rephrase as "A file that is being removed from a level must have been added to that level before (in a prior version edit)."

And "Note that a given file can be deleted from a level and added to another level in a single version edit" (is this what you meant?)
</blockquote></details>

Done.
___
*[`internal/manifest/version_edit.go` line 557 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ29RlaESYqgCsnrK3c:-NQ6rz-w6_2vR5iFZTud:bq709al) ([raw file](https://github.com/cockroachdb/pebble/blob/22ff4ab9b730455a721484f3723122bcda549241/internal/manifest/version_edit.go#L557)):*

// Zombies tracks files that are deleted across the version edits accumulated in the BulkVersionEdit.
// It includes files that were both added and deleted across these version edits -- this is necessary since BulkVersionEdit is also used at Pebble reopen, where knowing these zombie files allows us to delete them from disk.
</blockquote></details>

Done. Changed wording a bit.
___
*[`internal/manifest/version_edit.go` line 557 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ2FbSy93VG8HH-AMiu:-NQ6rxhe8ol-ZsflmidP:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/22ff4ab9b730455a721484f3723122bcda549241/internal/manifest/version_edit.go#L557)):*

[nit] needs a comment explaining what a zombie is exactly (and what the `uint64` is)
</blockquote></details>

Done.

RaduBerinde (commented):
> It's a bit confusing that we always compute these zombie tables when BulkVersionEdit.Apply is called, but only use it from versionSet.logAndApply.

Would it make sense to take out the zombie stuff from Apply into a separate method?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384#-:-NQC-51C28LZaLJ661qk:b-3n3h0h)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @sumeerbhola)

___
*[`internal/manifest/version_edit.go` line 557 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ29RlaESYqgCsnrK3c:-NQC-mno9nzDPPEGmWCX:b-dh86fl) ([raw file](https://github.com/cockroachdb/pebble/blob/22ff4ab9b730455a721484f3723122bcda549241/internal/manifest/version_edit.go#L557)):*

Done. Changed wording a bit.
</blockquote></details>

[nit] Maybe mention that any deletion in `Deleted` either has a corresponding addition in `Added` or appears here (is that correct?)

[nit] Maybe `DeletedPermanently`?

I wonder if it would be cleaner to have `Added`, `Moved`, and `Deleted` instead.

sumeerbhola (changes requested):
> Reviewed 4 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384)* status: 4 of 9 files reviewed, 5 unresolved discussions (waiting on @bananabrick and @RaduBerinde)

___
*[`internal/manifest/version_edit.go` line 582 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ0tjJpBMk6ACBLz9k2:-NQQCp9-9iKHkSNS0F88:b-4cndvz) ([raw file](https://github.com/cockroachdb/pebble/blob/d52333ce1a349d2bc308c2c1e32a726adbef415e/internal/manifest/version_edit.go#L582)):*
> Why isn't the list complete? bve application always happens to the latest version, or a new empty version. If a bve contains a deleted entry, but no added entry for a file, then the file will be deleted from the latest version, and it's a zombie.

I was looking ahead to the virtual sst PR where knowing whether something is a zombie will require updating the ref counts on the latest Version, which we don't have at Accumulate time.

I like the separation of `AccumulateAndApplySingleVE`.
___
*[`internal/manifest/version_edit.go` line 558 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQQBrvU8RwLe2Q4Iwsq:-NQQBrvU8RwLe2Q4Iwsr:b-g2bm8t) ([raw file](https://github.com/cockroachdb/pebble/blob/8818b0b05443058e86f18744a700b0b41ca087ff/internal/manifest/version_edit.go#L558)):*
> ```Go
> 	Deleted [NumLevels]map[base.FileNum]*FileMetadata
> 	// DeletedAndNotAdded is a map of file numbers which have been deleted by some
> 	// VersionEdit ve but not Added in ve or any subsequent version edit.
> ```

"not Added in ... any subsequent version edit" does not sound right. We are not in the business of foretelling the future, and so if it does not get added in the same version edit it is gone.
(This is also why if we import the same foreign sst in disagg storage multiple times, we will assign it a new local filenum)
___
*[`internal/manifest/version_edit.go` line 559 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQQEhAWDBQXrryMcjpC:-NQQEhAX9ammmbGThcMv:b46f75f) ([raw file](https://github.com/cockroachdb/pebble/blob/8818b0b05443058e86f18744a700b0b41ca087ff/internal/manifest/version_edit.go#L559)):*
> ```Go
> 	// DeletedAndNotAdded is a map of file numbers which have been deleted by some
> 	// VersionEdit ve but not Added in ve or any subsequent version edit.
> 	DeletedAndNotAdded map[base.FileNum]uint64 // file number -> file size
> ```

I'm still confused about why we need this and how this will need to evolve with virtual ssts.

We've de-duped the changes in Added/Deleted. The resulting state represents *all* the ref counting updates we need to make.

We had previously added `Zombies` not for the ref counting but to preserve state for the future zombie computation that Apply was expected to do, that was no longer captured in BVE because of the de-duping. Now that use case has gone away, and all the ref-counting deltas needed are captured in Added/Deleted. Sure, we can compute something liked `DeletedAndNotAdded` now, like is done here, but I don't see a clear path to extending this in virtual ssts where the notion of zombie depends on the `Version` (since there are multiple refs from the latest version), that we don't have in `Accumulate`.

That is, I don't see how the code below can cleanly evolve.
```
	err := b.Accumulate(ve)
	zombies := b.DeletedAndNotAdded
```
I still think we should move the zombie computation to `BVE.Apply`. We can make it optional since it isn't needed for the DB initialization path.

Am I missing something?

bananabrick (commented):
> Would it make sense to take out the zombie stuff from Apply into a separate method?

For now, I've made required callers of Apply which want zombie tables calculated to explicitly pass in a non-nil map which will be populated with zombie tables.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384#-:-NQRQiYy37JcsMVVp-a6:b1ewozk)* status: 0 of 9 files reviewed, 5 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/manifest/version_edit.go` line 582 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ0tjJpBMk6ACBLz9k2:-NQR60qpDLSLMABg_DpM:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/d52333ce1a349d2bc308c2c1e32a726adbef415e/internal/manifest/version_edit.go#L582)):*

> Why isn't the list complete? bve application always happens to the latest version, or a new empty version. If a bve contains a deleted entry, but no added entry for a file, then the file will be deleted from the latest version, and it's a zombie.

I was looking ahead to the virtual sst PR where knowing whether something is a zombie will require updating the ref counts on the latest Version, which we don't have at Accumulate time.

I like the separation of `AccumulateAndApplySingleVE`.
</blockquote></details>

Done.
___
*[`internal/manifest/version_edit.go` line 557 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQ29RlaESYqgCsnrK3c:-NQR5xTsCtv0ijLHiKp4:bss4a2b) ([raw file](https://github.com/cockroachdb/pebble/blob/22ff4ab9b730455a721484f3723122bcda549241/internal/manifest/version_edit.go#L557)):*

[nit] Maybe mention that any deletion in `Deleted` either has a corresponding addition in `Added` or appears here (is that correct?)

[nit] Maybe `DeletedPermanently`?

I wonder if it would be cleaner to have `Added`, `Moved`, and `Deleted` instead.
</blockquote></details>

Removed.
___
*[`internal/manifest/version_edit.go` line 558 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQQBrvU8RwLe2Q4Iwsq:-NQR6-Is8oUyf4N9ykum:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8818b0b05443058e86f18744a700b0b41ca087ff/internal/manifest/version_edit.go#L558)):*

"not Added in ... any subsequent version edit" does not sound right. We are not in the business of foretelling the future, and so if it does not get added in the same version edit it is gone.
(This is also why if we import the same foreign sst in disagg storage multiple times, we will assign it a new local filenum)
</blockquote></details>

Done.
___
*[`internal/manifest/version_edit.go` line 559 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQQEhAWDBQXrryMcjpC:-NQR6A692ITXg5X6S47l:b-ul6v3w) ([raw file](https://github.com/cockroachdb/pebble/blob/8818b0b05443058e86f18744a700b0b41ca087ff/internal/manifest/version_edit.go#L559)):*

I'm still confused about why we need this and how this will need to evolve with virtual ssts.

We've de-duped the changes in Added/Deleted. The resulting state represents *all* the ref counting updates we need to make.

We had previously added `Zombies` not for the ref counting but to preserve state for the future zombie computation that Apply was expected to do, that was no longer captured in BVE because of the de-duping. Now that use case has gone away, and all the ref-counting deltas needed are captured in Added/Deleted. Sure, we can compute something liked `DeletedAndNotAdded` now, like is done here, but I don't see a clear path to extending this in virtual ssts where the notion of zombie depends on the `Version` (since there are multiple refs from the latest version), that we don't have in `Accumulate`.

That is, I don't see how the code below can cleanly evolve.
```
	err := b.Accumulate(ve)
	zombies := b.DeletedAndNotAdded
```
I still think we should move the zombie computation to `BVE.Apply`. We can make it optional since it isn't needed for the DB initialization path.

Am I missing something?

</blockquote></details>

If we don't perform zombie table computation in Apply, then we won't add files which were added/removed from the same level in the same bulk version edit to the zombie map. This is okay, because zombie tables are only used from `AccumulateAndApplySingleVE` and in a single version edit we'll never add/remove a file from the same level. If we start accumulating more than one version edit during regular Pebble operation, then our zombie table calculation will be incorrect.

I guess I shouldn't anticipate a problem which doesn't exist yet so I've moved the zombie table calculation back to `Apply`.

The only change is that the caller of `Apply` must now pass in a non-nil zombies map if they want zombie tables to be generated. A non-nil zombies map is only passed in to `Apply` by `AccumulateAndApplySingleVE`.

I've added a comment warning that zombie table computation only works if a single version edit was accumulated.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384#-:-NQRcmXPC2ZeosT9Bq70:bnfp4nl)* status: 0 of 9 files reviewed, 11 unresolved discussions (waiting on @bananabrick and @sumeerbhola)

___
*[`version_set.go` line 280 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRaPLpAp2CCFJ8xdbd:-NQRaPLpAp2CCFJ8xdbe:b-xf0d4o) ([raw file](https://github.com/cockroachdb/pebble/blob/fcfce8f03801e336dac6a1b429047b90be629158/version_set.go#L280)):*
> ```Go
> 	vs.markFileNumUsed(vs.minUnflushedLogNum)
> 
> 	newVersion, err := bve.Apply(nil, vs.cmp, opts.Comparer.FormatKey, opts.FlushSplitBytes, opts.Experimental.ReadCompactionRate, nil)
> ```

[nit] nil /* zombies */
___
*[`internal/manifest/version_edit.go` line 647 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRc1fwAuF0K3-yDZ6k:-NQRc1fwAuF0K3-yDZ6l:bdmcd25) ([raw file](https://github.com/cockroachdb/pebble/blob/fcfce8f03801e336dac6a1b429047b90be629158/internal/manifest/version_edit.go#L647)):*
> ```Go
> // regarding BulkVersionedit.Accumulate/Apply and zombie table generation, if we
> // know that exactly one version edit is being accumulated. It also helps us
> // reason about the correctness of new zombie table generation.
> ```

[nit] zombie table generation is mentioned twice
___
*[`internal/manifest/version_edit.go` line 651 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRcLXt8yoUepqcVlwa:-NQRcLXt8yoUepqcVlwb:b-l071qy) ([raw file](https://github.com/cockroachdb/pebble/blob/fcfce8f03801e336dac6a1b429047b90be629158/internal/manifest/version_edit.go#L651)):*
> ```Go
> // Invariant: Any file being deleted through ve must belong to the curr Version.
> // We can't have a delete for some arbitrary file which does not exist in curr.
> func AccumulateAndApplySingleVE(
> ```

[nit] Accumulate doesn't make sense for a single VE, maybe just `ApplySingleVersionEdit`?
___
*[`internal/manifest/version_edit.go` line 658 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRc_H11Z3T5zWS14Wy:-NQRc_H11Z3T5zWS14Wz:bykx8zj) ([raw file](https://github.com/cockroachdb/pebble/blob/fcfce8f03801e336dac6a1b429047b90be629158/internal/manifest/version_edit.go#L658)):*
> ```Go
> 	flushSplitBytes int64,
> 	readCompactionRate int64,
> ) (*Version, map[base.FileNum]uint64, error) {
> ```

[nit] add a `zombies` name for the map. Would be useful if the doc for the function would mention them, or at least point to bve.Apply for info on zombies
___
*[`internal/manifest/version_edit.go` line 664 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRad3Z4hMzb1jE9SgI:-NQRad3Z4hMzb1jE9SgJ:b4de2ki) ([raw file](https://github.com/cockroachdb/pebble/blob/fcfce8f03801e336dac6a1b429047b90be629158/internal/manifest/version_edit.go#L664)):*
> ```Go
> 		return nil, nil, err
> 	}
> 	// If a file was deleted in a version edit, and not added in that version edit
> ```

[nit] this comment seems off (especially the "subsequent version edit" part)
___
*[`tool/db.go` line 486 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRbBIx-kWfCpvN5jFd:-NQRbBIx-kWfCpvN5jFe:b87qvyv) ([raw file](https://github.com/cockroachdb/pebble/blob/fcfce8f03801e336dac6a1b429047b90be629158/tool/db.go#L486)):*
> ```Go
> 			}
> 		}
> 		v, err := bve.Apply(nil /* version */, cmp.Compare, d.fmtKey.fn, d.opts.FlushSplitBytes, d.opts.Experimental.ReadCompactionRate, nil)
> ```

[nit] /* zombies */

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384)* status: 0 of 9 files reviewed, 6 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/manifest/version_edit.go` line 651 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRcLXt8yoUepqcVlwa:-NQRiMvzEOj3CL-Zw8KB:b-6rsfds) ([raw file](https://github.com/cockroachdb/pebble/blob/fcfce8f03801e336dac6a1b429047b90be629158/internal/manifest/version_edit.go#L651)):*

[nit] Accumulate doesn't make sense for a single VE, maybe just `ApplySingleVersionEdit`?
</blockquote></details>

We still need to call `BulkVersionEdit.Accumulate`. Left this as it is.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384)* status: 0 of 9 files reviewed, 6 unresolved discussions (waiting on @bananabrick and @sumeerbhola)

___
*[`internal/manifest/version_edit.go` line 651 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRcLXt8yoUepqcVlwa:-NQRvo-85ujcYIXX_AD5:b19yyab) ([raw file](https://github.com/cockroachdb/pebble/blob/fcfce8f03801e336dac6a1b429047b90be629158/internal/manifest/version_edit.go#L651)):*

We still need to call `BulkVersionEdit.Accumulate`. Left this as it is.
</blockquote></details>

The fact that we're using `BulkVersionEdit` internally to reuse some code is irrelevant to the caller. In fact, maybe this should be an `Apply` method on `VersionEdit`?

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 9 files at r4, 5 of 8 files at r5, 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384#-:-NQS1FVQ5sFKXP4lxvwV:bg7igkv)* status: all files reviewed, 10 unresolved discussions (waiting on @bananabrick)

___
*[`internal/manifest/version_edit.go` line 683 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRwzCcDy6IncI1MK2q:-NQRwzCcDy6IncI1MK2r:b-2zmf8g) ([raw file](https://github.com/cockroachdb/pebble/blob/17da3f9759066852ae58d9854ffbb4bd52d4b903/internal/manifest/version_edit.go#L683)):*
> ```Go
> //
> // Note that zombie table computation is only accurate if exactly one version
> // edit was accumulated in the BulkVersionEdit.Accumulate step.
> ```

Or one could define zombies to be those files in `curr` that are no longer needed in the returned `Version`.
Which I believe is what the earlier paragraph is roughly saying. I think the computation is correct with that definition regardless of the number of version edits accumulated into BVE.
___
*[`internal/manifest/version_edit.go` line 747 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRxhI68kiZbZWXwHEO:-NQRxhI68kiZbZWXwHEP:b-mr5urn) ([raw file](https://github.com/cockroachdb/pebble/blob/17da3f9759066852ae58d9854ffbb4bd52d4b903/internal/manifest/version_edit.go#L747)):*
> ```Go
> 
> 		addedFilesMap := b.Added[level]
> 		deletedMap := b.Deleted[level]
> ```

deletedFilesMap?
to be consistent with addedFilesMap
___
*[`internal/manifest/version_edit.go` line 755 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRxr8l-GrkEyAepW0r:-NQRxr8l-GrkEyAepW0s:b-w9spfe) ([raw file](https://github.com/cockroachdb/pebble/blob/17da3f9759066852ae58d9854ffbb4bd52d4b903/internal/manifest/version_edit.go#L755)):*
> ```Go
> 
> 		// NB: addedFilesMap may be empty, and it is internally consistent with
> 		// deletedMap.
> ```

Please define "internally consistent".
___
*[`internal/manifest/version_edit.go` line 758 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRzV1fCkFQ9IplxuUE:-NQRzV1fCkFQ9IplxuUF:b-p5549a) ([raw file](https://github.com/cockroachdb/pebble/blob/17da3f9759066852ae58d9854ffbb4bd52d4b903/internal/manifest/version_edit.go#L758)):*
> ```Go
> 
> 		for _, f := range deletedMap {
> 			addZombie(f.FileNum, f.Size)
> ```

The reason we landed in this PR is the discussion in the other PR about ref-counting.

Since we will need to see `latestRefs` drop to 0, before marking something as zombie, one natural way to do that is to proceed level-by-level and process all the addedFilesMap. Then proceed level-by-level and process all the deletedFilesMap. Anything that has `latestRefs` drop to 0 in the latter step is a zombie (one doesn't need `removeZombie`).
I'm also ok with this level-by-level processing of the added/deleted pair, and keeping the `removeZombie`.

bananabrick (commented):
tftr.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384#-:-NQSNFyN51pksEVJnvHv:b-7ixx5z)* status: 8 of 9 files reviewed, 10 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/manifest/version_edit.go` line 651 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRcLXt8yoUepqcVlwa:-NQSL-_xCyGPHu8EnKkT:b-3ug7y6) ([raw file](https://github.com/cockroachdb/pebble/blob/fcfce8f03801e336dac6a1b429047b90be629158/internal/manifest/version_edit.go#L651)):*
I think calling it `AccumulateAndApply` is okay. Not convinced that this is an internal detail of the abstraction which needs to be hidden from the caller.

> In fact, maybe this should be an Apply method on VersionEdit?

I don't think I want a method on the VersionEdit to internally use a BulkVersionEdit.
___
*[`internal/manifest/version_edit.go` line 683 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRwzCcDy6IncI1MK2q:-NQSMOjYC2TFT6qO5YCp:b-ash261) ([raw file](https://github.com/cockroachdb/pebble/blob/17da3f9759066852ae58d9854ffbb4bd52d4b903/internal/manifest/version_edit.go#L683)):*
>// Note that zombie table computation is only accurate if exactly one version
// edit was accumulated in the BulkVersionEdit.Accumulate step.

Got rid of this.
___
*[`internal/manifest/version_edit.go` line 747 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRxhI68kiZbZWXwHEO:-NQSLiJx0Dn0ncx3fAXC:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/17da3f9759066852ae58d9854ffbb4bd52d4b903/internal/manifest/version_edit.go#L747)):*

deletedFilesMap?
to be consistent with addedFilesMap
</blockquote></details>

Done.
___
*[`internal/manifest/version_edit.go` line 755 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRxr8l-GrkEyAepW0r:-NQSMNon4XJ9Zv-afG8-:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/17da3f9759066852ae58d9854ffbb4bd52d4b903/internal/manifest/version_edit.go#L755)):*

Please define "internally consistent".
</blockquote></details>

Done.
___
*[`internal/manifest/version_edit.go` line 758 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2384#-NQRzV1fCkFQ9IplxuUE:-NQSMEaW92Uja4EhtGhz:b9fc437) ([raw file](https://github.com/cockroachdb/pebble/blob/17da3f9759066852ae58d9854ffbb4bd52d4b903/internal/manifest/version_edit.go#L758)):*

The reason we landed in this PR is the discussion in the other PR about ref-counting.

Since we will need to see `latestRefs` drop to 0, before marking something as zombie, one natural way to do that is to proceed level-by-level and process all the addedFilesMap. Then proceed level-by-level and process all the deletedFilesMap. Anything that has `latestRefs` drop to 0 in the latter step is a zombie (one doesn't need `removeZombie`).
I'm also ok with this level-by-level processing of the added/deleted pair, and keeping the `removeZombie`.

</blockquote></details>

Keeping `removeZombie`, at least in this pr.

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2384#-:-NQVSKlbChkzBuiWarLb:bg7igkv)* status: all files reviewed, 10 unresolved discussions (waiting on @RaduBerinde)

Issue #2388, opened 2023-03-08
db: empty table error when ingesting as flushable


nicktrav:
When a table is ingested as a flushable, new flushables are added to the flushable queue for each table in the ingestion that overlaps. The memtable is then rotated, adding a new memtable to the end of the queue.

Consider a single non-empty memtable, followed by two tables ingested as flushable. After the ingested tables are added, the memtable queue resembles:

```
[non-empty memtable, ingested table A, ingested table B, empty-memtable]
<-head
```

The ingestion triggers a flush, which first flushes the non-empty memtable to L0, followed by ingestions of tables A and B into their appropriate levels. The empty memtable is then flushed, which emits the following error to the event logger:

```
2023/03/08 14:44:53 [JOB 21] flush error: pebble: empty table
```

While this error is harmless, it points to a performance issue where we unnecessarily flush the empty memtable.

See #2387 for a minimal reproducer.

Jira issue: PEBBLE-219

bananabrick:
As discussed on slack, we should only rotate the mutable memtable if it is non-empty: https://github.com/cockroachdb/pebble/blob/master/ingest.go#L849. If it is empty, then the ingestion cannot overlap with it, and we should be able to place the ingested flushable beneath the current mutable memtable.

I think too many memtables even if empty, can cause write stalls.

itsbilal:
@bananabrick I agree, we should only rotate the mutable memtable if it's non-empty. I imagine we'll need to avoid calling `makeRoomForWrite` here:

https://github.com/cockroachdb/pebble/blob/2be11668575156bb3ba15c25592f0f4614a420e9/ingest.go#L855

bananabrick:
I think this is a little more complex than I hoped. If we don't rotate the mutable memtable, then the ingested flushable, which will have a log number greater than the mutable memtable, will be placed before the mutable memtable in the queue. This breaks the flush invariant that memtables which are flushed should have increasing log numbers.

jbowens:
Can we handle this here: https://github.com/cockroachdb/pebble/blob/c24246ffc0123cb66aac8f7199400be4da351fd4/ingest.go#L759-L782

and if `currMem.empty()`, do not `recycleWAL()` and instead perform surgery over the `d.mu.mem.queue` to place the ingested flushable followed by the current memtable? That way the ingested flushable adopts the current mutable memtable's log number and the current mutable memtable stays the mutable memtable.

jbowens:
[...]

PR #2389, opened 2023-03-09
db: add metrics for flushable ingests; improve logging

nicktrav:
The new flushable ingest is designed to allow tables to be lazily ingested in the case where there is overlap with the memtable. To provide visibility into how the feature is being used, add two new flush metrics - a counter for each flushable ingest, and a counter for the number of bytes ingested as flushables.

Add the two new counts to the `Metrics` output. The formatting was altered slightly to improve the alignment.

Enhance the event logging for flushable ingests by printing the level into which the table is ingested.

### DIFF
--- a/testdata/event_listener
+++ b/testdata/event_listener
@@ -232,8 +232,8 @@ __level_____count____size___score______in__ingest(sz_cnt)____move(sz_cnt)___writ
       5         0     0 B    0.00     0 B     0 B       0     0 B       0     0 B       0     0 B       0     0.0
       6         1   770 B       -   1.5 K     0 B       0     0 B       0   770 B       1   1.5 K       1     0.5
   total         3   2.3 K       -   934 B   826 B       1     0 B       0   3.9 K       4   1.5 K       3     4.3
-  flush         3
-compact         1   2.3 K     0 B       0          (size == estimated-debt, score = in-progress-bytes, in = num-in-progress)
+  flush         3                             0 B       0          (ingest = ingested-as-flushable)
+compact         1   2.3 K     0 B       0                          (size == estimated-debt, score = in-progress-bytes, in = num-in-progress)
   ctype         1       0       0       0       0       0       0  (default, delete, elision, move, read, rewrite, multi-level)
  memtbl         1   256 K
 zmemtbl         0     0 B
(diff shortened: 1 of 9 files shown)


testdata/event_listener
```diff
@@ -244,11 +244,89 @@ zmemtbl         0     0 B
...
+close: db/marker.manifest.000007.MANIFEST-000022
+remove: db/marker.manifest.000006.MANIFEST-000016
+sync: db
+[JOB 18] MANIFEST created 000022
+[JOB 18] flushed 1 ingested flushable L0:000017 (826 B) in 1.0s (2.0s total), output rate 826 B/s
+remove: db/MANIFEST-000014
+[JOB 18] MANIFEST deleted 000014
+[JOB 19] flushing 1 memtable to L0
+sync: db/MANIFEST-000022
+[JOB 19] flush error: pebble: empty table
```
nicktrav, line 297: #2388 should solve for these errors. We can follow up once a fix for that lands.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2389)* status: 0 of 9 files reviewed, 6 unresolved discussions (waiting on @bananabrick and @nicktrav)

___
*[`compaction.go` line 1855 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2389#-NQ6fwVy8i51-R47piu7:-NQ6fwVy8i51-R47piu8:b-j6zl5h) ([raw file](https://github.com/cockroachdb/pebble/blob/8b1a42184b445bdba0a7268257f6a07224d8c230/compaction.go#L1855)):*
> ```Go
> 		levelMetrics.TablesIngested++
> 		d.mu.versions.metrics.Flush.AsIngestCount++
> 		d.mu.versions.metrics.Flush.AsIngestBytes += file.Size
> ```

we're making the assumption here that this version edit _will_ commit. Other metrics we accumulate in `c.metrics` and only update `d.mu.versions.metrics` when the version edit is committed.

I think this structure is a problem if there are multiple ingested flushables and `ingestTargetLevel` fails on a latter one. I think we'll also need to buffer these new metrics on the `*compaction` and apply them to the version set on version edit commit.
___
*[`event.go` line 190 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2389#-NQ6hpKhBgOeHK0jBuzC:-NQ6hpKhBgOeHK0jBuzD:b7s4dhl) ([raw file](https://github.com/cockroachdb/pebble/blob/8b1a42184b445bdba0a7268257f6a07224d8c230/event.go#L190)):*
> ```Go
> 			w.Printf(" to L0")
> 		} else {
> 			w.Printf("flushing ingested tables")
> ```

should we use `plural` and `i.Input` in the ingested flushables log message too?
___
*[`event.go` line 198 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2389#-NQ6iNFP1jeurFieLmdF:-NQ6iNFP1jeurFieLmdG:bdpodj0) ([raw file](https://github.com/cockroachdb/pebble/blob/8b1a42184b445bdba0a7268257f6a07224d8c230/event.go#L198)):*
> ```Go
> 	if !i.Ingest {
> 		if invariants.Enabled && len(i.IngestLevels) > 0 {
> 			panic(errors.Newf("pebble: expected len(IngestedLevels) == 0"))
> ```

nit: `errors.AssertionFailedf`
___
*[`event.go` line 209 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2389#-NQ6iWwi7V3wMYG35cBL:-NQ6iWwi7V3wMYG35cBM:b-7ropln) ([raw file](https://github.com/cockroachdb/pebble/blob/8b1a42184b445bdba0a7268257f6a07224d8c230/event.go#L209)):*
> ```Go
> 	} else {
> 		if invariants.Enabled && len(i.IngestLevels) == 0 {
> 			panic(errors.Newf("pebble: expected len(IngestedLevels) > 0"))
> ```

ditto
___
*[`event.go` line 214 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2389#-NQ6iaV1B89oON8H9OZG:-NQ6iaV1B89oON8H9OZH:b-l3jui6) ([raw file](https://github.com/cockroachdb/pebble/blob/8b1a42184b445bdba0a7268257f6a07224d8c230/event.go#L214)):*
> ```Go
> 		if len(i.Output) == 1 {
> 			plural = ""
> 		}
> ```

is `i.Input` not set to the number of sstables being ingested as flushable? I notice `plural` should already be initialized based on `i.Input`, which I expected should be equal to `len(i.Output)`.

nicktrav (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2389)* status: 0 of 9 files reviewed, 5 unresolved discussions (waiting on @bananabrick and @jbowens)

___
*[`compaction.go` line 1855 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2389#-NQ6fwVy8i51-R47piu7:-NQDWyXv8Cz-3MNkuQ6p:b-rw2ask) ([raw file](https://github.com/cockroachdb/pebble/blob/8b1a42184b445bdba0a7268257f6a07224d8c230/compaction.go#L1855)):*

we're making the assumption here that this version edit _will_ commit. Other metrics we accumulate in `c.metrics` and only update `d.mu.versions.metrics` when the version edit is committed.

I think this structure is a problem if there are multiple ingested flushables and `ingestTargetLevel` fails on a latter one. I think we'll also need to buffer these new metrics on the `*compaction` and apply them to the version set on version edit commit.
</blockquote></details>

Turns out the metrics were already buffered on the `*c`. I reworked this to move the updating of the metrics until after the version edit has been committed successfully.
___
*[`event.go` line 190 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2389#-NQ6hpKhBgOeHK0jBuzC:-NQDEt6F7gDcjWBc4Z1t:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8b1a42184b445bdba0a7268257f6a07224d8c230/event.go#L190)):*

should we use `plural` and `i.Input` in the ingested flushables log message too?
</blockquote></details>

Done.
___
*[`event.go` line 198 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2389#-NQ6iNFP1jeurFieLmdF:-NQDF72D582DrSljSPPJ:bombmv2) ([raw file](https://github.com/cockroachdb/pebble/blob/8b1a42184b445bdba0a7268257f6a07224d8c230/event.go#L198)):*

nit: `errors.AssertionFailedf`
</blockquote></details>

Good call. Done.
___
*[`event.go` line 209 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2389#-NQ6iWwi7V3wMYG35cBL:-NQDF9-g8sOcKcAnW6Uu:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8b1a42184b445bdba0a7268257f6a07224d8c230/event.go#L209)):*

ditto
</blockquote></details>

Done.
___
*[`event.go` line 214 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2389#-NQ6iaV1B89oON8H9OZG:-NQDWBRT2kr2SeUCs4uK:b6afe25) ([raw file](https://github.com/cockroachdb/pebble/blob/8b1a42184b445bdba0a7268257f6a07224d8c230/event.go#L214)):*
> is i.Input not set to the number of sstables being ingested as flushable?

It was set to `n`, which for a n ingested flushable was always `1`. I tweaked this a bit so as to avoid the need to set `plural` here again.

jbowens (approved):
:lgtm_strong: 

> Reviewed 1 of 9 files at r1, 3 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2389#-:-NQ_zUGHAgDNKwV782YJ:b-uwkowd)* status: 4 of 9 files reviewed, 4 unresolved discussions (waiting on @bananabrick)

nicktrav:
TFTR!

PR #2391, opened 2023-03-13
metamorphic: add RatchetFormatMajorVersion operation

### LINKED ISSUE #2364: metamorphic: test RatchetFormatMajorVersion
The metamorphic tests set a random starting format major version for each Options, but don't exercise ratcheting the format major version during operation. We should work in calls to `RatchetFormatMajorVersion`, but it's a little unclear how exactly. Maybe an operation that ratchets to the next format major version, but is a no-op if already at `FormatNewest`?

jbowens:
Add a new operation that ratchets the format major version of the database while it's running. This mirrors Cocroach's use of the format major version: Cockroach ratchets successive versions as a part of cluster version finalization, while the database is online and compactions, etc, may be executing.

Close #2364.


RaduBerinde (approved):
Very cool! :lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2391#-:-NQQuuUUEkjk-0kTwQmp:b-6tk46f)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`internal/metamorphic/ops.go` line 1148 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2391#-NQQuLIv5osPghzPksJ4:-NQQuLIv5osPghzPksJ5:byrzm9i) ([raw file](https://github.com/cockroachdb/pebble/blob/764ccc16031d6fa01ae01554b456fdb784e52e1b/internal/metamorphic/ops.go#L1148)):*
> ```Go
> 
> func (o *dbRatchetFormatMajorVersionOp) run(t *test, h historyRecorder) {
> 	_ = t.db.RatchetFormatMajorVersion(o.vers)
> ```

I can see ignoring the `database already at format major version` error, but I think any other kind of migration error would be indicative of a bug no?

nicktrav (approved):
Nice!  :lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2391#-:-NQR02Lq7CdmjpQ4u0EN:b-ypbbbp)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

jbowens (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2391#-:-NQRDQ-4ALsZZpxRg82p:bz5mair)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @itsbilal, @nicktrav, and @RaduBerinde)

___
*[`internal/metamorphic/ops.go` line 1148 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2391#-NQQuLIv5osPghzPksJ4:-NQRDUhPCZGT5ge6rJMI:bp9xoh3) ([raw file](https://github.com/cockroachdb/pebble/blob/764ccc16031d6fa01ae01554b456fdb784e52e1b/internal/metamorphic/ops.go#L1148)):*

I can see ignoring the `database already at format major version` error, but I think any other kind of migration error would be indicative of a bug no?
</blockquote></details>

Good call; updated to elide the call to `RatchetFormatMajorVersion` if the current version is already at least as high, and to output the error, if any, otherwise.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2391)* status: 4 of 5 files reviewed, 1 unresolved discussion

___
*[`internal/metamorphic/ops.go` line 1148 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2391#-NQQuLIv5osPghzPksJ4:-NQRGYc-3vmKDou5ssfA:b-3uas5s) ([raw file](https://github.com/cockroachdb/pebble/blob/764ccc16031d6fa01ae01554b456fdb784e52e1b/internal/metamorphic/ops.go#L1148)):*

Trying to understand more about the error handling in the metamorphic test - what does recording it do? We still effectively ignore it (and effectively test that it doesn't negatively affect subsequent operations) no? I think that in this case, there is no acceptable error here - any error is an indication of a bug in the migrations and the test should fail on the spot.
</blockquote></details>

The metamorphic tests assert vary little. They work by recording the output, and ensuring that the output matches between many different independent runs with varying underlying mechanics. If one of the tests encounters an error, it is very likely that another of the ~30-40 runs (some of which no-oped this operation) will not. This produces a divergence, which fails the test.

It's nice to rely on the metamorphism to surface bugs, because it adapts as the code changes (will it always be the case that there are no other legitimate errors to be returned here?), and it avoids the need to produce an 'oracle', including asserting every invariant.

RaduBerinde:
*[`internal/metamorphic/ops.go` line 1148 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2391#-NQQuLIv5osPghzPksJ4:-NQRKNAg5UoUri6sqVVe:b-7c8ppr) ([raw file](https://github.com/cockroachdb/pebble/blob/764ccc16031d6fa01ae01554b456fdb784e52e1b/internal/metamorphic/ops.go#L1148)):*

The metamorphic tests assert vary little. They work by recording the output, and ensuring that the output matches between many different independent runs with varying underlying mechanics. If one of the tests encounters an error, it is very likely that another of the ~30-40 runs (some of which no-oped this operation) will not. This produces a divergence, which fails the test.

It's nice to rely on the metamorphism to surface bugs, because it adapts as the code changes (will it always be the case that there are no other legitimate errors to be returned here?), and it avoids the need to produce an 'oracle', including asserting every invariant.
</blockquote></details>

I see, very cool! Thanks!

PR #2393, opened 2023-03-14
sstable: fix RewriteKeySuffixes to disallow format changes

sumeerbhola:
The code was broken both for v2=>v3 and v3=>v2 transitions. The modified test would catch both bugs. The test now also verifies that the vanilla rewrite and this optimized rewrite produce the same output.

Informs https://github.com/cockroachdb/cockroach/issues/97076

### DIFF
--- a/sstable/data_test.go
+++ b/sstable/data_test.go
@@ -425,7 +425,7 @@ func runRewriteCmd(
 	}
 
 	f := &memFile{}
-	meta, err := rewriteKeySuffixesInBlocks(r, f, opts, from, to, 2)
+	meta, _, err := rewriteKeySuffixesInBlocks(r, f, opts, from, to, 2)
 	if err != nil {
 		return nil, r, errors.Wrap(err, "rewrite failed")
 	}
--- a/sstable/suffix_rewriter.go
+++ b/sstable/suffix_rewriter.go
@@ -9,24 +9,14 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/bytealloc"
+	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/internal/rangekey"
 	"github.com/cockroachdb/pebble/objstorage"
 )
 
-// RewriteKeySuffixes copies the content of the passed SSTable bytes to a new
-// sstable, written to `out`, in which the suffix `from` has is replaced with
-// `to` in every key. The input sstable must consist of only Sets or RangeKeySets
-// and every key must have `from` as its suffix as determined by the Split
-// function of the Comparer in the passed WriterOptions. Range deletes must not
-// exist in this sstable, as they will be ignored.
+// RewriteKeySuffixes is deprecated.
 //
-// Data blocks are rewritten in parallel by `concurrency` workers and then
-// assembled into a final SST. Filters are copied from the original SST without
-// modification as they are not affected by the suffix, while block and table
-// properties are only minimally recomputed.
-//
-// Any block and table property collectors configured in the WriterOptions must
-// implement SuffixReplaceableTableCollector/SuffixReplaceableBlockCollector.
+// TODO(sumeer): remove after switching CockroachDB to RewriteKeySuffixesAndReturnFormat.
 func RewriteKeySuffixes(
 	sst []byte,
 	rOpts ReaderOptions,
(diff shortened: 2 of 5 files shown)


sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2393)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @nicktrav and @RaduBerinde)

___
*[`sstable/suffix_rewriter.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2393#-NQWMWV8-ddSi45absDm:-NQWMWV8-ddSi45absDn:b-1nndit) ([raw file](https://github.com/cockroachdb/pebble/blob/0105c99e551e0bf7a74b2a8ee50ea5429228b9e7/sstable/suffix_rewriter.go#L47)):*
> ```Go
> // implement SuffixReplaceableTableCollector/SuffixReplaceableBlockCollector.
> //
> // The WriterOptions.TableFormat is ignored, and the output sstable has the
> ```

I considered allowing a change in the format, but it is not needed in CockroachDB, and is more error prone.

RaduBerinde (approved):
:lgtm:

This needs to be backported to the release branch, correct?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2393#-:-NQWOJnD7GLNUC7QCSg9:bo2yw23)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @nicktrav and @sumeerbhola)

___
*[`sstable/suffix_rewriter.go` line 18 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2393#-NQWNsc77fpiVEvIMKRW:-NQWNsc77fpiVEvIMKRX:baqx6sa) ([raw file](https://github.com/cockroachdb/pebble/blob/0105c99e551e0bf7a74b2a8ee50ea5429228b9e7/sstable/suffix_rewriter.go#L18)):*
> ```Go
> // RewriteKeySuffixes is deprecated.
> //
> // TODO(sumeer): remove after switching CockroachDB to RewriteKeySuffixesAndReturnFormat.
> ```

If we intend to remove, why not remove it now (or keep the name and add the return value)? It's only used in one place in CRDB so it will be easy enough to change when we bump pebble.
___
*[`sstable/suffix_rewriter.go` line 72 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2393#-NQWOuQ7AghaWfsTOnxL:-NQWOuQ7AghaWfsTOnxM:b-n7lcrd) ([raw file](https://github.com/cockroachdb/pebble/blob/0105c99e551e0bf7a74b2a8ee50ea5429228b9e7/sstable/suffix_rewriter.go#L72)):*
> ```Go
> 	if o.Comparer == nil || o.Comparer.Split == nil {
> 		return nil, TableFormatUnspecified,
> 			errors.New("a valid splitter is required to define suffix to replace replace suffix")
> ```

[nit] this error message is unreadable, maybe "a valid splitter is required to rewrite suffixes"

RaduBerinde:
*[`sstable/suffix_rewriter.go` line 222 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2393#-NQWPBvQ3nKRPA6adu3H:-NQWPBvQ3nKRPA6adu3I:b-wxmh4r) ([raw file](https://github.com/cockroachdb/pebble/blob/0105c99e551e0bf7a74b2a8ee50ea5429228b9e7/sstable/suffix_rewriter.go#L222)):*
> ```Go
> 			// !iter.lazyValueHandling.hasValuePrefix, it will return the raw value
> 			// in the block, which includes the 1-byte prefix. This is fine since bw
> 			// also does not know about the prefix and will preserve it in bw.add.
> ```

Maybe we can have an `invariants` check that if this is v3, the first value byte makes sense? (assuming not all 256 values are valid for it)

sumeerbhola (commented):
TFTR!

> This needs to be backported to the release branch, correct?

Correct

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2393#-:-NQWiJeC91YM19xgE6tL:by1agho)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @nicktrav and @RaduBerinde)

___
*[`sstable/suffix_rewriter.go` line 18 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2393#-NQWNsc77fpiVEvIMKRW:-NQWiQqf6EKMn9MfULc6:b-39r2gu) ([raw file](https://github.com/cockroachdb/pebble/blob/0105c99e551e0bf7a74b2a8ee50ea5429228b9e7/sstable/suffix_rewriter.go#L18)):*

If we intend to remove, why not remove it now (or keep the name and add the return value)? It's only used in one place in CRDB so it will be easy enough to change when we bump pebble.
</blockquote></details>

I don't want to create extra work for whoever happens to be doing the bump.
___
*[`sstable/suffix_rewriter.go` line 72 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2393#-NQWOuQ7AghaWfsTOnxL:-NQWiwGA6h2nN4coqGzg:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/0105c99e551e0bf7a74b2a8ee50ea5429228b9e7/sstable/suffix_rewriter.go#L72)):*

[nit] this error message is unreadable, maybe "a valid splitter is required to rewrite suffixes"
</blockquote></details>

Done
___
*[`sstable/suffix_rewriter.go` line 222 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2393#-NQWPBvQ3nKRPA6adu3H:-NQX9oNA0y7wjQktH1K6:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/0105c99e551e0bf7a74b2a8ee50ea5429228b9e7/sstable/suffix_rewriter.go#L222)):*

Maybe we can have an `invariants` check that if this is v3, the first value byte makes sense? (assuming not all 256 values are valid for it)
</blockquote></details>

Done

PR #2396, opened 2023-03-15
sstable: fix RewriteKeySuffixes to disallow format changes

sumeerbhola:
Backport of commit from https://github.com/cockroachdb/pebble/pull/2393
***
The code was broken both for v2=>v3 and v3=>v2 transitions. The modified test would catch both bugs. The test now also verifies that the vanilla rewrite and this optimized rewrite produce the same output.

Informs https://github.com/cockroachdb/cockroach/issues/97076

### DIFF
--- a/sstable/data_test.go
+++ b/sstable/data_test.go
@@ -425,7 +425,7 @@ func runRewriteCmd(
 	}
 
 	f := &memFile{}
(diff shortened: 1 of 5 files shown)


jbowens (approved):
:lgtm:

sumeerbhola:
TFTR!

PR #2400, opened 2023-03-15
crl-release-23.1: db: add metrics for flushable ingests; improve logging

nicktrav:
This is a backport of #2389 to 23.1.

---

The new flushable ingest is designed to allow tables to be lazily ingested in the case where there is overlap with the memtable. To provide visibility into how the feature is being used, add two new flush metrics - a counter for each flushable ingest, and a counter for the number of bytes ingested as flushables.

Add the two new counts to the `Metrics` output. The formatting was altered slightly to improve the alignment.

Enhance the event logging for flushable ingests by printing the level into which the table is ingested.


jbowens (approved):
:lgtm:

nicktrav:
TFTR!

PR #2399, opened 2023-03-15
vfs: Add support for OpenBSD

lcheylus:
- Add `vfs/disk_usage_openbsd.go` to use `unix.Statfs` with specific fields for OpenBSD => see https://cs.opensource.google/go/x/sys/+/master:unix/ztypes_openbsd_amd64.go
  - Fix bug #2341

Tests are OK with `make test`

PR #2404, opened 2023-03-16
ci: add build for OpenBSD

lcheylus:
Add a new target for build on OpenBSD : duplicate of FreeBSD/NetBSD build with `GOOS: openbsd`


lcheylus:
Closes this PR => merge with https://github.com/cockroachdb/pebble/pull/2399

PR #2408, opened 2023-03-20
db: ingest-as-flushable metric and logging tweaks

nicktrav:
**db: tweak ingest-as-flushable metrics**

Currently, the metric for ingest-as-flushable counts the number of
tables ingested, rather than the number of flushes that are handling
ingested tables. This can be confusing to the operator.

Switch the existing metric to count the flushes handling ingested
tables.

Introduce a separate metric tracking the count of tables ingested as
flushables.

---

**db: fix redaction of ingest-as-flushable filenum**

The filenum is currently not marked as safe for redaction, which results
in the field being delimited with redaction markers in Cockroach log
files.

As the file number is not sensitive, mark the field as safe.

PR #2410, opened 2023-03-21
crl-release-23.1: flushable metrics fixes

nicktrav:
This is a backport of #2408 to 23.1.

---

**db: tweak ingest-as-flushable metrics**

Currently, the metric for ingest-as-flushable counts the number of
tables ingested, rather than the number of flushes that are handling
ingested tables. This can be confusing to the operator.

Switch the existing metric to count the flushes handling ingested
tables.

Introduce a separate metric tracking the count of tables ingested as
flushables.

---

**db: fix redaction of ingest-as-flushable filenum**

The filenum is currently not marked as safe for redaction, which results
in the field being delimited with redaction markers in Cockroach log
files.

As the file number is not sensitive, mark the field as safe.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2045,9 +2045,10 @@ func (d *DB) flush1() (bytesFlushed uint64, err error) {
 		d.updateReadStateLocked(d.opts.DebugCheck)
 		d.updateTableStatsLocked(ve.NewFiles)
 		if ingest {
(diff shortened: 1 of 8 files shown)


itsbilal (approved):
LGTM

nicktrav:
TFTR!

PR #2411, opened 2023-03-22
db: adjust range-del bytes estimate for L6 files

jbowens:
Previously, a file in L6 that contained range deletion(s) was assumed to contain range deletions due to open snapshots at the time it was written. Table stats, and in particular the RangeDeletionBytesEstimate value, was calculated assuming the range deletion dropped all keys within the sstable that fall within the range deletion's bounds. This is a decent heuristic for when snapshots were indeed open, preventing the elision of the range deletion.

However, Cockroach KV snapshot reception writes range deletions to each of the ingested sstables (see cockroachdb/cockroach#44048). These range deletions delete nothing within the tables themselves. They exist only to clear out existing state below the table in the LSM. In these cases, the RangeDeletionBytesEstimate value would be a gross overestimate. Compacting these tables would only drop the tombstones themselves and none of the actual range data contained in point blocks.

This commit adjusts the logic for calculating RangeDeletionBytesEstimate to take the range deletion sequence number into account. This small change will prevent sstables ingested through Cockroach snapshot reception from having a RangeDeletionBytesEstimate that incorrectly sums all data blocks. This has two consequences:

a) In times of low compaction pressure, we will no longer rewrite the ingested sstables through elision-only compactions simply to remove the tombstones. b) When compacting from L5 into L6, we will no longer artificially deflate these sstable's size when calculating the min-overlapping ratio heuristic.

Together, these effects should reduce write amplification/bandwidth, especially in the presence of significant rebalancing.

### DIFF
--- a/internal/manifest/version.go
+++ b/internal/manifest/version.go
@@ -60,9 +60,11 @@ type TableStats struct {
 	// overlapping data in L0 and ignores L0 sublevels, but the error that
 	// introduces is expected to be small.
 	//
-	// Tables in the bottommost level of the LSM may have a nonzero estimate
-	// if snapshots or move compactions prevented the elision of their range
-	// tombstones.
+	// Tables in the bottommost level of the LSM may have a nonzero estimate if
+	// snapshots or move compactions prevented the elision of their range
+	// tombstones. A table in the bottommost level that was ingested into L6
+	// will have a zero estimate, because the file's sequence numbers indicate
+	// that the tombstone cannot drop any data contained within the file itself.
 	RangeDeletionsBytesEstimate uint64
 	// Total size of value blocks and value index block.
 	ValueBlocksSize uint64
--- a/table_stats.go
+++ b/table_stats.go
@@ -320,25 +320,50 @@ func (d *DB) loadTableRangeDelStats(
 	for s := iter.First(); s != nil; s = iter.Next() {
 		start, end := s.Start, s.End
 		// We only need to consider deletion size estimates for tables that contain
-		// point keys.
-		var hasPoints bool
+		// RANGEDELs.
+		var maxRangeDeleteSeqNum uint64
 		for _, k := range s.Keys {
-			if k.Kind() == base.InternalKeyKindRangeDelete {
-				hasPoints = true
+			if k.Kind() == base.InternalKeyKindRangeDelete && maxRangeDeleteSeqNum < k.SeqNum() {
+				maxRangeDeleteSeqNum = k.SeqNum()
 				break
 			}
 		}
 
 		// If the file is in the last level of the LSM, there is no data beneath
 		// it. The fact that there is still a range tombstone in a bottommost file
-		// suggests that an open snapshot kept the tombstone around. Estimate disk
-		// usage within the file itself.
+		// indicates two possibilites:
+		//   1. an open snapshot kept the tombstone around, and the data the
+		//      tombstone deletes is contained within the file itself.
+		//   2. the file was ingested.
+		// In the first case, we'd like to estimate disk usage within the file
+		// itself since compacting the file will drop that covered data. In the
+		// second case, we expect that compacting the file will NOT drop any
+		// data and rewriting the file is a waste of write bandwidth. We can
+		// distinguish these cases by looking at the file metadata's sequence
+		// numbers. A file's range deletions can only delete data within the
+		// file at lower sequence numbers. All keys in an ingested sstable adopt
+		// the same sequence number, preventing tombstones from deleting keys
+		// within the same file. We check here if the largest RANGEDEL sequence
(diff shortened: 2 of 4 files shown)


jbowens (commented):
I expect this to be a high reward and very low risk change, so I would like to sneak it into 23.1 (and even entertain backporting to 22.2).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2411#-:-NRA5Jz99rTwj1moGfPD:b-ka0vjj)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @sumeerbhola)

jbowens (commented):
This came out of inspecting the 23.1 test cluster and observing high elision-only compaction counts:

```
Store 1:
__level_____count____size___score______in__ingest(sz_cnt)____move(sz_cnt)___write(sz_cnt)____read___r-amp___w-amp__val-bl
    WAL         1    19 M       -   182 G       -       -       -       -   179 G       -       -       -     1.0
      0         0     0 B    0.00   179 G   1.0 G   183 K     0 B       0    23 G    84 K    78 M       0     0.1     0 B
      1         0     0 B    0.00     0 B     0 B       0     0 B       0     0 B       0     0 B       0     0.0     0 B
      2         0     0 B    0.00   358 M   359 M      90     0 B       0   595 M     204   608 M       0     1.7     0 B
      3        22    62 M    0.98    14 G   1.3 G   2.3 K    10 G   3.9 K   404 G   152 K   409 G       1    28.1   7.2 M
      4       118   603 M    1.00    18 G   3.8 G    26 K   4.0 G   1.4 K    64 G    14 K    71 G       1     3.6    83 M
      5       578   4.9 G    1.00    17 G   3.9 G    17 K   2.6 G   1.3 K    54 G   6.9 K    60 G       1     3.1   361 M
      6      2608    53 G       -    25 G   5.5 G    41 K   1.7 G   1.5 K    64 G   2.6 K    76 G       1     2.5   215 M
  total      3326    59 G       -   195 G    16 G   270 K    19 G   8.2 K   805 G   260 K   616 G       4     4.1   666 M
  flush     72483                          1019 M  199326          (ingest = ingested-as-flushable)
compact     94170     0 B     0 B       0                          (size == estimated-debt, score = in-progress-bytes, in = num-in-progress)
  ctype     43557     102   42333    8172       6       0       0  (default, delete, elision, move, read, rewrite, multi-level)
 memtbl         1    64 M
zmemtbl         0     0 B
   ztbl         0     0 B
 bcache     466 K   7.4 G   99.5%  (score == hit-rate)
 tcache     3.3 K   2.2 M  100.0%  (score == hit-rate)
  snaps         0       -       0  (score == earliest seq num)
 titers         0
 filter         -       -   92.8%  (score == utility)
```

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2411#-:-NRA5oov4ZdvqsOa02Mh:bkm3yhh)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @sumeerbhola)

sumeerbhola (approved):
nice find!

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2411#-:-NRZTvZCCTLefHTpam8u:b-udnwza)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`table_stats.go` line 349 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2411#-NRZU8s35Gxp-7Ev8CYD:-NRZU8s35Gxp-7Ev8CYE:b-zdjpm) ([raw file](https://github.com/cockroachdb/pebble/blob/52946f53ffb86b38e9cbb0ef9cc37c280ebcfc73/table_stats.go#L349)):*
> ```Go
> 		// number is greater than the file's smallest sequence number. If it is,
> 		// the RANGEDEL could conceivably (although inconclusively) delete data
> 		// within the same file.
> ```

Probably worth adding in the comment that if the ingest happened into L5 and then a compaction into L6 kept the deleted data already in L6 due to a snapshot then this heuristic would not work in that we would count everything overlapping in `RangeDeletionsBytesEstimate`.
___
*[`table_stats.go` line 361 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2411#-NRZTI8P5g00phaodPLF:-NRZTI8P5g00phaodPLG:bvc1gsd) ([raw file](https://github.com/cockroachdb/pebble/blob/52946f53ffb86b38e9cbb0ef9cc37c280ebcfc73/table_stats.go#L361)):*
> ```Go
> 				return nil, err
> 			}
> 			stats.RangeDeletionsBytesEstimate += size
> ```

worth updating the comment for RangeDeletionsBytesEstimate
```
	// Tables in the bottommost level of the LSM may have a nonzero estimate
	// if snapshots or move compactions prevented the elision of their range
	// tombstones.
```

PR #2416, opened 2023-03-24
objstorage: extract Provider interface, move implementation

RaduBerinde:
This change moves the provider implementation to a separate package and leaves behind in `objstorage` just its public interface.

The main motivation is technical - for obj IO tracing, the tracer needs to use things like `Readable`, `Writable` and the provider needs to use the tracer.

But it is also more clarifying to have an explicit interface and it makes it possible to move the provider implementation outside of pebble if we choose to do so (there were initial debates about where it should live).

This commit should not contain any functional changes.


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2416)* status: 0 of 51 files reviewed, 2 unresolved discussions (waiting on @itsbilal)

___
*[`objstorage/objstorage.go` line 122 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2416#-NRK1o1A7Z5nzOvJkhsW:-NRK1o1A7Z5nzOvJkhsX:b-pvlz0v) ([raw file](https://github.com/cockroachdb/pebble/blob/aa9837237ffada1f650aa9179a7a2efc579ac542/objstorage/objstorage.go#L122)):*
> ```Go
> // This ID is incorporated in backing object names.
> // Must be non-zero.
> type CreatorID uint64
> ```

This part is nicer too - instead of defining this deep into the catalog package and aliasing it, the root `objstorage` acts as a "base" package for everything inside.
___
*[`objstorage/objstorage.go` line 154 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2416#-NRK1x7J6nDvNJ8JckFw:-NRK1x7K5IxQXQ7FfPq6:b-bt1y61) ([raw file](https://github.com/cockroachdb/pebble/blob/aa9837237ffada1f650aa9179a7a2efc579ac542/objstorage/objstorage.go#L154)):*
> ```Go
> //
> // Objects are currently backed by a vfs.File or a shared.Storage object.
> type Provider interface {
> ```

This is the main thing worth reviewing.

sumeerbhola (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2416)* status: all files reviewed, 4 unresolved discussions (waiting on @itsbilal)

PR #2417, opened 2023-03-27
objstorage: implement tracing

RaduBerinde:
This commit adds an obj IO tracing subsystem. It works by wrapping `Readable` and `Writable` in implementations which generate tracing events. The events are dumped to `IOTRACE-<timestamp>` files in the store directory. Contexts are used to plumb various details, like LSM level or block type.

The tracing code is behind the `pebble_obj_io_tracing` build tag; this is to avoid causing any regressions to normal build.

The plumbing of details through contexts is not (nor it is intended to be) complete.

A new make target `testobjiotracing` runs the test (which requires the `pebble_obj_io_tracing` tag).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,10 @@ testmsan: export CC=clang
 testmsan: testflags += -msan -timeout 20m
 testmsan: test
 
+.PHONY: testobjiotracing
+testobjiotracing:
+	${GO} test -tags '$(TAGS) pebble_obj_io_tracing' ${testflags} -run ${TESTS} ./objstorage/objstorageprovider/objiotracing
+
 .PHONY: lint
 lint:
 	${GO} test -tags '$(TAGS)' ${testflags} -run ${TESTS} ./internal/lint
--- a/compaction.go
+++ b/compaction.go
@@ -25,6 +25,7 @@ import (
 	"github.com/cockroachdb/pebble/internal/rangedel"
 	"github.com/cockroachdb/pebble/internal/rangekey"
 	"github.com/cockroachdb/pebble/objstorage"
+	"github.com/cockroachdb/pebble/objstorage/objstorageprovider/objiotracing"
 	"github.com/cockroachdb/pebble/sstable"
 	"github.com/cockroachdb/pebble/vfs"
 )
@@ -2748,7 +2749,19 @@ func (d *DB) runCompaction(
 		pendingOutputs = append(pendingOutputs, fileMeta)
 		d.mu.Unlock()
 
-		writable, objMeta, err := d.objProvider.Create(context.TODO(), fileTypeTable, fileNum, objstorage.CreateOptions{} /* TODO */)
+		ctx := context.TODO()
+		if objiotracing.Enabled {
+			ctx = objiotracing.WithLevel(ctx, c.outputLevel.level)
+			switch c.kind {
+			case compactionKindFlush:
+				ctx = objiotracing.WithReason(ctx, objiotracing.ForFlush)
+			case compactionKindIngestedFlushable:
+				ctx = objiotracing.WithReason(ctx, objiotracing.ForIngestion)
+			default:
+				ctx = objiotracing.WithReason(ctx, objiotracing.ForCompaction)
+			}
+		}
+		writable, objMeta, err := d.objProvider.Create(ctx, fileTypeTable, fileNum, objstorage.CreateOptions{} /* TODO */)
 		if err != nil {
 			return err
 		}
--- a/objstorage/objstorage.go
+++ b/objstorage/objstorage.go
@@ -71,7 +71,8 @@ type ReadHandle interface {
 	MaxReadahead()
 
 	// RecordCacheHit informs the implementation that we were able to retrieve a
-	// block from cache.
+	// block from cache. This is useful for example when the implementation is
+	// trying to detect a sequential reading pattern.
 	RecordCacheHit(ctx context.Context, offset, size int64)
 }
 
--- a/objstorage/objstorageprovider/objiotracing/obj_io_tracing.go
+++ b/objstorage/objstorageprovider/objiotracing/obj_io_tracing.go
@@ -1,0 +1,68 @@
+// Copyright 2023 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package objiotracing
+
+import "github.com/cockroachdb/pebble/internal/base"
+
+// OpType indicates the type of operation.
+type OpType uint8
+
+// OpType values.
+const (
+	ReadOp OpType = iota
+	WriteOp
+	// RecordCacheHitOp happens when a read is satisfied from the block cache. See
+	// objstorage.ReadHandle.RecordCacheHit().
+	RecordCacheHitOp
+	// MaxReadaheadOp is a "meta operation" that configures a read handle for
+	// large sequential reads. See objstorage.ReadHandle.MaxReadahead().
+	MaxReadaheadOp
+)
+
+// Reason indicates the higher-level context of the operation.
+type Reason uint8
+
+// Reason values.
+const (
+	UnknownReason Reason = iota
+	ForFlush
+	ForCompaction
+	ForIngestion
+	// TODO(radu): add ForUserFacing.
+)
+
+// BlockType indicates the type of data block relevant to an operation.
+type BlockType uint8
+
+// BlockType values.
+const (
+	UnknownBlock BlockType = iota
+	DataBlock
+	ValueBlock
+	FilterBlock
+	MetadataBlock
+)
+
+// Event is the on-disk format of a tracing event. It is exported here so that
+// trace processing tools can use it by importing this package.
+type Event struct {
+	// Event start time as a Unix time (see time.Time.StartUnixNano()).
+	// Note that recorded events are not necessarily ordered by time - this is
+	// because separate event "streams" use local buffers (for performance).
+	StartUnixNano int64
+	Op            OpType
+	Reason        Reason
+	BlockType     BlockType
+	// LSM level plus one (with 0 indicating unknown level).
+	LevelPlusOne uint8
+	// Hardcoded padding so that struct layout doesn't depend on architecture.
+	_       uint32
+	FileNum base.FileNum
+	// HandleID is a unique identifier corresponding to an objstorage.ReadHandle;
+	// only set for read operations performed through a ReadHandle.
+	HandleID uint64
+	Offset   int64
+	Size     int64
+}
--- a/objstorage/objstorageprovider/objiotracing/obj_io_tracing_off.go
+++ b/objstorage/objstorageprovider/objiotracing/obj_io_tracing_off.go
@@ -1,0 +1,59 @@
+// Copyright 2023 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+//go:build !pebble_obj_io_tracing
+// +build !pebble_obj_io_tracing
+
+package objiotracing
+
+import (
+	"context"
+
+	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/objstorage"
+	"github.com/cockroachdb/pebble/vfs"
+)
+
+// Enabled is used to short circuit tracing-related code in regular builds.
+const Enabled = false
+
+// Tracer manages the writing of object IO traces to files.
+type Tracer struct{}
+
+// Open creates a Tracer which generates trace files in the given directory.
+// Each trace file contains a series of Events (as they are in memory).
+func Open(fs vfs.FS, fsDir string) *Tracer {
+	return nil
+}
+
+// Close the tracer, flushing any remaining events.
+func (*Tracer) Close() {}
+
+// WrapReadable wraps an objstorage.Readable with one that generates tracing
+// events.
+func (*Tracer) WrapReadable(
+	ctx context.Context, r objstorage.Readable, fileNum base.FileNum,
+) objstorage.Readable {
+	return r
+}
+
+// WrapWritable wraps an objstorage.Writable with one that generates tracing
+// events.
+func (t *Tracer) WrapWritable(
+	ctx context.Context, w objstorage.Writable, fileNum base.FileNum,
+) objstorage.Writable {
+	return w
+}
+
+// WithReason creates a context that has an associated Reason (which ends up in
+// traces created under that context).
+func WithReason(ctx context.Context, reason Reason) context.Context { return ctx }
+
+// WithBlockType creates a context that has an associated BlockType (which ends up in
+// traces created under that context).
+func WithBlockType(ctx context.Context, blockType BlockType) context.Context { return ctx }
+
+// WithLevel creates a context that has an associated level (which ends up in
+// traces created under that context).
+func WithLevel(ctx context.Context, level int) context.Context { return ctx }
--- a/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go
+++ b/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go
@@ -1,0 +1,405 @@
+// Copyright 2023 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+//go:build pebble_obj_io_tracing
+// +build pebble_obj_io_tracing
+
+package objiotracing
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"math/rand"
+	"sync"
+	"sync/atomic"
+	"time"
+	"unsafe"
+
+	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/objstorage"
+	"github.com/cockroachdb/pebble/vfs"
+)
+
+// Enabled is used to short circuit tracing-related code in regular builds.
+const Enabled = true
+
+// Tracer manages the writing of object IO traces to files.
+//
+// The tracer runs a background worker goroutine which receives trace event
+// buffers over a channel and dumps them to IOTRACES- files. Wrapper
+// implementations for Readable, ReadHandle, Writable are producers of traces;
+// they maintain internal buffers of events which get flushed to the buffered
+// channel when they get full. This allows for minimal synchronization per IO
+// (as for most of these structures, an instance only allows a single IO at a
+// time).
+type Tracer struct {
+	fs    vfs.FS
+	fsDir string
+
+	handleID atomic.Uint64
+
+	workerStopCh chan struct{}
+	workerDataCh chan eventBuf
+	workerWait   sync.WaitGroup
+}
+
+// Open creates a Tracer which generates trace files in the given directory.
+// Each trace file contains a series of Events (as they are in memory).
+func Open(fs vfs.FS, fsDir string) *Tracer {
+	t := &Tracer{
+		fs:           fs,
+		fsDir:        fsDir,
+		workerStopCh: make(chan struct{}),
+		workerDataCh: make(chan eventBuf, channelBufSize),
+	}
+
+	t.handleID.Store(uint64(rand.NewSource(time.Now().UnixNano()).Int63()))
+
+	t.workerWait.Add(1)
+	go t.workerLoop()
+	return t
+}
+
+// Close the tracer, flushing any remaining events.
+func (t *Tracer) Close() {
+	if t.workerStopCh == nil {
+		return
+	}
+	// Tell the worker to stop and wait for it to finish up.
+	close(t.workerStopCh)
+	t.workerWait.Wait()
+	t.workerStopCh = nil
+}
+
+// WrapWritable wraps an objstorage.Writable with one that generates tracing
+// events.
+func (t *Tracer) WrapWritable(
+	ctx context.Context, w objstorage.Writable, fileNum base.FileNum,
+) objstorage.Writable {
+	return &writable{
+		w:       w,
+		fileNum: fileNum,
+		g:       makeEventGenerator(ctx, t),
+	}
+}
+
+type writable struct {
+	w         objstorage.Writable
+	fileNum   base.FileNum
+	curOffset int64
+	g         eventGenerator
+}
+
+var _ objstorage.Writable = (*writable)(nil)
+
+// Write is part of the objstorage.Writable interface.
+func (w *writable) Write(p []byte) error {
+	w.g.add(context.Background(), Event{
+		Op:      WriteOp,
+		FileNum: w.fileNum,
+		Offset:  w.curOffset,
+		Size:    int64(len(p)),
+	})
+	return w.w.Write(p)
+}
+
+// Finish is part of the objstorage.Writable interface.
+func (w *writable) Finish() error {
+	w.g.flush()
+	return w.w.Finish()
+}
+
+// Abort is part of the objstorage.Writable interface.
+func (w *writable) Abort() {
+	w.g.flush()
+	w.w.Abort()
+}
+
+// WrapReadable wraps an objstorage.Readable with one that generates tracing
+// events.
+func (t *Tracer) WrapReadable(
+	ctx context.Context, r objstorage.Readable, fileNum base.FileNum,
+) objstorage.Readable {
+	res := &readable{
+		r:       r,
+		fileNum: fileNum,
+	}
+	res.mu.g = makeEventGenerator(ctx, t)
+	return res
+}
+
+type readable struct {
+	r       objstorage.Readable
+	fileNum base.FileNum
+	mu      struct {
+		sync.Mutex
+		g eventGenerator
+	}
+}
+
+var _ objstorage.Readable = (*readable)(nil)
+
+// ReadAt is part of the objstorage.Readable interface.
+func (r *readable) ReadAt(ctx context.Context, v []byte, off int64) (n int, err error) {
+	r.mu.Lock()
+	r.mu.g.add(ctx, Event{
+		Op:      ReadOp,
+		FileNum: r.fileNum,
+		Offset:  off,
+		Size:    int64(len(v)),
+	})
+	r.mu.Unlock()
+	return r.r.ReadAt(ctx, v, off)
+}
+
+// Close is part of the objstorage.Readable interface.
+func (r *readable) Close() error {
+	r.mu.g.flush()
+	return r.r.Close()
+}
+
+// Size is part of the objstorage.Readable interface.
+func (r *readable) Size() int64 {
+	return r.r.Size()
+}
+
+// NewReadHandle is part of the objstorage.Readable interface.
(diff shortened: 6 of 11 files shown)


joshimhoff (approved):
Love it! I am giving approval, but the parts of this PR that touch existing pebble code (e.g. in `reader.go` and `compaction.go`) are hard for me to review for correctness.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2417#-:-NRd_nlX9Ei1jxiPOKZD:b-hazhae)* status: 0 of 10 files reviewed, 17 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`table_cache.go` line 418 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRcNSOB9mAPaQksOhtn:-NRcNSOB9mAPaQksOhto:b-if9gub) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/table_cache.go#L418)):*
> ```Go
> 	var iter sstable.Iterator
> 	useFilter := true
> 	if opts != nil {
> ```

When is this `nil`?
___
*[`objstorage/objstorageprovider/provider.go` line 154 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRcNboz23SwFuuboqNi:-NRcNboz23SwFuuboqNj:b2me1lf) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/provider.go#L154)):*
> ```Go
> 	}
> 	if objiotracing.Enabled {
> 		if p.tracer != nil {
> ```

If `Enabled`, don't we expect `p.tracer` to be not equal to `nil`?
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing.go` line 16 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdKNGqClsCK45Kqwo5:-NRdKNGqClsCK45Kqwo6:b-zbth7w) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing.go#L16)):*
> ```Go
> 	ReadOp OpType = iota
> 	WriteOp
> 	RecordCacheHitOp
> ```

Can you say more about `RecordCacheHitOp` and `MaxReadaheadOp` in comments? I think I understand this but only bc/ we talked in 1:1.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing.go` line 25 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdKW0JAqRQ-nk8NPn3:-NRdKW0JAqRQ-nk8NPn4:b-iho4yx) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing.go#L25)):*
> ```Go
> // Reason values.
> const (
> 	UnknownReason Reason = iota
> ```

Do you have any exceptions about how often we will see the `UnknownReason` reason?
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdLAs8Cbn8QFQ72uL0:-NRdL5hv0dQAA9cOI1tK:bbm8p0x) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing.go#L28)):*
> ```Go
> 	ForFlush
> 	ForCompaction
> 	ForIngestion
> ```

Don't we want a `Reason` called `FromUserFacingRead` for user-facing reads?
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing.go` line 36 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdKoGnEpr6ObPDOeaf:-NRdKoGnEpr6ObPDOeag:b-t11bzh) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing.go#L36)):*
> ```Go
> // BlockType values.
> const (
> 	UnknownBlock BlockType = iota
> ```

Ditto here. Do you have any exceptions about how often we will see `UnknownBlock`?
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdXb_u5JvHsvRND9ef:-NRdXb_u5JvHsvRND9eg:b-ydnjn6) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing.go#L49)):*
> ```Go
> 	// Note that recorded events are not necessarily ordered by time - this is
> 	// because separate event "streams" use local buffers (for performance).
> 	UnixNano  int64
> ```

`StartUnixNano` or similar?
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing.go` line 54 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdLQpK1l9u9hxlRwwo:-NRdLQpK1l9u9hxlRwwp:bju3a68) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing.go#L54)):*
> ```Go
> 	BlockType BlockType
> 	// LSM level plus one (with 0 indicating unknown level).
> 	LevelPlusOne uint8
> ```

Why not just `Level`? I figure this is an optimization, tho I wonder if you can motivate the necessity of it in a comment?
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdYbGW9XcCEBRUcy0t:-NRdYbGW9XcCEBRUcy0u:b-gtwti9) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L28)):*
> ```Go
> const Enabled = true
> 
> // Tracer manages the writing of object IO traces to files.
> ```

A comment at top describing the design would be nice!

- A buffer of events per `readable`, `readHandle`, etc. with minimal locking due to most of these structures only allowing a single IO at a time.
- At certain cap, buffer flushes to a pebble-instance-wide buffered channel
- A worker thread reads from channel and writes to a single file
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdUG7y2N027TQ4ZvUf:-NRdUG7zBU69jli809Is:brrpd02) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L47)):*
> ```Go
> 		fsDir:        fsDir,
> 		workerStopCh: make(chan struct{}),
> 		workerDataCh: make(chan eventBuf, 32*1024),
> ```

nit: Make constant.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdZOH5AFOAg8JxePCI:-NRdZOH5AFOAg8JxePCJ:b-f3up2m) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L47)):*
> ```Go
> 		fsDir:        fsDir,
> 		workerStopCh: make(chan struct{}),
> 		workerDataCh: make(chan eventBuf, 32*1024),
> ```

Lots of constants. I wonder if we will want to configure any of these without making a pebble code change. I could see us needing to tweak settings to achieve sufficient perf to write out IO traces on loaded large clusters? Changing pebble then building CRDB with `--cross` and `DEPS.bzl` adjusted feedback loop is pretty slow. I think this is fine to ignore for now. We can always write a PR later that makes more of the settings configurable.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 138 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdOKprElBc1BrCVIwF:-NRdOKprElBc1BrCVIwG:b-w9kaip) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L138)):*
> ```Go
> // ReadAt is part of the objstorage.Readable interface.
> func (r *readable) ReadAt(ctx context.Context, v []byte, off int64) (n int, err error) {
> 	r.mu.Lock()
> ```

- How long-lived is `readable`?
- Do we expect to direct all reads to a single file to a single `readable`?
- Depending on answers to my above Qs, do you worry about contention of calls to `ReadAt` on a single file?
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 139 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdPF9v1NwU80pcM6_c:-NRdPF9v1NwU80pcM6_d:b-h46d3) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L139)):*
> ```Go
> func (r *readable) ReadAt(ctx context.Context, v []byte, off int64) (n int, err error) {
> 	r.mu.Lock()
> 	defer r.mu.Unlock()
> ```

Can we call `Unlock` after calling `add` but before calling the `ReadAt` of the underlying `objstorage.Readable`? I don't immediately see why not.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 182 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdT7hu-QhJN1RwrdhY:-NRdT7hu-QhJN1RwrdhZ:bezgluq) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L182)):*
> ```Go
> 
> // ReadAt is part of the objstorage.ReadHandle interface.
> func (rh *readHandle) ReadAt(ctx context.Context, p []byte, off int64) (n int, err error) {
> ```

So for a `readHandle` we can expect concurrency<=1, even for reads? Maybe add comments whenever can assume that, both in `readHandle` and in `readable`.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 305 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdWkRu3IVHLZG_ZQ44:-NRdWkRu3IVHLZG_ZQ45:b-7xr94q) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L305)):*
> ```Go
> func (g *eventGenerator) flush() {
> 	if g.buf.num > 0 {
> 		g.t.workerDataCh <- g.buf
> ```

Do you have a mental model of the likely contention incurred on this pebble-instance-wide buffered channel & the performance costs of that contention? If no, that is okay. Just curious.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 347 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdTuZBA8yQTaLfZqIj:-NRdTuZBA8yQTaLfZqIk:bv40nsq) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L347)):*
> ```Go
> 			return
> 
> 		case data := <-dataCh:
> ```

Any back of the envelope math about how much IO CRDB would need to do for this to fall behind? I am not sure I have necessary mental models re: perf of sequential writes to as single file / EBS, etc. to answer my own question. Obvi, it is fine to figure this out experimentally also... I am just curious if your mental models are more detailed than mine.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 374 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdSBvU7bmsDStXHksB:-NRdSBvU7bmsDStXHksC:brrpd02) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L374)):*
> ```Go
> 	}
> 	file = vfs.NewSyncingFile(file, vfs.SyncingFileOptions{
> 		BytesPerSync: 128 * 1024,
> ```

nit: Make constant.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2417)* status: 0 of 10 files reviewed, 15 unresolved discussions (waiting on @jbowens and @joshimhoff)

___
*[`table_cache.go` line 418 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRcNSOB9mAPaQksOhtn:-NRdbIZ6908pNGID-GVf:bagaox1) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/table_cache.go#L418)):*

When is this `nil`?
</blockquote></details>

No idea, maybe someone else can chime in. I can confirm it does happen, I had the code outside the if and it paniced in some tests. AFAICT the codepaths that we care about have it set.
___
*[`objstorage/objstorageprovider/provider.go` line 154 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRcNboz23SwFuuboqNi:-NRdbS86Dq8KQx33zzc_:bgply0r) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/provider.go#L154)):*

If `Enabled`, don't we expect `p.tracer` to be not equal to `nil`?
</blockquote></details>

we tolerate multiple Close() calls (other components do and it's easy enough)
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing.go` line 16 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdKNGqClsCK45Kqwo5:-NRdc9Yr935ddnnJIEPZ:b-ldk5sf) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing.go#L16)):*

Can you say more about `RecordCacheHitOp` and `MaxReadaheadOp` in comments? I think I understand this but only bc/ we talked in 1:1.
</blockquote></details>

Added some pointers.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing.go` line 25 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdKW0JAqRQ-nk8NPn3:-NRdcTgSDwPZ2l-OcrX7:bm59adf) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing.go#L25)):*

Do you have any exceptions about how often we will see the `UnknownReason` reason?
</blockquote></details>

All reads other than flushes/compactions/ingestions will be uknown for now. I expect that it's ok to assume all are for user facing operations.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdLAs8Cbn8QFQ72uL0:-NRdca3aADfEoqcthBou:b-2d23oi) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing.go#L28)):*

Don't we want a `Reason` called `FromUserFacingRead` for user-facing reads?
</blockquote></details>

I added a TODO because I think it's hard to plumb this and I think for now it's ok to just assume unknown = user facing.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing.go` line 36 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdKoGnEpr6ObPDOeaf:-NRdiMqKAwUqx-q7a2Qd:bi493hh) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing.go#L36)):*

Ditto here. Do you have any exceptions about how often we will see `UnknownBlock`?
</blockquote></details>

For reads - there are a lot of misc codepaths that are unaccounted for but I expect that we can ignore them and the bulk of the read IOs have the block type set.

For writes - I believe they'll be mostly unknown. We can work on improving this as needed.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdXb_u5JvHsvRND9ef:-NRdcrsp-JoCJC7FRRzK:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing.go#L49)):*

`StartUnixNano` or similar?
</blockquote></details>

Done.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing.go` line 54 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdLQpK1l9u9hxlRwwo:-NRdcsH-A2KoSbATbNAD:bxz82hc) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing.go#L54)):*

Why not just `Level`? I figure this is an optimization, tho I wonder if you can motivate the necessity of it in a comment?
</blockquote></details>

To differentiate between L0 and "unknown" (the comment already says that 0 = unknown).
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdYbGW9XcCEBRUcy0t:-NRde3uz3u5nkmbw1lT7:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L28)):*

A comment at top describing the design would be nice!

- A buffer of events per `readable`, `readHandle`, etc. with minimal locking due to most of these structures only allowing a single IO at a time.
- At certain cap, buffer flushes to a pebble-instance-wide buffered channel
- A worker thread reads from channel and writes to a single file
</blockquote></details>

Done.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdZOH5AFOAg8JxePCI:-NRdeaN325MchLJ4YHHY:bb6sh2i) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L47)):*

Lots of constants. I wonder if we will want to configure any of these without making a pebble code change. I could see us needing to tweak settings to achieve sufficient perf to write out IO traces on loaded large clusters? Changing pebble then building CRDB with `--cross` and `DEPS.bzl` adjusted feedback loop is pretty slow. I think this is fine to ignore for now. We can always write a PR later that makes more of the settings configurable.
</blockquote></details>

Ack.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 138 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdOKprElBc1BrCVIwF:-NRdeuChA1_s1FCsJ0FW:b-etxms3) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L138)):*

- How long-lived is `readable`?
- Do we expect to direct all reads to a single file to a single `readable`?
- Depending on answers to my above Qs, do you worry about contention of calls to `ReadAt` on a single file?
</blockquote></details>

I believe that most data block reads happen through read handles. Direct reads to Readable are mostly for metadata, and for misc operations like Layout() and such.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 139 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdPF9v1NwU80pcM6_c:-NRdfSSzEm1aCZoY-ScR:bwjy83i) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L139)):*

Can we call `Unlock` after calling `add` but before calling the `ReadAt` of the underlying `objstorage.Readable`? I don't immediately see why not.
</blockquote></details>

Nice catch, yes it was silly to hold the lock for the entire thing.
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 182 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdT7hu-QhJN1RwrdhY:-NRdf_Q_EKA6f3yDXRYZ:bws64sl) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L182)):*

So for a `readHandle` we can expect concurrency<=1, even for reads? Maybe add comments whenever can assume that, both in `readHandle` and in `readable`.
</blockquote></details>

These implement the `objstorage.ReadHandle` interface, where it's stated that "Parallel ReadAt calls on the same ReadHandle are not allowed"
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 305 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdWkRu3IVHLZG_ZQ44:-NRdfrXv06BIJlyNkCvX:bvqic1t) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L305)):*

Do you have a mental model of the likely contention incurred on this pebble-instance-wide buffered channel & the performance costs of that contention? If no, that is okay. Just curious.
</blockquote></details>

I expect that as long as the worker can keep up, it won't be a big problem - it's one synchronization per 16 IOs. I'd expect you can do millions of these operations per second against a channel (again, as long as the draining can keep up).
___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 347 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRdTuZBA8yQTaLfZqIj:-NRdgV6iEUFgAAGl0ygQ:b-otn045) ([raw file](https://github.com/cockroachdb/pebble/blob/5f24878fe43d954b149e0fdfd260fce081f6a933/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L347)):*

Any back of the envelope math about how much IO CRDB would need to do for this to fall behind? I am not sure I have necessary mental models re: perf of sequential writes to as single file / EBS, etc. to answer my own question. Obvi, it is fine to figure this out experimentally also... I am just curious if your mental models are more detailed than mine.
</blockquote></details>

Each event is 48 bytes. If we pretend our average IO size is 16KB, we'd be increasing the storage bandwidth by ~0.3%. On the other hand, we trace block cache hits so that can screw up the math.

In any case, ideally we'd want to gather traces from workloads which degrade reasonably even if we slow everything down a bit.

RaduBerinde (commented):
Thanks for the thorough review!

I'll wait for a review on the changes outside objiotracing.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2417#-:-NRdjfIqCeNKJ1t2tHfR:b-wqimid)* status: 0 of 11 files reviewed, 15 unresolved discussions (waiting on @jbowens and @joshimhoff)

sumeerbhola (approved):
:lgtm: for the changes outside objiotracing

> Reviewed 6 of 10 files at r1, 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2417#-:-NRx0aJ0D3YEjHpg31jt:b-aafn3i)* status: 7 of 11 files reviewed, 17 unresolved discussions (waiting on @jbowens, @joshimhoff, and @RaduBerinde)

___
*[`table_cache.go` line 420 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRx-KBOD4gwNxpfC6v-:-NRx-KBOD4gwNxpfC6v0:budgi8a) ([raw file](https://github.com/cockroachdb/pebble/blob/296ff5c82a9b68dbea91487151230aea8ccbd636/table_cache.go#L420)):*
> ```Go
> 	if opts != nil {
> 		useFilter = manifest.LevelToInt(opts.level) != 6 || opts.UseL6Filters
> 		ctx = objiotracing.WithLevel(ctx, manifest.LevelToInt(opts.level))
> ```

should this be gated by `objiotracing.Enabled`? Or are we relying on ob_io_tracing_off.go to make this a noop and zero cost?
___
*[`sstable/reader.go` line 538 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRx0In4952dw8zJ0TT-:-NRx0In4952dw8zJ0TT0:b-q4qw74) ([raw file](https://github.com/cockroachdb/pebble/blob/296ff5c82a9b68dbea91487151230aea8ccbd636/sstable/reader.go#L538)):*
> ```Go
> 		// blockIntersects
> 	}
> 	ctx := objiotracing.WithBlockType(i.ctx, objiotracing.DataBlock)
> ```

same question about these `With*` calls in the sstable package.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2417#-:-NRx7xK49zxp_i2aXRWC:b-814g8k)* status: 7 of 11 files reviewed, 17 unresolved discussions (waiting on @jbowens, @joshimhoff, and @sumeerbhola)

___
*[`table_cache.go` line 420 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2417#-NRx-KBOD4gwNxpfC6v-:-NRx7r1Z9KpL3OC0Mhx1:b-f1p1iv) ([raw file](https://github.com/cockroachdb/pebble/blob/296ff5c82a9b68dbea91487151230aea8ccbd636/table_cache.go#L420)):*

should this be gated by `objiotracing.Enabled`? Or are we relying on ob_io_tracing_off.go to make this a noop and zero cost?
</blockquote></details>

Correct, they are no-ops and should be inlined to nothing.

PR #2418, opened 2023-03-27
db: delay memtable memory allocation until the first write

bananabrick:
Flushable ingestions, and flushable batches can lead to the rotation
of empty memtables. Empty memtables still have memory allocated for
them, and thus can contribute to memtable write stalls. By delaying
memory allocation for memtables until the first write happens, we
can ensure that empty memtables don't contribute to write stalls.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -1702,7 +1702,7 @@ func (b *flushableBatch) inuseBytes() uint64 {
 	return uint64(len(b.data) - batchHeaderLen)
 }
 
-func (b *flushableBatch) totalBytes() uint64 {
+func (b *flushableBatch) allocatedBytes() uint64 {
 	return uint64(cap(b.data))
 }
 
--- a/compaction.go
+++ b/compaction.go
@@ -1718,7 +1718,7 @@ func (d *DB) passedFlushThreshold() bool {
 			// size. See minFlushSize below.
 			size += uint64(d.opts.MemTableSize)
 		} else {
-			size += d.mu.mem.queue[n].totalBytes()
+			size += d.mu.mem.queue[n].allocatedBytes()
 		}
 	}
 	if n == 0 {
--- a/data_test.go
+++ b/data_test.go
@@ -944,6 +944,9 @@ func runDBDefineCmd(td *datadriven.TestData, opts *Options) (*DB, error) {
 					d.updateReadStateLocked(nil)
 				}
 				mem = d.mu.mem.mutable
+				// We're writing using memTable.set, which bypasses
+				// memTable.prepare, so allocate manually.
+				mem.allocate()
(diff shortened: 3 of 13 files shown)


bananabrick:
Still needs some tweaking before review.

bananabrick:
This might not be needed if I'm right here: https://github.com/cockroachdb/pebble/issues/2388#issuecomment-1489414725

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2418)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on @bananabrick and @jbowens)

___
*[`mem_table.go` line 159 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2418#-NRnNX7eC9EAnryiE3E0:-NRnNX7f2CTmEKHUfbim:bvxo1f0) ([raw file](https://github.com/cockroachdb/pebble/blob/a2dbb9d7e4f6c7c9ff5fc6dba0f2c40422bab5ab/mem_table.go#L159)):*
> ```Go
> 		m.releaseCacheReserve = m.cacheReserve(m.capacity)
> 	}
> ```

should we add an invariants-only check that arenaBuf is nil? Or just tolerate multiple calls and add `if m.arenaBuf == nil { return }` (we do this in one call site)
___
*[`mem_table.go` line 172 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2418#-NRnOdmt-UCnlDvs9GJh:-NRnOdmt-UCnlDvs9GJi:ba6uo6p) ([raw file](https://github.com/cockroachdb/pebble/blob/a2dbb9d7e4f6c7c9ff5fc6dba0f2c40422bab5ab/mem_table.go#L172)):*
> ```Go
> 
> func (m *memTable) releaseMemAccounting() (allocatedSize int) {
> 	if m.arenaBuf != nil {
> ```

[nit] it would be a bit more clear if we just did `if m.arenaBuf == nil { return 0 }` in the beginning
___
*[`mem_table.go` line 180 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2418#-NRnOwgY2cuhYkPQDL47:-NRnOwgY2cuhYkPQDL48:b-xcvkr4) ([raw file](https://github.com/cockroachdb/pebble/blob/a2dbb9d7e4f6c7c9ff5fc6dba0f2c40422bab5ab/mem_table.go#L180)):*
> ```Go
> 	if m.arenaBuf != nil && m.allocateManually {
> 		manual.Free(m.arenaBuf)
> 		m.arenaBuf = nil
> ```

Don't we want to set it to `nil` even if allocateManually is false?

bananabrick:
TFTR. But closing this out, because we don't need this.

Issue #1943, opened 2022-09-06
db: batch write stats


sumeerbhola:
For the read path we have detailed stats for an Iterator, that can be used to understand the behavior of an operation that used that iterator. For the write path we don't have stats that can explain the latency incurred by a `Batch.Commit`.

The commit can encounter queueing in various places, like `makeRoomForWrite` and in the `LogWriter` due to queuing inside Pebble, and external queuing in writing and syncing to the WAL file. Some of this can be indirectly inferred by looking at store-level metrics related to write stall duration and the `InternalIntervalMetrics` (https://github.com/cockroachdb/cockroach/issues/85755), but it is not always possible to link exactly what happened to a particular write operation with aggregated metrics. For instance, a particular write may see the effect of multiple write stalls. This may also help in the investigation of https://github.com/cockroachdb/cockroach/issues/72375#issuecomment-1238106419

@nicktrav

sumeerbhola:
remaining work is to add a trace statement in CockroachDB

jbowens:
fyi, corresponding Pebble bump happening in cockroachdb/cockroach#100489.

Issue #2355, opened 2023-02-23
replay: TestCompactionsQuiesce is flaky under race/asan


nicktrav:
Seen [here](https://github.com/cockroachdb/pebble/actions/runs/4249409568/jobs/7389536849).

```console
--- FAIL: TestCompactionsQuiesce (124.35s)
    replay_test.go:515: 
        	Error Trace:	replay_test.go:515
        	            				replay_test.go:519
        	Error:      	Not equal: 
        	            	expected: 0
        	            	actual  : 1
        	Test:       	TestCompactionsQuiesce
FAIL
FAIL	github.com/cockroachdb/pebble/replay	126.535s
```

nicktrav:
A slightly different failure mode. This one was from the linux-race CI job ([link](https://github.com/cockroachdb/pebble/actions/runs/4251128566/jobs/7400462929)).

```console
--- FAIL: TestCompactionsQuiesce (183.80s)
    replay_test.go:510: 
        	Error Trace:	replay_test.go:510
        	            				replay_test.go:519
        	Error:      	Condition never satisfied
        	Test:       	TestCompactionsQuiesce
        	Messages:   	(*replay.Runner).Wait didn't terminate
FAIL
```

RaduBerinde:
This test runs in ~2 seconds locally; these race/asan builds are orders of magnitude slower. I think we need to increase the timeout and the [1 second grace period](https://github.com/cockroachdb/pebble/blob/master/replay/replay.go#L423) for quiescing compactions in these builds.

nicktrav:
Another one: https://github.com/cockroachdb/pebble/actions/runs/6884451979/job/18726978449

PR #2419, opened 2023-03-28
db,record: add BatchCommitStats to measure total and component durati…

sumeerbhola:
…ons for commit

The component durations expose implementation detail and are only for expert diagnosis. We may choose to trace slow batch commits in CockroachDB, and aggregate these durations into histograms for metrics.

Informs #1943

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -13,6 +13,7 @@ import (
 	"sort"
 	"sync"
 	"sync/atomic"
+	"time"
 	"unsafe"
 
 	"github.com/cockroachdb/errors"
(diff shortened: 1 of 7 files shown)


sumeerbhola (commented):
I'd like an initial opinion before I add tests.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2419#-:-NRdXA89EoJIBfpmltbX:bv6bc1)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @jbowens)

jbowens (commented):
took a quick look and the approach looks good to me

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2419#-:-NReIitu-0HRlNomsX1o:bj4pd23)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`batch.go` line 302 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2419#-NReHZjYAv3dCb7Zzhd1:-NReHZjYAv3dCb7Zzhd2:bkd25jl) ([raw file](https://github.com/cockroachdb/pebble/blob/3a69a661350fca92754f18c0dc82261c09ea5ed3/batch.go#L302)):*
> ```Go
> 	// include some duration in which real work was being done for the commit
> 	// and will not be included here. This missing time is considered acceptable
> 	// since the goal of these stats is to understand user-facing latency.
> ```

Is it worth recording _overall_ wall-time duration between when `{Apply,ApplyNoSyncWait}` is called and when `SyncWait` returns? It may encompass time doing other productive work outside the context of committing this batch, but it's representative of the true user-facing latency. Maybe this is already well-accounted above us, so no need to include an accounting here?

jbowens (approved):
:lgtm:

> Reviewed 4 of 6 files at r1, 2 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2419#-:-NRtBgAGD96SJZl188yq:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`db.go` line 2006 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2419#-NRtBMLTAx3iMSEeLZv6:-NRtBMLTAx3iMSEeLZv7:b-r1gldt) ([raw file](https://github.com/cockroachdb/pebble/blob/a10bccc44114818a3fe03b873c25619af397cd14/db.go#L2006)):*
> ```Go
> 	if b != nil && b.ingestedSSTBatch {
> 		panic("pebble: invalid function call")
> 	}
> ```

should we time `makeRoomForWrite` as a whole? during a WAL rotation we might see a latency spike due to syncing the new WAL in place.

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2419#-:-NRtdE9s057tL8vwFk5m:b-814g8k)* status: 5 of 7 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`db.go` line 2006 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2419#-NRtBMLTAx3iMSEeLZv6:-NRtbyUO4zqV3O1L6eJf:bdkgauy) ([raw file](https://github.com/cockroachdb/pebble/blob/a10bccc44114818a3fe03b873c25619af397cd14/db.go#L2006)):*

should we time `makeRoomForWrite` as a whole? during a WAL rotation we might see a latency spike due to syncing the new WAL in place.
</blockquote></details>

Good point. I've added `WALRotationDuration ` to stay consistent with the idea that all the component times are non-overlapping. I didn't cover it in the test since its trivial.

Though I am getting a bit concerned that I went too far with the component durations -- we have 6 of them now. I do think it helps us to know if our queues are too small (SemaphoreWaitDuration and WALQueueWaitDuration) or we are seeing stalls due to memtable or L0 read-amp thresholds, or slowness due to fsyncs. So I don't have good alternative ideas either.

Thoughts?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2419)* status: 5 of 7 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`db.go` line 2006 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2419#-NRtBMLTAx3iMSEeLZv6:-NRtfPxJ8sdLXD0R5KMR:bz1ul3k) ([raw file](https://github.com/cockroachdb/pebble/blob/a10bccc44114818a3fe03b873c25619af397cd14/db.go#L2006)):*

Good point. I've added `WALRotationDuration ` to stay consistent with the idea that all the component times are non-overlapping. I didn't cover it in the test since its trivial.

Though I am getting a bit concerned that I went too far with the component durations -- we have 6 of them now. I do think it helps us to know if our queues are too small (SemaphoreWaitDuration and WALQueueWaitDuration) or we are seeing stalls due to memtable or L0 read-amp thresholds, or slowness due to fsyncs. So I don't have good alternative ideas either.

Thoughts?
</blockquote></details>

Is the concern the overhead? In a stable state, the two stall durations won't be recorded. I think we can also expect the `WALQueueWaitDuration` duration to be rarely recorded too. I say ship it

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2419)* status: 5 of 7 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`db.go` line 2006 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2419#-NRtBMLTAx3iMSEeLZv6:-NRx1OCdCV12gTNSQAZb:b-ysgq1) ([raw file](https://github.com/cockroachdb/pebble/blob/a10bccc44114818a3fe03b873c25619af397cd14/db.go#L2006)):*

Is the concern the overhead? In a stable state, the two stall durations won't be recorded. I think we can also expect the `WALQueueWaitDuration` duration to be rarely recorded too. I say ship it
</blockquote></details>

I agree that of the components only `SemaphoreWaitDuration` and `CommitWaitDuration` will be timed every time and the others should be rarely timed -- the latter being a deliberate choice in the timing code to reduce overhead.

sumeerbhola:
The flake is https://github.com/cockroachdb/pebble/issues/2355

Issue #2421, opened 2023-03-30
db: IngestOperationStats absent for flushable ingests 


jbowens:
We should calculate approximate stats to accommodate admission control.
https://github.com/cockroachdb/pebble/blob/2e9d3f59c4a2a0616108273923187668fb33baab/ingest.go#L661-L663

sumeerbhola:
Glad you noticed this! I think it would be an ok approximation if in `ingestMemtableOverlaps` we computed the full list of files that overlapped with the memtable, and assumed that they will all go into L0 when later doing `runIngestFlush`.

PR #2422, opened 2023-03-30
db: add OverlappedMemtable indicator to IngestOperationStats

jbowens:
Add a new field to IngestOperationStats indicating whether or not any of the ingested sstables overlapped any memtables.

Informs cockroachdb/cockroach#99273.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -662,6 +662,10 @@ type IngestOperationStats struct {
 	// be approximate once https://github.com/cockroachdb/pebble/issues/25 is
 	// implemented.
 	ApproxIngestedIntoL0Bytes uint64
+	// OverlappedMemtable is true if one or more of the memtables contained keys
+	// within the bounds of one or more of the ingested sstables. This triggers
+	// an early flush of the mutable memtable.
+	OverlappedMemtable bool
 }
 
 // IngestWithStats does the same as Ingest, and additionally returns
@@ -785,7 +789,7 @@ func (d *DB) handleIngestAsFlushable(meta []*fileMetadata, seqNum uint64) error 
 
 func (d *DB) ingest(
 	paths []string, targetLevelFunc ingestTargetLevelFunc,
-) (IngestOperationStats, error) {
+) (stats IngestOperationStats, err error) {
 	// Allocate file numbers for all of the files being ingested and mark them as
 	// pending in order to prevent them from being deleted. Note that this causes
 	// the file number ordering to be out of alignment with sequence number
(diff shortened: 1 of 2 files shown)


jbowens (commented):
While here, I noticed #2421. @sumeerbhola — should this be a release blocker due to admission control's dependency on these stats?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2422#-:-NRmn9mZ8hG8VL0jw96G:biyu6k)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @RaduBerinde)

sumeerbhola (commented):
Yes, I think it is a release blocker. I also commented on the issue.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2422#-:-NRn40Bb65btHL8Xa9a8:b-gkzikn)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @RaduBerinde)

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2422)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`ingest.go` line 668 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2422#-NRn6ACqEuWW930H0DCI:-NRnOy3QEkkCiEN4Mgss:b-1x9qdn) ([raw file](https://github.com/cockroachdb/pebble/blob/7b36dc2f709e44ba0e7ccb865a38cd3cb420d8d2/ingest.go#L668)):*

I'm not sure how useful a bool is. Do you have a particular diagnosis approach in mind?
For diagnosis, it may be more useful how many sstable bytes overlapped with the memtable. That may help us figure out whether the overlap is in the CockroachDB local key space or the global key space. The same value may be sufficient for https://github.com/cockroachdb/pebble/issues/2421#issuecomment-1490481249

</blockquote></details>

I was imagining that this bool could be included in the log line summarizing snapshot reception (I think there is such a log line?). I agree this bool alone likely isn't enough to diagnose the memtable overlap.

Are you worried about the cost of calculating the total overlap? Today, we stop as soon as find 1 sstable that overlaps the memtable, and we also don't iterate over any of the overlapping memtable keys once we find 1.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2422)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`ingest.go` line 668 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2422#-NRn6ACqEuWW930H0DCI:-NRtYz7g3ke7m3eRSk6O:b-nvwoy7) ([raw file](https://github.com/cockroachdb/pebble/blob/7b36dc2f709e44ba0e7ccb865a38cd3cb420d8d2/ingest.go#L668)):*
> I agree this bool alone likely isn't enough to diagnose the memtable overlap.

I was hoping we'd do something here that you think (as a first approximation) will be good enough to diagnose.

> Are you worried about the cost of calculating the total overlap? Today, we stop as soon as find 1 sstable that overlaps the memtable

Not really. We are just seeking the memtable iter O(n) times for n files. Even if every ingest finds an overlap it is going to do half those seeks on average, and the good case where there is no overlap will have to do all those seeks. I suspect I have misunderstood your question.

> and we also don't iterate over any of the overlapping memtable keys once we find 1.

We could use the first key that overlaps as a sample key -- the iter is already positioned there.

jbowens:
Superseded by #2425.

PR #1814, opened 2022-07-15
db: add snapshot-pinned keys sstable properties and metrics

jbowens:
Open snapshots require flushes and compactions to preserve, and write,
additional keys. Add new sstable properties and global metrics surfacing
the count and size of these obsolete keys.

These can be used to improve observability and the ability to diagnose
write-amplification incurred due to open snapshots.

Additionally, future work may make use of the table-level properties to
improve compaction heuristics, because files with obsolete keys may be
cheaper to compact than their physical file size alone indicates.

Informs #1204.

PR #2423, opened 2023-03-30
crl-release-23.1: db: compute table properties for snapshot-pinned keys

jbowens:
23.1 backport of #1814.

----

Add three new built-in sstable properties that hold the count, cumulative raw size of keys and cumulative raw size of values that would've been elided were it not for open snapshots. For now these properties only include point keys and not range deletions or range keys.

Also, add two new fields to pebble.Metrics: monotonically increasing cumulative totals of the count and sum of keys written to sstables due to snapshots.


nicktrav (approved):

jbowens:
tftr!

PR #2428, opened 2023-04-03
db: check if the dirname contains ~

itcuihao:
In Unix-like systems, use `~` to display the path of the previous user's home directory.
When creating a dirname with `~`, the dirname creation failure will not be generated.

Example:

```
func Run() {
	db, err := pebble.Open("~/mydata/pebble", &pebble.Options{})
	if err != nil {
		log.Fatal(err)
	}
	key := []byte("hello")
	if err := db.Set(key, []byte("world"), pebble.Sync); err != nil {
		log.Fatal(err)
	}
	value, closer, err := db.Get(key)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("%s %s\n", key, value)
	if err := closer.Close(); err != nil {
		log.Fatal(err)
	}
	if err := db.Close(); err != nil {
		log.Fatal(err)
	}
}
```

This directory `~/mydata/pebble` was not created successfully, on my macos.

So I add the modification.

```
func checkTilde(dirname string) (string, error) {
	if len(dirname) > 0 && dirname[0] == '~' {
		usr, err := user.Current()
		if err != nil {
			return "", err
		}
		return filepath.Join(usr.HomeDir, dirname[1:]), nil
	}
	return dirname, nil
}
```

```
	// Check if the dirname contains ~.
	dirname, err := checkTilde(dirname)
	if err != nil {
		return nil, err
	}
```

### DIFF
--- a/open.go
+++ b/open.go
@@ -11,6 +11,8 @@ import (
 	"io"
 	"math"
 	"os"
+	"os/user"
+	"path/filepath"
 	"sort"
 	"sync/atomic"
 	"time"
@@ -55,6 +57,19 @@ func TableCacheSize(maxOpenFiles int) int {
 	return tableCacheSize
 }
 
+// In Unix-like systems, use "~" to display the path of the previous user's home directory.
+// When creating a dirname with "~", the dirname creation failure will not be generated.
+func checkTilde(dirname string) (string, error) {
+	if len(dirname) > 0 && dirname[0] == '~' {
+		usr, err := user.Current()
+		if err != nil {
+			return "", err
+		}
+		return filepath.Join(usr.HomeDir, dirname[1:]), nil
+	}
+	return dirname, nil
+}
+
 // Open opens a DB whose files live in the given directory.
 func Open(dirname string, opts *Options) (db *DB, _ error) {
 	// Make a copy of the options so that we don't mutate the passed in options.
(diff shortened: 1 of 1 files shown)


jbowens (changes requested):
> In Unix-like systems, use `~` to display the path of the previous user's home directory.

I believe this is a bash-ism; not a Unix-ism. If your application receives a directory path through a command-line argument, it should be expanded before reaching your program. See golang/go#4140 for some related discussion. I don't think this is something we should merge.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2428#-:-NS6r1zZ0aps53UKO3sW:b-t7utxy)* status: 0 of 1 files reviewed, all discussions resolved

itcuihao:
> > In Unix-like systems, use `~` to display the path of the previous user's home directory.
> 
> I believe this is a bash-ism; not a Unix-ism. If your application receives a directory path through a command-line argument, it should be expanded before reaching your program. See [golang/go#4140](https://github.com/golang/go/issues/4140) for some related discussion. I don't think this is something we should merge.
> 
> > _[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2428#-:-NS6r1zZ0aps53UKO3sW:b-t7utxy)_ status: 0 of 1 files reviewed, all discussions resolved

Thank you for reply, I will close it

itcuihao:
> > In Unix-like systems, use `~` to display the path of the previous user's home directory.
> 
> I believe this is a bash-ism; not a Unix-ism. If your application receives a directory path through a command-line argument, it should be expanded before reaching your program. See [golang/go#4140](https://github.com/golang/go/issues/4140) for some related discussion. I don't think this is something we should merge.
> 
> > _[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2428#-:-NS6r1zZ0aps53UKO3sW:b-t7utxy)_ status: 0 of 1 files reviewed, all discussions resolved

Hi~, should pebble limit this kind of behavior? I did not achieve the desired effect when I `Open` dirname on pebble. Whether err should be returned when `dirname[0]` contains `~`.

PR #2432, opened 2023-04-04
sstable: add AtomicFilterMetrics

RaduBerinde:
FilterMetrics is used both to pass around the current values of the metrics and to maintain them internally. In the latter case, the accesses are atomic.

This commit adds `AtomicFilterMetrics` for internal use, which clarifies the code and allows use of `atomic.Int64`.

### DIFF
--- a/sstable/filter.go
+++ b/sstable/filter.go
@@ -7,9 +7,6 @@ package sstable
 import "sync/atomic"
 
 // FilterMetrics holds metrics for the filter policy.
-// TODO(radu): in some contexts, the fields inside are used as atomics; in
-// others they are not (in particular, the struct gets copied around). Split the
-// type into two and use atomic.Int64.
 type FilterMetrics struct {
 	// The number of hits for the filter policy. This is the
 	// number of times the filter policy was successfully used to avoid access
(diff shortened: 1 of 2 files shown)


bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2432)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`sstable/filter.go` line 24 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2432#-NSHgOlhDn-yO_0VVEbn:-NSHgOlhDn-yO_0VVEbo:b-8mzgiq) ([raw file](https://github.com/cockroachdb/pebble/blob/08c14ec5814b8954b8214ce5343aaabb3f2ede0d/sstable/filter.go#L24)):*
> ```Go
> // be updated atomically. An instance of AtomicFilterMetrics can be passed to a
> // Reader as a ReaderOption.
> type AtomicFilterMetrics struct {
> ```

Imo calling this `AtomicFilterMetrics` is confusing. It's not a type which is being updated/read atomically. It's components are individually updated/read atomically. What do you think about choosing an alternative name?

I like the idea of separating the type though and the rest :lgtm:

RaduBerinde:
*[`sstable/filter.go` line 24 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2432#-NSHgOlhDn-yO_0VVEbn:-NSHjZ0fACrsnDWHCHIC:bbvp7vl) ([raw file](https://github.com/cockroachdb/pebble/blob/08c14ec5814b8954b8214ce5343aaabb3f2ede0d/sstable/filter.go#L24)):*

Imo calling this `AtomicFilterMetrics` is confusing. It's not a type which is being updated/read atomically. It's components are individually updated/read atomically. What do you think about choosing an alternative name?

I like the idea of separating the type though and the rest :lgtm: 
</blockquote></details>

Sure, any ideas for names? Maybe `FilterMetricsTracker`? Other ideas are `InternalFilterMetrics`, `SyncedFilterMetrics`

bananabrick (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2432)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`sstable/filter.go` line 24 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2432#-NSHgOlhDn-yO_0VVEbn:-NSHznY-DWapY6o94Hb9:b-p2eznl) ([raw file](https://github.com/cockroachdb/pebble/blob/08c14ec5814b8954b8214ce5343aaabb3f2ede0d/sstable/filter.go#L24)):*
> FilterMetricsTracker
I like `FilterMetricsTracker`.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2432)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @bananabrick)

___
*[`sstable/filter.go` line 24 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2432#-NSHgOlhDn-yO_0VVEbn:-NSIySxmD5N0PyxLvFi7:b1wjbw5) ([raw file](https://github.com/cockroachdb/pebble/blob/08c14ec5814b8954b8214ce5343aaabb3f2ede0d/sstable/filter.go#L24)):*

> FilterMetricsTracker
I like `FilterMetricsTracker`.
</blockquote></details>

Thanks, updated.

PR #2436, opened 2023-04-06
db: expose new metrics for compaction concurrency

### LINKED ISSUE #1934: db: record a metric of effective compaction concurrency
Record an 'effective compaction concurrency' that records the number of compactions running at a time, integrated over an interval. This can be useful for determining whether compaction concurrency is the limiting bottleneck, and when making decisions to lower temporarily increased concurrency values back down (eg, SREOPS-5039).

This new metric can be added to the existing [InternalItervalMetrics](https://pkg.go.dev/github.com/cockroachdb/pebble#DB.InternalIntervalMetrics). It should record the total time spent in compactions during the interval over the interval duration. For example, if the engine consistently has 3 compactions running over the course of the interval, the effective compaction concurrency is 3.

jbowens:
Compute and calculate two new metrics, `Uptime` and `Compact.Duration` which together may be used to compute the effective compaction concurrency over an interval.

**db: add Uptime metric**

Add a new Metric recording the database's uptime since Open. This can be used
in combination with other metrics to integrate over time.

**db: add Compact.Duration metric**

Add a new Compact.Duration metric that measures the cumulative time spent in
compactions since Open. This may be used in combination with Metrics.Uptime to
compute the effective compaction concurrency (https://github.com/cockroachdb/pebble/issues/1934). Callers may compute
effective compaction concurrency over an interval by measuring both Uptime and
Compact.Duration at the beginning and end of the interval and subtracting.

Close #1934.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -555,6 +555,11 @@ type compaction struct {
 	logger    Logger
 	version   *version
 	stats     base.InternalIteratorStats
+	beganAt   time.Time
+	// versionEditApplied is set to true when a compaction has completed and the
+	// resulting version has been installed (if successful), but the compaction
+	// goroutine is still cleaning up (eg, deleting obsolete files).
+	versionEditApplied bool
 
 	score float64
 
@@ -683,7 +688,7 @@ func (c *compaction) makeInfo(jobID int) CompactionInfo {
 	return info
 }
 
-func newCompaction(pc *pickedCompaction, opts *Options) *compaction {
+func newCompaction(pc *pickedCompaction, opts *Options, beganAt time.Time) *compaction {
 	c := &compaction{
 		kind:              compactionKindDefault,
 		cmp:               pc.cmp,
@@ -696,6 +701,7 @@ func newCompaction(pc *pickedCompaction, opts *Options) *compaction {
 		largest:           pc.largest,
 		logger:            opts.Logger,
 		version:           pc.version,
+		beganAt:           beganAt,
 		maxOutputFileSize: pc.maxOutputFileSize,
 		maxOverlapBytes:   pc.maxOverlapBytes,
 		l0SublevelInfo:    pc.l0SublevelInfo,
@@ -727,7 +733,9 @@ func newCompaction(pc *pickedCompaction, opts *Options) *compaction {
 	return c
 }
 
-func newDeleteOnlyCompaction(opts *Options, cur *version, inputs []compactionLevel) *compaction {
+func newDeleteOnlyCompaction(
+	opts *Options, cur *version, inputs []compactionLevel, beganAt time.Time,
+) *compaction {
 	c := &compaction{
 		kind:      compactionKindDeleteOnly,
 		cmp:       opts.Comparer.Compare,
@@ -736,6 +744,7 @@ func newDeleteOnlyCompaction(opts *Options, cur *version, inputs []compactionLev
 		formatKey: opts.Comparer.FormatKey,
 		logger:    opts.Logger,
 		version:   cur,
+		beganAt:   beganAt,
 		inputs:    inputs,
 	}
 
@@ -815,7 +824,9 @@ func adjustGrandparentOverlapBytesForFlush(c *compaction, flushingBytes uint64) 
 	}
 }
 
-func newFlush(opts *Options, cur *version, baseLevel int, flushing flushableList) *compaction {
+func newFlush(
+	opts *Options, cur *version, baseLevel int, flushing flushableList, beganAt time.Time,
+) *compaction {
 	c := &compaction{
 		kind:              compactionKindFlush,
 		cmp:               opts.Comparer.Compare,
@@ -824,6 +835,7 @@ func newFlush(opts *Options, cur *version, baseLevel int, flushing flushableList
 		formatKey:         opts.Comparer.FormatKey,
 		logger:            opts.Logger,
 		version:           cur,
+		beganAt:           beganAt,
 		inputs:            []compactionLevel{{level: -1}, {level: 0}},
 		maxOutputFileSize: math.MaxUint64,
 		maxOverlapBytes:   math.MaxUint64,
@@ -1637,7 +1649,8 @@ func (d *DB) addInProgressCompaction(c *compaction) {
 //
 // DB.mu must be held when calling this method. All writes to the manifest for
 // this compaction should have completed by this point.
-func (d *DB) removeInProgressCompaction(c *compaction, rollback bool) {
+func (d *DB) clearCompactingState(c *compaction, rollback bool) {
+	c.versionEditApplied = true
 	for _, cl := range c.inputs {
 		iter := cl.files.Iter()
 		for f := iter.First(); f != nil; f = iter.Next() {
@@ -1661,8 +1674,6 @@ func (d *DB) removeInProgressCompaction(c *compaction, rollback bool) {
 			f.IsIntraL0Compacting = false
 		}
 	}
-	delete(d.mu.compact.inProgress, c)
-
 	l0InProgress := inProgressL0Compactions(d.getInProgressCompactionInfoLocked(c))
 	d.mu.versions.currentVersion().L0Sublevels.InitCompactingFileInfo(l0InProgress)
 }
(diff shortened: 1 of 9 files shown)


sumeerbhola (changes requested):
> Reviewed 3 of 3 files at r1, 1 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2436)* status: 1 of 8 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`compaction.go` line 1677 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2436#-NSRdRQW4QNOLXxxMZ4o:-NSRdRQW4QNOLXxxMZ4p:bds91le) ([raw file](https://github.com/cockroachdb/pebble/blob/86929113488c4160ade17f9f15dc2b2e8915d367/compaction.go#L1677)):*
> ```Go
> 	// in-progress compactions. This ensures Metrics.Compact.Duration does not
> 	// miss or double count a completing compaction's duration.
> 	d.mu.compact.duration += d.timeNow().Sub(c.beganAt)
> ```

Should we delay this duration computation until where we do `d.mu.compact.compactingCount--`, since we could block in obsolete file deletion after this, and that would prevent starting of another compaction.

Also, isn't this code also executed for flushes?
___
*[`db.go` line 417 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2436#-NSReGurE4CtX7Xld4T5:-NSReGurE4CtX7Xld4T6:b-gqi5sb) ([raw file](https://github.com/cockroachdb/pebble/blob/86929113488c4160ade17f9f15dc2b2e8915d367/db.go#L417)):*
> ```Go
> 
> 			// The cumulative duration of all completed compactions since Open.
> 			// Does not include flushes.
> ```

relates to previous question

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2436)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`compaction.go` line 1677 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2436#-NSRdRQW4QNOLXxxMZ4o:-NT0HC5Q3PmH-bEt6Rhc:b-gklc4i) ([raw file](https://github.com/cockroachdb/pebble/blob/86929113488c4160ade17f9f15dc2b2e8915d367/compaction.go#L1677)):*
> Also, isn't this code also executed for flushes?

Oops, good catch.

> Should we delay this duration computation until where we do d.mu.compact.compactingCount--, since we could block in obsolete file deletion after this, and that would prevent starting of another compaction.

I took this suggestion, but I ran afoul of some undocumented invariant that I don't understand. If `getInProgressCompactionInfoLocked` returns compactions that have successfully completed but are now deleting obsolete files, deadlock is possible. I observed this in unit tests and the metamorphic tests, with a goroutine indefinitely blocked waiting for memtable stall to abate. Any ideas?

sumeerbhola (changes requested):
> Reviewed 2 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2436)* status: 2 of 8 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`compaction.go` line 1677 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2436#-NSRdRQW4QNOLXxxMZ4o:-NTFDvB_-wahl6ZYezLa:b4s7cnx) ([raw file](https://github.com/cockroachdb/pebble/blob/86929113488c4160ade17f9f15dc2b2e8915d367/compaction.go#L1677)):*

> Also, isn't this code also executed for flushes?

Oops, good catch.

> Should we delay this duration computation until where we do d.mu.compact.compactingCount--, since we could block in obsolete file deletion after this, and that would prevent starting of another compaction.

I took this suggestion, but I ran afoul of some undocumented invariant that I don't understand. If `getInProgressCompactionInfoLocked` returns compactions that have successfully completed but are now deleting obsolete files, deadlock is possible. I observed this in unit tests and the metamorphic tests, with a goroutine indefinitely blocked waiting for memtable stall to abate. Any ideas?

</blockquote></details>

IIUC, we are just removing from `d.mu.compact.inProgress` later than before. Is my understanding correct?
It seems unfortunate that we have some undocumented invariant which allows `DB.mu.compactingCount` to be decremented later than removal from `DB.mu.inProgress`.
Odd about the memtable stall -- flush should not be affected.
We may be worse in some performance metrics since we use in-progress compactions in figuring out the ingest target level.
I don't think we have any consistency invariant between the latest `Version` and `inProgress` since we are calling `logAndApply` with the current set of in-progress compactions and removing this compaction after that call. I am wondering why that is even correct (since we use these in-progress compactions to mark L0 files as compacting). Maybe we are just being more pessimistic than we need to be, and it isn't a correctness issue. Ah maybe it doesn't matter since those L0 files that were compacted to Lbase are gone now, so there is nothing there in L0 to mark.

I'll stare some more at the code.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2436)* status: 2 of 8 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`compaction.go` line 1677 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2436#-NSRdRQW4QNOLXxxMZ4o:-NUw6lqP3XHBDcgWMltB:b-t95of5) ([raw file](https://github.com/cockroachdb/pebble/blob/86929113488c4160ade17f9f15dc2b2e8915d367/compaction.go#L1677)):*
I took another look and the indefinite memtable stall seems to be a result of `InitCompactingFileInfo` being invoked with the extra L0 compaction that's actually completed. `InitCompactingFileInfo` marks all the intervals overlapping any L0 base compactions' bounds as base compacting, even if the contained `FileMetadata` are not marked as currently compacting:
```
	// Some intervals may be base compacting without the files contained within
	// those intervals being marked as compacting. This is possible if the files
	// were added after the compaction initiated, and the active compaction
	// files straddle the input file. Mark these intervals as base compacting.
```

> I don't think we have any consistency invariant between the latest `Version` and `inProgress` since we are calling `logAndApply` with the current set of in-progress compactions and removing this compaction after that call.

Deleting obsolete files may drop the database mutex. In the meantime, another compaction/flush may complete and call `logAndApply` with a set of in-progress compactions that includes the completed compaction whose goroutine is deleting obsolete files.

I'm still thinking about what to do about this. It would be nice and easier to reason about if all the post-compaction state transitions were atomic, but I believe we have test dependencies on a compaction deleting the files it made obsolete before being cleared from `inProgress`.

jbowens (commented):
@sumeerbhola — this is good for another look

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2436#-:-NVW12nkDYLiGMZpugRf:bqx5v82)* status: 1 of 9 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 8 files at r3, 5 of 8 files at r4, 2 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2436#-:-NV_yv11BTaR6IMtm3Pe:bg7igkv)* status: 8 of 9 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`compaction.go` line 2596 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2436#-NV_y8SwCMtte5f2T0WI:-NV_y8SwCMtte5f2T0WJ:bpkt8jl) ([raw file](https://github.com/cockroachdb/pebble/blob/06b07306568e3c8c76af01f568836898403e3370/compaction.go#L2596)):*
> ```Go
> 	d.maybeUpdateDeleteCompactionHints(c)
> 	// NB: clearing compacting state must occur before updating the read state;
> 	// L0Sublevels initialization depends on it.
> ```

L0Sublevels for the output version of this compaction has already been initialized in `logAndApply`. But that code filtered out `c` because of `d.getInProgressCompactionInfoLocked(c)` (I had overlooked that filtering in my previous round of comments).

PR #2437, opened 2023-04-06
*: reserve sequence numbers for foreign keys

itsbilal:
This change ratchets the start sequence number for all new keys forward to SeqNumStart (= 10). This reserves 1-9 for special use, of which 1-6 are going to be used for foreign sstable (i.e. shared sstables created by a different node) ingestions in a future change.

Really the only relevant code changes are in open.go and internal/base/seqnums.go. The rest are test changes.

Note that we don't need a Format major version change here; the only guarantee that needs to be held with these sequence numbers is that a foreign sstable ingestion and associated excise cannot be followed with a key being written to with a sequence number below 10. Since that ingestion will only be supported in future releases, as long as this change goes in before foreign sstable ingestions are supported, we're safe.

### DIFF
--- a/compaction_iter.go
+++ b/compaction_iter.go
@@ -957,7 +957,7 @@ func (i *compactionIter) maybeZeroSeqnum(snapshotIdx int) {
 		// This is not the last snapshot
 		return
 	}
-	i.key.SetSeqNum(0)
+	i.key.SetSeqNum(base.SeqNumZero)
 }
 
 // A frontier is used to monitor a compaction's progression across the user
--- a/db_test.go
+++ b/db_test.go
@@ -417,13 +417,13 @@ func TestLargeBatch(t *testing.T) {
 
 	// Verify this results in one L0 table being created.
 	require.NoError(t, try(100*time.Microsecond, 20*time.Second,
-		verifyLSM("0.0:\n  000005:[a#1,SET-a#1,SET]\n")))
+		verifyLSM("0.0:\n  000005:[a#10,SET-a#10,SET]\n")))
 
 	require.NoError(t, d.Set([]byte("b"), bytes.Repeat([]byte("b"), 512), nil))
 
 	// Verify this results in a second L0 table being created.
 	require.NoError(t, try(100*time.Microsecond, 20*time.Second,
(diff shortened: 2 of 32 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2437#-:-NSNPmHhA1DPIGX-kyMh:bnfp4nl)* status: 0 of 32 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`internal/base/seqnums.go` line 15 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2437#-NSNPU7Z8PgHfPgb0kJ8:-NSNPU7Z8PgHfPgb0kJ9:b-p49q3j) ([raw file](https://github.com/cockroachdb/pebble/blob/02fa56927ab57cc0a9646987a75c562105428648/internal/base/seqnums.go#L15)):*
> ```Go
> 	// SeqNumZero is the zero sequence number, set by compactions if they can
> 	// guarantee there are no keys underneath an internal key.
> 	SeqNumZero = 0
> ```

[nit] should we type these to the type used for seq nums (`uint64`)? Just to get some compiler help if we're using them in the wrong context. (Ideally we'd have a specific type)

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2437#-:-NSNWXXR0TsJAnWxA20y:b-814g8k)* status: 0 of 32 files reviewed, all discussions resolved (waiting on @jbowens)

___
*[`internal/base/seqnums.go` line 15 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2437#-NSNPU7Z8PgHfPgb0kJ8:-NSNWVsJC_bm2mpQObhH:b-ydonir) ([raw file](https://github.com/cockroachdb/pebble/blob/02fa56927ab57cc0a9646987a75c562105428648/internal/base/seqnums.go#L15)):*

[nit] should we type these to the type used for seq nums (`uint64`)? Just to get some compiler help if we're using them in the wrong context. (Ideally we'd have a specific type)
</blockquote></details>

Done. Good idea.

Issue #2329, opened 2023-02-15
db: use level iterators to determine overlap with L0 during ingestions


bananabrick:
See https://github.com/cockroachdb/pebble/blob/master/ingest.go#L508. Instead of iterating over every L0 file, we could compute file overlap with each sublevel. In almost all cases, we expect to have fewer sublevels compared to files.

PR #2438, opened 2023-04-10
*: check L0 sublevel metadata when ingesting files

jayshrivastava:
Previously, when determining the level in which to ingest a new file,
we would iterate over every file in L0 to check for overlap. With this change,
we instead iterate over every sublevel to check for overlap. This is more
efficient since the number of sublevels in L0 is <= the number of files in L0.

Closes: https://github.com/cockroachdb/pebble/issues/2329
Release note: None
Epic: None

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -5,7 +5,6 @@
 package pebble
 
 import (
-	"context"
 	"sort"
 	"time"
 
@@ -513,34 +512,30 @@ func ingestTargetLevel(
 
 	targetLevel := 0
 
-	// Do we overlap with keys in L0?
-	// TODO(bananabrick): Use sublevels to compute overlap.
-	iter := v.Levels[0].Iter()
-	for meta0 := iter.First(); meta0 != nil; meta0 = iter.Next() {
-		c1 := sstableKeyCompare(cmp, meta.Smallest, meta0.Largest)
-		c2 := sstableKeyCompare(cmp, meta.Largest, meta0.Smallest)
-		if c1 > 0 || c2 < 0 {
-			continue
-		}
+	// This assertion implicitly checks that we have the current version of
+	// the metadata.
+	if v.L0Sublevels == nil {
+		return 0, errors.AssertionFailedf("could not read L0 sublevels")
+	}
+	// Check for overlap over the keys of L0 by iterating over the sublevels.
+	for subLevel := 0; subLevel < len(v.L0SublevelFiles); subLevel++ {
+		iter := newLevelIter(iterOps, cmp, nil /* split */, newIters,
+			v.L0Sublevels.Levels[subLevel].Iter(), manifest.Level(0), nil)
 
-		// TODO(sumeer): ingest is a user-facing operation, so we should accept a
-		// context and plumb it through, for tracing.
-		iter, rangeDelIter, err := newIters(context.Background(), meta0, nil, internalIterOpts{})
-		if err != nil {
-			return 0, err
-		}
-		rkeyIter, err := newRangeKeyIter(meta0, nil)
-		if err != nil {
-			return 0, err
-		}
-		overlap := overlapWithIterator(iter, &rangeDelIter, rkeyIter, meta, cmp)
-		err = firstError(err, iter.Close())
-		if rangeDelIter != nil {
-			err = firstError(err, rangeDelIter.Close())
-		}
-		if rkeyIter != nil {
-			err = firstError(err, rkeyIter.Close())
-		}
+		var rangeDelIter keyspan.FragmentIterator
+		// Pass in a non-nil pointer to rangeDelIter so that levelIter.findFileGE
+		// sets it up for the target file.
+		iter.initRangeDel(&rangeDelIter)
+
+		levelIter := keyspan.LevelIter{}
+		levelIter.Init(
+			keyspan.SpanIterOptions{}, cmp, newRangeKeyIter,
+			v.L0Sublevels.Levels[subLevel].Iter(), manifest.Level(0), manifest.KeyTypeRange,
+		)
+
+		overlap := overlapWithIterator(iter, &rangeDelIter, &levelIter, meta, cmp)
+		err := iter.Close() // Closes range del iter as well.
+		err = firstError(err, levelIter.Close())
 		if err != nil {
 			return 0, err
 		}
--- a/internal/manifest/version.go
+++ b/internal/manifest/version.go
@@ -1014,7 +1014,7 @@ type Version struct {
 	// usage.
 	//
 	// L0Sublevels.Levels contains L0 files ordered by sublevels. All the files
-	// in Files[0] are in L0Sublevels.Levels. L0SublevelFiles is also set to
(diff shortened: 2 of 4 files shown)


jayshrivastava:
```
name                                                      old time/op  new time/op  delta
ManySSTables/ingest/sstables=10-10                         101µs ± 0%    98µs ± 0%   ~     (p=1.000 n=1+1)
ManySSTables/ingest/sstables=1000-10                       103µs ± 0%   100µs ± 0%   ~     (p=1.000 n=1+1)
ManySSTables/ingest/sstables=10000-10                      174µs ± 0%   152µs ± 0%   ~     (p=1.000 n=1+1)
ManySSTables/ingest/sstables=100000-10                    3.13ms ± 0%  2.01ms ± 0%   ~     (p=1.000 n=1+1)
ManySSTables/ingest/sstables=1000000-10                   44.1ms ± 0%  41.8ms ± 0%   ~     (p=1.000 n=1+1)
ManySSTables/calculateInuseKeyRanges/sstables=10-10       2.11µs ± 0%  2.05µs ± 0%   ~     (p=1.000 n=1+1)
ManySSTables/calculateInuseKeyRanges/sstables=1000-10     2.34µs ± 0%  2.29µs ± 0%   ~     (p=1.000 n=1+1)
ManySSTables/calculateInuseKeyRanges/sstables=10000-10    2.25µs ± 0%  2.24µs ± 0%   ~     (p=1.000 n=1+1)
ManySSTables/calculateInuseKeyRanges/sstables=100000-10   2.31µs ± 0%  2.27µs ± 0%   ~     (p=1.000 n=1+1)
ManySSTables/calculateInuseKeyRanges/sstables=1000000-10  2.71µs ± 0%  2.45µs ± 0%   ~     (p=1.000 n=1+1)
```

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2438#-:-NShlIw18Vju18wSBleD:bnfp4nl)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @jayshrivastava)

___
*[`ingest.go` line 526 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2438#-NShkfcyE6H3mKChYFkB:-NShkfcyE6H3mKChYFkC:bc3gfdi) ([raw file](https://github.com/cockroachdb/pebble/blob/e24d0c28b321c3f2d8abf3a3442dce2f7510d763/ingest.go#L526)):*
> ```Go
> 			iter.initRangeDel(&rangeDelIter)
> 
> 			rkeyIter := &keyspan.LevelIter{}
> ```

[nit] why does it need to be `&` ?

If it escapes, maybe we should move it outside the loop.
___
*[`ingest.go` line 526 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2438#-NShkyN2BoEc3ks03uk4:-NShkyN2BoEc3ks03uk5:b-i7ecwc) ([raw file](https://github.com/cockroachdb/pebble/blob/e24d0c28b321c3f2d8abf3a3442dce2f7510d763/ingest.go#L526)):*
> ```Go
> 			iter.initRangeDel(&rangeDelIter)
> 
> 			rkeyIter := &keyspan.LevelIter{}
> ```

[nit] should the variable be `levelIter` now?

jayshrivastava (commented):
TYFR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2438#-:-NSkefRIAUZhlYjsf9zB:b-80slv9)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @itsbilal)

___
*[`ingest.go` line 526 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2438#-NShkfcyE6H3mKChYFkB:-NSkeQA80PmI1e-jlyPC:bdmaitf) ([raw file](https://github.com/cockroachdb/pebble/blob/e24d0c28b321c3f2d8abf3a3442dce2f7510d763/ingest.go#L526)):*

[nit] why does it need to be `&` ?

If it escapes, maybe we should move it outside the loop.
</blockquote></details>

It escapes. I removed the `&` and verified that it does not escape anymore.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2438)* status: 0 of 4 files reviewed, 1 unresolved discussion

___
*[`ingest.go` line 528 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2438#-NSkfuXz5It29u9dGylc:-NSv82m18jiKFFNx5ZPf:bxxl1ht) ([raw file](https://github.com/cockroachdb/pebble/blob/d65ef526c22455b9874701e5410b07a6fd897357/ingest.go#L528)):*
<details><summary><i>Previously, jayshrivastava (Jayant) wrote…</i></summary><blockquote>

Ty for catching this. I can put up a fix.
</blockquote></details>

@itsbilal — I gave this another look, and I _think_ this is correct and relies on the same (way too subtle) invariants that the non-L0 cases rely on. The call to `overlapWithIterator` seeks the `levelIter` to the smallest bound of the ingested sstable. This should always land on the first file within the [sub]level with _boundary_ overlap. Currently, we rely on the invariant that if there's any _data_ overlap within the level, the first file within the level with _boundary_ overlap must contain data overlap. The comment at the top of the `overlapWithIterator` walks through the cases:

```
func overlapWithIterator(
	iter internalIterator,
	rangeDelIter *keyspan.FragmentIterator,
	rkeyIter keyspan.FragmentIterator,
	meta *fileMetadata,
	cmp Compare,
) bool {
	// Check overlap with point operations.
	//
	// When using levelIter, it seeks to the SST whose boundaries
	// contain meta.Smallest.UserKey(S).
	// It then tries to find a point in that SST that is >= S.
	// If there's no such point it means the SST ends in a tombstone in which case
	// levelIter.SeekGE generates a boundary range del sentinel.
	// The comparison of this boundary with meta.Largest(L) below
	// is subtle but maintains correctness.
	// 1) boundary < L,
	//    since boundary is also > S (initial seek),
	//    whatever the boundary's start key may be, we're always overlapping.
	// 2) boundary > L,
	//    overlap with boundary cannot be determined since we don't know boundary's start key.
	//    We require checking for overlap with rangeDelIter.
	// 3) boundary == L and L is not sentinel,
	//    means boundary < L and hence is similar to 1).
	// 4) boundary == L and L is sentinel,
	//    we'll always overlap since for any values of i,j ranges [i, k) and [j, k) always overlap.
```

While I think this code is correct today, I'm not sure it's guaranteed to remain correct with virtual sstables depending on the implementation.

itsbilal:
@jbowens Yeah that's a very good point. I missed the fact that LevelIter will emit the pausing synthetic boundary keys to stay fixed at the first file, but only when the rangedel iter pointer is set. So yes this isn't a bug. I don't think it'll be a bug with virtual sstables either because the FileMetadatas for VSSTs will still be tight on user keys.

Sorry for the sidetrack @jayshrivastava ! You can probably close #2443.

PR #2440, opened 2023-04-10
objiotracing: set curOffset when doing writes

joshimhoff:
**objiotracing: set curOffset when doing writes**

This commit fixes a buglet that led to IO traces with a zero offset set on all write events.

Release note: None.

```
$ go test . -count=1 -tags=pebble_obj_io_tracing
# github.com/cockroachdb/pebble/objstorage/objstorageprovider/objiotracing.test
ld: warning: -no_pie is deprecated when targeting new OS versions
ld: warning: non-standard -pagezero_size is deprecated when targeting macOS 13.0 or later
ok  	github.com/cockroachdb/pebble/objstorage/objstorageprovider/objiotracing	0.235s
```


joshimhoff:
TTFR!

PR #2443, opened 2023-04-11
*: decouple iterators when determining ingest target level 

jayshrivastava:
Previously, when determining the level at which to ingest an SST,
we would call `overlapWithIterator` with a level iterator
and range deletion interator for each level. The range deletion
iterator was taken from a field of the level iterator. This behavior
is incorrect because the state of the range deletion iterator will depend
on the state of the level iterator, which is is unexpected by
`overlapWithIterator`. This change makes is so that the range deletion
iterator is created independently from the level iterator.

Release note: None
Epic: None

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -5,6 +5,7 @@
 package pebble
 
 import (
+	"context"
 	"sort"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ingest.go
```diff
@@ -522,19 +523,21 @@ func ingestTargetLevel(
...
-		// Pass in a non-nil pointer to rangeDelIter so that levelIter.findFileGE
-		// sets it up for the target file.
-		iter.initRangeDel(&rangeDelIter)
+		rangeDelIter := keyspan.LevelIter{}
+		rangeDelIter.Init(keyspan.SpanIterOptions{RangeKeyFilters: iterOps.RangeKeyFilters}, cmp,
+			tableNewRangeDelIter(context.Background(), newIters), v.L0Sublevels.Levels[subLevel].Iter(),
+			manifest.Level(0), manifest.KeyTypePoint)
+		fragmentRangeDelIter := keyspan.FragmentIterator(&rangeDelIter)
 
 		levelIter := keyspan.LevelIter{}
```
itsbilal, line 532: nit: since there are two leveliters (three if you count the point levelIter), I'd rather call this `rangeKeyIter` to keep it obvious.

ingest.go
```diff
@@ -548,19 +551,22 @@ func ingestTargetLevel(
...
 			v.Levels[level].Iter(), manifest.Level(level), nil)
-		var rangeDelIter keyspan.FragmentIterator
-		// Pass in a non-nil pointer to rangeDelIter so that levelIter.findFileGE
-		// sets it up for the target file.
-		levelIter.initRangeDel(&rangeDelIter)
+
+		rangeDelIter := keyspan.LevelIter{}
+		rangeDelIter.Init(keyspan.SpanIterOptions{RangeKeyFilters: iterOps.RangeKeyFilters}, cmp,
+			tableNewRangeDelIter(context.Background(), newIters), v.Levels[level].Iter(),
+			manifest.Level(0), manifest.KeyTypePoint)
```
itsbilal, line 558: This would be `manifest.Level(level)`

itsbilal (approved):
LGTM save for comments, thanks!

jayshrivastava:
TYFR!

bananabrick (commented):
I think for a fix like this, the unit test should be in the same commit/pr.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2443#-:-NSrDrVF6x0yTJrI6auS:b-kcb7h7)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @jayshrivastava)

jayshrivastava (commented):
I can add one tmr!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2443#-:-NSrh4Qd2kkUxiwaiYa0:b-gvizsj)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @itsbilal)

jayshrivastava:
Closing as this is not required (see discussion in https://github.com/cockroachdb/pebble/pull/2438)

PR #2445, opened 2023-04-12
sstable: include SINGLEDELs within NumDeletions table property

jbowens:
Previously, the sstable property recording the number of deletion tombstones within a file included DELs and RANGEDELs but not SINGLEDELs. This caused SINGLEDELs to be excluded from point tombstone compaction heuristics that use these properties to estimate the amount of disk space that may be reclaimed by compacting a file.

PR #2447, opened 2023-04-12
crl-release-23.1: sstable: include SINGLEDELs within NumDeletions table property

jbowens:
Backport of #2445 for 23.1.x.

----

Previously, the sstable property recording the number of deletion tombstones within a file included DELs and RANGEDELs but not SINGLEDELs. This caused SINGLEDELs to be excluded from point tombstone compaction heuristics that use these properties to estimate the amount of disk space that may be reclaimed by compacting a file.


nicktrav (approved):

jbowens:
TFTR!

Issue #2340, opened 2023-02-16
db: optional point tombstone deletion size estimates


jbowens:
In Cockroach, when we delete a MVCC key with a point tombstone (eg, GC), we always know the size of the live key and value we're deleting. We use these sizes to update MVCC stats.

In Pebble, our point-tombstone compaction heuristics estimate the amount of data dropped by point tombstones by looking at the average logical KV size of either KVs in the same sstable or overlapping sstables in lower levels. In some cases (eg, cockroachlabs/support#2084) this estimate is poor and resulting space amplification can be excessive.

We could extend the `Delete` interface to optionally accept the logical key+value size of the KV pair being deleted and encode it in the value. Pebble could record the sum of these values and the count of these tombstones in new table properties. In compaction heuristics, Pebble could estimate the space reclamation of compacting these point tombstones using these known quantities rather than averages from the own table or lower-levelled tables, significantly improving the estimate.

When a point tombstone causes the elision of another key during a compaction, this size hint could be zeroed, reflecting that it no longer has the expectation of removing lower-level KVs (although it still might, if there were multiple keys with the same user key).

There's still potential for error in compressibility. A user may indicate that a point tombstone deletes a 4KB value, but that value might actually be uniquely compressible and its surrounding data very incompressible (or vice versa).

sumeerbhola:
I really like this idea, since it sidesteps most of the estimation issues.

PR #2448, opened 2023-04-13
db: use exact accounting of point tombstone key sizes

jbowens:
Pebble's compaction heuristics estimate the amount of disk space that will be reclaimed by compacting a file due to point tombstones contained within a file. This estimate is used to increase the priority of compacting these files by inflating the file size during level scoring and file picking.

Previously, these heuristics relied on average key sizes to estimate the space reclaimed by tombstones. Depending on the density of point tombstones, this average was computed over the table's own keys or over all overlapping tables in lower levels.

This commit updates this estimation to instead use a new sstable property that totals all point tombstones' key sizes within a file. This exact sum of the logical bytes that will be reclaimed through compaction, removing all error from uneven distributions of key sizes. This logical estimate must then be scaled to physical sizes to account for compression, metadata, etc. This source of error remains, since different keys may be unequally compressible.

This commit does not alter the logic for estimating the amount of disk space that will be reclaimed through removing deleted entries' values, which continues to rely exclusively on averages. Future work (#2340) will seek to remove the error introduced by value sizes when a user knows the size of the value they're removing.

### DIFF
--- a/sstable/properties.go
+++ b/sstable/properties.go
@@ -136,6 +136,9 @@ type Properties struct {
 	PropertyCollectorNames string `prop:"rocksdb.property.collectors"`
 	// Total raw key size.
 	RawKeySize uint64 `prop:"rocksdb.raw.key.size"`
+	// Total raw key size of point deletion tombstones. This value is comparable
+	// to RawKeySize.
+	RawPointTombstoneKeySize uint64 `prop:"pebble.raw.point-tombstone.key.size"`
 	// Total raw rangekey key size.
 	RawRangeKeyKeySize uint64 `prop:"pebble.raw.range-key.key.size"`
 	// Total raw rangekey value size.
@@ -297,7 +300,7 @@ func (p *Properties) saveString(m map[string][]byte, offset uintptr, value strin
 	m[propOffsetTagMap[offset]] = []byte(value)
 }
 
-func (p *Properties) save(w *rawBlockWriter) {
+func (p *Properties) save(tblFormat TableFormat, w *rawBlockWriter) {
 	m := make(map[string][]byte)
 	for k, v := range p.UserProperties {
 		m[k] = []byte(v)
@@ -347,6 +350,13 @@ func (p *Properties) save(w *rawBlockWriter) {
 	p.saveUvarint(m, unsafe.Offsetof(p.NumDeletions), p.NumDeletions)
 	p.saveUvarint(m, unsafe.Offsetof(p.NumMergeOperands), p.NumMergeOperands)
 	p.saveUvarint(m, unsafe.Offsetof(p.NumRangeDeletions), p.NumRangeDeletions)
+	// NB: We only write out the RawTombstoneKeySize for Pebble formats. This
+	// isn't strictly necessary because unrecognized properties are interpreted
+	// as user-defined properties, however writing them prevents byte-for-byte
+	// equivalence with RocksDB files that some of our testing requires.
+	if p.RawPointTombstoneKeySize > 0 && tblFormat >= TableFormatPebblev1 {
+		p.saveUvarint(m, unsafe.Offsetof(p.RawPointTombstoneKeySize), p.RawPointTombstoneKeySize)
+	}
 	if p.NumRangeKeys() > 0 {
 		p.saveUvarint(m, unsafe.Offsetof(p.NumRangeKeyDels), p.NumRangeKeyDels)
 		p.saveUvarint(m, unsafe.Offsetof(p.NumRangeKeySets), p.NumRangeKeySets)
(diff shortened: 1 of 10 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2448)* status: 7 of 10 files reviewed, 5 unresolved discussions (waiting on @nicktrav and @sumeerbhola)

___
*[`table_stats.go` line 467 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2448#-NTEjIrp6VKB18JZ3d1h:-NTF2e9l4KPws6O5thmR:bb4z4an) ([raw file](https://github.com/cockroachdb/pebble/blob/e17da58b6108318eb2c1cd59cf2e3f156193358f/table_stats.go#L467)):*

Can you change this example to val size since that is what is now being calculated below.

Also, the actual equation in the code is different from the comment and could be more straightforward:
```
compressionRatio := fileSum/uncompressedSum
avgCompressedValueSize := (valSum/entriesCount)*compressionRatio
```
</blockquote></details>

Won't we the precision through truncation when computing the compression ratio?  I think the comment below about rounding/truncation is referring to this.

I updated this and other function to just use floats for more understandable code.
___
*[`table_stats.go` line 651 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2448#-NTEmG4s1p067PxCzt4I:-NTF47At7RPsfhazjMVs:b199fy9) ([raw file](https://github.com/cockroachdb/pebble/blob/e17da58b6108318eb2c1cd59cf2e3f156193358f/table_stats.go#L651)):*

isn't `props.RawKeySize` the total size of keys in the whole sstable and not the mean?
</blockquote></details>

Oof, good catch.
___
*[`table_stats.go` line 684 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2448#-NTEoo6w2_fRYuFWzmHW:-NTFBXLZ4vUrma_hXiOU:bq5mv04) ([raw file](https://github.com/cockroachdb/pebble/blob/e17da58b6108318eb2c1cd59cf2e3f156193358f/table_stats.go#L684)):*

the code comment seems odd -- aren't we truncating with this division?
</blockquote></details>

Yeah, I think the thought was that the error introduced was less than the truncation when computing the compression ratio. Updated this to use floats.
___
*[`sstable/properties.go` line 58 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2448#-NTEfout4WrfuvKX5jvJ:-NTFBouM1OagdAbe2zDk:bftinue) ([raw file](https://github.com/cockroachdb/pebble/blob/e17da58b6108318eb2c1cd59cf2e3f156193358f/sstable/properties.go#L58)):*

unrelated: do you know the memory footprint of a Properties object including the `LoadedMap`? We have been liberally adding fields, and I worried that this is getting bloated and will effect how many files we can cache since it is a field in `Reader`. There seem some fields we never use: ColumnFamily*, ExternalFormatVersion, FixedKeyLen, IndexKeyIsUserKey, IndexValueIsDeltaEncoded, OldestKeyTime, and could just ignore when parsing, and not bother writing them for newer sstable versions.
</blockquote></details>

I'm not sure. I think it is probably worth discarding them to save memory. A couple weeks ago @nicktrav also raised the question of storage cost for these, and whether an alternative encoding/compression might save some disk space. I think the memory savings here are likely to be more impactful, but avoiding writing unused fields will get us that disk space savings for free.
___
*[`sstable/properties.go` line 141 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2448#-NTEet1cAcIG_e4cR3ZR:-NTFD6j011GpTnZdbcB0:b-vqyyph) ([raw file](https://github.com/cockroachdb/pebble/blob/e17da58b6108318eb2c1cd59cf2e3f156193358f/sstable/properties.go#L141)):*

just trying to understand: eventually there will also be a raw value size of value deleted by point deletion tombstone, yes?
But that requires a Pebble interface change, so we are starting with only the key size?
</blockquote></details>

yeah, that's right. I wanted to start with key size so that we have the benefit for keys that we won't know the size of the value of. I've been considering two options for the next property:

1. Add a `RawPointTombstoneValueSize` and use it in combination with `RawPointTombstoneKeySize`. When a tombstone with an encoded value size elides another key, we can subtract the elided value's size from the tombstone's value size to reflect that the 'target' KV pair was already dropped. However, awkwardly, when computing PointDeletionBytesEstimate, the key size estimate would still use `2*RawPointTombstoneKeySize`.
2. Add a `RawPointTombstoneHintSize` property that encodes both key and value size of the _shadowed_ key-value pair only. When the user writes a point tombstone with a value-size hint, the point tombstone's key length and the provided value size are summed and encoded in the value. The `RawPointTombstoneHintSize` property is the sum of all these encoded values. When one of these point tombstones meets the key it deletes, its encoded can be zeroed out to zero. The point tombstone itself can continue to contribute to `RawPointTombstoneKeySize`.

jbowens:
TFTR!

PR #2451, opened 2023-04-14
tool: add db io-bench command

RaduBerinde:
This commit adds an `io-bench` command to the `db` tool. This tool performs IOs against the sstables of an existing database. It is meant to study the IO size vs latency behavior of shared object storage.

Sample output (local SSD):
```
IO count: 10000  Parallelism: 16  Levels: L5,L6
   4KB  --  avg 0.01ms   stddev 0.17ms   p10 0.00ms   p50 0.00ms   p90 0.01ms   p95 0.01ms   p99 0.02ms
  16KB  --  avg 0.01ms   stddev 0.18ms   p10 0.01ms   p50 0.01ms   p90 0.01ms   p95 0.02ms   p99 0.03ms
  64KB  --  avg 0.03ms   stddev 0.15ms   p10 0.02ms   p50 0.03ms   p90 0.04ms   p95 0.05ms   p99 0.11ms
 256KB  --  avg 0.12ms   stddev 0.36ms   p10 0.06ms   p50 0.10ms   p90 0.14ms   p95 0.17ms   p99 0.30ms
1024KB  --  avg 0.43ms   stddev 0.40ms   p10 0.24ms   p50 0.40ms   p90 0.56ms   p95 0.67ms   p99 1.17ms
```

### DIFF
--- a/db.go
+++ b/db.go
@@ -2360,3 +2360,9 @@ func (d *DB) SetCreatorID(creatorID uint64) error {
 	}
 	return d.objProvider.SetCreatorID(objstorage.CreatorID(creatorID))
 }
+
+// ObjProvider returns the objstorage.Provider for this database. Meant to be
+// used for internal purposes only.
+func (d *DB) ObjProvider() objstorage.Provider {
+	return d.objProvider
+}
--- a/tool/db.go
+++ b/tool/db.go
@@ -37,6 +37,7 @@ type dbT struct {
 	Scan       *cobra.Command
 	Set        *cobra.Command
 	Space      *cobra.Command
+	IOBench    *cobra.Command
 
 	// Configuration.
 	opts      *pebble.Options
@@ -44,14 +45,18 @@ type dbT struct {
 	mergers   sstable.Mergers
 
 	// Flags.
-	comparerName string
-	mergerName   string
-	fmtKey       keyFormatter
-	fmtValue     valueFormatter
-	start        key
-	end          key
-	count        int64
-	verbose      bool
+	comparerName  string
+	mergerName    string
+	fmtKey        keyFormatter
+	fmtValue      valueFormatter
+	start         key
+	end           key
+	count         int64
+	allLevels     bool
+	ioCount       int
+	ioParallelism int
+	ioSizes       string
+	verbose       bool
 }
 
 func newDB(opts *pebble.Options, comparers sstable.Comparers, mergers sstable.Mergers) *dbT {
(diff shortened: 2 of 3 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2451#-:-NTEv_rx9ou2Hx0YY_lB:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @joshimhoff and @RaduBerinde)

___
*[`tool/db_io_bench.go` line 78 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2451#-NTErV3J0yG3uKAHw9TS:-NTErV3J0yG3uKAHw9TT:b-jezfz9) ([raw file](https://github.com/cockroachdb/pebble/blob/147c650286e67239c048834c65c929cb0413ffc1/tool/db_io_bench.go#L78)):*
> ```Go
> 	wg.Add(d.ioParallelism)
> 	remainingIOs := ios
> 	for i := 0; i < d.ioParallelism; i++ {
> ```

I assume the intention is to give `len(ios)/d.ioParallelism` to each goroutine, and this code is trying to deal with the fact that `d.ioParallelism` may not be a factor of `len(ios)`. I took me a minute to see how `n := len(remainingIOs) / (d.ioParallelism - i)` accomplishes this. Could you add a code comment to make this very obvious?
___
*[`tool/db_io_bench.go` line 140 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2451#-NTEugtbCuO1QyXpIpNC:-NTEugtbCuO1QyXpIpND:b-o1ceml) ([raw file](https://github.com/cockroachdb/pebble/blob/147c650286e67239c048834c65c929cb0413ffc1/tool/db_io_bench.go#L140)):*
> ```Go
> 			r := readables[idx]
> 			// Random ioSize aligned offset.
> 			ofs := ioSize * rand.Intn(size[idx]*maxIOSize/ioSize)
> ```

I assume the largest `ioSize` should typically be significantly smaller than `maxIOSize` so that these random offsets will rarely read the same part of the file repeatedly?
Should the benchmark sort the intervals generated for these reads and spit out the fraction of intervals that overlapped with other reads, so that if the user runs with a high "io-count" on a small database they realize their results may be effected by caching at the blobstore level.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2451)* status: all files reviewed, 2 unresolved discussions (waiting on @joshimhoff and @sumeerbhola)

___
*[`tool/db_io_bench.go` line 140 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2451#-NTEugtbCuO1QyXpIpNC:-NTF-h6j1JMn_hKsHq9k:bnnz040) ([raw file](https://github.com/cockroachdb/pebble/blob/147c650286e67239c048834c65c929cb0413ffc1/tool/db_io_bench.go#L140)):*

I assume the largest `ioSize` should typically be significantly smaller than `maxIOSize` so that these random offsets will rarely read the same part of the file repeatedly?
Should the benchmark sort the intervals generated for these reads and spit out the fraction of intervals that overlapped with other reads, so that if the user runs with a high "io-count" on a small database they realize their results may be effected by caching at the blobstore level.
</blockquote></details>

Largest ioSize is intended to be the same as maxIOSize, 1MB. `maxIOSize` doesn't do much here, we just round down the size of the sst to `maxIOSize` (which is exactly what we want for ioSize = maxIOSize anyway)

We're just generating random ioSize aligned offsets in the object. I don't think it's that useful to actually calculate overlap, we could print out a warning if the total object sizes is not a good factor larger than 1MB * `io-count`.

PR #2454, opened 2023-04-17
db: introduce interface for sstable readers

bananabrick:
This pr introduces an interface for sstable readers which will be shared by both virtual sstables and physical sstables for code where we don't need special casing for virtual or physical sstables.

Physical readers can still be created and used directly depending on the usecase.

### DIFF
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -1011,7 +1011,7 @@ func TestCompaction(t *testing.T) {
 				if err != nil {
 					return "", "", errors.WithStack(err)
 				}
-				r, err := sstable.NewReader(f, sstable.ReaderOptions{})
+				r, err := sstable.NewPhysicalReader(f, sstable.ReaderOptions{})
 				if err != nil {
 					return "", "", errors.WithStack(err)
 				}
--- a/data_test.go
+++ b/data_test.go
@@ -1104,7 +1104,7 @@ func runSSTablePropertiesCmd(t *testing.T, td *datadriven.TestData, d *DB) strin
 	if err != nil {
 		return err.Error()
 	}
-	r, err := sstable.NewReader(readable, d.opts.MakeReaderOptions())
+	r, err := sstable.NewPhysicalReader(readable, d.opts.MakeReaderOptions())
 	if err != nil {
 		return err.Error()
 	}
--- a/db.go
+++ b/db.go
@@ -1918,7 +1918,7 @@ func (d *DB) EstimateDiskUsage(start, end []byte) (uint64, error) {
 				var size uint64
 				err := d.tableCache.withReader(
 					file,
-					func(r *sstable.Reader) (err error) {
+					func(r sstable.Reader) (err error) {
 						size, err = r.EstimateDiskUsage(start, end)
 						return err
 					},
--- a/external_iterator.go
+++ b/external_iterator.go
@@ -99,7 +99,7 @@ func NewExternalIterWithContext(
 		}
 	}
 
-	var readers [][]*sstable.Reader
+	var readers [][]*sstable.PhysicalReader
 
 	// Ensure we close all the opened readers if we error out.
 	defer func() {
@@ -120,7 +120,7 @@ func NewExternalIterWithContext(
 		seqNumOffset += len(levelFiles)
 	}
 	for _, levelFiles := range files {
-		var subReaders []*sstable.Reader
+		var subReaders []*sstable.PhysicalReader
 		seqNumOffset -= len(levelFiles)
 		subReaders, err = openExternalTables(o, levelFiles, seqNumOffset, o.MakeReaderOptions(), extraReaderOpts...)
 		readers = append(readers, subReaders)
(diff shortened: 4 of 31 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2454)* status: 0 of 31 files reviewed, 1 unresolved discussion (waiting on @bananabrick and @itsbilal)

___
*[`sstable/reader.go` line 2572 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2454#-NTFLZeP5HxLLGyz6M8F:-NTFLZeP5HxLLGyz6M8G:b-i9zxwy) ([raw file](https://github.com/cockroachdb/pebble/blob/624805265e63696db6d8041d120b0be3883b7d0f/sstable/reader.go#L2572)):*
> ```Go
> // Reader is an interface used by both physical and virtual sstable readers.
> type Reader interface {
> 	NewIterWithBlockPropertyFilters(
> ```

[nit] I think all the comments for these methods need to be moved here now.

bananabrick (commented):
On second thought, I'm not certain this interface is needed. We only need to use a Reader for a virtual sstable 1. within the table cache, and 2. When `tableCacheContainer.WithReader` is called. Within the table cache we'll have to check if the sstable is virtual, and then manually construct a virtual sstable reader using the sstable reader cached by the table cache. So, the interface doesn't help with case 1. For case 2, my belief is that it's better to have two functions: 1. `tableCacheContainer.WithReader` and `tableCacheContainer.WithVirtualReader` instead of an interface. This will force the callers of these functions to explicitly handle the virtual sstable case as virtual sstables might need special handling anyway.

What do you think?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2454#-:-NTFW5T08zJs5SmN4Rgk:b-gl7xyl)* status: 0 of 31 files reviewed, all discussions resolved (waiting on @itsbilal)

___
*[`sstable/reader.go` line 2572 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2454#-NTFLZeP5HxLLGyz6M8F:-NTFWq5l42yx6jdnRw7o:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/624805265e63696db6d8041d120b0be3883b7d0f/sstable/reader.go#L2572)):*

[nit] I think all the comments for these methods need to be moved here now.
</blockquote></details>

Done.

bananabrick (commented):
By `Reader` I mean the new `Reader` interface. A virtual sstable reader will be an implementation of this `Reader`.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2454#-:-NTFYGf59ReVAzBkAM48:bmq7lzn)* status: 0 of 31 files reviewed, all discussions resolved (waiting on @itsbilal)

sumeerbhola:
> On second thought, I'm not certain this interface is needed.
...
> By Reader I mean the new Reader interface. A virtual sstable reader will be an implementation of this Reader.

Are you saying the new Reader interface is not needed? What about cases like `singleLevelIterator` that have a reference to a `Reader` that they use for things like reading a block. They can probably make do with a Reader on the backing sst, but it may also be nice to validate that they are not reading data blocks outside the bounds of the virtual sst (though that can be a later extension). I am ok with not having a Reader interface if you don't have a use case for it yet.

itsbilal:
I agree that we likely don't need the interface and can keep `sstable.Reader` referring to readers for ReadableFiles (i.e. physical or physical-ish like with blob storage); bounds enforcement will need to happen inside the iterators anyway as they deal with keys more directly. The table cache `newIters` can produce the appropriately-constrained sstable iterator(s). We could throw an invariant-gated check inside a hypothetical virtual reader that we're not reading a block outside vsst bounds in readers, but I don't see it as necessary for the design. And even then that can happen while keeping the main reader just called `Reader`.

Having a clear delineation that a Reader is just a shared handle to read a file, one that can even be shared across vssts (if they're backed by the same physical sst) is going to be less confusing than a PhysicalReader.

bananabrick (commented):
> Are you saying the new Reader interface is not needed? What about cases like singleLevelIterator that have a reference to a Reader that they use for things like reading a block. They can probably make do with a Reader on the backing sst, but it may also be nice to validate that they are not reading data blocks outside the bounds of the virtual sst (though that can be a later extension). I am ok with not having a Reader interface if you don't have a use case for it yet.

This is handled here https://reviewable.io/reviews/cockroachdb/pebble/2288. The sstable iterators just need some state about the bounds of the virtual sstable to enforce them.

> Having a clear delineation that a Reader is just a shared handle to read a file, one that can even be shared across vssts (if they're backed by the same physical sst) is going to be less confusing than a PhysicalReader.

Yea, agreed. Closing this one out.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2454#-:-NTMdYCZ-Lx0UKFwF0Kp:b-hbwh2b)* status: 0 of 31 files reviewed, all discussions resolved (waiting on @itsbilal and @RaduBerinde)

PR #2357, opened 2023-02-23
*: add skip-shared iteration mode to ScanInternal

itsbilal:
This change adds a skip-shared iteration mode to ScanInternal(), allowing the caller to pass in a Visitor function to expose files in shareable levels (defined as L5 and below) instead of individually visiting all the keys in those files. The file's bounds are truncated to keys within the scan bounds before calling the visitor function. If a file in a shareable level is not marked as shared according to objstorage.Provider, we return a special error, allowing the caller to restart scanning in non-skip-shared mode or so.

PR #2455, opened 2023-04-17
*: Implement iterators and seqnum substitution for foreign SSTs

itsbilal:
When sstables written to by other stores in shared storage are
read, we need to collapse range keys/range dels/point keys within
each SST and then return keys at fixed sequence numbers reserved
for that level. This change implements that path for all sstables
that are in shared storage according to objstorage.Provider
but have a creator ID that doesn't match ours.

First commit is #2357 .


itsbilal:
Not sure what would be the best way to test foreign sstable iterator creation. I will give a shot at a test that forces some SSTs to be foreign by mutating their creator IDs, but the other way to write a test is to just wait for foreign sstable ingestion to be implemented (next PR) and test it more end-to-end.

RaduBerinde:
> Not sure what would be the best way to test foreign sstable iterator creation. I will give a shot at a test that forces some SSTs to be foreign by mutating their creator IDs, but the other way to write a test is to just wait for foreign sstable ingestion to be implemented (next PR) and test it more end-to-end.

We'll want to have the more end-to-end tests anyway, so unless you think we'd be covering something extra, I'd skip the hackier tests.

RaduBerinde (commented):
Some minor comment. I can't help much with most of the code but I'll try to understand as much as I can to educate myself.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2455#-:-NTKRvTiC2gHQNN4irKl:bexgff0)* status: 0 of 19 files reviewed, 2 unresolved discussions (waiting on @bananabrick and @itsbilal)

___
*[`range_keys.go` line 66 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NTKQqGA0idhbSfxPaeQ:-NTKQqGA0idhbSfxPaeR:b-fjfss5) ([raw file](https://github.com/cockroachdb/pebble/blob/a820063d70835375f0e610579e0171a123f959ce/range_keys.go#L66)):*
> ```Go
> 	iter := current.RangeKeyLevels[0].Iter()
> 	for f := iter.Last(); f != nil; f = iter.Prev() {
> 		spanIterOpts := &keyspan.SpanIterOptions{
> ```

[nit] consider making a constructor to which you have to pass the level.
___
*[`objstorage/objstorage.go` line 125 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NTKQ1Te5M14NmhHrDlR:-NTKQ1Te5M14NmhHrDlS:b-k1363g) ([raw file](https://github.com/cockroachdb/pebble/blob/a820063d70835375f0e610579e0171a123f959ce/objstorage/objstorage.go#L125)):*
> ```Go
> // ID that doesn't match ours. The creator ID of the current Provider must
> // be passed in.
> func (meta *ObjectMetadata) IsForeign(creatorID CreatorID) bool {
> ```

[nit] I think this should be a method on the provider (we end up having to do `GetCreatorID()` against the provider anyway).

RaduBerinde (commented):
Do we expect this code to have significant overhead? If yes, then in the progressive restore we may want the "hydration" to happen at a higher level, so that the seqnums are fixed inside the hydrated objects? (that would throw a wrench in the current plans of doing it at the objstorage level)

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2455#-:-NTKTu5iElW2pqHY0T46:bweum8h)* status: 0 of 19 files reviewed, 2 unresolved discussions (waiting on @bananabrick and @itsbilal)

itsbilal (commented):
We won't need to do point collapsing for the progressive restore case, as those sstables will be guaranteed to have one internal key per user key and also their seqnums won't be going into the reserved sequence numbers ranges. That said we would need to apply prefix rewrite rules on them, as discussed elsewhere, but those are going to look very different than anything on here.

TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2455#-:-NTzM-1AEXQYRvc-duPM:b-w7ejcv)* status: 0 of 19 files reviewed, all discussions resolved (waiting on @bananabrick)

___
*[`range_keys.go` line 66 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NTKQqGA0idhbSfxPaeQ:-NTzMLj31g14E0NDeIVK:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/a820063d70835375f0e610579e0171a123f959ce/range_keys.go#L66)):*

[nit] consider making a constructor to which you have to pass the level.
</blockquote></details>

Done.
___
*[`objstorage/objstorage.go` line 125 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NTKQ1Te5M14NmhHrDlR:-NTzMMOg1AtxCrCyiafd:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/a820063d70835375f0e610579e0171a123f959ce/objstorage/objstorage.go#L125)):*

[nit] I think this should be a method on the provider (we end up having to do `GetCreatorID()` against the provider anyway).
</blockquote></details>

Done.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2455#-:-NU3YflDBf7MMmitIivK:b-814g8k)* status: 0 of 19 files reviewed, 1 unresolved discussion (waiting on @bananabrick and @RaduBerinde)

___
*[`scan_internal.go` line 226 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NTz_bI4Eh0XUwfM9l1t:-NU3X9mr0tg2U4DlzLw8:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b7c0c5445b794b38ace6983d2caf4acf5c483f01/scan_internal.go#L226)):*

This could be a method on `pointCollapsingIterator`, no? We have two (almost) copies of it
</blockquote></details>

Done.
___
*[`internal/rangekey/coalesce.go` line 401 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NTzd538ARrq9cER7LH5:-NU3XAhX80EvRBNu4pNV:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b7c0c5445b794b38ace6983d2caf4acf5c483f01/internal/rangekey/coalesce.go#L401)):*

[nit] this code can be shorter if the `switch` just sets a `seqNum` variable and then a single piece of code appends the key.
</blockquote></details>

Done.
___
*[`internal/rangekey/coalesce.go` line 406 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NTze_1F6q8Ng5-qT81D:-NU3XTjl8W4sE3uPvRyk:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b7c0c5445b794b38ace6983d2caf4acf5c483f01/internal/rangekey/coalesce.go#L406)):*

[nit] cleaner to just set it to the right value directly:
```
append(dst.Keys, Key{
  Trailer: base.MakeTrailer(),
  Suffix: keys[i].Suffix,
  Value: keys[i].Value,
})
```

</blockquote></details>

Done.
___
*[`internal/rangekey/coalesce.go` line 416 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NTzdkOJEFxqjcxOyT8j:-NU3WyucCxNN-CIE6V8N:bku55kb) ([raw file](https://github.com/cockroachdb/pebble/blob/b7c0c5445b794b38ace6983d2caf4acf5c483f01/internal/rangekey/coalesce.go#L416)):*

Is this invariant not expected for `InternalKeyKindRangeDelete`? Feels like we're just checking that `KeysBySuffix` does what it says.. We could do that outside the `switch` or in a separate loop on `keys`
</blockquote></details>

It's not because InternalKeyKindRangeDelete doesn't have any prefixes. This logic is the same as `UserIteratorConfig.Transform` higher up in this file btw.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2455)* status: 0 of 19 files reviewed, 1 unresolved discussion (waiting on @bananabrick and @itsbilal)

___
*[`internal/rangekey/coalesce.go` line 416 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NTzdkOJEFxqjcxOyT8j:-NU477lWAE_8-GFwfrk0:b-srb9yq) ([raw file](https://github.com/cockroachdb/pebble/blob/b7c0c5445b794b38ace6983d2caf4acf5c483f01/internal/rangekey/coalesce.go#L416)):*

It's not because InternalKeyKindRangeDelete doesn't have any prefixes. This logic is the same as `UserIteratorConfig.Transform` higher up in this file btw.
</blockquote></details>

Got it, thanks, I see now the logic in `coalesce()` where we add the range-delete key after sorting.

itsbilal:
Rebased after the virtual sstable changed. PTAL!

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2455)* status: 0 of 19 files reviewed, 2 unresolved discussions (waiting on @bananabrick and @itsbilal)

___
*[`scan_internal.go` line 100 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NV9xa2IAIPuAiIZ10IA:-NV9xa2JDkyXA_70e12X:b-5tx8tn) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/scan_internal.go#L100)):*
> ```Go
> 
> // pointCollapsingSSTIterator implements sstable.Iterator while composing
> // pointCollapsingIterator.
> ```

how does the code in this PR handle range keys?

itsbilal (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2455)* status: 0 of 19 files reviewed, 1 unresolved discussion (waiting on @bananabrick and @sumeerbhola)

___
*[`scan_internal.go` line 100 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NV9xa2IAIPuAiIZ10IA:-NVAm5gN7OXJvD0HwvC9:b-l4kdvm) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/scan_internal.go#L100)):*

how does the code in this PR handle range keys?
</blockquote></details>

We don't need `pointCollapsingSSTIterator` for range keys as only point keys flow through it. Range keys are instead passed through a `keyspan.TransformIter` that applies a `ForeignSSTTransformer` on the raw range key iter from the underlying SST. The foreign sst transformer 1) collapses sets/unsets/dels within the SST, similar to `UserIteratorConfig` and 2) exposes the remaining sets at the designated Set sequence number, and the remaining unsets/dels at the appropriate sequence number if L5 (for L6 we skip unsets/dels). Range key masking is handled by the top-level Iterator as always.

You can see the code in `table_cache.go` where we create the transformIter.

jbowens (commented):
> Reviewed 2 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2455)* status: 2 of 19 files reviewed, 9 unresolved discussions (waiting on @bananabrick, @itsbilal, and @sumeerbhola)

___
*[`options.go` line 227 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBZ4ebCkzcaYeO0Mh-:-NVBZ4ebCkzcaYeO0Mh0:br08dhy) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/options.go#L227)):*
> ```Go
> 
> // SpanIterOptions creates a SpanIterOptions from this IterOptions.
> func (o *IterOptions) SpanIterOptions(level manifest.Level) *keyspan.SpanIterOptions {
> ```

Should we pass this around as a struct, instead of a pointer? Otherwise, this allocs per level right?
___
*[`scan_internal.go` line 161 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVB_Gf_DQ17Tl80K-No:-NVB_Gf_DQ17Tl80K-Np:bsx0gx6) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/scan_internal.go#L161)):*
> ```Go
> 	// If fixedSeqNum is non-zero, all emitted points have this fixed sequence
> 	// number.
> 	fixedSeqNum uint64
> ```

when is this zero?
___
*[`scan_internal.go` line 214 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBafZP8iMZn6Xh-lRz:-NVBafZP8iMZn6Xh-lS-:b-ayqhts) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/scan_internal.go#L214)):*
> ```Go
> 
> func (p *pointCollapsingIterator) subSeqNum(key *base.InternalKey) *base.InternalKey {
> 	if p.fixedSeqNum == 0 || key == nil || key.Kind() == InternalKeyKindRangeDelete {
> ```

i'm missing something — how does range delete sequence number substitution work?
___
*[`scan_internal.go` line 241 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBa77o2h35KBjphppx:-NVBa77o2h35KBjphppy:bc795br) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/scan_internal.go#L241)):*
> ```Go
> 	if saveValue {
> 		p.savedValue = val
> 	}
> ```

the `saveValue` parameter is a little confusing; should the callers that need to save the value save it themselves?
___
*[`scan_internal.go` line 356 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBbwIK5ZyvjCJAOHjQ:-NVBbwIK5ZyvjCJAOHjR:b-wqfpz0) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/scan_internal.go#L356)):*
> ```Go
> 				p.iterKey, p.value = p.iter.Next()
> 				p.saveKey()
> 				continue
> ```

can you explain this logic?
___
*[`table_cache.go` line 517 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBdnJTBkkT4K7Shfz6:-NVBdnJTBkkT4K7Shfz7:b-c5o2tr) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/table_cache.go#L517)):*
> ```Go
> 		// the sequence number in this iter, as these range deletes will not be
> 		// exposed to anything other than the interleaving iter and
> 		// pointCollapsingIter.
> ```

how is this used? i'm confused as to how this is used and where keys shadowed by range deletes are dropped. it seems like this could be done using a span mask, but it looks like we pass a nil mask down below.
___
*[`table_cache.go` line 527 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBd-vd2rwjrBDfZgyl:-NVBd-vd2rwjrBDfZgym:b3c2qmm) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/table_cache.go#L527)):*
> ```Go
> 		}
> 		pcIter.iter.Init(dbOpts.opts.Comparer, iter, rangeDelIter, nil /* mask */, opts.LowerBound, opts.UpperBound)
> 		iter = &pointCollapsingSSTIterator{pointCollapsingIterator: pcIter, childIter: iter}
> ```

i think we'll need to pool these to avoid allocations in practice
___
*[`sstable/reader.go` line 3171 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBfT410tuzeoFjWeN3:-NVBfT410tuzeoFjWeN4:b-pt9wvv) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/sstable/reader.go#L3171)):*
> ```Go
> 	}
> 	i := &fragmentBlockIter{}
> 	if err := i.blockIter.initHandle(r.Compare, h, seqNum); err != nil {
> ```

i think this has a problem in that it can surface multiple RANGEDEL fragments with identical sequence numbers (a violation of the contract). we don't have this problem for ingested sstables since they cannot contain overlapping RANGEDEL fragments.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2455#-:-NVeWIh_Dlvdx0aiCOkW:b-814g8k)* status: 1 of 24 files reviewed, 9 unresolved discussions (waiting on @bananabrick, @itsbilal, @jbowens, and @sumeerbhola)

___
*[`options.go` line 227 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBZ4ebCkzcaYeO0Mh-:-NVVou9S1kw6q7Di7bkm:b-4mpj7r) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/options.go#L227)):*

Should we pass this around as a struct, instead of a pointer? Otherwise, this allocs per level right?
</blockquote></details>

Done, though I had to change the TableNewSpanIter contract for this. 
___
*[`scan_internal.go` line 161 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVB_Gf_DQ17Tl80K-No:-NVeRQIK7zhVb-9YykBp:b-8aygmr) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/scan_internal.go#L161)):*

when is this zero?
</blockquote></details>

Shouldn't happen in our uses of it. Should I assert on that?
___
*[`scan_internal.go` line 214 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBafZP8iMZn6Xh-lRz:-NVFWXmU31OCvUZUwZ5O:b8pi9zv) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/scan_internal.go#L214)):*

i'm missing something — how does range delete sequence number substitution work?
</blockquote></details>

We don't expose range deletes through this iterator when in foreign sstable mode. We use the interleaving iterator to apply local deletions but we throw away the rangedel keys (because `elideRangeDels = true`).  Instead they get exposed through `NewFixedSeqnumRangeDelIter`, see the place in newIters in `table_cache.go` where we create that iterator.
___
*[`scan_internal.go` line 241 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBa77o2h35KBjphppx:-NVGQsrZ9hRDtZHa0LfN:bhcrm6l) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/scan_internal.go#L241)):*

the `saveValue` parameter is a little confusing; should the callers that need to save the value save it themselves?
</blockquote></details>

I could make that change, yeah. It's just that it was easiest to follow the convention of `findPrevEntry` setting savedValue (and `findNextEntry` not doing that) and also be able to just do `return p.finishAndReturnMerge()`
___
*[`scan_internal.go` line 356 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBbwIK5ZyvjCJAOHjQ:-NVGRFER1_F95Ygegqvy:bsx9v4s) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/scan_internal.go#L356)):*

can you explain this logic?
</blockquote></details>

Sure. We use the interleavingIter to keep the range del iter in tandem with the point iter. If `p.elideRangeDeletes` is enabled (which it is for all cases except when we use ScanInternal for generating internal-key sstables for replication), we want to ignore the range del points generated by the interleavingIter, but we still want to be able to peek at the `iter.Span()` to confirm if the current key is locally deleted.
___
*[`table_cache.go` line 517 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBdnJTBkkT4K7Shfz6:-NVGRdCDCZ3GRz3QJpdY:bpkaay3) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/table_cache.go#L517)):*

how is this used? i'm confused as to how this is used and where keys shadowed by range deletes are dropped. it seems like this could be done using a span mask, but it looks like we pass a nil mask down below.
</blockquote></details>

My understanding is that using a span mask will also delete points that have a higher sequence number than the range del, as that's how range key masking works. It's why I manually interleave the rangedels and handle deletion in the findNextEntry/findPrevEntry calls to the point collapsing iterator (see the `s.CoversAt` calls there).
___
*[`table_cache.go` line 527 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBd-vd2rwjrBDfZgyl:-NVeRUs12as1CWMwiSsf:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/table_cache.go#L527)):*

i think we'll need to pool these to avoid allocations in practice
</blockquote></details>

Done.
___
*[`sstable/reader.go` line 3171 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBfT410tuzeoFjWeN3:-NVeW8VE79pX7VvyuVDi:b2jx248) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/sstable/reader.go#L3171)):*

i think this has a problem in that it can surface multiple RANGEDEL fragments with identical sequence numbers (a violation of the contract). we don't have this problem for ingested sstables since they cannot contain overlapping RANGEDEL fragments.
</blockquote></details>

This is a good point. Let me address this.

itsbilal (commented):
> Dismissed @jbowens from a discussion.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2455)* status: 1 of 25 files reviewed, 8 unresolved discussions (waiting on @bananabrick, @jbowens, and @sumeerbhola)

___
*[`sstable/reader.go` line 3171 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBfT410tuzeoFjWeN3:-NVfOFp9AkWgZtgF4liz:bl7550) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/sstable/reader.go#L3171)):*

This is a good point. Let me address this.
</blockquote></details>

Done. Let me know if this solution is a little too general.

jbowens (commented):
i'm still working on fully grokking some of the logic within the actual point-collapsing iterator

> Reviewed 1 of 19 files at r1, 1 of 2 files at r3, 6 of 15 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2455#-:-NVfl5QEEWuNIztpxq_N:b-wefw1p)* status: 9 of 25 files reviewed, 10 unresolved discussions (waiting on @bananabrick, @itsbilal, and @sumeerbhola)

___
*[`scan_internal.go` line 161 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVB_Gf_DQ17Tl80K-No:-NVfg4eiACkqgfBGxDtP:b757los) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/scan_internal.go#L161)):*

Shouldn't happen in our uses of it. Should I assert on that?
</blockquote></details>

i think so :+1: 
___
*[`scan_internal.go` line 356 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBbwIK5ZyvjCJAOHjQ:-NVfgX_p1cVqikViYkVz:b-k73z76) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/scan_internal.go#L356)):*

Sure. We use the interleavingIter to keep the range del iter in tandem with the point iter. If `p.elideRangeDeletes` is enabled (which it is for all cases except when we use ScanInternal for generating internal-key sstables for replication), we want to ignore the range del points generated by the interleavingIter, but we still want to be able to peek at the `iter.Span()` to confirm if the current key is locally deleted.
</blockquote></details>

thanks
___
*[`scan_internal.go` line 170 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVfgDrT432gzWzipZRl:-NVfgDrT432gzWzipZRm:bt22rfo) ([raw file](https://github.com/cockroachdb/pebble/blob/9ebe0c19b77826d521be91ac787f06a951ef9c1b/scan_internal.go#L170)):*
> ```Go
> 	elideRangeDeletes bool
> 	// Value at the current iterator position, at iterKey.
> 	value      base.LazyValue
> ```

nit: should we call this `iterValue` to parallel `iterKey` and make it a little more obvious that it's not a iterator-owned copy?
___
*[`scan_internal.go` line 426 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVfhl6pEYmH68N345C_:-NVfhl6pEYmH68N345Ca:b-fpmjh6) ([raw file](https://github.com/cockroachdb/pebble/blob/9ebe0c19b77826d521be91ac787f06a951ef9c1b/scan_internal.go#L426)):*
> ```Go
> 		switch p.iterKey.Kind() {
> 		case InternalKeyKindSet, InternalKeyKindDelete, InternalKeyKindSetWithDelete:
> 			p.saveKey()
> ```

I'm a bit confused about the contract surrounding `saveKey`. If this is the first iteration, we already saved the key before we entered the for loop. Or if we just skipped a key deleted by range deletion, it was just saved by the above for loop. But then here we save again. Down below I see other instances where p.iterKey is updated and we proceed to the next iteration without saving the key. I think it would be helpful to document (and enforce) some invariants, and maybe to split the key fields into `i.iterKey` (always points to the `i.iter`-owned buffer), `i.key` (always backed by `i.savedKeyBuf`) and `i.savedKeyBuf` that's just a buffer used to back `i.key`.
___
*[`scan_internal.go` line 445 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVfkB6QA3q6klyh5m9G:-NVfkB6QA3q6klyh5m9H:b-toj0os) ([raw file](https://github.com/cockroachdb/pebble/blob/9ebe0c19b77826d521be91ac787f06a951ef9c1b/scan_internal.go#L445)):*
> ```Go
> 			// Panic, as this iterator is not expected to observe single deletes.
> 			panic("cannot process singledel key in point collapsing iterator")
> 		case InternalKeyKindMerge:
> ```

at one point we thought we didn't need to support `MERGE`s, right? did we conclude that we _do_? or is it for expediency since timeseries data may be in the same sstable as ordinary mvcc data today?
___
*[`table_cache.go` line 517 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBdnJTBkkT4K7Shfz6:-NVfkXnr3RNKJSlqbAIX:b-yqk3ef) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/table_cache.go#L517)):*

My understanding is that using a span mask will also delete points that have a higher sequence number than the range del, as that's how range key masking works. It's why I manually interleave the rangedels and handle deletion in the findNextEntry/findPrevEntry calls to the point collapsing iterator (see the `s.CoversAt` calls there).
</blockquote></details>

i see. i was thinking a slight change to the span mask interface to surface the whole `InternalKey`, which the span mask could either ignore (range keys) or use to implement elision. but i'm not sure what dragons lie in that approach. and like the comment down below, in the spirit of not entangling too much with existing code, i think we should leave this as-is.
___
*[`sstable/reader.go` line 3171 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBfT410tuzeoFjWeN3:-NVfjY8IBLNP_tr-9vhX:b-nj6upw) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/sstable/reader.go#L3171)):*

Done. Let me know if this solution is a little too general.
</blockquote></details>

the plan is still to use #2465 to avoid the need for point-collapsing in production, right? if so, I think it would be nice to avoid intertwining this implementation too closely with the `fragmentBlockIter`, since we anticipate needing to untangle it.

itsbilal (commented):
TFTR! I'll crank out a cleanup of some of the point-collapsing iter tomorrow, but otherwise this is ready for review.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2455#-:-NW9n2JQAkm_I1-Vn0pI:b-dul0b2)* status: 9 of 26 files reviewed, 9 unresolved discussions (waiting on @bananabrick, @jbowens, and @sumeerbhola)

___
*[`scan_internal.go` line 161 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVB_Gf_DQ17Tl80K-No:-NW9fkTP3btYcRwr4MGo:b31fpeo) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/scan_internal.go#L161)):*

i think so :+1: 
</blockquote></details>

Ah, the `scanInternalIterator` use of `pointCollapsingIterator` does not (and should not) set `fixedSeqNum`.
___
*[`scan_internal.go` line 170 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVfgDrT432gzWzipZRl:-NW9fX-hDnxZP20NWicq:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9ebe0c19b77826d521be91ac787f06a951ef9c1b/scan_internal.go#L170)):*

nit: should we call this `iterValue` to parallel `iterKey` and make it a little more obvious that it's not a iterator-owned copy?
</blockquote></details>

Done.
___
*[`scan_internal.go` line 426 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVfhl6pEYmH68N345C_:-NWA-oFW6U2jdXP6YVNa:bxi3b1p) ([raw file](https://github.com/cockroachdb/pebble/blob/9ebe0c19b77826d521be91ac787f06a951ef9c1b/scan_internal.go#L426)):*

I'm a bit confused about the contract surrounding `saveKey`. If this is the first iteration, we already saved the key before we entered the for loop. Or if we just skipped a key deleted by range deletion, it was just saved by the above for loop. But then here we save again. Down below I see other instances where p.iterKey is updated and we proceed to the next iteration without saving the key. I think it would be helpful to document (and enforce) some invariants, and maybe to split the key fields into `i.iterKey` (always points to the `i.iter`-owned buffer), `i.key` (always backed by `i.savedKeyBuf`) and `i.savedKeyBuf` that's just a buffer used to back `i.key`.
</blockquote></details>

I incorporated your suggestion to have us not set `p.iterKey` to `p.savedKey` and to always have it point to the child iterator's key. It simplified some uses of both variables in this method. I didn't rename `p.savedKey` to `p.key` though, to keep it clearer that there will be some cases where we will continue to pass through `p.iterKey` instead of saving it, as an optimization. I'll give more thought to how I can enforce invariants tomorrow (and if so, which ones) and where I should be adding comments, but currently this works.
___
*[`scan_internal.go` line 445 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVfkB6QA3q6klyh5m9G:-NW9g0Cw4MH8s2QbawbV:bktnmsd) ([raw file](https://github.com/cockroachdb/pebble/blob/9ebe0c19b77826d521be91ac787f06a951ef9c1b/scan_internal.go#L445)):*

at one point we thought we didn't need to support `MERGE`s, right? did we conclude that we _do_? or is it for expediency since timeseries data may be in the same sstable as ordinary mvcc data today?
</blockquote></details>

That's a good point. We handled it just in case we wanted to `ScanInternal` over them, and the code for it was less complex than for SingleDelete. But the reverse case won't be exercised by ScanInternal anyway. I'll rip this out.
___
*[`sstable/reader.go` line 3171 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBfT410tuzeoFjWeN3:-NW9hpn63P0A44I3adJx:b-pi7ocs) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/sstable/reader.go#L3171)):*

the plan is still to use #2465 to avoid the need for point-collapsing in production, right? if so, I think it would be nice to avoid intertwining this implementation too closely with the `fragmentBlockIter`, since we anticipate needing to untangle it.
</blockquote></details>

When we do #2465 we'll still need some special logic to 1) substitute sequence numbers, and 2) exclude the locally-deleted key kinds, right? So I can see more changes along these lines where we just push the logic down to block / sstable iters, instead of doing it higher up in a pointCollapsingIter of some sorts. And this change is looking similar to the 2) case above.

or maybe I'm not understanding your point? Either way I'm happy to pull this out to a separate iterator, it's just that it'll have to be another wrapping iterator as rangedel iterators are otherwise just straight up `fragmentBlockIters` I believe.

jbowens (approved):
:lgtm:

looks good, sorry for the slow review!

> Reviewed 2 of 19 files at r1, 1 of 8 files at r4, 2 of 15 files at r5, 1 of 2 files at r6, 11 of 13 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2455#-:-NWD5IDF2dkGqXFoN9V7:b-ey4gsf)* status: 24 of 26 files reviewed, 3 unresolved discussions (waiting on @bananabrick, @itsbilal, and @sumeerbhola)

___
*[`scan_internal.go` line 426 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVfhl6pEYmH68N345C_:-NWD32Pd32wWKXm-v7LR:buhe890) ([raw file](https://github.com/cockroachdb/pebble/blob/9ebe0c19b77826d521be91ac787f06a951ef9c1b/scan_internal.go#L426)):*

I incorporated your suggestion to have us not set `p.iterKey` to `p.savedKey` and to always have it point to the child iterator's key. It simplified some uses of both variables in this method. I didn't rename `p.savedKey` to `p.key` though, to keep it clearer that there will be some cases where we will continue to pass through `p.iterKey` instead of saving it, as an optimization. I'll give more thought to how I can enforce invariants tomorrow (and if so, which ones) and where I should be adding comments, but currently this works.
</blockquote></details>

I think we still have some duplicate key saves here (eg, `saveKey` before entering the for loop, and then a `saveKey` on this line in the first case block. These are just optimizations though, so maybe just leave a TODO in case this code does end up in use in the release.
___
*[`sstable/reader.go` line 3171 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVBfT410tuzeoFjWeN3:-NWD5JFH8Hi6DswIFMyV:bdaek36) ([raw file](https://github.com/cockroachdb/pebble/blob/8ef56f34a1fb6ad636c9d203d0c038b8bf30782f/sstable/reader.go#L3171)):*

When we do #2465 we'll still need some special logic to 1) substitute sequence numbers, and 2) exclude the locally-deleted key kinds, right? So I can see more changes along these lines where we just push the logic down to block / sstable iters, instead of doing it higher up in a pointCollapsingIter of some sorts. And this change is looking similar to the 2) case above.

or maybe I'm not understanding your point? Either way I'm happy to pull this out to a separate iterator, it's just that it'll have to be another wrapping iterator as rangedel iterators are otherwise just straight up `fragmentBlockIters` I believe.
</blockquote></details>

good point

itsbilal (commented):
thanks a ton and and no worries! apologies for the slow response from me too

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2455#-:-NWDqRCr0c1IpvtGUSqE:bh5z674)* status: 24 of 26 files reviewed, 3 unresolved discussions (waiting on @bananabrick, @jbowens, and @sumeerbhola)

___
*[`scan_internal.go` line 426 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2455#-NVfhl6pEYmH68N345C_:-NWDtfEKE3KuHDwl9j1w:bgh0h20) ([raw file](https://github.com/cockroachdb/pebble/blob/9ebe0c19b77826d521be91ac787f06a951ef9c1b/scan_internal.go#L426)):*

I think we still have some duplicate key saves here (eg, `saveKey` before entering the for loop, and then a `saveKey` on this line in the first case block. These are just optimizations though, so maybe just leave a TODO in case this code does end up in use in the release.
</blockquote></details>

I just removed that duplication. Turns out there's an optimization where we can saveKey by just storing the trailer (as we're already guaranteed to be at the same user key here).

PR #2457, opened 2023-04-17
objstorage: fix ReadAt EOF behavior for shared objects and add test

RaduBerinde:
#### vfs: fix MemFS ReadAt EOF behavior

The MemFS implementation of `ReadAt` does not conform to the spec: an
error must be returned if `n < len(p)`. If we are at the end of the
file, we need to return `io.EOF`.

#### objstorage: fix shared object ReadAt EOF behavior

The current ReadAt implementation for shared objects uses
`io.ReadFull` which converts an `EOF` error into an `UnexpectedEOF`
error. We don't want this conversion so we implement the read loop
ourselves.

We also add a mutex around the internal read handle - unlike
`ReadHandle.ReadAt`, `Readable.ReadAt` must support concurrent calls.

### DIFF
--- a/objstorage/objstorageprovider/provider_test.go
+++ b/objstorage/objstorageprovider/provider_test.go
@@ -7,6 +7,8 @@ package objstorageprovider
 import (
 	"context"
 	"fmt"
+	"io"
+	"math/rand"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


RaduBerinde:
Closing this. Instead, we can remove the EOF behavior altogether, we don't need to ever read past the end of the object (the size is always available).

PR #2458, opened 2023-04-17
objstorage: simplify ReadAt for objects

RaduBerinde:
#### vfs: fix MemFS ReadAt EOF behavior

The MemFS implementation of `ReadAt` does not conform to the spec: an
error must be returned if `n < len(p)`. If we are at the end of the
file, we need to return `io.EOF`.

#### sstable: don't try to read past end of object

Minor change to the `readFooter` code to never try to read past
object EOF.

#### objstorage: simplify ReadAt for objects

Simplify `ReadAt` to only return an error; the `n` return value is no
longer used.

#### objstorage: add mutex in sharedReadable

Unlike `ReadHandle.ReadAt`, `Readable.ReadAt` must support concurrent
calls.

Note that most of the performance sensitive paths use `ReadHandle`.

This fixes an occasional race failure in CI.

### DIFF
--- a/objstorage/noop_readahead.go
+++ b/objstorage/noop_readahead.go
@@ -20,7 +20,7 @@ func MakeNoopReadHandle(r Readable) NoopReadHandle {
 var _ ReadHandle = (*NoopReadHandle)(nil)
 
 // ReadAt is part of the ReadHandle interface.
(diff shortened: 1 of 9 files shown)


itsbilal (approved):
LGTM, nice cleanup!

RaduBerinde:
TFTR!

PR #2460, opened 2023-04-18
db: fix bug with restricted checkpoints

RaduBerinde:
When a checkpoint is restricted to a set of spans, we append a record to the checkpoint's manifest removing all the excluded ssts.

We append the record after copying the raw data from the existing manifest. This is not quite ok: the `record` library works by breaking up records in chunks and packing chunks into 32KB blocks. Chunks cannot straddle a 32KB boundary, but this invariant is violated by our append method. In practice, this only happens if the record we are appending is big and/or we are very unlucky and the existing manifest is close to a 32KB boundary.

To fix this: instead of doing a raw data copy of the existing manifest, we copy at the record level (using a record reader and a record writer). Then we can add a new record using the same writer.

Informs https://github.com/cockroachdb/cockroach/issues/100935

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -371,8 +371,28 @@ func (d *DB) writeCheckpointManifest(
 		}
 		defer dst.Close()
 
-		if _, err := io.Copy(dst, &io.LimitedReader{R: src, N: manifestSize}); err != nil {
-			return err
+		// Copy all existing records. We need to copy at the record level in case we
+		// need to append another record with the excluded files (we cannot simply
+		// append a record after a raw data copy; see
+		// https://github.com/cockroachdb/cockroach/issues/100935).
+		r := record.NewReader(&io.LimitedReader{R: src, N: manifestSize}, manifestFileNum.FileNum())
+		w := record.NewWriter(dst)
+		for {
+			rr, err := r.Next()
+			if err != nil {
+				if err == io.EOF {
+					break
+				}
+				return err
+			}
+
+			rw, err := w.Next()
+			if err != nil {
+				return err
+			}
+			if _, err := io.Copy(rw, rr); err != nil {
+				return err
+			}
 		}
 
 		if len(excludedFiles) > 0 {
@@ -382,17 +402,16 @@ func (d *DB) writeCheckpointManifest(
 				RemovedBackingTables: removeBackingTables,
 			}
 
-			rw := record.NewWriter(dst)
-			w, err := rw.Next()
+			rw, err := w.Next()
 			if err != nil {
 				return err
 			}
-			if err := ve.Encode(w); err != nil {
+			if err := ve.Encode(rw); err != nil {
 				return err
 			}
-			if err := rw.Close(); err != nil {
-				return err
-			}
+		}
+		if err := w.Close(); err != nil {
+			return err
 		}
 		return dst.Sync()
 	}(); err != nil {
--- a/checkpoint_test.go
+++ b/checkpoint_test.go
@@ -7,6 +7,7 @@ package pebble
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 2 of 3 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2460)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`checkpoint_test.go` line 306 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2460#-NTKJn-74YQ4h5V8V9xi:-NTKJn-74YQ4h5V8V9xj:bsrh9j) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb1663a7f2a379b4186dc90292e2c3f0be7cd1d/checkpoint_test.go#L306)):*
> ```Go
> 	defer d.Close()
> 
> 	for i := 0; i < 4000; i++ {
> ```

This test takes about 2 seconds. Does this sound about right or should I spend some time finding where the bottleneck is?  We are creating 4000 sst files but they are tiny and in memory.

Or are there any suggestions on how to speed up this process? I need to have many SST files that I can exclude in the checkpoint.

RaduBerinde:
Note to self to backport to 23.1 branch

bananabrick (commented):
(drive-by)

Should we test checkpoints restricted to spans in the metamorphic test? We have a `metamorphic/ops.go` file where the checkpoint option is run. Maybe we can generate a checkpoint command in the test which also restricts the checkpoint to some spans.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2460#-:-NTQKOQEA5g0BDRm-2fp:bcrwee3)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jbowens)

RaduBerinde:
Good idea, done in https://github.com/cockroachdb/pebble/pull/2466

jbowens (approved):
:lgtm:

Thanks for tracking this down!

> Reviewed 1 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2460#-:-NTQRhc24rSfN-XoCknh:b-c7s5dj)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`checkpoint.go` line 386 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2460#-NTQQ-V0C9XLIBhRHMTv:-NTQQ-V0C9XLIBhRHMTw:b-y2emdx) ([raw file](https://github.com/cockroachdb/pebble/blob/d9fd066bc39f421ba38d5e643281f9fd00c1f5c8/checkpoint.go#L386)):*
> ```Go
> 				if err == io.EOF || record.IsInvalidRecord(err) {
> 					break
> 				}
> ```

We could omit the `record.IsInvalidRecord` here because we got `manifestSize` from the writer, so it must be aligned with the end of a record. If it's not, something is borked.
___
*[`checkpoint_test.go` line 306 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2460#-NTKJn-74YQ4h5V8V9xi:-NTQQgfSBD4CzHBOQB1z:b6gnbeq) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb1663a7f2a379b4186dc90292e2c3f0be7cd1d/checkpoint_test.go#L306)):*

This test takes about 2 seconds. Does this sound about right or should I spend some time finding where the bottleneck is?  We are creating 4000 sst files but they are tiny and in memory.

Or are there any suggestions on how to speed up this process? I need to have many SST files that I can exclude in the checkpoint.
</blockquote></details>

I think 2 seconds is fine. You might be able to speed it up by making your keys longer (eg, add a 100-byte prefix) and then reducing the number of sstables you create. The longer keys need to be encoded in the version edit (twice), so we should hit the the 32 KB offset sooner with longer file boundaries.

We don't care about the sstables themselves being realistic, so maybe you could get a bit of a perf boost by disabling block compression too.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2460#-:-NTQ_lMu-UG9oQ9PGyUa:b-814g8k)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`checkpoint.go` line 386 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2460#-NTQQ-V0C9XLIBhRHMTv:-NTQVVBQ5CKYofACU7p0:bcpd778) ([raw file](https://github.com/cockroachdb/pebble/blob/d9fd066bc39f421ba38d5e643281f9fd00c1f5c8/checkpoint.go#L386)):*

We could omit the `record.IsInvalidRecord` here because we got `manifestSize` from the writer, so it must be aligned with the end of a record. If it's not, something is borked.
</blockquote></details>

Indeed. I was thinking of what happens if the database is read-only (we'd have to be reading the existing manifest), but it seems that checkpoints won't work at all in that case - the writer would not be initialized.
___
*[`checkpoint_test.go` line 306 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2460#-NTKJn-74YQ4h5V8V9xi:-NTQW-pC388F__rJ5pO7:b-c7mj2c) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb1663a7f2a379b4186dc90292e2c3f0be7cd1d/checkpoint_test.go#L306)):*

I think 2 seconds is fine. You might be able to speed it up by making your keys longer (eg, add a 100-byte prefix) and then reducing the number of sstables you create. The longer keys need to be encoded in the version edit (twice), so we should hit the the 32 KB offset sooner with longer file boundaries.

We don't care about the sstables themselves being realistic, so maybe you could get a bit of a perf boost by disabling block compression too.
</blockquote></details>

Interesting ideas, thanks! While trying out the larger keys I realized I didn't actually need 4000, somewhere around 450 was enough to make the manifest just short of 32KB; made the choice randomized around that value so it's still testing something if the encoded sizes change. It's now less than 0.1s.

PR #2464, opened 2023-04-18
objstorage: support LinkOrCopyFromLocal to shared storage

RaduBerinde:
Implement missing functionality in LinkOrCopyFromLocal: we now pass `CreateOptions` and the destination object can be on shared storage.

### DIFF
--- a/objstorage/objstorageprovider/provider.go
+++ b/objstorage/objstorageprovider/provider.go
@@ -6,6 +6,7 @@ package objstorageprovider
 
 import (
 	"context"
+	"io"
 	"os"
 	"sort"
 	"sync"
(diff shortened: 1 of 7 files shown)


itsbilal (approved):
LGTM :shipit:

objstorage/objstorageprovider/provider.go
```diff
@@ -301,8 +308,39 @@ func (p *provider) LinkOrCopyFromLocal(
...
+	w, meta, err := p.Create(ctx, dstFileType, dstFileNum, opts)
+	if err != nil {
+		return objstorage.ObjectMetadata{}, err
+	}
+	f, err := srcFS.Open(srcFilePath, vfs.SequentialReadsOption)
+	if err != nil {
+		return objstorage.ObjectMetadata{}, err
+	}
+	defer f.Close()
+	buf := make([]byte, 64*1024)
```
itsbilal, line 321: Just wondering, any particular reason to do a copy manually manually instead of io.Copy() or io.CopyBuffer() ?

RaduBerinde: We're not an `io.Writer`

PR #2466, opened 2023-04-19
metamorphic: add spans to checkpoint op

RaduBerinde:
This commit adds restriction spans to the checkpoint op in the metamorphic test (with 50% probability).

### DIFF
--- a/metamorphic/generator.go
+++ b/metamorphic/generator.go
@@ -460,7 +460,26 @@ func (g *generator) dbClose() {
 }
 
 func (g *generator) dbCheckpoint() {
-	g.add(&checkpointOp{})
+	// 1/2 of the time we don't restrict the checkpoint;
+	// 1/4 of the time we restrict to 1 span;
+	// 1/8 of the time we restrict to 2 spans; etc.
+	numSpans := 0
+	for g.rng.Intn(2) == 0 {
+		numSpans++
+	}
+	spans := make([]pebble.CheckpointSpan, numSpans)
+	for i := range spans {
+		start := g.randKeyToRead(0.01)
+		end := g.randKeyToRead(0.01)
+		if g.cmp(start, end) > 0 {
+			start, end = end, start
+		}
+		spans[i].Start = start
+		spans[i].End = end
+	}
+	g.add(&checkpointOp{
+		spans: spans,
+	})
 }
 
 func (g *generator) dbCompact() {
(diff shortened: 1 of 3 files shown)


bananabrick (commented):
:lgtm: except the exclusive span end comment.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2466#-:-NTa0Hyr3uUb-Smig279:b-nbpt3z)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`metamorphic/generator.go` line 474 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2466#-NT_yyFK6MOYKJ1YPccZ:-NT_yyFK6MOYKJ1YPcc_:b-x1s2ue) ([raw file](https://github.com/cockroachdb/pebble/blob/a626f9263fb3e97229dcde86f383bc19051d874f/metamorphic/generator.go#L474)):*
> ```Go
> 		start := g.randKeyToRead(0.01)
> 		end := g.randKeyToRead(0.01)
> 		if g.cmp(start, end) > 0 {
> ```

Should we not allow `start == end`? IIRC, the Checkpoint span has an exclusive end.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2466#-:-NTdgCs37DEijNCl7CiO:b-814g8k)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @bananabrick)

___
*[`metamorphic/generator.go` line 474 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2466#-NT_yyFK6MOYKJ1YPccZ:-NTdgD_q5553ouN1LK3K:b-qoq3cp) ([raw file](https://github.com/cockroachdb/pebble/blob/a626f9263fb3e97229dcde86f383bc19051d874f/metamorphic/generator.go#L474)):*

Should we not allow `start == end`? IIRC, the Checkpoint span has an exclusive end.
</blockquote></details>

It shouldn't be a problem - if we have a span like that, it's as if we didn't have that span at all.

bananabrick (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2466#-:-NTesFGL0PGMQ8clvyWg:bnfp4nl)* status: 0 of 3 files reviewed, 1 unresolved discussion

PR #2484, opened 2023-04-26
manifest: support ve encoding/decoding for virtual sstables

bananabrick:
We introduce three new tags for virtual sstable encoding/decoding.

tagCreatedBackingTable, tagRemovedBackingTable are used to encode
VersionEdit.CreatedBackingTables and VersionEdit.RemovedBackingTables.
For CreatedBackingTables, we just need to encode the DiskFileNum and
the Size, and for RemovedBackingTable, we just need to encode the
DiskFileNum.

We introduce a customTag customTagVirtual, to encode whether the
FileMetadata belongs to a virtual sstable. We only use this tag for
virtual sstables. If this tag is absent, then the FileMetadata belongs
to a physical sstable. This tag is also used to encode
FileMetadata.FileBacking.DiskFileNum for virtual sstables.

Note that virtual sstable construction depends on state which may
not exist in the version edit in which the virtual sstable is created.
For this reason, the we need to store some persistent state across calls
to Decode.

### DIFF
--- a/internal/manifest/version_edit.go
+++ b/internal/manifest/version_edit.go
@@ -51,7 +51,9 @@ const (
 	tagMaxColumnFamily  = 203
 
 	// Pebble tags.
-	tagNewFile5 = 104 // Range keys.
+	tagNewFile5            = 104 // Range keys.
+	tagCreatedBackingTable = 105
+	tagRemovedBackingTable = 106
 
 	// The custom tags sub-format used by tagNewFile4 and above.
 	customTagTerminate         = 1
@@ -59,6 +61,7 @@ const (
 	customTagCreationTime      = 6
 	customTagPathID            = 65
 	customTagNonSafeIgnoreMask = 1 << 6
+	customTagVirtual           = 66
 )
 
 // DeletedFileEntry holds the state for a file deletion from a level. The file
@@ -73,6 +76,9 @@ type DeletedFileEntry struct {
 type NewFileEntry struct {
 	Level int
 	Meta  *FileMetadata
+	// BackingFileNum is only set during manifest replay, and only for virtual
+	// sstables.
+	BackingFileNum base.DiskFileNum
 }
 
 // VersionEdit holds the state for an edit to a Version along with other
@@ -143,7 +149,9 @@ type VersionEdit struct {
 
 // Decode decodes an edit from the specified reader.
 //
-// TODO(bananabrick): Support decoding of virtual sstable state.
+// Note that the Decode step will not set the FileBacking for virtual sstables
+// and the responsibility is left to the caller. However, the Decode step will
+// populate the NewFileEntry.BackingFileNum in VersionEdit.NewFiles.
 func (v *VersionEdit) Decode(r io.Reader) error {
 	br, ok := r.(byteReader)
 	if !ok {
@@ -196,6 +204,28 @@ func (v *VersionEdit) Decode(r io.Reader) error {
 			}
 			// NB: RocksDB does not use compaction pointers anymore.
 
+		case tagRemovedBackingTable:
+			n, err := d.readUvarint()
+			if err != nil {
+				return err
+			}
+			v.RemovedBackingTables = append(
+				v.RemovedBackingTables, base.FileNum(n).DiskFileNum(),
+			)
+		case tagCreatedBackingTable:
+			dfn, err := d.readUvarint()
+			if err != nil {
+				return err
+			}
+			size, err := d.readUvarint()
+			if err != nil {
+				return err
+			}
+			fileBacking := &FileBacking{
+				DiskFileNum: base.FileNum(dfn).DiskFileNum(),
+				Size:        size,
+			}
+			v.CreatedBackingTables = append(v.CreatedBackingTables, fileBacking)
 		case tagDeletedFile:
 			level, err := d.readLevel()
 			if err != nil {
@@ -301,6 +331,10 @@ func (v *VersionEdit) Decode(r io.Reader) error {
 			}
 			var markedForCompaction bool
 			var creationTime uint64
+			virtualState := struct {
+				virtual        bool
+				backingFileNum uint64
+			}{}
 			if tag == tagNewFile4 || tag == tagNewFile5 {
 				for {
 					customTag, err := d.readUvarint()
@@ -309,7 +343,16 @@ func (v *VersionEdit) Decode(r io.Reader) error {
 					}
 					if customTag == customTagTerminate {
 						break
+					} else if customTag == customTagVirtual {
+						virtualState.virtual = true
+						n, err := d.readUvarint()
+						if err != nil {
+							return err
+						}
+						virtualState.backingFileNum = n
+						continue
 					}
+
 					field, err := d.readBytes()
 					if err != nil {
 						return err
@@ -345,6 +388,7 @@ func (v *VersionEdit) Decode(r io.Reader) error {
 				SmallestSeqNum:      smallestSeqNum,
 				LargestSeqNum:       largestSeqNum,
 				MarkedForCompaction: markedForCompaction,
+				Virtual:             virtualState.virtual,
 			}
 			if tag != tagNewFile5 { // no range keys present
 				m.SmallestPointKey = base.DecodeInternalKey(smallestPointKey)
@@ -376,11 +420,18 @@ func (v *VersionEdit) Decode(r io.Reader) error {
 				}
 			}
 			m.boundsSet = true
-			m.InitPhysicalBacking()
-			v.NewFiles = append(v.NewFiles, NewFileEntry{
+			if !virtualState.virtual {
+				m.InitPhysicalBacking()
+			}
+
+			nfe := NewFileEntry{
 				Level: level,
 				Meta:  m,
-			})
+			}
+			if virtualState.virtual {
+				nfe.BackingFileNum = base.FileNum(virtualState.backingFileNum).DiskFileNum()
+			}
+			v.NewFiles = append(v.NewFiles, nfe)
 
 		case tagPrevLogNumber:
 			n, err := d.readUvarint()
@@ -442,8 +493,6 @@ func (v *VersionEdit) String() string {
 }
 
 // Encode encodes an edit to the specified writer.
-//
-// TODO(bananabrick): Support encoding of virtual sstable state.
 func (v *VersionEdit) Encode(w io.Writer) error {
 	e := versionEditEncoder{new(bytes.Buffer)}
 
@@ -463,6 +512,15 @@ func (v *VersionEdit) Encode(w io.Writer) error {
 		e.writeUvarint(tagNextFileNumber)
 		e.writeUvarint(uint64(v.NextFileNum))
 	}
+	for _, dfn := range v.RemovedBackingTables {
+		e.writeUvarint(tagRemovedBackingTable)
+		e.writeUvarint(uint64(dfn.FileNum()))
+	}
+	for _, fileBacking := range v.CreatedBackingTables {
+		e.writeUvarint(tagCreatedBackingTable)
+		e.writeUvarint(uint64(fileBacking.DiskFileNum.FileNum()))
+		e.writeUvarint(fileBacking.Size)
+	}
 	// RocksDB requires LastSeqNum to be encoded for the first MANIFEST entry,
 	// even though its value is zero. We detect this by encoding LastSeqNum when
 	// ComparerName is set.
@@ -476,7 +534,7 @@ func (v *VersionEdit) Encode(w io.Writer) error {
 		e.writeUvarint(uint64(x.FileNum))
 	}
 	for _, x := range v.NewFiles {
-		customFields := x.Meta.MarkedForCompaction || x.Meta.CreationTime != 0
+		customFields := x.Meta.MarkedForCompaction || x.Meta.CreationTime != 0 || x.Meta.Virtual
 		var tag uint64
 		switch {
 		case x.Meta.HasRangeKeys:
@@ -529,6 +587,10 @@ func (v *VersionEdit) Encode(w io.Writer) error {
 				e.writeUvarint(customTagNeedsCompaction)
 				e.writeBytes([]byte{1})
 			}
+			if x.Meta.Virtual {
+				e.writeUvarint(customTagVirtual)
+				e.writeUvarint(uint64(x.Meta.FileBacking.DiskFileNum.FileNum()))
+			}
 			e.writeUvarint(customTagTerminate)
 		}
 	}
@@ -536,6 +598,7 @@ func (v *VersionEdit) Encode(w io.Writer) error {
 	return err
 }
 
+// versionEditDecoder should be used to decode version edits.
 type versionEditDecoder struct {
 	byteReader
 }
@@ -634,7 +697,9 @@ type BulkVersionEdit struct {
 	Added   [NumLevels]map[base.FileNum]*FileMetadata
 	Deleted [NumLevels]map[base.FileNum]*FileMetadata
 
-	AddedFileBacking   []*FileBacking
+	// AddedFileBacking is a map to support lookup so that we can populate the
+	// FileBacking of virtual sstables during manifest replay.
+	AddedFileBacking   map[base.DiskFileNum]*FileBacking
 	RemovedFileBacking []base.DiskFileNum
 
 	// AddedByFileNum maps file number to file metadata for all added files
@@ -697,6 +762,16 @@ func (b *BulkVersionEdit) Accumulate(ve *VersionEdit) error {
 		}
 	}
 
+	// Generate state for Added backing files. Note that these must be generated
+	// before we loop through the NewFiles, because we need to populate the
+	// FileBackings which might be used by the NewFiles loop.
+	if b.AddedFileBacking == nil {
+		b.AddedFileBacking = make(map[base.DiskFileNum]*FileBacking)
+	}
+	for _, fb := range ve.CreatedBackingTables {
+		b.AddedFileBacking[fb.DiskFileNum] = fb
+	}
+
 	for _, nf := range ve.NewFiles {
 		// A new file should not have been deleted in this or a preceding
 		// VersionEdit at the same level (though files can move across levels).
@@ -705,6 +780,17 @@ func (b *BulkVersionEdit) Accumulate(ve *VersionEdit) error {
 				return base.CorruptionErrorf("pebble: file deleted L%d.%s before it was inserted", nf.Level, nf.Meta.FileNum)
 			}
 		}
+		if nf.Meta.Virtual && nf.Meta.FileBacking == nil {
+			// FileBacking for a virtual sstable must only be nil if we're performing
+			// manifest replay.
+			nf.Meta.FileBacking = b.AddedFileBacking[nf.BackingFileNum]
+			if nf.Meta.FileBacking == nil {
+				return errors.Errorf("FileBacking for virtual sstable must not be nil")
+			}
+		} else if nf.Meta.FileBacking == nil {
+			return errors.Errorf("Added file L%d.%s's has no FileBacking", nf.Level, nf.Meta.FileNum)
+		}
+
 		if b.Added[nf.Level] == nil {
 			b.Added[nf.Level] = make(map[base.FileNum]*FileMetadata)
 		}
@@ -716,9 +802,6 @@ func (b *BulkVersionEdit) Accumulate(ve *VersionEdit) error {
 			b.MarkedForCompactionCountDiff++
 		}
 	}
-
-	// Generate state for the backing files.
-	b.AddedFileBacking = append(b.AddedFileBacking, ve.CreatedBackingTables...)
 
 	// Since a file can be removed from backing files in exactly one version
 	// edit it is safe to just append without any de-duplication.
--- a/internal/manifest/version_edit_test.go
+++ b/internal/manifest/version_edit_test.go
@@ -38,6 +38,77 @@ func checkRoundTrip(e0 VersionEdit) error {
 	return nil
 }
 
+// Version edits with virtual sstables will not be the same after a round trip
+// as the Decode function will not set the FileBacking for a virtual sstable.
+// We test round trip + bve accumulation here, after which the virtual sstable
+// FileBacking should be set.
+func TestVERoundTripAndAccumulate(t *testing.T) {
+	cmp := base.DefaultComparer.Compare
+	m1 := (&FileMetadata{
+		FileNum:        810,
+		Size:           8090,
+		CreationTime:   809060,
+		SmallestSeqNum: 9,
+		LargestSeqNum:  11,
+	}).ExtendPointKeyBounds(
+		cmp,
+		base.MakeInternalKey([]byte("a"), 0, base.InternalKeyKindSet),
+		base.MakeInternalKey([]byte("m"), 0, base.InternalKeyKindSet),
+	).ExtendRangeKeyBounds(
+		cmp,
+		base.MakeInternalKey([]byte("l"), 0, base.InternalKeyKindRangeKeySet),
+		base.MakeExclusiveSentinelKey(base.InternalKeyKindRangeKeySet, []byte("z")),
+	)
+	m1.InitPhysicalBacking()
+
+	m2 := (&FileMetadata{
+		FileNum:        812,
+		Size:           8090,
+		CreationTime:   809060,
+		SmallestSeqNum: 9,
+		LargestSeqNum:  11,
+		Virtual:        true,
+		FileBacking:    m1.FileBacking,
+	}).ExtendPointKeyBounds(
+		cmp,
+		base.MakeInternalKey([]byte("a"), 0, base.InternalKeyKindSet),
+		base.MakeInternalKey([]byte("c"), 0, base.InternalKeyKindSet),
+	)
+
+	ve1 := VersionEdit{
+		ComparerName:         "11",
+		MinUnflushedLogNum:   22,
+		ObsoletePrevLogNum:   33,
+		NextFileNum:          44,
+		LastSeqNum:           55,
+		CreatedBackingTables: []*FileBacking{m1.FileBacking},
+		NewFiles: []NewFileEntry{
+			{
+				Level: 4,
+				Meta:  m2,
+				// Only set for the test.
+				BackingFileNum: m2.FileBacking.DiskFileNum,
+			},
+		},
+	}
+	var err error
+	buf := new(bytes.Buffer)
+	if err = ve1.Encode(buf); err != nil {
+		t.Error(err)
+	}
+	var ve2 VersionEdit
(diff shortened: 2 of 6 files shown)


bananabrick (commented):
The commit message explains why I needed to make the changes to the decoder/Decode.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484#-:-NU-wNRv5rx6tSsyJl4_:b15tpc0)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484)* status: 0 of 10 files reviewed, 3 unresolved discussions (waiting on @bananabrick)

___
*[`version_set.go` line 237 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NU4VYJM6zF97ABBTKp1:-NU4VYJM6zF97ABBTKp2:b-bhm0zt) ([raw file](https://github.com/cockroachdb/pebble/blob/d9f53d3c4da880412d5be0660f4024d9537895e7/version_set.go#L237)):*
> ```Go
> 		}
> 		var ve *versionEdit
> 		ve, err = ved.Decode(r)
> ```

It feels fragile to also store state in the decoder. Could we instead pass the current set of backings to `Decode`? We'd need to change `bve` to contain a map instead of a list so we can just pass it here
___
*[`internal/manifest/version_edit.go` line 58 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NU4Tv8T9CvVd1XEbwAP:-NU4Tv8T9CvVd1XEbwAQ:bbm2tv3) ([raw file](https://github.com/cockroachdb/pebble/blob/d9f53d3c4da880412d5be0660f4024d9537895e7/internal/manifest/version_edit.go#L58)):*
> ```Go
> 	customTagTerminate         = 1
> 	customTagNeedsCompaction   = 2
> 	customTagVirtual           = 3
> ```

I'd move it after NonSafeIgnoreMask so that older code doesn't try to silently skip it
___
*[`internal/manifest/version_edit.go` line 557 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NU4UE9G25KpeoqmS9GJ:-NU4UE9G25KpeoqmS9GK:b-icj8zd) ([raw file](https://github.com/cockroachdb/pebble/blob/d9f53d3c4da880412d5be0660f4024d9537895e7/internal/manifest/version_edit.go#L557)):*
> ```Go
> type VersionEditDecoder struct {
> 	byteReader
> 	backingTables map[base.FileNum]*FileBacking
> ```

[nit] a comment would be good here

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`version_set.go` line 237 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NU4VYJM6zF97ABBTKp1:-NUcmW9HDQv4wHtyCAn1:bov1r4a) ([raw file](https://github.com/cockroachdb/pebble/blob/d9f53d3c4da880412d5be0660f4024d9537895e7/version_set.go#L237)):*

It feels fragile to also store state in the decoder. Could we instead pass the current set of backings to `Decode`? We'd need to change `bve` to contain a map instead of a list so we can just pass it here
</blockquote></details>

The bve populates its `AddedFileBacking` by reading the `CreatedBackingTables` in the version edit in the `bve.Accumulate` step. But the same version edit may have a new backing table `b1` and a virtual sstable `v1` backed by `b1`. Since we only accumulate after a call to `Decode` `b1` wouldn't be present in `bve.AddedFileBacking`, so we'd have to look for `b1` in `ve.CreatedBackingTables`.

I think it's simpler to maintain another map which will contain all the backing tables necessary. We don't `Decode` often anyway. What do you think?

bananabrick:
Tests don't fail locally. Looking.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`version_set.go` line 237 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NU4VYJM6zF97ABBTKp1:-NUhVn5k4Ts_yF88b52M:bzgs3cq) ([raw file](https://github.com/cockroachdb/pebble/blob/d9f53d3c4da880412d5be0660f4024d9537895e7/version_set.go#L237)):*

The bve populates its `AddedFileBacking` by reading the `CreatedBackingTables` in the version edit in the `bve.Accumulate` step. But the same version edit may have a new backing table `b1` and a virtual sstable `v1` backed by `b1`. Since we only accumulate after a call to `Decode` `b1` wouldn't be present in `bve.AddedFileBacking`, so we'd have to look for `b1` in `ve.CreatedBackingTables`.

I think it's simpler to maintain another map which will contain all the backing tables necessary. We don't `Decode` often anyway. What do you think?
</blockquote></details>

I thought about this and I think the current approach might be causing an unnecessary allocation on every call to Decode. I'll test that out and fix if necessary before further review.

bananabrick (commented):
Got rid of the allocation during `Decode` and fixed failing tests. Should be good for another look.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484#-:-NUhzL6H7wkbP43qroBr:b-8z0ue1)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`internal/manifest/version_edit.go` line 604 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NUiBwxlD1vi6oik8G-m:-NUiBwxlD1vi6oik8G-n:bb7yb6h) ([raw file](https://github.com/cockroachdb/pebble/blob/438d58c5fb451e1853cd791513f2a154e48d7755/internal/manifest/version_edit.go#L604)):*
> ```Go
> 				// We need to store the number of entries because it is used to
> 				// perform extrapolation when constructing a virtual sstable
> 				// Reader.
> ```

Even once we have async table stats collections, it's possible for iterator creation for virtual sstables to occur before virtual sstable table stats collection.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`internal/manifest/version_edit.go` line 604 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NUiBwxlD1vi6oik8G-m:-NUiCLFiCyikxORuJMx2:b-lmfy0o) ([raw file](https://github.com/cockroachdb/pebble/blob/438d58c5fb451e1853cd791513f2a154e48d7755/internal/manifest/version_edit.go#L604)):*

Even once we have async table stats collections, it's possible for iterator creation for virtual sstables to occur before virtual sstable table stats collection.
</blockquote></details>

`NumEntries` is used for extrapolation here: https://github.com/cockroachdb/pebble/blob/master/sstable/reader.go#L2857.

itsbilal (commented):
Thanks for doing this! Mostly looking in great shape

As mentioned in standup, I'd encourage you to do a separate PR that replaces the use of `NumEntries` in `MakeVirtualReader` with the size-based inteprolation we talked about yesterday. That will likely merge quickly and then you can rebase this on top of that and make the change to not store the NumEntries in the vedit.

> Reviewed 12 of 12 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484#-:-NVeZaViCaS5FoBZCb9s:b-jn7ija)* status: all files reviewed, 7 unresolved discussions (waiting on @bananabrick and @RaduBerinde)

___
*[`version_set.go` line 237 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NU4VYJM6zF97ABBTKp1:-NVes72mADrNbNsxxrrv:bobopgn) ([raw file](https://github.com/cockroachdb/pebble/blob/d9f53d3c4da880412d5be0660f4024d9537895e7/version_set.go#L237)):*

I thought about this and I think the current approach might be causing an unnecessary allocation on every call to Decode. I'll test that out and fix if necessary before further review.
</blockquote></details>

We can still maintain a singular map as you're doing here, and I agree it's simpler. I also agree with Radu that it seems a little weirder to make the decoder more stateful. How about keeping this function with `versionEdit` as its receiver and not versionEditDecoder, and just passing in a map that it mutates? The map can be a throwaway one just created by `versionSet.load`. And then a lot of the boilerplate changes in `version_edit.go` can be reverted.
___
*[`internal/manifest/version_edit.go` line 335 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVerI_OAAqE5PYIk-by:-NVerI_OAAqE5PYIk-bz:bjnkasx) ([raw file](https://github.com/cockroachdb/pebble/blob/438d58c5fb451e1853cd791513f2a154e48d7755/internal/manifest/version_edit.go#L335)):*
> ```Go
> 			virtualState := struct {
> 				virtual        bool
> 				virtualFileNum uint64
> ```

s/virtualFileNum/backingFileNum/
___
*[`internal/manifest/version_edit.go` line 336 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVerLc10kJ4dJjo-mQc:-NVerLc10kJ4dJjo-mQd:b-wa2i0k) ([raw file](https://github.com/cockroachdb/pebble/blob/438d58c5fb451e1853cd791513f2a154e48d7755/internal/manifest/version_edit.go#L336)):*
> ```Go
> 				virtual        bool
> 				virtualFileNum uint64
> 				numEntries     uint64
> ```

This should be removed as discussed elsewhere
___
*[`internal/manifest/version_edit.go` line 436 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVeqNq57euZZZCkPUJA:-NVeqNq57euZZZCkPUJB:b-5gbqk4) ([raw file](https://github.com/cockroachdb/pebble/blob/438d58c5fb451e1853cd791513f2a154e48d7755/internal/manifest/version_edit.go#L436)):*
> ```Go
> 				}
> 				m.FileBacking = fb
> 				m.Stats.NumEntries = virtualState.numEntries
> ```

This is the only stat stored in the version edit itself, which on its own is enough reason to not store it there.
___
*[`internal/manifest/version_edit.go` line 525 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVerRjkDaMRIzF82tTv:-NVerRjkDaMRIzF82tTw:b8ur9ar) ([raw file](https://github.com/cockroachdb/pebble/blob/438d58c5fb451e1853cd791513f2a154e48d7755/internal/manifest/version_edit.go#L525)):*
> ```Go
> 	}
> 	for _, dfn := range v.RemovedBackingTables {
> 		e.writeUvarint(tagRemovedBackingTable)
> ```

This requires a new format major version, right?
___
*[`internal/manifest/version_edit.go` line 621 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVersnx4m4A_V_7I8ZI:-NVersnx4m4A_V_7I8ZJ:bh3nfek) ([raw file](https://github.com/cockroachdb/pebble/blob/438d58c5fb451e1853cd791513f2a154e48d7755/internal/manifest/version_edit.go#L621)):*
> ```Go
> 	// information about a created FileBacking which might be needed to decode
> 	// a virtual sstable FileMetadata in a subsequent version edit.
> 	backingTables map[base.FileNum]*FileBacking
> ```

the value type should be DiskFileNum (which really should get renamed to BackingFileNum at some point), right?

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484)* status: 1 of 12 files reviewed, 6 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`version_set.go` line 237 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NU4VYJM6zF97ABBTKp1:-NVgMeF6CXKGX9b1tPrW:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/d9f53d3c4da880412d5be0660f4024d9537895e7/version_set.go#L237)):*

We can still maintain a singular map as you're doing here, and I agree it's simpler. I also agree with Radu that it seems a little weirder to make the decoder more stateful. How about keeping this function with `versionEdit` as its receiver and not versionEditDecoder, and just passing in a map that it mutates? The map can be a throwaway one just created by `versionSet.load`. And then a lot of the boilerplate changes in `version_edit.go` can be reverted.
</blockquote></details>

Done.
___
*[`internal/manifest/version_edit.go` line 335 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVerI_OAAqE5PYIk-by:-NVg97cACJEY0feVs4AO:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/438d58c5fb451e1853cd791513f2a154e48d7755/internal/manifest/version_edit.go#L335)):*

s/virtualFileNum/backingFileNum/
</blockquote></details>

Done.
___
*[`internal/manifest/version_edit.go` line 336 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVerLc10kJ4dJjo-mQc:-NVg97zqABzhg1oRlgwh:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/438d58c5fb451e1853cd791513f2a154e48d7755/internal/manifest/version_edit.go#L336)):*

This should be removed as discussed elsewhere
</blockquote></details>

Done.
___
*[`internal/manifest/version_edit.go` line 436 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVeqNq57euZZZCkPUJA:-NVg98qn1El-D87qqhHR:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/438d58c5fb451e1853cd791513f2a154e48d7755/internal/manifest/version_edit.go#L436)):*

This is the only stat stored in the version edit itself, which on its own is enough reason to not store it there.
</blockquote></details>

Done.
___
*[`internal/manifest/version_edit.go` line 525 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVerRjkDaMRIzF82tTv:-NVg99cq4ysu_7olAOei:b-drmpvi) ([raw file](https://github.com/cockroachdb/pebble/blob/438d58c5fb451e1853cd791513f2a154e48d7755/internal/manifest/version_edit.go#L525)):*

This requires a new format major version, right?
</blockquote></details>

Yea. Not necessarily in this pr.
___
*[`internal/manifest/version_edit.go` line 621 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVersnx4m4A_V_7I8ZI:-NVgK1QV853HSHZXUnLW:b3sup60) ([raw file](https://github.com/cockroachdb/pebble/blob/438d58c5fb451e1853cd791513f2a154e48d7755/internal/manifest/version_edit.go#L621)):*

the value type should be DiskFileNum (which really should get renamed to BackingFileNum at some point), right?
</blockquote></details>

Used FileNum because we're reading some bytes from disk to populate this map, and converting that to a FileNum. We can't create a DiskFileNum, without creating a FileNum first.. All DiskFileNums are still a FileNums.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484)* status: 1 of 12 files reviewed, 8 unresolved discussions (waiting on @bananabrick, @itsbilal, and @RaduBerinde)

___
*[`internal/manifest/version_edit.go` line 52 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NU346OW0-044gZfBARo:-NU346OW0-044gZfBARp:btungeo) ([raw file](https://github.com/cockroachdb/pebble/blob/ba26ba0236dc1a49821d6e442fabfad1504a3d1e/internal/manifest/version_edit.go#L52)):*
> ```Go
> 	// Pebble tags.
> 	tagNewFile5            = 104 // Range keys.
> 	tagCreatedBackingTable = 106
> ```

is skipping 105 intentional?
___
*[`internal/manifest/version_edit.go` line 150 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVjat4Y6b055Z7jZnT7:-NVjat4Y6b055Z7jZnT8:b2mwhcg) ([raw file](https://github.com/cockroachdb/pebble/blob/4d94783900bc0e71b2a009b4de735fe6dc2ad49d/internal/manifest/version_edit.go#L150)):*
> ```Go
> // virtual sstable and FileBacking decoding, and the same map must be passed
> // across multiple calls to Decode. backingTables can be non-nil if no virtual
> // sstables will be decoded.
> ```

This new decoding statefulness is unfortunate. Is it really necessary? IMO, we should keep decoding stateless and push the responsibility of any additional initialization of outside state into the caller. Combining the two makes it more difficult to understand what truly was encoded within the manifest, versus what was derived from outside state.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484)* status: 1 of 12 files reviewed, 8 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`internal/manifest/version_edit.go` line 150 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVjat4Y6b055Z7jZnT7:-NVjfIaoEMJtS3DXKbFT:b3jcuv6) ([raw file](https://github.com/cockroachdb/pebble/blob/4d94783900bc0e71b2a009b4de735fe6dc2ad49d/internal/manifest/version_edit.go#L150)):*

This new decoding statefulness is unfortunate. Is it really necessary? IMO, we should keep decoding stateless and push the responsibility of any additional initialization of outside state into the caller. Combining the two makes it more difficult to understand what truly was encoded within the manifest, versus what was derived from outside state.
</blockquote></details>

The extra state is used for figuring out the `FileBacking` for a virtual sstable in given version edit. The `FileBacking` could've been created in a previous version edit.

We could not store the extra state, and instead set the `FileBacking` for the virtual sstables after we've decoded every version edit in a manifest. But then the decode process would be creating incomplete `FileMetadata`s. There's also multiple places in the code where the version edits are decoded, and all of those places will have to set the `FileBacking` for the virtual sstables after every version edit in the manifest has been decoded.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484)* status: 1 of 12 files reviewed, 8 unresolved discussions (waiting on @bananabrick, @itsbilal, and @RaduBerinde)

___
*[`internal/manifest/version_edit.go` line 150 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVjat4Y6b055Z7jZnT7:-NVjpDhIEzimX_2MMBxG:b-ecnyoy) ([raw file](https://github.com/cockroachdb/pebble/blob/4d94783900bc0e71b2a009b4de735fe6dc2ad49d/internal/manifest/version_edit.go#L150)):*

The extra state is used for figuring out the `FileBacking` for a virtual sstable in given version edit. The `FileBacking` could've been created in a previous version edit.

We could not store the extra state, and instead set the `FileBacking` for the virtual sstables after we've decoded every version edit in a manifest. But then the decode process would be creating incomplete `FileMetadata`s. There's also multiple places in the code where the version edits are decoded, and all of those places will have to set the `FileBacking` for the virtual sstables after every version edit in the manifest has been decoded.
</blockquote></details>

Understood—I think it's still important for decoding to be a pure function of the encoded data.

We can still incrementally populate the `FileBacking` field, but as a separate step after decoding, no? Can we use `BulkVersionEdit` for this purpose? Decode already leaves `VersionEdit.DeletedFiles` incomplete, because the data encoded within the version edit itself is not sufficient to reconstruct the deleted `*fileMetadata`.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484)* status: 1 of 12 files reviewed, 8 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`internal/manifest/version_edit.go` line 52 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NU346OW0-044gZfBARo:-NW-cpdH0w0MW-5og0eg:bssmdj3) ([raw file](https://github.com/cockroachdb/pebble/blob/ba26ba0236dc1a49821d6e442fabfad1504a3d1e/internal/manifest/version_edit.go#L52)):*

is skipping 105 intentional?
</blockquote></details>

Oversight. I had another tag previously which I got rid of.
___
*[`internal/manifest/version_edit.go` line 150 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVjat4Y6b055Z7jZnT7:-NW-cxAA0HGMKJYIVxMj:bctq5uk) ([raw file](https://github.com/cockroachdb/pebble/blob/4d94783900bc0e71b2a009b4de735fe6dc2ad49d/internal/manifest/version_edit.go#L150)):*

Understood—I think it's still important for decoding to be a pure function of the encoded data.

We can still incrementally populate the `FileBacking` field, but as a separate step after decoding, no? Can we use `BulkVersionEdit` for this purpose? Decode already leaves `VersionEdit.DeletedFiles` incomplete, because the data encoded within the version edit itself is not sufficient to reconstruct the deleted `*fileMetadata`. 
</blockquote></details>

Yea, that makes sense to me.

I made that information part of the `NewFileEntry` along with the file metadata. By not populating the FileBacking, we no longer have the property that Encoding a version edit, and then decoding it, will give us the original version edit. The round trip no longer works for virtual sstables. I guess it doesn't work for `DeletedFiles` either.

I've made the requested updates. We not populate the FileBacking in `bve.Accumulate` similar to populating the FileMetadata for a deleted file in bve.Accumulate.

jbowens (approved):
:lgtm:

> Reviewed 1 of 12 files at r4, 2 of 11 files at r7, 2 of 6 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484#-:-NW8FP_fEYt2Xdf3Of4k:bnfp4nl)* status: 5 of 12 files reviewed, 10 unresolved discussions (waiting on @bananabrick, @itsbilal, and @RaduBerinde)

___
*[`internal/manifest/version_edit.go` line 150 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVjat4Y6b055Z7jZnT7:-NW8FOOT2hlErLUGjLKx:b-k73z76) ([raw file](https://github.com/cockroachdb/pebble/blob/4d94783900bc0e71b2a009b4de735fe6dc2ad49d/internal/manifest/version_edit.go#L150)):*

Yea, that makes sense to me.

I made that information part of the `NewFileEntry` along with the file metadata. By not populating the FileBacking, we no longer have the property that Encoding a version edit, and then decoding it, will give us the original version edit. The round trip no longer works for virtual sstables. I guess it doesn't work for `DeletedFiles` either.

I've made the requested updates. We not populate the FileBacking in `bve.Accumulate` similar to populating the FileMetadata for a deleted file in bve.Accumulate.
</blockquote></details>

thanks
___
*[`internal/manifest/version_edit.go` line 434 at r8](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NW8DTdT2_AEYQtfH7J-:-NW8DTdT2_AEYQtfH7J0:b-rnasdv) ([raw file](https://github.com/cockroachdb/pebble/blob/67c0cd4914d80f67b9ef507ee588952b86cb18dc/internal/manifest/version_edit.go#L434)):*
> ```Go
> 			if virtualState.virtual {
> 				v.NewFiles[len(v.NewFiles)-1].BackingFileNum = base.FileNum(virtualState.backingFileNum).DiskFileNum()
> 			}
> ```

nit: i think this would read a little easier to use a local variable for `NewFileEntry` and set `BackingFileNum` before appending it to `v.NewFiles`
___
*[`internal/manifest/version_edit.go` line 792 at r8](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NW8EKZsDw8XZdAfLHp8:-NW8EKZsDw8XZdAfLHp9:b-76sh8m) ([raw file](https://github.com/cockroachdb/pebble/blob/67c0cd4914d80f67b9ef507ee588952b86cb18dc/internal/manifest/version_edit.go#L792)):*
> ```Go
> 		} else if nf.Meta.FileBacking == nil {
> 			return errors.Errorf("Added file L%d.%s's has no FileBacking", nf.Level, nf.Meta.FileNum)
> 		}
> ```

nit: worth nesting these conditions?

```
if nf.Meta.FileBacking == nil {
   if !nf.Meta.Virtual {
       return // ...
   }
   // ...
}
```
___
*[`internal/manifest/version_edit_test.go` line 28 at r8](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NW8F4X-9Mc7XT2Gfv23:-NW8F4X-9Mc7XT2Gfv24:bgt7t23) ([raw file](https://github.com/cockroachdb/pebble/blob/67c0cd4914d80f67b9ef507ee588952b86cb18dc/internal/manifest/version_edit_test.go#L28)):*
> ```Go
> func checkRoundTrip(e0 VersionEdit) error {
> 	var e1 VersionEdit
> 	var err error
> ```

nit: not sure if this was intentional? the previous code kept `err` values narrowly scoped to the conditionals

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484)* status: 3 of 13 files reviewed, 9 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`internal/manifest/version_edit.go` line 150 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NVjat4Y6b055Z7jZnT7:-NW9trFa0awjqBHY4OC-:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/4d94783900bc0e71b2a009b4de735fe6dc2ad49d/internal/manifest/version_edit.go#L150)):*

thanks
</blockquote></details>

Done.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2484)* status: 3 of 13 files reviewed, 8 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`internal/manifest/version_edit.go` line 792 at r8](https://reviewable.io/reviews/cockroachdb/pebble/2484#-NW8EKZsDw8XZdAfLHp8:-NW9vsRu-x-TW7GvO8WN:bizdwyl) ([raw file](https://github.com/cockroachdb/pebble/blob/67c0cd4914d80f67b9ef507ee588952b86cb18dc/internal/manifest/version_edit.go#L792)):*

nit: worth nesting these conditions?

```
if nf.Meta.FileBacking == nil {
   if !nf.Meta.Virtual {
       return // ...
   }
   // ...
}
```
</blockquote></details>

Tried this out. Seemed less clear to me.

bananabrick:
Okay, the flaky msan failure was due to an unclosed db reference in a test I wrote. Closed the db now.

Issue #2196, opened 2022-12-19
db: ingestion can break sequence number invariant


bananabrick:
Ingestions have a [prepare](https://github.com/cockroachdb/pebble/blob/master/ingest.go#L753) step where we look for memtable overlap with the ingestion, and an [apply](https://github.com/cockroachdb/pebble/blob/master/ingest.go#L778) step where we move the ingested sstables into the lsm.

I have an inkling that detecting overlap like this in the prepare step isn't enough.

Consider the case where the lsm has a single memtable `M`. In the `prepare` step, the ingestion doesn't detect overlap with `M`. Once the `apply` step starts to run, the ingesting goroutine isn't holding any mutex(`db.mu` or `commitPipeline.mu`). Now, batches, which contain keys which overlap with the ingestion, with sequence numbers higher than the ingestion sequence number can get applied to `M`, which could force a flush on `M`. So, we could end up flushing `M`, before the apply step of the ingestion finishes. Let's say the sstables generated due to the flush are called `S`.

In this case, the ingested sstables could be layered on top of `S` in the apply step. But the ingested sstables have keys which overlap with `S`, and have sequence numbers lower than `S`. Thus inverting the sequence number invariant of the lsm.

Note that the `d.ingestApply` function within the `apply` function does grab `db.mu`, but there's a window where no mutex is held.

sumeerbhola:
After re-reading the code it does look like a bug.
We do want flushes to proceed for the memtables we want flushed due to overlap. But we don't want the current mutable memtable after this loop https://github.com/cockroachdb/pebble/blob/6ca7956d84c1ce37036c015a4069aafc2f31e516/ingest.go#L763-L773 has finished, from being flushed (which will block future mutable memtables from also being flushed). So after that loop we could increment `memTable.writerRefs` and decrement it in `DB.ingestApply` after we grab `d.mu.versions.logLock()`. Once we have the lock on the `versionSet` we can allow the flush to proceed since it won't be able to apply the version edit until after this ingest.
 
But what prevents two concurrent ingestions from racing each other and allocating seqnums in one order Ingest1 before Ingest2, and then Ingest2 grabbing the `d.mu.versions.logLock()` before Ingest1? If they have overlapping files, we will place them such that Ingest1 files are at a higher level than Ingest2 but that violates the seqnum invariant. This should be

jbowens:
[...]

PR #2486, opened 2023-04-26
db: synchronize ingests with flushes of later memtables

jbowens:
Pebble's consistency relies on maintaining the sequence number invariant: for
two internal keys k#s1 and k#s2 with the same user key and s1 < s2, k#s2 must
be in a higher level of the LSM.

Previously, it was possible for a sequence number inversion to occur during a
concurrent ingest and batch application writing to the same key. If the ingest
acquired its sequence number before the batch, but the batch application
completed first AND the containing memtable flushed first, the batch's higher
sequenced key could exist below the older conflicting key.

Informs #2196.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -846,6 +846,7 @@ func (d *DB) ingest(
 	// approximate ingest-into-L0 stats when using flushable ingests.
 	metaFlushableOverlaps := make([]bool, len(meta))
 	var mem *flushableEntry
+	var mut *memTable
 	// asFlushable indicates whether the sstable was ingested as a flushable.
 	var asFlushable bool
 	prepare := func(seqNum uint64) {
@@ -890,8 +891,20 @@ func (d *DB) ingest(
 				d.opts.Logger.Infof("ingest error reading flushable for log %s: %s", m.logNum, err)
 			}
 		}
+
 		if mem == nil {
-			// No overlap with any of the queued flushables.
+			// No overlap with any of the queued flushables, so no need to queue
+			// after them.
+
+			// New writes with higher sequence numbers may be concurrently
+			// committed. We must ensure they don't flush before this ingest
+			// completes. To do that, we ref the mutable memtable as a writer,
+			// preventing its flushing (and the flushing of all subsequent
+			// flushables in the queue). Once we've acquired the manifest lock
+			// to add the ingested sstables to the LSM, we can unref as we're
+			// guaranteed that the flush won't edit the LSM before this ingest.
+			mut = d.mu.mem.mutable
+			mut.writerRef()
 			return
 		}
 		// The ingestion overlaps with some entry in the flushable queue.
@@ -903,6 +916,15 @@ func (d *DB) ingest(
 			if mem.flushable == d.mu.mem.mutable {
 				err = d.makeRoomForWrite(nil)
 			}
+			// New writes with higher sequence numbers may be concurrently
+			// committed. We must ensure they don't flush before this ingest
+			// completes. To do that, we ref the mutable memtable as a writer,
+			// preventing its flushing (and the flushing of all subsequent
+			// flushables in the queue). Once we've acquired the manifest lock
+			// to add the ingested sstables to the LSM, we can unref as we're
+			// guaranteed that the flush won't edit the LSM before this ingest.
+			mut = d.mu.mem.mutable
+			mut.writerRef()
 			mem.flushForced = true
 			d.maybeScheduleFlush()
 			return
@@ -917,6 +939,9 @@ func (d *DB) ingest(
 	apply := func(seqNum uint64) {
 		if err != nil || asFlushable {
 			// An error occurred during prepare.
+			if mut != nil {
+				mut.writerUnref()
+			}
 			return
 		}
 
@@ -927,6 +952,9 @@ func (d *DB) ingest(
 		if err = ingestUpdateSeqNum(
 			d.cmp, d.opts.Comparer.FormatKey, seqNum, meta,
 		); err != nil {
+			if mut != nil {
+				mut.writerUnref()
+			}
 			return
 		}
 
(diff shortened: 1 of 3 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2486)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[`ingest.go` line 907 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2486#-NTyf_Cn0nDPo15ubuMz:-NTyf_Cn0nDPo15ubuN-:bc1zrel) ([raw file](https://github.com/cockroachdb/pebble/blob/676e43ea14258537ee152f71981edde1b49906a0/ingest.go#L907)):*
> ```Go
> 			// guaranteed that the flush won't edit the LSM before this ingest.
> 			mut = d.mu.mem.mutable
> 			mut.writerRef()
> ```

Instead of ref-ing the memtable, we could allow the flush to proceed but refuse to let it commit until the ingest updates a condition variable. This would allow slightly more pipelining, and in practice we expect the flush to take much longer than it'll take for the ingest to complete. It might also result in more understandable invariants. I'm going to take a pass at it.

sumeerbhola (commented):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2486)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`ingest.go` line 907 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2486#-NTyf_Cn0nDPo15ubuMz:-NTyh3f81gEfuDkyxura:b-y303fx) ([raw file](https://github.com/cockroachdb/pebble/blob/676e43ea14258537ee152f71981edde1b49906a0/ingest.go#L907)):*

Instead of ref-ing the memtable, we could allow the flush to proceed but refuse to let it commit until the ingest updates a condition variable. This would allow slightly more pipelining, and in practice we expect the flush to take much longer than it'll take for the ingest to complete. It might also result in more understandable invariants. I'm going to take a pass at it.
</blockquote></details>

Ack. The change here also looks good.

There isn't any real work happening between the `writerRef` and the `writerUnref`, so I am not worried about loss of flush throughput. We do wait for `<-mem.flushed`, but then the other flush cannot start either.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2486)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`ingest.go` line 907 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2486#-NTyf_Cn0nDPo15ubuMz:-NTzOQou0qXLevQAj169:b6gad37) ([raw file](https://github.com/cockroachdb/pebble/blob/676e43ea14258537ee152f71981edde1b49906a0/ingest.go#L907)):*

Ack. The change here also looks good.

There isn't any real work happening between the `writerRef` and the `writerUnref`, so I am not worried about loss of flush throughput. We do wait for `<-mem.flushed`, but then the other flush cannot start either.
</blockquote></details>

Yeah, I think I'll follow up later to see if there's another approach, but I don't think it'll be as straightforwardly backportable so I rather merge+backport this as-is first.

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2486#-:-NU-g3x6DErGHt01HXF-:bg7igkv)* status: all files reviewed, 3 unresolved discussions (waiting on @bananabrick and @jbowens)

___
*[`ingest_test.go` line 1525 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2486#-NU-f9CJBY-eCqY7mopu:-NU-f9CJBY-eCqY7mopv:bvfi4dw) ([raw file](https://github.com/cockroachdb/pebble/blob/fc5aca51d1df043af78542647c249ccfd9d04703/ingest_test.go#L1525)):*
> ```Go
> 
> 	// Replay the manifest. Every flush and ingest is a separate version edit.
> 	// Since they all write the same key and compacitons are disabled, sequence
> ```

nit: compactions
___
*[`ingest_test.go` line 1526 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2486#-NU-f_IzAkgEEsdQFPr6:-NU-f_IzAkgEEsdQFPr7:b300tcx) ([raw file](https://github.com/cockroachdb/pebble/blob/fc5aca51d1df043af78542647c249ccfd9d04703/ingest_test.go#L1526)):*
> ```Go
> 	// Replay the manifest. Every flush and ingest is a separate version edit.
> 	// Since they all write the same key and compacitons are disabled, sequence
> 	// numbers of new files should be monotonically increasing.
> ```

Is this `VersionEdit` checking needed since all these files are in L0, which will order them by LargestSeqNums, so `CheckLevels` won't catch the inversion that happened when adding to L0?
If so, worth clarifying in a code comment.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2486#-:-NU37og26yWUsJUJ2cSD:b-814g8k)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @bananabrick and @sumeerbhola)

___
*[`ingest_test.go` line 1526 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2486#-NU-f_IzAkgEEsdQFPr6:-NU36O9eAXAI6tYrH7Oo:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/fc5aca51d1df043af78542647c249ccfd9d04703/ingest_test.go#L1526)):*

Is this `VersionEdit` checking needed since all these files are in L0, which will order them by LargestSeqNums, so `CheckLevels` won't catch the inversion that happened when adding to L0?
If so, worth clarifying in a code comment.
</blockquote></details>

Done.

RaduBerinde:
Do we need to update the comment for `writerRefs` with the new source of references?
```go
	// writerRefs tracks the write references on the memtable. The two sources of
	// writer references are the memtable being on DB.mu.mem.queue and from
	// inflight mutations that have reserved space in the memtable but not yet
	// applied. The memtable cannot be flushed to disk until the writer refs
	// drops to zero.
	writerRefs atomic.Int32
```

PR #2490, opened 2023-04-26
db_test: Add a unit test to exercise memtable/ingestion inversion

itsbilal:
This change adds a unit test that utilizes new compaction-blocking hooks in db/compaction.go to reproduce cockroachdb/cockroach#101896.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -1284,6 +1284,351 @@ func TestSSTables(t *testing.T) {
 	}
 }
 
+func TestMemtableIngestInversion(t *testing.T) {
+	memFS := vfs.NewMem()
+	opts := &Options{
+		FS:                          memFS,
+		MemTableSize:                256 << 10, // 4KB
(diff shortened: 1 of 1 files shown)


itsbilal:
@sumeerbhola @jbowens is this worth polishing up and merging? reviews welcome either way. this always fails in `crl-release-22.2` pre cockroachdb/cockroach#101896 fix, but passes after it.

jbowens:
Independently of whether we're able to backport a version of this integration unit test, maybe we should write a randomized L0Sublevels unit test that would've caught the issue and backport that.

itsbilal:
I'll close this out in favour of #2508 (which is an improved version of this test that doesn't rely on any non-test code changes, instead relying heavily on the EventListener), and #2509 which adds an invariant assertion which combined with the metamorphic test gives us coverage into future incremental-vs-regeneration divergences.

PR #2491, opened 2023-04-26
objstorage: don't use mutex in sharedReadable.ReadAt

RaduBerinde:
Instead of serializing all `ReadAt` calls against a `Readable`, use a temporary throw-away handle. The end result is similar (we don't expect to be able to reuse the underlying reader in the common case) but without serialization.

Reads without a ReadHandle happen for various sst metadata (like index or filter blocks).


itsbilal (approved):
LGTM

RaduBerinde:
TFTR!

PR #2492, opened 2023-04-26
sstable: print magic number when encountering corruption

nicktrav:
Currently, if the magic number in an SSTable footer is invalid, an error is printed indicating as such. However, it would be useful to know what the magic number was for the SSTable. This allows for potentially faster issue triaging.

Print the magic number bytes when encountering an invalid table magic number.

PR #2496, opened 2023-04-26
crl-release-23.1: sstable: print magic number when encountering corruption

nicktrav:
This is a backport of #2492 to 23.1.

---

Currently, if the magic number in an SSTable footer is invalid, an error is printed indicating as such. However, it would be useful to know what the magic number was for the SSTable. This allows for potentially faster issue triaging.

Print the magic number bytes when encountering an invalid table magic number.


RaduBerinde (approved):

nicktrav:
TFTR!

PR #2498, opened 2023-04-27
db: allow locking directory before Open

jbowens:
Allow callers of Open to acquire the database directory's file lock in advance of Open. This allows clients like Cockroach to use the existing LOCK file to protect additional on-disk data within the database directory. This is preferrable to a separate file lock, because in a mixed version scenario it avoids a newer version mutating state in use by a running process of a previous version.

Informs cockroachdb/cockroach#98294.

### DIFF
--- a/db.go
+++ b/db.go
@@ -264,7 +264,7 @@ type DB struct {
 	// objProvider is used to access and manage SSTs.
 	objProvider objstorage.Provider
 
-	fileLock io.Closer
+	fileLock *Lock
 	dataDir  vfs.File
 	walDir   vfs.File
 
--- a/open.go
+++ b/open.go
@@ -87,9 +87,22 @@ func Open(dirname string, opts *Options) (db *DB, _ error) {
 	}()
 
 	// Lock the database directory.
-	fileLock, err := opts.FS.Lock(base.MakeFilepath(opts.FS, dirname, fileTypeLock, base.FileNum(0).DiskFileNum()))
-	if err != nil {
-		return nil, err
+	var fileLock *Lock
+	if opts.Lock != nil {
+		// The caller already acquired the database lock. Ensure that the
+		// directory matches.
+		if dirname != opts.Lock.dirname {
+			return nil, errors.Newf("pebble: opts.Lock acquired in %q not %q", opts.Lock.dirname, dirname)
+		}
+		if err := opts.Lock.refForOpen(); err != nil {
+			return nil, err
+		}
+		fileLock = opts.Lock
+	} else {
+		fileLock, err = LockDirectory(dirname, opts.FS)
+		if err != nil {
+			return nil, err
+		}
 	}
 	defer func() {
 		if db == nil {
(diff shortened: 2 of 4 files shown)


bananabrick (commented):
:lgtm: But I'm unsure about how this helps with #98294. I could use an explanation for "Allow callers of Open to acquire the database directory's file lock in advance of Open. This allows clients like Cockroach to use the existing LOCK file to protect additional on-disk data within the database directory." for my own understanding.

1. Enc at rest will acquire the lock file, which will prevent Pebble from concurrently opening without the enc rest subsystem passing the Lock in as a pebble option. 
2. So, are we just using the lock file to guarantee that `pebble.Open` will only be called once encryption at rest is finished with its setup?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2498#-:-NU3645J-gxXBv2G1g_d:b-9ikzyo)* status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`open_test.go` line 163 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2498#-NU35BpP8UUu24Y3St07:-NU35BpP8UUu24Y3St08:b-7cqqco) ([raw file](https://github.com/cockroachdb/pebble/blob/c2530da94c69d8aa057385a73068df9535090738/open_test.go#L163)):*
> ```Go
> 		// Try to open the same database reusing the Options containing the same
> 		// Lock. It should error when it observes that it's already referenced.
> 		_, err = Open(dirname, opts)
> ```

nit: Should we also try to Open without passing in the same Opts?

RaduBerinde:
We'd be extending the "locked section" to include the part where we're messing with encryption-at-rest metadata, ensuring that two separate processes don't mess with it at the same time.

jbowens (commented):
thanks :smile_cat: 

TFTRs! Looks like it still needs that official github approval :heavy_check_mark: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2498#-:-NU3PRF06hWLMC-x3Hcg:b3f0ojs)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @bananabrick and @RaduBerinde)

___
*[`open.go` line 998 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2498#-NU2h1MT-_hqGhjzN1FH:-NU3PNgW-UBQm7CuM295:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c2530da94c69d8aa057385a73068df9535090738/open.go#L998)):*

Should we install a finalizer that checks that we have closed the file lock?
</blockquote></details>

Done.
___
*[`open.go` line 1038 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2498#-NU2hABj37fzxFxRp01r:-NU3LvRW5iATaPpdXx1A:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c2530da94c69d8aa057385a73068df9535090738/open.go#L1038)):*

maybe `defer func() { l.fileLock = nil }()`
</blockquote></details>

Done.
___
*[`options.go` line 711 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2498#-NU2f1gm0ZOYmUK4WLOS:-NU3MIp5C5vxT5_ktgHd:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c2530da94c69d8aa057385a73068df9535090738/options.go#L711)):*

[nit] Mention that we check that the locked directory matches, and that concurrent Open calls against the same Lock are detected and disallowed (otherwise it's unclear why this is not just a `skipLocking` flag)
</blockquote></details>

Done.
___
*[`open_test.go` line 155 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2498#-NU2gtjx0fFDasTQR351:-NU3ML8m1l5j7cfJ0jlT:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c2530da94c69d8aa057385a73068df9535090738/open_test.go#L155)):*

[nit] maybe move this at the end of the test, check that there's no returned error, and check that the ref count ends up as 0.
</blockquote></details>

Done.
___
*[`open_test.go` line 163 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2498#-NU35BpP8UUu24Y3St07:-NU3NahZ58F3LxfVpFKL:b-gh0ata) ([raw file](https://github.com/cockroachdb/pebble/blob/c2530da94c69d8aa057385a73068df9535090738/open_test.go#L163)):*

nit: Should we also try to Open without passing in the same Opts?
</blockquote></details>

the behavior of locking a file multiple times from within the same process is not well defined. `vfs.Default` jumps through some hoops to prevent it, but `MemFS` does not. we could unify their behavior in a follow up.

nicktrav (approved):
Proxy-LGTM ✅

jbowens:
TFTRs!

PR #2499, opened 2023-04-27
vfs: mark file basename as safe to avoid log redaction

nicktrav:
When a disk stall is detected, the operation and path of the file are logged. Currently, in Cockroach, the path is redacted as it is not marked as safe.

As a path in a production environment could contain sensitive information (we have no control over how a user configures the directory in which they use for the Pebble store), we want to avoid leaking too much information. Only take the basename of the file, and mark this as safe. The filenames that Pebble uses are well-formed, and should not contain sensitive information.

Touches cockroachdb/cockroach#99756.

### DIFF
--- a/vfs/disk_health.go
+++ b/vfs/disk_health.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -375,7 +376,8 @@ func (i DiskSlowInfo) String() string {
 // SafeFormat implements redact.SafeFormatter.
 func (i DiskSlowInfo) SafeFormat(w redact.SafePrinter, _ rune) {
 	w.Printf("disk slowness detected: %s on file %s (%d bytes) has been ongoing for %0.1fs",
-		redact.Safe(i.OpType.String()), i.Path, i.WriteSize, redact.Safe(i.Duration.Seconds()))
+		redact.Safe(i.OpType.String()), redact.Safe(filepath.Base(i.Path)),
+		redact.Safe(i.WriteSize), redact.Safe(i.Duration.Seconds()))
 }
 
 // diskHealthCheckingFS adds disk-health checking facilities to a VFS.


joshimhoff (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2499#-:-NU2sA4J4sGubGJBY6HZ:bg7igkv)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sumeerbhola)

joshimhoff (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2499)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nicktrav and @sumeerbhola)

___
*[`vfs/disk_health.go` line 378 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2499#-NU2xmSb4Mla9leCPW3X:-NU2xmSb4Mla9leCPW3Y:bbcfh9j) ([raw file](https://github.com/cockroachdb/pebble/blob/5fb8a81b3d8ac9b28d7bc2ca0e51ce4cf93184b3/vfs/disk_health.go#L378)):*
> ```Go
> func (i DiskSlowInfo) SafeFormat(w redact.SafePrinter, _ rune) {
> 	w.Printf("disk slowness detected: %s on file %s (%d bytes) has been ongoing for %0.1fs",
> 		redact.Safe(i.OpType.String()), i.Path, i.WriteSize, redact.Safe(i.Duration.Seconds()))
> ```

As is, will `i.WriteSize` be redacted? I don't think it should. Apologies that I didn't understand this API in detail, when I wrote this code.

nicktrav (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2499)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @joshimhoff and @sumeerbhola)

___
*[`vfs/disk_health.go` line 378 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2499#-NU2xmSb4Mla9leCPW3X:-NU3Jdsg3yT4Qs2oH4o9:b19t20y) ([raw file](https://github.com/cockroachdb/pebble/blob/5fb8a81b3d8ac9b28d7bc2ca0e51ce4cf93184b3/vfs/disk_health.go#L378)):*

As is, will `i.WriteSize` be redacted? I don't think it should. Apologies that I didn't understand this API in detail, when I wrote this code.
</blockquote></details>

Ah good catch. That would have been redacted too. Fixed.

nicktrav:
TFTR!

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2499)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`vfs/disk_health.go` line 378 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2499#-NU2xmSb4Mla9leCPW3X:-NU3QR_R2svcaHQ3CyQy:b-dkf6yi) ([raw file](https://github.com/cockroachdb/pebble/blob/5fb8a81b3d8ac9b28d7bc2ca0e51ce4cf93184b3/vfs/disk_health.go#L378)):*
<details><summary><i>Previously, nicktrav (Nick Travers) wrote…</i></summary><blockquote>

Ah good catch. That would have been redacted too. Fixed.
</blockquote></details>

I think adding the `redact.Safe` is the right move, but in the context of the cockroach binary, I think `i.WriteSize` previously would _not_ be redacted, because [Cockroach registers all the numeric primitive types as safe](https://github.com/cockroachdb/cockroach/blob/c2a51cc078e489bb8b232e525944bac703a003cf/pkg/util/log/redact.go#L142-L167)

nicktrav:
Ah! TIL. That's what I was looking for [over here](https://cockroachlabs.slack.com/archives/C01CNRP6TSN/p1682621006232139), but didn't know exactly what I was asking for.

Let me know if you'd prefer to revert the explicit redaction of the ints. Seems fine to be explicit about it here.

PR #2500, opened 2023-04-27
crl-release-23.1: db: synchronize ingests with flushes of later memtables

jbowens:
23.1 backport of #2486.

----

Pebble's consistency relies on maintaining the sequence number invariant: for two internal keys k#s1 and k#s2 with the same user key and s1 < s2, k#s2 must be in a higher level of the LSM.

Previously, it was possible for a sequence number inversion to occur during a concurrent ingest and batch application writing to the same key. If the ingest acquired its sequence number before the batch, but the batch application completed first AND the containing memtable flushed first, the batch's higher sequenced key could exist below the older conflicting key.

Informs #2196.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -847,6 +847,7 @@ func (d *DB) ingest(
 	// approximate ingest-into-L0 stats when using flushable ingests.
 	metaFlushableOverlaps := make([]bool, len(meta))
 	var mem *flushableEntry
+	var mut *memTable
 	// asFlushable indicates whether the sstable was ingested as a flushable.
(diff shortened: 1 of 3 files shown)


bananabrick (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2500#-:-NUc6uiVAzMqWQDmXrkV:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @RaduBerinde)

jbowens:
TFTR!

PR #486, opened 2020-01-10
internal/metamorphic: use the strict MemFS and add an operation to re…

sumeerbhola:
…set the DB

PR #2501, opened 2023-04-27
crl-release-22.2: db: synchronize ingests with flushes of later memtables

jbowens:
22.2 backport of #486.

----

Pebble's consistency relies on maintaining the sequence number invariant: for two internal keys k#s1 and k#s2 with the same user key and s1 < s2, k#s2 must be in a higher level of the LSM.

Previously, it was possible for a sequence number inversion to occur during a concurrent ingest and batch application writing to the same key. If the ingest acquired its sequence number before the batch, but the batch application completed first AND the containing memtable flushed first, the batch's higher sequenced key could exist below the older conflicting key.

Informs #2196.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -644,18 +644,18 @@ func ingestTargetLevel(
 //
 // The steps for ingestion are:
 //
-//   1. Allocate file numbers for every sstable being ingested.
-//   2. Load the metadata for all sstables being ingest.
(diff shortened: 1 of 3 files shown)


itsbilal (approved):
LGTM

jbowens:
TFTR!

PR #2281, opened 2023-01-26
vfs: include size of write in DiskSlowInfo

joshimhoff:
Closes https://github.com/cockroachdb/cockroach/issues/67856. Previous work at https://github.com/cockroachdb/cockroach/pull/95436 & https://github.com/cockroachdb/pebble/pull/2255.

**vfs: include size of write in DiskSlowInfo**

This commit adds the size of a write to DiskSlowInfo, in cases where a write is sized. A small write stalling out points at file system / disk issues, while a large write taking time to complete may indicate CRDB issues with a certain workload, etc.

PR #2503, opened 2023-04-27
crl-release-23.1: vfs: include size of write in DiskSlowInfo

nicktrav:
This is a backport of #2281 and #2499 to 23.1.

---

**vfs: include size of write in DiskSlowInfo**

This commit adds the size of a write to DiskSlowInfo, in cases where a write is
sized. A small write stalling out points at file system / disk issues, while
a large write taking time to complete may indicate CRDB issues with a certain
workload, etc.

**vfs: mark file basename as safe to avoid log redaction**

When a disk stall is detected, the operation and path of the file are
logged. Currently, in Cockroach, the path is redacted as it is not
marked as safe.

As a path in a production environment could contain sensitive
information (we have no control over how a user configures the directory
in which they use for the Pebble store), we want to avoid leaking too
much information. Only take the basename of the file, and mark this as
safe. The filenames that Pebble uses are well-formed, and should not
contain sensitive information.

### DIFF
--- a/event.go
+++ b/event.go
@@ -119,24 +119,7 @@ func (i levelInfos) SafeFormat(w redact.SafePrinter, _ rune) {
 
 // DiskSlowInfo contains the info for a disk slowness event when writing to a
 // file.
-type DiskSlowInfo struct {
-	// Path of file being written to.
-	Path string
-	// Operation being performed on the file.
-	OpType vfs.OpType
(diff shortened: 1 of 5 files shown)


nicktrav:
Figured these would be useful in 23.1.

RaduBerinde (approved):

PR #2505, opened 2023-04-27
crl-release-22.2: db: allow locking directory before Open

jbowens:
22.2 backport of #2498.

----

Allow callers of Open to acquire the database directory's file lock in advance of Open. This allows clients like Cockroach to use the existing LOCK file to protect additional on-disk data within the database directory. This is preferrable to a separate file lock, because in a mixed version scenario it avoids a newer version mutating state in use by a running process of a previous version.

Informs cockroachdb/cockroach#98294.

### DIFF
--- a/db.go
+++ b/db.go
@@ -244,7 +244,7 @@ type DB struct {
 	// The on-disk size of the current OPTIONS file.
 	optionsFileSize uint64
 
-	fileLock io.Closer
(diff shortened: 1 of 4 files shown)


nicktrav (approved):

jbowens:
TFTR!

PR #2506, opened 2023-04-28
objstorage: prepend a hash value to object names

RaduBerinde:
AWS (and probably some other blob storage providers) use a prefix of the object names to partition the buckets (the length of the prefix is chosen dynamically to get a desired number of partitions). We get best performance when IO load is distributed evenly across the partitions. To achieve this we prepend a hash value to the object name, consisting of 4 hex digits. The hash value is derived from the other parts of the name - the creator ID and creator file num.

In this change we also improve the functions that generate names to avoid unnecessary allocations.


objstorage/objstorageprovider/shared_obj_name.go
```diff
@@ -0,0 +1,80 @@
...
+
+// sharedObjectName returns the name of an object on shared storage.
+//
+// For sstables, the format is: <hash>/<creator-id>-<file-num>.sst
+// For example: 1a3f-00000000000000000002-000001.sst
+func sharedObjectName(meta objstorage.ObjectMetadata) string {
+	switch meta.FileType {
+	case base.FileTypeTable:
+		return fmt.Sprintf(
+			"%04x-%020d-%06d.sst",
```
itsbilal, line 22: This seems to be out of sync with the comment? It's `hash-creatorID-filenum.sst` while the comment says it's `hash/creatorid-filenum.sst`. The latter is probably necessary for performance.

objstorage/objstorageprovider/shared_obj_name.go
```diff
@@ -0,0 +1,80 @@
...
+		)
+	}
+	panic("unknown FileType")
+}
+
+// sharedObjectRefName returns the name of the object's ref marker associated
+// with a given referencing provider. This name is the object's name concatenated with
+// ".ref.<ref-creator-id>.<local-file-num>".
+//
+// For example: 1a3f-00000000000000000002-000001.sst.ref.00000000000000000005.000008
```
itsbilal, line 33: same issue here?

objstorage/objstorageprovider/shared_obj_name.go
```diff
@@ -0,0 +1,80 @@
...
+func sharedObjectRefName(
+	meta objstorage.ObjectMetadata, refCreatorID objstorage.CreatorID, refFileNum base.DiskFileNum,
+) string {
+	if meta.Shared.CleanupMethod != objstorage.SharedRefTracking {
+		panic("ref object used when ref tracking disabled")
+	}
+	switch meta.FileType {
+	case base.FileTypeTable:
+		return fmt.Sprintf(
+			"%04x-%020d-%06d.sst.ref.%020d.%06d",
```
itsbilal, line 43: We should use sharedObjectName instead of repeating the format here?

objstorage/objstorageprovider/shared_obj_name.go
```diff
@@ -0,0 +1,80 @@
...
+		)
+	}
+	panic("unknown FileType")
+}
+
+func sharedObjectRefPrefix(meta objstorage.ObjectMetadata) string {
+	switch meta.FileType {
+	case base.FileTypeTable:
+		return fmt.Sprintf(
+			"%04x-%020d-%06d.sst.ref.",
```
itsbilal, line 54: same, we can use `sharedObjectName`.

sumeerbhola (commented):
looks good to me, apart from @itsbilal 's comments
:lgtm: 

> Reviewed 1 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2506#-:-NU8f_aP7mYdZNy-0jpk:b-n3dr5d)* status: 1 of 9 files reviewed, 4 unresolved discussions (waiting on @joshimhoff and @RaduBerinde)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2506)* status: 1 of 9 files reviewed, 4 unresolved discussions (waiting on @itsbilal and @joshimhoff)

___
*[`objstorage/objstorageprovider/shared_obj_name.go` line 22 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2506#gh-1180762572:-NU8eGyeE3xfMVoCF04N:b-s1eg4i) ([raw file](https://github.com/cockroachdb/pebble/blob/d84a6a6c03198a45029d8e47f7ba2c13c57404a9/objstorage/objstorageprovider/shared_obj_name.go#L22)):*

This seems to be out of sync with the comment? It's `hash-creatorID-filenum.sst` while the comment says it's `hash/creatorid-filenum.sst`. The latter is probably necessary for performance.
</blockquote></details>

I was going with the dash, so that you can more easily browse the objects in the AWS/GCS UI. Otherwise, in most practical cases you would get one "folder" per object and have to click through.

Do you have any reference on why a slash would make a difference for performance? My understanding is that (for partitioning purposes) AWS treats the full object name as an opaque string and dynamically picks a prefix length and partitions objects based on the values of that prefix; the length of the prefix is chosen to get a certain target number of partitions. I used a few hex digits so it can get a lot of partitions if needed.

The AWS docs mention that "the Amazon S3 data model is a flat structure: You create a bucket, and the bucket stores objects. There is no hierarchy of subbuckets or subfolders" https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html
___
*[`objstorage/objstorageprovider/shared_obj_name.go` line 43 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2506#gh-1180763885:-NU8fnzx6EH8iuQQjjj3:b-40suwg) ([raw file](https://github.com/cockroachdb/pebble/blob/d84a6a6c03198a45029d8e47f7ba2c13c57404a9/objstorage/objstorageprovider/shared_obj_name.go#L43)):*

We should use sharedObjectName instead of repeating the format here?
</blockquote></details>

I wanted to avoid a lot of allocations; the test cross-checks the result against that cleaner implementation. Maybe this is silly and unnecessary?

RaduBerinde:
Just a friendly ping.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2506#-:-NV0ucsO4ONE62FNZiQq:b-814g8k)* status: 1 of 11 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @joshimhoff)

PR #2508, opened 2023-04-28
db_test: Add a unit test to exercise memtable/ingestion inversion

itsbilal:
This change adds a unit test that utilizes the EventListener to
slow down compactions/flushes to reproduce https://github.com/cockroachdb/cockroach/issues/101896.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -1374,6 +1374,378 @@ func TestTracing(t *testing.T) {
 	b.Close()
 }
 
+func TestMemtableIngestInversion(t *testing.T) {
+	memFS := vfs.NewMem()
+	opts := &Options{
+		FS:                          memFS,
+		MemTableSize:                256 << 10, // 4KB
+		MemTableStopWritesThreshold: 1000,
+		L0StopWritesThreshold:       1000,
+		L0CompactionThreshold:       2,
+		MaxConcurrentCompactions: func() int {
+			return 1000
+		},
+	}
+
+	const channelTimeout = 5 * time.Second
+
+	// We induce delay in compactions by passing in an EventListener that stalls on
+	// the first TableCreated event for a compaction job we want to block.
+	// FlushBegin and CompactionBegin has info on compaction start/output levels
+	// which is what we need to identify what compactions to block. However
+	// FlushBegin and CompactionBegin are called while holding db.mu, so we cannot
+	// block those events forever. Instead, we grab the job ID from those events
+	// and store it. Then during TableCreated, we check if we're creating an output
+	// for a job we have identified earlier as one to block, and then hold on a
+	// semaphore there until there's a signal from the test code to resume with the
+	// compaction.
+	//
+	// If nextBlockedCompaction is non-zero, we must block the next compaction
+	// out of the nextBlockedCompaction - 3 start level. 1 means block the next
+	// intra-L0 compaction and 2 means block the next flush (as flushes have
+	// a -1 start level).
+	var nextBlockedCompaction, blockedJobID int
+	var blockedCompactionsMu sync.Mutex // protects the above two variables.
+	nextSem := make(chan chan struct{}, 1)
+	var el EventListener
+	el.EnsureDefaults(DefaultLogger)
+	el.FlushBegin = func(info FlushInfo) {
+		blockedCompactionsMu.Lock()
+		defer blockedCompactionsMu.Unlock()
+		if nextBlockedCompaction == 2 {
+			nextBlockedCompaction = 0
+			blockedJobID = info.JobID
+		}
+	}
+	el.CompactionBegin = func(info CompactionInfo) {
+		// 0 = block nothing, 1 = block intra-L0 compaction, 2 = block flush,
+		// 3 = block L0 -> LBase compaction, 4 = block compaction out of L1, and so on.
+		blockedCompactionsMu.Lock()
+		defer blockedCompactionsMu.Unlock()
+		blockValue := info.Input[0].Level + 3
+		if info.Input[0].Level == 0 && info.Output.Level == 0 {
+			// Intra L0 compaction, denoted by casValue of 1.
+			blockValue = 1
+		}
+		if nextBlockedCompaction == blockValue {
+			nextBlockedCompaction = 0
+			blockedJobID = info.JobID
+		}
+	}
+	el.TableCreated = func(info TableCreateInfo) {
+		blockedCompactionsMu.Lock()
+		if info.JobID != blockedJobID {
+			blockedCompactionsMu.Unlock()
+			return
+		}
+		blockedJobID = 0
+		blockedCompactionsMu.Unlock()
+		sem := make(chan struct{})
+		nextSem <- sem
+		<-sem
+	}
+	tel := TeeEventListener(MakeLoggingEventListener(DefaultLogger), el)
+	opts.EventListener = &tel
+	opts.Experimental.L0CompactionConcurrency = 1
+	d, err := Open("", opts)
+	require.NoError(t, err)
+	defer func() {
+		if d != nil {
+			require.NoError(t, d.Close())
+		}
+	}()
+
+	printLSM := func() {
+		d.mu.Lock()
+		s := d.mu.versions.currentVersion().String()
+		d.mu.Unlock()
+		t.Logf("%s", s)
+	}
+
+	// Create some sstables. These should go into L6. These are irrelevant for
+	// the rest of the test.
+	require.NoError(t, d.Set([]byte("b"), []byte("foo"), nil))
+	require.NoError(t, d.Flush())
+	require.NoError(t, d.Set([]byte("d"), []byte("bar"), nil))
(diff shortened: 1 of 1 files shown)


RaduBerinde (commented):
Interesting approach!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2508#-:-NU87DXk-ZIPfi_fp6Ly:bfmwbf0)* status: 0 of 1 files reviewed, 7 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2508#-NU849zZCevz4LdbSIZ9):*
This seems stale (no hooks used)
___
*[`db_test.go` line 1377 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2508#-NU84WghBrg2b3uQka_9:-NU84WghBrg2b3uQka_A:bken3fb) ([raw file](https://github.com/cockroachdb/pebble/blob/ff930a8664f8a5e1fd5a02ae705c384134e988d8/db_test.go#L1377)):*
> ```Go
> }
> 
> func TestMemtableIngestInversion(t *testing.T) {
> ```

[nit] It would be useful to explain the high level approach, namely that we will stall completion of certain compactions in the `TableCreated` callback
___
*[`db_test.go` line 1394 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2508#-NU85h2M4IsUQ1MzKAFj:-NU85h2M4IsUQ1MzKAFk:b-ybpxsf) ([raw file](https://github.com/cockroachdb/pebble/blob/ff930a8664f8a5e1fd5a02ae705c384134e988d8/db_test.go#L1394)):*
> ```Go
> 	// intra-L0 compaction and 2 means block the next flush (as flushes have
> 	// a -1 start level).
> 	var nextBlockedCompaction int32
> ```

Why are we mixing mutexes and atomics? Can't everything be protected by the mutex? It would help to put all this state in a separate struct with a few helper methods.
___
*[`db_test.go` line 1397 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2508#-NU84gSR3mL9vHYQrUl3:-NU84gSR3mL9vHYQrUl4:b-3wz5ku) ([raw file](https://github.com/cockroachdb/pebble/blob/ff930a8664f8a5e1fd5a02ae705c384134e988d8/db_test.go#L1397)):*
> ```Go
> 	var el EventListener
> 	el.EnsureDefaults(DefaultLogger)
> 	nextSem := make(chan chan struct{}, 1)
> ```

[nit] This channel of channels deserves a comment. Btw given that we're only blocking one thing at a time, we could use the same fixed "resume" channel instead of creating a new one each time.
___
*[`db_test.go` line 1398 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2508#-NU87bSJDeoYJzxIiJtv:-NU87bSJDeoYJzxIiJtw:b-6g3g4) ([raw file](https://github.com/cockroachdb/pebble/blob/ff930a8664f8a5e1fd5a02ae705c384134e988d8/db_test.go#L1398)):*
> ```Go
> 	el.EnsureDefaults(DefaultLogger)
> 	nextSem := make(chan chan struct{}, 1)
> 	blockedCompactions := make(map[int]bool)
> ```

Unless I'm missing something, it seems that this map never has more than 1 entry that is `true`, is this right? If yes, we can just make it a single `blockedJobID`
___
*[`db_test.go` line 1463 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2508#-NU88GXu2xibG0tTNOgf:-NU88GXu2xibG0tTNOgg:b-a20ywp) ([raw file](https://github.com/cockroachdb/pebble/blob/ff930a8664f8a5e1fd5a02ae705c384134e988d8/db_test.go#L1463)):*
> ```Go
> 	// compactions later on.
> 	atomic.StoreInt32(&nextBlockedCompaction, 3)
> 	timeoutSem := time.After(1 * time.Second)
> ```

We should move this timeout to a constant (it's used multiple times), and make it bigger (we see even simpler stuff go over all the time, for reference CRDB's SucceedsSoon is 45s).
___
*[`db_test.go` line 1475 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2508#-NU86t6YEefOGJwOrrXW:-NU86t6YEefOGJwOrrXX:b3si122) ([raw file](https://github.com/cockroachdb/pebble/blob/ff930a8664f8a5e1fd5a02ae705c384134e988d8/db_test.go#L1475)):*
> ```Go
> 			t.Fatal("did not get blocked on an LBase compaction")
> 		default:
> 			require.NoError(t, d.Set([]byte("b"), []byte("foo"), nil))
> ```

This is a very hot loop.. might make sense to add a short sleep.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2508#-:-NU8ZAyO008s7srLswcE:b-814g8k)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2508#-NU849zZCevz4LdbSIZ9):*

This seems stale (no hooks used)
</blockquote></details>

Done.
___
*[`db_test.go` line 1377 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2508#-NU84WghBrg2b3uQka_9:-NU8Xb0x2ZV04WVn5_SN:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/ff930a8664f8a5e1fd5a02ae705c384134e988d8/db_test.go#L1377)):*

[nit] It would be useful to explain the high level approach, namely that we will stall completion of certain compactions in the `TableCreated` callback
</blockquote></details>

Done.
___
*[`db_test.go` line 1394 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2508#-NU85h2M4IsUQ1MzKAFj:-NU8Xblj9bVApPF8yvr5:b6etrwv) ([raw file](https://github.com/cockroachdb/pebble/blob/ff930a8664f8a5e1fd5a02ae705c384134e988d8/db_test.go#L1394)):*

Why are we mixing mutexes and atomics? Can't everything be protected by the mutex? It would help to put all this state in a separate struct with a few helper methods.
</blockquote></details>

Done, using the mutex for both.
___
*[`db_test.go` line 1397 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2508#-NU84gSR3mL9vHYQrUl3:-NU8XlGe0OnVS_ygLCPO:b-oeyk2g) ([raw file](https://github.com/cockroachdb/pebble/blob/ff930a8664f8a5e1fd5a02ae705c384134e988d8/db_test.go#L1397)):*

[nit] This channel of channels deserves a comment. Btw given that we're only blocking one thing at a time, we could use the same fixed "resume" channel instead of creating a new one each time.
</blockquote></details>

We're actually not blocking one thing at a time; the base compaction stays blocked throughout the test so we do need a separate channel for each blocked compaction.
___
*[`db_test.go` line 1398 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2508#-NU87bSJDeoYJzxIiJtv:-NU8Xvkn6kr8SxEBz2Bz:by03g6j) ([raw file](https://github.com/cockroachdb/pebble/blob/ff930a8664f8a5e1fd5a02ae705c384134e988d8/db_test.go#L1398)):*

Unless I'm missing something, it seems that this map never has more than 1 entry that is `true`, is this right? If yes, we can just make it a single `blockedJobID`
</blockquote></details>

Yep. Changed.
___
*[`db_test.go` line 1463 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2508#-NU88GXu2xibG0tTNOgf:-NU8Ykdc6HNsaj-GG-qm:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/ff930a8664f8a5e1fd5a02ae705c384134e988d8/db_test.go#L1463)):*

We should move this timeout to a constant (it's used multiple times), and make it bigger (we see even simpler stuff go over all the time, for reference CRDB's SucceedsSoon is 45s).
</blockquote></details>

Done.
___
*[`db_test.go` line 1475 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2508#-NU86t6YEefOGJwOrrXW:-NU8YlKt4FBPAcKLiHPM:bzdpnyz) ([raw file](https://github.com/cockroachdb/pebble/blob/ff930a8664f8a5e1fd5a02ae705c384134e988d8/db_test.go#L1475)):*

This is a very hot loop.. might make sense to add a short sleep.
</blockquote></details>

Added one, but the non-async Flush() should have already ensured it wouldn't be too CPU-hot.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2508#-:-NU8dywt1Bisae7r941i:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sumeerbhola)

itsbilal:
TFTR!

PR #2509, opened 2023-04-28
internal/manifest: verify incremental L0 sublevel output

### LINKED ISSUE #2475: internal/manifest: equivalence of incremental and non-incremental sublevels creation
There are two means through which L0 sublevels are created. One takes all of L0 files and builds the sublevels from scratch. The other builds off the existing sublevels incrementally. We should make these two methods always construct identical sublevels and add assertion invariants to the effect.

itsbilal:
This change verifies that the output of incrementally generated L0 sublevels matches that of a regenerated L0 sublevels, if invariants build flag is enabled.

Fixes #2475.

### DIFF
--- a/internal/manifest/version.go
+++ b/internal/manifest/version.go
@@ -1058,15 +1058,21 @@ func (v *Version) DebugString(format base.FormatKey) string {
 	return v.string(format, true)
 }
 
+func describeSublevels(format base.FormatKey, verbose bool, sublevels []LevelSlice) string {
+	var buf bytes.Buffer
+	for sublevel := len(sublevels) - 1; sublevel >= 0; sublevel-- {
+		fmt.Fprintf(&buf, "0.%d:\n", sublevel)
+		sublevels[sublevel].Each(func(f *FileMetadata) {
+			fmt.Fprintf(&buf, "  %s\n", f.DebugString(format, verbose))
+		})
+	}
+	return buf.String()
+}
+
 func (v *Version) string(format base.FormatKey, verbose bool) string {
 	var buf bytes.Buffer
 	if len(v.L0SublevelFiles) > 0 {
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2509#-:-NU8APV_1XDzVvZP3vkn:bnfp4nl)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`internal/manifest/version_edit.go` line 971 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2509#-NU8A7AXBwyqtQwCcJMP:-NU8A7AXBwyqtQwCcJMQ:b-eo34kb) ([raw file](https://github.com/cockroachdb/pebble/blob/0f694cabc2eb8d47e9bf7e5f0551054011545f0b/internal/manifest/version_edit.go#L971)):*
> ```Go
> 					err = v.InitL0Sublevels(cmp, formatKey, flushSplitBytes)
> 				} else if invariants.Enabled && err == nil {
> 					copyOfSublevels, err := NewL0Sublevels(&v.Levels[0], cmp, formatKey, flushSplitBytes)
> ```

Do you think this will slow down the tests significantly? Could you do a before/after `time make test`? If it does, we could do it randomly 1 in X times.

itsbilal:
Before:

```
go test ./... -run . -count 1 -tags 'invariants'  94.92s user 31.04s system 576% cpu 21.860 total
```

After:

```
go test ./... -run . -count 1 -tags 'invariants'  75.59s user 28.87s system 478% cpu 21.849 total
```

Re-ran both a bunch of times even after ensuring the test binary was already built and that we actually rerunning all tests, and it was hard to get the `after` state to be slower than `master`. Fair to say the time differential is negligible enough that per-run variance dominates.

TFTR!

PR #2522, opened 2023-05-11
db: avoid unnecessary lock reacquisition for paced file deletions

jbowens:
When doDeleteObsoleteFiles function deletes obsolete files directly, or if file-deletion pacing is enabled it launches a new goroutine to perform the paced deletion in the background. The function drops the DB.mu mutex while performing the deletions to avoid performing I/O while blocking other goroutines. However, in the case that pacing is enabled, there's no need to drop the mutex.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -3549,11 +3549,6 @@ func (d *DB) doDeleteObsoleteFiles(jobID int) {
 	obsoleteOptions := d.mu.versions.obsoleteOptions
 	d.mu.versions.obsoleteOptions = nil
 
-	// Release d.mu while doing I/O
-	// Note the unusual order: Unlock and then Lock.
-	d.mu.Unlock()
-	defer d.mu.Lock()
-
 	files := [4]struct {
 		fileType fileType
 		obsolete []fileInfo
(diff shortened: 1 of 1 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2522)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`compaction.go` line 3595 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2522#-NVC-DsP5MnjNfZTbAXn:-NVC-DsP5MnjNfZTbAXo:bnzi9qv) ([raw file](https://github.com/cockroachdb/pebble/blob/6b8f310b5f9d624323ede8fd4810957cc6e1021c/compaction.go#L3595)):*
> ```Go
> 			go d.paceAndDeleteObsoleteFiles(jobID, filesToDelete)
> 		} else {
> 			// Release d.mu while doing I/O
> ```

What's the point of keeping this alternate path (which is never relevant with CRDB)? Is it for determinism during testing?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2522)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`compaction.go` line 3595 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2522#-NVC-DsP5MnjNfZTbAXn:-NVFOalh6h4nHx4nn3hc:b-v3cn81) ([raw file](https://github.com/cockroachdb/pebble/blob/6b8f310b5f9d624323ede8fd4810957cc6e1021c/compaction.go#L3595)):*

What's the point of keeping this alternate path (which is never relevant with CRDB)? Is it for determinism during testing?
</blockquote></details>

yeah, we do rely on its determinism in tests.

PR #2523, opened 2023-05-11
db: don't delete files during ingest application

jbowens:
Previously, ingestions would trigger obsolete file deletions during the ingest application step. This has the potential to exacerbate the write "hiccup"/stall induced by sstable ingestion. The sequence numbers assigned to the ingestion are not published until after ingestion application. This prevents any concurrent writes that obtained later sequence numbers from publishing their own sequence numbers.

When Options.Experimental.MinDeletionRate is configured, this step to delete obsolete files should have minimal impact. However, when Experimental.MinDeletionRate is not configured, the ingesting goroutine would synchronously perform file deletions, extending the length of the commit stall.

Now ingestions don't trigger deletion of obsolete files. The only file deletion that an ingestion operation itself could trigger is deletion of a manifest that was made obsolete during a manifest rotation while committing the version edit. This is very rare, and delaying the deletion until a flush or compaction completes is okay.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -1108,7 +1108,6 @@ func (d *DB) ingestApply(
 	}
 	d.updateReadStateLocked(d.opts.DebugCheck)
 	d.updateTableStatsLocked(ve.NewFiles)
-	d.deleteObsoleteFiles(jobID, false /* waitForOngoing */)
 	// The ingestion may have pushed a level over the threshold for compaction,
 	// so check to see if one is necessary and schedule it.
 	d.maybeScheduleCompaction()
--- a/testdata/event_listener
+++ b/testdata/event_listener
@@ -215,8 +215,6 @@ close: db/marker.manifest.000006.MANIFEST-000016
 remove: db/marker.manifest.000005.MANIFEST-000014
 sync: db
 [JOB 12] MANIFEST created 000016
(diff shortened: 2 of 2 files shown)


bananabrick (approved):
:lgtm: Did we see this causing write stalls in practice?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2523#-:-NVBvmDS-6SVNzOyOhGt:b-c5oou0)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @itsbilal)

RaduBerinde (commented):
`MinDeletionRate` is always configured with CRDB (so I'm guessing no)

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2523#-:-NVBz2Nb1frqJyZ-mMj-:bd430pn)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @itsbilal)

jbowens (commented):
this was just from code inspection

TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2523#-:-NVGkepiEXimdtdf8Qjk:b-mw31s0)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @itsbilal)

PR #2524, opened 2023-05-12
shared: improve interface for more efficient reading

RaduBerinde:
The current `shared.Storage` interface forces the implementation to perform a metadata operation to look up the object on each read. Some experimental backends have a separate internal open step that doesn't need to be performed every time.

This commit improves the interface to allow opening the object once and then performing many reads.

### DIFF
--- a/objstorage/objstorageprovider/shared.go
+++ b/objstorage/objstorageprovider/shared.go
@@ -203,7 +203,7 @@ func (p *provider) sharedOpenForReading(
 		}
 	}
 	objName := sharedObjectName(meta)
-	size, err := p.sharedStorage().Size(objName)
+	reader, size, err := p.sharedStorage().ReadObject(ctx, objName)
 	if err != nil {
 		if opts.MustExist && p.sharedStorage().IsNotExistError(err) {
 			p.st.Logger.Fatalf("object %q does not exist", objName)
@@ -211,7 +211,7 @@ func (p *provider) sharedOpenForReading(
 		}
 		return nil, err
 	}
-	return newSharedReadable(p.sharedStorage(), objName, size), nil
+	return newSharedReadable(reader, size), nil
 }
 
 func (p *provider) sharedSize(meta objstorage.ObjectMetadata) (int64, error) {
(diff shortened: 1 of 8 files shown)


RaduBerinde:
Before merging I will prepare the master bump PR since it will require some changes in the external storage wrapper.

RaduBerinde:
TFTR!

sumeerbhola (commented):
> Reviewed 3 of 9 files at r1, 1 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2524)* status: 5 of 9 files reviewed, 1 unresolved discussion

___
*[`objstorage/shared/storage.go` line 70 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2524#-NVkXlUk8HhhMANa0VP3:-NVkXlUk8HhhMANa0VP4:b-ibbu5w) ([raw file](https://github.com/cockroachdb/pebble/blob/52aefa825ff9fd688a4bdda86380cb6221e8cabf/objstorage/shared/storage.go#L70)):*
> ```Go
> 	//
> 	// Clients of ReadAt can execute parallel ReadAt calls on the
> 	// same ObjectReader.
> ```

nit: maybe clarify that if `offset+len(p)` is past the end of the file, it returns an error without populating `p` (since it's a bit unusual, though totally makes sense for our needs).

RaduBerinde:
*[`objstorage/shared/storage.go` line 70 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2524#-NVkXlUk8HhhMANa0VP3:-NVkssdf5hT5U1aURJy2:by6c282) ([raw file](https://github.com/cockroachdb/pebble/blob/52aefa825ff9fd688a4bdda86380cb6221e8cabf/objstorage/shared/storage.go#L70)):*

nit: maybe clarify that if `offset+len(p)` is past the end of the file, it returns an error without populating `p` (since it's a bit unusual, though totally makes sense for our needs).
</blockquote></details>

Done (in #2535), also added to `Readable.ReadAt`.

sumeerbhola:
Thanks

PR #2525, opened 2023-05-12
crl-release-23.1: revert vfs: include size of write in DiskSlowInfo

jbowens:
This backport appears to be origin of recent disk-stall detection roachtest failures (eg, cockroachdb/cockroach#102946). I haven't identified what the exact bug is.


nicktrav (approved):
Thanks for digging into this. LGTM.

jbowens:
tftr!

PR #2529, opened 2023-05-16
objstorage: fix readheadState initialization

RaduBerinde:
When I extracted out the readahead logic from the sstable Reader, I accidentally omitted some initialization code - setting `readaheadState.sizs` to `initialReadaheadSize`. Fixing.

Informs #2531.


RaduBerinde:
@bilal, any suggestions about how we could test that readahead is working as expected, at the objstorage level?

The main difficulty is that we use the in-mem FS for tests, where the readahead code paths are not exercised. We could use a tmp filesystem, but I think it would still be arch dependent. Perhaps we should replace the `Fd()` check with some new capabilities functionality and have the in-mem files pretend that they `Prealloc()`?

RaduBerinde:
I opened separate #2532 that adds a test. It changes a few things so I don't think we want to backport that as well.

jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2529#-:-NVaDpNBDSjD5928HOih:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`-- commits` line 6 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2529#-NVaDlsUB0yS5uKdwrRC):*
nit: s/sizs/size/

itsbilal (approved):
LGTM, thanks for catching this!

As for testing ideas, one could be to write a test with a wrapper vfs.FS that stubs the `file.Prefetch` method to record the number of calls made to it in some global atomic. It could then 1) open Pebble and Set a whole bunch of keys, 2) maybe do a manual compaction, 3) do a sequential iteration from First() onwards, 4) assert that Prefetch() was called at least once. That should cover stuff like this, right?

RaduBerinde:
Merged as part of #2532 (sorry, thought I had merged this already).

Issue #2531, opened 2023-05-16
objstorage: read ahead is broken


RaduBerinde:
The `readaheadState` in the vfs read handle is not being initialized properly, resulting in no "automatic" read-ahead functionality (read-ahead for compactions still works).

nicktrav:
@jbowens @sumeerbhola - I recall we had a perf issue ~recently and we suspected readahead. Could this be related?

Or is this only due to recent changes to VFS related to objstorage refactorings, etc.

jbowens:
If we're talking about the same issue, it was on 22.2 and we ended up suspecting that it was an issue of the OS page cache being trashed instead of loss of readahead.

PR #2532, opened 2023-05-16
objstorage: improve test and add read ahead test

RaduBerinde:
First commit is #2529.

#### objstorage: remove genericFileReadable

Remove the `genericFileReadable` implementation; since we moved
`Prealloc` to be a method on `File`, the regular implementation should
always work. Using the same implementation for all cases helps testing
- e.g. we can test readahead logic with files in MemFS.

#### objstorage: improve test and add read ahead test

This change improves the objstorage test to allow reading arbitrary
parts of objects. We add a test that verifies that the read-ahead
implementation does the Prefetch and the reopen that is expected.

We also improve logging FS to log `ReadAt` and `Prefetch` calls.

Informs #2531.


RaduBerinde:
*[`objstorage/objstorageprovider/testdata/provider/local_readahead` line 37 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2532#-NVa5uhG6oiKz3qULRyS:-NVa5uhG6oiKz3qULRyT:b-khassn) ([raw file](https://github.com/cockroachdb/pebble/blob/793d6cb4b0419e9824dc2af3e3b55e13fae04fd2/objstorage/objstorageprovider/testdata/provider/local_readahead#L37)):*
> ```
> <local fs> read-at(1000, 15000): p1/000001.sst
> 1000 15000: ok (salt 1)
> <local fs> prefetch(16000, 65536): p1/000001.sst
> ```

I confirmed that without #2529, we don't see these prefetch calls (or the open below).

itsbilal (approved):
LGTM

Issue #2465, opened 2023-04-19
sstable: sstable-local key kinds for overwritten/deleted points


sumeerbhola:
We have various related problems caused by Pebble snapshots:
- P1. RANGEDELs that delete points in the same sstable, but the points were not deleted during the compaction because of an open snapshot. This results in expensive iteration. https://github.com/cockroachdb/pebble/issues/2424 discusses a possible solution using a block property collector for the range of seqnums in a block. It involves using a dynamic block property filter, based on the RANGEDELs at the current point iterator position, to skip blocks. A concern there is the code complexity of this dynamic application (we have significant code complexity in the other case of dynamic application -- range key masking).

- P2. When iterating over foreign sstables, we need to do (a) point collapsing to expose at most one point per user key, (b) apply RANGEDELs. The current PR for this https://github.com/cockroachdb/pebble/pull/2455 involves a `pointCollapsingSSTIter` that wraps the actual sst iterator. SST iteration is very performance sensitive code and adding extra wrappers is non ideal: there is no clear end to the life of a foreign sst and we'd rather not force a rewrite of that sstable -- one of the goals of disaggregated storage is to scale compute and disk bandwidth resources as a function of the hot (from a write perspective) data and not the whole data. Also, application of RANGEDELs in this case suffers from the same issue as in P1.

The ideal solution for P2 would allow user-facing reads to utilize the existing SST iterators (with slight modifications) and with no loss of efficiency. And for P1 and P2 we would like to skip whole blocks of overwritten/deleted points. And avoiding key comparisons would be very useful in general.

We observe that:
- Reads:
  - All user-facing reads in CockroachDB use iterators over the DB, hence have a higher read seqnum than all sstables
  - Even for the internal-facing reads that use snapshots, the snapshots are shortlived enough that most L5 and L6 sstables will have all seqnums lower than the snapshot seqnum.
- Writes:
  - we have already done the key comparisons between points that share the same user key and between points and RANGEDELs on the compaction path (in `compactionIter`). The idea below uses those comparisons, by encoding them in the written sst, so they can be cheaply utilized when reading.

We ignore MERGE in the following discussion since I don't believe it is worth optimizing:
- For P2, we should disallow it from foreign ssts.
- For regular ssts, we don't need to optimize for it. They can be handled by the upstream iterators in the iterator tree (specifically `pebble.Iterator`).

[...]

PR #2537, opened 2023-05-17
db: add DeleteSized and corresponding compaction heuristics

### LINKED ISSUE #2340: db: optional point tombstone deletion size estimates

jbowens:
Add a new write operation and corresponding internal key kind for delete operations that know the size of the value they're deleting. SSTable writers aggregate these estimates into table properties that are then used to improve compaction-picking heuristics. This is particularly useful in scenarios with heterogeneous value sizes that are easily misestimated by existing heuristics that rely on averages.

When a DELSIZED tombstone deletes a key, it's mutated to reflect that the deletion has already occurred. If the size carried within the tombstone's value was accurate, the tombstone's value is removed. If the size was not accurate, the tombstone is mutated into an ordinary DEL tombstone which will be subject to typical average-based heuristics.

Close #2340.


sumeerbhola (changes requested):
Flushing some comments. Looks good!
Haven't read table_stats.go or the tests yet.

> Reviewed 17 of 42 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2537#-:-NVksDba2u_p0eOrHlVG:brdcrzj)* status: 17 of 42 files reviewed, 11 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`batch.go` line 423 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkfaooEeHA-Fqc1ckV:-NVkfaooEeHA-Fqc1ckW:bf9kdum) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/batch.go#L423)):*
> ```Go
> 	b.countRangeDels = 0
> 	b.countRangeKeys = 0
> 	b.minimumFormatMajorVersion = 0x00
> ```

nit: why not 0, given it is uint64?
___
*[`batch.go` line 434 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkfoed7StCcqiazzcO:-NVkfoed7StCcqiazzcP:bzslbj) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/batch.go#L434)):*
> ```Go
> 		case InternalKeyKindRangeKeySet, InternalKeyKindRangeKeyUnset, InternalKeyKindRangeKeyDelete:
> 			b.countRangeKeys++
> 		case InternalKeyKindDeleteSized:
> ```

nice to see this improved compatibility check!
___
*[`batch.go` line 708 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkfxe-8fEjGKJRqnhh:-NVkfxe-8fEjGKJRqnhi:br5q9sy) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/batch.go#L708)):*
> ```Go
> //
> // DeleteSized will record the value size within the tombstone and use it
> // inform compaction-picking heuristics which strive to reduce space
> ```

nit: to inform ...
___
*[`compaction_iter.go` line 851 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVknqWY2DxkoHVfFzUH:-NVknqWY2DxkoHVfFzUI:b-ypbtm6) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/compaction_iter.go#L851)):*
> ```Go
> 	// In this case, we still peek forward in case there's another DELSIZED key
> 	// with a lower sequence number, in which case we'll adopt its value.
> 	if len(i.value) == 0 {
> ```

when did we set `i.value = i.iterValue`. Was this condition meant to be `if len(i.iterValue) == 0`?
___
*[`compaction_iter.go` line 863 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkopwEA2tLgXAZEne1:-NVkopwEA2tLgXAZEne2:b2bldum) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/compaction_iter.go#L863)):*
> ```Go
> 	var elidedSize uint64
> 	for i.nextInStripe() == sameStripeSkippable {
> 		elidedSize += uint64(len(i.iterKey.UserKey)) + uint64(len(i.iterValue))
> ```

So a.DELSIZED.9 a.DEL.8 a.SET.7, will result in elidedSize including both the DEL and SET, yes?
___
*[`compaction_iter.go` line 883 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkrVRE1motvfr7N-OA:-NVkrVRE1motvfr7N-OB:bikypaf) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/compaction_iter.go#L883)):*
> ```Go
> 	}
> 	if i.iterStripeChange == newStripeNewKey || i.iterStripeChange == newStripeSameKey {
> 		i.skip = false
> ```

can you copy the following comment here
```
	// Reset skip if we landed outside the original stripe. Otherwise, we landed
	// in the same stripe on a non-skippable key. In that case we should preserve
	// `i.skip == true` such that later keys in the stripe will continue to be
	// skipped.
```
___
*[`level_checker.go` line 168 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkepZ47hdFzvseP3z-:-NVkepZ47hdFzvseP3z0:b-8z4j5x) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/level_checker.go#L168)):*
> ```Go
> 			// Ongoing series of MERGE records.
> 			switch item.key.Kind() {
> 			case InternalKeyKindSingleDelete, InternalKeyKindDelete, InternalKeyKindDeleteSized:
> ```

I assume you checked all places we refer to `InternalKeyKindDelete`. I didn't try to verify the completeness of the changes.
___
*[`metamorphic/ops.go` line 612 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkZhyNEpsO22zo5lVj:-NVkZhyNEpsO22zo5lVk:bawemqx) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/metamorphic/ops.go#L612)):*
> ```Go
> 			case pebble.InternalKeyKindDeleteSized:
> 				v, _ := binary.Uvarint(value.InPlaceValue())
> 				err = collapsed.DeleteSized(key.UserKey, uint32(v-uint64(len(key.UserKey))), nil)
> ```

why this subtraction of user key length?
Ah, I see why now -- can you add a code comment here.
___
*[`replay/replay.go` line 1042 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkZ7GL-4IyPUtcruju:-NVkZ7GL-4IyPUtcrujv:b-9yv5b6) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/replay/replay.go#L1042)):*
> ```Go
> 		case base.InternalKeyKindDeleteSized:
> 			v, _ := binary.Uvarint(bufs.keys[i].value)
> 			err = b.DeleteSized(bufs.keys[i].UserKey, uint32(v-uint64(len(bufs.keys[i].UserKey))), nil)
> ```

This subtraction also needs a code comment
___
*[`sstable/properties.go` line 362 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkRwGM20NKQw8UYVYL:-NVkRwGN3CY3EPjzCIDs:bcb2lpf) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/sstable/properties.go#L362)):*
> ```Go
> 	// user-defined properties, however writing them prevents byte-for-byte
> 	// equivalence with RocksDB files that some of our testing requires.
> 	if p.NumSizedDeletions > 0 && tblFormat >= TableFormatPebblev1 {
> ```

nit: doesn't `p.NumSizedDeletions > 0` imply `tblFormat >= TableFormatPebblev4`?
___
*[`sstable/properties.go` line 370 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkYIKKAQ88Z4oa4djV:-NVkYIKKAQ88Z4oa4djW:b-7ropln) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/sstable/properties.go#L370)):*
> ```Go
> 		p.saveUvarint(m, unsafe.Offsetof(p.RawPointTombstoneKeySize), p.RawPointTombstoneKeySize)
> 	}
> 	if p.RawPointTombstoneValueSizeHint > 0 && tblFormat >= TableFormatPebblev1 {
> ```

ditto

sumeerbhola (changes requested):
> Reviewed 20 of 42 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2537)* status: 37 of 42 files reviewed, 15 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`table_stats.go` line 429 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVlUsCv3ynjE-v57bH4:-NVlUsCv3ynjE-v57bH5:be6k4wp) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/table_stats.go#L429)):*
> ```Go
> 	// summing their value sizes and entry counts.
> 	var fileSum, keySum, valSum, entryCount uint64
> 	// Include the file itself.
> ```

was not including the file itself an oversight before, or is there some subtlety that I am missing?
___
*[`table_stats.go` line 433 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVlUaey2sZT9d5I-hgi:-NVlUaey2sZT9d5I-hgj:bj6lhdm) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/table_stats.go#L433)):*
> ```Go
> 		meta,
> 		func(r *sstable.Reader) (err error) {
> 			fileSum += meta.Size
> ```

is this the same func as the one below? If yes, can it be defined once?
___
*[`table_stats.go` line 439 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVlUkPr0Z8sQPlAbUoL:-NVlUkPr0Z8sQPlAbUoM:b-el5ml4) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/table_stats.go#L439)):*
> ```Go
> 			return nil
> 		}); err != nil {
> 		return avgValueLogicalSize, compressionRatio, err
> ```

why is this not returning 0, 0, err like the code below?
___
*[`table_stats.go` line 717 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVlXjbF88zJnTUZwgIX:-NVlXjbG9zzdNEvkunzB:b-pddn6y) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/table_stats.go#L717)):*
> ```Go
> 		// many sstables).
> 		shadowedLogicalSize += float64(numUnsizedDels) * avgValLogicalSize
> 	}
> ```

The comments here are great and make it easy to follow along.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2537)* status: 37 of 42 files reviewed, 13 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`batch.go` line 423 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkfaooEeHA-Fqc1ckV:-NVog1UQADOc_uBu7ZoW:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/batch.go#L423)):*

nit: why not 0, given it is uint64?
</blockquote></details>

Done.
___
*[`batch.go` line 708 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkfxe-8fEjGKJRqnhh:-NVogMIv4DtTVXHPrRih:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/batch.go#L708)):*

nit: to inform ...
</blockquote></details>

Done.
___
*[`compaction_iter.go` line 851 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVknqWY2DxkoHVfFzUH:-NVoh87o4mKrKiHTFkDQ:b-sx0b7w) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/compaction_iter.go#L851)):*

when did we set `i.value = i.iterValue`. Was this condition meant to be `if len(i.iterValue) == 0`?
</blockquote></details>

Thanks, good catch.
___
*[`compaction_iter.go` line 863 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkopwEA2tLgXAZEne1:-NVpB0c26oYdH4ed7k9V:b-890fet) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/compaction_iter.go#L863)):*

So a.DELSIZED.9 a.DEL.8 a.SET.7, will result in elidedSize including both the DEL and SET, yes?
</blockquote></details>

yeah, that's right
___
*[`compaction_iter.go` line 883 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkrVRE1motvfr7N-OA:-NVogx2FAIoclCQbYoOx:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/compaction_iter.go#L883)):*

can you copy the following comment here
```
	// Reset skip if we landed outside the original stripe. Otherwise, we landed
	// in the same stripe on a non-skippable key. In that case we should preserve
	// `i.skip == true` such that later keys in the stripe will continue to be
	// skipped.
```
</blockquote></details>

Done.
___
*[`level_checker.go` line 168 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkepZ47hdFzvseP3z-:-NVoi5Q4B6XiH53TAEbJ:b-uqwqzo) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/level_checker.go#L168)):*

I assume you checked all places we refer to `InternalKeyKindDelete`. I didn't try to verify the completeness of the changes.
</blockquote></details>

Yeah, I did
___
*[`table_stats.go` line 429 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVlUsCv3ynjE-v57bH4:-NVpCo80D1q9rnXelpXG:b-f4qqq) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/table_stats.go#L429)):*

was not including the file itself an oversight before, or is there some subtlety that I am missing?
</blockquote></details>

there's a bit of a subtlety motivating this. the pointDeletionBytesEstimate calculation previously used the containing file's own compression ratio to scale the tombstones to physical sizes. I thought the code was a little excessively complicated with two compression ratios, since I refactored it to use just the one: either derived from the file itself or the one here derived from the LSM.

It introduced a problem in that there might be zero files lower in the LSM, in which case the compression ratio is zero and the tombstones' own contribution is zeroed.
___
*[`table_stats.go` line 433 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVlUaey2sZT9d5I-hgi:-NVpClyX3K3ViDBMu2F2:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/table_stats.go#L433)):*

is this the same func as the one below? If yes, can it be defined once?
</blockquote></details>

Done.
___
*[`table_stats.go` line 439 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVlUkPr0Z8sQPlAbUoL:-NVpCmQa4ryXGxXHs-iu:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/table_stats.go#L439)):*

why is this not returning 0, 0, err like the code below?
</blockquote></details>

Done.
___
*[`metamorphic/ops.go` line 612 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkZhyNEpsO22zo5lVj:-NVpAXkm0sbouu0tyg5H:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/metamorphic/ops.go#L612)):*

why this subtraction of user key length?
Ah, I see why now -- can you add a code comment here.
</blockquote></details>

Done.
___
*[`replay/replay.go` line 1042 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkZ7GL-4IyPUtcruju:-NVpAhD796ey6OQgHGz1:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/replay/replay.go#L1042)):*

This subtraction also needs a code comment
</blockquote></details>

Done.
___
*[`sstable/properties.go` line 362 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkRwGM20NKQw8UYVYL:-NVpArqb9SJmOLwHZ1uf:bb7lbs2) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/sstable/properties.go#L362)):*

nit: doesn't `p.NumSizedDeletions > 0` imply `tblFormat >= TableFormatPebblev4`?
</blockquote></details>

nice
___
*[`sstable/properties.go` line 370 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NVkYIKKAQ88Z4oa4djV:-NVpAxha32YlRjo_7ROX:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9fc4bedf7dbc50d5dfbfadb787008f72a63e12eb/sstable/properties.go#L370)):*

ditto
</blockquote></details>

Done.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2537)* status: 29 of 48 files reviewed, 14 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/format.go` line 27 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NW2SVdLA9PJ9li-AEBd:-NW2SVdLA9PJ9li-AEBe:b-c68nfd) ([raw file](https://github.com/cockroachdb/pebble/blob/99601ee637617269b001111d1f6e778e13585b1c/sstable/format.go#L27)):*
> ```Go
> 	TableFormatPebblev2 // Range keys.
> 	TableFormatPebblev3 // Value blocks.
> 	TableFormatPebblev4 // DELSIZED tombstones.
> ```

What do you think about rolling more changes into v4, like https://github.com/cockroachdb/pebble/issues/2465 -- I may have something ready for review this week?
The mental overhead of many sstable versions seems high to me, so we could try to keep it to roughly 1 per CockroachDB release.

sumeerbhola (approved):
:lgtm: 

> Reviewed 5 of 42 files at r1, 14 of 14 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2537#-:-NW49KIFBeh23cVH6xdD:bg7igkv)* status: all files reviewed, 14 unresolved discussions (waiting on @jbowens and @RaduBerinde)

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2537)* status: all files reviewed, 14 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`sstable/format.go` line 27 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2537#-NW2SVdLA9PJ9li-AEBd:-NW87yHGEHG30iTxgZ80:b27bc81) ([raw file](https://github.com/cockroachdb/pebble/blob/99601ee637617269b001111d1f6e778e13585b1c/sstable/format.go#L27)):*

What do you think about rolling more changes into v4, like https://github.com/cockroachdb/pebble/issues/2465 -- I may have something ready for review this week?
The mental overhead of many sstable versions seems high to me, so we could try to keep it to roughly 1 per CockroachDB release.
</blockquote></details>

Yeah, that sounds reasonable to me. What do you think of merging this with an "Experimental" label on the format major version? We could leave `FormatNewest` pointing to the preceding format major version.

I don't think we really make any guarantees of stability of the master branch, but I don't think that's very well communicated.


On second thought... for `DELSIZED` we only really need a new sstable version so that during `Ingest` we can check that `DELSIZED` keys aren't ingested before the format major version is adopted. We alternatively could check the new `pebble.num.deletions.sized` sstable property, but I think it's nice to limit compatibility checks to a few standardized places. Maybe we should introduce a mechanism for associating minimum format major versions with sstables without introducing new table formats. This could be a new property or a new byte in the sstable footer. It could encode the minimum format major version required by the sstable, or a bit field indicating which sstable "features" are in use, which then can be mapped to format major versions.

sumeerbhola:
> What do you think of merging this with an "Experimental" label on the format major version? We could leave FormatNewest pointing to the preceding format major version.

Sounds good.

> We alternatively could check the new pebble.num.deletions.sized sstable property, but I think it's nice to limit compatibility checks to a few standardized places.

I think it's simpler to do this with an sstable version, as you have done here.

jbowens:
TFTR!

PR #2539, opened 2023-05-18
secondarycache: implement a minimal functional secondary cache

joshimhoff:

First two commit are authored by @RaduBerinde.

**vfs: support WriteAt**

Add support for opening files in read/write mode and writing at
arbitrary offsets.

**secondarycache: implement a minimal skeleton**

This commit implements a minimal skeleton for the secondary cache. As of this
commit, it is non-functional.

**secondarycache: implement a minimal functional secondary cache**

This commit expands the skeleton added in the last commit to be functional.
This commit also adds basic tests. Concretely, as of this commit, the secondary
cache implements storing cached data on disk, sharding, configurable cache
block sizes, a random eviction scheme, and in-memory metadata (implying the
overall secondary cache empties at restart time).

**vfs: make in-mem vfs satisfy the vfs contract**

This commit updates the in-memory vfs to satisy the vfs contract. Concretely,
OpenReadWrite now opens a file for reading & writing, creating the file if it
does not yet exist, and Create creates a file that can be read, not just
written to.

### DIFF
--- a/internal/errorfs/errorfs.go
+++ b/internal/errorfs/errorfs.go
@@ -61,6 +61,8 @@ const (
 	OpFileReadAt
 	// OpFileWrite describes a file write operation.
 	OpFileWrite
+	// OpFileWriteAt describes a file seek write operation.
+	OpFileWriteAt
 	// OpFileStat describes a file stat operation.
 	OpFileStat
 	// OpFileSync describes a file sync operation.
@@ -74,7 +76,7 @@ func (o Op) OpKind() OpKind {
 	switch o {
 	case OpOpen, OpOpenDir, OpList, OpStat, OpGetDiskUsage, OpFileRead, OpFileReadAt, OpFileStat:
 		return OpKindRead
-	case OpCreate, OpLink, OpRemove, OpRemoveAll, OpRename, OpReuseForRewrite, OpMkdirAll, OpLock, OpFileClose, OpFileWrite, OpFileSync, OpFileFlush, OpFilePreallocate:
+	case OpCreate, OpLink, OpRemove, OpRemoveAll, OpRename, OpReuseForRewrite, OpMkdirAll, OpLock, OpFileClose, OpFileWrite, OpFileWriteAt, OpFileSync, OpFileFlush, OpFilePreallocate:
 		return OpKindWrite
 	default:
 		panic(fmt.Sprintf("unrecognized op %v\n", o))
@@ -206,6 +208,22 @@ func (fs *FS) Link(oldname, newname string) error {
 
 // Open implements FS.Open.
 func (fs *FS) Open(name string, opts ...vfs.OpenOption) (vfs.File, error) {
+	if err := fs.inj.MaybeError(OpOpen, name); err != nil {
+		return nil, err
+	}
+	f, err := fs.fs.Open(name)
+	if err != nil {
+		return nil, err
+	}
+	ef := &errorFile{name, f, fs.inj}
+	for _, opt := range opts {
+		opt.Apply(ef)
+	}
+	return ef, nil
+}
+
+// OpenReadWrite implements FS.OpenReadWrite.
+func (fs *FS) OpenReadWrite(name string, opts ...vfs.OpenOption) (vfs.File, error) {
 	if err := fs.inj.MaybeError(OpOpen, name); err != nil {
 		return nil, err
 	}
@@ -356,6 +374,13 @@ func (f *errorFile) Write(p []byte) (int, error) {
 		return 0, err
 	}
 	return f.file.Write(p)
+}
+
+func (f *errorFile) WriteAt(p []byte, ofs int64) (int, error) {
+	if err := f.inj.MaybeError(OpFileWriteAt, f.path); err != nil {
+		return 0, err
+	}
+	return f.file.WriteAt(p, ofs)
 }
 
 func (f *errorFile) Stat() (os.FileInfo, error) {
--- a/objstorage/objstorageprovider/provider.go
+++ b/objstorage/objstorageprovider/provider.go
@@ -93,6 +93,16 @@ type Settings struct {
 	// (experimental).
 	Shared struct {
 		Storage shared.Storage
+
+		// CacheSizeBytes is the size of the on-disk block cache for objects
+		// on shared storage. If it is 0, no cache is used.
+		CacheSizeBytes int64
+
+		// CacheBlockSize is the block size of the cache; if 0, the default of 32KB is used.
+		CacheBlockSize int
+
+		// TODO(radu): allow the cache to live on another FS/location (e.g. to use
+		// instance-local SSD).
 	}
 }
 
--- a/objstorage/objstorageprovider/provider.go
+++ b/objstorage/objstorageprovider/provider.go
@@ -101,6 +101,11 @@ type Settings struct {
 		// CacheBlockSize is the block size of the cache; if 0, the default of 32KB is used.
 		CacheBlockSize int
 
+		// The number of independent shards the cache leverages. Each shard is the same size,
+		// and a hash of filenum & offset map a read to a certain shard. If set to 0,
+		// 2*runtime.GOMAXPROCS is used as the shard count.
+		CacheShardCount int
+
 		// TODO(radu): allow the cache to live on another FS/location (e.g. to use
 		// instance-local SSD).
 	}
--- a/objstorage/objstorageprovider/shared.go
+++ b/objstorage/objstorageprovider/shared.go
@@ -6,6 +6,7 @@ package objstorageprovider
 
 import (
 	"context"
+	"runtime"
 	"sync"
 	"sync/atomic"
 
@@ -72,7 +73,8 @@ func (p *provider) sharedInit() error {
 		if blockSize == 0 {
 			blockSize = defaultBlockSize
 		}
-		p.shared.cache, err = openSharedCache(p.st.FS, p.st.FSDirName, blockSize, p.st.Shared.CacheSizeBytes)
+		numShards := 2 * runtime.GOMAXPROCS(0)
+		p.shared.cache, err = openSharedCache(p.st.FS, p.st.FSDirName, blockSize, p.st.Shared.CacheSizeBytes, numShards)
 		if err != nil {
 			return errors.Wrapf(err, "pebble: could not open shared object cache")
 		}
--- a/objstorage/objstorageprovider/shared.go
+++ b/objstorage/objstorageprovider/shared.go
@@ -73,7 +73,12 @@ func (p *provider) sharedInit() error {
 		if blockSize == 0 {
 			blockSize = defaultBlockSize
 		}
-		numShards := 2 * runtime.GOMAXPROCS(0)
+
+		numShards := p.st.Shared.CacheShardCount
+		if numShards == 0 {
+			numShards = 2 * runtime.GOMAXPROCS(0)
+		}
+
 		p.shared.cache, err = openSharedCache(p.st.FS, p.st.FSDirName, blockSize, p.st.Shared.CacheSizeBytes, numShards)
 		if err != nil {
 			return errors.Wrapf(err, "pebble: could not open shared object cache")
--- a/objstorage/objstorageprovider/shared.go
+++ b/objstorage/objstorageprovider/shared.go
@@ -21,6 +21,7 @@ import (
 // All fields remain unset if shared storage is not configured.
 type sharedSubsystem struct {
 	catalog *sharedobjcat.Catalog
+	cache   *sharedCache
 
 	// checkRefsOnOpen controls whether we check the ref marker file when opening
 	// an object. Normally this is true when invariants are enabled (but the provider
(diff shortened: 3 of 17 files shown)


RaduBerinde (commented):
Great progress!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2539#-:-NVjrWUyB13qkoe3Bx97:b-nnl4p0)* status: 0 of 10 files reviewed, 6 unresolved discussions (waiting on @itsbilal and @joshimhoff)

___
*[`objstorage/objstorageprovider/shared_cache.go` line 33 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjpN5VDEbhO0jiZPd9:-NVjpN5VDEbhO0jiZPdA:ba699s8) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache.go#L33)):*
> ```Go
> 	fs vfs.FS, fsDir string, blockSize int, sizeBytes int64,
> ) (*sharedCache, error) {
> 	// TODO(josh): Vertical scale ops can be done. IIUC, the same sstable reads won't
> ```

Can you explain this more? I don't understand the issue
___
*[`objstorage/objstorageprovider/shared_cache.go` line 36 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjpWip7e9K72TTczLH:-NVjpWip7e9K72TTczLI:bb4xeak) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache.go#L36)):*
> ```Go
> 	// map to the same shards necessarily, emptying the cache in effect. We may need
> 	// a solution to this issue eventually.
> 	numShards := 2 * runtime.GOMAXPROCS(0)
> ```

I think we should make this configurable and use a fixed value for tests
___
*[`objstorage/objstorageprovider/shared_cache.go` line 82 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjrPBQ6GEcp3YIsmSV:-NVjrPBQ6GEcp3YIsmSW:b-cyw5n7) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache.go#L82)):*
> ```Go
> 		return err
> 	}
> 	// TODO(josh): Currently, we do not read in increments of cache block size above,
> ```

Yeah we really need to round up and read entire blocks.
___
*[`objstorage/objstorageprovider/shared_cache.go` line 131 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjpljR8bck2UZVYvz7:-NVjpljR8bck2UZVYvz8:bx4z3qj) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache.go#L131)):*
> ```Go
> //
> // If all of p is not written to the shard, Set returns a non-nil error.
> func (sc *sharedCache) Set(fileNum base.FileNum, p []byte, ofs int64) error {
> ```

This really only makes sense when `ofs` and `len(p)` are multiples of the block size. Maybe leave a TODO to assert that
___
*[`objstorage/objstorageprovider/shared_cache.go` line 179 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjpu-Z3p6z71yz213p:-NVjpu-Z3p6z71yz213q:b-40fm9h) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache.go#L179)):*
> ```Go
> type metadataKey struct {
> 	filenum base.FileNum
> 	offset  int64
> ```

This is not an offset, it's a block index
___
*[`objstorage/objstorageprovider/shared_cache_test.go` line 74 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjqlSRCgQ57Oar_39D:-NVjqlSRCgQ57Oar_39E:b11875a) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache_test.go#L74)):*
> ```Go
> 	})
> 
> 	t.Run("small read with write to cache at offset also", func(t *testing.T) {
> ```

These would be a whole lot easier as provider data driven tests, check out existing uses of `read`, we can control exactly the ReadAt calls and the test outputs everything that happens on the filesystem.

RaduBerinde:
*[`objstorage/objstorageprovider/shared_cache.go` line 33 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjpN5VDEbhO0jiZPd9:-NVjsGcy-DUMAlmRz3TO:b-2ix9il) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache.go#L33)):*

Can you explain this more? I don't understand the issue
</blockquote></details>

Oh, you are talking about GOMAXPROCS changing? Well we only read it once at initialization.. I guess it may be an issue when we persist the cache (though we could keep the same shards in principle)

joshimhoff (commented):
PTAL!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2539#-:-NW2lDXOBjTfFRq6SMSo:b-83330z)* status: 0 of 17 files reviewed, 5 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`objstorage/objstorageprovider/shared_cache.go` line 33 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjpN5VDEbhO0jiZPd9:-NVploLiDiX8KG1w1KrP:bpjd5fd) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache.go#L33)):*

Oh, you are talking about GOMAXPROCS changing? Well we only read it once at initialization.. I guess it may be an issue when we persist the cache (though we could keep the same shards in principle)
</blockquote></details>

Yup, I am pointing out the issue of `GOMAXPROCS` changing leading to a new mapping of <filenum, offset> pairs to shards. This will empty the cache out in effect. I have tried to make the TODO clearer.
___
*[`objstorage/objstorageprovider/shared_cache.go` line 36 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjpWip7e9K72TTczLH:-NVpnuPcEAtI6LLSX-bc:bcvjgzp) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache.go#L36)):*

I think we should make this configurable and use a fixed value for tests
</blockquote></details>

Good idea. Done.
___
*[`objstorage/objstorageprovider/shared_cache.go` line 82 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjrPBQ6GEcp3YIsmSV:-NVq-IcKEssiVh8dvwW1:b-2ewtlx) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache.go#L82)):*

Yeah we really need to round up and read entire blocks.
</blockquote></details>

I've implemented this now. There is one edge, which I've added a TODO to address in a follow up PR:

// TODO(josh): To have proper EOF handling, we will need readable.ReadAt to return
// the number of bytes read successfully. As is, we cannot tell if the readable.ReadAt
// should be returned from sharedCache.ReadAt. For now, the cache just swallows all
// io.EOF errors.
___
*[`objstorage/objstorageprovider/shared_cache.go` line 131 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjpljR8bck2UZVYvz7:-NVpppBR3Zggt84Yh6i-:b-j5o5cc) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache.go#L131)):*

This really only makes sense when `ofs` and `len(p)` are multiples of the block size. Maybe leave a TODO to assert that
</blockquote></details>

Agreed & good idea re: assertion. Done.
___
*[`objstorage/objstorageprovider/shared_cache.go` line 179 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjpu-Z3p6z71yz213p:-NVpoii6Ehn1ZCq9rGl1:bl8lja6) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache.go#L179)):*

This is not an offset, it's a block index
</blockquote></details>

Good point. Fixed.
___
*[`objstorage/objstorageprovider/shared_cache_test.go` line 74 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjqlSRCgQ57Oar_39D:-NW2dXG_--KjPO2zZihE:bdf5khi) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache_test.go#L74)):*

These would be a whole lot easier as provider data driven tests, check out existing uses of `read`, we can control exactly the ReadAt calls and the test outputs everything that happens on the filesystem.
</blockquote></details>

Done. I haven't used that test framework before. Thanks for the suggestion!

I notice that `logging_fs.go` doesn't trace calls to ReadAt or WriteAt, which are the calls of interest made by the secondary cache. Should I change that, or is there some reason that `logging_fs.go` doesn't trace those calls? I can also add an option, if we just want to trace those calls from the tests I've added. PR sorta large already, so I'd push tracing fs activity to a follow up PR, if okay with you.

RaduBerinde (approved):
:lgtm:

Nice work!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2539#-:-NW3JCkn-6muufgkR1LS:b-8cucv6)* status: 0 of 17 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @joshimhoff)

___
*[`objstorage/objstorageprovider/shared_cache.go` line 36 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjpWip7e9K72TTczLH:-NW3GiIH-eqyqDz8P5IX:b-4n0d3p) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache.go#L36)):*

Good idea. Done.
</blockquote></details>

I was thinking it would be a true option like `CacheBlockSize`
___
*[`objstorage/objstorageprovider/shared_cache.go` line 82 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjrPBQ6GEcp3YIsmSV:-NW3Gxq69TaolGoHRKUy:b-j3lqqr) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache.go#L82)):*

I've implemented this now. There is one edge, which I've added a TODO to address in a follow up PR:

// TODO(josh): To have proper EOF handling, we will need readable.ReadAt to return
// the number of bytes read successfully. As is, we cannot tell if the readable.ReadAt
// should be returned from sharedCache.ReadAt. For now, the cache just swallows all
// io.EOF errors.
</blockquote></details>

`Readable` has a very cheap `Size()` method that we can use to cap the length that we read.
___
*[`objstorage/objstorageprovider/shared_cache.go` line 171 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NW3HTmt61LvSuzxbsQG:-NW3HTmt61LvSuzxbsQH:bykehu5) ([raw file](https://github.com/cockroachdb/pebble/blob/e298392f459f53ddd7be6b985946899e67b45ee3/objstorage/objstorageprovider/shared_cache.go#L171)):*
> ```Go
> 	// updates len(sc.shards); see openSharedCache. As a result, the behavior of this
> 	// function changes, and the cache empties out at restart time. We may want a better
> 	// story here eventually.
> ```

When the cache will become persistent, we would store the number of shards in the cache metadata and we would use the same number of shards even if GOMAXPROCS changes.
___
*[`objstorage/objstorageprovider/shared_cache_test.go` line 74 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2539#-NVjqlSRCgQ57Oar_39D:-NW3If_D4Vxj5CowdhAW:bw2ytud) ([raw file](https://github.com/cockroachdb/pebble/blob/e3c246c0266e6b41e3aa387355c8c5fb8a12270d/objstorage/objstorageprovider/shared_cache_test.go#L74)):*

Done. I haven't used that test framework before. Thanks for the suggestion!

I notice that `logging_fs.go` doesn't trace calls to ReadAt or WriteAt, which are the calls of interest made by the secondary cache. Should I change that, or is there some reason that `logging_fs.go` doesn't trace those calls? I can also add an option, if we just want to trace those calls from the tests I've added. PR sorta large already, so I'd push tracing fs activity to a follow up PR, if okay with you.
</blockquote></details>

Sure, separate PR is better.

I guess if logging_fs always traced reads it would make a lot of existing tests too verbose (and maybe even non-deterministic). We can add it behind a flag, or we can add some test-only logging in our cache implementation.

RaduBerinde (approved):
> Reviewed 5 of 6 files at r1, 7 of 10 files at r4, 2 of 2 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2539)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @itsbilal)

Issue #2549, opened 2023-05-22
db: unbounded L0 garbage accumulation


jbowens:
Special level scoring logic applies to L0. None of this logic accounts for potentially high compensated sizes (`PointDeletionBytesEstimate` and `RangeDeletionBytesEstimate`). This can result in tombstones deleting an unbounded volume of data lower in the LSM accumulating in L0, preventing the space reclamation.

TODO(jackson): Add a detailed outline of L0's existing scoring logic.

Jira issue: PEBBLE-49

Epic CRDB-40361

jbowens:
Yeah, I think that's reasonable.

Is there a downside to incorporating (compensated) level size into L0 scoring more generally, eg, maybe as a multiple of LBaseMaxBytes, and taking the MAX of the computed scores?

sumeerbhola:
> Is there a downside to incorporating (compensated) level size into L0 scoring more generally, eg, maybe as a multiple of LBaseMaxBytes, and taking the MAX of the computed scores?

Do you mean something like `score = L0_compensated_size / (k * LbaseMaxBytes)`? Should be ok too, and more principled. I was being paranoid that there may be some existing workloads with a sub-level count of 1 and > 64MB in L0 that manage to delay L0 => Lbase compactions (and hence reduce write amp), so was trying to scope the re-scoring logic narrowly to the garbage case.

PR #2550, opened 2023-05-22
db: point tombstone bug fixes

jbowens:
**db: remove experimental point tombstone weight option** 

Remove the experimental point_tombstone_weight option. The introduction of this
option accidentally caused the loss of point tombstone compensation applied to
level scoring. It's been inconculsive whether setting this value >1.0 has ever
helped customers.

**db: consider scheduling compaction when adding point tombstone compensation**

Previously, when the asynchronous table stats collector calculated a positive
PointDeletionsBytesEstimate, it did not consider scheduling compactions unless
it also calculated a positive RangeDeletionsBytesEstimate. If the database was
otherwise quiet with few flushes and no in-progress compactions, this could
delay the scheduling of a compaction until the next flush or ingest, despite
levels have scores ≥ 1.0. This was illustrated with CockroachDB's point
tombstone roachtest, with at times no running compactions despite levels having
scores as high as 50 due to high volumes of data dropped by point tombstonees.

Note, an issue still remains whereby L0 files that delete large amounts of data
do not trigger compactions out of L0 on their own. This is a consequence of
L0's different scoring heuristics which do not consider compensated size. See
#2549.


open.go
```diff
@@ -512,7 +512,7 @@ func Open(dirname string, opts *Options) (db *DB, _ error) {
 	}
 	d.mu.tableStats.cond.L = &d.mu.Mutex
 	d.mu.tableValidation.cond.L = &d.mu.Mutex
-	if !d.opts.ReadOnly && !d.opts.private.disableTableStats {
+	if !d.opts.ReadOnly {
```
itsbilal, line 515: Is this change intentional?

itsbilal (approved):
LGTM, thanks for doing this!

Rip the short-lived `PointTombstoneWeight`

PR #2552, opened 2023-05-22
tool: improve `wal dump` output

jbowens:
Improve the debug tool's output for dumping WALs. Previously, the contents of range keys (start key, end key, suffix and value) were never printed. Additionally, the file number associated with an INGESTSST (encoded within the user key as a varint) was not printed.

### DIFF
--- a/internal/keyspan/span.go
+++ b/internal/keyspan/span.go
@@ -368,6 +368,7 @@ func (s Span) String() string {
 
 // Pretty returns a formatter for the span.
 func (s Span) Pretty(f base.FormatKey) fmt.Formatter {
+	// TODO(jackson): Take a base.FormatValue to format Key.Value too.
 	return prettySpan{s, f}
 }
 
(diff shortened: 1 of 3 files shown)


bananabrick (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2552#-:-NW5CB1HA6MPug0PN51v:bg7igkv)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`tool/wal.go` line 144 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2552#-NW5B04m4_X77DkF2zn6:-NW5B04m4_X77DkF2zn7:bhmbwyj) ([raw file](https://github.com/cockroachdb/pebble/blob/6c84750aaf864cda16f4cfbb60b6439e0eec1fae/tool/wal.go#L144)):*
> ```Go
> 					case base.InternalKeyKindIngestSST:
> 						fileNum, _ := binary.Uvarint(ukey)
> 						fmt.Fprintf(stdout, "%s", base.FileNum(fileNum))
> ```

This makes sense. There might be multiple filenums in the batch, but they'll be printed out separately on each iteration of the loop.

jbowens:
TFTR!

PR #2553, opened 2023-05-23
sharedcache: fix two bugs, add testing & assertions, esp. a randomized test

joshimhoff:
```
$ go test . -tags=invariants -run='TestSharedCacheRandomized' --count=1000
ok  	github.com/cockroachdb/pebble/objstorage/objstorageprovider	50.720s
```

**vfs: add logging for OpenReadWrite & WriteAt**

This commit adds logging for OpenReadWrite & WriteAt to the logging vfs.

**sharedcache: fix two bugs, add assertions & test case**

This commit fixes two bugs which were uncovered initially by the randomized
test that is added in the next commit. This commit adds assertions & a
non-randomized test case also.

The first bug was an incorrect computation of the length of the adjusted buffer
with the desired length being a multiple of the block size. The necessary
adjustment by ofs-adjustedOfs happened in the wrong place in the mathematical
expression computing the desired length of the adjusted buffer.

The second bug was an error in bounding calls to shard.Get and shard.Set so
that the reads & writes don't cross a shard boundary. The computation did not
take into account n, so it did not take into account previous calls to
shard.Get & shard.Set made in the loop in cache.Get & cache.Set.

**sharedcache: implement a randomized test**

This commit implements a randomized test of the shared cache. The main goal is
to find correctness bugs. The test found two bugs fixed in the previous commit.

### DIFF
--- a/objstorage/objstorageprovider/shared_cache.go
+++ b/objstorage/objstorageprovider/shared_cache.go
@@ -76,7 +76,7 @@ func (sc *sharedCache) ReadAt(
 	// that all shards have the same block size.
 	blockSize := sc.shards[0].blockSize
 	adjustedOfs := ((ofs + int64(n)) / int64(blockSize)) * int64(blockSize)
-	adjustedP := make([]byte, (((len(p[n:]) + int(ofs-adjustedOfs))+(blockSize-1))/blockSize)*blockSize)
+	adjustedP := make([]byte, (((len(p[n:])+int(ofs-adjustedOfs))+(blockSize-1))/blockSize)*blockSize)
 
 	// Read the rest from the object.
 	sc.misses.Add(1)
@@ -112,7 +112,7 @@ func (sc *sharedCache) Get(fileNum base.FileNum, p []byte, ofs int64) (n int, _ 
 	for {
 		shard := sc.getShard(fileNum, ofs+int64(n))
 		cappedLen := len(p[n:])
-		if toBoundary := int(shardingBlockSize - ((ofs+int64(n)) % shardingBlockSize)); cappedLen > toBoundary {
+		if toBoundary := int(shardingBlockSize - ((ofs + int64(n)) % shardingBlockSize)); cappedLen > toBoundary {
 			cappedLen = toBoundary
 		}
 		numRead, err := shard.Get(fileNum, p[n:n+cappedLen], ofs+int64(n))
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):
:lgtm: modulo some comments

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2553#-:-NW8LH6c9GaaLyD_9yA8:b6eq0yt)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @jbowens and @joshimhoff)

___
*[`objstorage/objstorageprovider/shared_cache.go` line 79 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2553#-NW8JPrvCEU2k4x4xOG9:-NW8JPrvCEU2k4x4xOGA:b2aw3as) ([raw file](https://github.com/cockroachdb/pebble/blob/bcd697dddf2fe4d8c70de4812a284ee23bf4a258/objstorage/objstorageprovider/shared_cache.go#L79)):*
> ```Go
> 	blockSize := sc.shards[0].blockSize
> 	adjustedOfs := ((ofs + int64(n)) / int64(blockSize)) * int64(blockSize)
> 	adjustedP := make([]byte, (((len(p[n:]) + int(ofs-adjustedOfs))+(blockSize-1))/blockSize)*blockSize)
> ```

This is very hard to reason about.. I'd do something like this:

```
firstBlock := (ofs + int64(n)) / blockSize
lastBlock := (ofs + len(p) - 1) / blockSize

adjustedOfs := firstBlock * blockSize
adjustedLen := (lastBlock - firstBlock + 1) * blockSize
```
___
*[`vfs/logging_fs.go` line 50 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2553#-NW8IqEaAy9B29U3MRew:-NW8IqEaAy9B29U3MRex:b-brg5mb) ([raw file](https://github.com/cockroachdb/pebble/blob/0f5c5702c9ea8a90b89a821facd38da23346c144/vfs/logging_fs.go#L50)):*
> ```Go
> 
> func (fs *loggingFS) OpenReadWrite(name string, opts ...OpenOption) (File, error) {
> 	fs.logFn("openreadwrite: %s", name)
> ```

[nit] `open-read-write` for consistency with others
___
*[`objstorage/objstorageprovider/shared_cache_test.go` line 114 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2553#-NW8KxrW0dp-CnwO2Ok0:-NW8KxrW0dp-CnwO2Ok1:bb9n7rn) ([raw file](https://github.com/cockroachdb/pebble/blob/0fdec945fd0b3ae53d309a33809ed3716588aa83/objstorage/objstorageprovider/shared_cache_test.go#L114)):*
> ```Go
> 	rand.Seed(seed)
> 
> 	blockSize := 32 * 1024 // 32 KB
> ```

We could have a couple of subtests instead of randomizing the blockSize 
___
*[`objstorage/objstorageprovider/shared_cache_test.go` line 146 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2553#-NW8KoOb00vgJusT1gh9:-NW8KoOb00vgJusT1ghA:b-jr85wc) ([raw file](https://github.com/cockroachdb/pebble/blob/0fdec945fd0b3ae53d309a33809ed3716588aa83/objstorage/objstorageprovider/shared_cache_test.go#L146)):*
> ```Go
> 	defer readable.Close()
> 
> 	offset := rand.Int63n(size)
> ```

[nit] we should do this in a loop (e.g. 100 times).

Issue #2541, opened 2023-05-19
disagg: implement a basic secondary cache


RaduBerinde:
This issue tracks implementing a minimally viable secondary cache. The block size will be fixed (32KB by default) and all metadata will be kept in memory. The cache is not persistent upon restart.

joshimhoff:
Gonna call this bit done.

PR #2561, opened 2023-05-24
sharedcache: improved randomized test, fix bugs

joshimhoff:
Part of https://github.com/cockroachdb/pebble/issues/2541. Follow up on https://github.com/cockroachdb/pebble/pull/2553.

**sharedcache: improved randomized test, fix bugs**

This commit improves the randomized test based on a suggestion from Radu. The test now does 100 distinct reads to the same file. This leads to cache eviction happening, which has flushed out bugs in the handling of the case where we only read a prefix of data from the cache.

This commit fixes the found bugs, while also refactoring the logic for clarity, again based on a suggestion from Radu. The two found bugs were in the cache's ReadAt method. The first was the use of ofs instead of ofs+n in the computation of the length of the cache-block-multiple-sized buffer to read into. The second was the use of ofs instead of ofs+n in the slice expression in which we read what is read into p, which the user passes into ReadAt. The bugs are of the same flavor. To guard against this happening in the future, I have refactored the code to add n to ofs after fetching from the cache.

### DIFF
--- a/objstorage/objstorageprovider/shared_cache.go
+++ b/objstorage/objstorageprovider/shared_cache.go
@@ -22,6 +22,8 @@ type sharedCache struct {
 	shards []sharedCacheShard
 	logger base.Logger
 
+	blockSize int
+
 	// TODO(josh): Have a dedicated metrics struct. Right now, this
 	// is just for testing.
 	misses atomic.Int32
@@ -35,7 +37,9 @@ func openSharedCache(
 		return nil, errors.Errorf("cache size %d lower than min %d", sizeBytes, min)
 	}
 
-	sc := &sharedCache{}
+	sc := &sharedCache{
+		blockSize: blockSize,
+	}
 	sc.shards = make([]sharedCacheShard, numShards)
 	blocksPerShard := sizeBytes / int64(numShards) / int64(blockSize)
 	for i := range sc.shards {
@@ -62,21 +66,40 @@ func (sc *sharedCache) Close() error {
 func (sc *sharedCache) ReadAt(
 	ctx context.Context, fileNum base.FileNum, p []byte, ofs int64, readable objstorage.Readable,
 ) error {
-	n, err := sc.Get(fileNum, p, ofs)
-	if err != nil {
-		return err
-	}
-	if n == len(p) {
-		// Everything was in cache!
-		return nil
+	{
+		n, err := sc.Get(fileNum, p, ofs)
+		if err != nil {
+			return err
+		}
+		if n == len(p) {
+			// Everything was in cache!
+			return nil
+		}
+
+		// Note this. The below code does not need the original ofs, as with the earlier
+		// reading from the cache done, the relevant offset is ofs + int64(n). Same with p.
+		ofs += int64(n)
+		p = p[n:]
+
+		if invariants.Enabled {
+			if n != 0 && ofs%int64(sc.blockSize) != 0 {
+				panic(fmt.Sprintf("after non-zero read from cache, ofs is not block-aligned: %v %v", ofs, n))
+			}
+		}
 	}
 
 	// We must do reads with offset & size that are multiples of the block size. Else
-	// later cache hits may return incorrect zeroed results from the cache. We assume
-	// that all shards have the same block size.
-	blockSize := sc.shards[0].blockSize
-	adjustedOfs := ((ofs + int64(n)) / int64(blockSize)) * int64(blockSize)
-	adjustedP := make([]byte, (((len(p[n:])+int(ofs-adjustedOfs))+(blockSize-1))/blockSize)*blockSize)
+	// later cache hits may return incorrect zeroed results from the cache.
+	firstBlockInd := ofs / int64(sc.blockSize)
+	adjustedOfs := firstBlockInd * int64(sc.blockSize)
+
+	// Take the length of what is left to read plus the length of the adjustment of
+	// the offset plus the size of a block minus one and divide by the size of a block
+	// to get the number of blocks to read from the readable.
+	sizeOfOffAdjustment := int(ofs - adjustedOfs)
+	numBlocksToRead := ((len(p) + sizeOfOffAdjustment) + (sc.blockSize - 1)) / sc.blockSize
+	adjustedLen := numBlocksToRead * sc.blockSize
+	adjustedP := make([]byte, adjustedLen)
 
 	// Read the rest from the object.
 	sc.misses.Add(1)
@@ -87,7 +110,7 @@ func (sc *sharedCache) ReadAt(
 	if err := readable.ReadAt(ctx, adjustedP, adjustedOfs); err != nil && err != io.EOF {
 		return err
 	}
-	copy(p[n:], adjustedP[ofs%int64(blockSize):])
+	copy(p, adjustedP[sizeOfOffAdjustment:])
 
 	// TODO(josh): Writing back to the cache should be async with respect to the
 	// call to ReadAt.
@@ -138,7 +161,7 @@ func (sc *sharedCache) Get(fileNum base.FileNum, p []byte, ofs int64) (n int, _ 
 // If all of p is not written to the shard, Set returns a non-nil error.
 func (sc *sharedCache) Set(fileNum base.FileNum, p []byte, ofs int64) error {
 	if invariants.Enabled {
-		if ofs%int64(sc.shards[0].blockSize) != 0 || len(p)%sc.shards[0].blockSize != 0 {
+		if ofs%int64(sc.blockSize) != 0 || len(p)%sc.blockSize != 0 {
 			panic(fmt.Sprintf("Set with ofs & len not multiples of block size: %v %v", ofs, len(p)))
 		}
 	}
--- a/objstorage/objstorageprovider/shared_cache_test.go
+++ b/objstorage/objstorageprovider/shared_cache_test.go
@@ -111,49 +111,52 @@ func TestSharedCacheRandomized(t *testing.T) {
 	fmt.Printf("seed: %v\n", seed)
 	rand.Seed(seed)
 
-	blockSize := 32 * 1024 // 32 KB
-	if rand.Intn(2) == 0 {
-		blockSize = 1024 * 1024 // 1 MB
+	helper := func(blockSize int) func(t *testing.T) {
+		return func(t *testing.T) {
+			numShards := rand.Intn(64) + 1
+			cacheSize := int64(shardingBlockSize * numShards) // minimum allowed cache size
+
+			cache, err := openSharedCache(fs, "", blockSize, cacheSize, numShards)
+			require.NoError(t, err)
+			defer cache.Close()
+
+			file, err := fs.Create("test")
+			require.NoError(t, err)
+
+			size := rand.Int63n(cacheSize)
+
+			// With invariants on, Write will modify its input buffer.
+			toWrite := make([]byte, size)
+			wrote := make([]byte, size)
+			for i := 0; i < int(size); i++ {
+				toWrite[i] = byte(i)
+				wrote[i] = byte(i)
+			}
+
+			n, err := file.Write(wrote)
+			require.NoError(t, err)
+			require.Equal(t, size, int64(n))
+
+			readable, err := newFileReadable(file, fs, "test")
+			require.NoError(t, err)
+			defer readable.Close()
+
+			const numDistinctReads = 100
+			for i := 0; i < numDistinctReads; i++ {
+				offset := rand.Int63n(size)
+
+				got := make([]byte, size-offset)
+				err = cache.ReadAt(ctx, 1, got, offset, readable)
+				require.NoError(t, err)
+				require.Equal(t, toWrite[int(offset):], got)
+
+				got = make([]byte, size-offset)
+				err = cache.ReadAt(ctx, 1, got, offset, readable)
+				require.NoError(t, err)
+				require.Equal(t, toWrite[int(offset):], got)
+			}
+		}
 	}
-	numShards := rand.Intn(64) + 1
-	cacheSize := int64(shardingBlockSize * numShards) // minimum allowed cache size
-
-	cache, err := openSharedCache(fs, "", blockSize, cacheSize, numShards)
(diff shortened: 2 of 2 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2561#-:-NWDagzR9hNjnRqEtWCn:bnfp4nl)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @joshimhoff and @sumeerbhola)

___
*[`objstorage/objstorageprovider/shared_cache.go` line 75 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWD_Gfn0z6CbY-kRBwN:-NWD_Gfn0z6CbY-kRBwO:bjvo198) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache.go#L75)):*
> ```Go
> 	// Note this. The below code does not need the original ofs, as with the earlier
> 	// reading from the cache done, the relevant offset is ofs + int64(n).
> 	ofs += int64(n)
> ```

[nit] this can be confusing since we continue to use `n`. If we do this, we should also do `p = p[n:]` and then we can forget about `n`.
___
*[`objstorage/objstorageprovider/shared_cache.go` line 80 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWDaW2_453Ec9M_gm4P:-NWDaW2_453Ec9M_gm4Q:b-dbvmp2) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache.go#L80)):*
> ```Go
> 	// later cache hits may return incorrect zeroed results from the cache. We assume
> 	// that all shards have the same block size.
> 	blockSize := sc.shards[0].blockSize
> ```

[nit] we should put the `blockSize` in `sc` as well so we don't have to explain this
___
*[`objstorage/objstorageprovider/shared_cache.go` line 101 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWD_2sgAf6hxW6Sqjy3:-NWD_2sgAf6hxW6Sqjy4:b-s2uxyx) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache.go#L101)):*
> ```Go
> 		return err
> 	}
> 	copy(p[n:], adjustedP[ofs%int64(blockSize):])
> ```

`ofs % blockSize` is the same as `ofs - adjustedOfs` above. I say we make that a separate variable `leadingLen` and use it in both places. `adjustedOfs` can just be `ofs - leadingLen`

sumeerbhola (dismissed):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2561)* status: 0 of 2 files reviewed, 6 unresolved discussions (waiting on @joshimhoff and @RaduBerinde)

___
*[`objstorage/objstorageprovider/shared_cache.go` line 82 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWDs5TZ5-GBtU6aj-6e:-NWDs5TZ5-GBtU6aj-6f:b5b0v2f) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache.go#L82)):*
> ```Go
> 	blockSize := sc.shards[0].blockSize
> 
> 	firstBlockInd := ofs / int64(blockSize)
> ```

Perhaps we should add an assertion that if `n != 0` the `ofs` is already block aligned.
___
*[`objstorage/objstorageprovider/shared_cache.go` line 88 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWDtoC470ehBmOj9aoR:-NWDtoC55YxlMTkzYqpU:b-gz7mew) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache.go#L88)):*
> ```Go
> 	// the offset plus the size of a block minus one and divide by the size of a block
> 	// to get the number of blocks to read from the readable.
> 	numBlocksToRead := ((len(p[n:]) + int(ofs-adjustedOfs)) + (blockSize - 1)) / blockSize
> ```

There is a bunch of `blockSize` related arithmetic scattered over this method and `sharedCacheShard` methods. Is there some way to unify these?
It sounds like we either have an offset that is known to align with blockSize (later iterations of a loop) or not. And in the latter case we want a method that will return the index of the block (in the context of the file), the offset to start from in that block, and the bytes we can read in that block. And a method to convert the index of the block into a file offset (for the case here that is going to do a read from the file). It's easier from a readability perspective to have these be in (inlined) methods than stare at each unique piece of arithmetic carefully. 
___
*[`objstorage/objstorageprovider/shared_cache_test.go` line 161 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWDsTp58o_ZQHVzpE9S:-NWDsTp58o_ZQHVzpE9T:bip9wjr) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache_test.go#L161)):*
> ```Go
> 	}
> 	t.Run("32 KB", helper(32*1024))
> 	t.Run("1 MB", helper(1024*1024))
> ```

How about block sizes randomly chosen in the interval [1,1000)? Small block sizes should exercise all the loops relating to slowly filling up the read byte slice.

joshimhoff (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2561)* status: 0 of 2 files reviewed, 6 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`objstorage/objstorageprovider/shared_cache.go` line 88 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWDtoC470ehBmOj9aoR:-NWDwxST91hqB59_Jyg0:b-n9p1pb) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache.go#L88)):*

There is a bunch of `blockSize` related arithmetic scattered over this method and `sharedCacheShard` methods. Is there some way to unify these?
It sounds like we either have an offset that is known to align with blockSize (later iterations of a loop) or not. And in the latter case we want a method that will return the index of the block (in the context of the file), the offset to start from in that block, and the bytes we can read in that block. And a method to convert the index of the block into a file offset (for the case here that is going to do a read from the file). It's easier from a readability perspective to have these be in (inlined) methods than stare at each unique piece of arithmetic carefully. 
</blockquote></details>

Ya, that's a good idea, and something I have been thinking about lately too. I wonder if you are comfortable with a follow up PR?
___
*[`objstorage/objstorageprovider/shared_cache_test.go` line 161 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWDsTp58o_ZQHVzpE9S:-NWDxH5j98Vy7JM4YUD1:b10x7i) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache_test.go#L161)):*

How about block sizes randomly chosen in the interval [1,1000)? Small block sizes should exercise all the loops relating to slowly filling up the read byte slice.
</blockquote></details>

Good idea. Are you comfortable with a follow up PR? 

My idea is to keep expanding the tests over time, as well as the clarity of the code (as per you other comment), while iterating via fairly small PRs that make improvements to the code, rather than large PRs with very long review cycles. Let me know if my way of working is not how we want to on storage tho!

sumeerbhola:
@joshimhoff follow up PRs are fine

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2561)* status: 0 of 2 files reviewed, 6 unresolved discussions (waiting on @joshimhoff and @sumeerbhola)

___
*[`objstorage/objstorageprovider/shared_cache.go` line 82 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWDs5TZ5-GBtU6aj-6e:-NWEKYq8EBFBal1Oy_bn:b-d5lt0w) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache.go#L82)):*

Perhaps we should add an assertion that if `n != 0` the `ofs` is already block aligned.
</blockquote></details>

It doesn't have to be, this is normally the offset at which we read an SST block.
___
*[`objstorage/objstorageprovider/shared_cache.go` line 88 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWDtoC470ehBmOj9aoR:-NWEKs2U4LFmaqCK_zIA:bke7zx2) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache.go#L88)):*

Ya, that's a good idea, and something I have been thinking about lately too. I wonder if you are comfortable with a follow up PR?
</blockquote></details>

I like the idea too. As part of that we can consider creating a `type blockIndex int64` and have some helper methods to convert to/from offsets. The type safety will point out issues where we may interpret a block index as an offset or vice-versa.

RaduBerinde:
*[`objstorage/objstorageprovider/shared_cache.go` line 82 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWDs5TZ5-GBtU6aj-6e:-NWELIqJEF1Xyraifhkh:b-2youct) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache.go#L82)):*

It doesn't have to be, this is normally the offset at which we read an SST block.
</blockquote></details>

Sorry, I misread. That's right, good idea.

joshimhoff (commented):
TTFR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2561#-:-NWI84mnEJzLJ69F4qW1:b-80vssw)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`objstorage/objstorageprovider/shared_cache.go` line 75 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWD_Gfn0z6CbY-kRBwN:-NWI7Qyn6WogL4qvKALk:b-8clh22) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache.go#L75)):*

[nit] this can be confusing since we continue to use `n`. If we do this, we should also do `p = p[n:]` and then we can forget about `n`.
</blockquote></details>

Done. Tho I introduced the change, I'm unsure if I prefer this approach of mutating `ofs` and `p`. I guess we'll go with it for now.
___
*[`objstorage/objstorageprovider/shared_cache.go` line 80 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWDaW2_453Ec9M_gm4P:-NWI25nr0ibksc1spz20:bcvjgzp) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache.go#L80)):*

[nit] we should put the `blockSize` in `sc` as well so we don't have to explain this
</blockquote></details>

Good idea. Done.
___
*[`objstorage/objstorageprovider/shared_cache.go` line 82 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWDs5TZ5-GBtU6aj-6e:-NWI3RtL5Z5kGyljn5Ty:bddzlq5) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache.go#L82)):*

Sorry, I misread. That's right, good idea.
</blockquote></details>

Done. Thanks for this suggestion.
___
*[`objstorage/objstorageprovider/shared_cache.go` line 88 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWDtoC470ehBmOj9aoR:-NWI3UWaBg0Js0FcRtyU:bb3mwmu) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache.go#L88)):*

I like the idea too. As part of that we can consider creating a `type blockIndex int64` and have some helper methods to convert to/from offsets. The type safety will point out issues where we may interpret a block index as an offset or vice-versa.
</blockquote></details>

Good idea, Radu. Will work on this soon.
___
*[`objstorage/objstorageprovider/shared_cache.go` line 101 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2561#-NWD_2sgAf6hxW6Sqjy3:-NWI5G9hET8DiLzq82BO:b-4er1lf) ([raw file](https://github.com/cockroachdb/pebble/blob/ffc4d8f8811d5fe57a6a79e66085de040b76a7b8/objstorage/objstorageprovider/shared_cache.go#L101)):*

`ofs % blockSize` is the same as `ofs - adjustedOfs` above. I say we make that a separate variable `leadingLen` and use it in both places. `adjustedOfs` can just be `ofs - leadingLen`
</blockquote></details>

Good observation. I've named the new variable `sizeOfOffAdjustment` and used in the computation of `numBlocksToRead` & the call to `copy` later on. Partly thinking forward to Sumeer's suggestion of encapsulating more offset / block size logic in functions, I'd prefer to compute `adjustedOfs` as I have it now. I am thinking we have a function called `blockAlign` that does a divide by block size followed by a multiple. I don't feel strongly about any of this tho. Let me know if you have strong feelings.

RaduBerinde:
:lgtm:

PR #2563, opened 2023-05-24
db: fix size adjustment for in-progress compactions

jbowens:
In #2436 the semantics of d.mu.compact.inProgress changed, and it's now possible for that slice to hold compactions whose version edit has already been applied. During compaction picking, these compactions should not adjust level sizes because the current LSM already reflects the resulting LSM state.

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -935,6 +935,12 @@ func calculateSizeAdjust(inProgressCompactions []compactionInfo) [numLevels]int6
 	var sizeAdjust [numLevels]int64
 	for i := range inProgressCompactions {
 		c := &inProgressCompactions[i]
+		// If this compaction's version edit has already been applied, there's
+		// no need to adjust: The LSM we'll examine will already reflect the
+		// new LSM state.
+		if c.versionEditApplied {
+			continue
+		}
 
 		for _, input := range c.inputs {
 			real := int64(input.files.SizeSum())
--- a/compaction_picker_test.go
+++ b/compaction_picker_test.go
@@ -11,6 +11,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


jbowens (commented):
I'll merge #2550, rebase and try to extend its new unit test to cover this scenario.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2563#-:-NWDMEVa35hiXVoNub5y:brl5zqd)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sumeerbhola)

sumeerbhola (commented):
:lgtm: 

> I'll merge [#2550](https://github.com/cockroachdb/pebble/pull/2550), rebase and try to extend its new unit test to cover this scenario.

ack

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2563#-:-NWDnF074NUMOs4bojqa:bxyyxsg)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @jbowens)

jbowens (commented):
in writing the unit test, I'd run into the issue fixed in c4be581b807fbcb6d72dce63ba7562fd35277072. now that that's fixed, the unit test is here and this is ready for a final look

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2563#-:-NXGuIls9p_cER2MKBJU:bux8g4x)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @sumeerbhola)

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2563#-:-NXSBTeq73rtgxsJrROW:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`compaction.go` line 1685 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2563#-NXS8wBwAniZmiRvdvF5:-NXS8wBx0SAtwoGMBpzD:b4mtiq6) ([raw file](https://github.com/cockroachdb/pebble/blob/a929a42b9321f02a545f1a33c9d636a6849215a9/compaction.go#L1685)):*
> ```Go
> 	l0InProgress := inProgressL0Compactions(d.getInProgressCompactionInfoLocked(c))
> 	d.mu.versions.currentVersion().L0Sublevels.InitCompactingFileInfo(l0InProgress)
> 	d.mu.compact.cond.Broadcast()
> ```

Why do we need this signaling of the `cond`?

Also, I'm confused about why we do
```
d.maybeScheduleCompaction()
d.mu.compact.cond.Broadcast()
```
in `DB.compact` given that the compaction will run in a different goroutine. Shouldn't we be signaling when the compaction or flush finishes? Perhaps this is what this code is effectively doing?
___
*[`testdata/compaction_picker_scores` line 225 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2563#-NXSBQcc7r0jZ1kjUpVt:-NXSBQcc7r0jZ1kjUpVu:bhpm87k) ([raw file](https://github.com/cockroachdb/pebble/blob/a929a42b9321f02a545f1a33c9d636a6849215a9/testdata/compaction_picker_scores#L225)):*
> ```
> # The scores and sizes should be stable between when the version edit has been
> # applied but the compaction has not completed, and when the compaction is
> # finally complete.
> ```

nice test

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2563#-:-NXX5GOh5bfb_wqfiWA4:b-814g8k)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`compaction.go` line 1685 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2563#-NXS8wBwAniZmiRvdvF5:-NXX58LTCnhRfiMq0I4C:b-63m3px) ([raw file](https://github.com/cockroachdb/pebble/blob/a929a42b9321f02a545f1a33c9d636a6849215a9/compaction.go#L1685)):*

Why do we need this signaling of the `cond`?

Also, I'm confused about why we do
```
d.maybeScheduleCompaction()
d.mu.compact.cond.Broadcast()
```
in `DB.compact` given that the compaction will run in a different goroutine. Shouldn't we be signaling when the compaction or flush finishes? Perhaps this is what this code is effectively doing?

</blockquote></details>

this was to facilitate the test, which needs to wait until the version edit is committed. i removed this and updated the test to use a sleep loop.

PR #2565, opened 2023-05-24
crl-release-22.2: point tombstone bug fixes

jbowens:
22.2 backport of #2550.

----

**db: remove experimental point tombstone weight option**

Remove the experimental point_tombstone_weight option. The introduction of this
option accidentally caused the loss of point tombstone compensation applied to
level scoring. It's been inconculsive whether setting this value >1.0 has ever
helped customers.

**db: consider scheduling compaction when adding point tombstone compensation**

Previously, when the asynchronous table stats collector calculated a positive
PointDeletionsBytesEstimate, it did not consider scheduling compactions unless
it also calculated a positive RangeDeletionsBytesEstimate. If the database was
otherwise quiet with few flushes and no in-progress compactions, this could
delay the scheduling of a compaction until the next flush or ingest, despite
levels having scores ≥ 1.0. This was illustrated with CockroachDB's point
tombstone roachtest, which at times ran no compactions despite levels having
scores as high as 50 due to high volumes of data dropped by point tombstones.

Note, an issue still remains whereby L0 files that delete large amounts of data
do not trigger compactions out of L0 on their own. This is a consequence of
L0's different scoring heuristics which do not consider compensated size. See
#2549.

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -575,15 +575,16 @@ type candidateLevelInfo struct {
 	file manifest.LevelFile
 }
 
+func fileCompensation(f *fileMetadata) uint64 {
+	return uint64(f.Stats.PointDeletionsBytesEstimate) + f.Stats.RangeDeletionsBytesEstimate
+}
+
 // compensatedSize returns f's file size, inflated according to compaction
 // priorities.
 func compensatedSize(f *fileMetadata) uint64 {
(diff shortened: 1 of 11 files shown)


itsbilal (approved):
LGTM :shipit:

jbowens:
TFTR!

PR #2566, opened 2023-05-24
db: fix compaction file picking bounds checks

jbowens:
When picking a seed file within a level to build a compaction around, the compaction picker uses a 'min-overlapping ratio' heuristic to pick a file such that the sum of the overlapping output-level files have the least size relative to the input-level file's size. This requires the compaction picker to check for file overlap between levels as it walks across a level.

Previously, these overlap checks were performed in terms of internal key comparisons. However, internal key comparisons are not what ultimately determine whether or not a compaction overlaps. A table with a largest key b.SET.10 and a table with a smallest key b.SET.0 overlap (they both contain the key "b"), despite b.SET.10 sorting before b.SET.0 within the internal key keyspace. This bug did not break LSM invariants because compaction inputs setup would always separately calculate output-level overlaps, which would pull in the correct file. However, in these cases the compaction picker could pick a suboptimal file for compaction, mistakenly calculating an overlapping ratio that's lower than the file's true overlapping ratio.

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -1085,12 +1085,12 @@ func (p *compactionPickerByScore) pickFile(
 		var overlappingBytes uint64
 
 		// Trim any output-level files smaller than f.
-		for outputFile != nil && base.InternalCompare(cmp, outputFile.Largest, f.Smallest) < 0 {
+		for outputFile != nil && sstableKeyCompare(cmp, outputFile.Largest, f.Smallest) < 0 {
 			outputFile = outputIter.Next()
 		}
 
 		compacting := f.IsCompacting()
-		for outputFile != nil && base.InternalCompare(cmp, outputFile.Smallest, f.Largest) < 0 {
+		for outputFile != nil && sstableKeyCompare(cmp, outputFile.Smallest, f.Largest) <= 0 {
 			overlappingBytes += outputFile.Size
 			compacting = compacting || outputFile.IsCompacting()
 
@@ -1109,7 +1109,7 @@ func (p *compactionPickerByScore) pickFile(
 			// If the file in the next level extends beyond f's largest key,
 			// break out and don't advance outputIter because f's successor
 			// might also overlap.
-			if base.InternalCompare(cmp, outputFile.Largest, f.Largest) > 0 {
+			if sstableKeyCompare(cmp, outputFile.Largest, f.Largest) > 0 {
 				break
 			}
 			outputFile = outputIter.Next()
--- a/compaction_picker_test.go
+++ b/compaction_picker_test.go
@@ -1293,6 +1293,73 @@ func TestCompactionPickerCompensatedSize(t *testing.T) {
 	}
 }
 
+func TestCompactionPickerPickFile(t *testing.T) {
+	fs := vfs.NewMem()
+	opts := &Options{
+		Comparer:           testkeys.Comparer,
+		FormatMajorVersion: FormatNewest,
+		FS:                 fs,
+	}
+
+	d, err := Open("", opts)
+	require.NoError(t, err)
+	defer func() {
+		if d != nil {
+			require.NoError(t, d.Close())
+		}
+	}()
+
+	datadriven.RunTest(t, "testdata/compaction_picker_pick_file", func(t *testing.T, td *datadriven.TestData) string {
+		switch td.Cmd {
+		case "define":
+			require.NoError(t, d.Close())
+
+			d, err = runDBDefineCmd(td, opts)
+			if err != nil {
+				return err.Error()
+			}
+			d.mu.Lock()
+			s := d.mu.versions.currentVersion().String()
+			d.mu.Unlock()
+			return s
+
+		case "file-sizes":
+			return runTableFileSizesCmd(td, d)
+
+		case "pick-file":
+			s := strings.TrimPrefix(td.CmdArgs[0].String(), "L")
+			level, err := strconv.Atoi(s)
+			if err != nil {
+				return fmt.Sprintf("unable to parse arg %q as level", td.CmdArgs[0].String())
+			}
+			if level == 0 {
+				panic("L0 picking unimplemented")
+			}
+			d.mu.Lock()
+			defer d.mu.Unlock()
+
+			// Use maybeScheduleCompactionPicker to take care of all of the
+			// initialization of the compaction-picking environment, but never
+			// pick a compaction; just call pickFile using the user-provided
+			// level.
+			var lf manifest.LevelFile
+			var ok bool
+			d.maybeScheduleCompactionPicker(func(untypedPicker compactionPicker, env compactionEnv) *pickedCompaction {
+				p := untypedPicker.(*compactionPickerByScore)
+				lf, ok = p.pickFile(level, level+1, env.earliestSnapshotSeqNum)
+				return nil
+			})
+			if !ok {
+				return "(none)"
+			}
+			return lf.FileMetadata.String()
+
+		default:
+			return fmt.Sprintf("unknown command: %s", td.Cmd)
+		}
+	})
+}
+
 func TestCompactionPickerScores(t *testing.T) {
 	fs := vfs.NewMem()
 	opts := &Options{
--- a/testdata/compaction_picker_pick_file
+++ b/testdata/compaction_picker_pick_file
@@ -1,0 +1,107 @@
+# Simple test with a single file per level.
+
+define
+L1
+  b.SET.11:foo
+  c.SET.11:foo
+L2
+  c.SET.0:foo
+  d.SET.0:foo
+----
+1:
+  000004:[b#11,SET-c#11,SET]
+2:
+  000005:[c#0,SET-d#0,SET]
+
+file-sizes
+----
+L1:
+  000004:[b#11,1-c#11,1]: 778 bytes (778 B)
+L2:
+  000005:[c#0,1-d#0,1]: 777 bytes (777 B)
+
+pick-file L1
+----
+000004:[b#11,1-c#11,1]
+
+pick-file L2
+----
+000005:[c#0,1-d#0,1]
+
+# Test a scenario where we should pick a file with a tiny file size over one
+# with a larger file size, because the tiny sized one overlaps zero data in the
+# output level.
+
+define
+L5
+  b.SET.11:<rand-bytes=65536>
+  c.SET.11:<rand-bytes=65536>
+L5
+  e.SET.11:<rand-bytes=2>
+L6
+  a.SET.0:foo
+  d.SET.0:foo
+----
+5:
+  000004:[b#11,SET-c#11,SET]
+  000005:[e#11,SET-e#11,SET]
+6:
+  000006:[a#0,SET-d#0,SET]
+
+pick-file L5
+----
+000005:[e#11,1-e#11,1]
+
+# Test the same scenario as above, but the larger file that overlaps the next
+# level only overlaps on its end boundary key ("c"). c.SET.11 sorts before
+# c.SET.0, but the files should still be considered overlapping.
+
+define
+L5
+  b.SET.11:<rand-bytes=65536>
+  c.SET.11:<rand-bytes=65536>
+L5
+  e.SET.11:<rand-bytes=2>
+L6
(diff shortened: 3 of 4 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2566)* status: 0 of 4 files reviewed, 1 unresolved discussion

___
*[`testdata/compaction_picker_target_level` line 198 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2566#-NWE4tzk5bNmARcOgyZS:-NWE4tzk5bNmARcOgyZT:bck9oaa) ([raw file](https://github.com/cockroachdb/pebble/blob/80f3263b52f2de96a1b2801ff8af25d37c944db8/testdata/compaction_picker_target_level#L198)):*
> ```
> pick ongoing=(5,6)
> ----
> L4->L5: 12.0
> ```

I found this test to be a bit inscrutable and am having trouble determining whether this is the correct output or not—anyone understand the mechanics of this test?

sumeerbhola (changes requested):
> Reviewed 1 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2566)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`compaction_picker.go` line 1095 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2566#-NWINdfr3ri83nmjfVMp:-NWINdfr3ri83nmjfVMq:b-m09a5) ([raw file](https://github.com/cockroachdb/pebble/blob/4a21b62377313e2a7b80e3e891f38eb0e607760a/compaction_picker.go#L1095)):*
> ```Go
> 		for outputFile != nil && sstableKeyCompare(cmp, outputFile.Smallest, f.Largest) <= 0 {
> 			overlappingBytes += outputFile.Size
> 			compacting = compacting || outputFile.IsCompacting()
> ```

Why do we keep iterating in this inner for-loop once compacting is known to be true? I suppose we are going to have to iterate over `outputIter` anyway when considering the next file (to skip the non-overlapping ones), but it still seems odd, and it is also possible that all f's have `f.IsCompacting()` so we can be quite cheap in this overall iteration.
___
*[`compaction_picker.go` line 1112 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2566#-NWIOUNK2o_drFevG4WH:-NWIOUNK2o_drFevG4WI:bmkons0) ([raw file](https://github.com/cockroachdb/pebble/blob/4a21b62377313e2a7b80e3e891f38eb0e607760a/compaction_picker.go#L1112)):*
> ```Go
> 			// break out and don't advance outputIter because f's successor
> 			// might also overlap.
> 			if sstableKeyCompare(cmp, outputFile.Largest, f.Largest) > 0 {
> ```

say the input level is files
`[a#120,a#100] [a#80,a#60]`
and output level is files
`[a#55,a#45] [a#35,a#25] [a#15,a#5]`

And we are examining the first file in the input level. It will end up iterating over all these output files because this expression will return 0. So we will have nothing left to account for, for the second file in the input level?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2566)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @sumeerbhola)

___
*[`compaction_picker.go` line 1095 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2566#-NWINdfr3ri83nmjfVMp:-NWIhUTL3VO2pVn07jiP:bqsjely) ([raw file](https://github.com/cockroachdb/pebble/blob/4a21b62377313e2a7b80e3e891f38eb0e607760a/compaction_picker.go#L1095)):*
> I suppose we are going to have to iterate over `outputIter` anyway when considering the next file (to skip the non-overlapping ones)

yeah, iirc, that was the motivation

> it is also possible that all f's have `f.IsCompacting()` so we can be quite cheap in this overall iteration.

yeah, i think a fast path for the case where the remaining input files are compacting makes sense.
___
*[`compaction_picker.go` line 1112 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2566#-NWIOUNK2o_drFevG4WH:-NWIi2zH8qTCy7DVOFGe:b-flnpbd) ([raw file](https://github.com/cockroachdb/pebble/blob/4a21b62377313e2a7b80e3e891f38eb0e607760a/compaction_picker.go#L1112)):*

say the input level is files
`[a#120,a#100] [a#80,a#60]`
and output level is files
`[a#55,a#45] [a#35,a#25] [a#15,a#5]`

And we are examining the first file in the input level. It will end up iterating over all these output files because this expression will return 0. So we will have nothing left to account for, for the second file in the input level?
</blockquote></details>

good point. Split user keys within the LSM are no longer possible though... at least in all recent format major versions (guaranteed by `FormatSplitUserKeysMarkedCompacted`). I think the result is the same as the preexisting code for earlier format major versions.

sumeerbhola (changes requested):
> Reviewed 3 of 4 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2566)* status: all files reviewed, 4 unresolved discussions (waiting on @jbowens)

___
*[`compaction_picker.go` line 1112 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2566#-NWIOUNK2o_drFevG4WH:-NWgnvlR8nz79FBZd7d8:b6n01ob) ([raw file](https://github.com/cockroachdb/pebble/blob/4a21b62377313e2a7b80e3e891f38eb0e607760a/compaction_picker.go#L1112)):*
I looked again at our `expandToAtomicUnit` calls, and I am wondering whether we are doing something sub-optimal. If we consider a variant of the above example where the output level has only `[a#15,a#5]` and the input level has only `[a#35,a#25]`, and there are no range dels, one should be able to do a move compaction of the latter (but we don't). And if we keep accumulating such move compactions, we can end up in the above scenario and do a move compaction of `[a#80,a#60]`. Practically, this example is not relevant to CockroachDB's MVCC workload since keys have very few Pebble versions, so I am fine with ignoring it.

> I think the result is the same as the preexisting code for earlier format major versions.

Coming back to this PR, for earlier format major versions (with this PR) one could have the above example, and then `[a#120,a#100]` would look more expensive than it is, because it does not account for `[a#80,a#60]` needing to become part of the same compaction and `[a#80,a#60]`looks cheaper than it is, since it sees no files below it (since they `outputIter` has already consumed them). With the code on master, both look cheaper since `base.InternalCompare` will not see any of the files under them. Is my understanding correct? If so, this change is fine, since both behaviors are not fully correct. But it could use a code comment and an example on why this behavior is correct when we don't split user keys (and if you are feeling up to it, a quick example the incorrectness that can happen when we do split user keys).
___
*[`testdata/compaction_picker_pick_file` line 57 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2566#-NWgrW8j8JWTZFZD5-2t:-NWgrW8kA2a6A9u1Usmj:b-irywtp) ([raw file](https://github.com/cockroachdb/pebble/blob/6718d765044d40e23d417439e2d46fedc34ddca0/testdata/compaction_picker_pick_file#L57)):*
> ```
> # Test the same scenario as above, but the larger file that overlaps the next
> # level only overlaps on its end boundary key ("c"). c.SET.11 sorts before
> # c.SET.0, but the files should still be considered overlapping.
> ```

[a#0,SET-c#0,SET] overlaps with [b#11,SET-c#11,SET] not just due to c. Did you mean to have the L6 file be something like [c#0,SET-d#0,SET]?
___
*[`testdata/compaction_picker_target_level` line 198 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2566#-NWE4tzk5bNmARcOgyZS:-NWh2PbI-zJ0w1wxd5Ry:b-3bpyhv) ([raw file](https://github.com/cockroachdb/pebble/blob/80f3263b52f2de96a1b2801ff8af25d37c944db8/testdata/compaction_picker_target_level#L198)):*

I found this test to be a bit inscrutable and am having trouble determining whether this is the correct output or not—anyone understand the mechanics of this test?
</blockquote></details>

My understanding is based on re-reading the test code. It could use more code comments for sure, and print the files that are compacting and their keys.
That may explain why it was not finding a compaction before and now it is (which is odd given that the only thing that has changed is the per file scoring).

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2566)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @sumeerbhola)

___
*[`compaction_picker.go` line 1112 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2566#-NWIOUNK2o_drFevG4WH:-NXBICroAjHrw0aEuq49:bok836o) ([raw file](https://github.com/cockroachdb/pebble/blob/4a21b62377313e2a7b80e3e891f38eb0e607760a/compaction_picker.go#L1112)):*
> With the code on master, both look cheaper since `base.InternalCompare` will not see any of the files under them. Is my understanding correct?

Yeah, that's my understanding as well.

Updated this comment, and included your example for why this logic is incorrect in the presence of split user keys.
___
*[`testdata/compaction_picker_pick_file` line 57 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2566#-NWgrW8j8JWTZFZD5-2t:-NXBQ3323jbYXH1mekba:b9iyuk5) ([raw file](https://github.com/cockroachdb/pebble/blob/6718d765044d40e23d417439e2d46fedc34ddca0/testdata/compaction_picker_pick_file#L57)):*

[a#0,SET-c#0,SET] overlaps with [b#11,SET-c#11,SET] not just due to c. Did you mean to have the L6 file be something like [c#0,SET-d#0,SET]?
</blockquote></details>

thanks, yeah
___
*[`testdata/compaction_picker_target_level` line 198 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2566#-NWE4tzk5bNmARcOgyZS:-NXCga-L9w7P-N9KjiSX:b-8zksdc) ([raw file](https://github.com/cockroachdb/pebble/blob/80f3263b52f2de96a1b2801ff8af25d37c944db8/testdata/compaction_picker_target_level#L198)):*

My understanding is based on re-reading the test code. It could use more code comments for sure, and print the files that are compacting and their keys.
That may explain why it was not finding a compaction before and now it is (which is odd given that the only thing that has changed is the per file scoring).
</blockquote></details>

Previously, the compaction picking logic would not pick a compaction from a level if the first file with the best ratio could not be picked due to an in-progress compaction. Even though it didn't pursue the compaction, it recorded the best ratio. If another file couldn't beat the ratio, (tie or worse), then no compaction was picked at all.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2566)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @sumeerbhola)

___
*[`testdata/compaction_picker_target_level` line 198 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2566#-NWE4tzk5bNmARcOgyZS:-NXCpLk035kFzwqUl1rn:bvwyy9n) ([raw file](https://github.com/cockroachdb/pebble/blob/80f3263b52f2de96a1b2801ff8af25d37c944db8/testdata/compaction_picker_target_level#L198)):*

Previously, the compaction picking logic would not pick a compaction from a level if the first file with the best ratio could not be picked due to an in-progress compaction. Even though it didn't pursue the compaction, it recorded the best ratio. If another file couldn't beat the ratio, (tie or worse), then no compaction was picked at all. 
</blockquote></details>

I was too hungry when I wrote this ^. Previously the picker would think that 500001 did not overlap 400001 and would pick 400001, only to discover during compaction expansion that it actually did overlap 500001 which is already compacting.

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 5 files at r4, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2566#-:-NXCxlRb80niH6WBN5i9:bg7igkv)* status: all files reviewed, 4 unresolved discussions (waiting on @jbowens)

jbowens:
tftr!

PR #2567, opened 2023-05-25
sharedcache: move to a dedicated package

joshimhoff:

**sharedcache: move to a dedicated package**

This commit introduces a package for the shared cache.


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2567)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @jbowens and @joshimhoff)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache_test.go` line 1 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2567#-NWITPKoBDE3PuMqVgNc:-NWJ0kje9SEDZ5LqvDDx:b-nqmdob) ([raw file](https://github.com/cockroachdb/pebble/blob/b66b5f340b9d83c53c0c9d258df8f4d983f55bf7/objstorage/objstorageprovider/sharedcache/shared_cache_test.go#L1)):*

I don't love having a test package like this, as it may mean making various things public just to test them from this package, as has already happeend. I guess I could implement a test `Readable` to avoid the need. Radu, let me know if you have thoughts.
</blockquote></details>

Hm yeah this is annoying. Implementing a very basic Readable for tests might be a good idea, the implementation could live in `objstorage`. We could even add logging to it which would be nicer than relying on the FS-level logging.

joshimhoff (commented):
TTFR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2567#-:-NWOFmA23-y0yYhZ51-0:b-80vssw)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache_test.go` line 1 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2567#-NWITPKoBDE3PuMqVgNc:-NWOFcvcAWjpcSPFdIB_:bpc8ezi) ([raw file](https://github.com/cockroachdb/pebble/blob/b66b5f340b9d83c53c0c9d258df8f4d983f55bf7/objstorage/objstorageprovider/sharedcache/shared_cache_test.go#L1)):*

Hm yeah this is annoying. Implementing a very basic Readable for tests might be a good idea, the implementation could live in `objstorage`. We could even add logging to it which would be nicer than relying on the FS-level logging.
</blockquote></details>

I'm gonna leave as is for now, based on the assumption that since you approved you don't object.

PR #2538, opened 2023-05-17
*: add IngestAndExcise operation

itsbilal:
This change adds an IngestAndExcise operation that does the below additional things alongside a regular ingestion:

1) It ingests some SharedSSTMeta files, which are provider-backed
   sstables that could be owned by other nodes.
2) It excises existing sstables within the provided excise span (within
   which all sstables from 1 must fit) by creating new virtual sstables
   that exclude keys from the excise span.

While this change can be implemented independently of #2455, some of the end-to-end tests in future changes will rely on both that and this.

Fixes #2520.

PR #2574, opened 2023-05-26
db: Add IngestShared test to exercise shared ingestion 

itsbilal:
This is a test-oriented change that adds more end-to-end testing for
ingesting shared sstables and also reading keys from them.

### DIFF
--- a/ingest_test.go
+++ b/ingest_test.go
@@ -741,6 +741,274 @@ func TestExcise(t *testing.T) {
 }
 
 func TestIngestShared(t *testing.T) {
+	var d, d1, d2 *DB
+	defer func() {
+		if d1 != nil {
+			require.NoError(t, d1.Close())
+		}
+		if d2 != nil {
+			require.NoError(t, d2.Close())
+		}
+	}()
+	creatorIDCounter := uint64(1)
+	replicateCounter := 1
+
+	reset := func() {
+		if d1 != nil {
+			require.NoError(t, d1.Close())
+		}
+		if d2 != nil {
+			require.NoError(t, d2.Close())
+		}
+
+		sstorage := shared.NewInMem()
+		mem1 := vfs.NewMem()
+		mem2 := vfs.NewMem()
+		require.NoError(t, mem1.MkdirAll("ext", 0755))
+		require.NoError(t, mem2.MkdirAll("ext", 0755))
+		opts1 := &Options{
+			Comparer:              testkeys.Comparer,
+			LBaseMaxBytes:         1,
+			FS:                    mem1,
+			L0CompactionThreshold: 100,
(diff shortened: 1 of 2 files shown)


itsbilal:
Rebased this now that #2538 has merged.

jbowens (approved):
:lgtm:

> Reviewed 1 of 16 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2574#-:-NXuTFSo8nBjw0aT9Svy:bnfp4nl)* status: 1 of 16 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`testdata/ingest_shared` line 482 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2574#-NXuSxcy6vZLvQsGOSyt:-NXuSxcy6vZLvQsGOSyu:bkmpr4e) ([raw file](https://github.com/cockroachdb/pebble/blob/454724611fbdac8fa3cec926f5fd7e61109ddd70/testdata/ingest_shared#L482)):*
> ```
>   000004:[a@3#10,SET-c@9#10,SET]
> 5:
>   000005:[b#2,RANGEDEL-d#2,RANGEDEL]
> ```

worth a test that exercises truncating the rangedel to the ingest sst bounds, ensuring it doesn't delete keys that fall within the physical sst bounds.

itsbilal (commented):
TFTR! Rebased to work with the newest changes.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2574#-:-NYKMg7S6xQVMngVa2RQ:b-iren2e)* status: 0 of 16 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`testdata/ingest_shared` line 482 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2574#-NXuSxcy6vZLvQsGOSyt:-NYKMh0c7khIu0esYGfD:bchq36y) ([raw file](https://github.com/cockroachdb/pebble/blob/454724611fbdac8fa3cec926f5fd7e61109ddd70/testdata/ingest_shared#L482)):*

worth a test that exercises truncating the rangedel to the ingest sst bounds, ensuring it doesn't delete keys that fall within the physical sst bounds.
</blockquote></details>

We actually don't support that, as we force virtual sstable bounds to fall in places where there are no range keys/deletes:

https://github.com/cockroachdb/pebble/blob/6746f4117fbc5da3587b74ee5572014b0ce1cf12/internal/keyspan/truncate.go#L69

But on closer thought, we might need to revisit this; seeing as we could defragment [a,b) and [b,c) as [a,c) even if b is a crdb range boundary and we'd need to be able to create virtual sstable bounds at b.

Issue #2562, opened 2023-05-24
github.com/cockroachdb/pebble/internal/metamorphic: TestMeta failed


nicktrav:
[Seen here.](https://teamcity.cockroachdb.com/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=10217641&_focus=107639)

Trace:

```
===== SEED =====
1684902326168487385
===== ERR =====
exit status 2
===== OUT =====
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x180 pc=0x986cbc]

goroutine 43 [running]:
github.com/cockroachdb/pebble.(*DB).FormatMajorVersion(0xc00049dba0?)
	github.com/cockroachdb/pebble/external/com_github_cockroachdb_pebble/format_major_version.go:373 +0x3c
github.com/cockroachdb/pebble/metamorphic.(*test).isFMV(...)
	github.com/cockroachdb/pebble/metamorphic/external/com_github_cockroachdb_pebble/metamorphic/test.go:178
github.com/cockroachdb/pebble/metamorphic.(*deleteOp).run(0xc0003c4320, 0xc0000fc2a0, {0xc0003c4480?, 0x0?})
	github.com/cockroachdb/pebble/metamorphic/external/com_github_cockroachdb_pebble/metamorphic/ops.go:195 +0x6b
github.com/cockroachdb/pebble/metamorphic.RunOnce.func1()
	github.com/cockroachdb/pebble/metamorphic/external/com_github_cockroachdb_pebble/metamorphic/meta.go:498 +0x179
golang.org/x/sync/errgroup.(*Group).Go.func1()
	golang.org/x/sync/errgroup/external/org_golang_x_sync/errgroup/errgroup.go:75 +0x64
created by golang.org/x/sync/errgroup.(*Group).Go
	golang.org/x/sync/errgroup/external/org_golang_x_sync/errgroup/errgroup.go:72 +0xa5
```

Repro:

```
$ go test -tags 'invariants' -exec 'stress -p 1' -timeout 0 -test.v -run TestMeta$ ./internal/metamorphic -seed 1684902326168487385 -ops "uniform:5000-10000"
```

nicktrav:
As part of this investigation, we should see why we didn't get a GH issue for this.

nicktrav:
@bananabrick - assigning to you as L2.

jbowens:
[...]

PR #2575, opened 2023-05-27
metamorphic: close batches before closing db

bananabrick:
In concurrent runs of the metamorphic tests, a closeOp for the DB can race with a deleteOp for the batch. But if we close the batch before closing the db, then it acts as a synchronization point and ensures that all batch operations complete before the db is closed.

Fixes: https://github.com/cockroachdb/pebble/issues/2562

### DIFF
--- a/metamorphic/generator.go
+++ b/metamorphic/generator.go
@@ -456,6 +456,11 @@ func (g *generator) dbClose() {
 	for len(g.liveSnapshots) > 0 {
 		g.snapshotClose()
 	}
+	for len(g.liveBatches) > 0 {
+		batchID := g.liveBatches[0]
+		g.batchClose(batchID)
+		g.add(&closeOp{objID: batchID})
+	}
 	g.add(&closeOp{objID: makeObjID(dbTag, 0)})
 }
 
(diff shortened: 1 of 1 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2575#-:-NWiaHYPBxmIvxCcNJmG:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @bananabrick)

___
*[`metamorphic/generator.go` line 539 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2575#-NWiXyUJ-IpLYvWrJA2R:-NWiXyUJ-IpLYvWrJA2S:bhu79lp) ([raw file](https://github.com/cockroachdb/pebble/blob/865607efb619c26564f79ed2de05d3a59e1645c1/metamorphic/generator.go#L539)):*
> ```Go
> 		batchID := g.liveBatches[0]
> 		g.batchClose(batchID)
> 		g.add(&closeOp{objID: batchID})
> ```

so previously we never actually called `Close` on the "closed" batches, we just discarded them? Should we put this `g.add(&closeOp{objID: batchID})` call in `g,batchClose`?

Issue #1683, opened 2022-05-04
db: split physical sstables into virtual sstables at ingest time


sumeerbhola:
This issue was originally about the concept of virtual sstables in general, as well as the ingestion-time uses of it. Currently, this issue is about the creation of virtual sstables (which were implemented in #2288 , #2352, and other PRs)  from physical sstables at ingest time, when it is beneficial to do so to slot ingested sstables at lower levels. The original description of the issue, as well as the detailed technical description is retained below the horizontal line.

See #2336 for the former meta-issue describing different parts of this project.

----

Virtual sstables separate the logical key bounds of an sstable from the physical key bounds of the underlying sstable. They allow multiple virtual ssts to share the same underlying physical sst. Virtual ssts are part of the disaggregated shared storage design https://github.com/cockroachdb/cockroach/pull/70419 since they allow us to have ssts that conform to CockroachDB range bounds in lower levels of the LSM.
However, they can also be beneficial in the current Pebble setup which does not share ssts across Pebble instances: in a customer issue https://github.com/cockroachlabs/support/issues/1558 we noticed that a significant number of ssts were being ingested into L0, which caused Pebble overload. These ssts were due to CockroachDB snapshots (see https://github.com/cockroachdb/cockroach/issues/80589 and https://github.com/cockroachdb/cockroach/issues/80607). A reproduction revealed that snapshot ingestion has no data overlap preventing these from being ingested into L6, and all the overlap occurs due to file key bounds.

By introducing virtual ssts we can improve the level assignment for ingestion by:
- finding the lowest level with data overlap, say i.
- ingest into i-1. If there is a file boundary overlap, split the sst into 2 virtual ssts, leaving a gap between them that the ingested sst can fit into. The VersionEdit for the ingest would also include this split.
- At most one sst is split per ingested sst. We can reduce the risk of many small virtual ssts by:
  - For CockroachDB's snapshot ingestion, there is one large sst (up to 512MB) and many tiny ones. We can choose the apply this splitting logic only for the large sst. It is ok for the tiny ssts to be ingested into L0.
  - Split only if the ingested sst is at least half the size of the sst being split. So if we have a smaller ingested sst, we will pick a higher level to split at (where the ssts are smaller). The lifetime of virtual ssts at a higher level is smaller, so there is lower risk of littering the LSM with long-lived small virtual ssts.

Effect on DB state data-structures (Version, FileMetadata, VersionEdit ...)
[...]

PR #2582, opened 2023-05-31
db: add ingest-time splitting of ssts into virtual ones

itsbilal:
Currently, if we identify boundary overlap in a level
during ingest target level calculation, but no data overlap,
we are forced to find a target level above the file we saw
the overlap with (if we can't fall below it, such as if the
existing file is in L6, which happens commonly).

This change takes advantage of virtual sstables to split
existing sstables into two virtual sstables when an ingested
sstable would be able to go into the same level had the sstables
been split that way to begin with. Doing this split reduces a
lot of write-amp as it avoids us from having to compact the
newly-ingested sstable with the sstable it boundary-overlapped with.

Biggest part of https://github.com/cockroachdb/pebble/issues/1683. First commit is https://github.com/cockroachdb/pebble/pull/2538, which this shares
a lot of logic with (mostly just the excise function).

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1905,15 +1905,27 @@ func (d *DB) runIngestFlush(c *compaction) (*manifest.VersionEdit, error) {
 	ve := &versionEdit{}
 	var level int
 	var err error
+	var fileToSplit *fileMetadata
+	var ingestSplitFiles []ingestSplitFile
 	for _, file := range c.flushing[0].flushable.(*ingestedFlushable).files {
-		level, err = ingestTargetLevel(
+		suggestSplit := d.opts.Experimental.IngestSplit != nil && d.opts.Experimental.IngestSplit() &&
+			d.FormatMajorVersion() >= FormatVirtualSSTables
+		level, fileToSplit, err = ingestTargetLevel(
 			d.newIters, d.tableNewRangeKeyIter, iterOpts, d.cmp,
 			c.version, baseLevel, d.mu.compact.inProgress, file.FileMetadata,
+			suggestSplit,
 		)
 		if err != nil {
 			return nil, err
 		}
 		ve.NewFiles = append(ve.NewFiles, newFileEntry{Level: level, Meta: file.FileMetadata})
+		if fileToSplit != nil {
+			ingestSplitFiles = append(ingestSplitFiles, ingestSplitFile{
+				ingestFile: file.FileMetadata,
+				splitFile:  fileToSplit,
+				level:      level,
+			})
+		}
 		levelMetrics := c.metrics[level]
 		if levelMetrics == nil {
 			levelMetrics = &LevelMetrics{}
@@ -1921,6 +1933,28 @@ func (d *DB) runIngestFlush(c *compaction) (*manifest.VersionEdit, error) {
 		}
 		levelMetrics.BytesIngested += file.Size
 		levelMetrics.TablesIngested++
+	}
+
+	updateLevelMetricsOnExcise := func(m *fileMetadata, level int, added []newFileEntry) {
+		levelMetrics := c.metrics[level]
+		if levelMetrics == nil {
+			levelMetrics = &LevelMetrics{}
+			c.metrics[level] = levelMetrics
+		}
+		levelMetrics.NumFiles--
+		levelMetrics.Size -= int64(m.Size)
+		for i := range added {
+			levelMetrics.NumFiles++
+			levelMetrics.Size += int64(added[i].Meta.Size)
+		}
+	}
+
+	if len(ingestSplitFiles) > 0 {
+		ve.DeletedFiles = make(map[manifest.DeletedFileEntry]*manifest.FileMetadata)
+		replacedFiles := make(map[base.FileNum][]newFileEntry)
+		if err := d.ingestSplit(ve, updateLevelMetricsOnExcise, ingestSplitFiles, replacedFiles); err != nil {
+			return nil, err
+		}
 	}
 
 	return ve, nil
@@ -2091,6 +2125,24 @@ func (d *DB) flush1() (bytesFlushed uint64, err error) {
 				metrics := c.metrics[0]
 				for i := 0; i < n; i++ {
 					metrics.BytesIn += d.mu.mem.queue[i].logSize
+				}
+			}
+		} else if len(ve.DeletedFiles) > 0 {
+			// c.kind == compactionKindIngestedFlushable && we have deleted files due
+			// to ingest-time splits.
+			//
+			// Iterate through all other compactions, and check if their inputs have
+			// been replaced due to an ingest-time split. In that case, cancel the
+			// compaction.
+			for c2 := range d.mu.compact.inProgress {
+				for i := range c2.inputs {
+					iter := c2.inputs[i].files.Iter()
+					for f := iter.First(); f != nil; f = iter.Next() {
+						if _, ok := ve.DeletedFiles[deletedFileEntry{FileNum: f.FileNum, Level: c2.inputs[i].level}]; ok {
+							c2.cancel.Store(true)
+							break
+						}
+					}
 				}
 			}
 		}
--- a/data_test.go
+++ b/data_test.go
@@ -1356,8 +1356,9 @@ func runForceIngestCmd(td *datadriven.TestData, d *DB) error {
 		int,
 		map[*compaction]struct{},
 		*fileMetadata,
-	) (int, error) {
-		return level, nil
+		bool,
+	) (int, *fileMetadata, error) {
+		return level, nil, nil
 	}, nil /* shared */, KeyRange{}, nil /* external */)
 	return err
 }
--- a/ingest.go
+++ b/ingest.go
@@ -797,6 +797,10 @@ func overlapWithIterator(
 	return computeOverlapWithSpans(*rangeDelIter)
 }
 
+// ingestTargetLevel returns the target level for a file being ingested.
+// If suggestSplit is true, it accounts for ingest-time splitting as part of
+// its target level calculation, and if a split candidate is found, that file
+// is returned as the splitFile.
 func ingestTargetLevel(
 	newIters tableNewIters,
 	newRangeKeyIter keyspan.TableNewSpanIter,
@@ -806,7 +810,8 @@ func ingestTargetLevel(
 	baseLevel int,
 	compactions map[*compaction]struct{},
 	meta *fileMetadata,
-) (int, error) {
+	suggestSplit bool,
+) (targetLevel int, splitFile *fileMetadata, err error) {
 	// Find the lowest level which does not have any files which overlap meta. We
 	// search from L0 to L6 looking for whether there are any files in the level
 	// which overlap meta. We want the "lowest" level (where lower means
@@ -821,6 +826,14 @@ func ingestTargetLevel(
 	//   violate the sequence number invariant.
 	// - no file boundary overlap with level i, since that will violate the
 	//   invariant that files do not overlap in levels i > 0.
+	//   - if there is only a file overlap at a given level, and no data overlap,
+	//     we can still slot a file at that level. We return the fileMetadata with
+	//     which we have file boundary overlap (must be only one file, as sstable
+	//     bounds are usually tight on user keys) and the caller is expected to split
+	//     that sstable into two virtual sstables, allowing this file to go into that
+	//     level. Note that if we have file boundary overlap with two files, which
+	//     should only happen on rare occasions, we treat it as data overlap and
+	//     don't use this optimization.
 	//
 	// The file boundary overlap check is simpler to conceptualize. Consider the
 	// following example, in which the ingested file lies completely before or
@@ -865,12 +878,10 @@ func ingestTargetLevel(
 	// existing point that falls within the ingested table bounds as being "data
 	// overlap".
 
-	targetLevel := 0
-
 	// This assertion implicitly checks that we have the current version of
 	// the metadata.
 	if v.L0Sublevels == nil {
-		return 0, errors.AssertionFailedf("could not read L0 sublevels")
+		return 0, nil, errors.AssertionFailedf("could not read L0 sublevels")
 	}
 	// Check for overlap over the keys of L0 by iterating over the sublevels.
 	for subLevel := 0; subLevel < len(v.L0SublevelFiles); subLevel++ {
@@ -896,10 +907,10 @@ func ingestTargetLevel(
 		err := iter.Close() // Closes range del iter as well.
 		err = firstError(err, levelIter.Close())
 		if err != nil {
-			return 0, err
+			return 0, nil, err
 		}
 		if overlap {
-			return targetLevel, nil
+			return targetLevel, nil, nil
 		}
 	}
 
@@ -926,26 +937,47 @@ func ingestTargetLevel(
 		err := levelIter.Close() // Closes range del iter as well.
 		err = firstError(err, rkeyLevelIter.Close())
 		if err != nil {
-			return 0, err
+			return 0, nil, err
 		}
 		if overlap {
-			return targetLevel, nil
+			return targetLevel, splitFile, nil
 		}
 
 		// Check boundary overlap.
+		var candidateSplitFile *fileMetadata
 		boundaryOverlaps := v.Overlaps(level, cmp, meta.Smallest.UserKey,
 			meta.Largest.UserKey, meta.Largest.IsExclusiveSentinel())
 		if !boundaryOverlaps.Empty() {
-			continue
+			// We are already guaranteed to not have any data overlaps with files
+			// in boundaryOverlaps, otherwise we'd have returned in the above if
+			// statements. Use this, plus boundaryOverlaps.Len() == 1 to detect for
+			// the case where we can slot this file into the current level despite
+			// a boundary overlap, by splitting one existing file into two virtual
+			// sstables.
+			if suggestSplit && boundaryOverlaps.Len() == 1 {
+				iter := boundaryOverlaps.Iter()
+				candidateSplitFile = iter.First()
+			} else {
+				// We either don't want to suggest ingest-time splits (i.e.
+				// !suggestSplit), or we boundary-overlapped with more than one file.
+				continue
+			}
 		}
 
-		// Check boundary overlap with any ongoing compactions.
+		// Check boundary overlap with any ongoing compactions. We consider an
+		// overlapping compaction that's writing files to an output level as
+		// equivalent to boundary overlap with files in that output level.
 		//
-		// We cannot check for data overlap with the new SSTs compaction will
-		// produce since compaction hasn't been done yet. However, there's no need
-		// to check since all keys in them will either be from c.startLevel or
-		// c.outputLevel, both levels having their data overlap already tested
-		// negative (else we'd have returned earlier).
+		// We cannot check for data overlap with the new SSTs compaction will produce
+		// since compaction hasn't been done yet. However, there's no need to check
+		// since all keys in them will be from levels in [c.startLevel,
+		// c.outputLevel], and all those levels have already had their data overlap
+		// tested negative (else we'd have returned earlier).
+		//
+		// An alternative approach would be to cancel these compactions and proceed
+		// with an ingest-time split on this level if necessary. However, compaction
+		// cancellation can result in significant wasted effort and is best avoided
+		// unless necessary.
 		overlaps := false
 		for c := range compactions {
 			if c.outputLevel == nil || level != c.outputLevel.level {
@@ -959,9 +991,10 @@ func ingestTargetLevel(
 		}
 		if !overlaps {
 			targetLevel = level
+			splitFile = candidateSplitFile
 		}
 	}
-	return targetLevel, nil
+	return targetLevel, splitFile, nil
 }
 
 // Ingest ingests a set of sstables into the DB. Ingestion of the files is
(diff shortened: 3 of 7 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2582)* status: 0 of 18 files reviewed, 5 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`ingest.go` line 617 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWsp_3r-o3NwmV5Xh2w:-NWsp_3r-o3NwmV5Xh2x:b-bpqz82) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L617)):*
> ```Go
> 	meta *fileMetadata,
> 	suggestSplit bool,
> ) (int, *fileMetadata, error) {
> ```

[nit] name the return values and add a function level comment
___
*[`ingest.go` line 1578 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWspnCf8e2XueXuhar3:-NWspnCf8e2XueXuhar4:b-5e0r2a) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L1578)):*
> ```Go
> 	// ingestFile is the file being ingested, and splitFile is the file that
> 	// needs to be split to allow ingestFile to slot into `level` level.
> 	ingestFile, splitFile *fileMetadata
> ```

[nit] separate the fields and comments
___
*[`ingest.go` line 1599 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWsq-DTDdxqjqyGDO66:-NWsq-DTDdxqjqyGDO67:b-xv03vh) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L1599)):*
> ```Go
> ) error {
> 	// Note the opposite order: Unlock() followed by Lock().
> 	d.mu.Unlock()
> ```

I'm not sure this makes sense in here, we should move to the calling code instead
___
*[`ingest.go` line 1616 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWsqex91qG-5ufcNK-v:-NWsqex91qG-5ufcNK-w:bvhtqz4) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L1616)):*
> ```Go
> 			for i := range replaced {
> 				if replaced[i].Meta.Overlaps(d.cmp, s.ingestFile.Smallest.UserKey, s.ingestFile.Largest.UserKey, s.ingestFile.Largest.IsExclusiveSentinel()) {
> 					splitFile = replaced[i].Meta
> ```

We expect this to be the case at most once, right? We can assert that `updatedSplitFile` is false.
___
*[`ingest.go` line 1652 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWsr-i44p0r56nNJgxs:-NWsr-i50WQXwz2d2BGe:b-ot3psu) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L1652)):*
> ```Go
> 		replacedFiles[splitFile.FileNum] = added
> 		for i := range added {
> 			if s.ingestFile.Overlaps(d.cmp, added[i].Meta.Smallest.UserKey, added[i].Meta.Largest.UserKey, added[i].Meta.Largest.IsExclusiveSentinel()) {
> ```

Nice sanity checks!

sumeerbhola (changes requested):
flushing some preliminary comments

> Reviewed 2 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2582#-:-NWt2vYvA9ANGy7a571a:bge0iq5)* status: 2 of 18 files reviewed, 11 unresolved discussions (waiting on @itsbilal)

___
*[`compaction.go` line 1861 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWsyw_Q5SnxmnpVwZkk:-NWsyw_Q5SnxmnpVwZkl:b-j0pw0b) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/compaction.go#L1861)):*
> ```Go
> 			d.newIters, d.tableNewRangeKeyIter, iterOpts, d.cmp,
> 			c.version, baseLevel, d.mu.compact.inProgress, file.FileMetadata,
> 			false, /* suggestSplit */
> ```

why is suggestSplit false in this case. Don't we want to ingest as low as possible in this case too?
___
*[`ingest.go` line 769 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWszhNJCnlP8o1hdKSG:-NWszhNJCnlP8o1hdKSH:bpef1wu) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L769)):*
> ```Go
> 				// latter should never happen today, but possibly some future version has
> 				// started writing FileMetadatas with key bounds that are loose on user
> 				// keys. Handle this defensively.
> ```

Can we log a warning here with the file bound details?
___
*[`ingest.go` line 1591 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWt29SJAOkY1_iW0HgN:-NWt29SJAOkY1_iW0HgO:bk2zc0m) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L1591)):*
> ```Go
> // unconditionally released and then re-acquired during this method as it does
> // IO.
> func (d *DB) ingestSplit(
> ```

(note to self) read this method
___
*[`ingest.go` line 1699 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWt10gHD7sNiI8JGpFQ:-NWt10gHD7sNiI8JGpFR:btlq32d) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L1699)):*
> ```Go
> 	baseLevel := d.mu.versions.picker.getBaseLevel()
> 	iterOps := IterOptions{logger: d.opts.Logger}
> 	// filesToSplit is a map of files being ingested -> files being split.
> ```

nit: it's not quite a map, though I realize this comment was intending to be more abstract.
We could say something more verbose like:
// filesToSplit is a list where each element is a pair consisting of a file being ingested and an existing file that needs to be split to make room for it. Each ingested file will appear at most once in this list.
___
*[`ingest.go` line 1774 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWt1lu00D63DdSzLZAa:-NWt1lu1-NUNI-8P6dUY:b-7i50ap) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L1774)):*
> ```Go
> 	// replacedFiles maps files excised due to exciseSpan (or splitFiles returned
> 	// by ingestTargetLevel), to files that were created to replace it. This map
> 	//
> ```

incomplete sentence.
Why is this a map in that a file that is split is replaced with multiple files?
___
*[`options.go` line 527 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWt2Ffk2DEqkBYwdLii:-NWt2Ffk2DEqkBYwdLij:b-as1ozr) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/options.go#L527)):*
> ```Go
> 		// sstables into two virtual sstables to allow ingestion sstables to slot
> 		// into a lower level than they otherwise would have.
> 		IngestSplit bool
> ```

to be on the safe side, I think we should make this dynamically changeable in a running cluster.

btw, do we have a metric counting the unique backing ssts? We should be able to see if such splitting is causing the count of filemetadata / count of backing ssts to become high.

itsbilal (commented):
TFTR!

Should have addressed all the comments so far, and should have also addressed my own concerns around how this interacted with in-progress compactions. The cancel cases for compactions are getting more complex, so I'll take a stab at a unit test that better tests the cancel cases tomorrow. Otherwise this should be good to review.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2582#-:-NbWELFx7AfV5eg-oYqU:b-xiuddc)* status: 0 of 18 files reviewed, 5 unresolved discussions (waiting on @sumeerbhola)

___
*[`compaction.go` line 1861 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWsyw_Q5SnxmnpVwZkk:-NbWEdwWDmYK7103nq4a:bgezfy5) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/compaction.go#L1861)):*

why is suggestSplit false in this case. Don't we want to ingest as low as possible in this case too?
</blockquote></details>

Done. This was because this code didn't correctly handle metrics and cancel cases for ingest-time splitting, but now it does.
___
*[`ingest.go` line 769 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWszhNJCnlP8o1hdKSG:-NbWEkooAF1MdnFiNvhn:binzy74) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L769)):*

Can we log a warning here with the file bound details?
</blockquote></details>

Done. Rather N/A because we now do loose-file-bound virtual sstables (for online restore). Procrastination can have pros like that!
___
*[`ingest.go` line 1578 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWspnCf8e2XueXuhar3:-NbWFk_P11Ovi8LT_u6F:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L1578)):*

[nit] separate the fields and comments
</blockquote></details>

Done.
___
*[`ingest.go` line 1599 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWsq-DTDdxqjqyGDO66:-NbWFmTl7rttgmNo4qg3:b3m6o01) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L1599)):*

I'm not sure this makes sense in here, we should move to the calling code instead
</blockquote></details>

Done. (Removed for the same reason as this was removed in the `IngestAndExcise` PR - it messes with the view of in-progress compactions).
___
*[`ingest.go` line 1616 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWsqex91qG-5ufcNK-v:-NbWJNNf7BthQ9mawaJI:b-7b162z) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L1616)):*

We expect this to be the case at most once, right? We can assert that `updatedSplitFile` is false.
</blockquote></details>

Yes. Asserted.
___
*[`ingest.go` line 1699 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWt10gHD7sNiI8JGpFQ:-NbWFsEN6iYE2ECv5DEK:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L1699)):*

nit: it's not quite a map, though I realize this comment was intending to be more abstract.
We could say something more verbose like:
// filesToSplit is a list where each element is a pair consisting of a file being ingested and an existing file that needs to be split to make room for it. Each ingested file will appear at most once in this list.
</blockquote></details>

Done.
___
*[`ingest.go` line 1774 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWt1lu00D63DdSzLZAa:-NbWFtam5UctghkwTaen:b-o5u24o) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/ingest.go#L1774)):*

incomplete sentence.
Why is this a map in that a file that is split is replaced with multiple files?
</blockquote></details>

Clarified in the comments. It's because the entries are FileMetadatas, and we create two FileMetadatas every time we split a file.
___
*[`options.go` line 527 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NWt2Ffk2DEqkBYwdLii:-NbWJ_bD9gjxYWuBMNMM:btx9pou) ([raw file](https://github.com/cockroachdb/pebble/blob/ee0931ed618626fb1227478b7a07dd5e11674cc4/options.go#L527)):*

to be on the safe side, I think we should make this dynamically changeable in a running cluster.

btw, do we have a metric counting the unique backing ssts? We should be able to see if such splitting is causing the count of filemetadata / count of backing ssts to become high.
</blockquote></details>

Done.

We don't have a metric right now, but we probably should have one to be able to track stuff like this - this is a good call. Will file an issue shortly.

sumeerbhola (changes requested):
> Reviewed 1 of 7 files at r2, 7 of 17 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2582)* status: 8 of 18 files reviewed, 14 unresolved discussions (waiting on @itsbilal)

___
*[`-- commits` line 17 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-Nbtucdk7NfN6FMutKbI):*
nit: can you update this slightly stale comment
___
*[`ingest.go` line 807 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-Nbu9xroCeufQe25UHNT:-Nbu9xroCeufQe25UHNU:bi3obja) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L807)):*
> ```Go
> 	//   - if there is only a file overlap at a given level, and no data overlap,
> 	//     we can still slot a file at that level. We return the fileMetadata with
> 	//     which we have file boundary overlap (must be only one file by definition, as
> ```

is this "must be only one file by definition" stale due to loose online restore sst bounds?
___
*[`ingest.go` line 921 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-Nbu8gMo9UqkObCtU_Ux:-Nbu8gMo9UqkObCtU_Uy:bmlkw82) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L921)):*
> ```Go
> 
> 		// Check boundary overlap.
> 		splitFile = nil
> ```

why are we setting this to nil here -- if we had advanced `targetLevel` in a previous loop iteration, we don't want to lose that in case compactions (or something else) prevent us from advancing it here? I think we need to maintain consistency via setting `splitFile` exactly where we set `targetLevel`, otherwise it becomes to hard to reason about.

`candidateSplitFile` is nil at this point, then the following blocks will try to set and unset `candidateSplitFile` and if we eventually fall through we would do
```
		if !overlaps {
			targetLevel = level
			splitFile = candidateSplitFile
		}
```
___
*[`ingest.go` line 926 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-Nbu8GJA5oLBaNfSS3OR:-Nbu8GJA5oLBaNfSS3OS:b-gbsekw) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L926)):*
> ```Go
> 		if !boundaryOverlaps.Empty() {
> 			// We are already guaranteed to not have any data overlaps with files
> 			// in boundaryOverlaps, otherwise we'd have gone into the above if
> ```

nit: otherwise we'd have returned in the above if ...
___
*[`ingest.go` line 1870 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NbuK9Mx7FRnrg7zTxDN:-NbuK9Mx7FRnrg7zTxDO:bqfgvlz) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L1870)):*
> ```Go
> 				if replaced[i].Meta.Overlaps(d.cmp, s.ingestFile.Smallest.UserKey, s.ingestFile.Largest.UserKey, s.ingestFile.Largest.IsExclusiveSentinel()) {
> 					if updatedSplitFile {
> 						panic("updated with two files in ingestSplit")
> ```

This could use a comment with a visual example. I believe this is an invariant because the various ingested files in `files` are non-overlapping.
___
*[`ingest.go` line 1879 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NbuKP6v5FxEFCoCCeiu:-NbuKP6v5FxEFCoCCeiv:b-4ao3yn) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L1879)):*
> ```Go
> 				// None of the replaced files overlapped with the file being ingested.
> 				// This is only okay if we have an excise span that overlaps with
> 				// the ingested file.
> ```

I am not sure this is limited to the excise case. If we had a file in the version that had a span [a,z) and one ingested file at [b,c) caused it to split into [a,b) and [j,z) (because the file had no keys in [b,j)), then a second ingested file with span [d,e) can slide into the gap that was left.
___
*[`ingest.go` line 1895 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NbudzY48PStjfraSq2V:-NbudzY48PStjfraSq2W:b-h9xbhn) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L1895)):*
> ```Go
> 		// as we're guaranteed to not have any data overlap between splitFile and
> 		// s.ingestFile, so panic if we do see a newly added file with an endKey
> 		// equalling s.ingestFile.Largest, and !s.ingestFile.Largest.IsExclusiveSentinel()
> ```

I am getting a bit confused by how this tree structure gets reflected in the versionEdit. My understanding is that we can have something like:
file1 (from the existing version) is split into file11, file12. Then file12 is split into file121, file122. `replacedFiles` will have file1, file12 as keys. This is so that we can traverse down the tree when someone else tries to split file1, and we need to effectively split file121. This all make sense.

`d.excise` will add file1 to `ve.DeletedFiles` and then we are adding it again? It will also add file11, file12 to `ve.NewFiles`. Then later we will add file12 to `ve.DeletedFiles` and file121, file122 to `ve.NewFiles`. The versionEdit documentation says:
```
// A file num may be present in both deleted files and new files when it  
// is moved from a lower level to a higher level (when the compaction  
// found that there was no overlapping file at the higher level).
```
Perhaps this is working out ok, because the code happens to work, even though file12 was never in the version?
Or we could flatten this tree after all the updates to `replacedFiles` are done, and at that time update the versionEdit.

Do we have tests exercising this repeated splitting (I didn't see one in testdata/ingest)?
___
*[`ingest.go` line 2016 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NbumcupDMSPQ3C0wc9a:-NbumcupDMSPQ3C0wc9b:bimauxm) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L2016)):*
> ```Go
> 				// check from findTargetLevel, as that requires d.mu to be held.
> 				shouldIngestSplit := d.opts.Experimental.IngestSplit != nil &&
> 					d.opts.Experimental.IngestSplit() && d.FormatMajorVersion() >= ExperimentalFormatVirtualSSTables
> ```

I think we should hoist this to the top of the function, and not do this in every iteration. It is simpler that way, and we don't need the setting change to apply in the middle of an ingest. 
___
*[`ingest.go` line 2094 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NbunWSiBg8hLZvSJ612:-NbunWSiBg8hLZvSJ613:b-7vxsyf) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L2094)):*
> ```Go
> 
> 			for m := iter.First(); m != nil; m = iter.Next() {
> 				excised, err := d.excise(exciseSpan, m, ve, level)
> ```

nit: excised is not a good name for a slice containing `NewFileEntry`.

itsbilal (commented):
TFTR!

This should be ready for another look; I've rebased and addressed known issues so far. I'll keep seeing if there are more ways to harden it up or improve testing, but from my perspective this is looking good again.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2582#-:-NejX3SM0PhMcQ5SOwUv:bv2vghx)* status: 2 of 18 files reviewed, 11 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`-- commits` line 17 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-Nbtucdk7NfN6FMutKbI):*

nit: can you update this slightly stale comment
</blockquote></details>

Done.
___
*[`ingest.go` line 807 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-Nbu9xroCeufQe25UHNT:-Nei3gAwA2x19-ZGqxZV:b-87t4y7) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L807)):*

is this "must be only one file by definition" stale due to loose online restore sst bounds?
</blockquote></details>

Yep. Clarified that now.
___
*[`ingest.go` line 921 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-Nbu8gMo9UqkObCtU_Ux:-Nei8KKL6v6bhPLYLHQd:b-ol66ky) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L921)):*

why are we setting this to nil here -- if we had advanced `targetLevel` in a previous loop iteration, we don't want to lose that in case compactions (or something else) prevent us from advancing it here? I think we need to maintain consistency via setting `splitFile` exactly where we set `targetLevel`, otherwise it becomes to hard to reason about.

`candidateSplitFile` is nil at this point, then the following blocks will try to set and unset `candidateSplitFile` and if we eventually fall through we would do
```
		if !overlaps {
			targetLevel = level
			splitFile = candidateSplitFile
		}
```
</blockquote></details>

Good catch! Done. I added a `candidateSplitFile` for this purpose.
___
*[`ingest.go` line 1870 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NbuK9Mx7FRnrg7zTxDN:-NejQ_yABMaFntEZThgv:bmmw69z) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L1870)):*

This could use a comment with a visual example. I believe this is an invariant because the various ingested files in `files` are non-overlapping.
</blockquote></details>

I added a longer comment and referenced the visual examples in `ingestTargetLevel`, hope that's sufficient?
___
*[`ingest.go` line 1879 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NbuKP6v5FxEFCoCCeiu:-NejQPR6Ez_3r5JgZjL3:bp23k7q) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L1879)):*

I am not sure this is limited to the excise case. If we had a file in the version that had a span [a,z) and one ingested file at [b,c) caused it to split into [a,b) and [j,z) (because the file had no keys in [b,j)), then a second ingested file with span [d,e) can slide into the gap that was left.
</blockquote></details>

Done. Good catch; adding a test case with this, thanks!
___
*[`ingest.go` line 1895 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NbudzY48PStjfraSq2V:-NejQSiC6g9sQT7_sHq7:b3xi6k5) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L1895)):*

I am getting a bit confused by how this tree structure gets reflected in the versionEdit. My understanding is that we can have something like:
file1 (from the existing version) is split into file11, file12. Then file12 is split into file121, file122. `replacedFiles` will have file1, file12 as keys. This is so that we can traverse down the tree when someone else tries to split file1, and we need to effectively split file121. This all make sense.

`d.excise` will add file1 to `ve.DeletedFiles` and then we are adding it again? It will also add file11, file12 to `ve.NewFiles`. Then later we will add file12 to `ve.DeletedFiles` and file121, file122 to `ve.NewFiles`. The versionEdit documentation says:
```
// A file num may be present in both deleted files and new files when it  
// is moved from a lower level to a higher level (when the compaction  
// found that there was no overlapping file at the higher level).
```
Perhaps this is working out ok, because the code happens to work, even though file12 was never in the version?
Or we could flatten this tree after all the updates to `replacedFiles` are done, and at that time update the versionEdit.

Do we have tests exercising this repeated splitting (I didn't see one in testdata/ingest)?
</blockquote></details>

Very good catch. I did the flattening in `ingestSplit`. Also added tests for these multi-split cases that actually traverse the tree.
___
*[`ingest.go` line 2016 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NbumcupDMSPQ3C0wc9a:-NeiDDiHANjLaXWOZ1K4:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L2016)):*

I think we should hoist this to the top of the function, and not do this in every iteration. It is simpler that way, and we don't need the setting change to apply in the middle of an ingest.
</blockquote></details>

Done.
___
*[`ingest.go` line 2094 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NbunWSiBg8hLZvSJ612:-NeiDEcI2Min6GXIuxAu:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/d5eaf3c1d33504dffe507937346a4cc2da716c56/ingest.go#L2094)):*

nit: excised is not a good name for a slice containing `NewFileEntry`.
</blockquote></details>

Done.

sumeerbhola (approved):
:lgtm: 

> Reviewed 10 of 10 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2582#-:-NfDZiQ4D4_QfUpF8wS6:bg7igkv)* status: 12 of 18 files reviewed, 15 unresolved discussions (waiting on @itsbilal)

___
*[`ingest.go` line 955 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NfDRXEyDjxeTWjjA6SF:-NfDRXEyDjxeTWjjA6SG:b8convb) ([raw file](https://github.com/cockroachdb/pebble/blob/c54bb3ad137206d2be5b5684460dd95e36e21ca9/ingest.go#L955)):*
> ```Go
> 		// produce since compaction hasn't been done yet. However, there's no need
> 		// to check since all keys in them will either be from c.startLevel or
> 		// c.outputLevel, both levels having their data overlap already tested
> ```

since we now have multi-level compactions, consider saying something like:

... be from levels in [c.startLevel, c.outputLevel], and all those levels have their data overlap ...
___
*[`ingest.go` line 966 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NfDS0RLDgfnOCrRjoMs:-NfDS0RMExRm61Y7vLNv:b-9mmdlh) ([raw file](https://github.com/cockroachdb/pebble/blob/c54bb3ad137206d2be5b5684460dd95e36e21ca9/ingest.go#L966)):*
> ```Go
> 				overlaps = true
> 				break
> 			}
> ```

I'm forgetting something now: are we doing this check since we would prefer to not have to cancel compactions? In that if we didn't do this here and ingested, we would be ok?
I guess the answer is no -- these compactions could finish before the ingest and then this ingest would be incorrect since it could try to add a file that overlaps with others in the level.
And the check we do in terms of canceling compactions when installing the version edit for the ingest is for compactions that started after the ingest decision was made, yes?

The code could use some more commentary on this.
___
*[`ingest.go` line 1861 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NfDTzr13IkK0txyePbJ:-NfDTzr2Dj8MX_JN6P2I:b73atr9) ([raw file](https://github.com/cockroachdb/pebble/blob/c54bb3ad137206d2be5b5684460dd95e36e21ca9/ingest.go#L1861)):*
> ```Go
> 		// replacedFiles can be thought of as a tree, where we start iterating with
> 		// s.splitFile and run its fileNum through replacedFiles, then find which
> 		// of the replaced files overlaps with s.ingestFile, check its
> ```

I think this should say something about splitFile, since its mentioned at the end of the comment. Something like:

...overlaps with s.ingestFile, which becomes the new splitFile, check splitFile's replacements ...
___
*[`ingest.go` line 1863 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NfDUIPQ4oNjIqijTtrh:-NfDUIPQ4oNjIqijTtri:b-5cs9n7) ([raw file](https://github.com/cockroachdb/pebble/blob/c54bb3ad137206d2be5b5684460dd95e36e21ca9/ingest.go#L1863)):*
> ```Go
> 		// of the replaced files overlaps with s.ingestFile, check its
> 		// replacements in replacedFiles again for overlap with s.ingestFile, and
> 		// so on until we can't find the current splitFile in replacedFiles.
> ```

how about expanding this comment like:

... so on until either we can't find the current splitFile in replacedFiles (which means the splitFile should be split) or we don't find a file that overlaps with s.ingestFile (which means another file being ingested already split things enough to accommodate s.ingestFile).

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2582#-:-NfHkT-494p9rwyHhV7S:b-814g8k)* status: 5 of 18 files reviewed, 15 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`ingest.go` line 955 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NfDRXEyDjxeTWjjA6SF:-NfHhL84CzEpKnECSDOy:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c54bb3ad137206d2be5b5684460dd95e36e21ca9/ingest.go#L955)):*

since we now have multi-level compactions, consider saying something like:

... be from levels in [c.startLevel, c.outputLevel], and all those levels have their data overlap ...
</blockquote></details>

Done.
___
*[`ingest.go` line 966 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NfDS0RLDgfnOCrRjoMs:-NfHjCO85c3bJQQMOq-M:bbt8qyo) ([raw file](https://github.com/cockroachdb/pebble/blob/c54bb3ad137206d2be5b5684460dd95e36e21ca9/ingest.go#L966)):*

I'm forgetting something now: are we doing this check since we would prefer to not have to cancel compactions? In that if we didn't do this here and ingested, we would be ok?
I guess the answer is no -- these compactions could finish before the ingest and then this ingest would be incorrect since it could try to add a file that overlaps with others in the level.
And the check we do in terms of canceling compactions when installing the version edit for the ingest is for compactions that started after the ingest decision was made, yes?

The code could use some more commentary on this.
</blockquote></details>

This code actually predates compaction cancellation and isn't related to ingest splitting - so the understanding was that if we saw an overlapping compaction going into a level, we'd also consider that level as having overlaps with us and stay out of that level. Today we can probably remove this code but that'd come at the cost of greater compaction cancellations and/or more coordination with that compaction to not create a boundary overlap with this ingested file. Something that we can/should consider as a future expansion though.

And yes , the versionEdit install-time compaction cancellation will catch anything that gets started in the meantime (but also anything that's running with a split file as an input file, which this code won't catch. this conditional only concerns with output levels).

Added a comment here and in the compaction cancellation code, good call.
___
*[`ingest.go` line 1861 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NfDTzr13IkK0txyePbJ:-NfHkQh_8A7Ow41gdQb3:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c54bb3ad137206d2be5b5684460dd95e36e21ca9/ingest.go#L1861)):*

I think this should say something about splitFile, since its mentioned at the end of the comment. Something like:

...overlaps with s.ingestFile, which becomes the new splitFile, check splitFile's replacements ...
</blockquote></details>

Done.
___
*[`ingest.go` line 1863 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2582#-NfDUIPQ4oNjIqijTtrh:-NfHkQIK9HA0zo_MkGfL:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c54bb3ad137206d2be5b5684460dd95e36e21ca9/ingest.go#L1863)):*

how about expanding this comment like:

... so on until either we can't find the current splitFile in replacedFiles (which means the splitFile should be split) or we don't find a file that overlaps with s.ingestFile (which means another file being ingested already split things enough to accommodate s.ingestFile).
</blockquote></details>

Done.

PR #2584, opened 2023-06-01
objstorageprovider: use cache in ReadAt

RaduBerinde:
This commit adds code to optionally use the shared cache when reading from an object.

When the read is for a compaction, we don't want to populate the cache (but we do want to read from it if the data is already there).

### DIFF
--- a/objstorage/objstorageprovider/shared.go
+++ b/objstorage/objstorageprovider/shared.go
@@ -240,7 +240,7 @@ func (p *provider) sharedOpenForReading(
 		}
 		return nil, err
 	}
-	return newSharedReadable(reader, size), nil
+	return p.newSharedReadable(reader, size, meta.DiskFileNum), nil
 }
 
 func (p *provider) sharedSize(meta objstorage.ObjectMetadata) (int64, error) {
--- a/objstorage/objstorageprovider/shared_readable.go
+++ b/objstorage/objstorageprovider/shared_readable.go
@@ -7,7 +7,9 @@ package objstorageprovider
 import (
 	"context"
 
+	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/objstorage"
+	"github.com/cockroachdb/pebble/objstorage/objstorageprovider/sharedcache"
 	"github.com/cockroachdb/pebble/objstorage/shared"
 )
 
@@ -16,18 +18,40 @@ import (
 type sharedReadable struct {
 	objReader shared.ObjectReader
 	size      int64
+	fileNum   base.DiskFileNum
+	provider  *provider
 }
 
 var _ objstorage.Readable = (*sharedReadable)(nil)
 
-func newSharedReadable(objReader shared.ObjectReader, size int64) *sharedReadable {
+func (p *provider) newSharedReadable(
+	objReader shared.ObjectReader, size int64, fileNum base.DiskFileNum,
+) *sharedReadable {
 	return &sharedReadable{
 		objReader: objReader,
 		size:      size,
+		fileNum:   fileNum,
+		provider:  p,
 	}
 }
 
+// ReadAt is part of the objstorage.Readable interface.
 func (r *sharedReadable) ReadAt(ctx context.Context, p []byte, offset int64) error {
+	return r.readInternal(ctx, p, offset, false /* forCompaction */)
+}
+
+// readInternal performs a read for the object, using the cache when
+// appropriate.
+func (r *sharedReadable) readInternal(
+	ctx context.Context, p []byte, offset int64, forCompaction bool,
+) error {
+	if cache := r.provider.shared.cache; cache != nil {
+		flags := sharedcache.ReadFlags{
+			// Don't add data to the cache if this read is for a compaction.
+			ReadOnly: forCompaction,
+		}
+		return r.provider.shared.cache.ReadAt(ctx, r.fileNum, p, offset, r.objReader, flags)
+	}
 	return r.objReader.ReadAt(ctx, p, offset)
 }
 
@@ -59,6 +83,7 @@ type sharedReadHandle struct {
 
 var _ objstorage.ReadHandle = (*sharedReadHandle)(nil)
 
+// ReadAt is part of the objstorage.ReadHandle interface.
 func (r *sharedReadHandle) ReadAt(ctx context.Context, p []byte, offset int64) error {
 	readaheadSize := r.maybeReadahead(offset, len(p))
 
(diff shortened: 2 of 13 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2584)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @joshimhoff)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 138 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2584#-NWtaD0ODYVBB4cWEOpY:-NWtaD0ODYVBB4cWEOpZ:b-h2q28m) ([raw file](https://github.com/cockroachdb/pebble/blob/e78020d7f00e025017a76f40e85e25c7bd3ce3b7/objstorage/objstorageprovider/sharedcache/shared_cache.go#L138)):*
> ```Go
> // If data is partially available, a prefix of the data is read; returns n < len(p)
> // and no error. If no prefix is available, returns n = 0 and no error.
> func (c *Cache) TryReadPrefix(
> ```

not too happy with this name, open to suggestions. Maybe it should be "get" instead of "read" (`GetPrefix` or just `Get`?)

joshimhoff (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2584)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

___
*[`objstorage/objstorageprovider/shared_readable.go` line 99 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2584#-NWwotY06daA7YGXC7OL:-NWwotY06daA7YGXC7OM:b-plplbw) ([raw file](https://github.com/cockroachdb/pebble/blob/e78020d7f00e025017a76f40e85e25c7bd3ce3b7/objstorage/objstorageprovider/shared_readable.go#L99)):*
> ```Go
> 		}
> 
> 		if err := r.readInternal(ctx, r.readahead.data, offset); err != nil {
> ```

So read-ahead will work the same whether we are reading SSTs off the file system, reading SSTs from the secondary cache, or reading from blob storage! Very cool.

I do have a Q about this. I can imagine that for large scans, & esp. for compaction, we would actually prefer to read from blob storage. Blob storage is good at thruput. We don't care about time-to-first-byte for compaction. And also doing so will reduce load on the secondary cache, which might be good, e.g. less lock contention in the secondary cache.

Thoughts? I think good to start with this btw. Just a musing to guide our testing once we run this in production.
___
*[`objstorage/objstorageprovider/shared_readable.go` line 113 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2584#-NWwodyS0IpDh9YNAIgE:-NWwodyS0IpDh9YNAIgF:bgtsd2d) ([raw file](https://github.com/cockroachdb/pebble/blob/e78020d7f00e025017a76f40e85e25c7bd3ce3b7/objstorage/objstorageprovider/shared_readable.go#L113)):*
> ```Go
> // readInternal performs a read for the object, using the cache when
> // appropriate.
> func (r *sharedReadHandle) readInternal(ctx context.Context, p []byte, offset int64) error {
> ```

Is there some way to test this? Prob not necessary if below refactor is implemented.
___
*[`objstorage/objstorageprovider/shared_readable.go` line 130 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2584#-NWwnpG_5pq8oqzjUgki:-NWwnpG_5pq8oqzjUgkj:b-dprqi1) ([raw file](https://github.com/cockroachdb/pebble/blob/e78020d7f00e025017a76f40e85e25c7bd3ce3b7/objstorage/objstorageprovider/shared_readable.go#L130)):*
> ```Go
> 			return nil
> 		}
> 		p = p[n:]
> ```

This code is similar to what is in `cache.ReadAt`. Another option is to call `cache.ReadAt` whenever `cache != nil` but add a flag to `cache.ReadAt` called `writeBack` and set it equal to `!r.Compaction`. Then no need for `TryReadPrefix` IIUC.

Seems nice to locate all the block math inside the `cache` itself. No strong feelings.

RaduBerinde (commented):
TFTR, very useful comments!  Updated. I also modified the test to show just the new misses for each testcase (rather than the total).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2584#-:-NWxTXGXBDOvlF3Xbt86:b-53xanp)* status: 0 of 13 files reviewed, 4 unresolved discussions (waiting on @joshimhoff)

___
*[`objstorage/objstorageprovider/shared_readable.go` line 99 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2584#-NWwotY06daA7YGXC7OL:-NWxQ0CU1ygQijLNMHd9:b-pdmixt) ([raw file](https://github.com/cockroachdb/pebble/blob/e78020d7f00e025017a76f40e85e25c7bd3ce3b7/objstorage/objstorageprovider/shared_readable.go#L99)):*

So read-ahead will work the same whether we are reading SSTs off the file system, reading SSTs from the secondary cache, or reading from blob storage! Very cool.

I do have a Q about this. I can imagine that for large scans, & esp. for compaction, we would actually prefer to read from blob storage. Blob storage is good at thruput. We don't care about time-to-first-byte for compaction. And also doing so will reduce load on the secondary cache, which might be good, e.g. less lock contention in the secondary cache.

Thoughts? I think good to start with this btw. Just a musing to guide our testing once we run this in production.
</blockquote></details>

Interesting, I'll add a TODO
___
*[`objstorage/objstorageprovider/shared_readable.go` line 113 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2584#-NWwodyS0IpDh9YNAIgE:-NWxTW65282PEcE7n7h-:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e78020d7f00e025017a76f40e85e25c7bd3ce3b7/objstorage/objstorageprovider/shared_readable.go#L113)):*

Is there some way to test this? Prob not necessary if below refactor is implemented.
</blockquote></details>

Done.
___
*[`objstorage/objstorageprovider/shared_readable.go` line 130 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2584#-NWwnpG_5pq8oqzjUgki:-NWxPenM74fxTNuraUdC:bxmiyy1) ([raw file](https://github.com/cockroachdb/pebble/blob/e78020d7f00e025017a76f40e85e25c7bd3ce3b7/objstorage/objstorageprovider/shared_readable.go#L130)):*

This code is similar to what is in `cache.ReadAt`. Another option is to call `cache.ReadAt` whenever `cache != nil` but add a flag to `cache.ReadAt` called `writeBack` and set it equal to `!r.Compaction`. Then no need for `TryReadPrefix` IIUC.

Seems nice to locate all the block math inside the `cache` itself. No strong feelings.
</blockquote></details>

Very good point, much nicer this way.

joshimhoff (approved):
LGTM!

PR #2586, opened 2023-06-02
sharedcache: move to fine-grained locking

joshimhoff:

**sharedcache: fix bug that "leaks" cache block indexes**

This commit introduces assertions regarding in-memory shard state such as
s.mu.where & s.mu.free that caught an existing bug in the cache. The bug is:
If concurrent calls to ReadAt are made, then shard.set might be called for
a logical block that is already in the cache. Before this commit, such a
situation would "leak" a cache block, meaning the cache block index would
be removed from s.mu.where, implying the cache block could never be used
again. As of this commit, the bug is fixed.

**sharedcache: move to fine-grained locking**

Before this commit, s.mu was held when s.get & s.set were reading and
writing to s.file. This was done to simplify merging an initial implementation
with the necessary correctness properties. This commit moves to fine-grained
locking, to improve the performance of the cache. A table mapping each cache
block index to its "lock state" is introduced. During the reading and writing
to s.file, only the specific cache block being updated is locked. Both read and
write locks are supported, so that multiple reads of some cache block can be
done concurrently. Also, writing back to the cache is now done asynchronously
with respect to the call to ReadAt.

With this commit merged, I think we can run some production experiments
measuring the efficiency of the cache.

### DIFF
--- a/objstorage/objstorageprovider/shared.go
+++ b/objstorage/objstorageprovider/shared.go
@@ -80,7 +80,7 @@ func (p *provider) sharedInit() error {
 			numShards = 2 * runtime.GOMAXPROCS(0)
 		}
 
-		p.shared.cache, err = sharedcache.Open(p.st.FS, p.st.FSDirName, blockSize, p.st.Shared.CacheSizeBytes, numShards)
+		p.shared.cache, err = sharedcache.Open(p.st.FS, p.st.Logger, p.st.FSDirName, blockSize, p.st.Shared.CacheSizeBytes, numShards)
 		if err != nil {
 			return errors.Wrapf(err, "pebble: could not open shared object cache")
 		}
--- a/objstorage/objstorageprovider/sharedcache/shared_cache.go
+++ b/objstorage/objstorageprovider/sharedcache/shared_cache.go
@@ -19,8 +19,7 @@ import (
 )
 
 const (
-	// ShardingBlockSize is public to enable testing from sharedcache_test.
-	ShardingBlockSize = 1024 * 1024
+	shardingBlockSize = 1024 * 1024
 )
 
 // Cache is a persistent cache backed by a local filesystem. It is intended
@@ -32,21 +31,24 @@ type Cache struct {
 
 	blockSize int
 
+	writeBackWaitGroup sync.WaitGroup
 	// TODO(josh): Have a dedicated metrics struct. Right now, this
 	// is just for testing.
-	Misses atomic.Int32
+	misses atomic.Int32
 }
 
 // Open opens a cache. If there is no existing cache at fsDir, a new one
 // is created.
-func Open(fs vfs.FS, logger base.Logger, fsDir string, blockSize int, sizeBytes int64, numShards int) (*Cache, error) {
-	min := ShardingBlockSize * int64(numShards)
+func Open(
+	fs vfs.FS, logger base.Logger, fsDir string, blockSize int, sizeBytes int64, numShards int,
+) (*Cache, error) {
+	min := shardingBlockSize * int64(numShards)
 	if sizeBytes < min {
 		return nil, errors.Errorf("cache size %d lower than min %d", sizeBytes, min)
 	}
 
 	sc := &Cache{
-		logger: logger,
+		logger:    logger,
 		blockSize: blockSize,
 	}
 	sc.shards = make([]shard, numShards)
@@ -62,6 +64,8 @@ func Open(fs vfs.FS, logger base.Logger, fsDir string, blockSize int, sizeBytes 
 // Close closes the cache. Methods such as ReadAt should not be called after Close is
 // called.
 func (c *Cache) Close() error {
+	c.writeBackWaitGroup.Wait()
+
 	var retErr error
 	for i := range c.shards {
 		if err := c.shards[i].close(); err != nil && retErr == nil {
@@ -113,7 +117,7 @@ func (c *Cache) ReadAt(
 		}
 	}
 
-	c.Misses.Add(1)
+	c.misses.Add(1)
 
 	if flags.ReadOnly {
 		return objReader.ReadAt(ctx, p, ofs)
@@ -133,18 +137,28 @@ func (c *Cache) ReadAt(
 	adjustedP := make([]byte, adjustedLen)
 
 	// Read the rest from the object.
+	// TODO(josh): To have proper EOF handling, we will need to use the Size method of the
+	// readable to limit the size of a read when at the end of a file. For now, the cache
+	// just swallows all io.EOF errors.
 	if err := objReader.ReadAt(ctx, adjustedP, adjustedOfs); err != nil && err != io.EOF {
 		return err
 	}
 	copy(p, adjustedP[sizeOfOffAdjustment:])
 
-	// TODO(josh): Writing back to the cache should be async with respect to the
-	// call to ReadAt.
-	if err := c.set(fileNum, adjustedP, adjustedOfs); err != nil {
-		// TODO(josh): Would like to log at error severity, but base.Logger doesn't
-		// have error severity.
-		c.logger.Infof("writing back to cache after miss failed: %v", err)
-	}
+	// TODO(josh): For writing back to the cache, we may want a concurrency-limited approach
+	// that does batching, instead of what is below. I think it's reasonable though to
+	// run a production experiment with what is below, before making adjustments. Note that
+	// the writes being done are in multiples of the filesystem block size. The filesystem
+	// might do okay with them.
+	c.writeBackWaitGroup.Add(1)
+	go func() {
+		defer c.writeBackWaitGroup.Done()
+		if err := c.set(fileNum, adjustedP, adjustedOfs); err != nil {
+			// TODO(josh): Would like to log at error severity, but base.Logger doesn't
+			// have error severity.
+			c.logger.Infof("writing back to cache after miss failed: %v", err)
+		}
+	}()
 	return nil
 }
 
@@ -162,7 +176,7 @@ func (c *Cache) get(fileNum base.DiskFileNum, p []byte, ofs int64) (n int, _ err
 	for {
 		shard := c.getShard(fileNum, ofs+int64(n))
 		cappedLen := len(p[n:])
-		if toBoundary := int(ShardingBlockSize - ((ofs + int64(n)) % ShardingBlockSize)); cappedLen > toBoundary {
+		if toBoundary := int(shardingBlockSize - ((ofs + int64(n)) % shardingBlockSize)); cappedLen > toBoundary {
 			cappedLen = toBoundary
 		}
 		numRead, err := shard.get(fileNum, p[n:n+cappedLen], ofs+int64(n))
@@ -200,7 +214,7 @@ func (c *Cache) set(fileNum base.DiskFileNum, p []byte, ofs int64) error {
 	for {
 		shard := c.getShard(fileNum, ofs+int64(n))
 		cappedLen := len(p[n:])
-		if toBoundary := int(ShardingBlockSize - ((ofs + int64(n)) % ShardingBlockSize)); cappedLen > toBoundary {
+		if toBoundary := int(shardingBlockSize - ((ofs + int64(n)) % shardingBlockSize)); cappedLen > toBoundary {
 			cappedLen = toBoundary
 		}
 		err := shard.set(fileNum, p[n:n+cappedLen], ofs+int64(n))
@@ -219,7 +233,7 @@ func (c *Cache) set(fileNum base.DiskFileNum, p []byte, ofs int64) error {
 
 func (c *Cache) getShard(fileNum base.DiskFileNum, ofs int64) *shard {
 	const prime64 = 1099511628211
-	hash := uint64(fileNum.FileNum())*prime64 + uint64(ofs)/ShardingBlockSize
+	hash := uint64(fileNum.FileNum())*prime64 + uint64(ofs)/shardingBlockSize
 	// TODO(josh): Instance change ops are often run in production. Such an operation
 	// updates len(c.shards); see openSharedCache. As a result, the behavior of this
 	// function changes, and the cache empties out at restart time. We may want a better
@@ -233,17 +247,39 @@ type shard struct {
 	blockSize    int
 	mu           struct {
 		sync.Mutex
-		// TODO(josh): Neither of these datastructures are space-efficient.
+		// TODO(josh): None of these datastructures are space-efficient.
 		// Focusing on correctness to start.
-		where map[metadataKey]int64
-		free  []int64
+		where whereMap
+		locks []lockState
+		free  []cacheBlockIndex
 	}
 }
 
-type metadataKey struct {
-	fileNum    base.DiskFileNum
-	blockIndex int64
+// Maps a logical block in an SST to an index of the cache block with the
+// file contents (to the "cache block index").
+type whereMap map[logicalBlockID]cacheBlockIndex
+
+type logicalBlockID struct {
+	filenum base.DiskFileNum
+	// Offset into the file, in units of shard.BlockSize.
+	offsetInUnitsOfBlocks int64
 }
+
+type cacheBlockIndex int64
+
+type lockState int64
+
+const (
+	unlocked lockState = 0
+	// >0 lockState tracks the number of distinct readers of some cache block / logical block
+	// which is in the secondary cache. It is used to ensure that a cache block is not evicted
+	// and overwritten, while there are active readers.
+	readLockTakenInc = 1
+	// -1 lockState indicates that some cache block is currently being populated with data from
+	// blob storage. It is used to ensure that a cache block is not read or evicted again, while
+	// it is being populated.
+	writeLockTaken = -1
+)
 
 func (s *shard) init(
 	fs vfs.FS, fsDir string, shardIdx int, sizeInBlocks int64, blockSize int,
@@ -251,8 +287,8 @@ func (s *shard) init(
 	*s = shard{
 		sizeInBlocks: sizeInBlocks,
 	}
-	if blockSize < 1024 || ShardingBlockSize%blockSize != 0 {
-		return errors.Newf("invalid block size %d (must divide %d)", blockSize, ShardingBlockSize)
+	if blockSize < 1024 || shardingBlockSize%blockSize != 0 {
+		return errors.Newf("invalid block size %d (must divide %d)", blockSize, shardingBlockSize)
 	}
 	s.blockSize = blockSize
 	file, err := fs.OpenReadWrite(fs.PathJoin(fsDir, fmt.Sprintf("SHARED-CACHE-%03d", shardIdx)))
@@ -269,9 +305,10 @@ func (s *shard) init(
 	// TODO(josh): Right now, the secondary cache is not persistent. All existing
 	// cache contents will be over-written, since all metadata is only stored in
 	// memory.
-	s.mu.where = make(map[metadataKey]int64)
+	s.mu.where = make(whereMap)
 	for i := int64(0); i < sizeInBlocks; i++ {
-		s.mu.free = append(s.mu.free, i)
+		s.mu.locks = append(s.mu.locks, unlocked)
+		s.mu.free = append(s.mu.free, cacheBlockIndex(i))
 	}
 
 	return nil
@@ -291,32 +328,51 @@ func (s *shard) close() error {
 //
 // If data is partially available, a prefix of the data is read; returns n < len(p)
 // and no error. If no prefix is available, returns n = 0 and no error.
+//
+// TODO(josh): Today, if there are two cache blocks needed to satisfy a read, and the
+// first block is not in the cache and the second one is, we will read both from
+// blob storage. We should fix this. This is not an unlikely scenario if we are doing
+// a reverse scan, since those iterate over sstable blocks in reverse order and due to
+// cache block aligned reads will have read the suffix of the sstable block that will
+// be needed next.
 func (s *shard) get(fileNum base.DiskFileNum, p []byte, ofs int64) (n int, _ error) {
 	if invariants.Enabled {
-		if ofs/ShardingBlockSize != (ofs+int64(len(p))-1)/ShardingBlockSize {
+		if ofs/shardingBlockSize != (ofs+int64(len(p))-1)/shardingBlockSize {
 			panic(fmt.Sprintf("get crosses shard boundary: %v %v", ofs, len(p)))
 		}
 		s.assertShardStateIsConsistent()
 	}
 
-	// TODO(josh): Make the locking more fine-grained. Do not hold locks during calls
-	// to ReadAt.
-	s.mu.Lock()
-	defer s.mu.Unlock()
-
 	// The data extent might cross cache block boundaries, hence the loop. In the hot
 	// path, max two iterations of this loop will be executed, since reads are sized
 	// in units of sstable block size.
 	for {
-		cacheBlockInd, ok := s.mu.where[metadataKey{
-			fileNum:    fileNum,
-			blockIndex: (ofs + int64(n)) / int64(s.blockSize),
-		}]
+		k := logicalBlockID{
+			filenum:               fileNum,
+			offsetInUnitsOfBlocks: (ofs + int64(n)) / int64(s.blockSize),
+		}
+		s.mu.Lock()
+		cacheBlockInd, ok := s.mu.where[k]
+		// TODO(josh): Multiple reads within the same few milliseconds (anything that is smaller
+		// than blob storage read latency) that miss on the same logical block ID will not necessarily
+		// be rare. We may want to do only one read, with the later readers blocking on the first read
+		// completing. This could be implemented either here or in the primary block cache. See
+		// https://github.com/cockroachdb/pebble/pull/2586 for additional discussion.
 		if !ok {
+			s.mu.Unlock()
 			return n, nil
 		}
+		if s.mu.locks[cacheBlockInd] == writeLockTaken {
+			// In practice, if we have two reads of the same SST block in close succession, we
+			// would expect the second to hit in the in-memory block cache. So it's not worth
+			// optimizing this case here.
+			s.mu.Unlock()
+			return n, nil
+		}
+		s.mu.locks[cacheBlockInd] += readLockTakenInc
+		s.mu.Unlock()
 
-		readAt := cacheBlockInd * int64(s.blockSize)
+		readAt := int64(cacheBlockInd) * int64(s.blockSize)
 		if n == 0 { // if first read
 			readAt += ofs % int64(s.blockSize)
 		}
@@ -328,9 +384,11 @@ func (s *shard) get(fileNum base.DiskFileNum, p []byte, ofs int64) (n int, _ err
 
 		if len(p[n:]) <= readSize {
 			numRead, err := s.file.ReadAt(p[n:], readAt)
+			s.dropReadLock(cacheBlockInd)
 			return n + numRead, err
 		}
 		numRead, err := s.file.ReadAt(p[n:n+readSize], readAt)
+		s.dropReadLock(cacheBlockInd)
 		if err != nil {
 			return 0, err
 		}
@@ -347,7 +405,7 @@ func (s *shard) get(fileNum base.DiskFileNum, p []byte, ofs int64) (n int, _ err
 // If all of p is not written to the shard, set returns a non-nil error.
 func (s *shard) set(fileNum base.DiskFileNum, p []byte, ofs int64) error {
 	if invariants.Enabled {
-		if ofs/ShardingBlockSize != (ofs+int64(len(p))-1)/ShardingBlockSize {
+		if ofs/shardingBlockSize != (ofs+int64(len(p))-1)/shardingBlockSize {
 			panic(fmt.Sprintf("set crosses shard boundary: %v %v", ofs, len(p)))
 		}
 		if ofs%int64(s.blockSize) != 0 || len(p)%s.blockSize != 0 {
@@ -355,11 +413,6 @@ func (s *shard) set(fileNum base.DiskFileNum, p []byte, ofs int64) error {
 		}
 		s.assertShardStateIsConsistent()
 	}
-
-	// TODO(josh): Make the locking more fine-grained. Do not hold locks during calls
-	// to WriteAt.
-	s.mu.Lock()
-	defer s.mu.Unlock()
 
 	// The data extent might cross cache block boundaries, hence the loop. In the hot
 	// path, max two iterations of this loop will be executed, since reads are sized
@@ -376,24 +429,44 @@ func (s *shard) set(fileNum base.DiskFileNum, p []byte, ofs int64) error {
 		}
 
 		// If the logical block is already in the cache, we should skip doing a set.
-		k := metadataKey{
-			filenum:    fileNum,
-			blockIndex: (ofs + int64(n)) / int64(s.blockSize),
+		k := logicalBlockID{
+			filenum:               fileNum,
+			offsetInUnitsOfBlocks: (ofs + int64(n)) / int64(s.blockSize),
 		}
+		s.mu.Lock()
 		if _, ok := s.mu.where[k]; ok {
+			s.mu.Unlock()
 			n += s.blockSize
 			continue
 		}
 
-		var cacheBlockInd int64
+		// Determine cache block index by looking at the free list and randomly evicting if
+		// it is empty.
+		var cacheBlockInd cacheBlockIndex
 		if len(s.mu.free) == 0 {
 			// TODO(josh): Right now, we do random eviction. Eventually, we will do something
 			// more sophisticated, e.g. leverage ClockPro.
-			var k metadataKey
+			foundBlockToEvict := false
+			var k logicalBlockID
 			for k1, v := range s.mu.where {
+				// This implies that hot blocks will not be evicted as often, meaning the
+				// eviction scheme is not quite random. I think that's fine for now, since we plan
+				// to leverage a more sophisticated eviction scheme eventually anyway.
+				lock := s.mu.locks[v]
+				if lock >= readLockTakenInc || lock == writeLockTaken {
+					continue
+				}
+				foundBlockToEvict = true
 				cacheBlockInd = v
 				k = k1
 				break
+			}
+			// TODO(josh): We may want to block until a block frees up, instead of returning
+			// an error here. But I think we can do that later on, e.g. after running some production
+			// experiments.
+			if !foundBlockToEvict {
+				s.mu.Unlock()
+				return errors.New("no block to evict so skipping write to cache")
 			}
 			delete(s.mu.where, k)
 		} else {
@@ -401,48 +474,89 @@ func (s *shard) set(fileNum base.DiskFileNum, p []byte, ofs int64) error {
 			s.mu.free = s.mu.free[:len(s.mu.free)-1]
 		}
 
-		s.mu.where[metadataKey{
-			fileNum:    fileNum,
-			blockIndex: (ofs + int64(n)) / int64(s.blockSize),
-		}] = cacheBlockInd
+		s.mu.where[k] = cacheBlockInd
+		s.mu.locks[cacheBlockInd] = writeLockTaken
+		s.mu.Unlock()
 
-		writeAt := cacheBlockInd * int64(s.blockSize)
+		writeAt := int64(cacheBlockInd) * int64(s.blockSize)
 
 		writeSize := s.blockSize
 		if len(p[n:]) <= writeSize {
 			writeSize = len(p[n:])
 		}
 
-		numWritten, err := s.file.WriteAt(p[n:n+writeSize], writeAt)
+		_, err := s.file.WriteAt(p[n:n+writeSize], writeAt)
 		if err != nil {
+			s.freeBlock(k, cacheBlockInd)
 			return err
 		}
+		s.dropWriteLock(cacheBlockInd)
 
-		// Note that numWritten == writeSize, since we checked for an error above.
-		n += numWritten
+		n += writeSize
 	}
+}
+
+// Doesn't inline currently. This might be okay, but something to keep in mind.
+func (s *shard) dropReadLock(cacheBlockInd cacheBlockIndex) {
+	s.mu.Lock()
+	s.mu.locks[cacheBlockInd] -= readLockTakenInc
+	if invariants.Enabled {
+		if s.mu.locks[cacheBlockInd] < 0 {
+			panic(fmt.Sprintf("unexpected lock state %v in dropReadLock: %v", s.mu.locks[cacheBlockInd], s.mu.locks))
+		}
+	}
+	s.mu.Unlock()
+}
+
+// Doesn't inline currently. This might be okay, but something to keep in mind.
+func (s *shard) dropWriteLock(cacheBlockInd cacheBlockIndex) {
+	s.mu.Lock()
+	if invariants.Enabled {
+		if s.mu.locks[cacheBlockInd] != writeLockTaken {
+			panic(fmt.Sprintf("unexpected lock state %v in dropWriteLock: %v", s.mu.locks[cacheBlockInd], s.mu.locks))
+		}
+	}
+	s.mu.locks[cacheBlockInd] = unlocked
+	s.mu.Unlock()
+}
+
+// Doesn't inline currently. This might be okay, but something to keep in mind.
+func (s *shard) freeBlock(k logicalBlockID, cacheBlockInd cacheBlockIndex) {
+	s.mu.Lock()
+	delete(s.mu.where, k)
+	s.mu.locks[cacheBlockInd] = unlocked
+	s.mu.free = append(s.mu.free, cacheBlockInd)
+	s.mu.Unlock()
 }
 
 func (s *shard) assertShardStateIsConsistent() {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	cacheBlockInds := make(map[int64]bool)
+	cacheBlockInds := make(map[cacheBlockIndex]bool)
 	for _, v := range s.mu.where {
 		if cacheBlockInds[v] {
-			panic(fmt.Sprintf("repeated cache block index: %v %v %v %v", v, cacheBlockInds, s.mu.where, s.mu.free))
+			panic(fmt.Sprintf("repeated cache block index (where): %v %v %v %v", v, cacheBlockInds, s.mu.where, s.mu.free))
 		}
 		cacheBlockInds[v] = true
 	}
 	for _, v := range s.mu.free {
 		if cacheBlockInds[v] {
-			panic(fmt.Sprintf("repeated cache block index: %v %v %v %v", v, cacheBlockInds, s.mu.where, s.mu.free))
+			panic(fmt.Sprintf("repeated cache block index (free): %v %v %v %v", v, cacheBlockInds, s.mu.where, s.mu.free))
 		}
 		cacheBlockInds[v] = true
 	}
 	for i := int64(0); i < s.sizeInBlocks; i++ {
-		if !cacheBlockInds[i] {
+		if !cacheBlockInds[cacheBlockIndex(i)] {
 			panic(fmt.Sprintf("missing cache block index: %v %v %v %v", i, cacheBlockInds, s.mu.where, s.mu.free))
+		}
+	}
+	if int64(len(s.mu.locks)) != s.sizeInBlocks {
+		panic(fmt.Sprintf("lock table isn't correct size: %v %v", len(s.mu.locks), s.sizeInBlocks))
+	}
+	for _, ls := range s.mu.locks {
+		if ls < writeLockTaken {
+			panic(fmt.Sprintf("lock state %v is not allowed: %v", ls, s.mu.locks))
 		}
 	}
 }
--- a/objstorage/objstorageprovider/sharedcache/shared_cache.go
+++ b/objstorage/objstorageprovider/sharedcache/shared_cache.go
@@ -39,13 +39,14 @@ type Cache struct {
 
 // Open opens a cache. If there is no existing cache at fsDir, a new one
 // is created.
-func Open(fs vfs.FS, fsDir string, blockSize int, sizeBytes int64, numShards int) (*Cache, error) {
+func Open(fs vfs.FS, logger base.Logger, fsDir string, blockSize int, sizeBytes int64, numShards int) (*Cache, error) {
 	min := ShardingBlockSize * int64(numShards)
 	if sizeBytes < min {
 		return nil, errors.Errorf("cache size %d lower than min %d", sizeBytes, min)
 	}
 
 	sc := &Cache{
+		logger: logger,
 		blockSize: blockSize,
 	}
 	sc.shards = make([]shard, numShards)
@@ -295,6 +296,7 @@ func (s *shard) get(fileNum base.DiskFileNum, p []byte, ofs int64) (n int, _ err
 		if ofs/ShardingBlockSize != (ofs+int64(len(p))-1)/ShardingBlockSize {
 			panic(fmt.Sprintf("get crosses shard boundary: %v %v", ofs, len(p)))
 		}
+		s.assertShardStateIsConsistent()
 	}
 
 	// TODO(josh): Make the locking more fine-grained. Do not hold locks during calls
@@ -351,6 +353,7 @@ func (s *shard) set(fileNum base.DiskFileNum, p []byte, ofs int64) error {
 		if ofs%int64(s.blockSize) != 0 || len(p)%s.blockSize != 0 {
 			panic(fmt.Sprintf("set with ofs & len not multiples of block size: %v %v", ofs, len(p)))
 		}
+		s.assertShardStateIsConsistent()
 	}
 
 	// TODO(josh): Make the locking more fine-grained. Do not hold locks during calls
@@ -363,6 +366,25 @@ func (s *shard) set(fileNum base.DiskFileNum, p []byte, ofs int64) error {
 	// in units of sstable block size.
 	n := 0
 	for {
+		if n == len(p) {
+			return nil
+		}
+		if invariants.Enabled {
+			if n > len(p) {
+				panic(fmt.Sprintf("set with n greater than len(p): %v %v", n, len(p)))
+			}
+		}
+
+		// If the logical block is already in the cache, we should skip doing a set.
+		k := metadataKey{
+			filenum:    fileNum,
+			blockIndex: (ofs + int64(n)) / int64(s.blockSize),
+		}
+		if _, ok := s.mu.where[k]; ok {
+			n += s.blockSize
+			continue
+		}
+
 		var cacheBlockInd int64
 		if len(s.mu.free) == 0 {
 			// TODO(josh): Right now, we do random eviction. Eventually, we will do something
@@ -385,14 +407,12 @@ func (s *shard) set(fileNum base.DiskFileNum, p []byte, ofs int64) error {
 		}] = cacheBlockInd
 
 		writeAt := cacheBlockInd * int64(s.blockSize)
-		writeSize := s.blockSize
 
+		writeSize := s.blockSize
 		if len(p[n:]) <= writeSize {
-			// Ignore num written ret value, since if partial write, an error
-			// is returned.
-			_, err := s.file.WriteAt(p[n:], writeAt)
-			return err
+			writeSize = len(p[n:])
 		}
+
 		numWritten, err := s.file.WriteAt(p[n:n+writeSize], writeAt)
 		if err != nil {
 			return err
@@ -400,5 +420,29 @@ func (s *shard) set(fileNum base.DiskFileNum, p []byte, ofs int64) error {
 
 		// Note that numWritten == writeSize, since we checked for an error above.
 		n += numWritten
+	}
+}
+
+func (s *shard) assertShardStateIsConsistent() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	cacheBlockInds := make(map[int64]bool)
+	for _, v := range s.mu.where {
+		if cacheBlockInds[v] {
+			panic(fmt.Sprintf("repeated cache block index: %v %v %v %v", v, cacheBlockInds, s.mu.where, s.mu.free))
+		}
+		cacheBlockInds[v] = true
+	}
+	for _, v := range s.mu.free {
+		if cacheBlockInds[v] {
+			panic(fmt.Sprintf("repeated cache block index: %v %v %v %v", v, cacheBlockInds, s.mu.where, s.mu.free))
+		}
+		cacheBlockInds[v] = true
+	}
+	for i := int64(0); i < s.sizeInBlocks; i++ {
+		if !cacheBlockInds[i] {
+			panic(fmt.Sprintf("missing cache block index: %v %v %v %v", i, cacheBlockInds, s.mu.where, s.mu.free))
+		}
 	}
 }
--- a/objstorage/objstorageprovider/sharedcache/shared_cache_helpers_test.go
+++ b/objstorage/objstorageprovider/sharedcache/shared_cache_helpers_test.go
@@ -1,0 +1,13 @@
+package sharedcache
+
+func ShardingBlockSize() int {
+	return shardingBlockSize
+}
+
+func (c *Cache) Misses() int32 {
+	return c.misses.Load()
+}
+
+func (c *Cache) WaitForWritesToComplete() {
+	c.writeBackWaitGroup.Wait()
+}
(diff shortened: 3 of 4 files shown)


joshimhoff:
I have just realized that my read lock approach doesn't work as is. In the case of two concurrent reads, the first reader to unlock will unlock despite that the second reader's read is not done. I will fix this bug, perhaps by keeping a reference count. 

I'll leave the PR up for review, despite this bug being present right now, as early feedback will be helpful.

RaduBerinde (commented):
First commit LGTM

Yeah, the typical solution is to have a reader count where 0 = unlocked and -1 = write locked.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586#-:-NWxXT9X2Wla4Hel6SDp:bh45ji7)* status: 0 of 3 files reviewed, 6 unresolved discussions (waiting on @joshimhoff and @sumeerbhola)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 405 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NWxUtAWCYKC4yhx91C1:-NWxUtAWCYKC4yhx91C2:b-4puvck) ([raw file](https://github.com/cockroachdb/pebble/blob/447e324d2a5d6a7912cd10f10181b28cbfbfdf30/objstorage/objstorageprovider/sharedcache/shared_cache.go#L405)):*
> ```Go
> 		}
> 
> 		// Note that numWritten == writeSize, since we checked for an error above.
> ```

[nit] I'd just do `_, err :=` and use `writeSize`
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 423 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NWxUYp2DTlfNXKYSo4X:-NWxUYp2DTlfNXKYSo4Y:bgyzkn4) ([raw file](https://github.com/cockroachdb/pebble/blob/447e324d2a5d6a7912cd10f10181b28cbfbfdf30/objstorage/objstorageprovider/sharedcache/shared_cache.go#L423)):*
> ```Go
> 	for _, v := range s.mu.free {
> 		if cacheBlockInds[v] {
> 			panic(fmt.Sprintf("repeated cache block index: %v %v %v %v", v, cacheBlockInds, s.mu.where, s.mu.free))
> ```

[nit] use a different message (block both used and free?)
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 35 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NWxVJgt7PnqC74M2CMA:-NWxVJgt7PnqC74M2CMB:b-x4o2lq) ([raw file](https://github.com/cockroachdb/pebble/blob/450a89415286db7513f2750a4d5a9b901ca9a4f9/objstorage/objstorageprovider/sharedcache/shared_cache.go#L35)):*
> ```Go
> 	blockSize int
> 
> 	// WriteBackWaitGroup is public to enable testing from sharedcache_test.
> ```

What you can do to keep the non-testing code clean is to create a `_test` file that has `package sharedcache` and add getters, e.g. `func (c *Cache) WriteBackWaitGroup() sync.WaitGroup { return c.writeBackWaitGroup }`. These methods will be accessible from the other `_test` file.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 236 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NWxWLKyALVkmaCr3dFS:-NWxWLKyALVkmaCr3dFT:b8wk3po) ([raw file](https://github.com/cockroachdb/pebble/blob/450a89415286db7513f2750a4d5a9b901ca9a4f9/objstorage/objstorageprovider/sharedcache/shared_cache.go#L236)):*
> ```Go
> 		// TODO(josh): None of these datastructures are space-efficient.
> 		// Focusing on correctness to start.
> 		where map[whereMapKey]int64
> ```

[nit] consider making this a separate type and documenting what the value is
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 242 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NWxW4-z7ewjCLlGwhm9:-NWxW4-z7ewjCLlGwhmA:b-z28n9p) ([raw file](https://github.com/cockroachdb/pebble/blob/450a89415286db7513f2750a4d5a9b901ca9a4f9/objstorage/objstorageprovider/sharedcache/shared_cache.go#L242)):*
> ```Go
> }
> 
> type whereMapKey struct {
> ```

[nit] a better name would be descriptibe, e.g. `objectBlockID`
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 327 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NWxXEv21nU_GBhIoi6G:-NWxXEv21nU_GBhIoi6H:b9gpfv6) ([raw file](https://github.com/cockroachdb/pebble/blob/450a89415286db7513f2750a4d5a9b901ca9a4f9/objstorage/objstorageprovider/sharedcache/shared_cache.go#L327)):*
> ```Go
> 			// We could block on the write completing, but it is not necessary, since if two
> 			// reads are done in close succession, we would expect a hit on the second one due
> 			// to the in-memory block cache.
> ```

I was confused by this comment for a while, I'd rephrase along the lines of "In practice, if we have two reads of the same SST block in close succession, we would expect the second to hit in the in-memory block cache; so it's not worth optimizing this case here".

joshimhoff (commented):
The code is now properly tracking the number of readers.

PTAL!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586#-:-NXAZ2bB0ZYI4IXcZoZn:blwmm18)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 405 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NWxUtAWCYKC4yhx91C1:-NXAVvqf1ZwII2glkkhB:bcvjgzp) ([raw file](https://github.com/cockroachdb/pebble/blob/447e324d2a5d6a7912cd10f10181b28cbfbfdf30/objstorage/objstorageprovider/sharedcache/shared_cache.go#L405)):*

[nit] I'd just do `_, err :=` and use `writeSize`
</blockquote></details>

Good idea. Done.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 423 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NWxUYp2DTlfNXKYSo4X:-NXAW7rMAYkJu_YdsKl5:b-hzd16b) ([raw file](https://github.com/cockroachdb/pebble/blob/447e324d2a5d6a7912cd10f10181b28cbfbfdf30/objstorage/objstorageprovider/sharedcache/shared_cache.go#L423)):*

[nit] use a different message (block both used and free?)
</blockquote></details>

It doesn't necessarily mean block was used and in free, as block could be in free twice. The complete in-memory state is printed out tho, so I think the ambiguity is okay. Note tho that I did add a "(used)" or "(free)" suffix to the messages.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 35 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NWxVJgt7PnqC74M2CMA:-NXAYrwb-YNUOE8S5T1U:bclxk5d) ([raw file](https://github.com/cockroachdb/pebble/blob/450a89415286db7513f2750a4d5a9b901ca9a4f9/objstorage/objstorageprovider/sharedcache/shared_cache.go#L35)):*

What you can do to keep the non-testing code clean is to create a `_test` file that has `package sharedcache` and add getters, e.g. `func (c *Cache) WriteBackWaitGroup() sync.WaitGroup { return c.writeBackWaitGroup }`. These methods will be accessible from the other `_test` file.
</blockquote></details>

Done. Much better. Thank you.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 236 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NWxWLKyALVkmaCr3dFS:-NXAUCW9DBh0VXDQ7bYn:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/450a89415286db7513f2750a4d5a9b901ca9a4f9/objstorage/objstorageprovider/sharedcache/shared_cache.go#L236)):*

[nit] consider making this a separate type and documenting what the value is
</blockquote></details>

Done.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 242 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NWxW4-z7ewjCLlGwhm9:-NXAVLTBCZOFNwuZTzDM:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/450a89415286db7513f2750a4d5a9b901ca9a4f9/objstorage/objstorageprovider/sharedcache/shared_cache.go#L242)):*

[nit] a better name would be descriptibe, e.g. `objectBlockID`
</blockquote></details>

Done.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 327 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NWxXEv21nU_GBhIoi6G:-NXAVXcJ-5h7NnAaa_WJ:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/450a89415286db7513f2750a4d5a9b901ca9a4f9/objstorage/objstorageprovider/sharedcache/shared_cache.go#L327)):*

I was confused by this comment for a while, I'd rephrase along the lines of "In practice, if we have two reads of the same SST block in close succession, we would expect the second to hit in the in-memory block cache; so it's not worth optimizing this case here".
</blockquote></details>

Done.

RaduBerinde (approved):
> Reviewed 1 of 3 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586)* status: all files reviewed, 3 unresolved discussions (waiting on @joshimhoff and @sumeerbhola)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 274 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXCht2q3IY5kXepw3R0:-NXCht2q3IY5kXepw3R1:b-z47m0f) ([raw file](https://github.com/cockroachdb/pebble/blob/d96be4ea2e33b9286d8364352e08ad1d2d5ace0b/objstorage/objstorageprovider/sharedcache/shared_cache.go#L274)):*
> ```Go
> 	unlocked lockState = 0
> 	// locks tracks the number of distinct readers of the logical block in an SST.
> 	readLockTakenInc = 1
> ```

[nit] it doesn't make much sense to have this constant, I'd remove it and move the comment to `lockState`
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 484 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXCi1HTB3LA9IljI_65:-NXCi1HTB3LA9IljI_66:bh57mxn) ([raw file](https://github.com/cockroachdb/pebble/blob/d96be4ea2e33b9286d8364352e08ad1d2d5ace0b/objstorage/objstorageprovider/sharedcache/shared_cache.go#L484)):*
> ```Go
> func (s *shard) dropReadLock(cacheBlockInd int64) {
> 	s.mu.Lock()
> 	s.mu.locks[cacheBlockInd] -= readLockTakenInc
> ```

[nit] worth an `invariants` assertion that the count is > 0
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 491 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXCi47PBtbXjXHGIdEC:-NXCi47PBtbXjXHGIdED:b-179ues) ([raw file](https://github.com/cockroachdb/pebble/blob/d96be4ea2e33b9286d8364352e08ad1d2d5ace0b/objstorage/objstorageprovider/sharedcache/shared_cache.go#L491)):*
> ```Go
> func (s *shard) dropWriteLock(cacheBlockInd int64) {
> 	s.mu.Lock()
> 	s.mu.locks[cacheBlockInd] = unlocked
> ```

[nit] `invariants` assertion that the count is -1

RaduBerinde (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586)* status: all files reviewed, 3 unresolved discussions (waiting on @joshimhoff and @sumeerbhola)

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586#-:-NXDc5dVBaztRRvOchiP:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @joshimhoff and @sumeerbhola)

joshimhoff (commented):
TTFR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586#-:-NXFirK3-ky451Guli8W:b-80vssw)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 274 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXCht2q3IY5kXepw3R0:-NXFfLqV5doW0jDCYCQv:b-ayyllo) ([raw file](https://github.com/cockroachdb/pebble/blob/d96be4ea2e33b9286d8364352e08ad1d2d5ace0b/objstorage/objstorageprovider/sharedcache/shared_cache.go#L274)):*

[nit] it doesn't make much sense to have this constant, I'd remove it and move the comment to `lockState`
</blockquote></details>

Without it, I'd need to add a comment every time I did `lock += 1` or `lock > 0` to mention read locking. I agree it should always be zero... which is kind of odd. I'll leave it for now, based on the assumption you don't have strong feelings.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 484 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXCi1HTB3LA9IljI_65:-NXFfhX46jH1XdBRb9wd:bz5py0o) ([raw file](https://github.com/cockroachdb/pebble/blob/d96be4ea2e33b9286d8364352e08ad1d2d5ace0b/objstorage/objstorageprovider/sharedcache/shared_cache.go#L484)):*

[nit] worth an `invariants` assertion that the count is > 0
</blockquote></details>

Good idea! Done.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 491 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXCi47PBtbXjXHGIdEC:-NXFgQWICsOo7lw-zQYh:b-896fja) ([raw file](https://github.com/cockroachdb/pebble/blob/d96be4ea2e33b9286d8364352e08ad1d2d5ace0b/objstorage/objstorageprovider/sharedcache/shared_cache.go#L491)):*

[nit] `invariants` assertion that the count is -1
</blockquote></details>

Done!

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586#-:-NXGPBK05OlaZwg954HB:bnfp4nl)* status: 2 of 4 files reviewed, all discussions resolved (waiting on @sumeerbhola)

sumeerbhola (changes requested):
(apologies for the tardiness)

> Reviewed 1 of 3 files at r3, 1 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586#-:-NXGPPGe15XLXJFAjOyH:b-4li34y)* status: 3 of 4 files reviewed, 7 unresolved discussions (waiting on @joshimhoff and @RaduBerinde)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 143 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGFMP663fjcdn3ctfP:-NXGFMP663fjcdn3ctfQ:b-1bcm17) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L143)):*
> ```Go
> 	// the number of bytes read successfully. As is, we cannot tell if the readable.ReadAt
> 	// should be returned from cache.ReadAt. For now, the cache just swallows all
> 	// io.EOF errors.
> ```

I thought the `Readable.ReadAt` interface was simplified by @RaduBerinde to not return the bytes read successfully because the Pebble callers do not try to read beyond the end of the file, so we did not need to distinguish the different errors. Is the issue here that the cache does not know the size of the object and attempts to read whole blocks, so can go over the end? We could add a `Size()` method to `ReaderAt` and not read beyond. I don't have an opinion on which approach is preferable. 
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 150 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGGtDd9Bkhr0s6jS7f:-NXGGtDd9Bkhr0s6jS7g:b1ozj2t) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L150)):*
> ```Go
> 
> 	// TODO(josh): For writing back to the cache, we may want a concurrency-limited approach
> 	// that does batching, instead of what is below. I think it's reasonable tho to
> ```

nit: though
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 153 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGGwbD2Ct1JVJemfi5:-NXGGwbD2Ct1JVJemfi6:b1nezhv) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L153)):*
> ```Go
> 	// run a production experiment with what is below, before making adjustments. Note that
> 	// the writes being done are in multiples of the filesystem block size. The filesystem
> 	// might do okay with them.
> ```

Could you elaborate on "The filesystem might do ...". I don't understand what is being said there, and how it relates to the concurrency-limited issue.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 275 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGLwfGBz1Cayk1tijY:-NXGLwfGBz1Cayk1tijZ:b-y1fol) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L275)):*
> ```Go
> 	// locks tracks the number of distinct readers of the logical block in an SST.
> 	readLockTakenInc = 1
> 	writeLockTaken   = -1
> ```

I think we need a code comment explaining the purpose of these.
IIUC, the readLock (which is really a read reference count -- may be worth renaming it), is ensuring that a cache block is not evicted and overwritten while there are readers.
The writeLock is for the case where someone is currently populating the cache block with the desired data.

Semantically, I don't think these "locks" belong with the cache block index. They are representing what is happening with a cache block that is currently assigned to a `logicalBlockID`. At eviction time, we can evict a `logicalBlockID` if *it* has no readers or a writer. At read time, we have to wait when reading the data for a `logicalBlockID` if a writer is currently writing it. So I think they should be in a struct that is the value in the map.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 335 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGLDXN0Ii6-BnmXAGp:-NXGLDXN0Ii6-BnmXAGq:b-k7zel) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L335)):*
> ```Go
> 	// The data extent might cross cache block boundaries, hence the loop. In the hot
> 	// path, max two iterations of this loop will be executed, since reads are sized
> 	// in units of sstable block size.
> ```

currently if there are 2 cache blocks needed to satisfy this read, and the first block is not in the cache and the second one is, will we end up reading both?
If yes, I think we should fix this (TODO is fine). This is not an unlikely scenario if we are doing a reverse scan, since those iterate over sstable blocks in reverse order and due to cache block aligned reads will have read the suffix of the sstable block that will be needed next.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 347 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGNMGP61xNfrGZuL-2:-NXGNMGP61xNfrGZuL-3:b2v6dzk) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L347)):*
> ```Go
> 			return n, nil
> 		}
> 		if s.mu.locks[cacheBlockInd] == writeLockTaken {
> ```

Slightly unrelated to this if-block, multiple reads within the same few milliseconds (anything that is smaller than the object store read latency) that miss on the same `logicalBlockID` will not necessarily be rare. We should have a TODO to measure the occurrences of such racing reads that end up all doing the same object store read. IIUC, such racing reads will likely return with the previous `if !ok`.

Coming to this if-block, this seems to to be the case where someone else has already done the read, and is in the process of populating the cache block. By returning here, this caller will do a (wasted) read. We should consider waiting for that write to finish. Again a TODO is fine. 
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 481 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGP7s16Yiezd_eB5BX:-NXGP7s16Yiezd_eB5BY:b4j4zhi) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L481)):*
> ```Go
> }
> 
> // Should be inlined.
> ```

Does it inline? I don't know if this matters much. The latency cost of cache hits will probably be dominated by the disk read.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586)* status: 3 of 4 files reviewed, 8 unresolved discussions (waiting on @joshimhoff and @sumeerbhola)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 143 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGFMP663fjcdn3ctfP:-NXGR93f8KKD_ocLRgwb:b-24fq6r) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L143)):*

I thought the `Readable.ReadAt` interface was simplified by @RaduBerinde to not return the bytes read successfully because the Pebble callers do not try to read beyond the end of the file, so we did not need to distinguish the different errors. Is the issue here that the cache does not know the size of the object and attempts to read whole blocks, so can go over the end? We could add a `Size()` method to `ReaderAt` and not read beyond. I don't have an opinion on which approach is preferable. 
</blockquote></details>

This is correct, we may go over the size of the object here because of rounding up to the cache block size. The `Readable` already has a cheap `Size()` getter we can use for this.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 261 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGT_oi6g2VZRws_kVR:-NXGT_oi6g2VZRws_kVS:b-jfvycl) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L261)):*
> ```Go
> // Maps a logical block in an SST to an index of the cache block with the
> // file contents (to the "cache block index").
> type whereMap map[logicalBlockID]int64
> ```

[nit] consider defining a type for the cache block index
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 275 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGLwfGBz1Cayk1tijY:-NXGS6D-DLh3mRlSLHZk:b-3vdcku) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L275)):*

I think we need a code comment explaining the purpose of these.
IIUC, the readLock (which is really a read reference count -- may be worth renaming it), is ensuring that a cache block is not evicted and overwritten while there are readers.
The writeLock is for the case where someone is currently populating the cache block with the desired data.

Semantically, I don't think these "locks" belong with the cache block index. They are representing what is happening with a cache block that is currently assigned to a `logicalBlockID`. At eviction time, we can evict a `logicalBlockID` if *it* has no readers or a writer. At read time, we have to wait when reading the data for a `logicalBlockID` if a writer is currently writing it. So I think they should be in a struct that is the value in the map.
</blockquote></details>

I don't see the advantage of storing it in the map.. It is a property of a cache block and it's reasonable (and more memory-efficient, because of the hash map load factor) to keep the per-cache-block state in a simple table.

I support moving it to a struct that is documented as storing per-cache-block state. We will probably need to add the logical block ID to it too (to update the `whereMap` during eviction).
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 335 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGLDXN0Ii6-BnmXAGp:-NXGTm3b-ukt1Uiw5yVB:brtorln) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L335)):*

currently if there are 2 cache blocks needed to satisfy this read, and the first block is not in the cache and the second one is, will we end up reading both?
If yes, I think we should fix this (TODO is fine). This is not an unlikely scenario if we are doing a reverse scan, since those iterate over sstable blocks in reverse order and due to cache block aligned reads will have read the suffix of the sstable block that will be needed next.
</blockquote></details>

Correct..  We could look at what's available at the end of the range as well. If there is a piece available inside, we will re-read it (but I think that's what we want, we don't want to break up into multiple read requests for cost reasons). I agree a TODO is fine for now though.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 347 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGNMGP61xNfrGZuL-2:-NXGUcqi29acVLXRrTAm:b-v5bav4) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L347)):*

Slightly unrelated to this if-block, multiple reads within the same few milliseconds (anything that is smaller than the object store read latency) that miss on the same `logicalBlockID` will not necessarily be rare. We should have a TODO to measure the occurrences of such racing reads that end up all doing the same object store read. IIUC, such racing reads will likely return with the previous `if !ok`.

Coming to this if-block, this seems to to be the case where someone else has already done the read, and is in the process of populating the cache block. By returning here, this caller will do a (wasted) read. We should consider waiting for that write to finish. Again a TODO is fine. 
</blockquote></details>

Ah.. We assumed that the second request would wait for the first in the pebble block cache but I see that's not the case.. Seems like it would be a bigger win if we did this there no?

joshimhoff (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586)* status: 3 of 4 files reviewed, 8 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 153 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGGwbD2Ct1JVJemfi5:-NXGQcj07tGNWbS97PS3:b-u6an8p) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L153)):*

Could you elaborate on "The filesystem might do ...". I don't understand what is being said there, and how it relates to the concurrency-limited issue.
</blockquote></details>

Since the writes are done in units of filesystem block size already, and since the file system manages its own queue, I am not sure I can speak to the concrete benefit of either of the following changes:

- Batch up writing back blocks to the secondary cache
- Concurrency limit the number of writes to the secondary cache happening at once

I think it will be easier to speak to the need for those changes, once we benchmark the code as is.

Does it make sense now?

(I do think concurrency / rate limiting writes (maybe more the latter) could make sense as a prioritization scheme. Need fast reads more than fast writes, since reads face users. Maybe best to smooth out write rate as a result, when possible.)
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 275 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGLwfGBz1Cayk1tijY:-NXGYFPhB92Ug5wfzFIp:b-4mk0vq) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L275)):*
Will beef up commentary.

> So I think they should be in a struct that is the value in the map.

This is actually what I had initially :) 

Semantically, I don't see why it's wrong semantically to store a lock with the cache block index. What is locked is the range of bytes on disk pointed at by the cache block index, IMHO. Example: If there are active readers, it cannot be used to store a different logical block.

Sumeer, is your argument about clarity of code, or some concrete benefit to your suggested approach over what is here (e.g. less lock contention)? How strongly do you feel?
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 335 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGLDXN0Ii6-BnmXAGp:-NXGRMeWEEvI1qUeACk_:b-nv8v1i) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L335)):*

Correct..  We could look at what's available at the end of the range as well. If there is a piece available inside, we will re-read it (but I think that's what we want, we don't want to break up into multiple read requests for cost reasons). I agree a TODO is fine for now though.
</blockquote></details>

I will add a TODO. Thanks for noting the reverse scan use case.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 347 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGNMGP61xNfrGZuL-2:-NXGSPv65QUKlsQUHPcq:b-57a1uv) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L347)):*
Radu, not sure I follow what you have realized. I'm just gonna write out my current thinking re: Sumeer's Qs, so you can comment on it too.

> We should have a TODO to measure the occurrences of such racing reads that end up all doing the same object store read. IIUC, such racing reads will likely return with the previous `if !ok`.

Agree re: the behavior of the current code & will add a TODO.

> Coming to this if-block, this seems to to be the case where someone else has already done the read, and is in the process of populating the cache block. By returning here, this caller will do a (wasted) read. We should consider waiting for that write to finish. Again a TODO is fine.

An opinion I have related to this discussion is: Blocks should be written back to both the in-memory block cache AND the secondary cache, since having hot blocks in the secondary cache means no big latency hit in case of process restart. 

The reason I am not sure we need to optimize this case is: The seconder reader will hit the in-memory primary cache, since that is written to as part of the first sstable read, before the write back to the secondary cache (which is async unlike the write back the in-memory primary cache IIUC). 

Actually, ^^ is another reason that I think blocks should be written back to both the in-memory block cache AND the secondary cache.

Sumeer / Radu, let me know what you think about this, and either way I can beef up the comment / add a TODO.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 481 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGP7s16Yiezd_eB5BX:-NXGRn7G654aOAFLZ6-c:b-l03yfb) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L481)):*

Does it inline? I don't know if this matters much. The latency cost of cache hits will probably be dominated by the disk read.
</blockquote></details>

Do you suggest me checking if it inlines, or should I just remove the comment, since sort of a premature optimization anyway? Not sure how to check but could be a fun diversion.

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586)* status: 3 of 4 files reviewed, 8 unresolved discussions (waiting on @joshimhoff and @RaduBerinde)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 153 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGGwbD2Ct1JVJemfi5:-NXG_kIu0K0b_dq35Zvm:bkrlz00) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L153)):*

Since the writes are done in units of filesystem block size already, and since the file system manages its own queue, I am not sure I can speak to the concrete benefit of either of the following changes:

- Batch up writing back blocks to the secondary cache
- Concurrency limit the number of writes to the secondary cache happening at once

I think it will be easier to speak to the need for those changes, once we benchmark the code as is.

Does it make sense now?

(I do think concurrency / rate limiting writes (maybe more the latter) could make sense as a prioritization scheme. Need fast reads more than fast writes, since reads face users. Maybe best to smooth out write rate as a result, when possible.)
</blockquote></details>

Ack. Do we have a metric for the cumulative write (and read) cache bytes so we can see whether the cache write bandwidth is significant when compared to the compaction and flush bandwidth?
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 275 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGLwfGBz1Cayk1tijY:-NXGaJFdEEfxKqCz2qOF:b-b55ti8) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L275)):*
> I don't see the advantage of storing it in the map

I am not claiming it as an advantage. I am claiming that semantically it belongs in the `whereMap` entry since it only makes sense for a `logicalBlockID` that is present in the cache -- it represents that `logicalBlockID`'s readers and writers. More sophisticated eviction is also related to the state maintained for `logicalBlockID`s, so I don't think there is any need for any complex state associated with each cache block (other than the list of free cache blocks).

Regarding memory efficiency, I am not sure we need to optimize that now. And we have ideas to move the map to a kv store.

Coming back to where it semantically belongs, I personally find the following code a bit awkward when thinking about correctness, since I have to think about whether `s.mu.locks[cacheBlockInd]` means a lock for this logical block or some other logical block that may be using the same `cacheBlockInd`. The fact that a `cacheBlockInd` is the map value of at most one `logicalBlockID` is a higher level invariant that we are already relying on.
```
		k := logicalBlockID{
			filenum:               fileNum,
			offsetInUnitsOfBlocks: (ofs + int64(n)) / int64(s.blockSize),
		}
		s.mu.Lock()
		cacheBlockInd, ok := s.mu.where[k]
		if !ok {
			s.mu.Unlock()
			return n, nil
		}
		if s.mu.locks[cacheBlockInd] == writeLockTaken {
```

Same for the following eviction code. Whether something can be evicted is a property of the `logicalBlockID`, IMO, so having to look up something by the `cacheBlockInd` seems odd.

```
			for k1, v := range s.mu.where {
				// This implies that hot blocks will not be evicted as often, meaning the
				// eviction scheme is not quite random. I think that's fine for now, since we plan
				// to leverage a more sophisticated eviction scheme eventually anyway.
				lock := s.mu.locks[v]
				if lock >= readLockTakenInc || lock == writeLockTaken {
					continue
				}
```

I don't want to bike shed either, so if you all are unanimous with the current scheme, feel free to ignore me.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 481 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGP7s16Yiezd_eB5BX:-NXGdvWC2hE3WWoAomXj:bns1jpp) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L481)):*

Do you suggest me checking if it inlines, or should I just remove the comment, since sort of a premature optimization anyway? Not sure how to check but could be a fun diversion.
</blockquote></details>

In the sharedcache directory, do `go build -gcflags='-m -m' 2>  <output file>` and see what the output file says about inlining.
You may also find https://dave.cheney.net/high-performance-go-workshop/sydney-2019.html useful reading

joshimhoff (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586)* status: 3 of 4 files reviewed, 8 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 275 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGLwfGBz1Cayk1tijY:-NXGf0K15ryx8xPIqc7T:bls0zd0) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L275)):*

> I don't see the advantage of storing it in the map

I am not claiming it as an advantage. I am claiming that semantically it belongs in the `whereMap` entry since it only makes sense for a `logicalBlockID` that is present in the cache -- it represents that `logicalBlockID`'s readers and writers. More sophisticated eviction is also related to the state maintained for `logicalBlockID`s, so I don't think there is any need for any complex state associated with each cache block (other than the list of free cache blocks).

Regarding memory efficiency, I am not sure we need to optimize that now. And we have ideas to move the map to a kv store.

Coming back to where it semantically belongs, I personally find the following code a bit awkward when thinking about correctness, since I have to think about whether `s.mu.locks[cacheBlockInd]` means a lock for this logical block or some other logical block that may be using the same `cacheBlockInd`. The fact that a `cacheBlockInd` is the map value of at most one `logicalBlockID` is a higher level invariant that we are already relying on.
```
		k := logicalBlockID{
			filenum:               fileNum,
			offsetInUnitsOfBlocks: (ofs + int64(n)) / int64(s.blockSize),
		}
		s.mu.Lock()
		cacheBlockInd, ok := s.mu.where[k]
		if !ok {
			s.mu.Unlock()
			return n, nil
		}
		if s.mu.locks[cacheBlockInd] == writeLockTaken {
```

Same for the following eviction code. Whether something can be evicted is a property of the `logicalBlockID`, IMO, so having to look up something by the `cacheBlockInd` seems odd.

```
			for k1, v := range s.mu.where {
				// This implies that hot blocks will not be evicted as often, meaning the
				// eviction scheme is not quite random. I think that's fine for now, since we plan
				// to leverage a more sophisticated eviction scheme eventually anyway.
				lock := s.mu.locks[v]
				if lock >= readLockTakenInc || lock == writeLockTaken {
					continue
				}
```

I don't want to bike shed either, so if you all are unanimous with the current scheme, feel free to ignore me.

</blockquote></details>

Thanks for giving your reasoning in more detail! Gonna think it over a bit before responding.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 481 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGP7s16Yiezd_eB5BX:-NXGfA6LBXUAaVszAMyU:b-9755fn) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L481)):*

In the sharedcache directory, do `go build -gcflags='-m -m' 2>  <output file>` and see what the output file says about inlining.
You may also find https://dave.cheney.net/high-performance-go-workshop/sydney-2019.html useful reading
</blockquote></details>

Thank you!

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586)* status: 3 of 4 files reviewed, 8 unresolved discussions (waiting on @joshimhoff and @sumeerbhola)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 275 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGLwfGBz1Cayk1tijY:-NXGkBS82s0WdwdLbxQr:b-bnzso) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L275)):*

Thanks for giving your reasoning in more detail! Gonna think it over a bit before responding.
</blockquote></details>

I think you can frame it both ways. To me the lock state is telling me the state of the physical cache block (whether I can read it or not). Similarly, the eviction algorithm could be on top of cache blocks not logical blocks (e.g. if it's LRU we can have a LRU list of cache blocks not a LRU list of logical block IDs.

Maybe for now leave a TODO to investigate whether this should live in the map.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 347 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGNMGP61xNfrGZuL-2:-NXGjfHZ1D0cqhlWIDJH:bat61q4) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L347)):*

Radu, not sure I follow what you have realized. I'm just gonna write out my current thinking re: Sumeer's Qs, so you can comment on it too.

> We should have a TODO to measure the occurrences of such racing reads that end up all doing the same object store read. IIUC, such racing reads will likely return with the previous `if !ok`.

Agree re: the behavior of the current code & will add a TODO.

> Coming to this if-block, this seems to to be the case where someone else has already done the read, and is in the process of populating the cache block. By returning here, this caller will do a (wasted) read. We should consider waiting for that write to finish. Again a TODO is fine.

An opinion I have related to this discussion is: Blocks should be written back to both the in-memory block cache AND the secondary cache, since having hot blocks in the secondary cache means no big latency hit in case of process restart. 

The reason I am not sure we need to optimize this case is: The seconder reader will hit the in-memory primary cache, since that is written to as part of the first sstable read, before the write back to the secondary cache (which is async unlike the write back the in-memory primary cache IIUC). 

Actually, ^^ is another reason that I think blocks should be written back to both the in-memory block cache AND the secondary cache.

Sumeer / Radu, let me know what you think about this, and either way I can beef up the comment / add a TODO.
</blockquote></details>

The case is when we get a second read while we're reading data from blob storage. Currently we start a second read for the same data; it would be better if the second request would just wait for the first one to complete. But I'm arguing that it might be more beneficial to implement something like this at the in-memory block cache level rather than here.

sumeerbhola (changes requested):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586#-:-NXI0uaLB_ARLnr0zgYe:bg7igkv)* status: 3 of 4 files reviewed, 7 unresolved discussions (waiting on @joshimhoff and @RaduBerinde)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 275 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGLwfGBz1Cayk1tijY:-NXHpe2w0CCH0yz4muUN:b-lxoiya) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L275)):*
> the eviction algorithm could be on top of cache blocks not logical blocks

Agreed, that in some sense there are only cache blocks and some have data (and possibly readers/writers) and some are free. The map is only needed because we need a lookup data-structure for the cache blocks that have data. And those are the only ones that have ongoing readers/writers. Also those will be the only ones for which the hit related information is needed for eviction purposes. But yes, that unnecessary state can also be maintained for free blocks, and it doesn't cost much since in steady state the number of free blocks will be a tiny fraction of the total blocks.

So I withdraw my opinion -- it doesn't matter to me how this is setup now. I'll wait about having an opinion until we have the more sophisticated state for eviction, persistent map etc. :)
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 347 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGNMGP61xNfrGZuL-2:-NXHvIXY10wFcCfi9FGA:bkrqgfv) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L347)):*
> The reason I am not sure we need to optimize this case is: The seconder reader will hit the in-memory primary cache, since that is written to as part of the first sstable read, before the write back to the secondary cache (which is async unlike the write back the in-memory primary cache IIUC).

That's a good point.

> The case is when we get a second read while we're reading data from blob storage. Currently we start a second read for the same data; it would be better if the second request would just wait for the first one to complete. But I'm arguing that it might be more beneficial to implement something like this at the in-memory block cache level rather than here.

I agree that we could also put this in the in-memory block cache. But that cache is extremely performance sensitive wrt CPU cycles, so I am wary of adding waiting logic in there. My understanding of the lookup behavior is the following sequence:
- Get from block cache. If success or not a shared storage file, return.
- Get from shared cache. If success return.

Since we are falling through to the second step, and failure here requires a more expensive block storage read, we can do the waiting in the second step if there are other readers also encountering this miss. The CPU cycles for more complex code will matter less here.

btw, I've built such a cache in a past life -- all the readers of that block waited in a queue, and the first one was responsible for doing the read and populating the cache, and signaling the others when done. The one responsible for populating the cache could encounter a transient or permanent error. In that setting the code respected context cancellation, so the transient error could also be due to the higher-level request timing out -- we don't respect cancellation inside Pebble, but this has come up recently in https://github.com/cockroachdb/cockroach/issues/104262 (not that we need to change anything urgently in Pebble, but worth keeping in mind). A transient error would cause the reader to return the error to its caller and handoff the responsibility of the read to the next waiter. A permanent error would get recorded as a cache entry to avoid reading it again. I bring all this up because when reads were to the "local filesystem" we didn't worry much in the Pebble code about transient errors, but we need to handle them better now. I'm not quite sure what we are doing now for transient errors from object storage -- transparently retrying indefinitely on a transient error would makes the stuckness problem worse, given that contexts are not passed into Pebble. Glancing at a `cloud.ExternalStorage` implementation, s3_storage.go, it seems we don't retry at all. Is my understanding accurate?

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586#-:-NXI1FbVBS_ZTp5joYCS:bg7igkv)* status: 3 of 4 files reviewed, 7 unresolved discussions (waiting on @joshimhoff and @RaduBerinde)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586)* status: 3 of 4 files reviewed, 7 unresolved discussions (waiting on @joshimhoff and @sumeerbhola)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 347 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGNMGP61xNfrGZuL-2:-NXI5xwi9sLGtczatsM4:bn2mfpi) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L347)):*
> Glancing at a `cloud.ExternalStorage` implementation, s3\_storage.go, it seems we don't retry at all. Is my understanding accurate?

In `ReadFileAt` we use a `NewResumingReader` which retries. We will probably need to tweak the retry settings for our purposes.

joshimhoff (commented):
TTFRs! Interesting discussions all around.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586#-:-NXL2f1_CjthgQ9Pk3Ba:bs5n8tg)* status: 3 of 4 files reviewed, 7 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 143 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGFMP663fjcdn3ctfP:-NXL6EfMASuBGERMWMw3:b-smgriy) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L143)):*

This is correct, we may go over the size of the object here because of rounding up to the cache block size. The `Readable` already has a cheap `Size()` getter we can use for this.
</blockquote></details>

I updated the TODO to mention using `Size`. Radu, `ObjectReader` doesn't have a `Size` method. Do we just want to add one to the interface?

I'll try to follow up with a fix here before I go on vacation next week but a little short on time also.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 150 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGGtDd9Bkhr0s6jS7f:-NXL0bVs6avZGMkO1Hy-:bt2vr1b) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L150)):*

nit: though
</blockquote></details>

Fixed.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 153 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGGwbD2Ct1JVJemfi5:-NXL0m875DqSjA57drn-:bv4kc49) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L153)):*

Ack. Do we have a metric for the cumulative write (and read) cache bytes so we can see whether the cache write bandwidth is significant when compared to the compaction and flush bandwidth?
</blockquote></details>

Not yet. We should do an instrumentation PR soon. Agree with idea of having metrics for read & write BW coming from the cache.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 261 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGT_oi6g2VZRws_kVR:-NXL3bJs397mkmx9TREs:bcvjgzp) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L261)):*

[nit] consider defining a type for the cache block index
</blockquote></details>

Good idea. Done.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 275 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGLwfGBz1Cayk1tijY:-NXL1OuSEy9bxlnfQPWG:b-t8oxoq) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L275)):*

> the eviction algorithm could be on top of cache blocks not logical blocks

Agreed, that in some sense there are only cache blocks and some have data (and possibly readers/writers) and some are free. The map is only needed because we need a lookup data-structure for the cache blocks that have data. And those are the only ones that have ongoing readers/writers. Also those will be the only ones for which the hit related information is needed for eviction purposes. But yes, that unnecessary state can also be maintained for free blocks, and it doesn't cost much since in steady state the number of free blocks will be a tiny fraction of the total blocks.

So I withdraw my opinion -- it doesn't matter to me how this is setup now. I'll wait about having an opinion until we have the more sophisticated state for eviction, persistent map etc. :)

</blockquote></details>

Good discussion. Will leave as is for now.
___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 347 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGNMGP61xNfrGZuL-2:-NXL1e_U8Z1VjlCToYrR:b838kl7) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L347)):*
I've added a TODO about repeat reads. I noted that it could go in either the primary or secondary cache, and I linked this PR also, since there are additional details in this thread.

> I bring all this up because when reads were to the "local filesystem" we didn't worry much in the Pebble code about transient errors, but we need to handle them better now.

This is an interesting observation.

joshimhoff:
Still need to add a bit more to the comments as per earlier threads. Doing that now.

joshimhoff (commented):
Added additional commentary, TODOs, etc.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586#-:-NXLDHyEDdfDL24YmKoh:b1hlyiy)* status: 3 of 4 files reviewed, 7 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 481 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGP7s16Yiezd_eB5BX:-NXLD7093J2RQccMBi6m:b-2e9sfy) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L481)):*

Thank you!
</blockquote></details>

Doesn't inline:

```
~/g/s/g/c/p/o/o/sharedcache [fine-grained-locks] $ cat inline | grep 'drop'
./shared_cache.go:500:6: cannot inline (*shard).dropReadLock: function too complex: cost 163 exceeds budget 80
./shared_cache.go:512:6: cannot inline (*shard).dropWriteLock: function too complex: cost 163 exceeds budget 80
./shared_cache.go:338:7:     from (*shard).dropReadLock(s, cacheBlockInd) (call parameter) at ./shared_cache.go:387:18
./shared_cache.go:338:7:     from (*shard).dropReadLock(s, cacheBlockInd) (call parameter) at ./shared_cache.go:387:18
~/g/s/g/c/p/o/o/sharedcache [fine-grained-locks] $ cat inline | grep 'freeBlock'
./shared_cache.go:524:6: cannot inline (*shard).freeBlock: function too complex: cost 177 exceeds budget 80
./shared_cache.go:406:7:     from (*shard).freeBlock(s, k, cacheBlockInd) (call parameter) at ./shared_cache.go:490:15
./shared_cache.go:406:7:     from (*shard).freeBlock(s, k, cacheBlockInd) (call parameter) at ./shared_cache.go:490:15
```

I have adjusted the comment to note this.

RaduBerinde:
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 143 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGFMP663fjcdn3ctfP:-NXLGI6q1lGeCMwjumim:bx97m72) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L143)):*

I updated the TODO to mention using `Size`. Radu, `ObjectReader` doesn't have a `Size` method. Do we just want to add one to the interface?

I'll try to follow up with a fix here before I go on vacation next week but a little short on time also.
</blockquote></details>

`objstorage.Readable` has a `Size()` method. We can just pass the size to the cache, along with the `ObjectReader`

joshimhoff (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2586)* status: 3 of 4 files reviewed, 7 unresolved discussions

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 143 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGFMP663fjcdn3ctfP:-NXLG_9z4fdvXfIV-wwY:b-g6llew) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L143)):*

`objstorage.Readable` has a `Size()` method. We can just pass the size to the cache, along with the `ObjectReader`
</blockquote></details>

Makes sense. Thanks!

RaduBerinde:
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 143 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2586#-NXGFMP663fjcdn3ctfP:-NXLGThL68MG45gkMhYJ:b-j1bbka) ([raw file](https://github.com/cockroachdb/pebble/blob/cf8e66b7fba6685e9b1834559ef5b1471842ad5f/objstorage/objstorageprovider/sharedcache/shared_cache.go#L143)):*

`objstorage.Readable` has a `Size()` method. We can just pass the size to the cache, along with the `ObjectReader`
</blockquote></details>

More specifically -`sharedReadable` has a `size` field

PR #2589, opened 2023-06-02
docs: add nightly benchmark annotation for infra change

### LINKED ISSUE #2578: db: investigate unexplained ycsb/F/values=1024 perf improvement
Screenshot 2023-05-30 at 5 28 48 PM

We don't currently understand where this improvement came from. We should investigate it, ensure that we're not doing anything silly, and see if there were any regressions fixed that should then be backported.

jbowens:
A yet unidentified change to Cockroach's roachprod or roachtest tooling resulted in a significant change to the ycsb/F/values=1024 benchmark. Compactions started to be less effective, resulting in a larger build-up of read amplification and less back pressure on the workload.

Close #2578.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -65,6 +65,8 @@
             iterator structs (#1822)</div>
         <div class="annotation" data-date="20230215">Grandparent boundary
             compaction splitting</div>
+        <div class="annotation" data-date="20230516">Infrastructure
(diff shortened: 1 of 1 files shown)


nicktrav (approved):

jbowens:
TFTR!

PR #2601, opened 2023-06-06
ci: move asan / msan to scheduled workflow

nicktrav:
The ASAN and MSAN pipelines, while useful, often flake out during CI. This can block merging of a patch until fixed.

Move the ASAN and MSAN pipelines to run in a scheduled workflow, daily at midnight UTC.


.github/workflows/sanitizers.yaml
```diff
@@ -0,0 +1,32 @@
+name: Sanitizers
+
+on:
+  schedule:
+  - cron: "* * * * *" # Midnight UTC, daily.
```
nicktrav, line 5: Woops. This is wrong. Should be "0 0 * * *".

nicktrav:
TFTR!

PR #2604, opened 2023-06-06
objstorage: increase shared readahead to 1MB, fix EOF bug

RaduBerinde:
This change makes the read-ahead size configurable in the code and sets it to 1MB for shared objects.

We also fix a bug where read-ahead was causing us to read past the end of objects.

### DIFF
--- a/objstorage/objstorageprovider/readahead.go
+++ b/objstorage/objstorageprovider/readahead.go
@@ -11,15 +11,15 @@ const (
 	minFileReadsForReadahead = 2
 	// TODO(bilal): Have the initial size value be a factor of the block size,
 	// as opposed to a hardcoded value.
-	initialReadaheadSize = 64 << 10  /* 64KB */
-	maxReadaheadSize     = 256 << 10 /* 256KB */
+	initialReadaheadSize = 64 << 10 /* 64KB */
 )
 
 // readaheadState contains state variables related to readahead. Updated on
 // file reads.
 type readaheadState struct {
 	// Number of sequential reads.
-	numReads int64
+	numReads         int64
+	maxReadaheadSize int64
 	// Size issued to the next call to Prefetch. Starts at or above
 	// initialReadaheadSize and grows exponentially until maxReadaheadSize.
 	size int64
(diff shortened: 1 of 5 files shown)


joshimhoff (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2604#-:-NXMNtjfDxeqtxsL4c0d:b-uwkowd)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`objstorage/objstorageprovider/shared_readable.go` line 112 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2604#-NXMLKn38LJT9FxreIqH:-NXMLKn43T6iGko4vygJ:bceksu6) ([raw file](https://github.com/cockroachdb/pebble/blob/e45b290dacd4387ceae5381ec795c9efb828cc5a/objstorage/objstorageprovider/shared_readable.go#L112)):*
> ```Go
> 		if offset+int64(readaheadSize) > r.readable.size {
> 			readaheadSize = int(r.readable.size - offset)
> 			if readaheadSize < 0 {
> ```

Can you add some commentary re: when we expect this branch to get hit? In pebble today, would callers pass in an `offset` that is larger than the size of the file?
___
*[`objstorage/objstorageprovider/shared_readable.go` line 124 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2604#-NXMNaC-3mujEJbGEAaF:-NXMNaC-3mujEJbGEAaG:b-9fwqi3) ([raw file](https://github.com/cockroachdb/pebble/blob/e45b290dacd4387ceae5381ec795c9efb828cc5a/objstorage/objstorageprovider/shared_readable.go#L124)):*
> ```Go
> 		}
> 
> 		if err := r.readable.readInternal(ctx, r.readahead.data, offset, r.forCompaction); err != nil {
> ```

If `len(r.readahead.data` bytes are read and right after that is end of file, does `readInternal` return `nil`, as desired? I think so, reading the documentation of the called interfaces, but just confirming.

PR #2605, opened 2023-06-06
Makefile: separate target for generating test data

### LINKED ISSUE #2383: make generate is stale
One problem is https://github.com/cockroachdb/pebble/pull/1752#issuecomment-1448246697

The other problem is that generation of test fixtures is non-deterministic (and in some cases, the generated fixtures cause test failures).

RaduBerinde:
We use `go:generate` to invoke programs that generate tool test data. This test data is non-deterministic (and tests don't generally pass when data is regenerated). This is not appropriate for the `make generate` target which is supposed to be a no-op on a "clean" branch.

This change separates these into a new `generate-test-data` target.

Fixes #2383.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,8 @@ all:
 	@echo "  make stressmeta"
 	@echo "  make crossversion-meta"
 	@echo "  make mod-update"
+	@echo "  make generate"
+	@echo "  make generate-test-data"
 	@echo "  make clean"
 
 override testflags :=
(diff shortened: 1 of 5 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2605#-:-NXGp9o_9NQVpxKEfZR9:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

PR #2608, opened 2023-06-06
cmd/pebble: sample read amp during write-only ycsb workloads

### LINKED ISSUE #2590: cmd/pebble: calculate r-amp for ycsb/F workloads
The ycsb benchmarks measure the true read amplification of reads performed by the benchmark. The ycsb/F workload performs no reads, so it produces no r-amp calculation. This can produce deceiving results, like #2578 where slower compactions resulted in accumulation of more compaction debt and read amplification, improving the overall write throughput. If we had a r-amp statistic available, we would be able to observe that the improved throughput and w-amp came at the cost of r-amp.

A simple way to resolve this would be to allow a very small percentage of reads as a part of the ycsb/F workload. Alternatively, we could sample the r-amp reported by `DB.Metrics` periodically.

jbowens:
Previously, if a ycsb benchmark performed no reads (eg, workload F), then no read amplification was reported because no reads were performed. Now in these cases a separate goroutine will periodically sample the database's worst-case read amplification.

Close #2590.

### DIFF
--- a/cmd/pebble/ycsb.go
+++ b/cmd/pebble/ycsb.go
@@ -350,6 +350,14 @@ func (y *ycsb) init(db DB, wg *sync.WaitGroup) {
 	y.limiter = maxOpsPerSec.newRateLimiter()
 
 	wg.Add(concurrency)
+
+	// If this workload doesn't produce reads, sample the worst case read-amp
+	// from Metrics() periodically.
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2608#-:-NXI75_K-zB5_8FRBxDR:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @itsbilal)

jbowens:
tftr!

PR #2609, opened 2023-06-07
db: don't pace deletions of shared objects

RaduBerinde:
Delete pacing is used to avoid slowdowns on certain SSDs caused by
deleting many large files. We don't want this pacing when deleting
shared objects.


itsbilal (approved):
LGTM, good catch!

RaduBerinde:
TFTR!

PR #2610, opened 2023-06-07
db: Call maybeScheduleFlush before waiting on cond lock

### LINKED ISSUE #2597: metamorphic: recent metamorphic runs appear to deadlock
All recent nightlies of the metamorphic tests going back to ~April 27th seem to stall for hours.

Possibly related to https://github.com/cockroachdb/pebble/commit/497c8cc0f87e24902ebd00345c572b6ffebf51b4.

itsbilal:
After #2486 went in, it became more likely for past calls to `maybeScheduleFlush` to not do anything as the mutable memtable during an ingestion would no longer count towards a flush's bytes as it has a nonzero writerRef. If this is followed by the mutable memtable filling up and hitting the write stall threshold, we could end up in a situation where we (an ingesting goroutine) is holding the commit mutex and there is no remaining goroutine to call maybeScheduleFlush. There could possibly be calls to maybeScheduleDelayedFlush, but those would wait for the commit mutex and so wouldn't do anything.

This change calls maybeScheduledFlush before waiting on the compaction cond lock just to ensure a flush is happening.

Fixes #2597.

### DIFF
--- a/db.go
+++ b/db.go
@@ -2136,6 +2136,7 @@ func (d *DB) makeRoomForWrite(b *Batch) error {
 					})
 				}
 				now := time.Now()
+				d.maybeScheduleFlush()
 				d.mu.compact.cond.Wait()
 				if b != nil {
 					b.commitStats.MemTableWriteStallDuration += time.Since(now)


itsbilal:
@jbowens I'm not 100% sure this explanation makes a lot of sense - it feels somewhat handwavy. Feel free to dissect it and let me know if you want me to elaborate. Also not sure how I'd write a test to test this anyway, but the metamorphic test seeds that were earlier hanging in 1 out of every 10 runs, now go for thousands of runs without any issues.

itsbilal:
this is ready for another look, incorporated comments from slack

bananabrick (commented):
:lgtm: Assuming the tests pass.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2610#-:-NXRfHg-CUQs01dAOMfn:br9xkiv)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`ingest.go` line 1079 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2610#-NXRewv088AMPoHCU-iK:-NXRewv088AMPoHCU-iL:b-6lu1i3) ([raw file](https://github.com/cockroachdb/pebble/blob/2504572834987a534685fd7842a78f7a2524ee37/ingest.go#L1079)):*
> ```Go
> 	d.mu.versions.logLock()
> 
> 	scheduleFlush := false
> ```

nit: Are we delaying the call to `maybeScheduleFlush` to not slow down ingestion? I think calling `maybeScheduleFlush` should be cheap. Should we do it here rather than delaying?

bananabrick (approved):
oops

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2610#-:-NXRffOK0bVW4I0N7n4W:bb7m3jc)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

jbowens (commented):
filed #2614—I think eventually we should rework this to avoid abusing the `writerRefs` reference count. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2610#-:-NXRlUKvA0MP0-n9Jjmr:b9fgf0z)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @bananabrick and @itsbilal)

___
*[`ingest.go` line 1079 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2610#-NXRewv088AMPoHCU-iK:-NXRk6dM3HaANlDN5Zfp:b-t8nn94) ([raw file](https://github.com/cockroachdb/pebble/blob/2504572834987a534685fd7842a78f7a2524ee37/ingest.go#L1079)):*

nit: Are we delaying the call to `maybeScheduleFlush` to not slow down ingestion? I think calling `maybeScheduleFlush` should be cheap. Should we do it here rather than delaying?
</blockquote></details>

curious about this too.

jbowens (commented):
I think we'll need to backport this change to 23.1 because #2486 was backported to 23.1.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2610#-:-NXVTsuA1vwXvOOvagW8:bkj6cn3)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`ingest.go` line 1079 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2610#-NXRewv088AMPoHCU-iK:-NXRmH-O-siHls9Zkr8s:b-7mqtvu) ([raw file](https://github.com/cockroachdb/pebble/blob/2504572834987a534685fd7842a78f7a2524ee37/ingest.go#L1079)):*

It is just out of readability - it makes it abundantly clear that we will beat it to the manifest application. Also pairs well with `maybeScheduleCompaction`.
</blockquote></details>

I think we should move it up here then; it would be a correctness issue if the `writerUnref` call didn't. also bea

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2610)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`ingest.go` line 1079 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2610#-NXRewv088AMPoHCU-iK:-NXVUAIgEWa58uvsReg2:b-ujst41) ([raw file](https://github.com/cockroachdb/pebble/blob/2504572834987a534685fd7842a78f7a2524ee37/ingest.go#L1079)):*

I think we should move it up here then; it would be a correctness issue if the `writerUnref` call didn't. also bea
</blockquote></details>

You can ignore this; I would've preferred the explicitness of scheduling the flush immediately at the unref site, but it's not a big deal.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2610)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`ingest.go` line 1079 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2610#-NXRewv088AMPoHCU-iK:-NXuTvk-1yrsC4JohWVY:bj0keqg) ([raw file](https://github.com/cockroachdb/pebble/blob/2504572834987a534685fd7842a78f7a2524ee37/ingest.go#L1079)):*

What if an error happens below and we don't schedule the flush? Is that fine because the errors below are unrecoverable?
</blockquote></details>

Hrm, it looks like the `findTargetLevel` error path below _is_ recoverable, at least in that it won't crash the process like other commit pipeline errors. So I think it's possible for `findTargetLevel` to error, leaving the flush unscheduled. Seems like a bug.

itsbilal:
*[`ingest.go` line 1079 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2610#-NXRewv088AMPoHCU-iK:-NXvgrg35jiY5HFHX4Ve:b-jfuclc) ([raw file](https://github.com/cockroachdb/pebble/blob/2504572834987a534685fd7842a78f7a2524ee37/ingest.go#L1079)):*

Hrm, it looks like the `findTargetLevel` error path below _is_ recoverable, at least in that it won't crash the process like other commit pipeline errors. So I think it's possible for `findTargetLevel` to error, leaving the flush unscheduled. Seems like a bug.
</blockquote></details>

Yeah, that seems like a bug. I'll tackle it momentarily or file an issue about it. Thanks for catching

PR #2611, opened 2023-06-08
db: make version set virtual sstable tests deterministic

bananabrick:
We were creating two virtual sstables m1, m2, in these tests by virtualizing a file f. m1.Size and m2.Size were being set to f.Size / 2. If f.Size is odd, then m1.Size + m2.Size will be less than f.Size because we round down. This can cause test failures.

Instead, we make sure that m1.Size + m2.Size will equal f.Size.


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2611#-:-NXQTNIU3TLuwz_P1YAI:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @itsbilal and @RahulAggarwal1016)

bananabrick:
Thanks.

PR #2615, opened 2023-06-08
objstorage: support heteorogeneous Storage backends

### LINKED ISSUE #2594: provider: add support for non-standard buckets
Currently, the provider and its catalog implies that every shared file lives in one `shared.Storage` - the one the DB was instantiated with. This doesn't need to be the case; we could have multiple separate `shared.Storage` instances pointing to different buckets/regions/clouds/drivers for files present in backups that we need to progressively restore from. Provided (hah) that there's a constructor/factory for a `shared.Storage` available, Pebble should be able to create a `shared.Storage` based on info passed in during ingest-time.

As part of this change, the provider's Catalog/manifest would need to be updated to support storing cloud provider/bucket/region/credential info, and really anything else needed to create a new `shared.Storage`, and to disambiguate these files from the main `shared.Storage` (if one is available).

In the future, this functionality will also be used for multi-region disaggregated storage clusters.

RaduBerinde:
Currently objstorage uses a single `shared.Storage` implementation; consequently all shared objects must reside in a single backend/bucket, which must be known at store open time.

For the progressive restore case, these assumptions don't hold. We want to attach objects that come from an arbitrary bucket at runtime.

This change implements the following design:
 - instead of a single `shared.Storage` instance, we use a `shared.Factory` which is able to return multiple shared.Storage objects.
 - we define shared.Locator as an opaque string that is associated with shared.Storage backend.
 - all shared objects store the Locator in the catalog.
 - if we want Pebble to default to creating new SSTs on shared storage, we must enable it explicitly and provide a locator.
 - when we attach shared objects, we have to pass a locator.

Fixes #2594.


sumeerbhola (changes requested):
looks good

> Reviewed 17 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2615#-:-NXl0PiA5c98_prflNkS:b-47fafm)* status: all files reviewed, 5 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`objstorage/objstorage.go` line 235 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2615#-NXkviHu14YivsvntPF1:-NXkviHu14YivsvntPF2:brl5cur) ([raw file](https://github.com/cockroachdb/pebble/blob/806c8f1edc18f187d9a91f5b175d9d3adf7d593e/objstorage/objstorage.go#L235)):*
> ```Go
> 
> 	// AttachSharedObjects registers existing shared objects with this provider.
> 	AttachSharedObjects(locator shared.Locator, objs []SharedObjectToAttach) ([]ObjectMetadata, error)
> ```

Is there a reason the `shared.Locator` is not serialized in the `SharedObjectBacking`? I was assuming the `SharedObjectBacking` would be self-contained when passed around.
___
*[`objstorage/objstorageprovider/provider.go` line 294 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2615#-NXkwi9G4E7S_zCNjH03:-NXkwi9G4E7S_zCNjH04:b-6pqf1r) ([raw file](https://github.com/cockroachdb/pebble/blob/806c8f1edc18f187d9a91f5b175d9d3adf7d593e/objstorage/objstorageprovider/provider.go#L294)):*
> ```Go
> // IsNotExistError is part of the objstorage.Provider interface.
> func (p *provider) IsNotExistError(err error) bool {
> 	return oserror.IsNotExist(err)
> ```

Is this change a consequence of the `errors.Mark` changes earlier, so all "not exist" errors returned by the provider will have `oserror.IsNotExist(err)` return true?
___
*[`objstorage/objstorageprovider/sharedobjcat/version_edit.go` line 86 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2615#-NXkyiZsAMNGx8JuIbkx:-NXkyiZsAMNGx8JuIbky:blgqa) ([raw file](https://github.com/cockroachdb/pebble/blob/806c8f1edc18f187d9a91f5b175d9d3adf7d593e/objstorage/objstorageprovider/sharedobjcat/version_edit.go#L86)):*
> ```Go
> 			buf = append(buf, []byte(meta.Locator)...)
> 		}
> 		buf = binary.AppendUvarint(buf, 0)
> ```

why this `binary.AppendUvarint(buf, 0)`? I suppose it tells us that `tagNewObjectLocator` was not written, and serves as a record terminator. Could you add a code comment.
___
*[`objstorage/objstorageprovider/provider_test.go` line 322 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2615#-NXl-UIfE_Bjm97uG1sb:-NXl-UIfE_Bjm97uG1sc:b-2lb5oq) ([raw file](https://github.com/cockroachdb/pebble/blob/806c8f1edc18f187d9a91f5b175d9d3adf7d593e/objstorage/objstorageprovider/provider_test.go#L322)):*
> ```Go
> 	b1, err := h1.Get()
> 	require.NoError(t, err)
> 	h1.Close()
> ```

why are we closing the backing handle before the attach? Is it because in this test we do not need to worry about keeping the ref alive?
___
*[`objstorage/objstorageprovider/provider_test.go` line 331 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2615#-NXl-quO8aGQH9CiSL8P:-NXl-quO8aGQH9CiSL8Q:b-za38d2) ([raw file](https://github.com/cockroachdb/pebble/blob/806c8f1edc18f187d9a91f5b175d9d3adf7d593e/objstorage/objstorageprovider/provider_test.go#L331)):*
> ```Go
> 		require.NoError(t, r.ReadAt(ctx, data, 0))
> 		r.Close()
> 		require.Equal(t, salt, checkData(t, 0, data))
> ```

nit: can you hoist this `checkObject` earlier so that there isn't such a long gap between the comment `// Now attach p1's object (in the "foo" store) to p2.` and the actual attach.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2615#-:-NXl6YFg0dwtZpSTc6Zl:b-814g8k)* status: all files reviewed, 4 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`objstorage/objstorage.go` line 235 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2615#-NXkviHu14YivsvntPF1:-NXl4Tn21POVjtOMPUSA:bcs1x8z) ([raw file](https://github.com/cockroachdb/pebble/blob/806c8f1edc18f187d9a91f5b175d9d3adf7d593e/objstorage/objstorage.go#L235)):*

Is there a reason the `shared.Locator` is not serialized in the `SharedObjectBacking`? I was assuming the `SharedObjectBacking` would be self-contained when passed around.
</blockquote></details>

I am open to doing it either way.. My only hesitation is that depending how it's used `Locator` might contain secrets (which makes uses of Backing more sensitive). @itsbilal what do you think?
___
*[`objstorage/objstorageprovider/provider.go` line 294 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2615#-NXkwi9G4E7S_zCNjH03:-NXl5emS5QWu7aBIsCBq:b-lwlr47) ([raw file](https://github.com/cockroachdb/pebble/blob/806c8f1edc18f187d9a91f5b175d9d3adf7d593e/objstorage/objstorageprovider/provider.go#L294)):*

Is this change a consequence of the `errors.Mark` changes earlier, so all "not exist" errors returned by the provider will have `oserror.IsNotExist(err)` return true?
</blockquote></details>

Right, added a comment.
___
*[`objstorage/objstorageprovider/sharedobjcat/version_edit.go` line 86 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2615#-NXkyiZsAMNGx8JuIbkx:-NXl5yU58eAWkV82F9oq:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/806c8f1edc18f187d9a91f5b175d9d3adf7d593e/objstorage/objstorageprovider/sharedobjcat/version_edit.go#L86)):*

why this `binary.AppendUvarint(buf, 0)`? I suppose it tells us that `tagNewObjectLocator` was not written, and serves as a record terminator. Could you add a code comment.
</blockquote></details>

Done.
___
*[`objstorage/objstorageprovider/provider_test.go` line 322 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2615#-NXl-UIfE_Bjm97uG1sb:-NXl6Kkt6LUQiBFT9rMY:bt2vr1b) ([raw file](https://github.com/cockroachdb/pebble/blob/806c8f1edc18f187d9a91f5b175d9d3adf7d593e/objstorage/objstorageprovider/provider_test.go#L322)):*

why are we closing the backing handle before the attach? Is it because in this test we do not need to worry about keeping the ref alive?
</blockquote></details>

Fixed.

sumeerbhola (commented):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2615)* status: all files reviewed, 4 unresolved discussions (waiting on @itsbilal)

___
*[`objstorage/objstorage.go` line 235 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2615#-NXkviHu14YivsvntPF1:-NXlFGrdE5F1NFE0pJZt:bc7ovhs) ([raw file](https://github.com/cockroachdb/pebble/blob/806c8f1edc18f187d9a91f5b175d9d3adf7d593e/objstorage/objstorage.go#L235)):*

I am open to doing it either way.. My only hesitation is that depending how it's used `Locator` might contain secrets (which makes uses of Backing more sensitive). @itsbilal what do you think?
</blockquote></details>

I agree that we should constrain `SharedObjectBacking` to not have secrets.

But we'll need to pass around something like a locator name between the CRDB nodes for this attach to succeed. We could make it a requirement that the `shared.Locator` does not have any secrets, and any indirection, including finding and using any secrets, has to be done by the `shared.StorageFactory` implementation.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2615)* status: all files reviewed, 4 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`objstorage/objstorage.go` line 235 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2615#-NXkviHu14YivsvntPF1:-NXlMjzHE2cVIAyY4c03:bdw8ba8) ([raw file](https://github.com/cockroachdb/pebble/blob/806c8f1edc18f187d9a91f5b175d9d3adf7d593e/objstorage/objstorage.go#L235)):*
> We could make it a requirement that the `shared.Locator` does not have any secrets, and any indirection, including finding and using any secrets, has to be done by the `shared.StorageFactory` implementation.

Yes, Bilal and I agree we should go with this approach.

RaduBerinde (commented):
Updated to encode the locator in the backing and document that it can't contain secrets.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2615#-:-NXlTybp-7_sr-vW_R_O:bxmebd7)* status: 9 of 18 files reviewed, 4 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

itsbilal:
*[`objstorage/objstorage.go` line 235 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2615#-NXkviHu14YivsvntPF1:-NXmWjw0EkhqZmrlxJVl:b7xlzlh) ([raw file](https://github.com/cockroachdb/pebble/blob/806c8f1edc18f187d9a91f5b175d9d3adf7d593e/objstorage/objstorage.go#L235)):*

> We could make it a requirement that the `shared.Locator` does not have any secrets, and any indirection, including finding and using any secrets, has to be done by the `shared.StorageFactory` implementation.

Yes, Bilal and I agree we should go with this approach.
</blockquote></details>

Yes, we're good with requiring `implicit` auth and really anything that keeps credentials and secrets out of locators.

(Gonna review the rest soon, thanks for doing this!)

RaduBerinde:
TFTR!

PR #2623, opened 2023-06-11
sharedcache: refactor block math

RaduBerinde:
This change moves all offset/block-index math to a special `blockMath` type and switches to bit operations instead of division and multiplication.


joshimhoff:
Looks nice!

PR #2620, opened 2023-06-09
db: add Metric.Keys.MissizedTombstoneCount

jbowens:
Add a new metric that exposes the count of DELSIZED tombstones that are missized when they drop data from the LSM. This is useful within unit tests in client code to ensure that we're appropriately sizing tombstones.

PR #2634, opened 2023-06-14
db: tighten missized tombstone metric count

jbowens:
This commit refactors elements of #2620 and #2537 to more narrowly define what consitutes a 'missized tombstone.'

Previously, a DELSIZED tombstone could be counted as missized if it deleted a user key that had more than one internal key within the LSM. That lax accounting made it more difficult to use the count for asserting general correctness of client code.  The new calculation ensures that if the most recent version of a key was an accurately-sized SET when the DELSIZED tombstone was committed, the tombstone will not be regarded as missized regardless of internal LSM state.

### DIFF
--- a/compaction_iter.go
+++ b/compaction_iter.go
@@ -881,28 +881,104 @@ func (i *compactionIter) deleteSizedNext() (*base.InternalKey, []byte) {
 		i.value = i.valueBuf
 	}
 
-	// Loop through all the keys within this stripe that are skippable,
-	// totalling the size of their keys and values.
+	// Loop through all the keys within this stripe that are skippable.
 	i.pos = iterPosNext
-	var elidedSize uint64
 	for i.nextInStripe() == sameStripeSkippable {
-		elidedSize += uint64(len(i.iterKey.UserKey)) + uint64(len(i.iterValue))
-
-		if i.iterKey.Kind() == InternalKeyKindDeleteSized {
-			// We encountered a DELSIZED that's deleted by the original
-			// DELSIZED. These tombstones could've been intended to delete two
-			// distinct values, eg:
+		switch i.iterKey.Kind() {
+		case InternalKeyKindDelete, InternalKeyKindDeleteSized:
+			// We encountered a tombstone (DEL, or DELSIZED) that's deleted by
+			// the original DELSIZED tombstone. This can happen in two cases:
+			//
+			// (1) These tombstones were intended to delete two distinct values,
+			//     and this DELSIZED has already dropped the relevant key. For
+			//     example:
 			//
 			//     a.DELSIZED.9   a.SET.7   a.DELSIZED.5   a.SET.4
 			//
-			// If a.DELSIZED.9 has already deleted a.SET.7, its size has already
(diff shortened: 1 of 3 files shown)


sumeerbhola (changes requested):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2634#-:-NXzJKroA7ElE_UefQ1Y:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`compaction_iter.go` line 919 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2634#-NXzI3PlEOzsxjTCSGgd:-NXzI3PlEOzsxjTCSGge:b-k02t52) ([raw file](https://github.com/cockroachdb/pebble/blob/98840df932da94c232591af4033881511419a4c8/compaction_iter.go#L919)):*
> ```Go
> 				// The original DELSIZED key was missized. The key that the user
> 				// thought they were deleting does not exist.
> 				i.stats.countMissizedDels++
> ```

For  `a.DELSIZED.9   a.SET.7   a.DELSIZED.5   a.SET.4`
if the DELSIZED at 9 was missized we have incremented the count already and we will count it again here, yes?
I think that is fine, but worth a code comment that we cannot avoid such double counting -- we are mainly using this to check for bugs.

And we need to count here because the original could have been:
`a.DELSIZED.9   a.DELSIZED.5   a.SET.4`
and we have no way to distinguish these two original histories in the context of this compaction.

Is my understanding correct?
___
*[`compaction_iter.go` line 951 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2634#-NXzHXhZ-_emTiwu2wAo:-NXzHXh_70beNzPW_cM9:b-b2baeh) ([raw file](https://github.com/cockroachdb/pebble/blob/98840df932da94c232591af4033881511419a4c8/compaction_iter.go#L951)):*
> ```Go
> 				return &i.key, i.value
> 			}
> 			// The deleted key is some other key kind. Check that its size
> ```

Can this say "the deleted key is not a DEL, DELSIZED, and the DELSIZED has a positive size". The "deleted key is some other key kind" is a bit vague.

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2634#-:-NXzb6la01MmtN4L0PWe:bg7igkv)* status: all files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`compaction_iter.go` line 980 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2634#-NXzb3BF1ZptfEhR1fff:-NXzb3BF1ZptfEhR1ffg:b14c4a2) ([raw file](https://github.com/cockroachdb/pebble/blob/1a334455bf685e5978c6816bfce89ef443ddc45b/compaction_iter.go#L980)):*
> ```Go
> 			// NB: We remove the value regardless of whether the key was sized
> 			// appropriately. The size encoded is 'consumed' the first time it
> 			// meets a key that it deletes.
> ```

This comment is helpful.

jbowens:
TFTR!

PR #2636, opened 2023-06-14
objstorage: support custom object names

RaduBerinde:
For progressive restore the shared object names are not controlled by Pebble. This commit adds support in the objstorage provider for custom object names; these names are encoded into the shared object catalog and shared object backings.

In addition, we add a new `CreateSharedObjectBacking` API which can be used to create a `SharedObjectBacking` for such an object.


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2636)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @sumeerbhola)

___
*[`objstorage/objstorageprovider/shared.go` line 129 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2636#-NXulNvfB_bMfjK0Wub0:-NXulNvfB_bMfjK0Wub1:brqrjeh) ([raw file](https://github.com/cockroachdb/pebble/blob/04e94bd95aa87e074feeff336508235eac25a7fe/objstorage/objstorageprovider/shared.go#L129)):*
> ```Go
> 		return false
> 	}
> 	return meta.IsShared() && (meta.Shared.CustomObjectName != "" || meta.Shared.CreatorID != p.shared.creatorID)
> ```

@itsbilal An external object (like one from progressive restore) should be deemed foreign, correct? But I think we have some assertions that foreign sstables must always be virtual, which would not need to be the case for progressive restore.

RaduBerinde:
@itsbilal friendly ping.

itsbilal (approved):
LGTM, sorry for the late review! Thanks for doing this

objstorage/objstorageprovider/shared_obj_name.go
```diff
@@ -37,6 +40,11 @@ func sharedObjectRefName(
 	if meta.Shared.CleanupMethod != objstorage.SharedRefTracking {
 		panic("ref object used when ref tracking disabled")
 	}
+	if meta.Shared.CustomObjectName != "" {
+		return fmt.Sprintf(
+			"%s.ref.%d.%06d", meta.Shared.CustomObjectName, refCreatorID, refFileNum.FileNum(),
```
itsbilal, line 45: I'm good with this implementation, but we might not need to support the combination of `CustomObjectName != "" && CleanupMethod == SharedRefTracking`.

RaduBerinde: Agreed, it was just easy enough to include just in case.

RaduBerinde:
TFTR!

PR #2637, opened 2023-06-14
pebble: use token bucket for rate limiting

RaduBerinde:
Switch rate limiter from using a copy of `x/time/rate` to the
TokenBucket from Cockroach, which has a nicer interface.

We change `rate.Limiter` to a thin wrapper that provides thread safety
and a few convenience methods.


RaduBerinde:
TFTR!

PR #2638, opened 2023-06-15
ingest: Call maybeScheduleFlush immediately in ingestApply

itsbilal:
Previously we would delay the call to maybeScheduleFlush in ingestApply mostly out of readability. However it's safe to call it directly as the flush will wait for the manifest lock (which we are holding in that method) before it applies anyway. Furthermore, calling it directly avoids a bug where error cases in ingestApply could lead to us not scheduling a flush for a flushable memtable.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -1611,14 +1611,16 @@ func (d *DB) ingestApply(
 	// returns must unlock the manifest.
 	d.mu.versions.logLock()
 
-	scheduleFlush := false
 	if mut != nil {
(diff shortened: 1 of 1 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2638#-:-NXzqmsE2iN4sMIU4HAS:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bananabrick)

itsbilal:
TFTR!

PR #2641, opened 2023-06-16
db: refactor obsolete file deletion code

### LINKED ISSUE #2630: db: TestSharedObjectDeletePacing is flaky
Seen here: https://github.com/cockroachdb/pebble/actions/runs/5247409884/jobs/9477540951

Reproduces on Linux, locally:

```console
$ go test -run TestSharedObjectDeletePacing -count 1 -exec 'stress ' -timeout 0 -test.v .
...
=== RUN   TestSharedObjectDeletePacing
2023/06/12 16:20:38 shared storage configured; no creatorID yet
2023/06/12 16:20:38 shared storage creatorID set to 1
    compaction_test.go:3979: compaction timed out, possibly due to incorrect deletion pacing
--- FAIL: TestSharedObjectDeletePacing (60.00s)
FAIL


ERROR: exit status 1

1m10s: 8200 runs so far, 14 failures (0.17%)
```

RaduBerinde:
The code around obsolete file deletion is very convoluted and there is
one known race (described in #2630).

This change reworks the deletion code as follows:
 - we add a cleanup manager which has a background goroutine.
 - all file and object deletions happen in the manager's background
   goroutine.
 - callers can enqueue cleanup jobs with the manager
 - tests can wait for outstanding cleanup jobs to finish

We also add a missing call to `deleteObsoleteObjects` from the ingest
path (which now can cause files to be removed through excising).

Finally, we fix an incorrect `Remove` call in the error path of
`Checkpoint()` (which was missing a `PathJoin`).

Fixes #2630.


jbowens (commented):
Looking good!

> Reviewed 11 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2641#-:-NY5Gwch9NnZ6H5NMTbx:br2sbr6)* status: 11 of 17 files reviewed, 8 unresolved discussions (waiting on @RaduBerinde)

___
*[`cleaner.go` line 50 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5EBoG89pG9YpgxWKk:-NY5EBoG89pG9YpgxWKl:b-ew6rlu) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/cleaner.go#L50)):*
> ```Go
> // In practice, we should rarely have more than a couple of jobs (in most cases
> // we Wait() after queueing a job).
> const jobsChLen = 10000
> ```

is this maybe _too_ big? With 8-byte words, 80KB?
___
*[`cleaner.go` line 103 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5D_oeEPH5dA1MaZH_:-NY5D_oeEPH5dA1MaZHa:bfi5dlg) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/cleaner.go#L103)):*
> ```Go
> // Note that logs in obsoleteFiles might be recycled instead of deleted.
> //
> // Returns a channel that is closed when the job completes.
> ```

Looks like this comment is out-of-date.
___
*[`cleaner.go` line 120 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5FIP_DpX3w51-NAM1:-NY5ErmwA8w0WsBgOnMD:byhs613) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/cleaner.go#L120)):*
> ```Go
> 		}
> 		// Something is terribly wrong... Just drop the job.
> 		cm.opts.Logger.Infof("cleanup jobs queue full")
> ```

this makes me nervous even though it shouldn't happen. since enqueues are already mutex-protected, should we use a preallocated slice instead of a channel? if the slice becomes full, we can always suffer the allocation to grow it.
___
*[`compaction.go` line 2636 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5G9fMERDVukxoRKDC:-NY5G9fMERDVukxoRKDD:bbnypby) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/compaction.go#L2636)):*
> ```Go
> 		d.updateTableStatsLocked(ve.NewFiles)
> 	}
> 	d.deleteObsoleteFiles(jobID, true /* waitForCompletion */)
> ```

I think this is a behavior change. Previously, with deletion pacing enabled,`waitForOngoing` would determine whether or not we should wait for our turn to call `doDeleteObsoleteFiles`, but that function wouldn't wait for the files to physically be deleted.
___
*[`compaction.go` line 3488 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5B_QADtWZmDzCGde-:-NY5B_QADtWZmDzCGde0:b-h6s5dj) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/compaction.go#L3488)):*
> ```Go
> 	d.mu.Unlock()
> 	defer d.mu.Lock()
> 	d.cleanupManager.Wait()
> ```

I didn't follow this. why do we wait for the cleaner but only if `d.mu.disableFileDeletions <= 1`? couldn't another goroutine have _just_ incremented `disableFileDeletions` and dropped the mu to wait for the cleaner. Then when the next goroutine increments, it avoids waiting. Shouldn't they receive the same treatment?
___
*[`compaction.go` line 3569 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5AS7U9Nxfkoy5d3kH:-NY5AS7U9Nxfkoy5d3kI:b6sdmi0) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/compaction.go#L3569)):*
> ```Go
> 
> 	// Release d.mu while doing I/O
> 	// Note the unusual order: Unlock and then Lock.
> ```

this function no longer does I/O directly, right? Should we update this comment?
___
*[`compaction.go` line 3583 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5B7CKDml_M0uDUiJo:-NY5B7CKDml_M0uDUiJp:b-jsb4km) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/compaction.go#L3583)):*
> ```Go
> 	}
> 	_, noRecycle := d.opts.Cleaner.(base.NeedsFileContents)
> 	filesToDelete := make([]obsoleteFile, 0, len(files))
> ```

looking at this existing code, is there any reason for this `files` array? couldn't we populate a single `filesToDelete` slice up above, and then down here perform all the same logic we do within the below for loop?

relatedly, I think this capacity for the `filesToDelete` slice wasn't the intended capacity. `len(files)` is always 4.
___
*[`event_listener_test.go` line 115 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5D-DADGfI08GE7Xar:-NY5D-DADGfI08GE7Xas:bknr5fo) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/event_listener_test.go#L115)):*
> ```Go
> 				defer d.mu.Unlock()
> 				d.enableFileDeletions()
> 				d.TestOnlyWaitForCleaning()
> ```

It looks `d.mu` is held here which is prohibited when calling cleanupManager.Wait(), right?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2641)* status: 10 of 17 files reviewed, 9 unresolved discussions (waiting on @jbowens)

___
*[`cleaner.go` line 50 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5EBoG89pG9YpgxWKk:-NY5aXC50NAOoUImk0vO:b-xjqvyb) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/cleaner.go#L50)):*

is this maybe _too_ big? With 8-byte words, 80KB?
</blockquote></details>

Reduced to 1000
___
*[`cleaner.go` line 103 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5D_oeEPH5dA1MaZH_:-NY5a_j84ZYk2RjsESyU:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/cleaner.go#L103)):*

Looks like this comment is out-of-date.
</blockquote></details>

Done.
___
*[`cleaner.go` line 120 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5FIP_DpX3w51-NAM1:-NY5e27FC5CI00H7rgAU:bnydl69) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/cleaner.go#L120)):*

this makes me nervous even though it shouldn't happen. since enqueues are already mutex-protected, should we use a preallocated slice instead of a channel? if the slice becomes full, we can always suffer the allocation to grow it.
</blockquote></details>

I'm hesitant to stop using the channel because the synchronization is much easier.. We could also do a blocking send, so that we wait for some jobs to finish? We already call EnqueueJob without holding the mutex.
___
*[`compaction.go` line 2636 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5G9fMERDVukxoRKDC:-NY5ana6EzwQmKAnB1Vo:bxe5svy) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/compaction.go#L2636)):*

I think this is a behavior change. Previously, with deletion pacing enabled,`waitForOngoing` would determine whether or not we should wait for our turn to call `doDeleteObsoleteFiles`, but that function wouldn't wait for the files to physically be deleted.
</blockquote></details>

But it would wait for any deletion that was just queued (e.g. from `maybeScheduleObsoleteTableDeletion`) to physically delete files..  If we accept that `maybeScheduleObsoleteTableDeletion` can run at any time, then we should treat both cases the same..

One could argue that the waiting was just an artifact of the poor design (no queue) and we really don't want to wait anywhere.. But I think we want some kind of backpressure - with a low pacing limit, the deletions could become the bottleneck and we don't want them to fall behind indefinitely..
___
*[`compaction.go` line 3488 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5B_QADtWZmDzCGde-:-NY5cB9Z-R3X7Qc2jonU:br21pzf) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/compaction.go#L3488)):*

I didn't follow this. why do we wait for the cleaner but only if `d.mu.disableFileDeletions <= 1`? couldn't another goroutine have _just_ incremented `disableFileDeletions` and dropped the mu to wait for the cleaner. Then when the next goroutine increments, it avoids waiting. Shouldn't they receive the same treatment?
</blockquote></details>

Nice catch, fixed.
___
*[`compaction.go` line 3569 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5AS7U9Nxfkoy5d3kH:-NY5c_Yb5V4-zFE9imwS:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/compaction.go#L3569)):*

this function no longer does I/O directly, right? Should we update this comment?
</blockquote></details>

Done.
___
*[`compaction.go` line 3583 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5B7CKDml_M0uDUiJo:-NY5dUQ_1S1BxGIkQyIO:b9rcdij) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/compaction.go#L3583)):*

looking at this existing code, is there any reason for this `files` array? couldn't we populate a single `filesToDelete` slice up above, and then down here perform all the same logic we do within the below for loop?

relatedly, I think this capacity for the `filesToDelete` slice wasn't the intended capacity. `len(files)` is always 4.
</blockquote></details>

We'd have to do the "conversion" from `fileInfo` to `obsoleteFile` in each place (or at least somehow associate the `fileType` with each entry), and in the loop we'd have to delete recycled logs from the slice. Not sure it ends up better so I'll leave it as is. I fixed the capacity.
___
*[`compaction.go` line 2118 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5gCb_EDCt4BP0qdUw:-NY5gCb_EDCt4BP0qdUx:bmu107n) ([raw file](https://github.com/cockroachdb/pebble/blob/a01ee60d92638e4d62b55f3b87da329e9b0cc4e9/compaction.go#L2118)):*
> ```Go
> 	}
> 
> 	d.deleteObsoleteFiles(jobID, true /* waitForCompletion */)
> ```

Note that I switched this call to wait so that the comment below makes sense..  Not that it made sense with the previous code though (with pacing enabled)..
___
*[`event_listener_test.go` line 115 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5D-DADGfI08GE7Xar:-NY5dw-55H1vTG7234cP:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/event_listener_test.go#L115)):*

It looks `d.mu` is held here which is prohibited when calling cleanupManager.Wait(), right?
</blockquote></details>

Done.

jbowens (approved):
Looks good to me, just have an open question of what to do to avoid backpressuring compactions.

> Reviewed 3 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2641#-:-NYUBLTZ7CosE839vNQH:bwu0kri)* status: 12 of 17 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`cleaner.go` line 120 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5FIP_DpX3w51-NAM1:-NYUApWl2_Wjh7AfadiF:bsmozvy) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/cleaner.go#L120)):*

I'm hesitant to stop using the channel because the synchronization is much easier.. We could also do a blocking send, so that we wait for some jobs to finish? We already call EnqueueJob without holding the mutex.
</blockquote></details>

Yeah, I think a blocking send is a good option.
___
*[`compaction.go` line 2636 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5G9fMERDVukxoRKDC:-NYU5foQ1klcBPgZTVwt:b-ud9myz) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/compaction.go#L2636)):*

But it would wait for any deletion that was just queued (e.g. from `maybeScheduleObsoleteTableDeletion`) to physically delete files..  If we accept that `maybeScheduleObsoleteTableDeletion` can run at any time, then we should treat both cases the same..

One could argue that the waiting was just an artifact of the poor design (no queue) and we really don't want to wait anywhere.. But I think we want some kind of backpressure - with a low pacing limit, the deletions could become the bottleneck and we don't want them to fall behind indefinitely..

</blockquote></details>

If the pacing is limiting the rate of deletions, I think the disk-space utilization heuristic will at least prevent us from exhausting the storage. The deletion pacing exists to protect foreground write latency by avoiding sudden bursts of deletions. I don't think we actually want to backpressure compactions which can degrade foreground read performance.

IMO our current deletion pacing is a little lacking, because it doesn't adapt to disk throughput. We want to smooth the deletions over time but not outright limit them. We could consider adapting the rate according to the deletions requested  over a recent long window, so we allow the rate to increase if we're deleting more files, but we still smooth it. But I think that's for future work.
___
*[`compaction.go` line 3583 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5B7CKDml_M0uDUiJo:-NYUAjZGBnhSUx0qsArn:btn461w) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/compaction.go#L3583)):*

We'd have to do the "conversion" from `fileInfo` to `obsoleteFile` in each place (or at least somehow associate the `fileType` with each entry), and in the loop we'd have to delete recycled logs from the slice. Not sure it ends up better so I'll leave it as is. I fixed the capacity.
</blockquote></details>

Got it

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2641)* status: 12 of 17 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`cleaner.go` line 120 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5FIP_DpX3w51-NAM1:-NYUIFXC0qY5b5Ec6sai:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/cleaner.go#L120)):*

Yeah, I think a blocking send is a good option.
</blockquote></details>

Done.
___
*[`compaction.go` line 2636 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5G9fMERDVukxoRKDC:-NYUIR6t8Fw41eYIxObS:b-ysm2st) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/compaction.go#L2636)):*

If the pacing is limiting the rate of deletions, I think the disk-space utilization heuristic will at least prevent us from exhausting the storage. The deletion pacing exists to protect foreground write latency by avoiding sudden bursts of deletions. I don't think we actually want to backpressure compactions which can degrade foreground read performance.

IMO our current deletion pacing is a little lacking, because it doesn't adapt to disk throughput. We want to smooth the deletions over time but not outright limit them. We could consider adapting the rate according to the deletions requested  over a recent long window, so we allow the rate to increase if we're deleting more files, but we still smooth it. But I think that's for future work.
</blockquote></details>

Ok, I can try to switch to not wait in the compaction and ingest code paths. I'll see how I can make the tests deterministic..

Should an explicit `db.Compact()` call wait for cleanup?

RaduBerinde:
*[`compaction.go` line 2636 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5G9fMERDVukxoRKDC:-NYUKEPg28xTQUapaF_2:b-6e0rd9) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/compaction.go#L2636)):*

Ok, I can try to switch to not wait in the compaction and ingest code paths. I'll see how I can make the tests deterministic..

Should an explicit `db.Compact()` call wait for cleanup?
</blockquote></details>

Hm, I guess maybe not, because we may be compacting a tiny range but there might be background cleanup across the entire key space

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2641)* status: 6 of 19 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`compaction.go` line 2636 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5G9fMERDVukxoRKDC:-NYUazET9NkR4Ms4h3J_:bkrlvps) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/compaction.go#L2636)):*

Hm, I guess maybe not, because we may be compacting a tiny range but there might be background cleanup across the entire key space
</blockquote></details>

Ok, I think I got it working. For one test I had to use an "always wait" hack but for others waiting for cleanup after Compact calls and such was sufficient.

jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2641#-:-NYYb0vp9TpnRMGOC7dQ:bnfp4nl)* status: 6 of 19 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`cleaner.go` line 120 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5FIP_DpX3w51-NAM1:-NYYaVzyBiqzDY1mHFkS:bf4rbqi) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/cleaner.go#L120)):*

Done.
</blockquote></details>

Thanks! #2662 validates that this is the right move, because our fixed deletion pacing rate can cause us to fall behind unboundedly on file deletions.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2641)* status: 6 of 19 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`compaction.go` line 2118 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5gCb_EDCt4BP0qdUw:-NYYz7PP785xGzKjKPNJ:b-rstxk) ([raw file](https://github.com/cockroachdb/pebble/blob/a01ee60d92638e4d62b55f3b87da329e9b0cc4e9/compaction.go#L2118)):*

Note that I switched this call to wait so that the comment below makes sense..  Not that it made sense with the previous code though (with pacing enabled)..
</blockquote></details>

I'm still not very sure about this wait here. Maybe it doesn't matter much anyway? AFAICT we only use `Flush()` for testing in CRDB

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2641)* status: 6 of 19 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`compaction.go` line 2118 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5gCb_EDCt4BP0qdUw:-NYZj6a58iVvVkHP6bYI:b2l840l) ([raw file](https://github.com/cockroachdb/pebble/blob/a01ee60d92638e4d62b55f3b87da329e9b0cc4e9/compaction.go#L2118)):*

I'm still not very sure about this wait here. Maybe it doesn't matter much anyway? AFAICT we only use `Flush()` for testing in CRDB
</blockquote></details>

Thanks for calling this out again. This function will get called for non-explicit flushes when the memtable becomes full too. Waiting will prevent new flushes from scheduling. I don't think we do want to wait here.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2641)* status: 6 of 19 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`cleaner.go` line 120 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2641#-NY5FIP_DpX3w51-NAM1:-NYZl3xd3m1PTovwQ9uF:b-laknmk) ([raw file](https://github.com/cockroachdb/pebble/blob/07877ee6bcc74d0455e526deef63e88acd928c41/cleaner.go#L120)):*

Thanks! #2662 validates that this is the right move, because our fixed deletion pacing rate can cause us to fall behind unboundedly on file deletions. 
</blockquote></details>

Makes sense.. The only remaining place where we wait is during `Open()`.. I don't think we want to wait there either, do we?

RaduBerinde:
I removed the wait flag altogether; we now never wait for cleanup except for some tests.

jbowens (approved):
:lgtm:

> Reviewed 6 of 11 files at r4, 7 of 7 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2641#-:-NYs_s6O4USDYpK4bImK:bnfp4nl)* status: 19 of 22 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #2643, opened 2023-06-16
db: don't require holding d.mu to read format major version

jbowens:
Allow readers of the database's format major version to read the format major version through an atomic load, instead of acquiring the global d.mu. This is particularly important for batch application, which must read the format major version to determine if the batch holds any keys that cannot be applied. Mutex contention during batch application in DB.FormatMajorVersion has been observed in experiments.

Screenshot 2023-06-16 at 11 16 59 AM

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -187,7 +187,7 @@ func (d *DB) Checkpoint(
 	// file number.
 	memQueue := d.mu.mem.queue
 	current := d.mu.versions.currentVersion()
-	formatVers := d.mu.formatVers.vers
+	formatVers := d.FormatMajorVersion()
 	manifestFileNum := d.mu.versions.manifestFileNum
 	manifestSize := d.mu.versions.manifest.Size()
 	optionsFileNum := d.optionsFileNum
--- a/compaction.go
+++ b/compaction.go
@@ -2724,7 +2724,7 @@ func (d *DB) runCompaction(
 	}()
 
 	snapshots := d.mu.snapshots.toSlice()
-	formatVers := d.mu.formatVers.vers
+	formatVers := d.FormatMajorVersion()
 
 	// Release the d.mu lock while doing I/O.
 	// Note the unusual order: Unlock and then Lock.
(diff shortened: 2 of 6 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2643)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @jbowens and @nvanbenschoten)

___
*[`db.go` line 348 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2643#-NY4HiZz60FyX5s7_XPJ:-NY4HiZz60FyX5s7_XPK:bf54zb2) ([raw file](https://github.com/cockroachdb/pebble/blob/beb0b8761e227d655d085d78229d97fe8d0c0fd4/db.go#L348)):*
> ```Go
> 			// value through finalizeFormatVersUpgrade which requires d.mu is
> 			// held.
> 			vers atomic.Value
> ```

We should make this `atomic.Uint64` which is more efficient

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2643#-:-NY51PQ5-yZIJeLIMT7a:bnfp4nl)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @jbowens and @nvanbenschoten)

jbowens (commented):
tftr!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2643#-:-NY5bB-vAKuP9QLDOddw:b-7ixx6c)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @nvanbenschoten and @RaduBerinde)

___
*[`db.go` line 348 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2643#-NY4HiZz60FyX5s7_XPJ:-NY5b9w39N8-gFomos_Z:b7anmnw) ([raw file](https://github.com/cockroachdb/pebble/blob/beb0b8761e227d655d085d78229d97fe8d0c0fd4/db.go#L348)):*

We should make this `atomic.Uint64` which is more efficient
</blockquote></details>

good call

PR #2427, opened 2023-04-01
*: use atomic types

RaduBerinde:
This set of commits changes most usages of atomics to types like `atomic.Uint64`, `atomic.Bool` etc. Note that `atomic.Uint64 / atomic.Int64` are guaranteed to be 64-bit aligned on all platforms (unlike `int64/uint64`).

There are some usages remaining, mostly because of structs that get copied around.

PR #2645, opened 2023-06-16
cache: use atomic.Int32 wrappers, fix potential bug

RaduBerinde:

This change switches to atomic.Int32 wrapper and uses `Store` on init.

### DIFF
--- a/internal/cache/refcnt_normal.go
+++ b/internal/cache/refcnt_normal.go
@@ -15,26 +15,28 @@ import (
 // refcnt provides an atomic reference count. This version is used when the
 // "tracing" build tag is not enabled. See refcnt_tracing.go for the "tracing"
 // enabled version.
-type refcnt int32
+type refcnt struct {
+	val atomic.Int32
+}
 
 // initialize the reference count to the specified value.
 func (v *refcnt) init(val int32) {
-	*v = refcnt(val)
+	v.val.Store(val)
 }
 
 func (v *refcnt) refs() int32 {
-	return atomic.LoadInt32((*int32)(v))
+	return v.val.Load()
 }
 
 func (v *refcnt) acquire() {
-	switch v := atomic.AddInt32((*int32)(v), 1); {
+	switch v := v.val.Add(1); {
 	case v <= 1:
 		panic(fmt.Sprintf("pebble: inconsistent reference count: %d", v))
 	}
 }
 
 func (v *refcnt) release() bool {
-	switch v := atomic.AddInt32((*int32)(v), -1); {
+	switch v := v.val.Add(-1); {
 	case v < 0:
 		panic(fmt.Sprintf("pebble: inconsistent reference count: %d", v))
 	case v == 0:
--- a/internal/cache/refcnt_tracing.go
+++ b/internal/cache/refcnt_tracing.go
@@ -19,22 +19,22 @@ import (
 // debugging logic errors in manipulating the reference count. This version is
 // used when the "tracing" build tag is enabled.
 type refcnt struct {
-	val int32
+	val atomic.Int32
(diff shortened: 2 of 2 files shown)


bananabrick (commented):
Is the theory that if we write to the refcount during init, without the write being atomic, then we could have a write to the cache, but not memory, and then the future read could potentially not see the write because it hasn't been flushed to memory from the cache?

I think we follow this pattern of initializing these atomic variables without an atomic store all over our code: https://github.com/cockroachdb/pebble/blob/master/db.go#L2106 is another example.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2645#-:-NY4TFMT2kuQmlG9zeKN:bo9jxlu)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jbowens)

RaduBerinde:
Yeah, imagine one CPU calls `newEntry` which does a non-atomic assignment then another CPU calls `free` soon after.. CPUs have store buffers, STORES don't immediately go to memory unless there's some kind of barrier.

In most other usages, this kind of scenario is probably too rare or impossible, not sure. For example, if we always release a mutex before the first atomic operation, we would be ok.

In any case, we should just be using the `atomic.` wrappers everywhere. I changed most of them in #2427 but there are some left.

bananabrick (commented):
I wonder why the following pattern is okay then.

Thread 1 acquires mutex, writes to some variable x, and releases mutex. Thread 2 acquires mutex, reads variable x, and releases mutex.

When thread 1 releases the mutex, is it guaranteed that the write to variable x will be flushed to memory from the cache?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2645#-:-NY4X4q2ArzMNDs0lY1u:b-7aanrl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jbowens)

RaduBerinde:
> Thread 1 acquires mutex, writes to some variable x, and releases mutex. Thread 2 acquires mutex, reads variable x, and releases mutex.

My understanding is that on a CPU loads and stores are not reordered across "locked" (atomic) operations. So atomic operations effectively act as memory barriers. Releasing a mutex involves an atomic operation. When a CPU sees the result of an atomic operation, it will see everything that was stored before by that CPU. This is the case at least on x86/amd64, more details here: https://www.intel.com/content/www/us/en/architecture-and-technology/64-ia-32-architectures-software-developer-vol-3a-part-1-manual.html

RaduBerinde:
> When thread 1 releases the mutex, is it guaranteed that the write to variable x will be flushed to memory from the cache?

Also note that the issue is not really flushing from L2/L3 to memory (L2 caches run coherence protocols), but when the stores make it out from a CPU's store buffer to L2 in the first place.

bananabrick (approved):
Never thought about mutexes and reordering before. Pretty interesting.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2645#-:-NY4dq3i4TwtZeCtKW5U:bnwwyn9)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jbowens)

jbowens (approved):
:lgtm:

Seems very plausible!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2645#-:-NY56Mg51H3Mn-abKNAk:b36y7q6)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

RaduBerinde:
I will backport to to 22.2 and 23.1 if it fixes https://github.com/cockroachdb/cockroach/issues/84971

PR #2648, opened 2023-06-17
*: switch to using atomic wrappers

RaduBerinde:
This change switches atomics to use wrappers. In some cases, this forced us to fix incorrect copying of atomic values, missing Loads, etc.

The only remaining use is for the pointer in the commit queue. That will be done in a separate PR for more thorough review (`atomic.Pointer` involves generics).


RaduBerinde:
*[`error_test.go` line 38 at r16](https://reviewable.io/reviews/cockroachdb/pebble/2648#-NY9R46j1so3d9h9rZuL:-NY9R46j1so3d9h9rZuM:b-bl6j9q) ([raw file](https://github.com/cockroachdb/pebble/blob/7b5313aa6fc255ee4101e65fb474c4bc1199e5f0/error_test.go#L38)):*
> ```Go
> }
> 
> func (fs *corruptFS) maybeCorrupt(n int32, p []byte) {
> ```

This test was completely broken - the non-pointer receiver means we never modified `bytesRead`, plus the `ReadAt` implementation (which is the one we actually use for ssts) was not correct.

jbowens (approved):
:lgtm:

> Reviewed 5 of 5 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 4 of 4 files at r6, 2 of 2 files at r7, 1 of 1 files at r8, 1 of 1 files at r9, 1 of 1 files at r10, 2 of 2 files at r11, 1 of 1 files at r12, 2 of 2 files at r13, 1 of 1 files at r14, 5 of 5 files at r15, 3 of 3 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2648#-:-NYOLJKg5qgLT5bYiRqT:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`internal/manifest/btree.go` line 166 at r9](https://reviewable.io/reviews/cockroachdb/pebble/2648#-NYOK_zo8wH6asQKowWA:-NYOK_zo8wH6asQKowWB:brn4f7z) ([raw file](https://github.com/cockroachdb/pebble/blob/1a7fe39c04b4c2a4e74f1da4ab0ad3f07729c7e3/internal/manifest/btree.go#L166)):*
> ```Go
> 
> 		// Whenever a node will be mutated, reset its annotations to be marked
> 		// as uncached. This ensures any future calls to (*node).annotation
> ```

looks like an accidental line deletion

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2648#-:-NYOlYuABY_i9MylwH1H:b-814g8k)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`internal/manifest/btree.go` line 166 at r9](https://reviewable.io/reviews/cockroachdb/pebble/2648#-NYOK_zo8wH6asQKowWA:-NYOlX9Y2bfXiH4J-8KR:bralddm) ([raw file](https://github.com/cockroachdb/pebble/blob/1a7fe39c04b4c2a4e74f1da4ab0ad3f07729c7e3/internal/manifest/btree.go#L166)):*

looks like an accidental line deletion
</blockquote></details>

Nice catch, thanks!!

PR #2650, opened 2023-06-19
sstable: shrink buffer in readFooter before calling Readable

itsbilal:
The objstorage.Readable interface specifically disallows partial reads, as an EOF error could be returned if a buffer larger than the size of the remaining part of the file is passed in. This change shrinks the buffer before calling ReadAt to avoid an EOF error from being erroneously returned.

### DIFF
--- a/sstable/table.go
+++ b/sstable/table.go
@@ -337,6 +337,7 @@ func readFooter(f objstorage.Readable) (footer, error) {
 		off = 0
 		buf = buf[:size]
 	}
+	buf = buf[:size-off]
 	if err := f.ReadAt(context.TODO(), buf, off); err != nil {
 		return footer, errors.Wrap(err, "pebble/table: invalid table (could not read footer)")
 	}


RaduBerinde:
*[`sstable/table.go` line 340 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2650#-NYIzXc4EHHPTNw_PdaK:-NYIzXc4EHHPTNw_PdaL:b70rdgo) ([raw file](https://github.com/cockroachdb/pebble/blob/ce5f0686a658fa4fd28c1393db2915f9b0829db8/sstable/table.go#L340)):*
> ```Go
> 		buf = buf[:size]
> 	}
> 	buf = buf[:size-off]
> ```

I initially thought this is needed too.. but if size >= maxFooterLen then `buf` is already the right size (`maxFooterLen=size-off`), and if not, it gets capped in the block above..

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2650)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`sstable/table.go` line 340 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2650#-NYIzXc4EHHPTNw_PdaK:-NYJ2lcZ12bil1xwMo_T:b-lvg4su) ([raw file](https://github.com/cockroachdb/pebble/blob/ce5f0686a658fa4fd28c1393db2915f9b0829db8/sstable/table.go#L340)):*

I initially thought this is needed too.. but if size >= maxFooterLen then `buf` is already the right size (`maxFooterLen=size-off`), and if not, it gets capped in the block above..
</blockquote></details>

Maybe we should rewrite to make things more clear (like below), but I don't see any bug here. The only way I can see this hitting EOF is if `Size()` returns the wrong size..

```diff
diff --git a/sstable/table.go b/sstable/table.go
index 040efafd..6bbbcda5 100644
--- a/sstable/table.go
+++ b/sstable/table.go
@@ -327,17 +327,17 @@ type footer struct {
 func readFooter(f objstorage.Readable) (footer, error) {
        var footer footer
        size := f.Size()
-       if size < minFooterLen {
+       var buf []byte
+       switch {
+       case size < minFooterLen:
                return footer, base.CorruptionErrorf("pebble/table: invalid table (file size is too small)")
+       case size < maxFooterLen:
+               buf = make([]byte, size)
+       default:
+               buf = make([]byte, maxFooterLen)
        }
 
-       buf := make([]byte, maxFooterLen)
-       off := size - maxFooterLen
-       if off < 0 {
-               off = 0
-               buf = buf[:size]
-       }
-       if err := f.ReadAt(context.TODO(), buf, off); err != nil {
+       if err := f.ReadAt(context.TODO(), buf, size-int64(len(buf))); err != nil {
                return footer, errors.Wrap(err, "pebble/table: invalid table (could not read footer)")
        }
 
@@ -347,9 +347,9 @@ func readFooter(f objstorage.Readable) (footer, error) {
                        return footer, base.CorruptionErrorf(
                                "pebble/table: invalid table (footer too short): %d", errors.Safe(len(buf)))
                }
-               footer.footerBH.Offset = uint64(off+int64(len(buf))) - levelDBFooterLen
+               footer.footerBH.Offset = uint64(size) - levelDBFooterLen
                buf = buf[len(buf)-levelDBFooterLen:]
-               footer.footerBH.Length = uint64(len(buf))
+               footer.footerBH.Length = levelDBFooterLen
                footer.format = TableFormatLevelDB
                footer.checksum = ChecksumTypeCRC32c
 
@@ -359,9 +359,9 @@ func readFooter(f objstorage.Readable) (footer, error) {
                if len(buf) < rocksDBFooterLen {
                        return footer, base.CorruptionErrorf("pebble/table: invalid table (footer too short): %d", errors.Safe(len(buf)))
                }
-               footer.footerBH.Offset = uint64(off+int64(len(buf))) - rocksDBFooterLen
+               footer.footerBH.Offset = uint64(size) - rocksDBFooterLen
                buf = buf[len(buf)-rocksDBFooterLen:]
-               footer.footerBH.Length = uint64(len(buf))
+               footer.footerBH.Length = rocksDBFooterLen
                version := binary.LittleEndian.Uint32(buf[rocksDBVersionOffset:rocksDBMagicOffset])
 
                format, err := ParseTableFormat(magic, version)
```

itsbilal:
@RaduBerinde That's a good suggestion. But yeah you're right, on closer inspection there's no bug here. The real fix necessary is https://github.com/cockroachdb/cockroach/pull/105118 . I'll close this PR for now but I'll try to package your suggestion in a PR shortly just to reduce investigation time in bugs like these in the future.

PR #2651, opened 2023-06-19
ingest: Call Extend{Point,Range}KeyBounds in excise

itsbilal:
Previously when we called `excise()`, we created new FileMetadata objects without calling ExtendPointKeyBounds and ExtendRangeKeyBounds. During testing we quickly realized that we would result in some instances of us not having the `boundType` set correctly, resulting in panics when we tried to install a file that had a modified boundType.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -1368,20 +1368,9 @@ func (d *DB) excise(
 			FileBacking: m.FileBacking,
 			FileNum:     d.mu.versions.getNextFileNum(),
 		}
-		leftFile.Smallest = m.Smallest
-		leftFile.SmallestRangeKey = m.SmallestRangeKey
-		leftFile.SmallestPointKey = m.SmallestPointKey
-		leftFile.HasPointKeys = m.HasPointKeys
-		leftFile.HasRangeKeys = m.HasRangeKeys
-		if m.HasPointKeys && exciseSpan.Contains(d.cmp, m.SmallestPointKey) {
-			// This file will not contain any point keys, but will contain range keys.
-			leftFile.HasPointKeys = false
-			leftFile.Smallest = m.SmallestRangeKey
-		} else if m.HasRangeKeys && exciseSpan.Contains(d.cmp, m.SmallestRangeKey) {
-			leftFile.HasRangeKeys = false
-			leftFile.Smallest = m.SmallestPointKey
-		}
-		if leftFile.HasPointKeys {
+		if m.HasPointKeys && !exciseSpan.Contains(d.cmp, m.SmallestPointKey) {
+			// This file will contain point keys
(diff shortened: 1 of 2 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2651#-:-NYOMpfh9VnVLWa5kn7s:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal and @sumeerbhola)

___
*[`ingest.go` line 1420 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2651#-NYOLsI-BQ8KMWlfZSR8:-NYOLsI-BQ8KMWlfZSR9:b4hnbta) ([raw file](https://github.com/cockroachdb/pebble/blob/12bd79cf12073a1d944c346e6340ef87e3682f79/ingest.go#L1420)):*
> ```Go
> 			// leftFile.{Smallest,Largest}PointKey with themselves, which can be
> 			// avoided if we exported ExtendOverallKeyBounds or so.
> 			leftFile.ExtendPointKeyBounds(d.cmp, leftFile.SmallestPointKey, leftFile.LargestPointKey)
> ```

Alternatively, I think we could restructure the code to perform all bound-setting through `ExtendPointKeyBounds`.

eg, up above
```
if m.HasPointKeys && !exciseSpan.Contains(d.cmp, m.SmallestPointKey) {
     // Inspect the SST, calling `ExtendPointKeyBounds` once for range dels and once for point keys.
}
```
I suspect that would read a little easier.

itsbilal (commented):
TFTR!

And yes, added a test that'd have caught this.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2651#-:-NYUWXKB33eMpKDblxNE:bfsquvl)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`ingest.go` line 1420 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2651#-NYOLsI-BQ8KMWlfZSR8:-NYUWZrqEJvZ4XAOEw38:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/12bd79cf12073a1d944c346e6340ef87e3682f79/ingest.go#L1420)):*

Alternatively, I think we could restructure the code to perform all bound-setting through `ExtendPointKeyBounds`.

eg, up above
```
if m.HasPointKeys && !exciseSpan.Contains(d.cmp, m.SmallestPointKey) {
     // Inspect the SST, calling `ExtendPointKeyBounds` once for range dels and once for point keys.
}
```
I suspect that would read a little easier.
</blockquote></details>

Done.

PR #2660, opened 2023-06-21
db: Zero out bounds for key type not present in ScanInternal

itsbilal:
Previously, when we were exporting shared files through truncateSharedFile, and we found no point keys in the relevant span but found range keys (or vice versa), we would leave some of the bounds of the not-found key type set. This could trip up logic elsewhere that relied on null key bounds for keys not present.


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2660#-:-NYZ_Bsd6X_xQHESZYF-:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @sumeerbhola)

itsbilal:
TFTR!

PR #2661, opened 2023-06-21
db: Refactor compaction cancellation with concurrent excises

itsbilal:
This change reworks the logic to cancel concurrent compactions by moving to a per-compaction cancelled atomic boolean, instead of relying on a boolean in the input FileMetadatas. This is easier for the compaction to check during its execution, reducing the amount of wasted work. It's also more correct as it allows the excising goroutine to error out an entire compaction even if none of its inputs overlapped with the excised span, as there's no guarantee that there won't be an output that overlaps with the excise span even if no inputs overlapped with it.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -13,6 +13,7 @@ import (
 	"runtime/pprof"
 	"sort"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/cockroachdb/errors"
@@ -30,6 +31,7 @@ import (
 
 var errEmptyTable = errors.New("pebble: empty table")
 var errFlushInvariant = errors.New("pebble: flush next log number is unset")
+var errCancelledCompaction = errors.New("pebble: compaction cancelled by a concurrent operation, will retry compaction")
 
 var compactLabels = pprof.Labels("pebble", "compact")
 var flushLabels = pprof.Labels("pebble", "flush")
@@ -546,6 +548,11 @@ func rangeKeyCompactionTransform(
 // compaction is a table compaction from one level to the next, starting from a
 // given version.
 type compaction struct {
+	// cancel is a bool that can be used by other goroutines to signal a compaction
+	// to cancel, such as if a conflicting excise operation raced it to manifest
+	// application. Only holders of the manifest lock will write to this atomic.
+	cancel atomic.Bool
+
 	kind      compactionKind
 	cmp       Compare
 	equal     Equal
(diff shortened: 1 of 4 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2661)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`-- commits` line 12 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2661#-NYZ_mU46R32HweJJpcM):*
This concern applies to non-excising ingestions that need to virtualize sstables as well, right? Although I'm not sure in that case we necessarily want to cancel the compaction and waste its already-expended IO.

itsbilal:
*[`-- commits` line 12 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2661#-NYZ_mU46R32HweJJpcM):*

This concern applies to non-excising ingestions that need to virtualize sstables as well, right? Although I'm not sure in that case we necessarily want to cancel the compaction and waste its already-expended IO.
</blockquote></details>

That's correct. That PR is not merged yet so I phrased this just in the context of excise, but the exact same concern does apply there too.

jbowens (approved):
:lgtm:

modulo skipped `versionEditApplied` compactions

> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2661#-:-NYt8DMcCaWm340zv91X:b19iivw)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`compaction.go` line 2860 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2661#-NYt5TjI6iVK37fswmZs:-NYt5TjI6iVK37fswmZt:bo37dys) ([raw file](https://github.com/cockroachdb/pebble/blob/bebfa0ef367f657858233082a105b3776b42ebe5/compaction.go#L2860)):*
> ```Go
> 		if c.cancel.Load() {
> 			return errCancelledCompaction
> 		}
> ```

Should we hoist this up to the very beginning of `newOutput`, eg before we've acquired a file num?
___
*[`ingest.go` line 1735 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2661#-NYt7_M34Qf8pua_1qNU:-NYt7_M34Qf8pua_1qNV:b-pxhfa1) ([raw file](https://github.com/cockroachdb/pebble/blob/bebfa0ef367f657858233082a105b3776b42ebe5/ingest.go#L1735)):*
> ```Go
> 	}
> 	for c := range d.mu.compact.inProgress {
> 		// Check if this compaction overlaps with the excise span. Note that just
> ```

We can skip compactions for which `c.versionEditApplied` is true, because these files are already committed to the LSM and will simply be excised if necessary.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2661#-:-NYtPWDA5U2IF_1m_vOm:b-814g8k)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`compaction.go` line 2860 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2661#-NYt5TjI6iVK37fswmZs:-NYtQ8Fx5mrPbjtK5fSi:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/bebfa0ef367f657858233082a105b3776b42ebe5/compaction.go#L2860)):*

Should we hoist this up to the very beginning of `newOutput`, eg before we've acquired a file num?
</blockquote></details>

Done.
___
*[`ingest.go` line 1735 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2661#-NYt7_M34Qf8pua_1qNU:-NYtPXa4DPxRTTJA7G6o:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/bebfa0ef367f657858233082a105b3776b42ebe5/ingest.go#L1735)):*

We can skip compactions for which `c.versionEditApplied` is true, because these files are already committed to the LSM and will simply be excised if necessary.
</blockquote></details>

Done.

Issue #1810, opened 2022-07-14
perf: snapshot predicates


jbowens:
Currently, a snapshot always applies universally across all keys in the database. In CockroachDB, snapshots are used to preserve state within the context of a single range. An LSM snapshot constructed to read range _r<sub>1</sub>_ still prevents removal of obsolete keys in range _r<sub>2</sub>_.

We could extend `NewSnapshot` to allow supplying a predicate `p(k)→bool` that configures the snapshot to only snapshot keys for which `p` returns true. During compaction and flushes, when a snapshot appears to produce a new snapshot stripe within the same key, `p(k)` is consulted and a new stripe is produced only if `p(k)→true`. This would allow overwritten keys in a non-snapshotted CockroachDB range to be dropped while preserving overwritten keys in a snapshotted CockroachDB range.

There's still the question of what to do when an iterator constructed through `Snapshot.NewIter` reads a key for which the predicate `p(k)→false`:
1. Do nothing—the result of iterating over keys that do not match `p(k)` is undefined. The caller must be careful to never use iteration results outside the predicate. If the predicate is defined over swaths of the key space, this may be achieved through settings iterator bounds.
2. Filter—during iteration, before returning test `p(k)`, skipping the key if `p(k)→false`.
3. Read at higher sequence number—when the iterator is constructed from the snapshot, record two sequence numbers: the current visible sequence number on the database, and the snapshot's sequence number. User keys for which `p(k)→true` are filtered at the snapshot's sequence number. User keys for which `p(k)→false` are filtered at the database's visible sequence number.

I expect limiting the scope of active snapshots would reduce write amplification, in particular during periods of heavy rebalancing where there are open LSM snapshots and replicas are being simultaneously removed. Replica removal lays down range deletions, but those range deletions are unable to drop the replica's data. Compaction of these range deletions is still prioritized, because wide range deletions force ingested sstables into higher levels. The result is we suffer unnecessary write amplification moving the removed replica's data and the range tombstone into L6.

If we are to tackle this, I think we might want to expose a very limited interface, at least from the CockroachDB `pkg/storage` package that meets our specific snapshot usages. This can help avoid the possibility of reading unshapshotted keys while under the impression of reading through a consistent snapshot.

The amount of write amplification saved is still unknown. Adding metrics for the size of obsolete keys preserved during compactions (#1204) would help us prioritize.

sumeerbhola:
[...]

PR #1189, opened 2021-07-09
db: use uncompensated scores to smoothe compaction picking scores

jbowens:
When adjusting scores by smoothing scores, use a parallel uncompensated
score as the divisor for the smoothing adjustment. This helps avoid
range deletions in lower levels from starving compactions in higher
levels through excessive prioritization of lower-level disk space
reclamation.

Informs #1181.

PR #2663, opened 2023-06-21
db: use uncompensated scores to smoothe compaction picking scores

bananabrick:
When adjusting scores by smoothing scores, use a parallel uncompensated score as the divisor for the smoothing adjustment. This helps avoid range deletions in lower levels from starving compactions in higher levels through excessive prioritization of lower-level disk space reclamation.


bananabrick:
Previous approved pr: https://github.com/cockroachdb/pebble/pull/1189. I pushed changes to that branch too, but reviewable wasn't letting me leave comments.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2663)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`compaction_picker.go` line 1046 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2663#-NYVMiW1-n2yhoSyIin1:-NYVMiW1-n2yhoSyIin2:b-2hvilo) ([raw file](https://github.com/cockroachdb/pebble/blob/84c66c1775ccbbe7e2103023eb8ee9cf5e5e7aa0/compaction_picker.go#L1046)):*
> ```Go
> 			const minScore = 0.01
> 			if scores[level].score >= minScore {
> 				scores[prevLevel].score /= scores[level].rawScore
> ```

I think using score here doesn't make sense, because we aren't reducing the write amp by not compacting into a level which has a high score due to level size compensation. This was discussed in the previous approved pr.

bananabrick (commented):
The only change I made was fixed the tests which needed to be rebased.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2663#-:-NYVN5D2ASJ-wf4FBncO:b3qghyx)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

bananabrick:
Read amp is looking a bit better(100 vs 200): https://grafana.testeng.crdb.io/d/ngnKs_j7x/admission-control?orgId=1&from=1687393951622&to=1687398461047&var-cluster=arjunnair-1687393815-01-n10cpu16&var-instances=All

https://grafana.testeng.crdb.io/d/ngnKs_j7x/admission-control?orgId=1&var-cluster=arjunnair-1687207031-01-n10cpu16&var-instances=All&from=1687205345636&to=1687212953183

But the KV store requests graph seems choppy with this change. Will look into it.

jbowens:
Nice, makes sense.

There are still a lot of snapshot applications forcing flushes.
Screenshot 2023-06-22 at 11 19 36 AM
There are also still compactions wasting bandwidth writing data that is pinned by snapshots:
Screenshot 2023-06-22 at 12 47 28 PM

It looks like there's a spike of elastic admittance right at the same time as the snapshot applications. Are these range merges that are being allowed in?
Screenshot 2023-06-22 at 1 03 37 PM

I think the aggressive range merging we see here is a large part of why it's so disruptive. It looks like we're simultaneously dropping `RANGEDEL`s _and_ replacing most of the LSM by applying broad range snapshots of merged ranges.

So there's still a few sources of badness outside of compaction-picker scoring that I think we do want to address:
1. The write bandwidth wasted by open LSM snapshots: This would be fixed by #1810 or maybe reducing the length of time these snapshots are held.
2. The flushes forced by snapshot reception increase read-amp excessively: cockroachdb/cockroach#99273 is probably contributing. Or if these snapshots are for merged ranges that cover old replicas already in the store, that could contribute as well. If we fixed cockroachdb/cockroach#99273 _and_ use Ingest+Excise, there would be no increase in read-amp because the the snapshot sstables could be ingested directly into L6.
3. We seem to be aggressively merging ranges, despite the increasing read amplification. Should admission control be preventing these merge operations from overwhelming the storage engine?

In terms of compaction scoring, maybe we should consider putting a ceiling on the score that can result from compensated sizes. For example, the score of a level could be: `Max(MIN(3, Score(CompensatedSize)), Score(Size))`. During this test, L0 had large scores:

Screenshot 2023-06-22 at 1 14 09 PM

They just paled in comparison to the scores lower in the LSM due to the very large amount of garbage.

Screenshot 2023-06-22 at 1 14 50 PM

jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2663#-:-NYZKwat8teJgHJ94eBM:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal)

bananabrick:
Ran it again: https://grafana.testeng.crdb.io/d/ngnKs_j7x/admission-control?orgId=1&var-cluster=arjunnair-1687986297-01-n10cpu16&var-instances=All&from=now-1h&to=now. I don't see the any throughput drops.

I am seeing some high write amps, so looking into that. Maybe there were temporary spikes in disk throughput which led to fsync latency spikes.

bananabrick:
Not sure what's going on now. Maybe I'm doing something wrong, but here's a run on master without this pr with a bunch of liveness failures: https://grafana.testeng.crdb.io/d/ngnKs_j7x/admission-control?orgId=1&var-cluster=arjunnair-1687996100-01-n10cpu16&var-instances=All&from=now-1h&to=now

I think I'm just gonna do another rounds of runs tomorrow, to see what's going on. I think the clearrange roachtest also failed during its last nightly run.

bananabrick:
I wonder if these are just GCE disk stalls, the fsync latencies are 10s+. Didn't we have an issue with gce disk stalls? I did a cursory search in Cockroach and couldn't find that issue.

bananabrick:
Alright, I can't repro the liveness failure anymore. I've made some minor changes. PTAL, @jbowens

jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2663#-:-NZBgP94-PZT04jINZ61:bnfp4nl)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

bananabrick:
Thanks

PR #2665, opened 2023-06-22
crl-release-22.2: db: don't require holding d.mu to read format major version

jbowens:
22.2 backport of #2643.

----

Allow readers of the database's format major version to read the format major version through an atomic load, instead of acquiring the global d.mu. This is particularly important for batch application, which must read the format major version to determine if the batch holds keys any keys that cannot be applied. Mutex contention during batch application in DB.FormatMajorVersion has been observed in experiments.

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -142,7 +142,7 @@ func (d *DB) Checkpoint(
 	// file number.
 	memQueue := d.mu.mem.queue
 	current := d.mu.versions.currentVersion()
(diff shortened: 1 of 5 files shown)


jbowens:
Need to convert to use atomic.StoreUint64/atomic.LoadUint64.

RaduBerinde (approved):

jbowens:
TFTR!

PR #2670, opened 2023-06-23
ingest: Set {Smallest,Largest}SeqNum in Excise

itsbilal:
Previously, we were leaving the Smallest,Largest SeqNum fields unset in the FileMetadata after we did an excise. This broke snapshot handling even if we weren't reading an excised key range through a snapshot; this is because we could excise a file that has an open snapshot reading some keys but not the exciseed keys, and the new file would contain zero seqnums in its FileMetadata, resulting in incorrect results.

This change fixes this bug, and improves a lot of datadriven testing across the project to better test for seqnum fields in the FileMetadata.


itsbilal:
TFTR!

Issue #2593, opened 2023-06-05
ingest: Add invariant-gated checks for accurate use of virtual sstable bounds


itsbilal:
When #2538 merges, we will be synthesizing file bounds for virtual sstables that got keys masked through an Excise. And when #2582 merges, we'll be doing the same for sstables split by virtualizations We should add invariants.Enabled-gated checks to ensure that iterators created on top of virtual sstables do not return excised/masked keys, and that every key returned is within {Smallest,Largest}{PointKey,RangeKey}. We should also confirm if the bounds were calculated correctly to begin with.

After #2616 is complete, this issue only refers to verifying the tight calculation of sstable bounds, i.e. that there is a user key at each of the Smallest, Largest bounds. This check can happen in `fileManifest.ValidateVirtual`, however it would require creating iterators and is hence best to be invariants-gated as well.

RahulAggarwal1016:
Fixed with https://github.com/cockroachdb/pebble/pull/2698

PR #2672, opened 2023-06-23
*: allow virtual sstables to truncate range keys/dels

itsbilal:
Previously, our virtual sstable read code panicked if we saw a rangedel or rangekey get truncated at a virtual sstable file bound. While this is rare, this is an allowable use of virtual sstables in Cockroach and one that will (on rare occasion) happen organically, eg. when there are two range dels or range keys that abut each other while in different KV ranges (thereby getting defragmented by Pebble, which isn't aware of CockroachDB ranges).

Also add a guard against 0 filesizes, which can cause panics in compaction code.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1410,7 +1410,7 @@ func (c *compaction) newInputIter(
 			// any range tombstones completely outside file bounds.
 			rangeDelIter = keyspan.Truncate(
 				c.cmp, rangeDelIter, lowerBound.UserKey, upperBound.UserKey,
-				&f.Smallest, &f.Largest, false, /* panicOnPartialOverlap */
+				&f.Smallest, &f.Largest, false, /* panicOnUpperTruncate */
 			)
 		}
 		if rangeDelIter == nil {
--- a/ingest.go
+++ b/ingest.go
@@ -84,6 +84,10 @@ func ingestValidateKey(opts *Options, key *InternalKey) error {
 func ingestLoad1Shared(
 	opts *Options, sm SharedSSTMeta, fileNum base.DiskFileNum,
 ) (*fileMetadata, error) {
+	if sm.Size == 0 {
+		// Disallow 0 file sizes
+		return nil, errors.New("pebble: cannot ingest shared file with size 0")
+	}
 	// Don't load table stats. Doing a round trip to shared storage, one SST
 	// at a time is not worth it as it slows down ingestion.
 	meta := &fileMetadata{}
@@ -103,7 +107,9 @@ func ingestLoad1Shared(
 		meta.SmallestRangeKey = sm.SmallestRangeKey
 		meta.LargestRangeKey = sm.LargestRangeKey
 		meta.SmallestRangeKey.SetSeqNum(seqNum)
-		meta.LargestRangeKey.SetSeqNum(seqNum)
+		if !meta.LargestRangeKey.IsExclusiveSentinel() {
+			meta.LargestRangeKey.SetSeqNum(seqNum)
+		}
 		meta.SmallestSeqNum = seqNum
 		meta.LargestSeqNum = seqNum
 		// Initialize meta.{Smallest,Largest} and others by calling this.
@@ -114,7 +120,9 @@ func ingestLoad1Shared(
 		meta.SmallestPointKey = sm.SmallestPointKey
 		meta.LargestPointKey = sm.LargestPointKey
 		meta.SmallestPointKey.SetSeqNum(seqNum)
-		meta.LargestPointKey.SetSeqNum(seqNum)
+		if !meta.LargestPointKey.IsExclusiveSentinel() {
+			meta.LargestPointKey.SetSeqNum(seqNum)
+		}
 		meta.SmallestSeqNum = seqNum
 		meta.LargestSeqNum = seqNum
 		// Initialize meta.{Smallest,Largest} and others by calling this.
@@ -356,7 +364,7 @@ func ingestSortAndVerify(cmp Compare, lr ingestLoadResult, exciseSpan KeyRange) 
 	for i := range lr.sharedMeta {
 		f := lr.sharedMeta[i]
 		if !exciseSpan.Contains(cmp, f.Smallest) || !exciseSpan.Contains(cmp, f.Largest) {
-			return errors.AssertionFailedf("pebble: shared file outside of excise span")
+			return errors.AssertionFailedf("pebble: shared file outside of excise span, span [%s-%s), file = %s", exciseSpan.Start, exciseSpan.End, f.String())
 		}
 	}
 	if len(lr.localMeta) <= 1 || len(lr.localPaths) <= 1 {
@@ -453,6 +461,20 @@ func ingestLink(
 		return err
 	}
 	for i := range sharedObjMetas {
+		// One corner case around file sizes we need to be mindful of, is that
+		// if one of the shareObjs was initially created by us (and has boomeranged
+		// back from another node), we'll need to update the FileBacking's size
+		// to be the true underlying size. Otherwise, we could hit errors when we
+		// open the db again after a crash/restart (see checkConsistency in open.go),
+		// plus it more accurately allows us to prioritize compactions of files
+		// that were originally created by us.
+		if !objProvider.IsForeign(sharedObjMetas[i]) {
+			size, err := objProvider.Size(sharedObjMetas[i])
+			if err != nil {
+				return err
+			}
+			lr.sharedMeta[i].FileBacking.Size = uint64(size)
+		}
 		if opts.EventListener.TableCreated != nil {
 			opts.EventListener.TableCreated(TableCreateInfo{
 				JobID:   jobID,
@@ -1455,6 +1477,13 @@ func (d *DB) excise(
 			if err != nil {
 				return nil, err
 			}
+			if leftFile.Size == 0 {
+				// On occasion, estimateSize gives us a low estimate, i.e. a 0 file size,
+				// such as if the excised file only has range keys/dels and no point
+				// keys. This can cause panics in places where we divide by file sizes.
+				// Correct for it here.
+				leftFile.Size = 1
+			}
 			if err := leftFile.Validate(d.cmp, d.opts.Comparer.FormatKey); err != nil {
 				return nil, err
 			}
@@ -1557,6 +1586,13 @@ func (d *DB) excise(
 		rightFile.Size, err = d.tableCache.estimateSize(m, rightFile.Smallest.UserKey, rightFile.Largest.UserKey)
 		if err != nil {
 			return nil, err
+		}
+		if rightFile.Size == 0 {
+			// On occasion, estimateSize gives us a low estimate, i.e. a 0 file size,
+			// such as if the excised file only has range keys/dels and no point keys.
+			// This can cause panics in places where we divide by file sizes. Correct
+			// for it here.
+			rightFile.Size = 1
 		}
 		ve.NewFiles = append(ve.NewFiles, newFileEntry{Level: level, Meta: rightFile})
 		if !backingTableCreated {
--- a/internal/keyspan/truncate.go
+++ b/internal/keyspan/truncate.go
@@ -14,7 +14,7 @@ func Truncate(
 	iter FragmentIterator,
 	lower, upper []byte,
 	start, end *base.InternalKey,
-	panicOnPartialOverlap bool,
+	panicOnUpperTruncate bool,
 ) FragmentIterator {
 	return Filter(iter, func(in *Span, out *Span) (keep bool) {
 		out.Start, out.End = in.Start, in.End
@@ -57,7 +57,6 @@ func Truncate(
 		var truncated bool
 		// Truncate the bounds to lower and upper.
 		if cmp(in.Start, lower) < 0 {
-			truncated = true
 			out.Start = lower
 		}
 		if cmp(in.End, upper) > 0 {
@@ -65,8 +64,8 @@ func Truncate(
 			out.End = upper
 		}
 
-		if panicOnPartialOverlap && truncated {
-			panic("pebble: bounds should not be truncated")
+		if panicOnUpperTruncate && truncated {
+			panic("pebble: upper bound should not be truncated")
 		}
 
 		return !out.Empty() && cmp(out.Start, out.End) < 0
--- a/internal/manifest/version.go
+++ b/internal/manifest/version.go
@@ -288,6 +288,33 @@ type PhysicalFileMeta struct {
 // VirtualFileMeta is used by functions which want a guarantee that their input
 // belongs to a virtual sst and not a physical sst.
 //
+// A VirtualFileMeta inherits all the same fields as a FileMetadata. These
+// fields have additional invariants imposed on them, and/or slightly varying
+// meanings:
+//   - Smallest and Largest (and their counterparts
+//     {Smallest, Largest}{Point,Range}Key) remain tight bounds that represent a
+//     key at that exact bound. We make the effort to determine the next smallest
+//     or largest key in an sstable after virtualizing it, to maintain this
+//     tightness. If the largest is a sentinel key (IsExclusiveSentinel()), it
+//     could mean that a rangedel or range key ends at that user key, or has been
+//     truncated to that user key.
+//   - One invariant is that if a rangedel or range key is truncated on its
+//     upper bound, the virtual sstable *must* have a rangedel or range key
+//     sentinel key as its upper bound. This is because truncation yields
+//     an exclusive upper bound for the rangedel/rangekey, and if there are
+//     any points at that exclusive upper bound within the same virtual
+//     sstable, those could get uncovered by this truncation. We enforce this
+//     invariant in calls to keyspan.Truncate.
+//   - Size is an estimate of the size of the virtualized portion of this sstable.
+//     The underlying file's size is stored in FileBacking.Size, though it could
+//     also be estimated or could correspond to just the referenced portion of
+//     a file (eg. if the file originated on another node).
+//   - SmallestSeqNum and LargestSeqNum are loose bounds for virtual sstables.
+//     This means that all keys in the virtual sstable must have seqnums within
+//     [SmallestSeqNum, LargestSeqNum], however there's no guarantee that there's
+//     a key with a seqnum at either of the bounds. Calculating tight seqnum
+//     bounds would be too expensive and deliver little value.
+//
 // NB: This type should only be constructed by calling FileMetadata.VirtualMeta.
 type VirtualFileMeta struct {
 	*FileMetadata
--- a/open.go
+++ b/open.go
@@ -1086,9 +1086,15 @@ func checkConsistency(v *manifest.Version, dirname string, objProvider objstorag
 			dedup[backingState.DiskFileNum] = struct{}{}
 			fileNum := backingState.DiskFileNum
 			fileSize := backingState.Size
+			// We allow foreign objects to have a mismatch between sizes. This is
+			// because we might skew the backing size stored by our objprovider
+			// to prevent us from over-prioritizing this file for compaction.
 			meta, err := objProvider.Lookup(base.FileTypeTable, fileNum)
 			var size int64
 			if err == nil {
+				if objProvider.IsForeign(meta) {
+					continue
+				}
 				size, err = objProvider.Size(meta)
 			}
 			if err != nil {
--- a/scan_internal.go
+++ b/scan_internal.go
@@ -750,7 +750,7 @@ func (d *DB) truncateSharedFile(
 	if rangeDelIter != nil {
 		rangeDelIter = keyspan.Truncate(
 			cmp, rangeDelIter, lower, upper, nil, nil,
-			false, /* panicOnPartialOverlap */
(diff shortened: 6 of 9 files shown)


bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2672)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`sstable/reader.go` line 2963 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NYtcnq95wHeyv9yU8c1:-NYtcnq95wHeyv9yU8c2:b-55mr1) ([raw file](https://github.com/cockroachdb/pebble/blob/975fbcb07b9285309864318d64e02fa9c33f1511/sstable/reader.go#L2963)):*
> ```Go
> 	return keyspan.Truncate(
> 		v.reader.Compare, iter, v.vState.lower.UserKey, v.vState.upper.UserKey,
> 		&v.vState.lower, &v.vState.upper, true, /* panicOnPartialOverlap */
> ```

A `Span` has an inclusive start key, and an exclusive end key. The bounds used for truncation are `v.vState.lower.UserKey, v.vState.upper.UserKey`.

`v.vState.upper.UserKey` can be inclusive. Let's say it is.

Let's say we're truncating some `Span` `s` with bounds `s.Start`, and `s.End`. And `S.End` > `v.vState.upper.UserKey`.  So, we'll truncate `End` to `v.vState.upper.UserKey`. This is probably incorrect, because the end key of the span is always treated as exclusive, whereas we want to treat `v.vState.upper.UserKey` as inclusive.

Instead of passing in `&v.vState.upper` into the `Truncate` function, could we manufacture a key which is lexicographically one greater than > `v.vState.upper.UserKey` ?

itsbilal (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2672)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @bananabrick)

___
*[`sstable/reader.go` line 2963 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NYtcnq95wHeyv9yU8c1:-NYygbFX42OzRxi0lWvS:bksbk3n) ([raw file](https://github.com/cockroachdb/pebble/blob/975fbcb07b9285309864318d64e02fa9c33f1511/sstable/reader.go#L2963)):*

A `Span` has an inclusive start key, and an exclusive end key. The bounds used for truncation are `v.vState.lower.UserKey, v.vState.upper.UserKey`.

`v.vState.upper.UserKey` can be inclusive. Let's say it is.

Let's say we're truncating some `Span` `s` with bounds `s.Start`, and `s.End`. And `S.End` > `v.vState.upper.UserKey`.  So, we'll truncate `End` to `v.vState.upper.UserKey`. This is probably incorrect, because the end key of the span is always treated as exclusive, whereas we want to treat `v.vState.upper.UserKey` as inclusive.

Instead of passing in `&v.vState.upper` into the `Truncate` function, could we manufacture a key which is lexicographically one greater than > `v.vState.upper.UserKey` ?
</blockquote></details>

You bring up a good point, which is that a virtual sstable `[a#4,SET-b#7,SET]` that has a rangedel `[aa-e)` inside it would end up exposing the key at `b#7, SET`even if it is originally deleted by it, because we'll truncate the rangedel to an exclusive at the file upper bound even if it's inclusive there. Since the "generate a key that's exactly one higher" problem is not worth generalizing all the way up to Cockroach, we can achieve the same purpose of this PR by setting panicOnPartialOverlap to `!fileUpper.IsExclusiveSentinel()`, which is that we will only truncate range dels and range keys at a file bound key if the upper bound key is an exclusive sentinel. This is an easy contract to obey for both excises and ingest-time splits. We can relax the panic for any truncations on the lower end.

I've made this change, PTAL!

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2672)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @itsbilal)

___
*[`ingest.go` line 1467 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NZ7oDl45NTyHrrkxglU:-NZ7oDl45NTyHrrkxglV:b-xdy8ft) ([raw file](https://github.com/cockroachdb/pebble/blob/1f5f2a95c338f1586cc5e29652ec9b119cb6a258/ingest.go#L1467)):*
> ```Go
> 			}
> 			if leftFile.Size == 0 {
> 				// On occasion, estimateSize gives us a low estimate, i.e. a 0 file size. This
> ```

Do you know when this can return 0?

Looking at the `EstimateDiskUsage` function, the last return line is:

```
return includeInterpolatedValueBlocksSize(
    endBH.Offset + endBH.Length + blockTrailerLen - startBH.Offset), nil)
```
Shouldn't the `startBh.Offset` be <= `endBH.Offset`, so we should get a value > 0.

The function does say that only blocks containing point keys are considered. So, maybe that has something to do with this.
___
*[`sstable/reader.go` line 2963 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NYtcnq95wHeyv9yU8c1:-NYz0ixf9XOcHGpmzM22:b-gjk47v) ([raw file](https://github.com/cockroachdb/pebble/blob/975fbcb07b9285309864318d64e02fa9c33f1511/sstable/reader.go#L2963)):*
> Since the "generate a key that's exactly one higher" problem is not worth generalizing all the way up to Cockroach

Would we have to do that? Would it be sufficient to pass in the next lexicographically higher key of `vState.upper.UserKey` to the `upper` parameter of `Truncate`?

I'm fine with ensuring this invariant, but we should document/enforce these invariants on virtual sstable creation. Maybe in `FileMetadata.ValidateVirtual`. What do you think?

Could you also clarify why we can ensure this invariant for inclusive end keys but not exclusive end key?
___
*[`sstable/reader.go` line 2961 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NZ7TGyy1gSSe5BYw1E1:-NZ7TGyz7imdkZv4W89o:bykn3h3) ([raw file](https://github.com/cockroachdb/pebble/blob/1f5f2a95c338f1586cc5e29652ec9b119cb6a258/sstable/reader.go#L2961)):*
> ```Go
> 	// Truncation of spans isn't allowed at a user key that also contains points
> 	// in the same virtual sstable, as it would lead to covered points getting
> 	// uncovered. Set panicOnUpperTruncate to true if the file's upper bound
> ```

I think we should add an example here to clarify why we're doing this, and also in the comment below.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2672#-:-NZXdfL7EX_hCvXRwge7:b-814g8k)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @bananabrick)

___
*[`ingest.go` line 1467 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NZ7oDl45NTyHrrkxglU:-NZXZG3M1mafd7yk0MSt:b-4yqy6z) ([raw file](https://github.com/cockroachdb/pebble/blob/1f5f2a95c338f1586cc5e29652ec9b119cb6a258/ingest.go#L1467)):*

Do you know when this can return 0?

Looking at the `EstimateDiskUsage` function, the last return line is:

```
return includeInterpolatedValueBlocksSize(
    endBH.Offset + endBH.Length + blockTrailerLen - startBH.Offset), nil)
```
Shouldn't the `startBh.Offset` be <= `endBH.Offset`, so we should get a value > 0.

The function does say that only blocks containing point keys are considered. So, maybe that has something to do with this.

</blockquote></details>

Yeah that's basically it; we end up in cases where there are no point keys in the virtual sstable but there's a rangedel or range key. the code for `EstimateDiskUsage` has a bunch of `if key == nil { return 0, iter.Error() }` cases in it, which ends up returning `0, nil` if the iterator also didn't error out and just happened to have nothing in the range (or a seekGE at the start key put us past all the index entries).
___
*[`sstable/reader.go` line 2963 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NYtcnq95wHeyv9yU8c1:-NZXcIWc7nDt24xO1OVz:b-hvy0gl) ([raw file](https://github.com/cockroachdb/pebble/blob/975fbcb07b9285309864318d64e02fa9c33f1511/sstable/reader.go#L2963)):*

> Since the "generate a key that's exactly one higher" problem is not worth generalizing all the way up to Cockroach

Would we have to do that? Would it be sufficient to pass in the next lexicographically higher key of `vState.upper.UserKey` to the `upper` parameter of `Truncate`?

I'm fine with ensuring this invariant, but we should document/enforce these invariants on virtual sstable creation. Maybe in `FileMetadata.ValidateVirtual`. What do you think?

Could you also clarify why we can ensure this invariant for inclusive end keys but not exclusive end key?




</blockquote></details>

The issue is that generating the "next lexicographically higher key of x" is not a problem we want to solve, as it reduces generality. Pebble so far has stayed away from having a comparer method that does that, and we probably want to keep it that way.

We should check for this in `ValidateVirtual`, yes, but we already have https://github.com/cockroachdb/pebble/issues/2593  to do that.
___
*[`sstable/reader.go` line 2961 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NZ7TGyy1gSSe5BYw1E1:-NZXc3NH0TR_O6sADspn:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/1f5f2a95c338f1586cc5e29652ec9b119cb6a258/sstable/reader.go#L2961)):*

I think we should add an example here to clarify why we're doing this, and also in the comment below.
</blockquote></details>

Done.

sumeerbhola (changes requested):
> Reviewed 1 of 8 files at r2, 3 of 5 files at r3, 2 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2672)* status: 6 of 9 files reviewed, 6 unresolved discussions (waiting on @bananabrick and @itsbilal)

___
*[`-- commits` line 10 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NZ_zw-S0ajjb51p0sMt):*
I am assuming this could also happen because a CockroachDB range got split, and there was a rangedel  or rangekey that was written before the split and later we excised one of the post-split ranges?
___
*[`ingest.go` line 1467 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NZ7oDl45NTyHrrkxglU:-NZa6N8R6V5u3OXqTrVD:b-7j2g8w) ([raw file](https://github.com/cockroachdb/pebble/blob/1f5f2a95c338f1586cc5e29652ec9b119cb6a258/ingest.go#L1467)):*
> we end up in cases where there are no point keys in the virtual sstable but there's a rangedel or range key.

can you add this as a code comment here.
___
*[`ingest.go` line 1578 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NZa6S1nD8NuGTCN2wpg:-NZa6S1nD8NuGTCN2wph:b-7ropln) ([raw file](https://github.com/cockroachdb/pebble/blob/77fa822d62d334ec4f525db666e2d6924cd8880f/ingest.go#L1578)):*
> ```Go
> 			// On occasion, estimateSize gives us a low estimate, i.e. a 0 file size. This
> 			// can cause panics in places where we divide by file sizes. Correct for it
> 			// here.
> ```

ditto
___
*[`internal/keyspan/truncate.go` line 68 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NZa7tXmARcce3UpajRl:-NZa7tXmARcce3UpajRm:b-kdogrj) ([raw file](https://github.com/cockroachdb/pebble/blob/77fa822d62d334ec4f525db666e2d6924cd8880f/internal/keyspan/truncate.go#L68)):*
> ```Go
> 
> 		if panicOnUpperTruncate && truncated {
> 			panic("pebble: bounds should not be truncated")
> ```

nit: upper bound should not be truncated
___
*[`sstable/reader.go` line 2963 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NYtcnq95wHeyv9yU8c1:-NZaAaTD0iui9rXoZU74:b-32i932) ([raw file](https://github.com/cockroachdb/pebble/blob/975fbcb07b9285309864318d64e02fa9c33f1511/sstable/reader.go#L2963)):*

The issue is that generating the "next lexicographically higher key of x" is not a problem we want to solve, as it reduces generality. Pebble so far has stayed away from having a comparer method that does that, and we probably want to keep it that way.

We should check for this in `ValidateVirtual`, yes, but we already have https://github.com/cockroachdb/pebble/issues/2593  to do that.
</blockquote></details>

I am getting confused by the reasoning here, so I'll try to replay what I can re-construct from this discussion and code.
- The virtualization of a table always uses userkey bound of the form [start-user-key, end-user-key), whether this happens via excise (for the left and right tables) or when exporting a shared file.
- The largest point key in this sst must be < end-user-key.
- A rangedel is truncated by this virtualization on the right-hand-side iff it straddles end-user-key. In that case the end-user-key exclusive sentinel is the contribution of this rangedel to the virtual sst upper bound. Then truncating at that end-user-key does not expose any covered point key.
- A similar argument applies to rangekeys. However, when both a rangedel and rangekey straddle the end-user-key, the sst upper bound will be due to the rangedel sentinel (since rangedel InternalKeyKind sorts after). This is fine since we use the general `v.vState.upper.IsExclusiveSentinel()` call to decide whether truncation is permitted or not.

We have lost some fidelity in that we will permit truncation of a rangedel if there is a rangekey that straddled end-user-key but not a rangedel (and vice versa), but that is ok from a correctness perspective since there won't actually be any truncation of the rangedel in that case.

Is this accurate?

It would be nice if we had a code comment that laid this out clearly in some high-level place where we define sst virtualization.

itsbilal (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2672#-:-NZgAyDs9M7ZxrPV6B6H:bz5mair)* status: 6 of 9 files reviewed, 5 unresolved discussions (waiting on @bananabrick and @sumeerbhola)

___
*[`-- commits` line 10 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NZ_zw-S0ajjb51p0sMt):*

I am assuming this could also happen because a CockroachDB range got split, and there was a rangedel  or rangekey that was written before the split and later we excised one of the post-split ranges?
</blockquote></details>

That's right. That would also do it.
___
*[`ingest.go` line 1467 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NZ7oDl45NTyHrrkxglU:-NZg3xBDAg2f-i5kgX7_:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/1f5f2a95c338f1586cc5e29652ec9b119cb6a258/ingest.go#L1467)):*

> we end up in cases where there are no point keys in the virtual sstable but there's a rangedel or range key.

can you add this as a code comment here.
</blockquote></details>

Done.
___
*[`ingest.go` line 1578 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NZa6S1nD8NuGTCN2wpg:-NZg49kW95D7m_6bl0Ro:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/77fa822d62d334ec4f525db666e2d6924cd8880f/ingest.go#L1578)):*

ditto
</blockquote></details>

Done.
___
*[`internal/keyspan/truncate.go` line 68 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NZa7tXmARcce3UpajRl:-NZg4AMe-6AbN_tJzFX3:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/77fa822d62d334ec4f525db666e2d6924cd8880f/internal/keyspan/truncate.go#L68)):*

nit: upper bound should not be truncated
</blockquote></details>

Done.
___
*[`sstable/reader.go` line 2963 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NYtcnq95wHeyv9yU8c1:-NZg4GwaEdmncH2hRqc9:bsdvqd5) ([raw file](https://github.com/cockroachdb/pebble/blob/975fbcb07b9285309864318d64e02fa9c33f1511/sstable/reader.go#L2963)):*

I am getting confused by the reasoning here, so I'll try to replay what I can re-construct from this discussion and code.
- The virtualization of a table always uses userkey bound of the form [start-user-key, end-user-key), whether this happens via excise (for the left and right tables) or when exporting a shared file.
- The largest point key in this sst must be < end-user-key.
- A rangedel is truncated by this virtualization on the right-hand-side iff it straddles end-user-key. In that case the end-user-key exclusive sentinel is the contribution of this rangedel to the virtual sst upper bound. Then truncating at that end-user-key does not expose any covered point key.
- A similar argument applies to rangekeys. However, when both a rangedel and rangekey straddle the end-user-key, the sst upper bound will be due to the rangedel sentinel (since rangedel InternalKeyKind sorts after). This is fine since we use the general `v.vState.upper.IsExclusiveSentinel()` call to decide whether truncation is permitted or not.

We have lost some fidelity in that we will permit truncation of a rangedel if there is a rangekey that straddled end-user-key but not a rangedel (and vice versa), but that is ok from a correctness perspective since there won't actually be any truncation of the rangedel in that case.

Is this accurate?

It would be nice if we had a code comment that laid this out clearly in some high-level place where we define sst virtualization.
</blockquote></details>

What you described is mostly accurate. One of the things to keep in mind is that while excise/truncation of shared files happens with a `[start, end)` user key bound in mind, the actual `FileMetadata` we generate for the truncated file(s) is of a `[start, end]` bound that's tight on user keys. We actually iterate the sstable to find the new smallest and largest internal keys, and set the SST to that.

The complex cases are, say, when you have an sstable 00009:

```
00009.sst:
a.RANGEKEYSET.8:d
a.SET.5:foo
b.SET.6:bar
```

And you're excising `b-e`. In this case the new LARGEST will be `b#RANGEKEYSET,inf`and the smallest will stay the same. The point keys get excluded. Since excise always operates on a `[smallest, largest)` excised span, if we're in the case where we're truncating away the right half of a rangekey or rangedel, we can always set the largest for the sstable left of the excise to a range del/key sentinel. We can also observe that in *all* cases where we truncate a range key/del on the upper bound, we will always have a fileMetadata with a largest key that's a sentinel key. So we leverage that observation to update this assertion to allow upper-bound truncates only if fileMeta.Largest.IsExclusiveSentinel() is true.

We don't need to pay too much attention to whether that's a rangedel or range key sentinel, at least from a correctness perspective. 

I already added a comment with a similar example in `VirtualReader.NewRawRange{Del,Key}Iter`. The unfortunate thign is that there's no single place where we define sstable virtualization. Maybe we should also copy or reference this comment in the places where we create virtual sstables (currently it's only `ingestLoad1Shared/truncateSharedFiles` and `excise`).

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 9 files at r1, 1 of 3 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2672#-:-NZh5_yj9aY7nnjc3wk7:bg7igkv)* status: 8 of 9 files reviewed, 5 unresolved discussions (waiting on @bananabrick)

___
*[`sstable/reader.go` line 2963 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NYtcnq95wHeyv9yU8c1:-NZgzc1YEl3_9DeghbXh:bepppx2) ([raw file](https://github.com/cockroachdb/pebble/blob/975fbcb07b9285309864318d64e02fa9c33f1511/sstable/reader.go#L2963)):*
> We actually iterate the sstable to find the new smallest and largest internal keys, and set the SST to that.

Right. I didn't mention that in the comment since the reasoning about correctness did not seem to require that assumption. But considering that in the example definitely helps the reader.

I think we could use a long code comment on `VirtualFileMeta` -- since that is just a pointer to a `FileMetadata` we have implicitly inherited the comments on the `FileMetadata` fields, which don't tell us anything about what `Smallest*`, `Largest*` on a virtual sst mean and what invariants they must satisfy relative to the actual data in the physical sst.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2672#-:-NZlqgYo51L4P1BOt1lV:b-814g8k)* status: 8 of 10 files reviewed, 5 unresolved discussions (waiting on @bananabrick and @sumeerbhola)

___
*[`sstable/reader.go` line 2963 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2672#-NYtcnq95wHeyv9yU8c1:-NZlqiEC4wxDFxCv6xs3:b-t4k8ut) ([raw file](https://github.com/cockroachdb/pebble/blob/975fbcb07b9285309864318d64e02fa9c33f1511/sstable/reader.go#L2963)):*

> We actually iterate the sstable to find the new smallest and largest internal keys, and set the SST to that.

Right. I didn't mention that in the comment since the reasoning about correctness did not seem to require that assumption. But considering that in the example definitely helps the reader.

I think we could use a long code comment on `VirtualFileMeta` -- since that is just a pointer to a `FileMetadata` we have implicitly inherited the comments on the `FileMetadata` fields, which don't tell us anything about what `Smallest*`, `Largest*` on a virtual sst mean and what invariants they must satisfy relative to the actual data in the physical sst.
 
</blockquote></details>

Added the comment, good idea. Thanks!

PR #2673, opened 2023-06-24
db: improve delete pacing

### LINKED ISSUE #2662: db: adaptively throttle file deletions
Observed during a scale test that strove to write 50 MiB/s/node in user data. The goroutine count was increasing monotonically. The growth was all goroutines responsible for deleting files, which were being paced at the default rate 128 MiB of delete files per second. The already in-flight #2641 will resolve the goroutine growth, but it does not resolve the issue of falling behind in file deletions.

We should not be setting a fixed deletion rate, and instead should smooth deletions based on the rate file deletions observed. See https://github.com/cockroachdb/pebble/pull/2641#pullrequestreview-1491426251

Screenshot 2023-06-21 at 6 13 11 PM
Screenshot 2023-06-21 at 6 13 48 PM

RaduBerinde:
Delete pacing is currently an on or off decision. If we are running
out of space or have too many obsolete bytes in relation to live
bytes, we disable pacing. Otherwise we pace at the configured rate
(128MB by default in CRDB).

This change improves pacing by keeping track of the average deletion
rate over the last 5 minutes and increasing the target rate to match
this rate if necessary. The intention is to avoid deletions lagging
behind.

Fixes #2662.

### DIFF
--- a/cleaner.go
+++ b/cleaner.go
@@ -74,7 +74,7 @@ func openCleanupManager(
 		opts:            opts,
 		objProvider:     objProvider,
 		onTableDeleteFn: onTableDeleteFn,
-		deletePacer:     newDeletionPacer(getDeletePacerInfo),
+		deletePacer:     newDeletionPacer(time.Now(), int64(opts.TargetByteDeletionRate), getDeletePacerInfo),
 		jobsCh:          make(chan *cleanupJob, jobsChLen),
 	}
 	cm.mu.completedJobsCond.L = &cm.mu.Mutex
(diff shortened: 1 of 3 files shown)


jbowens (approved):
Looks great!

:lgtm:

> Reviewed 22 of 22 files at r1, 2 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2673#-:-NYxSlEN34UDyPCFaNk3:b8reftd)* status: 2 of 22 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`pacer.go` line 169 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2673#-NYxRA3Y62saeTR8U9Xz:-NYxRA3Y62saeTR8U9Y-:bul4zij) ([raw file](https://github.com/cockroachdb/pebble/blob/ca89ca270f4ee71fcd8531eb508b9c6c2ff93970/pacer.go#L169)):*
> ```Go
> 	for h.currEpoch < epoch {
> 		h.currEpoch++
> 		// Forget the data for the oldest second.
> ```

nit: s/second/epoch/

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2673#-:-NYxgji23jq-qK3sAYS3:b-814g8k)* status: 1 of 22 files reviewed, all discussions resolved (waiting on @jbowens)

PR #2682, opened 2023-06-26
scan_internal: return err if skip-shared breaks snapshot consistency

itsbilal:
Previously, we'd unintentionally expose newer keys in shared sstables through a snapshot if a compaction pushed keys newer than the snapshot down into a shared level. We can identify this case and return `ErrInvalidSkipSharedIteration`, forcing the caller to resort to non-skip-shared ways to replicate keys.


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2682#-:-NYtlX2pBC9xD4lN6WaF:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @sumeerbhola)

itsbilal:
TFTR!

PR #2686, opened 2023-06-27
scan_internal: Fix seqnum comparison

itsbilal:
Previously, we were returning an error if the largest seqnum of a file was strictly greater than the snapshot seqnum in skip-shared iteration and the file was in a shared level, for consistency reasons. However as snapshots only make visible keys that are strictly less than the snapshot sequence number, the comparison should have been greater than or equals.

This change resolves that by replacing the comparison with a call to base.Visible.


itsbilal:
TFTR!

PR #2694, opened 2023-06-29
metamorphic: test shared storage including the secondary cache

joshimhoff:
**metamorphic: test shared storage including the secondary cache**

This commit adds a standard option to the metamorphic tests that tests shared storage including the secondary cache.

This commit fixes a buglet in the vfs stack involving incorrect OpenReadWrite implementations, which affect the secondary cache and were caught by the changes to the metamorphic tests.

To test the test changes, I introduced a bug in the secondary cache like so:

```
b/objstorage/objstorageprovider/sharedcache/shared_cache.go @@ -127,6 +127,7 @@ func (c *Cache) ReadAt(
                // Note this. The below code does not need the original ofs, as with the earlier
                // reading from the cache done, the relevant offset is ofs + int64(n). Same with p.
                ofs += int64(n)
 +              ofs += 3
                p = p[n:]
```

With this change, the test failed with the following error:

```
backing file 000006 error: pebble/table: invalid table (bad magic number: ...)
```

### DIFF
--- a/metamorphic/ops.go
+++ b/metamorphic/ops.go
@@ -102,6 +102,16 @@ type checkpointOp struct {
 }
 
 func (o *checkpointOp) run(t *test, h historyRecorder) {
+	// TODO(josh): db.Checkpoint does not work with shared storage yet.
+	// It would be better to filter out ahead of calling run on the op,
+	// by setting the weight that generator.go uses to zero, or similar.
+	// But IIUC the ops are shared for ALL the metamorphic test runs, so
+	// not sure how to do that easily:
+	// https://github.com/cockroachdb/pebble/blob/master/metamorphic/meta.go#L177
+	if t.testOpts.sharedStorageEnabled {
+		h.Recordf("%s // %v", o, nil)
+		return
+	}
 	var opts []pebble.CheckpointOption
 	if len(o.spans) > 0 {
 		opts = append(opts, pebble.WithRestrictToSpans(o.spans))
@@ -1209,6 +1219,11 @@ func (o *dbRatchetFormatMajorVersionOp) syncObjs() objIDSlice { return nil }
 type dbRestartOp struct{}
 
 func (o *dbRestartOp) run(t *test, h historyRecorder) {
+	// TODO(josh): db.Restart does not work with shared storage yet.
+	if t.testOpts.sharedStorageEnabled {
+		h.Recordf("%s", o)
+		return
+	}
 	if err := t.restartDB(); err != nil {
 		h.Recordf("%s // %v", o, err)
 		h.history.err.Store(errors.Wrap(err, "dbRestartOp"))
--- a/metamorphic/options.go
+++ b/metamorphic/options.go
@@ -19,6 +19,7 @@ import (
 	"github.com/cockroachdb/pebble/bloom"
 	"github.com/cockroachdb/pebble/internal/cache"
 	"github.com/cockroachdb/pebble/internal/testkeys"
+	"github.com/cockroachdb/pebble/objstorage/shared"
 	"github.com/cockroachdb/pebble/sstable"
 	"github.com/cockroachdb/pebble/vfs"
 	"golang.org/x/exp/rand"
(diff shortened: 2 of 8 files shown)


joshimhoff:
@itsbilal says:

> Ah right. I’d disable checkpoints in the generator by lowering their weight to 0. That way the metamoprhic test ops generator won’t produce any checkpoint operations

Will do now.

joshimhoff:
OK, this is ready for an actual review now. The tests pass. The way I implemented disable checkpoints is odd. Prob there is a better way.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2694)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @joshimhoff)

___
*[`options.go` line 675 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2694#-NZ6ytRfDHBh2P84eszq:-NZ6ytRfDHBh2P84eszr:b-2jcg72) ([raw file](https://github.com/cockroachdb/pebble/blob/7b35bd2a8b855d1e78be5e4bfc846d973c65b545/options.go#L675)):*
> ```Go
> 		// This ID is incorporated in backing object names.
> 		// Must be non-zero.
> 		CreatorID objstorage.CreatorID
> ```

Could the test run `SetCreatorID` instead of adding this option? I'd like to avoid having two different ways of setting it if possible

joshimhoff (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2694)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`options.go` line 675 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2694#-NZ6ytRfDHBh2P84eszq:-NZBu0h80CA0JiHn182W:b-h990k8) ([raw file](https://github.com/cockroachdb/pebble/blob/7b35bd2a8b855d1e78be5e4bfc846d973c65b545/options.go#L675)):*

Could the test run `SetCreatorID` instead of adding this option? I'd like to avoid having two different ways of setting it if possible
</blockquote></details>

Good call. Fixed.

joshimhoff:
Gentle poke on review, @itsbilal! Radu is OOO for the week.

RaduBerinde (approved):
:lgtm:

> Reviewed 3 of 6 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2694#-:-NZbW0Wb5ctCMkCmQOZ0:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @itsbilal)

metamorphic/ops.go
```diff
@@ -102,6 +102,16 @@ type checkpointOp struct {
 }
 
 func (o *checkpointOp) run(t *test, h historyRecorder) {
+	// TODO(josh): db.Checkpoint does not work with shared storage yet.
+	// It would be better to filter out ahead of calling run on the op,
+	// by setting the weight that generator.go uses to zero, or similar.
+	// But IIUC the ops are shared for ALL the metamorphic test runs, so
+	// not sure how to do that easily:
```
itsbilal, line 109: This works for now. I was thinking that we just create a new `TestMetaShared` test that enables shared storage options and disables the checkpoint op. But that duplicates test churn without that much of an additional win, so faking a successful checkpoint in the record is good enough.

metamorphic/options.go
```diff
@@ -339,6 +362,11 @@ func standardOptions() []*TestOptions {
 		25: `
 [TestOptions]
   enable_value_blocks=true
+`,
+		26: `
+[TestOptions]
```
itsbilal, line 367: The standard option looks great! maybe we should also add an ability for the random options below to toggle these?

itsbilal (approved):
LGTM with one improvement suggestion, thanks for doing this!

metamorphic/ops.go
joshimhoff: Thanks!

metamorphic/options.go
joshimhoff: Will do.

joshimhoff (commented):
TTFR!

joshimhoff:
I've adjusted the tests to also skip `dbRestartOp`, if running with shared storage. Discussion thread created at https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1688999102475869. I will merge like this for now, as (i) even with `dbRestartOp` skipped, this PR strictly adds test coverage, and (ii) I have a PR queued behind this one that I would like to put up for review related to secondary cache benchmarking.

joshimhoff:
Follow up: https://github.com/cockroachdb/pebble/issues/2717

PR #2696, opened 2023-06-29
crl-release-23.1: db: improve delete pacing

RaduBerinde:
Delete pacing is currently an on or off decision. If we are running out of space or have too many obsolete bytes in relation to live bytes, we disable pacing. Otherwise we pace at the configured rate (128MB by default in CRDB).

This change improves pacing by keeping track of the average deletion rate over the last 5 minutes and increasing the target rate to match this rate if necessary. The intention is to avoid deletions lagging behind.

Informs #2662.

I have tested this change against a heavy write workload where deletion is known to lag behind (resulting in 10k+ goroutines waiting in the pacer); with the change, the goroutine count is small and stable (after the first 5-10 minutes):
image

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -3547,6 +3547,13 @@ func (d *DB) doDeleteObsoleteFiles(jobID int) {
 		}
 	}
 	if len(filesToDelete) > 0 {
+		var tablesSize uint64
+		for i := range filesToDelete {
+			if filesToDelete[i].fileType == fileTypeTable {
+				tablesSize += filesToDelete[i].fileSize
+			}
+		}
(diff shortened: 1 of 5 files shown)


RaduBerinde:
Note: this is a backport of #2673, adapted to the 23.1 codebase.

jbowens (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2696#-:-NZBiqyg0Y6g5oUyy1qi:b-uwkowd)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

PR #2697, opened 2023-06-29
sharedcache: data-driven test cosmetic improvements

RaduBerinde:
#### sharedcache: use arguments for reads and writes

This commit changes the read/write test directives to use named
arguments. This allows using K/M/G suffixes and makes the tests easier
to read. We also swap the order of offset and size in the tests since
the offset is typically first.


joshimhoff (approved):
Much nicer!

Issue #2693, opened 2023-06-28
cache: block allocations during compaction cause expensive TLB shootdowns


nvanbenschoten:
See Slack thread: https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1687921559962319.

> `perf` profiles from a write-heavy workload show a concerning amount of cpu time performing [TLB shootdowns](https://github.com/bitcharmer/tlb_shootdowns) (`native_flush_tlb_multi`) under `__madvise` (I believe this is a call from glibc by jemalloc and not the go runtime) and when page faulting in `cache.newValue`. We see an improvement with numa alignment, but the issue is still severe (27% of cpu without alignment, 18% with alignment).

This is reproducible on AWS with the following steps:
```bash
roachprod create $USER-test -n6 --clouds=aws --aws-machine-type-ssd='i4i.8xlarge' --aws-enable-multiple-stores=true
roachprod stage $USER-test cockroach # master
roachprod start $USER-test:1-5 --env=COCKROACH_ROCKSDB_CONCURRENCY=12
roachprod sql $USER-test:1 -- -e='ALTER RANGE default CONFIGURE ZONE USING num_replicas = 5'
roachprod run $USER-test:6 -- './cockroach workload init kv --splits=20 --scatter {pgurl:1-5}'
roachprod run $USER-test:6 -- './cockroach workload run kv --concurrency=1024 --max-rate=60000 --min-block-bytes=2048 --max-block-bytes=2048 --ramp=1m --target-compression-ratio=1.5 {pgurl:1-5}'
```

The leading theory here is that placing uncompressed blocks in the block cache during compaction leads to a high likelihood that it is freed by a different thread than the one that allocated it, and that this makes the tcache ineffective (page fault on rewrite, cross-core tlb flush on madvise). To verify this, we should prototype a change that does not populate the block cache during compaction reads.

----

![perf_aws_32](https://github.com/cockroachdb/pebble/assets/5438456/62b30383-6bad-4486-ba1d-94b856445c90)

Issue #1997, opened 2022-10-05
cache: investigate mutex contention with TPCC


sumeerbhola:
@ajwerner reported high mutex contention on the block cache when running TPCC with 16 core machines (internal link: https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1664480089719489), specifically with nodes under high load (running tpccbench). And also pointed to a discussion regarding poor `RWMutex` scaling with high core count https://github.com/golang/go/issues/17973.

The diff from the 2 profiles (that are 60s apart) in that internal link is shown below and shows a total contention of 122s. 34s is in the sstable cache.
<img width="1570" alt="Screen Shot 2022-10-05 at 5 28 00 PM" src="https://user-images.githubusercontent.com/54990988/194167673-4d09f55f-e3c8-41ef-b437-55cd03ae7d7f.png">

There doesn't seem to be an easy fix here but we should reproduce this experiment and measure (a) mutex contention, (b) mutex related overhead in a cpu profile, (c) the number of concurrent MVCCGet calls. If (b) is "small" and (c) is "large", the contention walltime may be acceptable, since it would not impact the peak throughput we can achieve on that node.

andrewbaptist:
In a recent customer investigation (https://github.com/cockroachlabs/support/issues/2066) I was looking at what the runnable goroutines on the system were doing to try and determine why requests were taking longer than expected. There are a large number (445) of runnable goroutines on the system. Looking more into what they were doing, about half of them were waiting in the pebble layer. Note that this does not include goroutines that are in the `semacquire` state which would be what I naively expected.

Instead, these were busy attempting to get a lock, not waiting for the OS signal.
There are a few things that they are waiting for, but breaking them down it looks like they are almost all waiting on the pebble shard lock. I see that it is set to 2xgoprocs, but maybe that is not enough? 

[goroutine_dump.2023-02-12T06_50_56.716.double_since_last_dump.000002714.txt](https://github.com/cockroachdb/pebble/files/10757799/goroutine_dump.2023-02-12T06_50_56.716.double_since_last_dump.000002714.txt)

Looking at what all the runnable threads are doing:
```
% grep -h -A1 "^goroutine.*runnable" goroutine_dump.2023-02-12T06_50_56.716.double_since_last_dump.000002714.txt | grep -v goroutine | grep -v "^--" | grep -o '^.*(' | sort | uniq -c | sort -n  | tail -10
[...]

PR #2699, opened 2023-06-29
db: do not cache compaction block reads

jbowens:
**db: do not cache compaction block reads**

During compactions, avoid populating the block cache with input files' blocks.
These files will soon be removed from the LSM, making it less likely any
iterator will need to read these blocks. While Pebble uses a scan-resistant
block cache algorithm (ClockPRO), the act of inserting the blocks into the
cache increases contention on the block cache mutexes (https://github.com/cockroachdb/pebble/issues/1997). This contention
has been observed to significantly contribute to tail latencies, both for reads
and for writes during memtable reservation. Additionally, although these blocks
may be soon replaced with more useful blocks due to ClockPRO's scan resistance,
they may be freed by a different thread inducing excessive TLB shootdowns
(https://github.com/cockroachdb/pebble/issues/2693).

A compaction only requires a relatively small working set of buffers during its
scan across input sstables. In this commit, we introduce a per-compaction
BufferPool that is used to allocate buffers during cache misses. Buffers are
reused throughout the compaction and only freed to the memory allocator when
they're too small or the compaction is finished. This reduces pressure on the
memory allocator and the block cache.

**sstable: avoid caching meta blocks**

When opening a sstable for the first time, we read two 'meta' blocks: one
describes the layout of sstable, encoding block handles for the index block,
properties block, etc. The other contains table properties. These blocks are
decoded, and the necessary state is copied onto the heap, and then blocks are
released. In typical configurations with sufficiently large table caches, these
blocks are never read again or only much later after the table has been evicted
from the table cache.

This commit uses a BufferPool to hold these temporary blocks, and refrains from
adding these blocks to the block cache. This reduces contention on the block
cache mutexes (https://github.com/cockroachdb/pebble/issues/1997).

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -566,6 +566,7 @@ type compaction struct {
 	// resulting version has been installed (if successful), but the compaction
 	// goroutine is still cleaning up (eg, deleting obsolete files).
 	versionEditApplied bool
+	bufferPool         sstable.BufferPool
 
 	score float64
 
@@ -1342,7 +1343,10 @@ func (c *compaction) newInputIter(
 		f manifest.LevelFile, _ *IterOptions, l manifest.Level, bytesIterated *uint64,
 	) (keyspan.FragmentIterator, error) {
 		iter, rangeDelIter, err := newIters(context.Background(), f.FileMetadata,
-			&IterOptions{level: l}, internalIterOpts{bytesIterated: &c.bytesIterated})
+			&IterOptions{level: l}, internalIterOpts{
+				bytesIterated: &c.bytesIterated,
+				bufferPool:    &c.bufferPool,
+			})
 		if err == nil {
 			// TODO(peter): It is mildly wasteful to open the point iterator only to
 			// immediately close it. One way to solve this would be to add new
@@ -1427,7 +1431,10 @@ func (c *compaction) newInputIter(
 	// to configure the levelIter at these levels to hide the obsolete points.
 	addItersForLevel := func(level *compactionLevel, l manifest.Level) error {
 		iters = append(iters, newLevelIter(iterOpts, c.cmp, nil /* split */, newIters,
-			level.files.Iter(), l, &c.bytesIterated))
+			level.files.Iter(), l, internalIterOpts{
+				bytesIterated: &c.bytesIterated,
+				bufferPool:    &c.bufferPool,
+			}))
 		// TODO(jackson): Use keyspan.LevelIter to avoid loading all the range
 		// deletions into memory upfront. (See #2015, which reverted this.)
 		// There will be no user keys that are split between sstables
@@ -2744,6 +2751,32 @@ func (d *DB) runCompaction(
 	// Note the unusual order: Unlock and then Lock.
 	d.mu.Unlock()
 	defer d.mu.Lock()
+
+	// Compactions use a pool of buffers to read blocks, avoiding polluting the
+	// block cache with blocks that will not be read again. We initialize the
+	// buffer pool with a size 12. This initial size does not need to be
+	// accurate, because the pool will grow to accommodate the maximum number of
+	// blocks allocated at a given time over the course of the compaction. But
+	// choosing a size larger than that working set avoids any additional
+	// allocations to grow the size of the pool over the course of iteration.
+	//
+	// Justification for initial size 12: In a two-level compaction, at any
+	// given moment we'll have 2 index blocks in-use and 2 data blocks in-use.
+	// Additionally, when decoding a compressed block, we'll temporarily
+	// allocate 1 additional block to hold the compressed buffer. In the worst
+	// case that all input sstables have two-level index blocks (+2), value
+	// blocks (+2), range deletion blocks (+n) and range key blocks (+n), we'll
+	// additionally require 2n+4 blocks where n is the number of input sstables.
+	// Range deletion and range key blocks are relatively rare, and the cost of
+	// an additional allocation or two over the course of the compaction is
+	// considered to be okay. A larger initial size would cause the pool to hold
+	// on to more memory, even when it's not in-use because the pool will
+	// recycle buffers up to the current capacity of the pool. The memory use of
+	// a 12-buffer pool is expected to be within reason, even if all the buffers
+	// grow to the typical size of an index block (256 KiB) which would
+	// translate to 3 MiB per compaction.
+	c.bufferPool.Init(12)
+	defer c.bufferPool.Release()
 
 	iiter, err := c.newInputIter(d.newIters, d.tableNewRangeKeyIter, snapshots)
 	if err != nil {
--- a/db_test.go
+++ b/db_test.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/cache"
 	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/objstorage/objstorageprovider"
 	"github.com/cockroachdb/pebble/sstable"
@@ -787,14 +788,12 @@ func TestIterLeakSharedCache(t *testing.T) {
 }
 
 func TestMemTableReservation(t *testing.T) {
-	cache := NewCache(128 << 10 /* 128 KB */)
-	defer cache.Unref()
-
 	opts := &Options{
-		Cache:        cache,
+		Cache:        NewCache(128 << 10 /* 128 KB */),
 		MemTableSize: initialMemTableSize,
 		FS:           vfs.NewMem(),
 	}
+	defer opts.Cache.Unref()
 	opts.testingRandomized(t)
 	opts.EnsureDefaults()
 	// We're going to be looking at and asserting the global memtable reservation
@@ -805,7 +804,7 @@ func TestMemTableReservation(t *testing.T) {
 	// cache size, so opening the DB should cause this block to be evicted.
 	tmpID := opts.Cache.NewID()
 	helloWorld := []byte("hello world")
-	value := opts.Cache.Alloc(len(helloWorld))
+	value := cache.Alloc(len(helloWorld))
 	copy(value.Buf(), helloWorld)
 	opts.Cache.Set(tmpID, base.FileNum(0).DiskFileNum(), 0, value).Release()
 
--- a/flushable.go
+++ b/flushable.go
@@ -172,7 +172,7 @@ func (s *ingestedFlushable) newIter(o *IterOptions) internalIterator {
 	// aren't truly levels in the lsm. Right now, the encoding only supports
 	// L0 sublevels, and the rest of the levels in the lsm.
 	return newLevelIter(
-		opts, s.cmp, s.split, s.newIters, s.slice.Iter(), manifest.Level(0), nil,
+		opts, s.cmp, s.split, s.newIters, s.slice.Iter(), manifest.Level(0), internalIterOpts{},
 	)
 }
 
--- a/ingest.go
+++ b/ingest.go
@@ -702,7 +702,7 @@ func ingestTargetLevel(
 	// Check for overlap over the keys of L0 by iterating over the sublevels.
 	for subLevel := 0; subLevel < len(v.L0SublevelFiles); subLevel++ {
 		iter := newLevelIter(iterOps, cmp, nil /* split */, newIters,
-			v.L0Sublevels.Levels[subLevel].Iter(), manifest.Level(0), nil)
+			v.L0Sublevels.Levels[subLevel].Iter(), manifest.Level(0), internalIterOpts{})
 
 		var rangeDelIter keyspan.FragmentIterator
 		// Pass in a non-nil pointer to rangeDelIter so that levelIter.findFileGE
@@ -733,7 +733,7 @@ func ingestTargetLevel(
 	level := baseLevel
 	for ; level < numLevels; level++ {
 		levelIter := newLevelIter(iterOps, cmp, nil /* split */, newIters,
-			v.Levels[level].Iter(), manifest.Level(level), nil)
+			v.Levels[level].Iter(), manifest.Level(level), internalIterOpts{})
 		var rangeDelIter keyspan.FragmentIterator
 		// Pass in a non-nil pointer to rangeDelIter so that levelIter.findFileGE
 		// sets it up for the target file.
--- a/internal/cache/clockpro.go
+++ b/internal/cache/clockpro.go
@@ -785,14 +785,14 @@ func (c *Cache) Size() int64 {
 // manually managed. The caller MUST either add the value to the cache (via
 // Cache.Set), or release the value (via Cache.Free). Failure to do so will
 // result in a memory leak.
-func (c *Cache) Alloc(n int) *Value {
+func Alloc(n int) *Value {
 	return newValue(n)
 }
 
 // Free frees the specified value. The buffer associated with the value will
 // possibly be reused, making it invalid to use the buffer after calling
 // Free. Do not call Free on a value that has been added to the cache.
-func (c *Cache) Free(v *Value) {
+func Free(v *Value) {
 	if n := v.refs(); n > 1 {
 		panic(fmt.Sprintf("pebble: Value has been added to the cache: refs=%d", n))
 	}
--- a/internal/cache/clockpro_test.go
+++ b/internal/cache/clockpro_test.go
@@ -41,7 +41,7 @@ func TestCache(t *testing.T) {
 		var hit bool
 		h := cache.Get(1, base.FileNum(uint64(key)).DiskFileNum(), 0)
 		if v := h.Get(); v == nil {
-			value := cache.Alloc(1)
+			value := Alloc(1)
 			value.Buf()[0] = fields[0][0]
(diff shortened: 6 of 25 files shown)


jbowens:
@cockroachdb/storage — I'd appreciate a look this afternoon on this. I don't need a thorough look, but would like to identify any egregious correctness issues.

RaduBerinde (approved):
:lgtm_strong: 

> Reviewed 12 of 15 files at r1, 2 of 4 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2699#-:-N_GTdNiEt2Dho1r5ZKl:b-uwkowd)* status: all files reviewed, 1 unresolved discussion (waiting on @bananabrick and @jbowens)

___
*[`sstable/buffer_pool.go` line 77 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_GTYxRCScuNXpqGUo9:-N_GTYxRCScuNXpqGUoA:b-hh2do0) ([raw file](https://github.com/cockroachdb/pebble/blob/9017c6da53551404125bb0bc2325799081c7fc30/sstable/buffer_pool.go#L77)):*
> ```Go
> 	*p = BufferPool{
> 		cache: cache,
> 		pool:  pool,
> ```

[nit] should we do `pool[:0]` here so there's no possibility of misuse?

sumeerbhola (changes requested):
> Reviewed 7 of 15 files at r1, 2 of 4 files at r2, 2 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2699)* status: all files reviewed, 8 unresolved discussions (waiting on @bananabrick and @jbowens)

___
*[`compaction.go` line 2756 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_Jac6K0XuyJMxmvG7Q:-N_Jac6K0XuyJMxmvG7R:brqfg85) ([raw file](https://github.com/cockroachdb/pebble/blob/6c2252521f6933ad14043dd2057d318f4167d34c/compaction.go#L2756)):*
> ```Go
> 
> 	// Compactions use a pool of buffers to read blocks, avoiding polluting the
> 	// block cache with blocks that will not be read again.
> ```

Can you add a comment on why 5 is a good value for initial size.
In the worst-case with 2 level compactions do we have up to 2 data blocks, 2 value blocks, 2 rangedel blocks and 2 rangekey blocks allocated?
___
*[`sstable/block.go` line 405 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_JbJ_K89UCKnKJilQR:-N_JbJ_K89UCKnKJilQS:b-5k2prd) ([raw file](https://github.com/cockroachdb/pebble/blob/6c2252521f6933ad14043dd2057d318f4167d34c/sstable/block.go#L405)):*
> ```Go
> 	cached      []blockEntry
> 	cachedBuf   []byte
> 	cacheHandle cache.Handle
> ```

is `cacheHandle` unused, since we have the `cache.Handle` inside `bufferHandle`?
___
*[`sstable/buffer_pool.go` line 49 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_J_48I2Mc_s5IBvTAf:-N_J_48I2Mc_s5IBvTAg:bxts0ub) ([raw file](https://github.com/cockroachdb/pebble/blob/6c2252521f6933ad14043dd2057d318f4167d34c/sstable/buffer_pool.go#L49)):*
> ```Go
> 	// pool contains all the buffers held by the pool, including buffers that
> 	// are in-use.
> 	pool []allocedBuffer
> ```

IIUC, the invariant is that for every `pool[i]` the `pool[i].v` value is non-nil. If so, can you add a comment.
___
*[`sstable/buffer_pool.go` line 66 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_JYweJ8ksxty97dNL1:-N_JYweJ8ksxty97dNL2:b-bbo17j) ([raw file](https://github.com/cockroachdb/pebble/blob/6c2252521f6933ad14043dd2057d318f4167d34c/sstable/buffer_pool.go#L66)):*
> ```Go
> func (p *BufferPool) Init(cache *cache.Cache, initialSize int) {
> 	*p = BufferPool{
> 		cache: cache,
> ```

I don't see why `cache.Alloc` and `cache.Free` are methods on the `Cache` struct. They don't seem to reference anything in the cache. Could they be functions. Then we don't need `cache` as a member.
___
*[`sstable/buffer_pool.go` line 139 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_JXT-j1_bqYcSBWXwm:-N_JXT-j1_bqYcSBWXwn:b-rpm11n) ([raw file](https://github.com/cockroachdb/pebble/blob/6c2252521f6933ad14043dd2057d318f4167d34c/sstable/buffer_pool.go#L139)):*
> ```Go
> 	// is no longer in use and a future call to BufferPool.Alloc may reuse this
> 	// buffer.
> 	b.p.pool[b.i].b = nil
> ```

This code is not setting `b.p` to nil, which makes me worry that if `Release` is accidentally called again it may free a buffer that no longer belongs to `Buf`. I suppose we can't do better because `Release` is a value receiver. I am guessing making it a pointer receiver is not going to cause `Buf` to be heap allocated, in which case we could fix this.
___
*[`sstable/reader.go` line 3429 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_JczqAC7DG0fJIy7QD:-N_JczqB0-JkfWrhJ1Fr:b-rzrpq3) ([raw file](https://github.com/cockroachdb/pebble/blob/6c2252521f6933ad14043dd2057d318f4167d34c/sstable/reader.go#L3429)):*
> ```Go
> 			decompressed = cacheValueOrBuf{buf: bufferPool.Alloc(decodedLen)}
> 		} else {
> 			decompressed = cacheValueOrBuf{
> ```

so we need two bufs for each block, which means my previous comment may have undercounted by 2x.
___
*[`sstable/value_block.go` line 753 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_Jejq14lZAuUPITjpj:-N_Jejq14lZAuUPITjpk:bpxr5t4) ([raw file](https://github.com/cockroachdb/pebble/blob/6c2252521f6933ad14043dd2057d318f4167d34c/sstable/value_block.go#L753)):*
> ```Go
> ) (bufferHandle, error) {
> 	ctx = objiotracing.WithBlockType(ctx, objiotracing.ValueBlock)
> 	return bpwc.r.readBlock(ctx, h, nil, nil, stats, nil /* buffer pool */)
> ```

this case should be rare enough but if we needed to use the buffer pool we could since the `blockProviderWhenClosed` is not allowed to outlive the iterator tree, which means it cannot outlive the buffer pool.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2699)* status: all files reviewed, 7 unresolved discussions (waiting on @bananabrick and @sumeerbhola)

___
*[`compaction.go` line 2756 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_Jac6K0XuyJMxmvG7Q:-N_Kf2gIBUKmVvOEq0QA:b-gfrntd) ([raw file](https://github.com/cockroachdb/pebble/blob/6c2252521f6933ad14043dd2057d318f4167d34c/compaction.go#L2756)):*

Can you add a comment on why 5 is a good value for initial size.
In the worst-case with 2 level compactions do we have up to 2 data blocks, 2 value blocks, 2 rangedel blocks and 2 rangekey blocks allocated?
</blockquote></details>

In the worst case for a two-level compaction, we have up to 2 index blocks, 2 data blocks, 2 value blocks, _n_ rangedel blocks and _n_ rangekey blocks + 1 ephemeral block used to hold the intermediary compressed data block during a block load. I've bumped this up to 12 and written up a justification.
___
*[`sstable/block.go` line 405 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_JbJ_K89UCKnKJilQR:-N_Kjm1ABXhpiy3AyM-8:b-oz8o77) ([raw file](https://github.com/cockroachdb/pebble/blob/6c2252521f6933ad14043dd2057d318f4167d34c/sstable/block.go#L405)):*

is `cacheHandle` unused, since we have the `cache.Handle` inside `bufferHandle`?
</blockquote></details>

Good catch
___
*[`sstable/buffer_pool.go` line 49 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_J_48I2Mc_s5IBvTAf:-N_Kk8DB5T_X44AlvyGF:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/6c2252521f6933ad14043dd2057d318f4167d34c/sstable/buffer_pool.go#L49)):*

IIUC, the invariant is that for every `pool[i]` the `pool[i].v` value is non-nil. If so, can you add a comment.
</blockquote></details>

Done.
___
*[`sstable/buffer_pool.go` line 66 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_JYweJ8ksxty97dNL1:-N_KvhXg4eY-dRjKRSrT:b-i9s80l) ([raw file](https://github.com/cockroachdb/pebble/blob/6c2252521f6933ad14043dd2057d318f4167d34c/sstable/buffer_pool.go#L66)):*

I don't see why `cache.Alloc` and `cache.Free` are methods on the `Cache` struct. They don't seem to reference anything in the cache. Could they be functions. Then we don't need `cache` as a member.
</blockquote></details>

Very good point. This also lets us remove `cache` as a member from `cacheValueOrBuf`
___
*[`sstable/buffer_pool.go` line 139 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_JXT-j1_bqYcSBWXwm:-N_KkDij6vXGUyseYIwk:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/6c2252521f6933ad14043dd2057d318f4167d34c/sstable/buffer_pool.go#L139)):*

This code is not setting `b.p` to nil, which makes me worry that if `Release` is accidentally called again it may free a buffer that no longer belongs to `Buf`. I suppose we can't do better because `Release` is a value receiver. I am guessing making it a pointer receiver is not going to cause `Buf` to be heap allocated, in which case we could fix this.
</blockquote></details>

Done.
___
*[`sstable/buffer_pool.go` line 77 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_GTYxRCScuNXpqGUo9:-N_KvxRs6TU8tnOO9rZ0:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/9017c6da53551404125bb0bc2325799081c7fc30/sstable/buffer_pool.go#L77)):*

[nit] should we do `pool[:0]` here so there's no possibility of misuse?
</blockquote></details>

Done
___
*[`sstable/reader.go` line 3429 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_JczqAC7DG0fJIy7QD:-N_Kl1xjCWoklFYrTqQ6:b-eo7br8) ([raw file](https://github.com/cockroachdb/pebble/blob/6c2252521f6933ad14043dd2057d318f4167d34c/sstable/reader.go#L3429)):*

so we need two bufs for each block, which means my previous comment may have undercounted by 2x.
</blockquote></details>

It should be a constant +1 buf since only one of the buffers survive to the end
___
*[`sstable/value_block.go` line 753 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_Jejq14lZAuUPITjpj:-N_Km-U6DY-1Otp3HsKc:befsf87) ([raw file](https://github.com/cockroachdb/pebble/blob/6c2252521f6933ad14043dd2057d318f4167d34c/sstable/value_block.go#L753)):*

this case should be rare enough but if we needed to use the buffer pool we could since the `blockProviderWhenClosed` is not allowed to outlive the iterator tree, which means it cannot outlive the buffer pool.
</blockquote></details>

Left a TODO here

bananabrick (commented):
> they may be freed by a different thread inducing excessive TLB shootdowns

I never thought about this.

Why is it necessary that a free happens on a different thread to observe a TLB shootdown? If more than one thread is working with some memory, and any one of those threads calls free, we'd have the same problem right? Doesn't necessarily have to be a different thread than the one which allocated the memory.

> A compaction only requires a relatively small working set of buffers during its  scan across input sstables.

Do we really need manual memory allocation for these buffers?

> Reviewed 6 of 15 files at r1, 1 of 5 files at r3, 6 of 25 files at r4, 14 of 17 files at r5, 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2699#-:-N_Kgcy3EFW3mmGzmeyj:b9xyg64)* status: all files reviewed, 8 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/buffer_pool.go` line 99 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_KxROl8_Js9_MxwEPS:-N_KxROl8_Js9_MxwEPT:b-lqm74c) ([raw file](https://github.com/cockroachdb/pebble/blob/c5fc0298a28eeb04e0c5ff1c93a0ddb08afaccbe/sstable/buffer_pool.go#L99)):*
> ```Go
> 	for i := 0; i < len(p.pool); i++ {
> 		if p.pool[i].b == nil {
> 			if len(p.pool[i].v.Buf()) >= n {
> ```

nit: I think there's an invariant that `v.buf` is never resized so `cap(v.Buf) == len(v.Buf)`, when a `Value` is used by the BufferPool. Could we mention that in the `allocatedBuffer`?

jbowens (commented):
> Why is it necessary that a free happens on a different thread to observe a TLB shootdown? If more than one thread is working with some memory, and any one of those threads calls free, we'd have the same problem right? Doesn't necessarily have to be a different thread than the one which allocated the memory.

I believe that's true, but:
1. for these blocks read during compactions, the expectation is that only the one thread running the compaction goroutine is reading this memory, so only the free from another thread is causing any TLB shootdown.
2. jemalloc has a thread-local 'tcache' of cached objects. Beyond the TLB shootdown, the memory being freed from another thread defeats the utility of the thread-local tcache, making it less likely that subsequent allocations can be served by the tcache.

> Do we really need manual memory allocation for these buffers?

Not strictly, but we already need to use a manual accounting of their lifecycles, so keeping them off the Go heap and out of GC's view is just a freebie.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2699#-:-N_L69VdA1yuK6tB18lP:b-s9tw8l)* status: all files reviewed, 8 unresolved discussions (waiting on @bananabrick and @sumeerbhola)

___
*[`sstable/buffer_pool.go` line 99 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_KxROl8_Js9_MxwEPS:-N_LBJsG8wQqrsMgPY0j:b-7penrc) ([raw file](https://github.com/cockroachdb/pebble/blob/c5fc0298a28eeb04e0c5ff1c93a0ddb08afaccbe/sstable/buffer_pool.go#L99)):*

nit: I think there's an invariant that `v.buf` is never resized so `cap(v.Buf) == len(v.Buf)`, when a `Value` is used by the BufferPool. Could we mention that in the `allocatedBuffer`?
</blockquote></details>

while true, it's not an invariant we use in any way

bananabrick (approved):
:lgtm: 

> the expectation is that only the one thread running the compaction goroutine is reading this memory, so only the free from another thread is causing any TLB shootdown.

If the compaction thread is still reading the memory, then another thread wouldn't call free on it, because the refcount of the Value isn't 0, right?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2699#-:-N_LPjzj6-WH33nJWl0x:b-ebj4f6)* status: all files reviewed, 8 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/buffer_pool.go` line 99 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2699#-N_KxROl8_Js9_MxwEPS:-N_LQKlo1Y13bMZV8bfi:bb450by) ([raw file](https://github.com/cockroachdb/pebble/blob/c5fc0298a28eeb04e0c5ff1c93a0ddb08afaccbe/sstable/buffer_pool.go#L99)):*

while true, it's not an invariant we use in any way
</blockquote></details>

I figured we're relying on it here: `len(p.pool[i].v.Buf()) >= n`, we use `len` instead of `cap`, and it's correct to use `len` instead of `cap`, because `len == cap`.

sumeerbhola (approved):
:lgtm: 

> Reviewed 3 of 25 files at r4, 9 of 17 files at r5, 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2699#-:-N_QGbGnAxOTzsms0xTD:bg7igkv)* status: all files reviewed, 8 unresolved discussions (waiting on @jbowens)

jbowens (commented):
>If the compaction thread is still reading the memory, then another thread wouldn't call free on it, because the refcount of the Value isn't 0, right?

That's right, but even if the compaction thread is done with the block and has moved on to a subsequent block, the block may still be in its TLB. I believe even in this scenario, the processor suffers the cost of a 'TLB shootdown.'

TFTRs! 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2699#-:-N_Rb4XgEVBEQLZvl4Vi:b57583f)* status: all files reviewed, 8 unresolved discussions (waiting on @sumeerbhola)

PR #2700, opened 2023-06-29
crl-release-23.1: db: do not cache compaction block reads

jbowens:


RaduBerinde (commented):
Looks correct to me.  I left some comments but they don't need to be addressed before testing.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2700#-:-NZ7pl5r-sk50mi_YjKD:blvaeq7)* status: 0 of 15 files reviewed, 4 unresolved discussions (waiting on @jbowens)

___
*[`sstable/buffer_pool.go` line 36 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2700#-NZ7mUMP3a3eAJD-0GKY:-NZ7mUMP3a3eAJD-0GKZ:bsif2xn) ([raw file](https://github.com/cockroachdb/pebble/blob/d677777e36af6aa6f65c2a0eb6d8ba17918dfe64/sstable/buffer_pool.go#L36)):*
> ```Go
> }
> 
> // A BufferPool holds a pool of buffers for holding sstable blocks.
> ```

[nit] worth mentioning that it is intended for a small number of buffers since many operations are O(maxSize)
___
*[`sstable/buffer_pool.go` line 41 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2700#-NZ7m0SZ68JhejWB73qF:-NZ7m0SZ68JhejWB73qG:b-gpm04u) ([raw file](https://github.com/cockroachdb/pebble/blob/d677777e36af6aa6f65c2a0eb6d8ba17918dfe64/sstable/buffer_pool.go#L41)):*
> ```Go
> type BufferPool struct {
> 	cache *cache.Cache
> 	size  int
> ```

[nit] why not size the slice appropriately instead of using a separate size field?
___
*[`sstable/buffer_pool.go` line 67 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2700#-NZ7mp9tDkq6l_4wfvJO:-NZ7mp9tDkq6l_4wfvJP:b-ob5lg3) ([raw file](https://github.com/cockroachdb/pebble/blob/d677777e36af6aa6f65c2a0eb6d8ba17918dfe64/sstable/buffer_pool.go#L67)):*
> ```Go
> func (p *BufferPool) Alloc(n int) *Buf {
> 	for i := p.size - 1; i >= 0; i-- {
> 		if len(p.pool[i].b) >= n {
> ```

should this be `cap()`? we are shortening the buffer on alloc then releasing back but we never restore the original size
___
*[`sstable/buffer_pool.go` line 101 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2700#-NZ7n-Js64RCIFhI50gy:-NZ7n-Js64RCIFhI50gz:b-kdassv) ([raw file](https://github.com/cockroachdb/pebble/blob/d677777e36af6aa6f65c2a0eb6d8ba17918dfe64/sstable/buffer_pool.go#L101)):*
> ```Go
> 	b.b = b.v.Buf()
> 	if b.p.size == len(b.p.pool) {
> 		// The pool is full. Evict a random item.
> ```

What's the point of evicting rather than dropping this buffer? Maybe this buffer is already in our cache? If yes, then it seems like we want an LRU policy?

Or, maybe we should evict the smallest buffer so that in time we retain the more useful (larger) buffers?

sstable/reader.go
```diff
@@ -2839,56 +2892,80 @@ func (r *Reader) readBlock(
...
 	}
 
 	if transform != nil {
-		// Transforming blocks is rare, so the extra copy of the transformed data
-		// is not problematic.
-		var err error
-		b, err = transform(b)
+		// Transforming blocks is very rare, so the extra copy of the
+		// transformed data is not problematic.
+		tmpTransformed, err := transform(decompressed.get())
```
itsbilal, line 2941: We were aiming to leverage transformations more for progressive restore, so ideally we'd want to clean this up in the future or have the temp block also use a pool. Nothing to do at the moment, but maybe worth a TODO if we are merging this in.

itsbilal (commented):
LGTM as far as correctness goes, couldn't eyeball any bugs

PR #2701, opened 2023-06-29
record: don't allocate errors during WAL rotation

jbowens:
A call to cockroachdb/errors.New produces an error with a stack trace which can require allocating several times. Avoid allocating every call to LogWriter.Close.

### DIFF
--- a/record/log_writer.go
+++ b/record/log_writer.go
@@ -20,6 +20,7 @@ import (
 )
 
 var walSyncLabels = pprof.Labels("pebble", "wal-sync")
+var errClosedWriter = errors.New("pebble/record: closed LogWriter")
 
 type block struct {
 	// buf[:written] has already been filled with fragments. Updated atomically.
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2701#-:-NZ7qZwIDUBxO9jGzaJT:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @nvanbenschoten and @sumeerbhola)

nvanbenschoten (approved):

record/log_writer.go
```diff
@@ -20,6 +20,7 @@ import (
 )
 
 var walSyncLabels = pprof.Labels("pebble", "wal-sync")
+var errClosed = errors.New("pebble/record: closed LogWriter")
```
nvanbenschoten, line 23: nit: Depending on the namespace we're operating in (the `record` package), we may want to consider giving this a more descriptive name. Something like `errClosedLogWriter`.

jbowens:
TFTRs!

PR #2702, opened 2023-06-30
replay: pass correct replay bool flags on reruns

msbutler:
Previously, if the user passed --stream-logs or --ignore-checkpoint and with a --count>2, the replay tool would fail on the second run because the bool args would get incorrectly passed to syscall.Exec(). Specifically, for these flags, an additional `true` arg would get passed, which the cobra parser does not appreciate.

### DIFF
--- a/cmd/pebble/replay.go
+++ b/cmd/pebble/replay.go
@@ -89,10 +89,10 @@ func (c *replayConfig) args() (args []string) {
 		args = append(args, "--max-writes", fmt.Sprint(c.maxWritesMB))
 	}
 	if c.streamLogs {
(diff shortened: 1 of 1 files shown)


jbowens (approved):
Whoops, thanks :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2702#-:-NZCONu4Dh82sCBEpyNg:b-hgy52u)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sumeerbhola)

msbutler:
unrelated flake. merging.

PR #2706, opened 2023-07-04
sharedobjcat: allow addition/deletion of object in same batch

itsbilal:
On rare occasions, if we cancel a compaction at a point right after creating a new object, we'll end up deleting the object shortly after. Sometimes the two can happen with no explicit Sync() call in between, resulting in the object being created and deleted in the same batch which currently results in an error. This change allows for this case by moving around logic in the catalog batch application code.

### DIFF
--- a/objstorage/objstorageprovider/sharedobjcat/catalog.go
+++ b/objstorage/objstorageprovider/sharedobjcat/catalog.go
@@ -218,28 +218,36 @@ func (c *Catalog) ApplyBatch(b Batch) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	for _, n := range b.ve.DeletedObjects {
-		if _, exists := c.mu.objects[n]; !exists {
-			return errors.AssertionFailedf("deleting non-existent object %s", n)
-		}
-	}
+	// Add new objects before deleting any objects. This allows for cases where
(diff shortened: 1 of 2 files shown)


objstorage/objstorageprovider/sharedobjcat/catalog.go
```diff
@@ -218,28 +218,38 @@ func (c *Catalog) ApplyBatch(b Batch) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
+	addedObjects := make([]base.DiskFileNum, 0, len(b.ve.NewObjects))
```
RaduBerinde, line 221: Do we need this map? Can't we just go over NewObjects to undo the add?

itsbilal: Oopsie, fixed!

RaduBerinde (approved):
:lgtm:

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2706#-:-N_-t_x22FYmelUe3tQR:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal)

PR #2707, opened 2023-07-04
open: Don't enforce size match for virtual sstables on db open

itsbilal:
Currently, we verify that file sizes match between disk and file metadatas upon db open. This is an unhelpful exercise with virtual sstables as we expect to see mismatches there between the underlying backing file and the virtualized view of the sstable.

### DIFF
--- a/open.go
+++ b/open.go
@@ -1084,6 +1084,10 @@ func checkConsistency(v *manifest.Version, dirname string, objProvider objstorag
 				continue
 			}
 			dedup[backingState.DiskFileNum] = struct{}{}
+			// No need to enforce sizes for virtual sstables.
+			if f.Virtual {
+				continue
+			}
 			fileNum := backingState.DiskFileNum
 			fileSize := backingState.Size
 			meta, err := objProvider.Lookup(base.FileTypeTable, fileNum)


bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2707)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2707#-NZX_ZQy6sTc4fgfgt60):*
We're using the size of backing file to do the comparison. Maybe I'm misunderstanding what this is saying.

```
			fileNum := backingState.DiskFileNum
			fileSize := backingState.Size
			meta, err := objProvider.Lookup(base.FileTypeTable, fileNum)
			var size int64
			if err == nil {
				size, err = objProvider.Size(meta)
			}
			if err != nil {
				buf.WriteString("L%d: %s: %v\n")
				args = append(args, errors.Safe(level), errors.Safe(fileNum), err)
				continue
			}

			if size != int64(fileSize) {

```

itsbilal (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2707)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @bananabrick)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2707#-NZX_ZQy6sTc4fgfgt60):*

We're using the size of backing file to do the comparison. Maybe I'm misunderstanding what this is saying.

```
			fileNum := backingState.DiskFileNum
			fileSize := backingState.Size
			meta, err := objProvider.Lookup(base.FileTypeTable, fileNum)
			var size int64
			if err == nil {
				size, err = objProvider.Size(meta)
			}
			if err != nil {
				buf.WriteString("L%d: %s: %v\n")
				args = append(args, errors.Safe(level), errors.Safe(fileNum), err)
				continue
			}

			if size != int64(fileSize) {

```
</blockquote></details>

You're right, the change we need is specific for foreign sstables. I remembered I had already done it but turns out it's part of https://github.com/cockroachdb/pebble/pull/2672 which hasn't merged yet, so there's no need to redo it.

PR #2715, opened 2023-07-07
internal/manifest: delete L0Sublevels_LargeImport test

### LINKED ISSUE #2710: l0_sublevels: unskip TestL0Sublevels_LargeImportL0
This test has been skipped for years. It needs to be fixed and un-skipped, or removed if equivalent test coverage is already provided by other means.

itsbilal:
Previously we had skipped the L0Sublevels_LargeImport test as
it was not well-written and buggy. On closer inspection it does
not provide any net new test coverage and is just a sanity test
with loading a large manifest and picking some compactions in a row,
something that will happen with metamorphic test runs too. Plus,
the same manifest is already loaded in some benchmarks. Fixing
the test added 60+ seconds to the runtime of the whole test suite.

This change deletes this test.

Fixes #2710

### DIFF
--- a/internal/manifest/l0_sublevels_test.go
+++ b/internal/manifest/l0_sublevels_test.go
@@ -57,56 +57,6 @@ func readManifest(filename string) (*Version, error) {
 	return v, nil
 }
 
-func TestL0Sublevels_LargeImportL0(t *testing.T) {
-	t.Skip("#2710")
-	v, err := readManifest("testdata/MANIFEST_import")
-	require.NoError(t, err)
-
-	sublevels, err := NewL0Sublevels(&v.Levels[0],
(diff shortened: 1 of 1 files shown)


itsbilal:
Though it's debatable if this test is useful to begin with. It takes 60 seconds on its own (!), most of which is taken up by the L0 sublevels initialization and the read of the giant manifest. I don't see what bugs this would expose that won't also be exposed by the metamorphic test, also all it does is ensure compaction picking doesn't panic, there aren't any assertions here. My initial hunch would be to delete it. Thoughts?

jbowens:
> My initial hunch would be to delete it. Thoughts?

🚮

itsbilal:
🚮 it is

jbowens (approved):
lgtm

itsbilal:
TFTR!

PR #2718, opened 2023-07-10
db,sstable: bug fixes for keys with obsolete bit

### LINKED ISSUE #2705: sstable: missing keys when setting obsolete bits
```

		===== SEED =====
		1688175492068207000
		===== DIFF =====
		_meta/230630-213812.0682724060275/{standard-000,standard-026}
		@@ -237,11 +237,11 @@
 db.Merge("sxvlqq@1", "sle") // <nil> #236
 iter2.Prev("") // [false] pebble: unsupported reverse prefix iteration #237
 iter2.Next("bcvxkinko@2") // [invalid] cannot use limit with prefix iteration #238
 iter2.SeekLT("kbdzmck", "") // [true,"ipptgioar@2","nysdtymiefduy",["bcvxkinko","kbdzmck")=>{"@1"=""}*] <nil> #239
 db.SingleDelete("dijofczaiws@4", true /* maybeReplaceDelete */) // <nil> #240
-iter2.SeekGE("dzkhkzwn@2", "") // [true,"dzkhkzwn@2","omwnllrnaaekxcatmlw",["bcvxkinko","kbdzmck")=>{"@1"=""}] <nil> #241
+iter2.SeekGE("dzkhkzwn@2", "") // [true,"dzkhkzwn@2","",["bcvxkinko","kbdzmck")=>{"@1"=""}] <nil> #241
 iter2.NextPrefix() // [true,"hnsrnnu@1","vcmshucrmtu",["bcvxkinko","kbdzmck")=>{"@1"=""}] <nil> #242
 iter2.SeekPrefixGE("kbdzmck@3") // [true,"kbdzmck@3","xmhwzwzlize",<no range>*] <nil> #243
 iter2.Last() // [true,"wwzibulql@3","bqesltimayqruwsbzr",<no range>] <nil> #244
 db.DeleteRange("ifbca@1", "xujqxudpzwy@2") // <nil> #245
 iter2.First() // [true,"aognljopzy@3","qodk",<no range>] <nil> #246
@@ -257,11 +257,11 @@
 db.Delete("ifbca@1") // <nil> #256
 iter2.Next("") // [true,"awiaql@2","efsdxoodwkl",["awiaql","bcvxkinko")=>{""="v","@1"=""}] <nil> #257
 db.Get("awiaql@1") // [""] pebble: not found #258
 iter2.SeekLT("kbdzmck@1", "") // [true,"kbdzmck@3","xmhwzwzlize",<no range>*] <nil> #259
 db.Merge("spmxspivm@4", "nklocdzvc") // <nil> #260
-iter2.SeekPrefixGE("dzkhkzwn@2") // [true,"dzkhkzwn@2","omwnllrnaaekxcatmlw",["dzkhkzwn","dzkhkzwn\x00")=>{"@1"=""}*] <nil> #261
+iter2.SeekPrefixGE("dzkhkzwn@2") // [true,"dzkhkzwn@2","",["dzkhkzwn","dzkhkzwn\x00")=>{"@1"=""}*] <nil> #261
 iter2.SetBounds("", "kbdzmck") // <nil> #262
 iter2.SeekLT("kbdzmck", "") // [true,"ipptgioar@2","nysdtymiefduy",["bcvxkinko","kbdzmck")=>{"@1"=""}*] <nil> #263
 iter2.Next("") // [false] <nil> #264
 iter2.Prev("") // [true,"ipptgioar@2","nysdtymiefduy",["bcvxkinko","kbdzmck")=>{"@1"=""}*] <nil> #265
 iter2.SetBounds("spmxspivm@4", "sxvlqq@1") // <nil> #266
```

sumeerbhola:
- MERGE followed by DEL* previously caused the DEL* to be marked obsolete. This is incorrect since there may be a SET*/MERGE is a lower level that ought to be deleted, but will now get merged.

- A pointer to IterOptions was passed to tableCacheContainer.newIters, and IterOptions.PointKeyFilters was modified by the callee to append the obsolete key block property filter. Since the levelIter reuses the same IterOptions (levelIter.tableOpts), the obsolete key block property filter was getting used even for a file whose seqnum does not permit it. tableCacheShard.newIters is now changed to reuse the memory of the PointKeyFilters slice without altering the original slice.

- getIter was previously not setting the levelIterBoundaryContext, which prevented it from realizing whether a key returned should be ignored. This was not previously a bug since (a) there are no iterator bounds, so isSyntheticIterBoundsKey is never true (this is still true), (b) isIgnorableBoundaryKey was not true for some unclear reasons. (b) is now true when we exhaust the iterator by skipping blocks due to block property filters. This bug is fixed by setting and using levelIterBoundaryContext in getIter.

- compactionIter.snapshotPinned is no longer used to set forceObsolete, since it has a false positive for MERGE followed by older SET, when the latter is preserved due to a snapshot. Instead, we introduce forceObsoleteDueToRangeDel, which only handles obsolescence due to RANGEDEL.

Fixes #2705


sumeerbhola (commented):
I owe some unit tests.
stressmeta has been running for ~7 hours

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2718#-:-N_-n6d6BzW3SRRAv3du:b-7k9fmz)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @jbowens)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2718)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/writer.go` line 855 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2718#-N_0qHtd26LjFsxhJEZb:-N_0pCjx4_6_4PySmDKG:bmie39) ([raw file](https://github.com/cockroachdb/pebble/blob/cefb2ef753c58f38d4aeb5edb69349d9c1b830f0/sstable/writer.go#L855)):*
> ```Go
> 		cmpUser = w.compare(prevPointUserKey, key.UserKey)
> 	}
> 	if !(keyKind == InternalKeyKindSet || keyKind == InternalKeyKindSetWithDelete ||
> ```

[nit] I'd make this a `switch` with empty cases and the panic in the `default` case
___
*[`sstable/writer.go` line 860 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2718#-N_0pxGNDrK7ATXG6F1o:-N_0pxGNDrK7ATXG6F1p:bq8e46y) ([raw file](https://github.com/cockroachdb/pebble/blob/cefb2ef753c58f38d4aeb5edb69349d9c1b830f0/sstable/writer.go#L860)):*
> ```Go
> 		panic(errors.AssertionFailedf("unexpected key kind %s", keyKind.String()))
> 	}
> 	// If same user key, then the current key is obsolete if any of the
> ```

[nit] To make these cases more readable, I'd separate into two `if` blocks, each with its own comment (with each block setting `isObsolete = true`.

sumeerbhola (commented):
Added unit tests for everything other than the `getIter` bug, since it doesn't have a convenient datadriven test to modify. I can add a non-datadriven test if desired.
btw, all these bugs get triggered within ~1min of running stressmeta, so we have very solid test coverage for these already via the metamorphic test.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2718#-:-N_1YTdH5PlMxHpt-KYt:b-uhn3n5)* status: 0 of 15 files reviewed, 2 unresolved discussions (waiting on @jbowens)

sstable/writer.go
```diff
@@ -852,42 +852,42 @@ func (w *Writer) makeAddPointDecisionV3(
...
+	// must also delete all the lower seqnums for the same user key. C3 triggers
 	// due to a point delete and that deletes all the lower seqnums for the same
 	// user key.
 	//
 	// n = 1: This is the common case. It happens when the first key is not a
 	// MERGE, or the current key is some kind of point delete.
 	//
 	// n > 1: This is due to a sequence of MERGE keys, potentially followed by a
-	// single SET or SETWITHDELETE.
+	// single non-MERGE key.
```
sumeerbhola, line 891: > [nit] To make these cases more readable, I'd separate into two if blocks, each with its own comment (with each block setting isObsolete = true.

I left the comment in one place since I want the analysis related to n with all the conditions. But split the conditions to make them clearer.

sstable/writer.go
```diff
@@ -852,42 +852,42 @@ func (w *Writer) makeAddPointDecisionV3(
...
 		cmpUser = w.compare(prevPointUserKey, key.UserKey)
 	}
-	keyKindIsSetOrMerge := keyKind == InternalKeyKindSet || keyKind == InternalKeyKindSetWithDelete ||
-		keyKind == InternalKeyKindMerge
-	keyKindIsPointDelete := keyKind == InternalKeyKindDelete ||
-		keyKind == InternalKeyKindSingleDelete || keyKind == InternalKeyKindDeleteSized
-	if !keyKindIsSetOrMerge && !keyKindIsPointDelete {
+	if !(keyKind == InternalKeyKindSet || keyKind == InternalKeyKindSetWithDelete ||
+		keyKind == InternalKeyKindMerge || keyKind == InternalKeyKindDelete ||
+		keyKind == InternalKeyKindSingleDelete || keyKind == InternalKeyKindDeleteSized) {
```
sumeerbhola: > [nit] I'd make this a switch with empty cases and the panic in the default case

Done

sumeerbhola (commented):
TFTR!

(reviewable is not working, so responding via github)

jbowens (approved):
:lgtm_strong: 

> Reviewed 9 of 10 files at r1, 4 of 5 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2718#-:-N_4fp6W4BFlaCzQb4F1:b-uwkowd)* status: 14 of 15 files reviewed, 5 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`-- commits` line 14 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2718#-N_4fKsI4ytksIZ_qz9Y):*
Oof, tricky. Nice find

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2718)* status: 14 of 15 files reviewed, 6 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`table_cache.go` line 421 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2718#-N_4jVABEfoN3szZHGqj:-N_4jVABEfoN3szZHGqk:b-pj6e64) ([raw file](https://github.com/cockroachdb/pebble/blob/8b8a91a5e139db5fb49dc079717ff5831d0c8a93/table_cache.go#L421)):*
> ```Go
> 		// An alternative would be to have different slices for different sstable
> 		// iterators, but that requires more work to avoid allocations.
> 		hideObsoletePoints, opts.PointKeyFilters =
> ```

Does this really need to modify `opts.PointKeyFilters`? We could just store it in a local var, this function doesn't pass the full `opts` to anything..

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2718)* status: 11 of 15 files reviewed, 6 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`table_cache.go` line 421 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2718#-N_4jVABEfoN3szZHGqj:-N_6BYH74PufPtlTdtOR:b76ru3l) ([raw file](https://github.com/cockroachdb/pebble/blob/8b8a91a5e139db5fb49dc079717ff5831d0c8a93/table_cache.go#L421)):*

Does this really need to modify `opts.PointKeyFilters`? We could just store it in a local var, this function doesn't pass the full `opts` to anything..
</blockquote></details>

very good point!

Fixed.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2718#-:-N_AzCEc1VQXhkOXYtXS:bnfp4nl)* status: 11 of 15 files reviewed, 5 unresolved discussions (waiting on @jbowens and @sumeerbhola)

sumeerbhola:
TFTRs!

Issue #2676, opened 2023-06-26
ingest: Allow excises with flushable ingests


itsbilal:
Currently, an `IngestAndExcise` requires that we wait on any flushables that overlap with our ingestion sstables as well as any flushables that overlap with our excise span. This is because the flushable ingestion code does not support doing an excise operation. This reduces the performance benefit of doing a flushable ingest.

We should update the logic for flushable ingests to support excises to also become flushable ingests. This could be tricky, but should be possible by having the flushable iterators expose a `RANGEDEL` and `RANGEKEYDEL` at the ingestion sequence number from their iterators. Then when the flushable ingest is applied, we do the excise/virtualization of sstables.

Jira issue: PEBBLE-53

Epic CRDB-37583

PR #2719, opened 2023-07-10
ingest: don't assume flushable ingests are locally present

### LINKED ISSUE #2717: metamorphic: don't skip DB restarts in case shared storage is enabled
https://github.com/cockroachdb/pebble/pull/2694 added metamorphic testing of shared storage + the secondary cache. The test sometimes failed like this:

```
--- FAIL: TestMeta (12.59s)
    --- FAIL: TestMeta/execution (0.00s)
        --- FAIL: TestMeta/execution/random-008 (0.52s)
            meta.go:208: 
                ===== SEED =====
                1688998824586968860
                ===== ERR =====
                exit status 1
                ===== OUT =====
                Seed: 0
                dbRestartOp: open _meta/230710-142024.5874202807356/random-008/data/000059.sst: file does not exist
```

So that PR was adjusted to NOT do DB restarts in case shared storage is enabled.

We should fix whatever bug is causing the above test failure & then enable DB restarts (in case shared storage is enabled).

itsbilal:
Previously, the flushable ingest code assumed that all flushable ingested SSTs were locally present on vfs.FS, and the replayWAL code for that type of recovery bypassed the objProvider. This change updates replayWAL and ingestLoad to allow for the case where an ingested SST has already been copied/linked into shared storage.

Fixes #2717.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -79,8 +79,8 @@ func ingestValidateKey(opts *Options, key *InternalKey) error {
 	return nil
 }
 
-// ingestLoad1Shared loads the fileMetadata for one shared sstable. It also
-// sets the sequence numbers for a shared sstable.
+// ingestLoad1Shared loads the fileMetadata for one shared sstable owned or
+// shared by another node. It also sets the sequence numbers for a shared sstable.
 func ingestLoad1Shared(
 	opts *Options, sm SharedSSTMeta, fileNum base.DiskFileNum,
 ) (*fileMetadata, error) {
@@ -135,19 +135,15 @@ func ingestLoad1Shared(
 	return meta, nil
 }
 
+// ingestLoad1 creates the FileMetadata for one file. This file will be owned
+// by this store.
 func ingestLoad1(
-	opts *Options, fmv FormatMajorVersion, path string, cacheID uint64, fileNum base.DiskFileNum,
+	opts *Options,
+	fmv FormatMajorVersion,
+	readable objstorage.Readable,
+	cacheID uint64,
+	fileNum base.DiskFileNum,
 ) (*fileMetadata, error) {
-	f, err := opts.FS.Open(path)
-	if err != nil {
-		return nil, err
-	}
-
-	readable, err := sstable.NewSimpleReadable(f)
-	if err != nil {
-		return nil, err
-	}
-
 	cacheOpts := private.SSTableCacheOpts(cacheID, fileNum).(sstable.ReaderOption)
 	r, err := sstable.NewReader(readable, opts.MakeReaderOptions(), cacheOpts)
 	if err != nil {
@@ -302,7 +298,16 @@ func ingestLoad(
 	meta := make([]*fileMetadata, 0, len(paths))
 	newPaths := make([]string, 0, len(paths))
 	for i := range paths {
-		m, err := ingestLoad1(opts, fmv, paths[i], cacheID, pending[i])
+		f, err := opts.FS.Open(paths[i])
+		if err != nil {
+			return ingestLoadResult{}, err
+		}
+
+		readable, err := sstable.NewSimpleReadable(f)
+		if err != nil {
+			return ingestLoadResult{}, err
+		}
+		m, err := ingestLoad1(opts, fmv, readable, cacheID, pending[i])
 		if err != nil {
 			return ingestLoadResult{}, err
 		}
@@ -1107,7 +1112,6 @@ func (d *DB) ingest(
 	var mut *memTable
 	// asFlushable indicates whether the sstable was ingested as a flushable.
 	var asFlushable bool
-	var overlapWithExciseSpan bool
 	prepare := func(seqNum uint64) {
 		// Note that d.commit.mu is held by commitPipeline when calling prepare.
 
@@ -1159,7 +1163,6 @@ func (d *DB) ingest(
 					if mem == nil {
 						mem = m
 					}
-					overlapWithExciseSpan = true
 				}
 			}
 			err := iter.Close()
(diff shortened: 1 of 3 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2719)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @itsbilal, @joshimhoff, and @sumeerbhola)

___
*[`ingest.go` line 138 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2719#-N_0rLbx20MqmmZNN5pX:-N_0rLbx20MqmmZNN5pY:bful88b) ([raw file](https://github.com/cockroachdb/pebble/blob/e49cf76140ad70bee06acbb83d0a7fcf6286d088/ingest.go#L138)):*
> ```Go
> }
> 
> func ingestLoad1(
> ```

[nit] Consider making this take a Readable in all cases and moving the Open/NewSimpleReadable at the call site or into a new wrapper `openAndIngestLoad1()`
___
*[`ingest.go` line 138 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2719#-N_0tAs88jGKoQ82lLTB:-N_0tAs88jGKoQ82lLTC:b9o8vrd) ([raw file](https://github.com/cockroachdb/pebble/blob/e49cf76140ad70bee06acbb83d0a7fcf6286d088/ingest.go#L138)):*
> ```Go
> }
> 
> func ingestLoad1(
> ```

[nit] We also have `ingestLoad1Shared()` and it's confusing when would you use that vs using `ingestLoad1` with a shared Readable. Some more comments would help.
___
*[`open.go` line 814 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2719#-N_0sdTICdgJly02SjEK:-N_0sdTICdgJly02SjEL:b-qlljtd) ([raw file](https://github.com/cockroachdb/pebble/blob/168079ac22fdeee28feba15e3cf65a647cc1b664/open.go#L814)):*
> ```Go
> 
> 				var lr ingestLoadResult
> 				lr, err = ingestLoad(d.opts, d.FormatMajorVersion(), paths, nil, d.cacheID, fileNums)
> ```

It feels strange that the only remaining callsite for `ingestLoad` is for WAL replay, seems like we should be using the same codepath?
___
*[`open.go` line 815 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2719#-N_0sjU0B_ZpK339uUpw:-N_0sjU0B_ZpK339uUpx:b-8vabof) ([raw file](https://github.com/cockroachdb/pebble/blob/e49cf76140ad70bee06acbb83d0a7fcf6286d088/open.go#L815)):*
> ```Go
> 					if objMeta, err := d.objProvider.Lookup(fileTypeTable, n); err != nil {
> 						return nil, 0, errors.Wrap(err, "pebble: error when looking up ingested SSTs")
> 					} else if objMeta.IsShared() {
> ```

[nit] no need for else

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2719#-:-N_4yYkR0iEdemqLr-Fe:b-814g8k)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @joshimhoff, @RaduBerinde, and @sumeerbhola)

___
*[`ingest.go` line 138 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2719#-N_0rLbx20MqmmZNN5pX:-N_4xZunC2SMDrmb68bO:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e49cf76140ad70bee06acbb83d0a7fcf6286d088/ingest.go#L138)):*

[nit] Consider making this take a Readable in all cases and moving the Open/NewSimpleReadable at the call site or into a new wrapper `openAndIngestLoad1()`
</blockquote></details>

Done.
___
*[`ingest.go` line 138 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2719#-N_0tAs88jGKoQ82lLTB:-N_4x_fi1IM4zFneO5jv:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e49cf76140ad70bee06acbb83d0a7fcf6286d088/ingest.go#L138)):*

[nit] We also have `ingestLoad1Shared()` and it's confusing when would you use that vs using `ingestLoad1` with a shared Readable. Some more comments would help.
</blockquote></details>

Done.
___
*[`open.go` line 814 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2719#-N_0sdTICdgJly02SjEK:-N_4xqAl8VJCVOCjZlKT:btgkqhi) ([raw file](https://github.com/cockroachdb/pebble/blob/168079ac22fdeee28feba15e3cf65a647cc1b664/open.go#L814)):*

It feels strange that the only remaining callsite for `ingestLoad` is for WAL replay, seems like we should be using the same codepath?
</blockquote></details>

It is definitely a little weird, and comes from the fact that we reconstruct the FileMetadata for the ingested sstable during WAL replay (which is the job of ingestLoad*, to construct the fileMetadata) as the WAL entry only contains the sstable's FileNum. We'll need to revisit a lot of this when https://github.com/cockroachdb/pebble/issues/2676 is done.
___
*[`open.go` line 815 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2719#-N_0sjU0B_ZpK339uUpw:-N_4xlvF6Dg14-t4IVdo:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e49cf76140ad70bee06acbb83d0a7fcf6286d088/open.go#L815)):*

[nit] no need for else
</blockquote></details>

Done.

PR #2720, opened 2023-07-10
cmd: enable benchmarking shared storage including the secondary cache

joshimhoff:
**cmd: enable benchmarking shared storage including the secondary cache**

This commit makes it possible to run commands like 'pebble bench ycsb' against shared storage & the secondary cache. To do this, I have introduced a version of shared.Storage that is backed by local disk, as opposed to memory or blob storage. This enables running too-large-for-memory benchmarks from a single machine without access to blob storage. One reason this is convenient is that today the code to communicate with blob storage is in the cockroachdb repo, not the pebble repo.

The concrete benchmarking plan I have in mind is worth mentioning as motivation. I will measure the maximum ycsb thruput of shared storage without a secondary cache and shared storage with a secondary cache. Since shared objects will be stored on a medium as fast as the secondary cache (both will be on local disk), we expect better thruput without the secondary cache. With that said, we hope the thruput difference in not large, as for the secondary cache to be effective, it must make efficient use of the local disk it is backed by.

### DIFF
--- a/cmd/pebble/db.go
+++ b/cmd/pebble/db.go
@@ -10,6 +10,8 @@ import (
 	"github.com/cockroachdb/pebble"
 	"github.com/cockroachdb/pebble/bloom"
 	"github.com/cockroachdb/pebble/internal/bytealloc"
+	"github.com/cockroachdb/pebble/objstorage/shared"
+	"github.com/cockroachdb/pebble/vfs"
 )
 
 // DB specifies the minimal interfaces that need to be implemented to support
@@ -98,9 +100,25 @@ func newPebbleDB(dir string) DB {
 		opts.EventListener.WALDeleted = nil
 	}
 
+	if pathToLocalSharedStorage != "" {
+		opts.Experimental.SharedStorage = shared.MakeSimpleFactory(map[shared.Locator]shared.Storage{
+			// Store all shared objects on local disk, for convenience.
+			"": shared.NewLocalFS(pathToLocalSharedStorage, vfs.Default),
+		})
+		opts.Experimental.CreateOnShared = true
+		if secondaryCacheSize != 0 {
+			opts.Experimental.SecondaryCacheSize = secondaryCacheSize
+		}
+	}
+
 	p, err := pebble.Open(dir, opts)
 	if err != nil {
 		log.Fatal(err)
+	}
+	if pathToLocalSharedStorage != "" {
+		if err := p.SetCreatorID(1); err != nil {
+			log.Fatal(err)
+		}
 	}
 	return pebbleDB{
 		d:       p,
--- a/cmd/pebble/main.go
+++ b/cmd/pebble/main.go
@@ -15,15 +15,19 @@ import (
 )
 
 var (
-	cacheSize       int64
-	concurrency     int
-	disableWAL      bool
-	duration        time.Duration
-	maxSize         uint64
-	maxOpsPerSec    = newRateFlag("")
-	verbose         bool
-	waitCompactions bool
-	wipe            bool
+	cacheSize                int64
+	concurrency              int
+	disableWAL               bool
+	duration                 time.Duration
+	maxSize                  uint64
+	maxOpsPerSec             = newRateFlag("")
+	verbose                  bool
+	waitCompactions          bool
+	wipe                     bool
+	pathToLocalSharedStorage string
+	// If zero, or if !sharedStorageEnabled, secondary cache is
+	// not used.
+	secondaryCacheSize int64
 )
 
 func main() {
@@ -59,6 +63,10 @@ func main() {
 	for _, cmd := range []*cobra.Command{replayCmd, scanCmd, syncCmd, tombstoneCmd, writeBenchCmd, ycsbCmd} {
 		cmd.Flags().BoolVarP(
 			&verbose, "verbose", "v", false, "enable verbose event logging")
+		cmd.Flags().StringVar(
+			&pathToLocalSharedStorage, "shared-storage", "", "path to local shared storage (empty for no shared storage)")
+		cmd.Flags().Int64Var(
+			&secondaryCacheSize, "secondary-cache", 0, "secondary cache size in bytes")
 	}
 	for _, cmd := range []*cobra.Command{scanCmd, syncCmd, tombstoneCmd, ycsbCmd} {
 		cmd.Flags().Int64Var(
--- a/objstorage/shared/localfs.go
+++ b/objstorage/shared/localfs.go
@@ -1,0 +1,118 @@
+// Copyright 2023 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package shared
+
+import (
+	"context"
+	"io"
+	"os"
+	"path"
+
+	"github.com/cockroachdb/pebble/vfs"
+)
+
+// NewLocalFS returns a vfs-backed implementation of the shared.Storage
+// interface (for testing). All objects will be stored at the directory
+// dirname.
+func NewLocalFS(dirname string, fs vfs.FS) Storage {
+	store := &localFSStore{
+		dirname: dirname,
+		vfs:     fs,
+	}
+	return store
+}
+
+// localFSStore is a vfs-backed implementation of the shared.Storage
+// interface (for testing).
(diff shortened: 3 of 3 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2720)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @jbowens and @joshimhoff)

___
*[`cmd/pebble/main.go` line 67 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2720#-N_0tggH1iq5k6Kl8ue2:-N_0tggH1iq5k6Kl8ue3:b9536q1) ([raw file](https://github.com/cockroachdb/pebble/blob/511d21dbf61613e0d010da2275100e92b15d2351/cmd/pebble/main.go#L67)):*
> ```Go
> 			&verbose, "verbose", "v", false, "enable verbose event logging")
> 		cmd.Flags().BoolVar(
> 			&sharedStorageEnabled, "shared-storage", false, "enable shared storage (storage backing is local disk)")
> ```

I think instead of a flag, we should pass the path where we want the fake shared storage to be. We'll probably want to test with separate disks
___
*[`objstorage/shared/localfs.go` line 41 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2720#-N_0tpdt63nYfQRhsItf:-N_0tpdt63nYfQRhsItg:b-o5f2ar) ([raw file](https://github.com/cockroachdb/pebble/blob/511d21dbf61613e0d010da2275100e92b15d2351/objstorage/shared/localfs.go#L41)):*
> ```Go
> }
> 
> func (s *localFSStore) ReadObject(
> ```

[nit] add `// XXX is part of the shared.Storage interface` to the methods
___
*[`objstorage/shared/localfs.go` line 63 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2720#-N_0uQHVBWbzCqFrkseA:-N_0uQHVBWbzCqFrkseB:b77f3f6) ([raw file](https://github.com/cockroachdb/pebble/blob/511d21dbf61613e0d010da2275100e92b15d2351/objstorage/shared/localfs.go#L63)):*
`ReaderAt` has this strange case documented (https://pkg.go.dev/io#ReaderAt):
> If the n = len(p) bytes returned by ReadAt are at the end of the input source, ReadAt may return either err == EOF or err == nil.

I'd check for the `err == EOF && n == len(p)` and return `nil` in that case, just to be safe.
___
*[`objstorage/shared/localfs.go` line 91 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2720#-N_0u5Sd4H0jj3_LmD1s:-N_0u5Sd4H0jj3_LmD1t:b-b6nuqe) ([raw file](https://github.com/cockroachdb/pebble/blob/511d21dbf61613e0d010da2275100e92b15d2351/objstorage/shared/localfs.go#L91)):*
> ```Go
> 
> func (s *localFSStore) Size(objName string) (int64, error) {
> 	f, err := s.vfs.Open(path.Join(s.dirname, objName))
> ```

We need to close `f`

joshimhoff (commented):
PTAL!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2720#-:-N_43bYzBjv3Ng2L_E97:b-83330z)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`cmd/pebble/main.go` line 67 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2720#-N_0tggH1iq5k6Kl8ue2:-N_41i7P3kxg5Ky0tYX1:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/511d21dbf61613e0d010da2275100e92b15d2351/cmd/pebble/main.go#L67)):*

I think instead of a flag, we should pass the path where we want the fake shared storage to be. We'll probably want to test with separate disks
</blockquote></details>

Done.
___
*[`objstorage/shared/localfs.go` line 41 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2720#-N_0tpdt63nYfQRhsItf:-N_42ADV-eNuPno9keIo:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/511d21dbf61613e0d010da2275100e92b15d2351/objstorage/shared/localfs.go#L41)):*

[nit] add `// XXX is part of the shared.Storage interface` to the methods
</blockquote></details>

Done.
___
*[`objstorage/shared/localfs.go` line 63 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2720#-N_0uQHVBWbzCqFrkseA:-N_42s_t67RN_GQXgmFm:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/511d21dbf61613e0d010da2275100e92b15d2351/objstorage/shared/localfs.go#L63)):*

`ReaderAt` has this strange case documented (https://pkg.go.dev/io#ReaderAt):
> If the n = len(p) bytes returned by ReadAt are at the end of the input source, ReadAt may return either err == EOF or err == nil.

I'd check for the `err == EOF && n == len(p)` and return `nil` in that case, just to be safe.
</blockquote></details>

Done.
___
*[`objstorage/shared/localfs.go` line 91 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2720#-N_0u5Sd4H0jj3_LmD1s:-N_42VGS1Z6nO_7CN4IE:b3mezzd) ([raw file](https://github.com/cockroachdb/pebble/blob/511d21dbf61613e0d010da2275100e92b15d2351/objstorage/shared/localfs.go#L91)):*

We need to close `f`
</blockquote></details>

Thanks for catching! Fixed.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2720)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`objstorage/shared/localfs.go` line 41 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2720#-N_0tpdt63nYfQRhsItf:-N_4i_-0-YN8OWvyierr:b-a9odzn) ([raw file](https://github.com/cockroachdb/pebble/blob/511d21dbf61613e0d010da2275100e92b15d2351/objstorage/shared/localfs.go#L41)):*

Done.
</blockquote></details>

I don't think you pushed all your changes.

joshimhoff (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2720)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`objstorage/shared/localfs.go` line 41 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2720#-N_0tpdt63nYfQRhsItf:-N_4k3jxELFsXreVeHSO:b-lfjkee) ([raw file](https://github.com/cockroachdb/pebble/blob/511d21dbf61613e0d010da2275100e92b15d2351/objstorage/shared/localfs.go#L41)):*

I don't think you pushed all your changes.
</blockquote></details>

Oh? Can you be more specific? I see "// XXX is part of..." messages in `localfs.go` now: https://github.com/cockroachdb/pebble/pull/2720/files#diff-6561f572e858531d59f53889ae0aac6f1a53edbbcd8e70ffae57ba2ebc0fd8ebR42

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2720#-:-N_4uIuCB0PSCVliexdp:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @jbowens)

___
*[`objstorage/shared/localfs.go` line 41 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2720#-N_0tpdt63nYfQRhsItf:-N_4uAOhC0cv4Mn47fFz:bm1ed4) ([raw file](https://github.com/cockroachdb/pebble/blob/511d21dbf61613e0d010da2275100e92b15d2351/objstorage/shared/localfs.go#L41)):*

Oh? Can you be more specific? I see "// XXX is part of..." messages in `localfs.go` now: https://github.com/cockroachdb/pebble/pull/2720/files#diff-6561f572e858531d59f53889ae0aac6f1a53edbbcd8e70ffae57ba2ebc0fd8ebR42
</blockquote></details>

It shows up now, not sure what was the issue.

joshimhoff:
TTFR!

PR #2724, opened 2023-07-11
db: smoother thresholds for the delete pacer

RaduBerinde:
@jbowens this is a proposal to improve the behavior of the delete pacer heuristics.

----
The delete pacer has two heuristics to disable pacing altogether: one is when the free space drops below 16GB and one when the ratio of obsolete to live bytes goes over 20%.

Disabling pacing altogether can lead to a very sudden burst of large deletions, causing large latency spikes. In particular the obsolete bytes heuristic can be triggered when a large table is deleted.

In this change we make the transition smoother, using the idea that we want to get back to the desired thresholds within a given timeframe (10 seconds for the free bytes heuristic, 5 minutes for the obsolete bytes heuristic). Instead of effectively increasing the pacing rate to infinity, we increase it based on how many bytes are exceeding the threshold.

Note that the previous status quo corresponds to setting infinitesimal timeframes (e.g. 1ns).

### DIFF
--- a/pacer.go
+++ b/pacer.go
@@ -25,12 +25,16 @@ type deletionPacerInfo struct {
 // mechanism helps mitigate that.
 type deletionPacer struct {
 	// If there are less than freeSpaceThreshold bytes of free space on
-	// disk, do not pace deletions at all.
+	// disk, increase the pace of deletions such that we delete enough bytes to
+	// get back to the threshold within the freeSpaceTimeframe.
 	freeSpaceThreshold uint64
+	freeSpaceTimeframe time.Duration
 
 	// If the ratio of obsolete bytes to live bytes is greater than
-	// obsoleteBytesMaxRatio, do not pace deletions at all.
-	obsoleteBytesMaxRatio float64
+	// obsoleteBytesMaxRatio, increase the pace of deletions such that we delete
+	// enough bytes to get back to the threshold within the obsoleteBytesTimeframe.
+	obsoleteBytesMaxRatio  float64
+	obsoleteBytesTimeframe time.Duration
 
 	mu struct {
 		sync.Mutex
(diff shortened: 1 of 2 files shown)


RaduBerinde:
Some back of the envelope examples for the chosen parameters:
 - if (as in https://github.com/cockroachdb/cockroach/issues/106258#issuecomment-1623876400) we suddenly delete 200GB, the pacing rate will be set at ~700MB/s.
 - if free space dips to 10GB, the pacing rate will be set at 600MB/s
 - if free space dips to 1GB, the pacing rate will be set at 1.5GB/s

RaduBerinde:
If we want to be more conservative (i.e. closer to the status quo), we can reduce the timeframes, e.g. to 1s for free space and 1 minute. The examples above become:
 - delete 200GB table: 3.5GB/s
 - free space 10GB: 6GB/s
 - free space 1GB: 15GB/s

jbowens (approved):
I like this approach.

An adaption that came to mind is to use the history-based rate as a baseline expected throughput, and whatever additional rate we determine through the free-space or obsolete-bytes heuristics is added on top of that. The idea being that as long as we're still removing files from the LSM, we'll never reach the free-space or obsolete-bytes thresholds, because those rates (in particular the obsolete-bytes rate) don't account for additional incoming deletions. Maybe this isn't the right way of thinking about it, because a sudden deletion of many files will be double counted within both the obsolete-bytes heuristic and the history-based heuristic.

I think what you have makes sense.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2724#-:-N_FVMFKC0nVXEmfC7_N:b-datvfi)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde (commented):
Updated.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2724#-:-N_Ff2sA-e6KF_S1UrvQ:b-bw8e14)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jbowens)

jbowens (approved):
:lgtm:

Nice

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2724#-:-N_FoG1_E1p2RUwEjKUo:b-jxf22a)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

PR #2725, opened 2023-07-11
crl-release-23.1: cache: use atomic.Int32 wrappers, fix potential bug

RaduBerinde:

This change switches to atomic.Int32 wrapper and uses `Store` on init.


jbowens (approved):
lgtm

PR #2726, opened 2023-07-11
crl-release-22.2: cache: use atomic store on init

RaduBerinde:

This change switches to using `StoreInt32` on init.


RaduBerinde:
This is a backport of #2645, though we don't switch to using the wrappers because they are not available in go 1.18 (which is used for 22.2).

jbowens (approved):

PR #2734, opened 2023-07-12
vfs: Deflake TestDiskHealthChecking_File*

### LINKED ISSUE #1718: vfs: flaky test TestDiskHealthChecking_Filesystem
See: https://github.com/cockroachdb/pebble/runs/6563748979?check_suite_focus=true

```console
--- FAIL: TestDiskHealthChecking_Filesystem (0.41s)
    --- FAIL: TestDiskHealthChecking_Filesystem/remove-all (0.05s)
        assertion_compare.go:211: 
            	Error Trace:	disk_health_test.go:276
            	Error:      	"0" is not greater than "0"
            	Test:       	TestDiskHealthChecking_Filesystem/remove-all
            	Messages:   	[]
```

itsbilal:
Previously we were relying on sleeps and timing-based ways of
synchronization between observing a stall in the disk health checking
goroutine and confirming for it in the test code itself. This change
adds a more direct synchronization between the two events through
the use of channels, to deflake both tests. Furthermore,
the TestDiskHealthChecking_Filesystem_Close test was previously
doing a relatively thread-unsafe use of a map, which increased
the chances of a flake.

Also closes diskHealthCheckingFiles created in some Create operations
to prevent goroutine leaks.

Removes some tests that tested for false positives on disk stalls,
even though scheduler delays can also cause perceived disk
stalls at the small thresholds that were used.

Also selectively skips two tests on windows that almost exclusively
just flaked there.

Fixes #1718

### DIFF
--- a/vfs/disk_health_test.go
+++ b/vfs/disk_health_test.go
@@ -8,6 +8,8 @@ import (
 	"io"
 	"math"
 	"os"
+	"runtime"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
@@ -216,6 +218,10 @@ var _ FS = &mockFS{}
 func TestDiskHealthChecking_File(t *testing.T) {
 	oldTickInterval := defaultTickInterval
 	defaultTickInterval = time.Millisecond
+	if runtime.GOOS == "windows" {
+		t.Skipf("skipped on windows due to unreliable runtimes")
+	}
+
 	defer func() { defaultTickInterval = oldTickInterval }()
 
 	const (
@@ -229,43 +235,23 @@ func TestDiskHealthChecking_File(t *testing.T) {
 		writeDuration    time.Duration
 		fn               func(f File)
 		createWriteDelta time.Duration
-		expectStall      bool
 	}{
-		// No false negatives.
 		{
 			op:            OpTypeWrite,
 			writeSize:     5 * writeSizePrecision, // five KB
 			writeDuration: 100 * time.Millisecond,
 			fn:            func(f File) { f.Write(fiveKB) },
-			expectStall:   true,
 		},
 		{
 			op:            OpTypeSync,
 			writeSize:     0,
 			writeDuration: 100 * time.Millisecond,
 			fn:            func(f File) { f.Sync() },
-			expectStall:   true,
-		},
-		// No false positives.
-		{
-			op:               OpTypeWrite,
-			writeSize:        5,
-			writeDuration:    25 * time.Millisecond,
-			fn:               func(f File) { f.Write([]byte("uh oh")) },
-			createWriteDelta: 100 * time.Millisecond,
-			expectStall:      false,
-		},
-		{
-			op:            OpTypeSync,
-			writeSize:     0,
-			writeDuration: 25 * time.Millisecond,
-			fn:            func(f File) { f.Sync() },
-			expectStall:   false,
 		},
 	}
 	for _, tc := range testCases {
 		t.Run(tc.op.String(), func(t *testing.T) {
-			diskSlow := make(chan DiskSlowInfo, 1)
+			diskSlow := make(chan DiskSlowInfo, 3)
 			mockFS := &mockFS{create: func(name string) (File, error) {
 				return mockFile{syncAndWriteDuration: tc.writeDuration}, nil
 			}}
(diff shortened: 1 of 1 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2734)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`vfs/disk_health_test.go` line 518 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2734#-N_9vA4TARM4S3xAZd2q:-N_9vA4TARM4S3xAZd2r:byzpgvt) ([raw file](https://github.com/cockroachdb/pebble/blob/8253aea0546dfa9c69ee060c59ff32525a946e2a/vfs/disk_health_test.go#L518)):*
> ```Go
> 			select {
> 			case <-onStall:
> 			case <-time.After(10 * stallThreshold):
> ```

In rare cases (or stress/race runs and such) we can get arbitrary delays on the order of this value (100ms). I'd add something like 10 seconds to it just for good measure.
___
*[`vfs/disk_health_test.go` line 565 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2734#-N_9vTJ880YSpcfPk2IT:-N_9vTJ880YSpcfPk2IU:b-7ropln) ([raw file](https://github.com/cockroachdb/pebble/blob/8253aea0546dfa9c69ee060c59ff32525a946e2a/vfs/disk_health_test.go#L565)):*
> ```Go
> 		case stalledPath := <-stalled:
> 			require.Equal(t, filename, stalledPath)
> 		case <-time.After(10 * stallThreshold):
> ```

ditto

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2734#-:-N_AIHGJ5PglXE3C82Kw:b-814g8k)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sumeerbhola)

___
*[`vfs/disk_health_test.go` line 518 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2734#-N_9vA4TARM4S3xAZd2q:-N_AII9n8E8VPVl7gtTS:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8253aea0546dfa9c69ee060c59ff32525a946e2a/vfs/disk_health_test.go#L518)):*

In rare cases (or stress/race runs and such) we can get arbitrary delays on the order of this value (100ms). I'd add something like 10 seconds to it just for good measure.
</blockquote></details>

Done.
___
*[`vfs/disk_health_test.go` line 565 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2734#-N_9vTJ880YSpcfPk2IT:-N_AIIfL9WhfZYJlqw1-:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8253aea0546dfa9c69ee060c59ff32525a946e2a/vfs/disk_health_test.go#L565)):*

ditto
</blockquote></details>

Done.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2734)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @sumeerbhola)

___
*[`vfs/disk_health_test.go` line 299 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2734#-N_AIyv9EZRu-Xtm7-Zj:-N_AIyv9EZRu-Xtm7-Zk:bjiiw9h) ([raw file](https://github.com/cockroachdb/pebble/blob/8253aea0546dfa9c69ee060c59ff32525a946e2a/vfs/disk_health_test.go#L299)):*
> ```Go
> 				}
> 			} else { // no false positives
> 				select {
> ```

This seems robust to me.. if the test is slow, we might get a false pass. But we should never get a false fail

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2734)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sumeerbhola)

___
*[`vfs/disk_health_test.go` line 299 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2734#-N_AIyv9EZRu-Xtm7-Zj:-N_ALdAG5Kz5vXdP__zr:bfhpk0d) ([raw file](https://github.com/cockroachdb/pebble/blob/8253aea0546dfa9c69ee060c59ff32525a946e2a/vfs/disk_health_test.go#L299)):*

That's what I thought at first. Brought it back for now with far more generous timeouts, let's see what gives :crossed_fingers: if it flakes again I swing the hammer at it.
</blockquote></details>

A 50ms scheduling delay can happen, so there would still be spurious stalls. We'd need to up the `slowThreshold` to a few seconds. I'm in favor of axing them though.

itsbilal (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2734)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sumeerbhola)

___
*[`vfs/disk_health_test.go` line 299 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2734#-N_AIyv9EZRu-Xtm7-Zj:-N_AQfwM0tqyVOo_popp:b-8ao94d) ([raw file](https://github.com/cockroachdb/pebble/blob/8253aea0546dfa9c69ee060c59ff32525a946e2a/vfs/disk_health_test.go#L299)):*

A 50ms scheduling delay can happen, so there would still be spurious stalls. We'd need to up the `slowThreshold` to a few seconds. I'm in favor of axing them though.
</blockquote></details>

Axed.

PR #2736, opened 2023-07-12
objstorage: close the secondary cache

joshimhoff:
**objstorage: close the secondary cache**

This commit closes the secondary cache from provider.Close. This way, calling db.Close closes the secondary cache. Since the secondary caches does writes asynchronously with respect to calls to ReadAt, this is required, to avoid metamorphic test failures. The metamorphic tests simulate server restarts via calling db.Close & then pebble.Open.

Thank you to Radu for finding this bug!

### DIFF
--- a/objstorage/objstorageprovider/provider.go
+++ b/objstorage/objstorageprovider/provider.go
@@ -184,9 +184,10 @@ func open(settings Settings) (p *provider, _ error) {
 
 // Close is part of the objstorage.Provider interface.
 func (p *provider) Close() error {
-	var err error
 	if p.fsDir != nil {
-		err = p.fsDir.Close()
+		if err := p.fsDir.Close(); err != nil {
+			return err
+		}
 		p.fsDir = nil
 	}
 	if objiotracing.Enabled {
@@ -195,7 +196,13 @@ func (p *provider) Close() error {
 			p.tracer = nil
 		}
 	}
-	return err
+	if p.shared.cache != nil {
+		if err := p.shared.cache.Close(); err != nil {
+			return err
+		}
+		p.shared.cache = nil
+	}
+	return nil
 }
 
 // OpenForReading opens an existing object.


RaduBerinde (approved):
`Close` calls won't get retried, we have to do our best to clean up even in case of errors. We don't want a half-closed object that's not usable and yet it still holds up associated resources (like goroutines). There is a `firstError` helper you can use.


I wonder if there's an easy way to open the cache files in an "exclusive" mode. That would have made this bug manifest much nicer (error during the second cache open).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2736#-:-N_A0uuV1sNLeUgTdzXa:bukcds0)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens and @joshimhoff)

___
*[`objstorage/objstorageprovider/provider.go` line 199 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2736#-N_A0iORBEhMBX7hODMm:-N_A0iORBEhMBX7hODMn:b-xvh5az) ([raw file](https://github.com/cockroachdb/pebble/blob/ca09b134b44e5e8366da06a8215dd16491e24061/objstorage/objstorageprovider/provider.go#L199)):*
> ```Go
> 		}
> 	}
> 	if p.shared.cache != nil {
> ```

[nit] we should put this in a `sharedClose()` so there is symmetry with the open code. And we should call that first, since `sharedInit` is called last

joshimhoff:
Will make the suggested changes. Thanks for explaining the reason for the pattern followed in `Close` methods.

joshimhoff (commented):
Radu, can you take one more look, before I merge? I adjusted the code to close the catalog also.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2736#-:-N_ARa0EED0gnKCvC-PS:b-dfs9b1)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jbowens)

___
*[`objstorage/objstorageprovider/provider.go` line 199 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2736#-N_A0iORBEhMBX7hODMm:-N_ARWvzDdx9wa8bmsk1:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/ca09b134b44e5e8366da06a8215dd16491e24061/objstorage/objstorageprovider/provider.go#L199)):*

[nit] we should put this in a `sharedClose()` so there is symmetry with the open code. And we should call that first, since `sharedInit` is called last
</blockquote></details>

Done.

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2736#-:-N_AYgu27lkasjiVEmxc:bnfp4nl)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jbowens and @joshimhoff)

___
*[`objstorage/objstorageprovider/shared.go` line 123 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2736#-N_AYW2fBDpp_enzKuRG:-N_AYW2fBDpp_enzKuRH:bj74sdr) ([raw file](https://github.com/cockroachdb/pebble/blob/687e3f1bb25c28316ed22bf597090378ea0d775f/objstorage/objstorageprovider/shared.go#L123)):*
> ```Go
> 		p.shared.cache = nil
> 	}
> 	err = firstError(err, p.shared.catalog.Close())
> ```

Let's do an `if` here as well, I think we allow most `Close`s to be called multiple times (not sure it would ever happen in practice).

joshimhoff (commented):
TTFR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2736#-:-N_Acm6_69c-gkeyq_FS:b-80vssw)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @jbowens)

___
*[`objstorage/objstorageprovider/shared.go` line 123 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2736#-N_AYW2fBDpp_enzKuRG:-N_AbCjmBJr5BrrZdcqf:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/687e3f1bb25c28316ed22bf597090378ea0d775f/objstorage/objstorageprovider/shared.go#L123)):*

Let's do an `if` here as well, I think we allow most `Close`s to be called multiple times (not sure it would ever happen in practice).
</blockquote></details>

Done.

Issue #2741, opened 2023-07-13
go-linux-race is flaky


RaduBerinde:
Failure on go-linux-race:
https://github.com/cockroachdb/pebble/actions/runs/5546996852/jobs/10128026245?pr=2724

Looks like a timeout after ~15 minutes, though not at the level of `go test` which has 20 minutes timeout. In any case, we can probably do much better here by skipping some tests in race mode (in particular `lint`).

```
?   	github.com/cockroachdb/pebble/internal/ackseq	[no test files]
?   	github.com/cockroachdb/pebble/internal/bytealloc	[no test files]
?   	github.com/cockroachdb/pebble/internal/crc	[no test files]
?   	github.com/cockroachdb/pebble/internal/datatest	[no test files]
?   	github.com/cockroachdb/pebble/internal/errorfs	[no test files]
?   	github.com/cockroachdb/pebble/internal/humanize	[no test files]
?   	github.com/cockroachdb/pebble/internal/invariants	[no test files]
ok  	github.com/cockroachdb/pebble	286.984s
ok  	github.com/cockroachdb/pebble/bloom	0.257s
ok  	github.com/cockroachdb/pebble/cmd/pebble	0.028s
ok  	github.com/cockroachdb/pebble/internal/arenaskl	0.566s
ok  	github.com/cockroachdb/pebble/internal/base	0.111s
ok  	github.com/cockroachdb/pebble/internal/batchskl	0.140s
ok  	github.com/cockroachdb/pebble/internal/cache	6.992s
ok  	github.com/cockroachdb/pebble/internal/fastrand	0.020s [no tests to run]
ok  	github.com/cockroachdb/pebble/internal/intern	0.028s
ok  	github.com/cockroachdb/pebble/internal/keyspan	1.458s
ok  	github.com/cockroachdb/pebble/internal/lint	208.[75](https://github.com/cockroachdb/pebble/actions/runs/5546996852/jobs/10128026245?pr=2724#step:4:76)4s
?   	github.com/cockroachdb/pebble/internal/manual	[no test files]
ok  	github.com/cockroachdb/pebble/internal/manifest	150.4[80](https://github.com/cockroachdb/pebble/actions/runs/5546996852/jobs/10128026245?pr=2724#step:4:81)s
?   	github.com/cockroachdb/pebble/internal/pacertoy/pebble	[no test files]
?   	github.com/cockroachdb/pebble/internal/pacertoy/rocksdb	[no test files]
[...]

PR #2745, opened 2023-07-17
sharedcache: reduce the scope of TestSharedCacheRandomized under race

RaduBerinde:
Reduce the max number of shards and skip the random block size tests under race. This significantly reduces the runtime and memory footprint of the test.

Informs: #2741


joshimhoff (approved):
LGTM

PR #2748, opened 2023-07-18
db: Add file counts to EstimateDiskUsage return values

dt:


db.go
```diff
@@ -2045,6 +2053,17 @@ func (d *DB) EstimateDiskUsage(start, end []byte) (uint64, error) {
 				d.opts.Comparer.Compare(file.Largest.UserKey, end) <= 0 {
 				// The range fully contains the file, so skip looking it up in
 				// table cache/looking at its indexes, and add the full file size.
+				totalFiles++
+				if file.Virtual {
+					remoteFiles++
```
itsbilal, line 2058: same as below

db.go
```diff
@@ -2068,13 +2087,24 @@ func (d *DB) EstimateDiskUsage(start, end []byte) (uint64, error) {
 					)
 				}
 				if err != nil {
-					return 0, err
+					return 0, 0, 0, 0, err
+				}
+				totalFiles++
+				if file.Virtual {
+					remoteFiles++
```
itsbilal, line 2094: Virtual =/= remote unfortunately, so this will have to be `meta.IsShared()` below (which would, under new terminology, more accurately be called `IsRemote()`)

Ps. you can always look up a file in the provider, it doesn't need to be a virtual file to do this indirection.

itsbilal (approved):
LGTM save for`Virtual` not meaning remote.

db.go
dt: done

dt:
I wonder if this API would be more useful if it was `func GetEstimatedDiskUsage(start, end) (totalBytes, remoteBytes, externalBytes)` ? externalBytes > 0 would be good enough for my purposes -- since > 0 bytes implies > files -- and is potentially a more useful number to other callers? I suppose I could also do file count _and_ byte count but 6 return values from one function feels messy and I'd probably switch to returning a struct and I just don't know how useful the number of files really is to anyone?

dt:
I changed this from file counts to sum of file sizes, so it is ready for a new look.

One question: is there any risk of file.Size being zero? I very much need to be able to assume that if files > 0 then size >0, so if file.Size is able to be zero, I'm tempted to add a case for that that just increments size by 1 so it is non-zero.

db.go
```diff
@@ -2068,13 +2086,23 @@ func (d *DB) EstimateDiskUsage(start, end []byte) (uint64, error) {
...
 				if err != nil {
-					return 0, err
+					return 0, 0, 0, err
+				}
+				meta, err := d.objProvider.Lookup(fileTypeTable, file.FileBacking.DiskFileNum)
+				if err != nil {
+					return 0, 0, 0, err
+				}
+				if meta.IsShared() {
+					remoteSize += file.Size
```
itsbilal, line 2096: This will be `size` not file.Size, as this is the partial overlap case and `size` is being returned by `reader.EstimateDiskUsage`.

itsbilal (commented):
> One question: is there any risk of file.Size being zero? I very much need to be able to assume that if files > 0 then size >0, so if file.Size is able to be zero, I'm tempted to add a case for that that just increments size by 1 so it is non-zero.

We've gone around and added some `if file.Size == 0 { file.Size = 1}` checks in code in places where we were at risk of ending up with zero file sizes for virtual sstables, as we were seeing panics in compaction picking where we were dividing by zero. So yes, it's safe to assume that we will never have a `file.Size == 0` situation as multiple things will be going wrong.

And yes, I think byte counts are generally more useful than file counts so I do prefer the new approach here.

db.go
dt: hm. I yeah, I suppose. I was thinking in terms of our download, and we need to download the whole file, but I guess we'll be passing the whole span, so we're never in this overlap case anyway, so yeah, `size`. 

Done.

itsbilal (approved):
LGTM :shipit:

dt:
TFTR!

RaduBerinde:
*[`db.go` line 2025 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2748#-N_iz7xwC8RkIwdFgxiy:-N_iz7xwC8RkIwdFgxiz:bn19g6j) ([raw file](https://github.com/cockroachdb/pebble/blob/19d5e3d6a2391dae65434c33bef413bfa6b5dd95/db.go#L2025)):*
> ```Go
> 
> // EstimateDiskUsageByBackingType is like EstimateDiskUsage but additionally
> // returns the subsets of that size in remote ane external files.
> ```

[nit] `and`. We need to define "remote" and "external", or point to some place which does.

In Pebble we use "shared" to refer to what "remote" means here. Should we consider renaming all that to remote? If not, we should probably stick to the existing terminology inside Pebble code.

PR #2750, opened 2023-07-18
cmd/pebble,replay: various benchmark replay adjustments

jbowens:
[cmd/pebble: use non-default block cache size in replay tool](https://github.com/cockroachdb/pebble/commit/43a29468e25782b0dcdfa5d7fd657799288f2bea) 

If a block cache size is configured within a checkpoint's OPTIONS file, or if a
block cache size provided through the options string command-line flag, create
a cache of the provided size instead. Also, add a `--max-cache-size` CLI flag
so that a replay run that used a much larger block cache can be used on a
machine with less memory available.

[replay: calculate separate write stall metrics for each reason](https://github.com/cockroachdb/pebble/commit/123981e84e01ced75dd238416830a4190c76b7f1) 

The reason for the stall is important when reasoning about why a change moves
the metric.

[cmd/replay: always log checkpoint initialization](https://github.com/cockroachdb/pebble/commit/965e84bb119ffbfb10ac97035576352a0bc7766c) 

Verbose mode is verbose enough to affect performance numbers, so it's not
typically used. It would be helpful to still be able to determine when the node
was performing a copy of the storage directory.

[cmd/pebble: allow passing an explicit checkpoint directory path](https://github.com/cockroachdb/pebble/commit/3b21f9e1709d185cf81d78449c4292125da72e4b) 

This allows for copying a checkpoint to a separate filesystem before beginning
a test. Then runs of the test may link instead of copy files.

### DIFF
--- a/cmd/pebble/replay.go
+++ b/cmd/pebble/replay.go
@@ -209,7 +209,7 @@ func (c *replayConfig) initRunDir(r *replay.Runner) error {
 	}
 	if !c.ignoreCheckpoint {
 		checkpointDir := r.WorkloadFS.PathJoin(r.WorkloadPath, `checkpoint`)
-		verbosef("Attempting to initialize with checkpoint %q.\n", checkpointDir)
+		fmt.Printf("Attempting to initialize with checkpoint %q.\n", checkpointDir)
 		ok, err := vfs.Clone(
 			r.WorkloadFS,
 			vfs.Default,
@@ -222,7 +222,7 @@ func (c *replayConfig) initRunDir(r *replay.Runner) error {
 		if !ok {
 			return errors.Newf("no checkpoint %q exists; you may re-run with --ignore-checkpoint", checkpointDir)
 		}
-		verbosef("Run directory initialized with checkpoint %q.\n", checkpointDir)
+		fmt.Printf("Run directory initialized with checkpoint %q.\n", checkpointDir)
 	}
 	return nil
 }
--- a/cmd/pebble/replay.go
+++ b/cmd/pebble/replay.go
@@ -20,7 +20,9 @@ import (
 
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble"
+	"github.com/cockroachdb/pebble/bloom"
 	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/cache"
 	"github.com/cockroachdb/pebble/replay"
 	"github.com/cockroachdb/pebble/vfs"
 	"github.com/spf13/cobra"
@@ -52,6 +54,8 @@ func initReplayCmd() *cobra.Command {
 		&c.optionsString, "options", "", "Pebble options to override, in the OPTIONS ini format but with any whitespace as field delimiters instead of newlines")
 	cmd.Flags().StringVar(
 		&c.runDir, "run-dir", c.runDir, "the directory to use for the replay data directory; defaults to a random dir in pwd")
+	cmd.Flags().Int64Var(
+		&c.maxCacheSize, "max-cache-size", c.maxCacheSize, "the max size of the block cache")
 	cmd.Flags().BoolVar(
 		&c.streamLogs, "stream-logs", c.streamLogs, "stream the Pebble logs to stdout during replay")
 	cmd.Flags().BoolVar(
@@ -68,6 +72,7 @@ type replayConfig struct {
 	streamLogs       bool
 	ignoreCheckpoint bool
 	optionsString    string
+	maxCacheSize     int64
 
 	cleanUpFuncs []func() error
 }
@@ -87,6 +92,9 @@ func (c *replayConfig) args() (args []string) {
 	}
 	if c.maxWritesMB != 0 {
 		args = append(args, "--max-writes", fmt.Sprint(c.maxWritesMB))
+	}
+	if c.maxCacheSize != 0 {
+		args = append(args, "--max-cache-size", fmt.Sprint(c.maxCacheSize))
 	}
 	if c.streamLogs {
 		args = append(args, "--stream-logs")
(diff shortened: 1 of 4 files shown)


replay/replay.go
```diff
@@ -576,17 +598,28 @@ const (
...
 			r.cancel()
 		},
-		WriteStallBegin: func(pebble.WriteStallBeginInfo) {
-			r.metrics.writeStalls.Add(1)
+		WriteStallBegin: func(info pebble.WriteStallBeginInfo) {
+			r.writeStallMetrics.Lock()
+			defer r.writeStallMetrics.Unlock()
+			writeStallReason = info.Reason
+			// Take just the first word of the reason.
+			if j := strings.IndexByte(writeStallReason, ' '); j != -1 {
```
msbutler, line 612: nit: could be worth enumerating the write stall reasons to future proof this from folks revising or adding the write stall reason strings

cmd/pebble/replay.go
```diff
@@ -209,7 +209,7 @@ func (c *replayConfig) initRunDir(r *replay.Runner) error {
 	}
 	if !c.ignoreCheckpoint {
 		checkpointDir := r.WorkloadFS.PathJoin(r.WorkloadPath, `checkpoint`)
-		verbosef("Attempting to initialize with checkpoint %q.\n", checkpointDir)
+		fmt.Printf("Attempting to initialize with checkpoint %q.\n", checkpointDir)
```
msbutler, line 222: nit: could we print time.now()?

cmd/pebble/replay.go
```diff
@@ -60,6 +60,8 @@ func initReplayCmd() *cobra.Command {
 		&c.streamLogs, "stream-logs", c.streamLogs, "stream the Pebble logs to stdout during replay")
 	cmd.Flags().BoolVar(
 		&c.ignoreCheckpoint, "ignore-checkpoint", c.ignoreCheckpoint, "ignore the workload's initial checkpoint")
+	cmd.Flags().StringVar(
+		&c.checkpointDir, "checkpoint-dir", c.checkpointDir, "path to the checkpoint to use if not <WORKLOAD_DIR>/checkpoint")
```
msbutler, line 64: could you explain, perhaps in the commit message, the workflow you were envisioning here? This patch is to prevent re-cloning the workload from the snapshot right? I was imagining that a user would run 
```
replay ... --count=10 --checkpoint-dir=foo
```
to slowly clone the checkpoint on the first trial, and then reuse the checkpoint on subsequent runs? Maybe I'm confused on the purpose of this patch.

msbutler (commented):
thanks for doing this!

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2750)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @msbutler and @sumeerbhola)

___
*[`cmd/pebble/replay.go` line 64 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2750#gh-1268385041:-N_jwWbh8c3CDRMPBLzu:b-tlsxin) ([raw file](https://github.com/cockroachdb/pebble/blob/3b21f9e1709d185cf81d78449c4292125da72e4b/cmd/pebble/replay.go#L64)):*

could you explain, perhaps in the commit message, the workflow you were envisioning here? This patch is to prevent re-cloning the workload from the snapshot right? I was imagining that a user would run 
```
replay ... --count=10 --checkpoint-dir=foo
```
to slowly clone the checkpoint on the first trial, and then reuse the checkpoint on subsequent runs? Maybe I'm confused on the purpose of this patch.
</blockquote></details>

Updated the commit: I made the copy of the checkpoint manual. We can't know whether the data already in the checkpoint directory is the right data or not, and passing an old checkpoint-dir seems like an easy mistake. Hopefully it's a little less easy to make if it's clear the onus is on the user to prepare the checkpoint dir.

sumeerbhola (approved):
:lgtm: 

> Reviewed 2 of 2 files at r1, 3 of 3 files at r7, 1 of 1 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2750#-:-N_l3H0qAvWlufoPOKfb:bg7igkv)* status: all files reviewed, 3 unresolved discussions (waiting on @jbowens and @msbutler)

___
*[`cmd/pebble/replay.go` line 288 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2750#-N_kzPfW1QkG7H4rm815:-N_kzPfW1QkG7H4rm816:b-rnelen) ([raw file](https://github.com/cockroachdb/pebble/blob/43a29468e25782b0dcdfa5d7fd657799288f2bea/cmd/pebble/replay.go#L288)):*
> ```Go
> 		},
> 		NewComparer: makeComparer,
> 		NewFilterPolicy: func(name string) (pebble.FilterPolicy, error) {
> ```

this is not mentioned in the commit, but was this an oversight in that we weren't using bloom filters?

jbowens (commented):
One realization I had—I don't think the small block cache size should have really mattered now that we no longer add compaction reads to the block cache.

TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2750#-:-N_nqMvn0J3JUvqSWlNn:b-chwz45)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @sumeerbhola)

___
*[`cmd/pebble/replay.go` line 288 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2750#-N_kzPfW1QkG7H4rm815:-N_nqKzO1atTW_gVKIZx:bbq6832) ([raw file](https://github.com/cockroachdb/pebble/blob/43a29468e25782b0dcdfa5d7fd657799288f2bea/cmd/pebble/replay.go#L288)):*

this is not mentioned in the commit, but was this an oversight in that we weren't using bloom filters?
</blockquote></details>

Yeah, it was.

PR #2746, opened 2023-07-18
db: Add StorageType to SSTableInfo

itsbilal:
The SSTables() method returns sstable info from a Pebble db instance. This change adds an additional field denoting the type of storage backing a given sstable in addition to all other sstable properties; i.e. whether an sstable is local, shared (non-foreign), shared+foreign, or external.

PR #2751, opened 2023-07-18
db: include shared.Locator in SSTableInfo

RaduBerinde:
`DB.SSTables()` now returns whether the sstable is on shared storage and the shared storage locator.


RaduBerinde:
Ah, I didn't notice #2746. Closing.

PR #2753, opened 2023-07-19
ingest: add ability to ingest external (not local or shared) files

itsbilal:
External files are files that can be referenced through the objprovider using a shared.Locator, and are not refcounted or have their lifecycle managed by any Pebble instances. This change adds the ability to ingest external files into Pebble through a new IngestExternalFile method on the DB.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -1238,7 +1238,7 @@ func runForceIngestCmd(td *datadriven.TestData, d *DB) error {
 		*fileMetadata,
 	) (int, error) {
 		return level, nil
-	}, nil, KeyRange{})
+	}, nil /* shared */, KeyRange{}, nil /* external */)
 	return err
 }
 
--- a/flushable_test.go
+++ b/flushable_test.go
@@ -58,7 +58,7 @@ func TestIngestedSSTFlushableAPI(t *testing.T) {
 
 		// We can reuse the ingestLoad function for this test even if we're
 		// not actually ingesting a file.
-		lr, err := ingestLoad(d.opts, d.FormatMajorVersion(), paths, nil, d.cacheID, pendingOutputs)
+		lr, err := ingestLoad(d.opts, d.FormatMajorVersion(), paths, nil, nil, d.cacheID, pendingOutputs, d.objProvider, jobID)
 		meta := lr.localMeta
 		if err != nil {
 			panic(err)
--- a/ingest.go
+++ b/ingest.go
@@ -15,6 +15,7 @@ import (
 	"github.com/cockroachdb/pebble/internal/manifest"
 	"github.com/cockroachdb/pebble/internal/private"
 	"github.com/cockroachdb/pebble/objstorage"
+	"github.com/cockroachdb/pebble/objstorage/shared"
 	"github.com/cockroachdb/pebble/sstable"
 )
 
@@ -128,6 +129,138 @@ func ingestLoad1Shared(
 		// Initialize meta.{Smallest,Largest} and others by calling this.
 		meta.ExtendPointKeyBounds(opts.Comparer.Compare, meta.SmallestPointKey, meta.LargestPointKey)
 	}
+
+	if err := meta.Validate(opts.Comparer.Compare, opts.Comparer.FormatKey); err != nil {
+		return nil, err
+	}
+	return meta, nil
+}
+
+// ingestLoad1External loads the fileMetadata for one external sstable.
+// Sequence number and target level calculation happens during prepare/apply.
+func ingestLoad1External(
+	opts *Options,
+	e ExternalFile,
+	fileNum base.DiskFileNum,
+	objprovider objstorage.Provider,
+	jobID int,
+) (*fileMetadata, error) {
+	if e.Size == 0 {
+		// Disallow 0 file sizes
+		return nil, errors.New("pebble: cannot ingest external file with size 0")
+	}
+	// Don't load table stats. Doing a round trip to shared storage, one SST
+	// at a time is not worth it as it slows down ingestion.
+	meta := &fileMetadata{}
+	meta.FileNum = fileNum.FileNum()
+	meta.CreationTime = time.Now().Unix()
+	meta.Virtual = true
+	meta.Size = e.Size
+	meta.InitProviderBacking(fileNum)
+
+	// Try to resolve a reference to the external file.
+	backing, err := objprovider.CreateSharedObjectBacking(e.Locator, e.ObjName)
+	if err != nil {
+		return nil, err
+	}
+	metas, err := objprovider.AttachSharedObjects([]objstorage.SharedObjectToAttach{{
+		FileNum:  fileNum,
+		FileType: fileTypeTable,
+		Backing:  backing,
+	}})
+	if err != nil {
+		return nil, err
+	}
+	if opts.EventListener.TableCreated != nil {
+		opts.EventListener.TableCreated(TableCreateInfo{
+			JobID:   jobID,
+			Reason:  "ingesting",
+			Path:    objprovider.Path(metas[0]),
+			FileNum: fileNum.FileNum(),
+		})
+	}
+	readable, err := objprovider.OpenForReading(context.TODO(), fileTypeTable, fileNum, objstorage.OpenOptions{MustExist: true})
+	if err != nil {
+		return nil, err
+	}
+	reader, err := sstable.NewReader(readable, opts.MakeReaderOptions())
+	if err != nil {
+		return nil, err
+	}
+	pointKeyIter, err := reader.NewIter(e.SmallestUserKey, e.LargestUserKey)
+	var smallest, largest InternalKey
+	if err != nil {
+		return nil, err
+	}
+	defer pointKeyIter.Close()
+	smallestKey, _ := pointKeyIter.SeekGE(e.SmallestUserKey, base.SeekGEFlagsNone)
+	if smallestKey != nil {
+		smallest.CopyFrom(*smallestKey)
+	}
+	largestKey, _ := pointKeyIter.SeekLT(e.LargestUserKey, base.SeekLTFlagsNone)
+	if largestKey != nil {
+		largest.CopyFrom(*largestKey)
+	}
+	cmp := opts.Comparer.Compare
+	if smallest.UserKey != nil || largest.UserKey != nil {
+		meta.ExtendPointKeyBounds(cmp, smallest, largest)
+	}
+	{
+		// Figure out the range del bounds. Note that the use of Truncate means
+		// that we will never end up with spans that go outside [e.SmallestUserKey,
+		// e.LargestUserKey).
+		rawRangeDelIter, err := reader.NewRawRangeDelIter()
+		if err != nil {
+			return nil, err
+		}
+		rangeDelIter := keyspan.Truncate(cmp,
+			rawRangeDelIter, e.SmallestUserKey, e.LargestUserKey, nil, /* start */
+			nil /* end */, false /* panicOnUpperTruncate */)
+		defer rangeDelIter.Close()
+		span := rangeDelIter.SeekGE(e.SmallestUserKey)
+		var smallestRangeDel, largestRangeDel InternalKey
+		if span != nil {
+			smallestRangeDel.CopyFrom(span.SmallestKey())
+		}
+		span = rangeDelIter.SeekLT(e.LargestUserKey)
+		if span != nil {
+			largestRangeDel.CopyFrom(span.LargestKey())
+		}
+		if smallestRangeDel.UserKey != nil || largestRangeDel.UserKey != nil {
+			meta.ExtendPointKeyBounds(cmp, smallestRangeDel, largestRangeDel)
+		}
+	}
+	{
+		// Figure out the range key bounds. Note that the use of Truncate means
+		// that we will never end up with spans that go outside [e.SmallestUserKey,
+		// e.LargestUserKey).
+		rawRangeKeyIter, err := reader.NewRawRangeKeyIter()
+		if err != nil {
+			return nil, err
+		}
+		rangeKeyIter := keyspan.Truncate(cmp,
+			rawRangeKeyIter, e.SmallestUserKey, e.LargestUserKey, nil, /* start */
+			nil /* end */, false /* panicOnUpperTruncate */)
+		defer rangeKeyIter.Close()
+		span := rangeKeyIter.SeekGE(e.SmallestUserKey)
+		var smallestRangeKey, largestRangeKey InternalKey
+		if span != nil {
+			smallestRangeKey.CopyFrom(span.SmallestKey())
+		}
+		span = rangeKeyIter.SeekLT(e.LargestUserKey)
+		if span != nil {
+			largestRangeKey.CopyFrom(span.LargestKey())
+		}
+		if smallestRangeKey.UserKey != nil || largestRangeKey.UserKey != nil {
+			meta.ExtendRangeKeyBounds(cmp, smallestRangeKey, largestRangeKey)
+		}
+	}
+
+	// Set the underlying FileBacking's size to the same size as the virtualized
+	// view of the sstable. This ensures that we don't over-prioritize this
+	// sstable for compaction just yet, as we do not have a clear sense of
+	// what parts of this sstable are referenced by other nodes.
+	meta.FileBacking.Size = e.Size
 
 	if err := meta.Validate(opts.Comparer.Compare, opts.Comparer.FormatKey); err != nil {
 		return nil, err
@@ -283,8 +416,10 @@ func ingestLoad1(
 
 type ingestLoadResult struct {
 	localMeta, sharedMeta []*fileMetadata
+	externalMeta          []*fileMetadata
 	localPaths            []string
 	sharedLevels          []uint8
+	fileCount             int
 }
 
 func ingestLoad(
@@ -292,8 +427,11 @@ func ingestLoad(
 	fmv FormatMajorVersion,
 	paths []string,
 	shared []SharedSSTMeta,
+	external []ExternalFile,
 	cacheID uint64,
 	pending []base.DiskFileNum,
+	objProvider objstorage.Provider,
+	jobID int,
 ) (ingestLoadResult, error) {
 	meta := make([]*fileMetadata, 0, len(paths))
 	newPaths := make([]string, 0, len(paths))
@@ -316,8 +454,8 @@ func ingestLoad(
 			newPaths = append(newPaths, paths[i])
 		}
 	}
-	if len(shared) == 0 {
-		return ingestLoadResult{localMeta: meta, localPaths: newPaths}, nil
+	if len(shared) == 0 && len(external) == 0 {
+		return ingestLoadResult{localMeta: meta, localPaths: newPaths, fileCount: len(meta)}, nil
 	}
 	sharedMeta := make([]*fileMetadata, 0, len(shared))
 	levels := make([]uint8, 0, len(shared))
@@ -332,11 +470,21 @@ func ingestLoad(
 		sharedMeta = append(sharedMeta, m)
 		levels = append(levels, shared[i].Level)
 	}
+	externalMeta := make([]*fileMetadata, 0, len(external))
+	for i := range external {
+		m, err := ingestLoad1External(opts, external[i], pending[len(paths)+len(shared)+i], objProvider, jobID)
+		if err != nil {
+			return ingestLoadResult{}, err
+		}
+		externalMeta = append(externalMeta, m)
+	}
 	result := ingestLoadResult{
 		localMeta:    meta,
 		sharedMeta:   sharedMeta,
+		externalMeta: externalMeta,
 		localPaths:   newPaths,
 		sharedLevels: levels,
+		fileCount:    len(meta) + len(sharedMeta) + len(externalMeta),
 	}
 	return result, nil
 }
@@ -360,7 +508,9 @@ func (m metaAndPaths) Less(i, j int) bool {
 
 func (m metaAndPaths) Swap(i, j int) {
 	m.meta[i], m.meta[j] = m.meta[j], m.meta[i]
-	m.paths[i], m.paths[j] = m.paths[j], m.paths[i]
+	if m.paths != nil {
+		m.paths[i], m.paths[j] = m.paths[j], m.paths[i]
+	}
 }
 
 func ingestSortAndVerify(cmp Compare, lr ingestLoadResult, exciseSpan KeyRange) error {
@@ -371,6 +521,23 @@ func ingestSortAndVerify(cmp Compare, lr ingestLoadResult, exciseSpan KeyRange) 
 		if !exciseSpan.Contains(cmp, f.Smallest) || !exciseSpan.Contains(cmp, f.Largest) {
 			return errors.AssertionFailedf("pebble: shared file outside of excise span, span [%s-%s), file = %s", exciseSpan.Start, exciseSpan.End, f.String())
 		}
+	}
+	if len(lr.externalMeta) > 0 {
+		if len(lr.localMeta) > 0 || len(lr.sharedMeta) > 0 {
+			// Currently we only support external ingests on their own. If external
+			// files are present alongside local/shared files, return an error.
+			return errors.AssertionFailedf("pebble: external files cannot be ingested atomically alongside other types of files")
+		}
+		sort.Sort(&metaAndPaths{
+			meta: lr.externalMeta,
+			cmp:  cmp,
+		})
+		for i := 1; i < len(lr.externalMeta); i++ {
+			if sstableKeyCompare(cmp, lr.externalMeta[i-1].Largest, lr.externalMeta[i].Smallest) >= 0 {
+				return errors.AssertionFailedf("pebble: external sstables have overlapping ranges")
+			}
+		}
+		return nil
 	}
 	if len(lr.localMeta) <= 1 || len(lr.localPaths) <= 1 {
 		return nil
@@ -384,7 +551,7 @@ func ingestSortAndVerify(cmp Compare, lr ingestLoadResult, exciseSpan KeyRange) 
 
 	for i := 1; i < len(lr.localPaths); i++ {
 		if sstableKeyCompare(cmp, lr.localMeta[i-1].Largest, lr.localMeta[i].Smallest) >= 0 {
-			return errors.AssertionFailedf("pebble: external sstables have overlapping ranges")
+			return errors.AssertionFailedf("pebble: local ingestion sstables have overlapping ranges")
 		}
 	}
 	if len(lr.sharedMeta) == 0 {
@@ -489,6 +656,8 @@ func ingestLink(
 			})
 		}
 	}
+	// We do not need to do anything about lr.externalMetas. Those were already
+	// linked in ingestLoad.
 
 	return nil
 }
@@ -881,7 +1050,7 @@ func (d *DB) Ingest(paths []string) error {
 	if d.opts.ReadOnly {
 		return ErrReadOnly
 	}
-	_, err := d.ingest(paths, ingestTargetLevel, nil /* shared */, KeyRange{})
+	_, err := d.ingest(paths, ingestTargetLevel, nil /* shared */, KeyRange{}, nil /* external */)
 	return err
 }
 
@@ -901,6 +1070,27 @@ type IngestOperationStats struct {
 	MemtableOverlappingFiles int
 }
 
+// ExternalFile are external sstables that can be referenced through
+// objprovider and ingested as remote files that will not be refcounted or
+// cleaned up. For use with online restore. Note that the underlying sstable
+// could contain keys outside the [Smallest,Largest) bounds; however Pebble
+// is expected to only read the keys within those bounds.
+type ExternalFile struct {
+	// Locator is the shared.Locator that can be used with objProvider to
+	// resolve a reference to this external sstable.
+	Locator shared.Locator
+	// ObjName is the unique name of this sstable on Locator.
+	ObjName string
+	// Size of the referenced proportion of the virtualized sstable. An estimate
+	// is acceptable in lieu of the backing file size.
+	Size uint64
+	// SmallestUserKey and LargestUserKey are the [smallest,largest) user key
+	// bounds of the sstable. Note that this Largest bound is exclusive, and must
+	// be resolved into an inclusive user key bound through a SeekLT at ingestion
+	// time, to obey virtual FileMetadata guarantees.
+	SmallestUserKey, LargestUserKey []byte
+}
+
 // IngestWithStats does the same as Ingest, and additionally returns
 // IngestOperationStats.
 func (d *DB) IngestWithStats(paths []string) (IngestOperationStats, error) {
@@ -910,7 +1100,25 @@ func (d *DB) IngestWithStats(paths []string) (IngestOperationStats, error) {
 	if d.opts.ReadOnly {
 		return IngestOperationStats{}, ErrReadOnly
 	}
-	return d.ingest(paths, ingestTargetLevel, nil /* shared */, KeyRange{})
+	return d.ingest(paths, ingestTargetLevel, nil /* shared */, KeyRange{}, nil /* external */)
+}
+
+// IngestExternalFiles does the same as IngestWithStats, and additionally
+// accepts external files (with locator info that can be resolved using
+// d.opts.SharedStorage). These files must also be non-overlapping with
+// each other, and must be resolvable through
+func (d *DB) IngestExternalFiles(external []ExternalFile) (IngestOperationStats, error) {
+	if err := d.closed.Load(); err != nil {
+		panic(err)
+	}
+
+	if d.opts.ReadOnly {
+		return IngestOperationStats{}, ErrReadOnly
+	}
+	if d.opts.Experimental.SharedStorage == nil {
+		return IngestOperationStats{}, errors.New("pebble: cannot ingest external files without shared storage configured")
+	}
+	return d.ingest(nil, ingestTargetLevel, nil /* shared */, KeyRange{}, external)
 }
 
 // IngestAndExcise does the same as IngestWithStats, and additionally accepts a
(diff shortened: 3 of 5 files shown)


itsbilal:
Added tests. This is now ready for a look.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753)* status: 2 of 7 files reviewed, 9 unresolved discussions (waiting on @dt, @itsbilal, and @msbutler)

___
*[`ingest.go` line 1024 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u5ZrN9INEZa30IiCV:-N_u5ZrN9INEZa30IiCW:biru6uz) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L1024)):*
> ```Go
> 	// SmallestUserKey and LargestUserKey are the [smallest,largest) user key
> 	// bounds of the sstable. Note that this Largest bound is exclusive, and must
> 	// be resolved into an inclusive user key bound through a SeekLT at ingestion
> ```

But we don't do this at ingestion time, do we?
___
*[`ingest.go` line 1044 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u5ezH5o0AVBkcCPPd:-N_u5ezH5o0AVBkcCPPe:bl76m6o) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L1044)):*
> ```Go
> // accepts external files (with locator info that can be resolved using
> // d.opts.SharedStorage). These files must also be non-overlapping with
> // each other, and must be resolvable through
> ```

[nit] unfinished comment
___
*[`ingest.go` line 1389 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u68MsDSBjfa6Wv8Oq:-N_u68MsDSBjfa6Wv8Or:b-bbcbgz) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L1389)):*
> ```Go
> 		}
> 
> 		// Update the sequence number for all local sstables in the
> ```

[nit] local or external
___
*[`ingest.go` line 1398 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u62EsBBCcyZef4lag:-N_u62EsBBCcyZef4lah:b-z0idqn) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L1398)):*
> ```Go
> 		metas := loadResult.localMeta
> 		if len(loadResult.externalMeta) > 0 {
> 			metas = loadResult.externalMeta
> ```

We are relying here that only one of `localMeta` and `externalMeta` is not empty? If yes we should add a comment
___
*[`ingest.go` line 1430 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u5zOw2vwA2zL9Iidb:-N_u5zOw2vwA2zL9Iidc:b-touhmb) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L1430)):*
> ```Go
> 	// during ingestion.
> 	d.commit.ingestSem <- struct{}{}
> 	d.commit.AllocateSeqNum(len(loadResult.localMeta)+len(loadResult.externalMeta), prepare, apply)
> ```

[nit] could this be `len(metas)`?
___
*[`ingest.go` line 1453 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u6E8q9FnJHrGt6HCV:-N_u6E8q9FnJHrGt6HCW:brk0ju5) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L1453)):*
> ```Go
> 		globalSeqNum = loadResult.localMeta[0].SmallestSeqNum
> 	}
> 	if len(loadResult.externalMeta) > 0 {
> ```

Again, we're relying on the fact that only one of these branches can be taken? Then maybe add an `else` to make that more clear
___
*[`table_cache.go` line 544 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u6jY53lZkEFEqA89R:-N_u6jY636SrcULYQsZY:b-61a4y9) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/table_cache.go#L544)):*
> ```Go
> 	// it was an external ingestion, even if its creator ID does not match our
> 	// own. Exclude such files from going through pointCollapsingIterator.
> 	if provider.IsForeign(objMeta) && file.SmallestSeqNum < base.SeqNumStart {
> ```

I'm a bit uneasy about the `IsForeign` method because it encompasses two cases - shared but foreign, and external. Do we want to include the external case here? In general, maybe we should have `IsSharedForeign` and have the callers use `IsSharedForeign() || IsExternal()` when appropriate?
___
*[`table_cache.go` line 545 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u6Xj83NdcjROlji9k:-N_u6Xj83NdcjROlji9l:b-nfzpog) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/table_cache.go#L545)):*
> ```Go
> 	// own. Exclude such files from going through pointCollapsingIterator.
> 	if provider.IsForeign(objMeta) && file.SmallestSeqNum < base.SeqNumStart {
> 		// NB: IsForeign() guarantees IsShared, so opts must not be nil as we've
> ```

guarantees `IsRemote`
___
*[`data_test.go` line 594 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u8ZdeBXz97dE6cqJp:-N_u7_Ap9RfP_bwVlcP1:b-qnehp2) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/data_test.go#L594)):*
> ```Go
> }
> 
> type noSyncWrapper struct {
> ```

Why do this instead of simply making an `objstorage.Writable` implementation?

We could export a `NewRemoteWritable` from `objstorageprovider`. We can update the `sharedWritable` code to support an empty `meta`.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753)* status: 2 of 7 files reviewed, 11 unresolved discussions (waiting on @dt, @itsbilal, @msbutler, and @RaduBerinde)

___
*[`ingest.go` line 192 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na8-oJ52zvViPevYyuj:-Na8-oJ52zvViPevYyuk:b-jy7y9z) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L192)):*
> ```Go
> 		base.MakeRangeDeleteSentinelKey(largestCopy))
> 	meta.ExtendRangeKeyBounds(opts.Comparer.Compare, base.MakeInternalKey(smallestCopy, 0, InternalKeyKindRangeKeySet),
> 		base.MakeExclusiveSentinelKey(InternalKeyKindRangeKeyDelete, largestCopy))
> ```

Does backup have any record of whether there are range keys in these files? When all the files have bounds indicating they contain range keys, the 'lazy combined iterator' optimization is defeated, adding significant overhead to all reads.

Always setting large range key bounds is likely cumulatively much more expensive than looking for the existence of range keys through loading the table into the table cache and then checking its properties.
___
*[`ingest.go` line 1054 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na81JFK4oRpDKc46-Kx:-Na81JFK4oRpDKc46-Ky:b-afm5ex) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L1054)):*
> ```Go
> 	}
> 	if d.opts.Experimental.SharedStorage == nil {
> 		return IngestOperationStats{}, errors.New("pebble: cannot ingest external files without shared storage configured")
> ```

Does `Experimental.SharedStorage` handle supporting two 'external' configurations, one for external files stored in a backup bucket and another for typical disaggregated shared storage? Or are we saying that a cluster using either the disaggregated tech preview or the progressive restore feature will be ineligible for upgrading to a subsequent release? Just concerned about painting ourselves into a corner here in which it's difficult to upgrade old 23.2 clusters.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753)* status: 2 of 7 files reviewed, 11 unresolved discussions (waiting on @dt, @itsbilal, @jbowens, and @msbutler)

___
*[`ingest.go` line 1054 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na81JFK4oRpDKc46-Kx:-Na8AN1j9eZhf7K57xZk:b-l8vga) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L1054)):*

Does `Experimental.SharedStorage` handle supporting two 'external' configurations, one for external files stored in a backup bucket and another for typical disaggregated shared storage? Or are we saying that a cluster using either the disaggregated tech preview or the progressive restore feature will be ineligible for upgrading to a subsequent release? Just concerned about painting ourselves into a corner here in which it's difficult to upgrade old 23.2 clusters.
</blockquote></details>

Yes, it's a factory that takes a locator string, an arbitrary number of buckets are supported. It's been renamed to `RemoteStorage`.

msbutler (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753)* status: 2 of 7 files reviewed, 11 unresolved discussions (waiting on @dt, @itsbilal, @jbowens, and @RaduBerinde)

___
*[`ingest.go` line 192 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na8-oJ52zvViPevYyuj:-Na82THA2huCjNk8GLdS:bpf223g) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L192)):*

Does backup have any record of whether there are range keys in these files? When all the files have bounds indicating they contain range keys, the 'lazy combined iterator' optimization is defeated, adding significant overhead to all reads.

Always setting large range key bounds is likely cumulatively much more expensive than looking for the existence of range keys through loading the table into the table cache and then checking its properties.
</blockquote></details>

@dt and I were just talking about this. For restore correctness (of certain flavors, see below), a virtual sst cannot materialize _both_ the point keys and overlapping range keys in the backing sst, because at compaction time, according to the soon to be suffix rewrite rule, all key timestamps in the virtual sst will get rewritten to the same timestamp (i.e. we'll lose the relative ordering of a point key and an overlapping range key). 

To address this, restore will need to call AddRemoteSST twice on a given span: first to create a virtual sst for the rangekeys in a given span, then for the point keys. To that end, we will probably need to add an arg to AddRemotSST so the caller can indicate what kind of keys the virtual sst should keep track of: point keys or range keys?

NB: this isn't a problem for Restores of Tenants (bc they dont need the suffix rewrite rule) or restores of full backups without revision history (bc they contain no rangekeys).

itsbilal (commented):
> Dismissed @jbowens from a discussion.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753)* status: 2 of 12 files reviewed, 5 unresolved discussions (waiting on @dt and @RaduBerinde)

___
*[`ingest.go` line 192 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na8-oJ52zvViPevYyuj:-Na8ZpV-EgVq5D4CELeU:bitgs96) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L192)):*

@dt and I were just talking about this. For restore correctness (of certain flavors, see below), a virtual sst cannot materialize _both_ the point keys and overlapping range keys in the backing sst, because at compaction time, according to the soon to be suffix rewrite rule, all key timestamps in the virtual sst will get rewritten to the same timestamp (i.e. we'll lose the relative ordering of a point key and an overlapping range key). 

To address this, restore will need to call AddRemoteSST twice on a given span: first to create a virtual sst for the rangekeys in a given span, then for the point keys. To that end, we will probably need to add an arg to AddRemotSST so the caller can indicate what kind of keys the virtual sst should keep track of: point keys or range keys?

NB: this isn't a problem for Restores of Tenants (bc they dont need the suffix rewrite rule) or restores of full backups without revision history (bc they contain no rangekeys). 
</blockquote></details>

@msbutler thanks for adding colour about that! I've updated the `ExternalFile` struct to allow specifying what type of keys are present in this file. Basically I've added `HasPointKeys` and `HasRangeKeys` as bools in that struct that we take from Cockroach. That should address this concern, as Backup would be able to tell us what type of keys this sstable contains.

Since we do load the properties block during ingestApply, maybe we should also assert against these bools matching  up with whatever we end up reading down there. (eg. it's cool if we have HasRangeKeys = false and the properties block says we have range keys, but it should be an error if it's the other way around).
___
*[`ingest.go` line 1024 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u5ZrN9INEZa30IiCV:-Na8Ih7i8jEdq-H2Q5Y-:b-2pkdut) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L1024)):*

But we don't do this at ingestion time, do we?
</blockquote></details>

Good catch, we don't. Comment is from the older iteration of this that did this, but DR advised us to remove this.
___
*[`ingest.go` line 1398 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u62EsBBCcyZef4lag:-Na8JIpv3aofxlyUQKGE:b-2znw2y) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L1398)):*

We are relying here that only one of `localMeta` and `externalMeta` is not empty? If yes we should add a comment
</blockquote></details>

Yes, that's right. The check above in `ingestSortAndVerify` would have returned an error if that wasn't the case.
___
*[`ingest.go` line 1430 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u5zOw2vwA2zL9Iidb:-Na8Jeyq-GzxDPc92Zv0:b-7003eu) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L1430)):*

[nit] could this be `len(metas)`?
</blockquote></details>

The exclusion of shared sstables is intentional though, as those don't get seqnums here.
___
*[`ingest.go` line 1453 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u6E8q9FnJHrGt6HCV:-Na8JjY2D4nk1swoCEyo:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L1453)):*

Again, we're relying on the fact that only one of these branches can be taken? Then maybe add an `else` to make that more clear
</blockquote></details>

Done.
___
*[`table_cache.go` line 544 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u6jY53lZkEFEqA89R:-Na8Km8-CmfETnnOJxI8:b-kf9lun) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/table_cache.go#L544)):*

I'm a bit uneasy about the `IsForeign` method because it encompasses two cases - shared but foreign, and external. Do we want to include the external case here? In general, maybe we should have `IsSharedForeign` and have the callers use `IsSharedForeign() || IsExternal()` when appropriate?
</blockquote></details>

I went with the IsSharedForeign case. That's the best use for that method anyway. 
___
*[`data_test.go` line 594 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-N_u8ZdeBXz97dE6cqJp:-Na8M3GH7JixyHHcFIvU:b-vqmh83) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/data_test.go#L594)):*

Why do this instead of simply making an `objstorage.Writable` implementation?

We could export a `NewRemoteWritable` from `objstorageprovider`. We can update the `sharedWritable` code to support an empty `meta`.
</blockquote></details>

It was only test code, so I just went with the quick thing. But what you suggested makes sense. Done.

msbutler (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753)* status: 2 of 12 files reviewed, 5 unresolved discussions (waiting on @dt and @RaduBerinde)

___
*[`ingest.go` line 192 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na8-oJ52zvViPevYyuj:-NaCx7up3cYwvzIKYJPJ:bcyny7j) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L192)):*

> all key timestamps in the virtual sst will get rewritten to the same timestamp (i.e. we'll lose the relative ordering of a point key and an overlapping range key).

I don't understand. Ignoring range keys, this sst has one version per point, yes? And so it is ok to rewrite all point versions (across all `roachpb.Keys`) to one timestamp t?
If my understanding is not flawed for point keys, then why do range keys change the picture? If there is a range key for a span there shouldn't be a point key in that span, since the range key represented the most recent MVCC state for that span.
What am I missing?
</blockquote></details>

You're understanding is correct for point keys: in a given physical sst there will be one version per point.

There could exist a range key at t1 that intersects with a point key at t2. Consider the following scenario: a non revision history incremental backup needs to grab the latest version of each key in the span [a,c) from time t1 to t3. Suppose there exists a rangekey a-c@t1 and a point key b@t2. To restore correctly to an AOST later than t1, both the range key and the point key need to go in the backup. We also do not rewrite the rangekey during export requests
https://github.com/cockroachdb/cockroach/blob/3c866e4e6b9d707128ab1aa211b25a8c3510bce1/pkg/storage/mvcc.go#L6736

Therefore, the physical backup sst can end up with a point key that intersects with a rangekey.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753)* status: 2 of 12 files reviewed, 5 unresolved discussions (waiting on @dt and @RaduBerinde)

___
*[`ingest.go` line 192 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na8-oJ52zvViPevYyuj:-NaHoxhO-0Z5PGZPfCVz:b-g4hcc3) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L192)):*
> We also do not rewrite the rangekey during export requests

This seems an odd choice, since it prevents us from having a nice invariant of a single version for every key. Pebble is already configured with `EngineComparer.ImmediateSuccessor` that we could use to split the range key.

@jbowens can you shed some light on this choice?

dt (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753)* status: 2 of 12 files reviewed, 5 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`ingest.go` line 192 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na8-oJ52zvViPevYyuj:-NaIxhSyAOh-TcJoKJMw:bde2hux) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L192)):*

> We also do not rewrite the rangekey during export requests

This seems an odd choice, since it prevents us from having a nice invariant of a single version for every key. Pebble is already configured with `EngineComparer.ImmediateSuccessor` that we could use to split the range key.

@jbowens can you shed some light on this choice?
</blockquote></details>

Can we put a TODO here (or add a flag like "containsNoRangekeys" and TODO to pass a value to it) and then move this discussion of what metatdata/content we think we want in backup ssts to a doc or something instead of just one comment thread?

itsbilal:
*[`ingest.go` line 192 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na8-oJ52zvViPevYyuj:-NaIyDlU9iegGzNLP008:btmnru1) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L192)):*

Can we put a TODO here (or add a flag like "containsNoRangekeys" and TODO to pass a value to it) and then move this discussion of what metatdata/content we think we want in backup ssts to a doc or something instead of just one comment thread?
</blockquote></details>

The latest push should have already done that (there are HasPointKeys and HasRangeKeys bools passed through the `ExternalFile` struct), effectively passing on this buck up to Cockroach.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753)* status: 2 of 12 files reviewed, 5 unresolved discussions (waiting on @dt, @itsbilal, @msbutler, @RaduBerinde, and @sumeerbhola)

___
*[`ingest.go` line 192 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na8-oJ52zvViPevYyuj:-NaNqmUO21Dsqy-1voj1:b-3atvkf) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L192)):*
>This seems an odd choice, since it prevents us from having a nice invariant of a single version for every key. Pebble is already configured with `EngineComparer.ImmediateSuccessor` that we could use to split the range key.
>
> [@jbowens](https://github.com/jbowens) can you shed some light on this choice?

I've no recollection of this choice. Are we sure this is the case? I did a little bit of searching and found #82873 which truncated range tombstones to the SST bounds.

msbutler (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753)* status: 2 of 12 files reviewed, 5 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`ingest.go` line 192 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na8-oJ52zvViPevYyuj:-NaHzRGp391_PYOL5yQ6:bgasyxi) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L192)):*

>This seems an odd choice, since it prevents us from having a nice invariant of a single version for every key. Pebble is already configured with `EngineComparer.ImmediateSuccessor` that we could use to split the range key.
>
> [@jbowens](https://github.com/jbowens) can you shed some light on this choice?

I've no recollection of this choice. Are we sure this is the case? I did a little bit of searching and found #82873 which truncated range tombstones to the SST bounds.
</blockquote></details>

Yeah, check out this test case https://github.com/cockroachdb/cockroach/blob/bfe7843aefd6f7d8c1e94ef589b1a52f1cbf5c0e/pkg/storage/testdata/mvcc_histories/export#L372

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753)* status: 2 of 12 files reviewed, 5 unresolved discussions (waiting on @dt, @itsbilal, @msbutler, @RaduBerinde, and @sumeerbhola)

___
*[`ingest.go` line 192 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na8-oJ52zvViPevYyuj:-NaO7-xC7dLDHt6x_RPm:behmgq9) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L192)):*

Yeah, check out this test case https://github.com/cockroachdb/cockroach/blob/bfe7843aefd6f7d8c1e94ef589b1a52f1cbf5c0e/pkg/storage/testdata/mvcc_histories/export#L372
</blockquote></details>

I'm probably missing something; isn't that within bounds? The bounds are `k=a end=z` and all the range keys fall within?

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753)* status: 2 of 12 files reviewed, 5 unresolved discussions (waiting on @4, @itsbilal, and @RaduBerinde)

___
*[`ingest.go` line 192 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na8-oJ52zvViPevYyuj:-NaP2lNO8rtm2hf3sdU1:b-n0pwhd) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L192)):*

I'm probably missing something; isn't that within bounds? The bounds are `k=a end=z` and all the range keys fall within?
</blockquote></details>

I may have been unclear -- I didn't mean to refer to the sst bounds.

What I meant to say was that if we have a range key [a,d)@4 and a point key c@7, we could have chosen to export [a,c)@4, c@7, [ImmediateSuccessor(c),d)@4, so there would be exactly one version for each point in the key space.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753)* status: 2 of 12 files reviewed, 4 unresolved discussions (waiting on @4 and @itsbilal)

___
*[`objstorage/objstorage.go` line 271 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2753#-NaTA1gpElPMOHJRt7L1:-NaTA1gpElPMOHJRt7L2:baxf4yd) ([raw file](https://github.com/cockroachdb/pebble/blob/0a927a9a1157ad7640c3f688832d13f095e6bd38/objstorage/objstorage.go#L271)):*
> ```Go
> 	SetCreatorID(creatorID CreatorID) error
> 
> 	// IsSharedForeign returns whether this object is owned by a different node. Return
> ```

[nit] line wrapping. I'd remove the "return value undefined" thing. It's correct to return false if we have no creator ID, it can't possibly be a shared object in that case.
___
*[`objstorage/objstorageprovider/remote.go` line 158 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2753#-NaTAPDOEJX0p_cYOi9t:-NaTAPDOEJX0p_cYOi9u:bkiksgz) ([raw file](https://github.com/cockroachdb/pebble/blob/0a927a9a1157ad7640c3f688832d13f095e6bd38/objstorage/objstorageprovider/remote.go#L158)):*
> ```Go
> 		return false
> 	}
> 	return meta.IsRemote() && !meta.IsExternal() && (meta.Remote.CustomObjectName != "" || meta.Remote.CreatorID != p.remote.shared.creatorID)
> ```

use `meta.IsShared()` instead of `IsRemote && !IsExternal` and remove `meta.Remote.CustomObjectName != ""` (which is the same as `IsExternal` which is already false here)
___
*[`objstorage/objstorageprovider/shared_writable.go` line 21 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2753#-NaTAoQR8EZzqcKSSN0z:-NaTAoQR8EZzqcKSSN1-:b-kugdzx) ([raw file](https://github.com/cockroachdb/pebble/blob/0a927a9a1157ad7640c3f688832d13f095e6bd38/objstorage/objstorageprovider/shared_writable.go#L21)):*
> ```Go
> // WriteCloser returned by remote.Storage.CreateObject.
> type sharedWritable struct {
> 	// Either both p and meta must be unset / zero values, or both must be set.
> ```

[nit] Explain that `p` and `meta` can be unset but only in testing scenarios.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753)* status: 2 of 12 files reviewed, 4 unresolved discussions (waiting on @4, @itsbilal, and @RaduBerinde)

___
*[`ingest.go` line 192 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na8-oJ52zvViPevYyuj:-NaTMPlbDNdH6-4TtW7i:bfdazsr) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L192)):*

I may have been unclear -- I didn't mean to refer to the sst bounds.

What I meant to say was that if we have a range key [a,d)@4 and a point key c@7, we could have chosen to export [a,c)@4, c@7, [ImmediateSuccessor(c),d)@4, so there would be exactly one version for each point in the key space.

</blockquote></details>

Ah, sorry, wasn't reading carefully enough.

That makes sense and is clean. If there are many keys at timestamps above the range key, we would end up with an explosion of range keys though. Our range key implementation is built around the assumption that they're sparse, which could translate to a pathological performance regression if we're online restoring a sstable with thousands of range key fragments.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753#-:-Nai0CWw9QPR_54zska-:b-814g8k)* status: 2 of 12 files reviewed, 1 unresolved discussion (waiting on @4, @itsbilal, and @RaduBerinde)

___
*[`ingest.go` line 192 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2753#-Na8-oJ52zvViPevYyuj:-NahzBZk08kQo-Ev_Kt9:b-ubq4mp) ([raw file](https://github.com/cockroachdb/pebble/blob/e19dbf48b037a53597dd9c3e495856bbc0191994/ingest.go#L192)):*

Ah, sorry, wasn't reading carefully enough.

That makes sense and is clean. If there are many keys at timestamps above the range key, we would end up with an explosion of range keys though. Our range key implementation is built around the assumption that they're sparse, which could translate to a pathological performance regression if we're online restoring a sstable with thousands of range key fragments.
</blockquote></details>

The broader conversation about efficient handling of range keys in online restore is best handled in a separate slack thread. @msbutler will be creating one soon. 
___
*[`objstorage/objstorage.go` line 271 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2753#-NaTA1gpElPMOHJRt7L1:-Nahz6TT6Wj2U-5S1qgr:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0a927a9a1157ad7640c3f688832d13f095e6bd38/objstorage/objstorage.go#L271)):*

[nit] line wrapping. I'd remove the "return value undefined" thing. It's correct to return false if we have no creator ID, it can't possibly be a shared object in that case.
</blockquote></details>

Done.
___
*[`objstorage/objstorageprovider/remote.go` line 158 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2753#-NaTAPDOEJX0p_cYOi9t:-Nahys4XAj8tUhJinaxc:b-376vyb) ([raw file](https://github.com/cockroachdb/pebble/blob/0a927a9a1157ad7640c3f688832d13f095e6bd38/objstorage/objstorageprovider/remote.go#L158)):*

use `meta.IsShared()` instead of `IsRemote && !IsExternal` and remove `meta.Remote.CustomObjectName != ""` (which is the same as `IsExternal` which is already false here)
</blockquote></details>

Done. Good catch.
___
*[`objstorage/objstorageprovider/shared_writable.go` line 21 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2753#-NaTAoQR8EZzqcKSSN0z:-Nahyt9p68G7ehiiTNlJ:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0a927a9a1157ad7640c3f688832d13f095e6bd38/objstorage/objstorageprovider/shared_writable.go#L21)):*

[nit] Explain that `p` and `meta` can be unset but only in testing scenarios.
</blockquote></details>

Done.

jbowens (approved):
:lgtm:

> Reviewed 2 of 5 files at r1, 1 of 5 files at r2, 1 of 5 files at r4, 2 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753#-:-NbQeqBb3D-XXej6gby0:bnfp4nl)* status: 5 of 12 files reviewed, 3 unresolved discussions (waiting on @4, @itsbilal, @RaduBerinde, and @sumeerbhola)

___
*[`ingest.go` line 150 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2753#-NalqGkDDp97FfcjrYDD:-NalqGkDDp97FfcjrYDE:b-s22d1v) ([raw file](https://github.com/cockroachdb/pebble/blob/4a19dfe63a4569923af4396792fd54e9b99b447e/ingest.go#L150)):*
> ```Go
> 	if e.Size == 0 {
> 		// Disallow 0 file sizes
> 		return nil, errors.New("pebble: cannot ingest external file with size 0")
> ```

Is this possible? If a file has a zero file size, will it necessarily hit the case immediately below.
___
*[`ingest.go` line 588 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2753#-NbQeZCLBW1Qs2aWu1S7:-NbQeZCLBW1Qs2aWu1S8:b-sy1onz) ([raw file](https://github.com/cockroachdb/pebble/blob/b42c010441bcaa13c9aa17b8025e6cbf02cdb226/ingest.go#L588)):*
> ```Go
> 	}
> 	// We do not need to do anything about lr.externalMetas. Those were already
> 	// linked in ingestLoad.
> ```

Not for now, but I suspect a refactoring of ingestion could reduce some of these special cases.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2753#-:-NbQhjL9E_F2e9sJahQ4:b-814g8k)* status: 5 of 12 files reviewed, 3 unresolved discussions (waiting on @4, @jbowens, @RaduBerinde, and @sumeerbhola)

___
*[`ingest.go` line 150 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2753#-NalqGkDDp97FfcjrYDD:-NbQhDZ991Fc188to9zo:b-7829hp) ([raw file](https://github.com/cockroachdb/pebble/blob/4a19dfe63a4569923af4396792fd54e9b99b447e/ingest.go#L150)):*

Is this possible? If a file has a zero file size, will it necessarily hit the case immediately below.
</blockquote></details>

We call `Reader.estimateDiskUsage` which ignores range keys, so this case is possible in some range-key-only cases. However we try to force the caller to force a nonzero size, as a zero size causes all sorts of divide-by-zero errors in compaction picking.

We have similar logic in `truncateSharedFiles` for shared files, as otherwise we end up with zero-size files that do have something in it.
___
*[`ingest.go` line 588 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2753#-NbQeZCLBW1Qs2aWu1S7:-NbQhaHLB5dSIZCEj0GD:be8li9e) ([raw file](https://github.com/cockroachdb/pebble/blob/b42c010441bcaa13c9aa17b8025e6cbf02cdb226/ingest.go#L588)):*

Not for now, but I suspect a refactoring of ingestion could reduce some of these special cases.
</blockquote></details>

Yeah, another way would be we'd need to get the provider to allow opening a FileBacking without necessarily attaching it.

PR #2754, opened 2023-07-19
objstorage: rename Shared to Remote in the objstorage provider API

RaduBerinde:
`IsShared()` is split into `IsRemote()`, `IsShared()`, `IsExternal()`.

I tried to update the former uses of `IsShared()` appropriately - use `IsRemote()` in most cases but keep using `IsShared()` in the disagg storage code for skip-shared iteration. These callsites can use reviewing though.

I can wait to merge this when it is convenient, given the other ongoing work.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -449,19 +449,19 @@ func ingestLink(
 			})
 		}
 	}
-	sharedObjs := make([]objstorage.SharedObjectToAttach, 0, len(shared))
+	sharedObjs := make([]objstorage.RemoteObjectToAttach, 0, len(shared))
 	for i := range shared {
 		backing, err := shared[i].Backing.Get()
 		if err != nil {
 			return err
 		}
-		sharedObjs = append(sharedObjs, objstorage.SharedObjectToAttach{
+		sharedObjs = append(sharedObjs, objstorage.RemoteObjectToAttach{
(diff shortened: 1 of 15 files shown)


RaduBerinde:
Updated to rename backing-related types and methods.

ingest.go
```diff
@@ -858,7 +858,7 @@ func ingestTargetLevel(
 //     local sstables. This is the step where overlap with memtables is
 //     determined. If there is overlap, we remember the most recent memtable
 //     that overlaps.
-//  6. Update the sequence number in the ingested local sstables. (Shared
```
itsbilal, line 861: I think this will stay Shared, because external sstables are assigned sequence numbers the same way as local sstables.

ingest.go
```diff
@@ -1295,7 +1295,7 @@ func (d *DB) ingest(
 		}
 	}
 
-	// NB: Shared-sstable-only ingestions do not assign a sequence number to
+	// NB: Remote-sstable-only ingestions do not assign a sequence number to
```
itsbilal, line 1298: Same here, this stays Shared.

itsbilal (approved):
LGTM

RaduBerinde:
Thanks, fixed. Let me know if/when I can merge.

PR #2755, opened 2023-07-19
db: minor update of comment for TargetByteDeletionRate

RaduBerinde:


jbowens (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2755)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`options.go` line 887 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2755#-N_jiEkR0T7W2atVYeB_:-N_jiEkR0T7W2atVYeBa:b-dhd3qp) ([raw file](https://github.com/cockroachdb/pebble/blob/8f415b224e4f4c6b6ce4991b909bb8dc9f882b76/options.go#L887)):*
> ```Go
> 	//
> 	// This value is only a best-effort target; the effective rate can be
> 	// higher if deletions are falling behind or disk space is running low.
> ```

After your changes, it can also be higher if the LSM is deleting files at a sustained rate higher than `TargetBytesDeletionRate`, right?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2755)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`options.go` line 887 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2755#-N_jiEkR0T7W2atVYeB_:-N_jibYtDPdqABnqQMzZ:b83a94q) ([raw file](https://github.com/cockroachdb/pebble/blob/8f415b224e4f4c6b6ce4991b909bb8dc9f882b76/options.go#L887)):*

After your changes, it can also be higher if the LSM is deleting files at a sustained rate higher than `TargetBytesDeletionRate`, right?
</blockquote></details>

Yeah, I consider that under "falling behind" umbrella

PR #2758, opened 2023-07-20
objstorage: rename objstorage/shared package to remote and rename SharedStorage to RemoteStorage in options

RaduBerinde:
#### db: rename Experimental.SharedStorage to RemoteStorage in Options

---

There will be some upcoming work inside the objstorage provider to rename things and separate the aspect of supporting remote storage vs supporting shared storage. We should allow external files even if shared storage is not enabled / we don't have a creator ID.

### DIFF
--- a/cmd/pebble/db.go
+++ b/cmd/pebble/db.go
@@ -101,7 +101,7 @@ func newPebbleDB(dir string) DB {
 	}
 
 	if pathToLocalSharedStorage != "" {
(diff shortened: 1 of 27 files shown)


RaduBerinde:
When this merges, I will bump Pebble on master and make the necessary updates. I will then post more changes renaming stuff inside objstorage.

itsbilal (approved):
LGTM, thanks for the cleanup!

RaduBerinde:
TFTR!

Issue #2540, opened 2023-05-19
db: pipeline WAL rotation


jbowens:
Let _L_ be the fsync latency of the WAL storage medium.

When the memtable and WAL are rotated, the first batch application to the new WAL may need to at worst wait:
1. For an inflight fsync of entries to the previous WAL to complete (at worst, _L_).
2. For a final fsync of entries to the previous WAL that did not make the in-flight fsync. ( _L_ )
3. A final fsync in `LogWriter.Close` to ensure the EOF trailer is synced. ( _L_ )
4. A fsync of the WAL directory to ensure the new WAL is durably linked into its new name. ( _L_ )
5. The fsync of this new batch. ( _L_  )

Cumulatively, these can cause commit tail latencies to increase 5x. There are a few ways this could be reduced.

(2) & (3) could be together bounded by 1 _L_ through more coordination between `LogWriter.Close` and the `LogWriter`'s flush loop. The final flush of log entries (2) can include the EOF trailer and sync:

https://github.com/cockroachdb/pebble/blob/f6eaf9a696e6344af4660b2ac7e30e70539ac2f5/record/log_writer.go#L638-L645

(4) & (5) could happen in parallel, but it would require some additional, delicate synchronization. 

Or alternatively we could prepare the next WAL ahead of time. In a steady state, Pebble would have two open WALs with log numbers `>= minUnflushedLogNum`: _current_ and _next_. The _next_ LogWriter's `flushLoop` would synchronize with _current_'s `Close`, refusing to signal to waiting syncQueuers until _current_'s `Close` has completed. By addressing (2) & (3) as well, this would eliminate any additional worst-case fsync latency from the WAL rotation itself, making it inline with ordinary WAL fsyncs.

In `Open`, we would need to relax/rework the `strictWALTail` option. Currently all replayed WALs besides the most recent one are required to have clean tails indicating that they were deliberately closed—anything else is interpreted as corruption. With this change, it would be possible for the second most recent WAL to have an unclean tail for some time. We could include a marker entry in the _next_ WAL that is written only once after the _next_ WAL observed that _current_'s `Close` completed, indicating that if recovery observed an unclean tail of the previous WAL, it should treat it as corruption.

Jira issue: PEBBLE-192

jbowens:
[...]

PR #2762, opened 2023-07-20
record: allow unbounded queued blocks in LogWriter

jbowens:
Previously, a LogWriter would allocate up to 16 blocks of 32 KiB for buffering WAL writes. If all 16 blocks had been allocated and no free blocks were available, a batch writing to the WAL would queue until the flushing goroutine freed blocks. In testing of write-heavy workloads, especially with larger value sizes, we've seen queueing at the LogWriter. This queueing blocks the commit pipeline, preventing any batches from committing regardless of priority and whether they require waiting for fsync.

This commit modifies LogWriter to allow the queueing of an unbounded number of blocks. In practice, for the current WAL, the memtable size serves as an upper bound. With a 64 MiB memtable, at most 64 MiB / 32 KiB = 2,048 blocks may queue. This is not an unreasonable of additional memory overhead for a write-heavy workload.

Beyond improving throughput for write-heavy workloads, removing this hard bound improves tolerance of momentary fsync stalls.

Informs cockroachdb/cockroach#88699.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -319,7 +319,11 @@ type BatchCommitStats struct {
 	// commitPipeline.Commit.
 	SemaphoreWaitDuration time.Duration
 	// WALQueueWaitDuration is the wait time for allocating memory blocks in the
-	// LogWriter (due to the LogWriter not writing fast enough).
+	// LogWriter (due to the LogWriter not writing fast enough). At the moment
+	// this is duration is always zero because a single WAL will allow
+	// allocating memory blocks up to the entire memtable size. In the future,
+	// we may pipeline WALs and bound the WAL queued blocks separately, so this
+	// field is preserved for that possibility.
 	WALQueueWaitDuration time.Duration
 	// MemTableWriteStallDuration is the wait caused by a write stall due to too
 	// many memtables (due to not flushing fast enough).
--- a/batch_test.go
+++ b/batch_test.go
@@ -23,7 +23,6 @@ import (
 	"github.com/cockroachdb/pebble/internal/batchskl"
 	"github.com/cockroachdb/pebble/internal/keyspan"
 	"github.com/cockroachdb/pebble/internal/testkeys"
-	"github.com/cockroachdb/pebble/record"
 	"github.com/cockroachdb/pebble/vfs"
 	"github.com/stretchr/testify/require"
 )
@@ -1233,28 +1232,6 @@ func TestBatchCommitStats(t *testing.T) {
 			}
 		}
 
-		// WAL queue stall funcs.
-		//
-		// The LogWriter gets changed when stalling/unstalling the memtable, so we
-		// need to use a hook to tell us about the latest LogWriter.
-		var unstallWALQueue func()
-		stallWALQueue := func() {
-			var unstallLatestWALQueue func()
-			db.mu.Lock()
-			defer db.mu.Unlock()
-			db.mu.log.registerLogWriterForTesting = func(w *record.LogWriter) {
-				// db.mu will be held when this is called.
-				unstallLatestWALQueue = w.ReserveAllFreeBlocksForTesting()
-			}
-			db.mu.log.registerLogWriterForTesting(db.mu.log.LogWriter)
-			unstallWALQueue = func() {
-				db.mu.Lock()
-				defer db.mu.Unlock()
-				db.mu.log.registerLogWriterForTesting = nil
-				unstallLatestWALQueue()
-			}
-		}
-
 		// Commit wait stall funcs.
 		var unstallCommitWait func()
 		stallCommitWait := func() {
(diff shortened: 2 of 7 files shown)


jbowens (commented):
We might also consider adding additional memory accounting for these blocks. I don't think it's urgent, given the 64 MiB bound. But if in the future we allow pipelining WALs and permit buffering writes across the WAL rotation (which I think would help tail latencies considerably), I think we'll need some kind of accounting.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2762#-:-N_tF8Ii8DRbvWm3cyFJ:b-yuxkbb)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @sumeerbhola)

sumeerbhola (changes requested):
Is it worth adding some commentary to https://github.com/cockroachdb/pebble/issues/2540?

Can you add a test that does allocate a large number of blocks, so we can be sure that (a) it is correct, (b) the slices are behaving reasonably in that we are not incurring slice allocations where we don't expect to.

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2762#-:-Na2ag9G6kHO9IcWfMup:b-xheya)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @jbowens)

___
*[`batch.go` line 323 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2762#-Na2aqkI7QqwyNELAVdi:-Na2aqkI7QqwyNELAVdj:b-m6qxdv) ([raw file](https://github.com/cockroachdb/pebble/blob/fb76a24ac0984653275fbe9a3d7afb7f3476d844/batch.go#L323)):*
> ```Go
> 	// WALQueueWaitDuration is the wait time for allocating memory blocks in the
> 	// LogWriter (due to the LogWriter not writing fast enough).
> 	WALQueueWaitDuration time.Duration
> ```

will this come back once we add some memory bound?
I am wondering whether we want to keep this for now given that we have the plumbing for it in CockroachDB already.
___
*[`record/log_writer.go` line 569 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2762#-Na2drpHCiJlpMjQBzeQ:-Na2drpHCiJlpMjQBzeR:b-dd9phd) ([raw file](https://github.com/cockroachdb/pebble/blob/3a8381951b7e36a39633e6227d155ec4a64e61b7/record/log_writer.go#L569)):*
> ```Go
> 	w.free.Lock()
> 	if len(w.free.blocks) == 0 {
> 		w.free.blocks = append(w.free.blocks, &block{})
> ```

should we consider allocating the free blocks from a `sync.Pool`, given the well-defined lifecycle and the larger number of blocks a `logWriter` may allocate? That is, two levels of free pool management -- what we have here, and then the `sync.Pool` for `&block{}` and return them all to the `sync.Pool` in `LogWriter.Close`.

jbowens (commented):
Added a comment to #2540 and a unit test + benchmark.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2762#-:-Na7a9X520omof365wQF:bt9ola6)* status: 3 of 6 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`batch.go` line 323 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2762#-Na2aqkI7QqwyNELAVdi:-Na7Cjh-56dyj55InTNq:bpadcw4) ([raw file](https://github.com/cockroachdb/pebble/blob/fb76a24ac0984653275fbe9a3d7afb7f3476d844/batch.go#L323)):*

will this come back once we add some memory bound?
I am wondering whether we want to keep this for now given that we have the plumbing for it in CockroachDB already.
</blockquote></details>

Yeah, I started to wonder whether we ultimately should only have `MemtableStopWritesThreshold` which will also bound on the queued WAL blocks. But I'll keep this around so we can make a decision then.
___
*[`record/log_writer.go` line 569 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2762#-Na2drpHCiJlpMjQBzeQ:-Na7ZWLc3cFE0FryNgHe:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/3a8381951b7e36a39633e6227d155ec4a64e61b7/record/log_writer.go#L569)):*

should we consider allocating the free blocks from a `sync.Pool`, given the well-defined lifecycle and the larger number of blocks a `logWriter` may allocate? That is, two levels of free pool management -- what we have here, and then the `sync.Pool` for `&block{}` and return them all to the `sync.Pool` in `LogWriter.Close`.
</blockquote></details>

Done.

sumeerbhola (approved):
:lgtm: 

> Reviewed 4 of 4 files at r3, 1 of 1 files at r4, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2762#-:-Na7shBG5ohihQk5jEQj:bg7igkv)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @jbowens)

jbowens:
TFTR!

PR #2769, opened 2023-07-24
objstorage: rename 'shared' to 'remote'

RaduBerinde:
Shared objects refer to objects that have been created by a Pebble instance and are owned by the Pebble instances in a cluster. Remote objects can be shared objects or external objects (such as those attached by online restore).

In this commit we rename uses of "shared" to "remote" in objstorage provider code that applies to all types of remote objects. We now conceptually have a remote subsystem which itself contains a shared subsystem.

### DIFF
--- a/db.go
+++ b/db.go
@@ -2561,7 +2561,7 @@ func firstError(err0, err1 error) error {
 }
 
 // SetCreatorID sets the CreatorID which is needed in order to use shared objects.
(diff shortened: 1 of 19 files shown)


itsbilal (approved):
LGTM :shipit:

RaduBerinde:
TFTR!

Issue #2646, opened 2023-06-16
db: investigate elevated batch commit tail latencies


jbowens:
The following graphs are from one particular instance of this, but we've seen it across several test clusters including the 23.1 test cluster.

![Screenshot 2023-06-16 at 4 27 01 PM](https://github.com/cockroachdb/pebble/assets/867352/9c0b8971-25b7-4778-9acf-3ce1ad9c9921)
![Screenshot 2023-06-16 at 4 27 34 PM](https://github.com/cockroachdb/pebble/assets/867352/058179bf-d97d-4912-8c80-7f9873686882)

Internal discussion: https://cockroachlabs.slack.com/archives/C057ULDSKC0/p1686947338066129?thread_ts=1686924156.296339&cid=C057ULDSKC0

jbowens:
(Potentially) related to:

* https://github.com/cockroachdb/pebble/issues/2540
* https://github.com/cockroachdb/pebble/issues/1997

Also, ingests can be disruptive to batch commit tail latencies, although it is definitely not the only factor because we've seen elevated batch commit tail latencies in the absence of ingests.:
* https://github.com/cockroachdb/pebble/issues/2487
* https://github.com/cockroachdb/pebble/issues/2266
* https://github.com/cockroachdb/pebble/issues/2112

jbowens:
As far as I can tell, there are two primary sources of tail latency:

1. There's contention on `d.commit.mu`. Some batch commits block on acquiring `d.commit.mu` for 90ms:
```
slow batch commit: ‹(TotDur: 93.940495ms, SemaWait: 206ns, WALQueueWait: 0s, MemtblStall: 0s, L0Stall: 0s, WALRot: 0s, CommitWait: 489.115µs, MemtblApply: 10.924µs, Publish: 489.572µs, CommitMu: 93.436399ms, WALWrite: 1.094µs, DBMu: 124ns, PendingEnqueue: 308ns, RoomForWrite: 376ns, MemtblRot: 0s, MemtblPrep: 219ns)›
```
This seems to be driven by slow `DB.makeRoomForWrite` calls, which tend to be dominated by the memtable rotation:
```
[...]

PR #2772, opened 2023-07-25
db: keep up to one memtable for recycling

jbowens:
We've observed large allocations like the 64MB memtable allocation take 10ms+.
This can add latency to the WAL/memtable rotation critical section during which
the entire commit pipeline is stalled, contributing to batch commit tail
latencies. This commit adapts the memtable lifecycle to keep the most recent
obsolete memtable around for use as the next mutable memtable.

This reduces the commit latency hiccup during a memtable rotation, and it also
reduces block cache mutex contention (https://github.com/cockroachdb/pebble/issues/1997) by reducing the number of times we
must reserve memory from the block cache.

```
goos: linux
goarch: amd64
cpu: Intel(R) Xeon(R) CPU @ 2.30GHz
                   │   old.txt   │               new.txt               │
                   │   sec/op    │   sec/op     vs base                │
RotateMemtables-24   120.7µ ± 2%   102.8µ ± 4%  -14.85% (p=0.000 n=25)

                   │   old.txt    │               new.txt               │
                   │     B/op     │     B/op      vs base               │
RotateMemtables-24   124.3Ki ± 0%   124.0Ki ± 0%  -0.27% (p=0.000 n=25)

                   │  old.txt   │              new.txt              │
                   │ allocs/op  │ allocs/op   vs base               │
RotateMemtables-24   114.0 ± 0%   111.0 ± 0%  -2.63% (p=0.000 n=25)
```

Informs https://github.com/cockroachdb/pebble/issues/2646.

### DIFF
--- a/db.go
+++ b/db.go
@@ -253,9 +253,14 @@ func (d defaultCPUWorkGranter) CPUWorkDone(_ CPUWorkHandle) {}
 type DB struct {
 	// The count and size of referenced memtables. This includes memtables
 	// present in DB.mu.mem.queue, as well as memtables that have been flushed
-	// but are still referenced by an inuse readState.
+	// but are still referenced by an inuse readState, as well as up to one
+	// memTable waiting to be reused and stored in d.memTableRecycle.
 	memTableCount    atomic.Int64
 	memTableReserved atomic.Int64 // number of bytes reserved in the cache for memtables
+	// memTableRecycle holds a pointer to an obsolete memtable. The next
+	// memtable allocation will reuse this memtable if it has not already been
+	// recycled.
+	memTableRecycle atomic.Pointer[memTable]
 
 	// The size of the current log file (i.e. db.mu.log.queue[len(queue)-1].
 	logSize atomic.Uint64
@@ -1540,6 +1545,10 @@ func (d *DB) Close() error {
 		// replay.
 		mem.readerUnrefLocked(false)
 	}
+	// If there's an unused, recycled memtable, we need to release its memory.
+	if obsoleteMemTable := d.memTableRecycle.Swap(nil); obsoleteMemTable != nil {
+		d.freeMemTable(obsoleteMemTable)
+	}
 	if reserved := d.memTableReserved.Load(); reserved != 0 {
 		err = firstError(err, errors.Errorf("leaked memtable reservation: %d", errors.Safe(reserved)))
 	}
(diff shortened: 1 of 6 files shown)


jbowens (commented):
I updated this to recycle the entire `memTable` struct. This helps also reduce contention on the block cache, since we don't need to `Reserve` memory from all the block cache shards on every memtable rotation.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2772#-:-NaNH3j1Ba3-CT5VF-gr:b-e0vdnb)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`-- commits` line 21 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2772#-NaHv4STCtlTOGfRP3wR):*

why is this not showing improvement in allocation? Is this because of cgo, so these allocations are being hidden from the go runtime?
</blockquote></details>

Yeah, that's right; the allocation reduction is only in cgo allocations.
___
*[`db.go` line 2229 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2772#-NaHusHHB1QM42XeOqBF:-NaNH0ZM3lkqn9k9tkxj:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/bb8f9e507285fee2af2a40b142e9c225c5abe66e/db.go#L2229)):*

s/not/now/
</blockquote></details>

Done.

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2772#-:-NaRpBwlELuC5f6ccOhO:bg7igkv)* status: all files reviewed, 3 unresolved discussions (waiting on @bananabrick and @jbowens)

___
*[`db.go` line 2244 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2772#-NaRnKKkCb8C3Ck4O-8F:-NaRnKKkCb8C3Ck4O-8G:b7pfapm) ([raw file](https://github.com/cockroachdb/pebble/blob/d7f0ff999e6014a8d0dc34ee82c3cfca0afad672/db.go#L2244)):*
> ```Go
> 
> func (d *DB) freeMemTable(m *memTable) {
> 	d.memTableCount.Add(-1)
> ```

Can you update the code commentary around the `ZombieCount` and `ZombieSize` metrics to clarify that they include up to 1 memtable that is saved for reducing allocations.

I don't think we export CRDB metrics for these.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2772#-:-NaS-iQ82X_QahVR5nc6:b-814g8k)* status: 3 of 6 files reviewed, 3 unresolved discussions (waiting on @bananabrick and @sumeerbhola)

___
*[`db.go` line 2244 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2772#-NaRnKKkCb8C3Ck4O-8F:-NaS-kMW58NSGLECdevr:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/d7f0ff999e6014a8d0dc34ee82c3cfca0afad672/db.go#L2244)):*

Can you update the code commentary around the `ZombieCount` and `ZombieSize` metrics to clarify that they include up to 1 memtable that is saved for reducing allocations.

I don't think we export CRDB metrics for these.
</blockquote></details>

Done.

PR #2773, opened 2023-07-26
db: introduce Download API

RaduBerinde:
This commit introduces a `DB.Download()` API which is used during online restore to copy off external data.


itsbilal (approved):
LGTM, though I'll let DR chime in if they want this to be more explicitly tied to notions of progress.

RaduBerinde:
Thanks!

PR #2774, opened 2023-07-26
internal/cache: reduce mutex contention

jbowens:
**internal/cache: reduce mutex contention from EvictFile**

We've observed block cache mutex contention originating from EvictFile. When
evicting a file with a significant count of blocks currently held within the
block cache, EvictFile may hold the block cache shard mutex for an extended
duration, increasing latency for requests that must access the same shard.

This commit mitigates this contention with two approaches. EvictFile now
periodically drops the mutex to provide these other requests with an
opportunity to make progress. Additionally, rather than free()-ing manually
managed memory back to the memory allocator while holding the mutex, the free
is deferred until the shard mutex has been dropped.

Informs #1997.

**internal/cache: increase shard count to 4x CPUs**

Increase the block cache shard count to reduce mutex contention.

### DIFF
--- a/internal/cache/clockpro.go
+++ b/internal/cache/clockpro.go
@@ -627,7 +627,7 @@ type Metrics struct {
 // Cache implements Pebble's sharded block cache. The Clock-PRO algorithm is
 // used for page replacement
 // (http://static.usenix.org/event/usenix05/tech/general/full_papers/jiang/jiang_html/html.html). In
-// order to provide better concurrency, 2 x NumCPUs shards are created, with
+// order to provide better concurrency, 4 x NumCPUs shards are created, with
 // each shard being given 1/n of the target cache size. The Clock-PRO algorithm
 // is run independently on each shard.
 //
@@ -683,7 +683,34 @@ type Cache struct {
 //	defer c.Unref()
 //	d, err := pebble.Open(pebble.Options{Cache: c})
 func New(size int64) *Cache {
-	return newShards(size, 2*runtime.GOMAXPROCS(0))
+	// How many cache shards should we create?
+	//
+	// Note that the probability two processors will try to access the same
+	// shard at the same time increases superlinearly with the number of
+	// processors (Eg, consider the brithday problem where each CPU is a person,
+	// and each shard is a possible birthday).
+	//
+	// We could consider growing the number of shards superlinearly, but
+	// increasing the shard count may reduce the effectiveness of the caching
+	// algorithm if frequently-accessed blocks are insufficiently distributed
+	// across shards. If a shard's size is smaller than a single frequently
+	// scanned sstable, then the shard will be unable to hold the entire
+	// frequently-scanned table in memory despite other shards still holding
+	// infrequently accessed blocks.
+	//
+	// Experimentally, we've observed contention contributing to tail latencies
+	// at 2 shards per processor. For now we use 4 shards per processor,
+	// recognizing this may not be final word.
+	m := 4 * runtime.GOMAXPROCS(0)
+
+	// In tests we can use large CPU machines with small cache sizes and have
+	// many caches in existence at a time. If sharding into m shards would
+	// produce too small shards, constrain the number of shards to 4.
+	const minimumShardSize = 4 << 20 // 4 MiB
+	if m > 4 && int(size)/m < minimumShardSize {
+		m = 4
+	}
+	return newShards(size, m)
 }
 
 func newShards(size int64, shards int) *Cache {
--- a/internal/cache/clockpro.go
+++ b/internal/cache/clockpro.go
@@ -71,9 +71,7 @@ func (h Handle) Get() []byte {
 
 // Release releases the reference to the cache entry.
 func (h Handle) Release() {
-	if h.value != nil {
-		h.value.release()
-	}
+	h.value.release()
 }
 
 type shard struct {
@@ -176,7 +174,7 @@ func (c *shard) Set(id uint64, fileNum base.DiskFileNum, offset uint64, value *V
 		// cache entry was a test page
 		c.sizeTest -= e.size
 		c.countTest--
-		c.metaDel(e)
+		c.metaDel(e).release()
 		c.metaCheck(e)
 
 		e.size = int64(len(value.buf))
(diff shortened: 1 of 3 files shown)


RaduBerinde (approved):
:lgtm: Just some nits.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2774#-:-NaNCrqs6uoRX35Kujzt:b-n6a9df)* status: 0 of 1 files reviewed, 6 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/cache/clockpro.go` line 180 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaNAdkK8FEabMVQlzYu:-NaNAdkK8FEabMVQlzYv:bq3c3m0) ([raw file](https://github.com/cockroachdb/pebble/blob/d27fa522e86a922094e84a11edf57d44a6bba8e9/internal/cache/clockpro.go#L180)):*
> ```Go
> 		c.countTest--
> 		if oldV := c.metaDel(e); oldV != nil {
> 			oldV.release()
> ```

`oldV` should always be `nil` here right? (that would be the case above)
Ok to keep so it's idiomatic but perhaps add a comment.
___
*[`internal/cache/clockpro.go` line 259 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaNB8tG0jIasdKvQcS7:-NaNB8tG0jIasdKvQcS8:bnp9u3j) ([raw file](https://github.com/cockroachdb/pebble/blob/d27fa522e86a922094e84a11edf57d44a6bba8e9/internal/cache/clockpro.go#L259)):*
> ```Go
> 	fkey := key{fileKey{id, fileNum}, 0}
> 	for c.evictFileRun(fkey) {
> 	}
> ```

maybe call `runtime.Gosched()` here so other goroutines get more of a chance to acquire the mutex?
___
*[`internal/cache/clockpro.go` line 264 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaNBvuxEaQ5yblaWf6S:-NaNBvuxEaQ5yblaWf6T:bmza653) ([raw file](https://github.com/cockroachdb/pebble/blob/7a4ed3c57cc5e5c3183845029beac00b7dfd18a4/internal/cache/clockpro.go#L264)):*
> ```Go
> 			break
> 		}
> 	}
> ```

Wouldn't it be a lot simpler to just add a
```
mu.Unlock()
runtime.GoSched()
mu.Lock()
```
in here every 5 blocks?
___
*[`internal/cache/clockpro.go` line 399 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaNCFfE0o75VSwkvI-j:-NaNCFfE0o75VSwkvI-k:b-49jx4f) ([raw file](https://github.com/cockroachdb/pebble/blob/d27fa522e86a922094e84a11edf57d44a6bba8e9/internal/cache/clockpro.go#L399)):*
> ```Go
> // Remove the entry from the cache. This removes the entry from the blocks map,
> // the files map, and ensures that hand{Hot,Cold,Test} are not pointing at the
> // entry. Returns the deleted value that must be released.
> ```

Can also return `nil`.. Or we could make `release` tolerate `nil` and then we can just call `metaDel().release()` in many cases
___
*[`internal/cache/clockpro.go` line 405 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaNAHlb3C70IvGfhUxB:-NaNAHlb3C70IvGfhUxC:b9z7ldc) ([raw file](https://github.com/cockroachdb/pebble/blob/d27fa522e86a922094e84a11edf57d44a6bba8e9/internal/cache/clockpro.go#L405)):*
> ```Go
> 	}
> 	// Remove the pointer to the value.
> 	deletedValue, e.val = e.val, nil
> ```

[nit] I find this harder to read than two separate statement
___
*[`internal/cache/clockpro.go` line 710 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaNCiM01I28FANiO8C2:-NaNCiM01I28FANiO8C3:b4qd0i8) ([raw file](https://github.com/cockroachdb/pebble/blob/84588ca5f74a8959f463b1050e7dfa2af240fa06/internal/cache/clockpro.go#L710)):*
> ```Go
> 	// In tests, we can use large CPU machines with small cache sizes and have
> 	// many caches in existence at a time. If sharding into m shards would
> 	// produce small of shards, constrain the number of shards to 4.
> ```

[nit] would produce shards that are too small

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2774#-:-NaNWUlH97x3rafKCQLN:b-814g8k)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/cache/clockpro.go` line 259 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaNB8tG0jIasdKvQcS7:-NaNQ_B02LZVMZGQFQAN:b-48vilg) ([raw file](https://github.com/cockroachdb/pebble/blob/d27fa522e86a922094e84a11edf57d44a6bba8e9/internal/cache/clockpro.go#L259)):*

maybe call `runtime.Gosched()` here so other goroutines get more of a chance to acquire the mutex?
</blockquote></details>

Good call, done.
___
*[`internal/cache/clockpro.go` line 264 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaNBvuxEaQ5yblaWf6S:-NaNRjMuBmjjKydDHTV1:ba8igxz) ([raw file](https://github.com/cockroachdb/pebble/blob/7a4ed3c57cc5e5c3183845029beac00b7dfd18a4/internal/cache/clockpro.go#L264)):*

Wouldn't it be a lot simpler to just add a
```
mu.Unlock()
runtime.GoSched()
mu.Lock()
```
in here every 5 blocks?
</blockquote></details>

Freeing the memory while the mutex is dropped and adding a `defer` to handle invariant violation panics while `Free`-ing memory make it not too simple. But I agree performing the frees and consistency checks in the defer was confusing. I moved it it inline here, but it's still not very simple :/
___
*[`internal/cache/clockpro.go` line 399 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaNCFfE0o75VSwkvI-j:-NaNRTJABoGSzlxy9Ojh:b-86mr6o) ([raw file](https://github.com/cockroachdb/pebble/blob/d27fa522e86a922094e84a11edf57d44a6bba8e9/internal/cache/clockpro.go#L399)):*

Can also return `nil`.. Or we could make `release` tolerate `nil` and then we can just call `metaDel().release()` in many cases
</blockquote></details>

I like it, done.

RaduBerinde:
*[`internal/cache/clockpro.go` line 264 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaNBvuxEaQ5yblaWf6S:-NaNi7cW3qJBCJ_cwrrv:bapuf14) ([raw file](https://github.com/cockroachdb/pebble/blob/7a4ed3c57cc5e5c3183845029beac00b7dfd18a4/internal/cache/clockpro.go#L264)):*

Freeing the memory while the mutex is dropped and adding a `defer` to handle invariant violation panics while `Free`-ing memory make it not too simple. But I agree performing the frees and consistency checks in the defer was confusing. I moved it it inline here, but it's still not very simple :/
</blockquote></details>

Ah.. right.. Another option would be for the function to return the slice and have the caller release. That would keep the entire function locked. We can pass in the prealloced slice and the function woulf fill it to the cap.

Also, I don't think we need to checkConsistency in the first `return false` case, so we could just move it before the second `return false`

sumeerbhola (approved):
:lgtm: 

> Reviewed 3 of 3 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2774#-:-NaRv2RG-Krbr-d-y1iz:bg7igkv)* status: all files reviewed, 8 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/cache/clockpro.go` line 264 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaRqccP4m8uWogsgAcR:-NaRqccP4m8uWogsgAcS:b-psl0yp) ([raw file](https://github.com/cockroachdb/pebble/blob/37e934206e720c6744d321202e581208c0e836ef/internal/cache/clockpro.go#L264)):*
> ```Go
> 	// If most of the file's blocks are held in the block cache, evicting all
> 	// the blocks may take a while. We don't want to block the entire cache
> 	// shard, forcing concurrent readers until we're finished. We drop the mutex
> ```

... readers to wait until ...
___
*[`internal/cache/clockpro.go` line 277 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaRsOTB-5YkfezLr-Xe:-NaRsOTB-5YkfezLr-Xf:b-p54tbx) ([raw file](https://github.com/cockroachdb/pebble/blob/37e934206e720c6744d321202e581208c0e836ef/internal/cache/clockpro.go#L277)):*
> ```Go
> 	obsoleteValues := obsoleteValuesAlloc[:0]
> 	defer func() {
> 		if !moreRemaining {
> ```

maybe add a comment
// Don't want to pay the cost of checkConsistency for each batch corresponding to blocksPerMutexAcquisition, so do it only when this was the last batch processed.
___
*[`internal/cache/clockpro.go` line 289 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaRsjp1-GV-AKY7oxR_:-NaRsjp1-GV-AKY7oxRa:b-9ov0ea) ([raw file](https://github.com/cockroachdb/pebble/blob/37e934206e720c6744d321202e581208c0e836ef/internal/cache/clockpro.go#L289)):*
> ```Go
> 	if blocks == nil {
> 		return false
> 	}
> ```

This code could use a few comments. Something like:

// b is the current head of the doubly linked list, and n is the entry after b. 
___
*[`internal/cache/clockpro.go` line 293 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaRt59ZD0QTvQ0Iw6kP:-NaRt59ZD0QTvQ0Iw6kQ:bevrugq) ([raw file](https://github.com/cockroachdb/pebble/blob/37e934206e720c6744d321202e581208c0e836ef/internal/cache/clockpro.go#L293)):*
> ```Go
> 		n = b.fileLink.next
> 		obsoleteValues = append(obsoleteValues, c.metaEvict(b))
> 		if b == n {
> ```

// b == n represents the case where b was the last entry remaining in the doubly linked list, which is why it pointed at itself. So no more entries left.
___
*[`internal/cache/clockpro.go` line 704 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaRtxmo4IKAdmgbc1_d:-NaRtxmo4IKAdmgbc1_e:bxni7ql) ([raw file](https://github.com/cockroachdb/pebble/blob/53839eef5473e509e966b9e64c4371f30285fb2c/internal/cache/clockpro.go#L704)):*
> ```Go
> 	// In tests we can use large CPU machines with small cache sizes and have
> 	// many caches in existence at a time. If sharding into m shards would
> 	// produce too small of shards, constrain the number of shards to 4.
> ```

too small shards ...

jbowens (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2774#-:-NaS77JtBBabeuEW_7ok:bz5mair)* status: 2 of 3 files reviewed, 8 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/cache/clockpro.go` line 289 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaRsjp1-GV-AKY7oxR_:-NaS1h_s9lHCzozUKeD2:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/37e934206e720c6744d321202e581208c0e836ef/internal/cache/clockpro.go#L289)):*

This code could use a few comments. Something like:

// b is the current head of the doubly linked list, and n is the entry after b. 
</blockquote></details>

Done.
___
*[`internal/cache/clockpro.go` line 293 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaRt59ZD0QTvQ0Iw6kP:-NaS2Jmg4G19X69G_d6O:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/37e934206e720c6744d321202e581208c0e836ef/internal/cache/clockpro.go#L293)):*

// b == n represents the case where b was the last entry remaining in the doubly linked list, which is why it pointed at itself. So no more entries left.
</blockquote></details>

Done.
___
*[`internal/cache/clockpro.go` line 704 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2774#-NaRtxmo4IKAdmgbc1_d:-NaS2L19B8odxmSRcM2u:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/53839eef5473e509e966b9e64c4371f30285fb2c/internal/cache/clockpro.go#L704)):*

too small shards ...
</blockquote></details>

Done.

Issue #2112, opened 2022-11-09
db: reduce Ingest I/O with manifest+db locks


jbowens:
To identify the target level of M ingested files in a LSM with N levels, we may need to seek in O(N × M) sstables to look for data overlap. This is all performed during application while holding `DB.mu` and the manifest lock, which prevents concurrent flushes and compactions from scheduling or completing while the mutex is held.

Ingest performs these data overlap checks while holding the manifest lock for a consistent view of the LSM. However, for the purposes of detecting data overlap, ingest could be optimistic about its view of the LSM.

When Ingest acquires the manifest lock during the apply step, Ingest can exclude any sstables or memtables with largest sequence numbers already observed during data overlap calculations. Ingest only needs to search in any new sstables for more recent overlap introduced since its optimistic scan.

Related to #25 and [this TODO to reduce overlap checks by using L0 sublevels](https://github.com/cockroachdb/pebble/blob/7b30bd86ff65c3edadda1fea4f1f34f1106193aa/ingest.go#L535).

Also related to #1683 which will remove the strict requirement of preventing file boundary overlap, making target level a function of data overlap exclusively.

Jira issue: PEBBLE-173

jbowens:
This may be used to improve the estimate of `ApproxIngestedIntoL0Bytes` for flushable ingests. The precomputed data overlap can be used to determine which sstables overlapped L0 before entering the commit pipeline.

jbowens:
![Screenshot 2023-05-12 at 3 33 53 PM](https://github.com/cockroachdb/pebble/assets/867352/c1ba4063-953e-4783-914f-2045d9b5a2cf)

I suspect in these graphs we see ingestions holding up the commit pipeline, causing a divergence between the WAL fsync latency and the raft log commit latency.

RaduBerinde:
This is my current thinking:
 - refactor the code to stop using `levelIter`. We can just find the overlapping files and check for overlap with those.
 - store an "overlap cache" in each `FileMetadata`. The overlap cache consists of a small number of non-overlapping `UserKeyBounds`; each one is either a known empty region of the file, or a known "full" region of the file (either just a point, or a range corresponding to a range del or range key).
 - when we check with overlap with a `FileMetadata`, we first take a look at the overlap cache and see if it can give us an answer (i.e. if we are completely inside of a "known empty" region, or if we intersect with any "known full" region). If not, we check for overlap and add one or two regions in the cache (one for any empty region we find, and one for any full region we find).

[...]

PR #2775, opened 2023-07-27
*: implement EventuallyFileOnlySnapshots

### LINKED ISSUE #2740: snapshot: Create new `EventuallyFileOnlySnapshot` that holds readState
`IngestAndExcise` breaks pebble snapshot semantics, as existing open snapshots can see keys get dropped that got excised after the snapshot went in. This would be observed if a new iterator was opened on the same snapshot after said excise, and it tried to read excised sstables; we will either return no keys in the excised span (even if there were keys there at time of snapshot creation), or we will return keys from the `IngestAndExcise` that went into the excise span. Either way, we'd break snapshot guarantees.

Moving to long-lived iterators is an easy albeit non-performant band-aid, as they pin sstables (less bad in a disaggregated storage world) but also memtables for their lifetime. A better solution is to create an `EventuallyFileOnlySnapshot` that records itself like a snapshot, takes in some protected key spans over which it monitors excises, and upon a flush of all keys < snapshotSeqNum it converts to a file-only snapshot by switching to holding a readState (like an iterator, except that we don't need to pin any memtables anymore) instead of having its snapshot in the snapshots list. If an excise of the protected key spans happens between snapshot instantiation and conversion to a file-only snapshot, the snapshot returns an error and expects the caller to try grabbing a snapshot again.

Here's an example interface for how this could look like (thanks @sumeerbhola for coming up with this)

```
// Returns snapshot immediately.
//
// Snapshot grabs the current readState, like an Iterator, *and* records itself as
// a snapshot, to preserve values at its seqnum during compactions.
//
// The eventual nature of this is based on a duration that is configured at the
// DB level, since we want the DB to control when the flush happens, say every
// 5s when there are snapshots waiting to be come file-only snapshots.
//
// The conversion to the file-only snapshot can fail if the key-ranges were
// excised before the conversion. The failure will be seen by NewIter. If the
// conversion succeeds, the snapshot pins the files in the current version
// that correspond to its key ranges and no additional files (so the files
// pinned are minimized, which is helpful when using this without disaggregated
// storage). Additionally, it removes itself as a "snapshot" that affects what
// seqnums need to be preserved by compactions.
DB.NewEventuallyFileOnlySnapshot(key-ranges) (EventuallyFileOnlySnapshot, error)

type EventuallyFileOnlySnapshot interface {
     // NewIter uses the readState if not yet a file-only snapshot. If the
     // conversion to a file-only snapshot fails, it will return an error.
     //
     // The error case will be rare, so when that occurs during a CockroachDB
     // outgoing range snapshot we can afford to fail that outgoing snapshot.
     NewIter(o *IterOptions) (*Iterator, error)

     // Callers who don't want to handle an error and don't mind waiting can
     // call this method before creating their first iterator. e.g. replica
     // consistency checker.
     WaitForFileOnlySnapshot() error
}
```

itsbilal:
This change adds a new type of snapshots, EventuallyFileOnlySnapshots, that only stay in the snapshot list until all memtables with keys less than the snapshot have been flushed. After that, these file-only snapshots switch to holding a Version ref (which is akin to holding a readState ref minus the memtables). This allows for a good balance between the write-amp implications of staying in the snapshot list for too long, and the space-amp concerns with holding onto obsolete files.

Fixes #2740.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -1071,8 +1071,8 @@ func (b *Batch) SetRepr(data []byte) error {
 // The returned Iterator observes all of the Batch's existing mutations, but no
 // later mutations. Its view can be refreshed via RefreshBatchSnapshot or
 // SetOptions().
-func (b *Batch) NewIter(o *IterOptions) *Iterator {
-	return b.NewIterWithContext(context.Background(), o)
+func (b *Batch) NewIter(o *IterOptions) (*Iterator, error) {
+	return b.NewIterWithContext(context.Background(), o), nil
 }
 
 // NewIterWithContext is like NewIter, and additionally accepts a context for
--- a/batch.go
+++ b/batch.go
@@ -1081,7 +1081,7 @@ func (b *Batch) NewIterWithContext(ctx context.Context, o *IterOptions) *Iterato
 	if b.index == nil {
 		return &Iterator{err: ErrNotIndexed}
 	}
-	return b.db.newIter(ctx, b, nil /* snapshot */, o)
+	return b.db.newIter(ctx, b, snapshotIterOpts{}, o)
 }
 
 // newInternalIter creates a new internalIterator that iterates over the
--- a/batch_test.go
+++ b/batch_test.go
@@ -270,7 +270,7 @@ func TestBatchEmpty(t *testing.T) {
 	require.NoError(t, err)
 	defer d.Close()
 	ib := newIndexedBatch(d, DefaultComparer)
-	iter := ib.NewIter(nil)
+	iter, _ := ib.NewIter(nil)
 	require.False(t, iter.First())
 	iter2, err := iter.Clone(CloneOptions{})
 	require.NoError(t, err)
@@ -418,7 +418,7 @@ func TestIndexedBatchReset(t *testing.T) {
 	require.Nil(t, b.rangeKeyIndex)
 
 	count := func(ib *Batch) int {
-		iter := ib.NewIter(nil)
+		iter, _ := ib.NewIter(nil)
 		defer iter.Close()
 		iter2, err := iter.Clone(CloneOptions{})
 		require.NoError(t, err)
@@ -433,7 +433,7 @@ func TestIndexedBatchReset(t *testing.T) {
 		return count[0]
 	}
 	contains := func(ib *Batch, key, value string) bool {
-		iter := ib.NewIter(nil)
+		iter, _ := ib.NewIter(nil)
 		defer iter.Close()
 		iter2, err := iter.Clone(CloneOptions{})
 		require.NoError(t, err)
@@ -497,13 +497,13 @@ func TestIndexedBatchMutation(t *testing.T) {
 			return ""
 		case "new-batch-iter":
 			name := td.CmdArgs[0].String()
-			iters[name] = b.NewIter(&IterOptions{
+			iters[name], _ = b.NewIter(&IterOptions{
 				KeyTypes: IterKeyTypePointsAndRanges,
 			})
 			return ""
 		case "new-db-iter":
 			name := td.CmdArgs[0].String()
-			iters[name] = d.NewIter(&IterOptions{
+			iters[name], _ = d.NewIter(&IterOptions{
 				KeyTypes: IterKeyTypePointsAndRanges,
 			})
 			return ""
@@ -581,7 +581,7 @@ func TestIndexedBatch_GlobalVisibility(t *testing.T) {
 	// Create an iterator over an empty indexed batch.
 	b := newIndexedBatch(d, DefaultComparer)
 	iterOpts := IterOptions{KeyTypes: IterKeyTypePointsAndRanges}
-	iter := b.NewIter(&iterOpts)
+	iter, _ := b.NewIter(&iterOpts)
 	defer iter.Close()
 
 	// Mutate the database's committed state.
@@ -1523,7 +1523,7 @@ func TestBatchSpanCaching(t *testing.T) {
 		case p < .55: /* 45 % */
 			// Create a new iterator directly from the batch and check that it
 			// observes the correct state.
-			iter := b.NewIter(&IterOptions{KeyTypes: IterKeyTypePointsAndRanges})
+			iter, _ := b.NewIter(&IterOptions{KeyTypes: IterKeyTypePointsAndRanges})
 			checkIter(iter, nextWriteKey)
 			iters[nextWriteKey] = append(iters[nextWriteKey], iter)
 		default: /* 45 % */
--- a/checkpoint_test.go
+++ b/checkpoint_test.go
@@ -139,7 +139,7 @@ func TestCheckpoint(t *testing.T) {
 			}
 			memLog.Reset()
 			d := dbs[td.CmdArgs[0].String()]
-			iter := d.NewIter(nil)
+			iter, _ := d.NewIter(nil)
 			for valid := iter.First(); valid; valid = iter.Next() {
 				memLog.Infof("%s %s", iter.Key(), iter.Value())
 			}
@@ -280,7 +280,7 @@ func TestCheckpointFlushWAL(t *testing.T) {
 	{
 		d, err := Open(checkpointPath, opts)
 		require.NoError(t, err)
-		iter := d.NewIter(nil)
+		iter, _ := d.NewIter(nil)
 		require.True(t, iter.First())
 		require.Equal(t, key, iter.Key())
 		require.Equal(t, value, iter.Value())
@@ -335,7 +335,7 @@ func TestCheckpointManyFiles(t *testing.T) {
 	{
 		d, err := Open(checkpointPath, opts)
 		require.NoError(t, err)
-		iter := d.NewIter(nil)
+		iter, _ := d.NewIter(nil)
 		require.True(t, iter.First())
 		require.NoError(t, iter.Error())
 		n := 1
--- a/cmd/pebble/db.go
+++ b/cmd/pebble/db.go
@@ -131,7 +131,8 @@ func (p pebbleDB) Flush() error {
 }
 
 func (p pebbleDB) NewIter(opts *pebble.IterOptions) iterator {
-	return p.d.NewIter(opts)
+	iter, _ := p.d.NewIter(opts)
+	return iter
 }
 
 func (p pebbleDB) NewBatch() batch {
--- a/compaction.go
+++ b/compaction.go
@@ -2137,6 +2137,7 @@ func (d *DB) flush1() (bytesFlushed uint64, err error) {
 				// ranges this snapshot is interested in, this EFOS cannot transition to
 				// a file-only snapshot as keys in that range could now be deleted. Move
 				// onto the next snapshot.
+				s.efos.releaseReadState()
 				s = s.next
 				continue
 			}
--- a/compaction.go
+++ b/compaction.go
@@ -2118,6 +2118,36 @@ func (d *DB) flush1() (bytesFlushed uint64, err error) {
 				d.mu.versions.metrics.Flush.AsIngestTableCount += l.TablesIngested
 			}
 		}
+
+		// Update if any eventually file-only snapshots have now transitioned to
+		// being file-only.
+		earliestUnflushedSeqNum := d.getEarliestUnflushedSeqNumLocked()
+		currentVersion := d.mu.versions.currentVersion()
+		for s := d.mu.snapshots.root.next; s != &d.mu.snapshots.root; {
+			if s.efos == nil {
+				s = s.next
+				continue
+			}
+			if base.Visible(earliestUnflushedSeqNum, s.efos.seqNum, InternalKeySeqNumMax) {
+				s = s.next
+				continue
+			}
+			if s.efos.excised.Load() {
+				// If a concurrent excise has happened that overlaps with one of the key
+				// ranges this snapshot is interested in, this EFOS cannot transition to
+				// a file-only snapshot as keys in that range could now be deleted. Move
+				// onto the next snapshot.
+				s = s.next
+				continue
+			}
+			currentVersion.Ref()
+
+			// NB: s.efos.transitionToFileOnlySnapshot could close s, in which
+			// case s.next would be nil. Save it before calling it.
+			next := s.next
+			_ = s.efos.transitionToFileOnlySnapshot(currentVersion)
+			s = next
+		}
 	}
 	// Signal FlushEnd after installing the new readState. This helps for unit
 	// tests that use the callback to trigger a read using an iterator with
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -1436,7 +1436,7 @@ func TestManualCompaction(t *testing.T) {
 					db:     d,
 					seqNum: InternalKeySeqNumMax,
 				}
-				iter := snap.NewIter(nil)
+				iter, _ := snap.NewIter(nil)
 				return runIterCmd(td, iter, true)
 
 			case "lsm":
@@ -2148,7 +2148,7 @@ func TestCompactionDeleteOnlyHints(t *testing.T) {
 					db:     d,
 					seqNum: InternalKeySeqNumMax,
 				}
-				iter := snap.NewIter(nil)
+				iter, _ := snap.NewIter(nil)
 				return runIterCmd(td, iter, true)
 
 			case "reset":
--- a/db.go
+++ b/db.go
@@ -74,7 +74,7 @@ type Reader interface {
 	// NewIter returns an iterator that is unpositioned (Iterator.Valid() will
 	// return false). The iterator can be positioned via a call to SeekGE,
 	// SeekLT, First or Last.
-	NewIter(o *IterOptions) *Iterator
+	NewIter(o *IterOptions) (*Iterator, error)
 
 	// Close closes the Reader. It may or may not close any underlying io.Reader
 	// or io.Writer, depending on how the DB was created.
@@ -1471,14 +1471,14 @@ func (d *DB) NewIndexedBatch() *Batch {
 // to maintain a long-lived point-in-time view of the DB state can lead to an
 // apparent memory and disk usage leak. Use snapshots (see NewSnapshot) for
 // point-in-time snapshots which avoids these problems.
-func (d *DB) NewIter(o *IterOptions) *Iterator {
+func (d *DB) NewIter(o *IterOptions) (*Iterator, error) {
 	return d.NewIterWithContext(context.Background(), o)
 }
 
 // NewIterWithContext is like NewIter, and additionally accepts a context for
 // tracing.
-func (d *DB) NewIterWithContext(ctx context.Context, o *IterOptions) *Iterator {
-	return d.newIter(ctx, nil /* batch */, snapshotIterOpts{}, o)
+func (d *DB) NewIterWithContext(ctx context.Context, o *IterOptions) (*Iterator, error) {
+	return d.newIter(ctx, nil /* batch */, snapshotIterOpts{}, o), nil
 }
 
 // NewSnapshot returns a point-in-time view of the current DB state. Iterators
--- a/db.go
+++ b/db.go
@@ -980,12 +980,31 @@ var iterAllocPool = sync.Pool{
 	},
 }
 
+// snapshotIterOpts denotes snapshot-related iterator options when calling
+// newIter. These are the possible cases for a snapshotIterOpts:
+//   - No snapshot: All fields are zero values.
+//   - Classic snapshot: Only `seqNum` is set. The latest readState will be used
+//     and the specified seqNum will be used as the snapshot seqNum.
+//   - EventuallyFileOnlySnapshot (EFOS) in pre-file-only state: `readState` and
+//     `seqNum` are set. The specified readState is used with the snapshot
+//     sequence number.
+//   - EFOS in file-only state: Only `seqNum` and `vers` are set. No readState
+//     is referenced, as all the relevant SSTs are referenced by the *version.
+type snapshotIterOpts struct {
+	seqNum    uint64
+	readState *readState
+	vers      *version
+}
+
 // newIter constructs a new iterator, merging in batch iterators as an extra
 // level.
-func (d *DB) newIter(ctx context.Context, batch *Batch, s *Snapshot, o *IterOptions) *Iterator {
+func (d *DB) newIter(
+	ctx context.Context, batch *Batch, sOpts snapshotIterOpts, o *IterOptions,
+) *Iterator {
 	if err := d.closed.Load(); err != nil {
 		panic(err)
 	}
+	seqNum := sOpts.seqNum
 	if o.rangeKeys() {
 		if d.FormatMajorVersion() < FormatRangeKeys {
 			panic(fmt.Sprintf(
@@ -997,7 +1016,7 @@ func (d *DB) newIter(ctx context.Context, batch *Batch, s *Snapshot, o *IterOpti
 	if o != nil && o.RangeKeyMasking.Suffix != nil && o.KeyTypes != IterKeyTypePointsAndRanges {
 		panic("pebble: range key masking requires IterKeyTypePointsAndRanges")
 	}
-	if (batch != nil || s != nil) && (o != nil && o.OnlyReadGuaranteedDurable) {
+	if (batch != nil || seqNum != 0) && (o != nil && o.OnlyReadGuaranteedDurable) {
 		// We could add support for OnlyReadGuaranteedDurable on snapshots if
 		// there was a need: this would require checking that the sequence number
 		// of the snapshot has been flushed, by comparing with
@@ -1007,14 +1026,20 @@ func (d *DB) newIter(ctx context.Context, batch *Batch, s *Snapshot, o *IterOpti
 	// Grab and reference the current readState. This prevents the underlying
 	// files in the associated version from being deleted if there is a current
 	// compaction. The readState is unref'd by Iterator.Close().
-	readState := d.loadReadState()
+	readState := sOpts.readState
+	if readState == nil && sOpts.vers == nil {
+		// NB: loadReadState() calls readState.ref().
+		readState = d.loadReadState()
+	} else if readState != nil {
+		readState.ref()
+	} else {
+		// s.vers != nil
+		sOpts.vers.Ref()
+	}
 
 	// Determine the seqnum to read at after grabbing the read state (current and
 	// memtables) above.
-	var seqNum uint64
-	if s != nil {
-		seqNum = s.seqNum
-	} else {
+	if seqNum == 0 {
 		seqNum = d.mu.versions.visibleSeqNum.Load()
 	}
 
@@ -1028,6 +1053,7 @@ func (d *DB) newIter(ctx context.Context, batch *Batch, s *Snapshot, o *IterOpti
 		merge:               d.merge,
 		comparer:            *d.opts.Comparer,
 		readState:           readState,
+		version:             sOpts.vers,
 		keyBuf:              buf.keyBuf,
 		prefixOrFullSeekKey: buf.prefixOrFullSeekKey,
 		boundsBuf:           buf.boundsBuf,
@@ -1057,7 +1083,10 @@ func (d *DB) newIter(ctx context.Context, batch *Batch, s *Snapshot, o *IterOpti
 func finishInitializingIter(ctx context.Context, buf *iterAlloc) *Iterator {
 	// Short-hand.
 	dbi := &buf.dbi
-	memtables := dbi.readState.memtables
+	var memtables flushableList
+	if dbi.readState != nil {
+		memtables = dbi.readState.memtables
+	}
 	if dbi.opts.OnlyReadGuaranteedDurable {
 		memtables = nil
 	} else {
@@ -1196,7 +1225,7 @@ func (d *DB) ScanInternal(
 			UpperBound: upper,
 		},
 	}
-	iter := d.newInternalIter(nil /* snapshot */, scanInternalOpts)
+	iter := d.newInternalIter(snapshotIterOpts{} /* snapshot */, scanInternalOpts)
 	defer iter.close()
 	return scanInternalImpl(ctx, lower, upper, iter, scanInternalOpts)
 }
@@ -1208,31 +1237,39 @@ func (d *DB) ScanInternal(
 // TODO(bilal): This method has a lot of similarities with db.newIter as well as
 // finishInitializingIter. Both pairs of methods should be refactored to reduce
 // this duplication.
-func (d *DB) newInternalIter(s *Snapshot, o *scanInternalOptions) *scanInternalIterator {
+func (d *DB) newInternalIter(sOpts snapshotIterOpts, o *scanInternalOptions) *scanInternalIterator {
 	if err := d.closed.Load(); err != nil {
 		panic(err)
 	}
 	// Grab and reference the current readState. This prevents the underlying
 	// files in the associated version from being deleted if there is a current
 	// compaction. The readState is unref'd by Iterator.Close().
-	readState := d.loadReadState()
+	readState := sOpts.readState
+	if readState == nil && sOpts.vers == nil {
+		readState = d.loadReadState()
+	} else if readState != nil {
+		readState.ref()
+	}
+	if sOpts.vers != nil {
+		sOpts.vers.Ref()
+	}
 
 	// Determine the seqnum to read at after grabbing the read state (current and
 	// memtables) above.
-	var seqNum uint64
-	if s == nil {
+	seqNum := sOpts.seqNum
+	if seqNum == 0 {
 		seqNum = d.mu.versions.visibleSeqNum.Load()
-	} else {
-		seqNum = s.seqNum
 	}
 
 	// Bundle various structures under a single umbrella in order to allocate
 	// them together.
 	buf := iterAllocPool.Get().(*iterAlloc)
 	dbi := &scanInternalIterator{
+		db:              d,
 		comparer:        d.opts.Comparer,
 		merge:           d.opts.Merger.Merge,
 		readState:       readState,
+		version:         sOpts.vers,
 		alloc:           buf,
 		newIters:        d.newIters,
 		newIterRangeKey: d.tableNewRangeKeyIter,
@@ -1251,7 +1288,10 @@ func (d *DB) newInternalIter(s *Snapshot, o *scanInternalOptions) *scanInternalI
 
 func finishInitializingInternalIter(buf *iterAlloc, i *scanInternalIterator) *scanInternalIterator {
 	// Short-hand.
-	memtables := i.readState.memtables
+	var memtables flushableList
+	if i.readState != nil {
+		memtables = i.readState.memtables
+	}
 	// We only need to read from memtables which contain sequence numbers older
 	// than seqNum. Trim off newer memtables.
 	for j := len(memtables) - 1; j >= 0; j-- {
@@ -1306,7 +1346,10 @@ func (i *Iterator) constructPointIter(
 	}
 	numMergingLevels += len(memtables)
 
-	current := i.readState.current
+	current := i.version
+	if current == nil {
+		current = i.readState.current
+	}
 	numMergingLevels += len(current.L0SublevelFiles)
 	numLevelIters += len(current.L0SublevelFiles)
 	for level := 1; level < len(current.Levels); level++ {
@@ -1435,7 +1478,7 @@ func (d *DB) NewIter(o *IterOptions) *Iterator {
 // NewIterWithContext is like NewIter, and additionally accepts a context for
 // tracing.
 func (d *DB) NewIterWithContext(ctx context.Context, o *IterOptions) *Iterator {
-	return d.newIter(ctx, nil /* batch */, nil /* snapshot */, o)
+	return d.newIter(ctx, nil /* batch */, snapshotIterOpts{}, o)
 }
 
 // NewSnapshot returns a point-in-time view of the current DB state. Iterators
@@ -1459,6 +1502,28 @@ func (d *DB) NewSnapshot() *Snapshot {
 	d.mu.snapshots.pushBack(s)
 	d.mu.Unlock()
 	return s
+}
+
+// NewEventuallyFileOnlySnapshot returns a point-in-time view of the current DB
+// state, similar to NewSnapshot. See the comment at EventuallyFileOnlySnapshot
+// for its semantics.
+func (d *DB) NewEventuallyFileOnlySnapshot(keyRanges []KeyRange) *EventuallyFileOnlySnapshot {
+	if err := d.closed.Load(); err != nil {
+		panic(err)
+	}
+
+	internalKeyRanges := make([]internalKeyRange, len(keyRanges))
+	for i := range keyRanges {
+		if i > 0 && d.cmp(keyRanges[i-1].End, keyRanges[i].Start) > 0 {
+			panic("pebble: key ranges for eventually-file-only-snapshot not in order")
+		}
+		internalKeyRanges[i] = internalKeyRange{
+			smallest: base.MakeInternalKey(keyRanges[i].Start, InternalKeySeqNumMax, InternalKeyKindMax),
+			largest:  base.MakeExclusiveSentinelKey(InternalKeyKindRangeDelete, keyRanges[i].End),
+		}
+	}
+
+	return d.makeEventuallyFileOnlySnapshot(keyRanges, internalKeyRanges)
 }
 
 // Close closes the DB.
@@ -1631,6 +1696,10 @@ func (d *DB) Compact(start, end []byte, parallelize bool) error {
 		}
 	}
 
+	keyRanges := make([]internalKeyRange, len(meta))
+	for i := range meta {
+		keyRanges[i] = internalKeyRange{smallest: m.Smallest, largest: m.Largest}
+	}
 	// Determine if any memtable overlaps with the compaction range. We wait for
 	// any such overlap to flush (initiating a flush if necessary).
 	mem, err := func() (*flushableEntry, error) {
@@ -1640,7 +1709,7 @@ func (d *DB) Compact(start, end []byte, parallelize bool) error {
 		// overlaps.
 		for i := len(d.mu.mem.queue) - 1; i >= 0; i-- {
 			mem := d.mu.mem.queue[i]
-			if ingestMemtableOverlaps(d.cmp, mem, meta) {
+			if ingestMemtableOverlaps(d.cmp, mem, keyRanges) {
 				var err error
 				if mem.flushable == d.mu.mem.mutable {
 					// We have to hold both commitPipeline.mu and DB.mu when calling
@@ -2740,7 +2809,7 @@ func (d *DB) ScanStatistics(
 		},
 		rateLimitFunc: rateLimitFunc,
 	}
-	iter := d.newInternalIter(nil /* snapshot */, scanInternalOpts)
+	iter := d.newInternalIter(snapshotIterOpts{}, scanInternalOpts)
 	defer iter.close()
 
 	err := scanInternalImpl(ctx, lower, upper, iter, scanInternalOpts)
--- a/db_test.go
+++ b/db_test.go
@@ -493,7 +493,7 @@ func TestMergeOrderSameAfterFlush(t *testing.T) {
 
 	key := []byte("a")
 	verify := func(expected string) {
-		iter := d.NewIter(nil)
+		iter, _ := d.NewIter(nil)
 		if !iter.SeekGE([]byte("a")) {
 			t.Fatal("expected one value, but got empty iterator")
 		}
@@ -696,7 +696,7 @@ func TestIterLeak(t *testing.T) {
 					if flush {
 						require.NoError(t, d.Flush())
 					}
-					iter := d.NewIter(nil)
+					iter, _ := d.NewIter(nil)
 					iter.First()
 					if !leak {
 						require.NoError(t, iter.Close())
@@ -746,7 +746,7 @@ func TestIterLeakSharedCache(t *testing.T) {
 
 					// Check if leak detection works with only one db closing.
 					{
-						iter1 := d1.NewIter(nil)
+						iter1, _ := d1.NewIter(nil)
 						iter1.First()
 						if !leak {
 							require.NoError(t, iter1.Close())
@@ -764,7 +764,7 @@ func TestIterLeakSharedCache(t *testing.T) {
 					}
 
 					{
-						iter2 := d2.NewIter(nil)
+						iter2, _ := d2.NewIter(nil)
 						iter2.First()
 						if !leak {
 							require.NoError(t, iter2.Close())
@@ -837,7 +837,7 @@ func TestMemTableReservation(t *testing.T) {
 	// Flush in the presence of an active iterator. The iterator will hold a
 	// reference to a readState which will in turn hold a reader reference to the
 	// memtable.
-	iter := d.NewIter(nil)
+	iter, _ := d.NewIter(nil)
 	require.NoError(t, d.Flush())
 	// The flush moved the recycled memtable into position as an active mutable
 	// memtable. There are now two allocated memtables: 1 mutable and 1 pinned
@@ -895,7 +895,7 @@ func TestCacheEvict(t *testing.T) {
 	}
 
 	require.NoError(t, d.Flush())
-	iter := d.NewIter(nil)
+	iter, _ := d.NewIter(nil)
 	for iter.First(); iter.Valid(); iter.Next() {
 	}
 	require.NoError(t, iter.Close())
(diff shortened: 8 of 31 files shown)


itsbilal:
This is ready for a look; the latest push keeps all Snapshot guarantees for the key spans passed in.

jbowens (commented):
> Reviewed 1 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2775)* status: 1 of 10 files reviewed, 9 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`compaction.go` line 2132 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamvKeOAzV8rLvrCiWj:-NamvKeOAzV8rLvrCiWk:b-805zrn) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/compaction.go#L2132)):*
> ```Go
> 			// The fact that this transition spins off its own goroutine is
> 			// unfortunate, but it's better than having the flush goroutine wait for
> 			// acquisition of s.efos' internal mutex while it holds the DB mutex.
> ```

do we do anything slow while holding the EFOS's mutex? I find the asynchrony harder to reason about. 
___
*[`ingest.go` line 71 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamOnJz0_6jbEBqaTvP:-NamOnJz0_6jbEBqaTvQ:b-7yi7ed) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/ingest.go#L71)):*
> ```Go
> 
> // OverlapsKeyRange checks if this span overlaps with the provided KeyRange.
> // Not that we aren't checking for full containment of either span in the other,
> ```

nit: s/Not/Note/
___
*[`ingest.go` line 1190 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamkwJn7RZfxKL44MPr:-NamkwJn7RZfxKL44MPs:b-xbsvo8) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/ingest.go#L1190)):*
> ```Go
> 			// EventuallyFileOnlySnapshot that has yet to be converted to a file-only
> 			// snapshot (as it's waiting on a memtable to flush). In that case, wait
> 			// on the mutable memtable just to be safe.
> ```

This could use an explanation for why it's necessary to wait on the mutable memtable. I'm also having trouble convincing myself this is sufficient.

My muddled understanding: If the `exciseSpan` overlaps with the mutable memtable, `mem` already points to the mutable memtable. So this code only has an effect when the `exciseSpan` does NOT overlap the mutable memtable. The problematic case is where the `exciseSpan` and the file-only snapshots protected ranges intersect but do not cover exactly the same keyspan(s). If the `exciseSpan` finds no overlap with the mutable memtable but the intersecting snapshot has not yet converted to a file-only snapshot, the snapshot must be waiting on some key to be flushed that the falls outside the `exciseSpan`'s bounds. If the excise proceeded, it would remove keys that the eventually-file-only snapshot requires before the eventually-file-only-snapshot has acquired its reference.

But does it need to be the _mutable_ memtable? We need the excise to wait for the most recent memtable that an intersecting eventually file-only snapshot is waiting for. This may not be the mutable memtable. It could be an immutable memtable currently being flushed. Suppose we wait for the most recent memtable that an intersecting eventually-only file snapshot is waiting for, and that's memtable _M_. Then another eventually-only file snapshot is opened that again overlaps the excise span. That EOFS _S_<sub>2</sub> intersects the mutable memtable _M+1_.  The excise has not yet applied, so the EOFS _S_<sub>2</sub> _also_ expects to observe the keys that the excise span plans on removing once _M+1_ is flushed.

If the user is continually opening EOFS for the same intersecting span and continually intersecting with the mutable memtable, won't it never be safe for the excise to proceed?
___
*[`snapshot.go` line 198 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamPbToCSUIjMQ0Wb9e:-NamPbToCSUIjMQ0Wb9f:b5zeowa) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L198)):*
> ```Go
> // An EventuallyFileOnlySnapshot also takes a list of KeyRanges at
> // instantiation time. These are saved inside the underlying Snapshot, and
> // any excises that happen during it
> ```

unfinished (stale?) comment
___
*[`snapshot.go` line 201 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamRPU1DRj3sgcPt6j5:-NamRPU1DRj3sgcPt6j6:bjrm2xa) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L201)):*
> ```Go
> type EventuallyFileOnlySnapshot struct {
> 	mu struct {
> 		sync.RWMutex
> ```

Is anything expensive done while holding the write mutex? Wondering if this should be an ordinary `sync.Mutex` to simplify the synchronization
___
*[`snapshot.go` line 214 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamQIPv0bRvx4hYVMPE:-NamQIPv0bRvx4hYVMPF:b-vig4fk) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L214)):*
> ```Go
> 		// The wrapped regular snapshot, if not a file-only snapshot yet.
> 		snap      *Snapshot
> 		readState *readState
> ```

nit: can you explicitly mention that the `EventuallyFileOnlySnapshot` has ref'd the `readState` if non-nil?
___
*[`snapshot.go` line 223 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamuoOLDYoSJeZSjU4t:-NamuoOLDYoSJeZSjU4u:b-utxsnb) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L223)):*
> ```Go
> 	// excised, if true, signals that the above ranges were excised during the
> 	// lifetime of this snapshot.
> 	excised atomic.Bool
> ```

How is this used other than the assertion?
___
*[`snapshot.go` line 238 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Namut38DHupp1g9GFsS:-Namut38DHupp1g9GFsT:b-1q3b6h) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L238)):*
> ```Go
> ) {
> 	if es.excised.Load() {
> 		panic("pebble: snapshot excised before conversion to file-only snapshot")
> ```

nit: `panic(errors.AssertionFailedf(...))`
___
*[`snapshot.go` line 351 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamSsyR8hyYeQQbVOL0:-NamSsyR8hyYeQQbVOL1:bl6ycq8) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L351)):*
> ```Go
> 	close(es.closed)
> 	es.mu.RLock()
> 	defer es.mu.RUnlock()
> ```

Should this be acquiring a write lock on the RWMutex?

sumeerbhola (changes requested):
still reading, flushing some initial comments

> Reviewed 2 of 10 files at r1, 2 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2775#-:-Namy2B128M4b4c3ivri:b2quafx)* status: 4 of 10 files reviewed, 17 unresolved discussions (waiting on @itsbilal)

___
*[`db.go` line 1018 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamsM3M8WAWUA7SQMHJ:-NamsM3M8WAWUA7SQMHK:bhpu7rn) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/db.go#L1018)):*
> ```Go
> 	// files in the associated version from being deleted if there is a current
> 	// compaction. The readState is unref'd by Iterator.Close().
> 	readState := sOpts.readState
> ```

Shouldn't the code be reffing `sOpts.readState`, to correspond to the unref that will happen when the iterator is closed?

Also be reffing `sOpts.vers`?
___
*[`db.go` line 1480 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Namsk6pEVQbU95yioZk:-Namsk6pEVQbU95yioZl:b-1gr1tt) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/db.go#L1480)):*
> ```Go
> // NewEventuallyFileOnlySnapshot returns a point-in-time view of the current DB
> // state, similar to NewSnapshot. See the comment at EventuallyFileOnlySnapshot
> // for its semantics.
> ```

if we require keyRanges to be sorted, we should say this here. And verify under `invariants.Enabled`.
___
*[`db.go` line 1488 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Namt3RHBLSMP7PhFjXr:-Namt3RHBLSMP7PhFjXs:bmlgur5) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/db.go#L1488)):*
> ```Go
> 	internalKeyRanges := make([]internalKeyRange, len(keyRanges))
> 	for i := range keyRanges {
> 		internalKeyRanges[i] = internalKeyRange{
> ```

Could you add a comment where `internalKeyRange` is declared that `largest` is inclusive.
___
*[`db.go` line 1500 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamuBEZ5rkCHbeq7cuz:-NamuBEZ5rkCHbeq7cv-:b-arg2c) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/db.go#L1500)):*
> ```Go
> 	for i := range d.mu.mem.queue {
> 		mem := d.mu.mem.queue[i]
> 		if ingestMemtableOverlaps(d.cmp, mem, internalKeyRanges) {
> ```

I wonder whether we could do this overlap checking without holding `d.mu`, since that will block all writes. I realize we do the same with ingests, and snapshot creation may not be a high rate activity, but just raising it as a potential improvement.
We have already picked `seqNum` so no future writes are relevant (except for excise). We could:
- Grab the list of memtables to check and readerRef them and release the mutex.
- Make a list of the ones that overlap (expensive step).
-  Acquire the mutex again and do pointer comparison of the overlapping ones and the current (unflushed) ones (set intersection). If the intersecting set of memtables is empty, then we can transition to file only now.

I guess the problem is that an excise could come in between the first mutex release and the subsequent acquisition and we've grabbed a seqnum already that is pre-excise. Perhaps we could push onto `DB.mu.snapshots` when holding the first mutex and use that to check that an excise did not happen. 
___
*[`db.go` line 1505 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Namx4HxCVtClzOn-pK7:-Namx4HxCVtClzOn-pK8:b-a3pj37) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/db.go#L1505)):*
> ```Go
> 		}
> 	}
> 	es := &EventuallyFileOnlySnapshot{
> ```

nit: can you move all this code to a `makeEventuallyFileOnlySnapshot` func in snapshot.go. It's easier to grok and maintain when its near to the struct declaration.
___
*[`iterator.go` line 2669 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamrFQ2EuCXHDhYcwT1:-NamrFQ2EuCXHDhYcwT2:b-r6tl9c) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/iterator.go#L2669)):*
> ```Go
> 	}
> 	if vers != nil {
> 		vers.Ref()
> ```

where do we unref the version in `Close`?

If this is a bug, I am surprised our "leaked iterators" checking on closing the DB did not catch it.
___
*[`snapshot.go` line 198 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Namomav8lzIb4XCY7NY:-Namomav8lzIb4XCY7NZ:b-m27vnh) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L198)):*
> ```Go
> // An EventuallyFileOnlySnapshot also takes a list of KeyRanges at
> // instantiation time. These are saved inside the underlying Snapshot, and
> // any excises that happen during it
> ```

nit: incomplete sentence.
___
*[`snapshot.go` line 219 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NampDUtEqy7e7Lmvzx6:-NampDUtEqy7e7Lmvzx7:bve185f) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L219)):*
> ```Go
> 	}
> 
> 	// Key ranges to watch for an excise on.
> ```

Are the ranges in sorted order?

sumeerbhola (changes requested):
> Reviewed 1 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2775)* status: 5 of 10 files reviewed, 20 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`ingest.go` line 1190 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamkwJn7RZfxKL44MPr:-Nan0KZs1IuhFp57DVPl:bt9i2ru) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/ingest.go#L1190)):*

This could use an explanation for why it's necessary to wait on the mutable memtable. I'm also having trouble convincing myself this is sufficient.

My muddled understanding: If the `exciseSpan` overlaps with the mutable memtable, `mem` already points to the mutable memtable. So this code only has an effect when the `exciseSpan` does NOT overlap the mutable memtable. The problematic case is where the `exciseSpan` and the file-only snapshots protected ranges intersect but do not cover exactly the same keyspan(s). If the `exciseSpan` finds no overlap with the mutable memtable but the intersecting snapshot has not yet converted to a file-only snapshot, the snapshot must be waiting on some key to be flushed that the falls outside the `exciseSpan`'s bounds. If the excise proceeded, it would remove keys that the eventually-file-only snapshot requires before the eventually-file-only-snapshot has acquired its reference.

But does it need to be the _mutable_ memtable? We need the excise to wait for the most recent memtable that an intersecting eventually file-only snapshot is waiting for. This may not be the mutable memtable. It could be an immutable memtable currently being flushed. Suppose we wait for the most recent memtable that an intersecting eventually-only file snapshot is waiting for, and that's memtable _M_. Then another eventually-only file snapshot is opened that again overlaps the excise span. That EOFS _S_<sub>2</sub> intersects the mutable memtable _M+1_.  The excise has not yet applied, so the EOFS _S_<sub>2</sub> _also_ expects to observe the keys that the excise span plans on removing once _M+1_ is flushed.

If the user is continually opening EOFS for the same intersecting span and continually intersecting with the mutable memtable, won't it never be safe for the excise to proceed?
</blockquote></details>

Also, isn't this the case where the ingest should not wait and the EFOS will eventually return an error due to the code later in this source file?
___
*[`snapshot.go` line 238 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Nan1jJt-Sib6nnSL-wP:-Nan1jJt-Sib6nnSL-wQ:b71qt0i) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L238)):*
> ```Go
> ) {
> 	if es.excised.Load() {
> 		panic("pebble: snapshot excised before conversion to file-only snapshot")
> ```

I wasn't expecting this to panic, since I thought we were admitting the possibility of this error.
___
*[`snapshot.go` line 261 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Nan2JjP0aseQR4s9SMx:-Nan2JjQ4SUd8hkm3-iQ:b-ei6cgo) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L261)):*
> ```Go
> 	earliestUnflushedSeqNum uint64, vers *version,
> ) error {
> 	if base.Visible(earliestUnflushedSeqNum, es.seqNum, InternalKeySeqNumMax) {
> ```

why should it fail here given we verified `!base.Visible(earliestUnflushedSeqNum, s.efos.seqNum, InternalKeySeqNumMax)` in compaction.go?
___
*[`snapshot.go` line 287 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Nan35PoArS6eBQ1s9m4:-Nan35Pp3hZae9fcFLwE:b-7th1a6) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L287)):*
> ```Go
> 		es.mu.transitioned.Broadcast()
> 	} else {
> 		vers.Unref()
> ```

Why would this else case happen? `NewEventuallyFileOnlySnapshot` doesn't use this method when creating a EFOS that is initially file only.
Is this because this is async so can be triggered by multiple flushes?
If we can't make the whole thing synchronous we should set some bit in `maybeTransitionToFileOnlySnapshotAsync` so that only the first caller needs to start a goroutine and then this method can be stricter about the state it expects to see.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2775#-:-NaqnHGa1Vw0O4Y-XTpq:b-814g8k)* status: 5 of 10 files reviewed, 15 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`compaction.go` line 2132 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamvKeOAzV8rLvrCiWj:-NaqnIz25BfrnWwuvxqr:b3zgfh0) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/compaction.go#L2132)):*

do we do anything slow while holding the EFOS's mutex? I find the asynchrony harder to reason about. 
</blockquote></details>

Nothing heavy. I just had to untangle a bunch of mutexes and add a `snap.closeLocked` to make things work well synchronously without deadlocks, but the latest change does it.
___
*[`db.go` line 1018 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamsM3M8WAWUA7SQMHJ:-NaqoNI-Dzgt3frHyq6i:b74bjjg) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/db.go#L1018)):*

Shouldn't the code be reffing `sOpts.readState`, to correspond to the unref that will happen when the iterator is closed?

Also be reffing `sOpts.vers`?
</blockquote></details>

You're right that we're missing one round of ref/unref of `vers`, but we don't need to ref readState in the case below
___
*[`db.go` line 1480 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Namsk6pEVQbU95yioZk:-NaqpGNPEu4Qtivr0YnG:bl0mxdv) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/db.go#L1480)):*

if we require keyRanges to be sorted, we should say this here. And verify under `invariants.Enabled`.
</blockquote></details>

We don't, as I expect the number of keyRanges to be a constant size and small for each EFOS (4?) so we just do a nested loop iteration each time. But if we want to require it to be sorted, it would allow us to do binary-searches to check for overlaps.
___
*[`db.go` line 1488 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Namt3RHBLSMP7PhFjXr:-NaqpbPr6aSpu0C_WbAB:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/db.go#L1488)):*

Could you add a comment where `internalKeyRange` is declared that `largest` is inclusive.
</blockquote></details>

Done.
___
*[`db.go` line 1505 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Namx4HxCVtClzOn-pK7:-NaquGi00Tv5blULMGId:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/db.go#L1505)):*

nit: can you move all this code to a `makeEventuallyFileOnlySnapshot` func in snapshot.go. It's easier to grok and maintain when its near to the struct declaration.
</blockquote></details>

Done.
___
*[`ingest.go` line 71 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamOnJz0_6jbEBqaTvP:-NaquJKl9R1PD5Ghk-2K:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/ingest.go#L71)):*

nit: s/Not/Note/
</blockquote></details>

Done.
___
*[`ingest.go` line 1190 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamkwJn7RZfxKL44MPr:-Narf6X52RaDvXdo0eaj:b-g6ktet) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/ingest.go#L1190)):*

Also, isn't this the case where the ingest should not wait and the EFOS will eventually return an error due to the code later in this source file?
</blockquote></details>

(conversation was continued on slack but bringing it back here)

Yes you're right @jbowens , the wait on the mutable memtable is not sufficient. I've removed this case entirely and brought back the error case in `WaitForFileOnlySnapshot` for when the excise races past the EFOS' transition. Also updated one of the tests to exercise the error case better.
___
*[`iterator.go` line 2669 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamrFQ2EuCXHDhYcwT1:-NaquMV64q9DOqwuqr1u:bh11ton) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/iterator.go#L2669)):*

where do we unref the version in `Close`?

If this is a bug, I am surprised our "leaked iterators" checking on closing the DB did not catch it.
</blockquote></details>

Done. We were missing an unref in Close. And since this is just the clone case, we weren't leaking an iterator.
___
*[`snapshot.go` line 198 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamPbToCSUIjMQ0Wb9e:-NaqvNy75J3O125Lw6zm:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L198)):*

unfinished (stale?) comment
</blockquote></details>

Done.
___
*[`snapshot.go` line 198 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Namomav8lzIb4XCY7NY:-NaqvOQC2mZEfvNcolUw:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L198)):*

nit: incomplete sentence.
</blockquote></details>

Done.
___
*[`snapshot.go` line 201 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamRPU1DRj3sgcPt6j5:-NaqvP_m9K-amEW0pAuE:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L201)):*

Is anything expensive done while holding the write mutex? Wondering if this should be an ordinary `sync.Mutex` to simplify the synchronization
</blockquote></details>

Done.
___
*[`snapshot.go` line 214 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamQIPv0bRvx4hYVMPE:-NaqvVCaBmka8XVuuyoE:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L214)):*

nit: can you explicitly mention that the `EventuallyFileOnlySnapshot` has ref'd the `readState` if non-nil?
</blockquote></details>

Done.
___
*[`snapshot.go` line 219 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NampDUtEqy7e7Lmvzx6:-NaqvW-WBeUN26gcVc4p:b-faxefj) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L219)):*

Are the ranges in sorted order?
</blockquote></details>

Nope (see comment above).
___
*[`snapshot.go` line 223 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamuoOLDYoSJeZSjU4t:-NaqvYNU7fbfF-sCpFEn:b-rg6yy4) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L223)):*

How is this used other than the assertion?
</blockquote></details>

Done (irrelevant given new changes)
___
*[`snapshot.go` line 238 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Nan1jJt-Sib6nnSL-wP:-Naqv_ux94mewhVyWfB8:b-21rb84) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L238)):*

I wasn't expecting this to panic, since I thought we were admitting the possibility of this error.
</blockquote></details>

Done (change to returning errors again)
___
*[`snapshot.go` line 238 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Namut38DHupp1g9GFsS:-NaqvctAB2O0zVyS-jgC:b-21rb84) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L238)):*

nit: `panic(errors.AssertionFailedf(...))`
</blockquote></details>

Done (change to returning errors again)
___
*[`snapshot.go` line 261 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Nan2JjP0aseQR4s9SMx:-Narb4rtDtb6YlISnbSt:b2ar3so) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L261)):*

why should it fail here given we verified `!base.Visible(earliestUnflushedSeqNum, s.efos.seqNum, InternalKeySeqNumMax)` in compaction.go?
</blockquote></details>

Done. (removed)
___
*[`snapshot.go` line 287 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Nan35PoArS6eBQ1s9m4:-Narb8Nw8D60WvmWNuSD:bljrc6k) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L287)):*

Why would this else case happen? `NewEventuallyFileOnlySnapshot` doesn't use this method when creating a EFOS that is initially file only.
Is this because this is async so can be triggered by multiple flushes?
If we can't make the whole thing synchronous we should set some bit in `maybeTransitionToFileOnlySnapshotAsync` so that only the first caller needs to start a goroutine and then this method can be stricter about the state it expects to see.

</blockquote></details>

In the change you reviewed, you were right this would never happen because only one flush would do the transition. In the oldest change and in the newest change as of this comment, it's helpful as `WaitForFileOnlySnapshot()` could also cause the transition to happen. However that's not necessary and we could leave that responsibility solely up to the flushing goroutine, in which case this else case would go away.
___
*[`snapshot.go` line 351 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamSsyR8hyYeQQbVOL0:-NarbmIe5dX5OVdKI46_:b8vcldg) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L351)):*

Should this be acquiring a write lock on the RWMutex?
</blockquote></details>

Yeah probably a good idea. Done.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2775#-:-NbPn20rBzVgKPHomDyP:b-814g8k)* status: 5 of 13 files reviewed, 19 unresolved discussions (waiting on @itsbilal, @jbowens, and @sumeerbhola)

___
*[`db.go` line 1480 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Namsk6pEVQbU95yioZk:-NbPn4ROA-n7lO4cxqEA:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/db.go#L1480)):*

Let's require them to be sorted (so we don't have to change the API) and it's fine to leave a TODO to improve the implementation later.
</blockquote></details>

Done.
___
*[`db.go` line 1500 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamuBEZ5rkCHbeq7cuz:-NbPn5o20ilDFOS8AUfo:bg4l448) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/db.go#L1500)):*

Thoughts about this? Should this be captured in a todo or an issue?
</blockquote></details>

I'll shortly file an issue enumerating this optimization, if that sounds good!
___
*[`db.go` line 986 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbL3-dN9rJNpOHeJ0ua:-NbPn2vG-J_b_ECekiEJ:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9e17bf2739f8c941b90758de8d26d8255a27e93c/db.go#L986)):*

I think it would be simpler to understand by enumerating the 3 cases. Something like:

```
// - Classic snapshot: only snapshot is set. Since this is a classic snapshot, the callee will use the latest readState.
// - EFOS that has not yet become file-only: only snapshot and readState are set. It is not safe for the callee to use the latest readState, since the transition to file-only can happen concurrently and cause the latest readState to stop preserving seqnums needed by the snapshot.
// - EFOS that has become file-only: only seqNum, vers are set.
//
// A consequence of the above cases is that at most one of vers and readState are non-nil, and both are nil for a classic snapshot.
```
</blockquote></details>

Done.
___
*[`db.go` line 1031 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbL1wOu8TyfAVpLu_My:-NbPn3_D0sYru83NX6gc:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9e17bf2739f8c941b90758de8d26d8255a27e93c/db.go#L1031)):*

nit: this nicely fits in as an `else` in the previous if-elseif block since when the first 2 conditions are not satisfied, what remains is `sOpts.vers != nil`.
</blockquote></details>

Done.
___
*[`ingest.go` line 1803 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbL6HDO2cx0vUhTxARZ:-NbPnFU6BcBq6rjwElhX:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9e17bf2739f8c941b90758de8d26d8255a27e93c/ingest.go#L1803)):*

nit: can break out of the loop after setting true
</blockquote></details>

Done.
___
*[`iterator.go` line 2669 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamrFQ2EuCXHDhYcwT1:-NbPnGO26le5kgXxKmJS:b-nyau35) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/iterator.go#L2669)):*

> And since this is just the clone case, we weren't leaking an iterator.

So do we not test clone in a way that would have triggered the bug? If so, can you create an issue?
</blockquote></details>

Added some tests for the clone case that'd have caught this. We do test Clone, we just weren't testing it with EFOS.
___
*[`snapshot.go` line 253 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbL9MCU0LiOUvGjXCqj:-NbPnR_K6OY4YNn0rFTH:b-mmy3t9) ([raw file](https://github.com/cockroachdb/pebble/blob/9e17bf2739f8c941b90758de8d26d8255a27e93c/snapshot.go#L253)):*

I'm unclear on why this is a channel. Can this be a bool inside mu? 
</blockquote></details>

We check it inside the loop in `WaitForFileOnlySnapshot` without holding the EFOS mutex, and I wanted to have the ability to break out of that loop without grabbing that mutex. That's really the only reason why it's not inside that mutex. I could make it an atomic though, if we want.
___
*[`snapshot.go` line 361 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbLA0tg6gXPKGG8vcKy:-NbPngIY2083uFeFc3o3:b-dhwhol) ([raw file](https://github.com/cockroachdb/pebble/blob/9e17bf2739f8c941b90758de8d26d8255a27e93c/snapshot.go#L361)):*

How easy would it be to make this a DB wide schedule (configured in opts) that would flush at a certain frequency if there was a waiting EFOS? Or can we convince ourselves that a EFOS created every 1ms will not cause excessive flushes?
</blockquote></details>

That's not that different from calling this method with a large value for `dur`, right? Say we do create an EFOS every 1ms, but we call `WaitForFileOnlySnapshot` with a 10s duration after creating each of those. The EFOS will all end up calling `maybeScheduleDelayedFlush` on the current memtable, and the first one to hit the 10s will end up going through with the memtable rotation/flush. The others will all realize the memtable they're waiting on has flushed and will do nothing, as `maybeScheduleDelayedFlush` is set up to handle that case.

We could reduce the goroutine proliferation by having it centrally scheduled in db, but functionally it won't be any different in causing or reducing excessive flushes. What do you think? It shouldn't be much harder to do it there so I'm down to make the change, but just wanted to confirm.
___
*[`snapshot.go` line 368 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbLARkSBlUMWux8T9XR:-NbPopeH9uud4WvQneXp:b-c382v8) ([raw file](https://github.com/cockroachdb/pebble/blob/9e17bf2739f8c941b90758de8d26d8255a27e93c/snapshot.go#L368)):*

Can we simplify by having a channel to signal that the transition to file only snapshot is done. That way we don't have to attempt the transition here.
</blockquote></details>

Done. I moved to a cond lock if that makes sense; reduces the chances that we wait on the channel _after_ a concurrent flush goroutine has done the transition.

jbowens (commented):
> Reviewed 1 of 10 files at r1, 1 of 7 files at r4, 1 of 5 files at r8, 1 of 4 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2775)* status: 6 of 13 files reviewed, 22 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`compaction.go` line 2135 at r9](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbQ8JF69uoc5PXb300Q:-NbQ8JF69uoc5PXb300R:b-v6snxh) ([raw file](https://github.com/cockroachdb/pebble/blob/2484eb019046744a3ae6768320a03271ac193286/compaction.go#L2135)):*
> ```Go
> 				continue
> 			}
> 			if s.efos.excised.Load() {
> ```

nit: can you add a comment explaining that the EFOS cannot transition to file-only in this case.
___
*[`db.go` line 1500 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamuBEZ5rkCHbeq7cuz:-NbQApFyAtotVNtEWfbn:b-iapj50) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/db.go#L1500)):*

I'll shortly file an issue enumerating this optimization, if that sounds good!
</blockquote></details>

When you file the issue, you can link it to #2112, the issue for the analogous improvement to Ingest.
___
*[`db.go` line 987 at r9](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbQ-Yrn8e9rJ7wKjxQ3:-NbQ-Yro-aNWKLVMiM5U:b-lmcprl) ([raw file](https://github.com/cockroachdb/pebble/blob/2484eb019046744a3ae6768320a03271ac193286/db.go#L987)):*
> ```Go
> //   - No snapshot: All fields are zero values.
> //   - Classic snapshot: Only `snapshot` is set. The latest readState will be used
> //     and the seqNum from inside the *Snapshot will be used.
> ```

nit: Would we be able to simplify further by removing `snapshot` and just copying `snapshot.seqNum` onto `seqNum`? Then we always use the provided `seqNum` if non-zero. If `readState` is provided, we fall into the EFOS pre-file-only state and use the readState. If `vers` is provided, we fall into the file-only state and use that version. If neither `readState` nor `vers` is provided, we use the latest read state and it's a classic snapshot.
___
*[`ingest.go` line 1791 at r9](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbQBXnyBlpJgr1t7mD-:-NbQBXnyBlpJgr1t7mD0:b-hzsd4l) ([raw file](https://github.com/cockroachdb/pebble/blob/2484eb019046744a3ae6768320a03271ac193286/ingest.go#L1791)):*
> ```Go
> 			// write a file overlapping with the excise span.
> 			if exciseSpan.OverlapsInternalKeyRange(d.cmp, c.smallest, c.largest) {
> 				c.cancel.Store(true)
> ```

The glorious code coverage integration that @RaduBerinde added shows that the cancel case is untested. I realize that predates this PR, but can you file an issue for a unit test that covers this case?
___
*[`iterator.go` line 2671 at r9](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbQG6c_DwhE2xCn02pk:-NbQG6c_DwhE2xCn02pl:b-casxc4) ([raw file](https://github.com/cockroachdb/pebble/blob/2484eb019046744a3ae6768320a03271ac193286/iterator.go#L2671)):*
> ```Go
> 	if readState != nil {
> 		readState.ref()
> 	}
> ```

When cloning an iterator constructed off a snapshot, should we try to create it off the snapshot instead of reusing the same readState? If the current read state has progressed, we might be able to release zombie [mem]tbls. I'm not sure if we ever Clone a snapshot iterator in Cockroach. Maybe leave a TODO?
___
*[`snapshot.go` line 287 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-Nan35PoArS6eBQ1s9m4:-NbQQ2q86Xu-PMmiFgTw:b-j8uo1i) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/snapshot.go#L287)):*
> we could leave that responsibility solely up to the flushing goroutine, in which case this else case would go away.

This seems worthwhile to reduce complexity.
___
*[`snapshot.go` line 217 at r9](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbQHg-9EDTbbMH8U-Wq:-NbQHg-9EDTbbMH8U-Wr:bisdzmw) ([raw file](https://github.com/cockroachdb/pebble/blob/2484eb019046744a3ae6768320a03271ac193286/snapshot.go#L217)):*
nit: I found this comment a little confusing and maybe a little too focused on the internal implementation. Wdyt of something like:

> EventuallyFileOnlySnapshot (aka EFOS) provides a read-only point-in-time view of the database state, similar to Snapshot. A EventuallyFileOnlySnapshot induces less write amplification, at the cost of increased space amplification. While a Snapshot may increase write amplification across all flushes and compactions, a EventuallyFileOnlySnapshot only prevents the elision of keys for flushes of memtables that a) are open when the snapshot is created and b) overlap the specified key ranges. Once all such memtables have been flushed, the "EventuallyFileOnlySnapshot" transitions to a file-only snapshot state in which it pins zombies sstables like an open Iterator would, without pinning any memtables. Callers that can tolerate the increased space amplification of pinning zombie sstables until the snapshot is closed may prefer EventuallyFileOnlySnapshots for their reduced write amplification.
>
> EventuallyFileOnlySnapshots interact with the [IngestAndExcise] operation in subtle ways. Unlike Snapshots, EventuallyFileOnlySnapshots guarantee that their read-only point-in-time view is unaltered by the excision. However, an excise  ...

(I didn't finish because I still don't fully grok the interaction with excise here (see my other comment).
___
*[`snapshot.go` line 346 at r9](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbQO5Gv0OnvVJMcsT-L:-NbQO5Gv0OnvVJMcsT-M:b-6zhdjh) ([raw file](https://github.com/cockroachdb/pebble/blob/2484eb019046744a3ae6768320a03271ac193286/snapshot.go#L346)):*
> ```Go
> //
> // Idempotent; can be called multiple times with no side effects.
> func (es *EventuallyFileOnlySnapshot) WaitForFileOnlySnapshot(dur time.Duration) error {
> ```

Checking understanding: 
- If you call `EventuallyFileOnlySnapshot`, you get a usable EFOS that you can call `NewIter` on, which will just use the pinned read state.
- If you call `WaitForFileOnlySnapshot` and receive a nil error, you're guaranteed that you can continue to use the EFOS as a file-only snapshot indefinitely without pinning memtables.
- If you created a EFOS, never call `WaitForFileOnlySnapshot` and keep calling `NewIter` on it, the behavior is identical to `Clone`-ing an iterator and the snapshot provides no utility. If a conflicting excise occurs, the `readState` stays ref'd.

I think this could use some more explicit documentation on the intended usage.
___
*[`snapshot.go` line 461 at r9](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbQRpZG6M9f8hMnvNO3:-NbQRpZG6M9f8hMnvNO4:b-5r971t) ([raw file](https://github.com/cockroachdb/pebble/blob/2484eb019046744a3ae6768320a03271ac193286/snapshot.go#L461)):*
> ```Go
> ) error {
> 	if es.db == nil {
> 		panic(ErrClosed)
> ```

is this case possible?

sumeerbhola (changes requested):
looks good -- only one small thing

> Reviewed 1 of 5 files at r8, 1 of 4 files at r9, 4 of 5 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2775#-:-NbVt0KkE5L17Czv-t4G:b-4tvu8b)* status: 8 of 13 files reviewed, 19 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`iterator.go` line 2671 at r9](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbQG6c_DwhE2xCn02pk:-NbVqT0K1VP6gBWKFfwo:b-4n9mmt) ([raw file](https://github.com/cockroachdb/pebble/blob/2484eb019046744a3ae6768320a03271ac193286/iterator.go#L2671)):*

Done (left TODO)
</blockquote></details>

We do not clone when we have a `pebbleSnapshot` in CockroachDB, so there is no urgency to do this.
___
*[`snapshot.go` line 361 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbLA0tg6gXPKGG8vcKy:-NbVs1az6XOMP_IO8j9T:b-bnlfex) ([raw file](https://github.com/cockroachdb/pebble/blob/9e17bf2739f8c941b90758de8d26d8255a27e93c/snapshot.go#L361)):*
> but functionally it won't be any different in causing or reducing excessive flushes

Agreed

I remembered now the other reason this was bothering me, and it relates to the use case of CRDB's outgoing snapshot. My assumption was that no one will call `WaitForFileOnlySnapshot` in that case, but we do want a transition to happen in a timely manner. We could change the CRDB code to spin up an extra goroutine, which is possibly fine given the limited concurrency. I don't have a strong opinion either way.
___
*[`snapshot.go` line 346 at r9](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbQO5Gv0OnvVJMcsT-L:-NbVoOdr3aWU2QlT0MgL:baov4xg) ([raw file](https://github.com/cockroachdb/pebble/blob/2484eb019046744a3ae6768320a03271ac193286/snapshot.go#L346)):*
> unless a conflicting excise happens in which case we keep using the readState.

I missed this. I think we should return an error. It's better than an OOM, which, however rare, is annoying.

itsbilal (commented):
TFTR!

Added a second commit with the change to return an error from NewIter, as it really increased the surface area of this PR.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2775#-:-Nb_StlO1DNrmayylAd6:bfy0uhz)* status: 4 of 30 files reviewed, 19 unresolved discussions (waiting on @jbowens, @RaduBerinde, and @sumeerbhola)

___
*[`snapshot.go` line 361 at r7](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbLA0tg6gXPKGG8vcKy:-Nb_OV396RhWWpbpCq9w:b-wnfiqc) ([raw file](https://github.com/cockroachdb/pebble/blob/9e17bf2739f8c941b90758de8d26d8255a27e93c/snapshot.go#L361)):*

> but functionally it won't be any different in causing or reducing excessive flushes

Agreed

I remembered now the other reason this was bothering me, and it relates to the use case of CRDB's outgoing snapshot. My assumption was that no one will call `WaitForFileOnlySnapshot` in that case, but we do want a transition to happen in a timely manner. We could change the CRDB code to spin up an extra goroutine, which is possibly fine given the limited concurrency. I don't have a strong opinion either way.
</blockquote></details>

We can probably leave that as a follow-up optimization, this change is large enough as-is. I'll file an issue to have one goroutine that ticks at a configurable interval and schedules a flush if there are waiting EFOS.
___
*[`snapshot.go` line 346 at r9](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbQO5Gv0OnvVJMcsT-L:-Nb_KDI7AFVI-ve_Wf10:bqiyto1) ([raw file](https://github.com/cockroachdb/pebble/blob/2484eb019046744a3ae6768320a03271ac193286/snapshot.go#L346)):*

> unless a conflicting excise happens in which case we keep using the readState.

I missed this. I think we should return an error. It's better than an OOM, which, however rare, is annoying.
</blockquote></details>

The one annoyance there is going to be threading the error value from `NewIter` through all the relevant Cockroach/Pebble interfaces. I've made that change, but the tests rely a little too heavily on the fact that non-EFOS implementations don't return an error in NewIter. If that changes the tests will fail or panic anyway so that's fine.

itsbilal (commented):
Added a third commit with more testing for concurrent excises and compaction / EFOS cancellation. Could have done that as a standalone PR, but since it came up here + it also tests for EFOS stuff, it makes sense to do it here.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2775#-:-NbaVIiX53tWv7FeEjHN:b-eec9ys)* status: 1 of 31 files reviewed, 19 unresolved discussions (waiting on @jbowens, @RaduBerinde, and @sumeerbhola)

___
*[`ingest.go` line 1791 at r9](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbQBXnyBlpJgr1t7mD-:-NbaVTNX93l48qxOzlvG:b-m539f) ([raw file](https://github.com/cockroachdb/pebble/blob/2484eb019046744a3ae6768320a03271ac193286/ingest.go#L1791)):*

Writing a unit test that blocks a compaction is relatively annoying but yes I can file an issue to do one. We'll get coverage with the metamorphic test of this case no matter what, in my experience it's very common for an excise to cancel a compaction in Cockroach too.
</blockquote></details>

This is done now.

sumeerbhola (approved):
:lgtm: 

> Reviewed 2 of 5 files at r8, 1 of 1 files at r11, 10 of 26 files at r13, 22 of 23 files at r14, 2 of 2 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2775#-:-NbpVcxD9UqXwnd2sk0Q:bg7igkv)* status: all files reviewed, 20 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`testdata/ingest_shared` line 739 at r13](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbpUFcR7G-Pq7EJcg4m:-NbpUFcR7G-Pq7EJcg4n:b956ca5) ([raw file](https://github.com/cockroachdb/pebble/blob/3e0491711c7fd4f82e42dc683c4ce599036d0bc7/testdata/ingest_shared#L739)):*
> ```
> .
> 
> # Eventually File-only snapshots error out when a concurrent excise happens.
> ```

where is the error below?

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2775#-:-Nbpn59j7G_a4HAASpMc:b-814g8k)* status: 6 of 31 files reviewed, 20 unresolved discussions (waiting on @jbowens, @RaduBerinde, and @sumeerbhola)

___
*[`db.go` line 1500 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NamuBEZ5rkCHbeq7cuz:-NbpxJYi7irTEuLntl7C:bqo5lzr) ([raw file](https://github.com/cockroachdb/pebble/blob/8223fdbc05599808f18df36acd333c8cc84a38b8/db.go#L1500)):*

On it!
</blockquote></details>

Filed https://github.com/cockroachdb/pebble/issues/2824. 
___
*[`testdata/ingest_shared` line 739 at r13](https://reviewable.io/reviews/cockroachdb/pebble/2775#-NbpUFcR7G-Pq7EJcg4m:-Nbpn6k5D7xNQoBdxDEj:baedn5o) ([raw file](https://github.com/cockroachdb/pebble/blob/3e0491711c7fd4f82e42dc683c4ce599036d0bc7/testdata/ingest_shared#L739)):*

where is the error below?
</blockquote></details>

AH oops. This comment is better described as "Tests for EFOS, which does ....., follow below". Reworded.

Issue #2756, opened 2023-07-19
db: use strict-obsolete sstables when writing to shared storage


sumeerbhola:
These were introduced recently, and permit using the vanilla sstable readers when reading foreign ssts, which is more efficient than wrappers. Documented in https://github.com/cockroachdb/pebble/blob/456a2a2ab692e57be2926280ed57c7779f00b6bb/sstable/format.go#L32-L181

@itsbilal

itsbilal:
Using strict-obsolete SSTs for the write path is a little more complex as Pebble is currently unaware of what part of the keyspace is shareable at write time. Maybe we can thread this through to Pebble as an Option. and we mark sstables that are completely inside the strict-obsolete keyspace as strict obsolete when writing to them. We currently also put timeseries keys in shared storage which will contain merges but that's okay because we will never share them across pebbles.

itsbilal:
Based on discussion in Storage weekly, the preference was to add compaction guards around the Table keyspace as the "shareable keyspace", which would be passed into Pebble as an Option. Sstable outputs will be split at these boundaries and any sstables created within this range will be strict-obsolete. This issue now tracks the implementation of that.

PR #2776, opened 2023-07-28
*: simplify pointCollapsingIter, use vanilla iters for foreign SSTs

itsbilal:
Previously we were using pointCollapsingIters to collapse obsolete points in foreign sstables. This was necessary because of snapshots that were open at the time those sstables were written, however they really blew up code complexity and performance. This change transitions to using vanilla sstable iterators for foreign sstables with `hideObsoletePoints` set to true. We error out if the sstable format for a foreign sstable is not high enough to support the obsolete flag.

Also remove all cases in the pointCollapsingIter that were not used by ScanInternal (which is the last remaining user of it).

Part of #2756.

### DIFF
--- a/scan_internal.go
+++ b/scan_internal.go
@@ -7,7 +7,6 @@ package pebble
 import (
 	"context"
 	"fmt"
-	"sync"
 
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
@@ -16,7 +15,6 @@ import (
 	"github.com/cockroachdb/pebble/internal/manifest"
 	"github.com/cockroachdb/pebble/objstorage"
 	"github.com/cockroachdb/pebble/rangekey"
-	"github.com/cockroachdb/pebble/sstable"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


itsbilal:
Enjoy the code deletions, y'all

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2776#-:-Nagr9l24zHqkec3bt_g:bnfp4nl)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @sumeerbhola)

itsbilal:
TFTR!

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2776)* status: 0 of 5 files reviewed, 1 unresolved discussion

___
*[`table_cache.go` line 523 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2776#-NarXWam6_moGZVlETbE:-NarXWam6_moGZVlETbF:b-5a5fv9) ([raw file](https://github.com/cockroachdb/pebble/blob/3531d8b1a318fd0204afd04759cf31942d2ec805/table_cache.go#L523)):*
> ```Go
> 	}
> 
> 	if provider.IsForeign(objMeta) {
> ```

(tardy drive-by comment)
this could use a code comment with justification and a possible optimization. Something like:

// A foreign sst is assigned a seqnum s when it is ingested into Pebble. Only non-obsolete point keys in that sst are potentially relevant to an iterator, and these non-obsolete keys have seqnum s, and are only relevant if opts.snapshotForHideObsoletePoints > s. So we can always set `hideObsoletePoints=true`. Additionally, we could optimize by returning `emptyIter` when opts.snapshotForHideObsoletePoints <=  s.

Issue #2711, opened 2023-07-05
metamorphic: Add two-instance variation to test disaggregated ingestion


itsbilal:
Similar to how the crossversion test adds a new dimension to the metamorphic test by changing versions of Pebble and running random operations on top of it, we can get greater test coverage of disaggregated ingestions / skip-shared replication by having a variant of the metamorphic test that runs two different Pebble instances in tandem and adds a new `replicate` operation that under some settings does a skip-shared `ScanInternal` from one node to another, and in others, does a plain old iteration + sstable write + ingestion.

Possible ways of implementing this two-instance metamorphic test:
1) Reuse all of the existing metamorphic test code, but add a new metamorphic test type that creates two Pebbles, and adds a `switch` operation that changes the active foreground pebble instance. All batches/snapshots/iterators/etc are created on the active pebble instance at the time of batch/snapshot/iterator opening. This option has the benefit of only having one history/operations file to look into in case of failures.

2) Generate two different sequences of operations for the two pebbles, but at runtime we ensure synchronization/determination by running one operation in each sequence of operations at a time: line 1 in pebble 1, then line 1 in pebble 2, then line 2 in pebble 1, then line 2 in pebble 2, etc. The `replicate` is the only operation that has one pebble talking to the other.

PR #2782, opened 2023-08-01
internal/rangekey: Fix sort of range keys in foreign sstable iterators

itsbilal:
Tests in #2698 combined with `TestIngestShared` caught that we were not returning range keys in the order expected by other internal iterators when reading from a foreign sstable. This change re-sorts range keys in trailer descending order once they've been coalesced in ForeignSSTTransformer.

### DIFF
--- a/internal/rangekey/coalesce.go
+++ b/internal/rangekey/coalesce.go
@@ -375,7 +375,8 @@ func coalesce(
 // it calls coalesce to remove range keys shadowed by other range keys, but also
 // retains the range key that does the shadowing. In addition, it elides
 // RangeKey unsets/dels in L6 as they are inapplicable when reading from a
-// different Pebble instance.
+// different Pebble instance. Finally, it returns keys sorted in trailer order,
+// not suffix order, as that's what the rest of the iterator stack expects.
 type ForeignSSTTransformer struct {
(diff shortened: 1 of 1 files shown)


jbowens (commented):
Do we have metamorphic test coverage that should have caught this?

Is this `ForeignSSTTransformer` still necessary? I thought that we could use ordinary ingest global seqnum replacement for range keys now?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2782#-:-Nan4cMd-m0bJI69r-TS:b-w8mec1)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @RahulAggarwal1016)

RahulAggarwal1016 (approved):

itsbilal:
@jbowens Yes, metamorphic testing of foreign sstable iterators would have caught this, so #2711 is probably the one to wait for.

And you're right, we don't need the coalescing logic anymore. There was one concern around excess rangekeydels/unsets being emitted due to snapshots (because the obsolete key kind logic doesn't really apply here), but I don't think it should functionally change anything given all keys in the sstable will have the same sequence number. I've changed this PR to just remove `ForeignSSTTransformer`.

itsbilal:
TFTR!

PR #2790, opened 2023-08-04
Makefile: add testcoverage target

RaduBerinde:
Convenience target to generate a unit test coverage profile.


nicktrav (approved):
Should we gitignore `coverprofile.out`? Otherwise LGTM. Thanks for adding this!

RaduBerinde:
TFTR! Good point, added to `.gitignore`.

PR #2817, opened 2023-08-11
crossversion: add more comments

RaduBerinde:


RaduBerinde:
*[`internal/metamorphic/crossversion/crossversion_test.go` line 53 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2817#-NbZvPyo0hr5SZt4VJDs:-NbZvPyo0hr5SZt4VJDt:bx2vy24) ([raw file](https://github.com/cockroachdb/pebble/blob/9d98d2fd086039ea191ae363fcb888329e292080/internal/metamorphic/crossversion/crossversion_test.go#L53)):*
> ```Go
> The expected format is <label>,<SHA>,<test-binary-path>.
> The label should be a human-readable label describing the
> version, for example, 'CRDB-22.1'. The SHA indicates the
> ```

I made this correction because it looks like we are splitting on spaces as well:
```
	fields := strings.FieldsFunc(value, func(r rune) bool { return r == ',' || unicode.IsSpace(r) })
```

RaduBerinde:
TFTR!

PR #2821, opened 2023-08-14
scripts: add code coverage script

RaduBerinde:
#### metamorphic: allow coverage instrumentation

`TestMeta` runs multiple instances of itself in other processes by
executing the same binary (`os.Args[0]`). This doesn't work with the
go test code coverage infrastructure - the "inner" call is not
covered.

To address this, we modify `TestMeta` to allow specifying the "inner
binary" and we create a regular binary `metarunner` that can run a
single instance of `TestMeta` (i.e. either `--run-dir` or `--compare`
is passed). This will allow building `metarunner` with binary coverage
instrumentation (see https://go.dev/testing/coverage/#building).

#### scripts: add code coverage script

Add a script that collects overall code coverage profiles from all
tests (including the metamorphic test).

It generates three profiles, in LCOV format: `profile-tests.lcov`,
`profile-meta.lcov`, and `profile-combined.lcov`.

The intention is to make this script run nightly and add a front end
to visualize the coverage.

### DIFF
--- a/internal/metamorphic/meta_test.go
+++ b/internal/metamorphic/meta_test.go
@@ -5,14 +5,11 @@
 package metamorphic
 
 import (
-	"flag"
-	"math"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"testing"
 
-	"github.com/cockroachdb/pebble/internal/randvar"
+	"github.com/cockroachdb/pebble/internal/metamorphic/metaflags"
 	"github.com/cockroachdb/pebble/metamorphic"
 )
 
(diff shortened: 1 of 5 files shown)


RaduBerinde:
@jbowens friendly ping for the metamorphic test commit.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2821)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @jbowens and @srosenberg)

___
*[`internal/metamorphic/metarunner/main.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2821#-Nc3F9288VjwCqFjTiaA:-Nc3YGZv4-jDGanYb_JY:b-y2o9w) ([raw file](https://github.com/cockroachdb/pebble/blob/8a8822d1e58fe2117480b8898f60cca970e7048e/internal/metamorphic/metarunner/main.go#L49)):*

It's a little difficult because some flags apply to both, some apply only to RunOnce and some apply only to Run (and we don't want the latter to be defined at all in this binary). I'll give it a shot though.
</blockquote></details>

Ok, I think it came out fairly clean. PTAL

jbowens (approved):
:lgtm_strong:

> Reviewed 2 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2821#-:-Nc3i_9H67kzX6fRNDjD:b-jbsf8z)* status: 3 of 5 files reviewed, all discussions resolved (waiting on @srosenberg)

RaduBerinde:
@rickystewart if you have some time, could you review the script (second commit)? Sorry, I don't know who else to ask.

srosenberg (approved):
Nice!

PR #2731, opened 2023-07-12
db: experimentally bump L0 priority in the presence of range deletes

bananabrick:

PR #2823, opened 2023-08-14
db: ignore compensation in lower levels during compaction picking

bananabrick:
If a lower levels score is higher than a higher levels score only due
to compensation, then we'll pick the higher level for compaction.

Compacting into a level which has a high score only due to compensation
shouldn't incur a higher write amp. We prioritize higher levels in this
way because we don't want L0 compaction bandwidth to vary due to
compensation in lower levels. Variation in L0 compaction bandwidth can
lead to AC queueing.

Consider two arbitrary candidateLevelInfos x, y where x.level < y.level.

Case 1: x.score > y.score && x.score > y.rawSmoothed.
The behaviour hasn't changed with this pr. x will be placed before y.

Case 2: x.score < y.score && x.score < y.rawSmoothed.
The behaviour hasn't changed with this pr. x will place after y.

Case 3: x.score < y.score && x.score > y.rawSmoothed.
In this case, we will place x before y, which is a behaviour change.
	
Case 3.1: x.rawSmoothed > y.rawSmoothed.
y.score is only higher because of compensation. As described above,
it is okay to pick the higher level in this case.

Case 3.2: x.rawSmoothed < y.rawSmoothed.
x.score is greater than y.rawSmoothed because of compensation in
x.level. But y.score > x.score because of compensation in
y.level. Since y.score is only greater than x.score due to
compensation it is okay to compact x.level first.

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -1334,7 +1334,7 @@ func (p *compactionPickerByScore) pickAuto(env compactionEnv) (pc *pickedCompact
 				p.addScoresToPickedCompactionMetrics(pc, scores)
 				pc.score = info.score
 				// TODO(bananabrick): Create an EventListener for logCompaction.
-				if false {
+				if true {
 					logCompaction(pc)
 				}
 				return pc
@@ -1355,7 +1355,7 @@ func (p *compactionPickerByScore) pickAuto(env compactionEnv) (pc *pickedCompact
 			p.addScoresToPickedCompactionMetrics(pc, scores)
 			pc.score = info.score
 			// TODO(bananabrick): Create an EventListener for logCompaction.
-			if false {
+			if true {
 				logCompaction(pc)
 			}
 			return pc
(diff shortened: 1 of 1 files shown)


bananabrick:
This is a stronger change compared to https://github.com/cockroachdb/pebble/pull/2731. 2371 only changed where L0 was placed, but this pr can also have an effect on other levels. There's not much benefit to changing how the other levels are placed, but having a heuristic apply uniformly to every level might be desirable to avoid cases where L0 -> Lbase compaction are prioritized too much, but Lbase won't compact because a level lower than Lbase has a high score due to compensation.

bananabrick:
I can add some tests if everything looks ok here

bananabrick:
Review can wait until the deep dive tomorrow

bananabrick:
We can see some improvement on clearrange/checks=true, compared to master.

master
https://grafana.testeng.crdb.io/d/ngnKs_j7x/admission-control?orgId=1&var-cluster=teamcity-11300088-1691905772-63-n10cpu16&var-instances=All&from=1691934438241&to=1691942063866

vs 

https://grafana.testeng.crdb.io/d/ngnKs_j7x/admission-control?orgId=1&var-cluster=arjunnair-1692046636-01-n10cpu16&var-instances=All&from=1692046858541&to=1692052820141

bananabrick:
commit #3 didn't really work: https://grafana.testeng.crdb.io/d/ngnKs_j7x/admission-control?orgId=1&var-cluster=arjunnair-1692057788-01-n10cpu16&var-instances=All&from=1692060200298&to=1692063014727

PR #2828, opened 2023-08-15
Implement APIs for creating batch with pre-allocated size

rjl493456442:
This PR introduces two APIs for creating batch object with pre-allocated memory space.

It's very useful in case of large batch construction if the rough batch size is aware in advance,
which can prevent several rounds of unnecessary memory allocations and mitigate memory
peak because of instant memory allocation in a short time.

This PR also fixes a tiny issue in batch.Init which always allocate the memory space. Pebble
has batchPool and indexedBatchPool internally in order to reuse the allocated memory. This
unexpected behavior is break the original intention.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -1294,12 +1294,16 @@ func (b *Batch) Indexed() bool {
 	return b.index != nil
 }
 
-func (b *Batch) init(cap int) {
+// init ensures that the batch data slice is initialized to meet the
+// minimum required size and allocates space for the batch header.
+func (b *Batch) init(bytes int) {
 	n := batchInitialSize
-	for n < cap {
+	for n < bytes {
 		n *= 2
 	}
-	b.data = rawalloc.New(batchHeaderLen, n)
+	if cap(b.data) < n {
+		b.data = rawalloc.New(batchHeaderLen, n)
+	}
 	b.setCount(0)
 	b.setSeqNum(0)
 	b.data = b.data[:batchHeaderLen]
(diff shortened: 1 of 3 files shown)


rjl493456442:
@jbowens can we merge it?

rjl493456442 (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2828)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`batch.go` line 1306 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2828#-NcNtbAN6COgQhrn3J3M:-Ncer4nc6wjP36vKyzOa:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/f2ba568968a2bf628b9248341c6c9655b524bdac/batch.go#L1306)):*

Nice find!
</blockquote></details>

Done.
___
*[`batch_test.go` line 292 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2828#-Ncc94vO9y4CodgPZshW:-Ncer6FD8UH_KSNi-GMB:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b155c138d206c26f27063910ae79441827e0337d/batch_test.go#L292)):*

The linter fails in CI on this line because this assigned value of `b` is never used. Can you please revert this to `b = Batch{}`?
</blockquote></details>

Done.

jbowens:
Thanks @rjl493456442!

PR #2829, opened 2023-08-15
sstable: fix two-level monotonic bounds bug

jbowens:
This commit fixes a sstable iterator bug surfaced by the metamorphic test in
the issue https://github.com/cockroachdb/pebble/issues/2816.

In the circumstances of this bug, a two-level sstable iterator with
monotonically advanced bounds could reuse the wrong index block if a previous
block was excluded due to a block-property filter. An accompanying regression
unit test is added too.

Additionally, it refactors some additional sstable tests to run against more
table formats and automatically against newly introduced table formats.

While this commit is sufficient to fix the bug, in subsequent work I anticipate
refactoring and tightening invariants. This refactoring will not be suitable
for a backport, so this commit first resolves the bug in a backportable way.


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2829)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`sstable/testdata/reader_bpf/Pebblev3/iter` line 114 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2829#-NbuAS3RB6p_Y-RTt-Xi:-NbuAS3RB6p_Y-RTt-Xj:bjr6vqs) ([raw file](https://github.com/cockroachdb/pebble/blob/3bea918af26da384d5b93ef9b469908598e33226/sstable/testdata/reader_bpf/Pebblev3/iter#L114)):*
> ```
> |  topLevelIndex.InPlaceValue() = (Offset: 161, Length: 27, Props: 00020201)
> |  topLevelIndex.isDataInvalidated()=false
> |  index.Key() = "x#72057594037927935,17"
> ```

Here the `SeekGE("wb@2")` call updates the position of `topLevelIndex`, but it does not invalidate the `i.index`, so `i.index` holds a later block, not the one indicated by `i.topLevelIndex.Value()`

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2829#-:-Nbw5IQd8HMEKudUZCse:bg7igkv)* status: all files reviewed, 6 unresolved discussions (waiting on @jbowens)

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2829#-Nbvyul-4mziKLi-tFN7):*
which two unit tests?
___
*[`sstable/testdata/reader_bpf/Pebblev3/iter` line 40 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2829#-Nbw-vLu7C7QtA5-2xGO:-Nbw-vLu7C7QtA5-2xGP:b8n3dg8) ([raw file](https://github.com/cockroachdb/pebble/blob/3bea918af26da384d5b93ef9b469908598e33226/sstable/testdata/reader_bpf/Pebblev3/iter#L40)):*
> ```
> # Use the same block property filter, but use seeks to find these entries.
> # With the bug the second seek-ge below would step to the second lower-level
> # index block and only see the entry in the data block 4.
> ```

I don't quite understand. Is the output demonstrating the bug? I assume not since with the [a,c) bound, nothing should be visible as the output indicates. And with the [c,g) bound the output looks correct.
Or is this test failing because the output here is the correct output, and seek-ge c does not show d@7?
___
*[`sstable/testdata/reader_bpf/Pebblev3/iter` line 98 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2829#-Nbw1npNERN4yid7V6El:-Nbw1npNERN4yid7V6Em:b-q68q4w) ([raw file](https://github.com/cockroachdb/pebble/blob/3bea918af26da384d5b93ef9b469908598e33226/sstable/testdata/reader_bpf/Pebblev3/iter#L98)):*
> ```
> .
> | *sstable.twoLevelIterator:
> |  topLevelIndex.Key() = "x#72057594037927935,17"
> ```

Have we also broken the expectation of the monotonic bounds optimization that the SeekGE key will not be > the upper bound?
That is, even after fixing the bug, will the monotonic bounds optimization be incorrect?
I suspect not, because even though we are breaking the rule specified in https://github.com/cockroachdb/pebble/blob/8801607debb0af697ba74e902a4e68d648cbdc39/internal/base/iterator.go#L62-L64
(but without documenting this breaking, which we should fix), we are doing so only because a rangedel at a higher level is deleting some span of keys. So those keys between `v` and `wz@8` won't be visible anyway.
___
*[`sstable/data_test.go` line 399 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2829#-Nbw5FD_5zNR16XVB4Nv:-Nbw5FD_5zNR16XVB4Nw:bb7lbs2) ([raw file](https://github.com/cockroachdb/pebble/blob/3bea918af26da384d5b93ef9b469908598e33226/sstable/data_test.go#L399)):*
> ```Go
> 			continue
> 		case "internal-iter-state":
> 			fmt.Fprintf(&b, "| %T:\n", origIter)
> ```

nice
___
*[`sstable/reader_test.go` line 563 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2829#-Nbw-2QB0bpoh6hclR2g:-Nbw-2QB0bpoh6hclR2h:b3ym61o) ([raw file](https://github.com/cockroachdb/pebble/blob/3bea918af26da384d5b93ef9b469908598e33226/sstable/reader_test.go#L563)):*
> ```Go
> 		[NumTableFormats]string{
> 			TableFormatUnspecified: "",
> 			TableFormatLevelDB:     "testdata/readerstats_LevelDB",
> ```

great to have this improved coverage!

jbowens (commented):
@sumeerbhola I folded a minimal, backportable bug fix for the metamorphic test into this PR. Want to take a look at that?

After that's merged, I want to spend some time trying to refactor/clarify the various invariants. I suspect whatever results from that will be too invasive to backport.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2829#-:-NbzG62MCpKUxqJhL5Dk:bsdevlc)* status: 6 of 11 files reviewed, 5 unresolved discussions (waiting on @sumeerbhola)

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2829#-Nbvyul-4mziKLi-tFN7):*

which two unit tests?
</blockquote></details>

Oops, fixed.
___
*[`sstable/testdata/reader_bpf/Pebblev3/iter` line 40 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2829#-Nbw-vLu7C7QtA5-2xGO:-Nbz3XWD6EdOpm7mRQDP:b7vrxea) ([raw file](https://github.com/cockroachdb/pebble/blob/3bea918af26da384d5b93ef9b469908598e33226/sstable/testdata/reader_bpf/Pebblev3/iter#L40)):*

I don't quite understand. Is the output demonstrating the bug? I assume not since with the [a,c) bound, nothing should be visible as the output indicates. And with the [c,g) bound the output looks correct.
Or is this test failing because the output here is the correct output, and seek-ge c does not show d@7?
</blockquote></details>

Ah, sorry this test was an existing test that got renamed. This test is verbatim what was already there, and the new test begins on line 58.
___
*[`sstable/testdata/reader_bpf/Pebblev3/iter` line 98 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2829#-Nbw1npNERN4yid7V6El:-NbzG2tL0vi0TQTD6k1z:bf34fa8) ([raw file](https://github.com/cockroachdb/pebble/blob/3bea918af26da384d5b93ef9b469908598e33226/sstable/testdata/reader_bpf/Pebblev3/iter#L98)):*

Have we also broken the expectation of the monotonic bounds optimization that the SeekGE key will not be > the upper bound?
That is, even after fixing the bug, will the monotonic bounds optimization be incorrect?
I suspect not, because even though we are breaking the rule specified in https://github.com/cockroachdb/pebble/blob/8801607debb0af697ba74e902a4e68d648cbdc39/internal/base/iterator.go#L62-L64
(but without documenting this breaking, which we should fix), we are doing so only because a rangedel at a higher level is deleting some span of keys. So those keys between `v` and `wz@8` won't be visible anyway.
</blockquote></details>

Yeah, I have the same understanding.
___
*[`sstable/data_test.go` line 399 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2829#-Nbw5FD_5zNR16XVB4Nv:-NbzFw5E-aVx_zfGwB1R:b-j1we20) ([raw file](https://github.com/cockroachdb/pebble/blob/3bea918af26da384d5b93ef9b469908598e33226/sstable/data_test.go#L399)):*

nice
</blockquote></details>

:raised_hands: 
___
*[`sstable/reader_test.go` line 563 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2829#-Nbw-2QB0bpoh6hclR2g:-NbzFtRt6160_FgteI8L:b-j1we20) ([raw file](https://github.com/cockroachdb/pebble/blob/3bea918af26da384d5b93ef9b469908598e33226/sstable/reader_test.go#L563)):*

great to have this improved coverage!
</blockquote></details>

:raised_hands:

sumeerbhola (approved):
:lgtm: 

> Reviewed 3 of 3 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2829#-:-NbzySAl8-nuFvtClXW1:bg7igkv)* status: all files reviewed, 7 unresolved discussions (waiting on @jbowens)

___
*[`-- commits` line 13 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2829#-NbzujSx4HpKlrt_WLWP):*
nit: can you reference the issue
___
*[`sstable/reader_iter_two_lvl.go` line 264 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2829#-NbzxTP2-ysHwGb-uu23:-NbzxTP2-ysHwGb-uu24:bfw0363) ([raw file](https://github.com/cockroachdb/pebble/blob/2f25ae003a8e1c9d613afe301f9f854f65c18146/sstable/reader_iter_two_lvl.go#L264)):*
> ```Go
> 
> 	var dontSeekWithinSingleLevelIter bool
> 	if i.topLevelIndex.isDataInvalidated() || !i.topLevelIndex.valid() || i.index.isDataInvalidated() || err != nil ||
> ```

We are slightly narrowing the seek optimization because anytime we exhaust the bounds due to BPF and then see monotonic bounds, we will fall into here and seek the `topLevelIndex` even though we could have continued to use its position (and restricted ourselves to loading the index block). Alternatively, one can argue that the loading of the index block is costly, so we may as well seek to the right position in the topLevelIndex and load the appropriate index block instead of wastefully loading one that may be earlier than what we need.

The choice here is fine, but could you add some code comments discussing the tradeoffs.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2829#-:-Nc-dDhDB2KijxyAP86_:b-814g8k)* status: 10 of 11 files reviewed, 6 unresolved discussions (waiting on @sumeerbhola)

___
*[`sstable/reader_iter_two_lvl.go` line 264 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2829#-NbzxTP2-ysHwGb-uu23:-Nc-dFgXAQALY52fEW_t:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/2f25ae003a8e1c9d613afe301f9f854f65c18146/sstable/reader_iter_two_lvl.go#L264)):*

We are slightly narrowing the seek optimization because anytime we exhaust the bounds due to BPF and then see monotonic bounds, we will fall into here and seek the `topLevelIndex` even though we could have continued to use its position (and restricted ourselves to loading the index block). Alternatively, one can argue that the loading of the index block is costly, so we may as well seek to the right position in the topLevelIndex and load the appropriate index block instead of wastefully loading one that may be earlier than what we need.

The choice here is fine, but could you add some code comments discussing the tradeoffs.
</blockquote></details>

Done.

PR #2831, opened 2023-08-15
arenaskl: disallow arenas larger than 4GB

RaduBerinde:
This commit disallows creation of arenas larger than 4GB in invariants mode and otherwise limits the size to 4GB. The existing code could allow a larger buffer to be used, which would result in uint32 offset overflows.

We also cleanup the alignment argument so it's actually the alignment and not the maximum padding.

### DIFF
--- a/internal/arenaskl/arena.go
+++ b/internal/arenaskl/arena.go
@@ -23,6 +23,7 @@ import (
 	"unsafe"
 
 	"github.com/cockroachdb/errors"
+	"github.com/cockroachdb/pebble/internal/invariants"
 )
 
 // Arena is lock-free.
@@ -31,9 +32,7 @@ type Arena struct {
 	buf []byte
 }
 
-const (
-	align4 = 3
-)
+const nodeAlignment = 4
 
 var (
 	// ErrArenaFull indicates that the arena is full and cannot perform any more
@@ -44,11 +43,17 @@ var (
 // NewArena allocates a new arena using the specified buffer as the backing
 // store.
 func NewArena(buf []byte) *Arena {
-	// Don't store data at position 0 in order to reserve offset=0 as a kind
-	// of nil pointer.
+	if len(buf) > math.MaxUint32 {
+		if invariants.Enabled {
+			panic(errors.AssertionFailedf("attempting to create arena of size %d", len(buf)))
+		}
+		buf = buf[:math.MaxUint32]
+	}
 	a := &Arena{
 		buf: buf,
 	}
+	// We don't store data at position 0 in order to reserve offset=0 as a kind of
+	// nil pointer.
 	a.n.Store(1)
 	return a
 }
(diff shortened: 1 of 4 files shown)


RaduBerinde:
Hm, `maxMemTableSize` is 4GB but here I'm limiting to `4GB - 1 byte`. I think the latter limit makes sense because there is code that constructs the offset after the end of a buffer into a 32-bit value: 
https://github.com/cockroachdb/pebble/blob/7fae04db11b04a97af5a49522c77716786126393/internal/arenaskl/arena.go#L95
https://github.com/cockroachdb/pebble/blob/7fae04db11b04a97af5a49522c77716786126393/internal/arenaskl/skl.go#L413
https://github.com/cockroachdb/pebble/blob/7fae04db11b04a97af5a49522c77716786126393/internal/arenaskl/skl.go#L440

internal/arenaskl/arena.go
```diff
@@ -76,15 +88,15 @@ func (a *Arena) alloc(size, align, overflow uint32) (uint32, uint32, error) {
...
+	padded := size + align - 1
 
 	newSize := a.n.Add(uint64(padded))
 	if int(newSize)+int(overflow) > len(a.buf) {
 		return 0, 0, ErrArenaFull
 	}
 
 	// Return the aligned offset.
-	offset := (uint32(newSize) - padded + align) & ^align
+	offset := (uint32(newSize) - size) & ^(align - 1)
```
itsbilal, line 99: super nit: maybe throw in an invariants.Enabled guarded panic for align not being a power of 2?

itsbilal (approved):
LGTM

RaduBerinde (commented):
TFTR. I decreased `maxBatchSize` and `maxMemTableSize` to 4GB - 1 as well. Let me know if that makes sense.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2831#-:-Nc380q13uxlk6KXuil2:b-elolhd)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`internal/arenaskl/arena.go` line 99 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2831#gh-1296614371:-Nc380KnBHb4FYHcMuz0:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/048b18daee4b43fd6599009c9a4b5e7f0c86402e/internal/arenaskl/arena.go#L99)):*

super nit: maybe throw in an invariants.Enabled guarded panic for align not being a power of 2?
</blockquote></details>

Done.

jbowens (approved):
:lgtm: 

> Let me know if that makes sense.

Makes sense to me.

> Reviewed 2 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2831#-:-Nc3HD6O3-oJBbSaIq_M:b-x7xk3s)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal)

PR #2834, opened 2023-08-17
vfs: always swap and verify in health checking updates

jbowens:
The diskHealthCheckingFile is designed with the expectation that file write operations are serial. Previously, under the invariants build tag we would verify this expectation through updating state with a CompareAndSwap, asserting that the previous value equaled the value previously written by the same goroutine.

This commit moves this logic out of the invariants.Enabled gating. While an atomic swap operation may be slower than a simple atomic store, any additional cost should be negligible relative to the latency of the I/O write operation itself. This can improve confidence that we're obeying this contract in all instances, including direct VFS-usage within CockroachDB.

Informs cockroachdb/cockroach#97968.

### DIFF
--- a/vfs/disk_health.go
+++ b/vfs/disk_health.go
@@ -13,7 +13,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/redact"
 )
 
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2834#-:-Nc3KFHW6dyv1UQhpFOF:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sumeerbhola)

jbowens:
TFTR!

PR #2837, opened 2023-08-17
db: add additional logic to the compaction picker

bananabrick:


bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2837)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`compaction_picker.go` line 1104 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2837#-Nc4UOPa63VyE3xn6hwp:-Nc4UOPa63VyE3xn6hwq:b-anwzg0) ([raw file](https://github.com/cockroachdb/pebble/blob/eba5a250ef2c892a22d0af7a4807af768fbb2cf2/compaction_picker.go#L1104)):*
> ```Go
> 		info.score = fileScore
> 	}
> 	info.origScore = info.score
> ```

Just used for logging, but we never set it for L0. Makes the logs confusing.
___
*[`compaction_picker.go` line 1280 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2837#-Nc4U_t4BwGNaPw4niSi:-Nc4U_t57Dlhb318XMmc:bwm8bxc) ([raw file](https://github.com/cockroachdb/pebble/blob/eba5a250ef2c892a22d0af7a4807af768fbb2cf2/compaction_picker.go#L1280)):*
> ```Go
> 					p.vers.Levels[info.level].Iter(),
> 				))),
> 				humanize.Bytes.Int64(int64(totalSize(
> ```

Useful to see the actual size of the level in the logs.

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2837)* status: all files reviewed, 2 unresolved discussions (waiting on @bananabrick)

___
*[`compaction_picker.go` line 1280 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2837#-Nc4U_t4BwGNaPw4niSi:-NcSEVhP3CviBMYZH9Ly:b8hf9ay) ([raw file](https://github.com/cockroachdb/pebble/blob/eba5a250ef2c892a22d0af7a4807af768fbb2cf2/compaction_picker.go#L1280)):*

Useful to see the actual size of the level in the logs.
</blockquote></details>

We should not be using `totalCompensatedSize` and `totalSize` here since they are linear in the number of files in the level. We should do what `calculateScores` does: 
```
int64(levelCompensatedSize(p.vers.Levels[level]))

p.levelSizes[level]
```

and probably do the `sizeAdjust` logic too. We could make `calculateScores` return the `sizeAdjust` and use it here.

PR #2839, opened 2023-08-18
sharedcache: Deflake TestSharedCacheRandomized

itsbilal:
We can sometimes see flakes if rand.Int63 chooses a size of zero, as we feed that size into another Int63 call and it expects nonzero arguments. An example is https://github.com/cockroachdb/pebble/actions/runs/5907184465

This change fixes it by forcing a nonzero size.


RaduBerinde (commented):
:lgtm: 

Thanks for the fix!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2839#-:-Nc9fS0RErxMQkXP75NH:b3wstmi)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @joshimhoff)

RaduBerinde (approved):

joshimhoff:
Thanks for the fix!!

PR #2842, opened 2023-08-22
snapshot: Prevent indefinite wait on compact sync.Cond in EFOS

itsbilal:
Previously, if we went down the maybeScheduleFlush path instead of the maybeScheduleDelayedFlush path, we could end up in an indefinite wait for a compaction. This change updates a conditional there to ensure we have a flush happening before we wait on the flush conditional.

Also add a context.Context in WaitForFileOnly() to allow callers to use context cancellation to avoid busy-waiting if necessary.

### DIFF
--- a/ingest_test.go
+++ b/ingest_test.go
@@ -1047,7 +1047,7 @@ func TestIngestShared(t *testing.T) {
 				panic("insufficient args for file-only-snapshot command")
 			}
 			name := td.CmdArgs[0].Key
-			err := efos[name].WaitForFileOnlySnapshot(1 * time.Millisecond)
+			err := efos[name].WaitForFileOnlySnapshot(context.TODO(), 1*time.Millisecond)
 			if err != nil {
 				return err.Error()
 			}
@@ -1519,7 +1519,7 @@ func TestConcurrentExcise(t *testing.T) {
 				panic("insufficient args for file-only-snapshot command")
 			}
 			name := td.CmdArgs[0].Key
-			err := efos[name].WaitForFileOnlySnapshot(1 * time.Millisecond)
+			err := efos[name].WaitForFileOnlySnapshot(context.TODO(), 1*time.Millisecond)
 			if err != nil {
 				return err.Error()
 			}
--- a/scan_internal_test.go
+++ b/scan_internal_test.go
@@ -351,7 +351,7 @@ func TestScanInternal(t *testing.T) {
 			}
 			name := td.CmdArgs[0].Key
 			es := efos[name]
-			if err := es.WaitForFileOnlySnapshot(1 * time.Millisecond); err != nil {
+			if err := es.WaitForFileOnlySnapshot(context.TODO(), 1*time.Millisecond); err != nil {
 				return err.Error()
 			}
 			return "ok"
--- a/snapshot.go
+++ b/snapshot.go
@@ -366,7 +366,9 @@ func (es *EventuallyFileOnlySnapshot) releaseReadState() {
 // a delayed flush will be scheduled at that duration if necessary.
 //
 // Idempotent; can be called multiple times with no side effects.
-func (es *EventuallyFileOnlySnapshot) WaitForFileOnlySnapshot(dur time.Duration) error {
+func (es *EventuallyFileOnlySnapshot) WaitForFileOnlySnapshot(
+	ctx context.Context, dur time.Duration,
+) error {
 	es.mu.Lock()
 	if es.mu.vers != nil {
 		// Fast path.
@@ -384,14 +386,20 @@ func (es *EventuallyFileOnlySnapshot) WaitForFileOnlySnapshot(dur time.Duration)
 			return ErrClosed
 		default:
 		}
+		if err := ctx.Err(); err != nil {
+			return err
+		}
 		// Check if the current mutable memtable contains keys less than seqNum.
 		// If so, rotate it.
 		if es.db.mu.mem.mutable.logSeqNum < es.seqNum && dur.Nanoseconds() > 0 {
 			es.db.maybeScheduleDelayedFlush(es.db.mu.mem.mutable, dur)
+			es.db.mu.compact.cond.Wait()
 		} else {
 			es.db.maybeScheduleFlush()
+			if es.db.mu.compact.flushing {
+				es.db.mu.compact.cond.Wait()
+			}
 		}
-		es.db.mu.compact.cond.Wait()
 
 		earliestUnflushedSeqNum = es.db.getEarliestUnflushedSeqNumLocked()
 	}


sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2842)* status: all files reviewed, 2 unresolved discussions (waiting on @itsbilal)

___
*[`snapshot.go` line 395 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcXthd513Dd-7U0SMCU:-NcXthd6ADC4PK3w4Jl9:bqzfsm9) ([raw file](https://github.com/cockroachdb/pebble/blob/73cd5040c348c0d9101fc0935eac7ff384f96483/snapshot.go#L395)):*
> ```Go
> 		// If so, rotate it.
> 		if es.db.mu.mem.mutable.logSeqNum < es.seqNum && dur.Nanoseconds() > 0 {
> 			es.db.maybeScheduleDelayedFlush(es.db.mu.mem.mutable, dur)
> ```

- in the implementation of `maybeScheduleDelayedFlush` we have one path that does `d.makeRoomForWrite(nil)` which will set `flushForced` to true for that memtable, but it doesn't contain a call to `maybeScheduleFlush`. Is that an oversight?
- The other path in `maybeScheduleDelayedFlush` calls `maybeScheduleFlush()` but that may suffer from the same problem that this PR is trying to fix, in that it may not schedule the flush, so we could wait here when there is no flush scheduled.

Once a caller has marked an immutable memtable with `flushForced=true`, the only thing preventing it from being flushed is the `writerRefs`. So if every `writerUnref` is followed by a call to `maybeScheduleFlush()` (which it seems to be), then the original intended pattern of the code here of:
- set flushForced
- call maybeScheduleFlush
- cond wait

should be sufficient. Since that will either start the flush (for the memtable that had no `writerRefs`), or ensure that once `writerRefs` becomes 0, a flush will start.

Is the first bullet in this comment the actual problem, in that it did not follow this pattern?
___
*[`snapshot.go` line 399 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcXvy2mCUZArqNo1_Tq:-NcXvy2mCUZArqNo1_Tr:bdemnqp) ([raw file](https://github.com/cockroachdb/pebble/blob/73cd5040c348c0d9101fc0935eac7ff384f96483/snapshot.go#L399)):*
> ```Go
> 		} else {
> 			es.db.maybeScheduleFlush()
> 			if es.db.mu.compact.flushing {
> ```

iiuc, this will busy loop until the `writerRefs` are removed and the flush can actually happen.

RaduBerinde:
*[`snapshot.go` line 399 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcYHBo7BikswbhZUcQV:-NcYHBo7BikswbhZUcQW:b-elqxwd) ([raw file](https://github.com/cockroachdb/pebble/blob/73cd5040c348c0d9101fc0935eac7ff384f96483/snapshot.go#L399)):*
> ```Go
> 		} else {
> 			es.db.maybeScheduleFlush()
> 			if es.db.mu.compact.flushing {
> ```

Should this `if` be a `while`? That's the typical pattern around Wait

itsbilal (commented):
TFTR! Took a closer look at the failure I was seeing and you were right, this change wasn't helpful in its initial form.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2842#-:-NcYsYSY1GTVg4abjsuo:b-mkljjw)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`snapshot.go` line 395 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcXthd513Dd-7U0SMCU:-NcYt8CO7WU96HoXjeSk:b412cjz) ([raw file](https://github.com/cockroachdb/pebble/blob/73cd5040c348c0d9101fc0935eac7ff384f96483/snapshot.go#L395)):*

- in the implementation of `maybeScheduleDelayedFlush` we have one path that does `d.makeRoomForWrite(nil)` which will set `flushForced` to true for that memtable, but it doesn't contain a call to `maybeScheduleFlush`. Is that an oversight?
- The other path in `maybeScheduleDelayedFlush` calls `maybeScheduleFlush()` but that may suffer from the same problem that this PR is trying to fix, in that it may not schedule the flush, so we could wait here when there is no flush scheduled.

Once a caller has marked an immutable memtable with `flushForced=true`, the only thing preventing it from being flushed is the `writerRefs`. So if every `writerUnref` is followed by a call to `maybeScheduleFlush()` (which it seems to be), then the original intended pattern of the code here of:
- set flushForced
- call maybeScheduleFlush
- cond wait

should be sufficient. Since that will either start the flush (for the memtable that had no `writerRefs`), or ensure that once `writerRefs` becomes 0, a flush will start.

Is the first bullet in this comment the actual problem, in that it did not follow this pattern?
</blockquote></details>

Took a closer look and you're right, the issue was the first path which wasn't actually scheduling a flush. Hard to catch since usually some other writer will end up calling `maybeScheduleFlush` and then we'd be good, but one of the tests in Cockroach does no writes beyond this point and also wait for the WaitForFileOnly to finish, which was tripping on this.

Thanks!
___
*[`snapshot.go` line 399 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcYHBo7BikswbhZUcQV:-NcYsiru0qZRe71zBrsH:b-hkxamm) ([raw file](https://github.com/cockroachdb/pebble/blob/73cd5040c348c0d9101fc0935eac7ff384f96483/snapshot.go#L399)):*

Should this `if` be a `while`? That's the typical pattern around Wait
</blockquote></details>

The outer loop is basically a while, and it's best to loop around it and make all the checks it makes.
___
*[`snapshot.go` line 399 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcXvy2mCUZArqNo1_Tq:-NcYso8OD5Uv3ONWMKa_:b-gbh8wb) ([raw file](https://github.com/cockroachdb/pebble/blob/73cd5040c348c0d9101fc0935eac7ff384f96483/snapshot.go#L399)):*

iiuc, this will busy loop until the `writerRefs` are removed and the flush can actually happen.
</blockquote></details>

That's right. Previously this was a problem because if `flushing` was false, we were actually never dropping the db.mu (because the Wait() was not being run), so a flush would never be scheduled and we'd remain in the infinite loop forever. Fixed.

itsbilal:
Made another change to explicitly mark the relevant flushable as `flushForced` in the `maybeScheduleFlush` case in `WaitForFileOnlySnapshot` as that was also a missing case of an indefinite wait. PTAL.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2842)* status: 2 of 4 files reviewed, 5 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`snapshot.go` line 372 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcZIO3-E3v4fYKFmw2s:-NcZIO3-E3v4fYKFmw2t:baxvtwd) ([raw file](https://github.com/cockroachdb/pebble/blob/5e55bf5ca23fc998a05cf295580fc6672ba323eb/snapshot.go#L372)):*
> ```Go
> 	ctx context.Context, dur time.Duration,
> ) error {
> 	es.mu.Lock()
> ```

[nit] this first block could be a separate method on `es`, and it could use `defer es.mu.Unlock()`. Could have a `wait` flag so it replaces the last block as well
___
*[`snapshot.go` line 380 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcZIZ1p0LKT2i3u77IT:-NcZIZ1p0LKT2i3u77IU:b-y1yen8) ([raw file](https://github.com/cockroachdb/pebble/blob/5e55bf5ca23fc998a05cf295580fc6672ba323eb/snapshot.go#L380)):*
> ```Go
> 	es.mu.Unlock()
> 
> 	es.db.mu.Lock()
> ```

[nit] I would put this inside a `func() error` which defers the Unlock given that we have multiple exit paths that need to unlock
___
*[`snapshot.go` line 389 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcZIFgf2TFAN2NNkvVp:-NcZIFgf2TFAN2NNkvVq:b-gx8ihz) ([raw file](https://github.com/cockroachdb/pebble/blob/5e55bf5ca23fc998a05cf295580fc6672ba323eb/snapshot.go#L389)):*
> ```Go
> 		default:
> 		}
> 		if err := ctx.Err(); err != nil {
> ```

[nit] you can consider moving above to a `case <-ctx.Done` so it looks similar to the other case

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2842#-:-NccAnn43jeV7IyVxyl1:b-814g8k)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`snapshot.go` line 372 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcZIO3-E3v4fYKFmw2s:-Ncc8Sap706LFZ3wH5yp:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/5e55bf5ca23fc998a05cf295580fc6672ba323eb/snapshot.go#L372)):*

[nit] this first block could be a separate method on `es`, and it could use `defer es.mu.Unlock()`. Could have a `wait` flag so it replaces the last block as well
</blockquote></details>

Done.
___
*[`snapshot.go` line 380 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcZIZ1p0LKT2i3u77IT:-NccAp4w5H4x7bBH56Ri:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/5e55bf5ca23fc998a05cf295580fc6672ba323eb/snapshot.go#L380)):*

[nit] I would put this inside a `func() error` which defers the Unlock given that we have multiple exit paths that need to unlock
</blockquote></details>

Done.
___
*[`snapshot.go` line 389 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcZIFgf2TFAN2NNkvVp:-Ncc8_-iBCK8mJxLJS5L:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/5e55bf5ca23fc998a05cf295580fc6672ba323eb/snapshot.go#L389)):*

[nit] you can consider moving above to a `case <-ctx.Done` so it looks similar to the other case
</blockquote></details>

Done.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2842)* status: 2 of 4 files reviewed, 5 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`snapshot.go` line 372 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcchOA0-m8mwr-exiDr:-NcchOA0-m8mwr-exiDs:b-ncvsud) ([raw file](https://github.com/cockroachdb/pebble/blob/2f853d9a86f5a64eb1e764077758c2a973baaa4f/snapshot.go#L372)):*
> ```Go
> 	}
> 	if wait {
> 		es.mu.transitioned.Wait()
> ```

Heh, the new coverage shows that we don't exercise this in any unit tests, how hard would that be to add?
___
*[`snapshot.go` line 373 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcchFVKCBeuTEtvBily:-NcchFVKCBeuTEtvBilz:bzb5ir4) ([raw file](https://github.com/cockroachdb/pebble/blob/2f853d9a86f5a64eb1e764077758c2a973baaa4f/snapshot.go#L373)):*
> ```Go
> 	if wait {
> 		es.mu.transitioned.Wait()
> 		return true
> ```

Should we have an `invariants` check here that `vers` is not nil?
___
*[`snapshot.go` line 440 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcchpkT5zDZ9zQjlJO8:-NcchpkT5zDZ9zQjlJO9:b-3o5t53) ([raw file](https://github.com/cockroachdb/pebble/blob/2f853d9a86f5a64eb1e764077758c2a973baaa4f/snapshot.go#L440)):*
> ```Go
> 
> 	// Wait for transition to file-only snapshot.
> 	es.hasTransitioned(true /* wait */)
> ```

[nit] Hm, I realize this was my suggestion but this reads a bit strange.. May be nicer to have two functions `hasTransitioned` and `ensureTransitioned` (up to you).

sumeerbhola (approved):
:lgtm: once @RaduBerinde 's comments are addressed

> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2842#-:-NccnoT3-nf0ghg5cGCe:bt3z39w)* status: all files reviewed, 5 unresolved discussions (waiting on @itsbilal)

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2842#-:-NccwPkY2jK2On5Cc4aw:b-814g8k)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`snapshot.go` line 372 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcchOA0-m8mwr-exiDr:-Nccj2jLDl_wizlqzIUg:bo2kqqz) ([raw file](https://github.com/cockroachdb/pebble/blob/2f853d9a86f5a64eb1e764077758c2a973baaa4f/snapshot.go#L372)):*

Heh, the new coverage shows that we don't exercise this in any unit tests, how hard would that be to add?
</blockquote></details>

It's pretty challenging because usually the flushing goroutine ends up transitioning.

That actually got me thinking, we will never end up here since we just wait for the appropriate flush. I removed `transitioned` as it was unnecessary. 
___
*[`snapshot.go` line 373 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcchFVKCBeuTEtvBily:-Nccw-th8h-gLVtZvlew:bou6bbv) ([raw file](https://github.com/cockroachdb/pebble/blob/2f853d9a86f5a64eb1e764077758c2a973baaa4f/snapshot.go#L373)):*

Should we have an `invariants` check here that `vers` is not nil?
</blockquote></details>

Done. (In `WaitForFileOnlySnapshot` instead).
___
*[`snapshot.go` line 440 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcchpkT5zDZ9zQjlJO8:-NccwLaq0PmQw5pngRTW:blr12yt) ([raw file](https://github.com/cockroachdb/pebble/blob/2f853d9a86f5a64eb1e764077758c2a973baaa4f/snapshot.go#L440)):*

[nit] Hm, I realize this was my suggestion but this reads a bit strange.. May be nicer to have two functions `hasTransitioned` and `ensureTransitioned` (up to you).
</blockquote></details>

Done (irrelevant as the `ensureTransitioned` case no longer exists)

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2842#-:-Ncd-h-TEknGKSqRD1DG:bnfp4nl)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`snapshot.go` line 440 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2842#-NcchpkT5zDZ9zQjlJO8:-Ncd-jco-wmi683X7Hp6:b-83sjoh) ([raw file](https://github.com/cockroachdb/pebble/blob/2f853d9a86f5a64eb1e764077758c2a973baaa4f/snapshot.go#L440)):*

Done (irrelevant as the `ensureTransitioned` case no longer exists)
</blockquote></details>

Nice!

PR #2847, opened 2023-08-24
metamorphic: add support for EventuallyFileOnlySnapshots

### LINKED ISSUE #2781: metamorphic: Add support for EventuallyFileOnlySnapshots in metamorphic tests
A new type of snapshots, `EventuallyFileOnlySnapshots` were implemented in #2775. `EventuallyFileOnlySnapshots` retain the same guarantees as snapshots over the key-spans passed into them at creation time.

As part of this issue, we need to update Pebble metamorphic tests to add testing for EventuallyFileOnlySnapshots (EFOS). One way to do this is to have a test option that swaps the implementations of all snapshots with `EventuallyFileOnlySnapshot([]KeyRange{KeyMin-KeyMax})`. Another change would be to add support for creating / reading from EFOS as a toplevel object, and adding the relevant operations to open/close EFOS. The only challenge with this last part will be to ensure that iterators opened on an EFOS are scoped to just the keyranges passed in at creation time.

jbowens:
Add support for testing EventuallyFileOnlySnapshots in the metamorphic tests. This commit alters generation of the NewSnapshot operation to assign a random set of key ranges to a subset of snapshots. When a snapshot is constructed with a set of key ranges, all generated read options to the snapshot are constrained to fall within the snapshot's key ranges.

A new option within the TestOptions stanza is added to determine what subset of snapshots with key ranges are created as eventually-file only snapshots at runtime. This option is used as a component of a deterministic hash function of the operation index, enabling a deterministic subset of snapshots to be created as an eventually file-only snapshot.

Close #2781.


jbowens (commented):
Currently, the metamorphic tests are failing readily in `snapshot.Get` calls with divergences between EFOS and non-EFOS snapshots. The EFOS.Get implementation provided in this commit simply creates an iterator and seeks to it. I'm not yet sure if this is a legitimate bug with the implementation of EventuallyFileOnlySnapshots, or a bug in the integration into the metamorphic tests.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2847#-:-Ncd4Zbu9MP9vnKnlPyZ:b-sihykn)* status: 0 of 13 files reviewed, all discussions resolved (waiting on @itsbilal)

jbowens (commented):
Immediately after writing the comment I realized I forgot to check the iterator's key is indeed the sought key. Fixed.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2847#-:-Ncd5_sg7bYZAnsDL_u9:b5vimn2)* status: 0 of 13 files reviewed, all discussions resolved (waiting on @itsbilal)

itsbilal (approved):
LGTM, thanks for doing this!

jbowens:
TFTR!

PR #2854, opened 2023-08-27
Fix build for 32 bit architecture

anacrolix:
This fixes building on 32 bit architecture, without breaking any tests for 64 bit.


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2854)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @anacrolix and @itsbilal)

___
*[`open.go` line 42 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2854#-Ncxd9hyDi0_D5D1dW9i:-Ncxd9hyDi0_D5D1dW9j:b-xnjhtt) ([raw file](https://github.com/cockroachdb/pebble/blob/8918d4bf53d3359194a861d0a530f6871fc407e6/open.go#L42)):*
> ```Go
> 	// We limit the size to MaxUint32 so that the exclusive end of an allocation
> 	// fits in uint32.
> 	maxBatchSize = math.MaxInt // just short of 4 GB
> ```

I don't think we want this change for 64-bit systems
___
*[`internal/arenaskl/arena.go` line 46 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2854#-NcxdIPG6ja9Ra-5p9qe:-NcxdIPG6ja9Ra-5p9qf:bj34ktv) ([raw file](https://github.com/cockroachdb/pebble/blob/8918d4bf53d3359194a861d0a530f6871fc407e6/internal/arenaskl/arena.go#L46)):*
> ```Go
> // store.
> func NewArena(buf []byte) *Arena {
> 	if len(buf) > math.MaxInt {
> ```

We don't want this change for 64-bit systems.

If you want to limit the size of the arena to MaxInt for 32-bit systems, you need to change `maxMemTableSize` and the `MaxUint32` in `Size()`.
___
*[`internal/arenaskl/arena_test.go` line 36 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2854#-NcxfO_l3DkKMoE2tdw5:-NcxfO_l3DkKMoE2tdw6:b-z3jjnc) ([raw file](https://github.com/cockroachdb/pebble/blob/8918d4bf53d3359194a861d0a530f6871fc407e6/internal/arenaskl/arena_test.go#L36)):*
> ```Go
> func TestArenaSizeOverflow(t *testing.T) {
> 	if bits.UintSize == 32 {
> 		t.Skip("32 bit system")
> ```

Leave a TODO here, we would ideally make the test pass

anacrolix:
Ok I bit. I've addressed all your concerns.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2854)* status: 0 of 10 files reviewed, 6 unresolved discussions (waiting on @anacrolix and @itsbilal)

___
*[`go.mod` line 47 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2854#-NdFxOA0-QgfCAD9FusE:-NdFxOA0-QgfCAD9FusF:b-8ibjry) ([raw file](https://github.com/cockroachdb/pebble/blob/18295997116096e79af27dd552c3a5ba195ec3aa/go.mod#L47)):*
> ```XML
> )
> 
> go 1.21
> ```

We don't want this change right now. Cockroach is still using 1.19, and potentially other users of the library might be on 1.19 or 1.20.
___
*[`open.go` line 38 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2854#-NdFvF5g9N9vFgXUDIr4:-NdFvF5g9N9vFgXUDIr5:b-jswck9) ([raw file](https://github.com/cockroachdb/pebble/blob/0b401ee526b8ebcf9238bb7e1b953f40447003d1/open.go#L38)):*
> ```Go
> 	initialMemTableSize = 256 << 10 // 256 KB
> 
> 	// The max batch size is limited by the uint32 offsets stored in
> ```

These comments belong here
___
*[`internal/arenaskl/arena.go` line 38 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2854#-NdFuWMX231tmOtuIUdK:-NdFuWMX231tmOtuIUdL:b-xwe9si) ([raw file](https://github.com/cockroachdb/pebble/blob/18295997116096e79af27dd552c3a5ba195ec3aa/internal/arenaskl/arena.go#L38)):*
> ```Go
> 	nodeAlignment = 4
> 
> 	// MaxMemTableSize is limited by the uint32 offsets stored in internal/arenaskl.node,
> ```

This package is a data structure and has nothing to do with memtables or batch ops. I'd rename to `MaxSize` and fix the comment.

Also, `min` is new in go 1.21, we still need Pebble to build with 1.19 for now. You can use `math.MaxInt32 + (math.MaxUint>>63)<<31` and keep the nicer `min()` expression in a comment.
___
*[`internal/arenaskl/arena.go` line 42 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2854#-NdFx4I0-eYrlkVgui3y:-NdFx4I0-eYrlkVgui3z:b81s0q2) ([raw file](https://github.com/cockroachdb/pebble/blob/18295997116096e79af27dd552c3a5ba195ec3aa/internal/arenaskl/arena.go#L42)):*
> ```Go
> 	// end of an allocation fits in uint32, and also to the maximum possible contiguous block of
> 	// memory that can be allocated. Should be just shy of 4 GiB on 64 bit systems, and exactly 2 GiB
> 	// on 32 bit systems.
> ```

The explanation here for 32-bit doesn't sound right - the reason is that we use `int` instead of `uint32` in many places so we want the offsets to fit inside an `int`.
___
*[`internal/batchskl/skl.go` line 73 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2854#-NdFvvtB4B1CC5CLhh2w:-NdFvvtB4B1CC5CLhh2x:btbet5s) ([raw file](https://github.com/cockroachdb/pebble/blob/18295997116096e79af27dd552c3a5ba195ec3aa/internal/batchskl/skl.go#L73)):*
> ```Go
> 	maxNodeSize  = int(unsafe.Sizeof(node{}))
> 	linksSize    = int(unsafe.Sizeof(links{}))
> 	maxNodesSize = math.MaxInt32
> ```

This change limits the size to 2GB on 64-bit systems.. We want the same thing as the max arena size.

anacrolix:
I'm fine if you want to finish the PR off, I don't particularly care for the specifics.

RaduBerinde:
With all due respect, if you don't particularly care about specifics (some of which involve introducing bugs in the default 64-bit build of a production system), don't open PRs.

PR #2858, opened 2023-08-28
sstable, sharedcache: reduce test memory usage

### LINKED ISSUE #2849: Windows CI out-of-memory
```
2023-08-25T17:19:00.6769113Z runtime: VirtualAlloc of 2147483648 bytes failed with errno=1455
2023-08-25T17:19:00.6769214Z fatal error: out of memory
```
https://github.com/cockroachdb/pebble/actions/runs/5978602527/job/16221019736?pr=2845

RaduBerinde:
Fixes #2849.

#### sstable: reduce test memory usage

Reduce memory usage of TestGetBytes which allocates a 2GB slice. We
don't actually do anything with the data so we can allocate a small
slice.

#### sharedcache: reduce random test sizes

Reduce the maximum sharding block size to 4MB from 10MB (we always use
1MB in production) and reduce the maximum number of shards from 64 to
32. This reduces the memory footprint of the test significantly.


RaduBerinde:
TFTR!

PR #2212, opened 2023-01-03
db: flushable ingested sstable implementation

bananabrick:
This pr implements the feature mentioned in RFC: https://github.com/cockroachdb/pebble/pull/1586/files

Closes https://github.com/cockroachdb/pebble/issues/25

PR #2866, opened 2023-08-31
crl-release-22.2: db: don't clear logNum when appending flushable batch

RaduBerinde:
When we have a large batch we append it separately to the queue as a flushableBatch, with the same `logNum` as the last memtable.

The code sets the last memtable's `logNum` to 0. This is intended as a defensive measure (in case we flush the memtable without the flushableBatch) but it is incorrect: on flush, if the memtable is the first that is not ready to flush, its `logNum` will be chosen as the `minUnflushedLogNum` which will trigger the invariant violation error.

Note that this condition seems to be very rare, likely because large (~32MB) batches in Cockroach are rare and because it requires a writer reference to the memtable to last longer than it takes for the large batch to be flushed to the WAL.

The fix is to leave the memtable's `logNum` as is; together with the `logNum` invariant check at flush time, this ensures that the memtable and flushableBatch are always flushed together. Note that this is already what we're doing in 23.1 (as of #2212).

Informs #109078.

Release note (bug fix): fixed very rare condition where the storage engine stops making progress, reporting a "flush next log number is unset" background error.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1636,6 +1636,20 @@ func (d *DB) flush() {
 // d.mu must be held when calling this, but the mutex may be dropped and
 // re-acquired during the course of this method.
 func (d *DB) flush1() (bytesFlushed uint64, err error) {
+	// NB: Large batches placed in the flushable queue share the WAL with the
+	// previous memtable in the queue. We must ensure the property that both the
+	// large batch and the memtable with which it shares a WAL are flushed
+	// together. The property ensures that the minimum unflushed log number
+	// isn't incremented incorrectly. Since a flushableBatch.readyToFlush always
+	// returns true, and since the large batch will always be placed right after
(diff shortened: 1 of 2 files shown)


RaduBerinde:
Note that the invariant error was reproduced with an artificial delay added to the code (see https://github.com/cockroachdb/pebble/pull/2212#issuecomment-1701676215). We are still discussing what's the best way to improve our tests (likely metamorphic test) to trigger this condition. That will happen separately from this PR.

bananabrick (approved):
:lgtm:

This change should be safe.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2866#-:-NdCQ_l9EqVQ9o7I2crg:bm7bxde)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jbowens)

PR #2869, opened 2023-09-01
go.mod: downgrade golang.org/x/exp

jbowens:
Downgrade golang.org/x/exp to v0.0.0-20230626212559-97b1e661b5df, the version currently in use by the Cockroach master branch.


itsbilal (approved):
LGTM, thanks for the quick fix!

PR #2871, opened 2023-09-01
manifest: grab manifest lock when calling InitCompactingFileInfo

itsbilal:
Previously, we had one case where we could be calling InitCompactingFileInfo (a method on L0Sublevels that must be called on the most-recently-created L0Sublevels) on a failed compaction while another version was being installed but was in the process of writing to the manifest, and had hence dropped the db mutex during IO. This had the potential to panic on a case where we expect the interval indices in FileMetadata to match exactly with those in the latest version's L0Sublevels. If there's a newer L0Sublevels being created that hasn't been installed into the versions list yet, the two are expected to be out of sync.

This change resolves this by grabbing the manifest lock before calling InitCompactingFileInfo in clearCompactionState.

Will address cockroachdb/cockroach#109866 when it lands in cockroach.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1667,8 +1667,9 @@ func (d *DB) addInProgressCompaction(c *compaction) {
 // indicates whether the compaction state should be rolled back to its original
 // state in the case of an unsuccessful compaction.
 //
-// DB.mu must be held when calling this method. All writes to the manifest for
-// this compaction should have completed by this point.
+// DB.mu must be held when calling this method, however this method can drop and
+// re-acquire that mutex. All writes to the manifest for this compaction should
+// have completed by this point.
 func (d *DB) clearCompactingState(c *compaction, rollback bool) {
 	c.versionEditApplied = true
 	for _, cl := range c.inputs {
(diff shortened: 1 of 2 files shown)


jbowens (approved):
:lgtm:

In the roachtest, what causes the compactions to error out? A concurrent excise? I'm a little surprised we're seeing so many instances.

> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2871#-:-NdabP4-5AVyBJsuW_VT:b-75byyq)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @RaduBerinde)

itsbilal:
TFTR!

@jbowens Yes that's right, it's the concurrent excises that causes some compactions to error out, and that increases the chances of this bug being hit. We see a lot of concurrent-excise-based compaction cancellations very early in a node's lifetime when it only has one range or so in the LSM and so sstables are very wide, and that's usually when we hit this panic. The likelihood of it was lower before the L0CompactionConcurrency change.

PR #2872, opened 2023-09-03
tool: better errors in case of encrypted store

### LINKED ISSUE #2868: tool: improve output when running on encrypted store without the correct flags
If you run a command like `cockroach debug pebble db scan` without the `--enterprise-encryption` flag, you get bogus and very unhelpful output, and it seems it is on stdout too (not on stderr).

We should improve the output in this case.

RaduBerinde:
This commit improves the output of the pebble tool when it is run against an encrypted store and the encryption key is not set up correctly.

Until now, we get a very obscure "invalid key=value syntax" error from inside the options parsing code. What's worse the error contains unescaped non-printable characters which can end up erasing the message prefix altogether.

Finally, the errors show up on stdout instead of stderr.

This commit
 - escapes the non-printable characters and limits the length in the above error message
 - adds more context as to where the error is coming from
 - marks the error as a corruption error
 - augments all corruption errors with a specific note about encrypted stores
 - switches to using stderr for all `pebble db` subcommand errors.

Before:
```
$ go run ./cmd/pebble db scan /home/radu/local/1/data
<invalid UTF-8 garbage>
```

After:
```
$ go run ./cmd/pebble db scan /home/radu/local/1/data
error loading options: invalid key=value syntax: "\x96N\xa71$D\x81p\x1cƻ\xc1U\xee6\x88\x80\xd1\xdf\xf7R\x9c\xe2\xe5\xa2\xd2H\xb4\xd1\r+('W\xfeu\x9b\xbd1\xcafC\xd6\x01r\x80c..."
```

Fixes #2868.

### DIFF
--- a/open.go
+++ b/open.go
@@ -80,7 +80,7 @@ func Open(dirname string, opts *Options) (db *DB, _ error) {
 	// Open the database and WAL directories first.
 	walDirname, dataDir, walDir, err := prepareAndOpenDirs(dirname, opts)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "error opening database at %q", dirname)
 	}
 	defer func() {
 		if db == nil {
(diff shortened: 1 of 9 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2872)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`tool/db.go` line 287 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2872#-Nd_vawQ7rofOZak1EmV:-NdaDYnj5uDx4fIDT_D-:b1u11vl) ([raw file](https://github.com/cockroachdb/pebble/blob/8a762642213d5fd809f485f606c8484ebaeaf92e/tool/db.go#L287)):*

Since encryption-at-rest is implemented entirely within the Cockroach repository, is it possible to add this additional context within the Cockroach CLI code?
</blockquote></details>

Will do, I changed this PR to add an option for an "open error enhancer" instead.

PR #2876, opened 2023-09-05
crl-release-22.2: db: crash on logNum invariant violation

RaduBerinde:
The flushing code currently returns an error if a logNum invariant is violated. However, it's unlikely that we'll be able to recover from a bug here, and we end up keeping the node up but without making any forward progress.

We switch to crashing in this case - a panic is much more visible and restarting the node has a better chance of making progress anyway.


jbowens (approved):
lgtm

Issue #2853, opened 2023-08-27
32-bit platform support


anacrolix:
I'm getting this building on a 32bit system:

```
# github.com/cockroachdb/pebble/internal/batchskl
../../go/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20230826001808-0b401ee526b8/internal/batchskl/skl.go:310:18: maxNodesSize (untyped int constant 4294967295) overflows int
../../go/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20230826001808-0b401ee526b8/internal/batchskl/skl.go:320:16: cannot use maxNodesSize (untyped int constant 4294967295) as int value in assignment (overflows)
# github.com/cockroachdb/pebble/internal/testkeys
../../go/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20230826001808-0b401ee526b8/internal/testkeys/testkeys.go:30:46: cannot use math.MaxInt64 (untyped int constant 9223372036854775807) as int value in argument to fmt.Sprintf (overflows)
# github.com/cockroachdb/pebble/internal/arenaskl
../../go/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20230826001808-0b401ee526b8/internal/arenaskl/arena.go:46:16: math.MaxUint32 (untyped int constant 4294967295) overflows int
../../go/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20230826001808-0b401ee526b8/internal/arenaskl/arena.go:50:14: math.MaxUint32 (untyped int constant 4294967295) overflows int
```

I didn't look into it, but this is very common when an `int` is used rather than `int64` and is most likely an oversight.

anacrolix:
See https://github.com/cockroachdb/pebble/pull/2854.

RaduBerinde:
Hi,

I looked into it and while fixing the build is not hard, making all tests pass on 32-bit would be quite a bit of work. Without the tests passing, we would not recommend using a 32-bit build at all.

Can you describe your usecase a bit?

anacrolix:
I'm embedding an application using pebble onto Android.

PR #2877, opened 2023-09-05
testkeys: use int64 indexes for KeySpace

RaduBerinde:
This switches the KeySpace key indexes and timestamps to use int64, necessary to make things work on 32-bit.

Informs #2853.


RaduBerinde:
Thanks for the quick review!

PR #2884, opened 2023-09-06
compaction: maintain a ref of current version in runCompaction

itsbilal:
Previously, compaction-picking invariants guaranteed us that we would never delete a compaction's input file while it was in progress. However, the introduction of Excise changes this; a compaction's inputs could get excised, and we could consider those files as obsolete because compactions don't explicitly grab a ref of either the current version or readState otherwise.

This change introduces a version ref/unref in runCompaction so we don't try to delete a compaction's inputs during a compaction, only to panic because there are sstable iterator refs still on the underlying sstables.

Will fix cockroachdb/cockroach#110090 when it lands in cockroach master.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2795,6 +2795,22 @@ func (d *DB) runCompaction(
 	snapshots := d.mu.snapshots.toSlice()
 	formatVers := d.FormatMajorVersion()
 
+	if c.flushing == nil {
+		// Before dropping the db mutex, grab a ref to the current version. This
+		// prevents any concurrent excises from deleting files that this compaction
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2884#-:-NdgAnc_0GLaS0BVLb7M:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sumeerbhola)

itsbilal:
TFTR!

PR #2887, opened 2023-09-07
event: remove infinite recursion in compactionAnnotations.SafeFormat

itsbilal:
We have an infinite recursion where we try to repeatedly call SafeFormat on compactionAnnotations, as we pass the unmodified ca variable as-is to redact.Safe. This change addresses that by breaking up the []string and formatting it individually.


msbutler (approved):
My b! Sad that testing didn't catch that. How did you hit this?

itsbilal:
TFTR!

@msbutler the nightly metamorphic tests caught it! see #2886 and the #storage-ops channel on slack

PR #2890, opened 2023-09-08
metamorphic: fix testkeys.RandomSeparator

bananabrick:
The RandomSeparator function wasn't obeying its contract. If there is an id between apIdx and bpIdx, we use it now. If there is no such id, then we fix the computation of timestamps. Note that if there is no id between apIdx, and bpIdx, we will not necessarily respect the suffix parameter passed to the function. This is stated in the function contract.

Fixes: https://github.com/cockroachdb/pebble/issues/2874

### DIFF
--- a/internal/testkeys/testkeys.go
+++ b/internal/testkeys/testkeys.go
@@ -384,13 +384,17 @@ func computeAlphabetKeyIndex(key []byte, alphabet map[byte]int64, n int) int64 {
 	return ret
 }
 
+func abs(a int64) int64 {
+	if a < 0 {
+		return -a
+	}
+	return a
+}
+
 // RandomSeparator returns a random alphabetic key k such that a < k < b,
 // pulling randomness from the provided random number generator. If dst is
 // provided and the generated key fits within dst's capacity, the returned slice
 // will use dst's memory.
-//
-// The keys a and b must be purely alphabetic, consisting only of characters
-// a-z.
 //
 // If a prefix P exists such that Prefix(a) < P < Prefix(b), the generated key
 // will consist of the prefix P appended with the provided suffix. A zero suffix
@@ -405,6 +409,10 @@ func computeAlphabetKeyIndex(key []byte, alphabet map[byte]int64, n int) int64 {
 //
 // RandomSeparator panics if a or b fails to decode.
 func RandomSeparator(dst, a, b []byte, suffix int64, maxLength int, rng *rand.Rand) []byte {
+	if Comparer.Compare(a, b) >= 0 {
+		return nil
+	}
+
 	// Determine both keys' logical prefixes and suffixes.
 	ai := Comparer.Split(a)
 	bi := Comparer.Split(b)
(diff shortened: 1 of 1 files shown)


bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2890)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`internal/testkeys/testkeys.go` line 439 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2890#-NdmgfY2Dfx8aaqbahvA:-NdmgfY37ZbL_Kzz7qF_:b-aolclw) ([raw file](https://github.com/cockroachdb/pebble/blob/6f6852d52f18802da6f41fc7c8a4ed895110edd4/internal/testkeys/testkeys.go#L439)):*
> ```Go
> 			generatedIdx = rng.Int63n(diff) + apIdx + 1
> 		} else {
> 			generatedIdx = rng.Int63n(diff+1) + apIdx
> ```

The test was failing because in this code path we were generating an id equal to bpIdx, even though we were in a case where the diff between bpIdx and apIdx was > 1.

Because bpIdx was equal to generatedIdx, in the code path below, we ignore the suffix of 0 passed into the function, and instead gave the key a suffix of 3.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2890)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`internal/testkeys/testkeys.go` line 392 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2890#-NdmhyZoCT7zQr_FcM4J:-NdmhyZoCT7zQr_FcM4K:b-yv9lmn) ([raw file](https://github.com/cockroachdb/pebble/blob/6f6852d52f18802da6f41fc7c8a4ed895110edd4/internal/testkeys/testkeys.go#L392)):*
> ```Go
> // will use dst's memory.
> //
> // The keys a and b must be purely alphabetic, consisting only of characters
> ```

Removed this as a and b are also allowed to have a timestamp, and `@`.

bananabrick (commented):
The test failure was due to something like this:

```
initial bounds:
start oakpstog@0 ockm

Call to random separator with the keys nrij, and ockm, with the suffix 0, changes the lower bound to ockm@3. It was just unfortunate that the randomly generated key we picked between [nrij, ockm] happened to be ockm.
ockm@3 ockm

upper bound remains unchanged due to probability.
end ockm@3 ockm
```

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2890#-:-NdpSUVN9Wg3NOE_F4f4:bhyid8h)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jbowens)

internal/testkeys/testkeys.go
```diff
@@ -431,31 +439,56 @@ func RandomSeparator(dst, a, b []byte, suffix int64, maxLength int, rng *rand.Ra
 	diff := bpIdx - apIdx
 	generatedIdx := bpIdx
 	if diff > 0 {
+		var start int64 = diff + 1
```
itsbilal, line 442: nit: can we use a different name for `start`, maybe `delta` or so? it took me a while to understand that `base` is what I thought `start` would be, off the name.

itsbilal (approved):
LGTM

bananabrick:
Thanks

PR #2895, opened 2023-09-13
remoteobjcat: move assertions behind invariants.Enabled

### LINKED ISSUE #2894: Loading v23.2.0-alpha.00000000-4069-g0310d119569CC on disaggregated storage demo cluster has led to a crash loop
Rollout of `v23.2.0-alpha.00000000-4069-g0310d119569` on the CC disaggregated storage demo cluster has led to a crash loop.

```
ERROR: server startup failed: failed to start server: failed to create engines: pebble: could not open remote object catalog: pebble: error when loading remote object catalog file "REMOTE-OBJ-CATALOG-000035": version edit adds existing object 1503319
SQLSTATE: XX000
HINT: You have encountered an unexpected error.

Please check the public issue tracker to check whether this problem is
already tracked. If you cannot find it there, please report the error
with details by creating a new issue.

If you would rather not post publicly, please contact us directly
using the support form.

We appreciate your feedback.
Failed running "start"
```

Note that before I ran the rollout I replaced a node that was down for many days due to an unrelated infra issue with a new one. But that worked fine, and all nodes were healthy, until I updated the nodes to a new version. The new node updated first without any issues. The second node to update hit this crash loop. There are unavailable ranges, since the new node hadn’t finished rebalancing yet, but I don’t think think that is particularly interesting, as the new node plus the second node in a crash loop. 

Due to above, if we want to fix the existing cluster, we need to load a fix. If we clear the crashing node, we would lose data, since we only have two copies of it, due to the first node replacement being done right before the second node crashed.

RaduBerinde:
We recently added some assertions when applying a VersionEdit. Unfortunately we used to have a bug where when we rotate a catalog file, we dump the up-to-date state *and* the last version edit. This was a no-op, but now when we upgrade a node we can hit these assertions.

Fixes #2894.

### DIFF
--- a/objstorage/objstorageprovider/remoteobjcat/version_edit.go
+++ b/objstorage/objstorageprovider/remoteobjcat/version_edit.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/objstorage"
 	"github.com/cockroachdb/pebble/objstorage/remote"
 )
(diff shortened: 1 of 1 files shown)


itsbilal (approved):
LGTM, thanks for the quick fix!

Issue #112, opened 2019-05-03
db: blob storage / value separation


petermattis:
## Background

The [WiscKey](https://www.usenix.org/system/files/conference/fast16/fast16-papers-lu.pdf) paper showed the value of segregating storage for large values in order to reduce write amplification. [Badger](https://github.com/dgraph-io/badger) is a Go KV store implementing that design. RocksDB has support for [storing blobs](https://github.com/facebook/rocksdb/tree/master/utilities/blob_db) (binary large objects) outside of the main LSM. Unfortunately, this blob-db is incompatible with the normal RocksDB API, making it unusable for us as-is. This issue outlines how I think blob storage should be added to Pebble.

## Overview

The base WiscKey idea is that instead of storing large values inline in the LSM, they are stored in a separate value-log. The LSM merely contains a value-pointer. Storing large values outside of the LSM significantly reduces the size of the LSM which improves caching and reduces the overhead of compactions.

The preliminary blob support in RocksDB has outlined the basic structure for how this could work in Pebble. A new
`InternalKeyKindBlobIndex` is added indicating that the value for a key is stored elsewhere. The value looks like:

```
+-------------+----------+----------+
| file number | offset   | size     |
+-------------+----------+----------+
| varint64    | varint64 | varint64 |
+-------------+----------+----------+
```

(Note that RocksDB also provides support for a TTL on blobs, but I don't think this is usable for CockroachDB and therefore not needed in Pebble).

Blobs are stored in a series of blob logs. A background process periodically walks over the blob logs, reading the entries and checking to see if the value is still "live" in the LSM. This process is called value GC. If the value is still "live", it is written to the active blob log, and the LSM is updated.

## Critique

An LSM does not allow mutating existing entries. In order to change the value for a key, a new value needs to be written which masks the old value. In Pebble/RocksDB/Badger, each version of a key/value pair has an associated sequence number (timestamp in Badger). RocksDB/Pebble take advantage of these sequence numbers to provide immutable snapshots. But the immutability of existing values presents a problem for blob storage. How do we update the LSM when a value is moved from one blob log to another? AFAICT, RocksDB's blob-db punts on this issue and just updates the current version. Badger at least attempts to deal with the problem:

```
// IMPORTANT: We should never write an entry with an older timestamp
// for the same key, We need to maintain this invariant to search for
// the latest value of a key, or else we need to search in all tables
// and find the max version among them.  To maintain this invariant,
// we also need to ensure that all versions of a key are always
[...]

PR #2897, opened 2023-09-14
db: prototype of separation of values into blob files

sumeerbhola:
The goal is to reduce write amp without significantly sacrificing locality of access.

Preliminary performance numbers are promising (see TestWriteAmpWithBlobs) which achieves a write amp of 6.8 (versus 15.1 with no blob files) with Lbase == L3 and a uniform random write workload.

There are numerous TODOs, which can be addressed when turning this into production code.

Suggested reading order is the sstable package, then the manifest package and version_set.go, and then compaction.go.

For workloads that care less about locality (e.g. MVCCGet), we could tune for further write amp improvements, but it may also increase the number of files to an unacceptable value -- unlike some other DBs, CockroachDB's use of Pebble grows the target file size by a factor of 2 when descending the levels, so reusing blob files written at higher levels results in higher file count.

Informs #112


sumeerbhola (commented):
The prototype was begun in Jan and I have not bothered rebasing, hence the conflicts.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2897#-:-NeG4df36fFBv5xrm3NU:b-2txvr0)* status: 0 of 68 files reviewed, all discussions resolved

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2897)* status: 0 of 68 files reviewed, 1 unresolved discussion

___
*[`db_test.go` line 1576 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2897#-Ng_b_OaCOMxlo8fXzgY:-Ng_b_Ob-TqWrSes7kxa:beixhrt) ([raw file](https://github.com/cockroachdb/pebble/blob/06743e63fdff79cf1a4a8a8c046828353764bfc1/db_test.go#L1576)):*
> ```Go
> BenchmarkReadWithBlobs/write-blobs=true/read-value=true/has-cache=false-10           	 5000000	       251.5 ns/op	         0.03252 misses/op	       0 B/op	       0 allocs/op
> BenchmarkReadWithBlobs/write-blobs=true/read-value=true/has-cache=true-10            	 5000000	       127.6 ns/op	         0.003208 misses/op	       0 B/op	       0 allocs/op
> ```

Read numbers are very promising.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2897)* status: 0 of 68 files reviewed, 2 unresolved discussions

___
*[`db_test.go` line 1578 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2897#-Ng_ev0Z5WIflmDVy_Ur:-Ng_ev0Z5WIflmDVy_Us:b-exmq1h) ([raw file](https://github.com/cockroachdb/pebble/blob/06743e63fdff79cf1a4a8a8c046828353764bfc1/db_test.go#L1578)):*
> ```Go
> 
> TODO(sumeer): the 127.6 ns/op in the last bench versus its counterpart with 172.4 ns/op needs
> investigation. I didn't expect blobs to be faster.
> ```

There is nothing special about the cpu profiles. With blob files the calls to `findNextEntry` and `nextUserKey` are much faster. It may be a CPU caching effect, since the 1000 byte value is not inline.

PR #2898, opened 2023-09-14
db: use invalidating iterators with low probability under invariants

### LINKED ISSUE #2896: metamorphic: use invalidating iterators in some runs
In some unit tests we use 'invalidating iterators' to ensure we're not relying on key/buffer stability where it's not guaranteed. Some runs of the metamorphic test could configure iterators with similar mechanics.

One possible explanation of cockroachlabs/support#2583 is corruption during compaction iteration. If there is a memory lifetime issue, use of an invalidating iterator in the metamorphic tests may help shake it out.

jbowens:
In invariants builds, with 10% probability wrap various internal iterators with
an invalidating iterator. This helps ensure we're not accidentally depending on
key byte slice stability where we shouldn't.

Close #2896.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/invalidating"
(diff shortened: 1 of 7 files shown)


itsbilal (approved):
LGTM :shipit:

jbowens:
TFTR!

PR #2900, opened 2023-09-14
metamorphic: vary sstable compression algorithm

### LINKED ISSUE #1705: metamorphic: consider running with configurations that alter the compression library
We currently test with the default codec (snappy at the time of writing). To give us better coverage, we should consider testing against configurations with a) a different codec to `standard-000`, and b) an LSM with a mixture of codecs for various levels (as opposed to the same library for all levels).

We'd update the code here, to set the codec, and possibly have a mixture of level configs: https://github.com/cockroachdb/pebble/blob/3355a02e7cec7a4cbf775421a89dfbe833818266/internal/metamorphic/options.go#L264-L270

jbowens:
In the metamorphic test runs with randomized options, vary the sstable compression algorithm rather than always using the default Snappy compression algorithm.

Close #1705.

### DIFF
--- a/metamorphic/options.go
+++ b/metamorphic/options.go
@@ -488,6 +488,15 @@ func randomOptions(
 	default:
 		lopts.FilterPolicy = newTestingFilterPolicy(1 << rng.Intn(5))
 	}
+	// We use either no compression, snappy compression or zstd compression.
+	switch rng.Intn(3) {
+	case 0:
+		lopts.Compression = pebble.NoCompression
+	case 1:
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2900#-:-NeLPQFT12chxwSZeDS4:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sumeerbhola)

jbowens:
TFTR!

PR #2901, opened 2023-09-14
db: double check file reference counts when loading file

jbowens:
Double check the file reference counts before attempting to find/create a table
cache node for a file. Once a file's reference count falls to zero, the file
becomes obsolete and may be deleted at any moment.

Today if we have a race, break this invariant and attempt to load a file with a
nonpositive reference count, it's relatively unlikely that it manifests as an
error. Typically tables remain open in the table cache, allowing the table
cache to serve the request even if the file is no longer linked into the data
directory. Additionally, even if it's not in the table cache presently,
deletion of obsolete files may be delayed due to deletion pacing, hiding the
race.

This commit preemptively asserts on the file reference counts. I opted for not
restricting this invariant check to invariants builds because it's cheap
relative to a table cache lookup, and it's a particularly tricky form of
corruption to debug otherwise.

Informs https://github.com/cockroachdb/cockroach/issues/110645.

### DIFF
--- a/flushable_test.go
+++ b/flushable_test.go
@@ -59,13 +59,18 @@ func TestIngestedSSTFlushableAPI(t *testing.T) {
 		// We can reuse the ingestLoad function for this test even if we're
 		// not actually ingesting a file.
 		lr, err := ingestLoad(d.opts, d.FormatMajorVersion(), paths, nil, nil, d.cacheID, pendingOutputs, d.objProvider, jobID)
-		meta := lr.localMeta
 		if err != nil {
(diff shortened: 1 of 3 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2901#-:-NeLPo5S7zu6p-m9jj3e:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved

jbowens:
TFTR!

Issue #389, opened 2019-11-10
perf: skip output level sstables with no data overlap with input level


petermattis:
See https://github.com/facebook/rocksdb/pull/6021.

The idea is that if we're doing a compaction from Ln->Ln+1, we can skip sstables in Ln+1 which do contain any keys in Ln. Right now this determination is done solely based on the sstable boundaries, but it is only a little more work to create a `levelIter` on the input sstables in Ln and seek to the boundary keys in Ln+1 to see if any actual key is contained within the sstables in Ln+1.

Here's an example of where this could be beneficial:

```
L0: a-----g------------t-----z
L1:   c----h j------q s-----y        
```

L0 has one sstable with the bounds `[a,z]` containing the keys `a`, `g`, `t`, and `z`. A compaction from L0 to L1 would normally involve every sstable in L1 because the bounds `[a,z]` covers every sstable in L1. But notice that the L1 sstable `[j,q]` doesn't overlap with any keys in L0.

After selecting the initial set of input sstables in L0 and L1, we could open a `levelIter` on the sstables in L0 and seek that iterator to the boundary keys of the tables in L1, trimming the L1 inputs of any sstables that do not contain an overlapping key in L0. The compaction logic would have to be changed to ensure that the tables output to L1 are cut so as not to overlap with the excluded sstables in L1.

Cc @sumeerbhola as this overlaps with other thinking we've been doing in this area. I'm not sure how often this situation occurs in practice. We'd want to do some experimentation to determine if it is common or rare.

sumeerbhola:
Is the first step to add some stats where we do this seeking and record how many sstables and bytes could have been eliminated? Is it reasonable to add these stats for every compaction, or do you expect this to be slow enough that one should sample?

petermattis:
> Is the first step to add some stats where we do this seeking and record how many sstables and bytes could have been eliminated? Is it reasonable to add these stats for every compaction, or do you expect this to be slow enough that one should sample?

Yes, I think that is the first step. I think the stats could be added for every compaction. The seeks should be fast enough that they will be lost in the noise of the compaction itself.

petermattis:
There will probably be dragons in the handling of range tombstones that overlap the skipped sstable.

PR #2903, opened 2023-09-14
*: implement local to remote/shared copy compaction

### LINKED ISSUE #2626: compaction: implement local -> shared move compaction
Currently, if we make a move compaction from a local level to a shared level, we keep the file where it was and just move the FileMetadata between levels. To work around this limitation, we've been preferring that all sstables go on shared storage if they can, in the off chance they get moved to L5/L6 down the line. This is slow and potentially wasteful, such as during ingestion where it would be preferable to leave local files locally instead of copying them to shared storage even if they will never be moved below L5/L6. The other option currently is to keep files local even when they get move-compacted to L5/L6, at the expense of having more `ScanInternal` calls return `ErrInvalidSkipSharedIteration` because of the presence of non-shared files in lower levels.

A middle ground is to change `runCompaction` to move files to shared storage if we're doing a move compaction from L[0-4] to L[5-6]. This can then be followed with setting `preferSharedStorage` to be whether the destination file is going into L5-6 rightaway, speeding up ingestions.

itsbilal:
This change adds the ability to do a copy compaction for a file from local to remote storage, by copying the underlying file and producing a new filenum + backing file. This gives us the flexibility to specify finer-grained placement rules for files in local vs remote storage, and we can use this to selectively place only L5/6 in shared storage, an option that is now Options.Experimental.CreateOnShared (which becomes an enum as opposed to the previous bool).

Fixes #2626.

### DIFF
--- a/cmd/pebble/db.go
+++ b/cmd/pebble/db.go
@@ -105,7 +105,7 @@ func newPebbleDB(dir string) DB {
 			// Store all shared objects on local disk, for convenience.
 			"": remote.NewLocalFS(pathToLocalSharedStorage, vfs.Default),
 		})
-		opts.Experimental.CreateOnShared = true
+		opts.Experimental.CreateOnShared = remote.CreateOnSharedAll
 		if secondaryCacheSize != 0 {
 			opts.Experimental.SecondaryCacheSizeBytes = secondaryCacheSize
 		}
--- a/compaction.go
+++ b/compaction.go
@@ -19,6 +19,7 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/invalidating"
+	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/internal/keyspan"
 	"github.com/cockroachdb/pebble/internal/manifest"
 	"github.com/cockroachdb/pebble/internal/private"
@@ -26,6 +27,7 @@ import (
 	"github.com/cockroachdb/pebble/internal/rangekey"
 	"github.com/cockroachdb/pebble/objstorage"
 	"github.com/cockroachdb/pebble/objstorage/objstorageprovider/objiotracing"
+	"github.com/cockroachdb/pebble/objstorage/remote"
 	"github.com/cockroachdb/pebble/sstable"
 	"github.com/cockroachdb/pebble/vfs"
 	"golang.org/x/exp/constraints"
@@ -460,7 +462,12 @@ type compactionKind int
 const (
 	compactionKindDefault compactionKind = iota
 	compactionKindFlush
+	// compactionKindMove denotes a move compaction where the input file is
+	// retained and linked in a new level without being obsoleted.
 	compactionKindMove
+	// compactionKindCopy denotes a copy compaction where the input file is
+	// copied byte-by-byte into a new file with a new FileNum in the output level.
+	compactionKindCopy
 	compactionKindDeleteOnly
 	compactionKindElisionOnly
 	compactionKindRead
@@ -486,6 +493,8 @@ func (k compactionKind) String() string {
 		return "rewrite"
 	case compactionKindIngestedFlushable:
 		return "ingested-flushable"
+	case compactionKindCopy:
+		return "copy"
 	}
 	return "?"
 }
@@ -712,7 +721,9 @@ func (c *compaction) makeInfo(jobID int) CompactionInfo {
 	return info
 }
 
-func newCompaction(pc *pickedCompaction, opts *Options, beganAt time.Time) *compaction {
+func newCompaction(
+	pc *pickedCompaction, opts *Options, beganAt time.Time, provider objstorage.Provider,
+) *compaction {
 	c := &compaction{
 		kind:              compactionKindDefault,
 		cmp:               pc.cmp,
@@ -747,15 +758,41 @@ func newCompaction(pc *pickedCompaction, opts *Options, beganAt time.Time) *comp
 			c.smallest.UserKey, c.largest.UserKey, c.largest.IsExclusiveSentinel())
 	}
 	c.setupInuseKeyRanges()
-
 	c.kind = pc.kind
+
 	if c.kind == compactionKindDefault && c.outputLevel.files.Empty() && !c.hasExtraLevelData() &&
 		c.startLevel.files.Len() == 1 && c.grandparents.SizeSum() <= c.maxOverlapBytes {
-		// This compaction can be converted into a trivial move from one level
+		// This compaction can be converted into a move or copy from one level
 		// to the next. We avoid such a move if there is lots of overlapping
 		// grandparent data. Otherwise, the move could create a parent file
 		// that will require a very expensive merge later on.
-		c.kind = compactionKindMove
+		iter := c.startLevel.files.Iter()
+		meta := iter.First()
+		isRemote := false
+		// We should always be passed a provider, except in some unit tests.
+		if provider != nil {
+			objMeta, err := provider.Lookup(fileTypeTable, meta.FileNum.DiskFileNum())
+			if err != nil {
+				panic(errors.Wrapf(err, "cannot lookup table %s in provider", meta.FileNum))
+			}
+			isRemote = objMeta.IsRemote()
+		}
+		// Avoid a trivial move or copy if all of these are true, as rewriting a
+		// new file is better:
+		//
+		// 1) The source file is a virtual sstable
+		// 2) The existing file `meta` is on non-remote storage
+		// 3) The output level prefers shared storage
+		mustCopy := !isRemote && remote.ShouldCreateShared(opts.Experimental.CreateOnShared, c.outputLevel.level)
+		if mustCopy {
+			// If the source is virtual, it's best to just rewrite the file as all
+			// conditions in the above comment are met.
+			if !meta.Virtual {
+				c.kind = compactionKindCopy
+			}
+		} else {
+			c.kind = compactionKindMove
+		}
 	}
 	return c
 }
@@ -2349,7 +2386,7 @@ func (d *DB) maybeScheduleCompactionPicker(
 		env.inProgressCompactions = d.getInProgressCompactionInfoLocked(nil)
 		pc, retryLater := pickManualCompaction(v, d.opts, env, d.mu.versions.picker.getBaseLevel(), manual)
 		if pc != nil {
-			c := newCompaction(pc, d.opts, d.timeNow())
+			c := newCompaction(pc, d.opts, d.timeNow(), d.ObjProvider())
 			d.mu.compact.manual = d.mu.compact.manual[1:]
 			d.mu.compact.compactingCount++
 			d.addInProgressCompaction(c)
@@ -2376,7 +2413,7 @@ func (d *DB) maybeScheduleCompactionPicker(
 		if pc == nil {
 			break
 		}
-		c := newCompaction(pc, d.opts, d.timeNow())
+		c := newCompaction(pc, d.opts, d.timeNow(), d.ObjProvider())
 		d.mu.compact.compactingCount++
 		d.addInProgressCompaction(c)
 		go d.compact(c, nil)
@@ -2780,6 +2817,77 @@ type compactStats struct {
 	countMissizedDels    uint64
 }
 
+// runCopyCompaction runs a copy compaction where a new FileNum is created that
+// is a byte-for-byte copy of the input file. This is used in lieu of a move
+// compaction when a file is being moved across the local/remote storage
+// boundary.
+//
+// d.mu must be held when calling this method.
+func (d *DB) runCopyCompaction(
+	jobID int,
+	c *compaction,
+	meta *fileMetadata,
+	objMeta objstorage.ObjectMetadata,
+	versionEdit *versionEdit,
+) (ve *versionEdit, pendingOutputs []physicalMeta, retErr error) {
+	ve = versionEdit
+	if objMeta.IsRemote() || !remote.ShouldCreateShared(d.opts.Experimental.CreateOnShared, c.outputLevel.level) {
+		panic("pebble: scheduled a copy compaction that is not actually moving files to shared storage")
+	}
+	// Note that based on logic in the compaction picker, we're guaranteed
+	// meta.Virtual is false.
+	if meta.Virtual {
+		panic(errors.AssertionFailedf("cannot do a copy compaction of a virtual sstable across local/remote storage"))
+	}
+	// We are in the relatively more complex case where we need to copy this
+	// file to remote/shared storage. Drop the db mutex while we do the
+	// copy.
+	//
+	// To ease up cleanup of the local file and tracking of refs, we create
+	// a new FileNum. This has the potential of making the block cache less
+	// effective, however.
+	metaCopy := new(fileMetadata)
+	*metaCopy = fileMetadata{
+		Size:           meta.Size,
+		CreationTime:   meta.CreationTime,
+		SmallestSeqNum: meta.SmallestSeqNum,
+		LargestSeqNum:  meta.LargestSeqNum,
+		Stats:          meta.Stats,
+		Virtual:        meta.Virtual,
+	}
+	if meta.HasPointKeys {
+		metaCopy.ExtendPointKeyBounds(c.cmp, meta.SmallestPointKey, meta.LargestPointKey)
+	}
+	if meta.HasRangeKeys {
+		metaCopy.ExtendRangeKeyBounds(c.cmp, meta.SmallestRangeKey, meta.LargestRangeKey)
+	}
+	metaCopy.FileNum = d.mu.versions.getNextFileNum()
+	metaCopy.InitPhysicalBacking()
+	c.metrics = map[int]*LevelMetrics{
+		c.outputLevel.level: {
+			BytesIn:         meta.Size,
+			BytesCompacted:  meta.Size,
+			TablesCompacted: 1,
+		},
+	}
+	pendingOutputs = append(pendingOutputs, metaCopy.PhysicalMeta())
+
+	d.mu.Unlock()
+	defer d.mu.Lock()
+	_, err := d.objProvider.LinkOrCopyFromLocal(context.TODO(), d.opts.FS,
+		d.objProvider.Path(objMeta), fileTypeTable, metaCopy.FileBacking.DiskFileNum,
+		objstorage.CreateOptions{PreferSharedStorage: true})
+	if err != nil {
+		return ve, pendingOutputs, err
+	}
+	ve.NewFiles[0].Meta = metaCopy
+
+	if err := d.objProvider.Sync(); err != nil {
+		return nil, pendingOutputs, err
+	}
+	return ve, pendingOutputs, nil
+}
+
 // runCompactions runs a compaction that produces new on-disk tables from
 // memtables or old on-disk tables.
 //
@@ -2828,13 +2936,22 @@ func (d *DB) runCompaction(
 		panic("pebble: runCompaction cannot handle compactionKindIngestedFlushable.")
 	}
 
-	// Check for a trivial move of one table from one level to the next. We avoid
+	// Check for a move or copy of one table from one level to the next. We avoid
 	// such a move if there is lots of overlapping grandparent data. Otherwise,
 	// the move could create a parent file that will require a very expensive
 	// merge later on.
-	if c.kind == compactionKindMove {
+	if c.kind == compactionKindMove || c.kind == compactionKindCopy {
 		iter := c.startLevel.files.Iter()
 		meta := iter.First()
+		if invariants.Enabled {
+			if iter.Next() != nil {
+				panic("got more than one file for a move or copy compaction")
+			}
+		}
+		objMeta, err := d.objProvider.Lookup(fileTypeTable, meta.FileNum.DiskFileNum())
+		if err != nil {
+			return ve, pendingOutputs, stats, err
+		}
 		c.metrics = map[int]*LevelMetrics{
 			c.outputLevel.level: {
 				BytesMoved:  meta.Size,
@@ -2848,6 +2965,12 @@ func (d *DB) runCompaction(
 			NewFiles: []newFileEntry{
 				{Level: c.outputLevel.level, Meta: meta},
 			},
+		}
+		if c.kind == compactionKindCopy {
+			ve, pendingOutputs, retErr = d.runCopyCompaction(jobID, c, meta, objMeta, ve)
+			if retErr != nil {
+				return ve, pendingOutputs, stats, retErr
+			}
 		}
 		return ve, nil, stats, nil
 	}
@@ -3029,15 +3152,8 @@ func (d *DB) runCompaction(
 			}
 		}
 		// Prefer shared storage if present.
-		//
-		// TODO(bilal): This might be inefficient for short-lived files in higher
-		// levels if we're only writing to shared storage and not double-writing
-		// to local storage. Either implement double-writing functionality, or
-		// set PreferSharedStorage to c.outputLevel.level >= 5. The latter needs
-		// some careful handling around move compactions to ensure all files in
-		// lower levels are in shared storage.
 		createOpts := objstorage.CreateOptions{
-			PreferSharedStorage: true,
+			PreferSharedStorage: remote.ShouldCreateShared(d.opts.Experimental.CreateOnShared, c.outputLevel.level),
 		}
 		writable, objMeta, err := d.objProvider.Create(ctx, fileTypeTable, fileNum.DiskFileNum(), createOpts)
 		if err != nil {
--- a/compaction_picker_test.go
+++ b/compaction_picker_test.go
@@ -536,7 +536,7 @@ func TestCompactionPickerL0(t *testing.T) {
 			var result strings.Builder
 			if pc != nil {
 				checkClone(t, pc)
-				c := newCompaction(pc, opts, time.Now())
+				c := newCompaction(pc, opts, time.Now(), nil /* provider */)
 				fmt.Fprintf(&result, "L%d -> L%d\n", pc.startLevel.level, pc.outputLevel.level)
 				fmt.Fprintf(&result, "L%d: %s\n", pc.startLevel.level, fileNums(pc.startLevel.files))
 				if !pc.outputLevel.files.Empty() {
@@ -756,7 +756,7 @@ func TestCompactionPickerConcurrency(t *testing.T) {
 			})
 			var result strings.Builder
 			if pc != nil {
-				c := newCompaction(pc, opts, time.Now())
+				c := newCompaction(pc, opts, time.Now(), nil /* provider */)
 				fmt.Fprintf(&result, "L%d -> L%d\n", pc.startLevel.level, pc.outputLevel.level)
 				fmt.Fprintf(&result, "L%d: %s\n", pc.startLevel.level, fileNums(pc.startLevel.files))
 				if !pc.outputLevel.files.Empty() {
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -515,7 +515,8 @@ func TestPickCompaction(t *testing.T) {
 		vs.picker = &tc.picker
 		pc, got := vs.picker.pickAuto(compactionEnv{diskAvailBytes: math.MaxUint64}), ""
 		if pc != nil {
-			c := newCompaction(pc, opts, time.Now())
+			c := newCompaction(pc, opts, time.Now(), nil /* provider */)
+
 			gotStart := fileNums(c.startLevel.files)
 			gotML := ""
 			observedMulti := len(c.extraLevels) > 0
@@ -1855,7 +1856,7 @@ func TestCompactionOutputLevel(t *testing.T) {
 				d.ScanArgs(t, "start", &start)
 				d.ScanArgs(t, "base", &base)
 				pc := newPickedCompaction(opts, version, start, defaultOutputLevel(start, base), base)
-				c := newCompaction(pc, opts, time.Now())
+				c := newCompaction(pc, opts, time.Now(), nil /* provider */)
 				return fmt.Sprintf("output=%d\nmax-output-file-size=%d\n",
 					c.outputLevel.level, c.maxOutputFileSize)
 
@@ -3828,7 +3829,7 @@ func TestSharedObjectDeletePacing(t *testing.T) {
 	opts.Experimental.RemoteStorage = remote.MakeSimpleFactory(map[remote.Locator]remote.Storage{
 		"": remote.NewInMem(),
 	})
-	opts.Experimental.CreateOnShared = true
+	opts.Experimental.CreateOnShared = remote.CreateOnSharedAll
 	opts.TargetByteDeletionRate = 1
 
 	d, err := Open("", &opts)
--- a/ingest_test.go
+++ b/ingest_test.go
@@ -786,7 +786,9 @@ func TestExcise(t *testing.T) {
 	})
 }
 
-func TestIngestShared(t *testing.T) {
+func testIngestSharedImpl(
+	t *testing.T, createOnShared remote.CreateOnSharedStrategy, fileName string,
+) {
 	var d, d1, d2 *DB
 	var efos map[string]*EventuallyFileOnlySnapshot
 	defer func() {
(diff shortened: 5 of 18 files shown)


itsbilal:
This PR looks big off of the line count, but the bulk of that is the replication of `testdata/ingest_shared` into `testdata/ingest_shared_lower`.

jbowens (commented):
> Reviewed 3 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2903)* status: 3 of 18 files reviewed, 4 unresolved discussions (waiting on @itsbilal)

___
*[`compaction.go` line 2777 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2903#-Nehpr5sCzTPOLm8dMhm:-Nehpr5sCzTPOLm8dMhn:bk0r1zv) ([raw file](https://github.com/cockroachdb/pebble/blob/d400232a35f2b3b314d4f8bc853e8e4caaf631e1/compaction.go#L2777)):*
> ```Go
> 	if shouldDoTrivialMove {
> 		iter := c.startLevel.files.Iter()
> 		trivialMoveMeta = iter.First()
> ```

Today we'll only produce a move compaction with a single start-level file, but that might not always be the case. Can we either perform this for every file in `iter`, or assert that `iter.Next()` exhausts the iterator?
___
*[`compaction.go` line 2789 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NehqDiwEvcLVpK4ZwDB:-NehqDiwEvcLVpK4ZwDC:b-r5mvb4) ([raw file](https://github.com/cockroachdb/pebble/blob/d400232a35f2b3b314d4f8bc853e8e4caaf631e1/compaction.go#L2789)):*
> ```Go
> 		// 2) The existing file `trivialMoveMeta` is on non-remote storage
> 		// 3) The output level prefers shared storage
> 		shouldDoTrivialMove = !trivialMoveMeta.Virtual || objMeta.IsRemote() || !d.objProvider.ShouldCreateShared(c.outputLevel.level)
> ```

Can we move this determination into compaction picking? It's bit wonky to create a compaction marked as a move compaction, only not to perform a move. Then we can also remove the `isTrivialMove` field.
___
*[`compaction.go` line 2812 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NehrKcP7gRAyv1bPfJr:-NehrKcP7gRAyv1bPfJs:bg5lekx) ([raw file](https://github.com/cockroachdb/pebble/blob/d400232a35f2b3b314d4f8bc853e8e4caaf631e1/compaction.go#L2812)):*
> ```Go
> 			return ve, pendingOutputs, stats, err
> 		}
> 		if !objMeta.IsRemote() && d.objProvider.ShouldCreateShared(c.outputLevel.level) {
> ```

Should we pull this `if` block out into a helper? `runCompaction` is a bit of a chonky function already
___
*[`options.go` line 676 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NehroSb0wQuxbMldtOf:-NehroSb0wQuxbMldtOg:bqtku9u) ([raw file](https://github.com/cockroachdb/pebble/blob/d400232a35f2b3b314d4f8bc853e8e4caaf631e1/options.go#L676)):*
> ```Go
> 		// (using CreateOnSharedLocator and with the appropriate
> 		// CreateOnSharedStrategy). These sstables can be shared between different
> 		// Pebble instances; the lifecycle of such objects is managed by the cluster.
> ```

what does "the lifecycle of such objects is managed by the cluster" refer to? In the context of the `Options` file, should we say "is managed by the `remote.Storage` constructed by `Options.RemoteStorage`" or similar?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2903)* status: 3 of 18 files reviewed, 5 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`objstorage/remote/storage.go` line 33 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2903#-Nei6yWNC5qM3CNUl4yx:-Nei6yWNC5qM3CNUl4yy:b-bk35we) ([raw file](https://github.com/cockroachdb/pebble/blob/d400232a35f2b3b314d4f8bc853e8e4caaf631e1/objstorage/remote/storage.go#L33)):*
> ```Go
> 	CreateOnSharedNone CreateOnSharedStrategy = iota
> 	// CreateOnSharedLower denotes the creation of files in lower levels of the
> 	// LSM (specifically, L5 and L6 as they're below sharedLevelsStart) on
> ```

sharedLevelsStart reference not appropriate here.. Maybe we need to move `sharedLevelsStart` definition here?

Though in general, this package doesn't seem the right place for this higher-level decision logic.. Maybe it belongs in `/options.go` or `internal/base/options.go`?

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2903#-:-Nexhen725zHmsNuWwbA:b-814g8k)* status: 1 of 21 files reviewed, 5 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`compaction.go` line 2777 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2903#-Nehpr5sCzTPOLm8dMhm:-NexbUYfDmgWdQGcYqcU:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/d400232a35f2b3b314d4f8bc853e8e4caaf631e1/compaction.go#L2777)):*

Today we'll only produce a move compaction with a single start-level file, but that might not always be the case. Can we either perform this for every file in `iter`, or assert that `iter.Next()` exhausts the iterator?
</blockquote></details>

Done.
___
*[`compaction.go` line 2789 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NehqDiwEvcLVpK4ZwDB:-NexbblR0SFH2Oh1F5gt:bbnvh0j) ([raw file](https://github.com/cockroachdb/pebble/blob/d400232a35f2b3b314d4f8bc853e8e4caaf631e1/compaction.go#L2789)):*

Can we move this determination into compaction picking? It's bit wonky to create a compaction marked as a move compaction, only not to perform a move. Then we can also remove the `isTrivialMove` field.
</blockquote></details>

Done. We kinda still need the `isTrivialMove` field to be able to distinguish between the two kinds of move compactions in `ClearCompactingState`, but `shouldDoTrivialMove` is now gone.
___
*[`compaction.go` line 2812 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NehrKcP7gRAyv1bPfJr:-NexdclT6aCnheAHJaOi:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/d400232a35f2b3b314d4f8bc853e8e4caaf631e1/compaction.go#L2812)):*

Should we pull this `if` block out into a helper? `runCompaction` is a bit of a chonky function already
</blockquote></details>

Done.
___
*[`options.go` line 676 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NehroSb0wQuxbMldtOf:-NexdeqcCHtVHLTOXgYL:b-foaplj) ([raw file](https://github.com/cockroachdb/pebble/blob/d400232a35f2b3b314d4f8bc853e8e4caaf631e1/options.go#L676)):*

what does "the lifecycle of such objects is managed by the cluster" refer to? In the context of the `Options` file, should we say "is managed by the `remote.Storage` constructed by `Options.RemoteStorage`" or similar?
</blockquote></details>

Oops, I meant to say the lifecycle of such objects is managed by the objstorage.Provider and can be coordinated with other Pebbles. But what you said is better.
___
*[`objstorage/remote/storage.go` line 33 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2903#-Nei6yWNC5qM3CNUl4yx:-Nexew5t-D6f0QXecTde:b2kniz6) ([raw file](https://github.com/cockroachdb/pebble/blob/d400232a35f2b3b314d4f8bc853e8e4caaf631e1/objstorage/remote/storage.go#L33)):*

sharedLevelsStart reference not appropriate here.. Maybe we need to move `sharedLevelsStart` definition here?

Though in general, this package doesn't seem the right place for this higher-level decision logic.. Maybe it belongs in `/options.go` or `internal/base/options.go`?
</blockquote></details>

Seeing as the provider contains `CreateOnShared` and this package seems to contain the data types used by both the provider and the outer db, I felt this is the best place for it. Also remote.Locator is right here. But I'm also good with moving it to objstorageprovider or so. I've moved `sharedLevelsStart` here though.

Options in the other package is hard to reference from the provider, and internal/* stuff isn't exported outside Cockroach (where we will be setting this). Thoughts?

RaduBerinde (commented):
It's a bit confusing - we have a non-move compaction, or a non-trivial move compaction, or a trivial move compaction. What's the material difference between the first two? That the copying happens at the byte level as opposed to the key level? I think this would be useful to document, eg above the `compactionKindMove` constant.

Given that we have related logic in multiple places (in `maybeDoMoveCompaction` and where we decide if it's trivial) - it feels like we should instead have two separate kinds of compactions (e.g. `Move` and `Copy`) and make the decision upfront?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2903#-:-NfD9Wes9Q77GMNSaKSh:b2djlev)* status: 1 of 21 files reviewed, 17 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`compaction.go` line 757 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD8-BN7Sr9YNi234TK:-NfD8-BN7Sr9YNi234TL:bj6j549) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L757)):*
> ```Go
> }
> 
> func (c *compaction) maybeDoMoveCompaction(provider objstorage.Provider) error {
> ```

Why make this function separate from `newCompaction`? Is it because of the error? Since `Lookup` only checks the in-memory catalog, I think it would be ok to convert to a `panic(AssertionFaledf())`.. or ignore the error and let the compaction fail when it tries to use the object
___
*[`compaction.go` line 763 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD5RSr-304nHuj0vXF:-NfD5RSr-304nHuj0vXG:bhdt0yr) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L763)):*
> ```Go
> 		// to the next. We avoid such a move if there is lots of overlapping
> 		// grandparent data. Otherwise, the move could create a parent file
> 		// that will require a very expensive merge later on.'
> ```

[nit] extra `'`
___
*[`compaction.go` line 778 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD5qs2-CLL9UhUn3q1:-NfD5qs2-CLL9UhUn3q2:bri698a) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L778)):*
> ```Go
> 				return err
> 			}
> 			if !trivialMoveMeta.Virtual || objMeta.IsRemote() || !provider.ShouldCreateShared(c.outputLevel.level) {
> ```

[nit] it would be easier to correlate with the comment if we did
```
if trivialMoveMeta.Virtual && objMeta.IsRemote() && provider.ShouldCreateShared() {
  return nil
```

Then we don't need the `else` either and we can move the comment about the provider above the `if provider != nil`
___
*[`compaction.go` line 2780 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfDA5sv86Q_aC1sp7GW:-NfDA5sv86Q_aC1sp7GX:b5q7cx8) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L2780)):*
> ```Go
> 	versionEdit *versionEdit,
> ) (ve *versionEdit, pendingOutputs []physicalMeta, retErr error) {
> 	// NB: we expect the input ve to have already been set like a trivial move
> ```

This comment is confusing. I don't think we need to mention the trivial version here at all.
___
*[`compaction.go` line 2784 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD9zzkEPunpbg257U2:-NfD9zzkEPunpbg257U3:bwn137r) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L2784)):*
> ```Go
> 	ve = versionEdit
> 	// Note that based on logic in the compaction picker, we're guaranteed
> 	// trivialMoveMeta.Virtual is false.
> ```

meta.Virtual
___
*[`compaction.go` line 2786 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD9wzN8opDAbzNXsZw:-NfD9wzN8opDAbzNXsZx:bmklyzv) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L2786)):*
> ```Go
> 	// trivialMoveMeta.Virtual is false.
> 	if meta.Virtual {
> 		panic("cannot do a trivial move compaction of a virtual sstable across local/remote storage")
> ```

add `AssertionFailedf` and remove "trivial"
___
*[`compaction.go` line 2793 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfDAGQS63JALDhSO70Z:-NfDAGQT1WUi5sQh6J0H:bnc7opb) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L2793)):*
> ```Go
> 	//
> 	// To ease up cleanup of the local file and tracking of refs, we create
> 	// a new FileNum. This has the potential of causing greater block cache
> ```

greater utilization sounds like a good thing, maybe "the potential of making the block cache less effective"
___
*[`compaction.go` line 2879 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfDB3DC-Fw5YS8ejKEL:-NfDB3DC-Fw5YS8ejKEM:b-pr5tjp) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L2879)):*
> ```Go
> 	if c.kind == compactionKindMove {
> 		iter := c.startLevel.files.Iter()
> 		trivialMoveMeta := iter.First()
> ```

This should be `moveMeta` since it may or may not be trivial at this point
___
*[`objstorage/objstorageprovider/provider.go` line 268 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD6_zU-kw6uwGhf_Nv:-NfD6_zU-kw6uwGhf_Nw:b-mbgf1r) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/objstorage/objstorageprovider/provider.go#L268)):*
> ```Go
> }
> 
> // ShouldCreateShared returns whether new table files at the specified level
> ```

I'm not sure this belongs in the provider. It's about a higher level policy of what flags to pass to the provider. We're not passing the level in `Create` and letting the provider figure it out, the caller decides whether it's shared or not.
___
*[`objstorage/remote/storage.go` line 33 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2903#-Nei6yWNC5qM3CNUl4yx:-NfD51KdCg6UrPhHeyFE:bb73djm) ([raw file](https://github.com/cockroachdb/pebble/blob/d400232a35f2b3b314d4f8bc853e8e4caaf631e1/objstorage/remote/storage.go#L33)):*

Seeing as the provider contains `CreateOnShared` and this package seems to contain the data types used by both the provider and the outer db, I felt this is the best place for it. Also remote.Locator is right here. But I'm also good with moving it to objstorageprovider or so. I've moved `sharedLevelsStart` here though.

Options in the other package is hard to reference from the provider, and internal/* stuff isn't exported outside Cockroach (where we will be setting this). Thoughts?
</blockquote></details>

SGTM
___
*[`objstorage/remote/storage.go` line 25 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD5DKeAFO1UCF8Q1-q:-NfD5DKeAFO1UCF8Q1-r:bynvmp8) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/objstorage/remote/storage.go#L25)):*
> ```Go
> }
> 
> const (
> ```

[super nit] no need for parens

itsbilal (commented):
If copying happens at a key level, it's not a move compaction at all. it's a regular/default compaction that just happens to have one input file. The only difference between trivial move and non-trivial move is that trivial move doesn't copy anything (it just links the same  FileMetadata in a different level), retaining the same FileNum. A non-trivial move is actually a Copy which does a byte-level copy and produces a new FileNum.

I've now updated this with a Copy compaction instead.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2903#-:-NfSJJrgCB-rLMiORrVE:b-ptub38)* status: 1 of 21 files reviewed, 7 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`compaction.go` line 757 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfCn9OT75n6_aGXTqSG:-NfSLcF96u7zm7KA1Bop:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L757)):*

could we tweak the name to make it clear that this function doesn't actually _perform_ the move compaction, it just alters the picked compaction?
</blockquote></details>

Done.
___
*[`compaction.go` line 757 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD8-BN7Sr9YNi234TK:-NfSLwuSD-P8NKgQ-aXc:b-arp1l6) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L757)):*

Why make this function separate from `newCompaction`? Is it because of the error? Since `Lookup` only checks the in-memory catalog, I think it would be ok to convert to a `panic(AssertionFaledf())`.. or ignore the error and let the compaction fail when it tries to use the object
</blockquote></details>

Sure. Rolled it into `newCompaction`.
___
*[`compaction.go` line 763 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD5RSr-304nHuj0vXF:-NfSR_Kq0ceP-EWG5qry:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L763)):*

[nit] extra `'`
</blockquote></details>

Done.
___
*[`compaction.go` line 778 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD5qs2-CLL9UhUn3q1:-NfSRpkg3FILCDf7Pehb:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L778)):*

[nit] it would be easier to correlate with the comment if we did
```
if trivialMoveMeta.Virtual && objMeta.IsRemote() && provider.ShouldCreateShared() {
  return nil
```

Then we don't need the `else` either and we can move the comment about the provider above the `if provider != nil`
</blockquote></details>

Done.
___
*[`compaction.go` line 1732 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD-AOT1V2iWsEug1GE:-NfSSC3hAcYI5L46zaPE:b3y3fll) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L1732)):*

a thought: what do you think about passing in the `*versionEdit` and checking `ve.NewFiles` for the file to determine whether or not we should return `f` to `StateNotCompacting`?  At such low counts, I don't think the O(n^2) loop is anything to sweat, and it'd remove the `isTrivialMove` field and automatically work should we implement something like #389.
</blockquote></details>

Good idea. Either way this became irrelevant because we now have a different compactionKind for copy compactions.
___
*[`compaction.go` line 2780 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfDA5sv86Q_aC1sp7GW:-NfSTTGl4lzaC35op1Pr:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L2780)):*

This comment is confusing. I don't think we need to mention the trivial version here at all.
</blockquote></details>

Done.
___
*[`compaction.go` line 2784 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD9zzkEPunpbg257U2:-NfSSNh7B3CGwts-ObkI:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L2784)):*

meta.Virtual
</blockquote></details>

Done.
___
*[`compaction.go` line 2786 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD9wzN8opDAbzNXsZw:-NfSTXav7Axs2Ozz4z4s:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L2786)):*

add `AssertionFailedf` and remove "trivial"
</blockquote></details>

Done.
___
*[`compaction.go` line 2793 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfDAGQS63JALDhSO70Z:-NfSTjXp5z4jvd-q5wtw:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L2793)):*

greater utilization sounds like a good thing, maybe "the potential of making the block cache less effective"
</blockquote></details>

Done.
___
*[`compaction.go` line 2879 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfDB3DC-Fw5YS8ejKEL:-NfSTsYk9_OCh8AtKQWH:bypnipa) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/compaction.go#L2879)):*

This should be `moveMeta` since it may or may not be trivial at this point
</blockquote></details>

Done. Just made it `meta`.
___
*[`objstorage/objstorageprovider/provider.go` line 268 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD6_zU-kw6uwGhf_Nv:-NfSVNOu31TWWAjLo3vj:b-jikhek) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/objstorage/objstorageprovider/provider.go#L268)):*

I'm not sure this belongs in the provider. It's about a higher level policy of what flags to pass to the provider. We're not passing the level in `Create` and letting the provider figure it out, the caller decides whether it's shared or not.
</blockquote></details>

I moved it out of the provider. I kept it here as first as it already owns Settings, but all the callers also have access to the Options themselves and can just use them. PTAL?
___
*[`objstorage/remote/storage.go` line 25 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfD5DKeAFO1UCF8Q1-q:-NfSUPBrDuHdRAn4XoUN:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/objstorage/remote/storage.go#L25)):*

[super nit] no need for parens
</blockquote></details>

Done.
___
*[`internal/manifest/version.go` line 459 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfCoZcrDhy0vjWSsi2Y:-NfSXEOS5-s-3M1Y_AZ0:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e950c5158053ae0adbc78c959f9e8f9c5af75473/internal/manifest/version.go#L459)):*

I think we might be better off just explicitly copying what we need in `runNonTrivialMoveCompaction`. Many of the copied fields are irrelevant to the uploaded file (`FileNum`, `FileBacking`, `InitAllowed`,  `Sublevel`, `L0Index`, `minIntervalIndex`, `maxIntervalIndex`, `IsIntraL0Compacting`, `CompactionState`, `MarkedForCompaction`).

I'm wary of maintaining this broad `Copy` func, given we need to remember to manually copy every field we add to the file metadata.
</blockquote></details>

Done.

RaduBerinde (commented):
Nice, it's better with a Copy kind!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2903#-:-NfTTyxiEcm5iPjMC_oe:b-ft5khu)* status: 1 of 21 files reviewed, 10 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`compaction.go` line 471 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfTSY3Y8n2PGo6TlqeG:-NfTSY3Y8n2PGo6TlqeH:bocy84j) ([raw file](https://github.com/cockroachdb/pebble/blob/74c6c5b4de2fc529d97c299d92ef9105d7096d51/compaction.go#L471)):*
> ```Go
> 	compactionKindRewrite
> 	compactionKindIngestedFlushable
> 	compactionKindCopy
> ```

I'd add an explanation here.. And maybe move it next to `compactionKindMove`.  I realize you're following precedent but IMO this is the place where all of these should be documented to some extent.
___
*[`compaction.go` line 789 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfTS-T31yY00JUkePVh:-NfTS-T31yY00JUkePVi:brwswrl) ([raw file](https://github.com/cockroachdb/pebble/blob/74c6c5b4de2fc529d97c299d92ef9105d7096d51/compaction.go#L789)):*
> ```Go
> 			}
> 		} else {
> 			// We have no provider. This should only be the case in some unit tests.
> ```

Feels strange that we don't check the three conditions even without a provider. We could simplify a bit by doing this above:
```
isRemote := false
if provider != nil {
  objMeta, err := provider.Lookup()
  isRemote = objMeta.isRemote
}
```
___
*[`compaction.go` line 2938 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfTTMoF8zz2kZ_LiXPz:-NfTTMoF8zz2kZ_LiXQ-:b-pdy2jt) ([raw file](https://github.com/cockroachdb/pebble/blob/74c6c5b4de2fc529d97c299d92ef9105d7096d51/compaction.go#L2938)):*
> ```Go
> 	if c.kind == compactionKindMove || c.kind == compactionKindCopy {
> 		iter := c.startLevel.files.Iter()
> 		trivialMoveMeta := iter.First()
> ```

Remove "trivial".. in fact it's either a copy or move meta, so maybe just `sourceMeta`?
___
*[`compaction.go` line 2963 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfTTmas1tZsQCiOTZhK:-NfTTmas1tZsQCiOTZhL:b6jktoh) ([raw file](https://github.com/cockroachdb/pebble/blob/74c6c5b4de2fc529d97c299d92ef9105d7096d51/compaction.go#L2963)):*
> ```Go
> 		}
> 		if c.kind == compactionKindCopy {
> 			if objMeta.IsRemote() || !remote.ShouldCreateShared(d.opts.Experimental.CreateOnShared, c.outputLevel.level) {
> ```

[nit] this check belongs inside runCopyCompaction

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2903#-:-NfWRamK0FMsqSGgy0nr:b-814g8k)* status: 1 of 21 files reviewed, 6 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`compaction.go` line 471 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfTSY3Y8n2PGo6TlqeG:-NfWPGzbBqp4pjoSnxr1:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/74c6c5b4de2fc529d97c299d92ef9105d7096d51/compaction.go#L471)):*

I'd add an explanation here.. And maybe move it next to `compactionKindMove`.  I realize you're following precedent but IMO this is the place where all of these should be documented to some extent.
</blockquote></details>

Done.
___
*[`compaction.go` line 789 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfTS-T31yY00JUkePVh:-NfWQN3J3HJCzEXLj_Es:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/74c6c5b4de2fc529d97c299d92ef9105d7096d51/compaction.go#L789)):*

Feels strange that we don't check the three conditions even without a provider. We could simplify a bit by doing this above:
```
isRemote := false
if provider != nil {
  objMeta, err := provider.Lookup()
  isRemote = objMeta.isRemote
}
```
</blockquote></details>

Done.
___
*[`compaction.go` line 2938 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfTTMoF8zz2kZ_LiXPz:-NfWQTNo7dvmZxXZbfED:b-t3xy3r) ([raw file](https://github.com/cockroachdb/pebble/blob/74c6c5b4de2fc529d97c299d92ef9105d7096d51/compaction.go#L2938)):*

Remove "trivial".. in fact it's either a copy or move meta, so maybe just `sourceMeta`?
</blockquote></details>

Done. It's just `meta` now.
___
*[`compaction.go` line 2963 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfTTmas1tZsQCiOTZhK:-NfWQbOc1FfBNvpGJKdt:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/74c6c5b4de2fc529d97c299d92ef9105d7096d51/compaction.go#L2963)):*

[nit] this check belongs inside runCopyCompaction
</blockquote></details>

Done.

itsbilal:
@RaduBerinde @jbowens friendly ping

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2903#-:-Nfvy5Lu-RReSWd-Dgqm:bnfp4nl)* status: 1 of 21 files reviewed, 8 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`compaction.go` line 765 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2903#-NfvweDJ-DGDTrLPTJxr:-NfvweDJ-DGDTrLPTJxs:b-47rdec) ([raw file](https://github.com/cockroachdb/pebble/blob/ed3a6258a69ae5b3073f43df06540d9c9f65fd47/compaction.go#L765)):*
> ```Go
> 	if c.kind == compactionKindDefault && c.outputLevel.files.Empty() && !c.hasExtraLevelData() &&
> 		c.startLevel.files.Len() == 1 && c.grandparents.SizeSum() <= c.maxOverlapBytes {
> 		// This compaction can be converted into a trivial move from one level
> ```

[nit] can be converted into a copy or a move
___
*[`compaction.go` line 786 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2903#-Nfvx08f1fuaEJ7w8zYB:-Nfvx08f1fuaEJ7w8zYC:bd6uucq) ([raw file](https://github.com/cockroachdb/pebble/blob/ed3a6258a69ae5b3073f43df06540d9c9f65fd47/compaction.go#L786)):*
> ```Go
> 		// 2) The existing file `meta` is on non-remote storage
> 		// 3) The output level prefers shared storage
> 		if meta.Virtual && !isRemote && remote.ShouldCreateShared(opts.Experimental.CreateOnShared, c.outputLevel.level) {
> ```

I'd add a variable
`mustCopy := !isRemote && remote.ShouldCreateShared()`

then the logic can be 
```go
if mustCopy {
  //  We have to copy the data anyway, it's better to rewrite if the source is virtual.
  if !meta.Virtual {
    c.kind = Copy
  }
} else {
  c.kind = Move
}
```

I think this makes the logic more clear and the code self-documenting. I also think it's a good idea to avoid the `return` here as it might be missed if we add more (unrelated) code below.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2903#-:-Nfw4ueHDs072ZrwZ9hi:b-814g8k)* status: 1 of 21 files reviewed, 7 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`compaction.go` line 786 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2903#-Nfvx08f1fuaEJ7w8zYB:-Nfw4sseBXAA-XoQHiAA:b-fznogt) ([raw file](https://github.com/cockroachdb/pebble/blob/ed3a6258a69ae5b3073f43df06540d9c9f65fd47/compaction.go#L786)):*

I'd add a variable
`mustCopy := !isRemote && remote.ShouldCreateShared()`

then the logic can be 
```go
if mustCopy {
  //  We have to copy the data anyway, it's better to rewrite if the source is virtual.
  if !meta.Virtual {
    c.kind = Copy
  }
} else {
  c.kind = Move
}
```

I think this makes the logic more clear and the code self-documenting. I also think it's a good idea to avoid the `return` here as it might be missed if we add more (unrelated) code below.
</blockquote></details>

Good idea, done.

PR #2904, opened 2023-09-14
vfs: enforce mutual exclusion in MemFS.Lock

jbowens:
Previously MemFS's Lock method did not implement mutual exclusion under the assumption that databases are given separate processes which do not share memory. This commit adapts Lock to enforce mutual exclusion, preventing accidentally using same FS concurrently from more than one Pebble instance. It's believed that cockroachdb/cockroach#110645 is the result of Cockroach acceptance tests opening a MemFS store twice, with one Pebble instance deleting files from beneath another Pebble instance.

### DIFF
--- a/vfs/mem_fs.go
+++ b/vfs/mem_fs.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/cockroachdb/errors"
@@ -78,6 +79,11 @@ type MemFS struct {
 	mu   sync.Mutex
 	root *memNode
 
+	// lockFiles holds a map of open file locks. Presence in this map indicates
+	// a file lock is currently held. Keys are strings holding the path of the
+	// locked file. The stored value is untyped and  unused; only presence of
+	// the key within the map is significant.
+	lockedFiles sync.Map
 	strict      bool
 	ignoreSyncs bool
 	// Windows has peculiar semantics with respect to hard links and deleting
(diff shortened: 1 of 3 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2904#-:-NeKwE1kEoEPrIFs8gOT:bnfp4nl)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`vfs/mem_fs.go` line 468 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2904#-NeKvZgl6ic-kKtnTVNO:-NeKvZgmBVORDupQvFYz:b3h0sz0) ([raw file](https://github.com/cockroachdb/pebble/blob/b4ccd7020ad564476d6d6f09474ad7c34f35e000/vfs/mem_fs.go#L468)):*
> ```Go
> 	// close the same MemFS-backed database multiple times. We want mutual
> 	// exclusion in this case too. See cockroachdb/cockroach#110645.
> 	_, loaded := y.lockedFiles.Swap(fullname, true)
> ```

[nit] the value could be `nil` here (I initially wondered if the true value has some significance)
___
*[`vfs/mem_fs.go` line 474 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2904#-NeKvq58-KD3scnnAwX9:-NeKvq58-KD3scnnAwXA:bu54spl) ([raw file](https://github.com/cockroachdb/pebble/blob/b4ccd7020ad564476d6d6f09474ad7c34f35e000/vfs/mem_fs.go#L474)):*
> ```Go
> 		return nil, syscall.EAGAIN
> 	}
> 	// Otherwise, we successfully acquired the lock. Locks are visible in the
> ```

"locks are visible in the file" - did you mean "filesystem"?

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2904#-:-NeOOvhX4JhUoiq5YQJV:b-814g8k)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @itsbilal)

___
*[`vfs/mem_fs.go` line 468 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2904#-NeKvZgl6ic-kKtnTVNO:-NeOOIIB6wmetm5xUUSh:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b4ccd7020ad564476d6d6f09474ad7c34f35e000/vfs/mem_fs.go#L468)):*

[nit] the value could be `nil` here (I initially wondered if the true value has some significance)
</blockquote></details>

Done.
___
*[`vfs/mem_fs.go` line 474 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2904#-NeKvq58-KD3scnnAwX9:-NeOOeok7A9C6wzsDGgx:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b4ccd7020ad564476d6d6f09474ad7c34f35e000/vfs/mem_fs.go#L474)):*

"locks are visible in the file" - did you mean "filesystem"?
</blockquote></details>

Done.

Issue #2885, opened 2023-09-06
metamorphic: integrate IngestWithExcise


jbowens:
The excise operation is currently untested by the metamorphic tests. Integration could've caught issues like the one fixed in #2884.

PR #2910, opened 2023-09-16
db: support virtual sstables in block checksum validation

bananabrick:
Supports validation of virtual sstable block checksums during ingestion.
It was easier to deduplicate the virtual sstables by file backing to
ensure that the checksums for each file are only validated once. We
could also implement a VirtualReader.ValidateVirtual function which only
validates blocks which fall within virtual sstable bounds, but that
seems like too much work for a feature which is off by default and was
never turned on.

This pr also prepares us for
https://github.com/cockroachdb/pebble/issues/2885. The metamorphic test
might set `ValidateOnIngest` to true in which case we'll hit a panic.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -2004,26 +2004,29 @@ func (d *DB) ingestApply(
 	// The ingestion may have pushed a level over the threshold for compaction,
 	// so check to see if one is necessary and schedule it.
 	d.maybeScheduleCompaction()
-	d.maybeValidateSSTablesLocked(ve.NewFiles)
+	var toValidate []manifest.NewFileEntry
+	dedup := make(map[base.DiskFileNum]struct{})
+	for _, entry := range ve.NewFiles {
+		if _, ok := dedup[entry.Meta.FileBacking.DiskFileNum]; !ok {
+			toValidate = append(toValidate, entry)
+			dedup[entry.Meta.FileBacking.DiskFileNum] = struct{}{}
+		}
+	}
+	d.maybeValidateSSTablesLocked(toValidate)
 	return ve, nil
 }
 
 // maybeValidateSSTablesLocked adds the slice of newFileEntrys to the pending
 // queue of files to be validated, when the feature is enabled.
-// DB.mu must be locked when calling.
 //
-// TODO(bananabrick): Make sure that the ingestion step only passes in the
-// physical sstables for validation here.
+// Note that if two entries with the same backing file are added twice, then the
+// block checksums for the backing file will be validated twice.
+//
+// DB.mu must be locked when calling.
 func (d *DB) maybeValidateSSTablesLocked(newFiles []newFileEntry) {
 	// Only add to the validation queue when the feature is enabled.
 	if !d.opts.Experimental.ValidateOnIngest {
 		return
-	}
-
-	for _, f := range newFiles {
-		if f.Meta.Virtual {
-			panic("pebble: invalid call to maybeValidateSSTablesLocked")
-		}
 	}
 
 	d.mu.tableValidation.pending = append(d.mu.tableValidation.pending, newFiles...)
@@ -2085,10 +2088,19 @@ func (d *DB) validateSSTables() {
 			}
 		}
 
-		err := d.tableCache.withReader(
-			f.Meta.PhysicalMeta(), func(r *sstable.Reader) error {
-				return r.ValidateBlockChecksums()
-			})
+		var err error
+		if f.Meta.Virtual {
+			err = d.tableCache.withVirtualReader(
+				f.Meta.VirtualMeta(), func(v sstable.VirtualReader) error {
+					return v.ValidateBlockChecksumsOnBacking()
+				})
+		} else {
+			err = d.tableCache.withReader(
+				f.Meta.PhysicalMeta(), func(r *sstable.Reader) error {
+					return r.ValidateBlockChecksums()
+				})
+		}
+
 		if err != nil {
 			// TODO(travers): Hook into the corruption reporting pipeline, once
 			// available. See pebble#1192.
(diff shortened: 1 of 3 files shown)


bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2910)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`ingest.go` line 2008 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2910#-NeUiTdkAos6X_4QTSzX:-NeUiTdkAos6X_4QTSzY:bmmcosv) ([raw file](https://github.com/cockroachdb/pebble/blob/3e851599bd16ceb42518f8dbf61a86a273f723dc/ingest.go#L2008)):*
> The virtual sstables created during ingestion are created from backing files which are present in the lsm

@itsbilal Does that sound right?

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2910)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`ingest.go` line 2008 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2910#-NeUiTdkAos6X_4QTSzX:-NeUk1JP9v5ol4wlEGEm:b-cdae3g) ([raw file](https://github.com/cockroachdb/pebble/blob/3e851599bd16ceb42518f8dbf61a86a273f723dc/ingest.go#L2008)):*

> The virtual sstables created during ingestion are created from backing files which are present in the lsm

@itsbilal Does that sound right?
</blockquote></details>

Scratch that. It's not right.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2910)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`ingest.go` line 2025 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2910#-NeUsjSH-oPHDoFj0eyQ:-NeUsjSH-oPHDoFj0eyR:bc484bo) ([raw file](https://github.com/cockroachdb/pebble/blob/6d6570bf1e25d0776f565498d2f43ca4dcf5cf4b/ingest.go#L2025)):*
> ```Go
> 	for _, f := range newFiles {
> 		if f.Meta.Virtual {
> 			panic("pebble: invalid call to maybeValidateSSTablesLocked")
> ```

We weren't hitting this panic because the `ValidateOnIngest` option is off by default, and only turned on in the metamorphic test. But the metamorphic test doesn't currently support excises.

bananabrick:
This will get tested by https://github.com/cockroachdb/pebble/issues/2885, but I could also add in some sort of a test here if that's necessary.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2910#-:-Net-UcN2D_QYAPZh3O1:bnfp4nl)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @bananabrick)

___
*[`ingest.go` line 2015 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2910#-Nesz44UBnt7Emd2BnFx:-Nesz44UBnt7Emd2BnFy:bm2d7fb) ([raw file](https://github.com/cockroachdb/pebble/blob/81d2673fd10f8e69ba4f94767dc9693c33665cc3/ingest.go#L2015)):*
> ```Go
> 		}
> 	}
> 	d.maybeValidateSSTablesLocked(toValidate)
> ```

The change looks fine, but it feels like there is a lower-level construct missing. We pass the Metadata to `findNode`, so we will only pass the metadata of the first virtual "user" of a backing. That's fine in practice because we are not using it for anything serious. But conceptually, we would really want to pass (to maybeValidateSSTablesLocked) just the backing file num and what's strictly necessary to validate the physical sst.
___
*[`sstable/reader_virtual.go` line 94 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2910#-Net-KRa5ywYmBuE7sb1:-Net-KRb83MIwnn2XeQZ:b2yjxxa) ([raw file](https://github.com/cockroachdb/pebble/blob/81d2673fd10f8e69ba4f94767dc9693c33665cc3/sstable/reader_virtual.go#L94)):*
> ```Go
> // ValidateBlockChecksums will call ValidateBlockChecksums on the underlying reader.
> // Note that block checksum validation is NOT restricted to virtual sstable bounds.
> func (v *VirtualReader) ValidateBlockChecksums() error {
> ```

[nit] maybe ValidateBlockChecksumsOnBacking or something similar to make it clear it's not really an operation on the virtual sst

bananabrick (commented):
Flipped the ValidateOnIngest bool to true in the excising tests, so this should now be getting tested.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2910#-:-NeuwVQ26n_dcIa2GM32:bxyl7o)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`ingest.go` line 2015 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2910#-Nesz44UBnt7Emd2BnFx:-NeuuBW0-uF4VRVgaFtq:b-earziq) ([raw file](https://github.com/cockroachdb/pebble/blob/81d2673fd10f8e69ba4f94767dc9693c33665cc3/ingest.go#L2015)):*

The change looks fine, but it feels like there is a lower-level construct missing. We pass the Metadata to `findNode`, so we will only pass the metadata of the first virtual "user" of a backing. That's fine in practice because we are not using it for anything serious. But conceptually, we would really want to pass (to maybeValidateSSTablesLocked) just the backing file num and what's strictly necessary to validate the physical sst.
</blockquote></details>

I agree. I wanted to pass in the FileBacking, but the table cache currently needs the FileMetadata to create a Reader.

It only needs the Smallest/Largest SeqNums in addition to what's present in the FileBacking(see the findNode function in the table cache), so should be possible to use a FileBacking if we really want to.

Issue #2323, opened 2023-02-13
db: test and implement compaction picking heuristics for virtual sstables


bananabrick:
Some of the current compaction picking heuristics don't necessarily make sense for virtual sstables. We also need to implement a compaction picking heuristic to track space amp due to virtual sstables.

PR #2912, opened 2023-09-18
db: wire virtual sstable space amp heuristic

bananabrick:
We compute the space used by the backing file, the virtual files for which have already been compacted.

Then, we divide this by the number of virtual sstables which have this backing file to get a given virtual sstable's contribution to the space amp, `VirtualSpaceAmpProportion`.

Then in pickCompactionSeedFile, we divide the overlapping ratio by compensated size + `VirtualSpaceAmpProportion`.

The larger the virtual space amp proportion the more likely it is that the file gets picked for compaction.

Fixes: https://github.com/cockroachdb/pebble/issues/2323

### DIFF
--- a/internal/manifest/version.go
+++ b/internal/manifest/version.go
@@ -365,9 +365,6 @@ type FileBacking struct {
 	// The intuition is that if FileBacking.Size - FileBacking.VirtualizedSize
 	// is high, then the space amplification due to virtual sstables is
 	// high, and we should pick the virtual sstable with a higher priority.
-	//
-	// TODO(bananabrick): Compensate the virtual sstable file size using
-	// the VirtualizedSize during compaction picking and test.
 	VirtualizedSize atomic.Uint64
 	DiskFileNum     base.DiskFileNum
 	Size            uint64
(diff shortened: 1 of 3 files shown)


bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2912)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[`compaction_picker.go` line 1093 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2912#-Ne_mOdI4IDSaYPVbH_-:-Ne_mOdI4IDSaYPVbH_0:b-v3xx22) ([raw file](https://github.com/cockroachdb/pebble/blob/c529f10a88354b28512ab01fba4a41781b29af06/compaction_picker.go#L1093)):*
> ```Go
> 	// the target level, which minimizes write amplification.
> 	//
> 	// It uses a "compensated size" + "virtual space amp" for the denominator.
> ```

Looking for feedback on whether this approach seems sound.

internal/manifest/version.go
```diff
@@ -453,6 +450,18 @@ func (m *FileMetadata) LatestRef() {
...
+// VirtualSpaceAmpProportion computes a virtual sstable's contribution to space
+// amplification. See comment above FileBacking.VirtualizedSize for the
+// reasoning behind this computation.
+func (m *FileMetadata) VirtualSpaceAmpProportion() uint64 {
+	if !m.Virtual {
+		return 0
+	}
+	virtualizedSize := m.FileBacking.VirtualizedSize.Load()
+	latestRefs := m.LatestRefs()
+	return (m.FileBacking.Size - virtualizedSize) / uint64(latestRefs)
```
itsbilal, line 462: I wonder if we should be dividing by `m.Size` instead of `latestRefs` here - that way it's easier to reason about smaller virtual sstables getting a greater size adjustment than bigger virtual sstables that reference the same underlying sstable. Thoughts?

itsbilal (commented):
General approach LGTM, just some suggestions

PR #2915, opened 2023-09-18
open: Ratchet nextFileNum ahead of objprovider objects

itsbilal:
Previously, we were ratcheting nextFileNum ahead of all locally present obsolete sstables, but not doing the same for obsolete sstables in shared storage. This would cause errors due to FileNum clashes on restart after a mid-compaction process abort where some new sstables were created and persisted in the provider.

This change addresses that by ratcheting nextFileNum during Open() ahead of all known objects in the provider too.

### DIFF
--- a/open.go
+++ b/open.go
@@ -381,6 +381,15 @@ func Open(dirname string, opts *Options) (db *DB, _ error) {
 		}
 	}
 
+	// Ratchet d.mu.versions.nextFileNum ahead of all known objects in the
+	// objProvider. This avoids FileNum collisions with obsolete sstables.
+	objects := d.objProvider.List()
+	for _, obj := range objects {
+		if d.mu.versions.nextFileNum <= obj.DiskFileNum.FileNum() {
+			d.mu.versions.nextFileNum = obj.DiskFileNum.FileNum() + 1
+		}
+	}
+
 	// Validate the most-recent OPTIONS file, if there is one.
 	var strictWALTail bool
 	if previousOptionsFilename != "" {
--- a/open_test.go
+++ b/open_test.go
@@ -6,6 +6,7 @@ package pebble
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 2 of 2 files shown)


RaduBerinde (approved):
:lgtm:

Thanks for the fix! Wondering if the metamorphic test should have caught this?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2915#-:-NeeH77jALsTypspGq6n:b-ee48ny)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`open.go` line 389 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2915#-NeeJFlnCYXHnL5w8h6m:-NeeJFlnCYXHnL5w8h6n:b-z8mz35) ([raw file](https://github.com/cockroachdb/pebble/blob/ab164edec5fe3823056b7718c5885934bba4e110/open.go#L389)):*
> ```Go
> 	for _, obj := range objects {
> 		if d.mu.versions.nextFileNum <= obj.DiskFileNum.FileNum() {
> 			d.mu.versions.nextFileNum = obj.DiskFileNum.FileNum() + 1
> ```

[nit] Would be nice if we could add a test for this.. Maybe open a database, then create a provider directly and create some objects, then re-open? We'd need to make it do a compaction during WAL replay though..

itsbilal (commented):
Thanks! Metamorphic test can catch this if it just has CreateOnShared = true and a RemoteStorage factory set. Currently those things aren't true, but they can be more easily than if we do shared ingestions.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2915#-:-NefPDcrE7qAB-8UcDwW:b-i1sx6j)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @RaduBerinde)

___
*[`open.go` line 389 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2915#-NeeJFlnCYXHnL5w8h6m:-NefPN4YBWy3YnK4Vq0f:b-f9tkux) ([raw file](https://github.com/cockroachdb/pebble/blob/ab164edec5fe3823056b7718c5885934bba4e110/open.go#L389)):*

[nit] Would be nice if we could add a test for this.. Maybe open a database, then create a provider directly and create some objects, then re-open? We'd need to make it do a compaction during WAL replay though..
</blockquote></details>

Done - added a test which repros this successfully. The flush/compaction on start is easy by just having some keys in the memtable at db close time, as that means we'll replay the WAL and flush the non-empty memtable which is sufficient to trigger this bug.

Issue #2832, opened 2023-08-16
pick a heuristic to avoid the effects of large level size compensation


bananabrick:
Link to the compaction scoring internal [doc](https://docs.google.com/document/d/1yqh2j3l6kjDVIkI5qxTH8xVTx9e2_pcFFJ7QW_JDgi0/edit?usp=sharing).

The setup for the experiments was to run `clearrange/checks=true`.

- [x] post experimental results with prs for the changes.
- [x] post a blurb about the results for each run, comparing it to the other runs.

bananabrick:
Heuristic 1: CapCompensated from the doc

bananabrick:
Heuristic 2: NoCompensateL0

bananabrick:
Heuristic 3: CompensatedOnlyAsThreshold

bananabrick:
Heuristic 4: CompensatedForPreferringHigherLevels

bananabrick:
Master

Grafana [run](https://grafana.testeng.crdb.io/d/ngnKs_j7x/admission-control?orgId=1&var-cluster=teamcity-11324151-1692078373-72-n10cpu16&var-instances=All&from=1692108746113&to=1692114964036).

jbowens:
> Heuristic 1: CapCompensated

One possible variant of this: Compute the raw score _x_. Compute (_total compensation_ / _level target score_ = _C_). Today, the compensated score is _x_ + _C_. Instead, compute the compensated score as _x_ + _ln_(_e_+_C_) - 1. This would make the compensated scores grow logarithmically with the amount of data they drop, and makes it continuous so that we sidestep any concerns about equal scores introduced by a hard cap.

bananabrick:
Internal link for logs: https://drive.google.com/file/d/1kvEuCyz7DlZij9QrH-CFdqHvbI19S6E4/view?usp=drive_link

The runs were for heuristics 1, 3, 4. The cluster name is present in the logs and can be used to view the grafana dashboard for the run.

TODO: Dump summaries about the logs.

bananabrick:
cap_compensation is the same as https://github.com/cockroachdb/pebble/issues/2832#issuecomment-1681231538.

raw_scores is https://github.com/cockroachdb/pebble/issues/2832#issuecomment-1681232618.

compensate_higher is https://github.com/cockroachdb/pebble/issues/2832#issuecomment-1681232935.

[...]

PR #2917, opened 2023-09-18
db: use uncompensated scores to prioritize levels for compaction

bananabrick:
During large table drops, we see level compensation become drastically high, which leads to L0 compaction starvation, which in turn leads to AC queueing of requests and increases in foreground latencies.

We completely ignore compensation during prioritization of levels during compaction, but we consider compensation when we try and pick a level for compaction.

The above approach ensures that L0 isn't starved during large table drops.

Note that there might be scenarios where a level say L3, has a high compensated score due to a wide range delete, and L3 will also drop a ton of data in L4, but we end up picking L4 for compaction because it has a higher raw score.

While this will make some of the compactions picked inefficient, most of the data which is dropped due to the range delete should be in L6 anyway, and the L5 -> L6 compaction which will contain this range delete will still be cheap.

We considered some other heuristics in
https://github.com/cockroachdb/pebble/issues/2832.

Other approaches considered include capping the level compensation to a constant, or accounting for compensation in higher levels, but not lower levels. The first approach might not work well, because the level compensation can be lower than the constant picked and still starve out L0. The second approach is more difficult to reason about compared to the current approach.

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -90,18 +90,38 @@ func (info compactionInfo) String() string {
 	return buf.String()
 }
 
-type sortCompactionLevelsDecreasingScore []candidateLevelInfo
+type sortCompactionLevelsByPriority []candidateLevelInfo
 
-func (s sortCompactionLevelsDecreasingScore) Len() int {
+func (s sortCompactionLevelsByPriority) Len() int {
 	return len(s)
 }
-func (s sortCompactionLevelsDecreasingScore) Less(i, j int) bool {
-	if s[i].score != s[j].score {
-		return s[i].score > s[j].score
+
+// A level should be picked for compaction if the compensatedScoreRatio is >= the
+// compactionScoreThreshold.
+const compactionScoreThreshold = 1
+
+// Less should return true if s[i] must be placed earlier than s[j] in the final
+// sorted list. The candidateLevelInfo for the level placed earlier is more likely
+// to be picked for a compaction.
+func (s sortCompactionLevelsByPriority) Less(i, j int) bool {
+	iShouldCompact := s[i].compensatedScoreRatio >= compactionScoreThreshold
+	jShouldCompact := s[j].compensatedScoreRatio >= compactionScoreThreshold
+	// Ordering is defined as decreasing on (shouldCompact, uncompensatedScoreRatio)
+	// where shouldCompact is 1 for true and 0 for false.
+	if iShouldCompact && !jShouldCompact {
+		return true
+	}
+	if !iShouldCompact && jShouldCompact {
+		return false
+	}
+
+	if s[i].uncompensatedScoreRatio != s[j].uncompensatedScoreRatio {
+		return s[i].uncompensatedScoreRatio > s[j].uncompensatedScoreRatio
 	}
 	return s[i].level < s[j].level
 }
-func (s sortCompactionLevelsDecreasingScore) Swap(i, j int) {
+
+func (s sortCompactionLevelsByPriority) Swap(i, j int) {
 	s[i], s[j] = s[j], s[i]
 }
 
@@ -153,6 +173,7 @@ func generateSublevelInfo(cmp base.Compare, levelFiles manifest.LevelSlice) []su
 
 // compactionPickerMetrics holds metrics related to the compaction picking process
 type compactionPickerMetrics struct {
+	// scores contains the compensatedScoreRatio from the candidateLevelInfo.
 	scores                      []float64
 	singleLevelOverlappingRatio float64
 	multiLevelOverlappingRatio  float64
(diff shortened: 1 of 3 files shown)


bananabrick:
I still like https://github.com/cockroachdb/pebble/issues/2832#issuecomment-1681232935 too, and it would allow us to perform cheaper compactions out of a level if the level has a high compensated score, but for now I'm sticking to the simpler heuristic.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2917)* status: 0 of 3 files reviewed, 6 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`compaction_picker.go` line 683 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2917#-Nemn0ov89rk15azlQ1p:-NetBCAp107SdLGS9IXV:b3pskyg) ([raw file](https://github.com/cockroachdb/pebble/blob/c2293f62fb0ce070503b2b8f342e439fa1107c8f/compaction_picker.go#L683)):*

We have 4 score related terms now, score, origScore, rawScore, rawSmoothed, and it is hard for me to remember what they mean since they are not descriptive. Also "smoothed" evokes ideas like exponential smoothing or other time based smoothing which is not what this is.

How about `rawScore`, `compensatedSizeScore` (instead of `origScore`), `rawScoreRatio` (instead of `rawSmoothed`), `compensatedSizeScoreRatio` (instead of `score`)?
</blockquote></details>

Done, mostly.

Went with slightly different names. Slightly more verbose, but more precise.
___
*[`compaction_picker.go` line 991 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2917#-NemodP_ARpUJxixAMOc:-Nf03ojrDX3MBMMs_euj:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c2293f62fb0ce070503b2b8f342e439fa1107c8f/compaction_picker.go#L991)):*

The code structuring here is a bit error prone, I think. It was fine when we had one score.
We should only assign `rawScore` and `compensatedSizeScore` here. Similarly in the for loop below, and not have commentary like `// will be smoothed later.`. Then where we compute the score ratios we should assign to both the ratio fields -- otherwise it is hard to see what has been (partially) initialized and what is left to do.
</blockquote></details>

Done.
___
*[`compaction_picker.go` line 1035 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2917#-NemxWHsDXee_UUE0qvG:-NeswyNl0GvPrkKS78lN:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c2293f62fb0ce070503b2b8f342e439fa1107c8f/compaction_picker.go#L1035)):*

let's elevate this 1 to a const as mentioned in an earlier comment.
</blockquote></details>

Done.
___
*[`compaction_picker.go` line 1042 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2917#-NemxaUT0GYvXm7VWvxQ:-NeswyzZ-oiePuCMg0pT:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c2293f62fb0ce070503b2b8f342e439fa1107c8f/compaction_picker.go#L1042)):*

ditto
</blockquote></details>

Done.
___
*[`compaction_picker.go` line 1297 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2917#-NemxcvoDrzLd7dq9wgZ:-Nesx-5b3PK9QpXjM4P8:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c2293f62fb0ce070503b2b8f342e439fa1107c8f/compaction_picker.go#L1297)):*

ditto
</blockquote></details>

Done.

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 2 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2917#-:-NfCX5bn92q0YEt8xtxJ:bg7igkv)* status: all files reviewed, 7 unresolved discussions (waiting on @bananabrick and @itsbilal)

___
*[`compaction_picker.go` line 1077 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2917#-NfCWJym8YVzBqEDSFjL:-NfCWJym8YVzBqEDSFjM:b6yzete) ([raw file](https://github.com/cockroachdb/pebble/blob/4437f771289a2ad769f59b369bb4e923488d064c/compaction_picker.go#L1077)):*
> ```Go
> 		prevLevel = level
> 	}
> 	scores[prevLevel].compensatedScoreRatio = scores[prevLevel].compensatedScore
> ```

Can you add a comment
// Set the score ratios for the lowest level.
// INVARIANT: prevLevel == numLevels-1

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2917)* status: 2 of 3 files reviewed, 7 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`compaction_picker.go` line 1077 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2917#-NfCWJym8YVzBqEDSFjL:-NfCjTGp3TuFQUqw67wp:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/4437f771289a2ad769f59b369bb4e923488d064c/compaction_picker.go#L1077)):*

Can you add a comment
// Set the score ratios for the lowest level.
// INVARIANT: prevLevel == numLevels-1
</blockquote></details>

Done.

PR #2922, opened 2023-09-20
ingest: Set bounds with zero sequence number internal keys

itsbilal:
Previously we were letting the inbound SharedMeta dictate what sequence numbers would go into shared file bounds, before resetting sequence numbers on those file bounds in a later step. This led to a rare case where we'd end up setting the overall file bounds based on a point key when we should have done it based on a range key (due to the Kind being greater).

This change addresses it by using zero sequence numbers when calling Extend{Point,Range}KeyBounds on shared metas.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -113,11 +113,28 @@ func ingestSynthesizeShared(
 	meta.FileBacking.Size = sm.Size
 	if sm.LargestRangeKey.Valid() && sm.LargestRangeKey.UserKey != nil {
 		// Initialize meta.{HasRangeKeys,Smallest,Largest}, etc.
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2922#-:-NengaMfDCd5z9XlDbNv:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sumeerbhola)

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2922#-:-Nenko1343jElDdGc_bv:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @itsbilal)

itsbilal:
TFTR!

PR #2926, opened 2023-09-21
db: get rid of readstate in EFOS

bananabrick:
It's unnecessary to grab a read state when the EFOS is first created.

Case 1: No excises.
If there's no excises, then any iterator created on the EFOS prior to it transition to an FOS can read using only a sequence number. As memtables are flushed, and compactions occur, the keys relevant to the EFOS will be preserved.

Once the EFOS transitions to an FOS, we don't need a read state, because we grab a version ref for the latest version, which contains all sstables with keys which are relevant to the FOS.

Case 2: Excises.
If a relevant excise occurs after transition to FOS, we don't care, because the FOS is performing reads using a version which contains all the relevant files.

If an excise occurs before transition to an FOS, then all iterators already created on the EFOS will still be valid because they read using a sequence number, and a read state which was grabbed prior to excision.

No new iterators are created once an excise occurs.

### DIFF
--- a/snapshot.go
+++ b/snapshot.go
@@ -205,7 +205,7 @@ func (l *snapshotList) remove(s *Snapshot) {
 }
 
 // EventuallyFileOnlySnapshot (aka EFOS) provides a read-only point-in-time view
-// of the database state, similar to Snapshot. A EventuallyFileOnlySnapshot
+// of the database state, similar to Snapshot. An EventuallyFileOnlySnapshot
 // induces less write amplification than Snapshot, at the cost of increased space
 // amplification. While a Snapshot may increase write amplification across all
 // flushes and compactions for the duration of its lifetime, an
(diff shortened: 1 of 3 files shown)


snapshot.go
```diff
@@ -311,6 +303,8 @@ func (d *DB) makeEventuallyFileOnlySnapshot(
 // call.
 //
 // d.mu must be held when calling this method.
+//
+// TODO(bananabrick): Must we also hold the log lock?
```
itsbilal, line 307: This is unnecessary because we only need the db lock to be able to close the snapshot at the same time as us doing the transition, which doesn't require us to finish installing a version.

snapshot.go
```diff
@@ -511,8 +479,15 @@ func (es *EventuallyFileOnlySnapshot) NewIterWithContext(
 	if es.excised.Load() {
```
itsbilal, line 479: We can remove this

itsbilal (approved):
LGTM save for minor comments

snapshot.go
bananabrick: Kept it in because the work done to perform an atomic load is a lot less than performing an unnecessary iterator construction. Left it in for now.

snapshot.go
```diff
@@ -511,8 +479,15 @@ func (es *EventuallyFileOnlySnapshot) NewIterWithContext(
...
 	}
-	sOpts := snapshotIterOpts{seqNum: es.seqNum, readState: es.mu.readState}
-	return es.db.newIter(ctx, nil /* batch */, sOpts, o), nil
+	sOpts := snapshotIterOpts{seqNum: es.seqNum}
+	iter := es.db.newIter(ctx, nil /* batch */, sOpts, o)
+
+	// If excised is true, then keys relevant to the snapshot might not be
+	// present in the readState being used by the iterator. Error out.
+	if es.excised.Load() {
+		return nil, ErrSnapshotExcised
```
bananabrick, line 488: Had forgotten to close the iterator here. Added that.

PR #2928, opened 2023-09-21
db: by default don't log to standard output

jbowens:
Previously, by default an Options struct without a set Logger would default to logging to stdout using Go's standard library logging. This made unit tests particularly verbose. This is a backwards incompatible change, and to restore previous behavior clients that leave Logger unset should explicitly set it to pebble.DefaultLogger.

### DIFF
--- a/error_test.go
+++ b/error_test.go
@@ -12,20 +12,10 @@ import (
 	"sync/atomic"
 	"testing"
 
-	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/vfs"
 	"github.com/cockroachdb/pebble/vfs/errorfs"
 	"github.com/stretchr/testify/require"
 )
-
-type panicLogger struct{}
-
-func (l panicLogger) Infof(format string, args ...interface{}) {
-}
-
-func (l panicLogger) Fatalf(format string, args ...interface{}) {
-	panic(errors.Errorf("fatal: "+format, args...))
-}
 
 // corruptFS injects a corruption in the `index`th byte read.
 type corruptFS struct {
--- a/ingest_test.go
+++ b/ingest_test.go
@@ -788,7 +788,7 @@ func TestIngestShared(t *testing.T) {
 			FormatMajorVersion:    ExperimentalFormatVirtualSSTables,
 		}
 		// lel.
-		lel := MakeLoggingEventListener(DefaultLogger)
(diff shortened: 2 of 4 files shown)


jbowens:
Thoughts on this? I don't think it's too burdensome of a change in defaults and makes our lives easier.

RaduBerinde (commented):
Don't we want these logs for unit tests? They should only be visible when using `-test.v` no?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2928#-:-NeskUvcCULNu2a1411q:b6ayj7j)* status: 0 of 4 files reviewed, all discussions resolved

jbowens (commented):
> Don't we want these logs for unit tests? They should only be visible when using `-test.v` no?

Running a single package's tests will show all standard output. But `go test ./...` will silence it. I think tests can (and ideally should) use `testLogger` which will log to `(*testing.T).Log`, so that it's only visible when using `-test.v` or during failures.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2928#-:-Nesr3Al5POepdG03Rqo:b-m0jff7)* status: 0 of 4 files reviewed, all discussions resolved

karalabe:
Would it be possible to also expose a NoopLogger / NoopLoggerAndTracer to allow setting those? They are defined in `base` but that's internal so I can't reference it. Would rather not define an empty logging interface in our code just to silence Pebble.

RaduBerinde:
> Running a single package's tests will show all standard output. But go test ./... will silence it. 

Doesn't that mean that if we have a CI failure, we won't see the logs?

jbowens (commented):
The Pebble EventListener logs, yes, unless the test uses a `testLogger` to direct log output to the `*testing.T`. I don't think the Pebble logs are very useful for debugging unit tests, with the exception being `Fatalf`s which will still be visible (since they panic). In my experience, the intertwined logs (especially since they recently became more verbose) just make it harder to find the actual test failure.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2928#-:-Ngz5cDy7PjbernOAF2e:b-2e8ye2)* status: 0 of 4 files reviewed, all discussions resolved

RaduBerinde:
Sounds good. I'm still confused how there is a difference in behavior depending what packages we `go test`?

xxmplus:
Close to due to age

Issue #2041, opened 2022-10-21
metamorphic: vary operation distribution


jbowens:
Currently, the metamorphic tests use [a fixed distribution of operations](https://github.com/cockroachdb/pebble/blob/f34af25a01873c9cf027e53147754e96ec8b2b53/internal/metamorphic/config.go#L79-L123). Varying the distributions among runs would help increase coverage.

For example, the current frequency of range deletions is just half of sets. Having some runs write many range deletions like this is useful for exercising edge cases around range deletions and interactions between range deletions. However it also significantly reduces the amount of data accumulated over the course of the metamorphic tests. We would improve coverage in other dimensions by including some runs with fewer or narrower range deletions, allowing the database working set to grow more substantially over the course of the test.

PR #2929, opened 2023-09-21
metamorphic: add variant that constructs many versions

### LINKED ISSUE #2921: metamorphic: increase number of key versions to exercise restart code in nextPrefixV3
The code in https://github.com/cockroachdb/pebble/blob/f509cd48c70c622238091f02b9e04be5cf108797/sstable/block.go#L1262-L1319 is not exercised in the metamorphic test.

For example, see the coverage in https://storage.googleapis.com/crl-codecover-public/pebble/2023-09-20%2008:16Z%20f509cd48%20-%20meta%20test%20only/pebble/sstable/block.go.gcov.html

jbowens:
Previously, the metamorphic test always generated operations according to a fixed distribution. This commit adds a second configuration and adjusts the test to randomly chose between available configurations. The new configuration strongly prefers generating versions of existing key prefixes when generating a new key. Additionally, it generates significantly fewer deletion write operations in order to allow a test run to accumulate a more sizable working set of live keys.

This new configuration is sufficient to exercise all the non-error cases within the ssblock nextPrefix code paths (#2921).

The overall impact on metamorphic test code coverage is minor, moving from 67.3% to 67.7%.

Close #2921.
Informs #2041.

### DIFF
--- a/metamorphic/config.go
+++ b/metamorphic/config.go
@@ -74,6 +74,39 @@ type config struct {
 	// upperBoundFrac float64
 }
 
+func (c config) withNewPrefixProbability(p float64) config {
+	c.newPrefix = p
+	return c
+}
+
+func (c config) withOpWeight(op opType, weight int) config {
+	c.ops[op] = weight
+	return c
+}
+
+var presetConfigs = []config{
+	defaultConfig(),
+	// Generate a configuration that helps exercise code paths dependent on many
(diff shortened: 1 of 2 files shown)


itsbilal (approved):
LGTM :shipit:

jbowens:
TFTR!

PR #2934, opened 2023-09-25
db: add SkipPoint iterator option

jbowens:
Add a new IterOption that allows a user to configure an arbitrary predicate of point keys that should be skipped in the form of a function. The Iterator consults this predicate for every internal key.

Use this new SkipPoint function to remove the complicated logic within the metamorphic test for ensuring determinisim of RangeKeyChanged() in the presence of time-bound block-property filters. A SkipPoint predicate may be combined with a block-property filter to achieve deterministic iteration.

### DIFF
--- a/metamorphic/retryable.go
+++ b/metamorphic/retryable.go
@@ -5,11 +5,8 @@
 package metamorphic
 
 import (
-	"bytes"
-
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble"
-	"github.com/cockroachdb/pebble/internal/testkeys"
 	"github.com/cockroachdb/pebble/vfs/errorfs"
 )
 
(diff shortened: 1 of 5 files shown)


metamorphic/retryable.go
```diff
@@ -31,68 +28,6 @@ func withRetries(fn func() error) error {
...
-	if n == len(k) {
-		// No suffix, don't filter it.
-		return false
-	}
-	v, err := testkeys.ParseSuffix(k[n:])
-	if err != nil {
-		panic(err)
-	}
-	ts := uint64(v)
-	return ts < i.filterMin || ts >= i.filterMax
```
itsbilal, line 95: Nice

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2934#-:-NfD_ngT4S6dSlHXxRB6:b-814g8k)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

PR #2937, opened 2023-09-26
db: don't ignore range key bounds during excise

bananabrick:


testdata/excise
```diff
@@ -162,3 +162,39 @@ would excise 2 files, use ingest-and-excise to excise.
...
+
+build ext2
+set z z
+----
+
+ingest-and-excise ext2 excise=b-c
+----
+
+# TODO(bananabrick): This looks insanely incorrect. Why is the second sstable
+# bound d-d?
```
itsbilal, line 193: TODO should be taken out now?

itsbilal (approved):
Thanks for finding and fixing this quickly! LGTM :shipit:

testdata/excise
bananabrick: Lol my bad

bananabrick:
race build probably not going to fail due to this. merging.

PR #2942, opened 2023-09-27
metamorphic: Only enable ingest split if FMV is >= VirtualSSTs

### LINKED ISSUE #2940: metamorphic: TestMeta failed
This reproduces readily with `make stressmeta`

```
=== CONT  TestMeta/execution/random-022
=== CONT  TestMeta/execution/standard-018
=== NAME  TestMeta/execution/random-026
    meta.go:219: 
        ===== SEED =====
        1695823945473781000
        ===== ERR =====
        exit status 1
        ===== OUT =====
        MANIFEST cannot contain virtual sstable records due to format major version
```

itsbilal:
Previously we were enabling ingest splits even if the format major version was < FormatVirtualSSTables, resulting in errors lower down in the manifest when we created backing tables. This change fixes it.

Fixes #2940.

### DIFF
--- a/metamorphic/options.go
+++ b/metamorphic/options.go
@@ -548,7 +548,7 @@ func randomOptions(
 		}
 	}
 	testOpts.seedEFOS = rng.Uint64()
-	testOpts.ingestSplit = rng.Intn(2) == 0
+	testOpts.ingestSplit = rng.Intn(2) == 0 && opts.FormatMajorVersion >= pebble.FormatVirtualSSTables
 	opts.Experimental.IngestSplit = func() bool { return testOpts.ingestSplit }
 
 	return testOpts


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2942)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`metamorphic/options.go` line 551 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2942#-NfMR6rtDtpxzqnkb1Nk:-NfMR6rtDtpxzqnkb1Nl:bw1j6cv) ([raw file](https://github.com/cockroachdb/pebble/blob/17e473fd8facec291acb46f5d94e863b0c2a7aa9/metamorphic/options.go#L551)):*
> ```Go
> 	}
> 	testOpts.seedEFOS = rng.Uint64()
> 	testOpts.ingestSplit = rng.Intn(2) == 0 && opts.FormatMajorVersion >= pebble.FormatVirtualSSTables
> ```

I think we need to prevent the split ingestion within `Ingest` instead.

Also, the active format major version of a DB in the metamorphic tests may change over the course of the test, so it'd be nice to allow runs to begin using ingest-with-split once they ratchet to a sufficiently high version.

itsbilal:
*[`metamorphic/options.go` line 551 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2942#-NfMR6rtDtpxzqnkb1Nk:-NfMRvLQ5JJbCsrNkXZn:bvs8nrm) ([raw file](https://github.com/cockroachdb/pebble/blob/17e473fd8facec291acb46f5d94e863b0c2a7aa9/metamorphic/options.go#L551)):*

I think we need to prevent the split ingestion within `Ingest` instead.

Also, the active format major version of a DB in the metamorphic tests may change over the course of the test, so it'd be nice to allow runs to begin using ingest-with-split once they ratchet to a sufficiently high version.
</blockquote></details>

We're already doing that as it turns out:

https://github.com/cockroachdb/pebble/blob/4df2ce806152dd350a54c6c790bd0844fd2926da/ingest.go#L2014

There's something else fishy that's going on and I don't even see the failing ingest in the metamorphic test's operations? Taking a closer look

itsbilal:
Ah, found it. The error conditional in the Manifest code is looking at the wrong FMV, the one the DB was created with, not the current one.

PR #2932, opened 2023-09-22
metamorphic: fix block property collectors integration

jbowens:
Fix a bug where the testkeys timestamp suffix block property collectors were unused. This is coverage that we had but was lost at some point.

I've added a focused unit test to cover this case. This increases overall meta-only test coverage from 67.9% to 68.6%.

Close #2931.

PR #2943, opened 2023-09-27
Revert "db: add SkipPoint iterator option"

jbowens:
This reverts commit 895ffed51f35c1e1867b7b6bf6330e8bec58210c.

This was insufficiently tested because #2932 wasn't merged yet, and there
appears to be an unidentified interaction. Reverting this will allow us to
restore testing of block property filters, and then I can reexamine this
refactor.


jbowens:
TFTR!

PR #2946, opened 2023-09-27
db: add SkipPoint iterator option

jbowens:
This commit was added in #2934 and reverted in #2943. It's brought back with a bug fix for the case where a range key is truncated to a user key that also exists as a point key, and that point key should be skipped by the predicate.

----

Add a new IterOption that allows a user to configure an arbitrary predicate of point keys that should be skipped in the form of a function. The Iterator consults this predicate for every internal key.

Use this new SkipPoint function to remove the complicated logic within the metamorphic test for ensuring determinisim of RangeKeyChanged() in the presence of time-bound block-property filters. A SkipPoint predicate may be combined with a block-property filter to achieve deterministic iteration.

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -550,6 +550,21 @@ func (i *Iterator) findNextEntry(limit []byte) {
 			return
 		}
 
+		// If the user has configured a SkipPoint function, invoke it to see
+		// whether we should skip over the current user key.
+		if i.opts.SkipPoint != nil && key.Kind() != InternalKeyKindRangeKeySet && i.opts.SkipPoint(i.iterKey.UserKey) {
+			// NB: We could call nextUserKey, but in some cases the SkipPoint
+			// predicate function might be cheaper than nextUserKey's key copy
+			// and key comparison. This should be the case for MVCC suffix
+			// comparisons, for example. In the future, we could expand the
+			// SkipPoint interface to give the implementor more control over
+			// whether we skip over just the internal key, the user key, or even
+			// the key prefix.
+			i.stats.ForwardStepCount[InternalIterCall]++
+			i.iterKey, i.iterValue = i.iter.Next()
+			continue
+		}
+
 		switch key.Kind() {
 		case InternalKeyKindRangeKeySet:
 			// Save the current key.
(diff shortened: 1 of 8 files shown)


jbowens:
Putting this in draft mode until I've stressed the metamorphic test sufficiently.

jbowens:
> Putting this in draft mode until I've stressed the metamorphic test sufficiently.

Only turned up unrelated failures, which I filed issues for.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2946)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`iterator.go` line 642 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2946#-NfND7GB58DusEyZDG4U:-NfND7GB58DusEyZDG4V:b-xqlplb) ([raw file](https://github.com/cockroachdb/pebble/blob/bb8dd233feac04d1404ce347aac1a90ce4283317/iterator.go#L642)):*
> ```Go
> 	if i.opts.SkipPoint != nil && i.opts.SkipPoint(i.key) {
> 		return false
> 	}
> ```

This is is the new bit that the metamorphic test uncovered. ^

itsbilal (approved):
:lgtm:

> Reviewed 5 of 5 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2946#-:-NfNL2kq007mKOf6GGe9:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`iterator.go` line 642 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2946#-NfND7GB58DusEyZDG4U:-NfNLE9P8E2mDEHucURS:b-7m74ox) ([raw file](https://github.com/cockroachdb/pebble/blob/bb8dd233feac04d1404ce347aac1a90ce4283317/iterator.go#L642)):*

This is is the new bit that the metamorphic test uncovered. ^
</blockquote></details>

nice!

Issue #2947, opened 2023-09-27
metamorphic: TestMeta failed [duplicate FileBacking present in obsolete list]


jbowens:
```
=== NAME  TestMeta/execution/random-020
    meta.go:218: 
        ===== SEED =====
        1695838611509616000
        ===== ERR =====
        exit status 2
        ===== OUT =====
        --- FAIL: TestMeta (0.60s)
        panic: pebble: duplicate FileBacking present in obsolete list [recovered]
        	panic: pebble: duplicate FileBacking present in obsolete list
        
        goroutine 34 [running]:
        testing.tRunner.func1.2({0x10075bca0, 0x10093eb28})
        	/usr/local/go/src/testing/testing.go:1526 +0x24e
        testing.tRunner.func1()
        	/usr/local/go/src/testing/testing.go:1529 +0x39f
        panic({0x10075bca0, 0x10093eb28})
        	/usr/local/go/src/runtime/panic.go:884 +0x213
        github.com/cockroachdb/pebble.(*versionSet).addObsoleteLocked(0xc00075ec00, {0xc000852e00, 0xf, 0xc000119801?})
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/version_set.go:807 +0x3aa
        github.com/cockroachdb/pebble/internal/manifest.(*Version).UnrefLocked(0xc0002d9b00)
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/internal/manifest/version.go:1229 +0x165
        github.com/cockroachdb/pebble.(*EventuallyFileOnlySnapshot).Close(0xc0005c1950)
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/snapshot.go:422 +0x10b
        github.com/cockroachdb/pebble/metamorphic.(*closeOp).run(0xc0004c2bcc, 0xc0001880e0, {0xc0001514e0?, 0xc00001db00?})
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/metamorphic/ops.go:158 +0x1e5
        github.com/cockroachdb/pebble/metamorphic.(*test).step(0xc0001880e0, 0xc0002849c0?)
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/metamorphic/test.go:251 +0x4f
[...]

PR #2951, opened 2023-09-28
db: only create one CreatedBackingTables entry per sstable

bananabrick:
Problem explained in detail in: https://github.com/cockroachdb/pebble/issues/2947#issuecomment-1738225516

We were adding too many CreatedBackingTables entries to the manifest. Such an
entry should only be added when an sstable is first virtualized. This is
mentioned as an invariant above VersionEdit.CreatedBackingTables.

Note that the excise function will create backing tables and add them to the
version edit. But these backing tables should only be added to the version edit,
if the sstable being virtualized is NOT virtual. If we're further virtualizing
an already virtual sstable, then a backing table entry associated with the file
backing will already be present in the manifest as part of an older version edit.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -1608,7 +1608,7 @@ func (d *DB) excise(
 	var iter internalIterator
 	var rangeDelIter keyspan.FragmentIterator
 	var rangeKeyIter keyspan.FragmentIterator
-	backingTableCreated := false
+	needsBacking := false
 	// Create a file to the left of the excise span, if necessary.
 	// The bounds of this file will be [m.Smallest, lastKeyBefore(exciseSpan.Start)].
 	//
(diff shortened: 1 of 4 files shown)


bananabrick:
Stressed the failing test for a while:
```
96 runs so far, 0 failures, over 30m0s
96 runs so far, 0 failures, over 30m5s
96 runs completed, 0 failures, over 30m7s
context canceled
```

Looks good.

bananabrick:
invariant check in 2nd commit is pretty cool: https://github.com/cockroachdb/pebble/pull/2951/commits/7ae0092deab8ed9197b81a1b67e362594b3e329d

It would've caught this problem by failing existing tests like `TestIngest`.

jbowens (approved):
Good sleuthing

:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2951#-:-NfRJqul5PoHwRnJeaHB:bllr1px)* status: all files reviewed, 2 unresolved discussions (waiting on @bananabrick and @itsbilal)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2951#-NfRJUrn0DFs04l-u5y-):*
nit: can we include a brief summary in addition to the link for posterity?
___
*[`internal/manifest/version_edit.go` line 790 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2951#-NfRJkMv3q0DBT216Ztr:-NfRJkMv3q0DBT216Zts:b-p6g52m) ([raw file](https://github.com/cockroachdb/pebble/blob/7ae0092deab8ed9197b81a1b67e362594b3e329d/internal/manifest/version_edit.go#L790)):*
> ```Go
> 			// This should never happen. There must always be only one FileBacking
> 			// associated with a backing sstable.
> 			panic("pebble: duplicate file backing")
> ```

nit: should we include the DiskFileNum in the panic message?

ingest.go
```diff
@@ -1722,8 +1722,7 @@ func (d *DB) excise(
 			leftFile.ValidateVirtual(m)
 			d.checkVirtualBounds(leftFile)
 			ve.NewFiles = append(ve.NewFiles, newFileEntry{Level: level, Meta: leftFile})
-			ve.CreatedBackingTables = append(ve.CreatedBackingTables, leftFile.FileBacking)
-			backingTableCreated = true
+			needsBacking = true
 			numCreatedFiles++
 		}
 	}
```
itsbilal, line 1728: We early-return if there's no file to the right of the excise span below. So we'll need to copy the `if needsBacking` case below the return down in line 1732.

itsbilal (changes requested):
Thanks for doing this! One key thing

ingest.go
bananabrick: Added a test for this case, and fixed it.

PR #2952, opened 2023-09-28
db: cleanup btree obsoletion logic

bananabrick:
1. By returning FileBackings from the LevelMetadata.release functions, we can clean up some duplication in Version.Unref and Version.UnrefLocked.

2.
It is not obvious how Pebble determines when it is okay to delete the backing sstable associated with a virtual sstable.

The flow is that Filemetadata refcounting actually happens on the FileBacking. Multiple virtual sstables with the same backing sstable will share the same FileBacking struct. Once the refcount on the FileBacking struct hits 0, we know that 0 virtual sstables require that FileBacking, and therefore it is safe to delete the backing sstable.

This pr adds some commentary which explains this.

It was a bit awkward prior to this pr when a btree returns an obsolete virtual metadata, and then we grab the filebacking from the obsolete virtual meta to delete that.

### DIFF
--- a/internal/manifest/btree.go
+++ b/internal/manifest/btree.go
@@ -195,7 +195,7 @@ func (n *node) incRef() {
 // new nodes pass contentsToo=false to preserve existing reference counts during
 // operations that should yield a net-zero change to descendant refcounts.
 // When a node is released, its contained files are dereferenced.
-func (n *node) decRef(contentsToo bool, obsolete *[]*FileMetadata) {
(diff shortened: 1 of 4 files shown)


jbowens (approved):
Nice

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2952#-:-NfRLt7V6QvjI9xvsPR0:b-293xf2)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @itsbilal)

PR #2957, opened 2023-09-28
db: add metrics for num backing sstables and size

bananabrick:
Pr just adds backing size/count into the Metrics struct and
grabs them from the version set. Needs loglock to grab these
as the fields are protected by the loglock.

### DIFF
--- a/db.go
+++ b/db.go
@@ -1963,6 +1963,17 @@ func (d *DB) Metrics() *Metrics {
 
 	d.mu.versions.logLock()
 	metrics.private.manifestFileSize = uint64(d.mu.versions.manifest.Size())
+	metrics.Table.BackingTableCount = uint64(len(d.mu.versions.backingState.fileBackingMap))
+	metrics.Table.BackingTableSize = d.mu.versions.backingState.fileBackingSize
+	if invariants.Enabled {
+		var totalSize uint64
+		for _, backing := range d.mu.versions.backingState.fileBackingMap {
+			totalSize += backing.Size
+		}
+		if totalSize != metrics.Table.BackingTableSize {
+			panic("pebble: invalid backing table size accounting")
+		}
+	}
 	d.mu.versions.logUnlock()
 
 	metrics.LogWriter.FsyncLatency = d.mu.log.metrics.fsyncLatency
--- a/metrics.go
+++ b/metrics.go
@@ -253,6 +253,10 @@ type Metrics struct {
 		ZombieSize uint64
 		// The count of zombie tables.
 		ZombieCount int64
+		// The count of the backing sstables.
+		BackingTableCount uint64
+		// The sum of the sizes of the all of the backing sstables.
+		BackingTableSize uint64
 	}
 
 	TableCache CacheMetrics
@@ -527,6 +531,10 @@ func (m *Metrics) SafeFormat(w redact.SafePrinter, _ rune) {
 	w.Printf("Zombie tables: %d (%s)\n",
 		redact.Safe(m.Table.ZombieCount),
 		humanize.Bytes.Uint64(m.Table.ZombieSize))
+
+	w.Printf("Backing tables: %d (%s)\n",
+		redact.Safe(m.Table.BackingTableCount),
+		humanize.Bytes.Uint64(m.Table.BackingTableSize))
 
 	formatCacheMetrics := func(m *CacheMetrics, name redact.SafeString) {
 		w.Printf("%s: %s entries (%s)  hit rate: %.1f%%\n",
--- a/metrics_test.go
+++ b/metrics_test.go
@@ -5,6 +5,7 @@
 package pebble
 
 import (
+	"bytes"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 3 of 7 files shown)


bananabrick (commented):
1. More metrics will be added soon. Can be a follow-up pr too.
2. Are these two metrics even useful? I guess once I add the virtual sstable size, the backing size can be used to compute space amp.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2957#-:-NfT6MwD1J0JYlJumIR8:b6vhx01)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @jbowens)

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2957)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`metrics_test.go` line 248 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2957#-NfT6wvs3J5yooaOUQj7:-NfT6wvt-msJjaW3nMXX:bvb3b7) ([raw file](https://github.com/cockroachdb/pebble/blob/4565d99fc1265dc53c86781999e24918e56023d4/metrics_test.go#L248)):*
> ```Go
> 
> 		case "metrics-value":
> 			// metrics-value confirms the value of a given metric. Note that there
> ```

Dunno what the best way to get rid of this non-determinism is. I don't want to add a flag which disables the invariant check just for this test case.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2957)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`metrics_test.go` line 248 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2957#-NfT6wvs3J5yooaOUQj7:-NfTP67a65SDrwMC5cm6:bgdjedl) ([raw file](https://github.com/cockroachdb/pebble/blob/4565d99fc1265dc53c86781999e24918e56023d4/metrics_test.go#L248)):*

Dunno what the best way to get rid of this non-determinism is. I don't want to add a flag which disables the invariant check just for this test case.
</blockquote></details>

Also don't want to selectively remove the cache hit rates from the level metrics. Instead it's just better to print the values I care about which is what I did here.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2957)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`db.go` line 1967 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2957#-NfhBWPe1Gj7thdmWuJW:-NfhBWPf2a9tSRqgUjAv:b-6pjwsu) ([raw file](https://github.com/cockroachdb/pebble/blob/86255356ab1d819f8dbaf99f4ced0ef712a05624/db.go#L1967)):*
> ```Go
> 	metrics.private.manifestFileSize = uint64(d.mu.versions.manifest.Size())
> 	metrics.Table.BackingTableCount = uint64(len(d.mu.versions.fileBackingMap))
> 	for _, backing := range d.mu.versions.fileBackingMap {
> ```

I think this loop is fine for a metrics call. Should I maintain this incrementally?

jbowens (commented):
> Reviewed 4 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2957)* status: 4 of 7 files reviewed, 2 unresolved discussions (waiting on @bananabrick)

___
*[`db.go` line 1967 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2957#-NfhBWPe1Gj7thdmWuJW:-NfkzjuF2BDm1u9zT-W7:b19xz65) ([raw file](https://github.com/cockroachdb/pebble/blob/86255356ab1d819f8dbaf99f4ced0ef712a05624/db.go#L1967)):*

I think this loop is fine for a metrics call. Should I maintain this incrementally?
</blockquote></details>

Yeah, we need to avoid loops that grow linearly with the size of the LSM. We've observed O(# files) loops showing up prominently in Cockroach metrics collection before. It will be especially important with disaggregated storage as node density increases.
___
*[`metrics_test.go` line 248 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2957#-NfT6wvs3J5yooaOUQj7:-Nfl0Gsh8MsSbBn5Thko:b-fmhngn) ([raw file](https://github.com/cockroachdb/pebble/blob/4565d99fc1265dc53c86781999e24918e56023d4/metrics_test.go#L248)):*

Also don't want to selectively remove the cache hit rates from the level metrics. Instead it's just better to print the values I care about which is what I did here.
</blockquote></details>

I think this is fine and agree with the TODO that a generic command that uses reflection would be nice.

jbowens (approved):
:lgtm_strong: 

> Reviewed 1 of 7 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2957#-:-NfmKhEY2VwSrbiU1c0B:b-uwkowd)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bananabrick)

bananabrick:
Thx.

PR #2961, opened 2023-10-01
db: keep track of virtual sstable count in metrics

bananabrick:
This pr adds an int to the LevelMetadata and incrementally
keeps track of the number of virtual sstables in the level.

Also adds a field in the LevelMetrics to keep track of virtual
sstables, and print the virtual sstables as part of the metrics
string.

PR #2962, opened 2023-10-02
db: keep track of virtual sstable size sum

bananabrick:
The virtual sstable size sum is useful because it along with the
backing table size can be used to compute the space amp associated
with virtual sstables.

### DIFF
--- a/internal/manifest/level_metadata.go
+++ b/internal/manifest/level_metadata.go
@@ -19,17 +19,20 @@ type LevelMetadata struct {
 	totalSize uint64
 	// NumVirtual is the number of virtual sstables in the level.
 	NumVirtual uint64
-	tree       btree
+	// VirtualSize is the size of the virtual sstables in the level.
+	VirtualSize uint64
+	tree        btree
 }
 
 // clone makes a copy of the level metadata, implicitly increasing the ref
 // count of every file contained within lm.
 func (lm *LevelMetadata) clone() LevelMetadata {
 	return LevelMetadata{
-		level:      lm.level,
-		totalSize:  lm.totalSize,
-		NumVirtual: lm.NumVirtual,
-		tree:       lm.tree.Clone(),
+		level:       lm.level,
+		totalSize:   lm.totalSize,
+		NumVirtual:  lm.NumVirtual,
+		VirtualSize: lm.VirtualSize,
+		tree:        lm.tree.Clone(),
 	}
 }
 
@@ -49,6 +52,7 @@ func makeLevelMetadata(cmp Compare, level int, files []*FileMetadata) LevelMetad
 		lm.totalSize += f.Size
 		if f.Virtual {
 			lm.NumVirtual++
+			lm.VirtualSize += f.Size
 		}
 	}
 	return lm
@@ -70,6 +74,7 @@ func (lm *LevelMetadata) insert(f *FileMetadata) error {
 	lm.totalSize += f.Size
 	if f.Virtual {
 		lm.NumVirtual++
+		lm.VirtualSize += f.Size
 	}
 	return nil
 }
@@ -78,6 +83,7 @@ func (lm *LevelMetadata) remove(f *FileMetadata) bool {
 	lm.totalSize -= f.Size
 	if f.Virtual {
 		lm.NumVirtual--
+		lm.VirtualSize -= f.Size
 	}
 	return lm.tree.Delete(f)
 }
@@ -326,6 +332,19 @@ func (ls *LevelSlice) NumVirtual() uint64 {
 		}
 	}
 	return n
+}
+
+// VirtualSizeSum returns the sum of the sizes of the virtual sstables in the
+// level.
+func (ls *LevelSlice) VirtualSizeSum() uint64 {
+	var sum uint64
+	iter := ls.Iter()
+	for f := iter.First(); f != nil; f = iter.Next() {
+		if f.Virtual {
+			sum += f.Size
+		}
+	}
+	return sum
 }
 
 // Reslice constructs a new slice backed by the same underlying level, with
--- a/metrics.go
+++ b/metrics.go
@@ -48,6 +48,8 @@ type LevelMetrics struct {
 	NumVirtualFiles uint64
 	// The total size in bytes of the files in the level.
 	Size int64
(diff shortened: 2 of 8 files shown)


bananabrick:
First two commits are from https://github.com/cockroachdb/pebble/pull/2961

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2962)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`db.go` line 1969 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2962#-NfmIPRWALTOoFptq5ve:-NfmIPRWALTOoFptq5vf:b-60how4) ([raw file](https://github.com/cockroachdb/pebble/blob/25a190a082b730c311439f57ed47b03cf832d32e/db.go#L1969)):*
> ```Go
> 	for _, backing := range d.mu.versions.fileBackingMap {
> 		metrics.Table.BackingTableSize += backing.Size
> 		metrics.Table.VirtualSize += backing.VirtualizedSize.Load()
> ```

This pr needs more work now that we're no longer looping through all of the backing tables in Metrics. Maybe we'll need a per level virtual sstable size metric, which we sum over here.

bananabrick (commented):
Should be good again, first commit is from https://github.com/cockroachdb/pebble/pull/2961.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2962#-:-NfmYysJ65n7s_R33KJ1:b-u3imsw)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

bananabrick (commented):
Merged the other commits, so just this commit should be good for review.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2962#-:-NfrTIVP8IvfabwumXSy:b68y5yd)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2962#-:-NfrnaDW7pnCGlIZdUFJ:bnfp4nl)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @bananabrick and @sumeerbhola)

___
*[`metrics.go` line 49 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2962#-NfrS6rIDNOFZJYow0Q1:-NfrS6rIDNOFZJYow0Q2:bp3ondk) ([raw file](https://github.com/cockroachdb/pebble/blob/bdc6a5555cc804d27cf5ed890134604e2dd19fc9/metrics.go#L49)):*
> ```Go
> 	// The total number of virtual sstables in the level.
> 	NumVirtualFiles uint64
> 	// The total size of the virtual sstables in the level.
> ```

super-ocd-nit but I'd put it below Size so it's the same relative ordering as the files
___
*[`metrics_test.go` line 75 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2962#-NfrnJqH3zbiJ38hCM3e:-NfrnJqH3zbiJ38hCM3f:b-lasfi3) ([raw file](https://github.com/cockroachdb/pebble/blob/bdc6a5555cc804d27cf5ed890134604e2dd19fc9/metrics_test.go#L75)):*
> ```Go
> 		l.VirtualSize = 10 << 10
> 		l.NumVirtualFiles = 1
> 		l.Size = int64(base) + 2
> ```

[nit] but why not use `base + x` like everything else. The values don't need to make sense, it's better if they vary from row to row.

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2962)* status: 0 of 9 files reviewed, all discussions resolved (waiting on @sumeerbhola)

___
*[`metrics.go` line 49 at r5](https://reviewable.io/reviews/cockroachdb/pebble/2962#-NfrS6rIDNOFZJYow0Q1:-NfruuEk9uOdmva3QLGY:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/bdc6a5555cc804d27cf5ed890134604e2dd19fc9/metrics.go#L49)):*

super-ocd-nit but I'd put it below Size so it's the same relative ordering as the files
</blockquote></details>

done

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2962)* status: 0 of 9 files reviewed, 1 unresolved discussion

___
*[`testdata/metrics` line 22 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2962#-NfvPnV27NIqY5Hv9FCq:-NfvPnV27NIqY5Hv9FCr:bu7mmsr) ([raw file](https://github.com/cockroachdb/pebble/blob/8fa7470c378ed5f9cb12b647cb471ab8a0f3c7a7/testdata/metrics#L22)):*
> ```
> Zombie tables: 16 (15B)
> Backing tables: 1 (2.0MB)
> Virtual tables: 2807 (2.8KB)
> ```

Just checking that I understand how to interpret this (of course this is rather an outlier)
There is 1 backing table which is being referenced by 2807 virtual tables. The total virtual sst size is 2.8KB and the total backing size is 2MB, so (2MB-2.8KB) is the wasted space that is not being referenced by anything.

Is that correct?

Are we only printing the virtual table size sum and not per-level since this would consume yet another column?

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2962)* status: 0 of 9 files reviewed, 1 unresolved discussion

___
*[`testdata/metrics` line 22 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2962#-NfvPnV27NIqY5Hv9FCq:-Nfw8sQp4cfP4-CJN-Xd:b-x537fj) ([raw file](https://github.com/cockroachdb/pebble/blob/8fa7470c378ed5f9cb12b647cb471ab8a0f3c7a7/testdata/metrics#L22)):*
That's right. The virtual sstable sizes are estimates, so the wasted space is also an estimate. 

> Are we only printing the virtual table size sum and not per-level since this would consume yet another column?

Yea, we could add that. Wasn't sure if we cared about per level space amp.

PR #2964, opened 2023-10-03
db: add test for virtual sstable checkpointing

bananabrick:
Virtual sstable checkpointing will remove the backing files which won't be required by the checkpoint. Need to make sure that the virtual sstables which are present in the checkpoint manifest are still readable, and that the backing files not required are deleted.

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -135,9 +135,6 @@ func mkdirAllAndSyncParents(fs vfs.FS, destDir string) (vfs.File, error) {
 // space overhead for a checkpoint if hard links are disabled. Also beware that
 // even if hard links are used, the space overhead for the checkpoint will
 // increase over time as the DB performs compactions.
-//
-// TODO(bananabrick): Test checkpointing of virtual sstables once virtual
-// sstables is running e2e.
 func (d *DB) Checkpoint(
 	destDir string, opts ...CheckpointOption,
 ) (
--- a/checkpoint_test.go
+++ b/checkpoint_test.go
@@ -5,6 +5,7 @@
 package pebble
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"math/rand"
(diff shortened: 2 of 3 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2964#-:-Nfr8AJ45DxqFO5P0dws:bnfp4nl)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @bananabrick)

___
*[`testdata/checkpoint` line 496 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2964#-Nfr6rs9EhrFVRzf2UJu:-Nfr6rs9EhrFVRzf2UJv:b56001r) ([raw file](https://github.com/cockroachdb/pebble/blob/b7826d90135f0fb0e6f551120ee699cc232697a9/testdata/checkpoint#L496)):*
> ```
> ----
> 6:
>   000012:[a#0,SET-b#0,SET]
> ```

[nit] we should improve this output to show which SSTs are virtual, and show the backing file ids. (maybe you can file an issue if it's not trivial)

bananabrick (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2964)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`testdata/checkpoint` line 496 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2964#-Nfr6rs9EhrFVRzf2UJu:-NfrLn_uBPd6N2cVt_Do:bq0ncok) ([raw file](https://github.com/cockroachdb/pebble/blob/b7826d90135f0fb0e6f551120ee699cc232697a9/testdata/checkpoint#L496)):*

[nit] we should improve this output to show which SSTs are virtual, and show the backing file ids. (maybe you can file an issue if it's not trivial)
</blockquote></details>

sounds good
___
*[`checkpoint_test.go` line 39 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2964#-NfqzCoREb9VTOmhedlT:-NfqzCoREb9VTOmhedlU:bxi79uf) ([raw file](https://github.com/cockroachdb/pebble/blob/b7826d90135f0fb0e6f551120ee699cc232697a9/checkpoint_test.go#L39)):*
> ```Go
> 		FormatMajorVersion:    internalFormatNewest,
> 		L0CompactionThreshold: 10,
> 		DisableAutomaticCompactions: true,
> ```

These were pretty annoying.

Basically the test shares a memLog for every single checkpoint and db. So async tasks can mess up the test output for commands like open, scan, etc.

I think we want to share the memLog, because we're sharing the filesystem, but we don't want unnecessary writes to show up. So I disabled these async tasks.
___
*[`checkpoint_test.go` line 150 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2964#-Nfr-dd67sL5Zn1YYRn9:-Nfr-dd67sL5Zn1YYRnA:b-al2qti) ([raw file](https://github.com/cockroachdb/pebble/blob/b7826d90135f0fb0e6f551120ee699cc232697a9/checkpoint_test.go#L150)):*
> ```Go
> 			return buf.String()
> 
> 		case "close":
> ```

closing a db associated with the checkpoint further ensures that random output from older checkpoints doesn't show up in test.

PR #2968, opened 2023-10-03
db: turn testingAlwaysWaitForCleanup into an option

bananabrick:
There are some potential racy problems by making it a variable which is written to after the db is opened. All of the existing use cases can just use this variable as an option.


bananabrick:
This was calling the CI to fail. I couldn't repro the failure locally even with the race flag enabled, but from what I can glean from the failure here: https://github.com/cockroachdb/pebble/actions/runs/6398528368/job/17368847209, the fix seems right.

RaduBerinde (approved):

bananabrick:
thanks

PR #2972, opened 2023-10-04
db: move l0SublevelInfo to the compactionLevel struct

bananabrick:
Don't necessarily think this is cleaner, but the compactionLevel already has a sublevel info which is used by the compaction struct, and this pr makes the compaction picker align with that.


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2972#-:-Ng1g7qy9hqJCPbp_8T6:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved

bananabrick:
Thanks.

PR #2979, opened 2023-10-05
db: convert adjustedOutputLevel into a function

bananabrick:
There's benefits to turning adjustedOutputLevel into a function.
1. Multilevel compactions will want to edit the adjustedOutputLevel, and currently does so in a hacky way. With this change, I can get rid of the hack.
2. Storing the baseLevel is much better because it doesn't change once set.


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2979#-:-Ng5jhbXB21_MH464pZu:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @bananabrick and @itsbilal)

___
*[`compaction_picker.go` line 268 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2979#-Ng5jPSpEgi2lw1TIvKP:-Ng5jPSpEgi2lw1TIvKQ:b-2nuquy) ([raw file](https://github.com/cockroachdb/pebble/blob/7e24231c4fc6eacf9c4156308d64fdfcca6b2209/compaction_picker.go#L268)):*
> ```Go
> // determining the target output file size, overlap bytes, and expanded
> // bytes, taking into account the base level.
> func adjustedOutputLevel(outputLevel int, baseLevel int) int {
> ```

[nit] I think this makes sense as a method on `pickedCompaction`

bananabrick (commented):
Thanks

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2979#-:-Ng5pBqy8JlvYk0WMl_J:b-zcjvte)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`compaction_picker.go` line 268 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2979#-Ng5jPSpEgi2lw1TIvKP:-Ng5p3cx2s0CjiaNWHQk:by3ggvp) ([raw file](https://github.com/cockroachdb/pebble/blob/7e24231c4fc6eacf9c4156308d64fdfcca6b2209/compaction_picker.go#L268)):*

[nit] I think this makes sense as a method on `pickedCompaction`
</blockquote></details>

I think it's slightly cleaner if the caller passes in the parameters rather than using the parameters in the pickedCompaction. Will leave this as is.

Issue #1115, opened 2021-04-19
db: better testing of error handling pathways


jbowens:
cockroachdb/cockroach#63847 is likely an example of improper error handling introducing corruption. As a part of the quality initiative, we should make an effort to audit and improve testing of error handling code paths.

We currently have some existing tests like [TestIngestError](https://github.com/cockroachdb/pebble/blob/108cd0260986c5972d20bd807a4d7192d912c97a/ingest_test.go#L581) that try injecting an error at every _i_-th operation. There are probably more code paths that deserve this testing treatment.

The metamorphic test has an [`-error-rate`](https://github.com/cockroachdb/pebble/blob/master/internal/metamorphic/meta_test.go#L53-L54) that will inject errors into _read-only_ operations and automatically retry them, but it currently defaults off. We can update it to default to some low rate.

Related to #270.

jbowens:
> The metamorphic test has an [-error-rate](https://github.com/cockroachdb/pebble/blob/master/internal/metamorphic/meta_test.go#L53-L54) that will inject errors into read-only operations and automatically retry them, but it currently defaults off. We can update it to default to some low rate.

^ This item is a quick win, so triaging as such. Can move to backlog afterwards.

jbowens:
Remaining work:

- We still need better integration with the metamorphic tests: The metamorphic tests should retry operations that fail with `ErrInjected`, but retrying relative positioning operations is a bit tricky.
- We don't have randomized testing of all write paths too. This has its own challenges since so many write paths are fatal if they error.

Issue #270, opened 2019-09-14
db: better handling of background errors


petermattis:
When a background error occurs during a flush or compaction, Pebble currently just logs the error and attempts to retry the operation indefinitely. Depending on the error which occurred, this isn't terribly helpful. If the error was due to lack of disk space, a retry loop can be helpful. But if the error is due to a logic bug in Pebble or some sort of corruption it would be better to place the the DB in a read-only mode. This is the strategy used by RocksDB.

Jira issue: PEBBLE-181

petermattis:
@rohansuri has volunteered to take a look at this issue.

rohansuri:
Overview of RocksDB's handling: https://github.com/facebook/rocksdb/wiki/Background-Error-Handling. I'll understand all the behaviours and put up the plan/any questions I have.

rohansuri:
@petermattis Does the scope of this also cover to not panic on manifest write failures and continue serving reads? That will require the following changes: https://github.com/facebook/rocksdb/issues/6316, https://github.com/facebook/rocksdb/pull/5379.

petermattis:
> @petermattis Does the scope of this also cover to not panic on manifest write failures and continue serving reads? That will require the following changes: facebook/rocksdb#6316, facebook/rocksdb#5379.

Yes, we'd want to be able to recover from out-of-disk errors during WAL writes, flushes/compactions, and MANIFEST writes.

sumeerbhola:
We also need more visible error reporting so that we can quickly find errors and not have to scan a very verbose Pebble log. This came up when discussing a CockroachDB postmortem.

PR #2982, opened 2023-10-06
db: tolerate errors during async ingest sstable validation

jbowens:
After a table is ingested, an asynchronous job reads the table and validates
its checksums. If corruption is uncovered, the job fatals the process.
Previously, a transient I/O error would also fatal the process, even if the I/O
error did not indicate corruption. This commit adapts this code path to log the
error to the BackgroundError event listener and re-queue the file for
validation.

Informs #270.
Informs #1115.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -2318,6 +2318,11 @@ func (d *DB) validateSSTables() {
 	// where we are starving IO from other tasks due to having to page through
 	// all the blocks in all the sstables in the queue.
 	// TODO(travers): Add some form of pacing to avoid IO starvation.
+
+	// If we fail to validate any files due to reasons other than uncovered
+	// corruption, accumulate them and re-queue them for another attempt.
+	var retry []manifest.NewFileEntry
+
 	for _, f := range pending {
 		// The file may have been moved or deleted since it was ingested, in
 		// which case we skip.
@@ -2353,9 +2358,24 @@ func (d *DB) validateSSTables() {
 		}
 
 		if err != nil {
-			// TODO(travers): Hook into the corruption reporting pipeline, once
-			// available. See pebble#1192.
-			d.opts.Logger.Fatalf("pebble: encountered corruption during ingestion: %s", err)
+			if IsCorruptionError(err) {
+				// TODO(travers): Hook into the corruption reporting pipeline, once
+				// available. See pebble#1192.
+				d.opts.Logger.Fatalf("pebble: encountered corruption during ingestion: %s", err)
+			} else {
+				// If there was some other, possibly transient, error that
+				// caused table validation to fail inform the EventListener and
+				// move on. We remember the table so that we can retry it in a
+				// subsequent table validation job.
+				//
+				// TODO(jackson): If the error is not transient, this will retry
+				// validation indefinitely. While not great, it's the same
+				// behavior as erroring flushes and compactions. We should
+				// address this as a part of #270.
+				d.opts.EventListener.BackgroundError(err)
+				retry = append(retry, f)
+				continue
+			}
 		}
 
 		d.opts.EventListener.TableValidated(TableValidatedInfo{
@@ -2364,9 +2384,9 @@ func (d *DB) validateSSTables() {
 		})
 	}
 	rs.unref()
-
 	d.mu.Lock()
 	defer d.mu.Unlock()
+	d.mu.tableValidation.pending = append(d.mu.tableValidation.pending, retry...)
 	d.mu.tableValidation.validating = false
 	d.mu.tableValidation.cond.Broadcast()
 	if d.shouldValidateSSTablesLocked() {
--- a/ingest_test.go
+++ b/ingest_test.go
@@ -11,7 +11,6 @@ import (
 	"io"
 	"math"
 	"os"
-	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
@@ -3169,7 +3168,6 @@ func TestIngestValidation(t *testing.T) {
 
 		ingestTableName = "ext"
 	)
-	ingestPath := filepath.Join(t.TempDir(), ingestTableName)
 
 	seed := uint64(time.Now().UnixNano())
 	rng := rand.New(rand.NewSource(seed))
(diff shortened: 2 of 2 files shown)


sumeerbhola (changes requested):
> Reviewed 1 of 1 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2982)* status: all files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`ingest_test.go` line 3163 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2982#-Ng5XGj1EHPFmsakJMEd:-Ng5XGj1EHPFmsakJMEe:b-35kvzk) ([raw file](https://github.com/cockroachdb/pebble/blob/7ca1c72e7b5f5720a90963edc5655f2c69131599/ingest_test.go#L3163)):*
> ```Go
> 		errLocationIngest
> 		errLocationValidation
> 		errLocationBackground
> ```

This could use a comment to distinguish it from `errLocationValidation`. It seems both are during validation, while one is a background error and does not cause a Fatalf.
Maybe call it `errLocationValidationBackgroundErr`.
___
*[`ingest_test.go` line 3217 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2982#-Ng5WiucCGy_g098Ujlj:-Ng5WiucCGy_g098Ujlk:b558pao) ([raw file](https://github.com/cockroachdb/pebble/blob/7ca1c72e7b5f5720a90963edc5655f2c69131599/ingest_test.go#L3217)):*
> ```Go
> 			wantErrType: errLocationBackground,
> 			errorfsInjector: errorfs.InjectorFunc(func(op errorfs.Op, path string) error {
> 				// Inject an error on the first read-at operation on an sstable.
> ```

Won't there be a call to `ReadAt` during the ingest to read certain blocks when we open an iterator to look for overlaps, which would make this errLocationIngest?
___
*[`ingest_test.go` line 3322 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2982#-Ng5W3GZ0bL9eVze4uyC:-Ng5W3GZ0bL9eVze4uyD:b-dpcaya) ([raw file](https://github.com/cockroachdb/pebble/blob/7ca1c72e7b5f5720a90963edc5655f2c69131599/ingest_test.go#L3322)):*
> ```Go
> 					f, err = fs.OpenReadWrite(ingestTableName)
> 					require.NoError(t, err)
> 					corrupt(f)
> ```

I am missing something -- why does corrupting before doing the ingest allow for corruption to be discovered in `errLocationValidation`? Won't it be discovered in ingest itself when opening the iterator and seeking?
Is this because this is always used with `corruptionLocationInternal`, so the relevant block does not get read during the ingest?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2982)* status: all files reviewed, 3 unresolved discussions (waiting on @sumeerbhola)

___
*[`ingest_test.go` line 3163 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2982#-Ng5XGj1EHPFmsakJMEd:-NgK8fZn4-IukpswrQkv:b-u4a9jy) ([raw file](https://github.com/cockroachdb/pebble/blob/7ca1c72e7b5f5720a90963edc5655f2c69131599/ingest_test.go#L3163)):*

This could use a comment to distinguish it from `errLocationValidation`. It seems both are during validation, while one is a background error and does not cause a Fatalf.
Maybe call it `errLocationValidationBackgroundErr`.
</blockquote></details>

I added some comments and did some renaming to make this clearer:

errLocation -> errReportLocation
errLocationValidation -> errReportLocationFatal
errLocationBackground -> errReportLocationBackgroundError
___
*[`ingest_test.go` line 3217 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2982#-Ng5WiucCGy_g098Ujlj:-NgK7kL71RMZPEx65goD:bibi056) ([raw file](https://github.com/cockroachdb/pebble/blob/7ca1c72e7b5f5720a90963edc5655f2c69131599/ingest_test.go#L3217)):*

Won't there be a call to `ReadAt` during the ingest to read certain blocks when we open an iterator to look for overlaps, which would make this errLocationIngest?
</blockquote></details>

This was subtle. This test creates to-be-ingested sstables at the path `ext` which doesn't have the `.sst` extension, and so are excluded from the error injection. I added an explicit `path != "ext"` condition and expanded the comment to make this clearer.
___
*[`ingest_test.go` line 3322 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2982#-Ng5W3GZ0bL9eVze4uyC:-NgKAHWlBw8mS-pQvOJa:b-8ofb7f) ([raw file](https://github.com/cockroachdb/pebble/blob/7ca1c72e7b5f5720a90963edc5655f2c69131599/ingest_test.go#L3322)):*

I am missing something -- why does corrupting before doing the ingest allow for corruption to be discovered in `errLocationValidation`? Won't it be discovered in ingest itself when opening the iterator and seeking?
Is this because this is always used with `corruptionLocationInternal`, so the relevant block does not get read during the ingest?
</blockquote></details>

That's right, `Ingest` won't read the corrupted blocks when used with `corruptionLocationInternal`. I added an explanatory comment at the `corruptionLocation` enum declaration.

sumeerbhola (approved):
:lgtm: 

> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2982#-:-NgL8JR2-mqKb0QS7VYg:bg7igkv)* status: all files reviewed, 4 unresolved discussions (waiting on @jbowens)

___
*[`ingest_test.go` line 3166 at r4](https://reviewable.io/reviews/cockroachdb/pebble/2982#-NgL7ot97fjlz3t1NNh7:-NgL7ot97fjlz3t1NNh8:bfey7ht) ([raw file](https://github.com/cockroachdb/pebble/blob/3fabe145657cd4258172c9cadcae6d5116d1408d/ingest_test.go#L3166)):*
> ```Go
> 	)
> 	// The errReportLocation type defines an enum to allow tests to enforce
> 	// expectations about how an error surfced during ingestion or validation is
> ```

nit: surfaced

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2982#-:-NgLOni32oKejYtXYpvS:b-814g8k)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @sumeerbhola)

PR #2984, opened 2023-10-07
metamorphic: Add TestMetaTwoInstance

### LINKED ISSUE #2711: metamorphic: Add two-instance variation to test disaggregated ingestion

itsbilal:
This change adds a two-instance metamorphic test as well as a new metamorphic test command, db.Replicate, that under some testOptions does a shared ingestion (i.e. ScanInternal with skip-shared iteration to only get keys above files in shared levels) along with an excise, and in other testOptions it does a normal ingestion with range deletes/rangekeydels to clear out the "replicated" range

Fixes #2711.

### DIFF
--- a/internal/metamorphic/meta_test.go
+++ b/internal/metamorphic/meta_test.go
@@ -70,3 +70,24 @@ func TestMeta(t *testing.T) {
 		metamorphic.RunAndCompare(t, runFlags.Dir, opts...)
 	}
 }
+
+func TestMetaTwoInstance(t *testing.T) {
+	switch {
+	case runOnceFlags.Compare != "":
+		runDirs := strings.Split(runOnceFlags.Compare, ",")
+		onceOpts := runOnceFlags.MakeRunOnceOptions()
+		metamorphic.Compare(t, runOnceFlags.Dir, runOnceFlags.Seed, runDirs, onceOpts...)
+
+	case runOnceFlags.RunDir != "":
+		// The --run-dir flag is specified either in the child process (see
+		// runOptions() below) or the user specified it manually in order to re-run
+		// a test.
+		onceOpts := runOnceFlags.MakeRunOnceOptions()
+		metamorphic.RunOnce(t, runOnceFlags.RunDir, runOnceFlags.Seed, filepath.Join(runOnceFlags.RunDir, "history"), onceOpts...)
+
+	default:
+		opts := runFlags.MakeRunOptions()
+		opts = append(opts, metamorphic.MultiInstance(2))
+		metamorphic.RunAndCompare(t, runFlags.Dir, opts...)
+	}
+}
--- a/internal/metamorphic/metaflags/meta_flags.go
+++ b/internal/metamorphic/metaflags/meta_flags.go
@@ -35,6 +35,9 @@ type CommonFlags struct {
 	Keep bool
 	// MaxThreads used by a single run. See "max-threads" flag below.
 	MaxThreads int
+	// NumInstances is the number of Pebble instances to create in one run. See
+	// "num-instances" flag below.
+	NumInstances int
 }
 
 func initCommonFlags() *CommonFlags {
@@ -64,6 +67,8 @@ func initCommonFlags() *CommonFlags {
 
 	flag.IntVar(&c.MaxThreads, "max-threads", math.MaxInt,
 		"limit execution of a single run to the provided number of threads; must be ≥ 1")
+
+	flag.IntVar(&c.NumInstances, "num-instances", 1, "number of pebble instances to create (default: 1)")
 
 	return c
 }
@@ -179,6 +184,9 @@ func (ro *RunOnceFlags) MakeRunOnceOptions() []metamorphic.RunOnceOption {
 	if ro.ErrorRate > 0 {
 		onceOpts = append(onceOpts, metamorphic.InjectErrorsRate(ro.ErrorRate))
 	}
+	if ro.NumInstances > 1 {
+		onceOpts = append(onceOpts, metamorphic.MultiInstance(ro.NumInstances))
+	}
 	return onceOpts
 }
 
@@ -203,6 +211,9 @@ func (r *RunFlags) MakeRunOptions() []metamorphic.RunOption {
 	}
 	if r.PreviousOps != "" {
 		opts = append(opts, metamorphic.ExtendPreviousRun(r.PreviousOps, r.InitialStatePath, r.InitialStateDesc))
+	}
+	if r.NumInstances > 1 {
+		opts = append(opts, metamorphic.MultiInstance(r.NumInstances))
 	}
 
 	// If the filesystem type was forced, all tests will use that value.
--- a/metamorphic/config.go
+++ b/metamorphic/config.go
@@ -38,6 +38,7 @@ const (
 	newIterUsingClone
 	newSnapshot
 	readerGet
+	replicate
 	snapshotClose
 	writerApply
 	writerDelete
@@ -64,6 +65,10 @@ type config struct {
 	// It's a dynamic randvar to roughly emulate workloads with MVCC timestamps,
 	// skewing towards most recent timestamps.
 	writeSuffixDist randvar.Dynamic
+
+	// numInstances defines the number of pebble instances created for this
+	// metamorphic test run.
+	numInstances int
 
 	// TODO(peter): unimplemented
 	// keyDist        randvar.Dynamic
@@ -107,6 +112,8 @@ var presetConfigs = []config{
 		withOpWeight(writerMerge, 0),
 }
 
+var multiInstancePresetConfig = multiInstanceConfig()
+
 func defaultConfig() config {
 	return config{
 		// dbClose is not in this list since it is deterministically generated once, at the end of the test.
@@ -139,6 +146,7 @@ func defaultConfig() config {
 			newIterUsingClone:           5,
 			newSnapshot:                 10,
 			readerGet:                   100,
+			replicate:                   0,
 			snapshotClose:               10,
 			writerApply:                 10,
 			writerDelete:                100,
@@ -159,6 +167,19 @@ func defaultConfig() config {
 		// for a particular prefix.
 		writeSuffixDist: mustDynamic(randvar.NewSkewedLatest(0, 1, 0.99)),
 	}
+}
+
+func multiInstanceConfig() config {
+	cfg := defaultConfig()
+	cfg.ops[replicate] = 5
+	cfg.ops[writerSingleDelete] = 0
+	cfg.ops[writerMerge] = 0
+	// TODO(bilal): The disabled operations below should also be supported
+	// in the two-instance test, once they're updated to work in multi-instance
+	// mode.
+	cfg.ops[newSnapshot] = 0
+	cfg.ops[snapshotClose] = 0
+	return cfg
 }
 
 func mustDynamic(dyn randvar.Dynamic, err error) randvar.Dynamic {
--- a/metamorphic/generator.go
+++ b/metamorphic/generator.go
@@ -56,6 +56,7 @@ type generator struct {
 
 	// keyManager tracks the state of keys a operation generation time.
 	keyManager *keyManager
+	dbs        objIDSlice
 	// Unordered sets of object IDs for live objects. Used to randomly select on
 	// object when generating an operation. There are 4 concrete objects: the DB
 	// (of which there is exactly 1), batches, iterators, and snapshots.
@@ -83,6 +84,9 @@ type generator struct {
 	// iterators. The iter set value will also be indexed by either the batches
 	// or snapshots maps.
 	iters map[objID]objIDSet
+	// objectID -> db: used to keep track of the DB a batch, iter, or snapshot
+	// was created on.
+	objDB map[objID]objID
 	// readerID -> reader iters: used to keep track of the open iterators on a
 	// reader. The iter set value will also be indexed by either the batches or
 	// snapshots maps. This map is the union of batches and snapshots maps.
@@ -103,10 +107,12 @@ func newGenerator(rng *rand.Rand, cfg config, km *keyManager) *generator {
 	g := &generator{
 		cfg:                   cfg,
 		rng:                   rng,
-		init:                  &initOp{},
+		init:                  &initOp{dbSlots: uint32(cfg.numInstances)},
 		keyManager:            km,
-		liveReaders:           objIDSlice{makeObjID(dbTag, 0)},
-		liveWriters:           objIDSlice{makeObjID(dbTag, 0)},
+		liveReaders:           objIDSlice{makeObjID(dbTag, 1)},
+		liveWriters:           objIDSlice{makeObjID(dbTag, 1)},
+		dbs:                   objIDSlice{makeObjID(dbTag, 1)},
+		objDB:                 make(map[objID]objID),
 		batches:               make(map[objID]objIDSet),
 		iters:                 make(map[objID]objIDSet),
 		readers:               make(map[objID]objIDSet),
@@ -115,6 +121,11 @@ func newGenerator(rng *rand.Rand, cfg config, km *keyManager) *generator {
 		itersLastOpts:         make(map[objID]iterOpts),
 		iterCreationTimestamp: make(map[objID]int),
 		iterReaderID:          make(map[objID]objID),
+	}
+	for i := 1; i < cfg.numInstances; i++ {
+		g.liveReaders = append(g.liveReaders, makeObjID(dbTag, uint32(i+1)))
+		g.liveWriters = append(g.liveWriters, makeObjID(dbTag, uint32(i+1)))
+		g.dbs = append(g.dbs, makeObjID(dbTag, uint32(i+1)))
 	}
 	// Note that the initOp fields are populated during generation.
 	g.ops = append(g.ops, g.init)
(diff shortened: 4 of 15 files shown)


itsbilal:
This ended up being a much bigger lift than I had hoped at first, but it achieves the primary objectives. I don't expect this to be in merge-ready state right now, but it's close; I expect to address some of the TODOs very soon, either imminently or in follow-up PRs (eg. I can see the work to enable snapshots in multi-instance mode being a followup PR).

Either way, this should be ready for a review!

itsbilal:
I've addressed all the TODOs that I had hoped to address in this PR. I expect the rest to be related to either checkpointing (which is disabled for shared storage anyway), or snapshots, and I believe a follow-up PR is best to address snapshots/EFOS.

jbowens (commented):
> Reviewed 4 of 14 files at r1, 1 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2984)* status: 5 of 14 files reviewed, 3 unresolved discussions (waiting on @itsbilal)

___
*[`metamorphic/generator.go` line 1176 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhDjYN5BDBXj04cBos-:-NhDjYN5BDBXj04cBos0:bn48c11) ([raw file](https://github.com/cockroachdb/pebble/blob/0c91aa23a86c2bef2d0e76a7a8bf735f743d4180/metamorphic/generator.go#L1176)):*
> ```Go
> 		start:  startKey,
> 		end:    endKey,
> 	})
> ```

Can we adjust the generator / parser tests to exercise these new code paths?
___
*[`metamorphic/ops.go` line 472 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhDl5U00RBeDC9vD0zE:-NhDl5U00RBeDC9vD0zF:b72vg1j) ([raw file](https://github.com/cockroachdb/pebble/blob/0c91aa23a86c2bef2d0e76a7a8bf735f743d4180/metamorphic/ops.go#L472)):*
> ```Go
> 	// a single batch. If we are ingesting multiple batches, the batches may
> 	// overlap which would cause ingestion to fail but apply would succeed.
> 	if t.testOpts.ingestUsingApply && len(o.batchIDs) == 1 && o.derivedDBID == o.dbID {
> ```

Can you explain this bit? How would `derivedDBID != o.dbID`?
___
*[`metamorphic/ops.go` line 572 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhDlcEyEQjgckSkkO8N:-NhDlcEyEQjgckSkkO8O:b-5m8gdc) ([raw file](https://github.com/cockroachdb/pebble/blob/0c91aa23a86c2bef2d0e76a7a8bf735f743d4180/metamorphic/ops.go#L572)):*
> ```Go
> 	//if err := f.Sync(); err != nil {
> 	//	return "", err
> 	//}
> ```

debugging detritus?

itsbilal (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2984)* status: 5 of 14 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`metamorphic/generator.go` line 1176 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhDjYN5BDBXj04cBos-:-NhSVxk1As0gXzqCdjI7:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0c91aa23a86c2bef2d0e76a7a8bf735f743d4180/metamorphic/generator.go#L1176)):*

Can we adjust the generator / parser tests to exercise these new code paths?
</blockquote></details>

Done.
___
*[`metamorphic/ops.go` line 472 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhDl5U00RBeDC9vD0zE:-NhTLb_e4j5f010UYOd5:b-rd4s8d) ([raw file](https://github.com/cockroachdb/pebble/blob/0c91aa23a86c2bef2d0e76a7a8bf735f743d4180/metamorphic/ops.go#L472)):*

Can you explain this bit? How would `derivedDBID != o.dbID`?
</blockquote></details>

derivedDBID is the ID of the DB the batch was created from, and dbID is the DB we're ingesting this batch into. If ingestUsingApply is true, the two DBs must match or else we can't do an ingestUsingApply (as it'll error out due to us applying a batch from a different DB).
___
*[`metamorphic/ops.go` line 572 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhDlcEyEQjgckSkkO8N:-NhTLrjX3Gf6VyMF7ZWZ:bvu6wiw) ([raw file](https://github.com/cockroachdb/pebble/blob/0c91aa23a86c2bef2d0e76a7a8bf735f743d4180/metamorphic/ops.go#L572)):*

debugging detritus?
</blockquote></details>

Oops. Removed.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2984)* status: 3 of 15 files reviewed, 9 unresolved discussions (waiting on @itsbilal)

___
*[`metamorphic/generator.go` line 1176 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhDjYN5BDBXj04cBos-:-NhXD9aEC8-enBgRogc1:beme8iu) ([raw file](https://github.com/cockroachdb/pebble/blob/0c91aa23a86c2bef2d0e76a7a8bf735f743d4180/metamorphic/generator.go#L1176)):*

Done.
</blockquote></details>

Looks like it's still not exercising the innards of this func because `len(g.dbs) < 2`
___
*[`metamorphic/generator.go` line 690 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhXDdQW1oT0K5iP8TgU:-NhXDdQW1oT0K5iP8TgV:b-ya7hnm) ([raw file](https://github.com/cockroachdb/pebble/blob/ff66c8146d1580e91f21c556240914376c17ae2f/metamorphic/generator.go#L690)):*
> ```Go
> 	if dbReaderID.tag() != dbTag {
> 		dbReaderID = g.objDB[dbReaderID]
> 	}
> ```

This seems like a common pattern throughout the generator. Should we have a `deriveDB` helper?
___
*[`metamorphic/meta.go` line 479 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhXEJCUBNyPsFUI_f_m:-NhXEJCUBNyPsFUI_f_n:b-k4co1u) ([raw file](https://github.com/cockroachdb/pebble/blob/ff66c8146d1580e91f21c556240914376c17ae2f/metamorphic/meta.go#L479)):*
> ```Go
> 	if opts.WALDir != "" {
> 		if runOpts.numInstances > 1 {
> 			opts.WALDir = ""
> ```

we could add the DB's prefix to the WAL dir and support this option in multi-instance runs, yeah?
___
*[`metamorphic/ops.go` line 472 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhDl5U00RBeDC9vD0zE:-NhXFJos06mcmF9dtrrB:b5b4q2d) ([raw file](https://github.com/cockroachdb/pebble/blob/0c91aa23a86c2bef2d0e76a7a8bf735f743d4180/metamorphic/ops.go#L472)):*

derivedDBID is the ID of the DB the batch was created from, and dbID is the DB we're ingesting this batch into. If ingestUsingApply is true, the two DBs must match or else we can't do an ingestUsingApply (as it'll error out due to us applying a batch from a different DB).
</blockquote></details>

Thanks. I suppose we could also use a `BatchReader` to copy its contents to a new `Batch` on the target DB (performing the same `DELSIZED`->`DEL` translation as necessary).
___
*[`metamorphic/ops.go` line 178 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhXF8Te0rD5GiX_0MQo:-NhXF8Te0rD5GiX_0MQp:bscbyny) ([raw file](https://github.com/cockroachdb/pebble/blob/ff66c8146d1580e91f21c556240914376c17ae2f/metamorphic/ops.go#L178)):*
> ```Go
> 		return []objID{o.derivedDBID}
> 	}
> 	return []objID{dbObjID}
> ```

is it much of a lift to remove `dbObjID` now? It's a little confusing in this new world of multiple instances.
___
*[`metamorphic/parser.go` line 548 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhXILKRDHuduUGV5jpo:-NhXILKRDHuduUGV5jpp:b5qmqls) ([raw file](https://github.com/cockroachdb/pebble/blob/ff66c8146d1580e91f21c556240914376c17ae2f/metamorphic/parser.go#L548)):*
> ```Go
> 		case *getOp:
> 			if derivedDBID, ok := objToDB[v.readerID]; ok {
> 				v.derivedDBID = derivedDBID
> ```

nit: do these need to be conditional? if we seed `objToDB` to include every DB mapping to itself, I think this could just be `v.derivedDBID = objToDB[v.readerID]`?
___
*[`metamorphic/parser.go` line 557 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhXH2knDCxn5-ofU6UE:-NhXH2knDCxn5-ofU6UF:begmplv) ([raw file](https://github.com/cockroachdb/pebble/blob/ff66c8146d1580e91f21c556240914376c17ae2f/metamorphic/parser.go#L557)):*
> ```Go
> 			}
> 		case *ingestOp:
> 			if len(v.batchIDs) == 1 {
> ```

I don't think it's possible for `len(v.batchIDs) == 0`. If `len(v.batchIDs) >1`, we don't need this because we won't ingest-using-apply? I don't love the subtlety, since this code is so far from the ingest-using-apply code path with the dependency. Should we derive the DB for every batch, even if we're not using it at the moment?
___
*[`metamorphic/test.go` line 403 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhXIlnk9x55RHRF7aVt:-NhXIlnk9x55RHRF7aVu:bwrbfg9) ([raw file](https://github.com/cockroachdb/pebble/blob/ff66c8146d1580e91f21c556240914376c17ae2f/metamorphic/test.go#L403)):*
> ```Go
> 		waitIndex, ok := lastOpReference[receiver]
> 		lastOpReference[receiver] = i
> 		if !ok && receiver.tag() != dbTag {
> ```

I think this change makes this if block dead code. I think even if `receiver.tag() == dbTag`, we want to enter this block. Down below though, we only want to panic `if receiver.tag() != dbTag`.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2984#-:-Nhmptmv3KyVI0Vdzqlu:b-814g8k)* status: 3 of 15 files reviewed, 9 unresolved discussions (waiting on @jbowens)

___
*[`metamorphic/generator.go` line 1176 at r2](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhDjYN5BDBXj04cBos-:-Nhmdtnu09YySGwQrFER:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0c91aa23a86c2bef2d0e76a7a8bf735f743d4180/metamorphic/generator.go#L1176)):*

Looks like it's still not exercising the innards of this func because `len(g.dbs) < 2`
</blockquote></details>

Done.
___
*[`metamorphic/generator.go` line 690 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhXDdQW1oT0K5iP8TgU:-Nhmadqn4t9hsjHw9BMi:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/ff66c8146d1580e91f21c556240914376c17ae2f/metamorphic/generator.go#L690)):*

This seems like a common pattern throughout the generator. Should we have a `deriveDB` helper?
</blockquote></details>

Done.
___
*[`metamorphic/meta.go` line 479 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhXEJCUBNyPsFUI_f_m:-NhmeHfw7Si_cb2cjhoL:br2bn5m) ([raw file](https://github.com/cockroachdb/pebble/blob/ff66c8146d1580e91f21c556240914376c17ae2f/metamorphic/meta.go#L479)):*

we could add the DB's prefix to the WAL dir and support this option in multi-instance runs, yeah?
</blockquote></details>

Yeah that's correct. I've left a TODO about this so it isn't lost for now.
___
*[`metamorphic/ops.go` line 178 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhXF8Te0rD5GiX_0MQo:-NhmeKMz78KXxB3aJwJB:brmflxb) ([raw file](https://github.com/cockroachdb/pebble/blob/ff66c8146d1580e91f21c556240914376c17ae2f/metamorphic/ops.go#L178)):*

is it much of a lift to remove `dbObjID` now? It's a little confusing in this new world of multiple instances.
</blockquote></details>

I've removed some of the ones where we were unnecessarily syncing on it, such as this case. The remaining ones are now specific to the operations that haven't been converted over (eg. `Snapshot`).
___
*[`metamorphic/parser.go` line 548 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhXILKRDHuduUGV5jpo:-Nhmj0V97a9bV9TRTtV5:b-izvfud) ([raw file](https://github.com/cockroachdb/pebble/blob/ff66c8146d1580e91f21c556240914376c17ae2f/metamorphic/parser.go#L548)):*

nit: do these need to be conditional? if we seed `objToDB` to include every DB mapping to itself, I think this could just be `v.derivedDBID = objToDB[v.readerID]`?
</blockquote></details>

The annoying thing is that the parser doesn't know how many DBs there are, at this point. I could do this but it'll be a bit more of a lift.
___
*[`metamorphic/parser.go` line 557 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhXH2knDCxn5-ofU6UE:-Nhmj-zEBeCOWEJXCJl_:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/ff66c8146d1580e91f21c556240914376c17ae2f/metamorphic/parser.go#L557)):*

I don't think it's possible for `len(v.batchIDs) == 0`. If `len(v.batchIDs) >1`, we don't need this because we won't ingest-using-apply? I don't love the subtlety, since this code is so far from the ingest-using-apply code path with the dependency. Should we derive the DB for every batch, even if we're not using it at the moment?
</blockquote></details>

Done.
___
*[`metamorphic/test.go` line 403 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhXIlnk9x55RHRF7aVt:-NhmjtK53cBbs9OMT3Qt:b-cifnwt) ([raw file](https://github.com/cockroachdb/pebble/blob/ff66c8146d1580e91f21c556240914376c17ae2f/metamorphic/test.go#L403)):*

I think this change makes this if block dead code. I think even if `receiver.tag() == dbTag`, we want to enter this block. Down below though, we only want to panic `if receiver.tag() != dbTag`.
</blockquote></details>

Done. I had to do something funky here to have `initOp` sync on all DBs, and have operations on `db2` treated correctly in this code, as otherwise we'd have the initOp fall through the continue case and we wouldn't treat operations on `db2` as waiting on the init operation either. PTAL.

jbowens (approved):
:lgtm:

> Reviewed 4 of 7 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2984#-:-Ni0x666DiBbxujFgjSj:bnfp4nl)* status: 7 of 15 files reviewed, 4 unresolved discussions (waiting on @itsbilal)

___
*[`metamorphic/meta.go` line 479 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhXEJCUBNyPsFUI_f_m:-Ni0wU2s-3KuU2pPAdeI:b-zcjvte) ([raw file](https://github.com/cockroachdb/pebble/blob/ff66c8146d1580e91f21c556240914376c17ae2f/metamorphic/meta.go#L479)):*

Yeah that's correct. I've left a TODO about this so it isn't lost for now.
</blockquote></details>

Thanks
___
*[`metamorphic/parser.go` line 548 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2984#-NhXILKRDHuduUGV5jpo:-Ni0wlrwE46RuBQ1mMr0:bjt4j4z) ([raw file](https://github.com/cockroachdb/pebble/blob/ff66c8146d1580e91f21c556240914376c17ae2f/metamorphic/parser.go#L548)):*

The annoying thing is that the parser doesn't know how many DBs there are, at this point. I could do this but it'll be a bit more of a lift.
</blockquote></details>

I see. I think it's probably worthwhile, but fine as a followup.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2984#-:-Ni16DDiBjEEoppB8com:b-814g8k)* status: 7 of 15 files reviewed, 4 unresolved discussions (waiting on @jbowens)

PR #2987, opened 2023-10-09
go.mod: revert change to require go 1.21

RaduBerinde:
Setting the version to 1.21 forces cockroach to require 1.21 as well; we shouldn't do this until cockroach moves to 1.21.


jbowens (approved):
lgtm

RaduBerinde:
Thanks @jbowens - do you think you can do the bump on cockroach? I'm afk for 1-2 hrs

PR #2989, opened 2023-10-09
db: return structured errors from checkConsistency

jbowens:
During Open, Pebble verifies the consistency between the Version and the object catalog. Previously if this function encountered an error or inconsistency, it would return a new error that erased the type(s) of the original errors. This commit modifies the function to use errors.Join and avoid the type erasure. This is necessary for the metamorphic test's error injection, which must be able to test an error to observe whether it is an errorfs.ErrInjected error.

Informs #1115.

### DIFF
--- a/open.go
+++ b/open.go
@@ -1122,9 +1122,7 @@ func IsCorruptionError(err error) bool {
 }
 
 func checkConsistency(v *manifest.Version, dirname string, objProvider objstorage.Provider) error {
-	var buf bytes.Buffer
-	var args []interface{}
-
+	var errs []error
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

Very cool!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2989#-:-NgLoF32Dle5jRb3olMH:btgo8jj)* status: 0 of 1 files reviewed, all discussions resolved

jbowens:
TFTR!

Issue #2994, opened 2023-10-11
db: iterator creation is fallible


jbowens:
Relates to #1115.

The pebble Iterator is generally designed to tolerate errors. If an iterator operation errors, resulting in `Iterator.Error()` returning a non-nil error, the expectation is that the user may seek the `Iterator` again, and the previous error is cleared. There's at least 1 violation of this contract in which an I/O error encountered during iterator construction results in an `Iterator` that will always error.

If L0 contains any range keys, each file in L0 containing the range keys is opened during iterator construction. If opening the file encounters an error, then a keyspan.Iter that always errors is used, preventing any `Iterator` operation from succeeding: 
https://github.com/cockroachdb/pebble/blob/36b3f57b6167eb4e25d29dcb7be976d603f47d51/range_keys.go#L73

We should not perform any I/O during the iterator creation itself.

PR #2995, opened 2023-10-11
vfs/errorfs: add facilities for error injection in datadriven tests

jbowens:
We frequently use the errorfs package to inject errors into filesystem operations to test error code paths. This commit builds off this package to support encoding error injection conditions within datadriven tests through parsing a small DSL. Future work will build off this to test handling of I/O errors during iteration.

Informs #1115.
Informs #2994.

### DIFF
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -2900,7 +2900,7 @@ func TestCompactionErrorCleanup(t *testing.T) {
 	mem := vfs.NewMem()
 	ii := errorfs.OnIndex(math.MaxInt32) // start disabled
 	opts := (&Options{
-		FS:     errorfs.Wrap(mem, ii),
+		FS:     errorfs.Wrap(mem, errorfs.ErrInjected.If(ii)),
 		Levels: make([]LevelOptions, numLevels),
 		EventListener: &EventListener{
 			TableCreated: func(info TableCreateInfo) {
@@ -3286,8 +3286,8 @@ func TestFlushError(t *testing.T) {
 	// Error the first five times we try to write a sstable.
 	var errorOps atomic.Int32
 	errorOps.Store(3)
-	fs := errorfs.Wrap(vfs.NewMem(), errorfs.InjectorFunc(func(op errorfs.Op, path string) error {
-		if op == errorfs.OpCreate && filepath.Ext(path) == ".sst" && errorOps.Add(-1) >= 0 {
+	fs := errorfs.Wrap(vfs.NewMem(), errorfs.InjectorFunc(func(op errorfs.Op) error {
+		if op.Kind == errorfs.OpCreate && filepath.Ext(op.Path) == ".sst" && errorOps.Add(-1) >= 0 {
 			return errorfs.ErrInjected
 		}
 		return nil
@@ -3704,19 +3704,24 @@ type createManifestErrorInjector struct {
 	enabled atomic.Bool
 }
 
+// TODO(jackson): Replace the createManifestErrorInjector with the composition
+// of primitives defined in errorfs. This may require additional primitives.
+
+func (i *createManifestErrorInjector) String() string { return "MANIFEST-Creates" }
+
 // enable enables error injection for the vfs.FS.
 func (i *createManifestErrorInjector) enable() {
 	i.enabled.Store(true)
 }
 
 // MaybeError implements errorfs.Injector.
-func (i *createManifestErrorInjector) MaybeError(op errorfs.Op, path string) error {
+func (i *createManifestErrorInjector) MaybeError(op errorfs.Op) error {
 	if !i.enabled.Load() {
 		return nil
 	}
 	// This necessitates having a MaxManifestSize of 1, to reliably induce
 	// logAndApply errors.
-	if strings.Contains(path, "MANIFEST") && op == errorfs.OpCreate {
+	if strings.Contains(op.Path, "MANIFEST") && op.Kind == errorfs.OpCreate {
 		return errorfs.ErrInjected
 	}
 	return nil
@@ -3884,6 +3889,11 @@ type WriteErrorInjector struct {
 	enabled atomic.Bool
 }
 
+// TODO(jackson): Replace WriteErrorInjector with use of primitives in errorfs,
+// adding new primitives as necessary.
+
+func (i *WriteErrorInjector) String() string { return "FileWrites(ErrInjected)" }
+
 // enable enables error injection for the vfs.FS.
 func (i *WriteErrorInjector) enable() {
 	i.enabled.Store(true)
@@ -3895,12 +3905,12 @@ func (i *WriteErrorInjector) disable() {
 }
 
 // MaybeError implements errorfs.Injector.
-func (i *WriteErrorInjector) MaybeError(op errorfs.Op, path string) error {
+func (i *WriteErrorInjector) MaybeError(op errorfs.Op) error {
 	if !i.enabled.Load() {
 		return nil
 	}
 	// Fail any future write.
-	if op == errorfs.OpFileWrite {
+	if op.Kind == errorfs.OpFileWrite {
 		return errorfs.ErrInjected
 	}
 	return nil
--- a/data_test.go
+++ b/data_test.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/cockroachdb/datadriven"
 	"github.com/cockroachdb/errors"
+	"github.com/cockroachdb/pebble/bloom"
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/humanize"
 	"github.com/cockroachdb/pebble/internal/keyspan"
@@ -29,6 +30,7 @@ import (
 	"github.com/cockroachdb/pebble/objstorage/remote"
 	"github.com/cockroachdb/pebble/sstable"
 	"github.com/cockroachdb/pebble/vfs"
+	"github.com/cockroachdb/pebble/vfs/errorfs"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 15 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2995)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`vfs/errorfs/errorfs.go` line 202 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2995#-Ng_2muD3gYseqYyKZ6b:-Ng_LuVvDXqcvXwkTKod:bt4yzb1) ([raw file](https://github.com/cockroachdb/pebble/blob/affd954686276d0beceb69267c64be00632814cf/vfs/errorfs/errorfs.go#L202)):*
> This would be cleaner and more powerful, e.g. just by swapping the `all` arguments you can switch from "error on the 2nd operation but only if it happens to be a read" to "error on the 2nd read"

Both are possible as-is with `onIndex(2, reads(always))` and `reads(onIndex(2, always))`, but I agree the `all(...)` syntax is a little easier on the eyes.

I think the tradeoff is today an injector may inject arbitrary errors, not just `ErrInjected`. This isn't something we take advantage of yet, but we could use it to test branches gated on particular errors. Updated with a change that replaces `always` with `ErrInjected`. Thoughts?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2995)* status: 0 of 15 files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`vfs/errorfs/errorfs.go` line 202 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2995#-Ng_2muD3gYseqYyKZ6b:-Ng_s5lL6oswKFC5QrTN:b-a27rny) ([raw file](https://github.com/cockroachdb/pebble/blob/affd954686276d0beceb69267c64be00632814cf/vfs/errorfs/errorfs.go#L202)):*

> This would be cleaner and more powerful, e.g. just by swapping the `all` arguments you can switch from "error on the 2nd operation but only if it happens to be a read" to "error on the 2nd read"

Both are possible as-is with `onIndex(2, reads(always))` and `reads(onIndex(2, always))`, but I agree the `all(...)` syntax is a little easier on the eyes.

I think the tradeoff is today an injector may inject arbitrary errors, not just `ErrInjected`. This isn't something we take advantage of yet, but we could use it to test branches gated on particular errors. Updated with a change that replaces `always` with `ErrInjected`. Thoughts?
</blockquote></details>

Ah, indeed.

I think the same is achievable by defining `all()` as returning the error of the last injector (as long as all of them returned an error).

I kind of liked `always`. Maybe it can take an optional parameter, e.g. `always()` is equivalent to `always(ErrInjected)` but in the future it could also be `always(ErrSomethingElse)`.

sumeerbhola (dismissed):
nice! :lgtm: once @RaduBerinde approves

> Reviewed 2 of 9 files at r1, 4 of 9 files at r2, 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2995#-:-NgdKDLr1fAykJqBf4-3:b-wjsqvv)* status: all files reviewed, 6 unresolved discussions (waiting on @jbowens)

___
*[`vfs/errorfs/dsl.go` line 62 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2995#-NgdGKEcAu2bB8ayhSWw:-NgdGKEcAu2bB8ayhSWx:bahhz1k) ([raw file](https://github.com/cockroachdb/pebble/blob/317b29e777d99d9df8108e4ce7224c4185a04d0f/vfs/errorfs/dsl.go#L62)):*
> ```Go
> 	label string
> }
> ```

so calling `Error()` will not include the `label` but calling `String()` only returns the label and nothing from the `error`.
It looks a bit odd, and maybe could use a code comment.
___
*[`vfs/errorfs/errorfs.go` line 132 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2995#-NgdIR99E50AJWHKzXHs:-NgdIR99E50AJWHKzXHt:bmayx3s) ([raw file](https://github.com/cockroachdb/pebble/blob/317b29e777d99d9df8108e4ce7224c4185a04d0f/vfs/errorfs/errorfs.go#L132)):*
> ```Go
> type opKindInj struct {
> 	kind OpReadWrite
> 	next Injector
> ```

just curious: why is this called `next`? Is the mental model to think of this wrapping as a chain?
___
*[`vfs/errorfs/errorfs.go` line 184 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2995#-NgdJ2JU1Xsl-Qq-z7ZR:-NgdJ2JU1Xsl-Qq-z7ZS:bx1ysl9) ([raw file](https://github.com/cockroachdb/pebble/blob/317b29e777d99d9df8108e4ce7224c4185a04d0f/vfs/errorfs/errorfs.go#L184)):*
> ```Go
> // into an ErrFS with the provided probability. p should be within the range
> // [0.0,1.0].
> func WithProbability(op OpReadWrite, p float64) Injector {
> ```

didn't need support for this in the DSL?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2995)* status: 10 of 15 files reviewed, 5 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`vfs/errorfs/dsl.go` line 62 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2995#-NgdGKEcAu2bB8ayhSWw:-Ngea5uo8T7tyIYS4C3x:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/317b29e777d99d9df8108e4ce7224c4185a04d0f/vfs/errorfs/dsl.go#L62)):*

so calling `Error()` will not include the `label` but calling `String()` only returns the label and nothing from the `error`.
It looks a bit odd, and maybe could use a code comment.
</blockquote></details>

Done.
___
*[`vfs/errorfs/errorfs.go` line 202 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2995#-Ng_2muD3gYseqYyKZ6b:-Ngempp6-R9t3EemGLsZ:b-6yr9uc) ([raw file](https://github.com/cockroachdb/pebble/blob/affd954686276d0beceb69267c64be00632814cf/vfs/errorfs/errorfs.go#L202)):*

Ah, indeed.

I think the same is achievable by defining `all()` as returning the error of the last injector (as long as all of them returned an error).

I kind of liked `always`. Maybe it can take an optional parameter, e.g. `always()` is equivalent to `always(ErrInjected)` but in the future it could also be `always(ErrSomethingElse)`.
</blockquote></details>

I think the semantics of `all` become a little weird and subtle if we're discarding the `error` value of all the prior calls. We could expand the DSL to separate the predicates and error types... And straight up lean into the lisp Eg:

```
(ErrInjected [PREDICATE])
(Or <PREDICATE> [PREDICATE]...)
(And <PREDICATE> [PREDICATE]...)
(PathMatch <STRING>)
(OnIndex <INTEGER>)
(Any <OPTION<ERROR>> [OPTION<ERROR>]...)
```

Some examples:
```
(ErrInjected (And (OnIndex 3) Reads))
(ErrInjected (And Reads (OnIndex 3)))
(Any (ENOSPC Writes) (ErrInjected (And Reads (OnIndex 2))))
```

Overboard?
___
*[`vfs/errorfs/errorfs.go` line 132 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2995#-NgdIR99E50AJWHKzXHs:-Ngea9CECQHlz1ybSP4J:b5vdn1k) ([raw file](https://github.com/cockroachdb/pebble/blob/317b29e777d99d9df8108e4ce7224c4185a04d0f/vfs/errorfs/errorfs.go#L132)):*

just curious: why is this called `next`? Is the mental model to think of this wrapping as a chain?
</blockquote></details>

yeah, I was thinking of it similar to the `net/http` `Handler` types that wrap another `Handler`, sometimes deferring to the inner `Handler`.
___
*[`vfs/errorfs/errorfs.go` line 184 at r3](https://reviewable.io/reviews/cockroachdb/pebble/2995#-NgdJ2JU1Xsl-Qq-z7ZR:-NgeabTR2NS1KcCmRflu:b-jy3elm) ([raw file](https://github.com/cockroachdb/pebble/blob/317b29e777d99d9df8108e4ce7224c4185a04d0f/vfs/errorfs/errorfs.go#L184)):*

didn't need support for this in the DSL?
</blockquote></details>

I suspect it won't be useful within the datadriven tests, since it uses external randomness that will change the results on every run
___
*[`vfs/errorfs/errorfs_test.go` line 18 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2995#-Ng_3ghv6jrGlDS6eC5J:-Ngen2wz4trd4oCwG6iE:b-9tdqsz) ([raw file](https://github.com/cockroachdb/pebble/blob/affd954686276d0beceb69267c64be00632814cf/vfs/errorfs/errorfs_test.go#L18)):*

I am biased from optgen but the lisp syntax worked very well there and was easy to parse.
eg to fail the second read and all writes we'd do `(Any (All (OnAllReads) (OnIndex 2)) (OnAllWrites))`

I don't immediately see a benefit here though just mentioning it as a potential suggestion.
</blockquote></details>

I think this is a little more legible, so I adopted it.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2995)* status: 6 of 15 files reviewed, 5 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`vfs/errorfs/errorfs.go` line 202 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2995#-Ng_2muD3gYseqYyKZ6b:-NgsyFb33EUZ2ofcBhux:b-hurg2o) ([raw file](https://github.com/cockroachdb/pebble/blob/affd954686276d0beceb69267c64be00632814cf/vfs/errorfs/errorfs.go#L202)):*

I think the semantics of `all` become a little weird and subtle if we're discarding the `error` value of all the prior calls. We could expand the DSL to separate the predicates and error types... And straight up lean into the lisp Eg:

```
(ErrInjected [PREDICATE])
(Or <PREDICATE> [PREDICATE]...)
(And <PREDICATE> [PREDICATE]...)
(PathMatch <STRING>)
(OnIndex <INTEGER>)
(Any <OPTION<ERROR>> [OPTION<ERROR>]...)
```

Some examples:
```
(ErrInjected (And (OnIndex 3) Reads))
(ErrInjected (And Reads (OnIndex 3)))
(Any (ENOSPC Writes) (ErrInjected (And Reads (OnIndex 2))))
```

Overboard?
</blockquote></details>

I decided to implement this because it's a little cleaner and more consistent.

jbowens (commented):
I'm good with either method—Maybe we won't ever have a need for multiple error types within the same test, so we should just nix it for now?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2995#-:-NgtZ5N49XSs3zDKWTwP:b-ywlyyc)* status: 6 of 15 files reviewed, 5 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`vfs/errorfs/errorfs.go` line 202 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2995#-Ng_2muD3gYseqYyKZ6b:-NgtYr-TAVTHPqQUhaky:b-fu0219) ([raw file](https://github.com/cockroachdb/pebble/blob/affd954686276d0beceb69267c64be00632814cf/vfs/errorfs/errorfs.go#L202)):*

Nice. Now that the predicate is separated out, is the only reason for existence of `Any` that someone may want to inject two different kinds of error, like in your ErrInjected and ENOSPC example?
</blockquote></details>

Yeah, that's the only reason for its existence.

RaduBerinde (approved):
:lgtm:

I like where it's at right now. We can always extend later if needed.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2995#-:-NguMGU10CRRr8N9QRjX:b-mvwhd1)* status: 6 of 15 files reviewed, 5 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`vfs/errorfs/errorfs.go` line 187 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2995#-NguMD838YPT4xMaAQXH:-NguMD838YPT4xMaAQXI:b-ag7ti7) ([raw file](https://github.com/cockroachdb/pebble/blob/01544e3e9f7896742d76ab714c1bf4bc72ab2075/vfs/errorfs/errorfs.go#L187)):*
> ```Go
> }
> 
> // Any returns an injector that injects an error if any of the provided
> ```

[nit] We don't need this anymore.

jbowens (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2995#-:-NguVJ950hVKhBf46UZM:bz5mair)* status: 6 of 15 files reviewed, 5 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`vfs/errorfs/errorfs.go` line 187 at r6](https://reviewable.io/reviews/cockroachdb/pebble/2995#-NguMD838YPT4xMaAQXH:-NguV-L151AojVFGNciG:b2hhscx) ([raw file](https://github.com/cockroachdb/pebble/blob/01544e3e9f7896742d76ab714c1bf4bc72ab2075/vfs/errorfs/errorfs.go#L187)):*

[nit] We don't need this anymore.
</blockquote></details>

I left it in to support the case that a user uses the list format of a datadriven argument, parsing each argument as a separate injector like you described.

PR #2992, opened 2023-10-11
event: mark `FormatMajorVersion` as safe from redaction

AlexTalks:
Previously nearly all events logged by the `EventListener` used types that had been marked as safe by implementing `SafeFormatter`, defining them as non-sensitive and not requiring redaction in logs, however `FormatMajorVersion` had been excluded. This change makes `FormatMajorVersion` a `redact.SafeValue` as well.

Requires the following in the [cockroachdb/cockroach `redactcheck.go`](https://github.com/cockroachdb/cockroach/blob/master/pkg/testutils/lint/passes/redactcheck/redactcheck.go):
```
"github.com/cockroachdb/pebble": {
	"FormatMajorVersion": {},
},

PR #2997, opened 2023-10-12
crl-release-23.2: event: mark `FormatMajorVersion` as safe from redaction

jbowens:
23.2 backport of #2992.

----

Previously nearly all events logged by the `EventListener` used types that had been marked as safe by implementing `SafeFormatter`, defining them as non-sensitive and not requiring redaction in logs, however `FormatMajorVersion` had been excluded. This change makes `FormatMajorVersion` a `redact.SafeValue` as well.

### DIFF
--- a/format_major_version.go
+++ b/format_major_version.go
@@ -28,10 +28,13 @@ import (
 // FormatVersion that the default corresponds to may change with time.
 type FormatMajorVersion uint64
 
+// SafeValue implements redact.SafeValue.
+func (v FormatMajorVersion) SafeValue() {}
+
 // String implements fmt.Stringer.
 func (v FormatMajorVersion) String() string {
-	// NB: This must not change. It's used as the value for the the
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
We should submit for approval using the bookmark in the slack channel. I think that makes more sense than approving pebble bumps with multiple unrelated changes.

jbowens (commented):
A change like this isn't permissible under the existing backport policy, so I've asked about it in Slack. https://cockroachlabs.slack.com/archives/C05B1BPMJLE/p1697137033649869?thread_ts=1696519932.519609&cid=C05B1BPMJLE

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2997#-:-Ng_swpT99SXamjW0o3b:b-pvnawc)* status: 0 of 1 files reviewed, all discussions resolved

PR #2998, opened 2023-10-12
objstorage: fix up redaction, error message formatting

jbowens:
Fix up a few instances where nonsensitive data could be redacted in logs or error messages, and a couple instances where errors were included twice (once as the target of Wrapf, and once as a formatting operand to Wrapf).

### DIFF
--- a/objstorage/objstorage.go
+++ b/objstorage/objstorage.go
@@ -13,6 +13,7 @@ import (
 	"github.com/cockroachdb/pebble/objstorage/objstorageprovider/sharedcache"
 	"github.com/cockroachdb/pebble/objstorage/remote"
 	"github.com/cockroachdb/pebble/vfs"
+	"github.com/cockroachdb/redact"
 )
 
 // Readable is the handle for an object that is open for reading.
@@ -166,6 +167,11 @@ type CreatorID uint64
 func (c CreatorID) IsSet() bool { return c != 0 }
 
 func (c CreatorID) String() string { return fmt.Sprintf("%d", c) }
+
+// SafeFormat implements redact.SafeFormatter.
+func (c CreatorID) SafeFormat(w redact.SafePrinter, _ rune) {
+	w.Printf("%d", redact.SafeUint(c))
+}
 
 // SharedCleanupMethod indicates the method for cleaning up unused shared objects.
 type SharedCleanupMethod uint8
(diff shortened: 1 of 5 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2998#-:-Ng_5Jem5frN2BUcO7Q7:bnfp4nl)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @sumeerbhola)

sumeerbhola (commented):
> Reviewed 4 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2998)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`objstorage/objstorage.go` line 174 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2998#-Ng_51OqCIlzGvFN950h:-NgdPDAECkcgaTsGyjEM:bobaezx) ([raw file](https://github.com/cockroachdb/pebble/blob/aa4bcb23c838c966c252e6aab30b56575f880afb/objstorage/objstorage.go#L174)):*

I implemented `SafeFormatter` rather than `SafeValue` to avoid needing to update the cockroach redaction linter. Since the linter is in a separate repository, it makes it difficult to rename or refactor types.
</blockquote></details>

My mental model is broken here.
In addition to implementing `SafeValue` does one need to add to the list in `redactcheck.go` too? Was this done as a second line of defence in case someone accidentally marked something as `SafeValue`?

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/2998#-:-NgeYl7cEIaczM7wc5_p:b-814g8k)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`objstorage/objstorage.go` line 174 at r1](https://reviewable.io/reviews/cockroachdb/pebble/2998#-Ng_51OqCIlzGvFN950h:-NgeYGQ97GUnFeuqzecv:b-r9yybf) ([raw file](https://github.com/cockroachdb/pebble/blob/aa4bcb23c838c966c252e6aab30b56575f880afb/objstorage/objstorage.go#L174)):*
> In addition to implementing `SafeValue` does one need to add to the list in `redactcheck.go` too?

Yeah, that's right.

> Was this done as a second line of defence in case someone accidentally marked something as `SafeValue`?

I don't really understand the motivation. Maybe it's just that `SafeValue` is considered to be a dangerous tool since a type can grow to contain non-safe values, whereas a SafeFormat implementation would need to be updated to serialize non-safe values?

PR #3002, opened 2023-10-17
compaction: remove the useless condition in check to delete compactions

lonng:
This PR removes the useless condition `d.mu.compact.compactingCount < maxConcurrentCompactions` in checking to delete compactions. Because the `db.mu` will be held while calling the function `maybeScheduleCompactionPicker`. The parallel compaction count limitation has been checked at L2308 and there is not any change between L2308 to L2337.

```go
	if d.mu.compact.compactingCount >= maxConcurrentCompactions {
		if len(d.mu.compact.manual) > 0 {
			// Inability to run head blocks later manual compactions.
			d.mu.compact.manual[0].retries++
		}
		return
	}
```


jbowens (approved):
LGTM

PR #3003, opened 2023-10-17
*: use the constant manifest.NumLevels instead of literal value

lonng:
There is a constant definition for `manifest.NumLevels`. But the literal value `7` is still hard code in the codebase. The PR tries to fix this issue and replace the literal value with the constant `manifest.NumLevels`.


jbowens (approved):
lgtm, thanks!

PR #3004, opened 2023-10-17
db: defer loading L0 range key blocks during iterator construction 

jbowens:
Previously, construction of an iterator over range keys that found sstables
containing range keys within L0 performed I/O to load the range key blocks
during iterator construction. This was less efficient: If the iterator
ultimately didn't need to read the keyspace overlapping the sstables containing
range keys, the block loads were unnecessary.

More significantly, if the I/O failed during iterator construction, the
resulting iterator was unusable. It would always error with the original error
returned by the failed block load. This is a deviation from iterator error
handling across the rest of the iterator stack, which allows an Iterator to be
re-seeked to clear the current iterator error.

Resolves https://github.com/cockroachdb/pebble/issues/2994.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -1453,6 +1453,15 @@ func runLSMCmd(td *datadriven.TestData, d *DB) string {
 func parseDBOptionsArgs(opts *Options, args []datadriven.CmdArg) error {
 	for _, cmdArg := range args {
 		switch cmdArg.Key {
+		case "auto-compactions":
+			switch cmdArg.Vals[0] {
+			case "off":
+				opts.DisableAutomaticCompactions = true
+			case "on":
+				opts.DisableAutomaticCompactions = false
+			default:
+				return errors.Errorf("Unrecognized %q arg value: %q", cmdArg.Key, cmdArg.Vals[0])
+			}
 		case "inject-errors":
 			injs := make([]errorfs.Injector, len(cmdArg.Vals))
 			for i := 0; i < len(cmdArg.Vals); i++ {
--- a/iterator_histories_test.go
+++ b/iterator_histories_test.go
@@ -32,6 +32,7 @@ func TestIterHistories(t *testing.T) {
 		}
 
 		var d *DB
+		var buf bytes.Buffer
 		iters := map[string]*Iterator{}
 		batches := map[string]*Batch{}
 		newIter := func(name string, reader Reader, o *IterOptions) *Iterator {
@@ -76,6 +77,7 @@ func TestIterHistories(t *testing.T) {
 		defer cleanup()
 
 		datadriven.RunTest(t, path, func(t *testing.T, td *datadriven.TestData) string {
+			buf.Reset()
 			switch td.Cmd {
 			case "define":
 				var err error
@@ -166,7 +168,6 @@ func TestIterHistories(t *testing.T) {
 						return fmt.Sprintf("unknown reader %q", arg.Vals[0])
 					}
 				}
-				var buf bytes.Buffer
 				for _, l := range strings.Split(td.Input, "\n") {
 					v, closer, err := reader.Get([]byte(l))
 					if err != nil {
@@ -187,6 +188,30 @@ func TestIterHistories(t *testing.T) {
 					return err.Error()
 				}
 				return ""
+			case "layout":
+				var verbose bool
+				var filename string
+				td.ScanArgs(t, "filename", &filename)
+				td.MaybeScanArgs(t, "verbose", &verbose)
+				f, err := opts.FS.Open(filename)
+				if err != nil {
+					return err.Error()
+				}
+				readable, err := sstable.NewSimpleReadable(f)
+				if err != nil {
+					return err.Error()
+				}
+				r, err := sstable.NewReader(readable, opts.MakeReaderOptions())
+				if err != nil {
+					return err.Error()
+				}
+				defer r.Close()
+				l, err := r.Layout()
+				if err != nil {
+					return err.Error()
+				}
+				l.Describe(&buf, verbose, r, nil)
+				return buf.String()
 			case "lsm":
 				return runLSMCmd(td, d)
 			case "metrics":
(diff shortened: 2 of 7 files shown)


jbowens:
Looks like I've got something to dig into still.

sumeerbhola (approved):
:lgtm: 

> Reviewed 3 of 3 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3004#-:-Nh7MMdfAs--JtCPgWpB:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`range_keys.go` line 70 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3004#-Nh7Kt3IAe_eGD9432Un:-Nh7Kt3J3MpDCIDH-94p:baorm5k) ([raw file](https://github.com/cockroachdb/pebble/blob/31c81a4ff565ae8cba7f8e161a0f8a84261608cb/range_keys.go#L70)):*
> ```Go
> 	// through the fileMetadata to determine that. Since L0's file count should
> 	// not significantly exceed ~1000 files (see L0CompactionFileThreshold),
> 	// this should be okay.
> ```

I wonder if we can quickly verify the cost of this by running kv0 and taking a CPU profile.
We need kv0 with block size that both overloads the LSM and doesn't run very cold wrt CPU. Maybe block size of 100 with very high client concurrency? May need some experimentation. Then we may have ~400 files in L0 with ~10 sub-levels. And the MVCCPuts are creating an iterator that needs to read range keys.
___
*[`testdata/iter_histories/errors` line 94 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3004#-Nh7M6ih31IIMbh-__ZO:-Nh7M6ih31IIMbh-__ZP:b-yg3k8j) ([raw file](https://github.com/cockroachdb/pebble/blob/31c81a4ff565ae8cba7f8e161a0f8a84261608cb/testdata/iter_histories/errors#L94)):*
> ```
> ----
> 
> combined-iter
> ```

loving this simple to read but sophisticated error injection test

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3004)* status: all files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`range_keys.go` line 70 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3004#-Nh7Kt3IAe_eGD9432Un:-Nh7s3rGDnIyqaKLN4hd:b-8nph64) ([raw file](https://github.com/cockroachdb/pebble/blob/31c81a4ff565ae8cba7f8e161a0f8a84261608cb/range_keys.go#L70)):*

I wonder if we can quickly verify the cost of this by running kv0 and taking a CPU profile.
We need kv0 with block size that both overloads the LSM and doesn't run very cold wrt CPU. Maybe block size of 100 with very high client concurrency? May need some experimentation. Then we may have ~400 files in L0 with ~10 sub-levels. And the MVCCPuts are creating an iterator that needs to read range keys.
</blockquote></details>

It seems tricky because we also need range keys to exist within L0. And there need to exist range keys within sstables that overlap the put key (in order to defeat the lazy-combined iterator optimization). Another option is to add one `keyspan.LevelIter` per L0 file containing a range key (ignoring the sublevels structure entirely) by iterating through `current.RangeKeyLevels`. This becomes O(files containing range keys) which should be significantly less, at the cost of more levels in the iterator stack.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3004)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`range_keys.go` line 70 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3004#-Nh7Kt3IAe_eGD9432Un:-Nh8bpANDxEYb8yzbZQ-:b-xaeeih) ([raw file](https://github.com/cockroachdb/pebble/blob/31c81a4ff565ae8cba7f8e161a0f8a84261608cb/range_keys.go#L70)):*
> This becomes O(files containing range keys) which should be significantly less, at the cost of more levels in the iterator stack.

It's no more than the levels we currently have on master, except that we'd be now wrapping each of those in a level iter, yes?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3004)* status: all files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`range_keys.go` line 70 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3004#-Nh7Kt3IAe_eGD9432Un:-NhCgZ4G7lTtWurFi-KA:b-y3uyk4) ([raw file](https://github.com/cockroachdb/pebble/blob/31c81a4ff565ae8cba7f8e161a0f8a84261608cb/range_keys.go#L70)):*

> This becomes O(files containing range keys) which should be significantly less, at the cost of more levels in the iterator stack.

It's no more than the levels we currently have on master, except that we'd be now wrapping each of those in a level iter, yes?
</blockquote></details>

Yeah, that's right. I gave this a try but unfortunately it's not viable today because the `keyspan.LevelIter` requires that the underlying `manifest.LevelIterator` be sorted by user keys, even if it's bounded to contain a single file. The `manifest.LevelIterator`'s bounding is implemented by first seeking within the b-tree, and then adjusting to the bounds. If the underlying b-tree is sorted by sequence numbers, it can return incorrect results.

Some additional alternatives: 
- implement a special `keyspan.FragmentIterator` that takes a single `fileMetadata`, defering the I/O until a relevant seek call is made.
- adjust the L0Sublevels code to construct a parallel b-tree containing only files with range keys for each sublevel. We don't need to actually recompute the L0 sublevels as if these range-key files are the only files in L0 to get the optimal read-amp.

jbowens:
TFTR!

Merging this, and filed #3007 for following up on potentially avoiding the O(# files in L0) iteration.

PR #3008, opened 2023-10-23
Fix cornercase in capping memory allowance

rjl493456442:
This pull request addresses a corner case when capping the pebble memory allowance. On a 32-bit machine, the memory usage is capped at MaxInt32 (note that this value is also inclusive), and on a 64-bit machine, it's capped at MaxUint32. These edge values should also be considered valid options.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -1381,7 +1381,7 @@ func (b *Batch) countData() []byte {
 
 func (b *Batch) grow(n int) {
 	newSize := len(b.data) + n
-	if uint64(newSize) >= maxBatchSize {
+	if uint64(newSize) > maxBatchSize {
 		panic(ErrBatchTooLarge)
 	}
 	if newSize > cap(b.data) {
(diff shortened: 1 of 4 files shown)


karalabe:
@jbowens Gentle ping if this is something that would go into the 23.2 release as a fix? We hit an issue in our code when setting the cap to exactly MaxUint32.

Also, the same issue applied to batch sizes. Perhaps that should also be allowed to reach the maxBatchSize constant?

karalabe:
Based on the comments for the two constants, it would seem that the max is inclusive as @rjl493456442 said. If it's exclusive, perhaps the comment / naming should be clarified a bit?

- https://github.com/cockroachdb/pebble/blob/a9906157653757d42f7bd1482f557001efb2fec6/open.go#L48
- https://github.com/cockroachdb/pebble/blob/a9906157653757d42f7bd1482f557001efb2fec6/open.go#L58

xxmplus:
Close to due to age

Issue #3014, opened 2023-10-25
db: categorized iterator stats


sumeerbhola:
We have had recent examples where a node exhibits high disk read bandwidth and read IOPS but we are unable to ascertain the origin of the work. The presence of the page cache muddies the picture, since misses in the block cache may hit in the page cache, and the page cache hit rate may differ for different readers on the same node (e.g. compactions that are rewriting a recently written file are likely to hit in the page cache).

But ignoring this page cache issue, we would like to know the source of block reads from the block cache and the block cache misses. For instance, in one case we think it was rangefeeds but don't have an exact measure. There are various categories that are useful in the CockroachDB context: sql-user, sql-stats, raft, rangefeed, mvcc-gc, range-snapshot. And Pebble internal categories include compactions, table-stats.

We would like such categorized iterator stats for all sstable iterators, not just `pebble.Iterator`, and expose them via `pebble.Metrics`. In the CockroachDB context, these would be then exported as CockroachDB metrics.

PR #3015, opened 2023-10-25
db: categorized stats for sstable iterators

sumeerbhola:
Informs #3014

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1342,7 +1342,13 @@ func (c *compaction) newInputIter(
 			}
 		}
 	}()
-	iterOpts := IterOptions{logger: c.logger}
+	iterOpts := IterOptions{
+		CategoryAndQoS: sstable.CategoryAndQoS{
+			Category: "pebble-compaction",
+			QoSLevel: sstable.NonLatencySensitiveQoSLevel,
+		},
+		logger: c.logger,
+	}
 
 	// Populate iters, rangeDelIters and rangeKeyIters with the appropriate
 	// constituent iterators. This depends on whether this is a flush or a
@@ -1367,8 +1373,8 @@ func (c *compaction) newInputIter(
 			// initRangeDel, the levelIter will close and forget the range
 			// deletion iterator when it steps on to a new file. Surfacing range
 			// deletions to compactions are handled below.
-			iters = append(iters, newLevelIter(iterOpts, c.comparer, newIters,
-				level.files.Iter(), l, internalIterOpts{
+			iters = append(iters, newLevelIter(context.Background(),
+				iterOpts, c.comparer, newIters, level.files.Iter(), l, internalIterOpts{
 					bytesIterated: &c.bytesIterated,
 					bufferPool:    &c.bufferPool,
 				}))
@@ -1409,7 +1415,8 @@ func (c *compaction) newInputIter(
 			// mergingIter.
 			iter := level.files.Iter()
 			for f := iter.First(); f != nil; f = iter.Next() {
-				rangeDelIter, closer, err := c.newRangeDelIter(newIters, iter.Take(), nil, l, &c.bytesIterated)
+				rangeDelIter, closer, err := c.newRangeDelIter(
+					newIters, iter.Take(), iterOpts, l, &c.bytesIterated)
 				if err != nil {
 					// The error will already be annotated with the BackingFileNum, so
 					// we annotate it with the FileNum.
@@ -1525,12 +1532,13 @@ func (c *compaction) newInputIter(
 func (c *compaction) newRangeDelIter(
 	newIters tableNewIters,
 	f manifest.LevelFile,
-	_ *IterOptions,
+	opts IterOptions,
 	l manifest.Level,
 	bytesIterated *uint64,
 ) (keyspan.FragmentIterator, io.Closer, error) {
+	opts.level = l
 	iter, rangeDelIter, err := newIters(context.Background(), f.FileMetadata,
-		&IterOptions{level: l}, internalIterOpts{
+		&opts, internalIterOpts{
 			bytesIterated: &c.bytesIterated,
 			bufferPool:    &c.bufferPool,
 		})
--- a/db.go
+++ b/db.go
@@ -1237,6 +1237,7 @@ func finishInitializingIter(ctx context.Context, buf *iterAlloc) *Iterator {
 // iteration is invalid in those cases.
 func (d *DB) ScanInternal(
 	ctx context.Context,
+	categoryAndQoS sstable.CategoryAndQoS,
 	lower, upper []byte,
 	visitPointKey func(key *InternalKey, value LazyValue, iterInfo IteratorLevel) error,
 	visitRangeDel func(start, end []byte, seqNum uint64) error,
@@ -1244,6 +1245,7 @@ func (d *DB) ScanInternal(
 	visitSharedFile func(sst *SharedSSTMeta) error,
 ) error {
 	scanInternalOpts := &scanInternalOptions{
+		CategoryAndQoS:   categoryAndQoS,
 		visitPointKey:    visitPointKey,
 		visitRangeDel:    visitRangeDel,
 		visitRangeKey:    visitRangeKey,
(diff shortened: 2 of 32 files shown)


jbowens (commented):
Approach looks good to me. I also now think it could be reasonable to elevate the combined Category/QualityOfService into an explicit non-Context type that we explicitly propagate separately from the context, so that we can avoid the alloc in all cases and providing a value becomes a requirement of the type system. Using the context also LGTM.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3015#-:-Ni00ARG3ZOI2Pj_Z-2q:b-2tkeam)* status: 0 of 25 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`sstable/category_stats.go` line 116 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3015#-Ni08Jco3IhpNVoA3273:-Ni04ZlIEau9HOZhxOgs:bdbjwoq) ([raw file](https://github.com/cockroachdb/pebble/blob/10445c1218ff9605e7492a2669995d70f9d157e0/sstable/category_stats.go#L116)):*
> ```Go
> 	v, ok := c.statsMap.Load(category)
> 	if !ok {
> 		c.mu.Lock()
> ```

probably excessive, but food for thought:

We could make `Category` a runtime-populated enum with clients calling `sstable.NewCategory(name string, qos QosLevel)` in package init funcs to assign values to the enum. We could then turn the `statsMap` into a slice to avoid the `sync.Map`, constructed at collector creation?

sumeerbhola (commented):
TFTR!
I've added a test and this is ready for a proper review.

> I also now think it could be reasonable to elevate the combined Category/QualityOfService into an explicit non-Context type that we explicitly propagate separately from the context, so that we can avoid the alloc in all cases and providing a value becomes a requirement of the type system.

Done. I added to `IterOptions` since that doesn't change the method signature for anyone creating `pebble.Iterator`. That also resulted in minimal signature changes to the internal methods since they usually already have `IterOptions`. It does mean it is easier to forget to pass it.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3015#-:-Ni5PDUc1WU1h-WwScUI:b-c3w6us)* status: 0 of 32 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`sstable/category_stats.go` line 116 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3015#-Ni08Jco3IhpNVoA3273:-Ni5PwCO1n8OT4m-yiJf:b-52q33i) ([raw file](https://github.com/cockroachdb/pebble/blob/10445c1218ff9605e7492a2669995d70f9d157e0/sstable/category_stats.go#L116)):*

probably excessive, but food for thought:

We could make `Category` a runtime-populated enum with clients calling `sstable.NewCategory(name string, qos QosLevel)` in package init funcs to assign values to the enum. We could then turn the `statsMap` into a slice to avoid the `sync.Map`, constructed at collector creation?
</blockquote></details>

I didn't do this, since it seems like an unnecessary complication. If we find that the map lookup is showing up in CPU profiles, we can revisit.

jbowens (approved):
:lgtm:

> Reviewed 2 of 25 files at r1, 30 of 30 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3015#-:-NiaTIkjDGEzMvnM0cCY:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @sumeerbhola)

___
*[`options.go` line 195 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3015#-NiaOlp3Bky8kC47q13S:-NiaOlp4D-Ao7wEVljbc:b-5lt742) ([raw file](https://github.com/cockroachdb/pebble/blob/8c566397cd86914a12875fa24a05c0a24f4be779/options.go#L195)):*
> ```Go
> 	// existing iterator, such as OnlyReadGuaranteedDurable, UseL6Filters, if
> 	// the point iterator has already been initialized. Add comments regarding
> 	// which can be changed.
> ```

We have logic in SetOptions that forces the point iterator to be reconstructed if `OnlyReadGuaranteedDurable` or `UseL6Filters`. Alternatively we could add similar logic for `CategroyAndQoS`
___
*[`sstable/category_stats.go` line 135 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3015#-NiaQiXr4LYXBj3CmcBo:-NiaQiXr4LYXBj3CmcBp:bjq8oko) ([raw file](https://github.com/cockroachdb/pebble/blob/8c566397cd86914a12875fa24a05c0a24f4be779/sstable/category_stats.go#L135)):*
> ```Go
> 		if len(s.Category) == 0 {
> 			// { makes it sort last.
> 			s.Category = "{unknown}"
> ```

will this be okay when converted into a timeseries name? I don't know what restrictions prometheus or the like have on series names, but in promql I know braces are used to filter on series metadata. Did a bit searching and it looks like the only non-alphanumeric characters that prometheus allows are underscores or colons: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels I think we define metric names in Cockroach using hypens, so something must automatically translate? Anyways maybe underscores or colons will play nicer.
___
*[`sstable/reader_iter_single_lvl.go` line 60 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3015#-NiaSt0q-YOWCdfkYjp3:-NiaSt0q-YOWCdfkYjp4:b-c9tamn) ([raw file](https://github.com/cockroachdb/pebble/blob/8c566397cd86914a12875fa24a05c0a24f4be779/sstable/reader_iter_single_lvl.go#L60)):*
> ```Go
> 	// being reused). It is currently only provided when the iterator tree is
> 	// rooted at pebble.Iterator. iterStats is this sstable iterators private
> 	// stats that are not reported to a CategoryStatsCollector when this
> ```

s/are not reported/are reported/?

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3015)* status: 14 of 32 files reviewed, 4 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`options.go` line 195 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3015#-NiaOlp3Bky8kC47q13S:-Nj9OZFE1XjM-sjijJEb:b-9tifpf) ([raw file](https://github.com/cockroachdb/pebble/blob/8c566397cd86914a12875fa24a05c0a24f4be779/options.go#L195)):*

We have logic in SetOptions that forces the point iterator to be reconstructed if `OnlyReadGuaranteedDurable` or `UseL6Filters`. Alternatively we could add similar logic for `CategroyAndQoS`
</blockquote></details>

I've simply removed this TODO. I didn't add any logic to `SetOptions` to recreate the point iterator if this is changed -- I am slightly worried that someone will forget to set `CategoryAndQoS` when calling `SetOptions` and cause a performance regression *and* stop collecting stats in the proper category. And we don't currently have a reuse use case that needs changing `CategoryAndQos`.
___
*[`sstable/category_stats.go` line 135 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3015#-NiaQiXr4LYXBj3CmcBo:-Nj9Pnx0Eylr5a-BYXLa:b-jxm4wr) ([raw file](https://github.com/cockroachdb/pebble/blob/8c566397cd86914a12875fa24a05c0a24f4be779/sstable/category_stats.go#L135)):*

will this be okay when converted into a timeseries name? I don't know what restrictions prometheus or the like have on series names, but in promql I know braces are used to filter on series metadata. Did a bit searching and it looks like the only non-alphanumeric characters that prometheus allows are underscores or colons: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels I think we define metric names in Cockroach using hypens, so something must automatically translate? Anyways maybe underscores or colons will play nicer.
</blockquote></details>

Changed to `_unknown`.
___
*[`sstable/reader_iter_single_lvl.go` line 60 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3015#-NiaSt0q-YOWCdfkYjp3:-Nj9Q9Xd0g42KkP_MoKX:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/8c566397cd86914a12875fa24a05c0a24f4be779/sstable/reader_iter_single_lvl.go#L60)):*

s/are not reported/are reported/?
</blockquote></details>

Done

PR #3019, opened 2023-10-27
vfs/errorfs: expand interface and extract parsing logic

jbowens:
The first commit adds a few additional errorfs injectors and predicates. Together these primitives allow the replacement of a few custom injectors across the codebase, by refactoring them in terms of these primitives.

The new primitives:

- A Toggle type implements errorfs.Injector and allows toggling whether an injector should be used at runtime.
- A Counter type implements errorfs.Injector and counts the number of injected errors. It allows tests to ensure that errors are surfaced appropriately and not swallowed.
- A Not predicate allows for negating other predicates.
- The OpKind type now implements the Predicate interface and is integrated into the DSL as a predicate constant that returns true if the operation is of the same kind.

The second commit extracts much of the DSL parsing logic into a separate package, internal/dsl. This allows other packages to define their own DSLs easily.

### DIFF
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -2913,7 +2913,7 @@ func TestCompactionErrorCleanup(t *testing.T) {
 				if initialSetupDone {
 					tablesCreated = append(tablesCreated, info.FileNum)
 					if len(tablesCreated) >= 2 {
-						ii.SetIndex(0)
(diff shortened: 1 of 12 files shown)


jbowens:
I'm going to close this out and resequence the stack of PRs.

PR #3020, opened 2023-10-27
vfs: pass start time in nanos to disk-health timing functions

### LINKED ISSUE #3009: vfs: include start time in disk health checker timing stack traces
When a disk stall results in the termination of a process and `GOTRACEBACK` is set appropriately, the panic includes a dump of stack traces. These stack traces can be used to confirm the presence of a goroutine stuck in the described syscall. However, there's no way to verify that the two syscall invocations are actually the same... The previous syscall could have completed and a new one introduced. We could add the unix timestamp in nanoseconds of the operation's start as a parameter to the function performing the timing. This would allow us to inspect the stack trace and verify that the start time matches the alleged start of the disk stall.

jbowens:
Pass the start time in the form of nanoseconds since the unix epoch as a parameter to timeDiskOp and timeFilesystemOp. This aids post-mortem debugging when the disk-health checker fatals the process and GOTRACEBACK is set to dump the stacks, including arguments. The start time argument will be printed in hex form, allowing us to decode the start time of the operation. This can be used to confirm that the timed operation was still inflight at the time stacks were collected.

Close #3009.

### DIFF
--- a/vfs/disk_health.go
+++ b/vfs/disk_health.go
@@ -135,7 +135,7 @@ type diskHealthCheckingFile struct {
 	// NB: this packing scheme is not persisted, and is therefore safe to adjust
 	// across process boundaries.
 	lastWritePacked atomic.Uint64
-	createTime      time.Time
+	createTimeNanos int64
 }
 
 // newDiskHealthCheckingFile instantiates a new diskHealthCheckingFile, with the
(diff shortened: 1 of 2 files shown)


itsbilal (approved):
LGTM :shipit:

jbowens:
TFTR!

PR #3021, opened 2023-10-27
internal/keyspan: fix InterleavingIter error handling

jbowens:
Previously, the InterleavingIter could violate the InternalIterator contract when an error is encountered. If one of its two child iterators encountered an error and returned a nil KV pair, the iterator could yield a non-nil KV to the caller. This commit updates the interleaving iterator error handling to accumulate any encountered errors on a field and always yield nil.

This bug was surfaced by a new 'invariants' tag assertion in Valid, asserting that Error() returns nil if the iterator is valid. With this invariants tag, existing tests surface the assertion failure. Additional test coverage will come with #1115.

### DIFF
--- a/internal/keyspan/interleaving_iter.go
+++ b/internal/keyspan/interleaving_iter.go
@@ -7,6 +7,7 @@ package keyspan
 import (
 	"fmt"
 
+	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/invariants"
 )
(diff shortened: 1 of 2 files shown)


itsbilal (approved):
LGTM

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3021#-:-Ni0xLZCAWsYx-vlITgr:b-814g8k)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @itsbilal)

___
*[`internal/keyspan/interleaving_iter.go` line 1004 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3021#gh-1376563386:-Ni0xN4o-LlsPkzZh71i:b-bb39q7) ([raw file](https://github.com/cockroachdb/pebble/blob/098370bf3ddeb3e411e2250d09b3ad072ada4b87/internal/keyspan/interleaving_iter.go#L1004)):*

Does it make sense to add an `if invariants.Enabled { if key != nil && i.pointIter.Error() { panic() }}` as well here?
</blockquote></details>

Good idea! I did that here for the pointIter, and elsewhere for the keyspanIter.

Issue #2873, opened 2023-09-04
Panic in log writer: `panic: sync: negative WaitGroup counter`


LINKIWI:
PebbleDB version: `v0.0.0-20230728201528-5e7f88523982`
Go version: `1.21.0`

Observation: A PebbleDB-based application sporadically panics during writes, with the following crash symptom:

```
panic: sync: negative WaitGroup counter
goroutine 10697 [running]:
sync.(*WaitGroup).Add(0x0?, 0xc00020e240?)
#011GOROOT/src/sync/waitgroup.go:62 +0xd8
sync.(*WaitGroup).Done(...)
#011GOROOT/src/sync/waitgroup.go:87
github.com/cockroachdb/pebble/record.(*syncQueue).pop(0xc00046a118, 0x121, 0x45e2?, {0x10a7700?, 0xc00051e0f0}, 0xc0002d64e0)
#011external/com_github_cockroachdb_pebble/record/log_writer.go:157 +0x109
github.com/cockroachdb/pebble/record.(*LogWriter).flushLoop(0xc00046a000, {0xe4d800, 0x17558e0})
#011external/com_github_cockroachdb_pebble/record/log_writer.go:465 +0x645
runtime/pprof.Do({0x10adca8?, 0x17558e0?}, {{0xc000298040?, 0xc0002f0190?, 0x10b1c38?}}, 0xc0001b9fc0)
#011GOROOT/src/runtime/pprof/runtime.go:51 +0x9d
github.com/cockroachdb/pebble/record.NewLogWriter.func2()
#011external/com_github_cockroachdb_pebble/record/log_writer.go:355 +0x56
created by github.com/cockroachdb/pebble/record.NewLogWriter in goroutine 10693
#011external/com_github_cockroachdb_pebble/record/log_writer.go:354 +0x416
```

The implicated callsite appears to be: https://github.com/cockroachdb/pebble/blob/a4fe05009372ba870cc6fe4e9b8c4873b576c549/record/log_writer.go#L157

Unfortunately, the crash is not deterministically reproducible, so I don't have a minimally reproducible example at this time.

Additional useful usage context:

* The application's only usage pattern is pretty simple: only involves creating new IndexedBatches, performing no more than ~10 R/W operations on the batch, followed by committing the batch with option `pebble.Sync`. Unfortunately this application is not open source.
* The above operations occur at a very low frequency: about 1 request every few minutes.
* With this throughput, the above crash occurs about once every few hours.

[...]

PR #3026, opened 2023-10-30
db: guard against concurrent batch writes during Commit

### LINKED ISSUE #3024: batch: guard against additional batch writes during Commit
Once a Batch has started to commit, it is not safe to mutate the batch. One possible explanation for cockroachdb/cockroach#113268 is higher-level code mutating a batch as it's being applied to the memtable. We should add safeguards against this where we can. At the very least we can ensure that a batch being applied to the memtable has the same count before and after it's applied.

jbowens:
Add a guardrail that ensures a batch isn't in the process of being committed
when applying a mutation to the Batch. We've observed panics during batch
application that suggest concurrent modification of the batch, and these
guardrails will help surface any races more readily.

Informs cockroachdb/cockroach#113268.
Resolve #3024.
Informs #2873.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -267,11 +267,6 @@ type batchInternal struct {
 	// memtable.
 	flushable *flushableBatch
 
-	// ingestedSSTBatch indicates that the batch contains one or more key kinds
-	// of InternalKeyKindIngestSST. If the batch contains key kinds of IngestSST
-	// then it will only contain key kinds of IngestSST.
-	ingestedSSTBatch bool
-
 	// minimumFormatMajorVersion indicates the format major version required in
 	// order to commit this batch. If an operation requires a particular format
 	// major version, it ratchets the batch's minimumFormatMajorVersion. When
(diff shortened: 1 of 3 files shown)


jbowens:
Need to rework the first commit

sumeerbhola (approved):
:lgtm: 

> Reviewed 2 of 2 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3026#-:-NiAfvUB0XkXdqeYyZVq:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`batch.go` line 306 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3026#-NiAfczG3Ay62lr2f8We:-NiAfczG3Ay62lr2f8Wf:b-efnli9) ([raw file](https://github.com/cockroachdb/pebble/blob/cf713ea4e8c3aff97d1560ec5a38fbb895a10b5e/batch.go#L306)):*
> ```Go
> 
> 	// committing is set to true when a batch begins to commit. It's used to
> 	// ensure the batch is mutated concurrently. It is not an atomic
> ```

is not mutated?

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3026#-:-NiAlg8K8-iiz620FASA:b-814g8k)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`batch.go` line 306 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3026#-NiAfczG3Ay62lr2f8We:-NiAlfbrB1v9s7IX5cW7:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/cf713ea4e8c3aff97d1560ec5a38fbb895a10b5e/batch.go#L306)):*

is not mutated?
</blockquote></details>

Done.

PR #3027, opened 2023-10-30
crl-release-23.2: internal/keyspan: fix InterleavingIter error handling

jbowens:
This is a 23.2 backport of #3021 and #3035.

----

Previously, the InterleavingIter could violate the InternalIterator contract when an error is encountered. If one of its two child iterators encountered an error and returned a nil KV pair, the iterator could yield a non-nil KV to the caller. This commit updates the interleaving iterator error handling to accumulate any encountered errors on a field and always yield nil.

This bug was surfaced by a new 'invariants' tag assertion in Valid, asserting that Error() returns nil if the iterator is valid. With this invariants tag, existing tests surface the assertion failure. Additional test coverage will come with #1115.

### DIFF
--- a/internal/keyspan/interleaving_iter.go
+++ b/internal/keyspan/interleaving_iter.go
@@ -112,7 +112,22 @@ type InterleavingIter struct {
 	nextPrefixBuf []byte
 	pointKey      *base.InternalKey
 	pointVal      base.LazyValue
-	err           error
+	// err holds an iterator error from either pointIter or keyspanIter. It's
+	// reset to nil on seeks. An overview of error-handling mechanics:
+	//
+	// Whenever either pointIter or keyspanIter is respositioned and a nil
+	// key/span is returned, the code performing the positioning is responsible
+	// for checking the iterator's Error() value. This happens in savePoint and
+	// saveSpan[Forward,Backward].
+	//
+	// Once i.err is non-nil, the computation of i.pos must set i.pos =
+	// posExhausted. This happens in compute[Smallest|Largest]Pos and
+	// [next|prev]Pos. Setting i.pos to posExhausted ensures we'll yield nil to
+	// the caller, which they'll interpret as a signal they must check Error().
+	//
+	// INVARIANTS:
+	// i.err != nil => i.pos = posExhausted
+	err error
 	// prefix records the iterator's current prefix if the iterator is in prefix
 	// mode. During prefix mode, Pebble will truncate spans to the next prefix.
 	// If the iterator subsequently leaves prefix mode, the existing span cached
@@ -600,6 +615,23 @@ func (i *InterleavingIter) computeLargestPos() {
 
 // nextPos advances the iterator one position in the forward direction.
 func (i *InterleavingIter) nextPos() {
+	if invariants.Enabled {
+		defer func() {
+			if i.err != nil && i.pos != posExhausted {
+				panic(errors.AssertionFailedf("iterator has accumulated error but i.pos = %d", i.pos))
+			}
+		}()
+	}
+	// NB: If i.err != nil or any of the positioning methods performed in this
+	// function result in i.err != nil, we must set i.pos = posExhausted. We
+	// perform this check explicitly here, but if any of the branches below
+	// advance either iterator, they must also check i.err and set posExhausted
+	// if necessary.
+	if i.err != nil {
+		i.pos = posExhausted
+		return
+	}
+
 	switch i.pos {
 	case posExhausted:
 		i.savePoint(i.pointIter.Next())
@@ -608,6 +640,10 @@ func (i *InterleavingIter) nextPos() {
 		i.computeSmallestPos()
 	case posPointKey:
 		i.savePoint(i.pointIter.Next())
+		if i.err != nil {
+			i.pos = posExhausted
+			return
+		}
 		// If we're not currently within the span, we want to chose the
 		// MIN(pointKey,span.Start), which is exactly the calculation performed
 		// by computeSmallestPos.
@@ -619,8 +655,6 @@ func (i *InterleavingIter) nextPos() {
 		// Since we previously were within the span, we want to choose the
 		// MIN(pointKey,span.End).
 		switch {
-		case i.err != nil:
-			i.pos = posExhausted
 		case i.span == nil:
 			panic("i.withinSpan=true and i.span=nil")
 		case i.pointKey == nil:
@@ -653,6 +687,23 @@ func (i *InterleavingIter) nextPos() {
 
 // prevPos advances the iterator one position in the reverse direction.
 func (i *InterleavingIter) prevPos() {
+	if invariants.Enabled {
+		defer func() {
+			if i.err != nil && i.pos != posExhausted {
+				panic(errors.AssertionFailedf("iterator has accumulated error but i.pos = %d", i.pos))
+			}
+		}()
+	}
+	// NB: If i.err != nil or any of the positioning methods performed in this
+	// function result in i.err != nil, we must set i.pos = posExhausted. We
+	// perform this check explicitly here, but if any of the branches below
+	// advance either iterator, they must also check i.err and set posExhausted
+	// if necessary.
+	if i.err != nil {
+		i.pos = posExhausted
+		return
+	}
+
 	switch i.pos {
 	case posExhausted:
 		i.savePoint(i.pointIter.Prev())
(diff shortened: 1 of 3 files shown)


nicktrav:
EM approval ✅

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3027)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/keyspan/interleaving_iter.go` line 221 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3027#-NiA7PXO5ern1A90ke-D:-NiA7PXO5ern1A90ke-E:bbzd2s6) ([raw file](https://github.com/cockroachdb/pebble/blob/dee732271c4e59f560971dabe73114fcb58e02fa/internal/keyspan/interleaving_iter.go#L221)):*
> ```Go
> 	i.clearMask()
> 	i.prefix = prefix
> 	i.savePoint(pointKey, pointValue)
> ```

will `pointKey` always be non-nil and the `pointIter` have no error in these `InitSeek*` cases?
___
*[`internal/keyspan/interleaving_iter.go` line 284 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3027#-NiA8zM20mPez6EdS1aY:-NiA8zM20mPez6EdS1aZ:bivze2r) ([raw file](https://github.com/cockroachdb/pebble/blob/dee732271c4e59f560971dabe73114fcb58e02fa/internal/keyspan/interleaving_iter.go#L284)):*
> ```Go
> 	i.dir = +1
> 	i.computeSmallestPos()
> 	return i.yieldPosition(key, i.nextPos)
> ```

We seem to have the behavior that:
A. once `i.err` becomes non-nil, some piece of code will set `i.pos = posExhausted`.
B. Something will notice `posExhausted` and return nil.

B is done in `yieldPosition` which is the last thing all the seeking/stepping method call.

A happens very clearly in `computeSmallestPos`, `computeLargestPos`, and less clearly in `nextPos`, `prevPos`.

For the various seek methods we call `computeSmallestPos` or `computeLargestPos` immediately before calling `yieldPosition`.  But the stepping methods call `nextPos`, `prevPos`, so I worry that we may miss some case where `i.err` is non-nil, but forget to set `posExhausted`.
nextPos` also calls `computeSmallestPos` and `prevPos` calls `computeLargestPos`, in some cases. One of the other cases in `nextPos` explicitly handles the error
```
case i.err != nil:  
 i.pos = posExhausted
```
There is another one, that doesn't seem to do so explicitly.
```
case posKeyspanStart:  
 // Either a point key or the span's end key comes next.  
 if i.pointKey != nil && i.cmp(i.pointKey.UserKey, i.span.End) < 0 {  
  i.pos = posPointKey  
 } else {  
  i.pos = posKeyspanEnd  
 }
```

Could you add some code comments (on master), documenting some invariants that make it slightly easier to see that the code is correct.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3027)* status: all files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/keyspan/interleaving_iter.go` line 221 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3027#-NiA7PXO5ern1A90ke-D:-NiGVfHh3MIw3rLtseS5:b-m4wxuv) ([raw file](https://github.com/cockroachdb/pebble/blob/dee732271c4e59f560971dabe73114fcb58e02fa/internal/keyspan/interleaving_iter.go#L221)):*

will `pointKey` always be non-nil and the `pointIter` have no error in these `InitSeek*` cases?
</blockquote></details>

I don't think so, no. It's possible for `pointKey == nil && i.pointIter.Error() != nil`.
___
*[`internal/keyspan/interleaving_iter.go` line 284 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3027#-NiA8zM20mPez6EdS1aY:-NiGcIBc64wXCtWi7GXF:b-3c49g) ([raw file](https://github.com/cockroachdb/pebble/blob/dee732271c4e59f560971dabe73114fcb58e02fa/internal/keyspan/interleaving_iter.go#L284)):*

We seem to have the behavior that:
A. once `i.err` becomes non-nil, some piece of code will set `i.pos = posExhausted`.
B. Something will notice `posExhausted` and return nil.

B is done in `yieldPosition` which is the last thing all the seeking/stepping method call.

A happens very clearly in `computeSmallestPos`, `computeLargestPos`, and less clearly in `nextPos`, `prevPos`.

For the various seek methods we call `computeSmallestPos` or `computeLargestPos` immediately before calling `yieldPosition`.  But the stepping methods call `nextPos`, `prevPos`, so I worry that we may miss some case where `i.err` is non-nil, but forget to set `posExhausted`.
nextPos` also calls `computeSmallestPos` and `prevPos` calls `computeLargestPos`, in some cases. One of the other cases in `nextPos` explicitly handles the error
```
case i.err != nil:  
 i.pos = posExhausted
```
There is another one, that doesn't seem to do so explicitly.
```
case posKeyspanStart:  
 // Either a point key or the span's end key comes next.  
 if i.pointKey != nil && i.cmp(i.pointKey.UserKey, i.span.End) < 0 {  
  i.pos = posPointKey  
 } else {  
  i.pos = posKeyspanEnd  
 }
```

Could you add some code comments (on master), documenting some invariants that make it slightly easier to see that the code is correct.
</blockquote></details>

Thanks, the complexity of nextPos|prevPos was hiding an additional bug when switching directions. Put up #3035.

RaduBerinde (commented):
:lgtm:

Definitely an improvement, but the general approach still feels very fragile.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3027#-:-NiGmFVU3CqvqvsVLq1y:b-bfrrik)* status: all files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/keyspan/interleaving_iter.go` line 221 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3027#-NiA7PXO5ern1A90ke-D:-NiGlkWf-Jrv-T35n94r:bvob429) ([raw file](https://github.com/cockroachdb/pebble/blob/dee732271c4e59f560971dabe73114fcb58e02fa/internal/keyspan/interleaving_iter.go#L221)):*

I don't think so, no. It's possible for `pointKey == nil && i.pointIter.Error() != nil`.
</blockquote></details>

It does feels strange that we check the iterator here though, when the key is coming from the caller. Anyway, is this sufficient or are there more missing error checks in `lazyCombinedIter`?
___
*[`internal/keyspan/interleaving_iter.go` line 622 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3027#-NiGl8co7UB15cflhmWB:-NiGl8co7UB15cflhmWC:bb1v7wz) ([raw file](https://github.com/cockroachdb/pebble/blob/dee732271c4e59f560971dabe73114fcb58e02fa/internal/keyspan/interleaving_iter.go#L622)):*
> ```Go
> 		// MIN(pointKey,span.End).
> 		switch {
> 		case i.err != nil:
> ```

We don't need to check this in the other `case`s below?

sumeerbhola (approved):
:lgtm: 
I'm assuming https://github.com/cockroachdb/pebble/pull/3035 will also be backported

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3027#-:-NiLjsWi4qgbvUnIRGcv:b-1xmk4k)* status: all files reviewed, 3 unresolved discussions (waiting on @jbowens)

jbowens (commented):
Updated to pull in #3035 too

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3027#-:-NiqAfke0WDwc8eqwmkJ:b-esyvdx)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

RaduBerinde (approved):
:lgtm:

> Reviewed 1 of 2 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3027#-:-Nir51hW6n7dH1sLU95S:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

jbowens:
TFTRs!

PR #3034, opened 2023-11-02
internal/keyspan: fix defragmenting iterator error handling

jbowens:
A single iterator operation performed on a keyspan defragmenting iterator may
result in many iterator operations performed against its inner iterator.
Previously, in some circumstances if an error occurred during an inner iterator
operation other than the final one, the error was silently swallowed. This
allowed incorrect results, including spans that were not fully defragmented.

These commits fix these bugs. They also introduce some internal keyspan testing
facilities that build on the new internal/dsl package to provide iterator
"probes" for use in the keyspan datadriven tests. These probes may be used to
conditionally inject errors, mutate return values or log operations.

Future work will use these probes to expand error-handling unit tests for the
other iterators within the keyspan package.

### DIFF
--- a/internal/keyspan/bounded_test.go
+++ b/internal/keyspan/bounded_test.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/cockroachdb/datadriven"
 	"github.com/cockroachdb/pebble/internal/testkeys"
-	"github.com/stretchr/testify/require"
 )
 
 func TestBoundedIter(t *testing.T) {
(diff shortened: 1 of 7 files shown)


RaduBerinde (approved):
:lgtm:

Very cool testing facility!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3034#-:-NiGHI_JAtGxRk3FKDwk:b-6wwhps)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`internal/keyspan/datadriven_test.go` line 56 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3034#-NiGGj8557HnvYGoh2PH:-NiGGj8557HnvYGoh2PI:b475nu9) ([raw file](https://github.com/cockroachdb/pebble/blob/87a83286c41eac2224bf934e69fa1ad959339db3/internal/keyspan/datadriven_test.go#L56)):*
> ```Go
> }
> 
> // probe defines an interface for probes that may inspect or mutate internal
> ```

[nit] move before the functions above

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3034#-:-NiMIJh2A1WeUROhUGF1:b-814g8k)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @sumeerbhola)

PR #3037, opened 2023-11-03
db: add SetContext method to Iterator to override existing context

sumeerbhola:
This is to allow for (short-lived) Iterator reuse in CockroachDB, and cases where the code that uses an Iterator is "far" from the code that uses the iterator (e.g. rangefeed.registration.maybeRunCatchUpScan).

Relates to https://github.com/cockroachdb/cockroach/issues/113257 and https://github.com/cockroachdb/cockroach/issues/113256

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -1702,6 +1702,8 @@ func (i *batchIter) SetBounds(lower, upper []byte) {
 	i.iter.SetBounds(lower, upper)
 }
 
+func (i *batchIter) SetContext(_ context.Context) {}
+
 type flushableBatchEntry struct {
 	// offset is the byte offset of the record within the batch repr.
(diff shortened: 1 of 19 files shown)


RaduBerinde (approved):
:lgtm:

We should probably plumb the context to each iterator operation rather than storing it in the iterator.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3037#-:-NiWAAKp5rqzMnCqc81n:b-o9250)* status: 0 of 19 files reviewed, all discussions resolved (waiting on @jbowens)

RaduBerinde:
Maybe we can add a Context to the Seek methods, and Next would always use the one from the last Seek.

PR #3040, opened 2023-11-07
release-23.2: db: add KeyStatistics.LatestKindsCount

sumeerbhola:
Also give one example of how to interpret these statistics. And clarify that SnapshotPinned* are not counting what they are claiming to count.

----

Release justification: Low risk observability improvement motivated by escalation involving CRDB liveness range.


sumeerbhola:
TFTR!

PR #3029, opened 2023-11-01
internal/dsl: new package

jbowens:
Extract the parsing logic of vfs/errorfs's DSL into a separate package providing generic interfaces for parsing lisp-esque DSLs.

Some error pathways are difficult to test only through filesystem-injected errors. This new package prepares for a similar DSL for errors injected by a special InternalIterator that wraps other internal iterators.

PR #3045, opened 2023-11-08
crl-release-23.2: internal/keyspan: fix defragmenting iterator error handling

jbowens:
This is a 23.2 backport of #3029 and #3034 in order to fix and test the range key defragmenting iterator's handling of errors.

----

**internal/dsl: new package**

Extract the parsing logic of vfs/errorfs's DSL into a separate package
providing generic interfaces for parsing lisp-esque DSLs.

Some error pathways are difficult to test only through filesystem-injected
errors. This new package prepares for a similar DSL for errors injected by a
special InternalIterator that wraps other internal iterators.

**internal/keyspan: consolidate datadriven iterator ops**

Previously, a few different datadriven test repeated logic for driving a
FragmentIterator through datadriven commands. This commit refactors them to
share code for driving these datadriven commands.

**internal/keyspan: fix defragmenting iterator error handling**

A single iterator operation performed on a keyspan defragmenting iterator may
result in many iterator operations performed against its inner iterator.
Previously, in some circumstances if an error occurred during an inner iterator
operation other than the final one, the error was silently swallowed. This
allowed incorrect results, including spans that were not fully defragmented.

This commit fixes these bugs. It also introduces some internal keyspan testing
facilities that build on the new internal/dsl package to provide iterator
"probes" for use in the keyspan datadriven tests. These probes may be used to
conditionally inject errors, mutate return values or log operations.

Future work will use these probes to expand error-handling unit tests for the
other iterators within the keyspan package.

### DIFF
--- a/internal/dsl/dsl.go
+++ b/internal/dsl/dsl.go
@@ -1,0 +1,160 @@
+// Copyright 2023 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+// Package dsl provides facilities for parsing lisp-like domain-specific
+// languages (DSL).
+package dsl
+
+import (
+	"fmt"
+	"go/scanner"
+	"go/token"
+	"strconv"
+	"strings"
+
+	"github.com/cockroachdb/errors"
+)
+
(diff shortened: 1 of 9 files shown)


jbowens:
TFTR!

PR #3044, opened 2023-11-08
metamorphic: add IngestAndExcise, make EFOS determinitic

itsbilal:
This change updates the metamorphic tests to support a new useExcise testOption and a new IngestAndExcise operation, which calls db.IngestAndExcise() if useExcise is true, or calls it with an sstable containing DeleteRange/RangeKeyDelete for the same span if useExcise is false.

Furthermore, a new private option is added, efosAlwaysCreatesIterators. If the test option for useExcise (or the previously added useSharedReplicate) is enabled, one of these two things happen to guarantee EFOS determinism around excises:

1) If efosAlwaysCreatesIterators is true, the behaviour of EFOS is changed to always hold onto a readState and never return errors from NewIter.
2) If efosAlwaysCreatesIterators is false, regular EFOS behaviour is retained, but a flush is completed after every EFOS instantiation to guarantee that it becomes excise-immune right at instantiation.

Note that regular snapshots are never created when useExcise or useSharedReplicate is true.

Fixes #2885.

PR #3046, opened 2023-11-08
internal/keyspan: obey Seek invariants in filteringIter

itsbilal:
Previously if filteringIter's FilterFunc mutated the passed-in span to no longer be a valid return value for a SeekLT or SeekGE call, we would still return that span even though it could be >=  the seek key (for SeekLT), or less than it (for SeekGE).

This change updates filteringIter to guard for this case before returning from a seek call.

Found with #3044.

Informs cockroachdb/cockroach#113973, cockroachdb/cockroach#114056.

### DIFF
--- a/internal/keyspan/filter.go
+++ b/internal/keyspan/filter.go
@@ -4,6 +4,8 @@
 
 package keyspan
 
+import "github.com/cockroachdb/pebble/internal/base"
+
 // FilterFunc defines a transform from the input Span into the output Span. The
 // function returns true if the Span should be returned by the iterator, and
 // false if the Span should be skipped. The FilterFunc is permitted to mutate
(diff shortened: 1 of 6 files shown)


itsbilal:
This needs a test. I'll write one within the next hour.

itsbilal:
Tests added.

jbowens (approved):
Nice find.

:lgtm:

Do you mind filing an issue about refactoring `keyspan.Truncate`? If we don't use `keyspan.Filter` and implement `keyspan.FragmentIterator` we can avoid the key comparison when we don't truncate.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3046#-:-NionCvLBwXabMbUKqnV:b-p84jod)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @erikgrinaker)

itsbilal:
TFTR! Filed #3048.

I should backport this too right?

Issue #3048, opened 2023-11-09
internal/keyspan: refactor Truncate to not use filteringIter


itsbilal:
Currently, `keyspan.Truncate` creates a `filteringIter` with a filter function. We uncovered a bug around certain uses of seeks in `filteringIter` in #3046 that forced us to add a check around Seek calls in the filteringIter, at the cost of one additional key comparison. If we refactored keyspan.Truncate to not use `filteringIter`, we could make that key comparison specific to just the `Truncate` case, and not other uses of `filteringIter` that don't mutate spans.

RaduBerinde:
Fixed by #3282.

PR #3049, opened 2023-11-09
crl-release-23.2: internal/keyspan: obey Seek invariants in filteringIter

itsbilal:
23.2 Backport of #3046.

----

Previously if filteringIter's FilterFunc mutated the passed-in span to no longer be a valid return value for a SeekLT or SeekGE call, we would still return that span even though it could be >=  the seek key (for SeekLT), or less than it (for SeekGE).

This change updates filteringIter to guard for this case before returning from a seek call.

Found with #3044.

Informs cockroachdb/cockroach#113973, cockroachdb/cockroach#114056.

### DIFF
--- a/internal/keyspan/filter.go
+++ b/internal/keyspan/filter.go
@@ -4,6 +4,8 @@
 
 package keyspan
 
+import "github.com/cockroachdb/pebble/internal/base"
+
 // FilterFunc defines a transform from the input Span into the output Span. The
 // function returns true if the Span should be returned by the iterator, and
 // false if the Span should be skipped. The FilterFunc is permitted to mutate
@@ -21,6 +23,7 @@ type FilterFunc func(in *Span, out *Span) (keep bool)
 type filteringIter struct {
 	iter     FragmentIterator
 	filterFn FilterFunc
+	cmp      base.Compare
 
 	// span is a mutable Span passed to the filterFn. The filterFn is free to
 	// mutate this Span. The slice of Keys in the Span is reused with every call
@@ -32,18 +35,32 @@ var _ FragmentIterator = (*filteringIter)(nil)
 
 // Filter returns a new filteringIter that will filter the Spans from the
 // provided child iterator using the provided FilterFunc.
-func Filter(iter FragmentIterator, filter FilterFunc) FragmentIterator {
-	return &filteringIter{iter: iter, filterFn: filter}
+func Filter(iter FragmentIterator, filter FilterFunc, cmp base.Compare) FragmentIterator {
+	return &filteringIter{iter: iter, filterFn: filter, cmp: cmp}
 }
 
 // SeekGE implements FragmentIterator.
 func (i *filteringIter) SeekGE(key []byte) *Span {
-	return i.filter(i.iter.SeekGE(key), +1)
+	span := i.filter(i.iter.SeekGE(key), +1)
+	// i.filter could return a span that's less than key, _if_ the filterFunc
+	// (which has no knowledge of the seek key) mutated the span to end at a key
+	// less than or equal to `key`. Detect this case and next/invalidate the iter.
+	if span != nil && i.cmp(span.End, key) <= 0 {
+		return i.Next()
+	}
+	return span
 }
 
 // SeekLT implements FragmentIterator.
 func (i *filteringIter) SeekLT(key []byte) *Span {
-	return i.filter(i.iter.SeekLT(key), -1)
+	span := i.filter(i.iter.SeekLT(key), -1)
+	// i.filter could return a span that's >= key, _if_ the filterFunc (which has
+	// no knowledge of the seek key) mutated the span to start at a key greater
+	// than or equal to `key`. Detect this case and prev/invalidate the iter.
+	if span != nil && i.cmp(span.Start, key) >= 0 {
+		return i.Prev()
+	}
+	return span
 }
 
 // First implements FragmentIterator.
--- a/internal/keyspan/filter_test.go
+++ b/internal/keyspan/filter_test.go
@@ -67,7 +67,7 @@ func TestFilteringIter(t *testing.T) {
 				}
 			}
 			innerIter := NewIter(cmp, spans)
-			iter := Filter(innerIter, filter)
+			iter := Filter(innerIter, filter, cmp)
 			defer iter.Close()
 			s := runFragmentIteratorCmd(iter, td.Input, nil)
 			return s
--- a/internal/keyspan/testdata/truncate
+++ b/internal/keyspan/testdata/truncate
@@ -196,3 +196,123 @@ truncate g-g
 truncate g-h
 ----
 3:       gh
+
+# Regression test for https://github.com/cockroachdb/cockroach/issues/113973.
+
+truncate-and-save-iter a-dd
+----
+ok
+
+saved-iter
+first
+next
+next
+next
+----
+b-d:{(#1,RANGEDEL)}
+d-dd:{(#2,RANGEDEL)}
+<nil>
+<nil>
+
+saved-iter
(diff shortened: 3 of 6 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3049)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`internal/keyspan/filter.go` line 60 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3049#-NiozWSPEDfFEsNGzyDt:-NiozWSPEDfFEsNGzyDu:b-wcx8l6) ([raw file](https://github.com/cockroachdb/pebble/blob/48360e6b9996fdcb41ff43af7ec38941324a4212/internal/keyspan/filter.go#L60)):*
> ```Go
> 	// no knowledge of the seek key) mutated the span to start at a key greater
> 	// than or equal to `key`. Detect this case and prev/invalidate the iter.
> 	if span != nil && i.cmp(span.Start, key) >= 0 {
> ```

Should this be `> 0`? `span.Start` is inclusive, no?

RaduBerinde:
*[`internal/keyspan/filter.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3049#-Nip-71CDG-nKLg5kuy_:-Nip-71CDG-nKLg5kuya:bsmm2wr) ([raw file](https://github.com/cockroachdb/pebble/blob/48360e6b9996fdcb41ff43af7ec38941324a4212/internal/keyspan/filter.go#L49)):*
> ```Go
> 	// less than or equal to `key`. Detect this case and next/invalidate the iter.
> 	if span != nil && i.cmp(span.End, key) <= 0 {
> 		return i.Next()
> ```

Is it not possible for `i.Next()` to still be less than the key? Like if we are seeking to `a0` and we have keys `a1`, `a2` which the filter truncates to `a`?

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3049#-:-Nip1zkuBUFQIMH0GgdF:b-814g8k)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/keyspan/filter.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3049#-Nip-71CDG-nKLg5kuy_:-Nip1Tc6-fdBO5jzyIxm:byhpqzy) ([raw file](https://github.com/cockroachdb/pebble/blob/48360e6b9996fdcb41ff43af7ec38941324a4212/internal/keyspan/filter.go#L49)):*

Is it not possible for `i.Next()` to still be less than the key? Like if we are seeking to `a0` and we have keys `a1`, `a2` which the filter truncates to `a`?
</blockquote></details>

We're relying on the invariant that the underlying iter returned a span that was the correct `SeekGE/SeekLT` span, but filter excluded it, so a Next would take us to something that's strictly forward. Right?
___
*[`internal/keyspan/filter.go` line 60 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3049#-NiozWSPEDfFEsNGzyDt:-Nip1lg_9CvMEsqy-3OX:b-7f4vrp) ([raw file](https://github.com/cockroachdb/pebble/blob/48360e6b9996fdcb41ff43af7ec38941324a4212/internal/keyspan/filter.go#L60)):*

Should this be `> 0`? `span.Start` is inclusive, no?
</blockquote></details>

SeekLT is supposed to return a span that covers a key strictly less than the seek key, so `>=` is correct to catch the case that's wrong.

RaduBerinde:
*[`internal/keyspan/filter.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3049#-Nip-71CDG-nKLg5kuy_:-Nip2KGw-l8yg4IWfX9V:bsfu48) ([raw file](https://github.com/cockroachdb/pebble/blob/48360e6b9996fdcb41ff43af7ec38941324a4212/internal/keyspan/filter.go#L49)):*

We're relying on the invariant that the underlying iter returned a span that was the correct `SeekGE/SeekLT` span, but filter excluded it, so a Next would take us to something that's strictly forward. Right?
</blockquote></details>

Can't the filter exclude the next one too? I guess I don't understand what's special about the first key returned by Seek that doesn't apply to the following key

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3049)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`internal/keyspan/filter.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3049#-Nip-71CDG-nKLg5kuy_:-Nip7cH0-z_Gw8r6wJXy:b42kbxl) ([raw file](https://github.com/cockroachdb/pebble/blob/48360e6b9996fdcb41ff43af7ec38941324a4212/internal/keyspan/filter.go#L49)):*

Can't the filter exclude the next one too? I guess I don't understand what's special about the first key returned by Seek that doesn't apply to the following key
</blockquote></details>

I guess it's not possible with the truncate, because only the first span can intersect the upper bound.. but I'm less clear what is the `filter` contract in general, w.r.t modification of the span.

itsbilal (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3049)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`internal/keyspan/filter.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3049#-Nip-71CDG-nKLg5kuy_:-Nip8MWa4-8ybXWP_85x:butgwub) ([raw file](https://github.com/cockroachdb/pebble/blob/48360e6b9996fdcb41ff43af7ec38941324a4212/internal/keyspan/filter.go#L49)):*

I guess it's not possible with the truncate, because only the first span can intersect the upper bound.. but I'm less clear what is the `filter` contract in general, w.r.t modification of the span.
</blockquote></details>

Yeah. You're right that this covers the `Truncate` case, but doesn't generally solve a `FilterFunc` case for some contracts of FilterFunc that are currently completely not spelled out. We've filed #3048 to separate the two implementations anyway. so I'm okay with a Truncate-only fix for now.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3049#-:-Nip9EID8djE-fA7d11Q:bnfp4nl)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`internal/keyspan/filter.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3049#-Nip-71CDG-nKLg5kuy_:-Nip9C0B56ouZc919sVn:b-kyju7f) ([raw file](https://github.com/cockroachdb/pebble/blob/48360e6b9996fdcb41ff43af7ec38941324a4212/internal/keyspan/filter.go#L49)):*

Yeah. You're right that this covers the `Truncate` case, but doesn't generally solve a `FilterFunc` case for some contracts of FilterFunc that are currently completely not spelled out. We've filed #3048 to separate the two implementations anyway. so I'm okay with a Truncate-only fix for now.
</blockquote></details>

Awesome! Makes sense now!

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3049#-:-NipLeED3hwfZQE-QQXG:b-814g8k)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @jbowens)

PR #3051, opened 2023-11-09
sstable: add error injection unit test, fix NextPrefix bug

### LINKED ISSUE #3052: sstable: error swallowed in two-level NextPrefix
If the single-level iterator returns no KV due to an I/O error, the two-level iterator ignores the error and continues. Uncovered by #3051.

```
    random_test.go:111: error swallowed during NextPrefix("nua") with stack goroutine 22 [running]:
        runtime/debug.Stack()
        	/usr/local/go/src/runtime/debug/stack.go:24 +0x64
        github.com/cockroachdb/pebble/sstable.TestIterator_RandomErrors.func1({0x14000198000?, {0x0?, 0x14000198000?}, 0x14000034000?})
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/sstable/random_test.go:58 +0x50
        github.com/cockroachdb/pebble/vfs/errorfs.InjectorFunc.MaybeError(0x1400014f7d8?, {0x102c7c4a4?, {0x0?, 0x102952bec?}, 0x1400014f7e8?})
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/vfs/errorfs/errorfs.go:152 +0x38
        github.com/cockroachdb/pebble/vfs/errorfs.(*errorFile).ReadAt(0x14000288150, {0x600002c28020, 0x79, 0x79}, 0x1400014f828?)
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/vfs/errorfs/errorfs.go:452 +0x54
        github.com/cockroachdb/pebble/sstable.(*simpleReadable).ReadAt(0x1400014f848?, {0x10298a14c?, 0x2d08215834b62?}, {0x600002c28020?, 0x932fee80014f858?, 0x654d4b49?}, 0x1400014f858?)
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/sstable/reader.go:1215 +0x3c
        github.com/cockroachdb/pebble/objstorage.(*NoopReadHandle).ReadAt(0x1400014f8b8?, {0x1030c8c10?, 0x103518940?}, {0x600002c28020?, 0x1?, 0x140000df120?}, 0x1400019f400?)
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/objstorage/noop_readahead.go:24 +0x34
        github.com/cockroachdb/pebble/sstable.(*Reader).readBlock(0x140000ce000, {0x1030c8c10, 0x103518940}, {0x102da34c4?, 0x14000100b98?}, 0x0, {0x1030c8cf0, 0x14000288198}, 0x1400015a460, 0x0)
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/sstable/reader.go:557 +0x1c8
        github.com/cockroachdb/pebble/sstable.(*singleLevelIterator).loadBlock(0x14000100b00, 0x38?)
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/sstable/reader_iter_single_lvl.go:402 +0x214
        github.com/cockroachdb/pebble/sstable.(*singleLevelIterator).NextPrefix(0x14000100b00, {0x140000cdf38, 0x3, 0x17})
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/sstable/reader_iter_single_lvl.go:1187 +0x1ec
        github.com/cockroachdb/pebble/sstable.(*twoLevelIterator).NextPrefix(0x14000100b00, {0x140000cdf38, 0x3, 0x17})
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/sstable/reader_iter_two_lvl.go:810 +0x50
        github.com/cockroachdb/pebble/sstable.(*opGenerator).generateNextPrefix.func1({0x1030ccf70?, 0x14000100b00?})
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/sstable/random_test.go:171 +0x3c
        github.com/cockroachdb/pebble/sstable.TestIterator_RandomErrors(0x14000198820)
        	/Users/jackson/go/src/github.com/cockroachdb/pebble/sstable/random_test.go:105 +0xcf0
        testing.tRunner(0x14000198820, 0x1030be828)
        	/usr/local/go/src/testing/testing.go:1595 +0xe8
        created by testing.(*T).Run in goroutine 1
        	/usr/local/go/src/testing/testing.go:1648 +0x33c
```

jbowens:
Add a unit test that constructs sstables with random point keys using random
options and runs a random set of iterator operations against them. Randomly
inject errors into the VFS operations performed, and assert that if an error
was injected, it's also surfaced by the sstable.Iterator.

This test surfaced one new bug in #3052. This bug is fixed in the third commit.

Informs #1115.
Fixes #3052.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/cespare/xxhash/v2 v2.2.0
 	github.com/cockroachdb/datadriven v1.0.3-0.20230413201302-be42291fc80f
 	github.com/cockroachdb/errors v1.11.1
+	github.com/cockroachdb/metamorphic v0.0.0-20231108215700-4ba948b56895
 	github.com/cockroachdb/redact v1.1.5
 	github.com/cockroachdb/tokenbucket v0.0.0-20230807174530-cc333fc44b06
 	github.com/ghemawat/stream v0.0.0-20171120220530-696b145b53b9
@@ -18,7 +19,7 @@ require (
 	github.com/prometheus/client_golang v1.12.0
 	github.com/prometheus/client_model v0.2.1-0.20210607210712-147c58e9608a
 	github.com/spf13/cobra v1.0.0
-	github.com/stretchr/testify v1.8.1
+	github.com/stretchr/testify v1.8.4
 	golang.org/x/exp v0.0.0-20230626212559-97b1e661b5df
 	golang.org/x/perf v0.0.0-20230113213139-801c7ef9e5c5
 	golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4
--- a/go.sum
+++ b/go.sum
@@ -75,6 +75,8 @@ github.com/cockroachdb/errors v1.11.1 h1:xSEW75zKaKCWzR3OfxXUxgrk/NtT4G1MiOv5lWZ
 github.com/cockroachdb/errors v1.11.1/go.mod h1:8MUxA3Gi6b25tYlFEBGLf+D8aISL+M4MIpiWMSNRfxw=
 github.com/cockroachdb/logtags v0.0.0-20230118201751-21c54148d20b h1:r6VH0faHjZeQy818SGhaone5OnYfxFR/+AzdY3sf5aE=
 github.com/cockroachdb/logtags v0.0.0-20230118201751-21c54148d20b/go.mod h1:Vz9DsVWQQhf3vs21MhPMZpMGSht7O/2vFW2xusFUVOs=
+github.com/cockroachdb/metamorphic v0.0.0-20231108215700-4ba948b56895 h1:XANOgPYtvELQ/h4IrmPAohXqe2pWA8Bwhejr3VQoZsA=
+github.com/cockroachdb/metamorphic v0.0.0-20231108215700-4ba948b56895/go.mod h1:aPd7gM9ov9M8v32Yy5NJrDyOcD8z642dqs+F0CeNXfA=
 github.com/cockroachdb/redact v1.1.5 h1:u1PMllDkdFfPWaNGMyLD1+so+aq3uUItthCFqzwPJ30=
 github.com/cockroachdb/redact v1.1.5/go.mod h1:BVNblN9mBWFyMyqK1k3AAiSxhvhfK2oOZZ2lK+dpvRg=
 github.com/cockroachdb/tokenbucket v0.0.0-20230807174530-cc333fc44b06 h1:zuQyyAKVxetITBuuhv3BI9cMrmStnpT18zmgmTxunpo=
@@ -313,16 +315,12 @@ github.com/spf13/pflag v1.0.5/go.mod h1:McXfInJRrz4CZXVZOBLb0bTZqETkiAhM9Iw0y3An
 github.com/spf13/viper v1.4.0/go.mod h1:PTJ7Z/lr49W6bUbkmS1V3by4uWynFiR9p7+dSq/yZzE=
 github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
 github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
-github.com/stretchr/objx v0.4.0/go.mod h1:YvHI0jy2hoMjB+UWwv71VJQ9isScKT/TqJzVSSt89Yw=
-github.com/stretchr/objx v0.5.0/go.mod h1:Yh+to48EsGEfYuaHDzXPcE3xhTkx73EhmCGUpEOglKo=
 github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
 github.com/stretchr/testify v1.4.0/go.mod h1:j7eGeouHqKxXV5pUuKE4zz7dFj8WfuZ+81PSLYec5m4=
 github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
-github.com/stretchr/testify v1.7.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
-github.com/stretchr/testify v1.8.0/go.mod h1:yNjHg4UonilssWZ8iaSj1OCr/vHnekPRkoO+kdMU+MU=
-github.com/stretchr/testify v1.8.1 h1:w7B6lhMri9wdJUVmEZPGGhZzrYTPvgJArz7wNPgYKsk=
-github.com/stretchr/testify v1.8.1/go.mod h1:w2LPCIKwWwSfY2zedu0+kehJoqGctiVI29o6fzry7u4=
+github.com/stretchr/testify v1.8.4 h1:CcVxjf3Q8PM0mHUKJCdn+eZZtm5yQwehR5yeSVQQcUk=
+github.com/stretchr/testify v1.8.4/go.mod h1:sz/lmYIOXD/1dqDmKjjqLyZ2RngseejIcXlSw2iwfAo=
 github.com/tmc/grpc-websocket-proxy v0.0.0-20190109142713-0ad062ec5ee5/go.mod h1:ncp9v5uamzpCO7NfCPTXjqaC+bZgJeR0sMTm6dMHP7U=
 github.com/ugorji/go v1.1.4/go.mod h1:uQMGLiO92mf5W77hV/PUCpI3pbzQx3CRekS0kk+RGrc=
 github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2/go.mod h1:UETIi67q53MR2AWcXfiuqkDkRtnGDLqkBTpCHuJHxtU=
--- a/sstable/random_test.go
+++ b/sstable/random_test.go
@@ -1,0 +1,395 @@
+// Copyright 2023 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package sstable
+
+import (
+	"fmt"
+	"math/rand"
+	"runtime/debug"
+	"slices"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/cockroachdb/metamorphic"
+	"github.com/cockroachdb/pebble/bloom"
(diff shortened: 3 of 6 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3051)* status: 0 of 6 files reviewed, 4 unresolved discussions

___
*[`sstable/reader_iter_single_lvl.go` line 793 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3051#-NiqiOB88PEtkNA6ZsTh:-NiqiOB88PEtkNA6ZsTi:b-ko1wld) ([raw file](https://github.com/cockroachdb/pebble/blob/3448009d20f3c1324f3c3472008dc2498b387b29/sstable/reader_iter_single_lvl.go#L793)):*
> ```Go
> 			i.data.invalidate()
> 			return nil, base.LazyValue{}
> 		}
> ```

TODO: Why isn't this code path being hit?
___
*[`sstable/reader_iter_two_lvl.go` line 428 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3051#-Niqiy3E6jO0YjuSsAuz:-Niqiy3E6jO0YjuSsAv-:b-ko1wld) ([raw file](https://github.com/cockroachdb/pebble/blob/3448009d20f3c1324f3c3472008dc2498b387b29/sstable/reader_iter_two_lvl.go#L428)):*
> ```Go
> 			i.data.invalidate()
> 			return nil, base.LazyValue{}
> 		}
> ```

TODO: Why isn't this code path being hit?
___
*[`sstable/reader_iter_two_lvl.go` line 640 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3051#-Niqj5Lz0UwObpdV58Qj:-Niqj5Lz0UwObpdV58Qk:b-ko1wld) ([raw file](https://github.com/cockroachdb/pebble/blob/3448009d20f3c1324f3c3472008dc2498b387b29/sstable/reader_iter_two_lvl.go#L640)):*
> ```Go
> 		if result == loadBlockFailed {
> 			return nil, base.LazyValue{}
> 		}
> ```

TODO: Why isn't this code path being hit?
___
*[`sstable/reader_iter_two_lvl.go` line 809 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3051#-NiqjE3W70WR_bs6surC:-NiqjE3W70WR_bs6surD:bbf6pks) ([raw file](https://github.com/cockroachdb/pebble/blob/3448009d20f3c1324f3c3472008dc2498b387b29/sstable/reader_iter_two_lvl.go#L809)):*
> ```Go
> 	if i.err != nil {
> 		return nil, base.LazyValue{}
> 	}
> ```

TODO: I don't think this case is possible, because it's not valid to call NextPRefix on an exhausted iterator. Remove.

jbowens:
@itsbilal friendly ping :)

sumeerbhola (changes requested):
nice test!

> Reviewed 2 of 2 files at r1, 1 of 2 files at r2, 2 of 3 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3051#-:-NjJQHqj-fA_VuJ2ouiL:b-j6rd9r)* status: all files reviewed (commit messages unreviewed), 10 unresolved discussions (waiting on @jbowens)

___
*[`sstable/reader_iter_single_lvl.go` line 1157 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3051#-NjJPkvkAhmIjNx73fQu:-NjJPkvkAhmIjNx73fQv:b-hrlayt) ([raw file](https://github.com/cockroachdb/pebble/blob/1d992abb3763127631fb12c02cd9c64166c14856/sstable/reader_iter_single_lvl.go#L1157)):*
> ```Go
> 	if i.err != nil {
> 		// TODO(jackson): Can this case be turned into a panic? Once an error is
> 		// encountered, the iterator must be re-seeked.
> ```

why doesn't this code path get hit by the test?
___
*[`sstable/random_test.go` line 122 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3051#-NjJMVfL0heqmfZKNb6H:-NjJMVfL0heqmfZKNb6I:bch0g8o) ([raw file](https://github.com/cockroachdb/pebble/blob/75c897b9854607cdf7523e4d3cade91aeec9b8f6/sstable/random_test.go#L122)):*
> ```Go
> 	for i := 0; i < 1000; i++ {
> 		beforeCount := counter.Load()
> 		for op := nextOp(); !op(); {
> ```

should this be `op()` since we want to continue the loop if `op()` returns true?
This could use a code comment. It seems that `op()` return false when the iter is exhausted, and in that case we stay in this loop. But if it is not exhausted we continue? Is this because exhaustion can also be because of an error?

Also, shouldn't we be checking the `afterCount` stuff after each call to `op()`?
___
*[`sstable/random_test.go` line 132 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3051#-NjJMn_C3thq50XFPlJG:-NjJMn_C3thq50XFPlJH:blu7efz) ([raw file](https://github.com/cockroachdb/pebble/blob/75c897b9854607cdf7523e4d3cade91aeec9b8f6/sstable/random_test.go#L132)):*
> ```Go
> 		// injection. Then we can assert the results are identical.
> 
> 		if afterCount > beforeCount {
> ```

just making sure I understand: even if `afterCount == beforeCount` we could have an error because this was a Next() op and the iterator already had a saved error?
___
*[`sstable/random_test.go` line 263 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3051#-NjJM1rDEEhy_Ve2YIs0:-NjJM1rDEEhy_Ve2YIs1:b-mp4m7o) ([raw file](https://github.com/cockroachdb/pebble/blob/75c897b9854607cdf7523e4d3cade91aeec9b8f6/sstable/random_test.go#L263)):*
> ```Go
> 	if cfg.wopts == nil {
> 		cfg.wopts = &WriterOptions{
> 			TableFormat:             TableFormat(cfg.rng.Intn(int(TableFormatMax)) + 1),
> ```

so are are allowing `TableFormatLevelDB`?
___
*[`sstable/random_test.go` line 296 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3051#-NjJMAUU0YjM2_eTY4yH:-NjJMAUU0YjM2_eTY4yI:b-505fou) ([raw file](https://github.com/cockroachdb/pebble/blob/75c897b9854607cdf7523e4d3cade91aeec9b8f6/sstable/random_test.go#L296)):*
> ```Go
> 		{Item: base.InternalKeyKindDelete, Weight: 5},
> 		{Item: base.InternalKeyKindSingleDelete, Weight: 2},
> 		{Item: base.InternalKeyKindMerge, Weight: 1},
> ```

what about rangekeys and rangedels? Does that deserve a TODO?
___
*[`sstable/random_test.go` line 312 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3051#-NjJL6Fp9fr51Ab14D5A:-NjJL6Fp9fr51Ab14D5B:b-dwljma) ([raw file](https://github.com/cockroachdb/pebble/blob/75c897b9854607cdf7523e4d3cade91aeec9b8f6/sstable/random_test.go#L312)):*
> ```Go
> 	}
> 	keyMap := make(map[keyID]bool)
> 	// Constrain the space we generate keys to the middle 90% of the keyspace.
> ```

60% ? The divide by 5 suggests we are exluding the first 20% and last 20%.

jbowens (commented):
tftr!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3051#-:-NkH3wYo7fJlmdLypQJM:b-7ixx6c)* status: 2 of 6 files reviewed, 6 unresolved discussions (waiting on @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 1157 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3051#-NjJPkvkAhmIjNx73fQu:-NkH38-k85Xq6ULxvSBo:b-ltzjag) ([raw file](https://github.com/cockroachdb/pebble/blob/1d992abb3763127631fb12c02cd9c64166c14856/sstable/reader_iter_single_lvl.go#L1157)):*

why doesn't this code path get hit by the test?
</blockquote></details>

It's not valid to call `NextPrefix` on an invalid iterator (and the test doesn't know what `succKey` to produce), so `opRunner.runNextPrefix` refuses to run if it doesn't know the current position's user key.
___
*[`sstable/random_test.go` line 122 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3051#-NjJMVfL0heqmfZKNb6H:-NkGymbHARDsNsQOWASQ:b-9p184g) ([raw file](https://github.com/cockroachdb/pebble/blob/75c897b9854607cdf7523e4d3cade91aeec9b8f6/sstable/random_test.go#L122)):*

should this be `op()` since we want to continue the loop if `op()` returns true?
This could use a code comment. It seems that `op()` return false when the iter is exhausted, and in that case we stay in this loop. But if it is not exhausted we continue? Is this because exhaustion can also be because of an error?

Also, shouldn't we be checking the `afterCount` stuff after each call to `op()`?
</blockquote></details>

yeah, this wasn't clear. I updated with a comment describing what's happening—the func returned by `nextOp` returns false if the operation was not run because it's not valid on the current iterator state.
___
*[`sstable/random_test.go` line 132 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3051#-NjJMn_C3thq50XFPlJG:-NkH35yPEv2_IhKLEu8g:b-890fet) ([raw file](https://github.com/cockroachdb/pebble/blob/75c897b9854607cdf7523e4d3cade91aeec9b8f6/sstable/random_test.go#L132)):*

just making sure I understand: even if `afterCount == beforeCount` we could have an error because this was a Next() op and the iterator already had a saved error?
</blockquote></details>

yeah, that's right
___
*[`sstable/random_test.go` line 263 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3051#-NjJM1rDEEhy_Ve2YIs0:-NkGzE_fENd-pJ5s-QMs:b-dizk03) ([raw file](https://github.com/cockroachdb/pebble/blob/75c897b9854607cdf7523e4d3cade91aeec9b8f6/sstable/random_test.go#L263)):*

so are are allowing `TableFormatLevelDB`?
</blockquote></details>

yeah, as-is we are. any concerns?
___
*[`sstable/random_test.go` line 296 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3051#-NjJMAUU0YjM2_eTY4yH:-NkH1_4b2DWz7TNeXIaW:bjkcqu1) ([raw file](https://github.com/cockroachdb/pebble/blob/75c897b9854607cdf7523e4d3cade91aeec9b8f6/sstable/random_test.go#L296)):*

what about rangekeys and rangedels? Does that deserve a TODO?
</blockquote></details>

added TODOs here and within the test.
___
*[`sstable/random_test.go` line 312 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3051#-NjJL6Fp9fr51Ab14D5A:-NkH2YFz6yHpuyJPgAju:bc61n44) ([raw file](https://github.com/cockroachdb/pebble/blob/75c897b9854607cdf7523e4d3cade91aeec9b8f6/sstable/random_test.go#L312)):*

60% ? The divide by 5 suggests we are exluding the first 20% and last 20%.
</blockquote></details>

whoops, good catch. Fixed the code to constrain to the middle 90%.

PR #3054, opened 2023-11-10
internal/rangekey: fix invariant check with onlySets=false

itsbilal:
In the invariants build, with onlySets set to false (i.e. in ScanInternal), we would see RangeKeyDels in the coalescer that would sort to the end of the span.Keys when in BySuffixAsc order. This is expected and should not trip up the invariant panic, but it was.

### DIFF
--- a/internal/rangekey/coalesce.go
+++ b/internal/rangekey/coalesce.go
@@ -212,8 +212,10 @@ func (ui *UserIteratorConfig) ShouldDefragment(equal base.Equal, a, b *keyspan.S
 				b.Keys[i].Kind() != base.InternalKeyKindRangeKeySet) {
 				panic("pebble: unexpected non-RangeKeySet during defragmentation")
 			}
-			if i > 0 && (ui.comparer.Compare(a.Keys[i].Suffix, a.Keys[i-1].Suffix) < 0 ||
-				ui.comparer.Compare(b.Keys[i].Suffix, b.Keys[i-1].Suffix) < 0) {
+			// Don't do a comparison with RangeKeyDeletes, as those will always sort
+			// to the end and have no suffix.
+			if i > 0 && ((a.Keys[i].Kind() != base.InternalKeyKindRangeKeyDelete && ui.comparer.Compare(a.Keys[i].Suffix, a.Keys[i-1].Suffix) < 0) ||
+				(b.Keys[i].Kind() != base.InternalKeyKindRangeKeyDelete && ui.comparer.Compare(b.Keys[i].Suffix, b.Keys[i-1].Suffix) < 0)) {
 				panic("pebble: range keys not ordered by suffix during defragmentation")
 			}
 		}


itsbilal:
Found with #3044 .

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3054#-:-Nir4g_f3vc3YCoJmA-i:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @itsbilal)

___
*[`internal/rangekey/coalesce.go` line 216 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3054#-Nir4mpR3wepvKSOsPsT:-Nir4mpR3wepvKSOsPsU:b49amcy) ([raw file](https://github.com/cockroachdb/pebble/blob/ceb577b9097fe7e9a27aefdcf415702edba75096/internal/rangekey/coalesce.go#L216)):*
> ```Go
> 			}
> 			// Don't do a comparison with RangeKeyDeletes, as those will always sort
> 			// to the end and have no prefix.
> ```

s/prefix/suffix?

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3054#-:-NitqfLa6XbpMJQ3t6Sf:b-814g8k)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @aadityasondhi)

___
*[`internal/rangekey/coalesce.go` line 216 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3054#-Nir4mpR3wepvKSOsPsT:-NitqgZF8F0gdlO_bIIN:b-qsx7uj) ([raw file](https://github.com/cockroachdb/pebble/blob/ceb577b9097fe7e9a27aefdcf415702edba75096/internal/rangekey/coalesce.go#L216)):*

s/prefix/suffix?
</blockquote></details>

Done. Oops!

jbowens (dismissed):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3054)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @itsbilal)

___
*[`internal/rangekey/coalesce.go` line 218 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3054#-Nitr1Lx20ZAJR7DEdjt:-Nitr1Lx20ZAJR7DEdju:b-gvlo5c) ([raw file](https://github.com/cockroachdb/pebble/blob/ceb577b9097fe7e9a27aefdcf415702edba75096/internal/rangekey/coalesce.go#L218)):*
> ```Go
> 			// to the end and have no prefix.
> 			if i > 0 && ((a.Keys[i].Kind() != base.InternalKeyKindRangeKeyDelete && ui.comparer.Compare(a.Keys[i].Suffix, a.Keys[i-1].Suffix) < 0) ||
> 				(b.Keys[i].Kind() != base.InternalKeyKindRangeKeyDelete && ui.comparer.Compare(b.Keys[i].Suffix, b.Keys[i-1].Suffix) < 0)) {
> ```

Shouldn't `RangeKeyDelete`s have already been removed by `UserIteratorConfig.Transform`?

itsbilal:
*[`internal/rangekey/coalesce.go` line 218 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3054#-Nitr1Lx20ZAJR7DEdjt:-NitrSPs7R39pR6dBbPe:b-5wcneh) ([raw file](https://github.com/cockroachdb/pebble/blob/ceb577b9097fe7e9a27aefdcf415702edba75096/internal/rangekey/coalesce.go#L218)):*

Shouldn't `RangeKeyDelete`s have already been removed by `UserIteratorConfig.Transform`?
</blockquote></details>

Not if `onlySets=false`. See the above invariant conditional.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3054)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`internal/rangekey/coalesce.go` line 218 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3054#-Nitr1Lx20ZAJR7DEdjt:-NittT1O90KdfyRY1jC-:bn1d36p) ([raw file](https://github.com/cockroachdb/pebble/blob/ceb577b9097fe7e9a27aefdcf415702edba75096/internal/rangekey/coalesce.go#L218)):*

Not if `onlySets=false`. See the above invariant conditional.
</blockquote></details>

I see — can we update the comment at the beginning of `ShouldDefragment` and above it, and the comment above `UserIteratorConfig.Transform`?

I believe there's a bug with `!onlySets`: This function relies on the absence of `RangeKeyUnset`s and `RangeKeyDelete`s below (hence the original invariant above): If you have a span `a-b:{(#2,RANGEKEYSET,@5)}` and a span `b-c:{(#2,RANGEKEYUNSET,@5)}`, these two spans cannot be defragmented without losing information. If we defragment in the forward direction, we'll replace the second span's unset with a set. We should add a unit test for this case too.

Instead of using `UserIteratorConfig.ShouldDefragment` during scan internal, I think we should use a custom `DefragmentMethod` that includes a key kind comparison.

itsbilal:
*[`internal/rangekey/coalesce.go` line 218 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3054#-Nitr1Lx20ZAJR7DEdjt:-Niv6mMz-lv3JkqPdxYq:b-jllv9e) ([raw file](https://github.com/cockroachdb/pebble/blob/ceb577b9097fe7e9a27aefdcf415702edba75096/internal/rangekey/coalesce.go#L218)):*

I see — can we update the comment at the beginning of `ShouldDefragment` and above it, and the comment above `UserIteratorConfig.Transform`?

I believe there's a bug with `!onlySets`: This function relies on the absence of `RangeKeyUnset`s and `RangeKeyDelete`s below (hence the original invariant above): If you have a span `a-b:{(#2,RANGEKEYSET,@5)}` and a span `b-c:{(#2,RANGEKEYUNSET,@5)}`, these two spans cannot be defragmented without losing information. If we defragment in the forward direction, we'll replace the second span's unset with a set. We should add a unit test for this case too.

Instead of using `UserIteratorConfig.ShouldDefragment` during scan internal, I think we should use a custom `DefragmentMethod` that includes a key kind comparison.
</blockquote></details>

Good catch, this is problematic even though we haven't run into it yet. Filed https://github.com/cockroachdb/pebble/issues/3058 so we get to it asap.

PR #3060, opened 2023-11-10
db: return iterator construction errors eagerly

jbowens:
Previously iterator construction methods did not have a return value to surface errors (like ErrNotIndexed), so errors encountered during construction would be surfaced through the first iterator operation by installing an errorIter and/or errorKeyspanIter within the iterator stack. We now have error return values for all the iterator construction methods and can return these errors eagerly.

Similarly, the ScanInternal methods have error return values available and do not need to propagate iterator construction errors through adding fake iterators to the iterator stack.

For now, the errorIter and errorKeyspanIter types remain for a few test cases and to serve as "empty" iterator implementations when all a table's keys are excluded.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -1133,7 +1133,7 @@ func (b *Batch) NewIter(o *IterOptions) (*Iterator, error) {
 // tracing.
 func (b *Batch) NewIterWithContext(ctx context.Context, o *IterOptions) (*Iterator, error) {
 	if b.index == nil {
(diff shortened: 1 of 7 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3060#-:-NjJRcbnESKyeXzMfDUC:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @itsbilal)

PR #3065, opened 2023-11-13
sstable: remove valueBlockReader.ctx field

sumeerbhola:
blockProviderWhenOpen.readBlockForVBR no longer takes a context.Context parameter. There are two implementations of this interface:
- singleLevelIterator, which is used when the sstable iterator has not been closed. singleLevelIterator has a context to use in the call to Reader.readBlock.
- blockProviderWhenClosed, which is used when the sstable iterator has been closed. This is the rare case. We use the background context in this case. Given that valueBlockReader is not meant to outlive the root iterator in the iterator tree, and the context is set and propagated down from the root iterator, we could have kept the valueBlockReader.ctx field and set it in the corresponding sstable iterator's SetContext() implementation, but we choose to err on the side of safety for this rare case (and fewer structs with a context field is generally better).

Informs https://github.com/cockroachdb/cockroach/issues/114235

### DIFF
--- a/sstable/reader_iter_single_lvl.go
+++ b/sstable/reader_iter_single_lvl.go
@@ -218,7 +218,6 @@ func (i *singleLevelIterator) init(
 			// separated to their callers, they can put this valueBlockReader into a
 			// sync.Pool.
 			i.vbReader = &valueBlockReader{
-				ctx:    ctx,
 				bpOpen: i,
 				rp:     rp,
 				vbih:   r.valueBIH,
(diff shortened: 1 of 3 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3065#-:-Nj8rTfuC6at5u-dZo5y:bnfp4nl)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @sumeerbhola)

___
*[`sstable/value_block.go` line 752 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3065#-Nj8r6d_2nrUcdUq8peV:-Nj8r6d_2nrUcdUq8peW:bf8ukmv) ([raw file](https://github.com/cockroachdb/pebble/blob/62ee498d75309800259a38eced038609d63c7db9/sstable/value_block.go#L752)):*
> ```Go
> 	h BlockHandle, stats *base.InternalIteratorStats,
> ) (bufferHandle, error) {
> 	ctx := objiotracing.WithBlockType(context.Background(), objiotracing.ValueBlock)
> ```

[nit] Maybe add a comment explaining that we expect this to be rare (otherwise, we'd want a context with node/store tags set).

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3065#-:-Nj9LNEp6bKgTF6WF5OG:b-814g8k)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @aadityasondhi)

___
*[`sstable/value_block.go` line 752 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3065#-Nj8r6d_2nrUcdUq8peV:-Nj9LDCB9vHcW_0s-2QW:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/62ee498d75309800259a38eced038609d63c7db9/sstable/value_block.go#L752)):*

[nit] Maybe add a comment explaining that we expect this to be rare (otherwise, we'd want a context with node/store tags set).
</blockquote></details>

Done

PR #3067, opened 2023-11-13
db: add simplified Download() that does copy compactions

itsbilal:
This change adds a simplified Download() API implementation that uses any spare compaction concurrency to schedule copy compactions on the next external file in the download spans passed in. It also tracks for progress in `Download()` and returns when all external files are gone.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1671,6 +1671,19 @@ type readCompaction struct {
 	fileNum base.FileNum
 }
 
+type downloadSpan struct {
+	start []byte
+	end   []byte
+	// doneChans contains a list of channels passed into compactions as done
+	// channels. Each channel has a buffer size of 1 and is only passed into
+	// one compaction. This slice can grow over the lifetime of a downloadSpan.
+	doneChans []chan error
+	// compactionsStarted is the number of compactions started for this
+	// downloadSpan. Must be equal to len(doneChans)-1, i.e. there's one spare
+	// doneChan created each time a compaction starts up, for the next compaction.
+	compactionsStarted int
+}
+
 func (d *DB) addInProgressCompaction(c *compaction) {
 	d.mu.compact.inProgress[c] = struct{}{}
 	var isBase, isIntraL0 bool
@@ -2305,6 +2318,66 @@ func pickElisionOnly(picker compactionPicker, env compactionEnv) *pickedCompacti
 	return picker.pickElisionOnlyCompaction(env)
 }
 
+// maybeScheduleDownloadCompaction schedules a download compaction.
+//
+// Requires d.mu to be held.
+func (d *DB) maybeScheduleDownloadCompaction(env compactionEnv, maxConcurrentCompactions int) {
+	for len(d.mu.compact.downloads) > 0 && d.mu.compact.compactingCount < maxConcurrentCompactions {
+		v := d.mu.versions.currentVersion()
+		download := d.mu.compact.downloads[0]
+		env.inProgressCompactions = d.getInProgressCompactionInfoLocked(nil)
+		var externalFile *fileMetadata
+		var err error
+		var level int
+		for i := range v.Levels {
+			overlaps := v.Overlaps(i, d.cmp, download.start, download.end, true /* exclusiveEnd */)
+			iter := overlaps.Iter()
+			provider := d.objProvider
+			for f := iter.First(); f != nil; f = iter.Next() {
+				var objMeta objstorage.ObjectMetadata
+				objMeta, err = provider.Lookup(fileTypeTable, f.FileBacking.DiskFileNum)
+				if err != nil {
+					break
+				}
+				if objMeta.IsExternal() {
+					if f.IsCompacting() {
+						continue
+					}
+					externalFile = f
+					level = i
+					break
+				}
+			}
+			if externalFile != nil || err != nil {
+				break
+			}
+		}
+		if err != nil {
+			d.mu.compact.downloads = d.mu.compact.downloads[1:]
+			download.doneChans[download.compactionsStarted] <- err
+			continue
+		}
+		if externalFile == nil {
+			// The entirety of this span is downloaded, or is being downloaded right
+			// now. No need to schedule additional downloads for this span.
+			d.mu.compact.downloads = d.mu.compact.downloads[1:]
+			continue
+		}
+		pc := pickDownloadCompaction(v, d.opts, env, d.mu.versions.picker.getBaseLevel(), download, level, externalFile)
+		if pc != nil {
+			doneCh := download.doneChans[download.compactionsStarted]
+			download.compactionsStarted++
+			// Create another doneChan for the next compaction.
+			download.doneChans = append(download.doneChans, make(chan error, 1))
+
+			c := newCompaction(pc, d.opts, d.timeNow(), d.ObjProvider())
+			d.mu.compact.compactingCount++
+			d.addInProgressCompaction(c)
+			go d.compact(c, doneCh)
+		}
+	}
+}
+
 // maybeScheduleCompactionPicker schedules a compaction if necessary,
 // calling `pickFunc` to pick automatic compactions.
 //
@@ -2399,6 +2472,8 @@ func (d *DB) maybeScheduleCompactionPicker(
 		d.addInProgressCompaction(c)
 		go d.compact(c, nil)
 	}
+
+	d.maybeScheduleDownloadCompaction(env, maxConcurrentCompactions)
 }
 
 // deleteCompactionHintType indicates whether the deleteCompactionHint was
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -1906,6 +1906,38 @@ func pickManualCompaction(
 	return pc, false
 }
 
+func pickDownloadCompaction(
+	vers *version,
+	opts *Options,
+	env compactionEnv,
+	baseLevel int,
+	download *downloadSpan,
+	level int,
+	file *fileMetadata,
+) (pc *pickedCompaction) {
+	// Check if the file is compacting already. In that case we don't need to do
+	// anything as it'll be downloaded in the process.
+	if file.CompactionState == manifest.CompactionStateCompacting {
+		return nil
+	}
+	pc = newPickedCompaction(opts, vers, level, level, baseLevel)
+	pc.kind = compactionKindRewrite
+	pc.startLevel.files = manifest.NewLevelSliceKeySorted(opts.Comparer.Compare, []*fileMetadata{file})
+	if !pc.setupInputs(opts, env.diskAvailBytes, pc.startLevel) {
+		// setupInputs returned false indicating there's a conflicting
+		// concurrent compaction.
+		return nil
+	}
+	if pc.outputLevel.level != level {
+		panic("pebble: download compaction picked unexpected output level")
+	}
+	// Fail-safe to protect against compacting the same sstable concurrently.
+	if inputRangeAlreadyCompacting(env, pc) {
+		return nil
+	}
+	return pc
+}
+
 func (p *compactionPickerByScore) pickReadTriggeredCompaction(
 	env compactionEnv,
 ) (pc *pickedCompaction) {
--- a/db.go
+++ b/db.go
@@ -427,6 +427,9 @@ type DB struct {
 			// The list of manual compactions. The next manual compaction to perform
 			// is at the start of the list. New entries are added to the end.
 			manual []*manualCompaction
+			// downloads is the list of suggested download tasks. The next download to
+			// perform is at the start of the list. New entries are added to the end.
+			downloads []*downloadSpan
 			// inProgress is the set of in-progress flushes and compactions.
 			// It's used in the calculation of some metrics and to initialize L0
 			// sublevels' state. Some of the compactions contained within this
@@ -1894,6 +1897,108 @@ type DownloadSpan struct {
 	EndKey []byte
 }
 
+func (d *DB) downloadSpan(ctx context.Context, span DownloadSpan) error {
+	dSpan := &downloadSpan{
+		start: span.StartKey,
+		end:   span.EndKey,
+		// Protected by d.mu.
+		doneChans: make([]chan error, 1),
+	}
+	dSpan.doneChans[0] = make(chan error, 1)
+	doneChan := dSpan.doneChans[0]
+	compactionIdx := 0
+
+	func() {
+		d.mu.Lock()
+		defer d.mu.Unlock()
+
+		d.mu.compact.downloads = append(d.mu.compact.downloads, dSpan)
+		d.maybeScheduleCompaction()
+	}()
+
+	// Requires d.mu to be held.
+	noExternalFilesInSpan := func() (noExternalFiles bool) {
+		vers := d.mu.versions.currentVersion()
+
+		for i := 0; i < len(vers.Levels); i++ {
+			if vers.Levels[i].Empty() {
+				continue
+			}
+			overlap := vers.Overlaps(i, d.cmp, span.StartKey, span.EndKey, true /* exclusiveEnd */)
+			foundExternalFile := false
+			overlap.Each(func(metadata *manifest.FileMetadata) {
+				objMeta, err := d.objProvider.Lookup(fileTypeTable, metadata.FileBacking.DiskFileNum)
+				if err != nil {
+					return
+				}
+				if objMeta.IsExternal() {
+					foundExternalFile = true
+				}
+			})
+			if foundExternalFile {
+				return false
+			}
+		}
+		return true
+	}
+
+	// Requires d.mu to be held.
+	removeUsFromList := func() {
+		// Check where we are in d.mu.compact.downloads. Remove us from the
+		// list.
+		for i := range d.mu.compact.downloads {
+			if d.mu.compact.downloads[i] != dSpan {
+				continue
+			}
+			copy(d.mu.compact.downloads[i:], d.mu.compact.downloads[i+1:])
+			d.mu.compact.downloads = d.mu.compact.downloads[:len(d.mu.compact.downloads)-1]
+			break
+		}
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			d.mu.Lock()
+			defer d.mu.Unlock()
+			removeUsFromList()
+			return ctx.Err()
+		case err := <-doneChan:
+			if err != nil {
+				d.mu.Lock()
+				defer d.mu.Unlock()
+				removeUsFromList()
+				return err
+			}
+			compactionIdx++
+			// Grab the next doneCh to wait on.
+			func() {
+				d.mu.Lock()
+				defer d.mu.Unlock()
+				doneChan = dSpan.doneChans[compactionIdx]
+			}()
+		default:
+			doneSpan := func() bool {
+				d.mu.Lock()
+				defer d.mu.Unlock()
+				// It's possible to have downloaded all files without writing to any
+				// doneChans. This is expected if there are a significant amount
+				// of overlapping writes that schedule regular, non-download compactions.
+				if noExternalFilesInSpan() {
+					removeUsFromList()
+					return true
+				}
+				d.maybeScheduleCompaction()
+				d.mu.compact.cond.Wait()
+				return false
+			}()
+			if doneSpan {
+				return nil
+			}
+		}
+	}
+}
+
 // Download ensures that the LSM does not use any external sstables for the
 // given key ranges. It does so by performing appropriate compactions so that
 // all external data becomes available locally.
@@ -1908,7 +2013,23 @@ type DownloadSpan struct {
 // TODO(radu): consider passing a priority/impact knob to express how important
 // the download is (versus live traffic performance, LSM health).
 func (d *DB) Download(ctx context.Context, spans []DownloadSpan) error {
-	return errors.Errorf("not implemented")
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	if err := d.closed.Load(); err != nil {
+		panic(err)
+	}
+	if d.opts.ReadOnly {
+		return ErrReadOnly
+	}
+	for i := range spans {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		if err := d.downloadSpan(ctx, spans[i]); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 // Flush the memtable to stable storage.
--- a/ingest_test.go
+++ b/ingest_test.go
@@ -1698,6 +1698,8 @@ func TestIngestExternal(t *testing.T) {
 		// Disable automatic compactions because otherwise we'll race with
 		// delete-only compactions triggered by ingesting range tombstones.
 		opts.DisableAutomaticCompactions = true
+		lel := MakeLoggingEventListener(DefaultLogger)
+		opts.EventListener = &lel
 
 		var err error
 		d, err = Open("", opts)
@@ -1747,6 +1749,21 @@ func TestIngestExternal(t *testing.T) {
 				return "memtable flushed"
 			}
 			return ""
+
+		case "download":
+			if len(td.CmdArgs) != 2 {
+				panic("insufficient args for download command")
+			}
+			l := []byte(td.CmdArgs[0].Key)
+			r := []byte(td.CmdArgs[1].Key)
+			spans := []DownloadSpan{{StartKey: l, EndKey: r}}
+			ctx, cancel := context.WithTimeout(context.TODO(), 1*time.Minute)
+			defer cancel()
+			err := d.Download(ctx, spans)
+			if err != nil {
+				return err.Error()
+			}
+			return "ok"
 
 		case "get":
 			return runGetCmd(t, td, d)
--- a/testdata/ingest_external
+++ b/testdata/ingest_external
@@ -120,3 +120,35 @@ lsm
   000004:[a#10,DELSIZED-c#inf,RANGEDEL]
   000007:[c#11,DELSIZED-f#inf,RANGEDEL]
   000008:[f#12,DELSIZED-hh#inf,RANGEDEL]
+
+download a j
+----
+ok
+
+lsm
+----
+6:
+  000009:[a#0,SET-b#0,SET]
+  000010:[c#0,SET-e#0,SET]
+  000011:[f#0,SET-h#0,SET]
+
+iter
+first
+next
+next
+next
+next
+next
+next
+next
+next
+----
+a: (foo, .)
+b: (bar, .)
+c: (foobarbaz, .)
+d: (haha, .)
+e: (something, .)
+f: (foo, .)
+g: (foo, .)
+h: (foo, .)
+.


itsbilal:
Not the biggest fan of the structure of channels here. I'll give it more thought but otherwise this is ready for review.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3067)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`compaction.go` line 1669 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjIsofJ-i4YvnrQ78SU:-NjIsofJ-i4YvnrQ78SV:b-mv1yo6) ([raw file](https://github.com/cockroachdb/pebble/blob/3ee13e9bf097f9291e0866e974176843faeb8901/compaction.go#L1669)):*
> ```Go
> 	end   []byte
> 	level int
> 	done  chan error
> ```

We need to document the semantics of this channel, as they are non-trivial. IIUC we will be sending errors (normally `nil`), one per file in the span and then close the channel; a closed channel without a non-nil error having been sent means success.
___
*[`compaction.go` line 2425 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjIqft4E0cQeSd1QVGh:-NjIqft5Du6_DWgA2VJ7:bx59w9y) ([raw file](https://github.com/cockroachdb/pebble/blob/3ee13e9bf097f9291e0866e974176843faeb8901/compaction.go#L2425)):*
> ```Go
> 		}
> 		if err != nil {
> 			// Process other download spans.
> ```

This two-step handling is a bit hard to follow, consider doing just `break` above and doing the rest of the error handling here.
___
*[`compaction.go` line 2441 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjIrvss4F5qrkfCDdZh:-NjIrvss4F5qrkfCDdZi:b-7uln1h) ([raw file](https://github.com/cockroachdb/pebble/blob/3ee13e9bf097f9291e0866e974176843faeb8901/compaction.go#L2441)):*
> ```Go
> 			go d.compact(c, download.done)
> 		}
> 	}
> ```

On the next iteration here, won't we process `downloads[0]` again and find the same external file?
___
*[`compaction_picker.go` line 1919 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjIr_5KA5M4AONoQ31J:-NjIr_5KA5M4AONoQ31K:b-ly8za6) ([raw file](https://github.com/cockroachdb/pebble/blob/3ee13e9bf097f9291e0866e974176843faeb8901/compaction_picker.go#L1919)):*
> ```Go
> 	outputLevel := download.level
> 	if download.level < baseLevel && download.level != 0 {
> 		// The start level for a compaction must be >= Lbase. A download could have
> ```

I'm a little confused by this comment, what process would go inside the `downloadSpan` and change the level?
___
*[`db.go` line 1925 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjIrVH2-2Uk4v9-mB07:-NjIrVH2-2Uk4v9-mB08:b-l3ltzj) ([raw file](https://github.com/cockroachdb/pebble/blob/3ee13e9bf097f9291e0866e974176843faeb8901/db.go#L1925)):*
> ```Go
> 			}
> 			for j := range spans {
> 				dSpan := &downloadSpan{level: i, start: spans[j].StartKey, end: spans[j].EndKey, done: make(chan error, d.opts.MaxConcurrentCompactions())}
> ```

[nit] break up into multiple lines
___
*[`db.go` line 1983 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjIuU-W0bHOGK5Cr2Hh:-NjIuU-W0bHOGK5Cr2Hi:b-fdxe79) ([raw file](https://github.com/cockroachdb/pebble/blob/3ee13e9bf097f9291e0866e974176843faeb8901/db.go#L1983)):*
> ```Go
> 				case err, ok := <-addedSpans[i].done:
> 					if err != nil {
> 						allDone <- err
> ```

If we get more than 10 errors, this would block forever because we just exit below and never read from `allDone` again. I don't think we need this goroutine at all, we can do something like:
```
waitingSpans := addedSpans
for {
  if len(waitingSpans) == 0 {
    // All done.  
    return nil
  }
  select {
    case err, ok := <- waitingSpans[0].done:
    if err != nil {
      // TODO: maybe clean up unfinished spans.
      return err
    }
    if !ok {
      waitingSpans = waitingSpans[1:]
    }
    case <-ctx.Done():
    ..
  }
}
```

itsbilal (commented):
TFTR!


I need to write a test for the "pluck addedSpans out of d.mu.compact.download" logic that I just wrote, but otherwise this should be ready for a look.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3067#-:-NjJo0mbDpCgscGtYV0w:b-m8uu88)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @dt)

___
*[`compaction.go` line 1669 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjIsofJ-i4YvnrQ78SU:-NjJicuc0g1Iy-F2DjKt:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/3ee13e9bf097f9291e0866e974176843faeb8901/compaction.go#L1669)):*

We need to document the semantics of this channel, as they are non-trivial. IIUC we will be sending errors (normally `nil`), one per file in the span and then close the channel; a closed channel without a non-nil error having been sent means success.
</blockquote></details>

Done.
___
*[`compaction.go` line 2425 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjIqft4E0cQeSd1QVGh:-NjJid_59KRnZqHqMPKz:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/3ee13e9bf097f9291e0866e974176843faeb8901/compaction.go#L2425)):*

This two-step handling is a bit hard to follow, consider doing just `break` above and doing the rest of the error handling here.
</blockquote></details>

Done.
___
*[`compaction.go` line 2441 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjIrvss4F5qrkfCDdZh:-NjJjFXKC4ga8NpnCeTI:b-k2rwtz) ([raw file](https://github.com/cockroachdb/pebble/blob/3ee13e9bf097f9291e0866e974176843faeb8901/compaction.go#L2441)):*

On the next iteration here, won't we process `downloads[0]` again and find the same external file?
</blockquote></details>

Good catch. Fixed.
___
*[`compaction_picker.go` line 1919 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjIr_5KA5M4AONoQ31J:-NjJjHS6Af9DD5QkXxfU:bv4whyc) ([raw file](https://github.com/cockroachdb/pebble/blob/3ee13e9bf097f9291e0866e974176843faeb8901/compaction_picker.go#L1919)):*

I'm a little confused by this comment, what process would go inside the `downloadSpan` and change the level?
</blockquote></details>

Removed this as this is unnecessary now that a `downloadSpan` no longer has a level associated with it. But yes the idea was that we scheduled a download for, say, L4, and L4 got emptied out by the time we got to the download conditional in the compaction picker.
___
*[`db.go` line 1925 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjIrVH2-2Uk4v9-mB07:-NjJj_ct1-u2ubTlf-AT:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/3ee13e9bf097f9291e0866e974176843faeb8901/db.go#L1925)):*

[nit] break up into multiple lines
</blockquote></details>

Done.
___
*[`db.go` line 1983 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjIuU-W0bHOGK5Cr2Hh:-NjJnzrQ9eLXtCYgdBXL:b-qsvnon) ([raw file](https://github.com/cockroachdb/pebble/blob/3ee13e9bf097f9291e0866e974176843faeb8901/db.go#L1983)):*

If we get more than 10 errors, this would block forever because we just exit below and never read from `allDone` again. I don't think we need this goroutine at all, we can do something like:
```
waitingSpans := addedSpans
for {
  if len(waitingSpans) == 0 {
    // All done.  
    return nil
  }
  select {
    case err, ok := <- waitingSpans[0].done:
    if err != nil {
      // TODO: maybe clean up unfinished spans.
      return err
    }
    if !ok {
      waitingSpans = waitingSpans[1:]
    }
    case <-ctx.Done():
    ..
  }
}
```
</blockquote></details>

Done. PTAL!

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3067)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`compaction.go` line 2417 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjONLTD0YlhegPaRL39:-NjONLTD0YlhegPaRL3A:b-2a3fj6) ([raw file](https://github.com/cockroachdb/pebble/blob/a38bfe9e0c40443d69d93f22037a3fabfe3dbeaf/compaction.go#L2417)):*
> ```Go
> 			provider := d.objProvider
> 			for f := iter.First(); f != nil; f = iter.Next() {
> 				if f.IsCompacting() {
> ```

If a file is already compacting (internally triggered), we still need to wait for that compaction to finish, no? Plus, that compaction could fail..
___
*[`db.go` line 1933 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjOOfpuBYieYLakKoNf:-NjOOfpuBYieYLakKoNg:b9zl5y6) ([raw file](https://github.com/cockroachdb/pebble/blob/a38bfe9e0c40443d69d93f22037a3fabfe3dbeaf/db.go#L1933)):*
> ```Go
> 	}
> 
> 	noExternalFilesInSpan := func() bool {
> ```

[nit] InSpans
___
*[`db.go` line 1947 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjOOuJ62v2n6mGW2oyV:-NjOOuJ62v2n6mGW2oyW:b-bsiiwv) ([raw file](https://github.com/cockroachdb/pebble/blob/a38bfe9e0c40443d69d93f22037a3fabfe3dbeaf/db.go#L1947)):*
> ```Go
> 			for j := range spans {
> 				overlap := vers.Overlaps(i, d.cmp, spans[j].StartKey, spans[j].EndKey, true /* exclusiveEnd */)
> 				overlap.Each(func(metadata *manifest.FileMetadata) {
> ```

[nit] move `done` here as `foundExternalFile` then just do `if foundExternalFile { return false }`.
___
*[`db.go` line 1986 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjOPX1770w_1ZAT4d3u:-NjOPX1770w_1ZAT4d3v:b89gqq0) ([raw file](https://github.com/cockroachdb/pebble/blob/a38bfe9e0c40443d69d93f22037a3fabfe3dbeaf/db.go#L1986)):*
> ```Go
> 				d.mu.Lock()
> 				defer d.mu.Unlock()
> 				// Clean off as much of addedSpans as we can from d.mu.compact.downloads.
> ```

Maybe we can add a `cancel`/`noLongerNeeded` flag to to `downloadSpan` and check that in `addInProgressCompaction` instead of messing with the queue?
___
*[`db.go` line 1991 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjOPP2_-emYfuUBwUWG:-NjOPP2_-emYfuUBwUWH:bvndzdq) ([raw file](https://github.com/cockroachdb/pebble/blob/a38bfe9e0c40443d69d93f22037a3fabfe3dbeaf/db.go#L1991)):*
> ```Go
> 				// the list's head.
> 				//
> 				// This is necessary to avoid cases where stale download stick around
> ```

[nit] stale downloads

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3067#-:-NjOt8WI2ZctzqJP7rtF:b-814g8k)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @dt and @RaduBerinde)

___
*[`compaction.go` line 2417 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjONLTD0YlhegPaRL39:-NjOW8XtCI-TBWJyssUp:b-vjiswr) ([raw file](https://github.com/cockroachdb/pebble/blob/a38bfe9e0c40443d69d93f22037a3fabfe3dbeaf/compaction.go#L2417)):*

If a file is already compacting (internally triggered), we still need to wait for that compaction to finish, no? Plus, that compaction could fail..
</blockquote></details>

Good catch; I refactored most of this method so it should be better now to handle that case, as well as the outer Download() too. The concurrency stuff isn't fun. PTAL!
___
*[`db.go` line 1933 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjOOfpuBYieYLakKoNf:-NjOWMiZ3wTkTHYp6bfW:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/a38bfe9e0c40443d69d93f22037a3fabfe3dbeaf/db.go#L1933)):*

[nit] InSpans
</blockquote></details>

Done.
___
*[`db.go` line 1947 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjOOuJ62v2n6mGW2oyV:-NjOWgqyC0L5Qxdyq-wv:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/a38bfe9e0c40443d69d93f22037a3fabfe3dbeaf/db.go#L1947)):*

[nit] move `done` here as `foundExternalFile` then just do `if foundExternalFile { return false }`.
</blockquote></details>

Done.
___
*[`db.go` line 1986 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjOPX1770w_1ZAT4d3u:-NjOdgGn2fGr3is6SLhh:b-qu6n94) ([raw file](https://github.com/cockroachdb/pebble/blob/a38bfe9e0c40443d69d93f22037a3fabfe3dbeaf/db.go#L1986)):*

Maybe we can add a `cancel`/`noLongerNeeded` flag to to `downloadSpan` and check that in `addInProgressCompaction` instead of messing with the queue?
</blockquote></details>

Good point. Done.
___
*[`db.go` line 1991 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjOPP2_-emYfuUBwUWG:-NjOWjkLEZhBORyH0f35:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/a38bfe9e0c40443d69d93f22037a3fabfe3dbeaf/db.go#L1991)):*

[nit] stale downloads
</blockquote></details>

Done.

RaduBerinde (commented):
I think it would simplify things if instead of having one channel (with multiple senders) in each span, we would he have a structure (with its own mutex) with one instance for the `Download` call. It would store the first error encountered and a list of channels that need to be closed for the `Download` to be finished. We would add one of those for each `download span` in the queue, and one for each compaction.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3067#-:-NjP5ZlDB6_mar_KFz_1:b-v8n2uj)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`compaction.go` line 2417 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjONLTD0YlhegPaRL39:-NjP10kI1DqUYpTYkw7I:bwuug6t) ([raw file](https://github.com/cockroachdb/pebble/blob/a38bfe9e0c40443d69d93f22037a3fabfe3dbeaf/compaction.go#L2417)):*

Good catch; I refactored most of this method so it should be better now to handle that case, as well as the outer Download() too. The concurrency stuff isn't fun. PTAL!
</blockquote></details>

I still don't get how it works.. If the existing compaction was not triggered by the `Download` call, where in the code do we wait for that to finish? Feels like we need to tell the existing compaction to notify us when it's done.
___
*[`compaction.go` line 2754 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjP26lU7iMbw5uBOb2M:-NjP26lU7iMbw5uBOb2N:b-7tu33r) ([raw file](https://github.com/cockroachdb/pebble/blob/daa722577ee0a79c3003a70d59791930b57bb378/compaction.go#L2754)):*
> ```Go
> 
> // compact runs one compaction and maybe schedules another call to compact.
> func (d *DB) compact(c *compaction, errChannel chan error, errChanClosed *bool) {
> ```

I don't think we need this `errChanClosed`.. we don't need to close the channel if there are still compactions happening
___
*[`db.go` line 1922 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjP2SYK9IeFt0zMSmq8:-NjP2SYK9IeFt0zMSmq9:bvk0pnz) ([raw file](https://github.com/cockroachdb/pebble/blob/daa722577ee0a79c3003a70d59791930b57bb378/db.go#L1922)):*
> ```Go
> 				end:    spans[j].EndKey,
> 				done:   make(chan error, d.opts.MaxConcurrentCompactions()),
> 				closed: new(bool),
> ```

Since this is a pointer, they should all point to the same value, associated with the `Download` instance. Otherwise, it doesn't need to be a pointer.
___
*[`db.go` line 1935 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjP2vKZBWhyhpWUhrvi:-NjP2vK_50cf6QAjFzDi:b-k9117q) ([raw file](https://github.com/cockroachdb/pebble/blob/daa722577ee0a79c3003a70d59791930b57bb378/db.go#L1935)):*
> ```Go
> 
> 	// Requires db.mu *not* be held. However, it leaves db.mu locked *if*
> 	// noExternalFiles is true, relying on the caller to release it.
> ```

This kind of locking semantics are error-prone and an extra cognitive load. We should do the locking outside and the function should assume it's locked.
___
*[`db.go` line 1988 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjP6g4v8RXJkC5Kghop:-NjP6g4v8RXJkC5Kghoq:b-vfegi9) ([raw file](https://github.com/cockroachdb/pebble/blob/daa722577ee0a79c3003a70d59791930b57bb378/db.go#L1988)):*
> ```Go
> 			// It's possible to have downloaded all files without closing off the
> 			// relevant channels. This is expected if there are a significant amount
> 			// of overlapping writes.
> ```

How can this happen? Any downloads we started need to finish. The only time period where this could be true is in-between the last compaction and the call to `maybeScheduleDownloadCompaction` which will find no external files in the spans.
___
*[`db.go` line 1992 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjP35MVCbkSU312NIAT:-NjP35MVCbkSU312NIAU:b-sxskhe) ([raw file](https://github.com/cockroachdb/pebble/blob/daa722577ee0a79c3003a70d59791930b57bb378/db.go#L1992)):*
> ```Go
> 				// noExternalFilesInFirstSpan() has already locked d.mu. We need to
> 				// unlock it in this function.
> 				defer d.mu.Unlock()
> ```

`defer` will run at the very end of the function, not at the end of the `case` or `for` block

itsbilal:
*[`db.go` line 1992 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjP35MVCbkSU312NIAT:-NjPFjae1SW7RDiClQfc:b-uljmc8) ([raw file](https://github.com/cockroachdb/pebble/blob/daa722577ee0a79c3003a70d59791930b57bb378/db.go#L1992)):*

`defer` will run at the very end of the function, not at the end of the `case` or `for` block
</blockquote></details>

yeah, this is an artifact from a refactor in between. oops.

itsbilal (commented):
I'll consider that approach - that sounds like a good refactor. Thanks for describing! For now I've made some improvements to address the more egreriously-bad concurrency things that were present in the previous iteration. The last push also worked pretty well in a test run of an online restore.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3067#-:-NjSqmAiAO2RDQPqR4AI:b5w0w4d)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @dt and @RaduBerinde)

___
*[`compaction.go` line 2417 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjONLTD0YlhegPaRL39:-NjSpUm1EiQfFKdCuyyx:b-rag678) ([raw file](https://github.com/cockroachdb/pebble/blob/a38bfe9e0c40443d69d93f22037a3fabfe3dbeaf/compaction.go#L2417)):*

I still don't get how it works.. If the existing compaction was not triggered by the `Download` call, where in the code do we wait for that to finish? Feels like we need to tell the existing compaction to notify us when it's done.
</blockquote></details>

I've moved onto waiting on the compaction condition lock to clear out `waitingSpans` if that works, does this sound better?
___
*[`compaction.go` line 2754 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjP26lU7iMbw5uBOb2M:-NjSp_yl2OUmIEtwpqf9:b9ficyk) ([raw file](https://github.com/cockroachdb/pebble/blob/daa722577ee0a79c3003a70d59791930b57bb378/compaction.go#L2754)):*

I don't think we need this `errChanClosed`.. we don't need to close the channel if there are still compactions happening
</blockquote></details>

I think we still need to be able to tell the compaction goroutine that the download goroutine has disappeared, so it doesn't try writing to a channel that's full (and block). I can remove this though, now that we are better at having Download clear its own `downloadSpans`, we really only rely on this bool now in case the context gets cancelled and we need to bow out ASAP. What do you think?
___
*[`db.go` line 1922 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjP2SYK9IeFt0zMSmq8:-NjPH4NL8qbi71_gRq3E:bok1hlt) ([raw file](https://github.com/cockroachdb/pebble/blob/daa722577ee0a79c3003a70d59791930b57bb378/db.go#L1922)):*

Since this is a pointer, they should all point to the same value, associated with the `Download` instance. Otherwise, it doesn't need to be a pointer.
</blockquote></details>

Removed the pointer. Was an artifact from the time when I was keeping one bool for all spans in one download job.
___
*[`db.go` line 1935 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjP2vKZBWhyhpWUhrvi:-NjPGwti2YgjyAe5MqbD:b-v6s1o3) ([raw file](https://github.com/cockroachdb/pebble/blob/daa722577ee0a79c3003a70d59791930b57bb378/db.go#L1935)):*

This kind of locking semantics are error-prone and an extra cognitive load. We should do the locking outside and the function should assume it's locked.
</blockquote></details>

Removed. 
___
*[`db.go` line 1988 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjP6g4v8RXJkC5Kghop:-NjPGyArBkMoKZxvbSXp:b51853w) ([raw file](https://github.com/cockroachdb/pebble/blob/daa722577ee0a79c3003a70d59791930b57bb378/db.go#L1988)):*

How can this happen? Any downloads we started need to finish. The only time period where this could be true is in-between the last compaction and the call to `maybeScheduleDownloadCompaction` which will find no external files in the spans.
</blockquote></details>

This is possible if the external files landed in a higher level and/or there are lots of overlapping writes happening, so we just chose to do non-download compactions instead.

RaduBerinde (commented):
Ok with me to merge as-is for now if it helps with the prototype.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3067#-:-NjTpZhT0jmpxdw3KRoR:bkw7qah)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`compaction.go` line 2754 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjP26lU7iMbw5uBOb2M:-NjTlwKGCHf6PFcAzO65:b-1fc4tk) ([raw file](https://github.com/cockroachdb/pebble/blob/daa722577ee0a79c3003a70d59791930b57bb378/compaction.go#L2754)):*

I think we still need to be able to tell the compaction goroutine that the download goroutine has disappeared, so it doesn't try writing to a channel that's full (and block). I can remove this though, now that we are better at having Download clear its own `downloadSpans`, we really only rely on this bool now in case the context gets cancelled and we need to bow out ASAP. What do you think?
</blockquote></details>

Yeah I would remove it. I think we'll want to make Download wait for all compactions that have been passed one of its channels in any case (so things are more robust).
___
*[`compaction.go` line 1665 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjTmKc995rlKv0vO_9H:-NjTmKc995rlKv0vO_9I:bv16wlo) ([raw file](https://github.com/cockroachdb/pebble/blob/6a3eecab1e8f38a6270292f4e3434ff9c526a8a5/compaction.go#L1665)):*
> ```Go
> }
> 
> type downloadSpan struct {
> ```

I think it would make things easier if we made this `downloadTask` or something and have it hold all the `[]DownloadSpan`s. The one channel per span just complicates things.
___
*[`compaction.go` line 1673 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjTq6Xb957c15a0q-Ds:-NjTq6Xb957c15a0q-Dt:b2pykf4) ([raw file](https://github.com/cockroachdb/pebble/blob/6a3eecab1e8f38a6270292f4e3434ff9c526a8a5/compaction.go#L1673)):*
> ```Go
> 	// downloadSpan is expected to send a nil error into the channel. Zero or more
> 	// nil errors followed by a channel close denotes success.
> 	done chan error
> ```

It might make things easier if this was a list of channels, and we pass distinct channels to each compaction. Each channel would have buffer 1 so we wouldn't need about blocking if we exit early.
___
*[`compaction.go` line 2368 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjToP6I6sjMPHKh3Mt7:-NjToP6I6sjMPHKh3Mt8:b-isvp6d) ([raw file](https://github.com/cockroachdb/pebble/blob/6a3eecab1e8f38a6270292f4e3434ff9c526a8a5/compaction.go#L2368)):*
> ```Go
> 			// those compactions could write to it.
> 			if !externalFileIsCompacting {
> 				close(download.done)
> ```

Then we will never close this channel, no?

It would be simpler to just never close the channel since we're not relying on it anyway.

dt:
This was subsumed by https://github.com/cockroachdb/cockroach/pull/114742 right?

itsbilal (commented):
DR would like to see this merge soon so I cleaned it up + addressed most of the comments.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3067#-:-NkwR4Lh7RQRvRbo4d1U:bxem5au)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @dt and @RaduBerinde)

___
*[`compaction.go` line 2754 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjP26lU7iMbw5uBOb2M:-NkwRQ8FES_4oEYbo3_v:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/daa722577ee0a79c3003a70d59791930b57bb378/compaction.go#L2754)):*

Yeah I would remove it. I think we'll want to make Download wait for all compactions that have been passed one of its channels in any case (so things are more robust).
</blockquote></details>

Done.
___
*[`compaction.go` line 1665 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjTmKc995rlKv0vO_9H:-NkwR8aY3l58cfWRUmL6:b-ffqyge) ([raw file](https://github.com/cockroachdb/pebble/blob/6a3eecab1e8f38a6270292f4e3434ff9c526a8a5/compaction.go#L1665)):*

I think it would make things easier if we made this `downloadTask` or something and have it hold all the `[]DownloadSpan`s. The one channel per span just complicates things.
</blockquote></details>

We now do one channel per compaction which is definitely cleaner and easier to reason about. I tried a `downloadTask` approach but it was more complex, so I just took a page out of the manual compactions book and made a helper function that queues one span at a time and waits for it to finish. Plus it looks like we only call `Download` with one span on the CRDB side anyway.
___
*[`compaction.go` line 1673 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjTq6Xb957c15a0q-Ds:-NkwRO2Y0vomqduadWr-:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/6a3eecab1e8f38a6270292f4e3434ff9c526a8a5/compaction.go#L1673)):*

It might make things easier if this was a list of channels, and we pass distinct channels to each compaction. Each channel would have buffer 1 so we wouldn't need about blocking if we exit early.
</blockquote></details>

Done.
___
*[`compaction.go` line 2368 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3067#-NjToP6I6sjMPHKh3Mt7:-NkwRP6P8T1Lq9-KTypU:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/6a3eecab1e8f38a6270292f4e3434ff9c526a8a5/compaction.go#L2368)):*

Then we will never close this channel, no?

It would be simpler to just never close the channel since we're not relying on it anyway.
</blockquote></details>

Done.

RaduBerinde (approved):
:lgtm: with one question

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3067#-:-Nl0dQ3v1VpQfMU6bivB:b-cn9p1x)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`compaction.go` line 2366 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3067#-Nl0cGaA7UgfzcBavUws:-Nl0cGaA7UgfzcBavUwt:bd1y6ub) ([raw file](https://github.com/cockroachdb/pebble/blob/0cbe94fb587a119957249ae19ebb3d417bd48dc5/compaction.go#L2366)):*
> ```Go
> 			// now. No need to schedule additional downloads for this span.
> 			d.mu.compact.downloads = d.mu.compact.downloads[1:]
> 			// NB: We can't close download.done if `externalFileIsCompacting`, as
> ```

[nit] This comment is stale right?
___
*[`db.go` line 1986 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3067#-Nl0cyYaAoD7jXXvsebK:-Nl0cyYaAoD7jXXvsebL:b-59ynrz) ([raw file](https://github.com/cockroachdb/pebble/blob/0cbe94fb587a119957249ae19ebb3d417bd48dc5/db.go#L1986)):*
> ```Go
> 			}
> 			compactionIdx++
> 			// Grab the next doneCh to wait on.
> ```

Where are we checking if we got to the end of the list?

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3067#-:-NlPwA0MA9Dm7BjKDXdN:b-814g8k)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @dt and @RaduBerinde)

___
*[`compaction.go` line 2366 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3067#-Nl0cGaA7UgfzcBavUws:-NlPvheM9ycVTQKXLxve:bobclp9) ([raw file](https://github.com/cockroachdb/pebble/blob/0cbe94fb587a119957249ae19ebb3d417bd48dc5/compaction.go#L2366)):*

[nit] This comment is stale right?
</blockquote></details>

Good catch. Removed.
___
*[`db.go` line 1986 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3067#-Nl0cyYaAoD7jXXvsebK:-NlPvjMK78c_tx5KaHsN:bn3fgju) ([raw file](https://github.com/cockroachdb/pebble/blob/0cbe94fb587a119957249ae19ebb3d417bd48dc5/db.go#L1986)):*

Where are we checking if we got to the end of the list?
</blockquote></details>

There will always be one additional doneCh past the ones that have been handed to compactions, so indexing into the array and waiting on the next channel is safe. We rely on the `default` case below to detect the termination condition.

Issue #13, opened 2018-08-03
db: add support for the OPTIONS file


petermattis:
The RocksDB OPTIONS file provides an easy facility for determining the options that are set for a database. Supporting all of the RocksDB options is a non-starter, but providing a similar facility is useful for debugging/introspection purposes.

Issue #12, opened 2018-08-03
db: support arbitrary size write batches


petermattis:
LevelDB and RocksDB both support write batches that are larger than the configured memtable size because the memtable size is a soft-limit. The Pebble memtable size is a hard limit. Supporting write batches larger than the memtable size could be accomplished by allocating a custom-sized memtable when a write batch is larger than the memtable.

petermattis:
Another option: if a write batch is larger than the memtable size, we could index the write batch (if it isn't already) and add it as a memtable layer and flush it to disk immediately. This would avoid copying the memory, while not blocking the commit pipeline for anything other than the large write batch.

Issue #10, opened 2018-08-03
compaction: sstable output size


petermattis:
The compaction sstable output size is unimplemented. The `ShouldStopBefore` mechanism/heuristic used in both RocksDB and LevelDB needs to be implemented. This heuristic adjusts the output sstable size based on the number of bytes the sstable overlaps in the "grandparent" level (the level below the output sstable level). Care should be taken to ensure that this takes into account range deletions in addition to puts, merges and point deletions.

PR #3069, opened 2023-11-14
*: handle non-Sets correctly in UserIteratorConfig

### LINKED ISSUE #3058: internal/rangekey: fix defragmentation of unsets/dels with ScanInternal
Currently, ScanInternal creates a `UserIteratorConfig` for range keys with `onlySets` set to false to also surface range key unsets/dels. However, the coalescing / defragmentation logic there is flawed, and crucially does not check for key kinds, only suffixes and values. A set (of an empty value) followed by an adjoining unset for the same suffix would get defragmented, losing information from the kind of one of those two spans.

To fix this, a variant of UserIteratorConfig could be used that correctly defragments all key kinds, similar to compactions, instead of relying on `onlySets=false`.

See comment in https://github.com/cockroachdb/pebble/pull/3054#pullrequestreview-1725093103 for context and details.

itsbilal:
This change moves the `scanInternalIterator` use case of
UserIteratorConfig over to something that's closer to
rangeKeyCompactionTransform, where internal keys are coalesced
using rangekey.coalesce, key order is maintained at ByTrailerDesc,
and defragmentation happens by internal key using
keyspan.DefragmentInternal. The onlySets var is renamed to internalKeys
and its meaning is reversed.

Fixes #3058.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -426,14 +426,18 @@ func runBatchDefineCmd(d *datadriven.TestData, b *Batch) error {
 			}
 			err = b.Merge([]byte(parts[1]), []byte(parts[2]), nil)
 		case "range-key-set":
-			if len(parts) != 5 {
-				return errors.Errorf("%s expects 4 arguments", parts[0])
+			if len(parts) < 4 || len(parts) > 5 {
+				return errors.Errorf("%s expects 3 or 4 arguments", parts[0])
+			}
+			var val []byte
+			if len(parts) == 5 {
+				val = []byte(parts[4])
 			}
 			err = b.RangeKeySet(
 				[]byte(parts[1]),
 				[]byte(parts[2]),
 				[]byte(parts[3]),
-				[]byte(parts[4]),
+				val,
 				nil)
 		case "range-key-unset":
 			if len(parts) != 4 {
--- a/external_iterator.go
+++ b/external_iterator.go
@@ -312,7 +312,7 @@ func finishInitializingExternal(ctx context.Context, it *Iterator) {
 					base.InternalKeySeqNumMax,
 					it.opts.LowerBound, it.opts.UpperBound,
 					&it.hasPrefix, &it.prefixOrFullSeekKey,
-					true /* onlySets */, &it.rangeKey.internal,
+					false /* internalKeys */, &it.rangeKey.internal,
 				)
 				for i := range rangeKeyIters {
 					it.rangeKey.iterConfig.AddLevel(rangeKeyIters[i])
--- a/internal/rangekey/coalesce.go
+++ b/internal/rangekey/coalesce.go
@@ -52,15 +52,15 @@ import (
 //	                        │
 //	                        ╰── <?>.Next
 type UserIteratorConfig struct {
-	snapshot   uint64
-	comparer   *base.Comparer
-	miter      keyspan.MergingIter
-	biter      keyspan.BoundedIter
-	diter      keyspan.DefragmentingIter
-	liters     [manifest.NumLevels]keyspan.LevelIter
-	litersUsed int
-	onlySets   bool
-	bufs       *Buffers
+	snapshot     uint64
+	comparer     *base.Comparer
+	miter        keyspan.MergingIter
+	biter        keyspan.BoundedIter
+	diter        keyspan.DefragmentingIter
+	liters       [manifest.NumLevels]keyspan.LevelIter
+	litersUsed   int
+	internalKeys bool
+	bufs         *Buffers
 }
 
 // Buffers holds various buffers used for range key iteration. They're exposed
@@ -79,9 +79,10 @@ func (bufs *Buffers) PrepareForReuse() {
 
 // Init initializes the range key iterator stack for user iteration. The
 // resulting fragment iterator applies range key semantics, defragments spans
-// according to their user-observable state and, if onlySets = true, removes all
+// according to their user-observable state and, if !internalKeys, removes all
 // Keys other than RangeKeySets describing the current state of range keys. The
-// resulting spans contain Keys sorted by Suffix.
+// resulting spans contain Keys sorted by suffix (unless internalKeys is true,
+// in which case they remain sorted by trailer descending).
 //
 // The snapshot sequence number parameter determines which keys are visible. Any
 // keys not visible at the provided snapshot are ignored.
@@ -91,16 +92,20 @@ func (ui *UserIteratorConfig) Init(
 	lower, upper []byte,
 	hasPrefix *bool,
 	prefix *[]byte,
-	onlySets bool,
+	internalKeys bool,
 	bufs *Buffers,
 	iters ...keyspan.FragmentIterator,
 ) keyspan.FragmentIterator {
 	ui.snapshot = snapshot
 	ui.comparer = comparer
-	ui.onlySets = onlySets
+	ui.internalKeys = internalKeys
 	ui.miter.Init(comparer.Compare, ui, &bufs.merging, iters...)
 	ui.biter.Init(comparer.Compare, comparer.Split, &ui.miter, lower, upper, hasPrefix, prefix)
-	ui.diter.Init(comparer, &ui.biter, ui, keyspan.StaticDefragmentReducer, &bufs.defragmenting)
+	if internalKeys {
+		ui.diter.Init(comparer, &ui.biter, keyspan.DefragmentInternal, keyspan.StaticDefragmentReducer, &bufs.defragmenting)
+	} else {
+		ui.diter.Init(comparer, &ui.biter, ui, keyspan.StaticDefragmentReducer, &bufs.defragmenting)
+	}
 	ui.litersUsed = 0
 	ui.bufs = bufs
 	return &ui.diter
@@ -136,7 +141,8 @@ func (ui *UserIteratorConfig) SetBounds(lower, upper []byte) {
 // of unset keys, removal of keys overwritten by a set at the same suffix, etc)
 // and then non-RangeKeySet keys are removed. The resulting transformed spans
 // only contain RangeKeySets describing the state visible at the provided
-// sequence number, and hold their Keys sorted by Suffix.
+// sequence number, and hold their Keys sorted by Suffix (except if internalKeys
+// is true, then keys remain sorted by trailer.
 func (ui *UserIteratorConfig) Transform(cmp base.Compare, s keyspan.Span, dst *keyspan.Span) error {
 	// Apply shadowing of keys.
 	dst.Start = s.Start
@@ -148,9 +154,16 @@ func (ui *UserIteratorConfig) Transform(cmp base.Compare, s keyspan.Span, dst *k
 	if err := coalesce(ui.comparer.Equal, &ui.bufs.sortBuf, ui.snapshot, s.Keys); err != nil {
 		return err
 	}
-	// During user iteration over range keys, unsets and deletes don't matter.
-	// Remove them if onlySets = true. This step helps logical defragmentation
-	// during iteration.
+	if ui.internalKeys {
+		if s.KeysOrder != keyspan.ByTrailerDesc {
+			panic("unexpected key ordering in UserIteratorTransform with internalKeys = true")
+		}
+		dst.Keys = ui.bufs.sortBuf.Keys
+		keyspan.SortKeysByTrailer(&dst.Keys)
+		return nil
+	}
+	// During user iteration over range keys, unsets and deletes don't matter. This
+	// step helps logical defragmentation during iteration.
 	keys := ui.bufs.sortBuf.Keys
 	dst.Keys = dst.Keys[:0]
 	for i := range keys {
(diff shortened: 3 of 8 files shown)


jbowens (commented):
> Reviewed 3 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3069)* status: 2 of 8 files reviewed, 8 unresolved discussions (waiting on @itsbilal)

___
*[`scan_internal.go` line 398 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJS8lw5aGeo4Yles41:-NjJS8lw5aGeo4Yles42:b-ax5d0u) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal.go#L398)):*
> ```Go
> // scanInternalRangeKeyIter holds rangeKey-relevant state for a
> // scanInternalIterator.
> type scanInternalRangeKeyIter struct {
> ```

Another approach would be to adapt the `UserIteratorConfig` adding a new initialization method that just installs a separate `Transformer` and `DefragmentingMethod`. That might be a little clearer, thoughts?
___
*[`scan_internal.go` line 413 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJOZ_MAAAELSla_qFL:-NjJOZ_MAAAELSla_qFM:b1i2pqp) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal.go#L413)):*
> ```Go
> func (s *scanInternalRangeKeyIter) Transform(
> 	cmp base.Compare, in keyspan.Span, out *keyspan.Span,
> ) error {
> ```

Can we copy and edit `Transform` from `UserIteratorConfig`?  It'll avoid an allocation (assuming we stash a `keyspan.KeysBySuffix` on `scanInternalRangeKeyIter` too), and if we're going to maintain a parallel implementation, it's worthwhile to mirror them as much as possible.
___
*[`scan_internal.go` line 414 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJPSzz7zjMeKUigNQh:-NjJPT--6YifYcU5DP17:bhmigkp) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal.go#L414)):*
> ```Go
> 	cmp base.Compare, in keyspan.Span, out *keyspan.Span,
> ) error {
> 	// in.Keys are in descending seqnum order. Call rangekey.Coalesce on the
> ```

Let's assert that the key order (see `(keyspan.Span).KeysOrder`) wherever we rely on it
___
*[`scan_internal.go` line 420 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJOsQ_B1R-5F7vzX6M:-NjJOsQ_B1R-5F7vzX6N:bz7kh0k) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal.go#L420)):*
> ```Go
> 	out.Keys = out.Keys[:0]
> 	start := 0
> 	for start < len(in.Keys) && !base.Visible(in.Keys[start].SeqNum(), s.snapshot, base.InternalKeySeqNumMax) {
> ```

This works as a way to filter visible keys only because scan internal doesn't support reading through batches, which is a bit subtle.
___
*[`scan_internal.go` line 425 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJO68wC9cMsGUPzOKf:-NjJO68wC9cMsGUPzOKg:buxn3mw) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal.go#L425)):*
> ```Go
> 	}
> 	if start < len(in.Keys) {
> 		keysDst := out.Keys[:cap(out.Keys)]
> ```

Is this necessary? `Coalesce` immediately reslices `keysDst` to `[:0]`.
___
*[`testdata/scan_internal` line 154 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJTigx-6u0aBuWnqvS:-NjJTigx-6u0aBuWnqvT:b-7fe43d) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/testdata/scan_internal#L154)):*
> ```
> ----
> a-b:{(#13,RANGEKEYDEL)}
> b-c:{(#12,RANGEKEYUNSET,@6) (#10,RANGEKEYSET,@8,foo)}
> ```

In order to test the original bug, I think we need the `RANGEKEYSET` to be value-less (which is how always are in Cockroach too).
___
*[`scan_internal_test.go` line 22 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJT9-M7hv3WxVhbYWO:-NjJT9-M7hv3WxVhbYWP:bo5b6n4) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal_test.go#L22)):*
> ```Go
> 	"github.com/cockroachdb/pebble/internal/keyspan"
> 	"github.com/cockroachdb/pebble/internal/private"
> 	rkeyinternal "github.com/cockroachdb/pebble/internal/rangekey"
> ```

I think we can get rid of this file's import of `github.com/cockroachdb/pebble/rangekey` by updating references to directly reference the underlying type `keyspan.Key`, and then we don't need the package alias here.
___
*[`scan_internal_test.go` line 404 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJTU1vD8ajr9LySG7m:-NjJTU1vD8ajr9LySG7n:biuk6bx) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal_test.go#L404)):*
> ```Go
> 				}()
> 			} else if ingest {
> 				points, rangeDels, rangeKeys := private.BatchSort(b)
> ```

Since we're within the main pebble package, we can use `batchSort` directly rather than going through the weird `internal/private` hook.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3069#-:-NjP0Zn66cBaaXGECFQP:b-814g8k)* status: 2 of 8 files reviewed, 8 unresolved discussions (waiting on @jbowens)

___
*[`scan_internal.go` line 398 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJS8lw5aGeo4Yles41:-NjOu4vbCyvo9iE-56Lj:bxmyrln) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal.go#L398)):*

Another approach would be to adapt the `UserIteratorConfig` adding a new initialization method that just installs a separate `Transformer` and `DefragmentingMethod`. That might be a little clearer, thoughts?
</blockquote></details>

That's one approach and it has the argument going for it that it's a more minimal (and therefore more backportable if we choose to do that) change. I just went with something lighter-weight so we don't shoot ourselves with more unintended UserIteratorConfig-related things (such as the BySuffixAsc sorting) that we really don't need. What do you think?
___
*[`scan_internal.go` line 413 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJOZ_MAAAELSla_qFL:-NjOu9nf66y3-NDO4FHF:bv868v7) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal.go#L413)):*

Can we copy and edit `Transform` from `UserIteratorConfig`?  It'll avoid an allocation (assuming we stash a `keyspan.KeysBySuffix` on `scanInternalRangeKeyIter` too), and if we're going to maintain a parallel implementation, it's worthwhile to mirror them as much as possible.
</blockquote></details>

I actually did copy and edit `rangeKeyCompactionTransform` which this should mimic 1-1 save for the optimization where we're not concerned about snapshot stripes, rather only the one snapshot seqNum. Maybe the change we need is to just pool `scanInternalRangeKeyIter` or the whole iterator?
___
*[`scan_internal.go` line 414 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJPSzz7zjMeKUigNQh:-NjOuXhjAZfylXZe4cOl:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal.go#L414)):*

Let's assert that the key order (see `(keyspan.Span).KeysOrder`) wherever we rely on it
</blockquote></details>

Done.
___
*[`scan_internal.go` line 420 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJOsQ_B1R-5F7vzX6M:-NjOzhWU4_7RabE-Y0zY:bv8rvxb) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal.go#L420)):*

This works as a way to filter visible keys only because scan internal doesn't support reading through batches, which is a bit subtle.
</blockquote></details>

Left a comment about this, but yes.
___
*[`scan_internal.go` line 425 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJO68wC9cMsGUPzOKf:-NjOuks80XINfPVzrSJT:b-hhbkm3) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal.go#L425)):*

Is this necessary? `Coalesce` immediately reslices `keysDst` to `[:0]`.
</blockquote></details>

Copy-paste artifact as `rangeKeyCompactionTransform` does the same thing. Removed.
___
*[`testdata/scan_internal` line 154 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJTigx-6u0aBuWnqvS:-NjOzk1A9Uc5nEzJa3eo:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/testdata/scan_internal#L154)):*

In order to test the original bug, I think we need the `RANGEKEYSET` to be value-less (which is how always are in Cockroach too).
</blockquote></details>

Done.
___
*[`scan_internal_test.go` line 22 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJT9-M7hv3WxVhbYWO:-NjP-dK280sYbxKC0tbA:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal_test.go#L22)):*

I think we can get rid of this file's import of `github.com/cockroachdb/pebble/rangekey` by updating references to directly reference the underlying type `keyspan.Key`, and then we don't need the package alias here.
</blockquote></details>

Done.
___
*[`scan_internal_test.go` line 404 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJTU1vD8ajr9LySG7m:-NjP-h5W4swyH5p5EpWh:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal_test.go#L404)):*

Since we're within the main pebble package, we can use `batchSort` directly rather than going through the weird `internal/private` hook.
</blockquote></details>

Done.

itsbilal (commented):
oops, didn't push. pushing now

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3069#-:-NjPF9lp0JbT64K7Lv_B:b-6f5lsm)* status: 2 of 9 files reviewed, 8 unresolved discussions (waiting on @jbowens)

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3069)* status: 2 of 9 files reviewed, 2 unresolved discussions (waiting on @itsbilal)

___
*[`scan_internal.go` line 398 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJS8lw5aGeo4Yles41:-NjT4IaC0wN4eV0j6wDK:blwrx9c) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal.go#L398)):*

That's one approach and it has the argument going for it that it's a more minimal (and therefore more backportable if we choose to do that) change. I just went with something lighter-weight so we don't shoot ourselves with more unintended UserIteratorConfig-related things (such as the BySuffixAsc sorting) that we really don't need. What do you think?
</blockquote></details>

But the `BySuffixAsc` is happening here too, it's just hidden within `Coalesce` (which is also the origin of the additional allocation mentioned below because `rangekey.Coalesce` doesn't have access to `UserIteratorConfig.bufs.sortBuf`. I think the lighter weight and more maintainable change is to adapt the `UserIteratorConfig`

itsbilal (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3069)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`scan_internal.go` line 398 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3069#-NjJS8lw5aGeo4Yles41:-NjUJ_-ZDEWAW01-A-Jd:blp1e3i) ([raw file](https://github.com/cockroachdb/pebble/blob/dcb84e07c25aae0e6de9f94eb91f553061289884/scan_internal.go#L398)):*

But the `BySuffixAsc` is happening here too, it's just hidden within `Coalesce` (which is also the origin of the additional allocation mentioned below because `rangekey.Coalesce` doesn't have access to `UserIteratorConfig.bufs.sortBuf`. I think the lighter weight and more maintainable change is to adapt the `UserIteratorConfig`
</blockquote></details>

Done. PR has shrunk now as a result, PTAL!

jbowens (approved):
:lgtm_strong: 

> Reviewed 3 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3069#-:-NjsQxUy9QILoor1aI4z:b-uwkowd)* status: 3 of 9 files reviewed, all discussions resolved

itsbilal:
TFTR!

PR #3071, opened 2023-11-15
internal/manifest: InitProviderBacking in VersionEdit decoding

dt:


itsbilal:
The comment on `ve.Decode` describes that it's the caller's job to set the `FileBacking`, and the manifest load code seems to do this correctly. It also looks like we do set the file backing correctly when ingesting external files. I'll have to dig deeper into where we tripped up this issue, as something else is likely going wrong.

dt:
Huh, why is it init’ing the physical backing here then?

dt:
Anyway, I see where to init it in tools/lsm.go, I’ll send that out in a bit.

dt:
Did that over in https://github.com/cockroachdb/pebble/pull/3073

PR #3073, opened 2023-11-15
tool/lsm: init provider backing for decoded version edits

dt:


tool/lsm.go
```diff
@@ -288,6 +288,9 @@ func (l *lsmT) buildEdits(edits []*manifest.VersionEdit) error {
 
 		for j := range ve.NewFiles {
 			nf := &ve.NewFiles[j]
+			if nf.Meta.Virtual {
+				nf.Meta.InitProviderBacking(nf.BackingFileNum)
```
itsbilal, line 292: I think what you want to do is to have a running map of BackingFileNums -> FileBackings, populated by `ve.CreatedBackingTables`, and then set `nf.Meta.FileBacking = backingsMap[nf.BackingFileNum]` here.

It's possible the backing is a non-provider backing, eg. virtual sstables backed by a local sstable (created due to an ingest split), in which case this won't work.

dt: done

tool/lsm.go
```diff
@@ -288,6 +293,9 @@ func (l *lsmT) buildEdits(edits []*manifest.VersionEdit) error {
 
 		for j := range ve.NewFiles {
 			nf := &ve.NewFiles[j]
+			if b, ok := backings[nf.BackingFileNum]; ok {
```
itsbilal, line 296: nit: probably helpful to throw this behind an `if nf.Meta.Virtual` just to avoid some zero value `nf.BackingFileNum` cases. but also alright as-is.

itsbilal (approved):
LGTM

tool/lsm.go
dt: done.

PR #3074, opened 2023-11-15
*: shared ingestion fixes

itsbilal:
Contains 3 commits, each of which contains some fixes around shared/foreign sstables.

### *: set hideObsoletePoints=true for shared foreign sstables

Previously, we weren't setting hideObsoletePoints to true for
compaction iters created out of shared foreign sstables. This
change addresses that by passing that state through virtualState
in VirtualReader through to the old Reader.

Unblocks https://github.com/cockroachdb/pebble/pull/3044.

### ingest: don't call checkVirtualBounds() in ingestion

Previously, we were calling checkVirtualBounds on FileMetadatas
before they were installed. This would lead to panics in invariant
builds as we would try to load an uninstalled FileMetadata
from the table cache.

This change removes the checkVirtualBounds call, making it a test-only
thing. Adding a check after version installation also doesn't guarantee
that refs will be nonzero, as a compaction immediately after ingestApply
is possible.

Only an issue with invariants builds. Unblocks https://github.com/cockroachdb/pebble/pull/3044.

### scan_internal: Sort keys by kind before calling visitor

Previously, we were passing range keys that were sorted by trailer
descending right from ScanInternal. However, the caller isn't
interested in seqnum-based sorting, only kind-based trailer sorting.
This change zeroes seqnums before sorting the keys.

Unblocks https://github.com/cockroachdb/pebble/pull/3044.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -1155,7 +1155,7 @@ func runSSTablePropertiesCmd(t *testing.T, td *datadriven.TestData, d *DB) strin
 	var v sstable.VirtualReader
 	props := r.Properties.String()
 	if m != nil && m.Virtual {
-		v = sstable.MakeVirtualReader(r, m.VirtualMeta())
+		v = sstable.MakeVirtualReader(r, m.VirtualMeta(), false /* isForeign */)
 		props = v.Properties.String()
 	}
 	if len(td.Input) == 0 {
--- a/db.go
+++ b/db.go
@@ -2942,6 +2942,15 @@ func (d *DB) checkVirtualBounds(m *fileMetadata) {
 		return
 	}
 
+	objMeta, err := d.objProvider.Lookup(fileTypeTable, m.FileBacking.DiskFileNum)
+	if err != nil {
+		panic(err)
+	}
+	if objMeta.IsExternal() {
+		// Nothing to do; bounds are expected to be loose.
+		return
+	}
+
 	if m.HasPointKeys {
 		pointIter, rangeDelIter, err := d.newIters(context.TODO(), m, nil, internalIterOpts{})
 		if err != nil {
--- a/ingest.go
+++ b/ingest.go
@@ -1529,12 +1529,6 @@ func (d *DB) ingest(
 		}
 	}
 
-	if invariants.Enabled {
-		for _, sharedMeta := range loadResult.sharedMeta {
-			d.checkVirtualBounds(sharedMeta)
-		}
-	}
-
 	info := TableIngestInfo{
 		JobID:     jobID,
 		Err:       err,
@@ -1737,7 +1731,6 @@ func (d *DB) excise(
 				return nil, err
 			}
 			leftFile.ValidateVirtual(m)
-			d.checkVirtualBounds(leftFile)
 			ve.NewFiles = append(ve.NewFiles, newFileEntry{Level: level, Meta: leftFile})
 			needsBacking = true
 			numCreatedFiles++
@@ -1859,7 +1852,6 @@ func (d *DB) excise(
 			rightFile.Size = 1
 		}
 		rightFile.ValidateVirtual(m)
-		d.checkVirtualBounds(rightFile)
 		ve.NewFiles = append(ve.NewFiles, newFileEntry{Level: level, Meta: rightFile})
 		needsBacking = true
 		numCreatedFiles++
--- a/scan_internal.go
+++ b/scan_internal.go
@@ -668,7 +668,17 @@ func scanInternalImpl(
 		case InternalKeyKindRangeKeyDelete, InternalKeyKindRangeKeyUnset, InternalKeyKindRangeKeySet:
 			if opts.visitRangeKey != nil {
 				span := iter.unsafeSpan()
-				if err := opts.visitRangeKey(span.Start, span.End, span.Keys); err != nil {
+				// NB: The caller isn't interested in the sequence numbers of these
+				// range keys. Rather, the caller wants them to be in trailer order
+				// _after_ zeroing of sequence numbers. Copy span.Keys, sort it, and then
+				// call visitRangeKey.
+				keysCopy := make([]keyspan.Key, len(span.Keys))
+				for i := range span.Keys {
+					keysCopy[i] = span.Keys[i]
+					keysCopy[i].Trailer = base.MakeTrailer(0, span.Keys[i].Kind())
+				}
+				keyspan.SortKeysByTrailer(&keysCopy)
+				if err := opts.visitRangeKey(span.Start, span.End, keysCopy); err != nil {
 					return err
 				}
 			}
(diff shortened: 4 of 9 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3074)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`table_cache.go` line 204 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3074#-NjJ0rw43RXvt11Tekns:-NjJ0rw43RXvt11Teknt:bcilqps) ([raw file](https://github.com/cockroachdb/pebble/blob/da893c5dd0d8e565da6c2dc64a554ef48e771175/table_cache.go#L204)):*
> ```Go
> }
> 
> func createCommonReader(
> ```

It would be useful to document what `isForeign` does exactly here.

It's kind of unfortunate to add it to this method which is supposed to hide away the virtual vs non-virtual differences.. Maybe this flag belongs in `FileMetadata`, alongside `Virtual`?

Is there a description of `hideObsoletePoints` somewhere?
___
*[`data_test.go` line 1158 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3074#-NjJ1B791Putxz8akRDf:-NjJ1B791Putxz8akRDg:b-kae055) ([raw file](https://github.com/cockroachdb/pebble/blob/da893c5dd0d8e565da6c2dc64a554ef48e771175/data_test.go#L1158)):*
> ```Go
> 	props := r.Properties.String()
> 	if m != nil && m.Virtual {
> 		v = sstable.MakeVirtualReader(r, m.VirtualMeta(), false /* isForeign */)
> ```

How do we know it's not foreign? I assume it doesn't matter here, but then the flag shouldn't be `isForeign`, it should be more descriptive of what it actually does.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3074#-:-NjJ80OQAzBtUFDLyPs6:b-814g8k)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`table_cache.go` line 204 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3074#-NjJ0rw43RXvt11Tekns:-NjJ6Bez70TgVSSGv7d1:b-23ei7m) ([raw file](https://github.com/cockroachdb/pebble/blob/da893c5dd0d8e565da6c2dc64a554ef48e771175/table_cache.go#L204)):*

It would be useful to document what `isForeign` does exactly here.

It's kind of unfortunate to add it to this method which is supposed to hide away the virtual vs non-virtual differences.. Maybe this flag belongs in `FileMetadata`, alongside `Virtual`?

Is there a description of `hideObsoletePoints` somewhere?
</blockquote></details>

Added a comment. And I do have a bigger refactor in mind where commonReader just becomes an interface, and this method encapsulates the isForeign determination logic instead of having the caller do it.  I'll try to do that later; this is a minimal change to fix the bugs.

And since IsForeign is a state variable that's derived from stuff that's already stored in the catalog, I think it's fair to just leave it as a calculated field instead of storing it in the FileMetadata. Though I'm open to either argument. For the sake of making this backportable, we should probably just pass in `isForeign`?

`hideObsoletePoints` is sort of defined in reader.go around the NewIter* methods, but not in great detail.
___
*[`data_test.go` line 1158 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3074#-NjJ1B791Putxz8akRDf:-NjJ7Dmh2pR5zdq6Mrf3:b-nky8p3) ([raw file](https://github.com/cockroachdb/pebble/blob/da893c5dd0d8e565da6c2dc64a554ef48e771175/data_test.go#L1158)):*

How do we know it's not foreign? I assume it doesn't matter here, but then the flag shouldn't be `isForeign`, it should be more descriptive of what it actually does.
</blockquote></details>

Yeah it's just a test helper function, and these tests don't create foreign SSTs, only `TestIngestShared` and `TestScanInternal` etc do.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3074#-:-NjJCAyF1B1zdrOev_fw:bnfp4nl)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @sumeerbhola)

___
*[`table_cache.go` line 204 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3074#-NjJ0rw43RXvt11Tekns:-NjJCCgZ0An2iwmvu2si:b6mq0us) ([raw file](https://github.com/cockroachdb/pebble/blob/da893c5dd0d8e565da6c2dc64a554ef48e771175/table_cache.go#L204)):*

Added a comment. And I do have a bigger refactor in mind where commonReader just becomes an interface, and this method encapsulates the isForeign determination logic instead of having the caller do it.  I'll try to do that later; this is a minimal change to fix the bugs.

And since IsForeign is a state variable that's derived from stuff that's already stored in the catalog, I think it's fair to just leave it as a calculated field instead of storing it in the FileMetadata. Though I'm open to either argument. For the sake of making this backportable, we should probably just pass in `isForeign`?

`hideObsoletePoints` is sort of defined in reader.go around the NewIter* methods, but not in great detail.
</blockquote></details>

I guess this is where I'm coming from: it doesn't feel right that we should tie semantics to whether a file is external or not, even though these are correlated in the current usecases.. It should be something that is provided separately when we attach the file. Maybe one day we will want to attach an external file as a "regular" SST - e.g. starting a "cloned cluster" from saved object files.

This may or may not make sense because I still don't understand the `hideObsoletePoints` semantics well.

Anyway this PR is ok to merge for now, just trying to figure out the desired long term state.

itsbilal (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3074)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`table_cache.go` line 204 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3074#-NjJ0rw43RXvt11Tekns:-NjJHaMkE8Db1D2qAWiW:b-wgf5dk) ([raw file](https://github.com/cockroachdb/pebble/blob/da893c5dd0d8e565da6c2dc64a554ef48e771175/table_cache.go#L204)):*

I guess this is where I'm coming from: it doesn't feel right that we should tie semantics to whether a file is external or not, even though these are correlated in the current usecases.. It should be something that is provided separately when we attach the file. Maybe one day we will want to attach an external file as a "regular" SST - e.g. starting a "cloned cluster" from saved object files.

This may or may not make sense because I still don't understand the `hideObsoletePoints` semantics well.

Anyway this PR is ok to merge for now, just trying to figure out the desired long term state.
</blockquote></details>

Yeah I can see that. Currently we've tied "should we only read the newest internal key for each user key" (which is what `hideObsoletePoints` really does), to "is this file shared foreign" (which btw is not the same as external), and does make sense to have them as two separate things. It also makes sense to have a FileMetadata field for this, seeing as that's usually the place to interpret *how* to read a file, while the catalog ObjectMetadata mostly speaks about _where_ the file is.

TFTR!

PR #3075, opened 2023-11-15
compaction: return correct error if copy/move compaction is cancelled

itsbilal:
In rare cases, it's possible for an excise to come in and cancel
a move/copy compaction, and for us to try to read an input file for the
compaction that no longer exists. In that case we previously
would return an `objprovider: file not found` error, when in reality
we should return ErrCancelledCompaction so it's handled correctly
higher up in the stack, such as in metamorphic tests.

Found with #3044.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2843,6 +2843,12 @@ func (d *DB) runCopyCompaction(
 		},
 	}
 	pendingOutputs = append(pendingOutputs, metaCopy.PhysicalMeta())
+	// Before dropping the db mutex, grab a ref to the current version. This
+	// prevents any concurrent excises from deleting files that this compaction
+	// needs to read/maintain a reference to.
+	vers := d.mu.versions.currentVersion()
+	vers.Ref()
+	defer vers.UnrefLocked()
 
 	d.mu.Unlock()
 	defer d.mu.Lock()
@@ -2919,6 +2925,9 @@ func (d *DB) runCompaction(
 			if iter.Next() != nil {
 				panic("got more than one file for a move or copy compaction")
 			}
+		}
+		if c.cancel.Load() {
+			return ve, nil, stats, ErrCancelledCompaction
 		}
 		objMeta, err := d.objProvider.Lookup(fileTypeTable, meta.FileBacking.DiskFileNum)
 		if err != nil {


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3075#-:-NjJNSd42RDfbCH2rC56:bnfp4nl)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`compaction.go` line 2720 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3075#-NjJM7uT2lwuTyaQPDvP:-NjJM7uT2lwuTyaQPDvQ:b-bpc28d) ([raw file](https://github.com/cockroachdb/pebble/blob/5c7cb979732e034737bac715fdf7dac50b676ba5/compaction.go#L2720)):*
> ```Go
> 			// the manifest lock, we don't expect this bool to change its value
> 			// as only the holder of the manifest lock will ever write to it.
> 			if c.cancel.Load() {
> ```

[nit] this code pretty obfuscated, can we just remove the `func` altogether and do

````
d.mu.versions.logLock()
if c.cancel.Load() {
  err = ErrCancelledCompaction
  d.mu.versions.logUnlock()
} else {
  // logAndApply calls logUnlock.
  err = logAndApply()
}
```
````

___
*[`compaction.go` line 2735 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3075#-NjJN9KNERsq7Hp1TwRk:-NjJN9KNERsq7Hp1TwRl:bv9sbpk) ([raw file](https://github.com/cockroachdb/pebble/blob/5c7cb979732e034737bac715fdf7dac50b676ba5/compaction.go#L2735)):*
> ```Go
> 		if err != nil {
> 			// TODO(peter): untested.
> 			for _, f := range pendingOutputs {
> ```

Just a question, but we need this cleanup only here because when `runCompaction()` returns an error, it cleans up after itself?

**Review by sumeerbhola — DISMISSED** [review]
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3075)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`compaction.go` line 2753 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3075#-NjJS7Zj7dQJV9b1-jYy:-NjJS7Zj7dQJV9b1-jYz:b8smne) ([raw file](https://github.com/cockroachdb/pebble/blob/5c7cb979732e034737bac715fdf7dac50b676ba5/compaction.go#L2753)):*
> ```Go
> 		// operation. It's possible the error returned by runCompaction is due to a
> 		// missing file that has been excised away. Replace it with
> 		// ErrCancelledCompaction.
> ```

Do we want to check the error from `runCompaction` and scope more narrowly which errors to swallow?
Or is the concern here that the file getting deleted from under the compaction could result in a wide variety of errors e.g. we may see the missing file when trying to read a block that isn't in the cache?

We also have https://github.com/cockroachdb/pebble/blob/ec76d8bf346db3484aacb6078dda144bbcfb8794/compaction.go#L2959-L2969 so I am quite confused about how the file disappeared.

**Review by itsbilal — COMMENTED** [review]
TFTR! Needs another look as the approach is very different now.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3075#-:-NjJqrVO2LDaCciDU0l8:b-erhps5)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`compaction.go` line 2720 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3075#-NjJM7uT2lwuTyaQPDvP:-NjJpwU2-k3lt5u-G3Pl:bl150lb) ([raw file](https://github.com/cockroachdb/pebble/blob/5c7cb979732e034737bac715fdf7dac50b676ba5/compaction.go#L2720)):*

[nit] this code pretty obfuscated, can we just remove the `func` altogether and do

````
d.mu.versions.logLock()
if c.cancel.Load() {
  err = ErrCancelledCompaction
  d.mu.versions.logUnlock()
} else {
  // logAndApply calls logUnlock.
  err = logAndApply()
}
```
````

</blockquote></details>

N/a as I undid the change in this method.
___
*[`compaction.go` line 2735 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3075#-NjJN9KNERsq7Hp1TwRk:-NjJq1Go0raD-EO-HuCF:bjber26) ([raw file](https://github.com/cockroachdb/pebble/blob/5c7cb979732e034737bac715fdf7dac50b676ba5/compaction.go#L2735)):*

Just a question, but we need this cleanup only here because when `runCompaction()` returns an error, it cleans up after itself?
</blockquote></details>

`runCompaction` doesn't actually clean up after itself it seems, but it does `nil` the pendingOutputs in a defer if the error being returned is non-nil. We probably should fix the non-cleanup.
___
*[`compaction.go` line 2753 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3075#-NjJS7Zj7dQJV9b1-jYy:-NjJq8l43aHYXNMAKxmv:b9edxjn) ([raw file](https://github.com/cockroachdb/pebble/blob/5c7cb979732e034737bac715fdf7dac50b676ba5/compaction.go#L2753)):*

Do we want to check the error from `runCompaction` and scope more narrowly which errors to swallow?
Or is the concern here that the file getting deleted from under the compaction could result in a wide variety of errors e.g. we may see the missing file when trying to read a block that isn't in the cache?

We also have https://github.com/cockroachdb/pebble/blob/ec76d8bf346db3484aacb6078dda144bbcfb8794/compaction.go#L2959-L2969 so I am quite confused about how the file disappeared.
</blockquote></details>

You bring up a good point, and digging in, I realized this bug was only possible with a move/copy compaction (which is where I saw it). The move compaction path didn't check for `c.cancel.Load()` at any point, so the first time it'd hit an error would be when the provider.Lookup fails. I updated this PR to just add a cancel check before we do a move compaction, and also grab the ref before we do a copy compaction. PTAL.

RaduBerinde (commented):
:lgtm: Much cleaner!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3075#-:-NjJrWoX-ZXl5aOupxZs:b7lsvxr)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @sumeerbhola)

itsbilal:
TFTR!

PR #3077, opened 2023-11-15
db: add CanDeterministicallySingleDelete

jbowens:
Add a new specialized CanDeterministicallySingleDelete function that allows a
user to inspect hidden internal keys within the current user key to determine
whether the behavior of a SingleDelete of the current Key would be
deterministic, assuming no writes are made to the key between Iterator creation
and commit of the SingleDelete. The CanDeterministicallySingleDelete func may
only be called on an Iterator oriented in the forward direction and does not
change the external Iterator position.

During intent resolution, CockroachDB will write a SINGLEDEL tombstone to
delete the intent if the intent's value indicates that it must have only been
written once. Subtle logic that cuts across the stack makes this optimization
possible, and one possible explanation for a recent instance of replica
divergence (cockroachdb/cockroach#114421) is a bug in this logic. We anticipate
using CanDeterministicallySingleDelete within CockroachDB's intent resolution
to validate this logic at runtime.

There is some potential for this function to drive the decision to use
SingleDelete when creating batches for local Engine application only.

Informs cockroachdb/cockroach#114492.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -132,6 +132,27 @@ func runIterCmd(d *datadriven.TestData, iter *Iterator, closeIter bool) string {
 			}
 			validityState = iter.NextWithLimit([]byte(parts[1]))
 			printValidityState = true
+		case "internal-next":
+			validity, keyKind := iter.internalNext()
+			switch validity {
+			case internalNextError:
+				fmt.Fprintf(&b, "err: %s\n", iter.Error())
+			case internalNextExhausted:
+				fmt.Fprint(&b, ".\n")
+			case internalNextValid:
+				fmt.Fprintf(&b, "%s\n", keyKind)
+			default:
+				panic("unreachable")
+			}
+			continue
+		case "can-deterministically-single-delete":
+			ok, err := CanDeterministicallySingleDelete(iter)
+			if err != nil {
+				fmt.Fprintf(&b, "err: %s\n", err)
+			} else {
+				fmt.Fprintf(&b, "%t\n", ok)
+			}
+			continue
 		case "prev-limit":
 			if len(parts) != 2 {
 				return "prev-limit <limit>\n"
@@ -402,7 +423,7 @@ func runBatchDefineCmd(d *datadriven.TestData, b *Batch) error {
 			err = b.Delete([]byte(parts[1]), nil)
 		case "del-sized":
 			if len(parts) != 3 {
-				return errors.Errorf("%s expects 1 argument", parts[0])
+				return errors.Errorf("%s expects 2 arguments", parts[0])
 			}
 			var valSize uint64
 			valSize, err = strconv.ParseUint(parts[2], 10, 32)
--- a/iterator.go
+++ b/iterator.go
@@ -449,6 +449,11 @@ const (
 	seekPrefixGELastPositioningOp
 	seekGELastPositioningOp
 	seekLTLastPositioningOp
+	// internalNextOp is a special internal iterator positioning operation used
+	// by CanDeterministicallySingleDelete. It exists for enforcing requirements
+	// around calling CanDeterministicallySingleDelete at most once per external
+	// iterator position.
+	internalNextOp
 	// invalidatedLastPositionOp is similar to unknownLastPositionOp and the
 	// only reason to distinguish this is for the wider set of SeekGE
 	// optimizations we permit for the external iterator Iterator.forwardOnly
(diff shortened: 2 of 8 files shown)


sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3077)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`iterator.go` line 1921 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjP6BIDDtRSOaLkUYvg:-NjP6BIDDtRSOaLkUYvh:b-wmhrt0) ([raw file](https://github.com/cockroachdb/pebble/blob/3f35cf1b9e67918226479e20e2731e976c36680d/iterator.go#L1921)):*
> ```Go
> 
> // InternalNext advances internal Iterator state forward to expose the
> // InternalKeyKind of the next internal key with a user key equal to Key().
> ```

So `InternalNext` behaves as if nothing was actually done to the iterator, yes?
This looks fine.

We may have to do something about stability of the key/value pair already returned, if we want the caller to be able to use them after calling `InternalNext`. I guess the key is already stable since we are not fiddling with `i.key`. And cloning the `LazyValue` does feel unnecessary since CockroachDB eagerly extracts the value and parses the `MVCCMetadata` proto. In which case we could just document the instability of the value.

jbowens (commented):
This is RFAL

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3077#-:-Njhpa4b7xcDfF0PVyt5:b-n27ya0)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`iterator.go` line 1921 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjP6BIDDtRSOaLkUYvg:-Njhp_AX1wKfAPaYhCK9:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/3f35cf1b9e67918226479e20e2731e976c36680d/iterator.go#L1921)):*

So `InternalNext` behaves as if nothing was actually done to the iterator, yes?
This looks fine.

We may have to do something about stability of the key/value pair already returned, if we want the caller to be able to use them after calling `InternalNext`. I guess the key is already stable since we are not fiddling with `i.key`. And cloning the `LazyValue` does feel unnecessary since CockroachDB eagerly extracts the value and parses the `MVCCMetadata` proto. In which case we could just document the instability of the value.
</blockquote></details>

Done.

sumeerbhola (approved):
:lgtm: 

> Reviewed 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3077#-:-Nji2sKJ0ozbFE3GLgCG:bg7igkv)* status: all files reviewed, 8 unresolved discussions (waiting on @jbowens)

___
*[`iterator.go` line 1931 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjhtlUO-nXeRl0z4B-y:-NjhtlUO-nXeRl0z4B-z:bhfkxza) ([raw file](https://github.com/cockroachdb/pebble/blob/000d39dfd6576d4b454e3c903c926cf322cc97eb/iterator.go#L1931)):*
> ```Go
> // InternalNext, unlike all other Iterator methods, exposes internal LSM state.
> // InternalNext advances the Iterator's internal iterator to the next shadowed
> // key with the a user key equal to Key(). When a key is overwritten or deleted,
> ```

nit: ... with a user ...
___
*[`iterator.go` line 1935 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjhtrwV3_vlXwSry7XM:-NjhtrwV3_vlXwSry7XN:b-wuocpy) ([raw file](https://github.com/cockroachdb/pebble/blob/000d39dfd6576d4b454e3c903c926cf322cc97eb/iterator.go#L1935)):*
> ```Go
> // allows the caller to see whether an obsolete internal key exists with the
> // current Key(), and what it's key kind is. Note that the existence of an
> // internal key is nondeterministic and depedent on internal LSM state. These
> ```

dependent
___
*[`iterator.go` line 1956 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjhusEzBTcMbWDZ9Sfz:-NjhusEzBTcMbWDZ9Sg-:b1je4e3) ([raw file](https://github.com/cockroachdb/pebble/blob/000d39dfd6576d4b454e3c903c926cf322cc97eb/iterator.go#L1956)):*
> ```Go
> 		return InternalNextExhausted, base.InternalKeyKindInvalid
> 	}
> 	i.lastPositioningOp = unknownLastPositionOp
> ```

I was thinking that we will lose the monotonic seek optimization during intent resolution of a batch of intents because of this, but that may be ok. But then I noticed the loop with the call to `ltIter.NextEngineKey` in https://github.com/cockroachdb/cockroach/blob/e23591614d7ae4447a1d788c0e23e68a24018878/pkg/storage/mvcc.go#L4912 which suggests we already don't hit that optimization.
___
*[`iterator.go` line 1965 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3077#-Nji07w428nbJq-kUSu6:-Nji07w428nbJq-kUSu7:b-g2p6cz) ([raw file](https://github.com/cockroachdb/pebble/blob/000d39dfd6576d4b454e3c903c926cf322cc97eb/iterator.go#L1965)):*
> ```Go
> 			// variant so that the caller does not need to check i.iter.Error()
> 			// in the common case that the next internal key has a new user key.
> 			if i.err = firstError(i.err, i.iter.Error()); i.err != nil {
> ```

isn't `i.err` guaranteed to be nil prior to this assignment, so we could do `i.err = i.iter.Error()`?
___
*[`metamorphic/ops.go` line 1294 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3077#-Nji1UioA9dlBwTnZe4K:-Nji1UioA9dlBwTnZe4L:b-du4twc) ([raw file](https://github.com/cockroachdb/pebble/blob/000d39dfd6576d4b454e3c903c926cf322cc97eb/metamorphic/ops.go#L1294)):*
> ```Go
> 	_, _ = i.InternalNext()
> 	// The return value of an InternalNext is dependent on internal LSM state
> 	// and non-deterministic, so we don't record it. Including the operation
> ```

should we also have a test option that makes `iterInternalNextOp` a noop?
___
*[`metamorphic/ops.go` line 1298 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3077#-Nji1krO6GwWZd9EtELQ:-Nji1krO6GwWZd9EtELR:b-x4qvc6) ([raw file](https://github.com/cockroachdb/pebble/blob/000d39dfd6576d4b454e3c903c926cf322cc97eb/metamorphic/ops.go#L1298)):*
> ```Go
> 	// not change the result of any other Iterator operation that should be
> 	// deterministic, regardless of the outcome of InternalNext.
> 	h.Recordf("%s // %v", o, i.Error())
> ```

why is the error deterministic?
Is it because the only errors we happen to encounter here in the metamorphic test are caused by `i.err` already being true prior to the call to `i.InternalNext()` or `errors.New("switching from reverse to forward via peeking is prohibited")`?
___
*[`testdata/iter_histories/internal_next` line 109 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3077#-Nji2g3FCnxUFn3Z3dCo:-Nji2g3FCnxUFn3Z3dCp:b4m8z5h) ([raw file](https://github.com/cockroachdb/pebble/blob/000d39dfd6576d4b454e3c903c926cf322cc97eb/testdata/iter_histories/internal_next#L109)):*
> ```
> flush
> ----
> ```

Can you add a comment stating that we have two versions for each user key.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3077#-:-NjiJmLmAha9b5fvQL1u:b-814g8k)* status: 6 of 9 files reviewed, 7 unresolved discussions (waiting on @sumeerbhola)

___
*[`iterator.go` line 1935 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjhtrwV3_vlXwSry7XM:-NjiESPy4rpbbGIiSW8H:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/000d39dfd6576d4b454e3c903c926cf322cc97eb/iterator.go#L1935)):*

dependent
</blockquote></details>

Done.
___
*[`iterator.go` line 1956 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjhusEzBTcMbWDZ9Sfz:-NjiJEZDA0WxKs4Gda0I:b8h7yfk) ([raw file](https://github.com/cockroachdb/pebble/blob/000d39dfd6576d4b454e3c903c926cf322cc97eb/iterator.go#L1956)):*

I was thinking that we will lose the monotonic seek optimization during intent resolution of a batch of intents because of this, but that may be ok. But then I noticed the loop with the call to `ltIter.NextEngineKey` in https://github.com/cockroachdb/cockroach/blob/e23591614d7ae4447a1d788c0e23e68a24018878/pkg/storage/mvcc.go#L4912 which suggests we already don't hit that optimization.
</blockquote></details>

Hmm, I don't see any reason why we couldn't leave `lastPositioningOp` untouched, but these things can be subtle.
___
*[`iterator.go` line 1965 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3077#-Nji07w428nbJq-kUSu6:-NjiFsgyE2AIO7gOXuhL:b-us92ps) ([raw file](https://github.com/cockroachdb/pebble/blob/000d39dfd6576d4b454e3c903c926cf322cc97eb/iterator.go#L1965)):*

isn't `i.err` guaranteed to be nil prior to this assignment, so we could do `i.err = i.iter.Error()`?
</blockquote></details>

Yeah, Done.
___
*[`metamorphic/ops.go` line 1294 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3077#-Nji1UioA9dlBwTnZe4K:-NjiHJin7isc30XMUe1J:bf9odme) ([raw file](https://github.com/cockroachdb/pebble/blob/000d39dfd6576d4b454e3c903c926cf322cc97eb/metamorphic/ops.go#L1294)):*

should we also have a test option that makes `iterInternalNextOp` a noop?
</blockquote></details>

I think we should omit it because it allows us to assert that errors are deterministic. Some of the InternalNext branches are already just about noops, and the metamorphic tests are good at constructing a variety of internal LSM states that'll hit them.
___
*[`metamorphic/ops.go` line 1298 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3077#-Nji1krO6GwWZd9EtELQ:-NjiH2J02bcKeL-JWUOr:beaojgx) ([raw file](https://github.com/cockroachdb/pebble/blob/000d39dfd6576d4b454e3c903c926cf322cc97eb/metamorphic/ops.go#L1298)):*

why is the error deterministic?
Is it because the only errors we happen to encounter here in the metamorphic test are caused by `i.err` already being true prior to the call to `i.InternalNext()` or `errors.New("switching from reverse to forward via peeking is prohibited")`?
</blockquote></details>

Yeah, that's right. Added a comment here.
___
*[`testdata/iter_histories/internal_next` line 109 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3077#-Nji2g3FCnxUFn3Z3dCo:-NjiGJfI5y0gZJbE5OiW:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/000d39dfd6576d4b454e3c903c926cf322cc97eb/testdata/iter_histories/internal_next#L109)):*

Can you add a comment stating that we have two versions for each user key.
</blockquote></details>

Done.

jbowens (commented):
@sumeerbhola — I reworked this to keep the "internal next" operation internal, instead exposing a purpose-built function  `pebble.CanDeterministicallySingleDelete` that also handles the knowledge of what internal key kind sequences are okay. PTAL!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3077#-:-Njj-vri1zd6JqEQ3RCC:b4vm6i0)* status: 0 of 9 files reviewed, 7 unresolved discussions (waiting on @sumeerbhola)

sumeerbhola (approved):
:lgtm: 

> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3077#-:-NjjWfZw4-Y7MLxVIIsa:bg7igkv)* status: all files reviewed, 10 unresolved discussions (waiting on @jbowens and @nvanbenschoten)

___
*[`iterator.go` line 2862 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjjU3RjAUG5LbHfPJbP:-NjjU3RjAUG5LbHfPJbQ:b32wt8) ([raw file](https://github.com/cockroachdb/pebble/blob/fa1f4d69e999f0468cca7b55fdb79217ff4b19f0/iterator.go#L2862)):*
> ```Go
> //
> // If CanDeterministicallySingleDelete returns true AND the key at the iterator
> // position is not modified between the creation ofi the Iterator and the commit
> ```

of
___
*[`iterator.go` line 2864 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjjU9NO3RUBDg2kRvAe:-NjjU9NO3RUBDg2kRvAf:bq9tfiv) ([raw file](https://github.com/cockroachdb/pebble/blob/fa1f4d69e999f0468cca7b55fdb79217ff4b19f0/iterator.go#L2864)):*
> ```Go
> // position is not modified between the creation ofi the Iterator and the commit
> // of a batch containing a SingleDelete over the key, then the caller can be
> // assured that SingleDelete is equivalent to Delete.
> ```

How about: 
... is equivalent to Delete on the local engine. For batches that will be replicated, `CanDeterministicallySingleDelete` can return true on one engine, but it may not be true on another engine, since the first engine may have had multiple SETs collapsed into one.
___
*[`iterator.go` line 2880 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjjV5Ts0Z8WyyZON-wD:-NjjV5Ts0Z8WyyZON-wE:byyyc4y) ([raw file](https://github.com/cockroachdb/pebble/blob/fa1f4d69e999f0468cca7b55fdb79217ff4b19f0/iterator.go#L2880)):*
> ```Go
> 			//   SINGLEDEL; SET; SINGLEDEL; SET;
> 			// is not deterministic, because the first SET may consume the
> 			// second SINGLEDEL resulting in two SETs shadowed by one SINGLEDEL.
> ```

It is safe because we have SETWITHDEL. And hopefully we did a similar check when writing the SINGLEDEL so we don't have to continue now.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3077#-:-Njjbvau9pQg087_Zp1g:b-814g8k)* status: 8 of 9 files reviewed, 10 unresolved discussions (waiting on @nvanbenschoten and @sumeerbhola)

___
*[`iterator.go` line 2862 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjjU3RjAUG5LbHfPJbP:-NjjZYEZ-FEte7wq0oND:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/fa1f4d69e999f0468cca7b55fdb79217ff4b19f0/iterator.go#L2862)):*

of
</blockquote></details>

Done.
___
*[`iterator.go` line 2864 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjjU9NO3RUBDg2kRvAe:-Njj_P_6A7q3_4YtbSfY:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/fa1f4d69e999f0468cca7b55fdb79217ff4b19f0/iterator.go#L2864)):*

How about: 
... is equivalent to Delete on the local engine. For batches that will be replicated, `CanDeterministicallySingleDelete` can return true on one engine, but it may not be true on another engine, since the first engine may have had multiple SETs collapsed into one.
</blockquote></details>

Done.
___
*[`iterator.go` line 2880 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjjV5Ts0Z8WyyZON-wD:-NjjaOeHDZeSUC1CYf6J:bcn3o5h) ([raw file](https://github.com/cockroachdb/pebble/blob/fa1f4d69e999f0468cca7b55fdb79217ff4b19f0/iterator.go#L2880)):*

It is safe because we have SETWITHDEL. And hopefully we did a similar check when writing the SINGLEDEL so we don't have to continue now.
</blockquote></details>

Ah, right!

nvanbenschoten (commented):
> Reviewed 6 of 9 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3077)* status: all files reviewed, 11 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`-- commits` line 18 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjjjI_90QDMo201FnIX):*
Is the thinking that we will add calls to `CanDeterministicallySingleDelete` in testing builds of CockroachDB? Behind a `buildutil.CrdbTestBuild`?

If so, where will we add those? Just during evaluation of ResolveIntent(Range) requests above Raft on the leaseholder? Or also before applying `WriteBatches` that contain `SingleDelete` operations on all replicas below Raft?

The second check would be more comprehensive, but also more difficult to introduce without adding logic to iterate over every WriteBatch during raft application to search for `SingleDelete` operations. It makes me wonder whether this kind of assertion would be more natural to push into Pebble. For example, I could imagine a mode where Pebble could be configured to check for determinism whenever applying _any_ SingleDelete operation. We might not want to be reading on the apply path though, so that may not work.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3077)* status: all files reviewed, 11 unresolved discussions (waiting on @nvanbenschoten and @sumeerbhola)

___
*[`-- commits` line 18 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjjjI_90QDMo201FnIX):*
<details><summary><i>Previously, nvanbenschoten (Nathan VanBenschoten) wrote…</i></summary><blockquote>

Is the thinking that we will add calls to `CanDeterministicallySingleDelete` in testing builds of CockroachDB? Behind a `buildutil.CrdbTestBuild`?

If so, where will we add those? Just during evaluation of ResolveIntent(Range) requests above Raft on the leaseholder? Or also before applying `WriteBatches` that contain `SingleDelete` operations on all replicas below Raft?

The second check would be more comprehensive, but also more difficult to introduce without adding logic to iterate over every WriteBatch during raft application to search for `SingleDelete` operations. It makes me wonder whether this kind of assertion would be more natural to push into Pebble. For example, I could imagine a mode where Pebble could be configured to check for determinism whenever applying _any_ SingleDelete operation. We might not want to be reading on the apply path though, so that may not work.
</blockquote></details>

The intention is to call it in production builds just in evaluation. We expect the regression introduced to be small, because we already advance over these internal keys in NextEngineKey. Although this function doesn’t change the externally-visible iterator position, it does reduce the work the subsequent Next needs to perform.

Long-term I would like to explore pushing the single deletion decision point into Pebble (some thoughts in cockroachdb/cockroach#114492). I think it’d be nice to either completely obscure the single deletion optimization to the user, or make it simply a suggestion that Pebble may ignore if it’s unsafe or too expensive to prove it’s safe. Doing so might regress in terms of resource utilization per intent resolved since we need to introduce a read somewhere there wasn’t previously, but that seems like it should be offset by its wider applicability, like the raft log.

jbowens:
*[`-- commits` line 18 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjjjI_90QDMo201FnIX):*

The intention is to call it in production builds just in evaluation. We expect the regression introduced to be small, because we already advance over these internal keys in NextEngineKey. Although this function doesn’t change the externally-visible iterator position, it does reduce the work the subsequent Next needs to perform.

Long-term I would like to explore pushing the single deletion decision point into Pebble (some thoughts in cockroachdb/cockroach#114492). I think it’d be nice to either completely obscure the single deletion optimization to the user, or make it simply a suggestion that Pebble may ignore if it’s unsafe or too expensive to prove it’s safe. Doing so might regress in terms of resource utilization per intent resolved since we need to introduce a read somewhere there wasn’t previously, but that seems like it should be offset by its wider applicability, like the raft log.
</blockquote></details>

To be clear, we intend to call it in production during evaluation only as an invariant assertion.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3077)* status: all files reviewed, 11 unresolved discussions (waiting on @nvanbenschoten and @sumeerbhola)

___
*[`iterator.go` line 2880 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3077#-NjjV5Ts0Z8WyyZON-wD:-NjmYpwD3YScTh-KP1dh:b-t22bxh) ([raw file](https://github.com/cockroachdb/pebble/blob/fa1f4d69e999f0468cca7b55fdb79217ff4b19f0/iterator.go#L2880)):*

Ah, right!
</blockquote></details>

On second thought, maybe we should still check beneath the observed SINGLEDEL because in the context of `MVCCResolveWriteIntentRange` we're going to iterate over those internal keys during the next call to `NextEngineKey` anyways: may as well validate that the already-written SINGLEDEL obeyed invariants too?

sumeerbhola (approved):
:lgtm: 

> Reviewed 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3077#-:-NjnKIrG8SEX3G8mvyV2:bg7igkv)* status: all files reviewed (commit messages unreviewed), 11 unresolved discussions (waiting on @nvanbenschoten)

PR #3078, opened 2023-11-16
objstorage/objstorageprovider: save custom object name to catalog

stevendanna:
Previously, we weren't adding the custom object name to the catalog. As a result, upon re-opening, various parts of pebble would treat this external object as a local object.

### DIFF
--- a/objstorage/objstorageprovider/provider_test.go
+++ b/objstorage/objstorageprovider/provider_test.go
@@ -593,7 +593,7 @@ func TestParallelSync(t *testing.T) {
 							}
 						}
 					}
-				}(numOps*n, shared)
+				}(numOps*(n+1), shared)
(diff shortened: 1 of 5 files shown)


objstorage/objstorageprovider/provider_test.go
```diff
@@ -593,7 +593,7 @@ func TestParallelSync(t *testing.T) {
 							}
 						}
 					}
-				}(numOps*n, shared)
+				}(numOps*(n+1), shared)
```
stevendanna, line 596: This change is to ensure that we don't get a FileNum of 0 which will trigger various assertions during the invariant build.

itsbilal (approved):
LGTM, thanks for fixing this!

Issue #1255, opened 2021-09-08
db: more deterministic SingleDelete semantics for Set


sumeerbhola:
Currently SingleDelete is only guaranteed to behave correctly if the key has been Set at most once.
- We say something about Merge in the code, but CockroachDB does not currently use Merge with SingleDelete, and this issue is not trying to make it behave better for Merge.
- The RockDB doc for SingleDelete says it can be used if the user key has only been Set once since the last SingleDelete https://github.com/facebook/rocksdb/wiki/Single-Delete. I haven't verified whether that doc is correct, but this is definitely not true for Pebble (see example below).

We need to accommodate cases where the key has been Set at most once since the last Delete/SingleDelete. That is, sequences like `Set => SingleDelete/Delete => Set => SingleDelete` where `=>` is happens-before. This will eliminate current bugs in CockroachDB related to early intent removal (https://github.com/cockroachdb/cockroach/issues/69891) which cause us to forget whether the user key has been used before, and will eliminate future concerns related to range movement using Delete to delete a range (instead of DeleteRange). The above sequences do not work correctly since it can be transformed into `Set => Set => SingleDelete` if the later `Set` and `SingleDelete/Delete` meet in a compaction. This can then be transformed into `Set` if the later `Set` and `SingleDelete` meet in a compaction.

The following informally sketches the solution.

Semantics: SingleDelete declares a desire by the caller to Delete that user key, where the caller expects that a Get at a sequence number after the SingleDelete should return nothing. In return, the caller is promising that there is at most one Set since the last Delete/SingleDelete.

Solution:
- Introduce a SET-WITH-DEL key-kind that represents a SET that has potentially consumed a DEL or SINGLEDEL key-kind in a compaction.
- The current SET key-kind will be interpreted as a SET that makes a promise that it has not consumed a DEL or SINGLEDEL. This is what will be written when a caller uses `pebble.Writer.Set`.
- A `SET => SINGLEDEL` in a compaction causes both to be consumed. A `SET-WITH-DEL => SINGLEDEL` in a compaction results in a `DEL`.
- Compatibility:
  - An alternative would be to interpret the current SET as a SET that makes no promise and introduce a SET-WITHOUT-DEL. This seems more natural since older versions of Pebble are not tracking these states. However such older versions are not correct with this broader use of SINGLEDEL either, so the broader use must be introduced *after* we are running the newer Pebble.
[...]

PR #3081, opened 2023-11-16
db: extend compaction iterator comments, tests

jbowens:
The compaction iterator had a few methods and complicated edge cases that were undocumented. This commit comments some of these and adds a new assertion. It also adds a few test cases around SINGLEDEL and DELSIZED, unfruitfully trying to reproduce some form of behavior that might explain cockroachdb/cockroach#114421.

### DIFF
--- a/compaction_iter.go
+++ b/compaction_iter.go
@@ -333,7 +333,7 @@ func (i *compactionIter) Next() (*InternalKey, []byte) {
 		return nil, nil
 	}
 
-	// Prior to this call to `Next()` we are in one of three situations with
+	// Prior to this call to `Next()` we are in one of four situations with
 	// respect to `iterKey` and related state:
 	//
 	// - `!skip && pos == iterPosNext`: `iterKey` is already at the next key.
@@ -342,11 +342,33 @@ func (i *compactionIter) Next() (*InternalKey, []byte) {
 	//   snapshot stripe.
 	// - `skip && pos == iterPosCurForward`: We are at the key that has been returned.
 	//   To move forward we skip skippable entries in the stripe.
+	// - `skip && pos == iterPosNext && i.iterStripeChange == sameStripeNonSkippable`:
+	//    This case may occur when skipping within a snapshot stripe and we
+	//    encounter either:
+	//      a) an invalid key kind; The previous call will have returned
+	//         whatever key it was processing and deferred handling of the
+	//         invalid key to this invocation of Next(). We're responsible for
+	//         ignoring skip=true and falling into the invalid key kind case
+	//         down below.
+	//      b) an interleaved range delete; This is a wart of the current code
+	//         structure. While skipping within a snapshot stripe, a range
+	//         delete interleaved at its start key and sequence number
+	//         interrupts the sequence of point keys. After we return the range
+	//         delete to the caller, we need to pick up skipping at where we
+	//         left off, so we preserve skip=true.
+	//    TODO(jackson): This last case is confusing and can be removed if we
+	//    interleave range deletions at the maximal sequence number using the
+	//    keyspan interleaving iterator. This is the treatment given to range
+	//    keys today.
 	if i.pos == iterPosCurForward {
 		if i.skip {
 			i.skipInStripe()
 		} else {
 			i.nextInStripe()
+		}
+	} else if i.skip {
+		if i.iterStripeChange != sameStripeNonSkippable {
+			panic(errors.AssertionFailedf("compaction iterator has skip=true, but iterator is at iterPosNext"))
 		}
 	}
 
@@ -430,6 +452,12 @@ func (i *compactionIter) Next() (*InternalKey, []byte) {
 					// can be elided skip skippable keys in the same stripe.
 					i.saveKey()
 					i.skipInStripe()
+					if i.iterStripeChange == newStripeSameKey {
+						panic(errors.AssertionFailedf("pebble: skipInStripe in last stripe found a new stripe within the same key"))
+					}
+					if !i.skip && i.iterStripeChange != newStripeNewKey {
+						panic(errors.AssertionFailedf("pebble: skipInStripe in last stripe disabled skip without advancing to new key"))
+					}
 					continue
 				} else {
 					// We're not at the last snapshot stripe, so the tombstone
@@ -831,6 +859,16 @@ func (i *compactionIter) mergeNext(valueMerger ValueMerger) stripeChangeType {
 	}
 }
 
+// singleDeleteNext processes a SingleDelete point tombstone. A SingleDelete, or
+// SINGLEDEL, is unique in that it deletes exactly 1 internal key. It's a
+// performance optimization when the client knows a user key has not been
+// overwritten, allowing the elision of the tombstone earlier, avoiding write
+// amplification.
+//
+// singleDeleteNext returns a boolean indicating whether or not the caller
+// should yield the SingleDelete key to the consumer of the compactionIter. If
+// singleDeleteNext returns false, the caller may consume/elide the
+// SingleDelete.
 func (i *compactionIter) singleDeleteNext() bool {
 	// Save the current key.
 	i.saveKey()
@@ -839,6 +877,8 @@ func (i *compactionIter) singleDeleteNext() bool {
 
 	// Loop until finds a key to be passed to the next level.
 	for {
+		// If we find a key that can't be skipped, return true so that the
+		// caller yields the SingleDelete to the caller.
 		if i.nextInStripe() != sameStripeSkippable {
 			i.pos = iterPosNext
 			return true
@@ -854,11 +894,26 @@ func (i *compactionIter) singleDeleteNext() bool {
 			return true
 
 		case InternalKeyKindSet:
+			// This SingleDelete deletes the Set, and we can now elide the
+			// SingleDel as well. We advance past the Set and return false to
+			// indicate to the main compaction loop that we should NOT yield the
+			// current SingleDel key to the compaction loop.
 			i.nextInStripe()
+			// TODO(jackson): We could assert that nextInStripe either a)
+			// stepped onto a new key, or b) stepped on to a Delete, DeleteSized
+			// or SingleDel key. This would detect improper uses of SingleDel,
+			// but only when all three internal keys meet in the same compaction
+			// which is not likely.
 			i.valid = false
 			return false
 
 		case InternalKeyKindSingleDelete:
+			// Two single deletes met in a compaction. With proper deterministic
+			// use of SingleDelete, this should never happen. The expectation is
+			// that there's exactly 1 set beneath a single delete. Currently, we
+			// opt to skip it.
+			// TODO(jackson): Should we make this an error? This would also
+			// allow us to simplify the code a bit by removing the for loop.
 			continue
 
 		default:
@@ -895,7 +950,7 @@ func (i *compactionIter) deleteSizedNext() (*base.InternalKey, []byte) {
 	i.pos = iterPosNext
 	for i.nextInStripe() == sameStripeSkippable {
 		switch i.iterKey.Kind() {
-		case InternalKeyKindDelete, InternalKeyKindDeleteSized:
+		case InternalKeyKindDelete, InternalKeyKindDeleteSized, InternalKeyKindSingleDelete:
 			// We encountered a tombstone (DEL, or DELSIZED) that's deleted by
 			// the original DELSIZED tombstone. This can happen in two cases:
 			//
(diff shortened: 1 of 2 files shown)


sumeerbhola (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3081#-:-NjSoMOU9ORd60DMXjCq:b-uwkowd)* status: 0 of 2 files reviewed, 7 unresolved discussions (waiting on @jbowens)

___
*[`compaction_iter.go` line 454 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjSeFgyBtnIxFpyP96p:-NjSeFgyBtnIxFpyP96q:b12mzrb) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L454)):*
> ```Go
> 					// can be elided skip skippable keys in the same stripe.
> 					i.saveKey()
> 					i.skipInStripe()
> ```

Can we assert after skipInStripe returns that
- i.iterStripeChange is not newStripeSameKey
- if i.skip=false then i.iterStripeChange is newStripeNewKey

I realize this is not something specific to SingleDel, but should be harmless to have additional cheap assertions.
___
*[`compaction_iter.go` line 910 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjSo7yiA8ZiAWpgIOeO:-NjSo7yiA8ZiAWpgIOeP:ba07gck) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L910)):*
> ```Go
> 			// opt to skip it.
> 			// TODO(jackson): Should we make this an error? This would also
> 			// allow us to simplify the code a bit by removing the for loop.
> ```

According to https://github.com/cockroachdb/pebble/issues/1255
"the caller is promising that there is at most one Set since the last Delete/SingleDelete"

We should document this in the code if we don't already.
___
*[`compaction_iter.go` line 947 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjSjYGUBstdBoO64QJ8:-NjSjYGUBstdBoO64QJ9:b-o02p3b) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L947)):*
> ```Go
> 	for i.nextInStripe() == sameStripeSkippable {
> 		switch i.iterKey.Kind() {
> 		case InternalKeyKindDelete, InternalKeyKindDeleteSized:
> ```

So SingleDelete would fall into the default case? Doesn't seem like a correctness issue, but curious why that is.
___
*[`compaction_iter.go` line 990 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjSihKk8ZZyooWNztk7:-NjSihKk8ZZyooWNztk8:bpfrwjj) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L990)):*
> ```Go
> 				// stripe.
> 				i.key.SetKind(i.iterKey.Kind())
> 				i.skipInStripe()
> ```

we didn't need to do a `i.skipInStripe()` here since we already have `i.skip=true`, yes?
That would make it similar to how we handle `case InternalKeyKindDelete` in `compactionIter.Next`. Though doing it eagerly looks correct too.
___
*[`compaction_iter.go` line 1007 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjSj2B3-4dp9nW1gLml:-NjSj2B3-4dp9nW1gLmm:b-j496nh) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L1007)):*
> ```Go
> 			if len(i.value) == 0 {
> 				i.key.SetKind(InternalKeyKindDelete)
> 				i.skipInStripe()
> ```

same question.
___
*[`compaction_iter.go` line 1014 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjSgn2u8JOQdoyYj-oy:-NjSgn2u8JOQdoyYj-oz:b-6mgzhg) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L1014)):*
> ```Go
> 			expectedSize, n := binary.Uvarint(i.value)
> 			if n != len(i.value) {
> 				i.err = base.CorruptionErrorf("DELSIZED holds invalid value: %x", errors.Safe(i.value))
> ```

Can we tickle this path in a test and make sure we are not losing the error?
A TODO for a later PR is fine.
___
*[`compaction_iter.go` line 1034 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjSj56xDIFljr4dTU7g:-NjSj56xDIFljr4dTU7h:b-t3ofl8) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L1034)):*
> ```Go
> 				// user-provided size for accuracy, so ordinary DEL heuristics
> 				// are safer.
> 				i.key.SetKind(InternalKeyKindDelete)
> ```

Odd that we did not do
```
i.skipInStripe()  
return &i.key, i.value
```

Though it looks harmless.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3081)* status: 0 of 2 files reviewed, 8 unresolved discussions (waiting on @jbowens)

___
*[`compaction_iter.go` line 763 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjTNjY-9bq5eWxzP86R:-NjTNjY-9bq5eWxzP86S:bftwwmt) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L763)):*
> ```Go
> 				i.skip = true
> 				return
> 			}
> ```

There are some switch statements in this file involving the key kind that do not result in an error in the default case or do not have a default case. How about we change all of them to enumerate all key kinds. So something like
```
switch key.Kind() {
case <c1>: do something for c1
case <c2>: do something for c2
}
```

becomes
```
switch key.Kind() {
case <c1>: do something for c1
case <c2>: do something for c2
case ...: // do nothing
default:  panic(...)
}
```

The above if-block would also be rewritten to become a switch statement with this form.

It may give us more confidence that we have not forgotten some case since we are explicitly enumerating the ones for which we are doing nothing.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3081)* status: 0 of 2 files reviewed, 8 unresolved discussions (waiting on @sumeerbhola)

___
*[`compaction_iter.go` line 454 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjSeFgyBtnIxFpyP96p:-NjT9Vo93EM_KnjO3xA5:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L454)):*

Can we assert after skipInStripe returns that
- i.iterStripeChange is not newStripeSameKey
- if i.skip=false then i.iterStripeChange is newStripeNewKey

I realize this is not something specific to SingleDel, but should be harmless to have additional cheap assertions.
</blockquote></details>

Done.
___
*[`compaction_iter.go` line 910 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjSo7yiA8ZiAWpgIOeO:-NjTBbO8DHxsMKmBWdeo:bdnoqep) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L910)):*

According to https://github.com/cockroachdb/pebble/issues/1255
"the caller is promising that there is at most one Set since the last Delete/SingleDelete"

We should document this in the code if we don't already.
</blockquote></details>

Do we need to relax it to _at most_ one Set? Requiring exactly 1 set would allow us to more easily detect cases where interleaved parallel processes think they're obeying the contract but ultimately end up writing `SINGLEDEL` `SINGLEDEL` `SET` `SET`.
___
*[`compaction_iter.go` line 947 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjSjYGUBstdBoO64QJ8:-NjTN9xZAeDbRl8LkFaE:b-mots3m) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L947)):*

So SingleDelete would fall into the default case? Doesn't seem like a correctness issue, but curious why that is.
</blockquote></details>

Yeah, I don't think there ever was a direct motivation for it.
___
*[`compaction_iter.go` line 990 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjSihKk8ZZyooWNztk7:-NjTIbKcDzCTKvjbHll8:b9xr0jj) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L990)):*

we didn't need to do a `i.skipInStripe()` here since we already have `i.skip=true`, yes?
That would make it similar to how we handle `case InternalKeyKindDelete` in `compactionIter.Next`. Though doing it eagerly looks correct too.
</blockquote></details>

It's required for the `Next` invariant around `i.skip=true` implying `iterPosCurForward` (with the exception of `sameStripeNonSkippable`). The `InternalKeyKindDelete` case doesn't advance ahead, so it's permitted. I think it should be possible to refactor to relax the invariant. 
___
*[`compaction_iter.go` line 1007 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjSj2B3-4dp9nW1gLml:-NjTL-PD-lt5cIzOTm0y:b-34vmyd) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L1007)):*

same question.
</blockquote></details>

See above
___
*[`compaction_iter.go` line 1014 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjSgn2u8JOQdoyYj-oy:-NjTLIPL2_FZgRwQAacg:b-ag2igs) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L1014)):*

Can we tickle this path in a test and make sure we are not losing the error?
A TODO for a later PR is fine.
</blockquote></details>

It's easy enough to exercise this, so I added a case in this PR.
___
*[`compaction_iter.go` line 1034 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjSj56xDIFljr4dTU7g:-NjTMqoU5ifXkKq0q5A_:b-lnwjzd) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L1034)):*

Odd that we did not do
```
i.skipInStripe()  
return &i.key, i.value
```

Though it looks harmless.
</blockquote></details>

That's more explicit and obviously correct, so I updated the code to do this (plus remove the value).

sumeerbhola:
> Do we need to relax it to at most one Set?

Probably don't need to relax it. But I need to think about it some more.

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3081#-:-NjTajzL74LHRYpdwBmp:bg7igkv)* status: all files reviewed, 8 unresolved discussions (waiting on @jbowens)

___
*[`compaction_iter.go` line 763 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3081#-NjTNjY-9bq5eWxzP86R:-NjTaeC-EP53ne32dNuJ:b7d1ffa) ([raw file](https://github.com/cockroachdb/pebble/blob/9a8f206a96410b80713c72ac0351631a19640fc1/compaction_iter.go#L763)):*

There are some switch statements in this file involving the key kind that do not result in an error in the default case or do not have a default case. How about we change all of them to enumerate all key kinds. So something like
```
switch key.Kind() {
case <c1>: do something for c1
case <c2>: do something for c2
}
```

becomes
```
switch key.Kind() {
case <c1>: do something for c1
case <c2>: do something for c2
case ...: // do nothing
default:  panic(...)
}
```

The above if-block would also be rewritten to become a switch statement with this form.

It may give us more confidence that we have not forgotten some case since we are explicitly enumerating the ones for which we are doing nothing.
</blockquote></details>

only remaining comment, in case you want to address it before merging.

jbowens:
TFTR!

I'm going to merge this before addressing the key kind switch comment, but I'll capture it in a broader issue about the compaction iterator and clean up.

Issue #3082, opened 2023-11-17
db: refactor/cleanup compaction iterator


jbowens:
- [x] Remove the compaction trailer optimization [[1](https://github.com/cockroachdb/pebble/blob/bfb2e6a97fdd2671bad76f818100dec79742806a/compaction_iter.go#L613-L623)]. It's not particularly important in the context of a compaction and we have very subtle invariants involved.
- [x] Explicitly enumerate all key kinds https://github.com/cockroachdb/pebble/pull/3081#pullrequestreview-1737674028
- [ ] Use the `keyspan.InterleavingIter` to interleave range deletions at the maximal sequence number. This will simplify the logic considerably and allow us to eliminate the `sameStripeNonSkippable` case.

PR #3084, opened 2023-11-17
db: remove compaction iter seqnum optimization, replace with assert

jbowens:
Previously the compaction iterator would avoid user key comparisons in some circumstances when invariants around sequence numbers implied user keys must be different. This commit removes this optimization, replacing it with an assertion that the sequence number invariants hold.

This is done out of caution and with the realization that these user key comparisons form a relatively small portion of the CPU cost of a compaction.

Informs cockroachdb/pebble#3082.
Informs cockroachdb/cockroach#114421.

### DIFF
--- a/compaction_iter.go
+++ b/compaction_iter.go
@@ -15,7 +15,6 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/bytealloc"
-	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/internal/keyspan"
 	"github.com/cockroachdb/pebble/internal/rangekey"
 )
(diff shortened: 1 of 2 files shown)


sumeerbhola (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3084#-:-NjdakIh03HPh3sWY6X0:b-uwkowd)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`compaction_iter.go` line 655 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3084#-NjdaXl_188mMtYif4Ye:-NjdaXl_188mMtYif4Yf:b-ip68w2) ([raw file](https://github.com/cockroachdb/pebble/blob/026fe1a2c487679d64f9bdff15854cb536945aad/compaction_iter.go#L655)):*
> ```Go
> 	// iterator's input iterator at the maximal sequence number so their
> 	// original sequence number will not be observed here.
> 	if prevSeqNum := base.SeqNumFromTrailer(i.keyTrailer); (prevSeqNum == 0 || prevSeqNum <= i.iterKey.SeqNum()) &&
> ```

nit: can you change this to `...prevSeqNum <= key.SeqNum() ...` since the comment is written using `key`.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3084#-:-NjhTsH77bTDFYbqKJMi:b-814g8k)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @sumeerbhola)

___
*[`compaction_iter.go` line 655 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3084#-NjdaXl_188mMtYif4Ye:-NjhTqZQ0psi5aXWRDNm:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/026fe1a2c487679d64f9bdff15854cb536945aad/compaction_iter.go#L655)):*

nit: can you change this to `...prevSeqNum <= key.SeqNum() ...` since the comment is written using `key`.
</blockquote></details>

Done.

Issue #3064, opened 2023-11-13
deprecate v1 format version


RaduBerinde:
So far we have kept compatibility all the way back to RocksDB and the first versions. However, this comes with a burden - there is a lot of complicated code that is no longer necessary with newer formats (and which is undertested as a consequence). In particular, newer formats no longer have the split key issue, where different instances of the same user key can be split into adjacent SSTs.

In order to simplify our code and improve quality for the newer versions, we should stop supporting these old versions. Note that Cockroach 23.1 guarantees no split keys, even on clusters upgraded from older versions.

We should come up with a plan on how to do this so that other users of Pebble can migrate to the newer versions. I propose the following:
 - [] create a `v1compat` branch which tracks `crl-release-23.2`; this will be the most up-to-date version which still supports `v1`. In the short term, any users who need compatibility can just use this branch.
 - [] deprecate the old versions on `master`, clean up code
 - [] work on a command line tool that can be used to upgrade an existing datastore.

RaduBerinde:
CC @jbowens @sumeerbhola Please add your feedback.

jbowens:
Linking to some related issues #1813, #1913, #1887, #1064, #2587.

I started to document format major versions and the upgrade path in #2162, but I haven't touched it in a year 😬

RaduBerinde:
Thanks, this is very helpful. Regarding #1064, I think it would be reasonable to "stamp" the version that will go out with 23.2 as `1.0.0` and subsequent 23.2 changes as `1.0.1` etc. Any subsequent version we stamp from the 24.1 series will be `2.0.0` as it will not support the v1 format anymore.

jbowens:
There’s also this release-engineering ticket where we were hoping to tie pebble tags to corresponding cockroach tags: https://cockroachlabs.atlassian.net/browse/RE-430

That might be a more convenient (for us) versioning scheme.

RaduBerinde:
The problem is that other users of the repo would expect semantic versioning, not sure how much the 2X.Y.Z scheme is of use to them. Note that the cockroach tags can be something like `cockroach-2X.Y.Z` and live alongside the semantic versions like `1.0.0`.

RaduBerinde:
If we deprecate all format major versions smaller than `FormatFlushableIngest`, it looks like we will also deprecate `TableFormatLevelDB` and `TableFormatRocksDBv2`

[...]

PR #3085, opened 2023-11-17
db: enforce no split user keys in CheckOrdering

jbowens:
Adapt CheckOrdering to perform its overlap comparisons using the stricter requirement that adjacent files not contain the same user key. Since we still support the earlier format major versions that predate this requirement, this behavior is gated behind an OrderingInvariants enum variant determined by the active format major version.

Some of the tests that use more recent format major versions still created test cases with split user keys, and they needed to be adjusted.

Informs #3064.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1277,16 +1277,22 @@ func (c *compaction) newInputIter(
 	newIters tableNewIters, newRangeKeyIter keyspan.TableNewSpanIter, snapshots []uint64,
 ) (_ internalIterator, retErr error) {
 	// Validate the ordering of compaction input files for defense in depth.
+	// TODO(jackson): Some of the CheckOrdering calls may be adapted to pass
+	// ProhibitSplitUserKeys if we thread the active format major version in. Or
+	// if we remove support for earlier FMVs, we can remove the parameter
+	// altogether.
 	if len(c.flushing) == 0 {
 		if c.startLevel.level >= 0 {
 			err := manifest.CheckOrdering(c.cmp, c.formatKey,
-				manifest.Level(c.startLevel.level), c.startLevel.files.Iter())
+				manifest.Level(c.startLevel.level), c.startLevel.files.Iter(),
+				manifest.AllowSplitUserKeys)
 			if err != nil {
 				return nil, err
 			}
 		}
 		err := manifest.CheckOrdering(c.cmp, c.formatKey,
-			manifest.Level(c.outputLevel.level), c.outputLevel.files.Iter())
+			manifest.Level(c.outputLevel.level), c.outputLevel.files.Iter(),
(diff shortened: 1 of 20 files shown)


jbowens (commented):
I wrote this because I was growing concerned about the complexity of the ingest codepath and the possibility we might violate this invariant by ingesting a user key into a level it already exists within. As far as I can tell, we're not doing that, but this stronger assertion will ensure it.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3085#-:-NjUVeKT5jYQ7VL-QKuX:b-xolnq)* status: 0 of 20 files reviewed, all discussions resolved (waiting on @aadityasondhi, @RaduBerinde, and @sumeerbhola)

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3085#-:-Njj29i_CNCZzuofeWuH:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @aadityasondhi, @jbowens, and @RaduBerinde)

___
*[`format_major_version.go` line 222 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3085#-Njj1A5T1R3MhuCOaguy:-Njj1A5T1R3MhuCOaguz:b-sqd7v9) ([raw file](https://github.com/cockroachdb/pebble/blob/6cc49f3b7792115e9cf8d185bc4ccfc9737d0a79/format_major_version.go#L222)):*
> ```Go
> }
> 
> func (v FormatMajorVersion) orderingInvariants() manifest.OrderingInvariants {
> ```

Maybe add a comment that OrderingInvariants only get stricter as the format major version advances. So it is ok that someone retrieves an invariant and by the time it checks the invariant the FMV has advanced.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3085#-:-NjjYSeM8Bbf6cedLCQL:b-814g8k)* status: 19 of 20 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi, @RaduBerinde, and @sumeerbhola)

___
*[`format_major_version.go` line 222 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3085#-Njj1A5T1R3MhuCOaguy:-NjjYR4aD_Pue3ezUnVP:b-48vilg) ([raw file](https://github.com/cockroachdb/pebble/blob/6cc49f3b7792115e9cf8d185bc4ccfc9737d0a79/format_major_version.go#L222)):*

Maybe add a comment that OrderingInvariants only get stricter as the format major version advances. So it is ok that someone retrieves an invariant and by the time it checks the invariant the FMV has advanced.
</blockquote></details>

Good call, done.

PR #3089, opened 2023-11-20
db: more invariant checking for InternalKeyKind in compactionIter

sumeerbhola:
Informs https://github.com/cockroachdb/cockroach/issues/114421

### DIFF
--- a/compaction_iter.go
+++ b/compaction_iter.go
@@ -18,6 +18,7 @@ import (
 	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/internal/keyspan"
 	"github.com/cockroachdb/pebble/internal/rangekey"
+	"github.com/cockroachdb/redact"
 )
 
 // compactionIter provides a forward-only iterator that encapsulates the logic
@@ -484,8 +485,14 @@ func (i *compactionIter) Next() (*InternalKey, []byte) {
 			case InternalKeyKindSingleDelete:
 				if i.singleDeleteNext() {
 					return &i.key, i.value
+				} else if i.err != nil {
+					return nil, nil
 				}
 				continue
+
+			default:
+				panic(errors.AssertionFailedf(
+					"unexpected kind %s", redact.SafeString(i.iterKey.Kind().String())))
 			}
 
 		case InternalKeyKindSet, InternalKeyKindSetWithDelete:
@@ -495,6 +502,9 @@ func (i *compactionIter) Next() (*InternalKey, []byte) {
 			// preserving the original value, and potentially mutating the key
 			// kind.
 			i.setNext()
+			if i.err != nil {
+				return nil, nil
+			}
 			return &i.key, i.value
 
 		case InternalKeyKindMerge:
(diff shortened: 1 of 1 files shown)


sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3089)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`compaction_iter.go` line 605 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3089#-NjhqxZg6DsWr6cSdWCH:-NjhqxZg6DsWr6cSdWCI:b-wpm46p) ([raw file](https://github.com/cockroachdb/pebble/blob/2b849638632c7e91a05c0d0d61257f77aaa1da94/compaction_iter.go#L605)):*
> ```Go
> //
> // TODO(sumeer): skipInStripe() can return with i.err != nil and i.iterKey set
> // to nil, but callers don't look at i.err.
> ```

I believe the call to `i.iterNext()` in `nextInStripeHelper()` could cause this even prior to this PR.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3089)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`compaction_iter.go` line 605 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3089#-NjhqxZg6DsWr6cSdWCH:-Nji66Ju0GbcmIWbQWZu:b-tv8fy2) ([raw file](https://github.com/cockroachdb/pebble/blob/2b849638632c7e91a05c0d0d61257f77aaa1da94/compaction_iter.go#L605)):*

I believe the call to `i.iterNext()` in `nextInStripeHelper()` could cause this even prior to this PR.
</blockquote></details>

I've removed this TODO after auditing the callers, and adding some comments.

jbowens (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3089#-:-NjiCmSc-OLQvZykkCQw:b-yncq6h)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`compaction_iter.go` line 622 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3089#-NjiCB4_9sOlBwaDD-TS:-NjiCB4_9sOlBwaDD-TT:b6fbcj0) ([raw file](https://github.com/cockroachdb/pebble/blob/72a0b419362865176d2b23e7b1690d837fdc2659/compaction_iter.go#L622)):*
> ```Go
> func (i *compactionIter) iterNext() bool {
> 	var iterValue LazyValue
> 	i.iterKey, iterValue = i.iter.Next()
> ```

Whose responsibility is it to check `i.iter.Error()` if `i.iterKey` is nil after this `Next`? Some of our iterators implement `Error()` with `firstError(i.err, i.iter.Error())`, but `compactionIter` implements `Error()` by directly returning `i.err`.

itsbilal:
This will also need to be backported to 23.2 right?

sumeerbhola (commented):
> This will also need to be backported to 23.2 right?

Yep

TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3089#-:-NjiZmV74dMciFdnQ5K9:b2i769v)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`compaction_iter.go` line 622 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3089#-NjiCB4_9sOlBwaDD-TS:-NjiZ5GSC1Kd-Cf-ooOj:b-ycrb55) ([raw file](https://github.com/cockroachdb/pebble/blob/72a0b419362865176d2b23e7b1690d837fdc2659/compaction_iter.go#L622)):*

Whose responsibility is it to check `i.iter.Error()` if `i.iterKey` is nil after this `Next`? Some of our iterators implement `Error()` with `firstError(i.err, i.iter.Error())`, but `compactionIter` implements `Error()` by directly returning `i.err`.
</blockquote></details>

Good question.
(just repeating in-person discussion) we check in `compactionIter.Close()` which will get called because the iter is exhausted. I was going to add more defensive code, but noticed that `compaction.go` never calls `compactionIter.Error()` and relies on exhaustion.

PR #3090, opened 2023-11-20
[crl-release-23.2] db: remove compaction iter seqnum optimization, replace with assert

itsbilal:
Previously the compaction iterator would avoid user key comparisons in some circumstances when invariants around sequence numbers implied user keys must be different. This commit removes this optimization, replacing it with an assertion that the sequence number invariants hold.

This is done out of caution and with the realization that these user key comparisons form a relatively small portion of the CPU cost of a compaction.

Informs cockroachdb/pebble#3082.
Informs cockroachdb/cockroach#114421.

### DIFF
--- a/compaction_iter.go
+++ b/compaction_iter.go
@@ -15,7 +15,6 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/bytealloc"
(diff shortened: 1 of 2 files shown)


jbowens (approved):
:lgtm:

Thanks 🙏

nicktrav:
EM LGTM ✅

itsbilal:
TFTRs!

I'll run the 23.2 backport workflow on the crdb bump PR.

PR #3095, opened 2023-11-21
crl-release-23.2: db: enforce no split user keys in CheckOrdering

jbowens:
23.2 backport of #3085.

----

Adapt CheckOrdering to perform its overlap comparisons using the stricter requirement that adjacent files not contain the same user key. Since we still support the earlier format major versions that predate this requirement, this behavior is gated behind an OrderingInvariants enum variant determined by the active format major version.

Some of the tests that use more recent format major versions still created test cases with split user keys, and they needed to be adjusted.


itsbilal (approved):
LGTM

jbowens:
TFTR!

PR #3096, opened 2023-11-21
crl-release-23.2: db: add CanDeterministicallySingleDelete

jbowens:
23.2 backport of #3077.

----

Add a new specialized CanDeterministicallySingleDelete function that allows a user to inspect hidden internal keys within the current user key to determine whether the behavior of a SingleDelete of the current Key would be deterministic, assuming no writes are made to the key between Iterator creation and commit of the SingleDelete. The CanDeterministicallySingleDelete func may only be called on an Iterator oriented in the forward direction and does not change the external Iterator position.

During intent resolution, CockroachDB will write a SINGLEDEL tombstone to delete the intent if the intent's value indicates that it must have only been written once. Subtle logic that cuts across the stack makes this optimization possible, and one possible explanation for a recent instance of replica divergence (cockroachdb/cockroach#114421) is a bug in this logic. We anticipate using CanDeterministicallySingleDelete within CockroachDB's intent resolution to validate this logic at runtime.

There is some potential for this function to drive the decision to use SingleDelete when creating batches for local Engine application only.

Informs cockroachdb/cockroach#114492.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -145,6 +145,27 @@ func runIterCmd(d *datadriven.TestData, iter *Iterator, closeIter bool) string {
 			}
 			validityState = iter.NextWithLimit([]byte(parts[1]))
 			printValidityState = true
+		case "internal-next":
+			validity, keyKind := iter.internalNext()
+			switch validity {
+			case internalNextError:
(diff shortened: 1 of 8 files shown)


itsbilal (approved):
LGTM

Issue #1877, opened 2022-08-13
record: improve TestMetricsWithSync


nicktrav:
This test flaked [here](https://github.com/cockroachdb/pebble/runs/7815846812?check_suite_focus=true):

```console
--- FAIL: TestMetricsWithSync (0.01s)
    assertion_compare.go:342: 
        	Error Trace:	log_writer_test.go:392
        	Error:      	"5000" is not less than or equal to "4031"
        	Test:       	TestMetricsWithSync
        	Messages:   	[]
```

Digging into the test, it looks like we make use of sleeps, which is most likely where the flakiness is coming from.

Consider re-working this test to remove the sleeps in favor of some other synchronization methods.

nicktrav:
https://github.com/cockroachdb/pebble/actions/runs/3254278159/jobs/5342343471

PR #3098, opened 2023-11-22
pebble: panic on incorrect strict MemFS methods use

pavelkalinnikov:
Strict `MemFS` methods can be used by mistake on a non-strict `MemFS`. Panic if this happens, instead of (previously) ignoring this silently. This saves a test writer time being confused if they mis-use the type.

### DIFF
--- a/vfs/mem_fs.go
+++ b/vfs/mem_fs.go
@@ -773,22 +773,23 @@ func (f *memFile) Stat() (os.FileInfo, error) {
 }
 
 func (f *memFile) Sync() error {
-	if f.fs != nil && f.fs.strict {
-		f.fs.mu.Lock()
-		defer f.fs.mu.Unlock()
-		if f.fs.ignoreSyncs {
-			return nil
+	if f.fs == nil || !f.fs.strict {
+		return nil
+	}
+	f.fs.mu.Lock()
+	defer f.fs.mu.Unlock()
+	if f.fs.ignoreSyncs {
+		return nil
+	}
+	if f.n.isDir {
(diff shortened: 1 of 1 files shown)


pavelkalinnikov:
@jbowens PTAL

pavelkalinnikov:
There is one test failing on Windows, is this an unrelated flake?

```
=== RUN   TestMetricsWithSync
    log_writer_test.go:452: 
        	Error Trace:	D:\a\pebble\pebble\record\log_writer_test.go:452
        	Error:      	"5000" is not less than or equal to "0"
        	Test:       	TestMetricsWithSync
--- FAIL: TestMetricsWithSync (0.03s)
```

jbowens (approved):
Yeah, the Windows CI failure appears to be #1877 an unsquashed flake.

Should we instead expose a `StrictMemFS` type so that the type system prevents calling these methods? We could just export the methods on the `MemFS`, and expose a new `StrictMemFS` type that embeds the `MemFS` and exports the methods.

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3098#-:-NkGPrpK9sppPEV5Yaks:b-r81cu2)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pavelkalinnikov)

pavelkalinnikov:
Good idea adding a new type, I can do in a follow-up.

Alternatively, we may choose to make `MemFS` strict by default, because the performance problems preventing this can be fixed by #3104.

PR #3099, opened 2023-11-22
pebble: fix file link names in MemFS

pav-kv:
Before this commit, a file link returned the original file's name from the `Name()` method. `MemFS` dump would also print links with the original file names, which was confusing.


pav-kv:
Superseded by #3208.

PR #3101, opened 2023-11-22
[crl-release-23.2] *: handle non-Sets correctly in UserIteratorConfig

### LINKED ISSUE #3058: internal/rangekey: fix defragmentation of unsets/dels with ScanInternal

itsbilal:
Backport of #3069.

----

This change moves the `scanInternalIterator` use case of UserIteratorConfig over to something that's closer to rangeKeyCompactionTransform, where internal keys are coalesced using rangekey.coalesce, key order is maintained at ByTrailerDesc, and defragmentation happens by internal key using
keyspan.DefragmentInternal. The onlySets var is renamed to internalKeys and its meaning is reversed.

Fixes #3058.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -576,14 +576,18 @@ func runBatchDefineCmd(d *datadriven.TestData, b *Batch) error {
 			}
 			err = b.Merge([]byte(parts[1]), []byte(parts[2]), nil)
 		case "range-key-set":
-			if len(parts) != 5 {
-				return errors.Errorf("%s expects 4 arguments", parts[0])
+			if len(parts) < 4 || len(parts) > 5 {
+				return errors.Errorf("%s expects 3 or 4 arguments", parts[0])
+			}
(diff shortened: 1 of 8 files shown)


itsbilal:
Actually, let me combine all the backport-worthy fixes in one PR.

PR #3102, opened 2023-11-22
crl-release-23.2: backport shared ingestion fixes

itsbilal:
This change backports #3069 (1/1 commits) and #3074 (2/3 commits, the third being irrelevant for non-invariant builds)

---

### [scan_internal: Sort keys by kind before calling visitor](https://github.com/cockroachdb/pebble/commit/4f32102227690b9b57e9dbebe35f4b4987ffdfa1)

Previously, we were passing range keys that were sorted by trailer
descending right from ScanInternal. However, the caller isn't
interested in seqnum-based sorting, only kind-based trailer sorting.
This change zeroes seqnums before sorting the keys.

Unblocks https://github.com/cockroachdb/pebble/pull/3044.

### [*: set hideObsoletePoints=true for shared foreign sstables](https://github.com/cockroachdb/pebble/commit/673953238a2b69c2fb5f8c50e1cfb3799dbfb439)

Previously, we weren't setting hideObsoletePoints to true for
compaction iters created out of shared foreign sstables. This
change addresses that by passing that state through virtualState
in VirtualReader through to the old Reader.

Unblocks https://github.com/cockroachdb/pebble/pull/3044.

### [*: handle non-Sets correctly in UserIteratorConfig](https://github.com/cockroachdb/pebble/commit/de5ea48cfe13701ad85419f1b1388e3eb4875ecf)

This change moves the `scanInternalIterator` use case of
UserIteratorConfig over to something that's closer to
rangeKeyCompactionTransform, where internal keys are coalesced
using rangekey.coalesce, key order is maintained at ByTrailerDesc,
and defragmentation happens by internal key using
keyspan.DefragmentInternal. The onlySets var is renamed to internalKeys
and its meaning is reversed.

Fixes https://github.com/cockroachdb/pebble/issues/3058.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -1309,7 +1309,7 @@ func runSSTablePropertiesCmd(t *testing.T, td *datadriven.TestData, d *DB) strin
 	var v sstable.VirtualReader
 	props := r.Properties.String()
 	if m != nil && m.Virtual {
-		v = sstable.MakeVirtualReader(r, m.VirtualMeta())
+		v = sstable.MakeVirtualReader(r, m.VirtualMeta(), false /* isForeign */)
 		props = v.Properties.String()
 	}
 	if len(td.Input) == 0 {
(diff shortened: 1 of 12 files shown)


itsbilal:
TFTR!

PR #3103, opened 2023-11-22
db: add an error return value to BatchReader.Next

### LINKED ISSUE #3023: batch: detect corrupt batches during reading
Currently we don't differentiate between ordinary EOF and corruption when reading a batch. We should distinguish and use an appropriate error message when corurption is detected.

Informs cockroachdb/cockroach#113268.

https://github.com/cockroachdb/pebble/blob/ed45a776717595a33cce15c787e8583762c577a6/batch.go#L1515-L1537

jbowens:
Previously BatchReader.Next had a subtle API. If Next returned ok=false, the caller is responsible for checking len(r) and interpreting it as corruption if len(r) > 0. This commit introduces an explicit error return value.

Fixes #3023.

### DIFF
--- a/testdata/batch_reader
+++ b/testdata/batch_reader
@@ -1,0 +1,98 @@
+scan
+----
+Count: 0
+eof
+
+scan
+ffffffffffffffffffffffffffffffffffffffffffffffff
+----
+Count: 4294967295
+err: invalid key kind 0xff: pebble: invalid batch
+
+scan
+0000000000000000 01000000   # Seqnum = 0, Count = 1
+00 01 61                    # DEL "a"
+----
+Count: 1
+DEL: "a": ""
+eof
+
+scan
+0000000000000000 01000000   # Seqnum = 0, Count = 1
+01 01 62 01 62              # SET "b" = "b"
+----
+Count: 1
+SET: "b": "b"
+eof
+
+scan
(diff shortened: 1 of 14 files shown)


testdata/batch_reader
```diff
@@ -0,0 +1,98 @@
+scan
+----
+Count: 0
+eof
+
+scan
+ffffffffffffffffffffffffffffffffffffffffffffffff
```
itsbilal, line 7: press f to pay respects

itsbilal (approved):
LGTM

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3103#-:-NkGnGq0DSnNLFYPEnwJ:b-814g8k)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`testdata/batch_reader` line 7 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3103#gh-1403598293:-NkGnEx9C11UnulsbPx_:b2urym9) ([raw file](https://github.com/cockroachdb/pebble/blob/6da919bf1fbf2131486c18aa543a93977eae8077/testdata/batch_reader#L7)):*

press f to pay respects
</blockquote></details>

fffffffff

PR #3110, opened 2023-11-28
pebble: add global sync to MemFS

pav-kv:
This commit introduces `MemFS.Sync` method which persists the entire filesystem. It makes it possible to emulate a process crash and restart, as follows:

```
fs.Sync()                        // capture/snapshot the filesystem
fs.SetIgnoreSyncs(true)          // prevent changes from taking effect
db.Close()                       // shutdown, with no writes taking effect
fs.ResetToSyncedState()          // rollback to the snapshot
strictFS.SetIgnoreSyncs(false)   // allow changes again
db = Open(..., &Options{FS: fs}) // reopen on top of the reverted data
```

### DIFF
--- a/vfs/mem_fs.go
+++ b/vfs/mem_fs.go
@@ -114,6 +114,46 @@ func (y *MemFS) String() string {
 	return s.String()
 }
 
+// Sync makes the entire filesystem persistent. This is equivalent to calling
+// Sync() on every file and directory.
+//
+// This method can be used in conjunction with others to emulate a restart after
+// a process crash, as follows:
+//
+//	fs.Sync()                        // capture/snapshot the filesystem
+//	fs.SetIgnoreSyncs(true)          // prevent changes from taking effect
+//	db.Close()                       // shutdown, with no writes taking effect
+//	fs.ResetToSyncedState()          // rollback to the snapshot
+//	strictFS.SetIgnoreSyncs(false)   // allow changes again
+//	db = Open(..., &Options{FS: fs}) // reopen on top of the reverted data
+//
+// Without the fs.Sync() step above, the sequence emulates a power outage / OS
+// crash: everything that was not synced will be lost. With the fs.Sync(), only
+// data that was not flushed to the FS will be lost, but things that were
+// flushed and not synced will survive the process restart.
+//
+// For non-strict MemFS, Sync() is effectively a no-op because the latter anyway
+// behaves as if every operation is synced immediately.
+//
+// On the contrary, a strict MemFS does not sync at all by default. A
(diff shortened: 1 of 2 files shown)


jbowens (commented):
I'm confused. Wouldn't correct modeling of a crash be to _not_ sync the entire filesystem? We want to ensure that we've issued the correct syncs to all the individual files?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3110#-:-NkMW99u8HdgPAV-yab2:bahqwml)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @sumeerbhola)

pavelkalinnikov:
@jbowens The point of `fs.Sync()` is not so much in syncing the filesystem, but in checkpointing/forking for the purposes of the test, it in order to be able to rollback to exactly the same state later. In a real crash I'm imagining something like this happens:

1. the process is writing and flushing/fsyncing files
   - typically during this operation there is some "dirty" state that is not fully synced

2. crash
3. everything flushed so far is still in OS/FS buffers, so likely will persist
4. a restart will observe the FS at the state just before step (2), including the partially-written files that were not fully synced

In an emulated/unit-test environment we can't guarantee an immediate stop of the DB/server in step (2). So we snapshot the state of the filesystem just before step (2), and terminate the DB/server gracefully in step (2). The graceful stop will write some extra state and sync files, but we will ignore it all. In step (4) we rollback the FS to state that it was in just before the "crash" in step (2).

pavelkalinnikov:
A strict `MemFS` persists **only** the data that was synced. This is useful to make sure the necessary syncs are in place, and we can recover after a full system crash.

A non-strict `MemFS` persists everything. This is not super useful to test crashes, but still can be used to test clean restarts.

There is nothing in between. The `fs.(Sync/Checkpoint/Stash/etc)` kind of method allows testing restarts in "dirty" state in between "only the synced data is persisted" and "everything is persisted".

jbowens (approved):
Thanks for explaining. That makes sense. No need to consider it here, but I can imagine it being worthwhile to reset a random subset of unsynced data as well.

I think it would be possible to implement these various testing semantics as vfs.FS middleware, like we do in [`vfs/errorfs`](https://github.com/cockroachdb/pebble/blob/master/vfs/errorfs/errorfs.go). But maybe we aren't gonna need it.

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3110#-:-NkR40sL19Jnnxx28NXf:b-up3q2x)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sumeerbhola)

xxmplus:
Close to due to age

PR #2162, opened 2022-12-02
README: expand on format major versions and version compatibility

jbowens:
Expand the README with documentation on format major versions and the support for them in various Pebble versions. This is in preparation for dropping support for earlier format major versions.

PR #3115, opened 2023-11-30
crl-release-23.2: README: expand on format major versions and version compatibility

RaduBerinde:
The reason for backporting this is that we will tag a `v1.0.0` release from this branch and it would be good if that included this information.

Expand the README with documentation on format major versions and the support for them in Cockroach versions.

This is a partial cherry-pick of #2162. For now we are not mentioning deprecating any version.

We also unexport an unused format constant.


RaduBerinde:
TFTR!

PR #3116, opened 2023-11-30
metamorphic: refactor key tracking for single deletes 

jbowens:
This commit refactors the key tracking performed across objects and
decision-making around when to perform single deletes. Previously, the
metamorphic test tracked inflight writes to batches and restricted operations
to keys with inflight deletes (both SINGLEDEL and DEL). This commit takes an
alternative approach of tracking entire key histories per object and only
restricting per-object single deletes. This tracking alone is sufficient to
maintain single delete invariants for operations only performed directly
against the DB object.

Operations performed to batches and then committed to the database require
additional care. Rather than attempting to prevent violations from ever being
generated, this commit adapts the generator to detect violations and correct
them. When an applying or ingesting batch containing a single delete would
cause a violation of the single delete invariants, the generator generates
delete operations on the destination to avoid the violation. This allows single
deletes to be generated in a wider variety of circumstances and plays well with
the new multi-DB variants of the metamorphic test.

### DIFF
--- a/metamorphic/config.go
+++ b/metamorphic/config.go
@@ -53,6 +53,10 @@ const (
 	writerSingleDelete
 )
 
+func (o opType) isDelete() bool {
+	return o == writerDelete || o == writerDeleteRange || o == writerSingleDelete
+}
+
 type config struct {
 	// Weights for the operation mix to generate. ops[i] corresponds to the
 	// weight for opType(i).
--- a/metamorphic/generator.go
+++ b/metamorphic/generator.go
@@ -7,6 +7,7 @@ package metamorphic
 import (
 	"bytes"
 	"fmt"
+	"os"
 	"slices"
 
 	"github.com/cockroachdb/pebble"
@@ -182,6 +183,13 @@ func generate(rng *rand.Rand, count uint64, cfg config, km *keyManager) []op {
 	// TPCC-style deck of cards randomization. Every time the end of the deck is
 	// reached, we shuffle the deck.
 	deck := randvar.NewDeck(g.rng, cfg.ops...)
+
+	defer func() {
+		if r := recover(); r != nil {
+			fmt.Fprintln(os.Stderr, formatOps(g.ops))
+			panic(r)
+		}
+	}()
 	for i := uint64(0); i < count; i++ {
 		generators[deck.Int()]()
 	}
@@ -200,7 +208,7 @@ func (g *generator) add(op op) {
 // TODO(peter): make the size and distribution of keys configurable. See
 // keyDist and keySizeDist in config.go.
 func (g *generator) randKeyToWrite(newKey float64) []byte {
-	return g.randKeyHelper(g.keyManager.eligibleWriteKeys(), newKey, nil)
+	return g.randKeyHelper(g.keyManager.knownKeys(), newKey, nil)
 }
 
 // prefixKeyRange generates a [start, end) pair consisting of two prefix keys.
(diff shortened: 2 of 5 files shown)


jbowens (commented):
The first commit is from #3114 and can be reviewed there.

I've not rebased over master due to the outstanding bug (#3113).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3116#-:-NkavR3kC5M2IB_qrnLz:b-wer608)* status: 0 of 5 files reviewed, all discussions resolved

jbowens (commented):
TFTR!

@sumeerbhola I made a few additional changes:
a) Ingest handling wasn't right because ingesting a batch and applying it are not equivalent in the metamorphic test (eg, range deletions are always applied first if the batch is ingested).
b) I relaxed the set of scenarios we allow single deletes to include MERGE keys as long as the MERGE key is the only value-carrying key since the last delete.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3116#-:-Nl9rWIa3fsOjQhVt8Lj:b-nfm5bx)* status: 1 of 5 files reviewed, 4 unresolved discussions (waiting on @sumeerbhola)

sumeerbhola (approved):
:lgtm: 

> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3116#-:-NlAgqaBAbm0TXqZ_Vhs:bg7igkv)* status: all files reviewed, 6 unresolved discussions (waiting on @jbowens)

___
*[`metamorphic/key_manager.go` line 458 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3116#-NlAgHjv5xiBnvNsBopp:-NlAgHjv5xiBnvNsBopq:bm5klkz) ([raw file](https://github.com/cockroachdb/pebble/blob/92b781a86895b5d8aec578b2255fa2157bfcc2ec/metamorphic/key_manager.go#L458)):*
> ```Go
> 			k.mergeKeysInto(batchID, s.dbID, func(src, dst *keyMeta, ts int) {
> 				collapsed := src.history.collapsed()
> 				for _, op := range collapsed {
> ```

is this loop any different from calling `mergeInto`?
___
*[`metamorphic/key_manager.go` line 626 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3116#-NlAgQWB9ycWIF0jp5QR:-NlAgQWB9ycWIF0jp5QS:bgwfs9l) ([raw file](https://github.com/cockroachdb/pebble/blob/92b781a86895b5d8aec578b2255fa2157bfcc2ec/metamorphic/key_manager.go#L626)):*
> ```Go
> 	var ret keyHistory
> 	// When collapsing a batch, any range deletes are semantically applied
> 	// first. Look for any range deletes any apply them.
> ```

nit: and apply ...

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3116#-:-NlLT5RvDr8GxpywQPcR:b-814g8k)* status: all files reviewed, 5 unresolved discussions (waiting on @sumeerbhola)

___
*[`metamorphic/key_manager.go` line 458 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3116#-NlAgHjv5xiBnvNsBopp:-NlLSTG--uH-l_GtadML:bm3jluk) ([raw file](https://github.com/cockroachdb/pebble/blob/92b781a86895b5d8aec578b2255fa2157bfcc2ec/metamorphic/key_manager.go#L458)):*

is this loop any different from calling `mergeInto`?
</blockquote></details>

Yeah the loop is the same. Refactored to call mergeInto on a temporarily constructed keyMeta with the collapsed src history.

Issue #3025, opened 2023-10-30
db: periodically sync the data directory during file deletions


jbowens:
We pace file deletions to avoid spikes in IO. However, the data directory is never synced, so IO is not necessarily scheduled yet. This could theoretically allow a buildup of unsynced file deletions. When the data directory is finally synced, IO may spike.

Noticed during the investigation of cockroachlabs/support#2673. It's unlikely to be the root cause of that particular.

Jira issue: PEBBLE-66

sidntrivedi012:
Hey @jbowens, I would like to contribute and fix this. Can you share some more context related to the issue? Thanks 🙌

jbowens:
@sidntrivedi012 — Sure! File deletions are performed by a `cleanupManager` goroutine that paces itself to smooth deletions out over time:

https://github.com/cockroachdb/pebble/blob/844f0582c2ebbbb0e862459e843ef144792ff97b/cleaner.go#L149-L172

Deletions are performed by calling [(objstorage.Provider).Remove](https://github.com/cockroachdb/pebble/blob/844f0582c2ebbbb0e862459e843ef144792ff97b/objstorage/objstorage.go#L235).

I think we want to edit the main loop to call [(objstorage.Provider).Sync](https://github.com/cockroachdb/pebble/blob/844f0582c2ebbbb0e862459e843ef144792ff97b/objstorage/objstorage.go#L240) after removing a file if the size of all the files deleted since the last call to `Sync` is greater than some threshold; say 64 MB.

sidntrivedi012:
@jbowens Got it. Thanks for the info. Hoping to get the PR up for it asap.

PR #3122, opened 2023-12-04
db: call Sync after deletion threshold

darshanime:
Add periodic calls to `(objstorage.Provider).Sync`

Closes: #3025

### DIFF
--- a/cleaner.go
+++ b/cleaner.go
@@ -27,10 +27,11 @@ type DeleteCleaner = base.DeleteCleaner
 type ArchiveCleaner = base.ArchiveCleaner
 
 type cleanupManager struct {
-	opts            *Options
-	objProvider     objstorage.Provider
-	onTableDeleteFn func(fileSize uint64)
-	deletePacer     *deletionPacer
+	opts                       *Options
+	objProvider                objstorage.Provider
+	onTableDeleteFn            func(fileSize uint64)
+	deletePacer                *deletionPacer
+	deletedFileSizePendingSync uint64
 
 	// jobsCh is used as the cleanup job queue.
 	jobsCh chan *cleanupJob
@@ -160,7 +161,7 @@ func (cm *cleanupManager) mainLoop() {
 			} else {
 				cm.maybePace(&tb, of.fileType, of.fileNum, of.fileSize)
 				cm.onTableDeleteFn(of.fileSize)
-				cm.deleteObsoleteObject(fileTypeTable, job.jobID, of.fileNum)
+				cm.deleteObsoleteObjectAndSyncIfRequired(job.jobID, of.fileNum, of.fileSize)
 			}
 		}
 		cm.mu.Lock()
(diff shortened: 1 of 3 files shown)


RaduBerinde (commented):
Thanks for the contribution!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3122#-:-NkvWXAX51dBEUzC8kB0:b3tf9g9)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @darshanime)

___
*[`cleaner.go` line 176 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3122#-NkvVbmr62Vuf11rZes3:-NkvVbmr62Vuf11rZes4:b-9r61kn) ([raw file](https://github.com/cockroachdb/pebble/blob/8db6a97aab7547365795a3e451cd1b28256e5f8e/cleaner.go#L176)):*
> ```Go
> 
> func (cm *cleanupManager) deleteObsoleteObjectAndSyncIfRequired(jobID int, fileNum base.DiskFileNum, fileSize uint64) {
> 	var deleteThresholdForSync uint64 = 64 * 1024 * 1024 // 64 MB
> ```

[nit] this should be a constant around the top of the file (next to jobsQueueDepth)
___
*[`cleaner.go` line 188 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3122#-NkvVk1sCTYvZkzyCGxa:-NkvVk1sCTYvZkzyCGxb:bobea02) ([raw file](https://github.com/cockroachdb/pebble/blob/8db6a97aab7547365795a3e451cd1b28256e5f8e/cleaner.go#L188)):*
> ```Go
> 	}
> 
> 	if err := cm.objProvider.Sync(); err == nil {
> ```

We should log the error here. I think we should still reset `deletedFileSizePendingSync` even in the error case, it's unlikely that retrying after every object will help.
___
*[`cleaner.go` line 265 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3122#-NkvW3QpA1ST02a2ucKm:-NkvW3QpA1ST02a2ucKn:b-b0k8lb) ([raw file](https://github.com/cockroachdb/pebble/blob/8db6a97aab7547365795a3e451cd1b28256e5f8e/cleaner.go#L265)):*
> ```Go
> }
> 
> func (cm *cleanupManager) deleteObsoleteObject(
> ```

This needs an explanation of the return value. Though I don't think it's worth it, the case of the object not existing is unlikely to happen in practice
___
*[`testdata/cleaner` line 190 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3122#-NkvWOzm8yXCOPtEuJST:-NkvWOzm8yXCOPtEuJSU:b-9t8pu5) ([raw file](https://github.com/cockroachdb/pebble/blob/8db6a97aab7547365795a3e451cd1b28256e5f8e/testdata/cleaner#L190)):*
> ```
> close: db1/000123.sst
> 
> create-bogus-file db1/000456.sst 100000000
> ```

Add a comment explaining that we're setting a large file size to verify that the cleaner issues a sync.

darshanime:
@RaduBerinde thanks for the review! I have addressed the comments (pardon the delay!)

PR #3123, opened 2023-12-04
testdata: regenerate db-stage-x at more recent format

RaduBerinde:
In preparation for deprecating all formats older than `FormatFlushableIngest`, we regenerate these test databases at this format. We replace the C++ program that used RocksDB with a Go program which uses Pebble.

Informs: #3064


testdata/make-db.cc
```diff
@@ -1,137 +0,0 @@
...
-}
-
-template <typename T>
-inline T FromString(const std::string &str, T val = T()) {
-  std::istringstream s(str);
-  s >> val;
-  return val;
-}
-
-int main(int argc, char** argv) {
```
itsbilal, line 29: goodbye

tool/testdata/wal_dump
```diff
@@ -3,106 +3,106 @@ wal dump
...
 --key=pretty:test-comparer
 --value=%x
 ----
-000006.log
-0(22) seq=6 count=1
+000005.log
+0(22) seq=15 count=1
     SET(test formatter: foo,66697665)
-29(22) seq=7 count=1
+33(22) seq=16 count=1
```
itsbilal, line 89: Most of this seems to be due to our sequence number reservation work for disagg storage, which is now unnecessary. We can consider undoing it if we think that's valuable.

itsbilal (approved):
LGTM :shipit:

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3123#-:-Nkqtvsx6BKkYRNve7rV:b-814g8k)* status: 0 of 62 files reviewed, 2 unresolved discussions (waiting on @itsbilal)

___
*[`tool/testdata/wal_dump` line 89 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3123#gh-1414458941:-NkqtsQSCYTA23StMWug:b-kb1iwg) ([raw file](https://github.com/cockroachdb/pebble/blob/b526f51160311d964e7d25f7458d5757bb31556e/tool/testdata/wal_dump#L89)):*

Most of this seems to be due to our sequence number reservation work for disagg storage, which is now unnecessary. We can consider undoing it if we think that's valuable.
</blockquote></details>

I don't think it matters much either way.

PR #3126, opened 2023-12-05
sstable: remove FieldByName call in Properties.load

RaduBerinde:
This call has a lot of overhead. We replace this call with `FieldByIndex`. For this to work, we have to make sure we generate the correct sequence of field struct indexes.

This should fix some microbenchmark regressions.

### DIFF
--- a/sstable/properties.go
+++ b/sstable/properties.go
@@ -25,7 +25,7 @@ var propBoolFalse = []byte{'0'}
 
 var propOffsetTagMap = make(map[uintptr]string)
 
-func generateTagMaps(t reflect.Type) {
+func generateTagMaps(t reflect.Type, indexPrefix []int) {
 	for i := 0; i < t.NumField(); i++ {
 		f := t.Field(i)
 		if f.Type.Kind() == reflect.Struct {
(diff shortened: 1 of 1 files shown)


sumeerbhola (approved):
nice!
is it possible to write a small benchmark to confirm the reduced allocation?

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3126#-:-NkwSYZh19RY2eTxlqiQ:b1i2pe7)* status: 0 of 1 files reviewed, all discussions resolved

RaduBerinde (commented):
Sure, done:
```
name               old time/op    new time/op    delta
PropertiesLoad-10    9.53µs ± 0%    3.57µs ± 0%  -62.54%  (p=0.016 n=5+4)

name               old alloc/op   new alloc/op   delta
PropertiesLoad-10    5.00kB ± 0%    2.95kB ± 0%  -40.97%  (p=0.008 n=5+5)

name               old allocs/op  new allocs/op  delta
PropertiesLoad-10       102 ± 0%        28 ± 0%  -72.55%  (p=0.008 n=5+5)
```

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3126#-:-NkwyHl88OjlePLo5bEJ:bsmersr)* status: 0 of 2 files reviewed, all discussions resolved

RaduBerinde (commented):
Still seems like a lot, I will try code generation as well and see how it compares.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3126#-:-NkwyXgU03VgoRdUWJat:b-4rt5jv)* status: 0 of 2 files reviewed, all discussions resolved

RaduBerinde:
I tried code generation and the allocations are the same (this is what I tried: https://github.com/cockroachdb/pebble/compare/master...RaduBerinde:pebble:load-reflect-fix-2?expand=1). I don't think it's worth the extra complexity.

PR #3127, opened 2023-12-06
crl-release-23.2: sstable: remove FieldByName call in Properties.load

RaduBerinde:
This call has a lot of overhead. We replace this call with `FieldByIndex`. For this to work, we have to make sure we generate the correct sequence of field struct indexes.

This should fix some microbenchmark regressions.

See benchmark results in https://github.com/cockroachdb/pebble/pull/3126


jbowens (approved):
lgtm

PR #3129, opened 2023-12-06
db: fix regression in table cache hot path

RaduBerinde:
#### db: add benchmark for table cache hit path

Add a benchmark for the hot path of `findNode` (when we hit the
cache).

#### db: fix regression in table cache hot path

A seemingly innocuous change in `tableCacheShard.findNode` resulted in
a new allocation. We named the return variable, then we captured it in
the `closeHook` callback, which I guess means that the pointer itself
needs to be put on the heap.

With this change `BenchmarkTableCacheHotPath` goes down from 1
alloc/op to 0, and the CRDB
`RefreshRange/linear-keys/refresh_window=[99.00,99.00]` benchmark goes
from 101 allocs/op to 61.

Informs https://github.com/cockroachdb/cockroach/issues/115215

### DIFF
--- a/table_cache.go
+++ b/table_cache.go
@@ -755,27 +755,30 @@ func (c *tableCacheShard) unrefValue(v *tableCacheValue) {
 // findNode returns the node for the table with the given file number, creating
 // that node if it didn't already exist. The caller is responsible for
 // decrementing the returned node's refCount.
-func (c *tableCacheShard) findNode(
-	meta *fileMetadata, dbOpts *tableCacheOpts,
-) (v *tableCacheValue) {
+func (c *tableCacheShard) findNode(meta *fileMetadata, dbOpts *tableCacheOpts) *tableCacheValue {
+	v := c.findNodeInternal(meta, dbOpts)
+
 	// Loading a file before its global sequence number is known (eg,
 	// during ingest before entering the commit pipeline) can pollute
 	// the cache with incorrect state. In invariant builds, verify
 	// that the global sequence number of the returned reader matches.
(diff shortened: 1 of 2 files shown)


sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3129#-:-Nl-wckE5teD4YdDpkcn:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`-- commits` line 13 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3129#-Nl-warrAdQYSy67NV8E):*
nice find!
___
*[`table_cache_test.go` line 1232 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3129#-Nl-vDXf5AtWNMIs4Z5d:-Nl-vDXf5AtWNMIs4Z5e:bmfnscl) ([raw file](https://github.com/cockroachdb/pebble/blob/469003d1f235485ae6efe00794b95269dc29990a/table_cache_test.go#L1232)):*
> ```Go
> 
> 	cache := &tableCacheShard{}
> 	// NB: The table cache size of 200 is required for the expected test values.
> ```

I didn't understand this "... is required for the expected test values" since (a) I don't see where this benchmark has any expectations, (b) we seem to be retrieving the same table repeatedly, so wouldn't a size of 1 be sufficient to only get cache hits? What am I missing?

PR #3133, opened 2023-12-07
db: fix flaky TestLatestRefCounting

### LINKED ISSUE #3043: db: TestLatestRefCounting is flaky
```
go test -run TestLatestRefCounting -exec 'stress -p 16'
```

```
--- FAIL: TestLatestRefCounting (0.00s)
    version_set_test.go:228:
        	Error Trace:	/Users/jackson/go/src/github.com/cockroachdb/pebble/version_set_test.go:228
        	Error:      	Not equal:
        	            	expected: 1
        	            	actual  : 0
        	Test:       	TestLatestRefCounting
FAIL


ERROR: exit status 1

362 runs completed, 1 failures, over 1s
context canceled
FAIL
exit status 1
FAIL	github.com/cockroachdb/pebble	0.864s
```

RaduBerinde:
This check is racy in both directions:
 - we can check "too early", in that an asynchronous `collectTableStats` might still be holding a reference
 - we can check "too late", in that cleanup of the obsolete table has already happened

Since object cleanup is not the main point of this test, this commit removes this check altogether.

Fixes #3043

### DIFF
--- a/version_set_test.go
+++ b/version_set_test.go
@@ -223,10 +223,6 @@ func TestLatestRefCounting(t *testing.T) {
 	require.False(t, ok)
 	require.Equal(t, 0, int(m2.FileBacking.VirtualizedSize.Load()))
 	checkBackingSize(t, d)
-
-	// Make sure that the backing file is added to the obsolete tables list.
-	require.Equal(t, 1, len(d.mu.versions.obsoleteTables))
-
 }
 
 // TODO(bananabrick): Convert TestLatestRefCounting and this test into a single


RaduBerinde:
> we can check "too late", in that cleanup of the obsolete table has already happened

To be clear, the last point is relevant if we add a retry loop, dropping the mutex while we wait.

itsbilal (approved):
LGTM. Thanks for doing this! I was trying to avoid the "too late" case from happening and didn't get around to it just yet.

RaduBerinde:
Yeah, I considered disabling cleanup but I don't think it's worth it. TFTR!

PR #3137, opened 2023-12-07
record: fix TestMetricsWithSync flakiness

### LINKED ISSUE #1877: record: improve TestMetricsWithSync

RaduBerinde:
This commit attempts to fix a very rare flake in this test (which I could not reproduce). It is conceivable that the background tasks get delayed and thus the write latency they observe is lower.

We improve this by running `Close()` at the start of the sleep, which will force a flush as soon as possible. We also up the sleep to 100ms.

Fixes #1877

### DIFF
--- a/record/log_writer_test.go
+++ b/record/log_writer_test.go
@@ -411,7 +411,6 @@ func TestMetricsWithoutSync(t *testing.T) {
 func TestMetricsWithSync(t *testing.T) {
 	f := &syncFileWithWait{}
 	f.syncWG.Add(1)
-	writeTo := &prometheusgo.Metric{}
 	syncLatencyMicros := prometheus.NewHistogram(prometheus.HistogramOpts{
 		Buckets: []float64{0,
 			float64(time.Millisecond),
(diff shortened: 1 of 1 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3137#-:-Nl92pe87vmhlbeNsIVH:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @itsbilal)

Issue #2780, opened 2023-08-01
Make batchMaxRetainedSize configurable


lahsivjar:
I have a use case with pebble where the `pebble.Batch` size would usually be greater than `1 MB`. The current logic in pebble would GC the data held by the `pebble.Batch` if the data exceeds `1MB` ([ref](https://github.com/cockroachdb/pebble/blob/9c487492d79eae251864a428450306a7c0e3c602/batch.go#L1331-L1337)).

I was hoping that we can make the `batchMaxRetainedSize` configurable, what do others think about this? (If it sounds like a feasible idea I can create a PR).

jbowens:
Hi @lahsivjar, yeah, I think it's feasible, at least for instances where `d.NewBatch` is used.

lahsivjar:
👋 folks (@RaduBerinde @jbowens ), curious about the release schedule for this. I see that the latest releases are cut from a 1.1 branch so don't have this commit (or other commits), is there any plan for v2 release or is it possible to get this in the 1.x release?

lahsivjar:
> I will need to add an upgrade command to v1 and you would need to use it to upgrade any old stores. Does that work?

That would be great and it does work for our use-cases 🙏

RaduBerinde:
@lahsivjar I tagged a v2.0.0 release. Please check out the updated README around fomat versions and upgrading existing stores: https://github.com/cockroachdb/pebble/blob/8d697d1a30f28c4f1ccdf1ee508f7c02d1f7f181/README.md

PR #3139, opened 2023-12-08
Configurable initial and max retained batch sizes

lahsivjar:
Closes: https://github.com/cockroachdb/pebble/issues/2780

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -30,10 +30,11 @@ import (
 )
 
 const (
-	batchInitialSize     = 1 << 10 // 1 KB
-	batchMaxRetainedSize = 1 << 20 // 1 MB
-	invalidBatchCount    = 1<<32 - 1
-	maxVarintLen32       = 5
+	invalidBatchCount = 1<<32 - 1
+	maxVarintLen32    = 5
+
+	defaultBatchInitialSize     = 1 << 10 // 1 KB
+	defaultBatchMaxRetainedSize = 1 << 20 // 1 MB
 )
 
 // ErrNotIndexed means that a read operation on a batch failed because the
@@ -271,6 +272,7 @@ type batchInternal struct {
 	cmp            Compare
 	formatKey      base.FormatKey
 	abbreviatedKey AbbreviatedKey
+	opts           batchOptions
 
 	// An upper bound on required space to add this batch to a memtable.
 	// Note that although batches are limited to 4 GiB in size, that limit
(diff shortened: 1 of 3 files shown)


lahsivjar:
@jbowens Would appreciate a quick review whenever you have time.

lahsivjar:
@jbowens Friendly ping, any feedback on the PR is appreciated.

inge4pres:
This is very cool and I also have a similar use case 👍🏼 
Plus, there is no breaking change to the public API so it would be nice to have this landed 🙏🏼

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3139)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @jbowens and @lahsivjar)

___
*[`batch.go` line 275 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3139#-NvCvqCl0RWOmVnsOQXH:-NvCwY_ZCgr9K4WndMzX:b-gfmz9p) ([raw file](https://github.com/cockroachdb/pebble/blob/43fd2e881a983e8dcc431eccf1fc878fdf79bb19/batch.go#L275)):*

I think this should not be a pointer—otherwise, we'll suffer an additional heap allocation for every new batch initialization
</blockquote></details>

+1. `ensureDefaults` should just modify in-place and return nothing.
___
*[`batch.go` line 2421 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3139#-NvCwbUvD-6Ljics6F4j:-NvCwbUvD-6Ljics6F4k:b-oo9i76) ([raw file](https://github.com/cockroachdb/pebble/blob/43fd2e881a983e8dcc431eccf1fc878fdf79bb19/batch.go#L2421)):*
> ```Go
> func WithInitialSizeBytes(s int) BatchOption {
> 	return func(opts *batchOptions) {
> 		if s > 0 {
> ```

[nit] The special handling of s <= 0 is odd here, it shouldn't be valid to use such a value.

lahsivjar:
@jbowens @RaduBerinde Thanks both for your reviews. I have addressed the comments now.

RaduBerinde (commented):
Thanks! Can you squash the commits?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3139#-:-NvD6fwi3-MkIFCIsW1a:b-nkc2z)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @jbowens and @lahsivjar)

___
*[`batch_test.go` line 1744 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3139#-NvD6e0o4KI2eHvFcLr-:-NvD6e0o4KI2eHvFcLr0:b-s3vmj1) ([raw file](https://github.com/cockroachdb/pebble/blob/b0e5e0c8163332f393990119b8492b759eec4792/batch_test.go#L1744)):*
> ```Go
> 		},
> 		{
> 			name: "with_invalid_values",
> ```

I think we need to remove this test.

lahsivjar:
> Thanks! Can you squash the commits?

@RaduBerinde Done! Also removed the bad test.

RaduBerinde (approved):
Thank you!

PR #3141, opened 2023-12-08
db: convert TestElideTombstone to datadriven

jbowens:
Convert the TestElideTombstone unit test to a datadriven test. One of the test cases was removed because it tests a case that will be imminently unsupported (split user keys).


RaduBerinde (approved):
Nice!

jbowens:
TFTR!

PR #3143, opened 2023-12-08
db: compactionIter.nextInStripe skips RANGEDEL covered keys in the sa…

sumeerbhola:
…me stripe

This change allows us to stop remembering to check for visibly covered keys in callers. For example:
- deleteSizedNext did not account for visibly covered SETs when it did the size computation.
- singleDeleteNext was willing to consume the SINGLEDEL even if the SET was covered by a RANGEDEL (though this is not a bug, since everything below the RANGEDEL is deleted).

This change does effect mergeNext in that MERGE plus an older RANGEDEL will not get converted to a SET. This is not a semantic change to MERGE behavior.

### DIFF
--- a/compaction_iter.go
+++ b/compaction_iter.go
@@ -423,6 +423,10 @@ func (i *compactionIter) Next() (*InternalKey, []byte) {
 			return &i.key, i.value
 		}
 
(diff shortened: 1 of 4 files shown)


sumeerbhola:
TFTR!

PR #3146, opened 2023-12-11
metamorphic: copy shared storage contents when saving data

itsbilal:
Previously, we'd copy the contents of MemFS if a metamorphic test failed, but not the contents of shared storage if that was enabled. This change updates maybeSaveData to also copy over contents of the shared in-mem storage into a directory of its own.

### DIFF
--- a/metamorphic/options.go
+++ b/metamorphic/options.go
@@ -100,9 +100,11 @@ func parseOptions(
 				return true
 			case "TestOptions.shared_storage_enabled":
 				opts.sharedStorageEnabled = true
+				sharedStorage := remote.NewInMem()
 				opts.Opts.Experimental.RemoteStorage = remote.MakeSimpleFactory(map[remote.Locator]remote.Storage{
-					"": remote.NewInMem(),
+					"": sharedStorage,
 				})
+				opts.sharedStorageFS = sharedStorage
 				if opts.Opts.Experimental.CreateOnShared == remote.CreateOnSharedNone {
 					opts.Opts.Experimental.CreateOnShared = remote.CreateOnSharedAll
 				}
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3146#-:-NlPlVUsE9YwINd-xQa6:bnfp4nl)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @itsbilal)

___
*[`metamorphic/test.go` line 272 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3146#-NlPlN0tEwlow1TrXdSO:-NlPlN0tEwlow1TrXdSP:b-sqq18a) ([raw file](https://github.com/cockroachdb/pebble/blob/7bc97dd6b7f5848e104d6b14a596c11969ca0b57/metamorphic/test.go#L272)):*
> ```Go
> 		objs, err := fs.List("", "")
> 		if err != nil {
> 			t.opts.Logger.Infof("unable to clone shared storage: %s: %v", t.dir, err)
> ```

[nit] might be worth moving all this code into a `saveDataInternal` which returns an error and do the logging above it

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3146#-:-NlUOJUD9gbGNdHVhvvw:b-814g8k)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @aadityasondhi)

___
*[`metamorphic/test.go` line 272 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3146#-NlPlN0tEwlow1TrXdSO:-NlUOKFU38EnYDqvT2_I:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/7bc97dd6b7f5848e104d6b14a596c11969ca0b57/metamorphic/test.go#L272)):*

[nit] might be worth moving all this code into a `saveDataInternal` which returns an error and do the logging above it
</blockquote></details>

Done.

PR #3147, opened 2023-12-11
sstable: fix incorrect range key mask in virtualLast() 

itsbilal:
Previously, when a virtual sstable's iterator was SeekLT'd
past its end key or had Last() called on it, and the virtual
sstable had an inclusive end bound, we'd do a SeekGE on the end
key in `virtualLast()`. This had the potential of being inconsistent
with the rest of the iterator stack on the positioning of the
BoundLimitedBlockPropertyFilter in range-key combined iteration
with range key masking enabled, as the sstable iterator expects
that to always be set to a range key that's behind of us in the
direction we're iterating in. This saves some key comparisons
when checking for whether the current key is masked.

This change passes a SeekGEFlag for cases where we translate a
Last/SeekLT to a SeekGE, that overrides the direction specified
to loadIndex/loadData, forcing both the lower and upper bounds
of a BoundLimitedBlockPropertyFilter to be checked for when
the direction is ambiguous.

Fixes https://github.com/cockroachdb/pebble/issues/3128.

### DIFF
--- a/ingest_test.go
+++ b/ingest_test.go
@@ -587,7 +587,7 @@ func TestExcise(t *testing.T) {
 	}()
 
 	var opts *Options
-	reset := func() {
+	reset := func(blockSize int) {
 		if d != nil {
 			require.NoError(t, d.Close())
 		}
@@ -595,6 +595,10 @@ func TestExcise(t *testing.T) {
 		mem = vfs.NewMem()
 		require.NoError(t, mem.MkdirAll("ext", 0755))
 		opts = &Options{
+			BlockPropertyCollectors: []func() BlockPropertyCollector{
+				sstable.NewTestKeysBlockPropertyCollector,
+			},
+			Comparer:              testkeys.Comparer,
 			FS:                    mem,
 			L0CompactionThreshold: 100,
 			L0StopWritesThreshold: 100,
@@ -603,7 +607,9 @@ func TestExcise(t *testing.T) {
 				flushed = true
 			}},
 			FormatMajorVersion: FormatVirtualSSTables,
-			Comparer:           testkeys.Comparer,
+		}
+		if blockSize != 0 {
+			opts.Levels = append(opts.Levels, LevelOptions{BlockSize: blockSize, IndexBlockSize: 32 << 10})
 		}
 		// Disable automatic compactions because otherwise we'll race with
 		// delete-only compactions triggered by ingesting range tombstones.
@@ -616,12 +622,21 @@ func TestExcise(t *testing.T) {
 		d, err = Open("", opts)
 		require.NoError(t, err)
 	}
-	reset()
+	reset(0 /* blockSize */)
 
 	datadriven.RunTest(t, "testdata/excise", func(t *testing.T, td *datadriven.TestData) string {
 		switch td.Cmd {
 		case "reset":
-			reset()
+			var blockSize int
+			for i := range td.CmdArgs {
+				switch td.CmdArgs[i].Key {
+				case "tiny-blocks":
+					blockSize = 1
+				default:
+					return fmt.Sprintf("unexpected arg: %s", td.CmdArgs[i].Key)
+				}
+			}
+			reset(blockSize)
 			return ""
 		case "reopen":
 			require.NoError(t, d.Close())
@@ -687,9 +702,23 @@ func TestExcise(t *testing.T) {
 			return runGetCmd(t, td, d)
 
 		case "iter":
-			iter, _ := d.NewIter(&IterOptions{
+			opts := &IterOptions{
 				KeyTypes: IterKeyTypePointsAndRanges,
-			})
+			}
+			for i := range td.CmdArgs {
+				switch td.CmdArgs[i].Key {
+				case "range-key-masking":
+					opts.RangeKeyMasking = RangeKeyMasking{
+						Suffix: []byte(td.CmdArgs[i].Vals[0]),
+						Filter: func() BlockPropertyFilterMask {
+							return sstable.NewTestKeysMaskingFilter()
+						},
+					}
+				default:
+					return fmt.Sprintf("unexpected argument: %s", td.CmdArgs[i].Key)
+				}
+			}
+			iter, _ := d.NewIter(opts)
 			return runIterCmd(td, iter, true)
 
 		case "lsm":
--- a/sstable/reader_iter_single_lvl.go
+++ b/sstable/reader_iter_single_lvl.go
@@ -5,6 +5,7 @@
 package sstable
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"unsafe"
@@ -440,7 +441,7 @@ func (i *singleLevelIterator) readBlockForVBR(
 // that a block is excluded according to its properties but only if its bounds
 // fall within the filter's current bounds.  This function consults the
 // apprioriate bound, depending on the iteration direction, and returns either
-// `blockIntersects` or `blockMaybeExcluded`.
+// `blockIntersects` or `blockExcluded`.
 func (i *singleLevelIterator) resolveMaybeExcluded(dir int8) intersectsResult {
 	// TODO(jackson): We could first try comparing to top-level index block's
 	// key, and if within bounds avoid per-data block key comparisons.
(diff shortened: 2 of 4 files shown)


itsbilal:
Based on conversation on Slack, it would make sense to make a VSST-specific virtualLastSeek() or seekLEVirtual() or so, instead of having `virtualLast` use the public-facing `SeekGE` with a custom flag. I'll make that change soon.

itsbilal:
This is RFAL now.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3147)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`sstable/reader_iter_single_lvl.go` line 466 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3147#-NlVxxtc7xqYS-iwamRr:-NlVxxtc7xqYS-iwamRs:b4p3s0d) ([raw file](https://github.com/cockroachdb/pebble/blob/2f75cd703694e241145f4d1a9fc104350e0dd3d0/sstable/reader_iter_single_lvl.go#L466)):*
> ```Go
> 	if dir > 0 {
> 		// Forward iteration.
> 		if i.bpfs.boundLimitedFilter.KeyIsWithinUpperBound(i.index.Key().UserKey) {
> ```

Consider restructuring along the lines of:

```
// We don't need to check the upper bound during reverse iteration.
if dir >= 0 && !KeyIsWithinUpperBound {
  return blockIntersects
}
// We don't need to check the lower bound during forward iteration.
if dir <= 0 {
  peekKey, _ := i.index.Prev();
  _, _ = i.index.Next()
  if peekKey == nil || !IsWithinLowerBound() {
    return blockIntersects
  }
}
return blockExcluded
```
___
*[`sstable/reader_iter_two_lvl.go` line 912 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3147#-NlVvJ6xDoT0r2AtN1Zq:-NlVvJ6xDoT0r2AtN1Zr:bz2atp3) ([raw file](https://github.com/cockroachdb/pebble/blob/2f75cd703694e241145f4d1a9fc104350e0dd3d0/sstable/reader_iter_two_lvl.go#L912)):*
> ```Go
> }
> 
> func (i *twoLevelIterator) skipForward(dir int8) (*InternalKey, base.LazyValue) {
> ```

By itself, this is very confusing. It has "forward" in the name, so what does the direction do? Can it be zero? Why doesn't skipBackward need it? We should document these functions the best we can.

It might also help to make the direction a type and document it and add some constants.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3147#-:-NlZhfKf4alwvm457nRp:b-814g8k)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 466 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3147#-NlVxxtc7xqYS-iwamRr:-NlZhgE--Py284nnVQiX:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/2f75cd703694e241145f4d1a9fc104350e0dd3d0/sstable/reader_iter_single_lvl.go#L466)):*

Consider restructuring along the lines of:

```
// We don't need to check the upper bound during reverse iteration.
if dir >= 0 && !KeyIsWithinUpperBound {
  return blockIntersects
}
// We don't need to check the lower bound during forward iteration.
if dir <= 0 {
  peekKey, _ := i.index.Prev();
  _, _ = i.index.Next()
  if peekKey == nil || !IsWithinLowerBound() {
    return blockIntersects
  }
}
return blockExcluded
```
</blockquote></details>

Done.
___
*[`sstable/reader_iter_two_lvl.go` line 912 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3147#-NlVvJ6xDoT0r2AtN1Zq:-NlZhh0u0cZhQOc0DR8S:bm8wnru) ([raw file](https://github.com/cockroachdb/pebble/blob/2f75cd703694e241145f4d1a9fc104350e0dd3d0/sstable/reader_iter_two_lvl.go#L912)):*

By itself, this is very confusing. It has "forward" in the name, so what does the direction do? Can it be zero? Why doesn't skipBackward need it? We should document these functions the best we can.

It might also help to make the direction a type and document it and add some constants.
</blockquote></details>

I added a comment here to help address this confusion. But basically, `dir` is indicating if we can assume a direction about the rest of the iterator stack. It's only used to resolve the boundLimitedBlockPropertyFilter, and nothing else.

`skipBackward` doesn't need it because we currently don't do any reverse iteration operations as part of a forward iteration operation, even for virtual SSTs. If we did that, we'd need a `dir` there too.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3147)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 464 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3147#-Nlifaze3mNy3Lgu6lRf:-Nlifaze3mNy3Lgu6lRg:b4p0rta) ([raw file](https://github.com/cockroachdb/pebble/blob/44c7fcbc8d4d607e1b3f962f920078c0014a42f4/sstable/reader_iter_single_lvl.go#L464)):*
> ```Go
> 	// keys are ≤ index.Key(). For forward iteration, this is all we
> 	// need.
> 	if dir >= 0 {
> ```

why the `>=`? 
___
*[`sstable/reader_iter_single_lvl.go` line 860 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3147#-NlihPq-9h4qPbDr7HJ0:-NlihPq-9h4qPbDr7HJ1:bd2iuau) ([raw file](https://github.com/cockroachdb/pebble/blob/44c7fcbc8d4d607e1b3f962f920078c0014a42f4/sstable/reader_iter_single_lvl.go#L860)):*
> ```Go
> 	// Seek to the first internal key.
> 	ikey, _ := i.virtualLastSeekGE(i.upper)
> 	if i.endKeyInclusive {
> ```

if `i.endKeyInclusive`, couldn't we just `SeekLT(i.upper)` instead?
___
*[`sstable/reader_iter_single_lvl.go` line 887 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3147#-NligJODBP0dhtgSxzdI:-NligJODBP0dhtgSxzdJ:bo15j6i) ([raw file](https://github.com/cockroachdb/pebble/blob/44c7fcbc8d4d607e1b3f962f920078c0014a42f4/sstable/reader_iter_single_lvl.go#L887)):*
> ```Go
> // virtualLastSeekGE is called by virtualLast to do a Next() as part of a Last() or
> // SeekLT().
> func (i *singleLevelIterator) virtualLastSeekGE(key []byte) (*InternalKey, base.LazyValue) {
> ```

rather than two special methods for this case, could we add just a single `seekLE` method to the sstable iterators? The `dir` parameter is very confusing when the methods are already supposed to be directional.

RaduBerinde (commented):
The code LGTM to me but I'm a bit out of my depth, I will let @jbowens approve when he's satisfied.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3147#-:-Nlj8-VV-nT5dCNp980t:baro48y)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`sstable/reader_iter_single_lvl.go` line 464 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3147#-Nlifaze3mNy3Lgu6lRf:-Nlj7KWu7kTIeqELCBm5:b30y1eg) ([raw file](https://github.com/cockroachdb/pebble/blob/44c7fcbc8d4d607e1b3f962f920078c0014a42f4/sstable/reader_iter_single_lvl.go#L464)):*

why the `>=`?
</blockquote></details>

0 means "indefinite direction" so we have to check both bounds. We should mention this in the function comment.
___
*[`sstable/reader_iter_single_lvl.go` line 489 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3147#-Nlj7WZO4HpAJTH_ZXb1:-Nlj7WZP8OlmfqbTYurZ:bsz7ejk) ([raw file](https://github.com/cockroachdb/pebble/blob/44c7fcbc8d4d607e1b3f962f920078c0014a42f4/sstable/reader_iter_single_lvl.go#L489)):*
> ```Go
> 		// the original block's keys. Afterwards, we step forward to restore our
> 		// index position.
> 		if peekKey, _ := i.index.Prev(); peekKey == nil {
> ```

[nit]
```
peekKey, _ := i.index.Prev()
_, _ = i.index.Next()
```
and no need for the `else`

itsbilal (commented):
TFTRs! I implemented a `virtualLastSeekLE` in both the single/two Level iterators so this is now a very different (but also more encapsulated to just VSSTs) change.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3147#-:-NljwsFZ26ioXjjN5mvo:b1xe341)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 464 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3147#-Nlifaze3mNy3Lgu6lRf:-Nlj-4-zC5VHD2IhgxHe:br0pikm) ([raw file](https://github.com/cockroachdb/pebble/blob/44c7fcbc8d4d607e1b3f962f920078c0014a42f4/sstable/reader_iter_single_lvl.go#L464)):*

0 means "indefinite direction" so we have to check both bounds. We should mention this in the function comment.
</blockquote></details>

It allows the == 0 case to check both bounds in a sense. Anyway it's gone now; we never use 0 for dir anymore.
___
*[`sstable/reader_iter_single_lvl.go` line 489 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3147#-Nlj7WZO4HpAJTH_ZXb1:-NljwoxnAELO1vwRi055:b-td19we) ([raw file](https://github.com/cockroachdb/pebble/blob/44c7fcbc8d4d607e1b3f962f920078c0014a42f4/sstable/reader_iter_single_lvl.go#L489)):*

[nit]
```
peekKey, _ := i.index.Prev()
_, _ = i.index.Next()
```
and no need for the `else`
</blockquote></details>

Undid changes to this method so N/A
___
*[`sstable/reader_iter_single_lvl.go` line 860 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3147#-NlihPq-9h4qPbDr7HJ0:-Nlj-Cci4gaYlX1CpDHj:bjg6x97) ([raw file](https://github.com/cockroachdb/pebble/blob/44c7fcbc8d4d607e1b3f962f920078c0014a42f4/sstable/reader_iter_single_lvl.go#L860)):*

if `i.endKeyInclusive`, couldn't we just `SeekLT(i.upper)` instead?
</blockquote></details>

Yep. Done.
___
*[`sstable/reader_iter_single_lvl.go` line 887 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3147#-NligJODBP0dhtgSxzdI:-NljwrUDEq0llD9mH53z:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/44c7fcbc8d4d607e1b3f962f920078c0014a42f4/sstable/reader_iter_single_lvl.go#L887)):*

rather than two special methods for this case, could we add just a single `seekLE` method to the sstable iterators? The `dir` parameter is very confusing when the methods are already supposed to be directional.
</blockquote></details>

Done.

RaduBerinde (approved):
:lgtm:

Nice, this looks much better to me!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3147#-:-NlkY0HYD9QzF1hD6Mvu:bdgpch9)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`sstable/reader_iter_single_lvl.go` line 860 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3147#-NlkWAFfApImBiP7xKQ-:-NlkWAFfApImBiP7xKQ0:bnquwgg) ([raw file](https://github.com/cockroachdb/pebble/blob/a4fbc5bcfe751053c0d1dfca4264e12fb508dd9d/sstable/reader_iter_single_lvl.go#L860)):*
> ```Go
> 		return i.SeekLT(i.upper, base.SeekLTFlagsNone)
> 	}
> 	// Seek to the first internal key.
> ```

[nit] comment needs updating
___
*[`sstable/reader_iter_single_lvl.go` line 892 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3147#-NlkXPkwDTqlcJUoTSH7:-NlkXPkwDTqlcJUoTSH8:b-ena9u2) ([raw file](https://github.com/cockroachdb/pebble/blob/a4fbc5bcfe751053c0d1dfca4264e12fb508dd9d/sstable/reader_iter_single_lvl.go#L892)):*
> ```Go
> 	// This might be challenging to do correctly, so impose regular operations
> 	// for now.
> 	for ikey != nil && bytes.Equal(ikey.UserKey, key) {
> ```

[nit] It took me a bit to understand what we're doing here; we should mention that we can have multiple keys with the same UserKey and we want the last one.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3147#-:-NlxZ3YgABRnIlwyQBTd:b-814g8k)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`sstable/reader_iter_single_lvl.go` line 860 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3147#-NlkWAFfApImBiP7xKQ-:-NlxZ4ar4zqWApb7Txb6:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/a4fbc5bcfe751053c0d1dfca4264e12fb508dd9d/sstable/reader_iter_single_lvl.go#L860)):*

[nit] comment needs updating
</blockquote></details>

Done.
___
*[`sstable/reader_iter_single_lvl.go` line 892 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3147#-NlkXPkwDTqlcJUoTSH7:-NlxZ5KiCT_xTUmTrbop:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/a4fbc5bcfe751053c0d1dfca4264e12fb508dd9d/sstable/reader_iter_single_lvl.go#L892)):*

[nit] It took me a bit to understand what we're doing here; we should mention that we can have multiple keys with the same UserKey and we want the last one.
</blockquote></details>

Done.

itsbilal:
Jackson is out so I'll merge this for now. Can address anything that comes up after the fact.

PR #3148, opened 2023-12-11
metamorphic: increase probability of non-move multilevel compactions

jbowens:
Adapt the multilevel compaction heuristic into a metamorphic option, allowing
random option runs to use no multilevel compactions, multilevel compactions
with the default write amplification heuristic, and multilevel compactions with
additional, artificially inflated propensity.

Previously, the metamorphic tests tended to exercise simple multilevel
compactions, and especially multilevel move compactions. Now some random runs
should schedule a higher portion of nontrivial multilevel compactions.


itsbilal (approved):
LGTM

jbowens:
TFTR!

PR #3154, opened 2023-12-13
metamorphic: track object key bounds

jbowens:
This commit adjusts operation generation to track the smallest and largest keys in all objects. This is used to determine at generation time whether an ingestion should succeed or fail. Only if the ingestion will succeed is key manager state updated as if the operation succeeded.

This increases the probability of multi-batch ingestions and ingestions involving deletion operations through removing the old canTolerateApplyFailure mechanic that avoided ingesting some batches as a multi-batch ingestion if they contained any delete operations.

### DIFF
--- a/metamorphic/generator.go
+++ b/metamorphic/generator.go
@@ -1427,57 +1427,41 @@ func (g *generator) writerIngest() {
 		return
 	}
 
-	// TODO(nicktrav): this is resulting in too many single batch ingests.
-	// Consider alternatives. One possibility would be to pass through whether
-	// we can tolerate failure or not, and if the ingestOp encounters a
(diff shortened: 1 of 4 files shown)


itsbilal (approved):
LGTM

jbowens:
I'm realizing this will probably break the crossversion metamorphic test because the `successful` parameter to the `Ingest` operation won't exist. I'll need to think about what to do there.

jbowens:
I removed the `successful` bool parameter in favor of recomputing whether any batches overlap in `update`. This removes the backwards-incompatible ingest op signature change, and will ensure we compute whether the ingest will succeed correctly when we parse previous versions' ops files.

jbowens:
TFTR!

PR #3156, opened 2023-12-13
crl-release-23.1: Makefile: update crossversion-meta to test 22.2 -> 23.1 only

jbowens:
Previously the `crossversion-meta` target would test 23.2 -> 23.1 (moving back releases) because it always used the latest release. Update the Makefile to always test 22.2 -> 23.1 on the 23.1 release branch.

PR #3157, opened 2023-12-13
crl-release-23.1: db: use uncompensated scores to prioritize levels for compaction

jbowens:
23.1 backport of #2917. The backport was not clean because we refactored several pieces of the compaction picker over the course of 23.2, but the overall structure of the patch is the same.

----

During large table drops, we see level compensation become drastically high, which leads to L0 compaction starvation, which in turn leads to AC queueing of requests and increases in foreground latencies.

We completely ignore compensation during prioritization of levels during compaction, but we consider compensation when we try and pick a level for compaction.

The above approach ensures that L0 isn't starved during large table drops.

Note that there might be scenarios where a level say L3, has a high compensated score due to a wide range delete, and L3 will also drop a ton of data in L4, but we end up picking L4 for compaction because it has a higher raw score.

While this will make some of the compactions picked inefficient, most of the data which is dropped due to the range delete should be in L6 anyway, and the L5 -> L6 compaction which will contain this range delete will still be cheap.

We considered some other heuristics in

Other approaches considered include capping the level compensation to a constant, or accounting for compensation in higher levels, but not lower levels. The first approach might not work well, because the level compensation can be lower than the constant picked and still starve out L0. The second approach is more difficult to reason about compared to the current approach.

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -77,18 +77,38 @@ func (info compactionInfo) String() string {
 	return buf.String()
 }
 
-type sortCompactionLevelsDecreasingScore []candidateLevelInfo
+type sortCompactionLevelsByPriority []candidateLevelInfo
 
-func (s sortCompactionLevelsDecreasingScore) Len() int {
+func (s sortCompactionLevelsByPriority) Len() int {
 	return len(s)
 }
-func (s sortCompactionLevelsDecreasingScore) Less(i, j int) bool {
-	if s[i].score != s[j].score {
-		return s[i].score > s[j].score
+
+// A level should be picked for compaction if the compensatedScoreRatio is >= the
+// compactionScoreThreshold.
+const compactionScoreThreshold = 1
+
+// Less should return true if s[i] must be placed earlier than s[j] in the final
+// sorted list. The candidateLevelInfo for the level placed earlier is more likely
+// to be picked for a compaction.
+func (s sortCompactionLevelsByPriority) Less(i, j int) bool {
+	iShouldCompact := s[i].compensatedScoreRatio >= compactionScoreThreshold
+	jShouldCompact := s[j].compensatedScoreRatio >= compactionScoreThreshold
+	// Ordering is defined as decreasing on (shouldCompact, uncompensatedScoreRatio)
+	// where shouldCompact is 1 for true and 0 for false.
+	if iShouldCompact && !jShouldCompact {
+		return true
+	}
+	if !iShouldCompact && jShouldCompact {
+		return false
+	}
+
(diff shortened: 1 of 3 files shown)


jbowens:
The `go-linux-crossversion` test failure is an issue with the test itself and will be fixed by #3156.

sumeerbhola (approved):
:lgtm: 

the comment doesn't affect this backport

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3157#-:-NldjgQ57iECvZgzWglT:betb7z0)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`compaction_picker.go` line 1081 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3157#-NldihXBBz85HfYwffOA:-NldihXBBz85HfYwffOB:b-xbpvv9) ([raw file](https://github.com/cockroachdb/pebble/blob/b06c3ee42a387f21aadfab346ca51ef328b7f4d2/compaction_picker.go#L1081)):*
> ```Go
> 			}
> 		}
> 		if scores[prevLevel].uncompensatedScoreRatio >= compactionScoreThreshold {
> ```

I suppose this is the same as master, but it seems a curious choice: we will use compensatedScoreRatio >= 1.0 as the threshold for whether to compact, but if `uncompensatedScore` is < 1, we won't divide it by `scores[level].uncompensatedScore`, so it will be a non-ratio score that will then be used for ranking.
Seems better to do this logic whenever the previous if-block is done.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3157#-:-NliP4reEmumBKumsP6u:b-814g8k)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal and @sumeerbhola)

___
*[`compaction_picker.go` line 1081 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3157#-NldihXBBz85HfYwffOA:-NliOyKj4TCFACxYlxiz:b-9cnjxi) ([raw file](https://github.com/cockroachdb/pebble/blob/b06c3ee42a387f21aadfab346ca51ef328b7f4d2/compaction_picker.go#L1081)):*

I suppose this is the same as master, but it seems a curious choice: we will use compensatedScoreRatio >= 1.0 as the threshold for whether to compact, but if `uncompensatedScore` is < 1, we won't divide it by `scores[level].uncompensatedScore`, so it will be a non-ratio score that will then be used for ranking.
Seems better to do this logic whenever the previous if-block is done.
</blockquote></details>

Yeah, that is curious. I'll come back to this on master.

PR #3158, opened 2023-12-13
crl-release-23.2: Makefile: update crossversion-meta to test 23.1 -> 23.2 only

jbowens:
Previously the `crossversion-meta` target would test 23.2 -> 23.2 (moving laterally) because it always used the latest release. Update the Makefile to always test 23.1 -> 23.2 on the 23.2 release branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ GOFLAGS :=
 STRESSFLAGS :=
 TAGS := invariants
 TESTS := .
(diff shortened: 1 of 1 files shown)


jbowens:
Friendly ping :)

RaduBerinde (approved):

RaduBerinde:
Should we do the same on `master`? At least have the 23.2 variant commented out so it's easy to do again next time we branch. Is there a check-list that we go through when creating a new `crl-release-` branch?

Issue #2632, opened 2023-06-13
sstable: table-wide prefix compression


jbowens:
Consider a new `Comparer` function `EqualBytePrefix(a, b []byte) (prefix []byte, ok bool)`. This function returns a shared byte prefix _p_ of both `a` and `b` such that for all _x_ and _y_:

```
Compare(p + x, p + y) = Compare(x, y) 
```

In larger databases with larger SQL tables, it's common for all of a compaction's inputs to fall within the same SQL table index. If there are very many rows with a shared prefix, it's reasonably likely that the current key and the largest key of an overlapping compaction input all have a relatively large shared byte prefix, in particular if the user is storing large keys. When starting a new output sstable, we could call `EqualBytePrefix(key, c.Largest.UserKey)` to determine what shared byte prefix all the remaining keys have, and for which key comparisons could rely solely on the remaining byte suffix.

When writing these sstables, we could stash the `EqualBytePrefix` somewhere (a property?) and proceed to write only the byte suffix to sstables. During iteration, when decoding keys solely to serve seeks (eg, see the special blockIter.SeekGE code path and its inlined entry decoding), we could perform comparisons solely on the key suffixes encoded in the block. When decoding keys to return up the iterator stack, the blockIter would use its `fullKey` buffer used for prefix compression to append the block-encoded byte suffix onto the table-wide byte prefix.

This has a few advantages over the block-level prefix compression:
1. Block-level prefix compression does not apply to index blocks, because we need to be able to seek among all of the keys within an index block, requiring restartInterval=1.
2. Block-level prefix compression duplicates the shared key prefix every `restartInterval` (16 keys in CockroachDB).
3. Key comparisons during seeks are performed on the fully materialized key which adds cost of materializing the key, plus a more expensive key comparison.

I'm unsure in practice how large a shared byte prefix would need to be for the overhead to be worth it, or how frequently we see such keys and sstables in CockroachDB. I think we should examine telemetry, large tpce workloads, etc to estimate whether this is useful in practice.

Jira issue: PEBBLE-59

jbowens:
@RaduBerinde @dt @sumeerbhola This is an older issue that describes a mechanic that could be repurposed for prefix rewriting.

dt:
We could also do something like this per-block, i.e. just include the shared prefix once in the block footer and rather than in every restart key. 

[...]

PR #3159, opened 2023-12-14
sstable: add NewPrefixReplacingIterator

dt:
These changes allow ingesting an sstable that contains keys that all have some prefix `old` as if it actually had been built with all of those keys with prefix `new`. This is done by changing "new" to "old" in the arguments to interactions with this sstable, and then changing "old" to "new" in the results of those interactions. 

There are plenty of potentially ways to implement similar behavior that might be more performant, however the goal of this initial implementation is more to demonstrate the functionality, and allow tests to start to be put in place to verify correctness, after which the performance can be optimized as needed (for example by pushing the replacement deeper into table reader, perhaps on specially-built tables that elide their common prefix, etc).

### DIFF
--- a/sstable/reader_iter_single_lvl.go
+++ b/sstable/reader_iter_single_lvl.go
@@ -321,6 +321,23 @@ func disableBoundsOpt(bound []byte, ptr uintptr) bool {
 // state and require this behavior to be deterministic.
 var ensureBoundsOptDeterminism bool
 
+// SetBoundsWithSyntheticPrefix indicates whether this iterator requires keys
+// passed to its SetBounds() method by a prefix rewriting wrapper to be *not*
+// rewritten to be in terms of this iterator's content, but instead be passed
+// as-is, i.e. with the synthetic prefix still on them.
+//
+// This allows an optimization when this iterator is passing these bounds on to
+// a vState to additionally constrain them. In said vState, passed bounds are
+// combined with the vState bounds which are in terms of the rewritten prefix.
+// If the caller rewrote bounds to be in terms of content prefix and SetBounds
+// passed those to vState, the vState would need to *un*rewrite them back to the
+// synthetic prefix in order to combine them with the vState bounds. Thus, if
+// this iterator knows bounds will be passed to vState, it can signal that it
+// they should be passed without being rewritten to skip converting to and fro.
+func (i singleLevelIterator) SetBoundsWithSyntheticPrefix() bool {
+	return i.vState != nil
+}
+
 // SetBounds implements internalIterator.SetBounds, as documented in the pebble
 // package. Note that the upper field is exclusive.
 func (i *singleLevelIterator) SetBounds(lower, upper []byte) {
--- a/sstable/reader_virtual.go
+++ b/sstable/reader_virtual.go
@@ -5,6 +5,7 @@
 package sstable
 
 import (
+	"bytes"
 	"context"
 
 	"github.com/cockroachdb/pebble/internal/base"
@@ -27,11 +28,12 @@ type VirtualReader struct {
 
 // Lightweight virtual sstable state which can be passed to sstable iterators.
 type virtualState struct {
-	lower     InternalKey
-	upper     InternalKey
-	fileNum   base.FileNum
-	Compare   Compare
-	isForeign bool
+	lower        InternalKey
+	upper        InternalKey
+	fileNum      base.FileNum
+	Compare      Compare
+	isForeign    bool
+	prefixChange *manifest.PrefixReplacement
 }
 
 func ceilDiv(a, b uint64) uint64 {
@@ -48,11 +50,12 @@ func MakeVirtualReader(
 	}
 
 	vState := virtualState{
-		lower:     meta.Smallest,
-		upper:     meta.Largest,
-		fileNum:   meta.FileNum,
-		Compare:   reader.Compare,
-		isForeign: isForeign,
+		lower:        meta.Smallest,
+		upper:        meta.Largest,
+		fileNum:      meta.FileNum,
+		Compare:      reader.Compare,
+		isForeign:    isForeign,
+		prefixChange: meta.PrefixReplacement,
 	}
 	v := VirtualReader{
 		vState: vState,
@@ -74,6 +77,7 @@ func MakeVirtualReader(
 	v.Properties.NumSizedDeletions = ceilDiv(reader.Properties.NumSizedDeletions*meta.Size, meta.FileBacking.Size)
 	v.Properties.RawPointTombstoneKeySize = ceilDiv(reader.Properties.RawPointTombstoneKeySize*meta.Size, meta.FileBacking.Size)
 	v.Properties.RawPointTombstoneValueSize = ceilDiv(reader.Properties.RawPointTombstoneValueSize*meta.Size, meta.FileBacking.Size)
+
 	return v
 }
 
@@ -85,8 +89,12 @@ func (v *VirtualReader) NewCompactionIter(
 	rp ReaderProvider,
 	bufferPool *BufferPool,
 ) (Iterator, error) {
-	return v.reader.newCompactionIter(
+	i, err := v.reader.newCompactionIter(
 		bytesIterated, categoryAndQoS, statsCollector, rp, &v.vState, bufferPool)
+	if err == nil && v.vState.prefixChange != nil {
+		i = newPrefixReplacingIterator(i, v.vState.prefixChange.ContentPrefix, v.vState.prefixChange.SyntheticPrefix, v.reader.Compare)
+	}
+	return i, err
 }
 
 // NewIterWithBlockPropertyFiltersAndContextEtc wraps
@@ -103,9 +111,13 @@ func (v *VirtualReader) NewIterWithBlockPropertyFiltersAndContextEtc(
 	statsCollector *CategoryStatsCollector,
 	rp ReaderProvider,
 ) (Iterator, error) {
-	return v.reader.newIterWithBlockPropertyFiltersAndContext(
+	i, err := v.reader.newIterWithBlockPropertyFiltersAndContext(
 		ctx, lower, upper, filterer, hideObsoletePoints, useFilterBlock, stats,
 		categoryAndQoS, statsCollector, rp, &v.vState)
+	if err == nil && v.vState.prefixChange != nil {
+		i = newPrefixReplacingIterator(i, v.vState.prefixChange.ContentPrefix, v.vState.prefixChange.SyntheticPrefix, v.reader.Compare)
+	}
+	return i, err
 }
 
 // ValidateBlockChecksumsOnBacking will call ValidateBlockChecksumsOnBacking on the underlying reader.
@@ -123,6 +135,19 @@ func (v *VirtualReader) NewRawRangeDelIter() (keyspan.FragmentIterator, error) {
 	if iter == nil {
 		return nil, nil
 	}
+	lower := &v.vState.lower
+	upper := &v.vState.upper
+
+	if v.vState.prefixChange != nil {
+		lower = &InternalKey{UserKey: v.vState.prefixChange.ReplaceArg(lower.UserKey), Trailer: lower.Trailer}
+		upper = &InternalKey{UserKey: v.vState.prefixChange.ReplaceArg(upper.UserKey), Trailer: upper.Trailer}
+
+		iter = keyspan.Truncate(
+			v.reader.Compare, iter, lower.UserKey, upper.UserKey,
+			lower, upper, !v.vState.upper.IsExclusiveSentinel(), /* panicOnUpperTruncate */
+		)
+		return newPrefixReplacingFragmentIterator(iter, v.vState.prefixChange.ContentPrefix, v.vState.prefixChange.SyntheticPrefix), nil
+	}
 
 	// Truncation of spans isn't allowed at a user key that also contains points
 	// in the same virtual sstable, as it would lead to covered points getting
@@ -135,8 +160,8 @@ func (v *VirtualReader) NewRawRangeDelIter() (keyspan.FragmentIterator, error) {
 	// includes both point keys), but not [a#2,SET-b#3,SET] (as it would truncate
 	// the rangedel at b and lead to the point being uncovered).
 	return keyspan.Truncate(
-		v.reader.Compare, iter, v.vState.lower.UserKey, v.vState.upper.UserKey,
-		&v.vState.lower, &v.vState.upper, !v.vState.upper.IsExclusiveSentinel(), /* panicOnUpperTruncate */
+		v.reader.Compare, iter, lower.UserKey, upper.UserKey,
+		lower, upper, !v.vState.upper.IsExclusiveSentinel(), /* panicOnUpperTruncate */
 	), nil
 }
 
@@ -148,6 +173,18 @@ func (v *VirtualReader) NewRawRangeKeyIter() (keyspan.FragmentIterator, error) {
 	}
 	if iter == nil {
 		return nil, nil
+	}
+	lower := &v.vState.lower
+	upper := &v.vState.upper
+
+	if v.vState.prefixChange != nil {
+		lower = &InternalKey{UserKey: v.vState.prefixChange.ReplaceArg(lower.UserKey), Trailer: lower.Trailer}
+		upper = &InternalKey{UserKey: v.vState.prefixChange.ReplaceArg(upper.UserKey), Trailer: upper.Trailer}
+		iter = keyspan.Truncate(
+			v.reader.Compare, iter, lower.UserKey, upper.UserKey,
+			lower, upper, !v.vState.upper.IsExclusiveSentinel(), /* panicOnUpperTruncate */
+		)
+		return newPrefixReplacingFragmentIterator(iter, v.vState.prefixChange.ContentPrefix, v.vState.prefixChange.SyntheticPrefix), nil
 	}
 
 	// Truncation of spans isn't allowed at a user key that also contains points
@@ -161,8 +198,8 @@ func (v *VirtualReader) NewRawRangeKeyIter() (keyspan.FragmentIterator, error) {
 	// includes both point keys), but not [a#2,SET-b#3,SET] (as it would truncate
 	// the range key at b and lead to the point being uncovered).
 	return keyspan.Truncate(
-		v.reader.Compare, iter, v.vState.lower.UserKey, v.vState.upper.UserKey,
-		&v.vState.lower, &v.vState.upper, !v.vState.upper.IsExclusiveSentinel(), /* panicOnUpperTruncate */
+		v.reader.Compare, iter, lower.UserKey, upper.UserKey,
+		lower, upper, !v.vState.upper.IsExclusiveSentinel(), /* panicOnUpperTruncate */
 	), nil
 }
 
@@ -195,6 +232,10 @@ func (v *virtualState) constrainBounds(
 			last = end
 		}
 	}
+	if v.prefixChange != nil {
+		first = v.prefixChange.ReplaceArg(first)
+		last = v.prefixChange.ReplaceArg(last)
+	}
 	// TODO(bananabrick): What if someone passes in bounds completely outside of
 	// virtual sstable bounds?
 	return lastKeyInclusive, first, last
@@ -204,6 +245,16 @@ func (v *virtualState) constrainBounds(
 // enforcing the virtual sstable bounds.
 func (v *VirtualReader) EstimateDiskUsage(start, end []byte) (uint64, error) {
 	_, f, l := v.vState.constrainBounds(start, end, true /* endInclusive */)
+	if v.vState.prefixChange != nil {
+		if !bytes.HasPrefix(f, v.vState.prefixChange.SyntheticPrefix) || !bytes.HasPrefix(l, v.vState.prefixChange.SyntheticPrefix) {
+			return 0, errInputPrefixMismatch
+		}
+		// TODO(dt): we could add a scratch buf to VirtualReader to avoid allocs on
+		// repeated calls to this.
+		f = append(append([]byte{}, v.vState.prefixChange.ContentPrefix...), f[len(v.vState.prefixChange.SyntheticPrefix):]...)
+		l = append(append([]byte{}, v.vState.prefixChange.ContentPrefix...), l[len(v.vState.prefixChange.SyntheticPrefix):]...)
+	}
+
 	return v.reader.EstimateDiskUsage(f, l)
 }
 
--- a/data_test.go
+++ b/data_test.go
@@ -1271,6 +1271,20 @@ func runIngestCmd(td *datadriven.TestData, d *DB, fs vfs.FS) error {
 
 func runIngestExternalCmd(td *datadriven.TestData, d *DB, locator string) error {
 	external := make([]ExternalFile, 0)
+	usePrefixChange := false
+	var fromPrefix, toPrefix []byte
+	for i := range td.CmdArgs {
+		switch td.CmdArgs[i].Key {
+		case "prefix-replace":
+			vals := td.CmdArgs[i].Vals
+			if len(vals) != 2 {
+				return errors.New("usage: prefix-replace=(from,to)")
+			}
+			fromPrefix = []byte(vals[0])
+			toPrefix = []byte(vals[1])
+			usePrefixChange = true
+		}
+	}
 	for _, arg := range strings.Split(td.Input, "\n") {
 		fields := strings.Split(arg, ",")
 		if len(fields) != 4 {
@@ -1287,6 +1301,10 @@ func runIngestExternalCmd(td *datadriven.TestData, d *DB, locator string) error 
 		ef.SmallestUserKey = []byte(fields[2])
 		ef.LargestUserKey = []byte(fields[3])
 		ef.HasPointKey = true
+		if usePrefixChange {
+			ef.ContentPrefix = fromPrefix
+			ef.SyntheticPrefix = toPrefix
+		}
 		external = append(external, ef)
 	}
 
--- a/format_major_version.go
+++ b/format_major_version.go
@@ -178,6 +178,13 @@ const (
 	// a format major version.
 	FormatVirtualSSTables
 
+	// FormatSyntheticPrefixes is a format major version that adds support for
+	// sstables to have their content exposed in a different prefix of keyspace
+	// than the actual prefix persisted in the keys in such sstables. The prefix
+	// replacement information is stored in new fields in the Manifest and thus
+	// requires a format major version.
+	FormatSyntheticPrefixes
+
 	// -- Add new versions here --
 
 	// FormatNewest is the most recent format major version.
@@ -209,7 +216,7 @@ func (v FormatMajorVersion) MaxTableFormat() sstable.TableFormat {
 	switch v {
 	case FormatDefault, FormatFlushableIngest, FormatPrePebblev1MarkedCompacted:
 		return sstable.TableFormatPebblev3
-	case FormatDeleteSizedAndObsolete, FormatVirtualSSTables:
+	case FormatDeleteSizedAndObsolete, FormatVirtualSSTables, FormatSyntheticPrefixes:
 		return sstable.TableFormatPebblev4
 	default:
 		panic(fmt.Sprintf("pebble: unsupported format major version: %s", v))
@@ -221,7 +228,7 @@ func (v FormatMajorVersion) MaxTableFormat() sstable.TableFormat {
 func (v FormatMajorVersion) MinTableFormat() sstable.TableFormat {
 	switch v {
 	case FormatDefault, FormatFlushableIngest, FormatPrePebblev1MarkedCompacted,
-		FormatDeleteSizedAndObsolete, FormatVirtualSSTables:
+		FormatDeleteSizedAndObsolete, FormatVirtualSSTables, FormatSyntheticPrefixes:
 		return sstable.TableFormatPebblev1
 	default:
 		panic(fmt.Sprintf("pebble: unsupported format major version: %s", v))
@@ -254,6 +261,9 @@ var formatMajorVersionMigrations = map[FormatMajorVersion]func(*DB) error{
 	},
 	FormatVirtualSSTables: func(d *DB) error {
 		return d.finalizeFormatVersUpgrade(FormatVirtualSSTables)
+	},
+	FormatSyntheticPrefixes: func(d *DB) error {
+		return d.finalizeFormatVersUpgrade(FormatSyntheticPrefixes)
 	},
 }
 
--- a/format_major_version_test.go
+++ b/format_major_version_test.go
@@ -23,11 +23,12 @@ func TestFormatMajorVersionStableValues(t *testing.T) {
 	require.Equal(t, FormatPrePebblev1MarkedCompacted, FormatMajorVersion(14))
 	require.Equal(t, FormatDeleteSizedAndObsolete, FormatMajorVersion(15))
 	require.Equal(t, FormatVirtualSSTables, FormatMajorVersion(16))
+	require.Equal(t, FormatSyntheticPrefixes, FormatMajorVersion(17))
 
 	// When we add a new version, we should add a check for the new version in
 	// addition to updating these expected values.
-	require.Equal(t, FormatNewest, FormatMajorVersion(16))
-	require.Equal(t, internalFormatNewest, FormatMajorVersion(16))
+	require.Equal(t, FormatNewest, FormatMajorVersion(17))
+	require.Equal(t, internalFormatNewest, FormatMajorVersion(17))
 }
 
 func TestFormatMajorVersion_MigrationDefined(t *testing.T) {
@@ -50,6 +51,8 @@ func TestRatchetFormat(t *testing.T) {
 	require.Equal(t, FormatDeleteSizedAndObsolete, d.FormatMajorVersion())
 	require.NoError(t, d.RatchetFormatMajorVersion(FormatVirtualSSTables))
 	require.Equal(t, FormatVirtualSSTables, d.FormatMajorVersion())
+	require.NoError(t, d.RatchetFormatMajorVersion(FormatSyntheticPrefixes))
+	require.Equal(t, FormatSyntheticPrefixes, d.FormatMajorVersion())
 
 	require.NoError(t, d.Close())
 
@@ -203,6 +206,7 @@ func TestFormatMajorVersions_TableFormat(t *testing.T) {
 		FormatPrePebblev1MarkedCompacted: {sstable.TableFormatPebblev1, sstable.TableFormatPebblev3},
 		FormatDeleteSizedAndObsolete:     {sstable.TableFormatPebblev1, sstable.TableFormatPebblev4},
 		FormatVirtualSSTables:            {sstable.TableFormatPebblev1, sstable.TableFormatPebblev4},
+		FormatSyntheticPrefixes:          {sstable.TableFormatPebblev1, sstable.TableFormatPebblev4},
 	}
 
 	// Valid versions.
--- a/ingest.go
+++ b/ingest.go
@@ -214,6 +214,13 @@ func ingestLoad1External(
 	// what parts of this sstable are referenced by other nodes.
 	meta.FileBacking.Size = e.Size
 
+	if len(e.SyntheticPrefix) != 0 {
+		meta.PrefixReplacement = &manifest.PrefixReplacement{
+			ContentPrefix:   e.ContentPrefix,
+			SyntheticPrefix: e.SyntheticPrefix,
+		}
+	}
+
 	if err := meta.Validate(opts.Comparer.Compare, opts.Comparer.FormatKey); err != nil {
 		return nil, err
 	}
@@ -1112,6 +1119,11 @@ type ExternalFile struct {
 	// or range keys. If both structs are false, an error is returned during
 	// ingestion.
 	HasPointKey, HasRangeKey bool
+	// ContentPrefix and SyntheticPrefix denote a prefix replacement rule causing
+	// a file, in which all keys have prefix ContentPrefix, to appear whenever it
+	// is accessed as if those keys all instead have prefix SyntheticPrefix.
+	// SyntheticPrefix must be a prefix of both SmallestUserKey and LargestUserKey.
+	ContentPrefix, SyntheticPrefix []byte
 }
 
 // IngestWithStats does the same as Ingest, and additionally returns
--- a/ingest.go
+++ b/ingest.go
@@ -1308,6 +1308,13 @@ func (d *DB) ingest(
 	if (exciseSpan.Valid() || len(shared) > 0 || len(external) > 0) && d.FormatMajorVersion() < FormatVirtualSSTables {
 		return IngestOperationStats{}, errors.New("pebble: format major version too old for excise, shared or external sstable ingestion")
 	}
+	if len(external) > 0 && d.FormatMajorVersion() < FormatSyntheticPrefixes {
+		for i := range external {
+			if len(external[i].SyntheticPrefix) > 0 {
+				return IngestOperationStats{}, errors.New("pebble: format major version too old for synthetic prefix ingestion")
+			}
+		}
+	}
 	// Allocate file numbers for all of the files being ingested and mark them as
 	// pending in order to prevent them from being deleted. Note that this causes
 	// the file number ordering to be out of alignment with sequence number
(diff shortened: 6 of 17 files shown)


sstable/reader_virtual_test.go
```diff
@@ -0,0 +1,142 @@
...
+		{from: []byte("zzz"), to: []byte("aa")},
+	} {
+		t.Run(fmt.Sprintf("%s_%s", tc.from, tc.to), func(t *testing.T) {
+			r := buildTestTable(t, 20, 256, 256, DefaultCompression, tc.from)
+			defer r.Close()
+			rawIter, err := r.NewIter(nil, nil)
+			require.NoError(t, err)
+			defer rawIter.Close()
+
+			raw := rawIter.(*singleLevelIterator)
```
itsbilal, line 31: do we need it to be a singleLevelIterator? I don't see why. Might be a good idea to have a version of this test that calls `buildTestTable` with tiny index/data block sizes so we get a twoLevelIterator here too.

itsbilal (approved):
LGTM

sstable/reader_virtual_test.go
dt: we don't; the cast is so I can assert the internal fields are indeed being set by Set*.

dt: > do we need it to be a singleLevelIterator

No, I'm just casting it so that I can verify Set* functions set the underlying fields since otherwise their effects are unobservable.

sstable/reader_virtual.go
```diff
@@ -204,10 +248,264 @@ func (v *virtualState) constrainBounds(
...
+func (p *prefixReplacingFragmentIterator) rewriteSpan(sp *keyspan.Span) *keyspan.Span {
+	if !bytes.HasPrefix(sp.Start, p.src) || !bytes.HasPrefix(sp.End, p.src) {
+		p.err = errInputPrefixMismatch
+		return sp
+	}
+	sp.Start = append(p.out1[:len(p.dst)], sp.Start[len(p.src):]...)
+	sp.End = append(p.out2[:len(p.dst)], sp.End[len(p.src):]...)
+
+	// TODO(dt): RESOLVE DURING CODE REVIEW
+	// do I need to touch sp.Keys? is sp.Start/End actually assured to both have dst prefix?
```
itsbilal, line 466: You don't need to touch sp.Keys, the bounds are only in sp.Start/End.

sstable/reader_virtual.go
```diff
@@ -204,10 +248,264 @@ func (v *virtualState) constrainBounds(
...
+}
+
+// Close implements the Iterator interface.
+func (p *prefixReplacingIterator) Close() error {
+	return p.i.Close()
+}
+
+// SetBounds implements the Iterator interface.
+func (p *prefixReplacingIterator) SetBounds(lower, upper []byte) {
+	p.i.SetBounds(p.rewriteArg(lower), p.rewriteArg2(upper))
```
itsbilal, line 409: This is redundant / slightly buggy as the underlying iterator will do the same rewrite as part of `virtualState.constrainBounds`.

sstable/reader_virtual.go
```diff
@@ -204,10 +248,264 @@ func (v *virtualState) constrainBounds(
...
+	}
+	return p.rewriteResult(p.i.SeekGE(p.rewriteArg(key), flags))
+}
+
+// SeekPrefixGE implements the Iterator interface.
+func (p *prefixReplacingIterator) SeekPrefixGE(
+	prefix, key []byte, flags base.SeekGEFlags,
+) (*InternalKey, base.LazyValue) {
+	cmp := p.cmp(key, p.dst)
+	if cmp < 0 {
```
itsbilal, line 349: I think the issue here is that the comparator isn't necessarily defined on `p.dst`; we require it to be defined on the prefix as returned by `Comparer.Split`, but this prefix could be much shorter than that. I would remove this optimization and just call the child method always.

sstable/reader_virtual.go
```diff
@@ -204,10 +248,264 @@ func (v *virtualState) constrainBounds(
...
+	p.res.Trailer = k.Trailer
+	p.res.UserKey = append(p.res.UserKey[:len(p.dst)], k.UserKey[len(p.src):]...)
+	return &p.res, v
+}
+
+// SeekGE implements the Iterator interface.
+func (p *prefixReplacingIterator) SeekGE(
+	key []byte, flags base.SeekGEFlags,
+) (*InternalKey, base.LazyValue) {
+	cmp := p.cmp(key, p.dst)
```
itsbilal, line 337: Same issue as below. I'd just remove this case.

itsbilal (approved):
LGTM save for comments

sstable/reader_virtual.go
dt: Something feels a bit off here? PrefixReplacingIterator claims to implement Iterator, which includes SetBounds. The contract is that takes "new" prefix keys and bounds results to be within them. 

The underlying iterator is over "old" keys, and has its contract for SetBounds , again in terms of old keys. They underlying iterator doesn't know it is being wrapped in a prefix iterator. 

So why would I pass "new" key bounds to the "old" key iterator?

itsbilal: You're right about the iterator _interface_. The issue is that the sstable iterators (both singleLevel and twoLevel) have a notion of what the VSST file bounds are (let's say a VSST has bounds `[c-e)`), and they get a set of bounds from the interface (let's say `[d-g)`), and the sstable iterator will call `constrainBounds` which will go "oh I need to impose `[d-e)` on myself because anything outside of that in this sstable could be garbage".

The VSST bounds in the `FileMetadata`, which is what `constrainBounds` uses, are in the post-transformation keyspace. The bounds coming from the iterator, with your change, were sometimes in the post-transformation keyspace (if they were passed in as a creation-time IterOption), or pre-transformation (if they got rewriteArg'd in SetBounds). In the latter case, constrainBounds would go "oh, the user wants `[de-di)`, I have `[ef-ei)`, therefore I have nothing of value and I'll return an empty iterator", which was inaccurate because it ignored the fact that there was a `d->e` prefix-transform rule. The reason why the tests are passing right now is because almost all bounds are set at iter creation time in the tests we've written, so SetBounds doesn't get called.

If we're transforming keys in reverse at the interface, we'd need to transform them back forward before we apply VSST file bounds in `constrainMeta`, and then in reverse again when we pass them to the `singleLevelIterator`. That's 3 one-way transformations for each bound change, so I think it's best to just leave the current logic in `constrainBounds` which just does it once and correctly.

dt: I see.

I'm still not happy with this though.

This feels like an abstraction leaking. Badly.

Lamme ponder a bit but if I can't find a better answer than just passing through (wrong, but then later right) raw keys I'll at least add giant comment.

dt: added optional `SetBoundsMaterialized` API for an underlying iter to indicate it wants these weird bounds and a big comment explaining the very subtle reason why. WDYT?

sstable/reader_virtual.go
```diff
@@ -204,10 +248,260 @@ func (v *virtualState) constrainBounds(
...
+}
+
+// Close implements the Iterator interface.
+func (p *prefixReplacingIterator) Close() error {
+	return p.i.Close()
+}
+
+// SetBounds implements the Iterator interface.
+func (p *prefixReplacingIterator) SetBounds(lower, upper []byte) {
+	//
```
itsbilal, line 400: dangling comment?

sstable/reader_iter_single_lvl.go
```diff
@@ -320,6 +320,24 @@ func disableBoundsOpt(bound []byte, ptr uintptr) bool {
...
+//
+// This allows an optimization in vState, whereby when this iterator passes it
+// bounds to constrain, since bounds are kept in terms of materialized keys, the
+// vState does not need to convert them from the "backing" prefix before it can
+// combine them with the virtual bounds, which are always in terms of
+// materialized keys. The vState then converts its constrained bounds to backing
+// bounds which are what SetBounds will ultimately apply, so essentially this is
+// just indicating that constrainBounds is responsible for the rewriting, rather
+// than the caller of SetBounds (the wrapping PrefixRewritingIterator).
+func (i singleLevelIterator) SetBoundsMaterialized() bool {
```
itsbilal, line 337: I guess this gets copied through to `twoLevelIterator` seeing as that struct extends/embeds `singleLevelIterator`, but yea this could be enough reason to update the test to pass in tiny block sizes in a second variant to force the use of two level iterators..

itsbilal (approved):
LGTM

sstable/reader_iter_single_lvl.go
dt: twoLevelIterator doesn't call `constrainBounds` directly; it _only_ calls it via the embed of SLI and its `SetBounds` call so I'm fine with this.

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3159)* status: 0 of 9 files reviewed, 10 unresolved discussions (waiting on @dt, @itsbilal, and @RaduBerinde)

___
*[`ingest.go` line 1120 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljCAHZ544UQFNqB9R5:-NljCAHZ544UQFNqB9R6:b9d71qf) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1120)):*
> ```Go
> 	HasPointKey, HasRangeKey bool
> 	// HasPrefixRule denotes whether this file has a prefix replacement rule.
> 	HasPrefixRule bool
> ```

Is it sensical to have `HasPrefixRule == true` when `len(Backing) == 0 && len(Materialized) == 0`? Could we get rid of `HasPrefixRule` and simply rely on `len(Backing) > 0 || len(Materialized) > 0` to indicate that a prefix replacement rule is present?
___
*[`ingest.go` line 1123 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljDFZG6BJKeUdY4ufl:-NljDFZG6BJKeUdY4ufm:b-y4afdd) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1123)):*
> ```Go
> 	// Backing and Materialized denote a prefix replacement rule, where the
> 	// external file contains the Backing prefix and Materialized contains the
> 	// prefix that iterators are expected to return.
> ```

Rather than talking about the behavior in terms of iterators I'd prefer to have the documentation reference the semantics. After ingestion the system will behave as if the sstable had been rewritten replacing the old prefix on keys with the new prefix. Every key in the ingested sstable must contain the old prefix (I think this can/should be checked by examining `SmallestUserKey` and `LargestUserKey`). Actual replacement of the prefix may occur lazily.
___
*[`ingest.go` line 1124 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljCWahDFbZc79_ePDT:-NljCWahDFbZc79_ePDU:bh8viar) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1124)):*
> ```Go
> 	// external file contains the Backing prefix and Materialized contains the
> 	// prefix that iterators are expected to return.
> 	Backing, Materialized []byte
> ```

The terms `Backing` and `Materialized` are not intuitive to me in this context. Looking at similar functionality in the stdlib (e.g. strings.Replace) I see the terminology `old` and `new`. Perhaps `OldPrefix` and `NewPrefix`.

dt (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3159)* status: 0 of 9 files reviewed, 10 unresolved discussions (waiting on @itsbilal, @petermattis, and @RaduBerinde)

___
*[`ingest.go` line 1123 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljDFZG6BJKeUdY4ufl:-NljEw8dBOODu3OyEK4c:b-er8jt2) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1123)):*
> I think this can/should be checked by examining `SmallestUserKey` and `LargestUserKey`

We do not open the file or read the footer during ingest so we can't check this, but yes, it is assumed/required.
___
*[`ingest.go` line 1124 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljCWahDFbZc79_ePDT:-NljG0TZE6yOqK40OiGb:bktvlq7) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1124)):*

The terms `Backing` and `Materialized` are not intuitive to me in this context. Looking at similar functionality in the stdlib (e.g. strings.Replace) I see the terminology `old` and `new`. Perhaps `OldPrefix` and `NewPrefix`.
</blockquote></details>

We had `old` and `new` in an early version of this feature but tried to find something else intentionally since replacement happens _in both directions_ in different layers, and `old` and `new` were creating some cognitive dissonance when doing the reverse transform. 

I'll ponder a little more.

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3159)* status: 0 of 9 files reviewed, 10 unresolved discussions (waiting on @dt, @itsbilal, and @RaduBerinde)

___
*[`ingest.go` line 1123 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljDFZG6BJKeUdY4ufl:-NljJQ0CCI0t1Jgqcy7v:bnc1de9) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1123)):*

> I think this can/should be checked by examining `SmallestUserKey` and `LargestUserKey`

We do not open the file or read the footer during ingest so we can't check this, but yes, it is assumed/required.
</blockquote></details>

`ExternalFile` does contain the smallest and largest key. You can at least sanity check against those values.

Do `{Smallest,Largest}UserKey` reflect the keys in the table being ingested, or the keys after the prefix replacement? (This should be documented).
___
*[`ingest.go` line 1124 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljCWahDFbZc79_ePDT:-NljJU8d9_Prmbs8h5Kp:b-7rvnr1) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1124)):*

We had `old` and `new` in an early version of this feature but tried to find something else intentionally since replacement happens _in both directions_ in different layers, and `old` and `new` were creating some cognitive dissonance when doing the reverse transform. 

I'll ponder a little more.
</blockquote></details>

The replacement happening in both directions in different layers is an implementation detail, right? Conceptually the replacement could be happening by rewriting the sstable at ingestion time. 

Related, are the smallest and largest key correct for these sstables in the output of `DB.SSTables()`?

dt (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3159)* status: 0 of 9 files reviewed, 10 unresolved discussions (waiting on @itsbilal, @petermattis, and @RaduBerinde)

___
*[`ingest.go` line 1120 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljCAHZ544UQFNqB9R5:-Nln6Har8oO0giuK7VNh:bpata0o) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1120)):*

Is it sensical to have `HasPrefixRule == true` when `len(Backing) == 0 && len(Materialized) == 0`? Could we get rid of `HasPrefixRule` and simply rely on `len(Backing) > 0 || len(Materialized) > 0` to indicate that a prefix replacement rule is present?
</blockquote></details>

Switched this to just len(materialized) != 0.
___
*[`ingest.go` line 1123 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljDFZG6BJKeUdY4ufl:-Nln4QPX1N6-amqWw_o0:bm10ib3) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1123)):*
> You can at least sanity check against those values.
We are enforcing this already, in the 	`if m.PrefixReplacement != nil` check in internal/manifest/version.go above. 

They're both in terms of the post-replacement key space. The only place where the client mentions the "old" (backing) prefix is in the replacement rule to specify what to replace, but everything else is kept in terms of the visible keys. I'll add this to the comment on the prefixes below.
___
*[`ingest.go` line 1124 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljCWahDFbZc79_ePDT:-Nln58PUB-u7XM8H7p_I:bcp7cef) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1124)):*
> an implementation detail

Indeed, but I think having the internal names used in that implementation match the public API names makes it a bit easier to follow?

What do you think about this?
```
// ContentPrefix and VirtualPrefix denote a prefix replacement rule causing a
// file, in which all keys have prefix ContentPrefix, to appear whenever it is
// accessed as if those keys all instead have prefix VirtualPrefix.
// VirtualPrefix must be a prefix of both SmallestUserKey and LargestUserKey.
ContentPrefix, VirtualPrefix \[\]byteContentPrefix, VirtualPrefix []byte
```

`DB.SSTables()` should be in terms of the virtual prefix, since that's what we stick in the manifest. We only every rewrite the span into the "old"/content prefix when we open an iterator.
___
*[`sstable/reader_virtual.go` line 337 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3159#gh-1427164704:-Nln8oJDCYn9mVfyceoB:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b237ab07aa76d1ade6ad3ad100b369a6fca7a1a7/sstable/reader_virtual.go#L337)):*

Same issue as below. I'd just remove this case.
</blockquote></details>

Done.
___
*[`sstable/reader_virtual.go` line 349 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3159#gh-1427164324:-Nln8ow-AbwL-KG2BEDW:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b237ab07aa76d1ade6ad3ad100b369a6fca7a1a7/sstable/reader_virtual.go#L349)):*

I think the issue here is that the comparator isn't necessarily defined on `p.dst`; we require it to be defined on the prefix as returned by `Comparer.Split`, but this prefix could be much shorter than that. I would remove this optimization and just call the child method always.
</blockquote></details>

Done.
___
*[`sstable/reader_virtual.go` line 466 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3159#gh-1427160198:-NlnDdTv59xt7ISoe_1q:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b237ab07aa76d1ade6ad3ad100b369a6fca7a1a7/sstable/reader_virtual.go#L466)):*

You don't need to touch sp.Keys, the bounds are only in sp.Start/End.
</blockquote></details>

Done.
___
*[`sstable/reader_virtual.go` line 400 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#gh-1428445772:-NlnDU7OBZ3dW-cotzS2:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/sstable/reader_virtual.go#L400)):*

dangling comment?
</blockquote></details>

Done.

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3159)* status: 0 of 9 files reviewed, 10 unresolved discussions (waiting on @dt, @itsbilal, and @RaduBerinde)

___
*[`ingest.go` line 1120 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljCAHZ544UQFNqB9R5:-NlnWLnEAMdYKp8GcWA4:b-dwgmbs) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1120)):*

Switched this to just len(materialized) != 0.
</blockquote></details>

Not sure if you forgot to push, but I'm not seeing that change.
___
*[`ingest.go` line 1123 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljDFZG6BJKeUdY4ufl:-NlnUajGAxBsLQxv3jyO:b9sgjjt) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1123)):*

> You can at least sanity check against those values.
We are enforcing this already, in the 	`if m.PrefixReplacement != nil` check in internal/manifest/version.go above. 

They're both in terms of the post-replacement key space. The only place where the client mentions the "old" (backing) prefix is in the replacement rule to specify what to replace, but everything else is kept in terms of the visible keys. I'll add this to the comment on the prefixes below.
</blockquote></details>

Ah, got it. Carry on.
___
*[`ingest.go` line 1124 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljCWahDFbZc79_ePDT:-NlnUidz5urzn7JgukF1:bsrkiy8) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1124)):*
> Indeed, but I think having the internal names used in that implementation match the public API names makes it a bit easier to follow?

I'm hoping we can get rid of the implementation detail that we're transforming keys in both directions. See https://github.com/cockroachdb/pebble/issues/2632#issuecomment-1858861943.

That comment looks better, though I'm not thrilled with the term "virtual" here as I suspect it can cause confusion with "virtual sstables". 

> `DB.SSTables()` should be in terms of the virtual prefix, since that's what we stick in the manifest. We only every rewrite the span into the "old"/content prefix when we open an iterator.

Good point, though this does seem like something worth testing/verifying.

dt (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3159)* status: 0 of 9 files reviewed, 10 unresolved discussions (waiting on @itsbilal, @petermattis, and @RaduBerinde)

___
*[`ingest.go` line 1120 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljCAHZ544UQFNqB9R5:-Nln_gln7LaNgWXYHmeL:b3ux5u1) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1120)):*

Not sure if you forgot to push, but I'm not seeing that change.
</blockquote></details>

Whoops, done now.
___
*[`ingest.go` line 1124 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljCWahDFbZc79_ePDT:-Nln_n570qrObG2AeePm:bdui6m) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1124)):*
> #2632

Yeah, this is what DR and Storage have discussed as the direction for the next iteration of this, when we start modifying the backup files specifically for online restore.

> cause confusion with "virtual sstables".

FWIW, virtual prefixes are only allowed on "virtual stables" for now so it doesn't worry me too much if it is read as "the prefix in which this virtual file appears" this that isn't entirely incorrect either.

I thought about "VisiblePrefix" or "ExposedPrefix" though my first reaction is the connotation of both of those was too specific to the appearance of returned results, leaving out behaviors like Seek.

Maybe SyntheticPrefix or SynthesizedPrefix?

petermattis (commented):
Did you consider extending the `sstable.Iterator` benchmarks (`BenchmarkTableIter*`) to measure the overhead of this approach?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3159#-:-Nlscp-rAWFkTYawgkZU:b-jykeh5)* status: 0 of 9 files reviewed, 11 unresolved discussions (waiting on @dt, @itsbilal, and @RaduBerinde)

___
*[`ingest.go` line 1124 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljCWahDFbZc79_ePDT:-NlsdqJXEo-UmpwGvfFS:bmgc14k) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1124)):*
> FWIW, virtual prefixes are only allowed on "virtual stables" for now so it doesn't worry me too much if it is read as "the prefix in which this virtual file appears" this that isn't entirely incorrect either.

Yet there is nothing truly tying prefix rewriting to virtual sstables. Seems better to keep the concepts distinct and compose them together to get the desired result. We could have prefix rewriting work normal ingested files for zero additional effort. Why is that beneficial? No concrete idea, but it keeps the base concepts simpler and composable.
___
*[`sstable/reader_virtual.go` line 269 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlsbhaA1u8Nd-eNJqAi:-NlsbhaA1u8Nd-eNJqAj:bofp9tv) ([raw file](https://github.com/cockroachdb/pebble/blob/3be1868a9776ab6d26a6b30f5f4e6f93da268f88/sstable/reader_virtual.go#L269)):*
> ```Go
> }
> 
> type prefixReplacingIterator struct {
> ```

Is there anything `VirtualReader` specific about this code? I'd move it to a separate `prefix_replacing_iter.go` file. That will make it more obvious what needs to be removed if we do replace the implementation in the future.
___
*[`sstable/reader_virtual.go` line 287 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3159#-Nlsbc5Y9VVR9ttxc8fF:-Nlsbc5Y9VVR9ttxc8fG:b-qh67g7) ([raw file](https://github.com/cockroachdb/pebble/blob/3be1868a9776ab6d26a6b30f5f4e6f93da268f88/sstable/reader_virtual.go#L287)):*
> ```Go
> // passed as an argument to methods on this iterator must have prefix `dst`, and
> // every key produced by the underlying iterator must have prefix `src`.
> func NewPrefixReplacingIterator(i Iterator, src, dst []byte, cmp base.Compare) Iterator {
> ```

Does this method need to be exported?
___
*[`sstable/reader_virtual_test.go` line 22 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlsbAcRBvzYXiDXAb8L:-NlsbAcRBvzYXiDXAb8M:b-skud8o) ([raw file](https://github.com/cockroachdb/pebble/blob/3be1868a9776ab6d26a6b30f5f4e6f93da268f88/sstable/reader_virtual_test.go#L22)):*
> ```Go
> 		{from: []byte("bb"), to: []byte("aa")},
> 		{from: []byte("aa"), to: []byte("zzz")},
> 		{from: []byte("zzz"), to: []byte("aa")},
> ```

Some additional cases worth testing: empty `from`, empty `to`, empty `from` and empty `to`, identical non-empty `from` and `to`. The last two should be a no-op and will likely never happen in practice.

dt (commented):
> extending the `sstable.Iterator` benchmarks

I'd lean towards adding microbenchmarks if/when we're ready to start working on performance, letting observations of the initial prototype once its integrated into CRDB inform which parts we want to benchmark and optimize first.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3159#-:-NlsqOrB5AtFnZyIc3Ij:b6we9iu)* status: 0 of 10 files reviewed, 11 unresolved discussions (waiting on @itsbilal, @petermattis, and @RaduBerinde)

___
*[`ingest.go` line 1124 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NljCWahDFbZc79_ePDT:-Nlsr2J7-kZVCUfhfRLN:b-8xp73p) ([raw file](https://github.com/cockroachdb/pebble/blob/d063b69f7ed508b74e8ba39da6d4f7f32b134a88/ingest.go#L1124)):*

> nothing truly tying prefix rewriting to virtual sstables

The only thing that vaguely links them is that we've said the requirement that the whole file have content that matches the prefix rule is OK, even when on-disk local SSTs often contain data for multiple different (crdb table) prefixes, because we can say to put the prefix rule on a virtual SST, as these can specify arbitrary bounds on the span of the underlying file they expose, and thus can elect to expose only the span that has the prefix being rewritten. If an underlying physical file has data for many different prefixes, you'd use multiple virtual tables pointing into it to expose each prefix, rewritten, separately.

But obviously that's only required if the whole file has multiple prefixes; if it all had one prefix we could, in theory, apply a prefix rule directly to it, although we're not exposing any facility to do so at this point.

Anyway, I've renamed the param to SyntheticPrefix.
___
*[`sstable/reader_virtual.go` line 269 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlsbhaA1u8Nd-eNJqAi:-NlxFzLiBOg3HlVbeuf_:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/3be1868a9776ab6d26a6b30f5f4e6f93da268f88/sstable/reader_virtual.go#L269)):*

Is there anything `VirtualReader` specific about this code? I'd move it to a separate `prefix_replacing_iter.go` file. That will make it more obvious what needs to be removed if we do replace the implementation in the future.
</blockquote></details>

Done.
___
*[`sstable/reader_virtual.go` line 287 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3159#-Nlsbc5Y9VVR9ttxc8fF:-NltyjuT1I6LDy08e9bl:bti227m) ([raw file](https://github.com/cockroachdb/pebble/blob/3be1868a9776ab6d26a6b30f5f4e6f93da268f88/sstable/reader_virtual.go#L287)):*

Does this method need to be exported?
</blockquote></details>

I had some vague hopes of switching key rewriting in CRDB's current (offline) RESTORE, that is done during iteration of backup ssts to produce ingest ssts, to use the same rewriting iterator wrapper so we'd just have one shared rewriter. That said, this is a vague plan and I can always export this later when I get to it, if I get to it at all (they're not quite 1:1 and ideally we're just deleting that implementation of RESTORE anyway so maybe this never happen).
___
*[`sstable/reader_virtual_test.go` line 22 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlsbAcRBvzYXiDXAb8L:-NlxDJPcD0JDBwTCIkzF:b9aysxd) ([raw file](https://github.com/cockroachdb/pebble/blob/3be1868a9776ab6d26a6b30f5f4e6f93da268f88/sstable/reader_virtual_test.go#L22)):*

Some additional cases worth testing: empty `from`, empty `to`, empty `from` and empty `to`, identical non-empty `from` and `to`. The last two should be a no-op and will likely never happen in practice.
</blockquote></details>

Added identity rewrite and empty from. 

Neither empty `to` is valid or possible currently, since a non-empty synthetic is how we decide we have a replacement rule and to use the iterator at all. Updated the doc string and add an assertion to the constructor to make this clearer.

petermattis (approved):
> I'd lean towards adding microbenchmarks if/when we're ready to start working on performance, letting observations of the initial prototype once its integrated into CRDB inform which parts we want to benchmark and optimize first.

Ack. I'm a bit uncomfortable with this approach given that we know this code is performance critical, but you've indicated this several times so I'll stand aside on any objection. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3159#-:-NlxIO640li3aDHDHKrI:b-g8lk44)* status: 0 of 10 files reviewed, 7 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

RaduBerinde (commented):
I agree we should benchmark this at some point but I'm ok delaying it (in addition to pebble-level benchmarks, we should add variants of the MVCC Get/Scan benchmarks). We will still need the benchmarks even if/when we push this down into the reader.

In general, I am also weary of going too far down the path of cutting corners in the name of getting a prototype running ASAP.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3159#-:-NlxcL_iEaeEMO6FCZw_:b-c97o6c)* status: 0 of 10 files reviewed, 9 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`internal/manifest/version_edit.go` line 66 at r16](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlxdD0SAGI0mQlIY1Bj:-NlxdD0SAGI0mQlIY1Bk:b-4num9) ([raw file](https://github.com/cockroachdb/pebble/blob/02c3b193cdf86f26c5c0eb5580dd75c0fadcc854/internal/manifest/version_edit.go#L66)):*
> ```Go
> 	customTagNonSafeIgnoreMask = 1 << 6
> 	customTagVirtual           = 66
> 	customTagPrefixRewrite     = 67
> ```

We are still on the 23.2 format version, we need to add a new one before we add new tags.
___
*[`sstable/prefix_replacing_iterator.go` line 80 at r16](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlxdNza4cspv9F2weCd:-NlxdNza4cspv9F2weCe:b-g0dgx3) ([raw file](https://github.com/cockroachdb/pebble/blob/02c3b193cdf86f26c5c0eb5580dd75c0fadcc854/sstable/prefix_replacing_iterator.go#L80)):*
> ```Go
> 	}
> 	if !bytes.HasPrefix(k.UserKey, p.src) {
> 		panic(errOutputPrefixMismatch)
> ```

I would keep the panic in `invariants.Enabled` but I'd just set the error otherwise. We don't want to bring everything down because of one bug that affects one table.

dt (commented):
> path of cutting corners in the name of getting a prototype running

I don't wan to cut any correctness corners to get this working, but I _do_ want to distinguish between correctness corners and performance improvements, with the latter being what I'm eager to keep in a separated future task. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3159#-:-NlzM9M57YUNPmawyAsg:b-zbj818)* status: 0 of 17 files reviewed, 7 unresolved discussions (waiting on @itsbilal)

___
*[`internal/manifest/version_edit.go` line 66 at r16](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlxdD0SAGI0mQlIY1Bj:-NlzLi7EAHBMPjp9LK6i:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/02c3b193cdf86f26c5c0eb5580dd75c0fadcc854/internal/manifest/version_edit.go#L66)):*

We are still on the 23.2 format version, we need to add a new one before we add new tags.
</blockquote></details>

Done.
___
*[`sstable/prefix_replacing_iterator.go` line 80 at r16](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlxdNza4cspv9F2weCd:-NlzLfLO0C-teewZdBPx:bg3rsyn) ([raw file](https://github.com/cockroachdb/pebble/blob/02c3b193cdf86f26c5c0eb5580dd75c0fadcc854/sstable/prefix_replacing_iterator.go#L80)):*
> invariants.Enabled

Done.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3159#-:-NlzZPRP8-d97D6CXv6m:bnfp4nl)* status: 0 of 17 files reviewed, 8 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`ingest.go` line 214 at r20](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlzZDXzAeW0PZg-rSFm:-NlzZDXzAeW0PZg-rSFn:bg0nlsz) ([raw file](https://github.com/cockroachdb/pebble/blob/166a4f9eab3a858851b430069110ff734abe19d5/ingest.go#L214)):*
> ```Go
> 	meta.FileBacking.Size = e.Size
> 
> 	if len(e.SyntheticPrefix) != 0 {
> ```

We would need to check the format version here, no?
___
*[`internal/manifest/version_edit.go` line 66 at r16](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlxdD0SAGI0mQlIY1Bj:-NlzZSzt7gtfSRUu9uUC:b-qbxl94) ([raw file](https://github.com/cockroachdb/pebble/blob/02c3b193cdf86f26c5c0eb5580dd75c0fadcc854/internal/manifest/version_edit.go#L66)):*

Done.
</blockquote></details>

Looks great, thanks!
___
*[`sstable/prefix_replacing_iterator.go` line 80 at r16](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlxdNza4cspv9F2weCd:-NlzZatH1R4gM49L8tHV:bxw6exn) ([raw file](https://github.com/cockroachdb/pebble/blob/02c3b193cdf86f26c5c0eb5580dd75c0fadcc854/sstable/prefix_replacing_iterator.go#L80)):*

> invariants.Enabled

Done.
</blockquote></details>

I think you may have applied it to the wrong check, I was thinking of the `bytes.HasPrefix()` check.

RaduBerinde:
*[`ingest.go` line 214 at r20](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlzZDXzAeW0PZg-rSFm:-Nm1O0In414dBAN__fne:b-46f30i) ([raw file](https://github.com/cockroachdb/pebble/blob/166a4f9eab3a858851b430069110ff734abe19d5/ingest.go#L214)):*

We would need to check the format version here, no?
</blockquote></details>

Sorry, I forgot to delete this comment, I saw afterwards that we are checking before we get here.

dt (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3159)* status: 0 of 17 files reviewed, 8 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`ingest.go` line 214 at r20](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlzZDXzAeW0PZg-rSFm:-NlzlYRp0VA_pitAV7db:b4c9lyr) ([raw file](https://github.com/cockroachdb/pebble/blob/166a4f9eab3a858851b430069110ff734abe19d5/ingest.go#L214)):*

Sorry, I forgot to delete this comment, I saw afterwards that we are checking before we get here.
</blockquote></details>

:thumbsup: 
___
*[`internal/manifest/version_edit.go` line 66 at r16](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlxdD0SAGI0mQlIY1Bj:-Nm2oVDu-Zk_3mJQGIdI:b-atpe3c) ([raw file](https://github.com/cockroachdb/pebble/blob/02c3b193cdf86f26c5c0eb5580dd75c0fadcc854/internal/manifest/version_edit.go#L66)):*

Looks great, thanks!
</blockquote></details>

Can you give this one last glance now that I rebased it over your refactor?
___
*[`sstable/prefix_replacing_iterator.go` line 80 at r16](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlxdNza4cspv9F2weCd:-Nlzo8iY6rpuIKsyecGK:blu2b1e) ([raw file](https://github.com/cockroachdb/pebble/blob/02c3b193cdf86f26c5c0eb5580dd75c0fadcc854/sstable/prefix_replacing_iterator.go#L80)):*

I think you may have applied it to the wrong check, I was thinking of the `bytes.HasPrefix()` check.
</blockquote></details>

Ah, I see. done.

RaduBerinde (approved):
:lgtm:

> Reviewed 2 of 3 files at r22, 1 of 3 files at r23, 6 of 7 files at r24, 8 of 8 files at r25, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3159#-:-Nm2uYyb-sA657nKoxRi:bnfp4nl)* status: all files reviewed, 8 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`internal/manifest/version_edit.go` line 66 at r16](https://reviewable.io/reviews/cockroachdb/pebble/3159#-NlxdD0SAGI0mQlIY1Bj:-Nm2uTRW1L8laDNAgR42:bb6ywmx) ([raw file](https://github.com/cockroachdb/pebble/blob/02c3b193cdf86f26c5c0eb5580dd75c0fadcc854/internal/manifest/version_edit.go#L66)):*

Can you give this one last glance now that I rebased it over your refactor?
</blockquote></details>

LGTM
___
*[`format_major_version_test.go` line 26 at r25](https://reviewable.io/reviews/cockroachdb/pebble/3159#-Nm2uHNX3Nj_BXcCGltU:-Nm2uHNX3Nj_BXcCGltV:b-3nkk91) ([raw file](https://github.com/cockroachdb/pebble/blob/0dd6490b1963f268bf753508d31ab3a6d638d651/format_major_version_test.go#L26)):*
> ```Go
> 	require.Equal(t, FormatDeleteSizedAndObsolete, FormatMajorVersion(15))
> 	require.Equal(t, FormatVirtualSSTables, FormatMajorVersion(16))
> ```

[nit] add a `require.Equal(t, FormatSyntheticPrefixes, FormatMajorVersion(17))`

dt:
TFTRs!

PR #3152, opened 2023-12-12
db: deprecate older format versions

RaduBerinde:
This change deprecates format versions below `FormatFlushableIngest`
(the 23.1 version) and sstable formats below `TableFormatPebblev1`.

As part of this change, we remove code that deals with split keys;
further simplifications will be made separately (e.g. removing atomic
unit logic, simplifying the truncation iterator).

We also remove all code related to the `CURRENT` file.

The plan is to merge this after we tag a `v1.0.0` release, which will
for the time being be the recommended version series for users other
than CockroachDB.

Informs #3064

PR #3161, opened 2023-12-14
db: remove code related to atomic units

### LINKED ISSUE #1913: db: remove code for atomic compaction units
After the 22.2 branch cut, we can begin removing code for supporting atomic compaction units. Cockroach 23.1 will have the guarantee that no user keys are split across any files, because the upgrade to 22.2 blocked on their removal.

RaduBerinde:
Note: the first commit is #3152.

Fixes #1913.

#### db: don't allow table overlaps in compaction tests

With all formats that tolerate split keys deprecated, we no longer
have to deal with tables overlapping in terms of user keys.

This commit adds a check to disallow any key overlap between tables in
compaction tests and removes offending test cases.

#### db: remove code related to atomic units

The concept of "atomic units" is related to expanding the boundary of
a compaction so that it does not cross a split user key; this can now
be removed.


RaduBerinde:
*[`testdata/compaction_setup_inputs` line 125 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3161#-NmBbHdT3vfcDsymV88I:-NmERUdj-O0M3sTRt8NO:b-wc61v9) ([raw file](https://github.com/cockroachdb/pebble/blob/8e0081d1fe3bd601e1e7655156005671c51045a0/testdata/compaction_setup_inputs#L125)):*

can you alter this test to not have overlapping user keys but continue to have the already compacting file sandwiched between two files that are going to be picked by the setup-inputs interval.
</blockquote></details>

Changing to
```
setup-inputs a d
L1
  a.SET.5-f.SET.6
L2
  a.SET.3-c.SET.4
  d.SET.3-e.SET.2 compacting
  f.SET.3-g.SET.6
----
```

Interestingly, the existing atomic units code did not deem it as `is-compacting`. It looks like it doesn't look at the files already in the range, only at the "expanding" edges.. Do you think this is a bug worth fixing in older versions?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3161#-:-NmEeqmd95RMALl3VIrq:b-814g8k)* status: 0 of 88 files reviewed, 5 unresolved discussions (waiting on @sumeerbhola)

___
*[`level_checker.go` line 280 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3161#-NmBjf2dCX6EVLVGpDZX:-NmEe5K5BKhD80BKb3_V:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/769e2ab8bfa4bd0c917a4264c41bdb1cb75f454e/level_checker.go#L280)):*

I think the tombstones should be already truncated when written by compactions and flushes, so we should not try to truncate before checking.
</blockquote></details>

Done.
___
*[`level_checker.go` line 392 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3161#-NmBjxVB6dFMVnU1mfb7:-NmEe91E9AkNxPNRTh5j:b-ybzxe9) ([raw file](https://github.com/cockroachdb/pebble/blob/769e2ab8bfa4bd0c917a4264c41bdb1cb75f454e/level_checker.go#L392)):*

why this commented out code?
</blockquote></details>

Ah, this was my draft way of disabling the tombstone truncation logic. Cleaned up.
___
*[`testdata/compaction_expand_inputs` line 66 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3161#-NmBaf5Z8BjIEbbzsYQr:-NmENfuf7jNSl-Usr4mH:bez9r9d) ([raw file](https://github.com/cockroachdb/pebble/blob/8e0081d1fe3bd601e1e7655156005671c51045a0/testdata/compaction_expand_inputs#L66)):*

this is the only RANGEDEL case in this file and this is not considered overlapping wrt user key so we should retain this part of it.
</blockquote></details>

Done, I adjusted the last table since it was overlapping in `d`
___
*[`testdata/compaction_setup_inputs_multilevel_write_amp` line 284 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3161#-NmBdesi6TeGij6PpKKB:-NmEdJmy3AB-0kUyIj4f:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8e0081d1fe3bd601e1e7655156005671c51045a0/testdata/compaction_setup_inputs_multilevel_write_amp#L284)):*

If we have
L1
a-c
L2
a-d
d-e
L3
a-b
c-g

I think the d-e file in L2 will not get picked if the maxSize limit is exceeded.
</blockquote></details>

Done.

RaduBerinde:
This is RFAL. Thanks!

sumeerbhola (commented):
:lgtm: 

thanks for doing this!

> Reviewed 2 of 88 files at r7, 8 of 86 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3161#-:-NnKddMvD-ncTTjAn38t:bgywee3)* status: 10 of 88 files reviewed, 6 unresolved discussions (waiting on @RaduBerinde)

___
*[`compaction_picker.go` line 420 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3161#-NnKdSA7-EFPj7pm13x7:-NnKdSA7-EFPj7pm13x8:b-a6vwsu) ([raw file](https://github.com/cockroachdb/pebble/blob/fed7d75f8289372207b58c8fda4de18101aac5da/compaction_picker.go#L420)):*
> ```Go
> 
> 	// Determine the sstables in the output level which overlap with the input
> 	// sstables, and then expand those tables to a clean cut. No need to do
> ```

nit: "and then expand those tables to a clean cut" is now stale

sumeerbhola (approved):

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3161#-:-NnKeYTA1qIvmxFMChBW:b-814g8k)* status: 10 of 88 files reviewed, 5 unresolved discussions (waiting on @sumeerbhola)

PR #3166, opened 2023-12-15
internal/manifest: move CalculateInuseKeyRanges

jbowens:
Move the pebble package's calculateInuseKeyRanges function into the manifest package as a method on Version. This is a more obvious place for this logic to live, but this is also done in preparation for a randomized test of CalculateInuseKeyRanges that will need to live in an external package (eg, manifest_test). Exporting it from the manifest package will allow the external test package to access the method.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1003,8 +1003,8 @@ func (c *compaction) setupInuseKeyRanges() {
 	}
 	// calculateInuseKeyRanges will return a series of sorted spans. Overlapping
 	// or abutting spans have already been merged.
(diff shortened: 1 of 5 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3166#-:-NlxqSZNCPlGShDI-BDY:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jbowens)

PR #3169, opened 2023-12-17
Fix mipsle target

dmytro-vovk:
Make it possible to build using `mipsle` as target arch when imported into a project:
```shell
GOOS=linux GOARCH=mipsle go build main.go
```


xxmplus:
Close to due to age

PR #3170, opened 2023-12-18
external_iterator: don't double-close Readers in case of err

itsbilal:
Previously, we'd sometimes double-close Readers if `finishInitializingExternal` returned an error. This is because we would put all the sstable.Readers in a slice so we could close them in case of error, but we also called dbi.Close() which closes them in one return case. Double-closing would lead to an `inconsistent reference count: -1` panic.

Informs cockroachdb/cockroach#116566 (will fix when lands in cockroach).

### DIFF
--- a/external_iterator.go
+++ b/external_iterator.go
@@ -101,16 +101,6 @@ func NewExternalIterWithContext(
 
 	var readers [][]*sstable.Reader
 
-	// Ensure we close all the opened readers if we error out.
-	defer func() {
-		if err != nil {
-			for i := range readers {
-				for j := range readers[i] {
-					_ = readers[i][j].Close()
-				}
-			}
-		}
-	}()
 	seqNumOffset := 0
 	var extraReaderOpts []sstable.ReaderOption
 	for i := range extraOpts {
@@ -120,13 +110,18 @@ func NewExternalIterWithContext(
 		seqNumOffset += len(levelFiles)
 	}
 	for _, levelFiles := range files {
-		var subReaders []*sstable.Reader
 		seqNumOffset -= len(levelFiles)
-		subReaders, err = openExternalTables(o, levelFiles, seqNumOffset, o.MakeReaderOptions(), extraReaderOpts...)
+		subReaders, err := openExternalTables(o, levelFiles, seqNumOffset, o.MakeReaderOptions(), extraReaderOpts...)
 		readers = append(readers, subReaders)
-	}
-	if err != nil {
-		return nil, err
+		if err != nil {
+			// Close all the opened readers.
+			for i := range readers {
+				for j := range readers[i] {
+					_ = readers[i][j].Close()
+				}
+			}
+			return nil, err
+		}
 	}
 
 	buf := iterAllocPool.Get().(*iterAlloc)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3170)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`external_iterator.go` line 104 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3170#-NlzYBrtEj3gFPihSXC-:-NlzYBrtEj3gFPihSXC0:b57a0mm) ([raw file](https://github.com/cockroachdb/pebble/blob/a987116b2c02c4c011d162344e3d99de2ce00767/external_iterator.go#L104)):*
> ```Go
> 	var readers [][]*sstable.Reader
> 
> 	// Ensure we close all the opened readers if we error out.
> ```

We have exactly one exit path where we need to free these readers, I would just move this there instead of the `defer`.
___
*[`external_iterator.go` line 125 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3170#-NlzY3Eu7APEydlONEEI:-NlzY3Eu7APEydlONEEJ:b-y75b51) ([raw file](https://github.com/cockroachdb/pebble/blob/a987116b2c02c4c011d162344e3d99de2ce00767/external_iterator.go#L125)):*
> ```Go
> 		var subReaders []*sstable.Reader
> 		seqNumOffset -= len(levelFiles)
> 		subReaders, err = openExternalTables(o, levelFiles, seqNumOffset, o.MakeReaderOptions(), extraReaderOpts...)
> ```

Are we ignoring errors here (other than the last file)?

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3170#-:-Nm2rsowDdML3dckUSFs:b-814g8k)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`external_iterator.go` line 104 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3170#-NlzYBrtEj3gFPihSXC-:-Nm2rrxB7IBFRUrg67Gx:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/a987116b2c02c4c011d162344e3d99de2ce00767/external_iterator.go#L104)):*

We have exactly one exit path where we need to free these readers, I would just move this there instead of the `defer`.
</blockquote></details>

Done.
___
*[`external_iterator.go` line 125 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3170#-NlzY3Eu7APEydlONEEI:-Nm2rcnx4D-DkjoiDXBR:b-xfwwts) ([raw file](https://github.com/cockroachdb/pebble/blob/a987116b2c02c4c011d162344e3d99de2ce00767/external_iterator.go#L125)):*

Are we ignoring errors here (other than the last file)?
</blockquote></details>

Good catch - fixed.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3170#-:-Nm2ujFc2EC3PPqRoLLW:bnfp4nl)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`external_iterator.go` line 121 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3170#-Nm2udjv0JuO_dSL5D1u:-Nm2udjv0JuO_dSL5D1v:b-ufv7ge) ([raw file](https://github.com/cockroachdb/pebble/blob/3342186f50e12254f0864ca9d99e9eaf2aa118ea/external_iterator.go#L121)):*
> ```Go
> 		}
> 	}
> 	if err != nil {
> ```

[supernit] we can just move this directly in the `err != nil` path above, and change to `subReaders, err := `

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3170#-:-Nm2zwlp3iwCcG4u-MaZ:b-814g8k)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`external_iterator.go` line 121 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3170#-Nm2udjv0JuO_dSL5D1u:-Nm2zxTW93cDlVmUr2_3:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/3342186f50e12254f0864ca9d99e9eaf2aa118ea/external_iterator.go#L121)):*

[supernit] we can just move this directly in the `err != nil` path above, and change to `subReaders, err := `
</blockquote></details>

Done.

PR #3171, opened 2023-12-18
sstable: update comment about SINGLEDEL, foreign ssts and compactions

sumeerbhola:
Also added a comment about optimizing compaction reads of these
foreign ssts.

### DIFF
--- a/sstable/format.go
+++ b/sstable/format.go
@@ -134,9 +134,31 @@ const (
 //
 // Note that we do not need to do anything special at write time for
 // SETWITHDEL and SINGLEDEL. This is because these key kinds are treated
-// specially only by compactions, which do not hide obsolete points. For
-// regular reads, SETWITHDEL behaves the same as SET and SINGLEDEL behaves the
-// same as DEL.
+// specially only by compactions, which typically do not hide obsolete points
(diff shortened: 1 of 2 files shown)


sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3171#-:-Nm1_EnoCniRuo1_T75f:b-814g8k)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`sstable/format.go` line 158 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3171#-Nlzb62o7LLPdP9uXHrd:-Nm1_1rk8zqH5eRyl2RC:b-lcrgmh) ([raw file](https://github.com/cockroachdb/pebble/blob/b224f98096d22ba042a62605fef212ffbace0407/sstable/format.go#L158)):*

"The fact that DELs can be hidden by SINGLEDELs is harmless"?
</blockquote></details>

Bad phrasing, since SINGLEDELs are not allowed to occur. I rephrased.

sumeerbhola:
TFTRs!

PR #3172, opened 2023-12-18
db: code comment about guaranteeing durability of initial state

sumeerbhola:
This is in response to the hazard raised in
https://cockroachlabs.slack.com/archives/C0KB9Q03D/p1702923954950209?thread_ts=1702921711.638549&cid=C0KB9Q03D (internal link).

### DIFF
--- a/open.go
+++ b/open.go
@@ -680,10 +680,17 @@ func GetVersion(dir string, fs vfs.FS) (string, error) {
 	return version, nil
 }
 
-// replayWAL replays the edits in the specified log file. If the DB is in
-// read only mode, then the WALs are replayed into memtables and not flushed. If
-// the DB is not in read only mode, then the contents of the WAL are guaranteed
-// to be flushed.
+// replayWAL replays the edits in the specified log file. If the DB is in read
+// only mode, then the WALs are replayed into memtables and not flushed. If
+// the DB is not in read only mode, then the contents of the WAL are
+// guaranteed to be flushed. Note that this flushing is very important for
+// guaranteeing durability: the application may have had a number of pending
+// fsyncs to the WAL before the process crashed, and those fsyncs may not have
+// happened but the corresponding data may now be readable from the WAL (while
+// sitting in write-back caches in the kernel or the storage device). By
+// reading the WAL (including the non-fsynced data) and then flushing all
+// these changes (flush does fsyncs), we are able to guarantee that the
+// initial state of the DB is durable.
 //
 // The toFlush return value is a list of flushables associated with the WAL
 // being replayed which will be flushed. Once the version edit has been applied


petermattis (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3172)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal, @pavelkalinnikov, and @sumeerbhola)

___
*[`open.go` line 693 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3172#-NlykDPu4JDx0_5cBiLe:-NlykDPu4JDx0_5cBiLf:b-eqire5) ([raw file](https://github.com/cockroachdb/pebble/blob/041e23f99493f63303505888c4e4cd02c3475175/open.go#L693)):*
> ```Go
> // reading the WAL (including the non-fsynced data) and then flushing all
> // these changes (flush does fsyncs), we are able to guarantee that the
> // initial state of the DB is durable.
> ```

While you're looking at this code, can you verify the MANIFEST is either synced or rotated at startup? I seem to recall one or the other is done.

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3172#-:-Nlz2ijM31LI9eIvyWa3:b-814g8k)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal, @pavelkalinnikov, and @petermattis)

___
*[`open.go` line 693 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3172#-NlykDPu4JDx0_5cBiLe:-Nlz2CP53RatZAeqdgdM:b-y5tysp) ([raw file](https://github.com/cockroachdb/pebble/blob/041e23f99493f63303505888c4e4cd02c3475175/open.go#L693)):*

While you're looking at this code, can you verify the MANIFEST is either synced or rotated at startup? I seem to recall one or the other is done.
</blockquote></details>

A new MANIFEST is created, and all the work in `versionSet.logAndApply` happens with the flushed files, so we do fsync that new MANIFEST.

petermattis (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3172)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @itsbilal and @pavelkalinnikov)

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3172#-:-Nlz_W9I6PBh7AHz5erH:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @itsbilal and @pavelkalinnikov)

open.go
```diff
@@ -680,10 +680,17 @@ func GetVersion(dir string, fs vfs.FS) (string, error) {
...
+// only mode, then the WALs are replayed into memtables and not flushed. If
+// the DB is not in read only mode, then the contents of the WAL are
+// guaranteed to be flushed. Note that this flushing is very important for
+// guaranteeing durability: the application may have had a number of pending
+// fsyncs to the WAL before the process crashed, and those fsyncs may not have
+// happened but the corresponding data may now be readable from the WAL (while
+// sitting in write-back caches in the kernel or the storage device). By
+// reading the WAL (including the non-fsynced data) and then flushing all
+// these changes (flush does fsyncs), we are able to guarantee that the
+// initial state of the DB is durable.
```
pavelkalinnikov, line 693: Is it necessary to flush/sync the memtables to give this guarantee? Why not just fsync the WAL?

Seemingly, fsyncing the SSTs is necessary only before WAL is truncated?

pavelkalinnikov (commented):
Would be good to elaborate the `Open()` comment too. It's surprisingly empty for a key function in the package:

```
// Open opens a DB whose files live in the given directory.
func Open(dirname string, opts *Options) (db *DB, _ error)
```

Outline at some level of detail what steps it consists of, and what guarantees it gives. Otherwise the caller needs to dig into the code.

open.go
```diff
@@ -680,10 +680,17 @@ func GetVersion(dir string, fs vfs.FS) (string, error) {
 	return version, nil
 }
 
-// replayWAL replays the edits in the specified log file. If the DB is in
-// read only mode, then the WALs are replayed into memtables and not flushed. If
-// the DB is not in read only mode, then the contents of the WAL are guaranteed
-// to be flushed.
+// replayWAL replays the edits in the specified log file. If the DB is in read
+// only mode, then the WALs are replayed into memtables and not flushed. If
```
pavelkalinnikov, line 684: How is read-only mode used? Is it mostly informational?

Do we need to guarantee that reads returned by read-only mode are durable? If so, can we [optionally?] fsync the WAL in read-only mode (even though it's read-only)?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3172)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @pavelkalinnikov)

___
*[`open.go` line 684 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3172#gh-1431259267:-Nm1N2hREnAIjyaS4QTs:bw5923t) ([raw file](https://github.com/cockroachdb/pebble/blob/041e23f99493f63303505888c4e4cd02c3475175/open.go#L684)):*
<details><summary><i>Previously, pavelkalinnikov (Pavel Kalinnikov) wrote…</i></summary><blockquote>

How is read-only mode used? Is it mostly informational?

Do we need to guarantee that reads returned by read-only mode are durable? If so, can we [optionally?] fsync the WAL in read-only mode (even though it's read-only)?
</blockquote></details>

It's mostly used by the command line tools.
___
*[`open.go` line 693 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3172#gh-1431248434:-Nm1N8ah8kfPQmpk8KB3:bnlh02u) ([raw file](https://github.com/cockroachdb/pebble/blob/041e23f99493f63303505888c4e4cd02c3475175/open.go#L693)):*
<details><summary><i>Previously, pavelkalinnikov (Pavel Kalinnikov) wrote…</i></summary><blockquote>

Is it necessary to flush/sync the memtables to give this guarantee? Why not just fsync the WAL?

Seemingly, fsyncing the SSTs is necessary only before WAL is truncated?
</blockquote></details>

This is what we were doing anyway, before this particular possibility (of data being currently available, but not durable) came to light.

We need to sync the SSTs for other reasons. They need to be durable before they can be referenced in the manifest. We don't have the ability to "drop" them later, even if the relevant data is in the WAL.

pavelkalinnikov (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3172)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`open.go` line 693 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3172#gh-1431248434:-Nm1igEV13yNofVBsiWe:b-pwcoiv) ([raw file](https://github.com/cockroachdb/pebble/blob/041e23f99493f63303505888c4e4cd02c3475175/open.go#L693)):*
I might misunderstand (please correct me). In my simplistic mental model, the "durable" lifetime of a row in normal operation looks like this (not considering ingestions and other complications):

```
1. write to WAL  // not yet durable
2. fsync the WAL // durable from now on. Currently just in the WAL
3. write to LSM (via SST files)  // memtable flush
4. fsync SST files and parent directory
5. mention new SST files in the manifest
6. fsync the manifest // now durable in the LSM too
7. truncate the WAL  // now the row is durable only in LSM/SSTs
```

There are probably shortcuts here, e.g. I imagine that ingestions enter this pipeline at step (3), and are never durable via the WAL (except when they do, with "flushable ingestions"?). Anyway.

Depending on where a crash occurs, we may have some data in the FS but not durable:
- for regular writes, that's between step (1) and (2)
- for ingestions, that's between step (3) and (5)

I make a few conclusions about the recovery procedure:

1. We need to do step (2) at startup unconditionally, to make sure what we read is durable.
2. Some SSTs might be missing from the manifest (if we crash before step 5). Those either need to be removed (and replayed later - the higher level knows it's still not durable), or followed through and added to the manifest (if it can be done based on the available information).
3. Those SSTs that are in the manifest after a crash are guaranteed to be synced (because step 5 "happens after" step 4, obv). So we don't need to sync those SSTs.
4. Either way, we need to do the step (6) fsync the manifest.

What I don't see is why replaying into memtables and flushing it is a necessary durability condition to return from `Open`. It strikes me as an unrelated implementation detail (maybe you need it for some other invariants).

> We need to sync the SSTs for other reasons. They need to be durable before they can be referenced in the manifest. We don't have the ability to "drop" them later, even if the relevant data is in the WAL.

Is it always possible to not drop an SST? What if the crash occurred between steps 3 and 5 and some information/validation is lost that must precede adding to the manifest?

RaduBerinde:
*[`open.go` line 693 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3172#gh-1431248434:-Nm1v25-9oQ5fi-CJFHr:b-b9dzry) ([raw file](https://github.com/cockroachdb/pebble/blob/041e23f99493f63303505888c4e4cd02c3475175/open.go#L693)):*
<details><summary><i>Previously, pavelkalinnikov (Pavel Kalinnikov) wrote…</i></summary><blockquote>

I might misunderstand (please correct me). In my simplistic mental model, the "durable" lifetime of a row in normal operation looks like this (not considering ingestions and other complications):

```
1. write to WAL  // not yet durable
2. fsync the WAL // durable from now on. Currently just in the WAL
3. write to LSM (via SST files)  // memtable flush
4. fsync SST files and parent directory
5. mention new SST files in the manifest
6. fsync the manifest // now durable in the LSM too
7. truncate the WAL  // now the row is durable only in LSM/SSTs
```

There are probably shortcuts here, e.g. I imagine that ingestions enter this pipeline at step (3), and are never durable via the WAL (except when they do, with "flushable ingestions"?). Anyway.

Depending on where a crash occurs, we may have some data in the FS but not durable:
- for regular writes, that's between step (1) and (2)
- for ingestions, that's between step (3) and (5)

I make a few conclusions about the recovery procedure:

1. We need to do step (2) at startup unconditionally, to make sure what we read is durable.
2. Some SSTs might be missing from the manifest (if we crash before step 5). Those either need to be removed (and replayed later - the higher level knows it's still not durable), or followed through and added to the manifest (if it can be done based on the available information).
3. Those SSTs that are in the manifest after a crash are guaranteed to be synced (because step 5 "happens after" step 4, obv). So we don't need to sync those SSTs.
4. Either way, we need to do the step (6) fsync the manifest.

What I don't see is why replaying into memtables and flushing it is a necessary durability condition to return from `Open`. It strikes me as an unrelated implementation detail (maybe you need it for some other invariants).

> We need to sync the SSTs for other reasons. They need to be durable before they can be referenced in the manifest. We don't have the ability to "drop" them later, even if the relevant data is in the WAL.

Is it always possible to not drop an SST? What if the crash occurred between steps 3 and 5 and some information/validation is lost that must precede adding to the manifest?
</blockquote></details>

Yes! That's what I was saying 

above- it is an unrelated implementation detail, it just happens to also address this issue without us having to sync the WAL at startup.

If we crash before adding an sst to the manifest, any unreferenced sst file will be treated as trash and cleaned up. It is not allowed to mention an sst in the manifest if it's not fully durable.

pavelkalinnikov (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3172)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`open.go` line 693 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3172#gh-1431248434:-Nm1xm01AvBl8p3GJWvO:b-n8dubh) ([raw file](https://github.com/cockroachdb/pebble/blob/041e23f99493f63303505888c4e4cd02c3475175/open.go#L693)):*

Yes! That's what I was saying 

above- it is an unrelated implementation detail, it just happens to also address this issue without us having to sync the WAL at startup.

If we crash before adding an sst to the manifest, any unreferenced sst file will be treated as trash and cleaned up. It is not allowed to mention an sst in the manifest if it's not fully durable.
</blockquote></details>

Aaah, right! I think I got it. We make the data durable by replaying a tail of it into the LSM.

However, shouldn't we still fsync the WAL? Assume we don't. Then after replaying and syncing the LSM, a tail of the WAL may still be not durable. It means that the LSM durability has "overtaken" the WAL durability. The WAL may regress after a crash, and we'll end up in a state where there is a gap between the WAL and LSM.

Shouldn't LSM durable state always be <= WAL durable state? Or pebble recovers correctly in both cases? I guess in such cases the WAL can be effectively considered empty.

petermattis:
Fsyncing the MANIFEST and truncating the WAL are an atomic operation. (The MANIFEST update mentions that a new WAL is active which implies the old WAL is garbage). My recollection is that we do have tests around durability at startup, but I could be mistaken.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3172)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @itsbilal, @pavelkalinnikov, @petermattis, and @RaduBerinde)

___
*[`open.go` line 693 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3172#gh-1431248434:-Nm2CCy_26_2vEHONqwM:b-yn1ido) ([raw file](https://github.com/cockroachdb/pebble/blob/041e23f99493f63303505888c4e4cd02c3475175/open.go#L693)):*
<details><summary><i>Previously, pavelkalinnikov (Pavel Kalinnikov) wrote…</i></summary><blockquote>

Aaah, right! I think I got it. We make the data durable by replaying a tail of it into the LSM.

However, shouldn't we still fsync the WAL? Assume we don't. Then after replaying and syncing the LSM, a tail of the WAL may still be not durable. It means that the LSM durability has "overtaken" the WAL durability. The WAL may regress after a crash, and we'll end up in a state where there is a gap between the WAL and LSM.

Shouldn't LSM durable state always be <= WAL durable state? Or pebble recovers correctly in both cases? I guess in such cases the WAL can be effectively considered empty.
</blockquote></details>

There are many WAL files (totally ordered). Each flush fully handles *all* the data in one or more WAL files.
When we update and fsync the MANIFEST after a flush, any sstables referred to in the new manifest entry are already guaranteed to be durable (since we've fsynced them), and the act of updating the MANIFEST also specifies up to what WAL we have flushed (MinUnflushedLogNum), so once the MANIFEST is synced it doesn't matter if we happened to have flushed some data in a WAL that was not fsynced (we will never look at those WAL files again) -- the presence of MinUnflushedLogNum is why @petermattis was saying "Fsyncing the MANIFEST and truncating the WAL are an atomic operation".

sumeerbhola:
TFTRs!

PR #3176, opened 2023-12-20
metamorphic: add --try-to-reduce flag

RaduBerinde:
This change adds a variation of `RunOnce` where we try to reduce the
number of ops that reproduce a problem.

Currently it can be used by tests that fail (e.g. with a panic or
internal error); extending it to "compare" mode in the future should
be easy.

Sample usage:
```
go test -run TestMetaTwoInstance ./internal/metamorphic \
  --run-dir _meta/231220-073015.8533726292752/random-027 \
  --tags invariants --try-to-reduce -v
```

The test will save the smallest reproduction found and print out the
relevant information.

### DIFF
--- a/internal/metamorphic/meta_test.go
+++ b/internal/metamorphic/meta_test.go
@@ -63,6 +63,10 @@ func TestMeta(t *testing.T) {
 		// runOptions() below) or the user specified it manually in order to re-run
 		// a test.
 		onceOpts := runOnceFlags.MakeRunOnceOptions()
+		if runOnceFlags.TryToReduce {
+			tryToReduce(t, runOnceFlags.Dir, runOnceFlags.RunDir)
+			return
+		}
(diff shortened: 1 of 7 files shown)


jbowens (approved):
Nice! :lgtm:

How quickly can it reduce? Removing whole objects (eg, a whole Iterator and all operations performed against it) could reduce quickly at the beginning without creating invalid operation sequences, but it might not be worth any additional complexity.

> Reviewed 5 of 7 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3176#-:-NmCKI7t8x_vcjQho-VR:b-b6b8ix)* status: 6 of 7 files reviewed, all discussions resolved

RaduBerinde:
In the issue I used it for, it reduced from 7000 to under 10 ops in (what I recall felt like) less than a minute.

PR #3178, opened 2023-12-20
internal/manifest: add TestInUseKeyRangesRandomized

jbowens:
This commit adds a randomized test for CalculateInUseKeyRanges. One possible explanation for #114421 could be a compaction incorrectly eliding a tombstone too early, before it's deleted the key it's intended to delete. When a compaction begins, we calculate the "in-use key ranges" beneath the compaction. If a tombstone overlaps an in-use key range, it cannot be elided because there may exist a key that the tombstone should delete within the in-use key range.

This randomized test is intended to shake out any unconsidered edge cases within the non-trivial calculation of in-use key ranges performed during compaction setup.

### DIFF
--- a/internal/manifest/manifest_test.go
+++ b/internal/manifest/manifest_test.go
@@ -1,0 +1,129 @@
+// Copyright 2023 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package manifest_test
+
+import (
+	"io"
+	"testing"
+	"time"
+
+	"github.com/cockroachdb/errors/oserror"
+	"github.com/cockroachdb/pebble"
+	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/manifest"
+	"github.com/cockroachdb/pebble/internal/testkeys"
+	"github.com/cockroachdb/pebble/metamorphic"
+	"github.com/cockroachdb/pebble/record"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/exp/rand"
+)
+
+// TestInuseKeyRangesRandomized is a randomized test that generates a random
+// database (using the Pebble metamorphic tests), and then tests calculating
+// in-use key ranges for random spans. It asserts that all files that overlap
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):
:lgtm:

Cool!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3178#-:-Nm8pUMr9VDK4FT60rqc:blc42av)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`internal/manifest/manifest_test.go` line 90 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3178#-Nm8p0Md7YI0KWMcj6I9:-Nm8p0Md7YI0KWMcj6IA:b-ah415a) ([raw file](https://github.com/cockroachdb/pebble/blob/215308449130be77efbcb55d68fde8f7cafc29ed/internal/manifest/manifest_test.go#L90)):*
> ```Go
> 				}
> 				if !containedWithin {
> 					t.Errorf("file L%d.%s overlaps [%s, %s] but no in-use key range contains it", l, f, smallest, largest)
> ```

[nit] Consider making this Fatal, and moving the `Logf` in here - we'd want to only see the info for the case that failed.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3178#-:-NmCXu1aBx3Nx_WkgZ3B:b-814g8k)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`internal/manifest/manifest_test.go` line 90 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3178#-Nm8p0Md7YI0KWMcj6I9:-NmCXifH9NFTUbzhuKyA:b-k2ulb0) ([raw file](https://github.com/cockroachdb/pebble/blob/215308449130be77efbcb55d68fde8f7cafc29ed/internal/manifest/manifest_test.go#L90)):*

[nit] Consider making this Fatal, and moving the `Logf` in here - we'd want to only see the info for the case that failed.
</blockquote></details>

Updated to Fatal, but kept the Logfs since I found them useful when working with the test (and they only show up if the verbose flag is passed to `go`)

Issue #3174, opened 2023-12-19
metamorphic test failure: inconsistent keys in range


RaduBerinde:
CI ran into a metamorphic test failure in PR #3173. It's unlikely (but not impossible) that the PR itself introduces the problem (it's mostly code cleanup).

https://github.com/cockroachdb/pebble/actions/runs/7266582327/job/19798611718?pr=3173

```

		1703013741613279788
		===== DIFF =====
		_meta/231219-192221.6133753307642/{standard-000,random-009}
		@@ -2103,11 +2103,11 @@
 iter20.SetBounds("rvxxunegblh@1", "tjpwuigvwge@1") // <nil> #2102
 iter20.SeekLT("tjpwuigvwge@1", "") // [true,"tjaj",<no point>,["tjaj","tjpwuigvwge@1")=>{"@6"="gqxiqxgpiskukyar"}*] <nil> #2103
 iter20.Next("") // [false] <nil> #2104
 iter20.Next("") // [false] <nil> #2105
 snap17.Close() // <nil> #2106
-iter22.Next("") // [true,"rpsllbp",<no point>,["rpsllbp","rvxxunegblh")=>{"@6"="iclhhyqvjdtfc","@4"="qumipltbbtehcoala","@3"="gtsviaw"}*] <nil> #2107
+iter22.Next("") // [true,"rpsllbp",<no point>,["rpsllbp","rvxxunegblh")=>{""="aevhlsxcjbxypnnas","@6"="iclhhyqvjdtfc","@4"="qumipltbbtehcoala","@3"="gtsviaw"}*] <nil> #2107
```

RaduBerinde:
Reproduced on master (69994ddb7235eb42f5377f3af66556e57944ed07):

```
		===== SEED =====
		1703020073158492000
		===== DIFF =====
		_meta/231219-130753.1583218548355/{standard-000,random-027}
		@@ -4276,11 +4276,11 @@
 iter44.Prev("cksgzvdspi@8") // [valid,"gwmb",<no point>,["gwmb","jjbudo")=>{""="irymnotaxnqtzfxjdv","@9"="kyet","@6"=""}*] <nil> #4275
 batch30.Delete("zsgbcatts@6") // <nil> #4276
 db2.Ingest(batch30) // <nil> #4277
 iter44.SeekPrefixGE("zkqrrqt@5") // [false] pebble: SeekPrefixGE supplied with key outside of upper bound #4278
[...]

PR #3179, opened 2023-12-21
db: set hideObsoletePoints for "boomerang" files

RaduBerinde:
#### db: add version checks around shared files

We now require the `VirtualSSTables` format to use shared files. This
helps with replication via shared ingest and excise, which requires V4
table format.

We update the metamorphic tests to use the correct version when shared
storage is enabled.

#### db: set hideObsoletePoints for "boomerang" files

We weren't setting this flag correctly When a shared file is ingested
by another store and then ingested back into the original store,
leading to an "out of order keys" error in `TestMetaTwoInstances`.

Informs #3174.

### DIFF
--- a/format_major_version.go
+++ b/format_major_version.go
@@ -197,6 +197,10 @@ const (
 // Pebble version.
 const FormatMinSupported = FormatFlushableIngest
 
+// FormatMinForSharedObjects it the minimum format version that supports shared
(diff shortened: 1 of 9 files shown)


itsbilal (approved):
LGTM, thanks for doing this!

RaduBerinde:
TFTR!

PR #3180, opened 2023-12-21
sstable: add back support for older formats

RaduBerinde:
Add back support for older sstable formats. These are still not supported by any non-deprecated Pebble format, but they can be used if the `sstable` package is used directly. CockroachDB uses it to read data from backups, which could be old.

This will fix a bunch of "bad magic number" test failures which use older backup fixtures.


dt (approved):

RaduBerinde:
TFTR!

PR #3184, opened 2023-12-26
*: avoid variable name conflict with imported package name 

mattisonchao:
This is a cleanup to avoid name conflict with imported package names.


xxmplus:
Close to due to age

PR #3185, opened 2023-12-26
*: Ignore JetBrains IDEA local configurations.

mattisonchao:
Ignore JetBrains IDEA local configurations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ mutex.prof
 coverprofile.out
 # Testing artifacts
 meta.*.test
+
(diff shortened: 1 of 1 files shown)


nicktrav (approved):

RaduBerinde:
You should use the global gitignore for this instead:
```
# cat ~/.gitignore_global 
.idea
```

RaduBerinde:
Well, this is all about the developer's environment and their choice of IDE. I'm not opposed to making the change, but the reasonable thing for any GoLand user is to set up their `.gitignore_global` instead of expecting all repos they work with to have it.

PR #3186, opened 2023-12-27
keyspan: add logging infrastructure for iterator stacks

RaduBerinde:
#### keyspan: add logging infrastructure for iterator stacks

This change adds a logging `FragmentIterator` and new infrastructure
that allows injecting logging iterators in an entire iterator stack.
This helps understand how and why the stack produces a certain result.

We do this by adding a `FragmentIterator.WrapChildren()` method which
uses a given wrap function to replace any children iterators. The
wrap function itself can call `WrapChildren` again to recursively do
the same to all iterators in the stack.

Some iterators that create new child iterators on the fly have to
remember the wrap function.

Sample output: https://gist.github.com/RaduBerinde/0200d3c4726966f24c1ef81ed02e440e

#### keyspan: use treeprinter for iterator stack logging

The treeprinter library was lifted from cockroach without changes.

Sample output: https://gist.github.com/RaduBerinde/cc744c9922868740169cd670bfdb8b2a

### DIFF
--- a/error_iter.go
+++ b/error_iter.go
@@ -75,12 +75,13 @@ type errorKeyspanIter struct {
 // errorKeyspanIter implements the keyspan.FragmentIterator interface.
 var _ keyspan.FragmentIterator = (*errorKeyspanIter)(nil)
 
-func (*errorKeyspanIter) SeekGE(key []byte) *keyspan.Span { return nil }
-func (*errorKeyspanIter) SeekLT(key []byte) *keyspan.Span { return nil }
-func (*errorKeyspanIter) First() *keyspan.Span            { return nil }
-func (*errorKeyspanIter) Last() *keyspan.Span             { return nil }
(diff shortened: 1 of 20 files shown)


jbowens:
Very cool. I wonder if we should make it a probe (once we’ve pulled the keyspan iter probes into the itertest package)

RaduBerinde:
Where can I find out more about iter probes?

RaduBerinde:
I think this would be useful to merge. We can integrate it with probes later.

jbowens (approved):
Sorry for the delay, :lgtm:

> Reviewed 15 of 18 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3186#-:-NnoBlEJ8q00ov4Rw0kV:bg1jfsl)* status: 19 of 20 files reviewed, all discussions resolved

PR #3187, opened 2023-12-27
sstable: hide obsolete range keys for shared ingested sstables

### LINKED ISSUE #3174: metamorphic test failure: inconsistent keys in range

itsbilal:
This change updates the virtual sstable range key iterator to
hide obsolete range keys for shared ingested files. It does
this by reusing the ForeignSSTTransformer that was removed
in https://github.com/cockroachdb/pebble/pull/2782.

This change also cleans up some dangling references to the older
`isForeign` way of determining whether to hide obsolete keys,
in lieu of the newer isSharedIngested approach to address
boomerang shared files.

Fixes https://github.com/cockroachdb/pebble/issues/3174.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -1172,7 +1172,7 @@ func runSSTablePropertiesCmd(t *testing.T, td *datadriven.TestData, d *DB) strin
 	var v sstable.VirtualReader
 	props := r.Properties.String()
 	if m != nil && m.Virtual {
-		v = sstable.MakeVirtualReader(r, m.VirtualMeta(), false /* isForeign */)
+		v = sstable.MakeVirtualReader(r, m.VirtualMeta(), false /* isShared */)
 		props = v.Properties.String()
 	}
 	if len(td.Input) == 0 {
--- a/event_listener_test.go
+++ b/event_listener_test.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"fmt"
 	"reflect"
+	"runtime"
 	"strings"
 	"sync"
 	"testing"
@@ -25,6 +26,9 @@ import (
 
 // Verify event listener actions, as well as expected filesystem operations.
 func TestEventListener(t *testing.T) {
+	if runtime.GOARCH == "386" {
+		t.Skip("skipped on 32-bit due to slightly varied output")
+	}
 	var d *DB
 	var memLog base.InMemLogger
 	mem := vfs.NewMem()
--- a/ingest_test.go
+++ b/ingest_test.go
@@ -12,6 +12,7 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"runtime"
 	"slices"
 	"sort"
 	"strconv"
@@ -2116,6 +2117,9 @@ func TestIngest(t *testing.T) {
 	var mem vfs.FS
 	var d *DB
 	var flushed bool
+	if runtime.GOARCH == "386" {
+		t.Skip("skipped on 32-bit due to slightly varied output")
+	}
 	defer func() {
 		require.NoError(t, d.Close())
 	}()
--- a/internal/keyspan/transformer.go
+++ b/internal/keyspan/transformer.go
@@ -48,3 +48,102 @@ func VisibleTransform(snapshot uint64) Transformer {
 		return nil
 	})
 }
+
+// TransformerIter is a FragmentIterator that applies a Transformer on all
+// returned keys. Used for when a caller needs to apply a transformer on an
+// iterator but does not otherwise need the mergingiter's merging ability.
+type TransformerIter struct {
+	FragmentIterator
+
+	// Transformer is applied on every Span returned by this iterator.
+	Transformer Transformer
+	// Comparer in use for this keyspace.
+	Compare base.Compare
+
+	span Span
+	err  error
+}
+
+func (t *TransformerIter) applyTransform(span *Span) *Span {
+	t.span = Span{
+		Start: t.span.Start[:0],
+		End:   t.span.End[:0],
+		Keys:  t.span.Keys[:0],
+	}
+	if err := t.Transformer.Transform(t.Compare, *span, &t.span); err != nil {
+		t.err = err
+		return nil
+	}
+	return &t.span
+}
+
+// SeekGE implements the FragmentIterator interface.
+func (t *TransformerIter) SeekGE(key []byte) *Span {
+	span := t.FragmentIterator.SeekGE(key)
+	if span == nil {
+		return nil
+	}
+	return t.applyTransform(span)
+}
+
+// SeekLT implements the FragmentIterator interface.
+func (t *TransformerIter) SeekLT(key []byte) *Span {
+	span := t.FragmentIterator.SeekLT(key)
+	if span == nil {
+		return nil
+	}
+	return t.applyTransform(span)
+}
+
+// First implements the FragmentIterator interface.
+func (t *TransformerIter) First() *Span {
+	span := t.FragmentIterator.First()
+	if span == nil {
+		return nil
+	}
+	return t.applyTransform(span)
+}
+
+// Last implements the FragmentIterator interface.
+func (t *TransformerIter) Last() *Span {
+	span := t.FragmentIterator.Last()
+	if span == nil {
+		return nil
+	}
+	return t.applyTransform(span)
+}
+
+// Next implements the FragmentIterator interface.
+func (t *TransformerIter) Next() *Span {
+	span := t.FragmentIterator.Next()
+	if span == nil {
+		return nil
+	}
+	return t.applyTransform(span)
+}
+
+// Prev implements the FragmentIterator interface.
+func (t *TransformerIter) Prev() *Span {
+	span := t.FragmentIterator.Prev()
+	if span == nil {
+		return nil
+	}
+	return t.applyTransform(span)
+}
+
+// Error implements the FragmentIterator interface.
+func (t *TransformerIter) Error() error {
+	if t.err != nil {
+		return t.err
+	}
+	return t.FragmentIterator.Error()
+}
+
+// Close implements the FragmentIterator interface.
+func (t *TransformerIter) Close() error {
+	err := t.FragmentIterator.Close()
+	if err != nil {
+		return err
+	}
+	return t.err
+}
--- a/internal/rangekey/coalesce.go
+++ b/internal/rangekey/coalesce.go
@@ -6,7 +6,6 @@ package rangekey
 
 import (
 	"bytes"
-	"fmt"
 	"math"
 	"sort"
 
@@ -381,16 +380,17 @@ func coalesce(
 }
 
 // ForeignSSTTransformer implements a keyspan.Transformer for range keys in
-// foreign sstables (i.e. shared sstables not created by us). It is largely
-// similar to the Transform function implemented in UserIteratorConfig in that
-// it calls coalesce to remove range keys shadowed by other range keys, but also
-// retains the range key that does the shadowing. In addition, it elides
-// RangeKey unsets/dels in L6 as they are inapplicable when reading from a
-// different Pebble instance.
+// shared ingested sstables. It is largely similar to the Transform function
+// implemented in UserIteratorConfig in that it calls coalesce to remove range
+// keys shadowed by other range keys, but also retains the range key that does
+// the shadowing. In addition, it elides RangeKey unsets/dels in L6 as they are
+// inapplicable when reading from a different Pebble instance. Finally, it
+// returns keys sorted in trailer order, not suffix order, as that's what the
+// rest of the iterator stack expects.
 type ForeignSSTTransformer struct {
-	Comparer *base.Comparer
-	Level    int
-	sortBuf  keyspan.KeysBySuffix
+	Equal   base.Equal
+	SeqNum  uint64
+	sortBuf keyspan.KeysBySuffix
 }
 
 // Transform implements the Transformer interface.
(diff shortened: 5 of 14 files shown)


RaduBerinde:
Thanks for doing this so quickly!

Is TestMetaTwoInstance stable now if you stress it?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3187)* status: 0 of 9 files reviewed, 5 unresolved discussions (waiting on @itsbilal)

___
*[`sstable/reader.go` line 426 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3187#-Nmhva7oBthUQJDrj2Lx:-Nmhva7oBthUQJDrj2Ly:b-k1cd7x) ([raw file](https://github.com/cockroachdb/pebble/blob/228d31f2c53ae57c4e17ff839ee722eeb058cfdb/sstable/reader.go#L426)):*
> ```Go
> 	}
> 	i := &fragmentBlockIter{elideSameSeqnum: true}
> 	if err := i.blockIter.initHandle(r.Compare, h, r.Properties.GlobalSeqNum, false); err != nil {
> ```

Will we have the same problem with RANGEDELs? I guess not, because after rewriting the sequence numbers, any RANGEDELs won't apply to the points in the same table. Maybe a comment here (explaining the `false`) would be useful though.
___
*[`sstable/reader.go` line 432 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3187#-NmhuevE8FPbh8z4Heqn:-NmhuevE8FPbh8z4Heqo:b-7nwd8z) ([raw file](https://github.com/cockroachdb/pebble/blob/228d31f2c53ae57c4e17ff839ee722eeb058cfdb/sstable/reader.go#L432)):*
> ```Go
> }
> 
> func (r *Reader) newRawRangeKeyIter(vState *virtualState) (keyspan.FragmentIterator, error) {
> ```

[nit] Why not just a `hideObsoletePoints` flag?
___
*[`sstable/writer.go` line 1265 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3187#-NmhyMN44ddHPgDvrRxm:-NmhyMN44ddHPgDvrRxn:b-7n37tn) ([raw file](https://github.com/cockroachdb/pebble/blob/228d31f2c53ae57c4e17ff839ee722eeb058cfdb/sstable/writer.go#L1265)):*
> ```Go
> }
> 
> func (w *Writer) addRangeKey(key InternalKey, value []byte) error {
> ```

[nit] add a comment here explaining that this can only be called with fragmented range keys
___
*[`sstable/writer.go` line 1302 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3187#-Nmhznzj15iLamAYp8by:-Nmhznzj15iLamAYp8bz:b9vu9gk) ([raw file](https://github.com/cockroachdb/pebble/blob/228d31f2c53ae57c4e17ff839ee722eeb058cfdb/sstable/writer.go#L1302)):*
> ```Go
> 				return w.err
> 			}
> 			// There are two cases in which the current internal key is obsolete.
> ```

It's a bit unexpected that `disableKeyOrderChecks` also disables this code.. It looks like it's a testing-only flag though. Maybe rename to `testingDisableKeyOrderChecks`
___
*[`sstable/writer.go` line 1303 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3187#-NmhynG0D7-MqtbBZNUm:-NmhynG0D7-MqtbBZNUn:bce0tj7) ([raw file](https://github.com/cockroachdb/pebble/blob/228d31f2c53ae57c4e17ff839ee722eeb058cfdb/sstable/writer.go#L1303)):*
> ```Go
> 			}
> 			// There are two cases in which the current internal key is obsolete.
> 			// Either we've already written a RangeKeyDelete for this span (w.lastSpanDeleted)
> ```

It is important here that InternalKeyKindRangeKeyDelete is smaller than the others right? It implies that if we saw one already, it must have been at a strictly higher seq num.

itsbilal (commented):
TFTR!

@RaduBerinde yep! Ran it for ~15 mins and couldn't see a single failure.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3187#-:-NmlDj_V6y3txaPnM_hI:b-7dzvf7)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`sstable/reader.go` line 426 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3187#-Nmhva7oBthUQJDrj2Lx:-NmlDml_4IJCBlM0gdUc:bucs578) ([raw file](https://github.com/cockroachdb/pebble/blob/228d31f2c53ae57c4e17ff839ee722eeb058cfdb/sstable/reader.go#L426)):*

Will we have the same problem with RANGEDELs? I guess not, because after rewriting the sequence numbers, any RANGEDELs won't apply to the points in the same table. Maybe a comment here (explaining the `false`) would be useful though.
</blockquote></details>

That's right, we won't have this issue with rangedels. Added the comment. The other is that currently rangedels won't have the bit set anyway.
___
*[`sstable/reader.go` line 432 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3187#-NmhuevE8FPbh8z4Heqn:-NmlFBGg9LI3JHwkUyx2:bsn0xw0) ([raw file](https://github.com/cockroachdb/pebble/blob/228d31f2c53ae57c4e17ff839ee722eeb058cfdb/sstable/reader.go#L432)):*

[nit] Why not just a `hideObsoletePoints` flag?
</blockquote></details>

Just mimics the same behaviour as some of the other new*Iter* methods on Reader that take a vState pointer.
___
*[`sstable/writer.go` line 1265 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3187#-NmhyMN44ddHPgDvrRxm:-NmlFaJT1XOz06u52H55:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/228d31f2c53ae57c4e17ff839ee722eeb058cfdb/sstable/writer.go#L1265)):*

[nit] add a comment here explaining that this can only be called with fragmented range keys
</blockquote></details>

Done.
___
*[`sstable/writer.go` line 1302 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3187#-Nmhznzj15iLamAYp8by:-NmlFay16Eyp6omQ4Y-C:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/228d31f2c53ae57c4e17ff839ee722eeb058cfdb/sstable/writer.go#L1302)):*

It's a bit unexpected that `disableKeyOrderChecks` also disables this code.. It looks like it's a testing-only flag though. Maybe rename to `testingDisableKeyOrderChecks`
</blockquote></details>

Done.
___
*[`sstable/writer.go` line 1303 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3187#-NmhynG0D7-MqtbBZNUm:-NmlFj0HAktO22vY3cky:b-x207gg) ([raw file](https://github.com/cockroachdb/pebble/blob/228d31f2c53ae57c4e17ff839ee722eeb058cfdb/sstable/writer.go#L1303)):*

It is important here that InternalKeyKindRangeKeyDelete is smaller than the others right? It implies that if we saw one already, it must have been at a strictly higher seq num.
</blockquote></details>

I'm not sure if I understand the question, but if we have a RangeKeySet that has a higher seqnum than a RangeKeyDelete, we will still consider the RangeKeyDelete to be non-obsolete because it applies to all prefixes and could obsolete future RangeKeySets/Unsets. Now if we have two RangeKeyDeletes over the same span then yes the latter is obsolete.

RaduBerinde:
What if we have a Set and an Unset at different sequence numbers, don't we need to resolve those?  Eg Unset(foo)@5 Set(foo,bar)@4 needs to become Unset(foo)@X Set(bar)@X

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3187)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`sstable/writer.go` line 1303 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3187#-NmhynG0D7-MqtbBZNUm:-NmlkJO4B27NyTM5nh55:bagsxbt) ([raw file](https://github.com/cockroachdb/pebble/blob/228d31f2c53ae57c4e17ff839ee722eeb058cfdb/sstable/writer.go#L1303)):*

I'm not sure if I understand the question, but if we have a RangeKeySet that has a higher seqnum than a RangeKeyDelete, we will still consider the RangeKeyDelete to be non-obsolete because it applies to all prefixes and could obsolete future RangeKeySets/Unsets. Now if we have two RangeKeyDeletes over the same span then yes the latter is obsolete.
</blockquote></details>

I mean if we have a Set and Delete at the same sequence number.. we rely on Delete kind being smaller so thay the Set "wins", right?

itsbilal (commented):
> What if we have a Set and an Unset at different sequence numbers, don't we need to resolve those? Eg Unset(foo)@5 Set(foo,bar)@4

That will present itself as Unset(foo)@5, Set(foo)@4, Set(bar)@4, and only the Set(foo)@4 will have its obsolete key kind set. But you bring up an important point, deletes on a span are easy to track as we already get the sequence number ordering and can use the current flag, but for each Set we might need to find the relevant prefix' Unset before it as it could be non-contiguous. I'll try to take a stab at that.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3187#-:-NmmPWSH81Ni1kYubmxh:b-iswec7)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`sstable/writer.go` line 1303 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3187#-NmhynG0D7-MqtbBZNUm:-NmmP96R05keIqKbLjhz:b-rfaujk) ([raw file](https://github.com/cockroachdb/pebble/blob/228d31f2c53ae57c4e17ff839ee722eeb058cfdb/sstable/writer.go#L1303)):*

I mean if we have a Set and Delete at the same sequence number.. we rely on Delete kind being smaller so thay the Set "wins", right?
</blockquote></details>

In that case both would be non-obsolete because the Delete could delete other prefixes underneath us. And yes the user iterator config won't delete the Set if it is at the same sequence number as the Delete.

RaduBerinde:
So there is something missing here w.r.t Set/Unset. I wonder why TestMetaTwoInstance doesn't catch that.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3187)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`sstable/writer.go` line 1266 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3187#-Nn9_tVgDR9PdeTmL_1Z:-Nn9_tVgDR9PdeTmL_1_:bjb6jrt) ([raw file](https://github.com/cockroachdb/pebble/blob/82dc4ef0420f814d7293b5c8a9819de90ece897a/sstable/writer.go#L1266)):*
> ```Go
> 
> // NB: Just like AddRangeKey(), this can only be called with fragmented range
> // keys.
> ```

Just a drive-by since I haven't really read the code in this PR.

My original motivation to do the obsolete key stuff for point keys was that it was most performance sensitive. The thought then was that for RANGEDELs and range keys, we could afford to pay the cost of iterating using the original sequence number, squash the obsolete ones, and then assign the sst sequence number to the non-obsolete ones (using some kind of wrapper iterator). So we would go with whatever was simplest from a code perspective (only iterator changes OR write obsolete bit and use them in the iterator). You all should decide what is simplest.

RaduBerinde:
*[`sstable/writer.go` line 1266 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3187#-Nn9_tVgDR9PdeTmL_1Z:-Nn9oecA7KtOKMSx6lvK:b1d0xbm) ([raw file](https://github.com/cockroachdb/pebble/blob/82dc4ef0420f814d7293b5c8a9819de90ece897a/sstable/writer.go#L1266)):*

Just a drive-by since I haven't really read the code in this PR.

My original motivation to do the obsolete key stuff for point keys was that it was most performance sensitive. The thought then was that for RANGEDELs and range keys, we could afford to pay the cost of iterating using the original sequence number, squash the obsolete ones, and then assign the sst sequence number to the non-obsolete ones (using some kind of wrapper iterator). So we would go with whatever was simplest from a code perspective (only iterator changes OR write obsolete bit and use them in the iterator). You all should decide what is simplest.
</blockquote></details>

I have a slight preference for doing it at read time - it is simpler conceptually, it has the relevant code in a single place and it's easier to update / fix bugs. I think it would be very efficient either way because we already have all the keys organized nto Spans

itsbilal:
*[`sstable/writer.go` line 1266 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3187#-Nn9_tVgDR9PdeTmL_1Z:-NnA63Ee9cS_YJ8zHpyH:b-5vj182) ([raw file](https://github.com/cockroachdb/pebble/blob/82dc4ef0420f814d7293b5c8a9819de90ece897a/sstable/writer.go#L1266)):*

I don't think we need a wrapper iterator, it would be part of the existing code that rewrites seqnums
</blockquote></details>

We actually did have a `ForeignSSTTransformer` at some point that did exactly that, and was implemented as a merging iter transform. We just convinced ourselves that we didn't need it when we ripped it out. I don't mind adding it back in, as it seems better to do that than to have some relatively fragile write-time logic. I'll make that change next; thanks for y'alls thoughts on this!

itsbilal (commented):
TFTRs! I've gone back to using the `coalesce` logic for reading range keys from shared ingested sstables, and brought back ForeignSSTTransformer which was removed in #2782. The first commit is a blanket revert of that PR, and the second one has additional fixes to bring it "up to the times" + integrate it with the table cache. I've also improved unit testing in TestIngestShared.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3187#-:-NnAnTKy6Ja57TuXskYB:b8lxboc)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`sstable/writer.go` line 1266 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3187#-Nn9_tVgDR9PdeTmL_1Z:-NnAnk_gD1ZDzoBkDg-0:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/82dc4ef0420f814d7293b5c8a9819de90ece897a/sstable/writer.go#L1266)):*

We actually did have a `ForeignSSTTransformer` at some point that did exactly that, and was implemented as a merging iter transform. We just convinced ourselves that we didn't need it when we ripped it out. I don't mind adding it back in, as it seems better to do that than to have some relatively fragile write-time logic. I'll make that change next; thanks for y'alls thoughts on this!
</blockquote></details>

Done.

jbowens (approved):
:lgtm: 

> Reviewed 2 of 15 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3187#-:-NnFwz4UE1kw_u6MOpiQ:bg7igkv)* status: 2 of 17 files reviewed, 2 unresolved discussions (waiting on @itsbilal, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/reader_virtual.go` line 190 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3187#-NnFwXI48ciThW0_bHYw:-NnFwXI597u3vw3bX61d:bqc3kkt) ([raw file](https://github.com/cockroachdb/pebble/blob/ab534cb803592a325bafe4fe71ba19f3413f16d8/sstable/reader_virtual.go#L190)):*
> ```Go
> 			Transformer:      transform,
> 			Compare:          v.reader.Compare,
> 		}
> ```

nit: Add a TODO to avoid these two allocations by hoisting `ForeignSSTTransformer` and `TransformerIter` onto the `VirtualReader` struct.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3187#-:-NnFxsDwB8Xa-qST2Y7f:b-814g8k)* status: 2 of 17 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`sstable/reader_virtual.go` line 190 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3187#-NnFwXI48ciThW0_bHYw:-NnFxtCDCX3gJWtOsfD5:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/ab534cb803592a325bafe4fe71ba19f3413f16d8/sstable/reader_virtual.go#L190)):*

nit: Add a TODO to avoid these two allocations by hoisting `ForeignSSTTransformer` and `TransformerIter` onto the `VirtualReader` struct.
</blockquote></details>

Done.

PR #3188, opened 2023-12-28
sstable: flush value blocks if 128 are buffered

dt:


sumeerbhola (changes requested):
This will break `EstimateDiskUsage` which assumes that the data blocks are contiguous, and interpolates the size contribution of the value blocks. There is a TODO to improve the situation but it will need a format change https://github.com/cockroachdb/pebble/blob/ccb96e586962015d136e33a8f7cb844078628505/sstable/reader.go#L1106-L1137

Is https://github.com/cockroachdb/cockroach/pull/117122 an adequate work around in the absence of online restore? For backups in online restore, is there any benefit to making file sizes smaller so that they often don't span range boundaries (I realize we will virtualize and can share files across ranges)? With online restore, we could benefit from value blocks until the file is materialized locally, but we could also live without that optimization.

> Reviewed 1 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3188#-:-NnKwmdwBgr6VnNl-C5O:b4u7w1r)* status: 1 of 6 files reviewed, 2 unresolved discussions (waiting on @dt)

___
*[`sstable/options.go` line 277 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3188#-NnKeYawAKRUc4mrO8JT:-NnKeYawAKRUc4mrO8JU:b-5xrwqw) ([raw file](https://github.com/cockroachdb/pebble/blob/593fc13ff7c5aa8a6fed58afeb859641ae73e03b/sstable/options.go#L277)):*
> ```Go
> 	// presence of prefetching/read-ahead, page caching, etc.
> 	//
> 	// A value of 0 implies the default of max(8MB/BlockSize, 16) while a value of
> ```

I am curious about the 8MB default. Versus something like 64MB, which would allow most Pebble written files to be unaffected by this change.
___
*[`sstable/options.go` line 282 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3188#-NnKiCQu-B976KRxPzQD:-NnKiCQv-q2h29Xzo1GB:b38t36e) ([raw file](https://github.com/cockroachdb/pebble/blob/593fc13ff7c5aa8a6fed58afeb859641ae73e03b/sstable/options.go#L282)):*
> ```Go
> }
> 
> func (o WriterOptions) ensureDefaults() WriterOptions {
> ```

`sstable.WriterOptions` are usually generated using `pebble.MakeWriterOptions` (including in CRDB's sst_writer.go), so we need to add this `ValueBlockBufferLimit` to the `pebble.Options` so that it is actually possible to configure it.

PR #3189, opened 2023-12-29
sstable: fix PrefixReplacingIterator.SetBounds memory aliasing

dt:
Previously SetBounds was passing the underlying iterator bounds that would mutate on the next call to any of the methods of the iterator, causing the unexpected value in the test case that is updated here as well.


itsbilal (approved):
LGTM

dt:
NB: this only applies to SetBounds calls that do _not_ get handed though as-is... which are none outside the unit tests since normally the underlying iterator will be passing them to vState so we don't mess with them. So this is just to make the test happy.

PR #3193, opened 2024-01-03
db: add Options.DisableTableStats

jbowens:
Export the previously private disableTableStats field to allow external users access. This is intended for use within external tests that require additional determinism.


RaduBerinde (approved):

jbowens:
TFTR!

PR #3197, opened 2024-01-04
db: avoid mutex-protected I/O in Ingest and NewEventuallyFileOnlySnapshot

jbowens:
Creation of an eventually file-only snapshot requires inspecting the flushable queue to look for overlap with the key ranges of the new snapshot. This is performed while holding the DB mutex to avoid racing with other goroutines that may mutate the flushable queue. Previously, if ingested sstables were queued as flushable ingests, checking for overlap could perform read I/O while holding the database mutex.

Similarly, ingestion requires determining which ingested sstables overlap any flushables. This is used for correctness (to ensure we sequence the sstable higher in the LSM than any overlapping flushables) and for ingest statistics that we return to the caller. Previously, if ingested sstables were already in the queue as flushable ingests, checking for overlap could perform read I/O while holding both the database mutex and the commit pipeline mutex.

This commit refactors these overlap checks to use a new computePossibleOverlaps method of the flushable interface. This allows the `ingestedFlushable` implementation to avoid I/O and determine overlap using simple file boundary comparisons. This required some gymnastics to support the IngestAndExcise case that requires knowledge of which files overlapped flushables and must detect overlaps with both ingested sstables and an excise span.

### DIFF
--- a/db.go
+++ b/db.go
@@ -1578,19 +1578,12 @@ func (d *DB) NewEventuallyFileOnlySnapshot(keyRanges []KeyRange) *EventuallyFile
 	if err := d.closed.Load(); err != nil {
 		panic(err)
 	}
-
-	internalKeyRanges := make([]internalKeyRange, len(keyRanges))
 	for i := range keyRanges {
 		if i > 0 && d.cmp(keyRanges[i-1].End, keyRanges[i].Start) > 0 {
 			panic("pebble: key ranges for eventually-file-only-snapshot not in order")
 		}
-		internalKeyRanges[i] = internalKeyRange{
-			smallest: base.MakeInternalKey(keyRanges[i].Start, InternalKeySeqNumMax, InternalKeyKindMax),
-			largest:  base.MakeExclusiveSentinelKey(InternalKeyKindRangeDelete, keyRanges[i].End),
-		}
 	}
-
-	return d.makeEventuallyFileOnlySnapshot(keyRanges, internalKeyRanges)
+	return d.makeEventuallyFileOnlySnapshot(keyRanges)
 }
 
 // Close closes the DB.
@@ -1748,25 +1741,17 @@ func (d *DB) Compact(start, end []byte, parallelize bool) error {
 		return errors.Errorf("Compact start %s is not less than end %s",
 			d.opts.Comparer.FormatKey(start), d.opts.Comparer.FormatKey(end))
 	}
-	iStart := base.MakeInternalKey(start, InternalKeySeqNumMax, InternalKeyKindMax)
-	iEnd := base.MakeInternalKey(end, 0, 0)
-	m := (&fileMetadata{}).ExtendPointKeyBounds(d.cmp, iStart, iEnd)
-	meta := []*fileMetadata{m}
 
 	d.mu.Lock()
 	maxLevelWithFiles := 1
 	cur := d.mu.versions.currentVersion()
 	for level := 0; level < numLevels; level++ {
-		overlaps := cur.Overlaps(level, d.cmp, start, end, iEnd.IsExclusiveSentinel())
+		overlaps := cur.Overlaps(level, d.cmp, start, end, false)
 		if !overlaps.Empty() {
 			maxLevelWithFiles = level + 1
 		}
 	}
 
-	keyRanges := make([]internalKeyRange, len(meta))
-	for i := range meta {
-		keyRanges[i] = internalKeyRange{smallest: m.Smallest, largest: m.Largest}
-	}
 	// Determine if any memtable overlaps with the compaction range. We wait for
 	// any such overlap to flush (initiating a flush if necessary).
 	mem, err := func() (*flushableEntry, error) {
@@ -1776,25 +1761,31 @@ func (d *DB) Compact(start, end []byte, parallelize bool) error {
 		// overlaps.
 		for i := len(d.mu.mem.queue) - 1; i >= 0; i-- {
 			mem := d.mu.mem.queue[i]
-			if ingestMemtableOverlaps(d.cmp, mem, keyRanges) {
-				var err error
-				if mem.flushable == d.mu.mem.mutable {
-					// We have to hold both commitPipeline.mu and DB.mu when calling
-					// makeRoomForWrite(). Lock order requirements elsewhere force us to
-					// unlock DB.mu in order to grab commitPipeline.mu first.
-					d.mu.Unlock()
-					d.commit.mu.Lock()
-					d.mu.Lock()
-					defer d.commit.mu.Unlock()
-					if mem.flushable == d.mu.mem.mutable {
-						// Only flush if the active memtable is unchanged.
-						err = d.makeRoomForWrite(nil)
-					}
-				}
-				mem.flushForced = true
-				d.maybeScheduleFlush()
-				return mem, err
+			var anyOverlaps bool
+			mem.computePossibleOverlaps(func(b bounded) shouldContinue {
+				anyOverlaps = true
+				return stopIteration
+			}, KeyRange{Start: start, End: end})
+			if !anyOverlaps {
+				continue
 			}
+			var err error
+			if mem.flushable == d.mu.mem.mutable {
+				// We have to hold both commitPipeline.mu and DB.mu when calling
+				// makeRoomForWrite(). Lock order requirements elsewhere force us to
+				// unlock DB.mu in order to grab commitPipeline.mu first.
+				d.mu.Unlock()
+				d.commit.mu.Lock()
+				d.mu.Lock()
+				defer d.commit.mu.Unlock()
+				if mem.flushable == d.mu.mem.mutable {
+					// Only flush if the active memtable is unchanged.
+					err = d.makeRoomForWrite(nil)
+				}
+			}
+			mem.flushForced = true
+			d.maybeScheduleFlush()
+			return mem, err
 		}
 		return nil, nil
 	}()
@@ -1811,7 +1802,7 @@ func (d *DB) Compact(start, end []byte, parallelize bool) error {
 	for level := 0; level < maxLevelWithFiles; {
 		for {
 			if err := d.manualCompact(
-				iStart.UserKey, iEnd.UserKey, level, parallelize); err != nil {
+				start, end, level, parallelize); err != nil {
 				if errors.Is(err, ErrCancelledCompaction) {
 					continue
 				}
--- a/batch.go
+++ b/batch.go
@@ -2005,6 +2005,13 @@ func (b *flushableBatch) readyForFlush() bool {
 	return true
 }
 
+// computePossibleOverlaps is part of the flushable interface.
+func (b *flushableBatch) computePossibleOverlaps(
+	fn func(bounded) shouldContinue, bounded ...bounded,
+) {
+	computePossibleOverlapsGenericImpl[*flushableBatch](b, b.cmp, fn, bounded)
+}
+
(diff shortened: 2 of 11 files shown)


jbowens (commented):
I'm thinking a bit about how to test.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3197#-:-NnLR7KxD0cuNobUJTlp:b5hnwme)* status: 0 of 9 files reviewed, all discussions resolved (waiting on @sumeerbhola)

db.go
```diff
@@ -1751,7 +1744,6 @@ func (d *DB) Compact(start, end []byte, parallelize bool) error {
 	iStart := base.MakeInternalKey(start, InternalKeySeqNumMax, InternalKeyKindMax)
 	iEnd := base.MakeInternalKey(end, 0, 0)
 	m := (&fileMetadata{}).ExtendPointKeyBounds(d.cmp, iStart, iEnd)
```
itsbilal, line 1746: (not relevant to this PR so feel free to ignore) this seems like an abuse of a FileMetadata. Maybe we should use `KeyRange` instead?

itsbilal (approved):
LGTM, thanks for doing this!

sumeerbhola (changes requested):
very minor comments only

> Reviewed 4 of 9 files at r1, 3 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3197#-:-NnjI-NFADV_5KXh6KZB:b-vtqe2g)* status: 7 of 9 files reviewed, 7 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`flushable.go` line 45 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3197#-NnjC4hwEcDNgjF1XMX9:-NnjC4hx1za8BwaUf17s:bigl984) ([raw file](https://github.com/cockroachdb/pebble/blob/9fb56fe7b40439e956e04a65b36728f278c4515f/flushable.go#L45)):*
> ```Go
> 
> type bounded interface {
> 	InternalKeyBounds() (InternalKey, InternalKey)
> ```

can you add a comment that these are both inclusive.
___
*[`flushable.go` line 303 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3197#-NnjCNcz3pSXLnanQW2v:-NnjCNcz3pSXLnanQW2w:b-b4l0dk) ([raw file](https://github.com/cockroachdb/pebble/blob/9fb56fe7b40439e956e04a65b36728f278c4515f/flushable.go#L303)):*
> ```Go
> 
> // computePossibleOverlapsGenericImpl is an implemention of the flushable
> // interface's computePossibleOverlaps function for flushable implementations
> ```

nit: perhaps qualify as "... flushable implementations with only in-memory state that do not ..."
___
*[`flushable.go` line 319 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3197#-NnjCdl-3a-1-Q6vL7jn:-NnjCdl-3a-1-Q6vL7jo:bxajmos) ([raw file](https://github.com/cockroachdb/pebble/blob/9fb56fe7b40439e956e04a65b36728f278c4515f/flushable.go#L319)):*
> ```Go
> 		kr := internalKeyRange{s, l}
> 		if overlapWithIterator(iter, &rangeDelIter, rkeyIter, kr, cmp) {
> 			fn(b)
> ```

its a bit odd that this interface does not tell the callee whether the caller has ensured that there is overlap (like this case), or that this could be a false positive. I suppose that is because the current callees never try to eliminate false positives? 
___
*[`ingest.go` line 1352 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3197#-NnjDRjB81YxkJ6LYMC7:-NnjDRjB81YxkJ6LYMC8:b-39nmfr) ([raw file](https://github.com/cockroachdb/pebble/blob/9fb56fe7b40439e956e04a65b36728f278c4515f/ingest.go#L1352)):*
> ```Go
> 	}
> 
> 	// metaFlushableOverlaps is a slice parallel to meta indicating which of the
> ```

stale comment (no longer a slice)
___
*[`ingest.go` line 1396 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3197#-NnjFE8oALedDFk_MI8g:-NnjFE8oALedDFk_MI8h:bgebixm) ([raw file](https://github.com/cockroachdb/pebble/blob/9fb56fe7b40439e956e04a65b36728f278c4515f/ingest.go#L1396)):*
> ```Go
> 				switch v := b.(type) {
> 				case *fileMetadata:
> 					metaFlushableOverlaps[v.FileNum] = true
> ```

so are we accepting some false positives?
And is this considered ok since overlap between an ingested flushable (caused by addsstable or range snapshot) and a new ingest will likely be a true positive (in the CRDB context)?
___
*[`snapshot.go` line 287 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3197#-NnjH0Jr1oHI6vJiq7tx:-NnjH0Jr1oHI6vJiq7ty:b-4rkd77) ([raw file](https://github.com/cockroachdb/pebble/blob/9fb56fe7b40439e956e04a65b36728f278c4515f/snapshot.go#L287)):*
> ```Go
> 	for i := range d.mu.mem.queue {
> 		d.mu.mem.queue[i].computePossibleOverlaps(func(bounded) {
> 			isFileOnly = false
> ```

should there be a bool return value from this func to avoid continuing to check the remaining key ranges?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3197)* status: 1 of 9 files reviewed, 6 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`db.go` line 1746 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3197#gh-1443391898:-NnjPJNy8PAeL80vqEee:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/2caaeec9ef223f39a93defc199552689bab2ec08/db.go#L1746)):*

(not relevant to this PR so feel free to ignore) this seems like an abuse of a FileMetadata. Maybe we should use `KeyRange` instead?
</blockquote></details>

Done.
___
*[`flushable.go` line 45 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3197#-NnjC4hwEcDNgjF1XMX9:-NnjQGyf-nVSZM9UuuW1:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9fb56fe7b40439e956e04a65b36728f278c4515f/flushable.go#L45)):*

can you add a comment that these are both inclusive.
</blockquote></details>

Done.
___
*[`flushable.go` line 303 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3197#-NnjCNcz3pSXLnanQW2v:-NnjQS7LCQ99k8CeQ1CQ:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9fb56fe7b40439e956e04a65b36728f278c4515f/flushable.go#L303)):*

nit: perhaps qualify as "... flushable implementations with only in-memory state that do not ..."
</blockquote></details>

Done.
___
*[`flushable.go` line 319 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3197#-NnjCdl-3a-1-Q6vL7jn:-NnjQUL96AOH7F211KhS:bm7e97f) ([raw file](https://github.com/cockroachdb/pebble/blob/9fb56fe7b40439e956e04a65b36728f278c4515f/flushable.go#L319)):*

its a bit odd that this interface does not tell the callee whether the caller has ensured that there is overlap (like this case), or that this could be a false positive. I suppose that is because the current callees never try to eliminate false positives?
</blockquote></details>

Yeah, that's right
___
*[`ingest.go` line 1352 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3197#-NnjDRjB81YxkJ6LYMC7:-NnjQdn-14-vDQbnB8rZ:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9fb56fe7b40439e956e04a65b36728f278c4515f/ingest.go#L1352)):*

stale comment (no longer a slice)
</blockquote></details>

Done.
___
*[`ingest.go` line 1396 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3197#-NnjFE8oALedDFk_MI8g:-NnjQhNr1QUgfBk7g2Z6:bamg8p8) ([raw file](https://github.com/cockroachdb/pebble/blob/9fb56fe7b40439e956e04a65b36728f278c4515f/ingest.go#L1396)):*

so are we accepting some false positives?
And is this considered ok since overlap between an ingested flushable (caused by addsstable or range snapshot) and a new ingest will likely be a true positive (in the CRDB context)?
</blockquote></details>

Yeah, that's right—also even if it were a true false positive, it's unclear which is more disruptive: blocking all commits to determine whether there's overlap, or the cost of handling this ingestion as a flushable ingest. Added some of this in a comment here.
___
*[`snapshot.go` line 287 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3197#-NnjH0Jr1oHI6vJiq7tx:-NnjUxHv8oqupoC5ENn6:b-6ohhur) ([raw file](https://github.com/cockroachdb/pebble/blob/9fb56fe7b40439e956e04a65b36728f278c4515f/snapshot.go#L287)):*

should there be a bool return value from this func to avoid continuing to check the remaining key ranges?
</blockquote></details>

added a `shouldContinue` enum return value to control it

sumeerbhola (changes requested):
> Reviewed 1 of 9 files at r1, 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3197)* status: all files reviewed, 7 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`flushable.go` line 48 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3197#-Nnjh0kB8XbiYOb3iidV:-Nnjh0kC4oBUjZwe4vlo:b-ncyd94) ([raw file](https://github.com/cockroachdb/pebble/blob/065169ee333d1c733f49b56aa350da6814217366/flushable.go#L48)):*
> ```Go
> const (
> 	continueIteration shouldContinue = true
> 	stopIteration
> ```

https://go.dev/play/p/NA16uxyR-ug claims this will also be true. Even if correct, relying on some default here is confusing (at least to me). If incorrect, I wonder why this didn't trip up a unit test.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3197)* status: 8 of 11 files reviewed, 7 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`flushable.go` line 48 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3197#-Nnjh0kB8XbiYOb3iidV:-NnkD8CgEpUd1WMsdEK-:boep15o) ([raw file](https://github.com/cockroachdb/pebble/blob/065169ee333d1c733f49b56aa350da6814217366/flushable.go#L48)):*

https://go.dev/play/p/NA16uxyR-ug claims this will also be true. Even if correct, relying on some default here is confusing (at least to me). If incorrect, I wonder why this didn't trip up a unit test.
</blockquote></details>

Oof, good catch. Yeah, this was a bug. It wasn't caught by any unit tests because continuing ignoring a `stopIteration` value is okay in all use cases, just less efficient. I fixed this and added a memtable unit test that could've caught it.

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3197#-:-NnkLrqp0hz4DryqZCj7:bg7igkv)* status: all files reviewed, 7 unresolved discussions (waiting on @itsbilal)

jbowens:
TFTRs!

PR #3200, opened 2024-01-05
db: test iterator error while switching to forward direction

jbowens:
This commit adds datadriven test cases for an iterator error encountered while
switching from backward to forward iteration. It also fixes a case within
NextPrefix in which an `invariants` build would panic, improperly assuming
iterKey is non-nil (the only case in which it may be nil is if an error
occurred).

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -1368,6 +1368,16 @@ func parseDBOptionsArgs(opts *Options, args []datadriven.CmdArg) error {
 			for i := range opts.Levels {
 				opts.Levels[i].BlockSize = v
 			}
+		case "cache-size":
+			if opts.Cache != nil {
+				opts.Cache.Unref()
+				opts.Cache = nil
+			}
+			size, err := strconv.ParseInt(cmdArg.Vals[0], 10, 64)
+			if err != nil {
+				return err
+			}
+			opts.Cache = NewCache(size)
 		case "index-block-size":
 			v, err := strconv.Atoi(cmdArg.Vals[0])
 			if err != nil {
--- a/iterator.go
+++ b/iterator.go
@@ -1772,6 +1772,16 @@ func (i *Iterator) nextPrefix() IterValidityState {
 			// key with the user key less than i.key, so we're guaranteed to
 			// land on the correct key with a single Next.
 			i.iterKey, i.iterValue = i.iter.Next()
+			if i.iterKey == nil {
+				// This should only be possible if i.iter.Next() encountered an
+				// error.
+				if i.iter.Error() == nil {
+					i.opts.logger.Fatalf("pebble: invariant violation: Nexting internal iterator from iterPosPrev found nothing")
+				}
+				// NB: Iterator.Error() will return i.iter.Error().
+				i.iterValidityState = IterExhausted
+				return i.iterValidityState
+			}
 			if invariants.Enabled && !i.equal(i.iterKey.UserKey, i.key) {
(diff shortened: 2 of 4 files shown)


RaduBerinde (approved):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3200)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`iterator.go` line 1777 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3200#-Nne3BDi8uZHrTp2Z-hY:-Nne3BDi8uZHrTp2Z-hZ:b-nzwg7d) ([raw file](https://github.com/cockroachdb/pebble/blob/9343416736ee506d40695602c4a0516ebfd79bf8/iterator.go#L1777)):*
> ```Go
> 			if i.iterKey == nil {
> 				// This should only be possible if i.iter.Next() encountered an
> 				// error.
> ```

Don't we want to propagate `i.iter.Error()` in this case? Or at least assert that it's not nil?

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3200#-:-NnekgviAAvGCWCj8no7:b-814g8k)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`iterator.go` line 1777 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3200#-Nne3BDi8uZHrTp2Z-hY:-NnejyZuDpOQuzLT1uQM:b-8uqbdo) ([raw file](https://github.com/cockroachdb/pebble/blob/9343416736ee506d40695602c4a0516ebfd79bf8/iterator.go#L1777)):*

Don't we want to propagate `i.iter.Error()` in this case? Or at least assert that it's not nil?
</blockquote></details>

The `Iterator.Error()` implementation will return `i.iter.Error()` if it's non-nil (like expected in this case). Added a comment. Good idea to add an assertion.

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3200#-:-NnelMS542ezn8kl7XxU:bnfp4nl)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`iterator.go` line 1778 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3200#-Nnel9TM9HuOjfsef8mW:-Nnel9TM9HuOjfsef8mX:bt84w9l) ([raw file](https://github.com/cockroachdb/pebble/blob/cf99b48688ac22cdfceefca9f7148b405c30860e/iterator.go#L1778)):*
> ```Go
> 				// This should only be possible if i.iter.Next() encountered an
> 				// error.
> 				if err := i.iter.Error(); err == nil {
> ```

super nit but the `err` variable is not useful
___
*[`iterator.go` line 1781 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3200#-NnelE447zfjGZgfvIQp:-NnelE447zfjGZgfvIQq:b-wbfdgj) ([raw file](https://github.com/cockroachdb/pebble/blob/cf99b48688ac22cdfceefca9f7148b405c30860e/iterator.go#L1781)):*
> ```Go
> 					i.opts.logger.Fatalf("pebble: invariant violation: Nexting internal iterator from iterPosPrev found nothing")
> 				}
> 				// NB: Iterator.Error() will return i.iter.Error() if it's
> ```

[nit] I'd remove "if it's non-nil"

PR #3206, opened 2024-01-08
metamorphic: add diagrams, key simplification, more useful messages

RaduBerinde:
This commit makes a few improvements to the metamorphic test:
 - as part of `--try-to-reduce`, we now also try to simplify keys to
   single characters; we also set a timeout and treat timeouts as
   "bad" op sequences.
 - we add support for diagrams that show the relevant keys or key
   ranges for each operation; `--try-to-reduce` generates diagrams
   when they would be of reasonable size.
 - we improve error messages to show the `go test` command line that
   can be used to reduce the ops; `--try-to-reduce` shows the
   `go test` command line that can be used to re-run with the final
   ops.

### DIFF
--- a/internal/metamorphic/reduce_test.go
+++ b/internal/metamorphic/reduce_test.go
@@ -16,6 +16,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/cockroachdb/pebble/metamorphic"
 	"github.com/stretchr/testify/require"
 )
 
@@ -137,6 +138,7 @@ func (r *reducer) try(t *testing.T, ops []string) bool {
 	args := []string{
 		"-test.run", t.Name() + "$",
 		"-test.v",
+		"-test.timeout", "10s",
 		"--keep",
 	}
 
@@ -155,12 +157,13 @@ func (r *reducer) try(t *testing.T, ops []string) bool {
 	cmd.Stderr = &output
 	cmd.Stdout = &output
 	err := cmd.Run()
-	// If the test succeeds or fails with an internal test error, we removed
-	// important ops.
+	// If the test succeeds or fails with an internal test error or a timeout, we
+	// removed important ops.
 	if err == nil ||
 		strings.Contains(output.String(), "metamorphic test internal error") ||
 		strings.Contains(output.String(), "leaked iterators") ||
-		strings.Contains(output.String(), "leaked snapshots") {
+		strings.Contains(output.String(), "leaked snapshots") ||
+		strings.Contains(output.String(), "test timed out") {
 		require.NoError(t, os.RemoveAll(testRootDir))
 		return false
 	}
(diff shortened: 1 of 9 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3206)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`metamorphic/simplify.go` line 39 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3206#-NndozkoB641ckcEWdkL:-NndozkoB641ckcEWdkM:b-h13ps4) ([raw file](https://github.com/cockroachdb/pebble/blob/830e1fc7d84a528e9dca173d6a65a728f2e942e4/metamorphic/simplify.go#L39)):*
> ```Go
> 	}
> 	return []byte(formatOps(ops)), nil
> }
> ```

To maintain semantic equivalence, I think we would need to decompose the key into prefix and suffix and simplify them separately. A `foo@5` that is converted to a simple `b` will no longer by shadowed by an iterator with range key masking and a covering range key with suffix `@6`
___
*[`metamorphic/testdata/diagram` line 31 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3206#-Nndoupq99YQqqs3Hvwx:-Nndoupq99YQqqs3Hvwy:b-jdo7t9) ([raw file](https://github.com/cockroachdb/pebble/blob/830e1fc7d84a528e9dca173d6a65a728f2e942e4/metamorphic/testdata/diagram#L31)):*
> ```
>                |-----------------------------|                  db2.Compact("cranberry", "pear", true /* parallelize */)
> |------|-------|----------|-------|----------|-----|----------|
> apple  banana  cranberry  durian  guanabana  pear  pineapple  raspberry
> ```

Very cool!

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3206)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`metamorphic/simplify.go` line 39 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3206#-NndozkoB641ckcEWdkL:-Nne0dWH9Uj3v3_EQiPq:bycjjsa) ([raw file](https://github.com/cockroachdb/pebble/blob/830e1fc7d84a528e9dca173d6a65a728f2e942e4/metamorphic/simplify.go#L39)):*

To maintain semantic equivalence, I think we would need to decompose the key into prefix and suffix and simplify them separately. A `foo@5` that is converted to a simple `b` will no longer by shadowed by an iterator with range key masking and a covering range key with suffix `@6`
</blockquote></details>

True, this does not guarantee semantic equivalence (added a comment to clarify that). The result is just one more thing to try to see if it leads to a reproduction.

Even if we decomposed, some keys sharing a prefix might be important; we'd always need to check that the problem still reproduces.

I added a TODO with your suggestion. Ideally we should try the big simplification here first, and if the problem no longer reproduces we should try to simplify just the prefixes.

jbowens (approved):
:lgtm:

> Reviewed 6 of 9 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3206#-:-Nnf0GlW6RsQJ3csa3Wb:bnfp4nl)* status: 7 of 9 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`metamorphic/simplify.go` line 39 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3206#-NndozkoB641ckcEWdkL:-Nnf-rMDCFEbdNujpm58:b-hfghm4) ([raw file](https://github.com/cockroachdb/pebble/blob/830e1fc7d84a528e9dca173d6a65a728f2e942e4/metamorphic/simplify.go#L39)):*

True, this does not guarantee semantic equivalence (added a comment to clarify that). The result is just one more thing to try to see if it leads to a reproduction.

Even if we decomposed, some keys sharing a prefix might be important; we'd always need to check that the problem still reproduces.

I added a TODO with your suggestion. Ideally we should try the big simplification here first, and if the problem no longer reproduces we should try to simplify just the prefixes.
</blockquote></details>

Sounds good
___
*[`metamorphic/simplify_test.go` line 1 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3206#-Nnf0D-n5srr1RuoiDVm:-Nnf0D-n5srr1RuoiDVn:b-liuim6) ([raw file](https://github.com/cockroachdb/pebble/blob/dc1c181dcc59034d934c81a63f7021a45d7191bc/metamorphic/simplify_test.go#L1)):*
> ```Go
> // Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
> ```

happy new year

RaduBerinde:
TFTR! Happy new year!

PR #3208, opened 2024-01-08
vfs: fix hard link names in MemFS.String()

pav-kv:
This PR fixes a bug in `MemFS.String()` method. For hard links created using the `Link()` method, `String()` previously printed the original file name rather than the name of the link.

### DIFF
--- a/vfs/mem_fs.go
+++ b/vfs/mem_fs.go
@@ -110,7 +110,7 @@ func (y *MemFS) String() string {
 	defer y.mu.Unlock()
 
 	s := new(bytes.Buffer)
-	y.root.dump(s, 0)
+	y.root.dump(s, 0, sep)
 	return s.String()
 }
 
@@ -209,9 +209,10 @@ func (y *MemFS) Create(fullname string) (File, error) {
 			if frag == "" {
 				return errors.New("pebble/vfs: empty file name")
 			}
-			n := &memNode{name: frag}
+			n := &memNode{}
 			dir.children[frag] = n
 			ret = &memFile{
+				name:  frag,
 				n:     n,
 				fs:    y,
 				read:  true,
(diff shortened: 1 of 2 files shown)


pav-kv:
Commit-by-commit review is recommended.

vfs/mem_fs.go
```diff
@@ -657,22 +652,24 @@ func (f *memNode) resetToSyncedState() {
...
 }
 
+var _ os.FileInfo = (*memFile)(nil)
 var _ File = (*memFile)(nil)
 
 func (f *memFile) Close() error {
 	if n := f.n.refs.Add(-1); n < 0 {
 		panic(fmt.Sprintf("pebble: close of unopened file: %d", n))
 	}
-	f.n = nil
```
pav-kv, line 675: Needed to remove this line because `memFile.Stat()` must return `FileInfo` that can outlive the `memFile` after its `Close()`. Without this line `TestCheckpoint` panics.

vfs/mem_fs_test.go
```diff
@@ -227,12 +227,17 @@ func TestList(t *testing.T) {
...
 
+	require.NoError(t, fs.Link("/foo/3", "/bar/link-to-3"))
+	require.NoError(t, fs.Link("/foo/3", "/bar/another-link-to-3"))
+
 	{
 		got := fs.String()
 		const want = `          /
        0    a
             bar/
+       0      another-link-to-3
```
pav-kv, line 238: We could also print some "node ID" so that there is a visual way to see that certain files are actually hard linked to the same node.

vfs/mem_fs.go
pav-kv: I have a PR out for this.

pav-kv:
@jbowens PTAL

vfs/mem_fs.go
```diff
@@ -657,8 +625,9 @@ func (f *memNode) resetToSyncedState() {
 	}
 }
 
-// memFile is a reader or writer of a node's data, and implements File.
+// memFile is a reader or writer of a node's data. Implements File.
 type memFile struct {
+	name        string
```
pav-kv, line 630: We might also need a follow-up to remove this field from `memFile`, and "query" it from `MemFS` instead, when needed. The file name can be changed concurrently by `Rename()`, so this name can be stale.

pav-kv: Nah, we don't need to do that. The "canonical" implementation of this interface in Go packages [returns](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/os/stat_unix.go;l=24) the cached file name that was used to open it.

RaduBerinde (approved):

PR #2146, opened 2022-11-20
internal/keyspan: modify FragmentIterator seek semantics

jbowens:
Alter the semantics of SeekGE and SeekLT on FragmentIterator. Previously, FragmentIterator's seek operations were defined only in terms of span start keys. This commit changes the seek operations to be defined in terms of the keys contained by the span. A SeekGE now seeks to the first span containing a key ≥ the seek key, and a SeekLT now seeks to the last span containing a key < seek key. These new semantics match the typical top-level iterator use.

With these new semantics, SeekLT can still be implemented in terms of a simple span start key seek. Seeking to the last span containing a key < seek key is equivalent to seeking to the last span with a start key less the given key.

However, SeekGE implementations now require an extra key comparison and sometimes a Next. Since top-level iterator requires the containment semantics anyways, this key comparison and Next is only being moved down the stack into the interface implementation. When using the keyspan merging iterator, the keyspan.MergingIter's SeekGE implementation performs a SeekLT per-level, which suffers no additional overhead.

With the MergingIter and DefragmentingIter implementations, these new semantics reduce the amount of work performed during a seek. The previous iterator stack's SeekGE looked like (left-to-right, top-down):

```
                         InterleavingIter.SeekGE
                                    │
                   ╭────────────────┴───────────────╮
                   │                                │
        DefragmentingIter.SeekLT         DefragmentingIter.Next()
                   │                                │
       ╭───────────┴───╮                            │
       │               │                            │
 MergingIter.SeekLT    ├── defragmentFwd            ├── defragmentFwd
       │               │                            │
       │               ╰── defragmentBwd            ╰── defragmentFwd
       ╰───────────╮
                   │
       ╭───────────┴───────────╮
       │                       │
 MergingIter.SeekGE      MergingIter.Prev
       │
       ╰─╶╶ per level╶╶ ─╮
                         │
             ╭───────────┴───────────╮
             │                       │
         <?>.SeekLT              <?>.Next
```

The new iterator stack's SeekGE, assuming it doesn't hit the new defragmenting fast path, looks like:
```
                         InterleavingIter.SeekGE
                                    │
                         DefragmentingIter.SeekGE
                                    │
                   ╭────────────────┴───────────────╮
                   │                                ├── defragmentBwd*
             MergingIter.SeekGE                     │
                   │                                ╰── defragmentFwd
                   ╰─╶╶ per level╶╶ ─╮
                                     │
                                     │
                                     ├── <?>.SeekLT
                                     │
[...]

PR #3209, opened 2024-01-09
internal/keyspan: simplify Get and fix error handling

jbowens:
The keyspan.Get helper is used by `DB.Get` to retrieve range deletions overlapping a particular key. It was written before #2146 when a `SeekGE(k)` seeked the iterator to the first span with a _start key_ ≥ k. In #2146, the semantics of `SeekGE(k)` was modified to seek to the first span with an end key greater than k (or, more intuitively, the first span to "cover" a key ≥ k). With the new semantics, a simple `SeekGE` seeks to the span covering `k` if any, and `keyspan.Get` only needs to verify that the returned span doesn't lie wholly after `k`.

Despite being written against the old iterator interface, the previous keyspan.Get implementation remained correct in the absence of iterator errors after #2146. However, if an iterator error occurred during the SeekLT, keyspan.Get would improperly step the iterator forward without consulting `FragmentIterator.Error()`.

### DIFF
--- a/get_iter.go
+++ b/get_iter.go
@@ -83,8 +83,9 @@ func (g *getIter) Next() (*InternalKey, base.LazyValue) {
 			// key. Every call to levelIter.Next() potentially switches to a new
 			// table and thus reinitializes rangeDelIter.
 			if g.rangeDelIter != nil {
-				g.tombstone = keyspan.Get(g.comparer.Compare, g.rangeDelIter, g.key)
-				if g.err = g.rangeDelIter.Close(); g.err != nil {
+				g.tombstone, g.err = keyspan.Get(g.comparer.Compare, g.rangeDelIter, g.key)
+				g.err = firstError(g.err, g.rangeDelIter.Close())
+				if g.err != nil {
 					return nil, base.LazyValue{}
 				}
 				g.rangeDelIter = nil
--- a/internal/keyspan/fragmenter_test.go
+++ b/internal/keyspan/fragmenter_test.go
@@ -131,7 +131,8 @@ func TestFragmenter(t *testing.T) {
 	// number. This is a simple version of what full processing of range
 	// tombstones looks like.
 	deleted := func(key []byte, seq, readSeq uint64) bool {
-		s := Get(cmp, iter, key)
+		s, err := Get(cmp, iter, key)
+		require.NoError(t, err)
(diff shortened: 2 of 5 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3209#-:-NnoccGe8Iwst22pHa3m:bnfp4nl)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi, @jbowens, and @sumeerbhola)

___
*[`internal/keyspan/get.go` line 20 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3209#-NnocM4tBl8s1Ekoo4kG:-NnocM4tBl8s1Ekoo4kH:b-vs8ipr) ([raw file](https://github.com/cockroachdb/pebble/blob/4b398734c797c0653f77103f202ec0cd96ac206c/internal/keyspan/get.go#L20)):*
> ```Go
> 	// seeking to the first span with an end key greater than the given key.
> 	iterSpan := iter.SeekGE(key)
> 	if iterSpan == nil {
> ```

[nit] might be nicer as a switch
```
switch {
  case iterSpan == nil: return nil, iter.Error()
  case cmp(iterSpan.Start, key) > 0: return nil, nil
  default: return iterSpan, nil
}
```

jbowens:
TFTR!

sumeerbhola (commented):
> Reviewed 3 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3209)* status: 4 of 5 files reviewed, 2 unresolved discussions

___
*[`internal/keyspan/get.go` line 15 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3209#-NnpWohI9JFiMKxu_cjU:-NnpWohI9JFiMKxu_cjV:b-dvo50y) ([raw file](https://github.com/cockroachdb/pebble/blob/eda4b7ac0cec74f3b6b61a15eefa51139cac07f9/internal/keyspan/get.go#L15)):*
> ```Go
> // If an error occurs while seeking iter, a nil span and non-nil error is
> // returned.
> func Get(cmp base.Compare, iter FragmentIterator, key []byte) (*Span, error) {
> ```

love that this PR both simplifies/reduces the code and fixes a bug!

PR #3213, opened 2024-01-10
internal/keyspan: simplify SeekLE, fix nil-pointer

jbowens:
Similar to #3209, this commit simplifies SeekLE for the new (as of #2146) FragmentIterator semantics. It also fixes a nil-pointer access if iter.Error()!= nil after Next-ing the iterator in the second else block.

### DIFF
--- a/internal/keyspan/iter.go
+++ b/internal/keyspan/iter.go
@@ -17,6 +17,11 @@ import (
 // positioning method. Some implementations (eg, keyspan.Iter) may provide
 // longer lifetimes but implementations need only guarantee stability until the
 // next positioning method.
+//
+// If any positioning method fails to find a span, the iterator is left
+// positioned at an exhausted position in the direction of iteration. For
+// example, a caller than finds SeekGE(k)=nil may call Prev to move the iterator
+// to the last span.
 type FragmentIterator interface {
 	// SeekGE moves the iterator to the first span covering a key greater than
 	// or equal to the given key. This is equivalent to seeking to the first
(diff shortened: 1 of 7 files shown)


jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3213#-:-NnpHO8lCduqA_yTVT6T:b-814g8k)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/keyspan/seek.go` line 19 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3213#-NnoZNJXCTN9PqsAVDmn:-NnpGPd509tY4a_Mc8Ia:b321l1d) ([raw file](https://github.com/cockroachdb/pebble/blob/c70e1e677e3a96d07949162cadc3477d3f9fccea/internal/keyspan/seek.go#L19)):*

Is SeekGE guaranteed to leave the iterator "pointing after the end" in this case? If yes we should clarify in `FragmentIterator.SeekGE` interface comments.
In any case, it would be more clear to just do `Last()` here instead.
</blockquote></details>

Yeah it is. I'll add this to the top-level `FragmentIterator` comment. `Last` is less efficient for some iterators, because it requires re-examining indexing data structures or in some cases re-filtering.
___
*[`internal/keyspan/seek.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3213#-NnoZuBTEd9O5twXikLd:-NnpHN7J7RnzzihagkeC:b-896fiz) ([raw file](https://github.com/cockroachdb/pebble/blob/c70e1e677e3a96d07949162cadc3477d3f9fccea/internal/keyspan/seek.go#L28)):*

[nit] worth commenting that this is not necessarily an error case, `iter.Error()` can be nil here
</blockquote></details>

Done,

jbowens:
TFTR!

PR #3214, opened 2024-01-10
db,metamorphic: add TestIteratorErrors

jbowens:
Add a new randomized test that ensures errors injected within iterator
operations are surfaced to end users. The test uses the metamorphic test
framework to first generate a random database. It again uses the metamorphic
test framework to run random read-only operations against the database,
injecting errors randomly to VFS operations. The test fails if an error was
injected during the course of an operation, but the operation did not surface
an error.

Informs #1115.

### DIFF
--- a/external_test.go
+++ b/external_test.go
@@ -1,0 +1,112 @@
+// Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package pebble_test
+
+import (
+	"bytes"
+	"io"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/cockroachdb/pebble/metamorphic"
+	"github.com/cockroachdb/pebble/vfs"
+	"github.com/cockroachdb/pebble/vfs/errorfs"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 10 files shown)


external_test.go
```diff
@@ -0,0 +1,112 @@
...
+// TestIteratorErrors is a randomized test designed to ensure that errors
+// encountered by reads are properly propagated through to the user. It uses the
+// metamorphic tests configured with only write operations to first generate a
+// random database. It then uses the metamorphic tests to run a random set of
+// read operations against the generated database, randomly injecting errors at
+// the VFS layer. If an error is injected over the course of an operation, it
+// expects the error to surface to the operation output. If it doesn't, the test
+// fails.
+func TestIteratorErrors(t *testing.T) {
+	seed := time.Now().UnixNano()
```
itsbilal, line 30: We should probably throw a `-seed` command line arg in this package too, just like `internal/metamorphic`.

itsbilal (approved):
Last commit LGTM

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3214#-:-NnueNGX9D0G0iZHXE7b:b-814g8k)* status: 0 of 21 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

PR #3216, opened 2024-01-10
db: allow value block writing to be disabled

sumeerbhola:
Previously there was no way to disable value blocks in table formats that supported value blocks. Now, it can be disabled using WriterOptions.DisableValueBlocks. This is never used by Pebble since we expect small sstables where the memory overhead of buffering value blocks in the sstable.Writer is acceptable.

Informs https://github.com/cockroachdb/cockroach/issues/117113

### DIFF
--- a/metamorphic/ops.go
+++ b/metamorphic/ops.go
@@ -652,10 +652,11 @@ func buildForIngest(
 
 	equal := t.opts.Comparer.Equal
 	tableFormat := db.FormatMajorVersion().MaxTableFormat()
-	w := sstable.NewWriter(
-		objstorageprovider.NewFileWritable(f),
-		t.opts.MakeWriterOptions(0, tableFormat),
-	)
+	wOpts := t.opts.MakeWriterOptions(0, tableFormat)
+	if t.testOpts.disableValueBlocksForIngestSSTables {
+		wOpts.DisableValueBlocks = true
+	}
+	w := sstable.NewWriter(objstorageprovider.NewFileWritable(f), wOpts)
 
 	var lastUserKey []byte
 	for key, value := iter.First(); key != nil; key, value = iter.Next() {
--- a/metamorphic/options.go
+++ b/metamorphic/options.go
@@ -95,6 +95,9 @@ func parseOptions(
 				opts.enableValueBlocks = true
 				opts.Opts.Experimental.EnableValueBlocks = func() bool { return true }
 				return true
+			case "TestOptions.disable_value_blocks_for_ingest_sstables":
+				opts.disableValueBlocksForIngestSSTables = true
+				return true
 			case "TestOptions.async_apply_to_db":
 				opts.asyncApplyToDB = true
 				return true
@@ -187,6 +190,9 @@ func optionsToString(opts *TestOptions) string {
 	}
 	if opts.enableValueBlocks {
 		fmt.Fprintf(&buf, "  enable_value_blocks=%t\n", opts.enableValueBlocks)
+	}
+	if opts.disableValueBlocksForIngestSSTables {
+		fmt.Fprintf(&buf, "  disable_value_blocks_for_ingest_sstables=%t\n", opts.disableValueBlocksForIngestSSTables)
 	}
 	if opts.asyncApplyToDB {
 		fmt.Fprint(&buf, "  async_apply_to_db=true\n")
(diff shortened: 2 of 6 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3216)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/options.go` line 269 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3216#-Nnqh7Pv2WtDkUKo4DIQ:-Nnqh7Pv2WtDkUKo4DIR:b-me41uk) ([raw file](https://github.com/cockroachdb/pebble/blob/4fd3ad596104db52a34290dd56d570059b7b951f/sstable/options.go#L269)):*
> ```Go
> 
> // SetDisableValueBlocks sets the value of WriterOptions.DisableValueBlocks.
> func (o WriterOptions) SetDisableValueBlocks(disableValueBlocks bool) WriterOptions {
> ```

[nit] This would be better as `DisableValueBlocks()`, as it's unlikely we'll ever need to call it with `false` in practice (since that is the default). Not sure the method is useful in the first place since it's just setting a public member.
___
*[`sstable/writer.go` line 211 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3216#-NnqiPQhBOxn81UzHKkn:-NnqiPQhBOxn81UzHKko:b-9k4gd8) ([raw file](https://github.com/cockroachdb/pebble/blob/4fd3ad596104db52a34290dd56d570059b7b951f/sstable/writer.go#L211)):*
> ```Go
> 	requiredInPlaceValueBound UserKeyPrefixBound
> 	valueBlockWriter          *valueBlockWriter
> 	disableValueBlocks        bool
> ```

[nit] Do we need this? Can't we just check `valueBlockWriter == nil`?
___
*[`sstable/writer.go` line 924 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3216#-NnqhbtO18KpQP7PR0Wz:-NnqhbtO18KpQP7PR0X-:bk297kj) ([raw file](https://github.com/cockroachdb/pebble/blob/4fd3ad596104db52a34290dd56d570059b7b951f/sstable/writer.go#L924)):*
> ```Go
> 	// optimization.
> 	const tinyValueThreshold = 0
> 	if considerWriteToValueBlock && (valueLen <= tinyValueThreshold || w.disableValueBlocks) {
> ```

[nit] would be more clear to just do `considerWriteToValueBlock = setHasSamePrefix && !w.disableValueBlocks setHasSamePrefix && valueLen > tinyValueThreshold`

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3216#-:-NnywrQg28xeQNlZPplT:b-814g8k)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/options.go` line 269 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3216#-Nnqh7Pv2WtDkUKo4DIQ:-NnyqwgBDH2RtEZjEuo4:ba3dfob) ([raw file](https://github.com/cockroachdb/pebble/blob/4fd3ad596104db52a34290dd56d570059b7b951f/sstable/options.go#L269)):*

[nit] This would be better as `DisableValueBlocks()`, as it's unlikely we'll ever need to call it with `false` in practice (since that is the default). Not sure the method is useful in the first place since it's just setting a public member.
</blockquote></details>

Removed
___
*[`sstable/writer.go` line 211 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3216#-NnqiPQhBOxn81UzHKkn:-NnyrrDu2Xe_a-V_sqIM:ba3dfob) ([raw file](https://github.com/cockroachdb/pebble/blob/4fd3ad596104db52a34290dd56d570059b7b951f/sstable/writer.go#L211)):*

[nit] Do we need this? Can't we just check `valueBlockWriter == nil`?
</blockquote></details>

Removed
___
*[`sstable/writer.go` line 924 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3216#-NnqhbtO18KpQP7PR0Wz:-NnyuSswAXfvQ0ojs72N:b-hidjjv) ([raw file](https://github.com/cockroachdb/pebble/blob/4fd3ad596104db52a34290dd56d570059b7b951f/sstable/writer.go#L924)):*

[nit] would be more clear to just do `considerWriteToValueBlock = setHasSamePrefix && !w.disableValueBlocks setHasSamePrefix && valueLen > tinyValueThreshold`
</blockquote></details>

Much cleaner! Done.

RaduBerinde (approved):
:lgtm:

> Reviewed 2 of 6 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3216#-:-NoHisIj4qc_sSJH73Yp:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jbowens)

PR #3217, opened 2024-01-10
db: fix build

RaduBerinde:
Fix a small merge skew.


RaduBerinde:
Ah, Jackson beat me to it. #3215

PR #3219, opened 2024-01-11
db: use InterleavingIter to interleave range deletions in compactions

### LINKED ISSUE #3082: db: refactor/cleanup compaction iterator

jbowens:
Remove the keyspan.InternalIteratorShim that was intended to be a temporary bridge to allow range deletions to be surfaced within the compaction iterator, replacing it with the keyspan.InterleavingIter. This mirrors the mechanics of range keys. Follow up work will be able to simplify the compaction iterator logic now that range deletions are always interleaved at the maximal sequence number.

Additionally, this commit removes the compaction iterator's concept of a key that lies within the same snapshot stripe but is non-skippable. These "non-skippable" keys lead to bizarre behavior, such as a sequence like:

a.MERGE.5 a.RANGEDEL.4 a.MERGE.4

yielding the separate unmerged a.MERGE.5 and a.MERGE.4 internal keys (Note a.RANGEDEL.4 does not delete a.MERGE.4 because they share the same sequence number.)

The sameStripeNonSkippable fell out of the fact that range deletions were previously interleaved by the input iterator at their start key with their sequence number. These sameStripeNonSkippable could interrupt any logic iterating across the keys of a stripe, complicating the compaction iterator logic.

In some instances INVALID keys were also returned with sameStripeNonSkippable. These keys are now treated similarly to other errors encountered during iteration: i.err is set and newStripeNewKey is returned.

Close #3082.


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3219)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`compaction.go` line 1383 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3219#-NntY-xF2VtV8LhcoD6A:-NntY-xF2VtV8LhcoD6B:b2nh76y) ([raw file](https://github.com/cockroachdb/pebble/blob/31ca00d09d1d6042f11c7ae6ad062bfa439b751a/compaction.go#L1383)):*
> 1

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3219)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`compaction.go` line 1383 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3219#-NntY-xF2VtV8LhcoD6A:-NntqGRq7kx1ZsiVoh6v:bediaiy) ([raw file](https://github.com/cockroachdb/pebble/blob/31ca00d09d1d6042f11c7ae6ad062bfa439b751a/compaction.go#L1383)):*

> 1
</blockquote></details>

Good catch; looks like a long latent bug
___
*[`testdata/manual_compaction_set_with_del_sstable_Pebblev4` line 705 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3219#-NntPOYcCM8hrlAWjH1L:-NntPOYcCM8hrlAWjH1M:b-oevql8) ([raw file](https://github.com/cockroachdb/pebble/blob/31ca00d09d1d6042f11c7ae6ad062bfa439b751a/testdata/manual_compaction_set_with_del_sstable_Pebblev4#L705)):*
> ```
> 2:
>   000011:[a#4,RANGEDEL-d#inf,RANGEDEL]
>   000012:[k#5,RANGEDEL-q#inf,RANGEDEL]
> ```

this `m` -> `q` diff is a consequence of now using the defragmenting iterator. Previously, the compaction iterator would see the individual fragmented spans from each sstable in L1 and could elide the m-q rangedel. Now it sees one broader

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3219)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`testdata/manual_compaction_set_with_del_sstable_Pebblev4` line 705 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3219#-NntPOYcCM8hrlAWjH1L:-NntuA8K3kcq8FDe6Oor:b-c2t88x) ([raw file](https://github.com/cockroachdb/pebble/blob/31ca00d09d1d6042f11c7ae6ad062bfa439b751a/testdata/manual_compaction_set_with_del_sstable_Pebblev4#L705)):*

this `m` -> `q` diff is a consequence of now using the defragmenting iterator. Previously, the compaction iterator would see the individual fragmented spans from each sstable in L1 and could elide the m-q rangedel. Now it sees one broader
</blockquote></details>

To keep the diff minimal, I removed the defragmenting iterator, and we can consider defragmenting range deletions in a separate patch.

sumeerbhola (changes requested):
> Reviewed 7 of 10 files at r1, 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3219)* status: all files reviewed (commit messages unreviewed), 7 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`compaction.go` line 630 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3219#-No-ERkv39tq5nZ_hy56:-No-ERkv39tq5nZ_hy57:b-8nm00v) ([raw file](https://github.com/cockroachdb/pebble/blob/c20b86e5b37c4f03b02ec625ce6fd98cf070823b/compaction.go#L630)):*
> ```Go
> 	// The range deletion tombstone iterator, that merges and fragments
> 	// tombstones across levels. This iterator is included within the compaction
> 	// input iterator as a single level.
> ```

this comment seems stale now that the point keys are also interleaved by this.
___
*[`compaction.go` line 631 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3219#-No-Ebav8O2aeDbV-AsI:-No-Ebav8O2aeDbV-AsJ:b-ictimj) ([raw file](https://github.com/cockroachdb/pebble/blob/c20b86e5b37c4f03b02ec625ce6fd98cf070823b/compaction.go#L631)):*
> ```Go
> 	// tombstones across levels. This iterator is included within the compaction
> 	// input iterator as a single level.
> 	rangeDelInterleaving keyspan.InterleavingIter
> ```

meta question about this change: how much additional risk are we incurring by using `InterleavingIter`, given that we cannot afford any bug in our compaction iteration code. I am assuming the answer is very low especially since we use only `First()` and `Next()`, but do we need some additional randomized testing of `InterleavingIter` to give us more confidence?
___
*[`testdata/compaction_delete_only_hints` line 224 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3219#-No-FLo23vtz0xxWAmkF:-No-FLo23vtz0xxWAmkG:b-7fjf73) ([raw file](https://github.com/cockroachdb/pebble/blob/9be83bd239f8f9d3cbae807a8bfe3536ec7a2ee8/testdata/compaction_delete_only_hints#L224)):*
> ```
> ----
> 6:
>   000004:[a#20,SETWITHDEL-z#inf,RANGEDEL]
> ```

is this diff is because we don't get interrupted by `sameStripeNonSkippable` after seeing the `SET`?
___
*[`testdata/compaction_delete_only_hints` line 258 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3219#-No-Fe-26mvP3jVWMmiW:-No-Fe-26mvP3jVWMmiX:b-rlp9nx) ([raw file](https://github.com/cockroachdb/pebble/blob/9be83bd239f8f9d3cbae807a8bfe3536ec7a2ee8/testdata/compaction_delete_only_hints#L258)):*
> ```
> 10
> ----
> [JOB 100] compacted(elision-only) L6 [000004] (741B) Score=0.00 + L6 [] (0B) Score=0.00 -> L6 [000005] (662B), in 1.0s (2.0s total), output rate 662B/s
> ```

why did the byte count change?
___
*[`testdata/range_del` line 1196 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3219#-No-FpZX1MXePqH8qiRs:-No-FpZX1MXePqH8qiRt:b-20sg36) ([raw file](https://github.com/cockroachdb/pebble/blob/9be83bd239f8f9d3cbae807a8bfe3536ec7a2ee8/testdata/range_del#L1196)):*
> ```
> ----
> 2:
>   000013:[a#11,SETWITHDEL-c#inf,RANGEDEL]
> ```

why did the file boundaries change?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3219)* status: 7 of 11 files reviewed, 7 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`compaction.go` line 631 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3219#-No-Ebav8O2aeDbV-AsI:-NoIRgWq0bkmePfC3sfc:bo68vvd) ([raw file](https://github.com/cockroachdb/pebble/blob/c20b86e5b37c4f03b02ec625ce6fd98cf070823b/compaction.go#L631)):*

meta question about this change: how much additional risk are we incurring by using `InterleavingIter`, given that we cannot afford any bug in our compaction iteration code. I am assuming the answer is very low especially since we use only `First()` and `Next()`, but do we need some additional randomized testing of `InterleavingIter` to give us more confidence?
</blockquote></details>

Yeah, I think it's very low. We use the interleaving iterator today in both user iteration and compaction iterator for range keys. Focused randomized testing would be nice, but I would be shocked if it picked up something that the metamorphic tests haven't yet.
___
*[`testdata/compaction_delete_only_hints` line 224 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3219#-No-FLo23vtz0xxWAmkF:-NoIQt39-taKzBZtj1VP:b-koex8) ([raw file](https://github.com/cockroachdb/pebble/blob/9be83bd239f8f9d3cbae807a8bfe3536ec7a2ee8/testdata/compaction_delete_only_hints#L224)):*

is this diff is because we don't get interrupted by `sameStripeNonSkippable` after seeing the `SET`?
</blockquote></details>

That's right
___
*[`testdata/compaction_delete_only_hints` line 258 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3219#-No-Fe-26mvP3jVWMmiW:-NoIQEcaBHquSzeY1-BM:ba02b67) ([raw file](https://github.com/cockroachdb/pebble/blob/9be83bd239f8f9d3cbae807a8bfe3536ec7a2ee8/testdata/compaction_delete_only_hints#L258)):*

why did the byte count change?
</blockquote></details>

Thanks for the prod here. This was a result of the `snapshot-pinned` sstable properties being written when they weren't previously. They were improperly considering the point key as obsolete, because it had the same user key.

Relatedly, I think I should combine this commit with the one from #3221 so that the compaction iterator test cases reflect the reality of range deletions being interleaved first.
___
*[`testdata/range_del` line 1196 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3219#-No-FpZX1MXePqH8qiRs:-NoIPBzL-alRTKoN7g2H:bigx1kn) ([raw file](https://github.com/cockroachdb/pebble/blob/9be83bd239f8f9d3cbae807a8bfe3536ec7a2ee8/testdata/range_del#L1196)):*

why did the file boundaries change?
</blockquote></details>

This was an artifact of range deletions being interleaved first and the compaction splitter refusing to "split" user keys across files. The splitter is naive and does not know that it can split if the previous key was a range deletion (and is buffered in the fragmenter; not already written to the current output file). I updated the file size splitter to _not_ special case range deletions, which allows this case to split here before the range deletion here. I think either behavior is reasonable here.

sumeerbhola (approved):
:lgtm: 

> Reviewed 3 of 4 files at r4, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3219#-:-NoOI79NCln0of_RzUQ3:bg7igkv)* status: all files reviewed, 11 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`compaction_iter.go` line 330 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3219#-NoNbtFn4UT7k4xbNewD:-NoNbtFn4UT7k4xbNewE:bo7zbbl) ([raw file](https://github.com/cockroachdb/pebble/blob/a237293562d030982f961f9d317da8a0013b4e49/compaction_iter.go#L330)):*
> ```Go
> 	}
> 
> 	// Prior to this call to `Next()` we are in one of four situations with
> ```

nit: three situations
___
*[`compaction_iter.go` line 375 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3219#-NoOD6Wl6FLquI9wG-ya:-NoOD6Wl6FLquI9wG-yb:bxqqsml) ([raw file](https://github.com/cockroachdb/pebble/blob/a237293562d030982f961f9d317da8a0013b4e49/compaction_iter.go#L375)):*
> ```Go
> 			// Note that `skip` must already be false here, because range keys and range
> 			// deletions are interleaved at the maximal sequence numbers and neither will
> 			// set `skip`=true.
> ```

can we assert that skip is false?
___
*[`compaction_iter.go` line 645 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3219#-NoOFCiD5GstY7rvZ_Eb:-NoOFCiD5GstY7rvZ_Ec:b-y90u30) ([raw file](https://github.com/cockroachdb/pebble/blob/a237293562d030982f961f9d317da8a0013b4e49/compaction_iter.go#L645)):*
> ```Go
> 		//    number ordering within a user key. If the previous key was one
> 		//    of these keys, we consider the new key a `newStripeNewKey` to
> 		//    reflect that it's the beginning of a new stream of point keys.
> ```

#2 is a bit subtle, but I guess it is unavoidable.

I glanced through all the callers of `nextInStripe` that care about the return value and it seems all of them call when they are already on a point key. Just curious if there is a test case failing without this `i.key.IsExclusiveSentinel()`?
___
*[`testdata/compaction_iter_delete_sized` line 308 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3219#-NoN__3Z33poO_cc8j8o:-NoN__3Z33poO_cc8j8p:b-s6kfpc) ([raw file](https://github.com/cockroachdb/pebble/blob/a237293562d030982f961f9d317da8a0013b4e49/testdata/compaction_iter_delete_sized#L308)):*
> ```
> tombstones
> ----
> a#72057594037927935,15:
> ```

We are not printing the seqnum of the keys in the span, so it is not possible to verify the correctness here. Could you add that? The fragmented tombstones printed below have the seqnums, but it would be nice if we also printed the unfragmented ones.
Hmm, these are already fragmented, so maybe it doesn't matter.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3219#-:-NoOYvL71VMnV2ZOJwTw:b-814g8k)* status: 9 of 14 files reviewed, 9 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`compaction_iter.go` line 375 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3219#-NoOD6Wl6FLquI9wG-ya:-NoOWo602Nil5iKZyfpA:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/a237293562d030982f961f9d317da8a0013b4e49/compaction_iter.go#L375)):*

can we assert that skip is false?
</blockquote></details>

Done.
___
*[`compaction_iter.go` line 645 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3219#-NoOFCiD5GstY7rvZ_Eb:-NoOX-qz4Ds-FyAbolaN:brbiplc) ([raw file](https://github.com/cockroachdb/pebble/blob/a237293562d030982f961f9d317da8a0013b4e49/compaction_iter.go#L645)):*

#2 is a bit subtle, but I guess it is unavoidable.

I glanced through all the callers of `nextInStripe` that care about the return value and it seems all of them call when they are already on a point key. Just curious if there is a test case failing without this `i.key.IsExclusiveSentinel()`?
</blockquote></details>

Yeah. without it, a compaction iterator that sees `a.RANGEDEL.1:z` and `a.SET.4:a4` will think the `a.SET.4` should be considered a "snapshot-pinned" key because it'll encounter `a.SET.4` with `newStripeSameKey` 
___
*[`testdata/compaction_iter_delete_sized` line 308 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3219#-NoN__3Z33poO_cc8j8o:-NoOYlYO9ZdO1I5WgPmu:bfiasaw) ([raw file](https://github.com/cockroachdb/pebble/blob/a237293562d030982f961f9d317da8a0013b4e49/testdata/compaction_iter_delete_sized#L308)):*

We are not printing the seqnum of the keys in the span, so it is not possible to verify the correctness here. Could you add that? The fragmented tombstones printed below have the seqnums, but it would be nice if we also printed the unfragmented ones.
Hmm, these are already fragmented, so maybe it doesn't matter.
</blockquote></details>

Added them because it at least makes it clearer what's happening

PR #3221, opened 2024-01-12
db: remove compactionIter's sameStripeNonSkippable

jbowens:
The first commit is from #3219 and should be reviewed there.

----

This commit removes the compaction iterator's concept of a key that lies within
the same snapshot stripe but is non-skippable. These "non-skippable" keys lead
to bizarre behavior, such as a sequence like:

  a.MERGE.5 a.RANGEDEL.4 a.MERGE.4

yielding the separate unmerged a.MERGE.5 and a.MERGE.4 internal keys (Note
a.RANGEDEL.4 does not delete a.MERGE.4 because they share the same sequence
number.)

The sameStripeNonSkippable fell out of the fact that range deletions were
previously interleaved by the input iterator at their start key with their
sequence number. These sameStripeNonSkippable could interrupt any logic
iterating across the keys of a stripe, complicating the compaction iterator
logic. With https://github.com/cockroachdb/pebble/pull/3219, range deletions began to be interleaved at their start key
with the maximal sequence number, ensuring they never interrupt the keys of a
snapshot stripe.

In some instances INVALID keys were also returned with sameStripeNonSkippable.
These keys are now treated similarly to other errors encountered during
iteration: i.err is set and newStripeNewKey is returned.

Close https://github.com/cockroachdb/pebble/issues/3082.


jbowens:
Merged into #3219 to produce a more coherent diff.

PR #3223, opened 2024-01-16
internal/keyspan: add error return value to FragmentIterator methods

jbowens:
This commit refactors the keyspan.FragmentIterator interface to propagate
errors from the positioning method that encounters the error, removing the
`Error()` method from the interface. This makes it more difficult to
accidentally ignore an error value by forgetting to check `Error()`.

### DIFF
--- a/batch_test.go
+++ b/batch_test.go
@@ -981,7 +981,8 @@ func TestBatchRangeOps(t *testing.T) {
 
 			var buf bytes.Buffer
 			if fragmentIter != nil {
-				for s := fragmentIter.First(); s != nil; s = fragmentIter.Next() {
+				s, err := fragmentIter.First()
+				for ; s != nil; s, err = fragmentIter.Next() {
 					for i := range s.Keys {
 						s.Keys[i].Trailer = base.MakeTrailer(
 							s.Keys[i].SeqNum()&^base.InternalKeySeqNumBatch,
@@ -989,6 +990,9 @@ func TestBatchRangeOps(t *testing.T) {
 						)
 					}
 					fmt.Fprintln(&buf, s)
+				}
+				if err != nil {
+					return err.Error()
 				}
 			} else {
 				for k, v := internalIter.First(); k != nil; k, v = internalIter.Next() {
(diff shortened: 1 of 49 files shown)


RaduBerinde (approved):
:lgtm:  Cool!

> Reviewed 49 of 49 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3223#-:-NoSgLbhAVNq1Rik2rfC:b-hq9zt)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`internal/keyspan/level_iter.go` line 509 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3223#-NoSfr1F8WE6y0eAP6qd:-NoSfr1F8WE6y0eAP6qe:b-tio1ow) ([raw file](https://github.com/cockroachdb/pebble/blob/aa9048549c51f7f2da125bc461c7825249ba9866/internal/keyspan/level_iter.go#L509)):*
> ```Go
> // caller.
> func (l *LevelIter) verify(s *Span, err error) (*Span, error) {
> 	// NB: Do not add any logic outside the invariants.Enabled conditional to
> ```

Shoudl we assert that `l.err` is set whenever `err` is? (or maybe even that they are the same)

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3223)* status: 48 of 49 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`internal/keyspan/level_iter.go` line 509 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3223#-NoSfr1F8WE6y0eAP6qd:-NoTAKEU4sf3R612hSHs:b-fznoh6) ([raw file](https://github.com/cockroachdb/pebble/blob/aa9048549c51f7f2da125bc461c7825249ba9866/internal/keyspan/level_iter.go#L509)):*

Shoudl we assert that `l.err` is set whenever `err` is? (or maybe even that they are the same)
</blockquote></details>

Good idea, done!

jbowens:
TFTR!

PR #3224, opened 2024-01-16
wal: failover write path code

sumeerbhola:
failover_manager.go contains the failoverManager (which implements
wal.Manager) for the write path, and helper classes. dirProber monitors
the primary dir when failed over to use the secondary, to decide when to
failback to the primary. failoverMonitor uses the latency and error seen
by the current *LogWriter, and probing state, to decide when to switch to
a different dir.

Informs #3230

Informs CRDB-35401

### DIFF
--- a/open.go
+++ b/open.go
@@ -322,19 +322,18 @@ func Open(dirname string, opts *Options) (db *DB, err error) {
 	})
 	walManager := &wal.StandaloneManager{}
 	err = walManager.Init(wal.Options{
-		Primary:                        wal.Dir{FS: opts.FS, Dirname: walDirname},
-		Secondary:                      wal.Dir{},
-		MinUnflushedWALNum:             wal.NumWAL(d.mu.versions.minUnflushedLogNum),
-		MaxNumRecyclableLogs:           opts.MemTableStopWritesThreshold + 1,
-		NoSyncOnClose:                  opts.NoSyncOnClose,
-		BytesPerSync:                   opts.WALBytesPerSync,
-		PreallocateSize:                d.walPreallocateSize,
-		MinSyncInterval:                opts.WALMinSyncInterval,
-		FsyncLatency:                   d.mu.log.metrics.fsyncLatency,
-		QueueSemChan:                   d.commit.logSyncQSem,
-		ElevatedWriteStallThresholdLag: 0,
-		Logger:                         opts.Logger,
-		EventListener:                  walEventListenerAdaptor{l: opts.EventListener},
+		Primary:              wal.Dir{FS: opts.FS, Dirname: walDirname},
+		Secondary:            wal.Dir{},
+		MinUnflushedWALNum:   wal.NumWAL(d.mu.versions.minUnflushedLogNum),
+		MaxNumRecyclableLogs: opts.MemTableStopWritesThreshold + 1,
+		NoSyncOnClose:        opts.NoSyncOnClose,
+		BytesPerSync:         opts.WALBytesPerSync,
+		PreallocateSize:      d.walPreallocateSize,
+		MinSyncInterval:      opts.WALMinSyncInterval,
+		FsyncLatency:         d.mu.log.metrics.fsyncLatency,
+		QueueSemChan:         d.commit.logSyncQSem,
+		Logger:               opts.Logger,
+		EventListener:        walEventListenerAdaptor{l: opts.EventListener},
 	})
 	if err != nil {
 		return nil, err
--- a/vfs/errorfs/errorfs.go
+++ b/vfs/errorfs/errorfs.go
@@ -83,6 +83,8 @@ const (
 	OpFileSync
 	// OpFileSyncData describes a file sync operation.
 	OpFileSyncData
+	// OpFileSyncTo describes a file sync operation.
+	OpFileSyncTo
 	// OpFileFlush describes a file flush operation.
 	OpFileFlush
 )
@@ -92,7 +94,7 @@ func (o OpKind) ReadOrWrite() OpReadWrite {
 	switch o {
 	case OpOpen, OpOpenDir, OpList, OpStat, OpGetDiskUsage, OpFileRead, OpFileReadAt, OpFileStat:
 		return OpIsRead
-	case OpCreate, OpLink, OpRemove, OpRemoveAll, OpRename, OpReuseForWrite, OpMkdirAll, OpLock, OpFileClose, OpFileWrite, OpFileWriteAt, OpFileSync, OpFileSyncData, OpFileFlush, OpFilePreallocate:
+	case OpCreate, OpLink, OpRemove, OpRemoveAll, OpRename, OpReuseForWrite, OpMkdirAll, OpLock, OpFileClose, OpFileWrite, OpFileWriteAt, OpFileSync, OpFileSyncData, OpFileSyncTo, OpFileFlush, OpFilePreallocate:
 		return OpIsWrite
 	default:
 		panic(fmt.Sprintf("unrecognized op %v\n", o))
@@ -535,7 +537,9 @@ func (f *errorFile) SyncData() error {
 }
 
 func (f *errorFile) SyncTo(length int64) (fullSync bool, err error) {
-	// TODO(jackson): Consider error injection.
+	if err := f.inj.MaybeError(Op{Kind: OpFileSyncTo, Path: f.path}); err != nil {
+		return false, err
+	}
 	return f.file.SyncTo(length)
 }
 
--- a/wal/failover_manager.go
+++ b/wal/failover_manager.go
@@ -9,7 +9,168 @@ import (
 	"time"
 
 	"github.com/cockroachdb/pebble/vfs"
+	"golang.org/x/exp/rand"
 )
+
+// dirProber probes the primary dir, until it is confirmed to be healthy. If
+// it doesn't have enough samples, it is deemed to be unhealthy. It is only
+// used for failback to the primary.
+type dirProber struct {
+	fs vfs.FS
+	// The full path of the file to use for the probe. The probe is destructive
+	// in that it deletes and (re)creates the file.
+	filename string
+	// The probing interval, when enabled.
+	interval time.Duration
+	timeSource
+	// buf holds the random bytes that are written during the probe.
+	buf [100 << 10]byte
+	// enabled is signaled to enable and disable probing. The initial state of
+	// the prober is disabled.
+	enabled chan bool
+	mu      struct {
+		sync.Mutex
+		// Circular buffer of history samples.
+		history [probeHistoryLength]time.Duration
+		// The history is in [firstProbeIndex, nextProbeIndex).
+		firstProbeIndex int
+		nextProbeIndex  int
+	}
+	iterationForTesting chan<- struct{}
+}
+
+const probeHistoryLength = 128
+
+// Large value.
+const failedProbeDuration = 24 * 60 * 60 * time.Second
+
+// There is no close/stop method on the dirProber -- use stopper.
+func (p *dirProber) init(
+	fs vfs.FS,
+	filename string,
+	interval time.Duration,
+	stopper *stopper,
+	ts timeSource,
+	notifyIterationForTesting chan<- struct{},
+) {
+	*p = dirProber{
+		fs:                  fs,
+		filename:            filename,
+		interval:            interval,
+		timeSource:          ts,
+		enabled:             make(chan bool),
+		iterationForTesting: notifyIterationForTesting,
+	}
+	// Random bytes for writing, to defeat any FS compression optimization.
+	_, err := rand.Read(p.buf[:])
+	if err != nil {
+		panic(err)
+	}
+	stopper.runAsync(func() {
+		p.probeLoop(stopper.shouldQuiesce())
+	})
+}
+
+func (p *dirProber) probeLoop(shouldQuiesce <-chan struct{}) {
+	ticker := p.timeSource.newTicker(p.interval)
+	ticker.stop()
+	tickerCh := ticker.ch()
+	done := false
+	var enabled bool
+	for !done {
+		select {
+		case <-tickerCh:
+			if !enabled {
+				// Could have a tick waiting before we disabled it. Ignore.
+				continue
+			}
+			probeDur := func() time.Duration {
+				// Delete, create, write, sync.
+				start := p.timeSource.now()
+				_ = p.fs.Remove(p.filename)
+				f, err := p.fs.Create(p.filename)
+				if err != nil {
+					return failedProbeDuration
+				}
+				defer f.Close()
+				n, err := f.Write(p.buf[:])
+				if err != nil {
+					return failedProbeDuration
+				}
+				if n != len(p.buf) {
+					panic("invariant violation")
+				}
+				err = f.Sync()
+				if err != nil {
+					return failedProbeDuration
+				}
+				return p.timeSource.now().Sub(start)
+			}()
+			p.mu.Lock()
+			nextIndex := p.mu.nextProbeIndex % probeHistoryLength
+			p.mu.history[nextIndex] = probeDur
+			p.mu.nextProbeIndex++
+			numSamples := p.mu.nextProbeIndex - p.mu.firstProbeIndex
+			if numSamples > probeHistoryLength {
+				// Length has exceeded capacity, i.e., overwritten the first probe.
+				p.mu.firstProbeIndex++
+			}
+			p.mu.Unlock()
+
+		case <-shouldQuiesce:
+			done = true
+
+		case enabled = <-p.enabled:
+			if enabled {
+				ticker.reset(p.interval)
+			} else {
+				ticker.stop()
+				p.mu.Lock()
+				p.mu.firstProbeIndex = 0
+				p.mu.nextProbeIndex = 0
+				p.mu.Unlock()
+			}
+		}
+		if p.iterationForTesting != nil {
+			p.iterationForTesting <- struct{}{}
+		}
+	}
+}
+
+func (p *dirProber) enableProbing() {
+	p.enabled <- true
+}
+
+func (p *dirProber) disableProbing() {
+	p.enabled <- false
+}
+
+func (p *dirProber) getMeanMax(interval time.Duration) (time.Duration, time.Duration) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	numSamples := p.mu.nextProbeIndex - p.mu.firstProbeIndex
+	samplesNeeded := int((interval + p.interval - 1) / p.interval)
+	if samplesNeeded == 0 {
+		panic("interval is too short")
+	} else if samplesNeeded > probeHistoryLength {
+		panic("interval is too long")
+	}
+	if samplesNeeded > numSamples {
+		// Not enough samples, so assume not yet healthy.
+		return failedProbeDuration, failedProbeDuration
+	}
+	offset := numSamples - samplesNeeded
+	var sum, max time.Duration
+	for i := p.mu.firstProbeIndex + offset; i < p.mu.nextProbeIndex; i++ {
+		sampleDur := p.mu.history[i%probeHistoryLength]
+		sum += sampleDur
+		if max < sampleDur {
+			max = sampleDur
+		}
+	}
+	mean := sum / time.Duration(samplesNeeded)
+	return mean, max
+}
 
 type dirIndex int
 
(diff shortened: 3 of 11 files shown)


jbowens (commented):
In this design the "virtual WAL" grows endlessly until the stall abates, right? What happens when the memtable fills up? Today memtables and WALs are assumed to be one-to-one, although I'm not sure what breaks when that relationship is broken.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224#-:-NoNvsRPC2GarETd6xVa:brbwxvd)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @nicktrav, @petermattis, and @sumeerbhola)

___
*[`record/log_writer.go` line 240 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3224#-NoNZw9d2PEOXvglGVjt:-NoNZw9e4p6FRM31Qwhv:b-bwp0c8) ([raw file](https://github.com/cockroachdb/pebble/blob/e1b66df49e55e395720d388e6ef2feedf968b907/record/log_writer.go#L240)):*
> ```Go
> }
> 
> // The implementation of pendingSyncs in failover mode.
> ```

Would it be tractable to unify "standalone" and "failover" mode more? I'm finding it a little difficult to follow these interfaces, and having very discrete separate modes makes it harder to test and exercise the "exceptional" mode.

sumeerbhola (commented):
> In this design the "virtual WAL" grows endlessly until the stall abates, right? What happens when the memtable fills up? Today memtables and WALs are assumed to be one-to-one, although I'm not sure what breaks when that relationship is broken.

The one-to-one relationship is between a memtable and virtual WAL -- we can't switch to a new virtual WAL and memtable until the previous one has successfully closed (and fsynced). The memtable filling up will dictate switching to a new virtual WAL (unchanged from what we have now).
In standalone mode (which is retained partly for cases where there is no secondary dir, and partly because there will be some performance regression in the other mode), the virtual WAL corresponds to a single physical WAL file.

In failover mode, the virtual WAL will be backed by potentially many files. Some of those LogWriters may still be open and their flush loops active when the virtual WAL successfully closes. Though typically there will be one failover and one stuck LogWriter.flushLoop.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224#-:-NoNyG5s6GFFy5vHYzDn:bmjrrek)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jbowens, @nicktrav, and @petermattis)

___
*[`record/log_writer.go` line 240 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3224#-NoNZw9d2PEOXvglGVjt:-NoNzQC-AhW6n2xfAKTv:ba4jl4k) ([raw file](https://github.com/cockroachdb/pebble/blob/e1b66df49e55e395720d388e6ef2feedf968b907/record/log_writer.go#L240)):*

Would it be tractable to unify "standalone" and "failover" mode more? I'm finding it a little difficult to follow these interfaces, and having very discrete separate modes makes it harder to test and exercise the "exceptional" mode.
</blockquote></details>

We could, once we have microbenchmarks and are convinced that moving the queuing outside the LogWriter even in standalone mode is not a performance regression (since there will be duplication of synchronization both in the `record.LogWriter` and in whoever has the sync queue).

Roughly, what I am suggesting we do here is (a) abstract where possible the existing code to minimize code copying (like in `record.LogWriter`), (b) build two implementations of the `wal` interfaces (the helper classes, e.g. for WAL reuse, will be shared), (c) switch everything in Pebble to the new `wal` interfaces (at this point there should be no performance regression for the standalone mode), (d) do benchmarking of the performance penalty for the normal case (no actual failover) when running in failover mode, (e) if benchmarking suggests the performance penalty is small, start merging code wherever merging doesn't sacrifice code readability and doesn't sacrifice performance for standalone deployments.

jbowens (commented):
> The memtable filling up will dictate switching to a new virtual WAL (unchanged from what we have now).

Got it, thanks

The WAL interface makes sense to me.

super nit: I have a preference towards removing the `WAL` prefix from the type names to prevent the stutter of `wal.WALWriter`, `wal.WALManager`, etc in packages that consume it

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224#-:-NoOUEvx8OwQqhg-UONr:bvy3hla)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @nicktrav, @petermattis, and @sumeerbhola)

___
*[`wal/wal.go` line 513 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3224#-NoOR4d3D7GekyemK9Bf:-NoOR4d3D7GekyemK9Bg:b-omrrpw) ([raw file](https://github.com/cockroachdb/pebble/blob/e1b66df49e55e395720d388e6ef2feedf968b907/wal/wal.go#L513)):*
> ```Go
> }
> 
> // dirProber. One of these in WALManager to probe the primary dir.
> ```

do you think we could rely on our previously blocked writes unblocking in order to determine whether to switch back? or is the concern that the volume might be back, but the in-flight write may not return for a much longer time? If in-flight writes remain blocked, we also need to contend with the possibility of inflight flushes, compactions or manifest writes remaining blocked too.

sumeerbhola (commented):
> super nit: I have a preference towards removing the `WAL` prefix from the type names to prevent the stutter of `wal.WALWriter`, `wal.WALManager`, etc in packages that consume it

removed the stutter from everything except `WALNum`, since `Num` seemed a bit too bare.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224#-:-NoSFAS38TUmzCqLU9yH:b-1n2zns)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @jbowens, @nicktrav, and @petermattis)

sumeerbhola (commented):
I've rebased on master.
There are no unit tests, so the code is probably littered with bugs. But I would value some early feedback while I start writing tests.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224#-:-NoluEnx8iMRMSJaOYP4:bz3fsj5)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @jbowens, @nicktrav, and @petermattis)

___
*[`wal/wal.go` line 513 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3224#-NoOR4d3D7GekyemK9Bf:-Nolt-060XrTNGaWv--C:birmyv5) ([raw file](https://github.com/cockroachdb/pebble/blob/e1b66df49e55e395720d388e6ef2feedf968b907/wal/wal.go#L513)):*

do you think we could rely on our previously blocked writes unblocking in order to determine whether to switch back? or is the concern that the volume might be back, but the in-flight write may not return for a much longer time? If in-flight writes remain blocked, we also need to contend with the possibility of inflight flushes, compactions or manifest writes remaining blocked too.
</blockquote></details>

I wasn't thinking about the in-flight writes staying blocked. Ideally, we want a signal that will represent both (a) in-flight writes being done and (b) new writes finishing quickly. (a) is a bit trickier from a code perspective and we can't rely solely on (a) since if there was only 1 outstanding write and it finishes after 60s, and there is no more work for that LogWriter, it doesn't tell us what will happen for new writes. So we need (b), which is done using the `dirProber`. The code currently ignores (a) altogether, under the assumption that (b) is sufficient (since the `dirProber` also declares the dir to be healthy after a minimum of 2 min of healthy probes). We can revisit this based on real experience.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224)* status: 2 of 10 files reviewed, 7 unresolved discussions (waiting on @jbowens, @nicktrav, and @petermattis)

___
*[`wal/failover_manager.go` line 71 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Np1K78v0gtQ2O6pV9du:-Np1am9E52Z6I74BARqh:b3cnqum) ([raw file](https://github.com/cockroachdb/pebble/blob/271e86aa7781595f1db0061bcca0d5f1c2faa4a3/wal/failover_manager.go#L71)):*

Is it necessary to create a new file? When encryption-at-rest is enabled, this will require a fsync of the registry for the removal of the old probe file and a fsync of the registry for the creation of the new one.
</blockquote></details>

I wanted to test the full lifecycle to be sure things are healthy. We do this infrequently -- it is currently set to 1Hz, and we could further reduce it. But yes, this is somewhat subjective.
___
*[`wal/failover_manager.go` line 297 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Np1PbyZ7oDjO565ldux:-Np1bgMuDOxBHVX3cAoL:b4k01sy) ([raw file](https://github.com/cockroachdb/pebble/blob/271e86aa7781595f1db0061bcca0d5f1c2faa4a3/wal/failover_manager.go#L297)):*

i'm starting to get a little nervous about the complexity involved in switching back and forth, probing, accounting for errors, tuning heuristics, etc, and it's making me second guess whether we should be trying to instead always write to two WALs on separate volumes in parallel, requiring fsync from one of the logs to inform the user. the read-time merging seems very manageable to me, especially if we rotate WALs in parallel, so one should always have a superset of the other. The extra write bandwidth on the secondary volume/store is a concern, but we should get reduced WAL sync latencies from hedging as a benefit. Just throwing this out there to see if your appetite for it has changed after thinking through this switching in detail
</blockquote></details>

I'm not worried about the complexity in this file. This is also just a heuristic and I may have over-thought it. Conceptually, the composition done by the the `failoverMonitor`, of the signals from the `failoverWriter` and the `dirProber` is quite straightforward -- the code commentary here is rather poor, and I suspect it is making it seem more complex than it is. I will add more code comments.

Always writing to two WAL files will simplify the `failoverWriter` code -- `Close` especially required a lot of care.

The cons I can think of are:
- It doesn't easily extend to more than one failover device. I wouldn't be surprised if the 10 store folks want to eventually failover the WAL on a store to 9 other stores.
- Since these virtual block devices are themselves a distributed system, it seems possible that stuckness on one file does not cause stuckness in another file on the same block device (this was the case in a system from a past life) -- with two WAL files we are limited to those files for the lifetime of the memtable.
- Will users be unhappy that their failover device is constantly being written to? They do need to provision it wrt IOPS and bandwidth to handle the failover, so perhaps they have no reason to complain.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224)* status: 2 of 10 files reviewed, 7 unresolved discussions (waiting on @jbowens, @nicktrav, and @petermattis)

___
*[`wal/failover_manager.go` line 297 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Np1PbyZ7oDjO565ldux:-Np1sGQ05Gx0fYPmk48Q:b6wgerg) ([raw file](https://github.com/cockroachdb/pebble/blob/271e86aa7781595f1db0061bcca0d5f1c2faa4a3/wal/failover_manager.go#L297)):*

I'm not worried about the complexity in this file. This is also just a heuristic and I may have over-thought it. Conceptually, the composition done by the the `failoverMonitor`, of the signals from the `failoverWriter` and the `dirProber` is quite straightforward -- the code commentary here is rather poor, and I suspect it is making it seem more complex than it is. I will add more code comments.

Always writing to two WAL files will simplify the `failoverWriter` code -- `Close` especially required a lot of care.

The cons I can think of are:
- It doesn't easily extend to more than one failover device. I wouldn't be surprised if the 10 store folks want to eventually failover the WAL on a store to 9 other stores.
- Since these virtual block devices are themselves a distributed system, it seems possible that stuckness on one file does not cause stuckness in another file on the same block device (this was the case in a system from a past life) -- with two WAL files we are limited to those files for the lifetime of the memtable.
- Will users be unhappy that their failover device is constantly being written to? They do need to provision it wrt IOPS and bandwidth to handle the failover, so perhaps they have no reason to complain.
</blockquote></details>

Another more compelling con just occurred to me while cooking dinner (I wasn't looking for one):
users with n stores (n > 1) may decide to provision a single extra failover volume (to reduce interference with the normal workload) under the assumption that n+1 is good enough to tolerate a max of one failure (standard provisioning/redundancy assumption). This also relates to the internal conversation about using the VM root disk. In that case we wouldn't want the load fanin on that extra volume at all times.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224)* status: 2 of 10 files reviewed, 7 unresolved discussions (waiting on @jbowens, @nicktrav, and @petermattis)

___
*[`wal/failover_manager.go` line 297 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Np1PbyZ7oDjO565ldux:-Np58SedABo6Uf3z-QAO:bvjnur0) ([raw file](https://github.com/cockroachdb/pebble/blob/271e86aa7781595f1db0061bcca0d5f1c2faa4a3/wal/failover_manager.go#L297)):*

Another more compelling con just occurred to me while cooking dinner (I wasn't looking for one):
users with n stores (n > 1) may decide to provision a single extra failover volume (to reduce interference with the normal workload) under the assumption that n+1 is good enough to tolerate a max of one failure (standard provisioning/redundancy assumption). This also relates to the internal conversation about using the VM root disk. In that case we wouldn't want the load fanin on that extra volume at all times.
</blockquote></details>

I've documented some of this discussion and the multiple design choices we have in https://docs.google.com/document/d/1vAsftzyPG-kDy-A2Ic1fZeKd4OKJIf6N7KvNXRpDAFA/edit#bookmark=id.ecuq6vf8g9bu. Feel free to add stuff there. I'm still leaning significantly towards WAL failover, but happy to discuss more.

I also realized that `Close` can probably be simplified. The complexity comes from the waiting for different conditions and the early decision to ignore errors from certain LogWriters. What `Close` really needs to do is (a) issue close on all LogWriters, (b) wait until its `recordQueue` is empty. But currently `Close` itself needs to empty the `recordQueue` since if the last record was not required to be synced, the LogWriter won't do the callback. If we added a `LogWriter.AsyncClose(PendingSync)` with the index of the last record, then `LogWriter` can do the callback. And the `Close` can chose later to do `LogWriter.WaitForClose()` (which would return `error`) if the queue was non-empty and no more `LogWriters` could be created, and return that `error`. I'm going to attempt this.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224)* status: 2 of 10 files reviewed, 7 unresolved discussions (waiting on @jbowens, @nicktrav, and @petermattis)

___
*[`wal/failover_manager.go` line 297 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Np1PbyZ7oDjO565ldux:-Np5rt0x-voKNbjMMD3S:b5r44re) ([raw file](https://github.com/cockroachdb/pebble/blob/271e86aa7781595f1db0061bcca0d5f1c2faa4a3/wal/failover_manager.go#L297)):*
> I'm going to attempt this.

I also need to rework the use of interfaces in `record.LogWriter` since they cause structs to escape to the heap.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224)* status: 1 of 10 files reviewed, 7 unresolved discussions (waiting on @jbowens, @nicktrav, and @petermattis)

___
*[`wal/failover_manager.go` line 297 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Np1PbyZ7oDjO565ldux:-NpAo3f54yuf470Oam5e:b-293qbk) ([raw file](https://github.com/cockroachdb/pebble/blob/271e86aa7781595f1db0061bcca0d5f1c2faa4a3/wal/failover_manager.go#L297)):*

> I'm going to attempt this.

I also need to rework the use of interfaces in `record.LogWriter` since they cause structs to escape to the heap.
</blockquote></details>

I've reworked all of this and put it in a new commit. `failoverWriter.Close` is now simpler (~75 lines of code minus comments).

sumeerbhola (commented):
TFTR!
I've addressed the `LogWriter` comments and will move it into a separate PR as we discussed.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224#-:-NpLjLkm9V9B_72BX-tq:bf46qrw)* status: 1 of 10 files reviewed, 7 unresolved discussions (waiting on @jbowens, @nicktrav, and @petermattis)

___
*[`record/log_writer.go` line 448 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3224#-NomN5qy0YwzM0O_dB5A:-NpL3zke2G6jx32LiV-Z:b-wbo039) ([raw file](https://github.com/cockroachdb/pebble/blob/05ea0d2b70183de6241145159992e51c9c0de446/record/log_writer.go#L448)):*

not sure how impactful it is, but I could imagine allocations contributing to tail latencies since this is the commit hot path and allocations may occasionally need to go to the operating system to get more memory.
</blockquote></details>

I've removed the TODO.
___
*[`record/log_writer.go` line 610 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Np1CBeO1D3urFl7pQyE:-NpL5-4r4gun7ndbFhfR:budkv0b) ([raw file](https://github.com/cockroachdb/pebble/blob/271e86aa7781595f1db0061bcca0d5f1c2faa4a3/record/log_writer.go#L610)):*

Is this related to the comment down below?:

```
		// Grab the list of sync waiters. Note that syncQueue.load() will return
		// 0,0 while we're waiting for the min-sync-interval to expire. This
		// allows flushing to proceed even if we're not ready to sync.
```

It seems like `q.load()` will return 0,0 and `pendingSyncs.empty()` will return true if syncing is blocked.
</blockquote></details>

I missed that `syncQueue.empty` is using `syncQueue.load` so the right thing is happening.
Removed the TODO.
___
*[`record/log_writer.go` line 661 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Np1GBtB06nDQaF_FcKa:-NpLj53v3MPiTG7ZgNiU:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/271e86aa7781595f1db0061bcca0d5f1c2faa4a3/record/log_writer.go#L661)):*

Was the move of this unlock to avoid holding the lock while popping in the `fErr != nil` case? And that's   necessary because popping may need to invoke the "external sync queue" callback, which may need to acquire a WAL manager mutex? Maybe a short comment above the `pop` would help here
</blockquote></details>

Done

jbowens (commented):
Gave it all a read but didn't fully process everything. I think splitting into another PR makes sense now

> Reviewed 5 of 11 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224#-:-Nq4YZEK3riN5DeGqTSm:b-ke46sd)* status: 5 of 12 files reviewed, 10 unresolved discussions (waiting on @nicktrav, @petermattis, and @sumeerbhola)

___
*[`wal/failover_manager.go` line 71 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Np1K78v0gtQ2O6pV9du:-Nq3vqux-kf2YcPrrjHv:b-dxcbts) ([raw file](https://github.com/cockroachdb/pebble/blob/271e86aa7781595f1db0061bcca0d5f1c2faa4a3/wal/failover_manager.go#L71)):*

I wanted to test the full lifecycle to be sure things are healthy. We do this infrequently -- it is currently set to 1Hz, and we could further reduce it. But yes, this is somewhat subjective.
</blockquote></details>

Makes sense.
___
*[`wal/failover_manager.go` line 174 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq41jMJ7wQyhTECBbIj:-Nq41jMJ7wQyhTECBbIk:b-opp2aa) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_manager.go#L174)):*
> ```Go
> 	unhealthyOperationLatencyThreshold time.Duration
> 
> 	elevatedWriteStallThresholdLag time.Duration
> ```

nit: comments for these durations would be helpful, as it wasn't obvious to me what they all represented
___
*[`wal/failover_manager.go` line 220 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq42RZIBMuscfcagwE2:-Nq42NjFAz4Zf-m4H-Uy:b-evaabe) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_manager.go#L220)):*
> ```Go
> }
> 
> // writerCreateFunc is allowed to return nil.
> ```

in what circumstances would it return nil?
___
*[`wal/failover_writer.go` line 28 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq49152-wYjkGJlmt_n:-Nq49152-wYjkGJlmt_o:b8u4evv) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_writer.go#L28)):*
> ```Go
> const initialBufferLen = 8192
> 
> // recordQueue is a variable size single producer multiple consumer queue. It
> ```

nittiest of nits: hyphenating as "variable-size single-producer multiple-consumer" makes this parse a little easier.
___
*[`wal/failover_writer.go` line 324 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq4RQe89wblsmEtfi9J:-Nq4RQe89wblsmEtfi9K:bbqvv2m) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_writer.go#L324)):*
> ```Go
> 	logicalOffset struct {
> 		latestWriterInWriteRecord  *record.LogWriter
> 		latestLogSizeInWriteRecord int64
> ```

I'm having trouble groking the flow of `latest[LogSize,Writer]InWriteRecord` and logical offsets more generally. I _think_ I understand the problem. When we call `SyncRecordGeneralized` on the current WAL, we get back a physical offset for that individual file, assuming it will eventually be synced. When we switch to a new physical file, some tail of that writer's log size will be repeated on the new file. We want to present to the caller a monotonically increasing "logical WAL size".

As far as I know, we only use this size to serve as the current WAL's size in our metrics. Do we even need a detailed accounting of the logSize or could we expose a `LogSize() (unsynced, synced int64)` here and on `*record.LogWriter`. Then `failoverWriter` could implement it by summing the `synced` sizes of all prior logs and the `unsynced` size of the most recent log.
___
*[`wal/failover_writer.go` line 451 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq4N3AR-aVZ7ApGyrAN:-Nq4N3AR-aVZ7ApGyrAO:b-1b8r84) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_writer.go#L451)):*
> ```Go
> 		// handleErrFunc is called when err != nil. It handles the multiple IO error
> 		// cases below.
> 		handleErrFunc := func() {
> ```

should we pass `err` in? it's super easy to accidentally shadow `err` by using `:=` to declare some other variable
___
*[`wal/failover_writer.go` line 594 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq4LiGG0Fk2e6aKhoZ7:-Nq4LiGG0Fk2e6aKhoZ8:b-1wtzhf) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_writer.go#L594)):*
> ```Go
> //
> // See the long comment about Close behavior where failoverWriter is declared.
> func (ww *failoverWriter) Close() (logicalOffset int64, err error) {
> ```

Would it be cleaner to "Close" discarded log writers at the moment we switch over? We're guaranteed to no longer rely on any persistence, and doing that at switchover time might make that clearer (and remove the `isLastWriter` special casing here?)

At the time we switch over, the current writer is either already created or not. If it's not, when it wakes it'll notice `ww.mu.closed` and close itself. If it is, we close it then asynchronously.

sumeerbhola (commented):
Addressed comments. I'll separate `failoverWriter` into a new PR.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224#-:-Nq9CvQYDPkWOIYMSSB4:b-j74ubl)* status: 4 of 12 files reviewed, 9 unresolved discussions (waiting on @jbowens, @nicktrav, and @petermattis)

___
*[`wal/failover_manager.go` line 97 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Np1MD5q63agSUqII5__:-Nq8hwNY81AljHy-tdYA:b-3v1tmk) ([raw file](https://github.com/cockroachdb/pebble/blob/271e86aa7781595f1db0061bcca0d5f1c2faa4a3/wal/failover_manager.go#L97)):*

how do `firstProbeIndex` and `nextProbeIndex` work? It seems like they'd always be equal?
</blockquote></details>

This was buggy (unsurprising, since I haven't written a test yet). This logic should have been written in terms of length and capacity. Fixed.
___
*[`wal/failover_manager.go` line 281 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Np1PFEv7NCGym85YU_8:-Nq8xxAgCoUK0N6D-cou:bgr2fnb) ([raw file](https://github.com/cockroachdb/pebble/blob/271e86aa7781595f1db0061bcca0d5f1c2faa4a3/wal/failover_manager.go#L281)):*

what does "two times" mean here?
</blockquote></details>

Rephrased

```
// Arbitrary value.  
const switchImmediatelyCountThreshold = 2  
// High latency. Switch immediately if the number of switches that  
// have been done is below the threshold, since that gives us an  
// observation of both dirs. Once above that threshold, decay the  
// switch rate by increasing the observed latency needed for a  
// switch.
```
___
*[`wal/failover_manager.go` line 174 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq41jMJ7wQyhTECBbIj:-Nq8tHZ4CCACu2bqxqN4:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_manager.go#L174)):*

nit: comments for these durations would be helpful, as it wasn't obvious to me what they all represented
</blockquote></details>

Done
___
*[`wal/failover_manager.go` line 220 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq42RZIBMuscfcagwE2:-Nq8vK1RAMDUe5xzlxsH:b8yikdy) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_manager.go#L220)):*

in what circumstances would it return nil?
</blockquote></details>

I've updated the comment:
```
// writerCreateFunc is allowed to return nil, if there is an error. It is not  
// the responsibility of failoverMonitor to handle that error. So this should not
// be due to an IO error (which failoverMonitor is interested in).
```

Currently `newFailoverWriter` never returns an error. I considered removing the error return value, but left it in since it seemed unproductive to start removing error plumbing before this code is mature.
___
*[`wal/failover_writer.go` line 28 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq49152-wYjkGJlmt_n:-Nq8z3Ag5JkFDVF3AIyr:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_writer.go#L28)):*

nittiest of nits: hyphenating as "variable-size single-producer multiple-consumer" makes this parse a little easier.
</blockquote></details>

Done.
___
*[`wal/failover_writer.go` line 324 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq4RQe89wblsmEtfi9J:-Nq94XmNAPN5d-4PJIqH:bcjbjqg) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_writer.go#L324)):*

I'm having trouble groking the flow of `latest[LogSize,Writer]InWriteRecord` and logical offsets more generally. I _think_ I understand the problem. When we call `SyncRecordGeneralized` on the current WAL, we get back a physical offset for that individual file, assuming it will eventually be synced. When we switch to a new physical file, some tail of that writer's log size will be repeated on the new file. We want to present to the caller a monotonically increasing "logical WAL size".

As far as I know, we only use this size to serve as the current WAL's size in our metrics. Do we even need a detailed accounting of the logSize or could we expose a `LogSize() (unsynced, synced int64)` here and on `*record.LogWriter`. Then `failoverWriter` could implement it by summing the `synced` sizes of all prior logs and the `unsynced` size of the most recent log.
</blockquote></details>

The synced size can change for a `LogWriter` even after a switch, so we'd need to grab the synced sizes at the moment of the switch (from the previous LogWriter). The moment of the switch is in `recordQueue.snapshotAndSwitchWriter` since that's where we have embedded all the synchronization. And `recordQueue` is also what a LogWriter calls to pop things. So we would need to grab the synced size of the LogWriter in that pop callback (and only when the pop is being done by the latest LogWriter). This seems complex and I haven't even fully thought about the locking implications.

In comparison what is here seems conceptually simple.
```
latestWriterOffset, ... := w.SyncRecordGeneralized() 
offset += latestWriterOffset - prevWriterOffset
prevWriterOffset = latestWriterOffset
```

The only tricky part is maintaining the `prevWriterOffset` for `w`, since `w` can change during a switch. So `prevWriterOffset` needs to be maintained by the `recordQueue` that handles the switch. We could hand `recordQueue` that `prevWriterOffset` after the `SyncRecordGeneralized`, but then we are doing two mutex acquisitions, one before to grab `w` from `recordQueue` and one after to tell `recordQueue` to update `prevWriterOffset`. And in between these two mutex acquisitions, it is possible that the writer has switched so `recordQueue` still needs the logic to reject stale updates from a stale writer. We've taken this a step further and eliminated the second mutex acquisition and simply remember the `prevWriterOffset` and `w` tentatively in `latestWriter*` and hand them to `recordQueue` when we next ask it for a `LogWriter`.

The `notEstimatedOffset` is an unrelated wrinkle to deal with the absence of any `LogWriter`. It is not something that `recordQueue` knows or cares about, and it is handled within `WriteRecord`. It also isn't perfect. I've added a longer comment in `WriteRecord`, and a todo there for a future PR.
___
*[`wal/failover_writer.go` line 451 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq4N3AR-aVZ7ApGyrAN:-Nq9-cmc6GiuXZqri_dW:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_writer.go#L451)):*

should we pass `err` in? it's super easy to accidentally shadow `err` by using `:=` to declare some other variable
</blockquote></details>

Done
___
*[`wal/failover_writer.go` line 594 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq4LiGG0Fk2e6aKhoZ7:-Nq92BXV-GP8pBrzxql5:bibq4kz) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_writer.go#L594)):*
> Would it be cleaner to "Close" discarded log writers at the moment we switch over? We're guaranteed to no longer rely on any persistence,

We can certainly do that, since the older writer will not be fed any more records.

> and doing that at switchover time might make that clearer (and remove the `isLastWriter` special casing here?)

The difficulty is that new log writers can be created while `failoverWriter.Close` is ongoing. Say `nextWriterIndex=5` when `failoverWriter.Close` is called. Then it will start closing index 4 (which is the last writer), but if that writer is stuck, a `switchToNewDir` can create index 5. Now we will need to coordinate with `switchToNewDir` so that it doesn't close index 4 (since the close has already been initiated). Keeping all the closing in `failoverWriter.Close` eliminates that coordination. And `failoverWriter.Close` still needs to have `lastWriterState` since the last writer is changing.

It would eliminate the `!isLastWriter` conditional, but there's only ~4 lines of code to handle that case.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224)* status: 4 of 12 files reviewed, 8 unresolved discussions (waiting on @nicktrav, @petermattis, and @sumeerbhola)

___
*[`wal/failover_writer.go` line 324 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq4RQe89wblsmEtfi9J:-NqEBJV585M5qekRLTPQ:bfjh6zm) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_writer.go#L324)):*
>  I've added a longer comment in `WriteRecord`, and a todo there for a future PR.

Sounds good. I'm not sure I have a better suggestion, just a sense that it's a lot of distracting complexity for a metric that's only used to surface the physical log size in DB.Metrics. The metric is inherently ambiguous to begin with in the failover case.

I think it would even be reasonable to:
1. remove the offset return value from the write record path replacing it with a `LogSize() (synced, unsynced int64)`
2. call `LogSize` on the current WAL / WAL manager in `DB.Metrics`
3. implement `LogSize` in the failover case by summing `synced` across older logs and `unsynced` from the current.

The older logs' synced values may grow after we've passed them, but that's arguably more accurate for the only usage of accounting physical size of the WAL.
___
*[`wal/failover_writer.go` line 594 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq4LiGG0Fk2e6aKhoZ7:-NqEESkk-Pv-WW9cLWoi:b1pim4o) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_writer.go#L594)):*

> Would it be cleaner to "Close" discarded log writers at the moment we switch over? We're guaranteed to no longer rely on any persistence,

We can certainly do that, since the older writer will not be fed any more records.

> and doing that at switchover time might make that clearer (and remove the `isLastWriter` special casing here?)

The difficulty is that new log writers can be created while `failoverWriter.Close` is ongoing. Say `nextWriterIndex=5` when `failoverWriter.Close` is called. Then it will start closing index 4 (which is the last writer), but if that writer is stuck, a `switchToNewDir` can create index 5. Now we will need to coordinate with `switchToNewDir` so that it doesn't close index 4 (since the close has already been initiated). Keeping all the closing in `failoverWriter.Close` eliminates that coordination. And `failoverWriter.Close` still needs to have `lastWriterState` since the last writer is changing.

It would eliminate the `!isLastWriter` conditional, but there's only ~4 lines of code to handle that case.
</blockquote></details>

Ah I see I forgot that attempting to close the last writer can create a new writer :)

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224)* status: 4 of 12 files reviewed, 8 unresolved discussions (waiting on @jbowens, @nicktrav, and @petermattis)

___
*[`wal/failover_writer.go` line 324 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3224#-Nq4RQe89wblsmEtfi9J:-NqNseqn1MgiTHRZaBw4:b2ja0wj) ([raw file](https://github.com/cockroachdb/pebble/blob/d31904a5a730d960dc026e188bc52b24e764c37a/wal/failover_writer.go#L324)):*
> The metric is inherently ambiguous to begin with in the failover case.

agreed

> The older logs' synced values may grow after we've passed them, but that's arguably more accurate for the only usage of accounting physical size of the WAL.

That is a fair point. There is an existing TODO in wal.go
```
// TODO(sumeer): with failover, Metrics.WAL.BytesWritten needs to be  
// maintained here.
```
how about revisiting this when fixing that todo?

jbowens (approved):
:lgtm:

Do you want to merge this as-is, and we can flesh out the remaining TODOs and tests incrementally afterwards?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224#-:-NqYoiR0AH4dXfA_P4xx:b-lhabbb)* status: 0 of 12 files reviewed, 7 unresolved discussions (waiting on @nicktrav, @petermattis, and @sumeerbhola)

sumeerbhola (commented):
Tests are ready. PTAL

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224#-:-NqeO-N79R74apy8x4yK:bwffwj8)* status: 0 of 16 files reviewed, 7 unresolved discussions (waiting on @jbowens, @nicktrav, and @petermattis)

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224)* status: 7 of 16 files reviewed, 8 unresolved discussions (waiting on @jbowens, @nicktrav, and @petermattis)

___
*[`wal/failover_manager.go` line 177 at r15](https://reviewable.io/reviews/cockroachdb/pebble/3224#-NqhQa8gCUUoz0cNt4q_:-Nqi5gEk2Xst3TJRFLxt:b-pbj1ek) ([raw file](https://github.com/cockroachdb/pebble/blob/2286d40b8d0e2e6ef8ff7b0a913baaadb996d575/wal/failover_manager.go#L177)):*

nit: could put this in a _test.go file since it's only used in this package's unit tests
</blockquote></details>

Done for this and `failoverMonitor.printStateForTesting`.
___
*[`wal/failover_manager.go` line 212 at r15](https://reviewable.io/reviews/cockroachdb/pebble/3224#-NqhS3zu6akTyopYJk6f:-Nqi7-AjEZXfB6APE1bl:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/2286d40b8d0e2e6ef8ff7b0a913baaadb996d575/wal/failover_manager.go#L212)):*

would it be cleaner to embed the FailoverOptions directly?
</blockquote></details>

Done

sumeerbhola (commented):
I also did some small test tweaks to eliminate some flakes. There are a couple of remaining flakes under stress, that existed prior to this PR (TestFailoverWriter), so I plan to merge and fix them later.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3224#-:-Nqi7_hfDH0c6frhh3Ko:b-8psifg)* status: 3 of 16 files reviewed, 8 unresolved discussions (waiting on @jbowens, @nicktrav, and @petermattis)

PR #3227, opened 2024-01-17
base: clean up FileNum <-> DiskFileNum conversions

RaduBerinde:
We rename the methods to `PhysicalTableDiskFileNum` and
`PhysicalTableFileNum` which makes the assumption behind the
conversion evident at every call site. We also clean up some
unnecessary conversions and improve some relevant comments.

### DIFF
--- a/checkpoint_test.go
+++ b/checkpoint_test.go
@@ -285,7 +285,7 @@ func TestCheckpointCompaction(t *testing.T) {
 					if tbl.Virtual {
 						continue
 					}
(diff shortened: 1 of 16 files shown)


itsbilal (approved):
LGTM

Should we also take this opportunity to rename the type to `BackingFileNum` from `DiskFileNum`?

RaduBerinde:
TFTR!

`DiskFileNum` is also used for other files (manifests, logs, etc). In that context, "backing file" makes less sense. IMO we should rename `FileNum` to `TableNum` (and `FileMetadata` to `TableMetadata`), and after that we can rename `DiskFileNum` back to `FileNum`. But I'm hesitating because these would be very disruptive changes.

PR #3235, opened 2024-01-19
objstorage: simplify AttachRemoteObjects

RaduBerinde:
#### ingest: use FileNum instead of DiskNum when loading tables

We pass around pending `FileNum`s instead of `DiskFileNum`s, which
better matches the usage.

#### ingest: attach external objects at link time

Move the attaching of external objects to the link phase. We already
attach the shared objects at this phase. We now attach all remote
objects with a single call.

#### objstorage: simplify AttachRemoteObjects

Simplify `AttachRemoteObjects` to not call `sharedSync`. We already
call `Sync` as part of ingestion.

Add a call `sharedSync` when closing the provider (this should be a
no-op in practice, other than tests).

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -187,27 +187,6 @@ func ingestLoad1External(
 	// DiskFileNum (even though this is a virtual sstable).
 	meta.InitProviderBacking(base.DiskFileNum(fileNum))
 
-	// Try to resolve a reference to the external file.
-	backing, err := objProvider.CreateExternalObjectBacking(e.Locator, e.ObjName)
-	if err != nil {
-		return nil, err
-	}
-	metas, err := objProvider.AttachRemoteObjects([]objstorage.RemoteObjectToAttach{{
-		FileNum:  meta.FileBacking.DiskFileNum,
-		FileType: fileTypeTable,
-		Backing:  backing,
-	}})
-	if err != nil {
-		return nil, err
-	}
-	if opts.EventListener.TableCreated != nil {
-		opts.EventListener.TableCreated(TableCreateInfo{
-			JobID:   jobID,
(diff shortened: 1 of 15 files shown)


ingest.go
```diff
@@ -182,36 +178,36 @@ func ingestLoad1External(
...
+	// For simplicity, we use the same number for both the FileNum and the
+	// DiskFileNum (even though this is a virtual sstable).
+	meta.InitProviderBacking(base.DiskFileNum(fileNum))
+
+	//// Try to resolve a reference to the external file.
+	//backing, err := objProvider.CreateExternalObjectBacking(e.Locator, e.ObjName)
+	//if err != nil {
+	//	return nil, err
+	//}
+	//metas, err := objProvider.AttachRemoteObjects([]objstorage.RemoteObjectToAttach{{
```
itsbilal, line 195: Should probably delete this comment?

itsbilal (approved):
LGTM

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3235#-:-Non8kZM6-YDPYK7Ke52:b-814g8k)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`ingest.go` line 195 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3235#gh-1462429172:-Non8iqF7-NEpsYNeoPw:b9cedjz) ([raw file](https://github.com/cockroachdb/pebble/blob/1488ea888326c98828c4028fcb3c5de2299aab0e/ingest.go#L195)):*

Should probably delete this comment?
</blockquote></details>

Done. Thanks for catching it.

PR #3237, opened 2024-01-19
sstable: add block level synthetic prefix support

dt:
This adds the concept of synthetic prefixes to sstable readers and their underlying block iterators, treating the synthetic prefix as an extra shared prefix, shared even by restart keys.

When a reader is configured with a synthetic prefix, it will assume that that prefix is implicitly prepended to every key in the underlying sst blocks when interacting with or returning those keys.

### DIFF
--- a/sstable/block.go
+++ b/sstable/block.go
@@ -5,6 +5,7 @@
 package sstable
 
 import (
+	"bytes"
 	"context"
 	"encoding/binary"
 	"unsafe"
@@ -404,20 +405,22 @@ type blockIter struct {
 	cachedBuf []byte
 	handle    bufferHandle
 	// for block iteration for already loaded blocks.
-	firstUserKey      []byte
-	lazyValueHandling struct {
+	firstUserKey           []byte
+	firstUserKeyWithPrefix []byte
+	lazyValueHandling      struct {
 		vbr            *valueBlockReader
 		hasValuePrefix bool
 	}
 	hideObsoletePoints bool
+	prefix             SyntheticPrefix
 }
 
 // blockIter implements the base.InternalIterator interface.
 var _ base.InternalIterator = (*blockIter)(nil)
 
-func newBlockIter(cmp Compare, block block) (*blockIter, error) {
+func newBlockIter(cmp Compare, block block, syntheticPrefix SyntheticPrefix) (*blockIter, error) {
 	i := &blockIter{}
-	return i, i.init(cmp, block, 0, false)
+	return i, i.init(cmp, block, 0, false, syntheticPrefix)
 }
 
 func (i *blockIter) String() string {
@@ -425,19 +428,24 @@ func (i *blockIter) String() string {
 }
 
 func (i *blockIter) init(
-	cmp Compare, block block, globalSeqNum uint64, hideObsoletePoints bool,
+	cmp Compare, block block, globalSeqNum uint64, hideObsoletePoints bool, syntheticPrefix SyntheticPrefix,
 ) error {
 	numRestarts := int32(binary.LittleEndian.Uint32(block[len(block)-4:]))
 	if numRestarts == 0 {
 		return base.CorruptionErrorf("pebble/table: invalid table (block has no restart points)")
 	}
+	i.prefix = syntheticPrefix
 	i.cmp = cmp
 	i.restarts = int32(len(block)) - 4*(1+numRestarts)
 	i.numRestarts = numRestarts
 	i.globalSeqNum = globalSeqNum
 	i.ptr = unsafe.Pointer(&block[0])
 	i.data = block
-	i.fullKey = i.fullKey[:0]
+	if i.prefix != nil {
+		i.fullKey = append(i.fullKey[:0], i.prefix...)
+	} else {
+		i.fullKey = i.fullKey[:0]
+	}
 	i.val = nil
 	i.hideObsoletePoints = hideObsoletePoints
 	i.clearCache()
@@ -457,11 +465,11 @@ func (i *blockIter) init(
 //     ingested.
 //   - Foreign sstable iteration: globalSeqNum is always set.
 func (i *blockIter) initHandle(
-	cmp Compare, block bufferHandle, globalSeqNum uint64, hideObsoletePoints bool,
+	cmp Compare, block bufferHandle, globalSeqNum uint64, hideObsoletePoints bool, syntheticPrefix SyntheticPrefix,
 ) error {
 	i.handle.Release()
 	i.handle = block
-	return i.init(cmp, block.Get(), globalSeqNum, hideObsoletePoints)
+	return i.init(cmp, block.Get(), globalSeqNum, hideObsoletePoints, syntheticPrefix)
 }
 
 func (i *blockIter) invalidate() {
@@ -482,10 +490,11 @@ func (i *blockIter) isDataInvalidated() bool {
 
 func (i *blockIter) resetForReuse() blockIter {
 	return blockIter{
-		fullKey:   i.fullKey[:0],
-		cached:    i.cached[:0],
-		cachedBuf: i.cachedBuf[:0],
-		data:      nil,
+		fullKey:                i.fullKey[:0],
+		cached:                 i.cached[:0],
+		cachedBuf:              i.cachedBuf[:0],
+		firstUserKeyWithPrefix: i.firstUserKeyWithPrefix[:0],
+		data:                   nil,
 	}
 }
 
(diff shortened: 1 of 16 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3237#-:-NoYGZ4FBspq8JxMQjOl:bnfp4nl)* status: 0 of 16 files reviewed, 3 unresolved discussions (waiting on @dt, @msbutler, and @petermattis)

___
*[`sstable/block.go` line 414 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3237#-NoYEOgiBDcn6lDPvcQT:-NoYEOgiBDcn6lDPvcQU:b-ammmfe) ([raw file](https://github.com/cockroachdb/pebble/blob/59931fd1129cc00cace2460a4fe58b10dbcacb39/sstable/block.go#L414)):*
> ```Go
> 	}
> 	hideObsoletePoints bool
> 	SyntheticPrefix
> ```

[nit] we should avoid embedding unless we want to specifically inherit the methods
___
*[`sstable/block.go` line 644 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3237#-NoYFUbt9lMc29-qFeY4:-NoYFUbt9lMc29-qFeY5:b-9w1xxo) ([raw file](https://github.com/cockroachdb/pebble/blob/59931fd1129cc00cace2460a4fe58b10dbcacb39/sstable/block.go#L644)):*
> ```Go
> 		return base.CorruptionErrorf("pebble/table: invalid firstKey in block")
> 	}
> 	if i.SyntheticPrefix != nil {
> ```

Should we use a reusable buffer (similar to `fullKey`) to avoid this alloc when `restForReuse` is used?
___
*[`sstable/reader.go` line 174 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3237#-NoYG8sK7OdFJ_DEGt5s:-NoYG8sK7OdFJ_DEGt5t:b-8okduq) ([raw file](https://github.com/cockroachdb/pebble/blob/59931fd1129cc00cace2460a4fe58b10dbcacb39/sstable/reader.go#L174)):*
> ```Go
> }
> 
> type SyntheticPrefix []byte
> ```

[nit] will need some comments

sstable/block.go
```diff
@@ -633,6 +641,9 @@ func (i *blockIter) readFirstKey() error {
 		i.firstUserKey = nil
 		return base.CorruptionErrorf("pebble/table: invalid firstKey in block")
 	}
+	if i.SyntheticPrefix != nil {
```
msbutler, line 644: nit: in other call sites, you have added the prefix before the `DecodeInternalKey` call. I don't think it matters, but for consistency, perhaps move this up.

sstable/block.go
```diff
@@ -693,6 +704,17 @@ func (i *blockIter) SeekGE(key []byte, flags base.SeekGEFlags) (*InternalKey, ba
 		panic(errors.AssertionFailedf("invalidated blockIter used"))
 	}
 
+	searchKey := key
```
msbutler, line 707: There are a few places further down in this function which use `key` instead of `searchKey`, and I'm not sure it's correct. One [example](https://github.com/cockroachdb/pebble/blob/97ec6a5118e3c25368a963054ed0839a244178f9/sstable/block.go#L830).  To prevent this sort of error, perhaps we don't need to introduce `searchKey` as a new variable?

msbutler: > To prevent this sort of error, perhaps we don't need to introduce searchKey as a new variable?
I revoke this suggestion. We don't want to change the where the client provided `key` points to.

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3237)* status: 0 of 16 files reviewed, 8 unresolved discussions (waiting on @dt and @petermattis)

___
*[`sstable/block.go` line 647 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3237#-NpNP7Rj5z61W4J3suJD:-NpNP7Rk1rrWyNHlJ9eI:b7pg4av) ([raw file](https://github.com/cockroachdb/pebble/blob/330bf6112f04b066355245950e4d277f26fa3742/sstable/block.go#L647)):*
> ```Go
> 	}
> 	if i.prefix != nil {
> 		i.firstUserKey = append(append(i.firstUserKeyWithPrefix[:0], i.prefix...), i.firstUserKey...)
> ```

nit: the naming here is confusing. I thought `firstUserKey` was without the prefix, but it seems it is also with prefix, and `firstUserKeyWithPrefix` is really `firstUserKeyWithPrefixBacking`.
___
*[`sstable/block.go` line 704 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3237#-NpNPTaQ6sR5ydGcbTXY:-NpNPTaQ6sR5ydGcbTXZ:b-pnq13s) ([raw file](https://github.com/cockroachdb/pebble/blob/330bf6112f04b066355245950e4d277f26fa3742/sstable/block.go#L704)):*
> ```Go
> // SeekGE implements internalIterator.SeekGE, as documented in the pebble
> // package.
> func (i *blockIter) SeekGE(key []byte, flags base.SeekGEFlags) (*InternalKey, base.LazyValue) {
> ```

what's the effect of these changes on benchmarks that don't have a prefix?
___
*[`sstable/block.go` line 863 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3237#-NpNNsW99ml3rvB-z14S:-NpNNsW99ml3rvB-z14T:bgva23t) ([raw file](https://github.com/cockroachdb/pebble/blob/330bf6112f04b066355245950e4d277f26fa3742/sstable/block.go#L863)):*
> ```Go
> 		if i.prefix != nil {
> 			if !bytes.HasPrefix(key, i.prefix) {
> 				if i.cmp(i.prefix, key) < 0 {
> ```

where do we say that `Compare` needs to be defined for a prefix and not just user keys? For example `EngineKeyCompare` (the CRDB implementation of `Compare`) is explicitly looking for the separator between the prefix and the suffix.

RaduBerinde:
*[`table_cache.go` line 884 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3237#-NpfhEuf8XtXyLcSzaqT:-NpfhEuf8XtXyLcSzaqU:b-4hppjd) ([raw file](https://github.com/cockroachdb/pebble/blob/330bf6112f04b066355245950e4d277f26fa3742/table_cache.go#L884)):*
> ```Go
> 	pprof.Do(context.Background(), tableCacheLabels, func(context.Context) {
> 		var prefix []byte
> 		if meta.PrefixReplacement != nil && len(meta.PrefixReplacement.ContentPrefix) == 0 {
> ```

There is something subtle going on here. We only cache *physical* SST readers. We are pushing down the prefix replacement rules from the *virtual* SST into the physical reader (which can be shared among all virtual SSTs).

In principle we can currently have multiple virtual SSTs with the same backing SST that have different prefix replacement rules. I believe we don't need that functionality so we should require (and enforce) that all virtual SSTs with the same backing file have the same replacement rules. Once we do that, it's ok to push down the rules into the physical reader. (but we should still add an assertion that these things match when we use a cached physical reader).

This is an existing issue w.r.t `SmallestSeqNum/LargestSeqNum` below.

RaduBerinde:
*[`table_cache.go` line 884 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3237#-NpfhEuf8XtXyLcSzaqT:-Npfm3DJ-kpKvlht55Zc:b-aj1rmi) ([raw file](https://github.com/cockroachdb/pebble/blob/330bf6112f04b066355245950e4d277f26fa3742/table_cache.go#L884)):*

There is something subtle going on here. We only cache *physical* SST readers. We are pushing down the prefix replacement rules from the *virtual* SST into the physical reader (which can be shared among all virtual SSTs).

In principle we can currently have multiple virtual SSTs with the same backing SST that have different prefix replacement rules. I believe we don't need that functionality so we should require (and enforce) that all virtual SSTs with the same backing file have the same replacement rules. Once we do that, it's ok to push down the rules into the physical reader. (but we should still add an assertion that these things match when we use a cached physical reader).

This is an existing issue w.r.t `SmallestSeqNum/LargestSeqNum` below.
</blockquote></details>

Sounds like we can't assume all virtual SSTs will have the same suffix replacement rules (we would like these to be plumbed together). Maybe we can pass these rules only when we create iterators (from VirtualReader)?

msbutler:
moving pr to https://github.com/cockroachdb/pebble/pull/3350

PR #3239, opened 2024-01-22
db: add ViaBackingFileDownload option to DownloadSpan

dt:


dt:
There's some additional work to do here as discussed offline, but wanted to throw this up in case it is useful and/or if we want to merge it in a disabled state so that the CRDB-side or limiter etc work can be unblocked and then we come back to add the de-dupe and robust tests.

RaduBerinde (commented):
Left some comments. Like we discussed, it's ok to merge this for now even if we decide to switch to a provider-based operation later.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3239#-:-NpuWcaV9qsBT3b0F8AR:b-eubwe1)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`compaction.go` line 2759 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3239#-NpuUA9G-_6tADzwAXk1:-NpuUA9G-_6tADzwAXk2:b-r1ekgm) ([raw file](https://github.com/cockroachdb/pebble/blob/5cbce10cae483bd83897db8f0896eda0ed08b434/compaction.go#L2759)):*
> ```Go
> 	}
> 	metaCopy.FileNum = d.mu.versions.getNextFileNum()
> 	metaCopy.Virtual = false
> ```

This is hacky, we should do
```
if metaCopy.Virtual {
  // Even though this will be a virtual SST, we use the same number for the FileNum and the backing DiskFileNum.
  metaCopy.InitProviderBacking(base.DiskFileNum(metaCopy.FileNum))
} else {
  metaCopy.InitPhysicalBacking()
}
```
___
*[`compaction.go` line 2811 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3239#-NpuVpwaAZBbUmapwDmC:-NpuVpwaAZBbUmapwDmD:b-yd39d9) ([raw file](https://github.com/cockroachdb/pebble/blob/5cbce10cae483bd83897db8f0896eda0ed08b434/compaction.go#L2811)):*
> ```Go
> 				n := min(size-pos, int64(len(buf)))
> 				readErr := r.ReadAt(context.TODO(), buf, pos)
> 				if readErr != nil && readErr != io.EOF {
> ```

io.EOF isn't an ok error here, if we got the size wrong we'd be writing garbage at the end

PR #3241, opened 2024-01-23
db: have excises wait for EFOS below them, reducing EFOS waits

### LINKED ISSUE #3210: ingest: assign sequence number to excises and make EFOS infallible
Currently, excises as part of an `IngestAndExcise` call "always win" unless they overlap with a memtable at the time of excise, in which case the excise waits for the memtable to be flushed and then the excise is applied. If we assign a sequence number for excises, we can optionally add them to the flushable queue (see #2676), but alternatively an assigned sequence number also allows the excise to wait on any EFOS with snapshot sequence numbers below the excise to transition to a file-only snapshot before the excise is applied.

This issue is about introducing the seqnum assignment for excises and the linear search for any overlapping EFOS if there are any. #2676 is about the bigger change to move excises into the flushable queue, similar to flushable ingests.

itsbilal:
Previously, we had to make EventuallyFileOnlySnapshots (EFOS) wait for all relevant keys in their key bounds to be flushed, just in case an excise came in before that happened. This change avoids the need to do that by assigning a seqnum to every excise (which will be useful when we do #2676), and by flushing any memtables that overlap with the protected ranges of any EFOS that has a sequence number below the excise.

The reduction in waiting is desirable as actual conflicts between excises and EventuallyFileOnlySnapshots are rare, while users of EventuallyFileOnlySnapshots were previously stuck waiting or stuck doing excessive flushes at times when no such excises were happening.

Fixes #3210.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2083,14 +2083,6 @@ func (d *DB) flush1() (bytesFlushed uint64, err error) {
 				s = s.next
 				continue
 			}
-			if s.efos.excised.Load() {
-				// If a concurrent excise has happened that overlaps with one of the key
-				// ranges this snapshot is interested in, this EFOS cannot transition to
-				// a file-only snapshot as keys in that range could now be deleted. Move
-				// onto the next snapshot.
-				s = s.next
-				continue
-			}
 			currentVersion.Ref()
 
 			// NB: s.efos.transitionToFileOnlySnapshot could close s, in which
--- a/ingest.go
+++ b/ingest.go
@@ -1392,6 +1392,52 @@ func (d *DB) ingest(
 		d.mu.Lock()
 		defer d.mu.Unlock()
 
+		// Check if any of the currently-open EventuallyFileOnlySnapshots overlap
+		// in key ranges with the excise span. If so, we need to check for memtable
+		// overlaps with all bounds of that EventuallyFileOnlySnapshot in addition
+		// to the ingestion's own bounds too.
+
+		if exciseSpan.Valid() {
+			for s := d.mu.snapshots.root.next; s != &d.mu.snapshots.root; s = s.next {
+				if s.efos == nil {
+					continue
+				}
+				if base.Visible(seqNum, s.efos.seqNum, base.InternalKeySeqNumMax) {
+					// We only worry about snapshots older than the excise. Any snapshots
+					// created after the excise should see the excised view of the LSM
+					// anyway.
+					//
+					// Since we delay publishing the excise seqnum as visible until after
+					// the apply step, this case will never be hit in practice until we
+					// make excises flushable ingests.
+					continue
+				}
+				if invariants.Enabled {
+					if s.efos.hasTransitioned() {
+						panic("unexpected transitioned EFOS in snapshots list")
+					}
+				}
+				for i := range s.efos.protectedRanges {
+					if !s.efos.protectedRanges[i].OverlapsKeyRange(d.cmp, exciseSpan) {
+						continue
+					}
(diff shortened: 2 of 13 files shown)


jbowens (approved):
> Reviewed 8 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3241)* status: 8 of 13 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`ingest.go` line 2305 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3241#-NpV_vbGDCldBIzME1VY:-NpV_vbGDCldBIzME1VZ:b-vdp9ec) ([raw file](https://github.com/cockroachdb/pebble/blob/6f1189c1c00ca2a2c9767a4d49c27bcb56e22889/ingest.go#L2305)):*
> ```Go
> 		}
> 		// Check for any EventuallyFileOnlySnapshots that could be watching for
> 		// an excise on this span.
> ```

nit: should we update this comment to make it a little clearer we're guarding against an invariant violation
___
*[`snapshot.go` line 376 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3241#-NpViAPB0R3dRlei650d:-NpViAPB0R3dRlei650e:bf0prx9) ([raw file](https://github.com/cockroachdb/pebble/blob/6f1189c1c00ca2a2c9767a4d49c27bcb56e22889/snapshot.go#L376)):*
> ```Go
> //
> // Idempotent; can be called multiple times with no side effects.
> func (es *EventuallyFileOnlySnapshot) WaitForFileOnlySnapshot(
> ```

Do we still need `WaitForFileOnlySnapshot` (are we planning on ever calling it within Cockroach)? Could we now refactor EFOS to be just an optimization for snapshots that are bounded (eg, renaming `NewEventuallyFileOnlySnapshot` to `NewBoundedSnapshot`)?

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3241#-:-NpVoXxCDgbYGBWAD_ub:b-814g8k)* status: 8 of 13 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`ingest.go` line 2305 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3241#-NpV_vbGDCldBIzME1VY:-NpVoG8fBe6XxIutZEgB:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/6f1189c1c00ca2a2c9767a4d49c27bcb56e22889/ingest.go#L2305)):*

nit: should we update this comment to make it a little clearer we're guarding against an invariant violation
</blockquote></details>

Done.
___
*[`snapshot.go` line 376 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3241#-NpViAPB0R3dRlei650d:-NpVoHTjB0ELGhbIm9UD:b-qi5grz) ([raw file](https://github.com/cockroachdb/pebble/blob/6f1189c1c00ca2a2c9767a4d49c27bcb56e22889/snapshot.go#L376)):*

Do we still need `WaitForFileOnlySnapshot` (are we planning on ever calling it within Cockroach)? Could we now refactor EFOS to be just an optimization for snapshots that are bounded (eg, renaming `NewEventuallyFileOnlySnapshot` to `NewBoundedSnapshot`)?
</blockquote></details>

I've still kept it (and the tests around it) in case a caller ever needs to confirm it's not pinning any memory. However yes, I plan on removing the WaitForFileOnlySnapshot calls in Cockroach as they're unnecessary now. Down to explore either of those renames in a follow-up PR?

PR #3244, opened 2024-01-25
sstable: detect double Close on sstable iterators

RaduBerinde:
A flag indicates if an iterator is in the pool; if we call `Close()`
twice and the iterator has not been reused yet, we panic.

We add a `CloseHelper` which can absorb multiple `Close()` calls,
which is convenient to use with `defer`.

Also update the `InternalIterator` documentation to reflect the
reality that multiple `Close()` calls are not supported by all
implementations.

### DIFF
--- a/internal/base/close_helper.go
+++ b/internal/base/close_helper.go
@@ -1,0 +1,30 @@
+// Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package base
+
+import "io"
+
+// CloseHelper wraps an io.Closer in a wrapper that ignores extra calls to
+// Close. It is useful to ensure cleanup in error paths (using defer) without
+// double-closing.
+func CloseHelper(closer io.Closer) io.Closer {
+	return &closeHelper{
+		Closer: closer,
+	}
+}
+
+type closeHelper struct {
+	Closer io.Closer
+}
+
+// Close the underlying Closer, unless it was already closed.
+func (h *closeHelper) Close() error {
+	closer := h.Closer
+	if closer == nil {
+		return nil
+	}
+	h.Closer = nil
+	return closer.Close()
+}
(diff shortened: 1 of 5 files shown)


jbowens (approved):
nice

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3244#-:-Np5PsKI8R1pgsfJY-4W:bk4fctq)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/base/close_helper.go` line 10 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3244#-Np5P_eeCUN9AAnhjgrH:-Np5P_ef3u-tIUNed52l:b-7m7m3a) ([raw file](https://github.com/cockroachdb/pebble/blob/91e21dd944e725bbc33a91d1a9f8790225a313f6/internal/base/close_helper.go#L10)):*
> ```Go
> 
> // CloseHelper wraps an io.Closer in a wrapper that ignores extra calls to
> // Close. It is useful to ensure cleanup in erorr paths (using defer) without
> ```

nit: s/erorr/error/

jbowens (commented):
Maybe we should use `CloseHelper` when we construct the point iterator in this case: https://github.com/cockroachdb/pebble/blob/5b280af78f31791bf0e5701c9e649e73fe6f5c82/iterator.go#L307-L316

I don't remember the full context but maybe @itsbilal does.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3244#-:-Np5mkyCClbO6mUmvzTI:b-nkuvo7)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

RaduBerinde (commented):
TFTR!

I'll merge this for now, we can do that cleanup separately.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3244#-:-Np6Swyp3V3AhR8FRqvG:b-ae3ws5)* status: 4 of 5 files reviewed, all discussions resolved (waiting on @jbowens)

PR #3246, opened 2024-01-26
vfs: add disk write metrics to disk health checking vfs

CheranMahalingam:
This commit introduces a disk write stats collector to the disk health checking virtual file system. It makes it possible to aggregate the bytes written on a `Write([]byte)` call and identify the behaviour that is responsible. This commit will be followed up with changes to pass down the write category to the vfs layer and changes to export categorized write metrics at the cockroach layer.

Informs cockroachdb/cockroach#115434

### DIFF
--- a/options.go
+++ b/options.go
@@ -1221,7 +1221,7 @@ func (o *Options) WithFSDefaults() *Options {
 	if o.FS == nil {
 		o.FS = vfs.Default
 	}
-	o.FS, o.private.fsCloser = vfs.WithDiskHealthChecks(o.FS, 5*time.Second,
+	o.FS, o.private.fsCloser = vfs.WithDiskHealthChecks(o.FS, 5*time.Second, nil,
 		func(info vfs.DiskSlowInfo) {
 			o.EventListener.DiskSlow(info)
 		})
--- a/vfs/disk_health.go
+++ b/vfs/disk_health.go
@@ -5,10 +5,12 @@
 package vfs
 
 import (
+	"cmp"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
+	"slices"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -100,8 +102,53 @@ func (o OpType) String() string {
 	}
 }
 
-// diskHealthCheckingFile is a File wrapper to detect slow disk operations, and
-// call onSlowDisk if a disk operation is seen to exceed diskSlowThreshold.
+// DiskWriteCategory is a user-understandable string used to identify and aggregate
+// stats for disk writes. The prefix "pebble-" is reserved for internal Pebble categories.
+//
+// Some examples include, pebble-wal, pebble-memtable-flush, pebble-manifest and in the
+// Cockroach context includes, sql-spill, range-snapshot, node-log.
+type DiskWriteCategory string
+
+// WriteCategoryUnspecified denotes a disk write without a significant category.
+const WriteCategoryUnspecified = "unspecified"
+
+// DiskWriteStatsAggregate is an aggregate of the bytes written to disk for a given category.
+type DiskWriteStatsAggregate struct {
+	Category     DiskWriteCategory
+	BytesWritten uint64
+}
+
+// DiskWriteStatsCollector collects and aggregates disk write metrics per category.
+type DiskWriteStatsCollector struct {
+	mu       sync.Mutex
+	statsMap map[DiskWriteCategory]*atomic.Uint64
+}
+
+// NewDiskWriteStatsCollector instantiates a new DiskWriteStatsCollector.
+func NewDiskWriteStatsCollector() *DiskWriteStatsCollector {
+	return &DiskWriteStatsCollector{
+		statsMap: make(map[DiskWriteCategory]*atomic.Uint64),
+	}
+}
+
+// GetStats returns the aggregated metrics for all categories.
+func (d *DiskWriteStatsCollector) GetStats() []DiskWriteStatsAggregate {
+	var stats []DiskWriteStatsAggregate
+	d.mu.Lock()
+	for category, numBytes := range d.statsMap {
+		stats = append(stats, DiskWriteStatsAggregate{
+			Category:     category,
+			BytesWritten: numBytes.Load(),
+		})
+	}
+	d.mu.Unlock()
+	slices.SortFunc(stats, func(a, b DiskWriteStatsAggregate) int { return cmp.Compare(a.Category, b.Category) })
+	return stats
+}
+
+// diskHealthCheckingFile is a File wrapper to collect disk write stats,
+// detect slow disk operations, and call onSlowDisk if a disk operation
+// is seen to exceed diskSlowThreshold.
 //
 // This struct creates a goroutine (in startTicker()) that, at every tick
 // interval, sees if there's a disk operation taking longer than the specified
@@ -136,15 +183,38 @@ type diskHealthCheckingFile struct {
 	// across process boundaries.
 	lastWritePacked atomic.Uint64
 	createTimeNanos int64
+
+	// aggBytesWritten points to an atomic that aggregates the bytes written
+	// for files that belong to a specific DiskWriteCategory. This pointer is also stored in the
+	// DiskWriteStatsCollector for metric collection.
+	aggBytesWritten *atomic.Uint64
 }
+
+// diskHealthCheckingFile implements File.
+var _ File = (*diskHealthCheckingFile)(nil)
 
 // newDiskHealthCheckingFile instantiates a new diskHealthCheckingFile, with the
 // specified time threshold and event listener.
 func newDiskHealthCheckingFile(
 	file File,
 	diskSlowThreshold time.Duration,
+	category DiskWriteCategory,
+	statsCollector *DiskWriteStatsCollector,
 	onSlowDisk func(OpType OpType, writeSizeInBytes int, duration time.Duration),
 ) *diskHealthCheckingFile {
+	var bytesWritten *atomic.Uint64
+	if statsCollector != nil {
+		statsCollector.mu.Lock()
+		if aggStats, ok := statsCollector.statsMap[category]; !ok {
+			bytesWritten = new(atomic.Uint64)
+			statsCollector.statsMap[category] = bytesWritten
+		} else {
+			bytesWritten = aggStats
+		}
+		statsCollector.mu.Unlock()
+	} else {
+		bytesWritten = new(atomic.Uint64)
+	}
 	return &diskHealthCheckingFile{
 		file:              file,
 		onSlowDisk:        onSlowDisk,
@@ -153,6 +223,8 @@ func newDiskHealthCheckingFile(
 
 		stopper:         make(chan struct{}),
 		createTimeNanos: time.Now().UnixNano(),
+
+		aggBytesWritten: bytesWritten,
 	}
 }
 
@@ -215,14 +287,16 @@ func (d *diskHealthCheckingFile) Write(p []byte) (n int, err error) {
 	d.timeDiskOp(OpTypeWrite, int64(len(p)), func() {
 		n, err = d.file.Write(p)
 	}, time.Now().UnixNano())
+	d.aggBytesWritten.Add(uint64(n))
 	return n, err
 }
 
-// Write implements the io.WriterAt interface.
+// WriteAt implements the io.WriterAt interface.
 func (d *diskHealthCheckingFile) WriteAt(p []byte, ofs int64) (n int, err error) {
 	d.timeDiskOp(OpTypeWrite, int64(len(p)), func() {
 		n, err = d.file.WriteAt(p, ofs)
 	}, time.Now().UnixNano())
+	d.aggBytesWritten.Add(uint64(n))
 	return n, err
 }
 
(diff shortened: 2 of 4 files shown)


CheranMahalingam (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3246)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @jbowens)

___
*[`vfs/disk_health.go` line 123 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3246#-NpKuklo182KdbX6HHuH:-NpKuklo182KdbX6HHuI:b-ccmafi) ([raw file](https://github.com/cockroachdb/pebble/blob/a87702be81e5e233e867073ea4887b1d8dc05dc7/vfs/disk_health.go#L123)):*
> ```Go
> // DiskWriteStatsCollector collects and aggregates disk write metrics per category.
> type DiskWriteStatsCollector struct {
> 	mu       sync.Mutex
> ```

Wondering whether this will be a bottleneck since every single write to disk needs to acquire the mutex.

aadityasondhi (changes requested):
Great work, this is looking good! A quick first pass review with some comments. I will let other chime in as well.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3246#-:-NpKw5zN9BwSXnxeecI4:b-rf70zq)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @CheranMahalingam and @jbowens)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3246#-NpKv8xbCFFweWh9BtxU):*
Add a description for the commit. You could copy what you wrote in the PR here. Also link it to the originating issue "Informs #115434". 
___
*[`vfs/disk_health.go` line 123 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3246#-NpKuklo182KdbX6HHuH:-NpKv_73DMxe9zis7iso:b-w6hylw) ([raw file](https://github.com/cockroachdb/pebble/blob/a87702be81e5e233e867073ea4887b1d8dc05dc7/vfs/disk_health.go#L123)):*

Wondering whether this will be a bottleneck since every single write to disk needs to acquire the mutex.
</blockquote></details>

Yeah this is non-ideal. We can look into some cpu/mutex profiles to verify. But just based on how this is used, we should know all the categories ahead of time, right? We can create the map with all of them in `NewDiskWriteStatsCollector`. Since bytes are `uint64`, we can use atomics. This design should then be lock free, especially on the write-path.

For reads, it depends on if we care about all category states to be read atomically. If we do, then we will still need a lock there. But this function will be called a lot less, so it should be ok.

jbowens (commented):
Looking great :sunglasses: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3246#-:-NpL2FYz-qI1_G2Nxjup:b-5hum8n)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @aadityasondhi and @CheranMahalingam)

___
*[`vfs/disk_health.go` line 123 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3246#-NpKuklo182KdbX6HHuH:-NpL-tr3Ee8JtZE9UpsV:b-91gcd) ([raw file](https://github.com/cockroachdb/pebble/blob/a87702be81e5e233e867073ea4887b1d8dc05dc7/vfs/disk_health.go#L123)):*
<details><summary><i>Previously, aadityasondhi (Aaditya Sondhi) wrote…</i></summary><blockquote>

Yeah this is non-ideal. We can look into some cpu/mutex profiles to verify. But just based on how this is used, we should know all the categories ahead of time, right? We can create the map with all of them in `NewDiskWriteStatsCollector`. Since bytes are `uint64`, we can use atomics. This design should then be lock free, especially on the write-path.

For reads, it depends on if we care about all category states to be read atomically. If we do, then we will still need a lock there. But this function will be called a lot less, so it should be ok.
</blockquote></details>

I agree with Aaditya, we should use an atomic. Since we're passing the category down during file creation, I think we should:

1. Make `DiskWriteStatsCollector.statsMap` a `map[DiskWriteCategory]*atomic.Uint64`.
2. Use `mu` to guard the `statsMap` look up in order to grab the `*atomic.Uint64` pointer (or populate it) during file creation.
3. In write paths, atomically increment the stats using `(*atomic.Uint64).Add`.

Acquiring the mutex during file creation should be fine.

CheranMahalingam:
Sounds good. Just to confirm the specifics of the improved map design, are my assumptions correct?
1. Since writes to the map are very frequent we don't want to have to acquire a lock and instead do,
```
func (d *DiskWriteStatsCollector) IncrementStats(category DiskWriteCategory, numBytes uint64) {
	d.StatsMap[category].Add(numBytes)
}
```
2. To enable part 1 we need to initialize the map whenever a new category is introduced with zero,
```
	statsCollector.Mu.Lock()
	if _, ok := statsCollector.StatsMap[category]; !ok {
		statsCollector.StatsMap[category] = new(atomic.Uint64)
	}
	statsCollector.Mu.Unlock()
```
3. Reads are less frequent than increments but since someone could create a new category at the same time as we copy all the metrics we need to acquire the lock,
```
	var stats []DiskWriteStatsAggregate
	d.Mu.Lock()
	for category, numBytes := range d.StatsMap {
		stats = append(stats, DiskWriteStatsAggregate{
			Category:     category,
			BytesWritten: numBytes.Load(),
		})
	}
	d.Mu.Unlock()
```

If this is correct my only worry is the safety of the `IncrementStats` call that could panic if the category doesn't exist in the map. Is there a way to resolve this without slowing the program down or is this concern insignificant?

CheranMahalingam:
Nevermind, ignore my concern, the IncrementStats method should be made private.

CheranMahalingam (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3246)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @CheranMahalingam)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3246#-NpKv8xbCFFweWh9BtxU):*
<details><summary><i>Previously, aadityasondhi (Aaditya Sondhi) wrote…</i></summary><blockquote>

Add a description for the commit. You could copy what you wrote in the PR here. Also link it to the originating issue "Informs #115434".
</blockquote></details>

Done.

jbowens (commented):
> Reviewed 1 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3246)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @aadityasondhi and @CheranMahalingam)

___
*[`vfs/disk_health.go` line 215 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3246#-NpPhgsP12BMMSD80lxF:-NpPhgsP12BMMSD80lxG:b-yaxjqk) ([raw file](https://github.com/cockroachdb/pebble/blob/e2d73acffae71825ec0791c17cdd08a3cc997d18/vfs/disk_health.go#L215)):*
> ```Go
> 		if _, ok := statsCollector.StatsMap[category]; !ok {
> 			statsCollector.StatsMap[category] = new(atomic.Uint64)
> 		}
> ```

here, we actually want to grab the value of `statsCollector.StatsMap[category]` and then down below stash it on the `diskHealthCheckingFile`. I think we should then be able to remove the `category` and `statsCollector` from the `diskHealthCheckingFile` because they'll be unused.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3246)* status: 1 of 4 files reviewed, 4 unresolved discussions (waiting on @aadityasondhi and @CheranMahalingam)

___
*[`vfs/disk_health.go` line 124 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3246#-NpPjbAl7wA4kpJrq4Fm:-NpPjbAl7wA4kpJrq4Fn:bfyfmo6) ([raw file](https://github.com/cockroachdb/pebble/blob/e2d73acffae71825ec0791c17cdd08a3cc997d18/vfs/disk_health.go#L124)):*
> ```Go
> type DiskWriteStatsCollector struct {
> 	Mu       sync.Mutex
> 	StatsMap map[DiskWriteCategory]*atomic.Uint64
> ```

let's un-export these fields (by lowercasing them)
___
*[`vfs/disk_health.go` line 139 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3246#-NpPiMmB1AGtOvLWSjkT:-NpPiMmB1AGtOvLWSjkU:b-v36upx) ([raw file](https://github.com/cockroachdb/pebble/blob/e2d73acffae71825ec0791c17cdd08a3cc997d18/vfs/disk_health.go#L139)):*
> ```Go
> // a member of a diskHealthCheckingFile with the specified category.
> func (d *DiskWriteStatsCollector) incrementStats(category DiskWriteCategory, numBytes uint64) {
> 	d.StatsMap[category].Add(numBytes)
> ```

This map lookup will still race with other goroutines that are mutating the map to insert other categories. This is why we want to load the pointer to the `atomic.Uint64` while we hold the mutex (see my other comment). Once we've retrieved the appropriate pointer for the category, we don't care if other goroutines mutate the map, because we're never going to read it again.

CheranMahalingam (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3246)* status: 1 of 4 files reviewed, 4 unresolved discussions (waiting on @aadityasondhi and @jbowens)

___
*[`vfs/disk_health.go` line 124 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3246#-NpPjbAl7wA4kpJrq4Fm:-NpPulzVE_VScBGBx7qe:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e2d73acffae71825ec0791c17cdd08a3cc997d18/vfs/disk_health.go#L124)):*

let's un-export these fields (by lowercasing them)
</blockquote></details>

Done.
___
*[`vfs/disk_health.go` line 139 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3246#-NpPiMmB1AGtOvLWSjkT:-NpPz2GaEi_wIOVf7y3L:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e2d73acffae71825ec0791c17cdd08a3cc997d18/vfs/disk_health.go#L139)):*

This map lookup will still race with other goroutines that are mutating the map to insert other categories. This is why we want to load the pointer to the `atomic.Uint64` while we hold the mutex (see my other comment). Once we've retrieved the appropriate pointer for the category, we don't care if other goroutines mutate the map, because we're never going to read it again.
</blockquote></details>

Done.
___
*[`vfs/disk_health.go` line 215 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3246#-NpPhgsP12BMMSD80lxF:-NpPz3EH5v2lMABysfQ1:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e2d73acffae71825ec0791c17cdd08a3cc997d18/vfs/disk_health.go#L215)):*

here, we actually want to grab the value of `statsCollector.StatsMap[category]` and then down below stash it on the `diskHealthCheckingFile`. I think we should then be able to remove the `category` and `statsCollector` from the `diskHealthCheckingFile` because they'll be unused.
</blockquote></details>

Done.

jbowens (commented):
Looks great, just one last comment about an allocation

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3246#-:-NpR0BCI9Vyi3OJu9Cz8:bx9v8tv)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @aadityasondhi and @CheranMahalingam)

___
*[`vfs/disk_health.go` line 205 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3246#-NpQzSgd8dW4nZ0us6O0:-NpQzSgd8dW4nZ0us6O1:bd93r0o) ([raw file](https://github.com/cockroachdb/pebble/blob/7c84591d6e90f39197d56e696d4b03831618c6a6/vfs/disk_health.go#L205)):*
> ```Go
> 	onSlowDisk func(OpType OpType, writeSizeInBytes int, duration time.Duration),
> ) *diskHealthCheckingFile {
> 	bytesWritten := new(atomic.Uint64)
> ```

this will suffer an allocation even if we already have one for a category. can we move this into the `!ok` case down below (and declare bytesWritten here with: `var bytesWritten *atomic.Uint64`)

CheranMahalingam (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3246)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @aadityasondhi and @jbowens)

___
*[`vfs/disk_health.go` line 205 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3246#-NpQzSgd8dW4nZ0us6O0:-NpR5YfY9YlyRN_Y6aWv:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/7c84591d6e90f39197d56e696d4b03831618c6a6/vfs/disk_health.go#L205)):*

this will suffer an allocation even if we already have one for a category. can we move this into the `!ok` case down below (and declare bytesWritten here with: `var bytesWritten *atomic.Uint64`)
</blockquote></details>

Done.

jbowens (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3246#-:-NpUq-4Z0NVMJUjiQ8dx:b-uwkowd)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @aadityasondhi)

aadityasondhi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3246#-:-Npv2sg56Qau5X-BxC42:bnfp4nl)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @CheranMahalingam and @jbowens)

Issue #2863, opened 2023-08-30
db: consider reworking rangedel, mergingIter and levelIter interactions


jbowens:
Currently, there's an intricate coordination dance performed between `mergingIter` and `levelIter` in order to make range deletions work. This dance has several complexities that make it fragile and difficult to grok. The introduction of range-key masking and "truthful" block-property filters also forces us into some [awkward contortions](https://github.com/cockroachdb/pebble/blob/1a45921accf7c4422d7214a8e6315c374d0725c6/level_iter.go#L126-L135) to deal with the fact that a file's point iterator and range deletion iterator may become exhausted at different times.

I suspect we could significantly clean things up by using the `keyspan.InterleavingIter` within the `levelIter` implementation. It would work something like:

- If a file has range deletions, the `levelIter` initializes a `keyspan.InterleavingIter` with the file's point iterator and rangedel iterator. It configures the `InterleavingIter` to interleave both start and end boundaries of range deletions.
- The `levelIter` keeps a file open until it exhausts the `InterleavingIter`. Since range deletions' start/end boundaries are interleaved, this ensures that we don't prematurely close the range deletion iterator during iteration. We can remove the code that manufactures "ignorable boundary keys." The interleaving iterator handles interleaving the rangedel end boundary if necessary.
- The merging iterator no longer keeps a direct handle on a file's range deletion iterator. Instead, every `mergingIterLevel` has a `getRangeDel() *keyspan.Span` func that returns the range deletion covering that level's iterator's current position, if any. When a `levelIter` is setup, this is set to `levelIter.Span` which in turn calls `InterelavingIter.Span`.
- The `filteredIter` interface and related tracking within the sstable iterator is removed. It's no longer necessary because range deletions are unconditionally interleaved, which ensures we never miss a range deletion due to block-property filters excluding point keys that exist beyond the smallest/largest range deletion bound.
- The merging iterator iteration is adjusted to skip over range deletion boundaries that are interleaved and rise to the top of the heap, much like the existing code skips over the "ignorable boundary keys".
- The mechanisms of applying range deletions also change. The `mergingIterLevel` no longer buffers a tombstone. It instead calls `getRangeDel()` which accesses the `InterleavingIter`'s buffered tombstone instead. We remove `mergingIter.init[Min,Max]RangeDelIters`, because the `InterleavingIter` is transparently ensuring the range deletion iterators are appropriately positioned.

[...]

PR #3247, opened 2024-01-26
db: refactor table cache newIters

jbowens:
This commit refactors the interface of the table cache `newIters` method to support creation of any of a sstable's iterators, including its point iterator, range deletion iterator and range key iterator. This is motivated by the refactor planned in #2863, which in turn requires refactoring some of the newIters call sites that determine whether ingested sstables overlap existing tables.

There's also a minor benefit that a compaction setup no longer needs to open point iterators and then immediately close them when initializing the range deletion iterators.

```
goos: linux
goarch: amd64
              │ old-NewItersAlloc.txt │       new-NewItersAlloc.txt        │
              │        sec/op         │   sec/op     vs base               │
NewItersAlloc             427.0n ± 1%   412.1n ± 1%  -3.47% (p=0.000 n=10)

              │ old-NewItersAlloc.txt │     new-NewItersAlloc.txt      │
              │         B/op          │    B/op     vs base            │
NewItersAlloc              0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal

              │ old-NewItersAlloc.txt │     new-NewItersAlloc.txt      │
              │       allocs/op       │ allocs/op   vs base            │
NewItersAlloc              0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1427,38 +1427,22 @@ func (c *compaction) newRangeDelIter(
 	bytesIterated *uint64,
 ) (keyspan.FragmentIterator, io.Closer, error) {
 	opts.level = l
-	iter, rangeDelIter, err := newIters(context.Background(), f.FileMetadata,
+	iterSet, err := newIters(context.Background(), f.FileMetadata,
 		&opts, internalIterOpts{
 			bytesIterated: &c.bytesIterated,
 			bufferPool:    &c.bufferPool,
-		})
+		}, iterRangeDeletions)
 	if err != nil {
 		return nil, nil, err
-	}
-	// TODO(peter): It is mildly wasteful to open the point iterator only to
-	// immediately close it. One way to solve this would be to add new
-	// methods to tableCache for creating point and range-deletion iterators
-	// independently. We'd only want to use those methods here,
-	// though. Doesn't seem worth the hassle in the near term.
-	if err = iter.Close(); err != nil {
-		if rangeDelIter != nil {
-			err = errors.CombineErrors(err, rangeDelIter.Close())
-		}
-		return nil, nil, err
-	}
-	if rangeDelIter == nil {
+	} else if iterSet.rangeDeletion == nil {
 		// The file doesn't contain any range deletions.
 		return nil, nil, nil
 	}
-
 	// Ensure that rangeDelIter is not closed until the compaction is
 	// finished. This is necessary because range tombstone processing
 	// requires the range tombstones to be held in memory for up to the
 	// lifetime of the compaction.
-	closer := rangeDelIter
-	rangeDelIter = noCloseIter{rangeDelIter}
-
-	return rangeDelIter, closer, nil
+	return noCloseIter{iterSet.rangeDeletion}, iterSet.rangeDeletion, nil
 }
 
 func (c *compaction) String() string {
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -2891,9 +2891,9 @@ func TestCompactionCheckOrdering(t *testing.T) {
 				}
 
 				newIters := func(
-					_ context.Context, _ *manifest.FileMetadata, _ *IterOptions, _ internalIterOpts,
-				) (internalIterator, keyspan.FragmentIterator, error) {
-					return &errorIter{}, nil, nil
+					_ context.Context, _ *manifest.FileMetadata, _ *IterOptions, _ internalIterOpts, _ iterKinds,
+				) (iterSet, error) {
+					return iterSet{point: &errorIter{}}, nil
 				}
 				result := "OK"
 				_, err := c.newInputIter(newIters, nil, nil)
--- a/db.go
+++ b/db.go
@@ -3076,42 +3076,33 @@ func (d *DB) checkVirtualBounds(m *fileMetadata) {
 		return
 	}
 
+	iters, err := d.newIters.all(context.TODO(), m, nil, internalIterOpts{})
+	if err != nil {
+		panic(errors.Wrap(err, "pebble: error creating iterators"))
+	}
+	defer iters.CloseAll()
+
 	if m.HasPointKeys {
-		pointIter, rangeDelIter, err := d.newIters(context.TODO(), m, nil, internalIterOpts{})
-		if err != nil {
-			panic(errors.Wrap(err, "pebble: error creating point iterator"))
-		}
-
-		defer pointIter.Close()
-		if rangeDelIter != nil {
-			defer rangeDelIter.Close()
-		}
-
-		pointKey, _ := pointIter.First()
-		var rangeDel *keyspan.Span
-		if rangeDelIter != nil {
-			rangeDel, err = rangeDelIter.First()
-			if err != nil {
-				panic(err)
-			}
-		}
+		pointIter := iters.Point()
+		rangeDelIter := iters.RangeDeletion()
 
 		// Check that the lower bound is tight.
+		pointKey, _ := pointIter.First()
+		rangeDel, err := rangeDelIter.First()
+		if err != nil {
+			panic(err)
+		}
 		if (rangeDel == nil || d.cmp(rangeDel.SmallestKey().UserKey, m.SmallestPointKey.UserKey) != 0) &&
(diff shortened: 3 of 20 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3247#-:-NpQtUahCqLV7I6vIxor:bg7igkv)* status: all files reviewed, 3 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`table_cache.go` line 166 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3247#-NpQprnh7uoh5mkT5jre:-NpQprnh7uoh5mkT5jrf:b59mlni) ([raw file](https://github.com/cockroachdb/pebble/blob/f0b9dd6b2688e3e7795ab50b4c59843d4f7a7dfa/table_cache.go#L166)):*
> ```Go
> 	}
> 	if iter == nil {
> 		iter = emptyKeyspanIter
> ```

if the caller is going to place this in an `iterSet` can't we just return nil, since the `iterSet` handles the nil => non-nil empty iter transformation itself?
___
*[`table_cache.go` line 546 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3247#-NpQs34V2iAKoqTdHfNh:-NpQs34V2iAKoqTdHfNi:b-ddmu0d) ([raw file](https://github.com/cockroachdb/pebble/blob/f0b9dd6b2688e3e7795ab50b4c59843d4f7a7dfa/table_cache.go#L546)):*
> ```Go
> 			// Return an empty iterator. This iterator has no mutable state, so
> 			// using a singleton is fine.
> 			return filteredAll, err
> ```

same question about letting `iterSet` handle the nil iter case.
___
*[`table_cache.go` line 1286 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3247#-NpQoGs4-8cw5j0RkrzR:-NpQoGs4-8cw5j0RkrzS:b-o6rh09) ([raw file](https://github.com/cockroachdb/pebble/blob/f0b9dd6b2688e3e7795ab50b4c59843d4f7a7dfa/table_cache.go#L1286)):*
> ```Go
> // bitwise-OR iterPointKeys, iterRangeDeletions and/or iterRangeKeys together to
> // represent a set of desired iterator kinds.
> type iterKinds int8
> ```

nit: how about uint8, given it is a bitmap.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3247)* status: all files reviewed, 6 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`level_iter.go` line 50 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3247#-NpR-Pp5CT4BMryVTFCk:-NpR-Pp6-1gwme23fc0k:b-6bmute) ([raw file](https://github.com/cockroachdb/pebble/blob/f0b9dd6b2688e3e7795ab50b4c59843d4f7a7dfa/level_iter.go#L50)):*
> ```Go
> // all returns an iterator set containing iterators over all the keys that meta
> // contains.
> func (f tableNewIters) all(
> ```

This method on a function is a bit dubious. I'd consider adding an `iterAll` constant and call it directly with that instead.
___
*[`level_iter.go` line 54 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3247#-NpR-G-Y8oV2raL2ujcs:-NpR-G-Y8oV2raL2ujct:b-iei5fb) ([raw file](https://github.com/cockroachdb/pebble/blob/f0b9dd6b2688e3e7795ab50b4c59843d4f7a7dfa/level_iter.go#L54)):*
> ```Go
> ) (iterSet, error) {
> 	var kinds iterKinds
> 	if file.HasPointKeys {
> ```

[nit] Feels like this optimization belongs inside the implementation, not here
___
*[`table_cache.go` line 1248 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3247#-NpQzlrHEHIw8AiaU3AS:-NpQzlrHEHIw8AiaU3AT:bxfh1lq) ([raw file](https://github.com/cockroachdb/pebble/blob/f0b9dd6b2688e3e7795ab50b4c59843d4f7a7dfa/table_cache.go#L1248)):*
> ```Go
> }
> 
> // RangeDeletion returns the contained range deletion iterator. If there is no
> ```

[nit] Will something like `HasRangeDeletion() bool` be useful for fast paths?

jbowens (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3247#-:-NpV7wLV8VP1o1mWVFtt:bz5mair)* status: 16 of 21 files reviewed, 4 unresolved discussions (waiting on @itsbilal, @RaduBerinde, and @sumeerbhola)

___
*[`level_iter.go` line 50 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3247#-NpR-Pp5CT4BMryVTFCk:-NpV0kCH-rRsEf_KIzD5:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/f0b9dd6b2688e3e7795ab50b4c59843d4f7a7dfa/level_iter.go#L50)):*

This method on a function is a bit dubious. I'd consider adding an `iterAll` constant and call it directly with that instead.
</blockquote></details>

Done.
___
*[`level_iter.go` line 54 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3247#-NpR-G-Y8oV2raL2ujcs:-NpV0lDCEO0lr7w27hFY:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/f0b9dd6b2688e3e7795ab50b4c59843d4f7a7dfa/level_iter.go#L54)):*

[nit] Feels like this optimization belongs inside the implementation, not here
</blockquote></details>

Done.
___
*[`table_cache.go` line 166 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3247#-NpQprnh7uoh5mkT5jre:-NpV2ZgT7u4tr7iLRchs:b-l9k1mx) ([raw file](https://github.com/cockroachdb/pebble/blob/f0b9dd6b2688e3e7795ab50b4c59843d4f7a7dfa/table_cache.go#L166)):*

if the caller is going to place this in an `iterSet` can't we just return nil, since the `iterSet` handles the nil => non-nil empty iter transformation itself?
</blockquote></details>

I removed this entire func; this was a remaining top-level table cache entry point that did not use `iterSet` in order to adapt to the function signature of `keyspan.TableNewSpanIter`. I replaced it with a new func `tableNewRangeKeyIter` that takes a `newIters` and returns a function with the appropriate signature.
___
*[`table_cache.go` line 546 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3247#-NpQs34V2iAKoqTdHfNh:-NpV3-Ec94NuGGO7J_fP:bp4ydg8) ([raw file](https://github.com/cockroachdb/pebble/blob/f0b9dd6b2688e3e7795ab50b4c59843d4f7a7dfa/table_cache.go#L546)):*

same question about letting `iterSet` handle the nil iter case.
</blockquote></details>

Added a comment here. This one is necessary because of the MaybeFilteredKeys() interface, but it will be go away with #2863.
___
*[`table_cache.go` line 1248 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3247#-NpQzlrHEHIw8AiaU3AS:-NpV3Srj1XB8MKt-wRzv:b4tj2zz) ([raw file](https://github.com/cockroachdb/pebble/blob/f0b9dd6b2688e3e7795ab50b4c59843d4f7a7dfa/table_cache.go#L1248)):*

[nit] Will something like `HasRangeDeletion() bool` be useful for fast paths?
</blockquote></details>

Hm, yeah, I think it might—I'll defer it for now but after #2863 we can look at the usages and adapt. Left a TODO
___
*[`table_cache.go` line 1286 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3247#-NpQoGs4-8cw5j0RkrzR:-NpV3R039bkBYBnRU8FQ:b-48vilg) ([raw file](https://github.com/cockroachdb/pebble/blob/f0b9dd6b2688e3e7795ab50b4c59843d4f7a7dfa/table_cache.go#L1286)):*

nit: how about uint8, given it is a bitmap.
</blockquote></details>

Good call, done.

Issue #3230, opened 2024-01-18
db: WAL failover to deal with transient unavailability


sumeerbhola:
We see transient write unavailability of block devices in the cloud (< 60s) that are sometimes detected as disk stalls resulting in node crashes. Whether the node crashes or not, this negatively impacts the user workload. Read have not been observed to stall in this manner, and additionally reads can often be satisfied using the Pebble block cache, or the OS page cache. 

WAL failover relies on more than one block devices configured for the node, say two block devices and two Pebble DBs. The WAL for one Pebble DB can temporarily failover to the block device of the other. Flushes and compactions will stall, but most workloads are writing at a rate that we can afford to buffer 60s of data in memtables. More details in https://docs.google.com/document/d/1vAsftzyPG-kDy-A2Ic1fZeKd4OKJIf6N7KvNXRpDAFA/edit#heading=h.8n1r6sehoqgk (internal doc).

Also see CRDB-35401

Starting a list to track the remaining Pebble work:
- [x] Integrate wal.failoverManager with Pebble
- [x] Metrics: cumulative number of switches; cumulative duration writing to primary and secondary
- [x] Config: Make failover threshold (`FailoverOptions.UnhealthyOperationLatencyThreshold`) dynamically changeable
- [ ] #3463 
- [x] Testing: metamorphic/randomized testing of WAL failover
- [x] [Optional] make excises flushable ingests (to not stall the commit pipeline now that we excise for range snapshots) https://github.com/cockroachdb/pebble/issues/3335
- [x] [Misc CockroachDB, Optional] AC: if storage AC is active, it will see a drop in compaction throughput out of L0 and start limiting new writes. Could allow unlimited tokens while writing to secondary?

PR #3249, opened 2024-01-29
record: LogWriter changes for use in failover WAL writing

sumeerbhola:
The syncQueue is generalized via the pendingSyncs and helper interfaces. pendingSyncs has two implementations, one for the standalone case where there is a real queue (backed by syncQueue), and one for the failover case where the LogWriter only tracks the latest record that has requested a sync.

LogWriterConfig.ExternalSyncQueueCallback is introduced for the failover case to inform the caller when a particular record index is synced. This callback is also utilized on close, to inform the caller of the tail of the records that have now been synced (even if they did not originally request to be synced).

SyncRecordGeneralized and CloseWithLastQueuedRecord are the new methods that will be used in the failover case.

Informs #3230

Informs CRDB-35401

### DIFF
--- a/record/log_writer.go
+++ b/record/log_writer.go
@@ -150,12 +150,14 @@ func (q *syncQueue) pop(head, tail uint32, err error, queueSemChan chan struct{}
 		*slot.err = err
 		slot.wg = nil
 		slot.err = nil
-		// We need to bump the tail count before signalling the wait group as
-		// signalling the wait group can trigger release a blocked goroutine which
-		// will try to enqueue before we've "freed" space in the queue.
+		// We need to bump the tail count before releasing the queueSemChan
+		// semaphore as releasing the semaphore can cause a blocked goroutine to
+		// acquire the semaphore and enqueue before we've "freed" space in the
+		// queue.
 		q.headTail.Add(1)
 		wg.Done()
-		// Is always non-nil in production.
+		// Is always non-nil in production, unless using wal package for WAL
+		// failover.
 		if queueSemChan != nil {
 			<-queueSemChan
 		}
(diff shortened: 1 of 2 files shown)


jbowens (approved):
:lgtm:

> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3249#-:-NpRYb0w7ipdRXK-Rm-6:bnfp4nl)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`record/log_writer.go` line 430 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3249#-NpQj77X4VIC5Z_mVcpl:-NpQj77X4VIC5Z_mVcpm:bqrrh3y) ([raw file](https://github.com/cockroachdb/pebble/blob/c9e6cfc917f89509e706a2ba97a2e562e6662166/record/log_writer.go#L430)):*
> ```Go
> 	// updated to reflect flusher.err when a block is full and getting enqueued.
> 	// Therefore, when flusher.err is reflected in the return value of
> 	// SyncRecord* is delayed. On close, it is set to errClosedWriter to inform
> ```

I think this sentence starting with "Therefore," is incomplete or mangled (and I didn't understand it)

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3249#-:-NpS7NVaBUvL_QJGdVn3:b-814g8k)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`record/log_writer.go` line 430 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3249#-NpQj77X4VIC5Z_mVcpl:-NpS61gc7XSmDvwMvtoc:bq1txfm) ([raw file](https://github.com/cockroachdb/pebble/blob/c9e6cfc917f89509e706a2ba97a2e562e6662166/record/log_writer.go#L430)):*

I think this sentence starting with "Therefore," is incomplete or mangled (and I didn't understand it)
</blockquote></details>

I changed this to
```
// Therefore, there is a lag between when flusher.err has a non-nil error, and  
// when that non-nil error is reflected in LogWriter.err.
```

PR #3253, opened 2024-01-30
metamorphic: consolidate key manager per-object structures

RaduBerinde:
This commit merges the `byObj`, `byObjMap` and `bounds` maps into a
single map to a consolidated `*objKeyMeta`.

We no longer store keys in two structures; we only use a map.  We
simplify and improve `mergeKeysInto` to use the target map instead of
sorting and merging all keys; this should be much faster when the
target is the database.

### DIFF
--- a/metamorphic/generator.go
+++ b/metamorphic/generator.go
@@ -1009,14 +1009,14 @@ func (g *generator) iterSeekGEWithLimit(iterID objID) {
 	})
 }
 
-func (g *generator) randKeyToReadWithinBounds(lower, upper []byte, readerID objID) []*keyMeta {
+// inRangeKeys returns all keys in the range [lower, upper) associated with the
+// given object.
+func (g *generator) inRangeKeys(lower, upper []byte, o objID) []*keyMeta {
 	var inRangeKeys []*keyMeta
-	for _, keyMeta := range g.keyManager.byObj[readerID] {
-		posKey := keyMeta.key
-		if g.cmp(posKey, lower) < 0 || g.cmp(posKey, upper) >= 0 {
-			continue
+	for _, keyMeta := range g.keyManager.sortedKeysForObj(o) {
+		if g.cmp(keyMeta.key, lower) >= 0 && g.cmp(keyMeta.key, upper) < 0 {
+			inRangeKeys = append(inRangeKeys, keyMeta)
 		}
-		inRangeKeys = append(inRangeKeys, keyMeta)
 	}
 	return inRangeKeys
 }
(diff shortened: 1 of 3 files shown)


RaduBerinde:
Actually, let me work on this some more and reopen.

RaduBerinde:
Ok, ready for review.

jbowens (approved):
Nice :lgtm:

> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3253#-:-NpVK9GMCxugLqbUjpi1:b-qcmwly)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`metamorphic/key_manager.go` line 235 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3253#-NpVI5-F3E8dLVQmDAqt:-NpVI5-F3E8dLVQmDAqu:b-oa5er9) ([raw file](https://github.com/cockroachdb/pebble/blob/689f96754fb8c3c97abfd3a9322affb0557c7b77/metamorphic/key_manager.go#L235)):*
> ```Go
> 	}
> 	slices.SortFunc(res, func(a, b *keyMeta) int {
> 		return bytes.Compare(a.key, b.key)
> ```

Using `bytes.Compare` doesn't seem to be a new behavior, but it seems like this should be a problem for `generator.randKeyToReadWithinBounds` which generate keys to fall within bounds according to `base.Compare`.

Ah, I see, `randKeyToReadWithinBounds` isn't actually making use of the sorted property. Maybe we should thread the `Compare` func through and always sort according to it to avoid confusion?

RaduBerinde:
*[`metamorphic/key_manager.go` line 235 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3253#-NpVI5-F3E8dLVQmDAqt:-NpVKmDX-PP9dY42BZPO:bt96px2) ([raw file](https://github.com/cockroachdb/pebble/blob/689f96754fb8c3c97abfd3a9322affb0557c7b77/metamorphic/key_manager.go#L235)):*

Using `bytes.Compare` doesn't seem to be a new behavior, but it seems like this should be a problem for `generator.randKeyToReadWithinBounds` which generate keys to fall within bounds according to `base.Compare`.

Ah, I see, `randKeyToReadWithinBounds` isn't actually making use of the sorted property. Maybe we should thread the `Compare` func through and always sort according to it to avoid confusion?
</blockquote></details>

Good point, I will work on that. I believe that currently we only use this to avoid nondeterminism (but I agree we might as well do it right). In the previous use, we sorted just to merge-join, the particular ordering didn't actually matter.

RaduBerinde:
*[`metamorphic/key_manager.go` line 235 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3253#-NpVI5-F3E8dLVQmDAqt:-NpW6MCtA54XzSJgsmO3:bprgsbe) ([raw file](https://github.com/cockroachdb/pebble/blob/689f96754fb8c3c97abfd3a9322affb0557c7b77/metamorphic/key_manager.go#L235)):*

Good point, I will work on that. I believe that currently we only use this to avoid nondeterminism (but I agree we might as well do it right). In the previous use, we sorted just to merge-join, the particular ordering didn't actually matter.
</blockquote></details>

Done. Also renamed `randKeyToReadWithinBounds` to `inRangeKeys`.

PR #3252, opened 2024-01-30
sstable: enforce lower bounds in virtualLastSeekLE()

itsbilal:
In the rare case where the data block iter does not return a key at the upper bound of the virtual sstable, such as due to the obsolete bit, we could go past the lower bound of the iterator in `virtualLastSeekLE`. This change addresses this by enforcing the lower bound in those conditionals in `virtualLast` where we previously weren't enforcing anything.

Fixes #3236.

PR #3254, opened 2024-01-30
sstable: document some invariants in virtualLastSeekLE

itsbilal:
Follow-up from #3252.

### DIFF
--- a/sstable/reader_iter_single_lvl.go
+++ b/sstable/reader_iter_single_lvl.go
@@ -880,24 +880,23 @@ func (i *singleLevelIterator) virtualLast() (*InternalKey, base.LazyValue) {
 		// Trivial case.
 		return i.SeekLT(i.upper, base.SeekLTFlagsNone)
 	}
-	return i.virtualLastSeekLE(i.upper)
+	return i.virtualLastSeekLE()
 }
 
 // virtualLastSeekLE is called by virtualLast to do a SeekLE as part of a
 // virtualLast. Consider generalizing this into a SeekLE() if there are other
-// uses of this method in the future.
-func (i *singleLevelIterator) virtualLastSeekLE(key []byte) (*InternalKey, base.LazyValue) {
+// uses of this method in the future. Does a SeekLE on the upper bound of the
+// file/iterator.
+func (i *singleLevelIterator) virtualLastSeekLE() (*InternalKey, base.LazyValue) {
 	// Callers of SeekLE don't know about virtual sstable bounds, so we may
 	// have to internally restrict the bounds.
 	//
 	// TODO(bananabrick): We can optimize this check away for the level iter
 	// if necessary.
-	if i.cmp(key, i.upper) >= 0 {
-		if !i.endKeyInclusive {
-			panic("unexpected virtualLastSeekLE with exclusive upper bounds")
-		}
-		key = i.upper
+	if !i.endKeyInclusive {
+		panic("unexpected virtualLastSeekLE with exclusive upper bounds")
 	}
+	key := i.upper
 
 	i.exhaustedBounds = 0
 	i.err = nil // clear cached iteration error
@@ -911,6 +910,17 @@ func (i *singleLevelIterator) virtualLastSeekLE(key []byte) (*InternalKey, base.
 	// We can have multiple internal keys with the same user key as the seek
 	// key. In that case, we want the last (greatest) internal key.
 	//
+	// INVARIANT: One of two cases:
+	// A. ikey == nil. There is no data block with index key >= key. So all keys
+	//    in the last data block are < key.
+	// B. ikey.userkey >= key. This data block may have some keys > key.
+	//
+	// Subcases of B:
+	//   B1. ikey.userkey == key. This is when loop iteration happens.
+	//       Since ikey.UserKey >= largest data key in the block, the largest data
+	//       key in this block is <= key.
+	//   B2. ikey.userkey > key. Loop iteration will not happen.
+	//
 	// NB: We can avoid this Next()ing if we just implement a blockIter.SeekLE().
 	// This might be challenging to do correctly, so impose regular operations
 	// for now.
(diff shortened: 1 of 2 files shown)


sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3254)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @itsbilal)

___
*[`sstable/reader_iter_single_lvl.go` line 936 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3254#-NpS19jfAwx2vxqf8Xei:-NpS19jfAwx2vxqf8Xej:b-uscupu) ([raw file](https://github.com/cockroachdb/pebble/blob/b56ae1734732a6cb19e8093582ca0dc4a38d3516/sstable/reader_iter_single_lvl.go#L936)):*
> ```Go
> 		return i.skipBackward()
> 	}
> 	// Case B. It could be B2, so there could be keys in the block > key. But the
> ```

I think there is a tighter version of my original comment. We are here either because it was (a) originally case B2, (b) originally case B1 and we iterated successfully to a block with `ikey.UserKey > key`. So the invariant here is `ikey.UserKey > key` and the block preceding this one has no keys > key.
___
*[`sstable/reader_iter_single_lvl.go` line 946 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3254#-NpS10Nz0fCoZIi3epTi:-NpS10Nz0fCoZIi3epTj:b3dih8h) ([raw file](https://github.com/cockroachdb/pebble/blob/b56ae1734732a6cb19e8093582ca0dc4a38d3516/sstable/reader_iter_single_lvl.go#L946)):*
> ```Go
> 		// Enforce the lower bound here, as we could have gone past it, if the
> 		// block property filter excluded the only relevant block within the
> 		// bounds.
> ```

So if `ikey.UserKey > key` then the only way for `ikey.UserKey < i.lower` to be true is if `key < i.lower`. Am I missing something?
___
*[`sstable/reader_iter_single_lvl.go` line 964 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3254#-NpS3M9U88FQNRMiKcfQ:-NpS3M9U88FQNRMiKcfR:bonfyoq) ([raw file](https://github.com/cockroachdb/pebble/blob/b56ae1734732a6cb19e8093582ca0dc4a38d3516/sstable/reader_iter_single_lvl.go#L964)):*
> ```Go
> 	if ikey != nil {
> 		// Enforce the lower bound here, as we could have gone past it. This happens
> 		// if keys between `i.blockLower` and `key` are obsolete, for instance.
> ```

Thanks, this is clearer to me now.
Perhaps extend the comment a bit. Something like the following (feel free to rephrase). Mentioning RANGEDEL below because if they were obsolete due to a more recent point key, we would have found a userkey higher than the lower bound.

// Even though `i.blockLower` (which is equal to `i.lower`) is <= `key`, all internal keys in the user key interval `[i.blockLower, key]` could be obsolete (due to a RANGEDEL). And `i.data.Prev` will skip all these obsolete keys, and could land on a key below the lower bound.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3254#-:-NpVvr3n9mv-YodGK8Ju:b-814g8k)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 936 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3254#-NpS19jfAwx2vxqf8Xei:-NpVuypBDGx7L8JXX-mq:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b56ae1734732a6cb19e8093582ca0dc4a38d3516/sstable/reader_iter_single_lvl.go#L936)):*

I think there is a tighter version of my original comment. We are here either because it was (a) originally case B2, (b) originally case B1 and we iterated successfully to a block with `ikey.UserKey > key`. So the invariant here is `ikey.UserKey > key` and the block preceding this one has no keys > key.
</blockquote></details>

Done.
___
*[`sstable/reader_iter_single_lvl.go` line 946 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3254#-NpS10Nz0fCoZIi3epTi:-NpVv9aN9HcC8gOr0r4V:bvl2g22) ([raw file](https://github.com/cockroachdb/pebble/blob/b56ae1734732a6cb19e8093582ca0dc4a38d3516/sstable/reader_iter_single_lvl.go#L946)):*

So if `ikey.UserKey > key` then the only way for `ikey.UserKey < i.lower` to be true is if `key < i.lower`. Am I missing something?
</blockquote></details>

I think that's accurate, but that's also the same logic that `SeekLT` applies. We're supposed to check for the lower bound in all cases so I think that's why we need this here.
___
*[`sstable/reader_iter_single_lvl.go` line 964 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3254#-NpS3M9U88FQNRMiKcfQ:-NpVvlZV7yDWhb8j_0VP:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b56ae1734732a6cb19e8093582ca0dc4a38d3516/sstable/reader_iter_single_lvl.go#L964)):*

Thanks, this is clearer to me now.
Perhaps extend the comment a bit. Something like the following (feel free to rephrase). Mentioning RANGEDEL below because if they were obsolete due to a more recent point key, we would have found a userkey higher than the lower bound.

// Even though `i.blockLower` (which is equal to `i.lower`) is <= `key`, all internal keys in the user key interval `[i.blockLower, key]` could be obsolete (due to a RANGEDEL). And `i.data.Prev` will skip all these obsolete keys, and could land on a key below the lower bound.
</blockquote></details>

Done.

sumeerbhola:
(reviewable stopped working for me)

> I think that's accurate, but that's also the same logic that SeekLT applies.

I think that for SeekLT, `key` is allowed to be lower than the lower bound. But here that is not allowed.

We should also consider simplifying the signature of `virtualLastSeekLE` to not accept a parameter since it is always called with `i.upper`. I also don't understand what is going on in the following (it seems unnecessary):
```
func (i *twoLevelIterator) virtualLastSeekLE(key []byte) (*InternalKey, base.LazyValue) {
	// Callers of SeekLE don't know about virtual sstable bounds, so we may
	// have to internally restrict the bounds.
	//
	// TODO(bananabrick): We can optimize this check away for the level iter
	// if necessary.
	if i.cmp(key, i.upper) >= 0 {
		if !i.endKeyInclusive {
			panic("unexpected virtualLastSeekLE with exclusive upper bounds")
		}
		key = i.upper
	}
```

We can always generalize later (if a general `SeekLE` is needed).

itsbilal:
> I think that for SeekLT, key is allowed to be lower than the lower bound. But here that is not allowed.

> We should also consider simplifying the signature of virtualLastSeekLE to not accept a parameter since it is always called with i.upper. I also don't understand what is going on in the following (it seems unnecessary)

Cleaned up both and removed the parameter in `virtualLastSeekLE`. Also removed the unnecessary upper bound check seeing as `key` will now always be the upper bound.

sumeerbhola (approved):
looks great
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3254#-:-Nq3uZcR2Pac8VGa86Hz:bmf8erq)* status: all files reviewed, 4 unresolved discussions (waiting on @itsbilal)

___
*[`-- commits` line 2 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3254#-Nq3tm7kBMxY1Z24Azn_):*
slightly stale commit message

PR #3255, opened 2024-01-31
vfs: add write category during file creation

CheranMahalingam:
This commit modifies the vfs file creation interfaces to tag files with the source of writes supporting fine-grained metrics for disk writes. This change will be followed up with vfs interface changes at the cockroach layer (https://github.com/cockroachdb/cockroach/pull/119024). Note that once this change is merged, the pebble version used in cockroach cannot be bumped up until the vfs interface changes are resolved.

Informs https://github.com/cockroachdb/cockroach/issues/115434

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -403,7 +403,7 @@ func (d *DB) writeCheckpointManifest(
 		}
 		defer src.Close()
 
-		dst, err := fs.Create(destPath)
+		dst, err := fs.Create(destPath, vfs.WriteCategoryUnspecified)
 		if err != nil {
 			return err
 		}
--- a/cleaner_test.go
+++ b/cleaner_test.go
@@ -123,7 +123,7 @@ func TestCleaner(t *testing.T) {
 			if len(td.CmdArgs) != 1 {
 				return "create-bogus-file <db/file>"
 			}
-			dst, err := fs.Create(td.CmdArgs[0].String())
+			dst, err := fs.Create(td.CmdArgs[0].String(), vfs.WriteCategoryUnspecified)
 			require.NoError(t, err)
 			_, err = dst.Write([]byte("bogus data"))
 			require.NoError(t, err)
--- a/cmd/pebble/fsbench.go
+++ b/cmd/pebble/fsbench.go
@@ -113,7 +113,7 @@ type fsBench struct {
 // createFile can be used to create an empty file.
 // Invariant: File shouldn't already exist.
 func createFile(filepath string) vfs.File {
-	fh, err := fsConfig.fs.Create(filepath)
+	fh, err := fsConfig.fs.Create(filepath, vfs.WriteCategoryUnspecified)
 	if err != nil {
 		log.Fatalln(err)
 	}
--- a/commit_test.go
+++ b/commit_test.go
@@ -224,7 +224,7 @@ func TestCommitPipelineWALClose(t *testing.T) {
 	// rotate and close the log.
 
 	mem := vfs.NewMem()
-	f, err := mem.Create("test-wal")
+	f, err := mem.Create("test-wal", vfs.WriteCategoryUnspecified)
 	require.NoError(t, err)
 
 	// syncDelayFile will block on the done channel befor returning from Sync
--- a/compaction.go
+++ b/compaction.go
@@ -2897,9 +2897,25 @@ func (d *DB) runCompaction(
 				ctx = objiotracing.WithReason(ctx, objiotracing.ForCompaction)
 			}
 		}
+		var writeCategory vfs.DiskWriteCategory
+		switch c.kind {
+		case compactionKindFlush:
+			if d.opts.EnableSQLRowSpillMetrics {
+				// In the scenario that the Pebble engine is used for SQL row spills the data written to
+				// the memtable will correspond to spills to disk and should be categorized as such.
+				writeCategory = "sql-row-spill"
+			} else {
+				writeCategory = "pebble-memtable-flush"
+			}
+		case compactionKindIngestedFlushable:
+			writeCategory = "pebble-ingestion"
+		default:
+			writeCategory = "pebble-compaction"
+		}
 		// Prefer shared storage if present.
 		createOpts := objstorage.CreateOptions{
 			PreferSharedStorage: remote.ShouldCreateShared(d.opts.Experimental.CreateOnShared, c.outputLevel.level),
+			WriteCategory:       writeCategory,
 		}
 		diskFileNum := base.PhysicalTableDiskFileNum(fileNum)
 		writable, objMeta, err := d.objProvider.Create(ctx, fileTypeTable, diskFileNum, createOpts)
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -2752,7 +2752,7 @@ func TestCompactionErrorCleanup(t *testing.T) {
 
 	ingest := func(keys ...string) {
 		t.Helper()
-		f, err := mem.Create("ext")
+		f, err := mem.Create("ext", vfs.WriteCategoryUnspecified)
 		require.NoError(t, err)
 
 		w := sstable.NewWriter(objstorageprovider.NewFileWritable(f), sstable.WriterOptions{
@@ -3477,7 +3477,7 @@ func TestCompaction_LogAndApplyFails(t *testing.T) {
 	ingestKeys := func(db *DB) error {
 		// Create an SST for ingestion.
 		const fName = "ext"
-		f, err := db.opts.FS.Create(fName)
+		f, err := db.opts.FS.Create(fName, vfs.WriteCategoryUnspecified)
 		require.NoError(t, err)
 		w := sstable.NewWriter(objstorageprovider.NewFileWritable(f), sstable.WriterOptions{})
 		require.NoError(t, w.Set(key, nil))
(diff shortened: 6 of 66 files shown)


CheranMahalingam (commented):
One concern that I brought up with @aadityasondhi is that there are a significant number of tedious changes that have to be made at the Pebble and Cockroach layer to support the new file creation interfaces. Most of these changes are in test files where an in-memory FS is used. If this impacts readability, an alternative solution we discussed was to add new methods to the FS interface to wrap file creation calls with the write category. Let me know if this solution is preferred.

> Reviewed 2 of 4 files at r1, 58 of 58 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3255#-:-NpVKp3d7j5LgfR2RcMC:bnjr9vl)* status: 58 of 60 files reviewed, all discussions resolved (waiting on @CheranMahalingam)

jbowens (commented):
> One concern that I brought up with [@aadityasondhi](https://github.com/aadityasondhi) is that there are a significant number of tedious changes that have to be made at the Pebble and Cockroach layer to support the new file creation interfaces. Most of these changes are in test files where an in-memory FS is used. If this impacts readability, an alternative solution we discussed was to add new methods to the FS interface to wrap file creation calls with the write category. Let me know if this solution is preferred.

I'd like to avoid adding methods to the FS interface, because we want to make it difficult to accidentally omit the write category in production code. Using the type system is the best way to do it. With separate methods, a caller can easily forget to set the write category, resulting in uncategorized writes. 

 If the churn is too high, there are alternatives with less protection (but more than a separate method).

One is to tag the _FS_ with the write category:
```
type writeBandwidthFS struct {
    vfs.FS
    writeBandwidthCounter *atomic.Uint64
}

// ...

type writeBandwidthFile struct {
    vfs.File
    writeBandwidthCounter *atomic.Uint64
}

func (f *writeBandwidthFile) Write(b []byte) (n int, error) {
    n, err := f.File.Write(b)
    f.writeBandwidthCounter.Add(n)
    return n, err
}
```
Then we'd add a method like `func (d *DB) fs(category string) vfs.FS` and prohibit direct use  of `opts.FS`. The `opts` has already propagated deeply across Pebble, so this is likely still significant churn. It does have the advantage that the knowledge of write categories is isolated and avoids leaking into the vfs.FS interface itself. Happy to chat these options through

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3255#-:-NpVaC-WA9kBoWPPWpzw:bx798i9)* status: 58 of 60 files reviewed, all discussions resolved (waiting on @aadityasondhi and @CheranMahalingam)

CheranMahalingam (commented):
I agree that the additional type safety is beneficial, my worry with changing the interface is also that it forces users of the interface to be familiar with the metrics changes even though most of the time the category should be the default, `unspecified`. This is made worse by not having the type safety of enums for the categories and I am wondering whether something similar to the `ReadCategory` code would help at the Pebble layer (https://github.com/cockroachdb/cockroach/blob/master/pkg/storage/engine.go#L2190). I am confused about the FS tagging strategy, is it possible an FS could create multiple files with different categories and in that case would I have to manage multiple of the `writeBandwidthFS` structs?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3255#-:-NpVlBjS3IR0vI61x25n:b-2ul3am)* status: 58 of 60 files reviewed, all discussions resolved (waiting on @aadityasondhi)

jbowens (commented):
> my worry with changing the interface is also that it forces users of the interface to be familiar with the metrics changes even though most of the time the category should be the default, `unspecified`

This is only true in tests. Production code should almost always be specifying a category.

> This is made worse by not having the type safety of enums for the categories and I am wondering whether something similar to the `ReadCategory` code would help at the Pebble layer ([https://github.com/cockroachdb/cockroach/blob/master/pkg/storage/engine.go#L2190](https://github.com/cockroachdb/cockroach/blob/master/pkg/storage/engine.go#L2190)).

We can and probably should do something similar within pkg/storage for Cockroach callers, but Pebble should not know about CockroachDB concepts and shouldn't try to enumerate all of CockroachDB's uses of of the filesystem.

> I am confused about the FS tagging strategy, is it possible an FS could create multiple files with different categories and in that case would I have to manage multiple of the `writeBandwidthFS` structs?

One `FS` would only create files within one category. This means the leaves of the system need only take a `vfs.FS` since a single subsystem will only use a single category. Higher levels would need to propagate a `func(category string) vfs.FS` to allow the creation of multiple categorized FSes.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3255#-:-NpW8JSa1qYEFxqYvCVe:bb6tgmb)* status: 58 of 60 files reviewed, all discussions resolved (waiting on @aadityasondhi and @CheranMahalingam)

CheranMahalingam (commented):
I like how tagging the FS indirectly helps avoid changing the file creation interfaces. Are there any concerns regarding people trying to wrap this vfs with another vfs since if there are 8 categories each vfs would need to be wrapped (or we need to enforce that the `writeBandwidthFS` is at the highest layer) and the worst case scenario would be a new category being created and the new `writeBandwidthFS` missing the higher layer vfs? An implementation specific concern I have is how do I pass DB around such that they can access the `fs(category string)` method. For instance, how would I pass down the pebble DB in the cockroach layer to access a valid vfs for sst ingestion (https://github.com/cockroachdb/cockroach/blob/master/pkg/kv/kvserver/logstore/sideload_disk.go#L94)?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3255#-:-Np_9zls1OxwbuoSwUSA:b-4z8gvf)* status: 58 of 60 files reviewed, all discussions resolved (waiting on @aadityasondhi)

jbowens (commented):
> An implementation specific concern I have is how do I pass DB around such that they can access the `fs(category string)` method.

A [`storage.Engine`](https://github.com/cockroachdb/cockroach/blob/cc4fdffa8532d16544c48ef036689763f737dc6b/pkg/storage/engine.go#L1069-L1070) is the CockroachDB abstraction over the storage layer, and it currently embeds the `vfs.FS` interface. We could adapt this interface to instead expose a `FS(category string) vfs.FS` method.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3255#-:-Np_sGiu3N02kVXB01Ii:bgx8xth)* status: 58 of 60 files reviewed, all discussions resolved (waiting on @aadityasondhi and @CheranMahalingam)

CheranMahalingam (commented):
Sounds good. In that case is it reasonable to continue with the existing approach or is it better to explore this new idea to maintain the interface?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3255#-:-NpaWi3gBooGL7z3iscL:bx1tg09)* status: 58 of 60 files reviewed, all discussions resolved (waiting on @aadityasondhi)

jbowens (commented):
> Sounds good. In that case is it reasonable to continue with the existing approach or is it better to explore this new idea to maintain the interface?

I think both are reasonable; up to you!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3255#-:-NpuD2RQ0ovQ-3LGMbG9:b-vjiykc)* status: 58 of 60 files reviewed, all discussions resolved (waiting on @aadityasondhi and @CheranMahalingam)

aadityasondhi (commented):
This is looking good! Just one small nit.

> Reviewed 25 of 58 files at r2, 2 of 2 files at r3, 1 of 5 files at r6, 38 of 38 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3255#-:-NstX99bCV_PSP8-_7py:bqubm96)* status: 66 of 68 files reviewed, 1 unresolved discussion (waiting on @CheranMahalingam and @jbowens)

___
*[`vfs/disk_health.go` line 138 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3255#-NstW0slEZCoG9CY7st3:-NstW0slEZCoG9CY7st4:b-uig1v0) ([raw file](https://github.com/cockroachdb/pebble/blob/a20aade3be0b5604948cafe6e5d0fde410c8a3c6/vfs/disk_health.go#L138)):*
> ```Go
> func (d *DiskWriteStatsCollector) CreateStat(category DiskWriteCategory) *atomic.Uint64 {
> 	var bytesWritten *atomic.Uint64
> 	d.mu.Lock()
> ```

Let's use `defer d.mu.Unlock()` here.

CheranMahalingam (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3255#-:-NsxOedqAZIdh6pgFL-H:b-814g8k)* status: 65 of 68 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @jbowens)

___
*[`vfs/disk_health.go` line 138 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3255#-NstW0slEZCoG9CY7st3:-NsxOgPvEuTNAB3pqGO8:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/a20aade3be0b5604948cafe6e5d0fde410c8a3c6/vfs/disk_health.go#L138)):*
<details><summary><i>Previously, aadityasondhi (Aaditya Sondhi) wrote…</i></summary><blockquote>

Let's use `defer d.mu.Unlock()` here.
</blockquote></details>

Done.

aadityasondhi (approved):
:lgtm:

Let's hold off on merging until the related cockroach changes are ready to merge as well.

> Reviewed 2 of 18 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3255#-:-NsyEi2N-aep8VEOY0Oe:b-buwrb0)* status: 67 of 68 files reviewed, 1 unresolved discussion (waiting on @jbowens)

PR #2093, opened 2022-11-04
metamorphic: run subtests in parallel

jbowens:
Adjust the metamorphic tests to run the variant subtests in parallel. This reduces the wall-time of the test somewhere around 20-40% on my laptop, and will improve utilization of host machines.

The benefit is not as pronounced as I'd hoped, because there are outlier subtests with considerably longer execution time than the other subtests (eg, `use_disk=true`).

Informs #2079.

PR #3256, opened 2024-01-31
metamorphic: fix passed test.run argument

RaduBerinde:
#### metamorphic: fix passed test.run argument

In #2093 we accidentally changed the argument to `test.run` to be the
full subtest name (e.g. `-test.run=TestMeta/execution/random-015$`)
instead of just the top-level name. This makes the `$` ineffective and
we end up running both `TestMeta` and `TestMetaTwoInstance`.

#### metamorphic: prevent write-after-close through history

The history object implements a logger object which can be used after
the end of the test if a database isn't closed. This can cause writes
to the history file after close (which can result in subtle issues).

This commit adds a `Close()` method to `history` which causes further
logging calls to be ignored.

### DIFF
--- a/metamorphic/history.go
+++ b/metamorphic/history.go
@@ -12,6 +12,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"unicode"
 
@@ -28,6 +29,10 @@ type history struct {
 	err    atomic.Value
 	failRE *regexp.Regexp
 	log    *log.Logger
+	mu     struct {
+		sync.Mutex
+		closed bool
+	}
 }
 
 func newHistory(failRE *regexp.Regexp, writers ...io.Writer) *history {
@@ -36,8 +41,18 @@ func newHistory(failRE *regexp.Regexp, writers ...io.Writer) *history {
 	return h
 }
 
-// Recordf records the results of a single operation.
+func (h *history) Close() {
+	h.mu.Lock()
+	h.mu.closed = true
+	h.mu.Unlock()
+}
+
 func (h *history) Recordf(op int, format string, args ...interface{}) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	if h.mu.closed {
+		panic("Recordf after close")
+	}
 	if strings.Contains(format, "\n") {
 		// We could remove this restriction but suffixing every line with "#<seq>".
 		panic(fmt.Sprintf("format string must not contain \\n: %q", format))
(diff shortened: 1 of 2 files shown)


RaduBerinde:
I still can't quite explain the failure I observed when running small tests under stress (e.g. `go test -tags 'invariants' -exec 'stress -p 2' -timeout 0 -test.v -run TestMeta$ ./internal/metamorphic --ops 'uniform:10-20'`) .

The history file would have a final line at the end with a lot of `\x00`s which fails parsing in the compare stage. This PR fixes it but I don't understand how it happened because `TestMeta` and `TestMetaTwoInstance` should run serially (and effectively do the same thing when `--run-dir` is passed)..

RaduBerinde:
Ah, I figured it out. We are still logging after we close the file. I added a writer that just panics in that case and got this:

```
        panic: write after close!

        goroutine 89 [running]:
        github.com/cockroachdb/pebble/metamorphic.(*foo).Write(0x140000720a0?, {0x1400003e720?, 0x3
          /Users/radu/go/src/github.com/cockroachdb/pebble/metamorphic/meta.go:672 +0x4c
        io.(*multiWriter).Write(0x1400003e720?, {0x1400003e720, 0x30, 0x60})
          /opt/homebrew/opt/go/libexec/src/io/multi.go:85 +0x74
        log.(*Logger).output(0x140000b5d10, 0x0, 0x1400003b800?, 0x14000219210)
          /opt/homebrew/opt/go/libexec/src/log/log.go:245 +0x3cc
        log.(*Logger).Output(0x140003ea4c0?, 0x10329cca9?, {0x1400003b800?, 0x10329a52c?})
          /opt/homebrew/opt/go/libexec/src/log/log.go:195 +0x4c
        github.com/cockroachdb/pebble/metamorphic.(*history).Infof(0x1400007f360, {0x10329a52c?, 0x
          /Users/radu/go/src/github.com/cockroachdb/pebble/metamorphic/history.go:83 +0x64
        github.com/cockroachdb/pebble.MakeLoggingEventListener.func13({0x140003cc000?})
          /Users/radu/go/src/github.com/cockroachdb/pebble/event.go:669 +0xa4
        github.com/cockroachdb/pebble.(*DB).collectTableStats(0x140003cc000)
          /Users/radu/go/src/github.com/cockroachdb/pebble/table_stats.go:123 +0x300
        created by github.com/cockroachdb/pebble.(*DB).maybeCollectTableStatsLocked in goroutine 50
 ```

I think we need to make `db.Close()` wait for any background workers to exit.

RaduBerinde:
I added a commit that circumvents the write-after-close issue from the test. Also filed https://github.com/cockroachdb/pebble/issues/3257 to fix logging after Close from the db side.

RaduBerinde:
I believe that when we ran `TestMetaTwoInstance` with ops from `TestMeta`, the second db remained open and that's where the messages came from. But I'd like to keep this safeguard for cases where we don't close the db (when using `--try-to-reduce`). I amended the comments.

jbowens (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3256#-:-NpvMu-S8_1qIc1WSZAC:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

PR #3262, opened 2024-02-02
db: use correct bounds for shared ingestion

### LINKED ISSUE #3167: db: incorrect bounds on sstables in TestMetaTwoInstances
PR #3162 adds bounds assertions. The stricter variant (`AssertBounds`) can fail in `TestMetaTwoInstances`. This issue tracks finding the bugs and enabling the stricter assertion.

Given that this only seems to manifest in `TestMetaTwoInstances`, the issue is likely specific to disaggregated storage.

Information on one problem found is here:
https://github.com/cockroachdb/pebble/pull/3162#issuecomment-1858031214

RaduBerinde:
#### db: remove levelIter reliance on SmallestPointKey kind

The level iterator relies on checking if `SmallestPointKey` is of kind
`InternalKeyKindRangeDelete`. But in the case of external files, we
don't know the kind and we use `InternalKeyKindMax`.

We modify the code to always issue a synthetic boundary at the
smallest boundary if we have a rangeDel iterator, and adjust the logic
to potentially issue two synthetic boundaries (one at
`SmallestPointKey` and one at `Smallest`).

We add an external ingestion testcase which fails against the existing
code.

#### db: use correct bounds for shared ingestion

The kinds in the key bounds for shared ingestion can be incorrect
because of sequence number rewriting. We loosen the bounds to use the
minimum and maximum relevant kinds, similar to external file
ingestion.

We also switch to using `AssertBounds` instead of
`AssertUserKeyBounds` which is now possible.

Fixes #3167

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -121,29 +121,25 @@ func ingestSynthesizeShared(
 	// sstable for compaction just yet, as we do not have a clear sense of what
 	// parts of this sstable are referenced by other nodes.
 	meta.FileBacking.Size = sm.Size
-	if sm.LargestRangeKey.Valid() && sm.LargestRangeKey.UserKey != nil {
-		// Initialize meta.{HasRangeKeys,Smallest,Largest}, etc.
+	if sm.LargestPointKey.Valid() && sm.LargestPointKey.UserKey != nil {
+		// Initialize meta.{HasPointKeys,Smallest,Largest}, etc.
 		//
-		// NB: We create new internal keys and pass them into ExternalRangeKeyBounds
+		// NB: We create new internal keys and pass them into ExtendPointKeyBounds
 		// so that we can sub a zero sequence number into the bounds. We can set
 		// the sequence number to anything here; it'll be reset in ingestUpdateSeqNum
 		// anyway. However, we do need to use the same sequence number across all
 		// bound keys at this step so that we end up with bounds that are consistent
 		// across point/range keys.
-		// Note that the kind of the smallest key might change because of the seqnum
-		// rewriting. For example, the sstable could start with a.SET.2 and
-		// a.RANGEDEL.1 (with smallest key being a.SET.2) but after rewriting the seqnum we have `a.RANGEDEL.1`a.SET.100
-		smallestRangeKey := base.MakeInternalKey(sm.SmallestRangeKey.UserKey, 0, sm.SmallestRangeKey.Kind())
-		largestRangeKey := base.MakeExclusiveSentinelKey(sm.LargestRangeKey.Kind(), sm.LargestRangeKey.UserKey)
-		meta.ExtendRangeKeyBounds(opts.Comparer.Compare, smallestRangeKey, largestRangeKey)
-	}
-	if sm.LargestPointKey.Valid() && sm.LargestPointKey.UserKey != nil {
-		// Initialize meta.{HasPointKeys,Smallest,Largest}, etc.
 		//
-		// See point above in the ExtendRangeKeyBounds call on why we use a zero
-		// sequence number here.
-		smallestPointKey := base.MakeInternalKey(sm.SmallestPointKey.UserKey, 0, sm.SmallestPointKey.Kind())
-		largestPointKey := base.MakeInternalKey(sm.LargestPointKey.UserKey, 0, sm.LargestPointKey.Kind())
+		// Because of the sequence number rewriting, we cannot use the Kind of
+		// sm.SmallestPointKey. For example, the original SST might start with
+		// a.SET.2 and a.RANGEDEL.1 (with a.SET.2 being the smallest key); after
+		// rewriting the sequence numbers, these keys become a.SET.100 and
+		// a.RANGEDEL.100, with a.RANGEDEL.100 being the smallest key. To create a
+		// correct bound, we just use the maximum key kind (which sorts first).
+		// Similarly, we use the smallest key kind for the largest key.
+		smallestPointKey := base.MakeInternalKey(sm.SmallestPointKey.UserKey, 0, base.InternalKeyKindMax)
+		largestPointKey := base.MakeInternalKey(sm.LargestPointKey.UserKey, 0, 0)
 		if sm.LargestPointKey.IsExclusiveSentinel() {
 			largestPointKey = base.MakeRangeDeleteSentinelKey(sm.LargestPointKey.UserKey)
 		}
@@ -156,6 +152,15 @@ func ingestSynthesizeShared(
 			smallestPointKey, largestPointKey = largestPointKey, smallestPointKey
 		}
 		meta.ExtendPointKeyBounds(opts.Comparer.Compare, smallestPointKey, largestPointKey)
+	}
+	if sm.LargestRangeKey.Valid() && sm.LargestRangeKey.UserKey != nil {
+		// Initialize meta.{HasRangeKeys,Smallest,Largest}, etc.
+		//
+		// See comment above on why we use a zero sequence number and these key
+		// kinds here.
+		smallestRangeKey := base.MakeInternalKey(sm.SmallestRangeKey.UserKey, 0, base.InternalKeyKindRangeKeyMax)
+		largestRangeKey := base.MakeExclusiveSentinelKey(base.InternalKeyKindRangeKeyMin, sm.LargestRangeKey.UserKey)
+		meta.ExtendRangeKeyBounds(opts.Comparer.Compare, smallestRangeKey, largestRangeKey)
 	}
 	if err := meta.Validate(opts.Comparer.Compare, opts.Comparer.FormatKey); err != nil {
 		return nil, err
--- a/internal/keyspan/truncate.go
+++ b/internal/keyspan/truncate.go
@@ -4,7 +4,11 @@
 
 package keyspan
 
-import "github.com/cockroachdb/pebble/internal/base"
+import (
+	"fmt"
+
+	"github.com/cockroachdb/pebble/internal/base"
+)
 
 // Truncate creates a new iterator where every span in the supplied iterator is
 // truncated to be contained within the range [lower, upper). If start and end
@@ -54,18 +58,16 @@ func Truncate(
 			}
 		}
 
-		var truncated bool
 		// Truncate the bounds to lower and upper.
 		if cmp(in.Start, lower) < 0 {
 			out.Start = lower
 		}
 		if cmp(in.End, upper) > 0 {
-			truncated = true
-			out.End = upper
-		}
+			if panicOnUpperTruncate {
+				panic(fmt.Sprintf("pebble: upper bound should not be truncated. span end: %q  bound %q", in.End, upper))
+			}
 
-		if panicOnUpperTruncate && truncated {
-			panic("pebble: upper bound should not be truncated")
+			out.End = upper
 		}
 
 		return !out.Empty() && cmp(out.Start, out.End) < 0
--- a/level_iter.go
+++ b/level_iter.go
@@ -1121,36 +1121,17 @@ func (l *levelIter) skipEmptyFileBackward() (*InternalKey, base.LazyValue) {
 				// bounds.
 				return nil, base.LazyValue{}
 			}
-			// If the boundary is a range deletion tombstone, return that key.
-			if l.iterFile.SmallestPointKey.Kind() == InternalKeyKindRangeDelete {
+			// If the boundary could be a range deletion tombstone, return the
+			// smallest point key as a special ignorable key to avoid advancing to the
+			// next file.
+			//
+			// It's possible the SmallestPointKey was already returned. Returning it
+			// again is a violation of the strict monotonicity normally provided. The
+			// mergingIter's heap can tolerate this repeat key and in this case will
+			// keep the level at the top of the heap and immediately skip the entry,
+			// advancing to the next file.
+			if *l.rangeDelIterPtr != nil {
 				l.smallestBoundary = &l.iterFile.SmallestPointKey
-				if l.boundaryContext != nil {
-					l.boundaryContext.isIgnorableBoundaryKey = true
-				}
-				return l.smallestBoundary, base.LazyValue{}
-			}
-			// If the last point iterator positioning op skipped keys, it's
-			// possible the file's range deletions are still relevant to other
-			// levels. Return the smallest boundary as a special ignorable key
-			// to avoid advancing to the next file.
-			//
-			// The sstable iterator cannot guarantee that keys were skipped.  A
-			// SeekGE that lands on a index separator k only knows that the
-			// block at the index entry contains keys ≤ k. We can't know whether
-			// there were actually keys between the seek key and the index
-			// separator key. If the block is then excluded due to block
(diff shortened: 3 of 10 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3262)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`level_iter.go` line 1151 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3262#-Npb_6vX-dru6Z2Z0Wdz:-Npb_6vX-dru6Z2Z0We-:bu28z7j) ([raw file](https://github.com/cockroachdb/pebble/blob/24844b4b1c279194d85a4886e888a9ff53657176/level_iter.go#L1151)):*
> ```Go
> 			// the next file.
> 			if *l.rangeDelIterPtr != nil && l.filteredIter != nil && l.filteredIter.MaybeFilteredKeys() {
> 				// Only emit this boundary if we haven't emitted it already (or
> ```

I tried a ton of variations before I had this idea of emitting both boundaries. What I don't understand is how we got away with only emitting at `SmallestPointKey` before, since this boundary also seems important. But I don't understand exactly how the mergingIter code that relies on this works.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3262)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`level_iter.go` line 967 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3262#-NpvV1nz1-_7ODRn8xKB:-NpvV1nz1-_7ODRn8xKC:b-5bwan4) ([raw file](https://github.com/cockroachdb/pebble/blob/e53e09a9edbe89c75cc9599a282f843851eb3cd4/level_iter.go#L967)):*
> ```Go
> 			return l.verify(l.skipEmptyFileBackward())
> 		}
> 		return nil, base.LazyValue{}
> ```

If we just always emit the smallest boundary (rather than both), I think we can keep this structure. In any case, I think it'd be nice to keep Prev and Next reciprocal in code structure.
___
*[`level_iter.go` line 1153 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3262#-NpvRZOj9Fs18NygBSLC:-NpvRZOj9Fs18NygBSLD:bnzzk38) ([raw file](https://github.com/cockroachdb/pebble/blob/4a51fb82f2b99a07c7c043c7896413ec198721cf/level_iter.go#L1153)):*
> ```Go
> 				// Only emit this boundary if we haven't emitted it already (or
> 				// emitted one for the same key in the case above).
> 				if l.smallestBoundary == nil && l.cmp(l.smallestBoundary.UserKey, l.iterFile.Smallest.UserKey) > 0 {
> ```

Won't this nil pointer when we access `l.smallestBoundary.UserKey` because `l.smallestBoundary == nil`? Am I parsing this correctly? If I am, it's a bit troublesome that no tests triggered it. This is a very difficult case to exercise, but I thought we had a unit test in testdata/iter_histories. Maybe we only cover the largest boundary.
___
*[`level_iter.go` line 1151 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3262#-Npb_6vX-dru6Z2Z0Wdz:-NpvRzpIDoQigJgPr01M:b-8jthq9) ([raw file](https://github.com/cockroachdb/pebble/blob/24844b4b1c279194d85a4886e888a9ff53657176/level_iter.go#L1151)):*
I _think_ with the change you made above we no longer need this second boundary condition. If we give the same treatment to the largest boundary, I don't see why we can't remove the `MaybeFilteredKeys` func and related interfaces altogether.

> What I don't understand is how we got away with only emitting at `SmallestPointKey` before, since this boundary also seems important.

It took me a while staring at it to realize this, but `SmallestPointKey` is the smallest range deletion and/or point key, so an iterator that's positioned at `SmallestPointKey` must not have any range deletions that delete smaller keys.

Only range keys can cause `Smallest` to be less than `SmallestPointKey`, and the merging iterator doesn't care about range keys at all.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3262)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`level_iter.go` line 1153 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3262#-NpvRZOj9Fs18NygBSLC:-NpvvFlz6alESGnItON2:b-xnmp32) ([raw file](https://github.com/cockroachdb/pebble/blob/4a51fb82f2b99a07c7c043c7896413ec198721cf/level_iter.go#L1153)):*

Won't this nil pointer when we access `l.smallestBoundary.UserKey` because `l.smallestBoundary == nil`? Am I parsing this correctly? If I am, it's a bit troublesome that no tests triggered it. This is a very difficult case to exercise, but I thought we had a unit test in testdata/iter_histories. Maybe we only cover the largest boundary.
</blockquote></details>

Oh, wow, this was supposed to be an `||`. So then we must never be exiting here. But I did try just with the previous boundary and got meta test failures. Maybe I didn't stress it enough.

RaduBerinde:
*[`level_iter.go` line 1151 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3262#-Npb_6vX-dru6Z2Z0Wdz:-NpvxaKf57zdoTqGF9N5:b-xo0eqr) ([raw file](https://github.com/cockroachdb/pebble/blob/24844b4b1c279194d85a4886e888a9ff53657176/level_iter.go#L1151)):*

I _think_ with the change you made above we no longer need this second boundary condition. If we give the same treatment to the largest boundary, I don't see why we can't remove the `MaybeFilteredKeys` func and related interfaces altogether.

> What I don't understand is how we got away with only emitting at `SmallestPointKey` before, since this boundary also seems important.

It took me a while staring at it to realize this, but `SmallestPointKey` is the smallest range deletion and/or point key, so an iterator that's positioned at `SmallestPointKey` must not have any range deletions that delete smaller keys.

Only range keys can cause `Smallest` to be less than `SmallestPointKey`, and the merging iterator doesn't care about range keys at all.
</blockquote></details>

Then why did we need this boundary before?

RaduBerinde:
*[`level_iter.go` line 1153 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3262#-NpvRZOj9Fs18NygBSLC:-NpvynOZ6mRzfMo7oeeT:bmx3lny) ([raw file](https://github.com/cockroachdb/pebble/blob/4a51fb82f2b99a07c7c043c7896413ec198721cf/level_iter.go#L1153)):*

Oh, wow, this was supposed to be an `||`. So then we must never be exiting here. But I did try just with the previous boundary and got meta test failures. Maybe I didn't stress it enough.
</blockquote></details>

I don't think it's possible to get into this path without having emitted the previous boundary (IOW `smallestBoundary` is never nil here), so that explains why we didn't hit a `nil` pointer issue. But it also means we don't actually need this. I'll keep playing with it and see what more I can find.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3262)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`level_iter.go` line 967 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3262#-NpvV1nz1-_7ODRn8xKB:-NpwFcDG2dVNtvUGu8v2:b-el1hbk) ([raw file](https://github.com/cockroachdb/pebble/blob/e53e09a9edbe89c75cc9599a282f843851eb3cd4/level_iter.go#L967)):*

If we just always emit the smallest boundary (rather than both), I think we can keep this structure. In any case, I think it'd be nice to keep Prev and Next reciprocal in code structure.
</blockquote></details>

Done.mak
___
*[`level_iter.go` line 1153 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3262#-NpvRZOj9Fs18NygBSLC:-NpwFFqJ3Pt99bKQ5-9I:b9813eo) ([raw file](https://github.com/cockroachdb/pebble/blob/4a51fb82f2b99a07c7c043c7896413ec198721cf/level_iter.go#L1153)):*

I don't think it's possible to get into this path without having emitted the previous boundary (IOW `smallestBoundary` is never nil here), so that explains why we didn't hit a `nil` pointer issue. But it also means we don't actually need this. I'll keep playing with it and see what more I can find.
</blockquote></details>

I simplified as suggested and it does seem to work. I'm not sure why I was seeing a failure before.

itsbilal (approved):
:lgtm: , thanks for finally resolving this! The simplification in level_iter is also a great side-effect.

> Reviewed 2 of 3 files at r1, 9 of 9 files at r3, 8 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3262#-:-NpyyamZB_1w0I4Oc476:b-p0l0nc)* status: all files reviewed, 4 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`ingest.go` line 127 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3262#-Npyy5Jv8tA5_isenfD2:-Npyy5JwAMYS-KgqZJ4V:b-nzy94z) ([raw file](https://github.com/cockroachdb/pebble/blob/ce80183c1b8af2db8a5571874ca52422995213b7/ingest.go#L127)):*
> ```Go
> 		// Initialize meta.{HasPointKeys,Smallest,Largest}, etc.
> 		//
> 		// NB: We create new internal keys and pass them into ExternalRangeKeyBounds
> ```

I know this isn't all your change but s/External/Extend/, and also s/Range/Point/ 
___
*[`table_cache.go` line 673 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3262#-NpyyTJBEJluI7OdcZJe:-NpyyTJBEJluI7OdcZJf:bb7ddct) ([raw file](https://github.com/cockroachdb/pebble/blob/e49380ba5068fd10cbfa9d6887f4540d9bf84865/table_cache.go#L673)):*
> ```Go
> 		// TODO(radu): we should be using AssertBounds, but it currently fails in
> 		// some cases (#3167).
> 		rangeDelIter = keyspan.AssertUserKeyBounds(
> ```

ayyy

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3262#-:-Npz1M6q1t1-tccdKm4A:b-814g8k)* status: all files reviewed, 3 unresolved discussions (waiting on @jbowens)

jbowens (approved):
:lgtm:

> Reviewed 2 of 3 files at r1, 1 of 9 files at r3, 7 of 7 files at r5, 7 of 7 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3262#-:-Npz5mtA-u-cAQ2xxjLQ:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`level_iter.go` line 1151 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3262#-Npb_6vX-dru6Z2Z0Wdz:-Npz4okV2Iw1X068kyCs:bb742ek) ([raw file](https://github.com/cockroachdb/pebble/blob/24844b4b1c279194d85a4886e888a9ff53657176/level_iter.go#L1151)):*

Then why did we need this boundary before?
</blockquote></details>

We needed it before because a range deletion being the smallest key in the table wasn't the only condition with which we might exhaust the point iterator before the range deletion iterator. If a block property filter excludes blocks containing keys smaller than the smallest range deletion, the point iterator could be exhausted while a range deletion in the file is still relevant to another level. Your change here to unconditionally interleave the boundary when there are any range deletions side steps the problem.

I think we should make the reciprocal change to forward iteration / the largest boundary, and then I think we'll be able to remove the entire concept of a `filteredIter`.

RaduBerinde:
Makes sense, thanks! I will investigate the upper boundary separately, I hit meta test failures last time I tried it. TFTRs!

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3262)* status: all files reviewed, 3 unresolved discussions

___
*[`level_iter.go` line 1151 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3262#-Npb_6vX-dru6Z2Z0Wdz:-Nq-pEHNAR1scH8l6LM_:b8smtut) ([raw file](https://github.com/cockroachdb/pebble/blob/24844b4b1c279194d85a4886e888a9ff53657176/level_iter.go#L1151)):*

We needed it before because a range deletion being the smallest key in the table wasn't the only condition with which we might exhaust the point iterator before the range deletion iterator. If a block property filter excludes blocks containing keys smaller than the smallest range deletion, the point iterator could be exhausted while a range deletion in the file is still relevant to another level. Your change here to unconditionally interleave the boundary when there are any range deletions side steps the problem.

I think we should make the reciprocal change to forward iteration / the largest boundary, and then I think we'll be able to remove the entire concept of a `filteredIter`.
</blockquote></details>

The [corresponding change](https://github.com/cockroachdb/pebble/compare/master...RaduBerinde:pebble:level-iter-boundary-upper?expand=1) for the upper bound fails various tests with:

```
checker failed with error: out of order keys wjhyruuycd@4#66,SET >= wjhyruuycd@4#66,SET in L6: fileNum=000005
```

I tried using just the UserKey and setting the trailer to `InternalKeyRangeDeleteSentinel` but I get a similar error:
```
checker failed with error: out of order keys h#3,SET >= h#inf,RANGEDEL in L2: fileNum=000010
```
Not particularly surprising - if the `LargestPointKey` is not an exclusive sentinel, it's incorrect to emit it. If we only emit this boundary when the largest point key is an exclusive sentinel, this amounts to the current code (the only possible key type is RangeDel) and we can't really remove the other boundary case..

So I don't see how we can apply the same simplification.

RaduBerinde:
*[`level_iter.go` line 1151 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3262#-Npb_6vX-dru6Z2Z0Wdz:-Nq3MQFaD_FkIjVpmswm:b-cxayi9) ([raw file](https://github.com/cockroachdb/pebble/blob/24844b4b1c279194d85a4886e888a9ff53657176/level_iter.go#L1151)):*

I think we might need to just teach the level checker (level_checker.go) about the 'ignorable boundary keys'. I don't think it knows about them currently, and a key that's exactly equal in `Trailer` is unexpected.
</blockquote></details>

OOh interesting, I'll try that.

RaduBerinde:
*[`level_iter.go` line 1151 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3262#-Npb_6vX-dru6Z2Z0Wdz:-NrF_NLT5lDApFCpNsMm:bk05uy) ([raw file](https://github.com/cockroachdb/pebble/blob/24844b4b1c279194d85a4886e888a9ff53657176/level_iter.go#L1151)):*

OOh interesting, I'll try that.
</blockquote></details>

I haven't been able to get back to this yet, I filed an issue so I don't lose track of it (https://github.com/cockroachdb/pebble/issues/3334).

PR #3264, opened 2024-02-02
internal/rangekey: remove coalesce's error return value

jbowens:
The coalesce function is infallible and never returns a non-nil error return value. Remove the error return value altogether.


jbowens:
TFTR!

PR #3270, opened 2024-02-05
metamorphic: add helper for random int with exponential distribution

RaduBerinde:
This change adds a function that generates an integer with the
exponential distribution, where larger values are exponentially less
probable. This is useful to provide some coverage for larger values
while keeping the value small in most cases.

### DIFF
--- a/metamorphic/generator.go
+++ b/metamorphic/generator.go
@@ -7,6 +7,7 @@ package metamorphic
 import (
 	"bytes"
 	"fmt"
+	"math"
 	"os"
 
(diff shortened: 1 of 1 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3270#-:-Npz4UCm7bwG6-GZrDg_:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

PR #3271, opened 2024-02-05
table cache: improve comments, simplify loadInfo

RaduBerinde:
Minor cleanup in the table cache reader to make it less likely that we
will plumb virtual table properties into the physical reader.


itsbilal (approved):
LGTM

I like this as it's strictly better than what we had before. But it'd be good to just move the seqnum out of `loadInfo` and pass it directly to iterators, instead of relying on setting it in the properties block. But that can be a separate follow-up change.

RaduBerinde:
TFTR!  Agreed, I filed #3278 so I don't forget.

RaduBerinde:
It looks like we definitely need to move the seqnum out if we are to reuse external object backings (the same backing can be ingested at different times).

PR #3274, opened 2024-02-06
db: add TestCommitPipelineLogDataSeqNum

jbowens:
Add a unit test covering the sequence number publishing of a batch containing only LogData. A batch containing only a LogData may adopt the sequence number of the most recently committed KV, but it must not publish the sequence number before the preceding KV has been applied to the memtable. This unit test helped convince me that this is the case.

### DIFF
--- a/commit_test.go
+++ b/commit_test.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/cockroachdb/pebble/internal/arenaskl"
+	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/record"
 	"github.com/cockroachdb/pebble/vfs"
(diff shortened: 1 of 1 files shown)


sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3274)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`commit_test.go` line 317 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3274#-NpzbH0j3VxshmkSQTuP:-NpzbH0j3VxshmkSQTuQ:bxlf8r5) ([raw file](https://github.com/cockroachdb/pebble/blob/adda085a889066f36c128b6bcf786436350312ae/commit_test.go#L317)):*
> ```Go
> 			// but it should not have published its sequence number since b was
> 			// enqueued to the commit queue before it.
> 			require.LessOrEqual(t, testEnv.visibleSeqNum.Load(), b.SeqNum())
> ```

can this be strengthened to
```
if b.Count() > 0 {
  require.Less(t, testEnv.visibleSeqNum.Load(), b.SeqNum())
}  else {
  require.LessOrEqual(t, testEnv.visibleSeqNum.Load(), b.SeqNum())
}
```

Or does `visibleSeqNum` mean that everything < `visibleSeqNum` is published, since the CAS is using `newSeqNum := t.SeqNum() + uint64(t.Count())`?

This `LessOrEqual` will also execute for the `LogData` batch -- won't it fail for that batch since the `visibleSeqNum` may already be past that batch?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3274)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @sumeerbhola)

___
*[`commit_test.go` line 317 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3274#-NpzbH0j3VxshmkSQTuP:-Nq-Tvyy0QpmNnJ7C6mq:bihkh22) ([raw file](https://github.com/cockroachdb/pebble/blob/adda085a889066f36c128b6bcf786436350312ae/commit_test.go#L317)):*
> Or does `visibleSeqNum` mean that everything < `visibleSeqNum` is published,

Yeah, this was right. I always forget this. I updated the test to use `base.Visible` which does the right thing and makes it more obviously correct within the test.

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3274#-:-Nq41gh9-jrgQ1m67X38:bg7igkv)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`commit_test.go` line 297 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3274#-Nq40JkiEcBAmgB8nFzC:-Nq40JkjBqyTwCcir-1V:b-sroy3u) ([raw file](https://github.com/cockroachdb/pebble/blob/8f11f006745a83b3ff94b561a2c1c5147bb7afcd/commit_test.go#L297)):*
> ```Go
> // finish applying to the memtable before the KV that shares its sequence
> // number. However, sequence number publishing ratchets the visible sequence
> // number to the batch's seqnum + 1 (indicating that all seqnums strictly less
> ```

the phrasing is slightly confusing.

How about including an example and saying "first seqnum. Something like:

... ratchets the visible sequence number to the batch's first seqnum + number of batch entries ..., so for e.g. with first seqnum = 5 and number of entries = 3, it will ratchet to 8. This means all seqnums strictly less than 8 are visible. So a LogData batch which also grabbed the first seqnum = 5 before this batch, will ratchet to 5 + 0, which is a noop.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3274#-:-Nq4j0zA9zk8rsmXmaBQ:b-814g8k)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`commit_test.go` line 297 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3274#-Nq40JkiEcBAmgB8nFzC:-Nq4j07E445VzXVxmRcB:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8f11f006745a83b3ff94b561a2c1c5147bb7afcd/commit_test.go#L297)):*

the phrasing is slightly confusing.

How about including an example and saying "first seqnum. Something like:

... ratchets the visible sequence number to the batch's first seqnum + number of batch entries ..., so for e.g. with first seqnum = 5 and number of entries = 3, it will ratchet to 8. This means all seqnums strictly less than 8 are visible. So a LogData batch which also grabbed the first seqnum = 5 before this batch, will ratchet to 5 + 0, which is a noop.
</blockquote></details>

Done.

PR #3276, opened 2024-02-06
wal: implement virtual WAL reader

jbowens:
Add a new virtualWALReader type that handles merging multiple physical log
files that together form a single virtual WAL file. This new reader
implementation hides the complexity of reading back a WAL that's undergone
failover between devices.

Informs https://github.com/cockroachdb/pebble/issues/3230.

### DIFF
--- a/internal/datadrivenutil/datadrivenutil.go
+++ b/internal/datadrivenutil/datadrivenutil.go
@@ -1,0 +1,113 @@
+// Package datadrivenutil defines facilities to improve ergonomics around
+// parsing datadriven test input.
+package datadrivenutil
+
+import (
+	"encoding/hex"
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// Lines wraps a string, providing facilities for parsing individual lines.
+type Lines string
+
+// Next returns the string up to the next newline. The receiver is modified to
+// point to the string's contents immediately after the newline.
+func (l *Lines) Next() (line Line) {
+	i := strings.IndexByte(string(*l), '\n')
+	if i == -1 {
+		line = Line(*l)
+		*l = ""
+	} else {
+		line = Line((*l)[:i])
+		*l = (*l)[i+1:]
+	}
+	return line
+}
+
+// A Line is a string with no newlines.
+type Line string
+
+// Fields breaks the line into fields delimited by whitespace.
+func (l Line) Fields() Fields { return Fields(strings.Fields(string(l))) }
+
+// Fields wraps a []string with facilities for parsing out values.
+type Fields []string
+
+// HasValue searches for a field that is exactly the provided string.
+func (fs Fields) HasValue(value string) bool {
+	for i := range fs {
+		if fs[i] == value {
+			return true
+		}
+	}
+	return false
+}
+
+// KeyValue looks for a field containing a key=value pair with the provided key.
+// If not found, KeyValue returns false for the second return value.
+func (fs Fields) KeyValue(key string) (Value, bool) {
+	for i := range fs {
+		if len(fs[i]) >= len(key) && strings.HasPrefix(fs[i], key) && fs[i][len(key)] == '=' {
+			return Value(fs[i][len(key)+1:]), true
+		}
+	}
+	return "", false
+}
+
+// MustKeyValue is like KeyValue but panics if the field is not found.
+func (fs Fields) MustKeyValue(key string) Value {
+	f, ok := fs.KeyValue(key)
+	if !ok {
+		panic(fmt.Sprintf("unable to find required key-value pair %q", key))
+	}
+	return f
+}
+
+// HexBytes parses all the fields as hex-encoded bytes, returning the
+// concatenated decoded bytes. It panics if any of the fields fail to parse as
+// hex-encoded bytes.
+func (fs Fields) HexBytes() []byte {
+	var b []byte
+	for _, f := range fs {
+		b = append(b, Value(f).HexBytes()...)
+	}
+	return b
+}
+
+// A Value represents a single string of unknown structure. A Value is sometimes
+// used to represent an entire element of a Fields and other times a substring
+// of an individual field. This blurring of semantics is convenient for
+// parsing.
+type Value string
+
+// Int parses the value as an int. It panics if the value fails to decode as an
+// integer.
+func (v Value) Int() int {
+	vi, err := strconv.Atoi(string(v))
+	if err != nil {
+		panic(err)
+	}
+	return vi
+}
+
+// Uint64 parses the value as an uint64. It panics if the value fails to decode
+// as an uint64.
+func (v Value) Uint64() uint64 {
+	vi, err := strconv.ParseUint(string(v), 10, 64)
+	if err != nil {
+		panic(err)
+	}
+	return vi
+}
+
+// HexBytes decodes the value as hex-encoded bytes. It panics if the value fails
+// to decode as hex-encoded.
+func (v Value) HexBytes() []byte {
+	b, err := hex.DecodeString(string(v))
+	if err != nil {
+		panic(err)
+	}
+	return b
+}
--- a/wal/reader.go
+++ b/wal/reader.go
@@ -5,13 +5,18 @@
 package wal
 
 import (
+	"bytes"
 	"cmp"
 	"fmt"
+	"io"
 	"slices"
 	"strings"
 
 	"github.com/cockroachdb/errors"
+	"github.com/cockroachdb/pebble/batchrepr"
 	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/record"
+	"github.com/cockroachdb/pebble/vfs"
 )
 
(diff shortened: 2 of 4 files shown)


sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 1 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3276#-:-Nq5-eT117XrUInI6LIF:bg7igkv)* status: all files reviewed, 6 unresolved discussions (waiting on @jbowens)

___
*[`internal/datadrivenutil/datadrivenutil.go` line 50 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3276#-Nq4pClK1FQ2LdBPaXps:-Nq4pClK1FQ2LdBPaXpt:b-av4hh8) ([raw file](https://github.com/cockroachdb/pebble/blob/89d44ef121f571f4b2d67df431211036332dcff6/internal/datadrivenutil/datadrivenutil.go#L50)):*
> ```Go
> // KeyValue looks for a field containing a key=value pair with the provided key.
> // If not found, KeyValue returns false for the second return value.
> func (fs Fields) KeyValue(key string) (Field, bool) {
> ```

This was a bit confusing. I initially assumed `Fields` is a set of `Field` structs. But `Field` is actually the value for a field that is a key-value pair.
But then in `HexBytes` we blur the distinction between a field that is simply a value and a field that is a key-value pair. I think that blurring is fine, but then the return value from this `KeyValue` method should not be `Field`. Perhaps call it `Value`? Having two types makes the parsing code awkward, and this is just a test helper. Code comments explaining these two different uses of `Field` are fine too.
___
*[`wal/reader.go` line 60 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3276#-Nq4tITEB_ORgKrRkTGj:-Nq4tITEB_ORgKrRkTGk:b-bqyz93) ([raw file](https://github.com/cockroachdb/pebble/blob/901a75db0eb59b7da531a6167d29e0f25605ae61/wal/reader.go#L60)):*
> ```Go
> 
> // Get retrieves the WAL with the given number if present. The second return
> // value indicates whether or not the WAL was found.
> ```

nit: does this need to be public?
___
*[`wal/reader.go` line 180 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3276#-Nq4vbzqDnrkIe4VWKyU:-Nq4vbzqDnrkIe4VWKyV:b-shfl23) ([raw file](https://github.com/cockroachdb/pebble/blob/901a75db0eb59b7da531a6167d29e0f25605ae61/wal/reader.go#L180)):*
> ```Go
> 
> 		// Copy the record into a buffer. This ensures we read its entirety so
> 		// that NextRecord returns the next record, even if the caller never
> ```

I didn't understand the "This ensures ..." sentence. Why does it matter whether the caller exhausts the `io.Reader` implemented by `singleReader`? I think `singleReader.Read` only advances `r.begin` and `Reader.Next` doesn't care about the value of `r.begin` and simply overwrites it `r.begin = r.end`.
___
*[`wal/testdata/reader` line 3 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3276#-Nq5-AubDgLm5RtGaAMf:-Nq5-AubDgLm5RtGaAMg:bvt66et) ([raw file](https://github.com/cockroachdb/pebble/blob/901a75db0eb59b7da531a6167d29e0f25605ae61/wal/testdata/reader#L3)):*
> ```
> # Test a simple WAL file that consists of a single segment with a clean WAL
> # tail. This is what we'd expect from a graceful process exit in a non-failover
> # scenario.
> ```

nice test!

can you add a case where the first log has batch0, batch1, batch2; second log has batch1; third log has batch1, batch2, batch3.
___
*[`wal/testdata/reader` line 29 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3276#-Nq4zJKZEY166K33prso:-Nq4zJKZEY166K33prsp:bwr04e1) ([raw file](https://github.com/cockroachdb/pebble/blob/901a75db0eb59b7da531a6167d29e0f25605ae61/wal/testdata/reader#L29)):*
> ```
> 
> # Add a new physical file for the same logical log, this one with a batch that
> # only contians a LogData. This exercises a subtlety in which a sequence number
> ```

nit: contains
___
*[`wal/testdata/reader` line 257 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3276#-Nq5-3hPATM2bAq-Cwjf:-Nq5-3hPATM2bAq-Cwjg:b-yql9uv) ([raw file](https://github.com/cockroachdb/pebble/blob/901a75db0eb59b7da531a6167d29e0f25605ae61/wal/testdata/reader#L257)):*
> ```
> r.NextRecord() = (rr, (000005-001.log: 692), pebble/record: invalid chunk)
> 
> # Read again, this time pretending we found a third segemnt with the
> ```

nit: segment

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3276#-:-Nq96dvZ06JBKGK1Sn-s:b-814g8k)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @sumeerbhola)

___
*[`internal/datadrivenutil/datadrivenutil.go` line 50 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3276#-Nq4pClK1FQ2LdBPaXps:-Nq96Vaj5XBSVXsOV4jq:bfg47g4) ([raw file](https://github.com/cockroachdb/pebble/blob/89d44ef121f571f4b2d67df431211036332dcff6/internal/datadrivenutil/datadrivenutil.go#L50)):*

This was a bit confusing. I initially assumed `Fields` is a set of `Field` structs. But `Field` is actually the value for a field that is a key-value pair.
But then in `HexBytes` we blur the distinction between a field that is simply a value and a field that is a key-value pair. I think that blurring is fine, but then the return value from this `KeyValue` method should not be `Field`. Perhaps call it `Value`? Having two types makes the parsing code awkward, and this is just a test helper. Code comments explaining these two different uses of `Field` are fine too.
</blockquote></details>

Yeah, agreed it's confusing. I renamed `Field` to `Value` and added a code comment.
___
*[`wal/reader.go` line 60 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3276#-Nq4tITEB_ORgKrRkTGj:-Nq8lnwX6rBwrfW3hkKR:b-p6v4vd) ([raw file](https://github.com/cockroachdb/pebble/blob/901a75db0eb59b7da531a6167d29e0f25605ae61/wal/reader.go#L60)):*

nit: does this need to be public?
</blockquote></details>

naw, it doesn't need to be. It's not actually accessible outside the package regardless since the `logicalWALs` type is unexported too. I lowercased it.
___
*[`wal/reader.go` line 180 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3276#-Nq4vbzqDnrkIe4VWKyU:-Nq8x2Xv4APvrRuuKubu:b-qwv45a) ([raw file](https://github.com/cockroachdb/pebble/blob/901a75db0eb59b7da531a6167d29e0f25605ae61/wal/reader.go#L180)):*

I didn't understand the "This ensures ..." sentence. Why does it matter whether the caller exhausts the `io.Reader` implemented by `singleReader`? I think `singleReader.Read` only advances `r.begin` and `Reader.Next` doesn't care about the value of `r.begin` and simply overwrites it `r.begin = r.end`.
</blockquote></details>

I think it matters if the record is composed of multiple chunks. `singleReader.Read` can call `nextChunk` which will update `r.begin`, `r.end` and `r.n`. Expanded this comment to explicitly discuss the need to read all chunks.
___
*[`wal/testdata/reader` line 3 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3276#-Nq5-AubDgLm5RtGaAMf:-Nq8p3Ye-GMSMrfMxbHk:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/901a75db0eb59b7da531a6167d29e0f25605ae61/wal/testdata/reader#L3)):*

nice test!

can you add a case where the first log has batch0, batch1, batch2; second log has batch1; third log has batch1, batch2, batch3.
</blockquote></details>

Done.
___
*[`wal/testdata/reader` line 29 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3276#-Nq4zJKZEY166K33prso:-Nq8ne3o8ViEgXzzt_eh:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/901a75db0eb59b7da531a6167d29e0f25605ae61/wal/testdata/reader#L29)):*

nit: contains
</blockquote></details>

Done.
___
*[`wal/testdata/reader` line 257 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3276#-Nq5-3hPATM2bAq-Cwjf:-Nq8ninX6KPIMVzp7ede:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/901a75db0eb59b7da531a6167d29e0f25605ae61/wal/testdata/reader#L257)):*

nit: segment
</blockquote></details>

Done.

PR #3277, opened 2024-02-06
db: skip LogData when building flushable batch iterators

### LINKED ISSUE #3275: metamorphic: test LogData
LogData has some subtleties related to sharing other batches' sequence numbers. See #3274. We should add LogData as an operation within our metamorphic tests.

jbowens:
**db: skip LogData when building batch iterators**

Previously, if a batch containing a LogData was sufficiently large to be
considered a 'large batch' and enqueued among the in-memory flushables, it
would be surfaced to the eventual memtable flush. With the current compaction
iterator assertions, the flush would fail when it observed an unexpected key
kind. The `Batch.refreshMemTableSize` and `Batch.Apply` functions also had bugs
in which they failed to consider LogData keys, which meant that a batch only
containing a single large LogData could be considered a large batch if it was
committed through these code paths.

This commit adjusts all memtable-size calculations to skip LogDatas and error
on key kinds not explicitly handled. It also adjusts the construction of a
flushable batch to avoid indexing the offsets of LogData entries within a
batch, hiding all LogDatas from flush iteration.

**metamorphic: add Writer.LogData operation**

Exercise the Writer.LogData operation within the metamorphic tests.

Close #3275.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -471,16 +471,29 @@ func (b *Batch) refreshMemTableSize() error {
 			b.countRangeDels++
 		case InternalKeyKindRangeKeySet, InternalKeyKindRangeKeyUnset, InternalKeyKindRangeKeyDelete:
 			b.countRangeKeys++
+		case InternalKeyKindSet, InternalKeyKindDelete, InternalKeyKindMerge, InternalKeyKindSingleDelete, InternalKeyKindSetWithDelete:
+			// fallthrough
 		case InternalKeyKindDeleteSized:
 			if b.minimumFormatMajorVersion < FormatDeleteSizedAndObsolete {
 				b.minimumFormatMajorVersion = FormatDeleteSizedAndObsolete
 			}
+		case InternalKeyKindLogData:
+			// LogData does not contribute to memtable size.
+			continue
 		case InternalKeyKindIngestSST:
 			if b.minimumFormatMajorVersion < FormatFlushableIngest {
 				b.minimumFormatMajorVersion = FormatFlushableIngest
 			}
 			// This key kind doesn't contribute to the memtable size.
 			continue
+		default:
+			// Note In some circumstances this might be temporary memory
+			// corruption that can be recovered by discarding the batch and
+			// trying again. In other cases, the batch repr might've been
+			// already persisted elsewhere, and we'll loop continuously trying
+			// to commit the same corrupted batch. The caller is responsible for
+			// distinguishing.
+			return errors.Wrapf(ErrInvalidBatch, "unrecognized kind %v", kind)
 		}
 		b.memTableSize += memTableEntrySize(len(key), len(value))
 	}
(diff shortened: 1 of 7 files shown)


sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 1 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3277#-:-Nq4F3wU4fxsL6gbRCpB:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @aadityasondhi and @jbowens)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3277#-Nq4Ea6o8tTi9wrC_zdU):*
any unit tests for these changes? Like the size calculation.
___
*[`batch.go` line 551 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3277#-Nq4E7yW1jr4c0EKFY6u:-Nq4E7yW1jr4c0EKFY6v:b-umdoo6) ([raw file](https://github.com/cockroachdb/pebble/blob/5c67f78972728137be698def868f0352578826c4/batch.go#L551)):*
> ```Go
> 				// again. In other cases, the batch repr might've been already
> 				// persisted elsewhere, and we'll loop continuously trying to
> 				// commit the same corrupted batch.
> ```

This comment is helpful, but should it be a TODO? It seems the caller is the one with enough context to decide whether to retry or fail.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3277#-:-Nq8hY2I9nYAXww-O82S:b-814g8k)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @aadityasondhi and @sumeerbhola)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3277#-Nq4Ea6o8tTi9wrC_zdU):*

any unit tests for these changes? Like the size calculation.
</blockquote></details>

added a simple unit test for memtable size calculations.
___
*[`batch.go` line 551 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3277#-Nq4E7yW1jr4c0EKFY6u:-Nq8fDUk0r9IHv87dV65:b-eoy68e) ([raw file](https://github.com/cockroachdb/pebble/blob/5c67f78972728137be698def868f0352578826c4/batch.go#L551)):*

This comment is helpful, but should it be a TODO? It seems the caller is the one with enough context to decide whether to retry or fail.
</blockquote></details>

Removed the TODO and improved the comments on the user-facing exported methods to make this responsibility clearer. I also updated all these instances to consistently return `ErrInvalidBatch`. This error is an `ErrCorruption` which we might want to revisit so that callers can rely on corruption errors meaning true corruption of the storage engine state, not the user asking the storage engine to write something that doesn't make sense.

PR #3279, opened 2024-02-07
testkeys: replace RandomSeparator

RaduBerinde:
The `RandomSeparator` function is complex and very fragile: it can
fail and return `nil`, and it can return incorrect results if one of
the input keys is longer than ~13 characters (because of overflows in
the `Alpha` keyspace calculations).

In this commit we:
 - add overflow assertions to `keyCount`
 - replace `RandomSeparator` with a simpler `RandomPrefixInRange`
   which returns a prefix in the `[a, b)` range. This function cannot
   fail and supports any key lengths.
 - change the metamorphic key generator to add a suffix to the
   generated prefix.

### DIFF
--- a/internal/testkeys/testkeys.go
+++ b/internal/testkeys/testkeys.go
@@ -17,6 +17,7 @@ import (
 	"cmp"
 	"fmt"
 	"math"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


RaduBerinde:
Friendly ping. Thanks!

jbowens (approved):
:lgtm:

> Reviewed 2 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3279#-:-NqY1Bq9Djy9ZlhjUVq4:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #3280, opened 2024-02-08
ingest: fix external backing mix-up during ingest

RaduBerinde:
#### db: minor fix to external ingestion

One of the validation checks was in the wrong place, leading to
errors.

#### ingest: fix external backing mix-up during ingest

During ingest we sort the `fileMetadata` for the external files
and then incorrectly assume they map 1-1 to the given `[]ExternalFile`.

This commit fixes this by refactoring the `loadResult` structure to
contain only three slices (for local, shared, and external) which
contain all the relevant information. We can now use `slices.SortFunc`
to sort and don't have to pass around the extra information for shared
and external files.

### DIFF
--- a/flushable_test.go
+++ b/flushable_test.go
@@ -62,7 +62,10 @@ func TestIngestedSSTFlushableAPI(t *testing.T) {
 		if err != nil {
 			panic(err)
 		}
-		meta := lr.localMeta
+		meta := make([]*fileMetadata, len(lr.local))
+		for i := range meta {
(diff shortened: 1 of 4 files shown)


RaduBerinde:
Friendly ping. Thanks!

itsbilal (approved):
LGTM, thanks for the cleanup and apologies for the delay!

RaduBerinde:
TFTR!

PR #3282, opened 2024-02-08
keyspan: simplify Truncate, Filter

RaduBerinde:
#### keyspan: simplify and reimplement Truncate

We can simplify the Truncate operator, as we will never truncate
within the same user key. It now only takes user key bounds and is
much easier to reason about.

We also reimplement it so that we can simplify Filter.

#### keyspan: simplify Filter

The `FilterFunc` contract is not adequately spelled out; the code
makes certain unverified assumptions about how the bounds can be
changed.

Now that Truncate is a separate iterator, we no longer need Filter to
allow for changing bounds. We simplify `FilterFunc` to just return the
list of keys that are retained. We no longer need the extra
comparisons in `SeekGE/SeekLT`.

### DIFF
--- a/internal/keyspan/filter.go
+++ b/internal/keyspan/filter.go
@@ -6,13 +6,10 @@ package keyspan
 
 import "github.com/cockroachdb/pebble/internal/base"
 
-// FilterFunc defines a transform from the input Span into the output Span. The
-// function returns true if the Span should be returned by the iterator, and
-// false if the Span should be skipped. The FilterFunc is permitted to mutate
-// the output Span, for example, to elice certain keys, or update the Span's
-// bounds if so desired. The output Span's Keys slice may be reused to reduce
-// allocations.
-type FilterFunc func(in *Span, out *Span) (keep bool)
+// FilterFunc is a callback that allows filtering keys from a Span. The result
+// is the set of keys that should be retained (using buf as a buffer). If the
+// result has no keys, the span is skipped altogether.
+type FilterFunc func(span *Span, buf []Key) []Key
 
 // filteringIter is a FragmentIterator that uses a FilterFunc to select which
 // Spans from the input iterator are returned in the output.
@@ -45,17 +42,7 @@ func (i *filteringIter) SeekGE(key []byte) (*Span, error) {
 	if err != nil {
 		return nil, err
 	}
-	s, err = i.filter(s, +1)
-	if err != nil {
-		return nil, err
-	}
-	// i.filter could return a span that's less than key, _if_ the filterFunc
-	// (which has no knowledge of the seek key) mutated the span to end at a key
-	// less than or equal to `key`. Detect this case and next/invalidate the iter.
-	if s != nil && i.cmp(s.End, key) <= 0 {
-		return i.Next()
-	}
-	return s, nil
+	return i.filter(s, +1)
 }
 
 // SeekLT implements FragmentIterator.
@@ -64,17 +51,7 @@ func (i *filteringIter) SeekLT(key []byte) (*Span, error) {
 	if err != nil {
 		return nil, err
 	}
-	span, err = i.filter(span, -1)
-	if err != nil {
-		return nil, err
-	}
-	// i.filter could return a span that's >= key, _if_ the filterFunc (which has
-	// no knowledge of the seek key) mutated the span to start at a key greater
-	// than or equal to `key`. Detect this case and prev/invalidate the iter.
-	if span != nil && i.cmp(span.Start, key) >= 0 {
-		return i.Prev()
-	}
-	return span, nil
+	return i.filter(span, -1)
 }
 
 // First implements FragmentIterator.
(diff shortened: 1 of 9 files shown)


RaduBerinde:
Interesting, I stressed the regular meta test but there is a failure in the crossversion one.

RaduBerinde:
I reworked this to use the new `UserKeyBounds`.

RaduBerinde:
Friendly ping.

jbowens (commented):
> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3282)* status: 6 of 9 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`internal/keyspan/truncate.go` line 129 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3282#-Nta_oA80pAauJLdQN9O:-Nta_oA80pAauJLdQN9P:b-s92kir) ([raw file](https://github.com/cockroachdb/pebble/blob/dba88bc95dea5e9912b9cf89c369eb9d1736484f/internal/keyspan/truncate.go#L129)):*
> ```Go
> 	for span != nil {
> 		if i.bounds.End.Kind == base.Inclusive && span.Contains(i.cmp, i.bounds.End.Key) {
> 			err := errors.AssertionFailedf("inclusive upper bound %q inside span %s", i.bounds.End.Key, span)
> ```

I think I must be being dense, but I didn't understand this assertion. If we have bounds `[a, b]` and call Last() on an sstable iterator, what's to prevent the span from having bounds `[a,c)` and hitting this condition on the first iteration?

EDIT: Ahh, I think I see—we're expecting the caller to externally guarantee this invariant. I guess I'm still confused as to how we provide that invariant?
___
*[`internal/keyspan/truncate.go` line 135 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3282#-NtaaMsw-quSLbovoOSM:-NtaaMsw-quSLbovoOSN:bhslun6) ([raw file](https://github.com/cockroachdb/pebble/blob/dba88bc95dea5e9912b9cf89c369eb9d1736484f/internal/keyspan/truncate.go#L135)):*
> ```Go
> 			return nil, false, err
> 		}
> 		// Intersect [span.Start, span.End) with [lower, upper).
> ```

nit: s/[lower, upper)/i.bounds/?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3282)* status: 6 of 9 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`internal/keyspan/truncate.go` line 129 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3282#-Nta_oA80pAauJLdQN9O:-Ntaknw00hTdHKRFYBBw:b-8iifmy) ([raw file](https://github.com/cockroachdb/pebble/blob/dba88bc95dea5e9912b9cf89c369eb9d1736484f/internal/keyspan/truncate.go#L129)):*
Right, it's mentioned for Truncate:

> // Note that fragment iterator Spans always have exclusive end-keys; if the
// given bounds have an inclusive end key the input iterator must not produce a
// span that contains that key.

Basically if you hit this case, you are doing something wrong. In practice this currently means that you have  and inclusive end bound for a virtual table (i.e. `Largest` is not a sentinel) and the sst has a RangeDel/RangeKey that contains that key. This is a bug as there's no way to express that the key is "covered" by the RangeDel while staying within the bounds.

Looking at it another way, the only reason a caller would provide an inclusive end bound to `Truncate` is to perform this check. It's equivalent to the previous `panicOnUpperTruncate`, just feels more elegant.

jbowens (approved):
:lgtm:

> Reviewed 1 of 3 files at r3, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3282#-:-NtbSkzyA-wVvUWvFB-e:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`internal/keyspan/truncate.go` line 129 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3282#-Nta_oA80pAauJLdQN9O:-NtbS82e9KHho_nPApTy:b-vo0hen) ([raw file](https://github.com/cockroachdb/pebble/blob/dba88bc95dea5e9912b9cf89c369eb9d1736484f/internal/keyspan/truncate.go#L129)):*

Right, it's mentioned for Truncate:

> // Note that fragment iterator Spans always have exclusive end-keys; if the
// given bounds have an inclusive end key the input iterator must not produce a
// span that contains that key.

Basically if you hit this case, you are doing something wrong. In practice this currently means that you have  and inclusive end bound for a virtual table (i.e. `Largest` is not a sentinel) and the sst has a RangeDel/RangeKey that contains that key. This is a bug as there's no way to express that the key is "covered" by the RangeDel while staying within the bounds.

Looking at it another way, the only reason a caller would provide an inclusive end bound to `Truncate` is to perform this check. It's equivalent to the previous `panicOnUpperTruncate`, just feels more elegant.
</blockquote></details>

Got it, thanks

RaduBerinde:
TFTR!

PR #3283, opened 2024-02-08
wal: failoverWriter, for switching across a sequence of record.LogWri…

sumeerbhola:
…ters

Except for Close, failoverWriter methods do not block on IO. Close can block on IO if we run out of failover slots (10 slots) and new LogWriters cannot be created.

Informs #3230

Informs CRDB-35401

### DIFF
--- a/db.go
+++ b/db.go
@@ -2727,7 +2727,7 @@ func (d *DB) recycleWAL() (newLogNum base.DiskFileNum, prevLogSize uint64) {
 	metrics := d.mu.log.writer.Metrics()
 
 	d.mu.Lock()
-	if err := d.mu.log.metrics.LogWriterMetrics.Merge(metrics); err != nil {
+	if err := d.mu.log.metrics.LogWriterMetrics.Merge(&metrics); err != nil {
 		d.opts.Logger.Errorf("metrics error: %s", err)
 	}
 
--- a/record/log_writer.go
+++ b/record/log_writer.go
@@ -1001,10 +1001,15 @@ func (w *LogWriter) emitFragment(n int, p []byte) (remainingP []byte) {
 	return p[r:]
 }
 
-// Metrics must be called after Close. The callee will no longer modify the
-// returned LogWriterMetrics.
-func (w *LogWriter) Metrics() *LogWriterMetrics {
-	return w.flusher.metrics
+// Metrics must typically be called after Close, since the callee will no
+// longer modify the returned LogWriterMetrics. It is also current if there is
+// nothing left to flush in the flush loop, but that is an implementation
+// detail that callers should not rely on.
+func (w *LogWriter) Metrics() LogWriterMetrics {
+	w.flusher.Lock()
+	defer w.flusher.Unlock()
+	m := *w.flusher.metrics
+	return m
 }
 
 // LogWriterMetrics contains misc metrics for the log writer.
--- a/vfs/errorfs/errorfs.go
+++ b/vfs/errorfs/errorfs.go
@@ -81,6 +81,8 @@ const (
 	OpFileStat
 	// OpFileSync describes a file sync operation.
 	OpFileSync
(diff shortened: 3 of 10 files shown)


jbowens (approved):
:lgtm:

> Reviewed 8 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3283#-:-NqEqzBb53v_-nMZRz6l:bnfp4nl)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`wal/failover_writer.go` line 188 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3283#-NqEo3K13mwVPkQsFSma:-NqEo3K13mwVPkQsFSmb:bhen3er) ([raw file](https://github.com/cockroachdb/pebble/blob/5d58f1d195e303aca4a96dfe3b0f23ee4768b899/wal/failover_writer.go#L188)):*
> ```Go
> 		h, t = unpackHeadTail(ht)
> 		tail = int(t)
> 		maxEntriesToPop = int(index) - tail + 1
> ```

`maxEntriesToPop` shouldn't be changing right? Only we are popping, and the only reason a CAS might fail is appends to the head? (and that's what makes it okay that `b` is untouched when we fail the CAS) a comment would help clarify
___
*[`wal/failover_writer.go` line 321 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3283#-NqEqcAE6CUBD1L4ckpq:-NqEqcAE6CUBD1L4ckpr:bcfjqri) ([raw file](https://github.com/cockroachdb/pebble/blob/5d58f1d195e303aca4a96dfe3b0f23ee4768b899/wal/failover_writer.go#L321)):*
> ```Go
> 	// synchronization, since we rely on external synchronization (like the
> 	// standaloneWriter).
> 	logicalOffset struct {
> ```

I have an inkling that we could simplify by being less meticulous with our accounting of logicalOffset, but it's always something we could explore later:

https://github.com/cockroachdb/pebble/pull/3224#pullrequestreview-1873025950

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3283#-:-NqO-wpc8y3ME1esHne7:b-814g8k)* status: 6 of 10 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`wal/failover_writer.go` line 188 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3283#-NqEo3K13mwVPkQsFSma:-NqO-VX77C7mE15CeODk:b-9akq6g) ([raw file](https://github.com/cockroachdb/pebble/blob/5d58f1d195e303aca4a96dfe3b0f23ee4768b899/wal/failover_writer.go#L188)):*

`maxEntriesToPop` shouldn't be changing right? Only we are popping, and the only reason a CAS might fail is appends to the head? (and that's what makes it okay that `b` is untouched when we fail the CAS) a comment would help clarify
</blockquote></details>

It's the race with other consumers, since any consumer that was given a record is allowed to pop it.
I've simplified this and added some commentary.
___
*[`wal/failover_writer.go` line 321 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3283#-NqEqcAE6CUBD1L4ckpq:-NqO-gMpEQqMapF0cORW:b27wucm) ([raw file](https://github.com/cockroachdb/pebble/blob/5d58f1d195e303aca4a96dfe3b0f23ee4768b899/wal/failover_writer.go#L321)):*

I have an inkling that we could simplify by being less meticulous with our accounting of logicalOffset, but it's always something we could explore later:

https://github.com/cockroachdb/pebble/pull/3224#pullrequestreview-1873025950
</blockquote></details>

I responded on the other review. I agree we probably need to revisit how we do the metric for this.

PR #3240, opened 2024-01-23
sstable: teach pebble about suffix replacement

msbutler:
See individual commits.

PR #3292, opened 2024-02-12
rename FormatSyntheticSuffixes major format version to FormatSynethicSuffixPrefix

msbutler:
PR https://github.com/cockroachdb/pebble/pull/3240 introduced synthetic suffix replacement, whose use should be gated on 
a major format version. This follow up patch gates its use on the same major
format version added for prefix replacement and renames the version to
FormatSyntheticSuffixPrefix. This patch also adds more testing around synthetic
suffix ingestion.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -1283,6 +1283,7 @@ func runIngestExternalCmd(td *datadriven.TestData, d *DB, locator string) error 
 	external := make([]ExternalFile, 0)
 	usePrefixChange := false
 	var fromPrefix, toPrefix []byte
+	var syntheticSuffix []byte
 	for i := range td.CmdArgs {
 		switch td.CmdArgs[i].Key {
 		case "prefix-replace":
@@ -1293,6 +1294,8 @@ func runIngestExternalCmd(td *datadriven.TestData, d *DB, locator string) error 
 			fromPrefix = []byte(vals[0])
 			toPrefix = []byte(vals[1])
 			usePrefixChange = true
+		case "suffix-replace":
+			syntheticSuffix = []byte(td.CmdArgs[i].Vals[0])
 		}
 	}
 	for _, arg := range strings.Split(td.Input, "\n") {
@@ -1315,6 +1318,7 @@ func runIngestExternalCmd(td *datadriven.TestData, d *DB, locator string) error 
 			ef.ContentPrefix = fromPrefix
 			ef.SyntheticPrefix = toPrefix
 		}
+		ef.SyntheticSuffix = syntheticSuffix
 		external = append(external, ef)
 	}
 
--- a/format_major_version.go
+++ b/format_major_version.go
@@ -178,12 +178,12 @@ const (
 	// a format major version.
 	FormatVirtualSSTables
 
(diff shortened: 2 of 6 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3292#-:-NqUfMV8-uK5TFMRARJb:bnfp4nl)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @msbutler and @sumeerbhola)

___
*[`format_major_version.go` line 193 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3292#-NqUf45z7pEJJqi4Vlw0:-NqUf45z7pEJJqi4Vlw1:b-3wf63a) ([raw file](https://github.com/cockroachdb/pebble/blob/fba89c99d5ae199afd0626b6d1c9ba9ec8523047/format_major_version.go#L193)):*
> ```Go
> 	// information is stored in new fields in the Manifest and thus
> 	// requires a format major version.
> 	FormatSyntheticSuffixes
> ```

You can also merge into the previous version and make it `FormatSyntheticPrefixSuffix` since that was added recently as well.
___
*[`ingest.go` line 1333 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3292#-NqUfHbr43Vc3Y2CCiEV:-NqUfHbr43Vc3Y2CCiEW:b-c15w3z) ([raw file](https://github.com/cockroachdb/pebble/blob/fba89c99d5ae199afd0626b6d1c9ba9ec8523047/ingest.go#L1333)):*
> ```Go
> 		for i := range external {
> 			if len(external[i].SyntheticSuffix) > 0 {
> 				return IngestOperationStats{}, errors.New("pebble: format major version too old for synthetic suffix ingestion")
> ```

[nit] it would be nice to have a test for this

msbutler (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3292)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`format_major_version.go` line 193 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3292#-NqUf45z7pEJJqi4Vlw0:-NqUnl6P342yS4EJAigA:b47rol5) ([raw file](https://github.com/cockroachdb/pebble/blob/fba89c99d5ae199afd0626b6d1c9ba9ec8523047/format_major_version.go#L193)):*

You can also merge into the previous version and make it `FormatSyntheticPrefixSuffix` since that was added recently as well.
</blockquote></details>

o lol. that's much simpler.

RaduBerinde (commented):
:lgtm:

> Reviewed 2 of 7 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3292#-:-NqYWSZ0AHSRkpAspTCX:bnfp4nl)* status: 6 of 10 files reviewed, all discussions resolved (waiting on @sumeerbhola)

Issue #3289, opened 2024-02-12
wal: TestFailoverWriter is flaky on windows


jbowens:
```
2024-02-12T20:39:05.8678369Z         	Error Trace:	D:/a/pebble/pebble/wal/failover_writer_test.go:227
2024-02-12T20:39:05.8678908Z         	            				D:/a/pebble/pebble/wal/failover_writer_test.go:350
2024-02-12T20:39:05.8680212Z         	            				C:/Users/runneradmin/go/pkg/mod/github.com/cockroachdb/datadriven@v1.0.3-0.20230413201302-be42291fc80f/datadriven.go:117
2024-02-12T20:39:05.8681482Z         	            				C:/Users/runneradmin/go/pkg/mod/github.com/cockroachdb/datadriven@v1.0.3-0.20230413201302-be42291fc80f/datadriven.go:336
2024-02-12T20:39:05.8682748Z         	            				C:/Users/runneradmin/go/pkg/mod/github.com/cockroachdb/datadriven@v1.0.3-0.20230413201302-be42291fc80f/datadriven.go:341
2024-02-12T20:39:05.8684012Z         	            				C:/Users/runneradmin/go/pkg/mod/github.com/cockroachdb/datadriven@v1.0.3-0.20230413201302-be42291fc80f/datadriven.go:210
2024-02-12T20:39:05.8685272Z         	            				C:/Users/runneradmin/go/pkg/mod/github.com/cockroachdb/datadriven@v1.0.3-0.20230413201302-be42291fc80f/datadriven.go:183
2024-02-12T20:39:05.8686523Z         	            				C:/Users/runneradmin/go/pkg/mod/github.com/cockroachdb/datadriven@v1.0.3-0.20230413201302-be42291fc80f/datadriven.go:143
2024-02-12T20:39:05.8687815Z         	            				C:/Users/runneradmin/go/pkg/mod/github.com/cockroachdb/datadriven@v1.0.3-0.20230413201302-be42291fc80f/datadriven.go:116
2024-02-12T20:39:05.8688346Z         	            				D:/a/pebble/pebble/wal/failover_writer_test.go:175
2024-02-12T20:39:05.8689627Z         	            				C:/Users/runneradmin/go/pkg/mod/github.com/cockroachdb/datadriven@v1.0.3-0.20230413201302-be42291fc80f/datadriven.go:424
[...]

PR #3293, opened 2024-02-13
keyspan: move MergingIter and LevelIter to new keyspanimpl package

RaduBerinde:
This change leaves only generic concepts and implementation in
`keyspan` which don't depend on Pebble internals (in particular the
`manifest` package).


RaduBerinde:
Windows failure is #3289.

jbowens (approved):
I've previously considered moving `keyspan.Span`, `keyspan.Key`, `keyspan.FragmentIterator` into `internal/base` as `Span`, `SpanKey` and `SpanIterator`. Would that also achieve a similar end?

Regardless, :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3293#-:-NqY-OtHCFvEWkygwOGH:b-o89nh)* status: 0 of 31 files reviewed, all discussions resolved (waiting on @itsbilal)

RaduBerinde (commented):
I believe so. It looks like `sstable` only depends on those. Do you prefer I do that?

I kind of like having more packages instead of adding more stuff to base. If anything, I'd change `keyspan.Span` to `keyspan.Fragment`.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3293#-:-NqY1Pos2qAIv4xnIgJs:b-1nt6e)* status: 0 of 31 files reviewed, all discussions resolved (waiting on @itsbilal)

jbowens (commented):
Naw, I think this is good. The distinction between X and Ximpl can be a little ambiguous, but it's mostly aesthetic.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3293#-:-NqYmpNCCmwFOk54pshv:b-ahqqat)* status: 0 of 31 files reviewed, all discussions resolved (waiting on @itsbilal)

internal/keyspan/test_utils.go
```diff
@@ -5,13 +5,13 @@
...
 
 import (
+	"bytes"
 	"fmt"
 	"go/token"
 	"io"
 	"reflect"
 	"strconv"
 	"strings"
-	"testing"
```
dt, line 14: while this doesn't depend `testing`, it does still depend on `datadriven` on the line below this, so moving it from _test.go to a linked file meant we picked up datadriven's init-time CLI flag registrations

dt: #3326

PR #3295, opened 2024-02-13
metamorphic: add external files ingest operation

RaduBerinde:
#### metamorphic: add external files ingest operation

This commit adds two new operations:
 - `newExternalObjOp` transforms a batch into an external object.
 - `ingestExtternalFilesOp` ingests random parts of random external
   objects. If external storage is not enabled, this operation is
   emulated by creating sst files in the local FS and ingesting those.

The support of external storage is a setting independent of the shared
storage setting, allowing us to test more combinations. The external
storage uses the `"external"` locator whereas shared storage uses the
empty locator.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -181,6 +181,10 @@ func ingestLoad1External(
 	if !e.HasRangeKey && !e.HasPointKey {
 		return nil, errors.New("pebble: cannot ingest external file with no point or range keys")
 	}
+	// #3287: range keys don't yet work correctly when the range key bounds are not tight.
+	if e.HasRangeKey {
+		return nil, errors.New("pebble: range keys not supported in external files")
+	}
 	// Don't load table stats. Doing a round trip to shared storage, one SST
 	// at a time is not worth it as it slows down ingestion.
 	meta := &fileMetadata{
--- a/level_iter.go
+++ b/level_iter.go
@@ -1200,7 +1200,7 @@ func (l *levelIter) SetContext(ctx context.Context) {
 
 func (l *levelIter) String() string {
 	if l.iterFile != nil {
-		return fmt.Sprintf("%s: fileNum=%s", l.level, l.iter.String())
+		return fmt.Sprintf("%s: fileNum=%s", l.level, l.iterFile.FileNum.String())
 	}
 	return fmt.Sprintf("%s: fileNum=<nil>", l.level)
 }
--- a/metamorphic/build.go
+++ b/metamorphic/build.go
@@ -5,7 +5,9 @@
 package metamorphic
 
 import (
+	"context"
 	"fmt"
+	"slices"
 
 	"github.com/cockroachdb/pebble"
 	"github.com/cockroachdb/pebble/internal/base"
@@ -32,7 +34,6 @@ func writeSSTForIngestion(
 	rangeKeyIter keyspan.FragmentIterator,
 	writable objstorage.Writable,
 	targetFMV pebble.FormatMajorVersion,
-	bounds pebble.KeyRange,
 ) (*sstable.WriterMetadata, error) {
 	writerOpts := t.opts.MakeWriterOptions(0, targetFMV.MaxTableFormat())
 	if t.testOpts.disableValueBlocksForIngestSSTables {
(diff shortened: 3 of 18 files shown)


RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3295#-:-NqdQgV6-CluLh_Upi-_:b-814g8k)* status: 0 of 18 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`metamorphic/build.go` line 78 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3295#-NqcHNIs9KarXO4Qoq6M:-NqdNSjLDHIKaO-XLAdF:bmxrlt3) ([raw file](https://github.com/cockroachdb/pebble/blob/b07fd593f5e862371c65a65790115f655bf767ac/metamorphic/build.go#L78)):*

nit: slices.Clone could make it a little simpler
</blockquote></details>

Done, much nicer, thanks!
___
*[`metamorphic/ops.go` line 28 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3295#-NqcMvyF79XzY0Fy4A51:-NqdNdRw9-jiM2evfBSY:b-l1ivvf) ([raw file](https://github.com/cockroachdb/pebble/blob/3308a5048c519cfb38a42b3e4ba3d2c9c62f8a05/metamorphic/ops.go#L28)):*

this can be "slices" now that it's a part of the standard library, right?
</blockquote></details>

Fixed, goland keeps adding this one instead.
___
*[`sstable/reader_iter_single_lvl.go` line 1105 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3295#-NqcVyiI5jLVNTO_3axF:-NqdOqxk9TGH2gldXSXB:b-rezqyw) ([raw file](https://github.com/cockroachdb/pebble/blob/3308a5048c519cfb38a42b3e4ba3d2c9c62f8a05/sstable/reader_iter_single_lvl.go#L1105)):*

I like this simplification but it's contrary to the godoc comment on `First` (and the contract of the internalIterator). We should update the comment here to explicitly call out that although the internalIterator contract dictates that the caller is reponsible for using SeekGE(k) with a k ≥ lower if necessary, the caller only knows it relative to the virtual bounds in the case of a virtual sstable.
</blockquote></details>

Updated the comment. Can you point me to the godoc? I couldn't find anything in the public APIs, just the comment for [InternalIterator](https://github.com/cockroachdb/pebble/blob/master/internal/base/iterator.go#L138)
___
*[`sstable/reader_iter_two_lvl.go` line 725 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3295#-NqcWdUj8DscMWRD7R-n:-NqdQhowBvM_CKe6uZ97:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/11b5d32f8eda5e3692879a85a8d1be9f883b419b/sstable/reader_iter_two_lvl.go#L725)):*

ditto about the comment
</blockquote></details>

Done.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3295)* status: 0 of 18 files reviewed, 3 unresolved discussions

___
*[`sstable/reader_iter_single_lvl.go` line 1105 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3295#-NqcVyiI5jLVNTO_3axF:-NqhOMafAp6uHW6AfdO5:b-mouemr) ([raw file](https://github.com/cockroachdb/pebble/blob/3308a5048c519cfb38a42b3e4ba3d2c9c62f8a05/sstable/reader_iter_single_lvl.go#L1105)):*

Updated the comment. Can you point me to the godoc? I couldn't find anything in the public APIs, just the comment for [InternalIterator](https://github.com/cockroachdb/pebble/blob/master/internal/base/iterator.go#L138)
</blockquote></details>

sorry, I just meant the comment right above the method signature: 

```
// First implements internalIterator.First, as documented in the pebble
// package. Note that First only checks the upper bound. It is up to the caller
// to ensure that key is greater than or equal to the lower bound (e.g. via a
// call to SeekGE(lower)).
```

PR #3298, opened 2024-02-13
db: remove syntheticSuffix from loadInfo

RaduBerinde:
This looks like an accidental leftover (it's not used anywhere).


msbutler (approved):
Ah good catch!

PR #3299, opened 2024-02-13
iterator: require prefix filtering in merging iterator

### LINKED ISSUE #2182: db: consider requiring prefix filtering within InternalIterator
This came up during [#2178](https://github.com/cockroachdb/pebble/pull/2178#pullrequestreview-1210803906).

Today, `SeekPrefixGE` is not required to return a key that matches the provided prefix. Levels that can't cheaply (eg, bloom-filter check) determine whether a level contains a prefix behave like a `SeekGE`. This avoids the need to compare prefixes in all of the leaf iterators.

However, every leaf iterator that returns a key increases the work required to build the merging heap. In the case where only zero or one levels contain a key with the provided prefix, checking the prefix at the leaf iterators would result in _n_ prefix checks. This is in contrast to the current 1 top-level prefix check + _n_ key comparisons to build the heap. If all of the keys with the prefix reside in a single level, iteration over the prefix also benefits from the reduced heap size removing _k_ log _n_ comparisons, where _k_ is the number of times the iterator is stepped.

In the common case where one levels contains relevant keys, the seek would only move the single prefix check into the leaf iterator and remove the log _n_ key comparisons.

CheranMahalingam:
Currently for `SeekPrefixGE` calls we filter out keys without a matching prefix at the top layer iterator. This causes the merging iterator to build a larger heap containing entries without the required prefixes. This commit shifts the prefix filtering responsibility to the merging iterator before inserting keys in the heap.

Fixes #2182.

### DIFF
--- a/db.go
+++ b/db.go
@@ -1483,7 +1483,7 @@ func (i *Iterator) constructPointIter(
 	buf.merging.snapshot = i.seqNum
 	buf.merging.batchSnapshot = i.batchSeqNum
 	buf.merging.combinedIterState = &i.lazyCombinedIter.combinedIterState
-	i.pointIter = invalidating.MaybeWrapIfInvariants(&buf.merging)
+	i.pointIter = invalidating.MaybeWrapIfInvariants(&buf.merging).(topLevelIterator)
 	i.merging = &buf.merging
 }
 
--- a/error_iter.go
+++ b/error_iter.go
@@ -25,6 +25,12 @@ func (c *errorIter) SeekGE(key []byte, flags base.SeekGEFlags) (*InternalKey, ba
 func (c *errorIter) SeekPrefixGE(
 	prefix, key []byte, flags base.SeekGEFlags,
 ) (*base.InternalKey, base.LazyValue) {
+	return c.SeekPrefixGEStrict(prefix, key, flags)
+}
+
+func (c *errorIter) SeekPrefixGEStrict(
+	prefix, key []byte, flags base.SeekGEFlags,
+) (*base.InternalKey, base.LazyValue) {
 	return nil, base.LazyValue{}
 }
 
--- a/external_iterator.go
+++ b/external_iterator.go
@@ -166,7 +166,7 @@ func validateExternalIterOpts(iterOpts *IterOptions) error {
 	return nil
 }
 
-func createExternalPointIter(ctx context.Context, it *Iterator) (internalIterator, error) {
+func createExternalPointIter(ctx context.Context, it *Iterator) (topLevelIterator, error) {
 	// TODO(jackson): In some instances we could generate fewer levels by using
 	// L0Sublevels code to organize nonoverlapping files into the same level.
 	// This would allow us to use levelIters and keep a smaller set of data and
@@ -235,15 +235,6 @@ func createExternalPointIter(ctx context.Context, it *Iterator) (internalIterato
 				rangeDelIter: nil,
 			})
 		}
-	}
-	if len(mlevels) == 1 && mlevels[0].rangeDelIter == nil {
-		// Set closePointIterOnce to true. This is because we're bypassing the
-		// merging iter, which turns Close()s on it idempotent for any child
-		// iterators. The outer Iterator could call Close() on a point iter twice,
-		// which sstable iterators do not support (as they release themselves to
-		// a pool).
-		it.closePointIterOnce = true
-		return mlevels[0].iter, nil
 	}
 
 	it.alloc.merging.init(&it.opts, &it.stats.InternalStats, it.comparer.Compare, it.comparer.Split, mlevels...)
--- a/get_iter.go
+++ b/get_iter.go
@@ -55,6 +55,12 @@ func (g *getIter) SeekGE(key []byte, flags base.SeekGEFlags) (*InternalKey, base
 func (g *getIter) SeekPrefixGE(
 	prefix, key []byte, flags base.SeekGEFlags,
 ) (*base.InternalKey, base.LazyValue) {
+	return g.SeekPrefixGEStrict(prefix, key, flags)
+}
+
+func (g *getIter) SeekPrefixGEStrict(
+	prefix, key []byte, flags base.SeekGEFlags,
+) (*base.InternalKey, base.LazyValue) {
 	panic("pebble: SeekPrefixGE unimplemented")
 }
 
--- a/internal.go
+++ b/internal.go
@@ -37,6 +37,8 @@ type InternalKey = base.InternalKey
 
 type internalIterator = base.InternalIterator
 
+type topLevelIterator = base.TopLevelIterator
+
 // ErrCorruption is a marker to indicate that data in a file (WAL, MANIFEST,
 // sstable) isn't in the expected format.
 var ErrCorruption = base.ErrCorruption
--- a/internal/base/iterator.go
+++ b/internal/base/iterator.go
@@ -214,6 +214,16 @@ type InternalIterator interface {
 	fmt.Stringer
 }
 
+// TopLevelIterator extends InternalIterator to include an additional absolute
+// positioning method, SeekPrefixGEStrict.
+type TopLevelIterator interface {
+	InternalIterator
+
+	// SeekPrefixGEStrict extends InternalIterator.SeekPrefixGE with a guarantee
+	// that the iterator only returns keys matching the prefix.
+	SeekPrefixGEStrict(prefix, key []byte, flags SeekGEFlags) (*InternalKey, LazyValue)
+}
+
 // SeekGEFlags holds flags that may configure the behavior of a forward seek.
 // Not all flags are relevant to all iterators.
 type SeekGEFlags uint8
--- a/internal/invalidating/iter.go
+++ b/internal/invalidating/iter.go
@@ -55,7 +55,7 @@ func IgnoreKinds(kinds ...base.InternalKeyKind) Option {
 
 // NewIter constructs a new invalidating iterator that wraps the provided
 // iterator, trashing buffers for previously returned keys.
-func NewIter(originalIterator base.InternalIterator, opts ...Option) base.InternalIterator {
+func NewIter(originalIterator base.InternalIterator, opts ...Option) base.TopLevelIterator {
 	i := &iter{iter: originalIterator}
 	for _, opt := range opts {
 		opt.apply(i)
@@ -115,6 +115,12 @@ func (i *iter) SeekGE(key []byte, flags base.SeekGEFlags) (*base.InternalKey, ba
 }
 
 func (i *iter) SeekPrefixGE(
+	prefix, key []byte, flags base.SeekGEFlags,
+) (*base.InternalKey, base.LazyValue) {
+	return i.update(i.iter.SeekPrefixGE(prefix, key, flags))
+}
+
+func (i *iter) SeekPrefixGEStrict(
 	prefix, key []byte, flags base.SeekGEFlags,
 ) (*base.InternalKey, base.LazyValue) {
 	return i.update(i.iter.SeekPrefixGE(prefix, key, flags))
--- a/iterator.go
+++ b/iterator.go
@@ -187,7 +187,7 @@ type Iterator struct {
 	merge     Merge
 	comparer  base.Comparer
 	iter      internalIterator
-	pointIter internalIterator
+	pointIter topLevelIterator
 	// Either readState or version is set, but not both.
 	readState *readState
 	version   *version
@@ -302,16 +302,6 @@ type Iterator struct {
 	batchJustRefreshed bool
 	// batchOnlyIter is set to true for Batch.NewBatchOnlyIter.
 	batchOnlyIter bool
-	// closePointIterOnce is set to true if this point iter can only be Close()d
-	// once, _and_ closing i.iter and then i.pointIter would close i.pointIter
-	// twice. This is necessary to track if the point iter is an internal iterator
-	// that could release its resources to a pool on Close(), making it harder for
-	// that iterator to make its own closes idempotent.
-	//
-	// TODO(bilal): Update SetOptions to always close out point key iterators when
-	// they won't be used, so that Close() doesn't need to default to closing
-	// point iterators twice.
-	closePointIterOnce bool
 	// Used in some tests to disable the random disabling of seek optimizations.
 	forceEnableSeekOpt bool
 	// Set to true if NextPrefix is not currently permitted. Defaults to false
@@ -529,11 +519,19 @@ func (i *Iterator) findNextEntry(limit []byte) {
 	for i.iterKey != nil {
 		key := *i.iterKey
 
-		if i.hasPrefix {
-			if n := i.split(key.UserKey); !i.equal(i.prefixOrFullSeekKey, key.UserKey[:n]) {
-				return
+		// The topLevelIterator.StrictSeekPrefixGE contract requires that in
+		// prefix mode [i.hasPrefix=t], every point key returned by the internal
+		// iterator must have the current iteration prefix.
+		if invariants.Enabled && i.hasPrefix {
+			// Range keys are an exception to the contract and may return a different
+			// prefix. This case is explicitly handled in the switch statement below.
+			if key.Kind() != base.InternalKeyKindRangeKeySet {
+				if n := i.split(key.UserKey); !i.equal(i.prefixOrFullSeekKey, key.UserKey[:n]) {
+					i.opts.logger.Fatalf("pebble: prefix violation: key %q does not have prefix %q\n", key.UserKey, i.prefixOrFullSeekKey)
+				}
 			}
 		}
+
 		// Compare with limit every time we start at a different user key.
 		// Note that given the best-effort contract of limit, we could avoid a
 		// comparison in the common case by doing this only after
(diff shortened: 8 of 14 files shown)


jbowens (commented):
> Reviewed 1 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3299)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @CheranMahalingam)

___
*[`merging_iter.go` line 617 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3299#-NqcErlaEOXOaGWGjnnD:-NqcErlaEOXOaGWGjnnE:bci2jff) ([raw file](https://github.com/cockroachdb/pebble/blob/f89a30acb3715112fdca620885218aaff319e35e/merging_iter.go#L617)):*
> ```Go
> 		m.heap.items = m.heap.items[:0]
> 		return nil
> 	}
> ```

I think we should be able to move the above check after the call to `m.nextEntry`, so that `maybeNextEntryWithinPrefix` now _requires_ that `l.iterKey.UserKey` has `m.prefix`, and _enforces_ that the iterator is advanced to either the next entry within the prefix or becomes exhausted. This will also allow us to remove the checks at the end of `Next` and `SeekPrefixGE`.

CheranMahalingam (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3299)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CheranMahalingam and @jbowens)

___
*[`merging_iter.go` line 617 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3299#-NqcErlaEOXOaGWGjnnD:-NqcFnIa1DH-64U5hTrV:bpp2aob) ([raw file](https://github.com/cockroachdb/pebble/blob/f89a30acb3715112fdca620885218aaff319e35e/merging_iter.go#L617)):*

I think we should be able to move the above check after the call to `m.nextEntry`, so that `maybeNextEntryWithinPrefix` now _requires_ that `l.iterKey.UserKey` has `m.prefix`, and _enforces_ that the iterator is advanced to either the next entry within the prefix or becomes exhausted. This will also allow us to remove the checks at the end of `Next` and `SeekPrefixGE`.
</blockquote></details>

I agree with that, the issue has been that the metamorphic tests fail and I haven't been able to diagnose why flipping the order would break correctness. For reference this is what those changes might look like (https://github.com/cockroachdb/pebble/compare/master...CheranMahalingam:pebble:prefix_filter_leaf_iterator_exp) and here is the diff from the failing test,
```
===== DIFF =====
_meta/240214-105414.4202603277808/{standard-000,standard-008}
@@ -1868,11 +1868,11 @@
 iter19.SeekLT("aukgxqe@11", "") // [true,"aukgxqe@12","iiyjvgmukboifs",["arxo","axiiua")=>{"@13"="fijbdii","@12"="gyb","@10"="jyatssphsroj"}*] <nil> #1867
 iter19.SeekGE("picgy", "") // [true,"picgy","eyvhxewzt",<no range>*] <nil> #1868
 snap15 = db1.NewSnapshot("ejzsqxtqr", "gtcbhfooj", "hfgelqczzvp", "lhed", "mpyljlnitrq", "vbellfwq") #1869
 db1.Get("aukgxqe@11") // ["emevevtykeij"] <nil> #1870
 iter19.Last() // [true,"zrksmyqfn@7","dhqnerih",<no range>] <nil> #1871
-iter19.SeekPrefixGE("fyoqpvnpm@16") // [true,"fyoqpvnpm@16","mihftvbvdiqdi",<no range>] <nil> #1872
+iter19.SeekPrefixGE("fyoqpvnpm@16") // [false] <nil> #1872
 iter19.SeekGE("impgjoqj@13", "") // [true,"impgjoqj@13","zmmuphftz",["impgjoqj","jftwe")=>{""="gd","@13"="fijbdii","@12"="gyb","@11"="gctssrbnwjxylyqg","@10"="eiddqtykzvbwrfkfek","@8"="vxeudyqyt"}*] <nil> #1873
 iter19.First() // [true,"abkdds@16","xmy",<no range>*] <nil> #1874
 iter19.NextPrefix() // [true,"angmayxmhi",<no point>,["angmayxmhi","arxo")=>{"@13"="fijbdii"}*] <nil> #1875
 iter19.NextPrefix() // [true,"arxo","niiwgklubchzb",["arxo","axiiua")=>{"@13"="fijbdii","@12"="gyb","@10"="jyatssphsroj"}*] <nil> #1876
 iter19.Last() // [true,"zrksmyqfn@7","dhqnerih",<no range>*] <nil> #1877
```

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3299)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @CheranMahalingam and @jbowens)

___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3299#-Nqil6Ix4hXv6gl-sX_b):*
Can't a `pebble.Iterator` in theory be constructed without a `mergingIter`? If we want to disallow that, we should consider making that explicit in the type of the `pointIter`, or some other way of enforcing it.

CheranMahalingam (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3299)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3299#-Nqil6Ix4hXv6gl-sX_b):*

Can't a `pebble.Iterator` in theory be constructed without a `mergingIter`? If we want to disallow that, we should consider making that explicit in the type of the `pointIter`, or some other way of enforcing it.
</blockquote></details>

That makes sense, are there any cases where we would explicitly avoid using a `mergingIter` as the `pointIter` or is it purely to make testing easier? If that becomes a concern we can still revert the change at the `Iterator` to do a prefix check (https://github.com/cockroachdb/pebble/blob/master/iterator.go#L532). That way the heap in the `mergingIter` remains small if it is used and hopefully the second check at the `Iterator` adds little overhead. My only concern with that option is that future correctness problems at the `mergingIter` will be covered up by the overlapping prefix check from the `Iterator`.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3299)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3299#-Nqil6Ix4hXv6gl-sX_b):*

That makes sense, are there any cases where we would explicitly avoid using a `mergingIter` as the `pointIter` or is it purely to make testing easier? If that becomes a concern we can still revert the change at the `Iterator` to do a prefix check (https://github.com/cockroachdb/pebble/blob/master/iterator.go#L532). That way the heap in the `mergingIter` remains small if it is used and hopefully the second check at the `Iterator` adds little overhead. My only concern with that option is that future correctness problems at the `mergingIter` will be covered up by the overlapping prefix check from the `Iterator`.
</blockquote></details>

I don't think we should keep the second check in `Iterator`.
This question was a drive-by, and more for @jbowens. I am forgetting whether there are existing production cases (I think not) or tests (I vaguely remember `fakeIter` but maybe we wrap it in a `mergingIter` and I am wrong) that don't use `mergingIter`. I am fine with enforcing that requirement, with some code that enforces it.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3299)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @CheranMahalingam and @sumeerbhola)

___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3299#-Nqil6Ix4hXv6gl-sX_b):*

I don't think we should keep the second check in `Iterator`.
This question was a drive-by, and more for @jbowens. I am forgetting whether there are existing production cases (I think not) or tests (I vaguely remember `fakeIter` but maybe we wrap it in a `mergingIter` and I am wrong) that don't use `mergingIter`. I am fine with enforcing that requirement, with some code that enforces it.
</blockquote></details>

Good catch. With range keys, the iter will be an interleaving iter, but I _think_ we should be able to make `Iterator.pointIter` a `*mergingIter`. @CheranMahalingam — do you want to try it? If it proves difficult, we can back it out and leave the check for the interim with a TODO and file an issue.

CheranMahalingam (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3299)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @CheranMahalingam, @jbowens, and @sumeerbhola)

___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3299#-Nqil6Ix4hXv6gl-sX_b):*

Good catch. With range keys, the iter will be an interleaving iter, but I _think_ we should be able to make `Iterator.pointIter` a `*mergingIter`. @CheranMahalingam — do you want to try it? If it proves difficult, we can back it out and leave the check for the interim with a TODO and file an issue.
</blockquote></details>

Just for clarification, the concern is that our implementation relies on all `Iterator`s having a `mergingIter` either as an `iter` when searching for point keys or as a `pointIter` when using the `InterleavingIterator`. If that is the case why do we currently include both the `pointIter` and `mergingIter` as separate fields in the `Iterator` and does this imply that they could be combined (https://github.com/cockroachdb/pebble/blob/master/iterator.go#L258)? Also, for builds with invariants enabled we occasionally wrap the `mergingIter` with additional iterators (i.e. invalidating iterator), so is there a clean way to make this change without removing them?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3299)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @CheranMahalingam and @sumeerbhola)

___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3299#-Nqil6Ix4hXv6gl-sX_b):*
> If that is the case why do we currently include both the `pointIter` and `mergingIter` as separate fields in the `Iterator` and does this imply that they could be combined ([https://github.com/cockroachdb/pebble/blob/master/iterator.go#L258](https://github.com/cockroachdb/pebble/blob/master/iterator.go#L258))?

Yeah, it seems like they could be combined, except for cases like that invalidating iterator you just pointed out.... I'm not sure what to do about that. One thought is that we could define an interface like:

```
type toplevelIterator interface {
    internalIterator

    // SeekPrefixGEStrict is like SeekPrefixGE but guarantees that it will
    // never return a key outside the iteration prefix.
    SeekPrefixGEStrict(prefix, key []byte, flags SeekGEFlags) (*InternalKey, LazyValue)
}
```

and implement it in the mergingIter and invalidating iterator, and update our `pointIter` field to be of type `toplevelIterator`.

CheranMahalingam (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3299)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @CheranMahalingam, @jbowens, and @sumeerbhola)

___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3299#-Nqil6Ix4hXv6gl-sX_b):*

> If that is the case why do we currently include both the `pointIter` and `mergingIter` as separate fields in the `Iterator` and does this imply that they could be combined ([https://github.com/cockroachdb/pebble/blob/master/iterator.go#L258](https://github.com/cockroachdb/pebble/blob/master/iterator.go#L258))?

Yeah, it seems like they could be combined, except for cases like that invalidating iterator you just pointed out.... I'm not sure what to do about that. One thought is that we could define an interface like:

```
type toplevelIterator interface {
    internalIterator

    // SeekPrefixGEStrict is like SeekPrefixGE but guarantees that it will
    // never return a key outside the iteration prefix.
    SeekPrefixGEStrict(prefix, key []byte, flags SeekGEFlags) (*InternalKey, LazyValue)
}
```

and implement it in the mergingIter and invalidating iterator, and update our `pointIter` field to be of type `toplevelIterator`.
</blockquote></details>

I think this is complicated by a branch in `createExternalPointIter` where a list of `mergingIterLevel`s are used to create a `mergingIter`, however, if there is only one level, the `pointIter` becomes a `mergingIterLevel.iter` which can't be converted to a `topLevelIterator` (https://github.com/cockroachdb/pebble/blob/master/external_iterator.go#L246). Wondering if this branch is important or if removing it and returning a single level `mergingIter` would reduce the burden.

CheranMahalingam (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3299)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @CheranMahalingam, @jbowens, and @sumeerbhola)

___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3299#-Nqil6Ix4hXv6gl-sX_b):*

I think this is complicated by a branch in `createExternalPointIter` where a list of `mergingIterLevel`s are used to create a `mergingIter`, however, if there is only one level, the `pointIter` becomes a `mergingIterLevel.iter` which can't be converted to a `topLevelIterator` (https://github.com/cockroachdb/pebble/blob/master/external_iterator.go#L246). Wondering if this branch is important or if removing it and returning a single level `mergingIter` would reduce the burden.
</blockquote></details>

Resolved.

jbowens (commented):
looks great!

> Reviewed 3 of 5 files at r2, 10 of 11 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3299#-:-NsJWs2l2tPYCQzbB_az:b6lqmcl)* status: 13 of 14 files reviewed, 3 unresolved discussions (waiting on @CheranMahalingam and @sumeerbhola)

___
*[`external_iterator.go` line 245 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3299#-NsJUNutBlUW1HSL8Okp:-NsJUNutBlUW1HSL8Okq:b-6g1uco) ([raw file](https://github.com/cockroachdb/pebble/blob/f1528d7cb863a5da145ba20dbf8429912d1e7809/external_iterator.go#L245)):*
> ```Go
> 		// which sstable iterators do not support (as they release themselves to
> 		// a pool).
> 		it.closePointIterOnce = true
> ```

I think we can now remove the `closePointIterOnce` field from `Iterator`, as it should it be unused.
___
*[`iterator.go` line 536 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3299#-NsJVEBu0dnifeFDr4of:-NsJV3Nd6Sgi_IJlsWfJ:b-ce9vcf) ([raw file](https://github.com/cockroachdb/pebble/blob/f1528d7cb863a5da145ba20dbf8429912d1e7809/iterator.go#L536)):*
> ```Go
> 				return
> 			}
> 		}
> ```

now that we're requiring i.iterKey to already have the appropriate prefix, can we add an `invariants` build assertion that it holds (and document the requirement here?). something like:

```
// The topLevelIterator.StrictSeekPrefixGE contract requires that in prefix mode [i.hasPrefix=t],
// every point key returned by the internal iterator must be have the current iteration prefix.
// In invariants builds, verify this.
if invariants.Enabled && i.hasPrefix {
    // Range keys are not point keys and may have a different prefix. See the
    // explicit handling of this case in switch below.
    if i.iterKey.Kind() != base.InternalKeyKindRangeKeySet {
        if n := i.split(i.iterKey.UserKey); !i.equal(i.prefixOrFullSeekKey, i.iterKey.UserKey[:n]) {
            i.opts.logger.Fatalf("pebble: invariant violation: key %q does not have prefix %q", i.iterKey.UserKey, i.prefixOrFullSeekKey)
        }
    }
}
```

CheranMahalingam (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3299#-:-NsKC4PkA0NTmjv0nIwp:b-814g8k)* status: 12 of 14 files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`external_iterator.go` line 245 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3299#-NsJUNutBlUW1HSL8Okp:-NsK3M0l8x9bamjuCjR6:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/f1528d7cb863a5da145ba20dbf8429912d1e7809/external_iterator.go#L245)):*

I think we can now remove the `closePointIterOnce` field from `Iterator`, as it should it be unused.
</blockquote></details>

Done.
___
*[`iterator.go` line 536 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3299#-NsJVEBu0dnifeFDr4of:-NsK8cEABFrEu2lIeZ2b:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/f1528d7cb863a5da145ba20dbf8429912d1e7809/iterator.go#L536)):*

now that we're requiring i.iterKey to already have the appropriate prefix, can we add an `invariants` build assertion that it holds (and document the requirement here?). something like:

```
// The topLevelIterator.StrictSeekPrefixGE contract requires that in prefix mode [i.hasPrefix=t],
// every point key returned by the internal iterator must be have the current iteration prefix.
// In invariants builds, verify this.
if invariants.Enabled && i.hasPrefix {
    // Range keys are not point keys and may have a different prefix. See the
    // explicit handling of this case in switch below.
    if i.iterKey.Kind() != base.InternalKeyKindRangeKeySet {
        if n := i.split(i.iterKey.UserKey); !i.equal(i.prefixOrFullSeekKey, i.iterKey.UserKey[:n]) {
            i.opts.logger.Fatalf("pebble: invariant violation: key %q does not have prefix %q", i.iterKey.UserKey, i.prefixOrFullSeekKey)
        }
    }
}
```
</blockquote></details>

Done.

jbowens (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3299#-:-NsKdcJYBJT2sMrzgX_M:b-uwkowd)* status: 12 of 14 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

PR #3300, opened 2024-02-13
sstable: add more commentary on synthetic suffix replacement

msbutler:
See individual commits

### DIFF
--- a/sstable/block.go
+++ b/sstable/block.go
@@ -304,7 +304,10 @@ type blockEntry struct {
 // guarantee is used by the range tombstone and range key code, which knows that
 // the respective blocks are always encoded with a restart interval of 1. This
 // per-block key stability guarantee is sufficient for range tombstones and
-// range deletes as they are always encoded in a single block.
+// range deletes as they are always encoded in a single block. Note: this
+// stability guarantee no longer holds for a block iter with synthetic suffix
+// replacement, but this doesn't matter, as the user will not open
+// an iterator with a synthetic suffix on a block with rangekeys (for now).
 //
 // A blockIter also provides a value stability guarantee for range deletions and
 // range keys since there is only a single range deletion and range key block
@@ -372,6 +375,11 @@ type blockIter struct {
 	// point directly to data stored in the block (for a key which has no prefix
 	// compression), to fullKey (for a prefix compressed key), or to a slice of
 	// data stored in cachedBuf (during reverse iteration).
+	//
+	// NB: In general, key contains the same logical content as ikey
+	// (i.e. ikey = decode(key)), but if the iterator contains a synthetic suffix
+	// replacement rule, this will not be the case. Therefore, key should never
+	// be used after ikey is set.
 	key []byte
 	// fullKey is a buffer used for key prefix decompression.
 	fullKey []byte
@@ -383,9 +391,10 @@ type blockIter struct {
 	lazyValue base.LazyValue
 	// ikey contains the decoded InternalKey the iterator is currently pointed
 	// at. Note that the memory backing ikey.UserKey is either data stored
-	// directly in the block, fullKey, or cachedBuf. The key stability guarantee
-	// for blocks built with a restart interval of 1 is achieved by having
-	// ikey.UserKey always point to data stored directly in the block.
+	// directly in the block, fullKey, cachedBuf, or synthSuffixBuf. The key
+	// stability guarantee for blocks built with a restart interval of 1 is
+	// achieved by having ikey.UserKey always point to data stored directly in the
+	// block.
 	ikey InternalKey
 	// cached and cachedBuf are used during reverse iteration. They are needed
 	// because we can't perform prefix decoding in reverse, only in the forward
@@ -411,8 +420,27 @@ type blockIter struct {
 		hasValuePrefix bool
 	}
 	hideObsoletePoints bool
-	syntheticSuffix    SyntheticSuffix
-	synthSuffixBuf     []byte
+
+	// syntheticSuffix, if not nil, will replace the decoded ikey.UserKey suffix
+	// before the key is returned to the user. A sequence of iter operations on a
+	// block with a syntheticSuffix rule should return keys as if those operations
+	// ran on a block with keys that all had the syntheticSuffix. As an example:
+	// any sequence of block iter cmds should return the same keys for the
+	// following two blocks:
+	//
+	// blockA: a@3,b@3,c@3
+	// blockB: a@1,b@2,c@1 with syntheticSuffix=3
+	//
+	// To ensure this, Suffix replacement will not change the ordering of keys in
+	// the block because the iter assumes the following about the block:
+	//
+	// (1) no two keys in the block share the same prefix.
+	// (2) pebble.Compare(keyPrefix{replacementSuffix},keyPrefix{originalSuffix}) < 0
+	//
+	// In addition, we also assume that any block with rangekeys will not contain
+	// a synthetic suffix.
+	syntheticSuffix SyntheticSuffix
+	synthSuffixBuf  []byte
 }
 
 // blockIter implements the base.InternalIterator interface.
@@ -442,9 +470,7 @@ func (i *blockIter) init(
 		return base.CorruptionErrorf("pebble/table: invalid table (block has no restart points)")
 	}
 	i.syntheticSuffix = syntheticSuffix
-	if i.syntheticSuffix != nil {
-		i.synthSuffixBuf = []byte{}
-	}
+	i.synthSuffixBuf = i.synthSuffixBuf[:0]
 	i.split = split
 	i.cmp = cmp
 	i.restarts = int32(len(block)) - 4*(1+numRestarts)
@@ -994,7 +1020,13 @@ func (i *blockIter) SeekLT(key []byte, flags base.SeekLTFlags) (*InternalKey, ba
 			// The binary search was conducted on keys without suffix replacement,
 			// implying the first key in the block may be less than the search key. To
 			// double check, get the first key in the block with suffix replacement
-			// and compare to the search key.
+			// and compare to the search key. Consider the following example: suppose
+			// the user searches with a@3, the first key in the block is a@2 and the
+			// block contains a suffix replacement rule of 4. Since a@3 sorts before
+			// a@2, the binary search would return index==0. Without conducting the
+			// suffix replacement, the SeekLT would incorrectly return nil. With
+			// suffix replacement though, a@4 should be returned as a@4 sorts before
+			// a@3.
 			ikey, lazyVal := i.First()
 			if i.cmp(ikey.UserKey, key) < 0 {
 				return ikey, lazyVal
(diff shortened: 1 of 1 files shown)


sstable/block.go
```diff
@@ -1008,7 +1008,27 @@ func (i *blockIter) SeekLT(key []byte, flags base.SeekLTFlags) (*InternalKey, ba
 		i.offset = decodeRestart(i.data[i.restarts+4*(index-1):])
 		if index < i.numRestarts {
 			targetOffset = decodeRestart(i.data[i.restarts+4*(index):])
+
```
msbutler, line 1011: If you like this approach better, I'll add a big comment explaining what's going on here.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3300)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @msbutler and @sumeerbhola)

___
*[`sstable/suffix_rewriter.go` line 204 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3300#-NqdQHppC_NR3PvwDYk9:-NqdQHppC_NR3PvwDYkA:bvdkyvc) ([raw file](https://github.com/cockroachdb/pebble/blob/6b0359662b26ddb4c8583832081f5d809ad5bb01/sstable/suffix_rewriter.go#L204)):*
> ```Go
> 			return err
> 		}
> 		if err := iter.init(r.Compare, r.Split, inputBlock, r.Properties.GlobalSeqNum, false /* hideObsoletePoints */, nil /* syntheticSuffix */); err != nil {
> ```

Sorry to be a pain but could you leave these call sites as they are? I am moving them to an `IterTransforms` struct which contains `HideObsoletePoints` and `SyntheticSuffix` (and the "global" sequence number), and this would just create merge conflicts. It will be addressed very soon.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3300)* status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @10, @20, @msbutler, and @sumeerbhola)

___
*[`sstable/block.go` line 305 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3300#-Nqh1nPAD4U5M-E3Am1X:-NqhLib_5A-siRwQGXWL:b-m349dj) ([raw file](https://github.com/cockroachdb/pebble/blob/52e000d13771e34ad535e843a7e644b5b68d5b10/sstable/block.go#L305)):*
> Are we only relying on the key stability guarantee when processing rangedels and rangekeys?

Yeah, I believe that's true

> What about rangekeys? If someone wrote a RANGEKEYSET `[b, g)@10`, and now we want to replace suffix `@10` with suffix `@20`, how is that accomplished?

Also, what happens when a range key fragment has a suffix in either the start or end key? That can happen in ordinary non-ingested sstables due to fragmentation, but maybe it cannot happen when we produce these sstables?

Or maybe are there no range keys in these sstables at all because they're produced without version history?

RaduBerinde:
*[`sstable/block.go` line 305 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3300#-Nqh1nPAD4U5M-E3Am1X:-NqhVoVlC31cjJqTpoji:b-6jcyed) ([raw file](https://github.com/cockroachdb/pebble/blob/52e000d13771e34ad535e843a7e644b5b68d5b10/sstable/block.go#L305)):*

> Are we only relying on the key stability guarantee when processing rangedels and rangekeys?

Yeah, I believe that's true

> What about rangekeys? If someone wrote a RANGEKEYSET `[b, g)@10`, and now we want to replace suffix `@10` with suffix `@20`, how is that accomplished?

Also, what happens when a range key fragment has a suffix in either the start or end key? That can happen in ordinary non-ingested sstables due to fragmentation, but maybe it cannot happen when we produce these sstables?

Or maybe are there no range keys in these sstables at all because they're produced without version history?
</blockquote></details>

Yeah we don't allow rangekeys in external ingest for now (we won't need to for the first OR version).

RaduBerinde:
*[`sstable/suffix_rewriter.go` line 204 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3300#-NqdQHppC_NR3PvwDYk9:-Nqmd3DQ3aRQCNXChmMt:b-lxt62f) ([raw file](https://github.com/cockroachdb/pebble/blob/6b0359662b26ddb4c8583832081f5d809ad5bb01/sstable/suffix_rewriter.go#L204)):*

Sorry to be a pain but could you leave these call sites as they are? I am moving them to an `IterTransforms` struct which contains `HideObsoletePoints` and `SyntheticSuffix` (and the "global" sequence number), and this would just create merge conflicts. It will be addressed very soon.
</blockquote></details>

https://github.com/cockroachdb/pebble/pull/3310

RaduBerinde:
*[`sstable/suffix_rewriter.go` line 204 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3300#-NqdQHppC_NR3PvwDYk9:-NqmfhQF5yaJVry5Ali1:b4qxvct) ([raw file](https://github.com/cockroachdb/pebble/blob/6b0359662b26ddb4c8583832081f5d809ad5bb01/sstable/suffix_rewriter.go#L204)):*

https://github.com/cockroachdb/pebble/pull/3310
</blockquote></details>

The PR above also adds a check that we never try to iterate through range dels or range keys when we have a synthetic suffix.

msbutler (commented):
Removed last commit.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3300#-:-NqnVbT53INgwG886Pks:b47h2nm)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @10, @20, @msbutler, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/block.go` line 305 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3300#-Nqh1nPAD4U5M-E3Am1X:-NqnYD-2Euo33mT67exy:b-164t4w) ([raw file](https://github.com/cockroachdb/pebble/blob/52e000d13771e34ad535e843a7e644b5b68d5b10/sstable/block.go#L305)):*

Yeah we don't allow rangekeys in external ingest for now (we won't need to for the first OR version).
</blockquote></details>

Added more commentary on this. 
___
*[`sstable/block.go` line 1011 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3300#gh-1489758567:-NqnYK1zDWAIZHYyQmJb:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/99105b73ef641db7988111667310fd40dc806a0c/sstable/block.go#L1011)):*

Yes, I have a preference for this since it sets up everything correctly for later iteration.
</blockquote></details>

Done.
___
*[`sstable/block.go` line 1016 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3300#-NqhABC51TwMS1AghUBo:-NqnYL3W8-8mTm2L_4UW:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/99105b73ef641db7988111667310fd40dc806a0c/sstable/block.go#L1016)):*

we don't need to set `i.nextOffset`. `i.readEntry()` will set that.
</blockquote></details>

Done.
___
*[`sstable/block.go` line 1025 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3300#-NqhAGCH3c_qBGDH5lSY:-NqnYLtb-41wY3IKSR8q:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/99105b73ef641db7988111667310fd40dc806a0c/sstable/block.go#L1025)):*

index+1 may be out of bounds.
</blockquote></details>

Done.
___
*[`sstable/block.go` line 1030 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3300#-NqhASzr0JJzt6nOxycG:-NqnVSX4EHa8tIbX-Dpr:bmobqgs) ([raw file](https://github.com/cockroachdb/pebble/blob/99105b73ef641db7988111667310fd40dc806a0c/sstable/block.go#L1030)):*

I think these blocks can be rearranged to be clearer (some of this predates your changes, but we need to simplify to compensate for the added complexity)

Let's lift the index==0 case before, so we can delay definition of the `targetOffset` and can clarify what it really means. And then we can unindent the main case. Something like:

```
if index == 0 {  
  if i.syntheticSuffix != nil {  
   // The binary search was conducted on keys without suffix replacement,  
   // implying the first key in the block may be less than the search key. To   // double check, get the first key in the block with suffix replacement   // and compare to the search key. Consider the following example: suppose   // the user searches with a@3, the first key in the block is a@2 and the   // block contains a suffix replacement rule of 4. Since a@3 sorts before   // a@2, the binary search would return index==0. Without conducting the   // suffix replacement, the SeekLT would incorrectly return nil. With   // suffix replacement though, a@4 should be returned as a@4 sorts before   // a@3.   ikey, lazyVal := i.First()  
   if i.cmp(ikey.UserKey, key) < 0 {  
    return ikey, lazyVal  
   }  
  }  
  // If index == 0 then all keys in this block are larger than the key  
  // sought.  i.offset = -1  
  i.nextOffset = 0  
  return nil, base.LazyValue{}  
}  
  
// INVARIANT: index > 0  
  
// Ignoring suffix replacement, we will search between the restart at index-1  
// and the restart at index, for the first key >= key, and then on finding it,  
// return i.Prev(). We need to know when we have hit the offset for index, since  
// then we can stop searching. targetOffset encodes that offset for index.  
targetOffset := i.restarts  
  
i.offset = decodeRestart(i.data\[i.restarts+4\*(index-1):\])  
if index < i.numRestarts {  
  targetOffset = decodeRestart(i.data\[i.restarts+4\*(index):\])  
  if i.syntheticSuffix != nil {  
   // It is possible that we should be searching between the restart at index  
   // and that at index+1. TODO: add longer comment with example  
   naiveOffset := i.offset  
   // Shift up to the original binary search result and decode the key.  
   i.offset = targetOffset  
   i.readEntry()  
   i.decodeInternalKey(i.key)  
   i.maybeReplaceSuffix(false /\* allowInPlace \*/)  
  
   // If the binary search point is actually less than the search key, post  
   // replacement, bump the target offset.   if i.cmp(i.ikey.UserKey, key) < 0 {  
    i.offset = targetOffset  
    if index+1 < i.numRestarts {  
     targetOffset = decodeRestart(i.data\[i.restarts+4\*(index+1):\])  
    } else {  
     targetOffset = i.restarts  
    }  
   } else {  
    i.offset = naiveOffset  
   }  
  }  
}
```
</blockquote></details>

Done, with added commentary for suffix replacement.
___
*[`sstable/suffix_rewriter.go` line 204 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3300#-NqdQHppC_NR3PvwDYk9:-NqnVXVm65mcLNnIvEZ2:bc4am8q) ([raw file](https://github.com/cockroachdb/pebble/blob/6b0359662b26ddb4c8583832081f5d809ad5bb01/sstable/suffix_rewriter.go#L204)):*

The PR above also adds a check that we never try to iterate through range dels or range keys when we have a synthetic suffix.
</blockquote></details>

Removed the third commit from this pr. Will deal with later.

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3300#-:-NqooYSs09MXb31mdqY2:bg7igkv)* status: 1 of 7 files reviewed, 5 unresolved discussions (waiting on @10, @20, @msbutler, and @RaduBerinde)

RaduBerinde:
*[`sstable/block.go` line 438 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3300#-Nr1inlP2rX3ULhUI3sm:-Nr1inlQ-aLM0ye3CpOg:b2jnkxr) ([raw file](https://github.com/cockroachdb/pebble/blob/c81fbe1f03d31a396f4376a506c6b09b1a9b0622/sstable/block.go#L438)):*
> ```Go
> 	//
> 	// (1) no two keys in the block share the same prefix.
> 	// (2) pebble.Compare(keyPrefix{replacementSuffix},keyPrefix{originalSuffix}) < 0
> ```

we also allow `originalSuffix` to be empty, correct?

msbutler:
*[`sstable/block.go` line 438 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3300#-Nr1inlP2rX3ULhUI3sm:-Nr1pMEO4JcwcYkqmaiV:bt6xfj9) ([raw file](https://github.com/cockroachdb/pebble/blob/c81fbe1f03d31a396f4376a506c6b09b1a9b0622/sstable/block.go#L438)):*

we also allow `originalSuffix` to be empty, correct?
</blockquote></details>

Ah, i should have clarified given our conversation.  Yes, that is true.

And to clarify, keys with an empty suffix sort before keys with a suffix
`a@,a@5,a@4`

In other words, conducting suffix replacement on a key without a suffix would make it less than the original value, which contradicts this invariant.

This makes me wonder if all my added off-by-one logic breaks down when we conduct suffix replacement on keys without a suffix.

msbutler:
*[`sstable/block.go` line 438 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3300#-Nr1inlP2rX3ULhUI3sm:-Nr1u5vK-zFRylniL-T_:b-g75072) ([raw file](https://github.com/cockroachdb/pebble/blob/c81fbe1f03d31a396f4376a506c6b09b1a9b0622/sstable/block.go#L438)):*

Ah, i should have clarified given our conversation.  Yes, that is true.

And to clarify, keys with an empty suffix sort before keys with a suffix
`a@,a@5,a@4`

In other words, conducting suffix replacement on a key without a suffix would make it less than the original value, which contradicts this invariant.

This makes me wonder if all my added off-by-one logic breaks down when we conduct suffix replacement on keys without a suffix.
</blockquote></details>

note: the only keys without a suffix are index keys

RaduBerinde:
*[`sstable/block.go` line 438 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3300#-Nr1inlP2rX3ULhUI3sm:-Nr1xkNn9o8ky6I55K0K:b-17oyxu) ([raw file](https://github.com/cockroachdb/pebble/blob/c81fbe1f03d31a396f4376a506c6b09b1a9b0622/sstable/block.go#L438)):*

note: the only keys without a suffix are index keys
</blockquote></details>

In the meta test we do have keys without prefix and it seems to work fine.

msbutler:
*[`sstable/block.go` line 438 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3300#-Nr1inlP2rX3ULhUI3sm:-Nr5zLv1DJ-An7_Xmd6u:b-5tjvtk) ([raw file](https://github.com/cockroachdb/pebble/blob/c81fbe1f03d31a396f4376a506c6b09b1a9b0622/sstable/block.go#L438)):*

In the meta test we do have keys without prefix and it seems to work fine.
</blockquote></details>

Alrighty, i feel a bit better about this after adding a few unit test cases here
https://github.com/cockroachdb/pebble/pull/3320

I think this all works is because we only need to trigger off by one handling when (integer cmp)  origSuffix < > searchkeySuffix <> replacementSuffix (i.e. in the searchkeySuffix is in between the other two suffixes). This condition will never trigger because the originalSuffix is the max suffix, the empty suffix.

PR #3301, opened 2024-02-14
add external file iteration to scanInternal

stevendanna:
Like the sharedFileVisitor, the new externalFileVisitor allows the caller to process external files directly.

In a follow up PR, we will teach ingest how to process these similar to how it processes SharedSSTs

In this PR, the externalFileVisitor only applies to L6 and L6 must be completely external or shared files.


stevendanna:
Closing this as it is probably easier to review with some calling code in place as well. WIP here https://github.com/cockroachdb/pebble/pull/3348

PR #3306, opened 2024-02-15
manifest: fix bug when parsing the suffix rewrite tag

RaduBerinde:
We were missing a `continue` which leads to an error when the suffix
rewrite tag is used (meaning we can't open a database).

This commit restructures the code into a single switch to make this
mistake less likely.

### DIFF
--- a/internal/manifest/version_edit_test.go
+++ b/internal/manifest/version_edit_test.go
@@ -50,6 +50,11 @@ func TestVERoundTripAndAccumulate(t *testing.T) {
 		CreationTime:   809060,
 		SmallestSeqNum: 9,
 		LargestSeqNum:  11,
+		PrefixReplacement: &PrefixReplacement{
+			ContentPrefix:   []byte("before"),
+			SyntheticPrefix: []byte("after"),
+		},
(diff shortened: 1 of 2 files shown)


internal/manifest/version_edit_test.go
```diff
@@ -45,11 +45,12 @@ func checkRoundTrip(e0 VersionEdit) error {
...
-		Size:           8090,
-		CreationTime:   809060,
-		SmallestSeqNum: 9,
-		LargestSeqNum:  11,
+		FileNum:         810,
+		Size:            8090,
+		CreationTime:    809060,
+		SmallestSeqNum:  9,
+		LargestSeqNum:   11,
+		SyntheticSuffix: []byte("foo"),
```
msbutler, line 53: while you're here, could you add a synthetic prefix?

msbutler (approved):
Nice find!

internal/manifest/version_edit_test.go
RaduBerinde: Good idea, done. TFTR!

PR #3297, opened 2024-02-13
sstable: move PrefixReplacement to sstable

RaduBerinde:
#### sstable: move PrefixReplacement to sstable

We change `MakeVirtualReader` to not use the file metadata types,
resolving the last dependency on `manifest`; and we move the
`PrefixReplacement` struct to `sstable`.

PR #3310, opened 2024-02-16
sstable: remove global seq num property

RaduBerinde:
This PR is only for the top commit (the rest is #3297).

#### sstable: remove global seq num property

This change removes the "global seq num" sstable property and instead
passes it to iterators, as a `SyntheticSeqNum`. This cleans up a lot
of the code that manipulated the global seq num in the reader and
allows the same external file backing to be shared.

As part of this change, we create a new struct `IterTransforms` and
move the synthetic suffix as well as `hideObsoletePoints` to this
struct. This cleans up the call sites, and allows easily adding more
transforms in the future (e.g prefix replacement).

We add a method on `FileMetadata` that generates the proper
`IterTransforms` for that table.


RaduBerinde:
We will be able to move PrefixReplacement from `vState` to `IterTransforms` as well, I just didn't want to tackle it in the same PR.

jbowens (approved):
Nice! :lgtm:

Sorry for the delay

> Reviewed 7 of 7 files at r1, 57 of 57 files at r4, 9 of 10 files at r5, 17 of 55 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3310#-:-NrDXRoN7Ve4VyMQfpvH:bfae8gd)* status: 27 of 65 files reviewed, all discussions resolved (waiting on @itsbilal and @msbutler)

PR #3311, opened 2024-02-16
db: add ViaBackingFileDownload option to DownloadSpan

itsbilal:
This change adds an option to `Download`, `ViaBackingFileDownload`, that just does a copy compaction of the backing file instead of doing a rewrite compaction that iterates on keys one at a time. This allows for greater concurrency as less CPU work per byte is necessary to move these files to disk, and gives us more tools to inform the appropriate

Subsumes #3239.


itsbilal:
For now I've just made fixes to #3239 and addressed @RaduBerinde's points as it will help unblock our testing. If we do decide to go in the copy direction, the ideal way to do this would be as a provider operation that doesn't require any of the compaction-picking machinery and db.mu lock/unlocks.

RaduBerinde (commented):
LGTM after squashing commits and some minor comments

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3311#-:-NquXvAg0rhIQfrPNL4p:b-27irvu)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`compaction.go` line 2831 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3311#-NquXCwG5R9QqfXkCJCZ:-NquXCwG5R9QqfXkCJC_:bqb2m7u) ([raw file](https://github.com/cockroachdb/pebble/blob/9710bbd732f6abac089a48c7b9533222d5555a5a/compaction.go#L2831)):*
> ```Go
> 			buf := make([]byte, 128<<10)
> 			var pos int64
> 			for {
> ```

[nit] This should be `for pos := int64(0); pos < size; {`
___
*[`internal/manifest/version.go` line 434 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3311#-NquXQ6i5J6_dJYfUSdL:-NquXQ6i5J6_dJYfUSdM:b-lf70ks) ([raw file](https://github.com/cockroachdb/pebble/blob/9710bbd732f6abac089a48c7b9533222d5555a5a/internal/manifest/version.go#L434)):*
> ```Go
> 
> // InitProviderBacking creates a new FileBacking for a file backed by
> // an objstorage.Provider. It's the caller's responsibility to set the size
> ```

It looks like in all existing callsites we set the size shortly after; let's make it an argument to this method.
___
*[`testdata/ingest_external` line 330 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3311#-NquXmJaClPggSsVM4Yb:-NquXmJaClPggSsVM4Yc:b-9cunrf) ([raw file](https://github.com/cockroachdb/pebble/blob/9710bbd732f6abac089a48c7b9533222d5555a5a/testdata/ingest_external#L330)):*
> ```
> ----
> 6:
>   000004:[gc#10,DELSIZED-gf#inf,RANGEDEL]
> ```

Maybe we should improve `lsm` to show the backing num when the table is virtual?

dt (commented):
Squashed.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3311#-:-Nr7cpsdAJTSA5lIZyCW:b-un4k39)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`compaction.go` line 2831 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3311#-NquXCwG5R9QqfXkCJCZ:-Nr7coer0R7HV4gxfyQQ:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9710bbd732f6abac089a48c7b9533222d5555a5a/compaction.go#L2831)):*

[nit] This should be `for pos := int64(0); pos < size; {`
</blockquote></details>

Done.
___
*[`internal/manifest/version.go` line 434 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3311#-NquXQ6i5J6_dJYfUSdL:-Nr7cnLq92FiAEqJYq2F:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9710bbd732f6abac089a48c7b9533222d5555a5a/internal/manifest/version.go#L434)):*

It looks like in all existing callsites we set the size shortly after; let's make it an argument to this method.
</blockquote></details>

Done.
___
*[`testdata/ingest_external` line 330 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3311#-NquXmJaClPggSsVM4Yb:-Nr7ctpp1GZ7q-V1dd5z:bqmjwcz) ([raw file](https://github.com/cockroachdb/pebble/blob/9710bbd732f6abac089a48c7b9533222d5555a5a/testdata/ingest_external#L330)):*

Maybe we should improve `lsm` to show the backing num when the table is virtual?
</blockquote></details>

Added.

itsbilal:
David's changes LGTM, does someone mind LGTMing this cuz I can't: 
 
Screenshot 2024-02-20 at 6 56 29 PM

dt (approved):

RaduBerinde:
Looks great, thanks!

PR #3315, opened 2024-02-19
tool: adapt find command to explicitly track paths, filenames

jbowens:
Previously, the find tool used a map[base.DiskFileNum]string to record file paths across all file types. With #3230, WALs will need separate file path tracking because a single file number may actually consist of several physical files at several unique file paths. To avoid any confusion after this change is made, paths are now propagated alongside disk file numbers.

### DIFF
--- a/tool/find.go
+++ b/tool/find.go
@@ -9,6 +9,7 @@ import (
 	"cmp"
 	"fmt"
 	"io"
+	"path/filepath"
 	"slices"
 	"sort"
 
@@ -24,9 +25,10 @@ import (
 )
 
 type findRef struct {
-	key     base.InternalKey
-	value   []byte
-	fileNum base.FileNum
+	key      base.InternalKey
+	value    []byte
+	fileNum  base.FileNum
+	filename string
 }
 
 // findT implements the find tool.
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3315#-:-Nr2mmhqCYLPZEyCskh0:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @aadityasondhi)

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3315#-:-Nr5sgeV35NJEJgzzk3n:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @jbowens)

___
*[`tool/find.go` line 80 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3315#-Nr5rhCy5kq6IJaUtmoa:-Nr5rhCy5kq6IJaUtmob:b-vx3k2z) ([raw file](https://github.com/cockroachdb/pebble/blob/ec4317d4101a40308167bc246dbf222d10b3c499/tool/find.go#L80)):*
> ```Go
> type fileLoc struct {
> 	base.DiskFileNum
> 	path string
> ```

won't we need the FS too? Is that for a future PR when we add a second FS to `pebble.Options`, and this tool lists logs in both?

jbowens (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3315#-:-Nr671PKEDoz5tB53_AS:bz5mair)* status: all files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @sumeerbhola)

___
*[`tool/find.go` line 80 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3315#-Nr5rhCy5kq6IJaUtmoa:-Nr66bGS-bnKclZclQe6:bhplaqw) ([raw file](https://github.com/cockroachdb/pebble/blob/ec4317d4101a40308167bc246dbf222d10b3c499/tool/find.go#L80)):*

won't we need the FS too? Is that for a future PR when we add a second FS to `pebble.Options`, and this tool lists logs in both?
</blockquote></details>

yeah, I'm planning on either adding it in that PR or further tweaking the `wal.Logs` type so that the `find` tool doesn't use `fileLoc` for WALs at all. I'm not sure what'll shake out yet, but I figured regardless it would be confusing to maintain the `base.DiskFileNum` -> path map if it didn't hold for all disk file nums.

PR #3318, opened 2024-02-20
metamorphic: special handling of assertion errors

RaduBerinde:
Check for assertion errors when recording history and display their
stacktrace and panic. This helps with debugging.

We also fix some cases where we return an assertion error when we are
checking an API input; assertion errors should indicate a bug inside
Pebble.


itsbilal (approved):
LGTM

RaduBerinde:
TFTR!

PR #3320, opened 2024-02-20
test  and add commentary on synthetic suffix replacement on keys without suffix

msbutler:
This patch adds a few test cases to ensure that when the block iterator adds a synthetic suffix to a key without a suffix, the block iterator calls behave as expected.


RaduBerinde (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3320)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @msbutler)

PR #3321, opened 2024-02-20
add method to retrieve count of external files

stevendanna:
This allows callers to operate differently if the database contains external files that might be slower to read than expected.


dt:
I think the second return value of `EstimateDiskUsageByBackingType` could probably serve the same purpose as this unless that method is too expensive?

stevendanna:
I'll start with that and only move forward with this function if I start seeing it actually being a problem.

PR #3326, opened 2024-02-21
internal/keyspan: remove datadriven runtime dep

dt:
In #3293 some helpers were exported for reuse in other tests and the file suffix of _test was removed, however this meant that the keyspan package started to link in datadriven even in non-test binaries. This is not desired, as datadriven has init-registered flags such as rewrite and should only be linked in test binaries.

Fortunately the helpers only needed a string field so we don't need to use 'datadriven' to pass it that.


stevendanna (approved):
Seems pretty reasonable to me.

RaduBerinde:
Nice catch, thank you!!

PR #3330, opened 2024-02-22
sstable: fix compaction iterator bug with prefix replacement

RaduBerinde:
#### manifest: rename PrefixReplacement methods

The `ReplaceArg / ReplaceResult` methods are not too descriptive, I
always have to look at the code to remember which does what. Renaming
to `Apply` and `Invert`.

#### sstable: fix compaction iterator bug with prefix replacement

The sstable compaction iterators stop when the key exceeds
`vState.upper`; however this is incorrect when we replace prefixes
(the key has the content prefix and `vState.upper` has the synthetic
prefix).

Switch to using `i.upper` which is in the "content prefix space" and
was already calculated at initialization.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -510,6 +510,11 @@ func runBuildRemoteCmd(td *datadriven.TestData, d *DB, storage remote.Storage) e
 	}
 
 	writeOpts := d.opts.MakeWriterOptions(0 /* level */, tableFormat)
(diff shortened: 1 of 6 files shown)


dt (approved):

RaduBerinde:
TFTRs!

PR #3338, opened 2024-02-22
wal: delete recovered logs in Open

jbowens:
Adjust the Open logic to delete logs discovered on Open without ever propagating them to the WAL manager. This has the advantage that each WAL manager queue only needs to maintain information about WALs of the type that it creates. For example, if WAL failover was previously enabled but is no longer, any logical WALs that are split across multiple physical files will be deleted by Open and never need to enter the standalone manager's log queue.

### DIFF
--- a/open.go
+++ b/open.go
@@ -339,7 +339,9 @@ func Open(dirname string, opts *Options) (db *DB, err error) {
 	if err != nil {
 		return nil, err
 	}
-	err = walManager.Init(walOpts, wals)
+	// We pass minRecycleLogNum=wals.MaxNum()+1 to avoid recycling any of the
+	// logs that exist at start.
+	err = walManager.Init(walOpts, wals.MaxNum()+1)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3338#-:-NrItSF72MiMuWs0U2QS:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`wal/failover_manager.go` line 457 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3338#-NrItGCp8LTBcDppwkqB:-NrItGCp8LTBcDppwkqC:b-plbi49) ([raw file](https://github.com/cockroachdb/pebble/blob/a82a1b14f24a4137607f83983147bede450699ed/wal/failover_manager.go#L457)):*
> ```Go
> 	}
> 	wm.recycler.Init(o.MaxNumRecyclableLogs)
> 	if wm.recycler.MinRecycleLogNum() <= minRecycleLogNum {
> ```

should this be < 
___
*[`wal/standalone_manager.go` line 57 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3338#-NrIsqXfEQ-OJykZzego:-NrIsqXfEQ-OJykZzegp:b-plbi49) ([raw file](https://github.com/cockroachdb/pebble/blob/a82a1b14f24a4137607f83983147bede450699ed/wal/standalone_manager.go#L57)):*
> ```Go
> 	}
> 	m.recycler.Init(o.MaxNumRecyclableLogs)
> 	if m.recycler.MinRecycleLogNum() <= minRecycleLogNum {
> ```

should this be <

jbowens (commented):
This approach was a bit fraught. It's difficult to preserve determinism of file deletions if we introduce the EnqueueJob after replaying the WALs. I'm closing this in favor of #3341.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3338#-:-NrLjURk4icVUn9Quq29:b-ggry3t)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens)

PR #3342, opened 2024-02-23
db: don't wait on compaction cond lock if db is closed

itsbilal:
Previously, if the db was closed and a Download() call was waiting on a download to finish, we could see an indefinite wait because `maybeScheduleCompactionPicker` would not schedule any compactions, and the cond lock would never be broadcasted to. This change addresses this by checking for db closure before waiting on the compaction cond lock in downloadSpan().

### DIFF
--- a/db.go
+++ b/db.go
@@ -1996,6 +1996,12 @@ func (d *DB) downloadSpan(ctx context.Context, span DownloadSpan) error {
 					return true
 				}
 				d.maybeScheduleCompaction()
+				if d.mu.compact.compactingCount == 0 {
+					// No compactions were scheduled above. Waiting on the cond lock below
+					// could possibly lead to a forever-wait. Return true if the db is
+					// closed so we exit out of this method.
+					return d.closed.Load() != nil
+				}
 				d.mu.compact.cond.Wait()
 				return false
 			}()
@@ -2015,7 +2021,7 @@ func (d *DB) downloadSpan(ctx context.Context, span DownloadSpan) error {
 // downloaded with high priority.
 //
 // The method returns once no external sstasbles overlap the given spans, the
-// context is canceled, or an error is hit.
+// context is canceled, the db is closed, or an error is hit.
 //
 // TODO(radu): consider passing a priority/impact knob to express how important
 // the download is (versus live traffic performance, LSM health).


RaduBerinde (commented):
Why is it legal to call `db.Close()` while `db.Download()` is also running?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3342#-:-NrMlkaW8XTi4lK0YhA6:bsi4m5w)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @msbutler)

___
*[`db.go` line 1999 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3342#-NrMkgCfBxCL6YkG427f:-NrMkgCfBxCL6YkG427g:b-j0jne4) ([raw file](https://github.com/cockroachdb/pebble/blob/85f6cea7e345b60f52b09972481170a5c6f47678/db.go#L1999)):*
> ```Go
> 				}
> 				d.maybeScheduleCompaction()
> 				if d.closed.Load() != nil {
> ```

This works because the only way to not have a compaction running here is if `maybeScheduleCompaction` noticed that `closed.Load() != nil`, is this right? Seems very fragile, couldn't we just check the `compactingCount` instead?

itsbilal (commented):
@RaduBerinde It happened in a crdb test and I imagine it could happen as another means of cancelling downloads. Probably best to support it.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3342#-:-NrfyKRA1HHED_yiFFEn:bf8amv2)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @msbutler and @RaduBerinde)

___
*[`db.go` line 1999 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3342#-NrMkgCfBxCL6YkG427f:-NrfyTOTACNjoT3KqU2V:b-1w6lnd) ([raw file](https://github.com/cockroachdb/pebble/blob/85f6cea7e345b60f52b09972481170a5c6f47678/db.go#L1999)):*

This works because the only way to not have a compaction running here is if `maybeScheduleCompaction` noticed that `closed.Load() != nil`, is this right? Seems very fragile, couldn't we just check the `compactingCount` instead?
</blockquote></details>

Good point. Made the change; ptal!

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3342#-:-NrgJqeZ3skW0CwLJUcz:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @msbutler)

itsbilal:
TFTR!

PR #3344, opened 2024-02-24
sstable: fixes to prefix replacing iterator

RaduBerinde:
We add more logic inside the prefix replacing iterators:
 - pass through `SeekGE` calls even when not strictly necessary, to
   enable optimizations indicated by the flags.
 - keep track of whether we are positioned before/after the synthetic
   prefix range, to enable expected behaviors like `SeekGE` after the
   range followed by `Prev` (or `SeekLT` to before the range followed
   by `Next`)

### DIFF
--- a/sstable/prefix_replacing_iterator.go
+++ b/sstable/prefix_replacing_iterator.go
@@ -23,27 +23,38 @@ type prefixReplacingIterator struct {
 	contentPrefix   []byte
 	syntheticPrefix []byte
 
-	// keyInRange is a valid key in the logical range that has the syntheticPrefix.
-	// When an argument to a seek function does not have the syntheticPrefix,
-	// keyInRange is used to determine if the argument key is before or after the
-	// range of keys produced by the iterator.
-	keyInRange []byte
+	// lower is a valid key that has syntheticPrefix and is a lower bound for all
+	// keys produced by i. It is used to determine if the argument key is before
+	// or after the range of keys produced by the iterator.
+	lower []byte
 
 	// arg and arg2 are buffers that are used to avoid allocations when rewriting
 	// keys that are provided as arguments. They always start with contentPrefix.
 	arg, arg2 []byte
 
+	state prefixReplacingIteratorState
+	// Last bounds received via SetBounds.
+	lowerBound, upperBound []byte
 	// res is used to avoid allocations when rewriting result keys. It always
 	// starts with syntheticPrefix.
 	res InternalKey
 	err error
-	// empty is set after a seek operation that returns no keys.
-	empty bool
 }
 
-func errInputPrefixMismatch() error {
-	return errors.AssertionFailedf("key argument does not have prefix required for replacement")
-}
+type prefixReplacingIteratorState int8
+
+const (
+	// inRange indicates that the prefix replacing iterator is "in sync" with the
+	// underlying iterator; any Next/Prev calls can be passed through.
+	inRange prefixReplacingIteratorState = iota
+	// afterRange indicates that our iterator is positioned after the synthetic
+	// prefix range. A Prev() call should return the last key/span in the range.
+	afterRange
+	// beforeRange indicates that our iterator is positioned after the synthetic
+	// prefix range. A Next() call should return the first key/span in the range.
+	beforeRange
+	empty
+)
 
 func errOutputPrefixMismatch() error {
 	return errors.AssertionFailedf("key returned does not have prefix required for replacement")
@@ -56,21 +67,20 @@ var _ Iterator = (*prefixReplacingIterator)(nil)
 // `syntheticPrefix`. Every key produced by the underlying iterator must have
 // `contentPrefix`.
 //
-// keyInRange is a valid key that starts with syntheticPrefix. When a seek
-// function is called with a key that does not start with syntheticPrefix,
-// keyInRange is used to determine if the key is before or after the synthetic
-// prefix range.
+// lower is a valid key that has syntheticPreficx and is a lower bound for all
+// keys produced by i. It is used to determine if an argument key is before
+// or after the range of keys produced by the iterator.
 //
 // INVARIANT: len(syntheticPrefix) > 0 && keyInRange stars with syntheticPrefix.
 func newPrefixReplacingIterator(
-	i Iterator, contentPrefix, syntheticPrefix []byte, keyInRange []byte, cmp base.Compare,
+	i Iterator, contentPrefix, syntheticPrefix []byte, lower []byte, cmp base.Compare,
 ) Iterator {
 	if invariants.Enabled {
 		if len(syntheticPrefix) == 0 {
 			panic("newPrefixReplacingIterator called without synthetic prefix")
 		}
-		if !bytes.HasPrefix(keyInRange, syntheticPrefix) {
-			panic(fmt.Sprintf("keyInRange %q does not have synthetic prefix %q", keyInRange, syntheticPrefix))
+		if !bytes.HasPrefix(lower, syntheticPrefix) {
+			panic(fmt.Sprintf("lower %q does not have synthetic prefix %q", lower, syntheticPrefix))
 		}
 	}
 	return &prefixReplacingIterator{
@@ -78,7 +88,7 @@ func newPrefixReplacingIterator(
 		cmp:             cmp,
 		contentPrefix:   contentPrefix,
 		syntheticPrefix: syntheticPrefix,
-		keyInRange:      keyInRange,
+		lower:           lower,
 		arg:             slices.Clone(contentPrefix),
 		arg2:            slices.Clone(contentPrefix),
 		res:             InternalKey{UserKey: slices.Clone(syntheticPrefix)},
@@ -121,14 +131,16 @@ func (p *prefixReplacingIterator) rewriteResult(
 func (p *prefixReplacingIterator) SeekGE(
 	key []byte, flags base.SeekGEFlags,
 ) (*InternalKey, base.LazyValue) {
-	p.empty = false
+	p.state = inRange
 	if !bytes.HasPrefix(key, p.syntheticPrefix) {
-		if p.cmp(key, p.keyInRange) > 0 {
-			p.empty = true
+		if p.cmp(key, p.lower) > 0 {
+			p.state = afterRange
 			return nil, base.LazyValue{}
 		}
-		// Key must be before the range; use First instead.
-		return p.rewriteResult(p.i.First())
+		// Key must be before the range; seek to the lower bound instead.
+		// We don't use First because we may miss out on optimizations passed
+		// through SeekEFlags.
+		key = p.lower
 	}
 	return p.rewriteResult(p.i.SeekGE(p.rewriteArg(key), flags))
 }
@@ -137,13 +149,15 @@ func (p *prefixReplacingIterator) SeekGE(
 func (p *prefixReplacingIterator) SeekPrefixGE(
 	prefix, key []byte, flags base.SeekGEFlags,
 ) (*InternalKey, base.LazyValue) {
-	p.empty = false
-	if invariants.Enabled && !bytes.HasPrefix(key, prefix) {
-		panic(fmt.Sprintf("key %q does not have prefix %q", key, prefix))
-	}
-	if !bytes.HasPrefix(prefix, p.syntheticPrefix) {
+	p.state = inRange
+	if !bytes.HasPrefix(prefix, p.syntheticPrefix) || !bytes.HasPrefix(key, p.syntheticPrefix) {
 		// We never produce keys with this prefix; we can return nil.
-		p.empty = true
+		if p.cmp(prefix, p.lower) < 0 {
+			// We still want to seek the underlying iterator to potentially enable
+			// optimizations passed through flags.
+			p.i.SeekGE(p.rewriteArg(p.lower), flags)
+		}
+		p.state = empty
 		return nil, base.LazyValue{}
 	}
 	return p.rewriteResult(p.i.SeekPrefixGE(p.rewriteArg2(prefix), p.rewriteArg(key), flags))
@@ -153,52 +167,84 @@ func (p *prefixReplacingIterator) SeekPrefixGE(
 func (p *prefixReplacingIterator) SeekLT(
 	key []byte, flags base.SeekLTFlags,
 ) (*InternalKey, base.LazyValue) {
-	p.empty = false
+	p.state = inRange
 	if !bytes.HasPrefix(key, p.syntheticPrefix) {
-		if p.cmp(key, p.keyInRange) < 0 {
-			p.empty = true
+		if p.cmp(key, p.lower) < 0 {
+			// Key before the range; no results.
+			p.state = beforeRange
 			return nil, base.LazyValue{}
 		}
 		// Key must be after the range. Use Last instead.
 		return p.rewriteResult(p.i.Last())
 	}
+
 	return p.rewriteResult(p.i.SeekLT(p.rewriteArg(key), flags))
 }
 
 // First implements the Iterator interface.
 func (p *prefixReplacingIterator) First() (*InternalKey, base.LazyValue) {
-	p.empty = false
+	p.state = inRange
 	return p.rewriteResult(p.i.First())
 }
 
 // Last implements the Iterator interface.
 func (p *prefixReplacingIterator) Last() (*InternalKey, base.LazyValue) {
-	p.empty = false
+	p.state = inRange
 	return p.rewriteResult(p.i.Last())
 }
 
 // Next implements the Iterator interface.
 func (p *prefixReplacingIterator) Next() (*InternalKey, base.LazyValue) {
-	if p.empty {
+	switch p.state {
+	case empty, afterRange:
 		return nil, base.LazyValue{}
+	case beforeRange:
+		p.state = inRange
+		if p.lowerBound != nil {
+			return p.rewriteResult(p.i.SeekGE(p.lowerBound, base.SeekGEFlagsNone))
+		}
+		return p.rewriteResult(p.i.First())
+	case inRange:
+		return p.rewriteResult(p.i.Next())
+	default:
+		panic("invalid iterator state")
 	}
-	return p.rewriteResult(p.i.Next())
 }
 
 // NextPrefix implements the Iterator interface.
 func (p *prefixReplacingIterator) NextPrefix(succKey []byte) (*InternalKey, base.LazyValue) {
-	if p.empty {
+	switch p.state {
+	case empty, afterRange:
 		return nil, base.LazyValue{}
+	case beforeRange:
+		p.state = inRange
+		if p.lowerBound != nil {
+			return p.rewriteResult(p.i.SeekGE(p.lowerBound, base.SeekGEFlagsNone))
+		}
+		return p.rewriteResult(p.i.First())
+	case inRange:
+		return p.rewriteResult(p.i.NextPrefix(succKey))
+	default:
+		panic("invalid iterator state")
 	}
-	return p.rewriteResult(p.i.NextPrefix(p.rewriteArg(succKey)))
 }
 
 // Prev implements the Iterator interface.
 func (p *prefixReplacingIterator) Prev() (*InternalKey, base.LazyValue) {
-	if p.empty {
+	switch p.state {
+	case empty, beforeRange:
 		return nil, base.LazyValue{}
+	case afterRange:
+		p.state = inRange
+		if p.upperBound != nil {
+			return p.rewriteResult(p.i.SeekLT(p.upperBound, base.SeekLTFlagsNone))
+		}
+		return p.rewriteResult(p.i.Last())
+	case inRange:
+		return p.rewriteResult(p.i.Prev())
+	default:
+		panic("invalid iterator state")
 	}
-	return p.rewriteResult(p.i.Prev())
 }
 
 // Error implements the Iterator interface.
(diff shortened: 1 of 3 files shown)


sstable/prefix_replacing_iterator.go
```diff
@@ -23,27 +23,36 @@ type prefixReplacingIterator struct {
...
+	err   error
+	state iteratorState
+}
+
+type iteratorState int8
+
+const (
+	// inSync indicates that the prefix replacing iterator is "in sync" with the
+	// underlying iterator; any Next/Prev calls can be passed through.
+	inSync iteratorState = iota
```
itsbilal, line 47: nit: this can be `inRange`, or alternatively `iterPosCur` if we want to really mimic wording in other iterators.

sstable/prefix_replacing_iterator.go
```diff
@@ -153,52 +165,75 @@ func (p *prefixReplacingIterator) SeekPrefixGE(
...
 
 // Next implements the Iterator interface.
 func (p *prefixReplacingIterator) Next() (*InternalKey, base.LazyValue) {
-	if p.empty {
+	switch p.state {
+	case empty, afterRange:
 		return nil, base.LazyValue{}
+	case beforeRange:
+		p.state = inSync
+		return p.rewriteResult(p.i.First())
```
itsbilal, line 201: This could be a violation of the internalIterator contract, right? We don't allow First() to be called if there's a lower bound, instead we want SeekGE(lower) to be called. Same with Last() and the upper bound. You might need to save the bounds in `SetBounds` so you can do the appropriate translation here instead of passing through a first/last to the child iter.

sstable/prefix_replacing_iterator.go
RaduBerinde: Done.

sstable/prefix_replacing_iterator.go
RaduBerinde: Good point, done. TFTR!

sstable/prefix_replacing_iterator.go
```diff
@@ -153,52 +167,84 @@ func (p *prefixReplacingIterator) SeekPrefixGE(
...
 // NextPrefix implements the Iterator interface.
 func (p *prefixReplacingIterator) NextPrefix(succKey []byte) (*InternalKey, base.LazyValue) {
-	if p.empty {
+	switch p.state {
+	case empty, afterRange:
 		return nil, base.LazyValue{}
+	case beforeRange:
+		p.state = inRange
+		if p.lowerBound != nil {
+			return p.rewriteResult(p.i.SeekGE(p.lowerBound, base.SeekGEFlagsNone))
```
itsbilal, line 222: this might need to be `SeekGE(succKey)` as the lower bound might not have the prefix we want. And same for the First() call below; we can actually unconditionally call `SeekGE(succKey)` here.

itsbilal (approved):
LGTM save for `NextPrefix` fix

sstable/prefix_replacing_iterator.go
RaduBerinde: I need to think about this some more. What I did is incorrect, we can't just pass a lower bound below without rewriting its prefix.. We need to basically figure out the intersection of the given bounds and the vtable bounds (most of that is happening today below us).

itsbilal:
Does it require that much more work? If it's just `NextPrefix` that's affected, the lazy solution is to just call the prefix replacing iter's own `SeekGE(succKey)` in that case.

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3344)* status: 0 of 3 files reviewed, 14 unresolved discussions (waiting on @dt, @itsbilal, @jbowens, and @RaduBerinde)

___
*[`sstable/prefix_replacing_iterator.go` line 157 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3344#-NrNzMqYE-ycxnKXG1vF:-NrbAYYD3m9QwhJRsf4c:bv7350p) ([raw file](https://github.com/cockroachdb/pebble/blob/e00f71528f7f8e66fc46ebcce9bc89fabbe61eea/sstable/prefix_replacing_iterator.go#L157)):*

@jbowens do you think this is enough to retain any seek optimizations?
</blockquote></details>

I'm a little confused here. I think all we need is to track whether we did the absolute positioning or ignored the SeekPrefixGE (see other comment below).
So we would remember positioned/unpositioned. The ignore case here is the only one which would result in a state transition to unpositioned. Any relative positioning call when unpositioned will return nil. Any absolute positioning call where the state was unpositioned would turn off the TrySeekUsingNext. This is following the same pattern as the next lower layer where we do
```
		if !i.lastBloomFilterMatched {
			// Iterator is not positioned based on last seek.
			flags = flags.DisableTrySeekUsingNext()
		}
```

I could be wrong, since we need to flesh out the reasoning in https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1708716407912329?thread_ts=1708549343.720779&cid=CAC6K3SLU
___
*[`sstable/prefix_replacing_iterator.go` line 26 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3344#-NrawHnZEYHX8eKkvckr:-NrawHnZEYHX8eKkvcks:b-e9uag6) ([raw file](https://github.com/cockroachdb/pebble/blob/74ccc4e6f9aeb7ddeac5d88460dd200545f552a7/sstable/prefix_replacing_iterator.go#L26)):*
> ```Go
> 	syntheticPrefix []byte
> 
> 	// lower is a valid key that has syntheticPreficx and is a lower bound for all
> ```

nit: syntheticPrefix

Aren't we requiring that `i` only produce keys with `contentPrefix`. So how can a key starting with `syntheticPrefix` be a lower bound for `i`?
___
*[`sstable/prefix_replacing_iterator.go` line 63 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3344#-NrawvwYEuiIvK2ijBMx:-NrawvwYEuiIvK2ijBMy:b-kae5yy) ([raw file](https://github.com/cockroachdb/pebble/blob/74ccc4e6f9aeb7ddeac5d88460dd200545f552a7/sstable/prefix_replacing_iterator.go#L63)):*
> ```Go
> var _ Iterator = (*prefixReplacingIterator)(nil)
> 
> // newPrefixReplacingIterator wraps an iterator over keys that have
> ```

nit: over point keys ...
___
*[`sstable/prefix_replacing_iterator.go` line 68 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3344#-NraxQ204EDtctBNok0Z:-NraxQ204EDtctBNok0_:b-2ofjmw) ([raw file](https://github.com/cockroachdb/pebble/blob/74ccc4e6f9aeb7ddeac5d88460dd200545f552a7/sstable/prefix_replacing_iterator.go#L68)):*
> ```Go
> // `contentPrefix`.
> //
> // lower is a valid key that has syntheticPreficx and is a lower bound for all
> ```

syntheticPrefix

Same question about "lower bound for all keys produced by i`.
___
*[`sstable/prefix_replacing_iterator.go` line 201 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3344#gh-1503185242:-Nrb5e360rSLQY6VtEWr:b-gz9foe) ([raw file](https://github.com/cockroachdb/pebble/blob/74ccc4e6f9aeb7ddeac5d88460dd200545f552a7/sstable/prefix_replacing_iterator.go#L201)):*

Good point, done. TFTR!
</blockquote></details>

I suspect we are trying to be too clever with this code.

The underlying iterator `i` already has the bounds, but they are in terms of `ContentPrefix`. So our difficulty arises when someone tries `SeekGE` above the upper bound or `SeekLT` below the lower bound, since we can't translate them into seek keys in terms of `ContentPrefix`. But `levelIter` shouldn't be doing that, since it loads the relevant file with the first key >= seek key in the case of `SeekGE` and < seek key in case of `SeekLT`. So how about we treat those seek cases as simply calls to `Last(); Next()` and `First(); Prev()`, since they are not performance sensitive (unless I am mistaken).

I think this eliminates the need for `beforeRange` and `afterRange` and `lowerBound, upperBound`.
___
*[`sstable/prefix_replacing_iterator.go` line 49 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3344#-Nrb2Hkr05GFj8v353Jw:-Nrb2Hkr05GFj8v353Jx:b-daox18) ([raw file](https://github.com/cockroachdb/pebble/blob/570d0ebd7c1ee2201c1c8a868a0598fbb6bc83e6/sstable/prefix_replacing_iterator.go#L49)):*
> ```Go
> 	// inRange indicates that the prefix replacing iterator is "in sync" with the
> 	// underlying iterator; any Next/Prev calls can be passed through.
> 	inRange prefixReplacingIteratorState = iota
> ```

is there a reason to default to `inRange` instead of making the iota value panic-worthy?
___
*[`sstable/prefix_replacing_iterator.go` line 56 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3344#-Nrb2C1wAzy9MR0vRDfF:-Nrb2C1wAzy9MR0vRDfG:bcb8agf) ([raw file](https://github.com/cockroachdb/pebble/blob/570d0ebd7c1ee2201c1c8a868a0598fbb6bc83e6/sstable/prefix_replacing_iterator.go#L56)):*
> ```Go
> 	// prefix range. A Next() call should return the first key/span in the range.
> 	beforeRange
> 	empty
> ```

comment for empty?
___
*[`sstable/prefix_replacing_iterator.go` line 136 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3344#-Nrb1aeb45RIwZjTNQFx:-Nrb1aeb45RIwZjTNQFy:bpicrj0) ([raw file](https://github.com/cockroachdb/pebble/blob/570d0ebd7c1ee2201c1c8a868a0598fbb6bc83e6/sstable/prefix_replacing_iterator.go#L136)):*
> ```Go
> 	p.state = inRange
> 	if !bytes.HasPrefix(key, p.syntheticPrefix) {
> 		if p.cmp(key, p.lower) > 0 {
> ```

this is an odd use of `p.lower` that is worth a comment. One would expect this check to happen against upper, but this works because lower and upper share `SyntheticPrefix`.
___
*[`sstable/prefix_replacing_iterator.go` line 142 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3344#-Nrb1Kz0ENpDvYrEGm7u:-Nrb1Kz0ENpDvYrEGm7v:bv1ye18) ([raw file](https://github.com/cockroachdb/pebble/blob/570d0ebd7c1ee2201c1c8a868a0598fbb6bc83e6/sstable/prefix_replacing_iterator.go#L142)):*
> ```Go
> 		// Key must be before the range; seek to the lower bound instead.
> 		// We don't use First because we may miss out on optimizations passed
> 		// through SeekEFlags.
> ```

SeekGEFlags
___
*[`sstable/prefix_replacing_iterator.go` line 143 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3344#-Nrb-psv3OBMPGt68T74:-Nrb-psv3OBMPGt68T75:b-iom83i) ([raw file](https://github.com/cockroachdb/pebble/blob/570d0ebd7c1ee2201c1c8a868a0598fbb6bc83e6/sstable/prefix_replacing_iterator.go#L143)):*
> ```Go
> 		// We don't use First because we may miss out on optimizations passed
> 		// through SeekEFlags.
> 		key = p.lower
> ```


Presumably `prefixReplacingIterator` is being used in the performance sensitive case after the sstable has been ingested, so this is in a level and being iterated over using `levelIter`. Is that accurate?
___
*[`sstable/prefix_replacing_iterator.go` line 226 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3344#-Nrb7QY62e6dAuCKwCXV:-Nrb7QY62e6dAuCKwCXW:b-jxs6h0) ([raw file](https://github.com/cockroachdb/pebble/blob/570d0ebd7c1ee2201c1c8a868a0598fbb6bc83e6/sstable/prefix_replacing_iterator.go#L226)):*
> ```Go
> 		return p.rewriteResult(p.i.First())
> 	case inRange:
> 		return p.rewriteResult(p.i.NextPrefix(succKey))
> ```

what is the contract regarding `succKey`. Are we guaranteed that `succKey` has `SyntheticPrefix`? I believe not since `levelIter` will just pass through. So we have to be able to handle a `succKey` that is greater than the upper, i.e.,

```
if !bytes.HasPrefix(succKey, p.syntheticPrefix) {
		if p.cmp(key, p.lower) > 0 {
                   // Go past the end of this sstable
                   ...
                   return
                 }  else {
                   //panic. Can't be calling NextPrefix on this iterator if it is already positioned after `succKey`
                 }
}
// succKey has synthetic prefix. So rewrite succKey and call on i.
```

___
*[`sstable/prefix_replacing_iterator.go` line 268 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3344#-Nrb3P-28r2xJYkJ6sDJ:-Nrb3P-3DEj49_Yvd-vg:bp9bg8x) ([raw file](https://github.com/cockroachdb/pebble/blob/570d0ebd7c1ee2201c1c8a868a0598fbb6bc83e6/sstable/prefix_replacing_iterator.go#L268)):*
> ```Go
> 	// is going to rewrite them itself or pass them to something e.g. vState that
> 	// will rewrite them.
> 	if x, ok := p.i.(interface{ SetBoundsWithSyntheticPrefix() bool }); ok && x.SetBoundsWithSyntheticPrefix() {
> ```

this seems error-prone. What if someone has a typo and causes us to fall into the path below. I guess tests will catch it quickly enough, but what iterator type do we wrap that does not implement `SetBoundsWithSyntheticPrefix`?
___
*[`sstable/reader_virtual.go` line 137 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3344#-Nray9Iy6LPmqaCXTR1J:-Nray9Iy6LPmqaCXTR1K:bmyrywt) ([raw file](https://github.com/cockroachdb/pebble/blob/74ccc4e6f9aeb7ddeac5d88460dd200545f552a7/sstable/reader_virtual.go#L137)):*
> ```Go
> 	i, err := v.reader.newIterWithBlockPropertyFiltersAndContext(
> 		ctx, transforms, lower, upper, filterer, useFilterBlock,
> 		stats, categoryAndQoS, statsCollector, rp, &v.vState)
> ```

Seems like since `virtualState.{lower,upper}` also have `SyntheticPrefix`, the narrowed bound in `virtualState.constrainBounds` also have `SyntheticPrefix`, so the inverted bounds will have `ContentPrefix`. Is my understanding correct?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3344)* status: 0 of 3 files reviewed, 14 unresolved discussions (waiting on @dt, @itsbilal, @jbowens, and @sumeerbhola)

___
*[`sstable/prefix_replacing_iterator.go` line 157 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3344#-NrNzMqYE-ycxnKXG1vF:-NrbHlAE3Uy32vYiRocM:blqhskz) ([raw file](https://github.com/cockroachdb/pebble/blob/e00f71528f7f8e66fc46ebcce9bc89fabbe61eea/sstable/prefix_replacing_iterator.go#L157)):*

I'm a little confused here. I think all we need is to track whether we did the absolute positioning or ignored the SeekPrefixGE (see other comment below).
So we would remember positioned/unpositioned. The ignore case here is the only one which would result in a state transition to unpositioned. Any relative positioning call when unpositioned will return nil. Any absolute positioning call where the state was unpositioned would turn off the TrySeekUsingNext. This is following the same pattern as the next lower layer where we do
```
		if !i.lastBloomFilterMatched {
			// Iterator is not positioned based on last seek.
			flags = flags.DisableTrySeekUsingNext()
		}
```

I could be wrong, since we need to flesh out the reasoning in https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1708716407912329?thread_ts=1708549343.720779&cid=CAC6K3SLU
</blockquote></details>

I see, so we would not call SeekGE here at all and for any further Seek we will `DisableTrySeekUsingNext`  in the flags. That makes sense but I don't have a good understanding of all the details.
___
*[`sstable/prefix_replacing_iterator.go` line 26 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3344#-NrawHnZEYHX8eKkvckr:-NrbHTnzDTDUws7jTML1:bedz36d) ([raw file](https://github.com/cockroachdb/pebble/blob/74ccc4e6f9aeb7ddeac5d88460dd200545f552a7/sstable/prefix_replacing_iterator.go#L26)):*

nit: syntheticPrefix

Aren't we requiring that `i` only produce keys with `contentPrefix`. So how can a key starting with `syntheticPrefix` be a lower bound for `i`?
</blockquote></details>

Yeah, the description is incorrect. It's a lower bound for all keys produced by i after prefix replacement.
___
*[`sstable/prefix_replacing_iterator.go` line 201 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3344#gh-1503185242:-NrbIH-b9UlT5evxEza5:bpn3nk1) ([raw file](https://github.com/cockroachdb/pebble/blob/74ccc4e6f9aeb7ddeac5d88460dd200545f552a7/sstable/prefix_replacing_iterator.go#L201)):*

I suspect we are trying to be too clever with this code.

The underlying iterator `i` already has the bounds, but they are in terms of `ContentPrefix`. So our difficulty arises when someone tries `SeekGE` above the upper bound or `SeekLT` below the lower bound, since we can't translate them into seek keys in terms of `ContentPrefix`. But `levelIter` shouldn't be doing that, since it loads the relevant file with the first key >= seek key in the case of `SeekGE` and < seek key in case of `SeekLT`. So how about we treat those seek cases as simply calls to `Last(); Next()` and `First(); Prev()`, since they are not performance sensitive (unless I am mistaken).

I think this eliminates the need for `beforeRange` and `afterRange` and `lowerBound, upperBound`.
</blockquote></details>

The problem is that if `SetBounds()` was passed down, then it's illegal to call `First` and/or `Last`.. We would need to Seek from a bound. But we need to "trim" those bounds to the synthetic prefix range and convert them to content-prefix.

Currently, this happens underneath us - we pass the unchanged bounds to `SingleLeveliterator.SetBounds` which deals with the prefix replacement. It's a big mess.
https://github.com/cockroachdb/pebble/blob/master/sstable/reader_iter_single_lvl.go#L367

I think the correct thing to do is move all that logic in here instead. But I'm kind of losing steam on making this work given that we may be able to switch to just prepending a synthetic prefix and doing that in `blockIter` (#3350).
___
*[`sstable/prefix_replacing_iterator.go` line 56 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3344#-Nrb2C1wAzy9MR0vRDfF:-NrbHQXVCnXBJMy7iqet:b-vnxt6n) ([raw file](https://github.com/cockroachdb/pebble/blob/570d0ebd7c1ee2201c1c8a868a0598fbb6bc83e6/sstable/prefix_replacing_iterator.go#L56)):*

comment for empty?
</blockquote></details>

Sorry, left over
___
*[`sstable/prefix_replacing_iterator.go` line 143 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3344#-Nrb-psv3OBMPGt68T74:-NrbHNn5CU9V0esKucZO:btqqu5l) ([raw file](https://github.com/cockroachdb/pebble/blob/570d0ebd7c1ee2201c1c8a868a0598fbb6bc83e6/sstable/prefix_replacing_iterator.go#L143)):*

Presumably `prefixReplacingIterator` is being used in the performance sensitive case after the sstable has been ingested, so this is in a level and being iterated over using `levelIter`. Is that accurate?
</blockquote></details>

Correct.
___
*[`sstable/prefix_replacing_iterator.go` line 226 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3344#-Nrb7QY62e6dAuCKwCXV:-NrbJmNy6L3EFT6jbzwI:be4y133) ([raw file](https://github.com/cockroachdb/pebble/blob/570d0ebd7c1ee2201c1c8a868a0598fbb6bc83e6/sstable/prefix_replacing_iterator.go#L226)):*

what is the contract regarding `succKey`. Are we guaranteed that `succKey` has `SyntheticPrefix`? I believe not since `levelIter` will just pass through. So we have to be able to handle a `succKey` that is greater than the upper, i.e.,

```
if !bytes.HasPrefix(succKey, p.syntheticPrefix) {
		if p.cmp(key, p.lower) > 0 {
                   // Go past the end of this sstable
                   ...
                   return
                 }  else {
                   //panic. Can't be calling NextPrefix on this iterator if it is already positioned after `succKey`
                 }
}
// succKey has synthetic prefix. So rewrite succKey and call on i.
```

</blockquote></details>

Hm, good point. The contract as written is that `succKey`  is the immediate successor key to the iterator's current key prefix: https://github.com/cockroachdb/pebble/blob/master/internal/base/iterator.go#L168

In practice at least, the immediate successor key should have the same synthetic prefix. But I learned to take these docs with a grain of salt :)
___
*[`sstable/prefix_replacing_iterator.go` line 268 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3344#-Nrb3P-28r2xJYkJ6sDJ:-NrbKuHtAmPX8FSFKjdK:bjibabg) ([raw file](https://github.com/cockroachdb/pebble/blob/570d0ebd7c1ee2201c1c8a868a0598fbb6bc83e6/sstable/prefix_replacing_iterator.go#L268)):*

this seems error-prone. What if someone has a typo and causes us to fall into the path below. I guess tests will catch it quickly enough, but what iterator type do we wrap that does not implement `SetBoundsWithSyntheticPrefix`?
</blockquote></details>

I don't think we wrap anything else. I agree this kind of thing is very error prone. The correct thing would be to do the bounds conversion here instead of below us.
___
*[`sstable/reader_virtual.go` line 137 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3344#-Nray9Iy6LPmqaCXTR1J:-NrbLR7C9i8ZJUF22_t6:b-jte40w) ([raw file](https://github.com/cockroachdb/pebble/blob/74ccc4e6f9aeb7ddeac5d88460dd200545f552a7/sstable/reader_virtual.go#L137)):*

Seems like since `virtualState.{lower,upper}` also have `SyntheticPrefix`, the narrowed bound in `virtualState.constrainBounds` also have `SyntheticPrefix`, so the inverted bounds will have `ContentPrefix`. Is my understanding correct?
</blockquote></details>

Right, the `constrainBounds` call in `single/twoLevelIterator.init()` will intersect `[lower, upper)` with `[vState.lower, upper)` then convert from SyntheticPrefix to ContentPrefix.

PR #3350, opened 2024-02-26
teach pebble about block level prefix synthesis

msbutler:
See individual commits

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -1130,6 +1130,9 @@ type ExternalFile struct {
 	// is accessed as if those keys all instead have prefix SyntheticPrefix.
 	//
 	// SyntheticPrefix must be a prefix of both SmallestUserKey and LargestUserKey.
+	//
+	// NB: If the SyntheticPrefix is non-empty and the ContentPrefix is empty,
+	// then the read path will conduct block level prefix synthesis.
 	ContentPrefix, SyntheticPrefix []byte
 
 	// SyntheticSuffix will replace the suffix of every key in the file during
--- a/internal/manifest/version.go
+++ b/internal/manifest/version.go
@@ -294,9 +294,14 @@ func (m *FileMetadata) SyntheticSeqNum() sstable.SyntheticSeqNum {
 
 // IterTransforms returns an sstable.IterTransforms that has SyntheticSeqNum set as needed.
 func (m *FileMetadata) IterTransforms() sstable.IterTransforms {
+	var syntheticPrefix []byte
+	if m.PrefixReplacement != nil && !m.PrefixReplacement.UsePrefixReplacementIterator() {
+		syntheticPrefix = m.PrefixReplacement.SyntheticPrefix
+	}
 	return sstable.IterTransforms{
 		SyntheticSeqNum: m.SyntheticSeqNum(),
 		SyntheticSuffix: m.SyntheticSuffix,
+		SyntheticPrefix: syntheticPrefix,
 	}
 }
 
--- a/sstable/block.go
+++ b/sstable/block.go
@@ -5,8 +5,10 @@
 package sstable
 
 import (
+	"bytes"
 	"context"
 	"encoding/binary"
+	"slices"
 	"unsafe"
 
 	"github.com/cockroachdb/errors"
@@ -416,7 +418,9 @@ type blockIter struct {
 	// replacement rule, this will not be the case. Therefore, key should never
 	// be used after ikey is set.
 	key []byte
-	// fullKey is a buffer used for key prefix decompression.
+	// fullKey is a buffer used for key prefix decompression. Note that if
+	// transforms.SyntheticPrifix is not nil, fullKey always starts with that
+	// prefix.
 	fullKey []byte
 	// val contains the value the iterator is currently pointed at. If non-nil,
 	// this points to a slice of the block data.
@@ -454,7 +458,8 @@ type blockIter struct {
 		vbr            *valueBlockReader
 		hasValuePrefix bool
 	}
-	synthSuffixBuf []byte
+	synthSuffixBuf            []byte
+	firstUserKeyWithPrefixBuf []byte
 }
 
 // blockIter implements the base.InternalIterator interface.
@@ -484,7 +489,11 @@ func (i *blockIter) init(cmp Compare, split Split, block block, transforms IterT
 	i.numRestarts = numRestarts
 	i.ptr = unsafe.Pointer(&block[0])
 	i.data = block
-	i.fullKey = i.fullKey[:0]
+	if i.transforms.SyntheticPrefix != nil {
+		i.fullKey = append(i.fullKey[:0], i.transforms.SyntheticPrefix...)
+	} else {
+		i.fullKey = i.fullKey[:0]
+	}
 	i.val = nil
 	i.clearCache()
 	if i.restarts > 0 {
@@ -528,10 +537,11 @@ func (i *blockIter) isDataInvalidated() bool {
 
 func (i *blockIter) resetForReuse() blockIter {
 	return blockIter{
-		fullKey:   i.fullKey[:0],
-		cached:    i.cached[:0],
-		cachedBuf: i.cachedBuf[:0],
-		data:      nil,
+		fullKey:                   i.fullKey[:0],
+		cached:                    i.cached[:0],
+		cachedBuf:                 i.cachedBuf[:0],
+		firstUserKeyWithPrefixBuf: i.firstUserKeyWithPrefixBuf[:0],
+		data:                      nil,
 	}
 }
 
@@ -602,7 +612,7 @@ func (i *blockIter) readEntry() {
 		value = uint32(e)<<28 | uint32(d)<<21 | uint32(c)<<14 | uint32(b)<<7 | uint32(a)
 		ptr = unsafe.Pointer(uintptr(ptr) + 5)
 	}
-
+	shared += uint32(len(i.transforms.SyntheticPrefix))
 	unsharedKey := getBytes(ptr, int(unshared))
 	// TODO(sumeer): move this into the else block below.
 	i.fullKey = append(i.fullKey[:shared], unsharedKey...)
@@ -678,6 +688,12 @@ func (i *blockIter) readFirstKey() error {
 	} else {
 		i.firstUserKey = nil
 		return base.CorruptionErrorf("pebble/table: invalid firstKey in block")
+	}
+	if i.transforms.SyntheticPrefix != nil {
+		i.firstUserKeyWithPrefixBuf = slices.Grow(i.firstUserKeyWithPrefixBuf[:0], len(i.transforms.SyntheticPrefix)+len(i.firstUserKey))
+		i.firstUserKeyWithPrefixBuf = append(i.firstUserKeyWithPrefixBuf, i.transforms.SyntheticPrefix...)
+		i.firstUserKeyWithPrefixBuf = append(i.firstUserKeyWithPrefixBuf, i.firstUserKey...)
+		i.firstUserKey = i.firstUserKeyWithPrefixBuf
 	}
 	return nil
 }
@@ -757,6 +773,24 @@ func (i *blockIter) SeekGE(key []byte, flags base.SeekGEFlags) (*InternalKey, ba
 	if invariants.Enabled && i.isDataInvalidated() {
 		panic(errors.AssertionFailedf("invalidated blockIter used"))
 	}
+	searchKey := key
+	if i.transforms.SyntheticPrefix != nil {
+		// The seek key is before or after the entire block of keys that start with
+		// SyntheticPrefix. To determine which, we need to compare against a valid
+		// key in the block. We use firstUserKey which has the synthetic prefix.
+		if !bytes.HasPrefix(key, i.transforms.SyntheticPrefix) {
+			if i.cmp(i.firstUserKey, key) >= 0 {
+				return i.First()
+			}
+			// Set the offset to the end of the block to mimic the offset of an
+			// invalid iterator. This ensures a subsequent i.Prev() returns a valid
+			// result.
+			i.offset = i.restarts
+			i.nextOffset = i.restarts
+			return nil, base.LazyValue{}
+		}
+		searchKey = key[len(i.transforms.SyntheticPrefix):]
+	}
 
 	i.clearCache()
 	// Find the index of the smallest restart point whose key is > the key
@@ -821,7 +855,7 @@ func (i *blockIter) SeekGE(key []byte, flags base.SeekGEFlags) (*InternalKey, ba
 			}
 			// Else k is invalid, and left as nil
 
-			if i.cmp(key, k) > 0 {
+			if i.cmp(searchKey, k) > 0 {
 				// The search key is greater than the user key at this restart point.
 				// Search beyond this restart point, since we are trying to find the
 				// first restart point with a user key >= the search key.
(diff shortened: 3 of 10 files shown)


RaduBerinde (commented):
Nice! This is so much simpler than the prefix iterator approach (which I'm still trying to get to work in all cases)..

We should add checks in `NewRawRangeDelIter` and `NewRawRangeKeyIter` and error out since we can't handle the transform there.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3350#-:-NrbB_SB3t3YzhyHbcnh:bse1s22)* status: 0 of 10 files reviewed, 5 unresolved discussions (waiting on @msbutler)

___
*[`-- commits` line 10 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3350#-Nrb8lPn2tlgTDDDMHi1):*
This slowdown is when prefix synthesis is not enabled, correct?
___
*[`sstable/block.go` line 774 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3350#-Nrb9-gW7zsAAqakzZGP:-Nrb9-gW7zsAAqakzZGQ:b-rrum95) ([raw file](https://github.com/cockroachdb/pebble/blob/bc01f73115c15c1553bc5417524a9e6b3621b522/sstable/block.go#L774)):*
> ```Go
> 	if i.transforms.SyntheticPrefix != nil {
> 		if !bytes.HasPrefix(key, i.transforms.SyntheticPrefix) {
> 			if i.cmp(i.transforms.SyntheticPrefix, key) >= 0 {
> ```

This is a problem - `cmp` is only defined for full keys. The CRDB comparer looks at the last byte and treats it as a suffix length so things would break.
What the prefix replacing iterator does is to also pass a "test" key in the synthetic prefix range (I use the vtable lower bound) and use that for the comparison.
___
*[`sstable/reader.go` line 173 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3350#-NrbA9kl6pdv23FukW8l:-NrbA9kl6pdv23FukW8m:b2ngu5r) ([raw file](https://github.com/cockroachdb/pebble/blob/bc01f73115c15c1553bc5417524a9e6b3621b522/sstable/reader.go#L173)):*
> ```Go
> }
> 
> // SyntheticPrefix represents a byte slice that it implicitly prepended to every
> ```

*is
___
*[`sstable/reader.go` line 175 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3350#-NrbAUHa5JhF4ZTbUP8k:-NrbAUHa5JhF4ZTbUP8l:b-3515r1) ([raw file](https://github.com/cockroachdb/pebble/blob/bc01f73115c15c1553bc5417524a9e6b3621b522/sstable/reader.go#L175)):*
> ```Go
> // SyntheticPrefix represents a byte slice that it implicitly prepended to every
> // key in a file being read or accessed by a reader. Specifically, the file is
> // implied to contain keys that do not explicitly include the prefix, and the
> ```

This explanation is a bit confusing.. maybe say that the table is assumed to contain pieces of keys that become full keys when prepended a valid prefix. The table's bloom filters are constructed only on the pieces of keys in the table, but interactions with the file ...


One question I have - are the keys / "key pieces" in the table required to be full keys (i.e. can they be passed to `Compare` or `Split`?). I believe they will be in practice but we don't depend on that as long as they're valid with the SyntheticPrefix?
___
*[`sstable/reader_common.go` line 90 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3350#-NrbB3jN4HT9reH6proj:-NrbB3jN4HT9reH6prok:b72kxtc) ([raw file](https://github.com/cockroachdb/pebble/blob/cea741726e4cae1d5339b2fda6c5cffea2038bd3/sstable/reader_common.go#L90)):*
> ```Go
> 		return false
> 	}
> 	if p.ContentPrefix != nil && p.SyntheticPrefix != nil {
> ```

I'd compare lengths just in case. I think it's enough to check ContentPrefix

msbutler (commented):
Added those invariants. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3350#-:-NrbbIaeALJLeep19zHV:b-ljojj7)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @msbutler and @RaduBerinde)

___
*[`-- commits` line 10 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3350#-Nrb8lPn2tlgTDDDMHi1):*

This slowdown is when prefix synthesis is not enabled, correct?
</blockquote></details>

Indeed. Amended the commit msg to clarify
___
*[`sstable/block.go` line 774 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3350#-Nrb9-gW7zsAAqakzZGP:-NrbWrBt34C3983NvCYo:bdw4ij6) ([raw file](https://github.com/cockroachdb/pebble/blob/bc01f73115c15c1553bc5417524a9e6b3621b522/sstable/block.go#L774)):*

This is a problem - `cmp` is only defined for full keys. The CRDB comparer looks at the last byte and treats it as a suffix length so things would break.
What the prefix replacing iterator does is to also pass a "test" key in the synthetic prefix range (I use the vtable lower bound) and use that for the comparison.
</blockquote></details>

Changed it to use i.FirstUserKey which gets hydrated at init. 

I believe we could also use bytes.Compare() since the syntheticPrefix will never contain suffix bytes (which have weird ordering), but I think the FirstUserKey approach is cleaner. 
___
*[`sstable/reader.go` line 175 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3350#-NrbAUHa5JhF4ZTbUP8k:-NrbY3PCDYZLPJtKtgo1:b-3nzwan) ([raw file](https://github.com/cockroachdb/pebble/blob/bc01f73115c15c1553bc5417524a9e6b3621b522/sstable/reader.go#L175)):*

This explanation is a bit confusing.. maybe say that the table is assumed to contain pieces of keys that become full keys when prepended a valid prefix. The table's bloom filters are constructed only on the pieces of keys in the table, but interactions with the file ...


One question I have - are the keys / "key pieces" in the table required to be full keys (i.e. can they be passed to `Compare` or `Split`?). I believe they will be in practice but we don't depend on that as long as they're valid with the SyntheticPrefix?
</blockquote></details>

Done.  Since a prefix-less key will always contain a full suffix, I believe all compare operations will be valid on it. Might be worth adding some invariants here.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3350)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @msbutler)

___
*[`sstable/block.go` line 774 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3350#-Nrb9-gW7zsAAqakzZGP:-Nrbf59Y8h5yNeG-hXVo:b-bws26t) ([raw file](https://github.com/cockroachdb/pebble/blob/bc01f73115c15c1553bc5417524a9e6b3621b522/sstable/block.go#L774)):*

Changed it to use i.FirstUserKey which gets hydrated at init. 

I believe we could also use bytes.Compare() since the syntheticPrefix will never contain suffix bytes (which have weird ordering), but I think the FirstUserKey approach is cleaner.
</blockquote></details>

This deserves a comment. Something like:

The  seek key is before or after the entire range of keys that start with SyntheticPrefix. To determine which, we need to compare against a valid key in this range. We use `firstUserKey` which has the synthetic prefix.
___
*[`sstable/reader_common.go` line 90 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3350#-Nrbgykq9Ohq0lWeKrqb:-Nrbgykq9Ohq0lWeKrqc:b-32yzm5) ([raw file](https://github.com/cockroachdb/pebble/blob/bea0a4f0624a4b41cbaf8d9686d84c53d54a82fa/sstable/reader_common.go#L90)):*
> ```Go
> 		return false
> 	}
> 	if len(p.ContentPrefix)>0{
> ```

[nit] formatting (do you have `crlfmt` or `gofmt` set up to run automatically?)
___
*[`testdata/ingest_external` line 525 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3350#-NrbhH_ECQ-iDuf8MK3r:-NrbhH_ECQ-iDuf8MK3s:b-alcz2g) ([raw file](https://github.com/cockroachdb/pebble/blob/bea0a4f0624a4b41cbaf8d9686d84c53d54a82fa/testdata/ingest_external#L525)):*
> ```
> ----
> 
> # Synthesize prefix with one greater than the original one.
> ```

These comments don't make sense, we are replacing the empty prefix in both cases.

msbutler (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3350)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @msbutler and @RaduBerinde)

___
*[`sstable/reader_common.go` line 90 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3350#-Nrbgykq9Ohq0lWeKrqb:-Nrf9YwuBJxtpXK2bqHy:b-1t256h) ([raw file](https://github.com/cockroachdb/pebble/blob/bea0a4f0624a4b41cbaf8d9686d84c53d54a82fa/sstable/reader_common.go#L90)):*

[nit] formatting (do you have `crlfmt` or `gofmt` set up to run automatically?)
</blockquote></details>

i've been running it manually with the following cmd:
```
❯ which mbfmt
mbfmt () {
	echo crlfmt
	git diff-tree --no-commit-id --name-only -r HEAD | grep --color=auto --exclude-dir={.bzr,CVS,.git,.hg,.svn,.idea,.tox} '.go' | xargs -n1 crlfmt -tab 2 -w
}
```

it's not great with multi commit prs. I should revisit running it automatically via my IDE.
___
*[`testdata/ingest_external` line 525 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3350#-NrbhH_ECQ-iDuf8MK3r:-NrfAP0f4a-xRAijwJDR:b-cr06b7) ([raw file](https://github.com/cockroachdb/pebble/blob/bea0a4f0624a4b41cbaf8d9686d84c53d54a82fa/testdata/ingest_external#L525)):*

These comments don't make sense, we are replacing the empty prefix in both cases.
</blockquote></details>

Oops, bad copy pasta.

msbutler:
@RaduBerinde @sumeerbhola friendly ping to review this!

@sumeerbhola let me know if I should wait to merge until you have taken a look at this.

msbutler (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3350)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`sstable/block.go` line 420 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3350#-NrqDFKk7hbK4npj_lCS:-NrrCxoBBIm0SHqCZjBp:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/2e2c7479aee816e46ace9779c39b5182d25081a7/sstable/block.go#L420)):*

Mention that when a synthetic prefix is used, this always starts with that prefix.
</blockquote></details>

Done.
___
*[`sstable/block.go` line 691 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3350#-NrqAs9J3GmHR44GXwHc:-NrrHios4kZaoAgn7syj:b-fjh93d) ([raw file](https://github.com/cockroachdb/pebble/blob/2e2c7479aee816e46ace9779c39b5182d25081a7/sstable/block.go#L691)):*

[nit] this might allocate twice (not sure if Go "merges" the appends into a single operation). More importantly, we never actually set `firstUserKeyWithPrefixBuf` to anything, so we won't reuse buffers.

I'd replace with:
```
i.firstUserKeyWithPrefixBuf = slices.Grow(i.firstUserKeyWithPrefixBuf[:0], len(i.transforms.SyntheticPrefix) + len(i.firstUserKey))
i.firstUserKeyWithPrefixBuf = append(i.firstUserKeyWithPrefixBuf, i.transforms.SyntheticPrefix...)
i.firstUserKeyWithPrefixBuf = append(i.firstUserKeyWithPrefixBuf, i.firstUserKey...)
i.firstUserKey = i.firstUserKeyWithPrefixBuf
```
</blockquote></details>

Done.

To clarify, this change will increase the capacity of `firstUserKeyWithPrefixBuf`, if necessary. But we do reuse the buf when we reset the iter for reuse, right?
https://github.com/msbutler/pebble/blob/butler-rebased-prefix/sstable/block.go#L541
___
*[`sstable/block.go` line 964 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3350#-NrqFU0CELFETmpmkBIz:-NrrIK9I29tazIUuWbza:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/2e2c7479aee816e46ace9779c39b5182d25081a7/sstable/block.go#L964)):*

Worth adding a comment like above, mentioning that Next will return the first user key.
</blockquote></details>

Done.
___
*[`sstable/reader_common.go` line 86 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3350#-NrqH3bdB-SntutAxwrz:-NrrSWm9Cafn_rQMAxcr:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/6932834392ea89d26668407055665d144aeefc11/sstable/reader_common.go#L86)):*

[nit] return p != nil && len(p.ContentPrefix) > 0
</blockquote></details>

Done.
___
*[`sstable/reader_test.go` line 1381 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3350#-NrqHcSc2DygDvNViKGd:-NrrToV04TA3JK3hOGo1:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/6932834392ea89d26668407055665d144aeefc11/sstable/reader_test.go#L1381)):*

[nit] `keyToWrite = append(slices.Clip(prefix), key...)`
</blockquote></details>

Done.

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3350#-:-Nrr_mnH5_MXjaxQpIK0:bnfp4nl)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @sumeerbhola)

___
*[`sstable/block.go` line 691 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3350#-NrqAs9J3GmHR44GXwHc:-Nrr_LOF56w3dlc0D5Jz:bk9og6p) ([raw file](https://github.com/cockroachdb/pebble/blob/2e2c7479aee816e46ace9779c39b5182d25081a7/sstable/block.go#L691)):*

Done.

To clarify, this change will increase the capacity of `firstUserKeyWithPrefixBuf`, if necessary. But we do reuse the buf when we reset the iter for reuse, right?
https://github.com/msbutler/pebble/blob/butler-rebased-prefix/sstable/block.go#L541
</blockquote></details>

Right. I was saying that the previous version of the change (r7 on reviewable) did not ever set `firstUserKeyWithPrefixBuf` to anything, so it would always be `nil`

sumeerbhola (approved):
I did a quick pass, and have only a few nits
:lgtm: 

> Reviewed 5 of 10 files at r10, 1 of 1 files at r11, 2 of 9 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3350#-:-NruaLEj-99zL5VxVhNN:bmo3w3o)* status: 3 of 11 files reviewed, 4 unresolved discussions (waiting on @msbutler)

___
*[`-- commits` line 72 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3350#-Nru_rdgB4DneVUDj_wa):*
different than ...

isn't this the buggy path (that @RaduBerinde was trying to fix in https://github.com/cockroachdb/pebble/pull/3344) that we are abandoning in favor of that PR?
___
*[`sstable/block.go` line 783 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3350#-NruYy9PBsWPT9VvcnLZ:-NruYy9PBsWPT9VvcnL_:b-wl58k7) ([raw file](https://github.com/cockroachdb/pebble/blob/dae65f2b5a5ccaccb4464503ca1a799e388f6fc1/sstable/block.go#L783)):*
> ```Go
> 				// with SyntheticPrefix. To determine which, we need to compare against a
> 				// valid key in the block. We use firstUserKey which has the synthetic
> 				// prefix.
> ```

nit: a comment like this belongs before the predicate, since that is the line of code this is talking about.
___
*[`sstable/block.go` line 965 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3350#-NruZJAl9dPDSguM0rUq:-NruZJAl9dPDSguM0rUr:b-2dclgc) ([raw file](https://github.com/cockroachdb/pebble/blob/dae65f2b5a5ccaccb4464503ca1a799e388f6fc1/sstable/block.go#L965)):*
> ```Go
> 			// with SyntheticPrefix. To determine which, we need to compare against a
> 			// valid key in the block. We use firstUserKey which has the synthetic
> 			// prefix.
> ```

same nit
___
*[`sstable/reader.go` line 181 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3350#-Nru_N0HES3ayuZkNfJj:-Nru_N0HES3ayuZkNfJk:b-q5hq75) ([raw file](https://github.com/cockroachdb/pebble/blob/dae65f2b5a5ccaccb4464503ca1a799e388f6fc1/sstable/reader.go#L181)):*
> ```Go
> // keys that did include the prefix. Note that all Compare operations may act on
> // a prefix-less key as the synthetic prefix will never modify key metadata
> // stored in the key suffix.
> ```

Should this also say that such files must not contain rangedels or range keys, since the transformation is only applied to the point keys.

RaduBerinde:
*[`-- commits` line 72 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3350#-Nru_rdgB4DneVUDj_wa):*

different than ...

isn't this the buggy path (that @RaduBerinde was trying to fix in https://github.com/cockroachdb/pebble/pull/3344) that we are abandoning in favor of that PR?
</blockquote></details>

Yes, the plan is to remove that once Restore is able to generate "prefix free" ssts.

PR #3351, opened 2024-02-26
pebble: improve syntax for external ingestion test

RaduBerinde:
Use a better syntax for the external ingestion test, which allows
specifying prefix and suffix replacement per object.


RaduBerinde:
@msbutler I added `synthetic-prefix` as well so you don't have to modify it in your PR.

RaduBerinde:
TFTR!

PR #3355, opened 2024-02-27
metamorphic: fix invalid DeleteRange bounds

RaduBerinde:
Fix invalid DeleteRange bounds where the keys can be equal; this
causes a problem when it results in an external object with invalid
bounds.

We also add more assertions around bounds.

Informs #3325


itsbilal (approved):
LGTM

RaduBerinde:
TFTR!

PR #3357, opened 2024-02-28
db: require ExternalFile bounds without suffixes

RaduBerinde:
We already require this when `SyntheticSuffix` is used.

Also, use `Bounds KeyRange` instead of `SmallestUserKey/LargestUserKey`
in `ExternalFile`.


msbutler (approved):
Makes sense to me!

RaduBerinde:
TFTR!

PR #3358, opened 2024-02-28
base: default Split to trivial implementation

RaduBerinde:
This change makes `Comparer.EnsureDefaults` set a trivial `Split`
implementation which returns the full length of the key. This
simplifies code which had to check `Split != nil`, and we no longer
have to think about what are the semantics of a given operation when
`Split` is  `nil`.


sumeerbhola (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3358#-:-NrlWBUBAurGUrk1GVSv:b-uwkowd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jbowens)

RaduBerinde:
TFTR!

PR #3361, opened 2024-02-29
internal/metamorphic: use WALRecoveryDirs in cross-version tests

jbowens:
Cross-version metamorphic tests use an initial database state produced by a previous run of the metamorphic test with a different configuration. Since the previous and current configurations may use differing directories for WALs, we need to account for this difference during setup. Previously the test would copy all WALs into place. With the introduction of Options.WALRecoveryDirs, we can simply add the previous WAL directory as a recovery-only directory to ensure the relevant WALs are replayed during Open.

### DIFF
--- a/metamorphic/options.go
+++ b/metamorphic/options.go
@@ -23,6 +23,7 @@ import (
 	"github.com/cockroachdb/pebble/objstorage/remote"
 	"github.com/cockroachdb/pebble/sstable"
 	"github.com/cockroachdb/pebble/vfs"
+	"github.com/cockroachdb/pebble/wal"
 	"golang.org/x/exp/rand"
 )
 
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3361#-:-NrqWNEsDAlrTDdy1qBI:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`metamorphic/options.go` line 740 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3361#-NrqWFTE4Fqag9op3ERz:-NrqWFTE4Fqag9op3ES-:bwatt08) ([raw file](https://github.com/cockroachdb/pebble/blob/d85f25e3ea9865850d7a1a9484c80176c10598d7/metamorphic/options.go#L740)):*
> ```Go
> 	// directory itself as a 'WAL recovery dir' so that we'll read any WALs if
> 	// the previous test was writing them to the data directory.
> 	if testOpts.Opts.WALDir != "" {
> ```

[nit] this can be `else`? We could pull out the setting of `WALRecoveryDir` and set a string to either walDirPath or dataDir

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3361#-:-Nrqs5uh1aZGHlkr7vIa:b-814g8k)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @itsbilal)

___
*[`metamorphic/options.go` line 740 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3361#-NrqWFTE4Fqag9op3ERz:-Nrqs2WD0M5lldXMBom4:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/d85f25e3ea9865850d7a1a9484c80176c10598d7/metamorphic/options.go#L740)):*

[nit] this can be `else`? We could pull out the setting of `WALRecoveryDir` and set a string to either walDirPath or dataDir
</blockquote></details>

Done.

PR #3367, opened 2024-03-01
db: fix TestVirtualReadsWiring flake

### LINKED ISSUE #3364: db: lower bound violation in pebble tests
```
panic: lower bound f#12,21 violated by key f#14,21; wraps *keyspan.assertIter

goroutine 78074 [running]:
github.com/cockroachdb/pebble/internal/keyspan.(*assertIter).panicf(0xc00097dc20, {0x10f22786e?, 0x1?}, {0xc0008ac648?, 0x1?, 0x1?})
	/Users/runner/work/pebble/pebble/internal/keyspan/assert_iter.go:82 +0xa8
github.com/cockroachdb/pebble/internal/keyspan.(*assertIter).check(0xc00097dc20, 0xc0008074b0)
	/Users/runner/work/pebble/pebble/internal/keyspan/assert_iter.go:99 +0x1d4
github.com/cockroachdb/pebble/internal/keyspan.(*assertIter).First(0xc00097dc20)
	/Users/runner/work/pebble/pebble/internal/keyspan/assert_iter.go:134 +0x45
github.com/cockroachdb/pebble/internal/keyspan.(*filteringIter).First(0xc000807500)
	/Users/runner/work/pebble/pebble/internal/keyspan/filter.go:82 +0x23
github.com/cockroachdb/pebble/internal/keyspan.(*assertIter).First(0xc00097dcb0)
	/Users/runner/work/pebble/pebble/internal/keyspan/assert_iter.go:133 +0x23
github.com/cockroachdb/pebble/internal/keyspan.(*DefragmentingIter).First(0xc0002acb40)
	/Users/runner/work/pebble/pebble/internal/keyspan/defragment.go:289 +0x24
github.com/cockroachdb/pebble/internal/keyspan/keyspanimpl.(*MergingIter).First(0xc0003b9508)
	/Users/runner/work/pebble/pebble/internal/keyspan/keyspanimpl/merging_iter.go:615 +0x6a
github.com/cockroachdb/pebble.(*DB).loadTableRangeDelStats(0xc0008e4008, {0x10f5220b8, 0xc0009043c0}, 0xc0008f4c08, 0x6, 0xc0002856c0, 0xc0008acc08)
	/Users/runner/work/pebble/pebble/table_stats.go:375 +0xfb
github.com/cockroachdb/pebble.(*DB).loadTableStats.func1({0x10f5220b8, 0xc0009043c0})
	/Users/runner/work/pebble/pebble/table_stats.go:316 +0xf9
github.com/cockroachdb/pebble.(*tableCacheContainer).withCommonReader(0xc000072188?, 0xc0002856c0, 0xc0008acc50)
	/Users/runner/work/pebble/pebble/table_cache.go:295 +0xf3
github.com/cockroachdb/pebble.(*DB).loadTableStats(0x5?, 0x0?, 0x0?, 0x0?)
	/Users/runner/work/pebble/pebble/table_stats.go:305 +0xd4
github.com/cockroachdb/pebble.(*DB).scanReadStateTableStats(0xc0008e4008, 0xc0000b4d20, {0xc00019f490, 0x0, 0x32})
	/Users/runner/work/pebble/pebble/table_stats.go:281 +0x925
github.com/cockroachdb/pebble.(*DB).collectTableStats(0xc0008e4008)
	/Users/runner/work/pebble/pebble/table_stats.go:113 +0x236
created by github.com/cockroachdb/pebble.(*DB).maybeCollectTableStatsLocked in goroutine 78089
FAIL	github.com/cockroachdb/pebble	44.976s
```

https://github.com/cockroachdb/pebble/actions/runs/8102029032/job/22143601288?pr=3363

RaduBerinde:
Fixes #3364

### DIFF
--- a/table_cache_test.go
+++ b/table_cache_test.go
@@ -263,13 +263,13 @@ func TestVirtualReadsWiring(t *testing.T) {
 	b := newBatch(d)
 	// Some combination of sets, range deletes, and range key sets/unsets, so
 	// all of the table cache iterator functions are utilized.
-	require.NoError(t, b.Set([]byte{'a'}, []byte{'a'}, nil))
-	require.NoError(t, b.Set([]byte{'d'}, []byte{'d'}, nil))
-	require.NoError(t, b.DeleteRange([]byte{'c'}, []byte{'e'}, nil))
-	require.NoError(t, b.Set([]byte{'f'}, []byte{'f'}, nil))
-	require.NoError(t, b.RangeKeySet([]byte{'f'}, []byte{'k'}, nil, []byte{'c'}, nil))
-	require.NoError(t, b.RangeKeyUnset([]byte{'j'}, []byte{'k'}, nil, nil))
-	require.NoError(t, b.Set([]byte{'z'}, []byte{'z'}, nil))
+	require.NoError(t, b.Set([]byte{'a'}, []byte{'a'}, nil))                           // SeqNum start.
+	require.NoError(t, b.Set([]byte{'d'}, []byte{'d'}, nil))                           // SeqNum: +1
+	require.NoError(t, b.DeleteRange([]byte{'c'}, []byte{'e'}, nil))                   // SeqNum: +2
+	require.NoError(t, b.Set([]byte{'f'}, []byte{'f'}, nil))                           // SeqNum: +3
+	require.NoError(t, b.RangeKeySet([]byte{'f'}, []byte{'k'}, nil, []byte{'c'}, nil)) // SeqNum: +4
+	require.NoError(t, b.RangeKeyUnset([]byte{'j'}, []byte{'k'}, nil, nil))            // SeqNum: +5
+	require.NoError(t, b.Set([]byte{'z'}, []byte{'z'}, nil))                           // SeqNum: +6
 	require.NoError(t, d.Apply(b, nil))
 	require.NoError(t, d.Flush())
 	require.NoError(t, d.Compact([]byte{'a'}, []byte{'b'}, false))
@@ -287,6 +287,13 @@ func TestVirtualReadsWiring(t *testing.T) {
 	f2 := f1 + 1
 	d.mu.versions.nextFileNum += 2
 
+	seqNumA := parentFile.Smallest.SeqNum()
+	// See SeqNum comments above.
+	seqNumCEDel := seqNumA + 2
+	seqNumRangeSet := seqNumA + 4
+	seqNumRangeUnset := seqNumA + 5
+	seqNumZ := seqNumA + 6
+
 	v1 := &manifest.FileMetadata{
 		FileBacking:      parentFile.FileBacking,
 		FileNum:          f1,
@@ -294,10 +301,10 @@ func TestVirtualReadsWiring(t *testing.T) {
 		Size:             parentFile.Size / 2,
 		SmallestSeqNum:   parentFile.SmallestSeqNum,
 		LargestSeqNum:    parentFile.LargestSeqNum,
-		Smallest:         base.MakeInternalKey([]byte{'a'}, parentFile.Smallest.SeqNum(), InternalKeyKindSet),
-		Largest:          base.MakeInternalKey([]byte{'a'}, parentFile.Smallest.SeqNum(), InternalKeyKindSet),
-		SmallestPointKey: base.MakeInternalKey([]byte{'a'}, parentFile.Smallest.SeqNum(), InternalKeyKindSet),
-		LargestPointKey:  base.MakeInternalKey([]byte{'a'}, parentFile.Smallest.SeqNum(), InternalKeyKindSet),
+		Smallest:         base.MakeInternalKey([]byte{'a'}, seqNumA, InternalKeyKindSet),
+		Largest:          base.MakeInternalKey([]byte{'a'}, seqNumA, InternalKeyKindSet),
+		SmallestPointKey: base.MakeInternalKey([]byte{'a'}, seqNumA, InternalKeyKindSet),
+		LargestPointKey:  base.MakeInternalKey([]byte{'a'}, seqNumA, InternalKeyKindSet),
 		HasPointKeys:     true,
 		Virtual:          true,
 	}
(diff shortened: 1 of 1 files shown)


sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3367)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`table_cache_test.go` line 313 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3367#-NruOsZ48DOUVVoZy_Mg:-NruOsZ48DOUVVoZy_Mh:b-86t1o5) ([raw file](https://github.com/cockroachdb/pebble/blob/52a70a9357e572216b51dbe2d3d53249c45a0c90/table_cache_test.go#L313)):*
> ```Go
> 		SmallestSeqNum:   parentFile.SmallestSeqNum,
> 		LargestSeqNum:    parentFile.LargestSeqNum,
> 		Smallest:         base.MakeInternalKey([]byte{'d'}, parentFile.Smallest.SeqNum()+100, InternalKeyKindRangeDelete),
> ```

I am having trouble understanding this test and the failure `
panic: lower bound f#12,21 violated by key f#14,21; wraps *keyspan.assertIter
`

My current understanding:
- we are writing 7 keys in the batch so the range of seqnums should be `[parentFile.Smallest.SeqNum(), parentFile.Smallest.SeqNum()+7)`. To simplify, say the smallest seqnum was 1.
- The `d` point is the second thing written to the batch so it should have seqnum 2. The RANGEDEL containing `d` is the third thing written to the batch so has seqnum 3. So both are included in this sst when `SmallestPointKey: base.MakeInternalKey([]byte{'d'}, parentFile.Smallest.SeqNum()+100, InternalKeyKindRangeDelete)`? Why did +2 not work? It would be d@3, so [c, e)@3 would be cleanly split into [c,d)@3 and [d,e)@3 for the virtualization.

Some code comments would help.

RaduBerinde:
*[`table_cache_test.go` line 313 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3367#-NruOsZ48DOUVVoZy_Mg:-NruRxKA-E50OnGShxgn:b-2c77va) ([raw file](https://github.com/cockroachdb/pebble/blob/52a70a9357e572216b51dbe2d3d53249c45a0c90/table_cache_test.go#L313)):*

I am having trouble understanding this test and the failure `
panic: lower bound f#12,21 violated by key f#14,21; wraps *keyspan.assertIter
`

My current understanding:
- we are writing 7 keys in the batch so the range of seqnums should be `[parentFile.Smallest.SeqNum(), parentFile.Smallest.SeqNum()+7)`. To simplify, say the smallest seqnum was 1.
- The `d` point is the second thing written to the batch so it should have seqnum 2. The RANGEDEL containing `d` is the third thing written to the batch so has seqnum 3. So both are included in this sst when `SmallestPointKey: base.MakeInternalKey([]byte{'d'}, parentFile.Smallest.SeqNum()+100, InternalKeyKindRangeDelete)`? Why did +2 not work? It would be d@3, so [c, e)@3 would be cleanly split into [c,d)@3 and [d,e)@3 for the virtualization.

Some code comments would help.
</blockquote></details>

I can't say I understand this test very much either..  

The failure reproduces rarely, and only when there are a lot of things running in parallel.

I made a change to copy the sst, this is what it looks like when the test passes:
```
a#10,SET [61]
c-e#12,RANGEDEL
f#13,SET [66]
z#16,SET [7a]
[f-j):
  #14,RANGEKEYSET:  [63]
[j-k):
  #15,RANGEKEYUNSET: 
```

It seems to look exactly the same even when it fails:
```
a#10,SET [61]
c-e#12,RANGEDEL
f#13,SET [66]
z#16,SET [7a]
[f-j):
  #14,RANGEKEYSET:  [63]
[j-k):
  #15,RANGEKEYUNSET: 
```

I'll investigate more.

RaduBerinde:
*[`table_cache_test.go` line 313 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3367#-NruOsZ48DOUVVoZy_Mg:-NruZUYm-gpJRdCiO4EU:bcm0auw) ([raw file](https://github.com/cockroachdb/pebble/blob/52a70a9357e572216b51dbe2d3d53249c45a0c90/table_cache_test.go#L313)):*

I can't say I understand this test very much either..  

The failure reproduces rarely, and only when there are a lot of things running in parallel.

I made a change to copy the sst, this is what it looks like when the test passes:
```
a#10,SET [61]
c-e#12,RANGEDEL
f#13,SET [66]
z#16,SET [7a]
[f-j):
  #14,RANGEKEYSET:  [63]
[j-k):
  #15,RANGEKEYUNSET: 
```

It seems to look exactly the same even when it fails:
```
a#10,SET [61]
c-e#12,RANGEDEL
f#13,SET [66]
z#16,SET [7a]
[f-j):
  #14,RANGEKEYSET:  [63]
[j-k):
  #15,RANGEKEYUNSET: 
```

I'll investigate more.
</blockquote></details>

Oh.. The problem is with the range key. We set `f#12` as the lower bound. Nothing in the test causes us to read it, but a background `collectTableStats` does.

RaduBerinde:
*[`table_cache_test.go` line 313 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3367#-NruOsZ48DOUVVoZy_Mg:-NrubKvwDfQJoMuxT0Zw:b-31vuav) ([raw file](https://github.com/cockroachdb/pebble/blob/52a70a9357e572216b51dbe2d3d53249c45a0c90/table_cache_test.go#L313)):*

Oh.. The problem is with the range key. We set `f#12` as the lower bound. Nothing in the test causes us to read it, but a background `collectTableStats` does.
</blockquote></details>

Updated. This should make a lot more sense now. I will also separately try to update `InternalKey.String()` to show the kind as a string, it would have helped to see that instead of `,21`.

sumeerbhola (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3367#-:-NrvpOOFDVjl_K4E3iLH:b-uwkowd)* status: 0 of 1 files reviewed, 1 unresolved discussion

RaduBerinde:
TFTR!

PR #3371, opened 2024-03-01
metamorphic: add testing for prefix synthesis

RaduBerinde:
This commit adds prefix synthesis (i.e. `PrefixReplacement` with empty
`ContentPrefix`) to the metamorphic test.

We generate a random prefix and prepend it to the initially-generated
bounds. We take it into account when emulating the ingestion and when
determining key histories.

### DIFF
--- a/metamorphic/build.go
+++ b/metamorphic/build.go
@@ -30,6 +30,7 @@ func writeSSTForIngestion(
 	rangeKeyIter keyspan.FragmentIterator,
 	uniquePrefixes bool,
 	syntheticSuffix sstable.SyntheticSuffix,
+	prefixChange *sstable.PrefixReplacement,
 	writable objstorage.Writable,
 	targetFMV pebble.FormatMajorVersion,
 ) (*sstable.WriterMetadata, error) {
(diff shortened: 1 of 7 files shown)


RaduBerinde:
Ran under stress with higher weight for external ingest:
```
1040 runs so far, 0 failures, over 3h7m45s
```

msbutler (commented):
Looks good overall! Left one blocking question about why we need to invert the key bounds.

msbutler (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3371)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`metamorphic/build.go` line 208 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3371#gh-1511259074:-Ns9yTF42r3FIUy_RhSk:boqa4qd) ([raw file](https://github.com/cockroachdb/pebble/blob/7e08062abaa9b7eaf4d2361faa95c58a5236ea49/metamorphic/build.go#L208)):*

I don't get it, what does synthetic prefix have to do with the fact that the file has at most one key per prefix?
</blockquote></details>

ah yeah, nvm, i think i've been thinking too much about synthetic prefixes. Even if we prepend the synthetic prefixes, that statement hold. carry on :D
___
*[`metamorphic/build.go` line 247 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3371#gh-1511264241:-NsA-OGW7vRKIHC-bEjI:b-gr1is7) ([raw file](https://github.com/cockroachdb/pebble/blob/7e08062abaa9b7eaf4d2361faa95c58a5236ea49/metamorphic/build.go#L247)):*

The bounds include the synthetic prefix.. How would things work at all if we didn't invert them? In your case, they were inverted [above the blockIter](https://github.com/cockroachdb/pebble/blob/master/sstable/reader_iter_single_lvl.go#L208).
</blockquote></details>

ah, forgive my ignorance. I hadn't done a close enough reading on how synthetic prefixes affected iter bounds.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3371#-:-NsA296AD744h-Xw1WZ9:b-814g8k)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @itsbilal)

PR #3373, opened 2024-03-02
sstable: generate streamlined blockIter code

RaduBerinde:
#### sstable: generate streamlined blockIter code

We automatically generate a "streamlined" version of the blockIter code (in
block_iter_streamlined.gen.go). The streamlined code can be used in the
common case and does not support all features. This improves performance by
reducing the conditionals in the hot path.

The streamlinedBlockIter constant can be used in if statements; in the
streamlined version, it gets replaced by "true" (allowing the compiler to
omit blocks of code from the streamlined version).


RaduBerinde:
*[`sstable/block_iter_streamlined.gen.go` line 19 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3373#-NrwZZbx9KvW869WbZyu:-NrwZZbx9KvW869WbZyv:bb3pfcx) ([raw file](https://github.com/cockroachdb/pebble/blob/9ee5a1f2e09e343a34b682b178119aee806c3b37/sstable/block_iter_streamlined.gen.go#L19)):*
> ```Go
> func (i *blockIter) streamlinedReadEntry() {
> 	ptr := unsafe.Pointer(uintptr(i.ptr) + uintptr(i.offset))
> ```

Comments directly inside the method are missing. I couldn't quite figure out how to fix that, but it doesn't really matter (everyone should always work with the original code).

petermattis (commented):
Looks straightforward and understandable. Other storage folks should definitely give this a review, but I'm thumbs-up on an approach like this. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3373#-:-Ns-i85B6t2m43-tOFTk:bjrp7v3)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/block-iter-codegen/main.go` line 5 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3373#-Ns-h_9T84F3C0cs3sZ9:-Ns-h_9T84F3C0cs3sZA:b-8p07ba) ([raw file](https://github.com/cockroachdb/pebble/blob/53a81d3d8065cd2b6eb070644e82e840319c428d/sstable/block-iter-codegen/main.go#L5)):*
> ```Go
> // the LICENSE file.
> 
> package main
> ```

It's worthwhile to add a comment explaining what this code is doing. It is relatively straightforward, but I still suspect our future selves will thank us. Especially if this code grows more complex over time.
___
*[`sstable/block-iter-codegen/main.go` line 18 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3373#-Ns-h9Fj0kws1_Ezsmz4:-Ns-h9Fj0kws1_Ezsmz5:behspch) ([raw file](https://github.com/cockroachdb/pebble/blob/53a81d3d8065cd2b6eb070644e82e840319c428d/sstable/block-iter-codegen/main.go#L18)):*
> ```Go
> )
> 
> var methods = map[string]struct{}{
> ```

Rather than maintaining a list of methods like this, could you automatically deduce the methods by walking over the function declarations and marking any method which references the `streamlinedBlockIter` constant and any method which transitively calls such a method as needing a rewrite? I'm hoping you'll get the same list (though you might want to exclude `init`), but then we don't have the possibility of introducing another `streamlinedBlockIter` path and forget to update this list. 
___
*[`sstable/block-iter-codegen/main.go` line 46 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3373#-Ns-hi0F28Eh0X_tc5zf:-Ns-hi0F28Eh0X_tc5zg:bbcldbr) ([raw file](https://github.com/cockroachdb/pebble/blob/53a81d3d8065cd2b6eb070644e82e840319c428d/sstable/block-iter-codegen/main.go#L46)):*
> ```Go
> 	}
> 
> 	// Rename streamlinedBlockIter ot streamlinedBlockIterTrue.
> ```

Is this comment still valid? Seems like you're doing more than what the comment states.

Nit: `s/ot/to/g`

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3373)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @petermattis)

___
*[`sstable/block-iter-codegen/main.go` line 5 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3373#-Ns-h_9T84F3C0cs3sZ9:-Ns-rVkECOOg1RhfT2Ub:b-we9ppu) ([raw file](https://github.com/cockroachdb/pebble/blob/53a81d3d8065cd2b6eb070644e82e840319c428d/sstable/block-iter-codegen/main.go#L5)):*

It's worthwhile to add a comment explaining what this code is doing. It is relatively straightforward, but I still suspect our future selves will thank us. Especially if this code grows more complex over time.
</blockquote></details>

Agreed, it's just an initial draft
___
*[`sstable/block-iter-codegen/main.go` line 18 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3373#-Ns-h9Fj0kws1_Ezsmz4:-Ns-rfaO2_lDdzZMwpjx:b-acio2t) ([raw file](https://github.com/cockroachdb/pebble/blob/53a81d3d8065cd2b6eb070644e82e840319c428d/sstable/block-iter-codegen/main.go#L18)):*

Rather than maintaining a list of methods like this, could you automatically deduce the methods by walking over the function declarations and marking any method which references the `streamlinedBlockIter` constant and any method which transitively calls such a method as needing a rewrite? I'm hoping you'll get the same list (though you might want to exclude `init`), but then we don't have the possibility of introducing another `streamlinedBlockIter` path and forget to update this list.
</blockquote></details>

Interesting. I think I'd need some more advanced walking infrastructure than Inspect. Might not be worth the trouble, but I'll give it a shot

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3373)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/block-iter-codegen/main.go` line 18 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3373#-Ns-h9Fj0kws1_Ezsmz4:-Ns4LvsU72N1yuIqqUir:b-ma8lti) ([raw file](https://github.com/cockroachdb/pebble/blob/53a81d3d8065cd2b6eb070644e82e840319c428d/sstable/block-iter-codegen/main.go#L18)):*

Interesting. I think I'd need some more advanced walking infrastructure than Inspect. Might not be worth the trouble, but I'll give it a shot
</blockquote></details>

Yeah, my suggestion may not be worth the effort. Perhaps just leave a TODO comment in case this becomes some sort of headache in the future.

RaduBerinde:
*[`sstable/block-iter-codegen/main.go` line 18 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3373#-Ns-h9Fj0kws1_Ezsmz4:-Ns53MxQ3Y0zpr8mUqTA:bon4vvg) ([raw file](https://github.com/cockroachdb/pebble/blob/53a81d3d8065cd2b6eb070644e82e840319c428d/sstable/block-iter-codegen/main.go#L18)):*

Yeah, my suggestion may not be worth the effort. Perhaps just leave a TODO comment in case this becomes some sort of headache in the future.
</blockquote></details>

I looked at it briefly and I think it's not bad at all, when I get it working I'll update and we can decide then.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3373)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @petermattis)

___
*[`sstable/block-iter-codegen/main.go` line 5 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3373#-Ns-h_9T84F3C0cs3sZ9:-Ns9z1Yd3HjFwoK6cPxL:bs7wumx) ([raw file](https://github.com/cockroachdb/pebble/blob/53a81d3d8065cd2b6eb070644e82e840319c428d/sstable/block-iter-codegen/main.go#L5)):*

Agreed, it's just an initial draft
</blockquote></details>

Done. 
___
*[`sstable/block-iter-codegen/main.go` line 18 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3373#-Ns-h9Fj0kws1_Ezsmz4:-Ns9z2id-9ILAjQT1adM:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/53a81d3d8065cd2b6eb070644e82e840319c428d/sstable/block-iter-codegen/main.go#L18)):*

I looked at it briefly and I think it's not bad at all, when I get it working I'll update and we can decide then.
</blockquote></details>

Done.

petermattis (commented):
This all looks good to me, but let's get some other Storage :eyes: on this PR. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3373#-:-NsAEpVYBMqAD5hGGqiK:bugnk1a)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/block_iter_streamlined.gen.go` line 19 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3373#-NrwZZbx9KvW869WbZyu:-NsADp-33aK5CWcznJh8:bg9jb9l) ([raw file](https://github.com/cockroachdb/pebble/blob/9ee5a1f2e09e343a34b682b178119aee806c3b37/sstable/block_iter_streamlined.gen.go#L19)):*

Comments directly inside the method are missing. I couldn't quite figure out how to fix that, but it doesn't really matter (everyone should always work with the original code).
</blockquote></details>

Hmm, I thought that `parser.ParseComments` flag would have handled keeping the comments around. Perhaps there is a similar flag that needs to be specified when outputting the AST. I see there is a `printer.CommentedNode` type. I agree that this isn't worth worrying about.
___
*[`sstable/block-iter-codegen/main.go` line 18 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3373#-Ns-h9Fj0kws1_Ezsmz4:-NsAEmOH9mCvcNbeJjT0:b-83sjoh) ([raw file](https://github.com/cockroachdb/pebble/blob/53a81d3d8065cd2b6eb070644e82e840319c428d/sstable/block-iter-codegen/main.go#L18)):*

Done.
</blockquote></details>

Nice!
___
*[`sstable/block-iter-codegen/main.go` line 10 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3373#-NsADNDq7eg7NSswyqOC:-NsADNDq7eg7NSswyqOD:b8gfb5c) ([raw file](https://github.com/cockroachdb/pebble/blob/ee9934ef6050555ed753ec69db460fc6014f6e83/sstable/block-iter-codegen/main.go#L10)):*
> ```Go
> //
> // The block_iter.go code uses a special constant streamlinedBlockIter which is
> // always false in th code. This program generates streamlined versions of the
> ```

Nit: `s/th/the/g`

RaduBerinde (commented):
TFTR! I will rope in some more reviewers once I wire it up and maybe demonstrate some gains by inlining the crdb comparer.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3373#-:-NsB7JXZ6NEc1FSdfZbU:b-2j0dat)* status: 0 of 7 files reviewed, 3 unresolved discussions (waiting on @petermattis)

___
*[`sstable/block_iter_streamlined.gen.go` line 19 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3373#-NrwZZbx9KvW869WbZyu:-NsB68o46Eqt-8xS0Hyv:byrf5uu) ([raw file](https://github.com/cockroachdb/pebble/blob/9ee5a1f2e09e343a34b682b178119aee806c3b37/sstable/block_iter_streamlined.gen.go#L19)):*

Hmm, I thought that `parser.ParseComments` flag would have handled keeping the comments around. Perhaps there is a similar flag that needs to be specified when outputting the AST. I see there is a `printer.CommentedNode` type. I agree that this isn't worth worrying about.
</blockquote></details>

Yeah, me too. It does keep most comments, just not the ones that are at the top level inside the body. Apparently they are in separate nodes or something like that.

RaduBerinde:
*[`sstable/block_iter_streamlined.gen.go` line 19 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3373#-NrwZZbx9KvW869WbZyu:-NsB9NvN-WHl1-p4nMIU:b-pmsjup) ([raw file](https://github.com/cockroachdb/pebble/blob/9ee5a1f2e09e343a34b682b178119aee806c3b37/sstable/block_iter_streamlined.gen.go#L19)):*

Yeah, me too. It does keep most comments, just not the ones that are at the top level inside the body. Apparently they are in separate nodes or something like that.
</blockquote></details>

Yeah, the missing ones all show up under `ast.File.Comments`

PR #3381, opened 2024-03-06
db: move skip-remote logic out of manifest.LevelIterator

itsbilal:
Previously, we were adding remote file skipping logic to LevelIterator alongside the logic to skip files without point/range keys. Since determining whether a file is remote or not requires an objprovider call, this change moves that to iterator construction and undo's the change to leveliterator.

### DIFF
--- a/db.go
+++ b/db.go
@@ -1335,7 +1335,9 @@ func finishInitializingInternalIter(
 	}
 	i.initializeBoundBufs(i.opts.LowerBound, i.opts.UpperBound)
 
-	i.constructPointIter(i.opts.CategoryAndQoS, memtables, buf)
+	if err := i.constructPointIter(i.opts.CategoryAndQoS, memtables, buf); err != nil {
+		return nil, err
+	}
 
 	// For internal iterators, we skip the lazy combined iteration optimization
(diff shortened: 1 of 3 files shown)


stevendanna (approved):
Much nicer, thanks for following up on this.

itsbilal:
TFTR!

RaduBerinde:
*[`scan_internal.go` line 936 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3381#-NsJpSkh6TCRJah54pAq:-NsJpSki9Ud29RO5OJMg:b-k8udn7) ([raw file](https://github.com/cockroachdb/pebble/blob/641914d2f0a846ec529edcf8d01ff400e920dc1f/scan_internal.go#L936)):*
> ```Go
> 			}
> 			levSlice := manifest.NewLevelSliceKeySorted(i.db.cmp, nonRemoteFiles)
> 			levIter = levSlice.Iter()
> ```

Do we need to close the old `levIter` here?

stevendanna:
@RaduBerinde It looked to me like LevelIterator didn't have a Close()?

RaduBerinde:
Ah, I accidentally looked at `LevelIter`.

PR #3383, opened 2024-03-06
db: rework backing refcount management

RaduBerinde:
Previously, the logic for maintaining the `virtualBackings` map was very convoluted; in addition, modification happened outside `DB.mu` which makes the map unusable outside the log lock (we will want to use it for finding existing external backings).

`Apply` is no longer responsible for updating the backing refcounts and the `virtualBackings` map (and for updating a zombies map). It is more straightforward to do that separately (and under `DB.mu`); this allows us to remove `AccumulateIncompleteAndApplySingleVE` as well, which was semantically dubious (e.g. it modified `ve`).

### DIFF
--- a/internal/keyspan/keyspanimpl/level_iter_test.go
+++ b/internal/keyspan/keyspanimpl/level_iter_test.go
@@ -310,7 +310,7 @@ func TestLevelIterEquivalence(t *testing.T) {
 				amap[metas[i].FileNum] = metas[i]
 			}
 			b.Added[6] = amap
-			v, err := b.Apply(nil, base.DefaultComparer, 0, 0, nil)
+			v, err := b.Apply(nil, base.DefaultComparer, 0, 0)
 			require.NoError(t, err)
 			levelIter.Init(
 				keyspan.SpanIterOptions{}, base.DefaultComparer.Compare, tableNewIters,
@@ -454,7 +454,7 @@ func TestLevelIter(t *testing.T) {
 					amap[metas[i].FileNum] = metas[i]
 				}
 				b.Added[6] = amap
-				v, err := b.Apply(nil, base.DefaultComparer, 0, 0, nil)
+				v, err := b.Apply(nil, base.DefaultComparer, 0, 0)
 				require.NoError(t, err)
 				iter = NewLevelIter(
 					keyspan.SpanIterOptions{}, base.DefaultComparer.Compare,
--- a/internal/manifest/l0_sublevels_test.go
+++ b/internal/manifest/l0_sublevels_test.go
@@ -51,7 +51,7 @@ func readManifest(filename string) (*Version, error) {
 		if err := bve.Accumulate(&ve); err != nil {
 			return nil, err
 		}
-		if v, err = bve.Apply(v, base.DefaultComparer, 10<<20, 32000, nil); err != nil {
+		if v, err = bve.Apply(v, base.DefaultComparer, 10<<20, 32000); err != nil {
 			return nil, err
 		}
 	}
--- a/internal/manifest/manifest_test.go
+++ b/internal/manifest/manifest_test.go
@@ -124,7 +124,7 @@ func replayManifest(t *testing.T, opts *pebble.Options, dirname string) *manifes
 	}
 	v, err := bve.Apply(
 		nil /* version */, cmp, opts.FlushSplitBytes,
-		opts.Experimental.ReadCompactionRate, nil /* zombies */)
+		opts.Experimental.ReadCompactionRate)
 	require.NoError(t, err)
 	return v
 }
--- a/internal/manifest/testdata/version_edit_apply
+++ b/internal/manifest/testdata/version_edit_apply
@@ -16,7 +16,6 @@ edit
 2:
   000001:[a#1,SET-b#2,SET]
   000004:[c#3,SET-d#4,SET]
-zombies []
 
 apply
  L0
@@ -59,7 +58,6 @@ edit
   000002:[c#3,SET-d#4,SET]
 0.0:
   000001:[a#1,SET-c#2,SET]
-zombies []
 
 apply
  L0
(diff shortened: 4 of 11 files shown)


jbowens (approved):
:lgtm:

> Reviewed 9 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3383#-:-NsTZosx0o9JzvWLiHoi:bnfp4nl)* status: 9 of 11 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`version_set.go` line 489 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3383#-NsTYZvb-nHxkoVHCUF1:-NsTYZvc-vIrFUujGTy9:bnm4p03) ([raw file](https://github.com/cockroachdb/pebble/blob/45a55c12738c7907a9422a46032d3a032c590348/version_set.go#L489)):*
> ```Go
> 
> 	// Update backing refcounts and populate RemovedBackingTables.
> 	zombies := make(map[base.DiskFileNum]uint64)
> ```

nit: should we try to avoid this allocation if there are no zombies produced the version edit?
___
*[`version_set.go` line 494 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3383#-NsTYuhv0w72I9Nku63m:-NsTYuhv0w72I9Nku63n:btdu0bs) ([raw file](https://github.com/cockroachdb/pebble/blob/45a55c12738c7907a9422a46032d3a032c590348/version_set.go#L494)):*
> ```Go
> 	}
> 	for _, nf := range ve.NewFiles {
> 		nf.Meta.LatestRef()
> ```

I guess the other users of BulkVersionEdit (eg, tool/) don't actually require maintenance of the latest refs? I wonder if we should have a `manifest.LatestVersion` type or the like that maintains the current virtual backings and latest refs. All users, including tools, could make use of it to avoid accidentally depending on LatestRef state that isn't maintained by the simple version edit application.
___
*[`internal/manifest/version_edit.go` line 896 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3383#-NsTZbPBCLnWLtme3a7R:-NsTZbPBCLnWLtme3a7S:b4qay90) ([raw file](https://github.com/cockroachdb/pebble/blob/c77952507506534f0174e547025105666bd69fbe/internal/manifest/version_edit.go#L896)):*
> ```Go
> // Invariant: Any file being deleted through ve must belong to the curr Version.
> // We can't have a delete for some arbitrary file which does not exist in curr.
> func AccumulateIncompleteAndApplySingleVE(
> ```

good riddance

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3383)* status: 9 of 11 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`version_set.go` line 489 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3383#-NsTYZvb-nHxkoVHCUF1:-NsTtNBg0tW5tA6cjVSY:b-r3ut74) ([raw file](https://github.com/cockroachdb/pebble/blob/45a55c12738c7907a9422a46032d3a032c590348/version_set.go#L489)):*

I actually had a question about this. I saw a few similar optimizations but moving to a new version in general feels like a very heavy-weight path to me (we do IO, we allocate a bunch in `Accumulate/Apply`, we do btree operations etc)
</blockquote></details>

Good point—it is heavyweight, but we hold `DB.mu` at this point, whereas the IO and accumulate happens once we've dropped `DB.mu` (while still holding the "log lock"). There are probably lots of other allocations while holding `DB.mu` elsewhere, but in general we should try to drive those down since an allocation may potentially need to go to the operating system for more memory. A locked `DB.mu` blocks batches entering the commit pipeline.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3383#-:-NsU9Gw49TYJJ6AGsUOO:b-814g8k)* status: 9 of 11 files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`version_set.go` line 489 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3383#-NsTYZvb-nHxkoVHCUF1:-NsU84rV-EaeG8GmEDwq:bq4l630) ([raw file](https://github.com/cockroachdb/pebble/blob/45a55c12738c7907a9422a46032d3a032c590348/version_set.go#L489)):*

Good point—it is heavyweight, but we hold `DB.mu` at this point, whereas the IO and accumulate happens once we've dropped `DB.mu` (while still holding the "log lock"). There are probably lots of other allocations while holding `DB.mu` elsewhere, but in general we should try to drive those down since an allocation may potentially need to go to the operating system for more memory. A locked `DB.mu` blocks batches entering the commit pipeline.
</blockquote></details>

Ah, good point about the locked sections. I'll keep that in mind. Done.
___
*[`version_set.go` line 494 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3383#-NsTYuhv0w72I9Nku63m:-NsU8jju59MUJ14dpRF8:b1azl86) ([raw file](https://github.com/cockroachdb/pebble/blob/45a55c12738c7907a9422a46032d3a032c590348/version_set.go#L494)):*

Good point, they shouldn't care about latest refs but I will audit all uses.

Heh, yes, I already have something that centralizes all this "latest" state, it will be my next PR: https://github.com/cockroachdb/pebble/commit/f7c3872864aad25d5e1bea7317d51c8dfc8af5b9
</blockquote></details>

So in the old code, the only consequence of the latest refs was whether backings were added to the zombie list. All other uses of `BulkVersionEdit` don't care about zombies and were passing a nil zombie map to `Apply`.
___
*[`internal/manifest/version_edit.go` line 896 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3383#-NsTZbPBCLnWLtme3a7R:-NsU8DI7C4yQYQh-uYOW:b-mulzg) ([raw file](https://github.com/cockroachdb/pebble/blob/c77952507506534f0174e547025105666bd69fbe/internal/manifest/version_edit.go#L896)):*

good riddance
</blockquote></details>

:beers:

PR #3384, opened 2024-03-07
scan_internal: don't visit empty external files

stevendanna:
When truncating external files to our scan bounds, we may find that the file has Start == End. Other validation in ingest asserts these are invalid bounds.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -1179,23 +1179,6 @@ type ExternalFile struct {
 	Level uint8
 }
 
-func (e *ExternalFile) cloneFromFileMeta(f *fileMetadata) {
-	*e = ExternalFile{
-		Bounds: KeyRange{
-			Start: append([]byte(nil), f.Smallest.UserKey...),
-			End:   append([]byte(nil), f.Largest.UserKey...),
-		},
-		HasPointKey: f.HasPointKeys,
-		HasRangeKey: f.HasRangeKeys,
-		Size:        f.Size,
-	}
-	e.SyntheticSuffix = append([]byte(nil), f.SyntheticSuffix...)
-	if pr := f.PrefixReplacement; pr != nil {
-		e.ContentPrefix = append([]byte(nil), pr.ContentPrefix...)
-		e.SyntheticPrefix = append([]byte(nil), pr.SyntheticPrefix...)
-	}
-}
-
 // IngestWithStats does the same as Ingest, and additionally returns
 // IngestOperationStats.
 func (d *DB) IngestWithStats(paths []string) (IngestOperationStats, error) {
--- a/scan_internal.go
+++ b/scan_internal.go
@@ -7,6 +7,7 @@ package pebble
 import (
 	"context"
 	"fmt"
+	"slices"
 
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
@@ -460,27 +461,43 @@ func (d *DB) truncateExternalFile(
 	level int,
 	file *fileMetadata,
 	objMeta objstorage.ObjectMetadata,
-) *ExternalFile {
+) (*ExternalFile, error) {
 	cmp := d.cmp
-	sst := &ExternalFile{}
-	sst.cloneFromFileMeta(file)
-	sst.Level = uint8(level)
-	sst.ObjName = objMeta.Remote.CustomObjectName
-	sst.Locator = objMeta.Remote.Locator
+	sst := &ExternalFile{
+		Level:           uint8(level),
+		ObjName:         objMeta.Remote.CustomObjectName,
+		Locator:         objMeta.Remote.Locator,
+		Bounds:          KeyRange{},
+		HasPointKey:     file.HasPointKeys,
+		HasRangeKey:     file.HasRangeKeys,
+		Size:            file.Size,
+		SyntheticSuffix: slices.Clone(file.SyntheticSuffix),
+	}
+	if pr := file.PrefixReplacement; pr != nil {
+		sst.ContentPrefix = slices.Clone(pr.ContentPrefix)
+		sst.SyntheticPrefix = slices.Clone(pr.SyntheticPrefix)
+	}
+
 	needsLowerTruncate := cmp(lower, file.Smallest.UserKey) > 0
-	needsUpperTruncate := cmp(upper, file.Largest.UserKey) < 0 || (cmp(upper, file.Largest.UserKey) == 0 && !file.Largest.IsExclusiveSentinel())
-
 	if needsLowerTruncate {
-		sst.Bounds.Start = sst.Bounds.Start[:0]
-		sst.Bounds.Start = append(sst.Bounds.Start, lower...)
+		sst.Bounds.Start = slices.Clone(lower)
+	} else {
+		sst.Bounds.Start = slices.Clone(file.Smallest.UserKey)
 	}
 
+	cmpUpper := cmp(upper, file.Largest.UserKey)
+	needsUpperTruncate := cmpUpper < 0 || (cmpUpper == 0 && !file.Largest.IsExclusiveSentinel())
 	if needsUpperTruncate {
-		sst.Bounds.End = sst.Bounds.End[:0]
-		sst.Bounds.End = append(sst.Bounds.End, upper...)
+		sst.Bounds.End = slices.Clone(upper)
+	} else {
+		sst.Bounds.End = slices.Clone(file.Largest.UserKey)
 	}
 
-	return sst
+	if cmp(sst.Bounds.Start, sst.Bounds.End) >= 0 {
+		return nil, errors.AssertionFailedf("pebble: invalid external file bounds after truncation [%q, %q)", sst.Bounds.Start, sst.Bounds.End)
+	}
+
+	return sst, nil
 }
 
 // truncateSharedFile truncates a shared file's [Smallest, Largest] fields to
@@ -693,6 +710,10 @@ func scanInternalImpl(
 		for level := firstLevelWithRemote; level < numLevels; level++ {
 			files := current.Levels[level].Iter()
 			for f := files.SeekGE(cmp, lower); f != nil && cmp(f.Smallest.UserKey, upper) < 0; f = files.Next() {
+				if cmp(lower, f.Largest.UserKey) == 0 && f.Largest.IsExclusiveSentinel() {
+					continue
+				}
+
 				var objMeta objstorage.ObjectMetadata
 				var err error
 				objMeta, err = provider.Lookup(fileTypeTable, f.FileBacking.DiskFileNum)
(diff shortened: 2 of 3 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3384)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @stevendanna)

___
*[`scan_internal.go` line 699 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3384#-NsOO6g26J4ti6axBUsV:-NsOO6g26J4ti6axBUsW:bc53303) ([raw file](https://github.com/cockroachdb/pebble/blob/e40ae1915a933252f796ef8c3a986c853004d8b1/scan_internal.go#L699)):*
> ```Go
> 		for level := firstLevelWithRemote; level < numLevels; level++ {
> 			files := current.Levels[level].Iter()
> 			for f := files.SeekGE(cmp, lower); f != nil && cmp(f.Smallest.UserKey, upper) < 0; f = files.Next() {
> ```

Shouldn't this guarantee the empty sst doesn't happen? Maybe it's just a corner case where `SeekGE` returns a file that ends exactly at lower? Feels like we should check that here and `continue` instead (and replace the check you added with an assertion)

stevendanna:
@RaduBerinde Yes, I believe that is exactly what is happening. The file ends exactly at lower and SeekGE returns it. We try to scan from something like `/Table/104/, Table/105` and end up picking up a file for `/Table/103,/Table/104`.

RaduBerinde (commented):
Then we should check for that condition first thing in the loop and `continue` instead. E.g.
```
		// SeekGE can return a file that ends at lower.
		if f.Largest.IsExclusiveSentinel() && cmp(f.Largest.UserKey, lower) == 0 {
			continue
		}
```

And we can change the new check you added into an assertion.

@jbowens Is this actually a bug in `LevelIterator.SeekGE`? Should it do what's below instead?
```diff
        m := i.seek(func(m *FileMetadata) bool {
-               return cmp(m.Largest.UserKey, userKey) >= 0
+               c := cmp(m.Largest.UserKey, userKey)
+               return c > 0 || (c == 0 && !m.Largest.IsExclusiveSentinel())
```

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3384#-:-NsTZrqaEGPumUeqgEVt:b-2mwggn)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @stevendanna)

___
*[`scan_internal.go` line 466 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3384#-NsTaPmW8h9LpMTsYQ3t:-NsTaPmW8h9LpMTsYQ3u:bbo4kx8) ([raw file](https://github.com/cockroachdb/pebble/blob/e40ae1915a933252f796ef8c3a986c853004d8b1/scan_internal.go#L466)):*
> ```Go
> 	cmp := d.cmp
> 	sst := &ExternalFile{}
> 	sst.cloneFromFileMeta(file)
> ```

NYC but this method is only used here and it's far away from this code. It just makes this code harder to read. I would just inline it and simplify the code (we can set most fields directly and then we can set Start/End to either key).

stevendanna (commented):
:+1: Will push this shortly. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3384#-:-NsTeyAr8qoqPo0WMf3l:bip0whe)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`scan_internal.go` line 466 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3384#-NsTaPmW8h9LpMTsYQ3t:-NsTe1Pn4swcwvNtS_ug:b-soj7ki) ([raw file](https://github.com/cockroachdb/pebble/blob/e40ae1915a933252f796ef8c3a986c853004d8b1/scan_internal.go#L466)):*

NYC but this method is only used here and it's far away from this code. It just makes this code harder to read. I would just inline it and simplify the code (we can set most fields directly and then we can set Start/End to either key).
</blockquote></details>

:D You are right. I put this in in a previous commit wanting symmetry with SharedSST.cloneFileFromMeta but I also dislike how far away it is.

jbowens (commented):
I don't think it's a bug, but maybe adjusting the semantics can clean up some code. We'd need to audit all the other callers. Today this SeekGE is only a seek on the largest user key and doesn't imply anything about inclusivity or exclusivity.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3384#-:-NsTfGH5EvROvIOCONkV:b-xp9vso)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @itsbilal, @RaduBerinde, and @stevendanna)

jbowens (commented):
Ugh, it looks like when the key type filtering was added, it implemented the semantics you're suggesting when the filter is not `KeyTypePointAndRange`?

```
	if i.filter != KeyTypePointAndRange && m != nil {
		b, ok := m.LargestBound(i.filter)
		if !ok {
			m = i.Next()
		} else if c := cmp(b.UserKey, userKey); c < 0 || c == 0 && b.IsExclusiveSentinel() {
			// This file does not contain any keys of the type ≥ lower. It
			// should be filtered, even though it does contain point keys.
			m = i.Next()
		}
	}
```

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3384#-:-NsTgfBi7sEuEmSOwI1M:b-isgtid)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @itsbilal, @RaduBerinde, and @stevendanna)

stevendanna (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3384)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`scan_internal.go` line 471 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3384#-NsTyHnE2fe8tT1NhmAC:-NsTyHnE2fe8tT1NhmAD:b-v35pwx) ([raw file](https://github.com/cockroachdb/pebble/blob/e40ae1915a933252f796ef8c3a986c853004d8b1/scan_internal.go#L471)):*
> ```Go
> 	sst.Locator = objMeta.Remote.Locator
> 	needsLowerTruncate := cmp(lower, file.Smallest.UserKey) > 0
> 	needsUpperTruncate := cmp(upper, file.Largest.UserKey) < 0 || (cmp(upper, file.Largest.UserKey) == 0 && !file.Largest.IsExclusiveSentinel())
> ```

Actually, isn't this wrong as well. This `!` seems wrong.

stevendanna (commented):
Implemented @RaduBerinde's suggestion. I wasn't sure whether y'all prefer error returns with AssertionFailedf or a panic for this sort of thing.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3384#-:-NsU2gpZ4qzgH6psQSlf:bdcswlv)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3384)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @itsbilal and @stevendanna)

___
*[`scan_internal.go` line 471 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3384#-NsTyHnE2fe8tT1NhmAC:-NsU5PgAEcF4t-ZK51TQ:b-m37y4j) ([raw file](https://github.com/cockroachdb/pebble/blob/e40ae1915a933252f796ef8c3a986c853004d8b1/scan_internal.go#L471)):*
<details><summary><i>Previously, stevendanna (Steven Danna) wrote…</i></summary><blockquote>

Actually, isn't this wrong as well. This `!` seems wrong.
</blockquote></details>

Looks like it's wrong but inconsequential.. The user key is the same so it doesn't matter which one we choose.
___
*[`scan_internal.go` line 473 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3384#-NsU5cfvB8dPiQU32kJR:-NsU5cfvB8dPiQU32kJS:b-i06nna) ([raw file](https://github.com/cockroachdb/pebble/blob/a7e634a317f29e2a6461de23a7fd9479b4ab90ec/scan_internal.go#L473)):*
> ```Go
> 		HasRangeKey:     file.HasRangeKeys,
> 		Size:            file.Size,
> 		SyntheticSuffix: append([]byte(nil), file.SyntheticSuffix...),
> ```

[nit] use slices.Clone (same for the bound keys)
___
*[`scan_internal.go` line 495 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3384#-NsU6bQVEtFrkwsG6zaJ:-NsU6bQVEtFrkwsG6zaK:bb6og5y) ([raw file](https://github.com/cockroachdb/pebble/blob/a7e634a317f29e2a6461de23a7fd9479b4ab90ec/scan_internal.go#L495)):*
> ```Go
> 	}
> 
> 	if cmp(sst.Bounds.Start, sst.Bounds.End) == 0 {
> ```

<= 0

stevendanna (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3384)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`scan_internal.go` line 473 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3384#-NsU5cfvB8dPiQU32kJR:-NsUCLJ27f1DuYFCeLeK:b-66zb49) ([raw file](https://github.com/cockroachdb/pebble/blob/a7e634a317f29e2a6461de23a7fd9479b4ab90ec/scan_internal.go#L473)):*

[nit] use slices.Clone (same for the bound keys)
</blockquote></details>

Nice, I really need to internalize the new slices package. makes things much better.
___
*[`scan_internal.go` line 495 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3384#-NsU6bQVEtFrkwsG6zaJ:-NsUCSZv0CR5J3Clz311:bs0lv1c) ([raw file](https://github.com/cockroachdb/pebble/blob/a7e634a317f29e2a6461de23a7fd9479b4ab90ec/scan_internal.go#L495)):*

<= 0
</blockquote></details>

I think you meant `>=`. Also changed the message a bit to line up with some others I found.

RaduBerinde (approved):
:lgtm:

Good call on returning it as an error, it is preferable (we'd rather have an operation fail than crash the node).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3384#-:-NsUCYyG6zaLOfdZtffE:b-b7xqa9)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @stevendanna)

PR #3387, opened 2024-03-10
manifest: use Lx for levels in version String/DebugString

RaduBerinde:
The version string shows levels as `1:` instead of `L1:` which is how
we usually display levels.

This change prepends the `L` and updates tests to expect it. This
makes it more clear what you are looking at and also makes the parsing
(`ParseVersionDebug`) less error-prone, as `1:` could also mean file
number 1.


RaduBerinde (commented):
The actual change is in `version,.go`, the rest are all mechanical updates.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3387#-:-Nsd0yEeB3Dh_swiy7Dk:b-g55q0o)* status: 0 of 56 files reviewed, all discussions resolved (waiting on @itsbilal and @jbowens)

RaduBerinde:
TFTR!

PR #3389, opened 2024-03-10
internal/manifest: extend debug string output with externalness

dt:


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3389)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @dt and @itsbilal)

___
*[`internal/manifest/version.go` line 794 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3389#-NsedPs24mWpzYvhMhz-:-NsedPs24mWpzYvhMhz0:b-ov10zn) ([raw file](https://github.com/cockroachdb/pebble/blob/842c86fadc2d13e337a83643f389500f4793eb94/internal/manifest/version.go#L794)):*
> ```Go
> // use in tests and debugging, returning the file number and the point, range
> // and overall bounds for the table.
> func (m *FileMetadata) DebugString(format base.FormatKey, verbose bool, ext IsExternal) string {
> ```

I don't like this very much, since "externality" is not something this layer deals with and especially since many code paths will pass `nil` and the output will be inconsistent.

I would just add a helper in pebble that calls `DebugString()` and appends an ` [external]` at the end as needed.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3389)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`internal/manifest/version.go` line 820 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3389#-NseeaQz5lOe9TP5boO8:-NseeaQz5lOe9TP5boO9:bwu8cqz) ([raw file](https://github.com/cockroachdb/pebble/blob/51e0471b72231bbb8a7b6255a0bf5ef7782d0d4e/internal/manifest/version.go#L820)):*
> ```Go
> 			m.SmallestRangeKey.Pretty(format), m.LargestRangeKey.Pretty(format))
> 	}
> 	if m.PrefixReplacement != nil {
> ```

Second commit LGTM.

Maybe handle the `ContentPrefix == ""` case separately and print it as `prefix:%s` (hopefully that will be the only possible case soon).

dt:
> since "externality" is not something this layer deals with

Could add something like `type FileDescriber interface { Describe(base.DiskFileNum) fmt.Formatter }` and then let when we have one we'd just delegate to our FileDescriber to print our diskFileNum?

RaduBerinde:
I still think it's cleaner for the upper layer to append whatever it needs. Note that we have a ParseFileMetadataDebug which tries to parse the output, it's nice to have a well defined output in this layer.

RaduBerinde:
Or feel free to add a separate DebugStringEx or DebugStringCustom or something which takes a backing formatting func (I would pass it as a func not as an interface, just like the key formatter).

RaduBerinde:
Well, we already have a separate function string. We can just rename to CustomString.

I think the way I would do it is to add a `extraBackingInfo(b *FileBacking) []string` argument and join the result with `DiskFileNum.String()` with commas. E.g. (00001,external) and in the future can also be (00001,shared) or just (shared) for physical ssts.

PR #3379, opened 2024-03-06
db: verify virtualBackings in invariants mode

RaduBerinde:
#### db: rename fileBackings to virtualBackings

Rename to clarify that this includes only backings for virtual
sstables.

#### db: verify virtualBackings in invariants mode

And fix the copy compaction code to create the backing in the
manifest.

PR #3390, opened 2024-03-10
ingest_test: support reopening engine in test

dt:
First commit is #3388

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -1286,7 +1286,9 @@ func runIngestCmd(td *datadriven.TestData, d *DB, fs vfs.FS) error {
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


dt:
This is basically a regression test for https://github.com/cockroachdb/cockroach/issues/120181 which it turns out was already fixed by https://github.com/cockroachdb/pebble/pull/3379 (I was confused as to why the test here was passing when I could reproduce the behavior so easily with CRDB, but that's because CRDB hasn't yet picked up those changes, until  https://github.com/cockroachdb/cockroach/pull/120025 merges). Now that I know this test is testing functionality also tested in invariants, I could also just abandon it. That said, I did verify that reverting `c77952507` does cause it to fail so it is working.

itsbilal (approved):
I'm good with a test that does an explicit reopen too. LGTM

PR #3392, opened 2024-03-11
manifest: improve VersionEdit tests

RaduBerinde:
Some background: the goal of these changes is to address [a versionSet test TODO](https://github.com/cockroachdb/pebble/blob/master/version_set_test.go#L216) as I am reworking where and how the "latest ref"s are maintained.

#### manifest: add a helper for DebugString parsing

We add a parsing helper and use it to simplify the
`ParseFileMetadataDebug` and `ParseVersionDebug` code. This will make
it easier to extend these functions (e.g. to support virtual
sstables).

#### manifest: improve VersionEdit tests

In this commit we add a `ParseVersionEditDebug` which can create a
`VersionEdit` from a `DebugString` output (similar to
`ParseFileMetadataDebug` and `ParseVersionDebug`).

We improve `TestVersionEditApply` to use the `DebugString` formats
instead of a one-off format.

### DIFF
--- a/internal/manifest/testdata/version_edit_apply
+++ b/internal/manifest/testdata/version_edit_apply
@@ -1,180 +1,171 @@
-apply
- L0
-  1:[a#1,SET-b#2,SET]
-  2:[c#3,SET-d#4,SET]
-edit
- delete
-  L0
-   1
- add
-  L2
-  1:[a#1,SET-b#2,SET]
-  4:[c#3,SET-d#4,SET]
+define v1
+L0:
+  000002:[c#3,SET-d#4,SET] seqnums:[3-4]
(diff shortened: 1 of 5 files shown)


jbowens (approved):
:lgtm_strong: 

> Reviewed 2 of 2 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3392#-:-NsibuOD7lDxCEl33jR8:b-uwkowd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sumeerbhola)

RaduBerinde:
Thanks for the quick review!

PR #3393, opened 2024-03-11
manifest: add VersionEdit tests with virtual tables

RaduBerinde:
#### manifest: improve VersionEdit stringification

This makes the format of the fields more consistent.

#### manifest: add VersionEdit tests with virtual tables

Improve the debug parsing methods to parse virtual tables and backing
operations and add some `version_edit_apply` tests with virtual
tables.

### DIFF
--- a/internal/manifest/testdata/version_edit_apply
+++ b/internal/manifest/testdata/version_edit_apply
@@ -169,3 +169,49 @@ new version edit
 ----
 L2:
   000005:[s#3,SET-z#4,SET] seqnums:[0-0] points:[s#3,SET-z#4,SET]
+
+define v6
+L1:
+  000001:[a#2,SET-e#2,SET]
+L2:
+  000002:[c#1,SET-f#1,SET]
+----
+L1:
+  000001:[a#2,SET-e#2,SET] seqnums:[0-0] points:[a#2,SET-e#2,SET]
+L2:
+  000002:[c#1,SET-f#1,SET] seqnums:[0-0] points:[c#1,SET-f#1,SET]
+
+# Convert a physical table to virtual tables.
+apply v6
+  del-table: L1 000001
+  add-table: L1 000003(000009):[a#2,SET-b#2,SET]
+  add-table: L1 000004(000009):[c#2,SET-e#2,SET]
+  add-backing: 000009
+----
+L1:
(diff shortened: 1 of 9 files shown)


jbowens (approved):
:lgtm:

> Reviewed 6 of 6 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3393#-:-Nsj34uw8vepqt3J6yeu:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/manifest/version.go` line 880 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3393#-Nsj1sKg3lvGbaU5CItp:-Nsj1sKg3lvGbaU5CItq:b-yap8gb) ([raw file](https://github.com/cockroachdb/pebble/blob/838eb825cf1b8a3e25f48efbe2c972057bfbbf1b/internal/manifest/version.go#L880)):*
> ```Go
> 		size := uint64(backingNum) * 10000
> 		m.InitProviderBacking(backingNum, size)
> 	}
> ```

I could imagine this catching someone by surprise since it's not information actually contained within the input. maybe we could make it opt-in by making this function a method on the debug parser and adding a field that enables deterministically synthesizing state not explicitly contained in the input?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3393#-:-Nsj8Ep_1CiGEQRS01Ip:b-814g8k)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`internal/manifest/version.go` line 880 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3393#-Nsj1sKg3lvGbaU5CItp:-Nsj7t1hANGRnDuTPB2C:b-v9b92b) ([raw file](https://github.com/cockroachdb/pebble/blob/838eb825cf1b8a3e25f48efbe2c972057bfbbf1b/internal/manifest/version.go#L880)):*

I could imagine this catching someone by surprise since it's not information actually contained within the input. maybe we could make it opt-in by making this function a method on the debug parser and adding a field that enables deterministically synthesizing state not explicitly contained in the input?
</blockquote></details>

Yeah, good point. There's no reason this has to happen here, the caller can just fill in the sizes as needed. Removed.

PR #3394, opened 2024-03-12
manifest: centralize virtual backing metadata

RaduBerinde:
#### manifest: move Ref/Unref to FileBacking

The refcounting of the `FileBacking` unnecessarily goes through
`FileMetadata` which just makes things more obscure.

#### manifest: centralize virtual backing metadata

Currently, we have several pieces of metadata related to virtual
backings in different places:
 - latestRef count in the backing
 - VirtualizedSize in the backing
 - set of virtual backings in versionSet

The first two are atomics, even though they are only used under the
log lock.

This commit moves them all into a single `VirtualBackings` structure.
We no longer maintain latestRef counts for physical table backings
(they are not necessary).


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3394)* status: 0 of 18 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`internal/manifest/virtual_backings.go` line 111 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3394#-NskqqFL1ftTiFeSDgYd:-NskqqFL1ftTiFeSDgYe:by6vapg) ([raw file](https://github.com/cockroachdb/pebble/blob/910754cee6a61787344e73734f3f3e6376c055d3/internal/manifest/virtual_backings.go#L111)):*
> ```Go
> // space amplification due to virtual sstables is high, and we should pick the
> // virtual sstable with a higher priority.
> func (bv *VirtualBackings) Usage(n base.DiskFileNum) (useCount int, virtualizedSize uint64) {
> ```

A reasonable question is why do we have this API instead of having `RemoveTable` return a flag indicating if it was the last user. The answer is that we will add a `Protect/Unprotect` API that temporarily prevents a backing from being removed from the latest version (in order to be reused during an ingestion). The `Unprotect` call can happen in contexts which are not prepared to handle the "last ref" case. `Unused()` allows us to handle the unused backings only when we are preparing a new version.

An extra bonus is that this API gives us (for free) a way to remove any leaked backings if we crash while some backings are protected - the first version update after Open will automatically remove them because they will be Unused.

sumeerbhola (commented):
Starting to look at this, and went back to look at https://github.com/cockroachdb/pebble/pull/3383. I think that PR is an improvement, but some historical context may be useful, since it could motivate writing some more code comments.

- Our ref counting, prior to introduction of virtual ssts, was all done at the time of BVE.Apply, since that is when a new version is being produced. It was also all hidden inside the btrees, so it didn't matter whether the new version actually was logged to the manifest or not (but it is, since a failure to write to the manifest causes a fail-stop failure).

- When adding virtual ssts, we needed a place to hold the backing files, and do so in an efficient manner. We could have written a copy-on-write data-structure, but since a map was good enough, *and* we only needed it for the latest version, we put that map in the `VersionSet` (FileBackings).

- And we needed ref counting from the latest version (latestVersionRefs), since concurrent compactions could be consuming virtual ssts that pointed to the same backing, and they could complete in any order. So who should modify these latest refs. We went with the "principle" that it was reasonable to continue to do so in a BVE "Apply" method. We also noted that unlike other VersionEdits (loaded from the manifest in open), this VersionEdit was incomplete, in that it didn't contain `RemovedBackingTables`. So `AccumulateIncompleteAndApplySingleVE` was born.

- I think the recent PR correctly notes that the code is not easy to understand, and lifts the job of completing the VersionEdit and managing the `latestVersionRefs` into `versionSet.logAndApply`, *before* it hands the complete `VersionEdit` to BVE.Accumulate and BVE.Apply. This simplifies the code.

Some suggestions for code comments:
- a code comment in the `logAndApply` declaration that `VersionEdit` needs to be completed (this was an oversight from before too).
- a code comment in `logAndApply` that the updating of `FileBacking` and latest refs before applying to the latest version and logging to the manifest file is ok since if any of those subsequent steps fail, we will fail stop.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3394#-:-NsnJUwS-Gdr2X5uDoGM:b-bav0sd)* status: 0 of 18 files reviewed, 1 unresolved discussion (waiting on @jbowens)

sumeerbhola (approved):
:lgtm_strong: 

great to have the additional abstraction and overall code simplification.

> Reviewed 5 of 7 files at r1, 6 of 12 files at r3, 12 of 12 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3394#-:-NsnxFeXBpyX7NsArUuU:b-u3acvd)* status: 16 of 18 files reviewed, 5 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/manifest/virtual_backings.go` line 43 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3394#-NsnsR4v5Nx7oH90DFN_:-NsnsR4v5Nx7oH90DFNa:bwp84u7) ([raw file](https://github.com/cockroachdb/pebble/blob/35df508c61c52fc83ef6fff322c2efcfd6f860f1/internal/manifest/virtual_backings.go#L43)):*
> ```Go
> 	virtualizedSize uint64
> }
> ```

 A comment describing the lifecycle of a backing file would be useful. That describes it is added with useCount 0, then can see increments and decrements in useCount (via *Table methods), and may exist in the manifest even if useCount is 0, etc.
___
*[`internal/manifest/virtual_backings.go` line 50 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3394#-NsnrgrNDH9zXQKMfLYb:-NsnrgrNDH9zXQKMfLYc:b-r1aqtl) ([raw file](https://github.com/cockroachdb/pebble/blob/35df508c61c52fc83ef6fff322c2efcfd6f860f1/internal/manifest/virtual_backings.go#L50)):*
> ```Go
> 		backing: backing,
> 	})
> 	bv.unused[backing] = struct{}{}
> ```

I suppose the `useCount` starts at 0 because it may exist in the manifest without any virtual sst pointing to it, because of the protect case you mentioned?
___
*[`internal/manifest/virtual_backings.go` line 55 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3394#-Nsns4YT6IqNIT2ly920:-Nsns4YT6IqNIT2ly921:buuxwt9) ([raw file](https://github.com/cockroachdb/pebble/blob/35df508c61c52fc83ef6fff322c2efcfd6f860f1/internal/manifest/virtual_backings.go#L55)):*
> ```Go
> 
> // Remove a backing. The backing must not be in use.
> func (bv *VirtualBackings) Remove(n base.DiskFileNum) {
> ```

So `Remove` is called after `RemoveTable` tells the caller that the `useCount` is 0?
___
*[`testdata/version_set` line 1 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3394#-NsnwLK-Aa1c1NpYRf_s:-NsnwLK-Aa1c1NpYRf_t:b-j6rd9r) ([raw file](https://github.com/cockroachdb/pebble/blob/35df508c61c52fc83ef6fff322c2efcfd6f860f1/testdata/version_set#L1)):*
> ```
> apply
> ```

nice test!

RaduBerinde (commented):
TFTR! The write-up above is very helpful. I actually considered making a copy-on-write structure just so it fits better with`Version` but it didn't feel worth it. I added more comments as suggested, let me know if you have more suggestions.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3394#-:-NspXQq-CL839BDPL43h:bsp6y1u)* status: 4 of 18 files reviewed, 5 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/manifest/virtual_backings.go` line 43 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3394#-NsnsR4v5Nx7oH90DFN_:-NspX79hEUI1RdB5Eik5:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/35df508c61c52fc83ef6fff322c2efcfd6f860f1/internal/manifest/virtual_backings.go#L43)):*

A comment describing the lifecycle of a backing file would be useful. That describes it is added with useCount 0, then can see increments and decrements in useCount (via *Table methods), and may exist in the manifest even if useCount is 0, etc.
</blockquote></details>

Done.
___
*[`internal/manifest/virtual_backings.go` line 50 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3394#-NsnrgrNDH9zXQKMfLYb:-NspVwP6Ccc968uIVqaR:byudq3e) ([raw file](https://github.com/cockroachdb/pebble/blob/35df508c61c52fc83ef6fff322c2efcfd6f860f1/internal/manifest/virtual_backings.go#L50)):*

I suppose the `useCount` starts at 0 because it may exist in the manifest without any virtual sst pointing to it, because of the protect case you mentioned?
</blockquote></details>

Also because when we Open we get a list of backings and their sizes, separately from any tables.
___
*[`internal/manifest/virtual_backings.go` line 55 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3394#-Nsns4YT6IqNIT2ly920:-NspWxW68nJfimpsbpjd:b54olx4) ([raw file](https://github.com/cockroachdb/pebble/blob/35df508c61c52fc83ef6fff322c2efcfd6f860f1/internal/manifest/virtual_backings.go#L55)):*

So `Remove` is called after `RemoveTable` tells the caller that the `useCount` is 0?
</blockquote></details>

Added a comment - RemoveTable doesn't tell you anything, only `Unused()` does.
___
*[`testdata/version_set` line 1 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3394#-NsnwLK-Aa1c1NpYRf_s:-NspXKHX3I5FkhglVm79:bf7ccpi) ([raw file](https://github.com/cockroachdb/pebble/blob/35df508c61c52fc83ef6fff322c2efcfd6f860f1/testdata/version_set#L1)):*

nice test!
</blockquote></details>

Thanks! It took a bit of work to get all the parsing together but it was worth it.

sumeerbhola (approved):
:lgtm_strong: 

> Reviewed 2 of 7 files at r1, 12 of 12 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3394#-:-Nsrzm8A2dVJHOYnZjTy:b-uwkowd)* status: all files reviewed, 7 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`version_set.go` line 383 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3394#-Nsryp6I5BBJDStLw7R3:-Nsryp6I5BBJDStLw7R4:bibljt4) ([raw file](https://github.com/cockroachdb/pebble/blob/206df58015169879fa30a16af7ba9eb193fcafa3/version_set.go#L383)):*
> ```Go
> // to the current version, and installs the new version.
> //
> // logAndApply fills in the NextFileNum and LastSeqNum fields; other than these,
> ```

and RemovedBackingTables field
___
*[`version_set.go` line 389 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3394#-Nsrz-LJC4ASJknX5sA5:-Nsrz-LJC4ASJknX5sA6:bkiq5zy) ([raw file](https://github.com/cockroachdb/pebble/blob/206df58015169879fa30a16af7ba9eb193fcafa3/version_set.go#L389)):*
> ```Go
> // BulkVersionEdit.Apply). The state of the virtual backings
> // (vs.virtualBackings) is not updated; it is expected that the caller does that
> // before calling logAndApply. Doing that before logging to the manifest and
> ```

it is updated in logAndApply, before calling BulkVersionEdit.Apply.

This must have been a typo and meant to say BulkVersionEdit.Apply instead of logAndApply.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3394)* status: all files reviewed, 7 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`version_set.go` line 383 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3394#-Nsryp6I5BBJDStLw7R3:-Nss7WOJ4Z8SqFfP7ICx:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/206df58015169879fa30a16af7ba9eb193fcafa3/version_set.go#L383)):*

and RemovedBackingTables field
</blockquote></details>

Done.
___
*[`version_set.go` line 389 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3394#-Nsrz-LJC4ASJknX5sA5:-Nss5OIE7BH5ZNhNvoO8:bcjmo48) ([raw file](https://github.com/cockroachdb/pebble/blob/206df58015169879fa30a16af7ba9eb193fcafa3/version_set.go#L389)):*

it is updated in logAndApply, before calling BulkVersionEdit.Apply.

This must have been a typo and meant to say BulkVersionEdit.Apply instead of logAndApply.
</blockquote></details>

Yeah I confused myself, I reworked the comment.

PR #3395, opened 2024-03-12
tool: add DirectoryLock option

jbowens:
Add a new DirectoryLock option so that users of the tool package's programmatic interface may pass an already-acquired directory lock to Pebble. In CockroachDB, we acquire the directory lock before opening Open in order to prevent concurrent modification of the encryption-at-rest state.

PR #3397, opened 2024-03-12
tool: replace DirectoryLock with OpenOptions

jbowens:
In #3395 I added a new tool.DirectoryLock option for passing an already-acquired directory lock. This was motivated by cockroachdb/cockroach#120156 which begins to acquire the lock before loading encryption-at-rest state. When adding that option, I forgot the nuance that the tool doesn't know what database you'll open at the time the tool is initialized. When a node has multiple stores, the operator might try to open any of them. CockroachDB will lazily initialize the encryptedFS for a store only once the tool attempts to access a path within the store.

This commit removes the DirectoryLock option and adds a new OpenOptions option that allows the caller to pass in arbitrary OpenOptions that are run when constructing the pebble.Options to use. CockroachDB will be able to implement an OpenOption that opens the relevant encryptedFS when invoked and settings Options.Lock directly.

### DIFF
--- a/tool/db.go
+++ b/tool/db.go
@@ -44,6 +44,7 @@ type dbT struct {
 	comparers       sstable.Comparers
 	mergers         sstable.Mergers
 	openErrEnhancer func(error) error
+	openOptions     []OpenOption
 
 	// Flags.
 	comparerName  string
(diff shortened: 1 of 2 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3397#-:-NsoZ0-BASw4VuqE_OCR:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aadityasondhi)

jbowens:
TFTR!

PR #3398, opened 2024-03-13
ingest,compaction: use excise when flushing flushableIngest

### LINKED ISSUE #3335: ingest: `IngestAndExcise` should use `flushableIngest` for ingests
This came up in an [internal thread](https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1708457620647379).

When we use `IngestAndExcise` as default and have an overlap with the memtable, we have to wait for the memtable to flush since we don't use flushable ingest. This is because for the general `IngestAndExcise` use case, the sstables are from another LSM and do not have a RANGEDEL.

A possible solution sketch (from the discussion):
> Option to pass to IngestAndExcise promising that the exciseSpan is also repeated explicitly via RANGEDELs in the sstables. Then we could layer these on top as flushable ingests, and when it came time to actually flush this flushable ingest, we would do the excise, and get the lower read amp.

aadityasondhi:
This patch amends the ingest pipeline to allow for certain cases to use `flushaleIngest` in order to avoid waiting on memtable flushes when there is overlap. The current use for this is range snapshots in Cockroach where we include `RANGEDEL`. When we use this path, we also pass the `exciseSpan` down down into the flushable to then excise at flush time to still benefit from the use of excise in the ingest path.

Fixes #3335.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1448,17 +1448,70 @@ func (d *DB) runIngestFlush(c *compaction) (*manifest.VersionEdit, error) {
 	var err error
 	var fileToSplit *fileMetadata
 	var ingestSplitFiles []ingestSplitFile
-	for _, file := range c.flushing[0].flushable.(*ingestedFlushable).files {
-		suggestSplit := d.opts.Experimental.IngestSplit != nil && d.opts.Experimental.IngestSplit() &&
-			d.FormatMajorVersion() >= FormatVirtualSSTables
-		level, fileToSplit, err = ingestTargetLevel(
-			d.newIters, d.tableNewRangeKeyIter, iterOpts, d.opts.Comparer,
-			c.version, baseLevel, d.mu.compact.inProgress, file.FileMetadata,
-			suggestSplit,
-		)
+	ingestFlushable := c.flushing[0].flushable.(*ingestedFlushable)
+
+	updateLevelMetricsOnExcise := func(m *fileMetadata, level int, added []newFileEntry) {
+		levelMetrics := c.metrics[level]
+		if levelMetrics == nil {
+			levelMetrics = &LevelMetrics{}
+			c.metrics[level] = levelMetrics
+		}
+		levelMetrics.NumFiles--
+		levelMetrics.Size -= int64(m.Size)
+		for i := range added {
+			levelMetrics.NumFiles++
+			levelMetrics.Size += int64(added[i].Meta.Size)
+		}
+	}
+
+	suggestSplit := d.opts.Experimental.IngestSplit != nil && d.opts.Experimental.IngestSplit() &&
+		d.FormatMajorVersion() >= FormatVirtualSSTables
+
+	replacedFiles := make(map[base.FileNum][]newFileEntry)
+	for _, file := range ingestFlushable.files {
+		// This file fits perfectly within the excise span, so we can slot it at L6.
+		if ingestFlushable.exciseSpan.Valid() &&
+			ingestFlushable.exciseSpan.Contains(d.cmp, file.FileMetadata.Smallest) &&
+			ingestFlushable.exciseSpan.Contains(d.cmp, file.FileMetadata.Largest) {
+			level = 6
+		} else {
+			level, fileToSplit, err = ingestTargetLevel(
+				d.newIters, d.tableNewRangeKeyIter, iterOpts, d.opts.Comparer,
+				c.version, baseLevel, d.mu.compact.inProgress, file.FileMetadata,
+				suggestSplit,
+			)
+		}
+
+		if ingestFlushable.exciseSpan.Valid() {
+			ve.DeletedFiles = map[manifest.DeletedFileEntry]*manifest.FileMetadata{}
+			// Iterate through all levels and find files that intersect with exciseSpan.
+			for level = range c.version.Levels {
+				overlaps := c.version.Overlaps(level, base.UserKeyBoundsEndExclusive(ingestFlushable.exciseSpan.Start, ingestFlushable.exciseSpan.End))
+				iter := overlaps.Iter()
+
+				for m := iter.First(); m != nil; m = iter.Next() {
+					newFiles, err := d.excise(ingestFlushable.exciseSpan, m, ve, level)
+					if err != nil {
+						return nil, err
+					}
+
+					if _, ok := ve.DeletedFiles[deletedFileEntry{
+						Level:   level,
+						FileNum: m.FileNum,
+					}]; !ok {
+						// We did not excise this file.
+						continue
+					}
+					replacedFiles[m.FileNum] = newFiles
+					updateLevelMetricsOnExcise(m, level, newFiles)
+				}
+			}
+		}
 		if err != nil {
 			return nil, err
 		}
+
+		// Add the current flushableIngest file to the version.
 		ve.NewFiles = append(ve.NewFiles, newFileEntry{Level: level, Meta: file.FileMetadata})
 		if fileToSplit != nil {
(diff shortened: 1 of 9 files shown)


aadityasondhi:
First pass PR. Welcoming feedback, especially on things marked with TODO. Tests and comments are missing.

data_test.go
```diff
@@ -1268,7 +1268,7 @@ func runIngestAndExciseCmd(td *datadriven.TestData, d *DB, fs vfs.FS) error {
 		}
 	}
 
-	if _, err := d.IngestAndExcise(paths, nil /* shared */, nil /* external */, exciseSpan); err != nil {
+	if _, err := d.IngestAndExcise(paths, nil, nil, exciseSpan, false); err != nil {
```
itsbilal, line 1271: nit: let's retain the /* shared */ and /* external */

compaction.go
```diff
@@ -1447,17 +1447,66 @@ func (d *DB) runIngestFlush(c *compaction) (*manifest.VersionEdit, error) {
...
+		}
+	}
+
+	suggestSplit := d.opts.Experimental.IngestSplit != nil && d.opts.Experimental.IngestSplit() &&
+		d.FormatMajorVersion() >= FormatVirtualSSTables
+
+	replacedFiles := make(map[base.FileNum][]newFileEntry)
+	for _, file := range ingestFlushable.files {
+		if ingestFlushable.exciseSpan.Valid() && ingestFlushable.exciseSpan.Contains(d.cmp, file.FileMetadata.Smallest) && ingestFlushable.exciseSpan.Contains(d.cmp, file.FileMetadata.Largest) {
+			level = 6
```
itsbilal, line 1472: Let's add a comment here explaining this case

ingest.go
```diff
@@ -2420,6 +2427,7 @@ func (d *DB) ingestApply(
 		}
 	}
 
+	// TODO(aaditya): should this metric also be incremented for flushableIngests?
```
itsbilal, line 2430: We count these under `Flush.AsIngest*` already which I guess covers this, but yeah it feels odd to not add to the ingestion count either. We should do this in the place where we increment the flush metric.

ingest.go
```diff
@@ -1359,6 +1365,7 @@ func (d *DB) ingest(
 	targetLevelFunc ingestTargetLevelFunc,
 	shared []SharedSSTMeta,
 	exciseSpan KeyRange,
+	doFlushableIngest bool,
```
itsbilal, line 1368: I think this should be called `localFilesContainExciseSpanTombstones`  or `sstsContainExciseTombstone` or something with an explanation of the requirement on the local sstable(s) containing a rangedel/rangekeydel spanning the entirety of the excise span. `doFlushableIngest` isn't appropriate as it also can't _force_ a flushable ingest to happen either.

ingest.go
```diff
@@ -1555,7 +1562,7 @@ func (d *DB) ingest(
 		}
 		// The ingestion overlaps with some entry in the flushable queue.
 		if d.FormatMajorVersion() < FormatFlushableIngest ||
-			d.opts.Experimental.DisableIngestAsFlushable() ||
+			d.opts.Experimental.DisableIngestAsFlushable() || !doFlushableIngest ||
```
itsbilal, line 1565: This is slightly inaccurate, as we still want to do a flushable ingest if there's no excise span or shared/external files, even if doFlushableIngest is false. Maybe we need:

```
canIngestFlushable := d.FormatMajorVersion() >= FormatFlushableIngest && (len(d.mu.mem.queue) < d.opts.MemTableStopWritesThreshold) && !d.opts.Experimental.DisableIngestAsFlushable()
hasExciseOrRemoteFiles := len(shared) > 0 || exciseSpan.Valid() || len(external) > 0

if !canIngestFlushable || (hasExciseOrRemoteFiles && !sstContainsExciseTombstone) {
...
}
```

itsbilal (commented):
Did a first pass and overall this looks great, just some minor comments. Looking forward to the tests

ingest.go
itsbilal: Actually, looks like after this change we _can_ do excise+local files, but we still can't flushable ingest shared or external files. So scratch what I wrote above and move the remote files stuff into `canIngestFlushable`.

ingest.go
aadityasondhi: Moving this to a separate PR since some of the metrics tests depend on these and I don't want to crowd this PR with that change.

aadityasondhi:
@itsbilal this is ready for another look.

ingest.go
aadityasondhi: https://github.com/cockroachdb/pebble/pull/3429

compaction.go
```diff
@@ -1448,17 +1448,68 @@ func (d *DB) runIngestFlush(c *compaction) (*manifest.VersionEdit, error) {
...
+		}
+	}
+
+	suggestSplit := d.opts.Experimental.IngestSplit != nil && d.opts.Experimental.IngestSplit() &&
+		d.FormatMajorVersion() >= FormatVirtualSSTables
+
+	replacedFiles := make(map[base.FileNum][]newFileEntry)
+	for _, file := range ingestFlushable.files {
+		// This file fits perfectly within the excise span, so we can slot it at L6.
+		if ingestFlushable.exciseSpan.Valid() && ingestFlushable.exciseSpan.Contains(d.cmp, file.FileMetadata.Smallest) && ingestFlushable.exciseSpan.Contains(d.cmp, file.FileMetadata.Largest) {
```
itsbilal, line 1473: nit: line wrap

metamorphic/ops.go
```diff
@@ -1988,7 +1988,7 @@ func (r *replicateOp) runSharedReplicate(
 		return
 	}
 
-	_, err = dest.IngestAndExcise([]string{sstPath}, sharedSSTs, nil /* external */, pebble.KeyRange{Start: r.start, End: r.end})
+	_, err = dest.IngestAndExcise([]string{sstPath}, sharedSSTs, nil /* external */, pebble.KeyRange{Start: r.start, End: r.end}, false)
```
itsbilal, line 1991: Let's open an issue to toggle this to true for some metamorphic runs (and if we're toggling this to true, we can throw in the relevant dels inside the ssts)

ingest.go
```diff
@@ -1162,7 +1162,7 @@ func (d *DB) Ingest(paths []string) error {
 	if d.opts.ReadOnly {
 		return ErrReadOnly
 	}
-	_, err := d.ingest(paths, ingestTargetLevel, nil /* shared */, KeyRange{}, nil /* external */)
+	_, err := d.ingest(paths, ingestTargetLevel, nil, KeyRange{}, false, nil)
```
itsbilal, line 1165: nit: let's not remove the `/* shared */` stuff

data_test.go
```diff
@@ -1258,6 +1258,7 @@ func (d *DB) waitTableStats() {
 
 func runIngestAndExciseCmd(td *datadriven.TestData, d *DB, fs vfs.FS) error {
 	var exciseSpan KeyRange
+	var doFlushableIngest bool
```
itsbilal, line 1261: We should rename this and the `case` below to `sstContainsExciseTombstone` and `contains-excise-tombstone` to improve readability between the datadriven testdata and the IngestAndExcise code itself.

itsbilal (approved):
LGTM, great work! Just some minor comments

ingest.go
aadityasondhi: artifact of goland refactor, caused more harm than good in this case 😅

metamorphic/ops.go
aadityasondhi: https://github.com/cockroachdb/pebble/issues/3432

PR #3399, opened 2024-03-13
db: allow Options.Lock's dirname to be equivalent path

jbowens:
When a user sets Options.Lock to pass in an already-acquired directory lock, Open validates that the lock was acquired for the same directory. Previously, this didn't account for uncleaned paths, relative paths, symlinks, etc. This commit adjusts the check to fall back to using os.SameFile to check whether the locked directories match if the paths are not identical.

### DIFF
--- a/open.go
+++ b/open.go
@@ -105,8 +105,8 @@ func Open(dirname string, opts *Options) (db *DB, err error) {
 	if opts.Lock != nil {
 		// The caller already acquired the database lock. Ensure that the
 		// directory matches.
-		if dirname != opts.Lock.dirname {
-			return nil, errors.Newf("pebble: opts.Lock acquired in %q not %q", opts.Lock.dirname, dirname)
+		if err := opts.Lock.pathMatches(dirname); err != nil {
+			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


sumeerbhola (approved):
:lgtm: 

I am curious what tripped up without this change.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3399#-:-NssWj7H3jZA-m0_7FMN:bfg7z5d)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @itsbilal)

jbowens (commented):
TFTR!

> I am curious what tripped up without this change.

the autoDecryptFS (used by the debug tool) converts paths to absolute when figuring out the right vfs.FS/fs.Env to use. now that we initialize the Env here, this means we acquire the directory lock using the absolute path. once we have the right VFS, we pass relative paths into the actual VFS methods

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3399#-:-NssXdAS2cU-xqe21EnC:b5vlwkj)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @itsbilal)

PR #3400, opened 2024-03-13
db: add virtual backing protection mechanism

RaduBerinde:
#### manifest: add Protect API to VirtualBackings

Add `Protect` and `Unprotect` which control a separate ref count on
each backing. This will allow external ingestions to temporarily
prevent a backing from being removed from the latest version.

#### db: handle protected backings in versionSet

This commit updates versionSet code to support protected virtual
backings. The main change is that we now also `Ref` all backings that
exist in `vs.virtualBackings`; we reorganize the code related to
zombie backings in `logAndApply` to facilitate maintaining these refs
correctly.

### DIFF
--- a/internal/manifest/testdata/virtual_backings
+++ b/internal/manifest/testdata/virtual_backings
@@ -2,14 +2,14 @@
 add n=1 size=100
 ----
 1 virtual backings, total size 100:
-  000001:  size=100  useCount=0  virtualizedSize=0
+  000001:  size=100  useCount=0  protectionCount=0  virtualizedSize=0
 unused virtual backings: 000001
 
 add n=2 size=200
 ----
 2 virtual backings, total size 300:
-  000001:  size=100  useCount=0  virtualizedSize=0
-  000002:  size=200  useCount=0  virtualizedSize=0
+  000001:  size=100  useCount=0  protectionCount=0  virtualizedSize=0
+  000002:  size=200  useCount=0  protectionCount=0  virtualizedSize=0
 unused virtual backings: 000001 000002
 
 add n=2 size=200
@@ -19,51 +19,107 @@ pebble: trying to add an existing file backing
 add-table n=1 size=10
 ----
 2 virtual backings, total size 300:
-  000001:  size=100  useCount=1  virtualizedSize=10
-  000002:  size=200  useCount=0  virtualizedSize=0
+  000001:  size=100  useCount=1  protectionCount=0  virtualizedSize=10
+  000002:  size=200  useCount=0  protectionCount=0  virtualizedSize=0
 unused virtual backings: 000002
 
 add-table n=1 size=10
 ----
 2 virtual backings, total size 300:
-  000001:  size=100  useCount=2  virtualizedSize=20
-  000002:  size=200  useCount=0  virtualizedSize=0
+  000001:  size=100  useCount=2  protectionCount=0  virtualizedSize=20
+  000002:  size=200  useCount=0  protectionCount=0  virtualizedSize=0
 unused virtual backings: 000002
 
 add-table n=1 size=10
 ----
 2 virtual backings, total size 300:
-  000001:  size=100  useCount=3  virtualizedSize=30
-  000002:  size=200  useCount=0  virtualizedSize=0
+  000001:  size=100  useCount=3  protectionCount=0  virtualizedSize=30
+  000002:  size=200  useCount=0  protectionCount=0  virtualizedSize=0
 unused virtual backings: 000002
 
 remove n=1
 ----
-backing 000001 still in use (useCount=3)
+backing 000001 still in use (useCount=3 protectionCount=0)
 
 remove-table n=1 size=10
 ----
 2 virtual backings, total size 300:
-  000001:  size=100  useCount=2  virtualizedSize=20
-  000002:  size=200  useCount=0  virtualizedSize=0
+  000001:  size=100  useCount=2  protectionCount=0  virtualizedSize=20
+  000002:  size=200  useCount=0  protectionCount=0  virtualizedSize=0
 unused virtual backings: 000002
 
 remove-table n=1 size=10
 ----
 2 virtual backings, total size 300:
-  000001:  size=100  useCount=1  virtualizedSize=10
-  000002:  size=200  useCount=0  virtualizedSize=0
+  000001:  size=100  useCount=1  protectionCount=0  virtualizedSize=10
+  000002:  size=200  useCount=0  protectionCount=0  virtualizedSize=0
 unused virtual backings: 000002
 
 remove-table n=1 size=10
 ----
 2 virtual backings, total size 300:
-  000001:  size=100  useCount=0  virtualizedSize=0
-  000002:  size=200  useCount=0  virtualizedSize=0
+  000001:  size=100  useCount=0  protectionCount=0  virtualizedSize=0
+  000002:  size=200  useCount=0  protectionCount=0  virtualizedSize=0
 unused virtual backings: 000001 000002
 
 remove n=1
 ----
 1 virtual backings, total size 200:
-  000002:  size=200  useCount=0  virtualizedSize=0
+  000002:  size=200  useCount=0  protectionCount=0  virtualizedSize=0
+unused virtual backings: 000002
+
+protect n=2
+----
+1 virtual backings, total size 200:
+  000002:  size=200  useCount=0  protectionCount=1  virtualizedSize=0
+
+protect n=2
+----
+1 virtual backings, total size 200:
+  000002:  size=200  useCount=0  protectionCount=2  virtualizedSize=0
+
+unprotect n=2
+----
+1 virtual backings, total size 200:
+  000002:  size=200  useCount=0  protectionCount=1  virtualizedSize=0
+
+remove n=2
+----
+backing 000002 still in use (useCount=0 protectionCount=1)
+
+unprotect n=2
+----
+1 virtual backings, total size 200:
+  000002:  size=200  useCount=0  protectionCount=0  virtualizedSize=0
+unused virtual backings: 000002
+
+add-table n=2 size=10
+----
+1 virtual backings, total size 200:
+  000002:  size=200  useCount=1  protectionCount=0  virtualizedSize=10
+
+add-table n=2 size=10
+----
+1 virtual backings, total size 200:
+  000002:  size=200  useCount=2  protectionCount=0  virtualizedSize=20
+
+protect n=2
+----
+1 virtual backings, total size 200:
+  000002:  size=200  useCount=2  protectionCount=1  virtualizedSize=20
+
+remove-table n=2 size=10
+----
+1 virtual backings, total size 200:
+  000002:  size=200  useCount=1  protectionCount=1  virtualizedSize=10
+
+remove-table n=2 size=10
+----
+1 virtual backings, total size 200:
+  000002:  size=200  useCount=0  protectionCount=1  virtualizedSize=0
+
+unprotect n=2
+----
+1 virtual backings, total size 200:
+  000002:  size=200  useCount=0  protectionCount=0  virtualizedSize=0
 unused virtual backings: 000002
--- a/internal/manifest/version.go
+++ b/internal/manifest/version.go
@@ -396,6 +396,9 @@ type FileBacking struct {
 	// backing across all versions that have a non-zero reference count. The tables
 	// in each version are maintained in a copy-on-write B-tree and each B-tree node
 	// keeps a reference on the respective backings.
+	//
+	// In addition, a reference count is taken for every backing in the latest
+	// version's VirtualBackings (necessary to support Protect/Unprotect).
 	refs atomic.Int32
 }
 
--- a/internal/manifest/version_edit.go
+++ b/internal/manifest/version_edit.go
@@ -924,9 +924,15 @@ func (b *BulkVersionEdit) Accumulate(ve *VersionEdit) error {
 		}
 	}
 
-	// Since a file can be removed from backing files in exactly one version
-	// edit it is safe to just append without any de-duplication.
-	b.RemovedFileBacking = append(b.RemovedFileBacking, ve.RemovedBackingTables...)
+	for _, n := range ve.RemovedBackingTables {
+		if _, ok := b.AddedFileBacking[n]; ok {
+			delete(b.AddedFileBacking, n)
+		} else {
+			// Since a file can be removed from backing files in exactly one version
+			// edit it is safe to just append without any de-duplication.
+			b.RemovedFileBacking = append(b.RemovedFileBacking, n)
+		}
+	}
 
 	return nil
 }
--- a/internal/manifest/virtual_backings.go
+++ b/internal/manifest/virtual_backings.go
@@ -15,7 +15,7 @@ import (
 )
 
 // VirtualBackings maintains information about the set of backings that support
-// virtual tables in a given version.
+// virtual tables in the latest version.
 //
 // The VirtualBackings set internally maintains for each backing the number of
 // virtual tables that use that backing and the sum of their virtual sizes. When
@@ -23,6 +23,48 @@ import (
 // tables. AddTable/RemoveTable are used to maintain the set of tables that are
 // associated with a backing. Finally, a backing can only be removed from the
 // set when it is no longer in use.
+//
+// -- Protection API --
+//
+// VirtualBackings exposes a Protect/Unprotect API. This is used to allow
+// external file ingestions to reuse existing virtual backings. Because
+// ingestions can run in parallel with other operations like compactions, it is
+// possible for a backing to "go away" in-between the time the ingestion decides
+// to use it and the time the ingestion installs a new version. The protection
+// API solves this problem by keeping backings alive, even if they become
+// otherwise unused by any tables.
+//
+// Backing protection achieves two goals:
+//   - it must prevent the removal of the backing from the latest version, where
+//     removal means becoming part of a VersionEdit.RemovedBackingTables. This
+//     is achieved by treating the backing as "in use", preventing Unused() from
+//     reporting it.
+//   - it must prevent the backing from becoming obsolete (i.e. reaching a ref
+//     count of 0). To achieve this, VirtualBackings takes a ref on each backing
+//     when it is added; this ref must be released after the backing is removed
+//     (when it is ok for the backing to be reported as obsolete).
+//
+// For example, say we have virtual table T1 with backing B1 and an ingestion tries
+// to reuse the file. This is what will usually happen (the happy case):
+//   - latest version is V1 and it contains T1(B1).
+//   - ingestion request comes for another virtual portion of B1. Ingestion process
+//     finds B1 and calls Protect(B1).
+//   - ingestion completes, installs version V2 which has T1(B1) and a new
+//     T2(B1), and calls Unprotect(B1).
+//
+// In this path, the Protect/Unprotect calls do nothing. But here is what could
+// happen (the corner case):
+//   - latest version is V1 and it contains T1(B1).
+//   - ingestion request comes for another virtual portion of B1. Ingestion process
+//     finds B1 and calls Protect(B1).
+//   - compaction completes and installs version V2 which no longer has T1.
+//     But because B1 is protected, V2 still has B1.
+//   - ingestion completes, installs version V3 which has a new T2(B1) and calls
+//     Unprotect(B1).
+//
+// If instead the ingestion fails to complete, the last step becomes:
+//   - ingestion fails, calls Unprotect(B1). B1 is now Unused() and the next
+//     version (applied by whatever next operation is) will remove B1.
 type VirtualBackings struct {
 	m map[base.DiskFileNum]backingWithMetadata
 
@@ -46,16 +88,25 @@ type backingWithMetadata struct {
 
 	// A backing initially has a useCount of 0. The useCount is increased by
 	// AddTable and decreased by RemoveTable. Backings that have useCount=0 are
-	// reported by Unused().
-	useCount        int
+
+	useCount int32
+	// protectionCount is used by Protect to temporarily prevent a backing from
+	// being reported as unused.
+	protectionCount int32
+	// virtualizedSize is the sum of the sizes of the useCount virtual tables
+	// associated with this backing.
 	virtualizedSize uint64
 }
 
-// Add adds a new backing that will be used by virtual tables. Another
+// AddAndRef adds a new backing to the set and takes a reference on it. Another
 // backing for the same DiskFilNum must not exist.
 //
-// The added backing is unused until it is associated with a table via AddTable.
-func (bv *VirtualBackings) Add(backing *FileBacking) {
+// The added backing is unused until it is associated with a table via AddTable
+// or protected via Protect.
+func (bv *VirtualBackings) AddAndRef(backing *FileBacking) {
+	// We take a reference on the backing because in case of protected backings
+	// (see Protect), we might be the only ones holding on to a backing.
+	backing.Ref()
 	bv.mustAdd(backingWithMetadata{
 		backing: backing,
 	})
@@ -63,12 +114,17 @@ func (bv *VirtualBackings) Add(backing *FileBacking) {
 	bv.totalSize += backing.Size
 }
 
-// Remove a backing; the backing must not be in use. Normally backings are
-// removed once they are reported by Unused().
+// Remove removes a backing. The backing must not be in use; normally backings
+// are removed once they are reported by Unused().
+//
+// It is up to the caller to release the reference took by AddAndRef.
 func (bv *VirtualBackings) Remove(n base.DiskFileNum) {
 	v := bv.mustGet(n)
 	if v.inUse() {
-		panic(errors.AssertionFailedf("backing %s still in use (useCount=%d)", v.backing.DiskFileNum, v.useCount))
+		panic(errors.AssertionFailedf(
+			"backing %s still in use (useCount=%d protectionCount=%d)",
+			v.backing.DiskFileNum, v.useCount, v.protectionCount,
+		))
 	}
 	delete(bv.m, n)
 	delete(bv.unused, v.backing)
@@ -98,9 +154,40 @@ func (bv *VirtualBackings) RemoveTable(m *FileMetadata) {
 	}
 	v := bv.mustGet(m.FileBacking.DiskFileNum)
 
+	if v.useCount <= 0 {
+		panic(errors.AssertionFailedf("invalid useCount"))
+	}
 	v.useCount--
 	v.virtualizedSize -= m.Size
 	bv.m[m.FileBacking.DiskFileNum] = v
+	if !v.inUse() {
+		bv.unused[v.backing] = struct{}{}
+	}
+}
+
+// Protect prevents a backing from being reported as unused until a
+// corresponding Unprotect call is made. The backing must be in the set.
+//
+// Multiple Protect calls can be made for the same backing; each must have a
+// corresponding Unprotect call before the backing can become unused.
+func (bv *VirtualBackings) Protect(n base.DiskFileNum) {
+	v := bv.mustGet(n)
+	if !v.inUse() {
+		delete(bv.unused, v.backing)
+	}
+	v.protectionCount++
+	bv.m[n] = v
+}
+
+// Unprotect reverses a Protect call.
+func (bv *VirtualBackings) Unprotect(n base.DiskFileNum) {
+	v := bv.mustGet(n)
+
+	if v.protectionCount <= 0 {
+		panic(errors.AssertionFailedf("invalid protectionCount"))
+	}
+	v.protectionCount--
+	bv.m[n] = v
 	if !v.inUse() {
 		bv.unused[v.backing] = struct{}{}
 	}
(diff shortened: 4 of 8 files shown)


sumeerbhola (changes requested):
> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400)* status: 2 of 7 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/manifest/virtual_backings.go` line 53 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3400#-NstAIcHC9-0FrbPMg6M:-NstAIcHC9-0FrbPMg6N:bkpr7lw) ([raw file](https://github.com/cockroachdb/pebble/blob/a86e47339e4dbbf18b218256e6548b7d2e1ca5dd/internal/manifest/virtual_backings.go#L53)):*
> ```Go
> 	// protectionCount is used by Protect to temporarily prevent a backing from
> 	// being reported as unused.
> 	protectionCount int32
> ```

I am curious why this needs to be a separate count field from `useCount`.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400)* status: 2 of 7 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`internal/manifest/virtual_backings.go` line 53 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3400#-NstAIcHC9-0FrbPMg6M:-Nstt4me3nwZYiak_OL5:b-ihn83t) ([raw file](https://github.com/cockroachdb/pebble/blob/a86e47339e4dbbf18b218256e6548b7d2e1ca5dd/internal/manifest/virtual_backings.go#L53)):*

I am curious why this needs to be a separate count field from `useCount`.
</blockquote></details>

`responsibleForGarbageBytes` wants to know how many tables use a backing. Also, I'd probably prefer to keep them separate anyway just for debuggability.

sumeerbhola (changes requested):
> Reviewed 5 of 5 files at r3, 2 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400)* status: 4 of 7 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

___
*[`version_set.go` line 529 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3400#-NsymuWt5dFx_d4Dp2gQ:-NsymuWt5dFx_d4Dp2gR:b-knxcv5) ([raw file](https://github.com/cockroachdb/pebble/blob/671cca10e67ca6a5eece13d4ae7a4241c1154dae/version_set.go#L529)):*
> ```Go
> 	// Update virtualBackings and populate ve.RemovedBackingTables.
> 	for _, b := range ve.CreatedBackingTables {
> 		b.Ref()
> ```

odd to see the `Ref` here, instead of inside `VirtualBackings.Add`.
___
*[`version_set.go` line 537 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nsyo7MJEfqh1Ry432WS:-Nsyo7MJEfqh1Ry432WT:b-ah5qbn) ([raw file](https://github.com/cockroachdb/pebble/blob/671cca10e67ca6a5eece13d4ae7a4241c1154dae/version_set.go#L537)):*
> ```Go
> 		}
> 	}
> 	for _, m := range ve.DeletedFiles {
> ```

I'm getting very confused here, in that the logic was nicely abstracted in `getZombiesAndUpdateVirtualBackings` and now it is harder to understand.
What is the key insight that I am missing in why this is needed? Is this to delay unreffing until after the write to the manifest? The previous version is still alive, so why does it matter if the unref happens now?
___
*[`internal/manifest/version.go` line 405 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nsyfi860J2z0N1VIj23:-Nsyfi860J2z0N1VIj24:b-ytih7z) ([raw file](https://github.com/cockroachdb/pebble/blob/671cca10e67ca6a5eece13d4ae7a4241c1154dae/internal/manifest/version.go#L405)):*
> ```Go
> 	//
> 	// In addition, a reference count is taken for every backing in the latest
> 	// version's VirtualBackings (necessary to support Protect/Unprotect).
> ```

I'm getting a bit confused about the lifetime. Something will be in the latest version's `VirtualBackings` if the version has a file reffing it. If it doesn't have a file reffing it, `unused` will return it before the version is installed and the file will be removed from `VirtualBackings`. I suppose with the change in this PR if the file is being protected, it will not be included in the return value from `unused` and so it will stay in the `VirtualBacking`. But if there isn't a ref protecting deletion of the file, it will get deleted while `VirtualBackings` still thinks it is alive.

Some code commentary would help:
- Pointing out that the only thing standing in the way of file deletion is `FileBacking.refs`.
- We don't know which of those refs are coming from the latest version and which from earlier versions, hence `backingFileMetadata.useCount` which counts just the latest ones, so we can compute the right `versionEdit`.
- If the latest version has useCount > 0, it implies `FileBacking.refs > 0` but not vice versa. 
- With this change if useCount + protectionCount > 0, it is possible that useCount = 0, so `VirtualBackings` needs to hold a ref to prevent file deletion.

The other question is why we need to delay returning something from `unused` until `useCount + protectionCount > 0`. Isn't it ok to return if `useCount == 0`? All it does is remove it from the version. I think the answer is that `VirtualBackings.Add` corresponds to it being added in a `VersionEdit.CreatedBackingTables`, so if it stops existing in the latest version and remains in `VirtualBackings` we will not add the metadata for it in a future version edit when the `useCount` increases above 0. That is, `VirtualBackings` needs to mirror what is in the latest version.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400)* status: 4 of 7 files reviewed, 4 unresolved discussions (waiting on @sumeerbhola)

___
*[`version_set.go` line 537 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nsyo7MJEfqh1Ry432WS:-Nsz21gi1zagdqufWdGm:bseet4p) ([raw file](https://github.com/cockroachdb/pebble/blob/671cca10e67ca6a5eece13d4ae7a4241c1154dae/version_set.go#L537)):*

I'm getting very confused here, in that the logic was nicely abstracted in `getZombiesAndUpdateVirtualBackings` and now it is harder to understand.
What is the key insight that I am missing in why this is needed? Is this to delay unreffing until after the write to the manifest? The previous version is still alive, so why does it matter if the unref happens now?
</blockquote></details>

Yes its so we Unref after. Some of the unused backings might not be in the previous version (thet's the entire point of the protection mechanism). Maybe we can Unref here and only do the addOnsolete below?
___
*[`internal/manifest/version.go` line 405 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nsyfi860J2z0N1VIj23:-Nsz33I66goYqbMNujOR:b-ba6xys) ([raw file](https://github.com/cockroachdb/pebble/blob/671cca10e67ca6a5eece13d4ae7a4241c1154dae/internal/manifest/version.go#L405)):*

I'm getting a bit confused about the lifetime. Something will be in the latest version's `VirtualBackings` if the version has a file reffing it. If it doesn't have a file reffing it, `unused` will return it before the version is installed and the file will be removed from `VirtualBackings`. I suppose with the change in this PR if the file is being protected, it will not be included in the return value from `unused` and so it will stay in the `VirtualBacking`. But if there isn't a ref protecting deletion of the file, it will get deleted while `VirtualBackings` still thinks it is alive.

Some code commentary would help:
- Pointing out that the only thing standing in the way of file deletion is `FileBacking.refs`.
- We don't know which of those refs are coming from the latest version and which from earlier versions, hence `backingFileMetadata.useCount` which counts just the latest ones, so we can compute the right `versionEdit`.
- If the latest version has useCount > 0, it implies `FileBacking.refs > 0` but not vice versa. 
- With this change if useCount + protectionCount > 0, it is possible that useCount = 0, so `VirtualBackings` needs to hold a ref to prevent file deletion.

The other question is why we need to delay returning something from `unused` until `useCount + protectionCount > 0`. Isn't it ok to return if `useCount == 0`? All it does is remove it from the version. I think the answer is that `VirtualBackings.Add` corresponds to it being added in a `VersionEdit.CreatedBackingTables`, so if it stops existing in the latest version and remains in `VirtualBackings` we will not add the metadata for it in a future version edit when the `useCount` increases above 0. That is, `VirtualBackings` needs to mirror what is in the latest version.
</blockquote></details>

Yes the point is to prevent it from being removed via DeletedBackingTables (not just to keep the object alive).  Perhaps another way is to add it back via CreatedBackingTables as necessary. There might be some problems with reusing backing numbers in that way.

To me it makes more sense for VirtualBackings to always mirror the backings you have in the latest version) those you would have if you read the manifest).

RaduBerinde:
*[`internal/manifest/version.go` line 405 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nsyfi860J2z0N1VIj23:-NszCPoN62fbd3Qm1eco:b-lbk81a) ([raw file](https://github.com/cockroachdb/pebble/blob/671cca10e67ca6a5eece13d4ae7a4241c1154dae/internal/manifest/version.go#L405)):*

Yes the point is to prevent it from being removed via DeletedBackingTables (not just to keep the object alive).  Perhaps another way is to add it back via CreatedBackingTables as necessary. There might be some problems with reusing backing numbers in that way.

To me it makes more sense for VirtualBackings to always mirror the backings you have in the latest version) those you would have if you read the manifest).
</blockquote></details>

Perhaps I should look at *always* delaying putting backings in DeletedBackimgTables until they are obsolete (ie ref count goes to 0)?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400)* status: 4 of 7 files reviewed, 4 unresolved discussions (waiting on @sumeerbhola)

___
*[`version_set.go` line 529 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3400#-NsymuWt5dFx_d4Dp2gQ:-Nsz4Gpw5YcI3gK-UHRr:b-udimei) ([raw file](https://github.com/cockroachdb/pebble/blob/671cca10e67ca6a5eece13d4ae7a4241c1154dae/version_set.go#L529)):*

odd to see the `Ref` here, instead of inside `VirtualBackings.Add`.
</blockquote></details>

Unref should happen at the same layer.. If we add Unref inside Delete, we have to handle the case where it's the last ref. Felt cleaner to not entangle these in VirtualBackings
___
*[`version_set.go` line 537 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nsyo7MJEfqh1Ry432WS:-NszOfcXDk2lojoq--Ma:bslx7uw) ([raw file](https://github.com/cockroachdb/pebble/blob/671cca10e67ca6a5eece13d4ae7a4241c1154dae/version_set.go#L537)):*
> Maybe we can Unref here and only do the addOnsolete below?

I tried this (in a separate commit r5) and reinstated the getZombies function, it does feel cleaner. Let me know what you think.

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400)* status: 4 of 7 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

___
*[`version_set.go` line 537 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nsyo7MJEfqh1Ry432WS:-Nt19sC-0XpZU_s7L-hl:bfy2i15) ([raw file](https://github.com/cockroachdb/pebble/blob/671cca10e67ca6a5eece13d4ae7a4241c1154dae/version_set.go#L537)):*

> Maybe we can Unref here and only do the addOnsolete below?

I tried this (in a separate commit r5) and reinstated the getZombies function, it does feel cleaner. Let me know what you think.
</blockquote></details>

I haven't looked at the latest code changes yet, but was thinking about it this morning and it looks to me that the current getZombiesAndUpdateVirtualBackings (gzauvb) structure is viable.

Let's say we are going from version V1 to version V2, using version edit VE2. The current VirtualBackings is VB1. VB1 contains exactly the virtual sst backings that are in V1. For a backing in V1, we have one of two cases:
- Both V1 and VB1 have refs.
- Only VB1 has a ref. This is when a protect call prevented it from becoming unused when V1 was being installed, even though useCount was 0, i.e., it was not returned by the unused method. It is possible that useCount+protectionCount is now 0, but that is fine since we still hold the ref. Consider two such file backings f1 and f2, that have useCount+protectionCount equal to 0.

VB2 is going to be the next value of VirtualBackings. We have the partial VE2 that does not contain RemovedBackingTables, and we call gzauvb(VE2, VB1):
- VE2 contains some new file(s) that are backed by f1. We only allow physical to virtual backing transitions, so these new file must be virtual. We have not yet taken a ref in V2, since V2 has not been created yet. But we will call AddFile here, that will prevent it from becoming unused. When we exit gzauvb, the only ref to f1 is held by VB2, and when V2 is generated it will additionally hold a ref.
- VE2 contains no new file that is backed by f2. gzauvb will see it returned from unused, and will call VirtualBackings.Remove. The Remove method should also return whether the ref count dropped to 0. If so, the file belongs in both zombies and obsolete, and both should be returned from gzauvb. After vs.append we would explicitly (inside logAndApply) call versionSet.addObsolete.

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400)* status: 4 of 7 files reviewed, 6 unresolved discussions (waiting on @RaduBerinde)

___
*[`version_set.go` line 689 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nt1Atg924RKEIBFz1YU:-Nt1Atg924RKEIBFz1YV:b6swqpx) ([raw file](https://github.com/cockroachdb/pebble/blob/e6d0a4200fad928382507a46d94baf80145d2602/version_set.go#L689)):*
> ```Go
> 			delete(zombies, nf.Meta.FileBacking.DiskFileNum)
> 		}
> 	}
> ```

Based on my latest comment, do we even need this structure to change?
It was simple to understand that we
- first handle all the physical backings, then move to virtual backings.
- for virtual backings the first thing is to call VirtualBackings.Add and it suffices to do those for CreatedBackingTables.
- then handle the AddTable and RemoveTable, and finally Unused.
- All the zombie and obsolete logic happen when we call Unused.

Perhaps this is because this code is trying to avoid entangling `VirtualBackings` in reffing and unreffing. I am not sure why. We need to maintain the invariant that everything in `VirtualBackings` has a ref held by it, so it should do it. If we add a code comment that explains things in a manner akin to my latest comment, I think it should be easy-ish for the reader.
___
*[`version_set.go` line 820 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nt1ClJM-nyMqQz0aZ7T:-Nt1ClJN8gAymhJlSF62:btj5m5w) ([raw file](https://github.com/cockroachdb/pebble/blob/dae13c6dd1d70280cc727bf28f56fc2db17a906e/version_set.go#L820)):*
> ```Go
> 	fileNum, minUnflushedLogNum base.DiskFileNum,
> 	nextFileNum uint64,
> 	virtualBackings []*fileBacking,
> ```

related to the other comments, IIUC, we shouldn't need this change to accomplish the protect stuff.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400)* status: 4 of 7 files reviewed, 6 unresolved discussions (waiting on @sumeerbhola)

___
*[`version_set.go` line 537 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nsyo7MJEfqh1Ry432WS:-Nt1Ow_p6G3Ly37kIhAs:b-w50gfn) ([raw file](https://github.com/cockroachdb/pebble/blob/671cca10e67ca6a5eece13d4ae7a4241c1154dae/version_set.go#L537)):*

I haven't looked at the latest code changes yet, but was thinking about it this morning and it looks to me that the current getZombiesAndUpdateVirtualBackings (gzauvb) structure is viable.

Let's say we are going from version V1 to version V2, using version edit VE2. The current VirtualBackings is VB1. VB1 contains exactly the virtual sst backings that are in V1. For a backing in V1, we have one of two cases:
- Both V1 and VB1 have refs.
- Only VB1 has a ref. This is when a protect call prevented it from becoming unused when V1 was being installed, even though useCount was 0, i.e., it was not returned by the unused method. It is possible that useCount+protectionCount is now 0, but that is fine since we still hold the ref. Consider two such file backings f1 and f2, that have useCount+protectionCount equal to 0.

VB2 is going to be the next value of VirtualBackings. We have the partial VE2 that does not contain RemovedBackingTables, and we call gzauvb(VE2, VB1):
- VE2 contains some new file(s) that are backed by f1. We only allow physical to virtual backing transitions, so these new file must be virtual. We have not yet taken a ref in V2, since V2 has not been created yet. But we will call AddFile here, that will prevent it from becoming unused. When we exit gzauvb, the only ref to f1 is held by VB2, and when V2 is generated it will additionally hold a ref.
- VE2 contains no new file that is backed by f2. gzauvb will see it returned from unused, and will call VirtualBackings.Remove. The Remove method should also return whether the ref count dropped to 0. If so, the file belongs in both zombies and obsolete, and both should be returned from gzauvb. After vs.append we would explicitly (inside logAndApply) call versionSet.addObsolete.
</blockquote></details>

Yeah this is exactly what the latest commit does. I haven't moved the Red/Unref inside Add/Remove but can easily do that now since they happen at the same place.
___
*[`version_set.go` line 820 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nt1ClJM-nyMqQz0aZ7T:-Nt1PGOv5zqCWF5mkrcO:b-mjaaf2) ([raw file](https://github.com/cockroachdb/pebble/blob/dae13c6dd1d70280cc727bf28f56fc2db17a906e/version_set.go#L820)):*

related to the other comments, IIUC, we shouldn't need this change to accomplish the protect stuff.
</blockquote></details>

We do.. this function tries to determine the backings in the latest version based on the tables. Protected backings will dissapear. If virtualBackings was a COW structure, this would just consult the list in the currenr version.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400)* status: 4 of 7 files reviewed, 6 unresolved discussions (waiting on @RaduBerinde)

___
*[`version_set.go` line 820 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nt1ClJM-nyMqQz0aZ7T:-Nt26F6Z9mpfwJCfmDlU:bpj1ujx) ([raw file](https://github.com/cockroachdb/pebble/blob/dae13c6dd1d70280cc727bf28f56fc2db17a906e/version_set.go#L820)):*

We do.. this function tries to determine the backings in the latest version based on the tables. Protected backings will dissapear. If virtualBackings was a COW structure, this would just consult the list in the currenr version.
</blockquote></details>

I think I see what is happening now. Let me try to play back my mental model and you can correct me.

We had this dedup structure, which meant we didn't need to appeal to anything in `VirtualBackings`. We discovered exactly the backings we needed and and wrote them as `CreatedBackingTables`. It didn't matter that we had already updated `VirtualBackings` before calling `createManifest`.

Now with the protect feature, we have a ref inside `VirtualBackings`. That ref could go away when `VirtualBackings` is updated, say for file f, which has no remaining refs. If the machine crashes after creating this manifest and before writing the version edit, we are still ok since we delete file f after writing the version edit.

But f, by definition, is one where the current version has no mentions. So if we reconstruct `CreatedBackingTables` purely based on the mentions in this version, we won't write f in `CreatedBackingTables`. This will create an inconsistency in the case when we don't crash, since the next version edit expects to find f.

We could have passed the zombies here, since it is a superset of files like f. But since we have to iterate over the whole version anyway, to write this snapshot, this code takes the simpler approach of grabbing all the backings from the `VirtualBackings` before editing it, and passing it in. And that means the dedup logic goes away. Sounds good.

RaduBerinde:
*[`version_set.go` line 820 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nt1ClJM-nyMqQz0aZ7T:-Nt2EqraAOh-KluIygRY:b5t1gde) ([raw file](https://github.com/cockroachdb/pebble/blob/dae13c6dd1d70280cc727bf28f56fc2db17a906e/version_set.go#L820)):*

I think I see what is happening now. Let me try to play back my mental model and you can correct me.

We had this dedup structure, which meant we didn't need to appeal to anything in `VirtualBackings`. We discovered exactly the backings we needed and and wrote them as `CreatedBackingTables`. It didn't matter that we had already updated `VirtualBackings` before calling `createManifest`.

Now with the protect feature, we have a ref inside `VirtualBackings`. That ref could go away when `VirtualBackings` is updated, say for file f, which has no remaining refs. If the machine crashes after creating this manifest and before writing the version edit, we are still ok since we delete file f after writing the version edit.

But f, by definition, is one where the current version has no mentions. So if we reconstruct `CreatedBackingTables` purely based on the mentions in this version, we won't write f in `CreatedBackingTables`. This will create an inconsistency in the case when we don't crash, since the next version edit expects to find f.

We could have passed the zombies here, since it is a superset of files like f. But since we have to iterate over the whole version anyway, to write this snapshot, this code takes the simpler approach of grabbing all the backings from the `VirtualBackings` before editing it, and passing it in. And that means the dedup logic goes away. Sounds good.
</blockquote></details>

Exactly. Thanks for the thorough review, it helps a lot. I will work on improving the comments and I will add an example of the "normal path" where all this refcounting is inconsequential and an example of the "corner case" path.

RaduBerinde:
*[`version_set.go` line 820 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nt1ClJM-nyMqQz0aZ7T:-Nt2FMqr03fxor_l-gvl:b-phdnij) ([raw file](https://github.com/cockroachdb/pebble/blob/dae13c6dd1d70280cc727bf28f56fc2db17a906e/version_set.go#L820)):*

Exactly. Thanks for the thorough review, it helps a lot. I will work on improving the comments and I will add an example of the "normal path" where all this refcounting is inconsequential and an example of the "corner case" path.
</blockquote></details>

(and I will move the Ref/Unref inside VirtualBackings)

RaduBerinde (commented):
Updated. Added a detailed comment about the protection mechanism in VirtualBackings.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400#-:-Nt32L723-2YdcmymY3z:bbj19ok)* status: 1 of 8 files reviewed, 6 unresolved discussions (waiting on @sumeerbhola)

sumeerbhola (approved):
looks great
:lgtm: 

> Reviewed 2 of 5 files at r4, 3 of 5 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400#-:-Nt8if4CEEhT3qjEdHzd:bmf8erq)* status: all files reviewed, 8 unresolved discussions (waiting on @RaduBerinde)

___
*[`version_set.go` line 922 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nt8gsB30v3mUN2426K9:-Nt8gsB30v3mUN2426KA:bp3g4hu) ([raw file](https://github.com/cockroachdb/pebble/blob/d7326d02add22a4ee7f4b73d9426a8f91542a29c/version_set.go#L922)):*
> ```Go
> 	}
> 	vs.virtualBackings.ForEach(func(b *fileBacking) {
> 		m[b.DiskFileNum] = struct{}{}
> ```

Could you add a comment that virtualBackings contains both backings that are referenced by some virtual tables in the latest version (which are handled above), and those that are not but stayed alive due to protection. And this loop is ensuring the latter get added to the map.
___
*[`internal/manifest/version_edit.go` line 929 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nt8dsTz-sX_awYCREWl:-Nt8dsTz-sX_awYCREWm:b947lxd) ([raw file](https://github.com/cockroachdb/pebble/blob/d7326d02add22a4ee7f4b73d9426a8f91542a29c/internal/manifest/version_edit.go#L929)):*
> ```Go
> 	for _, n := range ve.RemovedBackingTables {
> 		if _, ok := b.AddedFileBacking[n]; ok {
> 			delete(b.AddedFileBacking, n)
> ```

is this a bug in master in that when BVE accumulated many VersionEdits during open, it would not remove from the AddedFileBacking if a later VE had removed it? Or this was getting handled correctly when the BVE is applied to the Version?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400#-:-Nt8wJUk3vy4ec96Rx53:b-814g8k)* status: 7 of 8 files reviewed, 8 unresolved discussions (waiting on @sumeerbhola)

___
*[`version_set.go` line 922 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nt8gsB30v3mUN2426K9:-Nt8uVRcAyvPJTYcXryn:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/d7326d02add22a4ee7f4b73d9426a8f91542a29c/version_set.go#L922)):*

Could you add a comment that virtualBackings contains both backings that are referenced by some virtual tables in the latest version (which are handled above), and those that are not but stayed alive due to protection. And this loop is ensuring the latter get added to the map.
</blockquote></details>

Done.
___
*[`internal/manifest/version_edit.go` line 929 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3400#-Nt8dsTz-sX_awYCREWl:-Nt8vLrCEdaqjFYdmGKs:b-zgquf3) ([raw file](https://github.com/cockroachdb/pebble/blob/d7326d02add22a4ee7f4b73d9426a8f91542a29c/internal/manifest/version_edit.go#L929)):*

is this a bug in master in that when BVE accumulated many VersionEdits during open, it would not remove from the AddedFileBacking if a later VE had removed it? Or this was getting handled correctly when the BVE is applied to the Version?
</blockquote></details>

These aren't even used when applying, since the version doesn't have a separate record of the backings. They are only used separately during manifest replay. All the `AddedBackingTables` were added to the map and then the Removed ones were then removed. It wasn't incorrect, just potentially a bit wasteful.

RaduBerinde (commented):
There is a problem with moving the `Unref` to happen before the unlocked section (I hit it in the meta test, with backing deduplication on top of this PR). If some old version was also hanging on to the backing and that version gets unrefed, it could do the last `Unref` on the backing before the `zombieTables` get updated, and we get the `obsolete table not marked as zombie` fatal error.

We could potentially move up the updating of the zombie tables (but I don't like it since it's less cleanly defined), or we could remove that check.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400#-:-Nt9-G1-1E-1iJlvPxBs:bvo69jv)* status: 7 of 8 files reviewed, 8 unresolved discussions (waiting on @sumeerbhola)

RaduBerinde (commented):
Note that I can't move the Remove calls to after the manifest update because we don't want an ingestion to use any of the backings that are removed in the version edit.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400#-:-Nt937xm97O7QxzRrQB-:b7eobqp)* status: 7 of 8 files reviewed, 8 unresolved discussions (waiting on @sumeerbhola)

RaduBerinde (commented):
The problem is more than just that panic. Once something makes it into the obsolete tables, it could get deleted at any time. If we crash before writing to the manifest, we will find the backing in the manifest but the object could have been deleted already. We are going to delete it anyway so it won't matter in practice, but it feels subtle and not quite right.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400#-:-Nt94SAZ1KYtFpO5UFvh:b-gq4nds)* status: 7 of 8 files reviewed, 8 unresolved discussions (waiting on @sumeerbhola)

RaduBerinde (commented):
The smallest diff to address the issue is to restore the later `Unref` while keeping the existing structure the same: https://gist.github.com/RaduBerinde/9edca4a63605448c15d0c3e731a31009

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400#-:-Nt9A1WTD51G6_3IKgpR:b-b7vz84)* status: 7 of 8 files reviewed, 8 unresolved discussions (waiting on @sumeerbhola)

sumeerbhola (commented):
> The problem is more than just that panic. Once something makes it into the obsolete tables, it could get deleted at any time. If we crash before writing to the manifest, we will find the backing in the manifest but the object could have been deleted already. We are going to delete it anyway so it won't matter in practice, but it feels subtle and not quite right.

I agree this is not right. IIUC, the problem is not when the table becomes obsolete in the `getZombies` call, but that once we get rid of the ref in `getZombies` any other concurrent action could cause it to become obsolete and deleted, and then we could crash before writing the manifest.

The diff looks good.

> Reviewed 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400#-:-NtDDE6DC3eoCFTWX5mj:b3rwy66)* status: all files reviewed (commit messages unreviewed), 8 unresolved discussions (waiting on @RaduBerinde)

RaduBerinde (commented):
Thanks, updated. I fixed up some comments and made the code that calls `addObsolete` more straightforward, the DeleteFunc thing was hard to read.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400#-:-NtDf0GZ7uh8g1IGx02O:b4blk7q)* status: 5 of 8 files reviewed, 8 unresolved discussions (waiting on @sumeerbhola)

sumeerbhola (approved):
:lgtm: 

> Reviewed 3 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3400#-:-NtDg6XK8ZuDATFZUCWk:bg7igkv)* status: all files reviewed, 8 unresolved discussions (waiting on @RaduBerinde)

RaduBerinde:
Thanks again for your help!

PR #3401, opened 2024-03-13
internal/manifest: add size to debug string

dt:


itsbilal (approved):
LGTM

RaduBerinde (approved):
The brackets are unnecessary/confusing. Should be easy to make Parse work without them (just move the Expects inside the cases).

dt:
> The brackets are unnecessary/confusing

I think there's a future where we put backing size as well as size here. I'd prefer to keep them?

RaduBerinde (approved):
Thanks! :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3401#-:-NstuLtU9jhxk8d5uCKD:b6duv8n)* status: 0 of 8 files reviewed, all discussions resolved

PR #3402, opened 2024-03-13
sstable: cleanup/move code, fix two-level when calling sync add

dt:
Pulling these fixups out of an in-progress change to make it and them easier to review in isolation.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2509,29 +2509,11 @@ func (d *DB) runCopyCompaction(
 		}
 		pendingOutputs = append(pendingOutputs, newMeta)
 
-		if err := func() error {
-			size := src.Size()
-			r := src.NewReadHandle(ctx)
-			r.SetupForCompaction()
-			buf := make([]byte, 128<<10)
-			for pos := int64(0); pos < size; {
-				n := min(size-pos, int64(len(buf)))
-				if n == 0 {
-					break
-				}
-				readErr := r.ReadAt(ctx, buf[:n], pos)
-				if readErr != nil {
-					w.Abort()
-					return readErr
-				}
-				pos += n
-				if err := w.Write(buf[:n]); err != nil {
-					w.Abort()
-					return err
-				}
-			}
-			return w.Finish()
-		}(); err != nil {
+		if err := objstorage.Copy(ctx, src, w, 0, uint64(src.Size())); err != nil {
+			w.Abort()
+			return ve, pendingOutputs, err
+		}
+		if err := w.Finish(); err != nil {
 			return ve, pendingOutputs, err
 		}
 	} else {
--- a/objstorage/objstorage.go
+++ b/objstorage/objstorage.go
@@ -331,3 +331,26 @@ type RemoteObjectToAttach struct {
 	// implementation).
 	Backing RemoteObjectBacking
 }
+
+// Copy copies the specified range from the input to the output.
+func Copy(ctx context.Context, in Readable, out Writable, offset, length uint64) error {
+	r := in.NewReadHandle(ctx)
+	r.SetupForCompaction()
+	buf := make([]byte, 256<<10)
+	end := offset + length
+	for offset < end {
+		n := min(end-offset, uint64(len(buf)))
+		if n == 0 {
+			break
+		}
+		readErr := r.ReadAt(ctx, buf[:n], int64(offset))
+		if readErr != nil {
+			return readErr
+		}
+		offset += n
+		if err := out.Write(buf[:n]); err != nil {
+			return err
+		}
+	}
+	return nil
+}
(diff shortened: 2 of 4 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3402)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`sstable/suffix_rewriter.go` line 276 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3402#-Nstqzp20wXY8RplonnG:-Nstqzp20wXY8RplonnH:b-kqagfw) ([raw file](https://github.com/cockroachdb/pebble/blob/dd2913be0bf626a77a2d38c198caaa2b8c528d10/sstable/suffix_rewriter.go#L276)):*
> ```Go
> 
> func checkWriterFilterMatchesReader(r *Reader, w *Writer) error {
> 	if w.filter != nil {
> ```

[nit] Maybe leave this check around the call, otherwise the function should be `checkWriterFilterNotSetOrMatchesReader` :)
___
*[`sstable/writer.go` line 2021 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3402#-NstqQG-EYA3YJJ2tZxJ:-NstqQG-EYA3YJJ2tZxK:b-g60zy9) ([raw file](https://github.com/cockroachdb/pebble/blob/cb0bb41723807f99097dc0b90b92e2c693400394/sstable/writer.go#L2021)):*
> ```Go
> 
> 	{
> 		if len(w.props.UserProperties) == 0 {
> ```

Can you add a comment explaining in what case it will already be populated?

Do we want to check (in `invariants.Enabled` mode) that the props would be the same? Feels a bit strange that we don't finish the collectors.
___
*[`sstable/writer.go` line 1589 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3402#-Nstruxu0zQaYIubTL4_:-Nstruxu0zQaYIubTL4a:bdnwjw8) ([raw file](https://github.com/cockroachdb/pebble/blob/c05326f5599c4c46422dcb8d8b11cd5c22f455e7/sstable/writer.go#L1589)):*
> ```Go
> 		flushableIndexBlock = w.indexBlock
> 		w.indexBlock = newIndexBlockBuf(w.coordination.parallelismEnabled)
> 		w.twoLevelIndex = true
> ```

Are we missing a test for this case?

dt (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3402)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`sstable/suffix_rewriter.go` line 276 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3402#-Nstqzp20wXY8RplonnG:-NstvAoICTz1FMHmwlzE:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/dd2913be0bf626a77a2d38c198caaa2b8c528d10/sstable/suffix_rewriter.go#L276)):*

[nit] Maybe leave this check around the call, otherwise the function should be `checkWriterFilterNotSetOrMatchesReader` :)
</blockquote></details>

Done.
___
*[`sstable/writer.go` line 2021 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3402#-NstqQG-EYA3YJJ2tZxJ:-Nstsa500bxFkSqLEyd6:b-gwlr5l) ([raw file](https://github.com/cockroachdb/pebble/blob/cb0bb41723807f99097dc0b90b92e2c693400394/sstable/writer.go#L2021)):*
It never is (already populated) right now, but will be in an upcoming change where I'm driving `Writer` myself. Commented that this is only hit if someone is taking manual control over the props.

> Do we want to check (in `invariants.Enabled` mode)
No, I don't think so -- the point of this is that someone has taken direct control over choosing the props, so rather than computing them via the collectors, and indeed the collectors might not product the same results (or results at all, since we didn't iterate the keys for them).
___
*[`sstable/writer.go` line 1589 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3402#-Nstruxu0zQaYIubTL4_:-Nstw0zo9TAq3ZfiVgaB:b186drd) ([raw file](https://github.com/cockroachdb/pebble/blob/c05326f5599c4c46422dcb8d8b11cd5c22f455e7/sstable/writer.go#L1589)):*

Are we missing a test for this case?
</blockquote></details>

Seems like it. I added a TODO to explicitly unit test this function -- I don't really want to detour to that right now; my stacked changes do actually add [indirect] coverage of this line.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3402#-:-Nstx-22AH_7GAfXARYF:bnfp4nl)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @itsbilal)

PR #3405, opened 2024-03-13
remove support for prefix replacement

RaduBerinde:
Remove `PrefixReplacement` in favor of `SyntheticPrefix`. The general prefix replacement turns out to be too difficult so we will instead strip the prefix when writing out the backup files.


internal/manifest/version.go
```diff
@@ -865,14 +861,14 @@ func (m *FileMetadata) Validate(cmp Compare, formatKey base.FormatKey) error {
 		return base.CorruptionErrorf("file metadata FileBacking not set")
 	}
 
-	if m.PrefixReplacement != nil {
+	if m.SyntheticPrefix.IsSet() {
 		if !m.Virtual {
 			return base.CorruptionErrorf("prefix replacement rule set with non-virtual file")
```
msbutler, line 866: uber nit: these error messages could be updated to use the phrase "synthetic prefix" instead of "prefix replacement"

internal/manifest/version_edit.go
```diff
@@ -385,18 +385,19 @@ func (v *VersionEdit) Decode(r io.Reader) error {
 						}
 
 					case customTagPrefixRewrite:
+						// We used to have a content prefix; we no longer use it.
```
msbutler, line 388: why can't we  completely delete this content prefix logic?

msbutler (approved):
Thanks for doing this!!

RaduBerinde:
*[`internal/manifest/version_edit.go` line 388 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3405#gh-1525001122:-NsxVWKHCa0PMGPDAGcp:b21pykh) ([raw file](https://github.com/cockroachdb/pebble/blob/a7318cd5f3e8675b546f28693e239a83fbe6b0df/internal/manifest/version_edit.go#L388)):*

why can't we  completely delete this content prefix logic?
</blockquote></details>

We can, but everybody who is currently running experiments will have to wipe their stores.. Seemed like keeping that extra byte would be fine.

We could also add a new tag and then retire this tag a bit later, not sure it's worth the trouble though.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3405#-:-NsxofAe2waaiL6koSdK:b-814g8k)* status: 0 of 21 files reviewed, 2 unresolved discussions (waiting on @dt and @msbutler)

___
*[`internal/manifest/version.go` line 866 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3405#gh-1524997802:-NsxnD_zAbK8zRn1fKdX:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/a7318cd5f3e8675b546f28693e239a83fbe6b0df/internal/manifest/version.go#L866)):*

uber nit: these error messages could be updated to use the phrase "synthetic prefix" instead of "prefix replacement"
</blockquote></details>

Done.
___
*[`internal/manifest/version_edit.go` line 388 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3405#gh-1525001122:-Nsxob8DA0V3a010Tcrq:bt9goa1) ([raw file](https://github.com/cockroachdb/pebble/blob/a7318cd5f3e8675b546f28693e239a83fbe6b0df/internal/manifest/version_edit.go#L388)):*

We can, but everybody who is currently running experiments will have to wipe their stores.. Seemed like keeping that extra byte would be fine.

We could also add a new tag and then retire this tag a bit later, not sure it's worth the trouble though.
</blockquote></details>

Ok I cleaned this up as discussed offline.

msbutler (approved):

PR #3406, opened 2024-03-14
sstable: add CopySpan, use in runCopyCompaction

dt:


RaduBerinde (commented):
Very cool! At a high level this is great, will want another pair of eyes (@bilal or @jbowens or @sumeerbhola) to make sure we're getting all the details right.

We should add `downloadSpan` operations to the metamorphic test. I can look at doing that soon.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3406#-:-Nsue8LJ2FxOVjlVSfJU:b-w35vyv)* status: 0 of 4 files reviewed, 14 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`sstable/copier.go` line 22 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsueIZQEMSla5-5cNH0:-NsueIZQEMSla5-5cNH1:b-cwrn6t) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L22)):*
> ```Go
> // file that were adjacent to but outside that span.
> //
> // CopyAtLeastSpan differs from simply seeking a reader to start and iterating a
> ```

[int] CopySpan
___
*[`sstable/copier.go` line 25 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsua6AEBojiHgHJ75tf:-Nsua6AEBojiHgHJ75tg:bg3xm79) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L25)):*
> ```Go
> // until end passing the results to a writer in that it does not write the new
> // sstable from scratch, key-by-key, recompressing each key nto new blocks and
> // computing new filters and properties. Instead finds _block_ that intersect
> ```

[nit] Instead, it finds _blocks_
___
*[`sstable/copier.go` line 39 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsuesZGCawQj8SwXK32:-NsuesZGCawQj8SwXK33:bgm106z) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L39)):*
> ```Go
> 	o WriterOptions,
> 	start, end InternalKey,
> ) (uint64, error) {
> ```

The return isn't documented. It would also help if it was named, e.g. `(size uint64, _ error)`. Or just return the entire `w.meta`.
___
*[`sstable/copier.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsubAWeDOTVLv8g7jYd:-NsubAWeDOTVLv8g7jYe:b-z7bpgf) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L49)):*
> ```Go
> 
> 	if r.Properties.NumValueBlocks > 0 || r.Properties.NumRangeKeys() > 0 {
> 		return 0, errors.New("sstable value blocks or range keys")
> ```

`output.Abort()`
___
*[`sstable/copier.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsueek_6zcjWn7x4xaV:-Nsueeka1JaNnTt9N2PU:bq0dj69) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L62)):*
> ```Go
> 			// set w.err to any non-nil error just so it aborts instead of finishing.
> 			w.err = base.ErrNotFound
> 			// w.Close now owns calling output.Abort() or output.Finish().
> ```

[nit] just output.Abort
___
*[`sstable/copier.go` line 79 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsufjUR8zYLjtzT2wgK:-NsufjUR8zYLjtzT2wgL:bil8tao) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L79)):*
> ```Go
> 	if len(blocks) > 0 {
> 		offset := blocks[0].bh.Offset
> 		length := blocks[len(blocks)-1].bh.Offset + blocks[len(blocks)-1].bh.Length + blockTrailerLen - offset
> ```

[nit] maybe an invariants assertion that the sum of the lengths matches?
___
*[`sstable/copier.go` line 97 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsublLJ03SZSDk3fiL6:-NsublLJ03SZSDk3fiL7:b-8ajl0l) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L97)):*
> ```Go
> 		}
> 
> 		// TODO(dt): range keys.
> ```

[it] For now we should check and error out if we have these kinds of blocks.
___
*[`sstable/copier.go` line 103 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsubrr1BGbIsB972DpZ:-Nsubrr1BGbIsB972Dp_:b-p1pyd5) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L103)):*
> ```Go
> 		// NB: Some of these -- such as total keys -- will be over-counts since they
> 		// include the blocks we didn't copy, but over-estimating is safer than under,
> 		// e.g. if we tried to scale based on fraction of blocks copied.
> ```

In the future, we should consider having per-block props that then get aggregated to get the top-level props. We should keep that in mind if we implement a new format.
___
*[`sstable/copier.go` line 124 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsufGO28KTFCbyFJZtm:-NsufGO28KTFCbyFJZtn:b4gnltc) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L124)):*
> ```Go
> 	wrote := w.meta.Size
> 	w = nil
> 	return wrote, err
> ```

`nil` instead of err
___
*[`sstable/copier.go` line 127 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsug-DTBAEnlr-k9zvX:-Nsug-DTBAEnlr-k9zvY:b-1sxeyi) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L127)):*
> ```Go
> }
> 
> type indexEntry struct {
> ```

[nit] should this be `dataBlockInfo` so it's more descriptive?
___
*[`sstable/copier.go` line 129 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsug6ww1CCZQ6wx4tEE:-Nsug6ww1CCZQ6wx4tEF:b-z2ycnf) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L129)):*
> ```Go
> type indexEntry struct {
> 	bh  BlockHandleWithProperties
> 	key InternalKey
> ```

[nit] a comment about the key would be good (it's the "separator" key)
___
*[`sstable/copier.go` line 132 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsucCcTBo6ZvcqjDEhx:-NsubYpl96yz9J2wezg0:b-sxzrz0) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L132)):*
> ```Go
> }
> 
> func intersectingIndex(ctx context.Context, r *Reader, start, end InternalKey) ([]indexEntry, error) {
> ```

[nit] Worth adding a function comment. Also not sure the name is very good, maybe `getDataBlocksIntersectingRange`?
___
*[`sstable/copier.go` line 157 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsud7wI69iCHNICBENl:-Nsud7wI69iCHNICBENm:b-nnhj9h) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L157)):*
> ```Go
> 			res = append(res, entry)
> 		} else {
> 			subBlk, err := r.readBlock(ctx, bh.BlockHandle, nil, nil, nil, nil, nil)
> ```

This part very closely follows the outer part (as it should).. We could extract a helper function `forEachBlockInRange(.. func(bh))` that we can use twice
___
*[`sstable/copier.go` line 178 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsugSMQ88cKi7OJAJrD:-NsugSMQ88cKi7OJAJrE:blpvxxg) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L178)):*
> ```Go
> 				alloc, entry.key.UserKey = alloc.Copy(entry.key.UserKey)
> 				res = append(res, entry)
> 				if base.InternalCompare(r.Compare, end, *key) < 1 {
> ```

[nit] `<= 0` is what we use almost everywhere

RaduBerinde:
*[`sstable/copier.go` line 79 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsufjUR8zYLjtzT2wgK:-NsuiyTu6Xhri73qj2It:bquedoz) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L79)):*

[nit] maybe an invariants assertion that the sum of the lengths matches?
</blockquote></details>

I guess we don't really care about that (as we discussed offline).. Just that `blocks[0].bh.Offset` is the minimum offset and `blocks[len(blocks)-1].bh.Offset` is the maximum one.

dt (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3406)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`sstable/copier.go` line 79 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsufjUR8zYLjtzT2wgK:-Nsuwidk4BYw2xSYvHya:bxyul1j) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L79)):*

I guess we don't really care about that (as we discussed offline).. Just that `blocks[0].bh.Offset` is the minimum offset and `blocks[len(blocks)-1].bh.Offset` is the maximum one.
</blockquote></details>

Not sure if I'm following if you're suggesting there is something to check here? The _check_ seems like it would just be.. this same line, right?
___
*[`sstable/copier.go` line 97 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsublLJ03SZSDk3fiL6:-Nsuwvl98n8yzJP98xKQ:b7x28g1) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L97)):*

[it] For now we should check and error out if we have these kinds of blocks.
</blockquote></details>

We already did at the top of the function (based on the reader props). I jus have the TODOs here since this is where I was gonna come add it.
___
*[`sstable/copier.go` line 127 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsug-DTBAEnlr-k9zvX:-Nsux3iI7_V13svNN360:b-fvbkin) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L127)):*

[nit] should this be `dataBlockInfo` so it's more descriptive?
</blockquote></details>

I donno, the sstable format refers to these as the "index" (as distinct from the metaindex) and each instance of this struct corresponds to an index entry -- a sep and handle. I added a comment, but "dataBlockInfo" feels like new terminology to me vs sticking with the convention?
___
*[`sstable/copier.go` line 129 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsug6ww1CCZQ6wx4tEE:-NsuxYdvEaYJGWgG_egG:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L129)):*

[nit] a comment about the key would be good (it's the "separator" key)
</blockquote></details>

Done.
___
*[`sstable/copier.go` line 132 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsucCcTBo6ZvcqjDEhx:-NsuxamQCvE7Jh3lsUvZ:bmg24b6) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L132)):*

[nit] Worth adding a function comment. Also not sure the name is very good, maybe `getDataBlocksIntersectingRange`?
</blockquote></details>

Done.

I don't like "getDataBlocks" since I'm not gettin blocks -- I'm getting index entries, i.e. block _handles_, not blocks.
___
*[`sstable/copier.go` line 157 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsud7wI69iCHNICBENl:-NsuyK8H2Vm4HXrf7Qe4:bm3oo0y) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L157)):*

This part very closely follows the outer part (as it should).. We could extract a helper function `forEachBlockInRange(.. func(bh))` that we can use twice
</blockquote></details>

We do the two loops everywhere: estimate disk usage, layout, etc. If we want to make a shared helper, I'd rather do it as a standalone refractor and use it in all of them.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3406)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`sstable/copier.go` line 79 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsufjUR8zYLjtzT2wgK:-NsvEaPJE7wozmZBuQlR:b-3iz7sl) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L79)):*

Not sure if I'm following if you're suggesting there is something to check here? The _check_ seems like it would just be.. this same line, right?
</blockquote></details>

I'm suggesting verifying that the block offsets are increasing.
___
*[`sstable/copier.go` line 127 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsug-DTBAEnlr-k9zvX:-NsvF5EqAXD5u7JwGi4H:b4c9lyr) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L127)):*

I donno, the sstable format refers to these as the "index" (as distinct from the metaindex) and each instance of this struct corresponds to an index entry -- a sep and handle. I added a comment, but "dataBlockInfo" feels like new terminology to me vs sticking with the convention?
</blockquote></details>

:thumbsup: 
___
*[`sstable/copier.go` line 132 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsucCcTBo6ZvcqjDEhx:-NsvF7uj7xdSL1ef-Qjv:bn1wz0t) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L132)):*

Done.

I don't like "getDataBlocks" since I'm not gettin blocks -- I'm getting index entries, i.e. block _handles_, not blocks.
</blockquote></details>

:thumbsup:
___
*[`sstable/copier.go` line 103 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsvFeJA2OdTdJlS1ywN:-NsvFeJA2OdTdJlS1ywO:b-feb8po) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/sstable/copier.go#L103)):*
> ```Go
> 		return 0, err
> 	}
> 	if len(blocks) > 0 {
> ```

I found that some things break if we try to write out empty SSTs; maybe the 0 blocks case should be an error.. But it is a legitimate case in principle.. Maybe we should just pick out the first data block in that case(?)  In any case, this path should be tested.
___
*[`sstable/copier.go` line 106 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsvFGgN95Wbn_2F4XJB:-NsvFGgN95Wbn_2F4XJC:b-nc0mtt) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/sstable/copier.go#L106)):*
> ```Go
> 		offset := blocks[0].bh.Offset
> 		// The block lengths don't include their trailers, which just sit after the
> 		// block length, before the next offset; We get he ones between the blocks
> ```

the ones

Would be good to add an overview of what we're doing: "We are going to copy byte-for-byte the section of the input file that contains all the data blocks we care about, and then we will rebuild the index blocks."

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3406)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`sstable/copier.go` line 103 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsvFeJA2OdTdJlS1ywN:-NsxQJ0I-egNdh2ZfW8U:bpgpf7s) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/sstable/copier.go#L103)):*

Just include a block that we know we don't need and will ignore?  

I guess I'm not sure what the things that break are, but if they'd also break when we read an sst that is empty due to and only to virtual bounds, then I don't think it matters what we include here.

If we really can't write empty stables (I think we should be allowed to, so I'm surprised we think we can't) then I'd rather error here and make the caller (compaction.go) figure out how to just bubble up the lack of an sst instead of making a bogus sst with the wrong (according to what we were asked for) content.
</blockquote></details>

Yeah so I ran into this when doing the metamorphic testing for external ingest. Here's a relevant comment: https://github.com/cockroachdb/pebble/blob/9591f9384d50c08c7f525b0c07e134d792e7c378/metamorphic/ops.go#L949

So yes, I think things would break anyway if there are no keys in the range (I think it's the level iterator that assumes that once it reaches a file, it will produce a key). So it would be ok to just error out here in that case.

The other failure I found was that if you had an empty sst and did a `SeekGE` on it, you would sometimes get panics, see discussion here: https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1706130530093809?thread_ts=1706036242.466479&cid=CAC6K3SLU

We'd need to fix both of these issues before trying to write out and use an empty sst.

I filed issues https://github.com/cockroachdb/pebble/issues/3409 and https://github.com/cockroachdb/pebble/issues/3408 to track these, you can reference them in a comment.

itsbilal (commented):
Mostly looks good to me, just some minor comments

> Reviewed 2 of 4 files at r1, 2 of 3 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3406#-:-Nsy6vzd9L08YO8DGjR-:b-bma04)* status: all files reviewed, 9 unresolved discussions (waiting on @dt)

___
*[`compaction.go` line 2512 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsxlZbGElIXYguK2vER:-NsxlZbGElIXYguK2vES:bcdoe66) ([raw file](https://github.com/cockroachdb/pebble/blob/e6d0a4200fad928382507a46d94baf80145d2602/compaction.go#L2512)):*
> ```Go
> 		pendingOutputs = append(pendingOutputs, newMeta)
> 
> 		if err := objstorage.Copy(ctx, src, w, 0, uint64(src.Size())); err != nil {
> ```

should we maintain an objstorage.Copy version of this for when `sstable.CopySpan` returns an error? Otherwise we're silently losing the ability to copy-compact files with range keys. Currently this "never happens" in cockroach but I would at least like this to be commented somewhere (Eg. in runCopyCompaction's function comment), if we aren't doing the fallback thing.
___
*[`compaction.go` line 2528 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsxkpTPBYs30Kbn1W7Q:-NsxkpTPBYs30Kbn1W7R:bxrwdkh) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/compaction.go#L2528)):*
> ```Go
> 			start, end,
> 		)
> 		src = nil // We passed src to CopySpan; its responsible for closing it.
> ```

nit: it's
___
*[`sstable/copier.go` line 49 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsxmeUABGPtu-mRyO1b:-NsxmeUABGPtu-mRyO1c:b9m35s6) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/sstable/copier.go#L49)):*
> ```Go
> 
> 	// If our input has not filters, our output cannot have filters either.
> 	if r.filterBH.Offset == 0 {
> ```

should we do `r.tableFilter == nil` for consistency with the check down below?
___
*[`sstable/copier.go` line 64 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsxlxm2913v10a1k_xk:-Nsxlxm2913v10a1k_xl:blpd7fy) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/sstable/copier.go#L64)):*
> ```Go
> 
> 	if r.Properties.NumValueBlocks > 0 || r.Properties.NumRangeKeys() > 0 {
> 		return 0, errors.New("sstable value blocks or range keys")
> ```

nit: "cannot CopySpan sstables with value blocks or range keys"
___
*[`sstable/copier.go` line 172 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsy5Rj32cAzVN5QageP:-Nsy5Rj32cAzVN5QageQ:b-w2q5xq) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/sstable/copier.go#L172)):*
> ```Go
> 			res = append(res, entry)
> 		} else {
> 			subBlk, err := r.readBlock(ctx, bh.BlockHandle, nil, nil, nil, nil, nil)
> ```

if we're doing repeat readBlocks, we could benefit from a readHandle. Let's just do an `rh = r.readable.NewReadHandle()`at the top (possibly only if we're in the twoLevelIndex case) and pass that in here.
___
*[`sstable/copier.go` line 197 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsy6XCq--F8iHP3WEty:-Nsy6XCr2NbKcXURlrto:b-jghju4) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/sstable/copier.go#L197)):*
> ```Go
> 				}
> 			}
> 			if err := sub.Error(); err != nil {
> ```

nit: blockIter is infallible so we can skip this
___
*[`data_test.go` line 515 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsxmHCR48nGCBu1xIJ6:-NsxmHCR48nGCBu1xIJ7:b-6op7vs) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/data_test.go#L515)):*
> ```Go
> 			switch cmdArg.Vals[0] {
> 			case "1":
> 				blockSize = 1
> ```

why not `blockSize, err = strconv.ParseInt(cmdArg.Vals[0], 10, 64)` ?

dt (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3406)* status: 3 of 5 files reviewed, 5 unresolved discussions (waiting on @itsbilal)

___
*[`compaction.go` line 2512 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsxlZbGElIXYguK2vER:-Nsyoo3eA3PG9BwZa4Di:bkhkdo9) ([raw file](https://github.com/cockroachdb/pebble/blob/e6d0a4200fad928382507a46d94baf80145d2602/compaction.go#L2512)):*

should we maintain an objstorage.Copy version of this for when `sstable.CopySpan` returns an error? Otherwise we're silently losing the ability to copy-compact files with range keys. Currently this "never happens" in cockroach but I would at least like this to be commented somewhere (Eg. in runCopyCompaction's function comment), if we aren't doing the fallback thing.
</blockquote></details>

Not silently -- it fails loudly.

IMO we never had this ability. Copy compactions already had significant limitations on them before we added the limited ability we added (only local to shared) so I'm fine with saying external to local has this restriction. Comment updated.
___
*[`sstable/copier.go` line 79 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsufjUR8zYLjtzT2wgK:-NsylSYs7Ney0WfSNnd_:b4iu2ss) ([raw file](https://github.com/cockroachdb/pebble/blob/21d26541eac3803eb8ee65fb102d4ee8f4e11494/sstable/copier.go#L79)):*

I'm suggesting verifying that the block offsets are increasing.
</blockquote></details>

Done. I think it is worth checking all the time since one int compare is nothing compared (har har) to the copy we're about to do, and we'd rather know if we goofed. It'd be silent otherwise thanks of uint wrap around.
___
*[`sstable/copier.go` line 49 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsxmeUABGPtu-mRyO1b:-Nsyjt0zCpwkpsr78KP1:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/sstable/copier.go#L49)):*

should we do `r.tableFilter == nil` for consistency with the check down below?
</blockquote></details>

Done.
___
*[`sstable/copier.go` line 103 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsvFeJA2OdTdJlS1ywN:-NsyoigH2iz1gMPkWJZc:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/sstable/copier.go#L103)):*

Yeah so I ran into this when doing the metamorphic testing for external ingest. Here's a relevant comment: https://github.com/cockroachdb/pebble/blob/9591f9384d50c08c7f525b0c07e134d792e7c378/metamorphic/ops.go#L949

So yes, I think things would break anyway if there are no keys in the range (I think it's the level iterator that assumes that once it reaches a file, it will produce a key). So it would be ok to just error out here in that case.

The other failure I found was that if you had an empty sst and did a `SeekGE` on it, you would sometimes get panics, see discussion here: https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1706130530093809?thread_ts=1706036242.466479&cid=CAC6K3SLU

We'd need to fix both of these issues before trying to write out and use an empty sst.

I filed issues https://github.com/cockroachdb/pebble/issues/3409 and https://github.com/cockroachdb/pebble/issues/3408 to track these, you can reference them in a comment.
</blockquote></details>

Done.
___
*[`sstable/copier.go` line 172 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsy5Rj32cAzVN5QageP:-NsyfG4M1H6NOqmKO-Ll:buylzyp) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/sstable/copier.go#L172)):*

if we're doing repeat readBlocks, we could benefit from a readHandle. Let's just do an `rh = r.readable.NewReadHandle()`at the top (possibly only if we're in the twoLevelIndex case) and pass that in here.
</blockquote></details>

For _index_ blocks? I suppose
___
*[`sstable/copier.go` line 197 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsy6XCq--F8iHP3WEty:-NsyoTKj5iKgVOs8VPCw:b-qc0mma) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/sstable/copier.go#L197)):*

nit: blockIter is infallible so we can skip this
</blockquote></details>

hardly seems worth skipping a single nil check compared to to the copy we're about to do?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3406)* status: 1 of 5 files reviewed, 6 unresolved discussions (waiting on @dt and @itsbilal)

___
*[`sstable/copier.go` line 63 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NtE7XrzAlGm-Z3phODP:-NtE7XrzAlGm-Z3phODQ:b-9lgj8q) ([raw file](https://github.com/cockroachdb/pebble/blob/8caef20c4ef56da6c5fcba22c97f5800a1a24685/sstable/copier.go#L63)):*
> ```Go
> 	}()
> 
> 	if r.Properties.NumValueBlocks > 0 || r.Properties.NumRangeKeys() > 0 {
> ```

We need to check that there are no RangeDels either right? Currently the metamorphic test does generate those (maybe we can check for those upfront and not use CopySpan in that case?)

dt:
Indeed!

Maybe I'll just copy all those blocks wholesale, as-is, and unfiltered -- all zero of them in real backups for now?

RaduBerinde:
SGTM!

RaduBerinde:
Could you try this one line diff with online restore? Seems to work from the Pebble side.

```
--- a/sstable/copier.go
+++ b/sstable/copier.go
@@ -51,6 +51,12 @@ func CopySpan(
        }
        o.TableFormat = r.tableFormat
        w := NewWriter(output, o)
+       // We don't want the writer to attempt to write out block property data in
+       // index blocks. This data won't be valid since we're not passing the actual
+       // key data through the writer. The data blocks will still have their
+       // properties unchanged, and we will use the the table level properties from
+       // the original ssts as well.
+       w.blockPropCollectors = nil
```

RaduBerinde:
> Maybe I'll just copy all those blocks wholesale, as-is, and unfiltered -- all zero of them in real backups for now?

This is easier said than done, looks like it would require some refactoring in the writer. We can just disallow rangedels in external files for now and worry about it later,

dt:
> just disallow rangedels 

heh, just pushed that right before you commented 

> require some refactoring in the writer

Indeed: 
```
git stash show --stat
 sstable/writer.go | 161 ++++++++++++++++++++++++++++++------------------------
 1 file changed, 90 insertions(+), 71 deletions(-)
```

compaction.go
```diff
@@ -2509,13 +2516,34 @@ func (d *DB) runCopyCompaction(
 		}
 		pendingOutputs = append(pendingOutputs, newMeta)
 
-		if err := objstorage.Copy(ctx, src, w, 0, uint64(src.Size())); err != nil {
-			w.Abort()
-			return ve, pendingOutputs, err
-		}
-		if err := w.Finish(); err != nil {
-			return ve, pendingOutputs, err
+		if d.opts.Experimental.DownloadViaSpaCopy {
```
msbutler, line 2519: I suspect a sizable number of externalSSTs will exactly match the physical sst. Could we add a fast path where we do an old style byte for byte copy if there's an exact match? This could be a follow up PR. I also wonder how much overhead this block copy approach takes-- if it's small, this extra fast path may not be worth it.

msbutler (approved):

testdata/ingest_external
```diff
@@ -323,21 +327,21 @@ gg: (foo, .)
 gh: (foo, .)
 gi: (foo, .)
 
-lsm
+lsm verbose
 ----
 L6:
-  000004(000004):[gc#10,DELSIZED-gf#inf,RANGEDEL]
-  000005(000005):[gg#11,DELSIZED-gj#inf,RANGEDEL]
+  000004(000004):[gc#10,DELSIZED-gf#inf,RANGEDEL] seqnums:[10-10] points:[gc#10,DELSIZED-gf#inf,RANGEDEL] size:979
```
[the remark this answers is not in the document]
dt: The relevant bit here is `size`: without span copy the size is the same before and after download, with it we see the size of our multi-block sst drop by the two blocks were not copying.

compaction.go
RaduBerinde: I like this idea

dt: > I like this idea

Can you expand more on why?

I certainly don't think we should add it now, since it is strictly an optimization that we should let benchmarking tell us to add.

Unless we are only using some tiny number of blocks, the cost of the block copy is going to dominate the cost of building and writing the index. If we're copying a tiny number of blocks from a huge file, this block-copy+build-index is extremely worthwhile. If we're copying all the blocks from a large file, the added overhead of building an index vs copying it is negligible. If we're copying all the blocks from a tiny file, then sure, the overhead of building a new index _relative to the copy_ is something, but in the big picture of an online restore is inconsequential.

dt:
@RaduBerinde 

> try this one line diff

Looking good! My big restore has been downloading for the last 6 hours and no consistency failures so far.

I added it to the PR, and moved the new functionality behind an experimental feature flag.

compaction.go
RaduBerinde: No strong opinion when we should add it. I like it because it's very simple to implement (e.g. if the slice of blocks we need to copy is > 90% of the object, just copy everything) - much less work than it would take to benchmark things. I don't have data to go on so I don't know that the building of the index blocks is negligible. In aggregate, it might add a non-trivial CPU component to the process.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3406#-:-NtMeXkO2YNrn-AGEYnt:bnfp4nl)* status: 1 of 7 files reviewed, 9 unresolved discussions (waiting on @dt, @itsbilal, and @msbutler)

___
*[`options.go` line 789 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NtMeWaK8ciXu5-TI-rB:-NtMeWaL-mMAOtSTBH8h:b8s3r8) ([raw file](https://github.com/cockroachdb/pebble/blob/b76054aae6723fe6421af2154f305cb67fefc035/options.go#L789)):*
> ```Go
> 		// DownloadViaSpaCopy downloads backing files during Download() calls via
> 		// span copy.
> 		DownloadViaSpaCopy bool
> ```

SpanCopy
___
*[`testdata/ingest_external` line 333 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3406#gh-1529102940:-NtMeL8X3uY80d_mbeQT:b-l0ve2c) ([raw file](https://github.com/cockroachdb/pebble/blob/8caef20c4ef56da6c5fcba22c97f5800a1a24685/testdata/ingest_external#L333)):*

The relevant bit here is `size`: without span copy the size is the same before and after download, with it we see the size of our multi-block sst drop by the two blocks were not copying.
</blockquote></details>

Worth adding a comment saying that we should see the size go down after download.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3406)* status: 1 of 7 files reviewed, 9 unresolved discussions (waiting on @dt, @itsbilal, and @msbutler)

___
*[`sstable/copier.go` line 63 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NtE7XrzAlGm-Z3phODP:-NtMiVZb9mOYoa7RZoll:b-j51to) ([raw file](https://github.com/cockroachdb/pebble/blob/8caef20c4ef56da6c5fcba22c97f5800a1a24685/sstable/copier.go#L63)):*

We need to check that there are no RangeDels either right? Currently the metamorphic test does generate those (maybe we can check for those upfront and not use CopySpan in that case?)
</blockquote></details>

Could we make this fall back on the regular Copy() path? If not, to merge this we need to disallow `IngestExternal` when there are range dels and modify the meta test to not generate those:
```diff
--- a/metamorphic/generator.go
+++ b/metamorphic/generator.go
@@ -1056,7 +1056,7 @@ func (g *generator) newExternalObj() {
                batchID = g.liveBatches.rand(g.rng)
                okm := g.keyManager.objKeyMeta(batchID)
                // #3287: IngestExternalFiles currently doesn't support range keys.
-               if !okm.bounds.IsUnset() && !okm.hasRangeKeys {
+               if !okm.bounds.IsUnset() && !okm.hasRangeKeys && !okm.hasRangeDels {
```

dt (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3406)* status: 0 of 7 files reviewed, 8 unresolved discussions (waiting on @itsbilal, @msbutler, and @RaduBerinde)

___
*[`compaction.go` line 2519 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3406#gh-1530497494:-NtMoW-W4xZTwearBYSt:b-dqmxzz) ([raw file](https://github.com/cockroachdb/pebble/blob/b76054aae6723fe6421af2154f305cb67fefc035/compaction.go#L2519)):*

No strong opinion when we should add it. I like it because it's very simple to implement (e.g. if the slice of blocks we need to copy is > 90% of the object, just copy everything) - much less work than it would take to benchmark things. I don't have data to go on so I don't know that the building of the index blocks is negligible. In aggregate, it might add a non-trivial CPU component to the process.
</blockquote></details>

Eh, I still think I want to see it in a profile before I add complexity and potential different cases. The idea that some files have transformed props and some don't seems like more than I want to consider at least until the common case is well understood and battle-tested.
___
*[`sstable/copier.go` line 172 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-Nsy5Rj32cAzVN5QageP:-NtJwHavECpPvNPHodK3:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/sstable/copier.go#L172)):*

For _index_ blocks? I suppose
</blockquote></details>

Done.
___
*[`sstable/copier.go` line 63 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NtE7XrzAlGm-Z3phODP:-NtJwGIw1ghJAKhm42Qd:b-4ffxp2) ([raw file](https://github.com/cockroachdb/pebble/blob/8caef20c4ef56da6c5fcba22c97f5800a1a24685/sstable/copier.go#L63)):*

Could we make this fall back on the regular Copy() path? If not, to merge this we need to disallow `IngestExternal` when there are range dels and modify the meta test to not generate those:
```diff
--- a/metamorphic/generator.go
+++ b/metamorphic/generator.go
@@ -1056,7 +1056,7 @@ func (g *generator) newExternalObj() {
                batchID = g.liveBatches.rand(g.rng)
                okm := g.keyManager.objKeyMeta(batchID)
                // #3287: IngestExternalFiles currently doesn't support range keys.
-               if !okm.bounds.IsUnset() && !okm.hasRangeKeys {
+               if !okm.bounds.IsUnset() && !okm.hasRangeKeys && !okm.hasRangeDels {
```
</blockquote></details>

I added a copy of whole file in this case, via a function call that we can spot in profiles to know if it is happening.
___
*[`testdata/ingest_external` line 333 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3406#gh-1529102940:-NtMouIi3SvvgA5qpdL9:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8caef20c4ef56da6c5fcba22c97f5800a1a24685/testdata/ingest_external#L333)):*

Worth adding a comment saying that we should see the size go down after download.
</blockquote></details>

Done.
___
*[`data_test.go` line 515 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NsxmHCR48nGCBu1xIJ6:-NtJwJU4EF8WpvJN4Enh:b-8ty0d3) ([raw file](https://github.com/cockroachdb/pebble/blob/18676a2b875121256b6b9606347266cd69541c52/data_test.go#L515)):*

why not `blockSize, err = strconv.ParseInt(cmdArg.Vals[0], 10, 64)` ?
</blockquote></details>

Meh, I suppose

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3406#-:-NtNN_et-GlSNYx_NVuW:bnfp4nl)* status: 0 of 7 files reviewed, 8 unresolved discussions (waiting on @dt, @itsbilal, and @msbutler)

___
*[`sstable/copier.go` line 63 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3406#-NtE7XrzAlGm-Z3phODP:-NtNN2SA1Zp3CozwVBIT:b-z0mnjf) ([raw file](https://github.com/cockroachdb/pebble/blob/8caef20c4ef56da6c5fcba22c97f5800a1a24685/sstable/copier.go#L63)):*

I added a copy of whole file in this case, via a function call that we can spot in profiles to know if it is happening.
</blockquote></details>

SGTM. You can remove this check here now.

PR #3407, opened 2024-03-14
db: add local table size metrics

### LINKED ISSUE #3391: db: Metrics.DiskSpaceUsage incorrect with virtual, shared or external files
CockroachDB uses `(*pebble.Metrics).DiskSpaceUsage` to calculate the disk space used by Cockroach within the data directory. Today, the sstable portion of the calculation assumes that the `Size` of all the `fileMetadata`s in the current version accurately reflect the local physical file size. With the introduction of virtual SSTables (eg, created from ingest-time splitting or excises), the underlying physical size of the file may be larger than the logical size of the file metadata. With shared and external files, it may be larger (if the entire backing file is copied locally) or smaller (if the file only exists remotely).

We should adapt Pebble to keep an accurate, incrementally-updated record of the physical size of local sstables and use it in `DiskSpaceUsage`. We must take care to avoid any O(n) work during `GetMetrics()`, instead incrementally updating statistics as version edits are applied.

sumeerbhola:
The Metrics object lacks a total size of local sstables, which causes DiskSpaceUsage to overestimate the local disk space used. The new metrics track live, zombie and obsolete local file sizes.

Fixes #3391


sumeerbhola (commented):
This lacks tests with non-local files. I'll add if the approach looks ok.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3407#-:-NsxRcM5BvOYFBZH5TgM:bt730nz)* status: 0 of 13 files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde (approved):
:lgtm:

It feels like there's a more general problem, where in the `pebble` layer we should have easier access to the `ObjectMetadata` for a `FileMetadata` (ideally they'd be both part of the same structure). But I don't see a nice way to do it given that the container for them is `Version`.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3407#-:-Nsy1YM59AzcJset4Za4:b-yrpphs)* status: 0 of 14 files reviewed, 6 unresolved discussions (waiting on @sumeerbhola)

___
*[`compaction.go` line 2573 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3407#-NsxyzUa2TS4bWWKI4W8:-NsxyzUa2TS4bWWKI4W9:bg2srhj) ([raw file](https://github.com/cockroachdb/pebble/blob/6c071907c10c803d113ee45cfc586fa7cc090a4c/compaction.go#L2573)):*
> ```Go
> }
> 
> type fileMetaEtc struct {
> ```

[nit] maybe `compactionOutput`?
___
*[`compaction.go` line 2574 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3407#-Nsxz0MyAsFDfRlIm78k:-Nsxz0MyAsFDfRlIm78l:ba3u50m) ([raw file](https://github.com/cockroachdb/pebble/blob/6c071907c10c803d113ee45cfc586fa7cc090a4c/compaction.go#L2574)):*
> ```Go
> 
> type fileMetaEtc struct {
> 	*fileMetadata
> ```

Do we need to embed this? I don't think we use it in too many places.
___
*[`version_set.go` line 694 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3407#-Nsy063D09XKqEi1yWrY:-Nsy063D09XKqEi1yWrZ:b-hq7ycc) ([raw file](https://github.com/cockroachdb/pebble/blob/6c071907c10c803d113ee45cfc586fa7cc090a4c/version_set.go#L694)):*
> ```Go
> func getZombiesAndUpdateVirtualBackings(
> 	ve *versionEdit, virtualBackings *manifest.VirtualBackings, provider objstorage.Provider,
> ) (zombies map[base.DiskFileNum]tableInfo, localLiveSizeDelta int64) {
> ```

I think this will be easier to do with the small refactoring in https://github.com/cockroachdb/pebble/pull/3400 (we can just go through `removedPhysicalBackings`)
___
*[`version_set.go` line 768 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3407#-Nsy-_5B7p_ky2DhYdDQ:-Nsy-_5B7p_ky2DhYdDR:b-yklkws) ([raw file](https://github.com/cockroachdb/pebble/blob/6c071907c10c803d113ee45cfc586fa7cc090a4c/version_set.go#L768)):*
> ```Go
> }
> 
> func computeLocalSize(
> ```

[nit] maybe `sizeIfLocal`. Also add a comment, e.g. "returns backing.Size if the backing is a local file or 0 otherwise"
___
*[`version_set.go` line 773 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3407#-Nsy0PF_7MYmp4aJkRec:-Nsy0PF_7MYmp4aJkRed:b-vznuzg) ([raw file](https://github.com/cockroachdb/pebble/blob/6c071907c10c803d113ee45cfc586fa7cc090a4c/version_set.go#L773)):*
> ```Go
> 	meta, err := provider.Lookup(fileTypeTable, backing.DiskFileNum)
> 	if err != nil {
> 		panic(errors.Wrapf(err, "disk file %s not found", backing.DiskFileNum))
> ```

This error can be misleading, it's not really the file itself. I think the provider error already has all the information (`file %s (type %d) unknown to the objstorage provider`).

I would add an `objstorage.MustLookupTable(provider, diskFileNum)` and `objstorage.IsLocal(provider, diskFileNum)` convenience wrappers. I think we'd use them in more places.
___
*[`version_set.go` line 1003 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3407#-Nsy1BQTBxufIhQKVH4Y:-Nsy1BQTBxufIhQKVH4Z:b-olu16n) ([raw file](https://github.com/cockroachdb/pebble/blob/6c071907c10c803d113ee45cfc586fa7cc090a4c/version_set.go#L1003)):*
> ```Go
> 		// deleted from disk.
> 		//
> 		// TODO(sumeer): this means that the zombie metrics, like ZombieSize,
> ```

Good point.. I think it probably doesn't matter much in practice unless the deleter falls behind

sumeerbhola (commented):
TFTR!

> It feels like there's a more general problem, where in the `pebble` layer we should have easier access to the `ObjectMetadata` for a `FileMetadata` (ideally they'd be both part of the same structure).

Our current decomposition seems reasonable. The provider manifest has an object added before the Pebble manifest, and removed from the provider manifest after removal from the Pebble manifest. And we can do a cheap join between those two when an object is still known to Pebble.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3407#-:-Nt1HHZ55YK1cxZ90CnU:bo0hddz)* status: 0 of 15 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

___
*[`compaction.go` line 2573 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3407#-NsxyzUa2TS4bWWKI4W8:-Nt1J2UmAyznqbMd9E6-:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/6c071907c10c803d113ee45cfc586fa7cc090a4c/compaction.go#L2573)):*

[nit] maybe `compactionOutput`?
</blockquote></details>

Done
___
*[`compaction.go` line 2574 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3407#-Nsxz0MyAsFDfRlIm78k:-Nt1J3Az9w-VpKUQJz3F:bcheknm) ([raw file](https://github.com/cockroachdb/pebble/blob/6c071907c10c803d113ee45cfc586fa7cc090a4c/compaction.go#L2574)):*

Do we need to embed this? I don't think we use it in too many places.
</blockquote></details>

Removed embedding
___
*[`version_set.go` line 694 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3407#-Nsy063D09XKqEi1yWrY:-Nt1JB8t630-eng2htHY:b-isfvxu) ([raw file](https://github.com/cockroachdb/pebble/blob/6c071907c10c803d113ee45cfc586fa7cc090a4c/version_set.go#L694)):*

I think this will be easier to do with the small refactoring in https://github.com/cockroachdb/pebble/pull/3400 (we can just go through `removedPhysicalBackings`)
</blockquote></details>

Ack. I'll wait until that settles, since it seems to me that we don't need `removedPhysicalBackings` either.
___
*[`version_set.go` line 768 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3407#-Nsy-_5B7p_ky2DhYdDQ:-Nt1Jq5rBvqBctweMhXu:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/6c071907c10c803d113ee45cfc586fa7cc090a4c/version_set.go#L768)):*

[nit] maybe `sizeIfLocal`. Also add a comment, e.g. "returns backing.Size if the backing is a local file or 0 otherwise"
</blockquote></details>

Done
___
*[`version_set.go` line 773 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3407#-Nsy0PF_7MYmp4aJkRec:-Nt1QE5J4QNxlqg5dVcH:b-r8fm8o) ([raw file](https://github.com/cockroachdb/pebble/blob/6c071907c10c803d113ee45cfc586fa7cc090a4c/version_set.go#L773)):*
> This error can be misleading, it's not really the file itself. I think the provider error already has all the information (`file %s (type %d) unknown to the objstorage provider`).

Removed the wrapping

> I would add an `objstorage.MustLookupTable(provider, diskFileNum)` and `objstorage.IsLocal(provider, diskFileNum)` convenience wrappers. I think we'd use them in more places.

Added `objstorage.MustIsLocalTable` and used here and in one other place in compaction.go.
___
*[`version_set.go` line 1003 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3407#-Nsy1BQTBxufIhQKVH4Y:-Nt1S4md7oHArv3Ocb1K:bqog80t) ([raw file](https://github.com/cockroachdb/pebble/blob/6c071907c10c803d113ee45cfc586fa7cc090a4c/version_set.go#L1003)):*

Good point.. I think it probably doesn't matter much in practice unless the deleter falls behind
</blockquote></details>

Agreed

sumeerbhola (commented):
I've added a test. I'll rebase on your PR when it is merged.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3407#-:-NtDgZB_0ZxLSVZMGpaW:b-ldde9q)* status: 0 of 15 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3407)* status: 0 of 16 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`version_set.go` line 748 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3407#-NtHcfE-9ZKqkF7Yqv9b:-NtHj3qp6VJm4RcljL4g:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/1460fc593c6e1e2fab4bd9a11b7d13066bc80dd3/version_set.go#L748)):*

[nit] Feels like this belongs in the loop below, next to `AddAndRef`
</blockquote></details>

Done

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3407#-:-NtHkZGzCp3OnTYNJEfv:bnfp4nl)* status: 0 of 16 files reviewed, all discussions resolved

sumeerbhola:
created https://github.com/cockroachdb/pebble/issues/3421 for flaky test

sumeerbhola:
TFTR!

PR #3410, opened 2024-03-14
wal: allow runtime toggling of WAL failover

jbowens:
This commit adds a new bool return value to the option func UnhealthyOperationLatencyThreshold that allows the client to dynamically toggle whether failover is permitted. This will be used by Cockroach to prevent failover until cluster version finalization begins.

Informs #3230.


jbowens:
TFTR!

PR #3412, opened 2024-03-15
ingest: expose inclusive bounds for ExternalFile end keys

stevendanna:
An excise can result in an external file having an inclusive bounds. But, previously we would erroneously expose that as an exclusive bounds.

Here, we add a boolean that indicates whether the Bounds on and ExternalFile are inclusive.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -182,17 +182,17 @@ func ingestLoad1External(
 		return nil, errors.New("pebble: cannot ingest external file with no point or range keys")
 	}
 
-	if opts.Comparer.Compare(e.Bounds.Start, e.Bounds.End) > 0 {
-		return nil, errors.Newf("pebble: external file bounds [%q, %q) are invalid", e.Bounds.Start, e.Bounds.End)
+	if opts.Comparer.Compare(e.StartKey, e.EndKey) > 0 {
+		return nil, errors.Newf("pebble: external file bounds [%q, %q) are invalid", e.StartKey, e.EndKey)
 	}
-	if opts.Comparer.Compare(e.Bounds.Start, e.Bounds.End) == 0 && !e.BoundsHasInclusiveEndKey {
-		return nil, errors.Newf("pebble: external file bounds [%q, %q) are invalid", e.Bounds.Start, e.Bounds.End)
+	if opts.Comparer.Compare(e.StartKey, e.EndKey) == 0 && !e.EndKeyIsInclusive {
+		return nil, errors.Newf("pebble: external file bounds [%q, %q) are invalid", e.StartKey, e.EndKey)
 	}
-	if n := opts.Comparer.Split(e.Bounds.Start); n != len(e.Bounds.Start) {
-		return nil, errors.Newf("pebble: external file bounds start key %q has suffix", e.Bounds.Start)
+	if n := opts.Comparer.Split(e.StartKey); n != len(e.StartKey) {
+		return nil, errors.Newf("pebble: external file bounds start key %q has suffix", e.StartKey)
 	}
-	if n := opts.Comparer.Split(e.Bounds.End); n != len(e.Bounds.End) {
-		return nil, errors.Newf("pebble: external file bounds end key %q has suffix", e.Bounds.End)
+	if n := opts.Comparer.Split(e.EndKey); n != len(e.EndKey) {
+		return nil, errors.Newf("pebble: external file bounds end key %q has suffix", e.EndKey)
 	}
 
 	// #3287: range keys don't yet work correctly when the range key bounds are not tight.
@@ -219,10 +219,13 @@ func ingestLoad1External(
 	// In the name of keeping this ingestion as fast as possible, we avoid
 	// *all* existence checks and synthesize a file metadata with smallest/largest
 	// keys that overlap whatever the passed-in span was.
-	smallestCopy := slices.Clone(e.Bounds.Start)
-	largestCopy := slices.Clone(e.Bounds.End)
+	smallestCopy := slices.Clone(e.StartKey)
+	largestCopy := slices.Clone(e.EndKey)
 	if e.HasPointKey {
-		if e.BoundsHasInclusiveEndKey {
+		// Sequence numbers are updated later by
+		// ingestUpdateSeqNum, applying a squence number that
+		// is applied to all keys in the sstable.
+		if e.EndKeyIsInclusive {
 			meta.ExtendPointKeyBounds(
 				opts.Comparer.Compare,
 				base.MakeInternalKey(smallestCopy, 0, InternalKeyKindMax),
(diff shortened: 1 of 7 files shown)


ingest.go
```diff
@@ -218,11 +222,17 @@ func ingestLoad1External(
...
-		meta.ExtendPointKeyBounds(
-			opts.Comparer.Compare,
-			base.MakeInternalKey(smallestCopy, 0, InternalKeyKindMax),
-			base.MakeRangeDeleteSentinelKey(largestCopy),
-		)
+		if e.BoundsHasInclusiveEndKey {
+			meta.ExtendPointKeyBounds(
+				opts.Comparer.Compare,
+				base.MakeInternalKey(smallestCopy, 0, InternalKeyKindMax),
+				base.MakeInternalKey(largestCopy, 0, InternalKeyKindMax))
```
stevendanna, line 229: Wasn't sure what the best `Kind` would be here.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3412)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @bilal)

___
*[`ingest.go` line 229 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3412#gh-1526365635:-Nt1VOq251HQzvmGfsDA:bj9bg1f) ([raw file](https://github.com/cockroachdb/pebble/blob/d3692290cfea8b2033a742907ca52cbaa42519c4/ingest.go#L229)):*
<details><summary><i>Previously, stevendanna (Steven Danna) wrote…</i></summary><blockquote>

Wasn't sure what the best `Kind` would be here.
</blockquote></details>

We should make the Kind 0.

The trailers of the keys are `(seqNum << 8 + kind)` (see MakeInternalKey) and they sort in reverse order (see InternalCompare). The largest key needs to sort after any valid keys with that user key, so 0 would make that happen.

I believe the sequence numbers in these keys will get overwritten somewhere with the one sequence number that we'll assign to the entire table (@bilal is that correct?)

RaduBerinde (approved):
:lgtm:

I will update the meta test separately to generate inclusive bounds too. We don't need to do it in this PR (it might need some debugging if it finds an issue).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3412#-:-Nt1YOGd2gCTZLUrX46t:bag2nmh)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @bilal)

itsbilal (approved):
:LGTM: save for the kind change

(ps. I'm @itsbilal, `bilal` is someone else)

> Reviewed 5 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3412#-:-Nt1a6M2APN0jIgv9FXm:b-g973ck)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @bilal and @RaduBerinde)

___
*[`ingest.go` line 229 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3412#gh-1526365635:-Nt1aHqzCKfYGgdxRTD7:b-slhyyl) ([raw file](https://github.com/cockroachdb/pebble/blob/d3692290cfea8b2033a742907ca52cbaa42519c4/ingest.go#L229)):*

We should make the Kind 0.

The trailers of the keys are `(seqNum << 8 + kind)` (see MakeInternalKey) and they sort in reverse order (see InternalCompare). The largest key needs to sort after any valid keys with that user key, so 0 would make that happen.

I believe the sequence numbers in these keys will get overwritten somewhere with the one sequence number that we'll assign to the entire table (@bilal is that correct?)
</blockquote></details>

Yeah this is correct, we should use the kind 0. I'm fine with the literal 0 or `InternalKeyKindDelete` which is 0.

stevendanna (commented):
Done. Thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3412#-:-Nt1bWxY3zDgP7BlVsuS:boyexb3)* status: 3 of 6 files reviewed, 1 unresolved discussion (waiting on @bilal, @itsbilal, and @RaduBerinde)

RaduBerinde (commented):
Oops, sorry, was on my phone (MacOS is updating haha)

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3412#-:-Nt1c13U2vPT8jQ1I2IC:b8tvvzj)* status: 3 of 6 files reviewed, 1 unresolved discussion (waiting on @bilal and @itsbilal)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3412)* status: 3 of 6 files reviewed, 2 unresolved discussions (waiting on @bilal, @itsbilal, and @stevendanna)

___
*[`ingest.go` line 225 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3412#-Nt1cEjS7Ix6QQI_wRPm:-Nt1cEjS7Ix6QQI_wRPn:b-g8zxsj) ([raw file](https://github.com/cockroachdb/pebble/blob/afbeb53e34ca4fe41d99b2a3f5902a8e03f886c4/ingest.go#L225)):*
> ```Go
> 	largestCopy := slices.Clone(e.Bounds.End)
> 	if e.HasPointKey {
> 		if e.BoundsHasInclusiveEndKey {
> ```

Let's add a comment saying that the sequence numbers will be updated later (and will match the sequence number of all keys in the table).

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3412)* status: 3 of 6 files reviewed, 3 unresolved discussions (waiting on @bilal, @itsbilal, and @stevendanna)

___
*[`ingest.go` line 1146 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3412#-Nt8thU1DLy9wZNMy1HO:-Nt8thU1DLy9wZNMy1HP:bygkc93) ([raw file](https://github.com/cockroachdb/pebble/blob/afbeb53e34ca4fe41d99b2a3f5902a8e03f886c4/ingest.go#L1146)):*
> ```Go
> 	Size uint64
> 
> 	// Bounds of the sstable; the ingestion of this file will only result in keys
> ```

Let's take the keys back out since `KeyRange` is defined as having an exclusive end key. We can just have `StartKey, EndKey, EndKeyIsInclusive`

stevendanna (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3412)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @bilal, @itsbilal, and @RaduBerinde)

___
*[`ingest.go` line 225 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3412#-Nt1cEjS7Ix6QQI_wRPm:-NtFmBIl6mkZVFl9sfZD:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/afbeb53e34ca4fe41d99b2a3f5902a8e03f886c4/ingest.go#L225)):*

Let's add a comment saying that the sequence numbers will be updated later (and will match the sequence number of all keys in the table).
</blockquote></details>

Done.
___
*[`ingest.go` line 1146 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3412#-Nt8thU1DLy9wZNMy1HO:-NtFm6Ox8rlJN0j8iNRB:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/afbeb53e34ca4fe41d99b2a3f5902a8e03f886c4/ingest.go#L1146)):*

Let's take the keys back out since `KeyRange` is defined as having an exclusive end key. We can just have `StartKey, EndKey, EndKeyIsInclusive`
</blockquote></details>

Done.

RaduBerinde:
I started trying to exercise this in the meta test and I am a bit confused when this would be useful. The `EndKey` cannot have a suffix, and the keys inside the table have suffixes. So in what case would the inclusivity actually matter?

PR #3415, opened 2024-03-15
db: add support for checkpointing with remote files

### LINKED ISSUE #3362: db: have Checkpoint support shared/external files
Currently, we don't copy over or link shared/external files when `Checkpoint` is called. This is broadly desirable as Checkpoint is supposed to be a quick operation that is often called as part of a crash-worthy event (eg. replica corruption), and copying over gigabites of sstables from remote storage before crashing sounds undesirable.

As part of this change, we should ensure Pebble also copies over the objprovider Catalog so any references to external/shared files are copied over. Furthermore, we should ensure that we leave any references on shared files dangling after a Checkpoint so those files don't get deleted by other nodes.

itsbilal:
Previously, if we couldn't read a remote file's DiskFileNum locally, we'd error out on the Checkpoint() function even though the store was valid and the file just happened to be remote (shared or external). This change adds support for saving the catalog's state to the checkpoint directory so references to remote files from the checkpoint can still be resolved. To prevent unreffing of shared files, we need to add a dangling reference to them that doesn't get deleted.

Fixes #3362.

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -137,6 +137,11 @@ func mkdirAllAndSyncParents(fs vfs.FS, destDir string) (vfs.File, error) {
 // space overhead for a checkpoint if hard links are disabled. Also beware that
 // even if hard links are used, the space overhead for the checkpoint will
 // increase over time as the DB performs compactions.
+//
+// Note that shared files in a checkpoint could get deleted if the DB is
+// restarted after a checkpoint operation, as the reference for the checkpoint
+// is only maintained in memory. This is okay as long as users of Checkpoint
+// crash shortly afterwards with a "poison file" preventing further restarts.
 func (d *DB) Checkpoint(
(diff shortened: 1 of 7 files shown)


RaduBerinde (approved):
:lgtm:

Thanks for doing this!  We should document somewhere (probably for `db.Checkpoint()`) the limitation that shared objects can go away after the store is reopened.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3415#-:-Nt8ihYr1L3uI4dgmDJ8:bi5aeth)* status: 0 of 7 files reviewed, all discussions resolved

itsbilal:
TFTR! Added the comment. And yeah that's usually okay for the crdb use-case as we'll write a poison file when we crash after a checkpoint. That file will prevent restarts on that store.

PR #3417, opened 2024-03-17
base: add UserKeyBounds

RaduBerinde:
This commit adds `UserKeyBounds` which represents a range of user keys with inclusive start boundary and either inclusive or exclusive range boundary.

This replaces ad-hoc code in various places.

Note: I think there are more places where this can be used and I will look at them as a follow-up, I just didn't want it all in one PR.

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -75,7 +75,8 @@ func excludeFromCheckpoint(f *fileMetadata, opt *checkpointOptions, cmp Compare)
 		return false
 	}
 	for _, s := range opt.restrictToSpans {
-		if f.Overlaps(cmp, s.Start, s.End, true /* exclusiveEnd */) {
+		spanBounds := base.UserKeyBoundsEndExclusive(s.Start, s.End)
+		if f.Overlaps(cmp, &spanBounds) {
 			return false
 		}
 	}
--- a/compaction.go
+++ b/compaction.go
@@ -390,6 +390,10 @@ func (c *compaction) makeInfo(jobID int) CompactionInfo {
 	return info
 }
 
+func (c *compaction) userKeyBounds() base.UserKeyBounds {
+	return base.UserKeyBoundsFromInternal(c.smallest, c.largest)
+}
+
 func newCompaction(
 	pc *pickedCompaction, opts *Options, beganAt time.Time, provider objstorage.Provider,
 ) *compaction {
@@ -422,8 +426,7 @@ func newCompaction(
 	// Compute the set of outputLevel+1 files that overlap this compaction (these
 	// are the grandparent sstables).
 	if c.outputLevel.level+1 < numLevels {
-		c.grandparents = c.version.Overlaps(c.outputLevel.level+1,
-			c.smallest.UserKey, c.largest.UserKey, c.largest.IsExclusiveSentinel())
+		c.grandparents = c.version.Overlaps(c.outputLevel.level+1, c.userKeyBounds())
 	}
 	c.setupInuseKeyRanges()
 	c.kind = pc.kind
@@ -657,8 +660,7 @@ func newFlush(
 	if opts.FlushSplitBytes > 0 {
 		c.maxOutputFileSize = uint64(opts.Level(0).TargetFileSize)
 		c.maxOverlapBytes = maxGrandparentOverlapBytes(opts, 0)
-		c.grandparents = c.version.Overlaps(baseLevel, c.smallest.UserKey,
-			c.largest.UserKey, c.largest.IsExclusiveSentinel())
+		c.grandparents = c.version.Overlaps(baseLevel, c.userKeyBounds())
 		adjustGrandparentOverlapBytesForFlush(c, flushingBytes)
 	}
 
@@ -1866,8 +1868,9 @@ func (d *DB) maybeScheduleDownloadCompaction(env compactionEnv, maxConcurrentCom
 		var externalFile *fileMetadata
 		var err error
 		var level int
+		bounds := base.UserKeyBoundsEndExclusive(download.start, download.end)
 		for i := range v.Levels {
-			overlaps := v.Overlaps(i, download.start, download.end, true /* exclusiveEnd */)
+			overlaps := v.Overlaps(i, bounds)
 			iter := overlaps.Iter()
 			provider := d.objProvider
 			for f := iter.First(); f != nil; f = iter.Next() {
@@ -2256,7 +2259,7 @@ func checkDeleteCompactionHints(
 		// The hint h will be resolved and dropped, regardless of whether
 		// there are any tables that can be deleted.
 		for l := h.tombstoneLevel + 1; l < numLevels; l++ {
-			overlaps := v.Overlaps(l, h.start, h.end, true /* exclusiveEnd */)
+			overlaps := v.Overlaps(l, base.UserKeyBoundsEndExclusive(h.start, h.end))
 			iter := overlaps.Iter()
 			for m := iter.First(); m != nil; m = iter.Next() {
 				if m.IsCompacting() || !h.canDelete(cmp, m, snapshots) || files[m] {
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -226,6 +226,10 @@ type pickedCompaction struct {
 	pickerMetrics compactionPickerMetrics
 }
 
+func (pc *pickedCompaction) userKeyBounds() base.UserKeyBounds {
+	return base.UserKeyBoundsFromInternal(pc.smallest, pc.largest)
+}
+
 func defaultOutputLevel(startLevel, baseLevel int) int {
 	outputLevel := startLevel + 1
 	if startLevel == 0 {
@@ -421,8 +425,7 @@ func (pc *pickedCompaction) setupInputs(
 	// sstables. No need to do this for intra-L0 compactions; outputLevel.files is
 	// left empty for those.
 	if startLevel.level != pc.outputLevel.level {
-		pc.outputLevel.files = pc.version.Overlaps(pc.outputLevel.level, pc.smallest.UserKey,
-			pc.largest.UserKey, pc.largest.IsExclusiveSentinel())
+		pc.outputLevel.files = pc.version.Overlaps(pc.outputLevel.level, pc.userKeyBounds())
 		if anyTablesCompacting(pc.outputLevel.files) {
 			return false
 		}
@@ -515,8 +518,7 @@ func (pc *pickedCompaction) grow(
 	if pc.outputLevel.files.Empty() {
 		return false
 	}
-	grow0 := pc.version.Overlaps(startLevel.level, sm.UserKey,
-		la.UserKey, la.IsExclusiveSentinel())
+	grow0 := pc.version.Overlaps(startLevel.level, base.UserKeyBoundsFromInternal(sm, la))
 	if anyTablesCompacting(grow0) {
 		return false
 	}
@@ -529,8 +531,7 @@ func (pc *pickedCompaction) grow(
 	// We need to include the outputLevel iter because without it, in a multiLevel scenario,
 	// sm1 and la1 could shift the output level keyspace when pc.outputLevel.files is set to grow1.
 	sm1, la1 := manifest.KeyRange(pc.cmp, grow0.Iter(), pc.outputLevel.files.Iter())
-	grow1 := pc.version.Overlaps(pc.outputLevel.level, sm1.UserKey,
-		la1.UserKey, la1.IsExclusiveSentinel())
+	grow1 := pc.version.Overlaps(pc.outputLevel.level, base.UserKeyBoundsFromInternal(sm1, la1))
 	if anyTablesCompacting(grow1) {
 		return false
 	}
@@ -1603,8 +1604,7 @@ func pickAutoLPositive(
 	if pc.startLevel.level == 0 {
 		cmp := opts.Comparer.Compare
 		smallest, largest := manifest.KeyRange(cmp, pc.startLevel.files.Iter())
-		pc.startLevel.files = vers.Overlaps(0, smallest.UserKey,
-			largest.UserKey, largest.IsExclusiveSentinel())
+		pc.startLevel.files = vers.Overlaps(0, base.UserKeyBoundsFromInternal(smallest, largest))
 		if pc.startLevel.files.Empty() {
 			panic("pebble: empty compaction")
 		}
@@ -1811,7 +1811,7 @@ func pickManualCompaction(
 	}
 	pc = newPickedCompaction(opts, vers, manual.level, defaultOutputLevel(manual.level, baseLevel), baseLevel)
 	manual.outputLevel = pc.outputLevel.level
-	pc.startLevel.files = vers.Overlaps(manual.level, manual.start, manual.end, false)
+	pc.startLevel.files = vers.Overlaps(manual.level, base.UserKeyBoundsInclusive(manual.start, manual.end))
 	if pc.startLevel.files.Empty() {
 		// Nothing to do
 		return nil, false
@@ -1897,7 +1897,7 @@ func (p *compactionPickerByScore) pickReadTriggeredCompaction(
 func pickReadTriggeredCompactionHelper(
 	p *compactionPickerByScore, rc *readCompaction, env compactionEnv,
 ) (pc *pickedCompaction) {
-	overlapSlice := p.vers.Overlaps(rc.level, rc.start, rc.end, false /* exclusiveEnd */)
+	overlapSlice := p.vers.Overlaps(rc.level, base.UserKeyBoundsInclusive(rc.start, rc.end))
 	if overlapSlice.Empty() {
 		// If there is no overlap, then the file with the key range
 		// must have been compacted away. So, we don't proceed to
@@ -1929,9 +1929,7 @@ func pickReadTriggeredCompactionHelper(
 	pc.kind = compactionKindRead
 
 	// Prevent read compactions which are too wide.
-	outputOverlaps := pc.version.Overlaps(
-		pc.outputLevel.level, pc.smallest.UserKey,
-		pc.largest.UserKey, pc.largest.IsExclusiveSentinel())
+	outputOverlaps := pc.version.Overlaps(pc.outputLevel.level, pc.userKeyBounds())
 	if outputOverlaps.SizeSum() > pc.maxReadCompactionBytes {
 		return nil
 	}
--- a/compaction_picker_test.go
+++ b/compaction_picker_test.go
@@ -1047,8 +1047,10 @@ func TestPickedCompactionSetupInputs(t *testing.T) {
 				pc.outputLevel.level = pc.startLevel.level + 1
 			}
 			pc.version = newVersion(opts, files)
-			pc.startLevel.files = pc.version.Overlaps(pc.startLevel.level,
-				[]byte(args[0].String()), []byte(args[1].String()), false /* exclusiveEnd */)
+			pc.startLevel.files = pc.version.Overlaps(
+				pc.startLevel.level,
+				base.UserKeyBoundsInclusive([]byte(args[0].String()), []byte(args[1].String())),
+			)
 
 			var isCompacting bool
 			if !pc.setupInputs(opts, availBytes, pc.startLevel) {
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -1211,8 +1211,8 @@ func TestManualCompaction(t *testing.T) {
 		ongoingCompaction.outputLevel = &ongoingCompaction.inputs[1]
 		// Mark files as compacting.
 		curr := d.mu.versions.currentVersion()
-		ongoingCompaction.startLevel.files = curr.Overlaps(startLevel, start, end, false)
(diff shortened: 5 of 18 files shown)


sumeerbhola (changes requested):
Looks great. Some minor comments.

> Reviewed 18 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3417#-:-NtMCz104pmiZl8Oymo3:bm54kgh)* status: all files reviewed, 9 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`flushable.go` line 297 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM9HB4Bo25Hm5BZ_f-:-NtM9HB4Bo25Hm5BZ_f0:b96u0an) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/flushable.go#L297)):*
> ```Go
> 			fileBounds := f.UserKeyBounds()
> 			if fileBounds.End.Applies(s.comparer.Compare, bounds.Start) {
> 				// This file's largest key is larger than bounds.Start. Either the file
> ```

equal to or larger?
___
*[`flushable.go` line 301 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM9PIM4HtjrzWDAQSL:-NtM9PIM4HtjrzWDAQSM:b-3fhwr3) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/flushable.go#L301)):*
> ```Go
> 				// we can stop iterating since the files are sorted. But first,
> 				// determine if there's overlap and call fn if necessary.
> 				if bounds.End.Applies(s.comparer.Compare, fileBounds.Start) {
> ```

How about adding to the comment:
The bounds largest key is greater than equal to the file's smallest key. So the file overlaps ....
___
*[`ingest.go` line 781 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM6vIv4YmJjBxDYDmt:-NtM6vIv4YmJjBxDYDmu:b-p8k1ln) ([raw file](https://github.com/cockroachdb/pebble/blob/3453026711f90e73fe8bff98f16a72adf570b2a2/ingest.go#L781)):*
> ```Go
> 	computeOverlapWithSpans := func(rIter keyspan.FragmentIterator) (bool, error) {
> 		// NB: The spans surfaced by the fragment iterator are non-overlapping.
> 		span, err := rIter.SeekLT(keyRange.smallest.UserKey)
> ```

was this `SeekLT` unnecessary, because it predated https://github.com/cockroachdb/pebble/commit/a80169845786dc35b791d5071a165977c45b80d7?
___
*[`ingest.go` line 790 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM8SPO97ADleZPvPI-:-NtM8SPO97ADleZPvPI0:b-2mdlia) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/ingest.go#L790)):*
> ```Go
> 				// The span starts after our bounds.
> 				return false, nil
> 			}
> ```

SeekGE says
```
	// SeekGE moves the iterator to the first span covering a key greater than
	// or equal to the given key. This is equivalent to seeking to the first
	// span with an end key greater than the given key.
```
so aren't we guaranteed that span.End > bounds.Start, so the following comparison is unnecessary?
___
*[`internal/base/key_bounds.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM2EfqEMiyvWfq8mtN:-NtM2EfqEMiyvWfq8mtO:bootmp4) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/base/key_bounds.go#L62)):*
> ```Go
> 
> // Applies returns true if the given key is "before" the end boundary.
> func (eb UserKeyEndBoundary) Applies(cmp Compare, userKey []byte) bool {
> ```

would `GreaterOrEqual` fully capture the meaning?
___
*[`internal/base/key_bounds.go` line 69 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM2VHi9n6gDJzZG0Qe:-NtM2VHi9n6gDJzZG0Qf:b-ttik4m) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/base/key_bounds.go#L69)):*
> ```Go
> // AppliesToInternalKey returns true if the given internal key is "before" the
> // end boundary.
> func (eb UserKeyEndBoundary) AppliesToInternalKey(cmp Compare, key InternalKey) bool {
> ```

GreaterOrEqualThanInternalKey?
___
*[`internal/base/key_bounds.go` line 153 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM3Rrb2BS4wvWmE1ak:-NtM3Rrb2BS4wvWmE1al:bpafkw7) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/base/key_bounds.go#L153)):*
> ```Go
> }
> 
> // Format converts the bounds to a string of the form "[foo, bar)", using the
> ```

nit: "[foo, bar)" or "[foo, bar]" ...
___
*[`internal/manifest/version.go` line 985 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtMAigz7Ufu6Bd0WFSY:-NtMAigz7Ufu6Bd0WFSZ:b-txr19v) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/manifest/version.go#L985)):*
> ```Go
> 
> func overlaps(iter LevelIterator, cmp Compare, bounds base.UserKeyBounds) LevelSlice {
> 	startIter := iter.Clone()
> ```

how about adding a comment that `overlaps` is only for levels other than L0.
___
*[`internal/manifest/version.go` line 1006 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtMAtstDQB6kInyCStV:-NtMAtstDQB6kInyCStW:bu1zcvt) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/manifest/version.go#L1006)):*
> ```Go
> 		if bounds.End.Kind == base.Inclusive {
> 			// endIter is now pointing at the *first* file with a largest key >= end.
> 			// If there are multiple files including the user key `end`, we want all
> ```

Will there be multiple files, given we don't split user keys across files? should we add a panic?
We are also not checking if the `endIterFile.Largest` is exclusive, so I suppose it could be and we do need to `Next()`. Though, `LevelIterator.SeekGE` should give us some guarantees there. Probably best to leave this as-is for now, with a TODO to come back to this.

RaduBerinde:
*[`internal/base/key_bounds.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM2EfqEMiyvWfq8mtN:-NtMU6Q6DklaZX92I0r-:b8qvfcn) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/base/key_bounds.go#L62)):*

would `GreaterOrEqual` fully capture the meaning?
</blockquote></details>

GreaterOrEqual is confusing when it's exclusive. I wanted a more conceptual term that abstracts away the exact kind of comparison we need to make. If it was a method on the key I'd call it "Within(boundary)" or "CompliesWith(boundary)" I couldn't find a better reciprocal to that than "applies". Maybe "Contains" would work better?

sumeerbhola (changes requested):
> Reviewed 3 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3417)* status: 15 of 18 files reviewed, 9 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`flushable.go` line 297 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM9HB4Bo25Hm5BZ_f-:-NtMwor31v-EQHm9xeSV:b-gm90mg) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/flushable.go#L297)):*

Do you understand why it's ok to stop and `break` below? Is it not possible that a `bounded` overlaps multiple files on a level?
</blockquote></details>

Yes, it could overlap with multiple files in the `ingestedFlushable`.
The contract is that the `ingestedFlushable` is one entity, and the caller just wants to know the `bounded` objects that overlap with that entity. `fn` doesn't care which part of that entity overlapped.
___
*[`internal/base/key_bounds.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM2EfqEMiyvWfq8mtN:-NtMrdUaE7Q_wD7frEMV:b-ot6whu) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/base/key_bounds.go#L62)):*

GreaterOrEqual is confusing when it's exclusive. I wanted a more conceptual term that abstracts away the exact kind of comparison we need to make. If it was a method on the key I'd call it "Within(boundary)" or "CompliesWith(boundary)" I couldn't find a better reciprocal to that than "applies". Maybe "Contains" would work better?
</blockquote></details>

Admittedly, we are doing something slightly odd. All the callers of this method, and `AppliesToInternalKey`, from outside this file, have a whole `UserKeyBounds` and are choosing to call with the end key in that bounds struct (instead of calling a method on the whole struct).

So stepping back, if we have an interval [a, b) or [a, b], and a different key c, what we are asking is whether the interval extends up to or beyond c. But we also want it to return true when a > c, so there isn't a clear meaning to what we are asking even if we asked with a whole interval (which suggests "Contains" could be more confusing)

So coming back to the end key by itself. It represents a key b (inclusive) or immediatePredecessor(b) (exclusive). Yes, we can't compute the immediatePredecessor, but it is conceptually a key (I realize I am on shaky ground here, since one can't argue that a real number has an immediate predecessor, even if one can't compute it). So if we go with that interpretation, GreaterOrEqual is reasonable, just like 1.999... < 2 - delta as delta approaches 0.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3417#-:-NtNIicW3fns3oWv9ynL:b-814g8k)* status: 13 of 18 files reviewed, 8 unresolved discussions (waiting on @itsbilal, @jbowens, and @sumeerbhola)

___
*[`flushable.go` line 297 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM9HB4Bo25Hm5BZ_f-:-NtN892sBkrct2VF3TX8:b-whespq) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/flushable.go#L297)):*

Yes, it could overlap with multiple files in the `ingestedFlushable`.
The contract is that the `ingestedFlushable` is one entity, and the caller just wants to know the `bounded` objects that overlap with that entity. `fn` doesn't care which part of that entity overlapped.
</blockquote></details>

Ah, duh, thanks. I reworked the code a bit to make things more clear. Also added an invariants check that the files really are sorted.
___
*[`internal/base/key_bounds.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM2EfqEMiyvWfq8mtN:-NtNBVn39d7VWF6XjQOV:b-7e85yu) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/base/key_bounds.go#L62)):*

Admittedly, we are doing something slightly odd. All the callers of this method, and `AppliesToInternalKey`, from outside this file, have a whole `UserKeyBounds` and are choosing to call with the end key in that bounds struct (instead of calling a method on the whole struct).

So stepping back, if we have an interval [a, b) or [a, b], and a different key c, what we are asking is whether the interval extends up to or beyond c. But we also want it to return true when a > c, so there isn't a clear meaning to what we are asking even if we asked with a whole interval (which suggests "Contains" could be more confusing)

So coming back to the end key by itself. It represents a key b (inclusive) or immediatePredecessor(b) (exclusive). Yes, we can't compute the immediatePredecessor, but it is conceptually a key (I realize I am on shaky ground here, since one can't argue that a real number has an immediate predecessor, even if one can't compute it). So if we go with that interpretation, GreaterOrEqual is reasonable, just like 1.999... < 2 - delta as delta approaches 0.
</blockquote></details>

I was thinking about it like `EndBoundary` represents the range `[MinKey - Key]` or `[MinKey - Key)`, hence the Contains. But I agree that when using it from `UserKeyBounds` it becomes confusing.

I see your real line analogy but the readers of the code might not have the same analogy in mind..

Some other ideas:
 - `Includes() / IncludesInternalKey()`
 - `Encloses() / EnclosesInternalKey()`
 - reverse the meaning and call it `Excludes[InternalKey]`

Or reverse it to `base.UserKeyWithinEndBoundary()` and `InternalKey.WithinEndBoundary()`. Maybe one day we'll have a `UserKey` type and the first one would become `UserKey.WithinEndBoundary()`.
___
*[`internal/manifest/version.go` line 985 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtMAigz7Ufu6Bd0WFSY:-NtNFXszE3FXuXXrVwY8:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/manifest/version.go#L985)):*

how about adding a comment that `overlaps` is only for levels other than L0.
</blockquote></details>

Done.
___
*[`internal/manifest/version.go` line 1006 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtMAtstDQB6kInyCStV:-NtNHqXE6xba56Esb3Mn:b-dzfe3v) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/manifest/version.go#L1006)):*

Will there be multiple files, given we don't split user keys across files? should we add a panic?
We are also not checking if the `endIterFile.Largest` is exclusive, so I suppose it could be and we do need to `Next()`. Though, `LevelIterator.SeekGE` should give us some guarantees there. Probably best to leave this as-is for now, with a TODO to come back to this.
</blockquote></details>

Yeah I tried to simplify it and got some test failures. Leaving a TODO.

sumeerbhola (approved):
:lgtm: 

> Reviewed 2 of 6 files at r2, 1 of 2 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3417#-:-NtNbGNj5YLzVXXUn0Qp:bg7igkv)* status: all files reviewed, 8 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`internal/base/key_bounds.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM2EfqEMiyvWfq8mtN:-NtNZx3c5KeklOufj-oj:b-1061l1) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/base/key_bounds.go#L62)):*

I was thinking about it like `EndBoundary` represents the range `[MinKey - Key]` or `[MinKey - Key)`, hence the Contains. But I agree that when using it from `UserKeyBounds` it becomes confusing.

I see your real line analogy but the readers of the code might not have the same analogy in mind..

Some other ideas:
 - `Includes() / IncludesInternalKey()`
 - `Encloses() / EnclosesInternalKey()`
 - reverse the meaning and call it `Excludes[InternalKey]`

Or reverse it to `base.UserKeyWithinEndBoundary()` and `InternalKey.WithinEndBoundary()`. Maybe one day we'll have a `UserKey` type and the first one would become `UserKey.WithinEndBoundary()`.
</blockquote></details>

Go with whatever name you prefer -- I don't want to bike shed this.

One thought is that the method comment says "Applies returns true if the given key is "before" the end boundary." (though it should say "before or equal"), and how we explain the method can suggest the name. So perhaps `AfterOrEqual`?
___
*[`internal/manifest/version.go` line 1006 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtMAtstDQB6kInyCStV:-NtN_tB3BsoAF2M8y121:b-ucfkn4) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/manifest/version.go#L1006)):*

Yeah I tried to simplify it and got some test failures. Leaving a TODO.
</blockquote></details>

Thanks.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3417)* status: all files reviewed, 8 unresolved discussions (waiting on @itsbilal, @jbowens, and @sumeerbhola)

___
*[`internal/base/key_bounds.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM2EfqEMiyvWfq8mtN:-NtOB1TU9hSNdSWkQzoc:brletmy) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/base/key_bounds.go#L62)):*

Go with whatever name you prefer -- I don't want to bike shed this.

One thought is that the method comment says "Applies returns true if the given key is "before" the end boundary." (though it should say "before or equal"), and how we explain the method can suggest the name. So perhaps `AfterOrEqual`?
</blockquote></details>

I don't love either of the options so far.. I'll sleep on it a bit.

I wrote "before" because in my mind the boundary is conceptually an infinitely thin vertical line that is immediately either before of after `.Key` :)

RaduBerinde:
*[`internal/base/key_bounds.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM2EfqEMiyvWfq8mtN:-NtOK-1eDm0uLG1ZGTlE:b-7ci3u1) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/base/key_bounds.go#L62)):*

I don't love either of the options so far.. I'll sleep on it a bit.

I wrote "before" because in my mind the boundary is conceptually an infinitely thin vertical line that is immediately either before of after `.Key` :)
</blockquote></details>

Yet another option would be a method on `Compare` itself, `cmp.WithinEndBound(key, b.End)`

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3417)* status: 13 of 18 files reviewed, 8 unresolved discussions (waiting on @itsbilal, @jbowens, and @sumeerbhola)

___
*[`internal/base/key_bounds.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM2EfqEMiyvWfq8mtN:-NtOp7jODqqYpRpXbCDQ:b-n26ckj) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/base/key_bounds.go#L62)):*

Yet another option would be a method on `Compare` itself, `cmp.WithinEndBound(key, b.End)`
</blockquote></details>

Ok, I came up with something reasonable. I changed the boundary to be a more generic `UserKeyBoundary` (no longer specific to the end of the interval) and changed this method to `IsUpperBoundFor`. This makes the call sites more clear.

sumeerbhola (approved):
:lgtm: 

> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3417#-:-NtRzGlG3WPcdFbealo3:bg7igkv)* status: all files reviewed, 8 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`internal/base/key_bounds.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM2EfqEMiyvWfq8mtN:-NtRzA329jt9-26zEMxp:b-mp4sc1) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/base/key_bounds.go#L62)):*

Ok, I came up with something reasonable. I changed the boundary to be a more generic `UserKeyBoundary` (no longer specific to the end of the interval) and changed this method to `IsUpperBoundFor`. This makes the call sites more clear.
</blockquote></details>

`IsUpperBoundFor` is perfect.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3417)* status: all files reviewed, 8 unresolved discussions (waiting on @itsbilal, @jbowens, and @sumeerbhola)

___
*[`internal/base/key_bounds.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3417#-NtM2EfqEMiyvWfq8mtN:-NtS-QiM0vcYvvZ8iTgR:bz5m8k5) ([raw file](https://github.com/cockroachdb/pebble/blob/5f3957aef786c1577c2df94f05ad66c334da5a82/internal/base/key_bounds.go#L62)):*

`IsUpperBoundFor` is perfect.
</blockquote></details>

TFTR!!

PR #3418, opened 2024-03-18
db: reuse backings for external ingestions

RaduBerinde:
When we ingest the same external file multiple times (either in separate ingestions or within the same ingestion), we now use the same backing.

We use the recently added virtual backing protection mechanism to make sure that the backings don't disappear between the point where we decide to reuse and applying the version change. The metamorphic test provides good coverage for this.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -204,13 +204,6 @@ func ingestLoad1External(
 		Virtual:      true,
 		Size:         e.Size,
 	}
-	// For simplicity, we use the same number for both the FileNum and the
-	// DiskFileNum (even though this is a virtual sstable). Pass the underlying
-	// FileBacking's size to the same size as the virtualized view of the sstable.
-	// This ensures that we don't over-prioritize this sstable for compaction just
-	// yet, as we do not have a clear sense of what parts of this sstable are
-	// referenced by other nodes.
-	meta.InitProviderBacking(base.DiskFileNum(fileNum), e.Size)
 
 	// In the name of keeping this ingestion as fast as possible, we avoid
 	// *all* existence checks and synthesize a file metadata with smallest/largest
@@ -235,9 +228,9 @@ func ingestLoad1External(
 	meta.SyntheticPrefix = e.SyntheticPrefix
 	meta.SyntheticSuffix = e.SyntheticSuffix
 
-	if err := meta.Validate(opts.Comparer.Compare, opts.Comparer.FormatKey); err != nil {
-		return nil, err
-	}
+	//if err := meta.Validate(opts.Comparer.Compare, opts.Comparer.FormatKey); err != nil {
+	//	return nil, err
+	//}
 	return meta, nil
 }
 
@@ -407,6 +400,11 @@ type ingestSharedMeta struct {
 type ingestExternalMeta struct {
 	*fileMetadata
 	external ExternalFile
+	// tookRefOnBacking is true if the external file is reusing a backing
+	// that existed before this ingestion. In this case, we took a reference count
+	// on that backing that we need to restore. When we drop the reference, we set
+	// tookRefOnBacking back to false.
+	usedExistingBacking bool
 }
 
 func (r *ingestLoadResult) fileCount() int {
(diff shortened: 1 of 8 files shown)


sumeerbhola (approved):
:lgtm: 

> Reviewed 7 of 8 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3418#-:-NtHXoAd5CxbOHfanxUc:bg7igkv)* status: all files reviewed, 3 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`ingest.go` line 403 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3418#-NtHRtOX-4JQHkqE_oRO:-NtHRtOX-4JQHkqE_oRP:bm5ocoh) ([raw file](https://github.com/cockroachdb/pebble/blob/22b1244c69e7ccdcc88579e196e354b95008e368/ingest.go#L403)):*
> ```Go
> 	*fileMetadata
> 	external ExternalFile
> 	// tookRefOnBacking is true if the external file is reusing a backing
> ```

useExistingBacking?
___
*[`ingest.go` line 580 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3418#-NtHSqMS8I5B8FfyHCpq:-NtHSqMS8I5B8FfyHCpr:b-1g9pjh) ([raw file](https://github.com/cockroachdb/pebble/blob/22b1244c69e7ccdcc88579e196e354b95008e368/ingest.go#L580)):*
> ```Go
> 
> // ingestLink creates new objects which are backed by either hardlinks to or
> // copies of the ingested files. It also attaches shared objects to the provider.
> ```

If this is going to only handle local objects, does it need the whole `ingestLoadResult`. If it does, we should clarify in a code comment that it is only linking local tables.
___
*[`ingest.go` line 610 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3418#-NtHTVEf4gl4dHOqxSrd:-NtHTVEf4gl4dHOqxSre:b-5pmwgs) ([raw file](https://github.com/cockroachdb/pebble/blob/22b1244c69e7ccdcc88579e196e354b95008e368/ingest.go#L610)):*
> ```Go
> //
> // For external objects, we reuse existing FileBackings from the current version
> // when possible.
> ```

can we extend this in the future to shared tables too?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3418#-:-NtI0PziBq7aOXWLm2mD:b-814g8k)* status: 3 of 10 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`ingest.go` line 403 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3418#-NtHRtOX-4JQHkqE_oRO:-NtI-Mlz1WEbRE59ULVk:b-cisnhw) ([raw file](https://github.com/cockroachdb/pebble/blob/22b1244c69e7ccdcc88579e196e354b95008e368/ingest.go#L403)):*

useExistingBacking?
</blockquote></details>

Done. And fixed the comment.
___
*[`ingest.go` line 580 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3418#-NtHSqMS8I5B8FfyHCpq:-NtI-qj4DZ18YPjAtyRa:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/22b1244c69e7ccdcc88579e196e354b95008e368/ingest.go#L580)):*

If this is going to only handle local objects, does it need the whole `ingestLoadResult`. If it does, we should clarify in a code comment that it is only linking local tables.
</blockquote></details>

Done.
___
*[`ingest.go` line 610 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3418#-NtHTVEf4gl4dHOqxSrd:-NtI02fQ86GGorlLTpXc:bnxeuwt) ([raw file](https://github.com/cockroachdb/pebble/blob/22b1244c69e7ccdcc88579e196e354b95008e368/ingest.go#L610)):*

can we extend this in the future to shared tables too?
</blockquote></details>

I don't see why not. When would that be useful? When a range moves away from us and then comes back?

sumeerbhola (approved):
:lgtm: 

> Reviewed 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3418#-:-NtNkDXq5fnDj74_y_Ze:bg7igkv)* status: all files reviewed, 3 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`ingest.go` line 610 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3418#-NtHTVEf4gl4dHOqxSrd:-NtNjuOd9MvCu4_dQiBm:bpctrbd) ([raw file](https://github.com/cockroachdb/pebble/blob/22b1244c69e7ccdcc88579e196e354b95008e368/ingest.go#L610)):*

I don't see why not. When would that be useful? When a range moves away from us and then comes back?
</blockquote></details>

I was thinking of range splits, resulting in multiple ranges sharing the same sstable.

PR #3423, opened 2024-03-18
metamorphic: add DownloadOp

RaduBerinde:
Test `DB.Download` in the metamorphic test.

### DIFF
--- a/metamorphic/config.go
+++ b/metamorphic/config.go
@@ -17,6 +17,7 @@ const (
 	OpDBCheckpoint
 	OpDBClose
 	OpDBCompact
+	OpDBDownload
 	OpDBFlush
 	OpDBRatchetFormatMajorVersion
 	OpDBRestart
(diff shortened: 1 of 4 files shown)


RaduBerinde:
*[`metamorphic/generator.go` line 411 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3423#-NtIAg4-E1eVO5DDY1DF:-NtIAg4-E1eVO5DDY1DG:b-7282t0) ([raw file](https://github.com/cockroachdb/pebble/blob/8c8f4100d6f872d5710f366a55b0b13f707c0c96/metamorphic/generator.go#L411)):*
> ```Go
> 	spans := make([]pebble.DownloadSpan, numSpans)
> 	for i := range spans {
> 		keys := g.keyGenerator.UniqueKeys(2, func() []byte { return g.keyGenerator.RandKey(0.001) })
> ```

@itsbilal Do the spans need to be disjoint? (logically, of course, but in practice it seems to work fine?)

itsbilal (approved):
LGTM

Non-disjoint spans should be alright! I don't see a reason why they wouldn't just work, so this is good.

RaduBerinde:
TFTR!

PR #3424, opened 2024-03-19
db: allow runtime configuration of compression

jbowens:
Allow clients to change configuration of the compression algorithm at runtime. This will be used to tie the compression algorithm to a CockroachDB cluster setting, enabling opt-in use of zstd.


jbowens:
TFTR!

PR #3425, opened 2024-03-19
db: surface OPTIONS filename in pebble.Peek

jbowens:
The pebble.Peek function allows external users to retrieve metadata about an Engine without opening it. This commit adds the filename of the latest OPTIONS file to the output, so the caller may parse the options used by the most recent process. Additionally, this commit adds new methods for locating atomicfs.Marker that use an existing directory listing to avoid repeated listing of the data directory during Open and Peek.

### DIFF
--- a/format_major_version.go
+++ b/format_major_version.go
@@ -282,9 +282,9 @@ const formatVersionMarkerName = `format-version`
 // only acceptable if we are creating a new store (we no longer support
 // FormatMostCompatible which is the only one with no version marker file).
 func lookupFormatMajorVersion(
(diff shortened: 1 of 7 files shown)


RaduBerinde (commented):
Nice! :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3425#-:-NtMlZPb7rXWFspiP_gc:bfd61qr)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

PR #3433, opened 2024-03-20
vfs/vfstest: adjust WithOpenFileTracking to not wrap typed nils

jbowens:
Previously, in error cases that returned a typed nil, WithOpenFileTracking would wrap the typed nil file with open file tracking, incorrectly identifying file leaks that don't exist.


RaduBerinde:
:lgtm:

RaduBerinde (approved):

jbowens:
TFTR!

PR #3436, opened 2024-03-21
db: remove strictWALTail option

RaduBerinde:
All compatible versions are expected to have `strict_wal_tail=true` so
we can remove this option now. We also clean up some of the options
checking code.


jbowens (commented):
Hrm... you know, I think we should actually be able to remove the reading of the `strict_wal_tail` option and instead always assume it's true. That option was introduced before the existence of format major versions and would've been a format major version had we done it today. At this point, our minimum format major version requirements provide a guarantee that the last pebble process to open the database produced clean WAL tails.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3436#-:-NtX9XuF21BjZ5IlhGhQ:b6sn290)* status: 0 of 3 files reviewed, all discussions resolved

RaduBerinde:
Noob question but how do we ensure clean WAL tails?

jbowens:
> Noob question but how do we ensure clean WAL tails?

We only ensure it for WALs other than the most recent, and we ensure it by closing the previous WAL (including fsyncing a chunk that cleanly marks an EOF for the logical WAL even if the physical file continues due to log recycling) before creating the new WAL.

RaduBerinde (commented):
Thanks, makes sense. Updated to remove the setting (and left a historical comment where we pass the flag).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3436#-:-NtXTZOBAQ4mLB8t7aiE:b-1vgnmg)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @jbowens)

jbowens (approved):
:lgtm_strong: 

nice!

> Reviewed 1 of 3 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3436#-:-NtXYg-i6Db05GV1JsK1:bleuxg1)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

Issue #3432, opened 2024-03-20
metamorphic: add cases with `sstContainsExciseTombstone=true` for IngestAndExcise


aadityasondhi:
As a follow up to changes in https://github.com/cockroachdb/pebble/pull/3398. We should test with the option set to true with cases where the ingested SSTs contain range dels.

itsbilal:
This was fixed in https://github.com/cockroachdb/pebble/pull/3443 .

PR #3437, opened 2024-03-21
metamorphic: re-enable ingest split and ingestAndExcise in TestMeta

itsbilal:
Previously, we disabled these features in the main metamorphic test to try to tease out issues elsewhere. However, seeing as these features are being turned on in 24.1 and have been stable for a while, we should turn it on in the main metamorphic test.

This change also implements the bulk of #3432 (i.e. the case where we can flushable-ingest an excise) but leaves it off as it's failing at the moment.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1467,6 +1467,11 @@ func (d *DB) runIngestFlush(c *compaction) (*manifest.VersionEdit, error) {
 	suggestSplit := d.opts.Experimental.IngestSplit != nil && d.opts.Experimental.IngestSplit() &&
 		d.FormatMajorVersion() >= FormatVirtualSSTables
 
+	if suggestSplit || ingestFlushable.exciseSpan.Valid() {
+		// We could add deleted files to ve.
+		ve.DeletedFiles = make(map[manifest.DeletedFileEntry]*manifest.FileMetadata)
+	}
+
 	replacedFiles := make(map[base.FileNum][]newFileEntry)
 	for _, file := range ingestFlushable.files {
 		// This file fits perfectly within the excise span, so we can slot it at L6.
@@ -1483,7 +1488,6 @@ func (d *DB) runIngestFlush(c *compaction) (*manifest.VersionEdit, error) {
 		}
 
 		if ingestFlushable.exciseSpan.Valid() {
-			ve.DeletedFiles = map[manifest.DeletedFileEntry]*manifest.FileMetadata{}
 			// Iterate through all levels and find files that intersect with exciseSpan.
 			for level = range c.version.Levels {
 				overlaps := c.version.Overlaps(level, base.UserKeyBoundsEndExclusive(ingestFlushable.exciseSpan.Start, ingestFlushable.exciseSpan.End))
--- a/metamorphic/config.go
+++ b/metamorphic/config.go
@@ -175,7 +175,7 @@ func DefaultOpConfig() OpConfig {
 			OpWriterDelete:                100,
 			OpWriterDeleteRange:           50,
 			OpWriterIngest:                100,
-			OpWriterIngestAndExcise:       0, // TODO(bilal): Enable this.
+			OpWriterIngestAndExcise:       50,
 			OpWriterLogData:               10,
 			OpWriterMerge:                 100,
 			OpWriterRangeKeySet:           10,
@@ -313,7 +313,6 @@ func WriteOpConfig() OpConfig {
 func multiInstanceConfig() OpConfig {
 	cfg := DefaultOpConfig()
 	cfg.ops[OpReplicate] = 5
-	cfg.ops[OpWriterIngestAndExcise] = 50
 	// Single deletes and merges are disabled in multi-instance mode, as
 	// replicateOp and ingestAndExciseOp don't support them.
 	cfg.ops[OpWriterSingleDelete] = 0
--- a/metamorphic/generator.go
+++ b/metamorphic/generator.go
@@ -1283,6 +1283,8 @@ func (g *generator) writerIngestAndExcise() {
 		derivedDBID: derivedDBID,
 		exciseStart: start,
 		exciseEnd:   end,
+		// TODO(bilal): Uncomment this when known bugs are fixed.
+		//sstContainsExciseTombstone: g.rng.Intn(2) == 0,
 	})
 }
 
--- a/metamorphic/generator.go
+++ b/metamorphic/generator.go
@@ -1277,6 +1277,22 @@ func (g *generator) writerIngestAndExcise() {
 	start, end := g.prefixKeyRange()
 	derivedDBID := g.objDB[batchID]
 
+	// Check for any single delete conflicts. If this batch is single-deleting
+	// a key that isn't safe to single delete in the underlying db, _and_ this
+	// key is not in the excise span, we add a delete before the ingestAndExcise.
+	singleDeleteConflicts := g.keyManager.checkForSingleDelConflicts(batchID, dbID, true /* collapsed */)
+	for _, conflict := range singleDeleteConflicts {
+		if g.cmp(conflict, start) >= 0 && g.cmp(conflict, end) < 0 {
+			// This key will get excised anyway.
+			continue
+		}
+		g.add(&deleteOp{
+			writerID:    dbID,
+			key:         conflict,
+			derivedDBID: dbID,
+		})
+	}
+
 	g.add(&ingestAndExciseOp{
 		dbID:        dbID,
 		batchID:     batchID,
--- a/metamorphic/key_manager.go
+++ b/metamorphic/key_manager.go
@@ -629,9 +629,30 @@ func (k *keyManager) update(o op) {
 			k.objKeyMeta(batchID).CollapseKeys()
 			k.mergeObjectInto(batchID, s.dbID)
 		}
-		// TODO(bilal): Handle ingestAndExciseOp and replicateOp here. We currently
-		// disable SingleDelete when these operations are enabled (see
-		// multiInstanceConfig).
+	case *ingestAndExciseOp:
+		// IngestAndExcise does not ingest multiple batches, so we will not see
+		// a failure due to overlapping sstables. However we do need to merge
+		// the singular batch into the key manager.
+		//
+		// Remove all keys from the key manager within the excise span before
+		// merging the batch into the db.
+		ts := k.nextMetaTimestamp()
+		keyRange := pebble.KeyRange{Start: s.exciseStart, End: s.exciseEnd}
+		for _, key := range k.knownKeysInRange(keyRange) {
+			meta := k.getOrInit(s.batchID, key)
+			if meta.objID.tag() == dbTag {
+				meta.clear()
+			} else {
+				meta.history = append(meta.history, keyHistoryItem{
+					opType:        OpWriterDeleteRange, // Mimic a DeleteRange.
+					metaTimestamp: ts,
+				})
+			}
+		}
+		k.objKeyMeta(s.batchID).CollapseKeys()
+		k.mergeObjectInto(s.batchID, s.dbID)
+		// TODO(bilal): Handle replicateOp here. We currently disable SingleDelete
+		// when these operations are enabled (see multiInstanceConfig).
 	case *newExternalObjOp:
 		// Collapse and transfer the keys from the batch to the external object.
 		k.objKeyMeta(s.batchID).CollapseKeys()
--- a/metamorphic/ops.go
+++ b/metamorphic/ops.go
@@ -882,10 +882,11 @@ func (o *ingestOp) keys() []*[]byte                     { return nil }
 func (o *ingestOp) diagramKeyRanges() []pebble.KeyRange { return nil }
 
 type ingestAndExciseOp struct {
-	dbID                   objID
-	batchID                objID
-	derivedDBID            objID
-	exciseStart, exciseEnd []byte
+	dbID                       objID
+	batchID                    objID
+	derivedDBID                objID
+	exciseStart, exciseEnd     []byte
+	sstContainsExciseTombstone bool
 }
 
 func (o *ingestAndExciseOp) run(t *Test, h historyRecorder) {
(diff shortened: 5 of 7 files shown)


itsbilal:
Oh look, the test already found a bug :eyes: fixed in the second commit

aadityasondhi (approved):
:lgtm: thanks for adding these and fixing the bug!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3437#-:-NtXRKvZ3Pq_ozntKUFa:b-1o1793)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`metamorphic/options.go` line 786 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3437#-NtXRFNK-Sg-yKv_w0nF:-NtXRFNK-Sg-yKv_w0nG:b-jpfgrz) ([raw file](https://github.com/cockroachdb/pebble/blob/d6f85bd44c74a4df7bc7bbab38741cb99c236394/metamorphic/options.go#L786)):*
> ```Go
> 			testOpts.Opts.FormatMajorVersion = pebble.FormatVirtualSSTables
> 		}
> 		// TODO(bilal): Enable this when known bugs with this are fixed.
> ```

What are the known bugs?

RaduBerinde (commented):
Shouldn't this be a randomized option in `ingestAndExciseOp` instead?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3437#-:-NtXV10xD_bt1CnN9KeI:b-7i4fvi)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @itsbilal)

itsbilal:
@aadityasondhi just did a run with that line commented and got this:

```
        iter2 = iter1.Clone(false, "", "", 2 /* key types */, 0, 0, false /* use L6 filters */, "" /* masking suffix */) // <nil> #264
        iter1.Next("") // [true,"pzxvuimmhebd",<no point>,["pzxvuimmhebd","rrgbo@2")=>{"@2"="lkolob"}*] <nil> #266
        iter1.Next("vqycec") // [invalid] <nil> #267
        // INFO: [JOB 20] flushing 1 ingested table
        iter2.SeekLT("jqkcsohyrsk", "") // [true,"izmivk","wxceabllgx",["bruxsifrwpmn","zuspniljrda")=>{"@2"="lkolob"}*] <nil> #265
        iter2.SeekPrefixGE("msrkizyzr") // [true,"msrkizyzr","pmkapqquiqgsbgvereh",["msrkizyzr","msrkizyzr\x00")=>{"@2"="lkolob"}*] <nil> #269
        iter2.First() // [true,"bqzsq","nbevdjqdztzmcwehc",<no range>*] <nil> #273
        iter2.SeekPrefixGE("thtniqsfannt@2") // [true,"thtniqsfannt@2",<no point>,["thtniqsfannt","thtniqsfannt\x00")=>{"@2"="lkolob"}*] <nil> #274
        iter2.Next("") // [false] <nil> #275
        iter2.SeekGE("dipr", "irrsdssgqxva") // [valid,"dipr",<no point>,["bruxsifrwpmn","zuspniljrda")=>{"@2"="lkolob"}*] <nil> #276
        // FATAL: MANIFEST apply failed: pebble: internal error: L6 files 000026 and 000016 have overlapping ranges: [bruxsifrwpmn#0,SET-wkrdgkmfmngd#inf,RANGEKEYSET] vs [jajsfhrvbv#150,DEL-xurozrwva#inf,RANGEDEL]
        // INFO: [JOB 20] flush error: MANIFEST apply failed: pebble: internal error: L6 files 000026 and 000016 have overlapping ranges: [bruxsifrwpmn#0,SET-wkrdgkmfmngd#inf,RANGEKEYSET] vs [jajsfhrvbv#150,DEL-xurozrwva#inf,RANGEDEL]
        iter2.SeekGE("igkhwvzdhaw", "") // [true,"igkhwvzdhaw","jgmulakuuivovhmn",["bruxsifrwpmn","zuspniljrda")=>{"@2"="lkolob"}] <nil> #281

        To reduce:  go test ./internal/metamorphic -tags invariants -run 'TestMeta$' --run-dir _meta/240321-153438.111892373176/random-013 --try-to-reduce -v
```

It's possible we aren't applying the excise during ingest time? Or we might not be erroring out any conflicting compactions which we're supposed to also do

itsbilal:
@RaduBerinde I think that works too. I was hoping this'd get us equivalence with the non-flushable-ingest case while an operation parameter won't, but I think we already get some level of equivalence with `doExcise=false` so let's just do this as a parameter.

itsbilal:
@RaduBerinde it's now an argument

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3437#-:-NtXb5-YDF8pRxB_9wgm:bnfp4nl)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @aadityasondhi and @itsbilal)

___
*[`metamorphic/ops.go` line 960 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3437#-NtXamVA-01JqiAW6QhE:-NtXamVA-01JqiAW6QhF:b-z8nwmc) ([raw file](https://github.com/cockroachdb/pebble/blob/91e4a805168cd688054ca8ba5b9a81b401f4a4dd/metamorphic/ops.go#L960)):*
> ```Go
> 
> func (o *ingestAndExciseOp) String() string {
> 	return fmt.Sprintf("%s.IngestAndExcise(%s, %q, %q, %t)", o.dbID, o.batchID, o.exciseStart, o.exciseEnd, o.sstContainsExciseTombstone)
> ```

[nit] add `/* sstContainsExciseTombstone */`

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3437#-:-NtYDF4DC1TLknHmIZDu:b-814g8k)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @RaduBerinde)

___
*[`metamorphic/ops.go` line 960 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3437#-NtXamVA-01JqiAW6QhE:-NtYDDb738svxxDaRbVe:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/91e4a805168cd688054ca8ba5b9a81b401f4a4dd/metamorphic/ops.go#L960)):*

[nit] add `/* sstContainsExciseTombstone */`
</blockquote></details>

Done.

itsbilal:
Looks like we have a failure from ingest-split in the CI run already 😬

itsbilal:
Diff:

```
-snap37.Get("wyaf@17") // [""] pebble: not found #4026
+snap37.Get("wyaf@17") // ["p"] <nil> #4026
```

Relevant ops:

```
db1.Set("wyaf@17", "p")
db1.IngestAndExcise(batch10, "ktdo", "swckmgxsfpsx", false /* sstContainsExciseTombstone */)
db1.IngestAndExcise(batch12, "hpbae", "vpjtfatj", false /* sstContainsExciseTombstone */)
db1.DeleteRange("nctwv", "wcpsbrqptacb@1")
db1.IngestAndExcise(batch15, "nctwv", "vpjtfatj", false /* sstContainsExciseTombstone */)
db1.IngestAndExcise(batch18, "ccmbgorxbbkw", "fxowfmrbtahf", false /* sstContainsExciseTombstone */)
db1.DeleteRange("vpjtfatj@2", "vpjtfatj@1")
db1.IngestAndExcise(batch21, "drtkgkleta", "qboh", false /* sstContainsExciseTombstone */)
db1.Set("wyaf@17", "ujordaxdtbrnvdpmrld")
db1.IngestAndExcise(batch26, "uobnqthed", "wcpsbrqptacb", false /* sstContainsExciseTombstone */)
db1.IngestAndExcise(batch32, "ktdo", "pths", false /* sstContainsExciseTombstone */)
batch33.SingleDelete("wyaf@17", false /* maybeReplaceDelete */)
db1.IngestAndExcise(batch33, "gzkfzkykhup", "recuhyoq", false /* sstContainsExciseTombstone */)
db1.IngestAndExcise(batch36, "swckmgxsfpsx", "wxgbwovejzj", false /* sstContainsExciseTombstone */)
snap37.Get("wyaf@17")
```

I listed all delete ranges, ingestAndExcises, and any direct point writes operations on `wyaf@17`, in order. It looks like we should have disallowed the batch33.SingleDelete but we didn't? None of the range deletes or excises should have cleared away the first write.

RaduBerinde:
It shouldn't be hard to add. IngestAndExcise should be the same as ingest, except it deletes all prior history within the excise span. ReplicateOp should copy the history in the range, overwriting any prior history.

itsbilal:
@RaduBerinde I'm actually not sure if that'll address this particular failure. writerSingleDeleteOp should have still noticed the two SETs on the DB as they happened directly on the DB, not through an ingestAndExcise. It shouldn't matter that batch33 got IngestAndExcise'd as that SingleDelete operation was already generated by then. Let me address that TODO anyway and see what happens

itsbilal:
yep, made the change and we still see a failure with this seed. Need to dive deeper into what's overwriting history on the key

RaduBerinde:
(Note that the SingleDel doesn't go to the DB, it's in the batch that becomes the ingested file).

itsbilal:
@RaduBerinde yep, pushed a new commit that adds that case. The tests should pass now, thanks!

aadityasondhi (approved):

itsbilal:
TFTR!

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3437)* status: 0 of 7 files reviewed, 5 unresolved discussions

___
*[`metamorphic/config.go` line 317 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3437#-NtbtZ-v6uK_IVV3Hx_D:-NtbtZ-v6uK_IVV3Hx_E:b-2rtyy4) ([raw file](https://github.com/cockroachdb/pebble/blob/8328ff07fd213abea79fd868f40ea486be568753/metamorphic/config.go#L317)):*
> ```Go
> 	cfg.ops[OpReplicate] = 5
> 	// Single deletes and merges are disabled in multi-instance mode, as
> 	// replicateOp and ingestAndExciseOp don't support them.
> ```

[nit] remove `and ingestAndExciseOp`
___
*[`metamorphic/key_manager.go` line 641 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3437#-Ntbrxn8Bn6Ztt01py1b:-Ntbrxn8Bn6Ztt01py1c:b-twh60c) ([raw file](https://github.com/cockroachdb/pebble/blob/8328ff07fd213abea79fd868f40ea486be568753/metamorphic/key_manager.go#L641)):*
> ```Go
> 		ts := k.nextMetaTimestamp()
> 		keyRange := pebble.KeyRange{Start: s.exciseStart, End: s.exciseEnd}
> 		for _, key := range k.knownKeysInRange(keyRange) {
> ```

We should use `InRangeKeysFor(dbID)`
___
*[`metamorphic/key_manager.go` line 643 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3437#-Ntbt3Y41NtmtmTl2MJ-:-Ntbt3Y41NtmtmTl2MJ0:b-8mbx7n) ([raw file](https://github.com/cockroachdb/pebble/blob/8328ff07fd213abea79fd868f40ea486be568753/metamorphic/key_manager.go#L643)):*
> ```Go
> 		for _, key := range k.knownKeysInRange(keyRange) {
> 			meta := k.getOrInit(s.batchID, key)
> 			if meta.objID.tag() == dbTag {
> ```

This is not possible, `meta.objID` is `s.batchID`. I think we want to do `getOrInit(dbID, key)`
___
*[`metamorphic/key_manager.go` line 646 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3437#-NtbsTwmCj7wENzCtQ3b:-NtbsTwmCj7wENzCtQ3c:b-vtcp1) ([raw file](https://github.com/cockroachdb/pebble/blob/8328ff07fd213abea79fd868f40ea486be568753/metamorphic/key_manager.go#L646)):*
> ```Go
> 				meta.clear()
> 			} else {
> 				meta.history = append(meta.history, keyHistoryItem{
> ```

We are adding a DeleteRange on top of the key history in the batch, then we collapse all the histories below.. So won't we just end up with the delete instead of whatever was in the batch? I'm confused, why does this work?

We should just delete the keys from the db altogether instead.

itsbilal (commented):
I'll do a separate PR with these suggestions, thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3437#-:-NtbuwQMDG-M-PQ9r63x:b-5apmrw)* status: 0 of 7 files reviewed, 5 unresolved discussions

___
*[`metamorphic/key_manager.go` line 646 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3437#-NtbsTwmCj7wENzCtQ3b:-NtbuVRt8JRZVfH0rV8G:b-glfiar) ([raw file](https://github.com/cockroachdb/pebble/blob/8328ff07fd213abea79fd868f40ea486be568753/metamorphic/key_manager.go#L646)):*

We are adding a DeleteRange on top of the key history in the batch, then we collapse all the histories below.. So won't we just end up with the delete instead of whatever was in the batch? I'm confused, why does this work?

We should just delete the keys from the db altogether instead.
</blockquote></details>

The merge happens _after_ this append (there's a k.nextMetaTimestamp call inside the mergeObjectInto method), so the batch's objects go after the fake DeleteRange. Which seems like the best way to mimic IngestAndExcise - we clear everything then we replace it with the batch.

itsbilal:
*[`metamorphic/key_manager.go` line 646 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3437#-NtbsTwmCj7wENzCtQ3b:-NtbvUxR7xT_tEvxlCWf:biub9ks) ([raw file](https://github.com/cockroachdb/pebble/blob/8328ff07fd213abea79fd868f40ea486be568753/metamorphic/key_manager.go#L646)):*

The merge happens _after_ this append (there's a k.nextMetaTimestamp call inside the mergeObjectInto method), so the batch's objects go after the fake DeleteRange. Which seems like the best way to mimic IngestAndExcise - we clear everything then we replace it with the batch.
</blockquote></details>

ah, nevermind I see what you mean

itsbilal:
#3443 has the fixes.

PR #3443, opened 2024-03-22
db: flushable ingest excise fixes

itsbilal:
Previously, we'd repeatedly call Excise() if we had multiple files to ingest, and we'd mutate the ingestSplitFile param. This change fixes that.

Also update the metamorphic test to resume testing IngestAndExcise with sstContainsExciseTombstone=true.


itsbilal:
@RaduBerinde I incorporated your suggestions from #3437 in here as the second commit

RaduBerinde (approved):
Both commits :lgtm:

Thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3443#-:-NtbytQa7NbENved9Yl0:b-2iyygi)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @aadityasondhi)

itsbilal:
TFTR!

PR #3444, opened 2024-03-22
metamorphic: add per-op timeout

RaduBerinde:
Add a per-op timeout. The default value is a very conservative 5
minutes so we don't cause flakiness in stress/race builds. It can be
changed with `--op-timeout`.

This should help detect hung operations under stress, which I believe
would otherwise keep running.

### DIFF
--- a/internal/metamorphic/metaflags/meta_flags.go
+++ b/internal/metamorphic/metaflags/meta_flags.go
@@ -14,7 +14,9 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"time"
 
+	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/internal/randvar"
 	"github.com/cockroachdb/pebble/metamorphic"
 )
@@ -40,6 +42,8 @@ type CommonFlags struct {
 	// NumInstances is the number of Pebble instances to create in one run. See
 	// "num-instances" flag below.
 	NumInstances int
+	// OpTimeout is a per-operation timeout.
+	OpTimeout time.Duration
 }
 
 func initCommonFlags() *CommonFlags {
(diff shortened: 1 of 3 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3444#-:-Ntpxpjo3YLEeSC8p4NC:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`metamorphic/test.go` line 34 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3444#-NtpvM54C9CAyonAbWF4:-NtpvM54C9CAyonAbWF5:b2zccdj) ([raw file](https://github.com/cockroachdb/pebble/blob/831311a1b35e831f2d0828f8d77046828b0ff982/metamorphic/test.go#L34)):*
> ```Go
> 	t := newTest(ops)
> 	h := newHistory(nil /* failRegexp */, w)
> 	if err := t.init(h, dir, opts, 1 /* numInstances */, 0 /* opTimeout */); err != nil {
> ```

why 0 here? Is this because we don't yet need to support configuring this for tests outside this package?

RaduBerinde:
*[`metamorphic/test.go` line 34 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3444#-NtpvM54C9CAyonAbWF4:-NtpzRhD4-PUBK26QExl:b-wl1jfr) ([raw file](https://github.com/cockroachdb/pebble/blob/831311a1b35e831f2d0828f8d77046828b0ff982/metamorphic/test.go#L34)):*

why 0 here? Is this because we don't yet need to support configuring this for tests outside this package?
</blockquote></details>

Right.. Maybe I should look at putting it inside TestOptions

jbowens:
LGTM

looks like this will help with #3455.

RaduBerinde:
Updated the default timeout to 1 minute / 5 minutes if under race.

PR #3445, opened 2024-03-22
db: add separate download concurrency count

### LINKED ISSUE #3430: downloadOp leads to meta test timeouts
We have seen meta tests timing out since the introduction of `downloadOp` (e.g. https://github.com/cockroachdb/pebble/issues/3426).

With the patch below, we see cases where DownloadOp takes 20+ seconds.

```diff
--- a/metamorphic/meta.go
+++ b/metamorphic/meta.go
@@ -609,7 +609,11 @@ func Execute(m *Test) error {
                                        }
                                }
 
+                               start := time.Now()
                                m.ops[idx].run(m, m.h.recorder(t, idx, nil /* optionalRecordf */))
+                               if delta := time.Since(start); delta > 10*time.Second {
+                                       panic(fmt.Sprintf("%s took %s", m.ops[idx].String(), delta))
+                               }
```

RaduBerinde:
Fixes #3430

#### db: add separate download concurrency count

Currently downloads can be starved by compactions, as we only schedule
downloads when there are no compactions to run.

This change adds a separate concurrency count for downloads so they
run independently of non-download compactions. Downloads should be
bottlenecked on network traffic, so it makes sense that they shouldn't
count towards the compaction concurrency limit.

#### db: minor reorganization of downloadSpan code

Reorganize `downloadSpan` to make the main path more clear. Add a
short sleep in a path where we could potentially busy-loop.

#### metamorphic: reenable downloadOp

Reenable downloadOp and add code to fail the test if any single op
runs for too long. Currently the timeout is very conservative to
prevent flakiness in stress and/or race tests (but it can be turned
down manually to investigate a specific issue).


compaction.go
```diff
@@ -244,7 +244,11 @@ type compaction struct {
 	// application. Only holders of the manifest lock will write to this atomic.
 	cancel atomic.Bool
 
-	kind      compactionKind
+	kind compactionKind
+	// isDownloadCompaction is true if this compaction was started as part of a Download operation.
+	// In this case kind is compactionKindCopy or compactionKindRewrite.
+	isDownloadCompaction bool
```
itsbilal, line 250: nit: can just be isDownload

compaction.go
```diff
@@ -1898,79 +1902,86 @@ func pickElisionOnly(picker compactionPicker, env compactionEnv) *pickedCompacti
...
+		return nil, 0, nil
+	}()
 
-			c := newCompaction(pc, d.opts, d.timeNow(), d.ObjProvider())
-			d.mu.compact.compactingCount++
-			d.addInProgressCompaction(c)
-			go d.compact(c, doneCh)
-		}
+	if err != nil {
+		// Report and error and remove this download span.
```
itsbilal, line 1936: nit: report an error

itsbilal (approved):
LGTM

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3445#-:-Ntc07MY8jR_NfEQebmt:b-814g8k)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`compaction.go` line 250 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3445#gh-1536220004:-Ntc08HzEtVSAIKrwVi0:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/6af2aaa43d238df1636af24aa4e607ab2c1dc6aa/compaction.go#L250)):*

nit: can just be isDownload
</blockquote></details>

Done.
___
*[`compaction.go` line 1936 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3445#gh-1536220507:-Ntc09CHDflxpuEHT0Cx:bt2vr1b) ([raw file](https://github.com/cockroachdb/pebble/blob/6af2aaa43d238df1636af24aa4e607ab2c1dc6aa/compaction.go#L1936)):*

nit: report an error
</blockquote></details>

Fixed.

PR #3451, opened 2024-03-23
sstable: fix two level iterator bug with virtual tables

### LINKED ISSUE #3450: TestMeta failure:  key out of singleLeveliterator virtual bounds
This is 83e94440340e7711281e494cb7798e0befa7d0c4 plus the fix in #3449.
```
        panic: key "qkdpexjhv@27" out of singleLeveliterator virtual bounds qpzjqwf uzgoncjddisn
        
        goroutine 72 [running]:
        github.com/cockroachdb/pebble/sstable.(*singleLevelIterator).maybeVerifyKey(0x140007ed000, 0x140007ed328, {{0x600001b78978?, 0x140009489c8?, 0x104ad83e0?}, 0x0?})
        	/Users/radu/go/src/github.com/cockroachdb/pebble/sstable/reader_iter_single_lvl.go:264 +0x204
        github.com/cockroachdb/pebble/sstable.(*twoLevelIterator).skipForward(0x140007ed000)
        	/Users/radu/go/src/github.com/cockroachdb/pebble/sstable/reader_iter_two_lvl.go:917 +0xfc
        github.com/cockroachdb/pebble/sstable.(*twoLevelIterator).Next(0x140007ed000)
        	/Users/radu/go/src/github.com/cockroachdb/pebble/sstable/reader_iter_two_lvl.go:830 +0x48
        github.com/cockroachdb/pebble.(*levelIter).Next(0x140007d8850)
        	/Users/radu/go/src/github.com/cockroachdb/pebble/level_iter.go:870 +0x17c
        github.com/cockroachdb/pebble.(*mergingIter).switchToMinHeap(0x14000948bf8)
        	/Users/radu/go/src/github.com/cockroachdb/pebble/merging_iter.go:475 +0x560
        github.com/cockroachdb/pebble.(*mergingIter).Next(0x14000948bf8)
        	/Users/radu/go/src/github.com/cockroachdb/pebble/merging_iter.go:1240 +0x224
        github.com/cockroachdb/pebble/internal/keyspan.(*InterleavingIter).nextPos(0x140000e6628)
        	/Users/radu/go/src/github.com/cockroachdb/pebble/internal/keyspan/interleaving_iter.go:646 +0x88
        github.com/cockroachdb/pebble/internal/keyspan.(*InterleavingIter).Next(0x140000e6628)
        	/Users/radu/go/src/github.com/cockroachdb/pebble/internal/keyspan/interleaving_iter.go:480 +0x154
        github.com/cockroachdb/pebble.(*Iterator).nextUserKey(0x14000948600)
        	/Users/radu/go/src/github.com/cockroachdb/pebble/iterator.go:753 +0x164
        github.com/cockroachdb/pebble.(*Iterator).nextWithLimit(0x14000948600, {0x0, 0x0, 0x0})
        	/Users/radu/go/src/github.com/cockroachdb/pebble/iterator.go:1939 +0x274
        github.com/cockroachdb/pebble.(*Iterator).Next(...)
        	/Users/radu/go/src/github.com/cockroachdb/pebble/iterator.go:1684
```

RaduBerinde:
#### sstable: fix two level iterator bug with virtual tables

Consider the following iterator operations: SeekLT, Prev, Prev
(returns nil), Next.

For a top-level or regular index block, it is possible that
`skipBackward` moves beyond the last index block that intersects the
vtable.  Then, when we do Next, we move back to that block and return
the smallest key inside that block, but this is incorrect if the lower
bound is inside this block.

In normal circumstances, `skipBackward` can't move beyond the last
index block; but it is possible when range key masking is in use: on
the filter causes the index block that intersects the vtable to be
skipped. Then the `rangeKeyMasking` filter is updated, and on the Next
call, the block is no longer skipped.

Note that this can't happen in the other direction - this is because
the index key is an end bound for the indexed block and that makes the
boundary check "tight" in `skipForward` but "loose" in `skipBackward`.

This change makes it so that we use `SeekGE` instead of
`firstInternal` when appropriate.

Fixes #3450

#### sstable: TestVirtualReader cleanup

Some cleanup for this test:
 - use proper directive argument format for bounds
 - remove one-off `twoLevel` argument, replace with already supported
   `block-size` and `index-block-size`
 - use standard formatting for properties
 - rename `citer` dirwective to `compaction-iter`

#### sstable: split virtual reader tests

Split the virtual reader tests into one that is focused on the
properties and one that is focused on iterator correctness. We run the
latter with randomized block sizes.

#### sstable: add virtual reader tests with masking

Add support for controlling masking with an interator command and add
a test cases that reproduces #3450.

### DIFF
--- a/sstable/data_test.go
+++ b/sstable/data_test.go
@@ -269,9 +269,10 @@ func scanGlobalSeqNum(td *datadriven.TestData) (uint64, error) {
 type runIterCmdOption func(*runIterCmdOptions)
 
 type runIterCmdOptions struct {
-	everyOp      func(io.Writer)
-	everyOpAfter func(io.Writer)
-	stats        *base.InternalIteratorStats
+	everyOp       func(io.Writer)
+	everyOpAfter  func(io.Writer)
+	stats         *base.InternalIteratorStats
+	maskingFilter TestKeysMaskingFilter
 }
 
 func runIterCmdEveryOp(everyOp func(io.Writer)) runIterCmdOption {
@@ -284,6 +285,12 @@ func runIterCmdEveryOpAfter(everyOp func(io.Writer)) runIterCmdOption {
 
 func runIterCmdStats(stats *base.InternalIteratorStats) runIterCmdOption {
 	return func(opts *runIterCmdOptions) { opts.stats = stats }
+}
+
+// runIterCmdMaskingFilter associates a masking filter and enables use of the
+// "with-masking" command.
+func runIterCmdMaskingFilter(maskingFilter TestKeysMaskingFilter) runIterCmdOption {
+	return func(opts *runIterCmdOptions) { opts.maskingFilter = maskingFilter }
 }
 
 func runIterCmd(
@@ -299,6 +306,24 @@ func runIterCmd(
 
 	var b bytes.Buffer
 	var prefix []byte
+	var maskingSuffix []byte
+	skipMaskedKeys := func(direction int) {
+		if len(maskingSuffix) == 0 {
+			return
+		}
+		for iter.Valid() {
+			k := iter.Key().UserKey
+			suffix := k[testkeys.Comparer.Split(k):]
+			if len(suffix) == 0 || testkeys.Comparer.Compare(suffix, maskingSuffix) <= 0 {
+				return
+			}
+			if direction > 0 {
+				iter.Next()
+			} else {
+				iter.Prev()
+			}
+		}
+	}
 	for _, line := range strings.Split(td.Input, "\n") {
 		parts := strings.Fields(line)
 		if len(parts) == 0 {
@@ -319,6 +344,7 @@ func runIterCmd(
 				}
 			}
 			iter.SeekGE([]byte(strings.TrimSpace(parts[1])), flags)
+			skipMaskedKeys(+1)
 		case "seek-prefix-ge":
 			if len(parts) != 2 && len(parts) != 3 {
 				return "seek-prefix-ge <key> [<try-seek-using-next>]\n"
@@ -333,24 +359,30 @@ func runIterCmd(
 				}
 			}
 			iter.SeekPrefixGE(prefix, prefix /* key */, flags)
+			skipMaskedKeys(+1)
 		case "seek-lt":
 			if len(parts) != 2 {
 				return "seek-lt <key>\n"
 			}
 			prefix = nil
 			iter.SeekLT([]byte(strings.TrimSpace(parts[1])), base.SeekLTFlagsNone)
+			skipMaskedKeys(-1)
 		case "first":
 			prefix = nil
 			iter.First()
+			skipMaskedKeys(+1)
 		case "last":
 			prefix = nil
 			iter.Last()
+			skipMaskedKeys(-1)
 		case "next":
 			iter.Next()
+			skipMaskedKeys(+1)
 		case "next-ignore-result":
 			iter.NextIgnoreResult()
 		case "prev":
 			iter.Prev()
+			skipMaskedKeys(-1)
 		case "next-prefix":
 			if len(parts) != 1 {
 				return "next-prefix should have no parameter\n"
@@ -363,6 +395,7 @@ func runIterCmd(
 			k = k[:prefixLen]
 			kSucc := testkeys.Comparer.ImmediateSuccessor(nil, k)
 			iter.NextPrefix(kSucc)
+			skipMaskedKeys(+1)
 		case "set-bounds":
 			if len(parts) <= 1 || len(parts) > 3 {
 				return "set-bounds lower=<lower> upper=<upper>\n"
@@ -394,6 +427,12 @@ func runIterCmd(
 			continue
 		case "reset-stats":
 			*opts.stats = base.InternalIteratorStats{}
+			continue
+		case "mask-suffix":
+			maskingSuffix = []byte(parts[1])
+			if err := opts.maskingFilter.SetSuffix(maskingSuffix); err != nil {
+				return fmt.Sprintf("set-suffix error: %s", err)
+			}
 			continue
 		case "internal-iter-state":
 			fmt.Fprintf(&b, "| %T:\n", origIter)
--- a/sstable/reader_iter_single_lvl.go
+++ b/sstable/reader_iter_single_lvl.go
@@ -291,6 +291,7 @@ func (i *singleLevelIterator) initBounds() {
 	i.blockLower = i.lower
 	if i.blockLower != nil {
 		key, _ := i.data.First()
+		// TODO(radu): this should be <= 0
 		if key != nil && i.cmp(i.blockLower, key.UserKey) < 0 {
 			// The lower-bound is less than the first key in the block. No need
 			// to check the lower-bound again for this block.
@@ -298,15 +299,39 @@ func (i *singleLevelIterator) initBounds() {
 		}
 	}
 	i.blockUpper = i.upper
+	// TODO(radu): this should be >= 0 if blockUpper is inclusive.
 	if i.blockUpper != nil && i.cmp(i.blockUpper, i.index.Key().UserKey) > 0 {
 		// The upper-bound is greater than the index key which itself is greater
 		// than or equal to every key in the block. No need to check the
 		// upper-bound again for this block. Even if blockUpper is inclusive
 		// because of upper being inclusive, we can still safely set blockUpper
 		// to nil here.
-		//
-		// TODO(bananabrick): We could also set blockUpper to nil for the >=
-		// case, if blockUpper is inclusive.
+		i.blockUpper = nil
+	}
+}
+
+func (i *singleLevelIterator) initBoundsForAlreadyLoadedBlock() {
+	// TODO(radu): determine automatically if we need to call First or not and
+	// unify this function with initBounds().
+	if i.data.getFirstUserKey() == nil {
+		panic("initBoundsForAlreadyLoadedBlock must not be called on empty or corrupted block")
+	}
+	i.blockLower = i.lower
+	if i.blockLower != nil {
+		firstUserKey := i.data.getFirstUserKey()
+		// TODO(radu): this should be <= 0
+		if firstUserKey != nil && i.cmp(i.blockLower, firstUserKey) < 0 {
+			// The lower-bound is less than the first key in the block. No need
+			// to check the lower-bound again for this block.
+			i.blockLower = nil
+		}
+	}
+	i.blockUpper = i.upper
+	// TODO(radu): this should be >= 0 if blockUpper is inclusive.
+	if i.blockUpper != nil && i.cmp(i.blockUpper, i.index.Key().UserKey) > 0 {
+		// The upper-bound is greater than the index key which itself is greater
+		// than or equal to every key in the block. No need to check the
+		// upper-bound again for this block.
 		i.blockUpper = nil
 	}
 }
@@ -524,28 +549,6 @@ func (i *singleLevelIterator) resolveMaybeExcluded(dir int8) intersectsResult {
 	}
 	_, _ = i.index.Next()
 	return blockIntersects
-}
-
-func (i *singleLevelIterator) initBoundsForAlreadyLoadedBlock() {
-	if i.data.getFirstUserKey() == nil {
-		panic("initBoundsForAlreadyLoadedBlock must not be called on empty or corrupted block")
-	}
-	i.blockLower = i.lower
-	if i.blockLower != nil {
-		firstUserKey := i.data.getFirstUserKey()
-		if firstUserKey != nil && i.cmp(i.blockLower, firstUserKey) < 0 {
-			// The lower-bound is less than the first key in the block. No need
-			// to check the lower-bound again for this block.
-			i.blockLower = nil
-		}
-	}
-	i.blockUpper = i.upper
-	if i.blockUpper != nil && i.cmp(i.blockUpper, i.index.Key().UserKey) > 0 {
-		// The upper-bound is greater than the index key which itself is greater
-		// than or equal to every key in the block. No need to check the
-		// upper-bound again for this block.
-		i.blockUpper = nil
-	}
 }
 
 // The number of times to call Next/Prev in a block before giving up and seeking.
@@ -1351,8 +1354,8 @@ func (i *singleLevelIterator) Prev() (*InternalKey, base.LazyValue) {
 
 func (i *singleLevelIterator) skipForward() (*InternalKey, base.LazyValue) {
 	for {
-		var key *InternalKey
-		if key, _ = i.index.Next(); key == nil {
+		indexKey, _ := i.index.Next()
+		if indexKey == nil {
 			i.data.invalidate()
 			break
 		}
@@ -1363,7 +1366,7 @@ func (i *singleLevelIterator) skipForward() (*InternalKey, base.LazyValue) {
 			}
 			if result == loadBlockFailed {
 				// We checked that i.index was at a valid entry, so
-				// loadBlockFailed could not have happened due to to i.index
+				// loadBlockFailed could not have happened due to i.index
 				// being exhausted, and must be due to an error.
 				panic("loadBlock should not have failed with no error")
 			}
@@ -1374,7 +1377,7 @@ func (i *singleLevelIterator) skipForward() (*InternalKey, base.LazyValue) {
 			// separator, the same user key can span multiple blocks. If upper
 			// is exclusive we use >= below, else we use >.
 			if i.upper != nil {
-				cmp := i.cmp(key.UserKey, i.upper)
+				cmp := i.cmp(indexKey.UserKey, i.upper)
 				if (!i.endKeyInclusive && cmp >= 0) || cmp > 0 {
 					i.exhaustedBounds = +1
 					return nil, base.LazyValue{}
@@ -1382,7 +1385,42 @@ func (i *singleLevelIterator) skipForward() (*InternalKey, base.LazyValue) {
 			}
 			continue
 		}
-		if key, val := i.data.First(); key != nil {
+		var key *InternalKey
+		var val base.LazyValue
+		// It is possible that skipBackward went too far and the virtual table lower
+		// bound is after the first key in the block we are about to load, in which
+		// case we must use SeekGE.
+		//
+		// An example of how this can happen:
+		//
+		//   Data block 1 - contains keys a@1, c@1
+		//   Data block 2 - contains keys e@1, g@1
+		//   Data block 3 - contains keys i@2, k@2
+		//
+		//   The virtual table lower bound is f. We have a range key masking filter
+		//   that filters keys with @1 suffix. We are positioned inside block 3 then
+		//   we Prev(). Block 2 is entirely filtered out, which makes us move to
+		//   block 1. Now the range key masking filter gets an update (via
+		//   SpanChanged) and it no longer filters out any keys. At this point if a
+		//   Next happens, we will load block 2 but it would not be legal to return
+		//   "e@1" which is outside the virtual bounds.
+		//
+		//   The core of the problem is that skipBackward doesn't know it can stop
+		//   at block 2, because it doesn't know what keys are at the start of that
+		//   block. This is why we don't have this problem in the opposite
+		//   direction: skipForward will never go beyond the last relevant block
+		//   because it looks at the separator key which is an upper bound for the
+		//   block.
+		//
+		// Note that this is only a problem with virtual tables; we make no
+		// guarantees wrt an iterator lower bound when we iterate forward. But we
+		// must never return keys that are not inside the virtual table.
+		if i.vState != nil && i.blockLower != nil {
+			key, val = i.data.SeekGE(i.lower, base.SeekGEFlagsNone)
+		} else {
+			key, val = i.data.First()
+		}
+		if key != nil {
 			if i.blockUpper != nil {
 				cmp := i.cmp(key.UserKey, i.blockUpper)
 				if (!i.endKeyInclusive && cmp >= 0) || cmp > 0 {
@@ -1398,8 +1436,8 @@ func (i *singleLevelIterator) skipForward() (*InternalKey, base.LazyValue) {
 
 func (i *singleLevelIterator) skipBackward() (*InternalKey, base.LazyValue) {
 	for {
-		var key *InternalKey
-		if key, _ = i.index.Prev(); key == nil {
+		indexKey, _ := i.index.Prev()
+		if indexKey == nil {
 			i.data.invalidate()
 			break
 		}
@@ -1419,7 +1457,7 @@ func (i *singleLevelIterator) skipBackward() (*InternalKey, base.LazyValue) {
 			// bound is already exceeded. Note that the previous block starts with
 			// keys <= key.UserKey since even though this is the current block's
 			// separator, the same user key can span multiple blocks.
-			if i.lower != nil && i.cmp(key.UserKey, i.lower) < 0 {
+			if i.lower != nil && i.cmp(indexKey.UserKey, i.lower) < 0 {
 				i.exhaustedBounds = -1
 				return nil, base.LazyValue{}
 			}
--- a/sstable/reader_iter_two_lvl.go
+++ b/sstable/reader_iter_two_lvl.go
@@ -901,9 +901,42 @@ func (i *twoLevelIterator) skipForward() (*InternalKey, base.LazyValue) {
 		if i.err != nil || i.exhaustedBounds > 0 {
 			return nil, base.LazyValue{}
 		}
+
+		// It is possible that skipBackward went too far and the virtual table lower
+		// bound is after the first key in the block we are about to load, in which
+		// case we must use SeekGE below. The keys in the block we are about to load
+		// start right after the topLevelIndex key (before we Next).
+		//
+		// An example of how this can happen:
+		//
+		//   Second-level index block 1 - contains keys a@1, c@1
+		//   Second-level index block 2 - contains keys e@1, g@1
+		//   Second-level index block 3 - contains keys i@2, k@2
+		//
+		//   The virtual table lower bound is f. We have a range key masking filter
+		//   that filters keys with @1 suffix. We are positioned inside block 3 then
+		//   we Prev(). Block 2 is entirely filtered out, which makes us move to
+		//   block 1. Now the range key masking filter gets an update (via
+		//   SpanChanged) and it no longer filters out any keys. At this point if a
+		//   Next happens, we will load block 2 but it would not be legal to return
+		//   "e@1" which is outside the virtual bounds.
+		//
+		//   The core of the problem is that skipBackward doesn't know it can stop
+		//   at block 2, because it doesn't know what keys are at the start of that
+		//   block. This is why we don't have this problem in the opposite
+		//   direction: skipForward will never go beyond the last relevant block
+		//   because it looks at the separator key which is an upper bound for the
+		//   block.
+		//
+		// Note that this is only a problem with virtual tables; we make no
+		// guarantees wrt an iterator lower bound when we iterate forward. But we
+		// must never return keys that are not inside the virtual table.
+		useSeek := i.vState != nil &&
+			(!i.topLevelIndex.valid() || base.InternalCompare(i.cmp, i.topLevelIndex.ikey, i.vState.lower) < 0)
+
 		i.exhaustedBounds = 0
-		var ikey *InternalKey
-		if ikey, _ = i.topLevelIndex.Next(); ikey == nil {
+		topLevelKey, _ := i.topLevelIndex.Next()
+		if topLevelKey == nil {
 			i.data.invalidate()
 			i.index.invalidate()
 			return nil, base.LazyValue{}
@@ -913,7 +946,14 @@ func (i *twoLevelIterator) skipForward() (*InternalKey, base.LazyValue) {
 			return nil, base.LazyValue{}
 		}
 		if result == loadBlockOK {
-			if ikey, val := i.singleLevelIterator.firstInternal(); ikey != nil {
+			var ikey *InternalKey
+			var val base.LazyValue
+			if useSeek {
+				ikey, val = i.singleLevelIterator.SeekGE(i.lower, base.SeekGEFlagsNone)
+			} else {
+				ikey, val = i.singleLevelIterator.firstInternal()
+			}
+			if ikey != nil {
 				return i.maybeVerifyKey(ikey, val)
 			}
 			// Next iteration will return if singleLevelIterator set
@@ -927,7 +967,7 @@ func (i *twoLevelIterator) skipForward() (*InternalKey, base.LazyValue) {
 			// multiple index blocks. If upper is exclusive we use >=
 			// below, else we use >.
 			if i.upper != nil {
-				cmp := i.cmp(ikey.UserKey, i.upper)
+				cmp := i.cmp(topLevelKey.UserKey, i.upper)
 				if (!i.endKeyInclusive && cmp >= 0) || cmp > 0 {
 					i.exhaustedBounds = +1
 					// Next iteration will return.
@@ -943,8 +983,8 @@ func (i *twoLevelIterator) skipBackward() (*InternalKey, base.LazyValue) {
 			return nil, base.LazyValue{}
 		}
 		i.exhaustedBounds = 0
-		var ikey *InternalKey
-		if ikey, _ = i.topLevelIndex.Prev(); ikey == nil {
+		topLevelKey, _ := i.topLevelIndex.Prev()
+		if topLevelKey == nil {
 			i.data.invalidate()
 			i.index.invalidate()
 			return nil, base.LazyValue{}
@@ -954,9 +994,11 @@ func (i *twoLevelIterator) skipBackward() (*InternalKey, base.LazyValue) {
 			return nil, base.LazyValue{}
 		}
 		if result == loadBlockOK {
-			if ikey, val := i.singleLevelIterator.lastInternal(); ikey != nil {
+			ikey, val := i.singleLevelIterator.lastInternal()
+			if ikey != nil {
 				return i.maybeVerifyKey(ikey, val)
 			}
+
 			// Next iteration will return if singleLevelIterator set
 			// exhaustedBounds = -1.
 		} else {
@@ -966,7 +1008,7 @@ func (i *twoLevelIterator) skipBackward() (*InternalKey, base.LazyValue) {
 			// the previous entry starts with keys <= ikey.UserKey since even
 			// though this is the current block's separator, the same user key
 			// can span multiple index blocks.
-			if i.lower != nil && i.cmp(ikey.UserKey, i.lower) < 0 {
+			if i.lower != nil && i.cmp(topLevelKey.UserKey, i.lower) < 0 {
 				i.exhaustedBounds = -1
 				// Next iteration will return.
 			}
--- a/sstable/reader_test.go
+++ b/sstable/reader_test.go
@@ -178,6 +178,23 @@ func (i *iterAdapter) SetContext(ctx context.Context) {
 }
 
 func TestVirtualReader(t *testing.T) {
+	t.Run("props", func(t *testing.T) {
+		runVirtualReaderTest(t, "testdata/virtual_reader_props", 0 /* blockSize */, 0 /* indexBlockSize */)
+	})
+	t.Run("iter", func(t *testing.T) {
+		for run := 0; run < 100; run++ {
+			var blockSize, indexBlockSize int
+			if run > 0 {
+				blockSize = rand.Intn(200)
+				indexBlockSize = rand.Intn(200)
+			}
+			t.Logf("run %d: blockSize=%d indexBlockSize=%d", run, blockSize, indexBlockSize)
+			runVirtualReaderTest(t, "testdata/virtual_reader_iter", blockSize, indexBlockSize)
+		}
+	})
+}
+
+func runVirtualReaderTest(t *testing.T, path string, blockSize, indexBlockSize int) {
 	// A faux filenum used to create fake filemetadata for testing.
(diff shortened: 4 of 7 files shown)


RaduBerinde:
There is more subtlety around how this can happen. We have a `rangeKeyMasking` filter. On the last `Prev` call, the filter causes a second-level index block to be skipped. Then the `rangeKeyMasking` filter is updated, and on the `Next` cal, the block is no longer skipped.

jbowens (commented):
> There is more subtlety around how this can happen. We have a `rangeKeyMasking` filter. On the last `Prev` call, the filter causes a second-level index block to be skipped. Then the `rangeKeyMasking` filter is updated, and on the `Next` cal, the block is no longer skipped.

This makes sense. Range key masking applies within bounds `[s, e)`. When performing forward iteration, the sstable iterator is only responsible for comparing the current key to `e`. It assumes that the current key is ≥ `s`, because it's supposed to be externally guaranteed by the higher iterator stack. By moving the iterator onto a key before the lower virtual bound, we're violating the externally-guaranteed invariant.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451#-:-NtqCmRj4mXZ8qBbtkcH:b-c1wzb)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`sstable/reader_iter_two_lvl.go` line 909 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtqGKp71ThPwKJInGTS:-NtqGKp71ThPwKJInGTT:b-63wc3b) ([raw file](https://github.com/cockroachdb/pebble/blob/ba8c69f8f714bac2f23b872d67a1f8dd3a6e13b6/sstable/reader_iter_two_lvl.go#L909)):*
> ```Go
> 		// i.topLevelIndex.key. If the lower bound cuts through this block, we must
> 		// use SeekGE below.
> 		lowerBoundInsideBlock := i.lower != nil && i.cmp(i.topLevelIndex.ikey.UserKey, i.lower) < 0
> ```

I think we can restrict these key comparisons to when `i.virtualState != nil` because non-virtual bounds are known to the higher iterator stack, and the internal iterator interface indicates that they must translate user operations into the appropriate bounded operation (to prevent exactly these types of comparisons within leaves):

```
// Additionally, the caller needs to ensure that SeekGE/SeekPrefixGE are not
// called with a key > the upper bound, and SeekLT is not called with a key <
// the lower bound. InternalIterator implementations are required to respect
// the iterator bounds, never returning records outside of the bounds with one
// exception: an iterator may generate synthetic RANGEDEL marker records. See
// levelIter.syntheticBoundary for the sole existing example of this behavior.
// Specifically, levelIter can return synthetic keys whose user key is equal to
// the lower/upper bound.
```

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal, @jbowens, and @sumeerbhola)

___
*[`sstable/reader_iter_two_lvl.go` line 909 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtqGKp71ThPwKJInGTS:-NtqZMjs26JBf98nPoaB:b3k0okq) ([raw file](https://github.com/cockroachdb/pebble/blob/ba8c69f8f714bac2f23b872d67a1f8dd3a6e13b6/sstable/reader_iter_two_lvl.go#L909)):*

I think we can restrict these key comparisons to when `i.virtualState != nil` because non-virtual bounds are known to the higher iterator stack, and the internal iterator interface indicates that they must translate user operations into the appropriate bounded operation (to prevent exactly these types of comparisons within leaves):

```
// Additionally, the caller needs to ensure that SeekGE/SeekPrefixGE are not
// called with a key > the upper bound, and SeekLT is not called with a key <
// the lower bound. InternalIterator implementations are required to respect
// the iterator bounds, never returning records outside of the bounds with one
// exception: an iterator may generate synthetic RANGEDEL marker records. See
// levelIter.syntheticBoundary for the sole existing example of this behavior.
// Specifically, levelIter can return synthetic keys whose user key is equal to
// the lower/upper bound.
```
</blockquote></details>

That makes sense, I'll add the `virtualState` check. I am working separately on a test for this, I need to extend things so I can mess with the range masking during iteration.

itsbilal (commented):
Code LGTM, just waiting on tests and then I think this is gucci

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/reader_iter_two_lvl.go` line 974 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtrNIEz-Sx11oHoGdO2:-NtrNIF-0eFKTYdDWIYq:b98gth0) ([raw file](https://github.com/cockroachdb/pebble/blob/5473ca25dfc475fa9855f2eb25fc2b031dd2db4d/sstable/reader_iter_two_lvl.go#L974)):*
> ```Go
> 			var ikey *InternalKey
> 			var val base.LazyValue
> 			// The keys in the block we just loaded end at topLevelKey. If the upper
> ```

I am not sure this case is needed - at least I can't reproduce a similar failure.

I think the asymmetry here comes from separator keys - when we do `skipBackward`, we stop when `i.cmp(topLevelKey.UserKey, i.lower) < 0`, which is the block *before* the first relevant block (causing issues on the next `skipForward`).

But when we do `skipForward`, we stop when `i.cmp(topLevelKey.UserKey, i.upper) > 0` which is the last relevant block. There might still be an edge case when `UserKey == i.upper`, not sure.


Another difficulty I'm running into w.r.t testing is that I don't think the result is deterministic when we mess with masking.

For example, say the table has keys `a@2 b@2 c@1 d@1`. We set up mask suffix @2, do `First() -> a@2, then Next() -> b@2(), then Next() -> nil`. Then we change the mask suffix to @1 and do `Prev()`. What is this `Prev()` supposed to return? In practice, I believe it could be either `d@1` or `c@1` or `b@2` depending how index blocks line up.

RaduBerinde:
*[`sstable/reader_iter_two_lvl.go` line 910 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtrajUd9Rto16bPH33q:-NtrajUd9Rto16bPH33r:b94fyxd) ([raw file](https://github.com/cockroachdb/pebble/blob/5473ca25dfc475fa9855f2eb25fc2b031dd2db4d/sstable/reader_iter_two_lvl.go#L910)):*
> ```Go
> 		// use SeekGE below. This is only relevant for virtual tables; iterator
> 		// lower bounds are not enforced during forward iteration.
> 		lowerBoundInsideBlock := i.vState != nil && i.lower != nil && i.cmp(i.topLevelIndex.ikey.UserKey, i.lower) < 0
> ```

I don't think this would work if topLevelIndex is before the first key.. My testing found that we need a similar fix in the single level iterator and doing the same doesn't work.

RaduBerinde:
Updated. I removed the changes to `skipBackward` as I was not able to reproduce a problem there (we'll see if the meta test will run into something).
I added some commits that cleanup and refactor the tests. We now run the iterator tests with many random block sizes. The last test I added breaks without the changes in the two `skipForward` methods.

sumeerbhola (approved):
:lgtm: 

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451#-:-NtvNgQ70Y-8n3-hNPS8:bg7igkv)* status: 2 of 6 files reviewed, 11 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`-- commits` line 8 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-Ntv17QN3FYOfw1k_nTR):*
nit: first
___
*[`-- commits` line 13 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-Ntv1B7r0O6XQQji7DdV):*
ditto
___
*[`sstable/reader_iter_single_lvl.go` line 1355 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-Ntv1ZRM4xDpmhcg1mvM:-Ntv1ZRM4xDpmhcg1mvN:b9vkhgk) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_single_lvl.go#L1355)):*
> ```Go
> 	for {
> 		// It is possible that skipBackward went too far and the virtual table lower
> 		// bound cuts through block we are about to load, in which case we must use
> ```

nit: "cuts through" isn't super clear. How about using some more words and saying something like "lower bound is after the first key in the block we are ...". And consider adding a couple of examples of how this can happen, so it's clearer why we are only considering this for virtual tables, and why we need both the conditions in the disjunction.
___
*[`sstable/reader_iter_single_lvl.go` line 1358 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-Ntv3fLE6m4CqrjckhXT:-Ntv3fLE6m4CqrjckhXU:b-ncuk5h) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_single_lvl.go#L1358)):*
> ```Go
> 		// SeekGE below. The keys in the block we are about to load start right
> 		// after the topLevelIndex key (before we Next).
> 		lowerBoundInsideBlock := i.vState != nil &&
> ```

nit: lowerBoundMayBeInsideBlock?
since this is a necessary condition but not sufficient?
___
*[`sstable/reader_iter_single_lvl.go` line 1360 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvLpo4CY8Kg6xc-KDv:-NtvLpo4CY8Kg6xc-KDw:bev7e6s) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_single_lvl.go#L1360)):*
> ```Go
> 		lowerBoundInsideBlock := i.vState != nil &&
> 			(!i.index.valid() || base.InternalCompare(i.cmp, i.index.ikey, i.vState.lower) < 0)
> ```

singleLevelIterator.skipForward is potentially on the performance sensitive path since it is called whenever we transition blocks in `Next` or `NextPrefix`, and for large key-value pairs that could be very frequent (maybe even after every key-value). In comparison, we shouldn't be transitioning first level index blocks often so not worth worrying about twoLevelIterator.
We've added a key comparison for virtual sstables, which can be expensive for large keys. I wonder whether it is optimizable, in that if `i.data.valid()` before we called `i.data.Next()` in https://github.com/cockroachdb/pebble/blob/9c2da6d20e1178d0bf383f5b848eb1ded43db2de/sstable/reader_iter_single_lvl.go#L1238 then we know we don't need to do this key comparison. So it could be a parameter, say `lowerBoundKnownBeforeBlock` to `skipForward`. Similar before calling `i.data.NextPrefix`.

I am ok with it being a TODO.
___
*[`sstable/reader_iter_single_lvl.go` line 1384 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-Ntv9fAk3TYYVM5caaRh:-Ntv2lNq26l98s_p0OzT:bn0j4t) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_single_lvl.go#L1384)):*
> ```Go
> 			// is exclusive we use >= below, else we use >.
> 			if i.upper != nil {
> 				cmp := i.cmp(indexKey.UserKey, i.upper)
> ```

Just thinking out aloud: if this block contained a#10, b#20, d#30, we are at risk of the same kind of bug if the last key d#30 does not satisfy the (inclusive or exclusive) upper bound. But in that case the separator will also not satisfy the upper bound.
I think this is a worth a code comment somewhere so that the reader can quickly see why we need this "inside block" logic for `skipForward` and not `skipBackward`.

There is also something slightly subtle about this in that we decided not to load the block, and there is potentially a key in there that violates the upper bound, but that is ok since we will never load this block in the future and start from the end. Because if we did, we would need to check the upper bound.
___
*[`sstable/reader_iter_two_lvl.go` line 907 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvBSPh2I00L1jUUCDa:-NtvBSPh2I00L1jUUCDb:b-h2yl7r) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_two_lvl.go#L907)):*
> ```Go
> 
> 		// It is possible that skipBackward went too far and the virtual table lower
> 		// bound cuts through block we are about to load, in which case we must use
> ```

same nit about "cuts through"
___
*[`sstable/reader_iter_two_lvl.go` line 909 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvBgrT3PvrA_kQPpHV:-NtvBgrT3PvrA_kQPpHW:b-264ntu) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_two_lvl.go#L909)):*
> ```Go
> 		// bound cuts through block we are about to load, in which case we must use
> 		// SeekGE below. The keys in the block we are about to load start right
> 		// after the topLevelIndex key (before we Next).
> ```

right after or equal to the topLevelIndex key ...

(since the same user key can span blocks)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451)* status: 2 of 6 files reviewed, 12 unresolved discussions (waiting on @jbowens)

___
*[`sstable/testdata/virtual_reader_iter` line 664 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvPSa16EE3kM7dsDkY:-NtvSIvN2EV7M_Eopsy3:bnvdna0) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/testdata/virtual_reader_iter#L664)):*

@jbowens 
 
I am unclear about what real-world situation causes the bug that motivated the PR.
Say the range key masking filter caused block B2 to be irrelevant when doing Prev, and we moved the index to B1 and stopped because nothing in B1 satisfied the lower bound. Then when we do Next, why does the range key masking filter not make block B2 irrelevant again?

Is this because, the SpanChanged to activate the block masking happens after B2 is loaded (from the following comment). That is, we made the decision (I am vaguely remembering some discussions from a long time ago) that block masking does not need to be consistent on whether it masks a block because it is a performance optimization.

```
// The logic to ensure that a mask block-property filter is only applied within
// the bounds of the masking range key is subtle. The interleaving iterator
// guarantees that it never invokes SpanChanged until the point iterator is
// positioned within the range key. During forward iteration, this guarantees
// that any block that a sstable reader might attempt to load contains only keys
// greater than or equal to the range key's lower bound. During backward
// iteration, it provides the analagous guarantee on the range key's upper
// bound.
```
</blockquote></details>

Correct, we do `Prev` which returns nil, then `SpanChanged` changes the filtering, then the block is not longer filtered out when we do `Next`. This corresponds to the second `mask-suffix` directive in the test.

RaduBerinde:
*[`sstable/testdata/virtual_reader_iter` line 664 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvPSa16EE3kM7dsDkY:-NtvSl3YC-6aendoppSk:bjc72x1) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/testdata/virtual_reader_iter#L664)):*

Correct, we do `Prev` which returns nil, then `SpanChanged` changes the filtering, then the block is not longer filtered out when we do `Next`. This corresponds to the second `mask-suffix` directive in the test.
</blockquote></details>

IIRC SpanChanged was called with `nil` at that point

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451)* status: 0 of 6 files reviewed, 11 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`-- commits` line 13 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-Ntv1B7r0O6XQQji7DdV):*

ditto
</blockquote></details>

Done.
___
*[`sstable/reader_iter_single_lvl.go` line 1355 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-Ntv1ZRM4xDpmhcg1mvM:-Ntvf2apEhKyEf7vtqMD:b-ya4zpv) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_single_lvl.go#L1355)):*

nit: "cuts through" isn't super clear. How about using some more words and saying something like "lower bound is after the first key in the block we are ...". And consider adding a couple of examples of how this can happen, so it's clearer why we are only considering this for virtual tables, and why we need both the conditions in the disjunction.
</blockquote></details>

Improved the comment.
___
*[`sstable/reader_iter_single_lvl.go` line 1358 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-Ntv3fLE6m4CqrjckhXT:-Ntv_ITZ5YK6ho9ycKNl:b-dmto79) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_single_lvl.go#L1358)):*

nit: lowerBoundMayBeInsideBlock?
since this is a necessary condition but not sufficient?
</blockquote></details>

Renamed to `useSeek`
___
*[`sstable/reader_iter_single_lvl.go` line 1360 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvLpo4CY8Kg6xc-KDv:-Ntvf4Rx07QmS4xPzkRN:b8ap5ju) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_single_lvl.go#L1360)):*

singleLevelIterator.skipForward is potentially on the performance sensitive path since it is called whenever we transition blocks in `Next` or `NextPrefix`, and for large key-value pairs that could be very frequent (maybe even after every key-value). In comparison, we shouldn't be transitioning first level index blocks often so not worth worrying about twoLevelIterator.
We've added a key comparison for virtual sstables, which can be expensive for large keys. I wonder whether it is optimizable, in that if `i.data.valid()` before we called `i.data.Next()` in https://github.com/cockroachdb/pebble/blob/9c2da6d20e1178d0bf383f5b848eb1ded43db2de/sstable/reader_iter_single_lvl.go#L1238 then we know we don't need to do this key comparison. So it could be a parameter, say `lowerBoundKnownBeforeBlock` to `skipForward`. Similar before calling `i.data.NextPrefix`.

I am ok with it being a TODO.
</blockquote></details>

I thought about it and we can elide the check if we have a valid data block, which is the case when we transition blocks. Similarly, in the second-level iterator we can elide it if we have a valid index block.
___
*[`sstable/reader_iter_single_lvl.go` line 1384 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-Ntv9fAk3TYYVM5caaRh:-NtvfPZZ1i3f6riVoqIA:b7rny47) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_single_lvl.go#L1384)):*

Just thinking out aloud: if this block contained a#10, b#20, d#30, we are at risk of the same kind of bug if the last key d#30 does not satisfy the (inclusive or exclusive) upper bound. But in that case the separator will also not satisfy the upper bound.
I think this is a worth a code comment somewhere so that the reader can quickly see why we need this "inside block" logic for `skipForward` and not `skipBackward`.

There is also something slightly subtle about this in that we decided not to load the block, and there is potentially a key in there that violates the upper bound, but that is ok since we will never load this block in the future and start from the end. Because if we did, we would need to check the upper bound.
</blockquote></details>

I added to the comment in `skipForward` explaining why we don't have the symmetrical problem.
___
*[`sstable/reader_iter_two_lvl.go` line 907 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvBSPh2I00L1jUUCDa:-NtvfXC011l4_nkI1LpI:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_two_lvl.go#L907)):*

same nit about "cuts through"
</blockquote></details>

Done.
___
*[`sstable/reader_iter_two_lvl.go` line 909 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvBgrT3PvrA_kQPpHV:-NtvfXT30OjmYW7COYgL:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_two_lvl.go#L909)):*

right after or equal to the topLevelIndex key ...

(since the same user key can span blocks)
</blockquote></details>

Done.

RaduBerinde:
*[`sstable/reader_iter_single_lvl.go` line 1360 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvLpo4CY8Kg6xc-KDv:-NtvhCsJ0sV4OklQE-7a:bnir0iq) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_single_lvl.go#L1360)):*

I thought about it and we can elide the check if we have a valid data block, which is the case when we transition blocks. Similarly, in the second-level iterator we can elide it if we have a valid index block.
</blockquote></details>

Oh wait, you're right, I'd need to pass through the value of `data.valid()` before that `Next` call.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451)* status: 0 of 6 files reviewed, 11 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 1360 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvLpo4CY8Kg6xc-KDv:-NtvhyfZ-75c0HScSOJB:b-1mt0yr) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_single_lvl.go#L1360)):*

Oh wait, you're right, I'd need to pass through the value of `data.valid()` before that `Next` call.
</blockquote></details>

Ok, fixed, I used `isDataInvalidated` which means we have no block at all, not that we merely finished iterating it.

jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451#-:-Ntvvoin-VDgLS3fGWnh:bnfp4nl)* status: 0 of 6 files reviewed, 11 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`-- commits` line 48 at r18](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvqQcA-lKfANmr9wHJ):*
s/interator/iterator/
___
*[`sstable/testdata/virtual_reader_iter` line 664 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvPSa16EE3kM7dsDkY:-NtvueqG4TTW2ZeB_65M:bxldwpa) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/testdata/virtual_reader_iter#L664)):*

IIRC SpanChanged was called with `nil` at that point
</blockquote></details>

I forgot about this asymmetry. We require that the iterator observe a key within the range key's bounds (so that we know that the iterator has stepped into the bounds) before we begin applying masking. Asymmetry is a consequence of that decision.

sumeerbhola (changes requested):
> Reviewed 2 of 6 files at r15, 1 of 4 files at r18.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451)* status: 3 of 6 files reviewed, 14 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 1360 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvLpo4CY8Kg6xc-KDv:-NtvyIPe-vhDIqpdPsjh:bpg9he5) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_single_lvl.go#L1360)):*

Ok, fixed, I used `isDataInvalidated` which means we have no block at all, not that we merely finished iterating it.
</blockquote></details>

I don't quite see why this is correct. And my previous suggestion is probably wrong too.

If block B2 was irrelevant in skipBackward, we could have loaded B1 (which is not masked) and has all keys less than the lower bound. We would do
```
		key, val := i.data.Last()
		if key == nil {
			return nil, base.LazyValue{}
		}
		if i.blockLower != nil && i.cmp(key.UserKey, i.blockLower) < 0 {
			i.exhaustedBounds = -1
			return nil, base.LazyValue{}
		}
```
So i.data is valid.

I think we could make it work by explicitly grabbing `lowerBoundKnownToBeBeforeBlock = i.exhaustedBounds == 0` at the start of `Next`. I am also ok with just merging without optimizing with a TODO.
___
*[`sstable/reader_iter_single_lvl.go` line 1361 at r18](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvwoCvEtE6_BruVOyM:-NtvwoCvEtE6_BruVOyN:bmc0t1c) ([raw file](https://github.com/cockroachdb/pebble/blob/b96fb5789d3b41c47140abd7daea221c2fafc3b4/sstable/reader_iter_single_lvl.go#L1361)):*
> ```Go
> 		// An example of how this can happen:
> 		//
> 		//   Index block 1 - contains keys a@1, c@1
> ```

block 1?
since this is a single level iterator, so only knows of a single index block.
___
*[`sstable/reader_iter_two_lvl.go` line 912 at r18](https://reviewable.io/reviews/cockroachdb/pebble/3451#-Ntw-Z802mIl49mGg13O:-Ntw-Z802mIl49mGg13P:b-5xb5zu) ([raw file](https://github.com/cockroachdb/pebble/blob/b96fb5789d3b41c47140abd7daea221c2fafc3b4/sstable/reader_iter_two_lvl.go#L912)):*
> ```Go
> 		// An example of how this can happen:
> 		//
> 		//   Top level index block 1 - contains keys a@1, c@1
> ```

This is the second level index block 1, yes?
Which represents a tree that contains these keys.
That is, this comment is talking about the top level index block entry.
___
*[`sstable/reader_iter_two_lvl.go` line 936 at r18](https://reviewable.io/reviews/cockroachdb/pebble/3451#-Ntw0KoUDO6ISkh2CanH:-Ntw0KoUDO6ISkh2CanI:bn2cdhc) ([raw file](https://github.com/cockroachdb/pebble/blob/b96fb5789d3b41c47140abd7daea221c2fafc3b4/sstable/reader_iter_two_lvl.go#L936)):*
> ```Go
> 		useSeek := i.vState != nil &&
> 			(!i.topLevelIndex.valid() ||
> 				(i.index.isDataInvalidated() && base.InternalCompare(i.cmp, i.topLevelIndex.ikey, i.vState.lower) < 0))
> ```

I think the same issue happens here wrt the all these blocks including the data block being successfully loaded, since `twoLevelIterator.skipBackward` will call `i.singleLevelIterator.lastInternal()`. Also, I worry about getting too clever, and would slightly prefer not to try to optimize this twoLevelIterator case.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451)* status: 3 of 6 files reviewed, 14 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 1360 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvLpo4CY8Kg6xc-KDv:-Ntw6dcE2ivxDvIxRQAX:b-tqsve8) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_single_lvl.go#L1360)):*

I don't quite see why this is correct. And my previous suggestion is probably wrong too.

If block B2 was irrelevant in skipBackward, we could have loaded B1 (which is not masked) and has all keys less than the lower bound. We would do
```
		key, val := i.data.Last()
		if key == nil {
			return nil, base.LazyValue{}
		}
		if i.blockLower != nil && i.cmp(key.UserKey, i.blockLower) < 0 {
			i.exhaustedBounds = -1
			return nil, base.LazyValue{}
		}
```
So i.data is valid.

I think we could make it work by explicitly grabbing `lowerBoundKnownToBeBeforeBlock = i.exhaustedBounds == 0` at the start of `Next`. I am also ok with just merging without optimizing with a TODO.
</blockquote></details>

Ah.. you are correct. I made one of the keys before the sstable in the test `@2` and the panic triggered. I switched to making it a TODO for now. I will try to find a clean way of doing it separately.
___
*[`sstable/reader_iter_two_lvl.go` line 912 at r18](https://reviewable.io/reviews/cockroachdb/pebble/3451#-Ntw-Z802mIl49mGg13O:-Ntw4ZmS7IGhJPU80t_K:bohwc58) ([raw file](https://github.com/cockroachdb/pebble/blob/b96fb5789d3b41c47140abd7daea221c2fafc3b4/sstable/reader_iter_two_lvl.go#L912)):*

This is the second level index block 1, yes?
Which represents a tree that contains these keys.
That is, this comment is talking about the top level index block entry.
</blockquote></details>

Yes, it says "top level index block" :)

RaduBerinde:
*[`sstable/reader_iter_single_lvl.go` line 1360 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvLpo4CY8Kg6xc-KDv:-NtwA5_N2r5K9O-fkZfK:b-g2jjms) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_single_lvl.go#L1360)):*

Ah.. you are correct. I made one of the keys before the sstable in the test `@2` and the panic triggered. I switched to making it a TODO for now. I will try to find a clean way of doing it separately.
</blockquote></details>

Also I think if the skipped block was the first block of the physical sst, `exhaustedBounds` would be 0. So I think the correct check is `i.data.isDataInvalidated() || exhaustedBounds was -1`

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451)* status: 0 of 6 files reviewed, 14 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 1361 at r18](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvwoCvEtE6_BruVOyM:-NtwA0GEEQzcoKHyXcgc:bt2vr1b) ([raw file](https://github.com/cockroachdb/pebble/blob/b96fb5789d3b41c47140abd7daea221c2fafc3b4/sstable/reader_iter_single_lvl.go#L1361)):*

block 1?
since this is a single level iterator, so only knows of a single index block.
</blockquote></details>

Fixed.
___
*[`sstable/reader_iter_two_lvl.go` line 912 at r18](https://reviewable.io/reviews/cockroachdb/pebble/3451#-Ntw-Z802mIl49mGg13O:-NtwCCQ-9L6-pXsoXJ7t:b-efess6) ([raw file](https://github.com/cockroachdb/pebble/blob/b96fb5789d3b41c47140abd7daea221c2fafc3b4/sstable/reader_iter_two_lvl.go#L912)):*

Yes, it says "top level index block" :)
</blockquote></details>

Oh sorry, I missed the "entry" part. Yes, updated to `Second-level index block X`.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451)* status: 0 of 7 files reviewed, 14 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 1360 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtvLpo4CY8Kg6xc-KDv:-NtwOuYy0iCz6-leZVCx:blu5krz) ([raw file](https://github.com/cockroachdb/pebble/blob/0e8608097efe733fd47ffa38a2c1ccb81dd9748f/sstable/reader_iter_single_lvl.go#L1360)):*

Also I think if the skipped block was the first block of the physical sst, `exhaustedBounds` would be 0. So I think the correct check is `i.data.isDataInvalidated() || exhaustedBounds was -1`
</blockquote></details>

Ok, I found a very nice way to do it in the single level iterator. We already calculate the `blockLower` and set it if the lower bound intersects the block, so we can just use that. PTAL.

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 6 files at r24, 1 of 1 files at r25, 1 of 5 files at r28, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451#-:-Ntw_aBp0RdTFf4ZUE9P:bg7igkv)* status: 3 of 7 files reviewed, 15 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 1415 at r28](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtwZljkCNqrYjEXDLwt:-NtwZljkCNqrYjEXDLwu:b-hhr6vm) ([raw file](https://github.com/cockroachdb/pebble/blob/780e5edf308b9849b8cf408f5f3bffe03705256d/sstable/reader_iter_single_lvl.go#L1415)):*
> ```Go
> 		// guarantees wrt an iterator lower bound when we iterate forward. But we
> 		// must never return keys that are not inside the virtual table.
> 		if i.vState != nil && i.blockLower != nil {
> ```

looks good. Our `blockLower` initialization logic seems unnecessarily conservative:
```
	if i.blockLower != nil {
		key, _ := i.data.First()
		if key != nil && i.cmp(i.blockLower, key.UserKey) < 0 {
			// The lower-bound is less than the first key in the block. No need
			// to check the lower-bound again for this block.
			i.blockLower = nil
		}
	}
```
Shouldn't that be `<= 0`? Same for `initBoundsForAlreadyLoadedBlock`.

How about adding a TODO to change it -- I don't want you to have to kickoff another multi-hour meta test run for merging this PR, it it already passes.

RaduBerinde (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451#-:-NtxZF9v8HeZKtOpa6V6:bz5mair)* status: 3 of 7 files reviewed, 15 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 1415 at r28](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtwZljkCNqrYjEXDLwt:-NtxZ3EyEhg9ybu9ZEw_:b-hukn7c) ([raw file](https://github.com/cockroachdb/pebble/blob/780e5edf308b9849b8cf408f5f3bffe03705256d/sstable/reader_iter_single_lvl.go#L1415)):*

looks good. Our `blockLower` initialization logic seems unnecessarily conservative:
```
	if i.blockLower != nil {
		key, _ := i.data.First()
		if key != nil && i.cmp(i.blockLower, key.UserKey) < 0 {
			// The lower-bound is less than the first key in the block. No need
			// to check the lower-bound again for this block.
			i.blockLower = nil
		}
	}
```
Shouldn't that be `<= 0`? Same for `initBoundsForAlreadyLoadedBlock`.

How about adding a TODO to change it -- I don't want you to have to kickoff another multi-hour meta test run for merging this PR, it it already passes.
</blockquote></details>

Added TODOs

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 6 files at r29, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451#-:-NtxjSd6ETuiQySO0C10:bg7igkv)* status: 2 of 7 files reviewed, 16 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 302 at r32](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtxieeK6UNBbm3REBFA:-NtxieeK6UNBbm3REBFB:b-cvau5h) ([raw file](https://github.com/cockroachdb/pebble/blob/6f655895be550ed1663a6e67dcf1193b54bc4713/sstable/reader_iter_single_lvl.go#L302)):*
> ```Go
> 	}
> 	i.blockUpper = i.upper
> 	// TODO(radu): this should be >= 0
> ```

`>=0` only if `i.endKeyInclusive` is true.
I think our current code where we have to remember to use both `i.upper` and `i.endKeyInclusive` is asking for bugs. We should (not in this PR of course) try to hide comparisons involving the upper bound behind a method so we don't forget to use both fields.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3451)* status: 1 of 7 files reviewed, 16 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 302 at r32](https://reviewable.io/reviews/cockroachdb/pebble/3451#-NtxieeK6UNBbm3REBFA:-Ntxk1B81qTtUAgO900n:b568vc4) ([raw file](https://github.com/cockroachdb/pebble/blob/6f655895be550ed1663a6e67dcf1193b54bc4713/sstable/reader_iter_single_lvl.go#L302)):*

`>=0` only if `i.endKeyInclusive` is true.
I think our current code where we have to remember to use both `i.upper` and `i.endKeyInclusive` is asking for bugs. We should (not in this PR of course) try to hide comparisons involving the upper bound behind a method so we don't forget to use both fields.
</blockquote></details>

Yeah, we can use the new `UserKeyBoundary`

PR #3454, opened 2024-03-26
metamorphic: add a missing simulated excise

RaduBerinde:
We weren't simulating the excise when the batch wasn't empty but it
yielded an empty sst (e.g. it only contained a `LogData`). Fix that
and refactor the code a bit to avoid three copies of this code.

Informs #3448


itsbilal (approved):
LGTM, thanks for catching this! These empty no-op cases were always hairy.

RaduBerinde:
TFTR!

PR #3462, opened 2024-03-27
ingest: fix ingest-split with loose bounds on split candidate

### LINKED ISSUE #3456: TestMeta failure: did not split file that was expected to be split
This one is pretty rare, the test ran a lot before it was hit:
```
1706 runs so far, 0 failures, over 4h33m40s
```

itsbilal:
Previously, if we had an ingestion file that had an inclusive upper bound, and an existing file with a loose lower bound, we'd correctly see no data overlap on that existing file and suggest an ingest-split. However, if the upper bound of the incoming sstable equals the loose lower bound in user keys, the excise() method would be a no-op as it would expect the upper bound of the incoming sstable to be exclusive.

This change updates db.excise() to take in UserKeyBounds with info on whether the bound is exclusive or inclusive. For inclusive bounds, we have additional guards to ensure we don't have data overlap at the upper bound, as this would only happen in the ingest-split case.

Fixes #3456.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -1820,11 +1820,12 @@ func (d *DB) ingest(
 //
 // The manifest lock must be held when calling this method.
 func (d *DB) excise(
-	exciseSpan KeyRange, m *fileMetadata, ve *versionEdit, level int,
+	exciseSpan base.UserKeyBounds, m *fileMetadata, ve *versionEdit, level int,
 ) ([]manifest.NewFileEntry, error) {
 	numCreatedFiles := 0
 	// Check if there's actually an overlap between m and exciseSpan.
-	if !exciseSpan.Overlaps(d.cmp, m) {
+	mBounds := base.UserKeyBoundsFromInternal(m.Smallest, m.Largest)
+	if !exciseSpan.Overlaps(d.cmp, &mBounds) {
 		return nil, nil
 	}
 	ve.DeletedFiles[deletedFileEntry{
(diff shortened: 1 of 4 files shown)


ingest.go
```diff
@@ -2018,20 +2019,27 @@ func (d *DB) excise(
 				rangeDelIter = emptyKeyspanIter
 			}
 		}
-		key, _ := iter.SeekGE(exciseSpan.End, base.SeekGEFlagsNone)
+		key, _ := iter.SeekGE(exciseSpan.End.Key, base.SeekGEFlagsNone)
 		if key != nil {
+			if exciseSpan.End.Kind == base.Inclusive && d.equal(exciseSpan.End.Key, key.UserKey) {
+				return nil, errors.New("cannot excise with an inclusive end key and data overlap at end key")
```
itsbilal, line 2025: maybe these should be panics

RaduBerinde (approved):
:lgtm:  Nice! Thanks for the fix!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3462#-:-Nu-xsVs51FKTXanzlzH:b-a22aro)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`ingest.go` line 2025 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3462#gh-1541539819:-Nu-xaCH0emxFl6k4o9R:b-pgim1f) ([raw file](https://github.com/cockroachdb/pebble/blob/3fe550cb581ccd4777d33847150c14ca134c3b41/ingest.go#L2025)):*

maybe these should be panics
</blockquote></details>

Let's make it an "errors.AssertionFailedf" for now. I am thinking of adding a wrapper for that that panics in invariants.Enabled mode.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3462#-:-Nu-z6KACRPuoyLurqyw:b-814g8k)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @RaduBerinde)

___
*[`ingest.go` line 2025 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3462#gh-1541539819:-Nu-z7NWCRQSh1Jd24yx:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/3fe550cb581ccd4777d33847150c14ca134c3b41/ingest.go#L2025)):*

Let's make it an "errors.AssertionFailedf" for now. I am thinking of adding a wrapper for that that panics in invariants.Enabled mode.
</blockquote></details>

Done.

PR #3464, opened 2024-03-27
sstable: handle synthetic suffix when using CopySpan

### LINKED ISSUE #3459: metamorphic: CopySpan cannot copy an empty span
Ran into this one. This is proof that we can end up with an empty external virtual table even if we never ingest an empty one (CC @dt @msbutler @itsbilal).

RaduBerinde:
We were ignoring the synthetic suffix when using `CopySpan`. This can
cause it to miss keys or generate the "empty span" error.

We handle the synthetic suffix by using suffix-free keys, extending
the bounds as necessary.

Fixes #3459

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2627,6 +2627,13 @@ func (d *DB) runCopyCompaction(
 			start.UserKey = newMeta.SyntheticPrefix.Invert(start.UserKey)
 			end.UserKey = newMeta.SyntheticPrefix.Invert(end.UserKey)
 		}
+		if newMeta.SyntheticSuffix.IsSet() {
+			// Extend the bounds as necessary so that the keys don't include suffixes.
+			start.UserKey = start.UserKey[:c.comparer.Split(start.UserKey)]
+			if n := c.comparer.Split(end.UserKey); n < len(end.UserKey) {
+				end = base.MakeRangeDeleteSentinelKey(c.comparer.ImmediateSuccessor(nil, end.UserKey[:n]))
+			}
+		}
 
 		wrote, err := sstable.CopySpan(ctx,
 			src, d.opts.MakeReaderOptions(),
--- a/sstable/copier.go
+++ b/sstable/copier.go
@@ -29,6 +29,9 @@ import (
 // higher false positive rate than one computed just from the they keys in it.
 //
 // Closes input and finishes or aborts output, including on non-nil errors.
+//
+// Note that CopySpan is not aware of any suffix or prefix replacement; the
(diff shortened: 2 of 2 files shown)


RaduBerinde:
Actually, hold on, I think there's more work to do here. We don't want to use keys with the replaced suffix as index separators. I'll need to write a randomized test for CopySpan..

RaduBerinde:
Ok, I made a simpler change to just extend the bounds.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3464)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @dt, @itsbilal, and @RaduBerinde)

___
*[`compaction.go` line 2635 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3464#-Nu0_uhT74pRH9J01LHk:-Nu0_uhU110kh_8645G0:b-p34tvt) ([raw file](https://github.com/cockroachdb/pebble/blob/16a3d44db24cc00d4c3fd4342e3f63cb1167d79b/compaction.go#L2635)):*
> ```Go
> 			if n := c.comparer.Split(end.UserKey); n < len(end.UserKey) {
> 				end = base.MakeRangeDeleteSentinelKey(c.comparer.ImmediateSuccessor(nil, end.UserKey[:n]))
> 			}
> ```

Isn't it possible we'll extend into our neighboring file's bounds, violating the invariant that bounds are nonoverlapping within a level?

jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3464#-:-Nu0er5Q4y8yJkYo2Jm1:bnfp4nl)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @dt, @itsbilal, and @RaduBerinde)

___
*[`compaction.go` line 2635 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3464#-Nu0_uhT74pRH9J01LHk:-Nu0eba210ZXb6YOODOw:bvsu14i) ([raw file](https://github.com/cockroachdb/pebble/blob/16a3d44db24cc00d4c3fd4342e3f63cb1167d79b/compaction.go#L2635)):*

Isn't it possible we'll extend into our neighboring file's bounds, violating the invariant that bounds are nonoverlapping within a level?
</blockquote></details>

I misread this; we're just mutating the bounds passed into CopySpan.

RaduBerinde:
TFTR!

PR #3467, opened 2024-03-27
metamorphic: run the level checker less frequently

RaduBerinde:
The level checker can significantly slow down downloads and background
compactions in the metamorphic test.

This change makes the level checker run only once per 500ms (on
average).

Informs #3460

### DIFF
--- a/metamorphic/options.go
+++ b/metamorphic/options.go
@@ -13,6 +13,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/cockroachdb/errors"
@@ -280,7 +281,6 @@ func defaultOptions() *pebble.Options {
 		// Always use our custom comparer which provides a Split method,
 		// splitting keys at the trailing '@'.
 		Comparer:           testkeys.Comparer,
-		DebugCheck:         pebble.DebugCheckLevels,
 		FS:                 vfs.NewMem(),
 		FormatMajorVersion: defaultFormatMajorVersion,
 		Levels: []pebble.LevelOptions{{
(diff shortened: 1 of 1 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3467#-:-Nu13v2K3VJZAlsmatCz:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @itsbilal)

RaduBerinde:
Another idea is to make the checker run at most once per second in the meta test, what do you think?

jbowens:
> Another idea is to make the checker run at most once per second in the meta test, what do you think?

that sounds reasonable to me

RaduBerinde:
Done, I like that general solution more.

RaduBerinde:
Hm, I still see some Downloads taking more than 5 minutes when running under `stress -p 3` locally :/

RaduBerinde:
If I try to rerun just the one that failed, the entire test completes in 1-2s.. I wonder if there's some kind of race condition that makes it get stuck.

RaduBerinde:
Indeed, wrote up the problem here: https://github.com/cockroachdb/pebble/issues/3470

jbowens (approved):
makes sense

> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3467#-:-Nu5BVTF3I3N4p_Ysm1B:b-pqov82)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @itsbilal)

Issue #3470, opened 2024-03-28
pebble: Download can get stuck


RaduBerinde:
Here are logs from a meta test instance where the Download op got stuck:
```
        // Mar 27 21:30:19.982 INFO: [JOB 39] compacted(copy) L0 [000038] (803B) Score=0.00 + L0 [] (0B) Score=0.00 -> L0 [000050] (803B), in 0.0s (0.0s total), output rate 8.5MB/s
        // Mar 27 21:30:19.982 INFO: [JOB 41] compacting(move) L5 [000045] (97B) Score=10.80 + L6 [] (0B) Score=0.20; OverlappingRatio: Single 0.00, Multi 0.00
        // Mar 27 21:30:19.982 INFO: [JOB 41] compacted(move) L5 [000045] (97B) Score=10.80 + L6 [] (0B) Score=0.20 -> L6 [000045] (97B), in 0.0s (0.0s total), output rate 13MB/s
        // Mar 27 21:30:19.982 INFO: [JOB 42] compacting(copy) L0 [000047] (803B) Score=0.00 + L0 [] (0B) Score=0.00; OverlappingRatio: Single 0.00, Multi 0.00
        // Mar 27 21:30:19.982 INFO: [JOB 42] compacted(copy) L0 [000047] (803B) Score=0.00 + L0 [] (0B) Score=0.00 -> L0 [000051] (803B), in 0.0s (0.0s total), output rate 13MB/s
        // Mar 27 21:30:19.982 INFO: [JOB 43] compacting(copy) L0 [000040] (803B) Score=0.00 + L0 [] (0B) Score=0.00; OverlappingRatio: Single 0.00, Multi 0.00
        // Mar 27 21:30:19.982 INFO: [JOB 43] compacted(copy) L0 [000040] (803B) Score=0.00 + L0 [] (0B) Score=0.00 -> L0 [000052] (803B), in 0.0s (0.0s total), output rate 20MB/s
        // Mar 27 21:30:19.982 INFO: [JOB 46] compacting(copy) L5 [000039] (803B) Score=0.00 + L5 [] (0B) Score=0.00; OverlappingRatio: Single 0.00, Multi 0.00
        // Mar 27 21:30:19.983 INFO: [JOB 46] compacted(copy) L5 [000039] (803B) Score=0.00 + L5 [] (0B) Score=0.00 -> L5 [000053] (803B), in 0.0s (0.0s total), output rate 2.0MB/s
[...]

PR #3473, opened 2024-03-28
metamorphic: disable download op

RaduBerinde:
Disable download op until #3470 is fixed.


aadityasondhi (approved):
maybe we should add a TODO/issue to remind us to turn it back on?

itsbilal:
One possibility when we do fix the issue is to enable this operation only in the TwoInstance test. That test's additional coverage on top of `TestMeta` is just things that aren't GA'd anyway, like fast rebalances and so on, so it makes sense to let `downloadOp` bake there until we are confident in promoting it to the main meta test.

RaduBerinde:
I will reenable as part of the fix. TFTRs!

Issue #3334, opened 2024-02-22
db: simplify upper synthetic boundary in level iterator


RaduBerinde:
This issue tracks the work discussed in #3262, specifically this discussion: https://reviewable.io/reviews/cockroachdb/pebble/3262#-Npb_6vX-dru6Z2Z0Wdz

That PR simplified emitting the boundary at the lower limit. If we can make a similar change at the upper limit, we can remove `MaybeFilteredKeys` and related code.

RaduBerinde:
@jbowens I tried a change similar to what I did on the lower bound and I made the level checker ignore boundary keys. I get this failure:


```
Init(0 /* dbs */, 51 /* batches */, 70 /* iters */, 51 /* snapshots */, 0 /* externalObjs */)
db1.RangeKeyUnset("k", "s", "@11")
db1.RangeKeySet("d", "j", "@12", "ozqtetiistfhzckitwu")
db1.Set("g", "fmcic")
db1.RangeKeyUnset("m", "q", "@9")
db1.Set("p", "ntq")
db1.Compact("c", "f", true /* parallelize */)
db1.DeleteRange("i", "s")
snap23 = db1.NewSnapshot("a", "b", "e", "h", "l", "n", "o", "r")
db1.Set("p", "jqo")
db1.Flush()
snap23.Get("p")   <-- returns "ntq" instead of "not found"
```

[...]

PR #3475, opened 2024-03-29
db: unconditionally interleave ignorable boundaries

jbowens:
**db: unconditionally interleave ignorable boundaries**

The levelIter has a concept of ignorable boundary keys. When a levelIter
exhausts a file's point keys, it's possible that the same file still contains
range deletions relevant to other levels of the LSM. The levelIter will, under
some conditions, interleave the largest boundary key of the sstable into
iteration as an 'ignorable boundary key,' so that the file's range deletions
remain accessible until all other levels progress beyound the file's boundary.

When block-property filters are in use, a file's point key iterator may become
exhausted early, before the file's range deletions are irrelevant, even if the
file's largest key is not a range deletion. To work around this subtlety, the
sstable iterator previously would surface knowledge of whether any point keys
may have been skipped through a MaybeFilteredKeys method. The levelIter used
this method to determine when to interleave an ignorable largest boundary in
case there may be additional relevant range deletions.

This commit removes the conditioning on the output of MaybeFilteredKeys,
instead unconditionally interleaving a synthetic boundary at a file's largest
point key if the levelIter user is using the file's range deletion iterator.
This simplifies the logic and removes a fragile reliance on the sstable's
iterators accounting of when it may have filtered keys.

Future work (https://github.com/cockroachdb/pebble/issues/2863) will remove the need to interleave synthetic boundaries at
all, instead interleaving the range deletion bounds themselves among point
keys.

Informs https://github.com/cockroachdb/pebble/issues/2863.
Close https://github.com/cockroachdb/pebble/issues/3334.

**sstable: remove Iterator.MaybeFilteredKeys**

Remove the MaybeFilteredKeys method that surfaced knowledge of whether an
sstable iterator may have omitted keys due to block-property filters. With the
previous commit's refactor to unconditionally interleave sstable largest
boundaries when using a levelIter to drive range deletion iteration, we no
longer need to know whether keys might've been filtered. This reduces the
code complexity considerably.

**sstable: strengthen assertion around mutual exclusion of seek optimizations**

Previously an assertion that the TrySeekUsingNext and monotonic bounds
optimizations could not simultaneously be active was disabled due to the
TestIterRandomizedMaybeFilteredKeys test that could violate the invariant. The
test has been removed (along with the method it was testing), so we can
now uncondtionally restore the invariant.

### DIFF
--- a/external_iterator_test.go
+++ b/external_iterator_test.go
@@ -7,21 +7,16 @@ package pebble
 import (
 	"bytes"
 	"fmt"
-	"math"
 	"testing"
-	"time"
 
 	"github.com/cockroachdb/datadriven"
 	"github.com/cockroachdb/errors"
-	"github.com/cockroachdb/pebble/internal/base"
-	"github.com/cockroachdb/pebble/internal/cache"
 	"github.com/cockroachdb/pebble/internal/itertest"
 	"github.com/cockroachdb/pebble/internal/testkeys"
 	"github.com/cockroachdb/pebble/objstorage/objstorageprovider"
 	"github.com/cockroachdb/pebble/sstable"
 	"github.com/cockroachdb/pebble/vfs"
 	"github.com/stretchr/testify/require"
-	"golang.org/x/exp/rand"
 )
 
 func TestExternalIterator(t *testing.T) {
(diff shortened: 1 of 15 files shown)


RaduBerinde (approved):
Awesome!! Thanks for making this work and for the cleanup!

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3475#-:-NuA7t8kAYVIBcyKyeZ-:bpw40k2)* status: 0 of 15 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`level_iter.go` line 1037 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3475#-NuA7Hnb7rU004m6cGkQ:-NuA7Hnb7rU004m6cGkR:b8t2nth) ([raw file](https://github.com/cockroachdb/pebble/blob/4dcd10ebbe9eab3c9805d6aa9f8d042b74750526/level_iter.go#L1037)):*
> ```Go
> 			// the top of the heap and immediately skip the entry, advancing to
> 			// the next file.
> 			if l.iterFile.LargestPointKey.Kind() == InternalKeyKindRangeDelete ||
> ```

If the largest point key is a range delete, wouln't `*rangeDelIterPtr` always be non-nil?

Conversely, if the largest point key kind is not correct (the bound can be loose for vtables), is it correct if we still emit the boundary?
___
*[`merging_iter.go` line 955 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3475#-NuA7_450vsgLh8Vwqs6:-NuA7_450vsgLh8Vwqs7:b-1kzq13) ([raw file](https://github.com/cockroachdb/pebble/blob/4dcd10ebbe9eab3c9805d6aa9f8d042b74750526/merging_iter.go#L955)):*
> ```Go
> 			break
> 		}
> 		// Skip ignorable boundary keys. These are not real keys and exist to
> ```

I think this is what I was missing to make it work.

Isn't it strange that we need to do this in `findPrevEntry` but not for `findNextEntry`?

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3475#-:-NuBP3nw6hMeL8SkYdge:b-814g8k)* status: 0 of 15 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`level_iter.go` line 1037 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3475#-NuA7Hnb7rU004m6cGkQ:-NuBFq6I1QGCHqpGYWaI:b-8fhqq1) ([raw file](https://github.com/cockroachdb/pebble/blob/4dcd10ebbe9eab3c9805d6aa9f8d042b74750526/level_iter.go#L1037)):*
> If the largest point key is a range delete, wouln't `*rangeDelIterPtr` always be non-nil?

I think we should be able to reduce it to this, but there's currently warts in the getIter where it retains the span after it's nil'd the `rangeDelIterPtr`. I left a TODO.

> Conversely, if the largest point key kind is not correct (the bound can be loose for vtables), is it correct if we still emit the boundary?

Yeah, it is; we'll just pay the cost of an unnecessary extra step.
___
*[`merging_iter.go` line 955 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3475#-NuA7_450vsgLh8Vwqs6:-NuBFazV1waiAYhxBlo6:b8m7ckn) ([raw file](https://github.com/cockroachdb/pebble/blob/4dcd10ebbe9eab3c9805d6aa9f8d042b74750526/merging_iter.go#L955)):*

I think this is what I was missing to make it work.

Isn't it strange that we need to do this in `findPrevEntry` but not for `findNextEntry`?
</blockquote></details>

We actually already do it in `findNextEntry`:
https://github.com/cockroachdb/pebble/blob/1c7bcd1c03f1697d36f53b54476921309ccc9396/merging_iter.go#L806-L816

RaduBerinde (approved):
> Reviewed 5 of 6 files at r1, 9 of 10 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3475)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`merging_iter.go` line 955 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3475#-NuA7_450vsgLh8Vwqs6:-NuBgIUA7tbVUh626kgE:b-t2xa40) ([raw file](https://github.com/cockroachdb/pebble/blob/4dcd10ebbe9eab3c9805d6aa9f8d042b74750526/merging_iter.go#L955)):*

We actually already do it in `findNextEntry`:
https://github.com/cockroachdb/pebble/blob/1c7bcd1c03f1697d36f53b54476921309ccc9396/merging_iter.go#L806-L816
</blockquote></details>

Oh, I think I looked at the wrong `findNextEntry`. Great, that explains why I could do it for the lower boundary before.  Thanks!!

jbowens:
TFTR!

PR #3482, opened 2024-04-01
db: cancel compactions that overlap with flushable ingest+excise

itsbilal:
This change addresses the case where an excise that was ingested as a flushable needs to cancel any compactions that have boundary overlap with the excise, but no inputs of that compaction overlap with the excise.

Informs cockroachdb/cockroach#121263.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1722,12 +1722,24 @@ func (d *DB) flush1() (bytesFlushed uint64, err error) {
 			}
 		} else if len(ve.DeletedFiles) > 0 {
 			// c.kind == compactionKindIngestedFlushable && we have deleted files due
-			// to ingest-time splits.
+			// to ingest-time splits or excises.
 			//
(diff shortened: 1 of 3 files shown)


RaduBerinde:
*[`compaction.go` line 1732 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3482#-NuQ5PUM5E54IbYXNrrW:-NuQ5PUM5E54IbYXNrrX:bhl9jtm) ([raw file](https://github.com/cockroachdb/pebble/blob/ee1aae67c656a526eacab8d24f2e975bc15afb9c/compaction.go#L1732)):*
> ```Go
> 			ingestFlushable := c.flushing[0].flushable.(*ingestedFlushable)
> 			for c2 := range d.mu.compact.inProgress {
> 				// Check if this compaction overlaps with the excise span. Note that just
> ```

LGTM. Nice comment!

PR #3483, opened 2024-04-01
db: reimplement Download

### LINKED ISSUE #3470: pebble: Download can get stuck

RaduBerinde:
#### db: reimplement Download

This commits changes the download operation to keep track of where we
are in the download process and provide more clear completion
guarantees. The mechanics are described in the comment for
`downloadSpanTask`.

We also add events/log messages around the download operation and
re-enable download ops in the metamorphic test.

Fixes #3470.


sumeerbhola (commented):
just caught up on the existing implementation

> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3483#-:-NuVOzyTACsHbD5VccZT:b-e0dws9)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`db.go` line 1965 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuVOpWw4POgvuL9pWMr:-NuVOpWw4POgvuL9pWMs:b-gczus9) ([raw file](https://github.com/cockroachdb/pebble/blob/eae0efc2a3919f0a56861ea71471ce723f256726/db.go#L1965)):*
> ```Go
> 
> 	for {
> 		// Non-blocking select.
> ```

the logic in this for loop seems quite fragile/convoluted.

sumeerbhola (changes requested):
> Reviewed 8 of 12 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3483)* status: 8 of 12 files reviewed, 7 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`download.go` line 73 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuZrk1m-F1GKS_UHdrH:-NuZrk1m-F1GKS_UHdrI:b-5aqb73) ([raw file](https://github.com/cockroachdb/pebble/blob/6ce39d6e077ee3be9569f8d5f74ee4d762e03cb4/download.go#L73)):*
> ```Go
> 
> 	for info.RestartCount = 0; ; info.RestartCount++ {
> 		d.opts.EventListener.DownloadBegin(info)
> ```

So each download will always "restart" with RestartCount=1, and if there isn't any new work it will declare done immediately?

If we need policy driven downloads e.g. ensure all files for table A are local, then we really ought to "install" that policy in Pebble such that after the one time download completes, new ingested files have the same policy applied. Is that the plan for the next iteration of download?
___
*[`download.go` line 146 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuZtMbX9Sj0c7yBRFbI:-NuZtMbX9Sj0c7yBRFbJ:bvk3wux) ([raw file](https://github.com/cockroachdb/pebble/blob/6ce39d6e077ee3be9569f8d5f74ee4d762e03cb4/download.go#L146)):*
> ```Go
> 			case compactionErr := <-ch:
> 				// We ignore canceled compaction errors; the higher level code will
> 				// retry the task.
> ```

so its not just the race that motivated the retry logic. This is simple and effective.
___
*[`download.go` line 154 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuVQgaE70FTBKgxaBTV:-NuVQgaE70FTBKgxaBTW:bofkgng) ([raw file](https://github.com/cockroachdb/pebble/blob/6ce39d6e077ee3be9569f8d5f74ee4d762e03cb4/download.go#L154)):*
> ```Go
> 		if err != nil {
> 			d.removeDownloadTasks(tasks)
> 			// TODO: log this error.
> ```

todo for future PR?
___
*[`download.go` line 181 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuZvLTkDy1V9Wsy_pBO:-NuZvLTkDy1V9Wsy_pBP:bnghgbs) ([raw file](https://github.com/cockroachdb/pebble/blob/6ce39d6e077ee3be9569f8d5f74ee4d762e03cb4/download.go#L181)):*
> ```Go
> // guaranteed that all external files that were overlapping the download span at
> // the beginning of the task are either downloaded or have download compactions
> // running (which will report their completion status via compactionDoneChans).
> ```

Worth clarifying that:
we will consider the task "complete" once all these channels provide their completion state. It doesn't matter if some of these were cancelled, since retry happens at a higher layer by creating a new downloadSpanTask.

(when I was initially reading this I got tripped up by this two step completion, since it wasn't reversible)
___
*[`download.go` line 196 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuZwE7uEX7OVwjm0XKY:-NuZwE7uEX7OVwjm0XKZ:boqvhyh) ([raw file](https://github.com/cockroachdb/pebble/blob/6ce39d6e077ee3be9569f8d5f74ee4d762e03cb4/download.go#L196)):*
> ```Go
> // While a download task is running, tables with external backings can disappear
> // due to excises or compactions; tables can move *down* (to a lower LSM level);
> // or tables can have their bounds shrink due to excises. The top-down,
> ```

consider adding the following, since it clarifies that tracking the filenum is insufficient:
... bounds shrink due to excises (and these will appear as new tables, even though they have the same backing)
___
*[`event.go` line 256 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuVQwJz8d6CJsIBBOPw:-NuVQwJz8d6CJsIBBOPx:b-72nevk) ([raw file](https://github.com/cockroachdb/pebble/blob/6ce39d6e077ee3be9569f8d5f74ee4d762e03cb4/event.go#L256)):*
> ```Go
> 	DownloadCompactionsLaunched int
> 
> 	RestartCount int
> ```

could you add a comment stating what a "restart" means.

RaduBerinde (commented):
Updated the implementation to use "bookmarks". This is conceptually simpler to think about than the multiple cursors and the code is about the same in terms of complexity.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3483#-:-Nu_k88t2Yfgk7mQXYqo:bnbg71u)* status: 0 of 12 files reviewed, 7 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`download.go` line 73 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuZrk1m-F1GKS_UHdrH:-Nu_2BYSBGzBi8csgtYs:bvh74q1) ([raw file](https://github.com/cockroachdb/pebble/blob/6ce39d6e077ee3be9569f8d5f74ee4d762e03cb4/download.go#L73)):*

So each download will always "restart" with RestartCount=1, and if there isn't any new work it will declare done immediately?

If we need policy driven downloads e.g. ensure all files for table A are local, then we really ought to "install" that policy in Pebble such that after the one time download completes, new ingested files have the same policy applied. Is that the plan for the next iteration of download?
</blockquote></details>

Fixed to not issue that restart.

Yes, I think we will need something along those lines.
___
*[`download.go` line 146 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuZtMbX9Sj0c7yBRFbI:-Nu_kLco2SL-IY_aqbNQ:bfzuuc3) ([raw file](https://github.com/cockroachdb/pebble/blob/6ce39d6e077ee3be9569f8d5f74ee4d762e03cb4/download.go#L146)):*

so its not just the race that motivated the retry logic. This is simple and effective.
</blockquote></details>

In the new version, we retry as part of the task.
___
*[`download.go` line 154 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuVQgaE70FTBKgxaBTV:-Nu_2SJ98pIOfHvYdjxK:b-25l68e) ([raw file](https://github.com/cockroachdb/pebble/blob/6ce39d6e077ee3be9569f8d5f74ee4d762e03cb4/download.go#L154)):*

todo for future PR?
</blockquote></details>

Removed, this was actually done (through `DownloadEnd`).
___
*[`download.go` line 196 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuZwE7uEX7OVwjm0XKY:-Nu_2mhi-n6JJsyOkMmG:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/6ce39d6e077ee3be9569f8d5f74ee4d762e03cb4/download.go#L196)):*

consider adding the following, since it clarifies that tracking the filenum is insufficient:
... bounds shrink due to excises (and these will appear as new tables, even though they have the same backing)
</blockquote></details>

Done.
___
*[`event.go` line 256 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuVQwJz8d6CJsIBBOPw:-Nu_3A9XDmQBRKOQqHSG:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/6ce39d6e077ee3be9569f8d5f74ee4d762e03cb4/event.go#L256)):*

could you add a comment stating what a "restart" means.
</blockquote></details>

Done.

sumeerbhola (changes requested):
looks good! minor comments only.
I will read the test tomorrow.

> Reviewed 9 of 12 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3483#-:-Nub7--y-26Pdwdcwc-R:b-m274tj)* status: 9 of 12 files reviewed, 14 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`compaction.go` line 1920 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-Nub3zRhAKQX_NcBPUqj:-Nub3zRhAKQX_NcBPUqk:bimhfc8) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/compaction.go#L1920)):*
> ```Go
> 		case didNotLaunchCompaction:
> 			// See if we can launch a compaction for another download task.
> 			i++
> ```

it's not clear to me what `maxConcurrentDownloads` limits given that we move here to the next index. The previous one could already be running `maxConcurrentDownloads`.

Ah, I see it now: `tryScheduleDownloadCompaction` will be called if the `DB.mu.compact.downloadingCount` (representing real download compactions) is less than `maxConcurrentDownloads`, say as high as `maxConcurrentDownloads-1`. Then if the first `downloadSpanTask` did not launch any new compactions, and we switched to the next `downloadSpanTask`, it would launch at most one such compaction and return, and then the outer loop would check again.

I don't understand why we are limiting a `downloadSpanTask` to tracking only `maxConcurrentDownloads` bookmarks. If the caller here is asking for more download compactions, since it has looked at `DB.mu.compact.downloadingCount`, the task should give it what it wants. Otherwise we will wastefully ask the next task while the previous task still had the ability to start new compactions.
___
*[`download.go` line 161 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuavcnY3dC8B-6mRqEo:-NuavcnY3dC8B-6mRqEp:b3sa7pf) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/download.go#L161)):*
> ```Go
> // guaranteed that all external files that were overlapping the download span at
> // the beginning of the task are either downloaded or have download compactions
> // running (which will report their completion status via compactionDoneChans).
> ```

Looks like compactionDoneChans are not exposed anymore. Stale comment?
___
*[`download.go` line 210 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuavZIR-H1jwxxfV0cS:-NuavZIR-H1jwxxfV0cT:bst0ec1) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/download.go#L210)):*
> ```Go
> 
> 	// taskCompletedChan is signaled when we have finished download compactions
> 	// for all external files encountered within the bounds, or when we one of
> ```

nit: or when one ...
___
*[`download.go` line 343 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuaxwDLCZpb2T4lgTAJ:-NuaxwDLCZpb2T4lgTAK:bw1ce63) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/download.go#L343)):*
> ```Go
> // NextExternalFile returns the first file after the cursor, returning the file
> // and a cursor that is immediately before the file. If no such file exists,
> // returns a nil metadata and a cursor that is AtEnd.
> ```

nit: the second return value is not actually a cursor.
___
*[`download.go` line 469 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-Nub0mui66wgCbp0d6Yz:-Nub0mui66wgCbp0d6Z-:bga261v) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/download.go#L469)):*
> ```Go
> 				// files. It is possible that this download was started after the
> 				// bookmark was initially created and it might correspond to a smaller
> 				// file.
> ```

I didn't quite understand the second sentence in this comment. Can you elaborate with a concrete example?
I understand we want to check the bookmark since there may be nothing there, or something to do. Oh wait, maybe this comment was trying to say something about the compactionErr == nil case, in that we need to check whether there is something left to do? But it's placed in the code where we can fall through both for compactionErr == nil and for a non-nil err (and a cancelled compaction), hence a bit confusing.
___
*[`download.go` line 475 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-Nub1f7c5rVAC70ME0I5:-Nub1f7c5rVAC70ME0I6:bkzo1bi) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/download.go#L475)):*
> ```Go
> 				continue
> 			}
> 		}
> ```

we will also fall through to the following if it was a normal compaction and check the bookmark, since there is no completion indication, so we have to "poll", yes?
Can you add a code comment about this polling, including why this is ok (presumably because tryLaunchDownloadCompaction is not being called in a tight loop)?
___
*[`download.go` line 486 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-Nub2NRA0HR9OwOFV8mK:-Nub2NRB6nVB2AtxVQig:b-8fhfp2) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/download.go#L486)):*
> ```Go
> 		// Move up the bookmark position to point at this file.
> 		b.start = makeCursorAtFile(f, b.start.level)
> 		doneCh, ok := d.tryLaunchDownloadForFile(vers, env, download, b.start.level, f)
> ```

so if it is already compacting, ok will be false, and we will leave the bookmark in place for later polling?
Code comment would help, since the handling of non-download compacting files is a bit subtle.

RaduBerinde (commented):
Thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3483#-:-NubDUOu-YIHfoqsrFjW:b-ucfknh)* status: 9 of 12 files reviewed, 12 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`compaction.go` line 1920 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-Nub3zRhAKQX_NcBPUqj:-NubAA7j6bHJs53df_gL:b-djx31x) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/compaction.go#L1920)):*

it's not clear to me what `maxConcurrentDownloads` limits given that we move here to the next index. The previous one could already be running `maxConcurrentDownloads`.

Ah, I see it now: `tryScheduleDownloadCompaction` will be called if the `DB.mu.compact.downloadingCount` (representing real download compactions) is less than `maxConcurrentDownloads`, say as high as `maxConcurrentDownloads-1`. Then if the first `downloadSpanTask` did not launch any new compactions, and we switched to the next `downloadSpanTask`, it would launch at most one such compaction and return, and then the outer loop would check again.

I don't understand why we are limiting a `downloadSpanTask` to tracking only `maxConcurrentDownloads` bookmarks. If the caller here is asking for more download compactions, since it has looked at `DB.mu.compact.downloadingCount`, the task should give it what it wants. Otherwise we will wastefully ask the next task while the previous task still had the ability to start new compactions.
</blockquote></details>

It's just a limit for practical purposes.. We look at each bookmark every time a compaction ends so I hesitate allowing an arbitrary number (e.g. maybe we have a huge compaction happening). I figured it's a reasonable limit in practice, even if some of the compactions that are happening don't count towards the `downloadingCount`.

What would you suggest? I do think there should be some kind of limit here, even if it's much higher than the concurrency limit.
___
*[`download.go` line 161 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NuavcnY3dC8B-6mRqEo:-NubAvPg2E7z12z0JPmb:bt2vr1b) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/download.go#L161)):*

Looks like compactionDoneChans are not exposed anymore. Stale comment?
</blockquote></details>

Fixed.
___
*[`download.go` line 469 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-Nub0mui66wgCbp0d6Yz:-NubCPz70iH2rlQgFN-m:br8uf3n) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/download.go#L469)):*

I didn't quite understand the second sentence in this comment. Can you elaborate with a concrete example?
I understand we want to check the bookmark since there may be nothing there, or something to do. Oh wait, maybe this comment was trying to say something about the compactionErr == nil case, in that we need to check whether there is something left to do? But it's placed in the code where we can fall through both for compactionErr == nil and for a non-nil err (and a cancelled compaction), hence a bit confusing.
</blockquote></details>

Good point, I hopefully made it more clear.
___
*[`download.go` line 475 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-Nub1f7c5rVAC70ME0I5:-NubD7Is0jbP8PKXZQEC:bs7wumx) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/download.go#L475)):*

we will also fall through to the following if it was a normal compaction and check the bookmark, since there is no completion indication, so we have to "poll", yes?
Can you add a code comment about this polling, including why this is ok (presumably because tryLaunchDownloadCompaction is not being called in a tight loop)?
</blockquote></details>

Done. 
___
*[`download.go` line 486 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-Nub2NRA0HR9OwOFV8mK:-NubDTFT2Do5iJlUwU-v:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/download.go#L486)):*

so if it is already compacting, ok will be false, and we will leave the bookmark in place for later polling?
Code comment would help, since the handling of non-download compacting files is a bit subtle.
</blockquote></details>

Done.

sumeerbhola (approved):
nice tests!

:lgtm: 

> Reviewed 3 of 12 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3483#-:-Nudp8px7q-XarHET8_3:b90owhi)* status: all files reviewed, 13 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`compaction.go` line 1920 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-Nub3zRhAKQX_NcBPUqj:-NudkjJ92oJqkOysqhkW:b-qplvlw) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/compaction.go#L1920)):*

It's just a limit for practical purposes.. We look at each bookmark every time a compaction ends so I hesitate allowing an arbitrary number (e.g. maybe we have a huge compaction happening). I figured it's a reasonable limit in practice, even if some of the compactions that are happening don't count towards the `downloadingCount`.

What would you suggest? I do think there should be some kind of limit here, even if it's much higher than the concurrency limit.
</blockquote></details>

I think what you have is ok. But I think it could benefit for some code commentary of why there is a limit on the number of bookmarks in a task (as you have stated above).
___
*[`download.go` line 469 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-Nub0mui66wgCbp0d6Yz:-NudlZp4DQRGHa8hh7oQ:b-zcjvte) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/download.go#L469)):*

Good point, I hopefully made it more clear.
</blockquote></details>

Thanks
___
*[`testdata/download_task` line 170 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NudozCkDbzzi168VFv-:-NudozCkDbzzi168VFv0:b-j1f23p) ([raw file](https://github.com/cockroachdb/pebble/blob/fed37e1a1b491c68b9bd61c050a046f9fd328887/testdata/download_task#L170)):*
> ```
> cursor: level=2 key="a" seqNum=1
> 
> try-launch max-concurrent-downloads=2
> ```

was this intended to be a noop in the test, or copy-paste typo?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3483#-:-NueLqtQ54OkmNrQrhle:b-814g8k)* status: 0 of 12 files reviewed, 13 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`compaction.go` line 1920 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3483#-Nub3zRhAKQX_NcBPUqj:-NueL9wA-jgrc9Nuigh3:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c6d7594ca53aa95d223d51cee672cb192ccd7b65/compaction.go#L1920)):*

I think what you have is ok. But I think it could benefit for some code commentary of why there is a limit on the number of bookmarks in a task (as you have stated above).
</blockquote></details>

Done.
___
*[`testdata/download_task` line 170 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3483#-NudozCkDbzzi168VFv-:-NueL_N86tD8_1w9W5OZ:b-rrb74b) ([raw file](https://github.com/cockroachdb/pebble/blob/fed37e1a1b491c68b9bd61c050a046f9fd328887/testdata/download_task#L170)):*

was this intended to be a noop in the test, or copy-paste typo?
</blockquote></details>

Added a comment, I think I wanted to verify that the cursor doesn't move.

sumeerbhola (approved):
:lgtm: 

> Reviewed 3 of 12 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3483#-:-NuekujX6180dywrqpSD:bg7igkv)* status: 3 of 12 files reviewed, 13 unresolved discussions (waiting on @itsbilal)

PR #3488, opened 2024-04-02
sstable: clear user properties in CopySpan

### LINKED ISSUE #3485: TestMeta: incorrect result after downloading external file
```
Init(0 /* dbs */, 41 /* batches */, 57 /* iters */, 44 /* snapshots */, 1 /* externalObjs */)
batch5 = db1.NewIndexedBatch()
batch5.Set("t", "bdgepzgefnbmoadh")
batch5.Delete("n")
batch5.Merge("r", "bswhibfochbzhriv")
external0 = batch5.NewExternalObj()
db1.IngestExternalFiles(external0, "aoaaaiumi" /* start */, "avqkpipyc" /* end */, "" /* syntheticSuffix */, "a" /* syntheticPrefix */, external0, "fppaqamjaiyf" /* start */, "vph" /* end */, "" /* syntheticSuffix */, "" /* syntheticPrefix */)
db1.Download("a" /* start */, "p" /* end */, true /* viaBackingFileDownload */, "i" /* start */, "l" /* end */, false /* viaBackingFileDownload */, "b" /* start */, "g" /* end */, true /* viaBackingFileDownload */, "d" /* start */, "q" /* end */, false /* viaBackingFileDownload */, "c" /* start */, "k" /* end */, true /* viaBackingFileDownload */, "h" /* start */, "o" /* end */, false /* viaBackingFileDownload */, "e" /* start */, "f" /* end */, false /* viaBackingFileDownload */, "j" /* start */, "m" /* end */, false /* viaBackingFileDownload */)
iter34 = db1.NewIter("", "", 0 /* key types */, 59, 85, false /* use L6 filters */, "" /* masking suffix */)
iter34.SeekLT("s", "")
```

```
-iter34.SeekLT("s", "") // [true,"r","bswhibfochbzhriv",<no range>] <nil> #9
+iter34.SeekLT("s", "") // [true,"at","bdgepzgefnbmoadh",<no range>] <nil> #9
```

Issue seems to be in the CopySpan path. If I force it to fall back to `copyWholeFileBecauseOfUnsupportedFeature`, the problem goes away.

RaduBerinde:
The user properties contain the table block properties and the
presence of any such property implies that index entries have the same
properties. When we create a partial copy of an sst in `CopySpan`, we
don't calculate properties for index entries but non-existent
properties can still have meaning (for example, it means a
non-obsolete block for the obsolete property filter; and it means an
empty interval for the interval filter).

The fix is to just clear out the user properties so that the reader
knows there are no block properties collected.

Fixes #3485

### DIFF
--- a/sstable/copier.go
+++ b/sstable/copier.go
@@ -19,14 +19,17 @@ import (
 // [start, end), as well as potentially some additional keys from the original
 // file that were adjacent to but outside that span.
 //
-// CopySpan differs from simply seeking a reader to start and iterating a until
-// end passing the results to a writer in that it does not write the new sstable
-// from scratch, key-by-key, recompressing each key nto new blocks and computing
-// new filters and properties. Instead it finds _blocks_ that intersect the
-// requested span and copies those, whole, to the new file, avoiding all
-// decompression and recompression work. It then copies the original filter as
-// these can serve as a filter for the subset as well, just with potentially a
-// higher false positive rate than one computed just from the they keys in it.
+// CopySpan differs from simply seeking a reader to start and iterating until
+// the end passing the results to a writer in that it does not write the new
+// sstable from scratch, key-by-key, recompressing each key into new blocks and
+// computing new filters and properties. Instead, it finds data _blocks_ that
+// intersect the requested span and copies those, whole, to the new file,
+// avoiding all decompression and recompression work. It then copies the
+// original bloom filter - this filter is valid for the subset of data as well,
+// just with potentially a higher false positive rate compared to one that would
+// be computed just from the keys in it.
+//
+// The resulting sstable will have no block properties.
 //
 // Closes input and finishes or aborts output, including on non-nil errors.
 //
@@ -61,11 +64,11 @@ func CopySpan(
 
 	// We don't want the writer to attempt to write out block property data in
 	// index blocks. This data won't be valid since we're not passing the actual
-	// key data through the writer. The data blocks will still have their
-	// properties unchanged, and we will use the the table level properties from
-	// the original ssts as well.
-	// TODO(dt,radu): Figure out how to initialize the prop collector state from
-	// the decoded block props such that they can re-encode index block props.
+	// key data through the writer. We also remove the table-level properties
+	// below.
+	//
+	// TODO(dt,radu): Figure out how to populate the prop collector state with
+	// block props from the original sst.
 	w.blockPropCollectors = nil
 
 	defer func() {
@@ -101,6 +104,9 @@ func CopySpan(
 	// original props instead. This will result in over-counts but that is safer
 	// than under-counts.
 	w.props = r.Properties
+	// Remove all user properties to disable block properties, which we do not
+	// calculate.
+	w.props.UserProperties = nil
 	// Reset props that we'll re-derive as we build our own index.
 	w.props.IndexPartitions = 0
 	w.props.TopLevelIndexSize = 0
--- a/sstable/properties.go
+++ b/sstable/properties.go
@@ -180,7 +180,8 @@ type Properties struct {
 	SnapshotPinnedValueSize uint64 `prop:"pebble.raw.snapshot-pinned-values.size"`
 	// Size of the top-level index if kTwoLevelIndexSearch is used.
 	TopLevelIndexSize uint64 `prop:"rocksdb.top-level.index.size"`
-	// User collected properties.
+	// User collected properties. Currently, we only use them to store block
+	// properties aggregated at the table level.
 	UserProperties map[string]string
 
 	// Loaded set indicating which fields have been loaded from disk. Indexed by
--- a/sstable/writer.go
+++ b/sstable/writer.go
@@ -2028,7 +2028,7 @@ func (w *Writer) Close() (err error) {
 	{
 		// Finish and record the prop collectors if props are not yet recorded.
 		// Pre-computed props might have been copied by specialized sst creators
-		// like suffix replacer or a span copier.
+		// like suffix replacer.
 		if len(w.props.UserProperties) == 0 {
 			userProps := make(map[string]string)
 			for i := range w.blockPropCollectors {
--- a/testdata/ingest_external
+++ b/testdata/ingest_external
@@ -341,8 +341,8 @@ ok
 lsm verbose
 ----
 L6:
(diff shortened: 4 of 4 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3488#-:-NuVCWDV8wNaDyuul6kW:bg7igkv)* status: all files reviewed, 5 unresolved discussions (waiting on @dt, @msbutler, and @RaduBerinde)

___
*[`sstable/copier.go` line 24 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3488#-NuV9IlE0C1pufmGXeMr:-NuV9IlE0C1pufmGXeMs:bb7i8yt) ([raw file](https://github.com/cockroachdb/pebble/blob/f23f49cd2fea66d18fab2214c7065d1efa1f9954/sstable/copier.go#L24)):*
> ```Go
> // CopySpan differs from simply seeking a reader to start and iterating a until
> // end passing the results to a writer in that it does not write the new sstable
> // from scratch, key-by-key, recompressing each key nto new blocks and computing
> ```

into
___
*[`sstable/copier.go` line 27 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3488#-NuV9Bvj00NW_3ss_HXY:-NuV9Bvk7liZz7CJpDNw:b-6dme3r) ([raw file](https://github.com/cockroachdb/pebble/blob/f23f49cd2fea66d18fab2214c7065d1efa1f9954/sstable/copier.go#L27)):*
> ```Go
> // new filters and properties. Instead it finds _blocks_ that intersect the
> // requested span and copies those, whole, to the new file, avoiding all
> // decompression and recompression work. It then copies the original filter as
> ```

this is referring to the bloom filter, and not block property filters, yes?
___
*[`sstable/copier.go` line 29 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3488#-NuV9OTy8bWzszhhZbh7:-NuV9OTy8bWzszhhZbh8:b-7y6ck5) ([raw file](https://github.com/cockroachdb/pebble/blob/f23f49cd2fea66d18fab2214c7065d1efa1f9954/sstable/copier.go#L29)):*
> ```Go
> // decompression and recompression work. It then copies the original filter as
> // these can serve as a filter for the subset as well, just with potentially a
> // higher false positive rate than one computed just from the they keys in it.
> ```

s/they/the
___
*[`sstable/copier.go` line 66 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3488#-NuV9eUPElnIXQEfFGVf:-NuV9eUPElnIXQEfFGVg:b-6hoosm) ([raw file](https://github.com/cockroachdb/pebble/blob/f23f49cd2fea66d18fab2214c7065d1efa1f9954/sstable/copier.go#L66)):*
> ```Go
> 	// key data through the writer. The data blocks will still have their
> 	// properties unchanged, and we will use the the table level properties from
> 	// the original ssts as well.
> ```

this comment seems stale.
___
*[`sstable/copier.go` line 106 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3488#-NuVBldo2N5sp1YLMxUa:-NuVBldo2N5sp1YLMxUb:b-sr9ymv) ([raw file](https://github.com/cockroachdb/pebble/blob/f23f49cd2fea66d18fab2214c7065d1efa1f9954/sstable/copier.go#L106)):*
> ```Go
> 	// Remove all user properties to disable block properties, which we do not
> 	// calculate.
> 	w.props.UserProperties = nil
> ```

Can you add a comment where UserProperties is declared that they are only used for "block properties" that are aggregated to the table level.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3488#-:-NuVHRijAV8hShNE-f3u:b-814g8k)* status: 2 of 4 files reviewed, 5 unresolved discussions (waiting on @dt, @msbutler, and @sumeerbhola)

___
*[`sstable/copier.go` line 24 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3488#-NuV9IlE0C1pufmGXeMr:-NuVDpD41OMPTfsJtuRq:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/f23f49cd2fea66d18fab2214c7065d1efa1f9954/sstable/copier.go#L24)):*

into
</blockquote></details>

Done.
___
*[`sstable/copier.go` line 27 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3488#-NuV9Bvj00NW_3ss_HXY:-NuVDpPW81gGhTdMCimV:bcppo66) ([raw file](https://github.com/cockroachdb/pebble/blob/f23f49cd2fea66d18fab2214c7065d1efa1f9954/sstable/copier.go#L27)):*

this is referring to the bloom filter, and not block property filters, yes?
</blockquote></details>

Right, updated.
___
*[`sstable/copier.go` line 29 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3488#-NuV9OTy8bWzszhhZbh7:-NuVDq3WAgLqZ7WWBCOj:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/f23f49cd2fea66d18fab2214c7065d1efa1f9954/sstable/copier.go#L29)):*

s/they/the
</blockquote></details>

Done.
___
*[`sstable/copier.go` line 66 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3488#-NuV9eUPElnIXQEfFGVf:-NuVEW1S05j871Sb3Y-V:b-adenuq) ([raw file](https://github.com/cockroachdb/pebble/blob/f23f49cd2fea66d18fab2214c7065d1efa1f9954/sstable/copier.go#L66)):*

this comment seems stale.
</blockquote></details>

Updated. Part of it made no sense I think - we don't retain any data block properties since those are stored in the index block (along with the handle).
___
*[`sstable/copier.go` line 106 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3488#-NuVBldo2N5sp1YLMxUa:-NuVEK2R8aTSlEtTte1x:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/f23f49cd2fea66d18fab2214c7065d1efa1f9954/sstable/copier.go#L106)):*

Can you add a comment where UserProperties is declared that they are only used for "block properties" that are aggregated to the table level.
</blockquote></details>

Done.

PR #3496, opened 2024-04-08
sstable: clean up UpdateKeySuffixes

RaduBerinde:
This change cleans up the design of `SuffixReplaceableBlockCollector`: instead of having a separate optional interface, we fold this functionality in the `BlockPropertyCollector` and
`DataBlockIntervalCollector` interfaces. This makes things much easier to follow.

Note that the old design was especially problematic for `DataBlockIntervalCollector`: we were passing the encoded properties to the implementation but the encoding of intervals into properties is implemented by Pebble. This works only because the Cockroach implementation (`pebbleDataBlockMVCCTimeIntervalCollector`) does not need to look at the old properties at all.

### DIFF
--- a/iterator_test.go
+++ b/iterator_test.go
@@ -485,6 +485,16 @@ func (c *minSeqNumPropertyCollector) FinishTable(buf []byte) ([]byte, error) {
 	return binary.AppendUvarint(buf, c.minSeqNum), nil
 }
 
+func (c *minSeqNumPropertyCollector) AddCollectedWithSuffixReplacement(
+	oldProp []byte, oldSuffix, newSuffix []byte,
+) error {
+	return errors.Errorf("not implemented")
+}
+
+func (c *minSeqNumPropertyCollector) SupportsSuffixReplacement() bool {
+	return false
+}
+
 // minSeqNumFilter is a BlockPropertyFilter that uses the
 // minSeqNumPropertyCollector data to filter out entire tables.
 type minSeqNumFilter struct {
@@ -1121,6 +1131,8 @@ type testBlockIntervalCollector struct {
 	lower, upper  uint64
 }
 
+var _ sstable.DataBlockIntervalCollector = (*testBlockIntervalCollector)(nil)
+
 func (bi *testBlockIntervalCollector) Add(key InternalKey, value []byte) error {
 	k := key.UserKey
 	if len(k) < bi.numLength+bi.offsetFromEnd {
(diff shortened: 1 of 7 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3496#-:-Nuyikk87F5_2fLxJTvZ:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/block_property.go` line 106 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3496#-Nuyfvq14yeg7gJ_7hk4:-Nuyfvq14yeg7gJ_7hk5:b61nzuu) ([raw file](https://github.com/cockroachdb/pebble/blob/97318c5cf84e8494c1677659302ffd26b60e5310/sstable/block_property.go#L106)):*
> ```Go
> 	Add(key InternalKey, value []byte) error
> 
> 	// AddCollectedWithSuffixReplacement adds previously collected property data
> ```

"adds previously" can be a bit confusing since what is happening here is replacement. In that this is (a) expected to be called instead of Add, (b) even if someone did call Add, before calling AddCollectedWithSuffixReplacement, it replaces the state accumulated via Add for the data block. In a different sense, this is an addition, in that it contributes to the accumulated state for the sstable.
A bit longer code comment may help clarify this.
___
*[`sstable/block_property.go` line 245 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3496#-NuygoB8D1kYRukI9xXj:-NuygoB8D1kYRukI9xXk:b-oi0309) ([raw file](https://github.com/cockroachdb/pebble/blob/97318c5cf84e8494c1677659302ffd26b60e5310/sstable/block_property.go#L245)):*
> ```Go
> 	Add(key InternalKey, value []byte) error
> 
> 	// AddCollectedWithSuffixReplacement extends the interval with a given
> ```

replaces the interval?
Given the `DataBlockIntervalCollector` has a memory only for a data block, we could perhaps call this `SetCollectedWithSuffixReplacement`.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3496)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/block_property.go` line 106 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3496#-Nuyfvq14yeg7gJ_7hk4:-NuylZrFDtKsCIn11izn:blx1glj) ([raw file](https://github.com/cockroachdb/pebble/blob/97318c5cf84e8494c1677659302ffd26b60e5310/sstable/block_property.go#L106)):*

"adds previously" can be a bit confusing since what is happening here is replacement. In that this is (a) expected to be called instead of Add, (b) even if someone did call Add, before calling AddCollectedWithSuffixReplacement, it replaces the state accumulated via Add for the data block. In a different sense, this is an addition, in that it contributes to the accumulated state for the sstable.
A bit longer code comment may help clarify this.
</blockquote></details>

It is not designed to be replacement. It should be possible to call `Add` a few times then `AddCollectedWithSuffixReplacement` and they will "merge" the results. In practice, the only implementation we have that does this is the `pebbleDataBlockMVCCTimeIntervalCollector` where it does extend the existing interval.

I will separately propose a change that that removes knowledge of data blocks and index blocks from the collector interfaces, and adds an `AddCollected` method. The index blocks will use a separate collector instance that just gets `AddCollected` calls that accumulate to get the index block or table properties.

PR #3497, opened 2024-04-09
sstable: move empty block property logic to decoder

RaduBerinde:
We omit block properties that are empty when we encode. When we
decode, we have to handle any missing properties as empty. The code
that does this is hard to follow. This change moves this logic into
the decoder, which makes things much cleaner.

### DIFF
--- a/sstable/block_property.go
+++ b/sstable/block_property.go
@@ -12,6 +12,7 @@ import (
 	"unsafe"
 
 	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/internal/rangekey"
 )
 
@@ -524,11 +525,28 @@ func (b *BlockIntervalFilter) SetInterval(lower, upper uint64) {
 	b.filterInterval = interval{lower: lower, upper: upper}
 }
 
-// When encoding block properties for each block, we cannot afford to encode
-// the name. Instead, the name is mapped to a shortID, in the scope of that
-// sstable, and the shortID is encoded. Since we use a uint8, there is a limit
-// of 256 block property collectors per sstable.
-type shortID uint8
+// When encoding block properties for each block, we cannot afford to encode the
+// name. Instead, the name is mapped to a shortID, in the scope of that sstable,
+// and the shortID is encoded as a single byte (which imposes a limit of of 256
+// block property collectors per sstable).
+// Note that the in-memory type is int16 to avoid overflows (e.g. in loops) and
+// to allow special values like -1 in code.
+type shortID int16
+
+const invalidShortID shortID = -1
+const maxShortID shortID = math.MaxUint8
+const maxPropertyCollectors = int(maxShortID) + 1
+
+func (id shortID) IsValid() bool {
+	return id >= 0 && id <= maxShortID
+}
+
+func (id shortID) ToByte() byte {
+	if invariants.Enabled && !id.IsValid() {
+		panic(fmt.Sprintf("inavlid id %d", id))
+	}
+	return byte(id)
+}
 
 type blockPropertiesEncoder struct {
 	propsBuf []byte
@@ -544,6 +562,11 @@ func (e *blockPropertiesEncoder) resetProps() {
 }
 
 func (e *blockPropertiesEncoder) addProp(id shortID, scratch []byte) {
+	if len(scratch) == 0 {
+		// We omit empty properties. The decoder will know that any missing IDs had
+		// empty values.
+		return
+	}
 	const lenID = 1
 	lenProp := uvarintLen(uint32(len(scratch)))
 	n := lenID + lenProp + len(scratch)
@@ -558,7 +581,7 @@ func (e *blockPropertiesEncoder) addProp(id shortID, scratch []byte) {
 	}
 	pos := len(e.propsBuf)
 	b := e.propsBuf[pos : pos+lenID]
-	b[0] = byte(id)
+	b[0] = id.ToByte()
 	pos += lenID
 	b = e.propsBuf[pos : pos+lenProp]
 	n = binary.PutUvarint(b, uint64(len(scratch)))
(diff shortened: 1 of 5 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3497#-:-Nv346_v0TpsxyqMM_zK:bg7igkv)* status: all files reviewed, 6 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3497#-Nv2ssG50OykcWfAlxa-):*
nit: into the encoder/decoder, ...
___
*[`sstable/block_property.go` line 564 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3497#-Nv2quXg6WtdFL61n6eX:-Nv2quXg6WtdFL61n6eY:bvxiuie) ([raw file](https://github.com/cockroachdb/pebble/blob/3661c4117377227d888b1b0ca5862c07950c88f9/sstable/block_property.go#L564)):*
> ```Go
> type blockPropertiesDecoder struct {
> 	props []byte
> 	// numCollectedProps is the number of collectors that were used when writing
> ```

there in no `numCollectedProps` field.
But it is a param below.
I think `endID` is a better name for the param too since it is the exclusive upper bound of the ids. It doesn't require one to think about whether ids are densely numbered or not.

Oh, I see below that the dense numbering is used in `Next`. Then it is worth saying that as a code comment.
___
*[`sstable/block_property.go` line 589 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3497#-Nv2rpU79oU6wUKp-0-m:-Nv2rpU79oU6wUKp-0-n:bkczl4j) ([raw file](https://github.com/cockroachdb/pebble/blob/3661c4117377227d888b1b0ca5862c07950c88f9/sstable/block_property.go#L589)):*
> ```Go
> 	d.nextID++
> 
> 	if len(d.props) == 0 || shortID(d.props[0]) != id {
> ```

can you add an `invariants.Enabled` assertion that `shortID(d.props[0]) >= id`.
___
*[`sstable/suffix_rewriter.go` line 335 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3497#-Nv32RNQ64fUg5QEZQSO:-Nv32RNQ64fUg5QEZQSP:bjet5dt) ([raw file](https://github.com/cockroachdb/pebble/blob/3661c4117377227d888b1b0ca5862c07950c88f9/sstable/suffix_rewriter.go#L335)):*
> ```Go
> 	}
> 
> 	var oldShortIDs []shortID
> ```

// oldShortIDs maps the shortID for the block property collector in the old blocks to the shortID in the new blocks. Initialized once for the sstable.
___
*[`sstable/suffix_rewriter.go` line 336 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3497#-Nv33FK56ol1Xt-ytrvj:-Nv33FK56ol1Xt-ytrvk:b-kf8rdz) ([raw file](https://github.com/cockroachdb/pebble/blob/3661c4117377227d888b1b0ca5862c07950c88f9/sstable/suffix_rewriter.go#L336)):*
> ```Go
> 
> 	var oldShortIDs []shortID
> 	var oldProps [][]byte
> ```

// oldProps is the property value in an old block, indexed by the shortID in the new block. Slice is reused for each block.
___
*[`sstable/suffix_rewriter.go` line 348 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3497#-Nv32fXa5gUdTkbZN7Y2:-Nv32fXa5gUdTkbZN7Y3:b-cm4q6d) ([raw file](https://github.com/cockroachdb/pebble/blob/3661c4117377227d888b1b0ca5862c07950c88f9/sstable/suffix_rewriter.go#L348)):*
> ```Go
> 				oldShortIDs[was] = is
> 			} else {
> 				return errors.Errorf("sstable does not contain property %s", p.Name())
> ```

glad to see this error being handled, instead of incorrectly passing nil to UpdateKeySuffixes.

RaduBerinde (commented):
TFTR! I made a couple more related cleanups.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3497#-:-Nv3GpwaCF_37eOexDKL:bknta8w)* status: 2 of 5 files reviewed, 5 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`sstable/block_property.go` line 564 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3497#-Nv2quXg6WtdFL61n6eX:-Nv3CHc25x7yEx-43uCF:bkswjt9) ([raw file](https://github.com/cockroachdb/pebble/blob/3661c4117377227d888b1b0ca5862c07950c88f9/sstable/block_property.go#L564)):*

there in no `numCollectedProps` field.
But it is a param below.
I think `endID` is a better name for the param too since it is the exclusive upper bound of the ids. It doesn't require one to think about whether ids are densely numbered or not.

Oh, I see below that the dense numbering is used in `Next`. Then it is worth saying that as a code comment.
</blockquote></details>

renamed the field back to numCollectedProps
___
*[`sstable/block_property.go` line 589 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3497#-Nv2rpU79oU6wUKp-0-m:-Nv3CQpgCjg1yqDAu4_s:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/3661c4117377227d888b1b0ca5862c07950c88f9/sstable/block_property.go#L589)):*

can you add an `invariants.Enabled` assertion that `shortID(d.props[0]) >= id`.
</blockquote></details>

Done.
___
*[`sstable/suffix_rewriter.go` line 335 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3497#-Nv32RNQ64fUg5QEZQSO:-Nv3CXhoEO0G0YHNZnhi:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/3661c4117377227d888b1b0ca5862c07950c88f9/sstable/suffix_rewriter.go#L335)):*

// oldShortIDs maps the shortID for the block property collector in the old blocks to the shortID in the new blocks. Initialized once for the sstable.
</blockquote></details>

Done.
___
*[`sstable/suffix_rewriter.go` line 336 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3497#-Nv33FK56ol1Xt-ytrvj:-Nv3CekZ3VwNO_1-m-77:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/3661c4117377227d888b1b0ca5862c07950c88f9/sstable/suffix_rewriter.go#L336)):*

// oldProps is the property value in an old block, indexed by the shortID in the new block. Slice is reused for each block.
</blockquote></details>

Done.

sumeerbhola (approved):
:lgtm_strong: 

> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3497#-:-NvCPdjC2RSUbpuIZPXs:b-uwkowd)* status: all files reviewed, 5 unresolved discussions (waiting on @itsbilal)

PR #3498, opened 2024-04-09
*: return *InternalKV from internal iterators

jbowens:
Previously, internal iterators' positioning methods returned two return values, a pointer to an internal key and a LazyValue struct. This commit collapses these two return values into a single return value: a pointer to an InternalKV struct. The key and value buffers were already owned by the underlying iterator, so passing the base.LazyValue as a value did not decouple the two. Propagating a single *InternalKV pointer can be cheaper in some instances.

More importantly, it allows us to propagate additional information without increasing the amount of copying up and down the iterator stack. I anticipate this being useful in the implementation of cockroachdb/cockroach#111001, allowing us to propagate checksums alongside keys and values. Future work may add an explicit error return value.


RaduBerinde:
*[`internal/base/internal.go` line 511 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3498#-Nv4BV3V2wajK4qLU4gx:-Nv4BV3V2wajK4qLU4gy:b-z6ow9e) ([raw file](https://github.com/cockroachdb/pebble/blob/129a8bbc6b139e1ff20cfbc08ffa4d465ec77ed3/internal/base/internal.go#L511)):*
> ```Go
> // InternalKV represents a single internal key-value pair.
> type InternalKV struct {
> 	InternalKey
> ```

These types have quite a few methods that would be confusing against `InternalKV`.. E.g. `.Size()` would be just the key size. IMO we should not embed these. We can manually implement very common "shortcuts" like `InPlaceValue()` for `InternalKV`.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3498)* status: 0 of 60 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/base/internal.go` line 511 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3498#-Nv4BV3V2wajK4qLU4gx:-Nv8NwCS4_wL7TufTmX2:b-oqmgto) ([raw file](https://github.com/cockroachdb/pebble/blob/129a8bbc6b139e1ff20cfbc08ffa4d465ec77ed3/internal/base/internal.go#L511)):*

These types have quite a few methods that would be confusing against `InternalKV`.. E.g. `.Size()` would be just the key size. IMO we should not embed these. We can manually implement very common "shortcuts" like `InPlaceValue()` for `InternalKV`.
</blockquote></details>

That's reasonable. Done.

RaduBerinde:
*[`internal/base/internal.go` line 511 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3498#-Nv4BV3V2wajK4qLU4gx:-Nv8RrQo0K2moa-Hp6r4:b1f2914) ([raw file](https://github.com/cockroachdb/pebble/blob/129a8bbc6b139e1ff20cfbc08ffa4d465ec77ed3/internal/base/internal.go#L511)):*

That's reasonable. Done.
</blockquote></details>

nit: the field names can be just Key and Val or even K and V

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3498)* status: 0 of 60 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/base/internal.go` line 511 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3498#-Nv4BV3V2wajK4qLU4gx:-Nv8ht_JBrhCUktny8hZ:b8vpu09) ([raw file](https://github.com/cockroachdb/pebble/blob/129a8bbc6b139e1ff20cfbc08ffa4d465ec77ed3/internal/base/internal.go#L511)):*

nit: the field names can be just Key and Val or even K and V
</blockquote></details>

Sure, did K and V

PR #3499, opened 2024-04-10
all: replace path.Join with filepath.Join

imalasong:
path.Join is separated by '/', but the window system file separator is' \ '

filepath.Join can automatically select delimiters based on the system


xxmplus:
Close to due to age

PR #3501, opened 2024-04-11
tool: add db excise command

RaduBerinde:
#### tool: minor fix to the tests

The tests clone any paths in the arguments before running each
command. This in principle is wrong when the test modifies the
database and expects to observe these changes (as in `db_get_set`). It
just so happens that Clone does not remove any new files added, so new
markers and manifests will be there.

To eliminate this subtlety, we clone each path only the first time we
see it (within each test file).

#### tool: add db excise command

We add a command that can be used to excise a span.

In a follow-up change, we will extend this with a new option to allow
the span to be specified via other flags from Cockroach.


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3501#-:-NvCte2b1p2RFDgeZGXa:bnfp4nl)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`tool/db.go` line 534 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3501#-NvCrmTL1y_1xclyYVRM:-NvCrmTL1y_1xclyYVRN:baccq7m) ([raw file](https://github.com/cockroachdb/pebble/blob/7939d661579fbd5d0195d383c7f5760f89c0730d/tool/db.go#L534)):*
> ```Go
> 	dbOpts := d.opts.EnsureDefaults()
> 	dbDir := args[0]
> 	db, err := d.openDB(dbDir, nonReadOnly{})
> ```

Maybe we should disable table stats collection. If there's a corrupt sstable rather than just a missing sstable, we may try to asynchronously load it as a part of the table stats collection and Fatalf before we complete the excise. Or we could disable the fataling behavior when we encounter an ErrCorruption.
___
*[`tool/db.go` line 589 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3501#-NvCsRlQ8cMFz6Q9-2Km:-NvCsRlQ8cMFz6Q9-2Kn:bnkkfww) ([raw file](https://github.com/cockroachdb/pebble/blob/7939d661579fbd5d0195d383c7f5760f89c0730d/tool/db.go#L589)):*
> ```Go
> 		fmt.Fprintf(stderr, "Error excising: %s\n", err)
> 		return
> 	}
> ```

I was thinking about whether or not we're guaranteed that the sstable is actually gone from the Version (so that again in the corruption case, the rebooted Cockroach node doesn't panic trying to asynchronously load table stats for a table that's corrupt). I think we are guaranteed, because when we Open the engine we'll replay the WALs and then synchronously flush, which should ensure there's no memtable overlap when we run IngestAndExcise. Since there's no memtable overlap, the IngestAndExcise should not fall into the flushable ingest pathway and should synchronously remove the excised files from the version. Maybe put some of this in a comment here?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3501)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`tool/db.go` line 534 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3501#-NvCrmTL1y_1xclyYVRM:-NvCtxI9AjzwZUmYDibd:b-sn99e3) ([raw file](https://github.com/cockroachdb/pebble/blob/7939d661579fbd5d0195d383c7f5760f89c0730d/tool/db.go#L534)):*

Maybe we should disable table stats collection. If there's a corrupt sstable rather than just a missing sstable, we may try to asynchronously load it as a part of the table stats collection and Fatalf before we complete the excise. Or we could disable the fataling behavior when we encounter an ErrCorruption.
</blockquote></details>

Hrm, in the case the sstable is actually missing, will we still open okay? Or will we try to synchronously stat the file, fail and abort?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3501)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`tool/db.go` line 534 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3501#-NvCrmTL1y_1xclyYVRM:-NvElauy-4IPxw4h1vEB:b5rd0mh) ([raw file](https://github.com/cockroachdb/pebble/blob/7939d661579fbd5d0195d383c7f5760f89c0730d/tool/db.go#L534)):*

Hrm, in the case the sstable is actually missing, will we still open okay? Or will we try to synchronously stat the file, fail and abort?
</blockquote></details>

I think we should open ok. But I need to add a test with an actual external sst that's missing.

Disabling table stats is a good point. I should also disable automatic compactions.
___
*[`tool/db.go` line 589 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3501#-NvCsRlQ8cMFz6Q9-2Km:-NvElhW77QD5o0fB2z5v:b-v3b6mb) ([raw file](https://github.com/cockroachdb/pebble/blob/7939d661579fbd5d0195d383c7f5760f89c0730d/tool/db.go#L589)):*

I was thinking about whether or not we're guaranteed that the sstable is actually gone from the Version (so that again in the corruption case, the rebooted Cockroach node doesn't panic trying to asynchronously load table stats for a table that's corrupt). I think we are guaranteed, because when we Open the engine we'll replay the WALs and then synchronously flush, which should ensure there's no memtable overlap when we run IngestAndExcise. Since there's no memtable overlap, the IngestAndExcise should not fall into the flushable ingest pathway and should synchronously remove the excised files from the version. Maybe put some of this in a comment here?
</blockquote></details>

Hm, I thought that the excise on the LSM happens regardless (in `ingestApply`)?

Maybe I should just pass `false` for `sstsContainExciseTombstone`? I think that would always prevent a flushable ingest

RaduBerinde (commented):
Updated. We now disable stats, compactions, and the consistency check. We also needed some minor tweaks to avoid panics. We now have a test against a real database with an inaccessible external file.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3501#-:-NvIm5vd3J92U8R3DBqr:b-5g512s)* status: 0 of 21 files reviewed, 2 unresolved discussions (waiting on @jbowens)

jbowens (approved):
Nice!

:lgtm:

> Reviewed 6 of 21 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3501#-:-NvJIRJk5sDbKNYyXy2L:b2pss2z)* status: 6 of 21 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`objstorage/objstorage.go` line 371 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3501#-NvJHxDf06UstqawVwtP:-NvJHxDf06UstqawVwtQ:bspytrw) ([raw file](https://github.com/cockroachdb/pebble/blob/e6e52408bb4741258d65b1796abf9b758e33c8b2/objstorage/objstorage.go#L371)):*
> ```Go
> }
> 
> // IsLocalTable returns ture if a table with the given fileNum exists and is
> ```

nit: s/ture/true/

PR #3507, opened 2024-04-12
db: refactor Reader.Get

jbowens:
**db: refactor Reader.Get**

This commit refactors the implementation of getIter to be a little more
understandable and avoid the unnecessary use of levelIter. Current supported
format major versions guarantee that a user key is not split across sstables
within a level. This ensures that Get (which only retrieves one individual user
key) need only consult 1 sstable per level.

This is somewhat motivated by https://github.com/cockroachdb/pebble/issues/2863. Removing getIter's dependency on levelIter
will make that refactor easier.

**db: simplify levelIter skipEmptyFileForward**

Apply a small simplification to levelIter.skipEmptyFileForward's condition for
when to interleave a synthetic boundary. Previously we needed to check whether
a file's largest point key was a range deletion to work around getIter's use of
the rangeDelIterPtr. Now that getIter no longer uses levelIter it's
unnecessary.

### DIFF
--- a/db.go
+++ b/db.go
@@ -557,15 +557,25 @@ func (d *DB) getInternal(key []byte, b *Batch, s *Snapshot) ([]byte, io.Closer, 
 
 	get := &buf.get
 	*get = getIter{
-		logger:   d.opts.Logger,
 		comparer: d.opts.Comparer,
 		newIters: d.newIters,
 		snapshot: seqNum,
-		key:      key,
-		batch:    b,
-		mem:      readState.memtables,
-		l0:       readState.current.L0SublevelFiles,
-		version:  readState.current,
+		iterOpts: IterOptions{
+			// TODO(sumeer): replace with a parameter provided by the caller.
+			CategoryAndQoS: sstable.CategoryAndQoS{
+				Category: "pebble-get",
+				QoSLevel: sstable.LatencySensitiveQoSLevel,
+			},
+			logger:                        d.opts.Logger,
+			snapshotForHideObsoletePoints: seqNum,
+		},
+		key: key,
+		// Compute the key prefix for bloom filtering.
+		prefix:  key[:d.opts.Comparer.Split(key)],
+		batch:   b,
+		mem:     readState.memtables,
+		l0:      readState.current.L0SublevelFiles,
+		version: readState.current,
 	}
 
 	// Strip off memtables which cannot possibly contain the seqNum being read
--- a/get_iter.go
+++ b/get_iter.go
@@ -11,7 +11,6 @@ import (
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/keyspan"
 	"github.com/cockroachdb/pebble/internal/manifest"
-	"github.com/cockroachdb/pebble/sstable"
 )
 
 // getIter is an internal iterator used to perform gets. It iterates through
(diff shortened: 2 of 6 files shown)


RaduBerinde (approved):
:lgtm: Awesome! Some minor comments.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3507#-:-NvIpL0m-Ko0ocvIjLEL:bcyrst2)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`get_iter.go` line 77 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3507#-NvInvl5A9uWxsWedlh3:-NvInvl5A9uWxsWedlh4:b-1qn0fp) ([raw file](https://github.com/cockroachdb/pebble/blob/8549c49ea751ff1b55b3be49f4e5fca98d571d43/get_iter.go#L77)):*
> ```Go
> }
> 
> func (g *getIter) Next() *base.InternalKV {
> ```

Do we ever call `Next()` on this iterator? It seems strange, since we'd only care about the first result.
___
*[`get_iter.go` line 243 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3507#-NvIonxp548zQIbHqnt-:-NvIonxq69t5wuQBUAJy:bhlk7pp) ([raw file](https://github.com/cockroachdb/pebble/blob/8549c49ea751ff1b55b3be49f4e5fca98d571d43/get_iter.go#L243)):*
> ```Go
> }
> 
> func (g *getIter) getSSTableIterators(
> ```

This could use a comment - we are getting iterators for the one sstable that might contain the key.
___
*[`get_iter.go` line 255 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3507#-NvIoMLFE3E1iHokP5vz:-NvIoMLFE3E1iHokP5w-:b-7ek0tw) ([raw file](https://github.com/cockroachdb/pebble/blob/8549c49ea751ff1b55b3be49f4e5fca98d571d43/get_iter.go#L255)):*
> ```Go
> 	// the file doesn't actually contain any point keys equal to `key`. We next
> 	// to keep searching for a file that actually contains point keys ≥ key.
> 	if m.LargestPointKey.IsExclusiveSentinel() && g.comparer.Equal(m.LargestPointKey.UserKey, g.key) {
> ```

Not related to this change, but I think we should make `SeekGE` take a `base.UserKeyBoundary`

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3507#-:-NvJ1sWA0ROMOg5j-nYH:b-814g8k)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`get_iter.go` line 77 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3507#-NvInvl5A9uWxsWedlh3:-NvJ1bTC4FhfhcviwDo_:b4h3arl) ([raw file](https://github.com/cockroachdb/pebble/blob/8549c49ea751ff1b55b3be49f4e5fca98d571d43/get_iter.go#L77)):*

Do we ever call `Next()` on this iterator? It seems strange, since we'd only care about the first result.
</blockquote></details>

yeah, but I think only in the case the first key is a MERGE. added a comment
___
*[`get_iter.go` line 243 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3507#-NvIonxp548zQIbHqnt-:-NvJ1n51DZYRkF2CChBd:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8549c49ea751ff1b55b3be49f4e5fca98d571d43/get_iter.go#L243)):*

This could use a comment - we are getting iterators for the one sstable that might contain the key.
</blockquote></details>

Done.
___
*[`get_iter.go` line 255 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3507#-NvIoMLFE3E1iHokP5vz:-NvJ1oBKBwc5oVgioHU8:b-s6xpjt) ([raw file](https://github.com/cockroachdb/pebble/blob/8549c49ea751ff1b55b3be49f4e5fca98d571d43/get_iter.go#L255)):*

Not related to this change, but I think we should make `SeekGE` take a `base.UserKeyBoundary`
</blockquote></details>

yeah, makes sense

Issue #999, opened 2020-11-17
sstable: revisit writer block flushing heuristics


petermattis:
`sstable.shouldFlush` contains the heuristics for deciding whether a block should be flushed or not during sstable construction. The intention is to flush a block before it reaches the configured `blockSize`. The block size controls how large the block will be in memory (on disk the block is compressed and may be significantly smaller). For CRDB, a block size of 32 KiB is used. What is interesting is how this block size interacts with jemalloc (the memory allocator linked into CRDB for C memory allocations). Jemalloc has many size classes and an allocation is always performed in the smallest size class that will hold it. Of interest to the block cache are the size classes 24 KiB, 28 KiB, 32 KiB, 40 KiB. If a block is just a tiny bit larger than 32 KiB it will be allocated from the 40 KiB size class which will waste ~25% of the space. Much better for a block to be just a bit smaller than 32 KiB.

The `shouldFlush` heuristic attempts to flush a block just before it grows larger than the target block size. But there is a second heuristic that says "don't flush a block if it is smaller than 99% of the target block size". (Note this heuristic was inherited from RocksDB). 99% of 32 KiB is 31.68 KiB, a difference of only 328 bytes. So if we have a key/value pair that is larger than 328 bytes we'll flush the block when it is just a little bit larger than 32 KiB. If we want to minimize internal fragmentation in the block cache, we should instead allow the block to be flushed earlier. If the block was flushed at just over 28 KiB internal fragmentation would be 14%. 

Making `shouldFlush` aware of the jemalloc size classes could allow significantly reduce this memory wastage. Does this matter in practice? Maybe. CRDB is frequently run with multi-gigabyte block cache sizes. The internal fragmentation is not accounted for in the block cache size which makes memory usage higher than expected. Smarter block sizing heuristics could bring a tighter bound on memory usage which we could use to reduce the CRDB memory footprint, or we could increase the block cache size in order to improve read performance. With a multi-gigabyte block cache we're talking about hundreds of megabytes of memory. 

We'd want to make the allocator size class knowledge a configurable so that we don't hard code something specific to jemalloc.

petermattis:
Also related to internal fragmentation is this TODO in `internal/cache/value_normal.go`:

```go
func newValue(n int) *Value {
	if n == 0 {
		return nil
	}
	// When we're not performing leak detection, the lifetime of the returned
	// Value is exactly the lifetime of the backing buffer and we can manually
	// allocate both.
	//
[...]

PR #3508, opened 2024-04-12
sstable: reduce block cache fragmentation

CheranMahalingam:
Currently, the sstable writer contains heuristics to flush sstable blocks once the size reaches a specified threshold. In CRDB this is defined as 32KiB. However, when these blocks are loaded into memory additional metadata is allocated sometimes exceeding the 32KiB threshold. Since CRDB uses jemalloc, these allocations use a 40KiB size class which leads to significant internal fragmentation. In addition, since the system is unaware of these size classes we cannot design heuristics that prioritize reducing memory fragmentation. Reducing internal fragmentation can help reduce CRDB's memory footprint. This commit decrements the target block size to prevent internal fragmentation for small key-value pairs and adds support for optionally specifying size classes to enable a new set of heuristics that will reduce internal fragmentation for workloads with larger key-value pairs.

Fixes: https://github.com/cockroachdb/pebble/issues/999.

### DIFF
--- a/internal/base/options.go
+++ b/internal/base/options.go
@@ -6,9 +6,10 @@ package base
 
 // SSTable block defaults.
 const (
-	DefaultBlockRestartInterval = 16
-	DefaultBlockSize            = 4096
-	DefaultBlockSizeThreshold   = 90
+	DefaultBlockRestartInterval      = 16
+	DefaultBlockSize                 = 4096
+	DefaultBlockSizeThreshold        = 90
+	SizeClassAwareBlockSizeThreshold = 60
 )
 
 // FilterType is the level at which to apply a filter: block or table.
--- a/internal/cache/value.go
+++ b/internal/cache/value.go
@@ -4,6 +4,13 @@
 
 package cache
 
+import "unsafe"
+
+// ValueMetadataSize denotes the number of bytes of metadata allocated for a
+// cache entry. Note that for builds with cgo disabled no metadata is allocated,
+// however, we keep the value constant to reduce friction for writing tests.
+const ValueMetadataSize = int(unsafe.Sizeof(Value{}))
+
 // Value holds a reference counted immutable value.
 type Value struct {
 	buf []byte
--- a/internal/cache/value_cgo.go
+++ b/internal/cache/value_cgo.go
@@ -1,0 +1,39 @@
+// Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+//go:build ((!invariants && !tracing) || race) && cgo
+// +build !invariants,!tracing race
+// +build cgo
+
+package cache
+
+import (
+	"unsafe"
+
+	"github.com/cockroachdb/pebble/internal/manual"
+)
+
+func newValue(n int) *Value {
+	if n == 0 {
+		return nil
+	}
+
+	// When we're not performing leak detection, the lifetime of the returned
+	// Value is exactly the lifetime of the backing buffer and we can manually
+	// allocate both.
+	b := manual.New(ValueMetadataSize + n)
+	v := (*Value)(unsafe.Pointer(&b[0]))
+	v.buf = b[ValueMetadataSize:]
+	v.ref.init(1)
+	return v
+}
+
+func (v *Value) free() {
+	// When we're not performing leak detection, the Value and buffer were
+	// allocated contiguously.
+	n := ValueMetadataSize + cap(v.buf)
+	buf := (*[manual.MaxArrayLen]byte)(unsafe.Pointer(v))[:n:n]
+	v.buf = nil
+	manual.Free(buf)
+}
--- a/internal/cache/value_normal.go
+++ b/internal/cache/value_normal.go
@@ -2,56 +2,23 @@
 // of this source code is governed by a BSD-style license that can be found in
 // the LICENSE file.
 
-//go:build (!invariants && !tracing) || race
+//go:build ((!invariants && !tracing) || race) && !cgo
 // +build !invariants,!tracing race
+// +build !cgo
 
 package cache
-
-import (
-	"unsafe"
-
-	"github.com/cockroachdb/pebble/internal/manual"
-)
-
-const valueSize = int(unsafe.Sizeof(Value{}))
 
 func newValue(n int) *Value {
 	if n == 0 {
 		return nil
 	}
 
-	if !cgoEnabled {
-		// If Cgo is disabled then all memory is allocated from the Go heap and we
-		// can't play the trick below to combine the Value and buffer allocation.
-		v := &Value{buf: make([]byte, n)}
-		v.ref.init(1)
-		return v
-	}
-
-	// When we're not performing leak detection, the lifetime of the returned
-	// Value is exactly the lifetime of the backing buffer and we can manually
-	// allocate both.
-	//
-	// TODO(peter): It may be better to separate the allocation of the value and
-	// the buffer in order to reduce internal fragmentation in malloc. If the
-	// buffer is right at a power of 2, adding valueSize might push the
-	// allocation over into the next larger size.
-	b := manual.New(valueSize + n)
-	v := (*Value)(unsafe.Pointer(&b[0]))
-	v.buf = b[valueSize:]
+	// Since Cgo is disabled then all memory is allocated from the Go heap we
+	// can't play the trick below to combine the Value and buffer allocation.
+	v := &Value{buf: make([]byte, n)}
 	v.ref.init(1)
 	return v
 }
 
 func (v *Value) free() {
-	if !cgoEnabled {
-		return
-	}
-
-	// When we're not performing leak detection, the Value and buffer were
-	// allocated contiguously.
-	n := valueSize + cap(v.buf)
-	buf := (*[manual.MaxArrayLen]byte)(unsafe.Pointer(v))[:n:n]
-	v.buf = nil
-	manual.Free(buf)
 }
--- a/options.go
+++ b/options.go
@@ -1060,6 +1060,12 @@ type Options struct {
 	// to temporarily persist data spilled to disk for row-oriented SQL query execution.
 	EnableSQLRowSpillMetrics bool
 
+	// AllocatorSizeClasses provides a sorted list containing the supported size
+	// classes of the underlying memory allocator. This provides hints to the
+	// sstable block writer's flushing policy to select block sizes that
+	// preemptively reduce internal fragmentation when loaded into the block cache.
+	AllocatorSizeClasses []int
+
 	// private options are only used by internal tests or are used internally
 	// for facilitating upgrade paths of unconfigurable functionality.
 	private struct {
@@ -1970,6 +1976,7 @@ func (o *Options) MakeWriterOptions(level int, format sstable.TableFormat) sstab
 	writerOpts.FilterPolicy = levelOpts.FilterPolicy
 	writerOpts.FilterType = levelOpts.FilterType
 	writerOpts.IndexBlockSize = levelOpts.IndexBlockSize
+	writerOpts.AllocatorSizeClasses = o.AllocatorSizeClasses
 	return writerOpts
 }
 
--- a/sstable/options.go
+++ b/sstable/options.go
@@ -132,8 +132,16 @@ type WriterOptions struct {
 	// specified percentage of the target block size and adding the next entry
 	// would cause the block to be larger than the target block size.
 	//
-	// The default value is 90
+	// The default value is 90.
 	BlockSizeThreshold int
+
+	// SizeClassAwareThreshold imposes a minimum block size restriction for blocks
+	// to be flushed, that is computed as the percentage of the target block size.
+	// Note that this threshold takes precedence over BlockSizeThreshold when
+	// valid AllocatorSizeClasses are specified.
+	//
+	// The default value is 60.
+	SizeClassAwareThreshold int
 
 	// Cache is used to cache uncompressed blocks from sstables.
 	//
@@ -230,6 +238,12 @@ type WriterOptions struct {
 	// 750MB sstables -- see
 	// https://github.com/cockroachdb/cockroach/issues/117113).
 	DisableValueBlocks bool
+
+	// AllocatorSizeClasses provides a sorted list containing the supported size
+	// classes of the underlying memory allocator. This provides hints to the
+	// writer's flushing policy to select block sizes that preemptively reduce
+	// internal fragmentation when loaded into the block cache.
+	AllocatorSizeClasses []int
 }
 
 func (o WriterOptions) ensureDefaults() WriterOptions {
@@ -241,6 +255,9 @@ func (o WriterOptions) ensureDefaults() WriterOptions {
 	}
 	if o.BlockSizeThreshold <= 0 {
 		o.BlockSizeThreshold = base.DefaultBlockSizeThreshold
+	}
+	if o.SizeClassAwareThreshold <= 0 {
+		o.SizeClassAwareThreshold = base.SizeClassAwareBlockSizeThreshold
 	}
 	if o.Comparer == nil {
 		o.Comparer = base.DefaultComparer
--- a/sstable/testdata/flush_heuristics
+++ b/sstable/testdata/flush_heuristics
@@ -1,0 +1,54 @@
+# Block size exceeds target block size.
+build key-size=0 val-size=0 block-size=64 target-size=64 threshold=59 size-class-threshold=0
+----
+true
+
+# Block size does not exceed threshold size.
+build key-size=0 val-size=0 block-size=59 target-size=64 threshold=59 size-class-threshold=0
+----
+false
+
+# New block size exceeds the target size.
+build key-size=1 val-size=1 block-size=60 target-size=64 threshold=32 size-class-threshold=0
+----
+true
+
+# New block size does not exceed the target size.
+build key-size=1 val-size=1 block-size=40 target-size=64 threshold=32 size-class-threshold=0
+----
+false
+
+# New block size does not exceed the target size with hints enabled.
+build key-size=1 val-size=1 block-size=8 target-size=64 threshold=0 size-class-threshold=0 hints=8,16,32,64,128
+----
+false
+
+# Block size does meet size class aware threshold with hints enabled.
+build key-size=1 val-size=1 block-size=6 target-size=64 threshold=0 size-class-threshold=38 hints=8,16,32,64,128
+----
+false
+
+# New block size reduces internal fragmentation.
+build key-size=1 val-size=60 block-size=8 target-size=64 threshold=0 size-class-threshold=38 hints=8,16,32,64,128
+----
+false
+
+# New block size increases internal fragmentation.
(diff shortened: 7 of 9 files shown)


CheranMahalingam:
The internal fragmentation reduction is mostly observable for small key value pairs where data is packed efficiently near the 32KiB threshold. In these cases the 32B of metadata for a block cache entry can cause jemalloc to allocate a 40KiB block instead of a 32KiB block. We expect that these changes will not reduce fragmentation for large key value pairs, however, it should not introduce any memory regressions.

The following tests involved running kv-50 workloads on a single cockroach node with 4vCPUs and 16GB of RAM with varying block sizes. To illustrate the fragmentation improvements and shorten the test duration the block cache size was reduced to 400MB.


## TLDR
| KV Block Size | Old Average Fragmentation (B/Alloc) | New Average Fragmentation (B/Alloc) | Fragmentation Reduction (%) | Old Jemalloc Total Bytes (MB) | New Jemalloc Total Bytes (MB) | Jemalloc Memory Usage Reduction (%) |
| - | - | - | - | - | - | - |
| 1-16B | 3630 | 1221 | 63.4 | 650 | 510 | 21.5 |
| 1-1024B |

## Test 1 (KV Block Size: 1-16B)
### Previous System
Allocations: 14393295
Average Allocation Size (B): 29612
| Bin Size (B) | Count | Fragmentation (B/Alloc) |
| - | - | - |
| 48 | 535 | 4320 |
| 40 | 4239577 | 8178 |
| 32 | 3433477 | 15 |
| 28 | 6640636 | 2603 |
| 24 | 12007 | 1672 |
| 20 | 10329 | 2033 |

> [!NOTE]
> For small key value pairs a large number of allocations fall into the 40KiB size class and we are wasting ~8MiB of space on these allocations.

Screenshot 2024-04-12 at 3 29 18 PM

```
Block Cache Usage: 252MB
Jemalloc Allocated Bytes: ~420MB
Jemalloc Resident Bytes: ~650MB
```

### New Heuristics
Allocations: 30157396
Average Allocation Size (B): 29561
| Bin Size (B) | Count |  Fragmentation (B/Alloc) |
| - | - | - |
| 48 |  928  | 4432 |
| 40 | 3844 | 6006 |
| 32 | 15779481 | 30 |
| 28 | 14238856 | 2539 |
| 24 | 19798 | 1536 |
| 20 | 16900 | 1960 |

Screenshot 2024-04-12 at 2 43 54 PM

```
Block Cache Usage: 252MB
Jemalloc Allocated Bytes: ~386MB
Jemalloc Resident Bytes: ~510MB
```

## Test 2 (KV Block Size: 1-1024B)
### Previous System

### New Heuristics

## Test 3 (KV Block Size: 1-16384B)
### Previous System

### New Heuristics

CheranMahalingam (commented):
During my experiments I noticed that at runtime the `sizeThreshold` is now 90% of the target block size (29492B) and we seem to be using the default threshold (https://github.com/cockroachdb/pebble/blob/master/internal/base/options.go#L11). In this case there is about 3KiB of leeway, however, in cases where the key value pair is larger such as 4KiB, this threshold gets in the way and contributes to internal fragmentation. For key value pairs of 10-11KiB this isn't problematic since they will go over the 32KiB target but land just under the 40KiB size class.

The simplest solution I can think of when the block size is below the threshold is a check to see whether adding the next key value pair would decrease fragmentation (i.e. `32768 - blockSize > 40960 - futureBlockSize`). However, I am not sure what the behaviour should be for large key value pairs that set the block size beyond 40KiB and there is an argument that blocks above 32KiB could have less internal fragmentation if we waited until the size was closer to 40KiB. I am wondering whether we still want to pursue a size-class aware heuristic.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3508#-:-NvXT2C7-AAe32TpmLTP:bkt1gzl)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @CheranMahalingam)

CheranMahalingam (commented):
I can try a size-class aware heuristic and see how it performs with key value pair sizes that introduce a lot of fragmentation. Just for clarification should the `shouldFlush` logic look something like this, because I am not sure how we want to prioritize reducing internal fragmentation vs. keeping the sstable block size close to the 32KiB target and how performant this code path has to be?
```
if (size-class options are set) {
	if (blockSize > targetBlockSize) {
		nextSizeClass = sizeClassFunc(blockSize)
		if (nextSizeClass - blockSize > nextSizeClass - futureBlockSize 
                    && futureBlockSize <= nextSizeClass
                    // bound future block size to 40KiB to prevent rare scenario where blockSize continually increases without flushing because fragmentation decreases
                    && nextSizeClass == sizeClassFunc(targetBlockSize + 1)) {
			return false // wait because we can definitely reduce fragmentation in the future
		} else {
			return true
		}
	} else { // remove sizeThreshold heuristic
		if (futureBlockSize > targetBlockSize) {
			if (sizeClassFunc(futureBlockSize) - futureBlockSize < sizeClassFunc(blockSize) - blockSize) {
				return false // option that reduces fragmentation with a small risk of having a very large sstable block being flushed
			} else {
				return true
			}
		} else {
			return false
		}
	}
} else {
	existing logic…
}
```


Sorry, I reverted my changes while experimenting with changes that would have the Cockroach layer adjust the block size due to a significant number of test failures this change causes in Pebble, however, I found some workarounds in the latest revision.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3508#-:-NvbLXXd3HvwB205p2eH:bf49p43)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @CheranMahalingam)

petermattis (commented):
Yes, something like those heuristics is what I'm imagining. We do call `shouldFlush` on every KV pair, but I think you can have fast-path checks to make this very fast. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3508#-:-NvcdvY2-oCfWeZKv0jx:b-k5q5dr)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @CheranMahalingam)

CheranMahalingam (commented):
The size-class aware heuristics show a significant improvement in fragmentation for general workloads where the key-value pair size has a large range. However, these improvements do not directly correspond to a reduction in memory usage. The tests were run on a machine with 16vCPUs and 64GB of RAM where the block cache size was set to 16GB (also ran the following on the previous setup with similar results). We ran a kv-50 workload with key-value pair sizes ranging from 1B to 16KiB and enough concurrency/batching to saturate the cache.

## TLDR
| Version| Fragmentation (B/Alloc) | Space Wasted Per Allocation (%) | Jemalloc Allocated Bytes (GiB) | Jemalloc Total Bytes (GiB) |
| - | - | - | - | - |
| Current | 3755 | 11.26 | 16.621 | 19.096 |
| New Heuristics | 1730 | 6.21 | 15.785 | 18.726 |
| Improvement | | +5.05% | +5.03% | +1.94% |

## Results
### Current
```
Allocations: 16187686
Average Allocation Size (B): 33358

Jemalloc Stats
Average Allocated Bytes (GiB): 16.621
Average Resident Bytes (GiB): 19.096
```
| Bin Size (KiB) | Count | Fragmentation (B/Alloc) |
| - | - | - |
| 64 | 695 | 5024 |
| 56 | 4181 | 4816 |
| 48 | 1590425 | 6409 |
| 40 | 8343194 | 4658 |
| 32 | 5284819 | 1554 |
| 28 | 18039 | 2153 |
| 24 | 22626 | 2257 |
| 20 | 17984 | 2031 |
| 16 | 11573 | 920 |

Screenshot 2024-04-18 at 3 20 43 PM

### New Heuristics
```
Allocations: 16207665
Average Allocation Size (B): 27867

Jemalloc Stats
Average Allocated Bytes (GiB): 15.785
Average Resident Bytes (GiB): 18.726
```
| Bin Size (KiB) | Count | Fragmentation (B/Alloc) |
| - | - | - |
| 64 | 986 | 5173 |
| 56 | 6166 | 4151 |
| 48 | 101524 | 5263 |
| 40 | 2285861 | 1492 |
| 32 | 5855568 | 1818 |
| 28 | 3766744 | 1689 |
| 24 | 2492485 | 1776 |
| 20 | 740086 | 1346 |
| 16 | 13756 | 953 |

Screenshot 2024-04-18 at 3 51 59 PM

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3508#-:-NvmnNyu42ziNdwNgame:bjmjtb5)* status: 0 of 13 files reviewed, all discussions resolved (waiting on @CheranMahalingam)

petermattis (commented):
I didn't scrutinize the implementation, but the numbers here look compelling. I'm going to leave it to the Storage folks to do a detailed code review.

I'm surprised that Jemalloc Total Bytes didn't fall by the same size as Jemalloc Allocated Bytes. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3508#-:-NvqHjCRAwJGPbqgRurS:b5ie1m2)* status: 0 of 13 files reviewed, all discussions resolved (waiting on @CheranMahalingam)

sumeerbhola (changes requested):
> Reviewed 2 of 4 files at r2, 1 of 8 files at r3, 2 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3508)* status: 5 of 13 files reviewed, 1 unresolved discussion (waiting on @CheranMahalingam and @jbowens)

___
*[`sstable/writer.go` line 1656 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3508#-Nw7MEYZBI3Uu2N0sbKI:-Nw7MEYZBI3Uu2N0sbKJ:b-5cb16x) ([raw file](https://github.com/cockroachdb/pebble/blob/f874ba5b5d16ab44f0fd9f177be5a4db10adc394/sstable/writer.go#L1656)):*
> ```Go
> 	newSize += 4                            // varint for shared prefix length
> 	newSize += uvarintLen(uint32(keyLen))   // varint for unshared key bytes
> 	newSize += uvarintLen(uint32(valueLen)) // varint for value size
> ```

Worth adding a code comment that `newSize` is also an upper-bound, even though it is tighter than `newEstimatedSize` since it does not account for key prefix compression.
For very long keys with a long shared prefix, and tiny values, such as secondary indexes, this over-estimation could be a problem in that it could result in tiny blocks. I am not sure what to do about this given we don't want to incur the expense of computing the shared prefix (that we do in `blockWriter.storeWithOptionalValuePrefix`). We may want to use the `sizeThreshold` after all, and reduce it to something smaller, like 60% (and return false if `estimatedBlockSize <= sizeThreshold`).

I was expecting to see `cache.ValueMetadataSize` accounted for here in adjusting `estimatedBlockSize`, instead of in options. It will naturally bump up the effective `BlockSize` if someone sets a very tiny value, since we will add at least one entry. And after that entry is added the block will be bigger than `BlockSize` and all subsequent attempts to add will fall through to the size class based comparisons. Was there a reason to do it in options.go? The logic there seems not quite right in that if someone sets a block size of 2 bytes, they should be getting a tiny block, and not a block size of `DefaultBlockSize`.

sumeerbhola (commented):
Can you rerun the last experiment with these changes to validate that the results are still good.

> Reviewed 2 of 10 files at r6.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3508#-:-NwFSYF93HdpeeyD7hhO:brnylwu)* status: 6 of 14 files reviewed, 4 unresolved discussions (waiting on @CheranMahalingam and @jbowens)

___
*[`sstable/writer.go` line 1638 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3508#-NwFQOmr9zEhXc_pSXE3:-NwFQOmr9zEhXc_pSXE4:b5ft8ir) ([raw file](https://github.com/cockroachdb/pebble/blob/852a5dbff0920dc40d2db29cc3552aaf012c8bfb/sstable/writer.go#L1638)):*
> ```Go
> 	// loaded into the block cache we also allocate space for the cache entry
> 	// metadata. The new allocation of size ~1052B may now only fit within a
> 	// 2048B class, which increases fragmentation.
> ```

nit: increases internal fragmentation.
___
*[`sstable/writer.go` line 1649 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3508#-NwBxAJQDqr7JMNg9qPX:-NwFQpg41dxa5QZEic-V:b-79oz95) ([raw file](https://github.com/cockroachdb/pebble/blob/852a5dbff0920dc40d2db29cc3552aaf012c8bfb/sstable/writer.go#L1649)):*

This change is awkward because we already store size thresholds in the `Writer` and pass down this value to `shouldFlushWithHints`. We could make those thresholds dynamic by checking whether size class hints are empty, however, then if we are forced to fall back to the old heuristics the original 90% threshold is lost. Wondering whether there is a cleaner way to express this.
</blockquote></details>

Let's just compute this once in `Writer` and wrap them in `flushDecisionOptions` struct with 3 fields `blockSize`, `blockSizeThreshold`, `sizeClassAwareThreshold`, and pass that options down.
___
*[`sstable/writer.go` line 1674 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3508#-NwFRqZDEjzCmo9oyGva:-NwFRqZDEjzCmo9oyGvb:bgeqs30) ([raw file](https://github.com/cockroachdb/pebble/blob/852a5dbff0920dc40d2db29cc3552aaf012c8bfb/sstable/writer.go#L1674)):*
> ```Go
> 	if blockSizeWithMetadata < targetBlockSize {
> 		newSizeClass, ok := blockSizeClass(newSize, sizeClassHints)
> 		if !ok || newSizeClass-newSize >= sizeClass-blockSizeWithMetadata {
> ```

nit: shouldn't `!ok` result in returning false? Since then we should continue adding to this block and will eventually fall into one of the earlier cases that returns true. I realize `!ok` won't happen in a properly configured system.

CheranMahalingam (commented):
I can confirm that there are no significant regressions following these changes. There is a slight bump in fragmentation for the 40KiB size class which can be attributed to the 60% size threshold. There is also a slight increase in bytes allocated for the new heuristics and in total bytes for the old heuristics as reported by jemalloc. However, note that I ran this experiment almost 2x longer than the previous experiment to make sure that the computed averages were stable.

| Version| Fragmentation (B/Alloc) | Space Wasted Per Allocation (%) | Jemalloc Allocated Bytes (GiB) | Jemalloc Total Bytes (GiB) |
| - | - | - | - | - |
| Current | 3760 | 11.26 | 16.66 | 19.43 |
| New Heuristics | 1778 | 6.22 | 15.97 | 18.80 |
| Improvement | | +5.04% | +4.16% | +3.22% |

## Raw Results

<table>
<tr>
<th>Current</th>
<th>Heuristics v3</th>
</tr>

<tr>
<td>

```
Allocations: 30252804
Average Allocation Size (B): 33387

Jemalloc Stats
Average Allocated Bytes (GiB): 16.6588
Average Resident Bytes (GiB): 19.4298
```

| Bin Size (KiB) | Count | Fragmentation (B/Alloc) |
| - | - | - |
| 64 | 2136 | 3848 |
| 56 | 8812 | 4723 |
| 48 | 3025767 | 6402 |
| 40 | 15593149 | 4656 |
| 32 | 9793786 | 1546 |
| 28 | 44441 | 2137 |
| 24 | 47351 | 2206 |
| 20 | 34000 | 2047 |
| 16 | 21698 | 920 |

</td>
<td>

```
Allocations: 30246591
Average Allocation Size (B): 28582

Jemalloc Stats
Average Allocated Bytes (GiB): 15.9657
Average Resident Bytes (GiB): 18.8035
```

| Bin Size (KiB) | Count | Fragmentation (B/Alloc) |
| - | - | - |
| 64 | 1411 | 4191 |
| 56 | 8549 | 3880 |
| 48 | 101385 | 1879 |
| 40 | 5315069 | 2002 |
| 32 | 11033603 | 1794 |
| 28 | 7004610 | 1683 |
| 24 | 4602729 | 1774 |
| 20 | 536362 | 522 |
| 16 | 21589| 986 |

</td>
</tr>
</table>

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3508#-:-NwI3jWA2vaxo5apusXY:bw0uasr)* status: 3 of 14 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

CheranMahalingam (commented):
Pebble commit: https://github.com/CheranMahalingam/pebble/commit/43a74eec18ef063492005a0129f0ce97ad5c9154
Cockroach commit: https://github.com/CheranMahalingam/cockroach/commit/7ce83ae3570377b4890267acc19ec0c4e6aa7702

### Reproduction Steps
```
// Spin up single node cluster with sufficient memory
$ roachprod create --gce-machine-type=n2-standard-16 $CLUSTER -n 1

$ roachprod put $CLUSTER stats.sh
$ roachprod start $CLUSTER

// Run kv-50 workload
$ roachprod run $CLUSTER:1 -- './cockroach workload init kv {pgurl:1}'
$ roachprod run $CLUSTER:1 -- './cockroach workload run kv --read-percent 50 --tolerate-errors --max-block-bytes 16384 --min-block-bytes 1 --concurrency 256 --batch 10 --insert-count 100000 {pgurl:1}'

// SSH into the node to check progress
$ roachprod ssh $CLUSTER:1
// Wait until block cache has saturated in dbconsole and for sufficient time to pass, I used 30000000 allocations as a stopping point
$ cat logs/cockroach.stdout.log | grep "ALLOC" | wc -l
$ cat logs/cockroach.stdout.log > flush
// Note that there are some assumptions made in this script to get an accurate average for jemalloc stats which are documented inside the script
$ ./stats.sh
```

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3508#-:-NwL-FSGEiLNmvX1BNvI:b9a4ue0)* status: 3 of 15 files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/writer_test.go` line 675 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3508#-NwKTg_KEgwkLXtMJmIq:-NwKiPgg8HMhg_YobHbW:bn8xkyn) ([raw file](https://github.com/cockroachdb/pebble/blob/651178410a1d2c5b5f62278f8606b13ea71f6d0a/sstable/writer_test.go#L675)):*

nit: this is not the block size with metadata, yes?
</blockquote></details>

This is a hacky way of dealing with the fact that with certain builds metadata size is 0 (cgo disabled or a set of flags defined in `value_cgo.go`) and we want `blockSizeWithMetadata` to be consistent. Then, I decrement by `cache.ValueMetadataSize` (0 or 32) to compute the block size for the flush call.

I am realizing that this has been a huge pain for testing and to avoid future friction I have set `cache.ValueMetadataSize` to 32 for all builds. If cgo were ever disabled in production our heuristics would break anyways.

Note that there is still a large point of friction preventing extending the tests because the value of `cache.ValueMetadataSize` is reduced for 32-bit builds.
___
*[`sstable/writer_test.go` line 703 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3508#-NwKToHh05Hs-RkQKbyh:-NwKqjkU5wGxyNnYsQRX:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/651178410a1d2c5b5f62278f8606b13ea71f6d0a/sstable/writer_test.go#L703)):*

can you add some test cases where this sizeClassAwareThreshold is different from the blockSizeThreshold.
</blockquote></details>

Done.

PR #3511, opened 2024-04-12
db: fix levelIter boundaryContext comment

jbowens:
The comment about levelIter.boundaryContext was outdated and only discussed the use of sstable boundary keys to truncate untruncated range tombstones at read time. With the current version of Pebble, untruncated range tombstones no longer exist and the boundaryContext's usage is more constrained.

### DIFF
--- a/level_iter.go
+++ b/level_iter.go
@@ -107,52 +107,22 @@ type levelIter struct {
 	files            manifest.LevelIterator
 	err              error
 
-	// Pointer into this level's entry in `mergingIterLevel::levelIterBoundaryContext`.
-	// We populate it with the corresponding bounds for the currently opened file. It is used for
-	// two purposes (described for forward iteration. The explanation for backward iteration is
-	// similar.)
-	// - To limit the optimization that seeks lower-level iterators past keys shadowed by a range
-	//   tombstone. Limiting this seek to the file largestUserKey is necessary since
-	//   range tombstones are stored untruncated, while they only apply to keys within their
-	//   containing file's boundaries. For a detailed example, see comment above `mergingIter`.
-	// - To constrain the tombstone to act-within the bounds of the sstable when checking
-	//   containment. For forward iteration we need the smallestUserKey.
+	// Pointer into this level's mergingIterLevel.levelIterBoundaryContext.
+	// It's populated when the levelIter is in-use by a mergingIter. It's used
+	// to signal additional semantic meaning about the most recently returned
+	// key. It's currently used to pause at two different types of bounds:
 	//
-	// An example is sstable bounds [c#8, g#12] containing a tombstone [b, i)#7.
-	// - When doing a SeekGE to user key X, the levelIter is at this sstable because X is either within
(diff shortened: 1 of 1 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3511)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`level_iter.go` line 120 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3511#-NvJXj6eDrRGLu7GJOXG:-NvJXj6eDrRGLu7GJOXH:b-f08w1o) ([raw file](https://github.com/cockroachdb/pebble/blob/4aa3a30ce7c94d9077370e9d481ccea731a35ee9/level_iter.go#L120)):*
> ```Go
> 	//   has paused to avoid unnecessarily loading sstables outside the bounds.
> 	// - isIgnorableBoundaryKey is set to true when the levelIter returns a
> 	//   fake key at one of the bounds of an sstable with the level. It does
> ```

"within the level"?

These two sound very similar, from this description I can't tell in what way they are handled differently

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3511)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`level_iter.go` line 120 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3511#-NvJXj6eDrRGLu7GJOXG:-NvJnFMG8yj3Bid0aw2Z:bi96znm) ([raw file](https://github.com/cockroachdb/pebble/blob/4aa3a30ce7c94d9077370e9d481ccea731a35ee9/level_iter.go#L120)):*

"within the level"?

These two sound very similar, from this description I can't tell in what way they are handled differently
</blockquote></details>

Reworded a bit; the types of bounds are different. The first is the user-imposed iteration bounds, whereas the second is the bounds of an sstable within the level.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3511#-:-NvK-0w55RgkYkpezCQF:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`level_iter.go` line 120 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3511#-NvJXj6eDrRGLu7GJOXG:-NvJzsc80v4An-ySS19T:b-ybkfp3) ([raw file](https://github.com/cockroachdb/pebble/blob/4aa3a30ce7c94d9077370e9d481ccea731a35ee9/level_iter.go#L120)):*

Reworded a bit; the types of bounds are different. The first is the user-imposed iteration bounds, whereas the second is the bounds of an sstable within the level.
</blockquote></details>

Oh, I see, much better now, thanks!

jbowens:
TFTR!

PR #3512, opened 2024-04-12
*: remove bytesIterated

jbowens:
Remove the bytesIterated used in flushes and compactions to incrementally update compaction.bytesIterated. It was a vestige of an attempt at pacing flushes and compactions to reduce impact on foreground traffic. That pacing logic had been inactive for years and has been removed from the code base for a while.

The only remaining use was for the flush throughput metric, but this metric did not require an incrementally-updated value, only an aggregate after the flush has completed.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -2056,7 +2056,7 @@ func (b *flushableBatch) newIter(o *IterOptions) internalIterator {
 }
 
 // newFlushIter is part of the flushable interface.
-func (b *flushableBatch) newFlushIter(o *IterOptions, bytesFlushed *uint64) internalIterator {
+func (b *flushableBatch) newFlushIter(o *IterOptions) internalIterator {
 	return &flushFlushableBatchIter{
 		flushableBatchIter: flushableBatchIter{
 			batch:   b,
@@ -2065,7 +2065,6 @@ func (b *flushableBatch) newFlushIter(o *IterOptions, bytesFlushed *uint64) inte
 			cmp:     b.cmp,
 			index:   -1,
 		},
-		bytesIterated: bytesFlushed,
 	}
 }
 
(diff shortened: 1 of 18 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3512#-:-NvK1LTR0iyp70nKbWt5:bnfp4nl)* status: 0 of 18 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @jbowens)

___
*[`level_iter.go` line 20 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3512#-NvK198a8fap6g4Bs_DS:-NvK198a8fap6g4Bs_DT:b-16vxdt) ([raw file](https://github.com/cockroachdb/pebble/blob/91fc04e5898c83e50ee109014c28227f459b40f9/level_iter.go#L20)):*
> ```Go
> 
> type internalIterOpts struct {
> 	compaction         bool
> ```

[nit] maybe a small comment "// If compaction is set, we use a compaction iterators at the sstable level.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3512#-:-NvN9jsb4jq-NcyBbm91:b-814g8k)* status: 0 of 18 files reviewed, all discussions resolved (waiting on @aadityasondhi)

___
*[`level_iter.go` line 20 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3512#-NvK198a8fap6g4Bs_DS:-NvN9lhX2okQ4WGiJQ5D:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/91fc04e5898c83e50ee109014c28227f459b40f9/level_iter.go#L20)):*

[nit] maybe a small comment "// If compaction is set, we use a compaction iterators at the sstable level.
</blockquote></details>

Done

RaduBerinde:
[nit] We should update this comment too:

https://github.com/cockroachdb/pebble/blob/master/sstable/reader.go#L386

I think we should say it's optimized for the sequential reads pattern and only implements `Next` and `First` (btw should we have a separate interface for compaction iters?)

PR #3514, opened 2024-04-15
tool: add custom excise span option

RaduBerinde:
This commit adds an option to pass down a function for a custom method
to specify the excise span. On the CRDB side, we will add table/tenant
ID flags and when those flags are set, the function will return the
table span.

Informs [#122059](https://github.com/cockroachdb/cockroach/issues/122059)

### DIFF
--- a/tool/db.go
+++ b/tool/db.go
@@ -50,6 +50,7 @@ type dbT struct {
 	mergers         sstable.Mergers
 	openErrEnhancer func(error) error
 	openOptions     []OpenOption
+	exciseSpanFn    DBExciseSpanFn
 
 	// Flags.
 	comparerName  string
@@ -73,6 +74,7 @@ func newDB(
 	mergers sstable.Mergers,
 	openErrEnhancer func(error) error,
 	openOptions []OpenOption,
+	exciseSpanFn DBExciseSpanFn,
 ) *dbT {
 	d := &dbT{
 		opts:            opts,
@@ -80,6 +82,7 @@ func newDB(
 		mergers:         mergers,
 		openErrEnhancer: openErrEnhancer,
 		openOptions:     openOptions,
+		exciseSpanFn:    exciseSpanFn,
 	}
 	d.fmtKey.mustSet("quoted")
 	d.fmtValue.mustSet("[%x]")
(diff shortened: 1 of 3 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3514#-:-Nvhh5iV3UshE0ibJNva:bnfp4nl)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`tool/tool.go` line 120 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3514#-Nvhgwxg6AyXUploCyPf:-Nvhgwxg6AyXUploCyPg:bwhauzx) ([raw file](https://github.com/cockroachdb/pebble/blob/367febe4395b436c48a87306adc20b47b092fffe/tool/tool.go#L120)):*
> ```Go
> // WithDBExciseSpanFn specifies a function that returns the excise span for the
> // `db excise` command.
> func WithDBExciseSpanFn(fn DBExciseSpanFn) Option {
> ```

will the Cockroach code only pass in this option conditionally, so we still have the option of providing explicit --start, ---end flags?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3514#-:-NvhqaRfBUQz6j_KoAjc:b-814g8k)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`tool/tool.go` line 120 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3514#-Nvhgwxg6AyXUploCyPf:-NvhqTDX0GEsu6aJ72Zl:bjl3g9g) ([raw file](https://github.com/cockroachdb/pebble/blob/367febe4395b436c48a87306adc20b47b092fffe/tool/tool.go#L120)):*

will the Cockroach code only pass in this option conditionally, so we still have the option of providing explicit --start, ---end flags?
</blockquote></details>

If the function returns an invalid (zero) KeyRange, we fall back to using the `--start/end` flags.

PR #3517, opened 2024-04-16
wal,db: add FailoverWriteAndSyncLatency histogram

sumeerbhola:
FailoverWriteAndSyncLatency measures the latency of writing and syncing a set of writes that were synced together. Each sample represents the highest latency observed across the writes in the set of writes. It gives us a sense of the user-observed latency, which can be much lower than the underlying fsync latency, when WAL failover is working effectively.

Informs #122364

PR #3519, opened 2024-04-17
crl-release-24.1: wal,db: add FailoverWriteAndSyncLatency histogram

sumeerbhola:
24.1 backport of https://github.com/cockroachdb/pebble/pull/3517

---

FailoverWriteAndSyncLatency measures the latency of writing and syncing a set of writes that were synced together. Each sample represents the highest latency observed across the writes in the set of writes. It gives us a sense of the user-observed latency, which can be much lower than the underlying fsync latency, when WAL failover is working effectively.

Informs #122364


sumeerbhola:
TFTR!

PR #3520, opened 2024-04-17
db: remove fragmenters from compactionIter

RaduBerinde:
#### db: remove fragmenters from compactionIter

The tombstone and range key fragmenters in the compaction iterator are
no longer necessary (they spans are now fragmented by
`keyspan.MergingIter`).

This change replaces the fragmenters with simple lists of spans. This
makes the flow of how these spans are handled much easier to follow.

#### keyspan: remove no longer used Fragmenter code

Remove the `Fragmenter`'s `Covers` and `TruncateAndFlushTo` methods.


jbowens (approved):
:lgtm:

> Reviewed 3 of 3 files at r1, 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3520#-:-Nvhhh6t9V9cfqM2xEbD:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`compaction_iter.go` line 1411 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3520#-NvhhM6gEZr6h3Vliz4c:-NvhhM6gEZr6h3Vliz4d:b-itvhow) ([raw file](https://github.com/cockroachdb/pebble/blob/73a8789644aba5221c001d9e5b25d1d90d086834/compaction_iter.go#L1411)):*
> ```Go
> 	before = spans[:n:n]
> 	after = append(spans[n:], keyspan.Span{
> 		Start: key,
> ```

does `key` need to be cloned? if it's the key we're splitting at, it might be an blockIter-owned user key?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3520)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`compaction_iter.go` line 1411 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3520#-NvhhM6gEZr6h3Vliz4c:-Nvhoou2DNSkGRzQ7aOF:baraf7d) ([raw file](https://github.com/cockroachdb/pebble/blob/73a8789644aba5221c001d9e5b25d1d90d086834/compaction_iter.go#L1411)):*

does `key` need to be cloned? if it's the key we're splitting at, it might be an blockIter-owned user key?
</blockquote></details>

We were cloning it before the call, but I agree it makes more sense to do it here. I think we won't need to do it in the common case.

PR #3521, opened 2024-04-17
db: tweak merging iterator range deletion cascading logic

jbowens:
When the merging iterator finds a range deletion during a seek, it uses the tombstone to adjust the seek key of lower levels, skipping over deleted keys. Previously this code used (*keyspan.Span).Contains to ensure that the range deletion tombstone's bounds allowed the optimization. This had a couple minor unnecessary consequences:

 - It performed key comparisons against both bounds, despite keyspan.Seek[GE|LE] has already ensured one of the bounds is met.
 - It didn't handle the SeekLT(k) case where k is also the end bound of the range deletion. In this case, the optimization is still valid because SeekLT(k) is exclusive with respect to k.

### DIFF
--- a/merging_iter.go
+++ b/merging_iter.go
@@ -710,7 +710,7 @@ func (m *mergingIter) isNextEntryDeleted(item *mergingIterLevel) (bool, error) {
 			continue
 		}
 
-		if l.tombstone.VisibleAt(m.snapshot) && l.tombstone.Contains(m.heap.cmp, item.iterKV.K.UserKey) {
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3521#-:-NvhPIDo7YTM9jJJ5AQH:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @aadityasondhi)

jbowens:
TFTR!

PR #3522, opened 2024-04-17
crl-release-24.1: tool: add db excise command

RaduBerinde:
#### tool: add db excise command

We add a command that can be used to excise a span.

In a follow-up change, we will extend this with a new option to allow
the span to be specified via other flags from Cockroach.

Some of the code had to be adjusted to avoid panics during Open.

#### tool: add custom excise span option

This commit adds an option to pass down a function for a custom method
to specify the excise span. On the CRDB side, we will add table/tenant
ID flags and when those flags are set, the function will return the
table span.


jbowens (approved):
lgtm

PR #3523, opened 2024-04-17
base: move some test utils to base

RaduBerinde:
Move the fake iterator and the deletable sum value merger into base.
This will be useful when we move some of the compaction code into
another package.


RaduBerinde:
TFTR!

PR #3524, opened 2024-04-18
compact: move out compaction iterator

RaduBerinde:
#### compact: add Snapshots type

Add a `compact.Snapshots` type and move the `snapshotIndex`
implementation.

#### compact: move out compaction iterator

We move `compactionIter` to `compact.Iter`.

Note that the `compaction_iter` test file was not being run, which
explains all the unrelated output changes.


RaduBerinde:
TFTR!

PR #3528, opened 2024-04-18
internal/keyspan: refactor interleaving iterator tests

jbowens:
**internal/keyspan: refactor interleaving iterator tests**

Refactor the interleaving iterator tests to take the masking threshold on the
iterator command. Rename 'define-rangekeys' to 'define-spans'.

**internal/keyspan: fix interleaving iter intermediary state**

Previously in InterleavingIter.prevPos when
  a) moving from a point key
  b) the point key was contained within a span
  c) no lesser point key exists
prevPos would improperly move to posKeyspanEnd instead of posKeyspanStart. This
doesn't have any externally visible impact on the behavior of the
InterleavingIter because end boundaries are not interleaved: prevPos is called
again before returning to the user, ensuring we step to posKeyspanStart. With
cases.

### DIFF
--- a/internal/keyspan/interleaving_iter.go
+++ b/internal/keyspan/interleaving_iter.go
@@ -724,7 +724,7 @@ func (i *InterleavingIter) prevPos() {
 		case i.span == nil:
 			panic("withinSpan=true, but i.span == nil")
 		case i.pointKV == nil:
-			i.pos = posKeyspanEnd
+			i.pos = posKeyspanStart
 		default:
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3528#-:-NvnFNL7AWaV88YSx_py:bnfp4nl)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @aadityasondhi)

jbowens:
tftr!

Issue #1942, opened 2022-09-05
perf: optimize {Lower,Upper} bound handling in indexed batch


jbowens:
Similar to #98, we can use seeks to find the last node less than the lower bound and the first node greater than or equal to the upper bound. Next and Prev then only need to do a pointer equality check. Keys may be added to the skiplist while iterating, but those keys are not visible until a call to SetOptions is made to refresh the batch. Similar to #98, we can perform these seeks lazily. We might also consider a fast path that first compares against the first and last keys in the batch: a common pattern in Cockroach is to set iterator bounds to a KV range's bounds and only write keys within bounds. In these cases, we can perform 2 up front key comparisons to determine that all the batch keys are within bounds and elide all subsequent key comparisons.

Motivated by cockroachdb/cockroach#87277 which observed a significant performance hit to reading through an indexed batch.

Issue #98, opened 2019-04-28
perf: optimize {Lower,Upper}Bound handling in the memtable


petermattis:
The memtable currently checks the lower/upper bound on every iteration step. An alternative is to find the nodes which bound the lower and upper bound (using `SeekGE(lower)` and `SeekLT(upper)`) and then do pointer comparisons for the boundary checks. This will make the boundary checks a lot faster, but it does mean that we'll do an extra `Seek*` (we can do this lazily, but there will still be at least 1 extra seek). Note that the memtable can be inserted into concurrently with iteration, but those concurrent insertions are always for newer values and cannot affect the keys seen during an iteration.

@ajkr I was thinking about this after your comment on https://github.com/petermattis/pebble/pull/81#pullrequestreview-231332657. Thoughts?

ajkr:
Makes sense. I wonder if there are additional requirements on `SeekGE` and `SeekLT`. For example, before I believe the requirement on `SeekGE` was the seek key had to be at least the lower bound. Now I expect it'll also need to be at most the upper-bound. Otherwise the binary search could land after the saved upper-bound pointer and forward scan could proceed obliviously. I'd guess adding such a requirement (if we don't already have it) is fine though.

petermattis:
> I wonder if there are additional requirements on `SeekGE` and `SeekLT`. For example, before I believe the requirement on `SeekGE` was the seek key had to be at least the lower bound. Now I expect it'll also need to be at most the upper-bound. Otherwise the binary search could land after the saved upper-bound pointer and forward scan could proceed obliviously. I'd guess adding such a requirement (if we don't already have it) is fine though.

Yes, I think we would need additional checks as you suggest.

petermattis:
@Ryanfsdf This is something to experiment with. It isn't clear is worthwhile or not, so the first step would be to do some quick implementation and write benchmarks to see if there is an improvement. The code in question is in `internal/arenaskl`.

Ryanfsdf:
> The memtable currently checks the lower/upper bound on every iteration step. An alternative is to find the nodes which bound the lower and upper bound (using `SeekGE(lower)` and `SeekLT(upper)`) 
>

I think you meant `SeekGE(upper)` and `SeekLT(lower)` here, unless I'm misunderstanding something?

[...]

PR #3531, opened 2024-04-20
internal/{arenaskl,batchskl}: adjust behavior at bounds

jbowens:
**internal/batchskl: remove iterAdapter**

Remove the iterAdapter used in tests, testing the current batchskl.Iterator
interface directly in all unit tests.

**db: add variants of memtable iterator benchmarks with bounds**

Add variants of the memtable iterator benchmarks that have iteration bounds.

Informs #98.

**internal/{arenaskl,batchskl}: adjust behavior at bounds**

Previously, when an arenaskl.Iterator or batchskl.Iterator was exhausted at an
upper or lower bound, calling Prev or Next (respectively), would cause the
iterator to surface a key outside the bounds. This meant if one of these
iterators had a bound configured and became exhausted, the caller must re-seek
the iterator.

Informs #98.
Informs #1942.


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3531#-:-NvxHT9mBukPEBrgpf3C:bnfp4nl)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`internal/arenaskl/iterator.go` line 50 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3531#-NvxH7cy09mVgpWgnk0m:-NvxH7cz4ixp4vWHGtPL:b-x3udqb) ([raw file](https://github.com/cockroachdb/pebble/blob/affcde6c7145f9a76edab69d4828329fa7e34a61/internal/arenaskl/iterator.go#L50)):*
> ```Go
> 	// "arbitrary" because it may not be the first node that exceeds the bound.
> 	// Concurrent insertions into the skiplist may introduce new nodes with keys
> 	// that exceed the bounds but are closer to the  bounds than the current
> ```

[nit] double space

jbowens:
TFTR!

PR #3533, opened 2024-04-21
db: fix up a few more tests

RaduBerinde:
Add `CheckOrdering` and fix violations for some compaction tests.

PR #3534, opened 2024-04-21
manifest: fix LevelIterator.SeekGE semantics

RaduBerinde:
~Note: first commit is #3533.~

#### manifest: fix LevelIterator.SeekGE semantics

The `LevelIterator` (which iterates through files within a level)
`SeekGE` function had very murky semantics: as far as I can tell from
the code, a file that ends with an exclusive sentinel key that equals
the given user key is returned if and only if the filter is
`KeyTypePointAndRange`.

This change fixes the semantics to never return a file with an
exclusive sentinel key that equals userKey and improves the
documentation. The call sites which had to handle this case are
simplified.


jbowens (approved):
:lgtm:

> Reviewed 10 of 10 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3534#-:-Nw60Zm09DeCETVRzg5K:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/base/internal.go` line 384 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3534#-Nw6-qMh4imlM1RrY9MD:-Nw6-qMh4imlM1RrY9ME:b-o31abb) ([raw file](https://github.com/cockroachdb/pebble/blob/51ac285f292652850e3368f1b853e14053f3e4c3/internal/base/internal.go#L384)):*
> ```Go
> 
> // IsUpperBoundFor returns true if a range ending in k contains the userKey:
> // either k.UserKey < userKey or they are equal and k is not an exclusive
> ```

this should be reversed, right? "userKey < k.UserKey"

PR #3535, opened 2024-04-22
db: remove compaction.disableSpanElision

RaduBerinde:
We remove this field which is not very useful (tests can just pass a
large in-use key range instead).

We also consult `inUseEntireRange` in `elideRangeTombstone` which
seems like an omission. We also simplify `allowZeroSeqNum()`.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -288,10 +288,6 @@ type compaction struct {
 	// maxOverlapBytes is the maximum number of bytes of overlap allowed for a
 	// single output table with the tables in the grandparent level.
 	maxOverlapBytes uint64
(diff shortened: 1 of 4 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3535#-:-Nw572iN7-TKcF6uohIs:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #3538, opened 2024-04-22
db: simplify mergingIter direction switching

jbowens:
The mergingIter logic for switching the direction of the heap has special handling for the "synthetic iterator bounds" that the levelIter interleaves when it encounters the largest key in a table. These conditionals were more complicated and involved than they needed to be. The isSyntheticIterBoundsKey flag implies that the current key is an exclusive sentinel and that there's a bound in the previously-active direction of iteration.

### DIFF
--- a/level_iter.go
+++ b/level_iter.go
@@ -614,6 +614,17 @@ func (l *levelIter) verify(kv *base.InternalKV) *base.InternalKV {
 		if l.upper != nil && kv != l.largestBoundary && l.cmp(kv.K.UserKey, l.upper) > 0 {
 			l.logger.Fatalf("levelIter %s: upper bound violation: %s > %s\n%s", l.level, kv, l.upper, debug.Stack())
 		}
+
+		if l.boundaryContext != nil && l.boundaryContext.isSyntheticIterBoundsKey {
+			switch {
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3538#-:-Nw6SxYX5bt4TVATIKLy:bnfp4nl)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`merging_iter.go` line 57 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3538#-Nw6SfMX96bso0TqkmEz:-Nw6SfMY-r7s_TaCoeVC:boern91) ([raw file](https://github.com/cockroachdb/pebble/blob/a3797acb23007c3537dd2c6d2d344af670210f54/merging_iter.go#L57)):*
> ```Go
> 	// `isSyntheticIterBoundsKey || isIgnorableBoundaryKey`.
> 	//
> 	// When true, the coincident key IsExclusiveSentinel(), and the current
> ```

[nit] "must be an exclusive sentinel"

jbowens:
TFTR!

PR #3542, opened 2024-04-23
db: fix levelIter tests to obey bounds

jbowens:
The internal iterator contract requires callers to use SeekLT(upper) or SeekGE(lower) rather than First or Last when upper/lower bounds are set. This commit fixes the levelIter unit tests to avoid testing this case, and adds new invariant-build checks to ensure this contract is obeyed by levelIter users.

### DIFF
--- a/level_iter.go
+++ b/level_iter.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"runtime/debug"
 
+	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3542#-:-NwAlC3O5kCbdE7oxxAo:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @sumeerbhola)

PR #3543, opened 2024-04-23
manifest: more accurate in-use key ranges

RaduBerinde:
#### manifest: use UserKeyBounds for in use ranges

This is a mechanical change to remove `UserKeyRanges` and switch to
`UserKeyBounds` for in-use key ranges. They are still always
inclusive.

#### db: better printing of bounds in TestCompactionInuseKeyRanges

Use the `UserKeyBounds` format, which shows whether the end bound is
exclusive or inclusive.

#### manifest: more accurate in-use key ranges

We now take into account the exclusivity of the largest key, with more
accurate ("tighter") resulting ranges.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -322,7 +322,7 @@ type compaction struct {
 	// grandparent and lower levels. See setupInuseKeyRanges() for the
 	// construction. Used by elideTombstone() and elideRangeTombstone() to
 	// determine if keys affected by a tombstone possibly exist at a lower level.
-	inuseKeyRanges []manifest.UserKeyRange
+	inuseKeyRanges []base.UserKeyBounds
 	// inuseEntireRange is set if the above inuse key ranges wholly contain the
 	// compaction's key range. This allows compactions in higher levels to often
 	// elide key comparisons.
@@ -692,7 +692,7 @@ func (c *compaction) setupInuseKeyRanges() {
 	// compaction overlaps with an in-use span.
 	c.inuseEntireRange = len(c.inuseKeyRanges) > 0 &&
 		c.cmp(c.inuseKeyRanges[0].Start, c.smallest.UserKey) <= 0 &&
-		c.cmp(c.inuseKeyRanges[0].End, c.largest.UserKey) >= 0
+		c.inuseKeyRanges[0].End.IsUpperBoundFor(c.cmp, c.largest.UserKey)
 }
 
 // findGrandparentLimit takes the start user key for a table and returns the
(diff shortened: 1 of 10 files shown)


jbowens (approved):
:lgtm:

> Reviewed 8 of 8 files at r1, 2 of 2 files at r2, 1 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3543#-:-NwBZxu7C9huStQxtdSJ:bnfp4nl)* status: 6 of 10 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/manifest/version.go` line 1398 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3543#-NwBZWM63_QtGP1IxuGf:-NwBZWM63_QtGP1IxuGg:b-amlyrs) ([raw file](https://github.com/cockroachdb/pebble/blob/ed3496e6caf57ef159f32357605b9e4a7d02594e/internal/manifest/version.go#L1398)):*
> ```Go
> func seekGT(iter *LevelIterator, cmp base.Compare, boundary base.UserKeyBoundary) *FileMetadata {
> 	f := iter.SeekGE(cmp, boundary.Key)
> 	if f == nil || cmp(f.Largest.UserKey, boundary.Key) != 0 {
> ```

is this comparison still necessary? I think maybe when you updated the `LevelIterator.SeekGE` semantics this became unnecessary

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3543#-:-NwBmblQCjlQaX9T0eoc:b-814g8k)* status: 5 of 10 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`internal/manifest/version.go` line 1398 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3543#-NwBZWM63_QtGP1IxuGf:-NwBk3NMCqz-KTnfXmD0:b1rkuv6) ([raw file](https://github.com/cockroachdb/pebble/blob/ed3496e6caf57ef159f32357605b9e4a7d02594e/internal/manifest/version.go#L1398)):*

is this comparison still necessary? I think maybe when you updated the `LevelIterator.SeekGE` semantics this became unnecessary
</blockquote></details>

I added some comments, and restructured the code to perform the faster checks first. We want the first file that could extend our boundary, for example if our current boundary is `- u]` and the file also ends in `- u]`, we want the next file.

PR #3544, opened 2024-04-23
crl-release-24.1: db: check for empty external ingestions

RaduBerinde:
This commit adds a check that the external ingestions are not empty.
This is not allowed and leads to errors or correctness issues down the
line.

We don't want to open all remote ssts files in production, so the
check only happens in invariants mode or if a new
CheckExternalIngestions option is set. This will be set by crdb in
tests.


jbowens (approved):
lgtm

RaduBerinde:
TFTR!

PR #3547, opened 2024-04-23
db: clean up levelIter, mergingIter unit tests

jbowens:
The unit tests TestMergingIterCornerCases, TestLevelIterBoundaries and TestLevelIterSeek all fake tableNewIters implementations, but previously these implementations ignored the requested iterator types, always constructing both point and range deletion iterators.

Additionally, the output of TestLevelIterBoundaries is adjusted to make it clearer when the levelIter has actually opened a file.

### DIFF
--- a/internal/itertest/datadriven.go
+++ b/internal/itertest/datadriven.go
@@ -15,44 +15,65 @@ import (
 
 	"github.com/cockroachdb/datadriven"
 	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/keyspan"
 	"github.com/cockroachdb/pebble/internal/testkeys"
 	"github.com/stretchr/testify/require"
 )
 
 type iterCmdOpts struct {
-	fmtKV func(io.Writer, *base.InternalKey, []byte, base.InternalIterator)
-	stats *base.InternalIteratorStats
+	fmtKV           formatKV
+	withoutNewlines bool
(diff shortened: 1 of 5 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3547)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`internal/itertest/datadriven.go` line 33 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3547#-NwFjrwUEd0TQEXejGdi:-NwFnG4SDB-FN5miix6s:b-qryeub) ([raw file](https://github.com/cockroachdb/pebble/blob/5646b7fdec7e7af54515cb36f3182519746c2ac8/internal/itertest/datadriven.go#L33)):*

went the opposite way; we don't need `FormatKV` to be exported
</blockquote></details>

*`WithFormatKV`

RaduBerinde (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3547#-:-NwG25MGBfeZooeWcGYf:b-uwkowd)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

jbowens:
TFTR!

PR #3551, opened 2024-04-24
db: simplify mergingIter.switchTo{Max,Min}Heap

jbowens:
Previously, mergingIter.switchTo{Max,Min}Heap had complicated logic to avoid violating the levelIter's invariants when higher-level range deletions caused cascading seek keys to extend beyond iteration bounds. With 6fa3fe9d, the levelIter behaves like other internal iterators and is capable of being relative positioned away from an exhausted state without violating its bounds-checking invariants.

This commit applies the same logic of 6fa3fe9d to the synthetic iterator bound keys that are interleaved when a levelIter reaches a distant bound, allowing a relative positioning method to step away from these synthetic iterator bounds. This allows us to simplify mergingIter.switchTo{Max,Min}Heap.

### DIFF
--- a/level_iter.go
+++ b/level_iter.go
@@ -140,7 +140,7 @@ type levelIter struct {
 
 	// exhaustedDir is set to +1 or -1 when the levelIter has been exhausted in
 	// the forward or backward direction respectively. It is set when the
-	// underlying data is exhausted, not when iteration has reached the upper or
+	// underlying data is exhausted or when iteration has reached the upper or
 	// lower boundary and interleaved a synthetic iterator bound key. When the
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):
:lgtm: Nice cleanup!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3551#-:-NwGf5Qo335nruKdcYTt:b-d43tfz)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @sumeerbhola)

jbowens:
TFTR!

PR #3552, opened 2024-04-24
compaction: move tombstone elision logic to internal/compact

RaduBerinde:
We move the logic that sets up the in-use key ranges and then uses
those to determine which tombstones need to be elided.

We store the in-use key ranges in a new type `TombstoneElision`. The
`TombstoneElision` is passed to the compaction iterator, replacing the
`ElideTombstone` and `ElideRangeTombstone` callbacks.

New point and range elider types are used to check against the in-use
ranges. They assume the "queries" are ordered.

The minor test result changes are due to the new code being better at
taking into account exclusive boundaries.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -259,16 +259,8 @@ type compaction struct {
 	// L0Sublevels. If nil, flushes aren't split.
 	l0Limits [][]byte
 
-	// List of disjoint inuse key ranges the compaction overlaps with in
-	// grandparent and lower levels. See setupInuseKeyRanges() for the
-	// construction. Used by elideTombstone() and elideRangeTombstone() to
-	// determine if keys affected by a tombstone possibly exist at a lower level.
-	inuseKeyRanges []base.UserKeyBounds
-	// inuseEntireRange is set if the above inuse key ranges wholly contain the
-	// compaction's key range. This allows compactions in higher levels to often
-	// elide key comparisons.
-	inuseEntireRange    bool
-	elideTombstoneIndex int
+	delElision      compact.TombstoneElision
+	rangeKeyElision compact.TombstoneElision
 
 	// allowedZeroSeqNum is true if seqnums can be zeroed if there are no
 	// snapshots requiring them to be kept. This determination is made by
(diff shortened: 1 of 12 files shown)


RaduBerinde:
*[`internal/compact/tombstone_elision.go` line 121 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3552#-NwGf_cY95Ro9jTeKh4j:-NwGf_cY95Ro9jTeKh4k:brudlg5) ([raw file](https://github.com/cockroachdb/pebble/blob/9fe975dfd470e70a1989af40fa571184aee9a4b8/internal/compact/tombstone_elision.go#L121)):*
> ```Go
> }
> 
> // rangeTombstoneElider is used to check if range tombstones can be elided.
> ```

jbowens (approved):
:lgtm: 

> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3552#-:-NwGwY-I1fmouoc_6ioW:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`internal/compact/tombstone_elision.go` line 24 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3552#-NwGuMm55qeEZh5aGh7q:-NwGuMm55qeEZh5aGh7r:b6bal2j) ([raw file](https://github.com/cockroachdb/pebble/blob/9fe975dfd470e70a1989af40fa571184aee9a4b8/internal/compact/tombstone_elision.go#L24)):*
> ```Go
> // ranges can be elided.
> //
> // Note that the concept of "tombstone" appliesto range keys as well:
> ```

nit: "applies to"
___
*[`internal/compact/tombstone_elision.go` line 121 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3552#-NwGf_cY95Ro9jTeKh4j:-NwGucBP1dv6Fh4HZZmT:b2nhipw) ([raw file](https://github.com/cockroachdb/pebble/blob/9fe975dfd470e70a1989af40fa571184aee9a4b8/internal/compact/tombstone_elision.go#L121)):*


</blockquote></details>

LOL

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3552#-:-NwK_uNx2OTP4GADgRN1:b-814g8k)* status: 11 of 12 files reviewed, 1 unresolved discussion (waiting on @jbowens)

PR #3553, opened 2024-04-24
db: rework mergingIter and levelIter synthetic keys

jbowens:
The merging iterator and the level iterator are more tightly coupled than desired. Most of this coupling is a consequence of range deletions. The merging iterator is responsible for applying range deletions across levels and requires that the level iterator not progress to a new file until the file's range deletions are no longer relevant to other levels. It does this by interleaving "synthetic" keys. When these keys reach the top of the heap, it signals that all other levels have progressed to a point where the current file's range deletions are no longer relevant.

Previously, knowledge of which keys were interleaved "synthetic" keys was propagated indirectly—outside the internal iterator interface—through a pointer to a levelIterBoundaryContext struct with boolean fields. This commit instead always interleaves synthetic keys as range deletion exclusive sentinels. The merging iterator can infer which keys are synthetic by calling InternalKey.IsExclusiveSentinel. Propagating this information directly through the internal iterator interface is more direct and understandable.

Beyond needing to know which keys are synthetic, the merging iterator also must know when user-imposed iteration bounds have been reached. Ordinarily when bounds have been reached, internal iterators become exhausted. Since the levelIter's file's range deletions may still be relevant, it cannot and it interleaves a "synthetic" key at the iteration bound. When this key reaches the top of the merging iterator's heap, there are no more keys within bounds and the merging iterator is exhausted. To make this determination, we add a simple key comparison. The use of a key comparison is expected to be acceptable, because it's only performed when a synthetic key reaches the top of the heap. This additional key comparison should ultimately be eliminated when #2863 is complete.

Informs #2863.

### DIFF
--- a/db.go
+++ b/db.go
@@ -1468,7 +1468,6 @@ func (i *Iterator) constructPointIter(
 
 			li.init(ctx, i.opts, &i.comparer, i.newIters, files, level, internalOpts)
 			li.initRangeDel(&mlevels[mlevelsIndex].rangeDelIter)
-			li.initBoundaryContext(&mlevels[mlevelsIndex].levelIterBoundaryContext)
 			li.initCombinedIterState(&i.lazyCombinedIter.combinedIterState)
 			mlevels[mlevelsIndex].levelIter = li
 			mlevels[mlevelsIndex].iter = invalidating.MaybeWrapIfInvariants(li)
--- a/level_checker.go
+++ b/level_checker.go
@@ -49,7 +49,6 @@ import (
 type simpleMergingIterLevel struct {
 	iter         internalIterator
 	rangeDelIter keyspan.FragmentIterator
-	levelIterBoundaryContext
 
 	iterKV    *base.InternalKV
 	tombstone *keyspan.Span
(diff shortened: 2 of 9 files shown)


RaduBerinde (approved):
What a great idea to issue sentinel keys! This is fantastic! :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3553#-:-NwL6tKO2Ia8MyS-FGNR:bot61pt)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`level_iter.go` line 114 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3553#-NwL6Dn87FkEUQBfvZMC:-NwL6Dn87FkEUQBfvZMD:bk1znde) ([raw file](https://github.com/cockroachdb/pebble/blob/16950aa34874b14eeb1ab1e1f9ed8cc6491726a5/level_iter.go#L114)):*
> ```Go
> 	err              error
> 
> 	// Pointer into this level's mergingIterLevel.levelIterBoundaryContext.
> ```

So good to see this go away!!
___
*[`level_iter.go` line 200 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3553#-NwL6TmO0qJoAbygiwCQ:-NwL6TmO0qJoAbygiwCR:bf41fdp) ([raw file](https://github.com/cockroachdb/pebble/blob/a28f3361673804a19cc140ab1470c0570ab9fa9f/level_iter.go#L200)):*
> ```Go
> 
> func (l *levelIter) makeSyntheticBound(userKey []byte) *base.InternalKV {
> 	l.syntheticBoundary = base.InternalKV{
> ```

Could we do an `if invariants.Enabled && l.rangeDelIterPtr == nil { panic }` and explain that we only emit these keys when used under the merge iterator?

The code would be more clear if we added a `emitSyntheticBounds` bool and initialize that as `rangeDelIterPtr != nil` instead of using the latter directly.

PR #3558, opened 2024-04-26
compact: minor splitter simplifications

RaduBerinde:
#### compact: simplify splitter to use Writer.UnsafeLastPointKey

Simplify the splitter that prevents split user keys to use
`tw.UnsafeLastPointKey` directly.

#### compact: use only one limit splitter

There's no point in using two limit splitters in parallel. We can use
just one with the smaller of the two limits.


RaduBerinde:
Closing this, I have a bigger rewrite coming up.

PR #3559, opened 2024-04-26
meta: fail execution if Fatalf was used

RaduBerinde:
`Fatalf` was storing an error but we were never checking it. It will now fail the execution.


jbowens (approved):
lgtm

RaduBerinde:
TFTR!

PR #3563, opened 2024-04-29
compact: new output splitter implementation

RaduBerinde:
The current `OutputSplitter` implementations give the illusion of
arbitrary composability but there are some subtleties.  There is an
unwritten requirement that the suggested split key returned by
`OnNewOutput` is >= last key for which `ShouldSplitBefore` returned
true. This is hard to guarantee when implementations are composed. It
works today because of the particular splitter logic and
configuration.

This change replaces these with a single `OutputSplitter`
implementation which does everything. Its implementation is only a bit
more complicated than the file size splitter.

We simplify the API: we now create a separate splitter per file, and
it is illegal to call `ShouldSplitBefore` again after it returns
`SplitNow`. We also provide a `SplitKey()` method to get a suggested
split key.

The behavior is mostly the same, with some improvements on the split
key: when we split along a grandparent boundary, the split key will
match that boundary (instead of using the next key); this makes a
difference in the file boundaries when there are tombstones or range keys
intersecting that boundary.


RaduBerinde:
The macos test ran into a meta test failure. I got the ops and reproduced locally and I can repro on master. I will spend some time debugging and file a separate issue and/or PR.

jbowens (approved):
:lgtm:



:sunglasses: 

> Reviewed 5 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3563#-:-NwjvcXK6l_4DNUtgKIW:bmiobjo)* status: 5 of 8 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

___
*[`internal/compact/splitting.go` line 120 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3563#-NwjtY7cCUJ5p51vobNC:-NwjtY7cCUJ5p51vobND:br2hm47) ([raw file](https://github.com/cockroachdb/pebble/blob/2d869001fc93c682e7569313ffc20554aa725e21/internal/compact/splitting.go#L120)):*
> ```Go
> }
> 
> // NewOutputSplitter uses
> ```

nit: incomplete comment
___
*[`internal/compact/splitting.go` line 160 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3563#-NwjtsqZ-xqbdFkWW55h:-NwjtsqZ-xqbdFkWW55i:bfwttxt) ([raw file](https://github.com/cockroachdb/pebble/blob/2d869001fc93c682e7569313ffc20554aa725e21/internal/compact/splitting.go#L160)):*
> ```Go
> // the frontier advances past the current boundary.
> func (s *OutputSplitter) boundaryReached(key []byte) (nextBoundary []byte) {
> 	s.reachedBoundary = s.nextBoundary
> ```

worth an `invariants.Enabled` check that `nextBoundary` matches `key`?
___
*[`internal/compact/splitting.go` line 190 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3563#-NwjuHgm0I8KFIAnYYNe:-NwjuHgm0I8KFIAnYYNf:birg62l) ([raw file](https://github.com/cockroachdb/pebble/blob/2d869001fc93c682e7569313ffc20554aa725e21/internal/compact/splitting.go#L190)):*
> // copying each key internally. It is not performance sensitive, as it is only
> // called once we decide to split.

This comment is no longer correct, because `ShouldSplitBefore` is called once per key of a compaction, right?
___
*[`internal/compact/splitting.go` line 219 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3563#-Nwjv7FgDY_pbfNZKR9X:-Nwjv7FgDY_pbfNZKR9Y:b-c2iosd) ([raw file](https://github.com/cockroachdb/pebble/blob/2d869001fc93c682e7569313ffc20554aa725e21/internal/compact/splitting.go#L219)):*
> ```Go
> 	}
> 
> 	if s.shouldSplitBasedOnSize(estimatedFileSize, atGrandparentBoundary) == SplitNow {
> ```

should we just access `reachedBoundary.isGrandparent` rather than setting the separate `atGrandparentBoundary` bool? I think it becomes a little more obvious if we're only accessing `reachedBoundary`

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3563)* status: 5 of 8 files reviewed, 4 unresolved discussions (waiting on @jbowens)

___
*[`internal/compact/splitting.go` line 160 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3563#-NwjtsqZ-xqbdFkWW55h:-Nwk0oHJ7KJFgSn0L7K4:btyku7f) ([raw file](https://github.com/cockroachdb/pebble/blob/2d869001fc93c682e7569313ffc20554aa725e21/internal/compact/splitting.go#L160)):*

worth an `invariants.Enabled` check that `nextBoundary` matches `key`?
</blockquote></details>

Hm.. Is that the case? If the next boundary is `b` but the compaction sees keys `a` and `c`, we would get `key=c`, right? 

Which means that I'd need to loop in here in case we go over multiple boundaries in one jump. I'll see if I can show this in a test. I think the old `fileSizeSplitter` code didn't do this correctly either (in practice, it's not a big deal, as we'll reach one boundary on each subsequent advance).
___
*[`internal/compact/splitting.go` line 190 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3563#-NwjuHgm0I8KFIAnYYNe:-Nwk13qF-CQSiwwMGt7q:b-362y5h) ([raw file](https://github.com/cockroachdb/pebble/blob/2d869001fc93c682e7569313ffc20554aa725e21/internal/compact/splitting.go#L190)):*

> // copying each key internally. It is not performance sensitive, as it is only
> // called once we decide to split.

This comment is no longer correct, because `ShouldSplitBefore` is called once per key of a compaction, right?
</blockquote></details>

Ha, I was talking about `lasUserKeyFn`, I'll clarify.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3563#-:-NwlG3i9BLwuPdSGtKXZ:b-814g8k)* status: 4 of 8 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`internal/compact/splitting.go` line 160 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3563#-NwjtsqZ-xqbdFkWW55h:-NwlDEmwE2KXYl9VIsdX:b8o8lz9) ([raw file](https://github.com/cockroachdb/pebble/blob/2d869001fc93c682e7569313ffc20554aa725e21/internal/compact/splitting.go#L160)):*

Hm.. Is that the case? If the next boundary is `b` but the compaction sees keys `a` and `c`, we would get `key=c`, right? 

Which means that I'd need to loop in here in case we go over multiple boundaries in one jump. I'll see if I can show this in a test. I think the old `fileSizeSplitter` code didn't do this correctly either (in practice, it's not a big deal, as we'll reach one boundary on each subsequent advance).
</blockquote></details>

Oh, no, I was wrong. When we return the next boundary key, it will get placed in the heap and we will call the reached callback again. I added some tests and comments.
___
*[`internal/compact/splitting.go` line 219 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3563#-Nwjv7FgDY_pbfNZKR9X:-NwlEfL68dsMmyr9kA_J:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/2d869001fc93c682e7569313ffc20554aa725e21/internal/compact/splitting.go#L219)):*

should we just access `reachedBoundary.isGrandparent` rather than setting the separate `atGrandparentBoundary` bool? I think it becomes a little more obvious if we're only accessing `reachedBoundary`
</blockquote></details>

Done.

RaduBerinde:
*[`compaction.go` line 2780 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3563#-Nwp6buDCfiDpwHPFiVF:-Nwp6buDCfiDpwHPFiVG:b-o74lbs) ([raw file](https://github.com/cockroachdb/pebble/blob/3a03bff01359d23892a1ed4007e7505ceac473f3/compaction.go#L2780)):*
> ```Go
> 	// sstable or an empty key if this output is the final sstable.
> 	finishOutput := func(splitKey []byte) error {
> 		// If we haven't output any point records to the sstable (tw == nil) then the
> ```

Note - this code is redundant with the `if tw == nil { return }` check below (after the loops). In subsequent changes we will eliminate this case altogether.

jbowens (approved):
:lgtm: 

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3563#-:-NwpCheIDmBku-tmoSoZ:bg7igkv)* status: 6 of 8 files reviewed, 4 unresolved discussions

PR #3573, opened 2024-05-01
compact: pull out most of newOutput

RaduBerinde:
This commit pulls out most of the code from `newOutput`; only the code
that modifies captured variables is left.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2685,81 +2685,22 @@ func (d *DB) runCompaction(
 		if c.cancel.Load() {
 			return ErrCancelledCompaction
 		}
-		d.mu.Lock()
-		fileNum := d.mu.versions.getNextFileNum()
-		d.mu.Unlock()
-
-		ctx := context.TODO()
-		if objiotracing.Enabled {
-			ctx = objiotracing.WithLevel(ctx, c.outputLevel.level)
-			switch c.kind {
-			case compactionKindFlush:
-				ctx = objiotracing.WithReason(ctx, objiotracing.ForFlush)
-			default:
-				ctx = objiotracing.WithReason(ctx, objiotracing.ForCompaction)
-			}
-		}
-		var writeCategory vfs.DiskWriteCategory
-		switch c.kind {
-		case compactionKindFlush:
-			if d.opts.EnableSQLRowSpillMetrics {
-				// In the scenario that the Pebble engine is used for SQL row spills the data written to
-				// the memtable will correspond to spills to disk and should be categorized as such.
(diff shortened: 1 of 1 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3573#-:-NwpU1qf51F96Lsb8EtX:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`compaction.go` line 3017 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3573#-NwpThpU0Nqn8AOtKlJl:-NwpThpU0Nqn8AOtKlJm:b-qf9e0u) ([raw file](https://github.com/cockroachdb/pebble/blob/c30a8f4391daebe7c529d372b7c78d8fc4303059/compaction.go#L3017)):*
> ```Go
> 	if c.kind == compactionKindFlush {
> 		reason = "flushing"
> 		if d.opts.EnableSQLRowSpillMetrics {
> ```

Not new in your PR, but it seems like this should be applying not just to flushes but compactions too. If the SQL row-spilling engine performs compactions, I think we want those reported as `sql-row-spill`, not `pebble-compaction`

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3573#-:-NwpZs9G6RylymSpHRqC:b-814g8k)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`compaction.go` line 3017 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3573#-NwpThpU0Nqn8AOtKlJl:-NwpZcbZ9uUQGXoC1Dzo:bp594wq) ([raw file](https://github.com/cockroachdb/pebble/blob/c30a8f4391daebe7c529d372b7c78d8fc4303059/compaction.go#L3017)):*

Not new in your PR, but it seems like this should be applying not just to flushes but compactions too. If the SQL row-spilling engine performs compactions, I think we want those reported as `sql-row-spill`, not `pebble-compaction`
</blockquote></details>

Makes sense, added a commit that does this.

jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3573#-:-NwpcCEADYiXQhQdYEbI:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

PR #3574, opened 2024-05-01
compact: rework the flow of range key spans

RaduBerinde:
#### sstable: add ReadAll test util

Add a test routine to allow tests to easily read all
points/rangedels/rangekeys from an sstable.

#### compact: rework the flow of range key spans

Currently, range key spans are obtained by the caller of the compaction
iterator and then added back to the compaction iterator which acts
like a container for these. This last part doesn't belong in the
iterator at all; we replace it with a separate "range key span
compactor".

The compactor performs the coalescing of range keys within snapshot
stripes and elision of RangeKeyDel/RangeKeyUnset. It acts as a
container for the last compacted span (if any).

Instead of buffering all the spans and encoding them when we're ready
to finish the table, we encode each span as soon as we encounter a new
span (at which point we have advanced past the end of any previous
span).

The only functional change is a minor improvement when splitting based
on file size - we now take into account the size of these encoded
spans and thus we may split sooner.


RaduBerinde:
Actually, hold off on this for now. I ran into some difficulties with this approach, I will try something a bit different.

PR #3578, opened 2024-05-02
compact: minor fix to the splitter

RaduBerinde:
Make sure we never split on the first key. In this case we were
returning SplitKey equal to the first key, which is not allowed.

Note that the newly added check is not in a hot path.

This doesn't cause visible problems currently because the compaction
code initializes the output writer lazily, and a nil `Writer`
has an estimated size of 0. It becomes a problem if we initialize the
writer upfront, since an empty but initialized writer has a few bytes
of estimated size.

### DIFF
--- a/internal/compact/splitting.go
+++ b/internal/compact/splitting.go
@@ -95,6 +95,7 @@ func (s ShouldSplit) String() string {
 // ShouldSplitBefore call.
 type OutputSplitter struct {
 	cmp            base.Compare
+	startKey       []byte
 	limit          []byte
 	targetFileSize uint64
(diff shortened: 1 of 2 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3578#-:-NwvFgVd8viE39Ih7Vbq:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #3579, opened 2024-05-02
db: remove ingestTargetLevelFunc indirection

jbowens:
Previously ingest calls passed in ingestTargetLevel to ingest to facilitate some testing that needed to force an ingestion into a specific level. The relevant testing no longer exists. This commit removes the indirection.


RaduBerinde (approved):

jbowens:
TFTR!

PR #3581, opened 2024-05-02
compact: relax splitter frontier requirement

RaduBerinde:
`NewOutputSplitter` requires the `firstKey` to be at or after the
current frontier (because it relies on the frontier tripping the
boundaries before the corresponding `ShouldSplitBefore` call).

In practice, this requirement is bothersome: if there are range del or
range key spans, the current table would start at the previous split
key, which can be behind the frontier. The current code uses the next
point key but this is not ideal - we may want to split a span multiple
times in-between two very far-away points.

This change relaxes this requirement by manually checking the
boundaries the first time `ShouldSplitBefore` is called. The
compaction code is improved to provide the correct first key.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2912,14 +2912,10 @@ func (d *DB) runCompaction(
 	// progress guarantees ensure that eventually the input iterator will be
 	// exhausted and the range tombstone fragments will all be flushed.
 	for key, val := iter.First(); key != nil || iter.FirstTombstoneStart() != nil || iter.FirstRangeKeyStart() != nil; {
-		var firstKey []byte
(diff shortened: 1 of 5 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3581#-:-NwyqAEe1Vq6YInFh4X9:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #3585, opened 2024-05-03
compact: rework span handling

RaduBerinde:
#### compact: pull out code for compacting spans

This commit pulls out the pieces of code which compact range del and
range key spans in their own structures (which we can now test
independently).

#### sstable: add ReadAll test util

Add a test routine to allow tests to easily read all
points/rangedels/rangekeys from an sstable.

#### compact: add facility to split and encode spans

Add `SplitAndEncodeSpan` which can split a range del or range key span
and encode the first split.

#### compact: rework span handling

Prior to this change, the compaction iterator would produce
non-compacted range del and range key spans, the caller would add them
back to the iterator which would buffer the spans and later compact
them.

Now, the compaction iterator immediately compacts every span it
encounters and the caller retrieves the compacted span directly. The
caller maintains at most one span and encodes previous spans
on-the-fly. The compaction iterator separately remembers the last
non-compacted RANGEDEL span for its internal purposes. We always reuse
the span slices so we will not allocate for each span we encounter.

We also improve the compaction iterator to not return (and instead
skip) any RANGEDEL or range key when the span is elided after
compaction. With this improvement, we now know for certain the first
key in each output file and can simplify the compaction loop to create
the output file upfront.


jbowens (commented):
nice!

> Reviewed 7 of 7 files at r1, 3 of 3 files at r2, 4 of 4 files at r3, 1 of 10 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3585#-:-Nwzy8VGBxu2qFFu_MwK:b-7m74ox)* status: 10 of 19 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`internal/compact/iterator.go` line 453 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3585#-NwzwcB50D87586-yRsR:-NwzwcB50D87586-yRsS:boakqfw) ([raw file](https://github.com/cockroachdb/pebble/blob/13a1d5027539c012b8be897f3065f95835e08f6c/internal/compact/iterator.go#L453)):*
> ```Go
> 					// The range key span was elided entirely; don't return this key to the caller.
> 					i.saveKey()
> 					i.skipInStripe()
> ```

I'm a little confused on the control flow here; is this `skipInStripe` right? Even if `a.RANGEKEYDEL.5` gets elided, we still want to encounter `a.SET.3`
___
*[`internal/compact/spans.go` line 210 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3585#-NwzvaCWAzgLBlP1SUR_:-NwzvaCWAzgLBlP1SURa:buj78gb) ([raw file](https://github.com/cockroachdb/pebble/blob/820aa21f14b453d0ba6224f6090f24b943ee312c/internal/compact/spans.go#L210)):*
> ```Go
> 		return err
> 	}
> 	span.Start = append(span.Start[:0], upToKey...)
> ```

I think it's worth calling out in the function comment that it will re-use the memory backing `span.Start`, which may be surprising.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3585)* status: 5 of 19 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`internal/compact/iterator.go` line 453 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3585#-NwzwcB50D87586-yRsR:-Nx--kIk3dQpcLe0pkXB:bgerklk) ([raw file](https://github.com/cockroachdb/pebble/blob/13a1d5027539c012b8be897f3065f95835e08f6c/internal/compact/iterator.go#L453)):*

I'm a little confused on the control flow here; is this `skipInStripe` right? Even if `a.RANGEKEYDEL.5` gets elided, we still want to encounter `a.SET.3`
</blockquote></details>

I wasn't sure about this.. The comment says `skipInStripe skips over skippable keys in the same stripe and user key`, I thought "skippable" means we don't need them.. but looking at the code it does seem wrong. Should we do `nextInStripe()` instead? I tried your example with the current code but it still produced a.SET.3 which I find surprising:

```
define
a-b:{(#5,RANGEKEYDEL)}
a.SET.3:val
----

iter elide-tombstones=true
first
next
----
a#3,SET:val
.

iter snapshots=(2,10) elide-tombstones=true
first
next
----
a#3,SET:val
.

```
___
*[`internal/compact/spans.go` line 210 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3585#-NwzvaCWAzgLBlP1SUR_:-Nx-07u91HY72ypfa_yT:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/820aa21f14b453d0ba6224f6090f24b943ee312c/internal/compact/spans.go#L210)):*

I think it's worth calling out in the function comment that it will re-use the memory backing `span.Start`, which may be surprising.
</blockquote></details>

Done.

jbowens (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3585)* status: 5 of 19 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`internal/compact/iterator.go` line 453 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3585#-NwzwcB50D87586-yRsR:-Nx-Cycl5tCw0TjeoO2N:b-avotwq) ([raw file](https://github.com/cockroachdb/pebble/blob/13a1d5027539c012b8be897f3065f95835e08f6c/internal/compact/iterator.go#L453)):*

I wasn't sure about this.. The comment says `skipInStripe skips over skippable keys in the same stripe and user key`, I thought "skippable" means we don't need them.. but looking at the code it does seem wrong. Should we do `nextInStripe()` instead? I tried your example with the current code but it still produced a.SET.3 which I find surprising:

```
define
a-b:{(#5,RANGEKEYDEL)}
a.SET.3:val
----

iter elide-tombstones=true
first
next
----
a#3,SET:val
.

iter snapshots=(2,10) elide-tombstones=true
first
next
----
a#3,SET:val
.

```
</blockquote></details>

Ah I think this is working for two reasons.

1. The snapshot 10 actually separates the `#inf` seqnum at which the rangekeydel is interleaved from `a#3`

but even if we remove that snapshot it works because:

2. `nextInStripe` considers `IsExclusiveSentinel()` to be a new newKeyNewStripe:

```
		// Is this a new key? There are two cases:
		//
		// 1. The new key has a different user key.
		// 2. The previous key was an interleaved range deletion or range key
		//    boundary. These keys are interleaved in the same input iterator
		//    stream as point keys, but they do not obey the ordinary sequence
		//    number ordering within a user key. If the previous key was one
		//    of these keys, we consider the new key a `newStripeNewKey` to
		//    reflect that it's the beginning of a new stream of point keys.
		if i.key.IsExclusiveSentinel() || !i.cfg.Comparer.Equal(i.key.UserKey, kv.K.UserKey) {
			i.curSnapshotIdx, i.curSnapshotSeqNum = i.cfg.Snapshots.IndexAndSeqNum(kv.SeqNum())
			return newStripeNewKey
		}
```

so this code seems okay

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3585)* status: 5 of 19 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`internal/compact/iterator.go` line 453 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3585#-NwzwcB50D87586-yRsR:-Nx-by_85Ki2hHLLrAEz:br9dtcq) ([raw file](https://github.com/cockroachdb/pebble/blob/13a1d5027539c012b8be897f3065f95835e08f6c/internal/compact/iterator.go#L453)):*

Ah I think this is working for two reasons.

1. The snapshot 10 actually separates the `#inf` seqnum at which the rangekeydel is interleaved from `a#3`

but even if we remove that snapshot it works because:

2. `nextInStripe` considers `IsExclusiveSentinel()` to be a new newKeyNewStripe:

```
		// Is this a new key? There are two cases:
		//
		// 1. The new key has a different user key.
		// 2. The previous key was an interleaved range deletion or range key
		//    boundary. These keys are interleaved in the same input iterator
		//    stream as point keys, but they do not obey the ordinary sequence
		//    number ordering within a user key. If the previous key was one
		//    of these keys, we consider the new key a `newStripeNewKey` to
		//    reflect that it's the beginning of a new stream of point keys.
		if i.key.IsExclusiveSentinel() || !i.cfg.Comparer.Equal(i.key.UserKey, kv.K.UserKey) {
			i.curSnapshotIdx, i.curSnapshotSeqNum = i.cfg.Snapshots.IndexAndSeqNum(kv.SeqNum())
			return newStripeNewKey
		}
```

so this code seems okay
</blockquote></details>

Oh, nice. But then it should be equivalent and less subtle to just call `nextInStripe`?

jbowens (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3585)* status: 5 of 19 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`internal/compact/iterator.go` line 453 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3585#-NwzwcB50D87586-yRsR:-NxDK6vO0KNSDo95wg8E:b1ubcge) ([raw file](https://github.com/cockroachdb/pebble/blob/13a1d5027539c012b8be897f3065f95835e08f6c/internal/compact/iterator.go#L453)):*

Oh, nice. But then it should be equivalent and less subtle to just call `nextInStripe`?
</blockquote></details>

Makes sense

RaduBerinde:
TFTR!

PR #3586, opened 2024-05-03
db: document memTable.writerRefs subtlety

### LINKED ISSUE #3557: db: avoid loading skiplist size during concurrent apply
Currently, during the batch commit process we first attempt to reserve space in the current memtable through `memTable.prepare`.  Since we don't know the exact amount of space the batch will consume when written to the skiplist that backs the memtable, we reserve space using an upper bound. However, over time the amount of space reserved will not reflect the skiplist size. We combat this by setting the reserved value to the skiplist size. To do this safely we use the `writerRefs` count to  ensure that there are no concurrent apply operations. However, for `memtable.prepare` we increment `writerRefs` after reserving space which results in a potential race between `memtable.prepare` and `memtable.apply`.
```
func (m *memTable) prepare(batch *Batch) error {
	avail := m.availBytes()    // If there are no concurrent apply operations or ingests, writerRefs is 0
	if batch.memTableSize > uint64(avail) {
		return arenaskl.ErrArenaFull
	}
	m.reserved += uint32(batch.memTableSize)
	m.writerRef()
	return nil
}
...

func (m *memTable) apply(batch *Batch, seqNum uint64) error {
...
	var ins arenaskl.Inserter
...
	for r := batch.Reader(); ; seqNum++ {
		kind, ukey, value, ok, err := r.Next()
		if !ok {
			if err != nil {
				return err
			}
			break
		}
		ikey := base.MakeInternalKey(ukey, seqNum, kind)
		switch kind {
...
		default:
			err = ins.Add(&m.skl, ikey, value)
		}
		if err != nil {
			return err
		}
	}
...
}
...

func (m *memTable) availBytes() uint32 {
	a := m.skl.Arena()
	if m.writerRefs.Load() == 1 {
		// If there are no other concurrent apply operations, we can update the
		// reserved bytes setting to accurately reflect how many bytes of been
		// allocated vs the over-estimation present in memTableEntrySize.
		m.reserved = a.Size()
	}
	return a.Capacity() - m.reserved
}
```

jbowens:
Previously it was not obvious why memTable.availBytes() could reset the memtable's reserved size when memTable.writerRefs() == 1. In #3557, a code reading lead us to believe there was a race where competing writers could improperly double reserve the tail of the memtable's arena. However, in the code reading we missed that the mutable memtable always has a writerRefs()≥1, which is why a writerRefs()==1 indicates there are no concurrent applications to the memtable.

Additionally, add a unit test that exercises the race of two batches competing for the tail of the memtable arena.

Close #3557.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -393,6 +393,24 @@ func writeRangeKeys(b io.Writer, iter *Iterator) {
 	}
 }
 
+func parseValue(s string) []byte {
+	if strings.HasPrefix(s, "<rand-bytes=") {
+		s = strings.TrimPrefix(s, "<rand-bytes=")
+		s = strings.TrimSuffix(s, ">")
+		n, err := strconv.Atoi(s)
+		if err != nil {
+			panic(err)
+		}
+		b := make([]byte, n)
+		rnd := rand.New(rand.NewSource(int64(n)))
+		if _, err := rnd.Read(b); err != nil {
+			panic(err)
+		}
+		return b
+	}
+	return []byte(s)
+}
+
 func runBatchDefineCmd(d *datadriven.TestData, b *Batch) error {
 	for _, line := range strings.Split(d.Input, "\n") {
 		parts := strings.Fields(line)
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3586#-:-Nx-4g_G0Bie-zxuf9js:bnfp4nl)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @itsbilal)

jbowens:
TFTR!

PR #3588, opened 2024-05-06
internal/keyspan: prohibit NextPrefix at span boundary

jbowens:
Previously when the interleaving iterator interleaved a span boundary, the semantics of calling NextPrefix were undefined. Ordinarily NextPrefix guarantees that it advances to a key with a new prefix. The existing interleaving iterator implementation didn't guarantee this when positioned on a span boundary. Guaranteeing it would require otherwise unnecessary key comparisons.

In existing use cases (for range keys, which always begin and end at prefix keys), we never call NextPrefix while positioned at a span boundary. With the planned refactor of #2863, we'll begin to use the interleaving iterator to interleave range deletions which may have bounds within the keys of a prefix. This assertion will ensure we don't unintentionally rely on currently undefined behavior of calling NextPrefix on a span boundary.

### DIFF
--- a/internal/keyspan/interleaving_iter.go
+++ b/internal/keyspan/interleaving_iter.go
@@ -479,6 +479,8 @@ func (i *InterleavingIter) Next() *base.InternalKV {
 }
 
 // NextPrefix implements (base.InternalIterator).NextPrefix.
+//
+// Calling NextPrefix while positioned at a span boundary is prohibited.
 func (i *InterleavingIter) NextPrefix(succKey []byte) *base.InternalKV {
 	if i.dir == -1 {
 		panic("pebble: cannot switch directions with NextPrefix")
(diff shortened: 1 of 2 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3588)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`internal/keyspan/interleaving_iter.go` line 504 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3588#-NxDzXekCFFVDHXL2ckq:-NxDzXekCFFVDHXL2ckr:byk06ql) ([raw file](https://github.com/cockroachdb/pebble/blob/6ee9ad0d6d0a46404124e15edb41d948ef23cc98/internal/keyspan/interleaving_iter.go#L504)):*
> ```Go
> 		}
> 	case posKeyspanStart, posKeyspanEnd:
> 		panic(errors.AssertionFailedf("NextPrefix called while positioned on a span boundary"))
> ```

[nit] I think we should do `i.err = base.AssertionFailedf()` so we still panic in invariants mode but not in production

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3588#-:-NxDzgeS3Kitilu2JW-1:bnfp4nl)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

jbowens:
TFTR!

RaduBerinde:
We hit this panic in TestIteratorErrors: https://github.com/cockroachdb/pebble/issues/3592#issuecomment-2096646327

PR #3591, opened 2024-05-06
compaction: clean up runCompaction

RaduBerinde:
#### compaction: tolerate empty keys

The compaction iterator will sometimes convert empty keys with nil.
Apparently, we want to allow empty keys (but not nil keys).

We make a small change to the compaction iterator to avoid this corner
case and we update `base.MinUserKey` to treat empty non-nil keys as
regular keys.

#### compaction: remove redundant cleanup code

The main compaction code deletes all created tables in case of error.
The calling code also deletes them by declaring them obsolete; this is
unnecessary. This change makes the copy compaction code also delete
the object on error, and removes the outer handling code (along with
all `pendingOutputs` returns).

#### compact: move grandparent and L0 limits code

This change moves the code to limit grandparent overlap and to enforce
L0 split points. We create a `Runner` into which we will incorporate
more of the code that runs the compaction.

#### compaction: clean up runCompaction

In this change we separate the code that deals with the "data" portion
of the compaction from the code that deals with the "metadata"
portion. These were intertwined in the old code, leading to a lot of
clutter.

We move the "inner" loop of the compaction (writing one output table)
to `compact.Runner`. We then use the tables in `compact.Result` to
create the version edit.


jbowens (approved):
:lgtm:

> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, 3 of 3 files at r3, 8 of 8 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3591#-:-NxJgnn89eYiZDEOgHnp:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/compact/run.go` line 90 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3591#-NxJgCpd9yVZtdjt7Gb4:-NxJgCpe451Ni1tounij:b-uivn8) ([raw file](https://github.com/cockroachdb/pebble/blob/8b8eac80c60d0b3b627c1f63f6beee9b71a0105e/internal/compact/run.go#L90)):*
> ```Go
> //
> //	r := NewRunner(cfg, iter)
> //	r.Start()
> ```

The `Start()` seems a little subtle. is it possible to fold into `WriteTable`?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3591#-:-NxK15de5lImAcDNUA23:b-814g8k)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`internal/compact/run.go` line 90 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3591#-NxJgCpd9yVZtdjt7Gb4:-NxK-m2c-iG0y7d7rFFv:bo9s7ye) ([raw file](https://github.com/cockroachdb/pebble/blob/8b8eac80c60d0b3b627c1f63f6beee9b71a0105e/internal/compact/run.go#L90)):*

The `Start()` seems a little subtle. is it possible to fold into `WriteTable`?
</blockquote></details>

I folded it into `NewRunner`. Also changed `Done()` to `MoreDataToWrite()`

PR #3593, opened 2024-05-06
db: fix iterator error propagation, NextPrefix violation

### LINKED ISSUE #3592: TestIteratorErrors fails with SeekLT violates upper bound panic
`TestIteratorErrors` runs into
```
panic: levelIter SeekLT to key "yaxhijzhbl" violates upper bound "cgavvrurmtfg" [recovered]
```

Can reproduce at 0bc137ad1edc1608cee12dec20d4b147636b72eb with
```
go test -tags invariants -run TestIteratorErrors -v -exec stress -p 4
```

https://github.com/cockroachdb/pebble/actions/runs/8973148304/job/24642661778?pr=3590

jbowens:
This commit fixes a few issues surrounding NextPrefix and iterator errors. Previously, under some circumstances, an error propagated by the underlying iterator would not immediately be propagated up. Instead the Iterator could in some circumstances call Next or NextPrefix on an already exhausted, errored iterator, resulting in undefined behavior.

Similarly, the results of calling NextPrefix on an exhausted top-level pebble.Iterator are not defined. This commit adapts the implementation to early exit in this case. The metamorphic test may generate such operations.

Fix #3592.

### DIFF
--- a/flushable.go
+++ b/flushable.go
@@ -237,10 +237,10 @@ func (s *ingestedFlushable) constructRangeDelIter(
 	file *manifest.FileMetadata, _ keyspan.SpanIterOptions,
 ) (keyspan.FragmentIterator, error) {
 	iters, err := s.newIters(context.Background(), file, nil, internalIterOpts{}, iterRangeDeletions)
-	// Note that the keyspan level iter expects a non-nil iterator to be
-	// returned even if there is an error. So, we return iters.RangeDeletion()
-	// regardless of the value of err.
-	return iters.RangeDeletion(), err
+	if err != nil {
+		return nil, err
+	}
+	return iters.RangeDeletion(), nil
 }
 
 // newRangeDelIter is part of the flushable interface.
--- a/internal/keyspan/keyspanimpl/level_iter.go
+++ b/internal/keyspan/keyspanimpl/level_iter.go
@@ -132,15 +132,13 @@ const (
 )
 
 func (l *LevelIter) loadFile(file *manifest.FileMetadata, dir int) loadFileReturnIndicator {
-	indicator := noFileLoaded
 	if l.iterFile == file {
 		if l.err != nil {
 			return noFileLoaded
 		}
 		if l.iter != nil {
-			// We are already at the file, but we would need to check for bounds.
-			// Set indicator accordingly.
-			indicator = fileAlreadyLoaded
+			// We are already at the file.
+			return fileAlreadyLoaded
 		}
 		// We were already at file, but don't have an iterator, probably because the file was
 		// beyond the iteration bounds. It may still be, but it is also possible that the bounds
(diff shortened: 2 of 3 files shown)


RaduBerinde (approved):
:lgtm:

Just one thing. this does not fix the original `violates upper bound` issue in #3592.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3593#-:-NxEPVVI07YSUYJyuVN9:bwqr93n)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @aadityasondhi)

jbowens (commented):
Pushed a second commit to fix the `violates upper bound` issue :sweat_smile: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3593#-:-NxJ9SBW1XYlKXyTUiex:b-crc2sg)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @aadityasondhi)

RaduBerinde (approved):
Nice!! :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3593#-:-NxJV46c5_IeHimXPvly:busy3rq)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @jbowens)

___
*[`internal/keyspan/keyspanimpl/level_iter.go` line 158 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3593#-NxJUqVc2nk2LseJ3gph:-NxJUqVc2nk2LseJ3gpi:b-3oskzv) ([raw file](https://github.com/cockroachdb/pebble/blob/7498751c94c6610a620e26d067463f1a1c503d1f/internal/keyspan/keyspanimpl/level_iter.go#L158)):*
> ```Go
> 		return noFileLoaded
> 	}
> 	l.iter, l.err = l.newIter(file, l.tableOpts)
> ```

[nit] maybe do `iter, err :=` and only set `l.iter` in the no error case, so we don't rely on `newIter` returning `nil`

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3593)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @RaduBerinde)

___
*[`internal/keyspan/keyspanimpl/level_iter.go` line 158 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3593#-NxJUqVc2nk2LseJ3gph:-NxJeUWf2BcIB6nAfgLk:b-ulcfqc) ([raw file](https://github.com/cockroachdb/pebble/blob/7498751c94c6610a620e26d067463f1a1c503d1f/internal/keyspan/keyspanimpl/level_iter.go#L158)):*

[nit] maybe do `iter, err :=` and only set `l.iter` in the no error case, so we don't rely on `newIter` returning `nil`
</blockquote></details>

Good call

RaduBerinde:
*[`iterator.go` line 1743 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3593#-NxQ-3pa174C2_I7gKg4:-NxQ-3pa174C2_I7gKg5:b-8yosdm) ([raw file](https://github.com/cockroachdb/pebble/blob/78953f4efe06133a504f38a14ce27e81000738be/iterator.go#L1743)):*
> ```Go
> 		i.rangeKey.updated = i.rangeKey.hasRangeKey && !i.Valid() && i.opts.rangeKeys()
> 	}
> 	// NextPrefix from an exhausted position is undefined. We keep the exhausted
> ```

`Iterator.NextPrefix` says you can't call it in `IterAtLimit` state but doesn't say anything about an exhausted iterator. If it's exhausted in the backward direction, isn't `NextPrefix` valid?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3593)* status: 0 of 3 files reviewed, 2 unresolved discussions

___
*[`iterator.go` line 1743 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3593#-NxQ-3pa174C2_I7gKg4:-NxSiyx1176XdN3igbT3:bryzu1f) ([raw file](https://github.com/cockroachdb/pebble/blob/78953f4efe06133a504f38a14ce27e81000738be/iterator.go#L1743)):*

`Iterator.NextPrefix` says you can't call it in `IterAtLimit` state but doesn't say anything about an exhausted iterator. If it's exhausted in the backward direction, isn't `NextPrefix` valid?
</blockquote></details>

Hrm yeah it seems like we could allow it as an operation semantically equivalent to a plain Next, but I also don't think we require it. Maybe this implementation should literally call Iterator.Next if exhausted.

PR #3596, opened 2024-05-07
db: interleave range deletion boundaries in levelIter 

jbowens:
Previously, levelIter would sometimes interleave fake keys at file boundaries
or user-provided iteration boundaries when a file contains range deletions.
This commit reworks the levelIter to instead interleave the individual bounds
of the range deletions themselves, using an interleaving iterator. This
simplifies the levelIter. Because range deletions are now read in two places:
once for interleaving bounds and once to expose a rangeDelIter to the
mergingIter, this commit requires opening a file's range deletion iterator
twice. This is an intermediary state until https://github.com/cockroachdb/pebble/issues/2863 is completed when the
mergingIter will consult the levelIter's interleaving iterator to determine
which range deletions overlap the current iterator position.

Informs https://github.com/cockroachdb/pebble/issues/2863.

### DIFF
--- a/internal/keyspan/interleaving_iter.go
+++ b/internal/keyspan/interleaving_iter.go
@@ -465,12 +465,12 @@ func (i *InterleavingIter) Next() *base.InternalKV {
 			// Since we're positioned on a Span, the pointIter is positioned
 			// entirely behind the current iterator position. Reposition it
 			// ahead of the current iterator position.
-			i.savePoint(i.pointIter.Next())
+			i.switchPointIteratorIntoForward()
 		case posKeyspanEnd:
 			// Since we're positioned on a Span, the pointIter is positioned
 			// entirely behind of the current iterator position. Reposition it
 			// ahead the current iterator position.
-			i.savePoint(i.pointIter.Next())
+			i.switchPointIteratorIntoForward()
 		}
 		// Fallthrough to calling i.nextPos.
 	}
@@ -543,7 +543,7 @@ func (i *InterleavingIter) Prev() *base.InternalKV {
 			// Since we're positioned on a Span, the pointIter is positioned
 			// entirely ahead of the current iterator position. Reposition it
 			// behind the current iterator position.
-			i.savePoint(i.pointIter.Prev())
+			i.switchPointIteratorIntoReverse()
 			// Without considering truncation of spans to seek keys, the keyspan
 			// iterator is already in the right place. But consider span [a, z)
 			// and this sequence of iterator calls:
@@ -565,7 +565,7 @@ func (i *InterleavingIter) Prev() *base.InternalKV {
 			// Since we're positioned on a Span, the pointIter is positioned
 			// entirely ahead of the current iterator position. Reposition it
 			// behind the current iterator position.
-			i.savePoint(i.pointIter.Prev())
+			i.switchPointIteratorIntoReverse()
 		}
 
 		if i.spanMarkerTruncated {
@@ -637,7 +637,7 @@ func (i *InterleavingIter) nextPos() {
 
 	switch i.pos {
 	case posExhausted:
-		i.savePoint(i.pointIter.Next())
+		i.switchPointIteratorIntoForward()
 		i.saveSpanForward(i.keyspanIter.Next())
 		i.savedKeyspan()
 		i.computeSmallestPos()
@@ -709,7 +709,7 @@ func (i *InterleavingIter) prevPos() {
 
 	switch i.pos {
 	case posExhausted:
-		i.savePoint(i.pointIter.Prev())
+		i.switchPointIteratorIntoReverse()
 		i.saveSpanBackward(i.keyspanIter.Prev())
 		i.savedKeyspan()
 		i.computeLargestPos()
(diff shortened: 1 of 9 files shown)


RaduBerinde (approved):
:lgtm:

This is very cool, I love it! So much cleaner when you're thinking about what keys levelIter produces

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3596#-:-NxL-wnE9g32jkvZGXDt:b-6pzq0d)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`merging_iter.go` line 1202 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3596#-NxKzhBLCckTVsOtuDxe:-NxKzhBM6IMSjURqYbDI:buizsrf) ([raw file](https://github.com/cockroachdb/pebble/blob/d97a0689216b3892f321968871907efc6f4157e7/merging_iter.go#L1202)):*
> ```Go
> 	// The heap root necessarily must be positioned at a key < succKey, because
> 	// NextPrefix was invoked.
> 	root := &m.heap.items[0]
> ```

I think it'd be cleaner to just do `root := m.heap.items[0]` here and again first thing inside the loop.
___
*[`merging_iter.go` line 1217 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3596#-NxKxz7B-lCDXAepsiN2:-NxKxz7B-lCDXAepsiN3:b-uskgm5) ([raw file](https://github.com/cockroachdb/pebble/blob/d97a0689216b3892f321968871907efc6f4157e7/merging_iter.go#L1217)):*
> ```Go
> 	// prefix if we did not land on a range deletion's boundary. Range deletions
> 	// may have bounds within the prefix that are still surfaced by NextPrefix.
> 	m.levelsPositioned[prevRoot.index] = prevRoot.iterKV == nil || !prevRoot.iterKV.K.IsExclusiveSentinel()
> ```

Do we care about `levelPositioned` once we hit an error? If no, can't we just set it before `nextEntry`?
___
*[`merging_iter.go` line 1250 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3596#-NxKz0KHCD8GQP3VgE-c:-NxKz0KHCD8GQP3VgE-d:b6l4aqv) ([raw file](https://github.com/cockroachdb/pebble/blob/d97a0689216b3892f321968871907efc6f4157e7/merging_iter.go#L1250)):*
> ```Go
> 		// deletions may have bounds within the prefix that are still surfaced
> 		// by NextPrefix.
> 		m.levelsPositioned[prevRoot.index] = prevRoot.iterKV == nil || !prevRoot.iterKV.K.IsExclusiveSentinel()
> ```

We handled the `root.IterKV.K.IsExclusiveSentinel()` case above, so isn't this always true here?

jbowens (commented):
Nice, yeah, agreed it's more obvious what's going on

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3596#-:-NxOM1RJ5Drke_SMrjbl:b-9n22fl)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`merging_iter.go` line 1202 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3596#-NxKzhBLCckTVsOtuDxe:-NxOI_iT3_aqFZn_g0Rf:bohqzsr) ([raw file](https://github.com/cockroachdb/pebble/blob/d97a0689216b3892f321968871907efc6f4157e7/merging_iter.go#L1202)):*

I think it'd be cleaner to just do `root := m.heap.items[0]` here and again first thing inside the loop.
</blockquote></details>

Agreed, done.
___
*[`merging_iter.go` line 1217 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3596#-NxKxz7B-lCDXAepsiN2:-NxOIe4n5vvHP8Hwk3dh:b-neodsj) ([raw file](https://github.com/cockroachdb/pebble/blob/d97a0689216b3892f321968871907efc6f4157e7/merging_iter.go#L1217)):*

Do we care about `levelPositioned` once we hit an error? If no, can't we just set it before `nextEntry`?
</blockquote></details>

we don't, but we care about the `iterKV` _after_ the call to nextEntry.
___
*[`merging_iter.go` line 1250 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3596#-NxKz0KHCD8GQP3VgE-c:-NxOLtJb0K0lK_AYPSxu:bs60zbd) ([raw file](https://github.com/cockroachdb/pebble/blob/d97a0689216b3892f321968871907efc6f4157e7/merging_iter.go#L1250)):*

We handled the `root.IterKV.K.IsExclusiveSentinel()` case above, so isn't this always true here?
</blockquote></details>

same here, the call to `nextEntry` will update `iterKV`, and we want to test _that_ KV's key

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3596#-:-NxOP9-pCi7Ha6HkNTeA:bnfp4nl)* status: 0 of 9 files reviewed, all discussions resolved (waiting on @itsbilal)

jbowens:
TFTR!

PR #3600, opened 2024-05-09
db: rework merging iterator range deletion handling

### LINKED ISSUE #2863: db: consider reworking rangedel, mergingIter and levelIter interactions

jbowens:
Rework the merging iterator's handling of range deletions to use the
interleaved range deletion boundary keys to determine when the iterator is
positioned within a level's range deletion. This removes the direct
manipulation of a range deletion keyspan.FragmentIterator from the mergingIter,
delegating that to the child iterator's keyspan.InterleavingIter.

This factoring is a little cleaner and decouples the mergingIter from the
details of range deletion iterators, and in particular, the levelIter's
individual sstables. It also should reduce key comparisons, especially during
scans, by avoiding unnecessary key comparisons with range deletions that are
loaded but outside the keyspace being iterated over.

Close #2863.


jbowens (commented):
metamorphic test failure 30 minutes in :(

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3600#-:-NxYm20H4vsxd9aVMEHM:b-uluwin)* status: 0 of 24 files reviewed, all discussions resolved (waiting on @aadityasondhi, @RaduBerinde, and @sumeerbhola)

xxmplus:
Close to due to age

PR #3602, opened 2024-05-09
db: avoid acquiring DB.mu during batch commit, unless rotating memtable

### LINKED ISSUE #2680: db: avoid acquiring DB.mu in the commit pipeline
Currently during batch `prepare`, a batch acquires the commit mutex. It then allocates sequence numbers for its keys. Then, in `commitWrite` it locks the global database mutex in order to grab the current mutable memtable:

```
	d.mu.Lock()

	var err error
	if !b.ingestedSSTBatch {
		// Batches which contain keys of kind InternalKeyKindIngestSST will
		// never be applied to the memtable, so we don't need to make room for
		// write. For the other cases, switch out the memtable if there was not
		// enough room to store the batch.
		err = d.makeRoomForWrite(b)
	}

	if err == nil && !d.opts.DisableWAL {
		d.mu.log.bytesIn += uint64(len(repr))
	}

	// Grab a reference to the memtable while holding DB.mu. Note that for
	// non-flushable batches (b.flushable == nil) makeRoomForWrite() added a
	// reference to the memtable which will prevent it from being flushed until
	// we unreference it. This reference is dropped in DB.commitApply().
	mem := d.mu.mem.mutable

	d.mu.Unlock()
```

This can contribute to high tail latencies (#2646). All batch commits become stalled, waiting for the goroutine holding `DB.mu` to complete.

We could avoid this mutex acquisition in the common case where there's room in the existing memtable for the current batch. The committing goroutine can load the mutable memtable through an atomic load. Memtable reservation can also be performed through lock-free operations. If reservation finds that there is not sufficient space in the memtable, it falls back into acquiring `d.mu` and entering `makeRoomForWrite`. There are instances where we rotate the memtable before it's full. In these instances, `makeRoomForWrite` can set the mutable memtable to `nil`, reflecting that a rotation is in progress and incoming writers must queue in `makeRoomForWrite`. A version of this was hacked together in ef6e59b4d7b4d1006e441945415abdaaa817aac7.

jbowens:
**db: move log bytes-in to an atomic**

Move the log bytes-in metric out of the protection of db.mu and into an atomic.

**db: avoid acquiring DB.mu during batch commit, unless rotating memtable**

Previously, DB.commitWrite would acquire the DB.mu before calling
makeRoomForWrite to ensure the current memtable had sufficient space for the
batch's contents. When DB.commitWrite is called, the caller already holds the
commit pipeline's mutex providing mutual exclusion with other committers and,
crucially, preventing the concurrent rotation of the memtable.

This commit adjusts DB.commitWrite to attempt reserving space in the current
mutable memtable before acquiring DB.mu. In the common case, if successful,
there is no need to acquire DB.mu. If unsuccessful, the committer falls back to
acuqiring DB.mu and rotating the memtable through calling makeRoomForWrite.

Close #2680.

### DIFF
--- a/db.go
+++ b/db.go
@@ -271,6 +271,10 @@ type DB struct {
 
 	// The logical size of the current WAL.
 	logSize atomic.Uint64
+	// The number of input bytes to the log. This is the raw size of the
+	// batches written to the WAL, without the overhead of the record
+	// envelopes.
+	logBytesIn atomic.Uint64
 
 	// The number of bytes available on disk.
 	diskAvailBytes atomic.Uint64
@@ -371,10 +375,6 @@ type DB struct {
 			// manager is not protected by mu, but calls to Create must be
 			// serialized, and happen after the previous writer is closed.
 			manager wal.Manager
-			// The number of input bytes to the log. This is the raw size of the
-			// batches written to the WAL, without the overhead of the record
-			// envelopes. Requires DB.mu to be held when read or written.
-			bytesIn uint64
 			// The Writer is protected by commitPipeline.mu. This allows log writes
 			// to be performed without holding DB.mu, but requires both
 			// commitPipeline.mu and DB.mu to be held when rotating the WAL/memtable
@@ -941,10 +941,6 @@ func (d *DB) commitWrite(b *Batch, syncWG *sync.WaitGroup, syncErr *error) (*mem
 		err = d.makeRoomForWrite(b)
 	}
 
-	if err == nil && !d.opts.DisableWAL {
-		d.mu.log.bytesIn += uint64(len(repr))
-	}
-
 	// Grab a reference to the memtable while holding DB.mu. Note that for
 	// non-flushable batches (b.flushable == nil) makeRoomForWrite() added a
 	// reference to the memtable which will prevent it from being flushed until
@@ -955,10 +951,10 @@ func (d *DB) commitWrite(b *Batch, syncWG *sync.WaitGroup, syncErr *error) (*mem
 	if err != nil {
 		return nil, err
 	}
-
 	if d.opts.DisableWAL {
 		return mem, nil
 	}
+	d.logBytesIn.Add(uint64(len(repr)))
 
 	if b.flushable == nil {
 		size, err = d.mu.log.writer.WriteRecord(repr, wal.SyncOptions{Done: syncWG, Err: syncErr}, b)
@@ -1959,7 +1955,7 @@ func (d *DB) Metrics() *Metrics {
 	// d.atomic.logSize has exceeded that physical size. We allow for this
 	// anomaly.
 	metrics.WAL.PhysicalSize = walStats.LiveFileSize
-	metrics.WAL.BytesIn = d.mu.log.bytesIn // protected by d.mu
+	metrics.WAL.BytesIn = d.logBytesIn.Load()
 	metrics.WAL.Size = d.logSize.Load()
 	for i, n := 0, len(d.mu.mem.queue)-1; i < n; i++ {
 		metrics.WAL.Size += d.mu.mem.queue[i].logSize
--- a/db.go
+++ b/db.go
@@ -959,10 +959,6 @@ func (d *DB) commitWrite(b *Batch, syncWG *sync.WaitGroup, syncErr *error) (*mem
 			func() {
 				d.mu.Lock()
 				defer d.mu.Unlock()
-				// TODO(jackson): makeRoomForWrite will try mutable.prepare
-				// again unnecessarily. We can refactor makeRoomForWrite to make
-				// the assumption that the initial memtable is full, and only
-				// attempt prepare on the rotated memtable.
 				err = d.makeRoomForWrite(b)
 				mem = d.mu.mem.mutable
 			}()
@@ -2341,13 +2337,24 @@ func (d *DB) walPreallocateSize() int {
 	return int(size)
 }
 
-func (d *DB) newMemTable(logNum base.DiskFileNum, logSeqNum uint64) (*memTable, *flushableEntry) {
+func (d *DB) newMemTable(
+	logNum base.DiskFileNum, logSeqNum, minSize uint64,
+) (*memTable, *flushableEntry) {
+	targetSize := minSize + uint64(memTableEmptySize)
+	// The targetSize should be less than MemTableSize, because any batch >=
+	// MemTableSize/2 should be treated as a large flushable batch.
+	if targetSize > d.opts.MemTableSize {
+		panic(errors.AssertionFailedf("attempting to allocate memtable larger than MemTableSize"))
+	}
+	// Double until the next memtable size is at least large enough to fit
+	// minSize.
+	for d.mu.mem.nextSize < targetSize {
+		d.mu.mem.nextSize = min(2*d.mu.mem.nextSize, d.opts.MemTableSize)
+	}
 	size := d.mu.mem.nextSize
+	// The next memtable should be double the size, up to Options.MemTableSize.
 	if d.mu.mem.nextSize < d.opts.MemTableSize {
-		d.mu.mem.nextSize *= 2
-		if d.mu.mem.nextSize > d.opts.MemTableSize {
-			d.mu.mem.nextSize = d.opts.MemTableSize
-		}
+		d.mu.mem.nextSize = min(2*d.mu.mem.nextSize, d.opts.MemTableSize)
 	}
 
 	memtblOpts := memTableOptions{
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3602#-:-NxYn_Bn-b3MoqmWEPj5:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi, @jbowens, and @sumeerbhola)

___
*[`db.go` line 943 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3602#-NxYnAtiCLwtzWoBX_7U:-NxYnAtiCLwtzWoBX_7V:b-7xycta) ([raw file](https://github.com/cockroachdb/pebble/blob/b7cb5a46f47c44d60d63493af535889786dc89d7/db.go#L943)):*
> ```Go
> 	var err error
> 	// Grab a reference to the memtable. We don't hold DB.mu, but that's okay
> 	// for readers of d.mu.mem.mutable.
> ```

[nit] it's ok because we hold `commitPipeline.mu`, correct?

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3602)* status: 0 of 1 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi and @jbowens)

___
*[`db.go` line 965 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3602#-Nxm20XG0mz6pL1-2TPj:-Nxm20XG0mz6pL1-2TPk:b3fptej) ([raw file](https://github.com/cockroachdb/pebble/blob/3795cf7fb285ddc2aaab2ba798752077d4d3bf05/db.go#L965)):*
> ```Go
> 				// again unnecessarily. We can refactor makeRoomForWrite to make
> 				// the assumption that the initial memtable is full, and only
> 				// attempt prepare on the rotated memtable.
> ```

Why not do this now? Are we trying to backport this?
___
*[`db.go` line 2444 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3602#-Nxm2R9C8JtEb7tg6rEm:-Nxm2R9C8JtEb7tg6rEn:b-jbqn2h) ([raw file](https://github.com/cockroachdb/pebble/blob/3795cf7fb285ddc2aaab2ba798752077d4d3bf05/db.go#L2444)):*
> ```Go
> 	}
> 
> 	force := b == nil || b.flushable != nil
> ```

Looks like we call this either with b == nil, which is a forced rotation, or when the memtable is already full, which is also a rotation. So this could be easily simplified as you mentioned earlier.
___
*[`db.go` line 2451 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3602#-Nxm2r5Y51L3afw0kMOC:-Nxm2r5Y51L3afw0kMOD:b-jy3xi7) ([raw file](https://github.com/cockroachdb/pebble/blob/3795cf7fb285ddc2aaab2ba798752077d4d3bf05/db.go#L2451)):*
> ```Go
> 			if err != arenaskl.ErrArenaFull {
> 				if stalled {
> 					d.opts.EventListener.WriteStallEnd()
> ```

I think we should document that `WriteStallBegin/WriteStallEnd` are always paired in the context of a single call to `makeRoomForWrite` in that if it began a write stall, it will end it before returning.

I have initially confused that we had lost this call to `WriteStallEnd`, when lifting out the prepare call. 
___
*[`db.go` line 2455 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3602#-Nxm2hYA58fzwQMizOfR:-Nxm2hYA58fzwQMizOfS:b-xxl99o) ([raw file](https://github.com/cockroachdb/pebble/blob/3795cf7fb285ddc2aaab2ba798752077d4d3bf05/db.go#L2455)):*
> ```Go
> 				return err
> 			}
> 		} else if !force {
> ```

!force is equivalent to b != nil && b.flushable == nil, so isn't this the same condition as the if-block?

sumeerbhola (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3602#-:-Nxm3jmFBEemI0MmYam3:bg7igkv)* status: 0 of 1 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi and @jbowens)

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3602)* status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi and @sumeerbhola)

___
*[`db.go` line 965 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3602#-Nxm20XG0mz6pL1-2TPj:-NxnAmLoEt7Kji3QKWNZ:b-7ev3mj) ([raw file](https://github.com/cockroachdb/pebble/blob/7d63c4c495d56d572e0924d18c0e310bd22010b1/db.go#L965)):*

Why not do this now? Are we trying to backport this?
</blockquote></details>

Not trying to backport, just figured it might be a bit of a delicate refactor. I pushed an additional commit that attempts it.
___
*[`db.go` line 2444 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3602#-Nxm2R9C8JtEb7tg6rEm:-NxnAxwi1Nd4pwrIUj_0:b-1b4m6p) ([raw file](https://github.com/cockroachdb/pebble/blob/7d63c4c495d56d572e0924d18c0e310bd22010b1/db.go#L2444)):*

Looks like we call this either with b == nil, which is a forced rotation, or when the memtable is already full, which is also a rotation. So this could be easily simplified as you mentioned earlier.
</blockquote></details>

addressed in the new commit
___
*[`db.go` line 2451 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3602#-Nxm2r5Y51L3afw0kMOC:-NxnB-J7A41eJs7c_fIl:b-1b4m6p) ([raw file](https://github.com/cockroachdb/pebble/blob/7d63c4c495d56d572e0924d18c0e310bd22010b1/db.go#L2451)):*

I think we should document that `WriteStallBegin/WriteStallEnd` are always paired in the context of a single call to `makeRoomForWrite` in that if it began a write stall, it will end it before returning.

I have initially confused that we had lost this call to `WriteStallEnd`, when lifting out the prepare call.
</blockquote></details>

addressed in the new commit
___
*[`db.go` line 2455 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3602#-Nxm2hYA58fzwQMizOfR:-NxnAzrp3YHHScri_EFK:b-1b4m6p) ([raw file](https://github.com/cockroachdb/pebble/blob/7d63c4c495d56d572e0924d18c0e310bd22010b1/db.go#L2455)):*

!force is equivalent to b != nil && b.flushable == nil, so isn't this the same condition as the if-block?
</blockquote></details>

addressed in the new commit

sumeerbhola (approved):
:lgtm: 

> Reviewed 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3602#-:-NxnWKdv4H4wY21x0_zI:bg7igkv)* status: all files reviewed, 9 unresolved discussions (waiting on @aadityasondhi and @jbowens)

___
*[`db.go` line 2450 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3602#-NxnUsfCDS7nP2vVv_uu:-NxnUsfCDS7nP2vVv_uv:bn02trv) ([raw file](https://github.com/cockroachdb/pebble/blob/c10cd65f353a09f1f98193b0897d0fcdfd657b58/db.go#L2450)):*
> ```Go
> //     increase write pressure.
> //
> // maybeInduceWriteStall checks these stall conditions, and if present, wait for
> ```

nit: waits for ...
___
*[`db.go` line 2454 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3602#-NxnUwdb8eyVmyGxCrCG:-NxnUwdb8eyVmyGxCrCH:b-pqrorj) ([raw file](https://github.com/cockroachdb/pebble/blob/c10cd65f353a09f1f98193b0897d0fcdfd657b58/db.go#L2454)):*
> ```Go
> func (d *DB) maybeInduceWriteStall(b *Batch) {
> 	stalled := false
> 	// This function will call EventListener.WriteStallBegin at most once.  If
> ```

nit: It it does ...
___
*[`db.go` line 2456 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3602#-NxnV6Qa-v6NvIaziuvH:-NxnV6Qa-v6NvIaziuvI:b-bmk3ir) ([raw file](https://github.com/cockroachdb/pebble/blob/c10cd65f353a09f1f98193b0897d0fcdfd657b58/db.go#L2456)):*
> ```Go
> 	// This function will call EventListener.WriteStallBegin at most once.  If
> 	// does call it, it will EventListener.WriteStallEnd once before returning.
> 	for {
> ```

This is a nice simplification -- lifting the for loop out of `makeRoomForWrite`.
___
*[`db.go` line 2589 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3602#-NxnVjKA7vkb2njmQ70g:-NxnVjKA7vkb2njmQ70h:bb7lbs2) ([raw file](https://github.com/cockroachdb/pebble/blob/c10cd65f353a09f1f98193b0897d0fcdfd657b58/db.go#L2589)):*
> ```Go
> 		// memtable rotation allocates a memtable sufficiently large. We also
> 		// held d.commit.mu for the entirety of this function, ensuring that no
> 		// other committers may have reserved memory in the new memtable yet.
> ```

nice

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3602)* status: 0 of 4 files reviewed, 7 unresolved discussions (waiting on @aadityasondhi and @sumeerbhola)

___
*[`db.go` line 2450 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3602#-NxnUsfCDS7nP2vVv_uu:-NxnX9RFC96suiQqSUo4:b-7rl0jd) ([raw file](https://github.com/cockroachdb/pebble/blob/c10cd65f353a09f1f98193b0897d0fcdfd657b58/db.go#L2450)):*

nit: waits for ...
</blockquote></details>

done.
___
*[`db.go` line 2454 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3602#-NxnUwdb8eyVmyGxCrCG:-NxnX7cN0lDEPfHphGxu:b-7rl0jd) ([raw file](https://github.com/cockroachdb/pebble/blob/c10cd65f353a09f1f98193b0897d0fcdfd657b58/db.go#L2454)):*

nit: It it does ...
</blockquote></details>

done.

jbowens:
TFTRs!

PR #3605, opened 2024-05-10
metrics: count copy compactions

RaduBerinde:
Currently copy compactions don't count towards the total number of
compactions, even though the bytes they write do count towards the
total bytes written by compactions.

This change fixes this and adds a copy compaction metric.


itsbilal (approved):
LGTM

RaduBerinde:
TFTR!

PR #3606, opened 2024-05-10
db: tolerate empty external ingestions

### LINKED ISSUE #3408: db: support empty virtual tables
Things break if we ingest an external table that has no keys in the given range. I don't remember the exact failure but it was an incorrect result. I believe I tracked it down to some levelIter code that assumed that once we reached a file in the range, it will produce a key.

See: https://github.com/cockroachdb/pebble/blob/9591f9384d50c08c7f525b0c07e134d792e7c378/metamorphic/ops.go#L949

We should support this case because the API allows it, and since we don't open external files during ingest, we can't check for it before incorporating the virtual table.

RaduBerinde:
We now tolerate empty external ingestions.

When we encounter an empty table in a download compaction, we just
delete the table.

The iterator correctness issues that occurred before seem to no longer
apply, very likely thanks to the recent refactoring and cleanup of the
iterator stack (I ran the meta test for ~5hrs and saw no failures).

Fixes #3408

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -18,7 +18,6 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/compact"
-	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/internal/keyspan"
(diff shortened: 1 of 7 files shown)


itsbilal (approved):
LGTM, nice work!

RaduBerinde:
TFTR!

PR #3607, opened 2024-05-10
objstorageprovider: small readahead improvements

sumeerbhola:
- The limit was not being updated in certain cases, which delayed when readahead would occur. This is fixed, and new test case exercises this.
- The code in recordCacheHit is almost identical to the one in maybeReadahead and the main comprehension challenge of the code is the various predicates (and their code comments with examples), which are repeated (without the examples). These are now merged into a single helper method.
- A small side-effect of the previous change is that numReads is incremented on a cache hit when we are already above the threshold. This has no actual effect on when readahead will happen. And arguably, this new behavior is more principled.
- One of the test cases had a missing - and was not being exercised, and resulting in an incorrect error in a later test case. This is fixed.
- There is additional invariant documentation.

### DIFF
--- a/objstorage/objstorageprovider/readahead.go
+++ b/objstorage/objstorageprovider/readahead.go
@@ -40,40 +40,7 @@ func makeReadaheadState(maxReadaheadSize int64) readaheadState {
 }
 
 func (rs *readaheadState) recordCacheHit(offset, blockLength int64) {
-	currentReadEnd := offset + blockLength
-	if rs.numReads >= minFileReadsForReadahead {
-		if currentReadEnd >= rs.limit && offset <= rs.limit+rs.maxReadaheadSize {
-			// This is a read that would have resulted in a readahead, had it
-			// not been a cache hit.
-			rs.limit = currentReadEnd
-			return
-		}
-		if currentReadEnd < rs.limit-rs.prevSize || offset > rs.limit+rs.maxReadaheadSize {
-			// We read too far away from rs.limit to benefit from readahead in
-			// any scenario. Reset all variables.
-			rs.numReads = 1
-			rs.limit = currentReadEnd
-			rs.size = initialReadaheadSize
-			rs.prevSize = 0
-			return
-		}
-		// Reads in the range [rs.limit - rs.prevSize, rs.limit] end up
-		// here. This is a read that is potentially benefitting from a past
-		// readahead.
-		return
-	}
-	if currentReadEnd >= rs.limit && offset <= rs.limit+rs.maxReadaheadSize {
-		// Blocks are being read sequentially and would benefit from readahead
-		// down the line.
-		rs.numReads++
-		return
-	}
-	// We read too far ahead of the last read, or before it. This indicates
-	// a random read, where readahead is not desirable. Reset all variables.
-	rs.numReads = 1
-	rs.limit = currentReadEnd
-	rs.size = initialReadaheadSize
-	rs.prevSize = 0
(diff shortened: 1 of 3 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3607)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`objstorage/objstorageprovider/readahead.go` line 96 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3607#-NxYVyE16HjMkgg757Vp:-NxYVyE16HjMkgg757Vq:byvc99d) ([raw file](https://github.com/cockroachdb/pebble/blob/2a2e3e19c5e2b7a2f53c03a448ea46f98c156954/objstorage/objstorageprovider/readahead.go#L96)):*
> ```Go
> 			//
> 			rs.numReads++
> 			if readahead {
> ```

Could we extract these two blocks from this function? This would make the semantics of the function cleaner - it just figures out if the reads matches a readahead pattern and updates the state accordingly. Each caller then does its own thing.

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3607#-:-Nx_Ex8t5HYJ-rRwIGYK:b-814g8k)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`objstorage/objstorageprovider/readahead.go` line 96 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3607#-NxYVyE16HjMkgg757Vp:-Nx_ELMg4mf9d3yzxoo2:by058v) ([raw file](https://github.com/cockroachdb/pebble/blob/2a2e3e19c5e2b7a2f53c03a448ea46f98c156954/objstorage/objstorageprovider/readahead.go#L96)):*

Could we extract these two blocks from this function? This would make the semantics of the function cleaner - it just figures out if the reads matches a readahead pattern and updates the state accordingly. Each caller then does its own thing.
</blockquote></details>

I am wary of that since there are other paths in this function that modify `limit`, `prevSize`, and `size`. It is confusing that in some cases this function will modify those fields and in other cases the caller has to modify them. And the complexity in this function is from the various nested conditions that result in this action -- the caller would need to do the same or we would need to return some bool. These choices seem harder to grok.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3607#-:-NxmxU3bBgXqcm0GCpYO:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved

___
*[`objstorage/objstorageprovider/readahead.go` line 96 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3607#-NxYVyE16HjMkgg757Vp:-NxmxVOR3Y4nLOpZmB5V:b4c9lyr) ([raw file](https://github.com/cockroachdb/pebble/blob/2a2e3e19c5e2b7a2f53c03a448ea46f98c156954/objstorage/objstorageprovider/readahead.go#L96)):*

I am wary of that since there are other paths in this function that modify `limit`, `prevSize`, and `size`. It is confusing that in some cases this function will modify those fields and in other cases the caller has to modify them. And the complexity in this function is from the various nested conditions that result in this action -- the caller would need to do the same or we would need to return some bool. These choices seem harder to grok.
</blockquote></details>

:thumbsup:

PR #3608, opened 2024-05-11
objstorage,sstable: add read-before for reader creation and iter index/filter blocks

sumeerbhola:
We consider two use cases for read-before when using a
objstorageprovider.remoteReadable, given the high latency (and cost) of
each read operation:

- When a sstable.Reader is opened, it needs to read the footer, metaindex
  block and meta properties block. It starts by reading the footer which is
  at the end of the table and then proceeds to read the other two. Instead
  of doing 3 tiny reads, we would like to do one read.

- When a single-level or two-level iterator is opened, it reads the
  (top-level) index block first. When the iterator is used, it will
  typically follow this by reading the filter block (since SeeKPrefixGE is
  common in CockroachDB). For a two-level iterator it will also read the
  lower-level index blocks which are after the filter block and before the
  top-level index block. It would be ideal if reading the top-level index
  block read enough to include the filter block. And for two-level
  iterators this would also include the lower-level index blocks.

In both use-cases we want the first read from the remoteReadable to do a
larger read, and read bytes earlier than the requested read, hence
"read-before". Subsequent reads from the remoteReadable can use the usual
readahead logic (for the second use-case above, this can help with
sequential reads of the lower-level index blocks when the read-before was
insufficient to satisfy such reads).

Since remoteReadHandle already has a buffer for read-ahead, we utilize
it for this read-before buffering pattern too.

While here, we bump up the read-ahead size for compactions to 8MB, given
the lower concurrency (and thereby higher tolerance to memory usage).

Informs cockroachdb#2328

### DIFF
--- a/objstorage/objstorage.go
+++ b/objstorage/objstorage.go
@@ -38,8 +38,37 @@ type Readable interface {
 	// The ReadHandle must be closed before the Readable is closed.
 	//
 	// Multiple separate ReadHandles can be used.
-	NewReadHandle(ctx context.Context) ReadHandle
+	NewReadHandle(ctx context.Context, readBeforeSize ReadBeforeSize) ReadHandle
 }
+
+// ReadBeforeSize specifies whether the first read should read additional
+// bytes before the offset, and how big the overall read should be. This is
+// just a suggestion that the callee can ignore (and does ignore in
+// fileReadable).
+//
+// When 0, the first read will only read what it is asked to read, say n
+// bytes. When it is a value b > 0, if b > n, then the read will be padded by
+// an additional b-n bytes to the left, resulting in an overall read size of
+// b. This behavior is akin to what the read-ahead implementation does -- when
+// the n bytes are not buffered, and there is read-ahead of b > n, the read
+// length is b bytes.
+type ReadBeforeSize int64
+
+const (
+	// NoReadBefore specifies no read-before.
+	NoReadBefore ReadBeforeSize = 0
+	// ReadBeforeForNewReader is used for a new Reader reading the footer,
+	// metaindex, properties. 32KB is unnecessarily large, but it is still small
+	// when considering remote object storage.
+	ReadBeforeForNewReader = 32 * 1024
+	// ReadBeforeForIndexAndFilter is used for an iterator reading the top-level
+	// index, filter and second-level index blocks.
+	//
+	// Consider a 128MB sstable with 32KB blocks, so 4K blocks. Say keys are
+	// ~100 bytes, then the size of the index blocks is ~400KB. 512KB is a bit
+	// bigger, and not too large to be a memory concern.
+	ReadBeforeForIndexAndFilter = 512 * 1024
+)
 
 // ReadHandle is used to perform reads that are related and might benefit from
 // optimizations like read-ahead.
@@ -347,7 +376,7 @@ type RemoteObjectToAttach struct {
 
 // Copy copies the specified range from the input to the output.
 func Copy(ctx context.Context, in Readable, out Writable, offset, length uint64) error {
-	r := in.NewReadHandle(ctx)
+	r := in.NewReadHandle(ctx, NoReadBefore)
 	r.SetupForCompaction()
 	buf := make([]byte, 256<<10)
 	end := offset + length
--- a/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go
+++ b/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go
@@ -171,11 +171,13 @@ func (r *readable) Size() int64 {
 }
 
 // NewReadHandle is part of the objstorage.Readable interface.
-func (r *readable) NewReadHandle(ctx context.Context) objstorage.ReadHandle {
+func (r *readable) NewReadHandle(
+	ctx context.Context, readBeforeSize objstorage.ReadBeforeSize,
+) objstorage.ReadHandle {
 	// It's safe to get the tracer from the generator without the mutex since it never changes.
 	t := r.mu.g.t
 	return &readHandle{
-		rh:       r.r.NewReadHandle(ctx),
+		rh:       r.r.NewReadHandle(ctx, readBeforeSize),
 		fileNum:  r.fileNum,
 		handleID: t.handleID.Add(1),
 		g:        makeEventGenerator(ctx, t),
--- a/objstorage/objstorageprovider/provider_test.go
+++ b/objstorage/objstorageprovider/provider_test.go
@@ -181,7 +181,7 @@ func TestProvider(t *testing.T) {
 				if err != nil {
 					return err.Error()
 				}
-				rh := r.NewReadHandle(ctx)
+				rh := r.NewReadHandle(ctx, objstorage.NoReadBefore)
 				if forCompaction {
 					rh.SetupForCompaction()
 				}
--- a/objstorage/objstorageprovider/remote_readable.go
+++ b/objstorage/objstorageprovider/remote_readable.go
@@ -7,6 +7,7 @@ package objstorageprovider
 import (
 	"context"
 	"io"
+	"sync"
 
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/objstorage"
@@ -24,8 +25,15 @@ func NewRemoteReadable(objReader remote.ObjectReader, size int64) objstorage.Rea
 
 const remoteMaxReadaheadSize = 1024 * 1024 /* 1MB */
 
+// Number of concurrent compactions is bounded and significantly lower than
+// the number of concurrent queries, and compactions consume reads from a few
+// levels, so there is no risk of high memory usage due to a higher readahead
+// size. So set this higher than remoteMaxReadaheadSize
+const remoteReadaheadSizeForCompaction = 8 * 1024 * 1024 /* 8MB */
+
 // remoteReadable is a very simple implementation of Readable on top of the
-// ReadCloser returned by remote.Storage.CreateObject.
+// remote.ObjectReader returned by remote.Storage.ReadObject. It is stateless
+// and can be called concurrently.
 type remoteReadable struct {
 	objReader remote.ObjectReader
 	size      int64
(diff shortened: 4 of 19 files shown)


sumeerbhola (commented):
First commit is from https://github.com/cockroachdb/pebble/pull/3607.

I'll add a test if the approach looks ok.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3608#-:-Nx_SWMY8cUkKEEMcejq:bxr5lzd)* status: 0 of 19 files reviewed, all discussions resolved (waiting on @itsbilal)

RaduBerinde (commented):
This is a cool idea, I think it's worth doing it.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3608#-:-Nxn0d0I4lmJHdGw86Yu:bbp506m)* status: 0 of 20 files reviewed, 4 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`objstorage/objstorage.go` line 41 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxn0I43-RsyFmOiUUFa:-Nxn0I43-RsyFmOiUUFb:b-b8ff6i) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorage.go#L41)):*
> ```Go
> 	//
> 	// Multiple separate ReadHandles can be used.
> 	NewReadHandle(ctx context.Context, readBeforeSize ReadBeforeSize) ReadHandle
> ```

I wonder if this would be cleaner (and less intrusive) as a new method on `ReadHandle`, where it advises what happens on the next read. It would also be more flexible, where we could use it on more than the firs tread.
___
*[`objstorage/objstorage.go` line 48 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxn1wU_3olHtZksBfgI:-Nxn1wU_3olHtZksBfgJ:bg4p04n) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorage.go#L48)):*
> ```Go
> // just a suggestion that the callee can ignore (and does ignore in
> // fileReadable).
> type ReadBeforeSize int64
> ```

The semantics of this seem pretty clear here but in the implementation this is not the number of bytes we read before, but the total size of the extended read. We should document the semantics correctly here. Perhaps also rename to `LeftExtendedReadSize`.

Or perhaps reframe it as an "area of interest" with an offset and a length (especially if we switch to a method on the ReadHandle).
___
*[`objstorage/objstorage.go` line 56 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxn06Wd269NtVdu3kZ-:-Nxn06Wd269NtVdu3kZ0:b4lhfvw) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorage.go#L56)):*
> ```Go
> 	// metaindex, properties. 32KB is unnecessarily large, but it is still small
> 	// when considering remote object storage.
> 	ReadBeforeForNewReader = 32 << 10
> ```

[nit] I find 32 * 1024 easier to read
___
*[`objstorage/objstorageprovider/remote_readable.go` line 176 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxn3YokD-EXplgnod1p:-Nxn3YokD-EXplgnod1q:b-x2f854) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorageprovider/remote_readable.go#L176)):*
> ```Go
> 	// Also, since this is the first call, the buffer must be empty.
> 	if readBeforeSize > 0 {
> 		r.buffered.offset = offset - int64(readBeforeSize-len(p))
> ```

Since the exact size is speculative and we have some leeway, perhaps we should align the offset to a 4kb boundary?

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3608#-:-Nxs-9iA0R2zRhUzGeDV:b-814g8k)* status: 0 of 20 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`objstorage/objstorage.go` line 41 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxn0I43-RsyFmOiUUFa:-NxrxfGnERfDo1JRr_gO:b-rjik16) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorage.go#L41)):*

I wonder if this would be cleaner (and less intrusive) as a new method on `ReadHandle`, where it advises what happens on the next read. It would also be more flexible, where we could use it on more than the firs tread.
</blockquote></details>

Good point. I considered it, but the difficulty is that the logic of `remoteReadHandle.ReadAt` becomes more complicated when doing this for a later read. We need to then account for how much of the read is buffered, and if both read-ahead and read-before is active, which one to prefer, or do both. And there is no use case that guides us on what the behavior ought to be. With the first read, we know there is no read-ahead and nothing in the buffer. We can definitely extend this when there is a use-case, but for these reasons I've left this unchanged for now. 
___
*[`objstorage/objstorage.go` line 48 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxn1wU_3olHtZksBfgI:-NxrxQzjDY1D6x9l8wYD:brik80k) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorage.go#L48)):*

The semantics of this seem pretty clear here but in the implementation this is not the number of bytes we read before, but the total size of the extended read. We should document the semantics correctly here. Perhaps also rename to `LeftExtendedReadSize`.

Or perhaps reframe it as an "area of interest" with an offset and a length (especially if we switch to a method on the ReadHandle).
</blockquote></details>

I added the following comment since what is happening here is quite akin to how readahead works, at least in `remoteReadHandle`.
```
// When 0, the first read will only read what it is asked to read, say n
// bytes. When it is a value b > 0, if b > n, then the read will be padded by
// an additional b-n bytes to the left, resulting in an overall read size of
// b. This behavior is akin to what the read-ahead implementation does -- when
// the n bytes are not buffered, and there is read-ahead of b > n, the read
// length is b bytes.
```

I didn't change the name because of symmetry with readahead and some similarity.
___
*[`objstorage/objstorage.go` line 56 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxn06Wd269NtVdu3kZ-:-NxryYsTBMtRIFkoojzR:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorage.go#L56)):*

[nit] I find 32 * 1024 easier to read
</blockquote></details>

Done
___
*[`objstorage/objstorageprovider/remote_readable.go` line 176 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxn3YokD-EXplgnod1p:-Nxryyll7Ecf4-S2s4yU:b-9ahlnm) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorageprovider/remote_readable.go#L176)):*

Since the exact size is speculative and we have some leeway, perhaps we should align the offset to a 4kb boundary?
</blockquote></details>

ext4 block size is 4KB, and also the sector size for typical disks. But this is remote blob storage, which typically packs blobs into the same file, and has unknown alignment.

Our configured `readBeforeSize`s are multiples of 4KB and effective `readBeforeSize` will likely be the same (since the offset will typically be large compared to the configured `readBeforeSize`).

So I don't quite understand the benefit. Could you elaborate?

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3608#-:-Nxs5II37j5qiAhkVVVG:bnfp4nl)* status: 0 of 20 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`objstorage/objstorage.go` line 41 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxn0I43-RsyFmOiUUFa:-Nxs4MtL9-7filmbapop:b-2v26qo) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorage.go#L41)):*

Good point. I considered it, but the difficulty is that the logic of `remoteReadHandle.ReadAt` becomes more complicated when doing this for a later read. We need to then account for how much of the read is buffered, and if both read-ahead and read-before is active, which one to prefer, or do both. And there is no use case that guides us on what the behavior ought to be. With the first read, we know there is no read-ahead and nothing in the buffer. We can definitely extend this when there is a use-case, but for these reasons I've left this unchanged for now.
</blockquote></details>

Got it. We already have a method `SetupForCompaction()` that reconfigures the handle. We should be consistent and either 1) also use a method for read-before (with the specification that it can only be called before the handle is used) or 2) add a `ReadHandleOptions` struct and move "for compaction" to that.
___
*[`objstorage/objstorageprovider/remote_readable.go` line 163 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxs3bpkD_4vNFnjZ8qX:-Nxs3bpkD_4vNFnjZ8qY:b-fv9a2u) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorageprovider/remote_readable.go#L163)):*
> ```Go
> // ReadAt is part of the objstorage.ReadHandle interface.
> func (r *remoteReadHandle) ReadAt(ctx context.Context, p []byte, offset int64) error {
> 	readBeforeSize := int(r.readBeforeSize)
> ```

[nit] It would be easier if we calculated exactly how many extra bytes we're reading:

```
var extraBytesBefore int64
if r.readBeforeSize > 0 {
  if r.readBeforeSize > len(p) {
    extraBytesBefore = min(int64(r.readBeforeSize - len(p)), offset)
  }
  r.readBeforeSize = 0
}
```
___
*[`objstorage/objstorageprovider/remote_readable.go` line 176 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxn3YokD-EXplgnod1p:-Nxs23Kd2-KpXaMqSxhh:b377p9p) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorageprovider/remote_readable.go#L176)):*

ext4 block size is 4KB, and also the sector size for typical disks. But this is remote blob storage, which typically packs blobs into the same file, and has unknown alignment.

Our configured `readBeforeSize`s are multiples of 4KB and effective `readBeforeSize` will likely be the same (since the offset will typically be large compared to the configured `readBeforeSize`).

So I don't quite understand the benefit. Could you elaborate?
</blockquote></details>

Sorry, I guess I spaced out and was thinking of regular storage
___
*[`objstorage/objstorageprovider/remote_readable.go` line 182 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxs3A-P618cFYG4t3Wh:-Nxs3A-P618cFYG4t3Wi:bm2lwnf) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorageprovider/remote_readable.go#L182)):*
> ```Go
> 		}
> 		// TODO(radu): we need to somehow account for this memory.
> 		if cap(r.buffered.data) >= readBeforeSize {
> ```

[nit] I'd move this block to a helper function `readToBuffer(offset, length)` and use it below as well

sumeerbhola (commented):
Test is ready.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3608#-:-Nxt1nFnBatHnl0TzIFO:bl6nke0)* status: 0 of 22 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`objstorage/objstorage.go` line 41 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxn0I43-RsyFmOiUUFa:-NxsuvsD7YRh4DHmccRh:blx7i5i) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorage.go#L41)):*

Got it. We already have a method `SetupForCompaction()` that reconfigures the handle. We should be consistent and either 1) also use a method for read-before (with the specification that it can only be called before the handle is used) or 2) add a `ReadHandleOptions` struct and move "for compaction" to that.
</blockquote></details>

I've added a TODO
```
// TODO(sumeer): both readBeforeSize and ReadHandle.SetupForCompaction are
// initial configuration of a ReadHandle. So they should both be passed as
// Options to NewReadHandle. But currently the latter is a separate method.
// This is because of how the sstable.Reader calls setupForCompaction on the
// iterators after they are constructed. Consider fixing this oddity.
```
___
*[`objstorage/objstorageprovider/remote_readable.go` line 163 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxs3bpkD_4vNFnjZ8qX:-NxsxCOQEMblVhsz1Co5:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorageprovider/remote_readable.go#L163)):*

[nit] It would be easier if we calculated exactly how many extra bytes we're reading:

```
var extraBytesBefore int64
if r.readBeforeSize > 0 {
  if r.readBeforeSize > len(p) {
    extraBytesBefore = min(int64(r.readBeforeSize - len(p)), offset)
  }
  r.readBeforeSize = 0
}
```
</blockquote></details>

Done
___
*[`objstorage/objstorageprovider/remote_readable.go` line 182 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3608#-Nxs3A-P618cFYG4t3Wh:-Nxsy_ix1OqFmLadCeOp:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/ba1b9d3c98b2e575be972ce581648fb2f1f85bf5/objstorage/objstorageprovider/remote_readable.go#L182)):*

[nit] I'd move this block to a helper function `readToBuffer(offset, length)` and use it below as well
</blockquote></details>

Done

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3608#-:-Nxws67I0NqWm6lJVn0O:bnfp4nl)* status: 0 of 22 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @sumeerbhola)

___
*[`objstorage/objstorageprovider/remote_readable_test.go` line 25 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3608#-NxwrlXJ4LdPi9mGBM-b:-NxwrlXJ4LdPi9mGBM-c:b-vnnzze) ([raw file](https://github.com/cockroachdb/pebble/blob/4c91ad124d71b1dfb36a6798d33b74ff576b1bcb/objstorage/objstorageprovider/remote_readable_test.go#L25)):*
> ```Go
> }
> 
> func (r *testObjectReader) init(size int) {
> ```

[nit] I think we could have used a logging in mem remote storage for this. No need to change anything heree, just mentioning it for the future.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3608)* status: 0 of 22 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`objstorage/objstorageprovider/remote_readable_test.go` line 25 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3608#-NxwrlXJ4LdPi9mGBM-b:-Nxx70_1Ak5vxSHG766R:b-5kbawn) ([raw file](https://github.com/cockroachdb/pebble/blob/4c91ad124d71b1dfb36a6798d33b74ff576b1bcb/objstorage/objstorageprovider/remote_readable_test.go#L25)):*

[nit] I think we could have used a logging in mem remote storage for this. No need to change anything heree, just mentioning it for the future.
</blockquote></details>

Thanks for the pointer.

PR #3611, opened 2024-05-12
db: more readable iterator stats

RaduBerinde:
This change reformats the output of iterator stats which are very hard
to read. We omit reverse counts when they are zero and remove
redundant information.

### DIFF
--- a/internal/base/iterator.go
+++ b/internal/base/iterator.go
@@ -8,6 +8,9 @@ import (
 	"context"
 	"fmt"
 	"time"
+
+	"github.com/cockroachdb/pebble/internal/humanize"
+	"github.com/cockroachdb/redact"
 )
 
 // InternalIterator iterates over a DB's key/value pairs in key order. Unlike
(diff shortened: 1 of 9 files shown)


RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3611#-:-Nxrj-sB-iEQjVtWPnZq:b-814g8k)* status: 3 of 9 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`testdata/external_iterator` line 248 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3611#-NxrdjKK2snd7fvIx899:-Nxriz69C74rjPvnleDE:b-sald4v) ([raw file](https://github.com/cockroachdb/pebble/blob/be6961f4400bfddf6d2000bef19f036f84a8cf42/testdata/external_iterator#L248)):*

do you know how it's possible that the separated value count `15` is higher than the total point key count `10`?
</blockquote></details>

This comment explains it: https://github.com/cockroachdb/pebble/blob/be6961f4400bfddf6d2000bef19f036f84a8cf42/internal/base/iterator.go#L403-L402

PR #3614, opened 2024-05-13
db: avoid unnecessary second opening of range deletion iter

jbowens:
Since #3596 the levelIter began opening a file's range deletion iterator twice: once to pass off to the mergingIter and once to interleave range deletion boundaries among the levelIter output to ensure we don't advance to a subsequent file until the file's range deletions are irrelevant. The implementation in #3596 naively always requested the range deletion iterator from the table cache a second time, even when it already discovered the file contained no range deletions. This caused a small regression in the Pebble YCSB benchmarks.

### DIFF
--- a/level_iter.go
+++ b/level_iter.go
@@ -570,7 +570,7 @@ func (l *levelIter) loadFile(file *fileMetadata, dir int) loadFileReturnIndicato
 			return noFileLoaded
 		}
 		l.iter = iters.Point()
-		if l.rangeDelIterPtr != nil {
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3614#-:-Nxnjs2uB6aSy4Ft_zz8:bnfp4nl)* status: 0 of 4 files reviewed, all discussions resolved

jbowens:
TFTR!

PR #3616, opened 2024-05-13
db: add initMergingIterLevel to flushable interface

jbowens:
Add a new initMergingIterLevel method to the flushable interface for setting up a flushable's level within the mergingIter. This is used to update ingested flushables to use a levelIter to surface range deletions, like other levels of the LSM do. This is more efficient, avoiding unnecessarily loading range deletions in other files, and consistent.


xxmplus:
Close to due to age

PR #3620, opened 2024-05-15
crl-release-24.1: ingest: avoid unnecessary I/O when checking overlap

RaduBerinde:
This commit sets the upper bound on the level iterator used when checking for overlap. This avoids unnecessarily opening a file that starts after the key range of interest.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -905,7 +905,7 @@ func overlapWithIterator(
 	// It then tries to find a point in that SST that is >= S.
 	// If there's no such point it means the SST ends in a tombstone in which case
 	// levelIter.SeekGE generates a boundary range del sentinel.
-	// The comparison of this boundary with keyRange.largest(L) below
+	// The comparison of this boundary with bounds.End(L) below
 	// is subtle but maintains correctness.
 	// 1) boundary < L,
 	//    since boundary is also > S (initial seek),
@@ -916,10 +916,12 @@ func overlapWithIterator(
 	// 3) boundary == L and L is not sentinel,
 	//    means boundary < L and hence is similar to 1).
 	// 4) boundary == L and L is sentinel,
-	//    we'll always overlap since for any values of i,j ranges [i, k) and [j, k) always overlap.
+	//    means we either hit the iterator upper bound or the SST ends in a range
+	//    deletion at that same key. We require checking for overlap with
+	//    rangeDelIter.
 	key, _ := iter.SeekGE(bounds.Start, base.SeekGEFlagsNone)
 	if key != nil {
-		if bounds.End.IsUpperBoundForInternalKey(cmp, *key) {
+		if bounds.End.IsUpperBoundFor(cmp, key.UserKey) {
 			return true
 		}
 	}
@@ -971,7 +973,7 @@ func overlapWithIterator(
 func ingestTargetLevel(
 	newIters tableNewIters,
 	newRangeKeyIter keyspanimpl.TableNewSpanIter,
-	iterOps IterOptions,
+	iterOpts IterOptions,
 	comparer *Comparer,
 	v *version,
 	baseLevel int,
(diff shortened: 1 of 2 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3620)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`ingest.go` line 935 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3620#-NxtopJi12iAv3pDC53G:-NxtopJi12iAv3pDC53H:bzaeiqc) ([raw file](https://github.com/cockroachdb/pebble/blob/8f06cb795453f16b645ab0eedd1bcb0bd48e516d/ingest.go#L935)):*
> ```Go
> 	computeOverlapWithSpans := func(rIter keyspan.FragmentIterator) (bool, error) {
> 		// NB: The spans surfaced by the fragment iterator are non-overlapping.
> 		span, err := rIter.SeekGE(bounds.Start)
> ```

Not sure this fixes much, I think we'd still open the next file here?
___
*[`ingest.go` line 1109 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3620#-Nxtp3hp2W-hukZWPjVa:-Nxtp3hp2W-hukZWPjVb:b-y8h73v) ([raw file](https://github.com/cockroachdb/pebble/blob/8f06cb795453f16b645ab0eedd1bcb0bd48e516d/ingest.go#L1109)):*
> ```Go
> 		rkeyLevelIter.Init(
> 			keyspan.SpanIterOptions{}, comparer.Compare, newRangeKeyIter,
> 			v.Levels[level].Iter(), manifest.Level(level), manifest.KeyTypeRange,
> ```

should this be `v.RangeKeyLevels`?

RaduBerinde:
*[`ingest.go` line 935 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3620#-NxtopJi12iAv3pDC53G:-NxwSYpJ1UpTPNE0FpIO:btjxyz2) ([raw file](https://github.com/cockroachdb/pebble/blob/8f06cb795453f16b645ab0eedd1bcb0bd48e516d/ingest.go#L935)):*

Not sure this fixes much, I think we'd still open the next file here?
</blockquote></details>

Well.. only when there's another file on the level which range keys

RaduBerinde:
*[`ingest.go` line 935 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3620#-NxtopJi12iAv3pDC53G:-NxwSgTJ7jU2m5f4M-8F:bllw3sc) ([raw file](https://github.com/cockroachdb/pebble/blob/8f06cb795453f16b645ab0eedd1bcb0bd48e516d/ingest.go#L935)):*

Well.. only when there's another file on the level which range keys
</blockquote></details>

*with range keys

jbowens (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3620#-:-NxxRdd-79z7LV7W5S_V:bg7igkv)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`ingest.go` line 1109 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3620#-Nxtp3hp2W-hukZWPjVa:-NxxR4ZH2kqtv5yF4NG8:b-uhdf3l) ([raw file](https://github.com/cockroachdb/pebble/blob/8f06cb795453f16b645ab0eedd1bcb0bd48e516d/ingest.go#L1109)):*

should this be `v.RangeKeyLevels`?
</blockquote></details>

Good catch, yeah it should be. `v.Levels` is still correct since it contains a superset of the files in `v.RangeKeyLevels`, but the keyspanimpl.LevelIter needs to potentially wade through many files with no range keys

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3620#-:-Ny3VXMu-CP1Ym6RruTu:b-814g8k)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`ingest.go` line 1109 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3620#-Nxtp3hp2W-hukZWPjVa:-Ny3VY__5u9qYxZBFEyY:b-xj8u2) ([raw file](https://github.com/cockroachdb/pebble/blob/8f06cb795453f16b645ab0eedd1bcb0bd48e516d/ingest.go#L1109)):*

Good catch, yeah it should be. `v.Levels` is still correct since it contains a superset of the files in `v.RangeKeyLevels`, but the keyspanimpl.LevelIter needs to potentially wade through many files with no range keys
</blockquote></details>

I'll leave it as is here to minimize the changes. Shouldn't be a big deal, the `keyspanimpl.LevelIter` will filter out files that have no range keys.

PR #3622, opened 2024-05-15
manifest: minor cleanup

RaduBerinde:
This change cleans up a few aspects of the `manifest` package:
 - we move `overlaps` to a method on the `LevelSlice`
 - we simplify the interface and improve the implementation of
   `LevelIterator.Find()`
 - we remove no-op uses of `LevelSlice.Reslice()`

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -93,7 +93,7 @@ type compactionLevel struct {
 func (cl compactionLevel) Clone() compactionLevel {
 	newCL := compactionLevel{
 		level: cl.level,
-		files: cl.files.Reslice(func(start, end *manifest.LevelIterator) {}),
+		files: cl.files,
 	}
 	return newCL
 }
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -136,7 +136,7 @@ type sublevelInfo struct {
 func (cl sublevelInfo) Clone() sublevelInfo {
 	return sublevelInfo{
 		sublevel:   cl.sublevel,
-		LevelSlice: cl.LevelSlice.Reslice(func(start, end *manifest.LevelIterator) {}),
+		LevelSlice: cl.LevelSlice,
 	}
 }
 func (cl sublevelInfo) String() string {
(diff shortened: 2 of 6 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3622#-:-Ny0zy203gA3oXBNFBtn:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/manifest/level_metadata.go` line 124 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3622#-Ny0zCnHDY4lLX2BDvbL:-Ny0zCnHDY4lLX2BDvbM:b-i5pde7) ([raw file](https://github.com/cockroachdb/pebble/blob/144fdc391ae59dbb263d5abeeb5a59f0f033dedc/internal/manifest/level_metadata.go#L124)):*
> ```Go
> 		// We only need to look at the portion of files that are "equal" to m with
> 		// respect to the L0 ordering.
> 		f := iter.seek(func(f *FileMetadata) bool {
> ```

And the existing code was iterating over all the files in L0? Nice improvement.

And this is the `LevelMetadata` for all of L0 and not just a sub-level, yes?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3622#-:-Ny1B8M13ARSjvceolag:b-814g8k)* status: all files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`internal/manifest/level_metadata.go` line 124 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3622#-Ny0zCnHDY4lLX2BDvbL:-Ny1BAIM0E3MgXQTBq5P:bvzjqzt) ([raw file](https://github.com/cockroachdb/pebble/blob/144fdc391ae59dbb263d5abeeb5a59f0f033dedc/internal/manifest/level_metadata.go#L124)):*

And the existing code was iterating over all the files in L0? Nice improvement.

And this is the `LevelMetadata` for all of L0 and not just a sub-level, yes?
</blockquote></details>

Correct. I believe we never have a `LevelMetadata` for just a sublevel (only `LevelSlice` and `[]*FileMetadata`)

PR #3624, opened 2024-05-15
db: add SharedLowerUserKeyPrefix and WriteSharedWithStrictObsolete op…

### LINKED ISSUE #2756: db: use strict-obsolete sstables when writing to shared storage

sumeerbhola:
…tions

SharedLowerUserKeyPrefix, if specified, is an additional lower bound on constraint on key prefixes that should be written to shared files. It applies only when CreateOnShared permits some shared file creation. It will be used by CockroachDB to exclude keys below TableDataMin from shared files, for both correctness (they can contain MERGEs for which the obsolete bit does not work) and performance reasons (low latency is more important and the data volume is tiny).

WriteSharedWithStrictObsolete, when true, causes shared files to be written with WriterOptions.IsStrictObsolete set to true. This adds an extra measure of configuration protection to accidentally sharing files where the MERGE could become visible (we currently share such files, but file virtualization hides these MERGEs).

The enforcement of SharedLowerUserKeyPrefix changes how PreferSharedStorage is computed during flushes and compactions. It will only be set if the next key to be written by compact.Runner permits writing to shared storage. compact.Runner optimizes this computation for when a compaction is fully within the shared or unshared bounds. Additionally compact.Runner uses the existing OutputSplitter to decide when a split should happen when transitioning from unshared to shared.

While here, we do a tiny optimization in OutputSplitter to remove a key comparison on each iteration key.

Fixes #2756

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2518,14 +2518,23 @@ func (d *DB) compactAndWrite(
 		MaxGrandparentOverlapBytes: c.maxOverlapBytes,
 		TargetOutputFileSize:       c.maxOutputFileSize,
 	}
+	considerCreateShared := remote.ShouldCreateShared(d.opts.Experimental.CreateOnShared, c.outputLevel.level)
+	if considerCreateShared {
+		runnerCfg.ConsiderCreateShared = true
+		runnerCfg.SharedLowerUserKeyPrefix = d.opts.Experimental.SharedLowerUserKeyPrefix
+	}
 	runner := compact.NewRunner(runnerCfg, iter)
-	for runner.MoreDataToWrite() {
+	for {
+		moreData, keyShouldBeWrittenToShared := runner.MoreDataToWrite()
+		if !moreData {
+			break
+		}
 		if c.cancel.Load() {
 			return runner.Finish().WithError(ErrCancelledCompaction)
 		}
 		// Create a new table.
 		writerOpts := d.opts.MakeWriterOptions(c.outputLevel.level, tableFormat)
-		objMeta, tw, cpuWorkHandle, err := d.newCompactionOutput(jobID, c, writerOpts)
+		objMeta, tw, cpuWorkHandle, err := d.newCompactionOutput(jobID, c, writerOpts, keyShouldBeWrittenToShared)
 		if err != nil {
 			return runner.Finish().WithError(err)
 		}
@@ -2652,7 +2661,7 @@ func (c *compaction) makeVersionEdit(result compact.Result) (*versionEdit, error
 // newCompactionOutput creates an object for a new table produced by a
 // compaction or flush.
 func (d *DB) newCompactionOutput(
-	jobID JobID, c *compaction, writerOpts sstable.WriterOptions,
+	jobID JobID, c *compaction, writerOpts sstable.WriterOptions, preferSharedStorage bool,
 ) (objstorage.ObjectMetadata, *sstable.Writer, CPUWorkHandle, error) {
 	d.mu.Lock()
 	diskFileNum := d.mu.versions.getNextDiskFileNum()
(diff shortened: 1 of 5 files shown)


sumeerbhola (commented):
There is no test -- I'll add one once the configuration settings look ok.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3624#-:-Nxxv6DX2qmkOEHKTTe0:bqelw5)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde (approved):
LGTM, just some minor questions

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3624#-:-NxyrLEXArDidX-XMVY-:b-40lmar)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @sumeerbhola)

___
*[`db.go` line 1261 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3624#-NxyrVfs0lJ03tRmcGky:-NxyrVfs0lJ03tRmcGkz:b6zy707) ([raw file](https://github.com/cockroachdb/pebble/blob/38e36b2ca87fa08274c2636a52a0561b6e9cd1f5/db.go#L1261)):*
> ```Go
> //     can of course conform to the stronger remote.CreateOnSharedAll).
> //
> //   - If Options.Experimental.SharedLowerUserKeyPrefix is non-nil, the lower
> ```

Why not make this a generic user key rather than a prefix? We will still set it to a prefix in CRDB but it will make the boundary checks easier.
___
*[`internal/compact/run.go` line 110 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3624#-NxypnHt99KmVndPVOCU:-NxypnHt99KmVndPVOCV:bcq01jh) ([raw file](https://github.com/cockroachdb/pebble/blob/38e36b2ca87fa08274c2636a52a0561b6e9cd1f5/internal/compact/run.go#L110)):*
> ```Go
> 	split base.Split
> 	// At most one state transition from false => true, when the next key should
> 	// be written to shared storage.
> ```

[nit] I'd point to `cfg.SharedLowerUserKeyPrefix` so the reader can understand the transition
___
*[`internal/compact/run.go` line 142 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3624#-NxyqMpO3Umg4FcPzyTE:-NxyqMpO3Umg4FcPzyTF:bowpmms) ([raw file](https://github.com/cockroachdb/pebble/blob/38e36b2ca87fa08274c2636a52a0561b6e9cd1f5/internal/compact/run.go#L142)):*
> ```Go
> 		r.cfg.SharedLowerUserKeyPrefix = nil
> 	} else {
> 		// Check if the compaction will only write non-shared keys.
> ```

Can't we just check `compactionBounds.End.IsUpperBoundFor(SharedLowerUserKeyPrefix)`? Are you worried of an end key that has the `SharedLowerUserKeyPrefix` and a suffix and yet it compares before the prefix? (i guess that's allowed even if it won't happen with crdb?)
___
*[`internal/compact/splitting.go` line 265 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3624#-NxynhKlBkaWykXclkuT:-NxynhKlBkaWykXclkuU:b-bwwlxg) ([raw file](https://github.com/cockroachdb/pebble/blob/38e36b2ca87fa08274c2636a52a0561b6e9cd1f5/internal/compact/splitting.go#L265)):*
> ```Go
> 		// When the target file size limit is very small (in tests), we could end up
> 		// splitting at the first key, which is not allowed.
> 		if !s.pastStartKey && s.cmp(nextUserKey, s.startKey) <= 0 {
> ```

The change looks good but just to defend the existing code :) - this is not a performance-sensitive path: if we are here, we are not at a grandparent boundary and `shouldSplitBasedOnSize` said that we should split (and it will continue to do so). So the comparison only happens near the split boundaries.

xxmplus:
Close to due to age

PR #3628, opened 2024-05-21
sstable: handle synthetic prefix when checking bloom filter

### LINKED ISSUE #3627: external ingest: fix bloom filters when synthetic prefix is used
We found that we do not un-apply the synthetic prefix when checking bloom filters. The result is that `SeekPrefixGE` will more often than not miss keys. This is a major bug that affects Online Restore and is causing corruption on tpcc tests.

I investigated why the metamorphic test did not find this issue. I found that a `objDB` map wasn't updated properly which lead to the code generating SeekPrefixGE operations not being able to use the keys that were written.

In addition to fixing the bug, I am identifying some key testing misses that got us here:
 - the metamorphic test code is a bit lax in testing its assumptions; for example, we should have asserted when `objDB` returned a 0 value
 - there aren't enough tests for the metamorphic generation code
 - there is a tendency to over-rely on the metamorphic tests; in this instance we were missing test coverage for SeekPrefixGE in the context of synthetic prefixes

RaduBerinde:
Fixes #3627

#### metamorphic: fix derived db ID

We had a bug in `deriveDB` and we weren't setting the right entry in
the map. This resulted in `iterSeekPrefixGE` not being able to use
known keys; in this case it falls back to returning a previously
generated random key, which is highly unlikely to match a key from an
external object ingested with prefix replacement. This caused the meta
test to completely miss a severe bug with bloom filters and prefix
replacement.

#### sstable: handle synthetic prefix when checking bloom filter

Fix the sstable iterators to strip the synthetic prefix before
checking the bloom filter.

Fix the `TestRandomizedPrefixSuffixRewriter` to check `SeekPrefixGE`
and randomly enable filters.

### DIFF
--- a/metamorphic/generator.go
+++ b/metamorphic/generator.go
@@ -99,7 +99,7 @@ type generator struct {
 	// or snapshots maps.
 	iters map[objID]objIDSet
 	// objectID -> db: used to keep track of the DB a batch, iter, or snapshot
-	// was created on.
+	// was created on. It should be read through the dbIDForObj method.
 	objDB map[objID]objID
 	// readerID -> reader iters: used to keep track of the open iterators on a
 	// reader. The iter set value will also be indexed by either the batches or
@@ -318,7 +318,7 @@ func (g *generator) batchCommit() {
 	}
 
 	batchID := g.liveBatches.rand(g.rng)
-	dbID := g.objDB[batchID]
+	dbID := g.dbIDForObj(batchID)
 	g.removeBatchFromGenerator(batchID)
 
 	// The batch we're applying may contain single delete tombstones that when
@@ -574,15 +574,22 @@ func (g *generator) randKeyTypesAndMask() (keyTypes uint32, maskSuffix []byte) {
 }
 
 func (g *generator) deriveDB(readerID objID) objID {
-	if readerID.tag() == iterTag {
-		readerID = g.iterReaderID[readerID]
-	}
 	dbParentID := readerID
+	if readerID.tag() == iterTag {
+		dbParentID = g.iterReaderID[readerID]
+	}
 	if dbParentID.tag() != dbTag {
-		dbParentID = g.objDB[dbParentID]
+		dbParentID = g.dbIDForObj(dbParentID)
 	}
 	g.objDB[readerID] = dbParentID
 	return dbParentID
+}
+
+func (g *generator) dbIDForObj(objID objID) objID {
+	if g.objDB[objID] == 0 {
+		panic(fmt.Sprintf("object %s has no associated DB", objID))
+	}
+	return g.objDB[objID]
 }
 
 func (g *generator) newIterUsingClone() {
@@ -815,7 +822,7 @@ func (g *generator) iterSeekPrefixGE(iterID objID) {
 	// random key.
 	if g.rng.Intn(10) >= 1 {
 		possibleKeys := make([][]byte, 0, 100)
-		inRangeKeys := g.keyManager.InRangeKeysForObj(g.objDB[iterID], lower, upper)
+		inRangeKeys := g.keyManager.InRangeKeysForObj(g.dbIDForObj(iterID), lower, upper)
 		for _, keyMeta := range inRangeKeys {
 			visibleHistory := keyMeta.history.before(iterCreationTimestamp)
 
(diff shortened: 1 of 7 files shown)


dt (approved):

dt:
That was quick! Nice find!

sumeerbhola (approved):
:lgtm: 

> Reviewed 2 of 2 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3628#-:-NyS0HGC-KPk2rQ6juuA:bg7igkv)* status: all files reviewed, 3 unresolved discussions (waiting on @msbutler and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 792 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3628#-NyS0Dl1Ee1ZbVippOVp:-NyS0Dl1Ee1ZbVippOVq:b-qbxrht) ([raw file](https://github.com/cockroachdb/pebble/blob/0f634ac891503be9fd6d4cbe03fb2feb2143a5cf/sstable/reader_iter_single_lvl.go#L792)):*
> ```Go
> 		mayContain, i.err = i.bloomFilterMayContain(prefix)
> 		if i.err != nil || !mayContain {
> 			// This invalidation may not be necessary for correctness, and may
> ```

same adjustment as other comment.
___
*[`sstable/reader_iter_single_lvl.go` line 834 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3628#-NyS-q8r8JxsYlT8HutV:-NyS-q8r8JxsYlT8HutW:btywm1s) ([raw file](https://github.com/cockroachdb/pebble/blob/0f634ac891503be9fd6d4cbe03fb2feb2143a5cf/sstable/reader_iter_single_lvl.go#L834)):*
> ```Go
> 	prefixToCheck := prefix
> 	if i.transforms.SyntheticPrefix.IsSet() {
> 		// We have to remove the synthetic prefix.
> ```

nit: is it simpler to use `res, ok := bytes.CutPrefix(key, sp)` like is done in `SyntheticPrefix.Invert`?
___
*[`sstable/reader_iter_two_lvl.go` line 396 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3628#-NyRzukX4ndR3ilgmeOe:-NyRzukX4ndR3ilgmeOf:bttov97) ([raw file](https://github.com/cockroachdb/pebble/blob/0f634ac891503be9fd6d4cbe03fb2feb2143a5cf/sstable/reader_iter_two_lvl.go#L396)):*
> ```Go
> 		mayContain, i.err = i.bloomFilterMayContain(prefix)
> 		if i.err != nil || !mayContain {
> 			// This invalidation may not be necessary for correctness, and may
> ```

This "may not be necessary for correctness" should now be qualified to say that it may not be necessary if i.err == nil.

RaduBerinde (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3628#-:-NyTNVd90QcJkaE7yjs8:bz5mair)* status: 2 of 7 files reviewed, 2 unresolved discussions (waiting on @msbutler and @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 792 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3628#-NyS0Dl1Ee1ZbVippOVp:-NyTNM2XER4EVVxMU7jI:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0f634ac891503be9fd6d4cbe03fb2feb2143a5cf/sstable/reader_iter_single_lvl.go#L792)):*

same adjustment as other comment.
</blockquote></details>

Done.
___
*[`sstable/reader_iter_two_lvl.go` line 396 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3628#-NyRzukX4ndR3ilgmeOe:-NyTNLR7BQu25jYOJTIY:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0f634ac891503be9fd6d4cbe03fb2feb2143a5cf/sstable/reader_iter_two_lvl.go#L396)):*

This "may not be necessary for correctness" should now be qualified to say that it may not be necessary if i.err == nil.
</blockquote></details>

Done.

RaduBerinde:
Opened backport PR here: https://github.com/cockroachdb/pebble/pull/3629

PR #3626, opened 2024-05-16
ingest: perform LSM overlap checks outside of DB.mu

RaduBerinde:
This change separates the LSM overlap checking from
`ingestTargetLevel`, with the former now happening without holding
`DB.mu`.

We will make more improvements in this area soon, but this change
should be a significant improvement on its own.

Informs #2112

PR #3632, opened 2024-05-23
overlap: rework ingest LSM overlap check code

RaduBerinde:
~This PR is only for the last commit (it is based on #3626).~

#### overlap: rework ingest LSM overlap check code

Instead of using level iterators, we directly check overlap with files
separately. This requires less machinery and there is no chance of
opening files that are outside of the bounds of interest.

We put the new code in a new `internal/overlap` package and
`overlapChecker` is now a very small wrapper which interfaces with
`overlap.Checker`.

### DIFF
--- a/db.go
+++ b/db.go
@@ -3007,7 +3007,13 @@ func (d *DB) checkVirtualBounds(m *fileMetadata) {
 
 // DebugString returns a debugging string describing the LSM.
 func (d *DB) DebugString() string {
+	return d.DebugCurrentVersion().DebugString()
+}
+
+// DebugCurrentVersion returns the current LSM tree metadata. Should only be
+// used for testing/debugging.
+func (d *DB) DebugCurrentVersion() *manifest.Version {
 	d.mu.Lock()
 	defer d.mu.Unlock()
-	return d.mu.versions.currentVersion().DebugString()
+	return d.mu.versions.currentVersion()
 }
--- a/flushable.go
+++ b/flushable.go
@@ -394,3 +394,33 @@ func determineOverlapAllIters(
 	}
 	return false, nil
 }
+
+func determineOverlapPointIterator(
+	cmp base.Compare, bounds base.UserKeyBounds, iter internalIterator,
+) (bool, error) {
+	kv := iter.SeekGE(bounds.Start, base.SeekGEFlagsNone)
+	if kv == nil {
+		return false, iter.Error()
+	}
+	return bounds.End.IsUpperBoundForInternalKey(cmp, kv.K), nil
+}
+
+func determineOverlapKeyspanIterator(
+	cmp base.Compare, bounds base.UserKeyBounds, iter keyspan.FragmentIterator,
+) (bool, error) {
+	// NB: The spans surfaced by the fragment iterator are non-overlapping.
+	span, err := iter.SeekGE(bounds.Start)
+	if err != nil {
+		return false, err
+	}
+	for ; span != nil; span, err = iter.Next() {
+		if !bounds.End.IsUpperBoundFor(cmp, span.Start) {
+			// The span starts after our bounds.
+			return false, nil
+		}
+		if !span.Empty() {
+			return true, nil
+		}
+	}
+	return false, err
+}
--- a/ingest.go
+++ b/ingest.go
@@ -15,6 +15,7 @@ import (
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/internal/manifest"
+	"github.com/cockroachdb/pebble/internal/overlap"
 	"github.com/cockroachdb/pebble/internal/private"
 	"github.com/cockroachdb/pebble/objstorage"
 	"github.com/cockroachdb/pebble/objstorage/remote"
@@ -832,7 +833,7 @@ func ingestUpdateSeqNum(
 func ingestTargetLevel(
 	ctx context.Context,
 	cmp base.Compare,
-	lsmOverlap lsmOverlap,
+	lsmOverlap overlap.WithLSM,
 	baseLevel int,
 	compactions map[*compaction]struct{},
 	meta *fileMetadata,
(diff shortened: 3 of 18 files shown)


sumeerbhola (approved):
:lgtm: 

> Reviewed 11 of 14 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3632#-:-Nyf-UNeCrPk1VEzk5mO:bg7igkv)* status: all files reviewed, 5 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/overlap/checker.go` line 101 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3632#-NyevmAP0bAiFH5WO0k5:-NyevmAP0bAiFH5WO0k6:b8bsyfv) ([raw file](https://github.com/cockroachdb/pebble/blob/047da513a00f2ccdcd0a52428593589cfcd625ce/internal/overlap/checker.go#L101)):*
> ```Go
> 	}
> 
> 	overlap, err := c.PossibleOverlap(ctx, region, boundaryOverlaps)
> ```

nit: slight preference for `PossibleDataOverlap`
___
*[`internal/overlap/checker.go` line 147 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3632#-NyewRc16nVisnhX_Y5H:-NyewRc16nVisnhX_Y5I:bjvdtqp) ([raw file](https://github.com/cockroachdb/pebble/blob/047da513a00f2ccdcd0a52428593589cfcd625ce/internal/overlap/checker.go#L147)):*
> ```Go
> 		}
> 	}
> 	// Note: there are at most two files, or we would have returned above.
> ```

There can be (many) more than two files for L0 since files can be overlapping.
___
*[`internal/overlap/checker.go` line 253 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3632#-NyexLXq7DzFALb3ZZys:-NyexLXq7DzFALb3ZZyt:bjdn6t0) ([raw file](https://github.com/cockroachdb/pebble/blob/047da513a00f2ccdcd0a52428593589cfcd625ce/internal/overlap/checker.go#L253)):*
> ```Go
> 	var err error
> 	if c.cmp(region.Start, fragmentLowerBound) <= 0 {
> 		span, err = fragments.First()
> ```

Using `First()` is only a performance optimization, yes?
Since `fragmentLowerBound` when this is called from `c.emptyFragmentRegion(region, pointBounds.Start, rangeDels)` is not a tight lower bound.
___
*[`testdata/event_listener` line 189 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3632#-NyetzpBBGWN46L_wx87:-NyetzpBBGWN46L_wx88:b2jhc9v) ([raw file](https://github.com/cockroachdb/pebble/blob/2ac449bbd7e59d2c7a6d1402183fbdc3e02cf3e5/testdata/event_listener#L189)):*
> ```
> [JOB 10] ingesting: sstable created 000015
> sync: db
> open: db/000013.sst
> ```

Why did we not need to open this sst for overlap checking? Glancing at `TestEventListener` it seems that both flushed data and ingested data is only using user key "a", so there will be boundary overlap, which means we will need to open the existing files to determine file overlap. What am I missing?
___
*[`internal/overlap/testdata/checker` line 8 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3632#-Nyf-Kag7x2adrTJv2zQ:-Nyf-Kag7x2adrTJv2zR:b4mfxd7) ([raw file](https://github.com/cockroachdb/pebble/blob/047da513a00f2ccdcd0a52428593589cfcd625ce/internal/overlap/testdata/checker#L8)):*
> ```
> ----
> 
> overlap t1
> ```

how about outputting which iterators were opened for each overlap check, so we have better certainty that the algorithm is working as intended?

RaduBerinde:
*[`testdata/event_listener` line 189 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3632#-NyetzpBBGWN46L_wx87:-Nyffyzn3UvUirasMCjX:b-ay96fx) ([raw file](https://github.com/cockroachdb/pebble/blob/2ac449bbd7e59d2c7a6d1402183fbdc3e02cf3e5/testdata/event_listener#L189)):*

Why did we not need to open this sst for overlap checking? Glancing at `TestEventListener` it seems that both flushed data and ingested data is only using user key "a", so there will be boundary overlap, which means we will need to open the existing files to determine file overlap. What am I missing?
</blockquote></details>

We assume data overlap if the ingested file contains/encloses a file, which we now check first.

RaduBerinde (commented):
TFTR!  After discussing with David, I updated the change to assume data overlap whenever our region contains a file boundary (explained further in a comment in LevelOverlap).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3632#-:-NygQak76wgPs7B9HbQa:b4onu7f)* status: 13 of 19 files reviewed, 5 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/overlap/checker.go` line 101 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3632#-NyevmAP0bAiFH5WO0k5:-NygQWcf3XL7chQju3TR:bwf7owv) ([raw file](https://github.com/cockroachdb/pebble/blob/047da513a00f2ccdcd0a52428593589cfcd625ce/internal/overlap/checker.go#L101)):*

nit: slight preference for `PossibleDataOverlap`
</blockquote></details>

I folded this into `LevelOverlap`
___
*[`internal/overlap/checker.go` line 253 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3632#-NyexLXq7DzFALb3ZZys:-NygQZ5b7rA9NjwaeNXF:b-hp0x2x) ([raw file](https://github.com/cockroachdb/pebble/blob/047da513a00f2ccdcd0a52428593589cfcd625ce/internal/overlap/checker.go#L253)):*

Using `First()` is only a performance optimization, yes?
Since `fragmentLowerBound` when this is called from `c.emptyFragmentRegion(region, pointBounds.Start, rangeDels)` is not a tight lower bound.
</blockquote></details>

Correct, added a comment.
___
*[`internal/overlap/testdata/checker` line 8 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3632#-Nyf-Kag7x2adrTJv2zQ:-NygQUxj5A9F3sp4ePVA:bqqbert) ([raw file](https://github.com/cockroachdb/pebble/blob/047da513a00f2ccdcd0a52428593589cfcd625ce/internal/overlap/testdata/checker#L8)):*

how about outputting which iterators were opened for each overlap check, so we have better certainty that the algorithm is working as intended?
</blockquote></details>

Great idea, done.

RaduBerinde:
@sumeerbhola - did you want to take a look at the updates or it's ok to merge? Thanks!

RaduBerinde:
*[`internal/overlap/checker.go` line 91 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3632#-Nz52l7_7sdtS1UrF_YR:-Nz558fo7_j_DtelA1Hh:b-yqt663) ([raw file](https://github.com/cockroachdb/pebble/blob/9a226fc9a887084347b8ebf44da607618e58f99d/internal/overlap/checker.go#L91)):*

Worth adding a code comment that we don't populate the split file for L0 since there is no non-overlapping requirement.
</blockquote></details>

Makes sense, I'll add it in the next change.

PR #3634, opened 2024-05-23
compact: use frontier to "expire" the last range del span

RaduBerinde:
When the compaction iterator is inside a range del, it has to compare
each key to `Span.End`. In this commit we reduce these key comparisons
by registering a frontier for the end of the span.

### DIFF
--- a/internal/compact/iterator.go
+++ b/internal/compact/iterator.go
@@ -239,6 +239,9 @@ type Iter struct {
 	// lastRangeDelSpan stores the last, not compacted tombstone span. It is used
 	// to elide points or mark them as snapshot-pinned.
 	lastRangeDelSpan keyspan.Span
+	// lastRangeDelSpanFrontier is the frontier used to clear out lastRangeDelSpan
+	// when we move beyond its end key.
+	lastRangeDelSpanFrontier frontier
 
 	// span stores the last, compacted tombstone or range key span. It is provided
 	// to the caller via Span().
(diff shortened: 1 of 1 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3634#-:-Nz-57w9E0fGqyG4vg_H:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/compact/iterator.go` line 415 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3634#-Nz-3lYwEgonTZoA0yNg:-Nz-3lYwEgonTZoA0yNh:b-uutjt8) ([raw file](https://github.com/cockroachdb/pebble/blob/2335a56140248a20ddab3768380b81a537d95836/internal/compact/iterator.go#L415)):*
> ```Go
> 
> 	for i.iterKV != nil {
> 		i.frontiers.Advance(i.iterKV.K.UserKey)
> ```

Was there a reason we were earlier doing this `Advance` in `saveKey` instead of here? This location seems more principled. Did the earlier code reduce the number of times we would call `Advance`, i.e., was it an attempt at an optimization?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3634#-:-Nz-6koI76Wohsv0jB3o:b-814g8k)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`internal/compact/iterator.go` line 415 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3634#-Nz-3lYwEgonTZoA0yNg:-Nz-6FUaBjxLZORL5H9Y:b-i0pb3d) ([raw file](https://github.com/cockroachdb/pebble/blob/2335a56140248a20ddab3768380b81a537d95836/internal/compact/iterator.go#L415)):*

Was there a reason we were earlier doing this `Advance` in `saveKey` instead of here? This location seems more principled. Did the earlier code reduce the number of times we would call `Advance`, i.e., was it an attempt at an optimization?
</blockquote></details>

I wondered the same thing. Maybe @itsbilal or @jbowens remember some context? I doubt it would be an important optimization.

PR #3647, opened 2024-06-05
base: minor improvement to IteratorStats string

RaduBerinde:
Show the cached and not-cached block bytes instead of the total and
cached.


nicktrav:
Thoughts on adding the block count and the blocks missed too?

nicktrav (approved):
LGTM for the low risk piece.

RaduBerinde:
Ok, I added counts and switched to counting cached and uncached.

nicktrav (approved):
Nice.

sumeerbhola (approved):
worth scoping out how this will affect the existing CRDB code, before merging this.

I don't think we display the corresponding metrics in the dbconsole, and no one should be alerting on them, so we could probably just alter the existing CRDB metrics.

:lgtm: 

> Reviewed 10 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3647#-:-Nzhyn2pCDoU08YuZhKh:b8yenlx)* status: 10 of 14 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`internal/base/iterator.go` line 372 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3647#-Nzhxokm3i876t5qnMBh:-Nzhxokm3i876t5qnMBi:b-nephe4) ([raw file](https://github.com/cockroachdb/pebble/blob/ab925fa8729b16a1b14879fd6a49410d27d828a0/internal/base/iterator.go#L372)):*
> ```Go
> 	// Number of blocks that were accessed from cache.
> 	CachedBlocks uint32
> 	// Number of blocks that had to be read from disk.
> ```

Consider adding a code comment that says something to the effect:
These missed the Pebble internal block cache. These could be served from an OS/filesystem cache, like the page cache, and Pebble is ignorant of that.

RaduBerinde:
Good point. We would want to change `MVCCIteratorStats` and the `crdb_internal.node_transaction_statistics` table. That's definitely not something I will purse as a backport. Sigh. Will simplify this PR back to just improving the string.

RaduBerinde:
Changed back to just a string improvement, updated to use the "X cached, Y not cached" formulation.

sean- (commented):
> Reviewed 13 of 14 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3647)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`internal/base/iterator.go` line 439 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3647#-O-27G09E3my6tjRqABv:-O-27G09E3my6tjRqABw:bht6w2v) ([raw file](https://github.com/cockroachdb/pebble/blob/38bc391f7c2c209d208c3219a25c5be24bdf2948/internal/base/iterator.go#L439)):*
> ```Go
> // SafeFormat implements the redact.SafeFormatter interface.
> func (s *InternalIteratorStats) SafeFormat(p redact.SafePrinter, verb rune) {
> 	p.Printf("blocks: %s cached",
> ```

Is this string accurate?  Isn't`BlockBytesInCache` bytes?

https://github.com/cockroachdb/pebble/blob/47da75f027b9a48878893ed59f0d7d3b429765c8/internal/base/iterator.go#L365-L375

I think the right format should be something closer to:
```
"iterator cache stats: %s hit",
```

so the output looks like:

```
iterator cache stats: 384B hit, 4KiB bytes missed (read time: 11ms)
```

or something like that.  Using the prefix "blocks:" makes me think of the number of 32K pebble pages, not bytes, and I would be led astray in an investigation.

RaduBerinde:
*[`internal/base/iterator.go` line 439 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3647#-O-27G09E3my6tjRqABv:-O-2qJOJ3_qSpAfEkPLC:b-a2tn46) ([raw file](https://github.com/cockroachdb/pebble/blob/38bc391f7c2c209d208c3219a25c5be24bdf2948/internal/base/iterator.go#L439)):*
<details><summary><i>Previously, sean- (Sean Chittenden) wrote…</i></summary><blockquote>

Is this string accurate?  Isn't`BlockBytesInCache` bytes?

https://github.com/cockroachdb/pebble/blob/47da75f027b9a48878893ed59f0d7d3b429765c8/internal/base/iterator.go#L365-L375

I think the right format should be something closer to:
```
"iterator cache stats: %s hit",
```

so the output looks like:

```
iterator cache stats: 384B hit, 4KiB bytes missed (read time: 11ms)
```

or something like that.  Using the prefix "blocks:" makes me think of the number of 32K pebble pages, not bytes, and I would be led astray in an investigation.
</blockquote></details>

The humanized bytes will include a `B/KiB/MiB` suffix so it can't be confused with a count.

PR #3651, opened 2024-06-10
crl-release-23.2: sstable: add LoadBlockSema option

RaduBerinde:
Add option to use a semaphore to limit the number of block loads in parallel.


nicktrav:
There were some minor diffs in the test output on i386. I have skipped those tests on that architecture. We have the same skips in place on master already.

RaduBerinde:
Thanks!!

PR #3660, opened 2024-06-14
go: bump cockroachdb/errors to v1.11.3

### LINKED ISSUE #3659: Deleted upstream dependency; bump github.com/cockroachdb/errors
Hey there,

We've received a report that our builds are failing https://github.com/ethereum/go-ethereum/issues/29984, and it seems it's a dependency issue originating from Pebble, which in turn originates from `github.com/cockroachdb/errors` v1.11.1 (fixed in https://github.com/cockroachdb/errors/pull/137).

It *appears* that an upstream repo (`sentry-go -> iris`)... just... deleted it's published git version tag. /facepalm

Any chance of bumping the `github.com/cockroachdb/errors` package in mainline pebble to the one that has the fixed dependency, and then tagging a v1.1.1 patch release do we could also pick up the fix?

Thanks

nicktrav:
Closes #3659.


RaduBerinde (approved):
LGTM. We will need to do the same on `crl-release-23.2` if they want a new 1.1.1 tag

RaduBerinde:
Unfortunately this bring in some new dep versions which cause some issues on CRDB master:
https://github.com/cockroachdb/cockroach/pull/125864#issuecomment-2176763442

PR #3666, opened 2024-06-14
Makefile: add gen-bazel target

RaduBerinde:
This change adds a `gen-bazel` target which generates a `WORKSPACE`
file and runs gazelle to generate `BUILD.bazel `files. Generating
these files  allows a local clone of the repository to be used
directly when building Cockroach using the `override_repository` flag,
for example:

```
./dev build short -- --override_repository=com_github_cockroachdb_pebble=~/go/src/github.com/cockroachdb/pebble
```

The new files are added to `.gitignore` so that generating them
doesn't interfere with commits.

We also add a `clean-bazel` target that cleans up these files (useful 
when switching back to an older branch)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,9 @@ mutex.prof
 coverprofile.out
 # Testing artifacts
 meta.*.test
+
+# Bazel files, generated with 'make gen-bazel'.
+/WORKSPACE
+BUILD.bazel
+
+
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ all:
 	@echo "  make crossversion-meta"
 	@echo "  make testcoverage"
 	@echo "  make mod-update"
+	@echo "  make gen-bazel"
 	@echo "  make generate"
 	@echo "  make generate-test-data"
 	@echo "  make clean"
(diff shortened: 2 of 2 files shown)


itsbilal (approved):
LGTM

RaduBerinde:
I'm a little weary of having to run `make gen-bazel` every time some imports change.. I wonder if there's a better way.

RaduBerinde:
I will try something a little different - instead of checking in the new files, I will add them to `.gitignore` and we can run `make gen-bazel` only when we need to use it as a dependency. That way generating the files won't interfere with the commits on that branch.

RaduBerinde:
Updated to add just a new target and `.gitignore` the bazel files.

aadityasondhi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3666#-:-O09nVx2-ZGUoOLX1W04:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jbowens)

RaduBerinde:
TFTRs!

PR #3668, opened 2024-06-18
provider: consult readahead config when initializing handle

RaduBerinde:
We currently retrieve the readahead config when we create a `Readable`
and then we use that config for all read handles. This makes the
settings not responsive, as any tables already in the table cache will
have the `Readable` created.

We change `ReadaheadConfig` to an atomic that is loaded every time we
initialize a read handle.

### DIFF
--- a/objstorage/objstorageprovider/provider.go
+++ b/objstorage/objstorageprovider/provider.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"slices"
 	"sync"
+	"sync/atomic"
 
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/errors/oserror"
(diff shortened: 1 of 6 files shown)


objstorage/objstorageprovider/provider.go
```diff
@@ -135,22 +135,49 @@ type Settings struct {
...
+//     large consecutive chunk of a file.
+//   - Speculative is the type of read-ahead used automatically, when consecutive
+//     reads are detected.
+//
+// The settings can be changed and read atomically.
 type ReadaheadConfig struct {
-	// Informed is the type of read-ahead for operations that are known to read a
-	// large consecutive chunk of a file.
-	Informed ReadaheadMode
+	value atomic.Uint32
```
nicktrav, line 149: nit: do you want to put a comment in here about the bitstring format? e.g. one byte for speculative, another for informed?

nicktrav (approved):
LGTM.

jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3668#-:-O-qaxch99DOgpP1qiUB:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sean-)

PR #3672, opened 2024-06-19
sstable: fix double close of fragmentBlockIter

RaduBerinde:
This change addresses a TODO to fix double closing of the
`fragmentBlockIter`. We can now use a sync.Pool and we can remove the
wrapper type that used a sync pool only for range key iterators.

We add an `invariants.DoubleCloseCheck` type which checks for multiple
close calls in invariant builds. In these builds we sometimes don't
put objects back in the pool, so that this check can run.

We now always call `fragmentBlockIter.init()` (we were missing on a
minor optimization before).

### DIFF
--- a/internal/invariants/off.go
+++ b/internal/invariants/off.go
@@ -9,3 +9,10 @@ package invariants
 
 // Enabled is true if we were built with the "invariants" or "race" build tags.
 const Enabled = false
+
+// DoubleCloseCheck is used to check that objects are not double-closed.
+type DoubleCloseCheck struct{}
+
+// Close panics if called twice on the same object (if we were built with the
+// "invariants" or "race" build tags).
+func (d *DoubleCloseCheck) Close() {}
--- a/internal/invariants/on.go
+++ b/internal/invariants/on.go
@@ -9,3 +9,17 @@ package invariants
 
 // Enabled is true if we were built with the "invariants" or "race" build tags.
 const Enabled = true
+
+// DoubleCloseCheck is used to check that objects are not double-closed.
+type DoubleCloseCheck struct {
+	closed bool
+}
+
+// Close panics if called twice on the same object (if we were built with the
+// "invariants" or "race" build tags).
+func (d *DoubleCloseCheck) Close() {
+	if d.closed {
+		panic("double close")
+	}
+	d.closed = true
+}
--- a/internal/keyspan/iter.go
+++ b/internal/keyspan/iter.go
@@ -55,8 +55,9 @@ type FragmentIterator interface {
 	// previous call to SeekLT or Prev returned an invalid span.
 	Prev() (*Span, error)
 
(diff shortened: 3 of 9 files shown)


jbowens (approved):
nice :lgtm: 

> Reviewed 6 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3672#-:-O-q1idyAdey-YIAdCWD:bwrsmhi)* status: 6 of 9 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`sstable/block_fragment_iter.go` line 215 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3672#-O-q1CM_9Qx81CJkvFaF:-O-q1CM_9Qx81CJkvFaG:b-mxvlq0) ([raw file](https://github.com/cockroachdb/pebble/blob/6b2cd6f2c4051d2077db53b93929aadc5538a779/sstable/block_fragment_iter.go#L215)):*
> ```Go
> 		*i = fragmentBlockIter{
> 			blockIter: i.blockIter.resetForReuse(),
> 		}
> ```

if we zeroed the struct when we take it out of the pool, could we unconditionally call `i.doubleCloseCheck.Close()`? we'd get double close protection for the duration the iter is sitting in the pool. we can still sometimes not add to the pool to check for double closes that take more time between Closes than the duration an object spends in the pool.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3672#-:-O-qTHaD7Exy1P-Eas6I:b-814g8k)* status: 6 of 9 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`sstable/block_fragment_iter.go` line 215 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3672#-O-q1CM_9Qx81CJkvFaF:-O-qSt77DZpwL9p3y1Y_:b-fznogt) ([raw file](https://github.com/cockroachdb/pebble/blob/6b2cd6f2c4051d2077db53b93929aadc5538a779/sstable/block_fragment_iter.go#L215)):*

if we zeroed the struct when we take it out of the pool, could we unconditionally call `i.doubleCloseCheck.Close()`? we'd get double close protection for the duration the iter is sitting in the pool. we can still sometimes not add to the pool to check for double closes that take more time between Closes than the duration an object spends in the pool.
</blockquote></details>

Good idea, done.

jbowens (approved):
:lgtm:

> Reviewed 2 of 9 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3672#-:-O-qbgUFAyIEEnTm017d:bnfp4nl)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @RaduBerinde)

PR #3676, opened 2024-06-20
internal/binfmt: new package

jbowens:
Add new internal binfmt package for formatting binary data in a descriptive
way, suitable for datadriven tests.

----

This is motivated from the columnar block format unit tests which make use of it for pretty-printing the binary formats. I think it could make sense to pull this into the datadriven package itself eventually, but maybe we should wait. I expect we may want to iterate it on it in the short term.

### DIFF
--- a/batchrepr/testdata/writer
+++ b/batchrepr/testdata/writer
@@ -9,8 +9,9 @@ read-header
 
 set-seqnum 1052
 ----
-1c04000000000000 01000000 # [seqNum=1052,count=1]
-00 0161                   # DEL "a"
+00-08: x 1c04000000000000 # seqnum=1052
+08-12: x 01000000         # count=1
+12-15: x 00 0161          # DEL "a"
 
 read-header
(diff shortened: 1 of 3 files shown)


RaduBerinde (approved):
:lgtm:

Cool! Consider adding the binfmt stuff to `datadriven/diagram` instead if you think it might be useful in other contexts.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3676#-:-O-qgeNl8UXJ1CZ4Nx88:b-37kmt3)* status: 0 of 3 files reviewed, all discussions resolved

jbowens (commented):
TFTR!

Will do once it's relatively stable

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3676#-:-O-r2sql8UVakzO9rWez:blhscbp)* status: 0 of 3 files reviewed, all discussions resolved

PR #3677, opened 2024-06-20
base: add SeqNum type

RaduBerinde:
We add a `SeqNum` type for sequence numbers. This makes the code more
self-documenting, and will allow a custom `String` implementation.

All changes outside `base/internal.go` were mechanical.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -317,7 +317,7 @@ type batchInternal struct {
 	// tombstonesSeqNum. This is the case for all new iterators created over a
 	// batch but it's not the case for all cloned iterators.
 	tombstones       []keyspan.Span
-	tombstonesSeqNum uint64
+	tombstonesSeqNum base.SeqNum
 
 	// Fragmented range key spans. Cached the first time a range key iterator is
 	// requested. The cache is invalidated whenever a new range key
@@ -326,7 +326,7 @@ type batchInternal struct {
 	// tombstonesSeqNum. This is the case for all new iterators created over a
 	// batch but it's not the case for all cloned iterators.
 	rangeKeys       []keyspan.Span
-	rangeKeysSeqNum uint64
+	rangeKeysSeqNum base.SeqNum
 
 	// The flushableBatch wrapper if the batch is too large to fit in the
 	// memtable.
@@ -487,8 +487,8 @@ func newIndexedBatchWithSize(db *DB, comparer *Comparer, size int) *Batch {
 // with the base.InternalKeySeqNumBatch bit. These sequence numbers are only
 // used during iteration, and the keys are assigned ordinary sequence numbers
 // when the batch is committed.
-func (b *Batch) nextSeqNum() uint64 {
-	return uint64(len(b.data)) | base.InternalKeySeqNumBatch
+func (b *Batch) nextSeqNum() base.SeqNum {
+	return base.SeqNum(len(b.data)) | base.InternalKeySeqNumBatch
 }
 
 func (b *Batch) release() {
@@ -1310,7 +1310,7 @@ func (b *Batch) initInternalIter(o *IterOptions, iter *batchIter) {
 	}
 }
 
-func (b *Batch) newRangeDelIter(o *IterOptions, batchSnapshot uint64) *keyspan.Iter {
+func (b *Batch) newRangeDelIter(o *IterOptions, batchSnapshot base.SeqNum) *keyspan.Iter {
 	// Construct an iterator even if rangeDelIndex is nil, because it is allowed
 	// to refresh later, so we need the container to exist.
 	iter := new(keyspan.Iter)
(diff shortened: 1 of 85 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3677)* status: 0 of 85 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`internal/base/internal.go` line 18 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3677#-O-qiw5u04MRU2O62Y_L:-O-qiw5u04MRU2O62Y_M:b-87l5q5) ([raw file](https://github.com/cockroachdb/pebble/blob/382ef48a4d3149505571699da831f4203c97de10/internal/base/internal.go#L18)):*
> ```Go
> )
> 
> // SeqNum is a sequence number.
> ```

I'd like to include a better definition here, do we have something written somewhere?

jbowens (approved):
:lgtm:

> Reviewed 80 of 85 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3677#-:-O-v5cTg7I-NE5CdcBs5:bnfp4nl)* status: 80 of 85 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`internal/base/internal.go` line 18 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3677#-O-qiw5u04MRU2O62Y_L:-O-v0dh92xYC6y0_xVb_:b8okjp4) ([raw file](https://github.com/cockroachdb/pebble/blob/382ef48a4d3149505571699da831f4203c97de10/internal/base/internal.go#L18)):*

I'd like to include a better definition here, do we have something written somewhere?
</blockquote></details>

I don't think we do... you could say something like:

SeqNum is a sequence number defining precedence among identical keys. A key with a higher sequence number takes precedence over a key with an equal user key of a lower sequence number. Sequence numbers are stored durably within the internal key "trailer" as a 7-byte (uint56) uint, and the maximum sequence number is 2^56-1. As keys are committed to the database, they're assigned increasing sequence numbers. Readers use sequence numbers to read a consistent database state, ignoring keys with sequence numbers larger than the readers' "visible sequence number."

The database maintains an invariant that no two point keys with equal user keys may have equal sequence numbers. Keys with differing user keys may have equal sequence numbers. A key's sequence number may be changed to zero during compactions when it can be proven that no identical keys with lower sequence numbers exist.
___
*[`sstable/testdata/virtual_reader_iter` line 203 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3677#-O-v55FV2VNg5q4C0D3Q:-O-v55FV2VNg5q4C0D3R:bw0n0dd) ([raw file](https://github.com/cockroachdb/pebble/blob/08d8bdd9eac8bb77533e7d919a41d1eb12baf3e2/sstable/testdata/virtual_reader_iter#L203)):*
> ```
> virtualize lower=c.SET.3 upper=f.SET.1
> ----
> bounds:  [c#3,SET-f#1,SET]
> ```

do you know where this diff came from?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3677)* status: 80 of 85 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`internal/base/internal.go` line 18 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3677#-O-qiw5u04MRU2O62Y_L:-O-vgKSE0YdGE_gh4E7U:bofkasy) ([raw file](https://github.com/cockroachdb/pebble/blob/382ef48a4d3149505571699da831f4203c97de10/internal/base/internal.go#L18)):*

I don't think we do... you could say something like:

SeqNum is a sequence number defining precedence among identical keys. A key with a higher sequence number takes precedence over a key with an equal user key of a lower sequence number. Sequence numbers are stored durably within the internal key "trailer" as a 7-byte (uint56) uint, and the maximum sequence number is 2^56-1. As keys are committed to the database, they're assigned increasing sequence numbers. Readers use sequence numbers to read a consistent database state, ignoring keys with sequence numbers larger than the readers' "visible sequence number."

The database maintains an invariant that no two point keys with equal user keys may have equal sequence numbers. Keys with differing user keys may have equal sequence numbers. A key's sequence number may be changed to zero during compactions when it can be proven that no identical keys with lower sequence numbers exist.
</blockquote></details>

Very nice, I will add that! Thanks!
___
*[`sstable/testdata/virtual_reader_iter` line 203 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3677#-O-v55FV2VNg5q4C0D3Q:-O-vg7tv9m3hpH1UbbpD:b-4vt39r) ([raw file](https://github.com/cockroachdb/pebble/blob/08d8bdd9eac8bb77533e7d919a41d1eb12baf3e2/sstable/testdata/virtual_reader_iter#L203)):*

do you know where this diff came from?
</blockquote></details>

Yes, the accidental extra `:ff` on the upper key before was causing a silent parsing error and we had seq num 0.

PR #3679, opened 2024-06-21
sstable: wrap raw fragment iterators with checkers, fix issues found

RaduBerinde:
#### tool: use DeepClone instead of ShallowClone

We were performing shallow clones of spans, which is not sufficient
when the spans come from a FragmentIterator.

#### sstable: wrap raw fragment iterators with checkers

This change wraps raw range del and range key iterators with
`assertIter` and `invalidatingIter` and we fix a few cases of using
stale values (especially Prefix/Value for range keys).

### DIFF
--- a/internal/keyspan/assert_iter.go
+++ b/internal/keyspan/assert_iter.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/fastrand"
 	"github.com/cockroachdb/pebble/internal/invariants"
 )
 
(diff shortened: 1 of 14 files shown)


jbowens (approved):
:lgtm:

nice

> Reviewed 1 of 1 files at r1, 6 of 6 files at r2, 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3679#-:-O-v7YWU9rFiPLobu0-V:b-jwumhe)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/keyspan/span.go` line 16 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3679#-O-v6kNF8Z651LzRGivL:-O-v6kNF8Z651LzRGivM:bjwegbt) ([raw file](https://github.com/cockroachdb/pebble/blob/4d1d88290551d89aa441fe8602387b3d45eb3c2e/internal/keyspan/span.go#L16)):*
> ```Go
> 
> 	"github.com/cockroachdb/pebble/internal/base"
> 	"golang.org/x/exp/slices"
> ```

this can just be "slices" now that it's part of the stdlib, right?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3679#-:-O-x8n8N28FEq65fB-is:b-814g8k)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`internal/keyspan/span.go` line 16 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3679#-O-v6kNF8Z651LzRGivL:-O-x8LiI4nWIsV2OeAbN:b-5gq9p8) ([raw file](https://github.com/cockroachdb/pebble/blob/4d1d88290551d89aa441fe8602387b3d45eb3c2e/internal/keyspan/span.go#L16)):*

this can just be "slices" now that it's part of the stdlib, right?
</blockquote></details>

Nice catch, not sure why Goland chooses that one. I added it to the forbidden imports linter.

Issue #3678, opened 2024-06-20
TestMetaCrossVersion and TestMeta with no invariants failure


RaduBerinde:
https://github.com/cockroachdb/pebble/actions/runs/9601942213/job/26481618161?pr=3677

```
        -iter20.SeekLT("dlrsqhqtkugm@10", "") // [true,"cearmwwagb@15","ojtyravuafvuep",["cearmwwagb","gsohc")=>{"@13"="jkzofmypok"}*] <nil> #1700
        +iter20.SeekLT("dlrsqhqtkugm@10", "") // [true,"cearmwwagb@3","dwyydffdbimw",["cearmwwagb","gsohc")=>{"@13"="jkzofmypok"}*] <nil> #1700
```

```
SEED=1718905245421630211 FACTOR=10 ./scripts/run-crossversion-meta.sh crl-release-24.1 HEAD
```

RaduBerinde:
Couldn't reproduce with the command above :/

RaduBerinde:
Possibly introduced by https://github.com/cockroachdb/pebble/pull/3672

RaduBerinde:
I couldn't reproduce myself, but I found some cases of using FragmentIterator spans after Close which may be interacting with #3672

RaduBerinde:
This was very tricky to debug. The finding here is that there is code that compares fragment iterators by pointer to see if there was a change. This doesn't work with a pool.
https://github.com/cockroachdb/pebble/blob/master/merging_iter.go#L544

RaduBerinde:
#3689 fixed the main underlying issue. #3697 will close.

PR #3680, opened 2024-06-21
github: don't use -v flag for windows test run

RaduBerinde:
The windows test run is the only one that uses the `-v` flag which makes the output hard to use in the UI and likely makes the run much slower.

Adjust to not use `-v` and use `-tags invariants`.


itsbilal (approved):
LGTM

RaduBerinde:
Meta CI failures are tracked in #3678.

RaduBerinde:
TFTRs!

PR #3688, opened 2024-06-24
sstable/block: new package

jbowens:
Move a few sstable block-related types into a new sstable/block package. This is in preparation for moving the row-based block iterator and writer into a sstable/rowblk package.


jbowens:
TFTR!

PR #3689, opened 2024-06-24
db: change leveliter.rangeDelIterPtr to a function

RaduBerinde:
`rangeDelIterPtr` is a mechanism through which the level iter passes
a range deletion iterator to the merge iterator (which uses it to
maintain a heap of tombstones).

This change cleans up the mechanism by making it a function instead.
This makes the ownership much clearer and allows the caller to add
more logic whenever the iterator changes. The merge iterator now
maintains a generation counter to detect changes (instead of comparing
by pointer, which doesn't work when iterators are pooled).

### DIFF
--- a/db.go
+++ b/db.go
@@ -1479,7 +1479,7 @@ func (i *Iterator) constructPointIter(
 			li := &levels[levelsIndex]
 
 			li.init(ctx, i.opts, &i.comparer, i.newIters, files, level, internalOpts)
-			li.initRangeDel(&mlevels[mlevelsIndex].rangeDelIter)
+			li.initRangeDel(mlevels[mlevelsIndex].setRangeDelIter)
 			li.initCombinedIterState(&i.lazyCombinedIter.combinedIterState)
 			mlevels[mlevelsIndex].levelIter = li
 			mlevels[mlevelsIndex].iter = invalidating.MaybeWrapIfInvariants(li)
--- a/level_checker.go
+++ b/level_checker.go
@@ -54,6 +54,14 @@ type simpleMergingIterLevel struct {
 	tombstone *keyspan.Span
 }
 
+func (ml *simpleMergingIterLevel) setRangeDelIter(iter keyspan.FragmentIterator) {
+	ml.tombstone = nil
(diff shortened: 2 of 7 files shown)


jbowens (commented):
> Reviewed 5 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3689)* status: 5 of 7 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`level_iter.go` line 904 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3689#-O0F8L07CKgo1uU-2uQ3:-O0F8L07CKgo1uU-2uQ4:bw8hogi) ([raw file](https://github.com/cockroachdb/pebble/blob/c324bc294be6ceba9d49f13b314a0b0b4a297aa7/level_iter.go#L904)):*
> ```Go
> 	//if l.rangeDelIterFn != nil {
> 	//	l.rangeDelIterFn(nil)
> 	//}
> ```

should these be uncommented?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3689)* status: 5 of 7 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`level_iter.go` line 904 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3689#-O0F8L07CKgo1uU-2uQ3:-O0FCO057SojtXP9xHGG:b-s7u0gb) ([raw file](https://github.com/cockroachdb/pebble/blob/c324bc294be6ceba9d49f13b314a0b0b4a297aa7/level_iter.go#L904)):*

should these be uncommented?
</blockquote></details>

Sorry, forgot to remove these. No, it's not correct because we no longer use `rangeDelIterCopy` to set them back if we stay with the same file on the next operation. It's cleaner if the last `rangeDelIterFn()` call always corresponds to the currently opened file.

jbowens (approved):
lgtm

RaduBerinde:
TFTR!

PR #3700, opened 2024-06-25
sstable/rowblk: move Iter, RawIter

jbowens:
Move the blockIter and rawBlockIter into the rowblk package.


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3700#-:-O0GqzzeCMjOuC13Wpz1:bnfp4nl)* status: 0 of 19 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/rowblk/rowblk_iter.go` line 39 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3700#-O0GqEDI-ozgeW3uaqLG:-O0GqEDI-ozgeW3uaqLH:biyfbb1) ([raw file](https://github.com/cockroachdb/pebble/blob/31891b4294c1560c617d05ca38a1323a7f6a4b7c/sstable/rowblk/rowblk_iter.go#L39)):*
> ```Go
> // sufficient for range tombstones and range deletes as they are always encoded
> // in a single block. Note: this stability guarantee no longer holds for a block
> // iter with synthetic suffix replacement, but this doesn't matter, as the user
> ```

[nit] synthetic prefix/suffix replacement.

The synthetic suffix/prefix will be applied in the `fragmentBlockIter` (the layer "above"). Maybe just say "but we don't use the synthetic suffix/prefix functionality of Iter for range keys".

jbowens:
TFTR!

PR #3701, opened 2024-06-25
crdbtest: tolerate empty keys in the comparer

RaduBerinde:
The compaction picker compares keys to the empty key, so it must be
supported by the Compare function. We recently made this function more
strict but we have to account for the empty key case.

This was causing some nightly benchmarks to error out.


itsbilal (approved):
LGTM

RaduBerinde:
TFTR!

PR #3702, opened 2024-06-26
sstable: use MaybeScanArgs for global seq num

jbowens:
Use the datadriven package's MaybeScanArgs helper to scan the global sequence number, removing some test code duplication.


RaduBerinde (approved):

jbowens:
TFTR!

PR #3708, opened 2024-06-26
sstable: add FragmentIterTransforms

RaduBerinde:
We add a separate struct for transforms passed to the fragment
iterator. This includes the option to elide keys at the same sequence
number.

We also move the underlying `block.Iter` handle initialization to the
fragment iterator constructor. It's up to the iterator to decide what
transformations to pass to that "raw" iterator. For example, prefix
and suffix synthesis will happen in the `FragmentIter` and these
options won't be passed to the `block.Iter`.

### DIFF
--- a/external_iterator.go
+++ b/external_iterator.go
@@ -208,7 +208,9 @@ func createExternalPointIter(ctx context.Context, it *Iterator) (topLevelIterato
 			if err != nil {
 				return nil, err
 			}
-			rangeDelIter, err = r.NewRawRangeDelIter(transforms)
+			rangeDelIter, err = r.NewRawRangeDelIter(sstable.FragmentIterTransforms{
+				SyntheticSeqNum: sstable.SyntheticSeqNum(seqNum),
+			})
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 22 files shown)


jbowens (approved):
:lgtm:

Should we consider applying suffix+prefix rangedel/rangekey transformations as block transformations, so that the transformed block sits in the block cache?

> Reviewed 21 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3708#-:-O0LSIh6EyVYHVhk_lzr:b-y4srja)* status: 21 of 22 files reviewed, all discussions resolved

RaduBerinde (commented):
Interesting point. I'll keep it in mind. It feels easier to do it at the `keyspan.Span` level and I'm not sure we lose much perf-wise.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3708#-:-O0LcPii1vnCKvnaPlrv:bz5w8e7)* status: 21 of 22 files reviewed, all discussions resolved (waiting on @jbowens)

PR #3711, opened 2024-06-27
colblk: add Bitmap, BitmapBuilder

jbowens:
Introduce the sstable/colblk package for columnar block primitives, and add
a first primitive for storing bitmaps.

The Bitmap uses 1+1/64 physical bits per logical bit. Bits are organized into
64-bit words, allowing constant-time access of an individual bit. Additionally,
after the bitmap a summary bitmap consisting of 1 bit per 64-bit word of the
bitmap provides faster lookup of preceding or successive set bits. This will be
used by a bitmap indicating when a key prefix changes to quickly find the next
key within a block with a new prefix.


RaduBerinde (commented):
Nice! Some comments for the bitmap stuff for now

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3711#-:-O0QJdzt6jOeSwjQxy2u:b-3r4qs7)* status: 0 of 6 files reviewed, 9 unresolved discussions (waiting on @jbowens, @petermattis, and @sumeerbhola)

___
*[`sstable/colblk/base.go` line 10 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QG6Pt9zxg0tidakmY:-O0QG6Pt9zxg0tidakmZ:b-u6e7o2) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/base.go#L10)):*
> ```Go
> import "golang.org/x/exp/constraints"
> 
> func align[T constraints.Integer](offset, val T) T {
> ```

[nit] could use some function comments. Is it the first value >= offset that is a multiple of val (assuming val is a power of two)?
___
*[`sstable/colblk/base.go` line 15 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QG_ru7438AlFImb8L:-O0QG_rvAi3f-8pXWBNu:b-rcq2rm) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/base.go#L15)):*
> ```Go
> 
> func alignWithZeroes[T constraints.Integer](buf []byte, offset, val T) T {
> 	aligned := (offset + val - 1) & ^(val - 1)
> ```

[nit] could use the previously defined aligned
___
*[`sstable/colblk/base.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QGoH62cYhGK0QQAsG:-O0QGoH7AZWj-fWHV0aI:b-nalx3g) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/base.go#L28)):*
> ```Go
> )
> 
> // When multiplying or dividing by align{16,32,64}, it's faster to shift to the
> ```

I don't think this is really true. If we are multiplying or dividing by a constant power of two, the compiler knows it can use a shift instead.
___
*[`sstable/colblk/bitmap.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QHxSG7ieAZaTfQyIR:-O0QHxSG7ieAZaTfQyIS:bqqs387) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/bitmap.go#L28)):*
> ```Go
> type Bitmap struct {
> 	data  UnsafeRawSlice[uint64]
> 	total int
> ```

[nit] `size` might be more appropriate
___
*[`sstable/colblk/bitmap.go` line 30 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QI4zM5S98_3HQtAnY:-O0QI4zM5S98_3HQtAnZ:b-n0nyf2) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/bitmap.go#L30)):*
> ```Go
> 	total int
> }
> ```

[nit] We should have a constructor here.
___
*[`sstable/colblk/bitmap.go` line 37 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QHzON3j8p8yOI5Ps5:-O0QHzON3j8p8yOI5Ps6:b-ulwzvw) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/bitmap.go#L37)):*
> ```Go
> 
> // Successor returns the next bit greater than or equal to i set in the bitmap.
> // Returns b.total if no next bit is set.
> ```

[nit] we shouldn't mention private members, maybe just say "size".
___
*[`sstable/colblk/bitmap.go` line 74 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QIe2d2CEa8rGDOXnP:-O0QIe2d2CEa8rGDOXnQ:bi0nf25) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/bitmap.go#L74)):*
> ```Go
> 	// words that are summarized together have a set bit. We call [nextInWord]
> 	// on the summary word to get the index of which word has a set bit, if any.
> 	summaryTableOffset := (b.total + 63) >> 6
> ```

[nit] this and the summaryTableEnd calculation should be separate methods
___
*[`sstable/colblk/bitmap.go` line 107 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QJ4tY57MnMM3tyNkK:-O0QJ4tY57MnMM3tyNkL:br3qa5l) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/bitmap.go#L107)):*
> ```Go
> 	// provided word. The returned index is an index local to the word. Returns
> 	// -1 if no set bit is found.
> 	prevInWord := func(word uint64, bit uint) int {
> ```

[nit] could be a method, and take `wordIdx`
___
*[`sstable/colblk/bitmap.go` line 176 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QJGWa6jE9LjC5fWu8:-O0QJGWa6jE9LjC5fWu9:b-hpv712) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/bitmap.go#L176)):*
> ```Go
> 
> // BitmapBuilder constructs a Bitmap. Bits are default false.
> type BitmapBuilder []uint64
> ```

[nit] can this be a struct with a slice member, so that functions like `Set()` can just modify that directly?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3711)* status: 0 of 6 files reviewed, 9 unresolved discussions (waiting on @petermattis, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk/base.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QGoH62cYhGK0QQAsG:-O0QPlnDApz6vi9j3CEu:blyv59y) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/base.go#L28)):*

I don't think this is really true. If we are multiplying or dividing by a constant power of two, the compiler knows it can use a shift instead.
</blockquote></details>

Yeah, I don't understand _why_ it's true, but empirically on my mac switching to explicit shifts made a significant speed up. I didn't dig deeper but maybe the inline complexity analysis happens _before_ applying the shift optimization and some things were being inlined that previously weren't? It's hard to imagine the compiler not having this optimization. I'll revisit this to try to isolate more

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3711)* status: 0 of 6 files reviewed, 9 unresolved discussions (waiting on @jbowens, @petermattis, and @sumeerbhola)

___
*[`sstable/colblk/base.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QGoH62cYhGK0QQAsG:-O0Qd4Rr7QMawZCiebrk:bsu850p) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/base.go#L28)):*

Yeah, I don't understand _why_ it's true, but empirically on my mac switching to explicit shifts made a significant speed up. I didn't dig deeper but maybe the inline complexity analysis happens _before_ applying the shift optimization and some things were being inlined that previously weren't? It's hard to imagine the compiler not having this optimization. I'll revisit this to try to isolate more
</blockquote></details>

Interesting. I think it's because we're using it with `int` which is signed and it needs to do a bit more than a simple shift:
https://go.dev/play/p/QDR8yPSTR5K

https://godbolt.org/z/dh4odMWPG

The unsigned variants are using the same generated code.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3711#-:-O0S3UXw5rEEHu3RVYAH:bnfp4nl)* status: 0 of 6 files reviewed, 19 unresolved discussions (waiting on @jbowens, @petermattis, and @sumeerbhola)

___
*[`sstable/colblk/bitmap.go` line 103 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S-jLa7XR5bDwKTT40:-O0S-jLb0LhxBCaUyCWs:btc6wqr) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/bitmap.go#L103)):*
> ```Go
> // Predecessor returns the previous bit less than or equal to i set in the
> // bitmap. Returns -1 if no previous bit is set.
> func (b Bitmap) Predecessor(i int) int {
> ```

[nit] Is `i` required to be in `[0, total)`? same for Successor
___
*[`sstable/colblk/bitmap.go` line 188 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S2lsS36YbXasc5etb:-O0S2lsT-tyaVIJtLjYI:btmpm3q) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/bitmap.go#L188)):*
> ```Go
> // Set sets the bit at position i if v is true and clears the bit at position i
> // otherwise. Callers need not call Set if v is false and Set(i, true) has not
> // been set yet.
> ```

[nit] not been called yet
___
*[`sstable/colblk/bitmap.go` line 203 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S2Z15D7DJUMu7UWqi:-O0S2Z15D7DJUMu7UWqj:bmjcyi5) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/bitmap.go#L203)):*
> ```Go
> // Reset resets the bitmap to the empty state.
> func (b *BitmapBuilder) Reset() {
> 	for i := range b.words {
> ```

[nit] clear(b.words)
___
*[`sstable/colblk/bitmap.go` line 224 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S2410-nemgZ0YsWTO:-O0S2410-nemgZ0YsWTP:bjuqwhe) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/bitmap.go#L224)):*
> ```Go
> 	// If the tail of b is sparse, fill in zeroes before inverting.
> 	nBitmapWords := (nRows + 63) >> 6
> 	for i := len(b.words); i < nBitmapWords; i++ {
> ```

[nit] Could do `b.words = slices.Grow(b.words, nBitmapWords-len(b.words))[:nBitmapWords]`. Maybe same in `Set`. It might eliminate some unnecessary doublings when we append a lot, though I haven't run a benchmark.
___
*[`sstable/colblk/column.go` line 39 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S30qY9ZvZVl7lHB0t:-O0S30qY9ZvZVl7lHB0u:b-rtko2k) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/column.go#L39)):*
> ```Go
> 	// lexicographically-sorted byte strings, with prefix compression.
> 	DataTypePrefixBytes DataType = 7
> 	dataTypesCount      DataType = 8
> ```

[nit] mention that we can't go above 8 without further changes
___
*[`sstable/colblk/column.go` line 45 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S35yt6ph2QpQir6aE:-O0S35yt6ph2QpQir6aF:bb7psyn) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/column.go#L45)):*
> ```Go
> 	DataTypeInvalid:     "invalid",
> 	DataTypeBool:        "bool",
> 	DataTypeUint8:       "int8",
> ```

uint
___
*[`sstable/colblk/column.go` line 164 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S0jXQ5eCfHpUNwfcL:-O0S0jXQ5eCfHpUNwfcM:bhxok16) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/column.go#L164)):*
> ```Go
> 	// should be encoded. Finish also returns a column descriptor describing the
> 	// encoding of the column, which will be serialized within the block header.
> 	//
> ```

Mention that `buf` has to have enough space (would the caller use `Size()` beforehand to ensure that?)
___
*[`sstable/colblk/column.go` line 167 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S0aMaD2boY6XHsiYG:-O0S0aMaD2boY6XHsiYH:bwjcpss) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/column.go#L167)):*
> ```Go
> 	// The provided column index must be less than NumColumns(). Finish is
> 	// called for each index < NumColumns() in order.
> 	Finish(col int, rows int, offset uint32, buf []byte) (uint32, ColumnDesc)
> ```

[nit] name the first return `nextOffset`
___
*[`sstable/colblk/unsafe_slice.go` line 19 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S1W_HC8gKzDHHtHtM:-O0S1W_HC8gKzDHHtHtN:b6hdy8s) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/unsafe_slice.go#L19)):*
> ```Go
> }
> 
> func makeUnsafeRawSlice[T constraints.Integer](ptr unsafe.Pointer) UnsafeRawSlice[T] {
> ```

[nit] this can take a `T* firstElem` instead of an `unsafe.Pointer`
___
*[`sstable/colblk/bitmap_test.go` line 135 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S1Ni63Kk73x9TOesv:-O0S1Ni755Sz6ECVzEek:bw15xzz) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/bitmap_test.go#L135)):*
> ```Go
> 			}
> 			// Ensure there are no set bits between pred and i.
> 			for j := max(0, pred+1); j < i; j++ {
> ```

[nit] pred is at least -1, no?

petermattis (commented):
Just a quick skim. Deferring to Radu's :lgtm:, but had one small comment about `ColumnDesc`.

> Reviewed 1 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3711#-:-O0TzsA11QQDbMU-olQF:b9cisbi)* status: 1 of 6 files reviewed, 21 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/colblk/column.go` line 94 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0TyutL-lJJ9TWw8HqW:-O0TyutM8jr03zsmdv9Q:bozuro1) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/column.go#L94)):*
> ```Go
> // The bits at positions 5, 6, and 7 encode an enum describing the encoding of
> // the column. See the ColumnEncoding type and its constants for details.
> type ColumnDesc uint8
> ```

I wonder if we're trying to be overly parsimonious with space in storing the column description in a single byte. Blocks are many KBs in size, and the number of columns will be a small fixed number, at least as currently envisioned. 
___
*[`sstable/colblk/unsafe_slice.go` line 20 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0Tzh13EIQKPj_P8Bzb:-O0Tzh144mwE1ggNHUCY:bp6ig04) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/unsafe_slice.go#L20)):*
> ```Go
> 
> func makeUnsafeRawSlice[T constraints.Integer](ptr unsafe.Pointer) UnsafeRawSlice[T] {
> 	// TODO(peter): verify alignment
> ```

:grin: Maybe worth fixing. Or removing.

jbowens (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3711#-:-O0e6lCzE83giudiOH1s:bz5mair)* status: 0 of 6 files reviewed, 9 unresolved discussions (waiting on @petermattis, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk/base.go` line 10 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QG6Pt9zxg0tidakmY:-O0Vv_z_2myFiTy_dsno:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/base.go#L10)):*

[nit] could use some function comments. Is it the first value >= offset that is a multiple of val (assuming val is a power of two)?
</blockquote></details>

done
___
*[`sstable/colblk/base.go` line 15 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QG_ru7438AlFImb8L:-O0Vvb6HA6IJocg0F07B:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/base.go#L15)):*

[nit] could use the previously defined aligned
</blockquote></details>

done
___
*[`sstable/colblk/base.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QGoH62cYhGK0QQAsG:-O0Vw1YdEpmYIzssclMs:bt3qv39) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/base.go#L28)):*

Interesting. I think it's because we're using it with `int` which is signed and it needs to do a bit more than a simple shift:
https://go.dev/play/p/QDR8yPSTR5K

https://godbolt.org/z/dh4odMWPG

The unsigned variants are using the same generated code.
</blockquote></details>

Nice, thanks. I added that context and a TODO to consider using unsigned ints here without the explicit shifts
___
*[`sstable/colblk/bitmap.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QHxSG7ieAZaTfQyIR:-O0R07fm7zJJ6XGd0UDz:b-fwx8h7) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/bitmap.go#L28)):*

[nit] `size` might be more appropriate
</blockquote></details>

updated to `bitCount`, wdyt? worried size could be confused with the serialized size of the bitmap
___
*[`sstable/colblk/bitmap.go` line 30 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QI4zM5S98_3HQtAnY:-O0VwCd6DA_i48oZ8Lb0:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/bitmap.go#L30)):*

[nit] We should have a constructor here.
</blockquote></details>

done
___
*[`sstable/colblk/bitmap.go` line 37 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QHzON3j8p8yOI5Ps5:-O0R0b9O5lSq0e8-f9bD:bwg0dnl) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/bitmap.go#L37)):*

[nit] we shouldn't mention private members, maybe just say "size".
</blockquote></details>

reworded
___
*[`sstable/colblk/bitmap.go` line 74 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QIe2d2CEa8rGDOXnP:-O0e1C_E7YNC5VS0-dFC:b-ktn8w5) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/bitmap.go#L74)):*

[nit] this and the summaryTableEnd calculation should be separate methods
</blockquote></details>

Pulled them into a single method that returns the bounds of the summary table
___
*[`sstable/colblk/bitmap.go` line 107 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QJ4tY57MnMM3tyNkK:-O0e1YlI6NQ8tVIUpJnA:b-7fkrht) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/bitmap.go#L107)):*

[nit] could be a method, and take `wordIdx`
</blockquote></details>

Left it as an inline function since we only use it here. I also left it taking the word itself—when we're scanning the summary table, we've already loaded the word to check if != 0.
___
*[`sstable/colblk/bitmap.go` line 176 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QJGWa6jE9LjC5fWu8:-O0e2LgF8IU2uwL0PEAc:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/bitmap.go#L176)):*

[nit] can this be a struct with a slice member, so that functions like `Set()` can just modify that directly?
</blockquote></details>

done
___
*[`sstable/colblk/bitmap.go` line 103 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S-jLa7XR5bDwKTT40:-O0e1X1P7rp-zqGKBEpt:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/bitmap.go#L103)):*

[nit] Is `i` required to be in `[0, total)`? same for Successor
</blockquote></details>

done
___
*[`sstable/colblk/bitmap.go` line 188 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S2lsS36YbXasc5etb:-O0e2UR66_lfBTlvvIoY:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/bitmap.go#L188)):*

[nit] not been called yet
</blockquote></details>

done
___
*[`sstable/colblk/bitmap.go` line 203 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S2Z15D7DJUMu7UWqi:-O0e2V0EEe8wf67uY2aw:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/bitmap.go#L203)):*

[nit] clear(b.words)
</blockquote></details>

done
___
*[`sstable/colblk/bitmap.go` line 224 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S2410-nemgZ0YsWTO:-O0e2Vrd-mkMeadcs_DL:b3b3usr) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/bitmap.go#L224)):*

[nit] Could do `b.words = slices.Grow(b.words, nBitmapWords-len(b.words))[:nBitmapWords]`. Maybe same in `Set`. It might eliminate some unnecessary doublings when we append a lot, though I haven't run a benchmark.
</blockquote></details>

done (but not in `Set`)
___
*[`sstable/colblk/column.go` line 39 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S30qY9ZvZVl7lHB0t:-O0e2bgP8NGoZ7HbmviE:bb5jgvm) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/column.go#L39)):*

[nit] mention that we can't go above 8 without further changes
</blockquote></details>

made obsolete by separation of the data type and encoding
___
*[`sstable/colblk/column.go` line 45 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S35yt6ph2QpQir6aE:-O0e2fzt09Oe6MHALIk-:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/column.go#L45)):*

uint
</blockquote></details>

done
___
*[`sstable/colblk/column.go` line 94 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0TyutL-lJJ9TWw8HqW:-O0e3r74AOzTSFlKbTJJ:b-mqfw85) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/column.go#L94)):*

I wonder if we're trying to be overly parsimonious with space in storing the column description in a single byte. Blocks are many KBs in size, and the number of columns will be a small fixed number, at least as currently envisioned.
</blockquote></details>

Yeah, I think you're right. I think I'll split it into two bytes: one encoding the data type and the other the encoding.
___
*[`sstable/colblk/column.go` line 164 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S0jXQ5eCfHpUNwfcL:-O0e3i9_4GaFg-9MIqe0:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/column.go#L164)):*

Mention that `buf` has to have enough space (would the caller use `Size()` beforehand to ensure that?)
</blockquote></details>

done
___
*[`sstable/colblk/column.go` line 167 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S0aMaD2boY6XHsiYG:-O0e3idw9-vgRhhQ2RDF:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/column.go#L167)):*

[nit] name the first return `nextOffset`
</blockquote></details>

done
___
*[`sstable/colblk/unsafe_slice.go` line 19 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S1W_HC8gKzDHHtHtM:-O0e3jWK98vtUnBSy-xs:b-5b0gi2) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/unsafe_slice.go#L19)):*

[nit] this can take a `T* firstElem` instead of an `unsafe.Pointer`
</blockquote></details>

Left as `unsafe.Pointer` because that's what we have available at call sites
___
*[`sstable/colblk/unsafe_slice.go` line 20 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0Tzh13EIQKPj_P8Bzb:-O0e3osPBeZUOIhC9h7c:b-7ql1td) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/unsafe_slice.go#L20)):*

:grin: Maybe worth fixing. Or removing.
</blockquote></details>

fixed
___
*[`sstable/colblk/bitmap_test.go` line 135 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0S1Ni63Kk73x9TOesv:-O0e3pz6BOpERb11PWF0:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/bitmap_test.go#L135)):*

[nit] pred is at least -1, no?
</blockquote></details>

done

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3711)* status: 0 of 6 files reviewed, 9 unresolved discussions (waiting on @jbowens, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk/column.go` line 94 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0TyutL-lJJ9TWw8HqW:-O0eBO9_EEA3FG3yUaLU:bs6vssz) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/column.go#L94)):*

Yeah, I think you're right. I think I'll split it into two bytes: one encoding the data type and the other the encoding.
</blockquote></details>

This looks better. One additional thought is whether there should be something reserved for future extensibility. For example, the ColumnDesc could be variable length. Not sure I'd go down that path. I'll leave it for you and Radu to discuss. Could easily over-engineer an extensibility mechanism and then never use it.

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3711#-:-O0epj-QEgYzAr841Z1Y:bnfp4nl)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @jbowens, @petermattis, and @sumeerbhola)

___
*[`sstable/colblk/bitmap.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0QHxSG7ieAZaTfQyIR:-O0eoHTF-WsMJR035lD9:b-57cn97) ([raw file](https://github.com/cockroachdb/pebble/blob/4379d6b0f563a8c66bf3144d41d72119d518838e/sstable/colblk/bitmap.go#L28)):*

updated to `bitCount`, wdyt? worried size could be confused with the serialized size of the bitmap
</blockquote></details>

:thumbsup: :thumbsup: 
___
*[`sstable/colblk/bitmap.go` line 36 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0eoJda3FIQ-dqnVcNY:-O0eoJda3FIQ-dqnVcNZ:b-j54oaw) ([raw file](https://github.com/cockroachdb/pebble/blob/389d2f8d6d51684026441d8eaffa658121650394/sstable/colblk/bitmap.go#L36)):*
> ```Go
> // the bitmap is appropriately sized and the provided bitCount correctly
> // identifies the number of bits in the bitmap.
> func MakeBitmap(ptr unsafe.Pointer, bitCount int) Bitmap {
> ```

[nit] can this take a `[]byte` sllice instead? It could check the size instead of relying on the caller. It would also be nice to localize the `unsafe` usage as much as possible.
___
*[`sstable/colblk/column.go` line 94 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0TyutL-lJJ9TWw8HqW:-O0epHjW9wPTjunVK-N5:b6v201j) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/column.go#L94)):*

This looks better. One additional thought is whether there should be something reserved for future extensibility. For example, the ColumnDesc could be variable length. Not sure I'd go down that path. I'll leave it for you and Radu to discuss. Could easily over-engineer an extensibility mechanism and then never use it.
</blockquote></details>

LGTM. Maybe leave a TODO to consider future extensibility, which we could come back later before the format is finalized.
___
*[`sstable/colblk/column.go` line 131 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0epUWk4hykaLhrnpCw:-O0epUWk4hykaLhrnpCx:b52m0fw) ([raw file](https://github.com/cockroachdb/pebble/blob/389d2f8d6d51684026441d8eaffa658121650394/sstable/colblk/column.go#L131)):*
> ```Go
> 	//
> 	// The supplied buf must have enough space at the provided offset to fit the
> 	// column. The caller may pass Size to calculate the exact size required. If
> ```

[nit] may use Size()

jbowens (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3711#-:-O0evcvnAoHi5iZEsEhg:bz5mair)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @petermattis, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk/bitmap.go` line 36 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0eoJda3FIQ-dqnVcNY:-O0esgrdARtZyOU7AjEi:b-4vtk0t) ([raw file](https://github.com/cockroachdb/pebble/blob/389d2f8d6d51684026441d8eaffa658121650394/sstable/colblk/bitmap.go#L36)):*

[nit] can this take a `[]byte` sllice instead? It could check the size instead of relying on the caller. It would also be nice to localize the `unsafe` usage as much as possible.
</blockquote></details>

I think we can take a `[]byte`, but we can't infer the logical bit count since the physical representation is the same every 0-63 bits.
___
*[`sstable/colblk/column.go` line 94 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0TyutL-lJJ9TWw8HqW:-O0etbNJBF8LtIarN298:bqb9cc0) ([raw file](https://github.com/cockroachdb/pebble/blob/3d938b19ab15cb3c46de9f86ee09045eaa13753d/sstable/colblk/column.go#L94)):*

LGTM. Maybe leave a TODO to consider future extensibility, which we could come back later before the format is finalized.
</blockquote></details>

In the larger prototype branch that includes block writing/reading, I include a single byte per-block to represent the block encoding. I'll leave a TODO and we can discuss in whatever PR eventually resolves that TODO
___
*[`sstable/colblk/column.go` line 131 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0epUWk4hykaLhrnpCw:-O0euKhV4y974sdj7YAA:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/389d2f8d6d51684026441d8eaffa658121650394/sstable/colblk/column.go#L131)):*

[nit] may use Size()
</blockquote></details>

Done

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3711)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @jbowens, @petermattis, and @sumeerbhola)

___
*[`sstable/colblk/bitmap.go` line 36 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0eoJda3FIQ-dqnVcNY:-O0ez_EjBdgRiixiMwW7:bpoaibl) ([raw file](https://github.com/cockroachdb/pebble/blob/389d2f8d6d51684026441d8eaffa658121650394/sstable/colblk/bitmap.go#L36)):*

I think we can take a `[]byte`, but we can't infer the logical bit count since the physical representation is the same every 0-63 bits.
</blockquote></details>

Yeah I meant we'd still take the `bitCount` but assert that `len(b)` is enough

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3711)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @petermattis, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk/bitmap.go` line 36 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0eoJda3FIQ-dqnVcNY:-O0iiiaQ1GPdkmqDDZOD:b-u9938v) ([raw file](https://github.com/cockroachdb/pebble/blob/389d2f8d6d51684026441d8eaffa658121650394/sstable/colblk/bitmap.go#L36)):*

Yeah I meant we'd still take the `bitCount` but assert that `len(b)` is enough
</blockquote></details>

Ah,  gotcha! done

sumeerbhola (commented):
> Reviewed 1 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3711)* status: 1 of 6 files reviewed, 6 unresolved discussions

___
*[`sstable/colblk/base.go` line 11 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0u25JrE-DDWX3KhOn9:-O0u25JrE-DDWX3KhOnA:b-h4f8mx) ([raw file](https://github.com/cockroachdb/pebble/blob/f375da4856f19356cf919f43c168143f6962cc3f/sstable/colblk/base.go#L11)):*
> ```Go
> 
> // align returns the next value greater than or equal to offset that's divisible
> // by val.
> ```

This is only for val's that are powers of 2, yes?
___
*[`sstable/colblk/column.go` line 127 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0u2MdX1z8QAAcGwWSM:-O0u2MdY09JFtMbO7Woi:bqesc2n) ([raw file](https://github.com/cockroachdb/pebble/blob/f375da4856f19356cf919f43c168143f6962cc3f/sstable/colblk/column.go#L127)):*
> ```Go
> 	Encoder
> 	// NumColumns returns the number of columns the ColumnWriter will encode.
> 	NumColumns() int
> ```

if this is a writer for a single column, why would it ever return a value other than 1?

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3711)* status: 1 of 6 files reviewed, 11 unresolved discussions

___
*[`sstable/colblk/bitmap.go` line 26 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0u7m4u3YRp7RNg-HgY:-O0u7m4u3YRp7RNg-HgZ:b11noe0) ([raw file](https://github.com/cockroachdb/pebble/blob/f375da4856f19356cf919f43c168143f6962cc3f/sstable/colblk/bitmap.go#L26)):*
> ```Go
> // summary bitmap corresponds to 1 word in the primary bitmap. A bit is set in
> // the summary bitmap if the corresponding word in the primary bitmap is
> // non-zero. The summary bitmap accelerates predecessor and successor
> ```

So the layout is
[63 ... 0] word, [127 ... 64] word, ...? 
___
*[`sstable/colblk/bitmap.go` line 62 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0u56fW0uVmfYwRM78I:-O0u56fX2VYdfMoNFKG6:bbwoxvn) ([raw file](https://github.com/cockroachdb/pebble/blob/f375da4856f19356cf919f43c168143f6962cc3f/sstable/colblk/bitmap.go#L62)):*
> ```Go
> 	nextInWord := func(word uint64, bit uint) int {
> 		// We want to find the index of the next set bit. We can accomplish this
> 		// by clearing the trailing `bit` bits from the word and counting the
> ```

this is clearing the trailing `bit-1` bits if bit > 0, else clearing nothing, yes?
___
*[`sstable/colblk/bitmap.go` line 115 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0u6Qf8CpfCsDZDS0hf:-O0u6Qf99O8epgc8YuuQ:bvjls9t) ([raw file](https://github.com/cockroachdb/pebble/blob/f375da4856f19356cf919f43c168143f6962cc3f/sstable/colblk/bitmap.go#L115)):*
> ```Go
> 	// which bit is set.
> 	wordIdx = ((summaryWordIdx - summaryTableOffset) << 6) + summaryNext
> 	return (wordIdx << 6) + bits.TrailingZeros64(b.data.At(wordIdx))
> ```

under invariants enabled, it may be worth checking that this call to `bits.TrailingZeros64` is not returning 64 (to ensure there is no inconsistency with the summary).
___
*[`sstable/colblk/bitmap.go` line 124 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0u9LTfApEWi7hjle1Z:-O0u9LTfApEWi7hjle1_:b-gumk9f) ([raw file](https://github.com/cockroachdb/pebble/blob/f375da4856f19356cf919f43c168143f6962cc3f/sstable/colblk/bitmap.go#L124)):*
> ```Go
> 	// prevInWord returns the index of the largest set bit ≤ bit within the
> 	// provided word. The returned index is an index local to the word. Returns
> 	// -1 if no set bit is found.
> ```

bit must be [0,64)?
___
*[`sstable/colblk/bitmap.go` line 127 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3711#-O0u9UrE5yiVPDenjHTo:-O0u9UrE5yiVPDenjHTp:b-pnjkrw) ([raw file](https://github.com/cockroachdb/pebble/blob/f375da4856f19356cf919f43c168143f6962cc3f/sstable/colblk/bitmap.go#L127)):*
> ```Go
> 	prevInWord := func(word uint64, bit uint) int {
> 		// We want to find the index of the previous set bit. We can accomplish
> 		// this by clearing the leading `bit` bits from the word and counting
> ```

clearing everything except for the trailing `bit+1` bits? (so if bit=0, we are clearing everything except the trailing 1 bit)

PR #3715, opened 2024-06-30
keyspan: improve LevelIter return verification

RaduBerinde:
Instead of calling `l.verify` in each return path, we now defer a call at the beginning of each positioning method. This is less error-prone and keeps the code cleaner.

We also correct some paths that weren't setting `l.err` in error cases.


RaduBerinde:
Closing in favor of #3717.

PR #3716, opened 2024-06-30
sstable: simplify monotonic bounds optimization condition

jbowens:
When a sstable iterator's bounds move monotonically forward to an abutting keyspan, the iterator employs an optimization to use the existing iterator position. When checking whether this condition was possible, the single-level iterator previously checked !i.index.IsDataInvalidated().

This case will always be true within the context of the single-level iterator. No single-level iterator methods other than initialization routines ever load the index block.

### DIFF
--- a/sstable/reader_iter_single_lvl.go
+++ b/sstable/reader_iter_single_lvl.go
@@ -654,7 +654,7 @@ func (i *singleLevelIterator) seekGEHelper(
 	// by trySeekUsingNext, or by monotonically increasing bounds (i.boundsCmp).
 
 	var dontSeekWithinBlock bool
-	if !i.data.IsDataInvalidated() && !i.index.IsDataInvalidated() && i.data.Valid() && i.index.Valid() &&
+	if !i.data.IsDataInvalidated() && i.data.Valid() && i.index.Valid() &&
 		boundsCmp > 0 && i.cmp(key, i.index.Key().UserKey) <= 0 {
 		// Fast-path: The bounds have moved forward and this SeekGE is
 		// respecting the lower bound (guaranteed by Iterator). We know that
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3716#-:-O0etRuqEvEiF3ay9Puh:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`sstable/reader_iter_single_lvl.go` line 657 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3716#-O0etE2r76iXWy3h0Jhx:-O0etE2r76iXWy3h0Jhy:b-4l90zd) ([raw file](https://github.com/cockroachdb/pebble/blob/f58d4cf6b72372ba7d4011c0b7eafc6b99938ce3/sstable/reader_iter_single_lvl.go#L657)):*
> ```Go
> 
> 	var dontSeekWithinBlock bool
> 	if !i.data.IsDataInvalidated() && i.data.Valid() && i.index.Valid() &&
> ```

Do we want to add an `if invariants.Enabled && i.index.IsDataInvalidated { panic }` somewhere? (in case the second level iterator doesn't set things up properly)

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3716#-:-O1Hncf87f0BZJ5D72jM:b-814g8k)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 657 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3716#-O0etE2r76iXWy3h0Jhx:-O1HnW8P6CH637iv9q_c:bx9ppy9) ([raw file](https://github.com/cockroachdb/pebble/blob/f58d4cf6b72372ba7d4011c0b7eafc6b99938ce3/sstable/reader_iter_single_lvl.go#L657)):*

Do we want to add an `if invariants.Enabled && i.index.IsDataInvalidated { panic }` somewhere? (in case the second level iterator doesn't set things up properly)
</blockquote></details>

I'll follow up with something to the effect in my next, related change.

PR #3718, opened 2024-07-01
colblk: add RawBytes, RawBytesBuilder

jbowens:
Add a new columnar primitive for storing an array of variable-length, unordered byte slices. The new format stores an offset table to provide readers with constant-time lookup of the position of any item in the array.

### DIFF
--- a/internal/binfmt/binfmt.go
+++ b/internal/binfmt/binfmt.go
@@ -155,6 +155,12 @@ func (f *Formatter) Pointer(off int) unsafe.Pointer {
 	return unsafe.Pointer(&f.data[f.off+off])
 }
 
+// Data returns the original data slice. Offset may be used to retrieve the
+// current offset within the slice.
+func (f *Formatter) Data() []byte {
+	return f.data
+}
+
 func (f *Formatter) newline(binaryData, comment string) {
 	f.lines = append(f.lines, [2]string{binaryData, comment})
 	f.buf.Reset()
--- a/sstable/colblk/raw_bytes.go
+++ b/sstable/colblk/raw_bytes.go
@@ -1,0 +1,186 @@
+// Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package colblk
+
+import (
+	"fmt"
+	"io"
+	"unsafe"
+
+	"github.com/cockroachdb/pebble/internal/binfmt"
+)
+
+// RawBytes holds an array of byte slices, stored as a concatenated data section
+// and a series of offsets for each slice. Byte slices within RawBytes are
+// stored in their entirety without any compression, ensuring stability without
+// copying.
+//
+// # Representation
+//
+// An array of N byte slices encodes N+1 offsets. The beginning of the data
+// representation holds an offsets table, in the same encoding as a
+// DataTypeUint32 column. The Uint32 offsets may be delta-encoded to save space
+// if all offsets fit within an 8-bit or 16-bit uint. Each offset is relative to
+// the beginning of the string data section (after the offset table).
+//
+// The use of delta encoding conserves space in the common case. In the context
+// of CockroachDB, the vast majority of offsets will fit in 16-bits when using
+// 32 KiB blocks (the size in use by CockroachDB). However a single value larger
+// than 65535 bytes requires an offset too large to fit within 16 bits, in which
+// case offsets will be encoded as 32-bit integers.
+//
+//	+-------------------------------------------------------------------+
+//	|        a uint32 offsets table, possibly delta encoded,            |
+//	|                possibly padded for 32-bit alignment               |
+//	|                      (see DeltaEncoding)                          |
+//	+-------------------------------------------------------------------+
+//	|                           String Data                             |
+//	|  abcabcada....                                                    |
+//	+-------------------------------------------------------------------+
+//
+// The DeltaEncoding bits of the ColumnEncoding for a RawBytes column describes
+// the delta encoding of the offset table.
+type RawBytes struct {
+	slices  int
+	offsets UnsafeUint32s
+	start   unsafe.Pointer
+	data    unsafe.Pointer
+}
+
+// MakeRawBytes constructs an accessor for an array of byte slices constructed
+// by RawBytesBuilder. Count must be the number of byte slices within the array.
+func MakeRawBytes(count int, b []byte, offset uint32, enc ColumnEncoding) RawBytes {
+	if count == 0 {
+		return RawBytes{}
+	}
+	dataOff, offsets := readUnsafeIntegerSlice[uint32](count+1 /* +1 offset */, b, offset, enc.Delta())
+	return RawBytes{
+		slices:  count,
+		offsets: offsets,
+		start:   unsafe.Pointer(&b[offset]),
+		data:    unsafe.Pointer(&b[dataOff]),
+	}
+}
+
+func defaultSliceFormatter(x []byte) string {
+	return string(x)
+}
+
+func rawBytesToBinFormatter(
+	f *binfmt.Formatter, count int, enc ColumnEncoding, sliceFormatter func([]byte) string,
+) {
+	if sliceFormatter == nil {
+		sliceFormatter = defaultSliceFormatter
+	}
+
+	rb := MakeRawBytes(count, f.Data(), uint32(f.Offset()), enc)
+	dataOffset := uint64(f.Offset()) + uint64(uintptr(rb.data)-uintptr(rb.start))
+	f.CommentLine("RawBytes")
+	f.CommentLine("Offsets table")
+	uintsToBinFormatter(f, count+1, ColumnDesc{DataType: DataTypeUint32, Encoding: enc}, func(offset uint64) string {
+		return fmt.Sprintf("%d [%d overall]", offset, offset+dataOffset)
+	})
+	f.CommentLine("Data")
+	for i := 0; i < rb.slices; i++ {
+		s := rb.At(i)
+		f.HexBytesln(len(s), "data[%d]: %s", i, sliceFormatter(s))
+	}
+}
+
+func (b RawBytes) ptr(offset int) unsafe.Pointer {
+	return unsafe.Pointer(uintptr(b.data) + uintptr(offset))
+}
+
+func (b RawBytes) slice(start, end int) []byte {
+	return unsafe.Slice((*byte)(b.ptr(start)), end-start)
+}
+
+// At returns the []byte at index i. The returned slice should not be mutated.
+func (b RawBytes) At(i int) []byte {
+	return b.slice(int(b.offsets.At(i)), int(b.offsets.At(i+1)))
+}
+
+// Slices returns the number of []byte slices encoded within the RawBytes.
+func (b RawBytes) Slices() int {
+	return b.slices
+}
+
+// RawBytesBuilder encodes a column of byte slices.
+type RawBytesBuilder struct {
+	rows    int
+	data    []byte
+	offsets UintBuilder[uint32]
+}
+
+// Assert that *RawBytesBuilder implements ColumnWriter.
+var _ ColumnWriter = (*RawBytesBuilder)(nil)
+
+// Reset resets the builder to an empty state, preserving the existing bundle
+// size.
+func (b *RawBytesBuilder) Reset() {
+	b.rows = 0
+	b.data = b.data[:0]
+	b.offsets.Reset()
+	// Add an initial offset of zero to streamline the logic in RawBytes.At() to
+	// avoid needing a special case for row 0.
+	b.offsets.Set(0, 0)
+}
+
+// NumColumns implements ColumnWriter.
+func (b *RawBytesBuilder) NumColumns() int { return 1 }
+
+// Put appends the provided byte slice to the builder.
+func (b *RawBytesBuilder) Put(s []byte) {
+	b.data = append(b.data, s...)
+	b.rows++
+	b.offsets.Set(b.rows, uint32(len(b.data)))
+}
+
+// PutConcat appends a single byte slice formed by the concatenation of the two
+// byte slice arguments.
+func (b *RawBytesBuilder) PutConcat(s1, s2 []byte) {
+	b.data = append(append(b.data, s1...), s2...)
+	b.rows++
+	b.offsets.Set(b.rows, uint32(len(b.data)))
+}
+
+// Finish writes the serialized byte slices to buf starting at offset. The buf
+// slice must be sufficiently large to store the serialized output. The caller
+// should use [Size] to size buf appropriately before calling Finish.
+func (b *RawBytesBuilder) Finish(
+	col int, rows int, offset uint32, buf []byte,
+) (uint32, ColumnDesc) {
+	desc := ColumnDesc{DataType: DataTypeBytes, Encoding: EncodingDefault}
+	if rows == 0 {
+		return 0, desc
+	}
+	dataLen := b.offsets.Get(rows)
+	dataOffset, offsetColumnDesc := b.offsets.Finish(0, rows+1, offset, buf)
+	desc.Encoding = offsetColumnDesc.Encoding
+	// Copy the data section.
+	endOffset := dataOffset + uint32(copy(buf[dataOffset:], b.data[:dataLen]))
+	return endOffset, desc
+}
+
+// Size computes the size required to encode the byte slices beginning in a
+// buffer at the provided offset. The offset is required to ensure proper
+// alignment. The returned uint32 is the offset of the first byte after the end
+// of the encoded data. To compute the size in bytes, subtract the [offset]
+// passed into Size from the returned offset.
+func (b *RawBytesBuilder) Size(rows int, offset uint32) uint32 {
+	if rows == 0 {
+		return 0
+	}
+	// Get the size needed to encode the rows+1 offsets.
+	offset = b.offsets.Size(rows+1, offset)
+	// Add the value of offset[rows] since that is the accumulated size of the
+	// first [rows] slices.
+	return offset + b.offsets.Get(rows)
+}
+
+// WriteDebug implements Encoder.
+func (b *RawBytesBuilder) WriteDebug(w io.Writer, rows int) {
+	fmt.Fprintf(w, "bytes: %d rows set; %d bytes in data", b.rows, len(b.data))
+}
--- a/sstable/colblk/raw_bytes_test.go
+++ b/sstable/colblk/raw_bytes_test.go
@@ -1,0 +1,140 @@
+// Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package colblk
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"strings"
+	"testing"
+
+	"github.com/cockroachdb/datadriven"
+	"github.com/cockroachdb/pebble/internal/aligned"
+	"github.com/cockroachdb/pebble/internal/binfmt"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/exp/rand"
+)
+
+func TestRawBytes(t *testing.T) {
+	var out bytes.Buffer
+	var builder RawBytesBuilder
+	var rawBytes RawBytes
+	datadriven.RunTest(t, "testdata/raw_bytes", func(t *testing.T, td *datadriven.TestData) string {
+		out.Reset()
+		switch td.Cmd {
+		case "build":
+			builder.Reset()
+
+			var startOffset int
+			td.ScanArgs(t, "offset", &startOffset)
+
+			var count int
+			for _, k := range strings.Split(strings.TrimSpace(td.Input), "\n") {
+				builder.Put([]byte(k))
+				count++
+			}
+			td.MaybeScanArgs(t, "count", &count)
+
+			size := builder.Size(count, uint32(startOffset))
+			fmt.Fprintf(&out, "Size: %d\n", size-uint32(startOffset))
+
+			buf := aligned.ByteSlice(startOffset + int(size))
+			endOffset, desc := builder.Finish(0, count, uint32(startOffset), buf)
+
+			// Validate that builder.Size() was correct in its estimate.
+			require.Equal(t, size, endOffset)
+			f := binfmt.New(buf).LineWidth(20)
+			if startOffset > 0 {
+				f.HexBytesln(startOffset, "start offset")
+			}
+			rawBytesToBinFormatter(f, count, desc.Encoding, nil)
+			rawBytes = MakeRawBytes(count, buf[:], 0, desc.Encoding)
+			return f.String()
+		case "at":
+			var indices []int
+			td.ScanArgs(t, "indices", &indices)
+			for i, index := range indices {
+				if i > 0 {
+					fmt.Fprintln(&out)
+				}
+				fmt.Fprintf(&out, "%s", rawBytes.At(index))
+			}
+			return out.String()
+		default:
+			panic(fmt.Sprintf("unrecognized command %q", td.Cmd))
+		}
+	})
+}
+
+func BenchmarkRawBytes(b *testing.B) {
+	seed := uint64(205295296)
+	generateRandomSlices := func(sliceSize, totalByteSize int) [][]byte {
+		rng := rand.New(rand.NewSource(seed))
+		randInt := func(lo, hi int) int {
+			return lo + rng.Intn(hi-lo)
+		}
+		data := make([]byte, totalByteSize)
+		for i := range data {
+			data[i] = byte(randInt(int('a'), int('z')+1))
+		}
+		slices := make([][]byte, 0, totalByteSize/sliceSize)
+		for len(data) > 0 {
+			s := data[:min(sliceSize, len(data))]
+			slices = append(slices, s)
+			data = data[len(s):]
+		}
+		return slices
+	}
+
+	var builder RawBytesBuilder
+	var buf []byte
+	buildRawBytes := func(slices [][]byte) ([]byte, ColumnEncoding) {
+		builder.Reset()
+		for _, s := range slices {
+			builder.Put(s)
+		}
+		sz := builder.Size(len(slices), 0)
+		if cap(buf) >= int(sz) {
+			buf = buf[:sz]
+		} else {
+			buf = make([]byte, sz)
+		}
+		_, desc := builder.Finish(0, len(slices), 0, buf)
+		return buf, desc.Encoding
+	}
+
+	sizes := []int{8, 128, 1024}
+
+	b.Run("Builder", func(b *testing.B) {
+		for _, sz := range sizes {
+			b.Run(fmt.Sprintf("sliceLen=%d", sz), func(b *testing.B) {
+				slices := generateRandomSlices(sz, 32<<10 /* 32 KiB */)
+				b.ResetTimer()
+				b.SetBytes(32 << 10)
+				for i := 0; i < b.N; i++ {
+					data, _ := buildRawBytes(slices)
+					fmt.Fprint(io.Discard, data)
+				}
+			})
+		}
+	})
+	b.Run("At", func(b *testing.B) {
+		for _, sz := range sizes {
+			b.Run(fmt.Sprintf("sliceLen=%d", sz), func(b *testing.B) {
+				slices := generateRandomSlices(sz, 32<<10 /* 32 KiB */)
+				data, enc := buildRawBytes(slices)
+				rb := MakeRawBytes(len(slices), data, 0, enc)
+				b.ResetTimer()
+				b.SetBytes(32 << 10)
+				for i := 0; i < b.N; i++ {
+					for j := 0; j < len(slices); j++ {
+						fmt.Fprint(io.Discard, rb.At(j))
+					}
+				}
+			})
+		}
+	})
+}
--- a/sstable/colblk/testdata/raw_bytes
+++ b/sstable/colblk/testdata/raw_bytes
@@ -1,0 +1,216 @@
+# Test a simple RawBytes: a single byte slice with a single 'a' byte.
+
+build offset=0
+a
+----
+# RawBytes
+# Offsets table
+0-4: x 00000000 # 32-bit constant: 0
+4-5: x 00       # data[0] = 0 [6 overall]
+5-6: x 01       # data[1] = 1 [7 overall]
+# Data
+6-7: x 61       # data[0]: a
+
+# Try the same thing, but with offsets 1, 2, 3, 4.
+
+build offset=1
+a
+----
+00-01: x 00       # start offset
+# RawBytes
+# Offsets table
+# Padding
+01-04: x 000000   # aligning to 32-bit boundary
+04-08: x 00000000 # 32-bit constant: 0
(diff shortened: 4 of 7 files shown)


petermattis (commented):
Some drive-by comments on the comments. Best if @RaduBerinde or someone else from Storage gives this a thorough review.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3718#-:-O0jDCIB78xdK-z6Slue:baarp4j)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/colblk/raw_bytes.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0j9UMk9eZ5PCeq0WP7:-O0j9UMk9eZ5PCeq0WP8:b94abck) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L26)):*
> ```Go
> // An array of N byte slices encodes N+1 offsets. The beginning of the data
> // representation holds an offsets table, broken into two sections: 16-bit
> // offsets and 32-bit offsets. The first two bytes of the offset table hold the
> ```

Let's add some words about why the 32-bit offsets exist: in order to handle values that are larger than 65535 bytes. Such values are rare, but the rowblk format supports them and CRDB allows them as well. Conceptually, all of the offsets are 32-bits, but we're optimizing for the common case where they fit into 16-bits which materially reduces the space overhead bringing it more in line with the rowblk encoding overhead.

Does it simplify anything significantly to only support a single value larger than 65535 bytes? I don't think we'll ever see that happen outside of tests with the default block size, because as soon as you add the large value to the block we'll flush the block. At the time I prototyped this structure I felt the generality was worth it, but just want to raise the question again now that this is being productionized. 
___
*[`sstable/colblk/raw_bytes.go` line 35 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0jCIVeCENnVMNrc-5w:-O0jCIVeCENnVMNrc-5x:bx9lsl2) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L35)):*
> ```Go
> //	|                16-bit offsets (k) [16-bit aligned]                |
> //	|                                                                   |
> //	| off_0 | off_1 | off_2 | .................................| off_k  |
> ```

Nit: isn't the last offset here `off_{k-1}`? And the first offset of the 32-bit offsets should be `off_{k}`.
___
*[`sstable/colblk/raw_bytes.go` line 36 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0jCAKr1O2pSBJuvb41:-O0jCAKr1O2pSBJuvb42:b-ii9uoo) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L36)):*
> ```Go
> //	|                                                                   |
> //	| off_0 | off_1 | off_2 | .................................| off_k  |
> //	+-------------------------------------------------------------------+
> ```

I think there is aligning padding between the 16-bit and 32-bit offsets as well.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3718)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @petermattis and @RaduBerinde)

___
*[`sstable/colblk/raw_bytes.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0j9UMk9eZ5PCeq0WP7:-O0jYDa5D-0eE8daTxSY:bcvtgop) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L26)):*
> Does it simplify anything significantly to only support a single value larger than 65535 bytes? I don't think we'll ever see that happen outside of tests with the default block size, because as soon as you add the large value to the block we'll flush the block. At the time I prototyped this structure I felt the generality was worth it, but just want to raise the question again now that this is being productionized.

Interesting—I didn't really consider limiting the column size that aggressively. At read time we always know the end offset of the entire RawBytes (either the start offset of the next column, or the end of the block). We could avoid serializing the final offset at all and let the last offset always be implicitly the end of the RawBytes column.

I think that would be simpler for RawBytes, but PrefixBytes is trickier. A large key's bundle prefix and row suffix offsets could both be larger than 65535 (even if you limit the size of bundle prefixes, if it's not the first key in the block).
___
*[`sstable/colblk/raw_bytes.go` line 36 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0jCAKr1O2pSBJuvb41:-O0jXxd58Kp75HkZxTYC:b-yf5ll9) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L36)):*

I think there is aligning padding between the 16-bit and 32-bit offsets as well.
</blockquote></details>

The amount of padding is sufficient to align both 16-bit and 32-bit offsets, but it's all added in one place between the offset count and the beginning of the offsets table.

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3718)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/colblk/raw_bytes.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0j9UMk9eZ5PCeq0WP7:-O0jah-B7by5U1n27TTQ:b-7kunwn) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L26)):*
> Interesting—I didn't really consider limiting the column size that aggressively. At read time we always know the end offset of the entire RawBytes (either the start offset of the next column, or the end of the block). We could avoid serializing the final offset at all and let the last offset always be implicitly the end of the RawBytes column.

I was thinking of a single offset, but what you're describing could be even better. On the other hand, I think the generality may be warranted and isn't significantly more complicated. Treating the last offset differently isn't that much different than treating the first N offsets differently than the last M offsets. Anyways, this was just a question for you and Radu to consider that had occurred to me. I clearly went with the general approach in the prototype. 
___
*[`sstable/colblk/raw_bytes.go` line 36 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0jCAKr1O2pSBJuvb41:-O0jabWK9bG7GWLgSEKA:bk8lspq) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L36)):*

The amount of padding is sufficient to align both 16-bit and 32-bit offsets, but it's all added in one place between the offset count and the beginning of the offsets table.
</blockquote></details>

Oh, you're right. I had forgotten how that worked.

sumeerbhola (changes requested):
some basic high-level questions

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3718#-:-O0uGynQ-qCGlH9t3wE_:b-o9vpd)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/colblk/raw_bytes.go` line 24 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0uEewp7K6jqOWmSN3A:-O0uEewq4RmacFXY_vZx:bobx96j) ([raw file](https://github.com/cockroachdb/pebble/blob/c106b3bf807a8ac5bbb425e9e40ccf356dde9925/sstable/colblk/raw_bytes.go#L24)):*
> ```Go
> // # Representation
> //
> // An array of N byte slices encodes N+1 offsets. The beginning of the data
> ```

why N+1 offsets? Don't we need only either the start or end for each slice?

And is this the starting offset? If yes, is N+1 so that we know where the last one ends? Isn't the length of `RawBytes` representation known at a higher layer that is putting together the output of various column builders?

Would `RawBytesColumn` represent what this really is -- a struct that is an accessor for a column containing an array of raw-bytes?
___
*[`sstable/colblk/raw_bytes.go` line 36 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0uFQsBEcpGcVAfvPB8:-O0uFQsBEcpGcVAfvPB9:b-e0mzbz) ([raw file](https://github.com/cockroachdb/pebble/blob/c106b3bf807a8ac5bbb425e9e40ccf356dde9925/sstable/colblk/raw_bytes.go#L36)):*
> ```Go
> //
> //	+----------+--------------------------------------------------------+
> //	| uint16 k |                   (aligning padding)                   |
> ```

can you change this to say aligning padding to align to 32-bit, in light of the discussion below?
Does this mean k is always even?
And is the padding always 16 bit, and even if there are no 32-bit offsets?

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3718)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/colblk/raw_bytes.go` line 40 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0uMfr15aYuE1tbE1aX:-O0uMfr15aYuE1tbE1aY:b-bql72k) ([raw file](https://github.com/cockroachdb/pebble/blob/c106b3bf807a8ac5bbb425e9e40ccf356dde9925/sstable/colblk/raw_bytes.go#L40)):*
> ```Go
> //	|                16-bit offsets (k) [16-bit aligned]                |
> //	|                                                                   |
> //	| off_0 | off_1 | off_2 | ............................,.| off_{k-1} |
> ```

It seems from reading the code that these offsets are from the start of string data. Can you add a code comment clarifying that here.
___
*[`sstable/colblk/raw_bytes.go` line 164 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0uLR0U5GgwCEBJX1yu:-O0uLR0U5GgwCEBJX1yv:b-e2qeb9) ([raw file](https://github.com/cockroachdb/pebble/blob/c106b3bf807a8ac5bbb425e9e40ccf356dde9925/sstable/colblk/raw_bytes.go#L164)):*
> ```Go
> // At returns the []byte at index i. The returned slice should not be mutated.
> func (b RawBytes) At(i int) []byte {
> 	s := b.slice(b.offset(i), b.offset(i+1))
> ```

So if we have N elements, the N-th offset gives us the ending of that N-th element. I can see how this fits with the "avoid special-casing" comment earlier. But I don't understand the `i <= b.nOffsets16`. If we have 0..N offsets and the first 3 are 16-bit offsets then why isn't nOffsets16=3 and the comparison `i < b.nOffsets16`?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3718)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @jbowens, @petermattis, and @sumeerbhola)

___
*[`sstable/colblk/raw_bytes.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0j9UMk9eZ5PCeq0WP7:-O0ucXiB7GiNtKUY4v03:b-4fbvdz) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L26)):*

> Interesting—I didn't really consider limiting the column size that aggressively. At read time we always know the end offset of the entire RawBytes (either the start offset of the next column, or the end of the block). We could avoid serializing the final offset at all and let the last offset always be implicitly the end of the RawBytes column.

I was thinking of a single offset, but what you're describing could be even better. On the other hand, I think the generality may be warranted and isn't significantly more complicated. Treating the last offset differently isn't that much different than treating the first N offsets differently than the last M offsets. Anyways, this was just a question for you and Radu to consider that had occurred to me. I clearly went with the general approach in the prototype.
</blockquote></details>

Can we restrict things so that the total length minus the length of the last value is strictly < 64KB?  That way all *start* offsets would be 16bit (only the total length would not).

RaduBerinde:
*[`sstable/colblk/raw_bytes.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0j9UMk9eZ5PCeq0WP7:-O0udoeK7_YJgxhRnYQo:bqd16kw) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L26)):*

Can we restrict things so that the total length minus the length of the last value is strictly < 64KB?  That way all *start* offsets would be 16bit (only the total length would not).
</blockquote></details>

Oh, sorry, that's exactly what Jackson said. It's conceptually simple - we serialize all start offsets.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3718)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @jbowens, @petermattis, and @sumeerbhola)

___
*[`sstable/colblk/raw_bytes.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0j9UMk9eZ5PCeq0WP7:-O0ufQN41XkC4K33Sg-q:bc8okac) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L26)):*

Oh, sorry, that's exactly what Jackson said. It's conceptually simple - we serialize all start offsets.
</blockquote></details>

Oh, but we'd need to encode the total length separately, I see

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3718)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @jbowens, @petermattis, and @sumeerbhola)

___
*[`sstable/colblk/raw_bytes.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0j9UMk9eZ5PCeq0WP7:-O0ufbz63lLWy7qAPoSj:bbnzqne) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L26)):*

Oh, but we'd need to encode the total length separately, I see
</blockquote></details>

Hm.. instead of a specific encoding, could we reuse the uint encoding? That would automatically use 16-bits internally except in cases where we exceed the 64KB size. We could encode the lengths rather than the offsets. I think that would be a great place to start since it reuses other code.

RaduBerinde:
*[`sstable/colblk/raw_bytes.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0j9UMk9eZ5PCeq0WP7:-O0ugx-h2H7IgKENSAXT:bwu6pmv) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L26)):*

Hm.. instead of a specific encoding, could we reuse the uint encoding? That would automatically use 16-bits internally except in cases where we exceed the 64KB size. We could encode the lengths rather than the offsets. I think that would be a great place to start since it reuses other code.
</blockquote></details>

One nice thing about reusing the uint encoding with lengths is that it can go down to 8 bits in many cases.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3718)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @petermattis, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk/raw_bytes.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0j9UMk9eZ5PCeq0WP7:-O0xzHHM8xNKSXw9BogQ:bg8ju0l) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L26)):*

One nice thing about reusing the uint encoding with lengths is that it can go down to 8 bits in many cases.
</blockquote></details>

I was thinking about the same thing yesterday, and we could reuse it. There's a bit of a complication when we introduce the prefix-compression [PrefixBytes type](https://github.com/jbowens/pebble/blob/22e01ea6ebff6558d281b9076a6746a21a2d3c8f/sstable/colblk/column_bytes.go#L20) and its associated writer. In the wip branch, PrefixBytes have the same layout as RawBytes—it just encodes more slices than rows in order to encode prefixes for the entire column and each "bundle" of K keys.

The issue is that we don't know the actual offsets when we're building the column. Every offset can be changed by a reduction in the block-level prefix, and all the offsets that make up a bundle can be changed by a reduction in the bundle's prefix. We need the Size calls to be accurate as we build the column, so the prototype opts for committing to whether an offset gets encoded in 16 bits up front. This means we'll sometimes encode an offset that ultimately does fit in 16-bits in 32-bits. This scheme doesn't really jive with reusing the UintBuilder.

As I'm typing this though, I'm realizing the reduction in the block prefix when you add a key that has a shorter matching prefix could be modeled as reducing the 'base' constant of the uint column, or during building, the "minimum". The `PrefixBytesBuilder` already needs to compute an accurate maximal offset during `Size` operations, and this is the "maximum."

RaduBerinde:
*[`sstable/colblk/raw_bytes.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0j9UMk9eZ5PCeq0WP7:-O0yZY9F-qq6_gulIXzx:b9ibb83) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L26)):*

I was thinking about the same thing yesterday, and we could reuse it. There's a bit of a complication when we introduce the prefix-compression [PrefixBytes type](https://github.com/jbowens/pebble/blob/22e01ea6ebff6558d281b9076a6746a21a2d3c8f/sstable/colblk/column_bytes.go#L20) and its associated writer. In the wip branch, PrefixBytes have the same layout as RawBytes—it just encodes more slices than rows in order to encode prefixes for the entire column and each "bundle" of K keys.

The issue is that we don't know the actual offsets when we're building the column. Every offset can be changed by a reduction in the block-level prefix, and all the offsets that make up a bundle can be changed by a reduction in the bundle's prefix. We need the Size calls to be accurate as we build the column, so the prototype opts for committing to whether an offset gets encoded in 16 bits up front. This means we'll sometimes encode an offset that ultimately does fit in 16-bits in 32-bits. This scheme doesn't really jive with reusing the UintBuilder.

As I'm typing this though, I'm realizing the reduction in the block prefix when you add a key that has a shorter matching prefix could be modeled as reducing the 'base' constant of the uint column, or during building, the "minimum". The `PrefixBytesBuilder` already needs to compute an accurate maximal offset during `Size` operations, and this is the "maximum."
</blockquote></details>

Another thing I thought about is that we probably want to be able to look up offsets from the raw data directly? Storing lengths would require a full pass over all the lengths to calculate the offsets. I guess we could do that when decompressing the block.

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3718)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @jbowens, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk/raw_bytes.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0j9UMk9eZ5PCeq0WP7:-O0y_ttK2TSoOWUQ_dPR:b-b0k8r4) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L26)):*

Another thing I thought about is that we probably want to be able to look up offsets from the raw data directly? Storing lengths would require a full pass over all the lengths to calculate the offsets. I guess we could do that when decompressing the block.
</blockquote></details>

I was about to write something similar about the drawback of lengths. During iteration we need an array of offsets. I doubt the space savings after compression from storing lengths will be significant enough to outweigh the complexity of having to do some special transformation after decompression.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3718)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @jbowens, @petermattis, and @sumeerbhola)

___
*[`sstable/colblk/raw_bytes.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0j9UMk9eZ5PCeq0WP7:-O0ylmx82b1Hb1QYKBZy:buoxa8v) ([raw file](https://github.com/cockroachdb/pebble/blob/a3fb17605c8a734940abfda47d627e1deebd9c38/sstable/colblk/raw_bytes.go#L26)):*

I was about to write something similar about the drawback of lengths. During iteration we need an array of offsets. I doubt the space savings after compression from storing lengths will be significant enough to outweigh the complexity of having to do some special transformation after decompression.
</blockquote></details>

Agreed. Plus I would hope that block compression would be able to compress the 16-bit offsets fairly well given that large batches of them share the high byte.

sumeerbhola (approved):
:lgtm: 

> Reviewed 4 of 14 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3718#-:-O1X_UBI7Ikz2ybp_Gem:bg7igkv)* status: 4 of 14 files reviewed, 12 unresolved discussions (waiting on @jbowens, @petermattis, and @RaduBerinde)

___
*[`internal/binfmt/binfmt.go` line 99 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O1XX2lt1DFV5_OCBZsH:-O1XX2lt1DFV5_OCBZsI:byfwalf) ([raw file](https://github.com/cockroachdb/pebble/blob/1ccd777e15e3cfd4e6cafb3bf765999435e0c81e/internal/binfmt/binfmt.go#L99)):*
> ```Go
> 	commentLine := strings.TrimSpace(fmt.Sprintf(format, args...))
> 	printLine := func() {
> 		bytesInLine := min(f.lineWidth/2, n)
> ```

why `lineWidth/2` and not `lineWidth`?
Also why is there no looping in this method if n can be arbitrarily greater than `lineWidth`? There seems to be an assumption that calling `printLine` twice is sufficient.
___
*[`sstable/colblk/raw_bytes.go` line 34 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O1XUmOq7JCHkuDy__W9:-O1XUmOq7JCHkuDy__WA:b-ev6vho) ([raw file](https://github.com/cockroachdb/pebble/blob/1ccd777e15e3cfd4e6cafb3bf765999435e0c81e/sstable/colblk/raw_bytes.go#L34)):*
> ```Go
> //
> //	+-------------------------------------------------------------------+
> //	|                       32-bit aligning padding                     |
> ```

this padding is the responsibility of the offsets encoding so is internal to it, yes?
In that we don't do any alignment before calling `readUnsafeIntegerSlice`.
___
*[`sstable/colblk/raw_bytes.go` line 73 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O1XWXYH6fj0HRO6KA5-:-O1XWXYH6fj0HRO6KA50:bsow3d6) ([raw file](https://github.com/cockroachdb/pebble/blob/1ccd777e15e3cfd4e6cafb3bf765999435e0c81e/sstable/colblk/raw_bytes.go#L73)):*
> ```Go
> func rawBytesToBinFormatter(
> 	f *binfmt.Formatter, count int, enc ColumnEncoding, sliceFormatter func([]byte) string,
> ) int {
> ```

I should have asked this in an earlier PR, but I am a bit confused about these int return values, in terms of what they mean and why they are useful.
___
*[`sstable/colblk/raw_bytes.go` line 175 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O1XYdeh2Ip2631cIiNG:-O1XYdeh2Ip2631cIiNH:b2h5uhy) ([raw file](https://github.com/cockroachdb/pebble/blob/1ccd777e15e3cfd4e6cafb3bf765999435e0c81e/sstable/colblk/raw_bytes.go#L175)):*
> ```Go
> 		return 0
> 	}
> 	offset = b.offsets.Size(rows+1, offset)
> ```

(a comment would help)
// Get the size needed to encode the rows+1 offsets.
___
*[`sstable/colblk/raw_bytes.go` line 176 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O1XYhZv95cQ5TI6fcg4:-O1XYhZv95cQ5TI6fcg5:b-4p97z8) ([raw file](https://github.com/cockroachdb/pebble/blob/1ccd777e15e3cfd4e6cafb3bf765999435e0c81e/sstable/colblk/raw_bytes.go#L176)):*
> ```Go
> 	}
> 	offset = b.offsets.Size(rows+1, offset)
> 	return offset + b.offsets.Get(rows)
> ```

(a comment would help)
// Add the value of offset[rows] since that is the accumulated size of the slices.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3718)* status: 4 of 14 files reviewed, 11 unresolved discussions (waiting on @petermattis, @RaduBerinde, and @sumeerbhola)

___
*[`internal/binfmt/binfmt.go` line 99 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O1XX2lt1DFV5_OCBZsH:-O1XwKcw4uNFmg2hIege:bu1wfu1) ([raw file](https://github.com/cockroachdb/pebble/blob/1ccd777e15e3cfd4e6cafb3bf765999435e0c81e/internal/binfmt/binfmt.go#L99)):*

why `lineWidth/2` and not `lineWidth`?
Also why is there no looping in this method if n can be arbitrarily greater than `lineWidth`? There seems to be an assumption that calling `printLine` twice is sufficient.
</blockquote></details>

the /2 comes from 2-ascii characters per byte when representing it in hexadecimal

the `printLine` func doesn't loop, but right below we have a `for` loop calling `printLine`
___
*[`sstable/colblk/raw_bytes.go` line 36 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0uFQsBEcpGcVAfvPB8:-O1XwuLb2MILmzM-A0Rz:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c106b3bf807a8ac5bbb425e9e40ccf356dde9925/sstable/colblk/raw_bytes.go#L36)):*

can you change this to say aligning padding to align to 32-bit, in light of the discussion below?
Does this mean k is always even?
And is the padding always 16 bit, and even if there are no 32-bit offsets?
</blockquote></details>

Done.
___
*[`sstable/colblk/raw_bytes.go` line 40 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0uMfr15aYuE1tbE1aX:-O1XxBeV1EGmhui9vVJj:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c106b3bf807a8ac5bbb425e9e40ccf356dde9925/sstable/colblk/raw_bytes.go#L40)):*

It seems from reading the code that these offsets are from the start of string data. Can you add a code comment clarifying that here.
</blockquote></details>

Done.
___
*[`sstable/colblk/raw_bytes.go` line 164 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O0uLR0U5GgwCEBJX1yu:-O1XwwmJBgHawhMXqL1c:by0gkcx) ([raw file](https://github.com/cockroachdb/pebble/blob/c106b3bf807a8ac5bbb425e9e40ccf356dde9925/sstable/colblk/raw_bytes.go#L164)):*

So if we have N elements, the N-th offset gives us the ending of that N-th element. I can see how this fits with the "avoid special-casing" comment earlier. But I don't understand the `i <= b.nOffsets16`. If we have 0..N offsets and the first 3 are 16-bit offsets then why isn't nOffsets16=3 and the comparison `i < b.nOffsets16`?
</blockquote></details>

I think this got resolved through the refactor to use the uint-encoding
___
*[`sstable/colblk/raw_bytes.go` line 34 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O1XUmOq7JCHkuDy__W9:-O1XxS-i8eAWpPomJjhk:b-104q37) ([raw file](https://github.com/cockroachdb/pebble/blob/1ccd777e15e3cfd4e6cafb3bf765999435e0c81e/sstable/colblk/raw_bytes.go#L34)):*

this padding is the responsibility of the offsets encoding so is internal to it, yes?
In that we don't do any alignment before calling `readUnsafeIntegerSlice`.
</blockquote></details>

yeah, that's right. I adjusted this comment to make that it a little clearer
___
*[`sstable/colblk/raw_bytes.go` line 73 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O1XWXYH6fj0HRO6KA5-:-O1XxvyoDz3FJ4hOSIVl:btet7ua) ([raw file](https://github.com/cockroachdb/pebble/blob/1ccd777e15e3cfd4e6cafb3bf765999435e0c81e/sstable/colblk/raw_bytes.go#L73)):*

I should have asked this in an earlier PR, but I am a bit confused about these int return values, in terms of what they mean and why they are useful.
</blockquote></details>

They're intended to be the number of bytes consumed of the formatter, but they're not actually used and bugs cropped in as a result. I'll remove
___
*[`sstable/colblk/raw_bytes.go` line 175 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O1XYdeh2Ip2631cIiNG:-O1Xyi_89zohDt77baQt:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/1ccd777e15e3cfd4e6cafb3bf765999435e0c81e/sstable/colblk/raw_bytes.go#L175)):*

(a comment would help)
// Get the size needed to encode the rows+1 offsets.
</blockquote></details>

Done.
___
*[`sstable/colblk/raw_bytes.go` line 176 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O1XYhZv95cQ5TI6fcg4:-O1XyyCd7oGUxEYpmfUU:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/1ccd777e15e3cfd4e6cafb3bf765999435e0c81e/sstable/colblk/raw_bytes.go#L176)):*

(a comment would help)
// Add the value of offset[rows] since that is the accumulated size of the slices.
</blockquote></details>

Done.

sumeerbhola (approved):
:lgtm: 

> Reviewed 3 of 14 files at r6, 4 of 4 files at r7.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3718#-:-O1axxHc8oA5hRE1VrET:bg7igkv)* status: 9 of 14 files reviewed, 12 unresolved discussions (waiting on @jbowens, @petermattis, and @RaduBerinde)

___
*[`sstable/colblk/raw_bytes.go` line 178 at r7](https://reviewable.io/reviews/cockroachdb/pebble/3718#-O1avd4T0WF99p-yjvX_:-O1avd4T0WF99p-yjvXa:b3mpdk0) ([raw file](https://github.com/cockroachdb/pebble/blob/798f3e92cdf38cd9a0dc9c68846c699a9e7fa5e2/sstable/colblk/raw_bytes.go#L178)):*
> ```Go
> 	// Get the size needed to encode the rows+1 offsets.
> 	offset = b.offsets.Size(rows+1, offset)
> 	// Add the size of offset[rows] since that is the accumulate size of the
> ```

should this say "Add the value of ..."
nit: accumulated

jbowens:
TFTR!

Issue #918, opened 2020-09-17
perf: heuristics to reduce impact of point tombstones on read latency


jbowens:
High densities of point tombstones slow down iteration, wasting IO on loading keys that are discarded and cpu time skipping over them.

In RocksDB, sstables that contain majority deletion entries are prioritized for compaction through inflating these tables' compensated sizes. ~In Pebble, we never adjust compaction priorities to compact point tombstones.~ (_Update: In #1018 we adjusted compaction heuristics in way similar to RocksDB to increase the priority of compacting these point tombstones._)  We should investigate if any CockroachDB workloads accumulate high densities of point tombstones and those densities affect the workload in practice. If there is an impact, we might try triggering compactions of files based on the frequency with which an iterator skips over point tombstones in the file.

Related to #29, which also triggers compactions from reads but aimed at reducing read amplification.

Jira issue: PEBBLE-214

jbowens:
I updated the issue description to capture the fact that since #1018, Pebble does incorporate point tombstones into compaction heuristics by incorporating them into 'compensated size.'

It's still possible that read-triggered compactions of point tombstones would be beneficial, especially for point tombstones in a part of the keyspace where the average value size is small (resulting in less contribution to compensated size). The internal iterator stats added in #1538 can help us evaluate if the existing heuristics are enough.

sumeerbhola:
This could have helped in https://github.com/cockroachlabs/support/issues/2107#issuecomment-1452660323

jbowens:
This would've helped in cockroachlabs/support#1830.

jbowens:
I think this came up in another issue, but a more generalized approach could use CPU attribution to determine which sstables to compact. Seems tricky but possible?

sumeerbhola:
I think this is also coming up with rangedels that we are just skipping over in https://github.com/cockroachlabs/support/issues/2203#issuecomment-1498078068

andrewbaptist:
Can we add "IO attribution" to ranges? This would be similar to CPU attribution in the sense that it would use lower-level calls to determine how much each individual range is reading/writing per request and the KV layer could store and expose that information both at the range and node level. 

The general idea as we are attempting to keep systems balanced is that having more information can help us make better decisions. 

I think the way to do this best is to pass a "IOStats" object into all the mvcc operations that update it on the operation.

nicktrav:
> Can we add "IO attribution" to ranges? 

[...]

PR #3719, opened 2024-07-01
docs/rfc: add RFC for point tombstone density compaction heuristic

anish-shanbhag:
Adds an RFC for a new compaction heuristic that considers density of point tombstones.

Informs: #918

### DIFF
--- a/docs/RFCS/20240701_tombstone_density_heuristic.md
+++ b/docs/RFCS/20240701_tombstone_density_heuristic.md
@@ -1,0 +1,119 @@
+- Feature Name: Compaction heuristics for point tombstone density
+- Status: done as of 2024-08-15
+- Start Date: 2024-06-28
+- Authors: Anish Shanbhag
+- RFC PR: cockroachdb#3719
+- Pebble Issues: https://github.com/cockroachdb/pebble/issues/918
+
+# Summary
+
+This design document outlines a new compaction type in Pebble which specifically targets the reduction of point tombstone density to enhance read performance. The heuristic is designed to eliminate high densities of point tombstones which slow down iteration due to unnecessary I/O and CPU usage.
+
+The new heuristic was introduced in cockroachdb#3790.
+
+# Motivation
+
+Pebble currently contains a variety of compaction heuristics which are mostly based on SSTable file size. One weak point in the compaction heuristics involves the case where we have a large buildup of close-together point tombstones, which causes reads for keys after the tombstone cluster to become extremely slow.
+
+As identified [here](https://github.com/cockroachdb/pebble/issues/918#issuecomment-1564714073), uncompacted point tombstones reduce read performance in two ways: the extra CPU cycles needed to iterate over them during a seek, and the extra I/O needed to load more blocks which contain live keys.
+
+It's important that we reduce tombstone density before the tombstones are read because there are many cases where tombstones are written for very long periods before any reads are triggered. 
+
+Even though our current heuristics take the potential amount of reclaimed space from compacting point tombstones into account, there are specific situations where this is not sufficient to prevent point tombstone buildup. Specifically, these are some of the factors that lead to this issue:
+
+1. Tombstone buildup usually only happens in large stores (> 10 GB) once multiple levels of the LSM start to be populated.
+2. The LSM is "too well shaped" in that the size ratio between levels is at the proper value, so [compaction scores](https://github.com/cockroachdb/pebble/blob/3ef2e5b1f693dfbf78785e14f603a443af3c674b/compaction_picker.go#L919) for each level are all calculated to be <1 and thus no compactions are scheduled.
+	- Observed in [this escalation](https://github.com/cockroachlabs/support/issues/2628) (internal only)
+3. Read performance becomes especially bad when we have a high density of point tombstones in higher levels (L0-3) which span many SSTables in the bottommost levels (L4-6).
+4. The problem is especially apparent when there's one key range which we write to/delete from frequently and an adjacent key range which we read frequently.
+	- RocksDB describes how [queues commonly contain this type of workload](https://github.com/facebook/rocksdb/wiki/Implement-Queue-Service-Using-RocksDB#reclaiming-space-of-deleted-items-faster).
+	- We've seen this behavior multiple times with CockroachDB's implementation of the KV liveness range.
+		- Observed in multiple escalations including [here](https://github.com/cockroachlabs/support/issues/2107) and [here](https://github.com/cockroachlabs/support/issues/2640) (internal only)
+		- We had the raft log sandwiched between the frequently-read keys used for expiration-based leases. Nodes used to gossip their liveness every few seconds, which writes and deletes messages from the raft log over and over. This makes the raft log span numerous SSTables that fill up the cache with tombstones, removing the frequently-read lease expiration keys from the cache. Liveness has since been changed to use a different implementation but the root cause still exists here.
+	- Outbox/queue-based workloads also run into this issue because we continuously delete keys from the start of a range and add keys to the end of the range.
+5. Tombstones build up more often when KV pairs are small in size because more KVs fit into a single SSTable. In this case, heuristics that measure the possibility of disk space reclamation don't work because the tombstones take up little space despite filling up the key space.
+6. The problem is specific to operations including `SeekGE`/`SeekLT` and `Next`/`Prev` because we sequentially iterate over multiple keys during these. For example, if we `Next` into a swath of tombstones, we need to step through all of the tombstones before reaching a live key.
+
+# Design
+
+The new compaction heuristic introduces the concept of a "tombstone-dense" data block. A data block is considered tombstone-dense if it fulfills either of the following criteria:
+1. The block contains at least `N` point tombstones. The default value for `N` is 100, and is controlled through a new option `options.Experimental.NumDeletionsThreshold`.
+2. The ratio of the uncompressed size of point tombstones to the uncompressed size of the block is at least `Y`.  For example, with the default value of `0.5`, a data block of size 4KB would be considered tombstone-dense if it contains at least 2KB of point tombstones. The threshold is configured through a new option `options.Experimental.DeletionSizeRatioThreshold`.
+
+The two criteria above are meant to eliminate wasteful data blocks that would consume unnecessary resources during reads. The count-based threshold prevents CPU waste, and the size-based threshold prevents I/O waste.
+
+A table is considered eligible for the new tombstone compaction type if the percent of tombstone-dense data blocks (compared to the total number of data blocks in the table) is at least X%. The default is 5% and is configured via the new `options.Experimental.TombstoneDenseCompactionThreshold`. This option may also be set to zero or a negative value to disable tombstone density compactions.
+
+Tombstone-dense data blocks are identified during sstable write time, and their count is recorded in a new table property called `NumTombstoneDenseBlocks`. During table stats collection, this property is used in conjunction with the `NumDataBlocks` property to calculate the new `TombstoneDenseBlocksRatio` table stat. If this calculated ratio is above `TombstoneDenseCompactionThreshold`, the table becomes eligible for a tombstone density compaction.
+
+We use a `manifest.Annotator` in a similar way to elision-only compactions in order to prioritize compacting the table with the highest `TombstoneDenseBlocksRatio` if there are multiple eligible tables.
+
+Tombstone density compactions are executed the same as a default compaction once the candidate file has been chosen. They are prioritized strictly below default compactions and strictly above read-triggered compactions.
+
+## Design Exploration
+
+Below are some ideas which were previously explored when designing this heuristic. Although the heuristic introduced in cockroachdb#3790 differs from these, the ideas may be useful if we still see this problem in the future.
+
+<details>
+<summary>(Expand to view)</summary>
+
+### 1. Tombstone Ratio
+
+**Note: this heuristic was implemented in cockroachdb#3793 and was found to be somewhat ineffective based the benchmarks tested there**
+
+The simplest way to detect a buildup of point tombstones is to define some threshold percentage (`TOMBSTONE_THRESHOLD`) which indicates that any SSTable where `NumDeletions/NumEntries > TOMBSTONE_THRESHOLD` should be compacted. For example, if `TOMBSTONE_THRESHOLD = 0.6` an SSTable with 10,000 internal keys would be scheduled for compaction if it has at least 6000 tombstones.
+
+- Only considers tombstone density for one SSTable - overlaps with other tables aren't considered
+- Runs into issues if we have a swath of point tombstones in a very large SSTable - they still slow down reads but aren't big compared to the overall number of keys in the table
+- Probably insufficient to use on its own, but we could try combining this with other methods
+	- For example, both [RocksDB](https://github.com/facebook/rocksdb/blob/main/include/rocksdb/utilities/table_properties_collectors.h#L53) and [ScyllaDB](https://enterprise.docs.scylladb.com/stable/kb/garbage-collection-ics.html) use this in combination with more complex strategies
+
+### 2. More Granularity
+
+For more granularity on a per-SSTable basis, i.e. it's important to know where tombstones are clustered within an SSTable, there are two possible options:
+- Divide the key range of the SSTable into buckets of `X` keys and calculate how many have `>Y%` tombstones in them
+	- If at least `Z` buckets are tombstone dense, compact this table
+- Adapt the [sliding window approach from RocksDB](https://github.com/facebook/rocksdb/blob/22fe23edc89e9842ed72b613de172cd80d3b00da/utilities/table_properties_collectors/compact_on_deletion_collector.cc#L33)
+	- RocksDB uses an approach where they "slide a window" across the SSTable keys and schedule compaction if the window has a high enough ratio of tombstones. In other words, while writing if there are ever at least `X` tombstones in the last `Y` keys, compact this table
+	- We could adapt this 1:1 or try some modifications:
+		- Allow the window to expand while tombstones are still being written in order to check the exact key ranges which are tombstone dense. Prioritize compaction of SSTables based on the length of its tombstone swaths
+
+Like the `TOMBSTONE_THRESHOLD` strategy, this only considers single SSTables, so we can just calculate these metrics on the fly while writing the SSTable and immediately schedule them for compaction if they meet a density criteria.
+
+**Note: the heuristic we implemented in cockroachdb#3790 is an example of this category, as we measure tombstone density at block-level granularity with that approach.**
+
+### 3. Key Range Statistics
+
+Both methods above only consider tombstone density on a per SSTable basis. We could have a situation where a single continuous key range actually spans tombstone swaths across many levels of the LSM, in which case looking at a single SSTable may not indicate that the key range is tombstone-dense.
+
+In this case, we want the ability to query if a certain key range `a->b` is "tombstone dense", and if so then compact table(s) overlapping that range. Range annotations which were introduced in cockroachdb#3759 would be a good fit for the implementation.
+
+Given this method to query tombstone stats for arbitrary key ranges, here's a sketch of how the overall compaction process could look:
+- After writing an SSTable, add this SSTable and all SSTables which overlap with its key range (using `version.Overlaps`) to a global set `needsTombstoneCheck` which marks them as possibly eligible for a tombstone density compaction
+	- If the logic below ends up being fast enough, we could avoid having `needsTombstoneCheck` entirely and check whether compaction is needed during a write itself. But if not, we should defer the check in order to keep writes fast
+- Inside [`pickAuto`](https://github.com/cockroachdb/pebble/blob/4981bd0e5e9538a032a4caf3a12d4571abb8c206/compaction_picker.go#L1324), we'll check whether any SSTable in `needsTombstoneCheck` should be compacted
+	- For each SSTable `T` in `needsTombstoneCheck`, we can get the following info using the `Annotator` and the table statistics we already have (assuming this SSTable spans the key range `a->b`:
+		- This SSTable
+			- number of tombstones in `T`
+			- number of internal keys in `T`
+		- Whole LSM
+			- number of tombstones across the whole LSM between `a->b`
+			- total number of internal keys across the whole LSM between `a->b`
+			- Note: if we use the technique from \# 2 above to find the tombstone-dense range `m->n` within this SSTable, we could also get more granular stats:
+				- number of tombstones across the whole LSM between `m->n`
+				- number of internal keys across the whole LSM between `m->n`
+	- We now have the 6 stats above about the SSTable's key range, not including any other possible queries we could also make. I'm still unsure about which of these are actually important, and what's the best way to reconcile these into a single signal how how strongly we want to compact this SSTable. My current intuition is that we want to avoid situations where we have a large number of tombstones overlapping a large number of keys in lower levels, so maybe we output a signal proportional to `this sstable's tombstone count / total key count of overlapping sstables`, or `this sstable's tombstone count / total key count of sstables overlapping with this sstable's tombstone cluster(s)` in the more granular case? Open to suggestions here. I'm also wondering if we should explicitly prioritize compaction of lower levels vs. higher levels (e.g. L0 vs. L5)
+	- Similar to the check in read-based compaction, we'll want to make sure the SSTable still actually exists before compacting it, since it could have been compacted away between being added to `needsTombstoneCheck` and right now.
+
+### 4. Maximum Granularity
+
+If we find that the key range statistics method above works well but we want even more granularity for key ranges, i.e. because the overestimate of whole-LSM stats above becomes an issue, then we could include per-block tombstone/key counts in the index block of each SSTable, which would allow us to get a more precise count of tombstones for a given key range. This would look pretty similar to the logic separating partial vs. full overlaps in [`estimateReclaimedSizesBeneath`](https://github.com/cockroachdb/pebble/blob/master/table_stats.go#L606), except we'd be checking tombstone/key count instead of disk usage.
+- If we store a running total of the tombstones for each block in the index entry, making this query would be O(log n) or faster, not including the I/O overhead of reading the index block
+</details>
+
+# Testing
+
+Testing methodology and results can be found [here](https://github.com/cockroachdb/pebble/pull/3790#issuecomment-2251439492). Tombstone buildup has been frequently observed in queue-based workloads and the benchmark introduced in cockroachdb#3744 is meant to capture this case. Below are some further ideas for testing in case they're needed:
+- Idea from [here](https://github.com/cockroachdb/pebble/issues/918#issuecomment-1599478862) - seed a new node with a replication snapshot, which has the tendency for tomstone buildup
+- cockroachdb/cockroach#113069 measures performance of liveness range scans since those were often slowed down by tombstones. Even though it looks like liveness logic has changed since then to avoid this, we could adapt the test to induce slow reads.
+- Create reproduction roachtests for past slowdown scenarios we've observed


jbowens (commented):
still reading, but looking good so far

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3719#-:-O0oNC4nCIP9TEDw5X75:b-sl8wll)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @anish-shanbhag and @itsbilal)

___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 25 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O0oAQ7E26E66Fs6QFX4:-O0oAQ7E26E66Fs6QFX5:bpy00xs) ([raw file](https://github.com/cockroachdb/pebble/blob/75b849c48dd7571d606e2b1300312e41cc877df4/docs/RFCS/20240701_tombstone_density_heuristic.md#L25)):*
> ```Markdown
> 
> 1. Tombstone buildup usually only happens in large stores (> few hundred MB) once multiple levels of the LSM start to be populated.
> 2. The LSM is "too well shaped" in that the size ratio between levels is at the proper value, so [compaction scores](https://github.com/cockroachdb/pebble/blob/3ef2e5b1f693dfbf78785e14f603a443af3c674b/compaction_picker.go#L919) for each level are all calculated to be <0 and thus no compactions are scheduled.
> ```

nit : this should be "<1": compaction scores range from 0 to infinity and a score of ≥ 1.0 is sufficient to schedule a new compaction out of a level
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 27 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O0oKiUX1xhtyJRUy7lP:-O0oKiUX1xhtyJRUy7lQ:b-bwql6h) ([raw file](https://github.com/cockroachdb/pebble/blob/75b849c48dd7571d606e2b1300312e41cc877df4/docs/RFCS/20240701_tombstone_density_heuristic.md#L27)):*
> ```Markdown
> 2. The LSM is "too well shaped" in that the size ratio between levels is at the proper value, so [compaction scores](https://github.com/cockroachdb/pebble/blob/3ef2e5b1f693dfbf78785e14f603a443af3c674b/compaction_picker.go#L919) for each level are all calculated to be <0 and thus no compactions are scheduled.
> 	- Observed in [this escalation](https://github.com/cockroachlabs/support/issues/2628)
> 3. Read performance becomes especially bad when we have a high density of point tombstones in lower levels (L0-3) which span many SSTables in higher levels (L4-6).
> ```

nit: confusingly, we refer to L0-3 as "higher" levels, and refer to L6 as the bottommost level. 
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 34 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O0oLBAX6N3DG3G8uBJ9:-O0oLBAX6N3DG3G8uBJA:bbvxfa8) ([raw file](https://github.com/cockroachdb/pebble/blob/75b849c48dd7571d606e2b1300312e41cc877df4/docs/RFCS/20240701_tombstone_density_heuristic.md#L34)):*
> ```Markdown
> 		- We had the raft log sandwiched between the frequently-read keys used for expiration-based leases. Nodes used to gossip their liveness every few seconds, which writes and deletes messages from the raft log over and over. This makes the raft log span numerous SSTables that fill up the cache with tombstones, removing the frequently-read lease expiration keys from the cache. Liveness has since been changed to use a different implementation but the root cause still exists here.
> 5. Tombstones build up more often when KV pairs are small in size because more KVs fit into a single SSTable. In this case, heuristics that measure the possibility of disk space reclamation don't work because the tombstones take up little space despite filling up the key space.
> 6. The problem is specific to `SeekGE` and `SeekLT` because we have to iterate over all keys for these operations.
> ```

nit: this isn't quite true—it's also relevant for relative positioning operations Next and Prev; if you Next into a swath of tombstones, the iterator needs step through all the tombstones to arrive at the first live key.
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 56 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O0oM2g68m7kakLGpj5u:-O0oM2g68m7kakLGpj5v:b-qif2i6) ([raw file](https://github.com/cockroachdb/pebble/blob/75b849c48dd7571d606e2b1300312e41cc877df4/docs/RFCS/20240701_tombstone_density_heuristic.md#L56)):*
> ```Markdown
> 	- If at least `Z` buckets are tombstone dense, compact this table
> - Adapt the [sliding window approach from RocksDB](https://github.com/facebook/rocksdb/blob/22fe23edc89e9842ed72b613de172cd80d3b00da/utilities/table_properties_collectors/compact_on_deletion_collector.cc#L33)
> 	- RocksDB uses an approach where they "slide a window" across the SSTable keys and schedule compaction if the window has a high enough ratio of tombstones. In other words, while writing if there are ever at least `X` tombstones in the last `Y` keys, compact this table
> ```

TIL — nice RocksDB archeology

itsbilal (commented):
Broadly this is looking great! Some comments based on what's here so far. I think we're already in a stage where we can decide on the level of granularity we want, and start implementing soon. This RFC can still stay here as a design doc for a little bit longer, but it shouldn't be a blocker for implementation work for much longer I'd hope.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3719#-:-O1JEIGw7-rygs3_Wklu:b-35da4o)* status: 0 of 1 files reviewed, 11 unresolved discussions (waiting on @anish-shanbhag and @jbowens)

___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 24 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1IvX0W8RM_w4P648IK:-O1IvX0W8RM_w4P648IL:b-csq1z) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L24)):*
> ```Markdown
> Even though our current heuristics take the potential amount of reclaimed space from compacting point tombstones into account, there are specific situations where this is not sufficient to prevent point tombstone buildup. Specifically, these are some of the factors that lead to this issue:
> 
> 1. Tombstone buildup usually only happens in large stores (> few hundred MB) once multiple levels of the LSM start to be populated.
> ```

nit: we can say something bigger eg. > 10GB . 
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 32 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1IxQgv-l0GLZLf1_pD:-O1IxQgv-l0GLZLf1_pE:by1l4l2) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L32)):*
> ```Markdown
> 	- We've seen this behavior multiple times with CockroachDB's implementation of the KV liveness range.
> 		- Observed in multiple escalations including [here](https://github.com/cockroachlabs/support/issues/2107) and [here](https://github.com/cockroachlabs/support/issues/2640) 
> 		- We had the raft log sandwiched between the frequently-read keys used for expiration-based leases. Nodes used to gossip their liveness every few seconds, which writes and deletes messages from the raft log over and over. This makes the raft log span numerous SSTables that fill up the cache with tombstones, removing the frequently-read lease expiration keys from the cache. Liveness has since been changed to use a different implementation but the root cause still exists here.
> ```

we can also list outbox-style workloads here - workloads that add rows at the end of a table and delete keys off the start. Once GC'd, this leads to a dense region of tombstones near the start, that keeps sliding forward.
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 100 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1JBA9w4K-BTS5TlzZw:-O1JBA9w4K-BTS5TlzZx:bz8bdvo) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L100)):*
> ```Markdown
> Given this method to query tombstone stats for arbitrary key ranges, here's a sketch of how the overall compaction process could look:
> - After writing an SSTable, add this SSTable and all SSTables which overlap with its key range (using `version.Overlaps`) to a global set `needsTombstoneCheck` which marks them as possibly eligible for a tombstone density compaction
> 	- If the logic below ends up being fast enough, we could avoid having `needsTombstoneCheck` entirely and check whether compaction is needed during a write itself. But if not, we should defer the check in order to keep writes fast
> ```

The annotator framework should already take advantage of unchanged subtrees before/after the addition of an sstable, and retain the cached annotations. So I don't see the need to keep a separate bool. We should be able to capture any sstable-level metrics during sstable write time, and the compaction picker can just do the tombstone density query when it's picking a new compaction - lazily doing the calculation down the tree for any non-cached values.
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 102 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1J68Oj9--t6PZxPrzC:-O1J68Oj9--t6PZxPrzD:b-4s90kj) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L102)):*
> ```Markdown
> 	- If the logic below ends up being fast enough, we could avoid having `needsTombstoneCheck` entirely and check whether compaction is needed during a write itself. But if not, we should defer the check in order to keep writes fast
> - Inside [`pickAuto`](https://github.com/cockroachdb/pebble/blob/4981bd0e5e9538a032a4caf3a12d4571abb8c206/compaction_picker.go#L1324), we'll check whether any SSTable in `needsTombstoneCheck` should be compacted
> 	- Like elision-only and read-based compaction, we likely only want to do this check if there are no regular, size-based compactions that should be scheduled. Open question: how should these compactions be prioritized against elision-only and read-based compaction?
> ```

I can see a strong case for slotting this above read-triggered compactions in priority, and maybe even slightly ahead of elision-only compactions. Read-triggered compactions are truly the compactions we schedule when we have no other intuition on compactions to schedule, and make no assumption on keys being deleted or space being freed up. In this case there's a stronger case to be made about dropping the tombstones, so it should be higher priority than that.
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 112 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1J82aoAJLSbPsLsANu:-O1J82aoAJLSbPsLsANv:bend0uu) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L112)):*
> ```Markdown
> 			- Note: if we use the technique from \# 2 above to find the tombstone-dense range `m->n` within this SSTable, we could also get more granular stats:
> 				- number of tombstones across the whole LSM between `m->n`
> 				- number of internal keys across the whole LSM between `m->n`
> ```

This can make a lot of sense if we happen to have tombstone dense ranges of keys that are right at sstable bounds, but the rest of the sstable(s) are still full of sets. That way we can surface a tombstone-dense keyrange even if no individual sstable has too many tombstones.

The one thing we'll need to keep in mind is that pebble doesn't have a good sense of how close together two keys are in the keyspace, so inevitably we'll also have to track keys in the not-tombstone-dense keyrange abutting a tombstone-dense keyrange. Once we know that, we can coalesce two tombstone-dense keyranges in the annotator if there's a small number of non-tombstone-dense keys in between. Does this make sense? Wonder what your thoughts are on addressing this issue.

My own preference would be to not worry about identifying tombstone-dense keyranges _within_ sstables (because if we're heading in that direction we have to do more work of this sort to make it worthwhile), and to instead focus the range annotator work on better adding up tombstone ranges across sstables (treating each sstable as an indivisible unit) and across levels. This might be too coarse to perfectly solve all corner cases, but it'd solve the bulk of these issues and would be easier to envision. Thoughts?
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 113 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1JBZXC5flYQaP52HJ9:-O1JBZXC5flYQaP52HJA:bnmlpj1) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L113)):*
> ```Markdown
> 				- number of tombstones across the whole LSM between `m->n`
> 				- number of internal keys across the whole LSM between `m->n`
> 	- Note that stats about the whole LSM would be overestimates since we'd be including tombstones/keys from SSTables which only partially overlap the range. I think this should be fine?
> ```

This is fine because if we were to do a compaction for that range, we'd need to compact the entirety of that sstable anyway (and anything that overlaps), so some overestimation is okay.
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 131 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1J9L5q7_x9Tk4uu2Aq:-O1J9L5q7_x9Tk4uu2Ar:bszspdl) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L131)):*
> ```Markdown
> - cockroachdb/cockroach#113069 measures performance of liveness range scans since those were often slowed down by tombstones. Even though it looks like liveness logic has changed since then to avoid this, we could adapt the test to induce slow reads.
> - Create reproduction roachtests for past slowdown scenarios we've observed
> - RocksDB highlights that a [queue service built on top of a KV store](https://github.com/facebook/rocksdb/wiki/Implement-Queue-Service-Using-RocksDB) is a common situation where tombstones build up. We should write a test to capture performance for a queue
> ```

Yep, queue/outbox-style workloads definitely exhibit something like this. If you were to run a workload exhibiting this behaviour in cockroach, you could try a `kv50` workload with a high delete percentage too, and lower the mvcc gcttl using `ALTER RANGE default CONFIGURE ZONE USING gc.ttlseconds=3600` or so, so we start to lay down pebble tombstones in an hour of a key being deleted.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3719)* status: 0 of 1 files reviewed, 15 unresolved discussions (waiting on @anish-shanbhag and @jbowens)

___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 14 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1KOHVCB1D9p1dShUmC:-O1KOHVCB1D9p1dShUmD:b-wd3wvi) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L14)):*
> ```Markdown
> This design document outlines a proposal for improving compaction heuristics in Pebble, specifically targeting the reduction of point tombstone density to enhance read performance. It identifies the problem of high densities of point tombstones slowing down iteration due to unnecessary I/O and CPU usage. We propose various methods to identify regions of the LSM with high tombstone density and schedule the corresponding SSTables for compaction.
> 
> # Motivation
> ```

This section is great!
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 42 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1KP4Tv3XDE1yl9WcZR:-O1KP4Tv3XDE1yl9WcZS:booz4ot) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L42)):*
> ```Markdown
> ### 1. Tombstone Ratio
> 
> The simplest way to detect a buildup of point tombstones is to define some threshold percentage (`TOMBSTONE_THRESHOLD`) which indicates that any SSTable where `NumDeletions/NumEntries > TOMBSTONE_THRESHOLD` should be compacted. For example, if `TOMBSTONE_THRESHOLD = 0.6` an SSTable with 10,000 internal keys would be scheduled for compaction if it has at least 6000 tombstones.
> ```

Probably we would want different thresholds on different levels.
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 54 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1KPXqPDWtpGhK2Zy9m:-O1KPXqQ5Qz9FBzv7hx6:b-1jbqt) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L54)):*
> ```Markdown
> If we find that more granularity is needed on a per-SSTable basis, i.e. it's important to know where tombstones are clustered within an SSTable, there are two possible options:
> - Divide the key range of the SSTable into buckets of `X` keys and calculate how many have `>Y%` tombstones in them
> 	- If at least `Z` buckets are tombstone dense, compact this table
> ```

Here the "buckets" could just be data blocks. We can give each data block a tombstone density score and then aggregate it into some final metric (e.g. a very rudimentary histogram). We can use the existing property collectors for this.

We could even consult (for each data block) how many files in the levels below overlap that data block (at the time of writing that table) and work that into the heuristic. For each block the number of tombstones multiplied by the number of overlapping files below would probably be the "score" we care about.
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 62 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1KQrhw8SDL2RugpEyg:-O1KQrhx8bUb5z4GSbxj:betv0bx) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L62)):*
> ```Markdown
> Like the `TOMBSTONE_THRESHOLD` strategy, this only considers single SSTables, so we can just calculate these metrics on the fly while writing the SSTable and immediately schedule them for compaction if they meet a density criteria.
> 
> ### 3. Key Range Statistics
> ```

I would recommend at least experimenting with a prototype with one of the simpler approaches before going down this route.

anish-shanbhag (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3719)* status: 0 of 1 files reviewed, 15 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 25 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O0oAQ7E26E66Fs6QFX4:-O1OBG1n5x9dw9tLdc-O:b-7ql1td) ([raw file](https://github.com/cockroachdb/pebble/blob/75b849c48dd7571d606e2b1300312e41cc877df4/docs/RFCS/20240701_tombstone_density_heuristic.md#L25)):*

nit : this should be "<1": compaction scores range from 0 to infinity and a score of ≥ 1.0 is sufficient to schedule a new compaction out of a level
</blockquote></details>

fixed
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 27 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O0oKiUX1xhtyJRUy7lP:-O1OBJJm2aa7_fbhLBrl:b-ncys50) ([raw file](https://github.com/cockroachdb/pebble/blob/75b849c48dd7571d606e2b1300312e41cc877df4/docs/RFCS/20240701_tombstone_density_heuristic.md#L27)):*

nit: confusingly, we refer to L0-3 as "higher" levels, and refer to L6 as the bottommost level.
</blockquote></details>

Makes sense, fixed this
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 34 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O0oLBAX6N3DG3G8uBJ9:-O1OBxGw5VB92p9_Up1e:b-wwpqgc) ([raw file](https://github.com/cockroachdb/pebble/blob/75b849c48dd7571d606e2b1300312e41cc877df4/docs/RFCS/20240701_tombstone_density_heuristic.md#L34)):*

nit: this isn't quite true—it's also relevant for relative positioning operations Next and Prev; if you Next into a swath of tombstones, the iterator needs step through all the tombstones to arrive at the first live key.
</blockquote></details>

Good point, I'll look into testing these as well
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 24 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1IvX0W8RM_w4P648IK:-O1OAlM9AhG9Cc15q97q:b-7ql1td) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L24)):*

nit: we can say something bigger eg. > 10GB .
</blockquote></details>

fixed
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 32 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1IxQgv-l0GLZLf1_pD:-O1OBrqN0I_LiRT7ZBYe:b-q67rnn) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L32)):*

we can also list outbox-style workloads here - workloads that add rows at the end of a table and delete keys off the start. Once GC'd, this leads to a dense region of tombstones near the start, that keeps sliding forward.
</blockquote></details>

Yep, definitely planning to test this - also added a note to the doc here
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 42 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1KP4Tv3XDE1yl9WcZR:-O1OC2B92NVu6kHuLInq:b-3f6iv9) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L42)):*

Probably we would want different thresholds on different levels.
</blockquote></details>

Could you clarify why we might want different thresholds? I'm guessing it's because the bottommost tables could still have large swaths of tombstones despite simultaneously containing mostly live keys, since the tables are much bigger. So would we want a lower threshold for e.g. L5 vs L2?
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 54 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1KPXqPDWtpGhK2Zy9m:-O1OAQ6kAtQoBFRHzLeu:b63xvn4) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L54)):*

Here the "buckets" could just be data blocks. We can give each data block a tombstone density score and then aggregate it into some final metric (e.g. a very rudimentary histogram). We can use the existing property collectors for this.

We could even consult (for each data block) how many files in the levels below overlap that data block (at the time of writing that table) and work that into the heuristic. For each block the number of tombstones multiplied by the number of overlapping files below would probably be the "score" we care about.
</blockquote></details>

Great idea, I was looking into the property collectors and thought this might be a good approach. 

One potential drawback might be the added size of each index block entry. Looks like we also considered keeping track of key/value disk size for `Reader.EstimateDiskUsage` but decided against it for this reason:

https://github.com/cockroachdb/pebble/blob/26fe5a09bc145e7ad5e9b22b7167fe4717823805/sstable/reader.go#L915

But if we store the density stats in a small value like a `uint8`/`uint16`, the 1-2 added bytes hopefully shouldn't have a meaningful impact on DB size.
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 100 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1JBA9w4K-BTS5TlzZw:-O1OJIVw2Z3HJxG6XGBr:bwem2wy) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L100)):*

The annotator framework should already take advantage of unchanged subtrees before/after the addition of an sstable, and retain the cached annotations. So I don't see the need to keep a separate bool. We should be able to capture any sstable-level metrics during sstable write time, and the compaction picker can just do the tombstone density query when it's picking a new compaction - lazily doing the calculation down the tree for any non-cached values.
</blockquote></details>

Makes sense, I think we can go without this. I was mainly concerned about the possible case where e.g. we compact from L4->L5 and this changes the amount of keys that are beneath a tombstone-dense region in an L0 SSTable. But the number of keys should only decrease or stay the same after the compaction, so an overestimate here should be fine.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3719)* status: 0 of 1 files reviewed, 15 unresolved discussions (waiting on @anish-shanbhag, @itsbilal, and @jbowens)

___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 54 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O1KPXqPDWtpGhK2Zy9m:-O1Pr9ly9pJI2NOPeyyE:b-dqsj9t) ([raw file](https://github.com/cockroachdb/pebble/blob/7c592ccac1d6abf7c50466830232a9ed6295a363/docs/RFCS/20240701_tombstone_density_heuristic.md#L54)):*

Great idea, I was looking into the property collectors and thought this might be a good approach. 

One potential drawback might be the added size of each index block entry. Looks like we also considered keeping track of key/value disk size for `Reader.EstimateDiskUsage` but decided against it for this reason:

https://github.com/cockroachdb/pebble/blob/26fe5a09bc145e7ad5e9b22b7167fe4717823805/sstable/reader.go#L915

But if we store the density stats in a small value like a `uint8`/`uint16`, the 1-2 added bytes hopefully shouldn't have a meaningful impact on DB size.
</blockquote></details>

I don't think increasing the handle is a huge deal. I interpret that TODO as saying that we don't want to increase it without good reason (I have a feeling it would have stayed a TODO regardless).

Each data block is ~32KB (uncompressed). If we add say 8 bytes to each block handle, that's only 0.02% of the total uncompressed size.

anish-shanbhag (commented):
I've updated to reflect the final design of the heuristic at #3790 - hopefully we can merge this to serve as a future reference.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3719#-:-O4CfR9d-Hjf832HtRpG:b-m3mkax)* status: 0 of 1 files reviewed, 15 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

itsbilal (approved):
Great work! :lgtm: save for some minor comments. This should be good to merge after those are addressed.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3719#-:-O4Ml74A8cslR9BOX35n:b5h3nmz)* status: 0 of 1 files reviewed, 18 unresolved discussions (waiting on @anish-shanbhag, @jbowens, and @RaduBerinde)

___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 2 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O4MlByG4cgR9cxvftlu:-O4MlByG4cgR9cxvftlv:b-eggmeq) ([raw file](https://github.com/cockroachdb/pebble/blob/6094fa647055dce766ae00f5961830a9f02523c3/docs/RFCS/20240701_tombstone_density_heuristic.md#L2)):*
> ```Markdown
> - Feature Name: Compaction heuristics for point tombstone density
> - Status: in-progress
> ```

Status: done, and possibly an end date or so below?
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 26 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O4MjDiU60poBjr9fgDk:-O4MjDiU60poBjr9fgDl:bz06ga4) ([raw file](https://github.com/cockroachdb/pebble/blob/6094fa647055dce766ae00f5961830a9f02523c3/docs/RFCS/20240701_tombstone_density_heuristic.md#L26)):*
> ```Markdown
> 1. Tombstone buildup usually only happens in large stores (> 10 GB) once multiple levels of the LSM start to be populated.
> 2. The LSM is "too well shaped" in that the size ratio between levels is at the proper value, so [compaction scores](https://github.com/cockroachdb/pebble/blob/3ef2e5b1f693dfbf78785e14f603a443af3c674b/compaction_picker.go#L919) for each level are all calculated to be <1 and thus no compactions are scheduled.
> 	- Observed in [this escalation](https://github.com/cockroachlabs/support/issues/2628)
> ```

RFCs are public (cockroachdb org) while support issues are not, so while these links are totally cool and I don't think we've leaked any sensitive info in this RFC, we should still be suffixing all of these support repo links with "(internal only)".
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 45 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O4Mjsva49DFLC8B5YUX:-O4Mjsva49DFLC8B5YUY:b-xxgosz) ([raw file](https://github.com/cockroachdb/pebble/blob/6094fa647055dce766ae00f5961830a9f02523c3/docs/RFCS/20240701_tombstone_density_heuristic.md#L45)):*
> ```Markdown
> The two criteria above are meant to eliminate wasteful data blocks that would consume unnecessary resources during reads. The count-based threshold prevents CPU waste, and the size-based threshold prevents I/O waste.
> 
> A table is considered eligible for the new tombstone compaction type if the percent of tombstone-dense data blocks (compared to the total number of data blocks in the table) is at least X%. The default is 5% and is configured via the new `options.Experimental.TombstoneDenseCompactionThreshold`. This option may also be set to `-1` to disable tombstone density compactions.
> ```

"may also be set to zero or a negative value"
___
*[`docs/RFCS/20240701_tombstone_density_heuristic.md` line 73 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3719#-O4Mks5z5FQd-6hFxbWt:-O4Mks6-BrHF2kmUDeSa:b-1bfce9) ([raw file](https://github.com/cockroachdb/pebble/blob/6094fa647055dce766ae00f5961830a9f02523c3/docs/RFCS/20240701_tombstone_density_heuristic.md#L73)):*
> ```Markdown
> ### 2. More Granularity
> 
> If we find that more granularity is needed on a per-SSTable basis, i.e. it's important to know where tombstones are clustered within an SSTable, there are two possible options:
> ```

Worth mentioning that the preferred approach (block-level tombstone density) is an example of an approach with greater granularity than the above, sstable-level approach.

anish-shanbhag (commented):
I've addressed these comments - TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3719#-:-O4fImBC-D9cB4yJRYzk:bp90ckq)* status: 0 of 1 files reviewed, 18 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

PR #3707, opened 2024-06-26
metrics: fix Compact.Duration metric bug

RaduBerinde:
Don't include in-progress ingested flushables for Compact.Duration.
They don't contribute to the total when they complete, so we can see
non-monotonic blips in the value.

I reproduced this by keeping track of the last reported duration and
setting up a goroutine to poll the metrics frequently.

Fixes #3706

PR #3729, opened 2024-07-02
db: skip counting flushable ingest in compaction duration metric

aadityasondhi:
We do not include flushes in our compaction duration metrics, which should also exclude flushableIngests. When they are complete, we do not update the aggregate duration metric, but we were including them in our in-progress compaction adjustments.

This patch ignores flushableIngests in `db.Metrics()` to preserve monotonicity of the duration metric.


RaduBerinde:
This was already fixed in https://github.com/cockroachdb/pebble/pull/3707

PR #3731, opened 2024-07-03
colblk: add UintBuilder, DeltaEncoding

jbowens:
Add a UintBuilder column writer for building uint{8,16,32,64} columns. The UintBuilder will use a delta encoding when possible to represent an array of values as a constant C and an array of deltas using a smaller-width uint. This is expected to reduce the size of blocks in many instances. Internal key trailers have zero sequence numbers in ingested sstables, and almost always have zero sequence numbers in L6 sstables. This will allow most data blocks in these sstables to encode either a single 8-byte constant, or an 8-byte constant and an array of 1-byte deltas if the key kinds vary.

### DIFF
--- a/batchrepr/writer_test.go
+++ b/batchrepr/writer_test.go
@@ -54,8 +54,8 @@ func prettyBinaryRepr(repr []byte) string {
 	}
 
 	f := binfmt.New(repr).LineWidth(40)
-	f.HexBytesln(8, " seqnum=%d", f.PeekInt(8))
-	f.HexBytesln(4, " count=%d", f.PeekInt(4))
+	f.HexBytesln(8, " seqnum=%d", f.PeekUint(8))
+	f.HexBytesln(4, " count=%d", f.PeekUint(4))
 	for r := Read(repr); len(r) > 0; {
 		prevLen := len(r)
 		kind, ukey, _, ok, err := r.Next()
--- a/internal/aligned/aligned.go
+++ b/internal/aligned/aligned.go
@@ -1,0 +1,26 @@
+// Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package aligned
+
+import (
+	"fmt"
+	"unsafe"
+)
+
+// ByteSlice allocates a new byte slice of length n, ensuring the address of the
+// beginning of the slice is word aligned. Go does not guarantee that a simple
+// make([]byte, n) is aligned. In practice it often is, especially for larger n,
+// but small n can often be misaligned.
+func ByteSlice(n int) []byte {
+	a := make([]uint64, (n+7)/8)
+	b := unsafe.Slice((*byte)(unsafe.Pointer(&a[0])), n)
+
+	// Verify alignment.
+	ptr := uintptr(unsafe.Pointer(&b[0]))
+	if ptr%unsafe.Sizeof(int(0)) != 0 {
+		panic(fmt.Sprintf("allocated []uint64 slice not %d-aligned: pointer %p", unsafe.Sizeof(int(0)), &b[0]))
+	}
+	return b
+}
--- a/internal/binfmt/binfmt.go
+++ b/internal/binfmt/binfmt.go
@@ -59,18 +59,18 @@ func (f *Formatter) Offset() int {
 	return f.off
 }
 
-// PeekInt reads a little-endian integer of the specified width at the current
-// offset.
-func (f *Formatter) PeekInt(w int) int {
+// PeekUint reads a little-endian unsigned integer of the specified width at the
+// current offset.
+func (f *Formatter) PeekUint(w int) uint64 {
 	switch w {
 	case 1:
-		return int(f.data[f.off])
+		return uint64(f.data[f.off])
 	case 2:
-		return int(binary.LittleEndian.Uint16(f.data[f.off:]))
+		return uint64(binary.LittleEndian.Uint16(f.data[f.off:]))
 	case 4:
-		return int(binary.LittleEndian.Uint32(f.data[f.off:]))
+		return uint64(binary.LittleEndian.Uint32(f.data[f.off:]))
 	case 8:
-		return int(binary.LittleEndian.Uint64(f.data[f.off:]))
+		return binary.LittleEndian.Uint64(f.data[f.off:])
 	default:
 		panic("unsupported width")
 	}
--- a/sstable/colblk/base.go
+++ b/sstable/colblk/base.go
@@ -2,7 +2,6 @@
 // of this source code is governed by a BSD-style license that can be found in
 // the LICENSE file.
 
-// Package colblk implements a columnar block format.
 package colblk
 
 import "golang.org/x/exp/constraints"
--- a/sstable/colblk/block.go
+++ b/sstable/colblk/block.go
@@ -1,0 +1,16 @@
+// Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+// Package colblk implements a columnar block format.
+//
+// TODO(jackson): Describe the format once we start merging the higher-level
+// block format.
+//
+// # Alignment
+//
+// Block buffers are required to be word-aligned, during encoding and decoding.
+// This ensures that if any individual column or piece of data requires
+// word-alignment, the writer can align the offset into the block buffer
+// appropriately to ensure that the data is word-aligned.
+package colblk
--- a/sstable/colblk/column.go
+++ b/sstable/colblk/column.go
@@ -56,6 +56,17 @@ func (t DataType) String() string {
 	return dataTypeName[t]
 }
 
+func (t DataType) uintWidth() uint32 {
+	if t >= DataTypeUint8 && t <= DataTypeUint64 {
+		rv := 1 << (t - DataTypeUint8)
+		if rv > 8 {
+			panic("width greater than 8 bytes")
+		}
+		return uint32(rv)
+	}
+	panic("not a unit")
+}
+
 // ColumnDesc describes the column's data type and its encoding.
 type ColumnDesc struct {
 	DataType DataType
@@ -93,30 +104,92 @@ func (c ColumnDescs) String() string {
 
 // ColumnEncoding describes the encoding of a column.
 //
-// Null bitmap (bit 0):
+// Column data can optionally contain a NULL bitmap, where each bit corresponds
+// to a row in the column and a set bit indicates that the corresponding row is
+// NULL. Column data can optionally use a delta encoding where each value is a
+// delta from a base, constant value.
 //
-// The bit at position 0 indicates whether the column is prefixed with a null
-// bitmap. A null bitmap is a bitmap where each bit corresponds to a row in the
-// column, and a set bit indicates that the corresponding row is NULL.
+// The layout of the encoded data is as follows:
+//   - [optional] NULL bitmap
+//   - [optional] a "base" constant value for delta encodings
+//   - n data values
 //
-// TODO(jackson): Add additional column encoding types.
+// The ColumnEncoding byte is a combination of the following values:
+//   - The bits 0,1,2 encode the DeltaEncoding enum variant.
+//   - The bit 3 is set if the column has a NULL bitmap.
+//
+// The delta encoding is currently only used for uint columns.
 type ColumnEncoding uint8
 
 const (
 	// EncodingDefault indicates that the default encoding is in-use for a
-	// column, encoding n values for n rows.
+	// column, encoding n values for n rows. In EncodingDefault, no null bitmap
+	// is set, no delta encoding is used, etc.
 	EncodingDefault ColumnEncoding = 0
+
+	encodingUintDeltaMask        = 0b00000111
+	encodingUintDeltaInverseMask = 0b11111000
+	encodingNullBitmapBit        = 0b00001000
+
 	// TODO(jackson): Add additional encoding types.
-	encodingTypeCount = 1
 )
+
+// Delta returns the delta encoding of the column.
+func (e ColumnEncoding) Delta() DeltaEncoding {
+	return DeltaEncoding(e & encodingUintDeltaMask)
+}
+
+// WithDelta returns the column encoding with the provided delta encoding.
+func (e ColumnEncoding) WithDelta(d DeltaEncoding) ColumnEncoding {
+	return (e & encodingUintDeltaInverseMask) | ColumnEncoding(d)
+}
 
 // String returns the string representation of the column encoding.
 func (e ColumnEncoding) String() string {
-	return encodingName[e]
+	var sb strings.Builder
+	fmt.Fprint(&sb, e.Delta().String())
+	return sb.String()
 }
 
-var encodingName [encodingTypeCount]string = [encodingTypeCount]string{
-	EncodingDefault: "default",
+// DeltaEncoding indicates what delta encoding, if any is in use by a column to
+// reduce the per-row storage size.
+//
+// A uint delta encoding represents every non-NULL element in an array of uints
+// as a delta relative to the column's constant. The logical value of each row
+// is computed as C + D[i] where C is the column constant and D[i] is the delta.
+type DeltaEncoding uint8
+
+const (
+	// DeltaEncodingNone indicates no delta encoding is in use. N rows are
+	// represented using N values of the column's logical data type.
+	DeltaEncodingNone DeltaEncoding = 0
+	// DeltaEncodingConstant indicates that all rows of the column share the
+	// same value. The column data encodes the constant value and no deltas.
+	DeltaEncodingConstant DeltaEncoding = 1
+	// DeltaEncodingUint8 indicates each delta is represented as a 1-byte uint8.
+	DeltaEncodingUint8 DeltaEncoding = 2
+	// DeltaEncodingUint16 indicates each delta is represented as a 2-byte uint16.
+	DeltaEncodingUint16 DeltaEncoding = 3
+	// DeltaEncodingUint32 indicates each delta is represented as a 4-byte uint32.
+	DeltaEncodingUint32 DeltaEncoding = 4
+)
+
+// String implements fmt.Stringer.
+func (d DeltaEncoding) String() string {
+	switch d {
+	case DeltaEncodingNone:
+		return "none"
+	case DeltaEncodingConstant:
+		return "const"
+	case DeltaEncodingUint8:
+		return "delta8"
+	case DeltaEncodingUint16:
+		return "delta16"
+	case DeltaEncodingUint32:
+		return "delta32"
+	default:
+		panic("unreachable")
+	}
 }
 
 // ColumnWriter is an interface implemented by column encoders that accumulate a
@@ -137,6 +210,10 @@ type ColumnWriter interface {
 	//
 	// The provided column index must be less than NumColumns(). Finish is
 	// called for each index < NumColumns() in order.
+	//
+	// The provided buf must be word-aligned (at offset 0). If a column writer
+	// requires a particularly alignment, it's responsible for padding offset
+	// appropriately first.
 	Finish(col int, rows int, offset uint32, buf []byte) (nextOffset uint32, desc ColumnDesc)
 }
 
--- a/sstable/colblk/column_test.go
+++ b/sstable/colblk/column_test.go
@@ -1,0 +1,54 @@
+// Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package colblk
+
+import "testing"
+
+func TestColumnDesc(t *testing.T) {
+	type testCase struct {
+		desc  ColumnDesc
+		delta DeltaEncoding
+		str   string
+	}
+
+	testCases := []testCase{
+		{
+			desc:  ColumnDesc{DataType: DataTypeBool, Encoding: EncodingDefault},
+			delta: DeltaEncodingNone,
+			str:   "bool",
+		},
+		{
+			desc:  ColumnDesc{DataType: DataTypeUint32, Encoding: EncodingDefault},
+			delta: DeltaEncodingNone,
+			str:   "uint32",
+		},
+		{
+			desc: ColumnDesc{
+				DataType: DataTypeUint32,
+				Encoding: EncodingDefault.WithDelta(DeltaEncodingUint16),
+			},
+			delta: DeltaEncodingUint16,
+			str:   "uint32+delta16",
+		},
+		{
+			desc: ColumnDesc{
+				DataType: DataTypeUint64,
+				Encoding: EncodingDefault.WithDelta(DeltaEncodingConstant),
+			},
+			delta: DeltaEncodingConstant,
+			str:   "uint64+const",
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.str, func(t *testing.T) {
+			if got := tc.desc.String(); got != tc.str {
+				t.Errorf("String() = %q; want %q", got, tc.str)
+			}
+			if d := tc.desc.Encoding.Delta(); d != tc.delta {
+				t.Errorf(".Delta() = %q; want %q", d, tc.delta)
+			}
+		})
+	}
+}
--- a/sstable/colblk/testdata/uints
+++ b/sstable/colblk/testdata/uints
@@ -1,0 +1,617 @@
+# This test works by initializing a set of UintBuilders at configured logical
+# widths and setting the same elements to the same indexes. The builders' Sizes
+# are queried for various row sizes. Individual builders of specific widths may
+# be finished separately so the test can continue with testing higher-width
+# integers.
+
+# Initialize all four writers.
+
+init widths=(8, 16, 32, 64)
+----
+b8
+b16
+b32
+b64
+
+# Write a few zero values at index [0,4].
+
+write
+0:0 1:0 2:0 3:0 4:0
+----
+
+# At all row counts, the column should be encoded as a constant using the column
+# type width.
+
+size rows=(5, 4, 3, 2, 1, 0)
+----
+b8:
+  8: *colblk.UintBuilder[uint8].Size(5, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(4, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(3, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(2, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(1, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(0, 0) = 0
+b16:
+  16: *colblk.UintBuilder[uint16].Size(5, 0) = 2
+  16: *colblk.UintBuilder[uint16].Size(4, 0) = 2
+  16: *colblk.UintBuilder[uint16].Size(3, 0) = 2
+  16: *colblk.UintBuilder[uint16].Size(2, 0) = 2
+  16: *colblk.UintBuilder[uint16].Size(1, 0) = 2
+  16: *colblk.UintBuilder[uint16].Size(0, 0) = 0
+b32:
+  32: *colblk.UintBuilder[uint32].Size(5, 0) = 4
+  32: *colblk.UintBuilder[uint32].Size(4, 0) = 4
+  32: *colblk.UintBuilder[uint32].Size(3, 0) = 4
+  32: *colblk.UintBuilder[uint32].Size(2, 0) = 4
+  32: *colblk.UintBuilder[uint32].Size(1, 0) = 4
+  32: *colblk.UintBuilder[uint32].Size(0, 0) = 0
+b64:
+  64: *colblk.UintBuilder[uint64].Size(5, 0) = 8
+  64: *colblk.UintBuilder[uint64].Size(4, 0) = 8
+  64: *colblk.UintBuilder[uint64].Size(3, 0) = 8
+  64: *colblk.UintBuilder[uint64].Size(2, 0) = 8
+  64: *colblk.UintBuilder[uint64].Size(1, 0) = 8
+  64: *colblk.UintBuilder[uint64].Size(0, 0) = 0
+
+# Add a nonzero value. Size calls that include the new row count should
+# increase, but the size calls that don't include the new row count should not.
+# The increased sizes should reflect use of a uint8 delta encoding.
+
+write
+5:10
+6:0
+7:10
+----
+
+size rows=(8, 7, 6, 5, 4, 3, 2, 1, 0)
+----
+b8:
+  8: *colblk.UintBuilder[uint8].Size(8, 0) = 8
+  8: *colblk.UintBuilder[uint8].Size(7, 0) = 7
+  8: *colblk.UintBuilder[uint8].Size(6, 0) = 6
+  8: *colblk.UintBuilder[uint8].Size(5, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(4, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(3, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(2, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(1, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(0, 0) = 0
+b16:
+  16: *colblk.UintBuilder[uint16].Size(8, 0) = 10
+  16: *colblk.UintBuilder[uint16].Size(7, 0) = 9
+  16: *colblk.UintBuilder[uint16].Size(6, 0) = 8
+  16: *colblk.UintBuilder[uint16].Size(5, 0) = 2
+  16: *colblk.UintBuilder[uint16].Size(4, 0) = 2
+  16: *colblk.UintBuilder[uint16].Size(3, 0) = 2
+  16: *colblk.UintBuilder[uint16].Size(2, 0) = 2
+  16: *colblk.UintBuilder[uint16].Size(1, 0) = 2
+  16: *colblk.UintBuilder[uint16].Size(0, 0) = 0
+b32:
+  32: *colblk.UintBuilder[uint32].Size(8, 0) = 12
+  32: *colblk.UintBuilder[uint32].Size(7, 0) = 11
+  32: *colblk.UintBuilder[uint32].Size(6, 0) = 10
+  32: *colblk.UintBuilder[uint32].Size(5, 0) = 4
+  32: *colblk.UintBuilder[uint32].Size(4, 0) = 4
+  32: *colblk.UintBuilder[uint32].Size(3, 0) = 4
+  32: *colblk.UintBuilder[uint32].Size(2, 0) = 4
+  32: *colblk.UintBuilder[uint32].Size(1, 0) = 4
+  32: *colblk.UintBuilder[uint32].Size(0, 0) = 0
+b64:
+  64: *colblk.UintBuilder[uint64].Size(8, 0) = 16
+  64: *colblk.UintBuilder[uint64].Size(7, 0) = 15
+  64: *colblk.UintBuilder[uint64].Size(6, 0) = 14
+  64: *colblk.UintBuilder[uint64].Size(5, 0) = 8
+  64: *colblk.UintBuilder[uint64].Size(4, 0) = 8
+  64: *colblk.UintBuilder[uint64].Size(3, 0) = 8
+  64: *colblk.UintBuilder[uint64].Size(2, 0) = 8
+  64: *colblk.UintBuilder[uint64].Size(1, 0) = 8
+  64: *colblk.UintBuilder[uint64].Size(0, 0) = 0
+
+# Finish the b8 so we can test 16-bit encoding.
+finish widths=(8) rows=8
+----
+b8: *colblk.UintBuilder[uint8]:
+0-1: x 00 # data[0] = 0
+1-2: x 00 # data[1] = 0
+2-3: x 00 # data[2] = 0
+3-4: x 00 # data[3] = 0
+4-5: x 00 # data[4] = 0
+5-6: x 0a # data[5] = 10
+6-7: x 00 # data[6] = 0
+7-8: x 0a # data[7] = 10
+Keeping b16 open
+Keeping b32 open
+Keeping b64 open
+
+# Add 1000 which should force a 16-bit delta encoding.
+
+write
+8:1000
+----
+
+size rows=(9, 8)
+----
+b16:
+  16: *colblk.UintBuilder[uint16].Size(9, 0) = 18
+  16: *colblk.UintBuilder[uint16].Size(8, 0) = 10
+b32:
+  32: *colblk.UintBuilder[uint32].Size(9, 0) = 22
+  32: *colblk.UintBuilder[uint32].Size(8, 0) = 12
+b64:
+  64: *colblk.UintBuilder[uint64].Size(9, 0) = 26
+  64: *colblk.UintBuilder[uint64].Size(8, 0) = 16
+
+# We should be able to write up to 2^16-1 without triggering a 32-bit encoding.
+
+write
+9:65535
+----
+
+size rows=(10, 9, 8)
+----
+b16:
+  16: *colblk.UintBuilder[uint16].Size(10, 0) = 20
+  16: *colblk.UintBuilder[uint16].Size(9, 0) = 18
+  16: *colblk.UintBuilder[uint16].Size(8, 0) = 10
+b32:
+  32: *colblk.UintBuilder[uint32].Size(10, 0) = 24
+  32: *colblk.UintBuilder[uint32].Size(9, 0) = 22
+  32: *colblk.UintBuilder[uint32].Size(8, 0) = 12
+b64:
+  64: *colblk.UintBuilder[uint64].Size(10, 0) = 28
+  64: *colblk.UintBuilder[uint64].Size(9, 0) = 26
+  64: *colblk.UintBuilder[uint64].Size(8, 0) = 16
+
+# But 2^16 should trigger a 32-bit encoding. (Finish b16 so we can test 32-bit
+# encoding.)
+
+finish widths=(16) rows=10
+----
+b16: *colblk.UintBuilder[uint16]:
+00-02: x 0000 # data[0] = 0
+02-04: x 0000 # data[1] = 0
+04-06: x 0000 # data[2] = 0
+06-08: x 0000 # data[3] = 0
+08-10: x 0000 # data[4] = 0
+10-12: x 0a00 # data[5] = 10
+12-14: x 0000 # data[6] = 0
+14-16: x 0a00 # data[7] = 10
+16-18: x e803 # data[8] = 1000
+18-20: x ffff # data[9] = 65535
+Keeping b32 open
+Keeping b64 open
+
+write
+10:65536
+----
+
+size rows=(11, 10, 9, 8)
+----
+b32:
+  32: *colblk.UintBuilder[uint32].Size(11, 0) = 44
+  32: *colblk.UintBuilder[uint32].Size(10, 0) = 24
+  32: *colblk.UintBuilder[uint32].Size(9, 0) = 22
+  32: *colblk.UintBuilder[uint32].Size(8, 0) = 12
+b64:
+  64: *colblk.UintBuilder[uint64].Size(11, 0) = 52
+  64: *colblk.UintBuilder[uint64].Size(10, 0) = 28
+  64: *colblk.UintBuilder[uint64].Size(9, 0) = 26
+  64: *colblk.UintBuilder[uint64].Size(8, 0) = 16
+
+# We should be able to write up to 2^32-1 without triggering a 64-bit encoding.
+
+write
+11:4294967295
+----
+
+size rows=(12, 11, 10, 9, 8)
+----
+b32:
+  32: *colblk.UintBuilder[uint32].Size(12, 0) = 48
+  32: *colblk.UintBuilder[uint32].Size(11, 0) = 44
+  32: *colblk.UintBuilder[uint32].Size(10, 0) = 24
+  32: *colblk.UintBuilder[uint32].Size(9, 0) = 22
+  32: *colblk.UintBuilder[uint32].Size(8, 0) = 12
+b64:
+  64: *colblk.UintBuilder[uint64].Size(12, 0) = 56
+  64: *colblk.UintBuilder[uint64].Size(11, 0) = 52
+  64: *colblk.UintBuilder[uint64].Size(10, 0) = 28
+  64: *colblk.UintBuilder[uint64].Size(9, 0) = 26
+  64: *colblk.UintBuilder[uint64].Size(8, 0) = 16
+
+# But 2^32 should trigger a 64-bit encoding.
+
+finish widths=(32) rows=12
+----
+b32: *colblk.UintBuilder[uint32]:
+00-04: x 00000000 # data[0] = 0
+04-08: x 00000000 # data[1] = 0
+08-12: x 00000000 # data[2] = 0
+12-16: x 00000000 # data[3] = 0
+16-20: x 00000000 # data[4] = 0
+20-24: x 0a000000 # data[5] = 10
+24-28: x 00000000 # data[6] = 0
+28-32: x 0a000000 # data[7] = 10
+32-36: x e8030000 # data[8] = 1000
+36-40: x ffff0000 # data[9] = 65535
+40-44: x 00000100 # data[10] = 65536
+44-48: x ffffffff # data[11] = 4294967295
+Keeping b64 open
+
+write
+12:4294967296
+----
+
+size rows=(13, 12, 11, 10, 9, 8)
+----
+b64:
+  64: *colblk.UintBuilder[uint64].Size(13, 0) = 104
+  64: *colblk.UintBuilder[uint64].Size(12, 0) = 56
+  64: *colblk.UintBuilder[uint64].Size(11, 0) = 52
+  64: *colblk.UintBuilder[uint64].Size(10, 0) = 28
+  64: *colblk.UintBuilder[uint64].Size(9, 0) = 26
+  64: *colblk.UintBuilder[uint64].Size(8, 0) = 16
+
+finish widths=(64) rows=13
+----
+b64: *colblk.UintBuilder[uint64]:
+000-008: x 0000000000000000 # data[0] = 0
+008-016: x 0000000000000000 # data[1] = 0
+016-024: x 0000000000000000 # data[2] = 0
+024-032: x 0000000000000000 # data[3] = 0
+032-040: x 0000000000000000 # data[4] = 0
+040-048: x 0a00000000000000 # data[5] = 10
+048-056: x 0000000000000000 # data[6] = 0
+056-064: x 0a00000000000000 # data[7] = 10
+064-072: x e803000000000000 # data[8] = 1000
+072-080: x ffff000000000000 # data[9] = 65535
+080-088: x 0000010000000000 # data[10] = 65536
+088-096: x ffffffff00000000 # data[11] = 4294967295
+096-104: x 0000000001000000 # data[12] = 4294967296
+
+# Repeat the above tests but with a zero default value, and without explicitly
+# setting any of the zero values.
+
+init widths=(8, 16, 32, 64) default-zero
+----
+b8
+b16
+b32
+b64
+
+# At all row counts, the column should be encoded as a constant using the column
+# type width.
+
+size rows=(5, 4, 3, 2, 1, 0)
+----
+b8:
+  8: *colblk.UintBuilder[uint8].Size(5, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(4, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(3, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(2, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(1, 0) = 1
+  8: *colblk.UintBuilder[uint8].Size(0, 0) = 0
+b16:
+  16: *colblk.UintBuilder[uint16].Size(5, 0) = 2
+  16: *colblk.UintBuilder[uint16].Size(4, 0) = 2
+  16: *colblk.UintBuilder[uint16].Size(3, 0) = 2
+  16: *colblk.UintBuilder[uint16].Size(2, 0) = 2
+  16: *colblk.UintBuilder[uint16].Size(1, 0) = 2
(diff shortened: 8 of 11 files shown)


RaduBerinde (commented):
Cool! Made a first pass but haven't looked at the testing code yet.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731#-:-O0u0DUl8deWJmNQeeg5:bbug1gh)* status: 0 of 9 files reviewed, 10 unresolved discussions (waiting on @aadityasondhi, @jbowens, @petermattis, and @sumeerbhola)

___
*[`internal/aligned/aligned.go` line 16 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0t_PgK3Zl1GusBPkZX:-O0t_PgK3Zl1GusBPkZY:bw5k7j1) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/internal/aligned/aligned.go#L16)):*
> ```Go
> // but small n can often be misaligned.
> func ByteSlice(n int) []byte {
> 	// Allocate n + align64 - 1 bytes to ensure we have room to align.
> ```

Does Go guarantee that a `[]uint64` is aligned? (I would be surprised if it didn't). We could do `make([]uint64, (n+align64-1)/align64)` and do an unsafe cast to `[]byte`. That way we won't allocate weird lengths when our lengths are "nice" (eg powers of 2)
___
*[`sstable/colblk/column.go` line 61 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tZTVH1lhkDfr2Y3D7:-O0tZTVH1lhkDfr2Y3D8:bezz7je) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/column.go#L61)):*
> ```Go
> func (t DataType) uintWidth() uint32 {
> 	switch t {
> 	case DataTypeUint8:
> ```

[nit] could be
```
  if d >= DataTypeUint8 && d <= DataTypeUint64 {
    return 1 << (d - DataTypeUint8)
  }
  panic()
```
___
*[`sstable/colblk/column.go` line 111 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tsnUr8V6oTqlnb5Ty:-O0tsnUr8V6oTqlnb5Tz:bgu0tlj) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/column.go#L111)):*
> ```Go
> // ColumnEncoding describes the encoding of a column.
> //
> // Delta encoding (bits 0, 1):
> ```

Here we are trying to convey three things
 - how the possible encodings work at a high level
 - how the encoded data is organized
 - how we encode this information into the uint8

We shouldn't need to read through descriptions of the lower aspects in order to understand the higher aspects. I'd focus on these three aspects separately. Here's an attempt (assuming I understood the details correctly).

```
// ColumnEncoding describes the data encoding for a single column.
//
// Column data can optionally contain a NULL bitmap, where each bit corresponds
// to a row in the column and a set bit indicates that the corresponding row is
// NULL). Column data can optionally use delta encoding where each value is a
// delta from a base, constant value.
//
// The layout of the encoded data is as follows:
//   - [optional] NULL Bitmap
//   - [optional] a "base" constant value for delta encodings
//   - n data values.
//
// The ColumnEncoding byte is a combination of the following values:
//   - delta encoding indicator (0 to 1): indicates if delta encoding is used.
//   - delta encoding byte width (0 to 3): the byte width of the data values
//     representing the deltas; 0 indicates all values are equal.
//   - null bitmap indicator (0 to 1)
```
___
*[`sstable/colblk/column.go` line 175 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0u-O-WCZlX8sE5zwnS:-O0u-O-WCZlX8sE5zwnT:b-vk14pl) ([raw file](https://github.com/cockroachdb/pebble/blob/dc95e51b0638a9f12bb93f3ffd3476cbf6abc191/sstable/colblk/column.go#L175)):*
> ```Go
> 	var sb strings.Builder
> 	if e.HasConstant() {
> 		fmt.Fprint(&sb, "const")
> ```

[nit] wouldn't "delta" be more appropriate? In general, I find it strange that the focus is on whether there is a constant rather than whether we are using a delta encoding. I guess the constant will make more sense as we introduce more encodings/types? (for bytes values maybe it's a byte prefix for all of them? though I would also consider that a "delta" encoding at a high level)
___
*[`sstable/colblk/column.go` line 192 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0u0-R22PPO0G5pctDo:-O0u0-R22PPO0G5pctDp:bccml42) ([raw file](https://github.com/cockroachdb/pebble/blob/dc95e51b0638a9f12bb93f3ffd3476cbf6abc191/sstable/colblk/column.go#L192)):*
> ```Go
> 	// indicates a constant is present, the entirety of the column is a single
> 	// value. Otherwise, the column encodes n full-width values for n rows.
> 	DeltaEncodingNone DeltaEncoding = 0
> ```

[nit] I think it would be useful to differentiate between "no delta encoding" and "delta encoding with no data". Or rename to `DeltaEncodingWidth` but then don't define values like it's an enum.
___
*[`sstable/colblk/uints.go` line 32 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tvBL44NT0l3cp6iTE:-O0tvBL44NT0l3cp6iTF:b-ml5uda) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/uints.go#L32)):*
> ```Go
> 
> 	// array holds the underlying heap-allocated array in which values are
> 	// stored. elems provides access to elements without bounds checking. n is
> ```

[nit] move the comment for `n` above `n int`
___
*[`sstable/colblk/uints.go` line 37 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0twx4z3LnXsro5nEnI:-O0twx4z3LnXsro5nEnJ:b8vitas) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/uints.go#L37)):*
> ```Go
> 	array struct {
> 		n     int
> 		elems UnsafeRawSlice[T]
> ```

[nit] mention that elems is a slice that grows automatically
___
*[`sstable/colblk/uints.go` line 71 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tw1plAYew8Hn5fiRo:-O0tw1plAYew8Hn5fiRp:b-32837h) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/uints.go#L71)):*
> ```Go
> // Init initializes the UintBuilder. If useDefault=true, any rows that are not
> // explicitly set are assumed to be the provided default value. For the purpose
> // of determining whether a delta encoding is possible, the column is assumed to
> ```

It wasn't clear from this comment if the last statement is only true when `setDefault` is true.
It feels weird to have to pass an inconsequential value when `useDefault` is false.. Maybe split into `Init` and `InitWithDefaultValue`?


By the way, why is that assumption necessary? Can't we calculate all th delta stuff in `Finish`?
___
*[`sstable/colblk/unsafe_slice.go` line 43 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0txmRA8YDXm7Lusx6Z:-O0txmRA8YDXm7Lusx6_:b-4mua2y) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/unsafe_slice.go#L43)):*
> ```Go
> }
> 
> // UnsafeUint8s is an UnsafeRawSlice of uint8s.
> ```

UnsafeIntegerSlice. But maybe it would be more useful something like "is a slice of uint8s, possibly using delta encoding internally".
___
*[`sstable/colblk/unsafe_slice.go` line 81 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tzbLGELB4_yfvgJwS:-O0tzbLGELB4_yfvgJwT:b-li78hk) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/unsafe_slice.go#L81)):*
> ```Go
> // At returns the `i`-th element of the slice.
> func (s UnsafeIntegerSlice[T]) At(i int) T {
> 	switch s.deltaWidth {
> ```

This branching in the hot path is unfortunate.. Maybe the compiler optimizes it nicely with a table..

I wonder if it would be feasible (depending on how it's used) to make it generic in both the visible type and the underlying type (`UnsafeIntegerSlice[T constraints.Integer, U constraints.Integer]`).

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731)* status: 0 of 9 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi, @petermattis, @RaduBerinde, and @sumeerbhola)

___
*[`internal/aligned/aligned.go` line 16 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0t_PgK3Zl1GusBPkZX:-O0uT3486gB5Q26Bpt3k:bap0xug) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/internal/aligned/aligned.go#L16)):*

Does Go guarantee that a `[]uint64` is aligned? (I would be surprised if it didn't). We could do `make([]uint64, (n+align64-1)/align64)` and do an unsafe cast to `[]byte`. That way we won't allocate weird lengths when our lengths are "nice" (eg powers of 2)
</blockquote></details>

Yeah, from some reading I believe it does. That's a good idea. In real block cache allocations, we'll use `malloc` and have a word-alignment guarantee.
___
*[`sstable/colblk/column.go` line 61 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tZTVH1lhkDfr2Y3D7:-O0uTzXr7KR-mtm39cPd:b-esfcfc) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/column.go#L61)):*

[nit] could be
```
  if d >= DataTypeUint8 && d <= DataTypeUint64 {
    return 1 << (d - DataTypeUint8)
  }
  panic()
```
</blockquote></details>

nice, done
___
*[`sstable/colblk/column.go` line 111 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tsnUr8V6oTqlnb5Ty:-O0uS-ue-w-EL1_lWDvT:b3tylto) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/column.go#L111)):*

Here we are trying to convey three things
 - how the possible encodings work at a high level
 - how the encoded data is organized
 - how we encode this information into the uint8

We shouldn't need to read through descriptions of the lower aspects in order to understand the higher aspects. I'd focus on these three aspects separately. Here's an attempt (assuming I understood the details correctly).

```
// ColumnEncoding describes the data encoding for a single column.
//
// Column data can optionally contain a NULL bitmap, where each bit corresponds
// to a row in the column and a set bit indicates that the corresponding row is
// NULL). Column data can optionally use delta encoding where each value is a
// delta from a base, constant value.
//
// The layout of the encoded data is as follows:
//   - [optional] NULL Bitmap
//   - [optional] a "base" constant value for delta encodings
//   - n data values.
//
// The ColumnEncoding byte is a combination of the following values:
//   - delta encoding indicator (0 to 1): indicates if delta encoding is used.
//   - delta encoding byte width (0 to 3): the byte width of the data values
//     representing the deltas; 0 indicates all values are equal.
//   - null bitmap indicator (0 to 1)
```
</blockquote></details>

thanks, agreed that's much better
___
*[`sstable/colblk/column.go` line 175 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0u-O-WCZlX8sE5zwnS:-O0uS2j73V8M8cmp_J94:b-u33f6x) ([raw file](https://github.com/cockroachdb/pebble/blob/dc95e51b0638a9f12bb93f3ffd3476cbf6abc191/sstable/colblk/column.go#L175)):*

[nit] wouldn't "delta" be more appropriate? In general, I find it strange that the focus is on whether there is a constant rather than whether we are using a delta encoding. I guess the constant will make more sense as we introduce more encodings/types? (for bytes values maybe it's a byte prefix for all of them? though I would also consider that a "delta" encoding at a high level)
</blockquote></details>

Yeah, I agree I found this organization a little unsatisfying. I was trying to preserve the flexibility to use the 'constant bit' for other column types that could encode a single value representing the value for all rows. Eg, if all rows had the same byte slice, the RawBytes encoding could set the constant bit and only encode a single byte slice. I don't think it's realistic for most data types, so I've taken your suggestion of collapsing the the constant w/ no deltas option into the DeltaEncoding enum. 
___
*[`sstable/colblk/column.go` line 192 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0u0-R22PPO0G5pctDo:-O0uStyiBE51TqnblYgz:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/dc95e51b0638a9f12bb93f3ffd3476cbf6abc191/sstable/colblk/column.go#L192)):*

[nit] I think it would be useful to differentiate between "no delta encoding" and "delta encoding with no data". Or rename to `DeltaEncodingWidth` but then don't define values like it's an enum.
</blockquote></details>

Done
___
*[`sstable/colblk/uints.go` line 71 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tw1plAYew8Hn5fiRo:-O0uHjBa0wP7RcTAPY2D:bkedede) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/uints.go#L71)):*
> Maybe split into `Init` and `InitWithDefaultValue`?

Done.

> By the way, why is that assumption necessary? Can't we calculate all th delta stuff in `Finish`?

We need the Size to be accurate, but also performant, because we call it after every additional kv to determine whether we should finish the block.
___
*[`sstable/colblk/unsafe_slice.go` line 43 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0txmRA8YDXm7Lusx6Z:-O0uIW4d2WzfARGpPH2m:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/unsafe_slice.go#L43)):*

UnsafeIntegerSlice. But maybe it would be more useful something like "is a slice of uint8s, possibly using delta encoding internally".
</blockquote></details>

done
___
*[`sstable/colblk/unsafe_slice.go` line 81 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tzbLGELB4_yfvgJwS:-O0uIhDF4jvUHC88nAkD:b-udmxdj) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/unsafe_slice.go#L81)):*

This branching in the hot path is unfortunate.. Maybe the compiler optimizes it nicely with a table..

I wonder if it would be feasible (depending on how it's used) to make it generic in both the visible type and the underlying type (`UnsafeIntegerSlice[T constraints.Integer, U constraints.Integer]`).
</blockquote></details>

Yeah, it's unfortunate. I think the visible type will always be statically known, but the underlying type will never be.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731)* status: 0 of 9 files reviewed, 8 unresolved discussions (waiting on @aadityasondhi, @jbowens, @petermattis, and @sumeerbhola)

___
*[`sstable/colblk/column.go` line 175 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0u-O-WCZlX8sE5zwnS:-O0uWVxR5O4dpzwA7r1D:b93qqi2) ([raw file](https://github.com/cockroachdb/pebble/blob/dc95e51b0638a9f12bb93f3ffd3476cbf6abc191/sstable/colblk/column.go#L175)):*

Yeah, I agree I found this organization a little unsatisfying. I was trying to preserve the flexibility to use the 'constant bit' for other column types that could encode a single value representing the value for all rows. Eg, if all rows had the same byte slice, the RawBytes encoding could set the constant bit and only encode a single byte slice. I don't think it's realistic for most data types, so I've taken your suggestion of collapsing the the constant w/ no deltas option into the DeltaEncoding enum.
</blockquote></details>

The RawBytes example is analogous to a constant int column, so we can still call that "delta encoding". In fact for raw bytes we may want to make the constant be any common prefix of the values (with all equal being a special case of that).
___
*[`sstable/colblk/column.go` line 174 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0uX91v8_TSaSCLkaFP:-O0uX91v8_TSaSCLkaFQ:b-m809ht) ([raw file](https://github.com/cockroachdb/pebble/blob/01493604e56e69ce1ad93f8cf7c9179c8d760af8/sstable/colblk/column.go#L174)):*
> ```Go
> 	switch d {
> 	case DeltaEncodingNone:
> 		return ""
> ```

[nit] I'd make this "none" so we don't end up with confusing logs
___
*[`sstable/colblk/uints.go` line 71 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tw1plAYew8Hn5fiRo:-O0uXz9OAjPD3YJZBxtP:b-w6jxn4) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/uints.go#L71)):*

> Maybe split into `Init` and `InitWithDefaultValue`?

Done.

> By the way, why is that assumption necessary? Can't we calculate all th delta stuff in `Finish`?

We need the Size to be accurate, but also performant, because we call it after every additional kv to determine whether we should finish the block.
</blockquote></details>

I see.. It sounds like `Size` can suddenly double when adding a value.. I wonder if we should have an API that would allow you to finish the block before the doubling. Maybe configuring the builder with a maximum and allowing a `Set` to fail? But that assumes the rows are populated in order.
___
*[`sstable/colblk/uints.go` line 134 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0uXi0P19rLAbcfBUC-:-O0uXi0P19rLAbcfBUC0:b-onm10g) ([raw file](https://github.com/cockroachdb/pebble/blob/01493604e56e69ce1ad93f8cf7c9179c8d760af8/sstable/colblk/uints.go#L134)):*
> ```Go
> 			n2 <<= 1 /* double the size */
> 		}
> 		newData := make([]byte, n2*int(unsafe.Sizeof(T(0))))
> ```

Should we use `ByteSlice` here? Seems like we'd want it aligned.
___
*[`sstable/colblk/unsafe_slice.go` line 81 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tzbLGELB4_yfvgJwS:-O0uZ_CD05P-4qv2G0EY:b-lv8uvo) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/unsafe_slice.go#L81)):*

Yeah, it's unfortunate. I think the visible type will always be statically known, but the underlying type will never be.
</blockquote></details>

Could we always store a `uint64` slice internally and have some precomputed shifts and masks?
```
vals[i>>indexShift] >> ((i & indexMask) << indexShift) & valMask
```
where `indexShift` is log2(byte width), `indexMask = (1 << indexShift)-1`, `valMask = (1 << (1 << indexShift)) - 1`
___
*[`sstable/colblk/unsafe_slice.go` line 89 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0uZTwFA2hJ9jR8IIXK:-O0uZTwFA2hJ9jR8IIXL:bdd9k31) ([raw file](https://github.com/cockroachdb/pebble/blob/01493604e56e69ce1ad93f8cf7c9179c8d760af8/sstable/colblk/unsafe_slice.go#L89)):*
> ```Go
> 		return s.base
> 	case 1:
> 		return s.base + T(*(*int8)(unsafe.Pointer(uintptr(s.deltaPtr) + uintptr(i))))
> ```

Why are these ints signed?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731)* status: 0 of 9 files reviewed, 8 unresolved discussions (waiting on @aadityasondhi, @petermattis, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk/uints.go` line 71 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tw1plAYew8Hn5fiRo:-O0y2Usw-l4HtU0qst2_:b-1whnoo) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/uints.go#L71)):*

I see.. It sounds like `Size` can suddenly double when adding a value.. I wonder if we should have an API that would allow you to finish the block before the doubling. Maybe configuring the builder with a maximum and allowing a `Set` to fail? But that assumes the rows are populated in order.
</blockquote></details>

I probably did a poor job explaining this in the comments on `ColumnWriter.{Size,Finish}`, but the intention is that the caller can:

1. Add a kv `n-1` to a block and its columns.
2. `size := blockWriter.Size()`
3. Add a kv `n` to a block and its columns.
4. `newSize := blockWriter.Size()`
5. If `newSize` is too big (and, with the new size class heuristics, the old `size` is closer to the size class it falls into), the caller can allocate a buffer for `size` and `blockWriter.Finish(n-1)` to serialize the block without kv `n`.

This avoids some tricky (esp with prefix compression) estimation involved in trying to determine what the block size with a new KV would be.

There's still a wart in the above scheme where if you have large KVs, you end up memcpying everything an extra time: once into the previous block and once into the block it ultimately is written to. I was talking with @sumeerbhola about this in the office.  We might be able to avoid the actual copying of values and keys until we've committed to including the KV in the final block, but we decided that could be a follow-up consideration.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731)* status: 0 of 9 files reviewed, 8 unresolved discussions (waiting on @aadityasondhi, @jbowens, @petermattis, and @sumeerbhola)

___
*[`sstable/colblk/uints.go` line 71 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tw1plAYew8Hn5fiRo:-O0y_iY7662_kxjfh1bh:blfqo3a) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/uints.go#L71)):*

I probably did a poor job explaining this in the comments on `ColumnWriter.{Size,Finish}`, but the intention is that the caller can:

1. Add a kv `n-1` to a block and its columns.
2. `size := blockWriter.Size()`
3. Add a kv `n` to a block and its columns.
4. `newSize := blockWriter.Size()`
5. If `newSize` is too big (and, with the new size class heuristics, the old `size` is closer to the size class it falls into), the caller can allocate a buffer for `size` and `blockWriter.Finish(n-1)` to serialize the block without kv `n`.

This avoids some tricky (esp with prefix compression) estimation involved in trying to determine what the block size with a new KV would be.

There's still a wart in the above scheme where if you have large KVs, you end up memcpying everything an extra time: once into the previous block and once into the block it ultimately is written to. I was talking with @sumeerbhola about this in the office.  We might be able to avoid the actual copying of values and keys until we've committed to including the KV in the final block, but we decided that could be a follow-up consideration.
</blockquote></details>

Yeah, we could make setting a value a prepare/commit style API:

```
func SetKeyPrepare() SetKeyCommitter
func (SetKeyCommitter) NewSize()
func (SetKeyCommitter) Commit()
```

We only copy the actual data in `Commit`. The `SetKeyCommitter` stores the updated internal state for delta values, which we can just copy directly in `Commit`. We can just discard the `SetKeyCommitter` if we don't want the value in the block.

I think it's simple enough and makes the API cleaner to use - it might be overall less work to just do this from the start, especially if we'll have multiple implementation with similar API.
___
*[`sstable/colblk/uints.go` line 118 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0ybrbK4LWE1Xh9_syQ:-O0ybrbK4LWE1Xh9_syR:b-aztsbn) ([raw file](https://github.com/cockroachdb/pebble/blob/01493604e56e69ce1ad93f8cf7c9179c8d760af8/sstable/colblk/uints.go#L118)):*
> ```Go
> 	} else {
> 		b.delta.minimum = T(0) - 1
> 		b.delta.maximum = 0
> ```

Seems error prone that we don't reset the array values here, even if we expect all of them to get set.
___
*[`sstable/colblk/uints.go` line 125 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0yb8k39bAar2n_J_Iz:-O0yb8k39bAar2n_J_J-:bhxjj9) ([raw file](https://github.com/cockroachdb/pebble/blob/01493604e56e69ce1ad93f8cf7c9179c8d760af8/sstable/colblk/uints.go#L125)):*
> ```Go
> 
> // Set sets the value of the provided row index to v.
> func (b *UintBuilder[T]) Set(row int, v T) {
> ```

Will we ever need to set rows out-of-order? If not, it would be good to require increasing `row` here. It would make some things easier, especially for prefix compression.
In fact, if we're not using the default, we'd require `row` to be the last value plus one

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731)* status: 0 of 9 files reviewed, 7 unresolved discussions (waiting on @aadityasondhi, @petermattis, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk/uints.go` line 71 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tw1plAYew8Hn5fiRo:-O1HifrO9MSr22AHuFmG:b6wb0mf) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/uints.go#L71)):*

Yeah, we could make setting a value a prepare/commit style API:

```
func SetKeyPrepare() SetKeyCommitter
func (SetKeyCommitter) NewSize()
func (SetKeyCommitter) Commit()
```

We only copy the actual data in `Commit`. The `SetKeyCommitter` stores the updated internal state for delta values, which we can just copy directly in `Commit`. We can just discard the `SetKeyCommitter` if we don't want the value in the block.

I think it's simple enough and makes the API cleaner to use - it might be overall less work to just do this from the start, especially if we'll have multiple implementation with similar API.
</blockquote></details>

I think it may end up being a pessimization, since the block-building path is hot and the resulting API adds more overhead. I'm going to wait.
___
*[`sstable/colblk/uints.go` line 118 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0ybrbK4LWE1Xh9_syQ:-O1Hj86KEd3LskwSdD9j:b8b9q5h) ([raw file](https://github.com/cockroachdb/pebble/blob/01493604e56e69ce1ad93f8cf7c9179c8d760af8/sstable/colblk/uints.go#L118)):*

Seems error prone that we don't reset the array values here, even if we expect all of them to get set.
</blockquote></details>

In the data block building benchmarks, resetting these values comes at a significant cost.
___
*[`sstable/colblk/uints.go` line 125 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0yb8k39bAar2n_J_Iz:-O1HkF0R9aFeBtvyv4_g:b-gjqwp7) ([raw file](https://github.com/cockroachdb/pebble/blob/01493604e56e69ce1ad93f8cf7c9179c8d760af8/sstable/colblk/uints.go#L125)):*

Will we ever need to set rows out-of-order? If not, it would be good to require increasing `row` here. It would make some things easier, especially for prefix compression.
In fact, if we're not using the default, we'd require `row` to be the last value plus one
</blockquote></details>

No, but storing the row per-column on every new KV is a significant cost in the data block building benchmarks.
___
*[`sstable/colblk/uints.go` line 134 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0uXi0P19rLAbcfBUC-:-O1HluhHCYhVcSWf6V0i:b-6ced5m) ([raw file](https://github.com/cockroachdb/pebble/blob/01493604e56e69ce1ad93f8cf7c9179c8d760af8/sstable/colblk/uints.go#L134)):*

Should we use `ByteSlice` here? Seems like we'd want it aligned.
</blockquote></details>

updated to allocate `[]T`
___
*[`sstable/colblk/unsafe_slice.go` line 81 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tzbLGELB4_yfvgJwS:-O1HlknZDsm2Zsz6K3sY:bw41igr) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/unsafe_slice.go#L81)):*

Could we always store a `uint64` slice internally and have some precomputed shifts and masks?
```
vals[i>>indexShift] >> ((i & indexMask) << indexShift) & valMask
```
where `indexShift` is log2(byte width), `indexMask = (1 << indexShift)-1`, `valMask = (1 << (1 << indexShift)) - 1`
</blockquote></details>

not sure I follow. what do you mean store a `uint64` slice internally?
___
*[`sstable/colblk/unsafe_slice.go` line 89 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0uZTwFA2hJ9jR8IIXK:-O1HkcMi47fhOU_rWwJi:b5ttvwd) ([raw file](https://github.com/cockroachdb/pebble/blob/01493604e56e69ce1ad93f8cf7c9179c8d760af8/sstable/colblk/unsafe_slice.go#L89)):*

Why are these ints signed?
</blockquote></details>

Thanks—forgot to update these when I made a switch to using uints.

sumeerbhola (changes requested):
> Reviewed 1 of 7 files at r1, 1 of 5 files at r2, 1 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731)* status: 3 of 9 files reviewed, 8 unresolved discussions (waiting on @aadityasondhi, @jbowens, @petermattis, and @RaduBerinde)

___
*[`internal/aligned/aligned.go` line 16 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0t_PgK3Zl1GusBPkZX:-O1IYoE75fGaUVtvfbO1:b50rbx6) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/internal/aligned/aligned.go#L16)):*
does the unsafe cast approach not work?

> In real block cache allocations, we'll use `malloc` and have a word-alignment guarantee.

Is the idea here that if the block start is word aligned, there is nothing in the block layout that will have broken that alignment for individual columns (e.g. the Bitmap column)? If yes, is that documented somewhere?

And if `ByteSlice` is needed by the writer, the block cache is irrelevant, yes?
___
*[`sstable/colblk/column.go` line 61 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1IaHcyD2jSjkEV22nZ:-O1IaHcyD2jSjkEV22n_:bde75hg) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/column.go#L61)):*
> ```Go
> func (t DataType) uintWidth() uint32 {
> 	if t >= DataTypeUint8 && t <= DataTypeUint64 {
> 		return 1 << (t - DataTypeUint8)
> ```

can you add an assertion that this is <= 8.
___
*[`sstable/colblk/column.go` line 121 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1IazN80te1tzdhw3IW:-O1IazN80te1tzdhw3IX:b2463d7) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/column.go#L121)):*
> ```Go
> 	// EncodingDefault indicates that the default encoding is in-use for a
> 	// column, encoding n values for n rows.
> 	EncodingDefault ColumnEncoding = 0
> ```

What does "default encoding" mean when bits 0-3 are used to represent these other things?
It seems to me that that `ColumnDesc` has a 1 byte data type and a 1 byte encoding. The interpretation of the encoding byte is dependent on the data type. Some data types may have only one encoding so the encoding byte would be 0.
If this is correct, should there be a code comment stating that these DeltaEncoding bits are only for `DataTypeUint*`?

(flushing since we had a longer in-person discussion about this)
___
*[`sstable/colblk/uints.go` line 203 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1IZFaL9sW4uplY-Gw7:-O1IZFaL9sW4uplY-Gw8:b-fxp59b) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L203)):*
> ```Go
> }
> 
> func (b *UintBuilder[T]) Finish(col, rows int, offset uint32, buf []byte) (uint32, ColumnDesc) {
> ```

is there an alignment requirement on buf here?

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731)* status: 3 of 9 files reviewed, 12 unresolved discussions (waiting on @aadityasondhi, @jbowens, @petermattis, and @RaduBerinde)

___
*[`sstable/colblk/uints.go` line 54 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1Io12h5t6BM1w7vmEZ:-O1Io12h5t6BM1w7vmE_:b-naqdlm) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L54)):*
> ```Go
> 	//
> 	// Any call to Size or Finish that supplies [rows] that's inclusive of the
> 	// index stored in minMaxRow may use the stored width. Calls with fewer
> ```

what is `minMaxRow`?
___
*[`sstable/colblk/uints.go` line 55 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1IoBhD66gm1eQlenNq:-O1IoBhD66gm1eQlenNr:btxtx28) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L55)):*
> ```Go
> 	// Any call to Size or Finish that supplies [rows] that's inclusive of the
> 	// index stored in minMaxRow may use the stored width. Calls with fewer
> 	// [rows] must recompute the min/max. In expected usage, only Finish will be
> ```

regarding "expected usage", is there a code simplification if we only supported this usage?
For instance we would only need to store widthWithLastTwoRows [2]uint32, and no recomputation.
___
*[`sstable/colblk/uints.go` line 80 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1IpNP_CYbUK8mtcwu9:-O1IpNP_CYbUK8mtcwuA:b-czh3xp) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L80)):*
> ```Go
> // assumed to be the provided default value. For the purpose of determining
> // whether a delta encoding is possible, the column is assumed to contain at
> // least 1 default value.
> ```

do we need a default that is not 0?
If there is no default `init(false, ...)`, are rows that are not set equal to NULL?

If we follow the rough guideline of our in-person discussion the default would be the zero-value for the type and there would be no distinction between absence (NULL) and the zero-value in the encoding.
___
*[`sstable/colblk/uints.go` line 118 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1Ir2Nk-kx_jQ_tmMI2:-O1Ir2Nk-kx_jQ_tmMI3:b3x3jgu) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L118)):*
> ```Go
> 		}
> 	} else {
> 		b.delta.minimum = T(0) - 1
> ```

we are subtracting 1 from uint(0), so won't it wrap around to the max value?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731)* status: 1 of 11 files reviewed, 11 unresolved discussions (waiting on @aadityasondhi, @petermattis, @RaduBerinde, and @sumeerbhola)

___
*[`internal/aligned/aligned.go` line 16 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0t_PgK3Zl1GusBPkZX:-O1JLki81eOabEq83DVB:b6wlf63) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/internal/aligned/aligned.go#L16)):*
Updated this to do the cast

> Is the idea here that if the block start is word aligned, there is nothing in the block layout that will have broken that alignment for individual columns (e.g. the Bitmap column)? If yes, is that documented somewhere?

Yeah, expanded the `ColumnWriter.Finish` comment with this and added it to the beginning of a doc comment describing the overarching block format

> And if `ByteSlice` is needed by the writer, the block cache is irrelevant, yes?

Yeah, that's true
___
*[`sstable/colblk/column.go` line 61 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1IaHcyD2jSjkEV22nZ:-O1JAMAuETW8zWME2Ay-:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/column.go#L61)):*

can you add an assertion that this is <= 8.
</blockquote></details>

Done.
___
*[`sstable/colblk/uints.go` line 118 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0ybrbK4LWE1Xh9_syQ:-O1JC6ddDqSbDyDgW6rP:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/01493604e56e69ce1ad93f8cf7c9179c8d760af8/sstable/colblk/uints.go#L118)):*

[nit] It would be good to mention that in a comment here, e.g.
`// We could reset all values as a precaution, but it has a visible cost in benchmarks.`
</blockquote></details>

Done
___
*[`sstable/colblk/uints.go` line 54 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1Io12h5t6BM1w7vmEZ:-O1JAouV9KiJBCAnmxt2:b4h0aoo) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L54)):*

what is `minMaxRow`?
</blockquote></details>

oops, fixed
___
*[`sstable/colblk/uints.go` line 80 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1IpNP_CYbUK8mtcwu9:-O1JK1wV32iC-c_s_PAs:bu8m2nh) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L80)):*
updated to remove support for configuring the default value. 

> If there is no default `init(false, ...)`, are rows that are not set equal to NULL?

it's undefined (eg, the caller is required to set all values)
___
*[`sstable/colblk/uints.go` line 118 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1Ir2Nk-kx_jQ_tmMI2:-O1JCB2c79Mqe8V9CFgh:bx20sab) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L118)):*

we are subtracting 1 from uint(0), so won't it wrap around to the max value?
</blockquote></details>

Yeah, to initialize the minimum to the max representable value. Added a comment
___
*[`sstable/colblk/uints.go` line 143 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1Hzk507Ej964sGQPm6:-O1JCZzY4HWVKbTBZMZu:b-th2zke) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L143)):*

[nit] we could `copy` into `newDataTyped` (and set defaultValue) directly, before creating newElems. Same thing, just a bit more readable
</blockquote></details>

good call, done
___
*[`sstable/colblk/uints.go` line 203 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1IZFaL9sW4uplY-Gw7:-O1JL0_AAZybWu5VPhKE:b-padq1j) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L203)):*

is there an alignment requirement on buf here?
</blockquote></details>

yeah, added this to the `ColumnWriter.Finish` comment
___
*[`sstable/colblk/unsafe_slice.go` line 81 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O0tzbLGELB4_yfvgJwS:-O1JCqcl6Jtj-sfdFIX2:b-dwozp5) ([raw file](https://github.com/cockroachdb/pebble/blob/6c4ba4396cf3f891e8f58a6e0bfdecdfbb17d571/sstable/colblk/unsafe_slice.go#L81)):*

Sorry, I mean treat it as a uint64 slice in all cases. On x86 even that's not necessary, I believe we could just do
```
T(*(*uint64)(unsafe.Pointer(uintptr(s.deltaPtr) + uintptr(i) << deltaWidthShift)))
```

Anyway, maybe just leave a TODO for now to experiment with other implementations.
</blockquote></details>

Added the TODO!

sumeerbhola (changes requested):
> Reviewed 4 of 7 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731)* status: 5 of 11 files reviewed, 14 unresolved discussions (waiting on @aadityasondhi, @jbowens, @petermattis, and @RaduBerinde)

___
*[`sstable/colblk/column.go` line 61 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1JrsT8AuW18Bczgns_:-O1JrsT8AuW18Bczgnsa:b-3ownw1) ([raw file](https://github.com/cockroachdb/pebble/blob/bd78301679ed6e4d1a7a87c90399be668123c696/sstable/colblk/column.go#L61)):*
> ```Go
> func (t DataType) uintWidth() uint32 {
> 	if t >= DataTypeUint8 && t <= DataTypeUint64 {
> 		v := t - DataTypeUint8
> ```

DataTypeUint64 - DataTypeUint8 is currently 3, so the if-condition for the panic seems weak.
___
*[`sstable/colblk/column.go` line 201 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1Jt4ce2FbPBWm3xEeO:-O1Jt4ce2FbPBWm3xEeP:bvfn3w1) ([raw file](https://github.com/cockroachdb/pebble/blob/bd78301679ed6e4d1a7a87c90399be668123c696/sstable/colblk/column.go#L201)):*
> ```Go
> 	Encoder
> 	// NumColumns returns the number of columns the ColumnWriter will encode.
> 	NumColumns() int
> ```

(copying a question over from https://github.com/cockroachdb/pebble/pull/3711#pullrequestreview-2157202836)
if this is a writer for a single column, why would it ever return a value other than 1?
___
*[`sstable/colblk/uints.go` line 79 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1JtydKA7FA6OJE4oU8:-O1JtydKA7FA6OJE4oU9:by7teru) ([raw file](https://github.com/cockroachdb/pebble/blob/bd78301679ed6e4d1a7a87c90399be668123c696/sstable/colblk/uints.go#L79)):*
> ```Go
> // set are assumed to be zero. For the purpose of determining whether a delta
> // encoding is possible, the column is assumed to contain at least 1 default
> // value.
> ```

The last sentence "... the column is assumed ..." is unclear to me. Why does it matter whether there is a value with the default in this column?
Oh, I see -- it seems an optimization to set an ok-ish value of min and max, and not a correctness requirements. If we don't make this assumption we need to know if there were some skipped row indices in `Set`. Is my understanding correct?

Is there a reason why we don't require monotonically increasing `row` values in `Set`. That would allow us to know if there is a gap and deal with defaults more easily. It would eliminate the need to have a `useDefault` field -- if there is a gap, we would fill it in with defaults.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731)* status: 5 of 11 files reviewed, 14 unresolved discussions (waiting on @aadityasondhi, @petermattis, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk/column.go` line 61 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1JrsT8AuW18Bczgns_:-O1N84XF0fNEwU-Ldd9v:bvks2q5) ([raw file](https://github.com/cockroachdb/pebble/blob/bd78301679ed6e4d1a7a87c90399be668123c696/sstable/colblk/column.go#L61)):*

DataTypeUint64 - DataTypeUint8 is currently 3, so the if-condition for the panic seems weak.
</blockquote></details>

what did you have in mind for the assertion?

https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1IaHcyD2jSjkEV22nZ
___
*[`sstable/colblk/column.go` line 201 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1Jt4ce2FbPBWm3xEeO:-O1N9QRa086uD5Fnif8H:b-7l5ui2) ([raw file](https://github.com/cockroachdb/pebble/blob/bd78301679ed6e4d1a7a87c90399be668123c696/sstable/colblk/column.go#L201)):*

(copying a question over from https://github.com/cockroachdb/pebble/pull/3711#pullrequestreview-2157202836)
if this is a writer for a single column, why would it ever return a value other than 1?
</blockquote></details>

It doesn't necessarily write a single column. Specifically the `KeyWriter` interface will implement ColumnWriter. `KeyWriter` is defined by the application and encodes user keys in a user-defined schema
___
*[`sstable/colblk/uints.go` line 55 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1IoBhD66gm1eQlenNq:-O1ND-Hp8SMxeiO8Bxwx:b-lszzgz) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L55)):*

regarding "expected usage", is there a code simplification if we only supported this usage?
For instance we would only need to store widthWithLastTwoRows [2]uint32, and no recomputation.
</blockquote></details>

Expanded the TODO below; I've been trying to avoid that subtlety in the API, but it may be net cleaner that way
___
*[`sstable/colblk/uints.go` line 79 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1JtydKA7FA6OJE4oU8:-O1NCMdt1fMjcAphxdzk:b-ykcfiy) ([raw file](https://github.com/cockroachdb/pebble/blob/bd78301679ed6e4d1a7a87c90399be668123c696/sstable/colblk/uints.go#L79)):*

The last sentence "... the column is assumed ..." is unclear to me. Why does it matter whether there is a value with the default in this column?
Oh, I see -- it seems an optimization to set an ok-ish value of min and max, and not a correctness requirements. If we don't make this assumption we need to know if there were some skipped row indices in `Set`. Is my understanding correct?

Is there a reason why we don't require monotonically increasing `row` values in `Set`. That would allow us to know if there is a gap and deal with defaults more easily. It would eliminate the need to have a `useDefault` field -- if there is a gap, we would fill it in with defaults.
</blockquote></details>

Yeah, when I was working on getting down the block-building overhead, avoiding the per-column, per-row store was impactful to the benchmarks

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731)* status: 4 of 11 files reviewed, 14 unresolved discussions (waiting on @aadityasondhi, @petermattis, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk/uints.go` line 55 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1IoBhD66gm1eQlenNq:-O1NRTv7AojSV3n9-iEx:b64s58v) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L55)):*

Expanded the TODO below; I've been trying to avoid that subtlety in the API, but it may be net cleaner that way
</blockquote></details>

Ah, now that we use this encoding for the offsets in RawBytes and the soon-to-be-added PrefixBytes, we have an instance of an expected usage that requires the more permissive contract. Adding a new row to PrefixBytes may result in +2 additional offsets (+1 for a new bundle prefix), so the PrefixBytes implementation of the more constrained contract would require a slightly more permissive contract from the UintBuilder.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731)* status: 4 of 11 files reviewed, 13 unresolved discussions (waiting on @aadityasondhi, @jbowens, @petermattis, and @sumeerbhola)

___
*[`sstable/colblk/uints.go` line 143 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1Hzk507Ej964sGQPm6:-O1NiCgr6SqdLmbD4Z3O:b-tw0wjx) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L143)):*

good call, done
</blockquote></details>

We still need to do
```
if b.useDefault {
  for i := b.array.n; i < len(newDataTyped); i++ {
    newDataTyped[i] = b.defaultValue
  }
```


By the way, do we envision usage where the default value is not 0?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731)* status: 4 of 11 files reviewed, 13 unresolved discussions (waiting on @aadityasondhi, @petermattis, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk/uints.go` line 143 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1Hzk507Ej964sGQPm6:-O1Nlfhj2dkxbpDn6266:b-yalo24) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L143)):*

We still need to do
```
if b.useDefault {
  for i := b.array.n; i < len(newDataTyped); i++ {
    newDataTyped[i] = b.defaultValue
  }
```


By the way, do we envision usage where the default value is not 0?
</blockquote></details>

I already made the change where the default value is only zero (so we don't need to update the grown size)

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731#-:-O1Nylf-EZw-Mlwr7xxl:bnfp4nl)* status: 4 of 11 files reviewed, 13 unresolved discussions (waiting on @aadityasondhi, @jbowens, @petermattis, and @sumeerbhola)

___
*[`sstable/colblk/uints.go` line 143 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1Hzk507Ej964sGQPm6:-O1Nyj4s8zgnBi3GdtQp:b-khgprv) ([raw file](https://github.com/cockroachdb/pebble/blob/51074522fa4d5605024c9f71596f16cba2e4cd5f/sstable/colblk/uints.go#L143)):*

I already made the change where the default value is only zero (so we don't need to update the grown size)
</blockquote></details>

Ah, missed that, great :thumbsup:

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 5 files at r2, 1 of 5 files at r4, 1 of 7 files at r6, 1 of 2 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731#-:-O1P0it-0T0YoxT0w1Mw:bg7igkv)* status: 9 of 11 files reviewed, 14 unresolved discussions (waiting on @aadityasondhi, @jbowens, @petermattis, and @RaduBerinde)

___
*[`sstable/colblk/column.go` line 61 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1JrsT8AuW18Bczgns_:-O1OxW8Z8Y9zycVsaWn0:bqr96n8) ([raw file](https://github.com/cockroachdb/pebble/blob/bd78301679ed6e4d1a7a87c90399be668123c696/sstable/colblk/column.go#L61)):*

what did you have in mind for the assertion?

https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1IaHcyD2jSjkEV22nZ
</blockquote></details>

didn't meant to drag this on -- it was a just a minor opinion.
```
rv := 1 << (t - DataTypeUint8)
if rv > 8 {
  panic("width greater than 8 bytes")
}
return rv
```
___
*[`sstable/colblk/uints.go` line 79 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1JtydKA7FA6OJE4oU8:-O1OzcEz174PSVDEX12J:brr2yve) ([raw file](https://github.com/cockroachdb/pebble/blob/bd78301679ed6e4d1a7a87c90399be668123c696/sstable/colblk/uints.go#L79)):*
> avoiding the per-column, per-row store was impactful to the benchmarks

Ack. Possibly the tight loop that initializes to the default value has better cache locality. AFAIK, golang compiler does not auto-vectorize.
___
*[`sstable/colblk/uints.go` line 208 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1P0HUL1VcvszwRJ3yL:-O1P0HUL1VcvszwRJ3yM:b-weft8v) ([raw file](https://github.com/cockroachdb/pebble/blob/237b17894ea5ec0890f535c8952764510cc4e354/sstable/colblk/uints.go#L208)):*
> ```Go
> }
> 
> // Size implements ColumnWriter, serializing the column into offset [offset] of
> ```

nit: Finish implements ...

jbowens (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3731#-:-O1SHaYrDlEA1c9fNUyZ:bz5mair)* status: 7 of 11 files reviewed, 13 unresolved discussions (waiting on @aadityasondhi, @petermattis, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk/column.go` line 61 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3731#-O1JrsT8AuW18Bczgns_:-O1SHUQ51rwMnfP0WBoa:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/bd78301679ed6e4d1a7a87c90399be668123c696/sstable/colblk/column.go#L61)):*

didn't meant to drag this on -- it was a just a minor opinion.
```
rv := 1 << (t - DataTypeUint8)
if rv > 8 {
  panic("width greater than 8 bytes")
}
return rv
```
</blockquote></details>

Done.

PR #3735, opened 2024-07-07
sstable: clean up reader/writer options

RaduBerinde:
The current code uses a `preApply` marker method to change when
`readerApply/writerApply` is called. This is unnecessarily subtle.
We change `ReaderOption` to have `readerApplyPre` and
`readerApplyPost`. We change `WriterOption` to always be called at the
beginning (we only have one implementation).


RaduBerinde:
Actually, I think we should just move these to `ReaderOptions` and `WriterOptions`. I don't see any advantage to having two mechanisms. I'll close for now.

Issue #3287, opened 2024-02-09
db: range keys are not supported in external files


RaduBerinde:
If we use `IngestExternalFiles` with a file that has range keys, we can get incorrect results. This is because some iterators (like `keyspan.LevelIter`) assume that range key boundaries are tight (and `HasRangeKeys` is accurate). These assumptions don't hold in general for external files.

Jira issue: PEBBLE-114

RaduBerinde:
This is done.

PR #3737, opened 2024-07-08
db: support range keys in external files

RaduBerinde:
#### keyspan: modernize the LevelIter test

This change improves the data-driven LevelIter test and simplfies the
code. Instead of specifying point keys, we allow extending either
point key or range key bounds directly.

#### db: support range keys in external files

We now support range keys in external files, and we support synthetic
prefix for range keys. We don't yet support synthetic suffix.

Informs #3287

### DIFF
--- a/testdata/ingest_external
+++ b/testdata/ingest_external
@@ -996,3 +996,72 @@ p1_a@1: (va, .)
 c@1: (vc, .)
 a@1: (va, .)
 .
+
+# Tests with range keys.
+
+reset
+----
+
+build-remote points-and-ranges.sst
+set a@1 va
+set b@1 vb
+set c@1 vc
+set d@1 vc
+range-key-set b c1 @2 vrange
+----
+
+ingest-external
+points-and-ranges.sst bounds=(a,e) has-range-keys
+----
+
+iter
+first
+next
+next
+next
+next
+next
+----
+a@1: (va, .)
+b: (., [b-"c1") @2=vrange UPDATED)
+b@1: (vb, [b-"c1") @2=vrange)
+c@1: (vc, [b-"c1") @2=vrange)
+d@1: (vc, . UPDATED)
+.
+
+ingest-external
+points-and-ranges.sst synthetic-prefix=x_ bounds=(x_a,x_d) has-range-keys
+----
+
+iter
+last
+prev
+prev
+prev
+prev
+prev
+----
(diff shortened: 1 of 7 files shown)


jbowens (approved):
:lgtm:

> Reviewed 3 of 3 files at r1, 1 of 1 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3737#-:-O1SxsYEEI0jdnThMMkm:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @itsbilal)

testdata/ingest_external
```diff
@@ -996,3 +996,54 @@ p1_a@1: (va, .)
...
+----
+a@1: (va, .)
+b: (., [b-"c1") @2=vrange UPDATED)
+b@1: (vb, [b-"c1") @2=vrange)
+c@1: (vc, [b-"c1") @2=vrange)
+d@1: (vc, . UPDATED)
+.
+
+ingest-external
+points-and-ranges.sst synthetic-prefix=x_ bounds=(x_a,x_d) has-range-keys
```
itsbilal, line 1033: Let's try adding a case where the last key is the rangedel's end key, and yet we're at a loose bound? So something like `[a, c2)` for the vsst bounds.

itsbilal (approved):
LGTM

RaduBerinde (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3737#-:-O1TH6C-DgcQ25TfdxFk:bz5mair)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`testdata/ingest_external` line 1033 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3737#gh-1672794052:-O1THoj3DmqPUY4HAqTX:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/74eab178f98ace19b5d994351a35e518df47fa0e/testdata/ingest_external#L1033)):*

Let's try adding a case where the last key is the rangedel's end key, and yet we're at a loose bound? So something like `[a, c2)` for the vsst bounds.
</blockquote></details>

Done.

RaduBerinde:
TFTRs!

PR #3738, opened 2024-07-08
sstable: factor out common iterator init code

RaduBerinde:
We've had multiple issues with divergences between initializing the data block read handles. This PR removes the code duplication.

#### sstable: add constructors for iterators

Add `newSingleLevelIterator` and `newTwoLevelIterator` constructors.

#### sstable: factor out common iterator init code

This commits factors out most of the `singleLevelIterator`
initialization code, so that `twoLevelIterator` can use it too.

### DIFF
--- a/sstable/reader.go
+++ b/sstable/reader.go
@@ -221,23 +221,22 @@ func (r *Reader) newIterWithBlockPropertyFiltersAndContext(
 	// NB: pebble.tableCache wraps the returned iterator with one which performs
 	// reference counting on the Reader, preventing the Reader from being closed
 	// until the final iterator closes.
+	var res Iterator
+	var err error
 	if r.Properties.IndexType == twoLevelIndex {
-		i := twoLevelIterPool.Get().(*twoLevelIterator)
-		err := i.init(ctx, r, vState, transforms, lower, upper, filterer, useFilterBlock,
+		res, err = newTwoLevelIterator(ctx, r, vState, transforms, lower, upper, filterer, useFilterBlock,
 			stats, categoryAndQoS, statsCollector, rp, nil /* bufferPool */)
-		if err != nil {
-			return nil, err
-		}
-		return i, nil
+	} else {
+		res, err = newSingleLevelIterator(
+			ctx, r, vState, transforms, lower, upper, filterer, useFilterBlock,
+			stats, categoryAndQoS, statsCollector, rp, nil /* bufferPool */)
 	}
-
-	i := singleLevelIterPool.Get().(*singleLevelIterator)
-	err := i.init(ctx, r, vState, transforms, lower, upper, filterer, useFilterBlock,
-		stats, categoryAndQoS, statsCollector, rp, nil /* bufferPool */)
 	if err != nil {
+		// Note: we don't want to return res here - it will be a nil
+		// single/twoLevelIterator, not a nil Iterator.
 		return nil, err
 	}
-	return i, nil
+	return res, nil
 }
 
 // NewIter returns an iterator for the contents of the table. If an error
@@ -275,8 +274,7 @@ func (r *Reader) newCompactionIter(
 		transforms.HideObsoletePoints = true
 	}
 	if r.Properties.IndexType == twoLevelIndex {
-		i := twoLevelIterPool.Get().(*twoLevelIterator)
-		err := i.init(
+		i, err := newTwoLevelIterator(
 			context.Background(),
 			r, vState, transforms, nil /* lower */, nil /* upper */, nil,
 			false /* useFilter */, nil /* stats */, categoryAndQoS, statsCollector, rp, bufferPool,
@@ -287,8 +285,7 @@ func (r *Reader) newCompactionIter(
 		i.setupForCompaction()
 		return &twoLevelCompactionIterator{twoLevelIterator: i}, nil
 	}
-	i := singleLevelIterPool.Get().(*singleLevelIterator)
-	err := i.init(
+	i, err := newSingleLevelIterator(
 		context.Background(), r, vState, transforms, nil /* lower */, nil, /* upper */
 		nil, false /* useFilter */, nil /* stats */, categoryAndQoS, statsCollector, rp, bufferPool,
 	)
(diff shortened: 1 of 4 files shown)


sumeerbhola (approved):
:lgtm: 

> Reviewed 3 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3738#-:-O1P4jfZ35w5A8XxRcdu:bg7igkv)* status: all files reviewed, 3 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 204 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3738#-O1P2NfRCc-zbraMEt7g:-O1P2NfRCc-zbraMEt7h:b-odixq0) ([raw file](https://github.com/cockroachdb/pebble/blob/04dbbd033fe1540a3548d0df7a498f599938f551/sstable/reader_iter_single_lvl.go#L204)):*
> ```Go
> 		ctx, r, v, transforms, lower, upper, filterer, useFilter, stats, categoryAndQoS, statsCollector, rp, bufferPool,
> 	); err != nil {
> 		_ = i.Close()
> ```

There's a small change here in that on master we were not calling Close() if `init()` returned an error.
Do we have enough testing of error paths (or a careful read of the Close code) to have some certainty that calling `Close()` is safe, i.e., there isn't some partially initialized state that will blow up in Close (e.g. unexpected nil pointer or something).

---
edit: there are changes in the second commit that slightly alter the details of the question, but it still stands.
___
*[`sstable/reader_iter_two_lvl.go` line 166 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3738#-O1P3FJ3CsiIteNTPqqP:-O1P3FJ3CsiIteNTPqqQ:b-7ropln) ([raw file](https://github.com/cockroachdb/pebble/blob/04dbbd033fe1540a3548d0df7a498f599938f551/sstable/reader_iter_two_lvl.go#L166)):*
> ```Go
> 	topLevelIndexH, err := r.readIndex(ctx, i.indexFilterRH, stats, &i.iterStats)
> 	if err != nil {
> 		_ = i.Close()
> ```

ditto
___
*[`sstable/reader_iter_two_lvl.go` line 188 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3738#-O1P3G000X4O55iFCybn:-O1P3G000X4O55iFCybo:b-7ropln) ([raw file](https://github.com/cockroachdb/pebble/blob/04dbbd033fe1540a3548d0df7a498f599938f551/sstable/reader_iter_two_lvl.go#L188)):*
> ```Go
> 	if err != nil {
> 		// blockIter.Close releases topLevelIndexH and always returns a nil error
> 		_ = i.Close()
> ```

ditto

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3738)* status: all files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 204 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3738#-O1P2NfRCc-zbraMEt7g:-O1Pdd3PCmsgQ4WhA8vY:bptb4ps) ([raw file](https://github.com/cockroachdb/pebble/blob/04dbbd033fe1540a3548d0df7a498f599938f551/sstable/reader_iter_single_lvl.go#L204)):*

There's a small change here in that on master we were not calling Close() if `init()` returned an error.
Do we have enough testing of error paths (or a careful read of the Close code) to have some certainty that calling `Close()` is safe, i.e., there isn't some partially initialized state that will blow up in Close (e.g. unexpected nil pointer or something).

---
edit: there are changes in the second commit that slightly alter the details of the question, but it still stands.
</blockquote></details>

Good point, I added a test.

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 3 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3738#-:-O1RdAky1-4-Dmu8Tbx3:bg7igkv)* status: all files reviewed, 4 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 204 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3738#-O1P2NfRCc-zbraMEt7g:-O1Rd9oy0FtOWa4oy7fG:b-zcjvte) ([raw file](https://github.com/cockroachdb/pebble/blob/04dbbd033fe1540a3548d0df7a498f599938f551/sstable/reader_iter_single_lvl.go#L204)):*

Good point, I added a test.
</blockquote></details>

Thanks
___
*[`sstable/reader_iter_test.go` line 54 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3738#-O1Rd0Mi26H66KIMsSn5:-O1Rd0Mi26H66KIMsSn6:blhx0f3) ([raw file](https://github.com/cockroachdb/pebble/blob/12184567c14e65863d3e78d5d6685dd43b933cff/sstable/reader_iter_test.go#L54)):*
> ```Go
> 
> 	var stats base.InternalIteratorStats
> 	for k := 0; k < 20; k++ {
> ```

wondering why this iteration instead of simply creating each kind of iterator once?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3738)* status: all files reviewed, 4 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/reader_iter_test.go` line 54 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3738#-O1Rd0Mi26H66KIMsSn5:-O1S9DWpCKNKr-ne3r56:bvkdwtz) ([raw file](https://github.com/cockroachdb/pebble/blob/12184567c14e65863d3e78d5d6685dd43b933cff/sstable/reader_iter_test.go#L54)):*

wondering why this iteration instead of simply creating each kind of iterator once?
</blockquote></details>

Just so we reuse the same iterator from the pool (in case ae leave it in a bad state in this path)

PR #3739, opened 2024-07-09
compact: add simple threshold-based tombstone compaction heuristic

anish-shanbhag:

PR #3744, opened 2024-07-11
db: Add BenchmarkQueueWorkload

anish-shanbhag:
BenchmarkQueueWorkload benchmarks a workload consisting of multiple queues that are all being processed at the same time. Processing a queue entails either appending to the end of the queue (a `Set` operation) or deleting from the start of the queue (a `Delete` operation). The goal is to detect cases where we see a large buildup of point tombstones at the beginning of each queue, which leads to the slowdown of `SeekGE(<start of queue>)`. To that end, the test progresses through a series of steps that each 1) process the queues a certain number of times and then 2) benchmark both the queue processing throughput and `SeekGE` performance.

See https://github.com/facebook/rocksdb/wiki/Implement-Queue-Service-Using-RocksDB for more context.

### DIFF
--- a/iterator_test.go
+++ b/iterator_test.go
@@ -2737,3 +2737,163 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
 		iter.SeekPrefixGE(seekKey)
 	}
 }
+
+func runBenchmarkQueueWorkload(b *testing.B, deleteRatio float32, initOps int, valueSize int) {
+	const queueCount = 8
+	// These should be large enough to assign a unique key to each item in the
+	// queue.
+	const maxQueueIDLen = 1
+	const maxItemLen = 7
+	const maxKeyLen = maxQueueIDLen + 1 + maxItemLen
+	queueIDKeyspace := testkeys.Alpha(maxQueueIDLen)
+	itemKeyspace := testkeys.Alpha(maxItemLen)
+	key := make([]byte, maxKeyLen)
+	val := make([]byte, valueSize)
+	rng := rand.New(rand.NewSource(uint64(time.Now().UnixNano())))
+
+	getKey := func(q int, i int) []byte {
+		n := testkeys.WriteKey(key, queueIDKeyspace, int64(q))
+		key[n] = '/'
+		prefixLen := n + 1
+		n = testkeys.WriteKey(key[prefixLen:], itemKeyspace, int64(i))
+		return key[:prefixLen+n]
+	}
+
+	type Queue struct {
+		start int
+		end   int // exclusive
+	}
+	var queues = make([]*Queue, queueCount)
+	for i := 0; i < queueCount; i++ {
+		queues[i] = &Queue{}
+	}
+
+	o := (&Options{
+		DisableWAL:         true,
+		FS:                 vfs.NewMem(),
+		Comparer:           testkeys.Comparer,
+		FormatMajorVersion: FormatNewest,
+	}).EnsureDefaults()
+
+	d, err := Open("", o)
+	require.NoError(b, err)
+
+	processQueueOnce := func(batch *Batch) {
+		for {
+			// Randomly pick a queue to process.
+			q := rng.Intn(queueCount)
+			queue := queues[q]
+
+			isDelete := rng.Float32() < deleteRatio
+
+			if isDelete {
+				// Only process the queue if it's not empty. Otherwise, retry
+				// with a different queue.
+				if queue.start != queue.end {
+					require.NoError(b, batch.Delete(getKey(q, queue.start), nil))
+					queue.start = (queue.start + 1) % int(itemKeyspace.Count())
+					break
+				}
+			} else {
+				// Append to the queue.
+				require.NoError(b, batch.Set(getKey(q, queue.end), val, nil))
+				queue.end = (queue.end + 1) % int(itemKeyspace.Count())
+				break
+			}
+		}
+	}
+
+	// First, process queues initialOps times.
+	batch := d.NewBatch()
+	for i := 0; i < initOps; i++ {
+		processQueueOnce(batch)
+		// Use a large batch size to speed up initialization.
+		if batch.Len() >= 10<<24 /* 167 MiB */ {
+			require.NoError(b, batch.Commit(NoSync))
+			batch = d.NewBatch()
+		}
+	}
+	require.NoError(b, batch.Commit(NoSync))
+	// Manually flush in case the last batch was small.
+	_, err = d.AsyncFlush()
+	require.NoError(b, err)
+
+	waitForCompactions := func() {
+		d.mu.Lock()
+		// NB: Wait for table stats because some compaction types rely
+		// on table stats to be collected.
+		d.waitTableStats()
+		for d.mu.compact.compactingCount > 0 {
+			d.mu.compact.cond.Wait()
+			d.waitTableStats()
+		}
+		d.mu.Unlock()
+	}
+
+	waitForCompactions()
+
+	// Log the number of tombstones and live keys in each level after
+	// background compactions are complete.
+	b.Log("LSM after compactions:")
+	firstIter, _ := d.NewIter(nil)
+	firstIter.First()
+	lastIter, _ := d.NewIter(nil)
+	lastIter.Last()
+	stats, _ := d.ScanStatistics(context.Background(), firstIter.Key(), lastIter.Key(), ScanStatisticsOptions{})
+	require.NoError(b, firstIter.Close())
+	require.NoError(b, lastIter.Close())
+	metrics := d.Metrics()
+	for i := 0; i < numLevels; i++ {
+		numTombstones := stats.Levels[i].KindsCount[base.InternalKeyKindDelete]
+		numSets := stats.Levels[i].KindsCount[base.InternalKeyKindSet]
+		numTables := metrics.Levels[i].NumFiles
+		if numSets > 0 {
+			b.Logf("L%d: %d tombstones, %d sets, %d sstables\n", i, numTombstones, numSets, numTables)
+		}
+	}
+
+	// Seek to the start of each queue.
+	b.Run("seek", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			for q := 0; q < queueCount; q++ {
+				iter, _ := d.NewIter(nil)
+				iter.SeekGE(getKey(q, 0))
+				require.NoError(b, iter.Close())
+			}
+		}
+	})
+
+	require.NoError(b, d.Close())
+}
+
+// BenchmarkQueueWorkload benchmarks a workload consisting of multiple queues
+// that are all being processed at the same time. Processing a queue entails
+// either appending to the end of the queue (a Set operation) or deleting from
+// the start of the queue (a Delete operation). The goal is to detect cases
+// where we see a large buildup of point tombstones at the beginning of each
+// queue, which leads to the slowdown of SeekGE(<start of queue>). To that end,
+// the test subbenchmarks a series of configurations that each 1) process the
+// queues a certain number of times and then 2) benchmark both the queue
+// processing throughput and SeekGE performance. See
+// https://github.com/facebook/rocksdb/wiki/Implement-Queue-Service-Using-RocksDB
+// for more information.
+func BenchmarkQueueWorkload(b *testing.B) {
+	// The portion of processing ops that will be deletes for each subbenchmark.
+	var deleteRatios = []float32{0.1, 0.3, 0.5}
+	// The number of times queues will be processed before running each
+	// subbenchmark.
+	var initOps = []int{400_000, 800_000, 1_200_000, 2_000_000, 3_500_000, 5_000_000, 7_500_000, 10_000_000, 50_000_000}
+	// We vary the value size to identify how compaction behaves when the
+	// relative sizes of tombstones and the keys they delete are different.
+	var valueSizes = []int{128, 2048}
+
+	for _, deleteRatio := range deleteRatios {
+		for _, valueSize := range valueSizes {
+			for _, numInitOps := range initOps {
+				b.Run(fmt.Sprintf("initial_ops=%d/deleteRatio=%.2f/valueSize=%d", numInitOps, deleteRatio, valueSize), func(b *testing.B) {
+					runBenchmarkQueueWorkload(b, deleteRatio, numInitOps, valueSize)
+				})
+			}
+		}
+	}
+}


anish-shanbhag:
Initial results show that the simple tombstone density threshold heuristic from https://github.com/cockroachdb/pebble/pull/3739#issuecomment-2218429942 does not properly deal with this case. See [here](https://gist.github.com/anish-shanbhag/1aeb6b751d283871ab47c66dd0399002) for benchmark results from the current master, and [here](https://gist.github.com/anish-shanbhag/59c88cece3d74a78a097a285bbe958c3) for the results with the heuristic enabled. In both cases, we see tombstone buildup in L4/L5 and `SeekGE` performance gradually going from `200,000 ns/op` -> `700,000+ ns/op`.

iterator_test.go
```diff
@@ -2737,3 +2738,179 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+		d.waitTableStats()
+		for d.mu.compact.compactingCount > 0 {
+			d.mu.compact.cond.Wait()
+			d.waitTableStats()
+		}
+		d.mu.Unlock()
+	}
+
+	for step := 0; step < len(iterationsPerStep); step++ {
+		b.Run(fmt.Sprintf("step=%d", step), func(b *testing.B) {
```
itsbilal, line 2848: I'm not sure if the step-based approach works well within the context of how these benchmarks are named or report their results. I wonder if a better approach is to just have sub-benchmarks that modify the intiial number of operations that you run before you call the time-bound `b.N` operations. So you could have sub-benchmarks that do 1000, 10_000, 100_000, 1_000_000, etc pre-timer iterations on the queues, then do b.N iterations.

iterator_test.go
```diff
@@ -2737,3 +2738,179 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+// appending to the end of the queue (a Set operation) or deleting from the start of
+// the queue (a Delete operation). The goal is to detect cases where we see a large
+// buildup of point tombstones at the beginning of each queue, which leads to the
+// slowdown of SeekGE(<start of queue>). To that end, the test progresses through
+// a series of steps that each 1) process the queues a certain number of times
+// and then 2) benchmark both the queue processing throughput and SeekGE performance.
+// See https://github.com/facebook/rocksdb/wiki/Implement-Queue-Service-Using-RocksDB
+// for more information.
+func BenchmarkQueueWorkload(b *testing.B) {
+	const queueCount = 25
```
itsbilal, line 2753: A small number of queues than 25 would probably get us bigger clumps of tombstones. You could also vary this in sub-benchmarks, eg.

```
queueCounts := int[]{1,2,4,8,16}
for _, queueCount := range queueCounts {
  b.Run(fmt.Sprintf("queues=%d", queueCount), func(...){
    ....
  })
}
```

iterator_test.go
```diff
@@ -2737,3 +2738,179 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+						return err
+					}
+					return nil
+				})
+			}
+			require.NoError(b, grp.Wait())
+
+			waitForCompactions()
+
+			b.Run("queue throughput", func(b *testing.B) {
```
itsbilal, line 2867: Conventionally we try to not have spaces in benchmark names, so `throughput` or `queue_throughput` works here, as well as `seek` below.

itsbilal (commented):
Mostly looking great! Some relatively minor suggestions mostly around better targeting what's being benchmarked.

iterator_test.go
```diff
@@ -2737,3 +2738,179 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+// the queue (a Delete operation). The goal is to detect cases where we see a large
+// buildup of point tombstones at the beginning of each queue, which leads to the
+// slowdown of SeekGE(<start of queue>). To that end, the test progresses through
+// a series of steps that each 1) process the queues a certain number of times
+// and then 2) benchmark both the queue processing throughput and SeekGE performance.
+// See https://github.com/facebook/rocksdb/wiki/Implement-Queue-Service-Using-RocksDB
+// for more information.
+func BenchmarkQueueWorkload(b *testing.B) {
+	const queueCount = 25
+	const minValSize = 512
```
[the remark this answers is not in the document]
anish-shanbhag: I tried this initially, but increasingly found that this runs into the same issue of lack of comparability/interpretability that comes with randomizing the delete ratios. I opted to go without the randomization entirely and use a slice of value sizes. It's set to just `{128, 2048}` since that seems to be enough to show the difference that value size creates (i.e. tombstone buildup is worse with larger value sizes), but we can always run this with other value sizes in the future too.

iterator_test.go
```diff
@@ -2737,3 +2738,179 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+				queue.end++
+				require.NoError(b, err)
+				return
+			}
+		}
+	}
+
+	waitForCompactions := func() {
+		d.mu.Lock()
+		d.waitTableStats()
```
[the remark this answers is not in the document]
anish-shanbhag: I included this since the compaction heuristic I was testing relied on `stats.NumDeletions` to be populated before the compaction was scheduled. I added a comment to clarify this in the newest change, but I'll definitely keep this in mind in case the final heuristic implementation doesn't use table stats, which would let us delete this.

iterator_test.go
```diff
@@ -2737,3 +2738,179 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+		return key[:prefixLen+n]
+	}
+
+	rng := rand.New(rand.NewSource(0))
+	var queues []*Queue
+	for i := 0; i < queueCount; i++ {
+		q := &Queue{
+			// Each queue is assigned a random deleteRatio to vary the relative
+			// size of tombstone clusters that result from queue processing.
+			deleteRatio: rng.Float32() * maxDeleteRatio,
```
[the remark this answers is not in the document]
anish-shanbhag: Great idea, I agree that the randomly generated delete ratios lacked interpretability. The newest change uses a `RunOptions` struct that goes through a series of preset configurations, including the delete ratios `{0.2, 0.4, 0.6, 0.8}`.

iterator_test.go
```diff
@@ -2737,3 +2738,179 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+				// with a different queue.
+				if queue.start < queue.end {
+					err := batch.Delete(getKey(q, queue.start), nil)
+					require.NoError(b, err)
+					queue.start++
+					return
+				}
+			} else {
+				// Append to the queue.
+				err := batch.Set(getKey(q, queue.end), val, nil)
```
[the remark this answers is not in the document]
anish-shanbhag: fixed

iterator_test.go
```diff
@@ -2737,3 +2738,179 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+			q := rng.Intn(queueCount)
+			queue := queues[q]
+
+			isDelete := rng.Float32() < queue.deleteRatio
+
+			if isDelete {
+				// Only process the queue if it's not empty. Otherwise, retry
+				// with a different queue.
+				if queue.start < queue.end {
+					err := batch.Delete(getKey(q, queue.start), nil)
```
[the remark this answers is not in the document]
anish-shanbhag: fixed, thanks for the catch

iterator_test.go
anish-shanbhag: Agreed, I think the step-based approach was a bit unwieldy. I was originally using steps because performing 1 milliion+ ops before the timer produced an oddly shaped LSM with only a few tables, but after some digging I found that it was because I was using an uncapped batch size. In the newest change, I've gone with your approach where the `RunOptions` struct has a field for the `initialOps` we perform before each sub-benchmark.

iterator_test.go
anish-shanbhag: Lowered this to 8, which I think is working well

anish-shanbhag:
TFTR! I've updated with some changes, namely removing delete ratio/value size randomization and doing away with the step-based structure of the benchmark.

iterator_test.go
```diff
@@ -2737,3 +2739,200 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+				}
+				runOpts = append(runOpts, opts)
+
+				d, err := Open("", o)
+				require.NoError(b, err)
+				dbs = append(dbs, d)
+
+				// First we process queues initialOps times (parallelized to
+				// speed things up).
+				grp.Go(func() error {
```
itsbilal, line 2858: You're speeding up the initialization for all the different sub-benchmarks, which you wouldn't get if you moved it all into a helper function called once, but it's helpful to have each benchmark get as much in the way of system resources (eg. memory) as it could. It's also possible that you aren't interested in running some of the sub-benchmarks (eg. by passing a regex filter to `go test -bench`, which applies to the benchmark names) but want to run or re-run others. Currently you'll initialize them all even if you only use some of these in practice further down. So while it's a bit of a tradeoff to serialize initialization, it's a worthwhile one IMO.

itsbilal (commented):
Mostly looking great! Just one more suggestion

iterator_test.go
anish-shanbhag: For future testing, I'll probably focus on the `deleteRatio=0.1, valueSize=2048` config because it seems to be a clear example of tombstone buildup.

iterator_test.go
```diff
@@ -2737,3 +2737,176 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+			if numSets > 0 {
+				b.Logf("L%d: %d tombstones, %d sets, %d sstables\n", i, numTombstones, numSets, numTables)
+			}
+		}
+
+		b.Run("queue_throughput", func(b *testing.B) {
+			batch := d.NewBatch()
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				batch = processQueueOnce(batch, 10<<10 /* 10 KiB */)
```
itsbilal, line 2860: The last batch isn't getting committed, right? You could consider moving the batch commit outside of `processQueueOnce` just to make it more readable.

iterator_test.go
```diff
@@ -2737,3 +2737,176 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+				// Append to the queue.
+				require.NoError(b, batch.Set(getKey(q, queue.end), val, nil))
+				queue.end++
+				break
+			}
+		}
+
+		if batch.Len() >= batchSize {
+			require.NoError(b, batch.Commit(NoSync))
+			d.AsyncFlush()
```
itsbilal, line 2807: One flush per batch is probably unnecessary unless it's a huge batch. We can probably skip this part.

iterator_test.go
```diff
@@ -2737,3 +2737,176 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+		for i := 0; i < numLevels; i++ {
+			numTombstones := stats.Levels[i].KindsCount[base.InternalKeyKindDelete]
+			numSets := stats.Levels[i].KindsCount[base.InternalKeyKindSet]
+			numTables := metrics.Levels[i].NumFiles
+			if numSets > 0 {
+				b.Logf("L%d: %d tombstones, %d sets, %d sstables\n", i, numTombstones, numSets, numTables)
+			}
+		}
+
+		b.Run("queue_throughput", func(b *testing.B) {
```
itsbilal, line 2856: Since this part is just batch applications, which are basically a memtable (+ WAL write except wals are disabled), I don't see how it will be affected by any compaction heuristics or really any parameters. We can probably move this outside of b.Run, and also remove `seek` and just time the part currently within `seek` in the main benchmark.

The other reason not to time this part is that if we get unlucky with the randomization in `processQueueOnce`, and the deleteRatio is high, we could end up looping around the queues a bunch of times before we find one that has an object to delete.

iterator_test.go
```diff
@@ -2737,3 +2737,176 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+
+	// First, process queues initialOps times.
+	batch := d.NewBatch()
+	for i := 0; i < initOps; i++ {
+		// Use a large batch size to speed up initialization.
+		batch = processQueueOnce(batch, 10<<24 /* 167 MiB */)
+	}
+	require.NoError(b, batch.Commit(NoSync))
+
+	b.Run(fmt.Sprintf("initial_ops=%d,deleteRatio=%.2f,valueSize=%d", initOps, deleteRatio, valueSize), func(b *testing.B) {
```
itsbilal, line 2821: This should be in the outer function, ie. `BenchmarkQueueWorkload`:

```
for _, deleteRatio := range deleteRatios {
		for _, valueSize := range valueSizes {
			for _, numInitOps := range initOps {
			  b.Run(fmt.Sprintf("initial_ops=%d/deleteRatio=%.2f/valueSize=%d", initOps, deleteRatio, valueSize), func(b *testing.B) {
			  runBenchmarkQueueWorkload(...)
			 })
    }
   }
}
```

It might sound like a subtle difference but currently we'd be doing the initialization for all the combinations of the benchmark even if we're only filtering and running one of them.

iterator_test.go
```diff
@@ -2737,3 +2737,176 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+func runBenchmarkQueueWorkload(b *testing.B, deleteRatio float32, initOps int, valueSize int) {
+	const queueCount = 8
+	// These should be large enough to assign a unique key to each item in the
+	// queue.
+	const maxQueueIDLen = 1
+	const maxItemLen = 7
+	const maxKeyLen = maxQueueIDLen + 1 + maxItemLen
+	queueIDKeyspace := testkeys.Alpha(maxQueueIDLen)
+	itemKeyspace := testkeys.Alpha(maxItemLen)
+	key := make([]byte, maxKeyLen+1+len(fmt.Sprint(b.N)))
```
itsbilal, line 2750: what's the purpose of the `+1+len(fmt.Sprint(b.N))` part? the iteration number is already going into the `itemKeyspace` part, right?

iterator_test.go
```diff
@@ -2737,3 +2737,176 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+			queue := queues[q]
+
+			isDelete := rng.Float32() < deleteRatio
+
+			if isDelete {
+				// Only process the queue if it's not empty. Otherwise, retry
+				// with a different queue.
+				if queue.start < queue.end {
+					require.NoError(b, batch.Delete(getKey(q, queue.start), nil))
+					queue.start++
```
itsbilal, line 2794: We probably want some logic here where if `queue.start `exceeds `itemKeyspace.Count()`, we have it go back to 0. We'd also need to update the `queue.start < queue.end` conditional above to account for this overflow.

itsbilal (commented):
This is taking good shape; one more round of comments. Thanks!

iterator_test.go
anish-shanbhag: Yeah, the last batch wasn't being used - I've updated to move the commit logic outside

iterator_test.go
anish-shanbhag: My goal with this sub-benchmark was to surfaces cases where (a) queue throughput slows down as tombstones build up, and (b) throughput slows down due to background compactions that are scheduled as a result of any heuristic that I add. For (a), currently it looks like there is a clear throughput slowdown as `initialOps` increases for small value sizes, but for large values sizes the numbers are pretty random, so I agree that the usefulness is questionable.

For (b), I'm curious about your thoughts on whether this is a good way to test for performance hits due to excessive background compactions. One other idea I had was to time `waitForCompactions`, but this doesn't fit into the benchmarking model of repeating `b.N` times.

I've removed this for now since seek time is definitely the main number we care about, but open to thoughts on alternative ways to test for (b).

iterator_test.go
anish-shanbhag: Good catch, I've fixed this

iterator_test.go
anish-shanbhag: This was left behind from an earlier way I was writing keys - I've changed to just be `maxItemLen`

iterator_test.go
anish-shanbhag: I think this logic worked better in the section where we append to the queue, so I've added a check there to make sure `queue.end` never overflows.

iterator_test.go
anish-shanbhag: Update: I added back the `b.Run("seek", ...` subbenchmark so that the DB is only initialized once per configuration. Without it, the whole `runQueueBenchmarkWorkload` helper is called multiple times as the benchmark increases `b.N`. We can still test with multiple initializations to account for randomness using the `-count` flag.

iterator_test.go
anish-shanbhag: Update: I added a single `AsyncFlush` at the end of initialization so that we make sure the last batch is flushed even if it was small.

iterator_test.go
itsbilal: LSMs are inherently going to get as many things out of the write path as possible, and compactions are a background / nondeterministic process so benchmarking them directly is more challenging. I don't see any cases where queue throughput (as long as that only includes writes) would be slowed down by any buildup, unless you're hitting write stall thresholds which will be pretty hard.

So I think it still makes sense to not have a `queue_throughput` sub-benchmark, but one just for seeks is understandable. This sounds good, thanks!

iterator_test.go
```diff
@@ -2737,3 +2737,168 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+	// These should be large enough to assign a unique key to each item in the
+	// queue.
+	const maxQueueIDLen = 1
+	const maxItemLen = 7
+	const maxKeyLen = maxQueueIDLen + 1 + maxItemLen
+	queueIDKeyspace := testkeys.Alpha(maxQueueIDLen)
+	itemKeyspace := testkeys.Alpha(maxItemLen)
+	key := make([]byte, maxKeyLen)
+	val := make([]byte, valueSize)
+	rng := rand.New(rand.NewSource(0))
```
itsbilal, line 2752: Let's use `time.Now().UnixNano()` as the seed? Otherwise the rng will always use the same seed.

iterator_test.go
```diff
@@ -2737,3 +2737,168 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+				// with a different queue.
+				if queue.start < queue.end {
+					require.NoError(b, batch.Delete(getKey(q, queue.start), nil))
+					queue.start++
+					break
+				}
+			} else {
+				// Prevent queue overflow.
+				if queue.end+1 > int(itemKeyspace.Count()) {
+					queue.start = 0
```
itsbilal, line 2800: You probably want it to wrap around, right? So end would become 0 here, but you'd do a similar check for start and make it go to 0 too. Alternaitvely you could just do `(queue.end + 1) % itemKeyspace.Count()` anywhere you increment it. Your `start < end` logic can then become `start != end`.

iterator_test.go
```diff
@@ -2737,3 +2737,168 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+	// subbenchmark.
+	var initOps = []int{400_000, 800_000, 1_200_000, 2_000_000, 3_500_000, 5_000_000, 7_500_000, 10_000_000, 50_000_000}
+	// We vary the value size to identify how compaction behaves when the
+	// relative sizes of tombstones and the keys they delete are different.
+	var valueSizes = []int{128, 2048}
+
+	for _, deleteRatio := range deleteRatios {
+		for _, valueSize := range valueSizes {
+			for _, numInitOps := range initOps {
+				b.Run(fmt.Sprintf("initial_ops=%d,deleteRatio=%.2f,valueSize=%d", numInitOps, deleteRatio, valueSize), func(b *testing.B) {
```
itsbilal, line 2898: nit: let's use / as a separator instead of commas, in line with conventions elsewhere.

itsbilal (approved):
LGTM, feel free to merge this after addressing the remaining minor comments. Thanks a ton!

iterator_test.go
anish-shanbhag: I opted to use the modulo logic since it's cleaner - thanks for the suggestion!

Issue #3728, opened 2024-07-02
db: improve context propagation for tracing of slow block loads


sumeerbhola:
We sometimes see traces in CockroachDB which suggest slowness in Pebble, but the `InternalIteratorStats` printed don't indicate block cache misses. This is possibly because those stats are incomplete (don't include `Reader` creation or loading of various block types). We have a `LoggerAndTracer` passed to Pebble, and tracing of slow block loads (5ms is the slow threshold in `Reader.readBlock`), since https://github.com/cockroachdb/pebble/issues/2055. But there are some significant gaps:
- Many `Reader.readBlock` calls are using a background context, e.g. `Reader.readRangeDel`, `readRangeKey`, `readMetaindex`. These should use a real context.
- `tableCacheValue.load` which creates a new `Reader` should accept a context. We have one in the callpath that originates in `tableCacheShard.newIters`.

RaduBerinde:
Removing the P-2 label because the important paths have been addressed. There is only one path around value blocks that is more tricky.

PR #3745, opened 2024-07-11
sstable: trace/expensive log when sstable footer read is slow

sumeerbhola:
Relates to #3728

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -1437,7 +1437,7 @@ func TestTracing(t *testing.T) {
 	_, closer, err := d.Get([]byte("hello"))
 	require.NoError(t, err)
 	closer.Close()
-	readerInitTraceString := "reading 37 bytes took 5ms\nreading 419 bytes took 5ms\n"
+	readerInitTraceString := "reading 53 bytes took 5ms\nreading 37 bytes took 5ms\nreading 419 bytes took 5ms\n"
 	iterTraceString := "reading 27 bytes took 5ms\nreading 29 bytes took 5ms\n"
 	require.Equal(t, readerInitTraceString+iterTraceString, tracer.buf.String())
 
--- a/internal/base/logger.go
+++ b/internal/base/logger.go
@@ -94,11 +94,13 @@ func (b *InMemLogger) Fatalf(format string, args ...interface{}) {
 type LoggerAndTracer interface {
 	Logger
 	// Eventf formats and emits a tracing log, if tracing is enabled in the
-	// current context.
+	// current context. It can also emit to a regular log, if expensive
+	// logging is enabled.
 	Eventf(ctx context.Context, format string, args ...interface{})
-	// IsTracingEnabled returns true if tracing is enabled. It can be used as an
-	// optimization to avoid calling Eventf (which will be a noop when tracing
-	// is not enabled) to avoid the overhead of boxing the args.
+	// IsTracingEnabled returns true if tracing is enabled for this context,
+	// or expensive logging is enabled. It can be used as an optimization to
+	// avoid calling Eventf (which will be a noop when tracing or expensive
+	// logging is not enabled) to avoid the overhead of boxing the args.
 	IsTracingEnabled(ctx context.Context) bool
 }
 
--- a/sstable/reader.go
+++ b/sstable/reader.go
@@ -449,20 +449,14 @@ func (r *Reader) readBlock(
 	}
 
 	compressed := block.Alloc(int(bh.Length+block.TrailerLen), bufferPool)
-	readStartTime := time.Now()
+	readStopwatch := makeStopwatch()
 	var err error
 	if readHandle != nil {
 		err = readHandle.ReadAt(ctx, compressed.Get(), int64(bh.Offset))
 	} else {
 		err = r.readable.ReadAt(ctx, compressed.Get(), int64(bh.Offset))
 	}
-	readDuration := time.Since(readStartTime)
-	// TODO(sumeer): should the threshold be configurable.
-	const slowReadTracingThreshold = 5 * time.Millisecond
-	// For deterministic testing.
-	if deterministicReadBlockDurationForTesting {
-		readDuration = slowReadTracingThreshold
-	}
+	readDuration := readStopwatch.stop()
 	// Call IsTracingEnabled to avoid the allocations of boxing integers into an
 	// interface{}, unless necessary.
 	if readDuration >= slowReadTracingThreshold && r.logger.IsTracingEnabled(ctx) {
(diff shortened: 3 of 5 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3745#-:-O1YQd2XEgc0O_CSSfB3:bnfp4nl)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/base/logger.go` line 104 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3745#-O1YPdCj7ZdGvc5Gp4JW:-O1YPdCj7ZdGvc5Gp4JX:bpw96pb) ([raw file](https://github.com/cockroachdb/pebble/blob/fcbfbd22b6b7e6484fdfce9556e4a2ac1e4ecfaf/internal/base/logger.go#L104)):*
> ```Go
> 	// avoid calling Eventf (which will be a noop when tracing or expensive
> 	// logging is not enabled) to avoid the overhead of boxing the args.
> 	IsTracingEnabled(ctx context.Context) bool
> ```

BTW I noticed a possible issue in the `IsTracingEnabled` implementation in Cockroach - it calls `ExpensiveLogEnabled` which is meant to be called directly - specifically the `VDepth(1)` check gets the source file where `ExpensiveLogEnabled` is called. As things stand, you can only turn all pebble logs on (by using `pebble_logger_and_tracer=2`)
___
*[`sstable/table.go` line 330 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3745#-O1YOonE6A6yOzkyiU3R:-O1YOonE6A6yOzkyiU3S:b-xex4xo) ([raw file](https://github.com/cockroachdb/pebble/blob/fcbfbd22b6b7e6484fdfce9556e4a2ac1e4ecfaf/sstable/table.go#L330)):*
> ```Go
> 		err = f.ReadAt(ctx, buf, off)
> 	}
> 	readDuration := time.Since(readStartTime)
> ```

[nit] We should add a little helper type that would be used like:

```
l := MakeStopwatch(ctx)
...
if duration, shouldLog := l.Finish(); shouldLog {
}
```

This would also give us the flexibility to skip `time.Now()` if tracing is not enabled (not sure which is slower through - `time.Now` or `IsTracingEnabled`)

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3745#-:-O1b6GG_CLk_gSrSIyTC:b-814g8k)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/base/logger.go` line 104 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3745#-O1YPdCj7ZdGvc5Gp4JW:-O1ay4G1CEzhvcJaMjY7:b-fv5ir0) ([raw file](https://github.com/cockroachdb/pebble/blob/fcbfbd22b6b7e6484fdfce9556e4a2ac1e4ecfaf/internal/base/logger.go#L104)):*

BTW I noticed a possible issue in the `IsTracingEnabled` implementation in Cockroach - it calls `ExpensiveLogEnabled` which is meant to be called directly - specifically the `VDepth(1)` check gets the source file where `ExpensiveLogEnabled` is called. As things stand, you can only turn all pebble logs on (by using `pebble_logger_and_tracer=2`)
</blockquote></details>

Yes, I realized that when doing it, but thought it was not too harmful. I had convinced myself then that it was a simpler story than knowing the Pebble filename, but now I realize that was all wrong. Looking at the source code for `log.modulePat` there seems to be some support for pattern matching, so I suppose one can also include more of the path and we are not stuck with something like --vmodule=reader=2 picking up all reader.go files in CockroachDB and Pebble.

Since the log package functions do accept a depth parameter, we can fix this by changing the implementation in pebble_logger_and_tracer.go. I'll make that change.
___
*[`sstable/table.go` line 330 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3745#-O1YOonE6A6yOzkyiU3R:-O1b5fTCB2kXUCec_6oF:b-13xppm) ([raw file](https://github.com/cockroachdb/pebble/blob/fcbfbd22b6b7e6484fdfce9556e4a2ac1e4ecfaf/sstable/table.go#L330)):*

[nit] We should add a little helper type that would be used like:

```
l := MakeStopwatch(ctx)
...
if duration, shouldLog := l.Finish(); shouldLog {
}
```

This would also give us the flexibility to skip `time.Now()` if tracing is not enabled (not sure which is slower through - `time.Now` or `IsTracingEnabled`)
</blockquote></details>

I added a `stopwatch` to abstract out the `deterministicReadBlockDurationForTesting` logic.

Did not hide the call to `IsTracingEnabled` because it has the same problem (that we want to fix) that the original caller should be calling it. We can increase the depth when there is a constant extra level of indirection via pebble_logger_and_tracer.go but can't handle a variable amount of indirection.

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3745#-:-O1bCgGD3RPzvKMaeYlL:bnfp4nl)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/reader.go` line 1099 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3745#-O1bCZb24d4Hj9aTUjin:-O1bCZb24d4Hj9aTUjio:beh73t2) ([raw file](https://github.com/cockroachdb/pebble/blob/069a48b95fd4f15a27c2026d5647d884d25d950f/sstable/reader.go#L1099)):*
> ```Go
> func (w stopwatch) stop() time.Duration {
> 	dur := time.Since(w.startTime)
> 	if deterministicReadBlockDurationForTesting {
> ```

[nit] maybe rename to `deterministicDurationsforTesting` or `deterministicStopwatchForTesting`.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3745)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @jbowens)

___
*[`sstable/reader.go` line 1099 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3745#-O1bCZb24d4Hj9aTUjin:-O1bRMLl3Fh8LL15bAAF:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/069a48b95fd4f15a27c2026d5647d884d25d950f/sstable/reader.go#L1099)):*

[nit] maybe rename to `deterministicDurationsforTesting` or `deterministicStopwatchForTesting`.
</blockquote></details>

done

PR #3747, opened 2024-07-12
sstable: improve the property collector interface

RaduBerinde:
This change makes the following improvements:

 - we now pass range key spans through a separate `AddRangeKeys` API.
   This is cleaner and avoids an unnecessary decode step on a span we
   just encoded.  The Writer no longer has a public `AddRangeKey`
   method - `EncodeSpan` is used instead.

 - we simplify the interface for block interval collectors: we now
   have an interface to map point keys and range key spans to
   intervals, plus an optional interface to apply a synthetic suffix.
   This greatly simplifies the implementation on the CRDB side (and
   allows it to be stateless).

### DIFF
--- a/ingest_test.go
+++ b/ingest_test.go
@@ -103,7 +103,7 @@ func TestIngestLoad(t *testing.T) {
 				if strings.HasPrefix(data, "rangekey: ") {
 					data = strings.TrimPrefix(data, "rangekey: ")
 					s := keyspan.ParseSpan(data)
-					err := rangekey.Encode(&s, w.AddRangeKey)
+					err := w.EncodeSpan(&s)
 					if err != nil {
 						return err.Error()
 					}
@@ -1083,9 +1083,7 @@ func testIngestSharedImpl(
 						Keys:      keys,
 						KeysOrder: 0,
 					}
-					require.NoError(t, rangekey.Encode(&s, func(k base.InternalKey, v []byte) error {
-						return w.AddRangeKey(base.MakeInternalKey(k.UserKey, 0, k.Kind()), v)
-					}))
+					require.NoError(t, w.EncodeSpan(&s))
 					return nil
 				},
 				func(sst *SharedSSTMeta) error {
@@ -1581,9 +1579,7 @@ func TestConcurrentExcise(t *testing.T) {
 						Keys:      keys,
 						KeysOrder: 0,
 					}
-					require.NoError(t, rangekey.Encode(&s, func(k base.InternalKey, v []byte) error {
-						return w.AddRangeKey(base.MakeInternalKey(k.UserKey, 0, k.Kind()), v)
-					}))
+					require.NoError(t, w.EncodeSpan(&s))
 					return nil
 				},
 				func(sst *SharedSSTMeta) error {
@@ -2016,9 +2012,7 @@ func TestIngestExternal(t *testing.T) {
 						Keys:      keys,
 						KeysOrder: 0,
 					}
-					require.NoError(t, rangekey.Encode(&s, func(k base.InternalKey, v []byte) error {
-						return w.AddRangeKey(base.MakeInternalKey(k.UserKey, 0, k.Kind()), v)
-					}))
+					require.NoError(t, w.EncodeSpan(&s))
 					return nil
 				},
 				nil,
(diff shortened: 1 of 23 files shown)


RaduBerinde:
The CRDB-side change are here: https://github.com/cockroachdb/cockroach/pull/127047

jbowens (approved):
:lgtm_strong: 

> Reviewed 6 of 23 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3747#-:-O1s713P2rprJG8TKdGv:b-uwkowd)* status: 6 of 23 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/block_property.go` line 110 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3747#-O1bJ45x9eifcptt7iUb:-O1bJ45x9eifcptt7iUc:bsme354) ([raw file](https://github.com/cockroachdb/pebble/blob/33071d91010e3ac453760d195be3846d4d39cadd/sstable/block_property.go#L110)):*
> ```Go
> 	// key properties are stored separately and don't contribute to data block
> 	// properties. They are only used when FinishTable is called.
> 	// TODO(radu): clean up this subtle semantic.
> ```

just curious what you had in mind for this TODO?
___
*[`sstable/block_property.go` line 269 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3747#-O1rwkbo0uLgOSP2bdEp:-O1rwkbo0uLgOSP2bdEq:bmtksav) ([raw file](https://github.com/cockroachdb/pebble/blob/33071d91010e3ac453760d195be3846d4d39cadd/sstable/block_property.go#L269)):*
> ```Go
> // should be provided, rather than the same collector passed in twice (see the
> // comment on BlockIntervalCollector for more detail)
> // XXX update
> ```

wip comment?
___
*[`sstable/block_property.go` line 432 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3747#-O1rxlXQESPBaRhfDlv-:-O1rxlXQESPBaRhfDlv0:b-l1unka) ([raw file](https://github.com/cockroachdb/pebble/blob/33071d91010e3ac453760d195be3846d4d39cadd/sstable/block_property.go#L432)):*
> ```Go
> 	// ApplySuffixReplacement recalculates a previously calculated interval (which
> 	// corresponds to an arbitrary collection of keys) under the assumption
> 	// that those keys are rewritten with a new prefix.
> ```

nit: "new suffix"?
___
*[`sstable/writer.go` line 1157 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3747#-O1s6WQ86ewPSe9Ti6Gn:-O1s6WQ86ewPSe9Ti6Go:b-7108g6) ([raw file](https://github.com/cockroachdb/pebble/blob/33071d91010e3ac453760d195be3846d4d39cadd/sstable/writer.go#L1157)):*
> ```Go
> }
> 
> func (w *Writer) addRangeKeySpanToFragmenter(span keyspan.Span) error {
> ```

nice! I'm thinking about pulling the convenience functions like `Set`, `Delete`, `Merge`, `RangeKeySet`, etc out onto a separate type (maybe `Writer` and `RawWriter`?), so good riddance to `Writer.AddRangeKey`.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3747)* status: 6 of 23 files reviewed, 4 unresolved discussions (waiting on @jbowens)

___
*[`sstable/block_property.go` line 110 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3747#-O1bJ45x9eifcptt7iUb:-O1sEFzY4S2f8bv2_srD:b-w6z5mm) ([raw file](https://github.com/cockroachdb/pebble/blob/33071d91010e3ac453760d195be3846d4d39cadd/sstable/block_property.go#L110)):*

just curious what you had in mind for this TODO?
</blockquote></details>

I want to simplify the interface to just

```
AddPointKey
AddRangeKeys
AddCollectedWithSuffixReplacement
EncodeProperty
Reset
```

We would use multiple separate instances (one for data blocks, one for first-level index blocks, one for top-level index block, one for range key blocks). The higher level ones would use `AddCollectedWithSuffixReplacement` with the properties computed by the lower level.

Since the only implementation really is the interval one, we can use a `sync.Pool` to save on the allocations.

PR #3754, opened 2024-07-12
colblk: only align uint deltas

jbowens:
Previously, even when using a delta-encoding, the uint column encoding aligned its output to the width of the logical uint. This was necessary to read the full-width constant using a pointer cast. This commit adapts the encoding to only align to the width of the deltas, using ordinary binary.LittleEndian methods for setting and getting the constant. This get only needs to happen once, while parsing the column. This reduces the amount of alignment padding necessary when delta encoding is effective.

### DIFF
--- a/sstable/colblk/column.go
+++ b/sstable/colblk/column.go
@@ -192,6 +192,21 @@ func (d DeltaEncoding) String() string {
 	}
 }
 
+func (d DeltaEncoding) width() int {
+	switch d {
+	case DeltaEncodingConstant:
+		return 0
+	case DeltaEncodingUint8:
+		return 1
+	case DeltaEncodingUint16:
+		return 2
+	case DeltaEncodingUint32:
+		return 4
+	default:
+		panic("unreachable")
+	}
+}
+
 // ColumnWriter is an interface implemented by column encoders that accumulate a
 // column's values and then serialize them.
 type ColumnWriter interface {
--- a/sstable/colblk/testdata/raw_bytes
+++ b/sstable/colblk/testdata/raw_bytes
@@ -16,16 +16,14 @@ a
 build offset=1
 a
 ----
-00-01: x 00       # start offset
+0-1: x 00       # start offset
 # RawBytes
 # Offsets table
-# Padding
-01-04: x 000000   # aligning to 32-bit boundary
-04-08: x 00000000 # 32-bit constant: 0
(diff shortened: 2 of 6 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3754#-:-O1d5WED4cAHrypLSvvB:bnfp4nl)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`sstable/colblk/unsafe_slice.go` line 83 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3754#-O1d5FNQ29XTU3FS_8nr:-O1d5FNQ29XTU3FS_8ns:berqifq) ([raw file](https://github.com/cockroachdb/pebble/blob/cf87d1399936fc6612e86fb7415c08d13f0a71f3/sstable/colblk/unsafe_slice.go#L83)):*
> ```Go
> 		slice = makeUnsafeIntegerSlice[T](base, unsafe.Pointer(&b[off]), 0)
> 	case DeltaEncodingUint8:
> 		base := readLittleEndianNonaligned[T](b, off)
> ```

[nit] These cases (other than None) can be made into a single block if we add a function that converts from `DeltaEncodingConstant/Uint8/16/32` to `0/1/4/8` respectively.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3754)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @itsbilal)

___
*[`sstable/colblk/unsafe_slice.go` line 83 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3754#-O1d5FNQ29XTU3FS_8nr:-O1dhAYD7aACsM31bdfW:bln7q2m) ([raw file](https://github.com/cockroachdb/pebble/blob/cf87d1399936fc6612e86fb7415c08d13f0a71f3/sstable/colblk/unsafe_slice.go#L83)):*

[nit] These cases (other than None) can be made into a single block if we add a function that converts from `DeltaEncodingConstant/Uint8/16/32` to `0/1/4/8` respectively.
</blockquote></details>

done, although I kept the constant as a separate case because `off = align(off, 0)` will not do what we want

jbowens:
TFTR!

PR #3755, opened 2024-07-12
objstorage: fix race in vfsSync

RaduBerinde:
#### objstorage: improve ParallelSync test to expose race in vfsSync

We improve this test by using a strict MemFS, simulating a crash at
a random point, and checking that all objects that were synced are
accessible.

#### objstorage: fix race in vfsSync

We have a race in `vfsSync`: if two goroutines Sync at the same time,
one might see the flag=false and return immediately, before the other
goroutine actually runs/completes the Sync.

The fix is to store a change counter and wait for any in-progress
syncs as necessary.

Informs https://github.com/cockroachdb/cockroach/issues/124845

### DIFF
--- a/objstorage/objstorageprovider/provider.go
+++ b/objstorage/objstorageprovider/provider.go
@@ -40,9 +40,14 @@ type provider struct {
 
 		remote remoteLockedState
 
-		// localObjectsChanged is set if non-remote objects were created or deleted
-		// but Sync was not yet called.
-		localObjectsChanged bool
+		// TODO(radu): move these fields to a localLockedState struct.
+		// localObjectsChanged is incremented whenever non-remote objects are created.
+		// The purpose of this counter is to avoid syncing the local filesystem when
+		// only remote objects are changed.
+		localObjectsChangeCounter uint64
+		// localObjectsChangeCounterSynced is the value of localObjectsChangeCounter
+		// value at the time the last completed sync was launched.
+		localObjectsChangeCounterSynced uint64
 
 		// knownObjects maintains information about objects that are known to the provider.
 		// It is initialized with the list of files in the manifest when we open a DB.
@@ -583,7 +588,7 @@ func (p *provider) addMetadataLocked(meta objstorage.ObjectMetadata) {
 			p.mu.remote.addExternalObject(meta)
 		}
 	} else {
-		p.mu.localObjectsChanged = true
+		p.mu.localObjectsChangeCounter++
 	}
 }
 
@@ -602,7 +607,7 @@ func (p *provider) removeMetadata(fileNum base.DiskFileNum) {
 	if meta.IsRemote() {
 		p.mu.remote.catalogBatch.DeleteObject(fileNum)
 	} else {
-		p.mu.localObjectsChanged = true
+		p.mu.localObjectsChangeCounter++
 	}
 }
 
--- a/objstorage/objstorageprovider/provider_test.go
+++ b/objstorage/objstorageprovider/provider_test.go
@@ -10,7 +10,9 @@ import (
 	"math/rand"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/cockroachdb/datadriven"
 	"github.com/cockroachdb/pebble/internal/base"
@@ -581,7 +583,8 @@ func TestParallelSync(t *testing.T) {
 			name = "shared"
 		}
 		t.Run(name, func(t *testing.T) {
-			st := DefaultSettings(vfs.NewMem(), "")
+			fs := vfs.NewStrictMem()
+			st := DefaultSettings(fs, "")
 			st.Remote.StorageFactory = remote.MakeSimpleFactory(map[remote.Locator]remote.Storage{
 				"": remote.NewInMem(),
 			})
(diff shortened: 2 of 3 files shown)


jbowens (approved):
do we know how expensive a sync is that should semantically be a no-op? I'm wondering if we should just naively let the operating system do its thing, presumably avoiding the actual I/O if the directory inode's page is not dirty.

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3755#-:-O1hrIFR8b-Dhb_0wzm5:bes8r8k)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @pav-kv, @RaduBerinde, and @sumeerbhola)

___
*[`objstorage/objstorageprovider/provider.go` line 48 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3755#-O1hr7Qj7mRdwlx8Payy:-O1hr7Qj7mRdwlx8Payz:boq2hvn) ([raw file](https://github.com/cockroachdb/pebble/blob/0b36210cf91081d3c81b02a62c1f8f14f414c674/objstorage/objstorageprovider/provider.go#L48)):*
> ```Go
> 		// lastLocalSyncCounter is the value of localObjectsChangeCounter value at the
> 		// time the last completed sync was launched.
> 		lastLocalSyncCounter uint64
> ```

nit: maybe `localObjectsChangeCounterSynced` to make the relationship with `localObjectsChangeCounter` more obvious?
___
*[`objstorage/objstorageprovider/provider_test.go` line 644 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3755#-O1hpt0v-7rHqVHbk19P:-O1hpt0v-7rHqVHbk19Q:b-s3l217) ([raw file](https://github.com/cockroachdb/pebble/blob/d5e468fdf051ca423ea9505c1ca1fdc330b3682c/objstorage/objstorageprovider/provider_test.go#L644)):*
> ```Go
> 								panic(err)
> 							}
> 							// If we stopped during the Sync, we don't expect the file to
> ```

nit: trailing comment
___
*[`objstorage/objstorageprovider/provider_test.go` line 645 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3755#-O1hpuMd3Lnj_N84-cVY:-O1hpuMd3Lnj_N84-cVZ:b-f8jgzg) ([raw file](https://github.com/cockroachdb/pebble/blob/d5e468fdf051ca423ea9505c1ca1fdc330b3682c/objstorage/objstorageprovider/provider_test.go#L645)):*
> ```Go
> 							}
> 							// If we stopped during the Sync, we don't expect the file to
> 							setMustExist(num, true)
> ```

regarding the comment above, should this be something like:
```
setMustExist(num, !stop.Load())
```
?

RaduBerinde:
Good point. I'm not sure..  The main reason we had the flag is to avoid the Sync when only remote objects were changed. But maybe it doesn't matter

RaduBerinde (commented):
Or we could still keep track of the counter, just remove the extra channel synchronization stuff and sometimes do two Syncs? That would simplify the code with no real downsides

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3755#-:-O1r7tYe7fviKWrrgMW2:b7kfg3k)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @pav-kv and @sumeerbhola)

jbowens (commented):
Yeah, that sounds reasonable to me

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3755#-:-O1rAo8gBTioVVseaAMP:bunrrt6)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @pav-kv, @RaduBerinde, and @sumeerbhola)

RaduBerinde (commented):
Done, thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3755#-:-O1sUwc41T2OMqk7-aFl:b-dtkds3)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jbowens, @pav-kv, and @sumeerbhola)

___
*[`objstorage/objstorageprovider/provider_test.go` line 645 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3755#-O1hpuMd3Lnj_N84-cVY:-O1sUY7q0wsBRUAkZYzE:by78270) ([raw file](https://github.com/cockroachdb/pebble/blob/d5e468fdf051ca423ea9505c1ca1fdc330b3682c/objstorage/objstorageprovider/provider_test.go#L645)):*

regarding the comment above, should this be something like:
```
setMustExist(num, !stop.Load())
```
?
</blockquote></details>

No, the comment was left over from an earlier iteration where I wasn't taking a "snapshot" of `mustExist`

RaduBerinde:
@jbowens I will wait for one more pass over the final version in case I screwed something up. I plan to backport this to 23.2 and later.

jbowens (approved):
:lgtm_strong: 

> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3755#-:-O1xVMIk5D0o2UqN5mQc:b-uwkowd)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @pav-kv and @sumeerbhola)

RaduBerinde:
TFTR!

PR #3756, opened 2024-07-13
colblk: add PrefixBytes, PrefixBytesBuilder

jbowens:
Add a new column data type primitive, PrefixBytes, for encoding lexicographically-sorted byte slices. This data type applies prefix compression at two levels: a) column-wide and b) across "bundles" of N keys.

### DIFF
--- a/sstable/colblk/column.go
+++ b/sstable/colblk/column.go
@@ -220,8 +220,10 @@ type ColumnWriter interface {
 	//
 	// The supplied buf must have enough space at the provided offset to fit the
 	// column. The caller may use Size() to calculate the exact size required.
-	// If [rows] is ≤ the highest row index at which a value has been set,
-	// Finish will only serialize the first [rows] values.
+	// The caller passes the number of rows they want to serialize. All
+	// implementations of Finish must support cases where rows is the number of
+	// rows the caller has set, or one less. Some implementations may be more
+	// permissive.
 	//
 	// The provided column index must be less than NumColumns(). Finish is
 	// called for each index < NumColumns() in order.
--- a/sstable/colblk/prefix_bytes.go
+++ b/sstable/colblk/prefix_bytes.go
@@ -1,0 +1,956 @@
+// Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package colblk
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"io"
+	"math/bits"
+	"strings"
+	"unsafe"
+
+	"github.com/cockroachdb/errors"
+	"github.com/cockroachdb/pebble/internal/binfmt"
+	"github.com/cockroachdb/pebble/internal/invariants"
+)
+
+// PrefixBytes holds an array of lexicograpghically ordered byte slices. It
+// provides prefix compression. Prefix compression applies strongly to two cases
+// in CockroachDB: removal of the "[/tenantID]/tableID/indexID" prefix that is
+// present on all table data keys, and multiple versions of a key that are
+// distinguished only by different timestamp suffixes. With columnar blocks
+// enabling the timestamp to be placed in a separate column, the multiple
+// version problem becomes one of efficiently handling exact duplicate keys.
+// PrefixBytes builds offs of the RawBytes encoding, introducing n/bundleSize+1
+// additional slices for encoding n/bundleSize bundle prefixes and 1
+// column-level prefix.
+//
+// To understand the PrefixBytes layout, we'll work through an example using
+// these 15 keys:
+//
+//	   01234567890
+//	 0 aaabbbc
+//	 1 aaabbbcc
+//	 2 aaabbbcde
+//	 3 aaabbbce
+//	 4 aaabbbdee*
+//	 5 aaabbbdee*
+//	 6 aaabbbdee*
+//	 7 aaabbbeff
+//	 8 aaabbe
+//	 9 aaabbeef*
+//	10 aaabbeef*
+//	11 aaabc
+//	12 aabcceef*
+//	13 aabcceef*
+//	14 aabcceef*
+//
+// The total length of these keys is 127 bytes. There are 3 keys which occur
+// multiple times (delineated by the * suffix) which models multiple versions of
+// the same MVCC key in CockroachDB. There is a shared prefix to all of the keys
+// which models the "[/tenantID]/tableID/indexID" present on CockroachDB table
+// data keys. There are other shared prefixes which model identical values in
+// table key columns.
+//
+// The table below shows the components of the KeyBytes encoding for these 15
+// keys when using a bundle size of 4 which results in 4 bundles. The 15 keys
+// are encoded into 20 slices: 1 block prefix, 4 bundle prefixes, and 15
+// suffixes. The first slice in the table is the block prefix that is shared by
+// all keys in the block. The first slice in each bundle is the bundle prefix
+// which is shared by all keys in the bundle.
+//
+//	 idx   | row   | offset | data
+//	-------+-------+--------+------
+//	     0 |       |      2 | aa
+//	     1 |       |      7 | ..abbbc
+//	     2 |     0 |      7 | .......
+//	     3 |     1 |      8 | .......c
+//	     4 |     2 |     10 | .......de
+//	     5 |     3 |     11 | .......e
+//	     6 |       |     15 | ..abbb
+//	     7 |     4 |     19 | ......dee*
+//	     8 |     5 |     19 | ......
+//	     9 |     6 |     19 | ......
+//	    10 |     7 |     22 | ......eff
+//	    11 |       |     24 | ..ab
+//	    12 |     8 |     26 | ....be
+//	    13 |     9 |     31 | ....beef*
+//	    14 |    10 |     31 | ....
+//	    15 |    11 |     32 | ....c
+//	    16 |       |     39 | ..bcceef*
+//	    17 |    12 |     39 | .........
+//	    18 |    13 |     39 | .........
+//	    19 |    14 |     39 | .........
+//
+// The offset column in the table points to the start and end index within the
+// RawBytes data array for each of the 20 slices defined above (the 15 key
+// suffixes + 4 bundle key prefixes + block key prefix). Offset[0] is the length
+// of the first slice which is always anchored at data[0]. The data columns
+// display the portion of the data array the slice covers. For row slices, an
+// empty suffix column indicates that the slice is identical to the slice at the
+// previous index which is indicated by the slice's offset being equal to the
+// previous slice's offset. Due to the lexicographic sorting, the key at row i
+// can't be a prefix of the key at row i-1 or it would have sorted before the
+// key at row i-1. And if the key differs then only the differing bytes will be
+// part of the suffix and not contained in the bundle prefix.
+//
+// The end result of this encoding is that we can store the 127 bytes of the 15
+// keys plus their start and end offsets (which would naively consume 15*4=60
+// bytes for at least the key lengths) in 64 bytes (39 bytes of data + 4 bytes
+// of offset constant + 20 bytes of offset delta data + 1 byte of bundle size).
+//
+// # Physical representation
+//
+//	+==================================================================+
+//	|                        Bundle size (1 byte)                      |
+//	|                                                                  |
+//	| The bundle size indicates how many keys prefix compression may   |
+//	| apply across. Every bundleSize keys, prefix compression restarts.|
+//	| The bundleSize is required to be a power of two,  and this 1-    |
+//	| byte prefix stores log2(bundleSize).                             |
+//	+==================================================================+
+//	|                            RawBytes                              |
+//	|                                                                  |
+//	| A modified RawBytes encoding is used to store the data slices. A |
+//	| PrefixBytes column storing n keys will encode 2+n+n/bundleSize   |
+//	| slices. Unlike the RawBytes encoding, the first offset encoded   |
+//	| is not guaranteed to be zero. In the PrefixBytes encoding, the   |
+//	| first offset encodes the length of the column-wide prefix. The   |
+//	| column-wide prefix is stored in slice(0, offset(0)).             |
+//	|                                                                  |
+//	|  +------------------------------------------------------------+  |
+//	|  |                       Offset table                         |  |
+//	|  |                                                            |  |
+//	|  | A Uint32 column encoding offsets into the string data,     |  |
+//	|  | possibly delta8 or delta16 encoded. When a delta encoding  |  |
+//	|  | is used, the base constant is always zero.                 |  |
+//	|  +------------------------------------------------------------+  |
+//	|  | offsetDelta[0] | offsetDelta[1] | ... | offsetDelta[m]     |  |
+//	|  +------------------------------------------------------------+  |
+//	|  | prefix-compressed string data                              |  |
+//	|  | ...                                                        |  |
+//	|  +------------------------------------------------------------+  |
+//	+==================================================================+
+//
+// # Reads
+//
+// This encoding provides O(1) access to any row by calculating the bundle for
+// the row (5*(row/4)), then the row's index within the bundle (1+(row%4)). If
+// the slice's offset equals the previous slice's offset then we step backward
+// until we find a non-empty slice or the start of the bundle (a variable number
+// of steps, but bounded by the bundle size).
+//
+// Forward iteration can easily reuse the previous row's key with a check on
+// whether the row's slice is empty.  Reverse iteration can reuse the next row's
+// key by looking at the next row's offset to determine whether we are in the
+// middle of a run of equal keys or at an edge. When reverse iteration steps
+// over an edge it has to continue backward until a non-empty slice is found
+// (just as in absolute positioning).
+//
+// The Seek{GE,LT} routines first binary search on the first key of each bundle
+// which can be retrieved without data movement because the bundle prefix is
+// immediately adjacent to it in the data array. We can slightly optimize the
+// binary search by skipping over all of the keys in the bundle on prefix
+// mismatches.
+type PrefixBytes struct {
+	rows            int
+	bundleShift     int
+	bundleMask      int
+	sharedPrefixLen int
+	rawBytes        RawBytes
+}
+
+// MakePrefixBytes constructs an accessor for an array of lexicographically
+// sorted byte slices constructed by PrefixBytesBuilder. Count must be the
+// number of logical slices within the array.
+func MakePrefixBytes(count int, b []byte, offset uint32, enc ColumnEncoding) PrefixBytes {
+	// The first byte of a PrefixBytes-encoded column is the bundle size
+	// expressed as log2 of the bundle size (the bundle size must always be a
+	// power of two).
+	bundleShift := int(*((*uint8)(unsafe.Pointer(&b[offset]))))
+	nBundles := 1 + (count-1)>>bundleShift
+
+	pb := PrefixBytes{
+		rows:        count,
+		bundleShift: bundleShift,
+		bundleMask:  ^((1 << bundleShift) - 1),
+		rawBytes:    MakeRawBytes(count+nBundles, b, offset+1, enc),
+	}
+	// We always set the base to zero.
+	if pb.rawBytes.offsets.base != 0 {
+		panic(errors.AssertionFailedf("unexpected non-zero base in offsets"))
+	}
+	pb.sharedPrefixLen = int(pb.rawBytes.offsets.At(0))
+	return pb
+}
+
+// SharedPrefix return a []byte of the shared prefix that was extracted from
+// all of the values in the Bytes vector. The returned slice should not be
+// mutated.
+func (b PrefixBytes) SharedPrefix() []byte {
+	// The very first slice is the prefix for the entire column.
+	return b.rawBytes.slice(0, b.rawBytes.offsets.At(0))
+}
+
+// RowBundlePrefix takes a row index and returns a []byte of the prefix shared
+// among all the keys in the row's bundle. The returned slice should not be
+// mutated.
+func (b PrefixBytes) RowBundlePrefix(row int) []byte {
+	// AND-ing the row with the bundle mask removes the least significant bits
+	// of the row, which encode the row's index within the bundle.
+	i := (row >> b.bundleShift) + (row & b.bundleMask)
+	return b.rawBytes.slice(b.rawBytes.offsets.At(i), b.rawBytes.offsets.At(i+1))
+}
+
+// BundlePrefix returns the prefix of the i-th bundle in the column. The
+// provided i must be in the range [0, Bundles()). The returned slice should not
+// be mutated.
+func (b PrefixBytes) BundlePrefix(i int) []byte {
+	j := (i << b.bundleShift) + i
+	return b.rawBytes.slice(b.rawBytes.offsets.At(j), b.rawBytes.offsets.At(j+1))
+}
+
+// RowSuffix returns a []byte of the suffix unique to the row. A row's full key
+// is the result of concatenating SharedPrefix(), BundlePrefix() and
+// RowSuffix().
+//
+// The returned slice should not be mutated.
+func (b PrefixBytes) RowSuffix(row int) []byte {
+	i := 1 + (row >> b.bundleShift) + row
+	// Retrieve the low and high offsets indicating the start and end of the
+	// row's suffix slice.
+	lowOff := b.rawBytes.offsets.At(i)
+	highOff := b.rawBytes.offsets.At(i + 1)
+	// If there's a non-empty slice for the row, this row is different than its
+	// predecessor.
+	if lowOff != highOff {
+		return b.rawBytes.slice(lowOff, highOff)
+	}
+	// Otherwise, an empty slice indicates a duplicate key. We need to find the
+	// first non-empty predecessor within the bundle, or if all the rows are
+	// empty, return nil.
+	//
+	// Compute the index of the first row in the bundle so we know when to stop.
+	firstIndex := 1 + (row >> b.bundleShift) + (row & b.bundleMask)
+	for i > firstIndex {
+		// Step back a row, and check if the slice is non-empty.
+		i--
+		highOff = lowOff
+		lowOff = b.rawBytes.offsets.At(i)
+		if lowOff != highOff {
+			return b.rawBytes.slice(lowOff, highOff)
+		}
+	}
+	// All the rows in the bundle are empty.
+	return nil
+}
+
+// Rows returns the count of rows whose keys are encoded within the PrefixBytes.
+func (b PrefixBytes) Rows() int {
+	return b.rows
+}
+
+// BundleCount returns the count of bundles within the PrefixBytes.
+func (b PrefixBytes) BundleCount() int {
+	return 1 + (b.rows-1)>>b.bundleShift
+}
+
+// Search searchs for the first key in the PrefixBytes that is greater than or
+// equal to k, returning the index of the key and whether an equal key was
+// found.
+func (b PrefixBytes) Search(k []byte) (int, bool) {
+	// First compare to the block-level shared prefix.
+	n := min(len(k), b.sharedPrefixLen)
+	c := bytes.Compare(k[:n], unsafe.Slice((*byte)(b.rawBytes.data), b.sharedPrefixLen))
+	switch {
+	case c < 0 || (c == 0 && n < b.sharedPrefixLen):
+		// Search key is less than any prefix in the block.
+		return 0, false
+	case c > 0:
+		// Search key is greater than any key in the block.
+		return b.rows, false
+	}
+	// Trim the block-level shared prefix from the search key.
+	k = k[b.sharedPrefixLen:]
+
+	// Binary search among the first keys of each bundle.
+	//
+	// Define f(-1) == false and f(upper) == true.
+	// Invariant: f(bi-1) == false, f(upper) == true.
+	nBundles := b.BundleCount()
+	bi, upper := 0, nBundles
+	upperEqual := false
+	for bi < upper {
+		h := int(uint(bi+upper) >> 1) // avoid overflow when computing h
+		// bi ≤ h < upper
+
+		// Retrieve the first key in the h-th (zero-indexed) bundle. We take
+		// advantage of the fact that the first row is stored contiguously in
+		// the data array (modulo the block prefix) to slice the entirety of the
+		// first key:
+		//
+		//       b u n d l e p r e f i x f i r s t k e y r e m a i n d e r
+		//       ^                       ^                                 ^
+		//     offset(j)             offset(j+1)                       offset(j+2)
+		//
+		j := (h << b.bundleShift) + h
+		bundleFirstKey := b.rawBytes.slice(b.rawBytes.offsets.At(j), b.rawBytes.offsets.At(j+2))
+		c = bytes.Compare(k, bundleFirstKey)
+		switch {
+		case c > 0:
+			bi = h + 1 // preserves f(bi-1) == false
+		case c < 0:
+			upper = h // preserves f(upper) == true
+			upperEqual = false
+		default:
+			// c == 0
+			upper = h // preserves f(upper) == true
+			upperEqual = true
+		}
+	}
+	if bi == 0 {
+		// The very first key is ≥ k. Return it.
+		return 0, upperEqual
+	}
+	// The first key of the bundle bi is ≥ k, but any of the keys in the
+	// previous bundle besides the first could also be ≥ k. We can binary search
+	// among them, but if the seek key doesn't share the previous bundle's
+	// prefix there's no need.
+	j := ((bi - 1) << b.bundleShift) + (bi - 1)
+	bundlePrefix := b.rawBytes.slice(b.rawBytes.offsets.At(j), b.rawBytes.offsets.At(j+1))
+	if len(bundlePrefix) > len(k) || !bytes.Equal(k[:len(bundlePrefix)], bundlePrefix) {
+		// The search key doesn't share the previous bundle's prefix, so all of
+		// the keys in the previous bundle must be less than k. We know the
+		// first key of bi is ≥ k, so return it.
+		if bi<<b.bundleShift < b.rows {
+			return bi << b.bundleShift, upperEqual
+		} else {
+			return b.rows, false
+		}
+	}
+	// Binary search among bundle bi-1's key remainders after stripping bundle
+	// bi-1's prefix.
+	//
+	// Define f(l-1) == false and f(u) == true.
+	// Invariant: f(l-1) == false, f(u) == true.
+	k = k[len(bundlePrefix):]
+	l := 1
+	u := min(1<<b.bundleShift, b.rows-(bi-1)<<b.bundleShift)
+	for l < u {
+		h := int(uint(l+u) >> 1) // avoid overflow when computing h
+		// l ≤ h < u
+
+		// j is currently the index of the offset of bundle bi-i's prefix.
+		//
+		//     b u n d l e p r e f i x f i r s t k e y s e c o n d k e y
+		//     ^                       ^               ^
+		//  offset(j)              offset(j+1)     offset(j+2)
+		//
+		// The beginning of the zero-indexed i-th key of the bundle is at
+		// offset(j+i+1).
+		//
+		hStart := b.rawBytes.offsets.At(j + h + 1)
+		hEnd := b.rawBytes.offsets.At(j + h + 2)
+		// There's a complication with duplicate keys. When keys are repeated,
+		// the PrefixBytes encoding avoids re-encoding the duplciate key,
+		// instead encoding an empty slice. While binary searching, if we land
+		// on an empty slice, we need to back up until we find a non-empty slice
+		// which is the key at index h. We iterate with p. If we eventually find
+		// the duplicated key at index p < h and determine f(p) == true, then we
+		// can set u=p (rather than h). If we determine f(p)==false, then we
+		// know f(h)==false too and set l=h+1.
+		p := h
+		if hStart == hEnd {
+			// Back up looking for an empty slice.
+			for hStart == hEnd && p >= l {
+				p--
+				hEnd = hStart
+				hStart = b.rawBytes.offsets.At(j + p + 1)
+			}
+			// If we backed up to l-1, then all the rows in indexes [l, h] have
+			// the same keys as index l-1. We know f(l-1) == false [see the
+			// invariants above], so we can move l to h+1 and continue the loop
+			// without performing any key comparisons.
+			if p < l {
+				l = h + 1
+				continue
+			}
+		}
+		rem := b.rawBytes.slice(hStart, hEnd)
+		c = bytes.Compare(k, rem)
+		switch {
+		case c > 0:
+			l = h + 1 // preserves f(l-1) == false
+		case c < 0:
+			u = p // preserves f(u) == true
+			upperEqual = false
+		default:
+			// c == 0
+			u = p // preserves f(u) == true
+			upperEqual = true
+		}
+	}
+	i := (bi-1)<<b.bundleShift + l
+	if i < b.rows {
+		return i, upperEqual
+	} else {
+		return b.rows, false
+	}
+}
+
+func prefixBytesToBinFormatter(
+	f *binfmt.Formatter, count int, enc ColumnEncoding, sliceFormatter func([]byte) string,
+) {
+	if sliceFormatter == nil {
+		sliceFormatter = defaultSliceFormatter
+	}
+	pb := MakePrefixBytes(count, f.Data(), uint32(f.Offset()), enc)
+	f.CommentLine("PrefixBytes")
+	f.HexBytesln(1, "bundleSize: %d", 1<<pb.bundleShift)
+	f.CommentLine("Offsets table")
+	dataOffset := uint64(f.Offset()) + uint64(uintptr(pb.rawBytes.data)-uintptr(pb.rawBytes.start))
+	uintsToBinFormatter(f, pb.rawBytes.slices+1, ColumnDesc{DataType: DataTypeUint32, Encoding: enc},
+		func(offsetDelta, offsetBase uint64) string {
+			// NB: offsetBase will always be zero for PrefixBytes columns.
+			return fmt.Sprintf("%d [%d overall]", offsetDelta+offsetBase, offsetDelta+offsetBase+dataOffset)
+		})
+	f.CommentLine("Data")
+
+	// The first offset encodes the length of the block prefix.
+	blockPrefixLen := pb.rawBytes.offsets.At(0)
+	f.HexBytesln(int(blockPrefixLen), "data[00]: %s (block prefix)",
+		sliceFormatter(pb.rawBytes.slice(0, blockPrefixLen)))
+
+	k := 2 + (count-1)>>pb.bundleShift + count
+	startOff := blockPrefixLen
+	prevLen := blockPrefixLen
+
+	// Use dots to indicate string data that's elided because it falls within
+	// the block or bundle prefix.
+	dots := strings.Repeat(".", int(blockPrefixLen))
+	// Iterate through all the slices in the data section, annotating bundle
+	// prefixes and using dots to indicate elided data.
+	for i := 1; i < k; i++ {
+		endOff := pb.rawBytes.offsets.At(i)
+		if (i-1)%(1+(1<<pb.bundleShift)) == 0 {
+			// This is a bundle prefix.
+			dots = strings.Repeat(".", int(blockPrefixLen))
+			f.HexBytesln(int(endOff-startOff), "data[%02d]: %s%s (bundle prefix)", i, dots, sliceFormatter(pb.rawBytes.At(i-1)))
+			dots = strings.Repeat(".", int(endOff-startOff+blockPrefixLen))
+			prevLen = endOff - startOff + blockPrefixLen
+		} else if startOff == endOff {
+			// An empty slice that's not a block or bundle prefix indicates a
+			// repeat key.
+			f.HexBytesln(0, "data[%02d]: %s", i, strings.Repeat(".", int(prevLen)))
+		} else {
+			f.HexBytesln(int(endOff-startOff), "data[%02d]: %s%s", i, dots, sliceFormatter(pb.rawBytes.At(i-1)))
+			prevLen = uint32(len(dots)) + endOff - startOff
+		}
+		startOff = endOff
+	}
+}
+
+// PrefixBytesBuilder encodes a column of lexicographically-sorted byte slices,
+// applying prefix compression to reduce the encoded size.
+type PrefixBytesBuilder struct {
+	// TODO(jackson): If prefix compression is very effective, the encoded size
+	// may remain very small while the physical in-memory size of the
+	// in-progress data slice may grow very large. This may pose memory usage
+	// problems during block building.
+	data                      []byte // The raw, concatenated keys w/o any prefix compression
+	nKeys                     int    // The number of keys added to the builder
+	bundleSize                int    // The number of keys per bundle
+	bundleShift               int    // log2(bundleSize)
+	completedBundleLen        int    // The encoded size of completed bundles
+	currentBundleLen          int    // The raw size of the current bundle's keys
+	currentBundleKeys         int    // The number of physical keys in the current bundle
+	currentBundlePrefixOffset int    // The index of the offset for the current bundle prefix
+	blockPrefixLen            uint32 // The length of the block-level prefix
+	blockPrefixLenUpdated     int    // The row index of the last row that updated the block prefix
+	lastKeyLen                int    // The length of the last key added to the builder
+	lastLastKeyLen            int    // The length of the key before the last key added to the builder
+	offsets                   struct {
+		count int // The number of offsets in the builder
+		// elemsSize is the size of the array (in count of uint32 elements; not
+		// bytes)
+		elemsSize int
+		// elems provides access to elements without bounds checking. elems is
+		// grown automatically in addOffset.
+		elems UnsafeRawSlice[uint32]
+	}
+	maxShared uint16
+}
+
+// Init initializes the PrefixBytesBuilder with the specified bundle size. The
+// builder will produce a prefix-compressed column of data type
+// DataTypePrefixBytes. The [bundleSize] indicates the number of keys that form
+// a "bundle," across which prefix-compression is applied. All keys in the
+// column will share a column-wide prefix if there is one.
+func (b *PrefixBytesBuilder) Init(bundleSize int) {
+	if bundleSize > 0 && (bundleSize&(bundleSize-1)) != 0 {
+		panic(errors.AssertionFailedf("prefixbytes bundle size %d is not a power of 2", bundleSize))
+	}
+	*b = PrefixBytesBuilder{
+		data:       b.data[:0],
+		bundleSize: bundleSize,
+		offsets:    b.offsets,
+	}
+	b.offsets.count = 0
+	if b.bundleSize > 0 {
+		b.bundleShift = bits.TrailingZeros32(uint32(bundleSize))
+		b.maxShared = (1 << 16) - 1
+	}
+}
+
+// NumColumns implements ColumnWriter.
+func (b *PrefixBytesBuilder) NumColumns() int { return 1 }
+
+// Reset resets the builder to an empty state, preserving the existing bundle
+// size.
+func (b *PrefixBytesBuilder) Reset() {
+	*b = PrefixBytesBuilder{
+		data:        b.data[:0],
+		bundleSize:  b.bundleSize,
+		bundleShift: b.bundleShift,
+		offsets:     b.offsets,
+		maxShared:   b.maxShared,
+	}
+	b.offsets.count = 0
+}
+
+// Put adds the provided key to the column. The provided key must be
+// lexicographically greater than or equal to the previous key added to the
+// builder.
+//
+// The provided bytesSharedWithPrev must be the length of the byte prefix the
+// provided key shares with the previous key. The caller is required to provide
+// this because in the primary expected use, the caller will already need to
+// compute it for the purpose of determining whether successive blocks share the
+// same prefix.
+func (b *PrefixBytesBuilder) Put(key []byte, bytesSharedWithPrev int) {
+	if invariants.Enabled {
+		if b.maxShared == 0 {
+			panic(errors.AssertionFailedf("maxShared must be positive"))
+		}
+		if b.nKeys > 0 {
+			if bytes.Compare(key, b.data[len(b.data)-b.lastKeyLen:]) < 0 {
+				panic(errors.AssertionFailedf("keys must be added in order: %q < %q", key, b.data[len(b.data)-b.lastKeyLen:]))
+			}
+			if bytesSharedWithPrev != bytesSharedPrefix(key, b.data[len(b.data)-b.lastKeyLen:]) {
+				panic(errors.AssertionFailedf("bytesSharedWithPrev %d != %d", bytesSharedWithPrev,
+					bytesSharedPrefix(key, b.data[len(b.data)-b.lastKeyLen:])))
+			}
+		}
+	}
+
+	switch {
+	case b.nKeys == 0:
+		// We're adding the first key to the block. Initialize the
+		// block prefix to the length of this key.
+		b.blockPrefixLen = uint32(min(len(key), int(b.maxShared)))
+		b.blockPrefixLenUpdated = 0
+		// Set a placeholder offset for the block prefix length.
+		b.addOffset(0)
+		// Add an offset for the bundle prefix length.
+		b.addOffset(uint32(len(b.data) + min(len(key), int(b.maxShared))))
+		b.nKeys++
+		b.lastLastKeyLen = b.lastKeyLen
+		b.lastKeyLen = len(key)
+		b.currentBundleLen = len(key)
+		b.currentBundlePrefixOffset = 1
+		b.currentBundleKeys = 1
+		b.data = append(b.data, key...)
+		b.addOffset(uint32(len(b.data)))
+	case b.nKeys%b.bundleSize == 0:
+		// We're starting a new bundle so we can compute what the
+		// encoded size of the previous bundle will be.
+		bundlePrefixLen := b.offsets.elems.At(b.currentBundlePrefixOffset) - b.offsets.elems.At(b.currentBundlePrefixOffset-1)
+		b.completedBundleLen += b.currentBundleLen - (b.currentBundleKeys-1)*int(bundlePrefixLen)
+
+		// Update the block prefix length if necessary. The caller tells us how
+		// many bytes of prefix this key shares with the previous key. The block
+		// prefix can only shrink if the bytes shared with the previous key are
+		// less than the block prefix length, in which case the new block prefix
+		// is the number of bytes shared with the previous key.
+		if uint32(bytesSharedWithPrev) < b.blockPrefixLen {
+			b.blockPrefixLen = uint32(bytesSharedWithPrev)
+			b.blockPrefixLenUpdated = b.nKeys
+		}
+
+		// We're adding the first key to the current bundle. Initialize
+		// the bundle prefix to the length of this key.
+		b.currentBundlePrefixOffset = b.offsets.count
+		b.addOffset(uint32(len(b.data) + min(len(key), int(b.maxShared))))
+		b.nKeys++
+		b.lastLastKeyLen = b.lastKeyLen
+		b.lastKeyLen = len(key)
+		b.currentBundleLen = len(key)
+		b.currentBundleKeys = 1
+		b.data = append(b.data, key...)
+		b.addOffset(uint32(len(b.data)))
+	default:
+		// Adding a new key to an existing bundle.
+		// Update the bundle prefix length. Note that the shared prefix length
+		// can only shrink as new values are added. During construction, the
+		// bundle prefix value is stored contiguously in the data array so even
+		// if the bundle prefix length changes no adjustment is needed to that
+		// value or to the first key in the bundle.
+		bundlePrefixLen := b.offsets.elems.At(b.currentBundlePrefixOffset) - b.offsets.elems.At(b.currentBundlePrefixOffset-1)
+		if uint32(bytesSharedWithPrev) < bundlePrefixLen {
+			b.offsets.elems.set(b.currentBundlePrefixOffset, b.offsets.elems.At(b.currentBundlePrefixOffset-1)+uint32(bytesSharedWithPrev))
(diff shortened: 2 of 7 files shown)


RaduBerinde (commented):
This is very cool!!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756#-:-O1e7vBO0efL6cMzq9JP:b-2iwfty)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi and @jbowens)

___
*[`sstable/colblk/prefix_bytes.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1e6y5x9W_JhkkJYe-l:-O1e6y5x9W_JhkkJYe-m:bbt9m6h) ([raw file](https://github.com/cockroachdb/pebble/blob/fa490140b1eee337bcb7d3e062c8a9d863f2ac33/sstable/colblk/prefix_bytes.go#L68)):*
> ```Go
> //	 idx   | row   | offset | data
> //	-------+-------+--------+------
> //	     0 |       |      2 | aa
> ```

Wouldn't it make sense to shift down all the values in this column? Or name it `end-offset`
___
*[`sstable/colblk/prefix_bytes.go` line 240 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1e7JmxBL35m2ImzujT:-O1e7JmxBL35m2ImzujU:b-qounzq) ([raw file](https://github.com/cockroachdb/pebble/blob/fa490140b1eee337bcb7d3e062c8a9d863f2ac33/sstable/colblk/prefix_bytes.go#L240)):*
> ```Go
> 	firstIndex := 1 + (row >> b.bundleShift) + (row & b.bundleMask)
> 	for i > firstIndex {
> 		// Step back a row, and check if the slice is non-empty.
> ```

Not sure how big we expect that bundles will be, but it might make sense to try to jump e.g. 4-8 rows at a time first. Can be a TODO.
___
*[`sstable/colblk/prefix_bytes.go` line 40 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1gmQRoAQZZ8TDNSIGU:-O1gmQRoAQZZ8TDNSIGV:bc0siyl) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L40)):*
> ```Go
> //	 2 aaabbbcde
> //	 3 aaabbbce
> //	 4 aaabbbdee*
> ```

[nit] the `*` suffix here is a bit confusing at first glance. Maybe drop it and just reference the row numbers - "There are 3 keys which occur multiple times (rows 4-6, 9-10, and 12-14)"
___
*[`sstable/colblk/prefix_bytes.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1gnpqf384ZEr6kVcpC:-O1gnpqf384ZEr6kVcpD:behl4e3) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L75)):*
> ```Go
> //	     5 |     3 |     11 | .......e
> //	     6 |       |     15 | ..abbb
> //	     7 |     4 |     19 | ......dee*
> ```

I had a small idea. Say we allowed negative offsets and offset -4 meant that this row is a duplicate of `(row-4)`? It would be an indirection, but we won't need to search back (which involves a trade-off in bundle sizes, and we'd want large bundles if there are a lot of duplicate keys).

In this example we would have:
```
idx=6 bundle end-offset=15 "..abbb"
idx=7 row=6 end-offset=19 "......dee"
idx=8 row=7 end-offset=-1 ""
idx=9 row=8 end-offset=-2 ""
```

For delta encoding purposes, we'd probably want to reserve the lowest bit instead of negative representation, but the principle is the same.

Note also that all row offsets could be relative to the bundle offset (we have to look at the bundle slice in most cases anyway) and that way row offsets would be very small (so no problem if we reserve a bit). So it becomes:

```
idx=6 bundle end-offset=15 "..abbb"
idx=7 row=6 relative-offset=4 "......dee"
idx=8 row=7 relative-offset=-1 ""
idx=9 row=8 relative-offset=-2 ""
```
___
*[`sstable/colblk/prefix_bytes.go` line 148 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1gmssb9u0ycPUA76nR:-O1gmssb9u0ycPUA76nS:b-gqvsbo) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L148)):*
> ```Go
> //
> // Forward iteration can easily reuse the previous row's key with a check on
> // whether the row's slice is empty.  Reverse iteration can reuse the next row's
> ```

[supernit] double space not consistent with the rest of formatting

RaduBerinde:
*[`sstable/colblk/prefix_bytes.go` line 21 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1hDeIKB0NhY9IBH_Ql:-O1hDeIKB0NhY9IBH_Qm:ba5e4rx) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L21)):*
> ```Go
> )
> 
> // PrefixBytes holds an array of lexicograpghically ordered byte slices. It
> ```

[nit] lexicographically

Are they ordered lexicographically or are they ordered according to the (potentially custom) comparator?

I don't think anything here (other than the expected effectiveness of the prefix compression) depends on their ordering, does it? I guess we depend on a < b if a is a prefix of b so we'd need to add that requirement to `Compare`?

IMO we do have the option of going all the way and no longer supporting a custom comparator for the prefix portion (IOW `Compare` must equal `bytes.Compare` when keys have no suffix).

petermattis (commented):
Mostly just a drive-by skim to see what this code is looking like as it is going in.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756#-:-O1mviIU2zMCNdg5gezi:b-2o5khh)* status: 0 of 7 files reviewed, 9 unresolved discussions (waiting on @aadityasondhi, @jbowens, and @RaduBerinde)

___
*[`sstable/colblk/prefix_bytes.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1gnpqf384ZEr6kVcpC:-O1muqvU2ujMq00a3-Z0:b-4xsqjn) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L75)):*

I had a small idea. Say we allowed negative offsets and offset -4 meant that this row is a duplicate of `(row-4)`? It would be an indirection, but we won't need to search back (which involves a trade-off in bundle sizes, and we'd want large bundles if there are a lot of duplicate keys).

In this example we would have:
```
idx=6 bundle end-offset=15 "..abbb"
idx=7 row=6 end-offset=19 "......dee"
idx=8 row=7 end-offset=-1 ""
idx=9 row=8 end-offset=-2 ""
```

For delta encoding purposes, we'd probably want to reserve the lowest bit instead of negative representation, but the principle is the same.

Note also that all row offsets could be relative to the bundle offset (we have to look at the bundle slice in most cases anyway) and that way row offsets would be very small (so no problem if we reserve a bit). So it becomes:

```
idx=6 bundle end-offset=15 "..abbb"
idx=7 row=6 relative-offset=4 "......dee"
idx=8 row=7 relative-offset=-1 ""
idx=9 row=8 relative-offset=-2 ""
```
</blockquote></details>

This is a neat idea. Beyond the extra 1-bit per entry I can't think of any downsides, and that is a really small downside. Possible I'm missing something as I'm no longer fresh on the implementation complexities of this code.
___
*[`sstable/colblk/prefix_bytes.go` line 28 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1msbfIBE2ruRlMQigA:-O1msbfIBE2ruRlMQigB:bo4429u) ([raw file](https://github.com/cockroachdb/pebble/blob/41b9ffa1b7839df7cf146c82cb7d39ef8c3601ff/sstable/colblk/prefix_bytes.go#L28)):*
> ```Go
> // enabling the timestamp to be placed in a separate column, the multiple
> // version problem becomes one of efficiently handling exact duplicate keys.
> // PrefixBytes builds offs of the RawBytes encoding, introducing n/bundleSize+1
> ```

Nit: `s/offs/off/g`
___
*[`sstable/colblk/prefix_bytes.go` line 30 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1mtgyi03IDqzY62fLi:-O1mtgyi03IDqzY62fLj:b-mlvxf8) ([raw file](https://github.com/cockroachdb/pebble/blob/41b9ffa1b7839df7cf146c82cb7d39ef8c3601ff/sstable/colblk/prefix_bytes.go#L30)):*
> ```Go
> // PrefixBytes builds offs of the RawBytes encoding, introducing n/bundleSize+1
> // additional slices for encoding n/bundleSize bundle prefixes and 1
> // column-level prefix.
> ```

Later we describe this as a "block-level shared prefix for the column". I think that is a slightly clear description is it eliminates a possibility of a reader as interpreting this as a prefix across blocks. 
___
*[`sstable/colblk/prefix_bytes.go` line 358 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1muGlvD-AQ3xG0yPpu:-O1muGlvD-AQ3xG0yPpv:beg9m27) ([raw file](https://github.com/cockroachdb/pebble/blob/41b9ffa1b7839df7cf146c82cb7d39ef8c3601ff/sstable/colblk/prefix_bytes.go#L358)):*
> ```Go
> 		hEnd := b.rawBytes.offsets.At(j + h + 2)
> 		// There's a complication with duplicate keys. When keys are repeated,
> 		// the PrefixBytes encoding avoids re-encoding the duplciate key,
> ```

Nit: `s/duplciate/duplicate/g`

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 9 unresolved discussions (waiting on @aadityasondhi and @jbowens)

___
*[`sstable/colblk/prefix_bytes.go` line 21 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1hDeIKB0NhY9IBH_Ql:-O1qraVqBbuiDX7lg-FO:bq4m5i5) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L21)):*

[nit] lexicographically

Are they ordered lexicographically or are they ordered according to the (potentially custom) comparator?

I don't think anything here (other than the expected effectiveness of the prefix compression) depends on their ordering, does it? I guess we depend on a < b if a is a prefix of b so we'd need to add that requirement to `Compare`?

IMO we do have the option of going all the way and no longer supporting a custom comparator for the prefix portion (IOW `Compare` must equal `bytes.Compare` when keys have no suffix).
</blockquote></details>

Hm, if we will use just the bundle prefixes to binary search, we will need to be able to `Compare` arbitrary user key prefixes (as in prefixes of a key prefix), which is not possible currently. Or would need to fully build the first bundle key.

By the way, since the format separates the key prefix and key suffix, do we envision separating them throughout Pebble? E.g. a `type UserKey struct { prefix []byte, suffix uint64 }`  Or would we separate them only at the sstable level?

jbowens (commented):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi, @petermattis, and @RaduBerinde)

___
*[`sstable/colblk/prefix_bytes.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1e6y5x9W_JhkkJYe-l:-O1r-tD17grbHPErnMrb:b624yy6) ([raw file](https://github.com/cockroachdb/pebble/blob/fa490140b1eee337bcb7d3e062c8a9d863f2ac33/sstable/colblk/prefix_bytes.go#L68)):*

Wouldn't it make sense to shift down all the values in this column? Or name it `end-offset`
</blockquote></details>

renamed it end offset
___
*[`sstable/colblk/prefix_bytes.go` line 40 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1gmQRoAQZZ8TDNSIGU:-O1qziNv1RxDhZgMeGxT:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L40)):*

[nit] the `*` suffix here is a bit confusing at first glance. Maybe drop it and just reference the row numbers - "There are 3 keys which occur multiple times (rows 4-6, 9-10, and 12-14)"
</blockquote></details>

Done
___
*[`sstable/colblk/prefix_bytes.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1gnpqf384ZEr6kVcpC:-O1r-x_hEpxJg2Uq-d8U:b-r8tyt3) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L75)):*

This is a neat idea. Beyond the extra 1-bit per entry I can't think of any downsides, and that is a really small downside. Possible I'm missing something as I'm no longer fresh on the implementation complexities of this code.
</blockquote></details>

Nice, I'll give this a try today. @petermattis I think you had a similar thought when you were working on it: https://github.com/cockroachdb/pebble/compare/master...petermattis:pebble:pmattis/bytes-builder-v2#diff-fd1ca62f03cdc7da64497d1c6f1bf8eae794554368d1599b563b4210e5c9aea9R63-R67 but I haven't attempted it yet.
___
*[`sstable/colblk/prefix_bytes.go` line 30 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1mtgyi03IDqzY62fLi:-O1r-MNiDp-2QHCOMnBb:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/41b9ffa1b7839df7cf146c82cb7d39ef8c3601ff/sstable/colblk/prefix_bytes.go#L30)):*

Later we describe this as a "block-level shared prefix for the column". I think that is a slightly clear description is it eliminates a possibility of a reader as interpreting this as a prefix across blocks.
</blockquote></details>

Done.

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi, @jbowens, and @RaduBerinde)

___
*[`sstable/colblk/prefix_bytes.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1gnpqf384ZEr6kVcpC:-O1r38mN5PyjE7uEcXUG:b-gllw9e) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L75)):*

Nice, I'll give this a try today. @petermattis I think you had a similar thought when you were working on it: https://github.com/cockroachdb/pebble/compare/master...petermattis:pebble:pmattis/bytes-builder-v2#diff-fd1ca62f03cdc7da64497d1c6f1bf8eae794554368d1599b563b4210e5c9aea9R63-R67 but I haven't attempted it yet.
</blockquote></details>

Oh, ha! I knew this idea sounded familiar. :smile:

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi, @petermattis, and @RaduBerinde)

___
*[`sstable/colblk/prefix_bytes.go` line 21 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1hDeIKB0NhY9IBH_Ql:-O1r2l3xDcQYQyghRwtO:btnra28) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L21)):*
> I guess we depend on a < b if a is a prefix of b so we'd need to add that requirement to `Compare`?

I think this comes close to equivalence between `bytes.Compare` and the user-defined `Compare`, so we mind as well go all the way

> Or would we separate them only at the sstable level?

I think only at the sstable level for the near-term future, but I was pondering the same long term. I don't know how feasible it would be, but it does seem like a shame to re-compose and de-compose the full keys unnecessarily

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi, @petermattis, and @RaduBerinde)

___
*[`sstable/colblk/prefix_bytes.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1gnpqf384ZEr6kVcpC:-O1rSMUL8oaNGVzkW6pJ:bavs11o) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L75)):*

Oh, ha! I knew this idea sounded familiar. :smile:
</blockquote></details>

There's a bit of a wrinkle in that when retrieving a bundle prefix, we know its start offset within the data section by looking at the preceding offset. If the preceding offset is a duplicated key, we would now need to make the extra hop to follow the duplicated key's relative index to find the start offset. I'm going to leave a TODO for now, and we can revisit.

RaduBerinde:
*[`sstable/colblk/prefix_bytes.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1gnpqf384ZEr6kVcpC:-O1rUhnkEi3KlBV7649j:b1d3jbk) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L75)):*

There's a bit of a wrinkle in that when retrieving a bundle prefix, we know its start offset within the data section by looking at the preceding offset. If the preceding offset is a duplicated key, we would now need to make the extra hop to follow the duplicated key's relative index to find the start offset. I'm going to leave a TODO for now, and we can revisit.
</blockquote></details>

Oh, interesting. We could store two offsets for the bundle prefix.

RaduBerinde:
*[`sstable/colblk/prefix_bytes.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1gnpqf384ZEr6kVcpC:-O1rYLp2-cQ2LuoHLVEc:b-v5pdyk) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L75)):*

Oh, interesting. We could store two offsets for the bundle prefix.
</blockquote></details>

We could even separate out the block prefixes completely (the offsets AND the data). That will give us much better cache locality when searching for a bundle.

So we would store the prefixes (with two offsets per prefix), then separately the rows (with one offset per row).

Sorry for bikeshedding but I think this is the time :)

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi, @petermattis, and @RaduBerinde)

___
*[`sstable/colblk/prefix_bytes.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1gnpqf384ZEr6kVcpC:-O1rjlRpCmWpxTXIFXu4:bwi25uw) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L75)):*
> We could even separate out the block prefixes completely (the offsets AND the data).

Separating the bundle prefix offsets is interesting; it would remove the somewhat annoying/complicated indexing scheme.

I think we wouldn't want to separate out the data because we take advantage of the fact the prefix and the first key's suffix is stored contiguously to binary search among the first keys of each bundle. We wouldn't want to just binary search among the bundle prefixes because they can be very non-discriminating (eg, you might have 10 bundles with the same prefix, but your seek key falls into the final bundle). I guess you're limited to at most 256/BundleSize bundles with the  exactsame prefix if your keys are dense, like UUID primary key.

> Sorry for bikeshedding but I think this is the time :)

No worries at all! I think it's great to iterate on the format and lay out all the possibilities. Some of them (like the offset bit) have tradeoffs that depend on the data distribution, and I think might be better suited for evaluating later, once we can run an end-to-end representative cockroach workload.

With the offset bit, I'm not sure whether many versions of the same key are common enough to warrant trying to bound the number of offsets we look back beyond the existing bound of the bundle size, especially if it comes at any cost to ordinary lookups.

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi, @jbowens, and @RaduBerinde)

___
*[`sstable/colblk/prefix_bytes.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1gnpqf384ZEr6kVcpC:-O1ruSD_3Cjr8obaHmVh:b-k032ca) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L75)):*
This is the right time to bikeshed in my opinion, though we don't have to get everything perfect in this PR. 

> Separating the bundle prefix offsets is interesting; it would remove the somewhat annoying/complicated indexing scheme.

The existing indexing is complicated, though with separate bundle prefix offsets would we be moving the complication to figuring out the start offset for the first key in the bundle? Perhaps there is something clever to be done there. Seems worthwhile to sketch out what this would look like.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi, @jbowens, and @petermattis)

___
*[`sstable/colblk/prefix_bytes.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1gnpqf384ZEr6kVcpC:-O1ryh_i37ngUTiur9oY:bh7cz45) ([raw file](https://github.com/cockroachdb/pebble/blob/7662b290e94452f307ebab40e3f3eabf6918bce9/sstable/colblk/prefix_bytes.go#L75)):*

This is the right time to bikeshed in my opinion, though we don't have to get everything perfect in this PR. 

> Separating the bundle prefix offsets is interesting; it would remove the somewhat annoying/complicated indexing scheme.

The existing indexing is complicated, though with separate bundle prefix offsets would we be moving the complication to figuring out the start offset for the first key in the bundle? Perhaps there is something clever to be done there. Seems worthwhile to sketch out what this would look like.
</blockquote></details>

Right, the first key in a bundle would need to be special cased to use the bundle end prefix. Yeah I agree, this is best left for later experimentation since there is no clear big win.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 8 unresolved discussions (waiting on @aadityasondhi, @jbowens, and @petermattis)

___
*[`sstable/colblk/prefix_bytes.go` line 179 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1s5O9y2NfRfLOkhEP6:-O1s5O9y2NfRfLOkhEP7:b-g3wyfu) ([raw file](https://github.com/cockroachdb/pebble/blob/941d9d2c887b58ad385c5403bd397f755c35bbb8/sstable/colblk/prefix_bytes.go#L179)):*
> ```Go
> // number of logical slices within the array.
> func MakePrefixBytes(count int, b []byte, offset uint32, enc ColumnEncoding) PrefixBytes {
> 	// The first byte of a PrefixBytes-encoded column is the bundle size
> ```

[nit] an invariants check that count > 0 might be useful
___
*[`sstable/colblk/prefix_bytes.go` line 207 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1s5ef8D9x63EApfMW_:-O1s5ef8D9x63EApfMWa:b-b1t8n5) ([raw file](https://github.com/cockroachdb/pebble/blob/941d9d2c887b58ad385c5403bd397f755c35bbb8/sstable/colblk/prefix_bytes.go#L207)):*
> ```Go
> }
> 
> // RowBundlePrefix takes a row index and returns a []byte of the prefix shared
> ```

This should make it clear that it does not include the shared prefix (it reads the other way currently).
___
*[`sstable/colblk/prefix_bytes.go` line 231 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1s6CGO4eGm3YEwK_L8:-O1s6CGO4eGm3YEwK_L9:bo7mfjr) ([raw file](https://github.com/cockroachdb/pebble/blob/941d9d2c887b58ad385c5403bd397f755c35bbb8/sstable/colblk/prefix_bytes.go#L231)):*
> ```Go
> // The returned slice should not be mutated.
> func (b PrefixBytes) RowSuffix(row int) []byte {
> 	i := 1 + (row >> b.bundleShift) + row
> ```

[nit] it would be nice to have some accessors instead of inlining this math in multiple places. E.g. `bundleIdxForRow()` `offsetIdxForBundle`, `firstRowIdxForBundle`, `lastRowIdxForBundle`
___
*[`sstable/colblk/prefix_bytes.go` line 273 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1s6dbe9-DB_hsxJOlM:-O1s6dbe9-DB_hsxJOlN:bjn8qqs) ([raw file](https://github.com/cockroachdb/pebble/blob/941d9d2c887b58ad385c5403bd397f755c35bbb8/sstable/colblk/prefix_bytes.go#L273)):*
> ```Go
> // equal to k, returning the index of the key and whether an equal key was
> // found.
> func (b PrefixBytes) Search(k []byte) (int, bool) {
> ```

[nit] name the returns `(rowIndex int, isEqual bool)`. Also mention that we return `Rows()` if all rows are < k. And maybe make it explicit that if we have equal keys, we return the first one of those.
___
*[`sstable/colblk/prefix_bytes.go` line 334 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1sBEYK8jOFKbyPRgG-:-O1sBEYK8jOFKbyPRgG0:b9btn7r) ([raw file](https://github.com/cockroachdb/pebble/blob/941d9d2c887b58ad385c5403bd397f755c35bbb8/sstable/colblk/prefix_bytes.go#L334)):*
> ```Go
> 	bundlePrefix := b.rawBytes.slice(b.rawBytes.offsets.At(j), b.rawBytes.offsets.At(j+1))
> 	if len(bundlePrefix) > len(k) || !bytes.Equal(k[:len(bundlePrefix)], bundlePrefix) {
> 		// The search key doesn't share the previous bundle's prefix, so all of
> ```

I had to think about this for a bit. I think it would be more clear if we said "The row we are looking for could be in the previous bundle only iff the seek key is equal to a row or is in-between two rows from the previous bundle; in both cases the seek key must have the bundle prefix`
___
*[`sstable/colblk/prefix_bytes.go` line 337 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1s9Zcb0ugMMby1weKZ:-O1s9Zcb0ugMMby1weK_:b-7follj) ([raw file](https://github.com/cockroachdb/pebble/blob/941d9d2c887b58ad385c5403bd397f755c35bbb8/sstable/colblk/prefix_bytes.go#L337)):*
> ```Go
> 		// the keys in the previous bundle must be less than k. We know the
> 		// first key of bi is ≥ k, so return it.
> 		if bi<<b.bundleShift < b.rows {
> ```

[nit] can this just be if `bi < nBundles`? (also, I'd reverse the check so the more special case has a deeper indentation

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @aadityasondhi, @petermattis, and @RaduBerinde)

___
*[`sstable/colblk/prefix_bytes.go` line 207 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1s5ef8D9x63EApfMW_:-O1w7jge8SLDUEc7Gjvr:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/941d9d2c887b58ad385c5403bd397f755c35bbb8/sstable/colblk/prefix_bytes.go#L207)):*

This should make it clear that it does not include the shared prefix (it reads the other way currently).
</blockquote></details>

Done.
___
*[`sstable/colblk/prefix_bytes.go` line 231 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1s6CGO4eGm3YEwK_L8:-O1wDVt7DTgXi0znOJe_:bm1eqww) ([raw file](https://github.com/cockroachdb/pebble/blob/941d9d2c887b58ad385c5403bd397f755c35bbb8/sstable/colblk/prefix_bytes.go#L231)):*

[nit] it would be nice to have some accessors instead of inlining this math in multiple places. E.g. `bundleIdxForRow()` `offsetIdxForBundle`, `firstRowIdxForBundle`, `lastRowIdxForBundle`
</blockquote></details>

Extracted some of these indexing routines
___
*[`sstable/colblk/prefix_bytes.go` line 273 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1s6dbe9-DB_hsxJOlM:-O1wGmQXBo2Ks1urDgZU:b-95txdl) ([raw file](https://github.com/cockroachdb/pebble/blob/941d9d2c887b58ad385c5403bd397f755c35bbb8/sstable/colblk/prefix_bytes.go#L273)):*

[nit] name the returns `(rowIndex int, isEqual bool)`. Also mention that we return `Rows()` if all rows are < k. And maybe make it explicit that if we have equal keys, we return the first one of those.
</blockquote></details>


Done
___
*[`sstable/colblk/prefix_bytes.go` line 334 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1sBEYK8jOFKbyPRgG-:-O1wHxV8-uJVg7pkdhvl:b6yexf8) ([raw file](https://github.com/cockroachdb/pebble/blob/941d9d2c887b58ad385c5403bd397f755c35bbb8/sstable/colblk/prefix_bytes.go#L334)):*

I had to think about this for a bit. I think it would be more clear if we said "The row we are looking for could be in the previous bundle only iff the seek key is equal to a row or is in-between two rows from the previous bundle; in both cases the seek key must have the bundle prefix`
</blockquote></details>

Did a bit of rewording
___
*[`sstable/colblk/prefix_bytes.go` line 337 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1s9Zcb0ugMMby1weKZ:-O1wItr40rxM20AXgTk_:bp2fioi) ([raw file](https://github.com/cockroachdb/pebble/blob/941d9d2c887b58ad385c5403bd397f755c35bbb8/sstable/colblk/prefix_bytes.go#L337)):*

[nit] can this just be if `bi < nBundles`? (also, I'd reverse the check so the more special case has a deeper indentation
</blockquote></details>

Good call, done

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @aadityasondhi, @petermattis, and @RaduBerinde)

___
*[`sstable/colblk/prefix_bytes.go` line 231 at r8](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1s6CGO4eGm3YEwK_L8:-O1yCk95Ch6tUSWNI6px:bzgtafr) ([raw file](https://github.com/cockroachdb/pebble/blob/941d9d2c887b58ad385c5403bd397f755c35bbb8/sstable/colblk/prefix_bytes.go#L231)):*

They can be methods on `b` so you don't have to pass `bundleShift`,`bundleMask`
</blockquote></details>

Some of these are used by the builder as well.
___
*[`sstable/colblk/prefix_bytes.go` line 477 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1xjMLP311GRqk53Dnx:-O1yDbpc5OB43If_hZsA:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/09777de103cba6555d9c3e5f73fb6a06bfaecc45/sstable/colblk/prefix_bytes.go#L477)):*

[nit] Maybe add a limit for `len(data)` in `Reset()` so at least we don't hold on to a huge slice for a long time
</blockquote></details>

Done
___
*[`sstable/colblk/prefix_bytes.go` line 517 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1xPIN0AiQRHj7anmqO:-O1yCqPU-xqsnTfQdV2n:byf82ux) ([raw file](https://github.com/cockroachdb/pebble/blob/09777de103cba6555d9c3e5f73fb6a06bfaecc45/sstable/colblk/prefix_bytes.go#L517)):*

Why are we tolerating `bundleSize==0`?
</blockquote></details>

Thanks; this is a vestige from the prototype where `RawBytes` and `PrefixBytes` shared a builder. Removed.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 3 unresolved discussions (waiting on @aadityasondhi, @jbowens, and @petermattis)

___
*[`sstable/colblk/prefix_bytes.go` line 720 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1y66r_12TdfU9dSp5c:-O20Bsha0llG47EgpYab:bjwn3oc) ([raw file](https://github.com/cockroachdb/pebble/blob/09777de103cba6555d9c3e5f73fb6a06bfaecc45/sstable/colblk/prefix_bytes.go#L720)):*

There is so much complication with `b.nKeys-1`.. We could `computePrefixCompressionSizing` inside `Put()` (we expect a corresponding call to `Size()` anyway) and keep the last two values, then use whichever one applies.

The same could apply to more fields inside the `PrefixBytesBuilder`, we could have something like

```
type PrefixBytesBuilder struct {
  ...
  state, stateBeforeLastKey struct {
    currentBundlePrefixLen int
    currentBundleDistrinctLen int
    currentBundleDistinctKeys int
    blockPrefixLen int
    sizing prefixCompressionSizing
  }
```

We would do `b.stateBeforeLastRow = b.state` whenever we add a row which should be very fast.
</blockquote></details>

Actually, if we had those two structs we can make `computePrefixCompressionSizing` use either struct and we don't need to store the results. That would enable some higher-level optimization where we don't call `Size()` every time (e.g. we could have a simple upper bound that we check first).

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 10 unresolved discussions (waiting on @aadityasondhi, @jbowens, and @petermattis)

___
*[`sstable/colblk/prefix_bytes.go` line 868 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20EWJo8JyCRENhEbag:-O20EWJo8JyCRENhEbah:br03kmh) ([raw file](https://github.com/cockroachdb/pebble/blob/d4f446fd600ddf2b8d3f416ffec501319de8328f/sstable/colblk/prefix_bytes.go#L868)):*
> ```Go
> 		off := b.offsets.elems.At(i)
> 		var suffix []byte
> 		if (i-1)%(b.bundleSize+1) == 0 {
> ```

[nit] It would be cleaner (and maybe even a bit more efficient) to do something like
```
for bundleIdx := 0; bundleIdx < .. ; bundleIdx++ {
  <set bundle prefix offset>
  
  for rowIdx := 0; rowIdx < b.bundleIdx; rowIdx++ {
    <set row offset>
  }
}
```
___
*[`sstable/colblk/prefix_bytes.go` line 909 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20CzPn7Z1JvudygrUw:-O20CzPn7Z1JvudygrUx:bv75taz) ([raw file](https://github.com/cockroachdb/pebble/blob/d4f446fd600ddf2b8d3f416ffec501319de8328f/sstable/colblk/prefix_bytes.go#L909)):*
> ```Go
> func (b *PrefixBytesBuilder) Finish(
> 	col int, rows int, offset uint32, buf []byte,
> ) (uint32, ColumnDesc) {
> ```

[nit] `endOffset uint32`
___
*[`sstable/colblk/prefix_bytes.go` line 916 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20E3RhBuBfo92dBVDF:-O20E3RhBuBfo92dBVDG:bk32z0v) ([raw file](https://github.com/cockroachdb/pebble/blob/d4f446fd600ddf2b8d3f416ffec501319de8328f/sstable/colblk/prefix_bytes.go#L916)):*
> ```Go
> 	desc := ColumnDesc{DataType: DataTypePrefixBytes}
> 	if rows == 0 {
> 		return offset, desc
> ```

Will the rows==0 case actually show up in practice or should we disallow it?
___
*[`sstable/colblk/prefix_bytes.go` line 967 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20CtLR-_15SRnrXNMt:-O20CtLR-_15SRnrXNMu:b-1y4gk8) ([raw file](https://github.com/cockroachdb/pebble/blob/d4f446fd600ddf2b8d3f416ffec501319de8328f/sstable/colblk/prefix_bytes.go#L967)):*
> ```Go
> // provided offset. The offset is required to ensure proper alignment. The
> // returned uint32 is the offset of the first byte after the end of the encoded
> // data. To compute the size in bytes, subtract the [offset] passed into Size
> ```

These semantics are unexpected, why not return the size in bytes, which is what the method signature strongly suggests? This applies to `uintColumnSize` as well..  If we keep it this way, we should rename the method or at least name the return `endOffset`.
___
*[`sstable/colblk/prefix_bytes_test.go` line 110 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20Cf9l7kESos7IqUUc:-O20Cf9mC0NZ28u0IkXm:bo1thck) ([raw file](https://github.com/cockroachdb/pebble/blob/d4f446fd600ddf2b8d3f416ffec501319de8328f/sstable/colblk/prefix_bytes_test.go#L110)):*
> ```Go
> 	runTest := func(t *testing.T, maxKeyCount, maxKeyLen int) {
> 		rng := rand.New(rand.NewSource(seed))
> 		randInt := func(lo, hi int) int {
> ```

[nit] it would be simpler if `hi` was inclusive
___
*[`sstable/colblk/prefix_bytes_test.go` line 124 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20FLjC0vX3oD-GKn-M:-O20FLjC0vX3oD-GKn-N:b-g2hbvl) ([raw file](https://github.com/cockroachdb/pebble/blob/d4f446fd600ddf2b8d3f416ffec501319de8328f/sstable/colblk/prefix_bytes_test.go#L124)):*
> ```Go
> 		totalSize := len(userKeys[0])
> 		for j := range userKeys[0] {
> 			userKeys[0][j] = byte(randInt(int('a'), int('z')+1))
> ```

[nit] I think the range of characters should be passed to `runTest` and have multiple variants. We can exercise a lot more cornercases with just two possible characters.
___
*[`sstable/colblk/prefix_bytes_test.go` line 230 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20FdkE6VFij-aVvZfz:-O20FdkE6VFij-aVvZg-:b-lem1re) ([raw file](https://github.com/cockroachdb/pebble/blob/d4f446fd600ddf2b8d3f416ffec501319de8328f/sstable/colblk/prefix_bytes_test.go#L230)):*
> ```Go
> 		copy(userKeys[i], userKeys[0][:blockPrefixLen])
> 		for j := blockPrefixLen; j < len(userKeys[i]); j++ {
> 			userKeys[i][j] = byte(randInt(int('a'), int('z')+1))
> ```

Same here, I think using the entire alphabet will have very small prefixes, it would be worth having a version with a small alphabet.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 10 unresolved discussions (waiting on @aadityasondhi, @petermattis, and @RaduBerinde)

___
*[`sstable/colblk/prefix_bytes.go` line 868 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20EWJo8JyCRENhEbag:-O20bXdP8XBQJf_eFNDE:b6772h9) ([raw file](https://github.com/cockroachdb/pebble/blob/d4f446fd600ddf2b8d3f416ffec501319de8328f/sstable/colblk/prefix_bytes.go#L868)):*

[nit] It would be cleaner (and maybe even a bit more efficient) to do something like
```
for bundleIdx := 0; bundleIdx < .. ; bundleIdx++ {
  <set bundle prefix offset>
  
  for rowIdx := 0; rowIdx < b.bundleIdx; rowIdx++ {
    <set row offset>
  }
}
```
</blockquote></details>

I actually already tried that and found it wasn't cleaner nor more efficient.
___
*[`sstable/colblk/prefix_bytes.go` line 916 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20E3RhBuBfo92dBVDF:-O20bjyI-PuNPAWQS3-d:b-feupdv) ([raw file](https://github.com/cockroachdb/pebble/blob/d4f446fd600ddf2b8d3f416ffec501319de8328f/sstable/colblk/prefix_bytes.go#L916)):*

Will the rows==0 case actually show up in practice or should we disallow it?
</blockquote></details>

Yes, it will.
___
*[`sstable/colblk/prefix_bytes.go` line 967 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20CtLR-_15SRnrXNMt:-O20bmqO4cvkpq79KMPd:b-ibane3) ([raw file](https://github.com/cockroachdb/pebble/blob/d4f446fd600ddf2b8d3f416ffec501319de8328f/sstable/colblk/prefix_bytes.go#L967)):*

These semantics are unexpected, why not return the size in bytes, which is what the method signature strongly suggests? This applies to `uintColumnSize` as well..  If we keep it this way, we should rename the method or at least name the return `endOffset`.
</blockquote></details>

The usage will always want the semantics that are implemented.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756)* status: 0 of 7 files reviewed, 9 unresolved discussions (waiting on @aadityasondhi, @petermattis, and @RaduBerinde)

___
*[`sstable/colblk/prefix_bytes.go` line 720 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1y66r_12TdfU9dSp5c:-O20dZvwCJXY026a59aJ:bu38zu9) ([raw file](https://github.com/cockroachdb/pebble/blob/09777de103cba6555d9c3e5f73fb6a06bfaecc45/sstable/colblk/prefix_bytes.go#L720)):*

Actually, if we had those two structs we can make `computePrefixCompressionSizing` use either struct and we don't need to store the results. That would enable some higher-level optimization where we don't call `Size()` every time (e.g. we could have a simple upper bound that we check first).
</blockquote></details>

I implemented something similar here, although I didn't follow your second comment.
___
*[`sstable/colblk/prefix_bytes_test.go` line 110 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20Cf9l7kESos7IqUUc:-O20doMGA438iul3m9w7:bl7v2wf) ([raw file](https://github.com/cockroachdb/pebble/blob/d4f446fd600ddf2b8d3f416ffec501319de8328f/sstable/colblk/prefix_bytes_test.go#L110)):*

[nit] it would be simpler if `hi` was inclusive
</blockquote></details>

I don't think it would be? In some instances we pass existing exclusive values.
___
*[`sstable/colblk/prefix_bytes_test.go` line 124 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20FLjC0vX3oD-GKn-M:-O20e1vN08l31JQ625Ux:b-vyz8je) ([raw file](https://github.com/cockroachdb/pebble/blob/d4f446fd600ddf2b8d3f416ffec501319de8328f/sstable/colblk/prefix_bytes_test.go#L124)):*

[nit] I think the range of characters should be passed to `runTest` and have multiple variants. We can exercise a lot more cornercases with just two possible characters.
</blockquote></details>

I think most of the corner cases, especially in Search are with small prefixes, but noted and done.

RaduBerinde (approved):
:lgtm_strong: 

Really great work! I can tell there were a thousand small decisions that went into it, some of which tend to keep one up at night :)

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756#-:-O20tRfc3XjTD5tClY8B:bjdmwrl)* status: 0 of 7 files reviewed, 12 unresolved discussions (waiting on @aadityasondhi, @jbowens, and @petermattis)

___
*[`sstable/colblk/prefix_bytes.go` line 720 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O1y66r_12TdfU9dSp5c:-O20qj3l8YGtYfG1Iwl9:b7zaan8) ([raw file](https://github.com/cockroachdb/pebble/blob/09777de103cba6555d9c3e5f73fb6a06bfaecc45/sstable/colblk/prefix_bytes.go#L720)):*

I implemented something similar here, although I didn't follow your second comment.
</blockquote></details>

This looks great :thumbsup:
___
*[`sstable/colblk/prefix_bytes.go` line 169 at r11](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20uB0Y6gdL5ZpZCzor:-O20uB0Y6gdL5ZpZCzos:b2zp9a7) ([raw file](https://github.com/cockroachdb/pebble/blob/f75cda9610a3de5918f4fd4cef659181e702b23d/sstable/colblk/prefix_bytes.go#L169)):*
> ```Go
> type PrefixBytes struct {
> 	rows            int
> 	bundleShift     int
> ```

[nit] IMO `bundleShift/bundleMask` belong in a separate struct which can be embedded (both here and in the builder0, which has methods for all calculations that depend on `bundleShift` (and perhaps uses uint32 internally). Maybe leave a TODO.
___
*[`sstable/colblk/prefix_bytes.go` line 170 at r11](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20tvm_8xe-V38yc1Iw:-O20tvm_8xe-V38yc1Ix:bvesboz) ([raw file](https://github.com/cockroachdb/pebble/blob/f75cda9610a3de5918f4fd4cef659181e702b23d/sstable/colblk/prefix_bytes.go#L170)):*
> ```Go
> 	rows            int
> 	bundleShift     int
> 	bundleMask      int
> ```

[nit] this deserves a comment, one would assume it's `(1 << bundleShift) - 1` not the inverse
___
*[`sstable/colblk/prefix_bytes.go` line 453 at r11](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20uLq1BOJvDagPq1h3:-O20uLq3CTkCrzvkK_R-:b-qnq2do) ([raw file](https://github.com/cockroachdb/pebble/blob/f75cda9610a3de5918f4fd4cef659181e702b23d/sstable/colblk/prefix_bytes.go#L453)):*
> ```Go
> 	for i := 0; i < k-1; i++ {
> 		endOff := pb.rawBytes.offsets.At(i + 1)
> 		if i%(1+(1<<pb.bundleShift)) == 0 {
> ```

`i & (1<<pb.bundleShift-1)`
___
*[`sstable/colblk/prefix_bytes.go` line 587 at r11](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20uR2u5klw-c8Ad4c9:-O20uR2u5klw-c8Ad4cA:bnv3otg) ([raw file](https://github.com/cockroachdb/pebble/blob/f75cda9610a3de5918f4fd4cef659181e702b23d/sstable/colblk/prefix_bytes.go#L587)):*
> ```Go
> // same prefix.
> func (b *PrefixBytesBuilder) Put(key []byte, bytesSharedWithPrev int) {
> 	currIdx := b.nKeys % 2
> ```

[nit] `& 1` (since we established it's faster with signed ints)
___
*[`sstable/colblk/prefix_bytes.go` line 628 at r11](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20uUm60HLwAh7H_lZS:-O20uUm60HLwAh7H_lZT:bubjmk0) ([raw file](https://github.com/cockroachdb/pebble/blob/f75cda9610a3de5918f4fd4cef659181e702b23d/sstable/colblk/prefix_bytes.go#L628)):*
> ```Go
> 			offsetDeltaWidth:          max(1, deltaWidth(uint64(len(key)))),
> 		}
> 	case b.nKeys%b.bundleSize == 0:
> ```

`nKeys & (1 << b.bundleSize - 1) == 0`
___
*[`sstable/colblk/prefix_bytes.go` line 709 at r11](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20uevZ8Bls185xg08i:-O20uevZ8Bls185xg08j:b-zgv45a) ([raw file](https://github.com/cockroachdb/pebble/blob/f75cda9610a3de5918f4fd4cef659181e702b23d/sstable/colblk/prefix_bytes.go#L709)):*
> ```Go
> // PrevKey returns the previous key added to the builder through Put. The key is
> // guaranteed to be stable until Finish or Reset is called.
> func (b *PrefixBytesBuilder) PrevKey() []byte {
> ```

[nit] would `LastKey()` be more accurate?
___
*[`sstable/colblk/prefix_bytes.go` line 710 at r11](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20u_7W6ztGVNq2qlvG:-O20u_7W6ztGVNq2qlvH:btoc68r) ([raw file](https://github.com/cockroachdb/pebble/blob/f75cda9610a3de5918f4fd4cef659181e702b23d/sstable/colblk/prefix_bytes.go#L710)):*
> ```Go
> // guaranteed to be stable until Finish or Reset is called.
> func (b *PrefixBytesBuilder) PrevKey() []byte {
> 	return b.data[len(b.data)-b.sizings[(b.nKeys+1)%2].lastKeyLen:]
> ```

[nit] `&1`
___
*[`sstable/colblk/prefix_bytes.go` line 936 at r11](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20shNj4w_PpP8_-Sz0:-O20shNj4w_PpP8_-Sz1:b-7ew3xn) ([raw file](https://github.com/cockroachdb/pebble/blob/f75cda9610a3de5918f4fd4cef659181e702b23d/sstable/colblk/prefix_bytes.go#L936)):*
> ```Go
> 	var shared int
> 	n := min(len(a), len(b))
> 	for shared < n-7 && asUint64(a, shared) == asUint64(b, shared) {
> ```

[nit] can just use `bytes.Equal(a[shared:shared+8], b[shared:shared+8])` (or 16). BTW we no longer use this outside of tests or invariants so it's not performance critical.

jbowens (commented):
TFTR!

> some of which tend to keep one up at night :)

is it so obvious

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756#-:-O21cUCO374aXqR1M1Nt:b-46hj5w)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @aadityasondhi, @petermattis, and @RaduBerinde)

___
*[`sstable/colblk/prefix_bytes.go` line 169 at r11](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20uB0Y6gdL5ZpZCzor:-O21Yvor2Wle0uNY3-Si:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/f75cda9610a3de5918f4fd4cef659181e702b23d/sstable/colblk/prefix_bytes.go#L169)):*

[nit] IMO `bundleShift/bundleMask` belong in a separate struct which can be embedded (both here and in the builder0, which has methods for all calculations that depend on `bundleShift` (and perhaps uses uint32 internally). Maybe leave a TODO.
</blockquote></details>

Done
___
*[`sstable/colblk/prefix_bytes.go` line 453 at r11](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20uLq1BOJvDagPq1h3:-O21ZTJK8EoXG-CmEylh:b-3lfwjj) ([raw file](https://github.com/cockroachdb/pebble/blob/f75cda9610a3de5918f4fd4cef659181e702b23d/sstable/colblk/prefix_bytes.go#L453)):*

`i & (1<<pb.bundleShift-1)`
</blockquote></details>

off by one I think... (also this is a formatting debugging routine so the perf doesn't matter)
___
*[`sstable/colblk/prefix_bytes.go` line 587 at r11](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20uR2u5klw-c8Ad4c9:-O21_bwC9uqpN0IkSUEB:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/f75cda9610a3de5918f4fd4cef659181e702b23d/sstable/colblk/prefix_bytes.go#L587)):*

[nit] `& 1` (since we established it's faster with signed ints)
</blockquote></details>

Done
___
*[`sstable/colblk/prefix_bytes.go` line 628 at r11](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20uUm60HLwAh7H_lZS:-O21bh_E1XwAILXHf_Ki:b8uuyi2) ([raw file](https://github.com/cockroachdb/pebble/blob/f75cda9610a3de5918f4fd4cef659181e702b23d/sstable/colblk/prefix_bytes.go#L628)):*

`nKeys & (1 << b.bundleSize - 1) == 0`
</blockquote></details>

Done (using `b.bundleSize` instead `1 << b.bundleShift`)
___
*[`sstable/colblk/prefix_bytes.go` line 936 at r11](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O20shNj4w_PpP8_-Sz0:-O21cAXz10Z0FTtl2J-V:bsnsyzy) ([raw file](https://github.com/cockroachdb/pebble/blob/f75cda9610a3de5918f4fd4cef659181e702b23d/sstable/colblk/prefix_bytes.go#L936)):*

[nit] can just use `bytes.Equal(a[shared:shared+8], b[shared:shared+8])` (or 16). BTW we no longer use this outside of tests or invariants so it's not performance critical.
</blockquote></details>

It will be used outside of tests in a future PR

sumeerbhola (commented):
a few drive-by questions based on reading the comments.

> Reviewed 1 of 8 files at r12, 1 of 3 files at r13.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756#-:-O28GnJ_6qPdf4BjlH_T:b-qw8gsw)* status: 2 of 10 files reviewed, 9 unresolved discussions

___
*[`sstable/colblk/column.go` line 76 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O288-ZP3OVuisYBWIPw:-O288-ZP3OVuisYBWIPx:b-tlfqpv) ([raw file](https://github.com/cockroachdb/pebble/blob/453aa1879b928efcb5c73651265dc8e4888771c0/sstable/colblk/column.go#L76)):*
> ```Go
> 	// data to buf at offset, and returning the offset at which the next column
> 	// should be encoded. Finish also returns a column descriptor describing the
> 	// encoding of the column, which will be serialized within the block header.
> ```

nit: Finish also returns ... seems stale
___
*[`sstable/colblk/prefix_bytes.go` line 89 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O289HbF-_Q4UPoSENA0:-O289HbF-_Q4UPoSENA1:b-l79e6i) ([raw file](https://github.com/cockroachdb/pebble/blob/453aa1879b928efcb5c73651265dc8e4888771c0/sstable/colblk/prefix_bytes.go#L89)):*
> ```Go
> //	    19 |    14 |         36 | ........
> //
> // The offset column in the table points to the start and end index within the
> ```

is there really a start index, given the table above says "end offset"? Is "end offset" the same as "end index"?
If "end offset" is 2 for aa, then the offset is the end index + 1, yes?

IIRC, RawBytes was encoding the start offsets. By saying that "Offset[0] is the length of the first slice", is this encoding all the end offsets instead?
___
*[`sstable/colblk/prefix_bytes.go` line 119 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O28Alj11aYi1iqrGv2R:-O28Alj11aYi1iqrGv2S:b65d2fg) ([raw file](https://github.com/cockroachdb/pebble/blob/453aa1879b928efcb5c73651265dc8e4888771c0/sstable/colblk/prefix_bytes.go#L119)):*
> ```Go
> //	|                                                                  |
> //	| A modified RawBytes encoding is used to store the data slices. A |
> //	| PrefixBytes column storing n keys will encode 2+n+n/bundleSize   |
> ```

why is this not 1 + n + (n + bundleSize-1)/n?
___
*[`sstable/colblk/prefix_bytes.go` line 151 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O28ECTb91qJm_N1U4ti:-O28ECTb91qJm_N1U4tj:b-z2jy9z) ([raw file](https://github.com/cockroachdb/pebble/blob/453aa1879b928efcb5c73651265dc8e4888771c0/sstable/colblk/prefix_bytes.go#L151)):*
> ```Go
> // This encoding provides O(1) access to any row by calculating the bundle for
> // the row (5*(row/4)), then the row's index within the bundle (1+(row%4)). If
> // the slice's offset equals the previous slice's offset then we step backward
> ```

The offset of the bundle prefix could also be equal to the prefix of the previous slice, yes? In which case we don't have to step backwards and just use the block prefix?
___
*[`sstable/colblk/prefix_bytes.go` line 153 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O28El_e-1MyleB4uI7U:-O28El_e-1MyleB4uI7V:b-nu3c47) ([raw file](https://github.com/cockroachdb/pebble/blob/453aa1879b928efcb5c73651265dc8e4888771c0/sstable/colblk/prefix_bytes.go#L153)):*
> ```Go
> // the slice's offset equals the previous slice's offset then we step backward
> // until we find a non-empty slice or the start of the bundle (a variable number
> // of steps, but bounded by the bundle size).
> ```

Even if we find a non-empty slice we need to step backward until the bundle prefix, since keys within the bundle are delta encoded wrt to each other, yes?
___
*[`sstable/colblk/prefix_bytes.go` line 160 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O28FOuz6lY0oL3D3p51:-O28FOuz6lY0oL3D3p52:b-426yht) ([raw file](https://github.com/cockroachdb/pebble/blob/453aa1879b928efcb5c73651265dc8e4888771c0/sstable/colblk/prefix_bytes.go#L160)):*
> ```Go
> // middle of a run of equal keys or at an edge. When reverse iteration steps
> // over an edge it has to continue backward until a non-empty slice is found
> // (just as in absolute positioning).
> ```

There is probably some visual in your head related to this "step over an edge" -- could you elaborate?

Reverse iteration in the rowblk had this cacheEntry logic to undo the prefix encoding and cache that. Is this comment saying that with colblk in the case of "step over an edge", which could be quite common (when there aren't multiple MVCC versions of a key), we repeatedly do absolute positioning?

jbowens (commented):
@sumeerbhola  put up #3780

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3756#-:-O2AKUwz6iLTRLIwOXaY:b-9346nt)* status: 2 of 10 files reviewed, 8 unresolved discussions

___
*[`sstable/colblk/prefix_bytes.go` line 89 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O289HbF-_Q4UPoSENA0:-O2A5v_a8cw1DWDXM-bL:b-9vdyw8) ([raw file](https://github.com/cockroachdb/pebble/blob/453aa1879b928efcb5c73651265dc8e4888771c0/sstable/colblk/prefix_bytes.go#L89)):*

is there really a start index, given the table above says "end offset"? Is "end offset" the same as "end index"?
If "end offset" is 2 for aa, then the offset is the end index + 1, yes?

IIRC, RawBytes was encoding the start offsets. By saying that "Offset[0] is the length of the first slice", is this encoding all the end offsets instead?
</blockquote></details>

Good catch, this should be updated. In the PrefixBytes structure, only end offsets are stored because there's no simplification to the read paths be encoding the zero start offset. This is because the block-level prefix is always retrieved separately anyways, so we're not avoiding special casing.

I think the "index" language should also be updated. 
___
*[`sstable/colblk/prefix_bytes.go` line 119 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O28Alj11aYi1iqrGv2R:-O2A7EDY8fz-sdvI14tV:ba0fwx5) ([raw file](https://github.com/cockroachdb/pebble/blob/453aa1879b928efcb5c73651265dc8e4888771c0/sstable/colblk/prefix_bytes.go#L119)):*

why is this not 1 + n + (n + bundleSize-1)/n?
</blockquote></details>

do you mean `1 + n + (n + bundleSize - 1)/bundleSize`?  I think the comment is wrong and intended to say `2+n+(n-1)/bundleSize` which is equivalent. Your formulation is probably more obvious.
___
*[`sstable/colblk/prefix_bytes.go` line 151 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O28ECTb91qJm_N1U4ti:-O2A4rQy8BfKC8ebZHeY:byac771) ([raw file](https://github.com/cockroachdb/pebble/blob/453aa1879b928efcb5c73651265dc8e4888771c0/sstable/colblk/prefix_bytes.go#L151)):*

The offset of the bundle prefix could also be equal to the prefix of the previous slice, yes? In which case we don't have to step backwards and just use the block prefix?
</blockquote></details>

Yeah, if the bundle prefix offset is equal to the previous offset, then the bundle prefix is empty and only the block prefix forms the prefix.
___
*[`sstable/colblk/prefix_bytes.go` line 153 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O28El_e-1MyleB4uI7U:-O2A5CnPBMrCtld1f2e1:b1kmh1y) ([raw file](https://github.com/cockroachdb/pebble/blob/453aa1879b928efcb5c73651265dc8e4888771c0/sstable/colblk/prefix_bytes.go#L153)):*

Even if we find a non-empty slice we need to step backward until the bundle prefix, since keys within the bundle are delta encoded wrt to each other, yes?
</blockquote></details>

With the indexing scheme, we can index directly to extract bundle prefix. But yeah, the non-empty slice is still just encoding the per-row 'suffix', and the complete key is the concatenation of the block prefix, bundle prefix and per-row suffix.
___
*[`sstable/colblk/prefix_bytes.go` line 160 at r13](https://reviewable.io/reviews/cockroachdb/pebble/3756#-O28FOuz6lY0oL3D3p51:-O2AHn778acgCjM0dKZl:bhxutt3) ([raw file](https://github.com/cockroachdb/pebble/blob/453aa1879b928efcb5c73651265dc8e4888771c0/sstable/colblk/prefix_bytes.go#L160)):*

There is probably some visual in your head related to this "step over an edge" -- could you elaborate?

Reverse iteration in the rowblk had this cacheEntry logic to undo the prefix encoding and cache that. Is this comment saying that with colblk in the case of "step over an edge", which could be quite common (when there aren't multiple MVCC versions of a key), we repeatedly do absolute positioning?
</blockquote></details>

I'll push a PR with a clarification to this comment, but this is talking about the case where we were positioned on a non-empty slice and we step back onto an empty slice. We don't know what key the empty slice is duplicating, so we need to scan backwards until a non-empty slice or the bundle prefix is encountered. In the common case where both the current key and the previous key are non-empty, there's no need to do anything.

PR #3759, opened 2024-07-15
manifest: add range annotations

anish-shanbhag:
manifest: add range annotations

This change adds a "range annotation" feature to Annotators , which are computations that aggregate some value over a specific key range within a level. Range annotations use the same B-tree caching behavior as regular annotations, so queries remain fast even with thousands of tables because they avoid a sequential iteration over a level's files.

This PR only sets up range annotations without changing any existing behavior. See cockroachdb#3793 for some potential use cases.

`BenchmarkNumFilesRangeAnnotation` shows that range annotations are significantly faster than using `version.Overlaps` to aggregate over a key range:
```
BenchmarkNumFilesRangeAnnotation/annotator-10         	  306010	      4015 ns/op	      48 B/op	       6 allocs/op
BenchmarkNumFilesRangeAnnotation/overlaps-10          	    2223	    513519 ns/op	     336 B/op	       8 allocs/op
```

### DIFF
--- a/internal/manifest/annotator.go
+++ b/internal/manifest/annotator.go
@@ -4,6 +4,12 @@
 
 package manifest
 
+import (
+	"sort"
+
+	"github.com/cockroachdb/pebble/internal/base"
+)
+
 // The Annotator type defined below is used by other packages to lazily
 // compute a value over a B-Tree. Each node of the B-Tree stores one
 // `annotation` per annotator, containing the result of the computation over
@@ -24,6 +30,10 @@ package manifest
 // computed incrementally as edits are applied to a level.
 type Annotator[T any] struct {
 	Aggregator AnnotationAggregator[T]
+
+	// scratch is used to hold the aggregated annotation value when computing
+	// range annotations in order to avoid additional allocations.
+	scratch *T
 }
 
 // An AnnotationAggregator defines how an annotation should be accumulated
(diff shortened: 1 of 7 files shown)


jbowens (commented):
Do you mind extracting the conversion to generics into its own commit/PR so that it can be reviewed independently? 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3759#-:-O1rtCo91FDgCSPbg4GM:bcmneuk)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @itsbilal)

anish-shanbhag (commented):
Definitely - that PR is now at #3760. I'll rebase this one to just include the range annotation changes once that one is merged.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3759#-:-O1s4dcaAslG49k0KuHm:bw9ri77)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @itsbilal)

jbowens (commented):
Cool!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3759#-:-O3t9zHGCOD14P1H_Tig:b-89q77x)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @anish-shanbhag and @itsbilal)

___
*[`internal/manifest/annotator.go` line 104 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3759#-O3t6-_53cbFooMXTrtK:-O3t6-_53cbFooMXTrtL:boa0hao) ([raw file](https://github.com/cockroachdb/pebble/blob/2d16b801c70abcf6e7e3ab47e502793b10c66aa2/internal/manifest/annotator.go#L104)):*
> ```Go
> 	lowerBound []byte,
> 	// upperBound is a UserKeyBoundary that may be inclusive or exclusive.
> 	upperBound *base.UserKeyBoundary,
> ```

what impact does this have on the existing annotation benchmarks without any ranges?

I expect there is some overhead to combining these two routines, and we might be better off duplicating the function (using mutual recursion when an entire subtree is contained within the bounds). We're also less likely to accidentally begin caching an annotation that does not apply across the node's width.

I think we could also avoid adding a new `scratch` field to every `annotation`, and have the caller pass in a `*T` into which they want the value accumulated. Callers can avoid an allocation by allocating the `T` with their own data types.

anish-shanbhag (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3759)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`internal/manifest/annotator.go` line 104 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3759#-O3t6-_53cbFooMXTrtK:-O4BRHv-0YhA5l3Z02fG:bd70t1b) ([raw file](https://github.com/cockroachdb/pebble/blob/2d16b801c70abcf6e7e3ab47e502793b10c66aa2/internal/manifest/annotator.go#L104)):*

what impact does this have on the existing annotation benchmarks without any ranges?

I expect there is some overhead to combining these two routines, and we might be better off duplicating the function (using mutual recursion when an entire subtree is contained within the bounds). We're also less likely to accidentally begin caching an annotation that does not apply across the node's width.

I think we could also avoid adding a new `scratch` field to every `annotation`, and have the caller pass in a `*T` into which they want the value accumulated. Callers can avoid an allocation by allocating the `T` with their own data types.
</blockquote></details>

You're right, there was some extra overhead introduced:
```
goos: darwin
goarch: arm64
                     │     old     │                new                 │
NumFilesAnnotator-10   1.536µ ± 1%   1.664µ ± 1%  +8.33% (p=0.000 n=10)

                     │    old     │                new                │
                     │    B/op    │    B/op     vs base               │
NumFilesAnnotator-10   536.0 ± 0%   537.0 ± 0%  +0.19% (p=0.000 n=10)

                     │    old     │              new               │
                     │ allocs/op  │ allocs/op   vs base            │
NumFilesAnnotator-10   7.000 ± 0%   7.000 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

Agreed that it's better to separate the two functions - I've updated with that change, and the logic is definitely looking cleaner.

I realized that we can just use a single `scratch` field for the entire Annotator, rather than one for each annotation. My updated implementation takes that route, but let me know if you think there's still a benefit to allowing callers to pass a custom `*T`.

jbowens (approved):
:lgtm_strong: 

Nice!

> Reviewed 1 of 13 files at r2, 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3759#-:-O4GScu08QOdiBX9G73C:bkx9igh)* status: 3 of 13 files reviewed, all discussions resolved (waiting on @itsbilal)

anish-shanbhag (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3759#-:-O4G_nx8500FIY_EhsKg:b-814g8k)* status: 3 of 13 files reviewed, all discussions resolved (waiting on @itsbilal)

PR #3760, opened 2024-07-15
manifest: improve Annotator interface with generics

anish-shanbhag:
Refactors `manifest.Annotator` to use generics and a simplified API. This eliminates the need to perform pointer manipulation and unsafe typecasting when defining a new Annotator. The goal of this change is to improve the `Annotator` interface while not changing any existing behavior.

`BenchmarkNumFilesAnnotator` shows roughly the same performance as master when compared
to the equivalent implementation of `orderStatistic`:
```
                     │     old     │             new              │
                     │   sec/op    │   sec/op     vs base         │
NumFilesAnnotator-10   1.635µ ± 1%   1.572µ ± 7%  ~ (p=0.065 n=6)

                     │    old     │              new              │
                     │    B/op    │    B/op     vs base           │
NumFilesAnnotator-10   536.0 ± 0%   536.0 ± 0%  ~ (p=1.000 n=6) ¹
¹ all samples are equal

                     │    old     │              new              │
                     │ allocs/op  │ allocs/op   vs base           │
NumFilesAnnotator-10   7.000 ± 0%   7.000 ± 0%  ~ (p=1.000 n=6) ¹
¹ all samples are equal
```

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -639,39 +639,13 @@ func compensatedSize(f *fileMetadata) uint64 {
 	return f.Size + fileCompensation(f)
 }
 
-// compensatedSizeAnnotator implements manifest.Annotator, annotating B-Tree
-// nodes with the sum of the files' compensated sizes. Its annotation type is
-// a *uint64. Compensated sizes may change once a table's stats are loaded
-// asynchronously, so its values are marked as cacheable only if a file's
-// stats have been loaded.
-type compensatedSizeAnnotator struct {
-}
-
-var _ manifest.Annotator = compensatedSizeAnnotator{}
-
-func (a compensatedSizeAnnotator) Zero(dst interface{}) interface{} {
-	if dst == nil {
-		return new(uint64)
-	}
-	v := dst.(*uint64)
-	*v = 0
-	return v
-}
-
-func (a compensatedSizeAnnotator) Accumulate(
-	f *fileMetadata, dst interface{},
-) (v interface{}, cacheOK bool) {
-	vptr := dst.(*uint64)
-	*vptr = *vptr + compensatedSize(f)
-	return vptr, f.StatsValid()
-}
-
-func (a compensatedSizeAnnotator) Merge(src interface{}, dst interface{}) interface{} {
-	srcV := src.(*uint64)
-	dstV := dst.(*uint64)
-	*dstV = *dstV + *srcV
-	return dstV
-}
+// compensatedSizeAnnotator is a manifest.Annotator that annotates B-Tree
+// nodes with the sum of the files' compensated sizes. Compensated sizes may
+// change once a table's stats are loaded asynchronously, so its values are
+// marked as cacheable only if a file's stats have been loaded.
+var compensatedSizeAnnotator = manifest.SumAnnotator(func(f *fileMetadata) (uint64, bool) {
+	return compensatedSize(f), f.StatsValid()
+})
 
 // totalCompensatedSize computes the compensated size over a file metadata
 // iterator. Note that this function is linear in the files available to the
@@ -912,10 +886,6 @@ func calculateSizeAdjust(inProgressCompactions []compactionInfo) [numLevels]leve
 	return sizeAdjust
 }
 
-func levelCompensatedSize(lm manifest.LevelMetadata) uint64 {
-	return *lm.Annotation(compensatedSizeAnnotator{}).(*uint64)
-}
-
 func (p *compactionPickerByScore) calculateLevelScores(
 	inProgressCompactions []compactionInfo,
 ) [numLevels]candidateLevelInfo {
@@ -932,7 +902,7 @@ func (p *compactionPickerByScore) calculateLevelScores(
 	}
 	sizeAdjust := calculateSizeAdjust(inProgressCompactions)
 	for level := 1; level < numLevels; level++ {
-		compensatedLevelSize := levelCompensatedSize(p.vers.Levels[level]) + sizeAdjust[level].compensated()
+		compensatedLevelSize := *compensatedSizeAnnotator.LevelAnnotation(p.vers.Levels[level]) + sizeAdjust[level].compensated()
 		scores[level].compensatedScore = float64(compensatedLevelSize) / float64(p.levelMaxBytes[level])
 		scores[level].uncompensatedScore = float64(p.vers.Levels[level].Size()+sizeAdjust[level].actual()) / float64(p.levelMaxBytes[level])
 	}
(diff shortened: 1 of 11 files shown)


jbowens (commented):
Looking great! Much cleaner!

> Reviewed 5 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3760#-:-O1wQTYKA108eIAWb0gw:bts8ihg)* status: 5 of 11 files reviewed, 4 unresolved discussions (waiting on @anish-shanbhag and @itsbilal)

___
*[`-- commits` line 21 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3760#-O1wQ-BH9zzXlBI9w4bG):*
do you know where this additional allocation is coming from? A heap profile `go test -run XXX -bench BenchmarkNumFilesAnnotator -memprofile mem.prof` should help 
___
*[`compaction_picker.go` line 1390 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3760#-O1wKtgb0AKmh4PppXQ4:-O1wKtgb0AKmh4PppXQ5:bt9tlof) ([raw file](https://github.com/cockroachdb/pebble/blob/b0d03ebddca9232fb5100f97dfc24c772dc9c947/compaction_picker.go#L1390)):*
> ```Go
> 			// TODO(travers): Consider an alternative heuristic for elision of range-keys.
> 			deletionCriteria := f.Stats.RangeDeletionsBytesEstimate*10 >= f.Size || f.Stats.NumDeletions*10 > f.Stats.NumEntries
> 			return !f.IsCompacting() && f.StatsValid() && deletionCriteria, f.StatsValid()
> ```

nit: I think the previous structure with an early exit if `!f.StatsValid()` was a little clearer—the stats we use in computing `deletionCriteria` might not be initialized at all, and it's less obvious that we're not improperly using them
___
*[`internal/manifest/annotator.go` line 75 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3760#-O1wMrdpEycOoz7xGJs0:-O1wMrdpEycOoz7xGJs1:b-jbo1r0) ([raw file](https://github.com/cockroachdb/pebble/blob/b0d03ebddca9232fb5100f97dfc24c772dc9c947/internal/manifest/annotator.go#L75)):*
> ```Go
> 	n.annot = append(n.annot, annotation{
> 		annotator: a,
> 		vptr:      new(T),
> ```

The annotations where `T = *fileMetadata` end up double boxing (storing a `**fileMetadata` in `vptr`). Maybe the responsibility of boxing the value should be the responsibility of whoever defines `T` (eg, `T` should always be a pointer type to avoid an allocation). Then the `PickFileAnnotator` continues to use `T = *fileMetadata`, boxing once.
___
*[`internal/manifest/annotator.go` line 234 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3760#-O1wPDG91eTJm_j_CW_2:-O1wPDG91eTJm_j_CW_3:but3yb5) ([raw file](https://github.com/cockroachdb/pebble/blob/b0d03ebddca9232fb5100f97dfc24c772dc9c947/internal/manifest/annotator.go#L234)):*
> ```Go
> 		return f1
> 	}
> 	return f2
> ```

nit: worth using a switch:

```
switch {
case f1 == nil:
    return f2
case f2 == nil:
    return f1
case fa.Compare(f1, f2):
    return f1
default:
    return f2
}
```

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3760)* status: 3 of 11 files reviewed, 2 unresolved discussions (waiting on @anish-shanbhag and @itsbilal)

___
*[`-- commits` line 21 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3760#-O1wQ-BH9zzXlBI9w4bG):*

Spent a while trying to find the allocation in the Annotator, but turns out this was actually because `require.EqualValues` allocated a new `uint64` when two different types were passed for comparison (in this case `int` and `uint64`. This is fixed now; I also added the updated benchmark results above.
</blockquote></details>

Ah nice
___
*[`internal/manifest/annotator.go` line 75 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3760#-O1wMrdpEycOoz7xGJs0:-O20iCbu4eKIaye4c2hE:bbmitsg) ([raw file](https://github.com/cockroachdb/pebble/blob/b0d03ebddca9232fb5100f97dfc24c772dc9c947/internal/manifest/annotator.go#L75)):*
I think the most significant downside is that double boxing forces a new allocation. Without the additional boxing, we would be storing `*filetMetadata` directly as the annotation value without suffering any additional allocations.

> If we place the responsibility of boxing on whoever defines `T`, then they also have implement the initial allocation in `Zero` and merging into an existing value in `Accumulate` (i.e. very similar to the previous implementation, minus the typecasting).

IMO, this is worth it. In the case of the `*fileMetadata` annotations, there is no initial allocation. The zero value is just a nil pointer. In all the non-nil instances, we'll just be passing around pointers to the already-allocated `fileMetadata`.

At least the pointer handling will be encapsulated within the `sumAggregator` for those cases.

anish-shanbhag (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3760)* status: 2 of 11 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`internal/manifest/annotator.go` line 75 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3760#-O1wMrdpEycOoz7xGJs0:-O2Q-Jw04l32sM9pQjq1:bkatlru) ([raw file](https://github.com/cockroachdb/pebble/blob/b0d03ebddca9232fb5100f97dfc24c772dc9c947/internal/manifest/annotator.go#L75)):*

I think the most significant downside is that double boxing forces a new allocation. Without the additional boxing, we would be storing `*filetMetadata` directly as the annotation value without suffering any additional allocations.

> If we place the responsibility of boxing on whoever defines `T`, then they also have implement the initial allocation in `Zero` and merging into an existing value in `Accumulate` (i.e. very similar to the previous implementation, minus the typecasting).

IMO, this is worth it. In the case of the `*fileMetadata` annotations, there is no initial allocation. The zero value is just a nil pointer. In all the non-nil instances, we'll just be passing around pointers to the already-allocated `fileMetadata`.

At least the pointer handling will be encapsulated within the `sumAggregator` for those cases.
</blockquote></details>

Makes sense, I hadn't considered the fact that we avoid the extra allocation here. I've updated to avoid the double boxing.

jbowens (approved):
:lgtm:

nice, looks great!

> Reviewed 2 of 5 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3760#-:-O2fH705C5SwGdURduSk:b-ixeipq)* status: 6 of 11 files reviewed, 2 unresolved discussions (waiting on @anish-shanbhag and @itsbilal)

anish-shanbhag (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3760#-:-O2jgIxa6ObUBF7PcGt5:b-814g8k)* status: 6 of 11 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @jbowens)

PR #3762, opened 2024-07-16
colblk: define block structure

jbowens:
This commit defines the structure of a columnar block, defining a block header and footer, and routines for reading columns from an encoded block. Additionally, the interfaces around ColumnEncodings are refactored so that individual data types are responsible for encoding metadata about their encoding within the column data itself if necessary. Currently, this means the delta encoding for uint columns is moved into the first byte of the uint column data.


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3762#-:-O20xXXR4GQeYTmoI9bj:bnfp4nl)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/colblk/block.go` line 19 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3762#-O20onK463hA1iuTWaKF:-O20onK463hA1iuTWaKG:buqqy5d) ([raw file](https://github.com/cockroachdb/pebble/blob/0f50e444becf3ce3788206e31e18e13fd4c3a132/sstable/colblk/block.go#L19)):*
> ```Go
> // - The number of columns in the block (2 bytes)
> // - The number of rows in the block (4 bytes)
> //
> ```

Should we reserve some bytes here for future use?
___
*[`sstable/colblk/block.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3762#-O20pDKc8QRDwOvB7hD3:-O20pDKd6OMEtx0Ji1XL:bw33ast) ([raw file](https://github.com/cockroachdb/pebble/blob/0f50e444becf3ce3788206e31e18e13fd4c3a132/sstable/colblk/block.go#L75)):*
> ```Go
> // the column is the last column within the block, this pointer would point to a
> // byte outside of the block's allocation which is a violation of Go pointer
> // rules. The trailing version byte ensures that these pointers are always valid.
> ```

[nit] I would move the version to the header and just add a padding byte at the end. Presumably we might need to look at the version upfront and it's easier if it's in the header.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3762#-:-O21NMbjEIwX8eurifOr:b-814g8k)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`sstable/colblk/block.go` line 19 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3762#-O20onK463hA1iuTWaKF:-O21N5Tg9eoPXg391Jgk:bvgo1w2) ([raw file](https://github.com/cockroachdb/pebble/blob/0f50e444becf3ce3788206e31e18e13fd4c3a132/sstable/colblk/block.go#L19)):*

Should we reserve some bytes here for future use?
</blockquote></details>

I think we could use the version byte to expand the size of the header if we need to?
___
*[`sstable/colblk/block.go` line 75 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3762#-O20pDKc8QRDwOvB7hD3:-O21NE0WEWhfyD6coMv_:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/0f50e444becf3ce3788206e31e18e13fd4c3a132/sstable/colblk/block.go#L75)):*

[nit] I would move the version to the header and just add a padding byte at the end. Presumably we might need to look at the version upfront and it's easier if it's in the header.
</blockquote></details>

Done

PR #3765, opened 2024-07-17
crl-release-23.2: objstorage: fix race in vfsSync

RaduBerinde:
#### crl-release-23.2: objstorage: improve ParallelSync test to expose race in vfsSync

We improve this test by using a strict MemFS, simulating a crash at
a random point, and checking that all objects that were synced are
accessible.

#### crl-release-23.2: objstorage: fix race in vfsSync

We have a race in `vfsSync`: if two goroutines Sync at the same time,
one might see the flag=false and return immediately, before the other
goroutine actually runs/completes the Sync.

The fix is to store a change counter and wait for any in-progress
syncs as necessary.

Informs https://github.com/cockroachdb/cockroach/issues/124845

### DIFF
--- a/objstorage/objstorageprovider/provider.go
+++ b/objstorage/objstorageprovider/provider.go
@@ -45,9 +45,14 @@ type provider struct {
 			storageObjects map[remote.Locator]remote.Storage
 		}
 
-		// localObjectsChanged is set if non-remote objects were created or deleted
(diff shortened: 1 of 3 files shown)


jbowens (approved):

RaduBerinde:
TFTR!

PR #3766, opened 2024-07-17
sstable: plumb context to range iterator constructors

RaduBerinde:
This context covers the read of the range del or range key block.

Informs #3728

### DIFF
--- a/external_iterator.go
+++ b/external_iterator.go
@@ -166,7 +166,7 @@ func createExternalPointIter(ctx context.Context, it *Iterator) (topLevelIterato
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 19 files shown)


sumeerbhola (approved):
is there enough plumbing now that `TestTracing` can be updated to show the effect of a slow read of a rangedel and rangekey block?

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3766#-:-O2AfFK46AjNSZGGbLwd:b-jpov18)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @itsbilal)

RaduBerinde:
Good idea. Looks like not yet.. we need to plumb it through `keyspanimpl.TableNewSpanIter` and it's not trivial. I will do that in a separate PR.

PR #3770, opened 2024-07-17
crl-release-24.2: sstable: improve slow block read log

RaduBerinde:
Add the two callers to identify the code path (and implicitly what
kind of block it is).


RaduBerinde:
I also backported #3745 which didn't make it into the 24.2 branch.

nicktrav (approved):

PR #3777, opened 2024-07-18
colblk: impose consistent interfaces

jbowens:
**colblk: make decoder function signatures consistent**

Unify the various functions that decode data structures from a columnar block
so that they all have consistent function signatures, and define a generic
DecodeFunc that captures this consistency. In future work, the DecodeFunc will
be used to aid in composing decoders.

**colblk: implement a common interface across column data accessors**

Introduce a new generic `Array[T]` type that allows indexed access to data and
implement it across the various accessors.

**colblk: randomize BundleSizes in block writer randomized test**

Randomize the BundleSize used for PrefixBytes columns in the
TestBlockWriterRandomized test.

**colblk: add facility for cloning an Array to a []T**

Add a `Clone[T](Array[T], int)` func that clones an array of n elements into a
heap-allocated Go slice. It's primarily intended for use in tests.

### DIFF
--- a/sstable/colblk/bitmap.go
+++ b/sstable/colblk/bitmap.go
@@ -30,6 +30,9 @@ type Bitmap struct {
 	bitCount int
 }
 
+// Assert that Bitmap implements Array[bool].
+var _ Array[bool] = Bitmap{}
+
 // DecodeBitmap decodes the structure of a Bitmap and returns a Bitmap that
 // reads from b supporting bitCount logical bits. No bounds checking is
 // performed, so the caller must guarantee the bitmap is appropriately sized and
@@ -50,8 +53,8 @@ func DecodeBitmap(b []byte, off uint32, bitCount int) (bitmap Bitmap, endOffset 
 // Assert that DecodeBitmap implements DecodeFunc.
 var _ DecodeFunc[Bitmap] = DecodeBitmap
 
-// Get returns true if the bit at position i is set and false otherwise.
-func (b Bitmap) Get(i int) bool {
+// At returns true if the bit at position i is set and false otherwise.
+func (b Bitmap) At(i int) bool {
 	return (b.data.At(i>>6 /* i/64 */) & (1 << uint(i%64))) != 0
 }
 
--- a/sstable/colblk/bitmap.go
+++ b/sstable/colblk/bitmap.go
@@ -30,21 +30,25 @@ type Bitmap struct {
 	bitCount int
 }
 
-// MakeBitmap returns a Bitmap that reads from b supporting bitCount logical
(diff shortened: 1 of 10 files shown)


RaduBerinde (approved):
:lgtm:  Very nice!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3777#-:-O27nNPu7uocA8qdCmAn:b-8ymd1x)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @aadityasondhi and @jbowens)

___
*[`sstable/colblk/prefix_bytes.go` line 174 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3777#-O27lZdODQA2x2ivtUDf:-O27lZdP-zCJX71358BN:blaechz) ([raw file](https://github.com/cockroachdb/pebble/blob/da3ffb6f830762cf73dc4c6061b61af78f8c4168/sstable/colblk/prefix_bytes.go#L174)):*
> ```Go
> }
> 
> // DecodePrefixBytes decodes the structure of a PrefixBytes constructs an
> ```

[nit] `, constructing`
___
*[`sstable/colblk/block_test.go` line 365 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3777#-O27muo8C8S0KQZqwH6S:-O27muo8C8S0KQZqwH6T:b-5hlkjp) ([raw file](https://github.com/cockroachdb/pebble/blob/5868ffd3b34289d88264f06e9d5c05798f00872f/sstable/colblk/block_test.go#L365)):*
> ```Go
> 			schema[j].DataType = DataType(randInt(1, int(dataTypesCount)))
> 			if schema[j].DataType == DataTypePrefixBytes {
> 				schema[j].BundleSize = 1 << randInt(2, 6)
> ```

[nit] We should make sure the range includes the minimum supported value (is that 1?)

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3777#-:-O2AAlVW46-N3gBUcJ-8:b-814g8k)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @aadityasondhi)

___
*[`sstable/colblk/block_test.go` line 365 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3777#-O27muo8C8S0KQZqwH6S:-O2AAj3h3MFh3qi1XwZP:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/5868ffd3b34289d88264f06e9d5c05798f00872f/sstable/colblk/block_test.go#L365)):*

[nit] We should make sure the range includes the minimum supported value (is that 1?)
</blockquote></details>

Done

PR #3779, opened 2024-07-19
base: make Compare more restrictive

RaduBerinde:
The new block format assumes that prefixes are ordered lexicographically. We now require the `Compare` function to adhere to this assumption. This is already the case with our test comparators and the CRDB comparator.

We add a `Comparer` wrapper that verifies this assumption, and we use this in a low percentage of `invariant` test runs.


jbowens (approved):
Nice

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3779#-:-O2AURJN1aqYXMO3Clc-:bmu9b66)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

PR #3781, opened 2024-07-20
colblk: support absent column values [high-level review only]

jbowens:
Edit: I'm leaning against this additional complexity after realizing https://github.com/cockroachdb/pebble/pull/3781#pullrequestreview-2190047479

----

Prefix all columns with a mapping indicating which rows have values encoded within the column data. Introduce a new PresenceBitmap used to encode this mapping using 2-bits per row and supporting O(1) lookup of a present value's index within the column's data array.

When all values are present or absent, the presence bitmap is elided from the encoding, consuming just a single byte to encode an enum variant.

The addition of presence is intended as an optimization for columns with a reasonable expectation that many values will be absent. For example, in the case of keyspan (range key or range deletion) blocks, the suffix (Key.Suffix) and value (Key.Value) columns encode byte slices that are expected to often be empty slices. The RawBytes encoding can encode empty byte slices but at the cost of at least 1 offset per key. In the worst case, the presence bitmap adds a cost of 2-bits for all rows, while allowing absent values to suffer no additional encoding overhead.

Although all columns are prefixed with a presence encoding, column encoders may use AllPresentWriter to encode a column with no absent values. When a column is statically known to not contain absent values, readers may use AllPresentDecoder to obtain a column accessor that doesn't suffer any additional indirection on lookups.


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3781)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`-- commits` line 17 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3781#-O2GajlE1Gtnb2lai8Or):*
Only while polishing this commit message did I realize that with the new RawBytes encoding, a RawBytes column of all empty bytes slices benefits from the UintDeltaEncodingConstant encoding of the offsets table. I'm questioning how necessary the presence bitmap is now

RaduBerinde:
Will we ever have to distinguish between empty values and not-present values?

RaduBerinde:
Another question - how do we plan to encode timestamps? If we separate the Logical from the WallTime and encode them separately as integers, the logical might be 0 for a lot of rows so a bitmap might make a big difference there?

jbowens:
> Will we ever have to distinguish between empty values and not-present values?

Not in any of the immediate use cases. We could conceivably one day decompose KVs into their SQL column fields and need to express NULLs.

> Another question - how do we plan to encode timestamps? If we separate the Logical from the WallTime and encode them separately as integers, the logical might be 0 for a lot of rows so a bitmap might make a big difference there?

Yeah, we plan to encode it separately. The expected case is that a block will contain all 0s for logical components, so the logical column will be encoded using a constant-sized encoding. If there any keys with a nonzero logical time, we’ll devolve into storing at least a byte per row.

I also prototyped a group-varint32 encoding column that stores n/4 group varint32s. It uses RawBytes to encode the varints, and special cases the the case where all four ints are zero as an empty slice. This also results in a constant-sized encoding for a column of all zeroes. When it needs to store non-zero values, it adds at least n/4 offsets. It would be very unlikely to need more than 1-byte per offset for the logical column, so it would also work out to about 2 bits/row of overhead.

RaduBerinde:
Understood. Then these seem like minor optimizations in the grand scheme of things so I don't think we should pursue them now. They can be done later and we can incorporate them if we prove some reasonable benefit with benchmarks.

jbowens:
Agreed, going to close this out and  we can revive elements if we find needs for them

PR #3783, opened 2024-07-20
keyspan: add contexts to keyspan iterators

RaduBerinde:
Informs #3728

#### db: convert TestTracing to datadriven

Rewrite this test to use `datadriven`. This will make it much easier
to update and extend.

#### keyspan: add contexts to keyspan iterators

This change adds a `SetContext()` method to `FragmentIterator` which
updates any context stored in an iterator and its children. The
iterator context is now plumbed correctly through to the table range
key iterators.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -807,9 +807,8 @@ func (c *compaction) newInputIters(
 				}
 			}
 			if hasRangeKeys {
-				li := &keyspanimpl.LevelIter{}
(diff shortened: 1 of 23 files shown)


sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 2 files at r4, 3 of 22 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3783#-:-O2UewwR1sn-4oOxNRgf:bg7igkv)* status: 4 of 23 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

PR #3784, opened 2024-07-22
sstable: new property collector API

RaduBerinde:
#### sstable: new property collector API

We simplify the property collector API: an instance now only needs to
maintain a single state, not three (data block, index, table). The
sstable writer uses multiple separate instances in parallel. The
instance for index blocks uses a new `AddCollected` API which adds
computed properties from the data blocks. Similarly, the table
properties are calculated in a separate instance using `AddCollected`
with the properties from the index blocks and the range key block.

This new interface will allow re-deriving index and table properties
for a partial file download which copies whole data blocks. In the
future, this would also allow copying entire data blocks during
compactions.

### DIFF
--- a/iterator_test.go
+++ b/iterator_test.go
@@ -280,6 +280,9 @@ func (c *minSeqNumPropertyCollector) SupportsSuffixReplacement() bool {
 	return false
 }
 
+// Close is part of the BlockPropertyCollector interface.
+func (c *minSeqNumPropertyCollector) Close() {}
+
 // minSeqNumFilter is a BlockPropertyFilter that uses the
 // minSeqNumPropertyCollector data to filter out entire tables.
 type minSeqNumFilter struct {
--- a/iterator_test.go
+++ b/iterator_test.go
@@ -255,22 +255,23 @@ func (c *minSeqNumPropertyCollector) AddRangeKeys(span sstable.Span) error {
 	return nil
 }
 
-func (c *minSeqNumPropertyCollector) FinishDataBlock(buf []byte) ([]byte, error) {
-	return nil, nil
+func (c *minSeqNumPropertyCollector) Finish(buf []byte) []byte {
+	return binary.AppendUvarint(buf, uint64(c.minSeqNum))
 }
 
-func (c *minSeqNumPropertyCollector) AddPrevDataBlockToIndexBlock() {}
-
-func (c *minSeqNumPropertyCollector) FinishIndexBlock(buf []byte) ([]byte, error) {
-	return nil, nil
-}
-
-func (c *minSeqNumPropertyCollector) FinishTable(buf []byte) ([]byte, error) {
-	return binary.AppendUvarint(buf, uint64(c.minSeqNum)), nil
+func (c *minSeqNumPropertyCollector) AddCollected(prop []byte) error {
+	res, n := binary.Uvarint(prop)
+	if n <= 0 {
+		panic("could not decode")
+	}
+	if c.minSeqNum == 0 || c.minSeqNum > base.SeqNum(res) {
+		c.minSeqNum = base.SeqNum(res)
+	}
+	return nil
 }
 
 func (c *minSeqNumPropertyCollector) AddCollectedWithSuffixReplacement(
-	oldProp []byte, oldSuffix, newSuffix []byte,
+	oldProp []byte, newSuffix []byte,
 ) error {
 	return errors.Errorf("not implemented")
 }
--- a/sstable/block_property.go
+++ b/sstable/block_property.go
@@ -11,23 +11,26 @@ import (
 	"sync"
 	"unsafe"
 
+	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/internal/keyspan"
 )
 
 // Block properties are an optional user-facing feature that can be used to
-// filter data blocks (and whole sstables) from an Iterator before they are
-// loaded. They do not apply to range delete blocks. These are expected to
-// very concisely represent a set of some attribute value contained within the
-// key or value, such that the set includes all the attribute values in the
-// block. This has some similarities with OLAP pruning approaches that
-// maintain min-max attribute values for some column (which concisely
-// represent a set), that is then used to prune at query time. In Pebble's
-// case, data blocks are small, typically 25-50KB, so these properties should
-// reduce their precision in order to be concise -- a good rule of thumb is to
-// not consume more than 50-100 bytes across all properties maintained for a
-// block, i.e., a 500x reduction compared to loading the data block.
+// filter data blocks or index blocks or whole sstables from an Iterator before
+// they are loaded.
+//
+// Block properties are expected to very concisely represent a set of some
+// attribute value contained within the key or value, such that the set includes
+// all the attribute values in the block. This has some similarities with OLAP
+// pruning approaches that maintain min-max attribute values for some column
+// (which concisely represent a set), that is then used to prune at query time.
+// In Pebble's case, data blocks are small, typically 25-50KB, so these
+// properties should reduce their precision in order to be concise -- a good
+// rule of thumb is to not consume more than 50-100 bytes across all properties
+// maintained for a block, i.e., a 500x reduction compared to loading the data
+// block.
 //
 // A block property must be assigned a unique name, which is encoded and
 // stored in the sstable. This name must be unique among all user-properties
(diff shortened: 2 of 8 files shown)


jbowens (commented):
> Reviewed 5 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3784)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/block_property.go` line 48 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3784#-O3-0vOjCBsp1EN6jK9x:-O3-0vOjCBsp1EN6jK9y:bwztx9) ([raw file](https://github.com/cockroachdb/pebble/blob/232b685f610afd7a8a2d5120339b6ba870e79cb5/sstable/block_property.go#L48)):*
> ```Go
> // derivable just from the properties of the data blocks it contains. Similarly,
> // the table properties must be derivable just from the properties of the index
> // blocks and range block.
> ```

I'm trying to think through whether we might regret this loss of power and expressibility. Even if we did eventually come to want to use some property that can't be hierarchical, I suppose it wouldn't be much effort to work it back in.
___
*[`sstable/block_property.go` line 125 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3784#-O3-3Xva-MJTueNtjpdf:-O3-3Xva-MJTueNtjpdg:bmlr0co) ([raw file](https://github.com/cockroachdb/pebble/blob/232b685f610afd7a8a2d5120339b6ba870e79cb5/sstable/block_property.go#L125)):*
> ```Go
> 	// error). SupportsSuffixReplacement() can be used to check if this method is
> 	// implemented.
> 	AddCollectedWithSuffixReplacement(oldProp []byte, newSuffix []byte) error
> ```

It seems we could further simplify the suffix replacement bit by making it a stateless function like

```
type BlockPropertySuffixTransform func(buf []byte, oldProp, newSuffix []byte) []byte
```

It feels a bit off to force `BlockPropertyCollector`s to know about suffix replacement through `SupportsSuffixReplacement` and `AddCollectedWithSuffixReplacement`. Instead, the caller initiating the suffix replacement could supply a map of collector names to transforms. We error out if a block property exists for which there is no supplied transform.
___
*[`sstable/writer.go` line 2047 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3784#-O3-Gcfl0pVCyGljL2Ad:-O3-Gcfl0pVCyGljL2Ae:b-iu2yc) ([raw file](https://github.com/cockroachdb/pebble/blob/232b685f610afd7a8a2d5120339b6ba870e79cb5/sstable/writer.go#L2047)):*
> ```Go
> 				// Add range key properties.
> 				rangeKeyProp := w.blockPropCollectors[i].rangeKeyBlock.Finish(scratch)
> 				if err := tableCollector.AddCollected(rangeKeyProp); err != nil {
> ```

Is this right? My  recollection is that range key properties and point key properties are separate. Eg, a table-level block property should not reflect the state of the contained range keys. The reason is because the point and range key iterators are separate, so the decision of whether or not to examine a particular file should be too. For this reason `IterOptions` takes separate block-property filters for point keys and range keys: https://github.com/cockroachdb/pebble/blob/368636f0554c02076aeda2778e1905987c9de0fb/options.go#L133-L145

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3784)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`sstable/writer.go` line 2047 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3784#-O3-Gcfl0pVCyGljL2Ad:-O3-cz7Y3AbcfrSuRTYH:b1n7sdf) ([raw file](https://github.com/cockroachdb/pebble/blob/232b685f610afd7a8a2d5120339b6ba870e79cb5/sstable/writer.go#L2047)):*

Is this right? My  recollection is that range key properties and point key properties are separate. Eg, a table-level block property should not reflect the state of the contained range keys. The reason is because the point and range key iterators are separate, so the decision of whether or not to examine a particular file should be too. For this reason `IterOptions` takes separate block-property filters for point keys and range keys: https://github.com/cockroachdb/pebble/blob/368636f0554c02076aeda2778e1905987c9de0fb/options.go#L133-L145
</blockquote></details>

Hm.. but there is no separate place to store the range-key derived properties, is there?  We don't write out props for the range key block currently..

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3784)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`sstable/block_property.go` line 125 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3784#-O3-3Xva-MJTueNtjpdf:-O3-dxbs70vvlblEw6jg:bwupoq0) ([raw file](https://github.com/cockroachdb/pebble/blob/232b685f610afd7a8a2d5120339b6ba870e79cb5/sstable/block_property.go#L125)):*

It seems we could further simplify the suffix replacement bit by making it a stateless function like

```
type BlockPropertySuffixTransform func(buf []byte, oldProp, newSuffix []byte) []byte
```

It feels a bit off to force `BlockPropertyCollector`s to know about suffix replacement through `SupportsSuffixReplacement` and `AddCollectedWithSuffixReplacement`. Instead, the caller initiating the suffix replacement could supply a map of collector names to transforms. We error out if a block property exists for which there is no supplied transform.
</blockquote></details>

This mapping would have to be provided through the Options, yes?  Feels more clunky than requiring it innthe implementation.. but I'll experiment with it.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3784)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/writer.go` line 2047 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3784#-O3-Gcfl0pVCyGljL2Ad:-O3-pyPf8BQoy3oGml2Z:b38ru) ([raw file](https://github.com/cockroachdb/pebble/blob/232b685f610afd7a8a2d5120339b6ba870e79cb5/sstable/writer.go#L2047)):*

Hm.. but there is no separate place to store the range-key derived properties, is there?  We don't write out props for the range key block currently..
</blockquote></details>

Hmmm... maybe range key block property filters have just never worked. I know we don't use them in Cockroach

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3784)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`sstable/block_property.go` line 48 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3784#-O3-0vOjCBsp1EN6jK9x:-O30-bK17rLwEKQAod7Q:b-3kgx6h) ([raw file](https://github.com/cockroachdb/pebble/blob/232b685f610afd7a8a2d5120339b6ba870e79cb5/sstable/block_property.go#L48)):*

I'm trying to think through whether we might regret this loss of power and expressibility. Even if we did eventually come to want to use some property that can't be hierarchical, I suppose it wouldn't be much effort to work it back in.
</blockquote></details>

Yes.. I could imagine a case where the property is too large to store at the data block level but is reasonable for the table level. But if we want to be able to derive properties for partially downloaded sst files, we will need this stricter framework.
___
*[`sstable/writer.go` line 2047 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3784#-O3-Gcfl0pVCyGljL2Ad:-O300DWV0zHjYeHQihO6:b-ah1u65) ([raw file](https://github.com/cockroachdb/pebble/blob/232b685f610afd7a8a2d5120339b6ba870e79cb5/sstable/writer.go#L2047)):*

Hmmm... maybe range key block property filters have just never worked. I know we don't use them in Cockroach
</blockquote></details>

The RangeKeyFilters code use the table properties (newRangeKeyIter -> checkAndIntersectFilters -> sstable.IntersectsTable), so they should work (but of course the properties will include point keys).

It looks like we stopped using them in https://github.com/cockroachdb/cockroach/issues/86260 but I'm not sure I quite understand the problem.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3784)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/writer.go` line 2047 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3784#-O3-Gcfl0pVCyGljL2Ad:-O33nNNC5SawWQ5mhyQ6:b-9nz1sa) ([raw file](https://github.com/cockroachdb/pebble/blob/232b685f610afd7a8a2d5120339b6ba870e79cb5/sstable/writer.go#L2047)):*

The RangeKeyFilters code use the table properties (newRangeKeyIter -> checkAndIntersectFilters -> sstable.IntersectsTable), so they should work (but of course the properties will include point keys).

It looks like we stopped using them in https://github.com/cockroachdb/cockroach/issues/86260 but I'm not sure I quite understand the problem.
</blockquote></details>

The MVCCIncrementalIterator relies on very subtle semantics to keep the time-bound iterator and non-bounded iterator in sync. I believe the issue is that the user-observed fragmentation varies (because we're hiding some range keys), which breaks the assumptions that keep the two iterators in lock step.

In any case, I don't think we want to combine point and range key properties since we'll only be reducing the effectiveness of each.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3784)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`sstable/writer.go` line 2047 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3784#-O3-Gcfl0pVCyGljL2Ad:-O33qbPXCzSiS-rVhd86:bs6q3w) ([raw file](https://github.com/cockroachdb/pebble/blob/232b685f610afd7a8a2d5120339b6ba870e79cb5/sstable/writer.go#L2047)):*

The MVCCIncrementalIterator relies on very subtle semantics to keep the time-bound iterator and non-bounded iterator in sync. I believe the issue is that the user-observed fragmentation varies (because we're hiding some range keys), which breaks the assumptions that keep the two iterators in lock step.

In any case, I don't think we want to combine point and range key properties since we'll only be reducing the effectiveness of each.
</blockquote></details>

This is great, thanks. So in the short term, we can just not compute range key properties. And in a follow-up I will add the ability to compute them and store them in the range key block handle.

PR #3791, opened 2024-07-25
db: Add BenchmarkPointDeletedSwath

anish-shanbhag:
Adds a new benchmark that measures the performance of seeks and full-table scans after a swath of point tombstones are created in the middle of the keyspace.

This is more or less a copy of cockroachdb#2657 with a few fixes from the comments there. Initial results can be found at [this comment](https://github.com/cockroachdb/pebble/issues/918#issuecomment-2214565707).


iterator_test.go
```diff
@@ -2722,6 +2724,182 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+				t.Fatal(err)
+			}
+			if batch.Len() >= 10<<10 /* 10 kib */ {
+				if err := batch.Commit(NoSync); err != nil {
+					t.Fatal(err)
+				}
+				batch = d.NewBatch()
+			}
+		}
+		if err := batch.Commit(NoSync); err != nil {
```
itsbilal, line 2893: We should probably check if the batch isn't empty before doing this

iterator_test.go
```diff
@@ -2722,6 +2724,182 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+// BenchmarkPointDeletedSwath benchmarks iterator operations on large-ish
+// (hundreds of MBs) databases containing broad swaths of keys removed by point
+// tombstones.
+func BenchmarkPointDeletedSwath(b *testing.B) {
+	const maxKeyLen = 5
+	ks := testkeys.Alpha(maxKeyLen)
+
+	opts := func() *Options {
+		return (&Options{
+			DisableWAL:         true,
```
itsbilal, line 2748: We shouldn't disable the WAL if the test is frequently closing/opening new DBs while ensuring saved state stays on the same underlying FS. We're already doing NoSync everywhere, so let's at least keep the WAL.

iterator_test.go
```diff
@@ -2722,6 +2724,182 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+		for l := 0; l < len(loadKeyspaces); l++ {
+			l := l
+			grp.Go(func() error {
+				rng := rand.New(rand.NewSource(1))
+				batch := d.NewBatch()
+				key := make([]byte, ks.MaxLen())
+				var val [valSize]byte
+				for i := int64(0); i < loadKeyspaces[l].Count(); i++ {
+					rng.Read(val[:])
+					n := testkeys.WriteKey(key[:], loadKeyspaces[l], i)
```
itsbilal, line 2849: nit: no need to write `key[:]` as `key` is already a slice, `key` alone works.

iterator_test.go
```diff
@@ -2722,6 +2724,182 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+		fn   func(*DB, testkeys.Keyspace, *rand.Rand) error
+	}
+	var iterKeyBuf [maxKeyLen]byte
+
+	iterOps := []iteratorOp{
+		{
+			name: "prefix point lookup", fn: func(d *DB, ks testkeys.Keyspace, rng *rand.Rand) error {
+				n := testkeys.WriteKey(iterKeyBuf[:], ks, int64(rng.Intn(int(ks.Count()))))
+				iter, _ := d.NewIter(nil)
+				_ = iter.SeekPrefixGE(iterKeyBuf[:n])
```
itsbilal, line 2765: One issue with the way these operations are written, is that the dominant thing is going to be iter opening/closing (which grabs version locks, and grabs an object from the pool, etc.) and not the actual operation you're trying to benchmark. Try reusing the iterator and just doing `SeekPrefixGE` etc. `b.N` times?

itsbilal (commented):
Looking great! Just some relatively minor comments.

internal/testkeys/testkeys_test.go
```diff
@@ -65,6 +65,10 @@ func TestKeyCount(t *testing.T) {
 	}
 	testCases := map[params]int64{
 		{26, 1}: 26,
+		{26, 2}: 702,
+		{26, 3}: 18278,
+		{26, 4}: 475254,
+		{26, 5}: 12356630,
```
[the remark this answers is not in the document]
anish-shanbhag: I think this is just testing `keyCount` which uses `n + n^2 + ... + n^l` to calculate the size of the keyspace, rather than actually generating all of the keys themselves. This might still be overkill given we already have the other test cases; I included it mostly as a sanity check.

anish-shanbhag:
Thanks for the suggestions! I've addressed most of the things you mentioned

internal/testkeys/testkeys_test.go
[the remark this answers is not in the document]
itsbilal: Good point. Sounds good to me then.

iterator_test.go
```diff
@@ -2722,6 +2724,178 @@ func BenchmarkSeekPrefixTombstones(b *testing.B) {
...
+		},
+		{
+			name: "seek-ge", fn: func(iter *Iterator, ks testkeys.Keyspace, rng *rand.Rand) {
+				n := testkeys.WriteKey(iterKeyBuf[:], ks, int64(rng.Intn(int(ks.Count()))))
+				_ = iter.SeekGE(iterKeyBuf[:n])
+			},
+		},
+		{
+			name: "iterate", fn: func(iter *Iterator, ks testkeys.Keyspace, rng *rand.Rand) {
+				for valid := iter.First(); valid; valid = iter.Next() {
```
itsbilal, line 2774: This one's a little complex. It feels odd that the runtime per `b.N` is going to vary by how many keys there are. maybe we should only do one operation in here, _just_ a `Next()` and if it returns a non-valid return value, we do a `First()`? That way it should be easier to pick on the difference in runtime of a positioning operation alone.

itsbilal (approved):
LGTM save for one comment

iterator_test.go
anish-shanbhag: Good point - a single `Next()` is much easier to compare across the benchmark than an entire scan. This is updated now.

PR #3792, opened 2024-07-26
db: don't load large filter blocks for flushable ingests

### LINKED ISSUE #3787: don't use filter blocks for large flushable ingests

Jira issue: PEBBLE-225

RaduBerinde:

`filterBlockSizeLimit` value.

Fixes #3787

### DIFF
--- a/sstable/reader_iter_single_lvl.go
+++ b/sstable/reader_iter_single_lvl.go
@@ -157,11 +157,11 @@ type singleLevelIterator struct {
 	// singleLevelIterator, given that these two iterators share this field.
 	exhaustedBounds int8
 
-	// useFilter specifies whether the filter block in this sstable, if present,
-	// should be used for prefix seeks or not. In some cases it is beneficial
-	// to skip a filter block even if it exists (eg. if probability of a match
-	// is high).
-	useFilter              bool
+	// filterBlockSizeLimit controls whether the bloom filter block in this
+	// sstable, if present, should be used for prefix seeks or not (depending on
+	// its size). In some cases it is beneficial to skip a filter block even if it
+	// exists (eg. if probability of a match is high).
+	filterBlockSizeLimit   FilterBlockSizeLimit
 	lastBloomFilterMatched bool
 
 	transforms IterTransforms
@@ -187,7 +187,7 @@ func newSingleLevelIterator(
 	transforms IterTransforms,
 	lower, upper []byte,
 	filterer *BlockPropertiesFilterer,
-	useFilter bool,
+	filterBlockSizeLimit FilterBlockSizeLimit,
 	stats *base.InternalIteratorStats,
 	categoryAndQoS CategoryAndQoS,
 	statsCollector *CategoryStatsCollector,
@@ -199,7 +199,7 @@ func newSingleLevelIterator(
 	}
 	i := singleLevelIterPool.Get().(*singleLevelIterator)
 	i.init(
-		ctx, r, v, transforms, lower, upper, filterer, useFilter,
+		ctx, r, v, transforms, lower, upper, filterer, filterBlockSizeLimit,
 		stats, categoryAndQoS, statsCollector, rp, bufferPool,
 	)
 	indexH, err := r.readIndex(ctx, i.indexFilterRH, stats, &i.iterStats)
@@ -221,7 +221,7 @@ func (i *singleLevelIterator) init(
 	transforms IterTransforms,
 	lower, upper []byte,
 	filterer *BlockPropertiesFilterer,
-	useFilter bool,
+	filterBlockSizeLimit FilterBlockSizeLimit,
 	stats *base.InternalIteratorStats,
 	categoryAndQoS CategoryAndQoS,
 	statsCollector *CategoryStatsCollector,
@@ -233,7 +233,7 @@ func (i *singleLevelIterator) init(
 	i.lower = lower
 	i.upper = upper
 	i.bpfs = filterer
-	i.useFilter = useFilter
+	i.filterBlockSizeLimit = filterBlockSizeLimit
 	i.reader = r
 	i.cmp = r.Compare
 	i.stats = stats
@@ -793,11 +793,11 @@ func (i *singleLevelIterator) SeekPrefixGE(
 			key = i.lower
 		}
 	}
-	return i.seekPrefixGE(prefix, key, flags, i.useFilter)
+	return i.seekPrefixGE(prefix, key, flags, i.filterBlockSizeLimit)
 }
 
 func (i *singleLevelIterator) seekPrefixGE(
-	prefix, key []byte, flags base.SeekGEFlags, checkFilter bool,
+	prefix, key []byte, flags base.SeekGEFlags, filterBlockSizeLimit FilterBlockSizeLimit,
 ) (kv *base.InternalKV) {
 	// NOTE: prefix is only used for bloom filter checking and not later work in
 	// this method. Hence, we can use the existing iterator position if the last
@@ -805,7 +805,7 @@ func (i *singleLevelIterator) seekPrefixGE(
 
 	err := i.err
 	i.err = nil // clear cached iteration error
-	if checkFilter && i.reader.tableFilter != nil {
+	if shouldUseFilterBlock(i.reader, filterBlockSizeLimit) {
 		if !i.lastBloomFilterMatched {
 			// Iterator is not positioned based on last seek.
 			flags = flags.DisableTrySeekUsingNext()
@@ -851,6 +851,12 @@ func (i *singleLevelIterator) seekPrefixGE(
 	i.boundsCmp = 0
 	i.positionedUsingLatestBounds = true
 	return i.maybeVerifyKey(i.seekGEHelper(key, boundsCmp, flags))
+}
+
+// shouldUseFilterBlock returns whether we should use the filter block, based on
+// its length and the size limit.
+func shouldUseFilterBlock(reader *Reader, filterBlockSizeLimit FilterBlockSizeLimit) bool {
+	return reader.tableFilter != nil && reader.filterBH.Length <= uint64(filterBlockSizeLimit)
 }
 
 func (i *singleLevelIterator) bloomFilterMayContain(prefix []byte) (bool, error) {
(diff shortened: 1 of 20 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3792#-:-O2jS_1tAPLwDN4nc5Vr:bg7igkv)* status: all files reviewed, 4 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`internal/manifest/level.go` line 30 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3792#-O2jNjdB5EohlYUls8QY:-O2jNjdB5EohlYUls8QZ:bhalabh) ([raw file](https://github.com/cockroachdb/pebble/blob/75b65e1c79787ca2e8ecaa6a2a17d3f2917aa226/internal/manifest/level.go#L30)):*
> ```Go
> // LevelToInt returns the int representation of a Level. Returns -1 if the Level
> // refers to the flushable ingests pseudo-level.
> func LevelToInt(l Level) int {
> ```

there is a call to `LevelToInt` from `Iterator.sampleRead`, to decide whether a read-driven compaction is warranted. I think we should ignore these flushable ingests in the function called by `mi.ForEachLevelIter`, since we can't do any read-driven compactions starting at these files until after they are flushed.
___
*[`sstable/reader_common.go` line 60 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3792#-O2jOGtDAwLDSdVwrUE1:-O2jOGtDAwLDSdVwrUE2:b17xwst) ([raw file](https://github.com/cockroachdb/pebble/blob/75b65e1c79787ca2e8ecaa6a2a17d3f2917aa226/sstable/reader_common.go#L60)):*
> ```Go
> 	// NeverUseFilterBlock indicates that bloom filter blocks should never be used.
> 	NeverUseFilterBlock FilterBlockSizeLimit = 0
> 	// AlwaysUseFilterBlock indicates that bloom filter blocks shoudl always be
> ```

nit: should
___
*[`sstable/reader_iter_two_lvl.go` line 344 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3792#-O2jOloKAVeKSSR5_vlR:-O2jOloKAVeKSSR5_vlS:b-n0uj4d) ([raw file](https://github.com/cockroachdb/pebble/blob/75b65e1c79787ca2e8ecaa6a2a17d3f2917aa226/sstable/reader_iter_two_lvl.go#L344)):*
> ```Go
> 	i.err = nil // clear cached iteration error
> 
> 	useFilter := i.shouldUseBloomFilter(i.filterBlockSizeLimit)
> ```

It is probably worth adding a comment to `shouldUseBloomFilter` that its return value must not change for an iterator. We depend on this behavior here in the calculation of `filterUsedAndDidNotMatch`.

Alternatively, given we know whether `tableFilter` is non-nil at the time of iter construction, and we know the size, should we just compute the bool value of `useFilter` once? Not a strong opinion, so whatever you think is best and easy to backport.
___
*[`testdata/flushable_ingest` line 669 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3792#-O2jRGEe9HcBcMUyfmP0:-O2jRGEe9HcBcMUyfmP1:b-77ke6d) ([raw file](https://github.com/cockroachdb/pebble/blob/75b65e1c79787ca2e8ecaa6a2a17d3f2917aa226/testdata/flushable_ingest#L669)):*
> ```
> small-00001-does-not-exist: pebble: not found
> 
> # When looking inside the small table, we will not read the bloom filter which
> ```

large table?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3792#-:-O2ji5Rl8akT7z-Abpgs:b-814g8k)* status: 15 of 20 files reviewed, 3 unresolved discussions (waiting on @itsbilal, @jbowens, and @sumeerbhola)

___
*[`internal/manifest/level.go` line 30 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3792#-O2jNjdB5EohlYUls8QY:-O2jiGifDwDPC9uc0pXI:bpqj741) ([raw file](https://github.com/cockroachdb/pebble/blob/75b65e1c79787ca2e8ecaa6a2a17d3f2917aa226/internal/manifest/level.go#L30)):*

there is a call to `LevelToInt` from `Iterator.sampleRead`, to decide whether a read-driven compaction is warranted. I think we should ignore these flushable ingests in the function called by `mi.ForEachLevelIter`, since we can't do any read-driven compactions starting at these files until after they are flushed.
</blockquote></details>

Great catch, thanks!
___
*[`sstable/reader_iter_two_lvl.go` line 344 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3792#-O2jOloKAVeKSSR5_vlR:-O2jilQvBL6N1pNwCCFT:bj2z1og) ([raw file](https://github.com/cockroachdb/pebble/blob/75b65e1c79787ca2e8ecaa6a2a17d3f2917aa226/sstable/reader_iter_two_lvl.go#L344)):*

It is probably worth adding a comment to `shouldUseBloomFilter` that its return value must not change for an iterator. We depend on this behavior here in the calculation of `filterUsedAndDidNotMatch`.

Alternatively, given we know whether `tableFilter` is non-nil at the time of iter construction, and we know the size, should we just compute the bool value of `useFilter` once? Not a strong opinion, so whatever you think is best and easy to backport.
</blockquote></details>

I changed `shouldUseBloomFilter` to not depend on the iterator take in the `*Reader` instead.

I agree on calculating `useFilter` upfront, and I will look to follow up with some cleanup, but I think it might make things harder to backport (it's a bit subtle because the single level iterator behaves differently depending if it's under a two level iterator).
___
*[`testdata/flushable_ingest` line 669 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3792#-O2jRGEe9HcBcMUyfmP0:-O2jjpuXDzAapgeRzdQB:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/75b65e1c79787ca2e8ecaa6a2a17d3f2917aa226/testdata/flushable_ingest#L669)):*

large table?
</blockquote></details>

Done.

sumeerbhola (approved):
:lgtm: 

> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3792#-:-O2jmUTt0ix89_1GTzLQ:bg7igkv)* status: all files reviewed (commit messages unreviewed), 3 unresolved discussions (waiting on @itsbilal and @jbowens)

sstable/reader_iter_single_lvl.go
```diff
@@ -793,19 +793,19 @@ func (i *singleLevelIterator) SeekPrefixGE(
...
+	prefix, key []byte, flags base.SeekGEFlags, filterBlockSizeLimit FilterBlockSizeLimit,
 ) (kv *base.InternalKV) {
 	// NOTE: prefix is only used for bloom filter checking and not later work in
 	// this method. Hence, we can use the existing iterator position if the last
 	// SeekPrefixGE did not fail bloom filter matching.
 
 	err := i.err
 	i.err = nil // clear cached iteration error
-	if checkFilter && i.reader.tableFilter != nil {
+	if shouldUseBloomFilter(i.reader, filterBlockSizeLimit) {
```
itsbilal, line 808: supernit: we generally just use the term `filter`, not bloom filter, just to be impl-agnostic. So `shouldUseFilter` sounds better.

testdata/flushable_ingest
```diff
@@ -618,3 +618,70 @@ L0.1:
...
+build large
+set-multiple 100000 large
+----
+
+batch
+set small-00001 before-ingest
+set large-00001 before-ingest
+----
+
+blockFlush
```
itsbilal, line 641: nit: maybe the blockFlush can go before the batch commit, that way we are guaranteed we wouldn't flush right after the batch?

testdata/flushable_ingest
```diff
@@ -618,3 +618,70 @@ L0.1:
...
+# reads.
+get with-fs-logging
+small-00001-does-not-exist
+----
+read-at(194, 27): 000004.sst
+read-at(120, 74): 000004.sst
+small-00001-does-not-exist: pebble: not found
+
+# When looking inside the large table, we will not read the bloom filter which
+# is large. We read the top-level index block, a second-level index block, and a
```
itsbilal, line 670: nice

itsbilal (approved):
LGTM

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3792#-:-O2kyqpvB-p5MZc7cAC7:b-814g8k)* status: 14 of 20 files reviewed, 6 unresolved discussions (waiting on @itsbilal, @jbowens, and @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 808 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3792#gh-1693579712:-O2kxcClDDKadWIxm4JZ:b-j877ia) ([raw file](https://github.com/cockroachdb/pebble/blob/110448dfb42b778083d8b9558e3796b7c6a3a89c/sstable/reader_iter_single_lvl.go#L808)):*

supernit: we generally just use the term `filter`, not bloom filter, just to be impl-agnostic. So `shouldUseFilter` sounds better.
</blockquote></details>

Renamed to `shouldUseFilterBlock` (which is more consistent with the argument too).

I sometimes encounter confusion between block property filtering and bloom filtering. They're both called filters and sometimes it's not immediately obvious which one it is when looking at unfamiliar code. "filter block" instead of just "filter" helps.
___
*[`testdata/flushable_ingest` line 641 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3792#gh-1693582512:-O2kyNZR1gZowDU90Od-:b-8alql3) ([raw file](https://github.com/cockroachdb/pebble/blob/110448dfb42b778083d8b9558e3796b7c6a3a89c/testdata/flushable_ingest#L641)):*

nit: maybe the blockFlush can go before the batch commit, that way we are guaranteed we wouldn't flush right after the batch?
</blockquote></details>

I'm following the existing tests which do this. I think they're reasonably assuming we wouldn't flush a tiny number of keys.

RaduBerinde:
Backport PRs below. All of them required manual merging.

https://github.com/cockroachdb/pebble/pull/3798
https://github.com/cockroachdb/pebble/pull/3799
https://github.com/cockroachdb/pebble/pull/3800
https://github.com/cockroachdb/pebble/pull/3801

Issue #3794, opened 2024-07-26
db: strictly enforce prefix in internalIterator.SeekPrefixGE


jbowens:
We should tighten the internal iterator interface to require that if SeekPrefixGE returns a key, the key is guaranteed to have the provided prefix. Every key we avoid propagating up the iterator stack avoids a key comparison during initialization of the merging iterator heap. Some code already requires that SeekPrefixGE strict, necessitating the ToplevelIterator interface with its separate SeekPrefixGEStrict method. Enforcing the prefix at the leaves will simplify this.

In columnar blocks, enforcing the prefix strictly comes for free because we must first compare against the prefix alone.

https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1721918197337849

RaduBerinde:
This is done (and taken advantage of in iterv2).

PR #3795, opened 2024-07-26
db: strictly enforce prefix in [flushable]BatchIter.SeekPrefixGE

jbowens:
Every key we avoid propagating up the iterator stack avoids a key comparison
during initialization of the merging iterator heap.

Informs #3794.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -5,6 +5,7 @@
 package pebble
 
 import (
+	"bytes"
 	"context"
 	"encoding/binary"
 	"fmt"
@@ -1686,7 +1687,16 @@ func (i *batchIter) SeekGE(key []byte, flags base.SeekGEFlags) *base.InternalKV 
 }
 
 func (i *batchIter) SeekPrefixGE(prefix, key []byte, flags base.SeekGEFlags) *base.InternalKV {
-	return i.SeekGE(key, flags)
+	kv := i.SeekGE(key, flags)
+	if kv == nil {
+		return nil
+	}
+	// If the key doesn't have the sought prefix, return nil.
+	if !bytes.Equal(i.batch.comparer.Split.Prefix(kv.K.UserKey), prefix) {
+		i.kv = base.InternalKV{}
+		return nil
+	}
+	return kv
 }
 
 func (i *batchIter) SeekLT(key []byte, flags base.SeekLTFlags) *base.InternalKV {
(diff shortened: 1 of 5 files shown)


RaduBerinde (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3795)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`batch.go` line 1692 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3795#-O2kdUQe71zBbWmJIu6R:-O2kdUQe71zBbWmJIu6S:b-59o439) ([raw file](https://github.com/cockroachdb/pebble/blob/405410e9c072bdae7d571bf61594d6cb9abccf26/batch.go#L1692)):*
> ```Go
> 	kv := i.SeekGE(key, flags)
> 	if kv == nil {
> 		return kv
> ```

Don't we need to set `i.kv` here too? Also, `return nil`. Maybe consolidate the two blocks.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3795)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`batch.go` line 1692 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3795#-O2kdUQe71zBbWmJIu6R:-O2kh4P1546NMEB8AG-y:b5q4i44) ([raw file](https://github.com/cockroachdb/pebble/blob/405410e9c072bdae7d571bf61594d6cb9abccf26/batch.go#L1692)):*

Don't we need to set `i.kv` here too? Also, `return nil`. Maybe consolidate the two blocks.
</blockquote></details>

i.SeekGE already did it

jbowens:
TFTR!

RaduBerinde:
*[`batch.go` line 1692 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3795#-O2kdUQe71zBbWmJIu6R:-O2kz0X_8fjXOh2J2M2t:bnkvyot) ([raw file](https://github.com/cockroachdb/pebble/blob/405410e9c072bdae7d571bf61594d6cb9abccf26/batch.go#L1692)):*

i.SeekGE already did it
</blockquote></details>

Ah, right, LGTM!

PR #3797, opened 2024-07-27
manifest: clean up Level

RaduBerinde:
Level can now also refer to flushable ingests. This change renames it
to `Layer` and makes it easier and less error-prone to use.

This is a follow-up to #3792.


jbowens (approved):
"layer" is a _much_ better name!

:lgtm:

> Reviewed 18 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3797#-:-O2yyz5j17eKGfYqGf92:b61bx7g)* status: all files reviewed, 3 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`level_iter.go` line 64 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3797#-O2yyBcyCzFh-DmtZiVm:-O2yyBcyCzFh-DmtZiVn:b-n3w49n) ([raw file](https://github.com/cockroachdb/pebble/blob/138b70603c988ebaa98814ef095f1869f33bebaa/level_iter.go#L64)):*
> ```Go
> 	tableOpts IterOptions
> 	// The LSM level this levelIter is initialized for.
> 	level manifest.Layer
> ```

nit: rename `levelIter.level` too?
___
*[`options.go` line 201 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3797#-O2yyG55CIFTdQmFErjI:-O2yyG55CIFTdQmFErjJ:b-5v3ga9) ([raw file](https://github.com/cockroachdb/pebble/blob/138b70603c988ebaa98814ef095f1869f33bebaa/options.go#L201)):*
> ```Go
> 	// Level corresponding to this file. Only passed in if constructed by a
> 	// levelIter.
> 	level manifest.Layer
> ```

nit: ditto?
___
*[`internal/manifest/layer.go` line 25 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3797#-O2yyQ5l-EZssVGrlXln:-O2yyQ5l-EZssVGrlXlo:b3lbo6h) ([raw file](https://github.com/cockroachdb/pebble/blob/138b70603c988ebaa98814ef095f1869f33bebaa/internal/manifest/layer.go#L25)):*
> ```Go
> 
> // Level returns a Layer that represents an entire level (L0 through L6).
> func Level(level int) Layer {
> ```

nano nit: should we suffix with "Layer" like we do for "FlushableIngestsLayer"?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3797#-:-O2zhHoZCzR6a9uAtJYe:b-814g8k)* status: 10 of 20 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`internal/manifest/layer.go` line 25 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3797#-O2yyQ5l-EZssVGrlXln:-O2zhqKl27Hvu_rPf_sN:brdyqxw) ([raw file](https://github.com/cockroachdb/pebble/blob/138b70603c988ebaa98814ef095f1869f33bebaa/internal/manifest/layer.go#L25)):*

nano nit: should we suffix with "Layer" like we do for "FlushableIngestsLayer"?
</blockquote></details>

There are a lot of usages and I think it's not necessary because "level" is already clearly identifying a layer in the hierarchy, whereas "flushable ingests" is more generic.

PR #3798, opened 2024-07-27
crl-release-24.2: db: don't load large filter blocks for flushable ingests

### LINKED ISSUE #3787: don't use filter blocks for large flushable ingests

RaduBerinde:

`filterBlockSizeLimit` value.

Fixes #3787

### DIFF
--- a/testdata/flushable_ingest
+++ b/testdata/flushable_ingest
@@ -618,3 +618,70 @@ L0.1:
   000004:[a#12,SET-b#12,SET]
 L0.0:
   000007:[a#10,SET-b#11,SET]
+
+close
+----
+
+reset
+----
+
+build small
+set-multiple 10 small
+----
+
+build large
+set-multiple 100000 large
+----
+
+batch
+set small-00001 before-ingest
+set large-00001 before-ingest
+----
+
+blockFlush
+----
+
+ingest small large
+----
+
+allowFlush
+----
+
+# When looking inside the small table, we will read the index block, then we
+# should read the bloom filter, followed by a data block read.
+get with-fs-logging
+small-00001
+----
+read-at(194, 27): 000004.sst
+read-at(120, 74): 000004.sst
+read-at(0, 120): 000004.sst
+small-00001:val-00001
+
+# When the key doesn't pass the bloom filter, we should see only two block
+# reads.
+get with-fs-logging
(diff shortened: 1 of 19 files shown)


testdata/flushable_ingest
```diff
@@ -618,3 +618,70 @@ L0.1:
...
+# reads.
+get with-fs-logging
+small-00001-does-not-exist
+----
+read-at(194, 27): 000004.sst
+read-at(120, 74): 000004.sst
+small-00001-does-not-exist: pebble: not found
+
+# When looking inside the large table, we will not read the bloom filter which
+# is large. We read the top-level index block, a second-level index block, and a
```
itsbilal, line 670: I'm catching this late, but the first read is probably a footer or meta-index read and not a two-level filter block read, right? Either way, doesn't block this PR.

itsbilal (approved):
LGTM

jbowens (approved):

RaduBerinde (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3798#-:-O2z-tZw80BSpstceJFA:bz5mair)* status: 0 of 19 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @sumeerbhola)

___
*[`testdata/flushable_ingest` line 670 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3798#gh-1695382821:-O2z-7L62WOCZpyRib0P:b-n1she1) ([raw file](https://github.com/cockroachdb/pebble/blob/c723b0ff500fdc11b8c5c508dfe2ff0287b52eab/testdata/flushable_ingest#L670)):*

I'm catching this late, but the first read is probably a footer or meta-index read and not a two-level filter block read, right? Either way, doesn't block this PR.
</blockquote></details>

Hm, I don't think so, we don't see the footer/meta-index read in the small case either. I expect that we already have a Reader in the table cache, these are the reads that are caused by creating and using an iterator. If I make the table larger, we see this diff:
```diff
-read-at(775373, 117): 000005.sst
-read-at(765608, 2268): 000005.sst
+read-at(7783655, 911): 000005.sst
+read-at(7683904, 2164): 000005.sst
```

PR #3799, opened 2024-07-27
crl-release-24.1: db: don't load large filter blocks for flushable ingests

### LINKED ISSUE #3787: don't use filter blocks for large flushable ingests

RaduBerinde:

`filterBlockSizeLimit` value.

Fixes #3787

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -409,6 +409,23 @@ func runBatchDefineCmd(d *datadriven.TestData, b *Batch) error {
 				return errors.Errorf("%s expects 2 arguments", parts[0])
 			}
 			err = b.Set([]byte(parts[1]), []byte(parts[2]), nil)
+
+		case "set-multiple":
+			if len(parts) != 3 {
+				return errors.Errorf("%s expects 2 arguments (n and prefix)", parts[0])
+			}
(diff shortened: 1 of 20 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3799)* status: 0 of 20 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`testdata/flushable_ingest` line 655 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3799#-O2p3caNBTs0caaEvoV7:-O2p3caNBTs0caaEvoV8:b-519dyv) ([raw file](https://github.com/cockroachdb/pebble/blob/5e24488ffbe82c75f1335fd4d65aa5fdd87bcca3/testdata/flushable_ingest#L655)):*
> ```
> # The extra index block reads are due to inefficiencies of the iterator
> # implementations (fixed in later versions).
> iter with-fs-logging
> ```

I had to switch to `iter` because the `getIter` wasn't doing seek-prefix-ge for the ingested flushables.

jbowens (approved):

PR #3805, opened 2024-07-29
sstable: clean up use-filter-block logic

RaduBerinde:
#### sstable: clean up use-filter-block logic

We now store `useFilterBlock` in both the single level and the two
level iterators. When the single level iterator is part of a two-level
iterator, the "inner" `useFilterBlock` is always false.

### DIFF
--- a/sstable/data_test.go
+++ b/sstable/data_test.go
@@ -417,7 +417,7 @@ func runIterCmd(
 			fmt.Fprintf(&b, "| %T:\n", origIter)
 			si, _ := origIter.(*singleLevelIterator)
 			if twoLevelIter, ok := origIter.(*twoLevelIterator); ok {
-				si = &twoLevelIter.singleLevelIterator
+				si = &twoLevelIter.secondLevel
 				if twoLevelIter.topLevelIndex.Valid() {
 					fmt.Fprintf(&b, "|  topLevelIndex.Key() = %q\n", twoLevelIter.topLevelIndex.Key())
 					v := twoLevelIter.topLevelIndex.Value()
--- a/sstable/reader.go
+++ b/sstable/reader.go
@@ -268,7 +268,7 @@ func (r *Reader) newCompactionIter(
 		if err != nil {
 			return nil, err
 		}
-		i.setupForCompaction()
+		i.SetupForCompaction()
 		return i, nil
 	}
 	i, err := newSingleLevelIterator(
@@ -278,7 +278,7 @@ func (r *Reader) newCompactionIter(
 	if err != nil {
 		return nil, err
 	}
-	i.setupForCompaction()
+	i.SetupForCompaction()
 	return i, nil
 }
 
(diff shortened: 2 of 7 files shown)


sumeerbhola (approved):
:lgtm: 

> Reviewed 7 of 7 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3805#-:-O3-zZmWBb2TKbpnVwGv:bg7igkv)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`sstable/reader_iter_two_lvl.go` line 22 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3805#-O3-z0-H11g7CVg5PmKM:-O3-z0-IAOpu_iAdbgI1:bm6i9ao) ([raw file](https://github.com/cockroachdb/pebble/blob/6718b2d83b30808a2e1d4c96f7e8ad49ec923569/sstable/reader_iter_two_lvl.go#L22)):*
> ```Go
> 	secondLevel            singleLevelIterator
> 	topLevelIndex          rowblk.Iter
> 	useFilterBlock         bool
> ```

It's great to see this un-embedding, but could you provide some context on why we need to mirror `useFilterBlock` and `lastBloomFilterMatched` in `twoLevelIterator`? Is it because it is semantically cleaner in that these fields don't say anything about the second level in the context of `twoLevelIterator` so we shouldn't be reusing the `secondLevel` fields?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3805)* status: 5 of 7 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`sstable/reader_iter_two_lvl.go` line 22 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3805#-O3-z0-H11g7CVg5PmKM:-O301reUBdePXPLI3rn1:bpozbsh) ([raw file](https://github.com/cockroachdb/pebble/blob/6718b2d83b30808a2e1d4c96f7e8ad49ec923569/sstable/reader_iter_two_lvl.go#L22)):*

It's great to see this un-embedding, but could you provide some context on why we need to mirror `useFilterBlock` and `lastBloomFilterMatched` in `twoLevelIterator`? Is it because it is semantically cleaner in that these fields don't say anything about the second level in the context of `twoLevelIterator` so we shouldn't be reusing the `secondLevel` fields?
</blockquote></details>

`secondLevel.useFilterBlock` is always `false`: we don't want `secondLevel` to ever do any bloom filtering since we are doing it at the top level. Added a comment.

`lastBloomFilterMatched` was just preference given that all the code that manipulates it in the two-level case is in this file, and logically belongs with `useFilterBlock.

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3805#-:-O32f78688uPuSNXEa75:bg7igkv)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

PR #3808, opened 2024-07-30
manifest: fix incorrect Annotator pointer aggregation

anish-shanbhag:
cockroachdb#3760 contained a bug which causes Annotator
values to be incorrectly aggregated when pointer values
should be overwritten. This is because the `vtyped`
variable was not being modified due to being on the
stack.

This change fixes this and adds a unit test for
`PickFileAggregator` to catch this issue in the future.

cockroachdb#3759 should already not be affected by
this due to the different way it handles aggregation.

### DIFF
--- a/internal/manifest/annotator.go
+++ b/internal/manifest/annotator.go
@@ -84,33 +84,36 @@ func (a *Annotator[T]) findAnnotation(n *node) *annotation {
 // nodeAnnotation computes this annotator's annotation of this node across all
 // files in the node's subtree. The second return value indicates whether the
 // annotation is stable and thus cacheable.
-func (a *Annotator[T]) nodeAnnotation(n *node) (v *T, cacheOK bool) {
+func (a *Annotator[T]) nodeAnnotation(n *node) (_ *T, cacheOK bool) {
 	annot := a.findAnnotation(n)
-	vtyped := annot.v.(*T)
+	t := annot.v.(*T)
 	// If the annotation is already marked as valid, we can return it without
 	// recomputing anything.
 	if annot.valid {
-		return vtyped, true
+		return t, true
 	}
 
-	annot.v = a.Aggregator.Zero(vtyped)
-	annot.valid = true
+	t = a.Aggregator.Zero(t)
+	valid := true
 
 	for i := int16(0); i <= n.count; i++ {
 		if !n.leaf {
 			v, ok := a.nodeAnnotation(n.children[i])
-			annot.v = a.Aggregator.Merge(v, vtyped)
-			annot.valid = annot.valid && ok
+			t = a.Aggregator.Merge(v, t)
+			valid = valid && ok
 		}
 
 		if i < n.count {
-			v, ok := a.Aggregator.Accumulate(n.items[i], vtyped)
-			annot.v = v
-			annot.valid = annot.valid && ok
+			var ok bool
+			t, ok = a.Aggregator.Accumulate(n.items[i], t)
+			valid = valid && ok
 		}
 	}
 
-	return annot.v.(*T), annot.valid
+	annot.v = t
+	annot.valid = valid
+
+	return t, annot.valid
 }
 
 // InvalidateAnnotation removes any existing cached annotations from this
--- a/internal/manifest/annotator_test.go
+++ b/internal/manifest/annotator_test.go
@@ -56,3 +56,26 @@ func BenchmarkNumFilesAnnotator(b *testing.B) {
 		require.EqualValues(b, uint64(i), numFiles)
 	}
 }
+
+func TestPickFileAggregator(t *testing.T) {
+	const count = 1000
+	a := Annotator[FileMetadata]{
+		Aggregator: PickFileAggregator{
+			Filter: func(f *FileMetadata) (eligible bool, cacheOK bool) {
+				return true, true
+			},
+			Compare: func(f1 *FileMetadata, f2 *FileMetadata) bool {
+				return base.DefaultComparer.Compare(f1.Smallest.UserKey, f2.Smallest.UserKey) < 0
+			},
+		},
+	}
+
+	lm, files := makeTestLevelMetadata(1)
+
+	for i := 1; i <= count; i++ {
+		lm.tree.Insert(newItem(key(i)))
+		pickedFile := a.LevelAnnotation(lm)
+		// The picked file should always be the one with the smallest key.
+		require.Same(t, files[0], pickedFile)
+	}
+}


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3808)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @anish-shanbhag and @itsbilal)

___
*[`internal/manifest/annotator.go` line 89 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3808#-O33UKpP-I73DADVXbwF:-O33UKpP-I73DADVXbwG:b-6ucqha) ([raw file](https://github.com/cockroachdb/pebble/blob/4f062cb7188edd4343c2487c61f70f474dee2fb5/internal/manifest/annotator.go#L89)):*
> ```Go
> func (a *Annotator[T]) nodeAnnotation(n *node) (v *T, cacheOK bool) {
> 	annot := a.findAnnotation(n)
> 	// If the annotation is already marked as valid, we can return it without
> ```

I don't see any functional difference in this change, can you explain more?

anish-shanbhag (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3808)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`internal/manifest/annotator.go` line 89 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3808#-O33UKpP-I73DADVXbwF:-O33fWt_4dDK2Q7muHjs:b-lfbmks) ([raw file](https://github.com/cockroachdb/pebble/blob/4f062cb7188edd4343c2487c61f70f474dee2fb5/internal/manifest/annotator.go#L89)):*

I don't see any functional difference in this change, can you explain more?
</blockquote></details>

The logic here is meant to merge annotations from child nodes/files into `annot.v`, which stores the current node's annotation value. Since `vtyped` was a stack variable, it created a copy of the annotation value rather than referencing `annot.v` itself. So when we pass `vtyped` in `a.Aggregator.Merge(v, vtyped)` or `a.Aggregator.Accumulate(n.items[i], vtyped)` it always passes the initial value of `annot.v` when `vtyped` was defined.

This wasn't caught before because the annotator unit tests all used `SumAnnotator` which only sets the annotation value to a `*uint64` pointer once, and just modifies the underlying int value thereafter. On the other hand, `PickFileAggregator` does modify `annot.v` to point to a different `*fileMetadata`.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3808)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @anish-shanbhag and @itsbilal)

___
*[`internal/manifest/annotator.go` line 89 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3808#-O33UKpP-I73DADVXbwF:-O33iPxa-bRHR9ansTMu:bpho3ag) ([raw file](https://github.com/cockroachdb/pebble/blob/4f062cb7188edd4343c2487c61f70f474dee2fb5/internal/manifest/annotator.go#L89)):*

The logic here is meant to merge annotations from child nodes/files into `annot.v`, which stores the current node's annotation value. Since `vtyped` was a stack variable, it created a copy of the annotation value rather than referencing `annot.v` itself. So when we pass `vtyped` in `a.Aggregator.Merge(v, vtyped)` or `a.Aggregator.Accumulate(n.items[i], vtyped)` it always passes the initial value of `annot.v` when `vtyped` was defined.

This wasn't caught before because the annotator unit tests all used `SumAnnotator` which only sets the annotation value to a `*uint64` pointer once, and just modifies the underlying int value thereafter. On the other hand, `PickFileAggregator` does modify `annot.v` to point to a different `*fileMetadata`.
</blockquote></details>

It created a copy of the pointer `*T` and then we passed the pointer, I don't see the problem.

There is no difference between using `<anything can go here>` directly and doing `v := <anything can go here>` and then using `v`.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3808)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @anish-shanbhag and @itsbilal)

___
*[`internal/manifest/annotator.go` line 89 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3808#-O33UKpP-I73DADVXbwF:-O33s1Iv1CKN7GcEcAS7:bhr812t) ([raw file](https://github.com/cockroachdb/pebble/blob/4f062cb7188edd4343c2487c61f70f474dee2fb5/internal/manifest/annotator.go#L89)):*

Makes sense. I think the difference here is that we may modify `annot.v` to a different `*T` pointer on these lines:
```
annot.v = a.Aggregator.Merge(v, vtyped)
```
```
annot.v = v
```

For example, if we start with `annot.v = (*fileMetadata) 0x2`, then `vtyped` also starts as `(*fileMetadata) 0x2`. If `a.Aggregator.Merge(v, vtyped)` returns `(*fileMetadata) 0x3`, then `annot.v` will be overwritten to `0x3` but `vtyped` will stay as `0x2`.

When accumulating/merging in future loop iterations, we want to use this new `annot.v`. However, `vtyped` will still be equal to the old `annot.v`.

I can definitely see how this is confusing, so open to suggestions on alternate naming/logic which might be more clear.
</blockquote></details>

I see, I would do it this way:

```diff
-func (a *Annotator[T]) nodeAnnotation(n *node) (v *T, cacheOK bool) {
+func (a *Annotator[T]) nodeAnnotation(n *node) (_ *T, cacheOK bool) {
        annot := a.findAnnotation(n)
-       vtyped := annot.v.(*T)
+       t := annot.v.(*T)
        // If the annotation is already marked as valid, we can return it without
        // recomputing anything.
        if annot.valid {
-               return vtyped, true
+               return t, true
        }
 
-       annot.v = a.Aggregator.Zero(vtyped)
-       annot.valid = true
+       t = a.Aggregator.Zero(t)
+       valid := true
 
        for i := int16(0); i <= n.count; i++ {
                if !n.leaf {
                        v, ok := a.nodeAnnotation(n.children[i])
-                       annot.v = a.Aggregator.Merge(v, vtyped)
-                       annot.valid = annot.valid && ok
+                       t = a.Aggregator.Merge(v, t)
+                       valid = valid && ok
                }
 
                if i < n.count {
-                       v, ok := a.Aggregator.Accumulate(n.items[i], vtyped)
-                       annot.v = v
-                       annot.valid = annot.valid && ok
+                       var ok bool
+                       t, ok = a.Aggregator.Accumulate(n.items[i], t)
+                       valid = valid && ok
                }
        }
 
-       return annot.v.(*T), annot.valid
+       annot.v = t
+       annot.valid = valid
+
+       return t, valid
 }
```

anish-shanbhag (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3808)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`internal/manifest/annotator.go` line 89 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3808#-O33UKpP-I73DADVXbwF:-O33wKoY-vxNaGkADLdu:bvt4ec7) ([raw file](https://github.com/cockroachdb/pebble/blob/4f062cb7188edd4343c2487c61f70f474dee2fb5/internal/manifest/annotator.go#L89)):*

I see, I would do it this way:

```diff
-func (a *Annotator[T]) nodeAnnotation(n *node) (v *T, cacheOK bool) {
+func (a *Annotator[T]) nodeAnnotation(n *node) (_ *T, cacheOK bool) {
        annot := a.findAnnotation(n)
-       vtyped := annot.v.(*T)
+       t := annot.v.(*T)
        // If the annotation is already marked as valid, we can return it without
        // recomputing anything.
        if annot.valid {
-               return vtyped, true
+               return t, true
        }
 
-       annot.v = a.Aggregator.Zero(vtyped)
-       annot.valid = true
+       t = a.Aggregator.Zero(t)
+       valid := true
 
        for i := int16(0); i <= n.count; i++ {
                if !n.leaf {
                        v, ok := a.nodeAnnotation(n.children[i])
-                       annot.v = a.Aggregator.Merge(v, vtyped)
-                       annot.valid = annot.valid && ok
+                       t = a.Aggregator.Merge(v, t)
+                       valid = valid && ok
                }
 
                if i < n.count {
-                       v, ok := a.Aggregator.Accumulate(n.items[i], vtyped)
-                       annot.v = v
-                       annot.valid = annot.valid && ok
+                       var ok bool
+                       t, ok = a.Aggregator.Accumulate(n.items[i], t)
+                       valid = valid && ok
                }
        }
 
-       return annot.v.(*T), annot.valid
+       annot.v = t
+       annot.valid = valid
+
+       return t, valid
 }
```
</blockquote></details>

Looks much cleaner - thanks for the suggestion. This is pushed now

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3808#-:-O385bso3_6oB5aKUVpa:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @itsbilal)

PR #3809, opened 2024-07-30
base: add a Comparer checker

RaduBerinde:
We add a `CheckComparer` function which tests the `Compare` and
`Split` functions on combinations of prefixes and suffixes.

The cockroach comparer currently fails this test, as `Split()` on a
suffix incorrectly returns the suffix length instead of 0.

### DIFF
--- a/internal/base/comparer.go
+++ b/internal/base/comparer.go
@@ -6,10 +6,14 @@ package base
 
 import (
 	"bytes"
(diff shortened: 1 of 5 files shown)


RaduBerinde:
Closing for now - there is an issue to be addressed around calling Split on a "bare" suffix.

PR #3814, opened 2024-07-31
base: add CompareSuffixes to Comparer

RaduBerinde:
#### keyspan: use slices.Sort functions

This commit replaces `sort.Interface` implementations with calls to
the newer slices.Sort functions. This is much cleaner and should be a
bit more efficient.

#### base: add CompareSuffixes to Comparer

We currently use `Compare` to compare both keys and bare suffixes.
This makes the implementation more complicated as it has to handle
both cases. It also prevents adding many useful assertions - for
example, we can't verify that `Compare` does byte-wise comparison of
prefixes because we can't unconditionally call `Split`.

We add a mandatory `CompareSuffixes` method that is used when
comparing bare suffixes. The result of `Compare` can be defined
completely in terms of `Split` and `CompareSuffixes`; `Compare` now
has a default implementation.

We also add a `CheckComparer` testing facility that can be used to
verify `Split` and `Compare` on combinations of prefixes and suffixes.
We use it to add a test to all `Comparer`s and it will be used from
CRDB code as well.

Informs https://github.com/cockroachdb/cockroach/issues/127914


jbowens (approved):
:lgtm:

nice!

> Reviewed 14 of 14 files at r1, 27 of 33 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3814#-:-O38LFgH92_TIRY3McCg:blzn4tv)* status: 33 of 39 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

RaduBerinde:
TFTR!

PR #3815, opened 2024-07-31
sstable/block: refactor CompressAndChecksum

jbowens:
This commit moves logic around finishing a block (compressing and checksumming) into the block package. The logic is refactored to use a PhysicalBlock type that distinguishes a finished block from an unfinished, trailer-less block.


RaduBerinde (approved):

jbowens:
TFTR!

PR #3816, opened 2024-07-31
sstable/rowblk: extract block-level suffix rewriting

jbowens:
Move the suffix-rewriting logic that's specific to row-based blocks into a
rowblk.Rewriter type. In future work, a columnar equivalent will be introduced.


jbowens:
TFTR!

PR #3818, opened 2024-08-01
devtools: use separate module for dev tool dependencies

RaduBerinde:
We currently use `go run` directly to run tools like `crlfmt`,
`golint`, etc. This is convenient but there are a few downsides:
 - there is no centralized place which has the tool versions;
 - `go run` requires an internet connection every time it is run;
 - it's more tedious to update the versions (compared to `go get`).

This change adds an `internal/devtools` module with a bogus `.go` file
with import dependencies on the tools. We use `go install` in this
repository to install the correct versions of the tools, and then we
run them directly. This scheme is documented here:
https://play-with-go.dev/tools-as-dependencies_go119_en/


RaduBerinde:
Adding a few more folks to make sure there are no objections.

jbowens (approved):
Cool. No objection from me. It looks like the code coverage CI job is unsatisfied

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3818#-:-O3ELfcoDvHr_s4RAGrj:b-uvl99i)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aadityasondhi and @itsbilal)

RaduBerinde:
TFTR!  Yeah, the coverage job is trying to test the `devtools` "package". It shouldn't happen with PRs that don't touch that path.

PR #3819, opened 2024-08-01
colblk: add data block builder, iterator

jbowens:
Introduce facilities for reading and writing data blocks using user-specified columns for user keys. The full data block schema is not yet complete (eg, the 'is obsolete' bit, etc has not yet been integrated).

### DIFF
--- a/internal/crdbtest/crdbtest.go
+++ b/internal/crdbtest/crdbtest.go
@@ -132,14 +132,14 @@ func EncodeTimestamp(key []byte, walltime uint64, logical uint32) []byte {
 // DecodeTimestamp decodes a MVCC timestamp from a serialized MVCC key.
 func DecodeTimestamp(mvccKey []byte) ([]byte, []byte, uint64, uint32) {
 	tsLen := int(mvccKey[len(mvccKey)-1])
-	keyPartEnd := len(mvccKey) - 1 - tsLen
+	keyPartEnd := len(mvccKey) - tsLen
 	if keyPartEnd < 0 {
 		return nil, nil, 0, 0
 	}
 
 	key := mvccKey[:keyPartEnd]
 	if tsLen > 0 {
-		ts := mvccKey[keyPartEnd+1 : len(mvccKey)-1]
+		ts := mvccKey[keyPartEnd : len(mvccKey)-1]
 		switch len(ts) {
 		case 8:
 			return key, nil, binary.BigEndian.Uint64(ts[:8]), 0
--- a/sstable/colblk/block.go
+++ b/sstable/colblk/block.go
@@ -321,6 +321,11 @@ func (r *BlockReader) Init(data []byte, customHeaderSize uint32) {
 	}
 }
 
+// Rows returns the number of rows in the block, as indicated by the block header.
+func (r *BlockReader) Rows() int {
+	return int(r.header.Rows)
+}
+
 // DataType returns the data type of the col'th column. Every column's data type
 // is encoded within the block header.
 func (r *BlockReader) DataType(col int) DataType {
@@ -423,6 +428,8 @@ func (r *BlockReader) columnToBinFormatter(f *binfmt.Formatter, col, rows int) {
 		bitmapToBinFormatter(f, rows)
 	case DataTypeUint8, DataTypeUint16, DataTypeUint32, DataTypeUint64:
 		uintsToBinFormatter(f, rows, dataType, nil)
+	case DataTypePrefixBytes:
+		prefixBytesToBinFormatter(f, rows, nil)
 	case DataTypeBytes:
 		rawBytesToBinFormatter(f, rows, nil)
 	default:
--- a/sstable/colblk/block.go
+++ b/sstable/colblk/block.go
@@ -418,23 +418,14 @@ func (r *BlockReader) headerToBinFormatter(f *binfmt.Formatter) {
 	}
 }
 
-func (r *BlockReader) columnToBinFormatter(f *binfmt.Formatter, col, rows int) {
+func (r *BlockReader) formatColumn(
+	f *binfmt.Formatter, col int, fn func(*binfmt.Formatter, DataType),
+) {
 	f.CommentLine("data for column %d", col)
 	dataType := r.DataType(col)
 	colSize := r.pageStart(col+1) - r.pageStart(col)
 	endOff := f.Offset() + int(colSize)
-	switch dataType {
-	case DataTypeBool:
-		bitmapToBinFormatter(f, rows)
-	case DataTypeUint8, DataTypeUint16, DataTypeUint32, DataTypeUint64:
-		uintsToBinFormatter(f, rows, dataType, nil)
-	case DataTypePrefixBytes:
-		prefixBytesToBinFormatter(f, rows, nil)
-	case DataTypeBytes:
-		rawBytesToBinFormatter(f, rows, nil)
-	default:
-		panic("unimplemented")
-	}
+	fn(f, dataType)
 
 	// We expect formatting the column data to have consumed all the bytes
 	// between the column's pageOffset and the next column's pageOffset.
@@ -445,4 +436,22 @@ func (r *BlockReader) columnToBinFormatter(f *binfmt.Formatter, col, rows int) {
 	case -1:
 		panic(fmt.Sprintf("expected f.Offset() = %d, but found %d; did column %s format too many bytes?", endOff, f.Offset(), dataType))
 	}
+}
+
+func (r *BlockReader) columnToBinFormatter(f *binfmt.Formatter, col, rows int) {
+	r.formatColumn(f, col, func(f *binfmt.Formatter, dataType DataType) {
+		switch dataType {
+		case DataTypeBool:
+			bitmapToBinFormatter(f, rows)
+		case DataTypeUint8, DataTypeUint16, DataTypeUint32, DataTypeUint64:
+			uintsToBinFormatter(f, rows, dataType, nil)
+		case DataTypePrefixBytes:
+			prefixBytesToBinFormatter(f, rows, nil)
+		case DataTypeBytes:
+			rawBytesToBinFormatter(f, rows, nil)
+		default:
+			panic("unimplemented")
+		}
+	})
+
 }
--- a/sstable/colblk/cockroach_test.go
+++ b/sstable/colblk/cockroach_test.go
@@ -1,0 +1,467 @@
+// Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package colblk
+
+import (
+	"bytes"
+	"cmp"
+	"encoding/binary"
+	"fmt"
+	"io"
+	"sync"
+	"testing"
+	"time"
+	"unsafe"
+
+	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/crdbtest"
+	"github.com/cockroachdb/pebble/sstable/block"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/exp/rand"
+)
+
+const (
+	cockroachColPrefix int = iota
+	cockroachColMVCCWallTime
+	cockroachColMVCCLogical
+	cockroachColUntypedSuffix
+	cockroachColCount
+)
+
+var cockroachKeySchema = KeySchema{
+	ColumnTypes: []DataType{
+		cockroachColPrefix:        DataTypePrefixBytes,
+		cockroachColMVCCWallTime:  DataTypeUint64,
+		cockroachColMVCCLogical:   DataTypeUint32,
+		cockroachColUntypedSuffix: DataTypeBytes,
+	},
+	NewKeyWriter: func() KeyWriter {
+		kw := &cockroachKeyWriter{}
+		kw.prefixes.Init(16)
+		kw.wallTimes.Init()
+		kw.logicalTimes.InitWithDefault()
+		kw.untypedSuffixes.Init()
+		return kw
+	},
+	NewKeySeeker: func() KeySeeker {
+		return &cockroachKeySeeker{}
+	},
+}
+
+type cockroachKeyWriter struct {
+	prefixes        PrefixBytesBuilder
+	wallTimes       UintBuilder[uint64]
+	logicalTimes    UintBuilder[uint32]
+	untypedSuffixes RawBytesBuilder
+	prevSuffix      []byte
+}
+
+func (kw *cockroachKeyWriter) ComparePrev(key []byte) KeyComparison {
+	lp := kw.prefixes.LastKey()
+	var cmpv KeyComparison
+	cmpv.PrefixLen = int32(crdbtest.Split(key)) // TODO(jackson): Inline
+	cmpv.CommonPrefixLen = int32(bytesSharedPrefix(lp, key[:cmpv.PrefixLen]))
+	if cmpv.CommonPrefixLen == cmpv.PrefixLen {
+		cmpv.UserKeyComparison = int32(crdbtest.CompareSuffixes(key[cmpv.PrefixLen:], kw.prevSuffix))
+		return cmpv
+	}
+	// The keys have different MVCC prefixes. We haven't determined which is
+	// greater, but we know the index at which they diverge. The base.Comparer
+	// contract dictates that prefixes must be lexicograrphically ordered.
+	if len(lp) == int(cmpv.CommonPrefixLen) {
+		// cmpv.PrefixLen > cmpv.PrefixLenShared; key is greater.
+		cmpv.UserKeyComparison = +1
+	} else {
+		// Both keys have at least 1 additional byte at which they diverge.
+		// Compare the diverging byte.
+		cmpv.UserKeyComparison = int32(cmp.Compare(key[cmpv.CommonPrefixLen], lp[cmpv.CommonPrefixLen]))
+	}
+	return cmpv
+}
+
+func (kw *cockroachKeyWriter) WriteKey(
+	row int, key []byte, keyPrefixLen, keyPrefixLenSharedWithPrev int32,
+) {
+	// TODO(jackson): Avoid copying the previous suffix.
+	// TODO(jackson): Use keyPrefixLen to speed up decoding.
+	prefix, untypedSuffix, wallTime, logicalTime := crdbtest.DecodeTimestamp(key)
+	kw.prevSuffix = append(kw.prevSuffix[:0], key[keyPrefixLen:]...)
+	kw.prefixes.Put(prefix, int(keyPrefixLenSharedWithPrev))
+	kw.wallTimes.Set(row, wallTime)
+	// The w.logicalTimes builder was initialized with InitWithDefault, so if we
+	// don't set a value, the column value is implicitly zero. We only need to
+	// Set anything for non-zero values.
+	if logicalTime > 0 {
+		kw.logicalTimes.Set(row, logicalTime)
+	}
+	kw.untypedSuffixes.Put(untypedSuffix)
+}
+
+func (kw *cockroachKeyWriter) Reset() {
+	kw.prefixes.Reset()
+	kw.wallTimes.Reset()
+	kw.logicalTimes.Reset()
+	kw.untypedSuffixes.Reset()
+}
+
+func (kw *cockroachKeyWriter) WriteDebug(dst io.Writer, rows int) {
+	fmt.Fprint(dst, "prefixes: ")
+	kw.prefixes.WriteDebug(dst, rows)
+	fmt.Fprintln(dst)
+	fmt.Fprint(dst, "wall times: ")
+	kw.wallTimes.WriteDebug(dst, rows)
+	fmt.Fprintln(dst)
+	fmt.Fprint(dst, "logical times: ")
+	kw.logicalTimes.WriteDebug(dst, rows)
+	fmt.Fprintln(dst)
+	fmt.Fprint(dst, "untyped suffixes: ")
+	kw.untypedSuffixes.WriteDebug(dst, rows)
+	fmt.Fprintln(dst)
+}
+
+func (kw *cockroachKeyWriter) NumColumns() int {
+	return cockroachColCount
+}
+
+func (kw *cockroachKeyWriter) DataType(col int) DataType {
+	return cockroachKeySchema.ColumnTypes[col]
+}
+
+func (kw *cockroachKeyWriter) Size(rows int, offset uint32) uint32 {
+	offset = kw.prefixes.Size(rows, offset)
+	offset = kw.wallTimes.Size(rows, offset)
+	offset = kw.logicalTimes.Size(rows, offset)
+	offset = kw.untypedSuffixes.Size(rows, offset)
+	return offset
+}
+
+func (kw *cockroachKeyWriter) Finish(
+	col int, rows int, offset uint32, buf []byte,
+) (endOffset uint32) {
+	switch col {
+	case cockroachColPrefix:
+		return kw.prefixes.Finish(0, rows, offset, buf)
+	case cockroachColMVCCWallTime:
+		return kw.wallTimes.Finish(0, rows, offset, buf)
+	case cockroachColMVCCLogical:
+		return kw.logicalTimes.Finish(0, rows, offset, buf)
+	case cockroachColUntypedSuffix:
+		return kw.untypedSuffixes.Finish(0, rows, offset, buf)
+	default:
+		panic(fmt.Sprintf("unknown default key column: %d", col))
+	}
+}
+
+var cockroachKeySeekerPool = sync.Pool{
+	New: func() interface{} { return &cockroachKeySeeker{} },
+}
+
+type cockroachKeySeeker struct {
+	reader          *DataBlockReader
+	prefixes        PrefixBytes
+	mvccWallTimes   UnsafeUint64s
+	mvccLogical     UnsafeUint32s
+	untypedSuffixes RawBytes
+	sharedPrefix    []byte
+}
+
+func (ks *cockroachKeySeeker) Init(r *DataBlockReader) error {
+	ks.reader = r
+	ks.prefixes = r.r.PrefixBytes(cockroachColPrefix)
+	ks.mvccWallTimes = r.r.Uint64s(cockroachColMVCCWallTime)
+	ks.mvccLogical = r.r.Uint32s(cockroachColMVCCLogical)
+	ks.untypedSuffixes = r.r.RawBytes(cockroachColUntypedSuffix)
+	ks.sharedPrefix = ks.prefixes.SharedPrefix()
+	return nil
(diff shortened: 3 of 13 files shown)


RaduBerinde (commented):
Great stuff! Flushing some comments after an incomplete pass over `r1`

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3819#-:-O3IfrgVAo3ahqSoyIEc:bsexj94)* status: 0 of 8 files reviewed, 14 unresolved discussions (waiting on @jbowens)

___
*[`sstable/colblk/data_block.go` line 63 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IV4yN3dJ2vJc3ARb8:-O3IV4yN3dJ2vJc3ARb9:b-tzc0kn) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L63)):*
> ```Go
> type KeyComparison struct {
> 	// PrefixLen is the length of the prefix of the key.
> 	PrefixLen int
> ```

[nit] should we call this `Split` so it's consistent with `Comparer.Split`? Or at least mention the latter in a comment
___
*[`sstable/colblk/data_block.go` line 65 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IVwjH16A8ykootihg:-O3IVwjH16A8ykootihh:bu1n55r) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L65)):*
> ```Go
> 	PrefixLen int
> 	// PrefixLenShared is the length of the physical prefix of the logical
> 	// prefix that is shared with the previous key. For example, for "apple@1"
> ```

[nit] The struct is described as comparison to an arbitrary key, I'd use "the other key" instead of "previous key"
___
*[`sstable/colblk/data_block.go` line 67 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IVMgF7muXPQnPMTPA:-O3IVMgF7muXPQnPMTPB:b-1fimxh) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L67)):*
> ```Go
> 	// prefix that is shared with the previous key. For example, for "apple@1"
> 	// and "applied@3" the value is 4 (the length of "appl"). For "apple@1" and
> 	// "apple@10" the value is 5 (the length of "apple"), because shareed
> ```

[nit] shared

I like the logical vs physical way to distinguish. Maybe add `(byte-wise)` after the first `physical` and use logical key prefix
___
*[`sstable/colblk/data_block.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IVfWC7cQKdO1aimtD:-O3IVfWC7cQKdO1aimtE:b-ho9dm1) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L68)):*
> ```Go
> 	// and "applied@3" the value is 4 (the length of "appl"). For "apple@1" and
> 	// "apple@10" the value is 5 (the length of "apple"), because shareed
> 	// physical prefix within the suffix is not included.
> ```

Maybe "shared bytes within the suffix are not included"
___
*[`sstable/colblk/data_block.go` line 69 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IVDTL-FaCna9obYTL:-O3IVDTL-FaCna9obYTM:b-z84mzy) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L69)):*
> ```Go
> 	// "apple@10" the value is 5 (the length of "apple"), because shareed
> 	// physical prefix within the suffix is not included.
> 	PrefixLenShared int
> ```

[nit] CommonPrefixLen
___
*[`sstable/colblk/data_block.go` line 74 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IbRIqCDqvqjcO3E7U:-O3IbRIqCDqvqjcO3E7V:b-g3gosr) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L74)):*
> ```Go
> 	//
> 	//   Compare(key, otherKey)
> 	UserKeyComparison int
> ```

[nit] could make the fields int32s to make it smaller, since we're going to be passing it around in hot paths
___
*[`sstable/colblk/data_block.go` line 83 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IbbMN4Uxx2ndWazmq:-O3IbbMN4Uxx2ndWazmr:b59bp9z) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L83)):*
> ```Go
> // KeySeeker iterates over the keys in a columnar data block.
> //
> // Users of Pebble who define their own key schema and must implement
> ```

[nit] remove "and"
___
*[`sstable/colblk/data_block.go` line 105 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IeyVS9Plo7JsbILp0:-O3IeyVS9Plo7JsbILp1:bknoig7) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L105)):*
> ```Go
> 	// MaterializeUserKey materializes the user key of the specified row.
> 	// Implementations may use the memory of dst to avoid an allocation.
> 	MaterializeUserKey(dst []byte, row int) []byte
> ```

[nit] maybe remame `dst` to `buf` since we're *not* appending to it if it's non-zero length
___
*[`sstable/colblk/data_block.go` line 107 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IbhVo3mKU4_swXe6A:-O3IbhVo3mKU4_swXe6B:bnv7dv6) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L107)):*
> ```Go
> 	MaterializeUserKey(dst []byte, row int) []byte
> 	// Release releases the KeySeeker. It's called when the seeker is no longer
> 	// in use. Implementations may pool KeySeeker implementations.
> ```

[nit] KeySeeker objects
___
*[`sstable/colblk/data_block.go` line 128 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IbvUf1uzTQWKp0nNw:-O3IbvUf1uzTQWKp0nNx:b-lbkln0) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L128)):*
> ```Go
> 
> // DefaultKeySchema returns the default key schema that decomposes a user key
> // into its lexicographically sorted prefix, and a suffix.
> ```

[nit] "lexicographically sorted" is odd when we're talking about a single user key. Maybe just add a separate "Prefixes are stored in lexicographical order".
___
*[`sstable/colblk/data_block.go` line 163 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IcBcZ4snsxhCiXwX-:-O3IcBc_2SsOhqSfbwgR:bi5d990) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L163)):*
> ```Go
> 		// The keys share the same MVCC prefix. Compare the suffixes.
> 		cmpv.UserKeyComparison = w.comparer.CompareSuffixes(key[cmpv.PrefixLen:], w.suffixes.LastSlice())
> 		return cmpv
> ```

Here we are assuming that a key (logical) prefix can't be a byte-wise-prefix of another key prefix. I don't think this is always the case (definitely not with the default or testkeys comparer). We need to look at the previous key's split point - maybe `LastKey()` can return that too?

[nit] add an invariant cross-check against `comparer.Compare` (in both return paths)
___
*[`sstable/colblk/data_block.go` line 174 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3Id8o9221NK9Mv9LTs:-O3Id8oA-uRYhH56D2cn:bfxserp) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L174)):*
> ```Go
> 		// Both keys have at least 1 additional byte at which they diverge.
> 		// Compare the diverging byte.
> 		cmpv.UserKeyComparison = cmp.Compare(key[cmpv.PrefixLenShared], lp[cmpv.PrefixLenShared])
> ```

[nit] invariant check that the result is non-zero
___
*[`sstable/colblk/data_block.go` line 274 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3Idv_W82kD43SODoj-:-O3Idv_XDleYqqfohZeL:b-w1u50w) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L274)):*
> ```Go
> 		h := int(uint(l+u) >> 1) // avoid overflow when computing h
> 		// l ≤ h < u
> 		c := ki.comparer.CompareSuffixes(suffix, ki.suffixes.At(h))
> ```

[nit] I'd swap the argument so the "inclusive" case matches the GE condition we are implementing, e.g.
```
if CompareSuffixes(At(h), suffix) >= 0 {
  u = h
} else {
  l = h+1
}
```
___
*[`sstable/colblk/data_block.go` line 295 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IeZds2rG6CSIjjyEr:-O3IeZds2rG6CSIjjyEs:boqqyx6) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L295)):*
> ```Go
> 	// Ensure i.key has sufficient capacity.
> 	n := len(ki.sharedPrefix) + len(bundlePrefix) + len(rowSuffix) + len(suffix)
> 	if len(dst) < n {
> ```

We should use `cap(dst)` here instead of the caller doing it.

We should actually do `dst = slices.Grow(dst[:0], n)` then `append` to ensure we don't reallocate many times if we have a slowly increasing `n`.

Or if we want to actually have append semantics (which is more typical), we can do `slices.Grow(dst, len(dst) + n)`.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3819)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`sstable/colblk/data_block.go` line 63 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IV4yN3dJ2vJc3ARb8:-O3JKoEC-Xm8i1AOJ0mQ:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L63)):*

[nit] should we call this `Split` so it's consistent with `Comparer.Split`? Or at least mention the latter in a comment
</blockquote></details>

Done
___
*[`sstable/colblk/data_block.go` line 65 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IVwjH16A8ykootihg:-O3JKosZ3PHcEThCcdyV:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L65)):*

[nit] The struct is described as comparison to an arbitrary key, I'd use "the other key" instead of "previous key"
</blockquote></details>

Done
___
*[`sstable/colblk/data_block.go` line 67 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IVMgF7muXPQnPMTPA:-O3JLHlL4bqfO2p8JpJH:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L67)):*

[nit] shared

I like the logical vs physical way to distinguish. Maybe add `(byte-wise)` after the first `physical` and use logical key prefix
</blockquote></details>

Done
___
*[`sstable/colblk/data_block.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IVfWC7cQKdO1aimtD:-O3JLIDoBDf483wPqD6D:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L68)):*

Maybe "shared bytes within the suffix are not included"
</blockquote></details>

Done
___
*[`sstable/colblk/data_block.go` line 69 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IVDTL-FaCna9obYTL:-O3JLLnqBUcv9HowtY8d:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L69)):*

[nit] CommonPrefixLen
</blockquote></details>

Done
___
*[`sstable/colblk/data_block.go` line 74 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IbRIqCDqvqjcO3E7U:-O3JM9PB4JrhVzab-MXX:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L74)):*

[nit] could make the fields int32s to make it smaller, since we're going to be passing it around in hot paths
</blockquote></details>

Done
___
*[`sstable/colblk/data_block.go` line 83 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IbbMN4Uxx2ndWazmq:-O3JLSn606YxetwER-cQ:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L83)):*

[nit] remove "and"
</blockquote></details>

Done
___
*[`sstable/colblk/data_block.go` line 105 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IeyVS9Plo7JsbILp0:-O3JMV2H8qQ8q1NIH-ra:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L105)):*

[nit] maybe remame `dst` to `buf` since we're *not* appending to it if it's non-zero length
</blockquote></details>

Done
___
*[`sstable/colblk/data_block.go` line 107 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IbhVo3mKU4_swXe6A:-O3JMZW1CQS9YFKMyhuq:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L107)):*

[nit] KeySeeker objects
</blockquote></details>

Done
___
*[`sstable/colblk/data_block.go` line 128 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IbvUf1uzTQWKp0nNw:-O3JMhmP182dbs825913:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L128)):*

[nit] "lexicographically sorted" is odd when we're talking about a single user key. Maybe just add a separate "Prefixes are stored in lexicographical order".
</blockquote></details>

Done.
___
*[`sstable/colblk/data_block.go` line 163 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IcBcZ4snsxhCiXwX-:-O3JN4PFA4ANKU4FaTnL:bv4j6dk) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L163)):*

Here we are assuming that a key (logical) prefix can't be a byte-wise-prefix of another key prefix. I don't think this is always the case (definitely not with the default or testkeys comparer). We need to look at the previous key's split point - maybe `LastKey()` can return that too?

[nit] add an invariant cross-check against `comparer.Compare` (in both return paths)
</blockquote></details>

We're taking advantage of the lexicographical ordering. Since prev ≤ key, CommonPrefixLen == PrefixLen implies equality of prefix. Added a comment and assertion that clarifies this.

Cross-checking against `Compare` would requires an allocation (to materialize the previous key), which I think is too much even behind `invariants` in such a hot path
___
*[`sstable/colblk/data_block.go` line 174 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3Id8o9221NK9Mv9LTs:-O3JUU030jgJmqTPqEOU:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L174)):*

[nit] invariant check that the result is non-zero
</blockquote></details>

Done
___
*[`sstable/colblk/data_block.go` line 274 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3Idv_W82kD43SODoj-:-O3JYDIG4wkt32D0_k5W:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L274)):*

[nit] I'd swap the argument so the "inclusive" case matches the GE condition we are implementing, e.g.
```
if CompareSuffixes(At(h), suffix) >= 0 {
  u = h
} else {
  l = h+1
}
```
</blockquote></details>

Done

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3819)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`sstable/colblk/data_block.go` line 163 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3IcBcZ4snsxhCiXwX-:-O3NIfrZ6QXX_qRRmIgi:b7x5pp8) ([raw file](https://github.com/cockroachdb/pebble/blob/39bb1588d7642836f0630362a1b6bfaaddfeb47a/sstable/colblk/data_block.go#L163)):*

We're taking advantage of the lexicographical ordering. Since prev ≤ key, CommonPrefixLen == PrefixLen implies equality of prefix. Added a comment and assertion that clarifies this.

Cross-checking against `Compare` would requires an allocation (to materialize the previous key), which I think is too much even behind `invariants` in such a hot path
</blockquote></details>

Nice, makes sense, I missed that PrefixLen would be longer for this key.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3819)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`sstable/colblk/prefix_bytes.go` line 342 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3nouzO1aZ3g9c2g-U5:-O3nouzO1aZ3g9c2g-U6:bamyfnx) ([raw file](https://github.com/cockroachdb/pebble/blob/791b3749feb52978b973d79904fb79632b38046e/sstable/colblk/prefix_bytes.go#L342)):*
> ```Go
> 		it.cap = it.len << 1
> 		it.ptr = mallocgc(uintptr(it.cap), nil, false)
> 		memmove(it.ptr, b.rawBytes.data, uintptr(b.sharedPrefixLen))
> ```

I'm at a loss here. This commit seems to show a consistent ~3% _regression_ in the PrefixBytes iteration benchmark when we remove the logic to grow the buffer if necessary:

```
goos: darwin
goarch: arm64
pkg: github.com/cockroachdb/pebble/sstable/colblk
PrefixBytes/alphaLen=2/iteration-10    5.611n ± 1%   5.811n ± 1%  +3.56% (p=0.000 n=25)
PrefixBytes/alphaLen=5/iteration-10    5.701n ± 2%   5.877n ± 1%  +3.09% (p=0.000 n=25)
PrefixBytes/alphaLen=26/iteration-10   5.554n ± 1%   5.810n ± 1%  +4.61% (p=0.000 n=25)
geomean                                5.622n        5.833n       +3.75%
```

RaduBerinde (approved):
:lgtm: overall. I will try to do one more pass in more depth but feel free to merge before that.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3819#-:-O4C5vii4noTTmUmAny5:b-znzn1)* status: 0 of 13 files reviewed, 5 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`-- commits` line 17 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O4Bwne4Bnip4lRoZg_v):*
What a great idea!
___
*[`-- commits` line 18 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O4BwlYx0ycncTMRqLNH):*
[nit] contained
___
*[`sstable/colblk/data_block.go` line 59 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O4Bt6q80zr-1xh1niVg:-O4Bt6q80zr-1xh1niVh:biu5dqc) ([raw file](https://github.com/cockroachdb/pebble/blob/838b01bd2337eaec16bdcd7f8dffdc2475445361/sstable/colblk/data_block.go#L59)):*
> ```Go
> 	// be stable until the next call to WriteKey or the KeyWriter columns are
> 	// finished.
> 	WriteKey(row int, key []byte, keyPrefixLen, keyPrefixLenSharedWithPrev int32) (unsafePrefix []byte)
> ```

[nit] The return value is strange here, the caller already has this information. Maybe explain a bit more the usecase you have in mind.
___
*[`sstable/colblk/prefix_bytes.go` line 342 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3nouzO1aZ3g9c2g-U5:-O4C-Aij9rKuwBMSA2uJ:bn4n8ap) ([raw file](https://github.com/cockroachdb/pebble/blob/791b3749feb52978b973d79904fb79632b38046e/sstable/colblk/prefix_bytes.go#L342)):*

I'm at a loss here. This commit seems to show a consistent ~3% _regression_ in the PrefixBytes iteration benchmark when we remove the logic to grow the buffer if necessary:

```
goos: darwin
goarch: arm64
pkg: github.com/cockroachdb/pebble/sstable/colblk
PrefixBytes/alphaLen=2/iteration-10    5.611n ± 1%   5.811n ± 1%  +3.56% (p=0.000 n=25)
PrefixBytes/alphaLen=5/iteration-10    5.701n ± 2%   5.877n ± 1%  +3.09% (p=0.000 n=25)
PrefixBytes/alphaLen=26/iteration-10   5.554n ± 1%   5.810n ± 1%  +4.61% (p=0.000 n=25)
geomean                                5.622n        5.833n       +3.75%
```
</blockquote></details>

I tried it too and see something similar. I can recover the difference if I remove the `invariants.Enabled` from the first length check in SetNext. I looked at the assembly and it's pretty much the same other than the extra check and panic handling code..

FWIW I saw paradoxical results with benchmarks that have nanosecond-order times on multiple occasions (e.g. a code difference that wasn't even being run). I suspect that the way important instruction blocks align to cachelines might play a role. In any case, I would not worry about it.

One optimization here would be to maintain the value of `At(offsetIndex+1)` in the struct. We would only need to do a single `At` in the `SetNext` fast path instead of two.

I think it would be a good idea to do the length checks in all builds, if there is a bug somewhere (perhaps on the building side), the result could be impossible to debug.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3819)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`-- commits` line 17 at r12](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O4Bwne4Bnip4lRoZg_v):*

What a great idea!
</blockquote></details>

:raised_hands: 
___
*[`sstable/colblk/data_block.go` line 59 at r10](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O4Bt6q80zr-1xh1niVg:-O4LWldf8D0c39EjmVT3:b-68neri) ([raw file](https://github.com/cockroachdb/pebble/blob/838b01bd2337eaec16bdcd7f8dffdc2475445361/sstable/colblk/data_block.go#L59)):*

[nit] The return value is strange here, the caller already has this information. Maybe explain a bit more the usecase you have in mind.
</blockquote></details>

I think this is no longer necessary because I added `ComparePrev` to the `KeyWriter` interface. Previously, this return value was what allowed the sstable writer to compare against the previous key's prefix without having to copy it. I'll remove it and we can resurrect it if I'm overlooking something.
___
*[`sstable/colblk/prefix_bytes.go` line 342 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3nouzO1aZ3g9c2g-U5:-O4LVydS1QXjWhWPtI9U:b-z9b2ka) ([raw file](https://github.com/cockroachdb/pebble/blob/791b3749feb52978b973d79904fb79632b38046e/sstable/colblk/prefix_bytes.go#L342)):*
> One optimization here would be to maintain the value of `At(offsetIndex+1)` in the struct. We would only need to do a single `At` in the `SetNext` fast path instead of two.

I wonder if this won't be as important now with your adjustment to the offset encoding and specialized offset lookup? Eg, is it actually faster to add an additional store the value of `At(offsetIndex+1)` and then load it, rather than performing the `At(offsetIndex+1)` load?

> I think it would be a good idea to do the length checks in all builds, if there is a bug somewhere (perhaps on the building side), the result could be impossible to debug.

Yeah, I debated this. You're right, we should be conservative

RaduBerinde:
*[`sstable/colblk/prefix_bytes.go` line 342 at r9](https://reviewable.io/reviews/cockroachdb/pebble/3819#-O3nouzO1aZ3g9c2g-U5:-O4N0xU7D1Cw-mO86a7M:bovw88o) ([raw file](https://github.com/cockroachdb/pebble/blob/791b3749feb52978b973d79904fb79632b38046e/sstable/colblk/prefix_bytes.go#L342)):*

> One optimization here would be to maintain the value of `At(offsetIndex+1)` in the struct. We would only need to do a single `At` in the `SetNext` fast path instead of two.

I wonder if this won't be as important now with your adjustment to the offset encoding and specialized offset lookup? Eg, is it actually faster to add an additional store the value of `At(offsetIndex+1)` and then load it, rather than performing the `At(offsetIndex+1)` load?

> I think it would be a good idea to do the length checks in all builds, if there is a bug somewhere (perhaps on the building side), the result could be impossible to debug.

Yeah, I debated this. You're right, we should be conservative
</blockquote></details>

Yeah I think it won't matter with the new offset lookup

PR #3823, opened 2024-08-01
colblk: fix UintBuilder's construction of constant column with default

jbowens:
Previously if a UintBuilder was constructed with InitWithDefault, and the column contained no non-zero elements, the UintBuilder would never construct an array to back b.array.elems. The call to unsafe.Slice would panic when it observed a non-zero element count and a nil pointer.


RaduBerinde (approved):

jbowens:
TFTR!

PR #3832, opened 2024-08-08
sstable: use block cache / readhandle for downloads

### LINKED ISSUE #3758: db: Download should use block cache for metadata/index block lookups
Currently `Download` reads from the file separately and doesn't use any information that might already be in the block cache. We should use any cached Reader which should already have the layout info, and we should use the block cache for index blocks when seeking.

itsbilal:
Previously, if we had some blocks in the block cache at the time of starting a download, we would still read all blocks from object storage. This was pretty wasteful, especially for index/footer/etc blocks that are more likely to be in cache.

This change threads through the appropriate Reader through sstable.CopySpan, so that the cached blocks can be appropriately used. Data blocks are also read through the cache if they exist in the cache; if not, we fall back to the old sequence of reading and writing bytes blindly from object storage to local disk as fast as IO allows us to.

Fixes #3758.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2287,11 +2287,18 @@ func (d *DB) runCopyCompaction(
 			}
 		}
 
-		wrote, err := sstable.CopySpan(ctx,
-			src, d.opts.MakeReaderOptions(),
-			w, d.opts.MakeWriterOptions(c.outputLevel.level, d.FormatMajorVersion().MaxTableFormat()),
-			start, end,
-		)
+		// NB: external files are always virtual.
+		var wrote uint64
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):
:lgtm:

Very cool! It would be nice to have a test that ingests an external SST, performs a random set of reads, then copies it and then verifies the contents.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3832#-:-O4BorVmEIGrcMQJjhv8:b-q9lk8c)* status: 0 of 4 files reviewed, all discussions resolved

itsbilal:
@RaduBerinde we should be getting similar test coverage through the metamorphic test already, right?

RaduBerinde:
Ah, yes, we should be fine.

PR #3834, opened 2024-08-10
sstable: genericize sstable iterators by data block iterator

jbowens:
In preparation for columnar blocks, this commit introduces type parameters to the {single,two}LevelIterator types. Parameterizing the sstable iterator ensures we pay the cost of dynamic dispatch only when entering the sstable iterator (as we do today).

In order to continue to allocate the data block iterator struct within the single-level iterator, this commit introduces two type paramters: one for the non-pointer iterator type, and one for the pointer type on which iteration methods are defined. This adds some slightly cumbersome ergonomics to data block iterator usage sites.

### DIFF
--- a/sstable/block/block.go
+++ b/sstable/block/block.go
@@ -89,6 +89,45 @@ func (c *Checksummer) Checksum(block []byte, blockType []byte) (checksum uint32)
 	return checksum
 }
 
+// DataBlockIterator is a type constraint for implementations of block iterators
+// over data blocks. It's currently satisifed by the *rowblk.Iter type.
+//
+// DataBlockIterator requires that the type be a pointer to its type parameter,
+// D, to allow sstable iterators embed the block iterator within its struct. See
+// this example from the Go generics proposal:
+// https://go.googlesource.com/proposal/+/refs/heads/master/design/43651-type-parameters.md#pointer-method-example
+type DataBlockIterator[D any] interface {
+	base.InternalIterator
+
+	// Handle returns the handle to the block.
+	Handle() BufferHandle
+	// InitHandle initializes the block from the provided buffer handle.
+	InitHandle(base.Compare, base.Split, BufferHandle, IterTransforms) error
+	// Valid returns true if the iterator is currently positioned at a valid KV.
+	Valid() bool
+	// KV returns the key-value pair at the current iterator position. The
+	// iterator must be Valid().
+	KV() *base.InternalKV
+	// ResetForReuse resets the iterator so that it may be used for iteration
+	// over a new block. It returns the non-pointer D type to allow resetting
+	// while initializing the containing struct, eg::
+	//   iter = sstableIter{dataBlockIter: iter.dataBlockIter.ResetForReuse()}
+	ResetForReuse() D
+	// FirstUserKey returns the first user key contained within the data block.
+	FirstUserKey() []byte
+	// Invalidate invalidates the block iterator, removing references to the block
+	// it was initialized with.
+	Invalidate()
+	// IsDataInvalidated returns true when the iterator has been invalidated
+	// using an Invalidate call.
+	//
+	// NB: this is different from Valid which indicates whether the current *KV*
+	// is valid.
+	IsDataInvalidated() bool
+
+	*D // non-interface type constraint element
+}
+
 // IterTransforms allow on-the-fly transformation of data at iteration time.
 //
 // These transformations could in principle be implemented as block transforms
--- a/sstable/data_test.go
+++ b/sstable/data_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/cockroachdb/pebble/internal/testkeys"
 	"github.com/cockroachdb/pebble/objstorage"
 	"github.com/cockroachdb/pebble/objstorage/objstorageprovider"
+	"github.com/cockroachdb/pebble/sstable/rowblk"
 	"github.com/cockroachdb/pebble/vfs"
 )
 
@@ -427,8 +428,8 @@ func runIterCmd(
 			continue
 		case "internal-iter-state":
 			fmt.Fprintf(&b, "| %T:\n", origIter)
-			si, _ := origIter.(*singleLevelIterator)
-			if twoLevelIter, ok := origIter.(*twoLevelIterator); ok {
+			si, _ := origIter.(*singleLevelIterator[rowblk.Iter, *rowblk.Iter])
+			if twoLevelIter, ok := origIter.(*twoLevelIterator[rowblk.Iter, *rowblk.Iter]); ok {
 				si = &twoLevelIter.secondLevel
 				if twoLevelIter.topLevelIndex.Valid() {
 					fmt.Fprintf(&b, "|  topLevelIndex.Key() = %q\n", twoLevelIter.topLevelIndex.Key())
--- a/sstable/reader.go
+++ b/sstable/reader.go
@@ -207,10 +207,11 @@ func (r *Reader) newPointIter(
 	var res Iterator
 	var err error
 	if r.Properties.IndexType == twoLevelIndex {
-		res, err = newTwoLevelIterator(ctx, r, vState, transforms, lower, upper, filterer, filterBlockSizeLimit,
+		res, err = newRowBlockTwoLevelIterator(
+			ctx, r, vState, transforms, lower, upper, filterer, filterBlockSizeLimit,
 			stats, categoryAndQoS, statsCollector, rp, nil /* bufferPool */)
 	} else {
-		res, err = newSingleLevelIterator(
+		res, err = newRowBlockSingleLevelIterator(
 			ctx, r, vState, transforms, lower, upper, filterer, filterBlockSizeLimit,
 			stats, categoryAndQoS, statsCollector, rp, nil /* bufferPool */)
 	}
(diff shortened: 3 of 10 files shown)


sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3834)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`sstable/reader_iter_single_lvl.go` line 249 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3834#-O46-ehX0LlZDbPyd5oj:-O46-ehX0LlZDbPyd5ok:bjk6maw) ([raw file](https://github.com/cockroachdb/pebble/blob/9759e901e6ecc9c2916739309e6706b5649ab5f6/sstable/reader_iter_single_lvl.go#L249)):*
> ```Go
> 				stats:  stats,
> 			}
> 			(&i.data).SetGetLazyValue(i.vbReader.getLazyValueForPrefixAndValueHandle)
> ```

I must have missed this `SetGetLazyValue` change. This is probably an allocation, for the closure. Can you check? If an interface was used we would avoid that allocation.

Also, I am confused why this compiles -- how is *D implementing `SetGetLazyValue`? 
___
*[`sstable/reader_iter_single_lvl.go` line 472 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3834#-O460f3iBpk40753iZUt:-O460f3iBpk40753iZUu:b-ynhspx) ([raw file](https://github.com/cockroachdb/pebble/blob/9759e901e6ecc9c2916739309e6706b5649ab5f6/sstable/reader_iter_single_lvl.go#L472)):*
> ```Go
> 		// Ensure the data block iterator is invalidated even if loading of the block
> 		// fails.
> 		PD(&i.data).Invalidate()
> ```

How about also adding a `PD` member to `singleLevelIterator` when initializing, and avoiding all these casts, given many of these are on the critical performance path.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3834)* status: all files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 249 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3834#-O46-ehX0LlZDbPyd5oj:-O463-dF6utEhVoogFNa:b1r1z0) ([raw file](https://github.com/cockroachdb/pebble/blob/9759e901e6ecc9c2916739309e6706b5649ab5f6/sstable/reader_iter_single_lvl.go#L249)):*
> I must have missed this `SetGetLazyValue` change. This is probably an allocation, for the closure. Can you check? If an interface was used we would avoid that allocation.

Good catch: https://github.com/cockroachdb/pebble/pull/3838

> Also, I am confused why this compiles -- how is \*D implementing `SetGetLazyValue`?

I think because `D` and `PD` have been substituted with concrete types up above, it knows `&i.data` is `*rowblik.Iter` which implement `SetGetLazyValue`.
___
*[`sstable/reader_iter_single_lvl.go` line 472 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3834#-O460f3iBpk40753iZUt:-O462B4J9N4VD93ge-iN:b4thhs6) ([raw file](https://github.com/cockroachdb/pebble/blob/9759e901e6ecc9c2916739309e6706b5649ab5f6/sstable/reader_iter_single_lvl.go#L472)):*

How about also adding a `PD` member to `singleLevelIterator` when initializing, and avoiding all these casts, given many of these are on the critical performance path.
</blockquote></details>

Shouldn't the casts be free since they're just changing the type of the pointer address? My mental model might be off, but I thought a `PD` member would force an extra load of the member, whereas `PD(&i.data)` would just be the pointer of `i` added with a fixed offset into the struct.

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3834#-:-O46kpzPC0DNttRmrDOI:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`sstable/reader_iter_single_lvl.go` line 472 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3834#-O460f3iBpk40753iZUt:-O46YwGp237cJ5Kn2ii0:b-fdsf8w) ([raw file](https://github.com/cockroachdb/pebble/blob/9759e901e6ecc9c2916739309e6706b5649ab5f6/sstable/reader_iter_single_lvl.go#L472)):*
My mental model here is average at best, but here is what I believe:

Type assertions to a concrete type are very cheap since it involves a simple pointer comparison between a known concrete type and the real type available in the interface struct.

But this is not that kind of type assertion. It seems akin to a type assertion to an interface (based on the description in https://research.swtch.com/interfaces). The runtime code knows the concrete type of `&i.data`, and needs to (a) confirm that it implements PD, (b) construct the (or find the already constructed) itable for the (interface type, concrete type) pair, so it can set the itable pointer in the PD interface struct.

I suppose it can do (b) first and if is in the cache the answer to (a) must be yes. Perhaps the assembly for this could shed some light.

> I thought a `PD` member would force an extra load of the member

Hmm, perhaps I should be thinking of loads. I simplistically assume everything is in a register.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3834)* status: all files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 472 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3834#-O460f3iBpk40753iZUt:-O46lYgHBrYfsCJmNpuW:b-4ul6s8) ([raw file](https://github.com/cockroachdb/pebble/blob/9759e901e6ecc9c2916739309e6706b5649ab5f6/sstable/reader_iter_single_lvl.go#L472)):*
> Type assertions to a concrete type are very cheap since it involves a simple pointer comparison between a known concrete type and the real type available in the interface struct.

So I don't really know what I'm talking about, but since this is using generics I assumed at code generation time these _are_ both statically-known concrete types. I'll futz around with generating the assembly of a simplified example to see if I can parse what it's doing.

jbowens:
I'm struggling to parse the assembly, but the light purple is the assembly for a function call using PD(&i.data)

Screenshot 2024-08-13 at 11 21 27 AM

and then here the light purple is the assembly for a function call using an explicit `PD` member:

Screenshot 2024-08-13 at 11 21 08 AM

looks like there is an extra MOVQ instruction

jbowens:
TFTR!

RaduBerinde:
*[`sstable/block/block.go` line 115 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3834#-O4Bs0b4Ab1vatoLB1HQ:-O4Bs0b58X3AHkJFxtNw:byqjsjp) ([raw file](https://github.com/cockroachdb/pebble/blob/3ff865b4153a4d33f79ea6ad8471b67567038e3c/sstable/block/block.go#L115)):*
> ```Go
> 	// while initializing the containing struct, eg::
> 	//   iter = sstableIter{dataBlockIter: iter.dataBlockIter.ResetForReuse()}
> 	ResetForReuse() D
> ```

Is this the only reason for having the `D` parameter? Can't this be a `Reset()` function that mutates the object?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3834)* status: 6 of 10 files reviewed, 3 unresolved discussions

___
*[`sstable/block/block.go` line 115 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3834#-O4Bs0b4Ab1vatoLB1HQ:-O4C6HGu8JTcuGpURQe_:bw7d69z) ([raw file](https://github.com/cockroachdb/pebble/blob/3ff865b4153a4d33f79ea6ad8471b67567038e3c/sstable/block/block.go#L115)):*
> Is this the only reason for having the `D` parameter?

No, the `*D` down below is the reason for it. It allows us to allocate the block iterator with the single-level iterator, even though the pointer type is the one that implements the iterator interface.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3834)* status: 6 of 10 files reviewed, 3 unresolved discussions

___
*[`sstable/block/block.go` line 115 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3834#-O4Bs0b4Ab1vatoLB1HQ:-O4CAXKm5ZZD-iU82_ql:b-6gq9to) ([raw file](https://github.com/cockroachdb/pebble/blob/3ff865b4153a4d33f79ea6ad8471b67567038e3c/sstable/block/block.go#L115)):*

> Is this the only reason for having the `D` parameter?

No, the `*D` down below is the reason for it. It allows us to allocate the block iterator with the single-level iterator, even though the pointer type is the one that implements the iterator interface.
</blockquote></details>

:exploding_head:

RaduBerinde:
*[`sstable/block/block.go` line 115 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3834#-O4Bs0b4Ab1vatoLB1HQ:-O4CHldIETZTAxitbZhE:b147hnx) ([raw file](https://github.com/cockroachdb/pebble/blob/3ff865b4153a4d33f79ea6ad8471b67567038e3c/sstable/block/block.go#L115)):*

:exploding_head:
</blockquote></details>

It would be nice if we could have a type constraint like `[D | *D DataBlockIterator]`

PR #3836, opened 2024-08-12
go.mod, go.sum: upgrade to remove old dependencies

jba:
Upgrade cobra, prometheus and golang.org/x/perf to the latest versions. This removes a lot of clutter from go.sum, including the spiral of old cloud.google.com/go versions. Even though these old versions aren't part of the build, the go command is forced to fetch their go.mod files.


go.mod
```diff
@@ -36,16 +36,15 @@ require (
...
 	github.com/gogo/protobuf v1.3.2 // indirect
-	github.com/golang/protobuf v1.5.3 // indirect
-	github.com/inconshreveable/mousetrap v1.0.0 // indirect
+	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/kr/text v0.2.0 // indirect
-	github.com/matttproud/golang_protobuf_extensions v1.0.2-0.20181231171920-c182affec369 // indirect
-	github.com/prometheus/common v0.32.1 // indirect
-	github.com/prometheus/procfs v0.7.3 // indirect
-	github.com/rogpeppe/go-internal v1.9.0 // indirect
+	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
```
nicktrav, line 41: Where is this dep coming from, and what is it used for?

jba: `go mod why` says:
```
github.com/cockroachdb/pebble
github.com/prometheus/client_golang/prometheus                                                                                               
github.com/prometheus/common/expfmt
github.com/munnerz/goautoneg  
```

The expfmt package uses it twice:
- https://github.com/prometheus/common/blob/main/expfmt/encode.go#L67
- https://github.com/prometheus/common/blob/main/expfmt/encode.go#L100

`ParseAccept` parses an accept header:
https://github.com/munnerz/goautoneg/blob/master/autoneg.go#L93

xxmplus:
Close to due to age

PR #3839, opened 2024-08-12
db: fix mkdirAllAndSyncParents edge case

jbowens:
On real filesystems, relative paths were not handled correctly, sometimes attempting to sync the empty path "".


jbowens:
Eh, these MemFS semantics are pretty annoying. Maybe I should just bite the bullet and try to fix them outright.

PR #2178, opened 2022-12-07
db: refactor merging iterator's prefix handling for TrySeekUsingNext

jbowens:
Builds off of #2151. This commit refactors the merging iterator's
handling of prefix-related invariants for enabling TrySeekUsingNext
optimizations when using SeekPrefixGE.

The TrySeekUsingNext optimization notes successively larger seek keys,
and attempts to use next to find the appropriate key in the later seeks,
avoiding seeking from scratch. In a level's file metadata, we now use
the presence of the TrySeekUsingNext flag to signal that we can Next
through the file metadata. This saves key comparisons, especially in
larger LSMs with many files within a level.

Ever since the introduction of #2048, we've encountered a slew of
metamorphic test failures due to interactions between this optimization
and prefix iteration. We've made various fixes in #2087, #2111, #2123
and #2151.

Most of these failures revolve around the progression of a level
iterator beyond the iterator's current iteration prefix. This is
problematic, because the incomplete view of the LSM during prefix
iteration can result in inconsistent positioning outside of the current
iterator prefix. The internal iterators do not guarantee that keys or
range deletions outside the prefix are observed.

This commit fixes the latest instance of this (#2174), and attempts to
restructure the code to clarify the required invariants. Now, any method
that may advance the iterator during prefix iteration (eg, nextEntry or
isNextEntryDeleted) check that the iteration prefix has not already been
exceeded. This check is only performed when the merging iterator is
advancing internal iterators of its own volition (eg, due to non-visible
keys or keys deleted by range tombstones). It does NOT perform this
check when the top-level Iterator Nexts the merging iterator because
of a user Next, a point-delete, etc. This means these prefix checks are
only redundant with the top-level Iterator Nexts in two cases:

  a) the heap root is deleted by a range tombstone
  b) the heap root is not visible at the iterator's sequence number

Both of these cases are expected to be rare, and the redundant prefix
check is expected to be insignificant.

Additionally, to obey these new stricter invariants, the interleaving
iterator is adjusted to exhaust the iterator when the remainder of a
prefix is masked by a range key. Besides ensuring invariants are
respected, this change provides a large benefit to prefix iteration
within a large swath of masked point keys.

```
name                                                                     old time/op  new time/op  delta
Iterator_RangeKeyMasking/range-keys-suffixes=@10/forward/seekprefix-24   59.8ms ± 2%  60.3ms ± 2%   +0.92%  (p=0.013 n=25+25)
[...]

PR #3845, opened 2024-08-12
db: fix cases where SeekPrefixGE prefix doesn't match key

RaduBerinde:
This commit adds `SeekPrefixGE` assertions verifying that the `prefix` actually is the prefix for the `key`. This was not always the case so the offending code paths are adjusted.

In the future, we should create a wrapper iterator that verifies this sort of thing.

Informs #3794

### DIFF
--- a/internal/base/comparer.go
+++ b/internal/base/comparer.go
@@ -159,6 +159,12 @@ func (s Split) Prefix(k []byte) []byte {
 	return k[:i:i]
 }
 
+// HasPrefix returns true if the given key has the given prefix.
+func (s Split) HasPrefix(prefix, key []byte) bool {
+	i := s(key)
+	return bytes.Equal(prefix, key[:i:i])
+}
+
 // DefaultSplit is a trivial implementation of Split which always returns the
 // full key.
 var DefaultSplit Split = func(key []byte) int { return len(key) }
--- a/iterator.go
+++ b/iterator.go
@@ -1476,23 +1476,18 @@ func (i *Iterator) SeekPrefixGE(key []byte) bool {
 	}
 	// Make a copy of the prefix so that modifications to the key after
 	// SeekPrefixGE returns does not affect the stored prefix.
-	if cap(i.prefixOrFullSeekKey) < prefixLen {
-		i.prefixOrFullSeekKey = make([]byte, prefixLen)
-	} else {
-		i.prefixOrFullSeekKey = i.prefixOrFullSeekKey[:prefixLen]
-	}
 	i.hasPrefix = true
-	copy(i.prefixOrFullSeekKey, keyPrefix)
+	i.prefixOrFullSeekKey = append(i.prefixOrFullSeekKey[:0], keyPrefix...)
 
 	if lowerBound := i.opts.GetLowerBound(); lowerBound != nil && i.cmp(key, lowerBound) < 0 {
-		if p := i.comparer.Split.Prefix(lowerBound); !bytes.Equal(i.prefixOrFullSeekKey, p) {
+		if !i.comparer.Split.HasPrefix(i.prefixOrFullSeekKey, lowerBound) {
 			i.err = errors.New("pebble: SeekPrefixGE supplied with key outside of lower bound")
 			i.iterValidityState = IterExhausted
 			return false
 		}
 		key = lowerBound
 	} else if upperBound := i.opts.GetUpperBound(); upperBound != nil && i.cmp(key, upperBound) > 0 {
-		if p := i.comparer.Split.Prefix(upperBound); !bytes.Equal(i.prefixOrFullSeekKey, p) {
+		if !i.comparer.Split.HasPrefix(i.prefixOrFullSeekKey, upperBound) {
 			i.err = errors.New("pebble: SeekPrefixGE supplied with key outside of upper bound")
 			i.iterValidityState = IterExhausted
 			return false
--- a/level_iter.go
+++ b/level_iter.go
@@ -655,6 +655,9 @@ func (l *levelIter) SeekGE(key []byte, flags base.SeekGEFlags) *base.InternalKV 
 }
 
 func (l *levelIter) SeekPrefixGE(prefix, key []byte, flags base.SeekGEFlags) *base.InternalKV {
+	if invariants.Enabled && !l.split.HasPrefix(prefix, key) {
+		panic(fmt.Sprintf("invalid prefix %q for key %q", prefix, key))
+	}
 	if invariants.Enabled && l.lower != nil && l.cmp(key, l.lower) < 0 {
 		panic(errors.AssertionFailedf("levelIter SeekGE to key %q violates lower bound %q", key, l.lower))
 	}
--- a/merging_iter.go
+++ b/merging_iter.go
@@ -524,11 +524,9 @@ func (m *mergingIter) nextEntry(l *mergingIterLevel, succKey []byte) error {
 	// P2. Care is taken to avoid ever advancing the iterator beyond the current
 	// prefix. If nextEntry is ever invoked while we're already beyond the
 	// current prefix, we're violating the invariant.
-	if invariants.Enabled && m.prefix != nil {
-		if p := m.split.Prefix(l.iterKV.K.UserKey); !bytes.Equal(m.prefix, p) {
-			m.logger.Fatalf("mergingIter: prefix violation: nexting beyond prefix %q; existing heap root %q\n%s",
-				m.prefix, l.iterKV, debug.Stack())
-		}
+	if invariants.Enabled && m.prefix != nil && !m.split.HasPrefix(m.prefix, l.iterKV.K.UserKey) {
+		m.logger.Fatalf("mergingIter: prefix violation: nexting beyond prefix %q; existing heap root %q\n%s",
+			m.prefix, l.iterKV, debug.Stack())
 	}
 
 	oldTopLevel := l.index
@@ -905,6 +903,10 @@ func (m *mergingIter) findPrevEntry() *base.InternalKV {
 //
 // If an error occurs, seekGE returns the error without setting m.err.
 func (m *mergingIter) seekGE(key []byte, level int, flags base.SeekGEFlags) error {
+	if invariants.Enabled && m.lower != nil && m.heap.cmp(key, m.lower) < 0 {
+		m.logger.Fatalf("mergingIter: lower bound violation: %s < %s\n%s", key, m.lower, debug.Stack())
+	}
+
 	// When seeking, we can use tombstones to adjust the key we seek to on each
 	// level. Consider the series of range tombstones:
 	//
@@ -957,15 +959,11 @@ func (m *mergingIter) seekGE(key []byte, level int, flags base.SeekGEFlags) erro
 	}
 
 	for ; level < len(m.levels); level++ {
-		if invariants.Enabled && m.lower != nil && m.heap.cmp(key, m.lower) < 0 {
-			m.logger.Fatalf("mergingIter: lower bound violation: %s < %s\n%s", key, m.lower, debug.Stack())
-		}
-
 		l := &m.levels[level]
 		if m.prefix != nil {
 			l.iterKV = l.iter.SeekPrefixGE(m.prefix, key, flags)
 			if l.iterKV != nil {
-				if !bytes.Equal(m.prefix, m.split.Prefix(l.iterKV.K.UserKey)) {
+				if !m.split.HasPrefix(m.prefix, l.iterKV.K.UserKey) {
 					// Prevent keys without a matching prefix from being added to the heap by setting
 					// iterKey and iterValue to their zero values before calling initMinHeap.
 					l.iterKV = nil
@@ -999,7 +997,21 @@ func (m *mergingIter) seekGE(key []byte, level int, flags base.SeekGEFlags) erro
 				// Based on the containment condition tombstone.End > key, so
 				// the assignment to key results in a monotonically
 				// non-decreasing key across iterations of this loop.
-				//
+				if m.prefix != nil && !m.split.HasPrefix(m.prefix, l.tombstone.End) {
+					// Any keys with m.prefix on subsequent levels are under the tombstone.
+					// We still need to perform the seeks, in case the next seek uses
+					// the TrySeekUsingNext flag.
+					for level++; level < len(m.levels); level++ {
+						l := &m.levels[level]
+						if kv := l.iter.SeekPrefixGE(m.prefix, key, flags); kv == nil {
+							if err := l.iter.Error(); err != nil {
+								return err
+							}
+						}
+						l.iterKV = nil
+					}
+					break
+				}
 				// The adjustment of key here can only move it to a larger key.
 				// Since the caller of seekGE guaranteed that the original key
 				// was greater than or equal to m.lower, the new key will
@@ -1037,6 +1049,9 @@ func (m *mergingIter) SeekPrefixGE(prefix, key []byte, flags base.SeekGEFlags) *
 func (m *mergingIter) SeekPrefixGEStrict(
 	prefix, key []byte, flags base.SeekGEFlags,
 ) *base.InternalKV {
+	if invariants.Enabled && !m.split.HasPrefix(prefix, key) {
+		panic(fmt.Sprintf("invalid prefix %q for key %q", prefix, key))
+	}
 	m.prefix = prefix
 	m.err = m.seekGE(key, 0 /* start level */, flags)
 	if m.err != nil {
@@ -1044,10 +1059,8 @@ func (m *mergingIter) SeekPrefixGEStrict(
 	}
 
 	iterKV := m.findNextEntry()
-	if invariants.Enabled && iterKV != nil {
-		if !bytes.Equal(m.prefix, m.split.Prefix(iterKV.K.UserKey)) {
-			m.logger.Fatalf("mergingIter: prefix violation: returning key %q without prefix %q\n", iterKV, m.prefix)
-		}
+	if invariants.Enabled && iterKV != nil && !m.split.HasPrefix(m.prefix, iterKV.K.UserKey) {
+		m.logger.Fatalf("mergingIter: prefix violation: returning key %q without prefix %q\n", iterKV, m.prefix)
 	}
 	return iterKV
 }
--- a/sstable/reader_iter_single_lvl.go
+++ b/sstable/reader_iter_single_lvl.go
@@ -176,8 +176,13 @@ type singleLevelIterator[D any, PD block.DataBlockIterator[D]] struct {
 	// present, should be used for prefix seeks or not. In some cases it is
 	// beneficial to skip a filter block even if it exists (eg. if probability of
 	// a match is high).
-	useFilterBlock         bool
-	lastBloomFilterMatched bool
+	useFilterBlock bool
+
+	// didNotPositionOnLastSeekGE is set to true if we completed a call to SeekGE
+	// or SeekPrefixGE without positioning the iterator internally. If this flag
+	// is set, the TrySeekUsingNext optimization is disabled on the next seek.
+	// This happens for example when the bloom filter excludes a prefix.
+	didNotPositionOnLastSeekGE bool
 
 	transforms IterTransforms
 
@@ -665,6 +670,11 @@ func (i *singleLevelIterator[D, PD]) SeekGE(key []byte, flags base.SeekGEFlags) 
 			key = i.lower
 		}
 	}
+	if i.didNotPositionOnLastSeekGE {
+		// Iterator is not positioned based on last seek.
+		flags = flags.DisableTrySeekUsingNext()
+		i.didNotPositionOnLastSeekGE = false
+	}
 
 	if flags.TrySeekUsingNext() {
 		// The i.exhaustedBounds comparison indicates that the upper bound was
@@ -817,6 +827,12 @@ func (i *singleLevelIterator[D, PD]) SeekPrefixGE(
 		// TODO(bananabrick): We can optimize away this check for the level iter
 		// if necessary.
 		if i.cmp(key, i.lower) < 0 {
+			if !i.reader.Split.HasPrefix(prefix, i.lower) {
+				i.err = nil // clear any cached iteration error
+				// Disable the TrySeekUsingNext optimization next time.
+				i.didNotPositionOnLastSeekGE = true
+				return nil
+			}
 			key = i.lower
 		}
 	}
@@ -826,6 +842,15 @@ func (i *singleLevelIterator[D, PD]) SeekPrefixGE(
 func (i *singleLevelIterator[D, PD]) seekPrefixGE(
 	prefix, key []byte, flags base.SeekGEFlags,
 ) (kv *base.InternalKV) {
+	if invariants.Enabled && !i.reader.Split.HasPrefix(prefix, key) {
+		panic(fmt.Sprintf("invalid prefix %q for key %q", prefix, key))
+	}
+	if i.didNotPositionOnLastSeekGE {
+		// Iterator is not positioned based on last seek.
+		flags = flags.DisableTrySeekUsingNext()
+		i.didNotPositionOnLastSeekGE = false
+	}
+
 	// NOTE: prefix is only used for bloom filter checking and not later work in
 	// this method. Hence, we can use the existing iterator position if the last
 	// SeekPrefixGE did not fail bloom filter matching.
@@ -833,11 +858,6 @@ func (i *singleLevelIterator[D, PD]) seekPrefixGE(
 	err := i.err
 	i.err = nil // clear cached iteration error
 	if i.useFilterBlock {
-		if !i.lastBloomFilterMatched {
-			// Iterator is not positioned based on last seek.
-			flags = flags.DisableTrySeekUsingNext()
-		}
-		i.lastBloomFilterMatched = false
 		// Check prefix bloom filter.
 		var mayContain bool
 		mayContain, i.err = i.bloomFilterMayContain(prefix)
@@ -848,9 +868,10 @@ func (i *singleLevelIterator[D, PD]) seekPrefixGE(
 			// since the caller was allowed to call Next when SeekPrefixGE returned
 			// nil. This is no longer allowed.
 			PD(&i.data).Invalidate()
+			// Disable the TrySeekUsingNext optimization next time.
+			i.didNotPositionOnLastSeekGE = true
 			return nil
 		}
-		i.lastBloomFilterMatched = true
 	}
 	if flags.TrySeekUsingNext() {
 		// The i.exhaustedBounds comparison indicates that the upper bound was
--- a/sstable/reader_iter_two_lvl.go
+++ b/sstable/reader_iter_two_lvl.go
@@ -32,8 +32,12 @@ type twoLevelIterator[D any, PD block.DataBlockIterator[D]] struct {
 	// useFilterBlock controls whether we consult the bloom filter in the
 	// twoLevelIterator code. Note that secondLevel.useFilterBlock is always
 	// false - any filtering happens at the top level.
-	useFilterBlock         bool
-	lastBloomFilterMatched bool
+	useFilterBlock bool
+	// didNotPositionOnLastSeekGE is set to true if we completed a call to SeekGE
+	// or SeekPrefixGE without positioning the iterator internally. If this flag
+	// is set, the TrySeekUsingNext optimization is disabled on the next seek.
+	// This happens for example when the bloom filter excludes a prefix.
+	didNotPositionOnLastSeekGE bool
 }
 
 var _ Iterator = (*twoLevelIterator[rowblk.Iter, *rowblk.Iter])(nil)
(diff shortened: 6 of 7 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3845)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`merging_iter.go` line 907 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O47FfQW5jcMZg6KNQci:-O47FfQW5jcMZg6KNQcj:bejn95b) ([raw file](https://github.com/cockroachdb/pebble/blob/7cdedc80e5f7a2e2c12364db1f2608cc2e741470/merging_iter.go#L907)):*
> ```Go
> func (m *mergingIter) seekGE(key []byte, level int, flags base.SeekGEFlags) error {
> 	if invariants.Enabled && m.lower != nil && m.heap.cmp(key, m.lower) < 0 {
> 		m.logger.Fatalf("mergingIter: lower bound violation: %s < %s\n%s", key, m.lower, debug.Stack())
> ```

Any good reason for this code to use `Fatalf` instead of `panic`? I'm thinking that `panic` should be just fine under `invariants.Enabled` since it won't happen in production anyway

sumeerbhola (changes requested):
I'm looking at this strict stuff for the first time, and there is a question below related to that and rangedel handling. Rest looks great.

> Reviewed 5 of 7 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3845#-:-O4H8UOE3YYwsOXB97tM:b-7hnvgz)* status: all files reviewed, 7 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`merging_iter.go` line 907 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O47FfQW5jcMZg6KNQci:-O4GrcWVA4AL-w8MCaJz:bk8w15d) ([raw file](https://github.com/cockroachdb/pebble/blob/7cdedc80e5f7a2e2c12364db1f2608cc2e741470/merging_iter.go#L907)):*

Any good reason for this code to use `Fatalf` instead of `panic`? I'm thinking that `panic` should be just fine under `invariants.Enabled` since it won't happen in production anyway
</blockquote></details>

no good reason, afaik.
___
*[`merging_iter.go` line 961 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H4aAMB_sxdJgD6TG3:-O4H4aAMB_sxdJgD6TG4:beb9l2x) ([raw file](https://github.com/cockroachdb/pebble/blob/80a5615349fa999507e3637c6db2eb97583f30fc/merging_iter.go#L961)):*
> ```Go
> 	for ; level < len(m.levels); level++ {
> 		if invariants.Enabled && m.lower != nil && m.heap.cmp(key, m.lower) < 0 {
> 			m.logger.Fatalf("mergingIter: lower bound violation: %s < %s\n%s", key, m.lower, debug.Stack())
> ```

I think `key` is getting updated in this loop, which is why the assertion was inside the loop.
___
*[`merging_iter.go` line 986 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H69puE1B8Qf08inmU:-O4H69puE1B8Qf08inmV:bz62dwk) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/merging_iter.go#L986)):*
> ```Go
> 		// keys, and there might exist live range keys within the range
> 		// tombstone's span that need to be observed to trigger a switch to
> 		// combined iteration.
> ```

Is there an invariant that if we are doing a prefix seek, that `key` is always a prefix of `m.prefix`, even though it may advance in this loop?

Which is why we can't seek using `l.tombstone.End` if it does not match the prefix? Say the prefix was c and we were seeking to c@30. And level 1 landed at c@5, and had a tombstone [a,f). We can't seek the lower levels to f. So we will seek them using c@30. Say level 2 lands at c@20. Now we have in some sense lost the benefit of the tombstone, in that we have not seeked far enough. We won't bother comparing with the tombstone so that's ok-ish. Say the next prefix seek done by the user was c@3. Then level 1 may not have a c, but it still has the tombstone. Then we will seek these lower levels to c@3.

Is this strictness resulting in unnecessary seeking in a part of the key space which is known to be range deleted?
___
*[`sstable/reader_iter_single_lvl.go` line 676 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H1sBIEriQM21oseMe:-O4H1sBIEriQM21oseMf:b-b2d70v) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/sstable/reader_iter_single_lvl.go#L676)):*
> ```Go
> 		// Iterator is not positioned based on last seek.
> 		flags = flags.DisableTrySeekUsingNext()
> 		i.didNotPositionOnLastSeekGE = false
> ```

Is this just a defensive addition, in case we start optimizing by not positioning on the regular `SeekGE` path too?
___
*[`sstable/reader_iter_two_lvl.go` line 40 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4GsSM68JEpmomAaLCN:-O4GsSM71CbKDDoQWR26:b1ct83d) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/sstable/reader_iter_two_lvl.go#L40)):*
> ```Go
> 	// is set, the TrySeekUsingNext optimization is disabled on the next seek.
> 	// This happens for example when the bloom filter excludes a prefix.
> 	didNotPositionOnLastSeekGE bool
> ```

Is it ok for the default value for a new iterator to be false? Will that cause us to assume that it was positioned? Do we get away with it because the caller will not set `seekGEFlagTrySeekUsingNext`?
___
*[`sstable/reader_iter_two_lvl.go` line 404 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H0WKJ3LyTgyvkwO4O:-O4H0WKJ3LyTgyvkwO4P:bdv5h55) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/sstable/reader_iter_two_lvl.go#L404)):*
> ```Go
> 	if flags.TrySeekUsingNext() &&
> 		(i.secondLevel.exhaustedBounds == +1 || (PD(&i.secondLevel.data).IsDataInvalidated() && i.secondLevel.index.IsDataInvalidated())) &&
> 		err == nil {
> ```

How about adding a comment:

// INVARIANT: flags.TrySeekUsingNext(). Which implies `i.didNotPositionOnLastSeekGE` was false at the start of `SeekPrefixGE`, and is still false. So future seeks will continue to be able to use the current position to optimize.
___
*[`sstable/reader_iter_two_lvl.go` line 529 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H3AVI6uWdrwVLhLBg:-O4H3AVI6uWdrwVLhLBh:b-uavlmm) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/sstable/reader_iter_two_lvl.go#L529)):*
> ```Go
> 	}
> 
> 	if !dontSeekWithinSingleLevelIter {
> ```

How about adding something like. Otherwise it is a bit confusing to see `singleLevelIterator` also having a if-block that disables TrySeekUsingNext, and worrying that it will get executed.
```
if invariants.Enabled && i.didNotPositionOnLastSeekGE {
   // i.didNotPositionOnLastSeekGE must always be false, since we've fiddled with that setting, and we want
   // singleLevelIterator to blindly do what we are telling it to in flags.
   panic(...)
}
```

jbowens (commented):
> Reviewed 3 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3845)* status: all files reviewed, 7 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`merging_iter.go` line 986 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H69puE1B8Qf08inmU:-O4HNAHm7aZbDr8_GUK1:bk594ge) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/merging_iter.go#L986)):*

Is there an invariant that if we are doing a prefix seek, that `key` is always a prefix of `m.prefix`, even though it may advance in this loop?

Which is why we can't seek using `l.tombstone.End` if it does not match the prefix? Say the prefix was c and we were seeking to c@30. And level 1 landed at c@5, and had a tombstone [a,f). We can't seek the lower levels to f. So we will seek them using c@30. Say level 2 lands at c@20. Now we have in some sense lost the benefit of the tombstone, in that we have not seeked far enough. We won't bother comparing with the tombstone so that's ok-ish. Say the next prefix seek done by the user was c@3. Then level 1 may not have a c, but it still has the tombstone. Then we will seek these lower levels to c@3.

Is this strictness resulting in unnecessary seeking in a part of the key space which is known to be range deleted?
</blockquote></details>

Hm I think this is true. Could we get away with not performing the remaining lower-level seeks at all and instead remember on their `mergingIterLevel`s that these levels' iterators are unpositioned? I think no because a subsequent SeekGE(k2) may find that k2 is < the position of the higher-level iterators (if they too were re-positioned by range deletes), which could cause  the lower levels' keys to improperly appear to be undeleted by higher-level tombstones.

We would either need to a) remember `l.tombstone.End` so that a subsequent `SeekGE(k2)` seeks to `max(k2,l.tombstone.End)`, or b) maintain an invariant that all the higher levels remain in the heap and their `l.tombstone`s cascade down to `l.tombstone.End` such that a subsequent `SeekGE(k2)` will end up adjusting its seek key to the original `l.tombstone.End` by the time it reaches the unpositioned levels.  
___
*[`sstable/reader_iter_two_lvl.go` line 40 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4GsSM68JEpmomAaLCN:-O4HN22R4snCJgiROrIz:blu5mfz) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/sstable/reader_iter_two_lvl.go#L40)):*
> Do we get away with it because the caller will not set `seekGEFlagTrySeekUsingNext`?

I think this is true. The levelIter takes care to disable the TrySeekUsingNext flag when it opens a new file.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3845)* status: all files reviewed, 7 unresolved discussions (waiting on @sumeerbhola)

___
*[`merging_iter.go` line 961 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H4aAMB_sxdJgD6TG3:-O4HXtfG-333paQU1naN:b-enaozt) ([raw file](https://github.com/cockroachdb/pebble/blob/80a5615349fa999507e3637c6db2eb97583f30fc/merging_iter.go#L961)):*

I think `key` is getting updated in this loop, which is why the assertion was inside the loop.
</blockquote></details>

Right, but we only update it to a larger one.
___
*[`merging_iter.go` line 986 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H69puE1B8Qf08inmU:-O4HYtOB1IT9cCcwY1ae:bz7bkna) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/merging_iter.go#L986)):*

Hm I think this is true. Could we get away with not performing the remaining lower-level seeks at all and instead remember on their `mergingIterLevel`s that these levels' iterators are unpositioned? I think no because a subsequent SeekGE(k2) may find that k2 is < the position of the higher-level iterators (if they too were re-positioned by range deletes), which could cause  the lower levels' keys to improperly appear to be undeleted by higher-level tombstones.

We would either need to a) remember `l.tombstone.End` so that a subsequent `SeekGE(k2)` seeks to `max(k2,l.tombstone.End)`, or b) maintain an invariant that all the higher levels remain in the heap and their `l.tombstone`s cascade down to `l.tombstone.End` such that a subsequent `SeekGE(k2)` will end up adjusting its seek key to the original `l.tombstone.End` by the time it reaches the unpositioned levels.
</blockquote></details>

Would doing `SeekGE(tombstone.End)` on the lower levels be better? AFAICT that's the closest thing to what we were doing before?
___
*[`sstable/reader_iter_single_lvl.go` line 676 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H1sBIEriQM21oseMe:-O4HYMSf4GT48k5JB6DL:b-t6ej0t) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/sstable/reader_iter_single_lvl.go#L676)):*

Is this just a defensive addition, in case we start optimizing by not positioning on the regular `SeekGE` path too?
</blockquote></details>

No, in case we get a `SeekGE w/TrySeekUsingNext` after a `SeekPrefixGE` (which I ran into if I remember correctly).
___
*[`sstable/reader_iter_two_lvl.go` line 40 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4GsSM68JEpmomAaLCN:-O4HZ7yt0NfGPN9ojlf5:bsj7koj) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/sstable/reader_iter_two_lvl.go#L40)):*

> Do we get away with it because the caller will not set `seekGEFlagTrySeekUsingNext`?

I think this is true. The levelIter takes care to disable the TrySeekUsingNext flag when it opens a new file.
</blockquote></details>

Good point though.. I can change it to `allowSeekUsingNext`?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3845)* status: all files reviewed, 7 unresolved discussions (waiting on @sumeerbhola)

___
*[`merging_iter.go` line 986 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H69puE1B8Qf08inmU:-O4HeDQ92VRFb4r_SFx6:bie9mf7) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/merging_iter.go#L986)):*

Would doing `SeekGE(tombstone.End)` on the lower levels be better? AFAICT that's the closest thing to what we were doing before?
</blockquote></details>

Jackson, I don't understand your point - is there a correctness issue if we just remember which levels we are allowed to use TrySeekUsingNext? It seems equivalent to the case where the existing `SeekPrefixGE` did not match the bloom filter.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3845)* status: all files reviewed, 7 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`merging_iter.go` line 986 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H69puE1B8Qf08inmU:-O4INu0IDkzgE2T5HWgB:b-jqaxbk) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/merging_iter.go#L986)):*
> Would doing `SeekGE(tombstone.End)` on the lower levels be better? AFAICT that's the closest thing to what we were doing before?

We may want something that knows the prefix so the `levelIter` will seek the `manifest.LevelIter`, but realizing that the seek key is beyond the prefix, will not load the file containing key, and return nil.

RaduBerinde:
How important is this case to optimize? (given that we now use excise for rebalancing)

jbowens:
> Jackson, I don't understand your point - is there a correctness issue if we just remember which levels we are allowed to use TrySeekUsingNext? It seems equivalent to the case where the existing SeekPrefixGE did not match the bloom filter.

I think the difference is that in the simple bloom filter case, we still seeked all the levels to the correct file. This paragraph in that TrySeekUsingNext PR is referring to this problem but it's very high level: https://github.com/cockroachdb/pebble/pull/3329/files#diff-5f17b9292c13dad71775aec366c4df449009f7fc211d69ed6cb134e6880feff0R64-R69

Ah yeah, here's a more concrete example in the mergingIter comments: https://github.com/cockroachdb/pebble/blob/25ac6781ae36cfb9c81996d4f2db49fb12379882/merging_iter.go#L929-L954

I'm having trouble articulating the invariant we're relying on here. [That's never a good sign :)]

jbowens:
> How important is this case to optimize? (given that we now use excise for rebalancing)

I think range deletions are still common although maybe not particularly broad ones. I believe MVCC GC will write them if there's a sufficiently long span of all keys that need GC-ing. I wouldn't be surprised if it's _not_ important to optimize in practice, but I'm still a little wary of a regression.

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3845)* status: all files reviewed, 7 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`merging_iter.go` line 986 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H69puE1B8Qf08inmU:-O4KvHX357JxJAehBkU6:b89a1i3) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/merging_iter.go#L986)):*

> Would doing `SeekGE(tombstone.End)` on the lower levels be better? AFAICT that's the closest thing to what we were doing before?

We may want something that knows the prefix so the `levelIter` will seek the `manifest.LevelIter`, but realizing that the seek key is beyond the prefix, will not load the file containing key, and return nil.
</blockquote></details>

Hmm, I'm forgetting the benefit of restricting the key in `SeekPrefixGE` such that `prefix` must be a prefix of it.

Also, I was forgetting the motivation for the plan to change `SeekPrefixGE` so that every iterator has `SeekPrefixGEStrict` (actually, is that the plan?) and went back to looking at https://github.com/cockroachdb/pebble/commit/1f7e8ee45c8564509aea7f1f8265904351f755e7 and https://github.com/cockroachdb/pebble/issues/2182 and https://github.com/cockroachdb/pebble/pull/2178#pullrequestreview-1210803906, where we started doing strictness inside `mergingIter` and can't follow the reasoning anymore.

The logic to init a heap is
```
	n := h.len()
	for i := n/2 - 1; i >= 0; i-- {
		h.down(i, n)
	}
```
Say we have a heap of 8 elements (memtable plus all levels with 1 sub-level in L0). I think the min and max number of comparisons is 7 and 11 respectively.

A comparison between a key that does not match the prefix and one that does match the prefix will stop after the same number of bytes as the check that the former key matches the prefix. So prefix checks are not necessarily faster (in one of those threads there was a claim of "faster prefix checks").

For levels with bloom filter checking (everything except memtable, and L6?), the probability that the key satisfies the prefix is very high (let's assume 100% for now) -- I may be forgetting something but I think we don't do an explicit prefix check in the sstable iterators. I think the earlier analysis ignored this high probability.

So let's consider some cases:

Case 1: even with prefix comparisons there are 8 elements

The code before that commit would do [7, 11] key comparisons to build the heap. Now we do 8 comparisons to check the prefix, that will all say yes. Then [7, 11] key comparisons to build the heap. So we are doing more work. In fairness, the old comments do mention the possibility of regression.

Case 2:  prefix comparisons can eliminate memtable and L6, and there are 6 elements left

The code before that commit would do [7, 11] key comparisons to build the heap. Now we have done 8 comparisons for the prefix, and we need to build a heap with 6 elements. I think the latter is [5, 7] comparisons. So the total is still worse.

Case 3: prefix comparisons eliminate everything but L0 and L6

L1-L5 would not have returned a key. So the code before that commit would be working with 3 elements, corresponding to memtable, L0, L6. That is 2 key comparisons to build the heap. Now we do 3 prefix comparisons, and then 1 key comparison to build the heap, for a total of 4.

Case 4: prefix comparisons eliminate everything but L6

L0-L5 would not have returned a key. So the code before that commit would be working with 2 elements, corresponding to memtable, L6. That is 1 key comparison to build the heap. Now we do 2 prefix comparisons, and then no key comparison to build the heap, for a total of 2.

There is also a mention of benefit for iteration "If all of the keys with the prefix reside in a single level, iteration over the prefix also benefits from the reduced heap size removing k log n comparisons, where k is the number of times the iterator is stepped." I am not sure this is precise enough:
If everything except the top of the heap does not have the prefix, they are beyond the prefix. Iteration requires 2 key comparisons when there are at least 3 elements in the heap, since the top won't change. So we do have a benefit if we have managed to filter the heap down to 1 element, or 2 elements, since then iteration is doing 0 or 1 key comparison respectively. I think the CockroachDB cases related to MVCCGet which find a key (common case, I think), and doing a read for a MVCCPut, don't do a Next. If we wanted to optimize this iteration, we could have the mergingIter know (a) which levels already have a high probability of matching the prefix, so it doesn't waste prefix comparisons (b) remember which of the levels that don't have that high probability and for which it has not performed a prefix check. In a Next, it would do a prefix check for (b) and remove the entry if it does not match the prefix. If it does it would flip the bit, so it doesn't bother doing it again while that level iter is not nexted. 

I am probably missing an interesting case, or missing some other obvious thing.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3845)* status: all files reviewed, 7 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`merging_iter.go` line 986 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H69puE1B8Qf08inmU:-O4LGgq48xIF0EMZX-0w:bkv45ix) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/merging_iter.go#L986)):*
> So prefix checks are not necessarily faster (in one of those threads there was a claim of "faster prefix checks").

I haven't benchmarked to measure the difference, but because prefix checks are equality comparisons (eg, we know _n_ the length of the prefix, then we're determining equality of the first _n_ bytes), it seems like it should be faster. If `Split(key) != len(prefix)`, then we can early exit. Otherwise, we're performing an equality comparison, which is generally faster than a comparison.

> For levels with bloom filter checking (everything except memtable, and L6?), the probability that the key satisfies the prefix is very high (let's assume 100% for now)

Hrm, it'd be interesting to have some measure of this: how frequently the bloom filter fails to exclude a table but the table doesn't contain a key. I know the filter effectiveness in terms of percent of tables successfully excluded is high, and that's inversely related to the false positive rate, but I don't know if we can infer much from it without knowing the true frequency of a prefix in multiple levels.

I agree that if we replace the top-level Iterator prefix check with leaf prefix checks that never exclude keys, we're increasing the amount of work. Although with columnar blocks, this [won't be the case](https://github.com/cockroachdb/pebble/pull/3819/commits/838b01bd2337eaec16bdcd7f8dffdc2475445361#diff-11160516463869db3bcdb6aec55fbf8947926eba9bff6ae3fc75e268270bfae5R549-R554) because our initial seek will be performed on the prefixes themselves, and we can implement strictness for free as a part of the block seek.

> So we do have a benefit if we have managed to filter the heap down to 1 element, or 2 elements, since then iteration is doing 0 or 1 key comparison respectively.

I think the benefit is also not just constrained to a single level containing the prefix, but I haven't stepped through the possibilities to bound the comparisons for various combinations of levels containing prefixes or not.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3845)* status: all files reviewed, 7 unresolved discussions (waiting on @sumeerbhola)

___
*[`merging_iter.go` line 986 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H69puE1B8Qf08inmU:-O4LarkiBd9ZTV2VGvxl:bdaxcbe) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/merging_iter.go#L986)):*

> So prefix checks are not necessarily faster (in one of those threads there was a claim of "faster prefix checks").

I haven't benchmarked to measure the difference, but because prefix checks are equality comparisons (eg, we know _n_ the length of the prefix, then we're determining equality of the first _n_ bytes), it seems like it should be faster. If `Split(key) != len(prefix)`, then we can early exit. Otherwise, we're performing an equality comparison, which is generally faster than a comparison.

> For levels with bloom filter checking (everything except memtable, and L6?), the probability that the key satisfies the prefix is very high (let's assume 100% for now)

Hrm, it'd be interesting to have some measure of this: how frequently the bloom filter fails to exclude a table but the table doesn't contain a key. I know the filter effectiveness in terms of percent of tables successfully excluded is high, and that's inversely related to the false positive rate, but I don't know if we can infer much from it without knowing the true frequency of a prefix in multiple levels.

I agree that if we replace the top-level Iterator prefix check with leaf prefix checks that never exclude keys, we're increasing the amount of work. Although with columnar blocks, this [won't be the case](https://github.com/cockroachdb/pebble/pull/3819/commits/838b01bd2337eaec16bdcd7f8dffdc2475445361#diff-11160516463869db3bcdb6aec55fbf8947926eba9bff6ae3fc75e268270bfae5R549-R554) because our initial seek will be performed on the prefixes themselves, and we can implement strictness for free as a part of the block seek.

> So we do have a benefit if we have managed to filter the heap down to 1 element, or 2 elements, since then iteration is doing 0 or 1 key comparison respectively.

I think the benefit is also not just constrained to a single level containing the prefix, but I haven't stepped through the possibilities to bound the comparisons for various combinations of levels containing prefixes or not.
</blockquote></details>

1. The prefix not matching key problem.

The main motivation for me is for understandability of the code. SeekPrefixGE definition says "The supplied prefix will be the prefix of the given key returned by that Split function.".  I am trying to make the code match the definition. If we want to relax the definition, we have to define what the operation even means when the prefix doesn't match, what are the expectations from the iterator, and why would anyone ever call something like that. Is there a reasonable proposal around that definition?

A secondary practical motivation is that we will probably want to keep statistics about bloom filter false positives - i.e. when the bloom filter passes but we find no key with the prefix. We would need to check the prefix to know if really the bloom filter failed to exclude something, or we're looking for an entirely different key.


2. The strictness problem

Like before, the main motivation here is understandability. Some operators implement a `SeekPrefixGEStrict` which leads to a second `TopLevelIterator` interface which makes things more complicated. There are instances in the code where `SeekPrefixGE` is used with the expectation that it is strict: `pebble.Iterator.SeekPrefixGE` is strict and calls `l.iter.SeekPrefixGE` (which is strict because there's a merging iterator under there). This is a mess.

A practical motivation is that the new columnar format allows us to very cheaply know when the prefix changes, so if we push the strictness down to those iterators, we can avoid that comparison altogether. Even without that, prefix checks can be cheaper (like Jackson mentioned above, and also we could do a quick check of the last e.g. 8 bytes of the prefix - we expect some common prefix because the keys are ordered).

Back to the PR: I don't understand all the nuances, but it feels like we want this `SeekPrefixGE(tombstone.End)` call not necessarily for what it does at the sstable level, but for what it does inside the `levelIter`? If that is the case, can't we have a separate method on `levelIter` that does whatever it is that we need? The merging iterator already has a reference to each `*levelIter`.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3845)* status: all files reviewed, 7 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`merging_iter.go` line 986 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H69puE1B8Qf08inmU:-O4MVZ-N98RuONZGm7uZ:b-dynx3) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/merging_iter.go#L986)):*

1. The prefix not matching key problem.

The main motivation for me is for understandability of the code. SeekPrefixGE definition says "The supplied prefix will be the prefix of the given key returned by that Split function.".  I am trying to make the code match the definition. If we want to relax the definition, we have to define what the operation even means when the prefix doesn't match, what are the expectations from the iterator, and why would anyone ever call something like that. Is there a reasonable proposal around that definition?

A secondary practical motivation is that we will probably want to keep statistics about bloom filter false positives - i.e. when the bloom filter passes but we find no key with the prefix. We would need to check the prefix to know if really the bloom filter failed to exclude something, or we're looking for an entirely different key.


2. The strictness problem

Like before, the main motivation here is understandability. Some operators implement a `SeekPrefixGEStrict` which leads to a second `TopLevelIterator` interface which makes things more complicated. There are instances in the code where `SeekPrefixGE` is used with the expectation that it is strict: `pebble.Iterator.SeekPrefixGE` is strict and calls `l.iter.SeekPrefixGE` (which is strict because there's a merging iterator under there). This is a mess.

A practical motivation is that the new columnar format allows us to very cheaply know when the prefix changes, so if we push the strictness down to those iterators, we can avoid that comparison altogether. Even without that, prefix checks can be cheaper (like Jackson mentioned above, and also we could do a quick check of the last e.g. 8 bytes of the prefix - we expect some common prefix because the keys are ordered).

Back to the PR: I don't understand all the nuances, but it feels like we want this `SeekPrefixGE(tombstone.End)` call not necessarily for what it does at the sstable level, but for what it does inside the `levelIter`? If that is the case, can't we have a separate method on `levelIter` that does whatever it is that we need? The merging iterator already has a reference to each `*levelIter`.
</blockquote></details>

I suspect there is a path forward that allows for enforcing that the prefix matches the key, while not becoming a de-optimization. I am mostly thinking about this in the context of the other PR, since the correctness and optimization reasoning need to go hand in hand.

Regarding strictness, I am ok with whatever you all decide. I was merely trying to point out that some of the original performance motivation for strictness *may* be flawed, in that it didn't take into account that the sstable levels, which will typically dominate the levels, may already have a high probability that the key matches the prefix. I see the point of `colblk` allowing for cheap prefix change checks -- that is beneficial for SeekPrefixGE followed by Nexts, but won't help with just SeekPrefixGE (which *may* be a more common case for CRDB's OLTP workloads).

RaduBerinde:
> I see the point of colblk allowing for cheap prefix change checks -- that is beneficial for SeekPrefixGE followed by Nexts, but won't help with just SeekPrefixGE (which may be a more common case for CRDB's OLTP workloads).

I think it should come at no extra cost to keep track of whether the binary search of the prefix ended in an equality or non-equality for SeekPrefixGE.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3845)* status: all files reviewed, 7 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`merging_iter.go` line 986 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3845#-O4H69puE1B8Qf08inmU:-O4koRb3EakNuP_phw6-:bbnwjmq) ([raw file](https://github.com/cockroachdb/pebble/blob/92187e19958f1f7a5906672f60e81fead51f07c7/merging_iter.go#L986)):*
> won't help with just SeekPrefixGE

It still will if bloom filters are imperfect, yes?

PR #3847, opened 2024-08-12
db: fix Open issue with single-item relative paths

### LINKED ISSUE #3509: db: opening in non-read-only mode can fail depending how the path is specified
From `testdata`:
```
$ go run ../cmd/pebble db set db-stage-41 a a
error opening database at "db-stage-41": open : no such file or directory

$ go run ../cmd/pebble db set db-stage-41/ a a
(ok)
```
I think it has to do with the recent WAL changes. CC @jbowens 

Jira issue: PEBBLE-97

### LINKED ISSUE #3842: db: Open fails with 'open : no such file or directory'
If you pass a relative path to Open as "foo", for example, using the pebble command:
```
./pebble bench ycsb --workload=A --num-ops 1 --initial-keys 1000000000 foo
```
it fails with:
```
error opening database at "foo": open : no such file or directory
```

This is because [mkdirAllAndSyncParents](https://github.com/cockroachdb/pebble/blob/9a4f0c8ce0a14117d4f30db8bc4086bbb3816897/checkpoint.go#L90) will attempt to open the empty directory.

I think we should actually be syncing the "." or "/", depending on whether the path was relative or absolute, and the MemFS should tolerate it. Currently the MemFS is particular about paths and does not tolerate paths relative to ".".

Jira issue: PEBBLE-238

jbowens:
Previously Open with a relative path containing a single element would fail on real filesystems, because Open would attempt to open the empty path.

Fix #3509.

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -91,22 +91,19 @@ func mkdirAllAndSyncParents(fs vfs.FS, destDir string) (vfs.File, error) {
 	// Collect paths for all directories between destDir (excluded) and its
 	// closest existing ancestor (included).
 	var parentPaths []string
-	foundExistingAncestor := false
-	for parentPath := fs.PathDir(destDir); parentPath != "."; parentPath = fs.PathDir(parentPath) {
+	for parentPath := fs.PathDir(destDir); ; parentPath = fs.PathDir(parentPath) {
 		parentPaths = append(parentPaths, parentPath)
+		if fs.PathDir(parentPath) == parentPath {
+			break
+		}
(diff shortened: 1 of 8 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3847#-:-O47YKKt8aLzvBG9Wp4S:bnfp4nl)* status: 0 of 8 files reviewed, all discussions resolved (waiting on @sumeerbhola)

jbowens:
TFTR!

PR #3848, opened 2024-08-13
db: refactor disk usage estimate using range annotations

anish-shanbhag:
This change updates `db.EstimateDiskUsage` to use range annotations to
estimate the disk usage of a key range. This should improve the
performance of repeated disk usage estimates for similar or identical
key ranges.

At the Cockroach layer we use `db.EstimateDiskUsage` in a few places,
most notably when [computing MVCC span stats](https://github.com/cockroachdb/cockroach/blob/master/pkg/server/span_stats_server.go#L217).

Informs: cockroachdb#3793

### DIFF
--- a/db.go
+++ b/db.go
@@ -501,6 +501,14 @@ type DB struct {
 			// validating is set to true when validation is running.
 			validating bool
 		}
+
+		// annotators contains various instances of manifest.Annotator which
+		// should be protected from concurrent access.
+		annotators struct {
+			totalSize    *manifest.Annotator[uint64]
+			remoteSize   *manifest.Annotator[uint64]
+			externalSize *manifest.Annotator[uint64]
+		}
 	}
 
 	// Normally equal to time.Now() but may be overridden in tests.
@@ -2228,6 +2236,31 @@ func (d *DB) SSTables(opts ...SSTablesOption) ([][]SSTableInfo, error) {
 	return destLevels, nil
 }
 
+// makeFileSizeAnnotator returns an annotator that computes the total size of
+// files that meet some criteria defined by filter.
+func (d *DB) makeFileSizeAnnotator(filter func(f *fileMetadata) bool) *manifest.Annotator[uint64] {
+	return &manifest.Annotator[uint64]{
+		Aggregator: manifest.SumAggregator{
+			AccumulateFunc: func(f *fileMetadata) (uint64, bool) {
+				if filter(f) {
+					return f.Size, true
+				}
+				return 0, true
+			},
+			AccumulatePartialOverlapFunc: func(f *fileMetadata, bounds base.UserKeyBounds) uint64 {
+				if filter(f) {
+					size, err := d.tableCache.estimateSize(f, bounds.Start, bounds.End.Key)
+					if err != nil {
+						return 0
+					}
+					return size
+				}
+				return 0
+			},
+		},
+	}
+}
+
 // EstimateDiskUsage returns the estimated filesystem space used in bytes for
 // storing the range `[start, end]`. The estimation is computed as follows:
 //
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):
:lgtm: Nice!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3848#-:-O5BKTfm0ASTUhlc9vUd:b-8gk8uj)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @anish-shanbhag, @itsbilal, and @jbowens)

___
*[`db.go` line 508 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3848#-O5BJPFsDWa2PH4IvYmS:-O5BJPFsDWa2PH4IvYmT:b-7nbo6z) ([raw file](https://github.com/cockroachdb/pebble/blob/4c6c47b8a2fdbb1aa7deeb739ddf21537c250a78/db.go#L508)):*
> ```Go
> 		// should be protected from concurrent access.
> 		annotators struct {
> 			totalSizeAnnotator    *manifest.Annotator[uint64]
> ```

[nit] Annotator is redundant given that these are inside `annotators.`
___
*[`internal/manifest/annotator.go` line 281 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3848#-O5BKDvs0pnZLJT80MI8:-O5BKDvs0pnZLJT80MI9:borkreg) ([raw file](https://github.com/cockroachdb/pebble/blob/4c6c47b8a2fdbb1aa7deeb739ddf21537c250a78/internal/manifest/annotator.go#L281)):*
> ```Go
> // for all files within the given Version which are within the range
> // defined by bounds.
> func (a *Annotator[T]) VersionRangeAnnotation(v *Version, bounds base.UserKeyBounds) *T {
> ```

[nit] Why not return `T`?

anish-shanbhag (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3848)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`db.go` line 508 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3848#-O5BJPFsDWa2PH4IvYmS:-O5FIGVfDg0Obqab2CRh:b-ylkqta) ([raw file](https://github.com/cockroachdb/pebble/blob/4c6c47b8a2fdbb1aa7deeb739ddf21537c250a78/db.go#L508)):*

[nit] Annotator is redundant given that these are inside `annotators.`
</blockquote></details>

Thanks, fixed this.
___
*[`internal/manifest/annotator.go` line 281 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3848#-O5BKDvs0pnZLJT80MI8:-O5FILKY3HZg1r_Nj0xd:bvvz37b) ([raw file](https://github.com/cockroachdb/pebble/blob/4c6c47b8a2fdbb1aa7deeb739ddf21537c250a78/internal/manifest/annotator.go#L281)):*

[nit] Why not return `T`?
</blockquote></details>

This is because of the convention I went with in https://github.com/cockroachdb/pebble/pull/3760 where annotators always return a pointer type since a pointer is necessary for the internal caching behavior. Some annotators have a `T = fileMetadata` and in this case we would want to return a `*fileMetadata` instead of `fileMetadata`.

anish-shanbhag (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3848#-:-O5KB6BE5kEMbad6LASy:b-814g8k)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

PR #3853, opened 2024-08-14
crl-release-24.1: export fragmenter.Truncate, MakeInternalKeyTrailer

itsbilal:
This change exports these two methods to unblock https://github.com/cockroachdb/cockroach/pull/129018.


itsbilal:
🤦

PR #3855, opened 2024-08-15
colblk: rework Uint encodings

RaduBerinde:
Currently we have four uint column types; each one can be encoded with
either 0,1,2,4,or 8 bytes per value. The encoding contains a delta
that has the same width as the column type. This leads to a lot of
possible combinations of data formats (each one requiring at least a
bit of specific code, even if generated through generics).

Given that the encoding supports smaller widths transparently, there
is no real advantage to declaring column types for smaller integers.
And there is a disadvantage - it will prevent us from using larger
integers in the future without changing the type.

This change replaces the four uint column types with a single type.
We also add a specialization of the decoding type which is optimized
for offsets.

Benchmarks:
```
name                     time/op
UnsafeUints/const-10     0.59ns ± 0%
UnsafeUints/1b-10        0.70ns ± 0%
UnsafeUints/1b,delta-10  0.70ns ± 1%
UnsafeUints/2b-10        0.74ns ± 0%
UnsafeUints/2b,delta-10  0.74ns ± 0%
UnsafeUints/4b-10        0.94ns ± 0%
UnsafeUints/4b,delta-10  0.94ns ± 0%
UnsafeUints/8b-10        1.25ns ± 0%
UnsafeOffsets/2b-10      0.51ns ± 0%
UnsafeOffsets/4b-10      0.94ns ± 0%
```

### DIFF
--- a/internal/binfmt/binfmt.go
+++ b/internal/binfmt/binfmt.go
@@ -18,7 +18,7 @@ import (
 
 // New constructs a new binary formatter.
 func New(data []byte) *Formatter {
-	offsetWidth := strconv.Itoa(int(math.Log10(float64(len(data)-1))) + 1)
+	offsetWidth := strconv.Itoa(max(int(math.Log10(float64(len(data)-1)))+1, 1))
 	return &Formatter{
 		data:            data,
 		lineWidth:       40,
--- a/sstable/colblk/block.go
+++ b/sstable/colblk/block.go
@@ -352,28 +352,10 @@ func (r *BlockReader) PrefixBytes(col int) PrefixBytes {
 	return DecodeColumn(r, col, int(r.header.Rows), DataTypePrefixBytes, DecodePrefixBytes)
 }
 
-// Uint8s retrieves the col'th column as a column of uint8s. The column must be
-// of type DataTypeUint8.
-func (r *BlockReader) Uint8s(col int) UnsafeUint8s {
-	return DecodeColumn(r, col, int(r.header.Rows), DataTypeUint8, DecodeUnsafeIntegerSlice[uint8])
-}
-
-// Uint16s retrieves the col'th column as a column of uint8s. The column must be
-// of type DataTypeUint16.
-func (r *BlockReader) Uint16s(col int) UnsafeUint16s {
-	return DecodeColumn(r, col, int(r.header.Rows), DataTypeUint16, DecodeUnsafeIntegerSlice[uint16])
-}
-
-// Uint32s retrieves the col'th column as a column of uint32s. The column must be
-// of type DataTypeUint32.
-func (r *BlockReader) Uint32s(col int) UnsafeUint32s {
-	return DecodeColumn(r, col, int(r.header.Rows), DataTypeUint32, DecodeUnsafeIntegerSlice[uint32])
-}
-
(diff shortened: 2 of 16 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3855)* status: 15 of 16 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/colblk/block.go` line 130 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3855#-O4LncvrDJCDZYMQQhl1:-O4LncvrDJCDZYMQQhl2:b-od73ml) ([raw file](https://github.com/cockroachdb/pebble/blob/b79a021200e23ff1d3cd700c8a49e937d427b055/sstable/colblk/block.go#L130)):*
> ```Go
> // Both range deletions and range keys are encoded with the same schema. Range
> // deletion keyspan.Keys never contain suffixes or values. When encoded, the
> // RawBytes encoding uses the UintDeltaEncodingConstant encoding to avoid
> ```

nit: lingering reference to the old enum variant
___
*[`sstable/colblk/block.go` line 132 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3855#-O4LnWrxCc5_vAhb4KNx:-O4LnWrxCc5_vAhb4KNy:bkkbapl) ([raw file](https://github.com/cockroachdb/pebble/blob/b79a021200e23ff1d3cd700c8a49e937d427b055/sstable/colblk/block.go#L132)):*
> ```Go
> // RawBytes encoding uses the UintDeltaEncodingConstant encoding to avoid
> // materializing encoding N offsets. Each of these empty columns is encoded in
> // just ~5 bytes of column data.
> ```

this can be updated now to say 1 byte of column data
___
*[`sstable/colblk/raw_bytes.go` line 37 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3855#-O4Lnmqq9cssIBBtiCuO:-O4Lnmqq9cssIBBtiCuP:b-cu0t5y) ([raw file](https://github.com/cockroachdb/pebble/blob/b79a021200e23ff1d3cd700c8a49e937d427b055/sstable/colblk/raw_bytes.go#L37)):*
> ```Go
> //	|        a uint32 offsets table, possibly delta encoded,            |
> //	|                possibly padded for 32-bit alignment               |
> //	|                      (see DeltaEncoding)                          |
> ```

nit: some lingering references to the old enums and data types in this comment

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3855#-:-O4LrVDr6Ydiwf0Mopde:b-814g8k)* status: 10 of 22 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`sstable/colblk/block.go` line 132 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3855#-O4LnWrxCc5_vAhb4KNx:-O4Lqeyo79XSGwV_lXqR:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b79a021200e23ff1d3cd700c8a49e937d427b055/sstable/colblk/block.go#L132)):*

this can be updated now to say 1 byte of column data
</blockquote></details>

Done.

PR #3856, opened 2024-08-15
colblk: add index block writer, reader, iterator

jbowens:
Add writer, reader and iterator types for index blocks.

### DIFF
--- a/sstable/colblk/index_block.go
+++ b/sstable/colblk/index_block.go
@@ -1,0 +1,297 @@
+// Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package colblk
+
+import (
+	"bytes"
+
+	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/binfmt"
+	"github.com/cockroachdb/pebble/sstable/block"
+)
+
+const indexBlockCustomHeaderSize = 0
+
+// IndexBlockWriter writes columnar index blocks. The writer is used for both
+// first-level and second-level index blocks. The index block schema consists of
+// three primary columns:
+//   - Separators: a user key that is ≥ the largest user key in the
+//     corresponding entry, and ≤ the smallest user key in the next entry.
+//     Note that this allows consecutive separators to be equal. This is
+//     possible when snapshots required we preserve duplicate user keys at
+//     different sequence numbers.
+//   - Offsets: the offset of the end of the corresponding block.
+//   - Lengths: the length in bytes of the corresponding block.
+//   - Block properties: a slice encoding arbitrary user-defined block
+//     properties.
+//
+// TODO(jackson): Consider splitting separators into prefixes and suffixes (even
+// without user-defined columns). This would allow us to use prefix compression
+// for the prefix. Separators should typically be suffixless unless two KVs with
+// the same prefix straddle a block boundary. We would need to use a buffer to
+// materialize the separator key when we need to use it outside the context of
+// seeking within the block.
+type IndexBlockWriter struct {
+	separators      RawBytesBuilder
+	offsets         UintBuilder
+	lengths         UintBuilder
+	blockProperties RawBytesBuilder
+	numColumns      int
+	rows            int
+	enc             blockEncoder
+}
+
+const (
+	indexBlockColumnSeparator = iota
+	indexBlockColumnOffsets
+	indexBlockColumnLengths
+	indexBlockColumnBlockProperties
+	indexBlockColumnCount
+)
+
+// Init initializes the index block writer.
+func (w *IndexBlockWriter) Init(blockProperties ...ColumnWriter) {
+	w.separators.Init()
+	w.offsets.Init()
+	w.lengths.InitWithDefault()
+	w.numColumns = indexBlockColumnCount
+	w.blockProperties.Init()
+	w.rows = 0
+}
+
+// Reset resets the index block writer to its initial state, retaining buffers.
+func (w *IndexBlockWriter) Reset() {
+	w.separators.Reset()
+	w.offsets.Reset()
+	w.lengths.Reset()
+	w.blockProperties.Reset()
+	w.rows = 0
+	w.enc.reset()
+}
+
+// AddBlockHandle adds a new separator and end offset of a data block to the
+// index block.  Add returns the index of the row. The caller should add rows to
+// the block properties columns too if necessary.
+//
+// AddBlockHandle should only be used for first-level index blocks.
+func (w *IndexBlockWriter) AddBlockHandle(
+	separator []byte, handle block.Handle, blockProperties []byte,
+) int {
+	idx := w.rows
+	w.separators.Put(separator)
+	w.offsets.Set(w.rows, handle.Offset)
+	w.lengths.Set(w.rows, handle.Length)
+	w.blockProperties.Put(blockProperties)
+	w.rows++
+	return idx
+}
+
+// Size returns the size of the pending index block.
+func (w *IndexBlockWriter) Size() int {
+	off := blockHeaderSize(w.numColumns, indexBlockCustomHeaderSize)
+	off = w.separators.Size(w.rows, off)
+	off = w.offsets.Size(w.rows, off)
+	off = w.lengths.Size(w.rows, off)
+	off = w.blockProperties.Size(w.rows, off)
+	off++
+	return int(off)
+}
+
+// Finish serializes the pending index block.
+func (w *IndexBlockWriter) Finish() []byte {
+	w.enc.init(w.Size(), Header{
+		Version: Version1,
+		Columns: uint16(w.numColumns),
+		Rows:    uint32(w.rows),
+	}, indexBlockCustomHeaderSize)
+	w.enc.encode(w.rows, &w.separators)
+	w.enc.encode(w.rows, &w.offsets)
+	w.enc.encode(w.rows, &w.lengths)
+	w.enc.encode(w.rows, &w.blockProperties)
+	return w.enc.finish()
+}
+
+// An IndexReader reads columnar index blocks.
+type IndexReader struct {
+	separators RawBytes
+	offsets    UnsafeUints
+	lengths    UnsafeUints // only used for second-level index blocks
+	br         BlockReader
+}
+
+// Init initializes the index reader with the given serialized index block.
+func (r *IndexReader) Init(data []byte) {
+	r.br.Init(data, indexBlockCustomHeaderSize)
+	r.separators = r.br.RawBytes(indexBlockColumnSeparator)
+	r.offsets = r.br.Uints(indexBlockColumnOffsets)
+	r.lengths = r.br.Uints(indexBlockColumnLengths)
+}
+
+// DebugString prints a human-readable explanation of the keyspan block's binary
+// representation.
+func (r *IndexReader) DebugString() string {
+	f := binfmt.New(r.br.data).LineWidth(20)
+	f.CommentLine("index block header")
+	r.br.headerToBinFormatter(f)
+	for i := 0; i < indexBlockColumnCount; i++ {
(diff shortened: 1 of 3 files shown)


RaduBerinde (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3856#-:-O4MXIPN7GYs4B3VzSqr:b-uwkowd)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`sstable/colblk/index_block.go` line 23 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3856#-O4MUSeD4QxqycFprA3Q:-O4MUSeD4QxqycFprA3R:b-z9qv3t) ([raw file](https://github.com/cockroachdb/pebble/blob/7d5a247d8233967f7582c748cf79196e090c60d4/sstable/colblk/index_block.go#L23)):*
> ```Go
> // consists of three primary columns:
> //   - Separators: a user key that is ≥ the largest user key in the corresponding
> //     entry.
> ```

[nit] and < than the smallest key in the next entry ?
___
*[`sstable/colblk/index_block.go` line 24 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3856#-O4MUb-eAFNKFzuKQ-2D:-O4MUb-eAFNKFzuKQ-2E:bigu79f) ([raw file](https://github.com/cockroachdb/pebble/blob/7d5a247d8233967f7582c748cf79196e090c60d4/sstable/colblk/index_block.go#L24)):*
> ```Go
> //   - Separators: a user key that is ≥ the largest user key in the corresponding
> //     entry.
> //   - BlockEndOffset: the offset of the end of the corresponding block.
> ```

[nit] BlockEndOffsets (for consistency)?
___
*[`sstable/colblk/index_block.go` line 30 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3856#-O4MUuX19nRtyko5YC2P:-O4MUuX21LClwZlAhJEm:b-a58a7j) ([raw file](https://github.com/cockroachdb/pebble/blob/7d5a247d8233967f7582c748cf79196e090c60d4/sstable/colblk/index_block.go#L30)):*
> ```Go
> // iterators infer the start offset of the block by looking at the end offset of
> // the previous block. This is possible because data blocks are written
> // sequentially without any other interleaved blocks between first-level index
> ```

Do we want to lock ourselves into this constraint? There's an argument for not keeping all index blocks in memory when building ssts. Are the savings here significant? It's just 2 bytes per ~32KB data block no?
___
*[`sstable/colblk/index_block.go` line 244 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3856#-O4MX9Su88rjBFOjzrlA:-O4MX9Su88rjBFOjzrlB:b-hhqgzq) ([raw file](https://github.com/cockroachdb/pebble/blob/7d5a247d8233967f7582c748cf79196e090c60d4/sstable/colblk/index_block.go#L244)):*
> ```Go
> }
> 
> // SeekGE seeks the index iterator to the block entry with a separator key
> ```

[nit] the first block entry

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3856)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`sstable/colblk/index_block.go` line 30 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3856#-O4MUuX19nRtyko5YC2P:-O4Q3k8iCL-4qLkExIf7:bdg6ios) ([raw file](https://github.com/cockroachdb/pebble/blob/7d5a247d8233967f7582c748cf79196e090c60d4/sstable/colblk/index_block.go#L30)):*

Do we want to lock ourselves into this constraint? There's an argument for not keeping all index blocks in memory when building ssts. Are the savings here significant? It's just 2 bytes per ~32KB data block no?
</blockquote></details>

Do we need to store the lengths at all? (Is this what your question is Radu?) I seem to recall that RocksDB created a new index block format which had an alternate encoding for block handles which computed the length of a block from the offset of the next block. Perhaps I'm misremembering, but that idea should work here. Isn't it true that `BlockLength(i) = BlockOffset(i+1) - BlockOffset(i)`?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3856)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @itsbilal, @petermattis, and @RaduBerinde)

___
*[`sstable/colblk/index_block.go` line 23 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3856#-O4MUSeD4QxqycFprA3Q:-O4QBYeD6jksOA-88M4x:buyvwt8) ([raw file](https://github.com/cockroachdb/pebble/blob/7d5a247d8233967f7582c748cf79196e090c60d4/sstable/colblk/index_block.go#L23)):*

[nit] and < than the smallest key in the next entry ?
</blockquote></details>

it's not guaranteed, because with snapshots we can have many versions of the same user key at different sequence numbers, which can cause adjacent blocks to have the same separator
___
*[`sstable/colblk/index_block.go` line 24 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3856#-O4MUb-eAFNKFzuKQ-2D:-O4QBXeU7t09R7DQJtWk:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/7d5a247d8233967f7582c748cf79196e090c60d4/sstable/colblk/index_block.go#L24)):*

[nit] BlockEndOffsets (for consistency)?
</blockquote></details>

Done
___
*[`sstable/colblk/index_block.go` line 30 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3856#-O4MUuX19nRtyko5YC2P:-O4QBtfKBjVv-cJZcARG:bdqhife) ([raw file](https://github.com/cockroachdb/pebble/blob/7d5a247d8233967f7582c748cf79196e090c60d4/sstable/colblk/index_block.go#L30)):*
> By the way, can't value blocks be interspersed between data blocks?

Naw, today they're all buffered and then written at the tail of the sstable, with the intention of preserving locality of key data (eg, for readahead).

> Do we need to store the lengths at all? (Is this what your question is Radu?) I seem to recall that RocksDB created a new index block format which had an alternate encoding for block handles which computed the length of a block from the offset of the next block. Perhaps I'm misremembering, but that idea should work here. Isn't it true that `BlockLength(i) = BlockOffset(i+1) - BlockOffset(i)`?

@petermattis the original commit (7d5a247d8233967f7582c748cf79196e090c60d4) here implemented what you described, but it was a little cumbersome because second-level indexes don't have that guarantee `BlockLength(i) = BlockOffset(i+1) - BlockOffset(i)`, and first-level index blocks other than the first also need to encode an initial start offset.

Like @RaduBerinde mentioned, our uint encoding does ensure that in the common case of blocks all around ~32KB, the lengths column will be just 2-bytes per block
___
*[`sstable/colblk/index_block.go` line 244 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3856#-O4MX9Su88rjBFOjzrlA:-O4QBX5QE8Ru6YqR7Z9J:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/7d5a247d8233967f7582c748cf79196e090c60d4/sstable/colblk/index_block.go#L244)):*

[nit] the first block entry
</blockquote></details>

Done

RaduBerinde:
*[`sstable/colblk/index_block.go` line 23 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3856#-O4MUSeD4QxqycFprA3Q:-O4QFtiH6sq0XxRAXh5c:b963jyf) ([raw file](https://github.com/cockroachdb/pebble/blob/7d5a247d8233967f7582c748cf79196e090c60d4/sstable/colblk/index_block.go#L23)):*

it's not guaranteed, because with snapshots we can have many versions of the same user key at different sequence numbers, which can cause adjacent blocks to have the same separator
</blockquote></details>

then let's say "and <= the smallest key in the next" and add the explanation above for the equality case.

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3856)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @itsbilal, @jbowens, and @RaduBerinde)

___
*[`sstable/colblk/index_block.go` line 30 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3856#-O4MUuX19nRtyko5YC2P:-O4R7evjEco4oZlVQpVs:bma1p3k) ([raw file](https://github.com/cockroachdb/pebble/blob/7d5a247d8233967f7582c748cf79196e090c60d4/sstable/colblk/index_block.go#L30)):*

> By the way, can't value blocks be interspersed between data blocks?

Naw, today they're all buffered and then written at the tail of the sstable, with the intention of preserving locality of key data (eg, for readahead).

> Do we need to store the lengths at all? (Is this what your question is Radu?) I seem to recall that RocksDB created a new index block format which had an alternate encoding for block handles which computed the length of a block from the offset of the next block. Perhaps I'm misremembering, but that idea should work here. Isn't it true that `BlockLength(i) = BlockOffset(i+1) - BlockOffset(i)`?

@petermattis the original commit (7d5a247d8233967f7582c748cf79196e090c60d4) here implemented what you described, but it was a little cumbersome because second-level indexes don't have that guarantee `BlockLength(i) = BlockOffset(i+1) - BlockOffset(i)`, and first-level index blocks other than the first also need to encode an initial start offset.

Like @RaduBerinde mentioned, our uint encoding does ensure that in the common case of blocks all around ~32KB, the lengths column will be just 2-bytes per block
</blockquote></details>

Got it. Thanks for the context. Carry on.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3856)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`sstable/colblk/index_block.go` line 32 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3856#-O4RDitq02mEstpuAOQk:-O4RDitq02mEstpuAOQl:b-sfn1f2) ([raw file](https://github.com/cockroachdb/pebble/blob/7d5a247d8233967f7582c748cf79196e090c60d4/sstable/colblk/index_block.go#L32)):*
> ```Go
> // sequentially without any other interleaved blocks between first-level index
> // blocks.
> //
> ```

Will we add a column for block props here?

jbowens (commented):
Updated so that the `colblk.IndexBlockIter` satisfies the new `block.IndexBlockIterator` constraint

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3856#-:-O4vUF2uErkN10hcefX2:bcagx7v)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`sstable/colblk/index_block.go` line 23 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3856#-O4MUSeD4QxqycFprA3Q:-O4vUNQu6ADQ_FlCHv9S:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/7d5a247d8233967f7582c748cf79196e090c60d4/sstable/colblk/index_block.go#L23)):*

then let's say "and <= the smallest key in the next" and add the explanation above for the equality case.
</blockquote></details>

Done
___
*[`sstable/colblk/index_block.go` line 32 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3856#-O4RDitq02mEstpuAOQk:-O4vUOqFAGqEGNVKS61j:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/7d5a247d8233967f7582c748cf79196e090c60d4/sstable/colblk/index_block.go#L32)):*

Will we add a column for block props here?
</blockquote></details>

Done

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3856#-:-O4wC8UkBTSMDXBfQn15:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

jbowens:
TFTR!

PR #3862, opened 2024-08-19
db: fix WAL offset in replayWAL errors

jbowens:
If an error occurs while replaying the WALs during Open, the error is annotated with the log number and the offset within the log. Previously, the offset was always zero. This also caused the queued flushables to be created with a zero logSize. This didn't matter in practice because the queued flushables would be flushed before Open returned.

This commit adds a PreviousFilesBytes total to the wal.Offset type to describe the number of bytes replayed from previous WAL segement files.

### DIFF
--- a/open.go
+++ b/open.go
@@ -771,7 +771,7 @@ func (d *DB) replayWAL(
 		buf             bytes.Buffer
 		mem             *memTable
 		entry           *flushableEntry
-		offset          int64 // byte offset in rr
+		offset          wal.Offset
 		lastFlushOffset int64
 		keysReplayed    int64 // number of keys replayed
 		batchesReplayed int64 // number of batches replayed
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3862#-:-O4fD3825671MBcDh1JT:bnfp4nl)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`wal/wal.go` line 437 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3862#-O4fCweB0f3jFHC60WOM:-O4fCweB0f3jFHC60WON:b7hifx5) ([raw file](https://github.com/cockroachdb/pebble/blob/732a444f8fb30abb57972f573eeb1189c60b843e/wal/wal.go#L437)):*
> ```Go
> // offset.
> func (o Offset) String() string {
> 	return fmt.Sprintf("(%s: %d), %d from previous files", o.PhysicalFile, o.Physical, o.PreviousFilesBytes)
> ```

[nit] you can consider omitting the new part when `PrevioousFilesBytes == 0`

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3862#-:-O4fgc1t5yipOPPIlV6k:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

jbowens:
TFTRs!

PR #3864, opened 2024-08-19
tool: fix up wal dump, add wal dump-merged

jbowens:
Adjust the wal dump tool to error on unrecognized key kinds, and surface errors
encounted more prominently in the output.

Additionally, add a new `wal dump-merged` command that dumps the merged view of
a WAL that's composed of multiple physical segment files.

Informs #3865.

### DIFF
--- a/tool/testdata/wal_dump
+++ b/tool/testdata/wal_dump
@@ -6,15 +6,15 @@ wal dump
 ../testdata/db-stage-2/000002.log
 ----
 000002.log
-0(21) seq=10 count=1
+0(21) seq=10 count=1, len=21
     SET(test formatter: foo,test value formatter: one)
-32(21) seq=11 count=1
+32(21) seq=11 count=1, len=21
     SET(test formatter: bar,test value formatter: two)
-64(23) seq=12 count=1
+64(23) seq=12 count=1, len=23
     SET(test formatter: baz,test value formatter: three)
-98(22) seq=13 count=1
+98(22) seq=13 count=1, len=22
     SET(test formatter: foo,test value formatter: four)
-131(17) seq=14 count=1
+131(17) seq=14 count=1, len=17
     DEL(test formatter: bar)
 EOF
 
@@ -24,15 +24,15 @@ wal dump
 --value=size
 ----
 000002.log
-0(21) seq=10 count=1
+0(21) seq=10 count=1, len=21
     SET(foo,<3>)
-32(21) seq=11 count=1
+32(21) seq=11 count=1, len=21
     SET(bar,<3>)
-64(23) seq=12 count=1
+64(23) seq=12 count=1, len=23
     SET(baz,<5>)
-98(22) seq=13 count=1
+98(22) seq=13 count=1, len=22
     SET(foo,<4>)
-131(17) seq=14 count=1
+131(17) seq=14 count=1, len=17
     DEL(bar)
 EOF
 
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3864#-:-O4g3uCT-oV3HxVU3W2O:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aadityasondhi)

jbowens (commented):
I ended up expanding this while debugging #3865 so that I could reproduce #3865 using the debug tool.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3864#-:-O4klwG61XnHA0fRlLIv:bx3f7ra)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @aadityasondhi and @RaduBerinde)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3864)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @aadityasondhi and @jbowens)

___
*[`tool/wal.go` line 209 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3864#-O4lBwUa322bg4-SFw0w:-O4lBwUa322bg4-SFw0x:b7iccuc) ([raw file](https://github.com/cockroachdb/pebble/blob/221f62e7a357cf6b55cd7156a89e38540ec323e6/tool/wal.go#L209)):*
> ```Go
> 	errf := func(offset wal.Offset, format string, args ...interface{}) {
> 		prefix := fmt.Sprintf("%s: error: ", offset)
> 		fmt.Fprintf(stderr, prefix+format+"\n", args...)
> ```

This kind of non-constant format string can trip up newer versions of lint checks (see https://github.com/cockroachdb/cockroach/pull/129208).

I'd just pass an `error` and use `errors.Wrap` at the callsite and here to add info.
___
*[`tool/wal.go` line 289 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3864#-O4lBTcI4BsUyCSmukPj:-O4lBTcI4BsUyCSmukPk:b-mescit) ([raw file](https://github.com/cockroachdb/pebble/blob/221f62e7a357cf6b55cd7156a89e38540ec323e6/tool/wal.go#L289)):*
> ```Go
> 			s, err := rangekey.Decode(ik, value, nil)
> 			if err != nil {
> 				errf("%s: error decoding %s", w.fmtKey.fn(ukey), err)
> ```

[nit] `errf` prints to `stdout` with a newline, unlike the other cases here. Maybe keep the `fmt.Fprintf(stdout` here and remove it from `errf`.

If it's not hard, it would be nice to have this case show up in the test.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3864)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @aadityasondhi and @RaduBerinde)

___
*[`tool/wal.go` line 209 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3864#-O4lBwUa322bg4-SFw0w:-O4qAZjr4FzVWneycUmp:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/221f62e7a357cf6b55cd7156a89e38540ec323e6/tool/wal.go#L209)):*

This kind of non-constant format string can trip up newer versions of lint checks (see https://github.com/cockroachdb/cockroach/pull/129208).

I'd just pass an `error` and use `errors.Wrap` at the callsite and here to add info.
</blockquote></details>

Done
___
*[`tool/wal.go` line 289 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3864#-O4lBTcI4BsUyCSmukPj:-O4uje_586Zcm_xlxRcz:b3i9kvb) ([raw file](https://github.com/cockroachdb/pebble/blob/221f62e7a357cf6b55cd7156a89e38540ec323e6/tool/wal.go#L289)):*

[nit] `errf` prints to `stdout` with a newline, unlike the other cases here. Maybe keep the `fmt.Fprintf(stdout` here and remove it from `errf`.

If it's not hard, it would be nice to have this case show up in the test.
</blockquote></details>

Done

I gave the test case a stab but aborted for now because it is tricky

Issue #3865, opened 2024-08-19
wal: reader batch corruption


jbowens:
While stitching together physical WAL segments, the WAL reader reads each record's contents into a temporary buffer. There's a bug where an error encountered mid-copy of a record (eg, because the record was large enough to be
split across multiple chunks), the buffer was not reset before attempting to read the next file's record. This corruption could prevent replay of the WAL.

- [x] fix wal.Reader
- [ ] add randomized test of WAL failover and recovery

Jira issue: PEBBLE-244

dshjoshi:
What sequence of user actions led to this issue being uncovered?

dshjoshi:
@jbowens Have we already added " add randomized test of WAL failover and recovery"?

jbowens:
Yes

PR #3868, opened 2024-08-20
crl-release-24.1: wal: fix reader buffer reuse

jbowens:
While stitching together physical WAL segments, the WAL reader reads each record's contents into a temporary buffer. Previously, if an error was encountered mid-copy of a record (eg, because the record was large enough to be split across multiple chunks), the buffer was not reset before attempting to read the next file's record. This corruption could prevent replay of the WAL.

This commit fixes the bug moving the buffer reset to immediately precede its use.

Informs #3865.

### DIFF
--- a/wal/reader.go
+++ b/wal/reader.go
@@ -241,8 +241,6 @@ var _ Reader = (*virtualWALReader)(nil)
 // are no more records. The reader returned becomes stale after the next
 // NextRecord call, and should no longer be used.
 func (r *virtualWALReader) NextRecord() (io.Reader, Offset, error) {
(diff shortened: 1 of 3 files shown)


RaduBerinde (approved):

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3868#-:-O4kf6SL6_NpcJu8zdSs:bg7igkv)* status: 0 of 3 files reviewed, all discussions resolved

jbowens:
TFTRs!

PR #3869, opened 2024-08-20
colblk: special case zero bitmaps

RaduBerinde:
Some bitmaps are going to be all-zeros in common cases, for example
the `isValueExternal` bitmap. It is worth special-casing these to
speed up the decoding.

We also improve `Bitmap.At` a bit using a shift and a mask to get the
offset to the respective word.


jbowens (commented):
> Reviewed 2 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3869)* status: 2 of 8 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/colblk/bitmap.go` line 264 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3869#-O4q5oKf5Oizycm55jNB:-O4q5oKf5Oizycm55jNC:b-w56395) ([raw file](https://github.com/cockroachdb/pebble/blob/04617ecaab943eb07c5bf5c3c27fe9872f4de021/sstable/colblk/bitmap.go#L264)):*
> ```Go
> 	} else {
> 		b.unset(i)
> 	}
> ```

did these changes move the microbenchmarks at all?
___
*[`sstable/colblk/bitmap.go` line 310 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3869#-O4q6buuCpTxiYVamyIT:-O4q6buuCpTxiYVamyIU:b-lyqh4e) ([raw file](https://github.com/cockroachdb/pebble/blob/04617ecaab943eb07c5bf5c3c27fe9872f4de021/sstable/colblk/bitmap.go#L310)):*
> ```Go
> 	// First byte will be the encoding type.
> 	offset++
> 	if b.IsZero() {
> ```

`Size` is a bit performance sensitive during data block building, because we need to call it after every KV to determine whether we should finish the block. does the `IsZero` call have a perf impact? Should we consider maintaining a flag indicating whether any bits are set?

When I was iterating on block building  perf, I avoided ever calling `Unset` instead relying on all bits being implicitly unset. We could make that part of the contract, remove `unset` and make `IsZero() == (len(b.words)==0)`
___
*[`sstable/colblk/data_block.go` line 488 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3869#-O4q7c6E1M9x95TboTaN:-O4q7c6E1M9x95TboTaO:b-q511aa) ([raw file](https://github.com/cockroachdb/pebble/blob/04617ecaab943eb07c5bf5c3c27fe9872f4de021/sstable/colblk/data_block.go#L488)):*
> ```Go
> 	// Invert the prefix-same bitmap before writing it out, because we want it
> 	// to represent when the prefix changes.
> 	w.prefixSame.Invert(w.rows)
> ```

hrm, the `Size()` calls we make along the way will no longer be accurate since the size of the pre-inversion bitmap is not the same as the post-inversion bitmap. we could add support for 'all ones' to make it symmetric, or we could explicitly account for it in `DataBlockWriter.Size()`

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3869)* status: 0 of 10 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`sstable/colblk/data_block.go` line 488 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3869#-O4q7c6E1M9x95TboTaN:-O4qFT401SphN15fGi-K:bsmnhpg) ([raw file](https://github.com/cockroachdb/pebble/blob/04617ecaab943eb07c5bf5c3c27fe9872f4de021/sstable/colblk/data_block.go#L488)):*

hrm, the `Size()` calls we make along the way will no longer be accurate since the size of the pre-inversion bitmap is not the same as the post-inversion bitmap. we could add support for 'all ones' to make it symmetric, or we could explicitly account for it in `DataBlockWriter.Size()`
</blockquote></details>

Makes sense. I switched to only allowing setting bits and added an `InvertedSize()`. Added some more checks in various tests.

jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3869#-:-O4qrn6I5wKdCMNl0E8P:bnfp4nl)* status: 0 of 10 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/colblk/bitmap.go` line 308 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3869#-O4qrk5REzItyldDMHb7:-O4qrk5REzItyldDMHb8:bpzilqj) ([raw file](https://github.com/cockroachdb/pebble/blob/4e48f228f8b0d579857a3d57211458e780f66538/sstable/colblk/bitmap.go#L308)):*
> ```Go
> //
> // Note that Invert can affect the Size of the bitmap. Use InvertedSize() if you
> // intend to invert the buitmap before finishing.
> ```

nit: "bitmap"

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3869#-:-O4rRjLCEGbBuXe4a1CH:b-814g8k)* status: 0 of 10 files reviewed, 3 unresolved discussions (waiting on @jbowens)

PR #3872, opened 2024-08-21
colblk: use regular slice in PrefixBytesIter

RaduBerinde:
We currently store an `unsafe.Pointer` and separate length and
capacities. In many cases in the hot path we have to create a slice
using `unsafe.Slice`. However `unsafe.Slice` includes some checks that
we don't need (see [here](https://github.com/golang/go/blob/f38d42f2c4c6ad0d7cbdad5e1417cac3be2a5dcb/src/runtime/unsafe.go#L53)).

In this change we switch to storing a regular slice and converting to
an `unsafe.Pointer` using `unsafe.SliceData` (which directly accesses
the pointer field). This improves performance, even while gaining
boundary checks when appending the suffix.

```
CockroachDataBlockIter/AlphaLen=4,Shared=8,PrefixLen=32,Logical=0,value=8/Next    11.4ns ± 2%    10.6ns ± 1%  -7.32%  (p=0.008 n=5+5)
```

### DIFF
--- a/sstable/colblk/cockroach_test.go
+++ b/sstable/colblk/cockroach_test.go
@@ -265,25 +265,42 @@ func (ks *cockroachKeySeeker) MaterializeUserKey(ki *PrefixBytesIter, prevRow, r
 		ks.prefixes.SetAt(ki, row)
 	}
 
+	ptr := unsafe.Pointer(uintptr(unsafe.Pointer(unsafe.SliceData(ki.buf))) + uintptr(len(ki.buf)))
 	mvccWall := ks.mvccWallTimes.At(row)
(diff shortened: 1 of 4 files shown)


jbowens (approved):
:lgtm:

nice!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3872#-:-O4q1aC-4jHJOgKhrQ5f:blzn4tv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #3874, opened 2024-08-21
lint: remove golint

RaduBerinde:
golint is deprecated and frozen since 2021. I don't think it provides
much value on top of go vet and staticcheck and it is very slow (it
currently takes ~90 seconds on my machine).

This change removes `TestGolint`.


jbowens (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3874)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nicktrav)

PR #3878, opened 2024-08-21
colblk: store shared and bundle prefix length

RaduBerinde:
This PR is only for the last commit.

#### colblk: store shared and bundle prefix length

We currently store the shared prefix length and bundle length
separately and have to add them every time we move to a new key. We
switch to storing their sum.

Together with #3877, #3869, #3872 we get about 15-20% improvement on Next benchmarks and are now ~on par with the row iterator.
Full results on AMD64 (Ryzen 9):
https://gist.github.com/RaduBerinde/d94b28f87943d9258f950c8588cb6a98

### DIFF
--- a/sstable/colblk/prefix_bytes.go
+++ b/sstable/colblk/prefix_bytes.go
@@ -241,11 +241,10 @@ func (b PrefixBytes) At(i int) []byte {
 type PrefixBytesIter struct {
 	// buf is used for materializing a user key. It is preallocated to the maximum
 	// key length in the data block.
-	buf             []byte
-	bundlePrefixLen uint32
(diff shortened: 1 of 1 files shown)


jbowens (approved):
:lgtm_strong: 

Good idea!

> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3878#-:-O4v4A_K4MyDrwQgci9c:b-bi7fba)* status: 1 of 12 files reviewed, all discussions resolved

RaduBerinde:
TFTR!

PR #3887, opened 2024-08-24
vfs: make Unwrap part of FS, fix metamorphic restart op

RaduBerinde:
The metamorphic restart op never resolved the "base" `MemFS` because
`diskHealthCheckingFS` does not implement `Unwrap`. The code path
where we discard writes done by Close was inactive.

This change makes `Unwrap` part of `FS` and implements it correctly in
`diskHealthCheckingFS`.

In the metamorphic test, we now support the restartDB operation even
when the FS is not strict. When it is strict, we discard unsynced data
prior to Close. We no longer silently tolerate `Root` returning
something other than `*MemFS` in `strictFS` mode.


sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3887)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`metamorphic/test.go` line 276 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3887#-O558wwkBMukBabG_Pjy:-O558wwkBMukBabG_Pjz:b-gjxj0h) ([raw file](https://github.com/cockroachdb/pebble/blob/4c68ed5bb368e9f1adeaee8403db414feae27dbc/metamorphic/test.go#L276)):*
> ```Go
> 
> 	var memFS *vfs.MemFS
> 	if !t.testOpts.strictFS {
> ```

Shouldn't this say `if t.testOpts.strictFS`?
I am surprised this did not panic when calling `SetIgnoreSyncs` because of it not being a strict `MemFS`.

Oh, this "worked" but is actually broken because we return earlier in this function due to
```
if !t.testOpts.strictFS {
		return nil
	}
```
And because of this new if-block we will never do ignore syncs.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3887)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`metamorphic/test.go` line 276 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3887#-O558wwkBMukBabG_Pjy:-O55RGx140a_Evrt1ZG3:b-n00d1t) ([raw file](https://github.com/cockroachdb/pebble/blob/4c68ed5bb368e9f1adeaee8403db414feae27dbc/metamorphic/test.go#L276)):*

Shouldn't this say `if t.testOpts.strictFS`?
I am surprised this did not panic when calling `SetIgnoreSyncs` because of it not being a strict `MemFS`.

Oh, this "worked" but is actually broken because we return earlier in this function due to
```
if !t.testOpts.strictFS {
		return nil
	}
```
And because of this new if-block we will never do ignore syncs.
</blockquote></details>

*faceslap*  Fixed it.. but now I get meta failures.. I'll have to investigate.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3887)* status: 5 of 7 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`metamorphic/test.go` line 276 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3887#-O558wwkBMukBabG_Pjy:-O55WAvvCyjYqP8Q33bs:b-7brakd) ([raw file](https://github.com/cockroachdb/pebble/blob/4c68ed5bb368e9f1adeaee8403db414feae27dbc/metamorphic/test.go#L276)):*

*faceslap*  Fixed it.. but now I get meta failures.. I'll have to investigate.
</blockquote></details>

Oh, I see. Without `strictFS`, we are using `NoSync` for write options, so we can't really restart the database and expect a deterministic outcome.

Changing it back to only run this op in `strictFS` mode. I'm still running into some failures that I'm investigating.

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3887#-:-O55wlXOEnZQ7wCpoTon:bg7igkv)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @jbowens)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3887)* status: 4 of 7 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`metamorphic/test.go` line 276 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3887#-O558wwkBMukBabG_Pjy:-O562DrMEJQvADVpGRN_:bzbhase) ([raw file](https://github.com/cockroachdb/pebble/blob/4c68ed5bb368e9f1adeaee8403db414feae27dbc/metamorphic/test.go#L276)):*

Oh, I see. Without `strictFS`, we are using `NoSync` for write options, so we can't really restart the database and expect a deterministic outcome.

Changing it back to only run this op in `strictFS` mode. I'm still running into some failures that I'm investigating.
</blockquote></details>

Added a couple more exceptions and a fix for write-after-EOF in MemFS, seems to work now.

jbowens (approved):
:lgtm:

> Reviewed 4 of 6 files at r1, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3887#-:-O5Dw0MyEALSjaicH1wH:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

PR #3888, opened 2024-08-26
vfs: redesign MemFS strict mode

RaduBerinde:
Currently we use `MemFS` in "strict" mode to test crash recovery in
the following way:
 - at the desired crash point we call `SetIgnoreSyncs(true)`
 - after we close the database, we call `ResetToSyncedState()` and
   `SetIgnoreSyncs(false)` and proceed using the same filesystem.

This model is a bit fragile in the sense that both the previous
operation that we're simulating a crash of and the new operation use
the same filesystem. For example, a background operation that is
finishing up some cleanup could in principle interfere with the new
process.

We switch to a "crash clone" model, where we instead extract a
crash-consistent copy of the filesystem; further testing can proceed
on this independent copy. This allows for more usage patterns - e.g.
we can take multiple crash clones at various points and check them all
afterwards.

We also add functionality to randomly retain part of the unsynced
data (which is closer to what would happen in a real crash).

### DIFF
--- a/checkpoint_test.go
+++ b/checkpoint_test.go
@@ -327,7 +327,7 @@ func TestCheckpointCompaction(t *testing.T) {
 
 func TestCheckpointFlushWAL(t *testing.T) {
 	const checkpointPath = "checkpoints/checkpoint"
-	fs := vfs.NewStrictMem()
+	fs := vfs.NewCrashableMem()
 	opts := &Options{FS: fs, Logger: testLogger{t: t}}
 	key, value := []byte("key"), []byte("value")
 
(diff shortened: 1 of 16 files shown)


jbowens (approved):
:lgtm:

> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3888#-:-O5FA_Qm19C1xCFmo5jF:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`error_test.go` line 436 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3888#-O5F9gLM-9EBmyHkw9P3:-O5F9gLM-9EBmyHkw9P4:bc8haqo) ([raw file](https://github.com/cockroachdb/pebble/blob/4977c3c73cc751368714d227dc33f22fa2c07ce6/error_test.go#L436)):*
> ```Go
> 		inj := errorfs.InjectorFunc(func(op errorfs.Op) error {
> 			if op.Kind.ReadOrWrite() == errorfs.OpIsWrite && crashIndex.Add(-1) == -1 {
> 				// Allow an arbitrary subset of non-0synced state to survive beyond the
> ```

nit: "non-synced"

RaduBerinde:
TFTR!

PR #3897, opened 2024-08-28
db: allow excises to unconditionally be flushable ingests

### LINKED ISSUE #2676: ingest: Allow excises with flushable ingests

itsbilal:
Previously, we'd only allow IngestAndExcise to become flushable ingests if we had a guarantee from the caller that the ingestion contained rangedel/rangekeydel overlapping the excise span. This change reworks flushable ingests to return rangedels/rangekeydels from their own iterators, even if one does not exist in the sstables themselves. This, plus the update to the WAL to be able to persist the excise span, allows for IngestAndExcise to allow flushable ingests in more cases than before.

Also resolves TODOs in replayWAL() around properly sequencing flushable ingests so their sstables can go in levels lower than L0. This sequencing was necessary to make excises work correctly; as excises require an up-to-date view of the version to work correctly.

Fixes #2676.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -552,6 +552,12 @@ func (b *Batch) refreshMemTableSize() error {
 			}
 			// This key kind doesn't contribute to the memtable size.
 			continue
+		case InternalKeyKindExcise:
+			if b.minimumFormatMajorVersion < FormatFlushableIngestExcises {
+				b.minimumFormatMajorVersion = FormatFlushableIngestExcises
+			}
+			// This key kind doesn't contribute to the memtable size.
+			continue
 		default:
 			// Note In some circumstances this might be temporary memory
 			// corruption that can be recovered by discarding the batch and
@@ -608,7 +614,7 @@ func (b *Batch) Apply(batch *Batch, _ *WriteOptions) error {
 				b.countRangeDels++
 			case InternalKeyKindRangeKeySet, InternalKeyKindRangeKeyUnset, InternalKeyKindRangeKeyDelete:
 				b.countRangeKeys++
-			case InternalKeyKindIngestSST:
+			case InternalKeyKindIngestSST, InternalKeyKindExcise:
 				panic("pebble: invalid key kind for batch")
 			case InternalKeyKindLogData:
 				// LogData does not contribute to memtable size.
@@ -1191,6 +1197,28 @@ func (b *Batch) ingestSST(fileNum base.FileNum) {
 	// number of sstable paths which have been added to the batch.
 	b.memTableSize = origMemTableSize
 	b.minimumFormatMajorVersion = FormatFlushableIngest
+}
+
+// Excise adds the excise span for a flushable ingest containing an excise. The data
+// will only be written to the WAL (not added to memtables or sstables).
+func (b *Batch) excise(start, end []byte) {
+	if b.Empty() {
+		b.ingestedSSTBatch = true
+	} else if !b.ingestedSSTBatch {
+		// Batch contains other key kinds.
+		panic("pebble: invalid call to excise")
+	}
+
+	origMemTableSize := b.memTableSize
+	b.prepareDeferredKeyValueRecord(len(start), len(end), InternalKeyKindExcise)
+	copy(b.deferredOp.Key, start)
+	copy(b.deferredOp.Value, end)
+	// Since excise writes only to the WAL and does not affect the memtable,
+	// we restore b.memTableSize to its original value. Note that Batch.count
+	// is not reset because for the InternalKeyKindIngestSST/Excise the count
+	// is the number of sstable paths which have been added to the batch.
+	b.memTableSize = origMemTableSize
+	b.minimumFormatMajorVersion = FormatFlushableIngestExcises
 }
 
 // Empty returns true if the batch is empty, and false otherwise.
--- a/batchrepr/reader.go
+++ b/batchrepr/reader.go
@@ -97,7 +97,7 @@ func (r *Reader) Next() (kind base.InternalKeyKind, ukey []byte, value []byte, o
 	switch kind {
 	case base.InternalKeyKindSet, base.InternalKeyKindMerge, base.InternalKeyKindRangeDelete,
 		base.InternalKeyKindRangeKeySet, base.InternalKeyKindRangeKeyUnset, base.InternalKeyKindRangeKeyDelete,
-		base.InternalKeyKindDeleteSized:
+		base.InternalKeyKindDeleteSized, base.InternalKeyKindExcise:
 		*r, value, ok = DecodeStr(*r)
 		if !ok {
 			return 0, nil, nil, false, errors.Wrapf(ErrInvalidBatch, "decoding %s value", kind)
--- a/data_test.go
+++ b/data_test.go
@@ -1331,7 +1331,6 @@ func (d *DB) waitTableStats() {
 
 func runIngestAndExciseCmd(td *datadriven.TestData, d *DB, fs vfs.FS) error {
 	var exciseSpan KeyRange
-	var sstContainsExciseTombstone bool
 	paths := make([]string, 0, len(td.CmdArgs))
 	for i, arg := range td.CmdArgs {
 		switch td.CmdArgs[i].Key {
@@ -1345,14 +1344,14 @@ func runIngestAndExciseCmd(td *datadriven.TestData, d *DB, fs vfs.FS) error {
 			}
 			exciseSpan.Start = []byte(fields[0])
 			exciseSpan.End = []byte(fields[1])
-		case "contains-excise-tombstone":
-			sstContainsExciseTombstone = true
+		case "no-wait":
+			// Handled by callers.
 		default:
 			paths = append(paths, arg.String())
 		}
 	}
 
-	if _, err := d.IngestAndExcise(context.Background(), paths, nil /* shared */, nil /* external */, exciseSpan, sstContainsExciseTombstone); err != nil {
+	if _, err := d.IngestAndExcise(context.Background(), paths, nil /* shared */, nil /* external */, exciseSpan); err != nil {
 		return err
 	}
 	return nil
@@ -1361,6 +1360,10 @@ func runIngestAndExciseCmd(td *datadriven.TestData, d *DB, fs vfs.FS) error {
 func runIngestCmd(td *datadriven.TestData, d *DB, fs vfs.FS) error {
 	paths := make([]string, 0, len(td.CmdArgs))
 	for _, arg := range td.CmdArgs {
+		if arg.Key == "no-wait" {
+			// Handled by callers.
+			continue
+		}
 		paths = append(paths, arg.String())
 	}
 
--- a/flushable.go
+++ b/flushable.go
@@ -163,7 +163,8 @@ type ingestedFlushable struct {
 	hasRangeKeys bool
 	// exciseSpan is populated if an excise operation should be performed during
 	// flush.
-	exciseSpan KeyRange
+	exciseSpan   KeyRange
+	exciseSeqNum base.SeqNum
 }
 
 func newIngestedFlushable(
@@ -172,6 +173,7 @@ func newIngestedFlushable(
 	newIters tableNewIters,
 	newRangeKeyIters keyspanimpl.TableNewSpanIter,
 	exciseSpan KeyRange,
+	seqNum base.SeqNum,
 ) *ingestedFlushable {
 	if invariants.Enabled {
 		for i := 1; i < len(files); i++ {
@@ -200,6 +202,7 @@ func newIngestedFlushable(
 		slice:        manifest.NewLevelSliceKeySorted(comparer.Compare, files),
 		hasRangeKeys: hasRangeKeys,
 		exciseSpan:   exciseSpan,
+		exciseSeqNum: seqNum,
 	}
 
 	return ret
@@ -245,30 +248,69 @@ func (s *ingestedFlushable) constructRangeDelIter(
 // TODO(sumeer): *IterOptions are being ignored, so the index block load for
 // the point iterator in constructRangeDeIter is not tracked.
 func (s *ingestedFlushable) newRangeDelIter(_ *IterOptions) keyspan.FragmentIterator {
-	return keyspanimpl.NewLevelIter(
+	liter := keyspanimpl.NewLevelIter(
 		context.TODO(),
 		keyspan.SpanIterOptions{}, s.comparer.Compare,
 		s.constructRangeDelIter, s.slice.Iter(), manifest.FlushableIngestsLayer(),
 		manifest.KeyTypePoint,
 	)
+	if !s.exciseSpan.Valid() {
+		return liter
+	}
+	// We have an excise span to weave into the rangedel iterators.
+	//
+	// TODO(bilal): should this be pooled?
+	miter := &keyspanimpl.MergingIter{}
+	rdel := keyspan.Span{
+		Start: s.exciseSpan.Start,
+		End:   s.exciseSpan.End,
+		Keys:  []keyspan.Key{{Trailer: base.MakeTrailer(s.exciseSeqNum, base.InternalKeyKindRangeDelete)}},
+	}
+	rdelIter := keyspan.NewIter(s.comparer.Compare, []keyspan.Span{rdel})
+	miter.Init(s.comparer, keyspan.NoopTransform, new(keyspanimpl.MergingBuffers), liter, rdelIter)
+	return miter
 }
 
 // newRangeKeyIter is part of the flushable interface.
 func (s *ingestedFlushable) newRangeKeyIter(o *IterOptions) keyspan.FragmentIterator {
-	if !s.containsRangeKeys() {
-		return nil
+	var rkeydelIter keyspan.FragmentIterator
+	if s.exciseSpan.Valid() {
+		// We have an excise span to weave into the rangekey iterators.
+		rkeydel := keyspan.Span{
+			Start: s.exciseSpan.Start,
+			End:   s.exciseSpan.End,
+			Keys:  []keyspan.Key{{Trailer: base.MakeTrailer(s.exciseSeqNum, base.InternalKeyKindRangeKeyDelete)}},
+		}
+		rkeydelIter = keyspan.NewIter(s.comparer.Compare, []keyspan.Span{rkeydel})
 	}
 
-	return keyspanimpl.NewLevelIter(
+	if !s.hasRangeKeys {
+		if rkeydelIter == nil {
+			// NB: we have to return the nil literal as opposed to the nil
+			// value of rkeydelIter, otherwise callers of this function will
+			// have the return value fail == nil checks.
+			return nil
+		}
+		return rkeydelIter
+	}
+
+	liter := keyspanimpl.NewLevelIter(
 		context.TODO(),
 		keyspan.SpanIterOptions{}, s.comparer.Compare, s.newRangeKeyIters,
 		s.slice.Iter(), manifest.FlushableIngestsLayer(), manifest.KeyTypeRange,
 	)
+	if rkeydelIter == nil {
+		return liter
+	}
+	// TODO(bilal): should this be pooled?
+	miter := &keyspanimpl.MergingIter{}
+	miter.Init(s.comparer, keyspan.NoopTransform, new(keyspanimpl.MergingBuffers), liter, rkeydelIter)
+	return miter
 }
 
 // containsRangeKeys is part of the flushable interface.
 func (s *ingestedFlushable) containsRangeKeys() bool {
-	return s.hasRangeKeys
+	return s.hasRangeKeys || s.exciseSpan.Valid()
 }
 
 // inuseBytes is part of the flushable interface.
@@ -328,6 +370,10 @@ func (s *ingestedFlushable) anyFileOverlaps(bounds base.UserKeyBounds) bool {
 		// There is overlap. Note that UserKeyBounds.Overlaps() performs exactly the
 		// checks above.
 		return true
+	}
+	if s.exciseSpan.Valid() {
+		uk := s.exciseSpan.UserKeyBounds()
+		return uk.Overlaps(s.comparer.Compare, &bounds)
 	}
 	return false
 }
--- a/flushable_test.go
+++ b/flushable_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/cockroachdb/datadriven"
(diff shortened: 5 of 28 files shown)


itsbilal:
The earlier meta test issue is fixed now, so this should be RFAL.

RaduBerinde (commented):
:lgtm:  I think Jackson should take a look at the `replayWAL` changes.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3897#-:-O5yxHm6Et_gob1VACwb:bhfp4vj)* status: 0 of 28 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi and @itsbilal)

___
*[`ingest.go` line 1289 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O5yubnmBSFNebMKc3uc:-O5yubnmBSFNebMKc3ud:b-qu8eju) ([raw file](https://github.com/cockroachdb/pebble/blob/59ff3ebdc8ae9b4342a6a501b927eff86de7a709/ingest.go#L1289)):*
> ```Go
> 	// The excise span is going to outlive this ingestion call. Copy it.
> 	exciseSpan = KeyRange{
> 		Start: append([]byte(nil), exciseSpan.Start...),
> ```

[nit] `slices.Clone`
___
*[`open.go` line 874 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O5yw1oF3DW0JYF1d_zG:-O5yw1oF3DW0JYF1d_zH:b-bslnb5) ([raw file](https://github.com/cockroachdb/pebble/blob/59ff3ebdc8ae9b4342a6a501b927eff86de7a709/open.go#L874)):*
> ```Go
> 			return errors.Wrapf(err, "running compaction during WAL replay")
> 		}
> 		ve.NewFiles = append(ve.NewFiles, newVE.NewFiles...)
> ```

[nit] maybe this should be a `VersionEdit.Merge` method?
___
*[`open.go` line 883 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O5yvpS-408vi3hNWn5Q:-O5yvpS0CYkgYHR9JTZu:bt2ai3m) ([raw file](https://github.com/cockroachdb/pebble/blob/59ff3ebdc8ae9b4342a6a501b927eff86de7a709/open.go#L883)):*
> ```Go
> 			}
> 		}
> 		if newVE.CreatedBackingTables != nil {
> ```

[nit] no need for the nil check
___
*[`open.go` line 1007 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O5ywbTP7he33wkOO2tB:-O5ywbTP7he33wkOO2tC:b-z9guyy) ([raw file](https://github.com/cockroachdb/pebble/blob/59ff3ebdc8ae9b4342a6a501b927eff86de7a709/open.go#L1007)):*
> ```Go
> 				}
> 				if kind == base.InternalKeyKindExcise {
> 					exciseSpan.Start = append([]byte(nil), encodedFileNum...)
> ```

[nit] slices.Clone
___
*[`open.go` line 1028 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O5ywfcmA7S5bDz-_Lm6:-O5ywfcmA7S5bDz-_Lm7:b-z9guyy) ([raw file](https://github.com/cockroachdb/pebble/blob/59ff3ebdc8ae9b4342a6a501b927eff86de7a709/open.go#L1028)):*
> ```Go
> 							panic("pebble: multiple excise spans in a single batch")
> 						}
> 						exciseSpan.Start = append([]byte(nil), key...)
> ```

[nit] slices.Clone

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3897#-:-O67sjuO0pft2VvtxL_C:b-814g8k)* status: 0 of 28 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi, @jbowens, and @RaduBerinde)

___
*[`ingest.go` line 1289 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O5yubnmBSFNebMKc3uc:-O67skdT3yybaU-Zc2AA:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/59ff3ebdc8ae9b4342a6a501b927eff86de7a709/ingest.go#L1289)):*

[nit] `slices.Clone`
</blockquote></details>

Done.
___
*[`open.go` line 874 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O5yw1oF3DW0JYF1d_zG:-O67slNN9TLW7wE_w1bM:b-er718v) ([raw file](https://github.com/cockroachdb/pebble/blob/59ff3ebdc8ae9b4342a6a501b927eff86de7a709/open.go#L874)):*

[nit] maybe this should be a `VersionEdit.Merge` method?
</blockquote></details>

In this case we can make some simplifying assumptions that allows us to just directly merge two VEs (that is, we'll either create brand new files, or we'll delete files that existed from the very start, we will never delete a file that was created in the same VE). In other cases we might need something like BulkVersionEdit.Accumulate. So maybe it's best to just leave it here, to avoid misuse of a generic Merge method where these assumptions don't hold.
___
*[`open.go` line 883 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O5yvpS-408vi3hNWn5Q:-O67t4MjDCdG1F3UlWSY:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/59ff3ebdc8ae9b4342a6a501b927eff86de7a709/open.go#L883)):*

[nit] no need for the nil check
</blockquote></details>

Done.

jbowens (commented):
> Reviewed 14 of 28 files at r1, 2 of 6 files at r2, 2 of 3 files at r3, 3 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3897)* status: 21 of 28 files reviewed, 6 unresolved discussions (waiting on @aadityasondhi, @itsbilal, and @RaduBerinde)

___
*[`flushable.go` line 167 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O6by3Ga0id3Txd2Wiuq:-O6by3Ga0id3Txd2Wiur:b-ypv8oe) ([raw file](https://github.com/cockroachdb/pebble/blob/c76349c3b659191943850cd6f5087a1f17f046f8/flushable.go#L167)):*
> ```Go
> 	// flush.
> 	exciseSpan KeyRange
> 	seqNum     base.SeqNum
> ```

nit: should this be called `exciseSeqNum`? or add a comment above `seqNum` discussing the sequence number assignment (with files adopting the `len(files)` sequence numbers starting at `seqNum+1` if `exciseSpan.Valid()`)
___
*[`flushable.go` line 305 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O6byw1D4to-eztnZ-tQ:-O6byw1D4to-eztnZ-tR:b-xm7zjw) ([raw file](https://github.com/cockroachdb/pebble/blob/c76349c3b659191943850cd6f5087a1f17f046f8/flushable.go#L305)):*
> ```Go
> 	)
> 	miter.Init(s.comparer, keyspan.NoopTransform, new(keyspanimpl.MergingBuffers), liter)
> 	if rkeydelIter != nil {
> ```

if `rkdelIter == nil` should we just return `liter` directly (and avoid allocating a `MergingIter`)?
___
*[`open.go` line 936 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O6c09d99gN1zOcw59CP:-O6c09d99gN1zOcw59CQ:bckb21t) ([raw file](https://github.com/cockroachdb/pebble/blob/c76349c3b659191943850cd6f5087a1f17f046f8/open.go#L936)):*
> ```Go
> 			if kind, encodedFileNum, val, ok, err := br.Next(); err != nil {
> 				return nil, 0, err
> 			} else if ok && (kind == InternalKeyKindIngestSST || kind == InternalKeyKindExcise) {
> ```

can we pull this out into its own function? `replayWAL` is already a bit hard to follow
___
*[`open.go` line 940 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O6c0T-95NuQUiRRwT71:-O6c0T-95NuQUiRRwT72:bvgqmnf) ([raw file](https://github.com/cockroachdb/pebble/blob/c76349c3b659191943850cd6f5087a1f17f046f8/open.go#L940)):*
> ```Go
> 				//
> 				// Ingests require an up-to-date view of the LSM to determine the target
> 				// level of ingested sstables, and to accurately compute excises. Outside
> ```

would it simplify things to just flushable ingest these into L0 during recovery? I don't think there's much of a loss in doing that and would love to keep the recovery code as simple as we can.
___
*[`internal/base/internal.go` line 163 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O6c0lzA331qPbxEtnNA:-O6c0lzA331qPbxEtnNB:b2xbj1h) ([raw file](https://github.com/cockroachdb/pebble/blob/c76349c3b659191943850cd6f5087a1f17f046f8/internal/base/internal.go#L163)):*
> ```Go
> 	// if the current InternalKeyKindMax is a kind that is never added to an
> 	// sstable or memtable (eg. InternalKeyKindExcise).
> 	InternalKeyKindMaxForSstable InternalKeyKind = InternalKeyKindDeleteSized
> ```

nit: I think we tend to consistently use `SSTable` capitalization

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3897#-:-O6h-k8R9GcK96CKOcKx:b-814g8k)* status: 21 of 28 files reviewed, 4 unresolved discussions (waiting on @aadityasondhi, @jbowens, and @RaduBerinde)

___
*[`flushable.go` line 167 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O6by3Ga0id3Txd2Wiuq:-O6gpW4VCY0lAnC3q551:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c76349c3b659191943850cd6f5087a1f17f046f8/flushable.go#L167)):*

nit: should this be called `exciseSeqNum`? or add a comment above `seqNum` discussing the sequence number assignment (with files adopting the `len(files)` sequence numbers starting at `seqNum+1` if `exciseSpan.Valid()`)
</blockquote></details>

Done.
___
*[`flushable.go` line 305 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O6byw1D4to-eztnZ-tQ:-O6gplDw-IZfg0mzn8ag:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c76349c3b659191943850cd6f5087a1f17f046f8/flushable.go#L305)):*

if `rkdelIter == nil` should we just return `liter` directly (and avoid allocating a `MergingIter`)?
</blockquote></details>

Done.
___
*[`open.go` line 936 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O6c09d99gN1zOcw59CP:-O6h-dFyCgqceKrO-n-A:bsblcvp) ([raw file](https://github.com/cockroachdb/pebble/blob/c76349c3b659191943850cd6f5087a1f17f046f8/open.go#L936)):*

can we pull this out into its own function? `replayWAL` is already a bit hard to follow
</blockquote></details>

Done. I kept the versionEdit/toFlush interactions here but moved the flushableEntry construction and batch reading into a function and it's much cleaner now.
___
*[`open.go` line 940 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O6c0T-95NuQUiRRwT71:-O6gqi7nA87G6SV9801s:b-twgqp0) ([raw file](https://github.com/cockroachdb/pebble/blob/c76349c3b659191943850cd6f5087a1f17f046f8/open.go#L940)):*

would it simplify things to just flushable ingest these into L0 during recovery? I don't think there's much of a loss in doing that and would love to keep the recovery code as simple as we can.
</blockquote></details>

The issue isn't so much just the L0 ingestions, but also the fact that we need to excise files we've written up to this point. I think it's easier to reason about correctness when the recovery code shares more code with the regular ingest pipeline than to have it do something unique and special.
___
*[`internal/base/internal.go` line 163 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O6c0lzA331qPbxEtnNA:-O6gqzoy1hRge5NkWPEy:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/c76349c3b659191943850cd6f5087a1f17f046f8/internal/base/internal.go#L163)):*

nit: I think we tend to consistently use `SSTable` capitalization
</blockquote></details>

Done.

jbowens (commented):
> Reviewed 3 of 28 files at r1, 1 of 5 files at r4, 3 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3897)* status: 26 of 28 files reviewed, 7 unresolved discussions (waiting on @aadityasondhi, @itsbilal, and @RaduBerinde)

___
*[`open.go` line 940 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O6c0T-95NuQUiRRwT71:-O762v7OE_OA29I2-x4p:bq8rngi) ([raw file](https://github.com/cockroachdb/pebble/blob/c76349c3b659191943850cd6f5087a1f17f046f8/open.go#L940)):*

The issue isn't so much just the L0 ingestions, but also the fact that we need to excise files we've written up to this point. I think it's easier to reason about correctness when the recovery code shares more code with the regular ingest pipeline than to have it do something unique and special.
</blockquote></details>

Thanks, I see. I still think this code structure is getting too unwieldy, and we're doing something unique and special right here. IIUC, before this PR `replayWAL` performed no calls to `logAndApply`. Its responsibility was a) building up the `flushables` list and b) corresponding set of version edit mutations. Then the caller was responsible for either just going on its way with the `flushables` as-is (read-only mode) or committing the version edit (read-write mode). (Are we respecting read-only mode here?)
___
*[`open.go` line 792 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O75uiAy-kNw96wjnkA9:-O75uiAy-kNw96wjnkAA:b5jpo41) ([raw file](https://github.com/cockroachdb/pebble/blob/0775601c95fbd85b87591f9a86a94796bd4dab80/open.go#L792)):*
> ```Go
> 	}
> 	if kind == base.InternalKeyKindExcise {
> 		exciseSpan.Start = slices.Clone(encodedFileNum)
> ```

rename `encodedFileNum`? it's not an encoded file num here.
___
*[`open.go` line 795 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O75v3aaBFZ9_V_3kG7i:-O75v3aaBFZ9_V_3kG7j:bbgdttx) ([raw file](https://github.com/cockroachdb/pebble/blob/0775601c95fbd85b87591f9a86a94796bd4dab80/open.go#L795)):*
> ```Go
> 		exciseSpan.End = slices.Clone(val)
> 	} else {
> 		addFileNum(encodedFileNum)
> ```

I didn't follow why the first iteration is handled up outside the loop?
___
*[`open.go` line 823 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O75wLXC02WMfYwM1_3C:-O75wLXC02WMfYwM1_3D:bn60yno) ([raw file](https://github.com/cockroachdb/pebble/blob/0775601c95fbd85b87591f9a86a94796bd4dab80/open.go#L823)):*
> ```Go
> 		return nil, err
> 	} else if ok {
> 		panic("pebble: invalid number of entries in batch.")
> ```

nit: stray trailing period
___
*[`open.go` line 950 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O75wpIC4vI5FxOTpLgB:-O75wpIC4vI5FxOTpLgC:b-5plwq2) ([raw file](https://github.com/cockroachdb/pebble/blob/0775601c95fbd85b87591f9a86a94796bd4dab80/open.go#L950)):*
> ```Go
> 	// updateVE is used to update ve with information about new files created
> 	// during the flush of any flushable not of type ingestedFlushable. For the
> 	// flushable of type ingestedFlushable we use custom handling below.
> ```

this comment is outdated now
___
*[`open.go` line 1047 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O75yKXCAGcBeKuQCb9P:-O75yKXCAGcBeKuQCb9Q:b-xb09eh) ([raw file](https://github.com/cockroachdb/pebble/blob/0775601c95fbd85b87591f9a86a94796bd4dab80/open.go#L1047)):*
> ```Go
> 					entry = d.mu.mem.queue[len(d.mu.mem.queue)-1]
> 					d.mu.mem.mutable = nil
> 					d.mu.mem.queue = d.mu.mem.queue[:len(d.mu.mem.queue)-1]
> ```

this line is confusing. the mutable memtable is the last item in the queue. when we flush, we'll flush the entirety of the queue. why is this trimming off just the mutable memtable at the end? are we guaranteed that the queue is a single element, the mutable memtable? if so we should add a comment describing the invariant, enforce it and use `d.mu.mem.queue[:0]` here 
___
*[`open.go` line 1061 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O75zqhxBUxfSEj-aykr:-O75zqhxBUxfSEj-ayks:b-7b0tfa) ([raw file](https://github.com/cockroachdb/pebble/blob/0775601c95fbd85b87591f9a86a94796bd4dab80/open.go#L1061)):*
> ```Go
> 				if d.mu.versions.logSeqNum.Load() < maxSeqNum {
> 					d.mu.versions.logSeqNum.Store(maxSeqNum)
> 				}
> ```

why do we need to update the logSeqNum/visibleSeqNum early instead of waiting for the caller of `replayWAL` to do it?

jbowens (commented):
> I still think this code structure is getting too unwieldy

In thinking about this more, why do we need to actually perform the excise immediately in in replayWAL? The goal of ingested flushables is to defer the operation until flush time. Shouldn't we able to queue the appropriate flushables during `replayWAL` and then perform flush operations to drain the queue of flushables? I.e, it seems like we uncouple the replaying into memtables/flushables from the construction of the version edits. We'll still be able to use the existing ingestion excise logic to perform the excise. If we go this route, we should first refactor the existing code to have the structure we want before we attempt to extend it to support excises with flushable ingestions.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3897#-:-O79oNVvBnRiWzNAvx2C:beq7uv)* status: 26 of 28 files reviewed, 7 unresolved discussions (waiting on @aadityasondhi, @itsbilal, and @RaduBerinde)

itsbilal (commented):
TFTRs!

I agree that the cleaner approach is to just always replay WALs into the flushable queue, and then do flushes until the flushable queue is empty. That way we can maintain one code path for flushes and flushable ingests, and makes this replay code much easier to reason about. I've gone ahead and done that refactor.

There's still one "memtable buffer not freed" bug in one test (more like, when one test is run with other tests and not on its own) that I spent a lot of time taking a look at, but wasn't able to quite get to the bottom of yet. Other than that, this is ready for a look.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3897#-:-O7G7YFb3QxjX14qFYjO:b-3te50l)* status: 22 of 39 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi, @jbowens, and @RaduBerinde)

___
*[`open.go` line 792 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O75uiAy-kNw96wjnkA9:-O7G7rdx1o5_qwpogLsK:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0775601c95fbd85b87591f9a86a94796bd4dab80/open.go#L792)):*

rename `encodedFileNum`? it's not an encoded file num here.
</blockquote></details>

Done.
___
*[`open.go` line 823 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O75wLXC02WMfYwM1_3C:-O7G7sOW02ZTUaT4Ngdx:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0775601c95fbd85b87591f9a86a94796bd4dab80/open.go#L823)):*

nit: stray trailing period
</blockquote></details>

Done.
___
*[`open.go` line 950 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O75wpIC4vI5FxOTpLgB:-O7G7tAf4XYHwiwDzQbd:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0775601c95fbd85b87591f9a86a94796bd4dab80/open.go#L950)):*

this comment is outdated now
</blockquote></details>

Done.
___
*[`open.go` line 1047 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O75yKXCAGcBeKuQCb9P:-O7G7v8C1gUJzGfPY7Yx:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0775601c95fbd85b87591f9a86a94796bd4dab80/open.go#L1047)):*

this line is confusing. the mutable memtable is the last item in the queue. when we flush, we'll flush the entirety of the queue. why is this trimming off just the mutable memtable at the end? are we guaranteed that the queue is a single element, the mutable memtable? if so we should add a comment describing the invariant, enforce it and use `d.mu.mem.queue[:0]` here
</blockquote></details>

Done.
___
*[`open.go` line 1061 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3897#-O75zqhxBUxfSEj-aykr:-O7G7wSgCzvQT1wvQU_-:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0775601c95fbd85b87591f9a86a94796bd4dab80/open.go#L1061)):*

why do we need to update the logSeqNum/visibleSeqNum early instead of waiting for the caller of `replayWAL` to do it?
</blockquote></details>

Done.

itsbilal (commented):
> There's still one "memtable buffer not freed" bug in one test (more like, when one test is run with other tests and not on its own) that I spent a lot of time taking a look at, but wasn't able to quite get to the bottom of yet.

This is addressed now.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3897#-:-O7GDPRP8QKFjlNFsFRN:b-8saeas)* status: 22 of 39 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi, @jbowens, and @RaduBerinde)

jbowens (commented):
> If we go this route, we should first refactor the existing code to have the structure we want before we attempt to extend it to support excises with flushable ingestions.

Do you mind pulling out the refactor into a separate PR?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3897#-:-O7UYMcP-i_qkA5OSOaW:b-s5m7bm)* status: 22 of 39 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi, @itsbilal, and @RaduBerinde)

itsbilal:
> Do you mind pulling out the refactor into a separate PR?

Done: https://github.com/cockroachdb/pebble/pull/3955

itsbilal:
TFTR!

PR #3917, opened 2024-09-04
manual: add cgo memory accounting

### LINKED ISSUE #3908: heap profiler for manual allocation
We have a few uses of `manual.New`. It would be good when investigating memory issues to get an idea of the breakdown for each usage.

We would have a profiling mode that is off by default but can be turned on for a specific node. When it is on, `manual.New` randomly samples allocations and maintains estimates of the usage. We can use the same method Go uses: it samples one out of 512KB allocated bytes (on average). It does this by keeping a counter until the next sampling. Whenever the counter reaches 0, we generate a new value using an exponential distribution with the desired mean.

Jira issue: PEBBLE-255

RaduBerinde:
This commit adds a breakdown of cgo memory usage in Pebble. This will
show up as part of the periodic LSM metrics dump in the logs.

Fixes #3908

### DIFF
--- a/db.go
+++ b/db.go
@@ -2041,6 +2041,8 @@ func (d *DB) Metrics() *Metrics {
 
 	metrics.Uptime = d.timeNow().Sub(d.openedAt)
 
+	metrics.manualMemory = manual.GetMetrics()
+
 	return metrics
 }
 
@@ -2368,7 +2370,7 @@ func (d *DB) newMemTable(
 		memtblOpts.releaseAccountingReservation = mem.releaseAccountingReservation
 	} else {
 		mem = new(memTable)
-		memtblOpts.arenaBuf = manual.New(int(size))
+		memtblOpts.arenaBuf = manual.New(manual.MemTable, int(size))
 		memtblOpts.releaseAccountingReservation = d.opts.Cache.Reserve(int(size))
 		d.memTableCount.Add(1)
 		d.memTableReserved.Add(int64(size))
--- a/internal/cache/block_map.go
+++ b/internal/cache/block_map.go
@@ -27,14 +27,14 @@ type blockMapAllocator struct{}
 
 func (blockMapAllocator) Alloc(n int) []swiss.Group[key, *entry] {
 	size := uintptr(n) * unsafe.Sizeof(swiss.Group[key, *entry]{})
-	buf := manual.New(int(size))
+	buf := manual.New(manual.BlockCacheMap, int(size))
 	return unsafe.Slice((*swiss.Group[key, *entry])(unsafe.Pointer(unsafe.SliceData(buf))), n)
 }
 
 func (blockMapAllocator) Free(v []swiss.Group[key, *entry]) {
 	size := uintptr(len(v)) * unsafe.Sizeof(swiss.Group[key, *entry]{})
 	buf := unsafe.Slice((*byte)(unsafe.Pointer(unsafe.SliceData(v))), size)
-	manual.Free(buf)
+	manual.Free(manual.BlockCacheMap, buf)
 }
 
 var blockMapOptions = []swiss.Option[key, *entry]{
--- a/internal/cache/entry_normal.go
+++ b/internal/cache/entry_normal.go
@@ -80,7 +80,7 @@ func (c *entryAllocCache) alloc() *entry {
 		if entriesGoAllocated {
 			return &entry{}
 		}
-		b := manual.New(entrySize)
+		b := manual.New(manual.BlockCacheEntry, entrySize)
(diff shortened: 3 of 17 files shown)


RaduBerinde:
What Pebble-level benchmark do you recommend running to check the overhead? Something that churns the block cache would be best.

jbowens (approved):
:lgtm_strong: 

> Reviewed 11 of 15 files at r1, 3 of 3 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3917#-:-O62UevTE21vG-zeE5yL:b-uwkowd)* status: all files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`internal/manual/manual.go` line 30 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3917#-O62Su9L6aqDdChFhNR1:-O62Su9L6aqDdChFhNR2:b-7ae29p) ([raw file](https://github.com/cockroachdb/pebble/blob/86535183021c55fa9784f6d01f6070265de05fb1/internal/manual/manual.go#L30)):*
> ```Go
> 	InUseBytes uint64
> 
> 	// InUseBytes is the total cumulative number of bytes allocated since the
> ```

s/InUseBytes/TotalBytes/
___
*[`internal/manual/manual.go` line 50 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3917#-O62TMY2C-ktgq4gM3tE:-O62TMY2C-ktgq4gM3tF:b1unkat) ([raw file](https://github.com/cockroachdb/pebble/blob/86535183021c55fa9784f6d01f6070265de05fb1/internal/manual/manual.go#L50)):*
> ```Go
> 	for i := range res {
> 		res[i].TotalBytes = counters[i].TotalAllocated.Load()
> 		res[i].InUseBytes = res[i].TotalBytes - counters[i].TotalFreed.Load()
> ```

Since these loads are not together atomic, it's possible for total to exceed freed. Should we load freed first, so that any drift is considered 'in-use' and we don't underflow?

RaduBerinde:
*[`internal/manual/manual.go` line 50 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3917#-O62TMY2C-ktgq4gM3tE:-O62_CCW2Rbu7eq3V0SL:byh9trs) ([raw file](https://github.com/cockroachdb/pebble/blob/86535183021c55fa9784f6d01f6070265de05fb1/internal/manual/manual.go#L50)):*

Since these loads are not together atomic, it's possible for total to exceed freed. Should we load freed first, so that any drift is considered 'in-use' and we don't underflow?
</blockquote></details>

Good point..  Should I just switch to a single atomic that we increase/decrease? I doubt the total-since-process-start would ever be very helpful.

RaduBerinde (commented):
TFTR!

I ran a few variations of ycs-b (with different cache sizes and concurrencies) and didn't see any difference before and after. I think this makes sense since most allocations are big and the small ones for the cache entries are buffered into a pool (plus each one is associated with a cache miss which has a lot more associated work than a counter).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3917#-:-O62bJLyCEdlyoMsxUS4:b-uf9rl3)* status: 13 of 17 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`internal/manual/manual.go` line 30 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3917#-O62Su9L6aqDdChFhNR1:-O62auEaBw1LRVp6WZtM:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/86535183021c55fa9784f6d01f6070265de05fb1/internal/manual/manual.go#L30)):*

s/InUseBytes/TotalBytes/
</blockquote></details>

Done.
___
*[`internal/manual/manual.go` line 50 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3917#-O62TMY2C-ktgq4gM3tE:-O62bH3D0VdlphGUO3A7:b-5yq8du) ([raw file](https://github.com/cockroachdb/pebble/blob/86535183021c55fa9784f6d01f6070265de05fb1/internal/manual/manual.go#L50)):*

Good point..  Should I just switch to a single atomic that we increase/decrease? I doubt the total-since-process-start would ever be very helpful.
</blockquote></details>

I went ahead and did that.

jbowens (approved):
:lgtm:

> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3917#-:-O62sjGH6zdVPaFkf_eb:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

PR #32, opened 2019-01-04
sstable: remove support for block-based bloom filters

tbg:
Block-based bloom filters are commonly eschewed in favor of table-based
bloom filters. Continuing to support block-based bloom filters while
remaining bit-by-bit compatible to RocksDB's SSTs will be challenging due
to RocksDB's crufty block-based bloom filter code which notably diverges
from the table-based bloom filter code.

See https://github.com/facebook/rocksdb/wiki/RocksDB-Bloom-Filter for an
explanation of the two bloom filter types.

PR #3920, opened 2024-09-05
sstable: remove unnecessary loop while identifying filter block

jbowens:
When opening an sstable and reading the metaindex, we loop over the known filters looking for a block with a matching name. Previously there was an inner loop that statically only ever had one iteration. This inner loop was a vestige from early Pebble's support for 'block-based' bloom filters. There's only ever been a single filter type since #32.

### DIFF
--- a/sstable/reader.go
+++ b/sstable/reader.go
@@ -559,29 +559,9 @@ func (r *Reader) readMetaindex(
 	}
 
 	for name, fp := range filters {
-		types := []struct {
-			ftype  FilterType
-			prefix string
-		}{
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3920#-:-O62e_of8uyhJ1MC2xKQ:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @itsbilal)

jbowens:
TFTR!

PR #3926, opened 2024-09-06
db: use delete-only compaction hints to do excises

### LINKED ISSUE #3820: db: add ability to excise without ingest
Currently, the only way to do an Excise is to do an `IngestAndExcise`. If we added the ability to do excises without an accompanying ingest, likely a `db.Excise(exciseSpans []KeyRange)` method where all the passed-in spans are excised in one atomic VersionEdit, we can replace more uses of rangedels / rangekeydels in CockroachDB with excises and significantly reduce write-amp.

Ideally this would happen after #2676 is complete. This change also makes it easier to complete https://github.com/cockroachdb/cockroach/issues/128004 .

Jira issue: PEBBLE-231

Epic CRDB-37583

itsbilal:
Previously, we'd only act on delete-only compaction hints if the entirety of a file was covered by a delete-only compaction hint. This change leverages the fact that we can now partially excise files using virtual sstables, to also use delete-only compactions to do excises of partial sstables and reduce the w-amp impact of writing rangedels and rangekeydels.

Fixes #3820.

### DIFF
--- a/testdata/compaction_delete_only_hints
+++ b/testdata/compaction_delete_only_hints
@@ -127,7 +127,7 @@ maybe-compact
 Deletion hints:
   (none)
 Compactions:
-  [JOB 100] compacted(delete-only) L2 [000006] (605B) Score=0.00 + L3 [000007] (605B) Score=0.00 -> L6 [] (0B), in 1.0s (2.0s total), output rate 0B/s
+  [JOB 100] compacted(delete-only) multilevel L1 [000005] (650B) Score=0.00 + L2 [000006] (605B) Score=0.00 + L3 [000007] (605B) Score=0.00 -> L6 [000009] (1B), in 1.0s (2.0s total), output rate 1B/s
 
 # Test a range tombstone that is already compacted into L6.
 
@@ -431,3 +431,223 @@ Deletion hints:
   (none)
 Compactions:
   [JOB 100] compacted(delete-only) L6 [000006 000007 000008 000009 000011] (3.3KB) Score=0.00 -> L6 [] (0B), in 1.0s (2.0s total), output rate 0B/s
+
+# Verify that a delete-only compaction can partially excise a file.
+
+define
+L0
+a.RANGEDEL.300:k
+L1
+b.RANGEDEL.230:h h.RANGEDEL.200:r
+L2
+d.SET.110:d i.SET.140:i
+L3
+k.SET.90:k o.SET.150:o
+L4
+m.SET.30:m u.SET.60:u
+----
+L0.0:
+  000004:[a#300,RANGEDEL-k#inf,RANGEDEL]
+L1:
+  000005:[b#230,RANGEDEL-r#inf,RANGEDEL]
+L2:
+  000006:[d#110,SET-i#140,SET]
+L3:
+  000007:[k#90,SET-o#150,SET]
+L4:
+  000008:[m#30,SET-u#60,SET]
+
+get-hints
+----
+L0.000004 a-k seqnums(tombstone=300-300, file-smallest=110, type=point-key-only)
+L1.000005 b-r seqnums(tombstone=200-230, file-smallest=90, type=point-key-only)
+
+iter
+first
+next
+next
+next
+----
+u: (u, .)
+.
+.
+.
+
+maybe-compact
+----
+Deletion hints:
+  (none)
+Compactions:
+  [JOB 100] compacted(delete-only) multilevel L1 [000005] (650B) Score=0.00 + L2 [000006] (605B) Score=0.00 + L3 [000007] (605B) Score=0.00 -> L6 [000009] (1B), in 1.0s (2.0s total), output rate 1B/s
+
+describe-lsm
+----
+L0.0:
+  000004:[a#300,RANGEDEL-k#inf,RANGEDEL]
+L1:
+  000009(000005):[k#200,RANGEDEL-r#inf,RANGEDEL]
+L4:
+  000008:[m#30,SET-u#60,SET]
+
+iter
+first
+next
+next
+next
+----
+u: (u, .)
+.
+.
+.
+
+get-hints
+----
+(none)
+
+# Verify that a delete-only compaction hint for point keys does not excise a file
+# if it has range keys
+
+reset
+----
+
+ingest ext
+set k k
+set o o
+range-key-set k o @3 foo
+----
+OK
+
+ingest ext
+set d d
+set i i
+----
+OK
+
+ingest ext
+del-range b r
+----
+OK
+
+describe-lsm
+----
+L0.0:
+  000006:[b#12,RANGEDEL-r#inf,RANGEDEL]
+L6:
+  000005:[d#11,SET-i#11,SET]
+  000004:[k#10,RANGEKEYSET-o#10,SET]
+
+get-hints
+----
+L0.000006 b-r seqnums(tombstone=12-12, file-smallest=11, type=point-key-only)
+
+iter
+first
+next
+next
+next
+----
+.
+.
+.
+.
+
+maybe-compact
+----
+Deletion hints:
+  (none)
+Compactions:
+  [JOB 100] compacted(delete-only) L6 [000005] (597B) Score=0.00 -> L6 [] (0B), in 1.0s (2.0s total), output rate 0B/s
+
+describe-lsm
+----
+L0.0:
+  000006:[b#12,RANGEDEL-r#inf,RANGEDEL]
+L6:
+  000004:[k#10,RANGEKEYSET-o#10,SET]
+
+iter
+first
+next
+next
+next
+----
+.
+.
+.
+.
+
+get-hints
+----
+(none)
+
+# Verify that a delete-only compaction hint can excise a file twice
+
+reset
+----
+
+ingest ext
+set k k
+set m m
+set o o
+range-key-set k o @3 foo
+----
+OK
+
+ingest ext
+del-range b l
+del-range m n
+range-key-del b l
+range-key-del m n
+----
+OK
+
+describe-lsm
+----
+L0.0:
+  000005:[b#11,RANGEKEYDEL-n#inf,RANGEDEL]
+L6:
+  000004:[k#10,RANGEKEYSET-o#10,SET]
+
+get-hints
+----
+(none)
+
+iter
+first
+next
+next
+next
+----
+o: (o, .)
+.
+.
+.
+
+maybe-compact
+----
+Deletion hints:
+  (none)
+Compactions:
+  (none)
+
+describe-lsm
+----
+L0.0:
+  000005:[b#11,RANGEKEYDEL-n#inf,RANGEDEL]
+L6:
+  000004:[k#10,RANGEKEYSET-o#10,SET]
+
+iter
+first
+next
+next
+next
+----
+o: (o, .)
+.
+.
+.
+
+get-hints
+----
+(none)
--- a/compaction.go
+++ b/compaction.go
@@ -11,6 +11,7 @@ import (
 	"math"
 	"runtime/pprof"
 	"slices"
+	"sort"
 	"sync/atomic"
 	"time"
 
@@ -251,6 +252,11 @@ type compaction struct {
 	// lower level in the LSM during runCompaction.
 	allowedZeroSeqNum bool
 
+	// deletionHints are set if this is a compactionKindDeleteOnly. Used to figure
+	// out whether an input must be deleted in its entirety, or excised into
+	// virtual sstables.
+	deletionHints []deleteCompactionHint
+
 	metrics map[int]*LevelMetrics
 
 	pickerMetrics compactionPickerMetrics
@@ -390,18 +396,23 @@ func newCompaction(
 }
 
 func newDeleteOnlyCompaction(
-	opts *Options, cur *version, inputs []compactionLevel, beganAt time.Time,
+	opts *Options,
+	cur *version,
+	inputs []compactionLevel,
+	beganAt time.Time,
+	hints []deleteCompactionHint,
 ) *compaction {
 	c := &compaction{
-		kind:      compactionKindDeleteOnly,
-		cmp:       opts.Comparer.Compare,
-		equal:     opts.Comparer.Equal,
-		comparer:  opts.Comparer,
-		formatKey: opts.Comparer.FormatKey,
-		logger:    opts.Logger,
-		version:   cur,
-		beganAt:   beganAt,
-		inputs:    inputs,
+		kind:          compactionKindDeleteOnly,
+		cmp:           opts.Comparer.Compare,
+		equal:         opts.Comparer.Equal,
+		comparer:      opts.Comparer,
+		formatKey:     opts.Comparer.FormatKey,
+		logger:        opts.Logger,
+		version:       cur,
+		beganAt:       beganAt,
+		inputs:        inputs,
+		deletionHints: hints,
 	}
 
 	// Set c.smallest, c.largest.
@@ -1013,10 +1024,13 @@ func (d *DB) clearCompactingState(c *compaction, rollback bool) {
 				d.opts.Logger.Fatalf("L%d->L%d: %s not being compacted", c.startLevel.level, c.outputLevel.level, f.FileNum)
 			}
 			if !rollback {
-				// On success all compactions other than move-compactions transition the
-				// file into the Compacted state. Move-compacted files become eligible
-				// for compaction again and transition back to NotCompacting.
-				if c.kind != compactionKindMove {
+				// On success all compactions other than move and delete-only compactions
+				// transition the file into the Compacted state. Move-compacted files
+				// become eligible for compaction again and transition back to NotCompacting.
+				// Delete-only compactions could, on rare occasion, leave files untouched
+				// (eg. if files have a loose bound), so we revert them all to NotCompacting
+				// just in case they need to be compacted again.
+				if c.kind != compactionKindMove && c.kind != compactionKindDeleteOnly {
 					f.SetCompactionState(manifest.CompactionStateCompacted)
 				} else {
 					f.SetCompactionState(manifest.CompactionStateNotCompacting)
@@ -1282,7 +1296,7 @@ func (d *DB) runIngestFlush(c *compaction) (*manifest.VersionEdit, error) {
 			iter := overlaps.Iter()
 
 			for m := iter.First(); m != nil; m = iter.Next() {
-				newFiles, err := d.excise(context.TODO(), ingestFlushable.exciseSpan.UserKeyBounds(), m, ve, l)
+				newFiles, err := d.excise(context.TODO(), ingestFlushable.exciseSpan.UserKeyBounds(), m, ve, l, d.mu.versions.getNextFileNum)
 				if err != nil {
 					return nil, err
 				}
@@ -1759,11 +1773,11 @@ func (d *DB) maybeScheduleCompactionPicker(
 func (d *DB) tryScheduleDeleteOnlyCompaction() {
 	v := d.mu.versions.currentVersion()
 	snapshots := d.mu.snapshots.toSlice()
-	inputs, unresolvedHints := checkDeleteCompactionHints(d.cmp, v, d.mu.compact.deletionHints, snapshots)
+	inputs, resolvedHints, unresolvedHints := checkDeleteCompactionHints(d.cmp, v, d.mu.compact.deletionHints, snapshots, d.FormatMajorVersion())
 	d.mu.compact.deletionHints = unresolvedHints
 
 	if len(inputs) > 0 {
-		c := newDeleteOnlyCompaction(d.opts, v, inputs, d.timeNow())
+		c := newDeleteOnlyCompaction(d.opts, v, inputs, d.timeNow(), resolvedHints)
 		d.mu.compact.compactingCount++
 		d.addInProgressCompaction(c)
 		go d.compact(c, nil)
@@ -1914,8 +1928,8 @@ func (h deleteCompactionHint) String() string {
 	)
 }
 
-func (h *deleteCompactionHint) canDelete(
-	cmp Compare, m *fileMetadata, snapshots compact.Snapshots,
+func (h *deleteCompactionHint) canDeleteOrExcise(
+	cmp Compare, m *fileMetadata, snapshots compact.Snapshots, fmv FormatMajorVersion,
 ) bool {
 	// The file can only be deleted if all of its keys are older than the
 	// earliest tombstone aggregated into the hint. Note that we use
@@ -1960,18 +1974,30 @@ func (h *deleteCompactionHint) canDelete(
 	default:
 		panic(fmt.Sprintf("pebble: unknown delete compaction hint type: %d", h.hintType))
 	}
-
-	// The file's keys must be completely contained within the hint range.
-	return cmp(h.start, m.Smallest.UserKey) <= 0 && cmp(m.Largest.UserKey, h.end) < 0
+	if fmv < FormatVirtualSSTables {
+		// The file's keys must be completely contained within the hint range.
+		return cmp(h.start, m.Smallest.UserKey) <= 0 &&
+			base.UserKeyExclusive(h.end).CompareUpperBounds(cmp, m.UserKeyBounds().End) >= 0
+	}
+	// Check for any overlap. In cases of partial overlap, we excise the part of the file
+	// that overlaps with the deletion hint.
+	return cmp(h.end, m.Smallest.UserKey) > 0 &&
+		(m.UserKeyBounds().End.CompareUpperBounds(cmp, base.UserKeyInclusive(h.start)) >= 0)
 }
 
 func checkDeleteCompactionHints(
-	cmp Compare, v *version, hints []deleteCompactionHint, snapshots compact.Snapshots,
-) ([]compactionLevel, []deleteCompactionHint) {
+	cmp Compare,
+	v *version,
+	hints []deleteCompactionHint,
+	snapshots compact.Snapshots,
+	fmv FormatMajorVersion,
+) (levels []compactionLevel, resolved, unresolved []deleteCompactionHint) {
 	var files map[*fileMetadata]bool
 	var byLevel [numLevels][]*fileMetadata
 
 	unresolvedHints := hints[:0]
+	// Lazily populate resolvedHints, similar to files above.
+	resolvedHints := make([]deleteCompactionHint, 0)
 	for _, h := range hints {
 		// Check each compaction hint to see if it's resolvable. Resolvable
 		// hints are removed and trigger a delete-only compaction if any files
@@ -2025,7 +2051,7 @@ func checkDeleteCompactionHints(
 			overlaps := v.Overlaps(l, base.UserKeyBoundsEndExclusive(h.start, h.end))
 			iter := overlaps.Iter()
 			for m := iter.First(); m != nil; m = iter.Next() {
-				if m.IsCompacting() || !h.canDelete(cmp, m, snapshots) || files[m] {
+				if m.IsCompacting() || !h.canDeleteOrExcise(cmp, m, snapshots, fmv) || files[m] {
 					continue
 				}
 				if files == nil {
@@ -2037,6 +2063,7 @@ func checkDeleteCompactionHints(
 				byLevel[l] = append(byLevel[l], m)
 			}
 		}
+		resolvedHints = append(resolvedHints, h)
 	}
 
 	var compactLevels []compactionLevel
@@ -2049,7 +2076,7 @@ func checkDeleteCompactionHints(
 			files: manifest.NewLevelSliceKeySorted(cmp, files),
 		})
 	}
-	return compactLevels, unresolvedHints
+	return compactLevels, resolvedHints, unresolvedHints
 }
 
 // compact runs one compaction and maybe schedules another call to compact.
@@ -2349,24 +2376,137 @@ func (d *DB) runCopyCompaction(
 	return ve, compact.Stats{}, nil
 }
 
+func (d *DB) runDeleteOnlyCompactionForLevel(
+	c *compaction,
+	cl compactionLevel,
+	levelMetrics *LevelMetrics,
+	ve *versionEdit,
+	nextFileNum func() base.FileNum,
+	snapshots compact.Snapshots,
+) error {
+	iter := cl.files.Iter()
+	// c.deletionHints[curHint] is the hint that overlaps with f. There
+	// could be more than one hint overlapping with f.
+	curHint := 0
+	f := iter.First()
+	// curFile usually matches f, except if f got excised in which case
+	// it maps to the "current state" of f.
+	curFile := f
+
+	for f != nil && curHint < len(c.deletionHints) {
+		h := c.deletionHints[curHint]
+		if h.tombstoneLevel >= cl.level {
+			// We cannot excise out the deletion tombstone itself, or anything
+			// above it.
+			curHint++
+			continue
+		}
+		if d.cmp(h.end, f.Smallest.UserKey) <= 0 || !h.canDeleteOrExcise(d.cmp, f, snapshots, d.FormatMajorVersion()) {
+			curHint++
+			continue
+		}
+		if curFile.UserKeyBounds().End.CompareUpperBounds(d.cmp, base.UserKeyInclusive(h.start)) < 0 {
+			if _, ok := ve.DeletedFiles[deletedFileEntry{
+				Level:   cl.level,
+				FileNum: f.FileNum,
+			}]; !ok && f == curFile {
+				panic("pebble: deletion only compaction scheduled with a hint that didn't delete a file")
+			}
+			f = iter.Next()
+			curFile = f
+			continue
+		}
+
+		// The hint overlaps with at least part of the file.
+		if curFile == f && d.cmp(h.start, f.Smallest.UserKey) <= 0 &&
+			base.UserKeyExclusive(h.end).CompareUpperBounds(d.cmp, f.UserKeyBounds().End) >= 0 {
+			// The hint deletes the entirety of this file.
+			ve.DeletedFiles[deletedFileEntry{
+				Level:   cl.level,
+				FileNum: f.FileNum,
+			}] = f
+			f = iter.Next()
+			curFile = f
+			levelMetrics.TablesDeleted++
+			continue
+		}
+		// The hint overlaps with only a part of the file, not the entirety of it. We need
+		// to use d.excise.
+		if d.FormatMajorVersion() < FormatVirtualSSTables {
+			panic("pebble: format major version too low to do an excise in deletion only compaction")
+		}
+		levelMetrics.TablesExcised++
+		newFiles, err := d.excise(context.TODO(), base.UserKeyBoundsEndExclusive(h.start, h.end), curFile, ve, cl.level, nextFileNum)
+		if err != nil {
+			return errors.Wrap(err, "error when running excise for delete-only compaction")
+		}
+		if _, ok := ve.DeletedFiles[deletedFileEntry{
+			Level:   cl.level,
+			FileNum: curFile.FileNum,
+		}]; !ok {
+			// This hint did not touch this file. Try the next hint.
+			curHint++
+			continue
+		}
+		if curFile != f {
+			// We excised a file after having previously excised it. Remove curFile from both
+			// deletedFiles and from ve.NewFiles.
+			delete(ve.DeletedFiles, deletedFileEntry{
+				Level:   cl.level,
+				FileNum: curFile.FileNum,
+			})
+			for i := range ve.NewFiles {
+				if ve.NewFiles[i].Meta == curFile {
+					copy(ve.NewFiles[i:], ve.NewFiles[i+1:])
+					ve.NewFiles = ve.NewFiles[:len(ve.NewFiles)-1]
+					break
+				}
+			}
+		}
+		if len(newFiles) == 0 {
+			// The entirety of this file was deleted by excise.
+			f = iter.Next()
+			curFile = f
+			continue
+		}
+		// Since the deletion hints are sorted by start keys, we can set curFile to the
+		// last of the new files. We could potentially excise this twice.
+		curFile = newFiles[len(newFiles)-1].Meta
+	}
+	return nil
+}
+
+// Runs a delete-only compaction.
+//
+// d.mu must *not* be held when calling this.
 func (d *DB) runDeleteOnlyCompaction(
-	jobID JobID, c *compaction,
+	jobID JobID, c *compaction, snapshots compact.Snapshots,
 ) (ve *versionEdit, stats compact.Stats, retErr error) {
+	nextFileNumGetter := func() base.FileNum {
+		d.mu.Lock()
+		defer d.mu.Unlock()
+		return d.mu.versions.getNextFileNum()
+	}
+
 	c.metrics = make(map[int]*LevelMetrics, len(c.inputs))
+	// Sort the deletion hints by start key. runDeleteOnlyCompactionForLevel
+	// requires this.
+	sort.Slice(c.deletionHints, func(i, j int) bool {
+		return d.cmp(c.deletionHints[i].start, c.deletionHints[j].start) < 0
+	})
 	ve = &versionEdit{
 		DeletedFiles: map[deletedFileEntry]*fileMetadata{},
 	}
 	for _, cl := range c.inputs {
 		levelMetrics := &LevelMetrics{}
-		iter := cl.files.Iter()
-		for f := iter.First(); f != nil; f = iter.Next() {
-			ve.DeletedFiles[deletedFileEntry{
-				Level:   cl.level,
-				FileNum: f.FileNum,
-			}] = f
+		if err := d.runDeleteOnlyCompactionForLevel(c, cl, levelMetrics, ve, nextFileNumGetter, snapshots); err != nil {
+			return nil, stats, err
 		}
 		c.metrics[cl.level] = levelMetrics
 	}
+	// Refresh the disk available statistic whenever a compaction/flush
+	// completes, before re-acquiring the mutex.
+	d.calculateDiskAvailableBytes()
 	return ve, stats, nil
 }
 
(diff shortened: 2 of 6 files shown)


testdata/compaction_delete_only_hints
```diff
@@ -431,3 +431,223 @@ Deletion hints:
...
+describe-lsm
+----
+L0.0:
+  000005:[b#11,RANGEKEYDEL-n#inf,RANGEDEL]
+L6:
+  000004:[k#10,RANGEKEYSET-o#10,SET]
+
+get-hints
+----
+(none)
```
itsbilal, line 613: I still need to figure out why there are no hints being generated here, but the hint-generating code didn't change in this PR anyway so it's ready for a look.

itsbilal: This is addressed now

jbowens (commented):
> Reviewed 1 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3926)* status: 1 of 8 files reviewed, 6 unresolved discussions (waiting on @itsbilal)

___
*[`compaction.go` line 2404 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O6wmZJv03gkgr-KJlVe:-O6wmZJv03gkgr-KJlVf:bmeg835) ([raw file](https://github.com/cockroachdb/pebble/blob/61109326b0b70ac7cec0a1757d223d1a7682a350/compaction.go#L2404)):*
> ```Go
> 			continue
> 		}
> 		if d.cmp(h.end, f.Smallest.UserKey) <= 0 || !h.canDeleteOrExcise(d.cmp, f, snapshots, d.FormatMajorVersion()) {
> ```

canDeleteOrExcise doesn't only check bounds: it checks sequence number and key types too. doesn't that mean we don't necessarily want to step forward to the next hint? I.e., the next file may also fall wholly within `h`'s bounds and it may be eligible for deletion.
___
*[`compaction.go` line 2488 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O6wl1R12unx7e-kiRNC:-O6wl1R12unx7e-kiRND:b-91m3cs) ([raw file](https://github.com/cockroachdb/pebble/blob/61109326b0b70ac7cec0a1757d223d1a7682a350/compaction.go#L2488)):*
> ```Go
> 		d.mu.Lock()
> 		defer d.mu.Unlock()
> 		return d.mu.versions.getNextFileNum()
> ```

should we update the nextFileNum to be an atomic and rid ourselves of the mutex concern (and the threading of the closure through parameters)?
___
*[`compaction.go` line 2496 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O6wkW-Y9AZnRTVbOHa0:-O6wkW-Y9AZnRTVbOHa1:b4p7q68) ([raw file](https://github.com/cockroachdb/pebble/blob/61109326b0b70ac7cec0a1757d223d1a7682a350/compaction.go#L2496)):*
> ```Go
> 	sort.Slice(c.deletionHints, func(i, j int) bool {
> 		return d.cmp(c.deletionHints[i].start, c.deletionHints[j].start) < 0
> 	})
> ```

nit: use [`slices.SortFunc`](https://pkg.go.dev/slices#SortFunc)
___
*[`ingest.go` line 1710 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O6wid2ZCKTO2gsQrTI6:-O6wid2ZCKTO2gsQrTI7:b-ot76m9) ([raw file](https://github.com/cockroachdb/pebble/blob/61109326b0b70ac7cec0a1757d223d1a7682a350/ingest.go#L1710)):*
> ```Go
> 	ve *versionEdit,
> 	level int,
> 	nextFileNumGetter func() base.FileNum,
> ```

previously `excise` always held DB.mu. now the delete compaction body may call `excise` without holding `DB.mu`. I don't see anything, but just making sure `excise` is truly agnostic to whether `DB.mu` is held now.

Executing this function without the mutex honestly makes much more sense, since we're performing I/O to constrain the bounds. Maybe we should lift the off the `*DB`, taking only what we need to make the relationship to `DB` state more obvious.
___
*[`table_stats.go` line 636 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O6wgaRj3nqcEZIqaEsd:-O6wgaRj3nqcEZIqaEse:bb36r8x) ([raw file](https://github.com/cockroachdb/pebble/blob/61109326b0b70ac7cec0a1757d223d1a7682a350/table_stats.go#L636)):*
> ```Go
> 				if hintSeqNum > file.SmallestSeqNum {
> 					hintSeqNum = file.SmallestSeqNum
> 				}
> ```

should we pull out the logic of the other branch that handles the hint type and avoids updating the hint sequence number for files that should not be affected by the hint?

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3926#-:-O74nV_a3CsxGf2928eC:b-814g8k)* status: 1 of 8 files reviewed, 5 unresolved discussions (waiting on @jbowens)

___
*[`compaction.go` line 2404 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O6wmZJv03gkgr-KJlVe:-O70tGX6B9XKurupXJQp:bj5acf3) ([raw file](https://github.com/cockroachdb/pebble/blob/61109326b0b70ac7cec0a1757d223d1a7682a350/compaction.go#L2404)):*

canDeleteOrExcise doesn't only check bounds: it checks sequence number and key types too. doesn't that mean we don't necessarily want to step forward to the next hint? I.e., the next file may also fall wholly within `h`'s bounds and it may be eligible for deletion.
</blockquote></details>

Good catch and you're right. I had to significantly rework this method to address this, seeing as we basically had to go through every overlapping hint for a file before we could move on from a file.

I've moved towards a quadratic-ish function here (but only in particularly pathological cases), but also capped the number of deletion only compaction hints we resolve in one delete-only compaction to prevent us from scheduling really CPU-hot or massive delete-only compactions. PTAL!
___
*[`compaction.go` line 2488 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O6wl1R12unx7e-kiRNC:-O70mQVf6heJmIFxcQ-M:bkka9ik) ([raw file](https://github.com/cockroachdb/pebble/blob/61109326b0b70ac7cec0a1757d223d1a7682a350/compaction.go#L2488)):*

should we update the nextFileNum to be an atomic and rid ourselves of the mutex concern (and the threading of the closure through parameters)?
</blockquote></details>

I left a TODO for that here, as we seem to update it in a lot of places and can be left to a quick follow-up PR.
___
*[`compaction.go` line 2496 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O6wkW-Y9AZnRTVbOHa0:-O70foQE1CDWneFcy0Z5:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/61109326b0b70ac7cec0a1757d223d1a7682a350/compaction.go#L2496)):*

nit: use [`slices.SortFunc`](https://pkg.go.dev/slices#SortFunc)
</blockquote></details>

Done.
___
*[`ingest.go` line 1710 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O6wid2ZCKTO2gsQrTI6:-O70qc5m2yHrWHbyLrTM:bufq1g0) ([raw file](https://github.com/cockroachdb/pebble/blob/61109326b0b70ac7cec0a1757d223d1a7682a350/ingest.go#L1710)):*

previously `excise` always held DB.mu. now the delete compaction body may call `excise` without holding `DB.mu`. I don't see anything, but just making sure `excise` is truly agnostic to whether `DB.mu` is held now.

Executing this function without the mutex honestly makes much more sense, since we're performing I/O to constrain the bounds. Maybe we should lift the off the `*DB`, taking only what we need to make the relationship to `DB` state more obvious.
</blockquote></details>

Yeah, d.excise is now mu-agnostic. The only case where we needed the mutex was if we were getting the next file num, which is also abstracted out now. Clarified in a comment.

I tried just removing *DB as a receiver but I had to add a whole bunch of new arguments to this method, from options to table cache to newIters to comparer, so I figured it's best to just leave it as-is.
___
*[`table_stats.go` line 636 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O6wgaRj3nqcEZIqaEsd:-O70gX2r8eEzs2c9BdaL:b-oaqqw) ([raw file](https://github.com/cockroachdb/pebble/blob/61109326b0b70ac7cec0a1757d223d1a7682a350/table_stats.go#L636)):*

should we pull out the logic of the other branch that handles the hint type and avoids updating the hint sequence number for files that should not be affected by the hint?
</blockquote></details>

Good catch! Done - the hint seqnum updating is now also conditional on key type in this branch.

itsbilal:
*[`compaction.go` line 2404 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O6wmZJv03gkgr-KJlVe:-O74nX47DE4TJTsgeHJT:b2h3kd4) ([raw file](https://github.com/cockroachdb/pebble/blob/61109326b0b70ac7cec0a1757d223d1a7682a350/compaction.go#L2404)):*

Good catch and you're right. I had to significantly rework this method to address this, seeing as we basically had to go through every overlapping hint for a file before we could move on from a file.

I've moved towards a quadratic-ish function here (but only in particularly pathological cases), but also capped the number of deletion only compaction hints we resolve in one delete-only compaction to prevent us from scheduling really CPU-hot or massive delete-only compactions. PTAL!
</blockquote></details>

Maybe what we need is a min heap of files (sorted by file smallest keys), and the iteration to work through a sorted list of hints. If we part-excise a file, we add its remnants to the heap. I'll try to think of whether that has any edge cases that I couldn't think of, or if that's the strictly better solution

jbowens (commented):
Still reviewing `runDeleteOnlyCompactionForLevel`

> Reviewed 2 of 6 files at r1, 1 of 3 files at r2, 1 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3926#-:-O7BeJvrEuEXaYXdk9XC:b9uj37v)* status: 5 of 8 files reviewed, 8 unresolved discussions (waiting on @itsbilal)

___
*[`compaction.go` line 2395 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O7Balgc7oCA99wLuWes:-O7Balgc7oCA99wLuWet:bhesf6p) ([raw file](https://github.com/cockroachdb/pebble/blob/f6be059593de469c9d76789074fc67cc96c112ad/compaction.go#L2395)):*
> ```Go
> 	snapshots compact.Snapshots,
> ) (newFiles []manifest.NewFileEntry, err error) {
> 	if !h.canDeleteOrExcise(d.cmp, f, snapshots, d.FormatMajorVersion()) {
> ```

nit: I think we should return an enum from `canDeleteOrExcise` telling us whether we can delete the file, excise the file or neither. always nice to avoid the cost of the extra key comparisons, but more importantly we avoid duplicating fiddly bounds checking logic
___
*[`compaction.go` line 2413 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O7BbZse85HdQZZyVEjj:-O7BbZse85HdQZZyVEjk:bmesj1n) ([raw file](https://github.com/cockroachdb/pebble/blob/f6be059593de469c9d76789074fc67cc96c112ad/compaction.go#L2413)):*
> ```Go
> 	// to use d.excise.
> 	if d.FormatMajorVersion() < FormatVirtualSSTables {
> 		// A future hint will apply to this file and delete it in its entirety.
> ```

nit: is this comment meant to say "A future compaction will find the entirety of this file deleted by range tombstones." or the like?
___
*[`compaction.go` line 2425 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O7Bc4X00LnDzRH65knS:-O7Bc4X00LnDzRH65knT:b-43i060) ([raw file](https://github.com/cockroachdb/pebble/blob/f6be059593de469c9d76789074fc67cc96c112ad/compaction.go#L2425)):*
> ```Go
> 		FileNum: f.FileNum,
> 	}]; !ok {
> 		// This hint did not touch this file.
> ```

Can you add a comment describing under what circumstances this can happen?
___
*[`compaction.go` line 2443 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O7BA6V63pcW5wumugoa:-O7BA6V76cghJQv9Ntes:bvgmi1f) ([raw file](https://github.com/cockroachdb/pebble/blob/f6be059593de469c9d76789074fc67cc96c112ad/compaction.go#L2443)):*
> ```Go
> 	iter := cl.files.Iter()
> 
> 	// Outer loop loops on files. Middle lop loops on fragments. Inner loop
> ```

nit: s/lop/loop/

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3926#-:-O7_pAmrBC85CwQBAQ4o:b-814g8k)* status: 4 of 8 files reviewed, 6 unresolved discussions (waiting on @jbowens)

___
*[`compaction.go` line 2395 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O7Balgc7oCA99wLuWes:-O7_nsOD7Sq3fMvYT_Ua:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/f6be059593de469c9d76789074fc67cc96c112ad/compaction.go#L2395)):*

nit: I think we should return an enum from `canDeleteOrExcise` telling us whether we can delete the file, excise the file or neither. always nice to avoid the cost of the extra key comparisons, but more importantly we avoid duplicating fiddly bounds checking logic
</blockquote></details>

Done.
___
*[`compaction.go` line 2413 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O7BbZse85HdQZZyVEjj:-O7_ntEoCggFlt6Q8Qvj:b-gq9i0g) ([raw file](https://github.com/cockroachdb/pebble/blob/f6be059593de469c9d76789074fc67cc96c112ad/compaction.go#L2413)):*

nit: is this comment meant to say "A future compaction will find the entirety of this file deleted by range tombstones." or the like?
</blockquote></details>

I just meant that one of the other hints must be deleting the entirety of this file, otherwise we wouldn't have picked it for this delete-only compaction. If we don't find such a hint, the caller will panic anyway.
___
*[`compaction.go` line 2425 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O7Bc4X00LnDzRH65knS:-O7_p5vl5IlQm1i5zZd8:bgfmx8t) ([raw file](https://github.com/cockroachdb/pebble/blob/f6be059593de469c9d76789074fc67cc96c112ad/compaction.go#L2425)):*

Can you add a comment describing under what circumstances this can happen?
</blockquote></details>

Done  - removed this case as it should be impossible. Artifact from a previous iteration.
___
*[`compaction.go` line 2443 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O7BA6V63pcW5wumugoa:-O7_oP427MSoeLt8iQ8s:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/f6be059593de469c9d76789074fc67cc96c112ad/compaction.go#L2443)):*

nit: s/lop/loop/
</blockquote></details>

Done.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3926)* status: 4 of 8 files reviewed, 6 unresolved discussions (waiting on @itsbilal)

___
*[`compaction.go` line 2488 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O6wl1R12unx7e-kiRNC:-O82WQrn-w6tltByTwtd:bwa0ucs) ([raw file](https://github.com/cockroachdb/pebble/blob/61109326b0b70ac7cec0a1757d223d1a7682a350/compaction.go#L2488)):*

I left a TODO for that here, as we seem to update it in a lot of places and can be left to a quick follow-up PR.
</blockquote></details>

nit: it'd reduce the code churn within this PR (since this would no longer need to pass a func into `excise`) to do it first (and can still be a quick precursor PR)
___
*[`compaction.go` line 2413 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O7BbZse85HdQZZyVEjj:-O82VbkP9Qebam_pI_RV:b-b6c4yk) ([raw file](https://github.com/cockroachdb/pebble/blob/f6be059593de469c9d76789074fc67cc96c112ad/compaction.go#L2413)):*

I just meant that one of the other hints must be deleting the entirety of this file, otherwise we wouldn't have picked it for this delete-only compaction. If we don't find such a hint, the caller will panic anyway.
</blockquote></details>

I'm still confused by this comment then. Wouldn't that have been a different hint? And this hint would be `hintDoesNotApply`?  because we do:
```
if fmv < FormatVirtualSSTables {
		// The file's keys must be completely contained within the hint range; excises
		// aren't allowed.
		return hintDoesNotApply
}
```

seems like this should be a panic right here.
___
*[`compaction.go` line 2530 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O82X1RNAQ-6AKEeehB_:-O82X1RNAQ-6AKEeehBa:b-9fl0z2) ([raw file](https://github.com/cockroachdb/pebble/blob/9e12f8560e4e1eb8501f2e7f5b02ee67bd84a4d4/compaction.go#L2530)):*
> ```Go
> }
> 
> type deleteCompactionHintFragment struct {
> ```

i think we need some comments describing what this fragmentation is and what it's achieving. I didn't see that anywhere and it's not obvious to me

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3926#-:-O83RanA5W09sVwaYndk:b-814g8k)* status: 2 of 9 files reviewed, 5 unresolved discussions (waiting on @jbowens)

___
*[`compaction.go` line 2488 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O6wl1R12unx7e-kiRNC:-O83RW-m3SraLIxs2j4K:b-dus9ml) ([raw file](https://github.com/cockroachdb/pebble/blob/61109326b0b70ac7cec0a1757d223d1a7682a350/compaction.go#L2488)):*

nit: it'd reduce the code churn within this PR (since this would no longer need to pass a func into `excise`) to do it first (and can still be a quick precursor PR)
</blockquote></details>

Done #3986 , and rebased. 
___
*[`compaction.go` line 2413 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O7BbZse85HdQZZyVEjj:-O83P3To2uMbbxiB098q:bbkzs6s) ([raw file](https://github.com/cockroachdb/pebble/blob/f6be059593de469c9d76789074fc67cc96c112ad/compaction.go#L2413)):*

I'm still confused by this comment then. Wouldn't that have been a different hint? And this hint would be `hintDoesNotApply`?  because we do:
```
if fmv < FormatVirtualSSTables {
		// The file's keys must be completely contained within the hint range; excises
		// aren't allowed.
		return hintDoesNotApply
}
```

seems like this should be a panic right here.
</blockquote></details>

Yeah, now that we do hintDoesNotApply in this case, this should be a panic. The idea was that if you can't virtualize sstables, and you see a hint that partially overlaps with a file, you skip it because you'll definitely see another hint that fully overlaps with that file later on (otherwise we wouldn't have scheduled this file in the compaction to begin with). And if you didn't, the caller already has a "did I try deleting every single file in the compaction" check that would panic.

It's all a little easier to reason about if you think of delete-only compactions as how they were before this PR. They didn't have any hints in them, they just blindly deleted every file they were created with.

Turned this into a panic now.
___
*[`compaction.go` line 2530 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O82X1RNAQ-6AKEeehB_:-O83RIWvEiYB_PBvkSic:bjlnsg9) ([raw file](https://github.com/cockroachdb/pebble/blob/9e12f8560e4e1eb8501f2e7f5b02ee67bd84a4d4/compaction.go#L2530)):*

i think we need some comments describing what this fragmentation is and what it's achieving. I didn't see that anywhere and it's not obvious to me
</blockquote></details>

Added one. Let me know if that makes sense. The idea is to simplify the iteration instead of having to backtrack through the hints (or binary search through a list for every file) to answer "what all hints apply to this file"

RaduBerinde (commented):
What prevents us from excising for very small ranges, potentially leading to a ton of small vssts over time? Is it that a delete-only compaction hint only exists if there is a range-del that encompasses an entire file underneath (so that we eliminate at least a table for every excise?)

It feels like maybe we want to have a heuristic along the lines of: only do a delete-only compaction if the total number of ssts doesn't increase? (so e.g. eliminating one sst while splitting 4 other ssts into two each would not happen).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3926#-:-O8Dpyas7h5Po5xz1vI8:b-nzbupn)* status: 2 of 9 files reviewed, 11 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`compaction.go` line 1927 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8DotuO7V4JHSCyBS91:-O8DotuO7V4JHSCyBS92:bo2ori8) ([raw file](https://github.com/cockroachdb/pebble/blob/1be57fcaf50d9652148d2fc012b1721538a5f052/compaction.go#L1927)):*
> ```Go
> 	// hintDoesNotApply indicates that the hint does not apply to the file.
> 	hintDoesNotApply deletionHintOverlap = iota
> 	// hintExcisesFile indicates that the hint excises a portion of the file.
> ```

[nit] and the database format supports excises.
___
*[`compaction.go` line 1997 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8Dp0x8CWBVjBrswCAy:-O8Dp0x8CWBVjBrswCAz:b-o782yh) ([raw file](https://github.com/cockroachdb/pebble/blob/1be57fcaf50d9652148d2fc012b1721538a5f052/compaction.go#L1997)):*
> ```Go
> 		return hintDoesNotApply
> 	}
> 	// Check for any overlap. In cases of partial overlap, we excise the part of the file
> ```

[nit] we can excise
___
*[`compaction.go` line 2006 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8Dp6PO7FSDWtGsc92Z:-O8Dp6PO7FSDWtGsc92_:b-hkj7t5) ([raw file](https://github.com/cockroachdb/pebble/blob/1be57fcaf50d9652148d2fc012b1721538a5f052/compaction.go#L2006)):*
> ```Go
> }
> 
> func checkDeleteCompactionHints(
> ```

[nit] could use a comment, especially explaining the return values
___
*[`compaction.go` line 2016 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8DpCDQAUsvGrXZbpDA:-O8DpCDQAUsvGrXZbpDB:bnggalk) ([raw file](https://github.com/cockroachdb/pebble/blob/1be57fcaf50d9652148d2fc012b1721538a5f052/compaction.go#L2016)):*
> ```Go
> 	var byLevel [numLevels][]*fileMetadata
> 
> 	// Deletion only compactions can be quadratic (O(mn)) in terms of runtime
> ```

[nit] Delete-only
___
*[`compaction.go` line 2079 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8Du4dV75ujrIPH9vms:-O8Du4dV75ujrIPH9vmt:blswxy5) ([raw file](https://github.com/cockroachdb/pebble/blob/1be57fcaf50d9652148d2fc012b1721538a5f052/compaction.go#L2079)):*
> ```Go
> 			iter := overlaps.Iter()
> 			for m := iter.First(); m != nil; m = iter.Next() {
> 				if m.IsCompacting() || h.canDeleteOrExcise(cmp, m, snapshots, fmv) == hintDoesNotApply || files[m] {
> ```

Should we check/assert that for every hint, there is at least one file with `hintDeletesFile`?
___
*[`compaction.go` line 2404 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8DsPzgDsyPqk312RLI:-O8DsPzgDsyPqk312RLJ:bbq8g7r) ([raw file](https://github.com/cockroachdb/pebble/blob/1be57fcaf50d9652148d2fc012b1721538a5f052/compaction.go#L2404)):*
> ```Go
> }
> 
> func (d *DB) applyHintOnFile(
> ```

[nit] needs a comment

itsbilal (commented):
@RaduBerinde Good point. I was hoping to do some size-based heuristic as opposed to an sstable count one, but a count one is easier to implement like you've described so I've gone ahead and implemented it.  We should only be scheduling delete-only compactions if the sstable count for resolving a given hint goes down or stays flat. PTAL!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3926#-:-O8ItmWrAviRp4katWv5:b-wl5mb8)* status: 2 of 9 files reviewed, 6 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`compaction.go` line 1927 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8DotuO7V4JHSCyBS91:-O8Ie9WG9C_yEqZgKRin:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/1be57fcaf50d9652148d2fc012b1721538a5f052/compaction.go#L1927)):*

[nit] and the database format supports excises.
</blockquote></details>

Done.
___
*[`compaction.go` line 1997 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8Dp0x8CWBVjBrswCAy:-O8IeCnn4y2vo10XmIE3:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/1be57fcaf50d9652148d2fc012b1721538a5f052/compaction.go#L1997)):*

[nit] we can excise
</blockquote></details>

Done.
___
*[`compaction.go` line 2006 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8Dp6PO7FSDWtGsc92Z:-O8Ig44l4b2IBrIDcSLT:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/1be57fcaf50d9652148d2fc012b1721538a5f052/compaction.go#L2006)):*

[nit] could use a comment, especially explaining the return values
</blockquote></details>

Done.
___
*[`compaction.go` line 2016 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8DpCDQAUsvGrXZbpDA:-O8Ig4g196y4ulxsLnOz:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/1be57fcaf50d9652148d2fc012b1721538a5f052/compaction.go#L2016)):*

[nit] Delete-only
</blockquote></details>

Done.
___
*[`compaction.go` line 2079 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8Du4dV75ujrIPH9vms:-O8ItVO84Mmkrr2ZgXOg:b-udtdco) ([raw file](https://github.com/cockroachdb/pebble/blob/1be57fcaf50d9652148d2fc012b1721538a5f052/compaction.go#L2079)):*

Should we check/assert that for every hint, there is at least one file with `hintDeletesFile`?
</blockquote></details>

That shouldn't be an issue even if it were to happen. Realized that it would actually happen a fair bit right now because we throw all resolved hints into the delete-only compaction.
___
*[`compaction.go` line 2404 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8DsPzgDsyPqk312RLI:-O8IhW3l7PVxYqgCK3FX:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/1be57fcaf50d9652148d2fc012b1721538a5f052/compaction.go#L2404)):*

[nit] needs a comment
</blockquote></details>

Done.

RaduBerinde (commented):
Nice, this is great! I still need to look some more at the implementation.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3926#-:-O8Jdri24cMb2DD5RuBY:b-1dnhyk)* status: 2 of 9 files reviewed, 6 unresolved discussions (waiting on @jbowens)

RaduBerinde (approved):
:lgtm: Sorry for the delay. Nice work!

As always, all nits are optional suggestions.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3926#-:-O8hgUGX73I0rr_-cJwL:b-cgm0x9)* status: 2 of 9 files reviewed, 17 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`compaction.go` line 2530 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O82X1RNAQ-6AKEeehB_:-O8hZ3xB70s4d-y06CDU:b-3rknkr) ([raw file](https://github.com/cockroachdb/pebble/blob/9e12f8560e4e1eb8501f2e7f5b02ee67bd84a4d4/compaction.go#L2530)):*

Added one. Let me know if that makes sense. The idea is to simplify the iteration instead of having to backtrack through the hints (or binary search through a list for every file) to answer "what all hints apply to this file"
</blockquote></details>

Could use a comment on the struct as well, maybe
```
// deleteCompactionHintFragment represents a fragment of the key space and
// contains a set of deleteCompactionHints that apply to that fragment; a
// fragment starts at the start field and ends where the next fragment starts.
```
___
*[`compaction.go` line 2105 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hbcsb8KJDEmzbPt06:-O8hbcsb8KJDEmzbPt07:bj6a941) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/compaction.go#L2105)):*
> ```Go
> 					// file on the right, decrement the counter once.
> 					if cmp(h.start, m.Smallest.UserKey) > 0 {
> 						filesDeletedByCurrentHint--
> ```

[nit] to make it easier to follow, we could have a separate `virtualFilesAddedByCurrentHint` counter. The condition will become super clear: `if filesDeletedByCurrentHint < virtualFilesAddedByCurrentHint`
___
*[`compaction.go` line 2107 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hbMH78FO4dv6CwkJL:-O8hbMH78FO4dv6CwkJM:bvmq6tw) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/compaction.go#L2107)):*
> ```Go
> 						filesDeletedByCurrentHint--
> 					}
> 					if base.UserKeyExclusive(h.end).CompareUpperBounds(cmp, m.UserKeyBounds().End) < 0 {
> ```

`m.UserKeyBounds().End.IsUpperBoundFor(h.end)`
___
*[`compaction.go` line 2502 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hdNZa9iMYgM4wdp2w:-O8hdNZa9iMYgM4wdp2x:b24zsw6) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/compaction.go#L2502)):*
> ```Go
> 	curFragment := 0
> 	iter := cl.files.Iter()
> ```

[nit] perhaps assert that cl.level > 0?
___
*[`compaction.go` line 2509 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hcbzdCzVslp0Nwrmc:-O8hcbzdCzVslp0Nwrmd:b-ceft40) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/compaction.go#L2509)):*
> ```Go
> 	for f := iter.First(); f != nil; f = iter.Next() {
> 		// curFile usually matches f, except if f got excised in which case
> 		// it maps to the "current state" of f.
> ```

[nit] maps to a virtual file that replaces f, or nil if f got removed in its entirety
___
*[`compaction.go` line 2511 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hcQmu1G2KmrbrtqK5:-O8hcQmv3LV-mtZpkwU6:b-se7ymz) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/compaction.go#L2511)):*
> ```Go
> 		// it maps to the "current state" of f.
> 		curFile := f
> 		for curFragment < len(fragments) && d.cmp(fragments[curFragment].start, f.Smallest.UserKey) <= 0 {
> ```

[nit] It would be cleaner if `curFragment` never went back; we could do `curFragment < len()-1 && d.cmp(fragments[curFragment+1].start` here; below we can do
```
for {
  if curFragment == len(fragments)-1 || !f.UserKeyBounds().End.IsUpperBound(fragments[curFragment+1].start) {
    break
  }
  curFragment++
  ..
```
___
*[`compaction.go` line 2539 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hfKaPEYO9fTF2wl0K:-O8hfKaPEYO9fTF2wl0L:bs0x6xb) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/compaction.go#L2539)):*
> ```Go
> 					return err
> 				}
> 				if _, ok := ve.DeletedFiles[manifest.DeletedFileEntry{Level: cl.level, FileNum: curFile.FileNum}]; ok {
> ```

Couldn't we do this at the very end?
```go
	ve.NewFiles = slices.DeleteFunc(ve.NewFiles, func(e manifest.NewFileEntry) bool {
		deletedFileEntry := manifest.DeletedFileEntry{Level: cl.level, FileNum: e.Meta.FileNum}
		if _, deleted := ve.DeletedFiles[deletedFileEntry]; deleted {
			delete(ve.DeletedFiles, deletedFileEntry)
			return true
		}
		return false
	})
```

This could actually be a method on `VersionEdit` (with a unit test).
___
*[`compaction.go` line 2595 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hYm9r9IMOcYfMpL-t:-O8hYm9r9IMOcYfMpL-u:b-f3kpni) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/compaction.go#L2595)):*
> ```Go
> 		return cmp(i.start, j.start)
> 	})
> 	j := 0
> ```

[nit] `fragments = slices.CompactFunc(fragments, ...)`
___
*[`table_stats.go` line 559 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hai41C2jBwaEUyKEU:-O8hai41C2jBwaEUyKEV:b-o8e19r) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/table_stats.go#L559)):*
> ```Go
> 		iter := overlaps.Iter()
> 		for file := iter.First(); file != nil; file = iter.Next() {
> 			startCmp := d.cmp(start, file.Smallest.UserKey)
> ```

[nit] can move these close to where they are first used
___
*[`table_stats.go` line 605 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hanlV3ARQ1QEVmfrn:-O8hanlV3ARQ1QEVmfro:bnr1fbp) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/table_stats.go#L605)):*
> ```Go
> 					hintSeqNum = file.SmallestSeqNum
> 				}
> 			} else if d.cmp(file.Smallest.UserKey, end) <= 0 && d.cmp(start, file.Largest.UserKey) <= 0 {
> ```

Shouldn't the first one be `< 0`? And the second one should check if Largest is exclusive?
___
*[`testdata/compaction_delete_only_hints` line 468 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hgG2N-Ag5Ll0htfL2:-O8hgG2N-Ag5Ll0htfL3:bv7a54d) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/testdata/compaction_delete_only_hints#L468)):*
> ```
> first
> next
> next
> ```

[nit] remove the last two `next`s
___
*[`testdata/compaction_delete_only_hints` line 495 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hgIwg4fpgUIyQy0Qa:-O8hgIwg4fpgUIyQy0Qb:b-7ropln) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/testdata/compaction_delete_only_hints#L495)):*
> ```
> first
> next
> next
> ```

ditto
___
*[`testdata/compaction_delete_only_hints` line 571 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hgNjk5wEdefBrPiwJ:-O8hgNjk5wEdefBrPiwK:b-7ropln) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/testdata/compaction_delete_only_hints#L571)):*
> ```
> first
> next
> next
> ```

ditto

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3926#-:-O8lWrg354BqeBxXBe5Y:b-814g8k)* status: 2 of 9 files reviewed, 8 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`compaction.go` line 2530 at r4](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O82X1RNAQ-6AKEeehB_:-O8lXJ9qDc22E5hl09HB:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9e12f8560e4e1eb8501f2e7f5b02ee67bd84a4d4/compaction.go#L2530)):*

Could use a comment on the struct as well, maybe
```
// deleteCompactionHintFragment represents a fragment of the key space and
// contains a set of deleteCompactionHints that apply to that fragment; a
// fragment starts at the start field and ends where the next fragment starts.
```
</blockquote></details>

Done.
___
*[`compaction.go` line 2107 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hbMH78FO4dv6CwkJL:-O8lZX0e4tC9kcROdb9T:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/compaction.go#L2107)):*

`m.UserKeyBounds().End.IsUpperBoundFor(h.end)`
</blockquote></details>

Done.
___
*[`compaction.go` line 2502 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hdNZa9iMYgM4wdp2w:-O8lX0Wj4Kiz6k-bYuak:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/compaction.go#L2502)):*

[nit] perhaps assert that cl.level > 0?
</blockquote></details>

Done.
___
*[`compaction.go` line 2509 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hcbzdCzVslp0Nwrmc:-O8lX7VqC6XOMv0n4CNd:b-7rl0jd) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/compaction.go#L2509)):*

[nit] maps to a virtual file that replaces f, or nil if f got removed in its entirety
</blockquote></details>

done.
___
*[`compaction.go` line 2511 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hcQmu1G2KmrbrtqK5:-O8lhlg12LI_Kxx_ufOa:b-78ewpe) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/compaction.go#L2511)):*

[nit] It would be cleaner if `curFragment` never went back; we could do `curFragment < len()-1 && d.cmp(fragments[curFragment+1].start` here; below we can do
```
for {
  if curFragment == len(fragments)-1 || !f.UserKeyBounds().End.IsUpperBound(fragments[curFragment+1].start) {
    break
  }
  curFragment++
  ..
```
</blockquote></details>

There are a few subtleties here that makes this a little more complex than this. (Trying your approach directly results in an off-by-one). Let me think about this again as a follow-up PR.
___
*[`compaction.go` line 2539 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hfKaPEYO9fTF2wl0K:-O8lhAg38mqiHby-bXr5:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/compaction.go#L2539)):*

Couldn't we do this at the very end?
```go
	ve.NewFiles = slices.DeleteFunc(ve.NewFiles, func(e manifest.NewFileEntry) bool {
		deletedFileEntry := manifest.DeletedFileEntry{Level: cl.level, FileNum: e.Meta.FileNum}
		if _, deleted := ve.DeletedFiles[deletedFileEntry]; deleted {
			delete(ve.DeletedFiles, deletedFileEntry)
			return true
		}
		return false
	})
```

This could actually be a method on `VersionEdit` (with a unit test).
</blockquote></details>

Done.
___
*[`compaction.go` line 2595 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hYm9r9IMOcYfMpL-t:-O8lXajj3kiD7SR1JuLm:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/compaction.go#L2595)):*

[nit] `fragments = slices.CompactFunc(fragments, ...)`
</blockquote></details>

Done.
___
*[`table_stats.go` line 559 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hai41C2jBwaEUyKEU:-O8lXl0lAJRLZtGvk0pX:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/table_stats.go#L559)):*

[nit] can move these close to where they are first used
</blockquote></details>

Done.
___
*[`table_stats.go` line 605 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hanlV3ARQ1QEVmfrn:-O8lXx2R-yrJnAqwl_mF:bqu391q) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/table_stats.go#L605)):*

Shouldn't the first one be `< 0`? And the second one should check if Largest is exclusive?
</blockquote></details>

That seems right, but I didn't change this code and I was a little worried about changing existing table stats logic. It should be relatively harmless but let me look at this as a followup PR.
___
*[`testdata/compaction_delete_only_hints` line 468 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hgG2N-Ag5Ll0htfL2:-O8lZ0Hg0nj9jwpRIE9i:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/testdata/compaction_delete_only_hints#L468)):*

[nit] remove the last two `next`s
</blockquote></details>

Done.
___
*[`testdata/compaction_delete_only_hints` line 495 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hgIwg4fpgUIyQy0Qa:-O8lZ0x10uYGeGhRFIXv:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/testdata/compaction_delete_only_hints#L495)):*

ditto
</blockquote></details>

Done.
___
*[`testdata/compaction_delete_only_hints` line 571 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3926#-O8hgNjk5wEdefBrPiwJ:-O8lZ1Tx5pMf7xoLoKtp:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/e3fe2c7c0169ae425fc2abaf5d187d4f5ff027d8/testdata/compaction_delete_only_hints#L571)):*

ditto
</blockquote></details>

Done.

itsbilal:
🤔 
Screenshot 2024-10-09 at 11 10 52 AM

PR #3927, opened 2024-09-09
colblk: allow finishing data, index blocks without last row

jbowens:
Adapt the DataBlockWriter and IndexBlockWriter interfaces to allow a user to finish the block without the most recently added row. This will be used during sstable building when deciding whether to flush a block with or without an additional KV. The sstable writer will add a new KV to the writer, query the new size, and then decide whether to flush without the new KV (eg, if flushing without the KV reduces memory fragmentation by more closely fitting a configured size class).

### DIFF
--- a/internal/crdbtest/crdbtest.go
+++ b/internal/crdbtest/crdbtest.go
@@ -103,6 +103,30 @@ func EncodeMVCCKey(dst []byte, key []byte, walltime uint64, logical uint32) []by
 	return EncodeTimestamp(dst, walltime, logical)
 }
 
+// AppendTimestamp appends an encoded MVCC timestamp onto key, returning the new
+// key. The provided key should already have the 0x00 sentinel byte (eg, key
+// should be a proper prefix from the perspective of Pebble).
+func AppendTimestamp(key []byte, walltime uint64, logical uint32) []byte {
+	if key[len(key)-1] != 0 {
+		panic(errors.AssertionFailedf("key does not end with 0x00 sentinel byte: %x", key))
+	}
+	if logical == 0 {
+		if walltime == 0 {
+			return key
+		}
+		key = append(key, make([]byte, 9)...)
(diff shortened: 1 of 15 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3927#-:-O6OVoiJ9kpNiYAUuf0q:bnfp4nl)* status: 0 of 15 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`internal/crdbtest/crdbtest.go` line 107 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3927#-O6OEQeR-LWStnCh9kKf:-O6OEQeS5_g7rWQxGBu6:b-1m353q) ([raw file](https://github.com/cockroachdb/pebble/blob/6cda0dd426969a97fe9f2833a495c63b6b14bab6/internal/crdbtest/crdbtest.go#L107)):*
> ```Go
> 
> // AppendTimestamp appends an encoded MVCC timestamp onto key, returning the new
> // key. The provided key should already have the 0x00 sentinel byte (eg, key
> ```

[nit] `i.e.` (id est = that is) not `eg` (exempli gratia = for example)
___
*[`sstable/colblk/data_block.go` line 502 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3927#-O6OVTR92RCemywV4UXU:-O6OVTR92RCemywV4UXV:b-ypiytk) ([raw file](https://github.com/cockroachdb/pebble/blob/6cda0dd426969a97fe9f2833a495c63b6b14bab6/sstable/colblk/data_block.go#L502)):*
> ```Go
> // Finish serializes the pending data block, including the first [rows] rows.
> // The value of [rows] must be Rows() or Rows()-1. The provided size must be
> // the size of the data block with the provided row count (eg, the return value
> ```

[nit] `i.e.`

PR #3950, opened 2024-09-19
colblk: allow PrefixBytes with all empty slices

jbowens:
Fix PrefixBytes to allow construction of a PrefixBytes that only consists of empty slices. Although the PrefixBytes encoding imparts special significance to empty slices (to indicate a duplicate key), we can still represent the empty slice because the slices are required to be lexicographically ordered and the empty slice sorts before all other slices.

### DIFF
--- a/sstable/colblk/prefix_bytes.go
+++ b/sstable/colblk/prefix_bytes.go
@@ -746,9 +746,6 @@ func (b *PrefixBytesBuilder) Put(key []byte, bytesSharedWithPrev int) {
 	prev := &b.sizings[1-currIdx]
 
 	if invariants.Enabled {
-		if len(key) == 0 {
-			panic(errors.AssertionFailedf("key must be non-empty"))
-		}
 		if b.maxShared == 0 {
 			panic(errors.AssertionFailedf("maxShared must be positive"))
 		}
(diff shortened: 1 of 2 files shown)


RaduBerinde:
Why do we want to allow empty suffix keys? Seems like we're just inviting corner cases that should never happen in practice

jbowens:
>  Why do we want to allow empty prefix keys? Seems like we're just inviting corner cases that should never happen in practice

My concern is that with online restore and synthetic prefix, you might have a logical key that consists only of the synthetic prefix, so the physically stored key is the empty slice. Maybe we can convince ourselves that that's not possible in the CockroachDB key encoding, but we need to be very sure. Are there any SQL encodings of primary key values that encode to an empty slice? For example, if you have a primary key on a `bytes` column and you have a row with an empty column value, would it not get encoded to `/Tenant/Table/Index/`? Which then backed up to an online restore-compatible sstable would encode just the empty slice for the user key within the data block.

RaduBerinde:
No, all encoded keys have at least a tag. You can't have an all-null key (null index values have the PK appended).

jbowens:
> No, all encoded keys have at least a tag. You can't have an all-null key (null index values have the PK appended).

Ok, sounds good! Agreed it's much better if we can avoid it!

PR #3951, opened 2024-09-19
block: clarify the SyntheticPrefix IterTransform documentation

jbowens:
The term 'prefix' is loaded in the context of Pebble, and this commit clarifies the relationship of a SyntheticPrefix to the Comparer's concept of a prefix.

### DIFF
--- a/sstable/block/block.go
+++ b/sstable/block/block.go
@@ -282,14 +282,25 @@ func (ss SyntheticSuffix) IsSet() bool {
 }
 
 // SyntheticPrefix represents a byte slice that is implicitly prepended to every
-// key in a file being read or accessed by a reader.  Note that the table is
-// assumed to contain "prefix-less" keys that become full keys when prepended
-// with the synthetic prefix. The table's bloom filters are constructed only on
-// the "prefix-less" keys in the table, but interactions with the file including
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3951#-:-O7B2Lj64QPFXAypk0zs:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/block/block.go` line 298 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3951#-O7B2Ccy5MCyedSfUOmd:-O7B2Ccy5MCyedSfUOme:b-elafa4) ([raw file](https://github.com/cockroachdb/pebble/blob/e9f5b4dc7fcb9b1c2bbf2a48774f52a0d71381af/sstable/block/block.go#L298)):*
> ```Go
> // The table's bloom filters are constructed only on the partial keys physically
> // stored in the table, but interactions with the file including seeks and
> // reads, will all behave as if the file had been constructed from keys that
> ```

nit: extra comma after reads

jbowens:
TFTR!

PR #3955, opened 2024-09-23
db: refactor replayWAL to use flushes to make versionEdits

itsbilal:
Previously, replayWAL was too tightly coupled with the internals of a flush; it ran a flush directly and would produce a versionEdit that would then be accumulated manually (eg. without a BVE) using assumptions true only for memtable flushes. This required relatively intricate interactions with flushable ingests, and meant that flushable ingested sstables would always go into L0.

To pave the way for flushable ingests+excises, we refactor out all flush logic from replayWAL, and have replayWAL just replay WALs into the flushable queue. After this, Open() just schedules flushes until we're done.

Informs #2676.

### DIFF
--- a/obsolete_files.go
+++ b/obsolete_files.go
@@ -338,7 +338,7 @@ func (d *DB) onObsoleteTableDelete(fileSize uint64, isLocal bool) {
 // are not actually deleted by this method. A subsequent call to
 // deleteObsoleteFiles must be performed. Must be not be called concurrently
 // with compactions and flushes. db.mu must be held when calling this function.
-func (d *DB) scanObsoleteFiles(list []string) {
+func (d *DB) scanObsoleteFiles(list []string, flushableIngests []*ingestedFlushable) {
 	// Disable automatic compactions temporarily to avoid concurrent compactions /
 	// flushes from interfering. The original value is restored on completion.
 	disabledPrev := d.opts.DisableAutomaticCompactions
@@ -356,14 +356,11 @@ func (d *DB) scanObsoleteFiles(list []string) {
 	d.mu.versions.addLiveFileNums(liveFileNums)
 	// Protect against files which are only referred to by the ingestedFlushable
 	// from being deleted. These are added to the flushable queue on WAL replay
-	// during read only mode and aren't part of the Version. Note that if
-	// !d.opts.ReadOnly, then all flushables of type ingestedFlushable have
-	// already been flushed.
-	for _, fEntry := range d.mu.mem.queue {
-		if f, ok := fEntry.flushable.(*ingestedFlushable); ok {
-			for _, file := range f.files {
-				liveFileNums[file.FileBacking.DiskFileNum] = struct{}{}
-			}
+	// and handle their own obsoletion/deletion. We exclude them from this obsolete
+	// file scan to avoid double-deleting these files.
+	for _, f := range flushableIngests {
+		for _, file := range f.files {
+			liveFileNums[file.FileBacking.DiskFileNum] = struct{}{}
 		}
 	}
 
--- a/open.go
+++ b/open.go
@@ -232,6 +232,11 @@ func Open(dirname string, opts *Options) (db *DB, err error) {
 		// allocated memory resources. Note that rather than look for an error, we
 		// look for the return of a nil DB pointer.
 		if r := recover(); db == nil {
+			// If there's an unused, recycled memtable, we need to release its memory.
+			if obsoleteMemTable := d.memTableRecycle.Swap(nil); obsoleteMemTable != nil {
+				d.freeMemTable(obsoleteMemTable)
+			}
+
 			// Release our references to the Cache. Note that both the DB, and
 			// tableCache have a reference. When we release the reference to
 			// the tableCache, and if there are no other references to
@@ -492,8 +497,7 @@ func Open(dirname string, opts *Options) (db *DB, err error) {
 			break
 		}
 	}
-	var ve versionEdit
-	var toFlush flushableList
+	var flushableIngests []*ingestedFlushable
 	for i, lf := range replayWALs {
 		// WALs other than the last one would have been closed cleanly.
 		//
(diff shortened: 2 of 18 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3955)* status: 0 of 17 files reviewed, 4 unresolved discussions (waiting on @itsbilal)

___
*[`open.go` line 775 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3955#-O7VZWD-2wNSbixV5xc2:-O7VZWD-2wNSbixV5xc3:b-w6ovbx) ([raw file](https://github.com/cockroachdb/pebble/blob/52a81375c8f10c696b7a84201592092e5d219b2b/open.go#L775)):*
> ```Go
> 	addFileNum(encodedKey)
> 
> 	for i := 1; i < int(b.Count()); i++ {
> ```

I didn't understand why the logic handles the first key of the batch individually before we enter the loop. Can this start from `i := 0` if we remove the `br.Next()`` up above?
___
*[`open.go` line 803 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3955#-O7V_Ke9EqSt5Y_LXr3_:-O7V_Ke9EqSt5Y_LXr3a:b-bk2lt4) ([raw file](https://github.com/cockroachdb/pebble/blob/52a81375c8f10c696b7a84201592092e5d219b2b/open.go#L803)):*
> ```Go
> 		}
> 		if objMeta.IsRemote() {
> 			readable, err = d.objProvider.OpenForReading(context.TODO(), fileTypeTable, n, objstorage.OpenOptions{MustExist: true})
> ```

shouldn't we be able to do remove the explicit `objProvider.Lookup` call and just unconditionally call `OpenForReading`, letting the `objProvider` internally call into the vfs provider as necessary?
___
*[`open.go` line 879 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3955#-O7VcYvc13Y8fxFDINyF:-O7VcYvc13Y8fxFDINyG:bmjy73l) ([raw file](https://github.com/cockroachdb/pebble/blob/52a81375c8f10c696b7a84201592092e5d219b2b/open.go#L879)):*
> ```Go
> 			return
> 		}
> 		mem.writerUnref()
> ```

why do we need to unref here? won't `mem` still be sitting in the flushable queue, so it should maintain a reference until it's been flushed? I'm probably missing something obvious
___
*[`open.go` line 885 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3955#-O7Vc99u5DXCqltMUot-:-O7Vc99u5DXCqltMUot0:b1g1lor) ([raw file](https://github.com/cockroachdb/pebble/blob/52a81375c8f10c696b7a84201592092e5d219b2b/open.go#L885)):*
> ```Go
> 		if !d.opts.ReadOnly {
> 			entry.flushForced = true
> 		}
> ```

nit:
```
entry.flushForced = !d.opts.ReadOnly
```

jbowens:
Looks like your PR surfaced an unrelated failure on the 24.2 branch in the crossversion metamorphic test run. I'll file an issue.

itsbilal (commented):
Yeah, the failure was on this branch and it was a race between compactions deleting away the flushable ingested file during Open(), and the obsolete file scan in Open(). I've fixed it now, PTAL!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3955#-:-O7ZWQnJ4VVWDQpVc-vB:bo05zd4)* status: 0 of 18 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`open.go` line 775 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3955#-O7VZWD-2wNSbixV5xc2:-O7ZYDBg-JlvIAcSljcA:b-341rr0) ([raw file](https://github.com/cockroachdb/pebble/blob/52a81375c8f10c696b7a84201592092e5d219b2b/open.go#L775)):*

I didn't understand why the logic handles the first key of the batch individually before we enter the loop. Can this start from `i := 0` if we remove the `br.Next()`` up above?
</blockquote></details>

No reason; copy-paste code from before. Removed.
___
*[`open.go` line 803 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3955#-O7V_Ke9EqSt5Y_LXr3_:-O7ZYEXn2O7hlSCxwSlp:b-w5gmsl) ([raw file](https://github.com/cockroachdb/pebble/blob/52a81375c8f10c696b7a84201592092e5d219b2b/open.go#L803)):*

shouldn't we be able to do remove the explicit `objProvider.Lookup` call and just unconditionally call `OpenForReading`, letting the `objProvider` internally call into the vfs provider as necessary?
</blockquote></details>

Removed. Was copy-pasted from what was there before.
___
*[`open.go` line 879 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3955#-O7VcYvc13Y8fxFDINyF:-O7ZWlkO3Omk0-Mf3uQ7:brxmlwc) ([raw file](https://github.com/cockroachdb/pebble/blob/52a81375c8f10c696b7a84201592092e5d219b2b/open.go#L879)):*

why do we need to unref here? won't `mem` still be sitting in the flushable queue, so it should maintain a reference until it's been flushed? I'm probably missing something obvious
</blockquote></details>

The writer unref is supposed to get zeroed when we're done writing to the memtable, which is the case here. The memtable starts off with one writer ref on itself (see memTable.init), and one reader ref on the flushableEntry wrapping it. We still keep the reader ref around until we flush the memtable. If we don't writerUnref here, we leak the memtable.

jbowens (approved):
:lgtm:

Looks great!

> Reviewed 15 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3955#-:-O7ZyKUA6A1tX2bXlson:b-41f3vz)* status: 15 of 18 files reviewed, 3 unresolved discussions (waiting on @itsbilal)

___
*[`obsolete_files.go` line 363 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3955#-O7Zxvr44YdJWScZUx8m:-O7Zxvr44YdJWScZUx8n:bn1sx0z) ([raw file](https://github.com/cockroachdb/pebble/blob/3831e2ba44a808034147352a1739a62ea54efa9d/obsolete_files.go#L363)):*
> ```Go
> 	// already been flushed.
> 	for _, fEntry := range d.mu.mem.queue {
> 		if f, ok := fEntry.flushable.(*ingestedFlushable); ok {
> ```

I didn't follow what necessitated passing the flushable ingests into this function as an argument, or was that a stylistic choice?

Also should we leave a TODO to rework this so that ingested flushable sstables receive a ref when they're replayed (and when ingested) and are unref'd after they've been 'flushed' and are now part of a version.
___
*[`open.go` line 879 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3955#-O7VcYvc13Y8fxFDINyF:-O7ZuljcDQWP1e0vQX6q:b-qlceaj) ([raw file](https://github.com/cockroachdb/pebble/blob/52a81375c8f10c696b7a84201592092e5d219b2b/open.go#L879)):*

The writer unref is supposed to get zeroed when we're done writing to the memtable, which is the case here. The memtable starts off with one writer ref on itself (see memTable.init), and one reader ref on the flushableEntry wrapping it. We still keep the reader ref around until we flush the memtable. If we don't writerUnref here, we leak the memtable.
</blockquote></details>

Ah thanks, I wasn't thinking and missed that this was a _writer_ reference.
___
*[`testdata/checkpoint` line 21 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3955#-O7Zv3G_7wubPymUBHsO:-O7Zv3G_7wubPymUBHsP:b-cab8pb) ([raw file](https://github.com/cockroachdb/pebble/blob/3831e2ba44a808034147352a1739a62ea54efa9d/testdata/checkpoint#L21)):*
> ```
> sync: db
> open-dir: db
> sync: db/MANIFEST-000001
> ```

this sync is gone because we don't replay anything so we don't flush anything, whereas previously we'd unconditionally commit a version edit? is that right?

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3955#-:-O7_B0e32RtTg8nSFhPg:b-814g8k)* status: 15 of 18 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`obsolete_files.go` line 363 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3955#-O7Zxvr44YdJWScZUx8m:-O7_5iDPA4Ck9zytfoHw:b97sy4l) ([raw file](https://github.com/cockroachdb/pebble/blob/3831e2ba44a808034147352a1739a62ea54efa9d/obsolete_files.go#L363)):*

I didn't follow what necessitated passing the flushable ingests into this function as an argument, or was that a stylistic choice?

Also should we leave a TODO to rework this so that ingested flushable sstables receive a ref when they're replayed (and when ingested) and are unref'd after they've been 'flushed' and are now part of a version.
</blockquote></details>

The queue should at this point have no flushable ingests in it, because we flushed it away in the maybeScheduleFlush loop earlier in `Open()`. This didn't matter back when flushable ingests wouldn't leave the queue until after we had cleaned away obsolete files, but now we need to save the flushable ingests externally so we don't double-delete their files.

The ref you're mentioning in the TODO does get added. The issue is that our obsolete file scan in this method doesn't look at refs, it just looks at files not in the current version and deletes them. It's why it had to be explicitly taught about flushable ingests in the first place.
___
*[`testdata/checkpoint` line 21 at r2](https://reviewable.io/reviews/cockroachdb/pebble/3955#-O7Zv3G_7wubPymUBHsO:-O7_9ngn5g0mvZbLzbgf:bgnnk6x) ([raw file](https://github.com/cockroachdb/pebble/blob/3831e2ba44a808034147352a1739a62ea54efa9d/testdata/checkpoint#L21)):*

this sync is gone because we don't replay anything so we don't flush anything, whereas previously we'd unconditionally commit a version edit? is that right?
</blockquote></details>

that's right. we still create and sync the manifest once, we just don't do it as second time because we wouldn't do the empty versionEdit.

PR #3972, opened 2024-09-27
lint: add gcassert

RaduBerinde:
Add the `gcassert` linter, which allows us to use directives like
`gcassert:inline` to assert that the compiler is doing what we want.


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3972)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`sstable/colblk/raw_bytes.go` line 116 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3972#-O7lLsi2CemRd5lDHCLV:-O7lLsi2CemRd5lDHCLW:bsk4dmc) ([raw file](https://github.com/cockroachdb/pebble/blob/b509b79bc0d6351ae6ebb724ebced03d6e707a22/sstable/colblk/raw_bytes.go#L116)):*
> ```Go
> 
> // At returns the []byte at index i. The returned slice should not be mutated.
> func (b RawBytes) At(i int) []byte {
> ```

I have verified that we get an error if I add `gcassert:inline` for this function (unfortunately)

jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3972#-:-O7o-2lMBwCTVsRTdNjl:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`sstable/colblk/raw_bytes.go` line 116 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3972#-O7lLsi2CemRd5lDHCLV:-O7o---GEkCAXB2iybmn:b2nh4dy) ([raw file](https://github.com/cockroachdb/pebble/blob/b509b79bc0d6351ae6ebb724ebced03d6e707a22/sstable/colblk/raw_bytes.go#L116)):*

I have verified that we get an error if I add `gcassert:inline` for this function (unfortunately)
</blockquote></details>

:'(

PR #3978, opened 2024-09-27
sstable: add ComparePrev, SetSnapshotPinnedProperties to RawWriter

jbowens:
Extend the RawWriter interface to support two additional methods required by compactions. The RawRowWriter.UnsafeLastPointUserKey method is adapted into ComparePrev so that it may be implemented by RawColumnWriter without materializing the previous key.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2666,7 +2666,7 @@ func (c *compaction) makeVersionEdit(result compact.Result) (*versionEdit, error
 // compaction or flush.
 func (d *DB) newCompactionOutput(
 	jobID JobID, c *compaction, writerOpts sstable.WriterOptions,
-) (objstorage.ObjectMetadata, *sstable.RawRowWriter, CPUWorkHandle, error) {
+) (objstorage.ObjectMetadata, sstable.RawWriter, CPUWorkHandle, error) {
 	d.mu.Lock()
 	diskFileNum := d.mu.versions.getNextDiskFileNum()
 	d.mu.Unlock()
--- a/internal/compact/run.go
+++ b/internal/compact/run.go
@@ -132,7 +132,7 @@ func (r *Runner) MoreDataToWrite() bool {
 // Result.Tables. Should only be called if MoreDataToWrite() returned true.
 //
 // WriteTable always closes the Writer.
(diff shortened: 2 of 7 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3978#-:-O7p1v5B2tIMG4tZQ57v:bnfp4nl)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`internal/compact/splitting.go` line 209 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3978#-O7p1cHP8oe3wVqp0EZH:-O7p1cHP8oe3wVqp0EZI:b-bp4v7u) ([raw file](https://github.com/cockroachdb/pebble/blob/12dc88771d314a85097ce4df8c1d941f6dc411a7/internal/compact/splitting.go#L209)):*
> ```Go
> // INVARIANT: nextUserKey must match the current frontier.
> func (s *OutputSplitter) ShouldSplitBefore(
> 	nextUserKey []byte, estimatedFileSize uint64, comparePrevFn func([]byte) int,
> ```

[nit] I think we should make this `equalToPrevFn func([]byte) bool` so we don't have to worry about which key is on which side of the comparison

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3978#-:-O7p3ZCV2aINxiocMAkt:b-814g8k)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/compact/splitting.go` line 209 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3978#-O7p1cHP8oe3wVqp0EZH:-O7p3YPhC-B4u1w3xWaA:b-ul8jrf) ([raw file](https://github.com/cockroachdb/pebble/blob/12dc88771d314a85097ce4df8c1d941f6dc411a7/internal/compact/splitting.go#L209)):*

[nit] I think we should make this `equalToPrevFn func([]byte) bool` so we don't have to worry about which key is on which side of the comparison
</blockquote></details>

Good idea

PR #3980, opened 2024-09-27
colblk: ensure IndexIter releases BufferHandles

jbowens:
Previously, when an IndexIter was initialized with InitHandle, it erroneously overwrote the handle on the struct while initializing the rest of the struct. This would result in the handle never being Released.


itsbilal (approved):
LGTM

jbowens:
TFTR!

PR #3984, opened 2024-09-30
colblk: ensure BitmapBuilder Size and Finish agree on encoding

jbowens:
Previously, if a bitmap builder switched from the all-zero encoding to the default encoding, a subsequent call to Finish with a row count that should use the all-zero encoding would erroneously write the default encoding (attempting to write more bytes than were allocated). This commit updates the BitmapBuilder to keep track of the first set bit of the bitmap. During Finish, this is used to determine the appropriate encoding. Invert continues to force use of the default encoding.


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3984)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/colblk/bitmap.go` line 250 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3984#-O834mnv2PeFh-97ajMg:-O83WEjLA6PeJc_OIp73:b4jchf4) ([raw file](https://github.com/cockroachdb/pebble/blob/49fde14862b4d6762b680a76c33bc294954e5d1e/sstable/colblk/bitmap.go#L250)):*

[nit] maybe `minNonZeroRowCount`
</blockquote></details>

Sure, done
___
*[`sstable/colblk/bitmap.go` line 281 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3984#-O834zO80b5L8Aq3eiym:-O83VtYE8bv5Dek_fzTL:bkdf7d7) ([raw file](https://github.com/cockroachdb/pebble/blob/49fde14862b4d6762b680a76c33bc294954e5d1e/sstable/colblk/bitmap.go#L281)):*

else `x = min(x, i+1)` ?
</blockquote></details>

will the optimizer avoid the store if it's not updated (the expected case)?

RaduBerinde:
*[`sstable/colblk/bitmap.go` line 281 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3984#-O834zO80b5L8Aq3eiym:-O83qbAj573V0bISc0nm:bfbmbqx) ([raw file](https://github.com/cockroachdb/pebble/blob/49fde14862b4d6762b680a76c33bc294954e5d1e/sstable/colblk/bitmap.go#L281)):*

will the optimizer avoid the store if it's not updated (the expected case)?
</blockquote></details>

I would hope it gets converted in a conditional move (CMOV on amd64)

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3984)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/colblk/bitmap.go` line 281 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3984#-O834zO80b5L8Aq3eiym:-O84kJzy7-yBCSGN6BQD:bbcvdcb) ([raw file](https://github.com/cockroachdb/pebble/blob/49fde14862b4d6762b680a76c33bc294954e5d1e/sstable/colblk/bitmap.go#L281)):*

I would hope it gets converted in a conditional move (CMOV on amd64)
</blockquote></details>

I was misunderstanding this original comment until just now. I was under the assumption that a part of the contract was that calls to `Set(i)` must be made with monotonically increasing `i` but we don't actually have that documented. Done. We can always establish that as a part of the contract if we're looking for CPU optimizations here later.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3984#-:-O88B0yN0ddvDok4Kzwf:bnfp4nl)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`sstable/colblk/bitmap.go` line 278 at r3](https://reviewable.io/reviews/cockroachdb/pebble/3984#-O88As6G9CMvCTpINTeH:-O88As6HAyxWM6tgJ23B:b-17wzsr) ([raw file](https://github.com/cockroachdb/pebble/blob/3017d897b27322d33bd7add92a7ca27707b08a61/sstable/colblk/bitmap.go#L278)):*
> ```Go
> 	// to encode any set bits at all. This is used to determine whether the
> 	// bitmap should be encoded using the all-zeros encoding.
> 	if b.minNonZeroRowCount == 0 {
> ```

[nit] could just do `if b.isZero(i+1)`

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3984#-:-O88G46U1opMxz6zswlp:bnfp4nl)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @jbowens)

___
*[`sstable/colblk/bitmap.go` line 275 at r5](https://reviewable.io/reviews/cockroachdb/pebble/3984#-O88Fm871h-21z5FjCtf:-O88Fm871h-21z5FjCtg:b-gd7im9) ([raw file](https://github.com/cockroachdb/pebble/blob/8bab92d69b79e507fb6a9559298b9a48e55d512b/sstable/colblk/bitmap.go#L275)):*
> ```Go
> // Set sets the bit at position i to true.
> func (b *BitmapBuilder) Set(i int) {
> 	// If no bits have been set, record the row count at which the bitmap begins
> ```

[nit] // Update minNonZeroRowCount if necessary. This is used to ..
___
*[`sstable/colblk/bitmap.go` line 343 at r6](https://reviewable.io/reviews/cockroachdb/pebble/3984#-O88Fz0m9u1Y6iTXHfn_:-O88Fz0m9u1Y6iTXHfna:bkk2jkb) ([raw file](https://github.com/cockroachdb/pebble/blob/7cd9806a1a60d63537cea751890186ad051b992e/sstable/colblk/bitmap.go#L343)):*
> ```Go
> 	nBitmapWords := (nRows + 63) >> 6
> 	if len(b.words) < nBitmapWords {
> 		b.words = slices.Grow(b.words, nBitmapWords-len(b.words))
> ```

I wish Grow had the total size as the argument..

PR #3985, opened 2024-09-30
sstable: update unit tests to use block.IndexBlockIterator

jbowens:
Update unit tests to use the IndexBlockIterator interface and be agnostic to the underlying block format.

### DIFF
--- a/sstable/block_property_test.go
+++ b/sstable/block_property_test.go
@@ -24,8 +24,6 @@ import (
 	"github.com/cockroachdb/pebble/internal/keyspan"
 	"github.com/cockroachdb/pebble/internal/rangekey"
 	"github.com/cockroachdb/pebble/internal/testkeys"
-	"github.com/cockroachdb/pebble/sstable/block"
-	"github.com/cockroachdb/pebble/sstable/rowblk"
 	"github.com/stretchr/testify/require"
 )
 
@@ -859,7 +857,7 @@ func TestBlockProperties(t *testing.T) {
 			return runTablePropsCmd(r, td)
 
 		case "block-props":
-			return runBlockPropsCmd(r, td)
+			return runBlockPropsCmd(r)
 
 		case "filter":
 			syntheticSuffix := make([]byte, 0, binary.MaxVarintLen64)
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3985#-:-O837RJU6hMD3T56iSfz:bnfp4nl)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi, @itsbilal, and @jbowens)

___
*[`sstable/testdata/block_properties` line 59 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3985#-O837K1Z1yXv0JEiGk-r:-O837K1Z1yXv0JEiGk-s:b-mbnb44) ([raw file](https://github.com/cockroachdb/pebble/blob/3013f26b8a6427c3f8f16b6ed81c529c92cdd649/sstable/testdata/block_properties#L59)):*
> ```
> block-props
> ----
> c:
> ```

How come the previous test didn't have the correct separator key? I thought the UserKeys in the index block are separator keys?

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3985#-:-O83A3Ad5FqphHEwxg5F:b-814g8k)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi, @itsbilal, and @RaduBerinde)

___
*[`sstable/testdata/block_properties` line 59 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3985#-O837K1Z1yXv0JEiGk-r:-O839iFCCzn-n7x6IIY8:b42ya45) ([raw file](https://github.com/cockroachdb/pebble/blob/3013f26b8a6427c3f8f16b6ed81c529c92cdd649/sstable/testdata/block_properties#L59)):*

How come the previous test didn't have the correct separator key? I thought the UserKeys in the index block are separator keys?
</blockquote></details>

Turns out that they are when we're able to shorten the key in `InternalKey.Separator`, but if we're not able to construct a shorter key we just use the largest key in the data block as-is: https://github.com/cockroachdb/pebble/blob/d93251cce9d77ec15be908d76c96739a0ce37fd7/internal/base/internal.go#L389-L407

PR #3986, opened 2024-09-30
db: use atomic for versionSet.nextFileNum

itsbilal:
Currently, we have the nextFileNum counter protected by db.mu, which is pretty unnecessary and results in us grabbing the db mutex just for being able to generate a new filenum. This change moves that to an atomic.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2667,9 +2667,7 @@ func (c *compaction) makeVersionEdit(result compact.Result) (*versionEdit, error
 func (d *DB) newCompactionOutput(
 	jobID JobID, c *compaction, writerOpts sstable.WriterOptions,
 ) (objstorage.ObjectMetadata, sstable.RawWriter, CPUWorkHandle, error) {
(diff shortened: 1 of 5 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3986#-:-O83H_tBDc72MfVIHcha:bnfp4nl)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @sumeerbhola)

RaduBerinde:
OOC, what motivated this change? I'd think generating a new file happens much too rarely for lock vs atomic to matter?

itsbilal:
@RaduBerinde the motivation was #3926, where the only difference between the multiple callers to `db.excise` is whether they have `db.mu` held or not

itsbilal:
TFTR!

RaduBerinde:
Nice, that makes a lot of sense!

PR #3987, opened 2024-09-30
colblk: safeguard against Go pointer rule violations in DataBlockRewriter

jbowens:
Go pointer rules require that a Go pointer always point to a byte within a valid allocation. The DataBlockRewriter previously would allocate buffers for keys exactly sized for the largest user key contained within a block. If one were to use pointer arithmetic to compute an exclusive pointer to byte beyond the key, this pointer would violate Go's pointer rules.

As far as I can tell, we don't do this today because only PrefixBytesIter performs pointer arithmetic that might violate this by computing a pointer to the beginning of the suffix. This is only problematic if there is no suffix, but suffix-rewriting always requires a suffix.

To be safe, we allocate an extra byte.

### DIFF
--- a/sstable/colblk/data_block.go
+++ b/sstable/colblk/data_block.go
@@ -641,10 +641,14 @@ func (rw *DataBlockRewriter) RewriteSuffixes(
 	if err = rw.iter.Init(&rw.reader, rw.keySeeker, nil, block.IterTransforms{}); err != nil {
 		return base.InternalKey{}, base.InternalKey{}, nil, err
 	}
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3987#-:-O84QdAk95Py-ro8zN-T:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @itsbilal)

PR #3989, opened 2024-10-01
block,colblk: tighten iterator semantics

jbowens:
**block,colblk: tighten index iterator semantics**

This commit more tightly defines the IndexBlockIterator semantics around behavior after a call to Invalidate and around positioning operations called after exhaustion. These semantics match the rowblk implementation and are currently required by the sstable iterator.

In the future, we may consider alternative semantics through updating the sstable iterator.

**block,colblk: tighten data block iterator semantics**

Define the semantics of calling a positioning method on an invalidated
DataBlockIterator. The sstable iterator currently depends on these semantics.


RaduBerinde (approved):

jbowens:
TFTR!

PR #3991, opened 2024-10-02
colblk: change CompareFirstUserKey to IsLowerBound

RaduBerinde:
This commit changes the `CompareFirstUserKey` API to `IsLowerBound`
which can be defined when there are no keys to present. We fix the
columnar implementation to take into account `HideObsoletePoints` and
ignore any leading obsolete points in the block.

### DIFF
--- a/internal/itertest/datadriven.go
+++ b/internal/itertest/datadriven.go
@@ -142,6 +142,9 @@ func RunInternalIterCmdWriter(
 		parts := strings.Fields(line)
 		var key *base.InternalKey
 		var value []byte
+		if o.showCommands {
+			fmt.Fprintf(w, "%*s: ", min(maxCmdLen, 40), line)
+		}
 		switch parts[0] {
 		case "seek-ge":
 			if len(parts) < 2 || len(parts) > 3 {
@@ -228,12 +231,18 @@ func RunInternalIterCmdWriter(
 				*o.stats = base.InternalIteratorStats{}
 			}
 			continue
+		case "is-lower-bound":
+			// This command is specific to colblk.DataBlockIter.
+			if len(parts) != 2 {
+				fmt.Fprint(w, "is-lower-bound <key>\n")
+				return
+			}
+			i := iter.(interface{ IsLowerBound(key []byte) bool })
+			fmt.Fprintf(w, "%v\n", i.IsLowerBound([]byte(parts[1])))
+			continue
 		default:
 			fmt.Fprintf(w, "unknown op: %s", parts[0])
 			return
-		}
-		if o.showCommands {
-			fmt.Fprintf(w, "%*s: ", min(maxCmdLen, 40), line)
 		}
 		o.fmtKV(w, key, value, iter)
 		if !o.withoutNewlines {
--- a/sstable/block/block.go
+++ b/sstable/block/block.go
@@ -158,10 +158,14 @@ type DataBlockIterator interface {
 	// KV returns the key-value pair at the current iterator position. The
 	// iterator must be Valid().
 	KV() *base.InternalKV
-	// CompareFirstUserKey compares the provided key to the first user key
-	// contained within the data block. It's equivalent to performing
-	//   Compare(firstUserKey, k)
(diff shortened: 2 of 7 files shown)


jbowens (approved):
> Reviewed 1 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3991)* status: 1 of 7 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/colblk/data_block.go` line 899 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3991#-O8DBj_j8LDXa3bHHwYD:-O8DBj_j8LDXa3bHHwYE:b-yt9e2n) ([raw file](https://github.com/cockroachdb/pebble/blob/09149152ff44b31ab87d8149833ace382783b0db/sstable/colblk/data_block.go#L899)):*
> ```Go
> // IsLowerBound implements the block.DataBlockIterator interface.
> func (i *DataBlockIter) IsLowerBound(k []byte) bool {
> 	// FOO
> ```

accidental comment?
___
*[`sstable/rowblk/rowblk_iter.go` line 531 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3991#-O8DCAAl8tkJEgvBqt4_:-O8DCAAl8tkJEgvBqt4a:b40tm3t) ([raw file](https://github.com/cockroachdb/pebble/blob/09149152ff44b31ab87d8149833ace382783b0db/sstable/rowblk/rowblk_iter.go#L531)):*
> ```Go
> 	// TODO(radu): we're not taking HideObsoletePoints into consideration. Today
> 	// this is ok because the method is used for some optimizations and false
> 	// negatives are ok.
> ```

Should we just define `IsLowerBound` to allow false negatives? I don't think there's much (any?) of a performance advantage to making the tradeoff of seeking in IsLowerBound, and the code becomes a bit more complicated

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3991)* status: 1 of 7 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`sstable/rowblk/rowblk_iter.go` line 531 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3991#-O8DCAAl8tkJEgvBqt4_:-O8DDPakEmgY5uePMUhQ:bz9wjef) ([raw file](https://github.com/cockroachdb/pebble/blob/09149152ff44b31ab87d8149833ace382783b0db/sstable/rowblk/rowblk_iter.go#L531)):*

Should we just define `IsLowerBound` to allow false negatives? I don't think there's much (any?) of a performance advantage to making the tradeoff of seeking in IsLowerBound, and the code becomes a bit more complicated
</blockquote></details>

I think you're right. Synthetic suffix would also add some complexity too. I'll work on that.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3991)* status: 1 of 7 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`sstable/rowblk/rowblk_iter.go` line 531 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3991#-O8DCAAl8tkJEgvBqt4_:-O8DIz4mCqVJDlO8pHSe:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/09149152ff44b31ab87d8149833ace382783b0db/sstable/rowblk/rowblk_iter.go#L531)):*

I think you're right. Synthetic suffix would also add some complexity too. I'll work on that.
</blockquote></details>

Done.

jbowens (approved):
:lgtm:

> Reviewed 3 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3991#-:-O8DXX-i7oBmHZAvqBrf:bnfp4nl)* status: 4 of 7 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #3995, opened 2024-10-02
sstable: populate range deletion, range key props in RawColumnWriter

jbowens:
Update RawColumnWriter to update sstable properties when writing range deletions and range keys. The properties are computed differently in columnar sstables, because the old accounting was tightly coupled with the encoding.


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3995#-:-O8ERvxq52SaaYmYSGtU:bnfp4nl)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @itsbilal)

PR #3999, opened 2024-10-03
colblk: unify MultiDataBlockIter and DataBlockIter

jbowens:
Collapse MultiDataBlockIter and DataBlockIter into one type. This was motivated by chasing down an unintentional allocation in MultiDataBlockIter.InitHandle.

### DIFF
--- a/sstable/colblk/cockroach_test.go
+++ b/sstable/colblk/cockroach_test.go
@@ -370,10 +370,13 @@ func TestCockroachDataBlock(t *testing.T) {
 	serializedBlock, _ := w.Finish(w.Rows(), w.Size())
 	var reader DataBlockReader
 	var it DataBlockIter
-	reader.Init(cockroachKeySchema, serializedBlock)
-	it.Init(&reader, cockroachKeySchema.NewKeySeeker(), getLazyValuer(func([]byte) base.LazyValue {
+	it.InitOnce(cockroachKeySchema, getLazyValuer(func([]byte) base.LazyValue {
 		return base.LazyValue{ValueOrHandle: []byte("mock external value")}
-	}), block.IterTransforms{})
+	}))
+	reader.Init(cockroachKeySchema, serializedBlock)
+	if err := it.Init(&reader, block.IterTransforms{}); err != nil {
+		t.Fatal(err)
+	}
 
 	t.Run("Next", func(t *testing.T) {
 		// Scan the block using Next and ensure that all the keys values match.
@@ -560,10 +563,13 @@ func benchmarkCockroachDataBlockIter(
 	serializedBlock, _ := w.Finish(w.Rows(), w.Size())
 	var reader DataBlockReader
 	var it DataBlockIter
-	reader.Init(cockroachKeySchema, serializedBlock)
-	it.Init(&reader, cockroachKeySchema.NewKeySeeker(), getLazyValuer(func([]byte) base.LazyValue {
+	it.InitOnce(cockroachKeySchema, getLazyValuer(func([]byte) base.LazyValue {
 		return base.LazyValue{ValueOrHandle: []byte("mock external value")}
-	}), transforms)
+	}))
+	reader.Init(cockroachKeySchema, serializedBlock)
+	if err := it.Init(&reader, transforms); err != nil {
+		b.Fatal(err)
+	}
 	avgRowSize := float64(len(serializedBlock)) / float64(count)
 
 	b.Run("Next", func(b *testing.B) {
--- a/sstable/colblk/data_block.go
+++ b/sstable/colblk/data_block.go
@@ -592,6 +592,14 @@ type DataBlockRewriter struct {
 	initialized     bool
 }
 
+type assertNoExternalValues struct{}
+
+var _ block.GetLazyValueForPrefixAndValueHandler = assertNoExternalValues{}
+
+func (assertNoExternalValues) GetLazyValueForPrefixAndValueHandle(value []byte) base.LazyValue {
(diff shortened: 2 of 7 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3999)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`sstable/layout.go` line 334 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3999#-O8Ie70YAhSMTvgCLPdN:-O8Ie70Z9sFUK7TUON92:br9wmim) ([raw file](https://github.com/cockroachdb/pebble/blob/455176abe1a23fcabdc2e772d30940c9599fbaa4/sstable/layout.go#L334)):*
> ```Go
> 	if fmtRecord != nil {
> 		var iter colblk.DataBlockIter
> 		iter.KeySchema = r.keySchema
> ```

[nit] feels weird that these are done separately from `Init`
___
*[`sstable/colblk/data_block.go` line 787 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3999#-O8Ie_1k9lVo5DaarAVE:-O8Ie_1k9lVo5DaarAVF:bwqj1e1) ([raw file](https://github.com/cockroachdb/pebble/blob/455176abe1a23fcabdc2e772d30940c9599fbaa4/sstable/colblk/data_block.go#L787)):*
> ```Go
> // Init initializes the data block iterator, configuring it to read from the
> // provided reader.
> func (i *DataBlockIter) Init(r *DataBlockReader, transforms block.IterTransforms) error {
> ```

These two very different initialization paths are leading to a lot of subtlety.

Do we really ever need the caller to provide the `Reader`? I would change that to a special `InitWithDataSlice` and pass the slice and let `DataBlockIter` handle the reader. Or even better, a `colblk.ForEachKVInBlock()` function so that the caller doesn't even have to worry about the iterator.

Then we wouldn't need `r` and `readAlloc` which may or may not be the same, we would just have `r` as a non-pointer.

Then we could have `InitOnce` which sets the keyschema and allocates the key seeker, and `InitHandle` which can be called multiple times and assumes those fields are initialized.
___
*[`sstable/colblk/data_block.go` line 790 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3999#-O8IeRf8EDaO9truK__r:-O8IeRf8EDaO9truK__s:b-6cqb5a) ([raw file](https://github.com/cockroachdb/pebble/blob/455176abe1a23fcabdc2e772d30940c9599fbaa4/sstable/colblk/data_block.go#L790)):*
> ```Go
> 	numRows := int(r.r.header.Rows)
> 	*i = DataBlockIter{
> 		KeySchema:     i.KeySchema,
> ```

[nit] I noticed that in cases like this Go makes a copy of the entire struct, I would just pull any `i.xx` accesses out into local variables.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3999)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/layout.go` line 334 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3999#-O8Ie70YAhSMTvgCLPdN:-O8JOL5uAmUzLxfInIHH:b-8vwxs2) ([raw file](https://github.com/cockroachdb/pebble/blob/455176abe1a23fcabdc2e772d30940c9599fbaa4/sstable/layout.go#L334)):*

[nit] feels weird that these are done separately from `Init`
</blockquote></details>

refactored into your InitOnce suggestion from the other comment
___
*[`sstable/colblk/data_block.go` line 787 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3999#-O8Ie_1k9lVo5DaarAVE:-O8JOSG87Ni0FH7lF2Es:bepelsg) ([raw file](https://github.com/cockroachdb/pebble/blob/455176abe1a23fcabdc2e772d30940c9599fbaa4/sstable/colblk/data_block.go#L787)):*

These two very different initialization paths are leading to a lot of subtlety.

Do we really ever need the caller to provide the `Reader`? I would change that to a special `InitWithDataSlice` and pass the slice and let `DataBlockIter` handle the reader. Or even better, a `colblk.ForEachKVInBlock()` function so that the caller doesn't even have to worry about the iterator.

Then we wouldn't need `r` and `readAlloc` which may or may not be the same, we would just have `r` as a non-pointer.

Then we could have `InitOnce` which sets the keyschema and allocates the key seeker, and `InitHandle` which can be called multiple times and assumes those fields are initialized.
</blockquote></details>

Hmm yeah the pointer for `*DataBlockReader` was from the forward-looking lens of stashing these readers on the cache value, but we should cross that bridge when we get to it. Will follow up
___
*[`sstable/colblk/data_block.go` line 790 at r1](https://reviewable.io/reviews/cockroachdb/pebble/3999#-O8IeRf8EDaO9truK__r:-O8JORetD4dO2tLPT-4y:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/455176abe1a23fcabdc2e772d30940c9599fbaa4/sstable/colblk/data_block.go#L790)):*

[nit] I noticed that in cases like this Go makes a copy of the entire struct, I would just pull any `i.xx` accesses out into local variables.
</blockquote></details>

Done

RaduBerinde (approved):
:lgtm: Ok with merging this as-is for now

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/3999#-:-O8JYgu--21ox_nqNp2f:b-8db49a)* status: 0 of 7 files reviewed, all discussions resolved

jbowens:
TFTR!

PR #4005, opened 2024-10-04
sstable: clean up TestRewriteSuffixProps, fix subtle bug

jbowens:
This commit cleans up the TestRewriteSuffixProps unit test, refactoring it to use the testkey comparator and new block-property collectors that are functions of the testkey integer suffix.

It also fixes a bug where if the suffix rewriting WriterOptions contained fewer block property collectors than the original sstable had when it was written, suffix rewriting could fail to decode all the relevant properties from a block handle.

### DIFF
--- a/sstable/block_property_test.go
+++ b/sstable/block_property_test.go
@@ -11,7 +11,8 @@ import (
 	"fmt"
 	"io"
 	"math"
-	"math/rand"
+	"math/bits"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -25,6 +26,7 @@ import (
 	"github.com/cockroachdb/pebble/internal/rangekey"
 	"github.com/cockroachdb/pebble/internal/testkeys"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/exp/rand"
 )
 
 func TestIntervalEncodeDecode(t *testing.T) {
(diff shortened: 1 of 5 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4005#-:-O8OmfG67fZOHAgOJ7pL:bnfp4nl)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`sstable/suffix_rewriter.go` line 241 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4005#-O8OkzJUDffXjrfNKMe_:-O8OkzJUDffXjrfNKMea:b-1loaeu) ([raw file](https://github.com/cockroachdb/pebble/blob/293bfa74a71371ead5f189bba16b475e88c20911/sstable/suffix_rewriter.go#L241)):*
> ```Go
> // index of the corresponding block property collector in collectors.
> //
> // Additionally, getShortIDs returns n which is the number of block property
> ```

[nit] this description is confusing. Maybe: 

n is the number of encoded properties that we need to process in order to get values for all the given collectors. This can be different than len(collectors) when the latter is different from the number of block property collectors used when the sstable was written. Note that shortIDs[n:] are all invalid.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4005#-:-O8XfSyn1OSESICVd9GN:b-814g8k)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

PR #4011, opened 2024-10-07
sstable/block: remove FragmentIterTransforms.ElideSameSeqNum

jbowens:
ElideSameSeqNum was only ever set by (*sstable.Reader).NewRawRangeDelIter where it should be no-op (because two range deletions with the same sequence number would be duplicate keys). Remove the option.


itsbilal (approved):
LGTM

jbowens:
TFTR!

PR #4012, opened 2024-10-07
ci: use go1.23

rickystewart:
Part of: CRDB-41921


Makefile
```diff
@@ -132,7 +132,7 @@ endif
 
 .PHONY: format
 format:
-	go install -C internal/devtools github.com/cockroachdb/crlfmt && crlfmt -w -tab 2 ../..
```
rickystewart, line 135: I was pretty perplexed by this. To me this `../..` seems transparently obviously wrong and it should be `.`. This change fixes CI ([link](https://github.com/cockroachdb/pebble/actions/runs/11222624774/job/31195468780?pr=4012)).

RaduBerinde: Indeed, thanks for the fix. I think there was an intermediate version that used `go run` and the working dir would be `./internal/devtools`.

RaduBerinde:
Do we plan to switch to go 1.23 for 24.2.3? If not, we should either hold off on this, or test both 1.22 and 1.23 in CI

rickystewart:
Yes, we plan to switch to Go 1.23 for 24.3.

RaduBerinde (approved):
Thanks!

PR #4014, opened 2024-10-07
colblk: implement FragmentIterTransforms, fix SeekLT

jbowens:
Fix a bug in (*colblk.KeyspanIter).SeekLT that caused the iterator to surface the wrong span, and implement the FragmentIterTransforms on the colblk.KeyspanIter.

### DIFF
--- a/sstable/block/block.go
+++ b/sstable/block/block.go
@@ -272,6 +272,14 @@ type FragmentIterTransforms struct {
 	SyntheticSuffix SyntheticSuffix
 }
 
+// NoTransforms returns true if there are no transforms enabled.
+func (t *FragmentIterTransforms) NoTransforms() bool {
+	// NoTransforms returns true if there are no transforms enabled.
+	return t.SyntheticSeqNum == 0 &&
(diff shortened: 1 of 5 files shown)


jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4014#-:-O8gIlmt4wXy-M5twf6k:b-814g8k)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`sstable/colblk/keyspan.go` line 308 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4014#-O8cvjUF70ZMlMfVbTG7:-O8gInHrC2X2srC_PaO3:b-wl6fxy) ([raw file](https://github.com/cockroachdb/pebble/blob/f352749e62a10c4665c39bcc566e9b8edd055a69/sstable/colblk/keyspan.go#L308)):*

[nit] Why are we keeping both versions? We could just check if `len(syntheticPrefix) > 0` inside.
</blockquote></details>

Good question — done

PR #4016, opened 2024-10-07
*: switch to rand/v2

RaduBerinde:
Switch to using `math/rand/v2` instead of `math/rand` or
`golang.org/x/exp/rand`.

Note that with v2, `rand.Uint64()` is pretty much the same as
`runtime_fastrand` so we can remove the fastrand package, but that is
left for another change.


jbowens (approved):
:lgtm_strong: 

> Reviewed 2 of 109 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4016#-:-O8gJ1oUBh7wfmtgz4AI:b-uwkowd)* status: 2 of 109 files reviewed, all discussions resolved

PR #4018, opened 2024-10-08
internal: remove fastrand

RaduBerinde:
#### fastrand: add v2 benchmarks

This benchmark proves that the fastrand package is not needed anymore
and we can just use `math/rand/v2` methods directly (which now use the
runtime functions underneath).

```
name                     time/op
FastRand-10              0.59ns ± 2%
RandV2-10                0.62ns ± 1%
STFastRand-10            4.84ns ± 0%
STRandV2-10              5.33ns ± 1%
```

#### internal: remove fastrand

This package is no longer necessary, as in `math/rand/v2` the
`rand.Uint32` function is equivalent.


jbowens (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4018#-:-O8h5xrc8U1EU6b0CSGo:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aadityasondhi)

PR #4031, opened 2024-10-10
sstable: remove block transforms

RaduBerinde:
The code to transform a raw block at load time is not currently used
or tested. Removing it to keep the code a bit cleaner (it's easy
enough to add back if we ever need it again).

### DIFF
--- a/sstable/block_property_test.go
+++ b/sstable/block_property_test.go
@@ -1321,7 +1321,7 @@ func runBlockPropsCmd(r *Reader) string {
 		// block that bhp points to, along with its block properties.
 		if twoLevelIndex {
 			subIndex, err := r.readBlock(
-				context.Background(), bhp.Handle, nil, nil, nil, nil, nil)
+				context.Background(), bhp.Handle, nil, nil, nil, nil)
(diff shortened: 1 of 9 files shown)


aadityasondhi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4031#-:-O8rAERW6fP2pSTgVptO:bnfp4nl)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`sstable/reader.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4031#-O8rA8a9AY3LXXES8kD-:-O8rA8a9AY3LXXES8kD0:bny3epg) ([raw file](https://github.com/cockroachdb/pebble/blob/27acbe1d595cf434f9cc82f012bc96a1aa668416/sstable/reader.go#L47)):*
> ```Go
> )
> 
> type blockTransform func([]byte) ([]byte, error)
> ```

I'm guessing the code that would initialize this function no longer exists?

PR #4033, opened 2024-10-10
sstable: cleanup readBlock usage

RaduBerinde:
We move the stats and bufferPool arguments into their own
`readBlockEnv` struct. We add specific functions for different block
types.

A functional change is that now we use the buffer pool for the
top-level index block and filter block (this looks like an omission in
the existing code). This changed caused failures in some tests that
were not releasing a pool, which are now fixed.

### DIFF
--- a/sstable/block_property_test.go
+++ b/sstable/block_property_test.go
@@ -902,7 +902,7 @@ func TestBlockProperties(t *testing.T) {
 
 				// Enumerate point key data blocks encoded into the index.
 				if f != nil {
-					indexH, err := r.readIndex(context.Background(), nil, nil, nil)
+					indexH, err := r.readTopLevelIndexBlock(context.Background(), noEnv, noReadHandle)
 					if err != nil {
 						return err.Error()
 					}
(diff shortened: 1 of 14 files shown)


jbowens (approved):
> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4033)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`sstable/reader.go` line 538 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4033#-O8wOH7Y18kg2JoUd-bp:-O8wOH7Y18kg2JoUd-bq:b-oz8o77) ([raw file](https://github.com/cockroachdb/pebble/blob/ceb87d4876e3b71e3ef74cf84c669f560dab9c4e/sstable/reader.go#L538)):*
> ```Go
> 		compressed.Release()
> 		if err != nil {
> 			decompressed.Release()
> ```

Good catch

RaduBerinde:
TFTR!

PR #4050, opened 2024-10-15
colblk: fix UintBuilder inconsistent size bug

RaduBerinde:
#### colblk: add test demonstrating Size bug

UintBuilder has a clever fast path for determining the current
encoding. The fast path is a bit pessimistic when we are
useing the "default" mode: it assumes that there will be at least one
value of 0.

The "slow" path is not pessimistic and thus the results might not be
consistent depending on which path is taken. This is a problem because
we could get the Size, add a row, then get the Size for the previous
number of rows again; now we no longer use the fast path and can get a
different result if all the rows have been set to non-zero values.

This commit adds a test that demonstrates this issue.

#### colblk: fix UintBuilder inconsistent size bug

This commit fixes the UintBuilder bug described in the previous
commit. We make the non-fast-path consistent with the fast path so
that the encoding can't change unexpectedly.

### DIFF
--- a/sstable/colblk/testdata/uints
+++ b/sstable/colblk/testdata/uints
@@ -9,7 +9,7 @@
 init default-zero
 ----
 
-size rows=(100)
+size rows=100
 ----
 Size(100, 0) = 1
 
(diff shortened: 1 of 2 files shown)


jbowens (approved):
:lgtm:

nice find

> Reviewed 2 of 2 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4050#-:-O9GEFlQ5DSbaS2Mp6VF:b-ai8u1h)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #4053, opened 2024-10-15
db: delete compactions output in case of late cancellation

itsbilal:
Previously, if we had a late cancellation of a compaction (ie. right before we were gonna logAndApply), we would end up not marking compaction outputs as obsolete/zombie, and neither would we delete them rightaway. This change updates the compaction running code to also clean up compaction outputs in this case of compaction cancellation.

Fixes cockroachdb/cockroach#132668.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2174,6 +2174,53 @@ func (d *DB) compact(c *compaction, errChannel chan error) {
 	})
 }
 
+// cleanupVersionEdit cleans up any on-disk artifacts that were created
+// for the application of a versionEdit that is no longer going to be applied.
+//
+// d.mu must be held when calling this method.
+func (d *DB) cleanupVersionEdit(ve *versionEdit) {
+	obsoleteFiles := make([]*fileBacking, 0, len(ve.NewFiles))
+	deletedFiles := make(map[base.FileNum]struct{})
+	for key := range ve.DeletedFiles {
+		deletedFiles[key.FileNum] = struct{}{}
+	}
+	for i := range ve.NewFiles {
+		if ve.NewFiles[i].Meta.Virtual {
+			// We handle backing files separately.
+			continue
+		}
+		if _, ok := deletedFiles[ve.NewFiles[i].Meta.FileNum]; ok {
+			// This file is being moved in this ve to a different level.
+			// Don't mark it as obsolete.
+			continue
+		}
+		obsoleteFiles = append(obsoleteFiles, ve.NewFiles[i].Meta.PhysicalMeta().FileBacking)
+	}
+	for i := range ve.CreatedBackingTables {
+		if ve.CreatedBackingTables[i].IsUnused() {
+			obsoleteFiles = append(obsoleteFiles, ve.CreatedBackingTables[i])
+		}
+	}
+	for i := range obsoleteFiles {
+		// Add this file to zombie tables as well, as the versionSet
+		// asserts on whether every obsolete file was at one point
+		// marked zombie.
+		d.mu.versions.zombieTables[obsoleteFiles[i].DiskFileNum] = tableInfo{
+			fileInfo: fileInfo{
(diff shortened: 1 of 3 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4053)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @itsbilal)

___
*[`compaction.go` line 2216 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4053#-O9H5Q9F5EeIJwTgsBNd:-O9H5Q9F5EeIJwTgsBNe:b-biabis) ([raw file](https://github.com/cockroachdb/pebble/blob/dbeaf61568fc44ac2350c046fe8129e7a53312b7/compaction.go#L2216)):*
> ```Go
> 				// by delete-only compactions are virtual sstables, which cannot be cleaned
> 				// up here.
> 				if c.kind != compactionKindDeleteOnly && c.kind != compactionKindMove {
> ```

Can we pull this into a separate function and handle the virtualization case? We don't Ref the file backing until we're applying the version edit to a new version, right? So if we see any sstables in NewFiles backed by a file with a non-zero ref count, we know it's not a new backing file created over the course of this function. Conversely, if a backing file has zero refs it's safe to delete.

Otherwise I worry about this logic being too tied to unspoken invariants that hold today but may not hold tomorrow. (Eg, I can imagine default compactions virtualizing significant portions of input sstables that have no overlap with the other input levels.)
___
*[`compaction.go` line 2784 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4053#-O9H3Pw0DbRHSe6C_bOb:-O9H3Pw0DbRHSe6C_bOc:bz1io14) ([raw file](https://github.com/cockroachdb/pebble/blob/dbeaf61568fc44ac2350c046fe8129e7a53312b7/compaction.go#L2784)):*
> ```Go
> 				FileNum: result.Tables[i].ObjMeta.DiskFileNum,
> 				Path:    path,
> 			})
> ```

Is there any reason we're not enqueueing these tables as obsolete files so that they're subject to pacing and would benefit from emitting `TableDeleted` events without duplicating code?

jbowens (approved):
:lgtm:

> Reviewed 2 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4053#-:-O9M6RZr6PV237Qbjk7H:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @itsbilal)

___
*[`compaction.go` line 2216 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4053#-O9H5Q9F5EeIJwTgsBNd:-O9M5tod76iXc8lRlTu3:bb70w76) ([raw file](https://github.com/cockroachdb/pebble/blob/dbeaf61568fc44ac2350c046fe8129e7a53312b7/compaction.go#L2216)):*

Done.
</blockquote></details>

Nice
___
*[`compaction.go` line 2784 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4053#-O9H3Pw0DbRHSe6C_bOb:-O9M5uvh8HdtxSrsjE6H:byfghwt) ([raw file](https://github.com/cockroachdb/pebble/blob/dbeaf61568fc44ac2350c046fe8129e7a53312b7/compaction.go#L2784)):*

Good point. I think the main reason why we can't directly just reuse the same code as the other (outer) case is because we don't necessarily wanna produce a versionEdit and fileMetadata for sstables that could be partially written or empty. That said I've short-circuited a way to produce file backings and obsolete files out of result.Tables without producing full-on FileMetadatas that gets us to reuse the obsolete files path.
</blockquote></details>

Yeah, I didn't mean reusing the same code as the other section, just that enqueueing them as obsolete files seems more correct than deleting them outright.

itsbilal:
TFTR!

PR #4054, opened 2024-10-15
colblk: improve Uint encoding for very few rows, don't use InitWithDefault for block lengths

RaduBerinde:
#### sstable: fix unaligned slice panic

Fix to RewriteSuffixes code to make sure the block buffers we decode
are 8-byte aligned. Without this fix, the columnar decoder can panic.

#### colblk: improve Uint encoding for very few rows

We currently choose the best uint encoding assuming we have a decent
number of rows. For less than 8 rows, the best encoding might be
different: e.g.  a non-delta encoding with 2 bytes per row can be
smaller than an 8 byte delta base plus 1 byte per row. In rare cases,
this can make blocks *decrease* in size when we add a row which can be
confusing (especially in tests).

This change takes the number of rows into account to get the smallest
possible encoding.

#### colblk: don't use InitWithDefault for block lengths

Having a default value for lengths is unnecessary and can lead to
worse encodings.


jbowens (approved):
:lgtm_strong: 

> Reviewed 3 of 3 files at r1, 11 of 11 files at r2, 13 of 13 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4054#-:-O9GuyA02PmbV7MjmsUB:b-uwkowd)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`-- commits` line 5 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4054#-O9GrnXaDaYBk0bkdKXE):*
Good catch

RaduBerinde:
TFTQR!

PR #4057, opened 2024-10-16
db: support multiple KeySchemas

### LINKED ISSUE #4045: colblk: support multiple key schemas
User code (i.e, Cockroach) may want to modify the colblk.KeySchema used for new sstables. We should provide a mechanism for recording within an sstable which user-provided key schema should be used during decoding. We'll also need to a mechanism for the user to pass all KeySchemas in use in through the Options.

Jira issue: PEBBLE-276

jbowens:
Adapt the configuration of block columnar KeySchemas to allow the user to configure a set of known KeySchemas, in addition to specifying one for use in construction of new sstables. If a user wishes to adapt their key schema, they may introduce a new KeySchema while using the old KeySchema to interpret existing sstables.

In future work we may consider allowing users to partition the LSM, specifying separate KeySchemas for distinct portions of the keyspace.

Close #4045.


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4057#-:-O9LFI5p6zXf0iOnGaj7:bnfp4nl)* status: 0 of 45 files reviewed, all discussions resolved

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4057#-:-O9LdAS7-UR-Zilg_mQj:b-814g8k)* status: 0 of 45 files reviewed, all discussions resolved

RaduBerinde:
*[`options.go` line 1634 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4057#-O9NCi1PDUx8mmMraw1e:-O9NCi1QCcSnVJFyORto:b-s81l7n) ([raw file](https://github.com/cockroachdb/pebble/blob/8989e4e6c1764aea63ef66a37cfcb87f11f6b658/options.go#L1634)):*
> ```Go
> 				}
> 			case "comparer":
> 				switch value {
> ```

The previous code left `o.Comparer` alone if the name doesn't match and there is no `hooks.NewComparer`.. does that matter?

PR #4060, opened 2024-10-16
block: rewrite block flush logic

RaduBerinde:
#### sstable: set w.indexBlockSize in colBlkWriter

This field was never set.

#### block: rewrite block flush logic

Currently the block flush logic is duplicated between the row and
columnar formats, and there are subtle differences between the two
implementations.

This commit moves all the logic to a single `FlushGovernor` struct.
The logic is improved to ignore the in-memory overhead when not using
size classes (which is desirable for testing purposes), and to ignore
size classes when the classes are all above or below the target block
size.


jbowens (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4060#-:-O9M50xy3H_K-4Gv4egS:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4060#-O9M2OXY97OVaLMWyOdJ):*
yikes! Good find
___
*[`sstable/rowblk_writer.go` line 1651 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4060#-O9M3BhV15nQuAWK4pb3:-O9M3BhV15nQuAWK4pb4:bd30ifx) ([raw file](https://github.com/cockroachdb/pebble/blob/940296e0062ea36752d947215bb1c6f93198d6ee/sstable/rowblk_writer.go#L1651)):*
> ```Go
> 		if _, err := w.layout.WritePropertiesBlock(raw.Finish()); err != nil {
> 			return err
> 		}
> ```

good catch
___
*[`sstable/block/flush_governor.go` line 126 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4060#-O9M4aLLAGDETyufAaEb:-O9M4aLLAGDETyufAaEc:b-6az84w) ([raw file](https://github.com/cockroachdb/pebble/blob/940296e0062ea36752d947215bb1c6f93198d6ee/sstable/block/flush_governor.go#L126)):*
> ```Go
> // the block in the cache.
> func (fg *FlushGovernor) wastedSpace(size int) int {
> 	for i := 0; i < len(fg.boundaries); i++ {
> ```

should this `len(fg.boundaries)` be `fg.numBoundaries`?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4060)* status: all files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`sstable/block/flush_governor.go` line 126 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4060#-O9M4aLLAGDETyufAaEb:-O9MDkMg8DmJcJ3D-aet:biz8joc) ([raw file](https://github.com/cockroachdb/pebble/blob/940296e0062ea36752d947215bb1c6f93198d6ee/sstable/block/flush_governor.go#L126)):*

should this `len(fg.boundaries)` be `fg.numBoundaries`?
</blockquote></details>

Nice catch! It happens to work because the extra ones are zero

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4060#-:-O9MjkH2-FRXzoVdvKBb:b-814g8k)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @jbowens)

PR #4061, opened 2024-10-16
sstable: add index separator comparison methods

jbowens:
Extend the IndexBlockIterator interface with two new methods for performing comparisons against the separator at the current iterator position. This is a precursor to decomposing separator keys into prefix and suffix in columnar blocks, as it will allow the colblk index iterator to avoid materializing the separator key unnecessarily.

### DIFF
--- a/sstable/block/block.go
+++ b/sstable/block/block.go
@@ -209,6 +209,17 @@ type IndexBlockIterator interface {
 	// guaranteed to be greater than or equal to every key contained within the
 	// referenced block(s).
 	Separator() []byte
+	// SeparatorLT returns true if the separator at the iterator's current
+	// position is strictly less than the provided key. For some
+	// implementations, it may be more performant to call SeparatorLT rather
+	// than explicitly performing Compare(Separator(), key) < 0.
+	SeparatorLT(key []byte) bool
+	// SeparatorGT returns true if the separator at the iterator's current
+	// position is strictly greater than (or equal, if orEqual=true) the
+	// provided key. For some implementations, it may be more performant to call
+	// SeparatorGT rather than explicitly performing a comparison using the key
+	// returned by Separator.
+	SeparatorGT(key []byte, orEqual bool) bool
 	// BlockHandleWithProperties decodes the block handle with any encoded
 	// properties at the iterator's current position.
 	BlockHandleWithProperties() (HandleWithProperties, error)
--- a/sstable/colblk/index_block.go
+++ b/sstable/colblk/index_block.go
@@ -295,6 +295,19 @@ func (i *IndexIter) Separator() []byte {
 	return i.applyTransforms(key)
 }
 
+// SeparatorLT returns true if the separator at the iterator's current
+// position is strictly less than the provided key.
+func (i *IndexIter) SeparatorLT(key []byte) bool {
+	return i.compare(i.Separator(), key) < 0
+}
+
+// SeparatorGT returns true if the separator at the iterator's current position
+// is strictly greater than (or equal, if orEqual=true) the provided key.
+func (i *IndexIter) SeparatorGT(key []byte, inclusively bool) bool {
+	cmp := i.compare(i.Separator(), key)
+	return cmp > 0 || (cmp == 0 && inclusively)
+}
+
 func (i *IndexIter) applyTransforms(key []byte) []byte {
 	if i.syntheticSuffix.IsSet() {
 		key = key[:i.split(key)]
--- a/sstable/copier.go
+++ b/sstable/copier.go
@@ -255,7 +255,7 @@ func intersectingIndexEntries(
 				}
 			}
 		}
-		if r.Compare(end.UserKey, top.Separator()) <= 0 {
+		if top.SeparatorGT(end.UserKey, true /* inclusively */) {
 			break
 		}
 	}
--- a/sstable/reader_iter_single_lvl.go
+++ b/sstable/reader_iter_single_lvl.go
@@ -409,7 +409,7 @@ func (i *singleLevelIterator[I, PI, D, PD]) initBounds() {
 	}
 	i.blockUpper = i.upper
 	// TODO(radu): this should be >= 0 if blockUpper is inclusive.
-	if i.blockUpper != nil && i.cmp(i.blockUpper, PI(&i.index).Separator()) > 0 {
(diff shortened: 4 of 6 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4061)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`sstable/reader_iter_single_lvl.go` line 833 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4061#-O9N5qEv4397Spe0PoSv:-O9N5qEv4397Spe0PoSw:bvm12ub) ([raw file](https://github.com/cockroachdb/pebble/blob/6917dfc8cdac98a0b40d1c97a093f13fe4591d8f/sstable/reader_iter_single_lvl.go#L833)):*
> ```Go
> 			// below, else we require the separator to be strictly greater than
> 			// upper.
> 			if i.upper != nil && PI(&i.index).IsSeparatorUpperBound(i.upper, !i.endKeyInclusive) {
> ```

Feels like the bound here is `i.upper, endKeyInclusive` and we want to compare that to the separator, rather than comparing `i.upper` to a `separator, !endKeyInclusive` bound.. (it's equivalent, just feels backward).

Intuitively, a successful check that something conforms to a bound feels like it should be the "good" case, not a "bad" early return case.


I would change the new methods to `UpperBoundAppliesToSeparator(b UserKeyBoundary)` and `LoweBoundAppliesToSeparator(key)`. This code would be cleaner to read:

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4061)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 833 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4061#-O9N5qEv4397Spe0PoSv:-O9QT0YzBAXVGsBwzGNj:bu70oc5) ([raw file](https://github.com/cockroachdb/pebble/blob/6917dfc8cdac98a0b40d1c97a093f13fe4591d8f/sstable/reader_iter_single_lvl.go#L833)):*

Feels like the bound here is `i.upper, endKeyInclusive` and we want to compare that to the separator, rather than comparing `i.upper` to a `separator, !endKeyInclusive` bound.. (it's equivalent, just feels backward).

Intuitively, a successful check that something conforms to a bound feels like it should be the "good" case, not a "bad" early return case.


I would change the new methods to `UpperBoundAppliesToSeparator(b UserKeyBoundary)` and `LoweBoundAppliesToSeparator(key)`. This code would be cleaner to read:
</blockquote></details>

Yeah, there's a bit of tension here in that the IndexBlockIterator interface doesn't know what you're comparing against or why. The bulk of usages _are_ against iterator bounds, so I'm okay with inverting the semantics.

itsbilal (approved):
LGTM

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4061)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`sstable/colblk/index_block.go` line 308 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4061#gh-1805280591:-O9QrUSCEfSoQPZ4tM6d:b-mtlf1z) ([raw file](https://github.com/cockroachdb/pebble/blob/ee3d6be8d3b3f1dc7a6b821813661d2a60a3ce52/sstable/colblk/index_block.go#L308)):*

nit: maybe `SeparatorIsLowerBoundFor` and `SeparatorIsUpperBoundFor` ?
</blockquote></details>

See https://reviewable.io/reviews/cockroachdb/pebble/4061#-O9N5qEv4397Spe0PoSv

That would be the original, opposite semantics. I should update these comments that describe the semantics in terms of the separator bounding the provided key, which is maybe what motivated this suggestion

RaduBerinde:
*[`sstable/reader_iter_single_lvl.go` line 833 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4061#-O9N5qEv4397Spe0PoSv:-O9R39NV7tTsZuizlGoq:bnn18gp) ([raw file](https://github.com/cockroachdb/pebble/blob/6917dfc8cdac98a0b40d1c97a093f13fe4591d8f/sstable/reader_iter_single_lvl.go#L833)):*

Yeah, there's a bit of tension here in that the IndexBlockIterator interface doesn't know what you're comparing against or why. The bulk of usages _are_ against iterator bounds, so I'm okay with inverting the semantics.
</blockquote></details>

We could also do SeparatorLT(key, orEqual bool) and SeparatorGT(key, otEqual bool) which would be more generic.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4061)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 833 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4061#-O9N5qEv4397Spe0PoSv:-O9Rpxhs7I-ImYU8YyzT:b-nvhcj2) ([raw file](https://github.com/cockroachdb/pebble/blob/6917dfc8cdac98a0b40d1c97a093f13fe4591d8f/sstable/reader_iter_single_lvl.go#L833)):*

We could also do SeparatorLT(key, orEqual bool) and SeparatorGT(key, otEqual bool) which would be more generic.
</blockquote></details>

that feels right to me

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4061#-:-O9Rvgh-EUMv3qMPpuCC:bnfp4nl)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

jbowens (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4061#-:-O9S8bhQ32tnzkBYoa_8:bz5mair)* status: 0 of 6 files reviewed, 1 unresolved discussion

PR #4065, opened 2024-10-17
db: fix Options.Parse handling of unrecognized comparer

jbowens:
In b3993ff the behavior of Options.Parse when no NewComparer hook is provided changed. Parse began to nil out the Comparer field, clearing any existing Comparer.

Thanks @RaduBerinde for spotting this.

### DIFF
--- a/options.go
+++ b/options.go
@@ -1667,7 +1667,11 @@ func (o *Options) Parse(s string, hooks *ParseHooks) error {
 					}
 				}
 			case "comparer":
-				o.Comparer, err = parseComparer(value)
+				var comparer *Comparer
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):

RaduBerinde:
As a follow-up, could we improve the `Parse` documentation mentioning that we don't clear out any existing options? Also, it's pretty strange that we don't error out if we don't know the comparer. Feels like we could potentially allow opening a store that was created with another comparer? What about making `parseComparer` check against the name of any existing `opts.Comparer` and return that (and error otherwise)?

PR #4066, opened 2024-10-17
sstable: cache DataBlockDecoder alongside data blocks

RaduBerinde:
We reserve a chunk of each allocation in the cache as
`block.Metadata`. We use this to cache the `colblk.DataBlockDecoder`
so we don't have to reinitialize it each time we use the block.

In subsequent changes we will also cache the key seeker and the
`IndexBlockDecoder`.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -831,8 +831,8 @@ func TestMemTableReservation(t *testing.T) {
 		t.Fatalf("expected 2 refs, but found %d", refs)
 	}
 	// Verify the memtable reservation has caused our test block to be evicted.
-	if h := opts.Cache.Get(tmpID, base.DiskFileNum(0), 0); h.Get() != nil {
-		t.Fatalf("expected failure, but found success: %s", h.Get())
+	if h := opts.Cache.Get(tmpID, base.DiskFileNum(0), 0); h.Valid() {
+		t.Fatalf("expected failure, but found success: %#v", h)
 	}
 
 	// Flush the memtable. The memtable reservation should double because old
--- a/internal/cache/clockpro.go
+++ b/internal/cache/clockpro.go
@@ -683,7 +683,9 @@ type Cache struct {
 	}
 }
 
-// TODO
+// ID is a namespace for file numbers. It allows a single Cache to be shared
+// among multiple Pebble instances. NewID can be used to generate a new ID that
+// is unique in the context of this cache.
 type ID uint64
 
 // New creates a new cache of the specified size. Memory for the cache is
--- a/internal/cache/clockpro.go
+++ b/internal/cache/clockpro.go
@@ -59,14 +59,19 @@ type Handle struct {
 	value *Value
 }
 
-// Get returns the value stored in handle.
-func (h Handle) Get() []byte {
-	if h.value != nil {
-		// NB: We don't increment shard.hits in this code path because we only want
-		// to record a hit when the handle is retrieved from the cache.
-		return h.value.buf
-	}
-	return nil
+// Valid returns true if the handle holds a value.
+func (h Handle) Valid() bool {
+	return h.value != nil
+}
+
+// RawBuffer returns the value buffer. Note that this buffer holds the block
+// metadata and the data and should be used through a block.BufferHandle.
+//
+// RawBuffer can only be called if the handle is Valid().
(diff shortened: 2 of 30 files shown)


jbowens (approved):
:lgtm_strong: 

This is dope :fire: 

> Reviewed 29 of 29 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4066#-:-O9QDAg49i9GPWRVlRw_:b-e3teyr)* status: all files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4066#-O9QBlH5C6AgzDREgydF):*
It might be beneficial to do it for keyspan blocks too, since when present the block is read by ~all reads that touch the sstable.
___
*[`internal/cache/clockpro.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4066#-O9Q9x248wpCXLAEeetU:-O9Q9x248wpCXLAEeetV:bkujib5) ([raw file](https://github.com/cockroachdb/pebble/blob/223c492c17d64bdbee1d5c72f78d8cb6f3e7d265/internal/cache/clockpro.go#L68)):*
> ```Go
> 
> // RawBuffer returns the value buffer. Note that this buffer holds the block
> // metadata and the data and should be used through a block.BufferHandle.
> ```

So the contract is that the handle must be `Valid()` if calling `RawBuffer`? I think we should mention it in the comment
___
*[`sstable/block/block.go` line 164 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4066#-O9QBNnk04EMJRbvjQTi:-O9QBNnk04EMJRbvjQTj:brp2hkc) ([raw file](https://github.com/cockroachdb/pebble/blob/223c492c17d64bdbee1d5c72f78d8cb6f3e7d265/sstable/block/block.go#L164)):*
> ```Go
> // Assert that MetadataSize is a multiple of 8. This is necessary to keep the
> // block data buffer aligned.
> const _ uint = -(MetadataSize % 8)
> ```

Clever

RaduBerinde:
TFTR! Yeah I agree on the keyspan blocks, it's why I made sure they have their own loading methods.

I will add a crdbtest benchmark that does random seeks inside an sst and see the before/after.

RaduBerinde (commented):
Added benchmark:
```
RandSeekInSST/v4/single-level-10  1.21µs ± 1%  1.21µs ± 1%    ~     (p=0.161 n=8+7)
RandSeekInSST/v4/two-level-10     2.02µs ± 1%  2.05µs ± 2%  +1.11%  (p=0.013 n=8+8)
RandSeekInSST/v5/single-level-10  1.16µs ± 0%  1.10µs ± 1%  -4.60%  (p=0.000 n=8+8)
RandSeekInSST/v5/two-level-10     1.72µs ± 2%  1.66µs ± 1%  -3.24%  (p=0.000 n=8+8)
```

I think this is in line with what you were seeing in the YCSB profiles. Hoping for another 4-7% gain after caching the index decoder and key seeker.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4066#-:-O9SPOudCTtnlq77Uuoy:b-hy5rjs)* status: 28 of 30 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4066#-O9QBlH5C6AgzDREgydF):*

It might be beneficial to do it for keyspan blocks too, since when present the block is read by ~all reads that touch the sstable.
</blockquote></details>

:thumbsup: 
___
*[`internal/cache/clockpro.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4066#-O9Q9x248wpCXLAEeetU:-O9SP61b0OU12vKBxJ_A:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/d1ec34ab1bf07c4eed9a856488c4efffbba01724/internal/cache/clockpro.go#L68)):*

So the contract is that the handle must be `Valid()` if calling `RawBuffer`? I think we should mention it in the comment
</blockquote></details>

Done.

PR #4067, opened 2024-10-17
options: add EnableDeleteOnlyCompactionExcises Option

itsbilal:
This change adds a dynamic knob in Options to be able to toggle the functionality added in #3926. If the method is undefined or returns false, delete-only compactions only do full-file deletes and not excises.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -257,6 +257,10 @@ type compaction struct {
 	// virtual sstables.
 	deletionHints []deleteCompactionHint
 
+	// exciseEnabled is set to true if this is a compactionKindDeleteOnly and
+	// this compaction is allowed to excise files.
+	exciseEnabled bool
+
 	metrics map[int]*LevelMetrics
 
 	pickerMetrics compactionPickerMetrics
@@ -401,6 +405,7 @@ func newDeleteOnlyCompaction(
 	inputs []compactionLevel,
 	beganAt time.Time,
 	hints []deleteCompactionHint,
+	exciseEnabled bool,
 ) *compaction {
 	c := &compaction{
 		kind:          compactionKindDeleteOnly,
@@ -413,6 +418,7 @@ func newDeleteOnlyCompaction(
 		beganAt:       beganAt,
 		inputs:        inputs,
 		deletionHints: hints,
+		exciseEnabled: exciseEnabled,
 	}
 
 	// Set c.smallest, c.largest.
@@ -1773,11 +1779,15 @@ func (d *DB) maybeScheduleCompactionPicker(
 func (d *DB) tryScheduleDeleteOnlyCompaction() {
 	v := d.mu.versions.currentVersion()
 	snapshots := d.mu.snapshots.toSlice()
-	inputs, resolvedHints, unresolvedHints := checkDeleteCompactionHints(d.cmp, v, d.mu.compact.deletionHints, snapshots, d.FormatMajorVersion())
+	// We need to save the value of exciseEnabled in the compaction itself, as
+	// it can change dynamically between now and when the compaction runs.
+	exciseEnabled := d.FormatMajorVersion() >= FormatVirtualSSTables &&
+		d.opts.Experimental.EnableDeleteOnlyCompactionExcises != nil && d.opts.Experimental.EnableDeleteOnlyCompactionExcises()
+	inputs, resolvedHints, unresolvedHints := checkDeleteCompactionHints(d.cmp, v, d.mu.compact.deletionHints, snapshots, exciseEnabled)
 	d.mu.compact.deletionHints = unresolvedHints
 
 	if len(inputs) > 0 {
-		c := newDeleteOnlyCompaction(d.opts, v, inputs, d.timeNow(), resolvedHints)
+		c := newDeleteOnlyCompaction(d.opts, v, inputs, d.timeNow(), resolvedHints, exciseEnabled)
 		d.mu.compact.compactingCount++
 		d.addInProgressCompaction(c)
 		go d.compact(c, nil)
@@ -1941,7 +1951,7 @@ func (h deleteCompactionHint) String() string {
 }
 
 func (h *deleteCompactionHint) canDeleteOrExcise(
-	cmp Compare, m *fileMetadata, snapshots compact.Snapshots, fmv FormatMajorVersion,
+	cmp Compare, m *fileMetadata, snapshots compact.Snapshots, exciseEnabled bool,
 ) deletionHintOverlap {
 	// The file can only be deleted if all of its keys are older than the
 	// earliest tombstone aggregated into the hint. Note that we use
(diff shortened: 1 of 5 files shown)


jbowens (approved):
:lgtm_strong: 

> Reviewed 4 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4067#-:-O9Q8Wt97LZ7XLaEgPQG:b-uwkowd)* status: 4 of 5 files reviewed, all discussions resolved

itsbilal:
TFTR! Mind taking another look? I made some more changes to make the actual behaviour of having `!exciseEnabled` closer to the pre-#3926 behaviour.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4067)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @itsbilal)

___
*[`compaction.go` line 2580 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4067#-O9QjjRoEcqIZOeRh6Pr:-O9QjjRoEcqIZOeRh6Ps:b-a4dmj) ([raw file](https://github.com/cockroachdb/pebble/blob/504d0a337423a73885c730709d6321ca02923bfb/compaction.go#L2580)):*
> ```Go
> 		// it maps to a virtual file that replaces f, or nil if f got removed
> 		// in its entirety.
> 		if !exciseEnabled {
> ```

Hint generation and hint resolution happen independently. If excise _was_ enabled and now is not, might we attempt to delete a table that was only intended to be excised?
___
*[`compaction.go` line 2694 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4067#-O9QjQQA2qJcHp0psrdJ:-O9QjQQBCOecbfkKCebt:b-e63u2e) ([raw file](https://github.com/cockroachdb/pebble/blob/504d0a337423a73885c730709d6321ca02923bfb/compaction.go#L2694)):*
> ```Go
> 	if c.exciseEnabled {
> 		fragments = fragmentDeleteCompactionHints(d.cmp, c.deletionHints)
> 	}
> ```

if `!c.exciseEnabled` fragments will be empty?

itsbilal (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4067)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`compaction.go` line 2580 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4067#-O9QjjRoEcqIZOeRh6Pr:-O9QkDxb7-QGBJ9vLFYp:bgocxlg) ([raw file](https://github.com/cockroachdb/pebble/blob/504d0a337423a73885c730709d6321ca02923bfb/compaction.go#L2580)):*

Hint generation and hint resolution happen independently. If excise _was_ enabled and now is not, might we attempt to delete a table that was only intended to be excised?
</blockquote></details>

Hint generation happens at table stats time and has nothing to do with what tables the hint(s) apply to, if any. We've always generated hints for every rangedel/rangekeydel overlap. This change couples hint selection (compaction picking) and application (compaction) by propagating the bool.
___
*[`compaction.go` line 2694 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4067#-O9QjQQA2qJcHp0psrdJ:-O9QkUlU1iclVb1HFpct:b3barku) ([raw file](https://github.com/cockroachdb/pebble/blob/504d0a337423a73885c730709d6321ca02923bfb/compaction.go#L2694)):*

if `!c.exciseEnabled` fragments will be empty?
</blockquote></details>

That's right. We just skip fragmentation to avoid the computational burden (and also relax the limit of 10 hints per compaction).

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4067)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @itsbilal)

___
*[`compaction.go` line 2580 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4067#-O9QjjRoEcqIZOeRh6Pr:-O9Ql9Ws9zoS2z1zz2jp:b-jn42bc) ([raw file](https://github.com/cockroachdb/pebble/blob/504d0a337423a73885c730709d6321ca02923bfb/compaction.go#L2580)):*

Hint generation happens at table stats time and has nothing to do with what tables the hint(s) apply to, if any. We've always generated hints for every rangedel/rangekeydel overlap. This change couples hint selection (compaction picking) and application (compaction) by propagating the bool.
</blockquote></details>

I was thinking of the changes that were made to the table stats collector when you adapted it to permit excises. I re-read the code and I think those changes would just potentially cause us to wait longer to resolve the hint, because the associated `fileSmallestSeqNum` may have been made smaller by a file that only partially overlaps (and previously would've been excluded from consideration). I don't think this is a concern.

Relying on altered semantics in two locations based on a shared flag feels too fragile. If `!exciseEnabled` can we just construct fragments that map 1:1 with hints, so we still check down below to ensure this sstable can be deleted outright?

itsbilal:
*[`compaction.go` line 2580 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4067#-O9QjjRoEcqIZOeRh6Pr:-O9QtDkD7N5vCDh-9fT7:b-hr5bkt) ([raw file](https://github.com/cockroachdb/pebble/blob/504d0a337423a73885c730709d6321ca02923bfb/compaction.go#L2580)):*

I was thinking of the changes that were made to the table stats collector when you adapted it to permit excises. I re-read the code and I think those changes would just potentially cause us to wait longer to resolve the hint, because the associated `fileSmallestSeqNum` may have been made smaller by a file that only partially overlaps (and previously would've been excluded from consideration). I don't think this is a concern.

Relying on altered semantics in two locations based on a shared flag feels too fragile. If `!exciseEnabled` can we just construct fragments that map 1:1 with hints, so we still check down below to ensure this sstable can be deleted outright?
</blockquote></details>

We already have a panic if `exciseEnabled` and there are no fragments. I just wanted to avoid the `O(hints x files)` iteration in the compaction as that can really get out of control; the old code didn't propagate hints at all and just blindly deleted every file in the compaction, so I was trying to mimic that here.

However on a closer inspection it seems like that's likely a non-issue because we already do a bit of an `O(hints x files)` iteration _with_ the db mutex held in `checkDeleteCompactionHints` and have always done that; the trick is to just process hints asap. Alongside that logic we could probably do the hint fragmentation and file iteration here no problem, so I've made that change and now the code is a lot closer to what you had approved earlier.

jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4067#-:-O9R0I0IC_O3YKIuAlez:bnfp4nl)* status: 3 of 5 files reviewed, 1 unresolved discussion

itsbilal:
TFTR!

PR #4069, opened 2024-10-18
colblk: separate prefixes and suffixes in index blocks

jbowens:


RaduBerinde (commented):
Very cool! I know it's still a draft but posted some early comments.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4069#-:-O9V-TLBA8aFE5RG1mQS:bsfaxd1)* status: 0 of 22 files reviewed, 6 unresolved discussions (waiting on @jbowens)

___
*[`sstable/colblk/index_block.go` line 104 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4069#-O9UyC7iAmA-Ujusneid:-O9UyC7iAmA-Ujusneie:b-5ibyju) ([raw file](https://github.com/cockroachdb/pebble/blob/0fda49a28424bdf2df6dd70b56d26569aaac687c/sstable/colblk/index_block.go#L104)):*
> ```Go
> 	// Decompose the separator into prefix and suffix.
> 	s := w.split(separator)
> 	commonPrefixLen := s
> ```

[nit] feels like a strange value to pass to the first one, shouldn't 0 also work? It doesn't look like we care about its value for the first key.
___
*[`sstable/colblk/index_block.go` line 119 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4069#-O9UymjL7NNrFHd-8LBh:-O9UymjL7NNrFHd-8LBi:b8bqjrq) ([raw file](https://github.com/cockroachdb/pebble/blob/0fda49a28424bdf2df6dd70b56d26569aaac687c/sstable/colblk/index_block.go#L119)):*
> ```Go
> 
> // UnsafeSeparator returns the separator of the i'th entry. If r is the number
> // of rows, it is required that r-2 <= i < r.
> ```

[nit] `i=r-2 or i=r-1` is more clea
___
*[`sstable/colblk/index_block.go` line 121 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4069#-O9Uyftp-nJ60k5Q6wmV:-O9Uyftp-nJ60k5Q6wmW:b-9v1gol) ([raw file](https://github.com/cockroachdb/pebble/blob/0fda49a28424bdf2df6dd70b56d26569aaac687c/sstable/colblk/index_block.go#L121)):*
> ```Go
> // of rows, it is required that r-2 <= i < r.
> func (w *IndexBlockWriter) UnsafeSeparator(i int) []byte {
> 	if i < w.rows-2 || i >= w.rows {
> ```

[nit] we already have this check in `UnsafeGet`
___
*[`sstable/colblk/index_block.go` line 382 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4069#-O9UzIlRBffHNCvzTLoy:-O9UzIlRBffHNCvzTLoz:b37js3) ([raw file](https://github.com/cockroachdb/pebble/blob/0fda49a28424bdf2df6dd70b56d26569aaac687c/sstable/colblk/index_block.go#L382)):*
> ```Go
> 				return i.row < i.n
> 			}
> 			i.row = i.d.bd.Rows()
> ```

`i.row = n; return false`
___
*[`sstable/colblk/index_block.go` line 411 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4069#-O9V-1apBMZ_JyWy0Jx8:-O9V-1apBMZ_JyWy0Jx9:b-bnqvo2) ([raw file](https://github.com/cockroachdb/pebble/blob/0fda49a28424bdf2df6dd70b56d26569aaac687c/sstable/colblk/index_block.go#L411)):*
> ```Go
> 	// Fall back to a slow scan forward.
> 	//
> 	// TODO(jackson): This can be improved by adding a PrefixBytes.NextUniqueKey
> ```

Maybe we should always include a `prefixChanged` bitmap inside the prefix bytes?
___
*[`sstable/colblk/index_block.go` line 416 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4069#-O9Uzgxx1k5m2b98nqhc:-O9Uzgxx1k5m2b98nqhd:bl0wnzr) ([raw file](https://github.com/cockroachdb/pebble/blob/0fda49a28424bdf2df6dd70b56d26569aaac687c/sstable/colblk/index_block.go#L416)):*
> ```Go
> 	// with offset indexes, so it's deferred for now.
> 	for i.row++; i.row < i.d.bd.Rows(); i.row++ {
> 		if i.SeparatorGT(key, true /* orEqual */) {
> ```

We want to compare just the suffix, no? `SeparatorGT` checks the prefix again

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4069)* status: 0 of 54 files reviewed, 6 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/colblk/index_block.go` line 416 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4069#-O9Uzgxx1k5m2b98nqhc:-O9VICdC3IAwWQjr6NDi:bndehfk) ([raw file](https://github.com/cockroachdb/pebble/blob/0fda49a28424bdf2df6dd70b56d26569aaac687c/sstable/colblk/index_block.go#L416)):*

We want to compare just the suffix, no? `SeparatorGT` checks the prefix again
</blockquote></details>

Unfortunately as long as we don't know when the prefix changes, we have to do a prefix comparison too

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4069)* status: 0 of 54 files reviewed, 6 unresolved discussions (waiting on @jbowens)

___
*[`sstable/colblk/index_block.go` line 416 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4069#-O9Uzgxx1k5m2b98nqhc:-O9V_O7JBg742kJ4IFYl:bkais1v) ([raw file](https://github.com/cockroachdb/pebble/blob/0fda49a28424bdf2df6dd70b56d26569aaac687c/sstable/colblk/index_block.go#L416)):*

Unfortunately as long as we don't know when the prefix changes, we have to do a prefix comparison too
</blockquote></details>

Yeah, sorry, I meant to delete that comment

PR #4070, opened 2024-10-18
db: split CompareSuffixes into Compare{Point,Range}Suffixes

jbowens:
Due to historical reasons in the context of cockroach (see cockroachdb/cockroach#132849), Cockroach has diverged in how it compares the suffixes of range keys and point keys. There are circumstances where Pebble may want to compare point suffixes in isolation. This commit replaces the Comparer's CompareSuffixes func with two new funcs: CompareRangeSuffixes and ComparePointSuffixes.


RaduBerinde (commented):
:lgtm_strong:   Nice, this makes things more clear!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4070#-:-O9UwzB13H2ZNSwcqER_:b-aa3hlv)* status: 0 of 33 files reviewed, 3 unresolved discussions (waiting on @aadityasondhi and @jbowens)

___
*[`internal/base/comparer.go` line 458 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4070#-O9UvvxS3sj1t7XihfzV:-O9UvvxS3sj1t7XihfzW:bietynn) ([raw file](https://github.com/cockroachdb/pebble/blob/0adb4600beec90ab0031016676b43aee53edb915/internal/base/comparer.go#L458)):*
> ```Go
> 	}
> 	// Verify the suffixes have a consistent ordering.
> 	slices.SortFunc(suffixes, c.ComparePointSuffixes)
> ```

How about we skip this `Sort`, asserting that the stricter range ordering is also an ordering for point.
___
*[`internal/base/comparer.go` line 483 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4070#-O9UvY5uA6tdyo-jj8mH:-O9UvY5uA6tdyo-jj8mI:b-wztqu3) ([raw file](https://github.com/cockroachdb/pebble/blob/0adb4600beec90ab0031016676b43aee53edb915/internal/base/comparer.go#L483)):*
> ```Go
> 			b := slices.Concat(p, suffixes[i])
> 			// Make sure the Compare function agrees with ComparePointSuffixes, with the
> 			// caveat that it can consider some consecutive suffixes to be equal.
> ```

[nit] remove the caveat
___
*[`internal/base/comparer.go` line 507 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4070#-O9UvgluE9K-EVKxsWWN:-O9UvgluE9K-EVKxsWWO:b-4t3bjd) ([raw file](https://github.com/cockroachdb/pebble/blob/0adb4600beec90ab0031016676b43aee53edb915/internal/base/comparer.go#L507)):*
> ```Go
> 						}
> 					} else {
> 						// We allow Compare to return equality for suffixes even when
> ```

We should use `ComparePointSuffixes` and it should return the same result. We can also cross-check the `CompareRangeSuffixes` in that they should agree unless the result is 0.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4070#-:-O9WTOrX6Le-L_gVEASe:bnfp4nl)* status: 0 of 33 files reviewed, all discussions resolved (waiting on @aadityasondhi)

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4070#-:-O9b4Gjz33SWSvYb5Vp-:b-814g8k)* status: 0 of 33 files reviewed, all discussions resolved

PR #4073, opened 2024-10-18
crdbtest: fix MaterializeUserKey

jbowens:
The KeySeeker implementation accidentally materialized four bytes of the wall time as the logical time when encoding a MVCC timestamp with a non-zero logical time.

### DIFF
--- a/internal/crdbtest/crdbtest.go
+++ b/internal/crdbtest/crdbtest.go
@@ -833,10 +833,10 @@ func (ks *cockroachKeySeeker) MaterializeUserKey(
 	}
 
 	// Inline binary.BigEndian.PutUint32.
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):
There is a test that seeks to all keys in the  block, does that not check that the key matches?

jbowens (commented):
Looks like we were leaving PercentLogical as zero so not exercising the logical timestamps. Fixed

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4073#-:-O9W6FIc63TSL6Teu3hy:b-bh4168)* status: 0 of 2 files reviewed, all discussions resolved

RaduBerinde (approved):

PR #4084, opened 2024-10-21
colblk: add a KeySchema-specific header

RaduBerinde:
We add a fixed-size KeySchema-specific header to the beginning of the
custom block header. A KeySchema implementation can store there
information and retrieve it when initializing the key seeker.

We use this to add a byte to cockroach blocks which indicates what
types of suffixes are contained in the block (mvcc, empty, non-mvcc).
This will be used for optimized fast paths for the common cases.

### DIFF
--- a/internal/crdbtest/crdb.go
+++ b/internal/crdbtest/crdb.go
@@ -14,6 +14,7 @@ import (
 	"io"
 	"math/rand/v2"
 	"slices"
+	"strings"
 	"time"
 	"unsafe"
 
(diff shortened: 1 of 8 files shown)


jbowens (approved):
:lgtm:

nice

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4084#-:-O9jZa0K6MFPjUsyNU5w:b-jwumhe)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
I was wondering - should we encode the size of the keyschema header as well? Seems fragile in that if we change `KeySchema.HeaderSize` we will end up reading garbage.

jbowens:
Yeah I think that's a good idea

PR #4085, opened 2024-10-21
sstable: pass KeySchema by pointer

RaduBerinde:
We pass and store `KeySchema` by value in many places; this struct is
~64 bytes in size, so it adds a bit of size to various structs.

Switch to using `*KeySchema`, which is also consistent with how we use
the `Comparer`.

RabdSeekInSST benchmark is unaffected:
```
name                              old time/op  new time/op  delta
RandSeekInSST/v5/single-level-10   505ns ± 2%   504ns ± 0%   ~     (p=0.718 n=7+8)
RandSeekInSST/v5/two-level-10     1.02µs ± 3%  1.01µs ± 2%   ~     (p=0.591 n=8+8)
```

### DIFF
--- a/cmd/pebble/db.go
+++ b/cmd/pebble/db.go
@@ -63,7 +63,7 @@ func newPebbleDB(dir string) DB {
 		DisableWAL:                  disableWAL,
 		FormatMajorVersion:          pebble.FormatNewest,
 		KeySchema:                   crdbtest.KeySchema.Name,
-		KeySchemas:                  sstable.MakeKeySchemas(crdbtest.KeySchema),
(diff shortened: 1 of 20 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4085#-:-O9jeaw96O78cbC8mZaK:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #4100, opened 2024-10-22
colblk: don't require the header size for KeySchemaHeader

RaduBerinde:
`KeySchemaHeader` returns the header that was written instead of
relying on the caller to provide the correct size. This makes things
less error-prone and also allows for extending the header while
allowing backward compatibility.


RaduBerinde:
TFTR!

PR #4101, opened 2024-10-23
crdb: fix untyped version seek, add tests

RaduBerinde:
#### crdbtest: add tests

This change adds tests for block encoding, suffix types, and key seek.
It also fixes some bugs related to the suffix types.

#### crdb: fix untyped version seek

The comparison when searching for an untyped version was in the wrong
direction (suffixes are in reverse order).


jbowens (approved):
:lgtm:

good catch

> Reviewed 7 of 7 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4101#-:-O9uvgImD09vvnQeYt_Z:b-93r36f)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

PR #4102, opened 2024-10-24
crdb: better seek, fix untyped versions bug

RaduBerinde:
The untyped versions should not have the terminator byte but we were
writing it out.

The new code relies on the `suffixTypes` value for a fast path that
assumes everything is MVCC. The rest of the cases are handled by a
general path that compares suffixes by byte. The performance is
slightly better than the old code and we now support all cases.

```
name                                                                                                                 old time/op    new time/op    delta
CockroachDataBlockIterShort/AlphaLen=8,Prefix=8,Shared=4,KeysPerPrefix=4,Logical=10,ValueLen=8/Next-10                 11.5ns ± 3%    11.2ns ± 0%  -1.91%  (p=0.001 n=7+7)
CockroachDataBlockIterShort/AlphaLen=8,Prefix=8,Shared=4,KeysPerPrefix=4,Logical=10,ValueLen=8/SeekGE-10                119ns ± 2%     116ns ± 0%  -2.24%  (p=0.000 n=8+8)
CockroachDataBlockIterShort/AlphaLen=8,Prefix=8,Shared=4,KeysPerPrefix=4,Logical=10,ValueLen=8/SeekGELatest-10          106ns ± 0%     105ns ± 0%  -0.97%  (p=0.000 n=7+8)
CockroachDataBlockIterShort/AlphaLen=8,Prefix=128,Shared=64,KeysPerPrefix=4,Logical=10,ValueLen=128/Next-10            10.7ns ± 1%    10.6ns ± 0%  -1.26%  (p=0.000 n=8+8)
CockroachDataBlockIterShort/AlphaLen=8,Prefix=128,Shared=64,KeysPerPrefix=4,Logical=10,ValueLen=128/SeekGE-10          93.9ns ± 1%    92.4ns ± 0%  -1.56%  (p=0.000 n=8+8)
CockroachDataBlockIterShort/AlphaLen=8,Prefix=128,Shared=64,KeysPerPrefix=4,Logical=10,ValueLen=128/SeekGELatest-10    85.8ns ± 1%    83.3ns ± 0%  -2.96%  (p=0.000 n=8+7)
```


RaduBerinde:
I plan to address the TODO to add tests with mixed suffix types in a separate PR (I'll also add some randomized tests).

jbowens (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4102#-:-O9z3OIb0dF-La-Sh5cs:b-uwkowd)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/crdbtest/crdb.go` line 794 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4102#-O9z2peM4aHd3KuYbFFi:-O9z2peM4aHd3KuYbFFj:bwiowm1) ([raw file](https://github.com/cockroachdb/pebble/blob/bb1d59c8565f7ae276a14515a65f0f40e73b94a1/internal/crdbtest/crdb.go#L794)):*
> ```Go
> 			// Note: this path is very unlikely in practice: it should only happen in
> 			// a block that has both non-MVCC suffixes and MVCC suffixes. At most one
> 			// such block should exist on each entire level.
> ```

i'm not sure this is strictly true, because we might use MVCC keys within the local range keyspace too. I'm not sure. But regardless, it should be relatively unlikely

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4102#-:-O9zRBBNDcvMk9XBOudt:b-814g8k)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`internal/crdbtest/crdb.go` line 794 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4102#-O9z2peM4aHd3KuYbFFi:-O9zR3rB1T1mswlCydeL:b9yhnu6) ([raw file](https://github.com/cockroachdb/pebble/blob/bb1d59c8565f7ae276a14515a65f0f40e73b94a1/internal/crdbtest/crdb.go#L794)):*

i'm not sure this is strictly true, because we might use MVCC keys within the local range keyspace too. I'm not sure. But regardless, it should be relatively unlikely
</blockquote></details>

Updated the comment. I also added some `gcassert:inline` to make sure the Put calls are inlined. I also separated the empty key suffix case to only check it for the first key, and only if the block contains empty suffixes.

PR #4104, opened 2024-10-24
metamorphic: tweak L0StopWritesThreshold in case of small files

itsbilal:
Small sstables lead to slower compactions as more time is spent in compaction picking and opening/closing files and iterators than iterating over sstables. A combination of a tiny target file size and a low L0 stop writes threshold can cause a deep write stall that makes some operations time out. This change tweaks L0StopWritesThreshold to let us wait longer in those instances without timing out.

Partly seen in #3947.

### DIFF
--- a/metamorphic/options.go
+++ b/metamorphic/options.go
@@ -519,6 +519,8 @@ func standardOptions() []*TestOptions {
   index_block_size=1
 `,
 		15: `
+[Options]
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):

itsbilal:
TFTR!

PR #4105, opened 2024-10-24
sstable: assert KeyComparison correctness under invariants

jbowens:
When the invariants build tag is enabled, assert that the KeyComparison returned by the columnar data block writer is accurate by computing the KeyComparison from a buffered user key.

Informs #4103.

### DIFF
--- a/internal/invariants/off.go
+++ b/internal/invariants/off.go
@@ -27,3 +27,18 @@ func (d *CloseChecker) AssertClosed() {}
 
 // AssertNotClosed panics in invariant builds if Close was called.
 func (d *CloseChecker) AssertNotClosed() {}
+
+// Value is a generic container for a value that should only exist in invariant
+// builds. In non-invariant builds, storing a value is a no-op, retrieving a
+// value returns the type parameter's zero value, and the Value struct takes up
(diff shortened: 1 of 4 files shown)


itsbilal (approved):
LGTM

jbowens (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4105#-:-OA-xGSF8UbL7Sja_HNX:bz5mair)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/invariants/on.go` line 48 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4105#-OA-J1ct9GZoISWhlHnR:-OA-xE9aEmwcismyxIWS:b-e1a4jo) ([raw file](https://github.com/cockroachdb/pebble/blob/9b8bb231e6af31e1efdcc77c8276bee9b0540033/internal/invariants/on.go#L48)):*

[nit] This can be a generic `Value<T>`, there are definitely more potential uses. Get can return the zero value in the off case.
</blockquote></details>

Good idea; done

PR #4108, opened 2024-10-25
colblk: add DataBlockDecoder.Validate

jbowens:
Add a Validate method to DataBlockDecoder that validates invariants of the decoded data block.

Informs #4103.

### DIFF
--- a/sstable/colblk_writer.go
+++ b/sstable/colblk_writer.go
@@ -635,6 +635,14 @@ func (w *RawColumnWriter) enqueueDataBlock(
 	// it's unnecessary.
 	w.meta.SetLargestPointKey(lastKey.Clone())
 
+	if invariants.Enabled {
+		var dec colblk.DataBlockDecoder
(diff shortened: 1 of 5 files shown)


sstable/colblk_writer.go
```diff
@@ -635,6 +635,14 @@ func (w *RawColumnWriter) enqueueDataBlock(
 	// it's unnecessary.
 	w.meta.SetLargestPointKey(lastKey.Clone())
 
+	if invariants.Enabled {
+		var dec colblk.DataBlockDecoder
+		dec.Init(w.opts.KeySchema, serializedBlock)
```
RaduBerinde, line 640: Doesn't this initialize a KeySeeker that we can use?

RaduBerinde: Ah, no, I was thinking of BlockIter.Init

RaduBerinde (approved):

PR #4113, opened 2024-10-25
metamorphic: improve SeekPrefixGE key selection

jbowens:
Previously the generator and key manager maintained a concept of a 'meta timestamp' in an attempt to provide a chronology of a key history. This mechanism was flawed, failing to properly account for the fact that a batch's history applies atomically at the moment the batch is committed. It also didn't account for the fact that the key manager destroys history beneath deletes (to avoid having to maintain too much in-memory state), which would wipe out a record of the DB state at the time an iterator was opened.

This commit replaces this mechanism with a simpler approach of making a copy of the set of a reader's visible keys at the moment the iterator is created.


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4113#-:-OAA3xU08ZX2N3lKWT81:bnfp4nl)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`metamorphic/key_manager.go` line 613 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4113#-OAA46UG2-SoxugGPybi:-OAA46UG2-SoxugGPybj:bqv7fh0) ([raw file](https://github.com/cockroachdb/pebble/blob/fcf5b7008386aae9d9cbbfa4b16602aa09519a6f/metamorphic/key_manager.go#L613)):*
> ```Go
> 		// TODO(jackson): If s.writerID is a batch, we may not know the set of
> 		// all keys that WILL exist by the time the batch is committed. This
> 		// means the delete range history is incomplete. Fix this.
> ```

How do you envision fixing this? History could separately maintain a point history and a (fragmented) range history, perhaps?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4113)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`metamorphic/key_manager.go` line 613 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4113#-OAA46UG2-SoxugGPybi:-OAIYxULCP8QRQKqTnLn:btlefek) ([raw file](https://github.com/cockroachdb/pebble/blob/fcf5b7008386aae9d9cbbfa4b16602aa09519a6f/metamorphic/key_manager.go#L613)):*

How do you envision fixing this? History could separately maintain a point history and a (fragmented) range history, perhaps?
</blockquote></details>

Yeah, I have a wip branch that maintains the set of range-deleted spans in the writer

jbowens:
TFTR!

PR #4116, opened 2024-10-26
crdbtest: fix empty suffix bug, add lock and empty suffixes to randomized test

RaduBerinde:
#### crdbtest: fix incorrect assumption about empty suffixes

The SeekGE code assumes that among all the keys with the same prefix,
only the first one can have an empty suffix. But this is incorrect
because we can have multiple occurrences of the same key.

#### crdbtest: add lock and empty suffixes to randomized test


jbowens (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4116#-:-OAI_lvp0wpxcPZSIDqg:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #4117, opened 2024-10-27
invariants, cache: reorganize and simplify tag-dependent code

RaduBerinde:
#### invariants: reorganize code

We add a `buildtags` package which defines constants for the `race`,
`cgo`, `tracing` tags and we simplify the `invariants` package.

#### cache: simplify value code

We have three tag-dependent file variants for allocating and freeing
`Value`.  This commit greatly simplifies this code using constants
instead of build tags and consolidating the non-cgo and the invariant
implementations.

#### cache: simplify entry code

Consolidate the entry allocation paths using constants instead of
build tags.


sumeerbhola (approved):
:lgtm: 

> Reviewed 14 of 14 files at r1, 4 of 4 files at r2, 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4117#-:-OBb8GeKExIfLer2Ibyj:bg7igkv)* status: all files reviewed, 5 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/invariants/invariants.go` line 18 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4117#-OBaxJO-7RCObKfj5M38:-OBaxJO-7RCObKfj5M39:bscwfti) ([raw file](https://github.com/cockroachdb/pebble/blob/f74f4d4758fec871e40ad2aef8299cb2feb2e618/internal/invariants/invariants.go#L18)):*
> ```Go
> 
> // Sometimes returns true percent% of the time if we were built with the
> // "invariants" of "race" build tags
> ```

nit: s/of/or

nit: missing period.

It is probably worth reiterating here in a comment that invariant || race tags <=> `invariants.Enabled`.
___
*[`internal/invariants/invariants.go` line 34 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4117#-OBazVM5AAShRq-grafd:-OBazVM5AAShRq-grafe:b-bokexx) ([raw file](https://github.com/cockroachdb/pebble/blob/f74f4d4758fec871e40ad2aef8299cb2feb2e618/internal/invariants/invariants.go#L34)):*
> ```Go
> //
> // We exclude race builds because we historically ran into some race detector
> // bugs related to finalizers.
> ```

Presumably this inlines when `!UseFinalizers`, so there is zero overhead?
___
*[`internal/cache/value.go` line 50 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4117#-OBb153208W7L1q6_wJv:-OBb15335Ql9ft0L_Cj1:b-vqvk3x) ([raw file](https://github.com/cockroachdb/pebble/blob/6aa9b9d4758e8657db14ce7abeff5f2bee11c6a2/internal/cache/value.go#L50)):*
> ```Go
> 
> 	if valueEntryGoAllocated {
> 		b := manual.New(manual.BlockCacheData, n)
> ```

It is a bit confusing to see `manual.New` here when it is possible that we are here due to `!buildtags.Cgo`. Consider adding a comment that `manual.New` will use the Go allocator in that case.
I suppose what we were doing in `value_normal.go` by allocating the buf explicitly using `make([]byte, n)` was unnecessary since `manual.New` would have done it for us?
___
*[`internal/cache/entry.go` line 171 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4117#-OBb4Zmt5NUPpBRfs3PO:-OBb4Zmt5NUPpBRfs3PP:b-3j16qo) ([raw file](https://github.com/cockroachdb/pebble/blob/fc3bd37dabe03bef589cb9444e03957ce524f755/internal/cache/entry.go#L171)):*
> ```Go
> // We also use the Go allocator in race mode because the normal path relies on a
> // finalizer (and historically there have been some finalizer-related bugs in
> // the race detector).
> ```

nit: the old comment was more specific
```
	// Avoid using runtime.SetFinalizer in race builds as finalizers tickle a bug
	// in the Go race detector in go1.15 and earlier versions. This requires that
	// entries are Go allocated rather than manually allocated.
```
___
*[`internal/cache/entry.go` line 177 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4117#-OBb6_9UErrCmn3trUq0:-OBb6_9UErrCmn3trUq1:bcxt3u0) ([raw file](https://github.com/cockroachdb/pebble/blob/fc3bd37dabe03bef589cb9444e03957ce524f755/internal/cache/entry.go#L177)):*
> ```Go
> 
> func entryAllocNew() *entry {
> 	if invariants.UseFinalizers {
> ```

previously we were doing this in `entry_invariants.go` under the condition `//go:build (invariants && !race) || (tracing && !race)`.

Now this is 
```
const UseFinalizers = !RaceEnabled && (Enabled || buildtags.Tracing)
```
where `Enabled` is true when  `//go:build invariants || race`

So `!race && (invariants || race || tracing)` which simplifies to `!race && (invariants || tracing)`.

So seems the same. My head hurts though -- I realize it can't be avoided.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4117#-:-OBexzml1-7KQk5Frh5l:b-814g8k)* status: all files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/cache/entry.go` line 177 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4117#-OBb6_9UErrCmn3trUq0:-OBevCLy-Gl3aWulmF-7:b-8csax3) ([raw file](https://github.com/cockroachdb/pebble/blob/fc3bd37dabe03bef589cb9444e03957ce524f755/internal/cache/entry.go#L177)):*

previously we were doing this in `entry_invariants.go` under the condition `//go:build (invariants && !race) || (tracing && !race)`.

Now this is 
```
const UseFinalizers = !RaceEnabled && (Enabled || buildtags.Tracing)
```
where `Enabled` is true when  `//go:build invariants || race`

So `!race && (invariants || race || tracing)` which simplifies to `!race && (invariants || tracing)`.

So seems the same. My head hurts though -- I realize it can't be avoided.
</blockquote></details>

I added `buildtags.Invariants` and made this `!buildtags.Race && (buildtags.Invariants || buildtags.Tracing)`, hopefully this is a bit more clear.
___
*[`internal/cache/value.go` line 50 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4117#-OBb153208W7L1q6_wJv:-OBexH2N8D4RFmFmXJw0:b-wymc4r) ([raw file](https://github.com/cockroachdb/pebble/blob/6aa9b9d4758e8657db14ce7abeff5f2bee11c6a2/internal/cache/value.go#L50)):*

It is a bit confusing to see `manual.New` here when it is possible that we are here due to `!buildtags.Cgo`. Consider adding a comment that `manual.New` will use the Go allocator in that case.
I suppose what we were doing in `value_normal.go` by allocating the buf explicitly using `make([]byte, n)` was unnecessary since `manual.New` would have done it for us?
</blockquote></details>

Correct. Done.
___
*[`internal/invariants/invariants.go` line 34 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4117#-OBazVM5AAShRq-grafd:-OBewg4E4ZPjiYTonph_:bkekeur) ([raw file](https://github.com/cockroachdb/pebble/blob/f74f4d4758fec871e40ad2aef8299cb2feb2e618/internal/invariants/invariants.go#L34)):*

Presumably this inlines when `!UseFinalizers`, so there is zero overhead?
</blockquote></details>

Yes, it should become an empty function. Added a comment.

sumeerbhola (approved):
:lgtm: 

> Reviewed 6 of 15 files at r4, 2 of 9 files at r6.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4117#-:-OBgeLs908LwSNSlBq6_:bg7igkv)* status: 18 of 25 files reviewed, 3 unresolved discussions (waiting on @jbowens)

PR #4118, opened 2024-10-28
release-24.2: db: delete compactions output in case of late cancellation

itsbilal:
24.2 backport of #4053.

----

Previously, if we had a late cancellation of a compaction (ie. right before we were gonna logAndApply), we would end up not marking compaction outputs as obsolete/zombie, and neither would we delete them rightaway. This change updates the compaction running code to also clean up compaction outputs in this case of compaction cancellation.

Fixes cockroachdb/cockroach#132668.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2075,6 +2075,53 @@ func (d *DB) compact(c *compaction, errChannel chan error) {
 	})
 }
 
+// cleanupVersionEdit cleans up any on-disk artifacts that were created
+// for the application of a versionEdit that is no longer going to be applied.
+//
+// d.mu must be held when calling this method.
+func (d *DB) cleanupVersionEdit(ve *versionEdit) {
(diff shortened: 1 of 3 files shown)


itsbilal:
Interestingly enough, this bug is not present on 24.1 because `runCompaction` doesn't do any SSTable cleanup and we let `compact1` do it all. Which is to say this bug crept in with the compaction runner refactor in 24.2.

jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4118#-:-OAIYYlX8LPi_0eGaqEz:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aadityasondhi)

itsbilal:
TFTR!

PR #4123, opened 2024-10-30
sstable: lazily allocate ValueFetcher

RaduBerinde:
Currently the type `valueBlockReader` implements both
`block.GetLazyValueForPrefixAndValueHandler` and `base.ValueFetcher`.
The lifetime of the `ValueFetcher` needs to extend beyond that
of the iterator.

This commit splits it into two objects - `valueBlockReader` continues
to implement `block.GetLazyValueForPrefixAndValueHandler` but the new
`valueBlockFetcher` now implements `base.ValueFetcher`. The
`valueBlockReader` lazily allocates the `valueBlockFetcher` the first
time it is asked to produce a `LazyValue`. The result is that we
should avoid the allocation if we are never positioned on a KV with
non-in-place value (which is the norm when we get the latest version
of a single key).

```
name                              old time/op    new time/op    delta
RandSeekInSST/v4/single-level-10     691ns ± 2%     668ns ± 3%    -3.38%  (p=0.000 n=8+7)
RandSeekInSST/v4/two-level-10       1.57µs ± 2%    1.54µs ± 4%      ~     (p=0.067 n=7+8)
RandSeekInSST/v5/single-level-10     479ns ± 1%     425ns ± 1%   -11.28%  (p=0.001 n=7+7)
RandSeekInSST/v5/two-level-10        967ns ± 4%     868ns ± 4%   -10.25%  (p=0.001 n=7+7)

name                              old alloc/op   new alloc/op   delta
RandSeekInSST/v4/single-level-10      288B ± 0%      122B ± 2%   -57.81%  (p=0.000 n=8+8)
RandSeekInSST/v4/two-level-10         288B ± 0%      122B ± 1%   -57.74%  (p=0.000 n=8+7)
RandSeekInSST/v5/single-level-10      288B ± 0%       11B ± 0%   -96.18%  (p=0.000 n=8+7)
RandSeekInSST/v5/two-level-10         288B ± 0%       11B ± 0%   -96.18%  (p=0.000 n=8+8)

name                              old allocs/op  new allocs/op  delta
RandSeekInSST/v4/single-level-10      1.00 ± 0%      0.00       -100.00%  (p=0.000 n=8+8)
RandSeekInSST/v4/two-level-10         1.00 ± 0%      0.00       -100.00%  (p=0.000 n=8+8)
RandSeekInSST/v5/single-level-10      1.00 ± 0%      0.00       -100.00%  (p=0.000 n=8+8)
RandSeekInSST/v5/two-level-10         1.00 ± 0%      0.00       -100.00%  (p=0.000 n=8+8)
```

### DIFF
--- a/sstable/block/kv.go
+++ b/sstable/block/kv.go
@@ -72,5 +72,12 @@ func InPlaceValuePrefix(setHasSameKeyPrefix bool) ValuePrefix {
 // GetLazyValueForPrefixAndValueHandler is an interface for getting a LazyValue
 // from a value prefix and value.
 type GetLazyValueForPrefixAndValueHandler interface {
+	// GetLazyValueForPrefixAndValueHandle returns a LazyValue for the given value
+	// prefix and value.
+	//
+	// The result is only valid until the next call to
+	// GetLazyValueForPrefixAndValueHandle. Use LazyValue.Clone if the lifetime of
+	// the LazyValue needs to be extended. For more details, see the "memory
+	// management" comment where LazyValue is declared.
 	GetLazyValueForPrefixAndValueHandle(handle []byte) base.LazyValue
 }
--- a/sstable/reader_iter_single_lvl.go
+++ b/sstable/reader_iter_single_lvl.go
@@ -64,7 +64,7 @@ type singleLevelIterator[I any, PI indexBlockIterator[I], D any, PD dataBlockIte
 	// loading. It may not actually have loaded the block, due to an error or
 	// because it was considered irrelevant.
 	dataBH   block.Handle
-	vbReader *valueBlockReader
+	vbReader valueBlockReader
 	// vbRH is the read handle for value blocks, which are in a different
 	// part of the sstable than data blocks.
 	vbRH         objstorage.ReadHandle
@@ -228,22 +228,13 @@ func newColumnBlockSingleLevelIterator(
 	)
 	var getLazyValuer block.GetLazyValueForPrefixAndValueHandler
 	if r.Properties.NumValueBlocks > 0 {
-		// NB: we cannot avoid this ~248 byte allocation, since valueBlockReader
-		// can outlive the singleLevelIterator due to be being embedded in a
-		// LazyValue. This consumes ~2% in microbenchmark CPU profiles, but we
-		// should only optimize this if it shows up as significant in end-to-end
-		// CockroachDB benchmarks, since it is tricky to do so. One possibility
-		// is that if many sstable iterators only get positioned at latest
-		// versions of keys, and therefore never expose a LazyValue that is
-		// separated to their callers, they can put this valueBlockReader into a
-		// sync.Pool.
-		i.vbReader = &valueBlockReader{
+		i.vbReader = valueBlockReader{
 			bpOpen: i,
 			rp:     rp,
 			vbih:   r.valueBIH,
 			stats:  stats,
 		}
-		getLazyValuer = i.vbReader
+		getLazyValuer = &i.vbReader
 		i.vbRH = objstorageprovider.UsePreallocatedReadHandle(r.readable, objstorage.NoReadBefore, &i.vbRHPrealloc)
 	}
 	i.data.InitOnce(r.keySchema, i.cmp, r.Split, getLazyValuer)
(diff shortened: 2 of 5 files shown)


sumeerbhola (approved):
nice!

:lgtm: 

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4123#-:-OAYDoSA-O_C7WUc61jY:btr8bhx)* status: all files reviewed (commit messages unreviewed), 4 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/value_block.go` line 725 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4123#-OAYALx7-T9KcBYDORtE:-OAYALx89h4nQcVgO36B:bt1pjom) ([raw file](https://github.com/cockroachdb/pebble/blob/5e95188ddc95ba7b76d0e909789843476937b261/sstable/value_block.go#L725)):*
> ```Go
> 
> // valueBlockReader implements GetLazyValueForPrefixAndValueHandler; it is used
> // to create a LazyValue that can be used to retrieve a values in a value block.
> ```

nit: a value ...
___
*[`sstable/value_block.go` line 752 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4123#-OAYCEV4DjGyiVswJnNi:-OAYCEV4DjGyiVswJnNj:b-u7ecr3) ([raw file](https://github.com/cockroachdb/pebble/blob/5e95188ddc95ba7b76d0e909789843476937b261/sstable/value_block.go#L752)):*
> ```Go
> 		//
> 		// Since it is a relatively small object, we could allocate multiple
> 		// instances together, using a sync.Pool.
> ```

I didn't understand this TODO, in that we don't have a good way to put back in the sync.Pool. Could you elaborate?
___
*[`sstable/value_block.go` line 856 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4123#-OAYD2HOBDdA9NGo4c54:-OAYD2HOBDdA9NGo4c55:bmqrr1c) ([raw file](https://github.com/cockroachdb/pebble/blob/5e95188ddc95ba7b76d0e909789843476937b261/sstable/value_block.go#L856)):*
> ```Go
> 	f.vbiCache.Release()
> 	// Set the handle to empty since Release does not nil the Handle.value. If
> 	// we were to reopen this valueBlockReader and retrieve the same
> ```

nit: reopen this valueBlockFetcher`.
___
*[`sstable/block/kv.go` line 77 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4123#-OAYDRpr1B8nplSUT20n:-OAYDRpr1B8nplSUT20o:bystrgr) ([raw file](https://github.com/cockroachdb/pebble/blob/5e95188ddc95ba7b76d0e909789843476937b261/sstable/block/kv.go#L77)):*
> ```Go
> 	// GetLazyValueForPrefixAndValueHandle returns a LazyValue for the given value
> 	// prefix and value. The result is only valid until the next call to
> 	// GetLazyValueForPrefixAndValueHandle.
> ```

Can you add:
Use LazyValue.Clone if the lifetime of the LazyValue needs to be extended. For more details, see the "memory management" comment where LazyValue is declared.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4123#-:-OAZ7tckEEIWdvqnD06J:b-814g8k)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/value_block.go` line 725 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4123#-OAYALx7-T9KcBYDORtE:-OAZ8VSs2EeFvH0T4YhG:b-wsxmub) ([raw file](https://github.com/cockroachdb/pebble/blob/5e95188ddc95ba7b76d0e909789843476937b261/sstable/value_block.go#L725)):*

nit: a value ...
</blockquote></details>

Fixed. Does the name `valueBlockReader` still make sense? It's more like a "valueBlockLazyValueConstructor` (but that sounds pretty terrible)
___
*[`sstable/value_block.go` line 752 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4123#-OAYCEV4DjGyiVswJnNi:-OAZ6l4dBv-UzuVyCFmb:byh7hf1) ([raw file](https://github.com/cockroachdb/pebble/blob/5e95188ddc95ba7b76d0e909789843476937b261/sstable/value_block.go#L752)):*

I didn't understand this TODO, in that we don't have a good way to put back in the sync.Pool. Could you elaborate?
</blockquote></details>

Pool object would be something like:
```
struct {
  nUsed int
  instances[16]
}
```

Alloc is: get an object from the pool, use `&instances[nUsed]`, increase `nUsed`, then put it back in the pool if `nUsed < 16`.

Basically just a bulk allocator (but using sync.Pool instead of a single structure with a mutex).

Improved the comment.
___
*[`sstable/block/kv.go` line 77 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4123#-OAYDRpr1B8nplSUT20n:-OAZ7sd771qTnYvnA3aN:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/5e95188ddc95ba7b76d0e909789843476937b261/sstable/block/kv.go#L77)):*

Can you add:
Use LazyValue.Clone if the lifetime of the LazyValue needs to be extended. For more details, see the "memory management" comment where LazyValue is declared.
</blockquote></details>

Done.

PR #4126, opened 2024-10-31
db: verify ingest sst bounds are suffixless or adjacent

### LINKED ISSUE #3829: ingest: verify ingestion sstables have legal boundary range keys
Currently, we don't allow range keys to have suffixed keys as bounds at any read interface point with Cockroach; the start,end of every range key fragment must be a suffixless key. On the write side, this is enforced in batch.go, but there's no such enforcement of this invariant in the ingestion path at all.

While reading every range key during ingestion is likely too expensive to be worthwhile, it's probably appropriate to verify (at ingestLoad time) that the smallest/largest range keys of an sstable at least have bounds that are suffixless, or if the bounds do contain suffixes, that these range keys perfectly defragment with range keys in adjacent ingestion sstables, such that a read post-ingestion would only surface suffixless keys and not panic in the read path.

Motivated from a conversation around https://github.com/cockroachdb/cockroach/pull/127997 and how it can fragment sstables at suffixed keys.

Jira issue: PEBBLE-234

itsbilal:
We now allow sstables to have range key bounds that have suffixes in them, during ingestion. This is to allow for Cockroach-side defragmentation of range keys. This change ensures that such suffixed range keys are adjacent and defrag cleanly with rangekeys in other ingested sstables, to not cause panics later on in the read path.

Fixes #3829.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -15,6 +15,7 @@ import (
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/cache"
 	"github.com/cockroachdb/pebble/internal/invariants"
+	"github.com/cockroachdb/pebble/internal/keyspan"
 	"github.com/cockroachdb/pebble/internal/manifest"
 	"github.com/cockroachdb/pebble/internal/overlap"
 	"github.com/cockroachdb/pebble/internal/sstableinternal"
@@ -239,8 +240,60 @@ func ingestLoad1External(
 	return meta, nil
 }
 
+type rangeKeyIngestValidator struct {
+	// lastRangeKey is the last range key seen in the previous file.
+	lastRangeKey keyspan.Span
+	// comparer, if unset, disables range key validation.
+	comparer *base.Comparer
+}
+
+func disableRangeKeyChecks() rangeKeyIngestValidator {
+	return rangeKeyIngestValidator{}
+}
+
+func validateSuffixedBoundaries(
+	cmp *base.Comparer, lastRangeKey keyspan.Span,
+) rangeKeyIngestValidator {
+	return rangeKeyIngestValidator{
+		lastRangeKey: lastRangeKey,
+		comparer:     cmp,
+	}
+}
+
+// Validate valides if the stored state of this rangeKeyIngestValidator allows for
+// a file with the given nextFileSmallestKey to be ingested, such that the stored
+// last file's largest range key defragments cleanly with the next file's smallest
+// key if it was suffixed. If a value of nil is passed in for nextFileSmallestKey,
+// that denotes the next file does not have a range key or there is no next file.
+func (r *rangeKeyIngestValidator) Validate(nextFileSmallestKey *keyspan.Span) error {
+	if r.comparer == nil {
+		return nil
+	}
+	if r.lastRangeKey.Valid() {
+		if r.comparer.Split.HasSuffix(r.lastRangeKey.End) {
+			if nextFileSmallestKey == nil || !r.comparer.Equal(r.lastRangeKey.End, nextFileSmallestKey.Start) {
+				// The last range key has a suffix, and it doesn't defragment cleanly with this range key.
+				return errors.AssertionFailedf("pebble: ingest sstable has suffixed largest range key that does not match the start key of the next sstable: %s",
+					r.comparer.FormatKey(r.lastRangeKey.End))
+			} else if !keyspan.DefragmentInternal.ShouldDefragment(r.comparer.CompareRangeSuffixes, &r.lastRangeKey, nextFileSmallestKey) {
+				// The last range key has a suffix, and it doesn't defragment cleanly with this range key.
+				return errors.AssertionFailedf("pebble: ingest sstable has suffixed range key that won't defragment with next sstable: %s",
+					r.comparer.FormatKey(r.lastRangeKey.End))
+			}
+		}
+	} else if nextFileSmallestKey != nil && r.comparer.Split.HasSuffix(nextFileSmallestKey.Start) {
+		return errors.Newf("pebble: ingest sstable has suffixed range key start that won't defragment: %s",
+			r.comparer.FormatKey(nextFileSmallestKey.Start))
+	}
+	return nil
+}
+
 // ingestLoad1 creates the FileMetadata for one file. This file will be owned
 // by this store.
+//
+// prevLastRangeKey is the last range key from the previous file. It is used to
+// ensure that the range keys defragment cleanly across files. These checks
+// are disabled if disableRangeKeyChecks is true.
 func ingestLoad1(
 	ctx context.Context,
 	opts *Options,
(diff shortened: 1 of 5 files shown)


itsbilal:
We might need to figure out a way to exclude disaggregated storage sstables from these assertions. Rangekeydels don't play well with the current logic, and are tripping up the metamorphic tests.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4126)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`ingest.go` line 245 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4126#-OAdKOhG30DHUaNPD447:-OAdKOhG30DHUaNPD448:b-a25vx) ([raw file](https://github.com/cockroachdb/pebble/blob/42019712a1b29bc78a29ba6e62eb2f3c573dd5d4/ingest.go#L245)):*
> ```Go
> // ingestLoad1 creates the FileMetadata for one file. This file will be owned
> // by this store.
> //
> ```

We should move the checking logic to a separate struct

```sql
type rangeKeyCheck struct {
  lastRangeKeySpan keyspan.Span
}

func (c *rangeKeyCheck) Check(span *keyspan.Span) error { .. }
```
We would have a `checkNoSuffixGap` constructor, or use `rangeKeyCheck{}` when we don't want any checks.


But I wonder if there is a better way than doing this here. Couldn't we run this check on the `meta`s after we do `ingestLoad1`? I'm assuming the `SmallestRangeKey` and `LargestRangeKey` in each meta would be enough to check for gaps?
___
*[`ingest.go` line 258 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4126#-OAdLHR4Am7OssXHkig3:-OAdLHR4Am7OssXHkig4:bkv7wkg) ([raw file](https://github.com/cockroachdb/pebble/blob/42019712a1b29bc78a29ba6e62eb2f3c573dd5d4/ingest.go#L258)):*
> ```Go
> 	lastRangeKey keyspan.Span,
> 	disableRangeKeyChecks bool,
> ) (*fileMetadata, keyspan.Span, error) {
> ```

The Span return looks very random, we should at least name it `lastRangeKeySpan`

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4126#-:-OBkqqlgDr7UaJaD-SvP:b-814g8k)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @sumeerbhola)

___
*[`ingest.go` line 245 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4126#-OAdKOhG30DHUaNPD447:-OBkdbok9hbI_SeBOAgF:bti9loe) ([raw file](https://github.com/cockroachdb/pebble/blob/42019712a1b29bc78a29ba6e62eb2f3c573dd5d4/ingest.go#L245)):*

We should move the checking logic to a separate struct

```sql
type rangeKeyCheck struct {
  lastRangeKeySpan keyspan.Span
}

func (c *rangeKeyCheck) Check(span *keyspan.Span) error { .. }
```
We would have a `checkNoSuffixGap` constructor, or use `rangeKeyCheck{}` when we don't want any checks.


But I wonder if there is a better way than doing this here. Couldn't we run this check on the `meta`s after we do `ingestLoad1`? I'm assuming the `SmallestRangeKey` and `LargestRangeKey` in each meta would be enough to check for gaps?
</blockquote></details>

We can't look at just the metas, unfortunately. We need to ensure the RangeKeySets and Unsets in the `[]Key` inside the bound spans are equivalent, otherwise we'd have to surface two different fragmented spans and one of those could have a suffix.

Did the struct change though
___
*[`ingest.go` line 258 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4126#-OAdLHR4Am7OssXHkig3:-OBkmvnlEVSIO3sTNbNZ:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/42019712a1b29bc78a29ba6e62eb2f3c573dd5d4/ingest.go#L258)):*

The Span return looks very random, we should at least name it `lastRangeKeySpan`
</blockquote></details>

Done.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4126#-:-OC5IxEx6L2_3r-llpQI:bnfp4nl)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`ingest.go` line 255 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4126#-OC5I1uL3iSN27TrHwVT:-OC5I1uL3iSN27TrHwVU:bmb6hhp) ([raw file](https://github.com/cockroachdb/pebble/blob/2165485a50b20cbd492d9fc65ce007655da51d65/ingest.go#L255)):*
> ```Go
> 	}
> 	if r.lastRangeKey.Valid() {
> 		if r.comparer.Split.HasSuffix(r.lastRangeKey.End) && (!r.comparer.Equal(r.lastRangeKey.End, nextFileSmallestKey.Start) ||
> ```

[nit] consider breaking up into two cases so we know from the message if it's the boundary or the keys that are the problem
___
*[`ingest.go` line 290 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4126#-OC5IDQS01rzWEg-YUID:-OC5IDQS01rzWEg-YUIE:b7hz8to) ([raw file](https://github.com/cockroachdb/pebble/blob/2165485a50b20cbd492d9fc65ce007655da51d65/ingest.go#L290)):*
> ```Go
> 		FileNum: base.PhysicalTableDiskFileNum(fileNum),
> 	})
> 	rangeKeyValidator := rangeKeyIngestValidator{}
> ```

I was thinking we would pass a `rangeKeyIngestValidator` to the function instead of the bool and the optional argument. And we would have two constructors like `NoRangeKeyValidation()` and `ValidateSuffixedBoundaries(...)`

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4126#-:-OC5Sbt63IWoJ95eVOYW:b-814g8k)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @sumeerbhola)

___
*[`ingest.go` line 255 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4126#-OC5I1uL3iSN27TrHwVT:-OC5OWNZ5EHHQy55KrJw:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/2165485a50b20cbd492d9fc65ce007655da51d65/ingest.go#L255)):*

[nit] consider breaking up into two cases so we know from the message if it's the boundary or the keys that are the problem
</blockquote></details>

Done.
___
*[`ingest.go` line 290 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4126#-OC5IDQS01rzWEg-YUID:-OC5RKbE4zp-wk_gcnKI:bqyceou) ([raw file](https://github.com/cockroachdb/pebble/blob/2165485a50b20cbd492d9fc65ce007655da51d65/ingest.go#L290)):*

I was thinking we would pass a `rangeKeyIngestValidator` to the function instead of the bool and the optional argument. And we would have two constructors like `NoRangeKeyValidation()` and `ValidateSuffixedBoundaries(...)`
</blockquote></details>

done. This exercise also reminded me of another case where we could have improved this checking, so thanks for highlighting that!

PR #4140, opened 2024-11-06
crl-release-23.2: db: avoid loading later files for a level in SeekPrefixGE

### LINKED ISSUE #3575: db: avoid loading later files for a level in SeekPrefixGE
In `levelIter.SeekPrefixGE` we do `loadFileIndicator := l.loadFile(l.findFileGE(key, flags), +1)` which does not care about the prefix (in CockroachDB, these `MVCCGet` iterators don't have bounds set), so can load an arbitrarily later file in the level. This is something we should optimize, since there could be gaps in levels which are causing us to unnecessarily load files. This is likely to be true for CockroachDB local keys, including locks and intents which we always try to read, since that part of the key space is very sparse.

(noticed due to https://github.com/cockroachdb/cockroach/pull/123313 where we had a failed read of an AbortSpanKey due to slow shared storage file read, even though the AbortSpanKey is a replicated range-id local key, so will never be in a shared storage file)

Jira issue: PEBBLE-299

itsbilal:
Don't load arbitrarily later files during levelIter.SeekPrefixGE. This avoids unnecessary IO and CPU overhead.

Close #3575.

### DIFF
--- a/internal/base/comparer.go
+++ b/internal/base/comparer.go
@@ -210,6 +210,10 @@ var DefaultComparer = &Comparer{
 		return append(dst, a...)
 	},
 
+	Split: func(a []byte) int {
+		return len(a)
+	},
+
 	ImmediateSuccessor: func(dst, a []byte) (ret []byte) {
 		return append(append(dst, a...), 0x00)
 	},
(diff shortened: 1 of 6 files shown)


jbowens (approved):

itsbilal:
TFTR!

PR #4143, opened 2024-11-07
block: improve FlushGovernor logic

RaduBerinde:
Currently the FlushGovernor logic results in blocks that are too small
(about 60% of the target). This was a result of a misinterpretation of
the previous code.

We now find the allocation size boundary that is closest to our target
and consider that boundary and the following one. For the 32KB
cockroach default block size, the boundaries are 32KB and 40KB. We
will flush right before exceeding the 32KB boundary, unless we have a
large enough KV that fits better in the 40KB class.

Informs https://github.com/cockroachdb/cockroach/issues/134420

### DIFF
--- a/sstable/block/flush_governor.go
+++ b/sstable/block/flush_governor.go
@@ -13,22 +13,26 @@ import (
 
 // FlushGovernor is used to decide when to flush a block. It takes into
 // consideration a target block size and (optionally) allocation size classes.
+//
+// When allocation size classes are used, we use the allocation class that is
+// closest to the target block size. We also take into account the next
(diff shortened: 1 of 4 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4143#-:-OBB72ySBZOa9nBkVfGL:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #4146, opened 2024-11-08
sstable: fix nil pointer with DisableValueBlocks set

jbowens:
Previously, the EstimatedSize method of the columnar sstable writer would panic when the writer was configured with DisableValueBlocks enabled.


jbowens:
TFTR!

PR #4147, opened 2024-11-08
sstable: reconcile out-of-order key error messages

jbowens:
When keys are written to a sstable writer out-of-order, the writer returns an error message. The error messages returned from the row-based and column-based writers varied, in part because the columnar writer did not readily have the previous key at hand. This commit adapts the column writer to synthesize the previous user key for the purpose of embedding it within the returned error message.

This will ensure cockroach unit tests that assert on these errors behave correctly when metamorphically switching between row and column blocks.


RaduBerinde (approved):

jbowens:
TFTR!

PR #4149, opened 2024-11-08
cache: misc tiny cleanups

sumeerbhola:
- Removed entry.peekValue, since there was existing code in clockpro.go that directly accesses e.val, and the method indirection hinders readability.
- Updated the comment about cgo pointer rules since entry no longer points to shard.
- Removed the comment about shard size which suggested that all the blocks of a file are mapped to the same shard, since that is not true.
- Added an assertion when a test entry has a value being set.

### DIFF
--- a/internal/cache/clockpro.go
+++ b/internal/cache/clockpro.go
@@ -160,7 +160,7 @@ func (c *shard) Set(id ID, fileNum base.DiskFileNum, offset uint64, value *Value
 			e = nil
 		}
 
-	case e.peekValue() != nil:
+	case e.val != nil:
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4149#-:-OBDNM-65eo95f3MOnC3:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sumeerbhola)

sumeerbhola:
TFTR!

PR #4152, opened 2024-11-08
tool: use MakeReaderOptions

jbowens:
Use MakeReaderOptions when constructing the sstable.ReaderOptions in individual subcommands. This is more maintainable and fixes a bug whereby the find command was unable to read columnar sstables written in a non-default KeySchema.


jbowens:
TFTR!

PR #4154, opened 2024-11-10
table cache: use a long-lived ReaderProvider

RaduBerinde:
An `sstable.ReaderProvider` is necessary whenever an sstable has value
blocks. Currently, we allocate one every time we create a point
iterator.

This commit makes the `tableCacheShardReaderProvider` usable in
parallel and maintains a long-lived instance in the `tableCacheValue`.

### DIFF
--- a/table_cache.go
+++ b/table_cache.go
@@ -557,10 +557,6 @@ func (c *tableCacheShard) newPointIter(
 	if err != nil {
 		return nil, err
 	}
-	var rp sstable.ReaderProvider
-	if tableFormat >= sstable.TableFormatPebblev3 && v.reader.Properties.NumValueBlocks > 0 {
-		rp = &tableCacheShardReaderProvider{c: c, file: file, dbOpts: dbOpts}
-	}
 
 	if v.isShared && file.SyntheticSeqNum() != 0 {
 		if tableFormat < sstable.TableFormatPebblev4 {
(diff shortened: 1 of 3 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4154#-:-OBTKNGj07HWZYzglSv_:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`table_cache.go` line 706 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4154#-OBTKw6JEscHHyrJ969-:-OBTKw6JEscHHyrJ9690:bv05yrq) ([raw file](https://github.com/cockroachdb/pebble/blob/4a53cdf5df3aaf654bd6318c1392308b28b7354b/table_cache.go#L706)):*
> ```Go
> 	// Calling findNode gives us the responsibility of decrementing v's
> 	// refCount.
> 	v := rp.c.findNodeInternal(ctx, rp.backingFileNum, rp.dbOpts)
> ```

Consider adding a comment that `fineNodeInternal` can do IO, in order to initialize a `Reader`, and that we are deliberately holding `rp.mu` while doing so, in that concurrent calls to GetReader will block until this initialization is done.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4154)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`table_cache.go` line 706 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4154#-OBTKw6JEscHHyrJ969-:-OBTR3pTDOF5d6UtzCHA:b-qqx0w8) ([raw file](https://github.com/cockroachdb/pebble/blob/4a53cdf5df3aaf654bd6318c1392308b28b7354b/table_cache.go#L706)):*

Consider adding a comment that `fineNodeInternal` can do IO, in order to initialize a `Reader`, and that we are deliberately holding `rp.mu` while doing so, in that concurrent calls to GetReader will block until this initialization is done.
</blockquote></details>

Done. TFTR!

PR #4157, opened 2024-11-12
cache,db: de-dup concurrent attempts to read the same block

### LINKED ISSUE #4138: cache: don't load the same block in parallel
Currently when we have a block cache miss: we allocate a buffer, issue a read, then populate a cache. If multiple readers miss on the same block, they will all allocate and issue the read (and only the last result will stay in the cache).

We have run into a few support issues where this contributed to a problem (e.g. when a filter block was very large).

We could insert a phantom entry in the cache first and have any subsequent readers wait for it instead. Or have a separate data structure for in-progress blocks.

Jira issue: PEBBLE-297

sumeerbhola:
Concurrent reads of the same block have been observed to cause very high
memory usage, and cause significant CPU usage for allocations/deallocations.
We now coordinate across multiple concurrent attempts to read the same
block via a readEntry, which makes the readers take turns until one
succeeds.

The readEntries are embedded in a map that is part of a readShard, where
there is a readShard for each cache.Shard. See the long comment in the
readShard declaration for motivation.

Callers interact with this new behavior via Cache.GetWithReadHandle,
which is only for callers that intend to do a read and then populate
the cache. If this method returns a ReadHandle, the caller has permission
to do a read. See the ReadHandle comment for details of the contract.

Fixes #4138

### DIFF
--- a/internal/cache/clockpro.go
+++ b/internal/cache/clockpro.go
@@ -18,6 +18,7 @@
 package cache // import "github.com/cockroachdb/pebble/internal/cache"
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"runtime"
@@ -25,6 +26,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/invariants"
@@ -114,9 +116,31 @@ type shard struct {
 	countHot  int64
 	countCold int64
 	countTest int64
+
+	// Some fields in readShard are protected by mu. See comments in declaration
+	// of readShard.
+	readShard readShard
 }
 
-func (c *shard) Get(id ID, fileNum base.DiskFileNum, offset uint64) Handle {
+func (c *shard) init(maxSize int64) {
+	*c = shard{
+		maxSize:    maxSize,
+		coldTarget: maxSize,
+	}
+	if entriesGoAllocated {
+		c.entries = make(map[*entry]struct{})
+	}
+	c.blocks.Init(16)
+	c.files.Init(16)
+	c.readShard.Init(c)
+}
+
+// getWithMaybeReadEntry is the internal helper for implementing
+// Cache.{Get,GetWithReadHandle}. When desireReadEntry is true, and the block
+// is not in the cache (!Handle.Valid()), a non-nil readEntry is returned.
+func (c *shard) getWithMaybeReadEntry(
+	id ID, fileNum base.DiskFileNum, offset uint64, desireReadEntry bool,
+) (Handle, *readEntry) {
 	c.mu.RLock()
 	var value *Value
 	if e, _ := c.blocks.Get(key{fileKey{id, fileNum}, offset}); e != nil {
@@ -126,12 +150,30 @@ func (c *shard) Get(id ID, fileNum base.DiskFileNum, offset uint64) Handle {
 		}
 	}
 	c.mu.RUnlock()
+	var re *readEntry
+	if value == nil && desireReadEntry {
+		c.mu.Lock()
+		// After the c.mu.RUnlock(), someone could have inserted the value in the
+		// cache. We could tolerate the race and do a file read, or do another map
+		// lookup. We choose to do the latter, since the cost of a map lookup is
+		// insignificant compared to the cost of reading a block from a file.
+		if e, _ := c.blocks.Get(key{fileKey{id, fileNum}, offset}); e != nil {
+			value = e.acquireValue()
+			if value != nil {
+				e.referenced.Store(true)
+			}
+		}
+		if value == nil {
+			re = c.readShard.getReadEntryLocked(id, fileNum, offset)
+		}
+		c.mu.Unlock()
+	}
 	if value == nil {
 		c.misses.Add(1)
-		return Handle{}
+	} else {
+		c.hits.Add(1)
 	}
-	c.hits.Add(1)
-	return Handle{value: value}
+	return Handle{value: value}, re
 }
 
 func (c *shard) Set(id ID, fileNum base.DiskFileNum, offset uint64, value *Value) Handle {
@@ -170,6 +212,11 @@ func (c *shard) Set(id ID, fileNum base.DiskFileNum, offset uint64, value *Value
 			value.ref.trace("add-hot")
 			c.sizeHot += delta
 		} else {
+			// TODO(sumeer): unclear why we don't set e.ptype to etHot on this path.
+			// In the default case below, where the state is etTest we set it to
+			// etHot. But etTest is "colder" than etCold, since the only transition
+			// into etTest is etCold => etTest, so since etTest transitions to
+			// etHot, then etCold should also transition.
 			value.ref.trace("add-cold")
 			c.sizeCold += delta
 		}
@@ -737,15 +784,7 @@ func newShards(size int64, shards int) *Cache {
 	c.idAlloc.Store(1)
 	c.trace("alloc", c.refs.Load())
 	for i := range c.shards {
-		c.shards[i] = shard{
-			maxSize:    size / int64(len(c.shards)),
-			coldTarget: size / int64(len(c.shards)),
-		}
-		if entriesGoAllocated {
-			c.shards[i].entries = make(map[*entry]struct{})
-		}
-		c.shards[i].blocks.Init(16)
-		c.shards[i].files.Init(16)
+		c.shards[i].init(size / int64(len(c.shards)))
 	}
 
 	// Note: this is a no-op if invariants are disabled or race is enabled.
@@ -822,7 +861,49 @@ func (c *Cache) Unref() {
 // Get retrieves the cache value for the specified file and offset, returning
 // nil if no value is present.
 func (c *Cache) Get(id ID, fileNum base.DiskFileNum, offset uint64) Handle {
-	return c.getShard(id, fileNum, offset).Get(id, fileNum, offset)
+	h, re := c.getShard(id, fileNum, offset).getWithMaybeReadEntry(
+		id, fileNum, offset, false)
+	if invariants.Enabled && re != nil {
+		panic("readEntry should be nil")
+	}
+	return h
+}
+
+// GetWithReadHandle retrieves the cache value for the specified ID, fileNum
+// and offset. If found, a valid Handle is returned (with cacheHit set to
+// true), else a valid ReadHandle is returned.
+//
+// See the ReadHandle declaration for the contract the caller must satisfy
+// when getting a valid ReadHandle.
+//
+// This method can block before returning since multiple concurrent gets for
+// the same cache value will take turns getting a ReadHandle, which represents
+// permission to do the read. This blocking respects context cancellation, in
+// which case an error is returned (and not a valid ReadHandle).
+//
+// When blocking, the errorDuration return value can be non-zero and is
+// populated with the total duration that other readers that observed an error
+// (see ReadHandle.SetReadError) spent in doing the read. This duration can be
+// greater than the time spent blocked in this method, since some of these
+// errors could have occurred prior to this call. But it serves as a rough
+// indicator of whether turn taking could have caused higher latency due to
+// context cancellation of other readers.
+//
+// While waiting, someone else may successfully read the value, which results
+// in a valid Handle being returned. This is a case where cacheHit=false.
+func (c *Cache) GetWithReadHandle(
+	ctx context.Context, id ID, fileNum base.DiskFileNum, offset uint64,
+) (h Handle, rh ReadHandle, errorDuration time.Duration, cacheHit bool, err error) {
+	h, re := c.getShard(id, fileNum, offset).getWithMaybeReadEntry(
+		id, fileNum, offset, true)
+	if h.Valid() {
+		return h, ReadHandle{}, 0, true, nil
+	}
+	h, errorDuration, err = re.waitForReadPermissionOrHandle(ctx)
+	if err != nil || h.Valid() {
+		return h, ReadHandle{}, errorDuration, false, err
+	}
+	return Handle{}, ReadHandle{entry: re}, errorDuration, false, nil
 }
 
 // Set sets the cache value for the specified file and offset, overwriting an
--- a/internal/cache/read_shard.go
+++ b/internal/cache/read_shard.go
@@ -1,0 +1,414 @@
+// Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package cache
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/swiss"
+)
+
+// readShard coordinates the read of a block that will be put in the cache. It
+// ensures only one goroutine is reading a block, and other callers block
+// until that goroutine is done (with success or failure). In the case of
+// success, the other goroutines will use the value that was read, even if it
+// is too large to be placed in the cache, or got evicted from the cache
+// before they got scheduled. In the case of a failure (read error or context
+// cancellation), one of the waiters will be given a turn to do the read.
+//
+// This turn-taking ensures that a large number of concurrent attempts to read
+// the same block that is not in the cache does not result in the same number
+// of reads from the filesystem (or remote storage). We have seen large spikes
+// in memory usage (and CPU usage for memory allocation/deallocation) without
+// this turn-taking.
+//
+// It introduces a small risk related to context cancellation -- if many
+// readers assigned a turn exceed their deadline while doing the read and
+// report an error, a reader with a longer deadline can unnecessarily wait. We
+// accept this risk for now since the primary production use in CockroachDB is
+// filesystem reads, where context cancellation is not respected. We do
+// introduce an error duration metric emitted in traces that can be used to
+// quantify such wasteful waiting. Note that this same risk extends to waiting
+// on the Options.LoadBlockSema, so the error duration metric includes the
+// case of an error when waiting on the semaphore (as a side effect of that
+// waiting happening in the caller, sstable.Reader).
+//
+// Design choices and motivation:
+//
+//   - readShard is tightly integrated with a cache shard: At its core,
+//     readShard is a map with synchronization. For the same reason the cache is
+//     sharded (for higher concurrency by sharding the mutex), it is beneficial
+//     to shard synchronization on readShard. By making readShard a member of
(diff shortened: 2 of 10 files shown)


sumeerbhola (commented):
This lacks tests -- I would like an opinion on the approach and interfaces before adding those.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4157#-:-OBSrMO3EcZ2WsUo6X1h:bm90dww)* status: 0 of 8 files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4157)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`sstable/reader.go` line 525 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4157#-OBT4YxU2nJhpvkem473:-OBT4YxU2nJhpvkem474:bq1pgjw) ([raw file](https://github.com/cockroachdb/pebble/blob/0be60e5387f0703428389b13172d5c263737b69f/sstable/reader.go#L525)):*
> ```Go
> 		var err error
> 		var errorDuration time.Duration
> 		ch, errorDuration, err = crh.WaitForReadPermissionOrHandle(ctx)
> ```

Have you considered a `GetOrPopulate` API where we pass a func that does the actual read (or we can define a simple interface that `*Reader` implements)? Then the new logic would be internal to `GetOrPopulate`. All the caller cares is that at the end we either get a BufferHandle or an error. We can keep the simple `Get` for the buffer pool case.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4157)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/reader.go` line 525 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4157#-OBT4YxU2nJhpvkem473:-OBbxSs32ZjxWvpXmaqJ:b-phcvxa) ([raw file](https://github.com/cockroachdb/pebble/blob/0be60e5387f0703428389b13172d5c263737b69f/sstable/reader.go#L525)):*

I didn't seriously consider it since there is a lot of shared code in `readBlockInternal` but there is also enough branching for the `cache.ReadHandle` case. We would want to write one func that does the actual read in both cases. So it would need to have the branching too. It would avoid the need for the following blocks which precede some return statements:
```
if crh.Valid() {
			crh.SetReadError(err)
}
```
I've moved those into a defer call here, so that by itself is not a big motivator for `GetOrPopulate`.

I've now done an implementation of `readBlockInternal` (readBlockInternal2), using (an unimplemented) `GetOrPopulate` in https://github.com/sumeerbhola/pebble/commit/bac895b7859269659a2743af67fbca8e14cd597f. PTAL. By itself, it isn't simpler IMO. But it does hide `ReadHandle` so it is perhaps compelling enough. Thoughts?
</blockquote></details>

Yeah, that looks worse.. I was envisioning that the function it calls would be a separate method, but I haven't considered all the details.

Why do we need a separate `WaitForReadPermissionOrHandle`? Can't this happen inside `GetWithReadHandle` - it could return either the data from the cache (regardless if we had to wait for it), or the handle that we will use to populate the cache after we do the read. I think only the first to create the entry should do the read, and anyone who comes in after should wait.

CC @jbowens as well
___
*[`sstable/reader.go` line 505 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4157#-OBc1FJgDn9o9Uv01G_d:-OBc1FJh4JJ57_IwG5Ol:b-k742a3) ([raw file](https://github.com/cockroachdb/pebble/blob/23f4b7fa73852610ea13fbad9e2d45edb3e60453/sstable/reader.go#L505)):*
> ```Go
> 	if env.BufferPool == nil {
> 		ch, crh = r.cacheOpts.Cache.GetWithReadHandle(
> 			r.cacheOpts.CacheID, r.cacheOpts.FileNum, bh.Offset, r.loadBlockSema)
> ```

Why do we need to move the semaphore acquire into the cache code? That should just happen right before we issue the read, with the expectation that outside of strange conditions it won't do anything (and if it does, it should just behave like a slower IO). We can treat a ctx cancelation error the same way we would treat an IO error.
___
*[`sstable/reader.go` line 552 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4157#-OBc0aYy4ggGxTsdC5SX:-OBc0aYy4ggGxTsdC5SY:b-rury6p) ([raw file](https://github.com/cockroachdb/pebble/blob/23f4b7fa73852610ea13fbad9e2d45edb3e60453/sstable/reader.go#L552)):*
> ```Go
> 	// INVARIANT: !ch.Valid().
> 
> 	compressed := block.Alloc(int(bh.Length+block.TrailerLen), env.BufferPool)
> ```

It would be cleaner to separate out the code starting here into a method that does the actual read and returns the `decompressed` block handle. We can `SetReadError` more cleanly after that call (this is error prone if someone adds an `if err :=  ...` case).

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4157)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/cache/read_shard.go` line 92 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4157#-OBfUxGZ0BLw443dc_jT:-OBfUxGZ0BLw443dc_jU:byglk5g) ([raw file](https://github.com/cockroachdb/pebble/blob/23f4b7fa73852610ea13fbad9e2d45edb3e60453/internal/cache/read_shard.go#L92)):*
> NB: we cannot place the loadBlockSema in the readShard since the readShard can be shared across the DBs, just like the block cache.

Should we make the load block semaphore shared across DBs, like the block cache? The primary motivation was preventing excessive memory utilization from many concurrent reads, right? And memory is a shared resource across DBs.

RaduBerinde:
> Should we make the load block semaphore shared across DBs, like the block cache? The primary motivation was preventing excessive memory utilization from many concurrent reads, right? And memory is a shared resource across DBs.

That would have been easier to implement, but there was concern that one store that is broken and has very slow IOs will block out all other stores.
https://github.com/cockroachdb/cockroach/blob/3644f0d3fe77c03c20d3b603c1c0eb6f335e7e15/pkg/storage/pebble.go#L997

sumeerbhola (commented):
I've reworked the logic based on the comments. I'll start adding tests if this seems reasonable.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4157#-:-OBnByvr8sXCm_FdIHdF:b3co3ab)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/cache/read_shard.go` line 92 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4157#-OBfUxGZ0BLw443dc_jT:-OBnBaKI7wOW3ljpdQ0O:b-58funs) ([raw file](https://github.com/cockroachdb/pebble/blob/23f4b7fa73852610ea13fbad9e2d45edb3e60453/internal/cache/read_shard.go#L92)):*

> NB: we cannot place the loadBlockSema in the readShard since the readShard can be shared across the DBs, just like the block cache.

Should we make the load block semaphore shared across DBs, like the block cache? The primary motivation was preventing excessive memory utilization from many concurrent reads, right? And memory is a shared resource across DBs.
</blockquote></details>

As @RaduBerinde mentioned, this choice was made for performance isolation of stores.
___
*[`sstable/reader.go` line 525 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4157#-OBT4YxU2nJhpvkem473:-OBn7rzHBGXwP3YFFdAw:b-41l90s) ([raw file](https://github.com/cockroachdb/pebble/blob/0be60e5387f0703428389b13172d5c263737b69f/sstable/reader.go#L525)):*
> Why do we need a separate `WaitForReadPermissionOrHandle`? Can't this happen inside `GetWithReadHandle`.

Yes, it can. It involves passing the context and needs additional return values. I've made the change.
___
*[`sstable/reader.go` line 505 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4157#-OBc1FJgDn9o9Uv01G_d:-OBn89Vt-_OeFvoRJ-a7:b-1xr2n8) ([raw file](https://github.com/cockroachdb/pebble/blob/23f4b7fa73852610ea13fbad9e2d45edb3e60453/sstable/reader.go#L505)):*

Why do we need to move the semaphore acquire into the cache code? That should just happen right before we issue the read, with the expectation that outside of strange conditions it won't do anything (and if it does, it should just behave like a slower IO). We can treat a ctx cancelation error the same way we would treat an IO error.
</blockquote></details>

Initially it felt cleaner to move all the waiting in one place. And it required less explanation about the contract. But given that we need to still do semaphore acquisition for the `BufferPool` case, I've moved it back.

The contract comment now says:
```
// The caller must immediately start doing a read, or can first wait on a
// shared resource that would also block a different reader if it was assigned
// the turn instead (specifically, this refers to Options.LoadBlockSema).
```
___
*[`sstable/reader.go` line 552 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4157#-OBc0aYy4ggGxTsdC5SX:-OBn7meK66AE0w1jlmdL:b-7qmo62) ([raw file](https://github.com/cockroachdb/pebble/blob/23f4b7fa73852610ea13fbad9e2d45edb3e60453/sstable/reader.go#L552)):*

It would be cleaner to separate out the code starting here into a method that does the actual read and returns the `decompressed` block handle. We can `SetReadError` more cleanly after that call (this is error prone if someone adds an `if err :=  ...` case).
</blockquote></details>

Good point. Done

RaduBerinde (commented):
The approach looks good to me. I will make another pass in more detail but you can start adding tests.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4157#-:-OC0SFiX-nI95KOTOC43:b-4huu86)* status: 0 of 8 files reviewed, 5 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/cache/clockpro.go` line 206 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4157#-OC0Qvtm3v_qwBFvwtbu:-OC0Qvtm3v_qwBFvwtbv:b-1hu437) ([raw file](https://github.com/cockroachdb/pebble/blob/bf70930c7086d5ae83543d2ca9e4c2ba044cc423/internal/cache/clockpro.go#L206)):*
> ```Go
> 			// etHot. But etTest is "colder" than etCold, since the only transition
> 			// into etTest is etCold => etTest, so since etTest transitions to
> 			// etHot, then etCold should also transition.
> ```

I think you're right. The [paper](https://www.usenix.org/legacy/publications/library/proceedings/usenix05/tech/general/full_papers/jiang/jiang_html/html.html) says that if the page isn't in the list it is added as cold (the `e == nil` case) but if it is "the faulted page turns into a hot page and is placed at the head of the list"

Maybe file an issue? We'd want it in a separate PR

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4157)* status: 0 of 8 files reviewed, 5 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/cache/clockpro.go` line 206 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4157#-OC0Qvtm3v_qwBFvwtbu:-OCEF2gu7Nj8edRct3OR:byp4tw3) ([raw file](https://github.com/cockroachdb/pebble/blob/bf70930c7086d5ae83543d2ca9e4c2ba044cc423/internal/cache/clockpro.go#L206)):*

I think you're right. The [paper](https://www.usenix.org/legacy/publications/library/proceedings/usenix05/tech/general/full_papers/jiang/jiang_html/html.html) says that if the page isn't in the list it is added as cold (the `e == nil` case) but if it is "the faulted page turns into a hot page and is placed at the head of the list"

Maybe file an issue? We'd want it in a separate PR
</blockquote></details>

https://github.com/cockroachdb/pebble/issues/4178

sumeerbhola (commented):
Tests are ready.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4157#-:-ODJYnMa8uEA4SbN3zrF:b-w1azid)* status: 0 of 10 files reviewed, 5 unresolved discussions (waiting on @jbowens and @RaduBerinde)

RaduBerinde (approved):
:lgtm:  Very cool!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4157#-:-ODmJf580pH296i7DxZK:b-94k0ld)* status: 0 of 10 files reviewed, 4 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/cache/clockpro.go` line 141 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4157#-ODmHtdu25iZ9qgvlG8R:-ODmHtdu25iZ9qgvlG8S:b-whr3fm) ([raw file](https://github.com/cockroachdb/pebble/blob/f299ee0cca6bb62c25af3cf3af1374b7efb06ea8/internal/cache/clockpro.go#L141)):*
> ```Go
> // Cache.{Get,GetWithReadHandle}. When desireReadEntry is true, and the block
> // is not in the cache (!Handle.Valid()), a non-nil readEntry is returned.
> func (c *shard) GetWithMaybeReadEntry(
> ```

[nit] does it need to be exported?
___
*[`internal/cache/read_shard.go` line 69 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4157#-ODmI_ox4v9MHjZLKu7w:-ODmI_ox4v9MHjZLKu7x:b-tez5ur) ([raw file](https://github.com/cockroachdb/pebble/blob/f299ee0cca6bb62c25af3cf3af1374b7efb06ea8/internal/cache/read_shard.go#L69)):*
> ```Go
> //     separate map. This separation also results in more modular code, instead
> //     of piling more stuff into shard.
> type readShard struct {
> ```

[nit] Would it be cleaner to just add `readMap` to `shard` and just implement the methods on `*shard`? We can still keep them separate.
___
*[`internal/cache/read_shard.go` line 226 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4157#-ODmJlvI9XgdI-Dz7M11:-ODmJlvI9XgdI-Dz7M12:b-5ohykv) ([raw file](https://github.com/cockroachdb/pebble/blob/f299ee0cca6bb62c25af3cf3af1374b7efb06ea8/internal/cache/read_shard.go#L226)):*
> ```Go
> 		return errorDuration
> 	}
> start:
> ```

We should do `for {` and use `continue`. Well, we don't even need continue, we can do
```
case _, ok := <-ch:
  if !ok {
    ...
    return ..
  }
  // Probably granted permission to do the read; check again. NB: since isReading is
  // false, someone else can slip through before this thread acquires
  // e.mu, and take the turn.
```

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4157#-:-ODmoPcBBvfBVt6_kiww:b-814g8k)* status: 0 of 10 files reviewed, 4 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/cache/clockpro.go` line 141 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4157#-ODmHtdu25iZ9qgvlG8R:-ODmk9mg1wQgensqbKoa:b-2nfos0) ([raw file](https://github.com/cockroachdb/pebble/blob/f299ee0cca6bb62c25af3cf3af1374b7efb06ea8/internal/cache/clockpro.go#L141)):*

[nit] does it need to be exported?
</blockquote></details>

Fixed. I fell into the trap of following the pattern of other methods in `shard` being exported (`Set/Delete/...`), none of which need to be.
___
*[`internal/cache/read_shard.go` line 69 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4157#-ODmI_ox4v9MHjZLKu7w:-ODml-Sv5nbqcPx9h660:bxf63k7) ([raw file](https://github.com/cockroachdb/pebble/blob/f299ee0cca6bb62c25af3cf3af1374b7efb06ea8/internal/cache/read_shard.go#L69)):*

[nit] Would it be cleaner to just add `readMap` to `shard` and just implement the methods on `*shard`? We can still keep them separate.
</blockquote></details>

This is subjective, but I like the bigger separation. Putting this data-structure in `shard` will move a bunch of commentary above into the `shard` declaration. Some of the methods could still be in this file of course. Keeping this more separate I think makes our code more easy to maintain.
___
*[`internal/cache/read_shard.go` line 226 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4157#-ODmJlvI9XgdI-Dz7M11:-ODmlpok0CPTRSYaoUh4:b-qu6n94) ([raw file](https://github.com/cockroachdb/pebble/blob/f299ee0cca6bb62c25af3cf3af1374b7efb06ea8/internal/cache/read_shard.go#L226)):*

We should do `for {` and use `continue`. Well, we don't even need continue, we can do
```
case _, ok := <-ch:
  if !ok {
    ...
    return ..
  }
  // Probably granted permission to do the read; check again. NB: since isReading is
  // false, someone else can slip through before this thread acquires
  // e.mu, and take the turn.
```
</blockquote></details>

Good point. Done.

RaduBerinde:
CI occasionally encounters `entry was not freed` failures; it might be related to this PR.

https://github.com/cockroachdb/pebble/actions/runs/12304080308/job/34340708418?pr=4199
https://github.com/cockroachdb/pebble/actions/runs/12300880108/job/34330136010
https://github.com/cockroachdb/pebble/actions/runs/12300880108/job/34330132989

RaduBerinde:
*[`sstable/reader.go` line 512 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4157#-OI1tr-l1wNKMTZFncvW:-OI1tr-l1wNKMTZFncvX:b4bb6lx) ([raw file](https://github.com/cockroachdb/pebble/blob/4f7bb5f48114d50d64a6e1f5b79b179aac1d3394/sstable/reader.go#L512)):*
> ```Go
> 	var ch cache.Handle
> 	var crh cache.ReadHandle
> 	hit := true
> ```

I just noticed this which should be `false`. I'll fix.

RaduBerinde:
*[`sstable/reader.go` line 512 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4157#-OI1tr-l1wNKMTZFncvW:-OI1uvwl1BPhOmhDteUR:b-q7ptu5) ([raw file](https://github.com/cockroachdb/pebble/blob/4f7bb5f48114d50d64a6e1f5b79b179aac1d3394/sstable/reader.go#L512)):*

I just noticed this which should be `false`. I'll fix.
</blockquote></details>

Oh, never mind, it doesn't change anything (in the case of buffer pools, we only look at `hit` if `ch.Valid()`

PR #4159, opened 2024-11-12
metamorphic: add a test option to make downloadOp no-op

itsbilal:
Previously, if we had external storage enabled, we'd always download in a download op. This meant we didn't really test cases where a download could change the state of the LSM relative to not doing the download.

This change adds a metamorphic test option (toggled randomly) that disables the download operation in 50% of runs.

### DIFF
--- a/metamorphic/ops.go
+++ b/metamorphic/ops.go
@@ -203,6 +203,10 @@ type downloadOp struct {
 }
 
 func (o *downloadOp) run(t *Test, h historyRecorder) {
+	if t.testOpts.disableDownloads {
+		h.Recordf("%s // %v", o, nil)
+		return
+	}
 	db := t.getDB(o.dbID)
 	err := t.withRetries(func() error {
 		return db.Download(context.Background(), o.spans)
--- a/metamorphic/options.go
+++ b/metamorphic/options.go
@@ -169,6 +169,9 @@ func parseOptions(
 					return opts.useDeleteOnlyCompactionExcises
 				}
 				return true
+			case "TestOptions.disable_downloads":
+				opts.disableDownloads = true
+				return true
 			case "TestOptions.use_jemalloc_size_classes":
 				opts.Opts.AllocatorSizeClasses = pebble.JemallocSizeClasses
 				return true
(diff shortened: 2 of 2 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4159#-:-OBbLrGz84M6f9PkXlaF:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @itsbilal)

___
*[`metamorphic/ops.go` line 207 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4159#-OBbLWKd1fwfOwZC-x9T:-OBbLWKd1fwfOwZC-x9U:b-s545a7) ([raw file](https://github.com/cockroachdb/pebble/blob/7dafc20e8f6eaf4598bc224eac1fc617f5a49937/metamorphic/ops.go#L207)):*
> ```Go
> func (o *downloadOp) run(t *Test, h historyRecorder) {
> 	if t.testOpts.disableDownloads {
> 		h.Recordf("%s // %v", o, nil)
> ```

is there something that could cause the `Download` to fail and so a non-nil error would cause a false positive diff and test failure?

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4159#-:-OBzxKug-taciXyC3Jzn:b-814g8k)* status: all files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`metamorphic/ops.go` line 207 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4159#-OBbLWKd1fwfOwZC-x9T:-OBzxA2b0Nu1EOSNz7-3:b-uy4exw) ([raw file](https://github.com/cockroachdb/pebble/blob/7dafc20e8f6eaf4598bc224eac1fc617f5a49937/metamorphic/ops.go#L207)):*

is there something that could cause the `Download` to fail and so a non-nil error would cause a false positive diff and test failure?
</blockquote></details>

There shouldn't be any reasons for the download to fail, no. We'd want any errors to cause a test failure.

PR #4160, opened 2024-11-12
sstable: order range keys consistently in colblk and rowblk encodings

jbowens:
Previously the ordering of range keys within a range key block was a bit muddled and undefined. Before encoding a Span of range keys to the underlying RawWriter, the sstable.Writer type previously sorted the span's keys by suffix. However, the row-based sstable writer always serializes RANGEKEYSETs of a Span first, followed by RANGEKEYUNSETs and then RANGEKEYDELs.

Confusingly, the rowblk fragment iterator also sorted keys by trailer when iterating backwards, but not forwards. The columnar RawWriter preserved the order of keys in the span passed to EncodeSpan, and this order was preserved during iteration.

This commit adapts the sstable.Writer type to sort a range key Span's keys by trailer and then suffix before encoding. This provides determinism and matches the ordering of keys produced by compactions which sort the keys by trailer. Additionally, the rowblk fragment iterator is updated to always sort the returned keys by trailer.

### DIFF
--- a/internal/keyspan/span.go
+++ b/internal/keyspan/span.go
@@ -470,9 +470,14 @@ func SortKeysByTrailer(keys []Key) {
 	})
 }
 
-// SortKeysBySuffix sorts a Keys slice by suffix.
-func SortKeysBySuffix(suffixCmp base.CompareRangeSuffixes, keys []Key) {
+// SortKeysByTrailerAndSuffix sorts a Keys slice by trailer, and among keys with
+// equal trailers, by suffix.
+func SortKeysByTrailerAndSuffix(suffixCmp base.CompareRangeSuffixes, keys []Key) {
 	slices.SortFunc(keys, func(a, b Key) int {
+		// Trailer are ordered in decreasing number order.
+		if v := cmp.Compare(b.Trailer, a.Trailer); v != 0 {
+			return v
+		}
 		return suffixCmp(a.Suffix, b.Suffix)
 	})
 }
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4160#-:-OBXDmIJ2KkMBtWJzpCJ:bnfp4nl)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @itsbilal and @jbowens)

___
*[`internal/keyspan/span.go` line 478 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4160#-OBXD8cgC4PejUruJ_iH:-OBXD8cgC4PejUruJ_iI:b-rbm9oa) ([raw file](https://github.com/cockroachdb/pebble/blob/433b7cb54fb624df940a7b02aff7a0e407e15e7e/internal/keyspan/span.go#L478)):*
> ```Go
> 	slices.SortFunc(keys, func(a, b Key) int {
> 		// Trailer are ordered in decreasing number order.
> 		if v := -cmp.Compare(a.Trailer, b.Trailer); v != 0 {
> ```

[nit] can be `cmp.Compare(b.Trailer, a.Trailer)`

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4160#-:-OBXP_Vw0LyCqOVLKp5S:b-814g8k)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @itsbilal)

___
*[`internal/keyspan/span.go` line 478 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4160#-OBXD8cgC4PejUruJ_iH:-OBXPaNy7Pps-LYRbDAR:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/433b7cb54fb624df940a7b02aff7a0e407e15e7e/internal/keyspan/span.go#L478)):*

[nit] can be `cmp.Compare(b.Trailer, a.Trailer)`
</blockquote></details>

Done

Issue #4162, opened 2024-11-13
db: improve logging observability into WAL failover


jbowens:
We should improve the observability into WAL failover via diagnostic logs.

We could at least log when we switch between primary and secondary.

We should also look at improving the WAL replay logic to ensure we log sufficient information about which physical WALs were read during recovery.

PR #4166, opened 2024-11-13
db: log additional WAL details during replay

jbowens:
When we finish replaying a WAL, log the logical log which will encode all the replayed segment files (if WAL failover resulted in multiple physical files).

```
[JOB 1] WAL 000002: {(data,000)} stopped reading at offset: (data/000002.log: 0); replayed 0 keys in 0 batches
```

Additionally, implement the SafeFormatter interface to ensure that relevant details are not redacted.

Informs #4162.


RaduBerinde (approved):

jbowens:
TFTR!

PR #4168, opened 2024-11-14
bump crlib, switch to crlib/fifo

RaduBerinde:
The `fifo` library has moved into `crlib`; switch to the updated
location.


RaduBerinde:
TFTR!

PR #4181, opened 2024-11-25
sstable: fix improper buffer reuse in copyDataBlocks

itsbilal:
Previously, in the colblk implementation of copyDataBlocks, we were reusing a buffer that could under some cases be passed directly to the write queue and would get written to sstable while later blocks are being read into the same buffer.

This change also improves tests around CopySpan() to better test cache hit/miss cases.

Fixes https://github.com/cockroachdb/cockroach/issues/131332.


itsbilal:
TFTR!

jbowens:
We should backport to crl-release-24.3 too (which will eventually have columnar blocks toggleable in a patch release)

PR #4183, opened 2024-11-27
wal: prevent negative durations

jbowens:
The WAL failover monitor tracks the duration spent writing to the primary and secondary WAL directories. Previously it was possible for updates to these durations to be negative (despite implicitly using the time.Time's monotonic time), because the time was retrieved before acquiring the mutex guarding the accumulated durations. A goroutine that measured the current time first was not necessarily the first to enter the critical section and update the monitor's lastAccumulateIntoDurations field.

This commit moves these time measurements under the mutex and adds an assertion that the monitor's understanding of time moves monotonically forward.

Informs cockroachdb/cockroach#136317.

PR #4184, opened 2024-11-27
crl-release-24.3: wal: prevent negative durations

jbowens:
24.3 backport of #4183 to avoid test flakes on the 24.3 release branch.

----

The WAL failover monitor tracks the duration spent writing to the primary and secondary WAL directories. Previously it was possible for updates to these durations to be negative (despite implicitly using the time.Time's monotonic time), because the time was retrieved before acquiring the mutex guarding the accumulated durations. A goroutine that measured the current time first was not necessarily the first to enter the critical section and update the monitor's lastAccumulateIntoDurations field.

This commit moves these time measurements under the mutex and adds an assertion that the monitor's understanding of time moves monotonically forward.

Informs cockroachdb/cockroach#136317.

### DIFF
--- a/wal/failover_manager.go
+++ b/wal/failover_manager.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/invariants"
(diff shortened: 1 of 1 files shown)


sumeerbhola (approved):

jbowens:
TFTR!

PR #4188, opened 2024-12-03
db: rename TableCache to FileCache

jbowens:
With the introduction of blob files for value separation (https://github.com/cockroachdb/pebble/issues/112), Pebble will
need to mantain file descriptors into non-sstable blob files. CockroachDB
attempts to respect configured file descriptor limits by propagating them to
Pebble as the size of the table cache. To continue to respect file descriptor
limits, Pebble will need to constrain both open sstables and blob files.

Ths commit renames the TableCache to a FileCache in preparation for storing
open blob file readers in the same cache. Entries in the cache are keyed by
disk file number which will still be unique with the introduction of blob
files. This rename will help prevent confusion from storing 'blob files' in a
'table cache'.


itsbilal (approved):
LGTM

RaduBerinde:
BTW, one day we should bite the bullet and rename `FileNum` to `TableNum` (and then much later `DiskFileNum` back to `FileNum`).

jbowens:
TFTR!

>  BTW, one day we should bite the bullet and rename FileNum to TableNum (and then much later DiskFileNum back to FileNum).

Yeah, that's much better

PR #4189, opened 2024-12-04
sstable: use ComparePointSuffix to compare against synthetic suffix

jbowens:
This commit adapts colblk.DataBlockIter to use a base.ComparePointSuffixes func
to compare keys against the synthetic suffix. Previously we used base.Compare
which is no longer a valid use of the base.Compare contract.

Informs https://github.com/cockroachdb/cockroach/issues/136709.

### DIFF
--- a/internal/base/comparer.go
+++ b/internal/base/comparer.go
@@ -177,7 +177,8 @@ type ImmediateSuccessor func(dst, a []byte) []byte
 //  3. Suffixes themselves must be valid keys and comparable, respecting the same
 //     ordering as within a key:
 //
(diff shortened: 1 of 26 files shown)


itsbilal (approved):
LGTM

In general I like that we're heading towards passing around `*base.Comparer` instead of `base.Compare`.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4189#-:-ODIoNqN5u3tdx8q8df9:bnfp4nl)* status: 0 of 26 files reviewed, all discussions resolved

jbowens:
TFTRs!

PR #4192, opened 2024-12-05
crl-release-24.3: sstable: use ComparePointSuffix to compare against synthetic suffix

jbowens:
This commit adapts colblk.DataBlockIter to use a base.ComparePointSuffixes func
to compare keys against the synthetic suffix. Previously we used base.Compare
which is no longer a valid use of the base.Compare contract.

Informs https://github.com/cockroachdb/cockroach/issues/136709.

24.3 backport of #4189.


RaduBerinde (approved):

jbowens:
TFTR!

PR #4195, opened 2024-12-10
arenaskl: remove node.allocSize field

nvanbenschoten:
The field was added in c34894c4 and has been unused since c34894c4.

The `node` struct remains the same size due to padding.

I found this while looking into cache misses in the memtable's arena-backed skiplist and exploring ways (e.g. value separation, value + key suffix separation) to improve memory locality. So far, I haven't found anything that shows promise on microbenchmarks.


RaduBerinde (approved):

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4195#-:-ODlNSrjCvmdRNtd2FS2:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved

PR #4198, opened 2024-12-12
internal/compact: defer retrieval of external values

### LINKED ISSUE #4197: db: refactor compaction iteration interfaces for deferring value retrieval
With #112, we will sometimes copy a blob value reference as-is, without ever retrieving the referenced value. Today the compaction loop and iterator greedily load the value. This issue tracks work to refactor the compaction iterator interfaces to propagate LazyValues and references as necessary.

Jira issue: PEBBLE-313

jbowens:
During compaction iteration, defer the retrieval of values stored in value blocks until either:

a) we need to copy the value before stepping the iterator, or b) the compaction iterator yields the value to the main compaction loop.

This refactor ensures that when a KV is elided by a tombstone, we avoid unnecessarily loading its value from the external value block. Additionally, refactoring the compaction iterator interfaces to propagate LazyValues will be used by value separation (#112) in which we will sometimes propagate external value references to output sstables without ever retrieving the value.

Close #4197.
Informs #112.

### DIFF
--- a/internal/base/lazy_value.go
+++ b/internal/base/lazy_value.go
@@ -222,6 +222,12 @@ func (lv *LazyValue) fetchValue(
 	return f.value, f.callerOwned, f.err
 }
 
+// IsInPlaceValue returns true iff the value was stored in-place and does not
+// need to be fetched externally.
+func (lv *LazyValue) IsInPlaceValue() bool {
+	return lv.Fetcher == nil
+}
+
 // InPlaceValue returns the value under the assumption that it is in-place.
 // This is for Pebble-internal code.
 func (lv *LazyValue) InPlaceValue() []byte {
--- a/internal/base/merger.go
+++ b/internal/base/merger.go
@@ -7,7 +7,7 @@ package base
 import "io"
 
 // Merge creates a ValueMerger for the specified key initialized with the value
-// of one merge operand.
+// of one merge operand. The caller retains ownership of key and value.
 type Merge func(key, value []byte) (ValueMerger, error)
 
 // ValueMerger receives merge operands one by one. The operand received is either
--- a/internal/compact/iterator.go
+++ b/internal/compact/iterator.go
@@ -178,7 +178,7 @@ type Iter struct {
 	// original trailer (eg, before any sequence-number zeroing or changes to
 	// key kind).
 	keyTrailer  base.InternalKeyTrailer
-	value       []byte
+	value       base.LazyValue
 	valueCloser io.Closer
 	// Temporary buffer used for storing the previous user key in order to
 	// determine when iteration has advanced to a new user key and thus a new
@@ -189,7 +189,6 @@ type Iter struct {
 	// advanced.
 	valueBuf         []byte
 	iterKV           *base.InternalKV
-	iterValue        []byte
 	iterStripeChange stripeChangeType
 	// skip indicates whether the remaining entries in the current snapshot
 	// stripe should be skipped or processed. `skip` has no effect when `pos ==
@@ -362,16 +361,12 @@ func (i *Iter) Stats() IterStats {
 }
 
 // First has the same semantics as InternalIterator.First.
-func (i *Iter) First() (*base.InternalKey, []byte) {
+func (i *Iter) First() (*base.InternalKey, base.LazyValue) {
 	if i.err != nil {
-		return nil, nil
+		return nil, base.LazyValue{}
 	}
 	i.iterKV = i.iter.First()
 	if i.iterKV != nil {
-		i.iterValue, _, i.err = i.iterKV.Value(nil)
-		if i.err != nil {
-			return nil, nil
-		}
 		i.curSnapshotIdx, i.curSnapshotSeqNum = i.cfg.Snapshots.IndexAndSeqNum(i.iterKV.SeqNum())
 	}
 	i.pos = iterPosNext
@@ -382,14 +377,14 @@ func (i *Iter) First() (*base.InternalKey, []byte) {
 // Next has the same semantics as InternalIterator.Next. Note that when Next
 // returns a RANGEDEL or a range key, the caller can use Span() to get the
 // corresponding span.
-func (i *Iter) Next() (*base.InternalKey, []byte) {
+func (i *Iter) Next() (*base.InternalKey, base.LazyValue) {
 	if i.err != nil {
-		return nil, nil
+		return nil, base.LazyValue{}
 	}
 
 	// Close the closer for the current value if one was open.
 	if i.closeValueCloser() != nil {
-		return nil, nil
+		return nil, base.LazyValue{}
 	}
 
 	// Prior to this call to `Next()` we are in one of three situations with
@@ -477,12 +472,15 @@ func (i *Iter) Next() (*base.InternalKey, []byte) {
 			// This goes against the comment on i.key in the struct, and
 			// therefore warrants some investigation.
 			i.saveKey()
+			i.value = i.iterKV.V
+			if invariants.Enabled && !i.value.IsInPlaceValue() {
+				panic(errors.AssertionFailedf("pebble: span key's value is not in-place"))
+			}
 			// TODO(jackson): Handle tracking pinned statistics for range keys
 			// and range deletions. This would require updating
 			// emitRangeDelChunk and rangeKeyCompactionTransform to update
 			// statistics when they apply their own snapshot striping logic.
 			i.snapshotPinned = false
-			i.value = i.iterValue
 			return &i.key, i.value
 		}
 
@@ -545,7 +543,7 @@ func (i *Iter) Next() (*base.InternalKey, []byte) {
 			switch i.iterKV.Kind() {
 			case base.InternalKeyKindDelete:
 				i.saveKey()
-				i.value = i.iterValue
+				i.value = base.LazyValue{} // DELs are value-less.
 				i.skip = true
 				return &i.key, i.value
 
@@ -559,7 +557,7 @@ func (i *Iter) Next() (*base.InternalKey, []byte) {
 				if i.singleDeleteNext() {
 					return &i.key, i.value
 				} else if i.err != nil {
-					return nil, nil
+					return nil, base.LazyValue{} // SINGLEDELs are value-less.
 				}
 				continue
 
@@ -576,7 +574,7 @@ func (i *Iter) Next() (*base.InternalKey, []byte) {
 			// kind.
 			i.setNext()
 			if i.err != nil {
-				return nil, nil
+				return nil, base.LazyValue{}
 			}
 			return &i.key, i.value
 
@@ -585,7 +583,8 @@ func (i *Iter) Next() (*base.InternalKey, []byte) {
 			// advances the iterator, adjusting curSnapshotIdx.
 			origSnapshotIdx := i.curSnapshotIdx
 			var valueMerger base.ValueMerger
-			valueMerger, i.err = i.cfg.Merge(i.iterKV.K.UserKey, i.iterValue)
+			// MERGE values are always stored in-place.
+			valueMerger, i.err = i.cfg.Merge(i.iterKV.K.UserKey, i.iterKV.V.InPlaceValue())
 			if i.err == nil {
 				i.mergeNext(valueMerger)
 			}
@@ -607,7 +606,7 @@ func (i *Iter) Next() (*base.InternalKey, []byte) {
 			if i.err == nil {
 				if needDelete {
 					if i.closeValueCloser() != nil {
-						return nil, nil
+						return nil, base.LazyValue{}
 					}
 					continue
 				}
@@ -620,15 +619,15 @@ func (i *Iter) Next() (*base.InternalKey, []byte) {
 				// MERGE?
 				i.err = base.MarkCorruptionError(i.err)
 			}
-			return nil, nil
+			return nil, base.LazyValue{}
 
 		default:
 			i.err = base.CorruptionErrorf("invalid internal key kind: %d", errors.Safe(i.iterKV.Kind()))
-			return nil, nil
+			return nil, base.LazyValue{}
 		}
 	}
 
-	return nil, nil
+	return nil, base.LazyValue{}
 }
 
 // Span returns the range deletion or range key span corresponding to the
@@ -666,11 +665,8 @@ func (i *Iter) skipInStripe() {
 
 func (i *Iter) iterNext() bool {
 	i.iterKV = i.iter.Next()
-	if i.iterKV != nil {
-		i.iterValue, _, i.err = i.iterKV.Value(nil)
-		if i.err != nil {
-			i.iterKV = nil
-		}
+	if i.iterKV == nil {
+		i.err = i.iter.Error()
 	}
 	return i.iterKV != nil
 }
@@ -779,7 +775,7 @@ func (i *Iter) nextInStripeHelper() stripeChangeType {
 func (i *Iter) setNext() {
 	// Save the current key.
 	i.saveKey()
-	i.value = i.iterValue
+	i.value = i.iterKV.V
 	i.maybeZeroSeqnum(i.curSnapshotIdx)
 
 	// If this key is already a SETWITHDEL we can early return and skip the remaining
@@ -789,9 +785,8 @@ func (i *Iter) setNext() {
 		return
 	}
 
-	// We are iterating forward. Save the current value.
-	i.valueBuf = append(i.valueBuf[:0], i.iterValue...)
-	i.value = i.valueBuf
+	// We need to iterate forward. Save the current value so we don't lose it.
+	i.saveValue()
 
 	// Else, we continue to loop through entries in the stripe looking for a
 	// DEL. Note that we may stop *before* encountering a DEL, if one exists.
@@ -874,7 +869,16 @@ func (i *Iter) mergeNext(valueMerger base.ValueMerger) {
 			// value and return. We change the kind of the resulting key to a
 			// Set so that it shadows keys in lower levels. That is:
 			// MERGE + (SET*) -> SET.
-			i.err = valueMerger.MergeOlder(i.iterValue)
+			var v []byte
+			var callerOwned bool
+			v, callerOwned, i.err = i.iterKV.Value(i.valueBuf[:0])
+			if i.err != nil {
+				return
+			}
+			if callerOwned && cap(v) > cap(i.valueBuf) {
+				i.valueBuf = v
+			}
+			i.err = valueMerger.MergeOlder(v)
 			if i.err != nil {
 				return
 			}
@@ -885,7 +889,9 @@ func (i *Iter) mergeNext(valueMerger base.ValueMerger) {
 		case base.InternalKeyKindMerge:
 			// We've hit another Merge value. Merge with the existing value and
 			// continue looping.
-			i.err = valueMerger.MergeOlder(i.iterValue)
+			//
+			// MERGE values are always stored in-place.
+			i.err = valueMerger.MergeOlder(i.iterKV.InPlaceValue())
 			if i.err != nil {
 				return
 			}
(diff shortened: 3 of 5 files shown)


jbowens:
Hrm, two cache `entry was not freed` errors in CI: [one](https://github.com/cockroachdb/pebble/actions/runs/12300880108/job/34330136010) on 32-bit linux and [another](https://github.com/cockroachdb/pebble/actions/runs/12300880108/job/34330132989) on macOS. I've been unable to repro

RaduBerinde:
I saw that error in CI also: https://github.com/cockroachdb/pebble/pull/4199

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4198#-:-ODwarq8CVLkJv3b7hM-:bnfp4nl)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @itsbilal, @jbowens, and @sumeerbhola)

___
*[`internal/compact/iterator.go` line 1291 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwaVIx7CzB32DClA6y:-ODwaVIx7CzB32DClA6z:b-v5eviw) ([raw file](https://github.com/cockroachdb/pebble/blob/f4e9e04b0021bacbf80f009e9982ab1b50041ff9/internal/compact/iterator.go#L1291)):*
> ```Go
> 		i.value = base.LazyValue{}
> 		return
> 	} else if !callerOwned {
> ```

[nit] remove the `else` or the `return`
___
*[`internal/compact/iterator.go` line 1408 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwacjJCdZ_a0GnAxkN:-ODwacjK3yME2WOZsOT9:bvlm0wq) ([raw file](https://github.com/cockroachdb/pebble/blob/f4e9e04b0021bacbf80f009e9982ab1b50041ff9/internal/compact/iterator.go#L1408)):*
> ```Go
> 	}
> 	lv = base.MakeInPlaceValue(value)
> 	return
> ```

[nit] `return base.MakeInPlaceValue(value), needDelete, closer, err` and `_ base.LazyValue`

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4198)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @itsbilal and @sumeerbhola)

___
*[`internal/compact/iterator.go` line 1291 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwaVIx7CzB32DClA6y:-ODwfIAI6ShfvFrOC7R4:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/f4e9e04b0021bacbf80f009e9982ab1b50041ff9/internal/compact/iterator.go#L1291)):*

[nit] remove the `else` or the `return`
</blockquote></details>

Done
___
*[`internal/compact/iterator.go` line 1408 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwacjJCdZ_a0GnAxkN:-ODwfVWV8_SXiSZ3oPSq:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/f4e9e04b0021bacbf80f009e9982ab1b50041ff9/internal/compact/iterator.go#L1408)):*

[nit] `return base.MakeInPlaceValue(value), needDelete, closer, err` and `_ base.LazyValue`
</blockquote></details>

Done

sumeerbhola (changes requested):
> Reviewed 4 of 5 files at r1, 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4198)* status: 5 of 6 files reviewed, 11 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`internal/compact/run.go` line 218 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwsmNCCPo7BlLF6f02:-ODwsmNCCPo7BlLF6f03:bnm5vnt) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/run.go#L218)):*
> ```Go
> 			continue
> 		}
> 		v, _, err := value.Value(nil)
> ```

We used to call `LazyValue.Value(nil)` to initialize `compact.Iter.iterValue` for the current iterator position. That made sense since it would never be caller owned. And then any instability was handled by reusing `valueBuf`.

Now we have two kinds of `LazyValue`s here: those that have not been cloned, and those that have been. For the latter, the `valueBlockFetcher` may be closed, in which case it will need to allocate. So I think we need another scratch buffer in this `Runner` to avoid those allocations.
___
*[`internal/compact/iterator_test.go` line 204 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwu6ez61qYwc_FzsjH:-ODwu6ez61qYwc_FzsjI:b-s3pkw4) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator_test.go#L204)):*
> ```Go
> 						d.Fatalf(t, "unknown iter command: %s", parts[0])
> 					}
> 					value := lv.InPlaceValue()
> ```

why is the test assuming in-place instead of using the more general `LazyValue.Value`?
___
*[`internal/compact/iterator.go` line 560 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwkrGVD3gTDdkdXEIH:-ODwkrGVD3gTDdkdXEII:b-wlkn50) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L560)):*
> ```Go
> 					return &i.key, i.value
> 				} else if i.err != nil {
> 					return nil, base.LazyValue{}
> ```

nit: consider repeating the comment from above `// SINGLEDELs are value-less`.
___
*[`internal/compact/iterator.go` line 586 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwl7Pi-yNISepav-AR:-ODwl7Pi-yNISepav-AS:bwnt48) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L586)):*
> ```Go
> 			origSnapshotIdx := i.curSnapshotIdx
> 			var valueMerger base.ValueMerger
> 			// MERGE values are always stored in-place.
> ```

Can you add an invariants.Enabled assertion.
___
*[`internal/compact/iterator.go` line 668 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwlVmSDZjCG1uVebOi:-ODwlVmSDZjCG1uVebOj:b-376r0d) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L668)):*
> ```Go
> func (i *Iter) iterNext() bool {
> 	i.iterKV = i.iter.Next()
> 	if i.err != nil {
> ```

It seems that we should no longer need to look at `i.err` given we are not changing it, and it will be nil when this function is called.
___
*[`internal/compact/iterator.go` line 893 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwpEQ85cdNzmCUSCSt:-ODwpEQ90tSTkerUPuWK:byuglf7) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L893)):*
> ```Go
> 			// continue looping.
> 			//
> 			// MERGE values are always stored in-place.
> ```

some comment about adding an assertion
___
*[`internal/compact/iterator.go` line 919 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwqQ5GDgpkMEHSocJV:-ODwqQ5GDgpkMEHSocJW:bal58bg) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L919)):*
> ```Go
> 	// Save the current key.
> 	i.saveKey()
> 	i.value = i.iterKV.V
> ```

This `i.value = ...` seems dubious. We are going to immediately call `i.nextInStripe()`, and we haven't cloned the `LazyValue`, so this value is no longer backed by anything.
I think it worked because `i.iterValue` was always the empty slice in this case. And now with the `LazyValue` being in-place, the `LazyFetcher` will be nil, so it will work. But is in principle incorrect. I think we should explicitly assert that this is in-place, and the empty slice, and then set `i.value = LazyValue{}`.
___
*[`internal/compact/iterator.go` line 1114 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwpajm84-qWpNa237R:-ODwpajm84-qWpNa237S:bln64zx) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L1114)):*
> ```Go
> 	// In this case, we still peek forward in case there's another DELSIZED key
> 	// with a lower sequence number, in which case we'll adopt its value.
> 	// If the DELSIZED does have a value, it must be in-place.
> ```

assertion please
___
*[`internal/compact/iterator.go` line 1160 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwrlKu7Ur10Dy18cVd:-ODwrlKu7Ur10Dy18cVe:bln64zx) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L1160)):*
> ```Go
> 			// If the tombstone has a value, it must be in-place. To save it, we
> 			// can just copy the in-place value directly.
> 			i.valueBuf = append(i.valueBuf[:0], i.iterKV.InPlaceValue()...)
> ```

assertion please
___
*[`internal/compact/iterator.go` line 1286 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwmxmxAVIJq3ycAUHX:-ODwmxmxAVIJq3ycAUHY:b-b898tl) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L1286)):*
> ```Go
> // result into i.valueBuf).
> func (i *Iter) saveValue() {
> 	v, callerOwned, err := i.iterKV.V.Value(i.valueBuf[:0])
> ```

I don't think we should be fetching. And should be doing `i.value, i.valueBuf = i.iterKV.V.Clone(i.valueBuf[:0], &i.scratchFetcher)`
___
*[`internal/compact/iterator.go` line 192 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwjqby-holesD3-BzH:-ODwjqby-holesD3-BzI:bojoqwd) ([raw file](https://github.com/cockroachdb/pebble/blob/ab9741a3ded6ac566068c653f68c8d69945ffc8f/internal/compact/iterator.go#L192)):*
> ```Go
> 	valueBuf         []byte
> 	iterKV           *base.InternalKV
> 	iterValue        []byte
> ```

I'm glad `iterValue` is no more. It seemed to be redundant with `iterKV` but if `iterKV` was nil we wouldn't set it to nil, which seemed error prone.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4198)* status: 3 of 6 files reviewed, 10 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`internal/compact/iterator.go` line 560 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwkrGVD3gTDdkdXEIH:-OE-y7Hi0nqX7O9RrFdN:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L560)):*

nit: consider repeating the comment from above `// SINGLEDELs are value-less`.
</blockquote></details>

Done.
___
*[`internal/compact/iterator.go` line 586 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwl7Pi-yNISepav-AR:-OE-y-gv5xUbFaPOEvWJ:bezoqk4) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L586)):*

Can you add an invariants.Enabled assertion.
</blockquote></details>

`InPlaceValue()` already asserts.
___
*[`internal/compact/iterator.go` line 668 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwlVmSDZjCG1uVebOi:-OE-xYsJ0WiOmREFZZye:by0aazt) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L668)):*

It seems that we should no longer need to look at `i.err` given we are not changing it, and it will be nil when this function is called.
</blockquote></details>

You're right, although I'm confused as to why we're not checking `i.iter.Error()` here?
___
*[`internal/compact/iterator.go` line 893 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwpEQ85cdNzmCUSCSt:-OE-xUMr5VwqLecLFB2N:bezoqk4) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L893)):*

some comment about adding an assertion
</blockquote></details>

`InPlaceValue()` already asserts.
___
*[`internal/compact/iterator.go` line 919 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwqQ5GDgpkMEHSocJV:-OE-xQHh-JnxK3ha8djW:b-3glmgh) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L919)):*

This `i.value = ...` seems dubious. We are going to immediately call `i.nextInStripe()`, and we haven't cloned the `LazyValue`, so this value is no longer backed by anything.
I think it worked because `i.iterValue` was always the empty slice in this case. And now with the `LazyValue` being in-place, the `LazyFetcher` will be nil, so it will work. But is in principle incorrect. I think we should explicitly assert that this is in-place, and the empty slice, and then set `i.value = LazyValue{}`.
</blockquote></details>

Agreed. 
___
*[`internal/compact/iterator.go` line 1114 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwpajm84-qWpNa237R:-OE-wVuk-b_R8u7c5ILr:bezoqk4) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L1114)):*

assertion please
</blockquote></details>

`InPlaceValue()` already asserts.
___
*[`internal/compact/iterator.go` line 1160 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwrlKu7Ur10Dy18cVd:-OE-wY9H-yqFVEOvdMgJ:bezoqk4) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L1160)):*

assertion please
</blockquote></details>

`InPlaceValue()` already asserts.
___
*[`internal/compact/iterator.go` line 1286 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwmxmxAVIJq3ycAUHX:-OE-wbXz83tzGiad0b4O:b-4ibj3x) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L1286)):*

I don't think we should be fetching. And should be doing `i.value, i.valueBuf = i.iterKV.V.Clone(i.valueBuf[:0], &i.scratchFetcher)`
</blockquote></details>

I had that initially but decided I did not want introduce the possibility of needing to load the value from an external block reader in this PR. Fetching the value keeps the behavior closer to what was here originally, constraining the surface area of this PR.
___
*[`internal/compact/run.go` line 218 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwsmNCCPo7BlLF6f02:-OE-z-ET9-2q-Rg9pjLG:bgasmz8) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/run.go#L218)):*

We used to call `LazyValue.Value(nil)` to initialize `compact.Iter.iterValue` for the current iterator position. That made sense since it would never be caller owned. And then any instability was handled by reusing `valueBuf`.

Now we have two kinds of `LazyValue`s here: those that have not been cloned, and those that have been. For the latter, the `valueBlockFetcher` may be closed, in which case it will need to allocate. So I think we need another scratch buffer in this `Runner` to avoid those allocations.
</blockquote></details>

Since we're not currently `Clone`-ing, this value is necessarily in-place. My preference is to defer any use of Clone until a subsequent PR. I'm unsure it's ever worthwhile for compactions to Clone value block references. I don't think there's ever an instance where we call `saveValue` and then _don't_ end up reading the value, because we only need to iterate forward like this to determine `SET` vs `SETWITHDEL`
___
*[`internal/compact/iterator_test.go` line 204 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwu6ez61qYwc_FzsjH:-OE-yI_NDNC0bGJ6Keu0:bf3wiea) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator_test.go#L204)):*

why is the test assuming in-place instead of using the more general `LazyValue.Value`?
</blockquote></details>

This particular unit test just does not support anything other than in-place values. It's trivial to use `Value`. Done, so there's one less thing to edit if we extend the test to construct non in-place values.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4198)* status: 3 of 6 files reviewed, 10 unresolved discussions (waiting on @itsbilal and @sumeerbhola)

___
*[`internal/compact/run.go` line 218 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwsmNCCPo7BlLF6f02:-OE02n7i8aim4DsI-nPH:bz0zxh3) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/run.go#L218)):*
> Since we're not currently `Clone`\-ing, this value is necessarily in-place.

I mean, this value is necessarily in-place _or_ for the current iterator position.

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 2 files at r3, 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4198#-:-OE13cPd9OvN8HltlyZg:bg7igkv)* status: all files reviewed, 10 unresolved discussions (waiting on @itsbilal and @jbowens)

___
*[`internal/compact/iterator.go` line 668 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwlVmSDZjCG1uVebOi:-OE11qYkDB_joc2Kbi2S:bxqhoaj) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L668)):*

You're right, although I'm confused as to why we're not checking `i.iter.Error()` here?
</blockquote></details>

It's probably because what is returned in `i.iter.Error()` will also be returned in `i.iter.Close()`.
___
*[`internal/compact/iterator.go` line 1286 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwmxmxAVIJq3ycAUHX:-OE12alr6hxzRHpT65rN:b8ozr4) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/iterator.go#L1286)):*

I had that initially but decided I did not want introduce the possibility of needing to load the value from an external block reader in this PR. Fetching the value keeps the behavior closer to what was here originally, constraining the surface area of this PR.
</blockquote></details>

I'm slightly worried we will forget it. Could you add a TODO, and also include in that scope that run.go should use a scratch buffer?
___
*[`internal/compact/run.go` line 218 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4198#-ODwsmNCCPo7BlLF6f02:-OE1382t5Z_J8BBEIs33:bvlledv) ([raw file](https://github.com/cockroachdb/pebble/blob/2c1fd7e60e73d0dbb04d807dc57c0209c2a487f0/internal/compact/run.go#L218)):*
> I don't think there's ever an instance where we call `saveValue` and then _don't_ end up reading the value, because we only need to iterate forward like this to determine `SET` vs `SETWITHDEL`

But we will in the future, yes? When a compaction is not rewriting its references to blob files.

PR #4200, opened 2024-12-12
sstable: optimize allocation of uncompressed buffers

RaduBerinde:
When we don't compress a data buffer, the `RawColumnWriter` makes a
clone of the data buffer. In the sysbench workload, we see this
allocation taking place for almost half of the blocks (suggesting that
the data isn't very compressible).

Use the buffer that we would have used for the compressed data instead
of allocating a new one.

### DIFF
--- a/sstable/block/compression.go
+++ b/sstable/block/compression.go
@@ -208,6 +208,13 @@ func (b PhysicalBlock) Clone() PhysicalBlock {
 	return PhysicalBlock{data: data, trailer: b.trailer}
 }
 
+// CloneUsingBuf makes a copy of the block data, using the given slice if it has
+// enough capacity.
+func (b PhysicalBlock) CloneUsingBuf(buf []byte) (_ PhysicalBlock, newBuf []byte) {
+	newBuf = append(buf[:0], b.data...)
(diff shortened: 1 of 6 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4200#-:-OE00tf04naq2BKiSPUG:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`sstable/rowblk_writer.go` line 465 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4200#-OE00ZuI63x91L6PFul1:-OE00ZuI63x91L6PFul2:b-h9x8vw) ([raw file](https://github.com/cockroachdb/pebble/blob/26e87ab1965bd7cfedb2d1cd6f95f199e8aa87db/sstable/rowblk_writer.go#L465)):*
> ```Go
> 	tmp [blockHandleLikelyMaxLen]byte
> 	// dataBuf is the destination buffer for compression, or (in some cases where
> 	// decompression is not used) for storing a copy of the data. It is re-used
> ```

nit: "compression is not used"

RaduBerinde:
TFTR!

PR #4204, opened 2024-12-17
internal: add SetFinalizerWithStack and fix test bug

sumeerbhola:
SetFinalizerWithStack prints out the stack trace from when the finalizer was set. There was a bug in TestReadShard, in that it was not calling shard.Free, which is now fixed.

### DIFF
--- a/internal/cache/entry.go
+++ b/internal/cache/entry.go
@@ -177,10 +177,10 @@ func entryAllocNew() *entry {
 		// We want to allocate each entry independently to check that it has been
 		// properly cleaned up.
 		e := &entry{}
-		invariants.SetFinalizer(e, func(obj interface{}) {
+		invariants.SetFinalizerWithStack(e, func(obj interface{}, stack []byte) {
 			e := obj.(*entry)
 			if *e != (entry{}) {
-				fmt.Fprintf(os.Stderr, "%p: entry was not freed", e)
+				fmt.Fprintf(os.Stderr, "%p: entry was not freed\n%s", e, stack)
 				os.Exit(1)
 			}
 		})
--- a/internal/cache/read_shard_test.go
+++ b/internal/cache/read_shard_test.go
@@ -109,11 +109,18 @@ func TestReadShard(t *testing.T) {
 	var c *shard
 	var readers map[string]*testReader
 	var mu sync.Mutex
-
+	freeShard := func() {
+		if c != nil {
+			c.Free()
+			c = nil
+		}
+	}
+	defer freeShard()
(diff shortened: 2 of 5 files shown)


sumeerbhola:
I am at a loss wrt the `TestLint/TestSetFinalizer` failure. When logging the string pairs sent to `lintIgnore` on master, there is a single spurious string which doesn't have a filename printed with it. Since it is not part of a pair, the test does not fail.

```
=== RUN   TestLint/TestSetFinalizer
    lint_test.go:342: first: internal/cache/entry.go-		// lint:ignore SetFinalizer
    lint_test.go:345: second: internal/cache/entry.go:		runtime.SetFinalizer(c, freeEntryAllocCache)
    lint_test.go:342: first: --
```

With the changes here there is an additional spurious string, and since the check happens on the second string, the test fails.
```
=== RUN   TestLint/TestSetFinalizer
    lint_test.go:330: first: internal/cache/entry.go-		// lint:ignore SetFinalizer
    lint_test.go:332: second: internal/cache/entry.go:		runtime.SetFinalizer(c, freeEntryAllocCache)
    lint_test.go:330: first: --
    lint_test.go:332: second: --
```

RaduBerinde (approved):
:lgtm: 

Nice, I did this locally a few times to debug. Hopefully the extra allocation won't slow down tests a lot.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4204#-:-OELDHOc1DLwkLyUshIu:bjguonn)* status: 0 of 5 files reviewed, all discussions resolved

RaduBerinde:
I see tests are timing out.  Perhaps just printing stack traces for all goroutines at failure time would be helpful enough (we won't know the path that created the object, but we'll know what test is running).

RaduBerinde (commented):
:lgtm:

Consider keeping the stack code but behind a `const debugFinalizers = false`.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4204#-:-OEPKiMF4l_z2STj-A2U:bo5n1va)* status: 0 of 5 files reviewed, all discussions resolved

sumeerbhola:
Bizarrely, now that I removed the stack code, code coverage is failing with `0xc0006a83c0: entry was not freed`. And I can't reproduce with `stress`.
I'm going to merge this and will come back to this again if we still see this bug in the wild.

RaduBerinde:
> Bizarrely, now that I removed the stack code, code coverage is failing with 0xc0006a83c0: entry was not freed

Code coverage runs tests both with your changes and without. It's failing in the `Generate "before" coverage` step.

PR #4206, opened 2024-12-19
crl-release-24.3: valblk: fix Writer.Size() performance

RaduBerinde:
#### sstable: improve writer benchmark

Add varying `vals-per-key` parameter and add an `EstimatedSize()` call
per key (which happens during compactions).

#### crl-release-24.3: valblk: fix Writer.Size() performance

`Writer.Size()` unnecessarily loops through all the blocks instead of
using `totalBlockBytes`. This seems like an oversight: we only append
to `blocks` in `compressAndFlush` where we also update
`totalBlockBytes`. I verified that in all Pebble tests these values are
the same, using `if sz != w.totalBlockBytes { panic() }`.

### DIFF
--- a/sstable/colblk_writer.go
+++ b/sstable/colblk_writer.go
@@ -204,9 +204,7 @@ func (w *RawColumnWriter) EstimatedSize() uint64 {
 		sz += uint64(w.rangeKeyBlock.Size())
 	}
 	if w.valueBlock != nil {
-		for _, blk := range w.valueBlock.blocks {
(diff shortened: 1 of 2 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4206#-:-OEU3ZLfAwt1FTm-Iqqv:bg7igkv)* status: 0 of 2 files reviewed, all discussions resolved

PR #4207, opened 2024-12-19
sstable/blob: new package

jbowens:
Add blob FileWriter type for writing out blob files, consisting of a series of
value blocks, a value block index and a footer.

Add the initial, largely unimplemented stencil of a FileReader type.

Informs #112.

### DIFF
--- a/sstable/blob/blob.go
+++ b/sstable/blob/blob.go
@@ -1,0 +1,365 @@
+// Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package blob
+
+import (
+	"context"
+	"encoding/binary"
+	"sync"
+
+	"github.com/cockroachdb/errors"
+	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/objstorage"
+	"github.com/cockroachdb/pebble/sstable/block"
+	"github.com/cockroachdb/pebble/sstable/valblk"
+	"github.com/cockroachdb/redact"
+)
+
+var (
+	errClosed = errors.New("blob: writer closed")
+)
+
+// FileFormat identifies the format of a blob file.
+type FileFormat uint8
+
+// String implements the fmt.Stringer interface.
+func (f FileFormat) String() string {
+	switch f {
+	case FileFormatV1:
+		return "blobV1"
+	default:
+		return "unknown"
+	}
+}
+
+const (
+	// FileFormatV1 is the first version of the blob file format.
+	FileFormatV1 FileFormat = 1
+)
+
+const (
+	fileFooterLength = 29
+	fileMagic        = "\xf0\x9f\xaa\xb3\xf0\x9f\xa6\x80" // 🪳🦀
+)
+
+// FileWriterOptions are used to configure the FileWriter.
+type FileWriterOptions struct {
+	Compression   block.Compression
+	ChecksumType  block.ChecksumType
+	FlushGovernor block.FlushGovernor
+}
+
+// FileWriterStats aggregates statistics about a blob file written by a
+// FileWriter.
+type FileWriterStats struct {
+	BlockCount             uint32
+	ValueCount             uint32
+	BlockLenLongest        uint64
+	UncompressedValueBytes uint64
+	FileLen                uint64
+}
+
+// Handle describes the location of a value stored within a blob file.
+type Handle struct {
+	FileNum       base.DiskFileNum
+	BlockNum      uint32
+	OffsetInBlock uint32
+	ValueLen      uint32
+}
+
+// String implements the fmt.Stringer interface.
+func (h Handle) String() string {
+	return redact.StringWithoutMarkers(h)
+}
+
+// SafeFormat implements redact.SafeFormatter.
+func (h Handle) SafeFormat(w redact.SafePrinter, _ rune) {
+	w.Printf("(%s,blk%d[%d:%d])",
+		h.FileNum, h.BlockNum, h.OffsetInBlock, h.OffsetInBlock+h.ValueLen)
+}
+
+// A FileWriter writes a blob file.
+type FileWriter struct {
+	fileNum      base.DiskFileNum
+	w            objstorage.Writable
+	b            block.Buffer
+	stats        FileWriterStats
+	flushGov     block.FlushGovernor
+	blockOffsets []uint64
+	err          error
+	checksumType block.ChecksumType
+	writeQueue   struct {
+		wg  sync.WaitGroup
+		ch  chan compressedBlock
+		err error
+	}
+}
+
+type compressedBlock struct {
+	pb block.PhysicalBlock
+	bh *block.BufHandle
+}
+
+// NewFileWriter creates a new FileWriter.
+func NewFileWriter(fn base.DiskFileNum, w objstorage.Writable, opts FileWriterOptions) *FileWriter {
+	fw := writerPool.Get().(*FileWriter)
+	fw.fileNum = fn
+	fw.w = w
+	fw.b.Init(opts.Compression, opts.ChecksumType)
+	fw.flushGov = opts.FlushGovernor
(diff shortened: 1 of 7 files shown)


sumeerbhola (changes requested):
flushing some comments on the first commit.

> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4207#-:-OEb0qamAwOKZ9E_9fLt:b-u6ue98)* status: 2 of 7 files reviewed, 5 unresolved discussions (waiting on @jbowens)

___
*[`sstable/block/compression.go` line 357 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4207#-OEayvQ30bTIeOQwclrU:-OEayvQ30bTIeOQwclrV:b-crh5b3) ([raw file](https://github.com/cockroachdb/pebble/blob/4c43c52461373662695b27e8d4d0ec412da847be/sstable/block/compression.go#L357)):*
> ```Go
> 
> // CompressAndChecksum compresses and checksums the block data, returning a
> // PhysicalBuffer that is owned by the caller.  If non-nil, the returned
> ```

should this say `PhysicalBlock` instead of `PhysicalBuffer`?
I am guessing that the slice in `PhysicalBlock` is backed by `BufHandle` but it isn't explicit in this code comment.
___
*[`sstable/block/compression.go` line 377 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4207#-OEb-5Xk9KE08zaAfZGK:-OEb-5Xk9KE08zaAfZGL:be440tw) ([raw file](https://github.com/cockroachdb/pebble/blob/4c43c52461373662695b27e8d4d0ec412da847be/sstable/block/compression.go#L377)):*
> ```Go
> 	// Use the compressedBuf as the new b.h.
> 	pbHandle := b.h
> 	b.h = compressedBuf
> ```

This is a bit different from what we used to do in `valblk.Writer.compressAndFlush` in that given the difference in sizes, it was trying to use the compressed and uncompressed `*blockBuffer`s separate, and not mix the pools.
Perhaps that wasn't necessary, but I am curious about the motivation.
___
*[`sstable/block/compression.go` line 404 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4207#-OEaxJzy5ffK9jBknL2R:-OEaxJzy5ffK9jBknL2S:b-m77kx7) ([raw file](https://github.com/cockroachdb/pebble/blob/4c43c52461373662695b27e8d4d0ec412da847be/sstable/block/compression.go#L404)):*
> ```Go
> // Release releases the buffer back to the pool for reuse.
> func (h *BufHandle) Release() {
> 	if invariants.Enabled && (h.pool == nil) != (h.b == nil) {
> ```

this suggests we have an invariant `h != nil`, so the next `if` predicate shouldn't need to do `h != nil`.
___
*[`sstable/block/compression.go` line 410 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4207#-OEaxXwyE7gZtVVzc9f-:-OEaxXwyE7gZtVVzc9f0:b-xdns1x) ([raw file](https://github.com/cockroachdb/pebble/blob/4c43c52461373662695b27e8d4d0ec412da847be/sstable/block/compression.go#L410)):*
> ```Go
> 	// maximum to the pool. This avoids holding on to occassional large buffers
> 	// necesary for, for example, single large values.
> 	if h != nil && h.b != nil && (h.pool.Max == 0 || len(h.b) < h.pool.Max) {
> ```

is there a reason to allow `Max == 0` to mean unlimited, given we don't use it?
___
*[`sstable/block/compression.go` line 450 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4207#-OEawo9_01nls_y65oYR:-OEawo9_01nls_y65oYS:bcm257) ([raw file](https://github.com/cockroachdb/pebble/blob/4c43c52461373662695b27e8d4d0ec412da847be/sstable/block/compression.go#L450)):*
> ```Go
> 		return v.(*BufHandle)
> 	}
> 	return &BufHandle{b: make([]byte, 0, max(p.Default, 1024)), pool: p}
> ```

Do we really need this `max` in that we are only creating two of these `bufferSyncPool`s with reasonable defaults. And if someone uses a smaller default, they potentially have a good reason for it.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4207)* status: 1 of 7 files reviewed, 5 unresolved discussions (waiting on @sumeerbhola)

___
*[`sstable/block/compression.go` line 357 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4207#-OEayvQ30bTIeOQwclrU:-OFbwP377LQ0464mirM_:b-fmpgz6) ([raw file](https://github.com/cockroachdb/pebble/blob/4c43c52461373662695b27e8d4d0ec412da847be/sstable/block/compression.go#L357)):*

should this say `PhysicalBlock` instead of `PhysicalBuffer`?
I am guessing that the slice in `PhysicalBlock` is backed by `BufHandle` but it isn't explicit in this code comment.
</blockquote></details>

yeah, Done
___
*[`sstable/block/compression.go` line 377 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4207#-OEb-5Xk9KE08zaAfZGK:-OFbwg-qCUvwo3_tJX-y:buz8nwd) ([raw file](https://github.com/cockroachdb/pebble/blob/4c43c52461373662695b27e8d4d0ec412da847be/sstable/block/compression.go#L377)):*

This is a bit different from what we used to do in `valblk.Writer.compressAndFlush` in that given the difference in sizes, it was trying to use the compressed and uncompressed `*blockBuffer`s separate, and not mix the pools.
Perhaps that wasn't necessary, but I am curious about the motivation.
</blockquote></details>

I was thinking since we know compression was _not_ fruitful now, the size of this compressed buf is roughly the size of the uncompressed buffers and it's not worth retrieving a new uncompressed buf from the pool.

The `BufHandle` remembers the pool it originated from, so when we're done with the buf it will end up being returned to the compressed buf pool.
___
*[`sstable/block/compression.go` line 404 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4207#-OEaxJzy5ffK9jBknL2R:-OFbzJCz422l4VS8MykB:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/4c43c52461373662695b27e8d4d0ec412da847be/sstable/block/compression.go#L404)):*

this suggests we have an invariant `h != nil`, so the next `if` predicate shouldn't need to do `h != nil`.
</blockquote></details>

Done.
___
*[`sstable/block/compression.go` line 410 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4207#-OEaxXwyE7gZtVVzc9f-:-OFc8IAN8njqChdWiKnh:b-16a0j4) ([raw file](https://github.com/cockroachdb/pebble/blob/4c43c52461373662695b27e8d4d0ec412da847be/sstable/block/compression.go#L410)):*

is there a reason to allow `Max == 0` to mean unlimited, given we don't use it?
</blockquote></details>

the intention was to make the zero-value `bufferSyncPool` usable, but I think a better approach is probably to assert that `h.pool.Max`
___
*[`sstable/block/compression.go` line 450 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4207#-OEawo9_01nls_y65oYR:-OFc9v0K3vPvEHa_lIcu:b9lhkmw) ([raw file](https://github.com/cockroachdb/pebble/blob/4c43c52461373662695b27e8d4d0ec412da847be/sstable/block/compression.go#L450)):*

Do we really need this `max` in that we are only creating two of these `bufferSyncPool`s with reasonable defaults. And if someone uses a smaller default, they potentially have a good reason for it.
</blockquote></details>

Replaced with a non-zero assertion

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 6 files at r3, 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4207#-:-OG1GTv56fR3urHUe2iT:bg7igkv)* status: all files reviewed, 5 unresolved discussions (waiting on @jbowens)

jbowens:
TFTR!

PR #4208, opened 2024-12-20
db: improve pprof labels for compactions

RaduBerinde:
We add a label for the compaction output level.

We also remove the "table-cache" label which is not very useful - you can just focus on `load() / releaseLoop()` to get the same information. This label "erases" the top-level labels because the context is not plumbed correctly.

Below is an example from sysbench (using pprof flag `-tagroot pebble,output-level`). This suggests that we could enable `zstd` only on L6 and see ~4x less perf impact.

image

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -36,8 +36,15 @@ var errEmptyTable = errors.New("pebble: empty table")
 // concurrent excise or ingest-split operation.
 var ErrCancelledCompaction = errors.New("pebble: compaction cancelled by a concurrent operation, will retry compaction")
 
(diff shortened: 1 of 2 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4208#-:-OEX-Tzm4_aKuyQxzl_E:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jbowens)

RaduBerinde:
TFTR!

PR #4209, opened 2024-12-22
invariants: fix non-inlined empty function call

RaduBerinde:
#### invariants: fix non-inlined empty function call

The `invariants.Sometimes()` was not being inlined even though it
resolves to a `return false` when invariants aren't on. It looks like
the inline decision happens before some phases of dead code
elimination. This function is called in some very hot paths inside the
columnar block builder.

This change moves `Sometimes()` into the `on.go/off.go` files where it
can be simplified for the `off` case. We also add some
`invariants.Enabled` and `invariants.UseFinalizers` checks (which are
constant, unlike function calls) in hot paths.

```
CockroachDataBlockWriter/AlphaLen=4,Prefix=8,Shared=2,KeysPerPrefix=2,valueLen=8-10  75.3µs ± 1%  74.4µs ± 0%  -1.19%  (p=0.000 n=8+8)
```

### DIFF
--- a/internal/invalidating/iter.go
+++ b/internal/invalidating/iter.go
@@ -15,7 +15,7 @@ import (
 // MaybeWrapIfInvariants wraps some iterators with an invalidating iterator.
 // MaybeWrapIfInvariants does nothing in non-invariant builds.
 func MaybeWrapIfInvariants(iter base.InternalIterator) base.InternalIterator {
-	if invariants.Sometimes(10) {
+	if invariants.Enabled && invariants.Sometimes(10) {
 		return NewIter(iter)
 	}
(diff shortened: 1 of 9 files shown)


jbowens (approved):
:lgtm:

> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4209#-:-OFNNaRV2O4G39C2nfQm:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aadityasondhi)

RaduBerinde:
TFTR!

PR #4213, opened 2024-12-26
tool: add db upgrade command

RaduBerinde:
Add a `db upgrade` command that ratches up the DB version to the
newest one.

### DIFF
--- a/tool/db.go
+++ b/tool/db.go
@@ -34,6 +34,7 @@ import (
 type dbT struct {
 	Root       *cobra.Command
 	Check      *cobra.Command
+	Upgrade    *cobra.Command
 	Checkpoint *cobra.Command
 	Get        *cobra.Command
 	Logs       *cobra.Command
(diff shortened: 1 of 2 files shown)


RaduBerinde:
This will be backported to the `pebble-release-1.1` branch.

jbowens (approved):
In any documentation that refers to this, we should stress that users should use their own binary (not cmd/pebble) that uses `pebble/tool` and configures the appropriate comparers, merger, etc if they use any non-default implementations.

:lgtm:

> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4213#-:-OFOQm_G-lAE9AT7zuL4:b-vixzbf)* status: 1 of 2 files reviewed, all discussions resolved

RaduBerinde:
Good point.. The vanilla command should fail if the comparer/merger doesn't match, right?

jbowens:
Yeah, `pebble.Open` should fail if the OPTIONS file comparer / merger doesn't match. If they have custom block property collectors, any sstables compacted during the format major version upgrades won't have the relevant block properties.

RaduBerinde:
Thanks. Added a WARNING. I will make sure to note this in the README update.

PR #4224, opened 2025-01-03
compaction: fix release of compaction slots

itsbilal:
Previously, we did not release a base.CompactionSlot after move, copy or delete-only compactions, resulting in huge leaks in slot utilization in any non-noop implementation of slots. This change fixes that.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -330,7 +330,11 @@ func (c *compaction) userKeyBounds() base.UserKeyBounds {
 }
 
 func newCompaction(
-	pc *pickedCompaction, opts *Options, beganAt time.Time, provider objstorage.Provider,
+	pc *pickedCompaction,
+	opts *Options,
+	beganAt time.Time,
+	provider objstorage.Provider,
+	slot base.CompactionSlot,
 ) *compaction {
 	c := &compaction{
 		kind:              compactionKindDefault,
@@ -347,7 +351,7 @@ func newCompaction(
 		maxOutputFileSize: pc.maxOutputFileSize,
 		maxOverlapBytes:   pc.maxOverlapBytes,
 		pickerMetrics:     pc.pickerMetrics,
-		slot:              pc.slot,
+		slot:              slot,
 	}
 	c.startLevel = &c.inputs[0]
 	if pc.startLevel.l0SublevelInfo != nil {
@@ -607,11 +611,15 @@ func newFlush(
 		updatePointBounds(f.newIter(nil))
 		if rangeDelIter := f.newRangeDelIter(nil); rangeDelIter != nil {
 			if err := updateRangeBounds(rangeDelIter); err != nil {
+				c.slot.Release(0)
+				c.slot = nil
 				return nil, err
 			}
 		}
 		if rangeKeyIter := f.newRangeKeyIter(nil); rangeKeyIter != nil {
 			if err := updateRangeBounds(rangeKeyIter); err != nil {
+				c.slot.Release(0)
+				c.slot = nil
 				return nil, err
 			}
 		}
@@ -1040,6 +1048,9 @@ func (d *DB) addInProgressCompaction(c *compaction) {
 // have completed by this point.
 func (d *DB) clearCompactingState(c *compaction, rollback bool) {
 	c.versionEditApplied = true
+	if c.slot != nil {
+		panic("pebble: compaction slot should have been released before clearing compacting state")
+	}
 	for _, cl := range c.inputs {
 		iter := cl.files.Iter()
 		for f := iter.First(); f != nil; f = iter.Next() {
@@ -1224,6 +1235,10 @@ func (d *DB) runIngestFlush(c *compaction) (*manifest.VersionEdit, error) {
 	if len(c.flushing) != 1 {
 		panic("pebble: ingestedFlushable must be flushed one at a time.")
 	}
+	defer func() {
+		c.slot.Release(0 /* totalBytesWritten */)
+		c.slot = nil
+	}()
 
 	// Construct the VersionEdit, levelMetrics etc.
 	c.metrics = make(map[int]*LevelMetrics, numLevels)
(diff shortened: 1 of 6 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4224)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @aadityasondhi, @itsbilal, and @sumeerbhola)

___
*[`compaction.go` line 546 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4224#-OFh02Wj4adZ5iDgTsWE:-OFh02Wj4adZ5iDgTsWF:bcgvsw0) ([raw file](https://github.com/cockroachdb/pebble/blob/e43e068e87992bb3e531783854c3fe0e3ff2eda8/compaction.go#L546)):*
> ```Go
> 			}
> 			c.kind = compactionKindIngestedFlushable
> 			return c, nil
> ```

Who will release this slot? `runIngestFlush` doesn't.
___
*[`compaction.go` line 606 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4224#-OFgz3QmAdAt5qc90s7_:-OFgz3QnBGg4OiuBD6AD:b-45iboj) ([raw file](https://github.com/cockroachdb/pebble/blob/e43e068e87992bb3e531783854c3fe0e3ff2eda8/compaction.go#L606)):*
> ```Go
> 
> 	var flushingBytes uint64
> 	for i := range flushing {
> ```

[nit] Feels like this block (along with the two functions) could be a separate method or function.
___
*[`compaction.go` line 2806 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4224#-OFgzgQV1wYmUFOmMZkC:-OFgzgQV1wYmUFOmMZkD:b98c2u9) ([raw file](https://github.com/cockroachdb/pebble/blob/e43e068e87992bb3e531783854c3fe0e3ff2eda8/compaction.go#L2806)):*
> ```Go
> ) (ve *versionEdit, stats compact.Stats, retErr error) {
> 	defer func() {
> 		c.slot.Release(stats.CumulativeWrittenSize)
> ```

Releasing in another part of the code than where we acquire is fragile and usually leads to bugs.

We could acquire the slot in `tryScheduleAutoCompaction`, before calling `pickFunc` and release inside the `go`

Similarly, we could acquire before the call to `newFlush` and defer the release there.

RaduBerinde:
*[`compaction.go` line 2806 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4224#-OFgzgQV1wYmUFOmMZkC:-OFh0tK18-firnBRcxxA:bh7uar) ([raw file](https://github.com/cockroachdb/pebble/blob/e43e068e87992bb3e531783854c3fe0e3ff2eda8/compaction.go#L2806)):*

Releasing in another part of the code than where we acquire is fragile and usually leads to bugs.

We could acquire the slot in `tryScheduleAutoCompaction`, before calling `pickFunc` and release inside the `go`

Similarly, we could acquire before the call to `newFlush` and defer the release there.
</blockquote></details>

I see that `pickFunc` has some non-trivial logic about how to acquire the slot. We could just return it from `pickFunc` as a second return (to make it clear that the caller *has* to do something with it) and release it in `tryScheduleAutoCompaction`.

RaduBerinde:
*[`compaction.go` line 2806 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4224#-OFgzgQV1wYmUFOmMZkC:-OFh1LEc2V2MlolvpKIb:belqz6q) ([raw file](https://github.com/cockroachdb/pebble/blob/e43e068e87992bb3e531783854c3fe0e3ff2eda8/compaction.go#L2806)):*

I see that `pickFunc` has some non-trivial logic about how to acquire the slot. We could just return it from `pickFunc` as a second return (to make it clear that the caller *has* to do something with it) and release it in `tryScheduleAutoCompaction`.
</blockquote></details>

It's also conceivable that we would acquire after `pickAuto`.

sumeerbhola (dismissed):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4224)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @aadityasondhi, @itsbilal, and @RaduBerinde)

___
*[`compaction.go` line 2806 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4224#-OFgzgQV1wYmUFOmMZkC:-OG08x2P417l2zNPDBI-:b-x6ihhv) ([raw file](https://github.com/cockroachdb/pebble/blob/e43e068e87992bb3e531783854c3fe0e3ff2eda8/compaction.go#L2806)):*

It's also conceivable that we would acquire after `pickAuto`.
</blockquote></details>

I have a similar concern about releasing the slot.

The code follows a pattern where a `CompactionSlot` gets acquired in one place and stashed into something. For example, `pickedCompaction`, and then moved to another struct, `compaction`. I can see why lower-level pieces of code that use a `compaction` struct needs to know about `CompactionSlot` since they potentially needs to call methods other than `CompactionSlot.Release`. But that leaves us exposed to the possibility that someone forgets to use `CompactionSlot`, and worse, forgets to call `Release`. We could try to protect ourselves by adding some checking -- for instance, anyone who releases the `CompactionSlot` also sets `compaction.slot` to nil. Then higher level functions, like `DB.Compact` and `DB.flush1` can assert that `compaction.slot` is nil.

itsbilal (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4224#-:-OG2dktb90I9ukCRA5K7:b-814g8k)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @aadityasondhi, @RaduBerinde, and @sumeerbhola)

___
*[`compaction.go` line 546 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4224#-OFh02Wj4adZ5iDgTsWE:-OG1sXBx3YewXX8SbuRd:bbgpa6v) ([raw file](https://github.com/cockroachdb/pebble/blob/e43e068e87992bb3e531783854c3fe0e3ff2eda8/compaction.go#L546)):*

Who will release this slot? `runIngestFlush` doesn't.
</blockquote></details>

It is being released there?
___
*[`compaction.go` line 2806 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4224#-OFgzgQV1wYmUFOmMZkC:-OG1sZWp3SdaSEPl6HuV:bkxdc36) ([raw file](https://github.com/cockroachdb/pebble/blob/e43e068e87992bb3e531783854c3fe0e3ff2eda8/compaction.go#L2806)):*

I have a similar concern about releasing the slot.

The code follows a pattern where a `CompactionSlot` gets acquired in one place and stashed into something. For example, `pickedCompaction`, and then moved to another struct, `compaction`. I can see why lower-level pieces of code that use a `compaction` struct needs to know about `CompactionSlot` since they potentially needs to call methods other than `CompactionSlot.Release`. But that leaves us exposed to the possibility that someone forgets to use `CompactionSlot`, and worse, forgets to call `Release`. We could try to protect ourselves by adding some checking -- for instance, anyone who releases the `CompactionSlot` also sets `compaction.slot` to nil. Then higher level functions, like `DB.Compact` and `DB.flush1` can assert that `compaction.slot` is nil.
</blockquote></details>

Good catch! I've moved the slot out of pc entirely, so now responsibility for closing it is entirely within `*compaction`. I've also added a check in `clearCompactingState` to ensure the slot has been nil'd away.  That should allay a lot of these concerns - PTAL!

RaduBerinde (approved):
:lgtm:

I still think it would be best not to have to stash the slot anywhere and defer the release where we call `d.compact()` and `newFlush`, but we can also explore that later.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4224#-:-OG5imNW-Qv-_DHyH4kc:bbbenn0)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @aadityasondhi and @itsbilal)

___
*[`compaction.go` line 546 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4224#-OFh02Wj4adZ5iDgTsWE:-OG5iZeW-BaBZye3qTpC:bclgwp8) ([raw file](https://github.com/cockroachdb/pebble/blob/e43e068e87992bb3e531783854c3fe0e3ff2eda8/compaction.go#L546)):*

It is being released there?
</blockquote></details>

Ah, I missed the diff.

PR #4229, opened 2025-01-06
cache: use key type internally

RaduBerinde:
Minor cleanup to the cache code:
 - use the `key` type in the shard interface
 - remove the `fileKey` type that's not used directly
 - use fibonacci hash for shard hash function

### DIFF
--- a/internal/cache/block_map.go
+++ b/internal/cache/block_map.go
@@ -15,6 +15,7 @@ import (
 )
 
 func fibonacciHash(k *key, seed uintptr) uintptr {
+	// See https://probablydance.com/2018/06/16/fibonacci-hashing-the-optimization-that-the-world-forgot-or-a-better-alternative-to-integer-modulo/
 	const m = 11400714819323198485
 	h := uint64(seed)
 	h ^= uint64(k.id) * m
--- a/internal/cache/clockpro.go
+++ b/internal/cache/clockpro.go
@@ -32,15 +32,40 @@ import (
 	"github.com/cockroachdb/pebble/internal/invariants"
 )
 
-type fileKey struct {
+// key is associated with a specific block.
+type key struct {
 	// id is the namespace for fileNums.
 	id      ID
 	fileNum base.DiskFileNum
+	offset  uint64
 }
 
-type key struct {
-	fileKey
-	offset uint64
+func makeKey(id ID, fileNum base.DiskFileNum, offset uint64) key {
+	return key{
+		id:      id,
+		fileNum: fileNum,
+		offset:  offset,
+	}
+}
+
+// shardIdx determines the shard index for the given key.
+func (k *key) shardIdx(numShards int) int {
+	if k.id == 0 {
+		panic("pebble: 0 cache ID is invalid")
+	}
+	// Same as fibonacciHash() but without the cast to uintptr.
+	const m = 11400714819323198485
+	h := uint64(k.id) * m
+	h ^= uint64(k.fileNum) * m
+	h ^= k.offset * m
+
+	// We need a 32-bit value below; we use the upper bits as per
+	// https://probablydance.com/2018/06/16/fibonacci-hashing-the-optimization-that-the-world-forgot-or-a-better-alternative-to-integer-modulo/
+	h >>= 32
+
+	// This is a better alternative to (h % numShards); see
+	// https://lemire.me/blog/2016/06/27/a-fast-alternative-to-the-modulo-reduction/
+	return int(h * uint64(numShards) >> 32)
 }
 
 // file returns the "file key" for the receiver. This is the key used for the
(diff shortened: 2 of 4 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4229#-:-OG22Ubf0PR3X94on4FM:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`internal/cache/clockpro.go` line 57 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4229#-OG2-DNGDHy8YEMa1zEI:-OG2-DNGDHy8YEMa1zEJ:b9yqthi) ([raw file](https://github.com/cockroachdb/pebble/blob/72815d8449e5106fa24b3c6e03b2c26a29c8c6f5/internal/cache/clockpro.go#L57)):*
> ```Go
> 	}
> 	// Same as fibonacciHash() but without the cast to uintptr.
> 	const m = 11400714819323198485
> ```

So are we switching from FNV to Fibonacci hashing? I don't remember the details in https://probablydance.com/2018/06/16/fibonacci-hashing-the-optimization-that-the-world-forgot-or-a-better-alternative-to-integer-modulo/ but I was under the impression that FNV was a better hash function. We do use Fibonacci hashing is some cases now, but they seem mainly related to invariant testing. I suppose if we think it is good enough for `blockMap` it should be good enough here? All the other cases have a uintptr `seed` -- I am guessing that isn't very random input since pointers from the same region of memory will share a lot of bits.
___
*[`internal/cache/clockpro.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4229#-OG21e_L2OiRpDJtXMaE:-OG21e_L2OiRpDJtXMaF:b-nkkoin) ([raw file](https://github.com/cockroachdb/pebble/blob/72815d8449e5106fa24b3c6e03b2c26a29c8c6f5/internal/cache/clockpro.go#L62)):*
> ```Go
> 	h ^= k.offset * m
> 	// See https://lemire.me/blog/2016/06/27/a-fast-alternative-to-the-modulo-reduction/
> 	return int((h >> 32) * uint64(numShards) >> 32)
> ```

is there a reason we are using the higher 32 bits of h instead of the lower 32 bits?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4229)* status: all files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`internal/cache/clockpro.go` line 57 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4229#-OG2-DNGDHy8YEMa1zEI:-OG23rAI34I5Wefsj2fn:b-t84d1u) ([raw file](https://github.com/cockroachdb/pebble/blob/72815d8449e5106fa24b3c6e03b2c26a29c8c6f5/internal/cache/clockpro.go#L57)):*

So are we switching from FNV to Fibonacci hashing? I don't remember the details in https://probablydance.com/2018/06/16/fibonacci-hashing-the-optimization-that-the-world-forgot-or-a-better-alternative-to-integer-modulo/ but I was under the impression that FNV was a better hash function. We do use Fibonacci hashing is some cases now, but they seem mainly related to invariant testing. I suppose if we think it is good enough for `blockMap` it should be good enough here? All the other cases have a uintptr `seed` -- I am guessing that isn't very random input since pointers from the same region of memory will share a lot of bits.
</blockquote></details>

We are using fib for our block maps. If it's good enough for that, I figure it's good enough for the shards..
___
*[`internal/cache/clockpro.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4229#-OG21e_L2OiRpDJtXMaE:-OG244ua6Dq4q_2cAdeT:bdcxqwm) ([raw file](https://github.com/cockroachdb/pebble/blob/72815d8449e5106fa24b3c6e03b2c26a29c8c6f5/internal/cache/clockpro.go#L62)):*

is there a reason we are using the higher 32 bits of h instead of the lower 32 bits?
</blockquote></details>

My understanding from the blog post is that the higher bits are the better ones to use. For one, if a value is a multiple of a power of 2, the result of the multiply is also a multiple of that power of 2.

sumeerbhola:
> My understanding from the blog post is that the higher bits are the better ones to use. For one, if a value is a multiple of a power of 2, the result of the multiply is also a multiple of that power of 2.

Could you add a code comment?

RaduBerinde (commented):
Done, TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4229#-:-OG2ZfAO3RKhuy7OgUeS:bexegmu)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4229#-:-OG2a__c0ZVs3F4Wv67W:bg7igkv)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @RaduBerinde)

PR #4230, opened 2025-01-06
cockroachkvs: add FormatKey, ParseFormattedKey

jbowens:
Add functions to format cockroach keys into human-readable strings and parse the human-readable strings back out into the binary format. This will be used by the metamorphic test when it serializes operations.

For parity with the Cockroach repo's key pretty-printing, timestamps use a formatting matching the hlc timestamp's pretty printing format.

### DIFF
--- a/cockroachkvs/cockroachkvs.go
+++ b/cockroachkvs/cockroachkvs.go
@@ -10,8 +10,10 @@ import (
 	"bytes"
 	"cmp"
 	"encoding/binary"
+	"encoding/hex"
 	"fmt"
 	"io"
+	"strconv"
 	"strings"
 	"unsafe"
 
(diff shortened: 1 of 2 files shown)


jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4230#-:-OG0fbEXANjDD67Y5qEj:b-814g8k)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @itsbilal)

___
*[`cockroachkvs/cockroachkvs_test.go` line 711 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4230#-OG0Stn35Tn4p2txtX7K:-OG0f_y7DJquJ9ow0lSV:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/210ac51e5c0eb0c5463c3a619c48aabd1622c564/cockroachkvs/cockroachkvs_test.go#L711)):*

What happens when we parse `foo@01000000`? If we allow it, it won't roundtrip. Feels like we should be erroring out for such a case (in which case we should have a test for that)
</blockquote></details>

Done

PR #4231, opened 2025-01-07
cache: remove tryRemoveState in readShard

RaduBerinde:
This PR is only for the last commit (first one is https://github.com/cockroachdb/pebble/pull/4229).

#### cache: remove tryRemoveState in readShard

This structure is not necessary because it stores fields from
`readEntry` that don't change. What's important is not that we read
them under the lock, but that we read them before we release the
refcount; we can do this in the same method where they are used.

### DIFF
--- a/internal/cache/read_shard.go
+++ b/internal/cache/read_shard.go
@@ -208,14 +208,13 @@ func (e *readEntry) waitForReadPermissionOrHandle(
 		e.mu.readStart = time.Now()
 	}
 	unlockAndUnrefAndTryRemoveFromMap := func(readLock bool) (errorDuration time.Duration) {
-		removeState := e.makeTryRemoveStateLocked()
 		errorDuration = e.mu.errorDuration
 		if readLock {
 			e.mu.RUnlock()
 		} else {
 			e.mu.Unlock()
 		}
-		unrefAndTryRemoveFromMap(removeState)
+		e.unrefAndTryRemoveFromMap()
 		return errorDuration
 	}
 
@@ -267,26 +266,9 @@ func (e *readEntry) waitForReadPermissionOrHandle(
 	}
 }
 
-// tryRemoveState captures the state needed by tryRemoveFromMap. The caller
-// constructs it before calling unrefAndTryRemoveFromMap since it typically
-// held readEntry.mu, which avoids acquiring it again in
-// unrefAndTryRemoveFromMap.
-type tryRemoveState struct {
-	rs *readShard
-	k  key
-	e  *readEntry
-}
-
-// makeTryRemoveStateLocked initializes tryRemoveState.
-func (e *readEntry) makeTryRemoveStateLocked() tryRemoveState {
-	return tryRemoveState{
-		rs: e.readShard,
-		k:  e.key,
-		e:  e,
-	}
-}
-
-// unrefAndTryRemoveFromMap tries to remove s.k => s.e from the map in s.rs.
+// unrefAndTryRemoveFromMap reduces the reference count of e and removes e.key
+// => e from the readMap if necessary.
+//
 // It is possible that after unreffing that s.e has already been removed, and
 // is now back in the sync.Pool, or being reused (for the same or different
 // key). This is because after unreffing, which caused the s.e.refCount to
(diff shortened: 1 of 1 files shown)


sumeerbhola (changes requested):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4231)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`-- commits` line 15 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4231#-OG249en4z6Mab9Z7PJ-:-OG249en4z6Mab9Z7PJ0:b-h2mgoo):*
nice catch!

Not sure what I was thinking given that I wrote comments like
```
// tryRemoveState captures the state needed by tryRemoveFromMap. The caller
// constructs it before calling unrefAndTryRemoveFromMap since it typically
// held readEntry.mu, which avoids acquiring it again in
// unrefAndTryRemoveFromMap.
```
and then proceeded to grab the members that were not protected by `readEntry.mu`.
___
*[`internal/cache/read_shard.go` line 269 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4231#-OG22yFa7EpNlg_Q706Y:-OG22yFa7EpNlg_Q706Z:bwof1af) ([raw file](https://github.com/cockroachdb/pebble/blob/73f846f82a4b4dd14ef0ae14b130013d7719afbb/internal/cache/read_shard.go#L269)):*
> ```Go
> }
> 
> // unrefAndTryRemoveFromMap tries to remove s.k => s.e from the map in s.rs.
> ```

This comment needs updating.
___
*[`internal/cache/read_shard.go` line 305 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4231#-OG254DD7ewnKEFM2du9:-OG254DD7ewnKEFM2duA:b-wjp5i7) ([raw file](https://github.com/cockroachdb/pebble/blob/73f846f82a4b4dd14ef0ae14b130013d7719afbb/internal/cache/read_shard.go#L305)):*
> ```Go
> 
> 	// Free s.e.
> 	e.mu.v.release()
> ```

why is the code no longer calling `e.mu.Lock()` given `e.mu.v` is protected by `e.mu`?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4231#-:-OG2_jvz3Dj0uKi53v_K:b-814g8k)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @sumeerbhola)

___
*[`-- commits` line 15 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4231#-OG249en4z6Mab9Z7PJ-:-OG2ZlJq8UG7Mxg9k6RY:b-xoo90i):*

nice catch!

Not sure what I was thinking given that I wrote comments like
```
// tryRemoveState captures the state needed by tryRemoveFromMap. The caller
// constructs it before calling unrefAndTryRemoveFromMap since it typically
// held readEntry.mu, which avoids acquiring it again in
// unrefAndTryRemoveFromMap.
```
and then proceeded to grab the members that were not protected by `readEntry.mu`.
</blockquote></details>

I'm guessing there was an earlier iteration where that was accurate. It looks like this whole thing was pretty tricky to get right.
___
*[`internal/cache/read_shard.go` line 269 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4231#-OG22yFa7EpNlg_Q706Y:-OG2aBIm41bAJ7aIPNjK:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/73f846f82a4b4dd14ef0ae14b130013d7719afbb/internal/cache/read_shard.go#L269)):*

This comment needs updating.
</blockquote></details>

Done.
___
*[`internal/cache/read_shard.go` line 305 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4231#-OG254DD7ewnKEFM2du9:-OG2_-Dh8mIbRT5Asj0a:b-kc2ofy) ([raw file](https://github.com/cockroachdb/pebble/blob/73f846f82a4b4dd14ef0ae14b130013d7719afbb/internal/cache/read_shard.go#L305)):*

why is the code no longer calling `e.mu.Lock()` given `e.mu.v` is protected by `e.mu`?
</blockquote></details>

Nobody else can hold this entry at this point. It would be very bad if this wasn't the case, given that we zero out the mutex below.

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4231#-:-OG6K6WABLJYp6KhSsxS:bg7igkv)* status: all files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

___
*[`internal/cache/read_shard.go` line 305 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4231#-OG254DD7ewnKEFM2du9:-OG6JnbM0r90nsV4MNQG:b2nhb2c) ([raw file](https://github.com/cockroachdb/pebble/blob/73f846f82a4b4dd14ef0ae14b130013d7719afbb/internal/cache/read_shard.go#L305)):*

BTW I'm a bit surprised the linter lets us zero out the lock, I guess it has a special case for the zero value.
</blockquote></details>

Ack
___
*[`internal/cache/read_shard.go` line 270 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4231#-OG6JtTl4tj0G9tnh7nq:-OG6JtTl4tj0G9tnh7nr:blinhcc) ([raw file](https://github.com/cockroachdb/pebble/blob/19c57f7aa2cd86e54f9bf0100b93dc6c864feb94/internal/cache/read_shard.go#L270)):*
> ```Go
> 
> // unrefAndTryRemoveFromMap reduces the reference count of e and removes e.key
> // => e from the reedMap if necessary.
> ```

nit: readMap

PR #4232, opened 2025-01-07
sstable: fix restarts integer overflow

EdwardX29:
Fix bug with integer overflows while indexing into blocks with large KVs
in SeekGE() and SeekLT() in rowblk_iter. Updated members in blockEntry
that represent  offsets in blocks to be type offsetInBlock (alias for int64).

Added check in rowblk_writer to ensure that block sizes do not exceed
rowblk.MaximumSize before writing more data to the block.

Wrote unit tests to verify correct behavior for SeekGE() and SeekLT() with large
blocks >2GB.

### DIFF
--- a/sstable/rowblk/rowblk_iter.go
+++ b/sstable/rowblk/rowblk_iter.go
@@ -113,10 +113,10 @@ type Iter struct {
 
 	// offset is the byte index that marks where the current key/value is
 	// encoded in the block.
-	offset int32
+	offset offsetInBlock
 	// nextOffset is the byte index where the next key/value is encoded in the
 	// block.
-	nextOffset int32
+	nextOffset offsetInBlock
 	// A "restart point" in a block is a point where the full key is encoded,
 	// instead of just having a suffix of the key encoded. See readEntry() for
 	// how prefix compression of keys works. Keys in between two restart points
@@ -129,7 +129,10 @@ type Iter struct {
 	// 4 bytes of the block as a uint32 (i.ptr[len(block)-4:]). i.restarts can
 	// therefore be seen as the point where data in the block ends, and a list
 	// of offsets of all restart points begins.
-	restarts int32
+	//
+	// int64 is used to prevent overflow and preserve signedness for binary
+	// search invariants.
+	restarts offsetInBlock
 	// Number of restart points in this block. Encoded at the end of the block
 	// as a uint32.
 	numRestarts int32
@@ -192,12 +195,25 @@ type Iter struct {
 	firstUserKeyWithPrefixBuf []byte
 }
 
+// offsetInBlock represents an offset in a block
+//
+// While restart points are serialized as uint32's, it is possible for offsets to
+// be greater than math.MaxUint32 since they may point to an offset after the KVs.
+//
+// Previously, offsets were represented as int32, which causes problems with
+// integer overflows while indexing into blocks (i.data) with large KVs in SeekGE()
+// and SeekLT(). Using an int64 solves the problem of overflows as wraparounds will
+// be prevented. Additionally, the signedness of int64 allows repsentation of
+// iterators that have conducted backward interation and allows for binary search
+// invariants in SeekGE() and SeekLT() to be preserved.
+type offsetInBlock int64
+
 type blockEntry struct {
-	offset   int32
-	keyStart int32
-	keyEnd   int32
-	valStart int32
-	valSize  int32
+	offset   offsetInBlock
+	keyStart offsetInBlock
+	keyEnd   offsetInBlock
+	valStart offsetInBlock
+	valSize  uint32
 }
 
 // *Iter implements the block.DataBlockIterator interface.
@@ -236,7 +252,7 @@ func (i *Iter) Init(
 	i.synthSuffixBuf = i.synthSuffixBuf[:0]
 	i.split = split
 	i.cmp = cmp
-	i.restarts = int32(len(blk)) - 4*(1+numRestarts)
+	i.restarts = offsetInBlock(len(blk)) - 4*(1+offsetInBlock(numRestarts))
 	i.numRestarts = numRestarts
 	i.ptr = unsafe.Pointer(&blk[0])
 	i.data = blk
@@ -395,7 +411,7 @@ func (i *Iter) readEntry() {
 	}
 	ptr = unsafe.Pointer(uintptr(ptr) + uintptr(unshared))
 	i.val = unsafe.Slice((*byte)(ptr), int(value))
-	i.nextOffset = int32(uintptr(ptr)-uintptr(i.ptr)) + int32(value)
+	i.nextOffset = offsetInBlock(uintptr(ptr)-uintptr(i.ptr)) + offsetInBlock(value)
 }
 
 func (i *Iter) readFirstKey() error {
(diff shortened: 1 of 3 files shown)


RaduBerinde (commented):
It would be good to add a test with a very large block and verify we can iterate it.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4232#-:-OG1drex3r7I4G4cHe5M:b-f73hr0)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @EdwardX29 and @jbowens)

___
*[`sstable/rowblk/rowblk_iter.go` line 116 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4232#-OG1e2nA8xUCgMbO8Egy:-OG1e2nA8xUCgMbO8Egz:b690op2) ([raw file](https://github.com/cockroachdb/pebble/blob/69ab5b66047b7b05f91540ce2506683afd6c37f2/sstable/rowblk/rowblk_iter.go#L116)):*
> ```Go
> 	// offset is the byte index that marks where the current key/value is
> 	// encoded in the block.
> 	offset int64
> ```

You can consider adding a `type offsetInBlock int64` which would make it evident what fields are offsets.
___
*[`sstable/rowblk/rowblk_iter.go` line 200 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4232#-OG1dnVR1TDRq5ZSc_9o:-OG1dnVR1TDRq5ZSc_9p:b7dal6j) ([raw file](https://github.com/cockroachdb/pebble/blob/69ab5b66047b7b05f91540ce2506683afd6c37f2/sstable/rowblk/rowblk_iter.go#L200)):*
> ```Go
> type blockEntry struct {
> 	offset   int64
> 	keyStart uint32
> ```

Aren't the rest of these also offsets?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4232)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/rowblk/rowblk_iter.go` line 116 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4232#-OG1e2nA8xUCgMbO8Egy:-OG1kZ-9-VcaJcP1FPMl:bdyhsf) ([raw file](https://github.com/cockroachdb/pebble/blob/69ab5b66047b7b05f91540ce2506683afd6c37f2/sstable/rowblk/rowblk_iter.go#L116)):*

You can consider adding a `type offsetInBlock int64` which would make it evident what fields are offsets.
</blockquote></details>

Good idea—we can add a comment on the type discussing the motivation for using an `int64`, including our expectation that an offset may be > `math.MaxUint32` (even though restart points are serialized as uint32s, because the offset fields are sometimes updated to point _after_ the last KV in the block)
___
*[`sstable/rowblk/rowblk_iter.go` line 200 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4232#-OG1dnVR1TDRq5ZSc_9o:-OG1jbJy9vDcd3f_YUGi:b2mnowh) ([raw file](https://github.com/cockroachdb/pebble/blob/69ab5b66047b7b05f91540ce2506683afd6c37f2/sstable/rowblk/rowblk_iter.go#L200)):*

Aren't the rest of these also offsets?
</blockquote></details>

+1, I think they're all offsets except for `valSize`

jbowens (commented):
> Reviewed 1 of 1 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4232)* status: 2 of 3 files reviewed, 6 unresolved discussions (waiting on @EdwardX29 and @RaduBerinde)

___
*[`sstable/rowblk/rowblk_iter_test.go` line 481 at r8](https://reviewable.io/reviews/cockroachdb/pebble/4232#-OGBXUZtEM9pUq6Hiex1:-OGBXUZtEM9pUq6Hiex2:b36p394) ([raw file](https://github.com/cockroachdb/pebble/blob/99483ca54ed8d9b7daf6c41a2ac302c49cd0c513/sstable/rowblk/rowblk_iter_test.go#L481)):*
> ```Go
> 
> 	var largeKeySize int64 = 2 << 30   // 2GB key size
> 	var largeValueSize int64 = 2 << 30 // 2GB value size
> ```

nit: these can be `const` instead of `var`, and if you omit the `int64`too, then down below we won't have to cast into an `int` when calling `bytes.Repeat`
___
*[`sstable/rowblk/rowblk_iter_test.go` line 534 at r8](https://reviewable.io/reviews/cockroachdb/pebble/4232#-OGBYXdg7dF5LZXKPLtA:-OGBYXdg7dF5LZXKPLtB:bh9n5ob) ([raw file](https://github.com/cockroachdb/pebble/blob/99483ca54ed8d9b7daf6c41a2ac302c49cd0c513/sstable/rowblk/rowblk_iter_test.go#L534)):*
> ```Go
> 	}
> 
> 	KVTestPairs := func() []KVTestPair {
> ```

curious why use an anonymous function here?
___
*[`sstable/rowblk/rowblk_iter_test.go` line 538 at r8](https://reviewable.io/reviews/cockroachdb/pebble/4232#-OGBY0gqCKl0y8a_jAIc:-OGBY0gqCKl0y8a_jAId:bf3vknu) ([raw file](https://github.com/cockroachdb/pebble/blob/99483ca54ed8d9b7daf6c41a2ac302c49cd0c513/sstable/rowblk/rowblk_iter_test.go#L538)):*
> ```Go
> 		for i := 0; i < numKVs; i++ {
> 			key := fmt.Sprintf("key-%04d", i)
> 			value := strings.Repeat("a", valueSize)
> ```

nit: since we're memory constrained, maybe lift this to the outer scope as:
```
value4MB := bytes.Repeat([]byte("a"), valueSize)
```
so that every KV we add is reusing the same constructed byte slice rather than re-creating it
___
*[`sstable/rowblk/rowblk_iter_test.go` line 584 at r8](https://reviewable.io/reviews/cockroachdb/pebble/4232#-OGBZ088EGtEDGFdOJeF:-OGBZ088EGtEDGFdOJeG:b332elg) ([raw file](https://github.com/cockroachdb/pebble/blob/99483ca54ed8d9b7daf6c41a2ac302c49cd0c513/sstable/rowblk/rowblk_iter_test.go#L584)):*
> ```Go
> 	numKVs := 63
> 	valueSize := 4 * (1 << 20)
> 	var FourGB int64 = 4 * (1 << 30)
> ```

nit: use `const`s for those of these that are constant
___
*[`sstable/rowblk/rowblk_iter.go` line 198 at r8](https://reviewable.io/reviews/cockroachdb/pebble/4232#-OGBWUDaD0r74cZYpipe:-OGBWUDaD0r74cZYpipf:b-cr0flb) ([raw file](https://github.com/cockroachdb/pebble/blob/99483ca54ed8d9b7daf6c41a2ac302c49cd0c513/sstable/rowblk/rowblk_iter.go#L198)):*
> ```Go
> }
> 
> // Type representing an offset in a block
> ```

nit: it's idiomatic to start a Go comment on an identifier with the name of the identifier, in this case:

```
// offsetInBlock represents an offset in a block.
```

jbowens (approved):
:lgtm:

Make sure to squash the commits before merging

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4232#-:-OGC596E925DEQwyIQGH:b-j8np3n)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @EdwardX29 and @RaduBerinde)

___
*[`sstable/rowblk/rowblk_iter_test.go` line 481 at r8](https://reviewable.io/reviews/cockroachdb/pebble/4232#-OGBXUZtEM9pUq6Hiex1:-OGBzxfmDmSsWAjXB_4p:bkclf7z) ([raw file](https://github.com/cockroachdb/pebble/blob/99483ca54ed8d9b7daf6c41a2ac302c49cd0c513/sstable/rowblk/rowblk_iter_test.go#L481)):*
<details><summary><i>Previously, EdwardX29 (Edward Xiao) wrote…</i></summary><blockquote>

32 bit architectures are not happy with large constants being used as `int`s in `bytes.Repeat` as constant expressions are evaluated at compile time. Adding a check for `strconv.IntSize == 32` doesn't solve the issue either. [CI run](https://github.com/cockroachdb/pebble/actions/runs/12698201229/job/35396155670?pr=4232). However, using `var` and `int64` solves this issue as overflow is at runtime instead; run [here](https://github.com/cockroachdb/pebble/actions/runs/12698325102/job/35396564648?pr=4232)
</blockquote></details>

Ah, got it.

PR #4233, opened 2025-01-07
metamorphic: move block property collectors into KeyFormat

jbowens:
Move the configuration of block-property collectors into the KeyFormat, since it's dependent on the key format used.

### DIFF
--- a/metamorphic/config.go
+++ b/metamorphic/config.go
@@ -361,6 +361,7 @@ type KeyFormat struct {
 	Name                         string
 	Comparer                     *base.Comparer
 	KeySchema                    *pebble.KeySchema
+	BlockPropertyCollectors      []func() pebble.BlockPropertyCollector
 	FormatKey                    func(UserKey) string
 	FormatKeySuffix              func(UserKeySuffix) string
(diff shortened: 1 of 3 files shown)


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4233#-:-OG1vZubEqjg0Z_Km1r-:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jbowens)

RaduBerinde:
What would you say about - in the long term - moving block property collectors to an internal-only concept? With the new key format, we can internally implement a min/max suffix collector (which is the only one we define outside or Pebble).

jbowens:
> What would you say about - in the long term - moving block property collectors to an internal-only concept? With the new key format, we can internally implement a min/max suffix collector (which is the only one we define outside or Pebble).

Hrm, they are useful more generally, but I'd guess that no open source users actually are using them. One day I'd love to see us apply them for SQL push-down filtering if we could figure out how to thread through the KV interface.

jbowens:
TFTR!

PR #4238, opened 2025-01-08
cache: readShard: don't allow 0->1 refcount transitions

RaduBerinde:
The current code allows a new reader to take the refcount from 0 to 1.
This leads to non-intuitive cases on the release side: once we drop
the refcount to 0, someone might come in, increase the refcount, then
decrease it and free the entry.

We remove this case by adding a "release if not last" primitive to the
refcount and making sure that any transition to 0 happens together
with the removal from the map (while holding the shard mutex).


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4238)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @sumeerbhola)

___
*[`internal/cache/read_shard.go` line 61 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4238#-OGFVaUy6oVIzqxsouq-:-OGFi44q9u7XhhXKvLk-:b-rmafwr) ([raw file](https://github.com/cockroachdb/pebble/blob/9c5cc230c7dbaa2140cddb17caa9a25817103827/internal/cache/read_shard.go#L61)):*

I suppose for maximum concurrency one could create a `readEntry` using a `readShard.mu`, then acquire `shard.mu` in read-mode and confirm that the entry is not populated, and then proceed.

Then `unrefAndTryRemoveFromMap` only needs to acquire `readShard.mu`.
</blockquote></details>

Yes, this would work.  We still need to acquire the shard to call `Set`, but I guess one locks per read block is better than two..
___
*[`internal/cache/read_shard.go` line 277 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4238#-OGFWP1IBbqiCZIadbpO:-OGFgp410t0JTdtO4qQn:b-jkilhy) ([raw file](https://github.com/cockroachdb/pebble/blob/9c5cc230c7dbaa2140cddb17caa9a25817103827/internal/cache/read_shard.go#L277)):*

(Note that I read this after adding the comment in refcnt_normal.go)

I am having trouble wrapping my head around this. The caller has a ref, so why would it not want to release it?

Oh, I just realized something. I had been assuming that this change was switching to a mode where the container also held a ref. But it isn't -- the initial ref is the caller's, when it was added to the map. But then the last releaser seems to hand-off its ref to the container (without decrementing it). A change where the container holds a ref (the more standard technique) would not allow for the simplification in the removal code that is being done here, since multiple concurrent callers can still discover that the ref-count is 1 and then concurrently proceed to acquire the lock and removing (which is why I didn't bother with that in the first place).

What you are doing here is more subtle, and something I haven't seen before. This may be correct, but I don't have a good framework to think about correctness. Say multiple concurrent callers handed off their reference. Ah, it seems they have not handed it off to the container, but the rest of this method, which must unconditionally call `release`. The lock serializes the release calls, only one of which will succeed and also remove from the map. So it seems to ensure at most one will remove. Can there be a case where no one attempts to remove from the map -- say because everyone gets a true return value from `releaseIfNotLast` -- that isn't possible because someone has to observe the ref-count of 1 on its way down to 0.

This is clever. I think it's not a common pattern, so could use a long code comment explaining its correctness.
</blockquote></details>

Thanks, this is correct.

Unfortunately, clever is usually something to stay away from when it comes to synchronization and encryption :)  I haven't seen this pattern before either. I don't feel strongly that this is better than before.

What would definitely be much better is if the ref count is not atomic and always protected by the same mutex the map is. The only difference is this `releaseIfNotLast` case when there are many readers for the same block (I would argue that even if that happens frequently, it can't be frequent enough for acquiring the mutex vs not here to make a big difference). Perhaps if we add a `readShard.mu` and don't have to lock the entire shard, this would be even less of a concern.

sumeerbhola:
> What would definitely be much better is if the ref count is not atomic and always protected by the same mutex the map is.

It is a concurrency reduction of course. I suppose we could simplify it and if the concurrency reduction doesn't show up in mutex profiles we don't have to worry about it.

RaduBerinde:
It only reduces concurrency when we have multiple readers waiting for the same block (for 1 reader, we have to lock anyway).

RaduBerinde:
Let's decide what to do with this:
 1. We close this PR and leave it as is
 2. We switch to using a readShard mutex and protecting the refcount with that mutex
 3. We merge this PR

I'm leaning towards 1 or 2 (maybe a bit more towards 2).

sumeerbhola (commented):
I'm fine with either 1 or 2.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4238#-:-OGfl53S6Y960XQ8r6-B:b-3eacpz)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

RaduBerinde:
Thanks, sounds good. Closing this and I'll put up another PR for 2 and decide there if we want to merge it or not.

PR #4253, opened 2025-01-13
db: add a LowDiskSpace event

RaduBerinde:
Add an event when the low disk space runs low. We post an event when
we go below each threshold (10%, 5%, 2%, 1%) and we keep posting the
event periodically.

Informs https://github.com/cockroachdb/cockroach/issues/129319

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1092,13 +1092,18 @@ func (d *DB) clearCompactingState(c *compaction, rollback bool) {
 }
 
 func (d *DB) calculateDiskAvailableBytes() uint64 {
-	if space, err := d.opts.FS.GetDiskUsage(d.dirname); err == nil {
-		d.diskAvailBytes.Store(space.AvailBytes)
-		return space.AvailBytes
-	} else if !errors.Is(err, vfs.ErrUnsupported) {
-		d.opts.EventListener.BackgroundError(err)
+	space, err := d.opts.FS.GetDiskUsage(d.dirname)
+	if err != nil {
+		if !errors.Is(err, vfs.ErrUnsupported) {
+			d.opts.EventListener.BackgroundError(err)
+		}
+		// Return the last value we managed to obtain.
+		return d.diskAvailBytes.Load()
 	}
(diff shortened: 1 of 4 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4253#-:-OGWXuXE1NRXV1WoTCqH:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`event.go` line 580 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4253#-OGWW7mN1BFvVaPsU3X3:-OGWW7mN1BFvVaPsU3X4:b-w42m7h) ([raw file](https://github.com/cockroachdb/pebble/blob/88b3efbb66d2560824c2ff6dbcd9e86cd1402b91/event.go#L580)):*
> ```Go
> 	switch i.Kind {
> 	case IneffectualSingleDelete, NondeterministicSingleDelete:
> 		w.Printf("possible API misuse: %s (key=%q)", redact.Safe(i.Kind), redact.Safe(i.UserKey))
> ```

This `redact.Safe(i.UserKey)` isn't safe, right? We might leak user data into the string.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4253#-:-OGYE9rz4b0i6f3sznqx:b-814g8k)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`event.go` line 580 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4253#-OGWW7mN1BFvVaPsU3X3:-OGYEATE4rk1I7aXJJGp:bpqj741) ([raw file](https://github.com/cockroachdb/pebble/blob/88b3efbb66d2560824c2ff6dbcd9e86cd1402b91/event.go#L580)):*

This `redact.Safe(i.UserKey)` isn't safe, right? We might leak user data into the string.
</blockquote></details>

Great catch, thanks!

PR #4256, opened 2025-01-13
sstable: revert change to binary search calculation

EdwardX29:
Revert index calculation for binary search in rowblk_iter.go to use casting (like in old iterations).
Binary search was originally modified to factor for int32 overflow, however this is not relevant
because a cast to uint will fix the overflow error. Reverting back to old calculation for
consistency with other of binary search calculations and (presumably) less instructions.


RaduBerinde:
Can you explain the reasoning (in the commit message)?

jbowens (approved):

PR #4259, opened 2025-01-14
sstable: fix row writer mangling of cache block

RaduBerinde:
#### sstable, objstorageprovider: mangle the buffer when writing blocks

Pepper in some invariant-only buffer mangling on block write paths.
With this we can repro the failure in
https://github.com/cockroachdb/cockroach/issues/138662 with the meta
test.

#### sstable: fix row writer mangling of cache block

Fix `RawRowWriter.addDataBlock` to make a copy of the block data when
it's not compressed (similar to the columnar writer), because in that
case the writing out can mangle the buffer.

The sstable copier (during Download) uses `addDataBlock` with a buffer
from the cache, which we are not allowed to modify.

Informs https://github.com/cockroachdb/cockroach/issues/138662

### DIFF
--- a/objstorage/objstorageprovider/vfs_writable.go
+++ b/objstorage/objstorageprovider/vfs_writable.go
@@ -7,6 +7,7 @@ package objstorageprovider
 import (
 	"bufio"
 
+	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/objstorage"
 	"github.com/cockroachdb/pebble/vfs"
 )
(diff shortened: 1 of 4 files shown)


RaduBerinde:
I will need to backport to 24.3.

jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4259#-:-OGaSpEr0hfzTCzIjqRw:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

jbowens:
Good find!

RaduBerinde:
TFTR!

PR #4262, opened 2025-01-15
sstable: expand sstable iterator block stats

EdwardX29:
Added iterator stats for rangekey, rangedel, and value blocks. Replaced block.NoReadEnv in NewRawRangeDelIter, NewRawRangeKeyIterm and ReadValueBlockExternal with an instantiated block.ReadEnv with Stats and IterStats fields. Refactored NewPointIter to accept a block.ReadEnv instead of InternalIteratorStats and IterStatsAccumulator. Removed ChildIterStatsAccumulator from category_stats.go as mutex contention in iterator stat reporting has been empirically tested to represent ~0% of overall mutex time in the worst case scenario. Updated interfaces and function calls to reflect changes.

### DIFF
--- a/cockroachkvs/cockroachkvs_bench_test.go
+++ b/cockroachkvs/cockroachkvs_bench_test.go
@@ -117,7 +117,7 @@ func benchmarkRandSeekInSST(
 	rp := sstable.MakeTrivialReaderProvider(reader)
 	iter, err := reader.NewPointIter(
 		ctx, sstable.NoTransforms, nil, nil, nil, sstable.NeverUseFilterBlock,
-		&stats, nil, rp)
+		block.ReadEnv{Stats: &stats, IterStats: nil}, rp)
 	require.NoError(b, err)
 	n := 0
 	for kv := iter.First(); kv != nil; kv = iter.Next() {
@@ -133,7 +133,7 @@ func benchmarkRandSeekInSST(
 		key := queryKeys[i%numQueryKeys]
 		iter, err := reader.NewPointIter(
 			ctx, sstable.NoTransforms, nil, nil, nil, sstable.NeverUseFilterBlock,
-			&stats, nil, rp)
+			block.ReadEnv{Stats: &stats, IterStats: nil}, rp)
 		if err != nil {
 			b.Fatal(err)
 		}
--- a/external_iterator.go
+++ b/external_iterator.go
@@ -12,6 +12,7 @@ import (
 	"github.com/cockroachdb/pebble/internal/keyspan"
 	"github.com/cockroachdb/pebble/internal/manifest"
 	"github.com/cockroachdb/pebble/sstable"
+	"github.com/cockroachdb/pebble/sstable/block"
 )
 
 // NewExternalIter takes an input 2d array of sstable files which may overlap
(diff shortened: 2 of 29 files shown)


sumeerbhola (changes requested):
I really like this simplification. Comments mainly for potentially later PRs.

> Reviewed 29 of 29 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4262#-:-OGjNMuvDoHRTpAhcr3D:b8qopa2)* status: all files reviewed, 3 unresolved discussions (waiting on @EdwardX29 and @jbowens)

___
*[`sstable/block/category_stats.go` line 172 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4262#-OGjK4AK4ryKh-ZDR2fk:-OGjK4AK4ryKh-ZDR2fl:b-6ujyry) ([raw file](https://github.com/cockroachdb/pebble/blob/9f49b3d0de7971f8e9c2904ef9ca47fa05bfec0a/sstable/block/category_stats.go#L172)):*
> ```Go
> }
> 
> const numCategoryStatsShards = 16
> ```

How many vCPUs in the machine that you experimented with? And what was the workload and qps? I am wondering whether 16 is high enough. And whether it should be a constant.
As comparison, the Pebble block cache (which is shared across all Pebble instances on a node, so in multi-store settings needs more concurrency than these category-stats, which are per store), uses `4 * runtime.GOMAXPROCS(0)`. Perhaps we should make 16 the lower-bound and allow this to be as much as `runtime.GOMAXPROCS(0)`. It can be a followup PR.
___
*[`sstable/block/category_stats.go` line 257 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4262#-OGjLRvc0V1OkbIPXbp6:-OGjLRvd47BqrzdY4bqf:b-qtw3dk) ([raw file](https://github.com/cockroachdb/pebble/blob/9f49b3d0de7971f8e9c2904ef9ca47fa05bfec0a/sstable/block/category_stats.go#L257)):*
> ```Go
> type IterStatsAccumulator interface {
> 	// Accumulate accumulates the provided stats.
> 	Accumulate(blockBytes, blockBytesInCache uint64, blockReadDuration time.Duration)
> ```

Can we get rid of this interface? I can't remember why this existed in the first place. But there is only one implementation and it is in the same package, so seems unnecessary to have the interface overhead.
Maybe there is some plumbing trickiness, since the plumbing of this originates in file_cache.go, but we can export `categoryStatsWithMu` from this package. Again, it can be a mechanical change in a followup PR.
___
*[`testdata/merging_iter` line 53 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4262#-OGjNbtJAQ5U_cdJ7pIL:-OGjNbtJAQ5U_cdJ7pIM:b-xrlh14) ([raw file](https://github.com/cockroachdb/pebble/blob/9f49b3d0de7971f8e9c2904ef9ca47fa05bfec0a/testdata/merging_iter#L53)):*
> ```
> g#20,SET:20
> .
> {BlockBytes:200 BlockBytesInCache:0 BlockReadDuration:0s KeyBytes:4 ValueBytes:8 PointCount:4 PointsCoveredByRangeTombstones:0 SeparatedPointValue:{Count:0 ValueBytes:0 ValueBytesFetched:0}}
> ```

are all the testdata file changes because of the new accounting we weren't doing before?

jbowens (approved):
:lgtm:

all the testdata changes seem to check out!

nice!

> Reviewed 29 of 29 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4262#-:-OGkoouW9aG-qFhCIZ0D:b71gvbs)* status: all files reviewed, 3 unresolved discussions (waiting on @EdwardX29 and @sumeerbhola)

___
*[`sstable/block/category_stats.go` line 257 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4262#-OGjLRvc0V1OkbIPXbp6:-OGkobC06Tw78IajIBAH:b-euca0c) ([raw file](https://github.com/cockroachdb/pebble/blob/9f49b3d0de7971f8e9c2904ef9ca47fa05bfec0a/sstable/block/category_stats.go#L257)):*

Can we get rid of this interface? I can't remember why this existed in the first place. But there is only one implementation and it is in the same package, so seems unnecessary to have the interface overhead.
Maybe there is some plumbing trickiness, since the plumbing of this originates in file_cache.go, but we can export `categoryStatsWithMu` from this package. Again, it can be a mechanical change in a followup PR.
</blockquote></details>

Yeah, I think you're right, we can remove the interface

sumeerbhola (approved):
:lgtm: since the remaining questions can be dealt with in followups.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4262#-:-OGksibBBQBI5REHM5Oq:b-mpprzo)* status: all files reviewed, 3 unresolved discussions (waiting on @EdwardX29)

PR #4265, opened 2025-01-16
sstable: copy filter block during suffix rewriting

jbowens:
Previously, during suffix rewriting, both row and columnar sstable writers
could pass the input sstable's filter block into objstorage.Writable.Write by
passing a subslice of the original input sstables byte slice. The Write method
is permitted to mangle the input buffer, which in this case would corrupt the
original input sstable.

### DIFF
--- a/objstorage/test_utils.go
+++ b/objstorage/test_utils.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"context"
 
+	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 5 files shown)


RaduBerinde (approved):
:lgtm:

This optimization seems to haunt us, I can't imagine root-causing this kind of bug in production. I think we should probably revisit it. I'd imagine applying an encryption primitive is at least a factor slower than a memory copy.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4265#-:-OGkrDt_1xe63h4RASVP:b-8b6xin)* status: 0 of 5 files reviewed, all discussions resolved

jbowens (commented):
Yeah, agreed

TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4265#-:-OGksjBC946eSJJM2Ku0:b7uotwt)* status: 0 of 5 files reviewed, all discussions resolved

PR #4269, opened 2025-01-17
sstable: refactor ReaderOptions

jbowens:
Add block.ReaderOptions and refactor sstable.ReaderOptions to contain them, and the blob reader to also use block.Reader and block.ReaderOptions.

### DIFF
--- a/cockroachkvs/cockroachkvs_bench_test.go
+++ b/cockroachkvs/cockroachkvs_bench_test.go
@@ -100,14 +100,16 @@ func benchmarkRandSeekInSST(
 	defer c.Unref()
 	ctx := context.Background()
 	readerOpts := sstable.ReaderOptions{
+		ReaderOptions: block.ReaderOptions{
+			CacheOpts: sstableinternal.CacheOptions{
+				Cache:   c,
+				CacheID: c.NewID(),
+				FileNum: 1,
+			},
+		},
 		Comparer:   writerOpts.Comparer,
 		KeySchemas: sstable.MakeKeySchemas(&KeySchema),
 	}
-	readerOpts.SetInternalCacheOpts(sstableinternal.CacheOptions{
-		Cache:   c,
-		CacheID: c.NewID(),
-		FileNum: 1,
-	})
 	reader, err := sstable.NewReader(ctx, obj, readerOpts)
 	require.NoError(b, err)
 	defer reader.Close()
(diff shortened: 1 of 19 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4269)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`sstable/options.go` line 120 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4269#-OGpbw3q7GwkINfdf3o7:-OGpbw3rD8ojSaXNHslY:b-apwo8l) ([raw file](https://github.com/cockroachdb/pebble/blob/d4b77cbbf24807baafeebeffe7d7cc402c9d3f6b/sstable/options.go#L120)):*
> ```Go
> 
> 	// internal options can only be used from within the pebble package.
> 	internal sstableinternal.ReaderOptions
> ```

I removed this internal field, but I think it's alright to export the CacheOpts field since the referenced struct itself defined in the internal package.

RaduBerinde (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4269)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/options.go` line 116 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4269#-OGq7piG7oKs1zVw4kdb:-OGq7piG7oKs1zVw4kdc:b-8l9wze) ([raw file](https://github.com/cockroachdb/pebble/blob/4781c70f3d7616435565a8fd4408d8f5ad615b0d/sstable/options.go#L116)):*
> ```Go
> // this method is public, a caller outside the pebble package can't construct a
> // value to pass to it.
> func (o *ReaderOptions) SetInternalCacheOpts(cacheOpts sstableinternal.CacheOptions) {
> ```

[nit] We no longer need this method (we can just set `CacheOpts` directly now)

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4269#-:-OGqFEZ_1Knwa-bkq7XX:b-814g8k)* status: 0 of 19 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`sstable/options.go` line 116 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4269#-OGq7piG7oKs1zVw4kdb:-OGqFFhXDoSZIhzw_6-B:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/4781c70f3d7616435565a8fd4408d8f5ad615b0d/sstable/options.go#L116)):*

[nit] We no longer need this method (we can just set `CacheOpts` directly now)
</blockquote></details>

Done

PR #4275, opened 2025-01-21
metamorphic: fatal on nondeterministic single delete

jbowens:
Terminate and fail the metamorphic test if a nondeterministic single delete is observed. The generator is responsible for generating sequences of operations that ensure determinism across all operations, including single deletes.

Informs #4267.

### DIFF
--- a/metamorphic/test.go
+++ b/metamorphic/test.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"runtime/debug"
 	"sort"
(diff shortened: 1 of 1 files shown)


RaduBerinde:
I thought this can have false positives, as documented here: https://github.com/cockroachdb/pebble/blob/769e97786494f283be402b8ec0b2df7c4dc58002/event.go#L632

jbowens:
Whoops. I re-read that comment but somehow missed it.

Issue #4268, opened 2025-01-17
Calling `db.Checkpoint` with `destDir` like `./foo` returns `no such file or directory`


gawen:
IIUC, this is because `mkdirAllAndSyncParents` is called, append an empty string as parent path in `parentPaths`, and `fs.OpenDir` on an empty string fails with the error `no such file or directory`.

Using pebble v.1.1.3

Jira issue: PEBBLE-324

siviae:
I came across this issue too. Seems like it was already fixed in https://github.com/cockroachdb/pebble/issues/3842 in version 2.0.0.

However I can't import 2.0.0 because of https://github.com/cockroachdb/pebble/issues/4271. The simplest workaround for now is to define checkpoint directory like that: "../parentDir/foo".

RaduBerinde:
I will try to backport the fix https://github.com/cockroachdb/pebble/pull/3847 to the 1.0 branch.

RaduBerinde:
Done. You can try again with v1.1.4

PR #4276, opened 2025-01-21
pebble-release-1.1: db: fix Open issue with single-item relative paths

RaduBerinde:
Previously Open with a relative path containing a single element would fail on real filesystems, because Open would attempt to open the empty path.

Backport of #3847 for #4268.


jbowens (approved):

RaduBerinde:
TFTR!

PR #4277, opened 2025-01-22
pebble-release-2.0: change path to v2

### LINKED ISSUE #4271: v2 module incorrectly named
Good afternoon,

I noticed your v2 branch doesn't have github.com/cockroachdb/pebble/**v2** as it's module name.
Is this intentional?

I'm having some issues importing your v2 branch:

_command_
`go get github.com/cockroachdb/pebble@v2.0.0`

_output_
`go: github.com/cockroachdb/pebble@v2.0.0: invalid version: module contains a go.mod file, so module path must match major version ("github.com/cockroachdb/pebble/v2")`

I could be wrong but I think this is because your go.mod file shows **github.com/cockroachdb/pebble** as the module name instead of github.com/cockroachdb/pebble/**v2**

Jira issue: PEBBLE-326

RaduBerinde:
A go module with version `v2.0` must have a path that ends in `/v2`; otherwise `go get` gives you a "module path must match major version" error.

This commit enacts this change for the pebble-release-2.0 branch. In the future we will consider making this change on master as well so we don't have inconsistency between the branches.

Fixes #4271


jbowens (approved):
:lgtm: I had no idea

PR #4282, opened 2025-01-23
crossversion: collect fewer artifacts on failure

jbowens:
Adjust the metamorphic test to collect fewer artifacts on failure. Previously, we collected the entire root directory containing all in-progress intermediary runs. We only need the one that failed or the two that diverged.

This hopes to address out-of-memory errors in CI when uploading artifacts.

Informs #4202.


jbowens:
bump

PR #4284, opened 2025-01-24
sstable: change load semaphore to limit bytes

RaduBerinde:
I don't plan to merge until the crdb side is done and reviewed as well.

#### sstable: change load semaphore to limit bytes

Change the block load semaphore to limit the total number of bytes in
outstanding reads, instead of the number of reads. In practice there
shouldn't be a huge difference most of the time, but limiting bytes
protects against more cornercases: we did see cases where many reads
were trying to load the same huge block which led to memory
consumption issues (this particular case was addressed by
deduplicating reads but this mechanism would have saved us from OOMs).

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -2364,8 +2364,8 @@ func parseOrderingTokens(tokens []string) (orderingNode, int) {
 type readTrackFS struct {
 	vfs.FS
 
-	currReadCount atomic.Int32
-	maxReadCount  atomic.Int32
+	currReadBytes atomic.Int32
+	maxReadBytes  atomic.Int32
 }
 
 type readTrackFile struct {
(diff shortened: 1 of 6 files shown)


sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4284)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`-- commits` line 11 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4284#-OHeKYTN2AxrkvoGYEZv:-OHeKYTN2AxrkvoGYEZw:bf0w6xi):*
> where many reads were trying to load the same huge block

Shouldn't this case be addressed by the de-duplication of reads?
I am not doubting the need for this PR, since having a byte limit is more principled, and would come into play if all the readers were loading different blocks.

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4284#-:-OHk-rSi2pfqAPy97dIT:bg7igkv)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/block/block.go` line 371 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4284#-OHjrg2YCGw-qPhKZ7OY:-OHjrg2Z-N62QHHOnCvn:b-e1gazq) ([raw file](https://github.com/cockroachdb/pebble/blob/56bdac4a174fd0c016e629161fbbd544d88bb615/sstable/block/block.go#L371)):*
> ```Go
> 	// be loaded (i.e. read from the filesystem) in parallel. Each block load
> 	// acquires X units from the semaphore, where X is the on-disk size of the
> 	// block.
> ```

Is there a concern about head-of-line blocking if there is a huge request because a single k-v happens to be huge. The default value of `kv.raft.command.max_size` is 64MiB and I think that is the only constraint in the system on the size of a k-v. And someone could even increase the default.
We saw 84 64MiB allocations in https://github.com/cockroachlabs/support/issues/3153#issuecomment-2620207778, presumably all sitting in the block cache, so there could be many such blocks in real systems.

What fraction of memory will this semaphore be set to? I suppose with the expected multi-GB setting, head of line blocking shouldn't be a problem. Capping what is acquired per request (upper-bound akin to the lower-bound), seems unnecessary.

The lower-bound of 8KiB sounds ok-ish. I wonder if it should be adaptive based on the capacity of the semaphore, since the lower-bound is mainly trying to put a concurrency limit. 8KiB * 7500 (the default of `storage.block_load.node_max_active`) is only 59MiB, so isn't much of a concurrency limit.

RaduBerinde:
Thanks, these are good questions. I will ponder this some more. One option would be to keep the semaphore as is and add another limit of total read size per read shard.

RaduBerinde:
> We saw 84 64MiB allocations in https://github.com/cockroachlabs/support/issues/3153#issuecomment-2620207778, presumably all sitting in the block cache, so there could be many such blocks in real systems.

These are likely backing the swiss table, not data blocks.

RaduBerinde:
Also, it's just 2 of them (84 is the index of the size class).

sumeerbhola (commented):
> Also, it's just 2 of them (84 is the index of the size class).

I clearly can't read :)

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4284#-:-OHrqcv7EPJBhmlzR8gU:bzdnf7o)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @jbowens and @RaduBerinde)

RaduBerinde:
I gave this some more thought and we'll want to retain a IO count limiter which guarantees protection from the thread limit. Since there's no strong case for adding a byte-limiter on top of that, I'll close for now.

jbowens:
We could define the semaphore over ceil(<read-size>/4096) or something similar so that we're capturing the fact that underlying iops increases for bigger reads, and we're still guarding against high concurrent small reads.

RaduBerinde:
It's roughly equivalent to having a bytes limit with a lower bound on the amount we take from the semaphore. The problem is that even with a modest total byte limit of say 100MB, that translates to 25k IOs small IOs..  We'd have to use a value much larger than 4096 (more like 100+KB).

Issue #4287, opened 2025-01-24
Flush hangs indefinitely from FormatDeleteSizedAndObsolete


carsonip:
From format version FormatDeleteSizedAndObsolete, adding 2 keys that share the same prefix after comparer.Split would hang indefinitely with background error:
```
2025/01/24 18:08:04 background error: pebble: keys must be added in strictly increasing order: foo11:bar#11,SET, foo1:bar#10,SET
```

Minimally reproducible example:
```go
package main

import (
	"bytes"
	"testing"

	"github.com/cockroachdb/pebble"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
)

func comparer() *pebble.Comparer {
	comparer := *pebble.DefaultComparer
	comparer.Split = func(k []byte) int {
		return bytes.IndexByte(k, ':')
	}
	return &comparer
}

func openPebble(storageDir string, version pebble.FormatMajorVersion) (*pebble.DB, error) {
	opts := &pebble.Options{
		FormatMajorVersion: version,
		Comparer:           comparer(),
	}
	return pebble.Open(storageDir, opts)
}

func TestFlush(t *testing.T) {
	versions := []pebble.FormatMajorVersion{pebble.FormatDefault}
	for i := 14; i <= 19; i++ {
		versions = append(versions, pebble.FormatMajorVersion(i))
	}
	for _, version := range versions {
		t.Run(version.String(), func(t *testing.T) {
			db, err := openPebble(t.TempDir(), version)
			require.NoError(t, err)
			defer db.Close()

			err = db.Set([]byte("foo1:bar"), []byte("baz"), pebble.NoSync)
			assert.NoError(t, err)

			err = db.Set([]byte("foo11:bar"), []byte("baz"), pebble.NoSync)
			assert.NoError(t, err)

			// BUG: From format version 15, Flush hangs indefinitely with
			// background error: pebble: keys must be added in strictly increasing order: foo11:bar#11,SET, foo1:bar#10,SET
			err = db.Flush()
			require.NoError(t, err)
		})
	}
}
```

Jira issue: PEBBLE-330

jbowens:
There are strict requirements on the relationship between `Split` and the other `Comparer` funcs, and this comparer violates them:
```
func comparer() *pebble.Comparer {
	comparer := *pebble.DefaultComparer
	comparer.Split = func(k []byte) int {
		return bytes.IndexByte(k, ':')
	}
	return &comparer
}
```

See
[...]

PR #4292, opened 2025-01-27
Update stale comment in comparer.go

carsonip:
Refer to base.CheckComparer for usage.

### DIFF
--- a/comparer.go
+++ b/comparer.go
@@ -30,9 +30,5 @@ type Comparer = base.Comparer
 // DefaultComparer exports the base.DefaultComparer variable.
 var DefaultComparer = base.DefaultComparer
 
-// CheckComparer is a mini test suite that verifies a comparer implementation.
-//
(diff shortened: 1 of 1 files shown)


RaduBerinde:
The existing comment is actually useful. The new one is completely redundant with the line itself (the others like that exist just to appease the linter I believe).

I'd also discourage PRs that just change some comments.

carsonip:
sorry @RaduBerinde , I should have provided more context in the PR. This PR came out of the discussion in https://github.com/cockroachdb/pebble/issues/4287#issuecomment-2614473262 that there is no need for `prefixes` to contain nil as @jbowens said it is likely an outdated comment. I took the liberty to deduplicate the root CheckComparer comment so that the base CheckComparer comment can be the source of truth.

RaduBerinde:
Ah, I see. LGTM, will merge when CI is happy.

RaduBerinde (approved):

PR #4293, opened 2025-01-27
cockroachkvs: handle untimestamped keys in IsLowerBound

jbowens:
The untimestamped key is special and sorts before all other keys. Previously the cockroachkv KeySeeker treated it as sorting after all other keys for the purposes of IsLowerBound.

### DIFF
--- a/cockroachkvs/cockroachkvs.go
+++ b/cockroachkvs/cockroachkvs.go
@@ -743,34 +743,60 @@ func (ks *cockroachKeySeeker) IsLowerBound(k []byte, syntheticSuffix []byte) boo
 	if len(syntheticSuffix) > 0 {
 		return ComparePointSuffixes(syntheticSuffix, k[len(roachKey)+1:]) >= 0
 	}
+	firstRowWall := ks.mvccWallTimes.At(0)
+	firstLogical := ks.mvccLogical.At(0)
+	// If the first row's WallTime and LogicalTime are both zero, the first row
+	// is either (a) unversioned (and sorts before all other suffixes) or (b) is
+	// a non-MVCC key with an untyped version.
+	if firstRowWall == 0 && firstLogical == 0 {
+		return ComparePointSuffixes(ks.untypedVersions.At(0), version) >= 0
+	}
+
 	var wallTime uint64
 	var logicalTime uint32
 	switch len(version) {
 	case engineKeyNoVersion:
+		// The zero-length version is the smallest possible suffix.
+		return true
 	case engineKeyVersionWallTimeLen:
 		wallTime = binary.BigEndian.Uint64(version[:8])
 	case engineKeyVersionWallAndLogicalTimeLen, engineKeyVersionWallLogicalAndSyntheticTimeLen:
 		wallTime = binary.BigEndian.Uint64(version[:8])
 		logicalTime = binary.BigEndian.Uint32(version[8:12])
 	default:
-		// The provided key `k` is not a MVCC key. Assert that the first key in
-		// the block is also not an MVCC key. If it were, that would mean there
-		// exists both a MVCC key and a non-MVCC key with the same prefix.
+		// The provided key `k` is not a MVCC key.
 		//
-		// TODO(jackson): Double check that we'll never produce index separators
-		// that are invalid version lengths.
-		if invariants.Enabled && ks.mvccWallTimes.At(0) != 0 {
-			panic("comparing timestamp with untyped suffix")
+		// The first row IS an MVCC key. We don't expect this to happen in
+		// practice, so this path is not performance sensitive. We reconsistute
+		// the first row's MVCC suffix and then compare it to the provided key's
+		// non-MVCC suffix.
+
+		//gcassert:noescape
+		var buf [13]byte
+		//gcassert:inline
+		binary.BigEndian.PutUint64(buf[:], firstRowWall)
+		var firstRowSuffix []byte
+		if firstLogical == 0 {
+			buf[8] = 9
+			firstRowSuffix = buf[:9]
+		} else {
+			//gcassert:inline
+			binary.BigEndian.PutUint32(buf[8:], uint32(firstLogical))
+			buf[12] = 13
+			firstRowSuffix = buf[:13]
 		}
-		return ComparePointSuffixes(ks.untypedVersions.At(0), version) >= 0
+		return ComparePointSuffixes(firstRowSuffix, version) >= 0
 	}
 
 	// NB: The sign comparison is inverted because suffixes are sorted such that
 	// the largest timestamps are "smaller" in the lexicographical ordering.
-	if v := cmp.Compare(ks.mvccWallTimes.At(0), wallTime); v != 0 {
+	if v := cmp.Compare(firstRowWall, wallTime); v != 0 {
+		// The empty-suffixed zero-timestamped key sorts first. If the first row
+		// has zero wall and logical times, it sorts before k and k is not a
+		// lower bound.
 		return v < 0
 	}
-	return cmp.Compare(uint32(ks.mvccLogical.At(0)), logicalTime) <= 0
+	return cmp.Compare(uint32(firstLogical), logicalTime) <= 0
 }
 
 // SeekGE is part of the KeySeeker interface.
--- a/cockroachkvs/testdata/key_schema_key_seeker
+++ b/cockroachkvs/testdata/key_schema_key_seeker
@@ -169,3 +169,75 @@ MaterializeUserKey(-1, 0) = hex:6d6f6f0000000000b2d05e0109
 MaterializeUserKey(0, 1) = hex:6d6f6f0000000000b2d05e00000000020d
 MaterializeUserKey(1, 2) = hex:6d6f6f0000000000b2d05e00000000010d
 MaterializeUserKey(2, 3) = hex:6d6f6f0000000000b2d05e0009
+
+define-block
+moo
+moo @ 3000000002,1
+moo @ 3000000001,1
+moo @ 0000000001,0
+----
+Parse(moo) = hex:6d6f6f00
+Parse(moo @ 3000000002,1) = hex:6d6f6f0000000000b2d05e02000000010d
+Parse(moo @ 3000000001,1) = hex:6d6f6f0000000000b2d05e01000000010d
+Parse(moo @ 0000000001,0) = hex:6d6f6f00000000000000000109
+
+# IsLowerBound should return false for all keys because the first key of the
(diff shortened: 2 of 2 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4293)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`cockroachkvs/cockroachkvs.go` line 774 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4293#-OHccrYT88m1WH5KbYO1:-OHcelfN8GkRBjLFZXV8:bmdspts) ([raw file](https://github.com/cockroachdb/pebble/blob/9f49f63fb5f836c62d450881e0404c507ab5d5af/cockroachkvs/cockroachkvs.go#L774)):*

As sorting where? A bit rough to put the entire suffix keyspace in between the first WallTime=0 key and next WallTime=0 key...
</blockquote></details>

I'm surprised those KV test cases work in that case. We're stuck with the ordering that we have, but it seems extremely error prone to break up the suffix keyspace like that. I can adjust this for now, but I think we should ultimately prohibit those cases because it's not likely to do what's intended (eg, seeking will not get you the most recent version of the key)

RaduBerinde:
*[`cockroachkvs/cockroachkvs.go` line 774 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4293#-OHccrYT88m1WH5KbYO1:-OHd8EFnDVEUZdqWyg1I:bwmoshc) ([raw file](https://github.com/cockroachdb/pebble/blob/9f49f63fb5f836c62d450881e0404c507ab5d5af/cockroachkvs/cockroachkvs.go#L774)):*

I'm surprised those KV test cases work in that case. We're stuck with the ordering that we have, but it seems extremely error prone to break up the suffix keyspace like that. I can adjust this for now, but I think we should ultimately prohibit those cases because it's not likely to do what's intended (eg, seeking will not get you the most recent version of the key)
</blockquote></details>

I'm not sure I understand the concern. When we introduced crdb we conflated a non-existant suffix with a suffix that is a 0 timestamp Which is fine because the latter is not valid.

But a timestamp that has 0 wall time and non-zero logical time is a valid timestamp and sorts just like any other timestamp. That's why the rest of the code in this file checks both the logical and the wall time against 0, e.g. below we have
```
if wallTime == 0 && logicalTime == 0 {
				// This row has an empty suffix. Since the seek suffix is not empty, it comes after.
```

We Just need `&& !(firstRowWall == 0 && ks.mvccLogicalTimes.At(0) == 0)` I think

RaduBerinde:
*[`cockroachkvs/cockroachkvs.go` line 774 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4293#-OHccrYT88m1WH5KbYO1:-OHdARi-4BvfWKSa3HzU:b-8disuc) ([raw file](https://github.com/cockroachdb/pebble/blob/9f49f63fb5f836c62d450881e0404c507ab5d5af/cockroachkvs/cockroachkvs.go#L774)):*

I'm not sure I understand the concern. When we introduced crdb we conflated a non-existant suffix with a suffix that is a 0 timestamp Which is fine because the latter is not valid.

But a timestamp that has 0 wall time and non-zero logical time is a valid timestamp and sorts just like any other timestamp. That's why the rest of the code in this file checks both the logical and the wall time against 0, e.g. below we have
```
if wallTime == 0 && logicalTime == 0 {
				// This row has an empty suffix. Since the seek suffix is not empty, it comes after.
```

We Just need `&& !(firstRowWall == 0 && ks.mvccLogicalTimes.At(0) == 0)` I think
</blockquote></details>

*When we introduced the crdb key schema

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4293)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`cockroachkvs/cockroachkvs.go` line 774 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4293#-OHccrYT88m1WH5KbYO1:-OHdqU3H75B5mFMdqVeb:b-2eancl) ([raw file](https://github.com/cockroachdb/pebble/blob/9f49f63fb5f836c62d450881e0404c507ab5d5af/cockroachkvs/cockroachkvs.go#L774)):*
The only purpose of the logical component of the timestamp is to disambiguate timestamps within the same wall time, so setting a `walltime=x,logical=1` implies the existence of a transaction attempting to use `walltime=x,logical=0`. So any code in that in practice sets `walltime=0,logical=1` has already gone off the rails by trying to use  `walltime=0,logical=0` as a real MVCC timestamp. We rely on seeking to a user key landing on the most recent key of the version, but that's not true if there exist any keys with `walltime=0`, because `logical=0` sorts ahead of all other timestamps with `walltime=0`.

I'm saying that since we overload `walltime=0,logical=0` as the empty suffix, `walltime=0,logical=y` should be prohibited because it's not possible to use keys with `walltime=0,logical=y` reliably in practice.

> We Just need `&& !(firstRowWall == 0 && ks.mvccLogicalTimes.At(0) == 0)` I think

Hee, I think I currently have an equivalent (DeMorgan's Law) expression. Your construction mirrors the comment better, so I'll switch it.

----

On second thought, I'm going to refactor this to also account for the mixing of MVCC and non-MVCC keys. It's more obviously correct in all cases and simpler in this unversioned first row case.

RaduBerinde (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4293#-:-OHhn7hr3VsW6hebss41:b-uwkowd)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`cockroachkvs/cockroachkvs.go` line 774 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4293#-OHccrYT88m1WH5KbYO1:-OHhlE1MA8z2tYRyPaNK:b-cy0fxy) ([raw file](https://github.com/cockroachdb/pebble/blob/9f49f63fb5f836c62d450881e0404c507ab5d5af/cockroachkvs/cockroachkvs.go#L774)):*
> So any code in that in practice sets `walltime=0,logical=1` has already gone off the rails by trying to use `walltime=0,logical=0` as a real MVCC timestamp.

But the latter never happens..  We do use `walltime=0,logical=y` reliably as long as y != 0.

I agree that it would be cleaner to disallow this, and it could make some key schema code more efficient (we won't need to look at the logical value when walltime=0). But we'd need to add an assertion and then hunt down the test code that uses 0 walltimes. If you have too much on your plate right now, I can take that up separately.
___
*[`cockroachkvs/cockroachkvs.go` line 764 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4293#-OHhml1y1IHJCObNPDQ5:-OHhml1y1IHJCObNPDQ6:b-id31mu) ([raw file](https://github.com/cockroachdb/pebble/blob/70b6886b3f4f819cdce507066fe6984c27bde442/cockroachkvs/cockroachkvs.go#L764)):*
> ```Go
> 		firstWall := ks.mvccWallTimes.At(0)
> 		firstLogical := ks.mvccLogical.At(0)
> 		if firstWall == 0 && firstLogical == 0 {
> ```

We can move this before the `switch` instead of doing it in two paths.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4293)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`cockroachkvs/cockroachkvs.go` line 774 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4293#-OHccrYT88m1WH5KbYO1:-OHhnchu8vQeurSTN97f:b3rv4vy) ([raw file](https://github.com/cockroachdb/pebble/blob/9f49f63fb5f836c62d450881e0404c507ab5d5af/cockroachkvs/cockroachkvs.go#L774)):*
> But the latter never happens.. We do use `walltime=0,logical=y` reliably as long as y != 0.

It _can_ happen though, because if we serialize an MVCC key with `walltime=0,logical=0`, it _will_ work. Its encoding is just the empty suffix.

RaduBerinde:
*[`cockroachkvs/cockroachkvs.go` line 774 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4293#-OHccrYT88m1WH5KbYO1:-OHhoII4EvNLP_xwb55k:b-br7fbm) ([raw file](https://github.com/cockroachdb/pebble/blob/9f49f63fb5f836c62d450881e0404c507ab5d5af/cockroachkvs/cockroachkvs.go#L774)):*

> But the latter never happens.. We do use `walltime=0,logical=y` reliably as long as y != 0.

It _can_ happen though, because if we serialize an MVCC key with `walltime=0,logical=0`, it _will_ work. Its encoding is just the empty suffix.
</blockquote></details>

Inside crdb a zero timestamp is treated specially and is never serialized (definitely not with the expectation that it will function as a timestamp). It is considered just a marker for "no timestamp".

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4293#-:-OHhq--K5ABcI_hZUaQK:b-814g8k)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`cockroachkvs/cockroachkvs.go` line 774 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4293#-OHccrYT88m1WH5KbYO1:-OHhpYYc3ogBnBA15l1T:b-5m1u5w) ([raw file](https://github.com/cockroachdb/pebble/blob/9f49f63fb5f836c62d450881e0404c507ab5d5af/cockroachkvs/cockroachkvs.go#L774)):*
> Inside crdb a zero timestamp is treated specially and is never serialized (definitely not with the expectation that it will function as a timestamp). It is considered just a marker for "no timestamp".

Agreed it's treated specially and never serialized, but from the perspective of calling code there's no difference. A MVCCKey with a `hlc.Timestamp{}` can be written and read, and the caller is none the wiser that they're actually writing a special unversioned key.
___
*[`cockroachkvs/cockroachkvs.go` line 764 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4293#-OHhml1y1IHJCObNPDQ5:-OHhpz8829zWRtA9TZGA:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/70b6886b3f4f819cdce507066fe6984c27bde442/cockroachkvs/cockroachkvs.go#L764)):*

We can move this before the `switch` instead of doing it in two paths.
</blockquote></details>

Done

PR #4295, opened 2025-01-27
sstable: always use target buffer for CompressAndChecksum

RaduBerinde:
Currently `CompressAndChecksum` can alias the original data buffer if
we do not compress the data (either because compression is disabled,
or the data was not compressible enough). In most cases, we write out
the resulting data which can mangle the buffer. This leads most
callers to check if the buffer is not compressed and make a copy.

This change moves the copy into `CompressAndChecksum`; we always use
the dst buffer, even if we don't compress. This simplifies the callers
and makes things less fragile.

### DIFF
--- a/sstable/block/compression.go
+++ b/sstable/block/compression.go
@@ -212,18 +212,6 @@ func (b PhysicalBlock) Clone() PhysicalBlock {
 	return PhysicalBlock{data: data, trailer: b.trailer}
 }
 
-// CloneUsingBuf makes a copy of the block data, using the given slice if it has
-// enough capacity.
-func (b PhysicalBlock) CloneUsingBuf(buf []byte) (_ PhysicalBlock, newBuf []byte) {
-	newBuf = append(buf[:0], b.data...)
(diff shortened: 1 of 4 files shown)


jbowens (approved):
:lgtm:

Good idea

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4295#-:-OHdfzz39E3QWTCyaUQf:bc89ay1)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4295#-OHdf8041rox-u0lr3EF:-OHdf8041rox-u0lr3EG:b-xiyvsi):*
nit: `CompressAndChecksum`

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4295#-:-OHdhjYoB5I5FaGFc1o4:b-814g8k)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @RaduBerinde)

PR #4297, opened 2025-01-27
db: introduce CompactionScheduler and integrate with DB

sumeerbhola:
CompactionScheduler is an interface that encompasses (a) our current compaction scheduling behavior, (b) compaction scheduling in a multi-store setting that adds a per-node limit in addition to the per-store limit, and prioritizes across stores, (c) compaction scheduling that includes (b) plus is aware of resource usage and can prioritize across stores and across other long-lived work in addition to compactions (e.g. range snapshot reception).

CompactionScheduler calls into DB and the DB calls into the CompactionScheduler. This requires some care in specification of the synchronization expectations, to avoid deadlock. For the most part, the complexity is borne by the CompactionScheduler -- see the code comments for details.

ConcurrencyLimitScheduler is an implementation for (a), and is paired with a single DB. It has no knowledge of delete-only compactions, so we have redefined the meaning of Options.MaxConcurrentCompactions, as discussed in the code comment.

CompactionScheduler has some related interfaces/structs:
- CompactionGrantHandle is used to report the start and end of the compaction, and frequently report the written bytes, and CPU consumption. In the implementation of CompactionGrantHandle provided by CockroachDB's AC component, the CPU consumption will use the grunning package.
- WaitingCompaction is a struct used to prioritize the DB's compaction relative to other long-lived work (including compactions by other DBs). makeWaitingCompaction is a helper that constructs this struct.

For integrating the CompactionScheduler with DB, there are a number of changes:
- The entry paths to ask to schedule a compaction are reduced to 1, by removing DB.maybeScheduleCompactionPicker. The only path is DB.maybeScheduleCompaction.
- versionSet.{curCompactionConcurrency,pickedCompactionCache} are added to satisfy the interface expected by CompactionScheduler. Specifically, pickedCompactionCache allows us to safely cache a pickedCompaction that cannot be run. There is some commentary on the worst-case waste in compaction picking -- with the default ConcurrencyLimitScheduler on average there should be no wastage.
- versionSet.curCompactionConcurrency and DB.mu.compact.manualLen are two atomic variables introduced to implement DB.GetAllowedWithoutPermission, which allows the DB to adjust what minimum compaction concurrency it desires based on the backlog of automatic and manual compactions. The encoded logic is meant to be equivalent to our current logic.

The CompactionSlot and CompactionLimiter introduced in a recent PR are deleted. CompactionGrantHandle is analogous to CompactionSlot, and allows for measuring of CPU usage since the implementation of CompactionScheduler in AC will explicitly monitor usage and capacity. CompactionScheduler is analogous to CompactionLimiter. CompactionLimiter had a non-queueing interface in that it never called into the DB. This worked since the only events that allowed another compaction to run were also ones that caused another call to maybeScheduleCompaction. This is not true when a CompactionScheduler is scheduling across multiple DBs, or managing a compaction and other long-lived work (snapshot reception), since something unrelated to the DB can cause resources to become available to run a compaction.

There is a partial implementation of a resource aware scheduler in https://github.com/sumeerbhola/cockroach/tree/long_lived_granter/pkg/util/admission/admit_long.

Informs #3813, https://github.com/cockroachdb/cockroach/issues/74697, #1329


sumeerbhola (commented):
Consider this a prototype. Once I have some experimental results (with CockroachDB integration), I will start cleaning this up into smaller changes for review.
But it gives a good sense for the kind of changes this will entail.
 
Existing Pebble tests pass and stress meta ran for 3+ hours without failure. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4297#-:-OHdqruV7p15Cy8pl-xY:bpr4p87)* status: 0 of 22 files reviewed, all discussions resolved (waiting on @aadityasondhi)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4297)* status: 0 of 22 files reviewed, 4 unresolved discussions (waiting on @aadityasondhi)

___
*[`compaction_scheduler.go` line 76 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OIsJMtp99ooVWzNenlN:-OIsJMtp99ooVWzNenlO:bi8i4vy) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/compaction_scheduler.go#L76)):*
> ```Go
> 	// by CompactionScheduler is long-lived (often executing for multiple
> 	// seconds).
> 	GetWaitingCompaction() (bool, WaitingCompaction)
> ```

In practice, would this be called more than once after `TrySchedule`? (if not, we can just pass the compaction information via TrySchedule; if yes, when? periodically?)
___
*[`compaction_scheduler.go` line 77 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OIsGqfl9LQYX9vfz-eP:-OIsGqfmAr6FYaxfQJpy:b-b74fyw) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/compaction_scheduler.go#L77)):*
> ```Go
> 	// seconds).
> 	GetWaitingCompaction() (bool, WaitingCompaction)
> 	// Schedule grants the DB permission to run a compaction. The DB returns
> ```

Isn't this specifically permission to run the compaction it last returned via `GetWaitingCompaction`?
___
*[`compaction_scheduler.go` line 90 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OIsGAkV4JR-r7x3DCMw:-OIsGAkV4JR-r7x3DCMx:bfxjofs) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/compaction_scheduler.go#L90)):*
> ```Go
> 	// overall healthy LSM. This value can be compared across compactions and
> 	// other long-lived work.
> 	Optional bool
> ```

[nit] Isn't having a priority enough?
___
*[`internal/base/compaction_grant_handle.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OIsJ2KF7kYtpPaNTlVN:-OIsJ2KF7kYtpPaNTlVO:b-dyq6mo) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/internal/base/compaction_grant_handle.go#L26)):*
> ```Go
> 	// requirement to call it on the same goroutine that runs the compaction.
> 	Started()
> 	// MeasureCPU is called from each of the two goroutines that consume CPU
> ```

[nit] What does it do?

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4297)* status: 0 of 23 files reviewed, 4 unresolved discussions (waiting on @aadityasondhi and @RaduBerinde)

___
*[`compaction_scheduler.go` line 76 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OIsJMtp99ooVWzNenlN:-OJPv9km6LLR6pPa7e1N:bvlx7yw) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/compaction_scheduler.go#L76)):*

In practice, would this be called more than once after `TrySchedule`? (if not, we can just pass the compaction information via TrySchedule; if yes, when? periodically?)
</blockquote></details>

Yes it can be called many times with no bounds. The scheduler could keep calling it whenever it has some capacity, and pick some other long-lived work (another store's compaction, incoming range snapshot). Hence the caching logic in `pickedCompactionCache` which then results in a worst-case overhead (discussed in a comment below).

The reason we don't pass this info in `TrySchedule` is that the scheduler has no ability to know when it becomes stale. So it becomes Pebble's responsibility to know the freshest state (either in the cache or needs to pick a compaction) when asked.
___
*[`compaction_scheduler.go` line 77 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OIsGqfl9LQYX9vfz-eP:-OJPvzXa1e7aG3gcCm-j:b20sckd) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/compaction_scheduler.go#L77)):*

Isn't this specifically permission to run the compaction it last returned via `GetWaitingCompaction`?
</blockquote></details>

Kind of. Pebble's state may have changed and it may run some other compaction. In practice, we are assuming that a `DBForCompaction` will run either what was returned in `GetWaitingCompaction` or something even more important.
___
*[`compaction_scheduler.go` line 90 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OIsGAkV4JR-r7x3DCMw:-OJPwThS7Cu8EC02fP4O:bvqozhg) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/compaction_scheduler.go#L90)):*

[nit] Isn't having a priority enough?
</blockquote></details>

This part is very much subject to revision. The reason I went with this for the prototype is that we don't just have compactions that are vying for resources: incoming range snapshots also consume disk and cpu. And so we would need a way to come up with a prioritization scheme that worked across the two. Instead of that, we classify some Pebble compactions as optional, where this optional bit is decided by Pebble (not the integration code in CockroachDB), and then optional=true becomes the lowest importance. And when the optional bit is the same, compactions win over incoming range snapshots.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4297)* status: 0 of 23 files reviewed, 4 unresolved discussions (waiting on @aadityasondhi and @sumeerbhola)

___
*[`compaction_scheduler.go` line 77 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OIsGqfl9LQYX9vfz-eP:-OJQ15Dj6jpYO-1s2ONa:bqo89p7) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/compaction_scheduler.go#L77)):*

Kind of. Pebble's state may have changed and it may run some other compaction. In practice, we are assuming that a `DBForCompaction` will run either what was returned in `GetWaitingCompaction` or something even more important.
</blockquote></details>

I think that's reasonable. We could make that explicit. `Schedule(CompactionGrantHandle, Priority)`: permission granted to schedule a compaction of priority at least as high as the argument.
___
*[`compaction_scheduler.go` line 90 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OIsGAkV4JR-r7x3DCMw:-OJQ1eka1yLQC-wHTMHM:bauzzt) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/compaction_scheduler.go#L90)):*

This part is very much subject to revision. The reason I went with this for the prototype is that we don't just have compactions that are vying for resources: incoming range snapshots also consume disk and cpu. And so we would need a way to come up with a prioritization scheme that worked across the two. Instead of that, we classify some Pebble compactions as optional, where this optional bit is decided by Pebble (not the integration code in CockroachDB), and then optional=true becomes the lowest importance. And when the optional bit is the same, compactions win over incoming range snapshots.
</blockquote></details>

What I'm missing is why isn't that equivalent to having a single integer, where some of the integer range is only for optional compactions and some other part of the range is for non-optional compactions?

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4297)* status: 0 of 23 files reviewed, 4 unresolved discussions (waiting on @aadityasondhi and @RaduBerinde)

___
*[`compaction_scheduler.go` line 90 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OIsGAkV4JR-r7x3DCMw:-OJTb46hAwe8wQhIGLRb:bpqfvka) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/compaction_scheduler.go#L90)):*

What I'm missing is why isn't that equivalent to having a single integer, where some of the integer range is only for optional compactions and some other part of the range is for non-optional compactions?
</blockquote></details>


That is doable. But note the comparator for the heap is a cleaner tuple comparison due to keeping optional as a separate field https://github.com/sumeerbhola/cockroach/blob/435f3332e75c89fa5ec27b981edf92110ee8eb32/pkg/util/admission/admit_long/granter.go#L942-L955 (from https://github.com/cockroachdb/cockroach/pull/141501)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4297)* status: 0 of 23 files reviewed, 4 unresolved discussions (waiting on @aadityasondhi and @sumeerbhola)

___
*[`compaction_scheduler.go` line 90 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OIsGAkV4JR-r7x3DCMw:-OJU1_Y36v4NbrEfArRi:b-np8rwl) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/compaction_scheduler.go#L90)):*

That is doable. But note the comparator for the heap is a cleaner tuple comparison due to keeping optional as a separate field https://github.com/sumeerbhola/cockroach/blob/435f3332e75c89fa5ec27b981edf92110ee8eb32/pkg/util/admission/admit_long/granter.go#L942-L955 (from https://github.com/cockroachdb/cockroach/pull/141501)
</blockquote></details>

Understood. So we are comparing them like a (Optional, Priority, Score) tuple; it wasn't clear from the comments. Maybe say for priority: "it is only compared across compactions with the same Optional value" and for score: "it is only compared across compactions with the same Optional and Priority values. Or just say at the top that it represents and compares like an (Optional, Priority, Score).

sumeerbhola (commented):
@RaduBerinde @jbowens 
Now that the approach has been partially validated with the integration prototype in https://github.com/cockroachdb/cockroach/pull/141501 and results in https://github.com/cockroachdb/cockroach/issues/136615#issuecomment-2659557636 I'd like to get a detailed review and then merge this PR.
I owe a test for `ConcurrencyLimitScheduler` which will be quite straightforward, so I'd like to wait until the rest is fully reviewed, as interface or other changes can affect `ConcurrencyLimitScheduler`.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4297#-:-OKDqbSI2l34etgaEaL2:b-oc8nq)* status: 0 of 23 files reviewed, 3 unresolved discussions (waiting on @aadityasondhi, @jbowens, and @RaduBerinde)

___
*[`compaction_scheduler.go` line 77 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OIsGqfl9LQYX9vfz-eP:-OKDgNS72L9tDLAuWPGF:byy9h0f) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/compaction_scheduler.go#L77)):*

I think that's reasonable. We could make that explicit. `Schedule(CompactionGrantHandle, Priority)`: permission granted to schedule a compaction of priority at least as high as the argument.
</blockquote></details>

We could. But it isn't clear that this race is worth complicating the interface and bookkeeping. Since for Pebble compactions, the importance of what it wants to run can't really decrease (there is just a backlog of work that increases).
___
*[`compaction_scheduler.go` line 90 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OIsGAkV4JR-r7x3DCMw:-OKDh_4B2wH2IUS7KRmO:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/compaction_scheduler.go#L90)):*

Understood. So we are comparing them like a (Optional, Priority, Score) tuple; it wasn't clear from the comments. Maybe say for priority: "it is only compared across compactions with the same Optional value" and for score: "it is only compared across compactions with the same Optional and Priority values. Or just say at the top that it represents and compares like an (Optional, Priority, Score).
</blockquote></details>

Done
___
*[`internal/base/compaction_grant_handle.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OIsJ2KF7kYtpPaNTlVN:-OKDooT2Drbh6I_QXrfM:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/internal/base/compaction_grant_handle.go#L26)):*

[nit] What does it do?
</blockquote></details>

Done.
___
*[`internal/compact/run.go` line 194 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKDqArg6wiQd39PTlBB:-OKDqArg6wiQd39PTlBC:bqsuchk) ([raw file](https://github.com/cockroachdb/pebble/blob/a8b60cf3d370d1545eb1205575d61447b4b89ae6/internal/compact/run.go#L194)):*
> ```Go
> 			})
> 			// TODO: need to give the GrantHandle to the writer so it can account on
> 			// all its goroutines.
> ```

I will do this in a followup PR, so that main interface and refactoring can be merged first.

jbowens (commented):
> Reviewed 3 of 22 files at r1, 5 of 17 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4297)* status: 8 of 23 files reviewed, 9 unresolved discussions (waiting on @aadityasondhi and @RaduBerinde)

___
*[`compaction_scheduler.go` line 51 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKYQGCd4NTEdXQLF4hg:-OKYQGCd4NTEdXQLF4hh:btt43ey) ([raw file](https://github.com/cockroachdb/pebble/blob/a4a526ffffd8b197a029b05c62c7d0642b81803f/compaction_scheduler.go#L51)):*
> ```Go
> 	// DBForCompaction are finished, so Unregister must not be called while
> 	// holding locks that DBForCompaction acquires in those calls.
> 	Unregister()
> ```

does this not need the `DBForCompaction` to unregister? I suppose the intention is that if managing compactions across multiple stores, each gets its own implementation of `CompactionScheduler`, and internally those `CompactionScheduler`s communicate?
___
*[`compaction_scheduler.go` line 155 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKh1HKIBXZCnF2rEHHK:-OKh1HKIBXZCnF2rEHHL:b-6yut58) ([raw file](https://github.com/cockroachdb/pebble/blob/a4a526ffffd8b197a029b05c62c7d0642b81803f/compaction_scheduler.go#L155)):*
> ```Go
> func (h *noopGrantHandle) MeasureCPU(g int)                                      {}
> func (h *noopGrantHandle) CumulativeStats(stats base.CompactionGrantHandleStats) {}
> func (h *noopGrantHandle) Done()                                                 {}
> ```

Just curious, is there a reason to prefer implementing the interface on the pointer to the empty struct instead of on the empty struct itself?
___
*[`compaction_scheduler.go` line 183 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKgwWasCSEl62w22TTT:-OKgwWasCSEl62w22TTU:b-6e7ob9) ([raw file](https://github.com/cockroachdb/pebble/blob/a4a526ffffd8b197a029b05c62c7d0642b81803f/compaction_scheduler.go#L183)):*
> ```Go
> type pickedCompactionCache struct {
> 	// pc != nil => waiting
> 	waiting bool
> ```

can we add some commentary explaining that `waiting` may be true and while `pc == nil`? would `maybeWaiting` be more apt, because if `pc == nil`, there may or may not be another compaction to run, right?
___
*[`compaction_scheduler.go` line 244 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKgxhG637BQAV7yM9lq:-OKgxhG637BQAV7yM9lr:b-g861mz) ([raw file](https://github.com/cockroachdb/pebble/blob/a4a526ffffd8b197a029b05c62c7d0642b81803f/compaction_scheduler.go#L244)):*
> ```Go
> // DB.maybeScheduleCompaction is called on flush completion. Hence, we resort
> // to having a background thread in ConcurrencyLimitScheduler sample the value
> // every 250ms.
> ```

does this add nondeterminism where we didn't have it before, potentially leading to test flakes? Previously a flush completing in isolation would immediately schedule a compaction if necessary. I suppose in any case that was deterministic, it's still deterministic just with some additional latency of the waiting for the 250ms timer tick?
___
*[`db.go` line 1675 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKh0bLyDFl50PXdsVJw:-OKh0bLyDFl50PXdsVJx:bfvru8d) ([raw file](https://github.com/cockroachdb/pebble/blob/5c86fab0d2cd567e9ba6dec9447e942de5af6ab7/db.go#L1675)):*
> ```Go
> 	if err := d.closed.Load(); err != nil {
> 		panic(err)
> 	}
> ```

nit: I think this check was here, within these two mutex critical sections, to make it more likely that we detect two concurrent calls to Close. If two goroutines call Close, they'll get sequenced on the mutexes. The first one will observe the atomic is nil, and succeed in storing `ErrClosed`. The second will necessarily observe that `ErrClosed`.
___
*[`options.go` line 938 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKgKXEK8-tNQZXDVbQ5:-OKgKXEK8-tNQZXDVbQ6:b-tm8qmj) ([raw file](https://github.com/cockroachdb/pebble/blob/a4a526ffffd8b197a029b05c62c7d0642b81803f/options.go#L938)):*
> ```Go
> 	//
> 	// DB.GetAllowedWithoutPermission will return a value in the interval [1,
> 	// MaxConcurrentCompactions]. A value > 1 is returned:
> ```

How about _beginning_ this comment by describing MaxConcurrentCompactions as an upper bound on the value of the DB's GetAllowedWithoutPermission value? I think describing what it concretely bounds first would make it a little clearer when the comment starts describing its relationship to the actual number of concurrent compactions allowed.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4297)* status: 8 of 23 files reviewed, 10 unresolved discussions (waiting on @aadityasondhi, @RaduBerinde, and @sumeerbhola)

___
*[`compaction_scheduler.go` line 24 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKhSm9W8OSfm8fjULb0:-OKhSm9W8OSfm8fjULb1:b6qk8h9) ([raw file](https://github.com/cockroachdb/pebble/blob/a4a526ffffd8b197a029b05c62c7d0642b81803f/compaction_scheduler.go#L24)):*
> ```Go
> // compaction so that the scheduler can have more granular estimates. For
> // example, the input or output level could affect the write bandwidth if the
> // inputs are better cached (say at higher levels).
> ```

Meta comment: I am finding it tricky to follow the dance of the DB and scheduler calling into one another, but I'm struggling to think of anything simpler

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4297#-:-OKm1VPgCtpY-yjJzj96:b-814g8k)* status: 6 of 23 files reviewed, 10 unresolved discussions (waiting on @aadityasondhi, @jbowens, and @RaduBerinde)

___
*[`compaction_scheduler.go` line 24 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKhSm9W8OSfm8fjULb0:-OKlj3Af7AihJWdCZJPZ:b-u2x4xs) ([raw file](https://github.com/cockroachdb/pebble/blob/a4a526ffffd8b197a029b05c62c7d0642b81803f/compaction_scheduler.go#L24)):*

Meta comment: I am finding it tricky to follow the dance of the DB and scheduler calling into one another, but I'm struggling to think of anything simpler
</blockquote></details>

I don't have any simplification ideas unfortunately, even after implementing the CockroachDB changes.
btw, this isn't very different from the existing `admission.{granter,requester}` interfaces.
___
*[`compaction_scheduler.go` line 51 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKYQGCd4NTEdXQLF4hg:-OKllCHh47dZR6jzs--9:b-iibzoj) ([raw file](https://github.com/cockroachdb/pebble/blob/a4a526ffffd8b197a029b05c62c7d0642b81803f/compaction_scheduler.go#L51)):*

does this not need the `DBForCompaction` to unregister? I suppose the intention is that if managing compactions across multiple stores, each gets its own implementation of `CompactionScheduler`, and internally those `CompactionScheduler`s communicate?
</blockquote></details>

Yes, it gets its own scheduler. Otherwise it would need to pass a parameter to both `Unregister` and `TrySchedule`.

The prototype implementation is in https://github.com/sumeerbhola/cockroach/blob/435f3332e75c89fa5ec27b981edf92110ee8eb32/pkg/storage/pebble_granter.go
___
*[`compaction_scheduler.go` line 155 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKh1HKIBXZCnF2rEHHK:-OKlsZFe-QK-cjRv5qG0:b-enwk5c) ([raw file](https://github.com/cockroachdb/pebble/blob/a4a526ffffd8b197a029b05c62c7d0642b81803f/compaction_scheduler.go#L155)):*

Just curious, is there a reason to prefer implementing the interface on the pointer to the empty struct instead of on the empty struct itself?
</blockquote></details>

No good reason. I may have momentarily forgotten that interfaces can store the value directly when they are small enough and even if they didn't golang optimizes away the allocation for zero-sized structs. I've gotten rid of `singletonNoopGrantHandle`.
___
*[`compaction_scheduler.go` line 183 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKgwWasCSEl62w22TTT:-OKlung-2ErNbpnkeasa:b-lhb16) ([raw file](https://github.com/cockroachdb/pebble/blob/a4a526ffffd8b197a029b05c62c7d0642b81803f/compaction_scheduler.go#L183)):*

can we add some commentary explaining that `waiting` may be true and while `pc == nil`? would `maybeWaiting` be more apt, because if `pc == nil`, there may or may not be another compaction to run, right?
</blockquote></details>

Yes it is possible that there isn't another compaction to run, either because we ran the last picked compaction (`getForRunning`), or a different compaction that doesn't go through the scheduler started running and there is some file overlap that prevents us temporarily from picking a compaction (`invalidate`). But it still represents that this DB is `waiting` for something to happen -- either a compaction to finish or for the scheduler to call `GetWaitingCompaction`. So I think the naming here is fine.

But it is unclear why `pc == nil`, so I've added some more commentary.

I'll also look into whether we can be more bug-proof when it comes to calling `invalidate`. I'm suddenly concerned with what happens if we forget to call `invalidate` and try to use a stale `pickedCompaction`. I am forgetting whether it is harmless (the compaction will simply fail) or something bad will happen.
___
*[`compaction_scheduler.go` line 244 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKgxhG637BQAV7yM9lq:-OKlx9Bi8NnqKuxIhlp3:b-rt1r94) ([raw file](https://github.com/cockroachdb/pebble/blob/a4a526ffffd8b197a029b05c62c7d0642b81803f/compaction_scheduler.go#L244)):*

does this add nondeterminism where we didn't have it before, potentially leading to test flakes? Previously a flush completing in isolation would immediately schedule a compaction if necessary. I suppose in any case that was deterministic, it's still deterministic just with some additional latency of the waiting for the 250ms timer tick?
</blockquote></details>

Yes, some lag. I didn't see any test flakes yet.
In production, this 250ms lag will affect us when we are ramping up concurrency, and may be problematic if the workload is bursty and ramps up and down frequently. I've seen cases where we are flushing > 60MiB/s (sstable bytes), and so could be flushing in < 500ms. I've reduced this to 100ms.
I am wondering whether this is concerning enough to add a `SampleGetAllowedWithoutPermission` method on the scheduler, to be called when a flush ends. Thoughts?
___
*[`db.go` line 1675 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKh0bLyDFl50PXdsVJw:-OKlzZywCe2ZylnJEEu9:b-ctpphv) ([raw file](https://github.com/cockroachdb/pebble/blob/5c86fab0d2cd567e9ba6dec9447e942de5af6ab7/db.go#L1675)):*

nit: I think this check was here, within these two mutex critical sections, to make it more likely that we detect two concurrent calls to Close. If two goroutines call Close, they'll get sequenced on the mutexes. The first one will observe the atomic is nil, and succeed in storing `ErrClosed`. The second will necessarily observe that `ErrClosed`.
</blockquote></details>

Thanks. I added it back with a comment.
___
*[`options.go` line 938 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKgKXEK8-tNQZXDVbQ5:-OKm0ch80U5dRCkUyhKr:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/a4a526ffffd8b197a029b05c62c7d0642b81803f/options.go#L938)):*

How about _beginning_ this comment by describing MaxConcurrentCompactions as an upper bound on the value of the DB's GetAllowedWithoutPermission value? I think describing what it concretely bounds first would make it a little clearer when the comment starts describing its relationship to the actual number of concurrent compactions allowed.
</blockquote></details>

Done

jbowens (approved):
:lgtm:

Up to you if you want to do something about the delay post-flush

> Reviewed 2 of 17 files at r4, 5 of 8 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4297#-:-OL4o-KA7p0NBDggMJKy:b-wfru52)* status: 13 of 23 files reviewed, 8 unresolved discussions (waiting on @aadityasondhi, @RaduBerinde, and @sumeerbhola)

___
*[`compaction_scheduler.go` line 244 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKgxhG637BQAV7yM9lq:-OL4nAbh9OxmDTfcx5tm:b-gut9uw) ([raw file](https://github.com/cockroachdb/pebble/blob/a4a526ffffd8b197a029b05c62c7d0642b81803f/compaction_scheduler.go#L244)):*

Yes, some lag. I didn't see any test flakes yet.
In production, this 250ms lag will affect us when we are ramping up concurrency, and may be problematic if the workload is bursty and ramps up and down frequently. I've seen cases where we are flushing > 60MiB/s (sstable bytes), and so could be flushing in < 500ms. I've reduced this to 100ms.
I am wondering whether this is concerning enough to add a `SampleGetAllowedWithoutPermission` method on the scheduler, to be called when a flush ends. Thoughts?
</blockquote></details>

I don't follow the naming `SampleGetAllowedWithoutPermission`, but I do think it would be reasonable to add a method so that we can schedule a compaction at flush end without delay.

sumeerbhola (commented):
Unit test for ConcurrencyLimitScheduler is ready.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4297#-:-OLAzwOf7tLrzQqJQQZ5:bdy85ty)* status: 7 of 25 files reviewed, 8 unresolved discussions (waiting on @aadityasondhi, @jbowens, and @RaduBerinde)

___
*[`compaction_scheduler.go` line 244 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4297#-OKgxhG637BQAV7yM9lq:-OLARHAP9g8tlX5HVx_1:bqmjwcz) ([raw file](https://github.com/cockroachdb/pebble/blob/a4a526ffffd8b197a029b05c62c7d0642b81803f/compaction_scheduler.go#L244)):*

I don't follow the naming `SampleGetAllowedWithoutPermission`, but I do think it would be reasonable to add a method so that we can schedule a compaction at flush end without delay.
</blockquote></details>

Added.

PR #4303, opened 2025-01-29
cockroachkvs: always use non-empty index separators

jbowens:
Adapt the cockroachkv Successor implementation to always return non-empty keys
with a version length byte. Previously it was possible for Successor to return
an empty slice if passed an empty slice, resulting in an empty string index
separator. When combined with a synthetic prefix, this produced an invalid key:
a non-empty string without a version length byte. Additionally, document that
Successor may receive an empty key as an operand. Document and enforce under
invariants that Successor must never receive empty keys as operands.


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4303)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`cockroachkvs/cockroachkvs_test.go` line 86 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4303#-OHnVNzi0_GIfbaVVfLj:-OHnVNzi0_GIfbaVVfLk:b-yv8pci) ([raw file](https://github.com/cockroachdb/pebble/blob/fbb534f9d58e0ae956b3263c5dda572283c51ac6/cockroachkvs/cockroachkvs_test.go#L86)):*
> ```Go
> 
> 	var buf bytes.Buffer
> 	datadriven.RunTest(t, "testdata/comparer_funcs",
> ```

git add
___
*[`internal/base/comparer.go` line 319 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4303#-OHnV1AnD0JuA5htTE_Y:-OHnV1AnD0JuA5htTE_Z:bl4d49r) ([raw file](https://github.com/cockroachdb/pebble/blob/fbb534f9d58e0ae956b3263c5dda572283c51ac6/internal/base/comparer.go#L319)):*
> ```Go
> 
> 		minLen := min(len(a), len(b))
> 		if i >= minLen {
> ```

[nit] `i == len(a) || i == len(b)` is more clear

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4303)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/base/comparer.go` line 315 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4303#-OHnWBvV6s_r1IPd-kaW:-OHnWBvV6s_r1IPd-kaX:berqho6) ([raw file](https://github.com/cockroachdb/pebble/blob/fbb534f9d58e0ae956b3263c5dda572283c51ac6/internal/base/comparer.go#L315)):*
> ```Go
> 
> 	Separator: func(dst, a, b []byte) []byte {
> 		i, n := SharedPrefixLen(a, b), len(dst)
> ```

You can replace this with `crbytes.CommonPrefix` and remove that function.I'd also separate `n := len(dst)` since it's not directly related

RaduBerinde:
*[`internal/base/comparer.go` line 329 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4303#-OHnZn6Z1kwvCwW3982C:-OHnZn6Z1kwvCwW3982D:bt1yzww) ([raw file](https://github.com/cockroachdb/pebble/blob/fbb534f9d58e0ae956b3263c5dda572283c51ac6/internal/base/comparer.go#L329)):*
> ```Go
> 		}
> 
> 		if i < len(b)-1 || a[i]+1 < b[i] {
> ```

[nit I think `i < len(b)-1` is always true here

RaduBerinde:
*[`internal/base/comparer.go` line 329 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4303#-OHnZn6Z1kwvCwW3982C:-OHn_6lC2M2vIy66fiSt:bu581u) ([raw file](https://github.com/cockroachdb/pebble/blob/fbb534f9d58e0ae956b3263c5dda572283c51ac6/internal/base/comparer.go#L329)):*

[nit I think `i < len(b)-1` is always true here
</blockquote></details>

oh, no, never mind :)

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4303)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/base/comparer.go` line 315 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4303#-OHnWBvV6s_r1IPd-kaW:-OHn_uQM5mPLDgNX772g:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/fbb534f9d58e0ae956b3263c5dda572283c51ac6/internal/base/comparer.go#L315)):*

You can replace this with `crbytes.CommonPrefix` and remove that function.I'd also separate `n := len(dst)` since it's not directly related
</blockquote></details>

Done
___
*[`cockroachkvs/cockroachkvs_test.go` line 86 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4303#-OHnVNzi0_GIfbaVVfLj:-OHn_x3T87DbXGw0DbLU:b-66whp7) ([raw file](https://github.com/cockroachdb/pebble/blob/fbb534f9d58e0ae956b3263c5dda572283c51ac6/cockroachkvs/cockroachkvs_test.go#L86)):*

git add
</blockquote></details>

every damn time

RaduBerinde:
*[`internal/base/comparer.go` line 314 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4303#-OHn_bUXAKaHOVyE_5EV:-OHn_bUXAKaHOVyE_5EW:b-q08uzi) ([raw file](https://github.com/cockroachdb/pebble/blob/fbb534f9d58e0ae956b3263c5dda572283c51ac6/internal/base/comparer.go#L314)):*
> ```Go
> 	FormatKey: DefaultFormatter,
> 
> 	Separator: func(dst, a, b []byte) []byte {
> ```

I'm having trouble understanding exactly how/why this works.. Can a and b be full keys (or just prefixes)? I assume it's full keys, since sstable blocks can stop in-between different suffixes of the same prefix. In that case,  think the byte-wise thing we're doing here doesn't work in general, e.g for `foo@12` and `foo@10` we would get `foo@13` instead of `foo@11`

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4303)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/base/comparer.go` line 314 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4303#-OHn_bUXAKaHOVyE_5EV:-OHnacxf1Po4WFHu8pI1:b-56xwfq) ([raw file](https://github.com/cockroachdb/pebble/blob/fbb534f9d58e0ae956b3263c5dda572283c51ac6/internal/base/comparer.go#L314)):*

I'm having trouble understanding exactly how/why this works.. Can a and b be full keys (or just prefixes)? I assume it's full keys, since sstable blocks can stop in-between different suffixes of the same prefix. In that case,  think the byte-wise thing we're doing here doesn't work in general, e.g for `foo@12` and `foo@10` we would get `foo@13` instead of `foo@11`
</blockquote></details>

Sorry, it might've been confusing that I made any diff at all here—this is the DefaultComparer and we only call into it from cockroachkv by passing the prefix component only. The testkey comparer implements `Separator` similarly, splitting and then calling `DefaultComparer.Separator`

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4303)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/base/comparer.go` line 314 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4303#-OHn_bUXAKaHOVyE_5EV:-OHnefJq45vAABhjE3pm:b-xel0i9) ([raw file](https://github.com/cockroachdb/pebble/blob/fbb534f9d58e0ae956b3263c5dda572283c51ac6/internal/base/comparer.go#L314)):*

Sorry, it might've been confusing that I made any diff at all here—this is the DefaultComparer and we only call into it from cockroachkv by passing the prefix component only. The testkey comparer implements `Separator` similarly, splitting and then calling `DefaultComparer.Separator`
</blockquote></details>

Oh, yes, thanks!

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4303)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`cockroachkvs/cockroachkvs.go` line 77 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4303#-OHneu3q1k5b9KeQfGiy:-OHneu3q1k5b9KeQfGiz:bbgfcpt) ([raw file](https://github.com/cockroachdb/pebble/blob/7bd44c817938ab75990d642b33fdedd10695a7c7/cockroachkvs/cockroachkvs.go#L77)):*
> ```Go
> 		// empty keys are problematic when combined with a synthetic prefix
> 		// (they form a key that has no trailing sentinel byte).
> 		if len(a) == 0 && len(b) > 0 {
> ```

If len(b) == 0, don't we want the same? I think it would fall through the first `return` below and return the empty key a

I think the function should require b > a (and we verify that under invariants), what do you think?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4303)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`cockroachkvs/cockroachkvs.go` line 77 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4303#-OHneu3q1k5b9KeQfGiy:-OHnfvMa5thXR5cImlE5:bkp2r9w) ([raw file](https://github.com/cockroachdb/pebble/blob/7bd44c817938ab75990d642b33fdedd10695a7c7/cockroachkvs/cockroachkvs.go#L77)):*
> I think the function should require b > a (and we verify that under invariants), what do you think?

Unfortunately, we can't because the same user key may be split across multiple data blocks, in which case we must use `a` as the separator.

> If len(b) == 0, don't we want the same? I think it would fall through the first `return` below and return the empty key a

Yeah, if `len(b) == 0` we can't return `[]byte{0x00}` because it sorts greater than `b`. I _think_ we could add a requirement that `len(b) > 0` and verify it under invariants, because I think an empty key is only possible in the  case that there are no data keys at all (in which case we will use `Successor`, not `Separator`).

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4303)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`cockroachkvs/cockroachkvs.go` line 77 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4303#-OHneu3q1k5b9KeQfGiy:-OHnh8Cj6aWEx1g2MREU:b-y623sf) ([raw file](https://github.com/cockroachdb/pebble/blob/7bd44c817938ab75990d642b33fdedd10695a7c7/cockroachkvs/cockroachkvs.go#L77)):*

> I think the function should require b > a (and we verify that under invariants), what do you think?

Unfortunately, we can't because the same user key may be split across multiple data blocks, in which case we must use `a` as the separator.

> If len(b) == 0, don't we want the same? I think it would fall through the first `return` below and return the empty key a

Yeah, if `len(b) == 0` we can't return `[]byte{0x00}` because it sorts greater than `b`. I _think_ we could add a requirement that `len(b) > 0` and verify it under invariants, because I think an empty key is only possible in the  case that there are no data keys at all (in which case we will use `Successor`, not `Separator`).
</blockquote></details>

Hrm, I think we could prohibit empty keys from `Separator` altogether, and only `Successor` needs to handle it. Let me try it

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4303)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`cockroachkvs/cockroachkvs.go` line 77 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4303#-OHneu3q1k5b9KeQfGiy:-OHnwQHg3gLJKhnrrIUo:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/7bd44c817938ab75990d642b33fdedd10695a7c7/cockroachkvs/cockroachkvs.go#L77)):*

Hrm, I think we could prohibit empty keys from `Separator` altogether, and only `Successor` needs to handle it. Let me try it
</blockquote></details>

Done

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4303#-:-OHo2iXJ6cugee2kx375:bnfp4nl)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

jbowens:
TFTR!

PR #4308, opened 2025-01-30
db: fix TestCrashOpenCrashAfterWALCreation flake

### LINKED ISSUE #4306: TestCrashOpenCrashAfterWALCreation flaky under test
```
go test -tags invariants --exec 'stress -p 16' -timeout 0 -v -race . -run TestCrashOpenCrashAfterWALCreation

980 runs so far, 0 failures, over 1m5s

=== RUN   TestCrashOpenCrashAfterWALCreation
    options_test.go:37: Running TestCrashOpenCrashAfterWALCreation with format major version 016
    ingest_test.go:3019: [JOB 1] WAL 000002: {(,000)} stopped reading at offset: (000002.log: 29); replayed 1 keys in 1 batches
    open_test.go:1036: 
                Error Trace:    /Users/radu/go/src/github.com/cockroachdb/pebble/open_test.go:1036
                Error:          "2" is not greater than "2"
                Test:           TestCrashOpenCrashAfterWALCreation
--- FAIL: TestCrashOpenCrashAfterWALCreation (0.00s)
FAIL
```

RaduBerinde:
The errorfs-based instrumentation had a race between the wal creation actually happening on the underlying FS and a dir sync (and/or between the dir sync actually happening on the underlying FS and a subsequent operation).

We use a custom instrumented FS that acts *after* the operation happens on the MemFS.

Fixes #4306

### DIFF
--- a/open_test.go
+++ b/open_test.go
@@ -994,33 +994,18 @@ func TestCrashOpenCrashAfterWALCreation(t *testing.T) {
 	// file is created and after a subsequent directory sync occurs. This
 	// simulates a crash after the new log file is created and synced.
 	crashFS = nil
-	var crashFSAtomic atomic.Pointer[vfs.MemFS]
+	instrumentedFS := &crashAfterLogCreationFS{MemFS: fs}
 	{
-		var walCreated, dirSynced atomic.Bool
 		d, err := Open("", &Options{
-			FS: errorfs.Wrap(fs, errorfs.InjectorFunc(func(op errorfs.Op) error {
-				if dirSynced.Load() {
-					crashFSAtomic.CompareAndSwap(nil, fs.CrashClone(vfs.CrashCloneCfg{UnsyncedDataPercent: 0}))
-				}
-				if op.Kind == errorfs.OpCreate && filepath.Ext(op.Path) == ".log" {
-					walCreated.Store(true)
-				}
-				// Record when there's a sync of the data directory after the
-				// WAL was created. The data directory will have an empty
-				// path because that's what we passed into Open.
-				if op.Kind == errorfs.OpFileSync && op.Path == "" && walCreated.Load() {
-					dirSynced.Store(true)
-				}
-				return nil
-			})),
+			FS:     instrumentedFS,
 			Logger: testLogger{t},
 		})
 		require.NoError(t, err)
 		require.NoError(t, d.Close())
 	}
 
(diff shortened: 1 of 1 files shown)


RaduBerinde:
`48049 runs so far, 0 failures, over 53m5s`

jbowens (approved):
:lgtm: 

One thought is that we could add a new `errorfs.Predicate` like:

```
func After(pred Predicate) Predicate {
        return &after{pred: pred}
}

type after struct {
	pred      Predicate
        triggered atomic.Bool
}

func (a *after) String() string {
	return fmt.Sprintf("(After %s)", a.pred)
}

func (a *after) Evaluate(op Op) bool {
        if a.triggered.Load() {
                return true
        }
        if pred.Evaluate(op) {
                a.triggered.Store(true)
                return true
        }
	return false
}

```

and use it like
```
errorfs.And(
    errorfs.After(errorfs.And(errorfs.OpCreate, errorfs.PathMatch("*.log")),
    errorfs.After(errorfs.And(errorfs.OpFileSync, errorfs.PathMatch("")),
    errorfs.InjectorFunc(func(op errorfs.Op) error {
        // simulate crash
    }),
)
```

but I think the custom instrumented FS is good too

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4308#-:-OHscQlO2rUswlx1Bl7b:b-u08r80)* status: 0 of 1 files reviewed, all discussions resolved

RaduBerinde (commented):
Interesting, that looks very cool. I'm not quite sure how to combine predicates and injectors, should I add an `errorfs.If(pred, injector) Injector`? Or I could add a function-based predicate and make the last thing a predicate as well (that always returns false) and use `ErrInjected.If()`

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4308#-:-OHsgx34CurG-YDYNu3T:b-7q8aoz)* status: 0 of 1 files reviewed, all discussions resolved

RaduBerinde (commented):
I'd need to add a Predicate that matches the op type too. I added a TODO linking your suggestion for now. TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4308#-:-OHsowwD3au0eNGz-Jws:b-cjbfy8)* status: 0 of 1 files reviewed, all discussions resolved

PR #4310, opened 2025-01-30
objstorage: add FileTypeBlob

jbowens:
Introduce a new file type for blob files. These files will contain the values
  for keys stored in separate sstables. This commit introduces the new file type
  and plumbs some simple machinery, including around deleting obsolete blob files
  subject to the same deletion pacing as sstables.

Informs #112.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2320,7 +2320,7 @@ func (d *DB) cleanupVersionEdit(ve *versionEdit) {
 		// Add this file to zombie tables as well, as the versionSet
 		// asserts on whether every obsolete file was at one point
 		// marked zombie.
-		d.mu.versions.zombieTables[obsoleteFiles[i].DiskFileNum] = tableInfo{
+		d.mu.versions.zombieTables[obsoleteFiles[i].DiskFileNum] = objectInfo{
 			fileInfo: fileInfo{
 				FileNum:  obsoleteFiles[i].DiskFileNum,
 				FileSize: obsoleteFiles[i].Size,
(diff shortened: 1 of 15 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4310)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi)

___
*[`testdata/version_check_consistency` line 34 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4310#-OHt7XpB7nV2HfjzJCE-:-OHt7XpB7nV2HfjzJCE0:b-3kltic) ([raw file](https://github.com/cockroachdb/pebble/blob/166af5668fec9244d7acc5bf020d35e7651edd4c/testdata/version_check_consistency#L34)):*
> ```
>   000001:11
> ----
> L0: 000001: file 000001 (type ‹×›) unknown to the objstorage provider: file does not exist
> ```

I don't understand why this now gets redacted. Even if I `errors.Safe(...)` it, it gets redacted. It seems like it only previously did not get redacted because we were using the `%d` format specifier?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4310)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi)

___
*[`testdata/version_check_consistency` line 34 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4310#-OHt7XpB7nV2HfjzJCE-:-OHtBii94hMQY_SVTtgq:b6o6wii) ([raw file](https://github.com/cockroachdb/pebble/blob/166af5668fec9244d7acc5bf020d35e7651edd4c/testdata/version_check_consistency#L34)):*

I don't understand why this now gets redacted. Even if I `errors.Safe(...)` it, it gets redacted. It seems like it only previously did not get redacted because we were using the `%d` format specifier?
</blockquote></details>

Ah, duh, they weren't marked as safe when I passed them to `redact.SafePrinter`

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4310#-:-OIBLvaeDcdV3-hSD9ca:bnfp4nl)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi)

jbowens:
TFTR!

PR #4312, opened 2025-01-31
db: add CacheSize option

RaduBerinde:
Add a `CacheSize` option which is used when a `Cache` is not
specified. This removes the need for the awkward `NewCache` option
parsing hook, which creates a `Cache` that has to be `Unref`ed
elsewhere.

### DIFF
--- a/cmd/pebble/db.go
+++ b/cmd/pebble/db.go
@@ -55,10 +55,8 @@ type pebbleDB struct {
 }
 
 func newPebbleDB(dir string) DB {
-	cache := pebble.NewCache(cacheSize)
-	defer cache.Unref()
 	opts := &pebble.Options{
-		Cache:                       cache,
+		CacheSize:                   cacheSize,
 		Comparer:                    &cockroachkvs.Comparer,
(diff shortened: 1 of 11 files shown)


jbowens (approved):
Nice. When we bump Pebble in Cockroach, there's a `engineConfig.cacheSize` field that we should be able to remove, instead setting `Options.CacheSize`


:lgtm:

> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4312#-:-OI1NEU71387hX-er8ID:b20e2kl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aadityasondhi)

RaduBerinde (commented):
Nice, will do that. TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4312#-:-OI1y5sm1358GUH56xTH:bo8j1mj)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aadityasondhi)

PR #4315, opened 2025-02-02
cockroachkvs: tolerate empty logical keys in Separator

RaduBerinde:
CRDB has some tests for the Separator which pass empty logical
keys (which encode to a 00) with timestamps. Because the Separator
implementation calls into the default comparer implementation with
just the logical, it will result in a panic.

Tolerate this case by just returning `a`.


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4315#-:-OIBB9HU2Qtgq0o2Nxi5:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #4317, opened 2025-02-04
db: key fileCache by handle

RaduBerinde:
#### db: key fileCache by handle

The `fileCache` is integrated with the block Cache and uses the
`cache.ID` to separate files from different stores.

We can achieve the same by keying on the `*fileCacheHandle` instead of
`cache.ID`, which simplifies things a bit. It would also allow using
a single file cache and multiple block caches if we ever need that
(perhaps for value separation).

### DIFF
--- a/db.go
+++ b/db.go
@@ -1408,12 +1408,10 @@ func (i *Iterator) constructPointIter(
 	// accumulator for this iterator's configured category and QoS. All SSTable
 	// iterators created by this Iterator will accumulate their stats to it as
 	// they Close during iteration.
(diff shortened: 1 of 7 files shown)


jbowens (approved):
> Reviewed 3 of 3 files at r1, 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4317)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #4320, opened 2025-02-04
db: remove type redeclaration of base.FileType

jbowens:
We were inconsistent in our use of the package-local redeclarations of base.FileType and base.FileNum, and they didn't provide in much in convenience or legibility. Remove the base.FileType redeclaration entirely and remove uses of pebble.FileNum that aren't part of the package interface.


RaduBerinde (approved):
:lgtm:

Nice, I never liked these.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4320#-:-OIHZruQ-DxuavIvYGFP:bipqpd7)* status: 0 of 25 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @jbowens)

___
*[`filenames.go` line 12 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4320#-OIHZlib15ooy6mrWnT2:-OIHZlib15ooy6mrWnT3:bgpule5) ([raw file](https://github.com/cockroachdb/pebble/blob/35df19611c104934f3eca5777bc3421500f50952/filenames.go#L12)):*
> ```Go
> 
> // FileNum is an identifier for a file within a database.
> type FileNum = base.FileNum
> ```

CRDB uses this in `storage/pebble.go`

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4320)* status: 0 of 26 files reviewed, 1 unresolved discussion (waiting on @aadityasondhi and @RaduBerinde)

___
*[`filenames.go` line 12 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4320#-OIHZlib15ooy6mrWnT2:-OILwifVAKgazDWWdzuA:b-oz8o77) ([raw file](https://github.com/cockroachdb/pebble/blob/35df19611c104934f3eca5777bc3421500f50952/filenames.go#L12)):*

CRDB uses this in `storage/pebble.go`
</blockquote></details>

Good catch

PR #4323, opened 2025-02-05
colblk: optimizations to PrefixBytesBuilder.Put and DetermineUintEncoding

### LINKED ISSUE #4322: colblk: optimize DetermineUintEncoding
In a TPCC profile, `DetermineUintEncoding` accounts for 0.1% usage. Not a lot, but worth optimizing. A few of the hot path callers pass 0 for min value, so we could eliminate the delta part for those callers.

RaduBerinde:
#### colblk: improved byteWidth implementation

```
name              old time/op  new time/op  delta
ByteWidth/n=1-10  1.06ns ± 1%  0.63ns ± 0%  -40.84%  (p=0.001 n=7+7)
ByteWidth/n=2-10  1.05ns ± 0%  0.63ns ± 0%  -40.60%  (p=0.000 n=8+8)
ByteWidth/n=4-10  1.06ns ± 0%  0.63ns ± 0%  -40.71%  (p=0.000 n=8+7)
ByteWidth/n=8-10  1.05ns ± 0%  0.63ns ± 0%  -40.61%  (p=0.000 n=8+8)
```

#### colblk: add DetermineUintEncodingNoDelta

Add a more efficient variant used for offsets, where the minimum value
is always 0.

#### colblk: minor optimizations to PrefixBytesBuilder.Put

Reorganize the code so that we only check one `if` in the common path.
Benchmark shows a slight improvement (I expect it to be a higher when
the keys are very small).

```
PrefixBytes/alphaLen=2/building-10    710µs ± 0%   708µs ± 0%  -0.34%  (p=0.029 n=4+4)
```

#### colblk: move completedBundleLen out of sizings

This value is not used outside of `Put` so we don't need to keep two
of its values around.

----

Fixes #4322

### DIFF
--- a/sstable/colblk/prefix_bytes.go
+++ b/sstable/colblk/prefix_bytes.go
@@ -781,7 +781,7 @@ func (sz *prefixBytesSizing) String() string {
 func (b *PrefixBytesBuilder) Put(key []byte, bytesSharedWithPrev int) {
 	currIdx := b.nKeys & 1 // %2
 	curr := &b.sizings[currIdx]
-	prev := &b.sizings[1-currIdx]
+	prev := &b.sizings[currIdx^1]
 
 	if invariants.Enabled {
 		if len(key) == 0 {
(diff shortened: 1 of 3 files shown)


jbowens (approved):
:lgtm:

nice!

> Reviewed 3 of 3 files at r1, 3 of 3 files at r2, 3 of 3 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4323#-:-OIMYdRd13uoNhmCAIDf:blzn4tv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #4324, opened 2025-02-05
db: simplify EnsureDefaults

RaduBerinde:
Simplify the awkward `EnsureDefaults` API; it no longer accepts a nil
receiver.

### DIFF
--- a/compaction_picker_test.go
+++ b/compaction_picker_test.go
@@ -367,7 +367,7 @@ func TestCompactionPickerEstimatedCompactionDebt(t *testing.T) {
 }
 
 func TestCompactionPickerL0(t *testing.T) {
-	opts := (*Options)(nil).EnsureDefaults()
+	opts := DefaultOptions()
 	opts.Experimental.L0CompactionConcurrency = 1
 
 	parseMeta := func(s string) (*fileMetadata, error) {
@@ -585,7 +585,7 @@ func TestCompactionPickerL0(t *testing.T) {
 }
 
 func TestCompactionPickerConcurrency(t *testing.T) {
-	opts := (*Options)(nil).EnsureDefaults()
+	opts := DefaultOptions()
 	opts.Experimental.L0CompactionConcurrency = 1
 
 	parseMeta := func(s string) (*fileMetadata, error) {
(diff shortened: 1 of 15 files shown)


jbowens (approved):
:lgtm:

> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4324#-:-OILubXP0AtKo16hAlvT:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`options.go` line 1296 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4324#-OILuC_mDv4ZFkYpaxTs:-OILuC_mDv4ZFkYpaxTt:b-iasxgw) ([raw file](https://github.com/cockroachdb/pebble/blob/8b8192ee10b590406e4b05eb9fede4197f54aa4b/options.go#L1296)):*
> ```Go
> // WithFSDefaults configures the Options to wrap the configured filesystem with
> // the default virtual file system middleware, like disk-health checking.
> func (o *Options) WithFSDefaults() *Options {
> ```

should we also update `WithFSDefaults` to no longer return the `*Options` (and maybe rename to `EnsureFSDefaults`)?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4324)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`options.go` line 1296 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4324#-OILuC_mDv4ZFkYpaxTs:-OIN5KPfDG5a8t84hifp:bkp2irx) ([raw file](https://github.com/cockroachdb/pebble/blob/8b8192ee10b590406e4b05eb9fede4197f54aa4b/options.go#L1296)):*

should we also update `WithFSDefaults` to no longer return the `*Options` (and maybe rename to `EnsureFSDefaults`)?
</blockquote></details>

Good idea, done. I left the name since `EnsureFSDefaults` sounds like it'd be fine to call it multiple times which isn't the case.

PR #4325, opened 2025-02-06
db: relax type of fileCache's closeHook

jbowens:
The file cache implements a closure it calls the 'closeHook' which is propagated to opened iterators to be invoked when the iterators are closed. This hook is responsible for releasing hte ref count on the open sstable Reader. In race builds, it's also used to track the stack traces of iterator opens to aid debugging leaks of open iterators. This closeHook previously required the iterator pass in itself as a sstable.Iterator interface value, but the closeHook really only required its pointer and none of the interface methods. It also returned an error that was never non-nil. This commit removes the error return value and updates the iterator parameter to be any type.

This ensures that other usages of the file cache, such as blob files, can use close hooks with non-sstable.Iterator types.

### DIFF
--- a/file_cache.go
+++ b/file_cache.go
@@ -338,8 +338,6 @@ type fileCacheShard struct {
 	mu struct {
 		sync.RWMutex
 		nodes map[fileCacheKey]*fileCacheNode
-		// The iters map is only created and populated in race builds.
-		iters map[io.Closer][]byte
 
 		handHot  *fileCacheNode
 		handCold *fileCacheNode
@@ -349,6 +347,11 @@ type fileCacheShard struct {
 		sizeHot    int
 		sizeCold   int
 		sizeTest   int
+
+		race struct {
+			// The iters map is only created and populated in race builds.
+			iters map[any][]byte
+		}
 	}
 	releasing       sync.WaitGroup
 	releasingCh     chan *fileCacheValue
@@ -365,7 +368,7 @@ func (c *fileCacheShard) init(size int) {
 	go c.releaseLoop()
 
 	if invariants.RaceEnabled {
-		c.mu.iters = make(map[io.Closer][]byte)
+		c.mu.race.iters = make(map[any][]byte)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4325)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4325#-OIQzzekC45GDh8rH4iG:-OIQzzekC45GDh8rH4iH:bg7igkv) (no related file):*
:lgtm: 
___
*[`file_cache.go` line 341 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4325#-OIQzIwx2zDhvEtoFDsD:-OIQzIwx2zDhvEtoFDsE:b-n2y61p) ([raw file](https://github.com/cockroachdb/pebble/blob/d97e4b938cb4faaa52a46baa21036cfcd990d29e/file_cache.go#L341)):*
> ```Go
> 		sync.RWMutex
> 		nodes map[fileCacheKey]*fileCacheNode
> 		// The iters map is only created and populated in race builds.
> ```

[nit] maybe wrap it in a `struct raceOnly { }` and move it at the end.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4325)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4325#-OIQzzekC45GDh8rH4iG:-OIRmDTLBuZIkalyW1IH:b-814g8k) (no related file):*

:lgtm:
</blockquote></details>

TFTR!
___
*[`file_cache.go` line 341 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4325#-OIQzIwx2zDhvEtoFDsD:-OIRmEXZ1OUKsy9lXxQy:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/d97e4b938cb4faaa52a46baa21036cfcd990d29e/file_cache.go#L341)):*

[nit] maybe wrap it in a `struct raceOnly { }` and move it at the end.
</blockquote></details>

Done

PR #4326, opened 2025-02-06
cache: introduce cache.Handle

RaduBerinde:
We currently use the block cache by plumbing a `cache.ID` together
with the `*Cache` reference. This is a bit cumbersome and error-prone.

This change adds a `cache.Handle` as an encapsulation for a
`(*Cache, handleID)` pair.

In the future, the Cache can keep track of the open handles and we can
move more logic into the cache (for example redistributing the block
load semaphore capacity among stores).

We no longer need to expose reference counting -  we can replace
`Cache.Unref()` with a `Cache.Close()` which is called after all
handles are closed. This will require refactoring a few code paths
which don't have a convenient place to call `Close()` so it is kept
out of this PR.


jbowens (approved):
:lgtm:

> Reviewed 38 of 38 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4326#-:-OImzs2BEv68HFOVN4SG:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sumeerbhola)

RaduBerinde:
TFTR!

PR #4329, opened 2025-02-07
rowblk: remove unused Iter methods, clean up RawIter comments

jbowens:
Remove the unused and unnecessary Key and Value methods on the Iter type. Additionally, fix up the comments on RawIter which incorrectly referred to the internal iterator interface which RawIter no longer implements.


RaduBerinde (approved):

jbowens:
TFTR!

PR #4331, opened 2025-02-11
record: add a WAL read ahead to disambiguate WAL replay errors

EdwardX29:
readAheadForCorruption is a function added to record to disambiguate between errors when replaying WAL. The read ahead process scans ahead in the WAL to detect corruptions based a new promised offset wire format. When a potential corruption / garbage value is read in the WAL, this function reads ahead until it detects an offset that promises durability beyond the previous corruption offset. If the read ahead, finds a confirmation offset, then the original error will be returned. If reading ahead finds the end of file without corruption confirmation, then ErrUnexpectedEOF is returned to signify an abrupt end of the WAL. This helps to differenatiate between errors in replayWAL as ErrInvalidChunk and ErrZeroedChunk can be identified and are no longer silently ignored.

### DIFF
--- a/record/record.go
+++ b/record/record.go
@@ -68,12 +68,12 @@
 // first, middle or last chunk of a multi-chunk record. A multi-chunk record
 // has one first chunk, zero or more middle chunks, and one last chunk.
 //
-// The recyclyable chunk format is similar to the legacy format, but extends
+// The recyclable chunk format is similar to the legacy format, but extends
 // the chunk header with an additional log number field. This allows reuse
 // (recycling) of log files which can provide significantly better performance
 // when syncing frequently as it avoids needing to update the file
 // metadata. Additionally, recycling log files is a prequisite for using direct
-// IO with log writing. The recyclyable format is:
+// IO with log writing. The recyclable format is:
 //
 //	+----------+-----------+-----------+----------------+--- ... ---+
 //	| CRC (4B) | Size (2B) | Type (1B) | Log number (4B)| Payload   |
@@ -83,6 +83,21 @@
 // extra "recyclable" chunk types that map directly to the legacy chunk types
 // (i.e. full, first, middle, last). The CRC is computed over the type, log
 // number, and payload.
+//
+// The WAL sync chunk format allows for detection of data corruption in some
+// circumstances. The WAL sync format extends the recyclable header with an
+// additional offset field. This allows "reading ahead" to be done in order to
+// decipher whether an invalid or zeroed chunk was an artifact of corruption or the
+// logical end of the log. SyncOffset is a promise that the log should have been
+// synced up until the offset. A promised synced offset is needed because cloud
+// providers  may write blocks out of order, rendering "read aheads" scanning for
+// logNum inaccurate.
+// The WAL sync format is:
+//	+----------+-----------+-----------+----------------+------------------+--- ... ---+
+//	| CRC (4B) | Size (2B) | Type (1B) | Log number (4B)| Sync Offset (8B) | Payload   |
+//	+----------+-----------+-----------+----------------+------------------+--- ... ---+
+//
+
 package record
 
 // The C++ Level-DB code calls this the log, but it has been renamed to record
@@ -93,6 +108,7 @@ package record
 import (
 	"encoding/binary"
 	"io"
+	"math"
 
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
@@ -112,6 +128,11 @@ const (
 	recyclableFirstChunkEncoding  = 6
 	recyclableMiddleChunkEncoding = 7
 	recyclableLastChunkEncoding   = 8
+
+	walSyncFullChunkEncoding   = 9
+	walSyncFirstChunkEncoding  = 10
+	walSyncMiddleChunkEncoding = 11
+	walSyncLastChunkEncoding   = 12
 )
 
 const (
@@ -119,9 +140,10 @@ const (
 	blockSizeMask        = blockSize - 1
 	legacyHeaderSize     = 7
 	recyclableHeaderSize = legacyHeaderSize + 4
+	walSyncHeaderSize    = recyclableHeaderSize + 8
 )
 
-// chunkPosition represents the type of a chunk in the WAL.
+// chunkPosition represents the type of a chunk in the log.
 // Records can be split into multiple chunks and marked
 // by the following position types:
 // - invalidChunkPosition: an invalid chunk
@@ -149,6 +171,7 @@ const (
 	invalidWireFormat wireFormat = iota
 	legacyWireFormat
 	recyclableWireFormat
+	walSyncWireFormat
 )
 
 // headerFormat represents the format of a chunk which has
@@ -170,6 +193,10 @@ var headerFormatMappings = [...]headerFormat{
 	recyclableFirstChunkEncoding:  {chunkPosition: firstChunkPosition, wireFormat: recyclableWireFormat, headerSize: recyclableHeaderSize},
 	recyclableMiddleChunkEncoding: {chunkPosition: middleChunkPosition, wireFormat: recyclableWireFormat, headerSize: recyclableHeaderSize},
 	recyclableLastChunkEncoding:   {chunkPosition: lastChunkPosition, wireFormat: recyclableWireFormat, headerSize: recyclableHeaderSize},
+	walSyncFullChunkEncoding:      {chunkPosition: fullChunkPosition, wireFormat: walSyncWireFormat, headerSize: walSyncHeaderSize},
+	walSyncFirstChunkEncoding:     {chunkPosition: firstChunkPosition, wireFormat: walSyncWireFormat, headerSize: walSyncHeaderSize},
+	walSyncMiddleChunkEncoding:    {chunkPosition: middleChunkPosition, wireFormat: walSyncWireFormat, headerSize: walSyncHeaderSize},
+	walSyncLastChunkEncoding:      {chunkPosition: lastChunkPosition, wireFormat: walSyncWireFormat, headerSize: walSyncHeaderSize},
 }
 
 var (
@@ -219,6 +246,10 @@ type Reader struct {
 	err error
 	// buf is the buffer.
 	buf [blockSize]byte
+	// invalidOffset is the first encountered chunk offset found during nextChunk()
+	// that had garbage values. It is used to clarify whether or not a garbage chunk
+	// encountered during WAL replay was the logical EOF or confirmed corruption.
+	invalidOffset uint64
 }
 
 // NewReader returns a new reader. If the file contains records encoded using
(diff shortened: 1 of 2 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4331)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @aadityasondhi and @jbowens)

___
*[`record/record.go` line 406 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIqMvGD70QH962zXPA1:-OIqMvGD70QH962zXPA2:b-sbrxku) ([raw file](https://github.com/cockroachdb/pebble/blob/e5f0302a5a9e658190770f5768a326d91e3fe895/record/record.go#L406)):*
> ```Go
> // or encountering end of file when loading a new block.
> func (r *Reader) readAheadForCorruption() error {
> 	wantFirst := false
> ```

[nit] Could use a comment
___
*[`record/record.go` line 434 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIqMcNa2PzJgI3eNDoa:-OIqMcNa2PzJgI3eNDob:bom4ar6) ([raw file](https://github.com/cockroachdb/pebble/blob/e5f0302a5a9e658190770f5768a326d91e3fe895/record/record.go#L434)):*
> ```Go
> 			chunkType, wireFormat, headerSize := headerFormat.chunkPosition, headerFormat.wireFormat, headerFormat.headerSize
> 			if checksum == 0 && length == 0 && chunkType == invalidChunkPosition {
> 				wantFirst = false
> ```

These cases indicate corruption, no? Don't we need to stop here in this case and not continue reading more blocks?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4331)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @aadityasondhi, @EdwardX29, and @jbowens)

___
*[`record/record.go` line 406 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIqMvGD70QH962zXPA1:-OIqY9W71oNDPIUBOvZn:b-ok4bpp) ([raw file](https://github.com/cockroachdb/pebble/blob/e5f0302a5a9e658190770f5768a326d91e3fe895/record/record.go#L406)):*

[nit] Could use a comment
</blockquote></details>

It would be more clear if we flipped this and called it `insideMultiChunkRecord`
___
*[`record/record.go` line 417 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIqXKSF0bFKeca_3OO1:-OIqXKSF0bFKeca_3OO2:b-6ysdd6) ([raw file](https://github.com/cockroachdb/pebble/blob/e5f0302a5a9e658190770f5768a326d91e3fe895/record/record.go#L417)):*
> ```Go
> 			// when the io library makes calls to Reader.Read().
> 			return io.ErrUnexpectedEOF
> 		}
> ```

How come we ignore any I/O errors here?


Also, is it legal for the last block to be <= 32KB?
___
*[`record/record.go` line 456 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIqXctF-gUW8FbNpc5c:-OIqXctF-gUW8FbNpc5d:b2v37lo) ([raw file](https://github.com/cockroachdb/pebble/blob/e5f0302a5a9e658190770f5768a326d91e3fe895/record/record.go#L456)):*
> ```Go
> 			r.end = r.begin + int(length)
> 			if r.end > r.n {
> 				// The chunk straddles a 32KB boundary (or the end of file).
> ```

Is it legal for the file to end in a partial block? If no, we should just exit above when `io.ReadFull` returns `ErrUnexpectedEOF`.
___
*[`record/record.go` line 477 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIqYUkrBazAdkIhggTZ:-OIqYUkrBazAdkIhggT_:b-mv2gj) ([raw file](https://github.com/cockroachdb/pebble/blob/e5f0302a5a9e658190770f5768a326d91e3fe895/record/record.go#L477)):*
> ```Go
> 			if wantFirst {
> 				if chunkType != fullChunkPosition && chunkType != firstChunkPosition {
> 					break
> ```

How come we are not resetting `wantFirst` here? (we're giving up from reading the rest of the chunks in the block, maybe one of them would be a first chunk)

If we're checking this, should we be also checking that a middle chunk only comes after a first chunk?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4331)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @aadityasondhi and @jbowens)

___
*[`record/record.go` line 375 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIr2HJU9o38mwrPj6JE:-OIr2HJVCQKPPon5Qvyv:b546rld) ([raw file](https://github.com/cockroachdb/pebble/blob/4dd4f6580cbbc5a7146a9d72f68ea28f82c10e75/record/record.go#L375)):*
> ```Go
> // due to an offset (encoded in a chunk header) exceeding the invalid offset,
> // or encountering end of file when loading a new block.
> func (r *Reader) readAheadForCorruption() error {
> ```

[nit] I'd document the return cases. It seems like we always return either `UnexpectedEOF` or some other error?
___
*[`record/record.go` line 441 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIr2-X5-LKusReP95Xp:-OIr2-X5-LKusReP95Xq:b-62yitj) ([raw file](https://github.com/cockroachdb/pebble/blob/4dd4f6580cbbc5a7146a9d72f68ea28f82c10e75/record/record.go#L441)):*
> ```Go
> 				// the invalid offset must have been corruption.
> 				if offset > r.invalidOffset {
> 					return r.err
> ```

Who sets this `r.err`? If the "contract" is that this function is called only after an error is recorded in `r.err`, we should mention it in the function comment. It's a bit hard to review when I don't have a good understanding of how it's going to be used.. In the future I'd aim to have both sides of the change in a single PR (perhaps with multiple commits).

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4331#-:-OIrKGa-6a8TQEYCwopF:bnfp4nl)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @aadityasondhi, @EdwardX29, and @jbowens)

jbowens (approved):
:lgtm_strong: 

Just some comment nits

> Reviewed 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4331#-:-OIuzj5TC5qxTlJHep2Y:brqecme)* status: 1 of 2 files reviewed, 9 unresolved discussions (waiting on @aadityasondhi, @EdwardX29, and @RaduBerinde)

___
*[`record/record.go` line 89 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIut3xDB7F-FK5EO3wK:-OIut3xDB7F-FK5EO3wL:b-qdszjg) ([raw file](https://github.com/cockroachdb/pebble/blob/c28e9e5f61151c515542067482228e149541d32e/record/record.go#L89)):*
> ```Go
> // The wire format allows for limited recovery in the face of data corruption:
> // on a format error (such as a checksum mismatch), the reader moves to the
> // next block and looks for the next full or first chunk.
> ```

This paragraph could use some rewording now. Maybe remove the paragraph entirely and prefix the subsequent paragraph with "The wire format allows for detection of data corruption in some circumstances."
___
*[`record/record.go` line 93 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIutlUaDjkAzoB3QN-6:-OIutlUaDjkAzoB3QN-7:b-1wjsuc) ([raw file](https://github.com/cockroachdb/pebble/blob/c28e9e5f61151c515542067482228e149541d32e/record/record.go#L93)):*
> ```Go
> // The WAL sync chunk format extends the recyclable header with an additional
> // offset field. This allows "reading ahead" to be done in order to decipher
> // whether an invalid or zeroed chunk was a corruption or the logical end of WAL.
> ```

nit: "was a artifact of corruption or the logical end of the WAL"
___
*[`record/record.go` line 94 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIutqMAElhVD5v9tJ4k:-OIutqMAElhVD5v9tJ4l:b1a1dp) ([raw file](https://github.com/cockroachdb/pebble/blob/c28e9e5f61151c515542067482228e149541d32e/record/record.go#L94)):*
> ```Go
> // offset field. This allows "reading ahead" to be done in order to decipher
> // whether an invalid or zeroed chunk was a corruption or the logical end of WAL.
> // Offset is a promise that the WAL should have been synced up until the offset.
> ```

nit: can we refer to it as 'SyncOffset' or something of the like that helps indicate what the encoded offset _is_?
___
*[`record/record.go` line 394 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIuv8Y55s2CST5_Ny01:-OIuv8Y55s2CST5_Ny02:btgv0ev) ([raw file](https://github.com/cockroachdb/pebble/blob/c28e9e5f61151c515542067482228e149541d32e/record/record.go#L394)):*
> ```Go
> 		// Load the next block into r.buf.
> 		n, err := io.ReadFull(r.r, r.buf[:])
> 		if err == io.EOF {
> ```

nit: let's use `errors.Is(err, io.EOF)`; we have some nontrivial code that runs within the `vfs.FS` layer and may wrap errors surfaced
___
*[`record/record.go` line 399 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIuvTzoBVxzz_ps9BoW:-OIuvTzp9sO3BXJ5JGPx:b-d3vy6u) ([raw file](https://github.com/cockroachdb/pebble/blob/c28e9e5f61151c515542067482228e149541d32e/record/record.go#L399)):*
> ```Go
> 			// an error when it is io.EOF. io.ErrUnexpectedEOF
> 			// is returned so that the error is not cleared
> 			// when the io library makes calls to Reader.Read().
> ```

nit: let's also add a sentence or two here justifying it semantically, something like "Because we did not encounter a sync offset indicating the invalid chunk should've been valid, the invalid chunk is an unclean, abrupt ending of the logical log. A `io.ErrUnexpectedEOF` represents this abrupt end of file." 
___
*[`record/record.go` line 402 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIux27Q6bLwIaVsMhCg:-OIux27Q6bLwIaVsMhCh:b-5ok3os) ([raw file](https://github.com/cockroachdb/pebble/blob/c28e9e5f61151c515542067482228e149541d32e/record/record.go#L402)):*
> ```Go
> 			return io.ErrUnexpectedEOF
> 		}
> 		// The last block of a WAL can be less than 32KiB, which is
> ```

nit: let's use "log" instead of "WAL", because this log format is used for manfiest files too
___
*[`record/record.go` line 403 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4331#-OIuz72354_H3uRHXX1u:-OIuz72354_H3uRHXX1v:bfoyzdh) ([raw file](https://github.com/cockroachdb/pebble/blob/c28e9e5f61151c515542067482228e149541d32e/record/record.go#L403)):*
> ```Go
> 		}
> 		// The last block of a WAL can be less than 32KiB, which is
> 		// the length of r.buf. Thus, we should still read the data in
> ```

nit: maybe "parse the data" instead of "read the data," because we already _read_ it  into the buf and now need to parse it into chunks

PR #4332, opened 2025-02-11
cache: Load before Store

### LINKED ISSUE #4328: cache: check value before setting referenced atomic
In a TPCC profile the `referenced.Store(true)` in the cache hit path takes about 0.1%. Not a lot, but we should try to check the value first and only store if necessary. An atomic load is a simple load, whereas an atomic store is an atomic XCHG operation which involves a memory barrier.

In the profile, the main code paths where through reading an index and a filter block; it is likely that these are accessed much more frequently than the clockpro hands make a full rotation.

Image

Jira issue: PEBBLE-340

RaduBerinde:
#### cache: Load before Store

On x86 an atomic.Load is a simple load, whereas a Store is an atomic
`XCHG` operation which involves a memory barrier. Frequently accessed
blocks are likely to be accessed many times between two sweeps of the
clockpro hand.

On a gceworker:
```
name         old time/op    new time/op    delta
CacheGet-24    20.1ns ± 4%    18.9ns ± 5%  -5.82%  (p=0.002 n=10+10)
```

Fixes #4328

### DIFF
--- a/internal/cache/cache_test.go
+++ b/internal/cache/cache_test.go
@@ -251,13 +251,14 @@ func TestCacheStressSetExisting(t *testing.T) {
 func BenchmarkCacheGet(b *testing.B) {
 	const size = 100000
 
-	cache := newCache(size, 1)
+	n := runtime.GOMAXPROCS(0)
+	cache := newCache(size*int64(n), n)
 	defer cache.Unref()
 	h := cache.NewHandle()
 	defer h.Close()
 
 	for i := 0; i < size; i++ {
-		setTestValue(h, 0, 0, "a", 1)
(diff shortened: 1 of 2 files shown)


jbowens (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4332#-:-OIr70KlEURAAM4ONwN5:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #4334, opened 2025-02-11
db: embed block.ReadEnv in internalIterOpts

jbowens:
The internalIterOpts struct had all the fields of a block.ReadEnv. Replace them with a block.ReadEnv field. Additionally, update instances where we construct internalIterOpts to eagerly retrieve a categorized iterator stats accumulator.


RaduBerinde (approved):

jbowens:
TFTR!

PR #4337, opened 2025-02-12
internal/manifest: add ObsoleteFiles type

jbowens:
Rework the collection of obsolete files during a Version unref to accumulate the obsolete files into a new ObsoleteFiles type. This is in preparation for the introduction of blob files into manifest Versions. Unreferences will also need to unreference blob files and accumulate obsolete blob files.


jbowens:
TFTR!

PR #4340, opened 2025-02-13
db: rework fileCache to use genericcache

RaduBerinde:
#### genericcache: extract the generic part of the file cache

This code extracts the internal logic of the file cache and makes it
into a generally useful type.

#### db: rework fileCache to use genericcache

We no longer intermingle the cache infrastructure logic with the file
cache specifics. This will help when we extend the file cache to
support blob files.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -695,8 +695,10 @@ func TestIterLeak(t *testing.T) {
 		t.Run(fmt.Sprintf("leak=%t", leak), func(t *testing.T) {
 			for _, flush := range []bool{true, false} {
 				t.Run(fmt.Sprintf("flush=%t", flush), func(t *testing.T) {
+					fc := NewFileCache(10, 100)
 					d, err := Open("", testingRandomized(t, &Options{
-						FS: vfs.NewMem(),
+						FS:        vfs.NewMem(),
+						FileCache: fc,
 					}))
 					require.NoError(t, err)
 
@@ -710,7 +712,6 @@ func TestIterLeak(t *testing.T) {
 						require.NoError(t, iter.Close())
 						require.NoError(t, d.Close())
 					} else {
-						defer iter.Close()
 						if err := d.Close(); err == nil {
 							t.Fatalf("expected failure, but found success")
 						} else if !strings.HasPrefix(err.Error(), "leaked iterators:") {
@@ -718,7 +719,9 @@ func TestIterLeak(t *testing.T) {
 						} else {
 							t.Log(err.Error())
 						}
+						iter.Close()
 					}
+					fc.Unref()
 				})
 			}
 		})
@@ -732,13 +735,16 @@ func TestIterLeakSharedCache(t *testing.T) {
 		t.Run(fmt.Sprintf("leak=%t", leak), func(t *testing.T) {
 			for _, flush := range []bool{true, false} {
 				t.Run(fmt.Sprintf("flush=%t", flush), func(t *testing.T) {
+					fc := NewFileCache(10, 100)
 					d1, err := Open("", &Options{
-						FS: vfs.NewMem(),
+						FS:        vfs.NewMem(),
+						FileCache: fc,
 					})
 					require.NoError(t, err)
 
 					d2, err := Open("", &Options{
-						FS: vfs.NewMem(),
+						FS:        vfs.NewMem(),
+						FileCache: fc,
 					})
 					require.NoError(t, err)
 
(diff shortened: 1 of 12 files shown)


jbowens (approved):
> Reviewed 4 of 4 files at r1, 3 of 3 files at r2, 5 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4340)* status: 9 of 12 files reviewed, 8 unresolved discussions

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4340#-OJ-wFut74wgxfE6vFAa:-OJ-wFuu1yLQbNTf2pFP:b-r0pm8i) (no related file):*
nice cleanup

:lgtm: 
___
*[`file_cache.go` line 389 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4340#-OJ-tAjq-zbFbkpNymm4:-OJ-tAjq-zbFbkpNymm5:b31eecd) ([raw file](https://github.com/cockroachdb/pebble/blob/9e381c0f798b7501f5cda9125d4abee3b91261d9/file_cache.go#L389)):*
> ```Go
> //func (c *FileCache) getShard(fileNum base.DiskFileNum) *fileCacheShard {
> //	return c.shards[uint64(fileNum)%uint64(len(c.shards))]
> //}
> ```

nit: detritus?
___
*[`internal/genericcache/cache.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4340#-OJ-nWJTDN5p4WN6pzV5:-OJ-nWJTDN5p4WN6pzV6:b-yyr0wp) ([raw file](https://github.com/cockroachdb/pebble/blob/59682bc6f38f8417166454877a588508be31e46b/internal/genericcache/cache.go#L49)):*
> ```Go
> // New creates a Cache with the given capacity and number of shards.
> //
> // initFn is used to initialize a new value when it is added to the cache.
> ```

nit: s/initFn/initValueFn/
___
*[`internal/genericcache/cache.go` line 50 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4340#-OJ-ngNS0RWa2cm1ClvD:-OJ-ngNS0RWa2cm1ClvE:b-mr3trt) ([raw file](https://github.com/cockroachdb/pebble/blob/59682bc6f38f8417166454877a588508be31e46b/internal/genericcache/cache.go#L50)):*
> ```Go
> //
> // initFn is used to initialize a new value when it is added to the cache.
> // releaseFn is used to close a V when it is no longer needed.
> ```

nit: s/releaseFn/releaseValueFn/
___
*[`internal/genericcache/cache.go` line 80 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4340#-OJ-oA0KCPzuOe2BdVCE:-OJ-oA0KCPzuOe2BdVCF:bkcpny7) ([raw file](https://github.com/cockroachdb/pebble/blob/59682bc6f38f8417166454877a588508be31e46b/internal/genericcache/cache.go#L80)):*
> ```Go
> 
> // FindOrCreate retrieves an existing value or creates a new value for the given
> // key. The result can be accessed via Result.Value(). The caller must call
> ```

nit: ValueRef.Value()?
___
*[`internal/genericcache/cache.go` line 93 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4340#-OJ-pz3iDl4jk-fx_kYL:-OJ-pz3iDl4jk-fx_kYM:bntpsbs) ([raw file](https://github.com/cockroachdb/pebble/blob/59682bc6f38f8417166454877a588508be31e46b/internal/genericcache/cache.go#L93)):*
> ```Go
> 
> // ValueRef is returned by FindOrCreate. It holds a reference on a value; the
> // value will be kept "alive" even if it is evicted from the cache.
> ```

I'm unclear on whether it's okay to have a reference on a value that's been evicted or not... This comment makes it seem like it's ok, but `Evict` and `EvictAll` expect no outstanding references. Maybe 'evicted' in this comment is referring to when the cache itself decides to evict the value to make room for a new value?
___
*[`internal/genericcache/cache.go` line 101 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4340#-OJ-oroeBU0KrCSUZCCK:-OJ-oroeBU0KrCSUZCCL:b-uz7sut) ([raw file](https://github.com/cockroachdb/pebble/blob/59682bc6f38f8417166454877a588508be31e46b/internal/genericcache/cache.go#L101)):*
> ```Go
> 
> // Value returns the value. This method and the returned value can only be used
> // until ref.Close() or Cache.UnrefValue() is called.
> ```

nit: looks like `Cache.UnrefValue()` got removed
___
*[`testdata/metrics` line 78 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4340#-OJ-w3on6QSYXPwBVYU2:-OJ-w3on6QSYXPwBVYU3:b-rqq0b5) ([raw file](https://github.com/cockroachdb/pebble/blob/9e381c0f798b7501f5cda9125d4abee3b91261d9/testdata/metrics#L78)):*
> ```
> Compression types: snappy: 1
> Block cache: 2 entries (716B)  hit rate: 0.0%
> Table cache: 1 entries (952B)  hit rate: 0.0%
> ```

is this 50-byte increase from better accounting of the memory used within the table cache, or an increase in the memory used by the table cache?

PR #4341, opened 2025-02-13
internal/manifest: maintain BlobFileMetadata references

jbowens:
Extend the BlobReference type with a pointer to a shared BlobFileMetadata
struct describing the blob file. Apply reference counting to these
BlobFileMetadata to track the lifecycle of a blob file.

Informs #112.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -1324,17 +1324,15 @@ func (d *DB) newIngestedFlushableEntry(
 	// to.
 	entry := d.newFlushableEntry(f, logNum, seqNum)
 	// The flushable entry starts off with a single reader ref, so increment
-	// the FileBacking.Refs.
+	// the TableMetadata.Refs.
 	for _, file := range f.files {
-		file.FileBacking.Ref()
+		file.Ref()
 	}
 	entry.unrefFiles = func(of *manifest.ObsoleteFiles) {
-		// Invoke Unref on each table. If any files become obsolete, add them to
-		// the obsolete files list.
+		// Invoke Unref on each table. If any files become obsolete, they'll be
+		// added to the set of obsolete files.
 		for _, file := range f.files {
-			if file.FileBacking.Unref() == 0 {
-				of.AddBacking(file.FileBacking)
-			}
+			file.Unref(of)
 		}
 	}
 
--- a/internal/manifest/blob_metadata.go
+++ b/internal/manifest/blob_metadata.go
@@ -5,16 +5,31 @@
 package manifest
 
 import (
+	"sync/atomic"
+
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/humanize"
+	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/redact"
 )
 
 // A BlobReference describes a sstable's reference to a blob value file.
 type BlobReference struct {
-	FileNum   base.DiskFileNum
+	// FileNum identifies the referenced blob file.
+	FileNum base.DiskFileNum
+	// ValueSize is the sum of the lengths of the uncompressed values within the
+	// blob file for which there exists a reference in the sstable.
+	//
+	// INVARIANT: ValueSize <= Metadata.ValueSize
 	ValueSize uint64
+
+	// Metadata is the metadata for the blob file. It is non-nil for blob
+	// references contained within active Versions. It is expected to initially
+	// be nil when decoding a version edit as a part of manfiest replay. When
+	// the version edit is accumulated into a bulk version edit, the metadata
+	// is populated.
+	Metadata *BlobFileMetadata
 }
 
 // BlobFileMetadata is metadata describing a blob value file.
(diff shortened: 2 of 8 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4341#-:-OJ0p81J94ZxBEwciGcZ:bnfp4nl)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`-- commits` line 6 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4341#-OJ0fHu44KpO7-sfrSmg:-OJ0fHu51HNZ2UbiGQ_a:b-rq4ajz):*
[nit] It might be cleaner if you moved the refcount to a `BlobReferences` type that contains that ref count and the slice.

jbowens:
I decided the o(# sstables) additional heap objects, and it’s still conceptually clean to ref count the tables themselves

RaduBerinde:
Sure, though the struct I suggested would be a field, not a separate object.

sumeerbhola (changes requested):
> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4341)* status: 1 of 8 files reviewed, 2 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/manifest/version.go` line 246 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4341#-OJ4c_wu88d2S0B7PfGF:-OJ4c_wu88d2S0B7PfGG:b-mcjlje) ([raw file](https://github.com/cockroachdb/pebble/blob/7730dc8df8b5faf9b37a56f14cdfce1b0780d57a/internal/manifest/version.go#L246)):*
> ```Go
> 	// released.
> 	//
> 	// The reference count is the number of versions with non-zero reference
> ```

Is this really the number of versions, or the number of distinct nodes that hold a reference?

I am trying to understand the choice to add two levels of ref counting, versus continuing to do what we currently do by reaching into `FileBacking` in `node.decRef` and decrementing that ref count. The prototype in blobt2 had code in `Version.unrefFiles` for that. Was the choice in this PR motivated by having cleaner abstraction boundaries in that the btree would simply call `TableMetadata.Unref`? Hmm, maybe not, because we could have `TableMetadata.Unref` only do unreffing of the backing and blob references and add them to `obsoleteFiles`.

Could you elaborate?

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4341)* status: 1 of 8 files reviewed, 2 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/manifest/version.go` line 246 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4341#-OJ4c_wu88d2S0B7PfGF:-OJ4gY4J0cHRuNH-x1y3:b-xiu8zm) ([raw file](https://github.com/cockroachdb/pebble/blob/7730dc8df8b5faf9b37a56f14cdfce1b0780d57a/internal/manifest/version.go#L246)):*

Is this really the number of versions, or the number of distinct nodes that hold a reference?

I am trying to understand the choice to add two levels of ref counting, versus continuing to do what we currently do by reaching into `FileBacking` in `node.decRef` and decrementing that ref count. The prototype in blobt2 had code in `Version.unrefFiles` for that. Was the choice in this PR motivated by having cleaner abstraction boundaries in that the btree would simply call `TableMetadata.Unref`? Hmm, maybe not, because we could have `TableMetadata.Unref` only do unreffing of the backing and blob references and add them to `obsoleteFiles`.

Could you elaborate?
</blockquote></details>

Never mind the question about two levels of ref counting. The blobt2 branch preceded virtual ssts, so it was relying on the ref in the `FileMetadata`. It seems awkward to do refs of blob references when cloning a node, like we do now with the `FileBacking`. This seems cleaner.

sumeerbhola (commented):
> Reviewed 4 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4341)* status: 5 of 8 files reviewed, 4 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/manifest/version_edit.go` line 925 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4341#-OJ4i8dZCzFvyDY24xyi:-OJ4i8dZCzFvyDY24xyj:biv19ul) ([raw file](https://github.com/cockroachdb/pebble/blob/f14d0f0cac6264c8c00bd7a6c371e481e3b0b30b/internal/manifest/version_edit.go#L925)):*
> ```Go
> 		// Deleted holds a list of all blob files that became unreferenced by
> 		// any sstables, making them obsolete within the resulting version (a
> 		// zombie if still referenced by previous versions).
> ```

is there an invariant that Added intersection Deleted is empty?
___
*[`internal/manifest/version_edit.go` line 933 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4341#-OJ4jdQ39xEx1KVKR8Tp:-OJ4jdQ39xEx1KVKR8Tq:b8j9bwe) ([raw file](https://github.com/cockroachdb/pebble/blob/f14d0f0cac6264c8c00bd7a6c371e481e3b0b30b/internal/manifest/version_edit.go#L933)):*
> ```Go
> 		// to a file introduced in VersionEdit.AddedBlobFiles or a file
> 		// referenced by a deleted sstable.
> 		DeletedReferences map[base.DiskFileNum]*BlobFileMetadata
> ```

Is this so that we can make `BlobReference.Metadata` non-nil in `BulkVersionEdit.Accumulate` when opening the DB?

I think the prototype does this fixing in `BulkVersionEdit.Apply` and it was probably a quick decision without much thought. I definitely didn't think through the fact that we have the information in the VersionEdit.

btw, I didn't read the code in version_edit.go carefully since @RaduBerinde already reviewed it.

sumeerbhola (approved):

jbowens (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4341#-:-OK2-DI9A23j3Qeri8Dn:bz5mair)* status: 5 of 8 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`-- commits` line 6 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4341#-OJ0fHu44KpO7-sfrSmg:-OJ4GcE_-0p5CpsFSWa1:b-yxmku6):*

[nit] It might be cleaner if you moved the refcount to a `BlobReferences` type that contains that ref count and the slice.
</blockquote></details>

I decided that O(sstables) additional objects on the heap wasn’t worth it
___
*[`internal/manifest/version.go` line 246 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4341#-OJ4c_wu88d2S0B7PfGF:-OK1yw_lACyYKjtrt3w_:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/7730dc8df8b5faf9b37a56f14cdfce1b0780d57a/internal/manifest/version.go#L246)):*

Never mind the question about two levels of ref counting. The blobt2 branch preceded virtual ssts, so it was relying on the ref in the `FileMetadata`. It seems awkward to do refs of blob references when cloning a node, like we do now with the `FileBacking`. This seems cleaner.
</blockquote></details>

Done.
___
*[`internal/manifest/version_edit.go` line 925 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4341#-OJ4i8dZCzFvyDY24xyi:-OK2-AjrEc3-ZeRjstP3:b-s4o8my) ([raw file](https://github.com/cockroachdb/pebble/blob/f14d0f0cac6264c8c00bd7a6c371e481e3b0b30b/internal/manifest/version_edit.go#L925)):*

is there an invariant that Added intersection Deleted is empty?
</blockquote></details>

Yeah, documented
___
*[`internal/manifest/version_edit.go` line 933 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4341#-OJ4jdQ39xEx1KVKR8Tp:-OK1zRiJCG8ojUgAyJCY:bm7e97f) ([raw file](https://github.com/cockroachdb/pebble/blob/f14d0f0cac6264c8c00bd7a6c371e481e3b0b30b/internal/manifest/version_edit.go#L933)):*

Is this so that we can make `BlobReference.Metadata` non-nil in `BulkVersionEdit.Accumulate` when opening the DB?

I think the prototype does this fixing in `BulkVersionEdit.Apply` and it was probably a quick decision without much thought. I definitely didn't think through the fact that we have the information in the VersionEdit.

btw, I didn't read the code in version_edit.go carefully since @RaduBerinde already reviewed it.
</blockquote></details>

Yeah, that's right

PR #4343, opened 2025-02-14
db: add implementation warning about requirement imposed by SingleDelete

sumeerbhola:


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4343#-:-OJ5zWCpBedC8IMVWMoD:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @pav-kv and @petermattis)

petermattis (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4343)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @pav-kv)

___
*[`db.go` line 154 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4343#-OJ6f5S_2fbYumYFH8Y8:-OJ6f5S_2fbYumYFH8Y9:b-z7ic89) ([raw file](https://github.com/cockroachdb/pebble/blob/146383810c2cc35f6428d987366602cc0869c8d4/db.go#L154)):*
> ```Go
> 	// flushable are usually in one-to-one correspondence, and atomically
> 	// updating the MANIFEST when the flushable is flushed (which ensures the
> 	// WAL will never be replayed). There is one exception: a flushableBatch is
> ```

Might want to add some words to explain what a flushableBatch is. `a flushableBatch (a batch too large to fit in a memtable) ...`.

The logical memtable splitting idea would also violate the 1:1 correspondence between a flushable and WAL. If we decide to precede with that proposal I think we should find a way to refactor the code to more clearly indicate that a contiguous span of flushables needs to be flushed together. It is subtle the way `flushableBatch.readyToFlush()` achieves this. I'm pretty sure I'm to blame for that.

Cc @RaduBerinde as he often has good ideas for cleaner implementations.

sumeerbhola (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4343#-:-OJOpemi49IwNedJ4PiV:bz5mair)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @pav-kv and @RaduBerinde)

___
*[`db.go` line 154 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4343#-OJ6f5S_2fbYumYFH8Y8:-OJOpLah2l_JOUivIDTf:b-6rx197) ([raw file](https://github.com/cockroachdb/pebble/blob/146383810c2cc35f6428d987366602cc0869c8d4/db.go#L154)):*
> Might want to add some words to explain what a flushableBatch is. `a flushableBatch (a batch too large to fit in a memtable) ...`.

Done

> The logical memtable splitting idea would also violate the 1:1 correspondence between a flushable and WAL. If we decide to precede with that proposal I think we should find a way to refactor the code to more clearly indicate that a contiguous span of flushables needs to be flushed together. It is subtle the way `flushableBatch.readyToFlush()` achieves this. I'm pretty sure I'm to blame for that.

Ack

Issue #4285, opened 2025-01-24
db: surface corruption, missing sstable errors without fataling


jbowens:
Some areas of Pebble respond to corruption, including missing sstables, by fataling the process. If the corruption lies in externally-linked sstables, a user may wish to excise the files from the database and continue operating. To support this, we need to surface these errors (eg, through the Download API) and not panic.

Jira issue: PEBBLE-328

RaduBerinde:
https://github.com/cockroachdb/pebble/issues/1192 is related

jbowens:
#4346 refactors Pebble handling of these corruptions errors and introduces a new SSTableCorruption event.

RaduBerinde:
Before we close, we still have to look at how Pebble behaves while the corruption exists (even with the assumption that it will be excised "soon"). E.g. no hot loops on retrying compactions to those files.

PR #4346, opened 2025-02-25
db: rework SSTable corruption handling

RaduBerinde:
This commit reworks SSTable corruption handling. Instead of Pebble
calling `Fatalf` in some cases, and the higher layer (CRDB) calling it
in other cases, we add an `SSTableCorruption` event. The default
handler calls `Fatalf` but a custom handler can be used for different
behavior. We rework the `MustExist` semantics to mark the not-exist
error as a corruption error and rely on the same reporting
mechanism.

We implement this by adding a callback to report corruption to
`block.ReadEnv`. This was necessary because some of the information we
need (like the user key bounds) are only available at a higher level.

Informs #4285


RaduBerinde:
I don't plan to merge until I get the corresponding changes in CRDB working as well.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4346)* status: 0 of 19 files reviewed, 1 unresolved discussion

___
*[`-- commits` line 14 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4346#-OK2dwla-kQ17Lbr65hL:-OK2dwla-kQ17Lbr65hM:b-zcx8i4):*
In the context of online restore, I think we could also rely on the KV-driven `DB.Download` call eventually encountering the corruption error while downloading one of its `DownloadSpan`s, and then we can propagate the corruption error up annotated with the span's bounds.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4346)* status: 0 of 19 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`event.go` line 728 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4346#-OK2q68b8MM2hxVoi2Lj:-OK2q68b8MM2hxVoi2Lk:b2crsep) ([raw file](https://github.com/cockroachdb/pebble/blob/67fa3c9d0f59a4b0bcae5c02d5dc7ef9f68977ed/event.go#L728)):*
> ```Go
> 	// SSTableCorruption is invoked when a corruption is detected in an SSTable.
> 	// It should not block, as it is called synchronously in read paths.
> 	SSTableCorruption func(SSTableCorruptionInfo)
> ```

With the introduction of blob files, we also may encounter corruption within a blob file (or a missing blob file). Should these be modeled as separate events, or should we combine them into a generalized DataCorruption error event?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4346)* status: 0 of 19 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`-- commits` line 14 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4346#-OK2dwla-kQ17Lbr65hL:-OK2zXg84fV2oqNSKUl-:b7zhfxv):*

In the context of online restore, I think we could also rely on the KV-driven `DB.Download` call eventually encountering the corruption error while downloading one of its `DownloadSpan`s, and then we can propagate the corruption error up annotated with the span's bounds.
</blockquote></details>

Right.. But it could take much longer for Download to reach a corrupt span. And we'd still need some revamping of the current mechanism so CRDB doesn't crash when e.g. a non-download compaction fails because of a corruption.
___
*[`event.go` line 728 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4346#-OK2q68b8MM2hxVoi2Lj:-OK3-Imy8q-kRQmvyugD:b-hfhouc) ([raw file](https://github.com/cockroachdb/pebble/blob/67fa3c9d0f59a4b0bcae5c02d5dc7ef9f68977ed/event.go#L728)):*

With the introduction of blob files, we also may encounter corruption within a blob file (or a missing blob file). Should these be modeled as separate events, or should we combine them into a generalized DataCorruption error event?
</blockquote></details>

Good point, I'll work on generalizing in this PR.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4346)* status: 0 of 19 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`event.go` line 728 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4346#-OK2q68b8MM2hxVoi2Lj:-OK3tcpGBPEk9wBj43Wd:b-g4bu9r) ([raw file](https://github.com/cockroachdb/pebble/blob/67fa3c9d0f59a4b0bcae5c02d5dc7ef9f68977ed/event.go#L728)):*

Good point, I'll work on generalizing in this PR.
</blockquote></details>

Done. There wasn't any sstable-specific info in it anyway.

RaduBerinde:
*[`file_cache.go` line 283 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4346#-OK7KoyIDMCEWmEKeXkt:-OK7tIAoC5CRzf-TJb2O:b-l2ol77) ([raw file](https://github.com/cockroachdb/pebble/blob/dd05203dd29c575e9e1f2a49438763a85e484f25/file_cache.go#L283)):*

do any of the `withCommonReader`|`withReader`|`withVirtualReader` calls pass a `fn` that actually uses this env? I think it does make sense to propagate it, because some of the calls pass a `fn` that performs I/O today, just with an implicit `block.NoReadEnv`, but I just want to make sure I understand the motivation
</blockquote></details>

I want to experiment with removing the virtual reader altogether and having a sstable.ReadEnv that contains the block.ReadEnv plus a pointer to virtual table params (that would be cached in TableMetadata). Unless I'm missing something that will be a lot cleaner and more efficient (fewer allocations).

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4346#-:-OKTCaOrD1R14yb7x7ug:b-814g8k)* status: 6 of 19 files reviewed, 5 unresolved discussions (waiting on @jbowens)

___
*[`compaction.go` line 2565 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4346#-OK7HlZM2C4SjnPOvxRo:-OKTBe0z3YM9LUzxnoQE:b-7yd1q4) ([raw file](https://github.com/cockroachdb/pebble/blob/dd05203dd29c575e9e1f2a49438763a85e484f25/compaction.go#L2565)):*

shouldn't this be using a `ReadEnv` with the compaction's `BufferPool` and a CategoryCompaction iter stats accumulator?
</blockquote></details>

Added a TODO.
___
*[`ingest.go` line 2596 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4346#-OK7IvEqDltpQWU9O53Y:-OKTCTw64j9GUJEQuxVj:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/dd05203dd29c575e9e1f2a49438763a85e484f25/ingest.go#L2596)):*

and should these use a `ReadEnv` with  `CategoryIngest` stats collector configured?
</blockquote></details>

Done.
___
*[`internal/base/filenames.go` line 200 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4346#-OK7LYDbALMx0ofcIe9t:-OKTCYiT-I5dYslHCe_8:b41yuvb) ([raw file](https://github.com/cockroachdb/pebble/blob/dd05203dd29c575e9e1f2a49438763a85e484f25/internal/base/filenames.go#L200)):*

should we get rid of `MustExist` now? or rename `fataler` since we explicitly do not want to fatal anymore
</blockquote></details>

We're still using this for the WAL

jbowens (approved):
:lgtm:

> Reviewed 9 of 19 files at r1, 2 of 3 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4346#-:-OKWwP438zjNuFiZKLaC:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

RaduBerinde:
I made a few small changes to add the path as a hint to the error, and to deprecate the `ErrCorruption` export in favor if `IsCorruptionError()`.

PR #4347, opened 2025-02-25
lint: add some roachvet passes

RaduBerinde:
Add a linter test that uses some of the CRDB custom linter passes. We
will add more passes separately (each one requires some fixes).


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4347)* status: all files reviewed, 2 unresolved discussions (waiting on @aadityasondhi and @sumeerbhola)

___
*[`objstorage/objstorageprovider/sharedcache/shared_cache.go` line 716 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4347#-OK1qAPR8rXKt92ccEFX:-OK29ULzC5cuL7JtQD7R:bm0lk1q) ([raw file](https://github.com/cockroachdb/pebble/blob/94bcb32b22a21cb312194bda0cdfef9ce8994a36/objstorage/objstorageprovider/sharedcache/shared_cache.go#L716)):*

was this a real bug?
</blockquote></details>

No, there is a `return` right after. We could make the tool smarter but it seems like a lot of work for not that much benefit.

PR #4348, opened 2025-02-26
internal/manifest: add BlobReferenceDepth and encode in manifest

jbowens:
Add TableMetadata.BlobReferenceDepth, which is an upper bound on the number of blob files that a reader scanning the table would need to keep open if they only open and close each referenced blob file once. The BlobReferenceDepth will be approximated in practice, by assigning the outputs of a compaction a blob reference depth equal to the sum of the highest blob reference depth in each input level of the compaction.

This commit introduces the field and adds it to the encoding and debug output of TableMetadata and VersionEdits.

Informs #112.

### DIFF
--- a/internal/manifest/testdata/version_edit_apply
+++ b/internal/manifest/testdata/version_edit_apply
@@ -231,7 +231,7 @@ L2:
 # that is unknown.
 
 apply v8
-  add-table:     L0 000004:[a#9,SET-z#9,DEL] seqnums:[9-9] points:[a#9,SET-z#9,DEL] blobrefs:[(000003: 25935)]
+  add-table:     L0 000004:[a#9,SET-z#9,DEL] seqnums:[9-9] points:[a#9,SET-z#9,DEL] blobrefs:[(000003: 25935); depth:1]
 ----
 error during Accumulate: blob file 000003 referenced by L0.000004 not found
 
@@ -240,10 +240,10 @@ error during Accumulate: blob file 000003 referenced by L0.000004 not found
 
 apply v8 name=v9
   add-blob-file: 000003 size:[20535 (20KB)] vals:[25935 (25KB)]
-  add-table:     L0 000004:[a#9,SET-z#9,DEL] seqnums:[9-9] points:[a#9,SET-z#9,DEL] blobrefs:[(000003: 25935)]
+  add-table:     L0 000004:[a#9,SET-z#9,DEL] seqnums:[9-9] points:[a#9,SET-z#9,DEL] blobrefs:[(000003: 25935); depth:1]
 ----
 L0.0:
-  000004:[a#9,SET-z#9,DEL] seqnums:[9-9] points:[a#9,SET-z#9,DEL] blobrefs:[(000003: 25935)]
+  000004:[a#9,SET-z#9,DEL] seqnums:[9-9] points:[a#9,SET-z#9,DEL] blobrefs:[(000003: 25935); depth:1]
 L1:
   000001:[a#2,SET-e#2,SET] seqnums:[0-0] points:[a#2,SET-e#2,SET]
 L2:
@@ -254,7 +254,7 @@ L2:
 # same version edit, accumulate will error.
 
 apply v8
-  add-table:     L0 000005:[a#10,SET-a#10,SET] seqnums:[10-10] points:[a#10,SET-a#10,SET] blobrefs:[(000003: 205)]
+  add-table:     L0 000005:[a#10,SET-a#10,SET] seqnums:[10-10] points:[a#10,SET-a#10,SET] blobrefs:[(000003: 205); depth:1]
 ----
 error during Accumulate: blob file 000003 referenced by L0.000005 not found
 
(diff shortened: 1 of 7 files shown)


RaduBerinde (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4348)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @sumeerbhola)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4348#-OK3NHO14mbJ-_2483Dt:-OK3NHO14mbJ-_2483Du:bg7igkv) (no related file):*
:lgtm: 
___
*[`internal/manifest/testdata/version_edit_decode` line 78 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4348#-OK3N2gY5k7QOCkMaDuX:-OK3N2gY5k7QOCkMaDuY:b-yjdc6t) ([raw file](https://github.com/cockroachdb/pebble/blob/db48c562240dca3c29bfc5d729bbb1817501888a/internal/manifest/testdata/version_edit_decode#L78)):*
> ```
> 67061d000b626172000e0000000000000b666f6f010d0000000000000000
> 45010129d8a301016b29d8a301f9a006006c21
>   add-table:     L6 000029:[bar#14,DEL-foo#13,SET] seqnums:[0-0] points:[bar#14,DEL-foo#13,SET] blobrefs:1*[(000041: 20952)]
> ```

The `1*` is very cryptic.. Could we just add a `blobrefdepth: %d` at the end? Or `; depth=%d` before closing the `]`
___
*[`internal/manifest/version.go` line 240 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4348#-OK3Lig2EVkHcMPio8CH:-OK3Lig2EVkHcMPio8CI:b9ggayd) ([raw file](https://github.com/cockroachdb/pebble/blob/db48c562240dca3c29bfc5d729bbb1817501888a/internal/manifest/version.go#L240)):*
> ```Go
> 	// referenced by this sstable.
> 	BlobReferences []BlobReference
> 	// BlobReferenceDepth is an upper bound on the number of blob files that a
> ```

Another way to describe it is that this is the largest overlapping set of spans, where each span corresponds to one blob file, is this correct?

By the way, will we maintain as metadata the key span for each blob file?
___
*[`internal/manifest/version.go` line 247 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4348#-OK3Mdzn4fVZwURQ3bbt:-OK3Mdzn4fVZwURQ3bbu:b-2yeswz) ([raw file](https://github.com/cockroachdb/pebble/blob/db48c562240dca3c29bfc5d729bbb1817501888a/internal/manifest/version.go#L247)):*
> ```Go
> 	// reference depth of 1. When a compaction reuses blob references, the max
> 	// blob reference depth of the files in each level is used, and then the
> 	// depth is summed, and assigned to the output. This is a primitive
> ```

[nit] I'd say "loose upper-bound" instead of "primitive approximation"

RaduBerinde:
*[`internal/manifest/version.go` line 247 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4348#-OK3Mdzn4fVZwURQ3bbt:-OK7rCv1AyzzaA_qIBxj:bvu0yqa) ([raw file](https://github.com/cockroachdb/pebble/blob/db48c562240dca3c29bfc5d729bbb1817501888a/internal/manifest/version.go#L247)):*

I think this reuses the phrase "primitive calculation" from https://docs.google.com/document/d/1bcDMfUnguIxNh9Po0kxiyhHbiBJA-UxPnCn-K\_yKsmU/edit?tab=t.0#bookmark=id.q6b0ib3prbsn :). I originally wrote it that way because I was lazy and didn't try to do the analysis of whether this was actually an upper-bound. If we change it to upper-bound, please add some analysis explaining why it is an upper-bound.
</blockquote></details>

The calculation is assuming the worst-case overlap across levels no?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4348)* status: 1 of 7 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/manifest/version.go` line 240 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4348#-OK3Lig2EVkHcMPio8CH:-OK80yBL6iYOLyYaLtkC:bc1lce1) ([raw file](https://github.com/cockroachdb/pebble/blob/db48c562240dca3c29bfc5d729bbb1817501888a/internal/manifest/version.go#L240)):*
Yeah, that's right (for the referencing spans)

> By the way, will we maintain as metadata the key span for each blob file?

I don't think so
___
*[`internal/manifest/version.go` line 247 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4348#-OK3Mdzn4fVZwURQ3bbt:-OK80wdCCAhROjtpTN-D:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/db48c562240dca3c29bfc5d729bbb1817501888a/internal/manifest/version.go#L247)):*

The calculation is assuming the worst-case overlap across levels no?
</blockquote></details>

Done
___
*[`internal/manifest/testdata/version_edit_decode` line 78 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4348#-OK3N2gY5k7QOCkMaDuX:-OK7oWjpDUJwfAYwWIgr:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/db48c562240dca3c29bfc5d729bbb1817501888a/internal/manifest/testdata/version_edit_decode#L78)):*

The `1*` is very cryptic.. Could we just add a `blobrefdepth: %d` at the end? Or `; depth=%d` before closing the `]`
</blockquote></details>

Done

jbowens:
TFTRs!

PR #4360, opened 2025-03-04
db: move excise to a separate file

jbowens:
Move DB.excise to its own file. Additionally, refactor the accumulation of EFOS protected ranges of EFOSes that overlap the excise span by moving it into a helper function in excise.go.


aadityasondhi (approved):

jbowens:
TFTR!

PR #4361, opened 2025-03-04
db: add DB.Excise

### LINKED ISSUE #4286: db: add Excise API
Add a new API that allows excising a span of the database without performing an accompanying ingestion. This is useful for recovering from an instance where externally linked sstables no longer exist or are corrupt.

Jira issue: PEBBLE-329

jbowens:
Add an Excise method to the DB allowing an excise without an accompanying ingestion. This will be used in CockroachDB to recover from corruption by excising out corrupted sstables.

For now, this implementation uses DB.ingest. I think a future refactor should be able to clean these code paths up and potentially refactor excise to be independent of the ingest codepath. Today, it's necessary to reuse the logic to perform excises as 'flushable ingests.'

Closes #4286.

### DIFF
--- a/checkpoint_test.go
+++ b/checkpoint_test.go
@@ -108,7 +108,7 @@ func testCheckpointImpl(t *testing.T, ddFile string, createOnShared bool) {
 
 			// Hacky but the command doesn't expect a db string. Get rid of it.
 			td.CmdArgs = td.CmdArgs[1:]
-			if err := runIngestAndExciseCmd(td, d, mem); err != nil {
+			if err := runIngestAndExciseCmd(td, d); err != nil {
 				return err.Error()
 			}
 			return ""
--- a/compaction_picker_test.go
+++ b/compaction_picker_test.go
@@ -1386,7 +1386,7 @@ func TestCompactionPickerPickFile(t *testing.T) {
 			return ""
 
 		case "ingest-and-excise":
-			if err := runIngestAndExciseCmd(td, d, d.opts.FS); err != nil {
+			if err := runIngestAndExciseCmd(td, d); err != nil {
 				return err.Error()
 			}
 			return ""
--- a/data_test.go
+++ b/data_test.go
@@ -1252,7 +1252,18 @@ func (d *DB) waitTableStats() {
 	}
 }
 
-func runIngestAndExciseCmd(td *datadriven.TestData, d *DB, fs vfs.FS) error {
+func runExciseCmd(td *datadriven.TestData, d *DB) error {
+	if len(td.CmdArgs) != 2 {
+		return errors.New("excise expects two arguments: <start-key> <end-key>")
+	}
+	exciseSpan := KeyRange{
+		Start: []byte(td.CmdArgs[0].String()),
+		End:   []byte(td.CmdArgs[1].String()),
+	}
+	return d.Excise(context.Background(), exciseSpan)
+}
+
+func runIngestAndExciseCmd(td *datadriven.TestData, d *DB) error {
 	var exciseSpan KeyRange
 	paths := make([]string, 0, len(td.CmdArgs))
 	for i, arg := range td.CmdArgs {
--- a/db.go
+++ b/db.go
@@ -1619,11 +1619,21 @@ func (d *DB) NewIterWithContext(ctx context.Context, o *IterOptions) (*Iterator,
 // will not prevent memtables from being released or sstables from being
 // deleted. Instead, a snapshot prevents deletion of sequence numbers
 // referenced by the snapshot.
+//
+// There exists one violation of a Snapshot's point-in-time guarantee: An excise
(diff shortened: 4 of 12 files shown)


RaduBerinde (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4361)* status: 0 of 11 files reviewed, 4 unresolved discussions

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4361#-OKZ3cJu63gZ8EId5iQQ:-OKZ3cJu63gZ8EId5iQR:b-4qh7xm) (no related file):*
Nice, thanks for doing this! :lgtm: 
___
*[`excise.go` line 19 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4361#-OKZ3wzm0WQfERSsxyGv:-OKZ3wzm0WQfERSsxyGw:bjok3a1) ([raw file](https://github.com/cockroachdb/pebble/blob/d8cf203f1547fc07afc57dad7a564607fe32125c/excise.go#L19)):*
> ```Go
> // Excise atomically deletes all data overlapping with the provided span. All
> // data overlapping with the span is removed, including from open snapshots.
> // Only currently-open iterators will still observe the removed data. Excise may
> ```

The currently-open iterators observe the removed data because they use EFOS? And "including from open snapshots" refers to non-EFOS?
___
*[`excise.go` line 30 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4361#-OKZ47wL5wAuMIb_gU3q:-OKZ47wL5wAuMIb_gU3r:b-ldbfs1) ([raw file](https://github.com/cockroachdb/pebble/blob/d8cf203f1547fc07afc57dad7a564607fe32125c/excise.go#L30)):*
> ```Go
> 	if invariants.Enabled {
> 		// Excise is only supported on prefix keys.
> 		if d.opts.Comparer.Split(span.Start) != len(span.Start) {
> ```

Since this is an external API, we should just return errors.
___
*[`testdata/excise` line 711 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4361#-OKZ7k2k70M9zEnUXkBe:-OKZ7k2k70M9zEnUXkBf:bqdqbis) ([raw file](https://github.com/cockroachdb/pebble/blob/d8cf203f1547fc07afc57dad7a564607fe32125c/testdata/excise#L711)):*
> ```
> .
> .
> ```

[nit] maybe add a test that cuts inside the range, and one that cuts completely outside all existing keys

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4361)* status: 0 of 12 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4361#-OKZ3cJu63gZ8EId5iQQ:-OKb9hNw78XryABr3ATg:b-814g8k) (no related file):*

Nice, thanks for doing this! :lgtm:
</blockquote></details>

TFTR!
___
*[`excise.go` line 19 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4361#-OKZ3wzm0WQfERSsxyGv:-OKb7aut5pkIWSa2CrYU:b-8519hc) ([raw file](https://github.com/cockroachdb/pebble/blob/d8cf203f1547fc07afc57dad7a564607fe32125c/excise.go#L19)):*

The currently-open iterators observe the removed data because they use EFOS? And "including from open snapshots" refers to non-EFOS?
</blockquote></details>

It's because a currently-open iterator pins its readState (and so Version, memtables and sstables). So even though the excise has removed the data from the current _Version_, the iterator's _Version_ still contains the hold table metadatas. I clarified this comment some and added a bit more commentary to NewSnapshot too. I think we should try to eliminate the now weird semantics of NewSnapshot, since it's too easy to accidentally rely on a point-in-time guarantee that you accidentally violate through Excise.
___
*[`excise.go` line 30 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4361#-OKZ47wL5wAuMIb_gU3q:-OKb5U8c7kUae-LggIHh:bija1ts) ([raw file](https://github.com/cockroachdb/pebble/blob/d8cf203f1547fc07afc57dad7a564607fe32125c/excise.go#L30)):*

Since this is an external API, we should just return errors.
</blockquote></details>

I'll make these production build checks too then, and give the same treatment to the checks in `IngestAndExcise`
___
*[`testdata/excise` line 711 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4361#-OKZ7k2k70M9zEnUXkBe:-OKb9fXw5fPnXTMVbPi5:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/d8cf203f1547fc07afc57dad7a564607fe32125c/testdata/excise#L711)):*

[nit] maybe add a test that cuts inside the range, and one that cuts completely outside all existing keys
</blockquote></details>

Done

PR #4362, opened 2025-03-05
db: refactor race-build file cache reference tracking

jbowens:
Previously the file cache used a pointer to the sstable iterator to maintain the stack traces of the origination of open references during race builds. This was a bit ambiguous. With the introduction of blob files, references will also be maintained by non-sstable iterators (eg, a cached blob file reader within an iterator's blob value fetcher).

This commit refactors this reference tracking to propagate a uint64 reference ID along with the closeHook, requiring the user to pass the reference ID back in their invocation of the closeHook. In production builds, this reference ID is always zero and is not stored. This hopefully will make usages unambiguous.

### DIFF
--- a/file_cache.go
+++ b/file_cache.go
@@ -107,9 +107,12 @@ type fileCacheHandle struct {
 	// This struct is only populated in race builds.
 	raceMu struct {
 		sync.Mutex
-		// iters maps sstable.Iterator objects to the stack trace recorded at
-		// creation time.
-		iters map[any][]byte
+		// nextRefID is the next ID to allocate for a new reference.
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):
:lgtm: 

> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4362#-:-OKcFC98AkamFUQwfxir:bg7igkv)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @jbowens)

jbowens:
TFTR

Issue #3938, opened 2024-09-16
reduce severity for 'pebble: compaction cancelled by a concurrent operation, will retry compaction'


tbg:
Have been seeing these regularly recently:

> teamcity-16884024-1726292891-120-n7cpu4-0005> E240914 12:20:44.318336 17340 3@pebble/event.go:764 ⋮ [T1,n5,s5,pebble] 255  background error: pebble: compaction cancelled by a concurrent operation, will retry compaction

Apparently they are expected[^1], so they should likely be logged as informational, if at all.

[^1]: https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1726482169839989

Jira issue: PEBBLE-257

RaduBerinde:
We should instead expose some metrics about how much work was wasted (eg bytes written in canceled compactions).

jbowens:
@xxmplus - this seems like a good starter.

> We should instead expose some metrics about how much work was wasted (eg bytes written in canceled compactions).

We've since added metrics tracking wasted work in #4132, but as far as I can tell we haven't done anything about the severity of cancelled compactions error messages. 

In Cockroach, we construct an event listener that includes the pebble logging event listener here:
https://github.com/cockroachdb/cockroach/blob/e452d477319e7811886c0167b50e9d62bf2afaad/pkg/storage/pebble.go#L1000-L1006

The logging event listener logs all background errors using `log.Error` here:
https://github.com/cockroachdb/pebble/blob/5b12eaf3fcde59db6ede98c24c2d5a95b248d340/event.go#L864-L866

I think we probably want to continue to log these events, since we log ordinary compaction start and ends, and sometimes need to consult that level detail (or aggregate it using the pebble compaction log tool). I think it would be reasonable to simply adapt the logging event listener's BackgroundError func to check is errors.Is(err, ErrCancelledCompaction) and log it at the info severity if it is.

xxmplus:
Code change is made. Resolving.

PR #4363, opened 2025-03-06
db: reduce log level to info for cancelled compaction

xxmplus:
The cancelled compaction error is expected, so reduce the log level to info as suggested in #3938.

### DIFF
--- a/event.go
+++ b/event.go
@@ -774,6 +774,15 @@ type EventListener struct {
 	PossibleAPIMisuse func(PossibleAPIMisuseInfo)
 }
 
+func backgroundError(logger Logger, err error) {
+	if errors.Is(err, ErrCancelledCompaction) {
+		// ErrCancelledCompaction is not an unexpected error, hence severity INFO.
+		logger.Infof("background error: %s", err)
+	} else {
+		logger.Errorf("background error: %s", err)
+	}
+}
+
 // EnsureDefaults ensures that background error events are logged to the
 // specified logger if a handler for those events hasn't been otherwise
 // specified. Ensure all handlers are non-nil so that we don't have to check
@@ -782,7 +791,7 @@ func (l *EventListener) EnsureDefaults(logger Logger) {
 	if l.BackgroundError == nil {
 		if logger != nil {
 			l.BackgroundError = func(err error) {
-				logger.Errorf("background error: %s", err)
+				backgroundError(logger, err)
 			}
 		} else {
 			l.BackgroundError = func(error) {}
@@ -862,7 +871,7 @@ func MakeLoggingEventListener(logger Logger) EventListener {
 
 	return EventListener{
 		BackgroundError: func(err error) {
-			logger.Errorf("background error: %s", err)
+			backgroundError(logger, err)
 		},
 		CompactionBegin: func(info CompactionInfo) {
 			logger.Infof("%s", info)
--- a/event_listener_test.go
+++ b/event_listener_test.go
@@ -367,6 +367,67 @@ func TestMakeLoggingEventListenerSetsAllCallbacks(t *testing.T) {
 	testAllCallbacksSetInEventListener(t, e)
 }
 
+type mockLogger struct {
+	infoFunc  func(format string, args ...interface{})
+	errorFunc func(format string, args ...interface{})
+	fatalFunc func(format string, args ...interface{})
+}
+
+func (l *mockLogger) Infof(format string, args ...interface{}) {
+	if l.infoFunc != nil {
+		l.infoFunc(format, args...)
+	}
+}
+
+func (l *mockLogger) Errorf(format string, args ...interface{}) {
+	if l.errorFunc != nil {
+		l.errorFunc(format, args...)
+	}
+}
+
+func (l *mockLogger) Fatalf(format string, args ...interface{}) {
+	if l.fatalFunc != nil {
+		l.fatalFunc(format, args...)
+	}
+}
+
+func newCountingMockLogger(t *testing.T) (*mockLogger, *int, *int) {
+	var infoCount, errorCount int
+	return &mockLogger{
+		infoFunc: func(format string, args ...interface{}) {
+			infoCount++
+		},
+		errorFunc: func(format string, args ...interface{}) {
+			errorCount++
+		},
+		fatalFunc: func(format string, args ...interface{}) {
+			t.Fatal("Unexpected call to Fatalf")
+		},
+	}, &infoCount, &errorCount
+}
+
+func TestMakeLoggingEventListenerBackgroundErrorCancelledCompaction(t *testing.T) {
+	mockLogger, infoCount, errorCount := newCountingMockLogger(t)
+	e := MakeLoggingEventListener(mockLogger)
+
+	e.BackgroundError(ErrCancelledCompaction)
+	require.Equal(t, 1, *infoCount)
+	require.Equal(t, 0, *errorCount)
+
+	testAllCallbacksSetInEventListener(t, e)
+}
+
+func TestMakeLoggingEventListenerBackgroundErrorOtherError(t *testing.T) {
+	mockLogger, infoCount, errorCount := newCountingMockLogger(t)
+	e := MakeLoggingEventListener(mockLogger)
+
+	e.BackgroundError(errors.New("an example error"))
+	require.Equal(t, 0, *infoCount)
+	require.Equal(t, 1, *errorCount)
+
+	testAllCallbacksSetInEventListener(t, e)
+}
+
 func TestTeeEventListenerSetsAllCallbacks(t *testing.T) {
 	e := TeeEventListener(EventListener{}, EventListener{})
 	testAllCallbacksSetInEventListener(t, e)


xxmplus (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4363)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @sumeerbhola)

___
*[`event.go` line 786 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4363#-OKd5np67tfzPaq3dDhj:-OKe18jXDoR7vBETkoCh:b123usr) ([raw file](https://github.com/cockroachdb/pebble/blob/27c1230e2be669ef3e2c36c9e4bcf39af47cca78/event.go#L786)):*

I suspect we still want this to say "background error". That is, it is sufficient that this is now `logger.Infof`.
</blockquote></details>

ack, updated.
___
*[`event.go` line 789 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4363#-OKd6r0D1lzO475FwKQW:-OKe1HEGAoIc9B7ABDHA:b-kp25i6) ([raw file](https://github.com/cockroachdb/pebble/blob/27c1230e2be669ef3e2c36c9e4bcf39af47cca78/event.go#L789)):*

I'm not sure this `EnsureDefaults` path is sufficient for this to have the desired effect. We call `MakeLoggingEventListener` with a non-nil `Logger`, and then call `TeeEventListener`, so since the first call has already set `EventListener.BackgroundError` the latter call will do nothing when it calls `EnsureDefaults`. 

The unit test we add here should just confirm that this is working as expected with `MakeLoggingEventListener` with a `testLogger`. That is, something like

```
type testLogger struct {
  errCount, infoCount int
}
func (l *testLogger) Infof(format string, args ...interface{}) {
  l.infoCount++
}
...
```
</blockquote></details>

good catch. updated.

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4363)* status: all files reviewed, 7 unresolved discussions (waiting on @xxmplus)

___
*[`event_listener_test.go` line 407 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4363#-OKfg3qt0PcEUm6tuRZE:-OKfg3qt0PcEUm6tuRZF:b-5nlx9k) ([raw file](https://github.com/cockroachdb/pebble/blob/2407fe6d92ae32c3841cf09121f02d7775d11fff/event_listener_test.go#L407)):*
> ```Go
> 		},
> 	}, &infoCount, &errorCount
> }
> ```

No need to change what you have here, especially since this is test code, but in general we prefer to delay generalizing (whether it be production or test code), until most aspects of that generalization are actually in use. That is, we are willing to rewrite code, versus over-generalizing up front, since the latter has the cost of more to read and comprehend. Of course, this rule doesn't apply to modularizing, which reduces cognitive burden, but we will typically keep the interfaces across modules narrow to what is needed.
In this case, `mockLogger` is adding a level of generalization via the indirection of function pointers, which aren't needed for this one use case. But fine to leave this as-is.
___
*[`event_listener_test.go` line 414 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4363#-OKfh_Mb3vgmS_bgLQVV:-OKfh_Mb3vgmS_bgLQVW:bgnxggc) ([raw file](https://github.com/cockroachdb/pebble/blob/2407fe6d92ae32c3841cf09121f02d7775d11fff/event_listener_test.go#L414)):*
> ```Go
> 
> 	e.BackgroundError(ErrCancelledCompaction)
> 	require.Equal(t, 1, *infoCount, "Infof should be called once for ErrCancelledCompaction")
> ```

nit: we don't bother with a custom error message in these test assertions, except when there is some state that needs to be printed when the assertion fails *and* the failure is hard to reproduce (e.g. occurs in some CI runs, but not when we try to reproduce). The assertion failure includes the line number it failed on which is sufficient for investigation.
___
*[`event.go` line 869 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4363#-OKfitSy6J9jjxf3zoeG:-OKfitSy6J9jjxf3zoeH:b5wgqxe) ([raw file](https://github.com/cockroachdb/pebble/blob/2407fe6d92ae32c3841cf09121f02d7775d11fff/event.go#L869)):*
> ```Go
> 	return EventListener{
> 		BackgroundError: func(err error) {
> 			if errors.Is(err, ErrCancelledCompaction) {
> ```

Consider adding a comment like
// ErrCancelledCompaction is not an unexpected error, hence has severity INFO.
___
*[`event.go` line 874 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4363#-OKffSNwB9q0D9P1nKZ2:-OKffSNwB9q0D9P1nKZ3:bpdc0x0) ([raw file](https://github.com/cockroachdb/pebble/blob/2407fe6d92ae32c3841cf09121f02d7775d11fff/event.go#L874)):*
> ```Go
> 				logger.Errorf("background error: %s", err)
> 			}
> 		},
> ```

instead of duplicating the code how about separating into a func that can be used in both places. Something like
```
func backgroundError(logger Logger, err error) {
   ...
}
```

xxmplus (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4363)* status: 0 of 2 files reviewed, 7 unresolved discussions (waiting on @sumeerbhola)

___
*[`event.go` line 869 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4363#-OKfitSy6J9jjxf3zoeG:-OKgMJQADAQ0WheJpHb4:b-k5axh6) ([raw file](https://github.com/cockroachdb/pebble/blob/2407fe6d92ae32c3841cf09121f02d7775d11fff/event.go#L869)):*

Consider adding a comment like
// ErrCancelledCompaction is not an unexpected error, hence has severity INFO.
</blockquote></details>

ack, added.
___
*[`event.go` line 874 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4363#-OKffSNwB9q0D9P1nKZ2:-OKgMH5t5iutbKHVrNep:b123usr) ([raw file](https://github.com/cockroachdb/pebble/blob/2407fe6d92ae32c3841cf09121f02d7775d11fff/event.go#L874)):*

instead of duplicating the code how about separating into a func that can be used in both places. Something like
```
func backgroundError(logger Logger, err error) {
   ...
}
```
</blockquote></details>

ack, updated.
___
*[`event_listener_test.go` line 407 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4363#-OKfg3qt0PcEUm6tuRZE:-OKgMMXF7AgyfYWdDGOy:b-tc6d92) ([raw file](https://github.com/cockroachdb/pebble/blob/2407fe6d92ae32c3841cf09121f02d7775d11fff/event_listener_test.go#L407)):*

No need to change what you have here, especially since this is test code, but in general we prefer to delay generalizing (whether it be production or test code), until most aspects of that generalization are actually in use. That is, we are willing to rewrite code, versus over-generalizing up front, since the latter has the cost of more to read and comprehend. Of course, this rule doesn't apply to modularizing, which reduces cognitive burden, but we will typically keep the interfaces across modules narrow to what is needed.
In this case, `mockLogger` is adding a level of generalization via the indirection of function pointers, which aren't needed for this one use case. But fine to leave this as-is.
</blockquote></details>

I see. will keep in mind in the future.
___
*[`event_listener_test.go` line 414 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4363#-OKfh_Mb3vgmS_bgLQVV:-OKgMTw83i33sVDdMvL8:b-tanj2y) ([raw file](https://github.com/cockroachdb/pebble/blob/2407fe6d92ae32c3841cf09121f02d7775d11fff/event_listener_test.go#L414)):*

nit: we don't bother with a custom error message in these test assertions, except when there is some state that needs to be printed when the assertion fails *and* the failure is hard to reproduce (e.g. occurs in some CI runs, but not when we try to reproduce). The assertion failure includes the line number it failed on which is sufficient for investigation.
</blockquote></details>

ack, removed.

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4363)* status: all files reviewed, 8 unresolved discussions (waiting on @xxmplus)

___
*[`event.go` line 779 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4363#-OKh4xBw0jo9ym7DFviR:-OKh4xBw0jo9ym7DFviS:bo7noym) ([raw file](https://github.com/cockroachdb/pebble/blob/6ad040596493dea57b2c3fecaa6e85df909a2086/event.go#L779)):*
> ```Go
> // ErrCancelledCompaction is not an unexpected error, hence has severity INFO.
> func backgroundError(logger Logger, err error) {
> 	if errors.Is(err, ErrCancelledCompaction) {
> ```

The comment about `ErrCancelledCompaction` belongs inside this if-block.

xxmplus:
TFTR!

RaduBerinde:
Note that amending the commit is preferred to making more commits during the review. Reviewable already allows diffing between revisions if necessary.

PR #4366, opened 2025-03-08
internal/compact: preserve blob references

jbowens:
Adapt the compaction iterator to preserve blob file references by cloning the
LazyValues. Compactions that do not rewrite blob files will avoid ever
retrieving the corresponding values.

Additionally, fix LazyValue.Clone to copy the BlobFileNum of the LazyFetcher.

### DIFF
--- a/internal/base/lazy_value.go
+++ b/internal/base/lazy_value.go
@@ -216,7 +216,7 @@ func (lv *LazyValue) fetchValue(
 	if !f.fetched {
 		f.fetched = true
 		f.value, f.callerOwned, f.err = f.Fetcher.Fetch(ctx,
-			lv.ValueOrHandle, lv.Fetcher.BlobFileNum, lv.Fetcher.Attribute.ValueLen, buf)
+			lv.ValueOrHandle, f.BlobFileNum, f.Attribute.ValueLen, buf)
 	}
 	return f.value, f.callerOwned, f.err
 }
@@ -262,8 +262,9 @@ func (lv *LazyValue) Clone(buf []byte, fetcher *LazyFetcher) (LazyValue, []byte)
 	var lvCopy LazyValue
 	if lv.Fetcher != nil {
 		*fetcher = LazyFetcher{
-			Fetcher:   lv.Fetcher.Fetcher,
-			Attribute: lv.Fetcher.Attribute,
+			Fetcher:     lv.Fetcher.Fetcher,
+			Attribute:   lv.Fetcher.Attribute,
+			BlobFileNum: lv.Fetcher.BlobFileNum,
 			// Not copying anything that has been extracted.
 		}
 		lvCopy.Fetcher = fetcher
--- a/internal/base/lazy_value_test.go
+++ b/internal/base/lazy_value_test.go
@@ -58,9 +58,11 @@ func TestLazyValue(t *testing.T) {
 						numCalls++
 						require.Equal(t, []byte("foo-handle"), handle)
 						require.Equal(t, uint32(3), valLen)
+						require.Equal(t, DiskFileNum(90), blobFileNum)
 						return fooBytes1, callerOwned, nil
 					}),
-				Attribute: AttributeAndLen{ValueLen: 3, ShortAttribute: 7},
+				Attribute:   AttributeAndLen{ValueLen: 3, ShortAttribute: 7},
+				BlobFileNum: 90,
 			},
 		}
 		require.Equal(t, []byte("foo"), getValue(fooLV3, callerOwned))
--- a/internal/base/value.go
+++ b/internal/base/value.go
@@ -27,6 +27,13 @@ func MakeInPlaceValue(val []byte) InternalValue {
 	return InternalValue{lazyValue: LazyValue{ValueOrHandle: val}}
 }
 
+// IsBlobValueHandle returns true iff the value is a blob value handle, pointing
+// to a value stored externally in a blob file.
+func (v *InternalValue) IsBlobValueHandle() bool {
+	f := v.lazyValue.Fetcher
+	return f != nil && f.BlobFileNum > 0
+}
+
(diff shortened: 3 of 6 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4366)* status: 3 of 6 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/compact/iterator.go` line 286 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4366#-OL-dITjBz5QFfSE8zWB:-OL-p-cZ8bBEr8EnrJZT:b82vu3e) ([raw file](https://github.com/cockroachdb/pebble/blob/1122a0b05e827aca0664c2cdf40df227c2bce35e/internal/compact/iterator.go#L286)):*

I'm getting a bit confused by both seeing this in the iterator, and the various optionality in the comments ("may set this to true" etc.).
Shouldn't the iterator just expose LazyValue with the blob reference regardless of whether the compaction wants to preserve the reference or not, and let the caller make the decision? I think we were doing something like that in the prototype, and the only place where the `compactionIter` would retrieve the underlying value was in `mergeNext`.
</blockquote></details>

The logic in `saveValue` is the crux. When we're saving a value that is a blob reference, we can either Clone or retrieve it. There's no point in cloning if we're ultimately going to retrieve it in the caller.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4366)* status: 2 of 14 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4366#-OL-mI4yD72AHMFdchqD:-OLAAa5Z7otzd_QhkPOX:bbfubex) (no related file):*

Just a high-level question about the approach -- I'll read carefully after that is resolved.
</blockquote></details>

Alright, I removed the logic that avoided unnecessary clones. This is ready for a review. Apologies there were some intermediate diffs that are now extracted out into #4376. Probably best to just review the entirety of the diff against base.

RaduBerinde (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4366)* status: 2 of 14 files reviewed, 8 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4366#-OL-mI4yD72AHMFdchqD:-OLAW1CIDtNTXcJ3Y9Q1:bg7igkv) (no related file):*

Alright, I removed the logic that avoided unnecessary clones. This is ready for a review. Apologies there were some intermediate diffs that are now extracted out into #4376. Probably best to just review the entirety of the diff against base.
</blockquote></details>

:lgtm: 
___
*[`-- commits` line 15 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4366#-OLAUcUlA2SKBlpO2i8Y:-OLAUcUlA2SKBlpO2i8Z:bg35zt7):*
[nit] compaction
___
*[`internal/compact/testdata/iter` line 1132 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4366#-OLAVI588uJjiiDXLJX6:-OLAVI588uJjiiDXLJX7:bscx5kp) ([raw file](https://github.com/cockroachdb/pebble/blob/ebb2512be0314f6e05a589fae82423a66d5a1bae/internal/compact/testdata/iter#L1132)):*
> ```
> 
> define
> a.SET.9:blobref(000294, blk2, 10, 20)
> ```

[nit] In the future, we may want to implement a more ergonomic test facility for blob refs, e.g. something where you define it with a friendly name e.g. `blobref(a)`. The test can generate the filenum and stuff internally and keep it around so it can convert back when printing.
___
*[`internal/compact/iterator_test.go` line 280 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4366#-OLAVb_5DWzh8g-ZDaW7:-OLAVb_5DWzh8g-ZDaW8:b93nwre) ([raw file](https://github.com/cockroachdb/pebble/blob/ebb2512be0314f6e05a589fae82423a66d5a1bae/internal/compact/iterator_test.go#L280)):*
> ```Go
> }
> 
> type mockBlobValueFetcher struct{}
> ```

[nit] Could use a comment, like
// mockBlobValueFetcher is a dummy ValueFetcher implementation which produces string values 
// reference the blobref.
___
*[`internal/compact/iterator_test.go` line 287 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4366#-OLAVu138piNhV1VxK2h:-OLAVu138piNhV1VxK2i:bg1lvvq) ([raw file](https://github.com/cockroachdb/pebble/blob/ebb2512be0314f6e05a589fae82423a66d5a1bae/internal/compact/iterator_test.go#L287)):*
> ```Go
> 	ctx context.Context, handle []byte, fileNum base.DiskFileNum, valLen uint32, buf []byte,
> ) (val []byte, callerOwned bool, err error) {
> 	s := fmt.Sprintf("<a fetched value from blobref(%s, encodedHandle=%x, valLen=%d)>",
> ```

[nit] "a " seems extraneous

sumeerbhola (approved):
> Reviewed 1 of 6 files at r1, 3 of 3 files at r2, 2 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4366)* status: 8 of 14 files reviewed, 8 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4366#-OL-mI4yD72AHMFdchqD:-OLB2Pok8UbaWQlxCKch:bg7igkv) (no related file):*

:lgtm:
</blockquote></details>

:lgtm:

PR #4369, opened 2025-03-10
avoid timeouts and failures with asan builds

RaduBerinde:
The asan build has been mostly failing due to timeouts. These slower
tests automatically adjust for race builds, but not for asan builds.

This commit adds a `buildtags.Instrumented` constant which indicates
if this is a race or asan or msan build. We switch various
`RaceEnabled` test checks to the new flag so we include these builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,10 +36,12 @@ testrace: testflags += -race -timeout 20m
 testrace: test
 
 testasan: testflags += -asan -timeout 20m
+testasan: TAGS += slowbuild
 testasan: test
 
 testmsan: export CC=clang
 testmsan: testflags += -msan -timeout 20m
+testmsan: TAGS += slowbuild
 testmsan: test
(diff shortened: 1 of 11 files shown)


RaduBerinde:
TFTR!

I think the s390x test run will need this too as it will be pretty slow through qemu. Considering changing `Instrumented` to `SlowBuild` and just using a single tag `slowbuild` for all these, does that sound ok?

RaduBerinde:
Made that change, PTAL. Thanks!

jbowens (approved):
> Reviewed 10 of 10 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4369)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

PR #4371, opened 2025-03-10
colblk: support big endian

RaduBerinde:
With these changes the `colblk` and `cockroachkvs` tests pass on s390x (under qemu). I will work separately on adding a nightly run for s390x.

#### colblk: replace UnsafeRawSlice with unsafeUint64Decoder

Bitmaps are the only remaining use of `UnsafeRawSlice`, so we simplify
to a specialized type which will be easy to make work with big endian.

#### colblk: support big endian


RaduBerinde:
cockroachkv benchmarks here: https://gist.github.com/RaduBerinde/71fdd51446b2bf07311516797d6ed412

jbowens (approved):
:lgtm_strong: 

thanks for doing this!

> Reviewed 4 of 4 files at r1, 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4371#-:-OLA-l8BA2bL8QzCAfsg:blf0wnz)* status: all files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/colblk/uints_decode.go` line 32 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4371#-OL9xClk3U27k-sFgsBr:-OL9xClk3U27k-sFgsBs:bpnr34c) ([raw file](https://github.com/cockroachdb/pebble/blob/92bd33f1cb1e9a0a655f528974eb9a914b98fe38/sstable/colblk/uints_decode.go#L32)):*
> ```Go
> 
> //gcassert:inline
> func (s unsafeUint64Decoder) At(idx int) uint64 {
> ```

nit: we could use `invariants.Value[int]` to retain `n` and perform a bounds check in invariants builds
___
*[`sstable/colblk/endian.go` line 16 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4371#-OL9xoSN8VLl2nGy2If2:-OL9xoSN8VLl2nGy2If3:b-s6lnsk) ([raw file](https://github.com/cockroachdb/pebble/blob/2b74658c6691c6e8636665ff3e92809319271338/sstable/colblk/endian.go#L16)):*
> ```Go
> 	if len(s) >= 4 {
> 		// We convert the slice (up to the tail) to a slice of [4]uint16. This helps
> 		// the compiler elide bound checks.
> ```

clever

RaduBerinde:
*[`sstable/colblk/uints_decode.go` line 32 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4371#-OL9xClk3U27k-sFgsBr:-OLA0Uyo257s5-ve4Hcf:b-wa0tb4) ([raw file](https://github.com/cockroachdb/pebble/blob/92bd33f1cb1e9a0a655f528974eb9a914b98fe38/sstable/colblk/uints_decode.go#L32)):*

nit: we could use `invariants.Value[int]` to retain `n` and perform a bounds check in invariants builds
</blockquote></details>

The problem is that it will no longer fit in the cached block metadata in invariant builds.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4371)* status: all files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/colblk/uints_decode.go` line 32 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4371#-OL9xClk3U27k-sFgsBr:-OLAB4Jo6bjeMq-ZUNXO:b-4th3o9) ([raw file](https://github.com/cockroachdb/pebble/blob/92bd33f1cb1e9a0a655f528974eb9a914b98fe38/sstable/colblk/uints_decode.go#L32)):*

The problem is that it will no longer fit in the cached block metadata in invariant builds.
</blockquote></details>

Ah that makes sense. Ah well

RaduBerinde:
TFTR!

PR #4376, opened 2025-03-12
db: implement blob.ReaderProvider on fileCacheHandle

jbowens:
Adapt the file cache to implement the blob.ReaderProvider interface, supporting
the caching of blob files alongside sstable files. The two file types will
share a cache since open sstables and blob files both contribute towards the
same open file descriptor limits.

### DIFF
--- a/event.go
+++ b/event.go
@@ -1150,17 +1150,32 @@ func (r *lowDiskSpaceReporter) findThreshold(
 	return threshold, ok
 }
 
-func (d *DB) reportSSTableCorruption(meta *manifest.TableMetadata, err error) {
+func (d *DB) reportCorruption(meta any, err error) {
+	switch meta := meta.(type) {
+	case *manifest.TableMetadata:
+		d.reportFileCorruption(base.FileTypeTable, meta.FileBacking.DiskFileNum, meta.UserKeyBounds(), err)
+	case *manifest.BlobFileMetadata:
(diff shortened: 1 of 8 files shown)


RaduBerinde (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4376)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @xxmplus)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4376#-OLAX6ky5Tpq4xwGO3te:-OLAX6kz4RIjUIKwEaFE:b4jtrn4) (no related file):*
r2 and r2 LGTM
___
*[`sstable/blob/blob.go` line 60 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4376#-OLATtHU00hGNLfxdM2Q:-OLATtHU00hGNLfxdM2R:b-xzgxv1) ([raw file](https://github.com/cockroachdb/pebble/blob/34cc6f32fb3086902bca02a45d10ad77c88c9784/sstable/blob/blob.go#L60)):*
> ```Go
> 		o.Compression = block.SnappyCompression
> 	}
> 	if o.ChecksumType == block.ChecksumTypeNone {
> ```

This means we'll never be able to write without checksums, but that's probably a good idea..

RaduBerinde:

r2 and r2 LGTM
</blockquote></details>

*and r3

PR #4382, opened 2025-03-14
sstable: add AddWithBlobHandle

jbowens:
Add a new AddWithBlobHandle method to the sstable.RawWriter interface allowing
the writing of key-value pairs with values that encode the location of the
value out-of-band in an external blob file.

Informs #112.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -587,7 +587,7 @@ func runBuildRemoteCmd(td *datadriven.TestData, d *DB, storage remote.Storage) e
 	for kv := iter.First(); kv != nil; kv = iter.Next() {
 		tmp := kv.K
 		tmp.SetSeqNum(0)
-		if err := w.Raw().AddWithForceObsolete(tmp, kv.InPlaceValue(), false); err != nil {
+		if err := w.Raw().Add(tmp, kv.InPlaceValue(), false); err != nil {
 			return err
 		}
 	}
@@ -677,7 +677,7 @@ func runBuildCmd(td *datadriven.TestData, d *DB, fs vfs.FS) error {
 	for kv := iter.First(); kv != nil; kv = iter.Next() {
 		tmp := kv.K
 		tmp.SetSeqNum(0)
-		if err := w.Raw().AddWithForceObsolete(tmp, kv.InPlaceValue(), false); err != nil {
+		if err := w.Raw().Add(tmp, kv.InPlaceValue(), false); err != nil {
 			return err
 		}
 	}
--- a/file_cache_test.go
+++ b/file_cache_test.go
@@ -214,7 +214,7 @@ func (t *fileCacheTest) newTestHandle() (*fileCacheHandle, *fileCacheTestFS) {
 		}
 		tw := sstable.NewWriter(w, sstable.WriterOptions{TableFormat: sstable.TableFormatPebblev2})
 		ik := base.ParseInternalKey(fmt.Sprintf("k.SET.%d", i))
-		if err := tw.Raw().AddWithForceObsolete(ik, xxx[:i], false); err != nil {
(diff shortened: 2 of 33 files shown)


RaduBerinde (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4382)* status: 0 of 35 files reviewed, 2 unresolved discussions (waiting on @xxmplus)

___
*[`sstable/blob/handle.go` line 19 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4382#-OLLE5L75hX8skvUY2fe:-OLLE5L75hX8skvUY2ff:b-j91r0q) ([raw file](https://github.com/cockroachdb/pebble/blob/3003bbbd91b35f79f69bbb72c40afd8ec4436f5c/sstable/blob/handle.go#L19)):*
> ```Go
> //
> // Handle fields are varint encoded, so maximum 5 bytes each.
> const MaxInlineHandleLength = 5*4 + 1
> ```

[nit] We should use `encoding.MaxVarintLen32 + valblk.HandleMaxLen` which is self-documenting.

Though I don't understand where the +1 comes from. valblk says it's the "valuePrefix" but the encode/decode functions don't have anything about that
___
*[`sstable/blob/handle.go` line 103 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4382#-OLLEiW8-pRaQ-Abwg7t:-OLLEiW8-pRaQ-Abwg7u:b-ny5bq2) ([raw file](https://github.com/cockroachdb/pebble/blob/3003bbbd91b35f79f69bbb72c40afd8ec4436f5c/sstable/blob/handle.go#L103)):*
> ```Go
> 		refIdx = uint32(a)
> 		src = src[1:]
> 	} else if a, b := a&0x7f, *((*uint8)(unsafe.Pointer(uintptr(ptr) + 1))); b < 128 {
> ```

[nit] can replace `unsafe.Pointer(uintptr(ptr) + 1)` with `unsafe.Add(ptr, 1)`

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4382#-:-OLLfHBC-QTaM2oiHXEa:b-814g8k)* status: 0 of 35 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @xxmplus)

___
*[`sstable/blob/handle.go` line 19 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4382#-OLLE5L75hX8skvUY2fe:-OLLe-gn5B41c1WsPpq5:bodi7yo) ([raw file](https://github.com/cockroachdb/pebble/blob/3003bbbd91b35f79f69bbb72c40afd8ec4436f5c/sstable/blob/handle.go#L19)):*
Updated to
```
const MaxInlineHandleLength = 4 * binary.MaxVarintLen32
```
(I'm expecting we will diverge from the valblk.Handle).

> Though I don't understand where the +1 comes from. valblk says it's the "valuePrefix" but the encode/decode functions don't have anything about that

Yeah, I noticed the same and forgot to remove this `+ 1`. I think the callers that need to decode a value directly from a block (where it might have a valuePrefix byte ahead of it) should account for the `+1`. I'll follow up with that change to valblk.
___
*[`sstable/blob/handle.go` line 103 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4382#-OLLEiW8-pRaQ-Abwg7t:-OLLeb2b1mw_SBDQGP7m:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/3003bbbd91b35f79f69bbb72c40afd8ec4436f5c/sstable/blob/handle.go#L103)):*

[nit] can replace `unsafe.Pointer(uintptr(ptr) + 1)` with `unsafe.Add(ptr, 1)`
</blockquote></details>

Done

PR #4385, opened 2025-03-14
sstable: clean up and document maximum handle lengths

jbowens:
This commit removes the +1 from valblk.HandleMaxLen. The code treats the handle as the data after the valuePrefix byte, noninclusive of the byte. The assertion that blockHandleLikelyMaxLen is large enough now incorporates the +1 of the valuePrefix byte explicitly. This was the only usage the depended on HandleMaxLen being inclusive of the valuePrefix byte.

Additionally, this commit documents blockHandleMaxLenWithoutProperties and adds an assertion that blockHandleLikelyMaxLen is larger than the length of an inline blob value handle (blob.MaxInlineHandleLength) plus the valuePrefix byte.

### DIFF
--- a/sstable/internal.go
+++ b/sstable/internal.go
@@ -7,6 +7,7 @@ package sstable
 import (
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/keyspan"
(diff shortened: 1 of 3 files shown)


RaduBerinde (approved):

jbowens:
TFTR!

PR #4388, opened 2025-03-15
sstable: update testutil readall to return an error

kruskall:
ReadAll is importing testutils which is pulling testify and other test dependencies in the final binary

move function to separate package since it's only called from test files

### DIFF
--- a/internal/compact/spans_test.go
+++ b/internal/compact/spans_test.go
@@ -132,10 +132,11 @@ func TestSplitAndEncodeSpan(t *testing.T) {
 			tw := sstable.NewRawWriter(obj, wo)
 			require.NoError(t, SplitAndEncodeSpan(base.DefaultComparer.Compare, &span, upToKey, tw))
 			require.NoError(t, tw.Close())
-			_, rangeDels, rangeKeys := sstable.ReadAll(obj, sstable.ReaderOptions{
+			_, rangeDels, rangeKeys, err := sstable.ReadAll(obj, sstable.ReaderOptions{
 				Comparer:   wo.Comparer,
 				KeySchemas: sstable.MakeKeySchemas(wo.KeySchema),
 			})
+			require.NoError(t, err)
 			require.LessOrEqual(t, len(rangeDels)+len(rangeKeys), 1)
 			s := "."
 			if all := append(rangeDels, rangeKeys...); len(all) == 1 {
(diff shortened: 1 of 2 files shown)


RaduBerinde:
How about updating the function to return an `error` instead of using `testutils.CheckErr` and leaving it where it is? It's only used in one test which can be easily updated to check for the error.

RaduBerinde (commented):
Just a couple of comments. Please squash the commits when you update. Thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4388#-:-OLRgZqJAeueUfvOWy7O:b3m1t01)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`sstable/test_utils.go` line 47 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4388#-OLRgCKi2ApP-98dhShc:-OLRgCKjABkHD9AX7Oc8:b8v5pta) ([raw file](https://github.com/cockroachdb/pebble/blob/34e225e484141fcc7c79083f673fbb2f4a6a8601/sstable/test_utils.go#L47)):*
> ```Go
> 		return nil, nil, nil, err
> 	}
> 	defer rangeDelIter.Close()
> ```

rangeDelIter can still be nil, we have to check
___
*[`sstable/test_utils.go` line 48 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4388#-OLRgPzi47BzznpF5F1J:-OLRgPzi47BzznpF5F1K:b-98umrj) ([raw file](https://github.com/cockroachdb/pebble/blob/34e225e484141fcc7c79083f673fbb2f4a6a8601/sstable/test_utils.go#L48)):*
> ```Go
> 	}
> 	defer rangeDelIter.Close()
> 	d, err := rangeDelIter.First()
> ```

[nit] I'd keep the `s` or use `sp` (for "span"). Same below
___
*[`sstable/test_utils.go` line 61 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4388#-OLRgEcS0ZH8d2HQabm2:-OLRgEcS0ZH8d2HQabm3:b-2guby3) ([raw file](https://github.com/cockroachdb/pebble/blob/34e225e484141fcc7c79083f673fbb2f4a6a8601/sstable/test_utils.go#L61)):*
> ```Go
> 
> 	rangeKeyIter, err := reader.NewRawRangeKeyIter(ctx, NoFragmentTransforms)
> 	if err != nil {
> ```

same here

RaduBerinde:
Published v2.0.3.

PR #4393, opened 2025-03-18
db: add mechanics for compaction-time value separation

jbowens:
Adapt the compact.Runner interface to support the configuration of a value separation policy via a compact.ValueSeparation interface, and introduce three implementations of this interface:
  - `writeNewBlobFiles` for writing blob files, separating eligible values to at most 1 blob file per output sstable.
  - `preserveBlobReferences` for writing out existing blob references as they are, accumulating BlobReferences for each output sstable
  - `NeverSeparateValues` for always retrieving values and never writing external blob references.

This commit introduces the interface and implementations only. These new mechanics are not yet in-use.

Informs #112.


RaduBerinde (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4393)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @sumeerbhola)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OLeaLiy3nNPpCOpIE7e:-OLeaLiy3nNPpCOpIE7f:byuxzud) (no related file):*
:lgtm:  Clean integration!
___
*[`internal/testutils/blob_handles.go` line 177 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OLea9Ti7BtwmIOex8ZA:-OLea9Ti7BtwmIOex8ZB:b-3zpsv3) ([raw file](https://github.com/cockroachdb/pebble/blob/92c2c1db50888d4e659b803d27fac19113838ea3/internal/testutils/blob_handles.go#L177)):*
> ```Go
> const debugParserSeparators = ":-[](){};="
> 
> func makeDebugParser(s string) debugParser {
> ```

[nit] we should make this public and reuse it for manifest tests
___
*[`internal/compact/run.go` line 133 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OLe_zS6BaAdB5cZxL-b:-OLe_zS6BaAdB5cZxL-c:b-8fkohq) ([raw file](https://github.com/cockroachdb/pebble/blob/92c2c1db50888d4e659b803d27fac19113838ea3/internal/compact/run.go#L133)):*
> ```Go
> 	// table's TableMetadata, and stats and metadata describing a newly
> 	// constructed blob file if any.
> 	CloseOutput() (ValueSeparationMetadata, error)
> ```

[nit] maybe Finish or FinishOutput

sumeerbhola (approved):
> Reviewed 3 of 8 files at r1, 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4393)* status: all files reviewed, 10 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OLeaLiy3nNPpCOpIE7e:-OM3YuDw3m7k-aKPTTDM:bg7igkv) (no related file):*

TFTR!
</blockquote></details>

:lgtm: 
___
*[`internal/testutils/blob_handles.go` line 22 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3Eg9_0TRWS0AJIwW0:-OM3Eg9_0TRWS0AJIwW1:b-hkqz7c) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/internal/testutils/blob_handles.go#L22)):*
> ```Go
> type BlobValues struct {
> 	mostRecentHandle blob.Handle
> 	trackedHandles   map[blob.Handle]string
> ```

Can you add a comment that the value in this map is the blob value.
___
*[`internal/testutils/blob_handles.go` line 27 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3EMlS6oLCsjGKVuZ0:-OM3EMlTApwKEugT8Jij:b-sarsl1) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/internal/testutils/blob_handles.go#L27)):*
> ```Go
> // Fetch returns the value corresponding to the given handle.
> func (bv *BlobValues) Fetch(
> 	ctx context.Context, handle []byte, blobFileNum base.DiskFileNum, valLen uint32, buf []byte,
> ```

- `buf` seems unused.
- `handleSuffix` would be a better name than `handle`.
___
*[`value_separation.go` line 35 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3QptjE-D9oOIWaaMN:-OM3QptjE-D9oOIWaaMO:b-gwx8qg) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/value_separation.go#L35)):*
> ```Go
> 	// to the sstable (but may still be written to a value block within the
> 	// sstable).
> 	minimumSize int
> ```

If this is the policy specification, we will probably want to separate it into a separate struct eventually. No need to do it now, of course.
___
*[`value_separation.go` line 85 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3Xkko2Qhe-VRznbWr:-OM3Xkko2Qhe-VRznbWs:b-1oglcj) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/value_separation.go#L85)):*
> ```Go
> 	}
> 
> 	// Keys that are too small are never separated.
> ```

nit: Values that are ...
___
*[`value_separation.go` line 111 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3YNjBBquUzuQA_NRo:-OM3YNjBBquUzuQA_NRp:b-egt7vv) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/value_separation.go#L111)):*
> ```Go
> 	// If there's a configured short attribute extractor, extract the value's
> 	// short attribute.
> 	var shortAttr base.ShortAttribute
> ```

just realized we don't currently have
```
// TODO(sumeer): in CockroachDB use LongAttribute for the clock timestamp in
// MVCC values, since we want to be able to check the uncertainty interval for
// a read without doing the expensive work of retrieving the value.

// LongAttribute is an attribute stored with the key when the actual value is
// stored in a blob file.
type LongAttribute []byte

// LongAttributeMaxLen is the maximum length permitted for a LongAttribute.
const LongAttributeMaxLen = 24
```

To be added later?
___
*[`value_separation.go` line 219 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3W_sOAzlZ01frBL91:-OM3W_sOAzlZ01frBL92:b-ufq4q) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/value_separation.go#L219)):*
> ```Go
> 	// compressible data, it overestimates the disk space consumed by the blob
> 	// references. It also does not include the blob file's index block or
> 	// footer, so it can underestimate if values are completely incompressible.
> ```

completely compressible?
___
*[`value_separation.go` line 222 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3Wg-14d1p37VOn1zH:-OM3Wg-14d1p37VOn1zI:bfkd8ql) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/value_separation.go#L222)):*
> ```Go
> 	//
> 	// Should we compute a compression ratio per blob file and scale the
> 	// references appropriately?
> ```

I think that was the todo in the prototype too.
___
*[`sstable/blob/blob.go` line 156 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3DDd82pPRSxgM1HfD:-OM3DDd82pPRSxgM1HfE:b9mokis) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/sstable/blob/blob.go#L156)):*
> ```Go
> 	sz := w.stats.FileLen                                    // Completed blocks
> 	sz += uint64(w.b.Size()) + block.TrailerLen              // Pending uncompressed block
> 	sz += uint64(w.stats.BlockCount+1)*12 + block.TrailerLen // Index block (assuming worst case of 4 bytes per block)
> ```

Does this comment mean 4 bytes for each of blockNum, blockOffset, blockLen, since it is multiplying by 12? If yes, can you tweak the comment to elaborate.
___
*[`compaction.go` line 3281 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3UPsyBmUGmeh_VUF_:-OM3UPsyBmUGmeh_VUFa:b-i1xqb2) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/compaction.go#L3281)):*
> ```Go
> 	writerOpts.Parallelism =
> 		d.opts.Experimental.MaxWriterConcurrency > 0 &&
> 			(cpuWorkHandle.Permitted() || d.opts.Experimental.ForceWriterParallelism)
> ```

aside: We should delete the code for `CPUWorkPermissionGranter`, `CPUWorkHandle` and `MaxWriterConcurrency` and simply use the `ForceWriterParallelism` to toggle between 1 and 2 goroutines. It is extremely unlikely that we will ever support more than 2 goroutines per writer since the generality becomes harder to integrate with admission control.
The only potential benefit is faster multi-level compactions, but part of the problem with multi-level compactions was that a compaction slot was taken for too long, and not necessarily our inability to find a concurrent compaction. And a fixed concurrency will go away with better AC support. And if we can't increase concurrency because disk is fully utilized, we can't actually allow the multi-level compaction to go faster (even if we had the capability).

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4393)* status: 6 of 9 files reviewed, 9 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OLeaLiy3nNPpCOpIE7e:-OM8UvYO6qIDBWHsFHg6:bz5mair) (no related file):*

:lgtm:
</blockquote></details>

TFTRs!
___
*[`compaction.go` line 3281 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3UPsyBmUGmeh_VUF_:-OM8Ss3_-hCGMKzy4FXM:bvgxhqa) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/compaction.go#L3281)):*

aside: We should delete the code for `CPUWorkPermissionGranter`, `CPUWorkHandle` and `MaxWriterConcurrency` and simply use the `ForceWriterParallelism` to toggle between 1 and 2 goroutines. It is extremely unlikely that we will ever support more than 2 goroutines per writer since the generality becomes harder to integrate with admission control.
The only potential benefit is faster multi-level compactions, but part of the problem with multi-level compactions was that a compaction slot was taken for too long, and not necessarily our inability to find a concurrent compaction. And a fixed concurrency will go away with better AC support. And if we can't increase concurrency because disk is fully utilized, we can't actually allow the multi-level compaction to go faster (even if we had the capability).
</blockquote></details>

Yeah, the columnar writer won't respect `WriterOptions.Parallelism` regardless and does not perform parallel data block compression. It always uses 1 additional goroutine for performing the write syscalls out-of-band from the main compaction loop.
___
*[`value_separation.go` line 35 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3QptjE-D9oOIWaaMN:-OM8TUHd3fUZ4U3AmM86:b2nhb2c) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/value_separation.go#L35)):*

If this is the policy specification, we will probably want to separate it into a separate struct eventually. No need to do it now, of course.
</blockquote></details>

Ack
___
*[`value_separation.go` line 85 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3Xkko2Qhe-VRznbWr:-OM8TZZI8n2r00A5vpM_:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/value_separation.go#L85)):*

nit: Values that are ...
</blockquote></details>

Done.
___
*[`value_separation.go` line 111 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3YNjBBquUzuQA_NRo:-OM8TaTB3bvNOhFz4-9b:b-m7ijg5) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/value_separation.go#L111)):*

just realized we don't currently have
```
// TODO(sumeer): in CockroachDB use LongAttribute for the clock timestamp in
// MVCC values, since we want to be able to check the uncertainty interval for
// a read without doing the expensive work of retrieving the value.

// LongAttribute is an attribute stored with the key when the actual value is
// stored in a blob file.
type LongAttribute []byte

// LongAttributeMaxLen is the maximum length permitted for a LongAttribute.
const LongAttributeMaxLen = 24
```

To be added later?
</blockquote></details>

Yeah, I'll follow up with it. Didn't want to add until we nailed down the general schema of how we're going to represent the blob handles.
___
*[`value_separation.go` line 219 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3W_sOAzlZ01frBL91:-OM8TtXg2uwdr4NuejaR:b-gsve1i) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/value_separation.go#L219)):*

completely compressible?
</blockquote></details>

I think incompressible is right here — if the values are incompressible so the disk space of the values themselves is the same, then the blob file's index block and footer are additive, resulting the total value size being a net underestimate. 
___
*[`value_separation.go` line 222 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3Wg-14d1p37VOn1zH:-OM8U7HTCl804s8MsA9p:b2nhb2c) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/value_separation.go#L222)):*

I think that was the todo in the prototype too.
</blockquote></details>

Ack
___
*[`internal/testutils/blob_handles.go` line 22 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3Eg9_0TRWS0AJIwW0:-OM8UPDw81oMSEeKS7Wx:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/internal/testutils/blob_handles.go#L22)):*

Can you add a comment that the value in this map is the blob value.
</blockquote></details>

Done.
___
*[`internal/testutils/blob_handles.go` line 27 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3EMlS6oLCsjGKVuZ0:-OM8UcLP37ScwU_tICE8:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/internal/testutils/blob_handles.go#L27)):*

- `buf` seems unused.
- `handleSuffix` would be a better name than `handle`.
</blockquote></details>

Done.
___
*[`sstable/blob/blob.go` line 156 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4393#-OM3DDd82pPRSxgM1HfD:-OM8UqSmBUCXLoEqpC51:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8cccc92f15bbffd8a757d460dff7d90dfb8b1bdc/sstable/blob/blob.go#L156)):*

Does this comment mean 4 bytes for each of blockNum, blockOffset, blockLen, since it is multiplying by 12? If yes, can you tweak the comment to elaborate.
</blockquote></details>

Done.

PR #4395, opened 2025-03-18
crl-release-24.3: sstable: copy filter block during suffix rewriting

jbowens:
Previously, during suffix rewriting, both row and columnar sstable writers
could pass the input sstable's filter block into objstorage.Writable.Write by
passing a subslice of the original input sstables byte slice. The Write method
is permitted to mangle the input buffer, which in this case would corrupt the
original input sstable.

Informs cockroachdb/cockroach#141493.
Backport of #4265.

### DIFF
--- a/objstorage/test_utils.go
+++ b/objstorage/test_utils.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"context"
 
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):

jbowens:
TFTR!

PR #4396, opened 2025-03-18
sstable: add Compressor and Decompressor interfaces

EdwardX29:
add Compressor and Decompressor interfaces to refactor existing compressor and decompressor schemes.


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4396)* status: 0 of 7 files reviewed, 3 unresolved discussions

___
*[`sstable/block/compressor.go` line 9 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4396#-OLfigwy4B95uFegkBBE:-OLfigwy4B95uFegkBBF:b-3s9t3n) ([raw file](https://github.com/cockroachdb/pebble/blob/a58813d2c68e0f0d498cd6b73b20fde91df00bf2/sstable/block/compressor.go#L9)):*
> ```Go
> }
> 
> type NoCompressionCompressor struct{}
> ```

[nit] NoopCompressor
___
*[`sstable/block/compressor.go` line 39 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4396#-OLfi-0S4E8zMxoq-GTt:-OLfi-0S4E8zMxoq-GTu:bapv2b5) ([raw file](https://github.com/cockroachdb/pebble/blob/a58813d2c68e0f0d498cd6b73b20fde91df00bf2/sstable/block/compressor.go#L39)):*
> ```Go
> }
> 
> type Decoder interface {
> ```

`block.Decoder` suggests something else. I'd rename to `Decompressor`
___
*[`sstable/block/compressor.go` line 40 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4396#-OLfi9Oe8XuTsj4RuOB_:-OLfi9Oe8XuTsj4RuOBa:b-k8op2o) ([raw file](https://github.com/cockroachdb/pebble/blob/a58813d2c68e0f0d498cd6b73b20fde91df00bf2/sstable/block/compressor.go#L40)):*
> ```Go
> 
> type Decoder interface {
> 	Decode(buf, compressed []byte) ([]byte, error)
> ```

This needs to be documented, especially that buf needs to be large enough. (Actually, this implies that the caller needs to know the size, so why not require it is exactly the correct size instead of returning a `[]byte`?)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4396)* status: 0 of 7 files reviewed, 2 unresolved discussions

___
*[`sstable/block/compressor.go` line 45 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4396#-OLkFqbxEjtDpIAwNz7G:-OLkFqbxEjtDpIAwNz7H:b-davg6e) ([raw file](https://github.com/cockroachdb/pebble/blob/1aeb60afc466acc54083a1c607231de98e0441cf/sstable/block/compressor.go#L45)):*
> ```Go
> type Decompressor interface {
> 	// DecompressInto decompresses compressed into buf. The buf slice must have the
> 	// exact size as the decompressed value. Callers may use DecompressedLen to
> ```

We should make `DecompressedLen` part of this interface as well.
___
*[`sstable/block/compressor.go` line 50 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4396#-OLkG4mR7x_8U1JddRuW:-OLkG4mR7x_8U1JddRuX:b-fz2e8y) ([raw file](https://github.com/cockroachdb/pebble/blob/1aeb60afc466acc54083a1c607231de98e0441cf/sstable/block/compressor.go#L50)):*
> ```Go
> }
> 
> type NoopDecompressor struct{}
> ```

Do these types need to be exported?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4396)* status: 0 of 7 files reviewed, 1 unresolved discussion

___
*[`sstable/block/compressor.go` line 56 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4396#-OLkRbvPD4w4scIdzDQK:-OLkRbvPD4w4scIdzDQL:b-h6qxtk) ([raw file](https://github.com/cockroachdb/pebble/blob/7a4e4dd62b24488fe7120186fb3211279dee9d08/sstable/block/compressor.go#L56)):*
> ```Go
> 	// varint encoding the length of the decompressed block. In such cases, a
> 	// non-zero prefixLength is returned indicating the length of this prefix.
> 	DecompressedLen(b []byte) (decompressedLen int, prefixLength int, err error)
> ```

We should remove the `prefixLength`, this should be an internal detail that is handled by the zstd implementation.

RaduBerinde (approved):
:lgtm_strong:  Nice cleanup!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4396#-:-OLlHoob0CPbmbV5b8by:b24l8lv)* status: 0 of 10 files reviewed, all discussions resolved

PR #4401, opened 2025-03-19
crl-release-23.2: sstable: copy filter block during suffix rewriting

jbowens:
Previously, during suffix rewriting, both row and columnar sstable writers could pass the input sstable's filter block into objstorage.Writable.Write by passing a subslice of the original input sstables byte slice. The Write method is permitted to mangle the input buffer, which in this case would corrupt the original input sstable.

Backport of #4265.
Informs cockroachdb/cockroach#141493.


jbowens:
TFTR!

PR #4402, opened 2025-03-19
internal/strparse: new package

jbowens:
Pull out the manifest package's debugParser used to facilitate parsing of debug representation's of data structures into a new internal package. This is in preparation for an additional use in parsing debug representations of blob handles.


RaduBerinde (approved):
Nice, I can already use this with a change I'm working on.

PR #4403, opened 2025-03-19
db, manifest: clean up error cases when picking compactions

RaduBerinde:
I'm not sure this can fail currently, but we should still check it. A
WIP change caused `l0SublevelInfo not created for compaction out of
L0` because of this missing check.

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -1482,6 +1482,8 @@ func (p *compactionPickerByScore) pickedCompactionFromCandidateFile(
 	}
 
 	if !pc.setupInputs(p.opts, env.diskAvailBytes, pc.startLevel) {
+		// TODO(radu): do we expect this to happen? (it does seem to happen if I add
+		// a log here).
 		return nil
 	}
 
@@ -1610,6 +1612,7 @@ func pickAutoLPositive(
 	}
 
 	if !pc.setupInputs(opts, env.diskAvailBytes, pc.startLevel) {
+		opts.Logger.Errorf("%v", base.AssertionFailedf("setupInputs failed"))
 		return nil
 	}
 	return pc.maybeAddLevel(opts, env.diskAvailBytes)
@@ -1740,7 +1743,7 @@ func (wa WriteAmpHeuristic) String() string {
 
 // Helper method to pick compactions originating from L0. Uses information about
 // sublevels to generate a compaction.
-func pickL0(env compactionEnv, opts *Options, vers *version, baseLevel int) (pc *pickedCompaction) {
+func pickL0(env compactionEnv, opts *Options, vers *version, baseLevel int) *pickedCompaction {
 	// It is important to pass information about Lbase files to L0Sublevels
 	// so it can pick a compaction that does not conflict with an Lbase => Lbase+1
 	// compaction. Without this, we observed reduced concurrency of L0=>Lbase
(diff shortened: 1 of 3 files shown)


RaduBerinde:
*[`compaction_picker.go` line 1754 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4403#-OLjAXg1AlfzU2hym0Hq:-OLjAXg1AlfzU2hym0Hr:b-dfitlr) ([raw file](https://github.com/cockroachdb/pebble/blob/acfd0e6ca84c55e599b1f0348a5a91a9bfa09dbd/compaction_picker.go#L1754)):*
> ```Go
> 	if err != nil {
> 		opts.Logger.Errorf("error when picking base compaction: %s", err)
> 		return
> ```

Would it be better to fall through in these error cases and try to pick other types of compactions?

jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4403#-:-OLk6F438VJgHu9Da8oT:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`compaction_picker.go` line 1754 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4403#-OLjAXg1AlfzU2hym0Hq:-OLk5jPZ8LgkZPr_8n1x:b2hz993) ([raw file](https://github.com/cockroachdb/pebble/blob/acfd0e6ca84c55e599b1f0348a5a91a9bfa09dbd/compaction_picker.go#L1754)):*

Would it be better to fall through in these error cases and try to pick other types of compactions?
</blockquote></details>

I'm staring at the error cases of PickBaseCompaction struggling to comprehend. I'm wondering if these should be assertion failures or, maybe non-error cases (resulting in a nil `lcf` and a `nil` error).

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4403)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`compaction_picker.go` line 1754 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4403#-OLjAXg1AlfzU2hym0Hq:-OLo9mlW7Ff1FWh922li:b5ool52) ([raw file](https://github.com/cockroachdb/pebble/blob/acfd0e6ca84c55e599b1f0348a5a91a9bfa09dbd/compaction_picker.go#L1754)):*

I'm staring at the error cases of PickBaseCompaction struggling to comprehend. I'm wondering if these should be assertion failures or, maybe non-error cases (resulting in a nil `lcf` and a `nil` error).
</blockquote></details>

I'm trying to refactor and use assertions but I'm hitting `file %s chosen as seed file for compaction should not be compacting` in practice. Is it possible that `isBaseCompacting` is not up-to-date? If a compaction started since the `L0Sublevels` were calculated, that would be expected, no?

RaduBerinde:
*[`compaction_picker.go` line 1754 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4403#-OLjAXg1AlfzU2hym0Hq:-OLpeqNHAagwGnE_SR2E:b-tv30jc) ([raw file](https://github.com/cockroachdb/pebble/blob/acfd0e6ca84c55e599b1f0348a5a91a9bfa09dbd/compaction_picker.go#L1754)):*

I'm trying to refactor and use assertions but I'm hitting `file %s chosen as seed file for compaction should not be compacting` in practice. Is it possible that `isBaseCompacting` is not up-to-date? If a compaction started since the `L0Sublevels` were calculated, that would be expected, no?
</blockquote></details>

@sumeerbhola pointed out that we do update this in `UpdateStateForStartedCompaction` so this seems unexpected. I'll investigate some more.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4403)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens, @RaduBerinde, and @sumeerbhola)

___
*[`compaction_picker.go` line 1754 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4403#-OLjAXg1AlfzU2hym0Hq:-OLsf3Mn0H0lk6C8i_x7:b24zs5y) ([raw file](https://github.com/cockroachdb/pebble/blob/acfd0e6ca84c55e599b1f0348a5a91a9bfa09dbd/compaction_picker.go#L1754)):*

@sumeerbhola pointed out that we do update this in `UpdateStateForStartedCompaction` so this seems unexpected. I'll investigate some more.
</blockquote></details>

Thanks for investigating. I agree that it's worrisome that some information is stale and we're relying on some last line of defense to return an error, instead of that last line being an assertion.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4403)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`compaction_picker.go` line 1754 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4403#-OLjAXg1AlfzU2hym0Hq:-OLt6eLcDT7qJyf-J-Cx:ba266cx) ([raw file](https://github.com/cockroachdb/pebble/blob/acfd0e6ca84c55e599b1f0348a5a91a9bfa09dbd/compaction_picker.go#L1754)):*

Thanks for investigating. I agree that it's worrisome that some information is stale and we're relying on some last line of defense to return an error, instead of that last line being an assertion.
</blockquote></details>

I checked again and I think the only problem was the test I'm fixing in #4409. I will proceed to convert to assertions.

@sumeerbhola - any opinion about whether I should keep these assertions in `invariants` builds only? (and just don't pick the compaction in release builds).  It feels like there could be some cornercase bugs in updating the `L0Sublevels` but they would resolve when the in-progress compactions complete, so to me it feels preferable to continue instead of crashing. I can add an `Errorf` log so they are still visible.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4403)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens, @RaduBerinde, and @sumeerbhola)

___
*[`compaction_picker.go` line 1754 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4403#-OLjAXg1AlfzU2hym0Hq:-OM3B6QECI7miZbRHIV0:betvxu6) ([raw file](https://github.com/cockroachdb/pebble/blob/acfd0e6ca84c55e599b1f0348a5a91a9bfa09dbd/compaction_picker.go#L1754)):*

I checked again and I think the only problem was the test I'm fixing in #4409. I will proceed to convert to assertions.

@sumeerbhola - any opinion about whether I should keep these assertions in `invariants` builds only? (and just don't pick the compaction in release builds).  It feels like there could be some cornercase bugs in updating the `L0Sublevels` but they would resolve when the in-progress compactions complete, so to me it feels preferable to continue instead of crashing. I can add an `Errorf` log so they are still visible.
</blockquote></details>

+1 to error logging in the normal case and asserting in the `invariants` build.

RaduBerinde (commented):
Updated.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4403#-:-OM5LxawDBR9hRmN6Y0P:b-bw8e14)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @jbowens and @sumeerbhola)

RaduBerinde:
Sigh.. the metamorphic test ran into `panic: setupInputs failed`

jbowens (approved):
:lgtm_strong:

> Reviewed 2 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4403#-:-OMSASak6gjRjxLAoOUM:b-jbsf8z)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #4404, opened 2025-03-19
manifest: cleanup related to making level slices

RaduBerinde:
#### manifest: add separate makeLevelSlice

All callers of `makeBTree` either need just the tree or just the
slice, so make a separate function for the slice.


RaduBerinde:
TFTR!

PR #4405, opened 2025-03-20
manifest: add L0Sublevels sanity checks, fix L0Index

RaduBerinde:
Add sanity checking to `L0Sublevels`.

The `L0Index` check was failing, suggesting that the code below made
an incorrect assumption about files exactly matching the last files in
the new levelMetadata:

```
	// Update interval indices for new files.
	for i, f := range files {
		f.L0Index = s.levelMetadata.Len() + i
```

I added extra checks to `AddL0Files` to bail out if this assumption is
not correct.

### DIFF
--- a/internal/manifest/l0_sublevels.go
+++ b/internal/manifest/l0_sublevels.go
@@ -331,6 +331,7 @@ func NewL0Sublevels(
 	}
 
 	s.calculateFlushSplitKeys(flushSplitMaxBytes)
+	s.Check()
 	return s, nil
 }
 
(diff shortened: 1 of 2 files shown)


jbowens (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4405)* status: all files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4405#-OLotihwBZadCTi23wna:-OLotihwBZadCTi23wnb:bg7igkv) (no related file):*
:lgtm: 
___
*[`internal/manifest/l0_sublevels.go` line 461 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4405#-OLorwXPD2aIQtTOmvAt:-OLorwXPD2aIQtTOmvAu:b-mpq3pt) ([raw file](https://github.com/cockroachdb/pebble/blob/90ca6589ddd2f85308866e3c03b2ce91491b5116/internal/manifest/l0_sublevels.go#L461)):*
> ```Go
> 			panic("AddL0Files called twice on the same receiver")
> 		}
> 		return nil, errInvalidL0SublevelsOpt
> ```

we mind as well make theses panic unconditional in invariants and non-invariants builds, right?

RaduBerinde:
*[`internal/manifest/l0_sublevels.go` line 461 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4405#-OLorwXPD2aIQtTOmvAt:-OLpffBa7ll3D0TZ1vNs:b-x0knfl) ([raw file](https://github.com/cockroachdb/pebble/blob/90ca6589ddd2f85308866e3c03b2ce91491b5116/internal/manifest/l0_sublevels.go#L461)):*

we mind as well make theses panic unconditional in invariants and non-invariants builds, right?
</blockquote></details>

I was thinking since this function is only an optimization and we can fall back on the slower one, we can reliably continue even if there's a bug here.

Issue #2571, opened 2023-05-25
sstable: do a bit flip computation on a checksum mismatch in the Reader


bananabrick:
In issues like [this](https://github.com/cockroachlabs/support/issues/2323), where the customer is using an encrypted store it is difficult to determine if the corruption occurred due to a bit flip.

While it is likely that the corruption occurred due to a bit flip, the answer isn't satisfactory to all customers without proof.

Instead, we could perform the check to see if there's a bit flip, when we detect the corruption, and log that to disk. We're going to panic and crash the process anyway, so the computation should be harmless.

PR #4406, opened 2025-03-20
sstable: do a bit flip computation on a checksum mismatch in the Reader

EdwardX29:
When a checksum mismatch occurs, this update tries flipping each bit of the data to see if a single-bit error caused the issue. If found, the error message includes the bit and index details.

Fixes: https://github.com/cockroachdb/pebble/issues/2571

### DIFF
--- a/sstable/block/block.go
+++ b/sstable/block/block.go
@@ -7,8 +7,10 @@ package block
 import (
 	"context"
 	"encoding/binary"
+	"fmt"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"time"
 
 	"github.com/cespare/xxhash/v2"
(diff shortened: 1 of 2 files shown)


RaduBerinde (commented):
Nice! It would be good to exercise this in a test, you can extend `TestReaderChecksumErrors` a bit (can add two subtest variants, one that has a single bit flip in a random byte).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4406#-:-OLoDSjCC2qEooX0_dLa:b-o3tet5)* status: 0 of 1 files reviewed, 5 unresolved discussions (waiting on @jbowens)

___
*[`sstable/block/block.go` line 176 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4406#-OLoCTL1EEU_a_MhL20a:-OLoCTL1EEU_a_MhL20b:bqs67t8) ([raw file](https://github.com/cockroachdb/pebble/blob/9cbd95a8b6c1e9871a414415af16eab5af142237/sstable/block/block.go#L176)):*
> ```Go
> 		flipped, indexFlipped, bitFlipped := checkForBitFlip(b[:bh.Length+1], checksumType, expectedChecksum)
> 		if flipped {
> 			base.CorruptionErrorf("block %d/%d: %s checksum mismatch %x != %x. bit flip found: bit %d index %d",
> ```

[nit] I'd just have a `extraMsg := ""` and update that and add it below as a `%s` at the end.
[nit] More useful to print the byte index and the before/after byte value in hex
___
*[`sstable/block/block.go` line 189 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4406#-OLoDpnf9R5ijFTEBj8_:-OLoDpnf9R5ijFTEBj8a:bv1cxn7) ([raw file](https://github.com/cockroachdb/pebble/blob/9cbd95a8b6c1e9871a414415af16eab5af142237/sstable/block/block.go#L189)):*
> ```Go
> func checkForBitFlip(
> 	data []byte, checksumType ChecksumType, expectedChecksum uint32,
> ) (flipped bool, indexFlipped int, bitFlipped int) {
> ```

[nit] s/flipped/found
___
*[`sstable/block/block.go` line 190 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4406#-OLoBFyeC-0Lk7zhJYMN:-OLoBFyeC-0Lk7zhJYMO:b-ethnkv) ([raw file](https://github.com/cockroachdb/pebble/blob/9cbd95a8b6c1e9871a414415af16eab5af142237/sstable/block/block.go#L190)):*
> ```Go
> 	data []byte, checksumType ChecksumType, expectedChecksum uint32,
> ) (flipped bool, indexFlipped int, bitFlipped int) {
> 	for i := 0; i < len(data); i++ {
> ```

Let's do a `data = slice.Clone(data)` here, just to be safe. (This function doesn't impose any restrictions on how `b` might be used in parallel)
___
*[`sstable/block/block.go` line 190 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4406#-OLoBuxQ9zentAXPxMp0:-OLoBuxQ9zentAXPxMp1:bcyqo09) ([raw file](https://github.com/cockroachdb/pebble/blob/9cbd95a8b6c1e9871a414415af16eab5af142237/sstable/block/block.go#L190)):*
> ```Go
> 	data []byte, checksumType ChecksumType, expectedChecksum uint32,
> ) (flipped bool, indexFlipped int, bitFlipped int) {
> 	for i := 0; i < len(data); i++ {
> ```

I would set a limit on `len(data)`, if we have a multi-megabyte block (which is possible, e.g. filter blocks), this is unlikely to finish in any reasonable amount of time. The limit will have to be somewhere north of 40KB to be useful.

I'd add a benchmark to see how long it takes for whatever limit you set, hoping something like 128KB is ok.
___
*[`sstable/block/block.go` line 193 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4406#-OLoD8dr-VBUJpPR_zHY:-OLoD8dr-VBUJpPR_zHZ:bcqpwny) ([raw file](https://github.com/cockroachdb/pebble/blob/9cbd95a8b6c1e9871a414415af16eab5af142237/sstable/block/block.go#L193)):*
> ```Go
> 		for bit := 0; bit < 8; bit++ {
> 			data[i] = data[i] ^ (1 << bit)
> 			var computedChecksum uint32
> ```

[nit] I'd extract this in a separate function

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4406#-:-OLuZlf3C2TMHioQfmpA:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jbowens)

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4406)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`sstable/block/block.go` line 186 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4406#-OLyELh80O0N7hdHhNJ_:-OLyELh80O0N7hdHhNJa:b5f5f00) ([raw file](https://github.com/cockroachdb/pebble/blob/582a17a33ad84e2ef8f349c02f40b3c112c3f67c/sstable/block/block.go#L186)):*
> ```Go
> 		return base.CorruptionErrorf("block %d/%d: %s checksum mismatch %x != %x%s",
> 			errors.Safe(bh.Offset), errors.Safe(bh.Length), checksumType,
> 			expectedChecksum, computedChecksum, bitFlipExtraMsg)
> ```

I'm pretty sure this `blitFlipExtraMsg` will get redacted
I think we should do:

```
err := base.CorruptionErrorf("block %d/%d: %s checksum mismatch %x != %x",
			errors.Safe(bh.Offset), errors.Safe(bh.Length), checksumType,
			expectedChecksum, computedChecksum)
if found {
	err = errors.WithSafeDetails(". bit flip found: byte index %d. got: %x. want: %x.",
				indexFound, data[indexFound], data[indexFound]^(1<<bitFound))
}
return err
```

PR #4409, opened 2025-03-21
db: fix CompactionPickerTargetLevel test

RaduBerinde:
This test did not properly set up the `L0Sublevels` with the
compaction information (in regular operation, `L0Sublevels` is kept in
sync with in-progress compactions).

To fix this, we use `L0Sublevels.InitCompactingFileInfo()` before
picking a compaction. This prevents an error case during compaction
picking.

### DIFF
--- a/compaction_picker_test.go
+++ b/compaction_picker_test.go
@@ -290,18 +290,31 @@ func TestCompactionPickerTargetLevel(t *testing.T) {
 					}
 				}
 
+				var l0InProgress []manifest.L0Compaction
+
 				// Mark files as compacting for each in-progress compaction.
 				for i := range inProgress {
 					c := &inProgress[i]
 					for j, cl := range c.inputs {
 						iter := vers.Levels[cl.level].Iter()
-						for f := iter.First(); f != nil; f = iter.Next() {
+						for f := iter.First(); ; f = iter.Next() {
+							if f == nil {
+								d.Fatalf(t, "could not find any non-compacting files in L%d", cl.level)
+							}
 							if !f.IsCompacting() {
 								f.CompactionState = manifest.CompactionStateCompacting
 								c.inputs[j].files = iter.Take().Slice()
 								break
 							}
 						}
+					}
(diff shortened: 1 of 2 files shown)


sumeerbhola (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4409)* status: all files reviewed, 3 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`compaction_picker_test.go` line 320 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4409#-OLsdwMiA4Z71Q_9JXLQ:-OLtvQig0iLVWzkNd_IK:b-9ezfdh) ([raw file](https://github.com/cockroachdb/pebble/blob/6bd3daff841eb9c9b7813392b578fa7b93e45e18/compaction_picker_test.go#L320)):*

I found it curious as well. Wouldn't it be much better if we could explicitly specify the compactions in the test? We could use a syntax similar to the one we use in logs: `L1 [0001,0002]  L2 [0003,00004] -> L5`
</blockquote></details>

I've always found the `ongoing` spec confusing. I have to read the code every time I want change or understand anything in this test.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4409)* status: all files reviewed, 3 unresolved discussions

___
*[`compaction_picker_test.go` line 320 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4409#-OLsdwMiA4Z71Q_9JXLQ:-OLtyEFy4dagTqaSWB-X:b-sa35fn) ([raw file](https://github.com/cockroachdb/pebble/blob/6bd3daff841eb9c9b7813392b578fa7b93e45e18/compaction_picker_test.go#L320)):*

I've always found the `ongoing` spec confusing. I have to read the code every time I want change or understand anything in this test.
</blockquote></details>

Filed https://github.com/cockroachdb/pebble/issues/4411

PR #4412, opened 2025-03-21
internal/manual: sometimes Go-allocate memory in race builds

jbowens:
It's unclear if the Go race detector can detect race in memory that's allocated by cgo, but under the assumption it can't, this commit adapts the manual package to allocate some of its allocations using standard Go allocations in race builds. The hope is that if data races exist in the data structures allocated using the manual package, sometimes Go-allocating this memory will surface them in our race-build test runs.

Motivated by cockroachdb/cockroach#142868.

### DIFF
--- a/internal/manual/manual_cgo.go
+++ b/internal/manual/manual_cgo.go
@@ -6,6 +6,12 @@ package manual
 
 // #include <stdlib.h>
 import "C"
+import (
+	"math/rand/v2"
+	"unsafe"
+
+	"github.com/cockroachdb/pebble/internal/invariants"
+)
 
 // The go:linkname directives provides backdoor access to private functions in
 // the runtime. Below we're accessing the throw function.
@@ -13,18 +19,36 @@ import "C"
 //go:linkname throw runtime.throw
 func throw(s string)
 
+// useGoAllocation is used in race-enabled builds to configure the package to
+// use ordinary Go allocations with make([]byte, n). This is done under the
+// assumption that the Go race detector will detect races within cgo-allocated
+// memory. Performing some allocations using Go allows the race detector to
+// observe concurrent memory access to memory allocated by this package.
+//
+// TODO(jackson): Confirm that the race detector does not detect races within
+// cgo-allocated memory.
+var useGoAllocation = invariants.RaceEnabled && rand.Uint32()%2 == 0
+
 // TODO(peter): Rather than relying an C malloc/free, we could fork the Go
 // runtime page allocator and allocate large chunks of memory using mmap or
 // similar.
 
-// New allocates a slice of size n. The returned slice is from manually managed
-// memory and MUST be released by calling Free. Failure to do so will result in
-// a memory leak.
+// New allocates a slice of size n. The returned slice is from manually
+// managed memory and MUST be released by calling Free. Failure to do so will
+// result in a memory leak.
 func New(purpose Purpose, n uintptr) Buf {
 	if n == 0 {
 		return Buf{}
 	}
 	recordAlloc(purpose, n)
+
+	// In race-enabled builds, we sometimes make allocations using Go to allow
+	// the race detector to observe concurrent memory access to memory allocated
+	// by this package. See the definition of useGoAllocation for more details.
+	if invariants.RaceEnabled && useGoAllocation {
+		b := make([]byte, n)
+		return Buf{data: unsafe.Pointer(&b[0]), n: n}
+	}
 	// We need to be conscious of the Cgo pointer passing rules:
 	//
 	//   https://golang.org/cmd/cgo/#hdr-Passing_pointers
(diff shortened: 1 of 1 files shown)


RaduBerinde (commented):
Is there an advantage to deciding on a per-allocation basis? A global `var useGoAllocation = rand.RandN(2) == 0` would be much simpler

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4412#-:-OLu_ApN2w7QyHpAvKIH:b1zorep)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @xxmplus)

jbowens (commented):
I’m wary of excessively slowing race builds if we only allocate using Go (in part due to normal tradeoffs, and part due to presumably many, more memory accesses the race detector is tracking). This was a speculative concern and if it is a valid concern maybe not resolved just by halving the number of Go allocations. So we could also do the simpler thing and see what breaks.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4412#-:-OLw5eGf9YAfhvZPwROM:b2f3d8v)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @xxmplus)

internal/manual/manual_cgo.go
```diff
@@ -6,25 +6,74 @@ package manual
...
+// goAllocationSeed is an arbitrary value used to seed the hash function used to
+// determine which allocations should be made using Go. See useGoAllocation and
+// the init func below.
+var goAllocationSeed uint64
+
+func init() {
+	if !invariants.RaceEnabled {
+		return
+	}
+	goAllocationSeed = uint64(uintptr(unsafe.Pointer(&goAllocationSeed)))
```
RaduBerinde, line 54: What are we doing here? Using the address as the seed?

internal/manual/manual_cgo.go
```diff
@@ -6,25 +6,74 @@ package manual
...
+// The choice of which allocations to make using Go is made deterministically
+// using a fibonacci hash of the allocation size and a seed derived from an
+// arbitrary pointer (which will change from run to run).
+func useGoAllocation(n uintptr) bool {
+	if !invariants.RaceEnabled {
+		return false
+	}
+	// See https://probablydance.com/2018/06/16/fibonacci-hashing-the-optimization-that-the-world-forgot-or-a-better-alternative-to-integer-modulo/
+	const m = 11400714819323198485
+	h := goAllocationSeed
```
RaduBerinde, line 40: Here only the high bit of the seed makes a difference.. and given two allocations a and b if the fibonacci hash has different high bits for them they will never be go allocated together. I'd xor before multiplying.

We could also build a table of bools for all possible sizes up to a certain value (using a standard rng) and just use that table.

RaduBerinde (commented):
I'm ok with this but I'd probably move to a different file with the race tag so the extra code doesn't interfere with inlining.

internal/manual/manual_cgo.go
sumeerbhola: Do we need the `goAllocationSeed` because the `uintptr` values won't be sufficiently random?
In reader_iter_single_lvl.go and iterator.go, we've been simply doing `(11400714819323198485 * uint64(ptr)) >> 63`, so if that is wrong, we should fix those too.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4412)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jbowens, @RaduBerinde, and @xxmplus)

internal/manual/manual_cgo.go
jbowens: > Do we need the `goAllocationSeed` because the `uintptr` values won't be sufficiently random?

the `n` here isn't actually a pointer, it's a size.

internal/manual/manual_cgo.go
jbowens: Yeah, I was originally hoping the address of the global would be fixed within a single build (but vary per-build), but I think PIE means each process will have a different address. I thought it would be nice if the behavior was fixed for allocations of certain sizes within a build to make reproductions easier.

jbowens (commented):
Let's try the simpler thing and see what happens

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4412#-:-OMDXRQvDU6XC6XdgTmC:b1p92fj)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @xxmplus)

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4412#-:-OMDfPK82g1wc393M8qz:bnfp4nl)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @xxmplus)

jbowens:
TFTR!

PR #4414, opened 2025-03-22
sstable: add Minlz Compression

EdwardX29:
add Minlz Compression

### DIFF
--- a/db.go
+++ b/db.go
@@ -2075,6 +2075,7 @@ func (d *DB) Metrics() *Metrics {
 		metrics.Table.CompressedCountUnknown += int64(compressionTypes.unknown)
 		metrics.Table.CompressedCountSnappy += int64(compressionTypes.snappy)
 		metrics.Table.CompressedCountZstd += int64(compressionTypes.zstd)
+		metrics.Table.CompressedCountMinlz += int64(compressionTypes.minlz)
 		metrics.Table.CompressedCountNone += int64(compressionTypes.none)
 	}
 
--- a/go.mod
+++ b/go.mod
@@ -14,8 +14,9 @@ require (
 	github.com/ghemawat/stream v0.0.0-20171120220530-696b145b53b9
 	github.com/golang/snappy v0.0.5-0.20231225225746-43d5d4cd4e0e
 	github.com/guptarohit/asciigraph v0.5.5
-	github.com/klauspost/compress v1.16.7
+	github.com/klauspost/compress v1.17.11
 	github.com/kr/pretty v0.3.1
+	github.com/minio/minlz v1.0.0
 	github.com/olekukonko/tablewriter v0.0.5
 	github.com/pkg/errors v0.9.1
 	github.com/pmezard/go-difflib v1.0.0
--- a/go.sum
+++ b/go.sum
@@ -127,8 +127,8 @@ github.com/jung-kurt/gofpdf v1.0.3-0.20190309125859-24315acbbda5/go.mod h1:7Id9E
 github.com/kisielk/errcheck v1.1.0/go.mod h1:EZBBE59ingxPouuu3KfxchcWSUPOHkagtvWXihfKN4Q=
 github.com/kisielk/errcheck v1.5.0/go.mod h1:pFxgyoBC7bSaBwPgfKdkLd5X25qrDl4LWUI2bnpBCr8=
 github.com/kisielk/gotool v1.0.0/go.mod h1:XhKaO+MFFWcvkIS/tQcRk01m1F5IRFswLeQ+oQHNcck=
-github.com/klauspost/compress v1.16.7 h1:2mk3MPGNzKyxErAw8YaohYh69+pa4sIQSC0fPGCFR9I=
-github.com/klauspost/compress v1.16.7/go.mod h1:ntbaceVETuRiXiv4DpjP66DpAtAGkEQskQzEyD//IeE=
+github.com/klauspost/compress v1.17.11 h1:In6xLpyWOi1+C7tXUUWv2ot1QvBjxevKAaI6IXrJmUc=
+github.com/klauspost/compress v1.17.11/go.mod h1:pMDklpSncoRMuLFrf1W9Ss9KT+0rH90U12bZKk7uwG0=
 github.com/konsorten/go-windows-terminal-sequences v1.0.1/go.mod h1:T0+1ngSBFLxvqU3pZ+m/2kptfBszLMUkC4ZK/EgS/cQ=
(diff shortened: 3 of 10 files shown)


jbowens (commented):
> Reviewed 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4414)* status: all files reviewed, 3 unresolved discussions (waiting on @RaduBerinde and @xxmplus)

___
*[`sstable/block/compressor.go` line 42 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4414#-ONHKIlmEbDTUjNeifeO:-ONHKIlmEbDTUjNeifeP:bt2zaoe) ([raw file](https://github.com/cockroachdb/pebble/blob/2a26810e4e35dbf2239aafe971793e37e6277eec/sstable/block/compressor.go#L42)):*
> ```Go
> func (minlzCompressor) Compress(dst, src []byte) (CompressionIndicator, []byte) {
> 	// Minlz cannot encode blocks greater than 8MB. Fall back to Snappy in those cases.
> 	if len(src) >= 8*(1<<20) {
> ```

can we add a unit test that tries compressing a payloads of sizes (8<<20 - 1), (8<<20) and (8<<20+1) to ensure something breaks if we accidentally change this conditional?

nit: `(8<<20)`
___
*[`sstable/block/compressor.go` line 48 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4414#-ONHKtyzEuULwWgfHa2e:-ONHKtyzEuULwWgfHa2f:biu2c15) ([raw file](https://github.com/cockroachdb/pebble/blob/2a26810e4e35dbf2239aafe971793e37e6277eec/sstable/block/compressor.go#L48)):*
> ```Go
> 	compressed, err := minlz.Encode(dst, src, minlz.LevelBalanced)
> 	if err != nil {
> 		panic("Error in MinlzCompressor Compress.")
> ```

nit: `panic(errors.Wrap(err, "minlz compression"))`
___
*[`sstable/block/compressor.go` line 145 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4414#-ONHL7ya5UmrPYhrnPlB:-ONHL7ya5UmrPYhrnPlC:bfmd3mw) ([raw file](https://github.com/cockroachdb/pebble/blob/2a26810e4e35dbf2239aafe971793e37e6277eec/sstable/block/compressor.go#L145)):*
> ```Go
> 
> func (minlzDecompressor) DecompressedLen(b []byte) (decompressedLen int, err error) {
> 	l, err := minlz.DecodedLen(b)
> ```

nit: `return minlz.DecodedLen(b)`

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4414#-:-ONKfAr49cYBpOGZZD5T:bnfp4nl)* status: 8 of 10 files reviewed, 4 unresolved discussions (waiting on @EdwardX29, @jbowens, and @xxmplus)

___
*[`sstable/block/compressor.go` line 42 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4414#-ONHKIlmEbDTUjNeifeO:-ONKeVuaAASrJOgaT_g6:basqng7) ([raw file](https://github.com/cockroachdb/pebble/blob/2a26810e4e35dbf2239aafe971793e37e6277eec/sstable/block/compressor.go#L42)):*

can we add a unit test that tries compressing a payloads of sizes (8<<20 - 1), (8<<20) and (8<<20+1) to ensure something breaks if we accidentally change this conditional?

nit: `(8<<20)`
</blockquote></details>

agreed, and the constant should be `minlz.MaxBlockSize` (or even better, handle `minlz.ErrTooLarge` error)
___
*[`sstable/block/compressor.go` line 46 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4414#-ONKd-nq3Vmh_7X6zUDa:-ONKd-nq3Vmh_7X6zUDb:bp5ojf8) ([raw file](https://github.com/cockroachdb/pebble/blob/b7137ce4e9062f7a3f7fd2d20c851c82df61ab97/sstable/block/compressor.go#L46)):*
> ```Go
> 	}
> 
> 	compressed, err := minlz.Encode(dst, src, minlz.LevelBalanced)
> ```

I would use `LevelFastest` - in the minlz readme, it was ~2x faster and not much worse in compression (and still much better than Snappy).

More generally (not in this PR of course), I think we should support all levels. `block.Compression` should be a struct that has a `Family` and a `Level`. And the `Compress` methods would take a level. We can hardcode the current defaults for level 0 for backward compatibility. The CompressionIndicator can still be just the family for old formats, though it might be useful to encode both in v6 format.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4414)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @jbowens and @xxmplus)

___
*[`sstable/block/compressor.go` line 42 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4414#-ONHKIlmEbDTUjNeifeO:-ONKsKz915U8aiUVUdBe:b-h0gn7p) ([raw file](https://github.com/cockroachdb/pebble/blob/2a26810e4e35dbf2239aafe971793e37e6277eec/sstable/block/compressor.go#L42)):*
<details><summary><i>Previously, EdwardX29 (Edward Xiao) wrote…</i></summary><blockquote>

switched unit test (https://reviewable.io/reviews/cockroachdb/pebble/4414#file-ONHakJttOtHcMwQAJVw) and this check to use `minlz.MaxBlockSize` and `minlz.ErrTooLarge`
</blockquote></details>

I think Jackson's suggestion was to verify that we can Compress and then Decompress a block of limit-1, limit, limit+1. We do need to test that the minlz decompress can handle the data compressed by snappy.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4414)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @jbowens and @xxmplus)

___
*[`sstable/block/compressor.go` line 42 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4414#-ONHKIlmEbDTUjNeifeO:-ONKubyOExtNzVxL1yco:b-yck6ak) ([raw file](https://github.com/cockroachdb/pebble/blob/2a26810e4e35dbf2239aafe971793e37e6277eec/sstable/block/compressor.go#L42)):*
<details><summary><i>Previously, EdwardX29 (Edward Xiao) wrote…</i></summary><blockquote>

Before decompressing, we do a switch statement on `CompressionIndicator` in `GetDecompressor(c CompressionIndicator)`. So when minlz compressor compresses with Snappy, the SnappyIndicator is returned (not Minlz Indicator). I added checks to see that the correct `CompressionIndicator` was returned after compression.
</blockquote></details>

Ah, I see, great!

PR #4418, opened 2025-03-24
sstable: remove WriterOptions.Parallelism

jbowens:
Remove the MaxWriterConcurrency option and the CPUWorkPermissionGranter
interface. The new columnar-block sstable writer does not support parallel
compression. We expect we do not want to parallelize work beneath the
granularity of compactions, in part because it is difficult to integrate with
admission control.

Remove WriterOptions.Parallelism. Only the rowblk writer partially supported
this option, and we don't anticipate ever supporting it in the colblk writer.


RaduBerinde (approved):
:lgtm_strong: :

> Reviewed 2 of 9 files at r1, 12 of 12 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4418#-:-OM9RxZL-CfCIGFJat2C:b-yncq55)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @xxmplus)

jbowens:
TFTR!

Issue #1192, opened 2021-07-13
sstable: isolate and report corruption in sstables


itsbilal:
This is one of the bigger Pebble-side changes as part of the sstable / replica corruption project in cockroach.
As part of this, sstable readers / iterators will build out a defined reporting path for any corruption
errors encountered when reading an sstable. Common corruption errors are "bad magic number",
"checksum mismatch", etc., and lots of instances of them are marked as dupes of
https://github.com/cockroachdb/cockroach/issues/57934 .

Pebble needs to report (through a `CorruptionReporter` interface) the key bounds of an instance
of corruption encountered. Pebble could optionally deduplicate / merge these key-ranges of corruption,
or leave that up to the entity running Pebble (Cockroach).

The easy thing to do when reporting corruption key bounds is to just report the sstable's bounds from
its FileMetadata as corrupt, if a corruption error is encountered when reading it. However, in cases where
the corruption is in a data block, the bounds could be narrowed further by
looking at adjacent non-corrupt data blocks and figure out the closest adjoining keys to the corrupt block(s) that
can be read correctly.

RaduBerinde:
This is done.

PR #4419, opened 2025-03-25
db: add problem spans data structure, check when picking auto compactions

### LINKED ISSUE #4285: db: surface corruption, missing sstable errors without fataling

RaduBerinde:
Note: first commit is #4403.

The structure uses a library I wrote over the weekend: https://github.com/RaduBerinde/axisds
I hope to add (over time) more data structures around intervals, order maintenance, etc.

#### problemspans: add data structure

Package problemspans provides functionality for tracking and managing
key spans that have been identified as problematic. It allows users to
add spans with associated expiration times, check if a given key range
overlaps any active (non-expired) spans, and remove spans when issues
are resolved.

#### db: check problem spans when picking auto compactions

We check the problem spans to avoid any overlap when picking
compactions. Note that normally the problem spans are empty and these
checks are avoided.

We add a test that verifies that compactions still go through after
trying to compact a non-existent external file.

Fixes #4285
Informs #1192

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1788,7 +1788,7 @@ func (d *DB) makeCompactionEnvLocked() *compactionEnv {
 	if d.closed.Load() != nil || d.opts.ReadOnly {
 		return nil
 	}
-	return &compactionEnv{
+	env := &compactionEnv{
 		diskAvailBytes:          d.diskAvailBytes.Load(),
 		earliestSnapshotSeqNum:  d.mu.snapshots.earliest(),
 		earliestUnflushedSeqNum: d.getEarliestUnflushedSeqNumLocked(),
@@ -1799,6 +1799,10 @@ func (d *DB) makeCompactionEnvLocked() *compactionEnv {
 			rescheduleReadCompaction: &d.mu.compact.rescheduleReadCompaction,
 		},
 	}
+	if !d.problemSpans.IsEmpty() {
+		env.problemSpans = &d.problemSpans
+	}
+	return env
 }
 
 // pickAnyCompaction tries to pick a manual or automatic compaction.
@@ -2324,7 +2328,7 @@ func (d *DB) compact(c *compaction, errChannel chan error) {
 		d.mu.Lock()
 		c.grantHandle.Started()
 		if err := d.compact1(c, errChannel); err != nil {
-			d.handleCompactFailure(err)
+			d.handleCompactFailure(c, err)
 		}
 		if c.isDownload {
 			d.mu.compact.downloadingCount--
@@ -2362,13 +2366,34 @@ func (d *DB) compact(c *compaction, errChannel chan error) {
 	})
 }
 
-func (d *DB) handleCompactFailure(err error) {
+func (d *DB) handleCompactFailure(c *compaction, err error) {
 	if errors.Is(err, ErrCancelledCompaction) {
 		// ErrCancelledCompaction is expected during normal operation, so we don't
 		// want to report it as a background error.
 		d.opts.Logger.Infof("%v", err)
 		return
 	}
+
+	// Record problem spans for a short duration, unless the error is a corruption.
+	expiration := 30 * time.Second
+	if IsCorruptionError(err) {
+		// TODO(radu): ideally, we should be using the corruption reporting
+		// mechanism which has a tighter span for the corruption. We would need to
+		// somehow plumb the level of the file.
+		expiration = 5 * time.Minute
+	}
+	for i := range c.inputs {
+		level := c.inputs[i].level
+		if level == 0 {
+			// We do not set problem spans on L0, as they could block flushes.
+			continue
+		}
+		it := c.inputs[i].files.Iter()
+		for f := it.First(); f != nil; f = it.Next() {
+			d.problemSpans.Add(level, f.UserKeyBounds(), expiration)
+		}
+	}
+
 	// TODO(peter): count consecutive compaction errors and backoff.
 	d.opts.EventListener.BackgroundError(err)
 }
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -17,6 +17,7 @@ import (
 	"github.com/cockroachdb/pebble/internal/humanize"
 	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/internal/manifest"
+	"github.com/cockroachdb/pebble/internal/problemspans"
 )
 
 // The minimum count for an intra-L0 compaction. This matches the RocksDB
@@ -39,6 +40,10 @@ type compactionEnv struct {
 	earliestSnapshotSeqNum  base.SeqNum
 	inProgressCompactions   []compactionInfo
 	readCompactionEnv       readCompactionEnv
+	// problemSpans is checked by the compaction picker to avoid compactions that
+	// overlap an active "problem span". It can be nil when there are no problem
+	// spans.
+	problemSpans *problemspans.ByLevel
 }
 
 type compactionPicker interface {
@@ -414,7 +419,10 @@ func (pc *pickedCompaction) maybeExpandBounds(smallest InternalKey, largest Inte
 // setupInputs returns true if a compaction has been set up. It returns false if
 // a concurrent compaction is occurring on the start or output level files.
 func (pc *pickedCompaction) setupInputs(
-	opts *Options, diskAvailBytes uint64, startLevel *compactionLevel,
+	opts *Options,
+	diskAvailBytes uint64,
+	startLevel *compactionLevel,
+	problemSpans *problemspans.ByLevel,
 ) bool {
 	// maxExpandedBytes is the maximum size of an expanded compaction. If
 	// growing a compaction results in a larger size, the original compaction
@@ -423,7 +431,7 @@ func (pc *pickedCompaction) setupInputs(
 		opts, adjustedOutputLevel(pc.outputLevel.level, pc.baseLevel), diskAvailBytes,
 	)
 
-	if anyTablesCompacting(startLevel.files) {
+	if !canCompactTables(startLevel.files, startLevel.level, problemSpans) {
 		return false
 	}
 
@@ -434,7 +442,7 @@ func (pc *pickedCompaction) setupInputs(
 	// left empty for those.
 	if startLevel.level != pc.outputLevel.level {
 		pc.outputLevel.files = pc.version.Overlaps(pc.outputLevel.level, pc.userKeyBounds())
-		if anyTablesCompacting(pc.outputLevel.files) {
+		if !canCompactTables(pc.outputLevel.files, pc.outputLevel.level, problemSpans) {
 			return false
 		}
 
@@ -504,7 +512,7 @@ func (pc *pickedCompaction) setupInputs(
 				*pc.lcf = *oldLcf
 			}
 		}
-	} else if pc.grow(pc.smallest, pc.largest, maxExpandedBytes, startLevel) {
+	} else if pc.grow(pc.smallest, pc.largest, maxExpandedBytes, startLevel, problemSpans) {
 		pc.maybeExpandBounds(manifest.KeyRange(pc.cmp,
 			startLevel.files.All(), pc.outputLevel.files.All()))
 	}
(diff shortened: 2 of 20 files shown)


sumeerbhola (changes requested):
flushing some initial comments -- haven't yet read the last commit.

> Reviewed 9 of 9 files at r2, 4 of 6 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4419#-:-OMrekNdB1ieUlCmWV2X:bmoyhz6)* status: 7 of 20 files reviewed, 4 unresolved discussions (waiting on @jbowens)

___
*[`internal/problemspans/set.go` line 30 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4419#-OMrdQu9C3brCij9RwuM:-OMrdQu9C3brCij9RwuN:b-rlkxyt) ([raw file](https://github.com/cockroachdb/pebble/blob/13d034afaceec4c54b15d42fe6336ce28127d552/internal/problemspans/set.go#L30)):*
> ```Go
> // To avoid blow-up in pathological corner cases, the Set limits the amount of
> // keys it stores internally (see SetSizeLimit). When this limit is hit, the
> // spans (or fragments of spans) that would expire first are dropped.
> ```

An alternative would be have coarse grained expiry times, so that adjacent regions are merged. I suppose there are two problems with that (a) there could be healthy and unhealthy regions alternating, (b) coarser grained expiry times also delay when a compaction will again be attempted, which may not be desirable.

Is this pathological case really a problem, given we are already tracking all files in the per-level btrees in the latest version? In the worst-case, if each file contributes a key span here, we would have the same number of entries here. And since we won't be doing compactions for those unexpired key spans, we won't have keyspans that only existed in old versions. So we are bounded by the size of the current version's spans.
___
*[`internal/problemspans/set.go` line 19 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4419#-OMrd1TP5FTdGpahMNDR:-OMrd1TQCIS3XpCJ-yvw:bvld3lg) ([raw file](https://github.com/cockroachdb/pebble/blob/4281ef6b18e203896627de0f452b09b1e242a3f4/internal/problemspans/set.go#L19)):*
> ```Go
> 
> // Set maintains a set of spans with expiration times and allows checking for
> // overlap against non-expired spans.
> ```

is the expiration there to allow retrying a compaction after some time in case it was a transient error that went away without excising?
___
*[`internal/problemspans/set.go` line 51 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4419#-OMrafb_6-Ywsi39iaV1:-OMrafb_6-Ywsi39iaV2:b-jhh7b6) ([raw file](https://github.com/cockroachdb/pebble/blob/4281ef6b18e203896627de0f452b09b1e242a3f4/internal/problemspans/set.go#L51)):*
> ```Go
> 	propEqFn := func(a, b expirationTime) bool {
> 		return a == b ||
> 			crtime.Mono(a) <= s.now && crtime.Mono(b) <= s.now // Both are expired or 0.
> ```

I think it is worth noting in a code comment somewhere that there is no explicit garbage collection, and that updates to regions will result in comparing with adjacent regions and expired regions will get merged.

Hmm, maybe I missed something -- does `regiontree.T.Enumerate` also mutate the tree, and could be doing some GC?
___
*[`internal/problemspans/set.go` line 79 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4419#-OMrcNNx4dRq5TSlf-E6:-OMrcNNx4dRq5TSlf-E7:b3l5u3s) ([raw file](https://github.com/cockroachdb/pebble/blob/4281ef6b18e203896627de0f452b09b1e242a3f4/internal/problemspans/set.go#L79)):*
> ```Go
> 	overlaps := false
> 	s.rt.Enumerate(start, end, func(start, end axisds.Endpoint[[]byte], exp expirationTime) bool {
> 		overlaps = true
> ```

can't this return something where `exp < s.now`?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4419)* status: 3 of 20 files reviewed, 4 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/problemspans/set.go` line 19 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4419#-OMrd1TP5FTdGpahMNDR:-OMsEo6fDiEPhuttNdlW:bynovqf) ([raw file](https://github.com/cockroachdb/pebble/blob/4281ef6b18e203896627de0f452b09b1e242a3f4/internal/problemspans/set.go#L19)):*

is the expiration there to allow retrying a compaction after some time in case it was a transient error that went away without excising?
</blockquote></details>

Yes - we add problem spans for all failures. 5 minutes for corruption, 30s for other errors. Plus even a corruption (which includes missing external table) could resolve itself (e.g. fixing permissions on the S3 objects).
___
*[`internal/problemspans/set.go` line 51 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4419#-OMrafb_6-Ywsi39iaV1:-OMsGNgMB4K1WMK-dFfT:bq0jij7) ([raw file](https://github.com/cockroachdb/pebble/blob/4281ef6b18e203896627de0f452b09b1e242a3f4/internal/problemspans/set.go#L51)):*

I think it is worth noting in a code comment somewhere that there is no explicit garbage collection, and that updates to regions will result in comparing with adjacent regions and expired regions will get merged.

Hmm, maybe I missed something -- does `regiontree.T.Enumerate` also mutate the tree, and could be doing some GC?
</blockquote></details>

Yes, Enumerate did remove unnecessary boundaries if it finds them. This guarantees that an unnecessary boundary will be visited at most once (either by an Update or by Enumerate).

I [changed](https://github.com/RaduBerinde/axisds/commit/ecb85bedf677ba87e0383e9b2ec5b5d88484ce8e) axisds to provide `WithGC` variants and added `Any / AnyWithGC` variants which make this more evident
___
*[`internal/problemspans/set.go` line 79 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4419#-OMrcNNx4dRq5TSlf-E6:-OMsGq6C36qypeyB5bqR:b1l4isg) ([raw file](https://github.com/cockroachdb/pebble/blob/4281ef6b18e203896627de0f452b09b1e242a3f4/internal/problemspans/set.go#L79)):*

can't this return something where `exp < s.now`?
</blockquote></details>

No, it returns only regions with non-zero property. Anyway, switched to the new `Any` API.
___
*[`internal/problemspans/set.go` line 30 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4419#-OMrdQu9C3brCij9RwuM:-OMsFQvTE6iVHkJK4aBR:bislb4m) ([raw file](https://github.com/cockroachdb/pebble/blob/13d034afaceec4c54b15d42fe6336ce28127d552/internal/problemspans/set.go#L30)):*

An alternative would be have coarse grained expiry times, so that adjacent regions are merged. I suppose there are two problems with that (a) there could be healthy and unhealthy regions alternating, (b) coarser grained expiry times also delay when a compaction will again be attempted, which may not be desirable.

Is this pathological case really a problem, given we are already tracking all files in the per-level btrees in the latest version? In the worst-case, if each file contributes a key span here, we would have the same number of entries here. And since we won't be doing compactions for those unexpired key spans, we won't have keyspans that only existed in old versions. So we are bounded by the size of the current version's spans.
</blockquote></details>

You're right. For some reason I didn't realize that because of our span the tables won't go away. Removing.

It was a bit of Jurassic Park - "Your scientists were so preoccupied with whether they could, they didn't stop to think if they should" :)

sumeerbhola (approved):
:lgtm: 

> Reviewed 3 of 17 files at r5, 14 of 14 files at r7, 11 of 11 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4419#-:-ONGuASt59RS3tDwg_6A:bg7igkv)* status: all files reviewed, 8 unresolved discussions (waiting on @jbowens)

___
*[`internal/manifest/l0_sublevels.go` line 1434 at r8](https://reviewable.io/reviews/cockroachdb/pebble/4419#-ONGjwyo750ZYUlbMYXC:-ONGjwyo750ZYUlbMYXD:b-vlgglb) ([raw file](https://github.com/cockroachdb/pebble/blob/edfad40a4ad7ce4f5c1f9a901260ca36882a6cee/internal/manifest/l0_sublevels.go#L1434)):*
> ```Go
> 	scoredIntervals := make([]intervalAndScore, 0, len(s.orderedIntervals))
> 	sublevelCount := len(s.levelFiles)
> 	for i := range s.orderedIntervals[:len(s.orderedIntervals)-1] {
> ```

how about a comment like
// The last orderedInterval does not contain any files and only provides the end key for the preceding interval, so we exclude it from this iteration.
___
*[`compaction_picker.go` line 551 at r8](https://reviewable.io/reviews/cockroachdb/pebble/4419#-ONGq5iVCzg24FuEOL5c:-ONGq5iVCzg24FuEOL5d:bzen76p) ([raw file](https://github.com/cockroachdb/pebble/blob/edfad40a4ad7ce4f5c1f9a901260ca36882a6cee/compaction_picker.go#L551)):*
> ```Go
> 	}
> 	// We need to include the outputLevel iter because without it, in a multiLevel scenario,
> 	// sm1 and la1 could shift the output level keyspace when pc.outputLevel.files is set to grow1.
> ```

(unrelated to this PR)
I don't understand this comment about multi-level compactions.
Also, I am uncertain whether the code here is correct for multi-level compactions. If we have levels:
```
L1:  [c, d)  [e, g)
L2:   [c, d) [f, g)
L3:  [c, f)                [h, k)  
```
Then if we started with the following compaction
```
L1:  [c, d) 
L2:   [c, d)
L3:  [c, f)
```
We would include [e, g) from L1, but won't include [f, g) from L2, which would break the level invariant.

Trying to (re)read the compaction expanding logic, it seems to have devolved into a mess when it should be fairly straightforward from an algorithmic perspective.

Thoughts?

@jbowens 
___
*[`compaction_picker.go` line 1825 at r8](https://reviewable.io/reviews/cockroachdb/pebble/4419#-ONGmV8I5ev22t1j2mrA:-ONGmV8I5ev22t1j2mrB:b-ws7fuw) ([raw file](https://github.com/cockroachdb/pebble/blob/edfad40a4ad7ce4f5c1f9a901260ca36882a6cee/compaction_picker.go#L1825)):*
> ```Go
> 	// 1, since choosing a single sublevel intra-L0 compaction is
> 	// counterproductive.
> 	lcf = l0Organizer.PickIntraL0Compaction(env.earliestUnflushedSeqNum, minIntraL0Count)
> ```

Why does `PickIntraL0Compaction` not use `problemSpans` to filter? Presumably we can ingest external files into L0 when restoring many layers from backup, so external files could end up in L0, yes?
___
*[`compaction_test.go` line 2847 at r8](https://reviewable.io/reviews/cockroachdb/pebble/4419#-ONGl4Rc-nnrRXOYzE31:-ONGl4Rc-nnrRXOYzE32:bhpm87k) ([raw file](https://github.com/cockroachdb/pebble/blob/edfad40a4ad7ce4f5c1f9a901260ca36882a6cee/compaction_test.go#L2847)):*
> ```Go
> }
> 
> func TestCompactionCorruption(t *testing.T) {
> ```

nice test

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4419#-:-ONQ9zsL7FA0GAs6DyRw:b-814g8k)* status: all files reviewed, 8 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`compaction_picker.go` line 551 at r8](https://reviewable.io/reviews/cockroachdb/pebble/4419#-ONGq5iVCzg24FuEOL5c:-ONQAgnhBKlozMpSVbvu:b-sa34oy) ([raw file](https://github.com/cockroachdb/pebble/blob/edfad40a4ad7ce4f5c1f9a901260ca36882a6cee/compaction_picker.go#L551)):*

(unrelated to this PR)
I don't understand this comment about multi-level compactions.
Also, I am uncertain whether the code here is correct for multi-level compactions. If we have levels:
```
L1:  [c, d)  [e, g)
L2:   [c, d) [f, g)
L3:  [c, f)                [h, k)  
```
Then if we started with the following compaction
```
L1:  [c, d) 
L2:   [c, d)
L3:  [c, f)
```
We would include [e, g) from L1, but won't include [f, g) from L2, which would break the level invariant.

Trying to (re)read the compaction expanding logic, it seems to have devolved into a mess when it should be fairly straightforward from an algorithmic perspective.

Thoughts?

@jbowens
</blockquote></details>

Filed https://github.com/cockroachdb/pebble/issues/4511
___
*[`compaction_picker.go` line 1825 at r8](https://reviewable.io/reviews/cockroachdb/pebble/4419#-ONGmV8I5ev22t1j2mrA:-ONQ9eOg4VS1zx_wJUAY:bt2vr1b) ([raw file](https://github.com/cockroachdb/pebble/blob/edfad40a4ad7ce4f5c1f9a901260ca36882a6cee/compaction_picker.go#L1825)):*

Why does `PickIntraL0Compaction` not use `problemSpans` to filter? Presumably we can ingest external files into L0 when restoring many layers from backup, so external files could end up in L0, yes?
</blockquote></details>

Fixed.
___
*[`internal/manifest/l0_sublevels.go` line 1434 at r8](https://reviewable.io/reviews/cockroachdb/pebble/4419#-ONGjwyo750ZYUlbMYXC:-ONQ9e3n31qIsD8LozhY:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/edfad40a4ad7ce4f5c1f9a901260ca36882a6cee/internal/manifest/l0_sublevels.go#L1434)):*

how about a comment like
// The last orderedInterval does not contain any files and only provides the end key for the preceding interval, so we exclude it from this iteration.
</blockquote></details>

Done.

PR #4422, opened 2025-03-25
db: fix TestCheckpointManyFiles

### LINKED ISSUE #4421: nightly testrace failed
The nightly testrace run failed on 96302028f139b8e2b38b910b4e45daf16d86ae04. Please review the run [here](https://github.com/cockroachdb/pebble/actions/runs/14053641693).

sumeerbhola:
It was incorrectly using the same compaction scheduler across two DB open calls.

Fixes #4421

### DIFF
--- a/compaction_scheduler.go
+++ b/compaction_scheduler.go
@@ -237,7 +237,8 @@ func (c *pickedCompactionCache) add(pc *pickedCompaction) {
 // ConcurrencyLimitScheduler is the default scheduler used by Pebble. It
 // simply uses the concurrency limit retrieved from
 // DBForCompaction.GetAllowedWithoutPermission to decide the number of
-// compactions to schedule.
+// compactions to schedule. ConcurrencyLimitScheduler must have its Register
+// method called at most once -- i.e., it cannot be reused across DBs.
 //
 // Since the GetAllowedWithoutPermission value changes over time, the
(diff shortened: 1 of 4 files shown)


sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4422)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @xxmplus)

___
*[`checkpoint_test.go` line 467 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4422#-OMDg-6YDxi-gtR064nA:-OMNJxMECgn4zmvE9IrC:bl8lja6) ([raw file](https://github.com/cockroachdb/pebble/blob/724bf87a01e447490ae1a2924ecc7b2931e4e3ba/checkpoint_test.go#L467)):*

If this is the case, `EnsureDefaults` shouldn't set it.. We should have a separate field in `DB` and do `newConcurrencyLimitScheduler` in `Open` if the one in opts is nil.

There are likely other tests that reopen with the same options.
</blockquote></details>

Good point. Fixed.
___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4422#-OMNJqJS0w4aC4zaEnoB:-OMNJqJT6mk37sehaUUn:b-814g8k) (no related file):*
TFTR!

PR #4427, opened 2025-03-26
db: prohibit ingestion or external iteration of sstables with blob refs

jbowens:
This commit adds validation to NewExternalIter and the ingestion pathways prohibiting the external iteration or ingestion of sstables containing blob references.

This could conceivably be relaxed in the future if we provide a mechanism to provide the referenced blob files too, but for now this is an error condition.

Additionally the most recent experimental format major version that introduces the TableFormatPebblev6 is renamed to capture the fact that it most significantly will enable use of blob value separation (in addition to the introduction of a checksum covering sstable footer data).

### DIFF
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -1225,8 +1225,8 @@ func TestManualCompaction(t *testing.T) {
 		},
 		{
 			testData:   "testdata/manual_compaction_set_with_del_sstable_Pebblev6",
-			minVersion: formatChecksumFooter,
-			maxVersion: formatChecksumFooter,
+			minVersion: formatTableFormatV6,
+			maxVersion: formatTableFormatV6,
 		},
 	}
 
--- a/data_test.go
+++ b/data_test.go
@@ -18,7 +18,9 @@ import (
 	"strings"
 	"testing"
 	"time"
+	"unicode"
 
+	"github.com/cockroachdb/crlib/crstrings"
 	"github.com/cockroachdb/datadriven"
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/bloom"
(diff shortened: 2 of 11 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4427#-:-OMIxiUm39HsbucTVPlZ:bnfp4nl)* status: 0 of 11 files reviewed, 1 unresolved discussion

___
*[`format_major_version.go` line 226 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4427#-OMIxVcG0nyzpq-VpBOo:-OMIxVcG0nyzpq-VpBOp:b5r1z5s) ([raw file](https://github.com/cockroachdb/pebble/blob/8edbb88bc27c3de41cbfa84caf3c63faca18601c/format_major_version.go#L226)):*
> ```Go
> 	//
> 	// This format major version is unexported until value separation is stable.
> 	formatBlobValueSeparation
> ```

I wonder if we should keep these separate. We could add just the footer checksum in 25.2?

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4427#-:-OMOTOZO4OaSbMEXWi9m:b-814g8k)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`format_major_version.go` line 226 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4427#-OMIxVcG0nyzpq-VpBOo:-OMOTPb58-LDEaP7CJkF:bmmy0v4) ([raw file](https://github.com/cockroachdb/pebble/blob/8edbb88bc27c3de41cbfa84caf3c63faca18601c/format_major_version.go#L226)):*

I wonder if we should keep these separate. We could add just the footer checksum in 25.2?
</blockquote></details>

for now let me make this format major version what enables the new table format only; the new table format supports encoding handles but that'll be unused until the subsequent format major version

RaduBerinde:
*[`format_major_version.go` line 226 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4427#-OMIxVcG0nyzpq-VpBOo:-OMOU6st3AbFbqa0pUQU:b4c9lyr) ([raw file](https://github.com/cockroachdb/pebble/blob/8edbb88bc27c3de41cbfa84caf3c63faca18601c/format_major_version.go#L226)):*

for now let me make this format major version what enables the new table format only; the new table format supports encoding handles but that'll be unused until the subsequent format major version
</blockquote></details>

:thumbsup:

PR #4444, opened 2025-03-28
sstable: use correct footer size in EstimatedSize

jbowens:
Use the correct footer size in (*RawColumnarWriter).EstimatedSize, and refactor some of the footer size accesses to use a new FooterSize method on the TableFormat type.


RaduBerinde (approved):

jbowens:
TFTR!

PR #4452, opened 2025-03-29
manifest: don't crosscheck AddL0Files result every time

RaduBerinde:
In invariants.Enabled, we rebuild the L0 sublevels from scratch to
cross-check the result of the incremental updating. However, this
rebuilding reinitializes the fields inside TableMetadata (like
L0Index), correcting any mistake made by `AddL0Files` (but only in
invariant builds..)

This change adds an `invariant.Sometimes(10)` to only do this
sometimes.

In the future, we will want to move out the TableMetadata fields we're
modifying and cross-check those values too.


RaduBerinde:
TFTR!

PR #4454, opened 2025-03-30
manifest: simplify BulkVersionEdit.Apply

RaduBerinde:
Require a non-nil version to simplify the interface and the code a
bit.

We also remove unnecessary nil checks for `Version.L0Sublevels`, which
is never the case.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -549,9 +549,7 @@ func newFlush(
 		}
 	}
 
-	if cur.L0Sublevels != nil {
-		c.l0Limits = cur.L0Sublevels.FlushSplitKeys()
-	}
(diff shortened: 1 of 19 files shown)


jbowens (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4454)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4454#-OMgljhX9dec2HxBO4hf:-OMgljhYASO__0s2lV_y:b-uwkowd) (no related file):*
:lgtm_strong:

RaduBerinde:
TFTR!

PR #4456, opened 2025-03-31
manifest: add L0Organizer, remove L0Sublevels from Version

RaduBerinde:
This set of changes consolidates the logic for the construction or incremental updating of `L0Sublevels` into a top-level object that maintains the current L0 state, to be maintained alongside the latest version.

#### manifest: add L0Organizer

We introduce `L0Organizer` type which contains the logic for updating
`L0Sublevels`. A single `L0Organizer` per store keeps track of the
current L0 state (corresponding to the latest `Version`).

#### manifest: remove L0Sublevels from Version

The `L0Sublevels` field of `Version` is very peculiar: it is only
usable for the most recent version.

This commit removes this field and switches all usages to
`L0Organizer`. `L0Sublevels` is now unexported and is only an
implementation detail of `L0Organizer`. In time we will rewrite the
functionality of `L0Sublevels` into `L0Organizer` directly.


sumeerbhola (changes requested):
> Reviewed 1 of 22 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4456)* status: 0 of 31 files reviewed, 5 unresolved discussions (waiting on @jbowens)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMiFQzQ9O_Or13WgqUQ:-OMiFQzQ9O_Or13WgqUR:bfrd9am) (no related file):*
(still reading) flushing some comments before the end of day
___
*[`internal/manifest/l0_sublevels.go` line 2080 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMiAwG9B5O_eZlNVRHq:-OMiAwG9B5O_eZlNVRHr:b-ie98tj) ([raw file](https://github.com/cockroachdb/pebble/blob/ffae41d250c4006d223d9aee82f72f47243ea337/internal/manifest/l0_sublevels.go#L2080)):*
> ```Go
> // sublevels.
> //
> // It is designed to be used as a singleton (per store) which gets updated as
> ```

nit: s/store/DB/

(since store is CockroachDB terminology)
___
*[`internal/manifest/l0_sublevels.go` line 2146 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMiD09i1BzA_tE5kpC-:-OMiD09i1BzA_tE5kpC0:bleck56) ([raw file](https://github.com/cockroachdb/pebble/blob/ffae41d250c4006d223d9aee82f72f47243ea337/internal/manifest/l0_sublevels.go#L2146)):*
> ```Go
> 		files := make([]*TableMetadata, 0, len(addedL0Tables))
> 		iter := newLevelMeta.Iter()
> 		for t := iter.Last(); len(files) < len(addedL0Tables); t = iter.Prev() {
> ```

why do we need to iterate over everything in L0 given that we already have `addedL0Tables`? Can't we initialize the slice using the map and then do `SortBySeqNum(addedTables)` like on master? We use to have situations (rarer now) with 1000s of existing files in L0 and a flush adding a handful of files.
___
*[`internal/manifest/l0_sublevels.go` line 2152 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMiCyEs8kyz7Fz28VG8:-OMiCyEs8kyz7Fz28VG9:b-b4c5ou) ([raw file](https://github.com/cockroachdb/pebble/blob/ffae41d250c4006d223d9aee82f72f47243ea337/internal/manifest/l0_sublevels.go#L2152)):*
> ```Go
> 			files = append(files, t)
> 		}
> 		if len(files) == len(addedL0Tables) {
> ```

should this condition be an assertion?
___
*[`internal/manifest/l0_sublevels.go` line 2154 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMiDt2rAdPE1pVhX1gI:-OMiDt2rAdPE1pVhX1gJ:b-7duttt) ([raw file](https://github.com/cockroachdb/pebble/blob/ffae41d250c4006d223d9aee82f72f47243ea337/internal/manifest/l0_sublevels.go#L2154)):*
> ```Go
> 		if len(files) == len(addedL0Tables) {
> 			slices.Reverse(files)
> 			newSublevels, err := o.sublevels.AddL0Files(files, o.flushSplitBytes, newLevelMeta)
> ```

On master we also verify when `invariants.Enabled` that the incremental construction is equivalent to the non-incremental construction. This has caught bugs in the past in the former. That verification seems to be missing now.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4456)* status: 0 of 31 files reviewed, 5 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMiFQzQ9O_Or13WgqUQ:-OMiKHT17Nn7SBpk1Y2c:b-yz40g0) (no related file):*

(still reading) flushing some comments before the end of day
</blockquote></details>

Thanks for reviewing!
___
*[`internal/manifest/l0_sublevels.go` line 2146 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMiD09i1BzA_tE5kpC-:-OMiJ54k2fL0FZi1IaCw:b-pi6cb) ([raw file](https://github.com/cockroachdb/pebble/blob/ffae41d250c4006d223d9aee82f72f47243ea337/internal/manifest/l0_sublevels.go#L2146)):*

why do we need to iterate over everything in L0 given that we already have `addedL0Tables`? Can't we initialize the slice using the map and then do `SortBySeqNum(addedTables)` like on master? We use to have situations (rarer now) with 1000s of existing files in L0 and a flush adding a handful of files.
</blockquote></details>

Usually the new files sort last in the level metadata, but that is not always the case (seqnum stack vs simple stack). We are basically checking which of this cases it is. We are only iterating `len(addedL0Tables)` times.
___
*[`internal/manifest/l0_sublevels.go` line 2154 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMiDt2rAdPE1pVhX1gI:-OMiKJAM80WNfYTvcCvX:b4de5h5) ([raw file](https://github.com/cockroachdb/pebble/blob/ffae41d250c4006d223d9aee82f72f47243ea337/internal/manifest/l0_sublevels.go#L2154)):*

On master we also verify when `invariants.Enabled` that the incremental construction is equivalent to the non-incremental construction. This has caught bugs in the past in the former. That verification seems to be missing now.
</blockquote></details>

Ah, will add back, I did intend to keep it.

RaduBerinde:
*[`internal/manifest/l0_sublevels.go` line 2154 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMiDt2rAdPE1pVhX1gI:-OMiMj1eD-ZJeZg4wqd4:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/ffae41d250c4006d223d9aee82f72f47243ea337/internal/manifest/l0_sublevels.go#L2154)):*

Ah, will add back, I did intend to keep it.
</blockquote></details>

Done.

RaduBerinde:
*[`internal/manifest/l0_sublevels.go` line 2146 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMiD09i1BzA_tE5kpC-:-OMj7q6mCgpgWkHFLaaY:bpirie1) ([raw file](https://github.com/cockroachdb/pebble/blob/ffae41d250c4006d223d9aee82f72f47243ea337/internal/manifest/l0_sublevels.go#L2146)):*

Usually the new files sort last in the level metadata, but that is not always the case (seqnum stack vs simple stack). We are basically checking which of this cases it is. We are only iterating `len(addedL0Tables)` times.
</blockquote></details>

I added a commit that addresses the TODO and moves this into `addL0Files` which already has the same check. It's now more clear that we only look at `len(addedL0Tables)` files.

sumeerbhola (changes requested):
> Reviewed 6 of 22 files at r1, 8 of 25 files at r3, 23 of 25 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4456)* status: all files reviewed, 16 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/manifest/l0_sublevels.go` line 2146 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMiD09i1BzA_tE5kpC-:-OMmByMV0a_d6Wn6TW1u:b-gdz22t) ([raw file](https://github.com/cockroachdb/pebble/blob/ffae41d250c4006d223d9aee82f72f47243ea337/internal/manifest/l0_sublevels.go#L2146)):*
> We are only iterating `len(addedL0Tables)` times.

Oh, I missed that.
___
*[`internal/manifest/l0_sublevels.go` line 2082 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmEEdS3UDm_QNAGC1-:-OMmEEdS3UDm_QNAGC10:b-yft0pr) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/internal/manifest/l0_sublevels.go#L2082)):*
> ```Go
> //
> // The level 0 sstables are organized in a series of sublevels. Similar to the
> // seqnum invariant in normal levels, there is no internal key in a higher level
> ```

nit: In Pebble we consistently use the term "higher level" to represent the visually higher level, and not the numbering. That is, L0 is a higher level compared to L1.
There are other systems which use the term based on the numbering.

Based on Pebble's visual interpretation, this comment seems false.
___
*[`internal/manifest/l0_sublevels.go` line 2086 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmEsiXCaQk9iiivgCd:-OMmEsiXCaQk9iiivgCe:b-g54p2d) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/internal/manifest/l0_sublevels.go#L2086)):*
> ```Go
> // sublevel, tables are sorted by their internal key range and any two tables at
> // the same sublevel do not overlap. Unlike the normal levels, sublevel n
> // contains older tables (lower sequence numbers) than sublevel n+1.
> ```

Perhaps clarify that we do this numbering from the bottom since the number of sublevels is variable.
___
*[`internal/manifest/l0_sublevels.go` line 2127 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmc5V333ymtu-bNXLP:-OMmc5V333ymtu-bNXLQ:b-4474bl) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/internal/manifest/l0_sublevels.go#L2127)):*
> ```Go
> 
> // SublevelFiles returns the sublevels as LevelSlices.
> func (o *L0Organizer) SublevelFiles() []LevelSlice {
> ```

I assume the caller assume the return value is immutable, since this may be the thing initializing `Version.L0SublevelFiles` that is needed for reads of any version? If yes, can you add a comment.
___
*[`internal/manifest/l0_sublevels.go` line 2182 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmdTTO3TEFWKjjE1du:-OMmdTTO3TEFWKjjE1dv:b-ljuwsx) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/internal/manifest/l0_sublevels.go#L2182)):*
> ```Go
> 
> // Reset the L0Organizer to reflect a given L0 level. Used for testing.
> func (o *L0Organizer) Reset(levelMetadata *LevelMetadata) {
> ```

nit: can you rename this `ResetForTesting`.
___
*[`compaction_picker.go` line 361 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmgUUz-G6o1SwOn7YJ:-OMmgUUz-G6o1SwOn7YK:bvwbi7m) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/compaction_picker.go#L361)):*
> ```Go
> 		pickerMetrics: pc.pickerMetrics,
> 
> 		// Both copies see the same manifest, therefore, it's ok for them to se
> ```

nit: them to share ...
___
*[`compaction_picker.go` line 677 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmhcm83E_6PFi7pTHA:-OMmhcm83E_6PFi7pTHB:bfqrj53) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/compaction_picker.go#L677)):*
> ```Go
> 	opts            *Options
> 	vers            *version
> 	l0Organizer     *manifest.L0Organizer
> ```

How about a comment like.
// Unlike `vers`, which is immutable and the latest version when this picker is created, l0Organizer represents the mutable L0 state of the latest version. This means that at some point in the future a `compactionPickerByScore` created in the past will have mutually inconsistent state in `vers` and `l0Organizer`. This is not a problem since (a) a new picker is created in `logAndApply` when a new version is installed, and (b) only the latest picker is used for picking compactions. This is ensured by holding `versionSet.logLock` for both (a) and (b).
___
*[`replay/replay.go` line 582 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmmrnw2ZLLRg9kLTcA:-OMmmrnw2ZLLRg9kLTcB:b-svcifn) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/replay/replay.go#L582)):*
> ```Go
> 	kind stepKind
> 	ve   manifest.VersionEdit
> 	// readAmp estimation for the LSM *before* the workload was corrected.
> ```

s/corrected/collected/
___
*[`replay/replay.go` line 808 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmnCsY-ZpoUUyULI5P:-OMmnCsY-ZpoUUyULI5Q:b-2hxcud) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/replay/replay.go#L808)):*
> ```Go
> 					}
> 				}
> 				if previousVersion != nil {
> ```

// previousVersion contains the current version, and so l0Organizer is consistent with it. The subsequent call to currentVersion will replace it with a new current version.
___
*[`replay/replay.go` line 819 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmoZp81xIil3h767ft:-OMmoZp9DRb3sjqK_zxk:b-aobyk) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/replay/replay.go#L819)):*
> ```Go
> 				}
> 				s.l0Organizer = l0Organizer
> 				// On the first time through, we set the previous version to the current
> ```

... we set the previousReadAmp to the current version's read-amp.

The "otherwise we set ..." reads a bit odd, since the otherwise happened a few lines before. Consider adding a backwards reference in the comment.
___
*[`replay/replay.go` line 824 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmp3J_1gyCsEHhgBSY:-OMmp3J_1gyCsEHhgBSZ:b-8m1mu0) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/replay/replay.go#L824)):*
> ```Go
> 					s.previousReadAmp = computeReadAmp(s.v, l0Organizer)
> 				}
> 				previousVersion = s.v
> ```

btw, I don't think `s.v` is ever used outside this loop, and within this loop it isn't needed either.
___
*[`internal/manifest/l0_sublevels.go` line 2159 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMm98vvAQGpLt9IHocb:-OMm98vvAQGpLt9IHocc:b-jcgffs) ([raw file](https://github.com/cockroachdb/pebble/blob/75c49a06ab2fde42ec8ede9072eac263c4e0c25e/internal/manifest/l0_sublevels.go#L2159)):*
> ```Go
> 				// AddL0Files did the right thing. Note that NewL0Sublevels updates
> 				// fields in TableMetadata like L0Index, so we don't want to do this
> 				// every time.
> ```

I don't quite understand the point about "fields in TableMetadata". Is this related to the other comment in `TableMetadata`?:
```
	// For L0 files only. Protected by DB.mu. Used to generate L0 sublevels and
	// pick L0 compactions. Only accurate for the most recent Version.
	// TODO(radu): this is very hacky and fragile. This information should live
	// inside L0Sublevels.
```

- We have an incremental path because the non-incremental construction is costly and was slowing down version application when L0 was huge. I forget whether it was linear or more from a big-O perspective, and if the former the constant may have been high. I don't think it mattered that some fields were getting updated in `TableMetadata`.
- We seem to have poor or incorrect code documentation on what mutex is held when calling the sublevels construction methods. e.g. the following seems incorrect.
```
// This method can be called without DB.mu being held, so any DB.mu protected
// fields in TableMetadata cannot be accessed here, such as Compacting and
// IsIntraL0Compacting. Those fields are accessed in InitCompactingFileInfo
// instead.
```
- Related to the TODO, could you elaborate on what you had in mind. The reason these fields were placed in `FileMetadata` is that they are convenient to access when constructing compactions since we don't need another lookup to find this information for a file that is under consideration. Were you thinking of having a map keyed by `FileNum` in `L0Sublevels` to do a lookup? btw, I don't think we ever measured whether optimizing away this lookup by placing these fields in `FileMetadata` was actually necessary -- I think this was how the prototype was written and we inherited it in the production implementation.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4456)* status: 26 of 31 files reviewed, 12 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMiFQzQ9O_Or13WgqUQ:-OMmzJNBCoIDrTR4LCOw:bq4e6e4) (no related file):*

Thanks for reviewing!
</blockquote></details>

Thanks for the thorough review, it helps a lot!
___
*[`compaction_picker.go` line 677 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmhcm83E_6PFi7pTHA:-OMmyR6sCygbm6y9uIXe:b-8lqcvy) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/compaction_picker.go#L677)):*

How about a comment like.
// Unlike `vers`, which is immutable and the latest version when this picker is created, l0Organizer represents the mutable L0 state of the latest version. This means that at some point in the future a `compactionPickerByScore` created in the past will have mutually inconsistent state in `vers` and `l0Organizer`. This is not a problem since (a) a new picker is created in `logAndApply` when a new version is installed, and (b) only the latest picker is used for picking compactions. This is ensured by holding `versionSet.logLock` for both (a) and (b).
</blockquote></details>

Done. Is there a good reason why we create a new `compactionPickerByScore` each time and put the version inside? Instead of passing the version to it whenever we run it? It feels like it should be another "top-level" object like `L0Organizer`.
___
*[`internal/manifest/l0_sublevels.go` line 2159 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMm98vvAQGpLt9IHocb:-OMmuPbp-TFLBfbmY2Qi:b-g1gba2) ([raw file](https://github.com/cockroachdb/pebble/blob/75c49a06ab2fde42ec8ede9072eac263c4e0c25e/internal/manifest/l0_sublevels.go#L2159)):*

I don't quite understand the point about "fields in TableMetadata". Is this related to the other comment in `TableMetadata`?:
```
	// For L0 files only. Protected by DB.mu. Used to generate L0 sublevels and
	// pick L0 compactions. Only accurate for the most recent Version.
	// TODO(radu): this is very hacky and fragile. This information should live
	// inside L0Sublevels.
```

- We have an incremental path because the non-incremental construction is costly and was slowing down version application when L0 was huge. I forget whether it was linear or more from a big-O perspective, and if the former the constant may have been high. I don't think it mattered that some fields were getting updated in `TableMetadata`.
- We seem to have poor or incorrect code documentation on what mutex is held when calling the sublevels construction methods. e.g. the following seems incorrect.
```
// This method can be called without DB.mu being held, so any DB.mu protected
// fields in TableMetadata cannot be accessed here, such as Compacting and
// IsIntraL0Compacting. Those fields are accessed in InitCompactingFileInfo
// instead.
```
- Related to the TODO, could you elaborate on what you had in mind. The reason these fields were placed in `FileMetadata` is that they are convenient to access when constructing compactions since we don't need another lookup to find this information for a file that is under consideration. Were you thinking of having a map keyed by `FileNum` in `L0Sublevels` to do a lookup? btw, I don't think we ever measured whether optimizing away this lookup by placing these fields in `FileMetadata` was actually necessary -- I think this was how the prototype was written and we inherited it in the production implementation.
</blockquote></details>

I have some ideas on how to rework all this but they need to be fleshed out more. I need to study the code paths that use these fields. At first glance I doubt that using a `map[*TableMetadata]L0RelatedState` would be a performance problem.

IMO some of these fields are misguided in the first place. `L0Index` is something that necessitates O(N) to update on each L0 compaction and AFAICT the only usecase for that is to represent the files in an L0 compaction using a `[]bool` of the same length as L0 instead of a `map[*TableMetadata]struct{}` (the latter sounds more efficient to me).
___
*[`internal/manifest/l0_sublevels.go` line 2086 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmEsiXCaQk9iiivgCd:-OMmtzuo3keURwpit4sB:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/internal/manifest/l0_sublevels.go#L2086)):*

Perhaps clarify that we do this numbering from the bottom since the number of sublevels is variable.
</blockquote></details>

Done.
___
*[`internal/manifest/l0_sublevels.go` line 2127 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmc5V333ymtu-bNXLP:-OMmuGFk8tUuXe2MwRfZ:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/internal/manifest/l0_sublevels.go#L2127)):*

I assume the caller assume the return value is immutable, since this may be the thing initializing `Version.L0SublevelFiles` that is needed for reads of any version? If yes, can you add a comment.
</blockquote></details>

Done.
___
*[`internal/manifest/l0_sublevels.go` line 2182 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmdTTO3TEFWKjjE1du:-OMn-0lm-afR1Kvk9fne:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/internal/manifest/l0_sublevels.go#L2182)):*

nit: can you rename this `ResetForTesting`.
</blockquote></details>

Done.
___
*[`replay/replay.go` line 582 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmmrnw2ZLLRg9kLTcA:-OMmxIA4DbRzbiniXHHe:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/replay/replay.go#L582)):*

s/corrected/collected/
</blockquote></details>

Done.
___
*[`replay/replay.go` line 808 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmnCsY-ZpoUUyULI5P:-OMmxMrNAzh0EoJbTcVk:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/replay/replay.go#L808)):*

// previousVersion contains the current version, and so l0Organizer is consistent with it. The subsequent call to currentVersion will replace it with a new current version.
</blockquote></details>

Done.
___
*[`replay/replay.go` line 819 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmoZp81xIil3h767ft:-OMmxYZU7kzeGCmHwvgs:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/replay/replay.go#L819)):*

... we set the previousReadAmp to the current version's read-amp.

The "otherwise we set ..." reads a bit odd, since the otherwise happened a few lines before. Consider adding a backwards reference in the comment.
</blockquote></details>

Done.
___
*[`replay/replay.go` line 824 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmp3J_1gyCsEHhgBSY:-OMmxmbO2vj34KkvHfxy:bdgobio) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/replay/replay.go#L824)):*

btw, I don't think `s.v` is ever used outside this loop, and within this loop it isn't needed either.
</blockquote></details>

I can try to clean this up for a bit. But can you clarify what's the difference between:
```
 // Version describing the state of the LSM *before* the workload was
// collected.
```
and
```
// a Version describing the state of the LSM when the workload was
// collected.
```
?

Should the latter read *after* the workload was collected? And should they be talking about a "workload step" rather than "the workload"?

sumeerbhola (approved):
> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4456)* status: all files reviewed, 12 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMiFQzQ9O_Or13WgqUQ:-OMo9G4O5koB08ZPTI8P:bg7igkv) (no related file):*

Thanks for the thorough review, it helps a lot!
</blockquote></details>

:lgtm: 
___
*[`compaction_picker.go` line 677 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmhcm83E_6PFi7pTHA:-OMo7RKG7YmHslOyrjeH:bpwzu7b) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/compaction_picker.go#L677)):*

Done. Is there a good reason why we create a new `compactionPickerByScore` each time and put the version inside? Instead of passing the version to it whenever we run it? It feels like it should be another "top-level" object like `L0Organizer`.
</blockquote></details>

It is already a top-level object and sits in the `versionSet`, except that we replace it with a new object every time. I am not sure it is worth wrapping it to create the illusion that it isn't being replaced, but actually replace a pointer inside the top-level object.
___
*[`replay/replay.go` line 824 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmp3J_1gyCsEHhgBSY:-OMo8aof8WjaVuY_bVbI:b-cvx0hr) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/replay/replay.go#L824)):*

I can try to clean this up for a bit. But can you clarify what's the difference between:
```
 // Version describing the state of the LSM *before* the workload was
// collected.
```
and
```
// a Version describing the state of the LSM when the workload was
// collected.
```
?

Should the latter read *after* the workload was collected? And should they be talking about a "workload step" rather than "the workload"?
</blockquote></details>

Yes, I think it should say workload step (though I've never reviewed this code before so my understanding is not strong).

Also, feel free to leave the cleanup for a later time since `workloadStep.v` was effectively unused even before this PR.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4456)* status: 6 of 31 files reviewed, 12 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`replay/replay.go` line 824 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMmp3J_1gyCsEHhgBSY:-OMoXURc1jbMk1wwg8z9:b-dm43p4) ([raw file](https://github.com/cockroachdb/pebble/blob/0a857da9c03b8e86e006bc57c5ff222d1ee627cd/replay/replay.go#L824)):*

Yes, I think it should say workload step (though I've never reviewed this code before so my understanding is not strong).

Also, feel free to leave the cleanup for a later time since `workloadStep.v` was effectively unused even before this PR.
</blockquote></details>

Cleaned up a bit.

sumeerbhola (approved):
> Reviewed 2 of 25 files at r8, 2 of 5 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4456)* status: 10 of 31 files reviewed, 12 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[a discussion](https://reviewable.io/reviews/cockroachdb/pebble/4456#-OMiFQzQ9O_Or13WgqUQ:-OMrIk0C-h0CScFVyG4G:bg7igkv) (no related file):*

:lgtm:
</blockquote></details>

:lgtm:

RaduBerinde:
TFTR!

PR #4457, opened 2025-03-31
manifest: add proof sketch for L0 sublevels

sumeerbhola:


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4457#-:-OMiIvkr0dib5ev2fnLG:bnfp4nl)* status: 0 of 1 files reviewed, 2 unresolved discussions

___
*[`internal/manifest/doc.go` line 21 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4457#-OMiGqYO86FUSp5NOwo_:-OMiGqYO86FUSp5NOwoa:b-fxk5qj) ([raw file](https://github.com/cockroachdb/pebble/blob/d6e7a52ef4d9465b7ad12e397b3ac78a7cf7487c/internal/manifest/doc.go#L21)):*
> ```Go
> into a stack of sstables:
> 
> - Simple-stack: Files are simply stacked into levels based on their add
> ```

[nit] I find the discussion around "simple stack" (which is something we're not using) distracting. I wonder if things would be simpler if we just focus on proving what we need to prove for the seqnum-stack directly. Just a note for the future, I know this PR is just for checking in what we already have.
___
*[`internal/manifest/doc.go` line 84 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4457#-OMiHaW83efMpKMTZuJi:-OMiHaW83efMpKMTZuJj:bhdt21i) ([raw file](https://github.com/cockroachdb/pebble/blob/d6e7a52ef4d9465b7ad12e397b3ac78a7cf7487c/internal/manifest/doc.go#L84)):*
> ```Go
>     - S2 is from memtable(s) that flushed because the ingest was requested.
>       Seqnum of k in S2 < Seqnum of k in S1. Contradiction.
> T
> ```

[nit] extra `T`

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4457#-:-OMmrmidDWqpF95pamO9:b-814g8k)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/manifest/doc.go` line 21 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4457#-OMiGqYO86FUSp5NOwo_:-OMmqN1uDj5TeOMyV_bN:b-knm4fc) ([raw file](https://github.com/cockroachdb/pebble/blob/d6e7a52ef4d9465b7ad12e397b3ac78a7cf7487c/internal/manifest/doc.go#L21)):*

[nit] I find the discussion around "simple stack" (which is something we're not using) distracting. I wonder if things would be simpler if we just focus on proving what we need to prove for the seqnum-stack directly. Just a note for the future, I know this PR is just for checking in what we already have.
</blockquote></details>

The simple stack was considered because it is (a) considered the vanilla mode of operation (though even pre L0-sublevels we didn't quite operate like that), and (b) it admits the behaviors that can happen due to files getting compacted out of L0 (in the extreme, if every file added to L0 got compacted out of L0 to lower levels before the next file was added, then we still need the level invariant of the LSM as a whole to be satisfied). I had been hoping to use (b) as a way to prove something about "A real LSM behaves as a hybrid of these two stacks, since there are compactions out from L0 at arbitrary points in time."
___
*[`internal/manifest/doc.go` line 84 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4457#-OMiHaW83efMpKMTZuJi:-OMmrQlE1Km5Sza3H9QO:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/d6e7a52ef4d9465b7ad12e397b3ac78a7cf7487c/internal/manifest/doc.go#L84)):*

[nit] extra `T`
</blockquote></details>

Done

PR #4460, opened 2025-03-31
db: fix plumbing of CompactionGrantHandle

sumeerbhola:
CompactionGrantHandle now includes the CPUMeasurer interface, that has separate method for each kind of goroutine, instead of identifying the goroutine kind by an index. This includes the secondary goroutine for writing to blob files. This interface is plumbed to RawColumnWriter and blob.FileWriter, so that the secondary writing goroutines can report CPU after writing each block.


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4460)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`internal/base/compaction_grant_handle.go` line 23 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4460#-OMiFtSJ26LGJXbTz-I7:-OMiFtSJ26LGJXbTz-I8:b-2sk9li) ([raw file](https://github.com/cockroachdb/pebble/blob/ef975c7d99f6db5eb9fae0288d35acc879c24e95/internal/base/compaction_grant_handle.go#L23)):*
> ```Go
> 	// CumulativeStats.
> 	Started()
> 	CPUMeasurer
> ```

[nit] move to the top
___
*[`internal/base/compaction_grant_handle.go` line 45 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4460#-OMiFzalDQJQUeAqs7Yf:-OMiFzalDQJQUeAqs7Yg:b-wli7mc) ([raw file](https://github.com/cockroachdb/pebble/blob/ef975c7d99f6db5eb9fae0288d35acc879c24e95/internal/base/compaction_grant_handle.go#L45)):*
> ```Go
> // certain kind of goroutine, it can skip calling that particular MeasureCPU*
> // method.
> type CPUMeasurer interface {
> ```

I don't quite understand how this is supposed to work.. What is `MeasureCPUPrimary()` expected to do exactly? Does it add the amount of CPU consumed by the same goroutine since the last `MeasureCPUPrimary()` call? If so, wouldn't a `PrimaryStarted() / PrimaryStopped()` pair make more sense?

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4460#-:-OMmvjyQ0U8Nla5g6Bn_:b-814g8k)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/base/compaction_grant_handle.go` line 23 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4460#-OMiFtSJ26LGJXbTz-I7:-OMmvPwsEZnmnK6K4XZr:b6k8ekq) ([raw file](https://github.com/cockroachdb/pebble/blob/ef975c7d99f6db5eb9fae0288d35acc879c24e95/internal/base/compaction_grant_handle.go#L23)):*

[nit] move to the top
</blockquote></details>

See response below.
___
*[`internal/base/compaction_grant_handle.go` line 45 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4460#-OMiFzalDQJQUeAqs7Yf:-OMmu2cZ8q1MQ368CDqP:b-ohpuoe) ([raw file](https://github.com/cockroachdb/pebble/blob/ef975c7d99f6db5eb9fae0288d35acc879c24e95/internal/base/compaction_grant_handle.go#L45)):*

I don't quite understand how this is supposed to work.. What is `MeasureCPUPrimary()` expected to do exactly? Does it add the amount of CPU consumed by the same goroutine since the last `MeasureCPUPrimary()` call? If so, wouldn't a `PrimaryStarted() / PrimaryStopped()` pair make more sense?
</blockquote></details>

These will end up calling into https://github.com/cockroachdb/cockroach/blob/435f3332e75c89fa5ec27b981edf92110ee8eb32/pkg/util/admission/admit_long/measure.go#L102 and there is a `goroutineCPUMeasurer` per goroutine, where the first call to `goroutineCPUMeasurer.measure` is used to initialize state https://github.com/cockroachdb/cockroach/blob/435f3332e75c89fa5ec27b981edf92110ee8eb32/pkg/util/admission/admit_long/measure.go#L181.
So yes, each call to `Measure*` will for its goroutine consider measuring the cpu time since the last call.
We could add an explicit `Started` and `Stopped` pair for each kind of measurement but then we will have 9 methods instead of 3 and it seemed like software over-engineering.

And the measure* calls don't have to be sandwiched between `CompactionGrantHandle.{Started,Done}` in the current implementation of `CompactionGrantHandle` but conceptually that seemed a cleaner way to state the interface. Which is also why I ended up placing the `CPUMeasurer` after `Started` in the `CompactionGrantHandle` interface.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4460#-:-OMn0Dqr0OUFnxvT-szC:bnfp4nl)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/base/compaction_grant_handle.go` line 45 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4460#-OMiFzalDQJQUeAqs7Yf:-OMn-vdm9gvw-ZAbpB31:b-naxns4) ([raw file](https://github.com/cockroachdb/pebble/blob/ef975c7d99f6db5eb9fae0288d35acc879c24e95/internal/base/compaction_grant_handle.go#L45)):*

These will end up calling into https://github.com/cockroachdb/cockroach/blob/435f3332e75c89fa5ec27b981edf92110ee8eb32/pkg/util/admission/admit_long/measure.go#L102 and there is a `goroutineCPUMeasurer` per goroutine, where the first call to `goroutineCPUMeasurer.measure` is used to initialize state https://github.com/cockroachdb/cockroach/blob/435f3332e75c89fa5ec27b981edf92110ee8eb32/pkg/util/admission/admit_long/measure.go#L181.
So yes, each call to `Measure*` will for its goroutine consider measuring the cpu time since the last call.
We could add an explicit `Started` and `Stopped` pair for each kind of measurement but then we will have 9 methods instead of 3 and it seemed like software over-engineering.

And the measure* calls don't have to be sandwiched between `CompactionGrantHandle.{Started,Done}` in the current implementation of `CompactionGrantHandle` but conceptually that seemed a cleaner way to state the interface. Which is also why I ended up placing the `CPUMeasurer` after `Started` in the `CompactionGrantHandle` interface.
</blockquote></details>

So this means that only the first and the last call to `Measure*` matter (for a given compaction)?

IMO a Start/Stop and using an enum with three values (Primary/SSTableWriter/BlobWrtier) as the argument would be a lot cleaner. Or just a Start that returns a Closer.

Just a suggestion though, keeping it as-is is OK but clarifying these mechanics would be good, maybe something along the lines:

// MeasureCPUPrimary is used to measure CPU usage of the primary goroutine that does the work in the compaction. Every call to this method adds the usage of the goroutine since the last call; in essence the resulting measurement is the CPU used by the goroutine between the first and last call.

I think the verb `Measure` is also kind of confusing since the method doesn't by itself measure anything.. Some ideas: `UpdateCPU`, `TrackCPU`, `CheckpointCPU`, `CPUPulse`

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4460)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/base/compaction_grant_handle.go` line 45 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4460#-OMiFzalDQJQUeAqs7Yf:-OMo3tME7Xlz2W4YsPIL:bro9ty8) ([raw file](https://github.com/cockroachdb/pebble/blob/ef975c7d99f6db5eb9fae0288d35acc879c24e95/internal/base/compaction_grant_handle.go#L45)):*
> So this means that only the first and the last call to `Measure*` matter (for a given compaction)?

The granter continually measures the rate at which a compaction is consuming CPU, both for current accounting, and for future prediction. So no, all the intermediate calls also matter.

> IMO a Start/Stop and using an enum with three values (Primary/SSTableWriter/BlobWrtier) as the argument would be a lot cleaner.

I'll switch to using an enum.

> I think the verb `Measure` is also kind of confusing since the method doesn't by itself measure anything.. Some ideas: `UpdateCPU`, `TrackCPU`, `CheckpointCPU`, `CPUPulse`

The callee is measuring something. But it doesn't return the measurement. So agreed, this isn't a very good name.

I am wary of using anything with the word Checkpoint in the Pebble context. UpdateCPU suggests we are providing an updated value to the callee. I could go with TrackCPU, though between MeasureCPU and TrackCPU, it is hard for me to pick.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4460)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/base/compaction_grant_handle.go` line 45 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4460#-OMiFzalDQJQUeAqs7Yf:-OMoBw3XElLTxpsiSTHv:b-6vpc0u) ([raw file](https://github.com/cockroachdb/pebble/blob/ef975c7d99f6db5eb9fae0288d35acc879c24e95/internal/base/compaction_grant_handle.go#L45)):*
> The granter continually measures the rate at which a compaction is consuming CPU, both for current accounting, and for future prediction. So no, all the intermediate calls also matter.

Oh, I completely missed this aspect, it makes much more sense now. Yeah "track" and "measure" are pretty much the same. Maybe the description can be something like

// MeasureCPU allows the granter to keep track of CPU usage while a compaction is ongoing. It is to be called periodically from the compaction goroutine corresponding to the argument.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4460)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`internal/base/compaction_grant_handle.go` line 45 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4460#-OMiFzalDQJQUeAqs7Yf:-OMrGM5CBCNorYNNC_6r:bwesohj) ([raw file](https://github.com/cockroachdb/pebble/blob/ef975c7d99f6db5eb9fae0288d35acc879c24e95/internal/base/compaction_grant_handle.go#L45)):*

> The granter continually measures the rate at which a compaction is consuming CPU, both for current accounting, and for future prediction. So no, all the intermediate calls also matter.

Oh, I completely missed this aspect, it makes much more sense now. Yeah "track" and "measure" are pretty much the same. Maybe the description can be something like

// MeasureCPU allows the granter to keep track of CPU usage while a compaction is ongoing. It is to be called periodically from the compaction goroutine corresponding to the argument.
</blockquote></details>

Done, including switching to an enum.

RaduBerinde (commented):
:lgtm_strong: Thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4460#-:-OMrQOUfCbaTJaz9VXii:bfdlkw5)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @jbowens)

Issue #4394, opened 2025-03-18
db: add retrieval of blob file values in iterators, compactions


jbowens:
Retrieval of blob file values is not yet integrated into iterators or compactions. Both should initialize a `blob.ValueFetcher` and plumb it through.

Informs #112.

Jira issue: PEBBLE-354

PR #4461, opened 2025-03-31
db: add support for reading blob values

jbowens:
Initialize iterators and compactions with a blob.ValueFetcher that's propagated
through to sstable iterators. When an sstable iterator encounters a blob value
handle, it constructs an internal value with the blob.ValueFetcher configured
as the ValueFetcher.

Close https://github.com/cockroachdb/pebble/issues/4394.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -29,6 +29,7 @@ import (
 	"github.com/cockroachdb/pebble/objstorage/objstorageprovider/objiotracing"
 	"github.com/cockroachdb/pebble/objstorage/remote"
 	"github.com/cockroachdb/pebble/sstable"
+	"github.com/cockroachdb/pebble/sstable/blob"
 	"github.com/cockroachdb/pebble/sstable/block"
 	"github.com/cockroachdb/pebble/vfs"
 )
@@ -216,6 +217,9 @@ type compaction struct {
 	// blob files. This consumes more write bandwidth because all values are
 	// rewritten. However it restores locality.
 	getValueSeparation func(JobID, *compaction, sstable.TableFormat) compact.ValueSeparation
+	// valueFetcher is used to fetch values from blob files. It's propagated
+	// down the iterator tree through the internal iterator options.
+	valueFetcher blob.ValueFetcher
 
 	// startLevel is the level that is being compacted. Inputs from startLevel
 	// and outputLevel will be merged to produce a set of outputLevel files.
@@ -3085,16 +3089,19 @@ func (d *DB) compactAndWrite(
 	// translate to 3 MiB per compaction.
 	c.bufferPool.Init(12)
 	defer c.bufferPool.Release()
+	env := block.ReadEnv{
+		BufferPool: &c.bufferPool,
+		Stats:      &c.stats,
+		IterStats: d.fileCache.SSTStatsCollector().Accumulator(
+			uint64(uintptr(unsafe.Pointer(c))),
+			categoryCompaction,
+		),
+	}
+	c.valueFetcher.Init(d.fileCache, env)
 	iiopts := internalIterOpts{
-		compaction: true,
-		readEnv: block.ReadEnv{
-			BufferPool: &c.bufferPool,
-			Stats:      &c.stats,
-			IterStats: d.fileCache.SSTStatsCollector().Accumulator(
-				uint64(uintptr(unsafe.Pointer(c))),
-				categoryCompaction,
-			),
-		},
+		compaction:       true,
+		readEnv:          env,
+		blobValueFetcher: &c.valueFetcher,
 	}
 
 	pointIter, rangeDelIter, rangeKeyIter, err := c.newInputIters(d.newIters, d.tableNewRangeKeyIter, iiopts)
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -610,7 +610,7 @@ func TestCompaction(t *testing.T) {
 					return "", "", errors.WithStack(err)
 				}
 				defer r.Close()
-				iter, err := r.NewIter(sstable.NoTransforms, nil /* lower */, nil /* upper */)
+				iter, err := r.NewIter(sstable.NoTransforms, nil /* lower */, nil /* upper */, sstable.AssertNoBlobHandles)
 				if err != nil {
 					return "", "", errors.WithStack(err)
 				}
--- a/db.go
+++ b/db.go
@@ -1349,6 +1349,9 @@ type internalIterOpts struct {
 	compaction         bool
 	readEnv            block.ReadEnv
 	boundLimitedFilter sstable.BoundLimitedBlockPropertyFilter
+	// blobValueFetcher is the base.ValueFetcher to use when constructing
+	// internal values to represent values stored externally in blob files.
+	blobValueFetcher base.ValueFetcher
 }
 
 func finishInitializingInternalIter(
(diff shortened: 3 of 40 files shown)


sumeerbhola (changes requested):
preliminary comments

> Reviewed 4 of 43 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4461#-:-OMro_Xm2h3UNXoT_xCi:bjzq3vn)* status: 4 of 43 files reviewed, 3 unresolved discussions

___
*[`sstable/blob/fetcher.go` line 125 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4461#-OMrlora8B9_kiYkEYl-:-OMrlora8B9_kiYkEYl0:b-zd8hft) ([raw file](https://github.com/cockroachdb/pebble/blob/8bd94b34f05d615a6d828816516f1513a5187778/sstable/blob/fetcher.go#L125)):*
> ```Go
> 
> 	if r.env.Stats != nil {
> 		r.env.Stats.SeparatedPointValue.ValueBytesFetched += uint64(vh.ValueLen)
> ```

Can you add a TODO in `InternalIteratorStats` that we should consider distinguishing between values fetched from value blocks in the same sstable, and those from blob files.
___
*[`db.go` line 1354 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4461#-OMrn_Ph7Xzs65JZLVv3:-OMrn_Ph7Xzs65JZLVv4:b61iseg) ([raw file](https://github.com/cockroachdb/pebble/blob/8bd94b34f05d615a6d828816516f1513a5187778/db.go#L1354)):*
> ```Go
> 	// blobValueFetcher is the base.ValueFetcher to use when constructing
> 	// internal values to represent values stored externally in blob files.
> 	blobValueFetcher base.ValueFetcher
> ```

So are we sharing the same `blob.ValueFetcher` across all levels? If yes, the locality implications of this make me nervous.

I think we'd discussed this in-person in the past, and considered two alternatives (a) having a fetcher per level, (b) increasing `maxCachedReaders`. (b) can still allow a level with terrible locality to thrash the cache and negatively impact other levels. And there are some linear scans related to cache eviction which become more problematic with increasing of `maxCachedReaders`.
___
*[`sstable/blob/blob.go` line 167 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4461#-OMrlZnTDoGD_sEbvqwb:-OMrlZnTDoGD_sEbvqwc:bt28rk3) ([raw file](https://github.com/cockroachdb/pebble/blob/8bd94b34f05d615a6d828816516f1513a5187778/sstable/blob/blob.go#L167)):*
> ```Go
> // It's exposed so that tests can force flushes to construct blob files with
> // arbitrary structures.
> func (w *FileWriter) Flush() {
> ```

nit: can we expose FlushForTesting, which calls flush.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4461)* status: 4 of 43 files reviewed, 3 unresolved discussions (waiting on @sumeerbhola)

___
*[`db.go` line 1354 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4461#-OMrn_Ph7Xzs65JZLVv3:-OMrvPt62sB9c8MvrCUV:bcm571b) ([raw file](https://github.com/cockroachdb/pebble/blob/8bd94b34f05d615a6d828816516f1513a5187778/db.go#L1354)):*

So are we sharing the same `blob.ValueFetcher` across all levels? If yes, the locality implications of this make me nervous.

I think we'd discussed this in-person in the past, and considered two alternatives (a) having a fetcher per level, (b) increasing `maxCachedReaders`. (b) can still allow a level with terrible locality to thrash the cache and negatively impact other levels. And there are some linear scans related to cache eviction which become more problematic with increasing of `maxCachedReaders`.
</blockquote></details>

I thought we talked about this and determined using a single `blob.ValueFetcher` is strictly better in terms of locality. A single fetcher is able to reuse a cached reader for a blob file between levels. During iteration, you actually care about the locality of the merged iterator stream. A value fetcher per level may succeed in avoiding misses within that level, but at the cost of more misses in the overall merged iterator stream than if you'd used that allocation of cached readers elsewhere.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4461)* status: 4 of 43 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`db.go` line 1354 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4461#-OMrn_Ph7Xzs65JZLVv3:-OMw4FPTBTvFrW_BzBar:b-k9ecwc) ([raw file](https://github.com/cockroachdb/pebble/blob/8bd94b34f05d615a6d828816516f1513a5187778/db.go#L1354)):*

I thought we talked about this and determined using a single `blob.ValueFetcher` is strictly better in terms of locality. A single fetcher is able to reuse a cached reader for a blob file between levels. During iteration, you actually care about the locality of the merged iterator stream. A value fetcher per level may succeed in avoiding misses within that level, but at the cost of more misses in the overall merged iterator stream than if you'd used that allocation of cached readers elsewhere.
</blockquote></details>

we discussed this offline, and agreed -- my memory was flawed.
The only remaining issue is that the fetcher should probably not have a fixed `maxCachedReaders` and should be configurable at construction time, based on the number of levels. And perhaps reconsider the linear scan over all the cached readers for finding the right one and evicting the oldest one, in `ValueFetcher.retrieve`.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4461)* status: 2 of 44 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`db.go` line 1354 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4461#-OMrn_Ph7Xzs65JZLVv3:-OMwsyz83u8AcQcgd0I3:bu6byak) ([raw file](https://github.com/cockroachdb/pebble/blob/8bd94b34f05d615a6d828816516f1513a5187778/db.go#L1354)):*

we discussed this offline, and agreed -- my memory was flawed.
The only remaining issue is that the fetcher should probably not have a fixed `maxCachedReaders` and should be configurable at construction time, based on the number of levels. And perhaps reconsider the linear scan over all the cached readers for finding the right one and evicting the oldest one, in `ValueFetcher.retrieve`.
</blockquote></details>

Added a TODO for the linear scan
___
*[`sstable/blob/blob.go` line 167 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4461#-OMrlZnTDoGD_sEbvqwb:-OMwqeYsBHNcITIs-hi9:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/8bd94b34f05d615a6d828816516f1513a5187778/sstable/blob/blob.go#L167)):*

nit: can we expose FlushForTesting, which calls flush.
</blockquote></details>

Done
___
*[`sstable/blob/fetcher.go` line 125 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4461#-OMrlora8B9_kiYkEYl-:-OMwqxwQ5niWa6F8JJqk:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/8bd94b34f05d615a6d828816516f1513a5187778/sstable/blob/fetcher.go#L125)):*

Can you add a TODO in `InternalIteratorStats` that we should consider distinguishing between values fetched from value blocks in the same sstable, and those from blob files.
</blockquote></details>

Done.

sumeerbhola (approved):
:lgtm: 

> Reviewed 28 of 43 files at r2, 8 of 11 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4461#-:-ON1cYRc3b-5VY5zNb7L:bg7igkv)* status: 39 of 44 files reviewed, 2 unresolved discussions

jbowens:
TFTR!

PR #4462, opened 2025-04-01
go.mod: update github.com/prometheus/client_golang

hazzik:
Updated github.com/prometheus/client_golang to minimal version that does not have a mirriad of unnecessary dependencies.

This commit is just a result of:

```sh
go get github.com/prometheus/client_golang@v1.15
go mod tidy
```


RaduBerinde:
This is an indirect dependency and I'm hesitant to downgrade it. Also, what Pebble version do you use? You'd need this on a `pebble-release-` branch.

hazzik:
Thanks for feedback.

> This is an indirect dependency

It is pretty much direct, updating `github.com/prometheus/client_golang` to 0.15.0 would produce almost the same diff, because `prometheus/client_golang` and `prometheus/common` have circular dependency on each other. 

I could redo with the upgrade of client_golang to 0.15.0 as I did initially, that would be a direct dependency upgrade. Would it be acceptable?

> Also, what Pebble version do you use? 

v1.1.2 -- the version used by go-ethereum. We have dependency on it, and I'm trying to cleanup things from bottom up.

> You'd need this on a `pebble-release- branch`.

Thanks for suggestion, I'll re-target to that branch.

RaduBerinde:
> I could redo with the upgrade of client_golang to 0.15.0 as I did initially, that would be a direct dependency upgrade. Would it be acceptable?

Oh, I see.  Yes, updating that is good. I can do that on the other branches separately, you'll care about the `pebble-release-1.1` branch.

hazzik:
Thanks. Done.

RaduBerinde:
Looks like CI isn't set up properly on this branch. I pushed a commit that should fix it, could you rebase? Thanks!

RaduBerinde:
I verified that the diff is exactly what I get when I do `go get github.com/prometheus/client_golang@v1.15.0; go mod tidy`.

RaduBerinde:
I tagged a v1.1.5 you can upgrade to. https://github.com/cockroachdb/pebble/releases/tag/v1.1.5

PR #4466, opened 2025-04-01
internal/manifest: add test for zero allocs when ranging

jbowens:
Add a unit test that asserts ranging over a LevelMetadata or LevelSlice through their All methods result in zero allocations.

### DIFF
--- a/internal/manifest/version_test.go
+++ b/internal/manifest/version_test.go
@@ -783,3 +783,52 @@ func TestCalculateInuseKeyRangesRandomized(t *testing.T) {
 		}
 	}
 }
+
+func TestIterAllocs(t *testing.T) {
+	t.Run("LevelSlice", func(t *testing.T) {
+		var testLevelSlice = func() LevelSlice {
+			const n = 10
+			var tables []*TableMetadata
+			for i := 0; i < n; i++ {
+				tables = append(tables, &TableMetadata{
+					FileNum:     base.FileNum(i),
+					FileBacking: &FileBacking{},
+				})
+			}
+			return NewLevelSliceSeqSorted(tables)
+		}()
+		allocs := testing.AllocsPerRun(100, func() {
+			for v := range testLevelSlice.All() {
+				if v == nil {
+					panic("nil")
+				}
+			}
+		})
+		if allocs > 0 {
+			t.Fatalf("allocs=%f", allocs)
+		}
+	})
+	t.Run("LevelMetadata", func(t *testing.T) {
+		var testLevelMetadata = func() LevelMetadata {
+			const n = 1000
+			var tables []*TableMetadata
+			for i := 0; i < n; i++ {
+				tables = append(tables, &TableMetadata{
+					FileNum:     base.FileNum(i),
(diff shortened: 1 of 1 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4466)* status: 0 of 1 files reviewed, 2 unresolved discussions

___
*[`internal/manifest/version_test.go` line 800 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4466#-OMmWojS3N3ARVoao7iw:-OMmWojS3N3ARVoao7ix:b-87j6m6) ([raw file](https://github.com/cockroachdb/pebble/blob/20d3ff9dfdb52d67b0761112937d3f9bb6102b1d/internal/manifest/version_test.go#L800)):*
> ```Go
> 			return NewLevelSliceSeqSorted(tables)
> 		}()
> 		allocs := testing.AllocsPerRun(1, func() {
> ```

Why only one run? It might be flaky if there's something in the background that might allocate something every now and then.
___
*[`internal/manifest/version_test.go` line 807 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4466#-OMmXaucDzZ0Ag55m7Nk:-OMmXaucDzZ0Ag55m7Nl:b-14bsp1) ([raw file](https://github.com/cockroachdb/pebble/blob/20d3ff9dfdb52d67b0761112937d3f9bb6102b1d/internal/manifest/version_test.go#L807)):*
> ```Go
> 			}
> 		})
> 		if allocs > 0.0001 {
> ```

Note that `Although the return value has type float64, it will always be an integral value.`  So you can probably just compare against 0.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4466)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @RaduBerinde)

___
*[`internal/manifest/version_test.go` line 800 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4466#-OMmWojS3N3ARVoao7iw:-OMqmX3_-3Zc2k_YBTve:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/20d3ff9dfdb52d67b0761112937d3f9bb6102b1d/internal/manifest/version_test.go#L800)):*

Why only one run? It might be flaky if there's something in the background that might allocate something every now and then.
</blockquote></details>

Done
___
*[`internal/manifest/version_test.go` line 807 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4466#-OMmXaucDzZ0Ag55m7Nk:-OMqmatX82osKNaN7EhR:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/20d3ff9dfdb52d67b0761112937d3f9bb6102b1d/internal/manifest/version_test.go#L807)):*

Note that `Although the return value has type float64, it will always be an integral value.`  So you can probably just compare against 0.
</blockquote></details>

Done

RaduBerinde (approved):

RaduBerinde (approved):

Issue #4417, opened 2025-03-24
db: excise should not require reading inside overlapping sstables


jbowens:
If an excise span partially overlaps a sstable, excise attempts to narrow the bounds of the resulting virtual sstable by reading inside the sstable. In the case where we're recovering from a corrupt or deleted sstable, the sstable may not exist and this read will fail. We should be able to tolerate excises that partially remove a table that we can't read (under the expectation that a future excise will remove the rest of the table).

```
F250324 16:01:34.280020 69287 kv/kvserver/store_raft.go:756  [T1,Vsystem,n3,s3,r77/3:{d-g},raft] 1  error while excising span: e{-z}: pebble: backing file 000039 error: stat /tmp/cockroach/_tmp/577bfc8ed46850d1aaafa923fa37ce
78/TestGeneralOperationsWorkAsExpectedOnDeletedExternalSST2532445940/001/external-files/non-Existent-File: no such file or directory
-- stack trace:
| github.com/cockroachdb/pebble.NewFileCache.func1
|         external/com_github_cockroachdb_pebble/file_cache.go:460
| github.com/cockroachdb/pebble/internal/genericcache.(*shard[...]).findOrCreateValue
|         external/com_github_cockroachdb_pebble/internal/genericcache/shard.go:191
| github.com/cockroachdb/pebble/internal/genericcache.(*Cache[...]).FindOrCreate
|         external/com_github_cockroachdb_pebble/internal/genericcache/cache.go:84
| github.com/cockroachdb/pebble.(*fileCacheHandle).findOrCreateTable
|         external/com_github_cockroachdb_pebble/file_cache.go:242
| github.com/cockroachdb/pebble.(*fileCacheHandle).newIters
|         external/com_github_cockroachdb_pebble/file_cache.go:545
| github.com/cockroachdb/pebble.(*DB).excise.func1
|         external/com_github_cockroachdb_pebble/excise.go:81
| github.com/cockroachdb/pebble.(*DB).excise
|         external/com_github_cockroachdb_pebble/excise.go:222
| github.com/cockroachdb/pebble.(*DB).ingestApply      
```

Jira issue: PEBBLE-359

jbowens:
cc @iskettaneh

RaduBerinde:
@jbowens @iskettaneh I am close to finishing this, but I'm looking for some guidance on the following:

1) Do we want to avoid opening the excised tables altogether or do we merely want to tolerate read errors and only in that case fall back to loose bounds?
[...]

PR #4468, opened 2025-04-02
db: file excise code cleanup

RaduBerinde:
This set of changes attempts to tame the function that excises a file which is several screens long.

This is in preparation for work on #4417.

#### db: find strict overlap with excise span

`Version.Overlaps()` returns all "indirect" overlaps in L0. This is
not what we want when we look for files to excise.

This change replaces `Overlaps()` with using `LevelSlice.Overlaps()`
on each sublevel/level.

#### db: require overlap in exciseTable

Now that we only call it when there is overlap, `exciseTable` asserts
that the table is overlapping.

#### db: extract VersionEdit logic from exciseTable

Add a separate function for applying an excise to a `VersionEdit`;
`exciseTable` is simplified to return the left and right tables.

#### db: extract determineExcisedTableSize from exciseTable

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1345,13 +1345,6 @@ func (d *DB) runIngestFlush(c *compaction) (*manifest.VersionEdit, error) {
 				if err != nil {
 					return nil, err
 				}
-
-				if _, ok := ve.DeletedTables[deletedFileEntry{
-					Level:   layer.Level(),
-					FileNum: m.FileNum,
-				}]; !ok {
-					panic("did not excise file that overlaps with the excise span")
-				}
 				replacedFiles[m.FileNum] = newFiles
 				updateLevelMetricsOnExcise(m, layer.Level(), newFiles)
 			}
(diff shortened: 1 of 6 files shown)


jbowens (approved):
:lgtm:

> Reviewed 5 of 5 files at r1, 4 of 4 files at r2, 3 of 3 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 4 of 4 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4468#-:-OMrUbuSAXEMe-VySo0F:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`excise.go` line 333 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4468#-OMr4ETZBFlrPG76Liba:-OMr4ETZBFlrPG76Libb:bc72pnb) ([raw file](https://github.com/cockroachdb/pebble/blob/882747917a2526cfbb7b8cf25eacbea610514810/excise.go#L333)):*
> ```Go
> ) error {
> 	if originalTable.HasPointKeys && cmp(originalTable.SmallestPointKey.UserKey, exciseSpanStart) < 0 {
> 		// This file will probably contain point keys.
> ```

will it _probably_, or _necessarily_? afaict it necessarily will contain keys in the point key space
___
*[`excise.go` line 366 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4468#-OMr4v6h5DfWn_lDujas:-OMr4v6iCYn15PcvWcrH:byfbwhm) ([raw file](https://github.com/cockroachdb/pebble/blob/882747917a2526cfbb7b8cf25eacbea610514810/excise.go#L366)):*
> ```Go
> 				lastRangeKey = slices.Clone(rkey.End)
> 			}
> 			lastRangeKeyKind := rkey.Keys[0].Kind()
> ```

This was the existing code, but I'm confused about the key kind here. We're using whatever the first key's kind is. The first key is actually the smallest of the keys. I guess it doesn't really matter because we're constructing the largest key with the maximal sequence number regardless, but it's weird.

It seems like we should we should use `InternalKeyKindRangeKeyMax`?

PR #4482, opened 2025-04-04
blob: use colblk encoding for index block

jbowens:
Adapt the blob file format to use a colblk format for the index block,
replacing the previous use of the valblk index block format. The new index
block consists of a single Uints column indicating the offset at which each
block begins. A final value indicates the offset immediately after the last
value block. During lookups, the block length is inferred as the difference
between consecutive blocks.

The new colblk format is more easily extensible and is smaller than the valblk
index format (due to eliding the block numbers and block lengths). It is also
self-describing, avoiding the need for the index block's handle to encode
additional data about the width of each row.

Informs #112.

This is not expected to be the final blob index block format: see #4483.


sumeerbhola (commented):
Do you have a small microbenchmark on how this compares with `DecodeBlockHandleFromIndex`?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4482#-:-ONFevVNC9RCla1Ellmd:b-xmoaox)* status: 0 of 16 files reviewed, all discussions resolved (waiting on @xxmplus)

jbowens (commented):
Not yet, are you concerned this column lookup is actually impactful to value lookup latencies? It seems unlikely

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4482#-:-ONG6ljZAHCrvzkP-1YO:bpxwq98)* status: 0 of 16 files reviewed, all discussions resolved (waiting on @xxmplus)

jbowens (commented):
I didn't bother writing a vary narrow benchmark since it's probably not useful; but I added one covering `ValueFetcher.retrieve` and verified it's a wash:

```
goos: darwin
goarch: arm64
pkg: github.com/cockroachdb/pebble/sstable/blob
cpu: Apple M1 Pro
ValueFetcherRetrieve/sequential-10   405.0n ± 0%   402.1n ± 0%  -0.72% (p=0.000 n=25)
ValueFetcherRetrieve/random-10       1.014µ ± 1%   1.004µ ± 1%  -0.99% (p=0.022 n=25)
geomean                              640.8n        635.4n       -0.85%
```

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4482#-:-ONHR__UEXEAoZFGD6Qj:b-2kymtu)* status: 0 of 17 files reviewed, all discussions resolved (waiting on @xxmplus)

jbowens (commented):
Bumping this

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4482#-:-OPBJgdW7OL4hs5WU1Pl:bovfojb)* status: 0 of 18 files reviewed, all discussions resolved (waiting on @xxmplus)

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4482#-:-OPkbWk00Pz6Nl6Adx8v:bnfp4nl)* status: 0 of 22 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola and @xxmplus)

___
*[`sstable/blob/blob.go` line 338 at r6](https://reviewable.io/reviews/cockroachdb/pebble/4482#-OPkahvO5OrvHzm4O72u:-OPkahvO5OrvHzm4O72v:bwkatz9) ([raw file](https://github.com/cockroachdb/pebble/blob/e8c16ea6679910e5a6454c48f9b725a1feee5675/sstable/blob/blob.go#L338)):*
> ```Go
> 
> func (f *fileFooter) encode(b []byte) {
> 	binary.LittleEndian.PutUint64(b[4:], f.indexHandle.Offset)
> ```

I'd add an invariants check for the size of `b`. It's not evident that it has to be exactly the size of the footer and not larger. OR the crc below should constrain it to the correct length, e.g. `b[4:22+len(fileMagic)]`

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4482#-:-OPp925Y-ClFUqnVHIiT:b-814g8k)* status: 0 of 22 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde, @sumeerbhola, and @xxmplus)

___
*[`sstable/blob/blob.go` line 338 at r6](https://reviewable.io/reviews/cockroachdb/pebble/4482#-OPkahvO5OrvHzm4O72u:-OPp8-DZ3LIxBL2mOAjX:b-48vilg) ([raw file](https://github.com/cockroachdb/pebble/blob/e8c16ea6679910e5a6454c48f9b725a1feee5675/sstable/blob/blob.go#L338)):*

I'd add an invariants check for the size of `b`. It's not evident that it has to be exactly the size of the footer and not larger. OR the crc below should constrain it to the correct length, e.g. `b[4:22+len(fileMagic)]`
</blockquote></details>

Good call, done.

Issue #11, opened 2018-08-03
perf: faster table/block cache


petermattis:
The existing block cache implementation uses the Clock-PRO replacement policy. The block cache is important for performance. The raw speed of accessing the block cache is an important component of read performance. Concurrency of the block cache directly affects concurrent read performance. There are three areas worth exploring here:

1. Should the block cache be placed outside of the visibility of the Go GC. This could be accomplished by implementing our own hash table on top of arena allocated memory. 
2. Improve the concurrency of the block cache, either via sharding or [lock-free](http://myui.github.io/publications/ICDE10_conf_full_409.pdf) approaches.
3. Use a better cache replacement policy such as [TinyLFU](https://github.com/dgryski/go-tinylfu). See also [Caffeine](https://github.com/ben-manes/caffeine), a high-performance Java caching library.

olekukonko:
I believe the `dgraph` team wants to implement a `caffeine` port https://blog.dgraph.io/post/caching-in-go/

petermattis:
Pebble, like RocksDB, contains two primary caches: the block cache and the table cache. The table cache (implemented by `tableCache`) caches `sstable.Readers`. Performance experiments are currently demonstrating lock contention in the `tableCache` under concurrent read workloads. The `tableCache` internally contains an LRU list, and manipulation of the list is the problematic bit that requires mutual exclusion. The [ristretto](https://github.com/dgraph-io/ristretto/blob/master/design.md) project is looking to use the same technique as Caffeine and [BP-Wrapper](https://drive.google.com/file/d/0B8oWjCpZGTn3YVhCc2dZSU5DNFBJRlBZa0s1STdaUWR5emxj/view): record page accesses in a lock-free (or thread-local) structure and then apply them in batch. Some hacky experimentation in this direction showed a 25% speedup in a concurrent scan workload.

For `tableCache.findNode`, we'd have a fast-path check for when the table is already in the cache that uses a `RWMutex` instead of a `Mutex`. Something like:

```
func (c *tableCache) findNode(meta *fileMetadata) *tableCacheNode {
	c.mu.RLock()
	if n := c.mu.nodes[meta.fileNum]; n != nil {
		atomic.AddInt32(&n.refCount, 1)
		c.mu.RUnlock()
		// TODO(peter): record access of `n`.
		return n
	}
	c.mu.RUnlock()

	c.mu.Lock()
	defer c.mu.Unlock()
	// Handle miss: we need to check c.mu.nodes again...
}
[...]

PR #4484, opened 2025-04-05
sstable: make code for creating ssts in tests reusable

RaduBerinde:
#### sstable: export functionality to parse SST contents

We export `ParseTestKVsAndSpans` which can parse multi-line SST
contents and `ParseTestSST` which also writes the contents to a
RawWriter.

#### sstable: standardize format argument

There are multiple datadriven arguments with different values for
specifying the table format. This change standardizes all of them to
use `table-format`.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -584,40 +584,21 @@ func runBuildRemoteCmd(td *datadriven.TestData, d *DB, storage remote.Storage) e
 	}
 	path := td.CmdArgs[0].String()
 
-	// Override table format, if provided.
-	tableFormat := d.TableFormat()
-	var blockSize int64
-	for _, cmdArg := range td.CmdArgs[1:] {
-		switch cmdArg.Key {
-		case "format":
-			switch cmdArg.Vals[0] {
-			case "pebblev1":
-				tableFormat = sstable.TableFormatPebblev1
-			case "pebblev2":
-				tableFormat = sstable.TableFormatPebblev2
-			case "pebblev3":
-				tableFormat = sstable.TableFormatPebblev3
-			case "pebblev4":
-				tableFormat = sstable.TableFormatPebblev4
-			default:
-				return errors.Errorf("unknown format string %s", cmdArg.Vals[0])
-			}
-		case "block-size":
-			var err error
-			blockSize, err = strconv.ParseInt(cmdArg.Vals[0], 10, 64)
-			if err != nil {
-				return errors.Wrap(err, td.Pos)
-			}
-		}
+	// Use TableFormatMax here and downgrade after, if necessary. This ensures
+	// that all fields are set.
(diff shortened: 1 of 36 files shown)


jbowens (approved):
:lgtm:

> Reviewed 19 of 19 files at r1, 9 of 9 files at r2, 4 of 4 files at r3, 19 of 19 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4484#-:-ONHHypf8eEd-mhqkS5-:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`sstable/test_utils.go` line 143 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4484#-ONHEZ6cEb0OuxJTVVis:-ONHE76D8w04jfahhYO4:bor5ygs) ([raw file](https://github.com/cockroachdb/pebble/blob/1478e701351780eb2932d03bc52f7511a8e66df3/sstable/test_utils.go#L143)):*
> ```Go
> //	Span: a-d:{(#11,RANGEKEYSET,@10,foo)}
> //	Span: g-l:{(#5,RANGEDEL)}
> //	Span: y-z:{(#12,RANGEKEYSET,@11,foo)}
> ```

There are tests outside the context of a _sstable_ that sometimes detail with streams of internal keys. We could consider generalizing the parsing further... Some things like `force-obsolete` are specific to the domain of building an sstable writer, but could be modeled as generic 'tags' or something.

RaduBerinde:
*[`sstable/test_utils.go` line 143 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4484#-ONHEZ6cEb0OuxJTVVis:-ONHSEHK-YbhOA8xmD7f:b-gj65hg) ([raw file](https://github.com/cockroachdb/pebble/blob/1478e701351780eb2932d03bc52f7511a8e66df3/sstable/test_utils.go#L143)):*

There are tests outside the context of a _sstable_ that sometimes detail with streams of internal keys. We could consider generalizing the parsing further... Some things like `force-obsolete` are specific to the domain of building an sstable writer, but could be modeled as generic 'tags' or something.
</blockquote></details>

What about prepending force-obsolete: to the key itself, and handle that for sstables?

RaduBerinde:
*[`sstable/test_utils.go` line 143 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4484#-ONHEZ6cEb0OuxJTVVis:-ONKXje_BVK0fJ7E1wLS:b-psxo9s) ([raw file](https://github.com/cockroachdb/pebble/blob/1478e701351780eb2932d03bc52f7511a8e66df3/sstable/test_utils.go#L143)):*

What about prepending force-obsolete: to the key itself, and handle that for sstables?
</blockquote></details>

Anyway, I will merge this and keep that in mind. I found a couple of other tests that do similar parsing but they also write them to SSTs so they can just use this.

PR #4491, opened 2025-04-07
db: add missing logUnlock in some excise error paths

RaduBerinde:


RaduBerinde:
We clearly need to improve testing of these error paths (which can definitely happen in practice since they involve I/O). Filed https://github.com/cockroachdb/pebble/issues/4492

sumeerbhola (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4491#-:-ONGhAlxAFHPw4V76FAG:b-uwkowd)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @jbowens)

___
*[`ingest.go` line 2104 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4491#-ONGgqjR6ElGJjFtISv3:-ONGgqjR6ElGJjFtISv4:b-qjiy6a) ([raw file](https://github.com/cockroachdb/pebble/blob/aa247de25ccc7c43f98e1cecf322f20243b83867/ingest.go#L2104)):*
> ```Go
> 				leftTable, rightTable, err := d.exciseTable(ctx, exciseSpan.UserKeyBounds(), m, layer.Level())
> 				if err != nil {
> 					d.mu.versions.logUnlock()
> ```

In some methods should we add a `defer d.mu.versions.checkLogUnlocked()` so that even if we miss some error path bugs in testing, we will crash instead of getting deadlocked in prod?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4491)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`ingest.go` line 2104 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4491#-ONGgqjR6ElGJjFtISv3:-ONGlWMO4eiCXwH6C1_9:b-ydif40) ([raw file](https://github.com/cockroachdb/pebble/blob/aa247de25ccc7c43f98e1cecf322f20243b83867/ingest.go#L2104)):*

In some methods should we add a `defer d.mu.versions.checkLogUnlocked()` so that even if we miss some error path bugs in testing, we will crash instead of getting deadlocked in prod?
</blockquote></details>

I'm working on a change where this can't happen at all (it's how I found these cases).

PR #4493, opened 2025-04-07
db: rework logAndApply

RaduBerinde:
This commit replaces `logAndApply` with an `UpdateVersionLocked`
function that takes a closure and performs the log "locking" around
that closure. This makes it impossible for a caller to mess up the
locking.

I plan to look at the remaining uses of logLock/logUnlock separately.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1100,10 +1100,10 @@ func (d *DB) clearCompactingState(c *compaction, rollback bool) {
 		// It is a bit peculiar that we are fiddling with th current version state
 		// in a separate critical section from when this version was installed.
 		// But this fiddling is necessary if the compaction failed. When the
-		// compaction succeeded, we've already done this in logAndApply, so this
-		// seems redundant. Anyway, we clear the pickedCompactionCache since we
+		// compaction succeeded, we've already done this in UpdateVersionLocked, so
+		// this seems redundant. Anyway, we clear the pickedCompactionCache since we
 		// may be able to pick a better compaction (though when this compaction
-		// succeeded we've also cleared the cache in logAndApply).
+		// succeeded we've also cleared the cache in UpdateVersionLocked).
 		defer d.mu.versions.logUnlockAndInvalidatePickedCompactionCache()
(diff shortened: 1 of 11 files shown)


jbowens (approved):
> Reviewed 6 of 11 files at r1, 3 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4493)* status: 9 of 11 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`version_set.go` line 454 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4493#-ONQGtKeAnahJopzaXJt:-ONQGtKeAnahJopzaXJu:b-x0g6gc) ([raw file](https://github.com/cockroachdb/pebble/blob/04f72ea9acd29c4dc5ece127d96d9c28088268ce/version_set.go#L454)):*
> ```Go
> func (vs *versionSet) UpdateVersionLocked(updateFn func() (versionUpdate, error)) error {
> 	vs.logLock()
> 	defer vs.logUnlockAndInvalidatePickedCompactionCache()
> ```

this is expanding the cases in which we will invalidate the picked compaction cache right? I guess it's just to error cases that are insignificant?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4493)* status: 9 of 11 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`version_set.go` line 454 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4493#-ONQGtKeAnahJopzaXJt:-ONQTCn9DzkbxybD_1W2:btqqu5l) ([raw file](https://github.com/cockroachdb/pebble/blob/04f72ea9acd29c4dc5ece127d96d9c28088268ce/version_set.go#L454)):*

this is expanding the cases in which we will invalidate the picked compaction cache right? I guess it's just to error cases that are insignificant?
</blockquote></details>

Correct.

PR #4494, opened 2025-04-07
db: refactor obsolete, zombie file tracking

jbowens:
Add a zombieObjects type for tracking the lifecycle of zombie objects. Today this is used for zombie sstable objects. With the introduction of blob files (#112) we'll also need to track the lifecycle of zombie blob file objects.

Additionally, refactor the lifecycle so that a file is considered either zombie or obsolete but not both. And avoid scanning the list of zombie objects on every call to Metrics, instead incrementally maintaining stats on the zombieObjects type.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2406,7 +2406,7 @@ func (d *DB) cleanupVersionEdit(ve *versionEdit) {
 		// Add this file to zombie tables as well, as the versionSet
 		// asserts on whether every obsolete file was at one point
 		// marked zombie.
-		d.mu.versions.zombieTables[of.DiskFileNum] = objectInfo{
+		d.mu.versions.zombieTables.Add(objectInfo{
 			fileInfo: fileInfo{
 				FileNum:  of.DiskFileNum,
 				FileSize: of.Size,
@@ -2417,7 +2417,7 @@ func (d *DB) cleanupVersionEdit(ve *versionEdit) {
 			// disaggregated storage; if this becomes the norm, we should do
 			// an objprovider lookup here.
 			isLocal: true,
-		}
+		})
 	}
 	d.mu.versions.addObsoleteLocked(obsoleteFiles)
 }
@@ -3025,13 +3025,7 @@ func (d *DB) runCompaction(
 			// Add this file to zombie tables as well, as the versionSet
 			// asserts on whether every obsolete file was at one point
 			// marked zombie.
-			d.mu.versions.zombieTables[backing.DiskFileNum] = objectInfo{
-				fileInfo: fileInfo{
-					FileNum:  backing.DiskFileNum,
-					FileSize: backing.Size,
-				},
-				isLocal: true,
-			}
+			d.mu.versions.zombieTables.AddMetadata(&result.Tables[i].ObjMeta, backing.Size)
 		}
 		d.mu.versions.addObsoleteLocked(obsoleteFiles)
 		d.mu.Unlock()
--- a/db.go
+++ b/db.go
@@ -1644,6 +1644,7 @@ func (d *DB) Close() error {
 	if err := d.closed.Load(); err != nil {
 		panic(err)
 	}
+	d.compactionSchedulers.Wait()
 	// Compactions can be asynchronously started by the CompactionScheduler
 	// calling d.Schedule. When this Unregister returns, we know that the
 	// CompactionScheduler will never again call a method on the DB. Note that
@@ -1753,7 +1754,6 @@ func (d *DB) Close() error {
 	}
 
 	d.mu.Unlock()
-	d.compactionSchedulers.Wait()
 
 	// Wait for all cleaning jobs to finish.
 	d.cleanupManager.Close()
@@ -1771,7 +1771,7 @@ func (d *DB) Close() error {
 
 	// As a sanity check, ensure that there are no zombie tables. A non-zero count
 	// hints at a reference count leak.
-	if ztbls := len(d.mu.versions.zombieTables); ztbls > 0 {
+	if ztbls := d.mu.versions.zombieTables.Count(); ztbls > 0 {
 		err = firstError(err, errors.Errorf("non-zero zombie file count: %d", ztbls))
 	}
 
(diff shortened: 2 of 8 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4494)* status: 0 of 8 files reviewed, 5 unresolved discussions (waiting on @annrpom)

___
*[`obsolete_files.go` line 592 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4494#-ONH1T8DEt9x3i6vzAKn:-ONH1T8DEt9x3i6vzAKo:bmhjvi6) ([raw file](https://github.com/cockroachdb/pebble/blob/adf2794adacf61ed1e6782980820e8836b5f8206/obsolete_files.go#L592)):*
> ```Go
> 	slices.SortFunc(a, cmpObsoleteFileNumbers)
> 	return slices.CompactFunc(a, func(a, b obsoleteFile) bool {
> 		return a.nonLogFile.fileNum == b.nonLogFile.fileNum
> ```

When a and b are both `fileTypeLog`, this will be true. If this method can't be called with log files, we should document that and add an invariants check.
___
*[`obsolete_files.go` line 597 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4494#-ONH1gpm5ucwcIOVuABC:-ONH1gpm5ucwcIOVuABD:bqa4f2c) ([raw file](https://github.com/cockroachdb/pebble/blob/adf2794adacf61ed1e6782980820e8836b5f8206/obsolete_files.go#L597)):*
> ```Go
> 
> func cmpObsoleteFileNumbers(a, b obsoleteFile) int {
> 	return cmp.Compare(a.nonLogFile.fileNum, b.nonLogFile.fileNum)
> ```

Similar concern here.
___
*[`obsolete_files.go` line 631 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4494#-ONH-uIx-3MUIOynqZqO:-ONH-uIy-dypnxx8M8iu:b-nz3pto) ([raw file](https://github.com/cockroachdb/pebble/blob/adf2794adacf61ed1e6782980820e8836b5f8206/obsolete_files.go#L631)):*
> ```Go
> type zombieObjects struct {
> 	objs      map[base.DiskFileNum]objectInfo
> 	sizeTotal uint64
> ```

[supernit] totalSize, localSize reads more naturally
___
*[`obsolete_files.go` line 665 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4494#-ONH19nj9n4vklFeC894:-ONH19nj9n4vklFeC895:bttu9ci) ([raw file](https://github.com/cockroachdb/pebble/blob/adf2794adacf61ed1e6782980820e8836b5f8206/obsolete_files.go#L665)):*
> ```Go
> // Remove removes an object from the set of zombie objects, returning the object
> // that was removed.
> func (z *zombieObjects) Remove(fileNum base.DiskFileNum) objectInfo {
> ```

[nit] Remove can incorrectly suggest that we are removing the zombie object. Maybe rename to `Extract`
___
*[`obsolete_files.go` line 690 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4494#-ONH04sg6JIFaxmm4OnL:-ONH04sg6JIFaxmm4OnM:bblmfnu) ([raw file](https://github.com/cockroachdb/pebble/blob/adf2794adacf61ed1e6782980820e8836b5f8206/obsolete_files.go#L690)):*
> ```Go
> // Sizes returns the size of all objects in the set, and the size of all local
> // objects in the set.
> func (z *zombieObjects) Sizes() (all, local uint64) {
> ```

[nit] the pairing is awkward, we should have two separate methods

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4494)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @annrpom and @RaduBerinde)

___
*[`obsolete_files.go` line 592 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4494#-ONH1T8DEt9x3i6vzAKn:-ONHCHA24KrpslKvu48w:bfz279i) ([raw file](https://github.com/cockroachdb/pebble/blob/adf2794adacf61ed1e6782980820e8836b5f8206/obsolete_files.go#L592)):*

When a and b are both `fileTypeLog`, this will be true. If this method can't be called with log files, we should document that and add an invariants check.
</blockquote></details>

Removed the wonky log file / non-log file separation within the `obsoleteFile` type
___
*[`obsolete_files.go` line 631 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4494#-ONH-uIx-3MUIOynqZqO:-ONH7w1_3d-stWvPCs4O:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/adf2794adacf61ed1e6782980820e8836b5f8206/obsolete_files.go#L631)):*

[supernit] totalSize, localSize reads more naturally
</blockquote></details>

Done
___
*[`obsolete_files.go` line 665 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4494#-ONH19nj9n4vklFeC894:-ONH7vX10SQ6qyo71EiT:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/adf2794adacf61ed1e6782980820e8836b5f8206/obsolete_files.go#L665)):*

[nit] Remove can incorrectly suggest that we are removing the zombie object. Maybe rename to `Extract`
</blockquote></details>

Done
___
*[`obsolete_files.go` line 690 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4494#-ONH04sg6JIFaxmm4OnL:-ONH7um44DFvxurABoTY:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/adf2794adacf61ed1e6782980820e8836b5f8206/obsolete_files.go#L690)):*

[nit] the pairing is awkward, we should have two separate methods
</blockquote></details>

Done

RaduBerinde (approved):

RaduBerinde (approved):

jbowens:
TFTR!

PR #4497, opened 2025-04-08
[WIP] sstable: add adaptive compression

EdwardX29:
Add adaptive compression with zstd and snappy. zstd copmression is probed with exponential backoff when compression ratios are not better than 50%.

### DIFF
--- a/db.go
+++ b/db.go
@@ -2078,6 +2078,7 @@ func (d *DB) Metrics() *Metrics {
 		metrics.Table.CompressedCountSnappy += int64(compressionTypes.snappy)
 		metrics.Table.CompressedCountZstd += int64(compressionTypes.zstd)
 		metrics.Table.CompressedCountMinlz += int64(compressionTypes.minlz)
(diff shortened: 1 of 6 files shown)


petermattis (commented):
Out of curiosity, are you specifying target compression numbers for these benchmarks? I think we'd want to have benchmarks with values which compress well with both Zstd and Snappy, values which compress well only with Zstd and not Snappy, and values that don't compress well at all.

The adaptive heuristic you're using seems to be prioritizing Zstd compression even if that compression is not significantly better than Snappy. This might be a pessimization if Zstd is consuming more CPU than Snappy. Concretely, I can imagine data that gets better than 50% compression with both Zstd and Snappy (i.e. highly compressible data), but uses less CPU to compress using Snappy.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4497#-:-ONRATarA7qGdlY6RYCS:b90yd1r)* status: 0 of 6 files reviewed, all discussions resolved

Issue #4425, opened 2025-03-26
sstable: convert remaining rowblk blocks to colblk


jbowens:
There are a few sstable block types that still use the rowblk format (eg, the properties block, meta index block). If we convert these to use colblk, we'll eventually be able to remove support for rowblk altogether and delete the code.

Jira issue: PEBBLE-364

PR #4502, opened 2025-04-08
colblk: use key value columnar block for properties

annrpom:
This patch uses the new, drop-in columnar block to replace the row format of properties blocks.

Fixes: #4425


annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4502)* status: 7 of 29 files reviewed, 2 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/properties.go` line 313 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4502#-ONQhH-GD_efJeEaKB5D:-ONQu1YODPpSqAn2OOow:b-amfdnj) ([raw file](https://github.com/cockroachdb/pebble/blob/9a066d243c4e6a29b28097dbbd906de82a173c8d/sstable/properties.go#L313)):*

nit: a thought: we could try adding an `All() iter.Seq2[[]byte, []byte]` method to both the `colblk.KeyValueBlockDecoder` _and_ the  `rowblk.RawIter`, and then we can have one `load` method that takes an `iter.Seq2[[]byte, []byte]` and ranges over it. There's not a not lot of code duplication, so I think it's good either way.

If we keep the separate load funcs, my preference is to name this one `load` and the old one `loadFromRowBlock` or the like.
</blockquote></details>

done to changing names so we can get this in v6; always down to try adding `All() iter.Seq2[[]byte, []byte]` later on :eyes: 
___
*[`sstable/properties.go` line 483 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4502#-ONQi_Lb4gL2lXnIGbxt:-ONQq1M72-XKLrebhpSR:b-g9nb7u) ([raw file](https://github.com/cockroachdb/pebble/blob/9a066d243c4e6a29b28097dbbd906de82a173c8d/sstable/properties.go#L483)):*

nit: now that we're on Go 1.23 and we're in this part of the code, we could replace the above for loop with

```
keys := slices.Collect(maps.Keys(m))
```
</blockquote></details>

done oh cool! 
___
*[`sstable/properties.go` line 502 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4502#-ONQioSJBdWQwhcXIpwV:-ONQrENT-mZ1BMl6u10g:b-fncaun) ([raw file](https://github.com/cockroachdb/pebble/blob/9a066d243c4e6a29b28097dbbd906de82a173c8d/sstable/properties.go#L502)):*

I think this `[]byte(...)` cast will end up allocating, but we can avoid it with unsafe:

```
// Zero-length keys are unsupported. See below about StringData.
if len(key) == 0 {
    continue
}
// Use an unsafe conversion to avoid allocating. AddKV is not
// supposed to modify the given slice, so the unsafe conversion
// is okay. Note that unsafe.StringData panics if len(key) == 0,
// so we explicitly skipped zero-length keys above. They shouldn't
// occur in practice.
w.AddKV(unsafe.Slice(unsafe.StringData(key), len(key)), m[key])
```
</blockquote></details>

done; ty
___
*[`testdata/compaction_delete_only_hints` line 545 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4502#-ONQkpZMAwa-7Wq1yfdr:-ONQrg36EIicb1bFkbKE:bjf5duu) ([raw file](https://github.com/cockroachdb/pebble/blob/9a066d243c4e6a29b28097dbbd906de82a173c8d/testdata/compaction_delete_only_hints#L545)):*
done

> But here it's translating to disk space usage because apparently we don't apply ordinary compression to properties blocks either

ah ok i see that makes sense -- `writeNamedBlock` now takes in a Compression and the properties block (gated by `TableFormatPebblev6`) will use whatever compression is set in the `layoutWriter`

jbowens (approved):
:lgtm:

> Reviewed 4 of 29 files at r1, 23 of 23 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4502#-:-ONRTYyV09RifQAlUWZm:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

annrpom:
TFTR! ('-')7

PR #4503, opened 2025-04-08
rowblk: raise max block size to 1<<31; error rather than panic

jbowens:
Relax the maximum block size to reclaim 7-unused bits within the restart offsets. This provides additional leeway for massive blocks we're sometimes forced to construct due to very large key-value pairs. Additionally, this commit turns this check into an error that we propagate up and out of the sstable Writer rather than panicking with an assertion failure. In practice this means a DB that hits this limit may fail a compaction without bringing down the node. The DB may repeat the compaction in a busy loop, but it's possible for the DB to make some forward progress in the meantime.

Future work will look at adapting sstable splitting to split preemptively to avoid the maximum.

### DIFF
--- a/sstable/options.go
+++ b/sstable/options.go
@@ -15,8 +15,8 @@ import (
 )
 
 const (
-	// MaximumBlockSize is the maximum permissible size of a block.
-	MaximumBlockSize = rowblk.MaximumSize
+	// MaximumRestartOffset is the maximum permissible size of a block.
+	MaximumRestartOffset = rowblk.MaximumRestartOffset
 	// DefaultNumDeletionsThreshold defines the minimum number of point
 	// tombstones that must be present in a data block for it to be
 	// considered tombstone-dense.
--- a/cockroachkvs/rowblk_bench_test.go
+++ b/cockroachkvs/rowblk_bench_test.go
@@ -42,8 +42,11 @@ func benchmarkCockroachDataRowBlockWriter(b *testing.B, keyConfig KeyGenConfig, 
 			if j > 0 {
 				samePrefix = bytes.Equal(keys[j], keys[j-1])
 			}
-			w.AddWithOptionalValuePrefix(
+			err := w.AddWithOptionalValuePrefix(
 				ik, false, values[j], prevKeyLen, true, block.InPlaceValuePrefix(samePrefix), samePrefix)
+			if err != nil {
+				b.Fatalf("error adding key: %v", err)
+			}
 			j++
 			prevKeyLen = len(ik.UserKey)
 		}
@@ -75,8 +78,11 @@ func benchmarkCockroachDataRowBlockIter(b *testing.B, keyConfig KeyGenConfig, va
 		if count > 0 {
 			samePrefix = bytes.Equal(keys[count], keys[count-1])
 		}
-		w.AddWithOptionalValuePrefix(
+		err := w.AddWithOptionalValuePrefix(
 			ik, false, values[count], prevKeyLen, true, block.InPlaceValuePrefix(samePrefix), samePrefix)
+		if err != nil {
+			b.Fatalf("error adding key: %v", err)
+		}
 		count++
 		prevKeyLen = len(ik.UserKey)
 	}
--- a/options.go
+++ b/options.go
@@ -450,8 +450,8 @@ func (o *LevelOptions) EnsureDefaults() {
 	}
 	if o.BlockSize <= 0 {
 		o.BlockSize = base.DefaultBlockSize
-	} else if o.BlockSize > sstable.MaximumBlockSize {
-		panic(errors.Errorf("BlockSize %d exceeds MaximumBlockSize", o.BlockSize))
+	} else if o.BlockSize > sstable.MaximumRestartOffset {
+		panic(errors.Errorf("BlockSize %d exceeds MaximumRestartOffset", o.BlockSize))
 	}
 	if o.BlockSizeThreshold <= 0 {
 		o.BlockSizeThreshold = base.DefaultBlockSizeThreshold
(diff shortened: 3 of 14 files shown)


RaduBerinde (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4503)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`sstable/rowblk/rowblk_writer.go` line 24 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4503#-ONLLtg2DjVFlfdsHzhy:-ONLLtg30pD46Pi5IJrS:brah7dk) ([raw file](https://github.com/cockroachdb/pebble/blob/9f0be455f65dcf0e570b8fe4598cd8cc6b10c294/sstable/rowblk/rowblk_writer.go#L24)):*
> ```Go
> 	// restart points table will be unable to express the position of the pair,
> 	// resulting in undefined behavior and arbitrary corruption.
> 	MaximumRestartOffset = 1 << 31
> ```

I think it should be 1<<31-1

sstable/options.go
```diff
@@ -15,8 +15,8 @@ import (
 )
 
 const (
-	// MaximumBlockSize is the maximum permissible size of a block.
-	MaximumBlockSize = rowblk.MaximumSize
+	// MaximumRestartOffset is the maximum permissible size of a block.
```
RaduBerinde, line 18: nit: Maximum permissible value for a restart offset - i.e the maximum block size which allows adding a restart point.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4503#-:-ONLVlFV47RkPfk7yqCN:bnfp4nl)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

sumeerbhola (approved):
:lgtm: 

> Reviewed 12 of 14 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4503#-:-ONLW8z7DsqW_4-DrYet:bg7igkv)* status: all files reviewed, 4 unresolved discussions (waiting on @jbowens)

___
*[`sstable/rowblk/rowblk_iter_test.go` line 496 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4503#-ONLUzZs4WVHOi99xCo-:-ONLUzZs4WVHOi99xCo0:bwk41z5) ([raw file](https://github.com/cockroachdb/pebble/blob/66fd51f8bb1a640e9fef89274bda5aca1551fe40/sstable/rowblk/rowblk_iter_test.go#L496)):*
> ```Go
> 
> 	writer := &Writer{RestartInterval: 1}
> 	require.NoError(t, writer.Add(base.InternalKey{UserKey: largeKey}, largeValue))
> ```

I'm confused. Was this working and continues to work because of a single k-v and so the restart never has to encode a high value?
___
*[`sstable/rowblk/rowblk_iter_test.go` line 532 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4503#-ONLVSWW6QMIrtH98E1X:-ONLVSWW6QMIrtH98E1Y:b-76wbkq) ([raw file](https://github.com/cockroachdb/pebble/blob/66fd51f8bb1a640e9fef89274bda5aca1551fe40/sstable/rowblk/rowblk_iter_test.go#L532)):*
> ```Go
> 
> 	// Test that writing to a block that is already >= 2GiB
> 	// causes a panic to occur.
> ```

nit: no longer a panic

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4503)* status: 13 of 14 files reviewed, 3 unresolved discussions (waiting on @sumeerbhola)

___
*[`sstable/rowblk/rowblk_iter_test.go` line 496 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4503#-ONLUzZs4WVHOi99xCo-:-ONLYZUd-pc3p-_ZVoTd:b-yis1qo) ([raw file](https://github.com/cockroachdb/pebble/blob/66fd51f8bb1a640e9fef89274bda5aca1551fe40/sstable/rowblk/rowblk_iter_test.go#L496)):*

I'm confused. Was this working and continues to work because of a single k-v and so the restart never has to encode a high value?
</blockquote></details>

Yeah. The same backport that added the assertion fixed this to work because no individual restart point has to encode a high value. We previously had a bug where we couldn't handle the offset of the restart table itself exceeding 2^32-1 but the backport fixed that. I added a comment to the top of the test to explain what's happening here better.
___
*[`sstable/rowblk/rowblk_iter_test.go` line 532 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4503#-ONLVSWW6QMIrtH98E1X:-ONLYqrJARFYQqznBQUy:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/66fd51f8bb1a640e9fef89274bda5aca1551fe40/sstable/rowblk/rowblk_iter_test.go#L532)):*

nit: no longer a panic
</blockquote></details>

Done

jbowens:
TFTRs!

RaduBerinde:
It doesn't build with GOARCH=386:
```
Error: sstable/rowblk/rowblk_iter_test.go:497:40: cannot use largeKeySize (untyped int constant 2147483648) as int value in argument to bytes.Repeat (overflows)
Error: sstable/rowblk/rowblk_iter_test.go:498:42: cannot use largeValueSize (untyped int constant 2147483648) as int value in argument to bytes.Repeat (overflows)
Error: sstable/rowblk/rowblk_iter_test.go:624:86: cannot use fourGB (untyped int constant 4294967296) as int value in argument to bytes.Repeat (overflows)
```

jbowens:
Put up #4522

PR #4505, opened 2025-04-08
crl-release-24.3: rowblk: raise max block size to (1<<31)-1; error rather than panic

jbowens:
24.3 backport of #4503 and #4507.

----

Relax the maximum block size to reclaim 7-unused bits within the restart offsets. This provides additional leeway for massive blocks we're sometimes forced to construct due to very large key-value pairs. Additionally, this commit turns this check into an error that we propagate up and out of the sstable Writer rather than panicking with an assertion failure. In practice this means a DB that hits this limit may fail a compaction without bringing down the node. The DB may repeat the compaction in a busy loop, but it's possible for the DB to make some forward progress in the meantime.


RaduBerinde (approved):

jbowens:
TFTR!

PR #4522, opened 2025-04-10
rowblk: move 64-bit tests build flag

jbowens:
Move tests that should only run on 64-bit architectures behind a build flag.

### DIFF
--- a/sstable/rowblk/rowblk_64bit_test.go
+++ b/sstable/rowblk/rowblk_64bit_test.go
@@ -1,0 +1,174 @@
+// Copyright 2025 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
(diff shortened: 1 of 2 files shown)


sstable/rowblk/rowblk_64bit_test.go
```diff
@@ -0,0 +1,174 @@
...
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+//go:build amd64 || arm64 || arm64be || ppc64 || ppc64le || mips64 || mips64le || s390x || sparc64 || riscv64 || loong64
+
+package rowblk
+
+import (
+	"bytes"
+	"errors"
```
RaduBerinde, line 11: linter says to use github.com/cockroachdb/errors

RaduBerinde (approved):
Thanks for the quick fix!

sstable/rowblk/rowblk_64bit_test.go
jbowens: Done

PR #4523, opened 2025-04-10
sstable: refactor prop block load

annrpom:
We now pass in an iterator to `load` in order to range all over the keys and values.


annrpom:
TFTR! ('-')7

PR #4530, opened 2025-04-11
db: rename LevelMetrics table fields

jbowens:
Rename LevelMetrics fields that refer to 'files' to refer to 'tables' instead. With the introduction of value separation, significant volumes of data will live in blob files not accounted within these particular existing metrics.


annrpom (approved):
:lgtm:

jbowens:
TFTR!

PR #4535, opened 2025-04-11
db: respect EnableColumnarBlocks option on recent FMVs

jbowens:
Fix a bug that caused EnableColumnarBlocks to be ignored on recent format major versions. Later format major versions may require use of columnar blocks, but not any in use in CockroachDB v25.2.

And by default enable columnar blocks. In the future, the option to configure use of columnar blocks will be removed altogether, and columnar blocks will always be used.


jbowens:
Will backport to 25.2

PR #4538, opened 2025-04-11
db: add support for blob files to versionSet

jbowens:
**manifest: include BlobFileMetadata for deleted blob files**

Adjust the VersionEdit to include a pointer to deleted blob files'
*BlobFileMetadata struct. During runtime when a blob file is deleted, this
pointer will be populated. During MANIFEST replay, it will not. This mirrors
the mechanics of VersionEdit.DeletedTables. It will be used to faciliate that
accounting of zombie blob files at runtime.

**db: add support for blob files to versionSet**

Adapt the versionSet to write extant blob files to new manifests during a
manifest rollover and support tracking zombie and obsolete blob files.

Informs #/112.

### DIFF
--- a/internal/manifest/version_edit.go
+++ b/internal/manifest/version_edit.go
@@ -11,6 +11,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"maps"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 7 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4538#-:-ONcHWEk6NvUwNZgYPGP:bnfp4nl)* status: 0 of 7 files reviewed, all discussions resolved

PR #4547, opened 2025-04-12
metamorphic: fix key format when running a single config

RaduBerinde:
`RunOnce` takes the `KeyFormat` separately, but the one we pass is
derived from the flags, not the one that is passed to us from e.g.
`TestMetaCockroachKVs`. This change fixes this.

Note that when running the test normally, the top-level process passes
`--key-format` to the child process, so in that case it works out.


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4547#-:-ONo62bf0xpc6RzCkqXy:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #4562, opened 2025-04-15
manifest,record: disambiguate unexpected EOFs

### LINKED ISSUE #4561: manifest: version edit decode misinterprets ErrUnexpectedEOF
During version edit decoding, we read from an `io.Reader` backed by the `record` package. If a record in the MANIFEST is spread across multiple chunks, this reader handles reading new chunks as the caller pulls from the reader. It's possible for a record that is split across multiple chunks to be only partially persisted if it was in the process of being written when the process exited. When this happens, the underlying reader will surface an `io.ErrUnexpectedEOF`.

During version edit decoding, when we expect to read _n_ bytes from a record and find fewer than _n_ bytes in the payload, this gets surfaced from `io.ReadFull` as an `io.ErrUnexpectedEOF`. This case should be considered corruption, because the payload is improper.

Currently, because these scenarios both use `io.ErrUnexpectedEOF` these cases are indistinguishable. If the `io.ErrUnexpectedEOF` was surfaced by the underlying record reader, we want to treat this as a logical EOF, not an error.

Jira issue: PEBBLE-404

jbowens:
Disambiguate an unexpected EOF in the record envelope (eg, missing chunks) from an unexpected EOF within the body of the record. An unexpected EOF of the envelope may occur if the process crashes while appending a version edit to the manifest file. An unexpected EOF within the record indicates corruption.

Previously, some version edit code paths re-mapped an unexpected EOF while parsing the record envelope into a corruption error. In this case, recovery failed improperly when it should've ignored the incomplete version edit.

In the other direction, some code paths would return io.ErrUnexpectedEOF while decoding the version edit structure within a valid envelope record. In this case, the caller interpreted the ErrUnexpectedEOF as a sudden end to the record envelope, and recovery succeeded improperly when it should've aborted with a corruption error.

Close #4561.

### DIFF
--- a/internal/manifest/version_edit.go
+++ b/internal/manifest/version_edit.go
@@ -25,8 +25,6 @@ import (
 // TODO(peter): describe the MANIFEST file format, independently of the C++
 // project.
 
-var errCorruptManifest = base.CorruptionErrorf("pebble: corrupt manifest")
-
 type byteReader interface {
 	io.ByteReader
 	io.Reader
@@ -551,7 +549,7 @@ func (v *VersionEdit) Decode(r io.Reader) error {
 			return base.CorruptionErrorf("column families are not supported")
 
 		default:
-			return errCorruptManifest
+			return base.CorruptionErrorf("MANIFEST: unknown tag: %d", tag)
 		}
 	}
 	return nil
@@ -847,7 +845,7 @@ func (d versionEditDecoder) readBytes() ([]byte, error) {
 	_, err = io.ReadFull(d, s)
 	if err != nil {
 		if err == io.ErrUnexpectedEOF {
-			return nil, errCorruptManifest
+			return nil, base.CorruptionErrorf("pebble: corrupt manifest: failed to read %d bytes", n)
 		}
 		return nil, err
 	}
@@ -860,7 +858,7 @@ func (d versionEditDecoder) readLevel() (int, error) {
 		return 0, err
 	}
 	if u >= NumLevels {
-		return 0, errCorruptManifest
+		return 0, base.CorruptionErrorf("pebble: corrupt manifest: level %d >= %d", u, NumLevels)
 	}
 	return int(u), nil
 }
@@ -876,8 +874,8 @@ func (d versionEditDecoder) readFileNum() (base.FileNum, error) {
 func (d versionEditDecoder) readUvarint() (uint64, error) {
 	u, err := binary.ReadUvarint(d)
 	if err != nil {
-		if err == io.EOF {
-			return 0, errCorruptManifest
+		if err == io.EOF || err == io.ErrUnexpectedEOF {
+			return 0, base.CorruptionErrorf("pebble: corrupt manifest: failed to read uvarint")
 		}
 		return 0, err
 	}
--- a/open.go
+++ b/open.go
@@ -950,20 +950,29 @@ func (d *DB) replayWAL(
 			// surfaces record.ErrInvalidChunk or record.ErrZeroedChunk. These
 			// errors are always indicative of corruption and data loss.
 			//
-			// Otherwise, the reader surfaces io.ErrUnexpectedEOF indicating that
-			// the WAL terminated uncleanly and ambiguously. If the WAL is the
-			// most recent logical WAL, the caller passes in (strictWALTail=false),
-			// indicating we should tolerate the unclean ending. If the WAL is an
-			// older WAL, the caller passes in (strictWALTail=true), indicating that
-			// the WAL should have been closed cleanly, and we should interpret
-			// the `io.ErrUnexpectedEOF` as corruption and stop recovery.
+			// Otherwise, the reader surfaces record.ErrUnexpectedEOF indicating
+			// that the WAL terminated uncleanly and ambiguously. If the WAL is
+			// the most recent logical WAL, the caller passes in
+			// (strictWALTail=false), indicating we should tolerate the unclean
+			// ending. If the WAL is an older WAL, the caller passes in
+			// (strictWALTail=true), indicating that the WAL should have been
+			// closed cleanly, and we should interpret the
+			// `record.ErrUnexpectedEOF` as corruption and stop recovery.
 			if errors.Is(err, io.EOF) {
 				break
-			} else if errors.Is(err, io.ErrUnexpectedEOF) && !strictWALTail {
+			} else if errors.Is(err, record.ErrUnexpectedEOF) && !strictWALTail {
 				break
-			} else if errors.Is(err, record.ErrInvalidChunk) || errors.Is(err, record.ErrZeroedChunk) {
-				// If a read-ahead returns one of these errors, they should be marked with corruption.
-				// Other I/O related errors should not be marked with corruption and simply returned.
+			} else if (errors.Is(err, record.ErrUnexpectedEOF) && strictWALTail) ||
+				errors.Is(err, record.ErrInvalidChunk) || errors.Is(err, record.ErrZeroedChunk) {
+				// If a read-ahead returns record.ErrInvalidChunk or
+				// record.ErrZeroedChunk, then there's definitively corruption.
+				//
+				// If strictWALTail=true, then record.ErrUnexpectedEOF should
+				// also be considered corruption because the strictWALTail
+				// indicates we expect a clean end to the WAL.
+				//
+				// Other I/O related errors should not be marked with corruption
+				// and simply returned.
 				err = errors.Mark(err, ErrCorruption)
 			}
 
--- a/record/record.go
+++ b/record/record.go
@@ -215,13 +215,28 @@ var (
 	// header, length, or checksum. This usually occurs when a log is recycled,
 	// but can also occur due to corruption.
 	ErrInvalidChunk = errors.New("pebble/record: invalid chunk")
+
+	// ErrUnexpectedEOF is returned if a log file ends unexpectedly. It
+	// indicates the unexpected end of the log file or an in-progress record
+	// envelope itself. ErrUnexpectedEOF may be returned by Reader when it
+	// encounters an invalid chunk but observes no evidence that the invalid
+	// chunk is caused by corruption (i.e., no future chunk indicates that
+	// offset should be valid and durably synced.)
+	//
+	// This error is defined separately from io.ErrUnexpectedEOF to disambiguate
+	// this case from from the case of an unexpected end of the record's payload
+	// while decoding at a higher-level (eg, version edit decoding). If a
+	// higher-level decoding routine returns record.ErrUnexpectedEOF, it
+	// unambiguously indicates that the log file itself ended unexpectedly. The
+	// record.Reader will never return io.ErrUnexpectedEOF, just record.ErrUnexpectedEOF.
+	ErrUnexpectedEOF = errors.New("pebble/record: unexpected EOF")
 )
 
 // IsInvalidRecord returns true if the error matches one of the error types
 // returned for invalid records. These are treated in a way similar to io.EOF
 // in recovery code.
 func IsInvalidRecord(err error) bool {
-	return err == ErrZeroedChunk || err == ErrInvalidChunk || err == io.ErrUnexpectedEOF
+	return err == ErrZeroedChunk || err == ErrInvalidChunk || err == ErrUnexpectedEOF
 }
 
 // Reader reads records from an underlying io.Reader.
@@ -269,7 +284,7 @@ func NewReader(r io.Reader, logNum base.DiskFileNum) *Reader {
 		logNum:   uint32(logNum),
 		blockNum: -1,
 		// invalidOffset is initialized as MaxUint64 so that reading ahead
-		// with the old chunk wire formats results in io.ErrUnexpectedEOF.
+		// with the old chunk wire formats results in ErrUnexpectedEOF.
 		invalidOffset: math.MaxUint64,
 	}
 }
@@ -396,7 +411,7 @@ func (r *Reader) nextChunk(wantFirst bool) error {
 		if err != nil && err != io.ErrUnexpectedEOF {
 			if err == io.EOF && !wantFirst {
 				r.invalidOffset = uint64(r.blockNum)*blockSize + uint64(r.begin)
-				return io.ErrUnexpectedEOF
+				return ErrUnexpectedEOF
 			}
 			return err
 		}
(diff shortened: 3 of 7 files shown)


annrpom (approved):
:lgtm:

sumeerbhola (changes requested):
This looks good, but the error handling in this area of the code is quite subtle, hence a bunch of comments/questions.

> Reviewed 6 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4562#-:-ONzii_DEBVFsYaDTi8V:b-su24y4)* status: 6 of 7 files reviewed, 9 unresolved discussions

___
*[`version_set_test.go` line 346 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzgYCgBHjMv1rrHsWQ:-ONzgYCgBHjMv1rrHsWR:b8dtfrd) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/version_set_test.go#L346)):*
> ```Go
> 	rng := rand.New(rand.NewPCG(seed, 0))
> 
> 	var crashClone atomic.Pointer[vfs.MemFS]
> ```

Some commentary would help. Such as
// crashClone is nil until a clone of the memFS is constructed, where the clone will lose 50% of the unsynced data.
// Each iteration constructs one clone at a random time, and the DB keeps setting values until the clone is created.
// Then a new DB is opened with the cloned memFS.
___
*[`version_set_test.go` line 405 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzhDXu88I2D0h6JgwR:-ONzhDXu88I2D0h6JgwS:b-s0f6ir) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/version_set_test.go#L405)):*
> ```Go
> 		}()
> 	}
> }
> ```

I think we could also use a (hopefully randomized) test where `versionEditDecoder` encounters a `io.ErrUnexpectedEOF` or `io.EOF`.
___
*[`open.go` line 965 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONz_1mo-pZBb-tgUM5i:-ONz_1mo-pZBb-tgUM5j:b-ke4xdh) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/open.go#L965)):*
> ```Go
> 			} else if errors.Is(err, record.ErrUnexpectedEOF) && !strictWALTail {
> 				break
> 			} else if errors.Is(err, record.ErrInvalidChunk) || errors.Is(err, record.ErrZeroedChunk) {
> ```

why are we marking this as corruption when `!strictWALTail`? I expected this to be using `IsInvalidRecord`.
___
*[`internal/manifest/version_edit.go` line 884 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzho4jAFqUC7KPLlt7:-ONzho4jAFqUC7KPLlt8:b27xt51) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/internal/manifest/version_edit.go#L884)):*
> ```Go
> 	u, err := binary.ReadUvarint(d)
> 	if err != nil {
> 		if err == io.EOF {
> ```

`binary.ReadUvarint` reads a byte at a time and return `io.ErrUnexpectedEOF`. So we won't return `errCorruptManifest`.
___
*[`record/record.go` line 216 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzUJub5HFi8SsMvPQB:-ONzUJub5HFi8SsMvPQC:b8fl4pm) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/record/record.go#L216)):*
> ```Go
> 	// ErrInvalidChunk is returned if a chunk is encountered with an invalid
> 	// header, length, or checksum. This usually occurs when a log is recycled,
> 	// but can also occur due to corruption.
> ```

is it correct that this can happen due a torn write in a non-recycled case too, due to out of order writes on cloud disks?
___
*[`record/record.go` line 223 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzUoDjDBuoiTQoSAIK:-ONzUoDjDBuoiTQoSAIL:bva4c00) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/record/record.go#L223)):*
> ```Go
> 	// unexpected end of the record envelope (which may be expected during
> 	// recovery due to a torn write), from an unexpected end of the record's
> 	// payload while decoding at a higher-level (eg, version edit decoding).
> ```

which one of these two cases is this?
___
*[`record/record.go` line 225 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzcYHH5UrGu03beyEI:-ONzcYHH5UrGu03beyEJ:b69rv4p) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/record/record.go#L225)):*
> ```Go
> 	// payload while decoding at a higher-level (eg, version edit decoding).
> 	ErrUnexpectedEOF = errors.New("pebble/record: unexpected EOF")
> )
> ```

I think it would be easier to reason about things if we could say that `record.Reader` never returns `io.ErrUnexpectedEOF`. I suspect this is already true in that it swallows that error if returned from the underlying `io.Reader`.
___
*[`record/record.go` line 476 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzVpDZ05vqsx0AyOL6:-ONzVpDZ05vqsx0AyOL7:b-yc7x8a) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/record/record.go#L476)):*
> ```Go
> 				r.loggerForTesting.logf("\tEncountered io.EOF; returning ErrUnexpectedEOF since no sync offset found.\n")
> 			}
> 			return ErrUnexpectedEOF
> ```

So this is a torn write yes? And we are returning `ErrUnexpectedEOF`, so we must be returning `io.ErrUnexpectedEOF` for decoding issues at a higher-level?
___
*[`record/record.go` line 483 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzVlg_D_onik0sHH-f:-ONzVlg_D_onik0sHH-g:buo7lhx) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/record/record.go#L483)):*
> ```Go
> 		// However, if the error is not ErrUnexpectedEOF, then this
> 		// error should be surfaced.
> 		if err != nil && !errors.Is(err, io.ErrUnexpectedEOF) {
> ```

was this a bug? Do we have testing for this? Or did something start wrapping `io.ErrUnexpectedEOF` in this PR?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4562)* status: 4 of 7 files reviewed, 9 unresolved discussions (waiting on @sumeerbhola)

___
*[`open.go` line 965 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONz_1mo-pZBb-tgUM5i:-ONzsnLg5ddyaaperzMr:b-dm6mpa) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/open.go#L965)):*

why are we marking this as corruption when `!strictWALTail`? I expected this to be using `IsInvalidRecord`.
</blockquote></details>

There's a subtlety here that we should clean up in the interface (in a subsequent commit that will not be backported). With Edward's new readahead logic, the `record.Reader` will never return `record.ErrInvalidChunk` or `record.ErrZeroedChunk` unless it has verified that indeed it is corruption. Otherwise, it returns `record.ErrUnexpectedEOF` indicating that there was an unclean end to the log.
___
*[`internal/manifest/version_edit.go` line 884 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzho4jAFqUC7KPLlt7:-ONzrVbo7EISvKjElt0r:b7skvt3) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/internal/manifest/version_edit.go#L884)):*

`binary.ReadUvarint` reads a byte at a time and return `io.ErrUnexpectedEOF`. So we won't return `errCorruptManifest`.
</blockquote></details>

My reading of its code is that it _can_ return `io.EOF`, but only if it reads zero bytes. I updated this code to return a corruption error in both cases.
___
*[`record/record.go` line 216 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzUJub5HFi8SsMvPQB:-ONztVqG-TOsnmMbLnHa:b-acjze) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/record/record.go#L216)):*

is it correct that this can happen due a torn write in a non-recycled case too, due to out of order writes on cloud disks?
</blockquote></details>

That's my understanding—I'll defer updating this comment until my follow up because in my follow up this will no longer be exported.
___
*[`record/record.go` line 223 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzUoDjDBuoiTQoSAIK:-ONzpuCS5FwU0v3mQvbH:bvnrpcm) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/record/record.go#L223)):*

which one of these two cases is this?
</blockquote></details>

updated
___
*[`record/record.go` line 225 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzcYHH5UrGu03beyEI:-ONzpunu0Imr-7z-Eqxc:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/record/record.go#L225)):*

I think it would be easier to reason about things if we could say that `record.Reader` never returns `io.ErrUnexpectedEOF`. I suspect this is already true in that it swallows that error if returned from the underlying `io.Reader`.
</blockquote></details>

Done
___
*[`record/record.go` line 476 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzVpDZ05vqsx0AyOL6:-ONzqHKf9fMAXBki2ujC:bqli53) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/record/record.go#L476)):*

So this is a torn write yes? And we are returning `ErrUnexpectedEOF`, so we must be returning `io.ErrUnexpectedEOF` for decoding issues at a higher-level?
</blockquote></details>

Yeah that's right
___
*[`record/record.go` line 483 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzVlg_D_onik0sHH-f:-ONzqXpO6x56MsOD0aes:b-oexct2) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/record/record.go#L483)):*

was this a bug? Do we have testing for this? Or did something start wrapping `io.ErrUnexpectedEOF` in this PR?
</blockquote></details>

This was unnecessary, and I'll remove it.
___
*[`version_set_test.go` line 346 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzgYCgBHjMv1rrHsWQ:-ONzr0VqDFzoAhgymxMx:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/version_set_test.go#L346)):*

Some commentary would help. Such as
// crashClone is nil until a clone of the memFS is constructed, where the clone will lose 50% of the unsynced data.
// Each iteration constructs one clone at a random time, and the DB keeps setting values until the clone is created.
// Then a new DB is opened with the cloned memFS.
</blockquote></details>

Done.
___
*[`version_set_test.go` line 405 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONzhDXu88I2D0h6JgwR:-ONzsPhW4x0twginWWC6:b-kg3p68) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/version_set_test.go#L405)):*

I think we could also use a (hopefully randomized) test where `versionEditDecoder` encounters a `io.ErrUnexpectedEOF` or `io.EOF`.
</blockquote></details>

Note that it actually is OK now for versionEditDecoder to encounter an `io.ErrUnexpectedEOF` because the caller will _not_ treat that as benign (since it's not a `record.ErrUnexpectedEOF`).

sumeerbhola (approved):
:lgtm: 

> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4562#-:-OOOvTxE6GNwfBce2wbc:bg7igkv)* status: all files reviewed (commit messages unreviewed), 9 unresolved discussions (waiting on @jbowens)

___
*[`open.go` line 965 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONz_1mo-pZBb-tgUM5i:-OOOuSdb9VhUlqPQfYcM:b9y2p84) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/open.go#L965)):*

There's a subtlety here that we should clean up in the interface (in a subsequent commit that will not be backported). With Edward's new readahead logic, the `record.Reader` will never return `record.ErrInvalidChunk` or `record.ErrZeroedChunk` unless it has verified that indeed it is corruption. Otherwise, it returns `record.ErrUnexpectedEOF` indicating that there was an unclean end to the log.
</blockquote></details>

Ack.

What about the case `errors.Is(err, record.ErrUnexpectedEOF) && strictWALTail`? Is there a reason we are not marking it as `ErrCorruption`? I suppose this marking as `ErrCorruption` isn't necessary since the caller expects any error returned by replay WAL is a bad error. Is my reading accurate?

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4562#-:-OOUSFDg9_7fzjVjMM5z:b-814g8k)* status: 4 of 7 files reviewed, 9 unresolved discussions (waiting on @sumeerbhola)

___
*[`open.go` line 965 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4562#-ONz_1mo-pZBb-tgUM5i:-OOURn47DymoGM4cDb4m:b119u8t) ([raw file](https://github.com/cockroachdb/pebble/blob/db595c2c7742aec4ab362dbb2478f2e4fa82779f/open.go#L965)):*

Ack.

What about the case `errors.Is(err, record.ErrUnexpectedEOF) && strictWALTail`? Is there a reason we are not marking it as `ErrCorruption`? I suppose this marking as `ErrCorruption` isn't necessary since the caller expects any error returned by replay WAL is a bad error. Is my reading accurate?
</blockquote></details>

Ah, right. Yeah, I think all this marking as `ErrCorruption` isn't strictly necessary. I updated this logic to also mark the error as corruption in the case you described

Issue #4448, opened 2025-03-28
tool: support reading blob references (and their values)


jbowens:
The tool package has a few commands that read sstables and can encounter blob handles. Sometimes these commands read sstables in isolation (eg, `find`, `sstable check`, `sstable scan`). We should adjust these commands to support two methods of handling blob handles[1]:

1. Load the referenced value from blob files in the same directory, if found.
2. Print the raw blob handle without reading from any blob files.

[1]
https://github.com/cockroachdb/pebble/blob/d8b52a2b396575007710f0707a6c28eea2083d0b/tool/find.go#L471-L473
https://github.com/cockroachdb/pebble/blob/d8b52a2b396575007710f0707a6c28eea2083d0b/tool/sstable.go#L338-L340
https://github.com/cockroachdb/pebble/blob/d8b52a2b396575007710f0707a6c28eea2083d0b/tool/sstable.go#L167-L169

Jira issue: PEBBLE-373

Epic CRDB-20379

jbowens:
> Load the referenced value from blob files in the same directory, if found.

I just realized that this is tricky right now. We don't have a mapping from the `blob.ReferenceID` to the blob file number in the sstable itself. It's only in the MANIFEST. Even if we were to include the blob file numbers in the sstable, the blob file number may change once we have blob file rewriting. 🤔

annrpom:
^ + from standup (feel free to edit in any holes):
- standalone cmds that just take an sstable (so just `sstable scan`) need to be adapted to also take a path to the MANIFEST file to get our mappings

PR #4563, opened 2025-04-16
tool: add blob handle print mode to tools

annrpom:
This patch adds the ability for commands in the `tool` package (`find`, `sstable check`,`sstable scan`) that read sstables in isolation to support printing the inline blob handles of val-separated kvs.

Informs: #4448

### DIFF
--- a/sstable/values.go
+++ b/sstable/values.go
@@ -21,6 +21,24 @@ var AssertNoBlobHandles = TableBlobContext{
 	References: nil,
 }
 
+// DebugHandlesBlobContext is a TableBlobContext that configures a sstable
+// iterator to return the partially decoded inline blob handle upon encountering
+// a value that references an external blob file.
+var DebugHandlesBlobContext = TableBlobContext{
+	ValueFetcher: nil,
+	References:   nil,
+	// Passing a non-nil BlobHandleFn will return the partially decoded inline
+	// handle before panicking due to missing BlobReferences.
+	BlobHandleFn: func(preface blob.InlineHandlePreface, remainder []byte) base.InternalValue {
+		handleSuffix := blob.DecodeHandleSuffix(remainder)
+		ih := blob.InlineHandle{
+			InlineHandlePreface: preface,
+			HandleSuffix:        handleSuffix,
+		}
+		return base.MakeInPlaceValue([]byte(ih.String()))
+	},
+}
+
 // BlobReferences provides a mapping from an index to a file number for a
 // sstable's blob references. In practice, this is implemented by
 // manifest.BlobReferences.
@@ -41,6 +59,10 @@ type TableBlobContext struct {
 	// References provides a mapping from an index to a file number for a
 	// sstable's blob references.
 	References BlobReferences
+	// BlobHandleFn is an optional function that is invoked after an inline blob
+	// handle has had its preface decoded. Note that if this function is set,
+	// we will not do any work in making a lazy value.
+	BlobHandleFn func(preface blob.InlineHandlePreface, remainder []byte) base.InternalValue
 }
 
 // defaultInternalValueConstructor is the default implementation of the
@@ -76,13 +98,6 @@ func (i *defaultInternalValueConstructor) GetInternalValueForPrefixAndValueHandl
 		panic(errors.AssertionFailedf("block: %x is neither a valblk or blob handle prefix", vp))
 	}
 
-	// We can't convert a blob handle into an InternalValue without
-	// BlobReferences providing the mapping of a reference index to a blob file
-	// number.
-	if i.blobContext.References == nil {
-		panic(errors.AssertionFailedf("blob references not configured"))
-	}
-
 	// The first byte of [handle] is the valuePrefix byte.
 	//
 	// After that, is the inline-handle preface encoding a) the length of the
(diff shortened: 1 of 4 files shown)


annrpom:
I'll add a test I had going on with this in a later PR when we have the ability to add val-sep kvs while updating our MANIFEST

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4563)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[`sstable/values.go` line 85 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4563#-OO3BMcN8nFyp-J9XmDD:-OO3BMcN8nFyp-J9XmDE:b-emt6kb) ([raw file](https://github.com/cockroachdb/pebble/blob/37864a30bd8b3fc6dc887bbcb981e34e1bfd532f/sstable/values.go#L85)):*
> ```Go
> 	// number.
> 	if i.blobContext.References == nil {
> 		return base.MakeLazyValue(base.LazyValue{})
> ```

I think we'll hit this case and never actually return a value with our `i.blobContext.ValueFetcher`.

Maybe we should add an optional func to the `TableBlobContext` that has a signature like:

```
func(InlineHandlePreface, remainder []byte) base.InternalValue
```

If that func is present on the `TableBlobContext`, we decode the preface and return the result of calling that func. Otherwise, we do the logic in this func to return a proper InternalValue

jbowens (commented):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4563)* status: 1 of 3 files reviewed, 4 unresolved discussions

___
*[`sstable/values.go` line 89 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4563#-OO8eN8M4DeoiKc_iOl8:-OO8eN8M4DeoiKc_iOl9:bz4gryj) ([raw file](https://github.com/cockroachdb/pebble/blob/a56dabc4d6a2d0c82cd2e6f4b06200af20ba61e7/sstable/values.go#L89)):*
> ```Go
> 	// BlobReferences providing the mapping of a reference index to a blob file
> 	// number.
> 	if i.blobContext.References == nil && i.blobContext.BlobHandleFn == nil {
> ```

nit: how about moving this assertion below the new `i.blobContext.BlobHandleFn != nil` clause (and removing the new `&& i.blobContext.BlobHandleFn == nil` part of the conditional)
___
*[`sstable/values.go` line 143 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4563#-OO8fk4692g7ElLifnzy:-OO8fk4692g7ElLifnzz:b-gyalv) ([raw file](https://github.com/cockroachdb/pebble/blob/a56dabc4d6a2d0c82cd2e6f4b06200af20ba61e7/sstable/values.go#L143)):*
> ```Go
> )
> 
> func ConvertToBlobRefMode(s string) BlobRefMode {
> ```

nit: I'd say we keep this enum and the string <> enum conversion in the tool pkg, because I don't think we expect other uses outside the tool pkg
___
*[`sstable/values.go` line 166 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4563#-OO8eqyk6XyzNkrnG1Si:-OO8eqyk6XyzNkrnG1Sj:b9vkmj9) ([raw file](https://github.com/cockroachdb/pebble/blob/a56dabc4d6a2d0c82cd2e6f4b06200af20ba61e7/sstable/values.go#L166)):*
> ```Go
> 				}
> 				fmt.Fprintf(stdout, "%v\n", ih)
> 				return base.InternalValue{}
> ```

What do you think about returning the string representation of the inline handle as an in-place value (rather than writing it to stdout directly)? eg, I 'm thinking

```
return base.MakeInPlaceValue([]byte(ih.String()))
```

And then we could make this a global `var DebugHandlesBlobContext TableContext = ...`

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4563)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`sstable/values.go` line 89 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4563#-OO8eN8M4DeoiKc_iOl8:-OONrdfH5qK2M50PkaV_:bal358d) ([raw file](https://github.com/cockroachdb/pebble/blob/a56dabc4d6a2d0c82cd2e6f4b06200af20ba61e7/sstable/values.go#L89)):*

nit: how about moving this assertion below the new `i.blobContext.BlobHandleFn != nil` clause (and removing the new `&& i.blobContext.BlobHandleFn == nil` part of the conditional)
</blockquote></details>

done -- we will do a little extra work now in the AssertNoBlobHandles case, but i'm assuming that it's negligible/fine
___
*[`sstable/values.go` line 143 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4563#-OO8fk4692g7ElLifnzy:-OONsioR1QKpp1QEpJeg:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/a56dabc4d6a2d0c82cd2e6f4b06200af20ba61e7/sstable/values.go#L143)):*

nit: I'd say we keep this enum and the string <> enum conversion in the tool pkg, because I don't think we expect other uses outside the tool pkg
</blockquote></details>

done
___
*[`sstable/values.go` line 166 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4563#-OO8eqyk6XyzNkrnG1Si:-OONvmTF3d5HokORo3xk:b-7rl0jr) ([raw file](https://github.com/cockroachdb/pebble/blob/a56dabc4d6a2d0c82cd2e6f4b06200af20ba61e7/sstable/values.go#L166)):*

What do you think about returning the string representation of the inline handle as an in-place value (rather than writing it to stdout directly)? eg, I 'm thinking

```
return base.MakeInPlaceValue([]byte(ih.String()))
```

And then we could make this a global `var DebugHandlesBlobContext TableContext = ...`
</blockquote></details>

done

jbowens (approved):
:lgtm_strong: 

> Reviewed 3 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4563#-:-OOO3eiEC2Sy6_uTdQEp:b-uwkowd)* status: 3 of 4 files reviewed, 1 unresolved discussion

annrpom:
TFTR! ('-')7

PR #4564, opened 2025-04-16
db: add compaction value separation mechanics

jbowens:
Introduce a new experimental option and (unexported) format major verson that
together enable compaction-time separation of values into external blob files.

Informs #112.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -174,6 +174,15 @@ func (k compactionKind) String() string {
 	return "?"
 }
 
+// compactingOrFlushing returns "flushing" if the compaction kind is a flush,
+// otherwise it returns "compacting".
+func (k compactionKind) compactingOrFlushing() string {
+	if k == compactionKindFlush {
+		return "flushing"
+	}
+	return "compacting"
+}
+
 // compaction is a table compaction from one level to the next, starting from a
 // given version.
 type compaction struct {
@@ -3143,7 +3152,7 @@ func (d *DB) compactAndWrite(
 		}
 		// Create a new table.
 		writerOpts := d.opts.MakeWriterOptions(c.outputLevel.level, tableFormat)
-		objMeta, tw, err := d.newCompactionOutput(jobID, c, writerOpts)
+		objMeta, tw, err := d.newCompactionOutputTable(jobID, c, writerOpts)
 		if err != nil {
 			return runner.Finish().WithError(err)
 		}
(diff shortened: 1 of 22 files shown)


RaduBerinde (approved):
:lgtm:

I think eventually we will want to be able to change the policy during the compaction rather than making a decision for the entire compaction.

The way I see this working is that when we start a compaction, we call into a function that determines the "desired properties" for the span that starts at that key, and an end key at which the properties change. When/if we reach the end key, we finish the current sstable and switch to using the new properties. This mechanism would replace `RequiredInPlaceValueBound` and would be used for other features like tiering and others (e.g. different compression algorithms for specific key ranges).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4564#-:-OO-_wA-Ap_UaJfiGvrX:b-5cipvs)* status: 0 of 28 files reviewed, 1 unresolved discussion (waiting on @xxmplus)

___
*[`value_separation.go` line 74 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4564#-OO-ZzVl83L0AkMahlmZ:-OO-ZzVl83L0AkMahlm_:b-yrow8) ([raw file](https://github.com/cockroachdb/pebble/blob/f9f4a87731518d28f82a0b3dddd80b08ecadfbbb/value_separation.go#L74)):*
> ```Go
> 	c *compaction, policy ValueSeparationPolicy,
> ) (writeBlobs bool, referenceDepth manifest.BlobReferenceDepth) {
> 	// Flushes will have no existing references to blob files and should their
> ```

[nit] should write

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4564)* status: 0 of 28 files reviewed, 2 unresolved discussions

___
*[`value_separation.go` line 44 at r6](https://reviewable.io/reviews/cockroachdb/pebble/4564#-OO87Igf96UgzK4mefOt:-OO87Igf96UgzK4mefOu:b-b8weu0) ([raw file](https://github.com/cockroachdb/pebble/blob/001d4f5e0cba666a4c3324feb8a1c06d31b77a34/value_separation.go#L44)):*
> ```Go
> 	// We're allowed to write blob references. Determine whether we should carry
> 	// forward existing blob references, or write new ones.
> 	if writeBlobs, outputBlobReferenceDepth := shouldWriteBlobFiles(c, policy); !writeBlobs {
> ```

nit: so the `outputBlobReferenceDepth` is only used when `writeBlobs` is false?
The comment for `shouldWriteBlobFiles` and the code in there suggest otherwise.

RaduBerinde:
*[`value_separation.go` line 63 at r6](https://reviewable.io/reviews/cockroachdb/pebble/4564#-OOJaFwL4XsllZkSqcS1:-OOJaFwL4XsllZkSqcS2:b3ahuk9) ([raw file](https://github.com/cockroachdb/pebble/blob/001d4f5e0cba666a4c3324feb8a1c06d31b77a34/value_separation.go#L63)):*
> ```Go
> 		// TODO(jackson): Don't propagate these bounds if they don't
> 		// overlap with the compaction's bounds.
> 		requiredInPlaceValueBound: d.opts.Experimental.RequiredInPlaceValueBound,
> ```

I'd like to tease this apart a bit.

The existing usecase for requiring in-place values is to disable discriminating between keys based on their suffixes, which doesn't make sense for lock keys. From that perspective, the existing setting could be named `DisableSuffixBasedSeparation`.

Whether or not we write values to blobs is related but not the same thing. It may well be the case that we also want to disable writing to blobs for lock keys, but the reasoning would be different (we want to minimize read latency perhaps? or we know that the KVs are short-lived?)

I'm prototyping replacing this bound with a new "span policy" and trying to find the best way to express what we want as a policy. Ideally the policy would contain high-level hints rather than specific knobs.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4564)* status: 0 of 28 files reviewed, 3 unresolved discussions

___
*[`value_separation.go` line 44 at r6](https://reviewable.io/reviews/cockroachdb/pebble/4564#-OO87Igf96UgzK4mefOt:-OON5xeX4NFuJmZRIKrX:bnvrm4i) ([raw file](https://github.com/cockroachdb/pebble/blob/001d4f5e0cba666a4c3324feb8a1c06d31b77a34/value_separation.go#L44)):*

nit: so the `outputBlobReferenceDepth` is only used when `writeBlobs` is false?
The comment for `shouldWriteBlobFiles` and the code in there suggest otherwise.
</blockquote></details>

Updated `shouldWriteBlobFiles`'s comments in:

https://github.com/cockroachdb/pebble/pull/4587#issuecomment-2816170120
___
*[`value_separation.go` line 63 at r6](https://reviewable.io/reviews/cockroachdb/pebble/4564#-OOJaFwL4XsllZkSqcS1:-OON66R0A7bF3lpDwZcI:b3a6wl6) ([raw file](https://github.com/cockroachdb/pebble/blob/001d4f5e0cba666a4c3324feb8a1c06d31b77a34/value_separation.go#L63)):*
> The existing usecase for requiring in-place values is to disable discriminating between keys based on their suffixes, which doesn't make sense for lock keys. From that perspective, the existing setting could be named `DisableSuffixBasedSeparation`.

It's true discriminating based on suffix doesn't make sense in the lock table, but I thought we did it for the more general reason of wanting to preserve low-latency access to these keys' values. Although now that I'm interrogating that thought, maybe it doesn't make sense. I think we need to interleave the lock table keys with iteration far more frequently than we actually need to use intents' values.

cc @sumeerbhola

PR #4567, opened 2025-04-17
db: include blob files in checkpoints

### LINKED ISSUE #4555: db: copy blob files in checkpoints
Today taking a checkpoint with `DB.Checkpoint` does not copy blob files at all. We should adjust `Checkpoint` to copy all referenced blob files. When the caller specifies checkpoint spans, we exclude some tables from the checkpoint and amend the checkpointed manifest to remove the excluded tables. We should also exclude any blob files that have no references from _included_ tables, and include the removal of those excluded blob files in the version edit we append.

Jira issue: PEBBLE-400

jbowens:
When checkpointing the database through (*pebble.DB).Checkpoint, link or copy blob files referenced by sstables included in the checkpoint. When a checkpoint operation is limited by key range, the checkpoint will exclude blob files that are part of the current Version of the LSM but not referenced by any sstable overlapping the checkpointed key ranges. Similarly to excluded sstables, these excluded blob files will be marked as removed by a final version edit appended to the checkpoint's MANIFEST so that the MANIFEST accurately reflects the set of blob files in the checkpoint.

Close #4555.

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -12,6 +12,7 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/errors/oserror"
 	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/manifest"
 	"github.com/cockroachdb/pebble/record"
 	"github.com/cockroachdb/pebble/vfs"
 	"github.com/cockroachdb/pebble/vfs/atomicfs"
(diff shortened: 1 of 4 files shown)


annrpom (commented):
:lgtm:, but feel free to wait on another review

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4567#-:-OO4Y5c37bJWZmqyNF73:b-eh8u26)* status: 0 of 4 files reviewed, all discussions resolved

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4567#-:-OO4tf4O2UKFYZvnluKg:bnfp4nl)* status: 0 of 4 files reviewed, 1 unresolved discussion

___
*[`checkpoint.go` line 309 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4567#-OO4tDQjBcPniwplNfnD:-OO4tDQjBcPniwplNfnE:bz3upir) ([raw file](https://github.com/cockroachdb/pebble/blob/b259ca80982124c684f52b48e61285907761ae29/checkpoint.go#L309)):*
> ```Go
> 			}
> 
> 			// Copy and referenced blob files that have not already been copied.
> ```

[nit] any referenced

jbowens:
TFTRs!

Issue #4518, opened 2025-04-10
db: bound worst-case block sizes


jbowens:
Today top-level index blocks, range deletion blocks and range key blocks have potentially unbounded sizes. In the rowblk format, this can cause keys to be written at offsets that are inexpressible as a restart point. We should adjust flushes and compactions to split sstables to bounds these blocks’ sizes.

cockroachlabs/support#3247

Jira issue: PEBBLE-392

PR #4583, opened 2025-04-18
db: add TestLargeKeys

jbowens:
Add a new datadriven unit test that exercises large keys with a long common shared prefix. This test demonstrates the worst-case unbounded block sizes.

Informs #4518.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -1321,6 +1321,29 @@ func runSSTablePropertiesCmd(t *testing.T, td *datadriven.TestData, d *DB) strin
 	return buf.String()
 }
 
+func runLayoutCmd(t *testing.T, td *datadriven.TestData, d *DB) string {
+	var filename string
+	td.ScanArgs(t, "filename", &filename)
+	f, err := d.opts.FS.Open(filename)
+	if err != nil {
+		return err.Error()
+	}
+	readable, err := sstable.NewSimpleReadable(f)
+	if err != nil {
+		return err.Error()
+	}
+	r, err := sstable.NewReader(context.Background(), readable, d.opts.MakeReaderOptions())
+	if err != nil {
+		return err.Error()
+	}
+	defer r.Close()
+	l, err := r.Layout()
+	if err != nil {
+		return err.Error()
+	}
+	return l.Describe(td.HasArg("verbose"), r, nil)
+}
+
 func runPopulateCmd(t *testing.T, td *datadriven.TestData, b *Batch) {
 	var maxKeyLength, valLength int
 	var timestamps []int
--- a/iterator_histories_test.go
+++ b/iterator_histories_test.go
@@ -5,7 +5,6 @@ package pebble
 
 import (
 	"bytes"
-	"context"
 	"fmt"
 	"path/filepath"
 	"strconv"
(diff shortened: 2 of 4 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4583#-:-OO9l_oV48Zhm0DgDv78:bnfp4nl)* status: 0 of 4 files reviewed, 2 unresolved discussions

___
*[`testdata/large_keys` line 51 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4583#-OO9koLOCMO8al4QIzuZ:-OO9koLOCMO8al4QIzu_:b6i40y0) ([raw file](https://github.com/cockroachdb/pebble/blob/346c228bbc9c34906e32d7bb85cf1964cfc792c6/testdata/large_keys#L51)):*
> ```
> ----
> sstable
>  ├── data  offset: 0  length: 46944
> ```

I think we should disable compression to get a better idea of what is going on. I assume we have one KV per data block and then one KV per index block?
___
*[`testdata/large_keys` line 217 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4583#-OO9lRYk0YqNl8auzGIY:-OO9lRYl2ZquOWVukAKF:b-zhkf4m) ([raw file](https://github.com/cockroachdb/pebble/blob/346c228bbc9c34906e32d7bb85cf1964cfc792c6/testdata/large_keys#L217)):*
> ```
> ----
> sstable
>  ├── data  offset: 0  length: 46998
> ```

How come we can fit more data in these blocks? Won't we finish a data block after writing one KV?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4583)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`testdata/large_keys` line 51 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4583#-OO9koLOCMO8al4QIzuZ:-OO9sFEf7wP8hqHvfGfC:bg5gplb) ([raw file](https://github.com/cockroachdb/pebble/blob/346c228bbc9c34906e32d7bb85cf1964cfc792c6/testdata/large_keys#L51)):*

I think we should disable compression to get a better idea of what is going on. I assume we have one KV per data block and then one KV per index block?
</blockquote></details>

Yeah, 1 KV per data block and 1 KV per index block. The properties block gives us some insight into what's going on, but it's a bit limited.

Unfortunately, disabling compression won't reproduce it because the estimated physical file size will increase quickly. We need the effective compression to get the disparity between the file size we split on and the size of the pending top-level index block.
___
*[`testdata/large_keys` line 217 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4583#-OO9lRYk0YqNl8auzGIY:-OO9rhHJ9RVlSHVipdLL:b-ng5kex) ([raw file](https://github.com/cockroachdb/pebble/blob/346c228bbc9c34906e32d7bb85cf1964cfc792c6/testdata/large_keys#L217)):*

How come we can fit more data in these blocks? Won't we finish a data block after writing one KV?
</blockquote></details>

I think it's still 1 KV pair per block. The length of the block is approximately the same as the rowblk blocks. We just have fewer data blocks per sstable because the RawColumnWriter accounts for buffered index blocks resulting in a higher estimated size and more sstable-splitting.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4583#-:-OO9zNWs6BumBkCN4kii:bnfp4nl)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`testdata/large_keys` line 51 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4583#-OO9koLOCMO8al4QIzuZ:-OO9zR0k1KDiHrnVM8IM:blpu5mg) ([raw file](https://github.com/cockroachdb/pebble/blob/346c228bbc9c34906e32d7bb85cf1964cfc792c6/testdata/large_keys#L51)):*

Yeah, 1 KV per data block and 1 KV per index block. The properties block gives us some insight into what's going on, but it's a bit limited.

Unfortunately, disabling compression won't reproduce it because the estimated physical file size will increase quickly. We need the effective compression to get the disparity between the file size we split on and the size of the pending top-level index block.
</blockquote></details>

I see, thanks. I wonder if layout should show uncompressed block lengths as well (I believe we can obtain them without actually decompressing)

jbowens:
TFTR!

PR #4587, opened 2025-04-18
db: export FormatExperimentalValueSeparation

jbowens:
Export the format major version that allows use of value separation, but
leaving FormatNewest pointing to the previous format major version. Exporting
the format allows us to use the format major version from other packages (eg,
the metamorphic tests, the tool package, etc).

**db: avoid use of internal type in ValueSeparationPolicy**

Avoid use of the internal/manifest.BlobReferenceDepth type in the external
ValueSeparationPolicy type. Users are expected to construct a
ValueSeparationPolicy and must be able to construct a MaxBlobReferenceDepth
value.

Also, sneak in a small change to address this comment while we're here: https://github.com/cockroachdb/pebble/pull/4564#pullrequestreview-2778613182

### DIFF
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -1232,8 +1232,8 @@ func TestCompaction(t *testing.T) {
 			maxVersion: internalFormatNewest,
 		},
 		"value_separation": {
-			minVersion: formatValueSeparation,
-			maxVersion: formatValueSeparation,
+			minVersion: FormatExperimentalValueSeparation,
+			maxVersion: FormatExperimentalValueSeparation,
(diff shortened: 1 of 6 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4587#-:-OOOb2Na93LIVMh9sJ_c:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @xxmplus)

jbowens:
TFTRs!

PR #4589, opened 2025-04-18
base: introduce TableNum

RaduBerinde:
We should rename all uses of `FileNum` to `TableNum`. But changing all
uses would be a massive diff, causing problems with backports.

This change renames `base.FileNum` to `TableNum` and leaves a
`FileNum` alias that we will deprecate in time.

### DIFF
--- a/internal/base/filenames.go
+++ b/internal/base/filenames.go
@@ -16,28 +16,32 @@ import (
 	"github.com/cockroachdb/redact"
 )
 
-// FileNum is an internal DB identifier for a table. Tables can be physical (in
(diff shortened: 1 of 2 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4589#-:-OON9s9d3OQIX8w_tPpP:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #4592, opened 2025-04-21
db: introduce SpanPolicy and SpanPolicyFunc

RaduBerinde:
This change introduces the concept of a "span policy". Flushes and
compactions take into account the span policy when writing out ssts.

We replace the `RequiredInPlaceValueBound` option with a span policy.
The effective change is that we now split output sstables at the
boundaries where the policy changes, and the policy affects the writer
options for the entire sstable.

For now, CockroachDB will define a `SpanPolicyFunc` with a single
static span. The intention is that in the future, span policies will
include more things (e.g. tiering metadata) and be tied to KV span
configs.


RaduBerinde:
Draft of the CRDB side change: https://github.com/cockroachdb/cockroach/pull/144749

RaduBerinde:
Gentle ping :)

jbowens (approved):
:lgtm_strong: 

Nice, I like the clean up of options within the sstable writer

> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4592#-:-OPWnvYpEjtsRAe0HUrO:b-505ox)* status: all files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`internal/compact/run.go` line 201 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4592#-OPWmgi24c40zjvttp8L:-OPWmgi24c40zjvttp8M:b-tb0ew4) ([raw file](https://github.com/cockroachdb/pebble/blob/093527b780f56724b48560f752e210358071860c/internal/compact/run.go#L201)):*
> ```Go
> func (r *Runner) FirstKey() []byte {
> 	firstKey := base.MinUserKey(r.cmp, spanStartOrNil(&r.lastRangeDelSpan), spanStartOrNil(&r.lastRangeKeySpan))
> 	if r.kv != nil && firstKey == nil {
> ```

nit: how about leaving a comment explaining that if there was a `r.lastRangeDelSpan` or `r.lastRangeKeySpan`, it necessarily sorts before the first point key.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4592#-:-OPXQ8iJ6Cmff4Vll-Cp:b-814g8k)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sumeerbhola)

PR #4594, opened 2025-04-21
db: run score-based compactions before manual compactions

### LINKED ISSUE #4571: db: manual compactions starve auto compactions
We like to call `DB.Compact` with `parallelize=true` to create many smaller manual compactions. This results in enqueuing a large number of manual compactions for a level, which starve out auto compactions. When doing manual compactions on a CockroachDB node that is up, this is a severe problem. For example, see https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1744911010703789?thread_ts=1744831476.530489&cid=CAC6K3SLU for 1000+ read amp caused by such starvation.

Instead of starving auto compactions, manual compactions should run up to the maximum concurrent compaction capacity after score based compactions are picked.

Jira issue: PEBBLE-407

sumeerbhola:
Score-based compactions are important for maintaining LSM health.

Fixes #4571

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1807,11 +1807,20 @@ func (d *DB) makeCompactionEnvLocked() *compactionEnv {
 
 // pickAnyCompaction tries to pick a manual or automatic compaction.
 func (d *DB) pickAnyCompaction(env compactionEnv) (pc *pickedCompaction) {
-	pc = d.pickManualCompaction(env)
-	if pc == nil && !d.opts.DisableAutomaticCompactions {
-		pc = d.mu.versions.picker.pickAuto(env)
+	// Pick a score-based compaction first, since a misshapen LSM is bad.
+	if !d.opts.DisableAutomaticCompactions {
+		if pc = d.mu.versions.picker.pickAutoScore(env); pc != nil {
+			return pc
+		}
 	}
-	return pc
+	// Pick a manual compaction, if any.
+	if pc = d.pickManualCompaction(env); pc != nil {
+		return pc
+	}
+	if !d.opts.DisableAutomaticCompactions {
+		return d.mu.versions.picker.pickAutoNonScore(env)
+	}
+	return nil
 }
 
 // runPickedCompaction kicks off the provided pickedCompaction. In case the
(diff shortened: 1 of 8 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4594#-:-OOOkm_JEfbTiXoTbInB:bnfp4nl)* status: 0 of 7 files reviewed, 1 unresolved discussion

___
*[`compaction.go` line 1812 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4594#-OOOke02EhIG7dEVLY5J:-OOOke02EhIG7dEVLY5K:bv9g0lg) ([raw file](https://github.com/cockroachdb/pebble/blob/bb886b629fc296814a25fcce5f3e37e034231985/compaction.go#L1812)):*
> ```Go
> 	// Pick a score-based compaction first, since a misshapen LSM is bad.
> 	if !d.opts.DisableAutomaticCompactions {
> 		if pc = d.mu.versions.picker.pickAutoScore(env); pc != nil {
> ```

Is it possible that with a high-enough write load, there will always be a score-based compaction to choose? (starving out the manual compaction indefinitely?)

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4594)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`compaction.go` line 1812 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4594#-OOOke02EhIG7dEVLY5J:-OOOx3iTD4TLA75ANVc5:b-rve5z5) ([raw file](https://github.com/cockroachdb/pebble/blob/bb886b629fc296814a25fcce5f3e37e034231985/compaction.go#L1812)):*

Yes, definitely possible. I think the recommendation there would be to bring down the node and then do a manual compaction.

The production node that motivated this was configured to allow 7 concurrent compactions and only 2-3 auto compactions were typically running -- so there was spare capacity.
</blockquote></details>

It would be cool if instead of taking `parallelize bool`, `DB.Compact` took a `maxConcurrentCompactions` or something, restricting the concurrency that will be used for the manual compaction.

RaduBerinde:
Maybe we want to backport this to previous releases?

PR #4596, opened 2025-04-21
db: allow cancellation of manual compactions

### LINKED ISSUE #4569: db: DB.Compact should support context cancellation
It is not uncommon for users to start an expensive manual compaction and realize that due to the performance side effects of it, they would like to cancel it. In CockroachDB, this happens via kvserver's `CompactEngineSpan` which originates in SQL, and this path has a cancellable context. We need to allow Pebble to accept a context and to stop generating new `*manualCompaction` and remove the ones it has already queued.

Jira issue: PEBBLE-406

sumeerbhola:
Fixes #4569

### DIFF
--- a/checkpoint_test.go
+++ b/checkpoint_test.go
@@ -136,7 +136,7 @@ func testCheckpointImpl(t *testing.T, ddFile string, createOnShared bool) {
 			}
 			memLog.Reset()
 			d := dbs[td.CmdArgs[0].String()]
-			if err := d.Compact(nil, []byte("\xff"), false); err != nil {
+			if err := d.Compact(context.Background(), nil, []byte("\xff"), false); err != nil {
 				return err.Error()
 			}
 			d.TestOnlyWaitForCleaning()
@@ -308,7 +308,7 @@ func TestCheckpointCompaction(t *testing.T) {
 		defer cancel()
 		defer wg.Done()
 		for ctx.Err() == nil {
-			if err := d.Compact([]byte("key"), []byte("key999999"), false); err != nil {
+			if err := d.Compact(context.Background(), []byte("key"), []byte("key999999"), false); err != nil {
 				t.Error(err)
 				return
 			}
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -1137,6 +1137,83 @@ func TestCompaction(t *testing.T) {
 				}
 				return "manual compaction blocked until ongoing finished\n" + s
 
+			case "async-compact-with-cancellation":
+				// Cancels a manual compaction that is blocked by an ongoing
+				// compaction. There can be multiple manual compactions created when
+				// parallel is specified.
+				var s string
+				ch := make(chan error, 1)
+				// Number of blocked manual compactions.
+				var numBlocked int
+				td.ScanArgs(t, "num-blocked", &numBlocked)
+				var cancelFunc atomic.Pointer[context.CancelFunc]
+				go func() {
+					compactFunc, cf, err := runCompactCmdAsync(td, d, true)
+					if err == nil {
+						cancelFunc.Store(&cf)
+						err = compactFunc()
+					}
+					if err != nil {
+						ch <- err
+						close(ch)
+						return
+					}
+					d.mu.Lock()
+					s = d.mu.versions.currentVersion().String()
+					d.mu.Unlock()
+					close(ch)
+				}()
+				var compErr error
+				var compDone bool
+				err := try(100*time.Microsecond, 20*time.Second, func() error {
+					select {
+					case compErr = <-ch:
+						// Unexpected, since compaction should be blocked.
+						compDone = true
+						return nil
+					default:
+					}
+					if cancelFunc.Load() == nil {
+						return errors.New("not yet attempted to run manual compactions")
+					}
+					d.mu.Lock()
+					defer d.mu.Unlock()
+					if len(d.mu.compact.manual) != numBlocked {
+						return errors.Errorf("expected %d waiting manual compactions, versus actual %d",
+							numBlocked, len(d.mu.compact.manual))
+					}
+					// Expect to be back to the fake ongoing compactions when the
+					// non-blocked manual compactions are done.
+					if d.mu.compact.compactingCount != 1 {
+						return errors.Errorf("expected 1 ongoing compaction, versus actual %d",
+							d.mu.compact.compactingCount)
+					}
+					return nil
+				})
+				if err != nil {
+					return err.Error()
+				}
+				if compDone {
+					return "manual compaction did not block for ongoing\n" + s
+				}
+				// Cancel the manual compaction.
+				(*cancelFunc.Load())()
+				// Wait for the cancellation to succeed.
+				compErr = <-ch
+				if compErr == nil {
+					return "manual compaction did not have an error\n" + s
+				}
+				d.mu.Lock()
+				deleteOngoingCompaction(ongoingCompaction)
+				ongoingCompaction = nil
+				numQueuedManualCompactions := len(d.mu.compact.manual)
+				d.mu.Unlock()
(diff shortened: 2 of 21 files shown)


sumeerbhola (commented):
I owe a test for the cancellation functionality.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4596#-:-OOP5wje7u8X2feCABwO:bmcndl6)* status: 0 of 20 files reviewed, all discussions resolved (waiting on @annrpom)

sumeerbhola (commented):
Test is ready

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4596#-:-OOQGD4c5Dsd2J3kmmL0:b-qt8xsa)* status: 0 of 21 files reviewed, all discussions resolved (waiting on @annrpom)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4596)* status: 0 of 21 files reviewed, 2 unresolved discussions (waiting on @annrpom)

___
*[`db.go` line 1944 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4596#-OOT4Ym7Dbu6W3u_2VeA:-OOT4Ym7Dbu6W3u_2VeB:b-sguf6z) ([raw file](https://github.com/cockroachdb/pebble/blob/22e61d91f83e0ea9a52fa25a7d03f78aba28f5e1/db.go#L1944)):*
> ```Go
> 		d.mu.Lock()
> 		for i := 0; i < len(d.mu.compact.manual); {
> 			if d.mu.compact.manual[i].id >= manualIDStart && d.mu.compact.manual[i].id < manualIDEnd {
> ```

Would it be cleaner to add a `ctx` to `manualCompaction` and check `Err` in `tryScheduleManualCompaction`? In the future, we may want to plumb `ctx` further down so we can actually cancel an in-progress compaction.
___
*[`db.go` line 1964 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4596#-OOT41reDZXXc07q01hC:-OOT41reDZXXc07q01hD:bp6ypmj) ([raw file](https://github.com/cockroachdb/pebble/blob/22e61d91f83e0ea9a52fa25a7d03f78aba28f5e1/db.go#L1964)):*
> ```Go
> 		case <-ctx.Done():
> 			cancelPendingCompactions()
> 			return ctx.Err()
> ```

Don't we want the call to block until any compactions we started finish?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4596)* status: 0 of 21 files reviewed, 3 unresolved discussions (waiting on @annrpom and @sumeerbhola)

___
*[`compaction_test.go` line 1161 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4596#-OOT53cNBvLW6XkrsMsB:-OOT53cNBvLW6XkrsMsC:b-sou9d5) ([raw file](https://github.com/cockroachdb/pebble/blob/22e61d91f83e0ea9a52fa25a7d03f78aba28f5e1/compaction_test.go#L1161)):*
> ```Go
> 				var compErr error
> 				var compDone bool
> 				err := try(100*time.Microsecond, 5*time.Second, func() error {
> ```

Can you run the test under stressrace? 5s might not be enough

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4596)* status: 0 of 21 files reviewed, 3 unresolved discussions (waiting on @annrpom and @RaduBerinde)

___
*[`db.go` line 1944 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4596#-OOT4Ym7Dbu6W3u_2VeA:-OOVFB5ZAeQkB6yg8-W_:bhoea5n) ([raw file](https://github.com/cockroachdb/pebble/blob/22e61d91f83e0ea9a52fa25a7d03f78aba28f5e1/db.go#L1944)):*

Would it be cleaner to add a `ctx` to `manualCompaction` and check `Err` in `tryScheduleManualCompaction`? In the future, we may want to plumb `ctx` further down so we can actually cancel an in-progress compaction.
</blockquote></details>

That's an interesting alternative. I've added a comment

```
	// On context cancellation, we only cancel the compactions that have not yet
	// started. The assumption is that it is relatively harmless to have the
	// already started compactions run to completion. We don't wait for the
	// ongoing compactions to finish, since the assumption is that the caller
	// has already given up on the operation (and the cancellation error is
	// going to be returned anyway).
	//
	// An alternative would be to store the context in each *manualCompaction,
	// and have the goroutine that retrieves the *manualCompaction for running
	// notice the cancellation and write the cancellation error to
	// manualCompaction.done. That approach would require this method to wait
	// for all the *manualCompactions it has enqueued to finish before returning
	// (to not leak a context). Since there is no timeliness guarantee on when a
	// *manualCompaction will be retrieved for running, the wait until a
	// cancelled context causes this method to return is not bounded. Hence, we
	// don't adopt that approach.
```
Thoughts?
___
*[`db.go` line 1964 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4596#-OOT41reDZXXc07q01hC:-OOUwjOL64ZXswVFTdpP:b-draf5d) ([raw file](https://github.com/cockroachdb/pebble/blob/22e61d91f83e0ea9a52fa25a7d03f78aba28f5e1/db.go#L1964)):*

Don't we want the call to block until any compactions we started finish?
</blockquote></details>

Context cancellation means the caller has already stopped waiting (killed their SQL command), so waiting here seems unnecessary. Also, see response to other comment.
___
*[`compaction_test.go` line 1161 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4596#-OOT53cNBvLW6XkrsMsB:-OOUu82HDMKNBzt5KQxF:bne8a53) ([raw file](https://github.com/cockroachdb/pebble/blob/22e61d91f83e0ea9a52fa25a7d03f78aba28f5e1/compaction_test.go#L1161)):*

Can you run the test under stressrace? 5s might not be enough
</blockquote></details>

Changed it back to 20s, like the other `try` in this test. I had shrunk it to 5s during debugging.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4596#-:-OPVlvwW4oejtApxdNRo:bnfp4nl)* status: 0 of 21 files reviewed, 1 unresolved discussion (waiting on @annrpom)

___
*[`db.go` line 1932 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4596#-OPVlgeN3Wv0ePlb3AMo:-OPVlgeN3Wv0ePlb3AMp:b-w0rkei) ([raw file](https://github.com/cockroachdb/pebble/blob/45f10a651fecaf116dc66b0b31fcf93cc35bf8cb/db.go#L1932)):*
> ```Go
> 	// d.mu.compact.manual.
> 	var manualIDStart, manualIDEnd uint64
> 	if n := len(compactions); n > 0 {
> ```

[nit] if len(compactions) == 0, can we just exit now? (and we can move this check further up)

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4596#-:-OPW1eVHEV2NvUFLSRs1:b-814g8k)* status: 0 of 21 files reviewed, all discussions resolved (waiting on @annrpom)

___
*[`db.go` line 1932 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4596#-OPVlgeN3Wv0ePlb3AMo:-OPW1dDK2DhuJ7M90xoH:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/45f10a651fecaf116dc66b0b31fcf93cc35bf8cb/db.go#L1932)):*

[nit] if len(compactions) == 0, can we just exit now? (and we can move this check further up)
</blockquote></details>

Done

PR #4601, opened 2025-04-22
metrics: show both compensated and uncompensated scores

### LINKED ISSUE #4599: db: include level compensation factor in metrics
The compensation factor is necessary to debug some cases of range dels not getting compacted. This came up in the context of [cockroachlabs/support/issues/3264](https://github.com/cockroachlabs/support/issues/3264).

Jira issue: PEBBLE-421

RaduBerinde:
Add a column showing the compensation factor (ratio between
compensated and uncompensated score) for each level.

Fixes #4599


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4601)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`metrics.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4601#-OOUNUKj2qDhMsJgOtym:-OOUNUKj2qDhMsJgOtyn:by6b1n1) ([raw file](https://github.com/cockroachdb/pebble/blob/12682e61acc425ebf56cb523a92f0f88e7364905/metrics.go#L62)):*
> ```Go
> 	// The level's compensation factor. This is the ratio between the
> 	// compensatedScore and the uncompensatedScore in candidateLevelInfo.
> 	CompensationFactor float64
> ```

curious why print the ratio of the two scores rather than two score columns, one compensated and one uncompensated? seems like we care about the real value of both?

RaduBerinde:
*[`metrics.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4601#-OOUNUKj2qDhMsJgOtym:-OOUpaRkBQ8yikUUNOcB:bdhasft) ([raw file](https://github.com/cockroachdb/pebble/blob/12682e61acc425ebf56cb523a92f0f88e7364905/metrics.go#L62)):*

curious why print the ratio of the two scores rather than two score columns, one compensated and one uncompensated? seems like we care about the real value of both?
</blockquote></details>

Good point, not sure. Come to think of it, should we show the actual scores instead of the score ratios? We would show them for L6 as well, it would give us a bit more info.

RaduBerinde:
Updated to show two scores (the uncompensated and the compensated score).

jbowens (commented):
> Reviewed 8 of 15 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4601)* status: 8 of 16 files reviewed, 3 unresolved discussions (waiting on @sumeerbhola)

___
*[`replay/replay_test.go` line 532 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4601#-OOcMj34-73W0i7Z-b1M:-OOcMj34-73W0i7Z-b1N:bhal882) ([raw file](https://github.com/cockroachdb/pebble/blob/fc7f3a575752f98b6e405e839cc87e9c2a962ae9/replay/replay_test.go#L532)):*
> ```Go
> 			require.Equal(t, int64(0), m.Final.Compact.NumInProgress)
> 			for l := 0; l < len(m.Final.Levels)-1; l++ {
> 				require.Less(t, m.Final.Levels[l].CompensatedScore, max(m.Final.Levels[l+1].UncompensatedScore, 0.01))
> ```

Same question, shouldn't we be comparing `CompensatedScore` to `1.0`?
___
*[`cmd/pebble/test.go` line 358 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4601#-OOcMCpy42XVHxf-fk1q:-OOcMCpz2JrJaBV5ysa7:b-lptfij) ([raw file](https://github.com/cockroachdb/pebble/blob/fc7f3a575752f98b6e405e839cc87e9c2a962ae9/cmd/pebble/test.go#L358)):*
> ```Go
> 		// background compactions are still needed.
> 		for i := range p.Levels[:len(p.Levels)-1] {
> 			if p.Levels[i].CompensatedScore/p.Levels[i+1].UncompensatedScore > 1 {
> ```

I don't understand this; why isn't `p.Levels[i].CompensatedScore > 1`? My understanding is that if a level has `CompensatedScore > 1`, we'll pursue a compaction

jbowens (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4601)* status: 8 of 16 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`cmd/pebble/test.go` line 358 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4601#-OOcMCpy42XVHxf-fk1q:-OOcNBsj2KD_Mh3BCjPA:b-ss92pu) ([raw file](https://github.com/cockroachdb/pebble/blob/fc7f3a575752f98b6e405e839cc87e9c2a962ae9/cmd/pebble/test.go#L358)):*

I don't understand this; why isn't `p.Levels[i].CompensatedScore > 1`? My understanding is that if a level has `CompensatedScore > 1`, we'll pursue a compaction
</blockquote></details>

Ah, I see, I was confused by what we're calling the score in the code these days

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4601)* status: 8 of 16 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`metrics.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4601#-OOUNUKj2qDhMsJgOtym:-OOcOF11CKNFtSAyqsyy:bc0eb84) ([raw file](https://github.com/cockroachdb/pebble/blob/12682e61acc425ebf56cb523a92f0f88e7364905/metrics.go#L62)):*

+1 to showing the scores, given the computation of the ratio is trivial. But we would want both the compensatedScore and uncompensatedScore, so now we would have 3 columns for the score.
</blockquote></details>

This entire time I was confused by what we were calling a 'score' here. Before  a48ac0f47f34ee330652dc3ed45a60707c6d32b0 we were referred to the score to refer to both the pre- and post- ratio calculation values. I don't have any particular naming ideas, but I feel like we should clarify the terminology since it's confusing that a _score_ is no longer what it used to rank the levels by priority.

RaduBerinde:
*[`metrics.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4601#-OOUNUKj2qDhMsJgOtym:-OOcUEQcCQ5J0Aq0xsrG:b-bcbyc8) ([raw file](https://github.com/cockroachdb/pebble/blob/12682e61acc425ebf56cb523a92f0f88e7364905/metrics.go#L62)):*

This entire time I was confused by what we were calling a 'score' here. Before  a48ac0f47f34ee330652dc3ed45a60707c6d32b0 we were referred to the score to refer to both the pre- and post- ratio calculation values. I don't have any particular naming ideas, but I feel like we should clarify the terminology since it's confusing that a _score_ is no longer what it used to rank the levels by priority.
</blockquote></details>

It wasn't that before either - it was the score that we compare against 1 before using another score to rank levels haha  I'm open to suggestions. I chose to use the two values from which everything else is derived, given that we pretty much have to look at the code to understand them anyway..

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4601)* status: 8 of 16 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`metrics.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4601#-OOUNUKj2qDhMsJgOtym:-OOcWpWvBycQpjbktIPd:bo0iqlz) ([raw file](https://github.com/cockroachdb/pebble/blob/12682e61acc425ebf56cb523a92f0f88e7364905/metrics.go#L62)):*

It wasn't that before either - it was the score that we compare against 1 before using another score to rank levels haha  I'm open to suggestions. I chose to use the two values from which everything else is derived, given that we pretty much have to look at the code to understand them anyway..
</blockquote></details>

Heh, yeah, I think before a48ac0f the "score" described the post-ratio score including compensation, and the same value was both compared to 1 and used for ranking.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4601)* status: 8 of 16 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`metrics.go` line 62 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4601#-OOUNUKj2qDhMsJgOtym:-OOde4K78U536lAgd7yw:b-aiupdl) ([raw file](https://github.com/cockroachdb/pebble/blob/12682e61acc425ebf56cb523a92f0f88e7364905/metrics.go#L62)):*

Ok, I propose I add a column that shows the uncompensated score ratio whenever the compensated score ratio is above 1.0 and - otherwise. That is the value we rank by. And we leave the other two columns to have more details.

Separatrly, we can rename "score ratio" to "score" and "score" to something else, like "bloat"
</blockquote></details>

Updated to the three columns described above.
___
*[`cmd/pebble/test.go` line 358 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4601#-OOcMCpy42XVHxf-fk1q:-OOde9_EDt71H9NIDPIT:b-5n7foj) ([raw file](https://github.com/cockroachdb/pebble/blob/fc7f3a575752f98b6e405e839cc87e9c2a962ae9/cmd/pebble/test.go#L358)):*

Ah, I see, I was confused by what we're calling the score in the code these days
</blockquote></details>

This is much cleaner now that I included the real `Score`

jbowens (approved):
:lgtm:

Looks great

> Reviewed 1 of 15 files at r4, 7 of 12 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4601#-:-OOdllA1E_-FQrcoS4_S:b-9an2pc)* status: 11 of 16 files reviewed, all discussions resolved (waiting on @sumeerbhola)

RaduBerinde:
TFTR!

Issue #2047, opened 2022-10-24
manifest: optimize sizeof FileMetadata


sumeerbhola:
The FileMetadata struct has grown to sizeof 344 bytes, which means 344MB with 1M files (which can happen for various reasons), and is not an insignificant amount of memory. There is of course additional cost of FileMetadata due to the InternalKey.UserKey byte slices, which is something discussed in https://github.com/cockroachdb/pebble/issues/1741.

We can reduce the FileMetadata size to 216 bytes with some rearrangement and moving mostly empty sets of fields to structs with pointers in FileMetadata (see ** for the changes below)
```
type FileMetadata struct {
	// Atomic contains fields which are accessed atomically. Go allocations
	// are guaranteed to be 64-bit aligned which we take advantage of by
	// placing the 64-bit fields which we access atomically at the beginning
	// of the FileMetadata struct. For more information, see
	// https://golang.org/pkg/sync/atomic/#pkg-note-BUG.
	Atomic struct {
		// AllowedSeeks is used to determine if a file should be picked for
		// a read triggered compaction. It is decremented when read sampling
		// in pebble.Iterator after every after every positioning operation
		// that returns a user key (eg. Next, Prev, SeekGE, SeekLT, etc).
		AllowedSeeks int64
		// statsValid is 1 if stats have been loaded for the table. The
		// TableStats structure is populated only if valid is 1.
		statsValid uint32
		// ** MOVED refs into struct to make it 16 byte instead of 16 byte struct
		// + 8 byte refs **
		// Reference count for the file: incremented when a file is added to a
		// version and decremented when the version is unreferenced. The file is
		// obsolete when the reference count falls to zero.
		refs int32
	}

	// InitAllowedSeeks is the inital value of allowed seeks. This is used
	// to re-set allowed seeks on a file once it hits 0.
	InitAllowedSeeks int64

	// FileNum is the file number.
	FileNum base.FileNum
	// Size is the size of the file, in bytes.
	Size uint64
	// File creation time in seconds since the epoch (1970-01-01 00:00:00
	// UTC). For ingested sstables, this corresponds to the time the file was
	// ingested.
	CreationTime int64
	// Smallest and largest sequence numbers in the table, across both point and
	// range keys.
	SmallestSeqNum uint64
	LargestSeqNum  uint64
	// SmallestPointKey and LargestPointKey are the inclusive bounds for the
	// internal point keys stored in the table. This includes RANGEDELs, which
	// alter point keys.
[...]

PR #4603, opened 2025-04-22
manifest: optimize sizeof TableMetadata

annrpom:
### manifest: use new InternalKeyBounds struct in TableMetadata's range keys
This patch introduces (and uses) a new `InternalKeyBounds` struct to represent
both the smallest, largest range keys (if they exist) in our
`TableMetadata` struct. These keys are now stored as a single string --
spliced by a separator index for individual access.

This allows us to save 24 bytes for our `TableMetadata` struct.

---

### manifest: use InternalKeyBounds struct in TableMetadata's point keys
This patch uses our `InternalKeyBounds` struct to represent
both the smallest, largest point keys in our `TableMetadata` struct.
These keys are now stored as a single string -- spliced by a separator index
for individual access.

This allows us to save 24 bytes for our `TableMetadata` struct.

---

### manifest: make internal range key bound of TableMetadata a pointer
Since range keys are rare, we make our range key bounds a pointer to
save on 32 bytes.

---

### manifest: make virtual params field of TableMetadata a pointer
This field is also rare and only used for virtual sstables.

Fixes: https://github.com/cockroachdb/pebble/issues/2047

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2609,7 +2609,7 @@ func (d *DB) runCopyCompaction(
 		SyntheticPrefixAndSuffix: inputMeta.SyntheticPrefixAndSuffix,
 	}
 	if inputMeta.HasPointKeys {
-		newMeta.ExtendPointKeyBounds(c.cmp, inputMeta.SmallestPointKey, inputMeta.LargestPointKey)
+		newMeta.ExtendPointKeyBounds(c.cmp, inputMeta.PointKeyBounds.Smallest(), inputMeta.PointKeyBounds.Largest())
 	}
 	if inputMeta.HasRangeKeys {
 		newMeta.ExtendRangeKeyBounds(c.cmp, inputMeta.RangeKeyBounds.Smallest(), inputMeta.RangeKeyBounds.Largest())
--- a/compaction.go
+++ b/compaction.go
@@ -2612,7 +2612,7 @@ func (d *DB) runCopyCompaction(
 		newMeta.ExtendPointKeyBounds(c.cmp, inputMeta.SmallestPointKey, inputMeta.LargestPointKey)
 	}
 	if inputMeta.HasRangeKeys {
-		newMeta.ExtendRangeKeyBounds(c.cmp, inputMeta.SmallestRangeKey, inputMeta.LargestRangeKey)
+		newMeta.ExtendRangeKeyBounds(c.cmp, inputMeta.RangeKeyBounds.Smallest(), inputMeta.RangeKeyBounds.Largest())
 	}
 	newMeta.FileNum = d.mu.versions.getNextFileNum()
 	if objMeta.IsExternal() {
--- a/compaction_picker_test.go
+++ b/compaction_picker_test.go
@@ -1716,6 +1716,6 @@ func checkClone(t *testing.T, pc *pickedCompaction) {
 }
 
 func checkTableBoundary(a, b *tableMetadata, cmp base.Compare) (ok bool) {
-	c := cmp(a.LargestPointKey.UserKey, b.SmallestPointKey.UserKey)
-	return c < 0 || (c == 0 && a.LargestPointKey.IsExclusiveSentinel())
+	c := cmp(a.PointKeyBounds.LargestUserKey(), b.PointKeyBounds.SmallestUserKey())
+	return c < 0 || (c == 0 && a.PointKeyBounds.Largest().IsExclusiveSentinel())
 }
--- a/data_test.go
+++ b/data_test.go
@@ -1303,7 +1303,7 @@ func runSSTablePropertiesCmd(t *testing.T, td *datadriven.TestData, d *DB) strin
 	props := r.Properties.String()
 	env := sstable.ReadEnv{}
 	if m != nil && m.Virtual {
-		env.Virtual = &m.VirtualParams
+		env.Virtual = m.VirtualParams
 		scaledProps := r.Properties.GetScaledProperties(m.FileBacking.Size, m.Size)
 		props = scaledProps.String()
 	}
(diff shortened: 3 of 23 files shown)


jbowens (commented):
Looking great! Just a couple of comments around allocations and nil checks

> Reviewed 16 of 16 files at r1, 19 of 19 files at r2, 5 of 5 files at r3, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4603#-:-OOZ2NIm4ZTmtspB9-oB:bp6ir7t)* status: all files reviewed, 3 unresolved discussions (waiting on @sumeerbhola)

___
*[`internal/manifest/version.go` line 74 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4603#-OOYyMWzD3vDK0v-AaXK:-OOYyMWzD3vDK0v-AaXL:b-ena2i6) ([raw file](https://github.com/cockroachdb/pebble/blob/c5b190b2ac9c13614bb734b9d3dc7b032dfe0bbe/internal/manifest/version.go#L74)):*
> ```Go
> 	}
> 	return ik
> }
> ```

nit: can we add a `MakeInternalKeyBounds(smallest, largest InternalKey)` func so that we can allocate the full `len(smallest.UserKey)+len(largest.UserKey)` userKeyData all at once, avoiding the intermediary allocation when we use both `SetSmallest` and `SetLargest`?
___
*[`internal/manifest/version.go` line 791 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4603#-OOZ-IyDBr9fOrNYAIv7:-OOZ-IyDBr9fOrNYAIv8:byz5jc7) ([raw file](https://github.com/cockroachdb/pebble/blob/c5b190b2ac9c13614bb734b9d3dc7b032dfe0bbe/internal/manifest/version.go#L791)):*
> ```Go
> 	case KeyTypeRange:
> 		srk := m.RangeKeyBounds.Smallest()
> 		return &srk, m.HasRangeKeys
> ```

I think this will end up allocating, because we're returning a pointer to a local struct variable. But there's only one caller of `TableMetadata.SmallestBound`/`TableMetadata.LargestBound`, and I think we could easily update these methods to return a non-pointer `InternalKey` and that'll remove the allocation. Maybe add a commit before the others that changes the return type of `SmallestBound` and `LargestBound`?
___
*[`internal/manifest/version.go` line 59 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4603#-OOZ1Lrd-iKIYzeWFq_Q:-OOZ1Lrd-iKIYzeWFq_R:b3tf4t8) ([raw file](https://github.com/cockroachdb/pebble/blob/74ecc889b06a7ea068e9031d9bc22c142b31798c/internal/manifest/version.go#L59)):*
> ```Go
> }
> 
> func (ikr *InternalKeyBounds) Smallest() InternalKey {
> ```

How feasible would it be to move these nil checks (or equivalent `m.HasRangeKeys` checks) into the callers? I'm worried about a code path ultimately trying to call `base.Compare` with an operand that's nil from one of these methods.

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4603)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/manifest/version.go` line 74 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4603#-OOYyMWzD3vDK0v-AaXK:-OOws3if0ZGm6MrtetAw:b-f1i1zn) ([raw file](https://github.com/cockroachdb/pebble/blob/c5b190b2ac9c13614bb734b9d3dc7b032dfe0bbe/internal/manifest/version.go#L74)):*

nit: can we add a `MakeInternalKeyBounds(smallest, largest InternalKey)` func so that we can allocate the full `len(smallest.UserKey)+len(largest.UserKey)` userKeyData all at once, avoiding the intermediary allocation when we use both `SetSmallest` and `SetLargest`?
</blockquote></details>

we can indeed -- done; i made a `SetInternalKeyBounds(smallest, largest InternalKey)` and used that in any places that had an intermediary allocation
___
*[`internal/manifest/version.go` line 791 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4603#-OOZ-IyDBr9fOrNYAIv7:-OOwn9131IcdwvDjoCJN:b37xfy9) ([raw file](https://github.com/cockroachdb/pebble/blob/c5b190b2ac9c13614bb734b9d3dc7b032dfe0bbe/internal/manifest/version.go#L791)):*

I think this will end up allocating, because we're returning a pointer to a local struct variable. But there's only one caller of `TableMetadata.SmallestBound`/`TableMetadata.LargestBound`, and I think we could easily update these methods to return a non-pointer `InternalKey` and that'll remove the allocation. Maybe add a commit before the others that changes the return type of `SmallestBound` and `LargestBound`?
</blockquote></details>

ah good point  -- done
___
*[`internal/manifest/version.go` line 59 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4603#-OOZ1Lrd-iKIYzeWFq_Q:-OOx9n7r66KRTf4Yx9Ly:bpyukit) ([raw file](https://github.com/cockroachdb/pebble/blob/74ecc889b06a7ea068e9031d9bc22c142b31798c/internal/manifest/version.go#L59)):*

How feasible would it be to move these nil checks (or equivalent `m.HasRangeKeys` checks) into the callers? I'm worried about a code path ultimately trying to call `base.Compare` with an operand that's nil from one of these methods.
</blockquote></details>

very feasible/done -- i had to rewrite some tests since now we don't print out the range key bounds if none exist, but lmk if we would rather just keep the format of displaying a default/invalid InternalKey

jbowens (approved):
:lgtm_strong: 

> Reviewed 1 of 19 files at r7, 7 of 11 files at r8, 3 of 4 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4603#-:-OOxkMnH0RvLtalXiSB1:b-uwkowd)* status: 11 of 23 files reviewed, all discussions resolved (waiting on @sumeerbhola)

PR #4611, opened 2025-04-23
db: test replica removal of snapshot sstable

jbowens:
In CockroachDB snapshot reception ingests sstables. In some circumstances and verisons of CockroachDB, these sstables have RANGEDEL and RANGEKEYDEL keys spanning the entirety of the KV range's span in order to clear any previous range state.

When a replica is destroyed, CockroachDB may choose to write a RANGEDEL (and possibly a RANGEKEYDEL) over the replica's span to delete the data. If an original sstable from snapshot reception still exists, we would like this sstable to be dropped by a simple deletion-only compaction. This commit adds a unit test that demonstrates this behavior.

### DIFF
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -1354,6 +1354,17 @@ func TestCompactionDeleteOnlyHints(t *testing.T) {
 	datadriven.RunTest(t, "testdata/compaction_delete_only_hints",
 		func(t *testing.T, td *datadriven.TestData) string {
 			switch td.Cmd {
+			case "batch":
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4611#-:-OOYLuOF0dFfl3x1Pct5:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

PR #4612, opened 2025-04-23
crl-release-24.1: fix deletion-only compaction bound comparison

jbowens:
Previously resolving a deletion-only compaction hints did not consider an
sstable with an end boundary that's exclusive. This prevented the hint from
being applied in circumstances where it could have been used. In particular, in
the context of CockroachDB this bug prevented the range deletions of replica
removals from removing a sstable that was ingested as a part of snapshot
reception with bounds exactly equal to the range's bounds.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2125,10 +2125,9 @@ type deleteCompactionHint struct {
 	// The type of key span that generated this hint (point key, range key, or
 	// both).
 	hintType deleteCompactionHintType
-	// start and end are user keys specifying a key range [start, end) of
-	// deleted keys.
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):
:lgtm_strong: :

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4612#-:-OOYLfKE6MAmBwHi5a__:b-yncq55)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @annrpom)

jbowens:
TFTR!

PR #4617, opened 2025-04-23
record: add sync offsets to manifest writer

EdwardX29:
Add sync offsets to manifest writer to enable read ahead and detect corruption.


xxmplus:
Close to due to age

PR #4618, opened 2025-04-23
sstable: improve writer EstimatedSize

jbowens:
When computing the estimated size of a sstable being built, incorporate size of the pending index blocks, value blocks, range deletion block and range key block. None of these quantities were incorporated into the RawRowWriter's estimated size.

The RawColumnWriter incorporated these, but did not incorporate the current pending data block or current pending index block.

Most of these quantities are uncompressed. We consider this pessimism to be acceptable, because these quantities are expected to be small most of the time. When they're not, they increase memory usage and splitting earlier than we otherwise would have helps bound memory usage.

Informs #4518.

### DIFF
--- a/sstable/colblk_writer.go
+++ b/sstable/colblk_writer.go
@@ -79,8 +79,14 @@ type RawColumnWriter struct {
 	// flushed in order after all the data blocks, and the top-level index block
 	// is constructed to point to all the individual index blocks.
 	indexBuffering struct {
-		// partitions holds all the completed index blocks.
+		// partitions holds all the completed, uncompressed index blocks.
+		//
+		// TODO(jackson): We should consider compressing these index blocks now,
+		// while buffering, to reduce the memory usage of the writer.
 		partitions []bufferedIndexBlock
+		// partitionSizeSum is the sum of the sizes of all the completed index
+		// blocks (in `partitions`).
+		partitionSizeSum uint64
 		// blockAlloc is used to bulk-allocate byte slices used to store index
 		// blocks in partitions. These live until the sstable is finished.
 		blockAlloc []byte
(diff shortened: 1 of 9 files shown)


sumeerbhola (approved):
:lgtm: 

> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4618#-:-OOdD_620ZuI6WnciVKg:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`testdata/metrics` line 108 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4618#-OOdBRFqC8Xz0BKmkaFe:-OOdBRFqC8Xz0BKmkaFf:bregjhs) ([raw file](https://github.com/cockroachdb/pebble/blob/cf46c5f513df90d68d8db7651b88ff972e50b16b/testdata/metrics#L108)):*
> ```
> ----
> 
> compact a-z
> ```

Can you add a comment like:
\# Two files, since TargetFileSize is set to 50.
___
*[`sstable/colblk_writer.go` line 83 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4618#-OOdCYaG3xiEZHGGuzTW:-OOdCYaG3xiEZHGGuzTX:beeviz3) ([raw file](https://github.com/cockroachdb/pebble/blob/cf46c5f513df90d68d8db7651b88ff972e50b16b/sstable/colblk_writer.go#L83)):*
> ```Go
> 	indexBuffering struct {
> 		// partitions holds all the completed index blocks.
> 		partitions []bufferedIndexBlock
> ```

Do you know why we don't compress a buffered index block before buffering it -- I couldn't figure out a good reason, and I wonder if we just didn't bother (unlike the value block case where we were worried about the memory footprint).
Now that we are including these uncompressed sizes it may be worth considering.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4618#-:-OOdqh-t3o7nPmaevdkA:b-814g8k)* status: 7 of 9 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`sstable/colblk_writer.go` line 83 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4618#-OOdCYaG3xiEZHGGuzTW:-OOdnhmrEwM_IOB0iD82:bnw3oq9) ([raw file](https://github.com/cockroachdb/pebble/blob/cf46c5f513df90d68d8db7651b88ff972e50b16b/sstable/colblk_writer.go#L83)):*

Do you know why we don't compress a buffered index block before buffering it -- I couldn't figure out a good reason, and I wonder if we just didn't bother (unlike the value block case where we were worried about the memory footprint).
Now that we are including these uncompressed sizes it may be worth considering.
</blockquote></details>

I can't think of any reason either. I added a TODO here
___
*[`testdata/metrics` line 108 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4618#-OOdBRFqC8Xz0BKmkaFe:-OOdnmp43lmRca8R3uUm:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/cf46c5f513df90d68d8db7651b88ff972e50b16b/testdata/metrics#L108)):*

Can you add a comment like:
\# Two files, since TargetFileSize is set to 50.
</blockquote></details>

Done

Issue #2202, opened 2022-12-22
db: optimize heaps handling of Next keys within same level


jbowens:
In a quiescent state, each level of the LSM has 10x more bytes than the previous level. As a result, higher levels tend to be sparser. When positioned over a key in L6, it's fairly likely that the next key is also within L6.

During a Next on the merging iterator, the heap root is nexted to the next key, and then the heap is fixed up. Consider the case where the next key is from the same level. `mergingIterHeap.fix(0)` needs to determine whether or not it should swap the level at the heap root with either of its heap children. It first calls Less(_left_, _right_) to determine which child is lessor, and then calls Less again to compare the root with its children, resulting in two total key comparisons. If there's a string of keys all from the same level, the  Less(_left_, _right_) comparison to determine which child is lessor is redundant.

It seems like there should be some means of remembering the knowledge of which child is lessor, and reducing the key comparisons to 1 per Next during long strings of keys originating from the same level.

I'm not sure how best to make this optimization while keeping the code general and maintainable. 

Jira issue: PEBBLE-161

jbowens:
Related article: https://grafana.com/blog/2024/04/23/the-loser-tree-data-structure-how-to-optimize-merges-and-make-your-programs-run-faster/

Internal slack thread: https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1745281925209639

PR #4623, opened 2025-04-24
db: reduce comparisons in mergingIterHeap

sumeerbhola:
The comparisons are reduced by remembering which child was the winner in the comparison between two children. This can help both in the case where the top of the heap advances, and needs to compare with the children (and wins against the children), and when pushing down due to a loss. The latter happens in two cases:

- When building the heap, since the building starts from the lowest interior nodes and proceed upwards -- a node that loses to an interior child node will be pushed down, and that interior child node already knows which of its children is the winner.
- When fixing the top that loses to one of the children and is pushed down.

To never do more comparisons than a regular heap, we allow for the case where the winning child is unknown.

With TestMergingIterHeap, one can see cmp reductions of ~40%. Sample output from some runs:

    merging_iter_heap_test.go:102: cmp needed=859 called=474(frac=0.55)
    merging_iter_heap_test.go:102: cmp needed=827 called=457(frac=0.55)
    merging_iter_heap_test.go:102: cmp needed=647 called=378(frac=0.58)
    merging_iter_heap_test.go:102: cmp needed=435 called=236(frac=0.54)
    merging_iter_heap_test.go:102: cmp needed=425 called=273(frac=0.64)
    merging_iter_heap_test.go:102: cmp needed=360 called=259(frac=0.72)
    merging_iter_heap_test.go:102: cmp needed=571 called=323(frac=0.57)
    merging_iter_heap_test.go:102: cmp needed=754 called=439(frac=0.58)

There are some low-level cons with this optimization:
- The heap element sizeof increases to 16 bytes.
- There is more branching because we have to accommodate the case of the winning chid being unknown.
- There is more writing to record the state.

The existing benchmarks use tiny 8 byte keys, so are most susceptible to these cons. Here are the benchmark results (without the cmp counting code):
                                                     │    sec/op    │    sec/op     vs base               │
MergingIterSeekGE/restart=16/count=1-10                445.4n ±  7%   439.7n ± 20%       ~ (p=0.579 n=10)
MergingIterSeekGE/restart=16/count=2-10                954.8n ± 29%   933.8n ±  3%       ~ (p=0.393 n=10)
MergingIterSeekGE/restart=16/count=3-10                1.446µ ±  3%   1.489µ ± 11%       ~ (p=0.123 n=10)
MergingIterSeekGE/restart=16/count=4-10                1.984µ ±  3%   2.008µ ±  3%       ~ (p=0.315 n=10)
MergingIterSeekGE/restart=16/count=5-10                2.574µ ±  3%   2.573µ ±  4%       ~ (p=0.724 n=10)
MergingIterNext/restart=16/count=1-10                  20.64n ±  0%   20.70n ±  0%       ~ (p=0.223 n=10)
MergingIterNext/restart=16/count=2-10                  31.46n ±  1%   31.90n ±  1%  +1.40% (p=0.000 n=10)
MergingIterNext/restart=16/count=3-10                  39.37n ±  1%   40.08n ±  1%  +1.82% (p=0.000 n=10)
MergingIterNext/restart=16/count=4-10                  44.90n ±  1%   45.70n ±  0%  +1.78% (p=0.000 n=10)
MergingIterNext/restart=16/count=5-10                  51.68n ±  2%   52.59n ±  7%  +1.77% (p=0.003 n=10)
MergingIterPrev/restart=16/count=1-10                  28.86n ±  0%   29.27n ±  1%  +1.40% (p=0.000 n=10)
MergingIterPrev/restart=16/count=2-10                  41.02n ±  0%   41.52n ±  0%  +1.21% (p=0.000 n=10)
MergingIterPrev/restart=16/count=3-10                  48.56n ±  1%   49.43n ±  1%  +1.79% (p=0.000 n=10)
MergingIterPrev/restart=16/count=4-10                  53.70n ±  2%   54.57n ±  1%  +1.62% (p=0.022 n=10)
MergingIterPrev/restart=16/count=5-10                  58.52n ±  0%   60.41n ±  1%  +3.22% (p=0.000 n=10)
MergingIterSeqSeekGEWithBounds/levelCount=5-10         215.4n ±  0%   215.6n ±  0%       ~ (p=0.285 n=10)
MergingIterSeqSeekPrefixGE/skip=1/use-next=false-10    197.4n ±  0%   197.7n ±  1%       ~ (p=0.669 n=10)
MergingIterSeqSeekPrefixGE/skip=1/use-next=true-10     100.5n ±  1%   109.8n ±  0%  +9.36% (p=0.000 n=10)
MergingIterSeqSeekPrefixGE/skip=2/use-next=false-10    197.7n ±  0%   198.8n ±  0%  +0.56% (p=0.001 n=10)
MergingIterSeqSeekPrefixGE/skip=2/use-next=true-10     100.5n ±  0%   109.9n ±  1%  +9.40% (p=0.000 n=10)
MergingIterSeqSeekPrefixGE/skip=4/use-next=false-10    198.0n ±  0%   198.6n ±  0%  +0.30% (p=0.002 n=10)
MergingIterSeqSeekPrefixGE/skip=4/use-next=true-10     101.7n ±  2%   109.9n ±  1%  +8.17% (p=0.000 n=10)
MergingIterSeqSeekPrefixGE/skip=8/use-next=false-10    199.8n ±  1%   200.1n ±  0%       ~ (p=0.322 n=10)
MergingIterSeqSeekPrefixGE/skip=8/use-next=true-10     100.4n ±  0%   109.8n ±  1%  +9.41% (p=0.000 n=10)
MergingIterSeqSeekPrefixGE/skip=16/use-next=false-10   201.4n ±  0%   202.8n ±  1%  +0.67% (p=0.000 n=10)
MergingIterSeqSeekPrefixGE/skip=16/use-next=true-10    102.2n ±  0%   112.2n ±  1%  +9.73% (p=0.000 n=10)

Informs #2202


sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4623)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`merging_iter_test.go` line 813 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4623#-OOe9TiZASnqzn9WqfBx:-OOe9TiZASnqzn9WqfBy:b-qx27nq) ([raw file](https://github.com/cockroachdb/pebble/blob/0ccb6d7c0fca38d822816a8d933efefc4f45e0ff/merging_iter_test.go#L813)):*
> ```Go
> }
> 
> // Benchmarks seeking and nexting when almost all the data is in L6.
> ```

```
                                                                 │    sec/op     │   sec/op     vs base                │
MergingIterSeekAndNextWithDominantL6AndLongKey/key-prefix=0-10       8.811m ± 1%   6.030m ± 0%  -31.56% (p=0.000 n=10)
MergingIterSeekAndNextWithDominantL6AndLongKey/key-prefix=20-10     11.735m ± 0%   7.379m ± 0%  -37.13% (p=0.000 n=10)
MergingIterSeekAndNextWithDominantL6AndLongKey/key-prefix=50-10     16.932m ± 1%   9.082m ± 0%  -46.36% (p=0.000 n=10)
MergingIterSeekAndNextWithDominantL6AndLongKey/key-prefix=100-10     24.20m ± 1%   13.08m ± 0%  -45.94% (p=0.000 n=10)
```

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4623)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`merging_iter_heap.go` line 22 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4623#-OOeF2C52pY2GFncEQZo:-OOeF2C52pY2GFncEQZp:b-j7h18e) ([raw file](https://github.com/cockroachdb/pebble/blob/4494afc5b054e3e9fc1d97196e530dce5e0cbccb/merging_iter_heap.go#L22)):*
> ```Go
> type mergingIterHeapItem struct {
> 	*mergingIterLevel
> 	winnerChild winnerChild
> ```

You can consider using a separate array for the winners rather than fattening the node.
___
*[`merging_iter_heap.go` line 59 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4623#-OOeEN-N93eg9WcISOZy:-OOeEN-N93eg9WcISOZz:b-latw5i) ([raw file](https://github.com/cockroachdb/pebble/blob/4494afc5b054e3e9fc1d97196e530dce5e0cbccb/merging_iter_heap.go#L59)):*
> ```Go
> 
> // swap is an internal method, used to swap the elements at i and j.
> func (h *mergingIterHeap) swap(i, j int) {
> ```

I would require the caller to provide new values for `winnerChild` for both, so it has to be explicit in all cases (even when the current value is ok - there should be a comment explaining why the current value is still ok). Overall the maintenance of `winnerChild` feels pretty fragile. We should have some invariant checks around it.

The code would be cleaner if we implemented it as a generic data structure that takes a `T` and a `less(T, T) bool` function.

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4623#-:-OP3Phdm7RTcUR7QjxKV:b-814g8k)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`merging_iter_heap.go` line 22 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4623#-OOeF2C52pY2GFncEQZo:-OP3Ozm_4j5IUM7hy_7C:b-thkxdm) ([raw file](https://github.com/cockroachdb/pebble/blob/4494afc5b054e3e9fc1d97196e530dce5e0cbccb/merging_iter_heap.go#L22)):*

You can consider using a separate array for the winners rather than fattening the node.
</blockquote></details>

Pushed a commit with that change with benchmark results -- did not improve.
___
*[`merging_iter_heap.go` line 59 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4623#-OOeEN-N93eg9WcISOZy:-OP3P5Sk9qj1Zyrn8x-y:b-r2y5cl) ([raw file](https://github.com/cockroachdb/pebble/blob/4494afc5b054e3e9fc1d97196e530dce5e0cbccb/merging_iter_heap.go#L59)):*

I would require the caller to provide new values for `winnerChild` for both, so it has to be explicit in all cases (even when the current value is ok - there should be a comment explaining why the current value is still ok). Overall the maintenance of `winnerChild` feels pretty fragile. We should have some invariant checks around it.

The code would be cleaner if we implemented it as a generic data structure that takes a `T` and a `less(T, T) bool` function.
</blockquote></details>

I don't really want to update winnerChild when not necessary given how performance sensitive this code is. That said, it definitely needs better comments and invariants checking, which I've added in a new commit.

Regarding generics, I didn't want to muddy the situation by changing to a generic, and more so because we don't yet have another use for this heap. I've left a todo for the future if we feel like doing it.

RaduBerinde (commented):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4623#-:-OP3cA2N3guAqiTUP1zu:b-uwkowd)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`merging_iter_heap.go` line 22 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4623#-OOeF2C52pY2GFncEQZo:-OP3c_dk8u22W_T9-WZy:b-mfl3e8) ([raw file](https://github.com/cockroachdb/pebble/blob/4494afc5b054e3e9fc1d97196e530dce5e0cbccb/merging_iter_heap.go#L22)):*

Pushed a commit with that change with benchmark results -- did not improve.
</blockquote></details>

Thanks, feel free to revert it.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4623)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`merging_iter_heap.go` line 22 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4623#-OOeF2C52pY2GFncEQZo:-OP76ZRs9QUy968kHqi3:bqtmzpf) ([raw file](https://github.com/cockroachdb/pebble/blob/4494afc5b054e3e9fc1d97196e530dce5e0cbccb/merging_iter_heap.go#L22)):*

Thanks, feel free to revert it.
</blockquote></details>

TFTR! I will close this PR. Opened a cleaned up one for proper review https://github.com/cockroachdb/pebble/pull/4663

PR #4629, opened 2025-04-25
crl-release-25.2: metrics improvements

### LINKED ISSUE #4572: db: include progress and completion indicator of table stats collection in metrics
And output it in `Metrics.String()`. The progress indicator can tell us how many old files have had their stats collected.
Without this information one is unsure whether table stats collection is complete or not (for huge LSMs), which affect the believability of other stats, and the optimality of compaction picking. This came up in the context of https://github.com/cockroachlabs/support/issues/3264.

Jira issue: PEBBLE-408

### LINKED ISSUE #4599: db: include level compensation factor in metrics

### LINKED ISSUE #4605: db: include TableStats in SSTableInfo
This will cause it to be included in `crdb_internal.sstable_metrics`, and one can aggregate the values of `PointDeletionsBytesEstimate` and `RangeDeletionsBytesEstimate`.
Came up in the context of https://github.com/cockroachlabs/support/issues/3264.

Jira issue: PEBBLE-423

### LINKED ISSUE #4624: master: nightly test failed
The nightly  test run failed on 8a142e58d8acf53d3a764115a0c2f8e57e682dc5. Please review the run [here](https://github.com/cockroachdb/pebble/actions/runs/14662084404).

Jira issue: PEBBLE-425

RaduBerinde:
#### metrics: show info about pending stats collection

Add the count of sstables that are pending stats collection, and
whether the initial loading completed.

Fixes #4572

#### metrics: improve score visibility

The current score column shows the `compensatedScoreRatio` which is
only used to determine if we should compact a level or not; the
value used in ranking levels that need compaction is
`uncompensatedScoreRatio`. Furthermore, it does not provide enough
information to understand where the value is coming from.

This commit replaces this column with the `uncompensatedScoreRatio`
actually used for ranking, except when the level doesn't need
compaction in which case we show a `-`. We also add two columns that
show the `uncompensatedScore` and `compensatedScore` which can help
when debugging.

Fixes #4599

#### db: log extreme values for deletion estimates

Check `PointDeletionsBytesEstimate` and `RangeDeletionsBytesEstimate`
for extreme values and log them (at most once every 30s). This is to
help identify the cause of very high level scores observed in a
cluster.

#### db: deflake TestMetrics

Fix #4624.

#### db: include TableStats in SSTableInfo

Fixes #4605


jbowens (approved):

RaduBerinde:
TFTR!

PR #4632, opened 2025-04-26
db: clean up level compaction scoring

RaduBerinde:
#### db: fix for TestCompactionPickerTargetLevel

In the "queue" case, we weren't updating the L0 organizer, which can
lead to panics in the test if compaction heuristics change.

#### db: clean up level compaction scoring

This commit renames "score" to "fill factor" and "score ratio" to
"score".

We no longer calculate a "compensated score ratio"; it was used only
to determine whether the level is eligible for a compaction (and not
used for ranking the levels). We now use the compensated fill factor
(note that this can result in levels being more frequently eligible
for compaction, but not the other way around).

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -6,9 +6,11 @@ package pebble
 
 import (
 	"bytes"
+	"cmp"
 	"fmt"
 	"iter"
 	"math"
+	"slices"
 	"sort"
 	"strings"
 
@@ -48,11 +50,9 @@ type compactionEnv struct {
 
 type compactionPickerMetrics struct {
 	levels [numLevels]struct {
-		// needsCompaction indicates if compensatedScoreRatio is >= compactionScoreThreshold.
-		shouldCompact           bool
-		uncompensatedScoreRatio float64
-		uncompensatedScore      float64
-		compensatedScore        float64
+		score                 float64
+		fillFactor            float64
+		compensatedFillFactor float64
 	}
 }
 
@@ -104,41 +104,6 @@ func (info compactionInfo) String() string {
 		fmt.Fprintf(&buf, " -> L%d", info.outputLevel)
 	}
 	return buf.String()
-}
-
-type sortCompactionLevelsByPriority []candidateLevelInfo
-
-func (s sortCompactionLevelsByPriority) Len() int {
-	return len(s)
-}
-
-// A level should be picked for compaction if the compensatedScoreRatio is >= the
-// compactionScoreThreshold.
-const compactionScoreThreshold = 1
-
-// Less should return true if s[i] must be placed earlier than s[j] in the final
-// sorted list. The candidateLevelInfo for the level placed earlier is more likely
-// to be picked for a compaction.
-func (s sortCompactionLevelsByPriority) Less(i, j int) bool {
-	iShouldCompact := s[i].compensatedScoreRatio >= compactionScoreThreshold
-	jShouldCompact := s[j].compensatedScoreRatio >= compactionScoreThreshold
-	// Ordering is defined as decreasing on (shouldCompact, uncompensatedScoreRatio)
-	// where shouldCompact is 1 for true and 0 for false.
-	if iShouldCompact && !jShouldCompact {
-		return true
-	}
-	if !iShouldCompact && jShouldCompact {
-		return false
-	}
-
-	if s[i].uncompensatedScoreRatio != s[j].uncompensatedScoreRatio {
-		return s[i].uncompensatedScoreRatio > s[j].uncompensatedScoreRatio
-	}
-	return s[i].level < s[j].level
-}
-
-func (s sortCompactionLevelsByPriority) Swap(i, j int) {
-	s[i], s[j] = s[j], s[i]
 }
 
 // sublevelInfo is used to tag a LevelSlice for an L0 sublevel with the
@@ -631,23 +596,31 @@ func newCompactionPickerByScore(
 // Information about a candidate compaction level that has been identified by
 // the compaction picker.
 type candidateLevelInfo struct {
-	// The compensatedScore of the level after adjusting according to the other
-	// levels' sizes. For L0, the compensatedScoreRatio is equivalent to the
-	// uncompensatedScoreRatio as we don't account for level size compensation in
-	// L0.
-	compensatedScoreRatio float64
-	// The score of the level after accounting for level size compensation before
-	// adjusting according to other levels' sizes. For L0, the compensatedScore
-	// is equivalent to the uncompensatedScore as we don't account for level
-	// size compensation in L0.
-	compensatedScore float64
-	// The score of the level to be compacted, calculated using uncompensated file
-	// sizes and without any adjustments.
-	uncompensatedScore float64
-	// uncompensatedScoreRatio is the uncompensatedScore adjusted according to
-	// the other levels' sizes.
-	uncompensatedScoreRatio float64
-	level                   int
+	// The fill factor of the level, calculated using uncompensated file sizes and
+	// without any adjustments. A factor > 1 means that the level has more data
+	// than the ideal size for that level.
+	fillFactor float64
+
+	// The score of the level, used to rank levels.
+	//
+	// If the level doesn't require compaction, the score is 0.
+	// Otherwise:
+	//  - for L6, or if the fillFactor is < 1, the score is equal to the fillFactor.
+	//  - for L0-L5 with fillFactor >= 1, the score is the ratio between the
+	//    fillFactor and the next level's fillFactor.
+	score float64
+
+	// The fill factor of the level after accounting for level size compensation (i.e.
+	// including the estimated savings in the lower levels because of deletions).
+	//
+	// For L0, the compensatedFillFactor is equal to the fillFactor as we don't
+	// account for level size compensation in L0.
+	//
+	// The compensated fill factor is used to determine if the level should be
+	// compacted (by comparing it to 1).
+	compensatedFillFactor float64
+
+	level int
 	// The level to compact to.
 	outputLevel int
 	// The file in level that will be compacted. Additional files may be
@@ -657,7 +630,7 @@ type candidateLevelInfo struct {
 }
 
 func (c *candidateLevelInfo) shouldCompact() bool {
-	return c.compensatedScoreRatio >= compactionScoreThreshold
+	return c.score > 0
 }
 
 func fileCompensation(f *tableMetadata) uint64 {
@@ -721,10 +694,9 @@ var _ compactionPicker = &compactionPickerByScore{}
 func (p *compactionPickerByScore) getMetrics(inProgress []compactionInfo) compactionPickerMetrics {
 	var m compactionPickerMetrics
 	for _, info := range p.calculateLevelScores(inProgress) {
-		m.levels[info.level].shouldCompact = info.shouldCompact()
-		m.levels[info.level].uncompensatedScoreRatio = info.uncompensatedScoreRatio
-		m.levels[info.level].uncompensatedScore = info.uncompensatedScore
-		m.levels[info.level].compensatedScore = info.compensatedScore
+		m.levels[info.level].score = info.score
+		m.levels[info.level].fillFactor = info.fillFactor
+		m.levels[info.level].compensatedFillFactor = info.compensatedFillFactor
 	}
 	return m
 }
@@ -931,6 +903,8 @@ func calculateSizeAdjust(inProgressCompactions []compactionInfo) [numLevels]leve
 	return sizeAdjust
 }
 
+// calculateLevelScores calculates the candidateLevelInfo for all levels and
+// returns them in decreasing score order.
 func (p *compactionPickerByScore) calculateLevelScores(
 	inProgressCompactions []compactionInfo,
 ) [numLevels]candidateLevelInfo {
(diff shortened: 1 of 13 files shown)


RaduBerinde:
*[`compaction_picker_test.go` line 285 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OOx_o_D50unAaJWGzUd:-OOy7fxPB0e4tOrRJu4l:b-6nru8m) ([raw file](https://github.com/cockroachdb/pebble/blob/48248cedce87bbb2c6f9be54c147e229c7ba1192/compaction_picker_test.go#L285)):*

I had to do something similar in https://github.com/cockroachdb/pebble/pull/4644/commits/d240e41b16c9580e585b9d53677b0ac6c00a2038#diff-478d649c5535daae03316defce5b966fdb110eb38f9d6a84f8f8876bfa98b33b.
Don't you need to also change `resetCompacting` to call `l0Organizer.ResetForTesting(vers)`?
</blockquote></details>

The l0Organizer should already reflect the current version, they are both returned by `loadVersion()` above. I will pull out this fix so you can rebase.

RaduBerinde:
*[`compaction_picker_test.go` line 285 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OOx_o_D50unAaJWGzUd:-OOy88jI5zwIxW8QkLJE:b-lxsgs3) ([raw file](https://github.com/cockroachdb/pebble/blob/48248cedce87bbb2c6f9be54c147e229c7ba1192/compaction_picker_test.go#L285)):*

The l0Organizer should already reflect the current version, they are both returned by `loadVersion()` above. I will pull out this fix so you can rebase.
</blockquote></details>

https://github.com/cockroachdb/pebble/pull/4645

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4632)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`compaction_picker.go` line 601 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP29AC83CRRY0JX_Eqd:-OP29AC83CRRY0JX_Eqe:b4n95hi) ([raw file](https://github.com/cockroachdb/pebble/blob/684418fae516afbda4ca830e62eada1c18cc7d5b/compaction_picker.go#L601)):*
> ```Go
> 	// The fill factor of the level, calculated using uncompensated file sizes and
> 	// without any adjustments. A factor > 1 means that the level has more data
> 	// than the ideal size for that level.
> ```

I think this needs a comment about L0 since the `fillFactor` there is based on sub-levels and not byte size.
___
*[`compaction_picker.go` line 611 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP2ASAn3IQoU3cKaQ98:-OP2ASAn3IQoU3cKaQ99:b-bq6206) ([raw file](https://github.com/cockroachdb/pebble/blob/684418fae516afbda4ca830e62eada1c18cc7d5b/compaction_picker.go#L611)):*
> ```Go
> 	//  - for L0-L5 with fillFactor >= 1, the score is the ratio between the
> 	//    fillFactor and the next level's fillFactor.
> 	score float64
> ```

We had two cases earlier for the compensatedScoreRatio (which we were using to decide eligibility for compaction):
- If the compensatedFillFactor was < 1, we made the compensatedScoreRatio equal to the compensatedFillFactor. So the compensatedFillFactor is an exact replacement for that case.
- If the compensatedFillFactor was >= 1, we did compensatedFillFactor[i]/compensatedFillFactor[i+1]. This could drop the score of Li to < 1, and make it ineligible for compaction. I am assuming the hypothesis with this PR is that the fact that we are not doing this and keeping it eligible for compactions is ok, since it will likely be prioritized lower based on the score anyway. Is that accurate?

This score part is trying to replace uncompensatedScoreRatio. Is the calculation the same when eligible for compactions, or is there some subtle change? The commentary is a bit confusing (but so was the old code).
The part of the commentary above that is confusing is that is starts by saying "level doesn't require compaction, the score is 0". Then follows by "or if the fillFactor is < 1, the score is equal to the fillFactor". But if the fillFactor < 1, it isn't eligible for compactions, so why is the score not 0?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4632)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`compaction_picker.go` line 601 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP29AC83CRRY0JX_Eqd:-OP6IVTUEEevlC5dNlGe:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/684418fae516afbda4ca830e62eada1c18cc7d5b/compaction_picker.go#L601)):*

I think this needs a comment about L0 since the `fillFactor` there is based on sub-levels and not byte size.
</blockquote></details>

Done.
___
*[`compaction_picker.go` line 611 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP2ASAn3IQoU3cKaQ98:-OP6IVwZEaWEsyScKqpk:b-6etg9u) ([raw file](https://github.com/cockroachdb/pebble/blob/684418fae516afbda4ca830e62eada1c18cc7d5b/compaction_picker.go#L611)):*
The score is 0 when the *compensated*FillFactor < 1. I added to the comment.

> I am assuming the hypothesis with this PR is that the fact that we are not doing this and keeping it eligible for compactions is ok, since it will likely be prioritized lower based on the score anyway. Is that accurate?

Yes. Also, there didn't seem to be a strong reason for preferring the existing way, it seemed more like an artifact of how we calculated the scores.

Other than the eligbility decision, the score calculation should be unchanged.

These formulas are confusing because they're kind of ad-hoc - I can't think of a principled intuition behind taking ratios of fill factors. In particular, I dislike this sudden transition where we start dividing by the next level's `fillFactor`. If our fill factor is 0.99 then the score is 0.99. If our fill factor increases a tiny bit to go over 1 and the next level's fill factor is 2, our score is suddenly ~0.5. (or conversely, if the next level's fill factor is 0.5, our score suddenly goes to 2).

sumeerbhola (changes requested):
> Reviewed 3 of 9 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4632)* status: 4 of 13 files reviewed, 8 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`compaction_picker.go` line 611 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP2ASAn3IQoU3cKaQ98:-OP6maE65PLv2nKzgzOb:b-qujqcy) ([raw file](https://github.com/cockroachdb/pebble/blob/684418fae516afbda4ca830e62eada1c18cc7d5b/compaction_picker.go#L611)):*
> These formulas are confusing because they're kind of ad-hoc ... I dislike this sudden transition where we start dividing

Agreed. Can you add some of this critique as a code comment.
Changing these is not easy since any heuristic change requires a lot of experimentation validation, but we should also not forget the cons we have already thought about with the current scheme.
___
*[`testdata/compaction_picker_concurrency` line 52 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP6pbHl185oMbbNtKYP:-OP6pbHl185oMbbNtKYQ:b-o2vfp7) ([raw file](https://github.com/cockroachdb/pebble/blob/7194e3324f2b8d951072c600dad2357e2f4e462d/testdata/compaction_picker_concurrency#L52)):*
> If the compensatedFillFactor was >= 1, we did compensatedFillFactor\[i\]/compensatedFillFactor\[i+1\]. This could drop the score of Li to < 1, and make it ineligible for compaction. I am assuming the hypothesis with this PR is that the fact that we are not doing this and keeping it eligible for compactions is ok, since it will likely be prioritized lower based on the score anyway.

Ah, this is an effect of the above. It makes me slightly nervous.
If Lbase has a high score and we are compacting out of Lbase and therefore unable to compact into Lbase, the new heuristic results in an intra-L0 compactions. Intra-L0 compactions are subjectively quite useless. I see two possibilities:
- make this change anyway, assuming that in a real setting with large L0 and very large Lbase, we will have enough non-overlapping files to permit enough compactions (due to our sub-levels logic that also causes flushes to split into multiple files), so an intra-L0 compaction won't be needed.
- restore the original logic for what is eligible.

I am slightly leaning towards the first.

@jbowens 
___
*[`compaction_picker.go` line 949 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP6ndsxEhZnkueInBtT:-OP6ndsxEhZnkueInBtU:b-jdhaas) ([raw file](https://github.com/cockroachdb/pebble/blob/7194e3324f2b8d951072c600dad2357e2f4e462d/compaction_picker.go#L949)):*
> ```Go
> 	//
> 	// The effect of this adjustment is to help prioritize compactions in lower
> 	// levels. The following example shows the scores and the fill ratios. In this
> ```

nit: s/ratios/factors/
___
*[`compaction_picker.go` line 957 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP6njt790uq0t7F0rN5:-OP6njt790uq0t7F0rN6:b9kwlno) ([raw file](https://github.com/cockroachdb/pebble/blob/7194e3324f2b8d951072c600dad2357e2f4e462d/compaction_picker.go#L957)):*
> ```Go
> 	//
> 	// Note that in the example shown there is no level size compensation so the
> 	// compensatedScore and the uncompensatedScore is the same for each level.
> ```

nit: compensatedFillFactor and fillFactor ...
___
*[`compaction_picker.go` line 969 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP6o6Ou1ArzkroU9xTA:-OP6o6Ou1ArzkroU9xTB:b-matkjk) ([raw file](https://github.com/cockroachdb/pebble/blob/7194e3324f2b8d951072c600dad2357e2f4e462d/compaction_picker.go#L969)):*
> ```Go
> 			continue
> 		}
> 		if scores[level].compensatedFillFactor < 1 {
> ```

// Score remains 0.
___
*[`db.go` line 2054 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP6ogDG1YFjK66m7RSR:-OP6ogDG1YFjK66m7RSS:b-4jw9pp) ([raw file](https://github.com/cockroachdb/pebble/blob/7194e3324f2b8d951072c600dad2357e2f4e462d/db.go#L2054)):*
> ```Go
> 		m := p.getMetrics(compactions)
> 		for level, lm := range m.levels {
> 			metrics.Levels[level].Score = 0
> ```

redundant line since we are overwriting in the next line

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4632)* status: 4 of 13 files reviewed, 8 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`testdata/compaction_picker_concurrency` line 52 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP6pbHl185oMbbNtKYP:-OP6t8R4-0uGodcRgn7Z:bx9kz5a) ([raw file](https://github.com/cockroachdb/pebble/blob/7194e3324f2b8d951072c600dad2357e2f4e462d/testdata/compaction_picker_concurrency#L52)):*

> If the compensatedFillFactor was >= 1, we did compensatedFillFactor\[i\]/compensatedFillFactor\[i+1\]. This could drop the score of Li to < 1, and make it ineligible for compaction. I am assuming the hypothesis with this PR is that the fact that we are not doing this and keeping it eligible for compactions is ok, since it will likely be prioritized lower based on the score anyway.

Ah, this is an effect of the above. It makes me slightly nervous.
If Lbase has a high score and we are compacting out of Lbase and therefore unable to compact into Lbase, the new heuristic results in an intra-L0 compactions. Intra-L0 compactions are subjectively quite useless. I see two possibilities:
- make this change anyway, assuming that in a real setting with large L0 and very large Lbase, we will have enough non-overlapping files to permit enough compactions (due to our sub-levels logic that also causes flushes to split into multiple files), so an intra-L0 compaction won't be needed.
- restore the original logic for what is eligible.

I am slightly leaning towards the first.

@jbowens
</blockquote></details>

Good point. I think I'm leaning towards the second to keep this as a cosmetic cleanup change. It might cause us more trouble to have to think about two versions, especially if we are going to change it again later.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4632)* status: 4 of 13 files reviewed, 8 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`testdata/compaction_picker_concurrency` line 52 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP6pbHl185oMbbNtKYP:-OP7Afov-nLlxB96zMuK:b-n08d0w) ([raw file](https://github.com/cockroachdb/pebble/blob/7194e3324f2b8d951072c600dad2357e2f4e462d/testdata/compaction_picker_concurrency#L52)):*
> I think I'm leaning towards the second to keep this as a cosmetic cleanup change.

+1

jbowens (approved):
:lgtm:

> Reviewed 2 of 2 files at r1, 4 of 13 files at r2, 8 of 9 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4632#-:-OP7HgYbA8xzPKJYfGUD:bnfp4nl)* status: all files reviewed, 9 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`testdata/compaction_picker_concurrency` line 52 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP6pbHl185oMbbNtKYP:-OP7HdW866hjeggw86zB:b32v57) ([raw file](https://github.com/cockroachdb/pebble/blob/7194e3324f2b8d951072c600dad2357e2f4e462d/testdata/compaction_picker_concurrency#L52)):*

> I think I'm leaning towards the second to keep this as a cosmetic cleanup change.

+1
</blockquote></details>

+1
___
*[`compaction_picker.go` line 626 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP7GnUw4zxWeMXXVCtN:-OP7GnUw4zxWeMXXVCtO:b59prb) ([raw file](https://github.com/cockroachdb/pebble/blob/7194e3324f2b8d951072c600dad2357e2f4e462d/compaction_picker.go#L626)):*
> ```Go
> 	// account for level size compensation in L0.
> 	//
> 	// For l1+, the compensatefdFillFactor takes into account the estimated
> ```

nit: `compensatedFillFactor`

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4632)* status: 4 of 13 files reviewed, 9 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`testdata/compaction_picker_concurrency` line 52 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP6pbHl185oMbbNtKYP:-OPDlHwp7xw77unEVHuz:bttuv19) ([raw file](https://github.com/cockroachdb/pebble/blob/7194e3324f2b8d951072c600dad2357e2f4e462d/testdata/compaction_picker_concurrency#L52)):*

+1
</blockquote></details>

Updated and it should now (hopefully) be equivalent to what we had before. The differences in tests are because we now actually show the score that's used for ranking (not the compensated score).

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4632)* status: 4 of 13 files reviewed, 6 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`compaction_picker.go` line 611 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP2ASAn3IQoU3cKaQ98:-OPDmLNPEBPpI8CtaP-S:bsro0ij) ([raw file](https://github.com/cockroachdb/pebble/blob/684418fae516afbda4ca830e62eada1c18cc7d5b/compaction_picker.go#L611)):*

> These formulas are confusing because they're kind of ad-hoc ... I dislike this sudden transition where we start dividing

Agreed. Can you add some of this critique as a code comment.
Changing these is not easy since any heuristic change requires a lot of experimentation validation, but we should also not forget the cons we have already thought about with the current scheme.
</blockquote></details>

Done.  I reframed it as requiring both the compensated fill factor and the compensated score to be >= 1.0. This is equivalent to what we we were doing before but feels like it makes more sense.
___
*[`compaction_picker.go` line 969 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP6o6Ou1ArzkroU9xTA:-OPDmAMjBsUgCag5Hz7q:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/7194e3324f2b8d951072c600dad2357e2f4e462d/compaction_picker.go#L969)):*

// Score remains 0.
</blockquote></details>

Done.
___
*[`db.go` line 2054 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP6ogDG1YFjK66m7RSR:-OPDmGcqEAsycJOlbGjq:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/7194e3324f2b8d951072c600dad2357e2f4e462d/db.go#L2054)):*

redundant line since we are overwriting in the next line
</blockquote></details>

Done.

sumeerbhola (approved):
:lgtm: 

> Reviewed 4 of 9 files at r3, 7 of 9 files at r5, 1 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4632#-:-OPW7LE412TfFR3oxj5f:bg7igkv)* status: all files reviewed, 8 unresolved discussions (waiting on @RaduBerinde)

___
*[`compaction_picker.go` line 977 at r7](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OPW4qqU68bhJL8joNvq:-OPW4qqU68bhJL8joNvr:b-gqex0r) ([raw file](https://github.com/cockroachdb/pebble/blob/011c2d60df1d511d06d245f68435703ce2a53b32/compaction_picker.go#L977)):*
> ```Go
> 			continue
> 		}
> 		if scores[level].compensatedFillFactor < 1.0 {
> ```

nit: how about using compactionScoreThreshold instead of 1.0, like the old code.
___
*[`testdata/compaction_picker_scores` line 44 at r7](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OPW62VS5_i4LY5nvAFG:-OPW62VS5_i4LY5nvAFH:b22hq67) ([raw file](https://github.com/cockroachdb/pebble/blob/011c2d60df1d511d06d245f68435703ce2a53b32/testdata/compaction_picker_scores#L44)):*
> ```
> scores
> ----
> L       Size   Score
> ```

can you print the ff and cff too, so we see the effect of the RANGEDEL etc.

RaduBerinde (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4632#-:-OPYKAPrBEqJ_XLqmvqx:bz5mair)* status: 8 of 13 files reviewed, 7 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`testdata/compaction_picker_scores` line 44 at r7](https://reviewable.io/reviews/cockroachdb/pebble/4632#-OPW62VS5_i4LY5nvAFG:-OPYNeMkAGHV4Qn27_PW:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/011c2d60df1d511d06d245f68435703ce2a53b32/testdata/compaction_picker_scores#L44)):*

can you print the ff and cff too, so we see the effect of the RANGEDEL etc.
</blockquote></details>

Done.

sumeerbhola (approved):
:lgtm: 

> Reviewed 4 of 5 files at r8.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4632#-:-OP_u9MK3eaFIekoA0tz:bg7igkv)* status: 12 of 13 files reviewed, 7 unresolved discussions (waiting on @RaduBerinde)

RaduBerinde:
TFTRs!

PR #4635, opened 2025-04-27
db: change MaxConcurrentCompactions() to return a range

RaduBerinde:
`MaxConcurrentCompactions()` returns an upper limit on the compaction
concurrency. Within this upper limit, Pebble decides what the
concurrency limit is (depending on L0 amplification and compaction debt).

There are cases where we want more concurrency for other reasons (like
space amp). Having a knob can be useful to tweak things in a
production environment.

This change makes `MaxConcurrentCompactions()` return a "baseline
limit" and an "upper limit". The baseline limit is the "starting"
value for the max concurrency, which Pebble can increase dynamically
up to the upper limit. Prior to this change, the (implicit) baseline
limit was always 1.

Informs https://github.com/cockroachdb/cockroach/issues/144963


RaduBerinde:
Draft of the CRDB-side changes: https://github.com/cockroachdb/cockroach/pull/145301

sumeerbhola (changes requested):
> Reviewed 1 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4635)* status: 1 of 20 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`options.go` line 902 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4635#-OOxXhw08CaF0bXwkVKs:-OOxXhw08CaF0bXwkVKt:b-aduho8) ([raw file](https://github.com/cockroachdb/pebble/blob/9802c46fb01a3bd2363870746f8dc435261215cf/options.go#L902)):*
> ```Go
> 	Merger *Merger
> 
> 	// MaxConcurrentCompactions is a range for the upper bound on the number of
> ```

This wording about a range is a bit confusing to me.

The current MaxConcurrentCompactions serves purely as an upper bound (modulo the "rough" caveats).
The new thing reads to me as a lower-bound, upper-bound pair, both inclusive.

I am guessing the reason what I am thinking of as a lower-bound, is instead phrased as another upper-bound, is due to the fact that the DB may not be able to find enough concurrent compactions to reach even the lower-bound. Perhaps we should treat that issue as orthogonal, and bring back the verbiage of `DB.GetAllowedWithoutPermission` if that helps?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4635#-:-OOy7vqY8c0BLuT6PzhB:b-814g8k)* status: 1 of 20 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`options.go` line 902 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4635#-OOxXhw08CaF0bXwkVKs:-OOy6R3nDOOykNASraKV:b-sdpmaw) ([raw file](https://github.com/cockroachdb/pebble/blob/9802c46fb01a3bd2363870746f8dc435261215cf/options.go#L902)):*

This wording about a range is a bit confusing to me.

The current MaxConcurrentCompactions serves purely as an upper bound (modulo the "rough" caveats).
The new thing reads to me as a lower-bound, upper-bound pair, both inclusive.

I am guessing the reason what I am thinking of as a lower-bound, is instead phrased as another upper-bound, is due to the fact that the DB may not be able to find enough concurrent compactions to reach even the lower-bound. Perhaps we should treat that issue as orthogonal, and bring back the verbiage of `DB.GetAllowedWithoutPermission` if that helps?
</blockquote></details>

That's right. I renamed to `CompactionConcurrencyRange` and reworded the comment.

sumeerbhola (approved):
:lgtm: 

> Reviewed 3 of 20 files at r1, 5 of 13 files at r2, 12 of 12 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4635#-:-OP10fwE1Esn93VxZTnr:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4635)* status: 10 of 20 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`compaction_picker.go` line 1271 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4635#-OP1UTTe2xrj68Z0j6fJ:-OP1UTTe2xrj68Z0j6fK:b-r8vfv8) ([raw file](https://github.com/cockroachdb/pebble/blob/9fb2d6e5b1da9ca8fb144a6e8f895bb564ec5894/compaction_picker.go#L1271)):*
> ```Go
> 		garbageFraction := float64(compactableGarbageBytes) / float64(p.dbSizeBytes)
> 		compactableGarbageCompactions =
> 			int((garbageFraction / garbageFractionLimit) * float64(upper-lower))
> ```

I resolved conflicts with your change. I made this slight change to the formula (we were interpolating between 0 and upper, now we're interpolating between lower and upper).

RaduBerinde:
TFTR!

Issue #4602, opened 2025-04-22
db: compaction concurrency should account for space amp (due to garbage)


sumeerbhola:
Compaction concurrency uses the `estimatedCompactionDebt` (and L0 sublevels). The `estimatedCompactionDebt is based on uncompensated sizes of levels (it uses `levelMaxBytes` which is also based on uncompensated sizes). The `estimatedCompactionDebt` includes the effect of write amp, but not the space amp due to garbage. A compaction concurrency of k, decided by the debt, just needs to keep the debt stable.

We can have space amp caused by garbage. We have two kinds of ways to ensure large tombstones become eligible for compactions:
- The scoring for compactions uses the compensated sizes so should ensure that the large tombtones sitting at higher levels get compacted down.
- For older versions with the old-style snapshots, elision-only compactions can ensure L6 sstable do get picked for compactions.

But since the `estimatedCompactionDebt` does not account for these, we could have a situation where the above compactions don't get to run (especially the second kind, since they are not score-based). We should add extra compaction slots, say `sum(PointDeletionsBytesEstimate + RangeDeletionsBytesEstimate)*4/(total LSM size)` (i.e., garbage equal to 25% of the LSM size adds an extra compaction slot). These compactions may not immediately get to run, but the extra concurrency will ensure that scores come down and eventually will get to run.

Motivated by https://github.com/cockroachlabs/support/issues/3264.

Jira issue: PEBBLE-422

PR #4627, opened 2025-04-25
db: add metrics for garbage due to point and range deletions

sumeerbhola:
Informs #4602

PR #4644, opened 2025-04-28
db: add option for CompactionGarbageFractionForMaxConcurrency

sumeerbhola:
This increases the compaction concurrency up to the maximum concurrency
based on the fraction of garbage. The default is 0.4, i.e., if 40% of
the DB is garbage, it will allow up to MaxConcurrentCompactions.

The assumption is that compensated level sizes used in scoring have
captured this garbage state, so the levels responsible for garbage
have scores > 1.0, so will eventually get picked for compactions.

Informs https://github.com/cockroachdb/pebble/issues/4602

### DIFF
--- a/cmd/pebble/db.go
+++ b/cmd/pebble/db.go
@@ -80,6 +80,10 @@ func newPebbleDB(dir string) DB {
 	// writing), columnar blocks are only written if explicitly opted into.
 	opts.Experimental.EnableColumnarBlocks = func() bool { return true }
 
+	// Running the tool should not start compactions due to garbage.
+	opts.Experimental.CompactionGarbageFractionForMaxConcurrency = func() float64 {
+		return -1.0
+	}
 	for i := 0; i < len(opts.Levels); i++ {
 		l := &opts.Levels[i]
 		l.BlockSize = 32 << 10       // 32 KB
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -706,6 +706,7 @@ type compactionPickerByScore struct {
 	// levelMaxBytes holds the dynamically adjusted max bytes setting for each
 	// level.
 	levelMaxBytes [numLevels]int64
+	dbSizeBytes   uint64
 }
 
 var _ compactionPicker = &compactionPickerByScore{}
@@ -828,6 +829,7 @@ func (p *compactionPickerByScore) initLevelMaxBytes(inProgressCompactions []comp
 	}
 
 	dbSize += p.vers.Levels[0].Size()
+	p.dbSizeBytes = dbSize
 	bottomLevelSize := dbSize - dbSize/uint64(p.opts.Experimental.LevelMultiplier)
 
 	curLevelSize := bottomLevelSize
@@ -1252,7 +1254,20 @@ func (p *compactionPickerByScore) getCompactionConcurrency() int {
 		compactionDebt := p.estimatedCompactionDebt(0)
 		compactionDebtCompactions = int(compactionDebt/p.opts.Experimental.CompactionDebtConcurrency) + 1
 	}
-	return max(min(maxConcurrentCompactions, max(l0ReadAmpCompactions, compactionDebtCompactions)), 1)
+	// Concurrency based on L0 read-amp and compaction debt.
+	concurrency :=
+		max(min(maxConcurrentCompactions, max(l0ReadAmpCompactions, compactionDebtCompactions)), 1)
+	unusedConcurrency := maxConcurrentCompactions - concurrency
+	garbageFractionLimit := p.opts.Experimental.CompactionGarbageFractionForMaxConcurrency()
+	if unusedConcurrency > 0 && garbageFractionLimit > 0 {
+		compactableGarbageBytes :=
+			*pointDeletionsBytesEstimateAnnotator.MultiLevelAnnotation(p.vers.Levels[:]) +
+				*rangeDeletionsBytesEstimateAnnotator.MultiLevelAnnotation(p.vers.Levels[:])
+		garbageFraction := float64(compactableGarbageBytes) / float64(p.dbSizeBytes)
+		concurrency +=
+			min(int((garbageFraction/garbageFractionLimit)*float64(unusedConcurrency)), unusedConcurrency)
+	}
+	return concurrency
 }
 
 // TODO(sumeer): remove unless someone actually finds this useful.
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -661,7 +661,7 @@ func (c *candidateLevelInfo) shouldCompact() bool {
 }
 
 func fileCompensation(f *tableMetadata) uint64 {
-	return uint64(f.Stats.PointDeletionsBytesEstimate) + f.Stats.RangeDeletionsBytesEstimate
+	return f.Stats.PointDeletionsBytesEstimate + f.Stats.RangeDeletionsBytesEstimate
 }
 
 // compensatedSize returns f's file size, inflated according to compaction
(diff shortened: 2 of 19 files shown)


sumeerbhola (commented):
First commit is from https://github.com/cockroachdb/pebble/pull/4627

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4644#-:-OOx5no_CqHQ0IrjWXe9:bz760yq)* status: 0 of 19 files reviewed, all discussions resolved (waiting on @jbowens)

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4644)* status: 0 of 19 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`compaction_picker.go` line 1257 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4644#-OOxM6pM8LHQAKfPBycx:-OOxM6pM8LHQAKfPBycy:b-gpd5o6) ([raw file](https://github.com/cockroachdb/pebble/blob/dc19922c32e971867aa14006f6f9e858b56e69ab/compaction_picker.go#L1257)):*
> ```Go
> 		compactionDebtCompactions = int(compactionDebt/p.opts.Experimental.CompactionDebtConcurrency) + 1
> 	}
> 	// Concurrency based on L0 read-amp and compaction debt.
> ```

In https://github.com/cockroachdb/pebble/pull/4635/files#diff-98a265d530e342998a426ac73709793b0de7b169d4eb8af753f246076786ae9aL1220 I refactor the ones above as "additional concurrency on top of the baseline" (which is always 1 here). I think we should do the same for the 
___
*[`compaction_picker.go` line 1268 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4644#-OOxMdjC5fhkeHlk0gCP:-OOxMdjC5fhkeHlk0gCQ:bjx2zri) ([raw file](https://github.com/cockroachdb/pebble/blob/dc19922c32e971867aa14006f6f9e858b56e69ab/compaction_picker.go#L1268)):*
> ```Go
> 		garbageFraction := float64(compactableGarbageBytes) / float64(p.dbSizeBytes)
> 		concurrency +=
> 			min(int((garbageFraction/garbageFractionLimit)*float64(unusedConcurrency)), unusedConcurrency)
> ```

It's not intuitive that the result of this heuristic depends on the other adjustments (by scaling with `unusedConcurrency`). It would be cleaner to think about if it didn't, e.g. we could just add `garbageFraction / garbageFractionLimit * (maxConcurrency - 1)` to the concurrency (and cap the result at maxConcurrency).

The `1` will be replaced with `defaultLimit` in https://github.com/cockroachdb/pebble/pull/4635/files#diff-98a265d530e342998a426ac73709793b0de7b169d4eb8af753f246076786ae9aL1220

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4644#-:-OOyI1oo0JZu4VwrEEFS:b-814g8k)* status: 0 of 19 files reviewed, all discussions resolved (waiting on @jbowens)

___
*[`compaction_picker.go` line 1257 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4644#-OOxM6pM8LHQAKfPBycx:-OOyEPblDFPpS_yKytNJ:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/dc19922c32e971867aa14006f6f9e858b56e69ab/compaction_picker.go#L1257)):*

In https://github.com/cockroachdb/pebble/pull/4635/files#diff-98a265d530e342998a426ac73709793b0de7b169d4eb8af753f246076786ae9aL1220 I refactor the ones above as "additional concurrency on top of the baseline" (which is always 1 here). I think we should do the same for the
</blockquote></details>

Done
___
*[`compaction_picker.go` line 1268 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4644#-OOxMdjC5fhkeHlk0gCP:-OOyEQ7I6cwM_S0XbHOU:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/dc19922c32e971867aa14006f6f9e858b56e69ab/compaction_picker.go#L1268)):*

It's not intuitive that the result of this heuristic depends on the other adjustments (by scaling with `unusedConcurrency`). It would be cleaner to think about if it didn't, e.g. we could just add `garbageFraction / garbageFractionLimit * (maxConcurrency - 1)` to the concurrency (and cap the result at maxConcurrency).

The `1` will be replaced with `defaultLimit` in https://github.com/cockroachdb/pebble/pull/4635/files#diff-98a265d530e342998a426ac73709793b0de7b169d4eb8af753f246076786ae9aL1220
</blockquote></details>

Done

RaduBerinde:
*[`compaction_picker.go` line 1269 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4644#-OOyWDvX6Q1SPgEFRuNh:-OOyWDvX6Q1SPgEFRuNi:b-iy6mgw) ([raw file](https://github.com/cockroachdb/pebble/blob/f7445e0103b262e41a30f6118ea7de9078f9ac2b/compaction_picker.go#L1269)):*
> ```Go
> 	}
> 	concurrencyBasedOnSignals :=
> 		 max(l0ReadAmpCompactions, compactionDebtCompactions, compactableGarbageCompactions)
> ```

Do we want the max, or do we want `max(l0ReadAmpCompactions, compactionDebtCompactions) + compactableGarbageCompactions`? AFAIU we take the max of the first two because they are inter-related

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4644)* status: 0 of 19 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`compaction_picker.go` line 1269 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4644#-OOyWDvX6Q1SPgEFRuNh:-OOzNbcn5tOlFT8eRRVK:b-cg9zbm) ([raw file](https://github.com/cockroachdb/pebble/blob/f7445e0103b262e41a30f6118ea7de9078f9ac2b/compaction_picker.go#L1269)):*

Do we want the max, or do we want `max(l0ReadAmpCompactions, compactionDebtCompactions) + compactableGarbageCompactions`? AFAIU we take the max of the first two because they are inter-related
</blockquote></details>

The older code was trying to do something akin to bridging the gap between the first two and `MaxConcurrentCompactions` with the `unusedConcurrency` logic. I think your objection to it was reasonable, and I interpreted it more broadly as keeping these entirely independent.

The concurrency permitted by each of these input signals benefits any of them. So summing some and not others seems a bit arbitrary. For example, a L0 with a high sub-level count may not always contribute much to compaction debt, so one could argue they could also be summed.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4644)* status: 0 of 19 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`compaction_picker.go` line 1269 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4644#-OOyWDvX6Q1SPgEFRuNh:-OOzVGeP9fijGZdq5J0D:b1pwcoa) ([raw file](https://github.com/cockroachdb/pebble/blob/f7445e0103b262e41a30f6118ea7de9078f9ac2b/compaction_picker.go#L1269)):*

The older code was trying to do something akin to bridging the gap between the first two and `MaxConcurrentCompactions` with the `unusedConcurrency` logic. I think your objection to it was reasonable, and I interpreted it more broadly as keeping these entirely independent.

The concurrency permitted by each of these input signals benefits any of them. So summing some and not others seems a bit arbitrary. For example, a L0 with a high sub-level count may not always contribute much to compaction debt, so one could argue they could also be summed.
</blockquote></details>

Yeah I was a bit baffled by the old code too, it felt like it should sum them up, but the comment above seems to suggest the compaction debt was something that is supposed to "kick in" after the read amp one dies down ("We tack on the compaction debt as a second signal to prevent compaction concurrency from dropping significantly right after a base compaction finishes, and before those bytes have been compacted further down the LSM.").

Anyway, in practice it shouldn't matter too much unless the max concurrency is very large.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4644#-:-OOzZ21O24jD6fyg5qoR:bnfp4nl)* status: 0 of 19 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

PR #4645, opened 2025-04-28
db: fix for TestCompactionPickerTargetLevel

RaduBerinde:
In the "queue" case, we weren't updating the L0 organizer, which can
lead to panics in the test if compaction heuristics change.

### DIFF
--- a/compaction_picker_test.go
+++ b/compaction_picker_test.go
@@ -246,10 +246,8 @@ func TestCompactionPickerTargetLevel(t *testing.T) {
 			case "queue":
 				var b strings.Builder
 				var inProgress []compactionInfo
-				printConcurrency := false
-				if d.HasArg("print-compaction-concurrency") {
-					printConcurrency = true
-				}
+				var l0InProgress []manifest.L0Compaction
+				printConcurrency := d.HasArg("print-compaction-concurrency")
 				for {
 					env := compactionEnv{
 						diskAvailBytes:          math.MaxUint64,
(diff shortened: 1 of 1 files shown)


sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4645)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`compaction_picker_test.go` line 285 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4645#-OP10mXu07blgF6_Cy6P:-OP10mXvBkh7K9ZauyvB:b-yd0vrr) ([raw file](https://github.com/cockroachdb/pebble/blob/bc796a35aa37545d4cc3e5fa774ad94acfefd248/compaction_picker_test.go#L285)):*
> ```Go
> 						}
> 					}
> 					l0Organizer.InitCompactingFileInfo(l0InProgress)
> ```

I may have partially fixed this, but perhaps the fix was not right since it used `l0Organizer.UpdateStateForStartedCompaction`.

The fix in this PR doesn't call `L0Organizer.ResetForTesting` -- isn't that needed?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4645)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`compaction_picker_test.go` line 285 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4645#-OP10mXu07blgF6_Cy6P:-OP1M0PID-syVmlKq7gb:b-cn8rg3) ([raw file](https://github.com/cockroachdb/pebble/blob/bc796a35aa37545d4cc3e5fa774ad94acfefd248/compaction_picker_test.go#L285)):*

I may have partially fixed this, but perhaps the fix was not right since it used `l0Organizer.UpdateStateForStartedCompaction`.

The fix in this PR doesn't call `L0Organizer.ResetForTesting` -- isn't that needed?
</blockquote></details>

`InitCompactingFileInfo` resets the compaction state in the `L0Organizer` and reinitializes it using the in-progress L0 compactions. `UpdateStateForStartedCompaction` updates existing state with "new" compactions, so it makes sense you needed the ResetForTesting. I will close this PR since it's no longer necessary.

PR #4653, opened 2025-04-29
db: add TestWALHardCrashRandomized

### LINKED ISSUE #4651: db: add randomized testing of new wal sync offset
null

Jira issue: PEBBLE-428

jbowens:
Add a new randomized unit test adapted from TestWALFailoverRandomized that stresses recovery from a hard crash.

Fix #4651.

### DIFF
--- a/open_test.go
+++ b/open_test.go
@@ -1586,7 +1586,7 @@ func TestMkdirAllAndSyncParents(t *testing.T) {
 	})
 }
 
-// TestWALFailoverRandomized is a randomzied test exercising recovery in the
+// TestWALFailoverRandomized is a randomized test exercising recovery in the
 // presence of WAL failover. It repeatedly opens a database, writes a number of
 // batches concurrently and simulates a hard crash using vfs.NewCrashableMem. It
 // ensures that the resulting DB state opens successfully, and the contents of
@@ -1596,7 +1596,6 @@ func TestMkdirAllAndSyncParents(t *testing.T) {
 func TestWALFailoverRandomized(t *testing.T) {
 	seed := time.Now().UnixNano()
 	t.Logf("seed %d", seed)
-	mem := vfs.NewCrashableMem()
 	makeOptions := func(mem *vfs.MemFS) *Options {
 		failoverOpts := WALFailoverOptions{
 			Secondary: wal.Dir{FS: mem, Dirname: "secondary"},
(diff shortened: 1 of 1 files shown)


sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4653)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`open_test.go` line 1802 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4653#-OP6kppA6I4QaIIMqwbJ:-OP6kppA6I4QaIIMqwbK:b-e7cevb) ([raw file](https://github.com/cockroachdb/pebble/blob/9236bb4a95cf744acdf082700672d7ce75f3c4df/open_test.go#L1802)):*
> ```Go
> 			// The configuration options are randomized to exercise different failure
> 			// scenarios. Some runs result in sufficient number of memtable rotations
> 			// between crashes that we do get recycled logs.
> ```

Is it worth injecting a tiny bit of latency in the FS, so that the `Batch.Commit` doesn't always complete before the crash? I suppose the `Batch.Commit` happening on a separate goroutine should result in some racing, but maybe worth increasing the probability of concurrent actions.

jbowens (commented):
I made it 25% shorter, and made each test run it 4 times. I wanted to make sure the op count didn't get too low.

TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4653#-:-OP7K6Vg9MSr5WH-w8TP:b-kgdnmv)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`open_test.go` line 1780 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4653#-OP6CrCR7cwlxx7NuSU1:-OP7K9Ir4TIjQEsATZS7:b-e4zi0e) ([raw file](https://github.com/cockroachdb/pebble/blob/52806de2e29145da41f0c1fee52ac7847368db75/open_test.go#L1780)):*

I tried this under stress and ran into a panic on "binary.BigEndian.PutUint16(v\[:\], j)" pretty quickly (`index out of range [1] with length 1`). I guess valuesize should be at least 2?
</blockquote></details>

Thanks, I only stressed it for about a minute before :smiling_face_with_tear: 
___
*[`open_test.go` line 1802 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4653#-OP6kppA6I4QaIIMqwbJ:-OP7KGco6YFLf-IA_hZB:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/9236bb4a95cf744acdf082700672d7ce75f3c4df/open_test.go#L1802)):*

Is it worth injecting a tiny bit of latency in the FS, so that the `Batch.Commit` doesn't always complete before the crash? I suppose the `Batch.Commit` happening on a separate goroutine should result in some racing, but maybe worth increasing the probability of concurrent actions.
</blockquote></details>

Done.

PR #4655, opened 2025-04-29
crl-release-25.2: db: change MaxConcurrentCompactions() to return a range

RaduBerinde:
`MaxConcurrentCompactions()` returns an upper limit on the compaction
concurrency. Within this upper limit, Pebble decides what the
concurrency limit is (depending on L0 amplification and compaction debt).

There are cases where we want more concurrency for other reasons (like
space amp). Having a knob can be useful to tweak things in a
production environment.

This change changes `MaxConcurrentCompactions()` to
`CompactionConcurrencyRange()` which returns both a lower and upper
value. The lower limit is the "baseline" value for the concurrency,
which Pebble can increase dynamically up to the upper limit. Prior to
this change, the (implicit) lower limit was always 1.

### DIFF
--- a/cmd/pebble/db.go
+++ b/cmd/pebble/db.go
@@ -72,8 +72,8 @@ func newPebbleDB(dir string) DB {
 		Merger: &pebble.Merger{
 			Name: "cockroach_merge_operator",
 		},
-		MaxConcurrentCompactions: func() int {
-			return 3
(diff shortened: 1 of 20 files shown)


jbowens (approved):

RaduBerinde:
TFTR!

Issue #4595, opened 2025-04-21
db: preserve blob reference metadata across excise


jbowens:
When an excise (see excise.go) span overlaps a sstable, we construct up to 2 sstables (one on the left side and one on the right side) to represent the remaining spans of the sstable. These new tables are considered 'virtual sstables,' but each has their own `manifest.TableMetadata`.

Today, the `TableMetadata`'s blob-related metadata is _not_ updated across an ingest split.

We should copy all the excised table's `BlobReferences` and the table's `BlobReferenceDepth` to the new tables we create.

There's a complication that the `BlobReference.ValueSize` is the value size of the underlying sstable, not the blob file. I think we should interpolate the `ValueSize`s based on the `TableMetadata.Size`s.

Suppose we had a 100MB table that gets excises into two: the left with an estimated `TableMetadata.Size` of 2MB and the right with an estimated `TableMetadata.Size` of 10MB. Both the left and right would get a `BlobReference` to every blob file the original table referenced, but all the left table's reference's `ValueSize`s would be scaled to 2% of their original values and all the right table's reference's `ValueSize`s would be scaled to 10% of their original values.

We should also update the `BlobReference.ValueSize` comment to highlight that if any of the referencing tables are virtualized tables, the `ValueSize` may be approximate.

Jira issue: PEBBLE-418

Epic CRDB-20379

PR #4660, opened 2025-04-30
db: preserve blob reference metadata across excise

annrpom:
During excise, we now preserve the top-level blob references
across virtual ssts created. The value size of each blob
reference is approximated by the proportion of the virtual table's
size to the original table's size.

Fixes: #4595

### DIFF
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -1004,6 +1004,13 @@ func TestCompaction(t *testing.T) {
 				}
 				return s
 
+			case "excise-dryrun":
+				ve, err := runExciseDryRunCmd(td, d)
+				if err != nil {
+					td.Fatalf(t, err.Error())
+				}
+				return fmt.Sprintf("would excise %d files.\n%s", len(ve.DeletedTables), ve.DebugString(base.DefaultFormatter))
+
 			case "file-sizes":
 				return runTableFileSizesCmd(td, d)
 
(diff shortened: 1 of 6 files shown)


annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4660)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`internal/manifest/blob_metadata.go` line 30 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4660#-OP7TJnU-8fH0sR-Hxi5:-OP7VxiNC6btTfzI0dGo:b-26ur74) ([raw file](https://github.com/cockroachdb/pebble/blob/a17268ed04bfbe1997866e3ef3f40d1fd22aa5ee/internal/manifest/blob_metadata.go#L30)):*

nit: rogue `)`
</blockquote></details>

eliminated T_T
___
*[`data_test.go` line 1443 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4660#-OP7TVdMEG5dmnsL-UoD:-OP7W-vBBKe37BMDatTd:b42gjdy) ([raw file](https://github.com/cockroachdb/pebble/blob/a17268ed04bfbe1997866e3ef3f40d1fd22aa5ee/data_test.go#L1443)):*

let's move this up and into a deferred anonymous function so that if we hit the early return err case we still unlock
</blockquote></details>

done good point

annrpom:
TFTR! ('-')7

PR #4663, opened 2025-04-30
db: reduce comparisons in mergingIterHeap

### LINKED ISSUE #2202: db: optimize heaps handling of Next keys within same level

sumeerbhola:
The comparisons are reduced by remembering which child was the winner in the comparison between two children. This can help both in the case where the top of the heap advances, and needs to compare with the children (and wins against the children), and when pushing down due to a loss. The latter happens in two cases:

- When building the heap, since the building starts from the lowest interior nodes and proceed upwards -- a node that loses to an interior child node will be pushed down, and that interior child node already knows which of its children is the winner. TestMergingIterHeapInit shows that this can save ~3.7% of the comparisons.
- When fixing the top that loses to one of the children and is pushed down.

To never do more comparisons than a regular heap, we allow for the case where the winning child is unknown.

With TestMergingIterHeap, one can see cmp reductions of ~40%.

There are some low-level cons with this optimization:
- The heap element sizeof increases to 16 bytes.
- There is more branching because we have to accommodate the case of the winning chid being unknown.
- There is more writing to record the state.

The existing benchmarks use tiny 8 byte keys, so are more susceptible to these cons, and don't show any improvement. In contrast, the new benchmark MergingIterSeekAndNextWithDominantL6AndLongKey shows an improvement even with tiny keys since L6 stays at the top of the heap, and shows an improvement of 45% with longer keys.

                                                               │     sec/op      │    sec/op     vs base                │
MergingIterSeekGE/restart=16/count=1-10                              463.4n ± 8%   464.1n ±  4%        ~ (p=0.542 n=10)
MergingIterSeekGE/restart=16/count=2-10                              939.7n ± 2%   991.2n ±  4%   +5.49% (p=0.000 n=10)
MergingIterSeekGE/restart=16/count=3-10                              1.604µ ± 3%   1.512µ ±  7%        ~ (p=0.138 n=10)
MergingIterSeekGE/restart=16/count=4-10                              2.101µ ± 5%   2.126µ ±  3%        ~ (p=0.739 n=10)
MergingIterSeekGE/restart=16/count=5-10                              2.726µ ± 2%   2.864µ ±  8%   +5.04% (p=0.034 n=10)
MergingIterNext/restart=16/count=1-10                                21.07n ± 4%   20.72n ±  2%   -1.68% (p=0.012 n=10)
MergingIterNext/restart=16/count=2-10                                31.69n ± 0%   31.95n ±  0%   +0.82% (p=0.001 n=10)
MergingIterNext/restart=16/count=3-10                                40.27n ± 3%   40.35n ±  1%        ~ (p=0.754 n=10)
MergingIterNext/restart=16/count=4-10                                45.36n ± 1%   45.86n ±  1%   +1.11% (p=0.001 n=10)
MergingIterNext/restart=16/count=5-10                                52.72n ± 1%   52.79n ±  1%        ~ (p=0.447 n=10)
MergingIterPrev/restart=16/count=1-10                                29.41n ± 0%   29.02n ±  1%   -1.34% (p=0.001 n=10)
MergingIterPrev/restart=16/count=2-10                                41.59n ± 2%   41.56n ±  1%        ~ (p=0.726 n=10)
MergingIterPrev/restart=16/count=3-10                                49.33n ± 1%   49.75n ±  3%   +0.85% (p=0.019 n=10)
MergingIterPrev/restart=16/count=4-10                                54.37n ± 1%   55.01n ±  1%   +1.18% (p=0.000 n=10)
MergingIterPrev/restart=16/count=5-10                                59.40n ± 1%   60.77n ±  3%   +2.31% (p=0.000 n=10)
MergingIterSeqSeekGEWithBounds/levelCount=5-10                       452.6n ± 1%   451.2n ±  0%   -0.30% (p=0.045 n=10)
MergingIterSeqSeekPrefixGE/skip=1/use-next=false-10                  358.1n ± 2%   363.4n ±  2%        ~ (p=0.075 n=10)
MergingIterSeqSeekPrefixGE/skip=1/use-next=true-10                   227.4n ± 2%   228.9n ±  0%        ~ (p=0.109 n=10)
MergingIterSeqSeekPrefixGE/skip=2/use-next=false-10                  358.6n ± 0%   364.7n ±  2%   +1.73% (p=0.009 n=10)
MergingIterSeqSeekPrefixGE/skip=2/use-next=true-10                   226.9n ± 0%   228.4n ±  4%   +0.64% (p=0.002 n=10)
MergingIterSeqSeekPrefixGE/skip=4/use-next=false-10                  358.8n ± 1%   360.4n ±  0%   +0.43% (p=0.022 n=10)
MergingIterSeqSeekPrefixGE/skip=4/use-next=true-10                   226.9n ± 0%   231.5n ±  8%   +2.05% (p=0.000 n=10)
MergingIterSeqSeekPrefixGE/skip=8/use-next=false-10                  360.1n ± 0%   369.2n ±  1%   +2.51% (p=0.000 n=10)
MergingIterSeqSeekPrefixGE/skip=8/use-next=true-10                   227.5n ± 1%   233.9n ±  2%   +2.79% (p=0.000 n=10)
MergingIterSeqSeekPrefixGE/skip=16/use-next=false-10                 373.4n ± 3%   370.2n ±  1%        ~ (p=0.470 n=10)
MergingIterSeqSeekPrefixGE/skip=16/use-next=true-10                  231.2n ± 1%   234.3n ±  1%   +1.32% (p=0.007 n=10)

MergingIterSeekAndNextWithDominantL6AndLongKey/key-prefix=0-10       8.811m ± 1%   6.030m ± 0%  -31.56% (p=0.000 n=10)
MergingIterSeekAndNextWithDominantL6AndLongKey/key-prefix=20-10     11.735m ± 0%   7.379m ± 0%  -37.13% (p=0.000 n=10)
MergingIterSeekAndNextWithDominantL6AndLongKey/key-prefix=50-10     16.932m ± 1%   9.082m ± 0%  -46.36% (p=0.000 n=10)
MergingIterSeekAndNextWithDominantL6AndLongKey/key-prefix=100-10     24.20m ± 1%   13.08m ± 0%  -45.94% (p=0.000 n=10)

Fixes #2202

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -2413,7 +2413,7 @@ func (i *Iterator) Close() error {
 			keyBuf               []byte
 			boundsBuf            [2][]byte
 			prefixOrFullSeekKey  []byte
-			mergingIterHeapItems []*mergingIterLevel
+			mergingIterHeapItems []mergingIterHeapItem
 		)
 
 		// Avoid caching the key buf if it is overly large. The constant is fairly
--- a/iterator_test.go
+++ b/iterator_test.go
@@ -1726,7 +1726,7 @@ func setupForTwoLevelBloomTombstone(b *testing.B, keyOffset int) twoLevelBloomTo
 					levels = 1
 				}
 				readers[index], levelSlices[index], keys = buildLevelsForMergingIterSeqSeek(
-					b, ch, blockSize, restartInterval, levels, keyOffset, withTombstone, bloom, twoLevelIndex)
+					b, ch, blockSize, restartInterval, levels, nil, keyOffset, withTombstone, bloom, twoLevelIndex)
 			}
 		}
 	}
@@ -1892,7 +1892,7 @@ func BenchmarkIteratorSeqSeekGEWithBounds(b *testing.B) {
 				ch := c.NewHandle()
 				defer ch.Close()
 				readers, levelSlices, keys := buildLevelsForMergingIterSeqSeek(
-					b, ch, blockSize, restartInterval, levelCount, 0, /* keyOffset */
+					b, ch, blockSize, restartInterval, levelCount, nil, 0, /* keyOffset */
 					false, false, twoLevelIndex)
 				m := buildMergingIter(readers, levelSlices)
 				iter := Iterator{
@@ -1935,7 +1935,7 @@ func BenchmarkIteratorSeekGENoop(b *testing.B) {
 	ch := c.NewHandle()
 	defer ch.Close()
 	readers, levelSlices, _ := buildLevelsForMergingIterSeqSeek(
-		b, ch, blockSize, restartInterval, levelCount, keyOffset, false, false, false)
+		b, ch, blockSize, restartInterval, levelCount, nil, keyOffset, false, false, false)
 	var keys [][]byte
 	for i := 0; i < keyOffset; i++ {
 		keys = append(keys, []byte(fmt.Sprintf("%08d", i)))
--- a/merging_iter.go
+++ b/merging_iter.go
@@ -314,7 +314,7 @@ func (m *mergingIter) init(
 	m.split = split
 	m.stats = stats
 	if cap(m.heap.items) < len(levels) {
-		m.heap.items = make([]*mergingIterLevel, 0, len(levels))
+		m.heap.items = make([]mergingIterHeapItem, 0, len(levels))
 	} else {
 		m.heap.items = m.heap.items[:0]
 	}
@@ -327,7 +327,7 @@ func (m *mergingIter) initHeap() {
 	m.heap.items = m.heap.items[:0]
 	for i := range m.levels {
 		if l := &m.levels[i]; l.iterKV != nil {
-			m.heap.items = append(m.heap.items, l)
+			m.heap.items = append(m.heap.items, mergingIterHeapItem{mergingIterLevel: l})
 		}
 	}
 	m.heap.init()
@@ -419,7 +419,7 @@ func (m *mergingIter) switchToMinHeap() error {
 	// iteration, we want to return a key that is greater than a:2.
 
 	key := m.heap.items[0].iterKV.K
-	cur := m.heap.items[0]
+	cur := m.heap.items[0].mergingIterLevel
 
 	for i := range m.levels {
 		l := &m.levels[i]
@@ -471,7 +471,7 @@ func (m *mergingIter) switchToMaxHeap() error {
 	// The current key is b:2 and i2 is pointing at b:1. When we switch to
 	// reverse iteration, we want to return a key that is less than b:2.
 	key := m.heap.items[0].iterKV.K
-	cur := m.heap.items[0]
+	cur := m.heap.items[0].mergingIterLevel
 
 	for i := range m.levels {
 		l := &m.levels[i]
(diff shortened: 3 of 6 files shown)


RaduBerinde (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4663#-:-OPBtF4a9GWDAwKg6Ef9:b-jbsf8z)* status: 0 of 6 files reviewed, 1 unresolved discussion

___
*[`merging_iter_heap.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4663#-OPBsxpu-euOkR1J7b8F:-OPBsxpu-euOkR1J7b8G:bmgmj2l) ([raw file](https://github.com/cockroachdb/pebble/blob/cace8e6b84bf6fd24bc80254b0b761727ddfcd44/merging_iter_heap.go#L26)):*
> ```Go
> }
> 
> type winnerChild uint8
> ```

[nit] a comment would help. Mention that any value is acceptable for a node with < 2 children

jbowens (approved):
:lgtm:

> Reviewed 5 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4663#-:-OPHgN4aERQN-OJkV8SA:bnfp4nl)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

sumeerbhola (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4663#-:-OPVtclI1sd8sDDoiqPf:bz5mair)* status: 4 of 6 files reviewed, all discussions resolved (waiting on @jbowens and @RaduBerinde)

___
*[`merging_iter_heap.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4663#-OPBsxpu-euOkR1J7b8F:-OPVtbP7CZ3Y_VAFwbe0:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/cace8e6b84bf6fd24bc80254b0b761727ddfcd44/merging_iter_heap.go#L26)):*

[nit] a comment would help. Mention that any value is acceptable for a node with < 2 children
</blockquote></details>

Done

PR #4664, opened 2025-04-30
colblk: add more bound checks

RaduBerinde:
Add invariants-only bound checking, especially before accessing
`UnsafeOffset`s. We also change to using `At2()` instead of two `At()`
calls in a few places.

### DIFF
--- a/internal/invariants/off.go
+++ b/internal/invariants/off.go
@@ -49,3 +49,7 @@ func (*Value[V]) Get() V {
 
 // Set the value; no-op in non-invariant builds.
 func (*Value[V]) Set(v V) {}
+
+// CheckBounds panics if the index is not in the range [0, n). No-op in
+// non-invariant builds.
+func CheckBounds(i int, n int) {}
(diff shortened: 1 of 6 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4664#-:-OP7XFwM2XqDFIJBXF2G:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/invariants/off.go` line 55 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4664#-OP7WmDW1i6VNSq-dYTL:-OP7WmDW1i6VNSq-dYTM:bb7lbs2) ([raw file](https://github.com/cockroachdb/pebble/blob/1d2a1a0c0b66896fde67f1f47d960f5f1a97e7de/internal/invariants/off.go#L55)):*
> ```Go
> // CheckBounds panics if the index is not in the range [0, n). No-op in
> // non-invariant builds.
> func CheckBounds(i int, n int) {}
> ```

nice

RaduBerinde:
TFTR!

PR #4667, opened 2025-04-30
crl-release-25.2: colblk: add more bound checks

RaduBerinde:
Add invariants-only bound checking, especially before accessing
`UnsafeOffset`s. We also change to using `At2()` instead of two `At()`
calls in a few places.


jbowens (approved):

RaduBerinde:
TFTR!

PR #4669, opened 2025-05-01
manifest: ensure TableMetadata.{Smallest,Largest} inline

jbowens:
In a Pebble ycsb/C benchmark, 1.83% of CPU was spent in (*TableMetadata).Largest. These functions were just barely over the budget to be inlined. This commit refactors these functions to ensure they're inlined and adds gcassert assertions to ensure they remain inlined.

### DIFF
--- a/internal/manifest/l0_sublevels.go
+++ b/internal/manifest/l0_sublevels.go
@@ -728,24 +728,26 @@ func (s *l0Sublevels) InitCompactingFileInfo(inProgress []L0Compaction) {
 
 	for f := range s.levelMetadata.All() {
 		if invariants.Enabled {
-			if !bytes.Equal(s.orderedIntervals[f.minIntervalIndex].startKey.key, f.Smallest().UserKey) {
+			bounds := f.UserKeyBounds()
+			if !bytes.Equal(s.orderedIntervals[f.minIntervalIndex].startKey.key, bounds.Start) {
 				panic(fmt.Sprintf("f.minIntervalIndex in TableMetadata out of sync with intervals in L0Sublevels: %s != %s",
-					s.formatKey(s.orderedIntervals[f.minIntervalIndex].startKey.key), s.formatKey(f.Smallest().UserKey)))
+					s.formatKey(s.orderedIntervals[f.minIntervalIndex].startKey.key), s.formatKey(bounds.Start)))
 			}
-			if !bytes.Equal(s.orderedIntervals[f.maxIntervalIndex+1].startKey.key, f.Largest().UserKey) {
+			if !bytes.Equal(s.orderedIntervals[f.maxIntervalIndex+1].startKey.key, bounds.End.Key) {
 				panic(fmt.Sprintf("f.maxIntervalIndex in TableMetadata out of sync with intervals in L0Sublevels: %s != %s",
-					s.formatKey(s.orderedIntervals[f.maxIntervalIndex+1].startKey.key), s.formatKey(f.Smallest().UserKey)))
+					s.formatKey(s.orderedIntervals[f.maxIntervalIndex+1].startKey.key), s.formatKey(bounds.Start)))
 			}
 		}
 		if !f.IsCompacting() {
 			continue
 		}
 		if invariants.Enabled {
-			if s.cmp(s.orderedIntervals[f.minIntervalIndex].startKey.key, f.Smallest().UserKey) != 0 ||
-				s.cmp(s.orderedIntervals[f.maxIntervalIndex+1].startKey.key, f.Largest().UserKey) != 0 {
+			bounds := f.UserKeyBounds()
+			if s.cmp(s.orderedIntervals[f.minIntervalIndex].startKey.key, bounds.Start) != 0 ||
+				s.cmp(s.orderedIntervals[f.maxIntervalIndex+1].startKey.key, bounds.End.Key) != 0 {
 				panic(fmt.Sprintf("file %s has inconsistent L0 Sublevel interval bounds: %s-%s, %s-%s", f.FileNum,
 					s.orderedIntervals[f.minIntervalIndex].startKey.key, s.orderedIntervals[f.maxIntervalIndex+1].startKey.key,
-					f.Smallest().UserKey, f.Largest().UserKey))
+					bounds.Start, bounds.End.Key))
 			}
 		}
 		for i := f.minIntervalIndex; i <= f.maxIntervalIndex; i++ {
--- a/internal/manifest/version.go
+++ b/internal/manifest/version.go
@@ -56,10 +56,12 @@ func (ikr *InternalKeyBounds) SetInternalKeyBounds(smallest, largest InternalKey
 	ikr.userKeySeparatorIdx = len(smallest.UserKey)
 }
 
+//gcassert:inline
 func (ikr *InternalKeyBounds) SmallestUserKey() []byte {
 	return unsafe.Slice(unsafe.StringData(ikr.userKeyData), ikr.userKeySeparatorIdx)
 }
 
+//gcassert:inline
 func (ikr *InternalKeyBounds) Smallest() InternalKey {
 	return InternalKey{
 		UserKey: ikr.SmallestUserKey(),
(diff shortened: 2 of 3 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4669)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @annrpom and @sumeerbhola)

___
*[`internal/manifest/version.go` line 1206 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4669#-OPBCWeW4hVR03cb9gki:-OPBCWeW4hVR03cb9gkj:b-585ulv) ([raw file](https://github.com/cockroachdb/pebble/blob/4dfedb5c2553de13a31ec408e0564a06b2ca4ef6/internal/manifest/version.go#L1206)):*
> ```Go
> func (m *TableMetadata) Smallest() InternalKey {
> 	if m.boundTypeSmallest == boundTypeRangeKey {
> 		return m.RangeKeyBounds.Smallest()
> ```

You can consider manually inlining `Smallest()` all the way (including what it calls). It might save a bit of bookkeeping data around the code (the program is able to produce the stack traces if a panic hits deep down).

You can also save some code by determining the `*InternalKeyBounds`, and then calling `Smallest()/Largest()` once on the result.

We could store the bounds in a `[2]InternalKeyBounds` and index it by the bound type but that might be overkill.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4669)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @annrpom, @RaduBerinde, and @sumeerbhola)

___
*[`internal/manifest/version.go` line 1206 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4669#-OPBCWeW4hVR03cb9gki:-OPBHA9GCwCwkRKmry5R:b-9pdu88) ([raw file](https://github.com/cockroachdb/pebble/blob/4dfedb5c2553de13a31ec408e0564a06b2ca4ef6/internal/manifest/version.go#L1206)):*
I don't quite follow what you mean by bookkeeping data? Isn't it better if Go will maintain the inlining for us?

> You can also save some code by determining the `*InternalKeyBounds`, and then calling `Smallest()/Largest()` once on the result.

Some instances do a variant of this already. The one that shows up in profiles only needs one bound (binary searching among the dividers of a B-Tree).

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4669)* status: all files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/manifest/l0_sublevels.go` line 738 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4669#-OPBJ4-mB8kRFnkhjR4_:-OPBJ4-mB8kRFnkhjR4a:b2ir6kj) ([raw file](https://github.com/cockroachdb/pebble/blob/4dfedb5c2553de13a31ec408e0564a06b2ca4ef6/internal/manifest/l0_sublevels.go#L738)):*
> ```Go
> 			if !bytes.Equal(s.orderedIntervals[f.maxIntervalIndex+1].startKey.key, bounds.End.Key) {
> 				panic(fmt.Sprintf("f.maxIntervalIndex in TableMetadata out of sync with intervals in L0Sublevels: %s != %s",
> 					s.formatKey(s.orderedIntervals[f.maxIntervalIndex+1].startKey.key), s.formatKey(bounds.End.Key)))
> ```

should this be  `bounds.Start.Key`

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4669)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @annrpom, @RaduBerinde, and @sumeerbhola)

___
*[`internal/manifest/version.go` line 1206 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4669#-OPBCWeW4hVR03cb9gki:-OPBMB2VDcmfrClNykYr:b-f86qlz) ([raw file](https://github.com/cockroachdb/pebble/blob/4dfedb5c2553de13a31ec408e0564a06b2ca4ef6/internal/manifest/version.go#L1206)):*

I don't quite follow what you mean by bookkeeping data? Isn't it better if Go will maintain the inlining for us?

> You can also save some code by determining the `*InternalKeyBounds`, and then calling `Smallest()/Largest()` once on the result.

Some instances do a variant of this already. The one that shows up in profiles only needs one bound (binary searching among the dividers of a B-Tree).
</blockquote></details>

I'll look at specializing instances of `LevelIterator.seek`, and we might get an additional benefit there. That'd be a benefit beyond the regression addressed with this PR, so I'll do that as a follow up.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4669)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @annrpom, @jbowens, and @sumeerbhola)

___
*[`internal/manifest/version.go` line 1206 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4669#-OPBCWeW4hVR03cb9gki:-OPBRE0d28mFywDoUAZQ:bdxat0g) ([raw file](https://github.com/cockroachdb/pebble/blob/4dfedb5c2553de13a31ec408e0564a06b2ca4ef6/internal/manifest/version.go#L1206)):*

I'll look at specializing instances of `LevelIterator.seek`, and we might get an additional benefit there. That'd be a benefit beyond the regression addressed with this PR, so I'll do that as a follow up.
</blockquote></details>

I meant inside this function, instead of inlining `Smallest()` twice, we could do:
```
x := &m.PointKeyBounds
if .. { x = &m.RangeKeyBounds }
return x.Smallest()
```

Unless the compiler is smart enough to do something like this already, it should reduce the size of the function (and make any calling function more likely to be inlined itself)


As for the bookkeeping, I think I was wrong, it probably doesn't make a difference. My line of thinking was that if a() inlines b() which inlines c() and there's a panic inside c(), *something* somewhere has to remember that the inlined code for c() should report the `c()` stack and the `b()` stack, not just the `a()` stack. But I don't think these tables or whatever they are matter unless there really is a panic.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4669)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @annrpom, @RaduBerinde, and @sumeerbhola)

___
*[`internal/manifest/version.go` line 1206 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4669#-OPBCWeW4hVR03cb9gki:-OPBXlO39W3bTESWE-Ww:b-5smf8o) ([raw file](https://github.com/cockroachdb/pebble/blob/4dfedb5c2553de13a31ec408e0564a06b2ca4ef6/internal/manifest/version.go#L1206)):*

I meant inside this function, instead of inlining `Smallest()` twice, we could do:
```
x := &m.PointKeyBounds
if .. { x = &m.RangeKeyBounds }
return x.Smallest()
```

Unless the compiler is smart enough to do something like this already, it should reduce the size of the function (and make any calling function more likely to be inlined itself)


As for the bookkeeping, I think I was wrong, it probably doesn't make a difference. My line of thinking was that if a() inlines b() which inlines c() and there's a panic inside c(), *something* somewhere has to remember that the inlined code for c() should report the `c()` stack and the `b()` stack, not just the `a()` stack. But I don't think these tables or whatever they are matter unless there really is a panic.
</blockquote></details>

I see. That does decrease the compiler's cost estimate from 78 to 51.

Done

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4669#-:-OPBYcDC6wo5WaV6ms-G:bnfp4nl)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @annrpom and @sumeerbhola)

jbowens (commented):
TFTRs!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4669#-:-OPBal3E52ES75tqOuDx:bz5mair)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @annrpom, @RaduBerinde, and @sumeerbhola)

PR #4673, opened 2025-05-01
crl-release-25.1: sstable: fix deletion properties for virtual tables

### LINKED ISSUE #4670: virtual table property scaling causes compensated size underflow
We noticed some extremely large level scores in a production environment (v24.1). We eventually reproduced and root caused to the code below:

https://github.com/cockroachdb/pebble/blob/b6630de7f9339ab36cd05c8ff51363e33dd2bbb4/sstable/reader_virtual.go#L84

We are scaling the number of deletions, but the code that uses them assumes that they are still sane with respect to each other, in particular that NumDeletions >= NumRangeDeletions + NumSizedDeletions:

https://github.com/cockroachdb/pebble/blob/47c8e5262bb640d21fcadf09e65bcaa0d88d2bc0/table_stats.go#L773

This issue tracks fixing this and also adding some safeguards to these substractions.

Jira issue: PEBBLE-432

RaduBerinde:
The independent scaling of `NumDeletions`, `NumRangeDeletions`,
`NumSizedDeletions` for virtual tables can lead to situations where
`NumRangeDeletions + NumSizedDeletions > NumDeletions`, which causes
underflows in compensated size calculations.

This change fixes the scaling code to scale the three "components" of
`NumDeletions` (range, sized, other) and then sum them.

We also add some safeguards around the subtractions - underflow now
causes panics in invariant builds and is capped to 0 in non-invariant
builds.

Fixes #4670

### DIFF
--- a/internal/invariants/off.go
+++ b/internal/invariants/off.go
@@ -49,3 +49,17 @@ func (*Value[V]) Get() V {
 
 // Set the value; no-op in non-invariant builds.
 func (*Value[V]) Set(v V) {}
+
+// SafeSub returns a - b. If a < b, it panics in invariant builds and returns 0
+// in non-invariant builds.
+func SafeSub[T Integer](a, b T) T {
+	if a < b {
+		return 0
+	}
+	return a - b
+}
+
+// Integer is a constraint that permits any integer type.
+type Integer interface {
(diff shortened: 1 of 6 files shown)


jbowens (approved):

RaduBerinde:
TFTR!

Issue #4668, opened 2025-05-01
db: allocation creating a sstable iterator with blob refs


jbowens:
```
         .          .    668:	} else {
         .      14262    669:		iter, err = reader.NewPointIter(ctx, sstable.IterOptions{
         .          .    670:			Lower:                opts.GetLowerBound(),
         .          .    671:			Upper:                opts.GetUpperBound(),
         .          .    672:			Transforms:           transforms,
         .          .    673:			FilterBlockSizeLimit: filterBlockSizeLimit,
         .          .    674:			Filterer:             filterer,
         .          .    675:			Env:                  internalOpts.readEnv,
         .          .    676:			ReaderProvider:       &v.readerProvider,
         .          .    677:			BlobContext: sstable.TableBlobContext{
         .          .    678:				ValueFetcher: internalOpts.blobValueFetcher,
60245477958 60245477958    679:				References:   file.BlobReferences,
         .          .    680:			},
         .          .    681:		})
         .          .    682:	}
```
I think this must be because `manifest.BlobReferences` implements the `sstable.BlobReferences` interface with a non-pointer receiver, forcing the slice header to be heap-allocated.

PR #4678, opened 2025-05-02
manifest: avoid BlobReferences interface allocation

jbowens:
**db: add BenchmarkPointLookupSeparatedValues**

Add a microbenchmark that exercises a point lookup wihtin a database with
separated values.

Informs #112.

**manifest: avoid BlobReferences interface allocation**

The manifest.BlobReferences type is used to satisfy the sstable.BlobReferences
interface. Previously the interface was implemented on the slice receiver type.
This caused an allocation whenever we used this type to satisfy the
sstable.BlobReferences interface. This commit moves the implementation onto the
pointer type, avoiding the need to heap-allocate a copy of the slice header.

```
goos: darwin
goarch: arm64
cpu: Apple M1 Pro
                                                    │ before.txt  │              after.txt               │
                                                    │   sec/op    │   sec/op     vs base                 │
PointLookupSeparatedValues/keys=1m,valueLen=100-10    1.636µ ± 8%   1.585µ ± 2%  -3.09% (p=0.001 n=10)
PointLookupSeparatedValues/keys=1m,valueLen=1024-10   2.452µ ± 4%   2.373µ ± 4%  -3.22% (p=0.007 n=10+6)
geomean                                               2.003µ        1.939µ       -3.15%

                                                    │ before.txt  │               after.txt               │
                                                    │    B/op     │    B/op      vs base                  │
PointLookupSeparatedValues/keys=1m,valueLen=100-10    32.000 ± 0%   8.000 ±  0%  -75.00% (p=0.000 n=10)
PointLookupSeparatedValues/keys=1m,valueLen=1024-10    265.5 ± 7%   210.0 ± 22%  -20.90% (p=0.001 n=10+6)
geomean                                                92.17        40.99        -55.53%

                                                    │ before.txt │                after.txt                │
                                                    │ allocs/op  │ allocs/op   vs base                     │
PointLookupSeparatedValues/keys=1m,valueLen=100-10    1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)
PointLookupSeparatedValues/keys=1m,valueLen=1024-10   1.000 ± 0%   0.000 ± 0%  -100.00% (n=10+6)
geomean                                               1.000                    ?                       ¹ ²
¹ summaries must be >0 to compute geomean
² ratios must be >0 to compute geomean
```

Fix https://github.com/cockroachdb/pebble/issues/4668.

### DIFF
--- a/external_test.go
+++ b/external_test.go
@@ -12,7 +12,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/cockroachdb/pebble"
+	"github.com/cockroachdb/pebble/bloom"
+	"github.com/cockroachdb/pebble/cockroachkvs"
 	"github.com/cockroachdb/pebble/metamorphic"
+	"github.com/cockroachdb/pebble/sstable"
 	"github.com/cockroachdb/pebble/vfs"
 	"github.com/cockroachdb/pebble/vfs/errorfs"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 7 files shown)


sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 1 files at r1, 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4678#-:-OPWg3RxAKMfVupWdUTx:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4678#-OPWbX0u5iqKuagtWuID:-OPWbX0u5iqKuagtWuIE:b34r479):*
nit: within
___
*[`external_test.go` line 234 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4678#-OPWcMt40aquVUjz8A_7:-OPWcMt40aquVUjz8A_8:b-k1qe1e) ([raw file](https://github.com/cockroachdb/pebble/blob/ce4059370e06bb53780257ddc17de29cf6a8d149/external_test.go#L234)):*
> ```Go
> 		require.NoError(tb, b.Commit(nil))
> 		keysToWrite = keysToWrite[n:]
> 	}
> ```

How much of this LSM is the memtable? Curious about the output of `Metrics.String`, in case the reads are running on an inverted LSM. In the prototype, `BenchmarkReadWithBlobs` waited until the scores of each level were < 2.

Should we flush after the for loop, so everything is in sstables?

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4678#-:-OPXEMzP22uOv_yS22td:b-814g8k)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`external_test.go` line 234 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4678#-OPWcMt40aquVUjz8A_7:-OPXEF5g1ZddE2ZBVIc7:bxw9vms) ([raw file](https://github.com/cockroachdb/pebble/blob/ce4059370e06bb53780257ddc17de29cf6a8d149/external_test.go#L234)):*

How much of this LSM is the memtable? Curious about the output of `Metrics.String`, in case the reads are running on an inverted LSM. In the prototype, `BenchmarkReadWithBlobs` waited until the scores of each level were < 2.

Should we flush after the for loop, so everything is in sstables?
</blockquote></details>

Good idea, I added a flush and also waited for compactions to stabilize.

PR #4681, opened 2025-05-02
internal/manifest: specialize LevelIterator seek methods

jbowens:
Previously LevelIterator's seek methods accepted an opaque
func(*TableMetadata)bool to perform comparisons. This commit specializes each
instance as separate functions, ensuring comparisons can be inlined where
possible. Special attention is given to SeekGE, which is the hot path for point
lookups.

```
goos: darwin
goarch: arm64
cpu: Apple M1 Pro
                       │   old.txt   │             newest2.txt             │
                       │   sec/op    │   sec/op     vs base                │
LevelIteratorSeekGE-10   144.3n ± 2%   110.0n ± 1%  -23.74% (p=0.000 n=10)
```

### DIFF
--- a/internal/base/internal.go
+++ b/internal/base/internal.go
@@ -247,6 +247,12 @@ func (t InternalKeyTrailer) Kind() InternalKeyKind {
 	return InternalKeyKind(t & 0xff)
 }
 
+// IsExclusiveSentinel returns true if the trailer is a sentinel for an
+// exclusive boundary.
+func (t InternalKeyTrailer) IsExclusiveSentinel() bool {
+	return t.SeqNum() == SeqNumMax
+}
+
 // InternalKey is a key used for the in-memory and on-disk partial DBs that
 // make up a pebble DB.
 //
--- a/internal/manifest/bench_test.go
+++ b/internal/manifest/bench_test.go
@@ -1,0 +1,49 @@
+// Copyright 2025 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package manifest_test
+
+import (
+	"math/rand/v2"
+	"testing"
+	"time"
+
+	"github.com/cockroachdb/pebble/cockroachkvs"
+	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/manifest"
+)
+
+func BenchmarkLevelIteratorSeekGE(b *testing.B) {
+	const countTables = 10_000
+	fileAlloc := make([]manifest.TableMetadata, countTables)
+	files := make([]*manifest.TableMetadata, countTables)
+	rng := rand.New(rand.NewPCG(0, uint64(time.Now().UnixNano())))
+	keys, _ := cockroachkvs.RandomKVs(rng, 2*countTables, cockroachkvs.KeyGenConfig{
+		PrefixAlphabetLen:  26,
+		PrefixLenShared:    2,
+		RoachKeyLen:        16,
+		AvgKeysPerPrefix:   1,
+		BaseWallTime:       uint64(time.Now().UnixNano()),
+		PercentLogical:     0,
+		PercentEmptySuffix: 0,
+		PercentLockSuffix:  0,
+	}, 0)
+	for i := 0; i < countTables; i++ {
+		fileAlloc[i] = manifest.TableMetadata{
+			FileNum: base.FileNum(i),
+		}
+		fileAlloc[i].ExtendPointKeyBounds(cockroachkvs.Compare,
+			base.MakeInternalKey(keys[i*2], base.SeqNum(i), base.InternalKeyKindSet),
+			base.MakeInternalKey(keys[i*2+1], base.SeqNum(i), base.InternalKeyKindSet))
+		fileAlloc[i].InitPhysicalBacking()
+		files[i] = &fileAlloc[i]
+	}
+
+	lm := manifest.MakeLevelMetadata(cockroachkvs.Compare, 0, files)
+	iter := lm.Iter()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_ = iter.SeekGE(cockroachkvs.Compare, keys[i%len(keys)])
+	}
+}
--- a/internal/manifest/btree.go
+++ b/internal/manifest/btree.go
@@ -1095,37 +1095,22 @@ func (i *iterator) ascend() {
 	i.pos = f.pos
 }
 
-// seek repositions the iterator over the first file for which fn returns
-// true, mirroring the semantics of the standard library's sort.Search
-// function.  Like sort.Search, seek requires the iterator's B-Tree to be
-// ordered such that fn returns false for some (possibly empty) prefix of the
-// tree's files, and then true for the (possibly empty) remainder.
-func (i *iterator) seek(fn func(*TableMetadata) bool) {
+// find seeks the iterator to the provided table metadata if it exists in the
+// tree. It returns true if the table metadata is found and false otherwise. If
+// find returns false, the position of the iterator is undefined.
+func (i *iterator) find(m *TableMetadata) bool {
 	i.reset()
 	if i.r == nil {
-		return
+		return false
 	}
-
+	i.n = i.r
 	for {
-		// Logic copied from sort.Search.
-		j, k := 0, int(i.n.count)
-		for j < k {
-			h := int(uint(j+k) >> 1) // avoid overflow when computing h
-
-			// j ≤ h < k
-			if !fn(i.n.items[h]) {
-				j = h + 1 // preserves f(j-1) == false
-			} else {
-				k = h // preserves f(k) == true
-			}
-		}
-
+		j, found := i.n.find(i.cmp, m)
 		i.pos = int16(j)
-		if i.n.leaf {
-			if i.pos == i.n.count {
-				i.next()
-			}
-			return
+		if found {
+			return true
+		} else if i.n.leaf {
+			return false
 		}
 		i.descend(i.n, i.pos)
 	}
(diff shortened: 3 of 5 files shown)


RaduBerinde (approved):
Nice! :lgtm: 

Consider adding some unit tests for the new iterator methods

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4681#-:-OPIK06lDj3KPGP_e9fc:bp5k29s)* status: 0 of 4 files reviewed, 8 unresolved discussions (waiting on @annrpom)

___
*[`internal/manifest/btree.go` line 1099 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4681#-OPIHKE32jEQPQhuW6v7:-OPIHKE32jEQPQhuW6v8:b-lkxr5b) ([raw file](https://github.com/cockroachdb/pebble/blob/bb69b0fa897a0e4022285be106110a6b251d8dc5/internal/manifest/btree.go#L1099)):*
> ```Go
> 
> // seekSeqNumL0 seeks an iterator over L0 files (ordered by sequence number) to
> // the provided table metadata if it exists.
> ```

[nit] And if it doesn't? Next file in L0 ordering?
___
*[`internal/manifest/btree.go` line 1127 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4681#-OPIILjt94UEM_0P3jHI:-OPIILjt94UEM_0P3jHJ:be4ko1a) ([raw file](https://github.com/cockroachdb/pebble/blob/bb69b0fa897a0e4022285be106110a6b251d8dc5/internal/manifest/btree.go#L1127)):*
> ```Go
> 			return
> 		}
> 		i.descend(i.n, i.pos)
> ```

`descend()` pushes onto `i.s` and we don't need that here, we could just inline the part we need
___
*[`internal/manifest/btree.go` line 1131 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4681#-OPIMcGG9FhfivJ_MDeV:-OPIMcGG9FhfivJ_MDeW:bjjkj0d) ([raw file](https://github.com/cockroachdb/pebble/blob/bb69b0fa897a0e4022285be106110a6b251d8dc5/internal/manifest/btree.go#L1131)):*
> ```Go
> }
> 
> // seekLargest repositions the iterator over the first table whose largest key
> ```

seekLargestGE
___
*[`internal/manifest/btree.go` line 1133 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4681#-OPIIpqv3R-1rmqfzpu3:-OPIIpqv3R-1rmqfzpu4:b-ceuk82) ([raw file](https://github.com/cockroachdb/pebble/blob/bb69b0fa897a0e4022285be106110a6b251d8dc5/internal/manifest/btree.go#L1133)):*
> ```Go
> // seekLargest repositions the iterator over the first table whose largest key
> // is an upper bound for the given user key. seekLargest requires the iterator's
> // B-Tree to be ordered by user keys (i.e, L1+ or a single sublevel of L0).
> ```

[nit] If there is no such table, positions the iterator before the first table.
___
*[`internal/manifest/btree.go` line 1170 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4681#-OPIMfag9azW1QWnf1jv:-OPIMfag9azW1QWnf1jw:bnzoe95) ([raw file](https://github.com/cockroachdb/pebble/blob/bb69b0fa897a0e4022285be106110a6b251d8dc5/internal/manifest/btree.go#L1170)):*
> ```Go
> }
> 
> // seekSmallest repositions the iterator over the first table whose smallest key
> ```

seekSmallestGT
___
*[`internal/manifest/btree.go` line 1171 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4681#-OPIM1pu2m5-wvRtfB3S:-OPIM1pu2m5-wvRtfB3T:b-ekv3yw) ([raw file](https://github.com/cockroachdb/pebble/blob/bb69b0fa897a0e4022285be106110a6b251d8dc5/internal/manifest/btree.go#L1171)):*
> ```Go
> 
> // seekSmallest repositions the iterator over the first table whose smallest key
> // is a lower bound for the given user key. seekSmallest requires the iterator's
> ```

is NOT a lower bound! (better yet just say it is > userKey)

You can also consider making this return the last table with smallest key < userKey, and don't do `Prev()` in the calling code. We'd have to modify the code below like this:
```
			h := int(uint(j+k+1) >> 1) // avoid overflow when computing h
			// j < h ≤ k
			if cmp(i.n.items[h].Smallest().UserKey, userKey) < 0 {
				j = h  // preserves INVARIANT A
			} else {
				k = h-1 // preserves INVARIANT B
			}
```
___
*[`internal/manifest/btree.go` line 1173 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4681#-OPIM0z75jGEHsU5S5MQ:-OPIM0z75jGEHsU5S5MR:b-fl2bsa) ([raw file](https://github.com/cockroachdb/pebble/blob/bb69b0fa897a0e4022285be106110a6b251d8dc5/internal/manifest/btree.go#L1173)):*
> ```Go
> // is a lower bound for the given user key. seekSmallest requires the iterator's
> // B-Tree to be ordered by user keys (i.e, L1+ or a single sublevel of L0).
> func (i *iterator) seekSmallest(cmp base.Compare, userKey []byte) {
> ```

If there is no such table, positions the iterator after the last table.
___
*[`internal/manifest/btree.go` line 1188 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4681#-OPIKEPhDSaCh00-8s_b:-OPIKEPhDSaCh00-8s_c:b-trxvoo) ([raw file](https://github.com/cockroachdb/pebble/blob/bb69b0fa897a0e4022285be106110a6b251d8dc5/internal/manifest/btree.go#L1188)):*
> ```Go
> 			// j ≤ h < k
> 			if cmp(i.n.items[h].Smallest().UserKey, userKey) < 0 {
> 				j = h + 1 // preserves INVARIANT A
> ```

Are you sure about this? This table could still be the one we're looking for if the next table starts after `userKey`? Feels like this should be:

```
			h := int(uint(j+k+1) >> 1) // avoid overflow when computing h
			// j ≤ h < k
			if cmp(i.n.items[h].Smallest().UserKey, userKey) <= 0 {
				j = h  // preserves INVARIANT A
			} else {
				k = h-1 // preserves INVARIANT B
			}
```

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4681)* status: 0 of 5 files reviewed, 7 unresolved discussions (waiting on @annrpom and @RaduBerinde)

___
*[`internal/manifest/btree.go` line 1099 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4681#-OPIHKE32jEQPQhuW6v7:-OPWBo5MCmKvEYSiYJ5J:baaezim) ([raw file](https://github.com/cockroachdb/pebble/blob/bb69b0fa897a0e4022285be106110a6b251d8dc5/internal/manifest/btree.go#L1099)):*

[nit] And if it doesn't? Next file in L0 ordering?
</blockquote></details>

Updated this to use the btree node `find` func, so that all calls to `Find` can make use.
___
*[`internal/manifest/btree.go` line 1127 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4681#-OPIILjt94UEM_0P3jHI:-OPWBz4D5TTB6yvz_Ao9:b-jt1bkv) ([raw file](https://github.com/cockroachdb/pebble/blob/bb69b0fa897a0e4022285be106110a6b251d8dc5/internal/manifest/btree.go#L1127)):*

`descend()` pushes onto `i.s` and we don't need that here, we could just inline the part we need
</blockquote></details>

We do need the stack bit because the expectation is that the iterator is now positioned over the provided file.
___
*[`internal/manifest/btree.go` line 1131 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4681#-OPIMcGG9FhfivJ_MDeV:-OPWC3ju8h8pcCqVOINN:b3uo2x7) ([raw file](https://github.com/cockroachdb/pebble/blob/bb69b0fa897a0e4022285be106110a6b251d8dc5/internal/manifest/btree.go#L1131)):*

seekLargestGE
</blockquote></details>

I ended up inlining this func because it's only called once, and the semantics of the seekSmallestLT func are closely intertwined with the semantics of `LevelIterator.SeekLT`
___
*[`internal/manifest/btree.go` line 1170 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4681#-OPIMfag9azW1QWnf1jv:-OPWCGHOAGoB5c_9e8ym:be0invf) ([raw file](https://github.com/cockroachdb/pebble/blob/bb69b0fa897a0e4022285be106110a6b251d8dc5/internal/manifest/btree.go#L1170)):*

seekSmallestGT
</blockquote></details>

I ended up inlining this func because it's only called once, and the semantics of this func are closely intertwined with the semantics of `LevelIterator.SeekLT`
___
*[`internal/manifest/btree.go` line 1171 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4681#-OPIM1pu2m5-wvRtfB3S:-OPWCQAzDn9pyHjm_bXf:bdspzja) ([raw file](https://github.com/cockroachdb/pebble/blob/bb69b0fa897a0e4022285be106110a6b251d8dc5/internal/manifest/btree.go#L1171)):*

is NOT a lower bound! (better yet just say it is > userKey)

You can also consider making this return the last table with smallest key < userKey, and don't do `Prev()` in the calling code. We'd have to modify the code below like this:
```
			h := int(uint(j+k+1) >> 1) // avoid overflow when computing h
			// j < h ≤ k
			if cmp(i.n.items[h].Smallest().UserKey, userKey) < 0 {
				j = h  // preserves INVARIANT A
			} else {
				k = h-1 // preserves INVARIANT B
			}
```
</blockquote></details>

If our seek key is less than all tables smallest keys, we want to ultimately end up before the first table. This is also complicated by the iterator start and end bounds which are both inclusive

jbowens (commented):
@RaduBerinde i made some material changes if u want to take another look

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4681#-:-OPfSztI09XtikHHdNN6:bo58ci2)* status: 0 of 5 files reviewed, 7 unresolved discussions (waiting on @annrpom and @RaduBerinde)

jbowens:
TFTR!

PR #4685, opened 2025-05-05
cmd/pebble: enable value separation in benchmarks

jbowens:
Enable value separation in the cmd/pebble ycsb nightly benchmarks. Configure a minimum size that effectively means the value=1024 variants of the benchmarks will separate values, and the value=64 variants will not.

Informs #112.

### DIFF
--- a/cmd/pebble/db.go
+++ b/cmd/pebble/db.go
@@ -59,7 +59,7 @@ func newPebbleDB(dir string) DB {
 		CacheSize:                   cacheSize,
 		Comparer:                    &cockroachkvs.Comparer,
 		DisableWAL:                  disableWAL,
-		FormatMajorVersion:          pebble.FormatNewest,
+		FormatMajorVersion:          pebble.FormatExperimentalValueSeparation,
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

I wonder if it's useful to have two flavors of each variant (with and without value separation). It might make it more clear what the tradeoffs are.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4685#-:-OPXPiV63jFWWh3Avup_:bwfhyfr)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @xxmplus)

jbowens (commented):
Yeah, it would be useful. These benchmarks currently take 9h tho (6 variants; 1.5h each; each run 1x on 5 different machines). We could consider running a variety of shorter duration benchmarks too

TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4685#-:-OPjmuQY4_nT5AMbwfoY:b-y4fes8)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @xxmplus)

PR #4686, opened 2025-05-05
db: track aggregate size, inclusive of separated values

jbowens:
Add estimation of the physical size of a table's separated values and use this to compute an 'aggregate size' of a level. Use this size when determining level scores.

Informs #112.


RaduBerinde (approved):

jbowens:
TFTR!

PR #4688, opened 2025-05-07
compression: move low level compression code to separate package

RaduBerinde:
Move the compression code to a separate package. This package will
expose compression levels (useful for running experiments), while the
`sstable/block` options will remain higher level.

PR #4689, opened 2025-05-07
block: push down MinLZ->Snappy fallback

RaduBerinde:
#### block: push down MinLZ->Snappy fallback

MinLZ has an 8MB block limit. We fall back to Snappy for the unlikely
case of larger blocks. The MinLZ compressor `Compress()` call can
return `SnappyCompressionIndicator`, which is unintuitive.

This commit hides this unfortunate detail inside the MinLZ
implementation. The decompression relies on the fact that the MinLZ
library can detect when it is used to decode a Snappy block. The
MinLZ compressor now always returns `MinLZCompressionIndicator`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/guptarohit/asciigraph v0.5.5
 	github.com/klauspost/compress v1.17.11
 	github.com/kr/pretty v0.3.1
-	github.com/minio/minlz v1.0.0
+	github.com/minio/minlz v1.0.1-0.20250507153514-87eb42fe8882
 	github.com/olekukonko/tablewriter v0.0.5
 	github.com/pkg/errors v0.9.1
 	github.com/pmezard/go-difflib v1.0.0
--- a/go.sum
+++ b/go.sum
@@ -149,8 +149,8 @@ github.com/mattn/go-sqlite3 v1.14.5/go.mod h1:WVKg1VTActs4Qso6iwGbiFih2UIHo0ENGw
 github.com/matttproud/golang_protobuf_extensions v1.0.1/go.mod h1:D8He9yQNgCq6Z5Ld7szi9bcBfOoFv/3dc6xSMkL2PC0=
 github.com/matttproud/golang_protobuf_extensions v1.0.4 h1:mmDVorXM7PCGKw94cs5zkfA9PSy5pEvNWRP0ET0TIVo=
 github.com/matttproud/golang_protobuf_extensions v1.0.4/go.mod h1:BSXmuO+STAnVfrANrmjBb36TMTDstsz7MSK+HVaYKv4=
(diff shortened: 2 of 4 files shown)


jbowens (approved):
I don't feel strongly, but imo it's more less intuitive for the compression indicator byte to not actually indicate the compression algorithm. That byte's intention is to avoid the need for a check like `minlz.IsMinLZ(compressed)`. I don't think we should strive to maintain a 1:1 mapping between the `Compression` enum and the compression indicators.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4689#-:-OPfMIeq1sCnOQnEvB6C:b-chgd9p)* status: 0 of 2 files reviewed, all discussions resolved

RaduBerinde (commented):
The thing is that minlz can normally decompress Snappy/S2 data. We have the `IsMinLZ` check because of https://github.com/minio/minlz/issues/14 which just got fixed. If I update the library, we won't even need that. Would that make this situation better or do you still feel we should have it shown as `Snappy`?

The main reason is that I'm working to separate the lower-level compression code (see #4688) and it's non-intuitive to have APIs for compression where you specify the algorithm (and soon, level) but that can internally use some other algorithm..

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4689#-:-OPfQ8V3E0_j1MXskyCX:b2jmpi7)* status: 0 of 2 files reviewed, all discussions resolved

jbowens (commented):
With the reorganized APIs how would adaptive compression fit in? would the adaptive compressor encode a different compression indicator depending on what ends up being used?

> The thing is that minlz can normally decompress Snappy/S2 data

Yeah, that does make it cleaner

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4689#-:-OPfh6WM7jo907t5Yxcr:b-q5jqlw)* status: 0 of 2 files reviewed, all discussions resolved

RaduBerinde (commented):
> With the reorganized APIs how would adaptive compression fit in?

The adaptive compression would be "above" the basic compression library which just provides the functionality to compress or decompress with a specific algorithm. Anything more fancy would be implemented in `block`.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4689#-:-OPg-qdZ-eOKUcuV89Aa:b9196gh)* status: 0 of 2 files reviewed, all discussions resolved

RaduBerinde (commented):
> Yeah, that does make it cleaner

Done.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4689#-:-OPhBhGbC3pMrXDBZv28:bixpqyd)* status: 0 of 4 files reviewed, all discussions resolved

RaduBerinde:
TFTR!

PR #4690, opened 2025-05-07
sstable: fix invariants panic due to zero virtual table size

RaduBerinde:
Informs https://github.com/cockroachdb/cockroach/issues/146240

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -344,7 +344,7 @@ func ingestLoad1(
 
 	meta = &tableMetadata{}
 	meta.FileNum = fileNum
-	meta.Size = uint64(readable.Size())
+	meta.Size = max(uint64(readable.Size()), 1)
 	meta.CreationTime = time.Now().Unix()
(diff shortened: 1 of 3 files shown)


RaduBerinde:
Not exactly sure through which path we got a vtable with zero size in the issue..

RaduBerinde:
I will see if I can repro with more assertions to find out.

RaduBerinde:
So far no luck with `dev test pkg/backup -f TestOnlineRestoreBasic --count 100`

RaduBerinde:
Oh.. the problem is actually that size > backingSize: `panic: bad scale(): p.Size=24270  p.BackingSize=24262` so we turn 0 into 1.

RaduBerinde:
I filed https://github.com/cockroachdb/pebble/issues/4696 to track fixing the code that results in `Size > Backing.Size`. I will merge and backport this to prevent the assertion / invalid scaling results.

Issue #4566, opened 2025-04-17
metamorphic: enable value separation


jbowens:
Enable value separation within the metamorphic test and randomize its configuration within the random option runs.

Jira issue: PEBBLE-405

Epic CRDB-20379

jbowens:
I forget how the metamorphic test uses checkpoints but it probably requires #4555 first.

PR #4691, opened 2025-05-07
metamophic: integrate value separation

annrpom:
### db: update compaction's bytesWritten atomically
Due to value separation, a compaction can be writing to
blob files and sstables in parallel. This means that a race
condition can occur in which updates to our per-compaction
`written` accumulator might be lost. This patch fixes this
behavior by using the `atomic.Int64` type to synchronize our
`written` accumulator.

---

### db: prohibit blob references in shared sstables
This patch prohibits the creation of shared sstables with blob references and
prevents sstables with blob references from going through a
copy compaction.
If a compaction is configured to output to shared storage, it will run
with the `neverSeparateValues` `compact.ValueSeparation` implementation.

Fixes: https://github.com/cockroachdb/pebble/issues/4693

---

### sstable/blob: ensure no mutation after returning writer to pool

Previously, it was possible for a race condition to occur between
blob file writers because the logic to close a file writer made it
possible (in the defer) to mutate a writer after putting it back in
the writer pool. This patch ensures that this is no longer possible
by removing said defer logic.

---

### metamophic: integrate value separation
This patch enables value separation within the metamorphic
test.

Fixes: https://github.com/cockroachdb/pebble/issues/4566

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -109,12 +109,12 @@ func (cl compactionLevel) String() string {
 
 // compactionWritable is a objstorage.Writable wrapper that, on every write,
 // updates a metric in `versions` on bytes written by in-progress compactions so
-// far. It also increments a per-compaction `written` int.
+// far. It also increments a per-compaction `written` atomic int.
 type compactionWritable struct {
 	objstorage.Writable
 
 	versions *versionSet
-	written  *int64
+	written  *atomic.Int64
 }
 
 // Write is part of the objstorage.Writable interface.
@@ -123,7 +123,7 @@ func (c *compactionWritable) Write(p []byte) error {
 		return err
 	}
 
-	*c.written += int64(len(p))
+	c.written.Add(int64(len(p)))
 	c.versions.incrementCompactionBytes(int64(len(p)))
 	return nil
 }
@@ -259,7 +259,7 @@ type compaction struct {
 	// flushing contains the flushables (aka memtables) that are being flushed.
 	flushing flushableList
 	// bytesWritten contains the number of bytes that have been written to outputs.
-	bytesWritten int64
+	bytesWritten atomic.Int64
 
 	// The boundaries of the input data.
 	smallest InternalKey
@@ -1598,7 +1598,7 @@ func (d *DB) flush1() (bytesFlushed uint64, err error) {
 
 	d.clearCompactingState(c, err != nil)
 	delete(d.mu.compact.inProgress, c)
-	d.mu.versions.incrementCompactions(c.kind, c.extraLevels, c.pickerMetrics, c.bytesWritten, err)
+	d.mu.versions.incrementCompactions(c.kind, c.extraLevels, c.pickerMetrics, c.bytesWritten.Load(), err)
 
 	var flushed flushableList
 	if err == nil {
@@ -2536,8 +2536,8 @@ func (d *DB) compact1(c *compaction, errChannel chan error) (err error) {
 	// NB: clearing compacting state must occur before updating the read state;
 	// L0Sublevels initialization depends on it.
 	d.clearCompactingState(c, err != nil)
-	d.mu.versions.incrementCompactions(c.kind, c.extraLevels, c.pickerMetrics, c.bytesWritten, err)
-	d.mu.versions.incrementCompactionBytes(-c.bytesWritten)
+	d.mu.versions.incrementCompactions(c.kind, c.extraLevels, c.pickerMetrics, c.bytesWritten.Load(), err)
+	d.mu.versions.incrementCompactionBytes(-c.bytesWritten.Load())
 
 	info.TotalDuration = d.timeNow().Sub(c.beganAt)
 	d.opts.EventListener.CompactionEnd(info)
--- a/compaction.go
+++ b/compaction.go
@@ -302,6 +302,8 @@ type compaction struct {
 	pickerMetrics pickedCompactionMetrics
 
 	grantHandle CompactionGrantHandle
+
+	opts objstorage.CreateOptions
 }
 
 // inputLargestSeqNumAbsolute returns the maximum LargestSeqNumAbsolute of any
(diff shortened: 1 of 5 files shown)


annrpom:
we're still trying to track down why/how we can get a non-nil `BlobReferences` with a len of 0, but putting this up here while we do

lassenordahl:
@annrpom I hope you track it down!

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4691)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @jbowens)

___
*[`compaction.go` line 3390 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4691#-OQO7GNX1X6A4G5s77kU:-OQOYiEM00uC-EIDWsC2:b-kbrmae) ([raw file](https://github.com/cockroachdb/pebble/blob/2ec83bc4b5fb7ea600d080239e446d1ba1dc3431/compaction.go#L3390)):*

I think this initialization of the `CreateOptions` should be lifted up into `(*DB).compactAndWrite` so we do it once per-compaction. Then if `remote.ShouldCreateShared(d.opts.Experimental.CreateOnShared, c.outputLevel.level)` is true, we should set the `valueSeparation` to `neverSeparateValues`, so that the compaction won't separate values.
</blockquote></details>

what about if we initialize `CreateOptions` in `newCompaction/newFlush` and override the `valueSeparation` to `NeverSeparateValues` (given shared storage is preferred) there? we could also do it in `runCompaction` if we wanted to elimate some small code dupe, but this current way makes a bit more sense organizationally
___
*[`compaction.go` line 2584 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4691#-OQO99TnA0l-wCw-7nlu:-OQdId3F71k_PC0Cj5XN:b-awxhoz) ([raw file](https://github.com/cockroachdb/pebble/blob/62f816acc4f14c3baf881462ed96e80a93d13c13/compaction.go#L2584)):*
> In `invariants` builds, `errors.AssertionFailedf` will panic when it's called. In non-invariants builds, it'll return the error.

i see -- thanks for explaining; done 
___
*[`metamorphic/options.go` line 354 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4691#-OQOAGh75z83DTX8m-Oj:-OQdI22Z9wC7UA2t5n5s:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/62f816acc4f14c3baf881462ed96e80a93d13c13/metamorphic/options.go#L354)):*

The metamorphic test won't actually generate values larger than 20 bytes today, so I think we should bring this down to something like 5
</blockquote></details>

done
___
*[`metamorphic/options.go` line 917 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4691#-OQOAQgwDLFv8Cso-wfL:-OQdIUxv3ILV734MuNS5:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/62f816acc4f14c3baf881462ed96e80a93d13c13/metamorphic/options.go#L917)):*

and here, I think we should make this `minSize := 1+rng.Intn(maxValueSize)` so that we constrain it to the value sizes we use
</blockquote></details>

done
___
*[`sstable/blob/blob.go` line 241 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4691#-OQO8PWk0SF7Uofj72xX:-OQdF7f55TylBE-_ni7f:b-kkoaj2) ([raw file](https://github.com/cockroachdb/pebble/blob/636ed6f12d59b7b2f987b34d296d613c3c2736fa/sstable/blob/blob.go#L241)):*

I think at one point it might've been to allow `Close` to be called twice (eg, once in the normal success condition and once in a defer), but I don't think we actually need that and it's fine to remove.
</blockquote></details>

done; ty for insight

jbowens (approved):
:lgtm_strong: 

> Reviewed 1 of 1 files at r8, 4 of 4 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4691#-:-OQdkZxG6ef07SOMpexn:b-uwkowd)* status: all files reviewed, 3 unresolved discussions (waiting on @annrpom)

___
*[`compaction.go` line 3390 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4691#-OQO7GNX1X6A4G5s77kU:-OQdkKrf6fGAE_kPqJBK:bq0ncok) ([raw file](https://github.com/cockroachdb/pebble/blob/2ec83bc4b5fb7ea600d080239e446d1ba1dc3431/compaction.go#L3390)):*

what about if we initialize `CreateOptions` in `newCompaction/newFlush` and override the `valueSeparation` to `NeverSeparateValues` (given shared storage is preferred) there? we could also do it in `runCompaction` if we wanted to elimate some small code dupe, but this current way makes a bit more sense organizationally
</blockquote></details>

sounds good
___
*[`compaction.go` line 465 at r9](https://reviewable.io/reviews/cockroachdb/pebble/4691#-OQdk-6YDX85EUQ_NZ1H:-OQdk-6YDX85EUQ_NZ1I:bnm1bbh) ([raw file](https://github.com/cockroachdb/pebble/blob/5ef3e394e4c9aca02ff9aa467a58a9cee133937f/compaction.go#L465)):*
> ```Go
> 	}
> 	if c.opts.PreferSharedStorage {
> 		c.getValueSeparation = func(_ JobID, _ *compaction, _ sstable.TableFormat) compact.ValueSeparation {
> ```

nit: this can use the predefined func `neverSeparateValues`

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4691)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`compaction.go` line 465 at r9](https://reviewable.io/reviews/cockroachdb/pebble/4691#-OQdk-6YDX85EUQ_NZ1H:-OQdmDEQ8FmB-Uesi15H:bqmft0f) ([raw file](https://github.com/cockroachdb/pebble/blob/5ef3e394e4c9aca02ff9aa467a58a9cee133937f/compaction.go#L465)):*

nit: this can use the predefined func `neverSeparateValues`
</blockquote></details>

soooooo true (done)

annrpom:
TFTR! ('-')7

PR #4692, opened 2025-05-07
db: ensure deletesized keys don't get separated, fix zero size blob ref vals

annrpom:
### db: ensure deletesized do not get separated
This patch ensures that keys of the `DeleteSized` kind
do not get value-separated as such values are not worth
externalizing and sometimes used by compaction.

---

### db: ensure blob references have proper value size
When scaling blob references' `ValueSize` during excise, we need to ensure 
that we round up to the nearest integer. Otherwise, a low `ValueSize` we want to
scale could easily be rounded to `0`.

### DIFF
--- a/excise.go
+++ b/excise.go
@@ -445,9 +445,12 @@ func determineExcisedTableSize(
 func determineExcisedTableBlobReferences(
 	originalBlobReferences manifest.BlobReferences, originalSize uint64, excisedTable *tableMetadata,
 ) {
+	if len(originalBlobReferences) == 0 {
+		return
+	}
 	newBlobReferences := make(manifest.BlobReferences, len(originalBlobReferences))
 	for i, bf := range originalBlobReferences {
-		bf.ValueSize = bf.ValueSize * excisedTable.Size / originalSize
+		bf.ValueSize = max(bf.ValueSize*excisedTable.Size/originalSize, 1)
 		newBlobReferences[i] = bf
 	}
 	excisedTable.BlobReferences = newBlobReferences
--- a/testdata/value_separation_policy
+++ b/testdata/value_separation_policy
@@ -26,9 +26,20 @@ foo#209,SET:poiandyaya
 RawWriter.AddWithBlobHandle("baz#202,SET", "(f0,blk0[0:5])", 0, false)
 RawWriter.AddWithBlobHandle("foo#209,SET", "(f0,blk0[5:15])", 0, false)
 
+# Ensure that merge and deletesized keys do not get separated if they meet the
+# minimum size.
+add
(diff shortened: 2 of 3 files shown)


jbowens (approved):
:lgtm:

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4692#-:-OPgTuiZAq1CTN8wtLan:bnfp4nl)* status: 2 of 4 files reviewed, 2 unresolved discussions

___
*[`-- commits` line 13 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4692#-OPgTDhn2AvmPWdvg5FN:-OPgTDhn2AvmPWdvg5FO:b-zbiuv2):*
curious what did the 0 value size trip? an assertion somewhere?
___
*[`excise.go` line 457 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4692#-OPgSxPv6e1KPqaQRSR4:-OPgSxPv6e1KPqaQRSR5:b-96g3yw) ([raw file](https://github.com/cockroachdb/pebble/blob/536d2987eb4ab5822fded63d877c0b87e5c2f506/excise.go#L457)):*
> ```Go
> 				excisedTable.Size/originalSize))
> 		}
> 		bf.ValueSize = uint64(adjustedSize)
> ```

how about keeping the old integer math but slapping a `max(bf.ValueSize, 1)` on it?

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4692)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`-- commits` line 13 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4692#-OPgTDhn2AvmPWdvg5FN:-OPgW7yn8wNCkvm8l__9:bhwvw9p):*

curious what did the 0 value size trip? an assertion somewhere?
</blockquote></details>

yes; https://github.com/cockroachdb/pebble/blob/c6f4d697e3dc8b2f13bc54ce748761d7928f21e0/internal/manifest/blob_metadata.go#L53-L55
___
*[`excise.go` line 457 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4692#-OPgSxPv6e1KPqaQRSR4:-OPgWBiLComwyvNoK_2h:b-535t2j) ([raw file](https://github.com/cockroachdb/pebble/blob/536d2987eb4ab5822fded63d877c0b87e5c2f506/excise.go#L457)):*

how about keeping the old integer math but slapping a `max(bf.ValueSize, 1)` on it?
</blockquote></details>

done ok

annrpom:
TFTR! ('-')7

PR #4701, opened 2025-05-08
db: fix size of deduplicated external backings

RaduBerinde:
#### db: fix size of deduplicated external backings

We allow external ingestions that have the same underlying remote
object, and we deduplicate the backings in this case. However, the
logic to set the backing size did not take deduplication into account.

We fix this by summing the `Size` of external files sharing a backing
within the same ingestion operation. When a backing is reused across
separate operations, we cap the virtual table size to the backing
size.


jbowens (approved):
> Reviewed 5 of 5 files at r1, 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4701)* status: 5 of 6 files reviewed, all discussions resolved (waiting on @sumeerbhola)

RaduBerinde:
TFTR!

PR #4702, opened 2025-05-08
base: add callback to ArchiveCleaner

tbg:
For issues like https://github.com/cockroachdb/cockroach/issues/145953, I'm
anticipating that we'll want to use `ArchiveCleaner` by default in some tests.
To avoid data pile-up, I'd like to be able to a) remove data after a while and
b) filter on the file type (keep WALs more aggressively for example). This
callback will allow both.

### DIFF
--- a/internal/base/cleaner.go
+++ b/internal/base/cleaner.go
@@ -30,7 +30,12 @@ func (DeleteCleaner) String() string {
 }
 
 // ArchiveCleaner archives file instead delete.
-type ArchiveCleaner struct{}
+type ArchiveCleaner struct {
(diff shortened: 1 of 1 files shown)


RaduBerinde (commented):
It's a very simple implementation, why not implement a different `Cleaner` instead of adding callbacks to this one?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4702#-:-OPk_7Zt5wn9y_TrgEI7:b4cpkdj)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @annrpom)

Issue #4621, opened 2025-04-24
db: incorporate blob files into EstimateDiskUsage


jbowens:
When estimating the disk usage between two bounds, we should incorporate contribution from blob files referenced from overlapping sstables. When there's partial overlap with a sstable, we don't have a means of accurately determining how much of a blob reference comes from the span passed into EstimateDiskUsage specifically. I suppose we could crudely estimate it using linear interpolation.

We use `(*sstable.Reader).EstimateDiskUsage(start, end)` to get an accurate estimate of the disk usage within the sstable itself. We divide that by the `TableMetadata.Size` to get an estimate of the fraction of the table's blob references that fall within the range. Then for each blob reference, we scale the uncompressd `ValueLen` by that fraction to get estimates of the uncompressed value quantities. Then we scale each of those by the compression ratio of the blob file and sum them all together. 

Jira issue: PEBBLE-424

Epic CRDB-20379

PR #4703, opened 2025-05-08
db: incorporate blob files into EstimateDiskUsage

annrpom:
This patch ensures that blob files are also included in disk usage estimation.

Fixes: #4621

### DIFF
--- a/db.go
+++ b/db.go
@@ -501,9 +501,12 @@ type DB struct {
 		// annotators contains various instances of manifest.Annotator which
 		// should be protected from concurrent access.
 		annotators struct {
-			totalSize    *manifest.Annotator[uint64]
-			remoteSize   *manifest.Annotator[uint64]
-			externalSize *manifest.Annotator[uint64]
+			// totalFileSize is the sum of the size of all files in the
+			// database. This includes local, remote, and external sstables --
+			// along with blob files.
+			totalFileSize *manifest.Annotator[uint64]
+			remoteSize    *manifest.Annotator[uint64]
+			externalSize  *manifest.Annotator[uint64]
 		}
 	}
 
@@ -2375,24 +2378,30 @@ func (d *DB) SSTables(opts ...SSTablesOption) ([][]SSTableInfo, error) {
 	return destLevels, nil
 }
 
-// makeFileSizeAnnotator returns an annotator that computes the total size of
-// files that meet some criteria defined by filter.
+// makeFileSizeAnnotator returns an annotator that computes the total
+// storage size of files that meet some criteria defined by filter. When
+// applicable, this includes both the sstable size and the size of any
+// referenced blob files.
 func (d *DB) makeFileSizeAnnotator(filter func(f *tableMetadata) bool) *manifest.Annotator[uint64] {
 	return &manifest.Annotator[uint64]{
 		Aggregator: manifest.SumAggregator{
 			AccumulateFunc: func(f *tableMetadata) (uint64, bool) {
 				if filter(f) {
-					return f.Size, true
+					return f.Size + f.EstimatedReferenceSize(), true
 				}
 				return 0, true
 			},
 			AccumulatePartialOverlapFunc: func(f *tableMetadata, bounds base.UserKeyBounds) uint64 {
 				if filter(f) {
-					size, err := d.fileCache.estimateSize(f, bounds.Start, bounds.End.Key)
+					overlappingFileSize, err := d.fileCache.estimateSize(f, bounds.Start, bounds.End.Key)
 					if err != nil {
 						return 0
 					}
-					return size
+					overlapFraction := float64(overlappingFileSize) / float64(f.Size)
+					// Scale the blob reference size proportionally to the file
+					// overlap from the bounds to approximate only the blob
+					// references that overlap with the requested bounds.
+					return overlappingFileSize + uint64(float64(f.EstimatedReferenceSize())*overlapFraction)
 				}
 				return 0
 			},
@@ -2438,9 +2447,10 @@ func (d *DB) EstimateDiskUsageByBackingType(
 	readState := d.loadReadState()
 	defer readState.unref()
 
-	totalSize = *d.mu.annotators.totalSize.VersionRangeAnnotation(readState.current, bounds)
+	totalSize = *d.mu.annotators.totalFileSize.VersionRangeAnnotation(readState.current, bounds)
 	remoteSize = *d.mu.annotators.remoteSize.VersionRangeAnnotation(readState.current, bounds)
 	externalSize = *d.mu.annotators.externalSize.VersionRangeAnnotation(readState.current, bounds)
+
 	return
 }
 
--- a/open.go
+++ b/open.go
@@ -417,7 +417,7 @@ func Open(dirname string, opts *Options) (db *DB, err error) {
 	d.newIters = d.fileCache.newIters
 	d.tableNewRangeKeyIter = tableNewRangeKeyIter(d.newIters)
 
-	d.mu.annotators.totalSize = d.makeFileSizeAnnotator(func(f *manifest.TableMetadata) bool {
+	d.mu.annotators.totalFileSize = d.makeFileSizeAnnotator(func(f *manifest.TableMetadata) bool {
 		return true
 	})
 	d.mu.annotators.remoteSize = d.makeFileSizeAnnotator(func(f *manifest.TableMetadata) bool {
--- a/tool/testdata/db_space
+++ b/tool/testdata/db_space
@@ -2,37 +2,49 @@ db space
 ----
 accepts 1 arg(s), received 0
 
-# covers the whole 4.sst
-
+# Covers the whole 4.sst.
 db space --start=a --end=z
 ../testdata/db-stage-4
 ----
 709
 
-# covers from left of 4.sst to its only data block
-
+# Covers from left of 4.sst to its only data block.
 db space --start=a --end=bar
 ../testdata/db-stage-4
 ----
 62
 
-# covers from 4.sst's only data block to its right
-
+# Covers from 4.sst's only data block to its right.
 db space --start=foo --end=z
 ../testdata/db-stage-4
 ----
 62
 
-# covers non-overlapping range to left of 4.sst
-
+# Covers non-overlapping range to left of 4.sst.
 db space --start=a --end=a
 ../testdata/db-stage-4
 ----
 0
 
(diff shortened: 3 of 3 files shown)


sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4703)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`db.go` line 2457 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4703#-OPpOJaT1uyYGJ5La29X:-OPpOJaT1uyYGJ5La29Y:btpbtg7) ([raw file](https://github.com/cockroachdb/pebble/blob/72637766557038b8834991628890dde928e95448/db.go#L2457)):*
> ```Go
> 			blobSize += uint64(float64(f.EstimatedReferenceSize()) * overlapFraction)
> 		}
> 	}
> ```

This looks computationally correct to me, but the approach here is different from what was suggested in https://github.com/cockroachdb/pebble/issues/4621:
We use the `Annotator`s since they remember the state of the computation (when stable) in the btree nodes, allowing for efficient incremental computation. In this particular case, the `totalSize` `SumAggregator` has a `AccumulateFunc` and `AccumulatePartialOverlapFunc`, and when the bounds passed in `*d.mu.annotators.totalSize.VersionRangeAnnotation(readState.current, bounds)` completely contain a btree node, the remembered value will be used, instead of iterating over the files in that btree node (see the code in annotator.go `accumulateRangeAnnotation`, on how the calling works).

The approach suggested in that issue was to use the annotator and change the behavior of https://github.com/cockroachdb/pebble/blob/1b97e1b6dbabd0cd02dd9a3ffb88a3f526f60021/db.go#L2391, which eventually ends up in `Reader.EstimateDiskUsage`, so that is what needs to be changed. I think that suggestion overlooked that we will also need to change the return value from `AccumulateFunc` https://github.com/cockroachdb/pebble/blob/1b97e1b6dbabd0cd02dd9a3ffb88a3f526f60021/db.go#L2385 since the sstable size is not what we want.

While we are here, we should change the commentary on `makeFileSizeAnnotator` and possibly the name too (including `annotators.totalSize`). It is easy to imagine this is the size of just the sstables and later misuse it in some other context.

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4703)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`db.go` line 2457 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4703#-OPpOJaT1uyYGJ5La29X:-OPpRar25RNllozxlEfW:b-i7n90y) ([raw file](https://github.com/cockroachdb/pebble/blob/72637766557038b8834991628890dde928e95448/db.go#L2457)):*

This looks computationally correct to me, but the approach here is different from what was suggested in https://github.com/cockroachdb/pebble/issues/4621:
We use the `Annotator`s since they remember the state of the computation (when stable) in the btree nodes, allowing for efficient incremental computation. In this particular case, the `totalSize` `SumAggregator` has a `AccumulateFunc` and `AccumulatePartialOverlapFunc`, and when the bounds passed in `*d.mu.annotators.totalSize.VersionRangeAnnotation(readState.current, bounds)` completely contain a btree node, the remembered value will be used, instead of iterating over the files in that btree node (see the code in annotator.go `accumulateRangeAnnotation`, on how the calling works).

The approach suggested in that issue was to use the annotator and change the behavior of https://github.com/cockroachdb/pebble/blob/1b97e1b6dbabd0cd02dd9a3ffb88a3f526f60021/db.go#L2391, which eventually ends up in `Reader.EstimateDiskUsage`, so that is what needs to be changed. I think that suggestion overlooked that we will also need to change the return value from `AccumulateFunc` https://github.com/cockroachdb/pebble/blob/1b97e1b6dbabd0cd02dd9a3ffb88a3f526f60021/db.go#L2385 since the sstable size is not what we want.

While we are here, we should change the commentary on `makeFileSizeAnnotator` and possibly the name too (including `annotators.totalSize`). It is easy to imagine this is the size of just the sstables and later misuse it in some other context.
</blockquote></details>

ah ok this makes sense; done & ty for the annotator pointers

what do you think about renaming `makeFileSizeAnnotator` to something like `makeStorageSizeAnnotator`? i also renamed `annotators.totalSize` -> `annotators.totalDiskUsageSize`

sumeerbhola (changes requested):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4703)* status: all files reviewed, 4 unresolved discussions (waiting on @annrpom)

___
*[`db.go` line 2457 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4703#-OPpOJaT1uyYGJ5La29X:-OQ57SNM1rhwH6nlAen5:b-7jpzg5) ([raw file](https://github.com/cockroachdb/pebble/blob/72637766557038b8834991628890dde928e95448/db.go#L2457)):*
> what do you think about renaming `makeFileSizeAnnotator` to something like `makeStorageSizeAnnotator`?

These are files and it is measuring file size, so it isn't necessarily a bad name. So I don't mind keeping the name unchanged.
___
*[`tool/testdata/db_space` line 41 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4703#-OQ58guS0MNzFhCZYWvn:-OQ58guS0MNzFhCZYWvo:b-f1j7i0) ([raw file](https://github.com/cockroachdb/pebble/blob/81bd7549c67f5b37018ad406b67a794eb050c221/tool/testdata/db_space#L41)):*
> ```
> 
> # covers the whole 000005.sst
> ```

What was the output before the code changes in this PR? I am finding it hard to reason about whether this output is correct. Some comments in the testdata file may help.
___
*[`db.go` line 504 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4703#-OQ55i0G5gPXGqwFV9R-:-OQ55i0HDu84yTo3BWzK:bqclq8z) ([raw file](https://github.com/cockroachdb/pebble/blob/81bd7549c67f5b37018ad406b67a794eb050c221/db.go#L504)):*
> ```Go
> 		// should be protected from concurrent access.
> 		annotators struct {
> 			totalDiskUsageSize *manifest.Annotator[uint64]
> ```

I think we include remote and external sstable sizes in here too, since the filter returns true in the call to to `makeFileSizeAnnotator`. Perhaps this should be called `totalFileSize`, and add a comment that it includes sstable and blob files.
___
*[`db.go` line 2401 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4703#-OQ56u81DAobO076Y6FT:-OQ56u81DAobO076Y6FU:b8h29r3) ([raw file](https://github.com/cockroachdb/pebble/blob/81bd7549c67f5b37018ad406b67a794eb050c221/db.go#L2401)):*
> ```Go
> 					overlapFraction := float64(overlappingFileSize) / float64(f.Size)
> 					// Scale the blob reference size proportionally to the file
> 					// overlap from the bounds to include only the blob
> ```

perhaps rephrase as:
... from the bounds to approximate only the blob ...

sumeerbhola (approved):
:lgtm: 

> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4703#-:-OQEYqGF0GG4US0o0Owr:bg7igkv)* status: all files reviewed (commit messages unreviewed), 4 unresolved discussions (waiting on @annrpom)

___
*[`tool/testdata/db_space` line 41 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4703#-OQ58guS0MNzFhCZYWvn:-OQEXu_J4xir1dh-_XsQ:b-d808bw) ([raw file](https://github.com/cockroachdb/pebble/blob/81bd7549c67f5b37018ad406b67a794eb050c221/tool/testdata/db_space#L41)):*

before the code changes in this PR:
```
db space --start=a --end=z
testdata/find-db-val-sep
----
814

db space --start=c --end=d
testdata/find-db-val-sep
----
159
```
> I am finding it hard to reason about whether this output is correct

Yeah that's fair -- perhaps it seems a bit wonky since everything is in the same data block (which is what `d.fileCache.estimateSize` ends up returning for the partial coverage calculation):

```
./pebble sstable layout tool/testdata/find-db-val-sep/000005.sst
tool/testdata/find-db-val-sep/000005.sst
sstable
 ├── data  offset: 0  length: 154
 ├── index  offset: 159  length: 36
 ├── properties  offset: 200  length: 501
 ├── meta-index  offset: 706  length: 46
 └── footer  offset: 757  length: 57
```

Either way, good point about comments; added the below (thoughts?):
```
# Covers the whole 000005.sst and referenced blob file 000006.blob. The size of
# the referenced blob file is 74 bytes. Note that this sst only has 1 data block
# with in bounds [a, d].
...
# Covers part of 000005.sst and part of referenced blob file 000006.blob. Note 
# that the estimated size of the sst from [c, d] is the size of the entire data 
# block + the estimated partial size of 000006.blob (14 bytes as more bytes are 
# in [a, b]).
...
```
</blockquote></details>

Looks like even the earlier existing tests in this file are wonky. IIUC, the problem you are referring to is that if we cover the whole file, the whole file size is included, but if we have to peer in (using the calculation in `Reader.EstimateDiskUsage`), and it happens to cover all the data blocks, we only include the size of the data blocks, which for tiny test sstables like this can be much smaller than the file size. Correct?

Fine to leave the existing test cases alone.

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4703)* status: all files reviewed (commit messages unreviewed), 4 unresolved discussions (waiting on @sumeerbhola)

___
*[`tool/testdata/db_space` line 41 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4703#-OQ58guS0MNzFhCZYWvn:-OQJZTV_3CXFqGrxH9y5:b-st6p9s) ([raw file](https://github.com/cockroachdb/pebble/blob/81bd7549c67f5b37018ad406b67a794eb050c221/tool/testdata/db_space#L41)):*

Looks like even the earlier existing tests in this file are wonky. IIUC, the problem you are referring to is that if we cover the whole file, the whole file size is included, but if we have to peer in (using the calculation in `Reader.EstimateDiskUsage`), and it happens to cover all the data blocks, we only include the size of the data blocks, which for tiny test sstables like this can be much smaller than the file size. Correct?

Fine to leave the existing test cases alone.
</blockquote></details>

yes, correct nod nod

TFTR! ('-')7

PR #4705, opened 2025-05-09
db: expand blob file LevelMetrics

jbowens:
Clarify the LevelMetrics fields related to byte quantities as bytes of
sstables by prefixing them with 'Table'.

Add totals about blob files to LevelMetrics and use them to correct the write
amplification calculation in the presence of value separation.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -68,27 +68,27 @@ type LevelMetrics struct {
 	FillFactor float64
 	// The level's compensated fill factor. See candidateLevelInfo.
 	CompensatedFillFactor float64
-	// The number of incoming bytes from other levels read during
+	// The number of incoming bytes from other levels' sstables read during
 	// compactions. This excludes bytes moved and bytes ingested. For L0 this is
 	// the bytes written to the WAL.
-	BytesIn uint64
-	// The number of bytes ingested. The sibling metric for tables is
+	TableBytesIn uint64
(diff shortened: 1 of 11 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4705#-:-OPqseBY6alkGjC4Tffu:bnfp4nl)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @xinhaoz)

metrics.go
```diff
@@ -68,27 +68,27 @@ type LevelMetrics struct {
...
-	// metrics for tables is TablesFlushed. This metric is always
-	// zero for all levels other than L0.
-	BytesFlushed uint64
+	TableBytesRead uint64
+	// The number of bytes written to sstables during compactions. The sibling
+	// metric for tables is TablesCompacted. This metric may be summed with
+	// BytesFlushed to compute the total bytes written for the level.
+	TableBytesCompacted uint64
+	// The number of bytes written to sstables during flushes. The sibling
+	// metrics for tables is TablesFlushed. This metric is always zero for all
```
RaduBerinde, line 89: Was this meant to say `BlobBytesCompacted` instead of `BytesFlushed`? We are missing a `BlobBytesCompacted` metric

PR #4707, opened 2025-05-09
ci: use go1.24

rickystewart:


rickystewart:
❓ Do we want to keep 1.23 in `.github/workflows/nightlies.yaml`?

RaduBerinde (approved):

RaduBerinde:
Yeah, let's keep 1.23 in the nightlies until we change `go.mod` to require 1.24 (after CRDB has switched).

rickystewart:
TFTR!

RaduBerinde:
We shouldn't have changed the go mod version here until after crdb switches over.

PR #4718, opened 2025-05-13
compact: move tombstone density compactions

### LINKED ISSUE #4161: compact: move tombstone density compactions
The move compaction optimization doesn't apply to compactions picked as tombstone density compactions, resulting in unnecessarily rewriting sstables that could've been trivially moved to the next level.

```
I241101 19:33:30.159675 12035713 3@pebble/event.go:776 ⋮ [n30,s141,pebble] 67372  [JOB 2477] compacted(tombstone-density) L4 [5400119] (4.0MB) Score=0.00 + L5 [] (0B) Score=0.00 -> L5 [5400166] (4.0MB), in 0.0s (0.0s total), output rate 98MB/s
```

Jira issue: PEBBLE-301

xxmplus:
### Overview

This PR introduces a performance optimization for tombstone density compactions in Pebble. The optimization allows certain tombstone density compactions to be converted into a move operation, avoiding unnecessary file rewriting and improving compaction efficiency.

Fixes #4161 

---

#### 1. **Compaction Move Optimization Logic**
- **`compaction.go`**:  
  - Enhanced the compaction selection logic to allow tombstone density compactions (not just default compactions) to be optimized as a move when:
    - The compaction is of kind `tombstone-density` (and not targeting the last level).
    - There is a single input file with high tombstone density.
    - The output level has no overlapping files.
    - The grandparent overlap is below the configured threshold.
  - This change enables more efficient handling of files with high tombstone density by moving them instead of rewriting.

#### 2. **Unit Tests for Move Optimization**
- **`TestTombstoneDensityCompactionMoveOptimization`**: Verifies that the move optimization is applied when all criteria are met.
- **`TestTombstoneDensityCompactionMoveOptimization_NoMoveWithOverlap`**: Ensures the optimization is not applied if there are overlapping files in the output level.
- **`TestTombstoneDensityCompactionMoveOptimization_GrandparentOverlapTooLarge`**: Ensures the optimization is not applied if the grandparent overlap exceeds the threshold.
- **`TestTombstoneDensityCompactionMoveOptimization_BelowDensityThreshold`**: Ensures the optimization is not applied if the tombstone density is below the threshold.
- **`TestTombstoneDensityCompactionMoveOptimization_InvalidStats`**: Ensures the optimization is not applied if file stats are missing or invalid.

#### 3. **Data-Driven Tests**
- **`data_test.go`**:  
  - Added support for a `force-tombstone-density-ratio` argument in the data-driven test framework, allowing tests to simulate files with a specific tombstone density ratio for more precise testing and documentation.
- **`testdata/compaction_tombstones`**:  
  - Added detailed documentation and test cases demonstrating:
    - When the move optimization is applied (single input file, no overlap, low grandparent overlap).
    - When the optimization is not applied (overlap in output level, large grandparent overlap).
    - How the tombstone density ratio can be forced in tests for demonstration purposes.
  - These cases help clarify the feature's behavior and its limitations in the data-driven test environment.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -412,7 +412,8 @@ func newCompaction(
 	)
 	c.kind = pc.kind
 
-	if c.kind == compactionKindDefault && c.outputLevel.files.Empty() && !c.hasExtraLevelData() &&
+	if (c.kind == compactionKindDefault || (c.kind == compactionKindTombstoneDensity && c.outputLevel.level != numLevels-1)) &&
+		c.outputLevel.files.Empty() && !c.hasExtraLevelData() &&
 		c.startLevel.files.Len() == 1 && c.grandparents.AggregateSizeSum() <= c.maxOverlapBytes {
 		// This compaction can be converted into a move or copy from one level
 		// to the next. We avoid such a move if there is lots of overlapping
--- a/compaction.go
+++ b/compaction.go
@@ -412,6 +412,12 @@ func newCompaction(
 	)
 	c.kind = pc.kind
 
+	// In addition to the default compaction, we also check whether a tombstone density compaction can be optimized into
+	// a move compaction. However, we want to avoid performing a move compaction into the lowest level, since the goal
+	// there is to actually remove the tombstones.
+	//
+	// Tombstone density compaction is meant to address cases where tombstones don't reclaim much space but are still
+	// expensive to scan over. We can only remove the tombstones once there's nothing at all underneath them.
 	if (c.kind == compactionKindDefault || (c.kind == compactionKindTombstoneDensity && c.outputLevel.level != numLevels-1)) &&
 		c.outputLevel.files.Empty() && !c.hasExtraLevelData() &&
 		c.startLevel.files.Len() == 1 && c.grandparents.AggregateSizeSum() <= c.maxOverlapBytes {
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -3199,8 +3199,8 @@ func TestTombstoneDensityCompactionMoveOptimization(t *testing.T) {
 
 	// Create a file with high tombstone density.
 	meta := &tableMetadata{
-		FileNum: 1,
-		Size:    1024,
+		TableNum: 1,
+		Size:     1024,
 		Stats: manifest.TableStats{
 			NumEntries:                10,
 			NumDeletions:              8,
@@ -3245,7 +3245,7 @@ func TestTombstoneDensityCompactionMoveOptimization(t *testing.T) {
 	iter := c.startLevel.files.Iter()
 	inFile := iter.First()
 	require.NotNil(t, inFile)
-	require.Equal(t, meta.FileNum, inFile.FileNum, "file should be the one planned for move")
+	require.Equal(t, meta.TableNum, inFile.TableNum, "file should be the one planned for move")
 
 	t.Logf("Compaction kind: %v", c.kind)
 	t.Logf("Input file: %v", meta)
@@ -3261,7 +3261,7 @@ func TestTombstoneDensityCompactionMoveOptimization(t *testing.T) {
 	for meta := i.First(); meta != nil; meta = i.Next() {
 		ve.DeletedTables[manifest.DeletedTableEntry{
 			Level:   inputLevel,
-			FileNum: meta.FileNum,
+			FileNum: meta.TableNum,
 		}] = meta
 		ve.NewTables = append(ve.NewTables, manifest.NewTableEntry{
 			Level: outputLevel,
@@ -3297,8 +3297,8 @@ func TestTombstoneDensityCompactionMoveOptimization_NoMoveWithOverlap(t *testing
 
 	// Create a file with high tombstone density in L4.
 	metaL4 := &tableMetadata{
-		FileNum: 1,
-		Size:    1024,
+		TableNum: 1,
+		Size:     1024,
 		Stats: manifest.TableStats{
 			NumEntries:                10,
 			NumDeletions:              8,
@@ -3314,8 +3314,8 @@ func TestTombstoneDensityCompactionMoveOptimization_NoMoveWithOverlap(t *testing
 
 	// Create an overlapping file in L5.
 	metaL5 := &tableMetadata{
-		FileNum: 2,
-		Size:    1024,
+		TableNum: 2,
+		Size:     1024,
 	}
 	metaL5.ExtendPointKeyBounds(opts.Comparer.Compare,
 		base.ParseInternalKey("m.SET.1"),
@@ -3376,8 +3376,8 @@ func TestTombstoneDensityCompactionMoveOptimization_GrandparentOverlapTooLarge(t
 
 	// File in L4 with high tombstone density.
 	metaL4 := &tableMetadata{
-		FileNum: 1,
-		Size:    1024,
+		TableNum: 1,
+		Size:     1024,
 		Stats: manifest.TableStats{
 			NumEntries:                10,
 			NumDeletions:              8,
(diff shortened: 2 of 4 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4718)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @annrpom)

___
*[`compaction.go` line 415 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4718#-OQAra3A5yqv3xGoJSt5:-OQAra3A5yqv3xGoJSt6:brqterv) ([raw file](https://github.com/cockroachdb/pebble/blob/db1477420852a6aa0e68db4cd7aac978570b2b00/compaction.go#L415)):*
> ```Go
> 	c.kind = pc.kind
> 
> 	if (c.kind == compactionKindDefault || (c.kind == compactionKindTombstoneDensity && c.outputLevel.level != numLevels-1)) &&
> ```

Why the output level check?  I guess it wouldn't make much sense to have a tombstone density compaction that doesn't have anything underneath.. still, this doesn't seem like the place to check for it

xxmplus (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4718)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @annrpom and @RaduBerinde)

___
*[`compaction.go` line 415 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4718#-OQAra3A5yqv3xGoJSt5:-OQFZDiCArvGaxWJlnn8:b-h8z460) ([raw file](https://github.com/cockroachdb/pebble/blob/db1477420852a6aa0e68db4cd7aac978570b2b00/compaction.go#L415)):*

Why the output level check?  I guess it wouldn't make much sense to have a tombstone density compaction that doesn't have anything underneath.. still, this doesn't seem like the place to check for it
</blockquote></details>

It's to make sure that the tombstone density compaction rather than move is chosen when L5 -> L6. It should read "when L5 -> L6, it doesn't make sense to move, I will keep the tombstone density compaction."

RaduBerinde:
*[`compaction.go` line 415 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4718#-OQAra3A5yqv3xGoJSt5:-OQFyhfa39yrGKxXq8d5:brncaoc) ([raw file](https://github.com/cockroachdb/pebble/blob/db1477420852a6aa0e68db4cd7aac978570b2b00/compaction.go#L415)):*
<details><summary><i>Previously, xxmplus (Xuming Xu) wrote…</i></summary><blockquote>

It's to make sure that the tombstone density compaction rather than move is chosen when L5 -> L6. It should read "when L5 -> L6, it doesn't make sense to move, I will keep the tombstone density compaction."
</blockquote></details>

@jbowens wouldn't `compactionKindRead` also work here?

More generally, it feels like the `compactionKind` conflates two separate things that we should separate:
 - the mechanics of the compaction (regular, move, copy, etc)
 - the reason the compaction was chosen

RaduBerinde:
*[`compaction.go` line 415 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4718#-OQAra3A5yqv3xGoJSt5:-OQG2HyL9zKWVzkJH4RO:b-dhvbxv) ([raw file](https://github.com/cockroachdb/pebble/blob/db1477420852a6aa0e68db4cd7aac978570b2b00/compaction.go#L415)):*
> It's to make sure that the tombstone density compaction rather than move is chosen when L5 -> L6. It should read "when L5 -> L6, it doesn't make sense to move, I will keep the tombstone density compaction."

We only convert to a move compaction if there is no overlapping file in the output level (`c.outputLevel.files.Empty()`). In this case, there shouldn't be a `compactionKindTombstoneDensity` anyway (and even if there was, I don't see the problem with moving the file(s) just like on other levels).

xxmplus (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4718)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @annrpom and @jbowens)

___
*[`compaction.go` line 415 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4718#-OQAra3A5yqv3xGoJSt5:-OQGCKK7DHZIF0rAqzFW:b-w6p6i9) ([raw file](https://github.com/cockroachdb/pebble/blob/db1477420852a6aa0e68db4cd7aac978570b2b00/compaction.go#L415)):*

> It's to make sure that the tombstone density compaction rather than move is chosen when L5 -> L6. It should read "when L5 -> L6, it doesn't make sense to move, I will keep the tombstone density compaction."

We only convert to a move compaction if there is no overlapping file in the output level (`c.outputLevel.files.Empty()`). In this case, there shouldn't be a `compactionKindTombstoneDensity` anyway (and even if there was, I don't see the problem with moving the file(s) just like on other levels).
</blockquote></details>

I see. Is it correct to say that it's acceptable to leave an SSTable with a high tombstone density, even after it's been moved to the lowest level?

Regarding the first comment, it sounds like a major rewrite would be required to separate that. If we were to do it, should we do it in a separate pr?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4718)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @annrpom and @jbowens)

___
*[`compaction.go` line 415 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4718#-OQAra3A5yqv3xGoJSt5:-OQGHEoV1I36ZpSXl6r2:b-u9wj9y) ([raw file](https://github.com/cockroachdb/pebble/blob/db1477420852a6aa0e68db4cd7aac978570b2b00/compaction.go#L415)):*
<details><summary><i>Previously, xxmplus (Xuming Xu) wrote…</i></summary><blockquote>

I see. Is it correct to say that it's acceptable to leave an SSTable with a high tombstone density, even after it's been moved to the lowest level?

Regarding the first comment, it sounds like a major rewrite would be required to separate that. If we were to do it, should we do it in a separate pr?
</blockquote></details>

Yeah, AFAIU tombstone density is used as an indicator of data going away when the table is merged with a table underneath.

Yeah, changing the kind would be a separate PR (I'm not even suggesting doing it now or anytime soon, but it's always good to look out for things that can be improved).

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4718)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @annrpom, @RaduBerinde, and @xxmplus)

___
*[`compaction.go` line 415 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4718#-OQAra3A5yqv3xGoJSt5:-OQGg-yP6573DpRN4e-0:b-fexw2r) ([raw file](https://github.com/cockroachdb/pebble/blob/db1477420852a6aa0e68db4cd7aac978570b2b00/compaction.go#L415)):*

Yeah, AFAIU tombstone density is used as an indicator of data going away when the table is merged with a table underneath.

Yeah, changing the kind would be a separate PR (I'm not even suggesting doing it now or anytime soon, but it's always good to look out for things that can be improved).
</blockquote></details>

I think we want to avoid a move compaction into L6 in this case because we want to remove the tombstones themselves. The tombstone density compaction strives to fill in a gap where the tombstones don't reclaim much space, but there's many of them so they can still be expensive to scan over. We can only remove the tombstones once there's nothing at all underneath them. I guess the most principled version of this would be that we refuse to do a move compaction if the in-use key range beneath the file is empty––that would ensure that we still drop the tombstones when we can.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4718#-:-OQGmicHDb4TctVgj8_A:bnfp4nl)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @annrpom, @jbowens, and @xxmplus)

___
*[`compaction.go` line 415 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4718#-OQAra3A5yqv3xGoJSt5:-OQGmdXx8obs1qilPrnJ:bth5mwe) ([raw file](https://github.com/cockroachdb/pebble/blob/db1477420852a6aa0e68db4cd7aac978570b2b00/compaction.go#L415)):*

I think we want to avoid a move compaction into L6 in this case because we want to remove the tombstones themselves. The tombstone density compaction strives to fill in a gap where the tombstones don't reclaim much space, but there's many of them so they can still be expensive to scan over. We can only remove the tombstones once there's nothing at all underneath them. I guess the most principled version of this would be that we refuse to do a move compaction if the in-use key range beneath the file is empty––that would ensure that we still drop the tombstones when we can.
</blockquote></details>

Makes sense - let's add that as a comment

PR #4728, opened 2025-05-15
db: remove deletedFileEntry alias

RaduBerinde:


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4728)* status: all files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`version_set.go` line 29 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4728#-OQIsZeWBtSX4gNZkVA0:-OQJeA3VEVUE6LiBsEVr:b-9sw6g7) ([raw file](https://github.com/cockroachdb/pebble/blob/939f3193b60696df444b3c76e07f0020954ea87c/version_set.go#L29)):*

just curious: any reason to remove this one but not the others? Seems like the only reason these were defined is to reduce the length of these names, which I personally find dubious (others may have different opinions though, so not wanting to bikeshed).
</blockquote></details>

I strongly agree, we should remove all of them. We save a few characters but introduce confusion when reading the code. I just didn't want to make a huge change right now (thinking of merge issues during backports).

This one is worse because it has the wrong name (file vs table).

RaduBerinde:
TFTR!

PR #4731, opened 2025-05-16
db: don't require quotes around level number in options section

RaduBerinde:
The current format requires per-level sections to look like
`[Level "1"]`. The quotes can be a pain to escape, especially when
this is passed through the `--store` command line.

This change makes the parsing more tolerant to allow but not require
quotes, so that `[Level 1]` also works.

### DIFF
--- a/options.go
+++ b/options.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"regexp"
 	"runtime"
 	"sort"
 	"strconv"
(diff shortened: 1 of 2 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4731#-:-OQOgq6b9Qr3fzvn8h89:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

PR #4733, opened 2025-05-16
crl-release-25.2: db: don't require quotes around level number in options section

RaduBerinde:
Backport of #4731.

The current format requires per-level sections to look like
`[Level "1"]`. The quotes can be a pain to escape, especially when
this is passed through the `--store` command line.

This change makes the parsing more tolerant to allow but not require
quotes, so that `[Level 1]` also works.

### DIFF
--- a/options.go
+++ b/options.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"regexp"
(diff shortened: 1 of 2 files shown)


jbowens (approved):

RaduBerinde:
TFTR!

Issue #4581, opened 2025-04-18
db: adapt Metrics for value separation


jbowens:
There are a few updates to `pebble.Metrics` that we should make to account for value separation.

- [ ] We should update `Metrics.String()` to render the metrics included under `Metrics.BlobFIles`.
- [ ] We should update `Metrics.DiskSpaceUsage()` to include disk space usage of blob files (local live, zombie and obsolete).
- [ ] We should add fields to `LevelMetrics` that represent not just the size of the physical sstables but also the referenced value sizes. We should include one that includes the ReferencedValueSize across all sstables in a level. Maybe we should also add an ApproximatePhysicalSize that is the sum of the physical sstable sizes, plus the referenced value sizes scaled according to each value's compression ratio? We'll also need this sum internally for compaction level scoring. Many of the uses of `TableSize` in the cockroach repository should be updated to use `ApproximatePhysicalSize`

PR #4739, opened 2025-05-19
db: account blob files in Metrics.DiskSpaceUsage

jbowens:
Account blob files' sizes when calculating disk space in (*Metrics).DiskSpaceUsage.

Informs #4581.


RaduBerinde (approved):

jbowens:
TFTR!

PR #4747, opened 2025-05-20
github: fix bug in post-issue

RaduBerinde:
Change the "here-doc" `<<EOF` to `<<'EOF'` which prevents escapes and
expansion.


xinhaoz (approved):

RaduBerinde:
TFTR!

Issue #4459, opened 2025-03-31
replay: support blob files


jbowens:
The replay package handles capturing and replaying a workload through monitoring ingestions and flushes of a pebble database. With the introduction of value separation placing values external to a sstable in a blob file, capturing the flushed sstables is insufficient. We'll need to also capture the blob files constructed by a flush.

Jira issue: PEBBLE-375

Epic CRDB-20379

PR #4752, opened 2025-05-21
replay: support blob files 

annrpom:
This patch ensures that `replay` is able to monitor & replay flushes with
value separation enabled.

Fixes: https://github.com/cockroachdb/pebble/issues/4459

### DIFF
--- a/file_cache.go
+++ b/file_cache.go
@@ -24,11 +24,13 @@ import (
 	"github.com/cockroachdb/pebble/internal/manifest"
 	"github.com/cockroachdb/pebble/internal/sstableinternal"
 	"github.com/cockroachdb/pebble/objstorage"
+	"github.com/cockroachdb/pebble/objstorage/objstorageprovider"
 	"github.com/cockroachdb/pebble/objstorage/objstorageprovider/objiotracing"
 	"github.com/cockroachdb/pebble/sstable"
 	"github.com/cockroachdb/pebble/sstable/blob"
 	"github.com/cockroachdb/pebble/sstable/block"
 	"github.com/cockroachdb/pebble/sstable/valblk"
+	"github.com/cockroachdb/pebble/vfs"
 	"github.com/cockroachdb/redact"
 )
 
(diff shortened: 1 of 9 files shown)


jbowens (commented):
> Reviewed 8 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4752)* status: 8 of 9 files reviewed, 4 unresolved discussions

___
*[`file_cache.go` line 731 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4752#-OQtKxjv9zegE4Ucc7X9:-OQtKxjv9zegE4Ucc7XA:b-hfaiuo) ([raw file](https://github.com/cockroachdb/pebble/blob/4fc88f189abf8678a741760751925f1d9d3163c3/file_cache.go#L731)):*
> ```Go
> 
> // SetupBlobReaderProvider creates a fileCachHandle blob.ReaderProvider for
> // reading blob files. The caller is responsible for calling the returned cleanup
> ```

nit: I think we should say something about this func being used by tests and tools that need to read blob files outside of normal database operation, and that this func is _not_ used by a database opened through Open.
___
*[`internal/manifest/version.go` line 115 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4752#-OQtRmVB1u0sQv_3KtG6:-OQtRmVB1u0sQv_3KtG7:b3sufm2) ([raw file](https://github.com/cockroachdb/pebble/blob/4fc88f189abf8678a741760751925f1d9d3163c3/internal/manifest/version.go#L115)):*
> ```Go
> 	LargestSeqNum base.SeqNum
> 	// BlobReferences is the list of blob files referenced by the table.
> 	BlobReferences BlobReferences
> ```

the  `manifest.BlobFileMetadata`/`BlobReferences` types have some internal DB state that we probably don't want users accessing. could we unexport this field and add a method to TableInfo that returns the `[]base.DiskFileNum`s of the references? 
___
*[`replay/replay.go` line 1011 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4752#-OQtT8WBCqeS-oeY4AYW:-OQtT8WBCqeS-oeY4AYX:bdtfg16) ([raw file](https://github.com/cockroachdb/pebble/blob/4fc88f189abf8678a741760751925f1d9d3163c3/replay/replay.go#L1011)):*
> ```Go
> 	bufs *flushBuffers,
> ) error {
> 	provider, closeFunc, err := pebble.SetupBlobReaderProvider(fs, path, opts, readOpts)
> ```

can we create this up front, in the outer loop so that we reuse the same file cache throughout replay?

also I think we should configure Options.MaxOpenFiles to some small constant like 10 when calling `SetupBlobReaderProvider`. The file cache won't evict anything until it reaches `MaxOpenFiles`. When we're replaying, we'll only read each file once when replaying its flush and then never again. So we don't want the same large `MaxOpenFiles` value that we configure the DB with
___
*[`replay/workload_capture.go` line 238 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4752#-OQtVFdO0SJ73V4ncQZS:-OQtVFdPEM2NOItbPwwg:bfwg3sn) ([raw file](https://github.com/cockroachdb/pebble/blob/4fc88f189abf8678a741760751925f1d9d3163c3/replay/workload_capture.go#L238)):*
> ```Go
> 			// Copy any updates to the manifests files.
> 			w.copyManifests(index, pendingManifests)
> 			// Copy the SSTables provided in pending. copySSTables takes
> ```

nit: s/copySSTables/copySSTablesAndBlobs/

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4752)* status: 4 of 9 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`file_cache.go` line 731 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4752#-OQtKxjv9zegE4Ucc7X9:-OQtYj9A3Ui2550pDO23:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/4fc88f189abf8678a741760751925f1d9d3163c3/file_cache.go#L731)):*

nit: I think we should say something about this func being used by tests and tools that need to read blob files outside of normal database operation, and that this func is _not_ used by a database opened through Open.
</blockquote></details>

done
___
*[`internal/manifest/version.go` line 115 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4752#-OQtRmVB1u0sQv_3KtG6:-OQt_TQ3CdAE4iMndZHg:b9rt37s) ([raw file](https://github.com/cockroachdb/pebble/blob/4fc88f189abf8678a741760751925f1d9d3163c3/internal/manifest/version.go#L115)):*

the  `manifest.BlobFileMetadata`/`BlobReferences` types have some internal DB state that we probably don't want users accessing. could we unexport this field and add a method to TableInfo that returns the `[]base.DiskFileNum`s of the references?
</blockquote></details>

done

for posterity from offline convo: we were thinking about the state that was later moved into `CurrentBlobFileSet`, but still unexporting and adding a method 
___
*[`replay/replay.go` line 1011 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4752#-OQtT8WBCqeS-oeY4AYW:-OQtd4nX4UzTEGiEuu17:b-lks4zn) ([raw file](https://github.com/cockroachdb/pebble/blob/4fc88f189abf8678a741760751925f1d9d3163c3/replay/replay.go#L1011)):*

can we create this up front, in the outer loop so that we reuse the same file cache throughout replay?

also I think we should configure Options.MaxOpenFiles to some small constant like 10 when calling `SetupBlobReaderProvider`. The file cache won't evict anything until it reaches `MaxOpenFiles`. When we're replaying, we'll only read each file once when replaying its flush and then never again. So we don't want the same large `MaxOpenFiles` value that we configure the DB with
</blockquote></details>

yes, good point done to both: we now create the provider before the loop over our manifests
___
*[`replay/workload_capture.go` line 238 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4752#-OQtVFdO0SJ73V4ncQZS:-OQthrsICs2H3qZWLkn4:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/4fc88f189abf8678a741760751925f1d9d3163c3/replay/workload_capture.go#L238)):*

nit: s/copySSTables/copySSTablesAndBlobs/
</blockquote></details>

done

jbowens (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4752#-:-OQtwJlzCDoBZWUUM9oP:b-uwkowd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @annrpom)

annrpom:
TFTR! ('-')7

Issue #4732, opened 2025-05-16
investigate and fix gap in crossversion tests


RaduBerinde:
https://github.com/cockroachdb/pebble/issues/4727 was an egregious 24.2->master update bug but was not caught by crossversion tests. Investigate why and fix.

One possibility is that we never upgrade database formats so all our table formats are as old as the first build in the chain.

RaduBerinde:
Perhaps randomizing the chain length rather than always starting with the same first version would help too.

RaduBerinde:
It looks like the test is completely broken. When TestMetaCrossVersion runs TestMeta, it does not pass a `--previous-ops` flag:
https://github.com/cockroachdb/pebble/blob/master/internal/metamorphic/crossversion/crossversion_test.go#L380

The initial state flags are silently ignored if `--previous-ops` flag isn't passed:
https://github.com/cockroachdb/pebble/blob/master/internal/metamorphic/metaflags/meta_flags.go#L301

I tried a manual run with the right flags but immediately ran into issues that have to do with wal failover:
`reading "_meta/250520-195135.4683254040531/standard-024/data/wal": open wal/: file does not exist`
or
`directory "" may contain relevant WALs`

PR #4755, opened 2025-05-21
db: support store-relative paths for WAL dirs

RaduBerinde:
#### db: support store-relative paths for WAL dirs

Relative WAL paths (including the actual WAL, the failover path, and
the recovery paths) are (unfortunately) interpreted as relative to the
current working directory.

The cross-version metamorphic test copies a store from a previous run
as the initial state for a new test. The options will fail the
compatibility check since the path changes.

This change adds support for using a special `{store_path}` prefix to
the path. Any such prefix is replaced with the store directory.

We also improve the missing WAL recovery dir error to show what
directories are actually configured.

#### metamorphic: fix code around WAL recovery directories

Informs: #4732

### DIFF
--- a/metamorphic/meta.go
+++ b/metamorphic/meta.go
@@ -540,19 +540,13 @@ func RunOnce(t TestingT, runDir string, seed uint64, historyPath string, rOpts .
 			testOpts.Opts.WALFailover = nil
 		} else {
 			testOpts.Opts.WALFailover.Secondary.FS = opts.FS
-			testOpts.Opts.WALFailover.Secondary.Dirname = opts.FS.PathJoin(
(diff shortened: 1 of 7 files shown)


jbowens (approved):
:lgtm:

> Reviewed 5 of 5 files at r1, 2 of 2 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4755#-:-OQtJtUn1aC1NDJ1DrLm:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sumeerbhola)

RaduBerinde:
TFTR!

Issue #4383, opened 2025-03-14
sstable: reduce memory consumption from properties


jbowens:
An open Reader holds its `Properties` in-memory. Over time, we've accumulated many properties, and the memory usage of the `Properties` struct and its `UserProperties` and `Loaded` maps is non trivial. Especially with higher node densities, per open-sstable memory usage can be substantial. ~1 million sstables is reasonable at >10TB node densities: possibly much more in multi-store arrangements where each LSM is smaller, and the average sstable size is much smaller. If we consume ~500 bytes of memory for 1 million sstables, that translates to 512mb of memory.

Typically, we don't need these properties to be in-memory. Many of the properties are used to compute TableStats that are stored on the (also in-memory) `manifest.TableMetadata` and then never consulted again. Other properties are never read except on-demand by tooling (eg, through crdb_internal builtins). Additionally, the Properties struct contains two maps today which introduce two additional pointers per open sstable that the Go GC must chase.

There are a few ways to reduce the memory usage. Most of the properties we could load only on demand.

Jira issue: PEBBLE-351

Epic CRDB-49141

jbowens:
See #2047 for an analogous issue for the in-memory `TableMetadata`

jbowens:
I think step one here is to pull out this logic to read properties from the file out into its own ReadProperties method on the `sstable.Reader`. In `readMetaindex`, we'll also want to store the properties block handle `meta[metaPropertiesName]` as a new field on the `sstable.Reader` so that the new `ReadProperties` func can read it:

https://github.com/cockroachdb/pebble/blob/716613c26cea749670c7bb6540a1df9691020bae/sstable/reader.go#L484-L504

PR #4756, opened 2025-05-21
sstable: introduce ReadPropertiesBlock on reader

xinhaoz:
### sstable: introduce ReadPropertiesBlock on reader

Introduces ReadPropertiesBlock on the reader struct to prep for loading
properties on demand from various callers. For now we continue to populate
`Properties` on reader creation.

### tools: load properties on demand

tools: load properties on demand
To support the goal of removing the props struct from the reader, this commit
loads properties on demand when running the following commands instead of
reading from the in-memory struct:
- runCheck
- runProperties (sstable + db level)
- runLayout
- runScan

Part of: #4383

### DIFF
--- a/sstable/reader.go
+++ b/sstable/reader.go
@@ -482,16 +482,11 @@ func (r *Reader) readMetaindex(
 	}
 
 	if bh, ok := meta[metaPropertiesName]; ok {
-		b, err = r.blockReader.Read(ctx, metaEnv, readHandle, bh, noInitBlockMetadataFn)
-		if err != nil {
-			return err
-		}
 		r.propertiesBH = bh
-		r.Properties, err = decodePropertiesBlock(r.tableFormat, b.BlockData(), deniedUserProperties)
+		r.Properties, err = r.readPropertiesBlockInternal(ctx, metaEnv.BufferPool, readHandle, deniedUserProperties)
 		if err != nil {
 			return err
 		}
-		b.Release()
 	} else {
 		return errors.New("did not read any value for the properties block in the meta index")
 	}
@@ -544,6 +539,48 @@ func decodePropertiesBlock(
 		}
 	}
 	return props, nil
+}
+
+var PropertiesBlockBufPools = sync.Pool{
+	New: func() any {
+		bp := new(block.BufferPool)
+		// New pools are initialized with a capacity of 2 to accommodate
+		// both the compressed properties block (1) and decompressed
+		// properties block (1).
+		bp.Init(2)
+		return bp
+	},
+}
+
+// ReadPropertiesBlock reads the properties block from the table.
+// We always read the properties block into a buffer pool instead
+// of the block cache.
+func (r *Reader) ReadPropertiesBlock(
+	ctx context.Context, bufferPool *block.BufferPool, deniedUserProperties map[string]struct{},
+) (Properties, error) {
+	return r.readPropertiesBlockInternal(ctx, bufferPool, noReadHandle, deniedUserProperties)
+}
+
+func (r *Reader) readPropertiesBlockInternal(
+	ctx context.Context,
+	bufferPool *block.BufferPool,
+	readHandle objstorage.ReadHandle,
+	deniedUserProperties map[string]struct{},
+) (Properties, error) {
+	if bufferPool == nil {
+		// We always use a buffer pool when reading the properties block as
+		// we don't want it in the block cache.
+		bufferPool = PropertiesBlockBufPools.Get().(*block.BufferPool)
+		defer PropertiesBlockBufPools.Put(bufferPool)
+		defer bufferPool.Release()
+	}
+	env := block.ReadEnv{BufferPool: bufferPool}
+	b, err := r.blockReader.Read(ctx, env, readHandle, r.propertiesBH, noInitBlockMetadataFn)
+	if err != nil {
+		return Properties{}, err
+	}
+	defer b.Release()
+	return decodePropertiesBlock(r.tableFormat, b.BlockData(), deniedUserProperties)
 }
 
 // Layout returns the layout (block organization) for an sstable.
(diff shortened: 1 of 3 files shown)


xinhaoz:
Putting this up first for comments on the API.
Next: https://github.com/cockroachdb/pebble/pull/4757

sstable/reader.go
```diff
@@ -546,6 +541,48 @@ func decodePropertiesBlock(
 	return props, nil
 }
 
+var PropertiesBlockBufPools = sync.Pool{
```
xinhaoz, line 544: Not really sure about this or if we should just introduce a generic BufferPool pool... lmk if there's any other way we'd like to deal with the allocations of the buf pool required to read the properties block.

jbowens: could we get away with not exporting this, relying on callers who don't have a `BufferPool` ready passing nil and letting `readPropertiesBlockInternal` fetch the buffer pool?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4756)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @xinhaoz)

___
*[`tool/db.go` line 981 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4756#-OQt7CmMEuXbyKOTnrAi:-OQt7CmNCRYccFdzr8oE:bpj485t) ([raw file](https://github.com/cockroachdb/pebble/blob/33438dbeccb62ce6032497a9262c86b62075c946/tool/db.go#L981)):*
> ```Go
> 	properties, err := r.ReadPropertiesBlock(context.TODO(), nil /* buffer pool */, opts.DeniedUserProperties)
> 	if err != nil {
> 		_ = f.Close()
> ```

here I think we want to close the `Reader` instead (which will close the `Readable` (`f` too); so we should do `return errors.CombineErrors(err, r.Close())`
___
*[`tool/sstable.go` line 382 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4756#-OQt8c5E7_lThh6pkgJ8:-OQt8c5E7_lThh6pkgJ9:bfhtbjn) ([raw file](https://github.com/cockroachdb/pebble/blob/33438dbeccb62ce6032497a9262c86b62075c946/tool/sstable.go#L382)):*
> ```Go
> 	defer sstable.PropertiesBlockBufPools.Put(bp)
> 	deniedUserProps := s.opts.MakeReaderOptions().DeniedUserProperties
> 	s.foreachSstable(stderr, args, func(path string, r *sstable.Reader) {
> ```

wdyt about reading the properties block in `foreachSstable` and passing the props into this closure to reduce the repetition

sstable/reader.go
xinhaoz: Done.

xinhaoz (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4756)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`tool/db.go` line 981 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4756#-OQt7CmMEuXbyKOTnrAi:-OQt_RxH3IKjsC89IGqO:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/33438dbeccb62ce6032497a9262c86b62075c946/tool/db.go#L981)):*

here I think we want to close the `Reader` instead (which will close the `Readable` (`f` too); so we should do `return errors.CombineErrors(err, r.Close())`
</blockquote></details>

Done.
___
*[`tool/sstable.go` line 382 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4756#-OQt8c5E7_lThh6pkgJ8:-OQt_T3L6wOb5UL4F-gf:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/33438dbeccb62ce6032497a9262c86b62075c946/tool/sstable.go#L382)):*

wdyt about reading the properties block in `foreachSstable` and passing the props into this closure to reduce the repetition
</blockquote></details>

Done.

sstable/reader.go
xinhaoz: Done.

xinhaoz (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4756)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`tool/db.go` line 981 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4756#-OQt7CmMEuXbyKOTnrAi:-OQt_RxH3IKjsC89IGqO:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/33438dbeccb62ce6032497a9262c86b62075c946/tool/db.go#L981)):*

here I think we want to close the `Reader` instead (which will close the `Readable` (`f` too); so we should do `return errors.CombineErrors(err, r.Close())`
</blockquote></details>

Done.
___
*[`tool/sstable.go` line 382 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4756#-OQt8c5E7_lThh6pkgJ8:-OQt_T3L6wOb5UL4F-gf:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/33438dbeccb62ce6032497a9262c86b62075c946/tool/sstable.go#L382)):*

wdyt about reading the properties block in `foreachSstable` and passing the props into this closure to reduce the repetition
</blockquote></details>

Done.

jbowens (approved):
:lgtm_strong: 

> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4756#-:-OQtjt2FC_PmMPN1rPSZ:b-uwkowd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @xinhaoz)

xinhaoz:
TFTR!!

PR #4757, opened 2025-05-21
sstable: save UserProperties from Properties on sstable reader

xinhaoz:
### sstable: remove DeniedUserProperties as a reader opt
Let's just always use the `ignoredInternalProperties`.

----------------------------

### sstable: save UserProperties from Properties on sstable reader
To support removing the `Properties` field from the reader, we should save at
least the UserProperties to support checking the intersection of filters when
creating RangeKey and Point iterators

Part of: #4383

### DIFF
--- a/sstable/reader.go
+++ b/sstable/reader.go
@@ -65,9 +65,10 @@ type Reader struct {
 	metaindexBH  block.Handle
 	footerBH     block.Handle
 
-	Properties  Properties
-	tableFormat TableFormat
-	Attributes  Attributes
+	Properties     Properties
+	tableFormat    TableFormat
+	Attributes     Attributes
+	UserProperties map[string]string
 }
 
 type ReadEnv struct {
@@ -486,6 +487,7 @@ func (r *Reader) readMetaindex(
 		if err != nil {
 			return err
 		}
+		r.UserProperties = r.Properties.UserProperties
 	} else {
 		return errors.New("did not read any value for the properties block in the meta index")
 	}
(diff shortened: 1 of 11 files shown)


sstable/reader.go
```diff
@@ -554,13 +558,16 @@ var propertiesBlockBufPools = sync.Pool{
...
-// of the block cache.
-func (r *Reader) ReadPropertiesBlock(
-	ctx context.Context, bufferPool *block.BufferPool, deniedUserProperties map[string]struct{},
+// ReadPropertiesBlockExcludeUserProps reads the properties block
+// from the table. We always read the properties block into a buffer pool
+// instead of the block cache. Note that UserProperties should not be read
+// from returned properties as they do not filter out denied user properties
+// (those that pebble doesn't care about). UserProperties should be read
+// directly from the Reader struct.
+func (r *Reader) ReadPropertiesBlockExcludeUserProps(
```
xinhaoz, line 567: It was a little cumbersome to have callers pass this map in when most of them don't read `UserProperties`. I opted to change this function to exclude it for the common case but I might have to introduce it again later (maybe as a separate fn?) as there at least one place ([getTableProperties in filecache](https://github.com/cockroachdb/pebble/blob/master/file_cache.go#L864)) where we want a complete correctly populated `Properties`.

I could also keep the param and export `noDeniedUserProperties`. LMK what y'all think.

xinhaoz: Followed up with @jbowens offline and we can probably just get rid of `DeniedUserProperties` entirely and just use `ignoredInternalProperties` directly. I saw that in the PR introducing this the goal seemed to be reducing the mem footprint of the properties struct anyways -- https://github.com/cockroachdb/pebble/commit/2f73929887f0b57336ad29bed0e648f68579ad0d.

jbowens (approved):
:lgtm:

> Reviewed 1 of 7 files at r4, 5 of 15 files at r5.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4757#-:-OQu9RkJAWaO2h1iUxM1:bnfp4nl)* status: 4 of 15 files reviewed, 1 unresolved discussion (waiting on @xxmplus)

PR #4758, opened 2025-05-21
sstable/blob: add ValueID indirection

### LINKED ISSUE #4483: blob: use an integer ID to identify values
To tame space amplification within the value separation (#112) implementation, we would like to support the ability to rewrite a blob file in-place. In order to do this, the handles encoded within sstables need to be stable. If the blob file is re-compacted, a reader needs to be able to use the previously-encoded handle to find the value within the new, re-arranged blob file.

We can do this by changing the blob handle so that instead of encoding (referenceID, valueLen, blockNum, offsetInBlock), it encodes (referenceID, valueLen, valueID). The valueID is just a uint. When we write a blob file, we assign IDs sequentially. A block within the blob file can use a colblk encoding, storing the values in a `RawBytes` allowing constant-time lookup of a value by its index.

To accommodate blob file rewrites, each value block will have a base value ID encoded in a custom header and a NullBitmap (eg, https://github.com/cockroachdb/pebble/commit/622ebc17593a7bc35e098b684d57135bef6077db) column indicating which values are present. This will allow a reader to re-map the valueID into an index within the RawBytes column in constant time.

We will also need to adapt the blob index block so that a reader can go from the valueID to a block. One possibility is to add the smallest valueID of a block as a new colblk.Uints column. Readers would need to binary search within the array to find the appropriate block. During scans, this search is constrained to the direction of iteration (and linear scanning might be faster).

We expect the binary search won't be problematic. With the colblk Uints implementation, this binary search is as efficient as binary searching in a simple `[]uint`.

If the binary search does end up being problematic, we could add another small index column that buckets the original value ID space [0, n) into buckets of size _m_. The column would be a `Uints` column with _n_/_m_ rows, and each value would be the index of the block containing the last value in the block. A reader looking for value with ID _i_ would find the relevant bucket _b_ = _i_/_m_, and binary search between the blocks index[_b_-1] and index[_b_]. The bucket size _m_ can be variable, chosen by the writer depending on the count of remaining values. 

Jira issue: PEBBLE-386

Epic CRDB-20379

jbowens:
Introduce indirection between a blob value handle and the physical location of a value within the blob file. Previously, the exact block index and offset of a value within the block was encoded within the value handle embedded in a referencing sstable. This position was used during value retrieval.

This commit adapts the blob value handle to remove the value's offset within the block, replacing it with a BlobValueID (identifying the value as the i'th value in the block). This indirection will allow for a blob file to be rewritten and rearranged without rewriting referencing sstables.

When a blob file is rewritten, an additional column is added to the blob file's index block. This column remaps the original block IDs to a tuple of a) the index of the new physical block containing the original block's extant values and b) the BlockValueID offset of the blocks within the physical block.

This indirection adds some overhead to value retrieval.
```
goos: linux
goarch: amd64
pkg: github.com/cockroachdb/pebble/sstable/blob
                                                  │   old.txt   │               new.txt                │
                                                  │   sec/op    │    sec/op     vs base                │
ValueFetcherRetrieve/uncached/2MiB/sequential-24    349.2n ± 1%    365.6n ± 1%   +4.68% (p=0.000 n=20)
ValueFetcherRetrieve/uncached/2MiB/random-24        805.7n ± 1%    845.2n ± 1%   +4.90% (p=0.000 n=20)
ValueFetcherRetrieve/uncached/16MiB/sequential-24   394.0n ± 1%    408.6n ± 1%   +3.72% (p=0.000 n=20)
ValueFetcherRetrieve/uncached/16MiB/random-24       981.9n ± 1%   1018.5n ± 1%   +3.73% (p=0.000 n=20)
ValueFetcherRetrieve/uncached/64MiB/sequential-24   458.7n ± 1%    473.9n ± 1%   +3.32% (p=0.000 n=20)
ValueFetcherRetrieve/uncached/64MiB/random-24       1.211µ ± 1%    1.253µ ± 0%   +3.51% (p=0.000 n=20)
ValueFetcherRetrieve/cached/2MiB/sequential-24      49.20n ± 1%    49.20n ± 1%        ~ (p=0.774 n=20)
ValueFetcherRetrieve/cached/2MiB/random-24          87.81n ± 2%    87.27n ± 0%        ~ (p=0.616 n=20)
ValueFetcherRetrieve/cached/16MiB/sequential-24     63.05n ± 1%    71.58n ± 1%  +13.54% (p=0.000 n=20)
ValueFetcherRetrieve/cached/16MiB/random-24         127.9n ± 3%    136.5n ± 1%   +6.77% (p=0.000 n=20)
ValueFetcherRetrieve/cached/64MiB/sequential-24     73.95n ± 1%    77.66n ± 1%   +5.02% (p=0.000 n=20)
ValueFetcherRetrieve/cached/64MiB/random-24         191.2n ± 2%    207.1n ± 2%   +8.34% (p=0.000 n=20)
geomean                                             235.9n         247.0n        +4.68%
```

Close #4483.


lassenordahl:
Screenshot 2025-05-22 at 4 59 11 PM

annrpom (commented):
yaaay this is exciting! just had some questions

> Reviewed 1 of 31 files at r1, 31 of 31 files at r2, 30 of 31 files at r3, 31 of 31 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4758#-:-OQyLpcN1dZNBglkzyIs:b7azlfg)* status: all files reviewed, 5 unresolved discussions (waiting on @xinhaoz)

___
*[`sstable/blob/fetcher_test.go` line 240 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4758#-OQtwflp6qtjoKdRdbSW:-OQtwflq6AQo9MtPcZbK:bao6932) ([raw file](https://github.com/cockroachdb/pebble/blob/1f4c44ec3cf6d5c30aea5bebad699bde0eec7868/sstable/blob/fetcher_test.go#L240)):*
> ```Go
> 		})
> 		b.Run("16MiB", func(b *testing.B) {
> 			benchmarkValueFetcherRetrieve(b, 16<<20, 128<<20)
> ```

did you mean to change the cache size of the `16MiB` run from `32<<20`
___
*[`sstable/blob/fetcher.go` line 252 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4758#-OQxkqik8eMROfV8CHXs:-OQxkqil9Eo0PSeuEYv7:b-mnr8m5) ([raw file](https://github.com/cockroachdb/pebble/blob/278257f149e179e14aaa5486b41422b832cb04ee/sstable/blob/fetcher.go#L252)):*
> ```Go
> 	invariants.CheckBounds(int(valueID), cr.currentValueBlock.dec.bd.Rows())
> 	v := cr.currentValueBlock.dec.values.Slice(cr.currentValueBlock.dec.values.Offsets(int(valueID)))
> 	if invariants.Enabled && len(v) != int(vh.ValueLen) {
> ```

ooc: is there a reason why we don't want to do
```
if len(v) != int(vh.ValueLen) {
    return nil, base.AssertionFailedf("value length mismatch: %d != %d", len(v), vh.ValueLen)
}
```

here instead
___
*[`testdata/metrics` line 345 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4758#-OQyIgxJ5VGwRgAAg1Ln:-OQyIgxJ5VGwRgAAg1Lo:b-ko9l5l) ([raw file](https://github.com/cockroachdb/pebble/blob/278257f149e179e14aaa5486b41422b832cb04ee/testdata/metrics#L345)):*
> ```
> level | tables  size val-bl vtables | score  ff  cff |   in  | tables  size | tables  size | tables  size |  read |   r   w
> ------+-----------------------------+----------------+-------+--------------+--------------+--------------+-------+---------
>     0 |     7  5.3KB     0B       0 |    - 0.25 0.25 |  165B |     0     0B |     0     0B |     9  6.8KB |    0B |   1 10.0
> ```

how/why did the write amp change here? do these tests have value sep enabled? (i thought they do not)
___
*[`sstable/blob/blocks.go` line 252 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4758#-OQxbqwz8L5Jut44glhe:-OQxbqwz8L5Jut44glhf:b-c83t9d) ([raw file](https://github.com/cockroachdb/pebble/blob/278257f149e179e14aaa5486b41422b832cb04ee/sstable/blob/blocks.go#L252)):*
> ```Go
> // An indexBlockDecoder decodes blob index blocks.
> type indexBlockDecoder struct {
> 	// virtualBlockCount ...
> ```

^this comment looks in-progress
___
*[`testdata/compaction/value_separation` line 122 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4758#-OQyJSNy5kkzoHWeZU2c:-OQyJSNy5kkzoHWeZU2d:b-3gj058) ([raw file](https://github.com/cockroachdb/pebble/blob/278257f149e179e14aaa5486b41422b832cb04ee/testdata/compaction/value_separation#L122)):*
> ```
> level | tables  size val-bl vtables | score  ff  cff |   in  | tables  size | tables  size | tables  size |  read |   r   w  |    top   in  read
> ------+-----------------------------+----------------+-------+--------------+--------------+--------------+-------+----------+------------------
>     0 |     0     0B     0B       0 |    -    0    0 |   41B |     0     0B |     0     0B |     1   784B |    0B |   0 6.59 |    0B    0B    0B
> ```

cool to see these write amp numbers go down :oooo

RaduBerinde:
Are we never merging multiple blob files during rewrite? It feels like we may end up with many small files if we don't? Or would we incorporate that in the heuristics on when to generate fresh blob files?

RaduBerinde:
> We haven't been planning on it, and we might end up with small files. We've been thinking they wouldn't be much of a problem, because we don't need to seek over them like small sstables. I suppose if we wanted to combine them, we could also add a separate 'virtual blocks' column per file.

If we support that, we will need the metadata for blob references to contain not just a FileNum, but also a "section" inside that file. We could consider extending the metadata and encoding to support that, even if we don't tolerate the section being set to a non-zero value.

In any case, the blob file should somewhere contain the original `FileNum` (or `FileNum`s) and we should cross check that against the original FileNum used when the sstable was written (even if just in invariants mode).

PR #4760, opened 2025-05-22
metamorphic: remote storage fixes for crossversion

RaduBerinde:
These commits fix various aspects around external and shared storage in the context of the crossversion test, where we start with an initial state.

Informs #4732

#### metamorphic: use FS-based remote storage

The metamorphic test uses in-mem remote storage, which doesn't work
when starting with an initial state (as in the crossversion tests).

Note that there is code to save the remote storage contents to disk
when the store is in-memory, but they are not read back when used as
initial state.

This commit changes to using FS-based remote storage in `shared`
and `external` subdirs inside the data dir. Note that the store itself
can still be in-memory; the data gets saved automatically with the
store.

We improve the FS-based Storage implementation to sync data and list
objects. We can now allow simulating crashes in the metamorphic test
when shared storage is enabled.

#### metamorphic: support CreateOnShared on existing store

If the metamorphic test has `CreateOnShared` set to something other
than "none" and we start with an initial store which did not have
remote storage configured, there can be background errors right after
opening the store, before we get a chance to call `SetCreatorID()`.
The metamorphic test fails on these background errors.

To fix this, we always open with `CreateOnSharedNone` first, and if
necessary reopen after setting the creator ID.

#### metamorphic: add random number to external object names

External object names can collide with existing objects when starting
with an initial state. This change adds a random unique number to the
filenames.

### DIFF
--- a/metamorphic/build.go
+++ b/metamorphic/build.go
@@ -258,7 +258,8 @@ func openExternalObj(
 	rangeDelIter keyspan.FragmentIterator,
 	rangeKeyIter keyspan.FragmentIterator,
 ) {
-	objReader, objSize, err := t.externalStorage.ReadObject(context.Background(), externalObjName(externalObjID))
+	objMeta := t.getExternalObj(externalObjID)
+	objReader, objSize, err := t.externalStorage.ReadObject(context.Background(), objMeta.objName)
 	panicIfErr(err)
 	opts := t.opts.MakeReaderOptions()
 	reader, err = sstable.NewReader(
--- a/metamorphic/ops.go
+++ b/metamorphic/ops.go
@@ -7,10 +7,11 @@ package metamorphic
 import (
 	"bytes"
 	"context"
-	"crypto/rand"
+	cryptorand "crypto/rand"
 	"encoding/binary"
 	"fmt"
 	"io"
+	"math/rand/v2"
 	"path"
 	"path/filepath"
 	"slices"
@@ -1053,7 +1054,7 @@ func (o *ingestExternalFilesOp) run(t *Test, h historyRecorder) {
 			meta := t.getExternalObj(obj.externalObjID)
 			external[i] = pebble.ExternalFile{
 				Locator:           "external",
-				ObjName:           externalObjName(obj.externalObjID),
+				ObjName:           meta.objName,
 				Size:              meta.sstMeta.Size,
 				StartKey:          obj.bounds.Start,
 				EndKey:            obj.bounds.End,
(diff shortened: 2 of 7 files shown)


jbowens (approved):
:lgtm:

> Reviewed 3 of 3 files at r1, 1 of 1 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4760#-:-OQtqT416nN70FKCI5gp:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @sumeerbhola)

___
*[`-- commits` line 26 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4760#-OQtm4rUDFtLVsoO91oU:-OQtm4rUDFtLVsoO91oV:bdllo6d):*
I didn't follow this. Is this referring to background errors like from compactions which are attempting to write to shared storage before the creator ID has been set? Would it be simpler if we have `CreateOnShared != none` to first open just the object provider and assign a creator ID to the object provider? Bc in the context of the metamorphic test we don't have the bootstrapping problem to allocate a creator ID
___
*[`metamorphic/ops.go` line 1887 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4760#-OQtoRuYATzYHJOtLn6R:-OQtoRuYATzYHJOtLn6S:betfmee) ([raw file](https://github.com/cockroachdb/pebble/blob/900e5b93a3a5d7343dc23c8153b067bf953ca445/metamorphic/ops.go#L1887)):*
> ```Go
> 	//
> 	// Note that the number is not based on the seed, in case we run using the
> 	// same seed that was used in a previous run with the same store.
> ```

if we made the name based on the generation time seed, wouldn't the current state of the prng not be the same because the previous run's different operations? Or is this referring to previous runs of the same ops? In which case can't we reset the test state to isolate identical re-runs?
___
*[`metamorphic/test.go` line 198 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4760#-OQtldna73z_xM3Utzhf:-OQtldnb5PXavyFIlsVq:boug7bk) ([raw file](https://github.com/cockroachdb/pebble/blob/eb9f20b1b82a650e5dafd1c1f35da2c6f207422f/metamorphic/test.go#L198)):*
> ```Go
> 		})
> 		if err != nil {
> 			fmt.Printf("Open(): %v\n", err)
> ```

nit: debug detritus?
___
*[`metamorphic/test.go` line 207 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4760#-OQtlfc8BRKv3gDgKW9K:-OQtlfc8BRKv3gDgKW9L:boug7bk) ([raw file](https://github.com/cockroachdb/pebble/blob/eb9f20b1b82a650e5dafd1c1f35da2c6f207422f/metamorphic/test.go#L207)):*
> ```Go
> 			})
> 			if err != nil {
> 				fmt.Printf("SetCreatorID(): %v\n", err)
> ```

nit: debug detritus?

RaduBerinde:
Upon backporting, it looks that the first commit breaks the `TestMetaTwoinstance` variant, which makes sense because the "shared" storage is no longer shared. BUT the fact that this test doesn't fail on master means something is broken and it's not actually testing much :/

PR #4761, opened 2025-05-22
metamorphic: sanity check initial-state and previous-ops flags, pass --previous-ops flag

RaduBerinde:
This requires https://github.com/cockroachdb/pebble/pull/4760 and https://github.com/cockroachdb/pebble/pull/4755 (and those need to be backported).

Informs #4732

#### metamorphic: sanity check initial-state and previous-ops flags

Error out if `--initial-state` is used without `--previous-ops`.

#### crossversion: pass --previous-ops flag

Lack of this flag is the root cause for the crossversion tests not
actually testing across versions (instead, each run was separate).

### DIFF
--- a/internal/metamorphic/crossversion/crossversion_test.go
+++ b/internal/metamorphic/crossversion/crossversion_test.go
@@ -174,6 +174,9 @@ func runCrossVersion(
 		// version's metamorphic runs used the same seed, so all of the
 		// resulting histories should be identical.
 		histories := subrunResults.historyPaths()
+		if len(histories) == 0 {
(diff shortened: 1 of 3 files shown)


annrpom (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4761#-:-OQyOdTGBvMiFsIe99A5:b-uwkowd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jbowens and @sumeerbhola)

PR #4764, opened 2025-05-23
metrics: fix w-amp calculation w/ blob files

jbowens:
Fix the w-amp calculation to properly account for blob files. We should be including all bytes written to disk in the numerator, and only bytes "in" (flushed in the case of blob files) in the denominator.

Informs #4581.


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4764)* status: 0 of 3 files reviewed, 2 unresolved discussions

___
*[`metrics.go` line 189 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4764#-OR-7Ihv3i239q8APBOX:-OR-7Ihv3i239q8APBOY:bpa3ijm) ([raw file](https://github.com/cockroachdb/pebble/blob/d50c3fa1a6079182ba52694e84cbd3aa1df99210/metrics.go#L189)):*
> ```Go
> //	TableBytesFlushed + TableBytesCompacted + BlobBytesFlushed + BlobBytesWritten
> //	-----------------------------------------------------------------------------
> //	                      TableBytesIn + BlobBytesFlushed
> ```

For L0, TableBytesIn is defined as "the bytes written to the WAL.". This would include BlobBytesFlushed, no?

I checked the code and it's updated here:
https://github.com/cockroachdb/pebble/blob/88e7636353cb073cff1c89389dc45790ac571a8c/compaction.go#L1570

I think it's fine to keep the definition as is, but we need to divide only by `TableBytesIn` here, and we need to update this line to also add `BlobBytesFlushed` to it: https://github.com/cockroachdb/pebble/blob/88e7636353cb073cff1c89389dc45790ac571a8c/compaction.go#L1567
___
*[`metrics.go` line 194 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4764#-OR-6qbP9j7Fd74mHH60:-OR-6qbP9j7Fd74mHH61:bj40fk2) ([raw file](https://github.com/cockroachdb/pebble/blob/d50c3fa1a6079182ba52694e84cbd3aa1df99210/metrics.go#L194)):*
> ```Go
> 		return 0
> 	}
> 	return float64(m.TableBytesFlushed+m.TableBytesCompacted+m.BlobBytesFlushed+m.BlobBytesWritten) /
> ```

[nit] We should rename `BlobBytesWritten` to `BlobBytesCompacted`, it is confusing

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4764)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`metrics.go` line 189 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4764#-OR-7Ihv3i239q8APBOX:-ORHTpxJEGgRQ1oVkqX9:brct9kj) ([raw file](https://github.com/cockroachdb/pebble/blob/d50c3fa1a6079182ba52694e84cbd3aa1df99210/metrics.go#L189)):*

For L0, TableBytesIn is defined as "the bytes written to the WAL.". This would include BlobBytesFlushed, no?

I checked the code and it's updated here:
https://github.com/cockroachdb/pebble/blob/88e7636353cb073cff1c89389dc45790ac571a8c/compaction.go#L1570

I think it's fine to keep the definition as is, but we need to divide only by `TableBytesIn` here, and we need to update this line to also add `BlobBytesFlushed` to it: https://github.com/cockroachdb/pebble/blob/88e7636353cb073cff1c89389dc45790ac571a8c/compaction.go#L1567
</blockquote></details>

Good catch, done
___
*[`metrics.go` line 194 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4764#-OR-6qbP9j7Fd74mHH60:-ORHRvzm2p8TPm10UOZY:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/d50c3fa1a6079182ba52694e84cbd3aa1df99210/metrics.go#L194)):*

[nit] We should rename `BlobBytesWritten` to `BlobBytesCompacted`, it is confusing
</blockquote></details>

Done.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4764#-:-ORHcWW50Hs0vWSqzyxp:bnfp4nl)* status: 0 of 5 files reviewed, all discussions resolved

PR #4767, opened 2025-05-25
metamorphic: update error blacklist

RaduBerinde:
We see this error when we close the DB without closing all iterators
first (perhaps only if value separation is enabled?). This error
indicates that we did not reproduce the original problem when
reducing.


annrpom (approved):
:lgtm:

RaduBerinde:
TFTR!

PR #4768, opened 2025-05-25
invariants: add buffer mangler, user for blob.ValueFetcher

RaduBerinde:
This moves the buffer mangling logic from the valblk fetcher to a
generic construct and uses it for the blob fetcher as well.


RaduBerinde:
The CI failures are reproducing #4765. I won't merge until that's fixed.

RaduBerinde:
TFTR! CI failure is #4791.

PR #4769, opened 2025-05-27
sstable: fix layout footer printing

xinhaoz:
Fix reading attributes from footer for layout `Describe` method.


annrpom (approved):
:lgtm:

PR #4770, opened 2025-05-27
tool: print normal value in blob mode print

annrpom:
This ensures that we don't print encoded values by default.


RaduBerinde (approved):

annrpom:
TFTR! ('-')7

PR #4772, opened 2025-05-27
db: reformat Metrics.String and print value separation metrics

jbowens:
Reformat the string outputted by Metrics.String to add more organization, and
add the value separation related metrics to the string's output.

Informs https://github.com/cockroachdb/pebble/issues/4581.

### DIFF
--- a/internal/whiteboard/table/examples_test.go
+++ b/internal/whiteboard/table/examples_test.go
@@ -1,0 +1,90 @@
+// Copyright 2025 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package table_test
+
+import (
+	"fmt"
+
+	"github.com/cockroachdb/pebble/internal/whiteboard"
+	"github.com/cockroachdb/pebble/internal/whiteboard/table"
+)
+
+func ExampleDefine() {
+	type Cat struct {
+		Name     string
+		Age      int
+		Cuteness int
+	}
+
+	tbl := table.Define(
+		table.String("name", 7, table.AlignLeft, func(c Cat) string { return c.Name }),
+		table.Div[Cat](),
+		table.Int("age", 4, table.AlignRight, func(c Cat) int { return c.Age }),
+		table.Div[Cat](),
+		table.Int("cuteness", 8, table.AlignRight, func(c Cat) int { return c.Cuteness }),
+	)
+
+	board := whiteboard.Make(8, tbl.CumulativeFieldWidth)
+	fmt.Println("Cool cats:")
+	table.RenderAll(tbl.RenderFunc(board.At(0, 0), table.RenderOptions{}), []Cat{
+		{Name: "Chicken", Age: 5, Cuteness: 10},
+		{Name: "Heart", Age: 4, Cuteness: 10},
+		{Name: "Mai", Age: 2, Cuteness: 10},
+		{Name: "Poi", Age: 15, Cuteness: 10},
+		{Name: "Pigeon", Age: 2, Cuteness: 10},
+		{Name: "Sugar", Age: 8, Cuteness: 10},
+		{Name: "Yaya", Age: 5, Cuteness: 10},
+		{Name: "Yuumi", Age: 5, Cuteness: 10},
+	})
+	fmt.Println(board.String())
+	// Output:
+	// Cool cats:
+	// name    |  age | cuteness
+	// --------+------+---------
+	// Chicken |    5 |       10
+	// Heart   |    4 |       10
+	// Mai     |    2 |       10
+	// Poi     |   15 |       10
+	// Pigeon  |    2 |       10
+	// Sugar   |    8 |       10
+	// Yaya    |    5 |       10
+	// Yuumi   |    5 |       10
+}
+
+func ExampleHorizontally() {
+	type Cat struct {
+		Name     string
+		Age      int
+		Cuteness int
+	}
+
+	tbl := table.Define(
+		table.String("name", 7, table.AlignRight, func(c Cat) string { return c.Name }),
+		table.Div[Cat](),
+		table.Int("age", 4, table.AlignRight, func(c Cat) int { return c.Age }),
+		table.Int("cuteness", 8, table.AlignRight, func(c Cat) int { return c.Cuteness }),
+	)
+
+	board := whiteboard.Make(8, tbl.CumulativeFieldWidth)
+	fmt.Println("Cool cats:")
+	opts := table.RenderOptions{Orientation: table.Horizontally}
+	table.RenderAll(tbl.RenderFunc(board.At(0, 0), opts), []Cat{
+		{Name: "Chicken", Age: 5, Cuteness: 10},
+		{Name: "Heart", Age: 4, Cuteness: 10},
+		{Name: "Mai", Age: 2, Cuteness: 10},
+		{Name: "Poi", Age: 15, Cuteness: 10},
+		{Name: "Pigeon", Age: 2, Cuteness: 10},
+		{Name: "Sugar", Age: 8, Cuteness: 10},
+		{Name: "Yaya", Age: 5, Cuteness: 10},
+		{Name: "Yuumi", Age: 5, Cuteness: 10},
+	})
+	fmt.Println(board.String())
+	// Output:
+	// Cool cats:
+	// name     |  Chicken   Heart     Mai     Poi  Pigeon   Sugar    Yaya   Yuumi
+	// ---------+-----------------------------------------------------------------
+	// age      |        5       4       2      15       2       8       5       5
+	// cuteness |       10      10      10      10      10      10      10      10
+}
--- a/internal/whiteboard/table/table.go
+++ b/internal/whiteboard/table/table.go
@@ -1,0 +1,356 @@
+// Copyright 2025 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package table
+
+import (
+	"fmt"
+	"math"
+	"strconv"
+	"strings"
+
+	"github.com/cockroachdb/pebble/internal/humanize"
+	"github.com/cockroachdb/pebble/internal/whiteboard"
+)
+
+// Define defines a new table layout with the given fields.
+func Define[T any](fields ...Field[T]) Definition[T] {
+	var verticalHeader strings.Builder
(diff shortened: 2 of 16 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4772)* status: 0 of 16 files reviewed, 11 unresolved discussions

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORX4MKk47Rn7gJBaqhq:-ORX4MKk47Rn7gJBaqhr:b-uyl190):*
[nit] maybe a better name would be `internal/ascii` with `ascii.Board`
___
*[`internal/whiteboard/table/examples_test.go` line 23 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORYvnGX53xTEa-5uXCK:-ORYvnGX53xTEa-5uXCL:b-eosglk) ([raw file](https://github.com/cockroachdb/pebble/blob/58320ecd6610115b5aedce1c7f0ff03a351f01fa/internal/whiteboard/table/examples_test.go#L23)):*
> ```Go
> 	tbl := table.Define(
> 		table.String("name", 7, table.AlignLeft, func(c Cat) string { return c.Name }),
> 		table.Div[Cat](),
> ```

[nit] Having to say `[Cat]` here is unfortunate. I'd create a "base" `Element` interface that doesn't have a `renderValue` function, and two more specific interfaces `Field` and (e.g.) `StaticElement`; the latter has a renderStatic() func that doesn't depend on `T`. When we render, we cast each element to either a `Field[T]` or `StaticElement`.
___
*[`internal/whiteboard/table/examples_test.go` line 81 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORYuD7m4yTVwO0ur2xX:-ORYuD7m4yTVwO0ur2xY:b7274g2) ([raw file](https://github.com/cockroachdb/pebble/blob/58320ecd6610115b5aedce1c7f0ff03a351f01fa/internal/whiteboard/table/examples_test.go#L81)):*
> ```Go
> 		{Name: "Sugar", Age: 8, Cuteness: 10},
> 		{Name: "Yaya", Age: 5, Cuteness: 10},
> 		{Name: "Yuumi", Age: 5, Cuteness: 10},
> ```

[nit] Can we add a row that has a value that is wider than the field width?
___
*[`testdata/event_listener` line 261 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORYxf_p-1xP9BjN7Tv8:-ORYxf_p-1xP9BjN7Tv9:bahlevr) ([raw file](https://github.com/cockroachdb/pebble/blob/a50858418347c307f268ed11a6e02ae39e1dfb26/testdata/event_listener#L261)):*
> ```
> count     |         1        0        0        0        0        0        0        0        0
> --------------------------------------------------------------------------------------------------
> COMMIT PIPELINE           ITERATORS             TABLES                  CGO MEMORY        COMPACTIONS
> ```

This layout is much harder to read than before, it's not intuitive that we have to read down, and that there's no relation between different columns of the same row
___
*[`testdata/event_listener` line 361 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORYxKit2PuFsVdqbZSd:-ORYxKiu5vPPMGhJjjvA:b-eoy8v0) ([raw file](https://github.com/cockroachdb/pebble/blob/a50858418347c307f268ed11a6e02ae39e1dfb26/testdata/event_listener#L361)):*
> ```
>     6 |   1.5KB |      2 1.5KB |      0     0 |     0B     0B |  1.5KB |      1  750B |   1  0.50
> total |   4.4KB |      6 4.4KB |      0     0 |     0B     0B |  2.3KB |      3 2.2KB |   5  2.57
> -------------------------------------------------------------------------------------------------
> ```

[nit] These lines of dashes after tables are more clutter than anything; a blank line would make it easier on the eyes
___
*[`internal/whiteboard/table/table.go` line 17 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORYsNeo7PDn6f_r49cH:-ORYsNeo7PDn6f_r49cI:b1z004i) ([raw file](https://github.com/cockroachdb/pebble/blob/58320ecd6610115b5aedce1c7f0ff03a351f01fa/internal/whiteboard/table/table.go#L17)):*
> ```Go
> )
> 
> // Define defines a new table layout with the given fields.
> ```

[nit] Would be good to show a full example here. Ideally, the reader won't have to go further down to understand how they can use it, at least for the most basic cases.
___
*[`internal/whiteboard/table/table.go` line 60 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORYsXNHEuDY03rLqyXj:-ORYsXNHEuDY03rLqyXk:bwf57pm) ([raw file](https://github.com/cockroachdb/pebble/blob/58320ecd6610115b5aedce1c7f0ff03a351f01fa/internal/whiteboard/table/table.go#L60)):*
> ```Go
> 
> // A Definition defines the layout of a table.
> type Definition[T any] struct {
> ```

[nit] Layout seems like a better name
___
*[`internal/whiteboard/table/table.go` line 73 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORYrijI1qouHVuwfiTs:-ORYrijI1qouHVuwfiTt:b-nw6gdf) ([raw file](https://github.com/cockroachdb/pebble/blob/58320ecd6610115b5aedce1c7f0ff03a351f01fa/internal/whiteboard/table/table.go#L73)):*
> ```Go
> }
> 
> // RenderAll renders all the rows of the table, calling fn for each row.
> ```

I don't get what's the benefit of having `RenderAll` and `RenderFunc`. Why not just a `Definition.Render()` which takes `[]T`( or `iter.Seq[T]`)?
___
*[`internal/whiteboard/table/table.go` line 138 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORYshJWCQikc3ncyO4G:-ORYshJWCQikc3ncyO4H:b-8ivwbp) ([raw file](https://github.com/cockroachdb/pebble/blob/58320ecd6610115b5aedce1c7f0ff03a351f01fa/internal/whiteboard/table/table.go#L138)):*
> ```Go
> 
> // A RenderContext provides the context for rendering a table.
> type RenderContext[T any] struct {
> ```

It feels like we have a lot of `T` parameters. Do we really need the entire Definition here, or could we store just the MaxFieldWidth?
___
*[`internal/whiteboard/table/table.go` line 237 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORYt9UtDcNtejUJ6n_V:-ORYt9UtDcNtejUJ6n_W:b-sbrxku) ([raw file](https://github.com/cockroachdb/pebble/blob/58320ecd6610115b5aedce1c7f0ff03a351f01fa/internal/whiteboard/table/table.go#L237)):*
> ```Go
> )
> 
> type Orientation uint8
> ```

[nit] Could use a comment
___
*[`internal/whiteboard/table/table.go` line 267 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORYtK5dEK8aNsGPir8u:-ORYtK5dEK8aNsGPir8v:bq7k6hr) ([raw file](https://github.com/cockroachdb/pebble/blob/58320ecd6610115b5aedce1c7f0ff03a351f01fa/internal/whiteboard/table/table.go#L267)):*
> ```Go
> }
> 
> func Count[T any](header string, width int, align Align, fn func(r T) uint64) Field[T] {
> ```

We can use a `N numeric` parameter so we don't have to have int64 and uint64 variants, and use `crhumanize` functions.

RaduBerinde:
*[`testdata/event_listener` line 262 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORajXzaA7lpcfx7-2dt:-ORajXzaA7lpcfx7-2du:b-dozznw) ([raw file](https://github.com/cockroachdb/pebble/blob/a50858418347c307f268ed11a6e02ae39e1dfb26/testdata/event_listener#L262)):*
> ```
> --------------------------------------------------------------------------------------------------
> COMMIT PIPELINE           ITERATORS             TABLES                  CGO MEMORY        COMPACTIONS
> wals                      block cache           stats initial load        0B tot          2.2KB estimated debt
> ```

These compaction stats (estimated debt, in-progress, etc.) could be another one-row table after the count table. Maybe more of these could be one or two row tables.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4772)* status: 0 of 16 files reviewed, 15 unresolved discussions (waiting on @jbowens)

___
*[`testdata/event_listener` line 350 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORakw-GCdqIDo6AWDqG:-ORakw-GCdqIDo6AWDqH:b-xr8v2b) ([raw file](https://github.com/cockroachdb/pebble/blob/a50858418347c307f268ed11a6e02ae39e1dfb26/testdata/event_listener#L350)):*
> ```
> metrics
> ----
> LSM   |                        |    virtual   |   value sep   |        |   ingested   |    amp
> ```

[nit] "virtual" is too generic, I'd keep "vtables  size"
___
*[`testdata/event_listener` line 351 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORal7e2AJZGqpUDs2-3:-ORal7e2AJZGqpUDs2-4:b-yttnvj) ([raw file](https://github.com/cockroachdb/pebble/blob/a50858418347c307f268ed11a6e02ae39e1dfb26/testdata/event_listener#L351)):*
> ```
> ----
> LSM   |                        |    virtual   |   value sep   |        |   ingested   |    amp
> level | aggsize | tables  size |  count  size |  refsz valblk |     in | tables  size |   r     w
> ```

[nit] "aggsize" is a bit confusing, not obvious what we're "aggregating". Maybe simply "size" and remove the separator so it's together with the level number?  Btw this is always equal to tables size + refsz?

RaduBerinde:
*[`testdata/event_listener` line 261 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4772#-ORYxf_p-1xP9BjN7Tv8:-ORaqw6z8MxbEsejC42S:b-bsduid) ([raw file](https://github.com/cockroachdb/pebble/blob/a50858418347c307f268ed11a6e02ae39e1dfb26/testdata/event_listener#L261)):*

```
It's the column-oriented format    Anyway, maybe Sumeer can
that I find hard to read. Can't    chime in too since this is
we achieve the same organization   clearly subjective.
improvement but keep it row
oriented? Or at least add some
vertical separators.
```
</blockquote></details>

Also, not super important but something to consider:  we may want to copy/paste a relevant section in an escalation, the column-oriented format makes that harder

PR #4774, opened 2025-05-27
tool: ensure properly coupled blob references for find

annrpom:
### tool: add more data to db val sep directory
This adds an additional blob file in our `testdata/find-db-val-sep` directory
for testing.

---

### tool: ensure properly coupled blob references for find
Previously, when we created our TableBlobContext for the `find` tool,
we loaded all blob references from each manifest. Instead, we
should have been loading the blob references for each sst separately. Loading
all blob references was incorrect since a val-sep key's inline handle would
potentially point to the wrong blob file when referencing a value.


tool/find.go
```diff
@@ -286,23 +285,11 @@ func (f *findT) readManifests(stdout io.Writer) {
 					if _, ok := f.tableMeta[nf.Meta.TableNum]; !ok {
 						f.tableMeta[nf.Meta.TableNum] = nf.Meta
 					}
-				}
-				for _, bf := range ve.NewBlobFiles {
-					if _, ok := blobMetas[bf.FileNum]; !ok {
-						blobMetas[bf.FileNum] = struct{}{}
-					}
+					f.blobRefsMap[nf.Meta.TableNum] = &nf.Meta.BlobReferences
```
annrpom, line 288: the issue that exists in `findAndReadManifests` also existed in `readManifests` -- i think we can do something like this for a quick fix (and we can do the subcommand suggestion later), but this works under the assumption that manifests will be read in order from least recent to most recent version edits (i think) https://github.com/cockroachdb/pebble/blob/d742eb87709bc5e19f29a467f4df174ae6fbe5d1/tool/find.go#L220

cc: @jbowens @RaduBerinde what do u all think?

RaduBerinde: This looks good to me

RaduBerinde (approved):
:lgtm: Thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4774#-:-OROFkbhA9G0S8vZhMbL:b-g2a5fb)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`tool/testdata/find` line 200 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4774#-OROFGnh5RY5vMe2VhJu:-OROFGnh5RY5vMe2VhJv:b-32hjgn) ([raw file](https://github.com/cockroachdb/pebble/blob/03439ea7cf168ee3f880f65c240f4686ef5e5c91/tool/testdata/find#L200)):*
> ```
> testdata/find-db-val-sep
> crdb:eee
> --blob-mode=load
> ```

Not for this PR, but what do you think about having "print" be the default mode, and changing the flag to something like `--blob-load-path <dbdir>` so that the db dir is just for that flag?

PR #4773, opened 2025-05-27
sstable: use ReadPropertiesBlock when accessing reader props

xinhaoz:
### sstable: address TODO to use IsZero

----------------------
### sstable: add AttributePointKeys
AttributePointKeys is set if Properties.NumEntries > 0

----------------------
### sstable: use ReadPropertiesBlock when accessing reader props
Load reader properties from block instead of reading from the in-memory
structure within the sstable pkg.

Part of: #4383

PR #4375, opened 2025-03-12
sstable: remove VirtualReader and CommonReader

EdwardX29:
Refactor VirtualReaderParams into block.ReadEnv to get rid of VirtualReader and CommonReader.

PR #4775, opened 2025-05-27
pebble: use ReadPropertiesBlock when accessing reader props

xinhaoz:
### pebble: close file cache before closing the objstorage provider

It seems that during close, the file cache may attempt to access structures
managed by the objstroage provider during the eviction process, such as sending
to channels used for tracing events. We should close the file cache before the
obj storage provider.

----------------------------------

### pebble: use ReadPropertiesBlock when accessing reader props

Remove remaining reads of `Properites` on the reader struct. This commit also
removes the Properties field from the reader.

Closes: https://github.com/cockroachdb/pebble/issues/4383

----------
Previous: https://github.com/cockroachdb/pebble/pull/4773

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -1349,11 +1349,15 @@ func runSSTablePropertiesCmd(t *testing.T, td *datadriven.TestData, d *DB) strin
 	}
 	defer r.Close()
 
-	props := r.Properties.String()
+	loadedProps, err := r.ReadPropertiesBlock(context.Background(), nil /* buffer pool */)
+	if err != nil {
+		return err.Error()
+	}
+	props := loadedProps.String()
 	env := sstable.ReadEnv{}
 	if m != nil && m.Virtual {
 		env.Virtual = m.VirtualParams
-		scaledProps := r.Properties.GetScaledProperties(m.TableBacking.Size, m.Size)
+		scaledProps := loadedProps.GetScaledProperties(m.TableBacking.Size, m.Size)
 		props = scaledProps.String()
 	}
 	if len(td.Input) == 0 {
--- a/db.go
+++ b/db.go
@@ -1789,6 +1789,8 @@ func (d *DB) Close() error {
 		err = firstError(err, errors.Errorf("non-zero zombie blob count: %d", zblobs))
 	}
 
+	err = firstError(err, d.fileCache.Close())
+
 	err = firstError(err, d.objProvider.Close())
 
 	// If the options include a closer to 'close' the filesystem, close it.
@@ -1800,8 +1802,6 @@ func (d *DB) Close() error {
 	if v := d.mu.snapshots.count(); v > 0 {
 		err = firstError(err, errors.Errorf("leaked snapshots: %d open snapshots on DB %p", v, d))
 	}
-
-	err = firstError(err, d.fileCache.Close())
 
 	return err
 }
--- a/file_cache.go
+++ b/file_cache.go
@@ -946,7 +946,11 @@ func (h *fileCacheHandle) getTableProperties(file *tableMetadata) (*sstable.Prop
 	defer v.Unref()
 
 	r := v.Value().mustSSTableReader()
-	return &r.Properties, nil
+	props, err := r.ReadPropertiesBlock(context.TODO(), nil /* buffer pool */)
+	if err != nil {
+		return nil, err
+	}
+	return &props, nil
 }
 
 type fileCacheValue struct {
--- a/ingest.go
+++ b/ingest.go
@@ -281,16 +281,24 @@ func ingestLoad1(
 			tf, fmv, fmv.MinTableFormat(), fmv.MaxTableFormat(),
 		)
 	}
-	if tf.BlockColumnar() {
-		if _, ok := opts.KeySchemas[r.Properties.KeySchemaName]; !ok {
-			return nil, keyspan.Span{}, errors.Newf(
-				"pebble: table uses key schema %q unknown to the database",
-				r.Properties.KeySchemaName)
-		}
-	}
-	if r.Properties.NumValuesInBlobFiles > 0 {
+
+	if r.Attributes.Has(sstable.AttributeBlobValues) {
 		return nil, keyspan.Span{}, errors.Newf(
 			"pebble: ingesting tables with blob references is not supported")
+	}
+
+	props, err := r.ReadPropertiesBlock(ctx, nil /* buffer pool */)
+	if err != nil {
+		return nil, keyspan.Span{}, err
+	}
+
+	// If this is a columnar block, read key schema name from properties block.
+	if tf.BlockColumnar() {
+		if _, ok := opts.KeySchemas[props.KeySchemaName]; !ok {
+			return nil, keyspan.Span{}, errors.Newf(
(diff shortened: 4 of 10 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4775)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 323 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4775#-ORMv7TV1c4gQaqUyVeJ:-ORMv7TV1c4gQaqUyVeK:b-am6mqj) ([raw file](https://github.com/cockroachdb/pebble/blob/25b668583ba7d9a6df35a7bfc97fd6198c821555/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L323)):*
> ```Go
> func (g *eventGenerator) flush() {
> 	if g.buf.num > 0 && !g.t.closed.Load() {
> 		g.t.workerDataCh <- g.buf
> ```

In principle, there is still a race here if the goroutine gets preempted right before sending on the channel.

Do you have the stack trace with the crash when we flushed after `Close()`? It suggests we were doing object IO after closing the objstorage provider, which shouldn't happen.

xinhaoz (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4775)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 323 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4775#-ORMv7TV1c4gQaqUyVeJ:-ORNKm-FDxZEXYj63M5y:btlh1pg) ([raw file](https://github.com/cockroachdb/pebble/blob/25b668583ba7d9a6df35a7bfc97fd6198c821555/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L323)):*

In principle, there is still a race here if the goroutine gets preempted right before sending on the channel.

Do you have the stack trace with the crash when we flushed after `Close()`? It suggests we were doing object IO after closing the objstorage provider, which shouldn't happen.
</blockquote></details>

Ahh yes you're right. Here's the stack trace, I also just dropped the commit since it's not actually fixing what we want: https://github.com/cockroachdb/pebble/actions/runs/15286733028/job/42998483391?pr=4775

It seems that the objprovider is closed before the file cache -- should we just close the file cache first or must it be closed last here? https://github.com/cockroachdb/pebble/blob/2fd9454bed4c2d0eea216dbfe5ae4e8b5629550a/db.go#L1794

xinhaoz:
*[`objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go` line 323 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4775#-ORMv7TV1c4gQaqUyVeJ:-ORNdvkn8uoE-PWUrg1Y:b-ybm7wh) ([raw file](https://github.com/cockroachdb/pebble/blob/25b668583ba7d9a6df35a7bfc97fd6198c821555/objstorage/objstorageprovider/objiotracing/obj_io_tracing_on.go#L323)):*

Ahh yes you're right. Here's the stack trace, I also just dropped the commit since it's not actually fixing what we want: https://github.com/cockroachdb/pebble/actions/runs/15286733028/job/42998483391?pr=4775

It seems that the objprovider is closed before the file cache -- should we just close the file cache first or must it be closed last here? https://github.com/cockroachdb/pebble/blob/2fd9454bed4c2d0eea216dbfe5ae4e8b5629550a/db.go#L1794
</blockquote></details>

Should also be failing on CI now. Also I was unsure if this is an even an issue with objprovider tracing at all, but I couldn't see anything obvious with these changes since we didn't add any reader.close calls here. LMK if anyone sees anything odd  with the file cache interactions here.

jbowens (approved):
:lgtm:

> Reviewed 11 of 11 files at r3, 9 of 10 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4775#-:-ORSv95W74egaPjecj3d:bnfp4nl)* status: 10 of 11 files reviewed, 5 unresolved discussions (waiting on @RaduBerinde)

___
*[`-- commits` line 11 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4775#-ORSlqOn0zyqkt6uBnW-:-ORSlqOn0zyqkt6uBnW0:bjr29ev):*
nit: `Properties`
___
*[`table_stats.go` line 512 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4775#-ORSuT4y9ZYd2gB3tfGJ:-ORSuPuQ7Brg1IdeuU2A:b-tqgswh) ([raw file](https://github.com/cockroachdb/pebble/blob/e84181c2a7e6b0baaecec7ad3918935b5352294a/table_stats.go#L512)):*
> ```Go
> 
> 	addPhysicalTableStats := func(r *sstable.Reader, _ sstable.ReadEnv) (err error) {
> 		props, err := r.ReadPropertiesBlock(context.TODO(), nil /* buffer pool */)
> ```

This block read is good for now, but I think it's probably worth avoiding in a follow-up by adding `NumEntries`, `RawKeySize` and `RawValueSize` to `TableMetadata.TableStats`. Want to leave a TODO/ file an issue?
___
*[`table_stats.go` line 518 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4775#-ORSsqeIAa6rlhKz2DaR:-ORSsqeIAa6rlhKz2DaS:btef2ky) ([raw file](https://github.com/cockroachdb/pebble/blob/9c19c94c4f8dd3fddf6d825dc35c69ac8191afec/table_stats.go#L518)):*
> ```Go
> 		entryCount += file.Stats.NumEntries
> 		keySum += v.Properties.RawKeySize
> 		valSum += v.Properties.RawValueSize
> ```

I think the existing code here has a bug. It should be scaling the properties for virtual tables, like we do higher up:

```
			props := r.Properties.CommonProperties
			if meta.Virtual {
				props = r.Properties.GetScaledProperties(meta.TableBacking.Size, meta.Size)
			}
```

I filed https://github.com/cockroachdb/pebble/issues/4790 for this. You could fix it up in this PR, or leave it for now. Up to you. I think we can also combine `addPhysicalTableStats` and `addVirtualTableStats` into one now. I think we overlooked that in #4375 
___
*[`ingest.go` line 297 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4775#-ORSoF_M3NExFekyXKgR:-ORSoF_M3NExFekyXKgS:b-47em3h) ([raw file](https://github.com/cockroachdb/pebble/blob/e84181c2a7e6b0baaecec7ad3918935b5352294a/ingest.go#L297)):*
> ```Go
> 
> 	containsRangeDeletions := r.Attributes.Intersects(sstable.AttributeRangeDels | sstable.AttributeRangeKeyDels)
> 	if tf.BlockColumnar() || !containsRangeDeletions {
> ```

nit: I think we should make the loading of the properties block unconditional here, because going forward we expect almost all tables to be `BlockColumnar()` and it'll simplify the code a bit

PR #4776, opened 2025-05-27
metamorphic: fix shared and external directories, fix testing of multiple instances

RaduBerinde:
#### metamorphic: check for false values when parsing options

The test parsing code assumes that the presence of a flag means that
it is turned on. One of the standard options uses `=false` which is
actually interpreted as true.

This commit adds an assertion and fixes the problematic option.

#### metamorphic: tweak secondary cache settings

We don't use the secondary cache at all, so remove it from the
standard settings and reduce the probability of it being turned on
when randomizing options.

#### metamorphic: sometimes turn off value separation in randomizeOptions

We also remove support for `disable_value_separation` as we can just
set the Pebble setting directly.

#### metamorphic: fix shared and external directories

When multiple stores are used, the shared and external directories
were set up inside the data dir of each store, which is not correct.
This commit fixes the paths.

### DIFF
--- a/metamorphic/ops.go
+++ b/metamorphic/ops.go
@@ -2134,6 +2134,10 @@ func (r *replicateOp) run(t *Test, h historyRecorder) {
 	// Shared replication only works if shared storage is enabled and value
 	// separation is disabled.
 	useSharedIngest := t.testOpts.useSharedReplicate && t.testOpts.sharedStorageEnabled &&
+		// TODO(radu): allow shared ingest with value separation. We automatically
+		// disable value separation when the output is on shared storage, so it
+		// should work. We curently get a crash in ScanInternal in this combination
+		// which we need to root cause.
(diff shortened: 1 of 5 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4776)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @annrpom and @jbowens)

___
*[`metamorphic/options_test.go` line 162 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4776#-ORIjsDXAONsJeUiOwf7:-ORIllY-4Q1K0-TfxsvS:b-g7ecop) ([raw file](https://github.com/cockroachdb/pebble/blob/03674f8dbc6bf3308c11d3ab8d0fd9deba42be04/metamorphic/options_test.go#L162)):*

what is `defaultValue` referring to?
</blockquote></details>

Oops, fixed. TFTR!

PR #4779, opened 2025-05-28
db: improve file cache panic

RaduBerinde:
Improve the panic message when we try to evict a file that is still in
use.


annrpom (approved):
nice!!

RaduBerinde:
TFTR!

PR #4780, opened 2025-05-28
db: fix copy compaction logic for tables with blobs

RaduBerinde:
#### db: set up the blob value fetcher for internal scans

We also add a safeguard to the code to fall back on the error fetcher
when it is not set, and we update the error fetcher to panic in
invariant builds.

#### db: fix copy compaction logic for tables with blobs

The condition checking for blob references was incorrect; it caused
move compactions that shouldn't be allowed. We fix the logic and
refactor the code so it's more readable.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -415,48 +415,7 @@ func newCompaction(
 	c.kind = pc.kind
 
 	preferSharedStorage := remote.ShouldCreateShared(opts.Experimental.CreateOnShared, c.outputLevel.level)
-	// In addition to the default compaction, we also check whether a tombstone density compaction can be optimized into
-	// a move compaction. However, we want to avoid performing a move compaction into the lowest level, since the goal
-	// there is to actually remove the tombstones.
-	//
-	// Tombstone density compaction is meant to address cases where tombstones don't reclaim much space but are still
-	// expensive to scan over. We can only remove the tombstones once there's nothing at all underneath them.
-	if (c.kind == compactionKindDefault || (c.kind == compactionKindTombstoneDensity && c.outputLevel.level != numLevels-1)) &&
-		c.outputLevel.files.Empty() && !c.hasExtraLevelData() &&
-		c.startLevel.files.Len() == 1 && c.grandparents.AggregateSizeSum() <= c.maxOverlapBytes {
-		// This compaction can be converted into a move or copy from one level
-		// to the next. We avoid such a move if there is lots of overlapping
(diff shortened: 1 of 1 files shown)


jbowens (approved):
:lgtm:

> Reviewed 5 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4780#-:-ORNtE6P60vKxyjOI-fJ:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @annrpom)

___
*[`compaction.go` line 481 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4780#-ORNs5eQ84oyW68tLlKP:-ORNs5eQ84oyW68tLlKQ:bm4ekpx) ([raw file](https://github.com/cockroachdb/pebble/blob/161aaaad7497c61f23759175800134b845d39a13/compaction.go#L481)):*
> ```Go
> 	// Shared and external tables can always be moved. We can also move a local
> 	// table unless we need the result to be on shared storage.
> 	if isRemote || !preferSharedStorage {
> ```

if `isRemote=true` and `preferSharedStorage=false`, wouldn't we want to rewrite the table in an ordinary compaction? maybe this isn't possible unless you previously had shared storage enabled and then disable it?

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4780#-:-OROAn5S-lDzutEq1zoJ:b-814g8k)* status: all files reviewed, 1 unresolved discussion (waiting on @annrpom and @jbowens)

___
*[`compaction.go` line 481 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4780#-ORNs5eQ84oyW68tLlKP:-OROApHx6VDcaoaXBeMp:bcfpazk) ([raw file](https://github.com/cockroachdb/pebble/blob/161aaaad7497c61f23759175800134b845d39a13/compaction.go#L481)):*

if `isRemote=true` and `preferSharedStorage=false`, wouldn't we want to rewrite the table in an ordinary compaction? maybe this isn't possible unless you previously had shared storage enabled and then disable it?
</blockquote></details>

It is possible to get an ingestion of a shared table that goes to L0 (at least in the metamorphic test - probably not with CRDB as we'd always excise). I think since the file is shared already, there's no problem keeping it shared until it goes further down in the LSM.

PR #4783, opened 2025-05-28
tool: ensure properly coupled blob references for scan

annrpom:


annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4783)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`tool/sstable.go` line 172 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4783#-ORWpnC-DIxx87ID5Rpk:-ORX_i-P2E-7qpdkZ626:b-20fwe3) ([raw file](https://github.com/cockroachdb/pebble/blob/0ff2c4f0c32cba8ef408a18b0d07dd016163f307/tool/sstable.go#L172)):*

If we fail to parse the filename, could we leave the `FileNum` unset? Then down below where we construct the `TableBlobContext`  we could error out only if using `BlobRefModeLoad`?
</blockquote></details>

yeah we could -- i was mainly trying to ensure that the following comment stayed true even in these tests, but im unsure how strict one must be about obeying it:
https://github.com/cockroachdb/pebble/blob/9360e681647e7a79210dbc16380e95fe2c7f92b0/internal/sstableinternal/options.go#L14-L19

is the motivation behind the delayed error suggestion so that we don't have to rename a bunch of ssts?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4783)* status: all files reviewed, 1 unresolved discussion (waiting on @annrpom)

___
*[`tool/sstable.go` line 172 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4783#-ORWpnC-DIxx87ID5Rpk:-ORXe37wCG3tRbypdOoU:b-7oti71) ([raw file](https://github.com/cockroachdb/pebble/blob/0ff2c4f0c32cba8ef408a18b0d07dd016163f307/tool/sstable.go#L172)):*

yeah we could -- i was mainly trying to ensure that the following comment stayed true even in these tests, but im unsure how strict one must be about obeying it:
https://github.com/cockroachdb/pebble/blob/9360e681647e7a79210dbc16380e95fe2c7f92b0/internal/sstableinternal/options.go#L14-L19

is the motivation behind the delayed error suggestion so that we don't have to rename a bunch of ssts?
</blockquote></details>

ah, I see--I think renaming the test SSTs makes sense and will allow us to update those sstables with blob references if we need to or decide to. The motivation is to more generally allow reading ssts without numbered file numbers which I'm now remembering can exist when preparing an sstable for ingestion. For example CockroachDB will write a file with a different file naming scheme to an auxiliary directory, pass it into `Ingest` and _then_ pebble will link it into the datadirectory under a numbered file name: https://github.com/cockroachdb/cockroach/blob/067c0afd6e2dfb604754a3498958e4936cbb33dd/pkg/kv/kvserver/logstore/sideload_disk.go#L42-L56

Good point about the cache options. It requires the FileNum because that's used as in the key in the cache, and we need it to avoid populating the cache with one file's blocks and then accidentally reading those blocks from the cache when reading a different file. I think the only reason this works today is because `foreachSSTable` creates a new cache handle for every unique sstable:
```
		c := pebble.NewCache(128 << 20 /* 128 MB */)
		defer c.Unref()
		ch := c.NewHandle()
		defer ch.Close()
```
so we're isolating each sstable's blocks. I'm not really clear on why the existing code is using a Cache at all. It seems like it would be preferable to use a `BufferPool`. maybe we could leave a TODO or follow up on trying that?

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4783)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`tool/sstable.go` line 172 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4783#-ORWpnC-DIxx87ID5Rpk:-ORYydZ4DdYwagnEeV_B:b-t72fwq) ([raw file](https://github.com/cockroachdb/pebble/blob/0ff2c4f0c32cba8ef408a18b0d07dd016163f307/tool/sstable.go#L172)):*

ah, I see--I think renaming the test SSTs makes sense and will allow us to update those sstables with blob references if we need to or decide to. The motivation is to more generally allow reading ssts without numbered file numbers which I'm now remembering can exist when preparing an sstable for ingestion. For example CockroachDB will write a file with a different file naming scheme to an auxiliary directory, pass it into `Ingest` and _then_ pebble will link it into the datadirectory under a numbered file name: https://github.com/cockroachdb/cockroach/blob/067c0afd6e2dfb604754a3498958e4936cbb33dd/pkg/kv/kvserver/logstore/sideload_disk.go#L42-L56

Good point about the cache options. It requires the FileNum because that's used as in the key in the cache, and we need it to avoid populating the cache with one file's blocks and then accidentally reading those blocks from the cache when reading a different file. I think the only reason this works today is because `foreachSSTable` creates a new cache handle for every unique sstable:
```
		c := pebble.NewCache(128 << 20 /* 128 MB */)
		defer c.Unref()
		ch := c.NewHandle()
		defer ch.Close()
```
so we're isolating each sstable's blocks. I'm not really clear on why the existing code is using a Cache at all. It seems like it would be preferable to use a `BufferPool`. maybe we could leave a TODO or follow up on trying that?
</blockquote></details>

ahhhh thank you for the context -- for now, i applied the unset FileNum suggestion and left a TODO for the `BufferPool`

annrpom:
(TFTR! ('-')7)

jayshrivastava:
This is epic

PR #4794, opened 2025-05-30
compressionanalyzer: sample per byte instead of per block

RaduBerinde:
We record compression ratio and block compression/decompression time
per block. This is ok when blocks are about the same size but are less
useful when there is larger variance (as in >128KB buckets).

We want the compression ratio to be the ratio between the total
decompressed size and the total compressed size, not the average
compression ratio among blocks. Similarly, we want to measure
performance in terms of CPU time per byte.

We switch to using frequency-weighted sampling (where frequency is the
size of the block) and we now show compression/decompression
performance in MB/s. Note that the standard deviation for the latter
is derived from the CPU time per second metric, it doesn't apply
directly to the MB/s figure.

### DIFF
--- a/sstable/compressionanalyzer/block_analyzer.go
+++ b/sstable/compressionanalyzer/block_analyzer.go
@@ -88,10 +88,10 @@ func (a *BlockAnalyzer) runExperiment(
 	}
 	decompressionTime := t2.Elapsed()
 
-	// CPU times are in microseconds.
-	pa.CompressionTime.Add(compressionTime.Seconds() * 1e6)
-	pa.DecompressionTime.Add(decompressionTime.Seconds() * 1e6)
-	pa.CompressionRatio.Add(float64(len(block)) / float64(len(compressed)))
+	// CPU times are in nanoseconds / byte.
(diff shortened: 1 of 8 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4794#-:-ORl6-Zw20TzVYmb_lnk:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`sstable/compressionanalyzer/buckets.go` line 200 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4794#-ORl5ZPxDul2Yl0quQoc:-ORl5ZPxDul2Yl0quQod:bdx4d1) ([raw file](https://github.com/cockroachdb/pebble/blob/1a8295ecf1e971ffce199bb9caca11bd262be0bc/sstable/compressionanalyzer/buckets.go#L200)):*
> ```Go
> 		return 0
> 	}
> 	return float64(time.Second) / (nsPerByte * (2 << 20))
> ```

why `2 << 20`, shouldn't it be `1 << 20`?

PR #4800, opened 2025-06-01
db: clean up level options

RaduBerinde:
#### base: add NoFilterPolicy

This is cleaner than special-casing `nil` and will allow
differentiating between an unset value and a value that is
specifically set to `NoFilterPolicy`.

#### db: clean up level options

The `Options.Levels` field has questionable semantics. It's a slice of
arbitrary length and if not all levels are specified, the settings for
missing levels are derived at runtime based on the last specified
level (with target file size doubling for each level). But if we add
a level to the slice, that level gets the L0 defaults (including the
target file size).

This commits cleans this up: the field is now an array of size
`NumLevels` and the defaults are always calculated in the same way:
hardcoded defaults for L0, and for all other levels we take the
options from the level above and double the file size.


RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4800#-:-ORm6TtqBav5q-s9Xx7v:b-814g8k)* status: all files reviewed, 1 unresolved discussion (waiting on @annrpom)

___
*[`compaction_picker.go` line 264 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4800#-ORm2c2KAOM13OBW-ljw:-ORm6OLD0zWPsdNshgVO:b-z35gjr) ([raw file](https://github.com/cockroachdb/pebble/blob/e7a73108ee33d7f7eeb4354f9a15bf173de315fa/compaction_picker.go#L264)):*

i'm a bit confused here: i thought as long as `outputLevel` is 0, `baseLevel` can be whatever? if that is not the case, then we can delete the other `baseLevel == 0` branch below instead
</blockquote></details>

Ah, sorry, left over from some debugging (I missed the branch below).

RaduBerinde:
Hm, looks like there's a bug, I will investigate:
```
        copying filter: mismatched filters "testing_bloom_filter/bits_per_key=16" vs "rocksdb.BuiltinBloomFilter"
```

RaduBerinde:
Added a fix for that issue, which I can reproduce on `master`.

PR #4801, opened 2025-06-02
*: add BlobFileID

jbowens:
Add a BlobFileID type for identifying blob files. When we introduce rewriting and replacing blob files, a BlobReference's file identifier may not translate directly to a file on disk, so we should not use DiskFileNum.

Informs #112.


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4801#-:-ORkz6qD-qJ9ty7gmKNQ:bnfp4nl)* status: 0 of 26 files reviewed, all discussions resolved

jbowens:
TFTR!

PR #4804, opened 2025-06-02
block: clean up compression buffers

RaduBerinde:
#### block: rename BufHandle to TempBuffer, use single pool

Using two separate temp buffer pools for compressed and uncompressed
data seems unnecessary and likely leads to more memory usage. Switch
to a single temp buffer pool (which simplifies the code) and rename
`BufHandle` to `TempBuffer`.

#### block: remove Buffer

The `block.Buffer` type does very little and is barely used. Use
`TempBuffer` directly instead.

#### block: remove CompressAndChecksum variant without compressor

We now require a compressor for `CompressAndChecksum`.

### DIFF
--- a/sstable/blob/blob.go
+++ b/sstable/blob/blob.go
@@ -116,7 +116,7 @@ type FileWriter struct {
 
 type compressedBlock struct {
 	pb  block.PhysicalBlock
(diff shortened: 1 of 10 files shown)


RaduBerinde:
TFTR!

PR #4805, opened 2025-06-02
internal/manifest: genericize btree

jbowens:
Convert the B-Tree used for organizing and reference counting TableMetadata
into a generic collection. This is in preparation for storing blob file
metadata in a B-Tree for copy-on-write reference counting. The iterator type is
left specialized to *TableMetadata because we will not need arbitrary iteration
over the blob file B-Tree.

We can consider eventually using an alternative B-Tree implementation, like the
one in the btreemap package, but we would need to introduce support for
copy-on-write reference counting. The existing table metadata B-Tree
implementation supports the reference counting semantics we need.

### DIFF
--- a/internal/manifest/annotator.go
+++ b/internal/manifest/annotator.go
@@ -79,7 +79,7 @@ type annotation struct {
 	valid atomic.Bool
 }
 
-func (a *Annotator[T]) findExistingAnnotation(n *node) *annotation {
+func (a *Annotator[T]) findExistingAnnotation(n *node[*TableMetadata]) *annotation {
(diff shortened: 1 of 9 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4805#-:-ORm5xwmAe5JkOJOe0qR:bnfp4nl)* status: 0 of 9 files reviewed, all discussions resolved (waiting on @xxmplus)

jbowens:
TFTR!

Issue #4802, opened 2025-06-02
manifest: support recording replaced blob files


jbowens:
When a blob file is replaced, we need to record the replacement within the manifest. Additionally, we need to decouple the reference counting of the old file from the extant TableMetadata BlobReferences.

Jira issue: PEBBLE-464

PR #4806, opened 2025-06-02
manifest: rework blob file reference counting

jbowens:
Rework the reference counting of in-use blob files to decouple the reference counts from the TableMetadatas' blob references. With the introduction of blob file rewrites and replacement, the physical blob file and the logical blob file's references have distinct lifetimes. A physical blob file that has been replaced will need to be removed once all Versions that predate the replacement have been unreferenced.

This commit renames the struct previously known as BlobFileMetadata to PhysicalBlobFile, describing metadata particular to a physical file backing a logical blob file. A new BlobFileMetadata struct is introduced that holds a BlobFileID and a pointer to the PhysicalBlobFile.

This commit additionally adapts the physical blob file referencing to occur through a separate B-Tree of BlobFileMetadata structs. Similar to TableMetadata, references to blob files are maintained by copy-on-write B-Tree nodes which themselves are reference counted. A PhysicalBlobFile's reference count is incremented when a new B-Tree node references a containing BlobFileMetadata, and it's decremented when the B-Tree node's reference count falls to zero. This indirection ensures that a mutation to the set of blob files only performs log(n) work. Since addition and removal of blob files within a version is now modeled directly (as opposed to indirectly via TableMetadata BlobReferences), a blob file may be removed or replaced within the set before all referencing TableMetadata are removed.

Informs #4802.

### DIFF
--- a/checkpoint.go
+++ b/checkpoint.go
@@ -354,13 +354,12 @@ func (d *DB) Checkpoint(
 	// When we write the MANIFEST of the checkpoint, we'll include a final
 	// VersionEdit that removes these blob files so that the checkpointed
 	// manifest is consistent.
-	var excludedBlobFiles map[base.DiskFileNum]*manifest.BlobFileMetadata
+	var excludedBlobFiles map[base.BlobFileID]*manifest.PhysicalBlobFile
 	if len(includedBlobFiles) < len(versionBlobFiles) {
-		excludedBlobFiles = make(map[base.DiskFileNum]*manifest.BlobFileMetadata, len(versionBlobFiles)-len(includedBlobFiles))
-		for _, blobFile := range versionBlobFiles {
-			if _, ok := includedBlobFiles[blobFile.FileID]; !ok {
-				diskFileNum := blob.DiskFileNumTODO(blobFile.FileID)
-				excludedBlobFiles[diskFileNum] = blobFile
+		excludedBlobFiles = make(map[base.BlobFileID]*manifest.PhysicalBlobFile, len(versionBlobFiles)-len(includedBlobFiles))
+		for _, meta := range versionBlobFiles {
+			if _, ok := includedBlobFiles[meta.FileID]; !ok {
+				excludedBlobFiles[meta.FileID] = meta.Physical
 			}
 		}
 	}
@@ -471,7 +470,7 @@ func (d *DB) writeCheckpointManifest(
 	manifestSize int64,
 	excludedTables map[manifest.DeletedTableEntry]*manifest.TableMetadata,
 	removeBackingTables []base.DiskFileNum,
-	excludedBlobFiles map[base.DiskFileNum]*manifest.BlobFileMetadata,
+	excludedBlobFiles map[base.BlobFileID]*manifest.PhysicalBlobFile,
 ) error {
 	// Copy the MANIFEST, and create a pointer to it. We copy rather
 	// than link because additional version edits added to the
--- a/compaction.go
+++ b/compaction.go
@@ -2478,7 +2478,7 @@ func (d *DB) handleCompactFailure(c *compaction, err error) {
 func (d *DB) cleanupVersionEdit(ve *manifest.VersionEdit) {
 	obsoleteFiles := manifest.ObsoleteFiles{
 		TableBackings: make([]*manifest.TableBacking, 0, len(ve.NewTables)),
-		BlobFiles:     make([]*manifest.BlobFileMetadata, 0, len(ve.NewBlobFiles)),
+		BlobFiles:     make([]*manifest.PhysicalBlobFile, 0, len(ve.NewBlobFiles)),
 	}
 	deletedTables := make(map[base.TableNum]struct{})
 	for key := range ve.DeletedTables {
@@ -2488,10 +2488,10 @@ func (d *DB) cleanupVersionEdit(ve *manifest.VersionEdit) {
 		obsoleteFiles.AddBlob(ve.NewBlobFiles[i])
 		d.mu.versions.zombieBlobs.Add(objectInfo{
 			fileInfo: fileInfo{
-				FileNum:  base.DiskFileNum(ve.NewBlobFiles[i].FileID),
+				FileNum:  ve.NewBlobFiles[i].FileNum,
 				FileSize: ve.NewBlobFiles[i].Size,
 			},
-			isLocal: objstorage.IsLocalBlobFile(d.objProvider, base.DiskFileNum(ve.NewBlobFiles[i].FileID)),
+			isLocal: objstorage.IsLocalBlobFile(d.objProvider, ve.NewBlobFiles[i].FileNum),
 		})
 	}
 	for i := range ve.NewTables {
@@ -3105,7 +3105,7 @@ func (d *DB) runCompaction(
 		// Delete any created tables or blob files.
 		obsoleteFiles := manifest.ObsoleteFiles{
 			TableBackings: make([]*manifest.TableBacking, 0, len(result.Tables)),
-			BlobFiles:     make([]*manifest.BlobFileMetadata, 0, len(result.Blobs)),
+			BlobFiles:     make([]*manifest.PhysicalBlobFile, 0, len(result.Blobs)),
 		}
 		d.mu.Lock()
 		for i := range result.Tables {
@@ -3291,7 +3291,7 @@ func (c *compaction) makeVersionEdit(result compact.Result) (*manifest.VersionEd
 		}
 	}
 	// Add any newly constructed blob files to the version edit.
-	ve.NewBlobFiles = make([]*manifest.BlobFileMetadata, len(result.Blobs))
+	ve.NewBlobFiles = make([]*manifest.PhysicalBlobFile, len(result.Blobs))
 	for i := range result.Blobs {
 		ve.NewBlobFiles[i] = result.Blobs[i].Metadata
 	}
--- a/data_test.go
+++ b/data_test.go
@@ -927,7 +927,7 @@ func runDBDefineCmdReuseFS(td *datadriven.TestData, opts *Options) (*DB, error) 
 	// them to the final version edit.
 	valueSeparator := &defineDBValueSeparator{
 		pbr:   &preserveBlobReferences{},
-		metas: make(map[base.BlobFileID]*manifest.BlobFileMetadata),
+		metas: make(map[base.BlobFileID]*manifest.PhysicalBlobFile),
 	}
 
 	var mem *memTable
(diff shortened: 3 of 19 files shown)


sumeerbhola (changes requested):
Made an initial pass.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4806#-:-ORs7psU8F_MEj93kxUP:b4kovv8)* status: 2 of 18 files reviewed, 14 unresolved discussions

___
*[`internal/manifest/version_edit.go` line 166 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4806#-ORruQBp3kvMTX1dPgIG:-ORruQBp3kvMTX1dPgIH:bwhwe21) ([raw file](https://github.com/cockroachdb/pebble/blob/751e1d0be7e7bdb597616bb171ef48353ea24d2b/internal/manifest/version_edit.go#L166)):*
> ```Go
> 	// While replaying a MANIFEST, the values are nil. Otherwise the values must
> 	// not be nil.
> 	DeletedBlobFiles map[base.BlobFileID]*PhysicalBlobFile
> ```

I'm getting confused here on why `PhysicalBlobFile` is sufficient for `NewBlobFiles`.

When we create an original blob the `PhysicalBlobFile` is sufficient since the only additional information provided by the `BlobFileMetadata` is the `FileID` which is the same as in the `PhysicalBlobFile`. But when we rewrite an original or an already rewritten blob file, we need the `BlobFileMetadata` of the new blob, since we need to preserve that mapping from the original `FileID` as all the sstables that are not in this version edit use that original `FileID`.

For deleting a blob file it seems the `PhysicalBlobFile` is sufficient, since a physical blob file's FileID is never referenced by more than one original blob file FileID. Though perhaps it would be easier to have the full `BlobFileMetadata` to quickly check there are no dangling references.
___
*[`internal/manifest/blob_metadata.go` line 25 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4806#-ORriyse6PMCD6cItM7b:-ORriyse6PMCD6cItM7c:b-4l4mhk) ([raw file](https://github.com/cockroachdb/pebble/blob/751e1d0be7e7bdb597616bb171ef48353ea24d2b/internal/manifest/blob_metadata.go#L25)):*
> ```Go
> // A BlobReference describes a sstable's reference to a blob value file.
> type BlobReference struct {
> 	// FileID identifies the referenced blob file.
> ```

Consider saying that a `BlobReference` is immutable. And that this is always the original `FileID`, i.e., if a value reference gets reused in a compaction it continues to reference the original FileID, even if that FileID has already been rewritten.
___
*[`internal/manifest/blob_metadata.go` line 36 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4806#-ORrRHojBjURyBrtqnRo:-ORrRHojBjURyBrtqnRp:b-untwd1) ([raw file](https://github.com/cockroachdb/pebble/blob/751e1d0be7e7bdb597616bb171ef48353ea24d2b/internal/manifest/blob_metadata.go#L36)):*
> ```Go
> 
> 	// OriginalMetadata is the metadata for the original physical blob file. It
> 	// is non-nil for blob references contained within active Versions. It is
> ```

I am unsure about this "is non-nil" comment, that is inherited from the existing code. I don't see us setting `BlobReference.Metadata` to nil in `TableMetadata.Unref`. So isn't it always non-nil?
___
*[`internal/manifest/blob_metadata.go` line 48 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4806#-ORrSGGl2Jota04PagJf:-ORrSGGl2Jota04PagJg:biy66ai) ([raw file](https://github.com/cockroachdb/pebble/blob/751e1d0be7e7bdb597616bb171ef48353ea24d2b/internal/manifest/blob_metadata.go#L48)):*
> ```Go
> 	// TODO(jackson): We should either remove the OriginalMetadata from the
> 	// BlobReference or use it to infer when a blob file has definitely NOT been
> 	// replaced and a lookup in Version.BlobFileSet is unnecessary.
> ```

I suppose the interpolation we do in `BlobReference.EstimatedPhysicalSize` can continue to use the original metadata without translation? Is that the only reason this exists? Given our in-person discussion today, I'd lean towards removing this and recovering any performance loss in other ways.
___
*[`internal/manifest/blob_metadata.go` line 79 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4806#-ORrlhNq5rn1QmCEAFcV:-ORrlhNq5rn1QmCEAFcW:b-nndjnw) ([raw file](https://github.com/cockroachdb/pebble/blob/751e1d0be7e7bdb597616bb171ef48353ea24d2b/internal/manifest/blob_metadata.go#L79)):*
> ```Go
> // BlobFileMetadata encapsulates a blob file ID used to identify a particular
> // blob file, and a reference-counted physical blob file. See the BlobFileSet
> // documentation for more details.
> ```

Consider saying that different versions can have different `BlobFileMetadata` structs with the same `BlobFileMetadata.FileID`, and if so they necessarily refer to different `PhysicalBlobFile`s.
___
*[`internal/manifest/blob_metadata.go` line 82 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4806#-ORrkZIW30ih0Xlt2qIF:-ORrkZIW30ih0Xlt2qIG:biqwuga) ([raw file](https://github.com/cockroachdb/pebble/blob/751e1d0be7e7bdb597616bb171ef48353ea24d2b/internal/manifest/blob_metadata.go#L82)):*
> ```Go
> type BlobFileMetadata struct {
> 	// FileID is the file ID used to identify the blob file within
> 	// BlobReferences.
> ```

The phrase "used to identify the blob file within BlobReferences" could be made more precise. Consider saying these are always original blob file ids, i.e., the same domain as BlobReference.FileID.
___
*[`internal/manifest/blob_metadata.go` line 87 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4806#-ORrlOUnA_3waFhsf8bg:-ORrlOUnA_3waFhsf8bh:be545zl) ([raw file](https://github.com/cockroachdb/pebble/blob/751e1d0be7e7bdb597616bb171ef48353ea24d2b/internal/manifest/blob_metadata.go#L87)):*
> ```Go
> 	//
> 	// If the blob file has been replaced, Physical.FileNum ≠ FileID.
> 	Physical *PhysicalBlobFile
> ```

Always non-nil, yes?
___
*[`internal/manifest/blob_metadata.go` line 103 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4806#-ORrlRALB4s074c7n1Tr:-ORrlRALB4s074c7n1Ts:bp6g7az) ([raw file](https://github.com/cockroachdb/pebble/blob/751e1d0be7e7bdb597616bb171ef48353ea24d2b/internal/manifest/blob_metadata.go#L103)):*
> ```Go
> // contains one.
> func (m BlobFileMetadata) Ref() {
> 	if m.Physical != nil {
> ```

why should this ever be nil?
___
*[`internal/manifest/blob_metadata.go` line 112 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4806#-ORrlTT05fBYgXmFWenb:-ORrlTT05fBYgXmFWenc:b-7ropln) ([raw file](https://github.com/cockroachdb/pebble/blob/751e1d0be7e7bdb597616bb171ef48353ea24d2b/internal/manifest/blob_metadata.go#L112)):*
> ```Go
> // the provided obsolete files set.
> func (m BlobFileMetadata) Unref(of ObsoleteFilesSet) {
> 	if m.Physical != nil {
> ```

ditto
___
*[`internal/manifest/blob_metadata.go` line 312 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4806#-ORrnq3hBk9QXzsnvMoQ:-ORrnq3hBk9QXzsnvMoR:ba0kryp) ([raw file](https://github.com/cockroachdb/pebble/blob/751e1d0be7e7bdb597616bb171ef48353ea24d2b/internal/manifest/blob_metadata.go#L312)):*
> ```Go
> // entries remap a blob file ID to a new blob file. This remapping allows for
> // replacement of a physical blob file without replacing the referencing
> // TableMetadatas.
> ```

Consider saying 
- the key is BlobFileMetadata.FileID.
- whether it is a remapping or not is hidden inside BlobFileMetadata which follows the same structure for both.
___
*[`internal/manifest/blob_metadata.go` line 341 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4806#-ORrocsZ7Sqo-Ti0lWZ2:-ORrocsZ7Sqo-Ti0lWZ3:b2k846k) ([raw file](https://github.com/cockroachdb/pebble/blob/751e1d0be7e7bdb597616bb171ef48353ea24d2b/internal/manifest/blob_metadata.go#L341)):*
> ```Go
> 	// Version's B-Tree should maintain a reference. So a call to remove()
> 	// should never result in a file's reference count dropping to zero, and we
> 	// pass assertNoObsoleteFiles{} to assert such.
> ```

This is because this is only happening as part of creating the new `Version` so we are holding a ref to the current (soon to be previous) `Version`, yes? A comment like this in `BulkVersionEdit.Apply` is easier to grasp since the context is already there, but providing a bit more context here since this is a standalone function would be helpful.
___
*[`internal/manifest/blob_metadata.go` line 385 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4806#-ORrrM2X0BFO-_SRZmPM:-ORrrM2X0BFO-_SRZmPN:b-iw7szg) ([raw file](https://github.com/cockroachdb/pebble/blob/751e1d0be7e7bdb597616bb171ef48353ea24d2b/internal/manifest/blob_metadata.go#L385)):*
> ```Go
> 	// files is a map of blob file IDs to *currentBlobFiles, recording metadata
> 	// about the blob file's active references in the latest Version.
> 	files map[base.BlobFileID]*currentBlobFile
> ```

Worth saying that the key is `currentBlobFile.metadata.FileID`, so original blob file ids.
Given how often we are saying this, I wonder if we should make it a different type.
___
*[`internal/manifest/blob_metadata.go` line 391 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4806#-ORrqrTx3SIG_pZNF8pC:-ORrqrTx3SIG_pZNF8pD:b7rnc72) ([raw file](https://github.com/cockroachdb/pebble/blob/751e1d0be7e7bdb597616bb171ef48353ea24d2b/internal/manifest/blob_metadata.go#L391)):*
> ```Go
> 
> type currentBlobFile struct {
> 	metadata BlobFileMetadata
> ```

Is this a copy of the struct in the btree for the latest version?
___
*[`internal/manifest/table_metadata.go` line 232 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4806#-ORrtamFDVnnVJxF-GbS:-ORrtamFDVnnVJxF-GbT:b-d1z17o) ([raw file](https://github.com/cockroachdb/pebble/blob/507d3bd3d9e48239d782fadbb4fb958939dc8d40/internal/manifest/table_metadata.go#L232)):*
> ```Go
> 	if v := m.refs.Add(1); v == 1 {
> 		m.TableBacking.Ref()
> 		for _, blobRef := range m.BlobReferences {
> ```

I see why this is removed since the `BlobFileSet` refs on its behalf as the same `TableMetadata` can ref different blob files in different versions.
Do we do a invariants.Enabled consistency check somewhere that a version's `BlobFileSet` has exactly the `FileIDs` that are referenced in the `TableMetadatas` in that version?

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 6 files at r5, 15 of 18 files at r6, 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4806#-:-OS0ifnBArg2T9fAgGxm:bg7igkv)* status: all files reviewed, 14 unresolved discussions (waiting on @jbowens)

jbowens:
TFTR!

Issue #4792, opened 2025-05-29
pebble: add `RawKeySize`, `RawValueSize` fields to table stats


xinhaoz:
In #4775, we changed [estimateSizesBeneath](https://github.com/cockroachdb/pebble/blob/d1cb41205386e0fd2f8db0cca06baeed70f5aba4/table_stats.go#L486) to read the properties block to load these fields. Instead, we should just add them to the table stats [[1](https://github.com/cockroachdb/pebble/blob/d1cb41205386e0fd2f8db0cca06baeed70f5aba4/table_stats.go#L317-L320), [2](https://github.com/cockroachdb/pebble/blob/master/table_stats.go#L710)].

Jira issue: PEBBLE-460

PR #4809, opened 2025-06-02
sstable: add `RawKeySize` and `RawValueSize` to TableStats

xinhaoz:
Add these properties to table stats so we don't need to load the properties block when calculating table stats.

Closes: #4792

### DIFF
--- a/ingest_test.go
+++ b/ingest_test.go
@@ -216,7 +216,7 @@ func TestIngestLoadRand(t *testing.T) {
 			w := sstable.NewRawWriter(objstorageprovider.NewFileWritable(f), sstable.WriterOptions{
 				TableFormat: version.MaxTableFormat(),
 			})
-			var count uint64
+			var count, rawKeySize uint64
 			for i := range keys {
 				if i > 0 && base.InternalCompare(cmp, keys[i-1], keys[i]) == 0 {
 					// Duplicate key, ignore.
@@ -224,7 +224,9 @@ func TestIngestLoadRand(t *testing.T) {
 				}
 				require.NoError(t, w.Add(keys[i], nil, false /* forceObsolete */))
 				count++
+				rawKeySize += uint64(keys[i].Size())
 			}
+			expected[i].Stats.RawKeySize = rawKeySize
 			expected[i].Stats.NumEntries = count
 			require.NoError(t, w.Close())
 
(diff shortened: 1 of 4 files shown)


jbowens (commented):
> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4809)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @annrpom)

___
*[`table_stats.go` line 513 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4809#-ORqQuLx5X0rVRl2g0Gk:-ORqQuLx5X0rVRl2g0Gl:b96pzpr) ([raw file](https://github.com/cockroachdb/pebble/blob/bd32285fdc344eab9b0963f4726ef3790811006e/table_stats.go#L513)):*
> ```Go
> 
> 	for l := level + 1; l < numLevels; l++ {
> 		for tableBeneath := range v.Overlaps(l, meta.UserKeyBounds()).All() {
> ```

I forgot that the table stats aren't guaranteed to already be populated for these files. We can check if they're present using `TableMetadata.StatsValid()`.

If `TableMetadata.StatsValid()` we can use `Stats.{NumEntries,RawKeySize,RawValueSize}`, and we don't actually use the `*sstable.Reader` down below and can avoid calling `d.fileCache.withReader`. Unfortunately if they're not valid, I think we need to fall back to the previous logic that calls `withReader` and loads the properties block.

xinhaoz (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4809)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @annrpom and @jbowens)

___
*[`table_stats.go` line 513 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4809#-ORqQuLx5X0rVRl2g0Gk:-ORqX9Nx96RxW1JXDRfg:b-737fem) ([raw file](https://github.com/cockroachdb/pebble/blob/bd32285fdc344eab9b0963f4726ef3790811006e/table_stats.go#L513)):*

I forgot that the table stats aren't guaranteed to already be populated for these files. We can check if they're present using `TableMetadata.StatsValid()`.

If `TableMetadata.StatsValid()` we can use `Stats.{NumEntries,RawKeySize,RawValueSize}`, and we don't actually use the `*sstable.Reader` down below and can avoid calling `d.fileCache.withReader`. Unfortunately if they're not valid, I think we need to fall back to the previous logic that calls `withReader` and loads the properties block.
</blockquote></details>

Oops, noted - done!

PR #4812, opened 2025-06-03
sstable: add blob reference value liveness index block

annrpom:
### colblk: introduce value liveness block encoder, decoder
This commit adds an encoder and a decoder that will be utilized for our
blob reference index block later on.

### db: add blobRefValueLivenessWriter
We introduce the following special encoding that keeps track of value liveness:
```
  <reference ID> <block ID> <values size> <ellision tag byte> [<bitmap>]
```

### sstable: add blob reference value liveness index block
This commit adds a new block, `blob-reference-index`, that tracks the
which values within blob value blocks are referenced by our sstable.


annrpom:
TestCompaction/value_separation seems to be non-deterministic in sst size. It looks like this is due to the `<values size>` part of the encoding varying between runs - cc: @jbowens does this seem probable?

I am wondering if this observation is just a side effect of some bad code. For now, I'm commenting it out

jbowens (commented):
Didn't make a full pass yet, but left a few comments. Looking good! :sunglasses: 

> Reviewed 4 of 4 files at r1, 23 of 23 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4812#-:-OS6Khev-jJUdn_2coJH:b-tfeqxi)* status: 4 of 27 files reviewed, 8 unresolved discussions (waiting on @annrpom)

___
*[`sstable/colblk/value_liveness_block.go` line 1 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4812#-OS12tkN4DUSLym30Enl:-OS12tkN4DUSLym30Enm:bwci63u) ([raw file](https://github.com/cockroachdb/pebble/blob/43cb726db88dc72d16586112b7ce596c3e619e82/sstable/colblk/value_liveness_block.go#L1)):*
> ```Go
> // Copyright 2024 The LevelDB-Go and Pebble Authors. All rights reserved. Use
> ```

nit: happy new year!
___
*[`sstable/colblk/value_liveness_block.go` line 24 at r4](https://reviewable.io/reviews/cockroachdb/pebble/4812#-OS0zma48p38jkdrH6pA:-OS0zma48p38jkdrH6pB:bu0816z) ([raw file](https://github.com/cockroachdb/pebble/blob/43cb726db88dc72d16586112b7ce596c3e619e82/sstable/colblk/value_liveness_block.go#L24)):*
> ```Go
> // A value liveness block is a columnar block that contains a single column:
> // an array of bytes encoding the liveness of values within a blob files
> // referenced by an sstable.
> ```

nit: maybe clarify that the indexes into this array are the `blob.ReferenceID`s contained within the sstable
___
*[`value_separation.go` line 174 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4812#-OS6FPNr5bYMn5EXPcQS:-OS6FPNr5bYMn5EXPcQT:bhvt01u) ([raw file](https://github.com/cockroachdb/pebble/blob/66ad95fedc05e074af249668077eaab06d7d642e/value_separation.go#L174)):*
> ```Go
> 	//
> 	// These indexes let us quickly determine if the bitmap is:
> 	// - all zeros (idxLastZero is set and equals last index in our bitmap)
> ```

nit: wouldn't all zeros be `idxLastOne` is unset?
___
*[`value_separation.go` line 192 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4812#-OS6HyFq6Z6r-TfKj_ym:-OS6HyFq6Z6r-TfKj_yn:bsviah0) ([raw file](https://github.com/cockroachdb/pebble/blob/66ad95fedc05e074af249668077eaab06d7d642e/value_separation.go#L192)):*
> ```Go
> type blobRefValueLivenessWriter struct {
> 	bufs     [][]byte
> 	stateMap map[blobBlockIdentifier]blobRefValueLivenessState
> ```

because we should observe values in order, I think we can maintain 1 state struct per-`ReferenceID`. We could also replace this map with a slice, where the index is the `ReferenceID`. Then when we step into a new `blockID` within a particular blob file, we append what we've buffered to a running buffer and clear the block-specific state. 
___
*[`value_separation.go` line 203 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4812#-OS6DS57EZdnfWVYUDx6:-OS6DS57EZdnfWVYUDx7:bqrkkjo) ([raw file](https://github.com/cockroachdb/pebble/blob/66ad95fedc05e074af249668077eaab06d7d642e/value_separation.go#L203)):*
> ```Go
> }
> 
> // addLiveValue adds a live value to the state mantained by (refID, blockID).
> ```

nit: maintained
___
*[`value_separation.go` line 213 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4812#-OS6JbOY6fui8qjrIXdB:-OS6JbOY6fui8qjrIXdC:b-9xic2p) ([raw file](https://github.com/cockroachdb/pebble/blob/66ad95fedc05e074af249668077eaab06d7d642e/value_separation.go#L213)):*
> ```Go
> 	state.lastSeenValueID = valueID
> 	state.valuesSize += valueSize
> 	state.bitmap = append(state.bitmap, 1)
> ```

I think it might be valuable to define a separate bitmap type that handles maintaining the bitmap in-memory and then the serialization to a buffer. And instead of using 1-byte per bit, we could keep it down to a 1-bit per 1-bit. The `colblk.BitmapBuilder` type has some prior art to draw from.. maybe we should chat about it IRL Monday. 
___
*[`value_separation.go` line 224 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4812#-OS6ET5HDliYG_Nx9_s1:-OS6ET5HDliYG_Nx9_s2:b8et9a9) ([raw file](https://github.com/cockroachdb/pebble/blob/66ad95fedc05e074af249668077eaab06d7d642e/value_separation.go#L224)):*
> ```Go
> 
> // addDeadValue adds a dead value to the state mantained by (refID, blockID).
> func (w *blobRefValueLivenessWriter) addDeadValue(refID blob.ReferenceID, blockID blob.BlockID) {
> ```

i'm wondering if we need a `addDeadValue` func at all-- maybe `addLiveValue` should grow the bitmap as necessary, handling the job of implicitly updating state around dead values. This part will end up being performance-sensitive so we want to keep it as tight as possible
___
*[`value_separation.go` line 295 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4812#-OS6IDir8BGM-jkv2P5V:-OS6IDir8BGM-jkv2P5W:b-pbtn3f) ([raw file](https://github.com/cockroachdb/pebble/blob/66ad95fedc05e074af249668077eaab06d7d642e/value_separation.go#L295)):*
> ```Go
> func (w *blobRefValueLivenessWriter) finishOutput() error {
> 	var err error
> 	for k, state := range w.stateMap {
> ```

I think the nondeterminism is coming from this map iteration which will encode these in a random order (and then compress a different amount, depending on the order). we'll need all the blocks in-order by blockID.

I don't think we need to encode the reference ID, because the reference ID will be the index into that `RawBytes` column

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4812)* status: 0 of 32 files reviewed, 5 unresolved discussions (waiting on @jbowens)

___
*[`value_separation.go` line 192 at r5](https://reviewable.io/reviews/cockroachdb/pebble/4812#-OS6HyFq6Z6r-TfKj_ym:-OSUdhqA48wrw7B2PaSV:bhvt8d3) ([raw file](https://github.com/cockroachdb/pebble/blob/66ad95fedc05e074af249668077eaab06d7d642e/value_separation.go#L192)):*

trying to think this one through: so we spoke about observing `valueIDs` in order for a particular `(refID, blockID)` which makes sense since keys will maintain the same ordering through compactions. now we are assuming that we are also guaranteed that visited refIDs will be monotonically increasing? i think i need to think about this some more
</blockquote></details>

ah nvm! i misunderstood how `currReferences` are tracked; ok this makes sense

value_separation.go
```diff
@@ -150,6 +152,102 @@ func uniqueInputBlobMetadatas(levels []compactionLevel) []*manifest.PhysicalBlob
...
+}
+
+// addLiveValue adds a live value to the state maintained by refID. If the
+// current blockID for this in-progress state is different from the provided
+// blockID, a new state is created and the old one is preserved to the buffer
+// at w.bufs[refID].
+func (w *blobRefValueLivenessWriter) addLiveValue(
+	refID blob.ReferenceID, blockID blob.BlockID, valueID blob.BlockValueID, valueSize uint64,
+) {
+	state := w.refState[refID]
```
jbowens, line 213: I think we need to take a pointer here `&w.refState[refID]`, otherwise we'll be mutating the local struct copy down below

value_separation.go
```diff
@@ -150,6 +152,102 @@ func uniqueInputBlobMetadatas(levels []compactionLevel) []*manifest.PhysicalBlob
...
+	if state.blockID != blockID {
+		w.bufs[refID] = state.finishOutput(w.bufs[refID])
+		state = blobRefValueLivenessState{}
```
jbowens, line 216: nit: this should be unnecessary once `blobRefValueLivenessState.init` also clears valueID

value_separation.go
```diff
@@ -150,6 +152,102 @@ func uniqueInputBlobMetadatas(levels []compactionLevel) []*manifest.PhysicalBlob
...
+	w.refState[refID] = state
+}
+
+// maybeAddNewState adds a new state for the provided referenceID if one does
+// not already exist. It assumes that blob.ReferenceIDs are visited in
+// monotonically increasing order.
+func (w *blobRefValueLivenessWriter) maybeAddNewState(
+	refID blob.ReferenceID, blockID blob.BlockID, valueID blob.BlockValueID,
+) {
+	if len(w.refState) <= int(refID) {
```
jbowens, line 231: we expect `len(w.refState) >= int(refID)` and want to add a new reference if `len(w.refState) == int(refID)`. I think we should panic with an assertion error if `len(w.refState) < int(refID)`

value_separation.go
```diff
@@ -150,6 +152,102 @@ func uniqueInputBlobMetadatas(levels []compactionLevel) []*manifest.PhysicalBlob
...
+	blockID    blob.BlockID
+	valueID    blob.BlockValueID
+	valuesSize uint64
+}
+
+// init initializes the state, resetting all fields to their initial values.
+func (s *blobRefValueLivenessState) init(refID blob.ReferenceID, blockID blob.BlockID) {
+	s.bitmap.Init()
+	s.refID = refID
+	s.blockID = blockID
```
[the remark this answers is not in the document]
jbowens: on second thought, maybe we don't need the valueID field at all?

value_separation.go
```diff
@@ -150,6 +152,102 @@ func uniqueInputBlobMetadatas(levels []compactionLevel) []*manifest.PhysicalBlob
 	return metadatas
 }
 
+// blobRefValueLivenessState tracks the liveness of values within a blob value
+// block via a sstable.BitmatRunLengthEncoder.
+type blobRefValueLivenessState struct {
```
jbowens, line 157: what do you think about moving this onto the `sstable.RawColumnWriter` and moving the integration point to `RawColumnWriter.AddWithBlobHandle`? That would ensure that every blob file has this value liveness state for all the blob handles it encodes, even if it's constructed through some other means (eg, if someday we allow ingesting a sstable and a blob file it references together).

jbowens (commented):
I got lost in reviewable so I switched to github reviews

sstable/blob_reference_index.go
```diff
@@ -0,0 +1,126 @@
...
+	refID blob.ReferenceID, blockID blob.BlockID, valueID blob.BlockValueID, valueSize uint64,
+) {
+	state := &w.refState[refID]
+	if state.blockID != blockID {
+		w.bufs[refID] = state.finishOutput(w.bufs[refID])
+		state.init(refID, blockID)
+	}
+	state.valuesSize += valueSize
+	state.bitmap.Set(int(valueID))
+	w.refState[refID] = *state
```
jbowens, line 81: this line should be unnecessary, since `state` is already a pointer to the struct stored in `refState[refID]`

sstable/colblk_writer.go
```diff
@@ -443,6 +444,10 @@ func (w *RawColumnWriter) AddWithBlobHandle(
 		return err
 	}
 	w.props.NumValuesInBlobFiles++
+	if err := w.blobRefLivenessIndexBlock.maybeAddNewState(h.ReferenceID, h.BlockID); err != nil {
+		return err
+	}
+	w.blobRefLivenessIndexBlock.addLiveValue(h.ReferenceID, h.BlockID, h.ValueID, uint64(h.ValueLen))
```
jbowens, line 450: thoughts on moving the call to `maybeAddNewState` into `addLiveValue`? Then we can hide a bit more of the internals of the liveness index block. Unfortunately since this is a hot path, I also think we should lift the check to see whether we need to add a new state also into addLiveValue. That would ensure that in the common happy case that we've already seen the referenceID, we don't need to pay the cost of a function call

sstable/colblk_writer.go
```diff
@@ -1002,6 +1007,19 @@ func (w *RawColumnWriter) Close() (err error) {
 		w.props.ValueBlocksSize = vbStats.ValueBlocksAndIndexSize
 	}
 
+	// Write the blob reference index block if non-empty.
+	if len(w.blobRefLivenessIndexBlock.bufs) > 0 {
```
jbowens, line 1011: nit: thoughts on exposing a `NumReferences()` method and calling it here to make this conditional a little clearer?

sstable/table.go
```diff
@@ -225,11 +226,13 @@ const (
...
-	metaPropertiesName = "rocksdb.properties"
-	metaRangeDelV1Name = "rocksdb.range_del"
-	metaRangeDelV2Name = "rocksdb.range_del2"
+	metaRangeKeyName        = "pebble.range_key"
+	metaValueIndexName      = "pebble.value_index"
+	metaPropertiesName      = "rocksdb.properties"
+	metaRangeDelV1Name      = "rocksdb.range_del"
+	metaRangeDelV2Name      = "rocksdb.range_del2"
+	metaBlobRefIndexName    = "pebble.blob_ref_index"
+	metaBlobRefLivenessName = "pebble.blob_ref_liveness"
```
jbowens, line 235: this guy is unused, right? I think we use `metaBlobRefIndexName` but not `metaBlobRefLivenessName`

sstable/blob_reference_index.go
```diff
@@ -0,0 +1,126 @@
...
+// block to the provided buffer, returning the modified buffer. The encoding is:
+//
+//	<block ID> <values size> <n bytes of bitmap> [<bitmap>]
+func (s *blobRefValueLivenessState) finishOutput(buf []byte) []byte {
+	buf = binary.AppendUvarint(buf, uint64(s.blockID))
+	buf = binary.AppendUvarint(buf, s.valuesSize)
+	var bitmap []byte
+	// Save the results of FinishAndAppend in case we haven't flushed in-progress
+	// bytes to the buf yet.
+	bitmap = s.bitmap.FinishAndAppend(bitmap)
```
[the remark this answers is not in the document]
annrpom: ah right we get that from `allSetRunLength` now

sstable/blob_reference_index.go
annrpom: oops; done

sstable/blob_reference_index.go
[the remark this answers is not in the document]
jbowens: ah, no, it was my brain that was mush. I forgot that bytes that aren't 0x00 or 0xFF don't use a run-length encoding. We'll need to keep the encoded length of the bitmap. Maybe we can compute and append the size of the encoded bitmap before calling FinishAndAppend, so we can still avoid the extra copy

sstable/colblk_writer.go
annrpom: > That would ensure that in the common happy case that we've already seen the referenceID, we don't need to pay the cost of a function call

ah cool that makes sense; done

sstable/blob_reference_index.go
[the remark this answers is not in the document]
annrpom: yeah i think size can change to precompute this stuff -- i took a stab at it

jbowens (approved):
:lgtm_strong:

annrpom:
strong TFTR! ('-')7 ('-')7

PR #4813, opened 2025-06-03
crl-release-24.3: metamorphic: tolerate extra arg for IngestAndExcise

RaduBerinde:

Informs #4729

### DIFF
--- a/open.go
+++ b/open.go
@@ -173,12 +173,6 @@ func Open(dirname string, opts *Options) (db *DB, err error) {
 					dirname)
 			}
 		}()
-	} else {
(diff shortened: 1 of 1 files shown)


RaduBerinde:
Got another failure, I will investigate and try to fix in the same PR.
```
database "/tmp/TestMetaCrossVersion2046501096/001/1748966661952068137/HEAD_HEAD_9407900762868009557_005/_meta/250603-160432.242512161165/standard-029/data" configured with shared objects but written in too old format major version 15
```

RaduBerinde:
Fixed that, now another: `shared ingested sstable has a lower table format than expected`

RaduBerinde:
The `shared ingested sstable has a lower table format than expected` failure is pretty hard to reproduce. I will investigate separately.

TFTR!

PR #4815, opened 2025-06-03
crl-release-25.1: metamorphic: more crossversion fixes

RaduBerinde:
Informs #4729

#### open: remove version check for shared objects

Remove a check that was too strict; even if the existing store version
is older, we will ratchet it up to `opts.FormatMajorVersion`, which
must support shared objects (checked by `opts.Validate()`).

#### metamorphic: tolerate integers for user key suffixes

This prevents a crossversion failure when parsing the NewIter op,
which switched from integers to strings in the filterMin/filterMax
arguments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,8 +62,8 @@ stressmeta: override TESTS = TestMeta$$
 stressmeta: stress
 
 .PHONY: crossversion-meta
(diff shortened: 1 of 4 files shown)


RaduBerinde:
TFTR!

PR #4818, opened 2025-06-03
crl-release-25.2: metamorphic: show op name on parsing panics

RaduBerinde:
#### open: remove version check for shared objects

Remove a check that was too strict; even if the existing store version
is older, we will ratchet it up to `opts.FormatMajorVersion`, which
must support shared objects (checked by `opts.Validate()`).

#### metamorphic: show op name on parsing panics

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ crossversion-meta:
 
 .PHONY: stress-crossversion
 stress-crossversion:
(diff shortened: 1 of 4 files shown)


jbowens (approved):

RaduBerinde:
TFTR!

PR #4821, opened 2025-06-03
db: avoid provider lookup when pacing file deletions

jbowens:
When deciding whether to pace a file deletion, avoid looking up the object in the objstorage provider. We now have an isLocal field readily available on the obsoleteFile.


RaduBerinde (approved):

jbowens:
TFTR!

PR #4827, opened 2025-06-05
github: fix post-issue

RaduBerinde:
The `jq` command was printing out `null` and we were treating that as
an existing issue number and trying to add a comment to it.


RaduBerinde:
TFTR!

PR #4836, opened 2025-06-06
crl-release-25.1: crossversion: debugging improvements

RaduBerinde:
#### metamorphic: support --initial-state with --run/compare

Allow supplying an initial state when trying to reproduce a failure.

#### crossversion: only save the relevant run dir

We were saving the entire root dir which is huge; switch to saving
just the relevant run(s) (along with any initial states). Note that
the `rootDir -> r.dir` change was already made on master in #4282.


RaduBerinde:
TFTR!

PR #4838, opened 2025-06-06
options: fix missing case in WAL recovery dir check

RaduBerinde:
If the existing store did not have a `WALDir` set and the new options
have one, the store path itself must be either a failover secondary or
a recovery dir. This was not covered by the compatibility check (it
would have saved a lot of time debugging a crossversion failure).

We also make sure we check the previous WAL dir even if the `wal_dir=`
entry doesn't appear (as was the case in tests).

### DIFF
--- a/open.go
+++ b/open.go
@@ -480,7 +480,7 @@ func Open(dirname string, opts *Options) (db *DB, err error) {
 		if err != nil {
 			return nil, err
 		}
-		if err := opts.CheckCompatibility(previousOptions); err != nil {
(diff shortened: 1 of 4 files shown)


jbowens (approved):
:lgtm_strong: 

> Reviewed 2 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4838#-:-OS5vFUGEyn1WS0UO3ng:b-uwkowd)* status: 2 of 4 files reviewed, all discussions resolved

RaduBerinde:
TFTR!

PR #4840, opened 2025-06-06
manifest: explicitly encode BlobFileID in manifest

jbowens:
Adapt the NewBlobFiles section of a version edit to explicitly encode a BlobFileID rather than inferring it from the disk file number. This is in preparation for blob file replacement.

Informs #4802.


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4840)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @annrpom)

___
*[`internal/manifest/testdata/version_edit_apply` line 242 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4840#-OS61JbK-uLYrPpgj8VM:-OS61JbK-uLYrPpgj8VN:b-wrsg66) ([raw file](https://github.com/cockroachdb/pebble/blob/13e87ba7ba1caf260dc29689390866126abf343c/internal/manifest/testdata/version_edit_apply#L242)):*
> ```
> 
> apply v8 name=v9
>   add-blob-file: 000003 -> 000003 size:[20535 (20KB)] vals:[25935 (25KB)]
> ```

@RaduBerinde I'm thinking about keeping it so that it's unambiguous that there is this mapping going on. WDYT? Maybe the `BlobFileID` shouldn't even be the same number for blob files that _are_ backed by the same file... and different formatting for the `BlobFileID`?

RaduBerinde:
*[`internal/manifest/testdata/version_edit_apply` line 242 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4840#-OS61JbK-uLYrPpgj8VM:-OS6EvHs9uJyvHq0mTDq:b3y0mrs) ([raw file](https://github.com/cockroachdb/pebble/blob/13e87ba7ba1caf260dc29689390866126abf343c/internal/manifest/testdata/version_edit_apply#L242)):*

@RaduBerinde I'm thinking about keeping it so that it's unambiguous that there is this mapping going on. WDYT? Maybe the `BlobFileID` shouldn't even be the same number for blob files that _are_ backed by the same file... and different formatting for the `BlobFileID`?
</blockquote></details>

Maybe it can be just a blob-file-id:3 field like the others

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4840)* status: 0 of 20 files reviewed, 1 unresolved discussion (waiting on @annrpom and @RaduBerinde)

___
*[`internal/manifest/testdata/version_edit_apply` line 242 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4840#-OS61JbK-uLYrPpgj8VM:-OSK8RewEZD5csfKGDf7:b-tlnp63) ([raw file](https://github.com/cockroachdb/pebble/blob/13e87ba7ba1caf260dc29689390866126abf343c/internal/manifest/testdata/version_edit_apply#L242)):*

Maybe it can be just a blob-file-id:3 field like the others
</blockquote></details>

updated it so that the entirety of the physical blob file information to be under a `physical:{}` field

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4840#-:-OSKipe197vcEbhmwFmy:b-814g8k)* status: 0 of 27 files reviewed, 2 unresolved discussions (waiting on @annrpom)

PR #4851, opened 2025-06-09
metamorphic: fix WAL recovery dirs logic

RaduBerinde:
When we set up an initial state and the `WALDir` changed, we add the
previous directory as a WAL recovery dir. But the logic uses the
existence of a `wal` subdirectory; this is incorrect in the case where
the initial store used a `WALDir` in the past, but not in the most
recent iteration.

We fix this by finding the latest OPTIONS file and parsing it instead.
In addition, if the previous store did not use a WAL dir, we remove
any `wal` subdirectory that was left over. We make the analogous fixes
for the WAL secondary.

### DIFF
--- a/metamorphic/options.go
+++ b/metamorphic/options.go
@@ -7,6 +7,7 @@ package metamorphic
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"math"
 	"math/rand/v2"
 	"os"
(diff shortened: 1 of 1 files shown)


RaduBerinde:
*[`metamorphic/options.go` line 1002 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4851#-OSJsr_I6mVfgsziJEkg:-OSK01yu1W1ubMQXzWUQ:b-sy1yhb) ([raw file](https://github.com/cockroachdb/pebble/blob/4121c64e081b4348486a5ba93b47022492d99fe8/metamorphic/options.go#L1002)):*

did the existence of the stale WAL directory cause issues? My expectation would be that we would ignore any WALs in the directory because their file numbers indicated that they were old, and we would remove them shortly after start.
</blockquote></details>

Heh, no, I was going to ask you if they would be ignored, but then I realized I can just delete them. I'll try without this.

PR #4854, opened 2025-06-09
db: create shared objects only once the format is new enough

RaduBerinde:
We currently have a race when opening an older (pre-shared-storage)
store and having shared storage enabled in the object. We ratchet the
FMV to the appropriate one, but until that is done any
flushes/compactions can still create shared sstables. These sstables
will have an old format that is not supported with shared storage.

We modify the code paths that decide where to create objects to check
the FMV as well.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -303,7 +303,8 @@ type compaction struct {
 
 	grantHandle CompactionGrantHandle
 
-	opts objstorage.CreateOptions
+	tableFormat   sstable.TableFormat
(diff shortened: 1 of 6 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4854#-:-OSJuot4-FuHdQyq5rNq:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

PR #4857, opened 2025-06-09
pebble-release-2.0: db: fix obsolete file metric underflows

jbowens:
Previously a race existed in the calculation of obsolete file metrics resulting in underflow. The values within versionSet.metrics were reset and recalculated to reflect the set of files in versionSet.obsoleteTables whenever new files were added to obsoleteFiles. Additionally, the cleanup manager invoked a callback to decrease versionSet.metrics whenever a table was deleted.

The recalculation of versionSet.metrics could reset the metrics to less than the sum of outstanding pending deletes. When the cleanup manager eventually deleted the pending tables, these metrics would underflow.

This commit fixes the bug by maintaining separate stats for all files that have been enqueued for the cleanup manager and all files that have been successfully deleted. The volume of outstanding, pending deletions is the difference between the two.

For now, there's an additional wart that the set of files that are sitting in versionSet.obsoleteTables are still separately tracked.

Informs cockroachdb#4811.

### DIFF
--- a/db.go
+++ b/db.go
@@ -459,10 +459,16 @@ type DB struct {
 			noOngoingFlushStartTime time.Time
 		}
 
-		// Non-zero when file cleaning is disabled. The disabled count acts as a
-		// reference count to prohibit file cleaning. See
-		// DB.{disable,Enable}FileDeletions().
-		disableFileDeletions int
+		fileDeletions struct {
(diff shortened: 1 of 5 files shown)


RaduBerinde (approved):

jbowens:
TFTR!

PR #4858, opened 2025-06-09
crl-release-25.2: crossversion: debugging improvements

RaduBerinde:
#### metamorphic: support --initial-state with --run/compare

Allow supplying an initial state when trying to reproduce a failure.

#### metamorphic: support --initial-state with --try-to-reduce


RaduBerinde:
TFTR!

PR #4862, opened 2025-06-09
manifest: clean up SortBySeqNum, SortBySmallest

jbowens:
Update both SortBySeqNum and SortBySmallest to use slices.SortFunc instead of the more verbose, less-efficient sort.Interface. Additionally, unexport SortBySeqNum and move it into a test-only file because it is now only used by a unit test.


xinhaoz (approved):

jbowens:
TFTR!

PR #4863, opened 2025-06-10
crl-release-24.3: more fixes around shared objects and upgrade

RaduBerinde:
#### compaction: relax copy compaction check

If we are racing with the db version being ratcheted, it is possible
that we plan to use shared storage when planning the compaction but
the old table format prevents us from doing so.

#### db: remove format check for shared objects

This check is too strict: we could have a file that was ingested
through a non-shared ingestion (and which has no obsolete points)
which was then copied to shared storage.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -3076,7 +3076,8 @@ func (d *DB) newCompactionOutput(
 
 	// Prefer shared storage if present.
 	createOpts := objstorage.CreateOptions{
-		PreferSharedStorage: d.shouldCreateShared(c.outputLevel.level),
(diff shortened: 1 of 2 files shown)


jbowens (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4863#-:-OSPLHXSAdKz1deo_aej:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

PR #4864, opened 2025-06-10
crl-release-25.1: scripts: run-crossversion-meta fix aliasing of test binaries

RaduBerinde:
We add the SHA to the binary name to avoid collisions and only extract
the version when it's an crl-release branch.


RaduBerinde:
TFTRs!

PR #4871, opened 2025-06-10
manifest: fix bug in accumulating blob file deletions

jbowens:
Previously when accumulating multiple version edits in a BulkVersionEdit, later calls to Accumulate would wipe blob file deletions that had already been accumulated.


jbowens:
TFTR!

PR #4874, opened 2025-06-10
block: add FastestCompression profile

RaduBerinde:
Add a profile that automatically selects between Snappy and MinLZ
depending on the platform. This will replace the similar code in CRDB
(which is causing CI failures because of different cluster setting
defaults for different platforms).


RaduBerinde:
TFTR!

PR #4875, opened 2025-06-10
manifest: support blob file replacement

jbowens:
**manifest: explicitly encode deleted physical blob file numbers**

Adapt the version edit encoding of blob file deletions to encode both the
BlobFileID and the deleted physical blob file's file number. This reduces
ambiguity and makes explicit the intention when we encode a blob file
replacement.

**manifest: support blob file replacement**

Adapt CurrentBlobFileSet and VersionEdit application to support version edits
that remove a physical blob file and add a new physical blob file under the
same BlobFileID.

Informs #112.


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4875#-:-OSZzIXqAbLw1gxp3E7H:bnfp4nl)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @xxmplus)

___
*[`internal/manifest/version_edit.go` line 172 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4875#-OSZy4Uu9-fkmhZxScE5:-OSZy4Uu9-fkmhZxScE6:b-vdbn05) ([raw file](https://github.com/cockroachdb/pebble/blob/22ae27585b028ada56f576ac34d971fd8f57b2bf/internal/manifest/version_edit.go#L172)):*
> ```Go
> 	//
> 	// A physical blob file may become unused if the corresponding BlobFileID
> 	// becomes unreferenced during the version edit. In this case must the
> ```

[nit] s/must//

PR #4881, opened 2025-06-11
db: maintain a Version reference during compactions

jbowens:
Today, a compaction has an associated version (c.version) that is used sparingly in a few compaction edge cases. This version is never referenced and may become unreferenced and cleared while a compaction is running. This has generally been okay, because compaction picking must explicitly coordinate with running compactions so that sstables that are inputs to compactions aren't inputs to some other compactions.

With the introduction of blob files (#112) and blob file rewriting, a compaction may need to use the Version's BlobFileSet to lookup the mapping from a BlobFileID to a physical disk file number. To ensure the BlobFileSet is still valid to read, the Version needs to remain referenced.

This version ref-ing also allows us to remove some existing, subtle code that would reference an arbitrary latest version when a compaction began so that excises could read sstables without their removal.

Informs #4802.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -369,6 +369,10 @@ func (c *compaction) userKeyBounds() base.UserKeyBounds {
 
 type getValueSeparation func(JobID, *compaction, sstable.TableFormat) compact.ValueSeparation
 
+// newCompaction constructs a compaction from the provided picked compaction.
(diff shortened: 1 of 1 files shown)


jbowens:
TFTR!

PR #4887, opened 2025-06-12
crossversion: pass all previous ops

RaduBerinde:
#### metamorphic: tolerate extra arg for IngestAndExcise

Informs #4729

#### metamorphic: tolerate integers for user key suffixes

This prevents a crossversion failure when parsing the NewIter op,
which switched from integers to strings in the filterMin/filterMax
arguments.

#### crossversion: pass all previous ops

We are passing the ops from the previous run. We need to pass ops from
all the runs that culminated in the initial state; this is necessary
for issuing correct SingleDel operations.

We change `--previous-ops` to take a list of files, separated by a
comma.

### DIFF
--- a/internal/metamorphic/crossversion/crossversion_test.go
+++ b/internal/metamorphic/crossversion/crossversion_test.go
@@ -226,9 +226,9 @@ type subrunResult struct {
 }
 
 type initialState struct {
-	desc    string
-	path    string
-	opsPath string
+	desc     string
+	path     string
+	opsPaths []string
 }
 
 func (s initialState) String() string {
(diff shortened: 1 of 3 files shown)


jbowens (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4887#-:-OSZkUwa3LbIOqLOvRuu:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

PR #4888, opened 2025-06-12
crl-release-25.1: crossversion: pass all previous ops

RaduBerinde:
#### metamorphic: tolerate extra arg for IngestAndExcise

Informs #4729

#### crossversion: pass all previous ops

We are passing the ops from the previous run. We need to pass ops from
all the runs that culminated in the initial state; this is necessary
for issuing correct SingleDel operations.

We change `--previous-ops` to take a list of files, separated by a
comma.


RaduBerinde:
TFTR!

PR #4891, opened 2025-06-12
manifest: remove BlobReference.OriginalMetadata

jbowens:
Previously a TableMetadata's BlobReferences maintained a pointer to the original PhysicalBlobFile that existed when the reference was created. This was confusing state and only existed for the purpose of calculating the estimated physical size of the referenced values.

This commit replaces the *PhysicalBlobFile pointer with a EstimatedPhysicalSize field that's calculated once upfront. Some of the version edit accumulation and application logic is also simplified, removing the need to track referenced *PhysicalBlobFiles.

Informs #112.

### DIFF
--- a/internal/manifest/blob_metadata.go
+++ b/internal/manifest/blob_metadata.go
@@ -32,50 +32,40 @@ type BlobReference struct {
 	// blob file for which there exists a reference in the sstable. Note that if
 	// any of the referencing tables are virtualized tables, the ValueSize may
 	// be approximate.
-	//
-	// INVARIANT: ValueSize <= Metadata.ValueSize
 	ValueSize uint64
-
-	// OriginalMetadata is the metadata for the original physical blob file. It
-	// is non-nil for blob references contained within active Versions. It is
-	// expected to initially be nil when decoding a version edit as a part of
-	// manfiest replay. When the version edit is accumulated into a bulk version
-	// edit, the metadata is populated.
-	//
-	// The OriginalMetadata describes the physical blob file that existed when
-	// the reference was originally created. The original physical blob file may
-	// no longer exist. The BlobReference.FileID should be translated using a
-	// Version's BlobFileSet.
-	//
-	// TODO(jackson): We should either remove the OriginalMetadata from the
-	// BlobReference or use it to infer when a blob file has definitely NOT been
-	// replaced and a lookup in Version.BlobFileSet is unnecessary.
-	OriginalMetadata *PhysicalBlobFile
+	// EstimatedPhysicalSize is an estimate of the physical size of the blob
+	// reference, in bytes. It's calculated by scaling the blob file's physical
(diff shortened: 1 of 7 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4891)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`tool/testdata/db_space` line 50 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4891#-OS_sQPl0EmpfwEa8Te-:-OS_sQPl0EmpfwEa8Te0:b-38qgzv) ([raw file](https://github.com/cockroachdb/pebble/blob/3747415bcf4f5fddd0df9800021c1b59f5964816/tool/testdata/db_space#L50)):*
> ```
> testdata/find-val-sep-db
> ----
> 157
> ```

I'm confused by the change in these values.

jbowens:
Looks like I broke something. Converting back to draft.

sumeerbhola (commented):
:lgtm_strong: 

> Reviewed 5 of 7 files at r1, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4891#-:-OSnsOwq8Yni745vQyZw:b-uwkowd)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`value_separation.go` line 326 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4891#-OSnquSL8rE8oXmfAkmE:-OSnquSL8rE8oXmfAkmF:bu96ulp) ([raw file](https://github.com/cockroachdb/pebble/blob/97ca0474b299bae34bb3210863952bf97535e9f2/value_separation.go#L326)):*
> ```Go
> 	// values from each blob file.
> 	//
> 	// INVARIANT: len(currReferences) == len(currReferenceValueSizes)
> ```

nit: can we define a struct and put both in the same slice instead of having to define and check an invariant?

sumeerbhola (approved):

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4891#-:-OStSDut15veVKmk1-8k:b-814g8k)* status: 6 of 8 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola)

___
*[`value_separation.go` line 326 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4891#-OSnquSL8rE8oXmfAkmE:-OStSGVO-pBytGSG-KXh:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/97ca0474b299bae34bb3210863952bf97535e9f2/value_separation.go#L326)):*

nit: can we define a struct and put both in the same slice instead of having to define and check an invariant?
</blockquote></details>

Done

PR #4892, opened 2025-06-13
metamorphic: remove obj metadata created from previous ops

RaduBerinde:
We partially "replay" previous ops. This populates the `objKeyMeta`
map for various objects from the previous run. These object IDs will
conflict with objects from a new run, so we have to clear them out. We
retain only the keys for the db itself, since that's all that carries
over from the previous run.

A symptom of this problem was
`metamorphic test internal error: external object empty`
This is because a batch that had just been created and was empty
already had an `objKeyMeta` that had bounds.

### DIFF
--- a/metamorphic/key_manager.go
+++ b/metamorphic/key_manager.go
@@ -935,6 +935,13 @@ func loadPrecedingKeys(ops []op, kg KeyGenerator, m *keyManager) {
 		// Update key tracking state.
 		m.update(op)
 	}
+	// We want to retain the keys that made it to the db and clear out everything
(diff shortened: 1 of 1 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4892#-:-OSe2WcW4VDBaOxnW2K3:bg7igk9)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #4899, opened 2025-06-13
pebble: update comments about the max batch size

kcalvinalvin:
While in the code the batch size is limited to 4GB, this wasn't stated in the documentation. We update the comments that'll be included in godoc.


kcalvinalvin:
> We could also include a note that if one needs to atomically commit a set of changes larger than 4GB, one can `Ingest` pre-constructed sstables.

Updated to include the note about `Ingest`.

jbowens:
Thanks for the contribution!

PR #4901, opened 2025-06-13
metamorphic: error out if flags are used incorrectly

### LINKED ISSUE #4845: metamorphic: error if unexpected flags are provided
There are metamorphic test flags that can only be used in conjunction with `--run/--compare`, and some flags that can only be used without `--run/--compare`. Currently, the ones that don't apply are being silently ignored. This has the potential to cause confusion. We should error out if any unexpected flags are provided.

Jira issue: PEBBLE-476

RaduBerinde:
The metamorphic test has some flags specific to the "outer" Run, and
some flags specific to `--run-dir/--compare`. If a flag is passed that
doesn't apply to the mode, it is ignored.

In this change we instead error out the test. This will reduce sources
of confusion, especially when trying to run things manually.

Examples:
```
--- FAIL: TestMeta (0.00s)
    meta_test.go:96: flag "previous-ops" is not allowed with --run-dir

--- FAIL: TestMeta (0.00s)
    meta_test.go:118: flag "try-to-reduce" is only allowed with --compare or --run-dir
```

Fixes #4845

### DIFF
--- a/internal/metamorphic/meta_test.go
+++ b/internal/metamorphic/meta_test.go
@@ -5,6 +5,7 @@
 package metamorphic
 
 import (
+	"flag"
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4901#-:-OSfCOsv8GvxlRmbJ8IV:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sumeerbhola)

RaduBerinde:
TFTR!

PR #4902, opened 2025-06-13
db: define a new value separation format major version

jbowens:
Since the original experimental value separation format major version was introduced, we've made backwards incompatible changes to value separation. In particular, value separation now requires TableFormatPebblev7 (which introduces the value liveness block that blob file replacement will make use of).

To avoid ambiguity, we define a new FormatValueSeparation that's required for enabling value separation. The old format major version becomes unexported.

Informs #112.


jbowens:
TFTRs!

PR #4903, opened 2025-06-13
block: support different setting for value blocks

RaduBerinde:
#### block: support different setting for value blocks

A possible "balanced" setting could be to use Zstd1 for value blocks
and MinLZ for everything else.


RaduBerinde:
TFTR!

Issue #4521, opened 2025-04-10
tool: support inspecting blob files


jbowens:
We should have a debug tool (`pebble/tool` pkg) for examining a blob file and its layout. In the current `valblk` format, we can't distinguish individual values but we can at least dump the layout indicated by the index block. If we convert to the colblk format, we can print individual values.

Jira issue: PEBBLE-393

PR #4906, opened 2025-06-16
tool: add tool for inspecting blob files

annrpom:
This patch adds tool support for inspecting the layout of blob files, dumping the layout indicated by the index block.

Fixes: #4521


jbowens (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4906#-:-OTmtN8lDhW2F_32JZ1_:b-uwkowd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @xxmplus)

annrpom:
i just snuck in another `sstable_layout` test for an sstable with a more interesting blob ref liveness block hence the large diff

would be nice if we could filter certain block layouts in the `sstable layout` cmd; i can look into that

TFTR! ('-')7

PR #4907, opened 2025-06-16
sstable: include all filter blocks in Reader.Layout

jbowens:
Previously, (*sstable.Reader).Layout only included 1 filter block and only if it matched the reader's configuration. This commit updates the Layout method to collect the block handles of all filter blocks contained within the metaindex block.

### DIFF
--- a/sstable/copier_test.go
+++ b/sstable/copier_test.go
@@ -30,8 +30,35 @@ func TestCopySpan(t *testing.T) {
 	defer cacheHandle.Close()
 	fileNameToNum := make(map[string]base.FileNum)
 	nextFileNum := base.FileNum(1)
-
 	keySchema := colblk.DefaultKeySchema(testkeys.Comparer, 16)
+
+	getReader := func(d *datadriven.TestData) (*Reader, error) {
+		f, err := fs.Open(d.CmdArgs[0].Key)
+		if err != nil {
+			return nil, err
+		}
+		readable, err := NewSimpleReadable(f)
+		if err != nil {
+			return nil, err
+		}
+
+		rOpts := ReaderOptions{
+			ReaderOptions: block.ReaderOptions{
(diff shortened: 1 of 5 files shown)


RaduBerinde (approved):
:lgtm:

OOC, do we ever have more than one filter block in practice?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4907#-:-OSuwupe392HXHl88dh9:b-t0hu83)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @sumeerbhola)

___
*[`sstable/layout.go` line 728 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4907#-OSuwfx7DiSjIhW9fq3e:-OSuwfx7DiSjIhW9fq3f:b8751pc) ([raw file](https://github.com/cockroachdb/pebble/blob/8e4b71f735b00aa9c1822534d9394cf3114b60d5/sstable/layout.go#L728)):*
> ```Go
> 	defer func() { err = firstError(err, i.Close()) }()
> 
> 	keysAlloc := make(bytealloc.A, 0, len(data))
> ```

This entire slice will be kept alive by the keys in the map, do we really want it to be `len(data)`?
___
*[`sstable/test_utils.go` line 264 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4907#-OSuxXiQ8rZG-QwMCuqw:-OSuxXiQ8rZG-QwMCuqx:b-rw8vok) ([raw file](https://github.com/cockroachdb/pebble/blob/8e4b71f735b00aa9c1822534d9394cf3114b60d5/sstable/test_utils.go#L264)):*
> ```Go
> 
> 		case "filter":
> 			fields := strings.FieldsFunc(value, func(r rune) bool {
> ```

Some tests need fixup, or we need to allow an empty value.

PR #4910, opened 2025-06-18
tool: analyze-data: support remote storage

RaduBerinde:
#### tool: improvements to analyze-data

We use a "real" `objstorage.ReadHandle` and set it up for readahead.

#### tool: analyze-data: support remote storage

We add support for analyzing ssts from cloud storage (backups). A tool
option is used to register a function that can take a URI and return a
`remote.Storage` implementation.

With remote storage, retrieving the size of each file one by one takes
too long when there are many files, so we don't do that upfront. Most
of the refactoring is to support tracking progress by file count only
in case of remote storage.


RaduBerinde:
TFTR!

Issue #4511, opened 2025-04-09
db: revisit pickedCompaction.grow logic


RaduBerinde:
The comment below concerns this code:
https://github.com/cockroachdb/pebble/blob/d8b52a2b396575007710f0707a6c28eea2083d0b/compaction_picker.go#L539

```
I don't understand this comment about multi-level compactions.
Also, I am uncertain whether the code here is correct for multi-level compactions. If we have levels:
L1:  [c, d)  [e,    g)
L2:  [c, d)     [f, g)
L3:  [c,         f)     [h, k)  

Then if we started with the following compaction
L1:  [c, d) 
L2:  [c, d)
L3:  [c,         f)

We would include [e, g) from L1, but won't include [f, g) from L2, which would break the level invariant.
Trying to (re)read the compaction expanding logic, it seems to have devolved into a mess when it should be fairly straightforward from an algorithmic perspective.
```
CC @sumeerbhola @jbowens 

Jira issue: PEBBLE-389

xinhaoz:
It seems that for ML compactions, we'd set up L1->L2 inputs and [then L2->L3](https://github.com/cockroachdb/pebble/blob/703ca0dd8eab3483cdd21a1e62bba3bf3ea2f42a/compaction_picker.go#L565) so I don't think this scenario could occur. I think the naming of the param makes it seem like we're using L1. Still poking around but I can see if there are any tests around to confirm and clean this up to be more straightforward.

PR #4912, opened 2025-06-18
pebble: clarify some multilevel compaction areas

xinhaoz:
This commit adds some comments to some of the existing ML compaction code. It
also removes the `extraLevels` fields from the pickedCompaction struct, since we
can read `inputs` directly.

Closes: https://github.com/cockroachdb/pebble/issues/4511

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -181,10 +181,8 @@ type pickedCompaction struct {
 	//    - in multilevel compaction, the output level is the lowest level involved in
 	//      the compaction
 	outputLevel *compactionLevel
-	// extraLevels contain additional levels in between the input and output
-	// levels that get compacted in multi level compactions
-	extraLevels []*compactionLevel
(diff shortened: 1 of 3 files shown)


compaction_picker.go
```diff
@@ -181,10 +181,8 @@ type pickedCompaction struct {
 	//    - in multilevel compaction, the output level is the lowest level involved in
 	//      the compaction
 	outputLevel *compactionLevel
-	// extraLevels contain additional levels in between the input and output
-	// levels that get compacted in multi level compactions
-	extraLevels []*compactionLevel
-	inputs      []compactionLevel
+	// Levels involved in the compaction in ascending order of depth.
```
annrpom, line 184: nit: `inputs contain levels involved in the compaction in ascending order of depth` (or something)

annrpom (approved):
:lgtm:

PR #4914, opened 2025-06-19
db: add fallback behavior if ShortAttributeExtractor errors

jbowens:
When separating a value, we call a user-provided 'ShortAttributeExtractor' to extract a few bits of data from the value. These bits, known as the 'short attribute', are stored in-place with the sstable key for fast retrieval.

Previously, if the user-provided ShortAttributeExtractor errored, the calling flush or compaction would abort and bubble the error up. If this was during a flush, the engine was unable to make progress, repeatedly attempting to flush a key that cannot be flushed.

This commit updates the logic to fallback to storing a value in-place if the short attribute extractor is unable to parse a value. The error is still surfaced through the PossibleAPIMisuse event.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -32,6 +32,7 @@ import (
 	"github.com/cockroachdb/pebble/sstable/blob"
 	"github.com/cockroachdb/pebble/sstable/block"
 	"github.com/cockroachdb/pebble/vfs"
+	"github.com/cockroachdb/redact"
 )
 
 var errEmptyTable = errors.New("pebble: empty table")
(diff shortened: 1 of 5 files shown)


sumeerbhola (approved):
:lgtm: 

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4914#-:-OTEVmVE1U8hBXDHbwB_:bg7igkv)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`event.go` line 770 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4914#-OTEVLHi4q9EneC1KyuR:-OTEVLHi4q9EneC1KyuS:b-n2wmj1) ([raw file](https://github.com/cockroachdb/pebble/blob/948902865dc443729dda847beec456bd94d834a0/event.go#L770)):*
> ```Go
> 	// ShortAttributeExtractor) returns an error for a committed value. This
> 	// suggests that either the callback is not implemented for all possible
> 	// values or a malformed value was committed to the engine.
> ```

nit: committed to the DB.
(since engine is CRDB terminology)

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4914#-:-OTEXQjlDMpdZVV6kUC2:b-814g8k)* status: 4 of 5 files reviewed, all discussions resolved (waiting on @RaduBerinde and @sumeerbhola)

___
*[`event.go` line 770 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4914#-OTEVLHi4q9EneC1KyuR:-OTEXT4N8rrse_fxah1B:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/948902865dc443729dda847beec456bd94d834a0/event.go#L770)):*

nit: committed to the DB.
(since engine is CRDB terminology)
</blockquote></details>

Done

Issue #3248, opened 2024-01-29
sstable: index block always loaded; even if unused


jbowens:
The `singleLevelIterator` and `twoLevelIterator` `init` methods both load the the highest index block. When CockroachDB is performing an MVCC Get, this may be unnecessary. If the iterator is only ever used for a call to `SeekPrefixGE` and the bloom filter is effective, the index block may never be seeked. It's unclear if this is at all impactful, and lazily loading the index block will inevitably add some small overhead and code complexity to seek operations.

Alternatively, we could implement a special `Reader.GetPrefix` method this optimization and others (#2002), without needing to edit the codepaths for other seek operations.

Jira issue: PEBBLE-116

xxmplus:
#4916 covers lazy loading in single-level iterator. The same feature in two-level iterator will come soon.

xxmplus:
It seems the lazy loading logic (lll) in single- and two-level iterators makes a few tests flaky. Taking another look while the commits are reverted.

lll in single-level: https://github.com/cockroachdb/pebble/commit/4ad0e5fbad0ad7792743cc695e2b5dc2392a89af
lll in two-level: https://github.com/cockroachdb/pebble/commit/f53515645abb563bacd662830f9eb44e598b77b5

Example failed run: https://github.com/cockroachdb/pebble/actions/runs/15969967046/job/45038765161

sumeerbhola:
> When CockroachDB is performing an MVCC Get, this may be unnecessary.

This turns out to be very unnecessary when running kv0 with random writes, which only does MVCCPut, since the key is never in the LSM so all the bloom filters miss. 49+% of the block loads from the cache in levels higher than L6 are for the index block, and 49+% are for the filter block, which is checked, the key is not found, so the data block is not loaded.

PR #4916, opened 2025-06-19
sstable: lazy load the index block in single- and two- level iterator

xxmplus:
This commit implements the index block lazy loading in single-level iterator as default behavior. The index block loading is now deferred until first access. At the same time, it leads to semantic changes as the file reading ordering changes because of this.

Here is a quick overview of the change.

Before (eager loading)
1. Iterator Construction:
    1.1 the constructor calls readTopLevelIndexBlock()
    1.2 readTopLevelIndexBlock() calls readIndexBlock()
    1.3 readIndexBlock(() calls blockReader.Read()
    1.4 blockReader.Read() calls objstorage ReadHandle ReadAt()
    1.5 ReadAt() loads the index block
    1.6 construction completes

2. First Positioning (e.g., First(), SeekGE(), SeekPrefixGE()):
    2.1 the positioning function uses the already-loaded index via PI(&i.index)
    2.2 file opened for data read
    2.3 ReadAt() loads the data block

After (lazy loading)
1. Iterator Construction:
    1.1 the constructor sets indexLoaded = false to defer loading
    1.2 construction completes (no I/O)

2. First Positioning (e.g., First(), SeekGE(), SeekPrefixGE()):
    2.1 the positioning function calls the new indexIter() wrapper
    2.2 indexIter() calls ensureIndexLoaded()
    2.3 ensureIndexLoaded() calls ReadAt()
    2.4 ReadAt() loads the index block
    2.5 uses the newly-loaded index for positioning
    2.6 ReadAt() loads the data block

As seen above, the I/O operation is now deferred. While the overall behavior is transparent to users, the internal semantic is different now. And the data-driven tests have to change accordingly. See the order of events change in file: checkpoint, cleaner, and event_listener.

What's interesting (optimized) is the change observed in flushable_ingest, which demonstrate the I/O saving thanks to the optimization.
```
-# When the key doesn't pass the bloom filter, we should see only two block
-# reads.
+# When the key doesn't pass the bloom filter, we should see only one block
+# read due to lazy loading optimization - bloom filter is checked first,
+# and if it rejects the key, we can avoid loading the index block entirely.
 get with-fs-logging
 small-00001-does-not-exist
 ----
-read-at(158, 41): 000004.sst
 read-at(199, 74): 000004.sst
```

Next is a comparison between eager loading and lazy loading. We should collect metrics on how often SeekPrefixGE() saved I/O; how often an iterator is created but not used; or how much the deferred operations improve memory usage if they are not available today (TODO).

Benchmark will be moved to a separate PR.
~~While one may argue the amortized cost is similar in the long run, lazy load could also help with perceived performance potentially. In other words, the average might be same in theory, but the common case is indeed optimized.~~

implements #3248

### DIFF
--- a/sstable/reader_iter_single_lvl.go
+++ b/sstable/reader_iter_single_lvl.go
@@ -165,6 +165,9 @@ type singleLevelIterator[I any, PI indexBlockIterator[I], D any, PD dataBlockIte
 	useFilterBlock         bool
 	lastBloomFilterMatched bool
 
+	// Lazy loading flag
+	indexLoaded bool
+
 	transforms IterTransforms
 
 	// All fields above this field are cleared when resetting the iterator for reuse.
@@ -214,14 +217,9 @@ func newColumnBlockSingleLevelIterator(
 		i.vbRH = r.blockReader.UsePreallocatedReadHandle(objstorage.NoReadBefore, &i.vbRHPrealloc)
 	}
 	i.data.InitOnce(r.keySchema, r.Comparer, &i.internalValueConstructor)
-	indexH, err := r.readTopLevelIndexBlock(ctx, i.readEnv.Block, i.indexFilterRH)
-	if err == nil {
-		err = i.index.InitHandle(r.Comparer, indexH, opts.Transforms)
-	}
-	if err != nil {
-		_ = i.Close()
-		return nil, err
-	}
+
+	// Use lazy loading by default - index will be loaded on first access
+	i.indexLoaded = false
 	return i, nil
 }
 
@@ -254,14 +252,8 @@ func newRowBlockSingleLevelIterator(
 		i.data.SetHasValuePrefix(true)
 	}
 
-	indexH, err := r.readTopLevelIndexBlock(ctx, i.readEnv.Block, i.indexFilterRH)
-	if err == nil {
-		err = i.index.InitHandle(r.Comparer, indexH, opts.Transforms)
-	}
-	if err != nil {
-		_ = i.Close()
-		return nil, err
-	}
+	// Use lazy loading by default - index will be loaded on first access
+	i.indexLoaded = false
 	return i, nil
 }
 
@@ -449,7 +441,7 @@ func (i *singleLevelIterator[I, PI, P, PD]) SetContext(ctx context.Context) {
 // unpositioned. If unsuccessful, it sets i.err to any error encountered, which
 // may be nil if we have simply exhausted the entire table.
 func (i *singleLevelIterator[I, PI, P, PD]) loadDataBlock(dir int8) loadBlockResult {
-	if !PI(&i.index).Valid() {
+	if i.err != nil || !PI(&i.index).Valid() {
 		// Ensure the data block iterator is invalidated even if loading of the block
 		// fails.
 		PD(&i.data).Invalidate()
@@ -521,6 +513,12 @@ func (i *singleLevelIterator[I, PI, D, PD]) ReadValueBlock(
 // apprioriate bound, depending on the iteration direction, and returns either
 // `blockIntersects` or `blockExcluded`.
 func (i *singleLevelIterator[I, PI, D, PD]) resolveMaybeExcluded(dir int8) intersectsResult {
+	if !i.indexLoaded {
+		if err := i.ensureIndexLoaded(); err != nil {
+			i.err = err
+			return blockExcluded
+		}
+	}
 	// TODO(jackson): We could first try comparing to top-level index block's
 	// key, and if within bounds avoid per-data block key comparisons.
 
@@ -682,6 +680,13 @@ func (i *singleLevelIterator[I, PI, D, PD]) SeekGE(
 func (i *singleLevelIterator[I, PI, D, PD]) seekGEHelper(
 	key []byte, boundsCmp int, flags base.SeekGEFlags,
 ) *base.InternalKV {
+	if !i.indexLoaded {
+		if err := i.ensureIndexLoaded(); err != nil {
+			i.err = err
+			return nil
+		}
+	}
+
 	// Invariant: trySeekUsingNext => !i.data.isDataInvalidated() && i.exhaustedBounds != +1
 
 	// SeekGE performs various step-instead-of-seeking optimizations: eg enabled
@@ -818,6 +823,7 @@ func (i *singleLevelIterator[I, PI, D, PD]) seekPrefixGE(
 			flags = flags.DisableTrySeekUsingNext()
 		}
 		i.lastBloomFilterMatched = false
+
 		// Check prefix bloom filter.
 		var mayContain bool
 		mayContain, i.err = i.bloomFilterMayContain(prefix)
@@ -922,6 +928,13 @@ func (i *singleLevelIterator[I, PI, D, PD]) virtualLastSeekLE() *base.InternalKV
 	i.boundsCmp = 0
 	i.positionedUsingLatestBounds = true
 
+	if !i.indexLoaded {
+		if err := i.ensureIndexLoaded(); err != nil {
+			i.err = err
+			return nil
+		}
+	}
+
 	indexOk := PI(&i.index).SeekGE(key)
 	// We can have multiple internal keys with the same user key as the seek
 	// key. In that case, we want the last (greatest) internal key.
@@ -993,6 +1006,12 @@ func (i *singleLevelIterator[I, PI, D, PD]) virtualLastSeekLE() *base.InternalKV
 func (i *singleLevelIterator[I, PI, D, PD]) SeekLT(
 	key []byte, flags base.SeekLTFlags,
 ) *base.InternalKV {
+	if !i.indexLoaded {
+		if err := i.ensureIndexLoaded(); err != nil {
+			i.err = err
+			return nil
+		}
+	}
 	if i.readEnv.Virtual != nil {
 		// Might have to fix upper bound since virtual sstable bounds are not
 		// known to callers of SeekLT.
@@ -1116,6 +1135,12 @@ func (i *singleLevelIterator[I, PI, D, PD]) First() *base.InternalKV {
 // index file. For the latter, one cannot make any claims about absolute
 // positioning.
 func (i *singleLevelIterator[I, PI, D, PD]) firstInternal() *base.InternalKV {
+	if !i.indexLoaded {
+		if err := i.ensureIndexLoaded(); err != nil {
+			i.err = err
+			return nil
+		}
+	}
 	i.exhaustedBounds = 0
 	i.err = nil // clear cached iteration error
 	// Seek optimization only applies until iterator is first positioned after SetBounds.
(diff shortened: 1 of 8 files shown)


RaduBerinde (commented):
I'd expect the (constructor + first access) time to be similar, any ideas why it's so much larger after the change?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4916#-:-OTEoeZZ2S-CvhOc3Wvv:b-s22a8)* status: 0 of 10 files reviewed, 5 unresolved discussions (waiting on @jbowens)

___
*[`sstable/simple_benchmark_test.go` line 93 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4916#-OTEpa8i61k1CTXlCDPD:-OTEpa8i61k1CTXlCDPE:b-c19tnw) ([raw file](https://github.com/cockroachdb/pebble/blob/f203b4437d55590547652ea657ddaf7821bfcec8/sstable/simple_benchmark_test.go#L93)):*
> ```Go
> // BenchmarkConstructionPlusFirstAccess measures construction + first index access
> // How to: go test -bench=BenchmarkConstructionPlusFirstAccess -run=^$ ./sstable
> func BenchmarkConstructionPlusFirstAccess(b *testing.B) {
> ```

[nit] all these benchmarks should start with `BenchmarkIterator...`
___
*[`sstable/comparison_benchmark_test.go` line 44 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4916#-OTEs5OA9Mfb1wYgrcMz:-OTEs5OA9Mfb1wYgrcN-:b-c2vkq9) ([raw file](https://github.com/cockroachdb/pebble/blob/f203b4437d55590547652ea657ddaf7821bfcec8/sstable/comparison_benchmark_test.go#L44)):*
> ```Go
> 	})
> 
> 	b.Run("SimulatedEagerLoading_ConstructOnly", func(b *testing.B) {
> ```

I don't understand all these benchmark variants (and the three files). We should have just:
 - construction
 - construction+first
 - construction+seekge
 - construction+seek-prefix-ge with no hit in filter
 - maybe construction+seek-prefix-ge with existing key

They should be in a different commit so we can easily obtain and show before and after values.
___
*[`sstable/reader_iter_single_lvl.go` line 173 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4916#-OTEq45c8o7SWVta0Xcp:-OTEq45c8o7SWVta0Xcq:bgu7be) ([raw file](https://github.com/cockroachdb/pebble/blob/f203b4437d55590547652ea657ddaf7821bfcec8/sstable/reader_iter_single_lvl.go#L173)):*
> ```Go
> 	clearForResetBoundary struct{}
> 
> 	// Lazy loading flag - must be below clearForResetBoundary to survive pool reuse
> ```

Why does this need to survive pool reuse?

In any case, we should put it next to another bool so it doesn't unnecessarily increase the struct size by a word
___
*[`sstable/reader_iter_single_lvl.go` line 451 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4916#-OTEqpRz0fVN75m7H56Y:-OTEqpRz0fVN75m7H56Z:bwtp4i) ([raw file](https://github.com/cockroachdb/pebble/blob/f203b4437d55590547652ea657ddaf7821bfcec8/sstable/reader_iter_single_lvl.go#L451)):*
> ```Go
> 	}
> 	// Load the next block.
> 	bhp, err := i.indexIter().BlockHandleWithProperties()
> ```

We don't need to keep calling the function each time..
___
*[`sstable/reader_iter_single_lvl.go` line 1585 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4916#-OTEqY9h60ton1VNxPq7:-OTEqY9h60ton1VNxPq8:b2yqrv5) ([raw file](https://github.com/cockroachdb/pebble/blob/f203b4437d55590547652ea657ddaf7821bfcec8/sstable/reader_iter_single_lvl.go#L1585)):*
> ```Go
> 
> func (i *singleLevelIterator[I, PI, D, PD]) indexIter() PI {
> 	if err := i.ensureIndexLoaded(); err != nil {
> ```

[nit] since it's a hot path, I'd move the `if i.indexLoaded` here, in case `ensureIndexLoaded()` doesn't get inlined.

xxmplus (commented):
Let me address that in the separate PR after split.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4916#-:-OTTYRJOADy8mwxQCMAj:bm7xkno)* status: 0 of 10 files reviewed, 5 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 173 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4916#-OTEq45c8o7SWVta0Xcp:-OTT9tkZ1uON77ZlEK2f:byx5pdw) ([raw file](https://github.com/cockroachdb/pebble/blob/f203b4437d55590547652ea657ddaf7821bfcec8/sstable/reader_iter_single_lvl.go#L173)):*

Why does this need to survive pool reuse?

In any case, we should put it next to another bool so it doesn't unnecessarily increase the struct size by a word
</blockquote></details>

You are right. This is unnecessary. Moved it above clearForResetBoundary.
___
*[`sstable/reader_iter_single_lvl.go` line 451 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4916#-OTEqpRz0fVN75m7H56Y:-OTTLmso6xoRzb9GARqy:bo4c8mg) ([raw file](https://github.com/cockroachdb/pebble/blob/f203b4437d55590547652ea657ddaf7821bfcec8/sstable/reader_iter_single_lvl.go#L451)):*

We don't need to keep calling the function each time..
</blockquote></details>

This is existing code replaced with the indexIter() wrapper. Test would fail if use PI(&i.index).
___
*[`sstable/reader_iter_single_lvl.go` line 1585 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4916#-OTEqY9h60ton1VNxPq7:-OTTStZnByG9kq_NfOwg:bgreapz) ([raw file](https://github.com/cockroachdb/pebble/blob/f203b4437d55590547652ea657ddaf7821bfcec8/sstable/reader_iter_single_lvl.go#L1585)):*

[nit] since it's a hot path, I'd move the `if i.indexLoaded` here, in case `ensureIndexLoaded()` doesn't get inlined.
</blockquote></details>

updated. I'm always under the impression that the compiler is smart enough these days, anyway :p
___
*[`sstable/comparison_benchmark_test.go` line 44 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4916#-OTEs5OA9Mfb1wYgrcMz:-OTTV-kR1MzX2-C6tlPG:b50wzd8) ([raw file](https://github.com/cockroachdb/pebble/blob/f203b4437d55590547652ea657ddaf7821bfcec8/sstable/comparison_benchmark_test.go#L44)):*

I don't understand all these benchmark variants (and the three files). We should have just:
 - construction
 - construction+first
 - construction+seekge
 - construction+seek-prefix-ge with no hit in filter
 - maybe construction+seek-prefix-ge with existing key

They should be in a different commit so we can easily obtain and show before and after values.
</blockquote></details>

hmmm, ok. I will split the PR.
___
*[`sstable/simple_benchmark_test.go` line 93 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4916#-OTEpa8i61k1CTXlCDPD:-OTTUzL_EFkH-C3e6Fef:b2nhysk) ([raw file](https://github.com/cockroachdb/pebble/blob/f203b4437d55590547652ea657ddaf7821bfcec8/sstable/simple_benchmark_test.go#L93)):*

[nit] all these benchmarks should start with `BenchmarkIterator...`
</blockquote></details>

ack

xxmplus (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4916)* status: 0 of 10 files reviewed, 6 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 1582 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4916#-OTTfB637R0gPsz8cM1X:-OTZT7jJDqtFSG1TDm1U:bo64xja) ([raw file](https://github.com/cockroachdb/pebble/blob/86145b61c9e347a08b13c5743df14ae4c223f870/sstable/reader_iter_single_lvl.go#L1582)):*
> Is it always safe to use the `PI(&i.index)` returned value even if the load failed?

It is not. This is indeed a good catch! We should always check for nil to be safe. Let me fix that in the next revision. Btw, when nil is detected, say, before calling `i.indexIter().BlockHandleWithProperties()`, what is the correct behavior? Is returning nil fine? I assume panicking is not ok, right?

> existing code might be modifying `i.err` with the assumption that it was nil before.

I don't see existing code check `i.err` before setting it, what's the correct handling behavior?

> I think it would be cleaner to just call `ensureIndexLoaded()` and return any error at the beginning of every function that needs it, and then proceed to use `PI(&i.index)` as usual.

I'm afraid it'd defeat the purpose of lazy loading. We do want to delay index loading until it's actually required. Also, all reference to `PI(&i.index)` is replaced by `indexIter()`. If we mixes the usage of them, it will be really confusing in the future.

RaduBerinde:
*[`sstable/reader_iter_single_lvl.go` line 1582 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4916#-OTTfB637R0gPsz8cM1X:-OTZnU8z1QLquZOpDow7:b-x7n58c) ([raw file](https://github.com/cockroachdb/pebble/blob/86145b61c9e347a08b13c5743df14ae4c223f870/sstable/reader_iter_single_lvl.go#L1582)):*

> Is it always safe to use the `PI(&i.index)` returned value even if the load failed?

It is not. This is indeed a good catch! We should always check for nil to be safe. Let me fix that in the next revision. Btw, when nil is detected, say, before calling `i.indexIter().BlockHandleWithProperties()`, what is the correct behavior? Is returning nil fine? I assume panicking is not ok, right?

> existing code might be modifying `i.err` with the assumption that it was nil before.

I don't see existing code check `i.err` before setting it, what's the correct handling behavior?

> I think it would be cleaner to just call `ensureIndexLoaded()` and return any error at the beginning of every function that needs it, and then proceed to use `PI(&i.index)` as usual.

I'm afraid it'd defeat the purpose of lazy loading. We do want to delay index loading until it's actually required. Also, all reference to `PI(&i.index)` is replaced by `indexIter()`. If we mixes the usage of them, it will be really confusing in the future.
</blockquote></details>

Of course we would call it later (after the bloom filter check) in the one method where it matters (SeekPrefixGE).  Methods need to return a nil KV on errors. We can change indexIter() to assert (when invariants.Enabled) that the filter is loaded, and use a separate error-returning ensureIndexLoaded() to do the load.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4916#-:-OTmrUj6EuKgFhOP3wmb:bnfp4nl)* status: 0 of 11 files reviewed, 5 unresolved discussions (waiting on @jbowens)

PR #4917, opened 2025-06-20
db: fix Options.Clone()

RaduBerinde:
`Clone()` was allowing a shallow copy of `*WALFailover`. This caused a
lot of time debugging a test which unwittingly aliased the WAL
failover FS and dir of two stores.

This commit fixes this and adds a test that would have caught it.

We should probably make `WALFailover` a non-pointer (with an `Enabled`
field); even if we do, the new test is still valuable.

### DIFF
--- a/external_test.go
+++ b/external_test.go
@@ -7,9 +7,12 @@ package pebble_test
 import (
 	"bytes"
 	"io"
+	randv1 "math/rand"
 	"math/rand/v2"
+	"reflect"
 	"strings"
 	"testing"
+	"testing/quick"
 	"time"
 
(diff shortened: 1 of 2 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4917#-:-OTChCO46bmpXf0vnwbC:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #4921, opened 2025-06-20
db: fix incorrect cleanup of initial WALs

RaduBerinde:
#### db: add TestDoubleRestart

Add a test that performs two restarts in quick sequence and checks the
KVs are correct. Reproduces the bug in cockroachdb/cockroach#148419

We improve `RandomOptions` to allow a bit more control on what
features we enable.

#### db: fix incorrect cleanup of initial WALs

We were incorrectly deleting all initial WALs after the first flush
completes. This is a problem when we have multiple flushes to perform,
which can happen if the WALs contain a batch that doesn't fit in the
memtable.

Informs cockroachdb/cockroach#148419

### DIFF
--- a/external_test.go
+++ b/external_test.go
@@ -10,6 +10,7 @@ import (
 	randv1 "math/rand"
 	"math/rand/v2"
 	"reflect"
+	"slices"
 	"strings"
 	"testing"
 	"testing/quick"
@@ -18,6 +19,7 @@ import (
 	"github.com/cockroachdb/pebble"
 	"github.com/cockroachdb/pebble/bloom"
 	"github.com/cockroachdb/pebble/cockroachkvs"
+	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/metamorphic"
 	"github.com/cockroachdb/pebble/sstable"
 	"github.com/cockroachdb/pebble/vfs"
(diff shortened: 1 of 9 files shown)


RaduBerinde:
Only if we had batches in the WAL larger than the memtable size.. Perhaps there are other ways in which the bug could manifest. It was pretty hard to get a test that reproduces it though.

jbowens (approved):
:lgtm_strong: 

nice test

> Reviewed 4 of 7 files at r1, 5 of 5 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4921#-:-OTXZnNx7hyrmkR50vaG:b-iin6vy)* status: all files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`external_test.go` line 260 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4921#-OTXYbTh7olUQQAGadgY:-OTXYbTh7olUQQAGadgZ:b-ifcjo2) ([raw file](https://github.com/cockroachdb/pebble/blob/1dd1be585ecb80b563de169e5cbe8a991ee6e8e4/external_test.go#L260)):*
> ```Go
> // flushes are complete. Steps:
> //
> //  1. Use metamorphic to run opretaions on a store; the resulting filesystem is
> ```

nit: "operations"
___
*[`external_test.go` line 330 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4921#-OTXZ2GK5-g68fYmxrf8:-OTXZ2GK5-g68fYmxrf9:b-qen84y) ([raw file](https://github.com/cockroachdb/pebble/blob/1dd1be585ecb80b563de169e5cbe8a991ee6e8e4/external_test.go#L330)):*
> ```Go
> 	}
> 
> 	// Open the "golden" filesystem and scan it to get the expected VKs.
> ```

nit: KVs

Issue #4807, opened 2025-06-02
db: implement blob file replacement


jbowens:
Implement blob file replacement with a simple configurable heuristic.

Jira issue: PEBBLE-466

Epic CRDB-20379

annrpom:
Fixed by https://github.com/cockroachdb/pebble/pull/4922, https://github.com/cockroachdb/pebble/pull/4908, https://github.com/cockroachdb/pebble/pull/4940

PR #4922, opened 2025-06-20
manifest: add blob file rewrite heuristic

jbowens:
Adapt CurrentBlobFileSet to maintain state in support of picking blob files for
rewrite. Specifically, the CurrentBlobFileSet is extended with two heaps. One
heap holds a set of files that are not fully referenced but are considered
'recently created,' ordered within the heap by the blob file creation time. The
other heap holds files that are not fully referenced, ordered within the heap
by the fraction of the file's data that is referenced.

These heaps will be used during compaction picking to pick blob files to
rewrite when value separation space amplification becomes intolerably high.

Informs https://github.com/cockroachdb/pebble/issues/4807.

### DIFF
--- a/cmd/pebble/db.go
+++ b/cmd/pebble/db.go
@@ -6,6 +6,7 @@ package main
 
 import (
 	"log"
+	"time"
 
 	"github.com/cockroachdb/pebble"
 	"github.com/cockroachdb/pebble/bloom"
@@ -86,6 +87,7 @@ func newPebbleDB(dir string) DB {
 			Enabled:               true,
 			MinimumSize:           512,
 			MaxBlobReferenceDepth: 10,
+			RewriteMinimumAge:     5 * time.Minute,
 		}
 	}
 
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -64,7 +64,9 @@ func newVersionWithLatest(
 		l0Organizer:     l0Organizer,
 		virtualBackings: manifest.MakeVirtualBackings(),
 	}
-	latest.blobFiles.Init(nil)
+	latest.blobFiles.Init(nil, manifest.BlobRewriteHeuristic{
+		CurrentTime: time.Now,
+	})
 	return v, latest
 }
 
--- a/db.go
+++ b/db.go
@@ -2168,7 +2168,7 @@ func (d *DB) Metrics() *Metrics {
 	backingCount, backingTotalSize := d.mu.versions.latest.virtualBackings.Stats()
 	metrics.Table.BackingTableCount = uint64(backingCount)
 	metrics.Table.BackingTableSize = backingTotalSize
-	blobStats := d.mu.versions.latest.blobFiles.Stats()
+	blobStats, _ := d.mu.versions.latest.blobFiles.Stats()
 	d.mu.versions.logUnlock()
 	metrics.BlobFiles.LiveCount = blobStats.Count
 	metrics.BlobFiles.LiveSize = blobStats.PhysicalSize
--- a/external_test.go
+++ b/external_test.go
@@ -199,6 +199,7 @@ func buildSeparatedValuesDB(
 			Enabled:               true,
 			MinimumSize:           50,
 			MaxBlobReferenceDepth: 10,
+			RewriteMinimumAge:     15 * time.Minute,
 		}
 	}
 	o.Levels[0].BlockSize = 32 << 10       // 32 KB
--- a/internal/manifest/blob_metadata.go
+++ b/internal/manifest/blob_metadata.go
@@ -6,10 +6,13 @@ package manifest
 
 import (
 	stdcmp "cmp"
+	"container/heap"
 	"fmt"
 	"iter"
 	"slices"
+	"strings"
 	"sync/atomic"
+	"time"
 
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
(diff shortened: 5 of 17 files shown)


RaduBerinde (approved):
:lgtm: Nice!

Seems like we need a generic heap in `crlib` :smile: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4922#-:-OTEnC8N5E8ygyxzFr3w:b-b82f9e)* status: 0 of 17 files reviewed, 5 unresolved discussions (waiting on @xinhaoz)

___
*[`internal/manifest/blob_metadata.go` line 468 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4922#-OTElMnw2u-cy6tJWmaa:-OTElMnx8gSdOdBg2V54:b-1i3c5k) ([raw file](https://github.com/cockroachdb/pebble/blob/f32f164bf128b7e94a44fb30031fc4890df774f6/internal/manifest/blob_metadata.go#L468)):*
> ```Go
> 		// space reclaimed per byte written.
> 		candidates currentBlobFileHeap[byReferencedRatio]
> 		// recentlyCreated holds a set of blob files that are not yet candidates
> ```

They also must contain some unreferenced values, right?
___
*[`internal/manifest/blob_metadata.go` line 528 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4922#-OTElzRD-ZwtASFNwMrg:-OTElzRD-ZwtASFNwMrh:b-r5pqfu) ([raw file](https://github.com/cockroachdb/pebble/blob/f32f164bf128b7e94a44fb30031fc4890df774f6/internal/manifest/blob_metadata.go#L528)):*
> ```Go
> type BlobRewriteHeuristic struct {
> 	// CurrentTimeSecs returns the current time in seconds since the unix epoch.
> 	CurrentTimeSecs func() uint64
> ```

Can this return `time.Time`, or introduce a special type? It would be good to have the compiler check that we e.g. never directly compare a time with a duration.

Ideally, we'd be using `time.Time` everywhere except at the manifest encode/decode level.. That way we would also get the benefit of the monotonic timer.
___
*[`internal/manifest/blob_metadata.go` line 536 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4922#-OTElrJb0KkgE9xhJYu-:-OTElrJb0KkgE9xhJYu0:b-f45ojr) ([raw file](https://github.com/cockroachdb/pebble/blob/f32f164bf128b7e94a44fb30031fc4890df774f6/internal/manifest/blob_metadata.go#L536)):*
> ```Go
> 	// heap. Raising the value is more complex: we would need to iterate over
> 	// all the blob files in the candidates heap.
> 	MinimumAgeSecs uint64
> ```

Can this be a time.Duration?
___
*[`options.go` line 1175 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4922#-OTEj53k6igJUwtKMe6q:-OTEj53k6igJUwtKMe6r:b-kmshc1) ([raw file](https://github.com/cockroachdb/pebble/blob/f32f164bf128b7e94a44fb30031fc4890df774f6/options.go#L1175)):*
> ```Go
> 	// values into new blob files.
> 	MaxBlobReferenceDepth int
> 	// RewriteMinimumAge is the minimum age of a blob file that is eligible to
> ```

[nit] "RewriteMinimumAge specifies how old a blob file must be in order for it to be eligible for a rewrite that reclaims disk space. Lower values reduce space amplification at the cost of write amplification".

Also, mention the default (for the ones above as well).
___
*[`open.go` line 307 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4922#-OTEis521_tJFIfqO37G:-OTEis535Gmb_aQ38kba:bal6nqr) ([raw file](https://github.com/cockroachdb/pebble/blob/f32f164bf128b7e94a44fb30031fc4890df774f6/open.go#L307)):*
> ```Go
> 		return uint64(d.timeNow().UnixNano() / int64(time.Second))
> 	}
> 	if opts.Experimental.ValueSeparationPolicy != nil {
> ```

[nit] We should make `EnsureDefaults` always initialize it like other fields.

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4922#-:-OTIj8o--e8rySOpxeht:bodgj64)* status: 0 of 18 files reviewed, all discussions resolved (waiting on @xinhaoz)

___
*[`open.go` line 307 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4922#-OTEis521_tJFIfqO37G:-OTIUvOiA3KinPc4Tbhe:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/f32f164bf128b7e94a44fb30031fc4890df774f6/open.go#L307)):*

[nit] We should make `EnsureDefaults` always initialize it like other fields.
</blockquote></details>

Done
___
*[`options.go` line 1175 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4922#-OTEj53k6igJUwtKMe6q:-OTIanmFExl2Hllh1kuh:b85pxx4) ([raw file](https://github.com/cockroachdb/pebble/blob/f32f164bf128b7e94a44fb30031fc4890df774f6/options.go#L1175)):*

[nit] "RewriteMinimumAge specifies how old a blob file must be in order for it to be eligible for a rewrite that reclaims disk space. Lower values reduce space amplification at the cost of write amplification".

Also, mention the default (for the ones above as well).
</blockquote></details>

Done.

There's no default, since it's disabled by default. `Options.Experimental.ValueSeparation` documents that it defaults to disabled. I included some documentation of the permitted bounds (and added validation to Options.Validate).
___
*[`internal/manifest/blob_metadata.go` line 468 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4922#-OTElMnw2u-cy6tJWmaa:-OTIbYK93mA9QdZMEts2:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/f32f164bf128b7e94a44fb30031fc4890df774f6/internal/manifest/blob_metadata.go#L468)):*

They also must contain some unreferenced values, right?
</blockquote></details>

Done.
___
*[`internal/manifest/blob_metadata.go` line 528 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4922#-OTElzRD-ZwtASFNwMrg:-OTIj6fs0tXi1SEQZlBy:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/f32f164bf128b7e94a44fb30031fc4890df774f6/internal/manifest/blob_metadata.go#L528)):*

Can this return `time.Time`, or introduce a special type? It would be good to have the compiler check that we e.g. never directly compare a time with a duration.

Ideally, we'd be using `time.Time` everywhere except at the manifest encode/decode level.. That way we would also get the benefit of the monotonic timer.
</blockquote></details>

Done
___
*[`internal/manifest/blob_metadata.go` line 536 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4922#-OTElrJb0KkgE9xhJYu-:-OTIj7MF3wo2D3ELc7bk:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/f32f164bf128b7e94a44fb30031fc4890df774f6/internal/manifest/blob_metadata.go#L536)):*

Can this be a time.Duration?
</blockquote></details>

Done

PR #4926, opened 2025-06-21
db: add SpanPolicy flag to prefer faster compression

RaduBerinde:
For the same reasons we don't want value separation for some of the
local keys, we want to use the faster compression since these keys
shouldn't take up a significant amount of space anyway.

More generally, we probably want to use the fastest compression for
all local keys.


RaduBerinde:
TFTR! By the way, why do we disable value separation just for lock keys? Don't we want it for e.g. raft logs too? Oh or I guess read latency is not critical for those.

jbowens:
> Oh or I guess read latency is not critical for those.

Yeah, we rarely read the raft logs. They also make up ~1/2 of all writes, so @sumeerbhola and I were discussing setting a span policy to always separate them to maximize the write-amp savings.

RaduBerinde:
Nice! We have the technology!

PR #4928, opened 2025-06-21
crossversion: don't parallelize calls to the meta test

RaduBerinde:
The meta test itself already runs many configs in parallel. Running
one meta test per core is overkill and can lead to a lot more memory
usage (we have seen OOMs in CI).

I verified that a crossversion run still saturates all 10 cores on my
machine.


RaduBerinde:
TFTR!

PR #4929, opened 2025-06-22
db: move TargetFileSize out of LevelOptions

RaduBerinde:
`LevelOptions.TargetFileSize` is used in a way that is not consistent
with the rest of the level options. Whereas other level options apply
to the expected level (e.g. `Levels[1]` applies to `L1`), the target
file size applies to the level relative to Lbase. Specifically for
i > 0, `Levels[i].TargetFileSize` is the target file size of Lbase+i-1.

We move this field to its own separate array and add a convenience
`TargetFileSize()` method.

We don't change the options encoding.


RaduBerinde:
TFTR!

PR #4933, opened 2025-06-22
db: clean up compaction, pickedCompaction structs

jbowens:
These commits attempt to organize and comment the compaction and pickedCompaction structs. This is motivated by blob-file rewrite 'compactions' which share little of the same state as compactions but need to be considered similarly in some instances (eg, concurrent compaction limiting).

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -198,14 +198,11 @@ type compaction struct {
 	// compactionKindRewrite.
 	isDownload bool
 
-	cmp       Compare
-	equal     Equal
-	comparer  *base.Comparer
-	formatKey base.FormatKey
-	logger    Logger
-	version   *manifest.Version
-	stats     base.InternalIteratorStats
-	beganAt   time.Time
+	comparer *base.Comparer
+	logger   Logger
+	version  *manifest.Version
+	stats    base.InternalIteratorStats
+	beganAt  time.Time
 	// versionEditApplied is set to true when a compaction has completed and the
 	// resulting version has been installed (if successful), but the compaction
 	// goroutine is still cleaning up (eg, deleting obsolete files).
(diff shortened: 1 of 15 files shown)


RaduBerinde (approved):
:lgtm_strong:  Nice!

> Reviewed 8 of 13 files at r3, 3 of 5 files at r4, 2 of 3 files at r6, 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4933#-:-OTNT4TX1GRpr1itBh5Y:b-t3mbzd)* status: all files reviewed, 1 unresolved discussion (waiting on @xinhaoz)

___
*[`internal/base/key_bounds.go` line 236 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4933#-OTNRIZM2B6_QhvBLqKm:-OTNRIZM2B6_QhvBLqKn:bshp32s) ([raw file](https://github.com/cockroachdb/pebble/blob/e4416dc9a3a6238216cb66b677cf44846bfe5124/internal/base/key_bounds.go#L236)):*
> ```Go
> func (b *UserKeyBounds) Union(cmp Compare, other UserKeyBounds) UserKeyBounds {
> 	union := *b
> 	if other.Start != nil && (union.Start == nil || cmp(union.Start, other.Start) > 0) {
> ```

[nit] Do we need to tolerate `other` having empty keys? I'd rather not allow this since other methods don't support it. Maybe just have a special case where if `*b == (UserKeyBounds{})` then we set to `other`.

We could also consider a `UserKeyBoundUnion` helper instead of this method:
```
var u UserKeyBoundUnion
u.Add(bound1)
u.Add(bound2)
u.Get()  // panics if Add was not called
```

Then we won't have a case in the API where an unset `UserKeyBounds` is valid in some way.

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4933#-:-OTOGH_C2dNSFJBCNAX1:bnfp4nl)* status: 0 of 15 files reviewed, all discussions resolved

___
*[`internal/base/key_bounds.go` line 236 at r3](https://reviewable.io/reviews/cockroachdb/pebble/4933#-OTNRIZM2B6_QhvBLqKm:-OTOGF5l2bphBbsITnXT:b-83sjoh) ([raw file](https://github.com/cockroachdb/pebble/blob/e4416dc9a3a6238216cb66b677cf44846bfe5124/internal/base/key_bounds.go#L236)):*

I added a `manifest.ExtendKeyRange` func that allows us to remove support for the passed in bounds being empty.
</blockquote></details>

Nice!

PR #4936, opened 2025-06-23
pebble: setupInputs checks key range conflicts with in progress compa…

xinhaoz:
…ctions

Previously, checking if a compaction had conflicts with an ongoing compaction was done in `setupInputs`, where we check if any of the input files are already marked as compacting, and also `inputRangeAlreadyCompacting`, where we check if the compaction range chosen overlaps with an ongoing compaction's range.

`inputRangeAlreadyCompacting` was typically called right after `setupInputs`. This commit consolidates `inputRangeAlreadyCompacting` into `setupInputs`.

Epic: none

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -339,7 +339,6 @@ func newPickedCompactionFromL0(
 
 	pc := newPickedTableCompaction(opts, vers, l0Organizer, 0, outputLevel, baseLevel)
 	pc.lcf = lcf
-	pc.outputLevel.level = outputLevel
 
 	// Manually build the compaction as opposed to calling
 	// pickAutoHelper. This is because L0Sublevels has already added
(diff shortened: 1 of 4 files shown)


jbowens (approved):
:lgtm:

> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4936#-:-OTmvBzW0g5jgk7VSqxv:bnfp4nl)* status: 3 of 4 files reviewed, 1 unresolved discussion

___
*[`testdata/compaction_setup_inputs_multilevel_write_amp` line 85 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4936#-OTmuxc_8cuB4DdgX2Xj:-OTmuxc_8cuB4DdgX2Xk:b1x6107) ([raw file](https://github.com/cockroachdb/pebble/blob/a4f8ad83f55b1b46ec71b02ba5da267571172371/testdata/compaction_setup_inputs_multilevel_write_amp#L85)):*
> ```
>   e.SET.3-f.SET.2 size=100
> ----
> malformed table spec: L2 max-size=5. usage: <optional-file-num>:start.SET.1-end.SET.2
> ```

looks like there's a parsing bug for this case where we have a `max-size=` setting on the level line.

PR #4938, opened 2025-06-23
sstable: fix bug in serializing blob reference index

jbowens:
Fix a bug in serializing the blob reference index that resulted in the per-reference encodings being omitted.


annrpom (approved):
T_T

jbowens:
TFTR!!

PR #4940, opened 2025-06-24
db: add blob file rewrite compaction

jbowens:
Introduce a new compaction kind: blob-file rewrite compactions. These compactions pick a physical blob file with unreferenced values, rewrite it and output a new physical blob file. This commit adds the compaction and compaction picking mechanics without the sophisticated blob file rewrite. For now these rewrite compactions only copy the blob file verbatim.

Informs #112.


jbowens:
TFTR!!

PR #4941, opened 2025-06-24
tool/testdata: regen val-sep db fixture

annrpom:
These fixtures now include the blob reference index block fixes merged recently.


annrpom:
TFTR! ('-')7

Issue #4889, opened 2025-06-12
tool: add support for inspecting sstable's blob value liveness index


jbowens:
The new blob value liveness index block in an sstable indicates which values in external blob files a sstable references. We should support inspecting the block with the pebble debug tools.

Jira issue: PEBBLE-483

PR #4943, opened 2025-06-24
tool: add support for inspecting sstable's blob value liveness index

annrpom:
The new blob value liveness index block in an sstable indicates which values in external blob files a sstable references. This commit adds support for inspecting the block within `sstable layout`.

Fixes: #4889


annrpom:
TFTR! ('-')7

Issue #4797, opened 2025-05-30
db: adapt compression stats to be per-block


jbowens:
We collect counts of sstables that use each of the various compression algorithms. Today within a sstable, some meta blocks are never compressed and some data blocks may not be compressed if compression was ineffective. With adaptive compression, data blocks within a sstable may be compressed with different algorithms. We should adapt our compression stats to be in terms of blocks to capture this granularity.

Jira issue: PEBBLE-462

RaduBerinde:
We want statistics for total compressed/decompressed byte size of each algorithm, and per-table stats of the "profile" used for the table (which includes knobs for any adaptive algorithm).

RaduBerinde:
We now have stats in the sstables. This issue still tracks adding Pebble metrics that expose them.

RaduBerinde:
Done in https://github.com/cockroachdb/pebble/pull/5133

PR #4945, opened 2025-06-24
sstable: track and record compression stats

RaduBerinde:
#### block: remove NoopCompressor

Replace the NoopCompressor with new variants for `ChecksumAndCompress`
and inform the `block.Compressor` of any uncompressed blocks.

#### sstable: track and record compression stats

Add `Properties.CompressionStats` which encodes the number of
compressed and uncompressed bytes for each compression setting used.

Informs #4797


RaduBerinde:
*[`sstable/block/compressor.go` line 116 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4945#-OTYOMx_9xDEm4CvjEg-:-OTYOMx_9xDEm4CvjEg0:bao7cmw) ([raw file](https://github.com/cockroachdb/pebble/blob/d441c60e6fa79a734b87191d7f8717ef00e79c3d/sstable/block/compressor.go#L116)):*
> ```Go
> 	n int
> 	// Compression profiles have three settings (data, value, other) and
> 	// NoCompression can also be used for data that didn't compress.
> ```

I could also store `UncompressedBytes` separately and make the stats string more readable, e.g. `None:X,Snapy:Y/Z`.

Issue #4872, opened 2025-06-10
db: always prefer L0 -> Lbase move compactions when possible


jbowens:
In #4785 it was observed that we sometimes pursue a L0 -> LBase compaction that rewrites input files when the input files have no overlap in LBase. We could've instead done individual move compactions.

Jira issue: PEBBLE-478

jbowens:
(Created a separate issue since #4785 was closed)

jbowens:
@xinhaoz is this good to close out?

xinhaoz:
Oops yeah my bad looks like I put two `#` in front of the issue num which may have prevented auto closing.

Issue #4785, opened 2025-05-29
Question: batched 4 x 4KB writes faster than single 16KB write


ijsong:
I'm seeing a performance difference between writing four 4KB values in a batch versus a single 16KB value. The batched writes are significantly faster, and I'd like to understand why.

- Batched 4x4KB: ~1 GB/s
- Single 16KB: ~600 MB/s

Both tests use the identical Pebble (v2.0.3) configuration:
```txt
use_sync=false
wal_bytes_per_sync=512KB
wal_min_sync_interval=0s
sst_bytes_per_sync=512KB
memtable_size=64MB
memtable_stop_writes_threshold=4
max_concurrent_compactions=4
l0_target_file_size=64MB
l0_compaction_concurrency=2
lbase_max_bytes=64MB
enable_multi_level_compaction_heuristic=false
```

Pebble Metrics:
Case 1: batched writes (4 x 4KB)
```txt
...

      |                             |       |       |   ingested   |     moved    |    written   |       |    amp
level | tables  size val-bl vtables | score |   in  | tables  size | tables  size | tables  size |  read |   r   w
------+-----------------------------+-------+-------+--------------+--------------+--------------+-------+---------
    0 |     0     0B     0B       0 |  0.00 |  71GB |     0     0B |     0     0B |   997   15GB |    0B |   0  0.2
    1 |     0     0B     0B       0 |  0.00 |    0B |     0     0B |     0     0B |     0     0B |    0B |   0  0.0
    2 |     0     0B     0B       0 |  0.00 |    0B |     0     0B |     0     0B |     0     0B |    0B |   0  0.0
    3 |     4   52MB     0B       0 |  0.82 |    0B |     0     0B |   544  7.6GB |     0     0B |    0B |   1  0.0
    4 |    23  380MB     0B       0 |  1.00 |    0B |     0     0B |   947   14GB |     0     0B |    0B |   1  0.0
    5 |   166  2.2GB     0B       0 |  1.00 |    0B |     0     0B |   965   14GB |     0     0B |    0B |   1  0.0
    6 |   804   12GB     0B       0 |     - |    0B |     0     0B |   804   12GB |     0     0B |    0B |   1  0.0
total |   997   15GB     0B       0 |     - |  71GB |     0     0B |  3.3K   48GB |   997   85GB |    0B |   4  1.2
-------------------------------------------------------------------------------------------------------------------
WAL: 2 files (105MB)  in: 71GB  written: 71GB (0% overhead)
[...]

PR #4947, opened 2025-06-24
pebble: prefer L0->Lbase move compactions when possible

xinhaoz:
The existing compaction grow logic attempts to add additional start level files
by expanding the output level's initial key range, which is based off of
overlapping files with L0. Previously, we would always try to expand the
compaction range even when the output key range was empty, causing the L0 files
in those empty output ranges to be rewritten instead of being moved.

This commit makes it such that we only expand the compaction output range and
inputs when there is at least one overlapping file in lbase, preferring move
compactions when possible. This behaviour mirrors the grow logic for
LPositive-LPositive compactions.

Fixes: #https://github.com/cockroachdb/pebble/issues/4872

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -527,29 +527,26 @@ func (pc *pickedTableCompaction) growL0ForBase(cmp base.Compare, maxExpandedByte
 			panic(fmt.Sprintf("pc.startLevel.level is %d, expected 0", pc.startLevel.level))
 		}
 	}
+
+	if pc.outputLevel.files.Empty() {
+		// If there are no overlapping fields in the output level, we do not
+		// attempt to expand the compaction to encourage move compactions.
+		return false
+	}
+
 	smallestBaseKey := base.InvalidInternalKey
 	largestBaseKey := base.InvalidInternalKey
-	if pc.outputLevel.files.Empty() {
-		baseIter := pc.version.Levels[pc.outputLevel.level].Iter()
-		if sm := baseIter.SeekLT(cmp, pc.bounds.Start); sm != nil {
+	// NB: We use Reslice to access the underlying level's files, but
+	// we discard the returned slice. The pc.outputLevel.files slice
+	// is not modified.
+	_ = pc.outputLevel.files.Reslice(func(start, end *manifest.LevelIterator) {
+		if sm := start.Prev(); sm != nil {
 			smallestBaseKey = sm.Largest()
 		}
-		if la := baseIter.SeekGE(cmp, pc.bounds.End.Key); la != nil {
+		if la := end.Next(); la != nil {
 			largestBaseKey = la.Smallest()
 		}
-	} else {
-		// NB: We use Reslice to access the underlying level's files, but
-		// we discard the returned slice. The pc.outputLevel.files slice
-		// is not modified.
-		_ = pc.outputLevel.files.Reslice(func(start, end *manifest.LevelIterator) {
-			if sm := start.Prev(); sm != nil {
-				smallestBaseKey = sm.Largest()
-			}
-			if la := end.Next(); la != nil {
-				largestBaseKey = la.Smallest()
-			}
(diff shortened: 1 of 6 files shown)


RaduBerinde (commented):
Is there a specific test that we can add (insipred from #4785 workload)? Something that would behave differently without the change.

There is a potential downside that we end up with too many small files in L6. I'm not sure how much of a problem this can be in practice.. But maybe we want to wait until 25.3 branching and only merge this on master. CC @jbowens 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4947#-:-OTYM5y_BxB2neB0s6yw:b-6jqo8c)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @xxmplus)

___
*[`testdata/compaction/value_separation` line 358 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4947#-OTYLztH1tuxw96-IX7A:-OTYLztH1tuxw96-IX7B:b-1o5ue6) ([raw file](https://github.com/cockroachdb/pebble/blob/4568349692cf218ad5a4ebf1cbcf53ba0529acdf/testdata/compaction/value_separation#L358)):*
> ```
> 
> # Schedule automatic compactions. These compactions should write data to L6. The
> # resulting sstables will reference multiple blob files but maintain a blob
> ```

Feels like this is no longer testing what we want? (multiple blob refs)

xinhaoz (commented):
> Is there a specific test that we can add (insipred from [ Question: batched 4 x 4KB writes faster than single 16KB write #4785](https://github.com/cockroachdb/pebble/issues/4785) workload)? Something that would behave differently without the change.

I just extracted the blobref's populate cmd as the test for the move compactions, since it generates sequential, non-overlapping keys, similar to the workload from the issue - lmk if that works or if we want something more specific.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4947#-:-OTcRoLVEe5iOWtd0ZlD:b1vuxnm)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @xxmplus)

___
*[`testdata/compaction/value_separation` line 358 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4947#-OTYLztH1tuxw96-IX7A:-OTcSXKkCp3BfNGaXiET:b-kcdlrz) ([raw file](https://github.com/cockroachdb/pebble/blob/4568349692cf218ad5a4ebf1cbcf53ba0529acdf/testdata/compaction/value_separation#L358)):*

Feels like this is no longer testing what we want? (multiple blob refs)
</blockquote></details>

Reverted to force the old compaction behaviour in this test.

RaduBerinde (approved):
:lgtm: Nice! Thanks!

Up to @jbowens if we want this in 25.3 or not. If not, I'd hold off merging for a day or two until we have a `crl-release-25.3` branch.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4947#-:-OTcWdBHEXtlURkx0FpR:b-h462h0)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @xxmplus)

___
*[`testdata/compaction/value_separation` line 358 at r2](https://reviewable.io/reviews/cockroachdb/pebble/4947#-OTcWYk-1-XB9DQ864Nb:-OTcWYk-1-XB9DQ864Nc:bx9561t) ([raw file](https://github.com/cockroachdb/pebble/blob/8891f47af26842ffedd56b52cc43ae44833cfade/testdata/compaction/value_separation#L358)):*
> ```
> 
> # Manual compaction of key range so we merge files instead of moving so we
> # rewrite the files. This compactions should write data to L6. The
> ```

[nit] This compaction

xinhaoz:
Tried to deflake `auto-compact` cmd but it seems like blob rewrites are missing in compaction count metrics - will put up a separate PR to resolve that.

PR #4948, opened 2025-06-24
db: remove redundant blob files in datadriven output

jbowens:
Now that the Version.DebugString prints a Version's blob files, the datadriven test helpers don't need to explicitly print it too.


jbowens:
TFTR!

PR #4953, opened 2025-06-24
test: introduce single level iterator lazy load benchmark test

xxmplus:
Here is the benchmark between eager load and lazy load behavior of the single level iterator. The tests just run against the iterator, in order to get the metrics, you will need to run before and after the commit #4916 

Ideally I expect to see the difference between construction time (or the index block load overhead, or 444 ns here) is similar to other differences. However, it seems they are higher. I believe the main reason is the Timer overhead introduced given the way the code is written+other noise. Given the difference is small (~200 ns at most), I'd argue it's good enough.

I'm only interested in the first-time access of these functions, which will trigger index block loading when the lazy load logic is applied. Therefore I cannot organize the test in form of

```
b.StartTimer()
for i := 0; i < b.N; i++ {
    doSomething()
}
b.StopTimer()
```
But have to do it like below
```
for i := 0; i < b.N; i++ {
    prepare()
    b.StartTimer()
    doWork()
    b.StopTimer()
}
```

| Operation | Lazy Loading | Eager Loading | Performance |
| ---------- | -------------- | ------------- | ------------- |
| Constructor Only | 55 ns/op | 499 ns/op  | -444 ns |
| First Access (First()) | 2000 ns/op | 1451 ns/op |  +549 ns |
| First Access (SeekGE()) | 2240 ns/op | 1617 ns/op | +623 ns |
| SeekPrefixGE() hit | 2254 ns/op | 1623 ns/op | +631 ns |
| SeekPrefixGE() no hit | 771 ns/op | 120 ns/op | +651 ns |

Related to #3248 and #4916

### DIFF
--- a/sstable/single_lvl_iter_benchmark_test.go
+++ b/sstable/single_lvl_iter_benchmark_test.go
@@ -1,0 +1,188 @@
+// Copyright 2025 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package sstable
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/cockroachdb/pebble/bloom"
+	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/objstorage/objstorageprovider"
+	"github.com/cockroachdb/pebble/sstable/block"
+	"github.com/cockroachdb/pebble/vfs"
+)
+
+var (
+	benchReader   *Reader
+	benchIterOpts IterOptions
+)
+
+func init() {
+	mem := vfs.NewMem()
+	f, err := mem.Create("bench.sst", vfs.WriteCategoryUnspecified)
+	if err != nil {
+		panic(err)
+	}
+
+	w := NewWriter(objstorageprovider.NewFileWritable(f), WriterOptions{
+		BlockSize:      4096,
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

Separate PR is ok too, I just want to clarify that my suggestion in the other PR was to separate the benchmark in a *commit* within that same PR.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4953#-:-OT_5iOgAYjPrZa2wfRy:brikaz5)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/single_lvl_iter_benchmark_test.go` line 178 at r1](https://reviewable.io/reviews/cockroachdb/pebble/4953#-OT_3Sda045fBO8audHZ:-OT_3Sda045fBO8audH_:bcxhprk) ([raw file](https://github.com/cockroachdb/pebble/blob/e709b3852e0961aebb0a60e39a292d50909bc10d/sstable/single_lvl_iter_benchmark_test.go#L178)):*
> ```Go
> 	b.ResetTimer()
> 	for i := 0; i < b.N; i++ {
> 		b.StopTimer()
> ```

Starting and stopping the timer so often might affect the results. Why not measure both the construction and the first seek instead of trying to isolate the seek? IMO that's what we're interested in anyway.

xxmplus:
TFTR!

PR #4956, opened 2025-06-25
db: TestCompaction: fix async-compact usage

### LINKED ISSUE #4939: master: nightly linux failed
The nightly linux test run failed on 70aa8f70b9bc3a7fc478585a1513692890d8f694. Please review the run [here](https://github.com/cockroachdb/pebble/actions/runs/15847479863).

RaduBerinde:
The `async-compact` directive is supposed to be used when the
compaction is expected to block. It is currently also used when the
compaction does not block; in this mode it is flaky because it checks
if the compaction is done immediately after seeing it go into the
queue.  There's a chance the compaction is not blocked but takes a
little bit to complete.

We change this case to a `Fatalf` and switch to using `compact` for
the cases where it's not supposed to block.

Fixes #4939

### DIFF
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -1113,18 +1113,12 @@ func TestCompaction(t *testing.T) {
 					close(ch)
 				}()
 
-				manualDone := func() bool {
+				// Wait until the manual compaction is queued.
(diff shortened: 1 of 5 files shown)


RaduBerinde:
TFTR!

PR #4937, opened 2025-06-23
compression: add an adaptive compressor

RaduBerinde:
Informs #4925

#### compression: return Setting from Compress

Flip-flopping on my earlier change :)

#### ewma: add a per-bye EWMA estimator

This will be used to estimate compression ratio of blocks based on the
compression ratios of recently sampled blocks.

#### compression: add an adaptive compressor

Add AdaptiveCompressor, which estimates the size reduction of using a
slower algorithm vs a faster one and chooses automatically (on a
per-block basis).

We will separately update the compression analyzer to run experiments
with the adaptive compressors.

#### compressionanalyzer: support adaptive compressors

Switch experiments from using `compression.Setting` to
`block.CompressionProfile` and add two adaptive profiles.

PR #4964, opened 2025-06-25
compression: fix zstd Compression to return Setting

annrpom:
One of the commits in https://github.com/cockroachdb/pebble/pull/4937 accidentally reverted the change to return Setting from Compress.

### DIFF
--- a/internal/compression/zstd_nocgo.go
+++ b/internal/compression/zstd_nocgo.go
@@ -34,7 +34,7 @@ func getZstdCompressor(level int) *zstdCompressor {
 // relies on CGo.
 const UseStandardZstdLib = false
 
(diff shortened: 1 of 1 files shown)


annrpom:
my oops too; i didn't notice :oopshehe:

TFTR! ('-')7

RaduBerinde:
This is still broken:
```
radu master+ pebble # CGO_ENABLED=0 go build ./...
# github.com/cockroachdb/pebble/internal/compression
internal/compression/zstd_nocgo.go:43:54: z.level undefined (type *zstdCompressor has no field or method level)
```

RaduBerinde:
I'll put up another PR, it's a nontrivial change.

annrpom:
oki

RaduBerinde:
#4965

PR #4965, opened 2025-06-25
compression: fix nocgo build

RaduBerinde:
Also add `make testnocgo` target and update the workflow (note: we
were running this without invariants before, that was not
intentional).


annrpom (approved):
:lgtm:

RaduBerinde:
TFTR!

PR #4966, opened 2025-06-25
crl-release-25.2: Makefile: add testnocgo

RaduBerinde:
We need this on the crl-release-25.2 branch because the 25.2 nightlies
workflow on master uses it.


RaduBerinde:
TFTR!

PR #4968, opened 2025-06-25
blob: uncouple blob file rewriting logic

jbowens:
Uncouple the logic around actually writing out a rewritten blob file from the
logic around identifying which values remain live. This will help us test both
parts in isolation.


jbowens:
TFTR!

PR #4972, opened 2025-06-26
db: don't ref version until blob rewrite compaction is scheduled

jbowens:
Blob rewrite compactions reference the version they use for the duration of the compaction. Previously the reference was added during compaction picking. This allowed the reference to be leaked if the compaction was picked and cached but never got an opportunity to run before the DB was closed. This commit defers referencing the Version until the compaction is scheduled.


annrpom (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4972#-:-OTmc3mL1r50WZ02Y25m:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @xinhaoz)

PR #4979, opened 2025-06-27
crl-release-24.2: metamorphic: bound time spent level checking 

xinhaoz:
The metamorphic test runs with the level checker validating every new
readState. This level checking happens while holding the database mutex,
preventing the test from making forward progress. Some configurations can
create pathological LSMs with tens of thousands of sstables, which are slow to
validate with the level checker.

This commit addresses this issue by tracking the test wall time and the
cumulative wall time spent in the level checker. It avoids running the level
checker if the cumulative time spent in the level checker exceeds 25% of the
overall run's run time.

Additionally, regardless of the cumulative time budget, it skips the level
checker half the time. This is intended to ensure that inserting artifical
latency during read-state installation through level checking doesn't obscure
subtle races.

Fix https://github.com/cockroachdb/pebble/issues/4517.
Fix https://github.com/cockroachdb/pebble/issues/4338.
Fix https://github.com/cockroachdb/pebble/issues/4202.

### DIFF
--- a/metamorphic/options.go
+++ b/metamorphic/options.go
@@ -289,27 +289,34 @@ func defaultOptions() *pebble.Options {
 		BlockPropertyCollectors: blockPropertyCollectorConstructors,
 	}
 
-	// We don't want to run the level checker every time because it can slow down
-	// downloads and background compactions too much.
+	// The level checker runs every time a new read state is installed: every
+	// compaction, flush, ingest completion, etc. It runs while the database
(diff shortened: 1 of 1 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/4979#-:-OTmmT1Q7UM8zXUP_gRJ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @annrpom)

PR #4982, opened 2025-06-27
crl-release-25.3: db: don't ref version until blob rewrite compaction is scheduled

jbowens:
Blob rewrite compactions reference the version they use for the duration of the compaction. Previously the reference was added during compaction picking. This allowed the reference to be leaked if the compaction was picked and cached but never got an opportunity to run before the DB was closed. This commit defers referencing the Version until the compaction is scheduled.


annrpom (approved):

jbowens:
TFTR!

PR #5007, opened 2025-06-30
crl-release-25.3: db: incorporate separated values when estimating compression ratios

jbowens:
Table stats collection estimates compression ratios for the purpose of
calculating the size of data deleted by tombstones. Previously, this
compression ratio calculation included the size of both keys and values in the
pre-compression sum but only included sstables in the post-compression sum.
This resulted in wildly overestimating the effectiveness of compression when
values were separated into external blob files.

This commit incorporates a table's references estimated physical size into the
calculation.

Informs https://github.com/cockroachdb/cockroach/issues/149147.


annrpom (approved):
:lgtm:

jbowens:
TFTR!

Issue #4915, opened 2025-06-19
db: calculate virtual sstable blob value liveness


jbowens:
With #4807 we will begin to rewrite and replace physical blob files, reclaiming disk space of values that are no longer referenced. A value may become unreferenced through the conversion of a sstable into one or more virtual sstables. The backing sstable's blob value liveness block reflects the liveness of the physical sstable with its original bounds, and cannot take into account the virtual sstable's bounds.

We could...

1. During rewriting we could seek into the virtual sstable to identify the minimum and maximum (blockID,valueID) tuple for each blob file the sstable references. This would likely require a full scan of the virtual sstable (if a blob file is referenced by the backing file but unreferenced within the virtual sstable's bounds, we have no means of knowing there are no references).
2. We could extend our existing blob liveness index to encode additional information about which data blocks reference which blob files. That would prevent the above scan from devolving into a full scan of the virtual sstable's bounds.

Jira issue: PEBBLE-487

Epic CRDB-53453

jbowens:
Recapping our meeting, instead of trying to improve the fidelity of the estimate of virtual sstables references to blob files we're going to instead more eagerly materialize virtual sstables as physical sstables (thereby removing any error due to the linear interpolation).

First we want to add a new `BackingValueSize` to `manifest.BlobReference` that will hold the maximum possible volume of values that a sstable references within a blob file. For non-virtual sstables, `BackingValueSize == ValueSize`. For virtual sstables, the `ValueSize` is the interpolated estimate and the `BackingValueSize` is the underlying backing file's `ValueSize`. We'll use the `BackingValueSize` in the `manifest.CurrentBlobFileSet` when estimating the percent of a blob file that's referenced (preventing unproductive rewrites of blob files that are referenced by virtual sstables).

Second we want to add a new kind of compaction that takes a backing file, finds all the referencing virtual sstables and then rewrites the virtual sstables replacing them each with a materialized version. We'll pick these compactions at a low priority. Since we will be compacting all the referencing virtual sstables at once, we ensure when the compaction completes we can reclaim the disk space of the backing file.

PR #5011, opened 2025-06-30
blob: deduplicate values during rewrite

jbowens:
Because of the existence of virtual sstables and the lack of accurate blob
value liveness for virtual sstables, a blob file rewrite may observe multiple
references to the same value. Blob file rewriting must deduplicate these value
IDs to preserve the correct indexing of values within the virtual block.

Note that it's not sufficient to deduplicate referencing sstables by backing
sstable. Consider a value V contained within virtual sstable s1, backed by b1.
Imagine there's another virtual sstable s2 also backed by b1. If s1 is
compacted, the reference to value V will be carried forward to a new physical
sstable b2 which will encode the reference in its blob value liveness index.
However the original backing file b1 is still in use, and its blob value
liveness index continues to be in-use.

Ultimately, we will fix this issue with accurate virtual sstable blob value
liveness (https://github.com/cockroachdb/pebble/issues/4915). For now, we deduplicate the valueIDs at rewrite time.

Additionally, enable blob file rewrite compactions during the metamorphic
tests.


jbowens:
TFTR!

PR #5014, opened 2025-07-01
compression: deflake TestAdaptiveCompressorCompressible

### LINKED ISSUE #5012: master: nightly linux-stress tests failed
The nightly linux-stress test run failed on 8e65e154c20743604dc579510de85e1dd6cbc51f. Please review the run [here](https://github.com/cockroachdb/pebble/actions/runs/15996232149).

jbowens:
Deflake TestAdaptiveCompressorCompressible by increasing the minimum size of the payload that's compressed. If the payload is too small, compression cannot result in a significant enough reduction in the size of the compressed payload and the compressor yields the uncompressed payload instead.

Fix #5012.


jbowens:
TFTR!

PR #5016, opened 2025-07-01
cmd/pebble: enable blob file rewriting

jbowens:
Adjust the TargetGarbageRatio setting to enable blob file rewrites. This effectively enables blob file rewrite compactions in the ycsb nightly benchmarks that make use of value separation.


annrpom (approved):
:lgtm: !

jbowens:
TFTR!

PR #5018, opened 2025-07-01
pebble: make MultiLevelCompactionHeuristic a function option

xinhaoz:
Previously, the `MultiLevelCompactionHeuristic` option was an interface field representing which multilevel compaction heuristic to use. This commit changes this field to be a function returning the heuristic, so that we may add more logic in determining which heuristic to use during runtime.


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5018#-:-OUBtrQE99x-4d3WPojR:bnfp4nl)* status: 0 of 8 files reviewed, all discussions resolved

Issue #4139, opened 2024-11-06
db: limit concurrency of multilevel compactions


itsbilal:
On a test cluster, we observed that having all compaction slots be taken up by relatively long-running (but not excessively long running) multilevel compactions can make the store very susceptible to transient increases in L0 sublevels. This is concerning because L0 sublevel count is used as a tell-all measure of store overload, and even a temporary spike can have undesirable side-effects in CockroachDB.

For instance, these three ML compactions all overlapped around 08:35:30-08:36:00, resulting in an L0 sublevel spike from 2 to 8 before the next (L0->LBase) compaction could be scheduled:

```
I241106 08:35:59.319548 693190905 3@pebble/event.go:776 ⋮ [n5,s5,pebble] 2825380  [JOB 552595] compacted(default) multilevel L3 [7131468] (14MB) Score=1.45 + L4 [7130082 7130092 7130097 7130121 713
0126 7130139 7130143 7130149 7130151 7130153 7097821 7076086] (152MB) Score=3.13 + L5 [7130352 7130361 6982483 6982597 6982684 6982729 6982819 6982828 6976618 6976623 7001341 7001361 7001417 700153
4 7001574 7001575 6825834 6990001 6990012 6990088 6990131 6990135 6990249] (963MB) Score=1.13 -> L5 [7133772 7133816 7133893 7133985 7133998 7134080 7134162 7134197 7134268 7134358 7134451 7134562
7134654 7134678 7134746 7134774 7134775 7134803 7134819 7134820] (931MB), in 109.0s (109.0s total), output rate 8.5MB/s
...
I241106 08:36:13.404917 693626967 3@pebble/event.go:776 ⋮ [n5,s5,pebble] 2825888  [JOB 552986] compacted(default) multilevel L3 [7134712] (11MB) Score=1.45 + L4 [6997130 7113394 7113395 7113411 7113417] (79MB) Score=3.10 + L5 [6561966 6989463 6989617 6989684 6960989 6960999 6961000 6562124 7015976 7015998] (328MB) Score=1.13 -> L5 [7134725 7134747 7134776 7134801 7134818 7134884 7134948 7134956] (378MB), in 47.7s (47.7s total), output rate 7.9MB/s
....
[...]

PR #5020, opened 2025-07-01
pebble: add additional constraints to running multiLevel compactions

xinhaoz:
Previously, we decide to run a multilevel compaction if the write amp with the additional level is less than that of the original compaction. We now add these additional constraints:
- The max compaction concurrency for the db must be greater than 1. This is to prevent the ML compaction from being the only compaction running, which could spike L0.
- At most one multiLevel compaction may be running at any time.

Part of: #4139

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -587,6 +587,25 @@ func (pc *pickedTableCompaction) estimatedInputSize() uint64 {
 
 // setupMultiLevelCandidate returns true if it successfully added another level
 // to the compaction.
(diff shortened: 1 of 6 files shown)


annrpom (approved):
:lgtm:

> Reviewed 5 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5020#-:-OUFuScz5JugRD6iCFV7:bnfp4nl)* status: 5 of 6 files reviewed, all discussions resolved (waiting on @xxmplus)

Issue #5015, opened 2025-07-01
db: collect effective read-amp metric


jbowens:
We maintain a read amplification metric that we surface in CockroachDB as a timeseries metric. This metric is the worst-case read amplification in terms of the number of sstables consulted.

We should collect additional metrics that allow us to measure the effective read-amp actually observed by readers too. Optimizations like #2366 do not reduce the worst-case read amplification but can significantly reduce the number of read I/Os in practice.

PR #5024, opened 2025-07-02
db: maintain per-block kind internal iterator stats

### LINKED ISSUE #5021: db: use iterator stats to back compaction bytes read metric
Currently the compaction bytes read metric is based on the size of compaction inputs, not based on the actual volume read from disk. This is confusing, especially with value separation when we may not need to read the values at all. We can use the iterator block load bytes metric instead, maintaining a total per-level.

Jira issue: PEBBLE-531

jbowens:
**db: maintain per-block kind internal iterator stats**

Expand the internal iterator's per-block stats to include counts of blocks
loaded cached and uncached, and convert the block stats to be accumulated
per-block kind.

Informs #5015.

**db: use internal iterator stats to populate compaction bytes read**

The LevelMetrics exposes two fields describing the volume of bytes read during
compactions: one from sstables and one from blob files. Confusing, these
metrics weren't based on what a compaction actually read but computed based on
the input of the sstable. Blocks that didn't needed to be loaded (eg, including
blob value blocks when preserving existing blob references) were still
recorded.

This commit adapts these metrics to use the block bytes metric maintained by
the internal iterator stack, reflecting the volume of bytes actually loaded by
a compaction. This will make it easier to understand what portion of disk read
volume is driven by compactions.

Close #5021.


sumeerbhola (changes requested):
> Reviewed 11 of 11 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5024)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`-- commits` line 23 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5024#-OUeFGis2rt-7D7sU91X:-OUeFGis2rt-7D7sU91Y:b-lcknyq):*
Definitely a big improvement. I wish we could somehow know what missed the page cache.
___
*[`metrics.go` line 85 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5024#-OUeG5Or475-dSA0yxTe:-OUeG5Or475-dSA0yxTf:bar1nh8) ([raw file](https://github.com/cockroachdb/pebble/blob/f2e3b3ff24da7d9de073322de149040f46ccb59d/metrics.go#L85)):*
> ```Go
> 	// The number of bytes read for compactions at the level. This includes bytes
> 	// read from other levels (BytesIn), as well as bytes read for the level.
> 	TableBytesRead uint64
> ```

Looks like this is still populated based on the `TableSizeSum`. Should this be switched to the new approach since that will exclude things like the bloomfilter block, the blob reference liveness block, and will may correct for any inaccuracy related to virtual ssts being inputs to compactions?

sumeerbhola (approved):
:lgtm: 

> Reviewed 1 of 6 files at r3, 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5024#-:-OUkNrD_-Dv39S3cNmfs:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @jbowens)

jbowens:
TFTR!

Issue #4919, opened 2025-06-20
db: lock WAL failover dir


RaduBerinde:
We use locking on the db dir to make sure multiple processes don't use the same store.

We should do the same with the WAL failover dir.  If two stores are incorrectly configured to use the same WAL failover directory, the two stores can apply and delete each other's WALs, leading to corruption. This is a double footgun.

Jira issue: PEBBLE-489

PR #5045, opened 2025-07-08
pebble: lock wal directories on Open

xinhaoz:
Ensure that we acquire locks for WAL directories when they are configured.

Fixes: #4919

### DIFF
--- a/db.go
+++ b/db.go
@@ -290,8 +290,8 @@ type DB struct {
 	// objProvider is used to access and manage SSTs.
 	objProvider objstorage.Provider
 
-	fileLock *Lock
-	dataDir  vfs.File
+	dataDirLock *base.DirLock
+	dataDir     vfs.File
 
 	fileCache            *fileCacheHandle
 	newIters             tableNewIters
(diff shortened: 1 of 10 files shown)


jbowens (approved):
:lgtm_strong: :

> Reviewed 2 of 6 files at r1, 5 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5045#-:-OUl6HgB4ydPvGIjKcd_:b-yncq55)* status: 7 of 10 files reviewed, 2 unresolved discussions (waiting on @annrpom and @xinhaoz)

___
*[`options.go` line 1348 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5045#-OUjw8pZBb1Eqhy4EiRl:-OUl64ff9yss3KoNsx3W:b-dizjzs) ([raw file](https://github.com/cockroachdb/pebble/blob/011275003b5ee80f055067763258558464f20ddd/options.go#L1348)):*

Done - I assume we only need to lock the recovery dirs during Open, but lmk if that's not accurate.
</blockquote></details>

Yeah, just for the duration of Open is good
___
*[`internal/base/directory_lock.go` line 57 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5045#-OUl0K7y6xE5ldcEfJ4r:-OUl0K7y6xE5ldcEfJ4s:b-7bpkdl) ([raw file](https://github.com/cockroachdb/pebble/blob/5a77c90f391df35ac3e3d5456b185c50f9d0c77c/internal/base/directory_lock.go#L57)):*
> ```Go
> }
> 
> // Lock represents a file lock on a directory. It may be passed to Open through
> ```

nit: s/Lock represents/DirLock represents/

Issue #5043, opened 2025-07-08
db: support compactions in response to policy changes


jbowens:
Relates to #4814.
Related to cockroachdb/cockroach#74804.

At runtime, the policies that configure the behavior of a compaction may change. Value separation may become enabled or disabled. The minimum size that triggers separation may be raised or lowered. The same applies to rollout of previous configurations like columnar blocks.

In particular, a customer may configure value separation to be more aggressive than they can actually tolerate, resulting in read latency. If the customer reverts the setting, sstables will only be outputted using the new policy under new compactions. Ideally we would rewrite these sstables at low priority to restore the old latency characteristics automatically, without requiring new incoming writes to churn compactions.

RaduBerinde:
Compression changes require the same.

jbowens:
This is a duplicate of #5657. (cc @xinhaoz)

PR #5047, opened 2025-07-09
sstable: include smallest separated value size in properties

jbowens:
Record a new sstable property containing the size of the smallest value separated. This can be used to schedule the rewrite of sstables that contain separated values that are smaller than what is permitted by the current policy. See #5043.

Value separation doesn't lend itself well to a generalized approach to #5043 because when a sstable preserves blob references, the input sstables may have been written according to a variety of different policies.


annrpom (approved):
:lgtm:

> Reviewed 16 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5047#-:-OUkhimR7W6Z9AAXP1i8:bnfp4nl)* status: 16 of 18 files reviewed, all discussions resolved

RaduBerinde (approved):

Issue #4798, opened 2025-05-30
manifest: explicitly encode excise operations in the MANIFEST


jbowens:
The replay package is used for compaction benchmarking. It captures data from flushes and ingestions to create a stable compaction workload. It then replays it on a Pebble instance, allowing for focused benchmarking of compactions and their heuristics. However, this process was not updated with the addition of excise, and excises are not repeated. This results in a deviation between the replay database and the original workload.

Trying to deduce the exicse bounds from existing version edit data is a bit fraught and likely fragile. We could instead explicitly encode the excise operation within the version edit, mapping a span to the sequence number at which the excise was performed. This would also allow us to strengthen invariant checking on manifest replay, and then would also allow the replay package to replay excises.

Jira issue: PEBBLE-463

PR #5048, opened 2025-07-09
manifest: explicitly encode excise operations in the MANIFEST

Sachuman:
This adds a new struct field that maps excise span to the sequence number at which it was performed, this will allow to strengthen invariant checking on manifest replay as it was not updated with excise operations, this field is also implemented in version_edit_apply and version_edit_decode tests.

Addresses #4798

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1587,6 +1587,11 @@ func (d *DB) runIngestFlush(c *tableCompaction) (*manifest.VersionEdit, error) {
 	}
 	if ingestFlushable.exciseSpan.Valid() {
 		exciseBounds := ingestFlushable.exciseSpan.UserKeyBounds()
+		ve.ExciseBoundsRecord = append(ve.ExciseBoundsRecord, manifest.ExciseOpEntry{
+			Bounds: exciseBounds,
+			SeqNum: ingestFlushable.exciseSeqNum,
+		})
+
 		// Iterate through all levels and find files that intersect with exciseSpan.
 		for layer, ls := range version.AllLevelsAndSublevels() {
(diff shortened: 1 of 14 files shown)


compaction.go
```diff
@@ -1587,6 +1587,11 @@ func (d *DB) runIngestFlush(c *tableCompaction) (*manifest.VersionEdit, error) {
 	}
 	if ingestFlushable.exciseSpan.Valid() {
 		exciseBounds := ingestFlushable.exciseSpan.UserKeyBounds()
+		ve.ExciseBoundsRecord = append(ve.ExciseBoundsRecord, manifest.ExciseOpEntry{
+			Bounds: ingestFlushable.exciseSpan.UserKeyBounds(),
```
annrpom, line 1591: nit: we can use `exciseBounds` here (so:

```
Bounds: exciseBounds,
```

compaction.go
```diff
@@ -1602,7 +1607,7 @@ func (d *DB) runIngestFlush(c *tableCompaction) (*manifest.VersionEdit, error) {
 	}
 
 	if len(ingestSplitFiles) > 0 {
-		if err := d.ingestSplit(context.TODO(), ve, updateLevelMetricsOnExcise, ingestSplitFiles, replacedTables); err != nil {
+		if err := d.ingestSplit(context.TODO(), ve, updateLevelMetricsOnExcise, ingestSplitFiles, replacedTables, ingestFlushable.exciseSeqNum); err != nil {
```
annrpom, line 1610: i'm not sure if this will be `ingestFlushable.exciseSeqNum`; @jbowens thoughts?

annrpom (commented):
yay! nice job! just had a small question

annrpom (approved):
:lgtm:

sjain022:
TFTRs!

PR #5049, opened 2025-07-09
sstable: add separated value retrieval internal iterator stats

jbowens:
Add internal iterator stats for the count of separated value retrievals and the subset of them that miss the iterator's reader cache.

Informs cockroachdb/cockroach#149422.


jbowens:
TFTR!

PR #5060, opened 2025-07-11
db: fix double Close of directory lock during failed Open

jbowens:
Previously if an error occurred during Open after the WAL manager had been initialized, it was possible for a directory lock to be Closed multiple times. This commit refactors the error handling to ensure we release the WALDir and WALFailover.Secondary locks at most once.


jbowens:
TFTR!

PR #5067, opened 2025-07-13
metamorphic: propagate FS changes to TestOptions

RaduBerinde:
Tests that run a meta test and then run another one on the same
filesystem currently start over from the place of the first restart
operation.

This commit propagates the restart FS changes to the `*TestOptions`.
Note that `New()` was already modifying them (by calling `WithFSDefaults()`).


RaduBerinde:
TFTR!

PR #5072, opened 2025-07-14
sstable: mangle some more buffers in invariants builds

jbowens:
Mangle some more byte buffers when they become unused to help shake out memory safety bugs.


jbowens:
TFTR!

Issue #4970, opened 2025-06-26
db: better observability for slow ingests


sumeerbhola:
An ingest produces a `TableIngestInfo` for the `EventListener`, that in CockroachDB is directed to the Pebble log. When an ingest operation is slow, it is not obvious why it was slow. We should measure and include these durations in the `TableIngestInfo` event. For example:
- how long did it spend waiting for memtable flushes: this would have made diagnosing https://github.com/cockroachdb/pebble/issues/4944 trivial.
- The total time spent in reading blocks for the various iterators it opened, as represented by the sum of `InternalIteratorStats.BlockReadDuration`, and the number of blocks read `InternalIteratorStats.BlockBytes-InternalIteratorStats.BlockBytesInCache`.
- The time spent in updating the MANIFEST.

We were unable to diagnose the root cause of https://github.com/cockroachdb/cockroach/issues/148337#issuecomment-2993053472, and having the above would have helped.

Jira issue: PEBBLE-506

PR #5073, opened 2025-07-14
db: track memtable flush, manifest update durations in TableIngestInfo

annrpom:
This patch tracks the duration of memtable flushes (for flushable ingests) and MANIFEST updates in our TableIngestInfo. This gives us more observability into slow ingests.

Informs: #4970


annrpom:
i'll open this one for a review now and get the last duration in in a diff PR:
```
The total time spent in reading blocks for the various iterators it opened, as represented by the sum of InternalIteratorStats.BlockReadDuration, and the number of blocks read InternalIteratorStats.BlockBytes-InternalIteratorStats.BlockBytesInCache.
```

compaction.go
```diff
@@ -1368,23 +1368,27 @@ func (d *DB) calculateDiskAvailableBytes() uint64 {
 	return space.AvailBytes
 }
 
-// maybeScheduleFlush schedules a flush if necessary.
+// maybeScheduleFlush schedules a flush if necessary, returning the time stamp
```
annrpom, line 1371: do we hate this change?

jbowens: i don't love it--feels like grabbing the time in the caller would be cleaner

jbowens (commented):
> Reviewed 8 of 8 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5073)* status: all files reviewed, 4 unresolved discussions (waiting on @annrpom)

___
*[`event.go` line 582 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5073#-OWHh6ur8o-G0R78yDDy:-OWHh6ur8o-G0R78yDDz:bcofstc) ([raw file](https://github.com/cockroachdb/pebble/blob/f275179ee12e8f83d41f37524bd594e965fb91fa/event.go#L582)):*
> ```Go
> 			redact.Safe(humanize.Bytes.Uint64(t.Size)))
> 	}
> 	w.Printf("; manifest update took %.1fs; block read took %.1fs with %s block bytes read",
> ```

nit: "block read**s**"
___
*[`ingest.go` line 1926 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5073#-OWHfslWBLH4ZNFEipU9:-OWHfslWBLH4ZNFEipUA:bq7p9zd) ([raw file](https://github.com/cockroachdb/pebble/blob/dea631b579649e1d0570ac15565ae0e22a055a44/ingest.go#L1926)):*
> ```Go
> 	var metrics levelMetricsDelta
> 
> 	manifestUpdateStart := crtime.NowMono()
> ```

i think the section we're measuring is too wide--it includes the contents of the function we pass into `UpdateVersionLocked` which may issue reads. I think we should measure from [here](https://github.com/cockroachdb/pebble/blob/master/version_set.go#L473) until UpdateVersionLocked returns. We could add a new field to `versionUpdate` that holds the duration
___
*[`ingest.go` line 466 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5073#-OWHhge-79PYidq92R-6:-OWHhge-79PYidq92R-7:b58ai2m) ([raw file](https://github.com/cockroachdb/pebble/blob/f275179ee12e8f83d41f37524bd594e965fb91fa/ingest.go#L466)):*
> ```Go
> 
> 	externalFilesHaveLevel bool
> 	blockreadStats         base.BlockReadStats
> ```

nit: `blockReadStats`

compaction.go
annrpom: done

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5073)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`event.go` line 582 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5073#-OWHh6ur8o-G0R78yDDy:-OWLfYfK8Kj_ZNzKwbp5:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/f275179ee12e8f83d41f37524bd594e965fb91fa/event.go#L582)):*

nit: "block read**s**"
</blockquote></details>

done
___
*[`ingest.go` line 1926 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5073#-OWHfslWBLH4ZNFEipU9:-OWNBzXP0f5h1Df3BfHa:b-9e4qni) ([raw file](https://github.com/cockroachdb/pebble/blob/dea631b579649e1d0570ac15565ae0e22a055a44/ingest.go#L1926)):*

i think the section we're measuring is too wide--it includes the contents of the function we pass into `UpdateVersionLocked` which may issue reads. I think we should measure from [here](https://github.com/cockroachdb/pebble/blob/master/version_set.go#L473) until UpdateVersionLocked returns. We could add a new field to `versionUpdate` that holds the duration
</blockquote></details>

done -- i ended up wrapping the duration in a struct pointer so that we actually mutate it inside of UpdateVersionLocked, but if that's gross let me know
___
*[`ingest.go` line 466 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5073#-OWHhge-79PYidq92R-6:-OWLfZPq3QYLo9ofteyD:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/f275179ee12e8f83d41f37524bd594e965fb91fa/ingest.go#L466)):*

nit: `blockReadStats`
</blockquote></details>

done

annrpom:
TFTR! ('-')7

PR #5076, opened 2025-07-15
blob: verify virtual BlockID is in bounds

jbowens:
During invariant builds, when looking up a BlockID in a blob file that's been rewritten, validate that the BlockID is within bounds of the number of virtual blocks encoded within the custom block header.


jbowens:
Oops, the lookup function already does this.

PR #5077, opened 2025-07-15
manual: sometimes mangle freed buffers in invariants builds

jbowens:
In invariants builds, sometimes mangle manually-managed buffers just before freeing them. This can help surface code paths that are inadvertently using the memory after it's been freed.


annrpom (approved):
:lgtm:

jbowens:
TFTR!

RaduBerinde:
We can also consider adding some testing runs with `GODEBUG=clobberfree=1` which does something similar for Go objects.

PR #5080, opened 2025-07-16
manual: expand useGoAllocation's comment

jbowens:
Document the restriction on allocating all of the manual package's allocations either via Go or via CGo, but not a combination. A combination could run afoul of CGo pointer passing rules.


jbowens:
TFTR!

PR #5085, opened 2025-07-17
sstable: prevent block metadata pointers to garbage

jbowens:
When allocating memory for a new block, we allocate an additional block.MetadataSize bytes for holding a Go structure containing decoded data describing the block and maintaining pointers into the block's contents. This allows us to decode the block structure once when the block is loaded into the block cache. Readers that find the block in the cache simply cast this 'block metadata' preamble into the appropriate type.

When a block is not found within the block cache, we: load the block, decode the block's header and initialize the block's unique metadata. We initialize the metadata by first casting the allocation's pointer to the appropriate Go pointer type. Then we decode the block's data, initializing the metadata struct accordingly. In the interim between the pointer cast and the initialization, the contents of the metadata are undefined.

The per-block metadata types typically contain pointer fields. Once initialized, these pointers point into the block's data, all part of the same CGo manual allocation. If the Go garbage collector examines this Go struct before it's been fully initialized, it may observe arbitrary garbage within the pointer fields. If one of these values looks like a pointer to a Go-allocated part of the address space, the Go garbage collector crashes the process. It's a little unclear the circumstances in which the Go garbage collector will examine these metadata structs that exist within CGo memory, but experimentally it appears to be the case.

This commit fixes the issue by always zeroing block metadata memory before casting it into the appropriate metadata struct. Additionally, it solves an analogous problem during de-allocation by nil-ing metadata pointers into blocks before we release the blocks back to the cache or buffer pool. Together, these changes ensure that whenever we have a pointer to a block metadata Go structure, the structure's backing memory is either zeroed or the result of initialization based on decoding the block.

Informs cockroachdb/cockroach#149955.
Informs cockroachdb/cockroach#150216.

### DIFF
--- a/sstable/blob/blocks.go
+++ b/sstable/blob/blocks.go
@@ -236,10 +236,7 @@ const _ uint = block.MetadataSize - uint(unsafe.Sizeof(indexBlockDecoder{}))
 
 // initIndexBlockMetadata initializes the index block metadata.
 func initIndexBlockMetadata(md *block.Metadata, data []byte) (err error) {
-	if uintptr(unsafe.Pointer(md))%8 != 0 {
-		return errors.AssertionFailedf("metadata is not 8-byte aligned")
-	}
-	d := (*indexBlockDecoder)(unsafe.Pointer(md))
+	d := block.CastMetadataZero[indexBlockDecoder](md)
 	// Initialization can panic; convert panics to corruption errors (so higher
 	// layers can add file number and offset information).
 	defer func() {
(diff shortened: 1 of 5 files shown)


jbowens:
I'm going to run the reproducing test one more time with this SHA–previously I tested using the larger hammer of zeroing the entire buffer before giving a buffer out of a BufferPool.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5085#-:-OVPJaSW0KCBtG-pj8YL:bnfp4nl)* status: 0 of 5 files reviewed, 1 unresolved discussion

___
*[`sstable/block/block.go` line 230 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5085#-OVPJSmL-wwDoa6-QMam:-OVPJSmL-wwDoa6-QMan:bo5w85q) ([raw file](https://github.com/cockroachdb/pebble/blob/d828cefc53e1b41785e44bfa024c6b275a4a57ec/sstable/block/block.go#L230)):*
> ```Go
> 
> // CastMetadataZero casts the provided metadata to the type parameter T, zeroing
> // the memory backing the metadata first. This zeroing is necessary to ensure
> ```

[nit] This zeroing is necessary when first initializing the data structure

jbowens:
TFTR!

RaduBerinde:
For posterity, this article is relevant: https://ihagopian.com/posts/write-barriers-in-the-go-garbage-collector#block-ed4084c886ca47a7b2799cd8a4e8babb

TL;DR: if the GC is running and at a certain stage, any pointer assignment will cause the old value of the pointer to be saved for the GC to take into consideration. In the case of this bug, the old pointer was ~garbage from a previous block metadata structure (possibly for a different block type).

PR #5087, opened 2025-07-17
replay: adding ingest and ingest-and-exicse testing

Sachuman:
this commit adds ingestion and ingest-and-excise tests on replay package for improving benchmarks and accounts for excise operations during ingestion as well.

see #5048

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -1593,6 +1593,7 @@ func (d *DB) runIngestFlush(c *tableCompaction) (*manifest.VersionEdit, error) {
 				Bounds: exciseBounds,
 				SeqNum: ingestFlushable.exciseSeqNum,
 			})
+			d.mu.versions.metrics.Ingest.ExciseIngestCount++
 		}
 		// Iterate through all levels and find files that intersect with exciseSpan.
 		for layer, ls := range version.AllLevelsAndSublevels() {
--- a/ingest.go
+++ b/ingest.go
@@ -2082,6 +2082,7 @@ func (d *DB) ingestApply(
 		var exciseBounds base.UserKeyBounds
 		if exciseSpan.Valid() {
 			exciseBounds = exciseSpan.UserKeyBounds()
+			d.mu.versions.metrics.Ingest.ExciseIngestCount++
 			// Iterate through all levels and find files that intersect with exciseSpan.
 			//
 			// TODO(bilal): We could drop the DB mutex here as we don't need it for
--- a/internal/datatest/datatest.go
+++ b/internal/datatest/datatest.go
@@ -7,12 +7,16 @@
 package datatest
 
 import (
+	"context"
 	"strings"
 	"sync"
 
 	"github.com/cockroachdb/datadriven"
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble"
+	"github.com/cockroachdb/pebble/objstorage/objstorageprovider"
+	"github.com/cockroachdb/pebble/sstable"
+	"github.com/cockroachdb/pebble/vfs"
 )
 
 // TODO(jackson): Consider a refactoring that can consolidate this package and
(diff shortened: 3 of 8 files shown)


sjain022:
@RaduBerinde Hey, I saw this line in data_test.go // remove this in favor of runBuildSSTCmd.
I was wondering if it would be okay if I rewrote tests that use runBuildCmd in favor of runBuildSSTCmd.

RaduBerinde:
Sure! There might be some work needed to add features to runBuildSSTCmd (not sure).

sjain022:
> Sure! There might be some work needed to add features to runBuildSSTCmd (not sure).

I see I will look into it ! Thanks :) Though I was able to use it for ingestion in replay.

annrpom (approved):
:lgtm:

> Reviewed 3 of 8 files at r1, 4 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5087#-:-OWH3bJ98xjZ0yBEk42P:bnfp4nl)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`replay/replay.go` line 704 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5087#-OWGUoLv2WTumhlidADa:-OWGUoLwCNUJGutcfuzy:bl95567) ([raw file](https://github.com/cockroachdb/pebble/blob/dfe14a10ce7272c00a68e158f8da6b1139c3e1e0/replay/replay.go#L704)):*
> ```Go
> 			r.stepsApplied <- step
> 		case ingestAndExciseStepKind:
> 			if _, err := r.d.IngestAndExcise(context.Background(), step.tablesToIngest, nil, nil, step.exciseSpan); err != nil {
> ```

nit: for arguments where we cannot tell the type/usage, we usually annotate it with the name of parameter. can we do that here? something like:

```
if _, err := r.d.IngestAndExcise(context.Background(), step.tablesToIngest, nil /* shared */, nil /* external */, step.exciseSpan); err != nil {
```
___
*[`internal/datatest/datatest.go` line 147 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5087#-OWH3JNDCB8GW0ryZZWs:-OWH3JNDCB8GW0ryZZWt:b-suotnf) ([raw file](https://github.com/cockroachdb/pebble/blob/dfe14a10ce7272c00a68e158f8da6b1139c3e1e0/internal/datatest/datatest.go#L147)):*
> ```Go
> 
> // Below functions are more or less replica from data_test.go pebble package
> func RunBuildSSTCmd(
> ```

are we able to dedupe this code? i think it would need to involve putting this in a diff package to avoid cyclic dependencies; we can put this as a TODO for now and worry about it later perhaps

Sachuman (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5087)* status: 6 of 8 files reviewed, 1 unresolved discussion (waiting on @annrpom and @RaduBerinde)

___
*[`internal/datatest/datatest.go` line 147 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5087#-OWH3JNDCB8GW0ryZZWs:-OWHW5Ue1BaNO8BhRRCA:bgpejx8) ([raw file](https://github.com/cockroachdb/pebble/blob/dfe14a10ce7272c00a68e158f8da6b1139c3e1e0/internal/datatest/datatest.go#L147)):*

are we able to dedupe this code? i think it would need to involve putting this in a diff package to avoid cyclic dependencies; we can put this as a TODO for now and worry about it later perhaps
</blockquote></details>

I believe since data_test is not exported, I had to rewrite it. I will look into it, make a PR addressing this issue and also address the issue about removing runBuildCmd in favor of runBuildSSTCmd.

sjain022:
TFTR !

PR #5089, opened 2025-07-18
constants: remove package

RaduBerinde:
The constant expression `min(math.MaxUint32, math.MaxInt)` can just be
used directly instead (and is better self-documenting).


RaduBerinde:
TFTR!

PR #5091, opened 2025-07-19
metrics: improvements to the string representation

RaduBerinde:
#### metrics: improve exampleMetrics

Reorganize the code to mirror the order of the fields in the struct,
and fill in more of them (in particular the blob file metrics).

#### metrics: use compact humanization

Use the `Compact` and `OmitI` formatter flags, since we have to fit
many of these metrics on one line. This used to be the case
historically, before we switched some uses to `crhumanize`.

#### metrics: improve tables for compaction kinds and files

The compaction kinds table looks funny and is too long. We no longer
use the horizontal orientation feature (which has some issues, and
unnecessarily makes all columns have the same width); instead, we
transpose the table.

We also improve the files table to fit all information.


annrpom (approved):
:lgtm:

> Reviewed 3 of 3 files at r1, 8 of 8 files at r2, 8 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5091#-:-OVnLvF2CwcJJCHOefDa:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jbowens)

RaduBerinde:
TFTR!

PR #5093, opened 2025-07-20
colblk: remove unnecessary start ptr from RawBytes

jbowens:
Remove an unnecessary 'start' pointer from RawBytes. It was only used when formatting the structure. This allows us to reduce the size of the block.Metadata by 24 bytes. Additionally, we add test assertions on 64-bit platforms that ensure we don't leave MetadataSize larger than is necessary.

MetadataSize is allocated per block in the block cache (~ on the order of hundreds of thousands), so its overhead can be material.

### DIFF
--- a/cockroachkvs/cockroachkvs_64bit_test.go
+++ b/cockroachkvs/cockroachkvs_64bit_test.go
@@ -1,0 +1,17 @@
+// Copyright 2025 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+//go:build arm64 || amd64
+
+package cockroachkvs
+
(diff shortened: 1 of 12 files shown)


RaduBerinde (approved):
:lgtm: Very cool!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5093#-:-OVc-G3D5-W_igpQBP3j:b-kkp8rn)* status: 0 of 12 files reviewed, all discussions resolved

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5093#-:-OVcpPDOE8NSM6EKoIii:b-814g8k)* status: 0 of 12 files reviewed, all discussions resolved

PR #5110, opened 2025-07-23
table: automatically widen columns if needed

RaduBerinde:
#### table: change renderValue to return a string

Pull out all the rendering code and simplify the interface to just
return the string.

#### table: automatically widen columns if needed

The table code currently defines a fixed column width and effectively
truncates any value that is longer (except in the last column). It
would be unfortunate to miss some useful part of a metric because of
this.

This change makes the table code automatically widen the columns if
necessary. We also automatically add a vertical space between columns
that don't have a divider.


RaduBerinde:
Just bumping this in case any notifications got lost in the noise last week.

RaduBerinde:
TFTR!

PR #5115, opened 2025-07-27
colblk: add an assertion

RaduBerinde:
This is to help debug #5096.


RaduBerinde:
Never mind, I can repro on gceworker, don't need to check this in.

PR #5116, opened 2025-07-27
colblk: fix 386 build

### LINKED ISSUE #5096: master: nightly linux-32bit test failed
The nightly linux-32bit test run failed on c21fd8d1bcf0119e3657c63f3613d9c90cd1e98c. Please review the run [here](https://github.com/cockroachdb/pebble/actions/runs/16414068396).

RaduBerinde:
When we cast the block metadata, we incorrectly calculate the key
seeker metadata offset without accounting for the possible padding. A
recent change (#5093) caused padding to take effect.

We fix this and make the code simpler and more robust.

Fixes #5096.


RaduBerinde:
TFTR!

PR #5125, opened 2025-07-28
sstable: use correct blockkind for checksum validation of blob ref index

annrpom:
Previously, we were using a read function with `blockkind.Metadata` when we should have passed in the `readBlobRefIndexBlock` read function (the only difference here is the proper blockkind, `blockkind.BlobReferenceValueLivenessIndex`, is used)


annrpom:
TFTR! ('-')7

PR #5126, opened 2025-07-28
block: rework physical block code

RaduBerinde:
We change `PhysicalBlock` to always be backed by a `TempBuffer`; this
simplifies the code related to creating the blocks. The trailer is now
also stored directly in this buffer, and we no longer have to issue
two separate `Write` calls.

Note that we were already using a `TempBuffer` in the most important
write paths (columnar sstable writer and blob value writer).

We add a `PhysicalBlockMaker` type which encompasses the compressor
and the checksummer and a single `Make()` method replaces all
`CompressAndChecksum/CopyAndChecksum` variations.


sumeerbhola (changes requested):
> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5126)* status: all files reviewed, 7 unresolved discussions (waiting on @jbowens)

___
*[`sstable/rowblk_writer.go` line 435 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5126#-OWIF0-T6WqXrk3kkMkc:-OWIF0-T6WqXrk3kkMkd:b-c27jxz) ([raw file](https://github.com/cockroachdb/pebble/blob/7ce2e17d2647c6ef4fed788c1303c82dbc65ef00/sstable/rowblk_writer.go#L435)):*
> ```Go
> 
> 	// physical holds the (possibly) compressed block and its trailer. The
> 	// underlying block data's byte slice is owned by the dataBlockBuf. It  may
> ```

is this ownership comment stale?
___
*[`sstable/rowblk_writer.go` line 1927 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5126#-OWIEJ6y0pCwrlUK-CKp:-OWIEJ6y0pCwrlUK-CKq:b9pgyj6) ([raw file](https://github.com/cockroachdb/pebble/blob/7ce2e17d2647c6ef4fed788c1303c82dbc65ef00/sstable/rowblk_writer.go#L1927)):*
> ```Go
> 	// layout.WriteDataBlock keeps layout.offset up-to-date for us.
> 	// Note that this can mangle the pb data.
> 	bh, err := w.layout.writePhysicalBlock(pb)
> ```

`pb.Release`?
I'm generally struggling with how to identify whether there are missing calls to `Release`, since the higher allocations because of things not being returned to the `sync.Pool` seems non-trivial to test.

For instance, I don't see where `RawRowWriter.rewriteSuffixes` calls release on the PhysicalBlocks in the slice of `blockWithSpan`.
___
*[`sstable/blob/blob.go` line 290 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5126#-OWIDYz0BuxGUnjb9foT:-OWIDYz0BuxGUnjb9foU:bh9og23) ([raw file](https://github.com/cockroachdb/pebble/blob/7ce2e17d2647c6ef4fed788c1303c82dbc65ef00/sstable/blob/blob.go#L290)):*
> ```Go
> 			}
> 			return FileWriterStats{}, err
> 		}
> ```

missing `pb.Release`?
___
*[`sstable/valblk/writer.go` line 167 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5126#-OWICe8bDipiOcMH59xL:-OWICe8bDipiOcMH59xM:b-t2oyzp) ([raw file](https://github.com/cockroachdb/pebble/blob/7ce2e17d2647c6ef4fed788c1303c82dbc65ef00/sstable/valblk/writer.go#L167)):*
> ```Go
> 		panic("incorrect length calculation")
> 	}
> 	pb := w.physBlockMaker.Make(w.buf.Data(), blockkind.Metadata, block.DontCompress)
> ```

is this DontCompress a change from before? I don't recollect doing anything special wrt compression of the value blocks index.
___
*[`sstable/block/physical.go` line 56 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5126#-OWHi2-rAQMNgh9vld6n:-OWHi2-rAQMNgh9vld6o:bo3l0m9) ([raw file](https://github.com/cockroachdb/pebble/blob/7ce2e17d2647c6ef4fed788c1303c82dbc65ef00/sstable/block/physical.go#L56)):*
> ```Go
> 
> // AlreadyEncodedPhysicalBlock creates a PhysicalBlock from the provided on-disk
> // data; used when copying existing blocks.
> ```

Can you add to the comment saying that this will make a copy of `dataWithTrailer`. This is my first time looking at this code, so I don't have an existing mental model, and it isn't clear when we want a shallow `PhysicalBlock` and when we want a deep copy.
___
*[`sstable/block/physical.go` line 82 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5126#-OWHie8d9CAPXv8VHOV9:-OWHie8d9CAPXv8VHOVA:btkli49) ([raw file](https://github.com/cockroachdb/pebble/blob/7ce2e17d2647c6ef4fed788c1303c82dbc65ef00/sstable/block/physical.go#L82)):*
> ```Go
> //		copy(data, b.data)
> //		return PhysicalBlock{data: data, trailer: b.trailer}
> //	}
> ```

Was the above meant to be deleted?
And do we no longer need the ability to allocate using a `bytealloc.A` because using `tempBufferPool` should be good enough?
___
*[`sstable/block/physical.go` line 115 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5126#-OWICIPt3-lqCKjC4-HB:-OWICIPt3-lqCKjC4-HC:b5gci3z) ([raw file](https://github.com/cockroachdb/pebble/blob/7ce2e17d2647c6ef4fed788c1303c82dbc65ef00/sstable/block/physical.go#L115)):*
> ```Go
> const (
> 	NoFlags      PhysicalBlockFlags = 0
> 	DontCompress PhysicalBlockFlags = 1 << (iota - 1)
> ```

will there be other bits set in this mask in the future?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5126)* status: all files reviewed, 8 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/colblk_writer.go` line 886 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5126#-OWIeZzZ5gdFUUTmgZPg:-OWIr2mWE8ie7RmQp9Bp:b-5rez9z) ([raw file](https://github.com/cockroachdb/pebble/blob/ded87d65ced15c0ab3ae18c8825112c088e7f663/sstable/colblk_writer.go#L886)):*

looks like every caller of `WritePrecompressedDataBlock` immediately releases the `PhysicalBlock` after the call. Would it be less error-prone to rename it `WriteAndReleasePrecompressedDataBlock` and release inside it?

Hmm, it seems that every caller of the underlying `layoutWriter.writePhysicalBlock` immediately releases. So perhaps we could transfer ownership in all cases if we chose to.
</blockquote></details>

I considered it, but `Release()` modifies the receiver (and we do want that modification to happen), so we would have to change them all to take a pointer which would make things awkward. Then there is the question of what happens in error cases, which is annoying to consider and document - I guess we could release it in all cases, but it seems a bit non-idiomatic.

One interesting idea would be if `writePhysicalBlock` took a `*TempBuffer` directly and the only way you could obtain that from the `PhysicalBlock` would be to "take" it:
```go
func (b *PhysicalBlock) Take() *TempBuffer {
   tb := b.tb
   b.tb = nil
   return tb
}
```
(or it could be some other type around a `*TempBuffer`, e.g. `OwnedPhysicalBlock`)
___
*[`sstable/block/physical.go` line 56 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5126#-OWHi2-rAQMNgh9vld6n:-OWIrZYH7zLKrG37-yYi:bwduw9n) ([raw file](https://github.com/cockroachdb/pebble/blob/7ce2e17d2647c6ef4fed788c1303c82dbc65ef00/sstable/block/physical.go#L56)):*

Why is `RawColumnWriter.copyDataBlocks` not performance sensitive? If I am reading correctly, that is the caller, and is used by `sstable.CopySpan` for copy compactions.
</blockquote></details>

In practice, we use that to copy a remote file during online restore - I just figured the overhead of a local memory copy (which is probably somewhere in the 5-10GB/s range) is probably insignificant in relation to the actual download.

Also note that the flip side of this change is that we now reuse the read buffer in `copyDataBlocks` so we avoid an allocation and a zeroing of the entire buffer on each iteration.

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5126#-:-OWLjzo38wlMq_qphABg:bg7igkv)* status: all files reviewed, 8 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/colblk_writer.go` line 886 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5126#-OWIeZzZ5gdFUUTmgZPg:-OWLibYO4RD9kqXjhKri:b-brri6v) ([raw file](https://github.com/cockroachdb/pebble/blob/ded87d65ced15c0ab3ae18c8825112c088e7f663/sstable/colblk_writer.go#L886)):*

I considered it, but `Release()` modifies the receiver (and we do want that modification to happen), so we would have to change them all to take a pointer which would make things awkward. Then there is the question of what happens in error cases, which is annoying to consider and document - I guess we could release it in all cases, but it seems a bit non-idiomatic.

One interesting idea would be if `writePhysicalBlock` took a `*TempBuffer` directly and the only way you could obtain that from the `PhysicalBlock` would be to "take" it:
```go
func (b *PhysicalBlock) Take() *TempBuffer {
   tb := b.tb
   b.tb = nil
   return tb
}
```
(or it could be some other type around a `*TempBuffer`, e.g. `OwnedPhysicalBlock`)
</blockquote></details>

If it took a `*TempBuffer` in `writePhysicalBlock`, the `TempBuffer` would need a `LengthWithoutTrailer` method, yes?
In some sense, what you outlined with `Take` may be less error prone. But we also are not very disciplined about using methods within the same package, and given there is no pointer ownership either, I'm worried that I may be asking for supeficial improvements. So up to you whether to do it or not -- I am ok either way.
___
*[`sstable/block/physical.go` line 56 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5126#-OWHi2-rAQMNgh9vld6n:-OWLjxhs-LnRdBUx0F01:bq0ncok) ([raw file](https://github.com/cockroachdb/pebble/blob/7ce2e17d2647c6ef4fed788c1303c82dbc65ef00/sstable/block/physical.go#L56)):*

In practice, we use that to copy a remote file during online restore - I just figured the overhead of a local memory copy (which is probably somewhere in the 5-10GB/s range) is probably insignificant in relation to the actual download.

Also note that the flip side of this change is that we now reuse the read buffer in `copyDataBlocks` so we avoid an allocation and a zeroing of the entire buffer on each iteration.
</blockquote></details>

sounds good

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5126)* status: all files reviewed, 8 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/colblk_writer.go` line 886 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5126#-OWIeZzZ5gdFUUTmgZPg:-OWLs-Dk2VA7EtxXD0S5:b-xehesz) ([raw file](https://github.com/cockroachdb/pebble/blob/ded87d65ced15c0ab3ae18c8825112c088e7f663/sstable/colblk_writer.go#L886)):*

If it took a `*TempBuffer` in `writePhysicalBlock`, the `TempBuffer` would need a `LengthWithoutTrailer` method, yes?
In some sense, what you outlined with `Take` may be less error prone. But we also are not very disciplined about using methods within the same package, and given there is no pointer ownership either, I'm worried that I may be asking for supeficial improvements. So up to you whether to do it or not -- I am ok either way.
</blockquote></details>

`writePhysicalBlock` can just subtract `block.TrailerLen`.

I'll merge this as-is and explore the alternative separately. If I come up with something that feels worth it, I'll post another PR. Thanks!

PR #5130, opened 2025-07-29
sstable: free physical buffers on write

RaduBerinde:
This change moves the responsibility of releasing the `PhysicalBlock`s
to the code that writes out the block. This makes the higher level
code simpler and more robust.

To avoid misuse, we add an `OwnedPhysicalBlock` type which is used to
indicate that once passed, the function takes responsibility for
releasing the block. The only way to create an `OwnedPhysicalBlock`
clears the `PhysicalBlock` so it can't be incorrectly used again.


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5130#-:-OWSQ2Eq41WbDnqIQpBD:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #5131, opened 2025-07-30
db: re-observe L6 in metrics

annrpom:
Previously, our `total` line was overriding the output for L6 in our LSM & COMPACTION metrics. This patch ensures we move our cursor to a new line before writing out our total.


annrpom:
> Helloo L6 my old frieeend 🎶

i've come to observe you once again

TFTR! ('-')7

Issue #5127, opened 2025-07-28
db: blob file corruption results in a panic


jbowens:
Corruption is reported through the `block.ReadEnv`:
```
	// ReportCorruptionFn is called with ReportCorruptionArg and the error
	// whenever an SSTable corruption is detected. The argument is used to avoid
	// allocating a separate function for each object. It returns an error with
	// more details.
	ReportCorruptionFn  func(opaque any, err error) error
	ReportCorruptionArg any
```

We never set `ReportCorruptionArg` for blob files, so we hit the default case here resulting in a panic:
```
// reportCorruption reports a corruption of a TableMetadata or BlobFileMetadata
// to the event listener and also adds a DataCorruptionInfo payload to the error.
func (d *DB) reportCorruption(meta any, err error) error {
	switch meta := meta.(type) {
	case *manifest.TableMetadata:
		return d.reportFileCorruption(base.FileTypeTable, meta.TableBacking.DiskFileNum, meta.UserKeyBounds(), err)
	case *manifest.PhysicalBlobFile:
		// TODO(jackson): Add bounds for blob files.
		return d.reportFileCorruption(base.FileTypeBlob, meta.FileNum, base.UserKeyBounds{}, err)
	default:
		return errors.AssertionFailedf("unknown metadata type: %T", meta)
	}
}
```
Additionally, we should give some thought to that TODO to add or infer bounds.

PR #5135, opened 2025-07-30
db: propagate *PhysicalBlobFile to reportCorruption

jbowens:
If corruption is encountered while reading a blob file, propagate the blob
file's *PhysicalBlobFile metadata struct to DB.reportCorruption. This fixes a
nil pointer that would previously result while handling blob file corruption.

Fixes https://github.com/cockroachdb/pebble/issues/5127.

### DIFF
--- a/blob_rewrite.go
+++ b/blob_rewrite.go
@@ -181,11 +181,12 @@ func (c *blobFileRewriteCompaction) Execute(jobID JobID, d *DB) error {
 			// Now that we have the manifest lock, check if the blob file is
 			// still current. If not, we bubble up ErrCancelledCompaction.
 			v := d.mu.versions.currentVersion()
-			currentDiskFileNum, ok := v.BlobFiles.Lookup(c.input.FileID)
+			currentDiskFile, ok := v.BlobFiles.Lookup(c.input.FileID)
 			if !ok {
 				return versionUpdate{}, errors.Wrapf(ErrCancelledCompaction,
 					"blob file %s became unreferenced", c.input.FileID)
 			}
+			currentDiskFileNum := currentDiskFile.DiskFileNum()
 			// Assert that the current version's disk file number for the blob
 			// matches the one we rewrote. This compaction should be the only
 			// rewrite compaction running for this blob file.
--- a/blob_rewrite_test.go
+++ b/blob_rewrite_test.go
@@ -296,7 +296,7 @@ func TestBlobRewriteRandomized(t *testing.T) {
 			base.MakeInternalKey(keys[i], base.SeqNum(i), base.InternalKeyKindSet),
 			blob.InlineHandle{
 				InlineHandlePreface: blob.InlineHandlePreface{
-					ReferenceID: blob.ReferenceID(0),
+					ReferenceID: base.BlobReferenceID(0),
 					ValueLen:    uint32(len(values[i])),
 				},
 				HandleSuffix: blob.HandleSuffix{
@@ -415,11 +415,11 @@ func TestBlobRewriteRandomized(t *testing.T) {
 	}
 }
 
-// constantFileMapping implements blob.FileMapping and always maps to itself.
+// constantFileMapping implements base.BlobFileMapping and always maps to itself.
 type constantFileMapping base.DiskFileNum
 
-// Assert that (*inputFileMapping) implements blob.FileMapping.
-var _ blob.FileMapping = constantFileMapping(0)
+// Assert that (*inputFileMapping) implements base.BlobFileMapping.
+var _ base.BlobFileMapping = constantFileMapping(0)
 
 func (m constantFileMapping) Lookup(fileID base.BlobFileID) (base.DiskFileNum, bool) {
 	return base.DiskFileNum(m), true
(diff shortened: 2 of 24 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5135)* status: 0 of 24 files reviewed, 4 unresolved discussions (waiting on @annrpom)

___
*[`internal/base/filenames.go` line 68 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5135#-OWRunRZ742ci8kiLR06:-OWRunRZ742ci8kiLR07:bbeo82u) ([raw file](https://github.com/cockroachdb/pebble/blob/79442aac3c8df893253145e1b9fd662f4a508c07/internal/base/filenames.go#L68)):*
> ```Go
> 
> // DiskFile is an interface that provides access to a disk file number.
> type DiskFile interface {
> ```

Consider making this more specific to how we are using it, maybe a `DiskFileObject` that only `TableMetadata` and `BlobMetadata` implement (and update the reporting callback to take a `DiskFileObject` argument).

This interface can contain a `FileInfo() (fileType, fileNum, userKeyBounds)` method and `reportCorruption` can simply call that instead of expecting specific types.
___
*[`internal/base/filenames.go` line 92 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5135#-OWRy4BCD_O-FOXXinYm:-OWRy4BCD_O-FOXXinYn:b-87fuow) ([raw file](https://github.com/cockroachdb/pebble/blob/79442aac3c8df893253145e1b9fd662f4a508c07/internal/base/filenames.go#L92)):*
> ```Go
> }
> 
> // DiskFileNum implements the DiskFile interface.
> ```

This seems error-prone, do we really want to be able to use just a DiskFile, like before?
___
*[`file_cache.go` line 122 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5135#-OWRuHTY19ssqbndiXsi:-OWRuHTZ1h20UMoapfbw:bdik3xr) ([raw file](https://github.com/cockroachdb/pebble/blob/79442aac3c8df893253145e1b9fd662f4a508c07/file_cache.go#L122)):*
> ```Go
> 
> 	// reportCorruptionFn is used for block.ReadEnv.ReportCorruptionFn. It expects
> 	// the first argument to be a `*TableMetadata`. It returns an error that
> ```

[nit] Can you update the comment about the argument
___
*[`file_cache.go` line 390 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5135#-OWRxfaW9jTV-KKIxPR8:-OWRxfaW9jTV-KKIxPR9:b-36szd1) ([raw file](https://github.com/cockroachdb/pebble/blob/79442aac3c8df893253145e1b9fd662f4a508c07/file_cache.go#L390)):*
> ```Go
> // GetValueReader returns a blob.ValueReader for blob file identified by fileNum.
> func (h *fileCacheHandle) GetValueReader(
> 	ctx context.Context, diskFile base.DiskFile,
> ```

Is it now possible to change this to take a `*manifest.PhysicalBlobFile`?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5135)* status: 0 of 28 files reviewed, 2 unresolved discussions (waiting on @annrpom and @RaduBerinde)

___
*[`file_cache.go` line 122 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5135#-OWRuHTY19ssqbndiXsi:-OWb7LY-7gR7s84h7Dmr:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/79442aac3c8df893253145e1b9fd662f4a508c07/file_cache.go#L122)):*

[nit] Can you update the comment about the argument
</blockquote></details>

Done
___
*[`file_cache.go` line 390 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5135#-OWRxfaW9jTV-KKIxPR8:-OWS-cxPCZTqaORgCW7c:b-olxjc2) ([raw file](https://github.com/cockroachdb/pebble/blob/79442aac3c8df893253145e1b9fd662f4a508c07/file_cache.go#L390)):*

Is it now possible to change this to take a `*manifest.PhysicalBlobFile`?
</blockquote></details>

No -- this is satisfying an interface, and the chain of interface and import dependencies prevent it.
___
*[`internal/base/filenames.go` line 68 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5135#-OWRunRZ742ci8kiLR06:-OWb7M_O3QJdutS65LJh:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/79442aac3c8df893253145e1b9fd662f4a508c07/internal/base/filenames.go#L68)):*

Consider making this more specific to how we are using it, maybe a `DiskFileObject` that only `TableMetadata` and `BlobMetadata` implement (and update the reporting callback to take a `DiskFileObject` argument).

This interface can contain a `FileInfo() (fileType, fileNum, userKeyBounds)` method and `reportCorruption` can simply call that instead of expecting specific types.
</blockquote></details>

Done

jbowens:
TFTR!

PR #5137, opened 2025-07-30
sstable: reimplement index block lazy loading in single-level and two-level iterator

xxmplus:
This attempt addresses the data corruption issues in the previous implementation (527eebf4 and fe43e7b5, reverted) successfully.

### What's changed?
The main improvement is to preserve the context at construction time for deferred loading via the new `lazyState` and `topLevelLazyState` structs, respectively. This approach avoids the params being changed between construction and first access and addresses the data corruption issues. Also, proper state management now ensures the iterator can be reused from the pool.

In terms of testing, lazy loading defers I/O operations. While the behavior is transparent to users, the internal semantic is different now. The of events in some data-driven tests have to change accordingly. See changes in `checkpoint`, `cleaner`, and `event_listener`. There are also tests added to demonstrate how bloom filter can benefit from lazy loading by avoid loading index block entirely. See changes in `flushable_ingest`, and `reader_lazy_loading`.
    
Implements #3248


RaduBerinde (commented):
CC @jbowens - please take a look as well given that there were subtle problems with the previous iteration.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5137#-:-OWSnZod7jrfW7DBn8_H:b-za5s2c)* status: 0 of 12 files reviewed, 5 unresolved discussions

___
*[`sstable/reader_iter_single_lvl.go` line 27 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWSmn6H7YJPTFMDNQPM:-OWSmn6H7YJPTFMDNQPN:b82p1q9) ([raw file](https://github.com/cockroachdb/pebble/blob/11ed23ff5913f43f1229567b73c0eb51e3f3683c/sstable/reader_iter_single_lvl.go#L27)):*
> ```Go
> // lazyState holds the state for lazy loading of index blocks.
> type lazyState struct {
> 	// loaded indicates if the index block has been loaded
> ```

[nit] Mention that when an error is hit during index load, `loaded` will be true and `err` will be set
___
*[`sstable/reader_iter_single_lvl.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWSkout3zcAn_RYHUFV:-OWSkout3zcAn_RYHUFW:b-gqvpql) ([raw file](https://github.com/cockroachdb/pebble/blob/11ed23ff5913f43f1229567b73c0eb51e3f3683c/sstable/reader_iter_single_lvl.go#L28)):*
> ```Go
> type lazyState struct {
> 	// loaded indicates if the index block has been loaded
> 	loaded bool
> ```

[nit] I'd separate the "constant" fields from the fields we are updating (i.e. move `loaded` next to `err`)
___
*[`sstable/reader_iter_single_lvl.go` line 36 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWSkkxY1fEbKUvRtvMj:-OWSkkxY1fEbKUvRtvMk:bn5gthu) ([raw file](https://github.com/cockroachdb/pebble/blob/11ed23ff5913f43f1229567b73c0eb51e3f3683c/sstable/reader_iter_single_lvl.go#L36)):*
> ```Go
> 	comparer      *base.Comparer
> 	transforms    blockiter.Transforms
> 	err           error
> ```

[nit] the err field could use a comment
___
*[`sstable/reader_iter_single_lvl.go` line 237 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWSl0BSB5TX1spwlnAR:-OWSl0BSB5TX1spwlnAS:b-afqp0c) ([raw file](https://github.com/cockroachdb/pebble/blob/11ed23ff5913f43f1229567b73c0eb51e3f3683c/sstable/reader_iter_single_lvl.go#L237)):*
> ```Go
> 
> 	// Perserve context for lazy loading.
> 	i.indexLazyState = &lazyState{
> ```

Why does `indexLazyState` need to be a pointer? We need to avoid allocations in these hot paths.
___
*[`sstable/reader_iter_single_lvl.go` line 241 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWSlU2h8LM2j5foAVi2:-OWSlU2h8LM2j5foAVi3:b-mgp066) ([raw file](https://github.com/cockroachdb/pebble/blob/11ed23ff5913f43f1229567b73c0eb51e3f3683c/sstable/reader_iter_single_lvl.go#L241)):*
> ```Go
> 		ctx:           ctx,
> 		reader:        r,
> 		indexFilterRH: i.indexFilterRH,
> ```

Do we need this to store `reader`, `indexFilerRH`, `readEnv`, `transforms`? Can't we use them directly from `i`? (they shouldn't change)  Same for `comparer` which is just `i.reader.Comparer`

Hm.. that doesn't leave much.. Do you know specifically which parameters changing were causing problems in the previous iteration of the change?

xxmplus (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5137)* status: 0 of 12 files reviewed, 5 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/reader_iter_single_lvl.go` line 27 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWSmn6H7YJPTFMDNQPM:-OWWXzxB0f2YqktK738d:b-t948zx) ([raw file](https://github.com/cockroachdb/pebble/blob/11ed23ff5913f43f1229567b73c0eb51e3f3683c/sstable/reader_iter_single_lvl.go#L27)):*

[nit] Mention that when an error is hit during index load, `loaded` will be true and `err` will be set
</blockquote></details>

added.
___
*[`sstable/reader_iter_single_lvl.go` line 28 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWSkout3zcAn_RYHUFV:-OWWY08KDPx_x-n2z9bL:b-ne7jz4) ([raw file](https://github.com/cockroachdb/pebble/blob/11ed23ff5913f43f1229567b73c0eb51e3f3683c/sstable/reader_iter_single_lvl.go#L28)):*

[nit] I'd separate the "constant" fields from the fields we are updating (i.e. move `loaded` next to `err`)
</blockquote></details>

moved.
___
*[`sstable/reader_iter_single_lvl.go` line 36 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWSkkxY1fEbKUvRtvMj:-OWWY2sZ2icC3uApq2Ih:b-t948zx) ([raw file](https://github.com/cockroachdb/pebble/blob/11ed23ff5913f43f1229567b73c0eb51e3f3683c/sstable/reader_iter_single_lvl.go#L36)):*

[nit] the err field could use a comment
</blockquote></details>

added.
___
*[`sstable/reader_iter_single_lvl.go` line 241 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWSlU2h8LM2j5foAVi2:-OWWY7Rq32aZdLGw34JJ:bit5b8x) ([raw file](https://github.com/cockroachdb/pebble/blob/11ed23ff5913f43f1229567b73c0eb51e3f3683c/sstable/reader_iter_single_lvl.go#L241)):*

Do we need this to store `reader`, `indexFilerRH`, `readEnv`, `transforms`? Can't we use them directly from `i`? (they shouldn't change)  Same for `comparer` which is just `i.reader.Comparer`

Hm.. that doesn't leave much.. Do you know specifically which parameters changing were causing problems in the previous iteration of the change?
</blockquote></details>

This is actually a good point. I should verify them individually instead of preserving everything. Let me test it more.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5137#-:-OWbyXi9CDgJF7h92XMK:bnfp4nl)* status: 0 of 12 files reviewed, 9 unresolved discussions (waiting on @jbowens)

___
*[`sstable/data_test.go` line 378 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbxdFU-Z1AQVXf4A0a:-OWbxdFU-Z1AQVXf4A0b:b-83hzgg) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/data_test.go#L378)):*
> ```Go
> 				fmt.Fprintf(&b, "|  index.isDataInvalidated()=%t\n", si.index.IsDataInvalidated())
> 			}
> 			if si != nil {
> ```

[nit] these should all go in a single block
___
*[`sstable/reader_iter_test.go` line 74 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbxyiWAhkz3mxUM0bO:-OWbxyiWAhkz3mxUM0bP:b-lwkrsy) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_test.go#L74)):*
> ```Go
> 			})
> 			if err != nil {
> 				// If construction fails, that's also valid (for non-lazy implementations)
> ```

We no longer have a non-lazy implementation, no?
___
*[`sstable/reader_iter_test.go` line 77 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbyQR0BFSwVoIArHTa:-OWbyQR0BFSwVoIArHTb:bn549n7) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_test.go#L77)):*
> ```Go
> 				continue
> 			}
> 			// With lazy loading, the error should happen on first use
> ```

[nit] `.` at the end
___
*[`sstable/reader_iter_two_lvl.go` line 184 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbwII741rYBkslyi0v:-OWbwII741rYBkslyi0w:b-cdem8x) ([raw file](https://github.com/cockroachdb/pebble/blob/f318b2daaaf8692deb990aaa73ecb1df54d9ca51/sstable/reader_iter_two_lvl.go#L184)):*
> ```Go
> 	}
> 	i.secondLevel.data.InitOnce(r.keySchema, r.Comparer, &i.secondLevel.internalValueConstructor)
> 	topLevelIndexH, err := r.readTopLevelIndexBlock(ctx, i.secondLevel.readEnv.Block, i.secondLevel.indexFilterRH)
> ```

Is there a benefit to lazy loading the top-level block? I think we only expect two-level indexes on L6 which doesn't normally use bloom filters. @jbowens, @sumeerbhola WDYT?
___
*[`sstable/reader_iter_two_lvl.go` line 1063 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbw68w8G24ViTN2wnn:-OWbw68w8G24ViTN2wno:b-r76so3) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_two_lvl.go#L1063)):*
> ```Go
> // Close implements internalIterator.Close, as documented in the pebble
> // package.
> func (i *twoLevelIterator[I, PI, D, PD]) ensureTopLevelIndexLoaded() error {
> ```

This function could set `i.secondLevel.err` instead of every call site doing it.
___
*[`sstable/reader_iter_single_lvl.go` line 175 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbv8kn8inTgnyOZygN:-OWbv8ko9d4X4IZqINY3:b-ml3qgp) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_single_lvl.go#L175)):*
> ```Go
> 	index I
> 	data  D
> ```

[nit] you can consider something like
```
lazyIndexLoad struct {
  enabled bool
  loaded bool
  err error
}
```
___
*[`sstable/reader_iter_single_lvl.go` line 178 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbx-XsA8dhKEIRuGR5:-OWbx-XsA8dhKEIRuGR6:b7et7vn) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_single_lvl.go#L178)):*
> ```Go
> 	// indexLazyEnabled is set to true if the index block is lazy-loaded. This is the
> 	// default for single-level iterators. But it can be false for two-level iterators.
> 	indexLazyEnabled bool
> ```

This is a bit confusing as it suggests the type also supports eager loading. Instead, why not just set `indexLoaded = true` from `loadSecondLevelIndexBlock()`?
___
*[`sstable/reader_iter_single_lvl.go` line 184 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbv1mNCZ-aY4cFQdMy:-OWbv1mNCZ-aY4cFQdMz:bstguvh) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_single_lvl.go#L184)):*
> ```Go
> 	indexLoaded bool
> 	// indexLoadError saves the error from the index block load operation. Note that
> 	// the error is not reset for reuse.
> ```

What does "is not reset for reuse" mean here?
___
*[`sstable/reader_iter_single_lvl.go` line 1626 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbwAAZ-ImVnn26t2bK:-OWbwAAZ-ImVnn26t2bL:b6nuehn) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_single_lvl.go#L1626)):*
> ```Go
> }
> 
> func (i *singleLevelIterator[I, PI, D, PD]) ensureIndexLoaded() error {
> ```

This function could set `i.err` instead of every call site doing it.

xxmplus (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5137)* status: 0 of 12 files reviewed, 9 unresolved discussions (waiting on @jbowens, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 178 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbx-XsA8dhKEIRuGR5:-OWg0dnyAhR36xybpbVJ:bnzmlpl) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_single_lvl.go#L178)):*

This is a bit confusing as it suggests the type also supports eager loading. Instead, why not just set `indexLoaded = true` from `loadSecondLevelIndexBlock()`?
</blockquote></details>

good point, removed it to avoid confusion.
___
*[`sstable/reader_iter_single_lvl.go` line 184 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbv1mNCZ-aY4cFQdMy:-OWg13SK6xIX7ZlLpqvE:brmfzr5) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_single_lvl.go#L184)):*

What does "is not reset for reuse" mean here?
</blockquote></details>

corrected. it should read "Note that the error is not reset when the iterator is reused."
___
*[`sstable/reader_iter_single_lvl.go` line 1626 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbwAAZ-ImVnn26t2bK:-OWhAW1B2gU4d-K5iNHk:b-cypxrc) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_single_lvl.go#L1626)):*

This function could set `i.err` instead of every call site doing it.
</blockquote></details>

updated.
___
*[`sstable/reader_iter_two_lvl.go` line 1063 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbw68w8G24ViTN2wnn:-OWhAY6Q84lW6Og4WG3c:b-cypxrc) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_two_lvl.go#L1063)):*

This function could set `i.secondLevel.err` instead of every call site doing it.
</blockquote></details>

updated.
___
*[`sstable/data_test.go` line 378 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbxdFU-Z1AQVXf4A0a:-OWhA_-fA_-Edd2M9hKr:b-cypxrc) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/data_test.go#L378)):*

[nit] these should all go in a single block
</blockquote></details>

updated.
___
*[`sstable/reader_iter_test.go` line 74 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbxyiWAhkz3mxUM0bO:-OWhAahd6J_Izz1TfTGU:b-ain59p) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_test.go#L74)):*

We no longer have a non-lazy implementation, no?
</blockquote></details>

removed to avoid confusion.
___
*[`sstable/reader_iter_test.go` line 77 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbyQR0BFSwVoIArHTa:-OWhBBi09osGzvFHw_-f:b-t948zx) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_test.go#L77)):*

[nit] `.` at the end
</blockquote></details>

added.

RaduBerinde (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5137)* status: 0 of 12 files reviewed, 6 unresolved discussions (waiting on @jbowens, @sumeerbhola, and @xxmplus)

xxmplus (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5137)* status: 0 of 12 files reviewed, 4 unresolved discussions (waiting on @jbowens, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 184 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbv1mNCZ-aY4cFQdMy:-OWvvHn935kDyvVQPccm:b-6px31z) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_single_lvl.go#L184)):*

`indexLoadError` is removed now.
</blockquote></details>

All absolute positioning methods reset the err (existing code) so that retry is allowed.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5137)* status: 0 of 12 files reviewed, 5 unresolved discussions (waiting on @jbowens, @sumeerbhola, and @xxmplus)

___
*[`sstable/reader_iter_single_lvl.go` line 1595 at r12](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWzWVx2Bl-emLoA3Nka:-OWzWVx2Bl-emLoA3Nkb:b-iglrpg) ([raw file](https://github.com/cockroachdb/pebble/blob/77d72b60d80b091d27174a3c252d2d1600b6ecc8/sstable/reader_iter_single_lvl.go#L1595)):*
> ```Go
> 
> func (i *singleLevelIterator[I, PI, D, PD]) ensureIndexLoaded() bool {
> 	if i.indexLoaded && i.err == nil {
> ```

Why `i.err == nil` here? Won't this mean that if we hit some other error, we'll reload the index? I think just doing `if i.indexLoaded { return true }` should work (and we can remove the `indexLoaded = false` below)

jbowens (commented):
> Reviewed 2 of 10 files at r1, 3 of 7 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5137)* status: 5 of 12 files reviewed, 11 unresolved discussions (waiting on @RaduBerinde, @sumeerbhola, and @xxmplus)

___
*[`sstable/reader_iter_single_lvl.go` line 184 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWbv1mNCZ-aY4cFQdMy:-OWzoSQT1VbMoe-D_ZB5:b-wr23zr) ([raw file](https://github.com/cockroachdb/pebble/blob/a65aa460992802a0b371333301ed4bf3f3c905f2/sstable/reader_iter_single_lvl.go#L184)):*

All absolute positioning methods reset the err (existing code) so that retry is allowed.
</blockquote></details>

i'm confused by this; why is it that we don't want this to be reset when the iterator is reset? the call to `resetForReuse` won't zero `singleLevelIterator.index`, but during `Close()` we'll call `index.Close()` which _will_ zero its state.
___
*[`sstable/reader_iter_single_lvl.go` line 321 at r12](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWzo3I52EGt6Hu-BnOI:-OWzo3I52EGt6Hu-BnOJ:bgdcfva) ([raw file](https://github.com/cockroachdb/pebble/blob/77d72b60d80b091d27174a3c252d2d1600b6ecc8/sstable/reader_iter_single_lvl.go#L321)):*
> ```Go
> 	if !i.ensureIndexLoaded() {
> 		return
> 	}
> ```

are there actually circumstances where we call initBounds and the index block is _not_ loaded?
___
*[`sstable/reader_iter_single_lvl.go` line 351 at r12](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWznxto0zxJS2T_pF8A:-OWznxto0zxJS2T_pF8B:b-2km4zj) ([raw file](https://github.com/cockroachdb/pebble/blob/77d72b60d80b091d27174a3c252d2d1600b6ecc8/sstable/reader_iter_single_lvl.go#L351)):*
> ```Go
> 	if !i.ensureIndexLoaded() {
> 		return
> 	}
> ```

are there actually circumstances where we call initBoundsForAlreadyLoadedBlock and the index block is _not_ loaded?
___
*[`sstable/reader_iter_single_lvl.go` line 700 at r12](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWzjfR86Ycn-BGz_5bk:-OWzjfR86Ycn-BGz_5bl:bv54nb1) ([raw file](https://github.com/cockroachdb/pebble/blob/77d72b60d80b091d27174a3c252d2d1600b6ecc8/sstable/reader_iter_single_lvl.go#L700)):*
> ```Go
> 	if !i.ensureIndexLoaded() {
> 		return nil
> 	}
> ```

this is redundant with the above conditional right?
___
*[`sstable/reader_iter_single_lvl.go` line 757 at r12](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWzjrTj-XVSHJ8H5ZSF:-OWzjrTj-XVSHJ8H5ZSG:bwp6gr6) ([raw file](https://github.com/cockroachdb/pebble/blob/77d72b60d80b091d27174a3c252d2d1600b6ecc8/sstable/reader_iter_single_lvl.go#L757)):*
> ```Go
> 		if !i.ensureIndexLoaded() {
> 			return nil
> 		}
> ```

is this also redundant? I don't see where we would've potentially cleared the loaded index block.
___
*[`sstable/reader_iter_single_lvl.go` line 1044 at r12](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWzmPI82f65Yn69Lzcx:-OWzmPI82f65Yn69Lzcy:bbufmwn) ([raw file](https://github.com/cockroachdb/pebble/blob/77d72b60d80b091d27174a3c252d2d1600b6ecc8/sstable/reader_iter_single_lvl.go#L1044)):*
> ```Go
> 
> 	var dontSeekWithinBlock bool
> 	if !i.ensureIndexLoaded() {
> ```

redundant with the call to `ensureIndexLoaded` a few lines above?
___
*[`sstable/reader_iter_single_lvl.go` line 1074 at r12](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OWzmWmGEruyqDlF3Lei:-OWzmWmGEruyqDlF3Lej:b-7psoge) ([raw file](https://github.com/cockroachdb/pebble/blob/77d72b60d80b091d27174a3c252d2d1600b6ecc8/sstable/reader_iter_single_lvl.go#L1074)):*
> ```Go
> 		// Intersects=false irrespective of the block props provided) during
> 		// seeks.
> 		if !i.ensureIndexLoaded() {
> ```

ditto, also redundant?

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5137#-:-OYgFUrn02wapGoz1kpi:bnfp4nl)* status: 5 of 12 files reviewed, 11 unresolved discussions (waiting on @jbowens, @sumeerbhola, and @xxmplus)

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5137)* status: 5 of 12 files reviewed, 15 unresolved discussions (waiting on @RaduBerinde, @sumeerbhola, and @xxmplus)

___
*[`sstable/reader_iter_test.go` line 539 at r14](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OYhBo7W61ZIMJXaZII8:-OYhBo7W61ZIMJXaZII9:b-7hi1o1) ([raw file](https://github.com/cockroachdb/pebble/blob/2f908908f6776eab5f873a0a0459033af6cb7a17/sstable/reader_iter_test.go#L539)):*
> ```Go
> 			}
> 		case 3:
> 			iter.Next()
> ```

It's not legal to perform a `Next` on an iterator that hasn't been positioned, so I think we need to adjust the test to first perform a `iter.First()` or something, or make the Next conditional on having already performed a seek.
___
*[`sstable/reader_iter_single_lvl.go` line 444 at r14](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OYh9Sj_BcNahcg8zlG0:-OYh9Sja0jwM_tbkEnv9:b27fhh0) ([raw file](https://github.com/cockroachdb/pebble/blob/2f908908f6776eab5f873a0a0459033af6cb7a17/sstable/reader_iter_single_lvl.go#L444)):*
> ```Go
> 		PD(&i.data).Invalidate()
> 		return loadBlockFailed
> 	}
> ```

This should be unnecessary; the caller is required to have positioned the index iterator already so we necessarily must have already loaded the block if the caller is obeying the contract. We could make it an assertion instead:

```
	if !i.indexLoaded {
		i.err = errors.AssertionFailedf("index block is not loaded")
		return loadBlockFailed
	}
```
___
*[`sstable/reader_iter_single_lvl.go` line 1290 at r14](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OYhAA0V37f0QMxBd3ZC:-OYhAA0V37f0QMxBd3ZD:b-63an0y) ([raw file](https://github.com/cockroachdb/pebble/blob/2f908908f6776eab5f873a0a0459033af6cb7a17/sstable/reader_iter_single_lvl.go#L1290)):*
> ```Go
> 	if !i.ensureIndexLoaded() {
> 		return nil
> 	}
> ```

This is also unnecessary in the relative positioning operations (`NextPrefix`,`Next`, `Prev`), because the iterator must've already been seeked. Otherwise it's undefined where we end up, since there is no current position to move relative to. We can make these calls to `ensureIndexLoaded` in `NextPrefix`, `Next` and `Prev` assertion failures too:

```
		if !i.indexLoaded {
			i.err = errors.AssertionFailedf("index block is not loaded")
			return nil
		}
```
___
*[`sstable/reader_iter_single_lvl.go` line 1445 at r14](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OYhCLLT5Jtzinb4CuZA:-OYhCLLT5Jtzinb4CuZB:bdyb96b) ([raw file](https://github.com/cockroachdb/pebble/blob/2f908908f6776eab5f873a0a0459033af6cb7a17/sstable/reader_iter_single_lvl.go#L1445)):*
> ```Go
> 		if !i.ensureIndexLoaded() {
> 			return nil
> 		}
> ```

ditto here and `skipForward`—I think these should be assertions.
___
*[`sstable/reader_iter_two_lvl.go` line 857 at r14](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OYhCahT6LexHrBMX7e5:-OYhCahT6LexHrBMX7e6:bvongay) ([raw file](https://github.com/cockroachdb/pebble/blob/2f908908f6776eab5f873a0a0459033af6cb7a17/sstable/reader_iter_two_lvl.go#L857)):*
> ```Go
> 	if !i.ensureTopLevelIndexLoaded() {
> 		return nil
> 	}
> ```

ditto for the twoLevelIterator--I think

- the relative positioning calls should assert the top-level index has already been loaded (and in this case we should lift the assertion up to above the call to `i.secondLevel.NextPrefix`.)
- the `skipForward` and `skipBackward` funcs can similarly assert

xxmplus (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5137)* status: 5 of 12 files reviewed, 15 unresolved discussions (waiting on @jbowens, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/reader_iter_single_lvl.go` line 444 at r14](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OYh9Sj_BcNahcg8zlG0:-OZBcKSS4Lcye8R2jz6H:b-cypxrc) ([raw file](https://github.com/cockroachdb/pebble/blob/2f908908f6776eab5f873a0a0459033af6cb7a17/sstable/reader_iter_single_lvl.go#L444)):*

This should be unnecessary; the caller is required to have positioned the index iterator already so we necessarily must have already loaded the block if the caller is obeying the contract. We could make it an assertion instead:

```
	if !i.indexLoaded {
		i.err = errors.AssertionFailedf("index block is not loaded")
		return loadBlockFailed
	}
```
</blockquote></details>

updated.
___
*[`sstable/reader_iter_single_lvl.go` line 1290 at r14](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OYhAA0V37f0QMxBd3ZC:-OZBcPDl8G6YU5fuyq_3:b-cypxrc) ([raw file](https://github.com/cockroachdb/pebble/blob/2f908908f6776eab5f873a0a0459033af6cb7a17/sstable/reader_iter_single_lvl.go#L1290)):*

This is also unnecessary in the relative positioning operations (`NextPrefix`,`Next`, `Prev`), because the iterator must've already been seeked. Otherwise it's undefined where we end up, since there is no current position to move relative to. We can make these calls to `ensureIndexLoaded` in `NextPrefix`, `Next` and `Prev` assertion failures too:

```
		if !i.indexLoaded {
			i.err = errors.AssertionFailedf("index block is not loaded")
			return nil
		}
```
</blockquote></details>

updated.
___
*[`sstable/reader_iter_single_lvl.go` line 1445 at r14](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OYhCLLT5Jtzinb4CuZA:-OZBcUc7-Soj0cWv0RkB:b-cypxrc) ([raw file](https://github.com/cockroachdb/pebble/blob/2f908908f6776eab5f873a0a0459033af6cb7a17/sstable/reader_iter_single_lvl.go#L1445)):*

ditto here and `skipForward`—I think these should be assertions.
</blockquote></details>

updated.
___
*[`sstable/reader_iter_test.go` line 539 at r14](https://reviewable.io/reviews/cockroachdb/pebble/5137#-OYhBo7W61ZIMJXaZII8:-OZBqFXh7yVdXP6cWoZj:b-cypxrc) ([raw file](https://github.com/cockroachdb/pebble/blob/2f908908f6776eab5f873a0a0459033af6cb7a17/sstable/reader_iter_test.go#L539)):*

It's not legal to perform a `Next` on an iterator that hasn't been positioned, so I think we need to adjust the test to first perform a `iter.First()` or something, or make the Next conditional on having already performed a seek.
</blockquote></details>

updated.

jbowens (approved):
:lgtm:

> @jbowens reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5137#-:-OZEfoGUCITniZldkIsz:bnfp4nl)* status: 5 of 13 files reviewed, 15 unresolved discussions (waiting on @RaduBerinde, @sumeerbhola, and @xxmplus)

xxmplus:
TFTR!

PR #5138, opened 2025-07-30
crl-release-25.3: sstable: use a single compressor in columnar writer

RaduBerinde:
The columnar writer has its own compressor and the layout writer has
one too. Using both leads to incomplete stats. Consolidate and use
only the layout writer.


RaduBerinde:
TFTR!

PR #5140, opened 2025-07-31
compression: improve compression stats string

RaduBerinde:
Improve the compression stats string to be shorter and more readable:
 - `None` instead of `NoCompression`;
 - remove redundant second number for `None`;
 - order algorithms according to the definition.

We also rename some of the compression constants.

We still support the previous format. The previous format is written
out by v25.3, but is never read back/parsed by that version.


annrpom (approved):
:lgtm:

> Reviewed 46 of 46 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5140#-:-OWWwJTB88wxonDc2IQ1:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #5141, opened 2025-07-31
arenaskl,batchskl: add test ensuring node contains no pointer types

jbowens:
Add a test asserting that the node structs used in both skiplist implementations do not contain any pointer types. We reuse memory backing both of these types without zeroing it. If we introduce a pointer type, we could expose ourselves to an issue similar to the one fixed in #5085: garbage memory could be misinterpreted as a pointer value before the pointer value has been appropriately initialized.

### DIFF
--- a/internal/arenaskl/skl_test.go
+++ b/internal/arenaskl/skl_test.go
@@ -22,6 +22,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"math/rand/v2"
+	"reflect"
 	"strconv"
 	"sync"
 	"sync/atomic"
@@ -29,6 +30,7 @@ import (
 	"time"
 
 	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/testutils"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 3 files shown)


RaduBerinde (approved):
:lgtm: Nice!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5141#-:-OWX8y-U8T-B164EILJ5:b-8gk8uj)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @xinhaoz)

___
*[`internal/testutils/reflect.go` line 45 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5141#-OWX8cXK2srfbG3vx2Ah:-OWX8cXK2srfbG3vx2Ai:b-pygnx1) ([raw file](https://github.com/cockroachdb/pebble/blob/ff34d32573c4e80ea7b482d1e550e2cabfa374e1/internal/testutils/reflect.go#L45)):*
> ```Go
> 
> var kindPointers = []anyPointers{
> 	reflect.Invalid:       kindNoPointer,
> ```

[nit] it would be easier if we just inlined these in the switch (`case refect.Bool, reflect.Int, ..`). The code is already expecting that exactly Struct and Array are "maybe"s.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5141)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @xinhaoz)

___
*[`internal/testutils/reflect.go` line 45 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5141#-OWX8cXK2srfbG3vx2Ah:-OWXfH4JBq7lt-fduMzP:b-l4ccy5) ([raw file](https://github.com/cockroachdb/pebble/blob/ff34d32573c4e80ea7b482d1e550e2cabfa374e1/internal/testutils/reflect.go#L45)):*

[nit] it would be easier if we just inlined these in the switch (`case refect.Bool, reflect.Int, ..`). The code is already expecting that exactly Struct and Array are "maybe"s.
</blockquote></details>

hehe I had that at first but I didn't like the formatting when you have so many variants in the same case statement

jbowens:
TFTRs!

PR #5144, opened 2025-08-02
metrics: move Compression to Table

RaduBerinde:
#### metrics: move Compression to Table

It will be useful to expose separate compression metrics for sstable
and blob value data; move `Metrics.Compression` to `Table` and add a
TODO to add one for blob files.

#### block: minor change to CompressionRatio()

Return 0 if there is no data. This won't make a difference to Pebble,
but it enables usage from CRDB when populating metrics.


RaduBerinde:
TFTR!

PR #5146, opened 2025-08-04
blob: add FileFormatV2 which has a properties block

RaduBerinde:
We extend the blob file format by adding a new V2 footer which encodes
a properties block handle. The V2 footer precedes the V1 footer when
the version in that footer is >= 2. The properties block contains a
key/value for compression statistics.

### DIFF
--- a/blob_rewrite.go
+++ b/blob_rewrite.go
@@ -291,7 +291,7 @@ func (d *DB) runBlobFileRewriteLocked(
 		env,
 		objMeta.DiskFileNum,
 		writable,
-		d.opts.MakeBlobWriterOptions(6),
+		d.opts.MakeBlobWriterOptions(6, d.BlobFileFormat()),
 		c.referencingTables,
 		c.input,
 	)
--- a/data_test.go
+++ b/data_test.go
@@ -1163,12 +1163,18 @@ func runDBDefineCmdReuseFS(td *datadriven.TestData, opts *Options) (*DB, error) 
 		return nil, err
 	}
 
-	if len(ve.NewTables) > 0 {
-		// Collect any blob files created.
-		fileStats, err := valueSeparator.bv.WriteFiles(func(fileNum base.DiskFileNum) (objstorage.Writable, error) {
+	if len(ve.NewTables) == 0 {
+		return d, nil
+	}
+
+	// Collect any blob files created.
+	if !valueSeparator.bv.IsEmpty() {
+		newBlobObject := func(fileNum base.DiskFileNum) (objstorage.Writable, error) {
 			writable, _, err := d.objProvider.Create(context.Background(), base.FileTypeBlob, fileNum, objstorage.CreateOptions{})
 			return writable, err
-		}, d.opts.MakeBlobWriterOptions(0))
+		}
+		blobWriterOpts := d.opts.MakeBlobWriterOptions(0, d.BlobFileFormat())
+		fileStats, err := valueSeparator.bv.WriteFiles(newBlobObject, blobWriterOpts)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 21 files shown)


RaduBerinde:
*[`sstable/blob/blob.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5146#-OWpSQjo8iAhDzQMK25A:-OWpaJLUEpBvudOkEa4m:bvlb1ju) ([raw file](https://github.com/cockroachdb/pebble/blob/8c330b95a26f2b132a909fa303b1791de24b5a69/sstable/blob/blob.go#L47)):*

This is a drive-by comment since I haven't read the code.

This approach is different from what we do for sstable footers, where we distinguish the various footers with a different magic number and read enough bytes at the end of the file corresponding to the maximum footer length. Also, we've not had much need to update that footer since once one has a metaindex block, one can do various format extensions without changing the footer.
That approach, though clunkier in some ways, maybe better, since we don't need a second CRC. Also, while we are at it, should we throw in support for a metaindex block, which can initially say it isn't present? That may mean never needing footer v3.
</blockquote></details>

Ah, right. I like that, will do

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5146)* status: 0 of 21 files reviewed, 1 unresolved discussion (waiting on @annrpom, @jbowens, and @sumeerbhola)

___
*[`sstable/blob/blob.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5146#-OWpSQjo8iAhDzQMK25A:-OWpu4kt3yT5EJ-BqqaW:b-gjyjoa) ([raw file](https://github.com/cockroachdb/pebble/blob/8c330b95a26f2b132a909fa303b1791de24b5a69/sstable/blob/blob.go#L47)):*

Ah, right. I like that, will do
</blockquote></details>

Done. I don't know the story behind the crab, but I tried to stay in theme

sumeerbhola (approved):
:lgtm: 

> Reviewed 14 of 21 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5146#-:-OWuxagWEAjxJD2FKX-o:bg7igkv)* status: all files reviewed, 5 unresolved discussions (waiting on @annrpom, @jbowens, and @RaduBerinde)

___
*[`sstable/blob/blob.go` line 47 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5146#-OWpSQjo8iAhDzQMK25A:-OWuuSNcDFAl-nRVNQ_N:b32vhw) ([raw file](https://github.com/cockroachdb/pebble/blob/8c330b95a26f2b132a909fa303b1791de24b5a69/sstable/blob/blob.go#L47)):*

Done. I don't know the story behind the crab, but I tried to stay in theme
</blockquote></details>

:)
___
*[`sstable/blob/blob.go` line 47 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5146#-OWuqBGk7T9krFF7q2Xj:-OWuqBGk7T9krFF7q2Xk:bdhkyk3) ([raw file](https://github.com/cockroachdb/pebble/blob/0399108f0cafbf1eabe83454095cfb35914dd927/sstable/blob/blob.go#L47)):*
> ```Go
> 	FileFormatV1 FileFormat = 1
> 	// FileFormatV2 adds a property block. The property block offset and length
> 	// are encoded in a separate v2 footer that comes before the v1 file footer.
> ```

This comment is stale wrt "separate v2 footer that comes before ...".

Also, consider mentioning the metaindex block in the comment too.
___
*[`sstable/blob/blob.go` line 366 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5146#-OWus5tqC08zLRQw3oH6:-OWus5tqC08zLRQw3oH7:b-honec5) ([raw file](https://github.com/cockroachdb/pebble/blob/0399108f0cafbf1eabe83454095cfb35914dd927/sstable/blob/blob.go#L366)):*
> ```Go
> 	w.valuesEncoder.Reset()
> 	w.w = nil
> 	stats := w.stats
> ```

nit: can you lift this non-cleanup line `stats := w.stats` above this cleanup section of the code.
___
*[`sstable/blob/blob.go` line 390 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5146#-OWusNwrEOA_KdaxmzkA:-OWusNws9OaF0wMPXZB1:bigb1ll) ([raw file](https://github.com/cockroachdb/pebble/blob/0399108f0cafbf1eabe83454095cfb35914dd927/sstable/blob/blob.go#L390)):*
> ```Go
> 
> // fileFooter contains the information contained within the footer of a blob
> // file. See fileFooterLengthV1 an fileFooterLengthV2 for the footer format.
> ```

s/an/and/
___
*[`sstable/blob/blob.go` line 504 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5146#-OWuvhKY3C9juMX9_QbK:-OWuvhKY3C9juMX9_QbL:b7vytxn) ([raw file](https://github.com/cockroachdb/pebble/blob/0399108f0cafbf1eabe83454095cfb35914dd927/sstable/blob/blob.go#L504)):*
> ```Go
> 	fileNum := ro.CacheOpts.FileNum
> 
> 	footerBuf := make([]byte, fileFooterLengthV2)
> ```

why the heap allocation instead of having a longer array of `fileFooterLengthV2`, since we're going to just use this temporarily to decode?

PR #5148, opened 2025-08-04
db: fix bug in scaling compression stats for virtual tables

RaduBerinde:
#### db: fix bug in scaling compression stats for virtual tables

A copy/pasta. I found this while working on a separate change that
makes the API around table stats much less fragile.

### DIFF
--- a/data_test.go
+++ b/data_test.go
@@ -1253,6 +1253,10 @@ func runTableStatsCmd(td *datadriven.TestData, d *DB) string {
 				fmt.Fprintf(&b, "tombstone-dense-blocks-ratio: %0.1f\n", f.Stats.TombstoneDenseBlocksRatio)
 			}
 
+			if !f.Stats.CompressionStats.IsEmpty() {
+				fmt.Fprintf(&b, "compression: %s\n", f.Stats.CompressionStats.String())
+			}
+
 			return b.String()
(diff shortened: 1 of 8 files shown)


annrpom (approved):
:lgtm:

> Reviewed 7 of 7 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5148#-:-OWqtx0r2ukH8OpfzAyZ:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`testdata/table_stats` line 986 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5148#-OWqtqANDMJXYywQUsol:-OWqtqANDMJXYywQUsom:bb7lc5i) ([raw file](https://github.com/cockroachdb/pebble/blob/2e120468a0d4c30ab801cee5f4ccc80a31201445/testdata/table_stats#L986)):*
> ```
> 
> # Use a small block-size so that excising the table results in significantly
> # virtual tables which cover significantly fewer blocks.
> ```

nit:



_Suggestion:_
```
# Use a small block-size so that excising the table results in
# virtual tables which cover significantly fewer blocks.
```

PR #5149, opened 2025-08-04
manifest: improve API around TableStats

RaduBerinde:
Using `TableMetadata.TableStats` is very fragile; any code that uses
it must first check that the stats are valid. Code that doesn't do so
relies on subtle higher level synchronization assumptions.

We improve this with a simple change: we unexport the stats and add an
getter which only returns the `*TableStats` if they have been
populated, and a setter with can be called at most once.

All access through these methods is safe (without assumptions about
any external locks), as long as we populate the stats only once.


RaduBerinde:
TFTR!

PR #5150, opened 2025-08-05
sstable: remove CompressionOptions

RaduBerinde:
This property is never used; we just write a fixed value for RocksDB
compatibility. This commit removes the property and writes it out in a
special code path with other rocksdb-speific properties, only when the
table format is an old rocksdb version.


RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5150#-:-OWusigPC2zDPW8aEaXh:b-814g8k)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sumeerbhola)

PR #5151, opened 2025-08-05
go.mod: require go1.24

RaduBerinde:
Cockroach now uses 1.24.5; change go.mod to use the same version and
stop testing nightlies on 1.23.


RaduBerinde:
TFTR!

PR #5156, opened 2025-08-07
block: don't obtain decompressor twice

RaduBerinde:
When we read, we use `DecompressedLen` and `DecompressedInto`
wrappers; each one gets a decompressor then closes it. This is more
than a no-op for Zstd, where we use a pool.

This change removes these wrappers so the caller can reuse the
decompressor against both calls. We now also correctly mark a
`DecompressedLen` error as corruption.

This should be a very minor perf improvement (on the ballpark of 0.05%
for TPCC).

### DIFF
--- a/sstable/block/block.go
+++ b/sstable/block/block.go
@@ -482,18 +482,22 @@ func (r *Reader) doRead(
 	if typ == NoCompressionIndicator {
 		decompressed = compressed
 	} else {
+		decompressor := GetDecompressor(typ)
 		// Decode the length of the decompressed value.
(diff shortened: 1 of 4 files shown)


jbowens (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5156#-:-OX4DDewEzzjiidDVWuu:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #5159, opened 2025-08-07
db: take ScanInternal options through an options struct

jbowens:
Update ScanInternal methods to take options via the (now exported) ScanInternalOptions struct. Previously the function signature was a bit unwieldy.


jbowens:
TFTR!

PR #5160, opened 2025-08-07
db: separate mvcc garbage into blob files

annrpom:
This patch separates SET keys that have already appeared in the sstable into a blob file.


RaduBerinde:
*[`internal/compact/run.go` line 137 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5160#-OXUFf9d3axAlUJIFyhu:-OXUFf9d3axAlUJIFyhv:b45z8fj) ([raw file](https://github.com/cockroachdb/pebble/blob/55bac997ee24c349888c86a953efe34318632f85/internal/compact/run.go#L137)):*
> ```Go
> 	// Add adds the provided key-value pair to the provided sstable writer,
> 	// possibly separating the value into a blob file.
> 	Add(tw sstable.RawWriter, kv *base.InternalKV, forceObsolete bool, isMVCCKey bool) error
> ```

This needs to be `isLikelyMVCCGarbage` not `isMVCCKey` no?

sumeerbhola (changes requested):
> Reviewed 1 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5160)* status: 0 of 15 files reviewed, 4 unresolved discussions (waiting on @annrpom and @jbowens)

___
*[`value_separation.go` line 175 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5160#-OXUugzC129ud0cj_w7w:-OXUugzC129ud0cj_w7x:bldczhe) ([raw file](https://github.com/cockroachdb/pebble/blob/55bac997ee24c349888c86a953efe34318632f85/value_separation.go#L175)):*
> ```Go
> 
> // writeNewBlobFiles implements the strategy and mechanics for separating values
> // into external blob files. We will always separate MVCC garbage values into
> ```

- I think this should say "potential MVCC garbage values" since we can't be sure that what we are separating is actually garbage (the more recent SET may have been DEL'd in a higher level since the transaction aborted).
- I prefer pointing to the comment in RawColumnWriter.evaluatePoint for the exact criteria we use to decide on potential garbage, since "any SET key that has prefix that we have encountered before" is too strong e.g. we don't do it if the previous was a DEL, of for key spans with `!SpanPolicy.DisableValueSeparationBySuffix`.
___
*[`value_separation.go` line 256 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5160#-OXUwyLP63QruNlfUqhU:-OXUwyLP63QruNlfUqhV:b-mnchm8) ([raw file](https://github.com/cockroachdb/pebble/blob/55bac997ee24c349888c86a953efe34318632f85/value_separation.go#L256)):*
> ```Go
> 	//
> 	// TODO(annie): Also allow SetWithDelete keys to be separated.
> 	isSeparableMVCCKey := isMVCCKey && len(v) != 0 && kv.K.Kind() == base.InternalKeyKindSet
> ```

I am confused by this predicate since it doesn't say anything about the preceding key and this being an older version.

edit: say @RaduBerinde's comment. Looks like it is a parameter naming issue.
___
*[`internal/compact/run.go` line 540 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5160#-OXUyJHFAxuQZSGDrLio:-OXUyJHGEhZ3GP3dnWXL:bi0gykh) ([raw file](https://github.com/cockroachdb/pebble/blob/55bac997ee24c349888c86a953efe34318632f85/internal/compact/run.go#L540)):*
> ```Go
> 	split := d.split(key.UserKey)
> 	if d.lastKeyPrefix == nil {
> 		d.lastKeyPrefix = key.UserKey[:split]
> ```

Is the memory backing `key.UserKey` stable? If not, we will need to make a copy of the prefix.
I wonder if there is a way to lift out what we do in `KeyWriter.ComparePrev`, or make the `RawWriter` perform that comparison first in a separate method and give back part of the result and stash it for its own future use. Then we can use that to make this determination and don't need to duplicate the key comparison or make a copy. There will be false negatives when we cross output file boundaries, i.e., we won't consider something garbage that is garbage, but that will be rare.

Also, its unfortunate that we are duplicating this determination in this new code and in the existing `RawColumnWriter.evaluatePoint`. Minimally, we should have a helper function that we can call from both.

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5160)* status: all files reviewed, 9 unresolved discussions (waiting on @jbowens, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk_writer.go` line 1283 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5160#-OXzgsOH3ChnXNxRmLCk:-OY1l8QtCV-grFOWcsOS:b-g2oy3b) ([raw file](https://github.com/cockroachdb/pebble/blob/1c60477a061b7d9c518cdc17afce699cbdf7b262/sstable/colblk_writer.go#L1283)):*

This "If there are bounds ..." bullet is stale, and should be removed. I think it predates the `SpanPolicy` stuff which is now evaluating this condition in higher layer code and adjusting `DisableValueBlocks`.
</blockquote></details>

ah oki done
___
*[`testdata/compaction/l0_to_lbase_compaction` line 22 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5160#-OXzhuUf1Sym-6oxCrpi:-OY1l_2c0RqKKXpuLGjh:b-23k9l5) ([raw file](https://github.com/cockroachdb/pebble/blob/eaca9d0cc5eabb9fea19431e80ac997a0a3af5b4/testdata/compaction/l0_to_lbase_compaction#L22)):*

why did this test result change?
</blockquote></details>

hrm good question - i think the change to override our comparer to `testkeys.Comparer` caused a bunch of these changes. i ended up creating a new datadriven test file so that we can separate our usage of this comparer as i'm not 100% sure of how much of our existing tests would actually exercise mvcc garbage blob behavior
___
*[`testdata/compaction/l0_to_lbase_compaction` line 29 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5160#-OXzhxcr6WUG_OedjwSj:-OY1x2Ar1BlZrffOy5yE:bj87hls) ([raw file](https://github.com/cockroachdb/pebble/blob/eaca9d0cc5eabb9fea19431e80ac997a0a3af5b4/testdata/compaction/l0_to_lbase_compaction#L29)):*

ditto
</blockquote></details>

ditto to `testkeys.Comparer` comment
___
*[`testdata/compaction/value_separation` line 604 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5160#-OXziUmJBf-tA51SUc_q:-OY1x8Fs6u5glNWDQ-dG:bj87hls) ([raw file](https://github.com/cockroachdb/pebble/blob/eaca9d0cc5eabb9fea19431e80ac997a0a3af5b4/testdata/compaction/value_separation#L604)):*

why did blob file 42 decrease significantly in size?
</blockquote></details>

ditto to `testkeys.Comparer` comment

sumeerbhola (approved):
:lgtm: 

> Reviewed 10 of 11 files at r5.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5160#-:-OY2bi1j4HzRE6cHVE41:bg7igkv)* status: 18 of 19 files reviewed, 9 unresolved discussions (waiting on @annrpom, @jbowens, and @RaduBerinde)

___
*[`testdata/compaction/l0_to_lbase_compaction` line 22 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5160#-OXzhuUf1Sym-6oxCrpi:-OY2b6RiAWoW2ugyLTzy:be4c64n) ([raw file](https://github.com/cockroachdb/pebble/blob/eaca9d0cc5eabb9fea19431e80ac997a0a3af5b4/testdata/compaction/l0_to_lbase_compaction#L22)):*

hrm good question - i think the change to override our comparer to `testkeys.Comparer` caused a bunch of these changes. i ended up creating a new datadriven test file so that we can separate our usage of this comparer as i'm not 100% sure of how much of our existing tests would actually exercise mvcc garbage blob behavior
</blockquote></details>

btw, I wasn't suggesting to undo the comparer change in our in-person conversation this morning. But this is fine too.
Thanks.

sumeerbhola:
The test failure is unrelated and will be fixed by https://github.com/cockroachdb/pebble/pull/5212

annrpom:
TFTRs! ('-')7

PR #5162, opened 2025-08-07
block,cache: trace and stats for wait duration (for concurrent reads)

sumeerbhola:
cache.Handle.GetWithReadHandle returns a non-zero wait duration on
a cache miss, which resulted in the request waiting for a turn
to do the read, or to wait for the block to be read by another
reader.

Tracing now happens on high wait duration. Wait duration is included
in iterator stats, since this is a block miss. And wait duration is
included in the read duration when the block was read by the caller
after waiting.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -30,7 +30,6 @@ import (
 	"github.com/cockroachdb/pebble/internal/testutils"
 	"github.com/cockroachdb/pebble/objstorage/objstorageprovider"
 	"github.com/cockroachdb/pebble/sstable"
-	"github.com/cockroachdb/pebble/sstable/block"
 	"github.com/cockroachdb/pebble/vfs"
 	"github.com/cockroachdb/pebble/vfs/errorfs"
 	"github.com/cockroachdb/pebble/wal"
@@ -1493,7 +1492,7 @@ func (t *testTracer) IsTracingEnabled(ctx context.Context) bool {
 }
 
 func TestTracing(t *testing.T) {
-	defer block.DeterministicReadBlockDurationForTesting()()
+	defer base.DeterministicReadDurationForTesting()()
 
 	var tracer testTracer
 	buf := &tracer.buf
--- a/internal/base/stopwatch.go
+++ b/internal/base/stopwatch.go
@@ -1,0 +1,43 @@
+// Copyright 2025 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package base
+
+import (
+	"time"
+
+	"github.com/cockroachdb/crlib/crtime"
+)
+
+// DeterministicReadDurationForTesting is for tests that want a deterministic
+// value of the time to read (that is not in the cache). The return value is a
+// function that must be called before the test exits.
+func DeterministicReadDurationForTesting() func() {
+	drbdForTesting := deterministicReadDurationForTesting
+	deterministicReadDurationForTesting = true
+	return func() {
+		deterministicReadDurationForTesting = drbdForTesting
+	}
+}
+
+var deterministicReadDurationForTesting = false
+
+type deterministicStopwatchForTesting struct {
+	startTime crtime.Mono
+}
+
+func MakeStopwatch() deterministicStopwatchForTesting {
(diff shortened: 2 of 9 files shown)


sumeerbhola (commented):
Added another commit

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5162#-:-OX63AcyDP_I_ZX__sAa:b4iwji1)* status: 0 of 9 files reviewed, all discussions resolved (waiting on @jbowens)

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5162#-:-OX7XxLh4Q4bcyCaPwdD:bnfp4nl)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`internal/cache/cache.go` line 291 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5162#-OX7XGOHCWszlQlBuBEk:-OX7XGOHCWszlQlBuBEl:b-tfh33m) ([raw file](https://github.com/cockroachdb/pebble/blob/8ac4385864c2f9e22faa4dfc7bece5e91c93be56/internal/cache/cache.go#L291)):*
> ```Go
> // errors could have occurred prior to this call. But it serves as a rough
> // indicator of whether turn taking could have caused higher latency due to
> // context cancellation of other readers. The waitDuration returns the
> ```

[supernit] I'd move this at the end of the previous paragraph.
___
*[`sstable/block/block.go` line 385 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5162#-OX7XPuW5QJaUFhpB7mt:-OX7XPuW5QJaUFhpB7mu:bkpfcpd) ([raw file](https://github.com/cockroachdb/pebble/blob/8ac4385864c2f9e22faa4dfc7bece5e91c93be56/sstable/block/block.go#L385)):*
> ```Go
> 	if (errorDuration > slowDur || waitDuration > slowDur) && r.opts.LoggerAndTracer.IsTracingEnabled(ctx) {
> 		r.opts.LoggerAndTracer.Eventf(
> 			ctx, "waited for turn for %s when %s time wasted by failed reads", waitDuration.String(),
> ```

[nit] I think separate events would be better, we don't want "failed reads" to show up when there were no errors

sumeerbhola (commented):
TFTR!

Squashed the commits since they are intertwined

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5162#-:-OX8rwY7A9DO2MonBTFO:b-q6euj6)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`internal/cache/cache.go` line 291 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5162#-OX7XGOHCWszlQlBuBEk:-OX8clnF2XHP3iY1LpZI:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/8ac4385864c2f9e22faa4dfc7bece5e91c93be56/internal/cache/cache.go#L291)):*

[supernit] I'd move this at the end of the previous paragraph.
</blockquote></details>

Done
___
*[`sstable/block/block.go` line 385 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5162#-OX7XPuW5QJaUFhpB7mt:-OX8fPtzD0KO1DTeswmv:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/8ac4385864c2f9e22faa4dfc7bece5e91c93be56/sstable/block/block.go#L385)):*

[nit] I think separate events would be better, we don't want "failed reads" to show up when there were no errors
</blockquote></details>

Done
___
*[`sstable/block/block.go` line 471 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5162#-OX7auc9E9IPME1q4s3g:-OX8g9kU8F562FpOafe2:bi8w07u) ([raw file](https://github.com/cockroachdb/pebble/blob/741225f1be17aa9b0103e937a65a18ab8a8fd49d/sstable/block/block.go#L471)):*

Let's show the `kind` too. Maybe we can then remove the callers, their point was mainly to infer the kind of block.
</blockquote></details>

Added the kind. I've kept the callers since the rarity of this probably means any extra information is ok.

PR #5163, opened 2025-08-08
blockkind: remove the Index Kind

sumeerbhola:
It is used in the same way as the SSTableIndex Kind, and its entry in the kindString was never populated.


sumeerbhola:
TFTR!

PR #5164, opened 2025-08-08
db: add initFileOpts for initializing in the FileCache

sumeerbhola:
initFileOpts only contains InitFileReadStats, which allows the caller to provide stats collected by iterators. Reads done when creating a Reader specifically because an iterator was needed on that file, will now be captured in the corresponding iterator stats.

This reduces the uncertainty when we see slowness in traces around iteration, and are unsure whether it happened prior to operations on the iterator. We continue to have some unaccounted latency:
- Footer reads are not included in the stats.
- File cache entry initialization is shared when there are concurrent attempts to create the entry, and the latency (via iterator stats) is only accounted in the one that did the Reader creation.

initFileOpts are also plumbed to valblk.blockProviderWhenClosed, when
it needs to open a sstable.Reader. As part of this change, we also
fix the missing stats for valblk.ReadValueBlockExternal, and the
missing category stats in general when reading value blocks.

### DIFF
--- a/file_cache.go
+++ b/file_cache.go
@@ -205,7 +205,7 @@ func (h *fileCacheHandle) Close() error {
 
 // openFile is called when we insert a new entry in the file cache.
 func (h *fileCacheHandle) openFile(
-	ctx context.Context, fileNum base.DiskFileNum, fileType base.FileType,
+	ctx context.Context, fileNum base.DiskFileNum, fileType base.FileType, opts initFileOpts,
 ) (io.Closer, objstorage.ObjectMetadata, error) {
 	f, err := h.objProvider.OpenForReading(
 		ctx, fileType, fileNum, objstorage.OpenOptions{MustExist: true},
@@ -223,6 +223,7 @@ func (h *fileCacheHandle) openFile(
 		CacheHandle: h.blockCacheHandle,
 		FileNum:     fileNum,
 	}
+	o.InitFileReadStats = opts.stats
 	switch fileType {
 	case base.FileTypeTable:
 		r, err := sstable.NewReader(ctx, f, o)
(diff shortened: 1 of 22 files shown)


sumeerbhola (commented):
also related to https://cockroachlabs.slack.com/archives/C04U1BTF8/p1754564947355449

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5164#-:-OX9PGLq-8VO_3sZA8oE:bgqmqso)* status: 0 of 15 files reviewed, all discussions resolved (waiting on @RaduBerinde)

jbowens (approved):
:lgtm:

> Reviewed 13 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5164#-:-OXdECsZ8aSfGm4JaIxk:bnfp4nl)* status: 13 of 15 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`file_cache.go` line 283 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5164#-OXdDzya6kTrwv-KxbXd:-OXdDzya6kTrwv-KxbXe:bcnczmd) ([raw file](https://github.com/cockroachdb/pebble/blob/75ee25f7e9b04308fe5fc70a90ddb1e7b49e3eb7/file_cache.go#L283)):*
> ```Go
> 		fileType: ftyp,
> 	}
> 	valRef, err := h.fileCache.c.FindOrCreate(ctx, key, initFileOpts{})
> ```

relates to my other question, but should we collect these stats for blob file opens too--or leave a TODO to do so?
___
*[`sstable/options.go` line 138 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5164#-OXdDeED-CCuWthZh-Bb:-OXdDeEE6lDypqB_8d5Z:b-t32l3u) ([raw file](https://github.com/cockroachdb/pebble/blob/75ee25f7e9b04308fe5fc70a90ddb1e7b49e3eb7/sstable/options.go#L138)):*
> ```Go
> // and we are able to capture these initialization stats in the iterator
> // stats.
> type InitFileReadStats struct {
> ```

should we put this in the `block` package and collect it for blob file opens too?

PR #5169, opened 2025-08-11
wal: hint sequential reads when reading WAL files

jbowens:
When reading WAL files, pass the vfs.SequentialReadsOption option to indicate that reads will be sequential. This enables OS-mediated readahead on Linux.


xinhaoz (approved):

jbowens:
TFTR!

PR #5177, opened 2025-08-12
db: ensure checkpoints copy a consistent snapshot

jbowens:
Adapt Checkpoint to copy a consistent snapshot of the database. Previously Checkpoint acquired a Version, ensuring a consistent snapshot of the sstables within the version, but copied the WALs arbitrarily. This allowed Checkpoint to race with inflight WAL writes and allowed for a few sources of inconsistency in checkpoints:

- The checkpointing goroutine racing with the LogWriter could copy an old block from a WAL file's recycled history, followed by a new block containing recent writes. If a chunk in the new block indicated that the old block should've been durably synced, the WAL was considered corrupt on replay when the checkpoint was opened.
- The checkpointing goroutine could copy WAL records containing batches that were committed after version edits (eg, ingests, excises) that were committed after the checkpointed Version. This would violate the checkpoint's consistency.

Now Checkpoint reads the current visible sequence number and only copies the prefix of the WAL that is visible at the sequence number. This ensures that a Checkpoint's snapshot isolation semantics mirror that of an Iterator.


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5177#-:-OXUsASR-KbvPD2yugXs:bnfp4nl)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @annrpom)

___
*[`wal/reader.go` line 392 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5177#-OXUrEBJCXNcO08RxVCt:-OXUrEBJCXNcO08RxVCu:bpxwt6g) ([raw file](https://github.com/cockroachdb/pebble/blob/084b89b21ad5c1f032f1771f422b2dc413994450/wal/reader.go#L392)):*
> ```Go
> // Copy does NOT sync the destination directory, and the caller must explicitly
> // sync it if they require the new WAL file to be durably linked.
> func Copy(
> ```

Should we mention that we allow the logical log to be updated synchronously?
___
*[`wal/reader.go` line 415 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5177#-OXUrMZG2XVQhlu_xx7T:-OXUrMZG2XVQhlu_xx7U:b-9l6md4) ([raw file](https://github.com/cockroachdb/pebble/blob/084b89b21ad5c1f032f1771f422b2dc413994450/wal/reader.go#L415)):*
> ```Go
> 		// A batch assigns sequence numbers beginning at the header's SeqNum,
> 		// increasing for successive internal keys. We only copy the batch if
> 		// the entirety of the batch's keys are visible.
> ```

Is it possible to have a batch that has some visible and some non-visible keys? There would be no way to get an "atomic" WAL for the given sequence key.. Should we error out or assert in this case?
___
*[`wal/reader.go` line 428 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5177#-OXUs12W23ZEzQq6TESo:-OXUs12W23ZEzQq6TESp:busqpr6) ([raw file](https://github.com/cockroachdb/pebble/blob/084b89b21ad5c1f032f1771f422b2dc413994450/wal/reader.go#L428)):*
> ```Go
> 		}
> 	}
> 	if err := w.Close(); err != nil {
> ```

[nit] Add a comment mentioning that `Close()` syncs the file

PR #5179, opened 2025-08-12
[tiered-prototype] db: partial code for new column format

sumeerbhola:
ColumnFormatv2 adds tiering (spanId, attribute) columns. These columns are 
native to Pebble, and not part of the code in cockroachkvs.go, since after 
initial extraction, these will be used internally to Pebble for tiering.

There is a long comment in data_block.go explaining the tentative choice
to have a spanID column (which will be revisited later). The corresponding
sstable formats and format major version is also added.

InternalKV is enhanced with a KVMeta field to carry these columns through
the iterator tree. This is just a prototyping choice as explained in the
todo.

The code changes are all just plumbing. The new format is not yet tested.

### DIFF
--- a/blob_rewrite_test.go
+++ b/blob_rewrite_test.go
@@ -292,21 +292,16 @@ func TestBlobRewriteRandomized(t *testing.T) {
 		tw := sstable.NewRawWriter(w, sstable.WriterOptions{
 			TableFormat: sstable.TableFormatMax,
 		})
-		require.NoError(t, tw.AddWithBlobHandle(
-			base.MakeInternalKey(keys[i], base.SeqNum(i), base.InternalKeyKindSet),
-			blob.InlineHandle{
-				InlineHandlePreface: blob.InlineHandlePreface{
-					ReferenceID: base.BlobReferenceID(0),
-					ValueLen:    uint32(len(values[i])),
-				},
-				HandleSuffix: blob.HandleSuffix{
-					BlockID: handles[i].BlockID,
-					ValueID: handles[i].ValueID,
-				},
+		require.NoError(t, tw.AddWithBlobHandle(base.MakeInternalKey(keys[i], base.SeqNum(i), base.InternalKeyKindSet), blob.InlineHandle{
+			InlineHandlePreface: blob.InlineHandlePreface{
+				ReferenceID: base.BlobReferenceID(0),
+				ValueLen:    uint32(len(values[i])),
 			},
-			base.ShortAttribute(0),
-			false, /* forceObsolete */
-		))
+			HandleSuffix: blob.HandleSuffix{
+				BlockID: handles[i].BlockID,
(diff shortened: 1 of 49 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5179)* status: 0 of 44 files reviewed, 4 unresolved discussions (waiting on @sumeerbhola)

___
*[`internal/base/internal.go` line 637 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5179#-OXUcL455aKbW6ESxeMr:-OXV584Y8E8H1fatEaS5:b-dl3q91) ([raw file](https://github.com/cockroachdb/pebble/blob/dbf658537d0f0785cf37da88b497eeb6c21da8be/internal/base/internal.go#L637)):*

This is temporary plumbing, that I'll undo later, as indicated in this TODO.
For sstables, I am not keen to have:
- more fields thrown into the various handles when the value is not inline, i.e., blob.Handle or valblk.Handle. That design predates columnar blocks.
- when the value is inline, have to extract it by parsing some part of the value, since I don't want to make any compromises wrt performance.

By extracting once, and throwing them into columns at the end of the block, we should get good compression, and fast access during compactions.

> Conceptually, it would be a property of the _value_ not of the KV.

Isn't that an implementation detail, in that CockroachDB happens to place it in the MVCCValueHeader, for convenient extraction? If the SQL timestamp column was part of the key, it would be actually part of the key.

I think it is preferable to think of it as a property of the KV, since that is what will enable us to throw the keys into cold sstables in the future.
</blockquote></details>

Ack, makes sense
___
*[`sstable/colblk/data_block.go` line 532 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5179#-OXV1otgDe6Tp9mebVIH:-OXV5Xjy4WGUvViY7LtV:b-1y1edw) ([raw file](https://github.com/cockroachdb/pebble/blob/dbf658537d0f0785cf37da88b497eeb6c21da8be/sstable/colblk/data_block.go#L532)):*

Related to extraction, this is the model I had in mind.

I realize it means the `SpanPolicyFunc` configuration approach can't be fully reused to configure this, but I somewhat prefer not getting affected by trying to unify abstractions at this early stage. Storing the tiering attribute in both places seemed more natural to me.
</blockquote></details>

Yes, this is what I had in mind too. I was just thinking of the attt as being part of the value handle, but that's just a conceptual distinction. It is still stored with and associated with the key no matter what (and with the value in blob files).

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5179#-:-OXZpSiR8TLCizRHcwzL:bnfp4nl)* status: 0 of 49 files reviewed, 3 unresolved discussions (waiting on @sumeerbhola)

PR #5180, opened 2025-08-12
compact: convert MERGE keys to SET when bottommost

### LINKED ISSUE #5178: compact: reduce allocations for lone MERGE keys
Currently both iterators and compactions unconditionally send `MERGE` keys' values through a roundtrip of the user-configured `ValueMerger`, even when there is a single `MERGE` key (so the value should be unchanged). This results in unnecessary allocations, especially in the context of CockroachDB and its allocation-heavy `ValueMerger` implementation.

There are two potential resolutions:
1. We could explicitly defer beginning the merging of values until we observe two internal keys with the same user key.
2. We could adapt the compaction iterator to transform `MERGE` keys into `SET`s if we know a key is the oldest key at that user key (i.e., `allowZeroSeqNum=true` and the key is in the final snapshot stripe).

The code changes for approach (2) are a lot simpler, and changing the key to a SET has other advantages (eg, iteration can stop as soon as it hits the SET, avoiding an additional Next of the internal iterator).

Jira issue: PEBBLE-1139

jbowens:
This commit refactors the compaction iterator to perform an additional transformation to MERGE keys that are known to be the oldest version of their key in the LSM, transforming their key kind to SET. This ensures that iterators or compactions stepping over the key avoid invoking the merge operator when unnecessary.

Fix #5178.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -964,11 +964,16 @@ func (c *tableCompaction) errorOnUserKeyOverlap(ve *manifest.VersionEdit) error 
 	return nil
 }
 
-// allowZeroSeqNum returns true if seqnum's can be zeroed if there are no
-// snapshots requiring them to be kept. It performs this determination by
-// looking at the TombstoneElision values which are set up based on sstables
-// which overlap the bounds of the compaction at a lower level in the LSM.
-func (c *tableCompaction) allowZeroSeqNum() bool {
+// isBottommostDataLayer returns true if the compaction's inputs are known to be
+// the bottommost layer of data for the compaction's key range. If true, this
+// allows the compaction iterator to perform transformations to keys such as
+// setting a key's sequence number to zero.
+//
+// This function performs this determination by looking at the TombstoneElision
+// values which are set up based on sstables which overlap the bounds of the
+// compaction at a lower level in the LSM. This function always returns false
+// for flushes.
+func (c *tableCompaction) isBottommostDataLayer() bool {
 	// TODO(peter): we disable zeroing of seqnums during flushing to match
 	// RocksDB behavior and to avoid generating overlapping sstables during
 	// DB.replayWAL. When replaying WAL files at startup, we flush after each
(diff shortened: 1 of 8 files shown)


xinhaoz (approved):

jbowens:
TFTRs!

RaduBerinde:
*[`internal/compact/iterator.go` line 629 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5180#-OXZREr54_7OKRoFm4AN:-OXZREr54_7OKRoFm4AO:b-b7h4kf) ([raw file](https://github.com/cockroachdb/pebble/blob/58635e798945b13ba6cefabbaa23255b3cb7334b/internal/compact/iterator.go#L629)):*
> ```Go
> 				// SET so that iteration and future compactions do not need to
> 				// invoke the user's Merge operator.
> 				if i.isBottommostSnapshotStripe(origSnapshotIdx) {
> ```

Just a curiosity - if we add a `|| rand.IntN(100)` here, does the metamorphic test fail? If not, we should file an issue to improve merge-related coverage.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5180)* status: 2 of 8 files reviewed, 1 unresolved discussion

___
*[`internal/compact/iterator.go` line 629 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5180#-OXZREr54_7OKRoFm4AN:-OXZtuPqBYWjha2H6SB_:bnxhjsq) ([raw file](https://github.com/cockroachdb/pebble/blob/58635e798945b13ba6cefabbaa23255b3cb7334b/internal/compact/iterator.go#L629)):*

Just a curiosity - if we add a `|| rand.IntN(100)` here, does the metamorphic test fail? If not, we should file an issue to improve merge-related coverage.
</blockquote></details>

Yeah, it failed on the 3rd run

PR #5183, opened 2025-08-13
crl-release-25.2: compact: convert MERGE keys to SET when bottommost

### LINKED ISSUE #5178: compact: reduce allocations for lone MERGE keys

jbowens:
This commit refactors the compaction iterator to perform an additional transformation to MERGE keys that are known to be the oldest version of their key in the LSM, transforming their key kind to SET. This ensures that iterators or compactions stepping over the key avoids invoking the merge operator when unnecessary.

Fix #5178.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -273,12 +273,6 @@ type compaction struct {
 	delElision      compact.TombstoneElision
 	rangeKeyElision compact.TombstoneElision
 
-	// allowedZeroSeqNum is true if seqnums can be zeroed if there are no
-	// snapshots requiring them to be kept. This determination is made by
-	// looking for an sstable which overlaps the bounds of the compaction at a
-	// lower level in the LSM during runCompaction.
-	allowedZeroSeqNum bool
-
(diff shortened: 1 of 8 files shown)


RaduBerinde (approved):

jbowens:
TFTR!

PR #5185, opened 2025-08-13
[tiered prototype] db: add blob.FileFormatV3

sumeerbhola:
This format stores the base.TieringMeta with each value. Also added
plumbing to write and read the TieringMeta and propagate it during
compactions (including blob rewrite compactions).


sumeerbhola:
TFTR!

PR #5186, opened 2025-08-13
db: include size of input, output files for blob file rewrites

jbowens:
When logging a blob file rewrite compaction, include the size of the input and output blob files to make it clear how much disk space the rewrite compaction was able to reclaim.


jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5186#-:-OX_V2FX5l_DLISl4iuQ:b-814g8k)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @xxmplus)

PR #5187, opened 2025-08-13
db: don't smooth compaction picker compensated fill factor

jbowens:
When deciding whether a level is even eligible for compaction picking, we consider the 'compensated fill factor': a statistic about the size of a level plus its garbage relative to the level's ideal size given the level multiplier and database size. Currently, we consider a level eligible to be the input level of a compaction if the compensated fill factor is > 1.0 AND the compensated fill factor divided by the next level's fill factor is also > 1.0.

We've preserved this behavior for historical consistency but have no clear justification for it. Experimentally, we've observed instances where the LSM is known to contain significant garbage (>14%) but no compactions are pursued because most of the data is in L6 and L6's fill factor is positive (see cockroachdb/cockroach#151633).

This commit removes the separate concept of a 'compensated score,' instead allowing the compaction picker to pursue a compaction out of a level so long as its compensated fill factor is > 1.0.

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -1038,31 +1038,18 @@ func (p *compactionPickerByScore) calculateLevelScores(
 			continue
 		}
 		const compensatedFillFactorThreshold = 1.0
+		// The level requires compaction iff the compensatedFillFactor is >= 1.0.
 		if scores[level].compensatedFillFactor < compensatedFillFactorThreshold {
 			// No need to compact this level; score stays 0.
 			continue
 		}
 		score := scores[level].fillFactor
-		compensatedScore := scores[level].compensatedFillFactor
 		if level < numLevels-1 {
 			nextLevel := scores[level].outputLevel
 			// Avoid absurdly large scores by placing a floor on the factor that we'll
 			// adjust a level by. The value of 0.01 was chosen somewhat arbitrarily.
 			denominator := max(0.01, scores[nextLevel].fillFactor)
 			score /= denominator
-			compensatedScore /= denominator
-		}
-		// The level requires compaction iff both compensatedFillFactor and
-		// compensatedScore are >= 1.0.
-		//
-		// TODO(radu): this seems ad-hoc. In principle, the state of other levels
-		// should not come into play when we're determining this level's eligibility
-		// for compaction. The score should take care of correctly prioritizing the
-		// levels.
-		const compensatedScoreThreshold = 1.0
-		if compensatedScore < compensatedScoreThreshold {
-			// No need to compact this level; score stays 0.
-			continue
 		}
 		scores[level].score = score
 	}
--- a/testdata/compaction_picker_concurrency
+++ b/testdata/compaction_picker_concurrency
@@ -49,7 +49,9 @@ compactions
 pick-auto l0_compaction_threshold=10
 ----
 picker.getCompactionConcurrency: 4
-nil
+L0 -> L0
+L0: 000301,000302,000303,000304,000305
+grandparents: 000201
 
 # Test that lowering L0CompactionConcurrency opens up more compaction slots.
 


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5187)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`testdata/compaction_picker_concurrency` line 52 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5187#-OX_hJ4a3CgzR4W1hdJN:-OX_hJ4a3CgzR4W1hdJO:bg2znek) ([raw file](https://github.com/cockroachdb/pebble/blob/7ef6f668f4dfb11b6acf799cf1eda86ff3b58eb6/testdata/compaction_picker_concurrency#L52)):*
Here's a link to a discussion about this when the same change was propsed before:

https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP6pbHl185oMbbNtKYP:-OP6pbHl185oMbbNtKYQ:b-o2vfp7

> If Lbase has a high score and we are compacting out of Lbase and therefore unable to compact into Lbase, the new heuristic results in an intra-L0 compactions. Intra-L0 compactions are subjectively quite useless.

sumeerbhola (commented):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5187)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`testdata/compaction_picker_concurrency` line 52 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5187#-OX_hJ4a3CgzR4W1hdJN:-OXi2rGi4WMWzbb3ySy2:bwobyic) ([raw file](https://github.com/cockroachdb/pebble/blob/7ef6f668f4dfb11b6acf799cf1eda86ff3b58eb6/testdata/compaction_picker_concurrency#L52)):*

Here's a link to a discussion about this when the same change was propsed before:

https://reviewable.io/reviews/cockroachdb/pebble/4632#-OP6pbHl185oMbbNtKYP:-OP6pbHl185oMbbNtKYQ:b-o2vfp7

> If Lbase has a high score and we are compacting out of Lbase and therefore unable to compact into Lbase, the new heuristic results in an intra-L0 compactions. Intra-L0 compactions are subjectively quite useless.
</blockquote></details>

I'm ok with keeping `compensatedScore` only for L0 to determine eligibility for compaction.

petermattis (commented):
> but I wonder if any workloads will regress..

Should the changes be made behind a config setting so they can be reverted if a problem arises? Sort of messy, but good to have controlled rollout of changes like this.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5187#-:-OXiISOB8NBdK31-hqrT:b-335r46)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @RaduBerinde)

RaduBerinde (commented):
:lgtm:  Nice.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5187#-:-OXjDiU49NnWP5PSt7lT:b-bsmg0)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

sumeerbhola (approved):
> Should the changes be made behind a config setting so they can be reverted if a problem arises? Sort of messy, but good to have controlled rollout of changes like this.

+1 since these compaction heuristic changes are so hard to fully predict

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5187#-:-OXjrb4122ycXjJqunsN:b-c8pzgt)* status: all files reviewed (commit messages unreviewed), 4 unresolved discussions (waiting on @RaduBerinde)

___
*[`-- commits` line 17 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5187#-OXjrraP8IsTXPpGfsko:-OXjrraP8IsTXPpGfskp:b-xalqu3):*
nit: this paragraph is slightly stale
___
*[`compaction_picker.go` line 1041 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5187#-OXjsNOi9kLRFcLC4kzN:-OXjsNOi9kLRFcLC4kzO:b-eomkql) ([raw file](https://github.com/cockroachdb/pebble/blob/6eec91f9e9b034eeb68cbcfafad535e4003a34a7/compaction_picker.go#L1041)):*
> ```Go
> 		}
> 		const compensatedFillFactorThreshold = 1.0
> 		// The level requires compaction iff the compensatedFillFactor is >= 1.0.
> ```

nit: may require? "iff" isn't quite right, since L0 has an additional condition below.
___
*[`compaction_picker.go` line 1058 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5187#-OXjsd4EAyzeRiEVpRJ8:-OXjsd4EAyzeRiEVpRJ9:b-vb1g0) ([raw file](https://github.com/cockroachdb/pebble/blob/6eec91f9e9b034eeb68cbcfafad535e4003a34a7/compaction_picker.go#L1058)):*
> ```Go
> 		// after dividing by the next level's fill factor) is less than 1.0.
> 		// This helps avoid over eager L0 compaction picking (and intra L0
> 		// compactions) when LBase is large.
> ```

nit: consider elaborating on the "intra L0 compaction when Lbase is large". The concern specifically is that there are compactions happening out of Lbase to Lbase+1 because Lbase is large, and we are not able to find a key range for an L0 => Lbase compaction.

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5187#-:-OXzd_z2AF4zY3TnMFZr:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

PR #5189, opened 2025-08-14
[tiered prototype] options: change SpanPolicyFunc to take UserKeyBounds

RaduBerinde:
If the span policy function is non-trivial, it might have to do a lot
of work to find the end key at which e.g. the empty policy doesn't
apply anymore.

We switch to passing `UserKeyBounds`, since we only care about the
policy up to the compaction end bound.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -3389,7 +3389,10 @@ func (d *DB) compactAndWrite(
 			(len(spanPolicy.KeyRange.End) > 0 && d.cmp(firstKey, spanPolicy.KeyRange.End) >= 0) {
 			var err error
 			if d.opts.Experimental.SpanPolicyFunc != nil {
-				spanPolicy, err = d.opts.Experimental.SpanPolicyFunc(firstKey)
+				spanPolicy, err = d.opts.Experimental.SpanPolicyFunc(base.UserKeyBounds{
+					Start: firstKey,
+					End:   c.bounds.End,
+				})
 				if err != nil {
 					return runner.Finish().WithError(err)
 				}
--- a/internal.go
+++ b/internal.go
@@ -42,6 +42,9 @@ type InternalKey = base.InternalKey
 // KeyRange exports the base.KeyRange type.
 type KeyRange = base.KeyRange
 
+// UserKeyBounds exports the base.UserKeyBounds type.
+type UserKeyBounds = base.UserKeyBounds
+
 // MakeInternalKey constructs an internal key from a specified user key,
 // sequence number and kind.
 func MakeInternalKey(userKey []byte, seqNum SeqNum, kind InternalKeyKind) InternalKey {
--- a/options.go
+++ b/options.go
@@ -1210,11 +1210,15 @@ type ValueSeparationPolicy struct {
 // SpanPolicy contains policies that can vary by key range. The zero value for
 // all fields, other than the KeyRange, is the default policy.
 type SpanPolicy struct {
-	// KeyRange defines the key range for which this policy is valid. The end
-	// key can be empty, in which case the policy is valid for the entire
-	// keyspace after KeyRange.Start. The Start and End keys are not required to
-	// encompass the whole KeyRange over which this policy applies, i.e., they
-	// should be interpreted as a subset of the real interval for the policy.
+	// KeyRange defines the key range for which this policy is valid.
+	//
+	// If Start is empty, the policy is valid for the entire keyspace up to End.
+	// If End is empty, the policy is valid for the entire keyspace after Start.
+	// If both are empty, this is the only policy across the entire keyspace.
+	//
+	// The Start and End keys are not required to encompass the whole KeyRange over
+	// which this policy applies, i.e., they should be interpreted as a subset of
+	// the real interval for the policy.
 	KeyRange KeyRange
 
 	// Prefer a faster compression algorithm for the keys in this span.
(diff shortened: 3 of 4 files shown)


sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5189)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5189#-OXdW-W_B01aSLuACYp5:-OXdW-W_B01aSLuACYp6:b-7i6he0):*
I am curious why this was needed, given the `storage.spanPolicyFunc` trivially determines the end key of the policy.
The signature has become more complicated, since now the end key return has to be interpreted in the context of the paramter.

And in https://github.com/cockroachdb/pebble/pull/5190 I was taking a different approach. These policies conceptually have bounds, so trying to hide them for performance reasons seems like a premature optimization unless we have a good existing example where that optimization is useful.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5189)* status: all files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5189#-OXdW-W_B01aSLuACYp5:-OXdb3BE156qjvz6TcCP:b-14kcqb):*

There is definitely a real problem I'm trying to solve here. When the span policy will depend on the span config, we will need to use this API:
https://github.com/cockroachdb/cockroach/blob/c68c559859be738efead9971f5e11f62a8c69d06/pkg/spanconfig/spanconfig.go#L286

This returns one span config at a time. In principle, there could be thousands. We may need to go through a lot of them on every call until we find one that requires a "split point" (and in the common case of no tiering being used, we would always need to go through all of them). We would need to build a caching layer in-between that holds just the tiering information which seems unnecessary.

At a high level, the span policy is something we are interested in a localized key space, and the current API is asking for information that can cover essentially the entire keyspace.
</blockquote></details>

An alternative would be to change SpanPolicyFunc to be "best effort" in terms of how long the policy span is. Meaning it could arbitrarily break up a span into smaller pieces. We would need to add a flag that indicates whether we want a split point at the end or not. But I don't quite see the benefit.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5189)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5189#-OXdW-W_B01aSLuACYp5:-OXdpN994oZ78FMVVgib:bj2rkyv):*

An alternative would be to change SpanPolicyFunc to be "best effort" in terms of how long the policy span is. Meaning it could arbitrarily break up a span into smaller pieces. We would need to add a flag that indicates whether we want a split point at the end or not. But I don't quite see the benefit.
</blockquote></details>

I see. Do you have more details on how this information will be provided by the user as part of the ZONE config? My understanding was that CONFIGURE ZONE could not specify expressions referencing sql columns, unlike ttl_expiration_expression (used for row level TTL), so we may need different plumbing than the ZoneConfig to SpanConfig transformation.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5189)* status: all files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5189#-OXdW-W_B01aSLuACYp5:-OXdrqd53gBugpKV-uYH:b-vvrbdq):*

I see. Do you have more details on how this information will be provided by the user as part of the ZONE config? My understanding was that CONFIGURE ZONE could not specify expressions referencing sql columns, unlike ttl_expiration_expression (used for row level TTL), so we may need different plumbing than the ZoneConfig to SpanConfig transformation.
</blockquote></details>

The span config is not determined solely from the zone config. Anything from the table descriptor can affect the span config. The interface for doing this is `spanconfig.SQLTranslator` and this is the implementation: 
https://github.com/cockroachdb/cockroach/blob/c68c559859be738efead9971f5e11f62a8c69d06/pkg/spanconfig/spanconfigsqltranslator/sqltranslator.go#L31

The user would use DDL statements to set up the tiering. There would be something in the table descriptor related to tiering and that would make it into the span config.

For the prototype (at least for now), I am special casing a specific table name so we don't have to worry about the descriptor changes yet.

sumeerbhola:
@RaduBerinde in light of our discussion yesterday:
- I believe you'd mentioned that this PR maybe better to merge to the tiered branch.
- I think we'd agreed to moving the KeyRange into the SpanPolicy, which I've done in https://github.com/cockroachdb/pebble/pull/5190, and has been revised in light of our discussion.
- We'd agreed that passing in UserKeyBounds does make sense. So this PR LGTM, modulo the question of whether it merges to master or tiered (whatever you prefer).

RaduBerinde:
👍 I'll rebase onto `tiered` after 5190 merges.

RaduBerinde:
Updated.

sumeerbhola (approved):
:lgtm: 

> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5189#-:-OXoSj_T-BlXwLDwM_BX:bg7igkv)* status: 4 of 10 files reviewed, 1 unresolved discussion (waiting on @jbowens)

PR #5190, opened 2025-08-14
[tiered prototype] db: add tiering policy

sumeerbhola:
The policy is specified as a TieringPolicy struct and in combination with
an extractor func, to extract the tiering attribute, for use in
compactions and flushes. It is exposed via a TieringPolicyAndExtractor
interface embedded in the SpanPolicy.

SpanPolicy now contains a KeyRange, which was conceptually already in
existence, but was only partly exposed (via the end key). Exposing both
bounds allows Pebble to check that the invariant related to an immutable
spanID are not violated.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -3374,11 +3374,10 @@ func (d *DB) compactAndWrite(
 	}
 	runner := compact.NewRunner(runnerCfg, iter)
 
+	// If spanPolicyValid is true and spanPolicy.KeyRange.End is empty, then
+	// spanPolicy applies for the rest of the keyspace.
 	var spanPolicyValid bool
 	var spanPolicy SpanPolicy
-	// If spanPolicyValid is true and spanPolicyEndKey is empty, then spanPolicy
-	// applies for the rest of the keyspace.
-	var spanPolicyEndKey []byte
 
 	for runner.MoreDataToWrite() {
 		if c.cancel.Load() {
@@ -3386,12 +3385,18 @@ func (d *DB) compactAndWrite(
 		}
 		// Create a new table.
 		firstKey := runner.FirstKey()
-		if !spanPolicyValid || (len(spanPolicyEndKey) > 0 && d.cmp(firstKey, spanPolicyEndKey) >= 0) {
+		if !spanPolicyValid ||
+			(len(spanPolicy.KeyRange.End) > 0 && d.cmp(firstKey, spanPolicy.KeyRange.End) >= 0) {
 			var err error
-			spanPolicy, spanPolicyEndKey, err = d.opts.Experimental.SpanPolicyFunc(firstKey)
-			if err != nil {
-				return runner.Finish().WithError(err)
+			if d.opts.Experimental.SpanPolicyFunc != nil {
+				spanPolicy, err = d.opts.Experimental.SpanPolicyFunc(firstKey)
+				if err != nil {
+					return runner.Finish().WithError(err)
+				}
 			}
+			// Else SpanPolicyFunc is nil, so use the empty policy which already has
+			// an empty KeyRange.End.
+
 			spanPolicyValid = true
 		}
 
@@ -3417,7 +3422,7 @@ func (d *DB) compactAndWrite(
 		if err != nil {
 			return runner.Finish().WithError(err)
 		}
-		runner.WriteTable(objMeta, tw, spanPolicyEndKey, vSep)
+		runner.WriteTable(objMeta, tw, spanPolicy.KeyRange.End, vSep)
 	}
 	result = runner.Finish()
 	if result.Err == nil {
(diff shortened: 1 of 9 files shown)


RaduBerinde:
*[`options.go` line 1249 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5190#-OXdM44JBsfKdvfgSahh:-OXdM44JBsfKdvfgSahi:bfu0zp5) ([raw file](https://github.com/cockroachdb/pebble/blob/de6b8a41910ed5ed4b7d61f9f12c408b65816985/options.go#L1249)):*
> ```Go
> 	// key can be empty, in which case the policy is valid for the entire
> 	// keyspace after KeyRange.Start.
> 	KeyRange KeyRange
> ```

The end key of a policy can be expensive to get - e.g. we may need to dredge through all span configs (there can be up to 5000 of them I believe) to confirm that no remaining tables have tiering enabled. Hence https://github.com/cockroachdb/pebble/pull/5189 which limits the key range that the `SpanPolicyFunc` needs to look at.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5190#-:-OXi595w8a-N_bn4AKJU:bnfp4nl)* status: 0 of 9 files reviewed, 2 unresolved discussions

___
*[`options.go` line 1241 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5190#-OXi4ZNI3TFMGn5mWAZK:-OXi4ZNJARdrm342xUQ8:b-89ti63) ([raw file](https://github.com/cockroachdb/pebble/blob/fa8db23bc06763715079cb3bb2d1f673fcb2a133/options.go#L1241)):*
> ```Go
> 	// TieringPolicy is an optional policy for specifying which key-value pairs
> 	// should be stored in the warm or cold tier. Once a SpanPolicy specifies a
> 	// non-nil TieringPolicy, all subsequent requests for a SpanPolicy for a key
> ```

"all subsequent" seems like it's forever.. a table could get erased then its ID reused. Not sure what we want to say here.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5190)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`options.go` line 1241 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5190#-OXi4ZNI3TFMGn5mWAZK:-OXjiNfe0y1mLOaAb1Ey:b3zg5v7) ([raw file](https://github.com/cockroachdb/pebble/blob/fa8db23bc06763715079cb3bb2d1f673fcb2a133/options.go#L1241)):*

"all subsequent" seems like it's forever.. a table could get erased then its ID reused. Not sure what we want to say here.
</blockquote></details>

I added a comment
```
	// Additionally, a span may stop being used permanently by a higher layer,
	// and the data deleted, in which case it can reuse the same SpanID for a
	// different span. This must only be done when the deletion has destroyed
	// the data in the LSM, say via an excise.
```

It's risky if there is still data in the LSM, since the reused spanID could have a different policy which could dangerously cause a cold => hot transition for the existing data, even if there is a wide RANGEDEL sitting on top of it.
___
*[`options.go` line 1445 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5190#-OXiDoDLDepd2yE-Yr0J:-OXjjNTy4uFBhfgH_XEa:bez33ch) ([raw file](https://github.com/cockroachdb/pebble/blob/fa8db23bc06763715079cb3bb2d1f673fcb2a133/options.go#L1445)):*

[nit] Any reason this is an interface? `ExtractAttribute` will have a single universal implementation, we could have just a `TieringPolicy` and a `func`
</blockquote></details>

That's a good point. I'm conflicted in that I don't like over-generalizing, but this is more of a contract with the outside world, so being more general may be better.
I have been thinking of one use case recently, where this may be useful -- some users may be ok with using the mvcc timestamp to designate cold data, since it also means they don't need to do a schema change. That policy will have a different extraction function.
I'll merge this for now, and we can revisit.

RaduBerinde:
*[`options.go` line 1445 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5190#-OXiDoDLDepd2yE-Yr0J:-OXogJxaEw6_Tdk2F13U:bdm0tq1) ([raw file](https://github.com/cockroachdb/pebble/blob/fa8db23bc06763715079cb3bb2d1f673fcb2a133/options.go#L1445)):*

That's a good point. I'm conflicted in that I don't like over-generalizing, but this is more of a contract with the outside world, so being more general may be better.
I have been thinking of one use case recently, where this may be useful -- some users may be ok with using the mvcc timestamp to designate cold data, since it also means they don't need to do a schema change. That policy will have a different extraction function.
I'll merge this for now, and we can revisit.
</blockquote></details>

Nice idea.

PR #5192, opened 2025-08-15
[tiered prototype] compact: extract (spanId, attribute) if not already extracted

sumeerbhola:
There is some additional commentary on what this tuple represents. And all
the tiering declarations are moved to base so that they can be utilized in
lower level packages.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -3422,7 +3422,7 @@ func (d *DB) compactAndWrite(
 		if err != nil {
 			return runner.Finish().WithError(err)
 		}
-		runner.WriteTable(objMeta, tw, spanPolicy.KeyRange.End, vSep)
+		runner.WriteTable(objMeta, tw, spanPolicy.KeyRange.End, vSep, spanPolicy.TieringPolicy)
 	}
 	result = runner.Finish()
 	if result.Err == nil {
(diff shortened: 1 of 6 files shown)


sumeerbhola (commented):
The first commit is from https://github.com/cockroachdb/pebble/pull/5190

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5192#-:-OXhpGowCcPFyoTATf5T:baxobz1)* status: 0 of 12 files reviewed, all discussions resolved (waiting on @RaduBerinde)

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5192#-:-OXjnPl-6d0Et0e-YyEe:b-814g8k)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`options.go` line 1382 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5192#-OXi5_MaA_cwk6ygqzCL:-OXjlKlXEdg59WJeCqOQ:bsl0mfw) ([raw file](https://github.com/cockroachdb/pebble/blob/fa8db23bc06763715079cb3bb2d1f673fcb2a133/options.go#L1382)):*

We would need to exclude the zero value for the attribute. So it should be `0 < attribute < threshold`. An interval would make this more explicit, so we don't forget about the "exclude zero" part in some of the code.
</blockquote></details>

Good point -- I've updated the comment. I didn't change it to an interval, since we are not giving them a choice to choose an arbitrary interval (which could include 0). We're explicitly saying 0 is to not be returned as a valid attribute.

PR #5195, opened 2025-08-16
internal/rangekey: opt lenVarint

cuiweixie:


RaduBerinde:
We should just replace any uses of the function with `crencoding.UvarintLen32()` (from `crlib`).

cuiweixie:
> We should just replace any uses of the function with `crencoding.UvarintLen32()` (from `crlib`).

done

RaduBerinde:
LGTM, thanks!

cuiweixie:
I am not very sure. But it seems t the go-linux-no-cgo test failed is not caused by this pr. 
@RaduBerinde  What do we need to do?

RaduBerinde:
Looks like an unrelated test timeout, filed #5199.

Merging this. Thanks!

PR #5198, opened 2025-08-18
db: don't re-read sstable properties

RaduBerinde:
We add properties to `TableBacking`, used instead of opening the
sstable and reading the property block (which happens in multiple code
paths). The backing stats are populated directly when a table is
created (through compaction) or ingested. On startup, they are
asynchronously populated for existing tables (as part of calculating
table stats).

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -1761,17 +1761,21 @@ func (p *compactionPickerByScore) pickTombstoneDensityCompaction(
 			if f.IsCompacting() || f.Size == 0 {
 				continue
 			}
-			stats, statsValid := f.Stats()
-			if !statsValid || stats.TombstoneDenseBlocksRatio < p.opts.Experimental.TombstoneDenseCompactionThreshold {
+			props, propsValid := f.TableBacking.Properties()
+			if !propsValid {
+				continue
+			}
+			tombstoneDenseBlocksRatio := props.TombstoneDenseBlocksRatio()
+			if tombstoneDenseBlocksRatio < p.opts.Experimental.TombstoneDenseCompactionThreshold {
 				continue
 			}
 			overlaps := p.vers.Overlaps(lastNonEmptyLevel, f.UserKeyBounds())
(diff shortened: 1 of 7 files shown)


RaduBerinde:
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5198#-OXxny_zCyDT73PsC6-A:-OXxqQoI35WxVMRZzDm4:b-xxoaon):*

We recently purposely removed the properties from the `sstable.Reader`, moving to loading the properties from disk in the rare instances they're needed, to reduce memory. I think we can need to be careful about any per-sstable memory usage, and the properties struct is a substantial struct used rarely. Historically we've pulled only what we really need at runtime into the metadata's table stats.
</blockquote></details>

The intention is to keep only what we need at runtime. Currently this is `sstable.CommonProperties` (though it could be trimmed a little, at least the CompressionName for which I left a TODO).

On the flip side, this allows us to remove some of the fields of `TableStats`. Before the change, TableMetadata + TableBacking was 344+24=368. After the change, they are 288+176=464.

We can improve on this further by removing more of the duplication (some of the fields are only used for annotations, and we could just as well look up the backing stats and scale as necessary). I can work on these improvements to reduce the footprint increase in the WDYT?

RaduBerinde:
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5198#-OXxny_zCyDT73PsC6-A:-OXxwMEf-4-PJW2uZ5vN:b-cew17t):*

The intention is to keep only what we need at runtime. Currently this is `sstable.CommonProperties` (though it could be trimmed a little, at least the CompressionName for which I left a TODO).

On the flip side, this allows us to remove some of the fields of `TableStats`. Before the change, TableMetadata + TableBacking was 344+24=368. After the change, they are 288+176=464.

We can improve on this further by removing more of the duplication (some of the fields are only used for annotations, and we could just as well look up the backing stats and scale as necessary). I can work on these improvements to reduce the footprint increase in the WDYT?
</blockquote></details>

*in the same PR

RaduBerinde:
TFTR!

PR #5203, opened 2025-08-18
[tiered prototype] tieredmeta: add some code for a tiering histogram block

sumeerbhola:
The block will map TieringSpanID to a histogram. The support is
incomplete, with some todos for future PRs.

### DIFF
--- a/sstable/tieredmeta/block.go
+++ b/sstable/tieredmeta/block.go
@@ -1,0 +1,210 @@
+// Copyright 2025 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package tieredmeta
+
+import (
+	"cmp"
+	"fmt"
+	"slices"
+	"strconv"
+
+	"github.com/cockroachdb/errors"
+	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/sstable/colblk"
+)
+
+// A TieringHistogramBlockWriter writes a tiering histogram block. The block
+// records StatsHistograms for rows being written to a sstable or blob file.
+// If the key and value are being written to different files, there will be a
+// different TieringHistogramBlockWriter for each file.
+//
+// The tier of the (sstable or blob) file to which the tiering histogram block
+// is being written is implicit in the context, and does not need to be stored
+// in the histogram.
+//
+// We do want to distinguish between key bytes and value bytes, since even
+// though the keys will always be in the hot tier for now, it helps us
+// understand how much gains we could have by storing sstables in the cold
+// tier too.
+//
+// For a sstable, we want to distinguish which values it refers to are in hot
+// storage and which in cold storage. This is useful since before starting a
+// future compaction, we want to decide whether to write new blob files, and
+// in which tier, and which input blob files will have some spanIDs rewritten.
+// This decision needs to be made with the help of these histograms to balance
+// the cost of rewriting with the benefit, and to avoid producing tiny blob
+// files. Additionally, we want to know which values are inside the sstable
+// and which are blob references, since if most of the hot => cold transitions
+// are for values inside the sstable, we may choose to not rewrite the blob
+// references since the gain is small.
+//
+// To handle the above cases, an sstable can have up to 4 histograms for a
+// spanID, the ones prefixed with SSTable in the KindAndTier enum. A blob file
+// has one histogram for the values it contains, BlobFileValueBytes.
+//
+// For observability of the live bytes and their TieringAttribute, one can
+// sum:
+//
+// SSTableKeyBytes + SSTableValueBytes + SSTableBlobReferenceHotBytes +
+// SSTableBlobReferenceColdBytes
+//
+// Additionally,
+//
+// SSTableBlobReferenceHotBytes + SSTableBlobReferenceColdBytes <=
+// BlobFileValueBytes
+//
+// since the blob file may contain values that are not referenced by any
+// sstable.
+//
+// Why do we need the BlobFileValueBytes histogram? We could use it to copy a
+// blob file from hot to cold tier, or vice versa. Additionally, it could be
+// used to compare the sum of the stats for a spanID for the input sstables of
+// a compaction, with the sum of the stats of the spanID for blob files
+// referenced by the sstables in the compaction, to compute whether enough of
+// a spanID will move between tiers to justify rewriting the references.
+//
+// TODO(sumeer): the above is quite handwavy. It will become clearer when we
+// start implementing the decision functions. We err on the side of more
+// information for now, since in addition to decision making, it helps with
+// observability.
+
+type KindAndTier uint8
+
+const (
+	// SSTableKeyBytes is the histogram for sstable keys. The tier is implicit
+	// based on whether the sstable is in hot or cold storage.
+	SSTableKeyBytes KindAndTier = iota
+	// SSTableValueBytes is the histogram for values stored inside the
+	// sstable. The tier is implicit based on whether the sstable is in hot or
+	// cold storage.
+	SSTableValueBytes
+	// SSTableBlobReferenceHotBytes is the histogram for blob references from the
+	// sstable to hot tier blob files. The size is the size of the value.
+	SSTableBlobReferenceHotBytes
+	// SSTableBlobReferenceColdBytes is the histogram for blob references from the
+	// sstable to cold tier blob files. The size is the size of the value.
+	SSTableBlobReferenceColdBytes
+	// BlobFileValueBytes is the histogram for values stored inside the blob
+	// file. The tier is implicit based on whether the blob file is in hot or
+	// cold storage.
+	BlobFileValueBytes
+	NumKindAndTiers
+)
+
+type ColdTierThresholdRetriever interface {
+	// GetColdTierLTThreshold returns the current cold threshold for the given
+	// spanID. See the comment in histogram.go for how this is currently used,
+	// which is subject to revision.
+	//
+	// REQUIRES: spanID > 0.
+	GetColdTierLTThreshold(spanID base.TieringSpanID) base.TieringAttribute
+}
+
+// TieringHistogramBlockWriter is instantiated for each sstable or blob file
+// being written.
+type TieringHistogramBlockWriter struct {
+	retriever ColdTierThresholdRetriever
+	writers   map[Key]*histogramWriter
+}
+
+type Key struct {
+	KindAndTier
+	base.TieringSpanID
+}
+
+func (k *Key) encode() []byte {
+	return append([]byte{byte(k.KindAndTier)}, []byte(fmt.Sprintf("%08d", k.TieringSpanID))...)
+}
+
+func (k *Key) decode(data []byte) error {
+	if len(data) == 0 {
+		return errors.AssertionFailedf("key too short")
+	}
+	kt := KindAndTier(data[0])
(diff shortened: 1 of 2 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5203)* status: 0 of 8 files reviewed, 8 unresolved discussions

___
*[`sstable/tieredmeta/block.go` line 18 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY2nQdv3zxjfJKFoqw4:-OY2nQdv3zxjfJKFoqw5:b-8o4hrw) ([raw file](https://github.com/cockroachdb/pebble/blob/518557e262e1c62caf05a7ae9320cb3c11f9e020/sstable/tieredmeta/block.go#L18)):*
> ```Go
> )
> 
> // A TieringHistogramBlockWriter records the bytes for a row, being written to
> ```

We should mention early on here that all histograms we are talking about are bucketed on the tiering attribute axis.
___
*[`sstable/tieredmeta/block.go` line 19 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY2mazi6nNp0OvJQSjm:-OY2mazi6nNp0OvJQSjn:bkzfgs5) ([raw file](https://github.com/cockroachdb/pebble/blob/518557e262e1c62caf05a7ae9320cb3c11f9e020/sstable/tieredmeta/block.go#L19)):*
> ```Go
> 
> // A TieringHistogramBlockWriter records the bytes for a row, being written to
> // a sstable or blob file. If the key and value are both being written to
> ```

[nit] remove "both"
___
*[`sstable/tieredmeta/block.go` line 23 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY2mhsU-A4dznBBHJkn:-OY2mhsU-A4dznBBHJko:b-kss8r6) ([raw file](https://github.com/cockroachdb/pebble/blob/518557e262e1c62caf05a7ae9320cb3c11f9e020/sstable/tieredmeta/block.go#L23)):*
> ```Go
> // each file.
> //
> // The tier of the file for which the block is being written is implicit in
> ```

For blob file histograms, ...
___
*[`sstable/tieredmeta/block.go` line 26 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY2mmrg5ZU5QuhorKgB:-OY2mmrg5ZU5QuhorKgC:bfxu4a6) ([raw file](https://github.com/cockroachdb/pebble/blob/518557e262e1c62caf05a7ae9320cb3c11f9e020/sstable/tieredmeta/block.go#L26)):*
> ```Go
> // the context, and does not need to be stored in the histogram.
> //
> // We do want to distinguish between key bytes and value bytes, since even
> ```

We'd need to separate hot vs cold keys (perhaps together with how many bytes their values take up, whether they are small in-place values or value handles).

I agree this would be very useful information to aggregate for an LSM, but we don't need a histogram - some cold/hot stable properties would do just as well.
___
*[`sstable/tieredmeta/block.go` line 76 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY2lOFwEntdAwtKFoL3:-OY2lOFwEntdAwtKFoL4:b-1oy91o) ([raw file](https://github.com/cockroachdb/pebble/blob/518557e262e1c62caf05a7ae9320cb3c11f9e020/sstable/tieredmeta/block.go#L76)):*
> ```Go
> const (
> 	// SSTableKeyBytes is the histogram for sstable keys. The tier is implicit
> 	// based on whether the sstable is hot or cold.
> ```

[nit] The sstable itself won't be hot or cold, but a mix.
___
*[`sstable/tieredmeta/block.go` line 83 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY2ngBi8_Piekfb_pOo:-OY2ngBi8_Piekfb_pOp:b-pxqn0c) ([raw file](https://github.com/cockroachdb/pebble/blob/518557e262e1c62caf05a7ae9320cb3c11f9e020/sstable/tieredmeta/block.go#L83)):*
> ```Go
> 	// SSTableBlobReferenceHotBytes is the histogram for blob references from the
> 	// sstable to hot tier blob files. The size is the size of the value.
> 	SSTableBlobReferenceHotBytes
> ```

SSTableBlobReferenceHotBytes and SSTableBlobReferenceColdBytes - wouldn't they in practice look just like a single histogram for all references that was cut into two pieces along the threshold that was used when generating the sst? We could just store the threshold instead of having two separate histograms.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5203)* status: 0 of 8 files reviewed, 7 unresolved discussions (waiting on @RaduBerinde)

___
*[`sstable/tieredmeta/block.go` line 18 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY2nQdv3zxjfJKFoqw4:-OY42U2OCPVcFG-GywHB:b-z3pvea) ([raw file](https://github.com/cockroachdb/pebble/blob/518557e262e1c62caf05a7ae9320cb3c11f9e020/sstable/tieredmeta/block.go#L18)):*

We should mention early on here that all histograms we are talking about are bucketed on the tiering attribute axis.
</blockquote></details>

Added this at the start of the StatsHistogram comment, and referenced StatsHistogram here.
___
*[`sstable/tieredmeta/block.go` line 19 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY2mazi6nNp0OvJQSjm:-OY42gRY9eTjGkRzZEH2:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/518557e262e1c62caf05a7ae9320cb3c11f9e020/sstable/tieredmeta/block.go#L19)):*

[nit] remove "both"
</blockquote></details>

Done
___
*[`sstable/tieredmeta/block.go` line 23 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY2mhsU-A4dznBBHJkn:-OY43NP97c0lPnDHvE8r:b-y1ydls) ([raw file](https://github.com/cockroachdb/pebble/blob/518557e262e1c62caf05a7ae9320cb3c11f9e020/sstable/tieredmeta/block.go#L23)):*

For blob file histograms, ...
</blockquote></details>

This was meant for both files. I've revised this to
```
// The tier of the (sstable or blob) file to which the tiering histogram block
// is being written is implicit in the context, and does not need to be stored
// in the histogram.
```

The comment in general could probably be improved further. I'd like to leave that to later, since this may change significantly.
___
*[`sstable/tieredmeta/block.go` line 26 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY2mmrg5ZU5QuhorKgB:-OY3V6Ry1EDYLorc3iH9:bs5llg8) ([raw file](https://github.com/cockroachdb/pebble/blob/518557e262e1c62caf05a7ae9320cb3c11f9e020/sstable/tieredmeta/block.go#L26)):*

We'd need to separate hot vs cold keys (perhaps together with how many bytes their values take up, whether they are small in-place values or value handles).

I agree this would be very useful information to aggregate for an LSM, but we don't need a histogram - some cold/hot stable properties would do just as well.
</blockquote></details>

I am not sure how properties for hot or cold keys work without a histogram since time advances and what is hot or cold changes.
___
*[`sstable/tieredmeta/block.go` line 76 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY2lOFwEntdAwtKFoL3:-OY44K5F9630D9EeUyaD:b-mwfw6a) ([raw file](https://github.com/cockroachdb/pebble/blob/518557e262e1c62caf05a7ae9320cb3c11f9e020/sstable/tieredmeta/block.go#L76)):*

[nit] The sstable itself won't be hot or cold, but a mix.
</blockquote></details>

Here we are referring to whether the sstable itself is in hot or cold storage. The histograms are just TieringAttribute histograms so don't say anything about cold and hot by themselves (they need the TieringPolicy.ColdTierLTThreshold to interpret).

There is probably a way to improve the terminology. 
___
*[`sstable/tieredmeta/block.go` line 83 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY2ngBi8_Piekfb_pOo:-OY3VUnS3DAEJ724SJZm:butq32j) ([raw file](https://github.com/cockroachdb/pebble/blob/518557e262e1c62caf05a7ae9320cb3c11f9e020/sstable/tieredmeta/block.go#L83)):*

SSTableBlobReferenceHotBytes and SSTableBlobReferenceColdBytes - wouldn't they in practice look just like a single histogram for all references that was cut into two pieces along the threshold that was used when generating the sst? We could just store the threshold instead of having two separate histograms.
</blockquote></details>

Not necessarily. We may not rewrite all blob references when rewriting the sstable, so there may be cold data (based on the threshold) sitting in hot blob files (and vice versa).

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5203#-:-OY4PeqX7-z-kLdEzFED:bnfp4nl)* status: 0 of 8 files reviewed, 8 unresolved discussions (waiting on @sumeerbhola)

___
*[`sstable/tieredmeta/block.go` line 26 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY2mmrg5ZU5QuhorKgB:-OY4IDhg9ZaMbAndF9u0:bourh0y) ([raw file](https://github.com/cockroachdb/pebble/blob/518557e262e1c62caf05a7ae9320cb3c11f9e020/sstable/tieredmeta/block.go#L26)):*

I am not sure how properties for hot or cold keys work without a histogram since time advances and what is hot or cold changes.
</blockquote></details>

It would be cold or hot at the time the sstable was written. That would answer the question of how much space in the hot tier would we save if when we last compacted the sstable we would have moved the keys along with the values.

The histogram would let us answer the question of how much space would we save if everything was tiered perfectly. But it's unclear how we would use the histogram to derive LSM-wide metrics without doing `O(#files)` every time.
___
*[`sstable/tieredmeta/block.go` line 83 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY2ngBi8_Piekfb_pOo:-OY4JGHm1KvgbbC8KAzi:b-qv43bv) ([raw file](https://github.com/cockroachdb/pebble/blob/518557e262e1c62caf05a7ae9320cb3c11f9e020/sstable/tieredmeta/block.go#L83)):*

Not necessarily. We may not rewrite all blob references when rewriting the sstable, so there may be cold data (based on the threshold) sitting in hot blob files (and vice versa).
</blockquote></details>

Are you thinking that we may choose to rewrite the sstable plus just a subset of blob files? That sounds useful.
___
*[`sstable/tieredmeta/histogram.go` line 54 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5203#-OY4LJIe2vrzQXtN4ZRr:-OY4LJIe2vrzQXtN4ZRs:b4bavss) ([raw file](https://github.com/cockroachdb/pebble/blob/b8c6253137fd4dd717e0a472977a306ab7ac0b24/sstable/tieredmeta/histogram.go#L54)):*
> ```Go
> // happens, they are permanently in the unaccounted bucket.
> //
> // Age plumbing and age changes: TODO(sumeer): we will need to interpolate the
> ```

I'm not crazy about having the histograms depend on the "age" parameter..  Ok for now of course, but ideally we'd want something general.

I guess the conundrum is that usually 90+% of the data will be cold and we don't really care about distinguishing between very cold values (so we don't want to waste the "resolution" on those). But there must be a generic thing we can do, where the lowest-value buckets are more coarse than the highest-value buckets (somewhat similar to exponential bucket histograms).

sumeerbhola:
TFTR!

PR #5204, opened 2025-08-19
manifest: store the tombstone-dense block ratio in TableBackingStats

RaduBerinde:
#### manifest: don't embed CommonProperties in TableBackingProperties

This will allow us to derive more concise backing properties (for
example the tombstone-dense ratio).

We no longer rely on scaling the entire `CommonProperties`; this was
really only used for the point deletion bytes estimate, which we now
scale after calculating.

#### manifest: store the tombstone-dense block ratio in TableBackingStats


RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5204#-:-OY2ZdEo2cUwKrT6CBuQ:b-814g8k)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

PR #5211, opened 2025-08-19
compact: include expected size in MissizedDeleteCallback

jbowens:
Fix one of the instances where we report a missized DELSIZED tombstone but did not include the expected tombstone size. If a DELSIZED tombstone with a non-zero value meets another tombstone during a compaction, it indicates the more recent DELSIZED was missized--it deleted a key that did not exist (and was already deleted). Previously when invoking a callback to report this missizing, the compaction iterator passed expectedSize=0. This was misleading because the DELSIZED explicitly carried a non-zero size which allowed to us to determine there was a missizing in the first place.

This commit adapts the compaction iterator to decode the DELSIZED's value in this case and propagate it into the missized delete callback.

### DIFF
--- a/internal/compact/iterator.go
+++ b/internal/compact/iterator.go
@@ -1201,12 +1201,23 @@ func (i *Iter) deleteSizedNext() *base.InternalKV {
 			//
 			// We treat both cases the same functionally, adopting the identity
 			// of the lower-sequence numbered tombstone. However in the second
-			// case, we also increment the stat counting missized tombstones.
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5211#-:-OY2ZGQVEePLXfj-uV_O:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jbowens)

PR #5216, opened 2025-08-20
db: always estimate deletions via the table stats collector

jbowens:
When a new sstable is added to the LSM we need to populate its table stats, including PointDeletionBytesEstimate and RangeDeletionBytesEstimate. These statistics are typically calculated by the table stats collector asynchronously, because their calculation requires walking the LSM table metadata beneath the file and sometimes I/O.

Previously there was a fast path for sstables for which unsized point tombstones (DELs, SINGLEDELs but not DELSIZEDs) made up less than 10% of all of the table's entries. The previous logic reasoned that if there were not many deletions within the file, we could use the average value size and compression ratio of the file itself without introducing too much error overall. However, an incorrect compression ratio can introduce significant error, including for DELSIZED keys. In the previous logic, we could take the fast path even if the entirety of the table was DELSIZED tombstones.

This commit removes the fast path for all sstables except for those with no deletions at all. The 'slow path' of allowing the asynchronous table stats collector to calculate these stats is not significantly more resource intensive, and does not need to perform any I/O for point tombstones. I suspect the fast path was a premature optimization.

Informs cockroachdb/cockroach#151633.

### DIFF
--- a/table_stats.go
+++ b/table_stats.go
@@ -480,6 +480,16 @@ func (d *DB) loadTableRangeDelStats(
 	return compactionHints, rangeDeletionsBytesEstimate, nil
 }
 
+// estimateSizesBeneath calculates two statistics describing the data in the LSM
+// below the provided table metadata:
+//
+//  1. The average logical size of values: This is a precompression sum of
+//     non-tombstone values. It's helpful for estimating how much data a DEL
+//     might delete.
+//  2. The compression ratio of the data beneath the table.
+//
+// estimateSizesBeneath walks the LSM table metadata for all tables beneath meta
+// (plus the table itself), computing the above statistics.
 func (d *DB) estimateSizesBeneath(
 	ctx context.Context,
 	v *manifest.Version,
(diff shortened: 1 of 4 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5216#-:-OY8IwQx9-kLwgFgnC0i:bnfp4nl)* status: 0 of 6 files reviewed, 3 unresolved discussions

___
*[`table_stats.go` line 483 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5216#-OY8HPu_AeZXO55DqT_W:-OY8HPu_AeZXO55DqT_X:b-f9xgv5) ([raw file](https://github.com/cockroachdb/pebble/blob/d262c77ea6ff4b3b5334098def5d1a5e978bdae4/table_stats.go#L483)):*
> ```Go
> }
> 
> func (d *DB) estimateSizesBeneath(
> ```

[nit] It would be useful to add a comment explaining exactly what we're calculating
___
*[`table_stats.go` line 562 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5216#-OY8IiDF-K5x6m6EcbkN:-OY8IiDF-K5x6m6EcbkO:b-i506vp) ([raw file](https://github.com/cockroachdb/pebble/blob/d262c77ea6ff4b3b5334098def5d1a5e978bdae4/table_stats.go#L562)):*
> ```Go
> 	// When calculating the average value size, we subtract the number of
> 	// deletions from the total number of entries.
> 	avgValueLogicalSize = float64(valSum) / float64(max(1, invariants.SafeSub(entryCount, deletionCount)))
> ```

Should we subtract `∑RawPointTombstoneValueSize` from `valSum`?
___
*[`table_stats.go` line 720 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5216#-OY8H6yqBtOi-hMuTRkS:-OY8H6yqBtOi-hMuTRkT:b-mdm1y3) ([raw file](https://github.com/cockroachdb/pebble/blob/ef02ea02ccea9643d707c9c7f9d25bcbcf98b79f/table_stats.go#L720)):*
> ```Go
> 	}
> 	var stats manifest.TableStats
> 	meta.PopulateStats(&stats)
> ```

[nit] can just pass `&manifest.TableStats{}`

PR #5217, opened 2025-08-20
db: reduce TestBlobRewriteRandomized log spam

jbowens:
TestBlobRewriteRandomized logs a lot of context on its random state and operations and it's overwhelming when running tests with -v. This commit updates the test to write this context into an in-memory buffer, logging the contents only if the test fails.

### DIFF
--- a/blob_rewrite_test.go
+++ b/blob_rewrite_test.go
@@ -344,6 +344,14 @@ func TestBlobRewriteRandomized(t *testing.T) {
 		referencingTables: originalTables,
 	}}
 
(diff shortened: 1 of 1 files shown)


jbowens:
tftr!

RaduBerinde (commented):
:lgtm: Thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5217#-:-OY8JVi21Yw3tLYyLm4p:b-g2a5fb)* status: 0 of 1 files reviewed, all discussions resolved

Issue #4424, opened 2025-03-26
db: consider separating mvcc garbage values into separate blob files


jbowens:
The current (in-progress) design of value separation (#112) writes all the values of a sstable to a single blob file during compactions that write blob files. If there exists significant MVCC garbage, the MVCC garbage can reduce locality of access for reads at recent MVCC timestamps. We could consider instead storing these MVCC garbage values in a separate blob file.

This has the added advantage that in CockroachDB, MVCC garbage is expected to be deleted once the GC ttl elapses. That means we expect all the values within a garbage blob file to eventually be deleted. We may be able to delete the blob file even before a relevant blob-rewriting compaction is scheduled if non-blob-rewriting compactions compact the tombstones written by MVCC GC first.

Jira issue: PEBBLE-363

petermattis:
There is similarity to https://github.com/cockroachdb/pebble/issues/847. That issue was thinking about obsolete records pinned by snapshots, but the same idea could be done for historic MVCC keys. We already record the MVCC bounds in the sstable properties to do table-level filtering on reads. Separating out the historic MVCC keys into separate sstables has the nice property that you don't even have to skip over the keys during reads.

sumeerbhola:
> Separating out the historic MVCC keys into separate sstables has the nice property that you don't even have to skip over the keys during reads.

I suspect @jbowens was narrowly thinking of writing to two blob files for the separated values in a sstable -- that sounds like a good idea.
Regarding skipping over keys, I think this is similar to https://github.com/cockroachdb/pebble/issues/1170 which is now closed, post implementation of value blocks in sstables. The reason we did not separate the keys is that we have no way of knowing when writing a sstable in a compaction about which keys represent committed and resolved values. If we separate what we think are older keys and values, and the latest one is actually not committed and resolved, the MVCC get/scan will be incorrect.

petermattis:
Ack that a compaction can't know if the latest MVCC version is committed or not, but isn't it true that if there are two MVCC versions for a key that the second (older version) must be committed?

jbowens:
I filed #5040 for a couple ideas around separating _keys_ too. The first idea seems tractable but requires still storing the largest key in every prefix (even if it's a MVCC tombstone).

The second idea is a lot more convoluted. It optimistically assumes everything is committed but leaves enough information to take a slower detour during a read if we're wrong.

PR #5219, opened 2025-08-20
sstable: allow SetWithDelete keys to be considered as MVCC garbage

annrpom:
This patch allows the `SetWithDelete` key kind to be eligible for MVCC garbage consideration. This effectively allows `SetWithDelete` keys to be separated into value blocks and blob files.

Fixes: #4424

### DIFF
--- a/internal/compact/run.go
+++ b/internal/compact/run.go
@@ -338,7 +338,6 @@ func (r *Runner) writeKeysToTable(
 		isLikelyMVCCGarbage := func() bool {
 			return sstable.IsLikelyMVCCGarbage(kv.K.UserKey, prevKeyKind, kv.K.Kind(), valueLen, prefixEqual)
 		}
-		prevKeyKind = kv.K.Kind()
 		// Add the value to the sstable, possibly separating its value into a
 		// blob file. The ValueSeparation implementation is responsible for
 		// writing the KV to the sstable.
@@ -348,6 +347,7 @@ func (r *Runner) writeKeysToTable(
 		if err := valueSeparation.Add(tw, kv, r.iter.ForceObsoleteDueToRangeDel(), isLikelyMVCCGarbage); err != nil {
 			return nil, err
 		}
+		prevKeyKind = kv.K.Kind()
 		if r.iter.SnapshotPinned() {
 			// The kv pair we just added to the sstable was only surfaced by
 			// the compaction iterator because an open snapshot prevented
--- a/sstable/colblk_writer.go
+++ b/sstable/colblk_writer.go
@@ -1277,8 +1277,8 @@ func (w *RawColumnWriter) copyProperties(props Properties) {
 //
 // We require:
 //
-//	. The previous key to be a SET.
-//	. The current key to be a SET.
+//	. The previous key to be a SET/SETWITHDEL.
+//	. The current key to be a SET/SETWITHDEL.
 //	. The value to be sufficiently large. (Currently we simply require a
 //	  non-zero length, so all non-empty values are eligible for storage
 //	  out-of-band in a value block.)
@@ -1294,8 +1294,11 @@ func IsLikelyMVCCGarbage(
 	prefixEqual func(k []byte) bool,
 ) bool {
 	const tinyValueThreshold = 0
-	return prevKeyKind == InternalKeyKindSet &&
-		keyKind == InternalKeyKindSet &&
+	isSetStarKind := func(k base.InternalKeyKind) bool {
+		return k == InternalKeyKindSet || k == InternalKeyKindSetWithDelete
+	}
+	return isSetStarKind(prevKeyKind) &&
+		isSetStarKind(keyKind) &&
 		valueLen > tinyValueThreshold &&
 		prefixEqual(k)
 }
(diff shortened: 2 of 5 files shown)


annrpom (commented):
i also made a change to ensure that we use the same data (when specified in comments) in `writer_value_blocks`

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5219#-:-OYCJiPb0xlV9BATTA2s:b-jya1sw)* status: 2 of 5 files reviewed, 6 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/compact/run.go` line 350 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5219#-OYC0tSB8p09VW3B7c8s:-OYC5KkJ53KmLXzOTyyV:bebtosr) ([raw file](https://github.com/cockroachdb/pebble/blob/b8f242fb035b025088fd6994fac7abb5b2f550fb/internal/compact/run.go#L350)):*

bug caught by a new test?
</blockquote></details>

yes T_T
___
*[`sstable/colblk_writer.go` line 1297 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5219#-OYC10jAEEWI2WVQzXjr:-OYC5a1Z9Tk5ajhfP6jO:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/b8f242fb035b025088fd6994fac7abb5b2f550fb/sstable/colblk_writer.go#L1297)):*

nit: isValidKind suggests other kinds are invalid. How about `isSetStarKind` or something that hints that we are talking about {SET,SETWITHDEL}.
</blockquote></details>

done
___
*[`testdata/compaction/mvcc_garbage_blob` line 11 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5219#-OYC23fO-1Llftjy4OiX:-OYC8a4d572cNiLpz4ca:b3i7o5k) ([raw file](https://github.com/cockroachdb/pebble/blob/b8f242fb035b025088fd6994fac7abb5b2f550fb/testdata/compaction/mvcc_garbage_blob#L11)):*

I am assuming this is outputting:
SET yay@3, SET yay@2, SET zoo@4, DEL zoo@2, SET zoo@1.

Is this what caught the bug above, in that SET zoo@1 bah got separated?

Hmm, the comment suggests I am wrong, since it mentions both bag and bah. Why is bag being written, given it is already DELeted in the memtable?
</blockquote></details>

Ah - this is not what caught the bug above
Let me add a comment, but 
```
del zoo@2
set zoo@2 bag
```
turns into a `SetWithDel` in https://github.com/annrpom/pebble/blob/b8f242fb035b025088fd6994fac7abb5b2f550fb/internal/compact/iterator.go#L853-L860

So "bag" becomes eligible for separation into a blob file; same with "bah" since the prevKeyKind is a `SetWithDel`
___
*[`testdata/compaction/mvcc_garbage_blob` line 27 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5219#-OYC2z_f2hW0B7sNyymI:-OYC8zt72k2PqqBmKSEb:bz4zxov) ([raw file](https://github.com/cockroachdb/pebble/blob/b8f242fb035b025088fd6994fac7abb5b2f550fb/testdata/compaction/mvcc_garbage_blob#L27)):*

Or perhaps this caught the bug.

I think we could use a test with blobs and SETWITHDEL.
</blockquote></details>

yes this caught the bug; re: `SETWITHDEL` see above comment
___
*[`sstable/testdata/writer_value_blocks` line 21 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5219#-OYC3Vh4El3AthdrI6bD:-OYCAPah1my5C2dgAIKz:b-99s68d) ([raw file](https://github.com/cockroachdb/pebble/blob/b8f242fb035b025088fd6994fac7abb5b2f550fb/sstable/testdata/writer_value_blocks#L21)):*

is this in-place? It doesn't say in-place or value-handle, unlike the SET case. Perhaps the test code needs improvement?
</blockquote></details>

Let me add a comment, but this code uses the rowblk writer; we did not modify the value block separation logic for rowblk blocks since we no longer need to support writing such blocks
___
*[`sstable/testdata/writer_value_blocks` line 22 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5219#-OYC3f9TCTLIco5WZXQ6:-OYCBrUM0IFqVw_Cv72L:b-t3yigh) ([raw file](https://github.com/cockroachdb/pebble/blob/b8f242fb035b025088fd6994fac7abb5b2f550fb/sstable/testdata/writer_value_blocks#L22)):*

hmm, I'm confused why b@1 is in in-place.
</blockquote></details>

see above; the `v` preface is confusing here as well so i removed that
___
*[`sstable/testdata/writer_value_blocks` line 49 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5219#-OYC4EoK2SK-MDcnMkhd:-OYCGoce9-9vi5MufLOQ:bilcuc9) ([raw file](https://github.com/cockroachdb/pebble/blob/b8f242fb035b025088fd6994fac7abb5b2f550fb/sstable/testdata/writer_value_blocks#L49)):*

This comment would be more convincing with a scan-raw like we have in the previous test case.
</blockquote></details>

agreed, but it looks like column iterators don't support raw scan https://github.com/annrpom/pebble/blob/b8f242fb035b025088fd6994fac7abb5b2f550fb/sstable/writer_test.go#L421-L422; maybe this might be convincing enough: https://github.com/annrpom/pebble/blob/52962b13d89653bd1977faef986c22862e776256/sstable/testdata/writer_value_blocks#L80-L82

annrpom:
TFTR! ('-')7

PR #5220, opened 2025-08-20
db: aggregate blob file compression stats

RaduBerinde:
#### db: collect blob file properties

Collect blob file properties as part of initial table stats
collection. Properties for newly written blob files are populated
directly.

#### db: aggregate blob file compression stats

We rename `manifest.Annotator` to `TableAnnotator` and add a
`BlobFileAnnotator` type. Internally, we use a generic `annotator`
type to share code.

We use a `BlobFileAnnotator` to report compression statistics for blob
files.

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -1557,62 +1557,58 @@ func (p *compactionPickerByScore) addScoresToPickedCompactionMetrics(
 	}
 }
 
-// elisionOnlyAnnotator is a manifest.Annotator that annotates B-Tree
+// elisionOnlyAnnotator is a manifest.TableAnnotator that annotates B-Tree
 // nodes with the *fileMetadata of a file meeting the obsolete keys criteria
 // for an elision-only compaction within the subtree. If multiple files meet
 // the criteria, it chooses whichever file has the lowest LargestSeqNum. The
 // lowest LargestSeqNum file will be the first eligible for an elision-only
 // compaction once snapshots less than or equal to its LargestSeqNum are closed.
-var elisionOnlyAnnotator = &manifest.Annotator[manifest.TableMetadata]{
-	Aggregator: manifest.PickFileAggregator{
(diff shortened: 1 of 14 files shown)


jbowens (approved):
:lgtm:

> @jbowens reviewed 11 of 11 files at r1, 13 of 13 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5220#-:-OYI4Z45DuACcUSUF77h:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @xxmplus)

___
*[`table_stats.go` line 318 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5220#-OYI2vh492OAolMGcCt4:-OYI2vh5EphZOUrxhZse:btlfgdw) ([raw file](https://github.com/cockroachdb/pebble/blob/71562160cce71b91336ca0ad09694eb07e951509/table_stats.go#L318)):*
> ```Go
> 		}
> 		if maxNum == 0 {
> 			return moreRemain
> ```

am I missing it or do we never set `moreRemain=true` in the case that we exhaust `maxNum`

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5220)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens and @xxmplus)

___
*[`table_stats.go` line 318 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5220#-OYI2vh492OAolMGcCt4:-OYIxOdM5N3B9t1LWDNs:b-c45w6w) ([raw file](https://github.com/cockroachdb/pebble/blob/71562160cce71b91336ca0ad09694eb07e951509/table_stats.go#L318)):*

am I missing it or do we never set `moreRemain=true` in the case that we exhaust `maxNum`
</blockquote></details>

Good catch! I'll work on adding a test that would have caught this.

PR #5221, opened 2025-08-21
manifest: remove btree memory abuse

### LINKED ISSUE #5200: manifest: avoid unallocated memory backing b-tree node
For leaf nodes, the children array is garbage unallocated memory. 

https://github.com/cockroachdb/pebble/blob/efc47891e28916f1a2b9c22f4d5360a934489a31/internal/manifest/btree.go#L112

Our B-Tree implementation was copied and then adapted from the cockroach repository. The cockroach B-Tree implementation has since been updated to avoid this memory abuse.

https://github.com/cockroachdb/cockroach/commit/77fbc752049f0c7528fb754c6ef0cadf155b7ba1

Jira issue: PEBBLE-1145

RaduBerinde:
We currently allocate either a `leafNode` or a `node` and cast between
them. This is potentially unsafe, since at times we are holding a
`*node`, part of which is outside a `leafNode` object.

We remove this hack by moving the children list to a separate object
and having a pointer to it inside `node`. We also move the
`subtreeCount` to save space, since it is not useful in the leaves
(this offsets the size of the extra pointer for leaves, which are the
vast majority).

The separate object is still allocated together with the inner node,
so this does not increase the number of allocations.

Fixes #5200

### DIFF
--- a/internal/manifest/annotator.go
+++ b/internal/manifest/annotator.go
@@ -126,8 +126,8 @@ func (a *Annotator[T]) nodeAnnotation(n *node[*TableMetadata]) (t *T, cacheOK bo
 	valid := true
 
 	for i := int16(0); i <= n.count; i++ {
-		if !n.leaf {
-			v, ok := a.nodeAnnotation(n.children[i])
+		if !n.isLeaf() {
+			v, ok := a.nodeAnnotation(n.child(i))
 			t = a.Aggregator.Merge(v, t)
 			valid = valid && ok
 		}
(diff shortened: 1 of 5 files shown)


jbowens (approved):
:lgtm_strong: 

> @jbowens reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5221#-:-OYCA_XZ3RWKAMk8zj1f:b-uwkowd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #5223, opened 2025-08-21
manifest: replace marked for compaction annotator with a set

RaduBerinde:
*This PR is just for the top commit.*

The marked-for-compaction annotator is an overcomplicated way to
implement an ordered set. It does not fit neatly in the annotator
pattern - unlike all other annotators, it can become invalidated. It
also requires separate logic to maintain a count of marked files.

We remove this annotator and replace it with an ordered set (using the
google btree) stored in the Version. This is a lot cleaner and also
addresses a limitation of the annotator: we were only able to obtain
one candidate per level (and thus only one rewrite compaction
per level at any time).

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -1524,7 +1524,7 @@ func (p *compactionPickerByScore) pickAutoNonScore(env compactionEnv) (pc picked
 	// MarkedForCompaction field is persisted in the manifest. That's okay. We
 	// previously would've ignored the designation, whereas now we'll re-compact
 	// the file in place.
-	if p.vers.Stats.MarkedForCompaction > 0 {
+	if p.vers.MarkedForCompaction.Count() > 0 {
 		if pc := p.pickRewriteCompaction(env); pc != nil {
 			return pc
 		}
@@ -1591,19 +1591,6 @@ var elisionOnlyAnnotator = manifest.NewTableAnnotator[manifest.TableMetadata](ma
 		// TODO(travers): Consider an alternative heuristic for elision of range-keys.
 		eligible = stats.RangeDeletionsBytesEstimate*10 >= f.Size || backingProps.NumDeletions*10 > backingProps.NumEntries
 		return eligible, true
-	},
-	Compare: func(f1 *manifest.TableMetadata, f2 *manifest.TableMetadata) bool {
-		return f1.LargestSeqNum < f2.LargestSeqNum
-	},
-})
-
-// markedForCompactionAnnotator is a manifest.TableAnnotator that annotates B-Tree
-// nodes with the *fileMetadata of a file that is marked for compaction
-// within the subtree. If multiple files meet the criteria, it chooses
-// whichever file has the lowest LargestSeqNum.
-var markedForCompactionAnnotator = manifest.NewTableAnnotator[manifest.TableMetadata](manifest.PickFileAggregator{
-	Filter: func(f *manifest.TableMetadata) (eligible bool, cacheOK bool) {
-		return f.MarkedForCompaction, true
 	},
 	Compare: func(f1 *manifest.TableMetadata, f2 *manifest.TableMetadata) bool {
 		return f1.LargestSeqNum < f2.LargestSeqNum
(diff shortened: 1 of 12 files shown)


jbowens (approved):
top commit lgtm but left a suggestion to avoid the mutation of the latest version

> @jbowens reviewed 3 of 11 files at r1, 11 of 12 files at r3.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5223#-:-OYDELo2Dy1-IwuAz85b:bb2gnah)* status: 12 of 29 files reviewed, 1 unresolved discussion

___
*[`format_major_version.go` line 646 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5223#-OYDDJrN7NYvQ37u6z-U:-OYDDJrN7NYvQ37u6z-V:b-5afwrv) ([raw file](https://github.com/cockroachdb/pebble/blob/5a60d4a2456380391bfabb3d6a69f1bd3eb632fc/format_major_version.go#L646)):*
> ```Go
> 				// ok because we are holding the DB lock and all code that uses the
> 				// MarkedForCompaction set runs under the DB lock.
> 				// TODO(radu): find a less sketchy way to do this.
> ```

one suggestion: we could avoid the weird MANIFEST rotation down below and instead just add a version edit that encodes the file nums of files that should now be considered marked for compaction. I think we could even remove the `MarkedForCompaction` field from the table metadata and just have something new in the `latestVersionState` struct that maintains the set of files that have been marked for compaction

RaduBerinde:
*[`format_major_version.go` line 646 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5223#-OYDDJrN7NYvQ37u6z-U:-OYDpa_zC0MEyJ77c7HB:b-13n0gz) ([raw file](https://github.com/cockroachdb/pebble/blob/5a60d4a2456380391bfabb3d6a69f1bd3eb632fc/format_major_version.go#L646)):*

one suggestion: we could avoid the weird MANIFEST rotation down below and instead just add a version edit that encodes the file nums of files that should now be considered marked for compaction. I think we could even remove the `MarkedForCompaction` field from the table metadata and just have something new in the `latestVersionState` struct that maintains the set of files that have been marked for compaction
</blockquote></details>

We would also need some kind of new encoding for that, no?

If we want to reuse the existing stuff, we could remove and add back the table on the same level. There's some code that assumes this never happens but maybe it's not hard to fix, WDYT?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5223)* status: 10 of 29 files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`format_major_version.go` line 646 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5223#-OYDDJrN7NYvQ37u6z-U:-OYDwqXHEnoJgxNAX771:bszj9i7) ([raw file](https://github.com/cockroachdb/pebble/blob/5a60d4a2456380391bfabb3d6a69f1bd3eb632fc/format_major_version.go#L646)):*

We would also need some kind of new encoding for that, no?

If we want to reuse the existing stuff, we could remove and add back the table on the same level. There's some code that assumes this never happens but maybe it's not hard to fix, WDYT?
</blockquote></details>

Another possibility is to have an in-memory-only version edit that updates the map, and we apply that edit and then we do a second no-op update with manifest rotation to ensure all table metadata is fully re-encoded. Then we get rid of in-place modification (but keep the rotation hack).

I might just file an issue for now.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5223)* status: 10 of 29 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`format_major_version.go` line 646 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5223#-OYDDJrN7NYvQ37u6z-U:-OYHFM9I--vheHrVfaDD:bm3za89) ([raw file](https://github.com/cockroachdb/pebble/blob/5a60d4a2456380391bfabb3d6a69f1bd3eb632fc/format_major_version.go#L646)):*

Another possibility is to have an in-memory-only version edit that updates the map, and we apply that edit and then we do a second no-op update with manifest rotation to ensure all table metadata is fully re-encoded. Then we get rid of in-place modification (but keep the rotation hack).

I might just file an issue for now.
</blockquote></details>

I think it's a bit preferable to make this _not_ a property of the table metadata, since the manifest-encoded table metadata is intended to be immutable and we don't actually have any need for hanging it off the TableMetadata. Yeah it would require a new version edit encoding tag. Filing an issue for now sounds fine to me

PR #5231, opened 2025-08-22
[tiered prototype] db: read and aggregate tiering histograms

sumeerbhola:
As a temporary step, we read the histograms into TableStats and then use an AnnotationAggregator to aggregate them per level, so they can seen in metrics.

We will use the per sstable TableStats to make tiering decisions for compactions in future PRs.


sumeerbhola:
TFTR!

RaduBerinde:
Btw, even if we just keep a global histogram to which we add or subtract, we still have to either store each filehistogram in memory, or re-read it when a file is removed..

RaduBerinde:
There are some approximate datastructures (like count-sketch) that we could look into.

PR #5233, opened 2025-08-23
event: add a WALReplayed event

RaduBerinde:
We currently log some information about WALs which shows up more than
we'd like in verbose tests. This change adds a `WALReplayed` event to
`EventListener` and uses that instead.


RaduBerinde:
Hm, I think I should incorporate this log too:
```
[JOB %d] WAL %s stopped reading at offset: %s; replayed %d keys in %d batches
```

xinhaoz (approved):

PR #5234, opened 2025-08-23
manifest: clean up marked-for-compaction infrastructure

### LINKED ISSUE #5228: rework mark-for-compaction in the manifest
This issue tracks the changes suggested in #5223:
 - remove MarkedForCompaction from TableMetadata since the encoded part of TableMetadata is supposed to be immutable
 - store the tables that have been marked or unmarked for compaction since the last version update in `latestVersionState`
 - add version encoding tag for changing the marked-for-compaction state.

RaduBerinde:
In this change we remove `TableMetadata.MarkedForCompaction` (which is
now redundant with the marked-for-compaction set in the version); and
we add VersionEdit support and encoding for marking tables for
compaction (previously we would rely on forcing manifest rotation).

There are no backward-compatibility concerns because 1) none
of the supported Pebble versions marked anything for
compaction, and 2) the last version which marked files for
compaction was followed by a "barrier" version that ensured all
marked files were compacted.

Fixes #5228


RaduBerinde:
A new FMV seemed unnecessary, because we would only mark in some future version which will definitely support the new tags.  But we could in principle try to use an old pebble version to read the newer file (maybe through the pebble cmd), so maybe I should add one?

jbowens (approved):
:lgtm: 

> But we could in principle try to use an old pebble version to read the newer file (maybe through the pebble cmd), so maybe I should add one?

Yeah, I think would be okay without, but it seems prudent to add one.

> @jbowens reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5234#-:-OYc1_oP8FGHjgXyO2KN:b-gchdv6)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde (commented):
TFTR! Done.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5234#-:-OYgU1pQ58CMqTdZdl28:boopawo)* status: 4 of 18 files reviewed, all discussions resolved (waiting on @jbowens)

PR #5236, opened 2025-08-25
db: consolidate annotators

RaduBerinde:
This series of changes consolidates related annotators.

#### manifest: remove AnnotationAggregator.Zero() argument

We always pass `nil` to this method.

#### db: consolidate the file size annotators

We currently have three file size annotators which are used for
`EstimateDiskUsage`. Partially overlapping remote files are inspected
twice and external remote files three times. It also means 3x more
annotation-related objects per B-Tree node.

This change consolidates the three aggregators into a single one. We
generalize `SumAggregator` to allow arbitrary value types.

#### db: consolidate point and range bytes annotators

Using two annotators leads to two traversals and twice the allocated
objects per node. Consolidate into a single annotator.

#### manifest: remove SumAggregator

We also move the test-only NumFilesAnnotator to the test package. This
move triggered some false positives for `gcassert` which the rest of
the commit addresses (by including the `.` package).


jbowens (approved):
:lgtm_strong: 

> @jbowens reviewed 2 of 2 files at r1, 2 of 2 files at r2, 4 of 4 files at r3, 3 of 3 files at r4, 2 of 2 files at r5, 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5236#-:-OYWyvvM-kTXZ_3V52-k:b-uwkowd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #5239, opened 2025-08-25
db, manifest: rework annotation implementation, don't hold DB.mu when annotating

RaduBerinde:
#### manifest: rework annotation implementation

The current annotation implementation purports to be thread-safe but
in practice it is not: we store atomics in a slice which can get
copied over when it is resized.

We replace the dynamic slice with a static array. This requires
assigning each annotator a unique index.

The current implementation has a `valid` flag in addition to the
`atomic.Value`. Storing a non-valid value is never necessary, so
instead of using an extra flag we simply don't update `atomic.Value`
unless the value is valid.

We also rework the APIs to be as simple as possible and with less
boilerplate.

#### db: don't hold DB.mu when annotating

While we expect annotations to be very cheap for the most part, they
can be expensive after startup when many of the tables don't have the
properties or stats populated.

Now that the annotation infrastructure is thread-safe, we can
annotate without holding the DB mutex.

### DIFF
--- a/db.go
+++ b/db.go
@@ -505,8 +505,9 @@ type DB struct {
 			// validating is set to true when validation is running.
 			validating bool
 		}
-		fileSizeAnnotator manifest.TableAnnotator[fileSizeByBacking]
 	}
+
+	fileSizeAnnotator manifest.TableAnnotator[fileSizeByBacking]
 
 	// problemSpans keeps track of spans of keys within LSM levels where
 	// compactions have failed; used to avoid retrying these compactions too
@@ -1506,16 +1507,13 @@ func (d *DB) Close() error {
 	// Wait for all cleaning jobs to finish.
 	d.cleanupManager.Close()
 
-	// Sanity check metrics.
+	d.mu.Lock()
+	// Sanity check compaction metrics.
 	if invariants.Enabled {
-		m := d.Metrics()
-		if m.Compact.NumInProgress > 0 || m.Compact.InProgressBytes > 0 {
-			d.mu.Lock()
-			panic(fmt.Sprintf("invalid metrics on close:\n%s", m))
+		if d.mu.compact.compactingCount > 0 || d.mu.compact.downloadingCount > 0 || d.mu.versions.atomicInProgressBytes.Load() > 0 {
+			panic("compacting counts not 0 on close")
 		}
 	}
-
-	d.mu.Lock()
 
 	// As a sanity check, ensure that there are no zombie tables or blob files.
 	// A non-zero count hints at a reference count leak.
@@ -1794,6 +1792,9 @@ func (d *DB) Metrics() *Metrics {
 
 	d.mu.Lock()
 	vers := d.mu.versions.currentVersion()
+	vers.Ref()
+	defer vers.Unref()
+
 	*metrics = d.mu.versions.metrics
 	metrics.Compact.EstimatedDebt = d.mu.versions.picker.estimatedCompactionDebt()
 	metrics.Compact.InProgressBytes = d.mu.versions.atomicInProgressBytes.Load()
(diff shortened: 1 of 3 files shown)


RaduBerinde:
Oops, I forgot to update some documentation/comments. Converting to draft for now.

RaduBerinde:
Ok, updated.

jbowens (approved):
:lgtm:

> @jbowens reviewed 9 of 9 files at r1, 4 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5239#-:-OYXi2kD8eVkD8QVVTYl:bnfp4nl)* status: 8 of 10 files reviewed, 5 unresolved discussions

___
*[`db.go` line 1905 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5239#-OYXheufEQxDIyjcvhx3:-OYXheufEQxDIyjcvhx4:b-t9k40h) ([raw file](https://github.com/cockroachdb/pebble/blob/38bbe1c030e4d4a27108dcd44c8fc783d7305170/db.go#L1905)):*
> ```Go
> 	// files in the associated version from being deleted if there is a concurrent
> 	// compaction.
> 	readState := d.loadReadState()
> ```

why the readState instead of the previous `vers`? we don't need the memtables right? also I think we probably still want to use a version that we retrieved while holding the mutex to ensure consistency between metrics that are protected by `d.mu` and those that are computed from the version
___
*[`internal/manifest/annotator.go` line 167 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5239#-OYXgo_Y9qqhGeRL92YA:-OYXgo_Y9qqhGeRL92YB:b-gshyso) ([raw file](https://github.com/cockroachdb/pebble/blob/c1c5efb36b5874262fb16d506b0db71f5f6b62cf/internal/manifest/annotator.go#L167)):*
> ```Go
> //	AccumulateFunc               func(f *TableMetadata) (v T, cacheOK bool)
> //	AccumulatePartialOverlapFunc func(f *TableMetadata, bounds base.UserKeyBounds) T
> //}
> ```

did you intend to delete this in this revision?
___
*[`internal/manifest/annotator_blob.go` line 24 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5239#-OYXgvaw3KZWjs_yxG1Z:-OYXgvaxApLaAq8BKXd9:bpu7lin) ([raw file](https://github.com/cockroachdb/pebble/blob/c1c5efb36b5874262fb16d506b0db71f5f6b62cf/internal/manifest/annotator_blob.go#L24)):*
> ```Go
> }
> 
> // BlobFileAnnotatorFuncs groups the functions that define a table annotator.
> ```

nit: "blob file annotator"
___
*[`internal/manifest/annotator_blob.go` line 39 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5239#-OYXgzxn7BbQvZLB1cZV:-OYXgzxn7BbQvZLB1cZW:b-khz2y9) ([raw file](https://github.com/cockroachdb/pebble/blob/c1c5efb36b5874262fb16d506b0db71f5f6b62cf/internal/manifest/annotator_blob.go#L39)):*
> ```Go
> // all blob files in the given set.
> // A pointer to the TableAnnotator is used as the key for pre-calculated values,
> // so the same TableAnnotator must be used to avoid duplicate computation.
> ```

nit: s/TableAnnotator/BlobFileAnnotator/ in this comment
___
*[`internal/manifest/annotator_blob.go` line 55 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5239#-OYXh4eBBNyMY9QewO4z:-OYXh4eBBNyMY9QewO5-:b-i7w5pr) ([raw file](https://github.com/cockroachdb/pebble/blob/c1c5efb36b5874262fb16d506b0db71f5f6b62cf/internal/manifest/annotator_blob.go#L55)):*
> ```Go
> type BlobFileAnnotationIdx int
> 
> // NewBlobAnnotationIdx generates a unique AnnotationIdx that can be used with
> ```

nit: s/AnnotationIdx/BlobFileAnnotationIdx/

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5239#-:-OYXs6FzCeClRu5r6vCK:b-814g8k)* status: 4 of 10 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`db.go` line 1905 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5239#-OYXheufEQxDIyjcvhx3:-OYXqnKP8Q70sDz2NqRf:beao5s3) ([raw file](https://github.com/cockroachdb/pebble/blob/38bbe1c030e4d4a27108dcd44c8fc783d7305170/db.go#L1905)):*

why the readState instead of the previous `vers`? we don't need the memtables right? also I think we probably still want to use a version that we retrieved while holding the mutex to ensure consistency between metrics that are protected by `d.mu` and those that are computed from the version
</blockquote></details>

It was getting cleaned up from under us. I switched back to using `vers` but added `Ref()/Unref()`.
___
*[`internal/manifest/annotator.go` line 167 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5239#-OYXgo_Y9qqhGeRL92YA:-OYXrnBI3zBWub-yRAjs:b-1adqkr) ([raw file](https://github.com/cockroachdb/pebble/blob/c1c5efb36b5874262fb16d506b0db71f5f6b62cf/internal/manifest/annotator.go#L167)):*

did you intend to delete this in this revision?
</blockquote></details>

Yes, sorry, fixed.

PR #5243, opened 2025-08-26
db: clear a subset of Iterator fields on Close

jbowens:
Adjust the logic that zeroes an Iterator to only clear fields after a marker `clearForReuseBoundary` field, and use this to delegate clearing of the blob.ValueFetcher to the blob.ValueFetcher's Close method. When initializing the Iterator, only set the fields that need to be set (instead relying on the allocated Iterator's other fields having already been zeroed.  This in preparation for recycling a buffer within the blob.ValueFetcher across reuses of the containing Iterator.

```
goos: darwin
goarch: arm64
cpu: Apple M1 Pro
                    │   old.txt    │             fetcher.txt             │
                    │    sec/op    │   sec/op     vs base                │
NewIterClose/1-10     1014.0n ± 1%   849.2n ± 2%  -16.25% (p=0.000 n=25)
NewIterClose/10-10     5.246µ ± 0%   5.184µ ± 1%   -1.18% (p=0.000 n=25)
NewIterClose/100-10    78.68µ ± 1%   79.25µ ± 2%        ~ (p=0.151 n=25)
geomean                7.480µ        7.040µ        -5.89%
```


xinhaoz (approved):

jbowens:
TFTR!

PR #5248, opened 2025-08-27
blob: vary max cached readers by read amplification

### LINKED ISSUE #5181: blob: vary number of cached readers
The blob.ValueFetcher caches blob file readers and the most recently accessed blocks per reader. Currently it always caches a maximum of 5 readers, regardless of the read-amplification. We should vary the number of readers based on the number of levels in the LSM. We can experiment to see what impact this has, if any, on scan performance.

Jira issue: PEBBLE-1140

jbowens:
The blob.ValueFetcher maintains a small cache of open blob file readers. This allows a scan that's retrieving values to avoid repeatedly re-opening the same blob file and re-retrieving the index block from the block cache. This cache previously used a fixed size of 5 in all contexts. This commit updates the fetcher to configure the size of the cache dynamically. Iterator construction now sets a cache size of 5 times the number of non-empty levels in the LSM, capturing the fact that deeper LSMs will reference more blob files over a span.

Close #5181.

```
                   │   all.txt    │
                   │   ops/sec    │
ycsb/A/values=1024   383.6k ± ∞ ¹
ycsb/B/values=1024   619.7k ± ∞ ¹
ycsb/C/values=1024   1.005M ± ∞ ¹
ycsb/D/values=1024   232.8k ± ∞ ¹
ycsb/E/values=1024   96.19k ± ∞ ¹
ycsb/F/values=1024   82.91k ± ∞ ¹
geomean              276.2k
```

### DIFF
--- a/blob_rewrite_test.go
+++ b/blob_rewrite_test.go
@@ -410,7 +410,8 @@ func TestBlobRewriteRandomized(t *testing.T) {
 		// Verify that the rewritten blob file contains the correct values, and
 		// that they may still be accessed using the original handles.
 		var valueFetcher blob.ValueFetcher
-		valueFetcher.Init(constantFileMapping(newBlobFileNum), fch, readEnv)
+		valueFetcher.Init(constantFileMapping(newBlobFileNum), fch, readEnv,
+			blob.SuggestedCachedReaders(1))
 		func() {
 			defer func() { _ = valueFetcher.Close() }()
 			for _, valueIndex := range newFile.valueIndices {
--- a/compaction.go
+++ b/compaction.go
@@ -3321,7 +3321,8 @@ func (d *DB) compactAndWrite(
 		),
 	}
 	if c.version != nil {
-		c.iterationState.valueFetcher.Init(&c.version.BlobFiles, d.fileCache, blockReadEnv)
+		c.iterationState.valueFetcher.Init(&c.version.BlobFiles, d.fileCache, blockReadEnv,
+			blob.SuggestedCachedReaders(len(c.inputs)))
 	}
 	iiopts := internalIterOpts{
 		compaction:       true,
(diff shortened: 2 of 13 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5248#-:-OYgGqiA6wVdrmGkm9PT:bnfp4nl)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola)

___
*[`sstable/blob/fetcher.go` line 87 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5248#-OYgGiHY0zO24TCGiaD9:-OYgGiHY0zO24TCGiaDA:b76xqc6) ([raw file](https://github.com/cockroachdb/pebble/blob/602d3e69fe35827f965e1db53f74903b707a70a4/sstable/blob/fetcher.go#L87)):*
> ```Go
> 		panic("readerProvider is nil")
> 	}
> 	if cap(r.readers) >= maxCachedReaders {
> ```

[nit] This block can be `slices.Grow(r.readers[:0], maxCachedReaders)[:maxCachedReaders]` and it can reduce allocations

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5248#-:-OYgci7G9mudTNoG8m2I:b-814g8k)* status: 0 of 13 files reviewed, all discussions resolved (waiting on @sumeerbhola)

___
*[`sstable/blob/fetcher.go` line 87 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5248#-OYgGiHY0zO24TCGiaD9:-OYgcjv7Av4KPvXTZ5lY:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/602d3e69fe35827f965e1db53f74903b707a70a4/sstable/blob/fetcher.go#L87)):*

[nit] This block can be `slices.Grow(r.readers[:0], maxCachedReaders)[:maxCachedReaders]` and it can reduce allocations
</blockquote></details>

Done

PR #5253, opened 2025-08-28
cache: use malloc instead of calloc

RaduBerinde:
Many profiles have a percentage or two in `manual.Alloc`. Switch to
using `malloc` instead of `calloc`, which should speed up the
allocations. We clear only the part that we cast to `*cache.Value`. We
already clear the part that we use for block metadata (via
`block.CastMetadataZero`).

### DIFF
--- a/internal/cache/value.go
+++ b/internal/cache/value.go
@@ -53,7 +53,7 @@ func Alloc(n int) *Value {
 
 	if valueEntryGoAllocated {
 		// Note: if cgo is not enabled, manual.New will do a regular Go allocation.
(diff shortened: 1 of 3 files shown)


RaduBerinde:
I am a little concerned that most of our testing is with invariants on, and we don't use `manual` for the block cache at all in that case. Maybe I should change things so that half the time we use it even with invariants?

RaduBerinde:
> Maybe I should change things so that half the time we use it even with invariants?

Opened https://github.com/cockroachdb/pebble/pull/5254

RaduBerinde:
TFTR! 🤞

PR #5254, opened 2025-08-28
cache: invariants: use Go allocated values half the time

RaduBerinde:
Currently if the `invariants` tag is used (which is true for most of
our testing), the cache uses Go-backed memory. This reduces test
coverage on the production path of using "manual" memory via cgo.

This commit changes things so that we only use Go-backed memory half
the time in invariants mode. We do this while still allowing the
Go-backed code to be statically elided in production builds.

### DIFF
--- a/internal/cache/clockpro.go
+++ b/internal/cache/clockpro.go
@@ -122,7 +122,7 @@ func (c *shard) init(maxSize int64) {
 		maxSize:    maxSize,
 		coldTarget: maxSize,
 	}
-	if entriesGoAllocated {
+	if entriesCanBeGoAllocated && entriesGoAllocated {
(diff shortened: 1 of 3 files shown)


jbowens (approved):
good idea

:lgtm:

> @jbowens reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5254#-:-OYlhqcL2iNhCwUGy-Aw:by60rtt)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #5256, opened 2025-08-28
db: SeekPrefixGE lazy positioning

Sachuman:
WIP ( still writing down detailed comments in the code ) 

1. db: SeekPrefixGE lazy positioning

2. db: add probe tests to TestIterHistories

### DIFF
--- a/file_cache.go
+++ b/file_cache.go
@@ -352,6 +352,7 @@ func createReader(
 		env.Virtual = meta.VirtualParams
 		env.IsSharedIngested = v.isShared && meta.SyntheticSeqNum() != 0
 	}
+	env.InternalBounds = &meta.PointKeyBounds
 	return r, env
 }
 
@@ -588,6 +589,7 @@ func (h *fileCacheHandle) newIters(
 	r, env := createReader(v, file)
 	internalOpts.readEnv.Virtual = env.Virtual
 	internalOpts.readEnv.IsSharedIngested = env.IsSharedIngested
+	internalOpts.readEnv.InternalBounds = env.InternalBounds
 
 	var iters iterSet
 	if kinds.RangeKey() && file.HasRangeKeys {
@@ -730,6 +732,7 @@ func (h *fileCacheHandle) newPointIter(
 				ValueFetcher: internalOpts.blobValueFetcher,
 				References:   blobReferences,
 			},
+			MaximumSuffixProperty: sstable.MaxTestKeysSuffixProperty{},
 		})
 	}
 	if err != nil {
--- a/internal.go
+++ b/internal.go
@@ -19,6 +19,7 @@ const (
 	InternalKeyKindMerge          = base.InternalKeyKindMerge
 	InternalKeyKindLogData        = base.InternalKeyKindLogData
 	InternalKeyKindSingleDelete   = base.InternalKeyKindSingleDelete
+	InternalKeyKindSyntheticKey   = base.InternalKeyKindSyntheticKey
 	InternalKeyKindRangeDelete    = base.InternalKeyKindRangeDelete
 	InternalKeyKindMax            = base.InternalKeyKindMax
 	InternalKeyKindSetWithDelete  = base.InternalKeyKindSetWithDelete
--- a/internal/base/internal.go
+++ b/internal/base/internal.go
@@ -96,7 +96,8 @@ const (
 	//InternalKeyKindRollbackXID              InternalKeyKind = 12
 	//InternalKeyKindNoop                     InternalKeyKind = 13
 	//InternalKeyKindColumnFamilyRangeDelete  InternalKeyKind = 14
-	InternalKeyKindRangeDelete InternalKeyKind = 15
+	InternalKeyKindSyntheticKey InternalKeyKind = 14
+	InternalKeyKindRangeDelete  InternalKeyKind = 15
 	//InternalKeyKindColumnFamilyBlobIndex    InternalKeyKind = 16
 	//InternalKeyKindBlobIndex                InternalKeyKind = 17
 
@@ -200,6 +201,7 @@ var internalKeyKindNames = []string{
 	InternalKeyKindMerge:          "MERGE",
 	InternalKeyKindLogData:        "LOGDATA",
 	InternalKeyKindSingleDelete:   "SINGLEDEL",
+	InternalKeyKindSyntheticKey:   "SYNTHETIC",
 	InternalKeyKindRangeDelete:    "RANGEDEL",
 	InternalKeyKindSeparator:      "SEPARATOR",
 	InternalKeyKindSetWithDelete:  "SETWITHDEL",
@@ -307,6 +309,7 @@ func MakeExclusiveSentinelKey(kind InternalKeyKind, userKey []byte) InternalKey 
 var kindsMap = map[string]InternalKeyKind{
 	"DEL":           InternalKeyKindDelete,
 	"SINGLEDEL":     InternalKeyKindSingleDelete,
+	"SYNTHETIC":     InternalKeyKindSyntheticKey,
 	"RANGEDEL":      InternalKeyKindRangeDelete,
 	"LOGDATA":       InternalKeyKindLogData,
 	"SET":           InternalKeyKindSet,
@@ -541,7 +544,8 @@ func (k InternalKey) IsExclusiveSentinel() bool {
 	}
 	switch kind := k.Kind(); kind {
 	case InternalKeyKindRangeDelete, InternalKeyKindRangeKeyDelete,
-		InternalKeyKindRangeKeyUnset, InternalKeyKindRangeKeySet:
+		InternalKeyKindRangeKeyUnset, InternalKeyKindRangeKeySet,
+		InternalKeyKindSyntheticKey:
 		return true
 	default:
 		return false
--- a/internal/itertest/testdata/probes
+++ b/internal/itertest/testdata/probes
@@ -21,6 +21,30 @@ bar#2,SET:value
 bar#2,SET:value
 bar#2,SET:value
 
+# Create probes to exercise the UserKey predicate. On Next/Prev return a KV
+# with user key "bar"; on other ops return a KV with user key "foo". Then if
+# the returned KV's user key is "bar", override the result with a specific KV.
+
+new
+(If (Or OpNext OpPrev) (ReturnKV "bar.SET.2" "value") (ReturnKV "foo.SET.1" "value"))
+(If (UserKey "bar") (ReturnKV "ok.SET.1" "ok") noop)
+----
+
+iter
+first
+next
+seek-ge foo
+seek-lt bar
+last
+prev
+----
+foo#1,SET:value
+ok#1,SET:ok
+foo#1,SET:value
+foo#1,SET:value
+foo#1,SET:value
+ok#1,SET:ok
+
 # Create an identical probe as above, and then wrap it in a "noop" probe.
 # A noop probe does nothing; it passes through the child iterator's results
 # verbatim, so the iteration results should be the same.
--- a/iterator_histories_test.go
+++ b/iterator_histories_test.go
@@ -5,6 +5,7 @@ package pebble
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"path/filepath"
 	"strconv"
@@ -15,6 +16,8 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/invariants"
+	"github.com/cockroachdb/pebble/internal/itertest"
+	"github.com/cockroachdb/pebble/internal/manifest"
 	"github.com/cockroachdb/pebble/internal/testkeys"
 	"github.com/cockroachdb/pebble/internal/testutils"
 	"github.com/cockroachdb/pebble/sstable"
@@ -38,6 +41,7 @@ func TestIterHistories(t *testing.T) {
 		var buf bytes.Buffer
 		iters := map[string]*Iterator{}
 		batches := map[string]*Batch{}
+		parser := itertest.NewParser()
 		newIter := func(name string, reader Reader, o *IterOptions) *Iterator {
 			it, _ := reader.NewIter(o)
 			iters[name] = it
@@ -288,6 +292,8 @@ func TestIterHistories(t *testing.T) {
 				o := &IterOptions{KeyTypes: IterKeyTypePointsAndRanges}
 				var reader Reader = d
 				var name string
+				pointProbes := make(map[base.TableNum][]itertest.Probe)
+				rangeDelProbes := make(map[base.TableNum][]keyspanProbe)
 				for _, arg := range td.CmdArgs {
 					switch arg.Key {
 					case "mask-suffix":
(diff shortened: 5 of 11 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5256)* status: 4 of 13 files reviewed, 16 unresolved discussions (waiting on @12, @13, @Sachuman, and @xinhaoz)

___
*[`sstable/reader_iter_two_lvl.go` line 861 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZAirKuA2Iw35MohkjM:-OZFtED26UH-g9olu51-:b-x0djv9) ([raw file](https://github.com/cockroachdb/pebble/blob/269b84df2e891da86cc80f517efa6fa5703775a8/sstable/reader_iter_two_lvl.go#L861)):*
<details><summary><i>Previously, Sachuman (Sachin) wrote…</i></summary><blockquote>

During the two level SeekPrefixGE, we never loaded its index block. So during next we must load the index block before calling single level iterator.
If you were talking  about replacing secondLevel.Next with seekPrefixGE. We can certainly do that and get some results.
</blockquote></details>

Ah I think I was confusing the code structure of twoLevelIterator and singleLevelIterator.

Can we extract the body of the seek prefix GE and share the code between `twoLevelIterator.SeekPrefixGE` and `twoLevelIterator.Next`?
___
*[`sstable/reader_iter_single_lvl.go` line 1299 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFv-V9CFchWb87SuTN:-OZFv-V9CFchWb87SuTO:b-l27bgn) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/sstable/reader_iter_single_lvl.go#L1299)):*
> ```Go
> func (i *singleLevelIterator[I, PI, D, PD]) Next() *base.InternalKV {
> 
> 	if i.synthetic.atSyntheticKey {
> ```

nit: let's put a comment above this if statement explaining the optimization and that SeekPrefixGE may have returned a synthetic key with the "smallest" (smallest according to base.Compare) suffix contained in the sstable, avoiding performing the actual seek. If the caller is calling Next, they want to move past the synthetic key and Next is responsible for performing the actual seek.
___
*[`internal/itertest/testdata/probes` line 26 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFr-ne0yqOUCZbure5:-OZFr-ne0yqOUCZbure6:b-kfyu5e) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/internal/itertest/testdata/probes#L26)):*
> ```
> # Create probes to exercise the UserKey predicate. On Next/Prev return a KV
> # with user key "bar"; on other ops return a KV with user key "foo". Then if
> # the returned KV's user key is "bar", override the result with a specific KV.
> ```

I'm a bit confused as to what this test case is seeking to test?
___
*[`iterator_histories_test.go` line 358 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFqXjf0fWmV7F2ayDQ:-OZFqXjf0fWmV7F2ayDR:b3dc7jd) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/iterator_histories_test.go#L358)):*
> ```Go
> 						o.UseL6Filters = true
> 					case "probe-points":
> ```

nit: remove empty line
___
*[`iterator_histories_test.go` line 367 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFqV9k7rtjDIlAX6kC:-OZFqV9k7rtjDIlAX6kD:b3dc7jd) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/iterator_histories_test.go#L367)):*
> ```Go
> 						pointProbes[base.TableNum(i)] = itertest.MustParseProbes(parser, arg.Vals[1:]...)
> 					case "probe-rangedels":
> ```

nit: remove empty line
___
*[`sstable/reader_iter_two_lvl.go` line 429 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFut2n2H3qJfvpSygk:-OZFut2n2H3qJfvpSygl:b-zh9up2) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/sstable/reader_iter_two_lvl.go#L429)):*
> ```Go
> 	// We do so because there might be cases which result in generating a synthetic key which was not supposed to be generated
> 	// But due to the Bounds we might end up creating the synthetic key, which will result in a seek failure. A case that can
> 	// be considered is when we have a RangeDel key at the upper bound of the sstable, which possibly can result in fake synthetic key.
> ```

nit: ditto the comment on the singleLevelIterator
___
*[`sstable/reader_iter_two_lvl.go` line 436 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFu_uw5-Z75pQ1eCEI:-OZFu_uw5-Z75pQ1eCEJ:b-g0e6wu) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/sstable/reader_iter_two_lvl.go#L436)):*
> ```Go
> 		largest = i.secondLevel.reader.Comparer.Split.Prefix(largest)
> 
> 		if i.secondLevel.cmp(prefix, smallest) > 0 && i.secondLevel.cmp(prefix, largest) < 0 {
> ```

nit: let's move checking this condition to be the innermost. the vast majority of the time this condition will be met, whereas the other conditions (eg, the seek key's suffix being < the max suffix) are less likely to be met.
___
*[`testdata/iter_histories/probe_prefix` line 178 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFsC1E8qBe_O6LTAzF:-OZFsC1F9_4qMIac8935:b-cz7k5h) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/testdata/iter_histories/probe_prefix#L178)):*
> ```
> ----
> 
> combined-iter probe-points=(4,(Log "OpSeekPrefixGE")) probe-rangedels=(4,(Log "OpSeekPrefixGE"))
> ```

these `"OpSeekPrefixGE"` prefixes are a bit confusing. how about
```
combined-iter probe-points=(4,(Log "000004.")) probe-rangedels=(4,(Log "000004.RangeDels."))
```

(and analogous changes throughout the test file)

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5256)* status: 4 of 13 files reviewed, 18 unresolved discussions (waiting on @12, @13, @Sachuman, and @xinhaoz)

___
*[`sstable/reader_iter_two_lvl.go` line 448 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFvpn3-TSWehLeG5PT:-OZFvpn3-TSWehLeG5PU:bsqqb5x) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/sstable/reader_iter_two_lvl.go#L448)):*
> ```Go
> 					}
> 				} else {
> 					maxSuffix, ok, err = i.secondLevel.maximumSuffixProperty.Extract(maxSuffix, unsafe.Slice(unsafe.StringData(prop), len(prop)))
> ```

in order to use the `dst` parameter to reduce allocations, we need to retain a buffer between calls to `SeekPrefixGE`. We can add a `maxSuffixBuf []byte` to `i.synthetic`.

Then we would do something like:
```
maxSuffix, ok, err = i.secondLevel.maximumSuffixProperty.Extract(i.synthetic.maxSuffixBuf[:0], unsafe.Slice(unsafe.StringData(prop), len(prop)))
if err != nil {
	i.secondLevel.err = err
	return nil
} else if ok {
	i.synthetic.maxSuffixBuf = maxSuffix
}
```
___
*[`sstable/reader_iter_two_lvl.go` line 471 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFwYpR-KCXM6i-QBzd:-OZFwYpR-KCXM6i-QBze:bk4e0za) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/sstable/reader_iter_two_lvl.go#L471)):*
> ```Go
> 				// to still be open by the time singleLevelIterator.Next is called
> 				// and we use the seek key to actually perform the seek.
> 				i.secondLevel.synthetic.seekKey = append(i.secondLevel.synthetic.seekKey[:0], key...)
> ```

We should retain `seekKey` and the new `maxSuffix` field on `i.synthetic` across Closes. To do that, I think we should move `i.synthetic` below `i.clearForResetBoundary` in the struct definition (so `i.synthetic` no longer has its memory zeroed automatically in `resetForReuse`), and then explicitly update `resetForReuse` to do:
```
clear(synthetic.seekKey[:cap(synthetic.seekKey)])
clear(synthetic.maxSuffix[:cap(synthetic.maxSuffix)])
synthetic.seekKey = synthetic.seekKey[:0]
synthetic.maxSuffix = synthetic.maxSuffix[:0]
```
and explicitly clear out the other fields of `i.synthetic`

Sachuman (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5256)* status: 1 of 13 files reviewed, 11 unresolved discussions (waiting on @12, @13, @jbowens, and @xinhaoz)

___
*[`sstable/reader_iter_single_lvl.go` line 801 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFiNqb7wWtFJ45b_8i:-OZLL3gX-xeAKaNcbxPD:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/sstable/reader_iter_single_lvl.go#L801)):*

nit: wrap the comment lines at ~80-100 chars, and format the TODO prefix as `TODO(sachin): ...`
</blockquote></details>

Done.
___
*[`sstable/reader_iter_single_lvl.go` line 819 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFix2W-02UaDLm6YI9:-OZLL59M4tFpRBMILfas:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/sstable/reader_iter_single_lvl.go#L819)):*

nit: I think this should say something like "However, we can only perform ..."

(also wrap at 80-100 chars)
</blockquote></details>

Done.
___
*[`sstable/reader_iter_single_lvl.go` line 1299 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFv-V9CFchWb87SuTN:-OZLL6Vg5gHtRs5IoH0Y:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/sstable/reader_iter_single_lvl.go#L1299)):*

nit: let's put a comment above this if statement explaining the optimization and that SeekPrefixGE may have returned a synthetic key with the "smallest" (smallest according to base.Compare) suffix contained in the sstable, avoiding performing the actual seek. If the caller is calling Next, they want to move past the synthetic key and Next is responsible for performing the actual seek.
</blockquote></details>

Done.
___
*[`sstable/reader_iter_two_lvl.go` line 861 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZAirKuA2Iw35MohkjM:-OZLLbD89Xjms1aBjGiu:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/269b84df2e891da86cc80f517efa6fa5703775a8/sstable/reader_iter_two_lvl.go#L861)):*

Ah I think I was confusing the code structure of twoLevelIterator and singleLevelIterator.

Can we extract the body of the seek prefix GE and share the code between `twoLevelIterator.SeekPrefixGE` and `twoLevelIterator.Next`?
</blockquote></details>

Done.
___
*[`sstable/reader_iter_two_lvl.go` line 429 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFut2n2H3qJfvpSygk:-OZLL9582vvrvPZWYPUA:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/sstable/reader_iter_two_lvl.go#L429)):*

nit: ditto the comment on the singleLevelIterator
</blockquote></details>

Done.
___
*[`sstable/reader_iter_two_lvl.go` line 436 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFu_uw5-Z75pQ1eCEI:-OZLL9msAioAkuBoAiBb:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/sstable/reader_iter_two_lvl.go#L436)):*

nit: let's move checking this condition to be the innermost. the vast majority of the time this condition will be met, whereas the other conditions (eg, the seek key's suffix being < the max suffix) are less likely to be met.
</blockquote></details>

Done.
___
*[`sstable/reader_iter_two_lvl.go` line 448 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFvpn3-TSWehLeG5PT:-OZLLB846X8kCA7NNArm:bprhzu4) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/sstable/reader_iter_two_lvl.go#L448)):*

in order to use the `dst` parameter to reduce allocations, we need to retain a buffer between calls to `SeekPrefixGE`. We can add a `maxSuffixBuf []byte` to `i.synthetic`.

Then we would do something like:
```
maxSuffix, ok, err = i.secondLevel.maximumSuffixProperty.Extract(i.synthetic.maxSuffixBuf[:0], unsafe.Slice(unsafe.StringData(prop), len(prop)))
if err != nil {
	i.secondLevel.err = err
	return nil
} else if ok {
	i.synthetic.maxSuffixBuf = maxSuffix
}
```
</blockquote></details>

I completely missed the point of reducing allocations with buffer. Thanks for this, made changes as suggested.
___
*[`testdata/iter_histories/probe_prefix` line 178 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFsC1E8qBe_O6LTAzF:-OZLLcp6-YTNY3iVvi3G:bg4foot) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/testdata/iter_histories/probe_prefix#L178)):*

these `"OpSeekPrefixGE"` prefixes are a bit confusing. how about
```
combined-iter probe-points=(4,(Log "000004.")) probe-rangedels=(4,(Log "000004.RangeDels."))
```

(and analogous changes throughout the test file)
</blockquote></details>

Agreed,  Done.
___
*[`internal/itertest/testdata/probes` line 26 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFr-ne0yqOUCZbure5:-OZLLfQDDMjf-1fKPznu:biz82qj) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/internal/itertest/testdata/probes#L26)):*

I'm a bit confused as to what this test case is seeking to test?
</blockquote></details>

I added a test case for testing dsl function 'UserKey' since there wasn't one I believe. Made some change to the test case to make it more clear.
___
*[`iterator_histories_test.go` line 358 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFqXjf0fWmV7F2ayDQ:-OZLM1Xs947Yh0Z4QN3u:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/iterator_histories_test.go#L358)):*

nit: remove empty line
</blockquote></details>

Done.
___
*[`iterator_histories_test.go` line 367 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5256#-OZFqV9k7rtjDIlAX6kC:-OZLM2797hj5fCyQka5m:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/992d0556d440d93d9fb332843cd50f62ac0a11bf/iterator_histories_test.go#L367)):*

nit: remove empty line
</blockquote></details>

Done.

sjain022:
Added plumbing missing during level iter initialization.

TFTR!

PR #5250, opened 2025-08-28
[tiered prototype] db: implementation of ValueSeparation for tiering

sumeerbhola:
generalizedValueSeparation subsumes the functionality of
preserveBlobReferences and writeNewBlobFiles, and can additionally
preserve cold references or write new cold blob files.

There are some changes to the ValueSeparation interface to go with this
change.

PR #5257, opened 2025-08-29
db,compact: change sstable splitting behavior derived from SpanPolicy

sumeerbhola:
We had two deficiencies in how SpanPolicy was causing sstable splitting:
- When not writing new cold blob files, which is always the case for
  higher levels, and will often be the case for lower levels (since
  data will be moved from hot => cold in large chunks), we were splitting
  sstables when the TieringPolicy changed. This is a result of splits
  always occurring at SpanPolicy boundaries (in addition to other causes
  of splits).
  For this case, we now extract the attributes needed to switch sstables,
  into ExtractedSpanPolicy, and walk the SpanPolicys until the extracted
  attributes change, and then split.

- When writing cold blob files, we did not have a notion of a shard
  boundary (e.g. CockroachDB ranges), which can be deleted as a whole
  by the upper layer (e.g. CockroachDB). This shard boundary is needed
  to prevent cold blob files from spanning shards (best-effort, since
  such boundaries can change), so they can be deleted as a whole.
  For this case, the SpanPolicyFunc accepts a breakAtShardEndBoundary
  parameter that causes SpanPolicys to be broken at these boundaries.

These changes are accompanied with some reworking of the compact.Runner
and ValueSeparation interfaces.

### DIFF
--- a/blob_rewrite_test.go
+++ b/blob_rewrite_test.go
@@ -81,26 +81,27 @@ func TestBlobRewrite(t *testing.T) {
 				bv = blobtest.Values{}
 				switch x := d.CmdArgs[0].String(); x {
 				case "preserve-blob-references":
-					pbr := &preserveBlobReferences{}
 					lines := crstrings.Lines(d.Input)
-					pbr.inputBlobPhysicalFiles = make(map[base.BlobFileID]*manifest.PhysicalBlobFile, len(lines))
+					inputBlobPhysicalFiles := make(map[base.BlobFileID]*manifest.PhysicalBlobFile, len(lines))
 					for _, line := range lines {
 						bfm, err := manifest.ParseBlobFileMetadataDebug(line)
 						require.NoError(t, err)
 						fn = max(fn, bfm.Physical.FileNum)
-						pbr.inputBlobPhysicalFiles[bfm.FileID] = bfm.Physical
+						inputBlobPhysicalFiles[bfm.FileID] = bfm.Physical
 					}
+					pbr := newPreserveAllHotBlobReferences(inputBlobPhysicalFiles, 0)
 					vs = pbr
+					vs.StartOutput(compact.ValueSeparationOutputConfig{})
 				case "write-new-blob-files":
-					newSep := &writeNewBlobFiles{
-						comparer: testkeys.Comparer,
-						newBlobObject: func() (objstorage.Writable, objstorage.ObjectMetadata, error) {
+					var minimumSize int
+					d.MaybeScanArgs(t, "minimum-size", &minimumSize)
+					newSep := newWriteNewBlobFiles(nil, testkeys.Comparer,
+						func() (objstorage.Writable, objstorage.ObjectMetadata, error) {
 							fn++
 							return objStore.Create(ctx, base.FileTypeBlob, fn, objstorage.CreateOptions{})
-						},
-					}
-					d.MaybeScanArgs(t, "minimum-size", &newSep.minimumSize)
+						}, nil, blob.FileWriterOptions{}, minimumSize, nil, nil)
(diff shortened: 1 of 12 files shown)


sumeerbhola (commented):
First commit is from https://github.com/cockroachdb/pebble/pull/5250

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5257#-:-OYnNtH8EzgJOBP1eqmn:bz76l1v)* status: 0 of 12 files reviewed, all discussions resolved

RaduBerinde (commented):
I feel like things would be cleaner (and more efficient in terms of implementing the span policy func) if we had two functions `SpanPolicyForUpperLevels()` which retrieves just the "core" policy and `SpanPolicyForLowerLevels()` which also includes the tiering policy and breaks at range boundaries. That way we wouldn't have to introduce the concept of ranges/shards into Pebble, and we wouldn't need to "merge" core policies in Pebble. And we wouldn't waste time looking up span configs for upper levels.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5257#-:-OYnY4A31fcWl95U8Cs_:bm4s676)* status: 0 of 12 files reviewed, 2 unresolved discussions

___
*[`options.go` line 1222 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5257#-OYnYVFT4WBMV7uVcNdm:-OYnYVFT4WBMV7uVcNdn:b-2tn8vf) ([raw file](https://github.com/cockroachdb/pebble/blob/d813a49439c94f9904fce5f9f04801a6dedef61a/options.go#L1222)):*
> ```Go
> virtualize them as recovering space requires a rewrite.
> 
> Call it shardEndBoundary.
> ```

[nit] Why not call it RangeEndBoundary so there's less confusion inside CRDB?
___
*[`options.go` line 1316 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5257#-OYnW11i0saZTPMmfcBc:-OYnW11i0saZTPMmfcBd:b708suh) ([raw file](https://github.com/cockroachdb/pebble/blob/d813a49439c94f9904fce5f9f04801a6dedef61a/options.go#L1316)):*
> ```Go
> // ExtractedSpanPolicy is like SpanPolicy but focused on the policy fields,
> // minus the TieringPolicyAndExtractor.
> type ExtractedSpanPolicy struct {
> ```

[nit] should we embed this in `SpanPolicy` (and maybe rename to `CoreSpanPolicy` and define it as the part of the policy where any change requires a split in output files)

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5257)* status: 0 of 13 files reviewed, all discussions resolved (waiting on @RaduBerinde)

___
*[`options.go` line 1222 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5257#-OYnYVFT4WBMV7uVcNdm:-OYrMGekDVD9w8FmZg8A:b7ynerl) ([raw file](https://github.com/cockroachdb/pebble/blob/d813a49439c94f9904fce5f9f04801a6dedef61a/options.go#L1222)):*
> Perhaps "partition" instead of "shard"?

I have a mild preference for shard given it is more specifically used in such distributed settings, while partition seems to be less specific (e.g. partitioning data on a single node DB to drop partitions easily). https://en.wikipedia.org/wiki/Shard_(database_architecture) seems to reinforce this.
I can change it if you have a strong preference.

> Or go more general with `potentialExciseBoundary`? As in a boundary where it is likely we will receive an excise for at some point (which is what happens when a range moves between nodes).

I was imagining we would also use this in the future to split disaggregated storage sstables, at least in L6, so it isn't solely about excise or RANGEDELs.

PR #5258, opened 2025-08-29
db: add high-priority blob file rewrite compactions

jbowens:
This commit introduces a new trigger for blob-file rewrite compactions. Previously blob-file rewrite compactions were triggered when
  a) no default score-based compactions were available to run
  b) no tombstone density or elision-only compaction were available to run
  c) total unreferenced blob values exceeded the 'garbage ratio' threshold
  d) a blob file meets the criteria for rewrite (including the minimum age)

We've observed that score-based compactions can starve blob-file rewrite compactions, leaving blob-file—induced space amplification unbounded. This commit renames the old 'garbage ratio' threshold to GarbageRatioLowPriority and introduces a new GarbageRatioHighPriority setting. When the ratio of unreferenced values reaches the high priority setting, Pebble will schedule up to 1 blob file rewrite compaction at a time as long as eligible blob files exist. To avoid reducing the concurrency available for default compactions, compaction concurrency is temporarily inflated for the duration of the blob file rewrite compaction.

Informs #112.

### DIFF
--- a/blob_rewrite.go
+++ b/blob_rewrite.go
@@ -7,6 +7,7 @@ package pebble
 import (
 	"container/heap"
 	"context"
+	"fmt"
 	"iter"
 	"runtime/pprof"
 	"slices"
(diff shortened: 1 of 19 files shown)


jbowens:
Got some test nondeterminism to work out

jbowens:
TFTR!

Issue #5242, opened 2025-08-26
db: read performance degradation when sharding a Pebble DB


leowkong:
To improve scalability, we sharded a single Pebble DB (~600 GB) into multiple instances (initially 8 shards). 
Total block cache and  handles are evenly split across the 8 DB instances (i.e., each shard gets 1/6 of the cache and handle budget). Each instance uses a separate cache.

Workload (blockchain system): predominantly serial/low-concurrency reads ，write batch.

However, during system read/write operations, The multi-shard version shows worse read performance compared to the single DB version.
1）Read latency is about 10% worse in the multi-shard setup. 
2） By checking Pebble’s built-in block cache metrics: The overall block cache hit rate in the multi-shard setup is about 91.4%. In contrast, the single DB setup achieves about 93.2% hit rate. 

We suspect cache locality/fragmentation is the main culprit. We tried creating one shared pebble.Cache instance and passing it into all shard DBs (8-shard test), but the results still didn’t improve meaningfully 

**Quesions:**
1. What tuning approaches or design patterns are recommended to make multi-DB/sharded Pebble perform well?
2. Is there guidance on effective cache sharing (config, sizing, admission/eviction) across DBs? 
3.   Are there known gotchas when sharing a cache (e.g., fairness, contention, memory accounting) that we should watch out for?
Jira issue: PEBBLE-1154

RaduBerinde:
Sharing the cache is the recommended setup. The cache is internally sharded into 4*GOMAXPROCS shards. What is the hit ratio when using a shared cache?

RaduBerinde:
What are you trying to achieve by splitting into multiple stores? Are the node resources far from saturation with a single store?

You can always increase compaction concurrency to allow more resources to be used for compactions. The read path should scale with the load even on one store.

leowkong:
> Sharing the cache is the recommended setup. The cache is internally sharded into 4*GOMAXPROCS shards. What is the hit ratio when using a shared cache?

Thank you .  When reading BlockCache.CacheMetrics, the hit ratio was reported at only ~2%, which doesn’t look correct. We suspect the metric may not be aggregated properly across multiple DBs (or perhaps we are misinterpreting it).

And the observed read latency is still ~10% worse compared to a single large DB.

[...]

PR #5261, opened 2025-09-01
cache: split filter and index cache when sharing cache

galaio:
Hi guys, while using Pebble, we've discovered that sharing the cache across multiple instances can be a better solution. However, we're based on Pebble 1.0, and the current pull request is a proof-of-concept. If you agree on the changes, I'll continue optimizing it to meet the merge requirements.

In our scenario, we've been using a single database, accumulating around 1.5TB of data. To reduce the pressure on the single database, we've used parallel commits and other methods. We plan to run on shards in the future, so we'll split the single database into eight shards, which will of course reduce the cache size and handles proportionally.

Although parallel commits work well, read performance and cache hit rate have decreased slightly. We've tried using a shared cache across eight shards.

Version 1:
image

image

The performance has deteriorated significantly. The flame graph shows a significant drop in the cache hit rate for readFilter. Analysis suggests this may be due to frequent evicts of the filter cache when using a shared cache.  see https://github.com/cockroachdb/pebble/issues/5242

Version 2:
image

image

We forked the Pebble code to isolate the filter and index caches. The performance difference is minimal compared to a single-DB cache.

Thank you for your review.

### DIFF
--- a/cmd/pebble/db.go
+++ b/cmd/pebble/db.go
@@ -57,9 +57,12 @@ func newPebbleDB(dir string) DB {
 	defer cache.Unref()
 	filterCache := pebble.NewCache(cacheSize)
 	defer filterCache.Unref()
+	indexCache := pebble.NewCache(cacheSize)
+	defer indexCache.Unref()
 	opts := &pebble.Options{
 		Cache:                       cache,
 		FilterCache:                 filterCache,
+		IndexCache:                  indexCache,
 		Comparer:                    mvccComparer,
 		DisableWAL:                  disableWAL,
 		FormatMajorVersion:          pebble.FormatNewest,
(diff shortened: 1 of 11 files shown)


RaduBerinde:
Thank you! These findings are interesting. I am reluctant of going down the separate caches route, as that will involve a non-trivial sizing decision. It is worth investigating why the cache policy didn't appropriately keep the filter blocks in the cache (we may benefit from a better replacement algorithm, or perhaps we need to somehow make certain misses more expensive than others).

RaduBerinde:
Do you by chance have a program that reproduces this behavior which you can share (so we can investigate more)?

RaduBerinde:
It would also be useful to see if you can repro the same improvement with a more recent Pebble (preferably 2.1)

xxmplus:
Close to due to age

Issue #5260, opened 2025-08-31
panic: Release called on a BufferPool with in-use buffers


RTS5227:
I encountered this error after upgrading Pebble from v2.0.6 to #crl-release-25.3 and changing the format major from FormatSyntheticPrefixSuffix to FormatValueSeparation. The app crashes several times a day after it occurs. I'm not sure when exactly it happens and what exactly causes it yet - I will do more testing - but wanted to put this here as a reminder.

```
panic: too many compression settings
	panic: Release called on a BufferPool with in-use buffers

goroutine 11177 [running]:
github.com/cockroachdb/pebble/sstable/block.(*BufferPool).Release(0xc0001fd7f0)
	/Users/pacx/go/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20250717220203-36d0909a10db/sstable/block/buffer_pool.go:177 +0xc5
panic({0x10a3a9400?, 0x10a655710?})
	/usr/local/Cellar/go/1.24.4/libexec/src/runtime/panic.go:792 +0x132
github.com/cockroachdb/pebble/sstable/block.(*CompressionStats).add(...)
	/Users/pacx/go/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20250717220203-36d0909a10db/sstable/block/compressor.go:139
github.com/cockroachdb/pebble/sstable/block.(*Compressor).UncompressedBlock(...)
	/Users/pacx/go/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20250717220203-36d0909a10db/sstable/block/compressor.go:102
github.com/cockroachdb/pebble/sstable/block.CopyAndChecksum(0xc0014d8c18, {0xc0055d0040?, 0x35, 0xc000e7f448?}, 0x0?, 0xc000fa8928, 0xc000fa8910)
	/Users/pacx/go/pkg/mod/github.com/c:349 +0x225
github.com/cockroachdb/pebble/sstable/blob.(*FileWriter).Close(0xc000fa8788)
	/Users/pacx/go/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20250717220203-36d0909a10db/sstable/blob/blob.go:288 +0x1a5
github.com/cockroachdb/pebble.(*writeNewBlobFiles).FinishOutput(0xc001814b60)
	/Users/pacx/go/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20250717220203-36d0909a10db/value_separation.go:303 +0x59
[...]

PR #5262, opened 2025-09-01
block: fix off-by-one bug in CompressionStats

RaduBerinde:
CompressionStats is supposed to hold up to four settings but because of an off-by-one bug it only allows three. The "built-in" profiles use at most three, but it is possible for a user to define their own profile which might require four.

Fix and add a test.

Informs #5260


xinhaoz (approved):

RaduBerinde:
TFTR!

PR #5263, opened 2025-09-02
docs: add annotation for variable blob reader count

jbowens:
Add an annotation for #5248, increasing the size of a blob.ValueFetcher's cache of readers to a variable count dependent on the read amplification. This change had a perceptible impact on write-only workloads by avoiding duplicate reads during compactions that write new blob files.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -44,6 +44,8 @@
           blocks</div>
         <div class="annotation" data-date="20250509">Enabled value
           separation (values=1024 variant)</div>
+        <div class="annotation" data-date="20250827">Vary count of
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5263#-:-OZAG3yn0qvAcqaZqLUa:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @annrpom)

jbowens:
TFTR!

PR #5264, opened 2025-09-02
blob: pool []cachedReader slices separately

jbowens:
Previously a blob.ValueFetcher's cached readers were implicitly pooled through the pooling of a pebble.Iterator. This commit updates the ValueFetcher to explicitly pool cached reader slices separately. The fetcher's cached readers are only retrieved if an external value is actually retrieved.

### DIFF
--- a/sstable/blob/fetcher.go
+++ b/sstable/blob/fetcher.go
@@ -7,6 +7,7 @@ package blob
 import (
 	"context"
 	"slices"
+	"sync"
 
 	"github.com/cockroachdb/errors"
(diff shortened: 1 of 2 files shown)


RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5264#-:-OZAGFCk-XigMoLnEcx3:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

jbowens:
TFTR!

PR #5265, opened 2025-09-02
db: add tests for EstimateDiskUsage

Sachuman:
Added tests for EstimateDiskUsage and EstimateDiskUsageByBackingType. 
Added calls to metamorphic tests, not for results but for races/panics.


RaduBerinde (approved):
:lgtm: Very nice!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5265#-:-OZG-RvGC0DzcL3Zdc_V:b-kerl87)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @jbowens, @Sachuman, and @xxmplus)

___
*[`disk_usage_test.go` line 155 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5265#-OZFyCg98xlpuc3_GXDb:-OZG-7_L7bm4gxdwVpVh:bvxtjx1) ([raw file](https://github.com/cockroachdb/pebble/blob/e3803911579713b871727637fad0ec726ddef6fd/disk_usage_test.go#L155)):*

rather than just expecting zero or nonzero, can we include an explicit value in the test results? It'll help catch instances where the calculated size changes unexpected/unintentionally.
</blockquote></details>

+1 We'll probably need to set `DisableAutomaticCompactions` in the options.

Sachuman (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5265)* status: 5 of 7 files reviewed, 2 unresolved discussions (waiting on @jbowens and @xxmplus)

___
*[`disk_usage_test.go` line 155 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5265#-OZFyCg98xlpuc3_GXDb:-OZKJmgC5HWEQKENsuRi:bfqyov9) ([raw file](https://github.com/cockroachdb/pebble/blob/e3803911579713b871727637fad0ec726ddef6fd/disk_usage_test.go#L155)):*

+1 We'll probably need to set `DisableAutomaticCompactions` in the options.
</blockquote></details>

Seems like even after DisableAutomaticCompactions set, for ingest-external, there is non-determinism for the size,
So I added the expect parameter as optional.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5265)* status: 5 of 7 files reviewed, 2 unresolved discussions (waiting on @jbowens, @Sachuman, and @xxmplus)

___
*[`disk_usage_test.go` line 155 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5265#-OZFyCg98xlpuc3_GXDb:-OZKNT3V0WqhXPVJV15b:b-n5y06k) ([raw file](https://github.com/cockroachdb/pebble/blob/e3803911579713b871727637fad0ec726ddef6fd/disk_usage_test.go#L155)):*
<details><summary><i>Previously, Sachuman (Sachin) wrote…</i></summary><blockquote>

Seems like even after DisableAutomaticCompactions set, for ingest-external, there is non-determinism for the size,
So I added the expect parameter as optional.
</blockquote></details>

There are other tests that use ingestions and are deterministic. I think we need to specify the block sizes to avoid the randomization in `runBuildRemoteCmd`

If it still doesn't work, instead of `expect-total` you can add a flag that changes any non-zero value to a special `<non-zero>` string when outputting (so we can continue to rely on matching test output)

sjain022:
TFTRs!

PR #5269, opened 2025-09-03
base: correctness comment about reads and tiering

sumeerbhola:


RaduBerinde:
If we marked all data blocks that contain only unique prefixes (unique across the sstable), we would be able to skip through such blocks in L6 as long as there are no overlapping keys in higher layers. I would imagine this is the common case with cold data.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5269)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`internal/base/internal.go` line 881 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5269#-OZFYRUa34hqOn3bDK-x:-OZKJXPZEeG_LsjpXsAi:b-2un2q) ([raw file](https://github.com/cockroachdb/pebble/blob/06118f5e4275fea639160f768ab655d50fe93a80/internal/base/internal.go#L881)):*

I currently don't think anything in this code comment is a real problem.

That is, it would be a problem if there were queries on a non-secondary index where the tiering attribute column is not part of the key (so only-value-cold), AND the query predicate included the tiering attribute column but the spans constructed by SQL could not use that predicate. In the examples we have, this doesn't happen.

In fact, all the examples where there is a predicate on the tiering attribute column or descending order specified for that column, are designed to use a secondary index, which either is doing an index join (so will naturally hit hot data in the primary index), or have a lot of storing columns.

One thing I overlooked in the phrasing of this comment is that for large secondary indexes with many storing columns, where the tiering column is part of the key (or can be made part of the key), we can successfully use only-value-cold, and we should apply block property filters, since it is correct.

We will need to strengthen the eventual consistency of the SpanPolicy to make solution part 2 work. That is, we can't allow k1@t2 to have successfully extracted its tiering attribute and k1@t1 to not have, resulting in k1@t1 in the hot tier and k1@t2 in the cold tier.

> This can be above Pebble, but we need to allow retrieving the tiering attribute (e.g. via `LazyValue`) without retrieving the value.

I think this can happen inside Pebble. The storage package above Pebble isn't currently adding any real SQL knowledge.
</blockquote></details>

Updated the comment to reflect the above.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5269#-:-OZLJ8UB1fH7WZZcUVVK:bnfp4nl)* status: 0 of 13 files reviewed, 1 unresolved discussion

___
*[`internal/base/internal.go` line 907 at r5](https://reviewable.io/reviews/cockroachdb/pebble/5269#-OZLGPIe3WKj5bsIy-qq:-OZLGPIe3WKj5bsIy-qr:bhp098u) ([raw file](https://github.com/cockroachdb/pebble/blob/49c53674fc889de0c4269689fc33f3f636c766e4/internal/base/internal.go#L907)):*
> ```Go
> // iterator using it to skip keys that are not in the interval.
> //
> // Solution part 2 requires removing the "eventual consistency" behavior of
> ```

I think we could block flushes and compactions if the relevant key range has an unknown span config. It should be very rare, only shortly after initialization or index/table creation.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5269)* status: 0 of 13 files reviewed, 1 unresolved discussion

___
*[`internal/base/internal.go` line 907 at r5](https://reviewable.io/reviews/cockroachdb/pebble/5269#-OZLGPIe3WKj5bsIy-qq:-OZLlAnt9tsSJTsmbndl:b4ine6q) ([raw file](https://github.com/cockroachdb/pebble/blob/49c53674fc889de0c4269689fc33f3f636c766e4/internal/base/internal.go#L907)):*

If we block a flush, we could have a write stall. We could disable write stalling, if we know this is < 5s or so, otherwise we risk an OOM.
</blockquote></details>

If compactions extract the attribute when it is missing, perhaps we can get away with tolerating values of 0 on L0 through L4 (and we would only block compactions into L5/L6).

PR #5270, opened 2025-09-03
db: pool scan internal iterators separately

jbowens:
The ScanInternal API allow the caller to scan the internal keys within the engine. Because these internal scans require maintaining additional state, these scans are backed by a separate 'scanInternalIterator' type. This commit bundles this along with iterAlloc into a new pooled scanInternalIterAlloc struct.

The inclusion of the scanInternalIterator struct avoids that allocation, but more importantly, the separate pool prevents ScanInternal from sharing objects with ordinary Iterators. The different access patterns of ScanInternal and Iterators can yield different buffer sizes. Since the code paths are separate, code drift can accidentally result in discarding buffers when returning an iterAlloc for reuse. Using separate pools sidesteps these concerns.

### DIFF
--- a/scan_internal.go
+++ b/scan_internal.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"slices"
+	"sync"
 
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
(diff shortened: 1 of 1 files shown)


jbowens:
TFTR!

PR #5278, opened 2025-09-05
compression: avoid msan crashes

### LINKED ISSUE #5275: master: nightly linux-msan failed
The nightly linux-msan test run failed on ff737ac39a0a36165c41c870cfd98cb911d17116 (go1.23). Please review the run [here](https://github.com/cockroachdb/pebble/actions/runs/17490118052).

Jira issue: PEBBLE-1163

RaduBerinde:
Some of the compression algorithms use hand-written assembly code,
which is "invislble" to the memory sanitizer. Together with the recent
change to use `malloc` instead of `calloc` for the block cache, this
results in uninitialized memory failures with `msan`.

The fix is to call `runtime.MSanWrite()` to inform `msan` that the
memory was written to. Note that `runtime.MSanWrite()` exists only
in builds with the `msan` tag.

Fixes #5275

### DIFF
--- a/internal/compression/minlz.go
+++ b/internal/compression/minlz.go
@@ -29,6 +29,7 @@ func (c *minlzCompressor) Compress(dst, src []byte) ([]byte, Setting) {
 	if err != nil {
 		panic(errors.Wrap(err, "minlz compression"))
 	}
+	msanWrite(compressed)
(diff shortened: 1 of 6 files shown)


RaduBerinde:
TFTR!

PR #5279, opened 2025-09-05
db: improve compaction test timeout observability

RaduBerinde:
- set up a 1 minute timeout for any datadriven compactions;
 - print db logs on test failures;
 - upgrade datadriven to include a recent change to periodically log
   when a single command takes a long time.

Informs #5277


RaduBerinde:
(pending merge of https://github.com/cockroachdb/datadriven/pull/57)

Issue #5172, opened 2025-08-12
test: stress test is not stressful enough


xxmplus:
The stress test is not stressful enough at the moment. It runs for 5 min and 100 runs maximum. We should press the codebase harder. See https://github.com/cockroachdb/pebble/blob/master/scripts/stress.sh 

```
make stress STRESSFLAGS='-maxtime 5m -maxruns 100' "PKG=$p"
```

Jira issue: PEBBLE-1134

Epic PEBBLE-1131

PR #5280, opened 2025-09-05
scripts: more improvements to stress.sh

RaduBerinde:
Increase the number of runs as many packages have very fast tests.
Also, use reduced parallelism only for the bigger packages.

Informs #5172


RaduBerinde:
TFTR!

PR #5282, opened 2025-09-05
internal/manifest: add heap maintaining virtual backing rewrite candidates

xinhaoz:
### manifest/internal: adjust VirtualBackings map to track pointers
To prepare the VirtualBackings struct to support maintaining a table backing
heap, convert the backingWithMetadata map to store pointers.

### internal/manifest: add heap maintaining virtual backing rewrite candidates
Add a rewriteCandidatesHeap which tracks the table backing with the lowest
estimated amount of referenced data.


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5282)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @annrpom, @jbowens, and @xinhaoz)

___
*[`internal/manifest/virtual_backings.go` line 363 at r5](https://reviewable.io/reviews/cockroachdb/pebble/5282#-OZjVeTm82vhHzLvXg8V:-OZjVeTm82vhHzLvXg8W:b-6nu3ua) ([raw file](https://github.com/cockroachdb/pebble/blob/a4ef2112af0e08e15cdea5f35d56307f9fcf58f8/internal/manifest/virtual_backings.go#L363)):*
> ```Go
> }
> 
> type virtualBackingRewriteCandidatesHeap struct {
> ```

[nit] Just a drive-by comment: consider using a `google/btree` instead (the [generic version](https://pkg.go.dev/github.com/google/btree#NewG)). It should be simpler to use - in particular we wouldn't need the annoying `heapIndex` field. It is probably a tad more efficient than a heap in practice too.

xinhaoz:
I decided not to pursue the btree since it seems like we would need to remove and then reinsert the element because there's no 'fixup' operation :(

PR #5289, opened 2025-09-08
db: clean up compaction test infrastructure

RaduBerinde:
#### db: fix TestTableStats options reuse

In `TestTableStats` any options passed to `define` persist through to
the next `define`, which is very unexpected.

We fix this by re-initializing the options each time.

#### db: split up TestCompaction

We split off the main part of `TestCompaction`. Now each test file
initializes new state (which is less fragile and allows reproducing a
specific test file failure with a given seed).

#### db: remove unused runIngestCmd argument

In some tests (`TestCompaction`), the passed argument is incorrect
anyway (because `runDBDefineCmd` overwrites the FS we think we're
using).

#### db: clean up test code for set-concurrent-compactions

We no longer require copying `CompactionConcurrencyRange` from the old
options.

#### db: clean up runDBDefineCmd

Consolidate the two variants of this function: we create a new `MemFS`
only if `opts.FS` is not set (which is much more intuitive).

Note that we no longer call `EnsureDefaults()` before modifying the
options passed to the `define` directive. This changes certain values
that have defaults that depends on other values, like
`FlushSplitBytes`. We add a `flush-split-byte` argument to correct
this.

### DIFF
--- a/compaction_picker_test.go
+++ b/compaction_picker_test.go
@@ -1463,20 +1463,8 @@ func TestCompactionPickerScores(t *testing.T) {
 	fs := vfs.NewMem()
 	cleaner := pausableCleaner{cleaner: DeleteCleaner{}}
 	cleaner.cond.L = &cleaner.mu
-	opts := &Options{
-		Cleaner:                     &cleaner,
-		Comparer:                    testkeys.Comparer,
-		DisableAutomaticCompactions: true,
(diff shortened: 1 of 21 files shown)


RaduBerinde:
Friendly ping

RaduBerinde:
TFTR!

PR #5292, opened 2025-09-09
db: periodically recheck the memtable write stall condition

sumeerbhola:
The period is set to 1s, by writing at that interval to the condition variable.

A future PR will make Options.MemTableStopWritesThreshold dynamic.

### DIFF
--- a/db.go
+++ b/db.go
@@ -2341,6 +2341,7 @@ func (d *DB) maybeInduceWriteStall(b *Batch) {
 	// This function will call EventListener.WriteStallBegin at most once.  If
 	// it does call it, it will call EventListener.WriteStallEnd once before
 	// returning.
+	var timer *time.Timer
 	for {
 		var size uint64
 		for i := range d.mu.mem.queue {
@@ -2357,7 +2358,18 @@ func (d *DB) maybeInduceWriteStall(b *Batch) {
 				})
 			}
 			beforeWait := crtime.NowMono()
+			// NB: In a rare case, we can start a write stall, and then the system
+			// may detect WAL failover, resulting in
+			// ElevateWriteStallThresholdForFailover returning true. So we want to
+			// recheck the predicate periodically, which we do by signaling the
+			// condition variable.
+			if timer == nil {
+				timer = time.AfterFunc(time.Second, d.mu.compact.cond.Broadcast)
+			} else {
+				timer.Reset(time.Second)
+			}
 			d.mu.compact.cond.Wait()
+			timer.Stop()
 			if b != nil {
 				b.commitStats.MemTableWriteStallDuration += beforeWait.Elapsed()
 			}
--- a/db_test.go
+++ b/db_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/cockroachdb/datadriven"
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/buildtags"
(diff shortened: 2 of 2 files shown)


sumeerbhola (commented):
I'll add a test if the approach looks ok.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5292#-:-OZk1es545BEhez5K3nM:bjs35l9)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @annrpom)

RaduBerinde:
*[`open.go` line 1259 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5292#-OZk4Qj11N-uocaKMh7I:-OZo6Dsi6i-kF_zCwKp8:bbm2yb) ([raw file](https://github.com/cockroachdb/pebble/blob/a8ecee8fbfeea8f7907c9a50734ae96fc7e462d9/open.go#L1259)):*

Good point about doing something inband. Definitely don't want to sleep for a second since the stall may be shorter and want to be reactive.

I tried that inband thing and it introduces 2 allocations, one 112 bytes (the call to `newTimer` inside `time.AfterFunc`) and the other 16 bytes, which looks to be the allocation for the closure which needs a pointer to `d.mu.compact.cond` and to the func being called. So it isn't viable given the hot path this code is on.
</blockquote></details>

Is this a hot path? I thought write stalls are rare

RaduBerinde:
*[`open.go` line 1259 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5292#-OZk4Qj11N-uocaKMh7I:-OZo6_AB7mvJDr2j7V6b:b-judpuv) ([raw file](https://github.com/cockroachdb/pebble/blob/a8ecee8fbfeea8f7907c9a50734ae96fc7e462d9/open.go#L1259)):*

Is this a hot path? I thought write stalls are rare
</blockquote></details>

I don't like adding another background goroutine per store if we can avoid it, stack/goroutine dumps have gotten very clunky.

RaduBerinde:
*[`open.go` line 1259 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5292#-OZk4Qj11N-uocaKMh7I:-OZo86WZ9EzPv_CIWftF:b-yb5s3r) ([raw file](https://github.com/cockroachdb/pebble/blob/a8ecee8fbfeea8f7907c9a50734ae96fc7e462d9/open.go#L1259)):*

I don't like adding another background goroutine per store if we can avoid it, stack/goroutine dumps have gotten very clunky.
</blockquote></details>

I believe the AfterFunc timer can be set up once then reused with Reset.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5292)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @annrpom and @RaduBerinde)

___
*[`open.go` line 1259 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5292#-OZk4Qj11N-uocaKMh7I:-OZoyk0k4U1lCmAMiZgL:b-szicpa) ([raw file](https://github.com/cockroachdb/pebble/blob/a8ecee8fbfeea8f7907c9a50734ae96fc7e462d9/open.go#L1259)):*
> I believe the AfterFunc timer can be set up once then reused with Reset.

Good point. Done.

sumeerbhola:
I increased the timeout from 15s to 1min under race, since even though `stressrace` on my laptop did not fail, each race run was taking ~7s, and CI did fail under race.

PR #5294, opened 2025-09-10
cache: don't update hit/miss metrics for compaction reads

RaduBerinde:
To improve the usefulness of the cache hit metrics, we no longer
update them for compaction reads and CopySpan cache accesses (these
paths are not latency-sensitive).


RaduBerinde:
I don't think there's agreement on this change, closing.

PR #5299, opened 2025-09-10
metrics: more granular block cache metrics

RaduBerinde:
#### metricsutil: add Window data structure

Window maintains instances of metric at regular intervals and is able
to return the metrics from ~10 minutes and ~1hr ago. This is useful to
provide statistics (e.g. cache hits) from these timeframes.

#### metrics: more granular block cache metrics

We improve the cache metrics in two ways:
 - we break down the misses by the type of block (for the most
   important block types)
 - we also show miss rates across the last 10 minutes and the last
   hour

In follow-up work, I will attempt to plumb the LSM level as well.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -857,7 +857,7 @@ func TestMemTableReservation(t *testing.T) {
 		t.Fatalf("expected 2 refs, but found %d", refs)
 	}
 	// Verify the memtable reservation has caused our test block to be evicted.
-	if cv := tmpHandle.Get(base.DiskFileNum(0), 0); cv != nil {
+	if cv := tmpHandle.Get(base.DiskFileNum(0), 0, cache.CategoryBackground); cv != nil {
 		t.Fatalf("expected failure, but found success: %#v", cv)
 	}
 
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -17,19 +17,8 @@ import (
 
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/invariants"
+	"github.com/cockroachdb/pebble/internal/metricsutil"
 )
-
-// Metrics holds metrics for the cache.
-type Metrics struct {
-	// The number of bytes inuse by the cache.
-	Size int64
-	// The count of objects (blocks or tables) in the cache.
-	Count int64
-	// The number of cache hits.
-	Hits int64
-	// The number of cache misses.
-	Misses int64
-}
 
 // Cache implements Pebble's sharded block cache. The Clock-PRO algorithm is
 // used for page replacement
@@ -77,6 +66,8 @@ type Cache struct {
 	maxSize int64
 	idAlloc atomic.Uint64
 	shards  []shard
+
+	metricsWindow *metricsutil.Window[HitsAndMisses]
 
 	// Traces recorded by Cache.trace. Used for debugging.
 	tr struct {
@@ -136,6 +127,8 @@ func NewWithShards(size int64, shards int) *Cache {
 	for i := range c.shards {
 		c.shards[i].init(size / int64(len(c.shards)))
 	}
+	c.metricsWindow = metricsutil.NewWindow[HitsAndMisses](c.hitsAndMisses)
+	c.metricsWindow.Start()
 
 	// Note: this is a no-op if invariants are disabled or race is enabled.
 	invariants.SetFinalizer(c, func(c *Cache) {
@@ -171,10 +164,19 @@ func (c *Cache) Unref() {
 	case v < 0:
 		panic(fmt.Sprintf("pebble: inconsistent reference count: %d", v))
 	case v == 0:
-		for i := range c.shards {
-			c.shards[i].Free()
-		}
+		c.destroy()
 	}
+}
+
+func (c *Cache) destroy() {
+	if c.metricsWindow != nil {
+		c.metricsWindow.Stop()
+		c.metricsWindow = nil
+	}
+	for i := range c.shards {
+		c.shards[i].Free()
+	}
+	c.shards = nil
 }
 
 func (c *Cache) NewHandle() *Handle {
@@ -205,21 +207,6 @@ func (c *Cache) Reserve(n int) func() {
 		}
 		shardN = -1
 	}
-}
-
-// Metrics returns the metrics for the cache.
-func (c *Cache) Metrics() Metrics {
-	var m Metrics
-	for i := range c.shards {
-		s := &c.shards[i]
-		s.mu.RLock()
-		m.Count += int64(s.blocks.Len())
-		m.Size += s.sizeHot + s.sizeCold
-		s.mu.RUnlock()
-		m.Hits += s.hits.Load()
-		m.Misses += s.misses.Load()
-	}
-	return m
 }
 
 // MaxSize returns the max size of the cache.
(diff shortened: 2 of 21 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5299)* status: 0 of 21 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`internal/metricsutil/window_test.go` line 18 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5299#-OZpXD1m5rrA0b0qhyv6:-OZpXD1m5rrA0b0qhyv7:bmnej8t) ([raw file](https://github.com/cockroachdb/pebble/blob/dd5a3aac10fb579e97f7d68b1a073660e0286a2e/internal/metricsutil/window_test.go#L18)):*
> ```Go
> 
> func TestWindow(t *testing.T) {
> 	synctest.Test(t, func(t *testing.T) {
> ```

This new synctest thing is awesome: it "freezes" time as long as any goroutine (spawned by the func) is doing work, then fast forwards to the earliest time that would wake up a timer, repeat.

sumeerbhola (approved):
:lgtm: 

> @sumeerbhola reviewed 2 of 21 files at r3, 2 of 2 files at r4, 21 of 21 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5299#-:-O_E3DCZDgdpArQy27Vn:bg7igkv)* status: all files reviewed, 5 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/metricsutil/window_test.go` line 18 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5299#-OZpXD1m5rrA0b0qhyv6:-O_E-wmQ9FVmRHnGH4D3:b-7m74ox) ([raw file](https://github.com/cockroachdb/pebble/blob/dd5a3aac10fb579e97f7d68b1a073660e0286a2e/internal/metricsutil/window_test.go#L18)):*

This new synctest thing is awesome: it "freezes" time as long as any goroutine (spawned by the func) is doing work, then fast forwards to the earliest time that would wake up a timer, repeat.
</blockquote></details>

nice!
___
*[`internal/cache/cache_test.go` line 256 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5299#-O_Dy-qP6z-GkGgQ5H3J:-O_Dy-qP6z-GkGgQ5H3K:b-jhe9eq) ([raw file](https://github.com/cockroachdb/pebble/blob/66894bb1f47607fbf204b64374be4431c606de83/internal/cache/cache_test.go#L256)):*
> ```Go
> 	// relatively few shards, the probability that any bucket is more than double
> 	// the expected size is vanishingly small.
> 	cache := New(2 * size)
> ```

I don't quite understand the change, in that earlier we were multiplying by GOMAXPROCS and now we are multiplying by 2. Could you elaborate on the reason for the change?
___
*[`internal/cache/cache_test.go` line 271 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5299#-O_DxqGiDzxlpLRbBNfc:-O_DxqGiDzxlpLRbBNfd:b22akjj) ([raw file](https://github.com/cockroachdb/pebble/blob/66894bb1f47607fbf204b64374be4431c606de83/internal/cache/cache_test.go#L271)):*
> ```Go
> 			offset := pcg.Uint64() % size
> 			v := h.Get(base.DiskFileNum(0), offset)
> 			if v != nil {
> ```

Given the `% size` above, when will this be `nil`?
___
*[`testdata/compaction/l0_to_lbase_compaction` line 73 at r5](https://reviewable.io/reviews/cockroachdb/pebble/5299#-O_E2oqm6bqSgW2F5dPB:-O_E2oqm6bqSgW2F5dPC:bwudxr5) ([raw file](https://github.com/cockroachdb/pebble/blob/c08c8de2522b890639eeafb52c9afe5bbc7faa2f/testdata/compaction/l0_to_lbase_compaction#L73)):*
> ```
>    all |  background |     sstdata |      sstval |     blobval |      filter |       index
> -------+-------------+-------------+-------------+-------------+-------------+------------
>  60.0% | 100% [0.2%] |   40% [99%] | 0.0% [0.0%] | 0.0% [0.0%] | 0.0% [0.0%] |  40% [0.6%]
> ```

shouldn't this have been in the background category for compactions?
___
*[`internal/cache/metrics.go` line 63 at r5](https://reviewable.io/reviews/cockroachdb/pebble/5299#-O_E1j6DDLx_-B-LTWn_:-O_E1j6DDLx_-B-LTWna:b-n4jk9x) ([raw file](https://github.com/cockroachdb/pebble/blob/c08c8de2522b890639eeafb52c9afe5bbc7faa2f/internal/cache/metrics.go#L63)):*
> ```Go
> 	// Recent contains cache hit metrics covering two recent periods (last ~10
> 	// minutes and last ~1 hour).
> 	Recent [2]struct {
> ```

Consider defining an enum for this, so we can index using 10MinWindow, 1hWindow instead of 0, 1.

RaduBerinde:
I will post a follow-up PR which also splits by the levels and I will merge them together. This is so I can cleanly run the benchmark before/after.

sumeerbhola (commented):
:thumbsup: 

> @sumeerbhola reviewed 14 of 22 files at r11.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5299#-:-O_IEdtX541yV63Edj7F:b4c9lyr)* status: 14 of 22 files reviewed, 3 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/cache/metrics.go` line 63 at r5](https://reviewable.io/reviews/cockroachdb/pebble/5299#-O_E1j6DDLx_-B-LTWn_:-O_ICU4mAIElR0IlX_YS:b2nhb2c) ([raw file](https://github.com/cockroachdb/pebble/blob/c08c8de2522b890639eeafb52c9afe5bbc7faa2f/internal/cache/metrics.go#L63)):*

I don't want to hardcode these periods at this layer. I have a TODO to switch to using the `Since` field instead of assuming they are 10m and 1hr.
</blockquote></details>

Ack
___
*[`internal/cache/cache_test.go` line 256 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5299#-O_Dy-qP6z-GkGgQ5H3J:-O_ICub_1DZPJ1ry1JAM:bdd6xw4) ([raw file](https://github.com/cockroachdb/pebble/blob/66894bb1f47607fbf204b64374be4431c606de83/internal/cache/cache_test.go#L256)):*
>We are trying to cache `size` objects.

Isn't this the block cache, so we are caching `size` bytes? 2MiB is very modest given that we have some unit tests like `TestWriterClearCache` that are using 64MiB caches, and there are existing benchmarks `BenchmarkIteratorScanNextPrefix` that use 200MiB caches.

> I wanted to use the standard calculation for the number of shards (the one in `New()`, which is 4\*GOMAXPROCS)

I see - so K is bigger now than before to be more realistic.

RaduBerinde:
*[`internal/cache/cache_test.go` line 256 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5299#-O_Dy-qP6z-GkGgQ5H3J:-O_JpDjX-HN-OVMGL6yv:bahu09m) ([raw file](https://github.com/cockroachdb/pebble/blob/66894bb1f47607fbf204b64374be4431c606de83/internal/cache/cache_test.go#L256)):*

>We are trying to cache `size` objects.

Isn't this the block cache, so we are caching `size` bytes? 2MiB is very modest given that we have some unit tests like `TestWriterClearCache` that are using 64MiB caches, and there are existing benchmarks `BenchmarkIteratorScanNextPrefix` that use 200MiB caches.

> I wanted to use the standard calculation for the number of shards (the one in `New()`, which is 4\*GOMAXPROCS)

I see - so K is bigger now than before to be more realistic.
</blockquote></details>

We're caching `size` objects of 1 byte each.

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5299#-:-O_NO4tu9zOfefNHmg4f:bg7igkv)* status: 14 of 22 files reviewed, 3 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/cache/cache_test.go` line 256 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5299#-O_Dy-qP6z-GkGgQ5H3J:-O_NO13Y8H9J9Aq3inrl:bj3xlwf) ([raw file](https://github.com/cockroachdb/pebble/blob/66894bb1f47607fbf204b64374be4431c606de83/internal/cache/cache_test.go#L256)):*

We're caching `size` objects of 1 byte each.
</blockquote></details>

ah, so the overhead is high

RaduBerinde:
Merged in #5321

PR #5310, opened 2025-09-13
metrics: minor improvements

RaduBerinde:
#### table: pass rows as slice

This simplifies most call sites.

#### table: support horizontal dividers

We add a `RenderOptions` field which allows customizing where the
horizontal dividers are put.

#### metrics: don't use horizontal table rendering

The horizontal table rendering is very unintuitive. We no longer need
it, given that we can customize both the vertical and horizontal
dividers. Change the only table that is using it.

#### table: remove horizontal orientation


RaduBerinde:
TFTR!

PR #5316, opened 2025-09-15
db: assert that blob file rewrites reduce value size

jbowens:
When performing a blob file rewrite compaction, assert that the rewritten blob file has a smaller value size. We should never choose to rewrite a blob file that is fully referenced. If we do, it reflects a bug in our statistics or compaction picking.

Informs #5306.


jbowens:
TFTR!

(this was directed at Radu, not the robot)

PR #5321, opened 2025-09-16
cache: separate hit/miss metrics by level

RaduBerinde:
This PR is only for the top 2 commits. The rest is #5299.

#### block: hide cache hits/misses from the level checker

We introduce a "reason" for using a `BufferPool`. We use this reason
to determine what category to pass to the cache. We introduce a
special `CategoryHidden` which doesn't update any metrics; we use this
for the level checker. Since the level checker is a testing facility
only, it only adds noise to the metrics.

#### cache: separate hit/miss metrics by level

We further break down the block cache metrics by level.

The level is plumbed through the level iterator. However, it is not
currently plumbed for the initial open of a file, and it is not
plumbed for blob value reads. I will investigate plumbing a level
through `LazyFetcher` separately.

The n/a level reflects accesses where the level is not known and any
accesses on flushable ingests.

Metrics after 30m of ycsb-A:
```
BLOCK CACHE: 592 entries (17MB)
                 miss rate [percentage of total misses] since start
 level |            |  background    sstdata       sstval      blobval       filter       index
-------+------------+------------------------------------------------------------------------------
   n/a | 2.3% [91%] | 9.6% [7.4%]                             4.1% [81%]               0.2% [3.3%]
    L0 | 22% [4.6%] |  36% [1.3%]    24% [2%]                                           15% [1.3%]
    L6 | 0% [4.1%]  |   0% [0%]     0% [2.8%]                                           0% [1.3%]
 total |    0.3%    |  10% [8.7%]   0% [4.8%]                 4.1% [81%]                0% [5.9%]
```

Benchmark (before = without the level AND the category changes). On a
GCE worker, `--benchtime=10s`:

```
name         old time/op  new time/op  delta
CacheGet-24  52.0ns ± 1%  44.0ns ± 1%  -15.40%  (p=0.008 n=5+5)
```


sumeerbhola (approved):
:lgtm: 

> @sumeerbhola reviewed 26 of 27 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5321#-:-O_X3w4y0OoNl3KLmq29:bg7igkv)* status: 26 of 28 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`-- commits` line 44 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5321#-O_Wz6uUD4upATxwu_Rt:-O_Wz6uUD4upATxwu_Ru:bm5f46t):*
surprised to see 91% of accesses being n/a. Any theory?
I suppose the low miss rate is because of the zipfian distribution, since the cache size of 17MB is probably much smaller than the size of the LSM. Does that seem correct?
___
*[`internal/cache/metrics.go` line 157 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5321#-O_X03pL6lPp7TSUeG5a:-O_X03pL6lPp7TSUeG5b:b1mko7r) ([raw file](https://github.com/cockroachdb/pebble/blob/5032fa050cd37b869bdb83711b033a312236110c/internal/cache/metrics.go#L157)):*
> ```Go
> // (across all levels).
> func (hm *HitsAndMisses) AggregateCategory(category Category) (hits, misses int64) {
> 	for level := range hm {
> ```

nit: it is less confusing to name this index `i`, instead of `level` since it isn't a `Level`
___
*[`sstable/block/block.go` line 277 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5321#-O_X0WztAlSENttugG9u:-O_X0WztAlSENttugG9v:b-wzhnd3) ([raw file](https://github.com/cockroachdb/pebble/blob/5032fa050cd37b869bdb83711b033a312236110c/sstable/block/block.go#L277)):*
> ```Go
> 	BufferPool *BufferPool
> 	// Level is the LSM level associated with the operation. It must be set whenever
> 	// the reader is using the block cache (i.e. ReaderOptions.CacheOpts is set).
> ```

Should this comment narrow its scope to additionally say that this is when reading a (potentially virtual) sstable, since that belongs to a level?
Since this approach will not work for blob files, when we extend coverage to blob files in the future.

Also, I am never clear on the sharing behavior of `ReadEnv`. Looks like it isn't shared because we copy the struct while sharing the pointers inside it. Something concrete about this in the declaration comment for ReadEnv would be helpful.

Unrelated to this PR, I don't know if `ReportCorruptionFn` and `ReportCorruptionArg` are being properly setup for blob files.
https://github.com/cockroachdb/pebble/blob/998a55aac5bb6da774c2a00b72a1551ac73c5340/blob_rewrite.go#L315 is setting `ReportCorruptionFn` but not the arg. And the `ReadEnv` passed to `blob.ValueFetcher.Init` doesn't set either -- possibly because the `ValueFetcher` is shared. Is there a gap in our corruption reporting? @jbowens

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5321#-:-O_XdiKI5i7Df-7iDrh8:b-814g8k)* status: 0 of 28 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`-- commits` line 44 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5321#-O_Wz6uUD4upATxwu_Rt:-O_XbLOK60SSCSQAyfKK:b-m82cbv):*

surprised to see 91% of accesses being n/a. Any theory?
I suppose the low miss rate is because of the zipfian distribution, since the cache size of 17MB is probably much smaller than the size of the LSM. Does that seem correct?
</blockquote></details>

These are blob values for which we are not plumbing the level (at least not yet).
___
*[`sstable/block/block.go` line 277 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5321#-O_X0WztAlSENttugG9u:-O_XcbvK8aXT9e7hxk2O:bautevq) ([raw file](https://github.com/cockroachdb/pebble/blob/5032fa050cd37b869bdb83711b033a312236110c/sstable/block/block.go#L277)):*

Should this comment narrow its scope to additionally say that this is when reading a (potentially virtual) sstable, since that belongs to a level?
Since this approach will not work for blob files, when we extend coverage to blob files in the future.

Also, I am never clear on the sharing behavior of `ReadEnv`. Looks like it isn't shared because we copy the struct while sharing the pointers inside it. Something concrete about this in the declaration comment for ReadEnv would be helpful.

Unrelated to this PR, I don't know if `ReportCorruptionFn` and `ReportCorruptionArg` are being properly setup for blob files.
https://github.com/cockroachdb/pebble/blob/998a55aac5bb6da774c2a00b72a1551ac73c5340/blob_rewrite.go#L315 is setting `ReportCorruptionFn` but not the arg. And the `ReadEnv` passed to `blob.ValueFetcher.Init` doesn't set either -- possibly because the `ValueFetcher` is shared. Is there a gap in our corruption reporting? @jbowens
</blockquote></details>

Updated the comment and added a comment for `ReadEnv`.

Filed https://github.com/cockroachdb/pebble/issues/5343 for the other thing.

RaduBerinde:
The tests are flaky because of `CheckLevels`. When it runs during move compactions, the same access can show up as different levels.

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5321#-:-O_Yj099C4sHfPrPmGpA:bg7igkv)* status: 0 of 29 files reviewed, all discussions resolved (waiting on @jbowens)

RaduBerinde:
I added a commit that hides the hits/misses from the level checker (and provides an easy way to better choose the category for any work that uses `BufferPool`s). PTAL

RaduBerinde:
*[`testdata/compaction/value_separation` line 436 at r13](https://reviewable.io/reviews/cockroachdb/pebble/5321#-O_nd_Tu2jB0Mjknj4H3:-O_nd_Tu2jB0Mjknj4H4:bpidhqq) ([raw file](https://github.com/cockroachdb/pebble/blob/ea36b2462ac3eeddcd0d4c9f8043516dbd8baffb/testdata/compaction/value_separation#L436)):*
> ```
> level      all     |  background    sstdata       sstval      blobval       filter       index
> -------------------+------------------------------------------------------------------------------
>  n/a   79% [100%]  |  100% [20%]   60% [20%]                                           82% [60%]
> ```

There is still an issue I'm looking into, where the level checker doesn't plumb the `ReadEnv` to `gatherBlobHandles` and some of that shows up here.

RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5321#-:-O_ng6L-Bm-Ds0XJy_Ns:b-814g8k)* status: 0 of 40 files reviewed, 2 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`sstable/block/block.go` line 387 at r12](https://reviewable.io/reviews/cockroachdb/pebble/5321#-O_ndGtF42x1EsVaUQoy:-O_nf0UzCkBb1AQ7rF1u:b-hk1e1z) ([raw file](https://github.com/cockroachdb/pebble/blob/065904710d4b0848432f7b855404c41a94dceb75/sstable/block/block.go#L387)):*

Why are these not in `kindToCacheCategory`?
</blockquote></details>

Added a comment. I initially wanted to hide compactions altogether but I decided to add a "background" category and put it all there.

PR #5323, opened 2025-09-16
metrics: hide empty level metrics

RaduBerinde:
#### table: replace AutoIncrement with StringWithTupleIndex

This is more powerful as it allows customization of the output. We no
longer need a hack to emit the "total" for the per-level metric
tables.

#### table: support filtering

Support filtering of rows, useful with `StringWithTupleIndex` which
will get the original tuple index.

We also support relative-to-end indices in `HorizontalDividers`.

#### metrics: hide empty level metrics

Hide rows that have all zeros in the level and compaction metrics.
Also, use `Lx` instead of just `x` for the level.


RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5323#-:-O_NMA3G62mEaz4CaoP4:b-814g8k)* status: all files reviewed, 1 unresolved discussion (waiting on @annrpom)

___
*[`testdata/metrics` line 461 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5323#-O_NFss85bLY_Td8vJ6y:-O_NM8EJCYl82RoWXPbZ:b8gj3mm) ([raw file](https://github.com/cockroachdb/pebble/blob/b33397f5f8147d47517679e6ddbd216def83fd74/testdata/metrics#L461)):*

we filter out L0 here, but it has some table bytes in & write amp info populated - should we keep this row if these are present?
</blockquote></details>

Good catch, fixed!

PR #5324, opened 2025-09-16
cockroachkvs: add functions for MaximumSuffixProperty interface to cockroachkvs

Sachuman:
**cockroachkvs: add functions for MaximumSuffixProperty interface to cockroachkvs**
 
**sstable: fix issue with comparing suffix during SeekPrefixGE lazy positioning**

During a RangeDelIter the prefix passed might not be the same the prefix of the key
therefore we need to split the prefix fromt the given key.

**metamorphic: add changes to iterOpts in metamorphic settings for new SeekPrefix lazy positioning**

Added metamorphic changes with 20% probability to add new lazy positioning optimization in iterator opts

**sstable: fix i.synthetic.atSyntheticKey check in Next**

we should check if key is synthetic before invalidation check

### DIFF
--- a/cockroachkvs/blockproperties.go
+++ b/cockroachkvs/blockproperties.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble"
+	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/sstable"
 )
 
@@ -143,4 +144,52 @@ func mapSuffixToInterval(b []byte) (sstable.BlockInterval, error) {
 		return sstable.BlockInterval{Lower: ts, Upper: ts + 1}, nil
 	}
 	return sstable.BlockInterval{}, nil
+}
+
+type MaxMVCCTimestampProperty struct{}
+
+// Name is part of the cockroachkvs.MaxMVCCTimestampProperty interface.
+func (MaxMVCCTimestampProperty) Name() string {
+	return mvccWallTimeIntervalCollector
+}
+
+// Extract is part of the cockroachkvs.MaxMVCCTimestampProperty interface.
+// It extracts the maximum MVCC timestamp from the encoded block property and
+// returns it as a CockroachDB-formatted suffix.
+func (MaxMVCCTimestampProperty) Extract(
+	dst []byte, encodedProperty []byte,
+) (suffix []byte, ok bool, err error) {
+	if len(encodedProperty) <= 1 {
+		return nil, false, nil
+	}
+	// First byte is shortID, skip it and decode interval from remainder.
+	buf := encodedProperty[1:]
+	if len(buf) == 0 {
+		return nil, false, nil
+	}
+	// Decode the block interval using the same logic as sstable.decodeBlockInterval
+	var interval sstable.BlockInterval
+	var n int
+	interval.Lower, n = binary.Uvarint(buf)
+	if n <= 0 || n >= len(buf) {
+		return nil, false, base.CorruptionErrorf("cannot decode interval from buf %x", buf)
+	}
+	pos := n
+	interval.Upper, n = binary.Uvarint(buf[pos:])
+	pos += n
+	if pos != len(buf) || n <= 0 {
+		return nil, false, base.CorruptionErrorf("cannot decode interval from buf %x", buf)
+	}
+	// Delta decode.
+	interval.Upper += interval.Lower
+	if interval.Upper < interval.Lower {
+		return nil, false, base.CorruptionErrorf("unexpected overflow, upper %d < lower %d", interval.Upper, interval.Lower)
+	}
+	if interval.IsEmpty() {
+		return nil, false, nil
+	}
+	dst = append(dst, make([]byte, 9)...)
+	binary.BigEndian.PutUint64(dst[len(dst)-9:], interval.Upper)
+	dst[len(dst)-1] = 9
+	return dst, true, nil
 }
--- a/metamorphic/cockroachkvs.go
+++ b/metamorphic/cockroachkvs.go
@@ -162,6 +162,11 @@ func (kg *cockroachKeyGenerator) IncMaxSuffix() []byte {
 	return kg.suffixSpace.ToMaterializedSuffix(s)
 }
 
+// MaximumSuffixProperty returns the maximum suffix property.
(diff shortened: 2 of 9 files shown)


jbowens (commented):
looking good!

Read up to the metamorphic test commit but not the last commit. I have some suggestions on the metamorphic test integration. If you want, you can pull the other commits out into separate PRs to get them merged earlier

> @jbowens reviewed 1 of 1 files at r1, 9 of 9 files at r9, 9 of 9 files at r10, 2 of 7 files at r11.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5324#-:-O_Nzf1k9KOIZl3OwTsl:b9hmtio)* status: 3 of 10 files reviewed, 8 unresolved discussions (waiting on @xxmplus)

___
*[`metamorphic/config.go` line 445 at r11](https://reviewable.io/reviews/cockroachdb/pebble/5324#-O_NzXZaCfOMTYJ_q00f:-O_NzXZaCfOMTYJ_q00g:b-zy845) ([raw file](https://github.com/cockroachdb/pebble/blob/4566a7247a1013feff5df876b2290377d9205082/metamorphic/config.go#L445)):*
> ```Go
> 	// MaximumSuffixProperty returns the maximum suffix property used during
> 	// the lazy position of SeekPrefixGE optimization.
> 	MaximumSuffixProperty() pebble.MaximumSuffixProperty
> ```

instead of putting this on the generator, I think this should be a field on the `KeyFormat` in metamorphic/config.go
___
*[`cockroachkvs/blockproperties.go` line 151 at r9](https://reviewable.io/reviews/cockroachdb/pebble/5324#-O_NwLC99eImAaBik6bW:-O_NwLC99eImAaBik6bX:bt008on) ([raw file](https://github.com/cockroachdb/pebble/blob/97fede945395c53b5aa2649f12fd1d3f2319f32f/cockroachkvs/blockproperties.go#L151)):*
> ```Go
> type MaxMVCCTimestampProperty struct{}
> 
> // Name is part of the cockroachkvs.MaxMVCCTimestampProperty interface.
> ```

nit: let's say "Name implements the pebble.MaximumSuffixProperty interface."
___
*[`cockroachkvs/blockproperties.go` line 156 at r9](https://reviewable.io/reviews/cockroachdb/pebble/5324#-O_Nw_6sBWZSfG_Cc22N:-O_Nw_6sBWZSfG_Cc22O:b6e45a7) ([raw file](https://github.com/cockroachdb/pebble/blob/97fede945395c53b5aa2649f12fd1d3f2319f32f/cockroachkvs/blockproperties.go#L156)):*
> ```Go
> }
> 
> // Extract is part of the cockroachkvs.MaxMVCCTimestampProperty interface.
> ```

s/Extract implements the pebble.MaximumSuffixProperty interface./
___
*[`cockroachkvs/blockproperties.go` line 170 at r9](https://reviewable.io/reviews/cockroachdb/pebble/5324#-O_Nx2n1D1u6Uubmy08k:-O_Nx2n1D1u6Uubmy08l:b-jd500p) ([raw file](https://github.com/cockroachdb/pebble/blob/97fede945395c53b5aa2649f12fd1d3f2319f32f/cockroachkvs/blockproperties.go#L170)):*
> ```Go
> 		return nil, false, nil
> 	}
> 	// Decode the block interval using the same logic as sstable.decodeBlockInterval
> ```

nit: I think we should export `sstable.DecodeBlockInterval` and call it here since we're tightly coupled to the sstable's block interval property encoding
___
*[`cockroachkvs/blockproperties.go` line 193 at r9](https://reviewable.io/reviews/cockroachdb/pebble/5324#-O_NxZzx71cHma18YRU0:-O_NxZzx71cHma18YRU1:b-xduozs) ([raw file](https://github.com/cockroachdb/pebble/blob/97fede945395c53b5aa2649f12fd1d3f2319f32f/cockroachkvs/blockproperties.go#L193)):*
> ```Go
> 	dst = append(dst, make([]byte, 9)...)
> 	binary.BigEndian.PutUint64(dst[len(dst)-9:], interval.Upper)
> 	dst[len(dst)-1] = 9
> ```

nit: instead of inlining `9`, let's use `suffixLenWithWall`
___
*[`sstable/reader_iter_two_lvl.go` line 470 at r10](https://reviewable.io/reviews/cockroachdb/pebble/5324#-O_NyQXa96kQhGqoXhkz:-O_NyQXa96kQhGqoXhl-:b-9ufxvw) ([raw file](https://github.com/cockroachdb/pebble/blob/fc77e758b5239b669d9e8a4a08c7da30d4d42f95/sstable/reader_iter_two_lvl.go#L470)):*
> ```Go
> 		// the top of the iterator's heap, and the iterator is Nexted.
> 		// During a RangeDelIter, the prefix passed and the prefix in the key might not be the same.
> 		if ok && maxSuffix != nil && i.secondLevel.reader.Comparer.ComparePointSuffixes(key[len(i.secondLevel.reader.Comparer.Split.Prefix(key)):], maxSuffix) < 0 {
> ```

ditto: regarding calling `Split` directly
___
*[`sstable/reader_iter_single_lvl.go` line 867 at r10](https://reviewable.io/reviews/cockroachdb/pebble/5324#-O_NyEsSBWRc2hqUZzpg:-O_NyEsSBWRc2hqUZzph:bfnp9up) ([raw file](https://github.com/cockroachdb/pebble/blob/fc77e758b5239b669d9e8a4a08c7da30d4d42f95/sstable/reader_iter_single_lvl.go#L867)):*
> ```Go
> 		// the top of the iterator's heap, and the iterator is Nexted.
> 		// During a RangeDelIter, the prefix passed and the prefix in the key might not be the same.
> 		if ok && maxSuffix != nil && i.reader.Comparer.ComparePointSuffixes(key[len(i.reader.Comparer.Split.Prefix(key)):], maxSuffix) < 0 {
> ```

nit: to extract the suffix index, we can just call `Split` directly like:`key[i.reader.Comparer.Split(key):]`
___
*[`metamorphic/ops.go` line 581 at r11](https://reviewable.io/reviews/cockroachdb/pebble/5324#-O_NysOz1TZtIVEePfu1:-O_NysOz1TZtIVEePfu2:b-96nqzt) ([raw file](https://github.com/cockroachdb/pebble/blob/4566a7247a1013feff5df876b2290377d9205082/metamorphic/ops.go#L581)):*
> ```Go
> }
> 
> func deserializeMaximumSuffixProperty(s string) pebble.MaximumSuffixProperty {
> ```

rather than serializing the name of the implementation, I think we just serialize "maxsuffixprop" vs "none"; then when `MaximumSuffixProperty` we use can be determined by the `KeyFormat` used by the test

PR #5329, opened 2025-09-17
db: make tests faster

RaduBerinde:
The runtime for the top-level package tests has been creeping up. This is an assorted set of small adjustments to improve the running time.

#### db: make TestCompaction/l0_to_lbase_compaction faster

This test took ~1s, now it takes 0.06s.

#### db: make TestCompaction/value_separation faster

This test was taking ~1s, now it's under 0.1s.

#### db: reduce TestOpenCloseOpenClose configurations

This saves about 0.5s.

#### db: remove duplicate TestFileCacheClockPro

The genericcache code already has this test.

#### db: reduce numOps in TestWALFailoverRandomized

This reduces run time from 4s to ~1.5s on average.

#### db: reduce numOps in TestWALHardCrashRandomized

This reduces the average run time from ~4s to ~0.5s.

### DIFF
--- a/file_cache_test.go
+++ b/file_cache_test.go
@@ -5,14 +5,11 @@
 package pebble
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"fmt"
 	"math/rand/v2"
-	"os"
 	"path"
-	"strconv"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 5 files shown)


RaduBerinde:
Hm, `TestCompaction/value_separation` is now flaky..  The reason is just file numbering. Seems like the creation of a new WAL can happen during the flush and that affects the table numbers. How do other compaction tests go around this issue? I guess previously we had bigger files so an sstable wouldn't finish so quickly.. but relying on that seems fragile.

```
    logger.go:15: [JOB 4] flushing 2 memtables (1.3MB) to L0
    logger.go:15: [JOB 4] flushing: sstable created 000005
    logger.go:15: [JOB 4] flushing: blob file created 000007
    logger.go:15: [JOB 5] WAL created 000006
```
vs
```
    logger.go:15: [JOB 4] WAL created 000005
    logger.go:15: [JOB 5] flushing 3 memtables (1.3MB) to L0
    logger.go:15: [JOB 5] flushing: sstable created 000006
    logger.go:15: [JOB 5] flushing: blob file created 000007
```

RaduBerinde:
Ah, the problem here was that sometimes we had 2 memtables to flush, sometimes 3. Fixed by setting a higher memtable size.

RaduBerinde:
TFTR!

PR #5331, opened 2025-09-17
manifest: apply a few small hygiene improvements to L0 sublevels 

jbowens:
**manifest: apply a few small hygiene improvements to L0 sublevels**

Apply a few small code hygiene improvements, like using min and max builtins
and the slices.Clone function.

**manifest: use slices.SortFunc**

Replace a lingering implementation of the sort.Interface with a use of
slices.SortFunc.


jbowens:
TFTR!

PR #5333, opened 2025-09-18
db: compactions: calculate eventual output level

RaduBerinde:
#### db: minor compaction cleanup

We no longer calculate `maxOutputFileSize` and `maxOverlapBytes`
inside `pickedTableCompaction`, as they are not used there. We
calculate them in `newCompaction` instead.

We no longer store a `maxReadCompactionBytes` field; we calculate it
in the one place we need it.

#### db: compactions: calculate eventual output level

We improve the compaction code to check if the levels below the
compaction bounds are empty, in which case we adjust the output file
size and other sstable writer options to correspond to the eventual
level (after move compactions).


jbowens (approved):
:lgtm:

> @jbowens reviewed 5 of 5 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5333#-:-O_Wx7NJ1pGbKYMCD_Wq:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sumeerbhola)

RaduBerinde:
*[`compaction.go` line 276 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5333#-O_XD2Ba8sBUTVuGI0Rl:-O_rKZgGBRlLTU_KDc_P:b-bkezx6) ([raw file](https://github.com/cockroachdb/pebble/blob/84d608d51380df063d653edb8a291b87b23c416e/compaction.go#L276)):*

> I can explore that. It would reduce some manifest churn as well

I haven't really given this more than passing thought, wrt whether there are write amp downsides to it.
It will require some care wrt coordination with ingests.
</blockquote></details>

Yeah, it looks like not an easy change. I will probably merge this as-is but after branch cut.

PR #5336, opened 2025-09-18
cockroachkvs: add functions for MaximumSuffixProperty interface to cockroachkvs

sjain022:
**_cockroachkvs: add functions for MaximumSuffixProperty interface to cockroachkvs_**

1. Changed decodeBlockInterval as exported function

**_sstable: fix issue with comparing suffix during SeekPrefixGE lazy positioning_**

1. During a RangeDelIter the prefix passed might not be the same the prefix of the key
therefore we need to split the prefix from the given key.

**_sstable: fix i.synthetic.atSyntheticKey check in Next_**

1. we should check if key is synthetic before invalidation check


sjain022:
@jbowens made this separate PR . need a review on this one to change metamorphic lazy load PR. Thanks!

sjain022:
TFTR!

PR #5337, opened 2025-09-18
sstable: optimize blobReferenceValues allocation

jbowens:
When writing a sstable containing references to blob files, the sstable writer constructs the blob reference liveness index—an accounting of which values in which blob files are referenced. Once we've finalized the data for an individual block within a blob file, we encode the data to a buffer through appending. These appends show up as 3% of all allocations in heap profiles from a recent run of the pebble ycsb/D/values=1024 benchmark. This commit adds a call to slices.Grow to avoid unnecessary intermediate allocations. Additionally, when constructing state for a new blob file, we allocate an initial 512B under the expectation that most referenced blob files will require around 512B to encode all the values that are referenced from the sstable.

### DIFF
--- a/sstable/blob_reference_index.go
+++ b/sstable/blob_reference_index.go
@@ -7,6 +7,7 @@ package sstable
 import (
 	"encoding/binary"
 	"iter"
+	"slices"
 
 	"github.com/cockroachdb/errors"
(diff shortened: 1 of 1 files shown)


annrpom (approved):
:lgtm:

jbowens:
TFTR!

PR #5338, opened 2025-09-19
[tiered prototype] db: misc fixes to write cold blob files

sumeerbhola:
The TestMetrics is flaky wrt compaction stats for one test case, which I've narrowed down the existing TieringMeta not being returned by the compact.Iter tree in some cases, which causes the compaction to reextract it (which requires fetching the value which would otherwise not need to be fetched). This changes the block read stats.

I don't plan to hunt down the bug immediately since it is harmless for our current purposes.


sumeerbhola:
TFTR!

PR #5341, opened 2025-09-19
Revert "cache: use malloc instead of calloc"

jbowens:
This reverts commit dd58b8432fe899bfeeb55feb3ee76cf9d129b0e0.

The pebble ycsb benchmarks show a regression around the time this commit was merged. A comparison of the CPU profiles before and after show ~5.32% time in _Cfunc_calloc before and ~12.10% time in _cgo_cmalloc after.

### DIFF
--- a/internal/cache/value.go
+++ b/internal/cache/value.go
@@ -59,7 +59,7 @@ func Alloc(n int) *Value {
 
 	if valueEntryCanBeGoAllocated && valueEntryGoAllocated {
 		// Note: if cgo is not enabled, manual.New will do a regular Go allocation.
-		b := manual.NewUninitialized(manual.BlockCacheData, uintptr(n))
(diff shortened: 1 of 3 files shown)


RaduBerinde (approved):
Wow.. that is surprising.  I'll see if I can figure out why.

RaduBerinde:
Hm, I'm reading that malloc can have more upfront cost whereas calloc can cause more page faults later. So the percentage in the profile might not tell the full picture. Really surprising this does cause an overall regression..

jbowens:
Hm, it's definitely possible the regression is caused by another commit in the same time, but given the CPU profile this does seem like the most likely culprit: https://github.com/cockroachdb/pebble/commits/master/?since=2025-09-03&until=2025-09-06

RaduBerinde:
I guess we'll know after this merges. If it doesn't resolve it, it's likely the lazy block load.

PR #5346, opened 2025-09-19
cache: don't mark blocks as accessed for background work

RaduBerinde:
We don't want block cache hits for compactions to cause the respective
block to stay in the cache longer (in fact, it is about to become
obsolete). This change adds a `Peek()` method to the block cache which
is like `Get()` but does not mark the block as recently accessed.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -858,7 +858,7 @@ func TestMemTableReservation(t *testing.T) {
 		t.Fatalf("expected 2 refs, but found %d", refs)
 	}
 	// Verify the memtable reservation has caused our test block to be evicted.
-	if cv := tmpHandle.Get(base.DiskFileNum(0), 0); cv != nil {
+	if cv := tmpHandle.Peek(base.DiskFileNum(0), 0); cv != nil {
 		t.Fatalf("expected failure, but found success: %#v", cv)
 	}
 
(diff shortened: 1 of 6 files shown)


sumeerbhola (approved):
:lgtm: 

> @sumeerbhola reviewed 5 of 6 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5346#-:-O_n1yCt2miBNcV21e5l:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens)

___
*[`internal/cache/clockpro.go` line 157 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5346#-O_n19eHBBoq2NwzTNt7:-O_n19eHBBoq2NwzTNt8:b-ggossh) ([raw file](https://github.com/cockroachdb/pebble/blob/6b6864054f2a7e3cd8d5627afa99b465dcad482a/internal/cache/clockpro.go#L157)):*
> ```Go
> // getWithReadEntry is the internal helper for implementing
> // Cache.{Get,GetWithReadHandle}. When desireReadEntry is true, and the block
> // is not in the cache (nil Value), a non-nil readEntry is returned (in which
> ```

nit: the `desireReadEntry` part of this comment is stale.

Issue #5347, opened 2025-09-22
DB.Close() hangs due to uninterruptible deletion pacing


ijsong:
When `TargetByteDeletionRate` is set to a low value, `DB.Close()` can hang for a long time because deletion pacing uses `time.Sleep()` that can't be interrupted.

The `cleanupManager.Close()` comment says ["Delete pacing is disabled for the remaining jobs"](https://github.com/cockroachdb/pebble/blob/489b133bc3a12164b261fd6d2447167cacb2a276/obsolete_files.go#L119-L121) but this isn't actually implemented.

```go
// obsolete_files.go
func (cm *cleanupManager) maybePace(...) {
    for {
        ok, d := tb.TryToFulfill(tokenbucket.Tokens(tokens))
        if ok {
            break
        }
        time.Sleep(d)  // Can't be interrupted during Close()
    }
}
```

Reproduction

1. Set `TargetByteDeletionRate` to 1MB/s
2. Create a DB with many obsolete files  
3. Call `DB.Close()` - it hangs until all sleep delays finish

PR #5348, opened 2025-09-22
fix cleanupManager.Close() hanging when deletion pacing is active

ijsong:
When `TargetByteDeletionRate` is set to a low value, `DB.Close()` can hang for a long time because deletion pacing uses `time.Sleep()` that can't be interrupted.

This change modifies `maybePace()` to check for early termination and use interruptible `select` instead of `time.Sleep()`.

Resolves: #5347

### DIFF
--- a/obsolete_files.go
+++ b/obsolete_files.go
@@ -41,6 +41,8 @@ type cleanupManager struct {
 	jobsCh chan *cleanupJob
 	// waitGroup is used to wait for the background goroutine to exit.
 	waitGroup sync.WaitGroup
+	// stopCh is closed when the pacer is disabled after closing the cleanup manager.
+	stopCh chan struct{}
 
 	mu struct {
 		sync.Mutex
@@ -103,6 +105,7 @@ func openCleanupManager(
 			getDeletePacerInfo,
 		),
 		jobsCh: make(chan *cleanupJob, jobsQueueDepth),
+		stopCh: make(chan struct{}),
 	}
 	cm.mu.completedJobsCond.L = &cm.mu.Mutex
 	cm.waitGroup.Add(1)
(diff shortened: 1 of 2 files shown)


ijsong:
I'm using the release versions, so I'd appreciate it if you could include this fix in the next release.

ijsong:
*[`obsolete_files_test.go` line 169 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5348#-O_lUbyv9KVijy73lFRR:-O_tFkXZB08r0JgBdOJz:bt2vr1b) ([raw file](https://github.com/cockroachdb/pebble/blob/f4125be3b5f795ba8679f434062e30cece137332/obsolete_files_test.go#L169)):*

The expected time should be significantly larger than the test timeout below which is also 30s. I'd bump it up to 100 files.
</blockquote></details>

Fixed.

RaduBerinde (commented):
> I'm using the release versions, so I'd appreciate it if you could include this fix in the next release.

Are you using 2.0.x or 2.1.0?  (2.1 is significantly newer, I'd use that if possible)

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5348#-:-O_tOFp_Ab4Qk9tVFvOS:bhrzgd7)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @annrpom)

ijsong:
Actually, I am using v1.1.5 :sweat_smile:  But I am preparing to bump to v2.1.0.

RaduBerinde:
Hm, I commented out `close(stopCh)` and the test still passes..

RaduBerinde:
https://github.com/cockroachdb/pebble/pull/5361

PR #5350, opened 2025-09-22
arenaskl: use pointer comparison to avoid key comparison

jbowens:
When inserting a new node into the memtable skiplist, we must determine which
nodes within the skiplist will be to the left and right of the new node. When
computing this "slice" we descend to each level, moving along the level until
we find a key greater than the one we're inserting.

Our skiplist implementation has a max height of 20, so we must consider at most
20 levels. With each additional level, the probability a node existing at that
level increases by the inverse of Euler's number (> 36%). It's not unlikely
that one node may point to the same node on consecutive levels of the skiplist.

This commit adapts the splice-finding code to take advantage of this porperty
by checking for pointer equality between next nodes of consecutive levels. If
we observe a pointer to a node at height h and we already recorded the node as
'next' within the splice at height h+1, we can elide the key comparison.

```
goos: linux
goarch: amd64
pkg: github.com/cockroachdb/pebble/internal/arenaskl
                      │     a.txt     │               c.txt                │
                      │    sec/op     │    sec/op     vs base              │
ReadWrite/frac_0-24      247.1n ±  2%   236.6n ±  1%  -4.23% (p=0.002 n=6)
ReadWrite/frac_10-24     233.8n ±  1%   225.5n ±  1%  -3.53% (p=0.002 n=6)
ReadWrite/frac_20-24     219.7n ±  1%   211.9n ±  0%  -3.55% (p=0.002 n=6)
ReadWrite/frac_30-24     206.7n ±  1%   199.0n ±  1%  -3.70% (p=0.002 n=6)
ReadWrite/frac_40-24     192.2n ±  1%   183.8n ±  2%  -4.37% (p=0.002 n=6)
ReadWrite/frac_50-24     176.2n ±  1%   169.2n ±  2%  -3.97% (p=0.002 n=6)
ReadWrite/frac_60-24     163.0n ±  2%   157.1n ±  2%  -3.62% (p=0.002 n=6)
ReadWrite/frac_70-24     143.7n ±  3%   140.7n ±  4%  -2.09% (p=0.015 n=6)
ReadWrite/frac_80-24     124.6n ±  4%   121.7n ±  6%  -2.37% (p=0.048 n=6)
ReadWrite/frac_90-24    101.81n ±  6%   93.70n ±  4%  -7.97% (p=0.026 n=6)
ReadWrite/frac_100-24    10.41n ± 22%   10.45n ± 57%       ~ (p=0.818 n=6)
OrderedWrite-24          101.5n ±  0%   101.0n ±  0%  -0.54% (p=0.004 n=6)
geomean                  132.0n         127.6n        -3.32%
```

### DIFF
--- a/internal/arenaskl/skl.go
+++ b/internal/arenaskl/skl.go
@@ -387,8 +387,18 @@ func (s *Skiplist) findSplice(key base.InternalKey, ins *Inserter) (found bool) 
 	for level = level - 1; level >= 0; level-- {
 		for {
 			// Assume prev.key < key.
-			next = s.getNext(prev, level)
-			if next == s.tail {
+			nextInLevel := s.getNext(prev, level)
+			// next is currently set to the previous level's `next`, the first
+			// node whose key is ≥ `key` within the previous level. If
+			// nextInLevel's equals the last level's `next`, then `next` is also
+			// the first node whose is key ≥ `key` within this level as well.
+			//
+			// This pointer comparison avoids an extra key comparison.
+			if nextInLevel == next {
+				break
+			}
+			next = nextInLevel
+			if nextInLevel == s.tail {
 				// Tail node, so done.
 				break
 			} else if next == nil {
--- a/internal/arenaskl/skl.go
+++ b/internal/arenaskl/skl.go
@@ -383,11 +383,41 @@ func (s *Skiplist) findSplice(key base.InternalKey, ins *Inserter) (found bool) 
 		}
 	}
 
+	var next *node
 	for level = level - 1; level >= 0; level-- {
-		var next *node
-		prev, next, found = s.findSpliceForLevel(key, level, prev)
-		if next == nil {
-			next = s.tail
+		for {
+			// Assume prev.key < key.
+			next = s.getNext(prev, level)
+			if next == s.tail {
+				// Tail node, so done.
+				break
+			} else if next == nil {
+				next = s.tail
+				break
+			}
+
+			offset, size := next.keyOffset, next.keySize
+			nextKey := s.arena.buf[offset : offset+size]
+			cmp := s.cmp(key.UserKey, nextKey)
+			if cmp < 0 {
+				// We are done for this level, since prev.key < key < next.key.
+				break
+			}
+			if cmp == 0 {
+				// User-key equality.
+				if key.Trailer == next.keyTrailer {
+					// Internal key equality.
+					found = true
+					break
+				}
+				if key.Trailer > next.keyTrailer {
+					// We are done for this level, since prev.key < key < next.key.
+					break
+				}
+			}
+
+			// Keep moving right on this level.
+			prev = next
 		}
 		ins.spl[level].init(prev, next)
 	}


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5350)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/arenaskl/skl.go` line 389 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5350#-O_l_X1ZCF20eYxEWYmz:-O_mmJTuAm0X-ifabsf9:b-uxlwye) ([raw file](https://github.com/cockroachdb/pebble/blob/489b133bc3a12164b261fd6d2447167cacb2a276/internal/arenaskl/skl.go#L389)):*

I'm a bit confused by the existing code - how can `findSpliceForLevel` return a nil `next`? (unless s.tail itself is `nil`, in which case `next = s.tail` is a no-op)
</blockquote></details>

Hmm, yeah, that was unnecessary.
___
*[`internal/arenaskl/skl.go` line 391 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5350#-O_laEOw4aUbe1W05pkx:-O_mmOpO0Cgs9oIc7RzU:b-frey54) ([raw file](https://github.com/cockroachdb/pebble/blob/845fe69927da55c95bc7f2dae9db2121b22642db/internal/arenaskl/skl.go#L391)):*

This is only true in the first iteration of the inner loop, no? I get that the pointer equality will never happen except in the first iteration but it's kind of confusing.

Consider doing a single `getNext` before the inner loop and making this special check then:
```
prevLevelNext := next
next = s.getNext(prev, level)
// check next == prevLevelNext
...
for {
  if next == s.tail { break }
  ...
  prev = next
  next = s.getNext(prev, level)
}
```
</blockquote></details>

It is possible to encounter the previous level's next after the first iteration if all the intermediate nodes are less than the sought key. I cleaned up the code and commentary a bit and included a little diagram.

RaduBerinde:
*[`internal/arenaskl/skl.go` line 377 at r13](https://reviewable.io/reviews/cockroachdb/pebble/5350#-O_rCx3u8tYgRUKKSwPC:-O_rLW-17pWwOe5VGdnF:b-efmj5l) ([raw file](https://github.com/cockroachdb/pebble/blob/c2a7682b46981a2b0794d6b82c4922688431d1c0/internal/arenaskl/skl.go#L377)):*

typo?
</blockquote></details>

It's concerning that tests passed, we should look at adding a test.

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5350)* status: 0 of 3 files reviewed, 8 unresolved discussions (waiting on @jbowens and @RaduBerinde)

___
*[`internal/arenaskl/skl.go` line 377 at r13](https://reviewable.io/reviews/cockroachdb/pebble/5350#-O_rCx3u8tYgRUKKSwPC:-O_wNT3n4mG__QeiDXtR:by6g463) ([raw file](https://github.com/cockroachdb/pebble/blob/c2a7682b46981a2b0794d6b82c4922688431d1c0/internal/arenaskl/skl.go#L377)):*

I agree that the skiplist is under-tested and a randomized unit test would be helpful, but imo it shouldn't block merging improvements here.
</blockquote></details>

I would definitely agree if this were a bug fix or pressing in some way. But since this is an optimization, and changing critical code which has not been changed in a while, getting very solid coverage seems prudent.
I am possibly being paranoid, so if @RaduBerinde doesn't see this as a risk, then :lgtm:

RaduBerinde:
*[`internal/arenaskl/skl.go` line 377 at r13](https://reviewable.io/reviews/cockroachdb/pebble/5350#-O_rCx3u8tYgRUKKSwPC:-O_wSQlq4DA1JzzZYLju:bc7vz8f) ([raw file](https://github.com/cockroachdb/pebble/blob/c2a7682b46981a2b0794d6b82c4922688431d1c0/internal/arenaskl/skl.go#L377)):*

I would definitely agree if this were a bug fix or pressing in some way. But since this is an optimization, and changing critical code which has not been changed in a while, getting very solid coverage seems prudent.
I am possibly being paranoid, so if @RaduBerinde doesn't see this as a risk, then :lgtm:
</blockquote></details>

Knowing Jackson, he will follow up with a test very soon regardless ^_^

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5350)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`internal/arenaskl/skl.go` line 377 at r13](https://reviewable.io/reviews/cockroachdb/pebble/5350#-O_rCx3u8tYgRUKKSwPC:-Oa6uDcX04RO5gJNdIWX:b-ugqs34) ([raw file](https://github.com/cockroachdb/pebble/blob/c2a7682b46981a2b0794d6b82c4922688431d1c0/internal/arenaskl/skl.go#L377)):*

Knowing Jackson, he will follow up with a test very soon regardless ^_^
</blockquote></details>

I was wrong, this isn't resolved by the CAS—it's resolved within `findSplice` itself. With `==` instead of `!=` we would ascend to `level=listHeight` and perform a fresh seek down below. It's not possible to catch this with a unit test without pulling the code out of the function.

jbowens:
TFTRs!

PR #5345, opened 2025-09-19
colblk: remove BlockDecoder allocation

jbowens:
```
goos: darwin
goarch: arm64
pkg: github.com/cockroachdb/pebble/sstable/colblk
cpu: Apple M1 Pro
                        │   old.txt   │               new.txt               │
                        │   sec/op    │   sec/op     vs base                │
DataBlockDecoderInit-10   94.53n ± 0%   82.12n ± 0%  -13.13% (p=0.000 n=25)

                        │  old.txt   │              new.txt               │
                        │    B/op    │   B/op     vs base                 │
DataBlockDecoderInit-10   48.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=25)

                        │  old.txt   │               new.txt               │
                        │ allocs/op  │ allocs/op   vs base                 │
DataBlockDecoderInit-10   1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=25)
```

Fix https://github.com/cockroachdb/pebble/issues/5339.

PR #5352, opened 2025-09-22
cockroachkvs: avoid cockroachKeysSeeker.init allocation

jbowens:
The change in https://github.com/cockroachdb/pebble/pull/5345 prevented BlockDecoder from escaping to the heap when using
the testkeys.KeySchema, but with the cockroachkvs.KeySchema the key seeker init
func's local BlockDecoder still escaped. This commit updates the key seeker to
call colblk.DecodeColumn directly, which is enough for the Go compiler to avoid
moving the arg to the heap.

```
goos: darwin
goarch: arm64
pkg: github.com/cockroachdb/pebble/cockroachkvs
cpu: Apple M1 Pro
InitDataBlockMetadata-10   135.5n ± 1%   110.8n ± 0%  -18.20% (p=0.002 n=6)

                         │  old.txt   │              new.txt              │
                         │    B/op    │   B/op     vs base                │
InitDataBlockMetadata-10   48.00 ± 0%   0.00 ± 0%  -100.00% (p=0.002 n=6)

                         │  old.txt   │              new.txt               │
                         │ allocs/op  │ allocs/op   vs base                │
InitDataBlockMetadata-10   1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.002 n=6)
```

### DIFF
--- a/cockroachkvs/cockroachkvs.go
+++ b/cockroachkvs/cockroachkvs.go
@@ -747,12 +747,20 @@ var _ uint = colblk.KeySeekerMetadataSize - uint(unsafe.Sizeof(cockroachKeySeeke
 
 var _ colblk.KeySeeker = (*cockroachKeySeeker)(nil)
 
-func (ks *cockroachKeySeeker) init(d *colblk.DataBlockDecoder, bd colblk.BlockDecoder) {
-	ks.roachKeys = bd.PrefixBytes(cockroachColRoachKey)
+func (ks *cockroachKeySeeker) init(
+	d *colblk.DataBlockDecoder,
+	//gcassert:noescape
+	bd colblk.BlockDecoder,
+) {
+	// Note that we call DecodeColumn directly rather than using the
+	// BlockDecoder's type-specific convenience methods (e.g. PrefixBytes,
+	// Uints, etc) to avoid bd from escaping to the heap.
+	rows := int(bd.Rows())
+	ks.roachKeys = colblk.DecodeColumn(&bd, cockroachColRoachKey, rows, colblk.DataTypePrefixBytes, colblk.DecodePrefixBytes)
 	ks.roachKeyChanged = d.PrefixChanged()
-	ks.mvccWallTimes = bd.Uints(cockroachColMVCCWallTime)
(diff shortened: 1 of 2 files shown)


jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5352#-:-O_m6ePp06dOxuCcxvXO:b-814g8k)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @xinhaoz)

___
*[`cockroachkvs/cockroachkvs.go` line 750 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5352#-O_m13O76VrKx3IcsuSx:-O_m6fYlCZjrwuB_yfCb:bg0g6ge) ([raw file](https://github.com/cockroachdb/pebble/blob/ca2d54860a3d3b6de412bdee30957ad540b20da1/cockroachkvs/cockroachkvs.go#L750)):*

We can do this:
```
func (ks *cockroachKeySeeker) init(
	d *colblk.DataBlockDecoder,
	//gcassert:noescape
	bd colblk.BlockDecoder,
) {
```

I verified that TestLint/GCAssert fails with this on master and passes with the change. I'd also add one on `DecodeColumn`:
```
func DecodeColumn[V any](
	//gcassert:noescape
	d *BlockDecoder,
	col int,
	rows int,
	dataType DataType,
	decodeFunc DecodeFunc[V],
) V {
```
</blockquote></details>

Good idea. Does the `DecodeColumn` one do what we want, since that's parameter is a pointer? Does it ensure that the pointer itself doesn't escape to the heap or that the value pointed to does not escape?
___
*[`cockroachkvs/cockroachkvs.go` line 751 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5352#-O_m18pV-O80b7eCHDNo:-O_m6qkS8U-PJhupUcTH:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/ca2d54860a3d3b6de412bdee30957ad540b20da1/cockroachkvs/cockroachkvs.go#L751)):*

[nit] add a comment explaining that we're not using `bd` methods to avoid escaping.
</blockquote></details>

Done

PR #5359, opened 2025-09-23
tieredmeta: add commentary and code for generalizing the histogram

sumeerbhola:
Just a proof-of-concept with no test (so bugs are likely).

### DIFF
--- a/sstable/tieredmeta/histogram.go
+++ b/sstable/tieredmeta/histogram.go
@@ -6,10 +6,10 @@ package tieredmeta
 
 import (
 	"encoding/binary"
+	"math"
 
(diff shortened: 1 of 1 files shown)


RaduBerinde (commented):
(KLL would need some relatively minor modifications to support weights, but there are also other options like t-digest: https://pkg.go.dev/github.com/caio/go-tdigest  https://arxiv.org/pdf/1902.04023)

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5359#-:-O_tPqao3YcqbJM7ILHu:b-88eo32)* status: 0 of 1 files reviewed, all discussions resolved

Issue #4416, opened 2025-03-24
wal: validate recovery directory through a stable identifier


jbowens:
When recovering a database previously configured with WAL failover, the OPTIONS file encodes the directory that was previously used as a secondary (which may contain relevant WAL files that need to be replayed). Open will error out if this directory encoded within the OPTIONS file is not provided as a recovery directory or the current WAL failover secondary.

However it's also possible for an operator to accidentally remount the wrong disks in the wrong places, so although the correct directory path is configured, its contents are incorrect. We should persist a stable identifier (a UUID?) to both the OPTIONS file and a file within the secondary directory. If recovery finds that the secondary directory does not contain a matching identifier, we can abort recovery indicating that the secondary seems incorrect / corrupt.

This would've helped with a recent DRT test cluster issue where the loss of a VM's host forced a migration of a VM, and the node came back up with incorrect disk mountpoints.

Jira issue: PEBBLE-358

Epic PEBBLE-1158

tbg:
Adding O-support due to possible connection to https://cockroachdb.zendesk.com/agent/tickets/27469

PR #5360, opened 2025-09-24
wal: validate recovery directory through a stable identifier

annrpom:
We now persist a stable identifier to both the OPTIONS file and a file within the secondary directory (failover_identifier). If recovery finds that the secondary directory does not contain a matching identifier, we abort recovery indicating that the secondary seems incorrect / corrupt.

Fixes: #4416

### DIFF
--- a/open.go
+++ b/open.go
@@ -372,6 +372,7 @@ func Open(dirname string, opts *Options) (db *DB, err error) {
 	}
 	if opts.WALFailover != nil {
 		walOpts.Secondary = opts.WALFailover.Secondary
+		walOpts.SecondaryIdentifier = opts.WALFailover.SecondaryIdentifier
 		// Lock the secondary WAL directory, if distinct from the data directory
 		// and primary WAL directory.
 		if secondaryWalDirName != dirname && secondaryWalDirName != walDirname {
@@ -420,6 +421,12 @@ func Open(dirname string, opts *Options) (db *DB, err error) {
 	walManager, err := wal.Init(walOpts, wals)
 	if err != nil {
 		return nil, err
+	}
+
+	// Update the pebble options with the secondary identifier if it was
+	// generated.
+	if opts.WALFailover != nil && walManager.Opts().SecondaryIdentifier != "" {
+		opts.WALFailover.SecondaryIdentifier = walManager.Opts().SecondaryIdentifier
 	}
 	defer maybeCleanUp(walManager.Close)
 	d.mu.log.manager = walManager
--- a/open_test.go
+++ b/open_test.go
@@ -256,6 +256,9 @@ func TestOpen_WALFailover(t *testing.T) {
 			if o.FS == nil {
 				return "no path"
 			}
+			// Set a constant identifier for testing to avoid flaky tests
+			wal.SetGenerateStableIdentifierForTesting("9f69f2c3ffb3c247767290a9b3215fc5")
+			defer wal.ResetGenerateStableIdentifierForTesting()
 			d, err := Open(dataDir, o)
 			if err != nil {
 				return err.Error()
--- a/options.go
+++ b/options.go
@@ -1380,6 +1380,13 @@ type WALFailoverOptions struct {
 	// preacquire the lock on the secondary directory.
 	Secondary wal.Dir
 
+	// SecondaryIdentifier is a stable UUID that uniquely identifies the secondary
+	// directory. This identifier is persisted both in the OPTIONS file in the
+	// primary directory and in a file (failover_identifier) within the secondary
+	// directory to detect incorrectness/corruptions (e.g. when the wrong disk
+	// has been mounted at the expected path during recovery).
+	SecondaryIdentifier string
+
 	// FailoverOptions provides configuration of the thresholds and intervals
 	// involved in WAL failover. If any of its fields are left unspecified,
 	// reasonable defaults will be used.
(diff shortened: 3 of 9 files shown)


jbowens (commented):
> @jbowens reviewed 1 file, made 5 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5360)* status: 1 of 10 files reviewed, 6 unresolved discussions (waiting on @annrpom and @xxmplus).

___
*[`open.go` line 99 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5360#-OiKFbQOEG1vpaiNgkNV:-OiKFbQOEG1vpaiNgkNW:b-o9tsnz) ([raw file](https://github.com/cockroachdb/pebble/blob/ad95561396034396f0ca577dc99ae61a8faf338e/open.go#L99)):*
> ```Go
> 
> 	// Recover the current database state.
> 	rs, err := recoverState(opts, dirname)
> ```

So `opts` is passed in by the caller of `Open`, so it doesn't reflect what's currently in the `OPTIONS` file. I think we need to update `recoverState` so that when it reads the previous `OPTIONS` file it also extracts the recovery IDs encoded within the options file and sticks it in the appropriate `wal.Dir`s in `recoveredState.resolvedDirs`. Then when we validate, that's what we validate against.
___
*[`wal/failover_manager.go` line 121 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5360#-OiKHQGT3XCqtTOYjNii:-OiKHQGT3XCqtTOYjNij:bl6yzb) ([raw file](https://github.com/cockroachdb/pebble/blob/ad95561396034396f0ca577dc99ae61a8faf338e/wal/failover_manager.go#L121)):*
> ```Go
> 	}
> 
> 	return errors.CombineErrors(f.Sync(), f.Close())
> ```

I think we also need to sync the containing directory to ensure the new file we added to the directory is durably synced too :disappointed: 
___
*[`wal/failover_manager.go` line 629 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5360#-OiKECU65f_a0RKxCcuo:-OiKECU65f_a0RKxCcup:b-k8hm70) ([raw file](https://github.com/cockroachdb/pebble/blob/ad95561396034396f0ca577dc99ae61a8faf338e/wal/failover_manager.go#L629)):*
> ```Go
> 		if err != nil {
> 			return Dir{}, errors.Newf("failed to read secondary identifier: %v", err)
> 		}
> ```

nit: in both of the cases of the outer conditional we call `readSecondaryIdentifier` immediately, so I think we should lift that up into the main body of the func.

we should probably also return out of one condition and outdent the other, stylistically
___
*[`options.go` line 2873 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5360#-OiKCAGo83YIC77QrHEn:-OiKCAGo83YIC77QrHEo:b-yy5l1b) ([raw file](https://github.com/cockroachdb/pebble/blob/ad95561396034396f0ca577dc99ae61a8faf338e/options.go#L2873)):*
> ```Go
> // ensure that we're using the correct directory.
> func (o *Options) validateWALRecoveryDirIdentifier(d wal.Dir) error {
> 	identifierFile := d.FS.PathJoin(d.Dirname, "failover_identifier")
> ```

nit: I think we should call this filename `stable_identifier` or something similar, because I could see us using a similar technique for the other directories which are not for failover.
___
*[`options.go` line 2891 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5360#-OiKDTgIB8IFG4_4oJpF:-OiKDTgIB8IFG4_4oJpG:bggtoo2) ([raw file](https://github.com/cockroachdb/pebble/blob/ad95561396034396f0ca577dc99ae61a8faf338e/options.go#L2891)):*
> ```Go
> 		return errors.Newf("failed to read secondary identifier from WALRecoveryDir %q: %v",
> 			d.Dirname, err)
> 	}
> ```

this `err != nil` branch is redundant w/ the one above

annrpom (commented):
> @annrpom made 5 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5360)* status: 0 of 11 files reviewed, 4 unresolved discussions (waiting on @jbowens and @xxmplus).

___
*[`open.go` line 99 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5360#-OiKFbQOEG1vpaiNgkNV:-OinUt7v87e7DP6CmPM3:bhel2ra) ([raw file](https://github.com/cockroachdb/pebble/blob/ad95561396034396f0ca577dc99ae61a8faf338e/open.go#L99)):*

So `opts` is passed in by the caller of `Open`, so it doesn't reflect what's currently in the `OPTIONS` file. I think we need to update `recoverState` so that when it reads the previous `OPTIONS` file it also extracts the recovery IDs encoded within the options file and sticks it in the appropriate `wal.Dir`s in `recoveredState.resolvedDirs`. Then when we validate, that's what we validate against.
</blockquote></details>

ah done
___
*[`options.go` line 2873 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5360#-OiKCAGo83YIC77QrHEn:-OinUuqf23dSSJ42SYS-:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/ad95561396034396f0ca577dc99ae61a8faf338e/options.go#L2873)):*

nit: I think we should call this filename `stable_identifier` or something similar, because I could see us using a similar technique for the other directories which are not for failover.
</blockquote></details>

done
___
*[`options.go` line 2891 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5360#-OiKDTgIB8IFG4_4oJpF:-OinUvTP69yEksWdhLI_:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/ad95561396034396f0ca577dc99ae61a8faf338e/options.go#L2891)):*

this `err != nil` branch is redundant w/ the one above
</blockquote></details>

done
___
*[`wal/failover_manager.go` line 121 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5360#-OiKHQGT3XCqtTOYjNii:-OinUwPlBLcVDqSUrb3i:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/ad95561396034396f0ca577dc99ae61a8faf338e/wal/failover_manager.go#L121)):*

I think we also need to sync the containing directory to ensure the new file we added to the directory is durably synced too :disappointed:
</blockquote></details>

done
___
*[`wal/failover_manager.go` line 629 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5360#-OiKECU65f_a0RKxCcuo:-OinUxHl16vpuma79iIg:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/ad95561396034396f0ca577dc99ae61a8faf338e/wal/failover_manager.go#L629)):*

nit: in both of the cases of the outer conditional we call `readSecondaryIdentifier` immediately, so I think we should lift that up into the main body of the func.

we should probably also return out of one condition and outdent the other, stylistically
</blockquote></details>

done

annrpom:
i just rebased to capture recent changes; TFTR! ('-')7

PR #5361, opened 2025-09-24
db: fix TestCleanupManagerCloseWithPacing

RaduBerinde:
Follow-up to #5348.

This test is ineffective because the `FreeSpaceThresholdBytes` option disables pacing. Fix by setting that option to 1 byte.

We also remove the extra checking of `stopCh` - if it is closed, we will exit the pacing loop right away so there's no need for this special check.


xinhaoz (approved):

RaduBerinde:
TFTR!

PR #5362, opened 2025-09-24
pebble-release-1.1: fix cleanupManager.Close() hanging when deletion pacing is active

RaduBerinde:
When `TargetByteDeletionRate` is set to a low value, `DB.Close()` can hang for a
long time because deletion pacing uses `time.Sleep()` that can't be interrupted.

This change modifies `maybePace()` to check for early termination and use
interruptible `select` instead of `time.Sleep()`.

Backport of #5348 and #5361.


RaduBerinde:
CC @ijsong

PR #5363, opened 2025-09-24
db: improve estimation of size of new cold blob files in shouldWriteB…

sumeerbhola:
…lobFiles

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -248,8 +248,7 @@ type tableCompaction struct {
 	// b) rewrite blob files: The compaction will write eligible values to new
 	// blob files. This consumes more write bandwidth because all values are
 	// rewritten. However it restores locality.
-	getValueSeparation func(
-		JobID, *tableCompaction, sstable.TableFormat, tieredmeta.ColdTierThresholdRetriever) compact.ValueSeparation
+	getValueSeparation getValueSeparation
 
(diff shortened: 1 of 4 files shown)


sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5363#-:-Oa0tfsu065tn4RXU30V:b-814g8k)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @RaduBerinde)

___
*[`compaction.go` line 3827 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5363#-O_x7Wo48PQ4U2pE6Hsp:-Oa0tT3j0_sdsBKEnKbG:b-d5iw2v) ([raw file](https://github.com/cockroachdb/pebble/blob/2b07dbc69cc2304a513cbc3d7ef7ddb28216172c/compaction.go#L3827)):*

[nit] Maybe rename `numShards` to `minOuptutFiles`? Either way, this could use a comment, maybe `numShards is the number of regions within the compaction range that have the same SpanID. These regions will go to different output files.`
</blockquote></details>

Done. I didn't use the "region" term, since I was worried it would cause more terminological confusion, but referenced the explanation of shards.

PR #5368, opened 2025-09-25
sstable: track value separation policy props used

xinhaoz:
Add ValueSeparationKind and ValueSeparationMinSize to sstable properties to
track the value separation policy used when writing the table.

### DIFF
--- a/internal/compact/run.go
+++ b/internal/compact/run.go
@@ -124,6 +124,11 @@ type ValueSeparation interface {
 	// sstable. It can be used to configure value separation specifically for
 	// the next compaction output.
 	SetNextOutputConfig(config ValueSeparationOutputConfig)
+	// Kind returns the kind of value separation strategy being used.
+	Kind() sstable.ValueSeparationKind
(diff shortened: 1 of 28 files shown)


jbowens (approved):
:lgtm:

xinhaoz:
*[`value_separation.go` line 413 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5368#-Oa6I_ni7RdmekCQMVX6:-OaL54xB08KfmJv36dg7:b-zco0h7) ([raw file](https://github.com/cockroachdb/pebble/blob/86ad0b693038a3454dfc594c48fe8a09235913b2/value_separation.go#L413)):*

when we're preserving blob references, I think we need to output whatever the input table's props were. So if all the inputs were (`sstable.ValueSeparationDefault`,`minimumSize=10`), we'd output the same. maybe that's for the next commit?
</blockquote></details>

Ack - this will be changed in the next PR when we add this field to the tablemetadata.

PR #5369, opened 2025-09-25
internal: add concurrentset

RaduBerinde:
Add a set data structure that supports adding/removing elements
with high concurrency.

This data structure will be used to keep track of iterators, so we can
periodically check for long-lived iterators. In that case the value
type will contain a timestamp and a few PCs from the creation caller
stack.

See bench_test.go for benchmark results.

### DIFF
--- a/internal/concurrentset/bench_test.go
+++ b/internal/concurrentset/bench_test.go
@@ -1,0 +1,98 @@
+// Copyright 2025 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package concurrentset
+
+import (
+	"fmt"
+	"math/rand/v2"
+	"runtime"
+	"sync"
+	"testing"
+)
+
+// Sample benchmark results.
+//
+// On Apple M1 (10 core):
+//   Set/1-10            59.3ns ± 4%
+//   Set/4-10            67.1ns ± 2%
+//   Set/10-10           66.9ns ± 1%
+//   Set/40-10           67.3ns ± 1%
+//   MapWithMutex/1-10   91.8ns ± 2%
+//   MapWithMutex/4-10    194ns ± 1%
+//   MapWithMutex/10-10   294ns ± 1%
+//   MapWithMutex/40-10   278ns ± 2%
+//
+// On Intel(R) Xeon(R) CPU @ 2.80GHz (GCP n2-custom-24-32768):
+//   Set/1-24            132ns ± 3%
+//   Set/4-24            154ns ± 3%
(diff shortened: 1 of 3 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5369)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jbowens and @sumeerbhola)

___
*[`internal/concurrentset/bench_test.go` line 34 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5369#-Oa1GvSf6ACotB2sGzay:-Oa1GvSf6ACotB2sGzaz:b4acbva) ([raw file](https://github.com/cockroachdb/pebble/blob/4d86e679936c3399397d963abf251608209553a4/internal/concurrentset/bench_test.go#L34)):*
> ```Go
> //   MapWithMutex/1-24   168ns ± 6%
> //   MapWithMutex/4-24   244ns ± 3%
> //   MapWithMutex/24-24  331ns ± 1%
> ```

Pretty surprised by how well both versions (but especially the simple one) scale on 24 cores..
Maybe there's something wrong with the benchmark?

It's surprising that b.N is about the same where I would expect it to grow with the higher parallelism (given that each worker does a fraction of b.N):
```
BenchmarkSet/1-10               24296702                60.42 ns/op
BenchmarkSet/4-10               23555757                66.02 ns/op
BenchmarkSet/10-10              23158496                67.32 ns/op
BenchmarkSet/40-10              23366930                66.29 ns/op
```

RaduBerinde:
*[`internal/concurrentset/bench_test.go` line 34 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5369#-Oa1GvSf6ACotB2sGzay:-Oa1IdSX2Am9LT8wXs6i:b-9zccz7) ([raw file](https://github.com/cockroachdb/pebble/blob/4d86e679936c3399397d963abf251608209553a4/internal/concurrentset/bench_test.go#L34)):*

Pretty surprised by how well both versions (but especially the simple one) scale on 24 cores..
Maybe there's something wrong with the benchmark?

It's surprising that b.N is about the same where I would expect it to grow with the higher parallelism (given that each worker does a fraction of b.N):
```
BenchmarkSet/1-10               24296702                60.42 ns/op
BenchmarkSet/4-10               23555757                66.02 ns/op
BenchmarkSet/10-10              23158496                67.32 ns/op
BenchmarkSet/40-10              23366930                66.29 ns/op
```
</blockquote></details>

Ah.. we are measuring the overall time instead of adding the time for each worker. So it's not `ns` of wall time per op, not CPU ns / op. I'll rework.

PR #5378, opened 2025-10-01
db: improve metrics and logs with multiple disks being used by a DB

### LINKED ISSUE #5328: db: improve metrics and logs with multiple disks being used by a DB
- `Metrics.LogWriter.FsyncLatency` does not distinguish between the primary WAL dir and secondary WAL dir. We should have separate metrics for each. Also, we should include all filesystem operations done for writing the log file in the corresponding latency histogram (i.e., create, write, sync etc.), since we've seen cases where deployments that use a noop fsync show low latency even though other operations may have higher latency.
- `DiskSlowInfo.SafeFormat` should include the immediate directory. Currently it does `filepath.Base(i.Path)`, so one cannot tell whether the WAL file is on the primary or secondary.

Jira issue: PEBBLE-1183

xxmplus:
Log full path to WAL file in disk slow info so it's explicit if the WAL file is on the primary or secondary.

Enhance the log writer to instrument all filesystem operations done for writing the log file in the latency histogram (create, write, sync, etc.).

Fixes #5328

### DIFF
--- a/commit_test.go
+++ b/commit_test.go
@@ -20,7 +20,6 @@ import (
 	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/record"
 	"github.com/cockroachdb/pebble/vfs"
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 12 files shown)


xxmplus:
>> I don't think we need per-operation histograms, but rather one histogram for the primary and one histogram for the secondary, reporting every operation against the individual file.
>
>These ops have quite different latency and frequency access patterns. If we consolidate them into one, we will lose visibility to the less frequent ops. Is that still ok?

I had a chance to sync up with Jackson. While they have different latency profiles, we care more about how file ops impact as a whole. I will update the code to consolidate them and also, not increase the number of metrics.

RaduBerinde:
> Cost: $0.0210 | Duration: 8.7s

Claude just gave new meaning to "my two cents"

xxmplus:
I believe the AI review bot found a real bug. The PR has been updated accordingly.

xxmplus:
TFTR!

PR #5380, opened 2025-10-01
db: track and report long-standing iterators

RaduBerinde:
I will try to figure out how to test this with CRDB. I also plan to run sysbench to see if there is any observable perf hit.

#### github: run tests against custom cockroach Go

Add a script and a nightly target that runs tests with the
custom Go runtime and the `cockroach_go` tag. The script is lifted
from crlib.

#### inflight: add data structure to detect long-running operations

This will be used to report long-lived iterators.

#### db: track and report long-standing iterators

We wire up iterators to an `inflight.Tracker`. Every 5 minutes we log
a report on iterators that have been open for more than 1 minute.

#### db: make iterator tracking configurable

We disable iterator tracking by default (since the overhead is
measurable). In addition, we allow disabling tracking for specific
iterators via `IterOptions`.

In CRDB we will enable iterator tracking but exempt those in the
hottest batch eval paths.

### DIFF
--- a/.github/workflows/nightlies.yaml
+++ b/.github/workflows/nightlies.yaml
@@ -49,3 +49,34 @@ jobs:
       sha: ${{ github.sha }}
       file_issue_branch: 'master'
       go_version: ${{ matrix.go }}
+
+  linux-cockroach-go:
+    runs-on: ubuntu-latest
+    env:
+      GO_BRANCH: cockroach-go1.23.12
+
+    steps:
+      - uses: actions/checkout@v4
+
+      # Step 1: Fetch the branch tip SHA for cache key
+      - name: Get cockroachdb/go commit hash
+        id: go-sha
+        run: |
+          SHA=$(git ls-remote https://github.com/cockroachdb/go.git refs/heads/$GO_BRANCH | cut -f1)
+          echo "GO_SHA=$SHA" >> $GITHUB_ENV
+
+      # Step 2: Restore cache (per branch + commit SHA)
+      - name: Cache custom Go toolchain
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/cockroachdb-go/${{ env.GO_SHA }}
+          key: cockroachdb-${{ env.GO_SHA }}
+
+      # Step 3: Install bootstrap Go (needed to build fork)
+      - name: Install bootstrap Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.23.x"
+
+      # Step 4: Run tests with custom Go
+      - run: ./scripts/run-tests-with-custom-go.sh -tags invariants ./...
--- a/db.go
+++ b/db.go
@@ -21,6 +21,7 @@ import (
 	"github.com/cockroachdb/pebble/internal/arenaskl"
 	"github.com/cockroachdb/pebble/internal/base"
 	"github.com/cockroachdb/pebble/internal/cache"
+	"github.com/cockroachdb/pebble/internal/inflight"
 	"github.com/cockroachdb/pebble/internal/invalidating"
 	"github.com/cockroachdb/pebble/internal/invariants"
 	"github.com/cockroachdb/pebble/internal/keyspan"
@@ -545,6 +546,8 @@ type DB struct {
 	// the time at database Open; may be used to compute metrics like effective
 	// compaction concurrency
 	openedAt time.Time
+
+	iterTracker *inflight.Tracker
 }
 
 var _ Reader = (*DB)(nil)
@@ -1111,6 +1114,10 @@ func (d *DB) newIter(
 		dbi.batch.batch = batch
 		dbi.batch.batchSeqNum = batch.nextSeqNum()
 	}
+	if !dbi.batchOnlyIter {
+		dbi.tracker = d.iterTracker
+		dbi.trackerHandle = d.iterTracker.Start()
+	}
 	return finishInitializingIter(ctx, buf)
 }
 
@@ -1121,6 +1128,7 @@ func (d *DB) newIter(
 func finishInitializingIter(ctx context.Context, buf *iterAlloc) *Iterator {
 	// Short-hand.
 	dbi := &buf.dbi
+
 	var memtables flushableList
 	if dbi.readState != nil {
 		memtables = dbi.readState.memtables
@@ -1615,6 +1623,9 @@ func (d *DB) Close() error {
 	if v := d.mu.snapshots.count(); v > 0 {
 		err = firstError(err, errors.Errorf("leaked snapshots: %d open snapshots on DB %p", v, d))
 	}
+
+	d.iterTracker.Close()
+	d.iterTracker = nil
 
 	return err
 }
--- a/db.go
+++ b/db.go
@@ -1114,8 +1114,8 @@ func (d *DB) newIter(
 		dbi.batch.batch = batch
 		dbi.batch.batchSeqNum = batch.nextSeqNum()
 	}
-	if !dbi.batchOnlyIter {
-		dbi.tracker = d.iterTracker
+	dbi.tracker = d.iterTracker
+	if !dbi.batchOnlyIter && d.iterTracker != nil && !dbi.opts.ExemptFromTracking {
 		dbi.trackerHandle = d.iterTracker.Start()
 	}
 	return finishInitializingIter(ctx, buf)
@@ -1624,8 +1624,10 @@ func (d *DB) Close() error {
 		err = firstError(err, errors.Errorf("leaked snapshots: %d open snapshots on DB %p", v, d))
 	}
 
-	d.iterTracker.Close()
-	d.iterTracker = nil
+	if d.iterTracker != nil {
+		d.iterTracker.Close()
+		d.iterTracker = nil
+	}
 
 	return err
 }
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ require (
 	github.com/HdrHistogram/hdrhistogram-go v1.1.2
 	github.com/RaduBerinde/axisds v0.0.0-20250419182453-5135a0650657
 	github.com/cespare/xxhash/v2 v2.2.0
-	github.com/cockroachdb/crlib v0.0.0-20250916151006-1094cb39adac
+	github.com/cockroachdb/crlib v0.0.0-20251001180057-2a49e1873587
 	github.com/cockroachdb/datadriven v1.0.3-0.20250911232732-d959cf14706c
 	github.com/cockroachdb/errors v1.11.3
 	github.com/cockroachdb/metamorphic v0.0.0-20231108215700-4ba948b56895
@@ -24,6 +24,7 @@ require (
 	github.com/pmezard/go-difflib v1.0.0
 	github.com/prometheus/client_golang v1.16.0
 	github.com/prometheus/client_model v0.3.0
+	github.com/puzpuzpuz/xsync/v3 v3.5.1
 	github.com/spf13/cobra v1.0.0
 	github.com/stretchr/testify v1.9.0
 	golang.org/x/exp v0.0.0-20230626212559-97b1e661b5df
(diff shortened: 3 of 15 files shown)


sumeerbhola (changes requested):
Nice. Only looked at the first commit -- flushing comments.

> @sumeerbhola reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5380#-:-OaWx9W0Bunbl2xeZz0q:bb55f4p)* status: 2 of 10 files reviewed, 3 unresolved discussions (waiting on @jbowens)

___
*[`internal/inflight/in_flight.go` line 29 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5380#-OaWrZ8M5fQj-tat8SKV:-OaWrZ8M5fQj-tat8SKW:b-wcuqg7) ([raw file](https://github.com/cockroachdb/pebble/blob/97d001e37405c0a1948e0a8053cd11b08d0eb9a2/internal/inflight/in_flight.go#L29)):*
> ```Go
> // Usage:
> //
> //	t := NewTrackerWithAutoReport(10*time.Minute, reportFn)
> ```

was this supposed to say `NewPollingTracker`?
___
*[`internal/inflight/in_flight.go` line 129 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5380#-OaWtRBC0OzQHebwDYyd:-OaWtRBC0OzQHebwDYye:b13fjw7) ([raw file](https://github.com/cockroachdb/pebble/blob/97d001e37405c0a1948e0a8053cd11b08d0eb9a2/internal/inflight/in_flight.go#L129)):*
> ```Go
> 		<-t.reportCh
> 		t.timer.Stop()
> 		t.reportCh = nil
> ```

The timer func can do a select on `t.reportCh` concurrent with this setting to nil. Isn't that a race?

It took me a moment to reason through the other concurrency aspects e.g. The timer.Reset in the other goroutine strictly precedes the timer.Stop here and subsequently the other goroutine will not fiddle with timer. A code comment making all of this clear would be great.
___
*[`internal/inflight/in_flight.go` line 153 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5380#-OaWvj7K6g_oTfPD3M92:-OaWvj7K6g_oTfPD3M93:bdnbs0s) ([raw file](https://github.com/cockroachdb/pebble/blob/97d001e37405c0a1948e0a8053cd11b08d0eb9a2/internal/inflight/in_flight.go#L153)):*
> ```Go
> // already stopped.
> func (t *Tracker) Stop(h Handle) {
> 	s := &t.shards[h>>56]
> ```

The `h>>56` could use a short comment. Something like
// The high byte is the shard number. See newTracker.

RaduBerinde (commented):
Thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5380#-:-OaXA_5N13QBXL1VaB_W:b-ucfknh)* status: 1 of 10 files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/inflight/in_flight.go` line 29 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5380#-OaWrZ8M5fQj-tat8SKV:-OaX5kJD4t0z99x5Gfkv:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/97d001e37405c0a1948e0a8053cd11b08d0eb9a2/internal/inflight/in_flight.go#L29)):*

was this supposed to say `NewPollingTracker`?
</blockquote></details>

Done.
___
*[`internal/inflight/in_flight.go` line 129 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5380#-OaWtRBC0OzQHebwDYyd:-OaX5sjN6H0-eDRNTOkL:b6qzven) ([raw file](https://github.com/cockroachdb/pebble/blob/97d001e37405c0a1948e0a8053cd11b08d0eb9a2/internal/inflight/in_flight.go#L129)):*

The timer func can do a select on `t.reportCh` concurrent with this setting to nil. Isn't that a race?

It took me a moment to reason through the other concurrency aspects e.g. The timer.Reset in the other goroutine strictly precedes the timer.Stop here and subsequently the other goroutine will not fiddle with timer. A code comment making all of this clear would be great.
</blockquote></details>

Nice catch. Yeah, it turned out a bit more subtle than I wanted. I switched to using a mutex.
___
*[`internal/inflight/in_flight.go` line 153 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5380#-OaWvj7K6g_oTfPD3M92:-OaX78ma3JAKbAQ7P2G4:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/97d001e37405c0a1948e0a8053cd11b08d0eb9a2/internal/inflight/in_flight.go#L153)):*

The `h>>56` could use a short comment. Something like
// The high byte is the shard number. See newTracker.
</blockquote></details>

Done.

RaduBerinde (commented):
Improved the memory layout of the structure. I will add a benchmark.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5380#-:-OaXEVVr5Xfu-CyyPkYR:b-obbev0)* status: 1 of 10 files reviewed, 3 unresolved discussions (waiting on @jbowens and @sumeerbhola)

sumeerbhola (approved):
:lgtm: 

> @sumeerbhola reviewed 2 of 4 files at r1, 7 of 8 files at r4, 1 of 1 files at r5, 4 of 4 files at r6, 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5380#-:-Oa_Imn77WuRgm6529xh:bg7igkv)* status: all files reviewed (commit messages unreviewed), 8 unresolved discussions (waiting on @jbowens)

___
*[`internal/inflight/in_flight_bench_test.go` line 22 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5380#-Oa_FRnY8RIvyF8_azvR:-Oa_FRnY8RIvyF8_azvS:b-caehan) ([raw file](https://github.com/cockroachdb/pebble/blob/b8583728f134929e007fcb9022690ed8dae57058/internal/inflight/in_flight_bench_test.go#L22)):*
> ```Go
> //	Tracker/p=1-10           230ns ± 0%       216ns ± 1%  -5.78%  (p=0.029 n=4+4)
> //	Tracker/p=5-10           318ns ± 0%       291ns ± 1%  -8.46%  (p=0.029 n=4+4)
> //	Tracker/p=10-10          543ns ± 3%       540ns ± 3%    ~     (p=0.886 n=4+4)
> ```

I was assuming that the cpu affinity would start to matter when the number of goroutines exceed the number of cpus. If correct, should this be run with say p=100?
___
*[`internal/inflight/in_flight_bench_test.go` line 27 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5380#-Oa_GAsr6tlQ1eYHUrVo:-Oa_GAsr6tlQ1eYHUrVp:b-jwqmfm) ([raw file](https://github.com/cockroachdb/pebble/blob/b8583728f134929e007fcb9022690ed8dae57058/internal/inflight/in_flight_bench_test.go#L27)):*
> ```Go
> 	for _, parallelism := range []int{1, procs / 2, procs} {
> 		b.Run(fmt.Sprintf("p=%d", parallelism), func(b *testing.B) {
> 			const batchSize = 1000
> ```

Is `batchSize` to amortize the cost of the channel operation, so it becomes insignificant compared to the Tracker cost?
___
*[`internal/inflight/in_flight_bench_test.go` line 34 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5380#-Oa_FHuw7R5EcafKuhyZ:-Oa_FHuw7R5EcafKuhy_:bepz1i7) ([raw file](https://github.com/cockroachdb/pebble/blob/b8583728f134929e007fcb9022690ed8dae57058/internal/inflight/in_flight_bench_test.go#L34)):*
> ```Go
> 			for range parallelism {
> 				wg.Add(1)
> 				tr := NewTracker()
> ```

Why is each goroutine getting its own Tracker?
___
*[`internal/inflight/in_flight.go` line 30 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5380#-Oa_DKzqBVtz9g_EaH7n:-Oa_DKzqBVtz9g_EaH7o:bgap1ce) ([raw file](https://github.com/cockroachdb/pebble/blob/b8583728f134929e007fcb9022690ed8dae57058/internal/inflight/in_flight.go#L30)):*
> ```Go
> // Usage:
> //
> //	t := NewPollingTracker(10*time.Minute, reportFn)
> ```

nit: missing a parameter.

Consider combining this with the example below, because it seems a bit repetitive.
___
*[`open.go` line 209 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5380#-Oa_IHIzAhRTQM5qEjME:-Oa_IHIzAhRTQM5qEjMF:bo0e97c) ([raw file](https://github.com/cockroachdb/pebble/blob/e80afdb1bf48d88383b852471695acf61642e924/open.go#L209)):*
> ```Go
> 	}
> 	const iterTrackerPollInterval = 5 * time.Minute
> 	const iterTrackerMaxAge = 1 * time.Minute
> ```

After validating the performance etc., how about making these dynamic Options. I suspect we will want to tune these to reduce noise in CockroachDB.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5380)* status: all files reviewed (commit messages unreviewed), 8 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/inflight/in_flight_bench_test.go` line 22 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5380#-Oa_FRnY8RIvyF8_azvR:-Oa_JcER5U_xoloUDNGE:b-gb6fgq) ([raw file](https://github.com/cockroachdb/pebble/blob/b8583728f134929e007fcb9022690ed8dae57058/internal/inflight/in_flight_bench_test.go#L22)):*

I was assuming that the cpu affinity would start to matter when the number of goroutines exceed the number of cpus. If correct, should this be run with say p=100?
</blockquote></details>

CPU affinity makes it so that the 10 goroutines don't contend with each other on the data structure (for the most part). Added a 2*P case. Note that both versions shown here have CPU affinity, the crdb one is just more efficient.

This is what it would look if we used a random shard:
```
Tracker/p=1-10    215ns ± 0%   225ns ± 0%   +4.86%  (p=0.008 n=5+5)
Tracker/p=5-10    332ns ± 1%   537ns ± 4%  +61.72%  (p=0.008 n=5+5)
Tracker/p=10-10   527ns ± 4%   826ns ± 1%  +56.85%  (p=0.008 n=5+5)
Tracker/p=20-10  1.07µs ± 1%  1.63µs ± 1%  +52.12%  (p=0.008 n=5+5)
```
___
*[`internal/inflight/in_flight_bench_test.go` line 27 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5380#-Oa_GAsr6tlQ1eYHUrVo:-Oa_MkQL9sor27hYHSjM:b-63v6yu) ([raw file](https://github.com/cockroachdb/pebble/blob/b8583728f134929e007fcb9022690ed8dae57058/internal/inflight/in_flight_bench_test.go#L27)):*

Is `batchSize` to amortize the cost of the channel operation, so it becomes insignificant compared to the Tracker cost?
</blockquote></details>

Yeah, added a comment.
___
*[`internal/inflight/in_flight_bench_test.go` line 34 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5380#-Oa_FHuw7R5EcafKuhyZ:-Oa_JHx54CIVwbprzMEV:b-575q54) ([raw file](https://github.com/cockroachdb/pebble/blob/b8583728f134929e007fcb9022690ed8dae57058/internal/inflight/in_flight_bench_test.go#L34)):*

Why is each goroutine getting its own Tracker?
</blockquote></details>

Oh, wow, nice catch :facepalm:

RaduBerinde:
Unfortunately the Sysbench benchmark shows a 2% regression:
https://github.com/cockroachdb/cockroach/actions/runs/18222532410?pr=154734

I don't know if we can make the data structure itself much more efficient. Maybe we need to look into "exempting" the hot paths from this tracking.

RaduBerinde:
I will try the idea of using a timer to only start tracking an iterator a few seconds after creation.

RaduBerinde:
A quick benchmark shows that a `timer.Reset(), runtime.Callers(), timer.Stop()` sequence takes about ~200ns vs ~500ns for this data structure. That would still mean a not-insignificant ~1% regression.

Just `runtime.Callers()` is ~100ns, which means we won't be able to get under that if we track all iterators.

RaduBerinde:
Updated. Tracking is now configurable and disabled by default. Individual iterators can be exempt, which will be used in the CRDB hottest paths (batch eval).

PR #5384, opened 2025-10-02
metamorphic: fix bug in suffix generation when prefix == startPrefix

annrpom:
While the suffixIdx < startSuffixIdx, generate a new suffixIdx - not the other way around.

Fixes: #5349


annrpom:
nope oops this isn't it

PR #5385, opened 2025-10-02
metamorphic: fix bug in suffix generation when prefix == startPrefix

annrpom:
We should also keep generating a new suffix if our suffixIdx is == 0 - since a zero suffix sorts before all non-zero suffixes.

Fixes: #5349


xinhaoz (approved):

annrpom:
TFTR! ('-')7

PR #5387, opened 2025-10-02
metrics: fix TestMetrics flake

RaduBerinde:
The disk usage counts the OPTIONS file as well. One of the values
`table_cache_shards` varies depending on machine and can lead to a 1
byte difference.

We fix this option in this test and show exact bytes values for disk
usage so this is easier to catch (it was caught with a separate change
where that 1 byte happened to make the difference between `4.5KB` and
`4.6KB`).

### DIFF
--- a/metrics_test.go
+++ b/metrics_test.go
@@ -16,6 +16,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/cockroachdb/crlib/crhumanize"
 	"github.com/cockroachdb/crlib/crstrings"
 	"github.com/cockroachdb/datadriven"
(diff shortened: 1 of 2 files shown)


jbowens (approved):
:lgtm:

> @jbowens reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5387#-:-Oaasy7yDZd3GdUDD05P:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sumeerbhola)

RaduBerinde:
TFTR!

PR #5388, opened 2025-10-02
wal: clean up all segment files

### LINKED ISSUE #5355: wal: failover writer may leak a file on the filesystem
https://github.com/cockroachdb/pebble/blob/489b133bc3a12164b261fd6d2447167cacb2a276/wal/failover_writer.go#L687-L694

https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1758549696285599

Jira issue: PEBBLE-1196

jbowens:
When WAL failover is configured, a single logical WAL may be composed of multiple physical segment files. The creation of these physical segment files occurs asynchronously. This asynchronous creation may race with the closing of the failover writer.

Specifically, an outstanding attempt to create a segment file may block. Meanwhile, writes persisting all the necessary records may complete on the other device. If the logical WAL is now finished, the failover writer may be closed by higher-level code while the outstanding attempt to create a file remains. The system progresses, accumulating a list of obsolete files based on the set that existed at the time that the writer was closed. Eventually, the outstanding file creation may complete, creating a new file.

Previously this race resulted in leaking the straggling file. Because we preallocate disk space for WAL files, this logically empty file could still consume significant disk space. This leaked file could not be discovered and deleted until process restart.

This commit adjusts the FailoverWriter to invoke a callback on the FailoverManager, propagating information about these abandoned segment files. This allows the FailoverManager to propagate these obsolete files to higher levels for deletion.

Fix #5355.

### DIFF
--- a/wal/failover_manager.go
+++ b/wal/failover_manager.go
@@ -5,6 +5,7 @@
 package wal
 
 import (
+	"cmp"
 	"fmt"
 	"io"
 	"math/rand/v2"
@@ -537,26 +538,22 @@ func (wm *failoverManager) init(o Options, initial Logs) error {
 func (wm *failoverManager) List() Logs {
 	wm.mu.Lock()
 	defer wm.mu.Unlock()
-	n := len(wm.mu.closedWALs)
-	if wm.mu.ww != nil {
-		n++
-	}
-	wals := make(Logs, n)
-	setLogicalLog := func(index int, llse logicalLogWithSizesEtc) {
+	wals := make(Logs, 0, len(wm.mu.closedWALs)+1)
+	setLogicalLog := func(llse logicalLogWithSizesEtc) {
 		segments := make([]segment, len(llse.segments))
 		for j := range llse.segments {
 			segments[j] = llse.segments[j].segment
 		}
-		wals[index] = LogicalLog{
+		wals = append(wals, LogicalLog{
 			Num:      llse.num,
 			segments: segments,
-		}
+		})
 	}
-	for i, llse := range wm.mu.closedWALs {
-		setLogicalLog(i, llse)
+	for _, llse := range wm.mu.closedWALs {
+		setLogicalLog(llse)
 	}
 	if wm.mu.ww != nil {
-		setLogicalLog(n-1, wm.mu.ww.getLog())
+		setLogicalLog(wm.mu.ww.getLog())
 	}
 	return wals
 }
@@ -637,6 +634,7 @@ func (wm *failoverManager) Create(wn NumWAL, jobID int) (Writer, error) {
 		stopper:                     wm.stopper,
 		failoverWriteAndSyncLatency: wm.opts.FailoverWriteAndSyncLatency,
 		writerClosed:                wm.writerClosed,
+		segmentClosed:               wm.segmentClosed,
 		writerCreatedForTest:        wm.opts.logWriterCreatedForTesting,
 		writeWALSyncOffsets:         wm.opts.WriteWALSyncOffsets,
 	}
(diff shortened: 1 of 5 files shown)


jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5388)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @xxmplus)

___
*[`wal/failover_manager.go` line 676 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5388#-Oab-iTf5Mu6UTZ-zrw7:-OafEp0M51E-UITNGPIS:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/33c42311fe5843310599d5b5bb22cea88dabad6f/wal/failover_manager.go#L676)):*

[nit] "segmentClosed is called by the failover writer; see failoverWriterOpts.segmentClosed."
</blockquote></details>

Done
___
*[`wal/failover_manager.go` line 683 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5388#-Oab0UFFDY0eiNt2M7vL:-OafEpjp5NJiZgLNjBPr:b-qoscuk) ([raw file](https://github.com/cockroachdb/pebble/blob/33c42311fe5843310599d5b5bb22cea88dabad6f/wal/failover_manager.go#L683)):*

[nit] Wouldn't `closedWALs` be better as a map?
</blockquote></details>

Hrm we're using its ordered nature elsewhere when trimming the prefix of logs that are obsolete. I updated this to use `slices.BinarySearchFunc`
___
*[`wal/failover_manager.go` line 687 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5388#-Oab02RfCCKBGdDTZhKj:-OafEzM01TKlPIEWvWne:b-46tfrt) ([raw file](https://github.com/cockroachdb/pebble/blob/33c42311fe5843310599d5b5bb22cea88dabad6f/wal/failover_manager.go#L687)):*

!alreadyExists?
</blockquote></details>

Updated to use `BinarySearchFunc`
___
*[`wal/failover_manager.go` line 702 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5388#-Oab11xGCXfBVFCS-Jia:-OafF0bHETO8OzWewhJQ:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/33c42311fe5843310599d5b5bb22cea88dabad6f/wal/failover_manager.go#L702)):*

[nit] BinarySearchFunc would return an index that you can use directly with Insert
</blockquote></details>

Done
___
*[`wal/failover_writer.go` line 704 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5388#-Oab-NKACyFHUPwmXp6k:-OafF1uN870gKlAtKnf_:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/33c42311fe5843310599d5b5bb22cea88dabad6f/wal/failover_writer.go#L704)):*

[nit] "Even though we never wrote to the file, it could be occupying non-negligible disk space because of preallocation."
</blockquote></details>

Done

jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5388#-:-OafWK2RCPoVTzJFLIda:b-814g8k)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @xxmplus)

___
*[`wal/failover_manager.go` line 687 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5388#-Oab02RfCCKBGdDTZhKj:-OafVOpy84G7LBaabOCP:bcf8mch) ([raw file](https://github.com/cockroachdb/pebble/blob/33c42311fe5843310599d5b5bb22cea88dabad6f/wal/failover_manager.go#L687)):*

Is it concerning that the version with the incorrect `alreadyExists` check didn't fail any tests?
</blockquote></details>

The previous code was bugged in more ways than one, and `alreadyExists` was always true (the `ContainsFunc` func was comparing the wrong thing and always returned true). This meant we always added the new segment. In this case two wrongs made a right, and if you fixed the contains func the test I added failed.

I'm not particularly concerned about the case where there is an existing segment; I think that's not actually a possible case in practice.

RaduBerinde:
*[`wal/failover_manager.go` line 687 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5388#-Oab02RfCCKBGdDTZhKj:-OafaM8wEAP8XawJRR5N:b-mulzg) ([raw file](https://github.com/cockroachdb/pebble/blob/33c42311fe5843310599d5b5bb22cea88dabad6f/wal/failover_manager.go#L687)):*

The previous code was bugged in more ways than one, and `alreadyExists` was always true (the `ContainsFunc` func was comparing the wrong thing and always returned true). This meant we always added the new segment. In this case two wrongs made a right, and if you fixed the contains func the test I added failed.

I'm not particularly concerned about the case where there is an existing segment; I think that's not actually a possible case in practice.
</blockquote></details>

:beers:

RaduBerinde:
Many nigtlies failed with TestManagerFailover flakes

PR #5392, opened 2025-10-03
objstorageprovider: propagate errIsNotExist fn in remoteReadable

annrpom:
Fixes: https://github.com/cockroachdb/cockroach/issues/154032


RaduBerinde (approved):
Thanks!

Issue #5390, opened 2025-10-03
db: bug in EventuallyFileOnlySnapshot and excise interaction


sumeerbhola:
Observed in https://github.com/cockroachdb/cockroach/issues/154566#issuecomment-3358551933 and https://github.com/cockroachdb/cockroach/issues/154596#issuecomment-3358963192.

```
* ERROR: a panic has occurred!
* panic: unexpected excise of an EventuallyFileOnlySnapshot's bounds
* (1) attached stack trace
*   -- stack trace:
*   | runtime.gopanic
*   | 	GOROOT/src/runtime/panic.go:791
*   | github.com/cockroachdb/cockroach/pkg/kv/kvserver.(*Replica).applySnapshotRaftMuLocked.func1
*   | 	pkg/kv/kvserver/replica_raftstorage.go:500
*   | runtime.gopanic
*   | 	GOROOT/src/runtime/panic.go:791
*   | github.com/cockroachdb/pebble.(*DB).ingestApply
*   | 	external/com_github_cockroachdb_pebble/ingest.go:2222
*   | github.com/cockroachdb/pebble.(*DB).ingest.func2
*   | 	external/com_github_cockroachdb_pebble/ingest.go:1720
*   | github.com/cockroachdb/pebble.(*commitPipeline).AllocateSeqNum
*   | 	external/com_github_cockroachdb_pebble/commit.go:421
*   | github.com/cockroachdb/pebble.(*DB).ingest
*   | 	external/com_github_cockroachdb_pebble/ingest.go:1733
*   | github.com/cockroachdb/pebble.(*DB).IngestAndExcise
*   | 	external/com_github_cockroachdb_pebble/ingest.go:1297
*   | github.com/cockroachdb/cockroach/pkg/storage.(*Pebble).IngestAndExciseFiles
*   | 	pkg/storage/pebble.go:2186
```

Jira issue: PEBBLE-1204

sumeerbhola:
Analysis in https://docs.google.com/document/d/1nA8UV4n5AjlFua7N3x8fZx6wUECVSMMNiSEiOZpzWmo

sumeerbhola:
Not fixed
```
*
* ERROR: a panic has occurred!
* unexpected excise of an EventuallyFileOnlySnapshot's bounds: excise [/Table/5/0,0, /Table/6/0,0) 17512, efos [/Min/0,0, /Max/0,0) 17511
* (1) attached stack trace
*   -- stack trace:
[...]

PR #5393, opened 2025-10-03
db: test to demonstrate excise and EFOS race bug

sumeerbhola:
Informs #5390

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -1715,6 +1715,9 @@ func (d *DB) ingest(ctx context.Context, args ingestArgs) (IngestOperationStats,
 			waitFlushDuration = waitFlushStart.Elapsed()
 		}
 
+		if d.opts.private.testingBeforeIngestApplyFunc != nil {
+			d.opts.private.testingBeforeIngestApplyFunc()
+		}
 		// Assign the sstables to the correct level in the LSM and apply the
 		// version edit.
 		ve, manifestUpdateDuration, err = d.ingestApply(ctx, jobID, loadResult, mut, args.ExciseSpan, args.ExciseBoundsPolicy, seqNum)
--- a/options.go
+++ b/options.go
@@ -1187,6 +1187,10 @@ type Options struct {
 		// obsolete file deletion (to make events deterministic).
 		testingAlwaysWaitForCleanup bool
 
+		// testingBeforeIngestApplyFunc when non-nil, is called when ingesting,
+		// before calling DB.ingestApply.
+		testingBeforeIngestApplyFunc func()
+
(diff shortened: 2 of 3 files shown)


jbowens (approved):
:lgtm:

> @jbowens reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5393#-:-OautyWZ5F6vZ9iabi2-:bnfp4nl)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`snapshot_test.go` line 436 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5393#-OautBrc3R2yLrlJ7dw8:-OautBrdCSdSABFz9-0t:bo20wp5) ([raw file](https://github.com/cockroachdb/pebble/blob/d073980927e4966239c43ed17ac0971900493ee9/snapshot_test.go#L436)):*
> ```Go
> 	// Flush and transition to FOS.
> 	require.NoError(t, d.Flush())
> 	require.NoError(t, snap.WaitForFileOnlySnapshot(context.Background(), 0))
> ```

I forgot about this `WaitForFileOnlySnapshot` method. Do we need this call? Since we synchronously call `d.Flush()` above, it necessarily must have transitioned to being a file-only snapshot, right? If we don't need it, maybe we should get rid of the `WaitForFileOnlySnapshot` method altogether. It's unused within Cockroach.

sumeerbhola (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5393#-:-Oav3Lbt2z_27xhvOPZV:b-814g8k)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`snapshot_test.go` line 436 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5393#-OautBrc3R2yLrlJ7dw8:-Oav2zttB62uJ5FquYvN:b-5s5vfh) ([raw file](https://github.com/cockroachdb/pebble/blob/d073980927e4966239c43ed17ac0971900493ee9/snapshot_test.go#L436)):*

I forgot about this `WaitForFileOnlySnapshot` method. Do we need this call? Since we synchronously call `d.Flush()` above, it necessarily must have transitioned to being a file-only snapshot, right? If we don't need it, maybe we should get rid of the `WaitForFileOnlySnapshot` method altogether. It's unused within Cockroach.
</blockquote></details>

I think it was for those users who wanted to only create an iterator on a FOS. I don't have any objection to removing it. I used it here mostly to ensure we were in a FOS mode. I could have used a `ConfirmThatFOS() bool` method.

PR #5394, opened 2025-10-03
db: EFOS and excise race bug fix

sumeerbhola:
Fixes https://github.com/cockroachdb/pebble/issues/5390

### DIFF
--- a/db.go
+++ b/db.go
@@ -501,6 +501,12 @@ type DB struct {
 			// sstables.
 			cumulativePinnedCount uint64
 			cumulativePinnedSize  uint64
+
+			// The ongoing excises. These are under snapshots since this is only
+			// needed to coordinate with EventuallyFileOnlySnapshot creation.
+			ongoingExcises map[SeqNum]KeyRange
+			// Signalled when an ongoingExcise is removed.
+			ongoingExcisesRemovedCond *sync.Cond
 		}
 
 		tableStats struct {
@@ -2811,4 +2817,15 @@ func (d *DB) DebugCurrentVersion() *manifest.Version {
 	d.mu.Lock()
 	defer d.mu.Unlock()
 	return d.mu.versions.currentVersion()
+}
+
+func (d *DB) removeFromOngoingExcises(seqNum base.SeqNum) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	_, ok := d.mu.snapshots.ongoingExcises[seqNum]
+	if !ok {
+		panic(errors.AssertionFailedf("pebble: no ongoing excise for seqnum %d", seqNum))
+	}
+	delete(d.mu.snapshots.ongoingExcises, seqNum)
+	d.mu.snapshots.ongoingExcisesRemovedCond.Broadcast()
 }
--- a/ingest.go
+++ b/ingest.go
@@ -1526,7 +1526,9 @@ func (d *DB) ingest(ctx context.Context, args ingestArgs) (IngestOperationStats,
 	// asFlushable indicates whether the sstable was ingested as a flushable.
 	var asFlushable bool
 	var waitFlushStart crtime.Mono
+	var assignedSeqNum SeqNum
 	prepare := func(seqNum base.SeqNum) {
+		assignedSeqNum = seqNum
 		// Note that d.commit.mu is held by commitPipeline when calling prepare.
 
 		// Determine the set of bounds we care about for the purpose of checking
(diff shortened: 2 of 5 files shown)


db.go
```diff
@@ -500,6 +500,12 @@ type DB struct {
 			// sstables.
 			cumulativePinnedCount uint64
 			cumulativePinnedSize  uint64
+
+			// The ongoing excises. These are under snapshots since this is only
+			// needed to coordinate with EventuallyFileOnlySnapshot creation.
+			ongoingExcises []ongoingExcise
```
RaduBerinde, line 506: Can"t this be a map from seqnum to span?

jbowens (commented):
I'm not sure I'm understanding the race fully, but would it be resolved if EFOS creation called `AllocateSeqNum` to allocate 1 sequence number for the EFOS, serializing the enqueueing of the new snapshot with the IngestAndExcise's `AllocateSeqNum` prepare call?

I had trouble keeping the interactions between excises, flushes and EFOSes in my head before, and we're fixing it with more complexity, so I'm wondering if we missed a simplification somewhere along our way to our current design.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5394#-:-OafTs7dEzZf5zEw1LY_:b9fzwwb)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @sumeerbhola and @xxmplus)

db.go
sumeerbhola: Done

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5394)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @xxmplus)

sumeerbhola:
TFTRs!

sumeerbhola (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5394)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde and @xxmplus)

___
*[`ingest.go` line 1575 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5394#-Oag2fUpBhYVtsUn6f3Y:-Oag2fUpBhYVtsUn6f3Z:be0udzb) ([raw file](https://github.com/cockroachdb/pebble/blob/76038ca5bc6cd8a9e6a7bc1296269d355fe2c6e7/ingest.go#L1575)):*
> ```Go
> 			// mutates the LSM in a non-additive way (unlike ingests and writes).
> 			//
> 			// Say there is an existing EFOS for the key-range [a, d) where a is in
> ```

Added this comment as promised in https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1759527039213429?thread_ts=1759363211.737159&cid=CAC6K3SLU

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5394)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @jbowens and @xxmplus)

___
*[`snapshot.go` line 241 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5394#-OavJy48ETRDXWuEc9td:-OavJy48ETRDXWuEc9te:b-8e79sl) ([raw file](https://github.com/cockroachdb/pebble/blob/330a4c57848b923da8a192fa9457cb8431110ef3/snapshot.go#L241)):*
> ```Go
> 	var es *EventuallyFileOnlySnapshot
> 	tryConstruct := func() {
> 		d.mu.Lock()
> ```

[nit] I don't see why we need to unlock and lock each time, this can stay outside of the function
___
*[`snapshot.go` line 272 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5394#-OavJCE5AoszI7ZNDmnS:-OavJCE5AoszI7ZNDmnT:baa3w6w) ([raw file](https://github.com/cockroachdb/pebble/blob/330a4c57848b923da8a192fa9457cb8431110ef3/snapshot.go#L272)):*
> ```Go
> 		snapshotSeqNum := d.mu.versions.visibleSeqNum.Load()
> 
> 		// Check if any of the keyRanges overlap with an ongoing
> ```

[nit] Half of this function runs only once, which makes things harder to follow. We could separate the code that returns the `exciseSeqNumToWaitFor` and refactor this like:

```
Lock()
defer Unlock()
for {
  exciseSeqNumToWaitFor := fooLocked()
  if exciseSeqNumToWaitFor == 0 {
    break
  }
  for !base.Visible(exciseSeqNumToWaitFor, d.mu.versions.visibleSeqNum.Load(), base.SeqNumMax) {
     d.mu.snapshots.ongoingExcisesRemovedCond.Wait()
   }
}
```
___
*[`snapshot.go` line 291 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5394#-OavLkwREqRRyY8o3ReP:-OavLkwREqRRyY8o3ReQ:b-wyogtm) ([raw file](https://github.com/cockroachdb/pebble/blob/330a4c57848b923da8a192fa9457cb8431110ef3/snapshot.go#L291)):*
> ```Go
> 				if keyRanges[i].OverlapsKeyRange(d.cmp, span) {
> 					exciseSeqNumToWaitFor = seqNum
> 					break
> ```

Instead of putting a loop around everything, can't we just do a loop right here?

```
for !base.Visible(exciseSeqNumToWaitFor, d.mu.versions.visibleSeqNum.Load(), base.SeqNumMax) {
   d.mu.snapshots.ongoingExcisesRemovedCond.Wait()
 }
```
___
*[`ingest.go` line 1557 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5394#-OavHzcN3M-k7_24kbOL:-OavHzcODYsfP9ksI3yz:b-7e9mcj) ([raw file](https://github.com/cockroachdb/pebble/blob/330a4c57848b923da8a192fa9457cb8431110ef3/ingest.go#L1557)):*
> ```Go
> 		d.mu.Lock()
> 		defer func() {
> 			if addToOngoingExcises {
> ```

[nit] Would it not be equivalent to do `if args.ExciseSpan.Valid() && !asFlushable` instead of another flag? And same for `if removeFromOngoingExcises` below

RaduBerinde:
*[`snapshot.go` line 291 at r7](https://reviewable.io/reviews/cockroachdb/pebble/5394#-OavLkwREqRRyY8o3ReP:-Ob5JD657jAhB5VWxXLk:b-75dkrr) ([raw file](https://github.com/cockroachdb/pebble/blob/330a4c57848b923da8a192fa9457cb8431110ef3/snapshot.go#L291)):*

Just a small preference to wait once for the highest seqnum we found, so that we delay reacquiring `DB.mu` since it may be a (mildly?) contended mutex.
</blockquote></details>

I don't see the difference, it's not like Wait knows what we're waiting for, it would be the same number of Wait calls

PR #5395, opened 2025-10-03
crl-release-25.4: block: minor improvements to compression profiles

RaduBerinde:
We change the "good" profile to use Zstd1 instead of adaptively
choosing between MinLZ1 and Zstd3. We also reduce some of the
thresholds and improve comments.

We also update the compression analyzer profiles. We no longer measure
MinLZ2 since it's almost as slow as Zstd1 with much smaller size
benefit over MinLZ1.


RaduBerinde:
TFTR!

PR #5402, opened 2025-10-05
db: clean up some nil Options paths

RaduBerinde:
We no longer use nil Options; clean up some paths that tolerated it.


RaduBerinde:
TFTR!

PR #5403, opened 2025-10-06
wal: deflake TestManagerFailover

### LINKED ISSUE #5401: master: nightly linux-race failed
The nightly linux-race test run failed on 3214797c0e799fa7fec3c9bfcea7deff8b55a0e6 (go1.23). Please review the run [here](https://github.com/cockroachdb/pebble/actions/runs/18257234316).

Jira issue: PEBBLE-1210

jbowens:
The invocation of the new segmentClosed callback introduced in #5388 occurs asynchronously with respect to the manager and the progression through logical log numbers. This test was flaky in two ways: If the segmentClosed callback was invoked /before/ the writerClosed callback for the same WAL, writerClosed would append a second record for the same logical log containing the set of WALs other than the one inserted by the segmentClosed callback. Conversely, if the segmentClosed callback was sufficiently delayed relative to the closing of the writer, it might not be invoked until after the test listed the set of obsolete logs.

The first race is fixed by a refactoring of the segmentClosed and writerClosed callbacks, adapting them to share the same logic for merging logs. The second race is fixed through use of (datadriven.TestData).Retry to account for the nondeterminism.

Fix #5401.

### DIFF
--- a/wal/failover_manager.go
+++ b/wal/failover_manager.go
@@ -440,6 +440,10 @@ type segmentWithSizeEtc struct {
 	synchronouslyClosed bool
 }
 
+func cmpSegmentWithSizeEtc(a, b segmentWithSizeEtc) int {
+	return cmp.Compare(a.segment.logNameIndex, b.segment.logNameIndex)
+}
+
 type failoverManager struct {
 	opts Options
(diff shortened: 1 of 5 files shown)


jbowens:
TFTR!

PR #5407, opened 2025-10-06
db: add DBCompressionFast

RaduBerinde:
Add `DBCompressionSettings` that uses `FastCompression` on L6 (which
automatically chooses between MinLZ or Zstd1 for value blocks).


RaduBerinde:
TFTR!

PR #5408, opened 2025-10-07
db: don't store table fromat in compaction

RaduBerinde:
We currently store a table format in compaction. In practice, this is
always `d.TableFormat()`, and we don't even end up using it when we
write the output (we use `d.TableFormat()` again).

This change removes the table format from the compaction to simplify
things a bit. This allows adding a simpler `DB.makeWriterOptions()`
(which makes future extensions easier).

### DIFF
--- a/blob_rewrite.go
+++ b/blob_rewrite.go
@@ -320,7 +320,7 @@ func (d *DB) runBlobFileRewriteLocked(
 	if err != nil {
 		return objstorage.ObjectMetadata{}, nil, err
 	}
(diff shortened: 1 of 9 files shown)


annrpom (approved):
:lgtm:

> @annrpom reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5408#-:-OazKEn0CieyOOXm3AZH:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!!

Issue #5377, opened 2025-09-30
db: make minimum bytes eligible for likely mvcc garbage separation configurable


annrpom:

annrpom:
We backported something for this in 25.4 - https://github.com/cockroachdb/pebble/pull/5505; going to get stuff merged on master before I close this out

PR #5410, opened 2025-10-07
db: min size for mvcc separation should be configurable

annrpom:
We will add a configurable option, `LatencyTolerantSpanPolicy`, that
specifies the minimum value size required for latency tolerant spans
and likely MVCC garbage to be separated into a blob file.

Fixes: #5377

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -3476,14 +3476,15 @@ func (d *DB) compactAndWrite(
 			writerOpts.Compression = block.FastestCompression
 		}
 		vSep := valueSeparation
-		switch spanPolicy.ValueStoragePolicy {
-		case ValueStorageLowReadLatency:
+		switch spanPolicy.ValueStoragePolicy.PolicyAdjustment {
+		case NoValueSeparation:
 			vSep = compact.NeverSeparateValues{}
-		case ValueStorageLatencyTolerant:
+		case Override:
 			// This span of keyspace is more tolerant of latency, so set a more
 			// aggressive value separation policy for this output.
 			vSep.SetNextOutputConfig(compact.ValueSeparationOutputConfig{
-				MinimumSize: latencyTolerantMinimumSize,
+				MinimumSize:            spanPolicy.ValueStoragePolicy.MinimumSize,
+				MinimumMVCCGarbageSize: spanPolicy.ValueStoragePolicy.MinimumMVCCGarbageSize,
 			})
 		}
 		objMeta, tw, err := d.newCompactionOutputTable(jobID, c, writerOpts)
--- a/data_test.go
+++ b/data_test.go
@@ -1783,21 +1783,27 @@ func parseDBOptionsArgs(opts *Options, args []datadriven.CmdArg) error {
 			}
 			policy := SpanPolicy{
 				DisableValueSeparationBySuffix: true,
-				ValueStoragePolicy:             ValueStorageLowReadLatency,
+				ValueStoragePolicy: ValueStoragePolicy{
+					PolicyAdjustment: NoValueSeparation,
+				},
 			}
 			spanPolicies = append(spanPolicies, SpanAndPolicy{
 				KeyRange: span,
 				Policy:   policy,
 			})
-		case "latency-tolerant-span":
+		case "override-span":
 			if len(cmdArg.Vals) != 2 {
-				return errors.New("latency-tolerant-span expects 2 arguments: <start-key> <end-key>")
+				return errors.New("override-span expects 2 arguments: <start-key> <end-key>")
 			}
 			span := KeyRange{
 				Start: []byte(cmdArg.Vals[0]),
 				End:   []byte(cmdArg.Vals[1]),
 			}
-			policy := SpanPolicy{ValueStoragePolicy: ValueStorageLatencyTolerant}
+			policy := SpanPolicy{ValueStoragePolicy: ValueStoragePolicy{
+				PolicyAdjustment:       Override,
+				MinimumSize:            latencyTolerantMinimumSize,
+				MinimumMVCCGarbageSize: 1,
+			}}
 			spanPolicies = append(spanPolicies, SpanAndPolicy{
 				KeyRange: span,
 				Policy:   policy,
@@ -1868,6 +1874,11 @@ func parseDBOptionsArgs(opts *Options, args []datadriven.CmdArg) error {
 							return err
 						}
 						policy.GarbageRatioHighPriority, err = strconv.ParseFloat(value[i+1:], 64)
+						if err != nil {
+							return err
+						}
+					case "min-mvcc-garbage-size":
+						policy.MinimumMVCCGarbageSize, err = strconv.Atoi(value)
 						if err != nil {
 							return err
 						}
--- a/event_listener_test.go
+++ b/event_listener_test.go
@@ -633,9 +633,10 @@ func TestBlobCorruptionEvent(t *testing.T) {
 			}
 			opts.Experimental.ValueSeparationPolicy = func() ValueSeparationPolicy {
 				return ValueSeparationPolicy{
-					Enabled:               true,
-					MinimumSize:           1,
-					MaxBlobReferenceDepth: 10,
+					Enabled:                true,
+					MinimumSize:            1,
+					MinimumMVCCGarbageSize: 1,
+					MaxBlobReferenceDepth:  10,
 				}
 			}
 			d, err := Open("", opts)
--- a/internal/compact/run.go
+++ b/internal/compact/run.go
@@ -115,6 +115,9 @@ type ValueSeparationOutputConfig struct {
 	// MinimumSize is the minimum size of a value that will be separated into a
 	// blob file.
 	MinimumSize int
+	// MinimumMVCCGarbageSize is the minimum size of a value that will be
+	// separated into a blob file if the value is likely MVCC garbage.
+	MinimumMVCCGarbageSize int
 }
 
 // ValueSeparation defines an interface for writing some values to separate blob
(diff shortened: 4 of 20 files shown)


RaduBerinde:
*[`options.go` line 822 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5410#-Ob9q-uME6dxYtsVISM3:-ObDRxIc3n4MfjZPohtj:b-q931r2) ([raw file](https://github.com/cockroachdb/pebble/blob/1c0a72cfc9fa1504eb353d6f1420559af1807fed/options.go#L822)):*

I am always a bit confused about our configuration approach to value separation.
- We have a globally applicable ValueSeparationPolicy with many parameters. This is very reasonable.
- We tweak it in two ways (a) no value separation, (b) the minimum size, based on the ValueStoragePolicy enum in the SpanPolicy. This tweaking happens in code, instead of the SpanPolicy explicitly specifying the pair (doValueSeparation, minimumSize), where both of the elements in that tuple would allow a special value that says "use global setting". In general, mapping things from enum into configs is probably best left to the highest layer (like CockroachDB).

Now we are adding another option that makes that minimum size configurable and it uses a func, when we already have a dynamic `ValueSeparationPolicy` via `Options.Experimental.ValueSeparationPolicy` func and we have a dynamic `SpanPolicy` to override the global behavior via `SpanPolicyFunc`. I think this is making the configuration more confusing. And we seem to have decided that the minimum size should be the same for latency tolerant and MVCC garbage -- that decision is better left to CockroachDB.

How about something like:

```
type ValueSeparationPolicy struct {
   ...
   MinimumSize int
   // Applies only to SpanPolicys that permit separation of MVCC garbage, which is also the default. 
   MinimumMVCCGarbageSize int
    ...
}

type SpanPolicy struct {
   ...
   ValueStoragePolicy ValueStoragePolicy
}

type ValueStoragePolicy {
   PolicyAdjustment ValueStoragePolicyAdjustment
   // Remaining fields are ignored, unless the adjustment is Override.
   HasMVCCGarbage bool
   MinimumSize int
   MinimumMVCCGarbageSize int.
}

// Possibe values are UseDefault, NoValueSeparation, Override.
type ValueStoragePolicyAdjustment uint8

```

@jbowens
</blockquote></details>

I agree - Pebble can define a "recommended" low latency value for `ValueStoragePolicy` that CRDB can use to populate the field - but it should all be returned through the span policy.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5410)* status: 3 of 26 files reviewed, 3 unresolved discussions (waiting on @annrpom, @jbowens, @RaduBerinde, and @sumeerbhola)

___
*[`value_separation.go` line 45 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5410#-Oc1jzYPEXe9YA3lOtdx:-Oc2K2uhE62bVHP-qxXL:bbnuk3h) ([raw file](https://github.com/cockroachdb/pebble/blob/7e2405577376a7c93efad8aaa3c59efe3ae1ceae/value_separation.go#L45)):*

> When there isn't a single policy, we used to pass `ValueStorageDefault`. Now we have a struct. The calling code of `determine...` seems unchanged

I didn't follow this; I didn't see any changes from enum to struct.

> This seems dubious since there could be many policies spanning the compaction interval and we are looking at one policy (perhaps the default) and it not everything matches we would rewrite the references?

We split sstables at new policies during Open, so we expect the vast majority of compactions to be within the same span policy. We establish the value separation of the entire compaction span upfront and pass that in:
```
        spanPolicy, spanPolicyEndKey, err = d.opts.Experimental.SpanPolicyFunc(c.bounds.Start)
	if err != nil {
		return runner.Finish().WithError(err)
	}
	if len(spanPolicyEndKey) == 0 || d.cmp(c.bounds.End.Key, spanPolicyEndKey) < 0 {
		compactionValueStoragePolicy = spanPolicy.ValueStoragePolicy
	}
```
This allows us to rewrite blob files if any of the inputs were encoded using a different policy.

(still parsing the comments here)
</blockquote></details>

> Then I noticed `preserveBlobReferences` doesn't do anything in `SetNextOutputConfig`, so the `preserveBlobReferences.MinimumSize` return value is quite arbitrary, which we then fill into the sstable properties `tw.SetValueSeparationProps`.

The existing logic looks right to me. We'll only preserve blob references if all the compaction inputs used the same minimum size for the purpose of separating values. So this propagated minimum size holds for the outputted sstables as well.

However by making this additional value size configurable, the policy can no longer be expressed by just 1 minimum size. We now also have to capture the minimum size of MVCC garbage at the time it was written.

sumeerbhola (changes requested):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5410)* status: 3 of 26 files reviewed, 3 unresolved discussions (waiting on @annrpom, @jbowens, and @RaduBerinde)

___
*[`options.go` line 822 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5410#-Ob9q-uME6dxYtsVISM3:-Oc39EwE5uprXCc_Ghz1:b-lqgns0) ([raw file](https://github.com/cockroachdb/pebble/blob/1c0a72cfc9fa1504eb353d6f1420559af1807fed/options.go#L822)):*

At the SpanPolicy level we currently have `DisableValueSeparationBySuffix`:
```
	// DisableValueSeparationBySuffix disables discriminating KVs depending on
	// suffix.
	//
	// Among a set of keys with the same prefix, Pebble's default heuristics
	// optimize access to the KV with the smallest suffix. This is useful for MVCC
	// keys (where the smallest suffix is the latest version), but should be
	// disabled for keys where the suffix does not correspond to a version.
	DisableValueSeparationBySuffix bool
```
Are you suggesting moving it into the `ValueStoragePolicy`? I think it's currently at the higher level because it also configures whether to separate values into value blocks within the same sstable. I don't really have an opinion where it lives.
</blockquote></details>

Good point. I don't have an opinion either. Just wanted it captured somehow, and we already do.
Do we use it in blob value separation, or are we relying on NoValueSeparation?
___
*[`value_separation.go` line 45 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5410#-Oc1jzYPEXe9YA3lOtdx:-Oc39mqOETaz2RKnjM8r:b-7rqcf0) ([raw file](https://github.com/cockroachdb/pebble/blob/7e2405577376a7c93efad8aaa3c59efe3ae1ceae/value_separation.go#L45)):*
> I didn't follow this; I didn't see any changes from enum to struct.

ValueStoragePolicy used to be an enum and now it is a struct. I may be missing something.

> We split sstables at new policies during Open, so we expect the vast majority of compactions to be within the same span policy

Yep, I understand that part (mainly because I had to heavily modify this code for the tiered storage prototype).

> The existing logic looks right to me. We'll only preserve blob references if all the compaction inputs used the same minimum size for the purpose of separating values.

This seems risky since if we happen to have files of 2 policies in a compaction, we will unnecessarily rewrite the values. That is if [a,c) has min-size 100 and [c,e) has min-size 500, and we have a compaction over the interval [b,d) we would rewrite the blob references. Why don't we check whether the minimum size is correct by iterating over the SpanPolicys in the compaction and coordinate that iteration with that of the input files?

I suspect what you are saying is that because every level splits at these policy boundaries, we should never be expanding a compaction to cover multiple policies (unless the policy changed)? Am I understanding correctly? But we also add manual compactions, which aren't necessarily working with such constrained spans. I think if we are relying on some assumptions it would be nice to spell them out.

Also maybe worth taking a look at the value_separation.go, SpanPolicy, SpanPolicyFunc and DB.compactAndWrite -- there is iteration over the policies that happens for other reasons. It may help refining this in a way that will not need to be rethought when we add the TieringPolicy.

> One thing to keep in mind, I don't think we should go too hard on configuration clean up in this PR, because our intent is to backport the knob to 25.4

Oh, I wasn't aware of that. In that case I'd prefer for @annrpom to do a small PR that you review that then gets backported. We can come back to this larger one after that.

xinhaoz:
*[`value_separation.go` line 45 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5410#-Oc1jzYPEXe9YA3lOtdx:-OcllUd5A65iQTkMTt5-:b-53vxq9) ([raw file](https://github.com/cockroachdb/pebble/blob/7e2405577376a7c93efad8aaa3c59efe3ae1ceae/value_separation.go#L45)):*

> I didn't follow this; I didn't see any changes from enum to struct.

ValueStoragePolicy used to be an enum and now it is a struct. I may be missing something.

> We split sstables at new policies during Open, so we expect the vast majority of compactions to be within the same span policy

Yep, I understand that part (mainly because I had to heavily modify this code for the tiered storage prototype).

> The existing logic looks right to me. We'll only preserve blob references if all the compaction inputs used the same minimum size for the purpose of separating values.

This seems risky since if we happen to have files of 2 policies in a compaction, we will unnecessarily rewrite the values. That is if [a,c) has min-size 100 and [c,e) has min-size 500, and we have a compaction over the interval [b,d) we would rewrite the blob references. Why don't we check whether the minimum size is correct by iterating over the SpanPolicys in the compaction and coordinate that iteration with that of the input files?

I suspect what you are saying is that because every level splits at these policy boundaries, we should never be expanding a compaction to cover multiple policies (unless the policy changed)? Am I understanding correctly? But we also add manual compactions, which aren't necessarily working with such constrained spans. I think if we are relying on some assumptions it would be nice to spell them out.

Also maybe worth taking a look at the value_separation.go, SpanPolicy, SpanPolicyFunc and DB.compactAndWrite -- there is iteration over the policies that happens for other reasons. It may help refining this in a way that will not need to be rethought when we add the TieringPolicy.

> One thing to keep in mind, I don't think we should go too hard on configuration clean up in this PR, because our intent is to backport the knob to 25.4

Oh, I wasn't aware of that. In that case I'd prefer for @annrpom to do a small PR that you review that then gets backported. We can come back to this larger one after that.
</blockquote></details>

Synced with Annie after I happened to see the discussion here, and I applied some of the suggestions in https://github.com/cockroachdb/pebble/pull/5518 since I just merged some of the changes in question last week.
We can move review/discussion of the preserving logic there!

PR #5412, opened 2025-10-07
metrics: add compression counters

RaduBerinde:
#### base: move cache.Level to base

We also rework the implementation so it's less confusing.

#### metrics: add compression counters

Add running counters that keep track of how many bytes were compressed
and decompressed. The counters are segregated along the same lines
where compression settings can differ: L5 vs L6 vs other levels, and
data vs value vs other blocks.

The intention is to estimate the CPU usage change for a different
compression profile (in conjunction with data about each algorithm's
performance, as obtained by the compression analyzer).

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -2945,6 +2945,8 @@ func (d *DB) runCopyCompaction(
 		var wrote uint64
 		err = d.fileCache.withReader(ctx, block.NoReadEnv, inputMeta.VirtualMeta(), func(r *sstable.Reader, env sstable.ReadEnv) error {
 			var err error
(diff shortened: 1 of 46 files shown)


annrpom (approved):
:lgtm:

> @annrpom reviewed 15 of 15 files at r1, 32 of 32 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5412#-:-Ob-J5624DxyUhj64RgT:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

PR #5413, opened 2025-10-07
db: factor out recovery logic

jbowens:
Factor out logic to recover the current database state from Open. Over time Open has grown involved, interleaving initialization of pebble.DB state and recovery of the database. Teasing apart these functions help improve readability and will help in refactoring to clean up obsolete files before creating new ones during Open.

Future work will pull loading of the version set into RecoverState as well.

### DIFF
--- a/format_major_version.go
+++ b/format_major_version.go
@@ -417,13 +417,13 @@ func lookupFormatMajorVersion(
 	}
 	vers := FormatMajorVersion(v)
 	if vers == FormatDefault {
-		return 0, nil, errors.Newf("pebble: default format major version should not persisted", vers)
+		return 0, nil, errors.Newf("default format major version should not persisted", vers)
 	}
 	if vers > internalFormatNewest {
-		return 0, nil, errors.Newf("pebble: database %q written in unknown format major version %d", dirname, vers)
+		return 0, nil, errors.Newf("written in unknown format major version %d", vers)
 	}
 	if vers < FormatMinSupported {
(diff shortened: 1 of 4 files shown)


jbowens (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5413#-:-Ob3FNkd7Ckk1tzToWQF:b-814g8k)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @RaduBerinde)

___
*[`open.go` line 222 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5413#-Ob-L_MbEal2cZGFlmYV:-Ob3E92-2uSqJ_AXVlYo:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/beaabb358aaa05bdb580b908244df29983778de8/open.go#L222)):*

In error cases after initializing `d` we will Close the objProvider two times - once here, once in `rs.Close()`
</blockquote></details>

Done
___
*[`open.go` line 1236 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5413#-Ob-MuG20RQ-JD0RB9r3:-Ob3FM0j38jygmRx6NVZ:bb6un1j) ([raw file](https://github.com/cockroachdb/pebble/blob/beaabb358aaa05bdb580b908244df29983778de8/open.go#L1236)):*

[nit] it might be nicer to move most of this function to a `fs.init()` method and leave just:
```
if err := rs.init(opts, dirname); err != nil {
  rs.Close()
  return errors.Wrapf(err, ...)
}
return &rs, nil
```
</blockquote></details>

Done

PR #5415, opened 2025-10-08
block: improve adaptive compression profiles

RaduBerinde:
We allow different adaptive ratios for data and value blocks and
improve the profiles.


RaduBerinde:
TFTR!

PR #5423, opened 2025-10-13
db: Add method to change value for option `WALDir`

### LINKED ISSUE #5421: db: Allow to rename `WALDir`
There might be a situation, when a directory with WAL will be lost or moved to another mount point.
Currently, `open` fails with error `WALDir changed from previous options`.

**Suggestion:**
Allow to skip this check with additional option `SkipCheckCompatibility` or introduce new public method `pebble.RenameWALDir`

Jira issue: PEBBLE-1214

annikiforov:
Close #5421.

### DIFF
--- a/open.go
+++ b/open.go
@@ -1317,3 +1317,64 @@ func (l walEventListenerAdaptor) LogCreated(ci wal.CreateInfo) {
 	}
 	l.l.WALCreated(wci)
 }
+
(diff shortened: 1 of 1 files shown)


annikiforov:
Addressed here https://github.com/cockroachdb/pebble/pull/5478.

Issue #5424, opened 2025-10-14
db: improve obsolete file deletion pacing and queuing to prevent backlog and blocking


sumeerbhola:
Motivated by observations from https://github.com/cockroachlabs/support/issues/3457 and https://cockroachlabs.slack.com/archives/C09KSGD2XNX, where the cleanupManager's finite queue of jobs (length 1000) became full and blocked compactions for up to a minute, and sometimes blocked flushes. The former resulted in no compactions from L0=>Lbase which eventually resulted in AC throttling in CockroachDB. The latter resulted in write stalls (this was rarer, since most flushes do not generate obsolete WALs due to WAL recycling).

Listing all the desired cleanup improvements:
- [x] Queue of cleanup jobs should not be bounded: Blocking compactions or flushes is dangerous. And we need to be able to absorb spikes. And if deletion is chronically slow because of the filesystem, but compactions/flushes are not slow, that is a very odd grey failure mode, and we should ignore it.

- [x] Obsolete bytes used to pace deletions should include the bytes queued in the `cleanupManager`.

- [x] Configurability: The lower bound pacing rate (`Options.TargetByteDeletionRate`) should be dynamically updatable without restarting all the nodes (and hooked up to a CockroachDB cluster setting).

- [x] Observability: The backlog of bytes and files in the `cleanupManager` should be observable via gauge metrics. The cumulative byte deletion should be observable via a counter.

Some subset of the above needs to be backported to CockroachDB v25.2.

Jira issue: PEBBLE-1216

PR #5430, opened 2025-10-14
db: add fail-safe to the delete pacer

RaduBerinde:
logic fails to keep up.

Informs #5424

### DIFF
--- a/obsolete_files.go
+++ b/obsolete_files.go
@@ -65,6 +65,8 @@ func (m *cleanupManager) CompletedStats() obsoleteObjectStats {
 
 // We can queue this many jobs before we have to block EnqueueJob.
 const jobsQueueDepth = 1000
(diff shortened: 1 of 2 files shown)


RaduBerinde:
*[`obsolete_files.go` line 228 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5430#-Ob_3M5mBD8vokmBCq-Z:-Ob_6k5gEHuPiMjKAEkF:b-kdjhbh) ([raw file](https://github.com/cockroachdb/pebble/blob/a5e7243b5fe226e9473457a5017460451ac70fd6/obsolete_files.go#L228)):*

do we need to print a log here to indicate the pacing is disabled?
</blockquote></details>

There is already a log at this threshold in `maybeLogLocked`

PR #5445, opened 2025-10-17
db: fix BlobBytesCompacted accounting

jbowens:
Previously flushes would contribute the size of written blob files to both BlobBytesFlushed and BlobBytesCompacted, artificially inflating the BytesCompacted total. This commit fixes the metric update to include the blob files in either the flushes total or the compacted total but not both.


sumeerbhola (approved):
:lgtm: 

> @sumeerbhola reviewed 1 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5445#-:-Oc1_0Pz2hYfHQAiQwP3:bg7igkv)* status: 2 of 3 files reviewed, 1 unresolved discussion

PR #5449, opened 2025-10-18
db: delete pacer rewrite

### LINKED ISSUE #5424: db: improve obsolete file deletion pacing and queuing to prevent backlog and blocking

RaduBerinde:
This change reimplements the delete pacer with an improved design. The
previous implementation was prone to falling behind on deletions; the
"obsolete ratio" feedback mechanism was broken because we didn't count
the queued deletions as obsolete bytes. The implementation was also
unnecessarily complicated and intertwined with the rest of the DB
code.

The new delete pacer is in a separate package with clear API
boundaries. It internally maintains all relevant metrics.  It is
designed to increase the deletion rate whenever deletions stay in the
queue for more than 5 minutes (it no longer depends on the obsolete
bytes ratio). In addition, instead of proportional control (where the
rate is proportional to the backlog, thus going down as we clear it)
we switch to constant-horizon control where we keep the same rate
until the backlog is cleared.

The implementation uses an unbounded `fifo.Queue` instead of a channel
of limited size (which can cause unwanted blocking). We still have a
safety valve to disable pacing if the queue gets too big.

We add delete pacer metrics so that we can monitor the deletion rate
and queue depth.

Fixes #5424


RaduBerinde:
linux-race failure caused by `TestIteratorErrors` not closing the DB and injecting an error in `GetDiskUsage`. I'm looking at it separately.

PR #5451, opened 2025-10-19
db: initialize more LevelMetrics fields on recovery

RaduBerinde:
We weren't initializing all `LevelMetrics` fields when reading a
manifest. This isn't a big deal in practice because on the first
version change, we will reinitialize these fields.

This change reuses the code that normally sets these fields when
updating the version.


RaduBerinde:
TFTR!

PR #5452, opened 2025-10-19
db: remove unnecessary metric updates

RaduBerinde:
When we apply a new version, we also pass "delta metrics" that are
added to the current `LevelMetrics`.

Some of `LevelMetrics` fields (like table counts/sizes, estimated
ref size) are reset from the `LevelMetadata` from the new version.
These fields don't require maintenance through the delta metrics.

This change removes the ineffectual code which attempts to keep these
fields updated through the delta metrics.


RaduBerinde:
TFTR!

PR #5453, opened 2025-10-19
crl-release-25.4: db: add fail-safe to the delete pacer

RaduBerinde:
logic fails to keep up.

Release justification: mitigation of serious condition encountered in production.

CC @cockroachdb/release 

Informs #5424


sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5453#-:-Oc0w4sD5rWzqxV5XEQL:bg7igkv)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jbowens)

PR #5456, opened 2025-10-19
crl-release-24.3: db: add fail-safe to the delete pacer

RaduBerinde:
logic fails to keep up.

Release justification: mitigation of serious condition encountered in
production.

Informs #5424


RaduBerinde:
TFTR!

PR #5458, opened 2025-10-20
compact: make `TombstoneDenseCompactionThreshold` dynamically reconfigurable

### LINKED ISSUE #5457: make tombstone compaction threshold configurable
On some workloads, tombstone compactions can lead to increased resource usage with no visible benefit on latencies. We need the threshold to be configurable so we can easily turn it down or off in these cases.

Jira issue: PEBBLE-1225

ever0de:
fixes #5457

Previously, TombstoneDenseCompactionThreshold was a static float64 value that could only be set at initialization time. This made it impossible to dynamically adjust the threshold based on changing workload characteristics or operational requirements.

This commit changes TombstoneDenseCompactionThreshold from a float64 to a function returning float64, following the pattern used by other dynamic options like CompactionGarbageFractionForMaxConcurrency and VirtualTableRewriteUnreferencedFraction. The default value remains 0.10.

### DIFF
--- a/metamorphic/options_test.go
+++ b/metamorphic/options_test.go
@@ -85,6 +85,7 @@ func TestOptionsRoundtrip(t *testing.T) {
 		"Experimental.RemoteStorage:",
 		"Experimental.SingleDeleteInvariantViolationCallback:",
 		"Experimental.EnableDeleteOnlyCompactionExcises:",
+		"Experimental.TombstoneDenseCompactionThreshold:",
 		"Experimental.ValueSeparationPolicy:",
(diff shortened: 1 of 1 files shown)


RaduBerinde (approved):
:lgtm:

Looks great, thanks!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5458#-:-Oc1RFFY7_UVhUtqxKfj:buvclpw)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @sumeerbhola)

PR #5465, opened 2025-10-20
db: don't inject GetDiskUsage errors in TestIteratorErrors

RaduBerinde:
Add a more general `OpKindIn` errorfs predicate and use it to disable
errors in `GetDiskUsage` (which causes a background error to be
reported, which causes a test failure).


RaduBerinde:
TFTR!

PR #5478, opened 2025-10-21
options: add Unsafe.AllowMissingWALDir option

### LINKED ISSUE #5421: db: Allow to rename `WALDir`

RaduBerinde:
This unsafe option allows opening a store after moving or losing the
previous WAL dir. This is understood to be risky and have associated
data loss if any of the necessary WALs are not in the new directory.

Fixes #5421

### DIFF
--- a/open_test.go
+++ b/open_test.go
@@ -242,6 +242,12 @@ func TestOpen_WALFailover(t *testing.T) {
 				switch cmdArg.Key {
 				case "path":
 					o.FS, dataDir = extractFSAndPath(cmdArg)
+				case "wal-dir":
+					fs, dir := extractFSAndPath(cmdArg)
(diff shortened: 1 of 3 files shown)


RaduBerinde:
CC @annikiforov - this should address your problem, correct?

annikiforov:
> CC @annikiforov - this should address your problem, correct?

Absolutely, perfect!

RaduBerinde:
TFTR!

PR #5482, opened 2025-10-22
valsep: introduce new generalized blob writer

xinhaoz:
### valsep: introduce new generalized blob writer

generalizedValueSeparation is a new ValueSeparation implementation that combines
the existing implementations from writeNewBlobs and preserveBlobReferences. It
is able to function in either mode (writing new blobs or preserving all blob
refs).

Currently, we either write new blob refs or preserve blob refs when writing an
sst. In the future (for tiered storage), we may do both.

The next commits will replace the old implementations with the generalized
separator.

### db: replace preserveBlobReferences

### DIFF
--- a/compaction_value_separation_test.go
+++ b/compaction_value_separation_test.go
@@ -78,15 +78,20 @@ func TestValueSeparationPolicy(t *testing.T) {
 				case "never-separate-values":
 					vs = valsep.NeverSeparateValues{}
 				case "preserve-blob-references":
-					pbr := &preserveBlobReferences{}
 					lines := crstrings.Lines(d.Input)
-					pbr.inputBlobPhysicalFiles = make(map[base.BlobFileID]*manifest.PhysicalBlobFile, len(lines))
+					inputBlobPhysicalFiles := make(map[base.BlobFileID]*manifest.PhysicalBlobFile, len(lines))
 					for _, line := range lines {
 						bfm, err := manifest.ParseBlobFileMetadataDebug(line)
 						require.NoError(t, err)
 						fn = max(fn, bfm.Physical.FileNum)
-						pbr.inputBlobPhysicalFiles[bfm.FileID] = bfm.Physical
+						inputBlobPhysicalFiles[bfm.FileID] = bfm.Physical
 					}
+					pbr := valsep.NewPreserveAllHotBlobReferences(
+						inputBlobPhysicalFiles, /* blob file set */
+						manifest.BlobReferenceDepth(0),
+						sstable.ValueSeparationDefault,
+						0, /* minimum size */
+					)
 					vs = pbr
 				case "write-new-blob-files":
 					var minimumSize int
@@ -212,7 +217,7 @@ func (w *loggingRawWriter) AddWithBlobHandle(
 type defineDBValueSeparator struct {
 	bv    blobtest.Values
 	metas map[base.BlobFileID]*manifest.PhysicalBlobFile
-	pbr   *preserveBlobReferences
+	pbr   valsep.ValueSeparation
 	kv    base.InternalKV
 }
 
@@ -274,10 +279,6 @@ func (vs *defineDBValueSeparator) Add(
 	meta.Size += uint64(lv.Fetcher.Attribute.ValueLen)
 	meta.ValueSize += uint64(lv.Fetcher.Attribute.ValueLen)
 
-	if vs.pbr.inputBlobPhysicalFiles == nil {
-		vs.pbr.inputBlobPhysicalFiles = make(map[base.BlobFileID]*manifest.PhysicalBlobFile)
-	}
-	vs.pbr.inputBlobPhysicalFiles[fileID] = meta
 	// Return a KV that uses the original key but our constructed blob reference.
 	vs.kv.K = kv.K
 	vs.kv.V = iv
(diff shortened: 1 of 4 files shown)


compaction_value_separation_test.go
xinhaoz: Planning on moving this test file to the valsep pkg in a follow-up PR.

annrpom (approved):
:lgtm: cool!!

> @annrpom reviewed 2 of 2 files at r1, 3 of 3 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5482#-:-OcHlema5ztqkAJkcw0E:bvwh8ig)* status: all files reviewed, 1 unresolved discussion

xinhaoz (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5482)* status: 1 of 6 files reviewed, 4 unresolved discussions (waiting on @annrpom, @jbowens, and @RaduBerinde)

___
*[`valsep/generalized_value_separation.go` line 36 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5482#-OcIVEonCbPacHvjQQMD:-OcMgqZn5ukgKOUhFzS_:b3rhet3) ([raw file](https://github.com/cockroachdb/pebble/blob/b4a6dd9bd8bb80415278d3efa60a1e4ad07e61e7/valsep/generalized_value_separation.go#L36)):*

nit: and maybe call the type `Separator` or something; the 'Generalized' makes sense in the context of the previous implementations, but less as a standalone name going forward
</blockquote></details>

Ah good idea! If it's alright, I think I'll consolidate NeverSeparateValues and remove the interface in the next PR when the tests are also moved, since there's also a testing impl of the interface right now. Renamed this to ValueSeparator.
___
*[`valsep/generalized_value_separation.go` line 39 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5482#-OcIW5EiBFxcyTbpUH5l:-OcMf316DAq6vsVDSea7:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b4a6dd9bd8bb80415278d3efa60a1e4ad07e61e7/valsep/generalized_value_separation.go#L39)):*

[nit] The name makes sense in light of the history of the code but it's a bit confusing when you just think of the latest state of the code. Maybe `valsep.Router` or `valsep.Separator`?
</blockquote></details>

Done.
___
*[`valsep/generalized_value_separation.go` line 65 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5482#-OcMOCE01aZQ5WTL_YF5:-OcMgG74DKVHNk7PmeoG:b17gioo) ([raw file](https://github.com/cockroachdb/pebble/blob/b4a6dd9bd8bb80415278d3efa60a1e4ad07e61e7/valsep/generalized_value_separation.go#L65)):*

> It may be overridden per-output by SetNextOutputConfig.

Is this true? My reading was that we'll update `minimumSize` but not `globalMinimumSize`
</blockquote></details>

Oops, your understanding is correct. IDK how that got there, fixed.
___
*[`valsep/generalized_value_separation.go` line 80 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5482#-OcMOZij1JcE2SwtUhJw:-OcMfRMw5fdzwt1Yxif9:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b4a6dd9bd8bb80415278d3efa60a1e4ad07e61e7/valsep/generalized_value_separation.go#L80)):*

nit: s/currPreservedReferenceValueSizes/currPendingReferences/

i think?
</blockquote></details>

Done.
___
*[`valsep/generalized_value_separation.go` line 202 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5482#-OcIVWr9DNQSMXbK5LUM:-OcMfZ_40yJ4ziunQNZu:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b4a6dd9bd8bb80415278d3efa60a1e4ad07e61e7/valsep/generalized_value_separation.go#L202)):*

[nit] remove `compact.` or replace with `valsep`.
</blockquote></details>

Done.
___
*[`valsep/generalized_value_separation.go` line 259 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5482#-OcMNBE97QuZd-XEYCxj:-OcMfcF_9w5ZCD4NMkNp:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b4a6dd9bd8bb80415278d3efa60a1e4ad07e61e7/valsep/generalized_value_separation.go#L259)):*

I'm noticing that in this code and in the existing code, the `SpanPolicy.DisableValueSeparationBySuffix` setting doesn't apply here; mind leaving a TODO to address that? I think when `DisableValueSeparationBySuffix=true`, we should _not_ separate values even if `isLikelyMVCCGarbage=true`.
</blockquote></details>

Done.
___
*[`valsep/generalized_value_separation.go` line 278 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5482#-OcMQ5oQ9oBdJNLr-gOI:-OcMgBnj0B4K4BJI--pb:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b4a6dd9bd8bb80415278d3efa60a1e4ad07e61e7/valsep/generalized_value_separation.go#L278)):*

nit: how about return `errors.AssertionFailedf("...")`
</blockquote></details>

Done.

PR #5485, opened 2025-10-23
db: re-export deletepacer.Options

jbowens:
Re-export the internal/deletepacer.Options struct from the pebble package so that clients can configure it.


xinhaoz (approved):

RaduBerinde:
Is it necessary? I thought we can set each field directly.

PR #5486, opened 2025-10-23
crl-release-25.4: db: add configurable latency tolerant minimum size

annrpom:
We'll allow both likely MVCC garbage and latency tolerant values to have their minimum size to be separated configurable via a field in our ValueSeparationPolicy.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -3484,7 +3484,7 @@ func (d *DB) compactAndWrite(
 			// This span of keyspace is more tolerant of latency, so set a more
 			// aggressive value separation policy for this output.
 			vSep.SetNextOutputConfig(compact.ValueSeparationOutputConfig{
-				MinimumSize: latencyTolerantMinimumSize,
+				MinimumSize: d.opts.Experimental.ValueSeparationPolicy().MinimumLatencyTolerantSize,
 			})
 		}
 		objMeta, tw, err := d.newCompactionOutputTable(jobID, c, writerOpts)
--- a/data_test.go
+++ b/data_test.go
@@ -1868,6 +1868,11 @@ func parseDBOptionsArgs(opts *Options, args []datadriven.CmdArg) error {
 						if err != nil {
 							return err
 						}
+					case "latency-tolerant-min-size":
+						policy.MinimumLatencyTolerantSize, err = strconv.Atoi(value)
+						if err != nil {
+							return err
+						}
 					default:
 						return errors.Newf("unrecognized value-separation argument %q", name)
 					}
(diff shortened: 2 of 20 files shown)


jbowens (approved):
> @jbowens reviewed 6 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5486)* status: 6 of 19 files reviewed, 1 unresolved discussion

___
*[`options.go` line 1834 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5486#-OcM_mO83bQSLAcPDPEG:-OcM_mO83bQSLAcPDPEH:b23adlc) ([raw file](https://github.com/cockroachdb/pebble/blob/d5954a99d1532d3e05ede9731a02569f7ae9dcb0/options.go#L1834)):*
> ```Go
> 			fmt.Fprintf(&buf, "  enabled=%t\n", policy.Enabled)
> 			fmt.Fprintf(&buf, "  minimum_size=%d\n", policy.MinimumSize)
> 			fmt.Fprintf(&buf, "  minimum_latency_tolerant_size=%d\n", policy.MinimumLatencyTolerantSize)
> ```

i think we should omit it from the OPTIONS file in the backport. IIRC, Pebble will error if it sees an option in the OPTIONS file that it doesn't recognize, which could interfere with an operator's ability to revert to a prior patch release.

RaduBerinde:
*[`options.go` line 1834 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5486#-OcM_mO83bQSLAcPDPEG:-OcMjK_jCNwFbwY4d1IH:b-ytr99y) ([raw file](https://github.com/cockroachdb/pebble/blob/d5954a99d1532d3e05ede9731a02569f7ae9dcb0/options.go#L1834)):*

i think we should omit it from the OPTIONS file in the backport. IIRC, Pebble will error if it sees an option in the OPTIONS file that it doesn't recognize, which could interfere with an operator's ability to revert to a prior patch release.
</blockquote></details>

If that's the case, it could still be a problem if a 25.4-26.1 upgrade is reverted before finalization, no?

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5486)* status: 6 of 19 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`options.go` line 1834 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5486#-OcM_mO83bQSLAcPDPEG:-OcMoVNC4EZDq3Y2gnrZ:b-3z6qp4) ([raw file](https://github.com/cockroachdb/pebble/blob/d5954a99d1532d3e05ede9731a02569f7ae9dcb0/options.go#L1834)):*

If that's the case, it could still be a problem if a 25.4-26.1 upgrade is reverted before finalization, no?
</blockquote></details>

Yeah, it's true. Maybe the answer is we should just tolerate unrecognized options

annrpom:
TFTRs! ('-')7

PR #5488, opened 2025-10-23
metrics: clean up setting of obsolete metrics

RaduBerinde:
There is no reason to actively maintain the stats for the
`obsoleteTables`/`obsoleteBlobs` lists. We can calculate them
when we get the metrics, making things much easier to follow.

The "dual" meaning of these fields - inside `versionSet.Metrics`
they do not include enqueued files but in the result of `Metrics()`
they do include them - was the root cause of the delete pacer
issue #5424.


jbowens (approved):
:lgtm_strong: 

> @jbowens reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5488#-:-OcM_-ST2dzXpMT5KjXl:b-uwkowd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaduBerinde)

RaduBerinde:
TFTR!

PR #5491, opened 2025-10-24
db: fix Metrics.String ingestion count

jbowens:
The Metrics.String output was printing the WAL.BytesIn+WAL.BytesWritten as the ingestion count.


annrpom (approved):
:lgtm: nice catch

Issue #5493, opened 2025-10-24
metrics: local blob files not initialized on open; affects DiskSpaceUsage


RaduBerinde:
Minimal reproduction: add to `testdata/metrics`:
```
init
----

batch
set a 012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789
set b 012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789
set c 012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789
set d 012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789
----


flush
----
L0.0:
  000005:[a#10,SET-a#10,SET]
  000007:[b#11,SET-b#11,SET]
  000009:[c#12,SET-c#12,SET]
  000011:[d#13,SET-d#13,SET]
Blob files:
  B000006 physical:{000006 size:[208 (208B)] vals:[120 (120B)]}
  B000008 physical:{000008 size:[208 (208B)] vals:[120 (120B)]}
  B000010 physical:{000010 size:[208 (208B)] vals:[120 (120B)]}
  B000012 physical:{000012 size:[208 (208B)] vals:[120 (120B)]}

disk-usage
----
7,614B

init reopen
----

disk-usage
----
6,265B
```

This affects `DB.DiskSpaceUsage()` which makes it into the visible store info (and probably used by the allocator?)

Related - we should be showing local tables and local blob files in the FILES metrics output.

Jira issue: PEBBLE-1236

RaduBerinde:
I confirmed with 25.4.0-rc.1:

<img width="429" height="207" alt="Image" src="https://github.com/user-attachments/assets/9c383856-c0a8-44d6-9cc2-7b49315029b3" />

After restart:

<img width="475" height="203" alt="Image" src="https://github.com/user-attachments/assets/38551d2a-dd29-4c88-95ce-337caf94de9e" />

PR #5494, opened 2025-10-24
crl-release-25.4: metrics: fix DiskSpaceUsage after reopen with value sep

RaduBerinde:
We weren't initializing `BlobFiles.Local.LiveSize` on open, which
should be the dominating term in disk space when values are separated.

Informs #5493


RaduBerinde:
TFTR!

PR #5503, opened 2025-10-28
crl-release-25.2: update post-issue action

RaduBerinde:
Update the post-issue action on this branch. It is currently
incompatible with the "parent" workflows that use it.


RaduBerinde:
TFTR!

PR #5504, opened 2025-10-28
metrics: redesign disk usage metrics

RaduBerinde:
The way we maintain the metrics that are used for disk usage
calculations are very messy. They are updated incrementally with
special code paths which can be fragile. And they are not consistent
with sibling metrics (for example: `Tables.Live.All` uses virtual table
sizes, whereas `Tables.Live.Local` uses backing sizes)

This change redesigns these metrics. We specialize the metrics to
focus on disk usage, and thus only include physical tables and
physical backings. We now use the (improved) disk space aggregator and
new blob size aggregator on demand, only when metrics are needed.

### DIFF
--- a/blob_rewrite.go
+++ b/blob_rewrite.go
@@ -259,8 +259,8 @@ func (c *blobFileRewriteCompaction) Execute(jobID JobID, d *DB) error {
 					// We don't know the size of the output blob file--it may have
 					// been half-written. We use the input blob file size as an
 					// approximation for deletion pacing.
-					FileSize: c.input.Physical.Size,
-					IsLocal:  true,
+					FileSize:  c.input.Physical.Size,
+					Placement: base.Local,
 				},
 			})
 		}
--- a/compaction.go
+++ b/compaction.go
@@ -2677,7 +2677,7 @@ func (d *DB) cleanupVersionEdit(ve *manifest.VersionEdit) {
 				FileNum:  ve.NewBlobFiles[i].Physical.FileNum,
 				FileSize: ve.NewBlobFiles[i].Physical.Size,
 			},
-			isLocal: objstorage.IsLocalBlobFile(d.objProvider, ve.NewBlobFiles[i].Physical.FileNum),
+			placement: objstorage.Placement(d.objProvider, base.FileTypeBlob, ve.NewBlobFiles[i].Physical.FileNum),
 		})
 	}
 	for i := range ve.NewTables {
@@ -2706,7 +2706,7 @@ func (d *DB) cleanupVersionEdit(ve *manifest.VersionEdit) {
 				FileNum:  of.DiskFileNum,
 				FileSize: of.Size,
 			},
-			isLocal: objstorage.IsLocalTable(d.objProvider, of.DiskFileNum),
+			placement: objstorage.Placement(d.objProvider, base.FileTypeTable, of.DiskFileNum),
 		})
 	}
 	d.mu.versions.addObsoleteLocked(obsoleteFiles)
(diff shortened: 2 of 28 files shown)


jbowens (approved):
:lgtm_strong: 

> @jbowens reviewed 28 of 28 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5504#-:-OclfMB57TqeVz8sxkMn:b-uwkowd)* status: all files reviewed, 6 unresolved discussions (waiting on @RaduBerinde)

___
*[`metrics.go` line 376 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5504#-OclavDv38TzRWITPVXI:-OclavDv38TzRWITPVXJ:buxj3lf) ([raw file](https://github.com/cockroachdb/pebble/blob/e3e228fd04aa1323d6d8a72dc5f9968207d84c92/metrics.go#L376)):*
> ```Go
> 		// ValueSize is the sum of the length of the uncompressed values in all live
> 		// (referenced by some sstable(s) within the current version) blob files.
> 		// ValueSize may be greater than ReferencedValueSize when compression is
> ```

I think this comment should say "ValueSize may be greater than Live.Total().Size when compression is effective."
___
*[`metrics/by_placement.go` line 77 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5504#-OclbeacA1NxLl3I7IqB:-OclbeacA1NxLl3I7IqC:b-897yru) ([raw file](https://github.com/cockroachdb/pebble/blob/e3e228fd04aa1323d6d8a72dc5f9968207d84c92/metrics/by_placement.go#L77)):*
> ```Go
> }
> 
> func (bp *ByPlacement[T]) Set(placement base.Placement, value T) {
> ```

nit: worth using `Ptr` to implement it?

```
{
    *bp.Ptr(placement) = value
}
```

___
*[`metrics/by_placement.go` line 187 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5504#-OclcqNTEB9xkQeBaXgY:-OclcqNTEB9xkQeBaXgZ:b-p314c5) ([raw file](https://github.com/cockroachdb/pebble/blob/e3e228fd04aa1323d6d8a72dc5f9968207d84c92/metrics/by_placement.go#L187)):*
> ```Go
> 		if !first {
> 			w.Printf("; ")
> 			first = false
> ```

this setting of first should be outside the conditional or in an else branch, right?
___
*[`metrics/by_placement.go` line 194 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5504#-Oclcz-AAWhxIuFWMGfE:-Oclcz-AAWhxIuFWMGfF:b-7ropln) ([raw file](https://github.com/cockroachdb/pebble/blob/e3e228fd04aa1323d6d8a72dc5f9968207d84c92/metrics/by_placement.go#L194)):*
> ```Go
> 		if !first {
> 			w.Printf("; ")
> 			first = false
> ```

ditto
___
*[`internal/base/placement.go` line 15 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5504#-Ocl_KTu0CnRtjqr_aHK:-Ocl_KTu0CnRtjqr_aHL:b-4far1g) ([raw file](https://github.com/cockroachdb/pebble/blob/e3e228fd04aa1323d6d8a72dc5f9968207d84c92/internal/base/placement.go#L15)):*
> ```Go
> // Placement identifies where a file/object is stored.
> //
> // The zero value is invalid (this is intentional to determine accidantelly
> ```

nit: "accidentally" (and maybe "to disambiguate accidentally uninitialized fields.")
___
*[`internal/base/placement.go` line 17 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5504#-Ocl_yVrDMMIvx64F_Pd:-Ocl_yVrDMMIvx64F_Pe:bsqerfe) ([raw file](https://github.com/cockroachdb/pebble/blob/e3e228fd04aa1323d6d8a72dc5f9968207d84c92/internal/base/placement.go#L17)):*
> ```Go
> // The zero value is invalid (this is intentional to determine accidantelly
> // uninitialized fields).
> type Placement uint8
> ```

love this new enum

Issue #5501, opened 2025-10-28
db: RangeDeletionBytesEstimate excludes separated values


jbowens:
null

Jira issue: PEBBLE-1239

PR #5507, opened 2025-10-28
crl-release-25.4: db: fix RangeDeletionBytesEstimate

jbowens:
Fix RangeDeletionBytesEstimate to incorporate an estimate of the referenced
blob values.

Fix https://github.com/cockroachdb/pebble/issues/5501.


jbowens:
TFTR!

PR #5515, opened 2025-10-29
go.mod: require Go 1.25

jbowens:
Require Go 1.25, mirroring Cockroach's master branch, and allowing us to make use of new 1.24 and 1.25 features such as the synctest package.

### DIFF
--- a/.github/workflows/nightlies.yaml
+++ b/.github/workflows/nightlies.yaml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go: [ '1.23', '1.25' ]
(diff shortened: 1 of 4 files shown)


RaduBerinde:
I think we should wait until there's high confidence we'll stay with 1.25 (last time we had to walk back). CC @rickystewart

xinhaoz (approved):

rickystewart:
We're going to have lots of problems if this Go upgrade is walked back, having to revert this PR being the least of them. I wouldn't be concerned about this.

RaduBerinde (approved):

jbowens:
TFTRs!

PR #5517, opened 2025-10-29
crl-release-25.4: db: tolerate unknown options

annrpom:
Instead of throwing an error when we encounter unfamiliar options, we will instead tolerate it and log a message. This will allow us to be more lenient about options handling (backporting, renaming, etc.). It also addresses the case in which a user could revert before a major upgrade finalization and have some new option written that the lesser version doesn't recognize and error out at.


annrpom:
TFTR! ('-')7

PR #5518, opened 2025-10-29
db, valsep: make span ValueStoragePolicy a struct defining the min size

xinhaoz:
ValueStoragePolicy in SpanPolicy is now a struct defining the policy adjustment type (defaultPolicy, noValSeparation, override) and the override value for the minimum size used for value separation. Compactions now use the ValueStoragePolicy returned by the span policy for a table's key range to decide whether we can preserve blob references or not.

Previously, we would rewrite blob references if a compaction contained tables with mixed value storage policies, even if those policies were unchanged.

We now preserve blob references if:
- The maximum blob reference depth is not exceeded.
- For each input table, the value separation policy used when writing the table matches the current value separation policy applied to the table span.

### db, valsep: do not separate mvcc values if DisableSeparationBySuffix is true 

Move this field into the ValueStoragePolicy struct and make sure it's being read
at the time of value separation.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -3438,7 +3438,7 @@ func (d *DB) compactAndWrite(
 			spanPolicyValid = true
 		}
 		writerOpts := d.makeWriterOptions(c.outputLevel.level)
-		if spanPolicy.DisableValueSeparationBySuffix {
+		if spanPolicy.ValueStoragePolicy.DisableSeparationBySuffix {
 			writerOpts.DisableValueBlocks = true
 		}
 		if spanPolicy.PreferFastCompression && writerOpts.Compression != block.NoCompression {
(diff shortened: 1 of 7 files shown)


jbowens (approved):
:lgtm_strong: 

> @jbowens reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5518#-:-OcuyLPJ5ZKalOfslE6B:b-uwkowd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sumeerbhola)

xinhaoz:
@jbowens I added another commit after your review to address the TODO on using the `DisableValueSeparationBySuffix` flag if you wouldn't mind taking another look 🙏

xinhaoz:
TFTRs! Followed up about the min value size for latency tolerant spans, and the value was chosen arbitrarily 🚀 . We can bump it up when we make the changes to make the value configurable. I think @annrpom is already on it.

PR #5521, opened 2025-11-02
db: narrow down versionSet.metrics

RaduBerinde:
We now store in `versionSet` only the subset of metrics that are
actually maintained by the version set, reducing confusion.


RaduBerinde:
TFTR!

PR #5524, opened 2025-11-02
deletepacer: don't use fifo.Queue

### LINKED ISSUE #5498: master: nightly linux-msan failed
The nightly linux-msan test run failed on 4cb1291805fb54e025da6408ff7d3dcb1bfcb8e8 (go1.23). Please review the run [here](https://github.com/cockroachdb/pebble/actions/runs/18816397870).

Jira issue: PEBBLE-1237

RaduBerinde:
There is a msan failure that goes away if we switch to using a slice
here. This isn't due to some bug in `fifo.Queue` but it looks like a
possible compiler bug; the problem will continue to be investigated as
part of https://github.com/golang/go/issues/76138

We don't really need the specialized `fifo.Queue` here, a slice is
just as good; we just make sure to allocate in reasonable chunks.

Fixes #5498

### DIFF
--- a/internal/deletepacer/delete_pacer.go
+++ b/internal/deletepacer/delete_pacer.go
@@ -7,11 +7,11 @@ package deletepacer
 import (
 	"context"
 	"runtime/pprof"
+	"slices"
 	"sync"
 	"time"
 
 	"github.com/cockroachdb/crlib/crtime"
-	"github.com/cockroachdb/crlib/fifo"
 	"github.com/cockroachdb/pebble/internal/base"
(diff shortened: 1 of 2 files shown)


annrpom (approved):
:lgtm:

> @annrpom reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5524#-:-Od9Zf5L3_Ylgy7M6EOn:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

RaduBerinde:
TFTR!

Issue #5444, opened 2025-10-17
db: add metric encapsulating blob file rewrite bytes read, written


jbowens:
Blob file rewrite compactions happen independent of the LSM, and so don't contribute to the per-level 'compacted bytes' total. We should have separate metrics recording the bytes read and bytes written during these compactions, and we should incorporate them into the CockroachDB compacted bytes timeseries metrics.

Jira issue: PEBBLE-1222

PR #5525, opened 2025-11-03
db: add metric encapsulating blob file rewrite bytes read, written

annrpom:
Blob file rewrite compactions happen independent of LSM levels, and so don't
contribute to the per-level 'compacted bytes' total. This patch adds separate
metrics recording the bytes read and bytes written during these compactions.

Fixes: #5444


annrpom:
TFTR! ('-')7

PR #5528, opened 2025-11-04
db: pull more recovery logic into recoverState

jbowens:
Additionally, remove the call to wal.Scan in Open, instead relying on the WALs
scanned in recoverState. This was accidentally overlooked during the refactor
that began scanning wals in recoverState.


jbowens:
TFTR!

PR #5529, opened 2025-11-04
sstable: in props, remove ValueSeparationKind, add ValueSeparationByS…

xinhaoz:
…uffixDisabled

Remove ValueSeparationKind, which was originally added for compaction decisions
but was not necessary in the end.

Add the property ValueSeparationBySuffixDisabled, which indicates if the table
used special value separation policies for mvcc garbage to inform value
separation decisions during compactions.


xinhaoz:
tftr!!

PR #5532, opened 2025-11-05
db: disable writing to value blocks if `DisableSeparationBySuffix` is…

xinhaoz:
… true

Make sure this field, which has been moved from SpanPolicy to ValueStoragePolicy, is read for the DisableValueBlocks option in the sst writer.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -3479,16 +3479,11 @@ func (d *DB) compactAndWrite(
 			writerOpts.Compression = block.FastestCompression
 		}
 		vSep := valueSeparation
-		switch spanPolicy.ValueStoragePolicy.PolicyAdjustment {
-		case UseDefaultValueStorage:
-			// No change to value separation.
-		case NoValueSeparation:
+		if spanPolicy.ValueStoragePolicy.DisableBlobSeparation {
 			vSep = valsep.NeverSeparateValues{}
-		case OverrideValueStorage:
-			// This span of keyspace is more tolerant of latency, so set a more
-			// aggressive value separation policy for this output.
+		} else if spanPolicy.ValueStoragePolicy.ContainsOverrides() {
 			vSep.SetNextOutputConfig(valsep.ValueSeparationOutputConfig{
-				MinimumSize:                    spanPolicy.ValueStoragePolicy.MinimumSize,
+				MinimumSize:                    spanPolicy.ValueStoragePolicy.OverrideBlobSeparationMinimumSize,
 				DisableValueSeparationBySuffix: spanPolicy.ValueStoragePolicy.DisableSeparationBySuffix,
 			})
 		}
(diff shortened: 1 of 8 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5532)* status: 0 of 7 files reviewed, 3 unresolved discussions

___
*[`options.go` line 1312 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5532#-OdKiIVa5Lr7whySuju8:-OdKiIVa5Lr7whySuju9:b8jwdpe) ([raw file](https://github.com/cockroachdb/pebble/blob/10c0e65c2f3fcfc806d9f0a414e6d00b6bf323da/options.go#L1312)):*
> ```Go
> // KVs, overriding global policies. Values can be configured to be stored in-place,
> // in value blocks, or in blob files.
> type ValueStoragePolicyAdjustment struct {
> ```

I think we can simplify this to just a set of fields.

```
type ValueStoragePolicyAdjustment struct {
   DisableBlobSeparation bool
   OverrideBlobSeparationMinimumSize int
   DisableSeparationBySuffix bool
}
```
Zero value means no adjustment.
___
*[`options.go` line 1320 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5532#-OdKifPa5N5-wxMfo2tI:-OdKifPa5N5-wxMfo2tJ:b7wu6lo) ([raw file](https://github.com/cockroachdb/pebble/blob/10c0e65c2f3fcfc806d9f0a414e6d00b6bf323da/options.go#L1320)):*
> ```Go
> 	// keys (where the smallest suffix is the latest version), but should be
> 	// disabled for keys where the suffix does not correspond to a version.
> 	DisableSeparationBySuffix bool
> ```

It would be good to document the of cases of where values can go, with or without this flag, for the three possibilities:
 - most recent suffix with value >= MinimumSize
 - older suffix with value >= MinimumSize
 - older suffix with value < MinimumSize
___
*[`compaction.go` line 3491 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5532#-OdKhOHn3DiAkqEvDDMy:-OdKhOHn3DiAkqEvDDMz:berzzz0) ([raw file](https://github.com/cockroachdb/pebble/blob/10c0e65c2f3fcfc806d9f0a414e6d00b6bf323da/compaction.go#L3491)):*
> ```Go
> 			// aggressive value separation policy for this output.
> 			vSep.SetNextOutputConfig(valsep.ValueSeparationOutputConfig{
> 				MinimumSize:                    spanPolicy.ValueStoragePolicy.MinimumSize,
> ```

[nit] maybe we should rename `MinimumSizeForBlob` or similar

xinhaoz:
TFTR!

PR #5534, opened 2025-11-05
internal/manifest: document relationship between LastSeqNum and WAL seqnums

jbowens:
Clarify commentary around logSeqNum/LastSeqNum. Unflushed WALs may contain sequence numbers less than the most recently persisted LastSeqNum.


annrpom (approved):

jbowens:
TFTR!

PR #5536, opened 2025-11-06
valsep: create SSTBlobWriter to write external sst with blob files

xinhaoz:
SSTBlobWriter is an SST writer that can be configured to separate values into
new blob files. On close, the writer can return metadata about the written sst
and its blob files.

### DIFF
--- a/internal/base/internal.go
+++ b/internal/base/internal.go
@@ -228,6 +228,11 @@ func (k InternalKeyKind) SafeFormat(w redact.SafePrinter, _ rune) {
 	w.Print(redact.SafeString(k.String()))
 }
 
+// IsSet returns true if the key kind is a SET type.
+func (k InternalKeyKind) IsSet() bool {
+	return k == InternalKeyKindSet || k == InternalKeyKindSetWithDelete
+}
+
 // InternalKeyTrailer encodes a SeqNum and an InternalKeyKind.
 type InternalKeyTrailer uint64
 
--- a/internal/compact/run.go
+++ b/internal/compact/run.go
@@ -258,12 +258,8 @@ func (r *Runner) writeKeysToTable(
 	equalPrev := func(k []byte) bool {
 		return tw.ComparePrev(k) == 0
 	}
-	prefixEqual := func(k []byte) bool {
-		return tw.IsPrefixEqualPrev(k)
-	}
 	var pinnedKeySize, pinnedValueSize, pinnedCount uint64
 	var iteratedKeys uint64
-	var prevKeyKind base.InternalKeyKind
 	kv := r.kv
 	for ; kv != nil; kv = r.iter.Next() {
 		iteratedKeys++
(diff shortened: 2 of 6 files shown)


xinhaoz:
Still working on the ingest side changes but going to put this up this now for feedback. This will be used in some tests after the ingest changes are done.

jbowens (approved):
:lgtm:

> @jbowens reviewed 5 of 5 files at r1, 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5536#-:-OerCZDR0a9wU9d80Dde:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @xinhaoz)

___
*[`valsep/sst_blob_writer_test.go` line 38 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5536#-OeqrWVe8LGxbkC8b6Do:-OeqrWVe8LGxbkC8b6Dp:bwa6qo7) ([raw file](https://github.com/cockroachdb/pebble/blob/0b11403e44628f02991b78906011b98a832af6a0/valsep/sst_blob_writer_test.go#L38)):*
> ```Go
> 	var policy base.ValueStoragePolicyAdjustment
> 	var err error
> 	for _, part := range spanPolicyParts {
> ```

nit: now that we're on Go 1.25, you could use `SplitSeq` like:
```
for part := range strings.SplitSeq(spanPolicyStr, ",") {
```
___
*[`valsep/sst_blob_writer.go` line 23 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5536#-OerC6Fr-dfdOO3Zd308:-OerC6Fr-dfdOO3Zd309:b-4retlg) ([raw file](https://github.com/cockroachdb/pebble/blob/0b11403e44628f02991b78906011b98a832af6a0/valsep/sst_blob_writer.go#L23)):*
> ```Go
> type SSTBlobWriter struct {
> 	// Promote methods from sstable.Writer.
> 	*sstable.Writer
> ```

how many of the underlying Writer's methods will we actually need to access from Cockroach? I'm wondering if we should write explicit pass through methods and make it unambiguous which will separate values and which will not. Specifically I'm worried about a caller using `Writer.Add` thinking that it'll separate values, but it won't.
___
*[`valsep/sst_blob_writer.go` line 145 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5536#-OeqmYoJ4yMTyuaottSg:-OeqmYoJ4yMTyuaottSh:b82rhim) ([raw file](https://github.com/cockroachdb/pebble/blob/0b11403e44628f02991b78906011b98a832af6a0/valsep/sst_blob_writer.go#L145)):*
> ```Go
> 	w.kvScratch.K = base.MakeInternalKey(key, 0, sstable.InternalKeyKindSet)
> 	w.kvScratch.V = base.MakeInPlaceValue(value)
> 	isLikelyMVCCGarbage := sstable.IsLikelyMVCCGarbage(
> ```

I think maybe we should move IsLikelyMVCCGarbage onto a method on the RawWriter rather than exporting `PrevPointKeyKind` from the `RawWriter`.

```
IsLikelyMVCCGarbage(nextKey []byte, keyKind base.InternalKeyKind)
```

I think we might be able to unexport `IsPrefixEqualPrev` too then
___
*[`valsep/sst_blob_writer.go` line 150 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5536#-Oeqqaed9qlmMfwcGsQ6:-Oeqqaed9qlmMfwcGsQ7:b-jrb6yf) ([raw file](https://github.com/cockroachdb/pebble/blob/0b11403e44628f02991b78906011b98a832af6a0/valsep/sst_blob_writer.go#L150)):*
> ```Go
> }
> 
> // BlobWriterMetas returns an array of blob.FileWriterStats describing the
> ```

nit: s/array/slice/

jbowens (approved):
:lgtm:

> @jbowens reviewed 1 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5536#-:-OerlxmZ90ZoylhK_IKV:bnfp4nl)* status: 1 of 12 files reviewed, 3 unresolved discussions (waiting on @xinhaoz)

PR #5539, opened 2025-11-07
db: add ability to collect a profile of separated value retrievals

### LINKED ISSUE #5535: db: add profile for value retrievals
Retrieving a separated value is more costly, so we try to only separate values that either

a. are rarely retrieved, or
b. are large enough that the additional indirection is insignificant

When we're retrieving more values than anticipated, it's difficult to determine where the retrievals are originating from. We could add a mechanism to collect a custom pprof of value retrieval stack traces. We would need to build the profile ourselves using:

https://pkg.go.dev/github.com/google/pprof/profile#Sample

jbowens:
With the introduction of value separation, retrievals of separated values are more expensive than retrievals of inlined values. Pebble exports an aggregate metric recording the total count of value retrievals that must load a value from an external blob file. This can be used to help attribute performance issues to excessive separated value retrievals, but it doesn't help with understanding what codepaths are performing the retrievals.

This commit adds support for collection of a 'separated value retrieval' profile, aggregating the count and byte-total of value retrievals per unique stack trace while profile collection is active. This additional instrumentation adds some overhead and is too expensive to perform by default.

Users may call DB.RecordSeparatedValueRetrievals to initiate collection of a profile. All subsequent opened iterators and compactions will record retrievals of separated values into the profile until the caller stops the collection with the invocation of a callback.

For now this profile can only be dumped to a text format through a String method. We could export it to a pprof file but it would require a fair bit of work.

Close #5535.
Informs cockroachdb/cockroach#144715.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -3390,6 +3390,7 @@ func (d *DB) compactAndWrite(
 			uint64(uintptr(unsafe.Pointer(c))),
 			categoryCompaction,
 		),
+		ValueRetrievalProfile: d.valueRetrievalProfile.Load(),
 	}
 	if c.version != nil {
 		c.iterationState.valueFetcher.Init(&c.version.BlobFiles, d.fileCache, blockReadEnv, suggestedCacheReaders)
--- a/db.go
+++ b/db.go
@@ -20,6 +20,7 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/cockroachdb/pebble/internal/arenaskl"
 	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/cockroachdb/pebble/internal/bytesprofile"
 	"github.com/cockroachdb/pebble/internal/cache"
 	"github.com/cockroachdb/pebble/internal/deletepacer"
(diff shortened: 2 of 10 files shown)


RaduBerinde:
I asked Claude-Code Van Damme if it would be easier to use custom pprof profiles and it had only great things to say about your change :)
```
  Using the pprof protobuf package would require:
  - Adding a new dependency
  - Manually managing Function and Location objects
  - Implementing deduplication logic
  - Maintaining ID mappings

  Here's what the refactored Record() would look like:
  [...]
  This is significantly more complex than the current ~9 lines.

  Recommendation

  The current implementation is actually well-designed for its purpose:
  - Simple and efficient
  - No external dependencies
  - Clear aggregation logic
  - Compact (91 lines total)
```

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5539#-:-OdVPAhbBTVH58hLnuLg:bnfp4nl)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @xinhaoz)

PR #5540, opened 2025-11-07
crl-release-25.2: db: fix Options.Clone()

annrpom:
`Clone()` was allowing a shallow copy of `*WALFailover`. This caused a lot of time debugging a test which unwittingly aliased the WAL failover FS and dir of two stores.

This commit fixes this and adds a test that would have caught it.

We should probably make `WALFailover` a non-pointer (with an `Enabled` field); even if we do, the new test is still valuable.

Fixes: #5538

### DIFF
--- a/external_test.go
+++ b/external_test.go
@@ -7,11 +7,15 @@ package pebble_test
 import (
 	"bytes"
 	"io"
+	randv1 "math/rand"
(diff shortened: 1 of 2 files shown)


annrpom:
np; TFTR! ('-')7

PR #5541, opened 2025-11-08
sstable: clean up lazy loading test

RaduBerinde:
Don't use unnecessary globals, and use `runIterCmd` instead of
reimplementing a good chunk of it.


RaduBerinde:
TFTR!

PR #5546, opened 2025-11-10
tests: use colon separator for KVs

RaduBerinde:
The vast majority of tests that specify a KV use the `key:value`
format. The parsing code also supports using a `=` separator and a
handful of tests use that format.

For the sake of consistency, remove this alternate format and fix up
the relevant tests.


RaduBerinde:
TFTR!

PR #5547, opened 2025-11-10
bytesprofile: add Profile.Collect

jbowens:
Add a method to bytesprofile.Profile to return the collected statistics in a structured format. This can be used to return results from a SQL built-in in CockroachDB.


jbowens:
TFTR!

PR #5548, opened 2025-11-10
*: update tests to use the newer internal key format

RaduBerinde:
When parsing internal keys, we switched to a new format (`k@2,SET`) but many tests are still using the old format (`k.SET.2`).

This PR updates all tests to use the new format and removes parsing support for the old format. Some test code is also cleaned-up to reuse functionality.

#### sstable: clean up tests to use new key format

Clean up the sstable tests to use the newer internal key format.
Update some test code to increase reuse.

#### internal: clean up tests to use the new key format

Also improve some of the update code to use existing routines.


RaduBerinde:
TFTR!  Claude was a lot of help :)

PR #5550, opened 2025-11-10
db: use (*testing.T).Chdir

jbowens:
Update a couple tests that chdir to use (*testing.T).Chdir (new in Go 1.24) to change directories with the guarantee that the Go test runner will change the directory back to the original after the test completes.


jbowens:
TFTR!

PR #5552, opened 2025-11-10
*: use (*sync.WaitGroup).Go

jbowens:
Update uses of WaitGroup to use the new (Go 1.25+) Go method to prevent accidental misuse and simplify call sites.

### DIFF
--- a/cmd/pebble/scan.go
+++ b/cmd/pebble/scan.go
@@ -90,11 +90,8 @@ func runScan(cmd *cobra.Command, args []string) {
 
 			limiter := maxOpsPerSec.newRateLimiter()
 
-			wg.Add(concurrency)
-			for i := 0; i < concurrency; i++ {
-				go func(i int) {
-					defer wg.Done()
-
+			for i := range concurrency {
+				wg.Go(func() {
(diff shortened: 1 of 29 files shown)


cmd/pebble/scan.go
```diff
@@ -90,11 +90,9 @@ func runScan(cmd *cobra.Command, args []string) {
 
 			limiter := maxOpsPerSec.newRateLimiter()
 
-			wg.Add(concurrency)
-			for i := 0; i < concurrency; i++ {
-				go func(i int) {
-					defer wg.Done()
-
+			for i := range concurrency {
+				i := i
```
RaduBerinde: [nit] I think this is no longer necessary

RaduBerinde (approved):
Nice!

jbowens:
TFTR!

RaduBerinde:
An annoying side-effect:
```
                created by sync.(*WaitGroup).Go in goroutine 434
                  /Users/radu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.3.darwin-arm64/src/sync/waitgroup.go:237 +0x70
```

Filed https://github.com/golang/go/issues/77173

PR #5553, opened 2025-11-11
treesteps: disable in non-invariants mode, assorted improvements

RaduBerinde:
#### treesteps: disable in non-invariants mode

Make the treesteps code only active in invariants mode, and add
minimal stubs for non-invariants builds. This will allow using
treesteps for iterators which are performance-sensitive.

#### treesteps: UpdateLastOpf

Add a convenience function that automatically uses the last op, to
avoid having to plumb the operation through layers of functions.

#### treeprinter: support indented lines under nodes

This change adds the vertical connector when a multi-line node leaves
spaces for it.

#### treeprinter: add DotDotDot

Add a feature to add a vertical ellipsis indicating that a node has
more hidden children.

#### treesteps: improved depth limiting

We improve the way operations are shown when we limit the operation
depth: at the last depth, we show just the result (the last step) of
that operation.

We also use a better representation for the properties so they don't
look like children, and use a vertical ellipsis to indicate hidden
children.

#### treesteps: allow passthrough

This change improves tresteps to allow eliding nodes like
`invalidating.Iter` or `keyspan.assertIter` from the hierarchy.

If we simply pass through the `NodeInfo` from the wrapped node, we
still register the elided node instead of the wrapped node (and thus
the wrapped node ops are not recorded). To fix this, we the `Node` to
`NodeInfo`.

Note that passthrough happens implicitly when a node embeds another
node and relies on the embedded `TreeSpepsNode()` implementation (like
`colblk.KeyspanIter` embeds `colblk.keyspanIter`).

### DIFF
--- a/internal/treesteps/testdata/passthrough
+++ b/internal/treesteps/testdata/passthrough
@@ -1,0 +1,18 @@
+# Here B is wrapped inside an invisible node; verify that we see the B op.
+run
+----
+step 1/5:
+  A
+   └── B
+step 2/5:
+  A  ← Op
+   └── B
+step 3/5:
+  A  ← Op
+   └── B  ← Op
+step 4/5:
+  A  ← Op
+   └── B  ← Op done
+step 5/5:
+  A  ← Op done
+   └── B
--- a/internal/treesteps/tree_steps_off.go
+++ b/internal/treesteps/tree_steps_off.go
@@ -23,7 +23,7 @@ type Node interface {
 }
 type NodeInfo struct{}
 
-func NodeInfof(format string, args ...any) NodeInfo                  { return NodeInfo{} }
+func NodeInfof(node Node, format string, args ...any) NodeInfo       { return NodeInfo{} }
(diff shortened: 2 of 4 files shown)


RaduBerinde (commented):
TFTR! I added one more commit ("allow passthrough"), PTAL

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5553#-:-Odz5fuREWLCINFb1gFN:b3a3uay)* status: 1 of 10 files reviewed, 1 unresolved discussion (waiting on @xinhaoz)

___
*[`internal/treesteps/doc.go` line 1 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5553#-Odyu7zO7qU6_8BWxnWu:-Odz5lg7DHPdgX-p5YeR:bt2vr1b) ([raw file](https://github.com/cockroachdb/pebble/blob/900399f8ae57978bb9e4388f9f47619f93aa6f2c/internal/treesteps/doc.go#L1)):*

2025
</blockquote></details>

Fixed.

PR #5555, opened 2025-11-12
db: use synctest in TestFlushDelay[Stress]

jbowens:
Use the new synctest package to marginally speed up TestFlushDelay and TestFlushDelayStress through the fast-forwarding of time while waiting for a flush to trigger.


jbowens:
TFTR!

PR #5588, opened 2025-11-14
tool: update compaction summary tool

### LINKED ISSUE #5514: tool: audit compaction summary tool
Seems like this tool is out-of-date as of at least v24.3 - we should fill in the gaps for missing support of other compaction types and potentially have some pebble test that collects logs and runs the summary tool.'

https://github.com/cockroachdb/pebble/issues/5502#issuecomment-3463006358

Jira issue: PEBBLE-1243

xxmplus:
add support for copy, tombstone-density, and rewrite compaction.

Note: ingested-flushable is already supported. Unlike other compaction types, ingested-flushable uses the flush logging format `flushed N ingested flushables` rather than the compaction format, and is handled by `parseIngestDuringFlush()` [(link)](https://github.com/cockroachdb/pebble/blob/master/tool/logs/compaction.go#L1290). It is treated as a flush event per [version_set.go:739](https://github.com/cockroachdb/pebble/blob/master/version_set.go#L739), not as a compaction type.

Fixes #5514.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -154,6 +154,9 @@ const (
 	compactionKindRewrite
 	compactionKindIngestedFlushable
 	compactionKindBlobFileRewrite
+	// compactionKindVirtualRewrite must be the last compactionKind.
+	// If a new kind has to be added after VirtualRewrite,
+	// update AllCompactionKindStrings() accordingly.
 	compactionKindVirtualRewrite
 )
 
@@ -194,6 +197,20 @@ func (k compactionKind) compactingOrFlushing() string {
 		return "flushing"
 	}
 	return "compacting"
+}
+
+// AllCompactionKindStrings returns all compaction kind string representations
+// for testing purposes. Used by tool/logs/compaction_test.go to verify the
+// compaction summary tool stays in sync with new compaction types.
+//
+// NOTE: This function iterates up to compactionKindVirtualRewrite. If a new
+// compactionKind is added after VirtualRewrite, update this function accordingly.
+func AllCompactionKindStrings() map[string]bool {
+	kinds := make(map[string]bool)
+	for k := compactionKindDefault; k <= compactionKindVirtualRewrite; k++ {
+		kinds[k.String()] = true
+	}
+	return kinds
 }
 
 type compaction interface {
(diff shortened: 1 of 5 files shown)


annrpom:
sorry for the long comment - was trying to create a gist instead, but kept getting errors 🤔

xxmplus:
> what do you think about adding something like the following so that we get some sort of warning whenever this tool is out of sync in the future (AI generated) :
> 

Sure, but I'd take a slightly different approach. If the goal is to be future proof, we will need a way to detect out-of-sync compactions without changing the test. Let me draft it.

annrpom (approved):
oops yeah sounds good; feel free to do it in a separate PR
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5588#-:-OeN8BNkDMd0Ovd0_mIR:b-3egk9l)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jbowens)

xxmplus:
> oops yeah sounds good; feel free to do it in a separate PR :lgtm:
> 
> > _[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5588#-:-OeN8BNkDMd0Ovd0_mIR:b-3egk9l)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jbowens)

lol, TFTR. I meant I'd update the PR soon.

xxmplus:
> > _[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5588)_ status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @jbowens)
> 
> _[`compaction.go` line 160 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5588#-OeS6F9n8JStCiu_VgxJ:-OeS6F9n8JStCiu_VgxK:bkyzakb) ([raw file](https://github.com/cockroachdb/pebble/blob/e7174d6bced7fe2953a1e26f22188d57a90970b7/compaction.go#L160)):_
> 
> > ```go
> > 	// If a new kind has to be added after VirtualRewrite,
> > 	// update AllCompactionKindStrings() accordingly.
> > 	compactionKindVirtualRewrite
> > ```
> 
> what do you think about adding a sentinel compactionKindMax instead? and then have a comment that enforces all new compaction kinds should get added above the compactionKindMax

I debated with myself about that idea and ended up against it. Here is a similar proposal in golang code itself, they declined the pattern. https://github.com/golang/go/issues/38831

xxmplus:
Chatted with @annrpom offline. We are ok with the current implementation. Merging the code. Thanks for the review.

PR #5594, opened 2025-11-17
bytesprofile: bump stack depth

jbowens:
Bump the maximum stack depth sampled from 20 frames to 30. We've observed that we could use more depth in Cockroach stack traces, as we're unable to see where a batch evaluation originated even if it's run on the same node.


annrpom (approved):

jbowens:
TFTR!

PR #5595, opened 2025-11-17
metamorphic: fix options ranges

xinhaoz:
Some bitshifts were off by 1 due to using rand.IntN, which returns a half open interval. Fix options ranges to match the one described by comments.


xinhaoz:
Addressed nits, TFTR!

PR #5597, opened 2025-11-18
objstorage: support cold tier

RaduBerinde:
#### objstorageprovider: minor local subsystem cleanup

Minor cleanup of the local/vfs subsystem.

#### objstorage: local settings cleanup

Cleanup for local settings.

#### objstorageprovider: improve "open" test directive

Make the creator-id an optional key-value.

#### objstorage: support cold tier

This change adds objstorage provider support for a cold tier.

`CreateOptions.UseColdTier` can be used to create cold objects; the
tier becomes part of the `ObjectMetadata`.

### DIFF
--- a/internal/base/internal.go
+++ b/internal/base/internal.go
@@ -767,5 +767,6 @@ type StorageTier uint8
 
 const (
 	HotTier StorageTier = iota
+	ColdTier
 	NumStorageTiers
 )
--- a/objstorage/objstorage.go
+++ b/objstorage/objstorage.go
@@ -120,6 +120,11 @@ type ObjectMetadata struct {
 	DiskFileNum base.DiskFileNum
 	FileType    base.FileType
 
+	// The fields below are only set if the object is on local storage.
+	Local struct {
+		Tier base.StorageTier
+	}
+
 	// The fields below are only set if the object is on remote storage.
 	Remote struct {
 		// CreatorID identifies the DB instance that originally created the object.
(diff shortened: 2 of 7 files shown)


RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5597#-:-OeYx4wUAPDAPr0tYD8V:b-814g8k)* status: 15 of 22 files reviewed, 2 unresolved discussions (waiting on @jbowens)

___
*[`objstorage/objstorage.go` line 254 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5597#-OeYPCXy-BOTqNXCracM:-OeYwDDN1psihbYstyUi:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/4f64f74f2978be617fc74ffcbb855b6607dda476/objstorage/objstorage.go#L254)):*

any reason to use a bool here instead of `base.StorageTier`? I understand the `PreferSharedStorage` takes precedence, but maybe we could define the tiers as a property of the object (independent from the question of where it ultimately is stored physically)?
</blockquote></details>

Done.
___
*[`objstorage/objstorageprovider/provider.go` line 412 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5597#-OeYP3RV2i70Z-Rd_GeZ:-OeYwuWGBf4fmuGPyKGH:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/4f64f74f2978be617fc74ffcbb855b6607dda476/objstorage/objstorageprovider/provider.go#L412)):*

should this use `opts.Tier` (/ be conditional on `opts.UseColdTier` as long as that's a bool)?

should this entire branch only be if `!opts.UseColdTier || p.Local.ColdTier.FS == srcFS`?
</blockquote></details>

Done.

RaduBerinde:
+O-AI-Review-Real-Issue-Found

PR #5598, opened 2025-11-18
Revert "db: add MinimumLatencyTolerantSize to ValueSeparationPolicy s…

annrpom:
…truct"

This reverts commit d5052bfb553e0cb39eba123773a0f720988e7ce0.

We can actually just make this configurable on the cockroach side via a cluster setting/through the span policy.


annrpom:
TFTR! ('-')7

PR #5602, opened 2025-11-20
cmd/pebble: ensure minimum mvcc is set for newPebbleDB

annrpom:
This patch ensures that pebble commands that ultimately call `newPebbleDB` get their MinimumMCCGarbageSize set.


annrpom:
TFTR! ('-')7

PR #5605, opened 2025-11-21
github: remove labeled type from pr bot review triggering condition

xxmplus:
The `labeled` condition was required when the triggering is off by default so that we can trigger the review by hand. Now that we are in opt-out mode, this condition should be removed to avoid unnecessary runs.


xxmplus:
TFTR!

PR #5606, opened 2025-11-21
github: add nightly crossversion smoke test

### LINKED ISSUE #4844: crossversion: add a smoke test
This issue tracks adding a special "smoke test' for the crossversion test iself, something that would have caught #4732.

It could be a nightly script that introduces an upgrade bug on the current branch and verifies that the crossversion test fails.

Jira issue: PEBBLE-475

Epic PEBBLE-1158

RaduBerinde:
Sample run: https://github.com/RaduBerinde/pebble/actions/runs/19582196207/job/56082638208

Fixes #4844

#### scripts: add crossversion smoke test

Add a smoke test that validates the crossversion test infrastructure
can detect backwards compatibility issues. The test works in three phases:

Phase 1: Run crossversion test with clean versions (should pass)
Phase 1.5: Run single-version test with patched version (should pass)
Phase 2: Run crossversion test with patched version (should fail)

The patch introduces an intentional metaindex encoding
incompatibility.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude <noreply@anthropic.com>

#### github: add nightly crossversion smoke test

### DIFF
--- a/.github/workflows/crossversion.yaml
+++ b/.github/workflows/crossversion.yaml
@@ -1,0 +1,62 @@
+name: Crossversion smoke test
+
+on:
+  workflow_call:
+    inputs:
+      sha:
+        description: 'SHA to run against'
+        required: false
+        type: string
+      file_issue_branch:
+        description: 'Branch used in filed issues; if unset, no issues are filed'
+        required: false
+        type: string
+      go_version:
+        description: 'Go version'
+        required: false
+        default: '1.25'
+        type: string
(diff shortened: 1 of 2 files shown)


RaduBerinde:
> I'm a little wary of maintaining the patch file, but I don't have any better ideas

Same. I will try an alternative where I put the code in behind a constant that's controlled by a build flag. I will merge this as-is for now and explore that separately. TFTR!

PR #5611, opened 2025-11-24
db: move file number assignment to just before metadata creation

xinhaoz:
Previously we preallocated all table numbers upfront based on the ingest args. The comment on this being required in order to mark the files as pending seems outdated, so we now assign the file number just before ingesting the sst.


xinhaoz:
tftr!

PR #5614, opened 2025-11-25
db: export SpanPolicy and ValueStoragePolicyAdjustment as public type aliases

xxmplus:
These types were moved to internal/base but are needed by external users like CockroachDB. Add public type aliases to maintain API compatibility.


xxmplus:
TFTR!

PR #5620, opened 2025-12-01
db: allow ingesting local SSTs with values in blob files

xinhaoz:
### db: add LocalSST struct containing metadata for local ingested files

Previously when ingesting local ssts we provided just the string paths as a
slice. Now that we want to ingest tables with blob files, we introduce a new
type that will track local sst paths and extra metadata like their blob file
paths.

### db: allow ingesting local SSTs with values in blob files

Allow ingesting local ssts and their associated blob files. Note that we do not
validate blob value handles. Each blob file is assumed to be fully referenced by
the SST.

### DIFF
--- a/ingest.go
+++ b/ingest.go
@@ -23,7 +23,9 @@ import (
 	"github.com/cockroachdb/pebble/objstorage"
 	"github.com/cockroachdb/pebble/objstorage/remote"
 	"github.com/cockroachdb/pebble/sstable"
+	"github.com/cockroachdb/pebble/sstable/blob"
 	"github.com/cockroachdb/pebble/sstable/block"
+	"github.com/cockroachdb/pebble/sstable/colblk"
 )
 
 func sstableKeyCompare(userCmp Compare, a, b InternalKey) int {
@@ -245,6 +247,13 @@ func (r *rangeKeyIngestValidator) Validate(nextFileSmallestKey *keyspan.Span) er
 	return nil
 }
 
+type ingestLocalResult struct {
+	meta           *manifest.TableMetadata
+	lastRangeKey   keyspan.Span
+	blockReadStats base.BlockReadStats
+	blobMetas      []manifest.BlobFileMetadata
+}
+
 // ingestLoad1 creates the TableMetadata for one file. This file will be owned
 // by this store.
 //
@@ -260,12 +269,10 @@ func ingestLoad1(
 	compressionCounters *block.CompressionCounters,
 	tableNum base.TableNum,
 	rangeKeyValidator rangeKeyIngestValidator,
-) (
-	meta *manifest.TableMetadata,
-	lastRangeKey keyspan.Span,
-	blockReadStats base.BlockReadStats,
-	err error,
-) {
+	blobPaths []string,
+	blobFileNums []base.DiskFileNum,
+) (res ingestLocalResult, err error) {
+	var meta *manifest.TableMetadata
 	o := opts.MakeReaderOptions()
 	o.CacheOpts = sstableinternal.CacheOptions{
 		CacheHandle: cacheHandle,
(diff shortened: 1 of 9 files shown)


jbowens (commented):
> @jbowens reviewed 2 of 2 files at r1, 5 of 5 files at r2, 7 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5620)* status: 10 of 12 files reviewed, 10 unresolved discussions (waiting on @xinhaoz and @xxmplus)

___
*[`ingest_with_blobs.go` line 38 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5620#-Ofj6FrX0ZJoxEmTElG0:-Ofj6FrX0ZJoxEmTElG1:br3p14w) ([raw file](https://github.com/cockroachdb/pebble/blob/40e9e8a2ab08681484f1645ee73492a190eb7888/ingest_with_blobs.go#L38)):*
> ```Go
> // IngestLocal ingests the provided local sstables into the DB.
> // If a valid exciseSpan is provided, the ingested sstables will have any keys
> // in that span excised (removed) during ingestion.
> ```

the sentence describing the `exciseSpan` behavior seems off. if exciseSpan is provided, any committed keys within the DB that fall in that span will be deleted atomically with ingestion. if the sstables being ingested have keys that fall within the excise span, those keys will not be removed
___
*[`ingest.go` line 487 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5620#-OfgnezFAQePzJ4ymp54:-OfgnezFAQePzJ4ymp55:b390d4b) ([raw file](https://github.com/cockroachdb/pebble/blob/40e9e8a2ab08681484f1645ee73492a190eb7888/ingest.go#L487)):*
> ```Go
> }
> 
> // ingestBlobFilesForTable creates table's blob references from the provided blob file paths.
> ```

nit: I think we should explicitly state that this function will update `tableMeta.BlobReferences` appropriately and return the `BlobFileMetadata` for the corresponding referenced files
___
*[`ingest.go` line 490 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5620#-OfghoABBjoNUE_lfi30:-OfghoAC3TAPDT1QzCtw:b-y2ny8d) ([raw file](https://github.com/cockroachdb/pebble/blob/40e9e8a2ab08681484f1645ee73492a190eb7888/ingest.go#L490)):*
> ```Go
> //
> // The following invariants are expected to hold:
> //   - The provided blob file paths are ordered by by their index in the table's blob references
> ```

nit: s/by by/by/
___
*[`ingest.go` line 494 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5620#-OfgnMC20h-JPxj-otF-:-OfgnMC20h-JPxj-otF0:b-ivkdam) ([raw file](https://github.com/cockroachdb/pebble/blob/40e9e8a2ab08681484f1645ee73492a190eb7888/ingest.go#L494)):*
> ```Go
> //     value, so the order must be maintained when mapping to the table's blob references.
> //   - Each blob file is fully referenced by the table.
> func ingestBlobFilesForTable(
> ```

nit: maybe call this something like `constructBlobFileMetadataForIngestedTable`? the existing name was reading to me like it was performing the ingestion itself
___
*[`ingest.go` line 508 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5620#-Ofgi2Wu5-U-m4k8KSch:-Ofgi2Wu5-U-m4k8KSci:b4vifr5) ([raw file](https://github.com/cockroachdb/pebble/blob/40e9e8a2ab08681484f1645ee73492a190eb7888/ingest.go#L508)):*
> ```Go
> 	}
> 
> 	// Read the the blob reference index block to determine the size of values
> ```

nit: s/the the/the/
___
*[`ingest.go` line 517 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5620#-OfglW4w9F3LWfKvn2OA:-OfglW4w9F3LWfKvn2OB:bhmkg05) ([raw file](https://github.com/cockroachdb/pebble/blob/40e9e8a2ab08681484f1645ee73492a190eb7888/ingest.go#L517)):*
> ```Go
> 	defer h.Release()
> 	if !h.Valid() {
> 		return nil, errors.New("pebble: table has invalid blob reference index block")
> ```

I think we should be able to omit this or at least convert it to an invariant assertion. The contract should be that `err==nil` implies `h.Valid()`
___
*[`ingest.go` line 580 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5620#-OfjWLJK4uIxTtw_6Uze:-OfjWLJK4uIxTtw_6Uzf:b-vke3gt) ([raw file](https://github.com/cockroachdb/pebble/blob/40e9e8a2ab08681484f1645ee73492a190eb7888/ingest.go#L580)):*
> ```Go
> 			Size:         max(uint64(fileSize), 1),
> 			ValueSize:    valueSize,
> 			CreationTime: uint64(time.Now().Unix()),
> ```

nit: `uint64(opts.private.timeNow().Unix())`
___
*[`ingest.go` line 849 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5620#-Ofgqbtd2Bxc3AzPRTTc:-Ofgqbtd2Bxc3AzPRTTd:b-d8mwfd) ([raw file](https://github.com/cockroachdb/pebble/blob/40e9e8a2ab08681484f1645ee73492a190eb7888/ingest.go#L849)):*
> ```Go
> 		// Link blob files first.
> 		for blobInd, bf := range localMetas[i].blobFiles {
> 			blobPaths := localMetas[i].local.BlobPaths
> ```

nit: maybe move `blobPaths :=` outside this inner for loop, since it's per-table state
___
*[`ingest.go` line 1699 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5620#-Ofgsf0i-rL1O7MMrPPD:-OfgsR3FC0toT0dFe-70:be3by9s) ([raw file](https://github.com/cockroachdb/pebble/blob/40e9e8a2ab08681484f1645ee73492a190eb7888/ingest.go#L1699)):*
> ```Go
> 	// the tables referenced in the MANIFEST, but not present in the provider.
> 	if err := d.objProvider.Sync(); err != nil {
> 		return IngestOperationStats{}, err
> ```

nit: this was an existing issue before this change, but I'm realizing that this error case and the one above don't `ingestCleanup` before returning. how about leaving a TODO to clean up the error handling in here?
___
*[`ingest_test.go` line 381 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5620#-OfjVW56CHWkejnnfTSj:-OfjVW56CHWkejnnfTSk:basmkd3) ([raw file](https://github.com/cockroachdb/pebble/blob/40e9e8a2ab08681484f1645ee73492a190eb7888/ingest_test.go#L381)):*
> ```Go
> 				for _, val := range arg.Vals {
> 					// Split by comma to handle comma-separated blob paths
> 					paths := strings.Split(val, ",")
> ```

nit: I think you could do

```
blobPaths = append(blobPaths, strings.FieldsFunc(val, func(r rune) bool {
    return r == ',' || unicode.IsSpace(r)
})
```

xinhaoz:
TFTR!

PR #5622, opened 2025-12-02
cockroachkvs: use bytes.Compare when comparing key version

annrpom:
This patch uses bytes.Compare to compare key versions instead of ComparePointSuffixes; the latter calls on
`normalizeEngineSuffixForCompare` - which could end up stripping additional bytes from the `suffix` (comparing only a prefix of the suffix) if the last byte of the version happens to match one of the sentinel lengths defined.

Fixes: [#158463](https://github.com/cockroachdb/cockroach/issues/158463)

Radu's analysis: https://github.com/cockroachdb/cockroach/issues/158463#issuecomment-3603307181

### DIFF
--- a/cockroachkvs/cockroachkvs_test.go
+++ b/cockroachkvs/cockroachkvs_test.go
@@ -30,17 +30,34 @@ import (
 	testify "github.com/stretchr/testify/require"
 )
 
-func TestComparer(t *testing.T) {
+func testPrefixes() [][]byte {
 	prefixes := [][]byte{
 		EncodeMVCCKey(nil, []byte("abc"), 0, 0),
 		EncodeMVCCKey(nil, []byte("d"), 0, 0),
 		EncodeMVCCKey(nil, []byte("ef"), 0, 0),
 	}
+	for range 10 {
+		roachKey := make([]byte, 1+rand.IntN(100))
+		for i := range roachKey {
+			roachKey[i] = byte(rand.Uint32())
+		}
+		prefixes = append(prefixes, EncodeMVCCKey(nil, roachKey, 0, 0))
+	}
+	return prefixes
+}
 
+func testSuffixes(t *testing.T) [][]byte {
 	suffixes := [][]byte{{}}
-	for walltime := 3; walltime > 0; walltime-- {
-		for logical := 2; logical >= 0; logical-- {
-			key := EncodeMVCCKey(nil, []byte("foo"), uint64(walltime), uint32(logical))
+
+	wallTimes := []uint64{1, 2, 3}
+	logicalTimes := []uint32{0, 1, 2}
+	for range 5 {
+		wallTimes = append(wallTimes, rand.Uint64())
+		logicalTimes = append(logicalTimes, rand.Uint32())
+	}
+	for _, walltime := range wallTimes {
+		for _, logical := range logicalTimes {
+			key := EncodeMVCCKey(nil, []byte("foo"), walltime, logical)
 			suffix := key[Comparer.Split(key):]
 			suffixes = append(suffixes, suffix)
 
@@ -51,7 +68,7 @@ func TestComparer(t *testing.T) {
 				if Comparer.CompareRangeSuffixes(suffix, newSuffix) != 1 {
 					t.Fatalf("expected suffixes %x < %x", suffix, newSuffix)
 				}
-				if Comparer.Compare(slices.Concat(prefixes[0], suffix), slices.Concat(prefixes[0], newSuffix)) != 0 {
+				if Comparer.Compare(slices.Concat([]byte("foo"), suffix), slices.Concat([]byte("foo"), newSuffix)) != 0 {
 					t.Fatalf("expected keys with suffixes %x and %x to be equal", suffix, newSuffix)
 				}
 				suffixes = append(suffixes, newSuffix)
@@ -66,15 +83,31 @@ func TestComparer(t *testing.T) {
 			if Comparer.CompareRangeSuffixes(suffix, newSuffix) != 1 {
 				t.Fatalf("expected suffixes %x < %x", suffix, newSuffix)
 			}
-			if Comparer.Compare(slices.Concat(prefixes[0], suffix), slices.Concat(prefixes[0], newSuffix)) != 0 {
+			if Comparer.Compare(slices.Concat([]byte("foo"), suffix), slices.Concat([]byte("foo"), newSuffix)) != 0 {
 				t.Fatalf("expected keys with suffixes %x and %x to be equal", suffix, newSuffix)
 			}
 			suffixes = append(suffixes, newSuffix)
 		}
 	}
 	// Add some lock table suffixes.
+	suffixes = append(suffixes, append(bytes.Repeat([]byte{0}, engineKeyVersionLockTableLen), suffixLenWithLockTable))
 	suffixes = append(suffixes, append(bytes.Repeat([]byte{1}, engineKeyVersionLockTableLen), suffixLenWithLockTable))
 	suffixes = append(suffixes, append(bytes.Repeat([]byte{2}, engineKeyVersionLockTableLen), suffixLenWithLockTable))
+	suffixes = append(suffixes, append(bytes.Repeat([]byte{0xFF}, engineKeyVersionLockTableLen), suffixLenWithLockTable))
+	for range 5 {
+		v := make([]byte, engineKeyVersionLockTableLen+1)
+		for i := 0; i < engineKeyVersionLockTableLen; i++ {
+			v[i] = byte(rand.Uint32())
+		}
+		v[engineKeyVersionLockTableLen] = suffixLenWithLockTable
+		suffixes = append(suffixes, v)
+	}
+	return suffixes
+}
+
+func TestComparer(t *testing.T) {
+	prefixes := testPrefixes()
+	suffixes := testSuffixes(t)
 	if err := base.CheckComparer(&Comparer, prefixes, suffixes); err != nil {
 		t.Error(err)
 	}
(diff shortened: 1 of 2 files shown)


annrpom (commented):
done

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5622#-:-OfXDZKp6S3gRpeJ2ouq:bb7f2mf)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @RaduBerinde)

___
*[`cockroachkvs/cockroachkvs.go` line 792 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5622#-OfVf4OnDbjeSUG3Q_CY:-OfXDbgO1GiHn1D8Qtzy:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/5c6a9c768cbb599d25c5ed22c58a56d004b41aa3/cockroachkvs/cockroachkvs.go#L792)):*

[nit] I'd put the `At()` result in a variable
</blockquote></details>

done
___
*[`cockroachkvs/cockroachkvs.go` line 794 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5622#-OfVfDdCDk1g6Ga5IXkW:-OfXDc4OE6hlMtI51uA_:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/5c6a9c768cbb599d25c5ed22c58a56d004b41aa3/cockroachkvs/cockroachkvs.go#L794)):*

`return len(version) <= len(..)`
</blockquote></details>

done

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5622)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`cockroachkvs/cockroachkvs.go` line 794 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5622#-OfVfDdCDk1g6Ga5IXkW:-OfZtKKN9YnqQ-Xz0d1N:brsjbo3) ([raw file](https://github.com/cockroachdb/pebble/blob/5c6a9c768cbb599d25c5ed22c58a56d004b41aa3/cockroachkvs/cockroachkvs.go#L794)):*

It might be easier to reason about if we did this (your call):
```
// Empty suffixes sort before non-empty suffixes.
switch {
case len(version) == 0:
  // This includes the case where both versions are empty.
  return true
case len(firstRowUntypedVer) == 0:
  return false
default:
  return bytes.Compare()
}
```
</blockquote></details>

ah i like - done
___
*[`cockroachkvs/testdata/key_schema_key_seeker` line 245 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5622#-OfZpxbh5HCXyCGeCCDL:-OfZuoU6AUs8NC034sCF:boj1ceb) ([raw file](https://github.com/cockroachdb/pebble/blob/e119ca2263953ce04969bbf857b5c458bf6a0e36/cockroachkvs/testdata/key_schema_key_seeker#L245)):*

Let's add a case where the first key in the block doesn't have a suffix.
</blockquote></details>

we have one https://github.com/annrpom/pebble/blob/e119ca2263953ce04969bbf857b5c458bf6a0e36/cockroachkvs/testdata/key_schema_key_seeker#L173-L188
___
*[`cockroachkvs/testdata/key_schema_key_seeker` line 270 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5622#-OfZq2Lt3Rz8KrbKKBJ-:-OfZwpVI7ZKvqvW-gcRs:btns4ct) ([raw file](https://github.com/cockroachdb/pebble/blob/e119ca2263953ce04969bbf857b5c458bf6a0e36/cockroachkvs/testdata/key_schema_key_seeker#L270)):*
> also a case with a suffix that is expected to return true.
done

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5622#-:-Of_3_6d30U_AMAovc0I:bnfp4nl)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @annrpom)

___
*[`cockroachkvs/testdata/key_schema_key_seeker` line 245 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5622#-OfZpxbh5HCXyCGeCCDL:-Of_3WG_7PQuAToGeKj4:b-1y0b30) ([raw file](https://github.com/cockroachdb/pebble/blob/e119ca2263953ce04969bbf857b5c458bf6a0e36/cockroachkvs/testdata/key_schema_key_seeker#L245)):*

we have one https://github.com/annrpom/pebble/blob/e119ca2263953ce04969bbf857b5c458bf6a0e36/cockroachkvs/testdata/key_schema_key_seeker#L173-L188
</blockquote></details>

Ah, nice!

RaduBerinde:
*[`cockroachkvs/cockroachkvs.go` line 838 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5622#-Of_eo1H-oG5_PUHHyFX:-Of_eo1H-oG5_PUHHyFY:b-u00a88) ([raw file](https://github.com/cockroachdb/pebble/blob/e19ac24fb16f42291cfafa03c95de8371c17f9b6/cockroachkvs/cockroachkvs.go#L838)):*
> ```Go
> 			firstRowSuffix = buf[:13]
> 		}
> 		return ComparePointSuffixes(firstRowSuffix, version) >= 0
> ```

Claude Code found another issue here: we're calling this with `version` which is not a suffix. We can just replace with `bytes.Compare(version, firstRowSuffix[:len(firstRowSuffix)-1]) >= 0` since neither is empty.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5622)* status: 1 of 2 files reviewed, 5 unresolved discussions (waiting on @annrpom and @sumeerbhola)

___
*[`cockroachkvs/cockroachkvs.go` line 791 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5622#-OfaK41V7NlnDMKyyvdo:-OfalDWM4Vfk1n9zg0nj:b-e6rkkn) ([raw file](https://github.com/cockroachdb/pebble/blob/2aabb29ad29228e234d26d5f2fab5f8732e908b8/cockroachkvs/cockroachkvs.go#L791)):*

Oh, I see https://github.com/cockroachdb/cockroach/issues/158463#issuecomment-3603324718.

And since this is an untyped version, we don't need to normalize it. We should add a comment where `normalizeEngineSuffixForCompare` is declared that it is a noop for a non-MVCC suffix.

But `suffix` could be a MVCC suffix so shouldn't we normalize it, just to be consistent with what we do in `ComparePointSuffixes` which allows comparing a MVCC suffix with a non-MVCC suffix and would normalize the former?
So we would do `normalizedVersion := normalizeEngineSuffixForCompare(suffix)` and use that below?
</blockquote></details>

You're right. I think we should separate out a `normalizeVersionForCompare` and have a `compareVersions` helper.

RaduBerinde:
*[`cockroachkvs/cockroachkvs.go` line 791 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5622#-OfaK41V7NlnDMKyyvdo:-OfaloHX2Ne8Tw9wK7wV:b-oez00h) ([raw file](https://github.com/cockroachdb/pebble/blob/2aabb29ad29228e234d26d5f2fab5f8732e908b8/cockroachkvs/cockroachkvs.go#L791)):*

You're right. I think we should separate out a `normalizeVersionForCompare` and have a `compareVersions` helper.
</blockquote></details>

Feels like all these paths can be tested with a randomized test. We can pair every suffix with every prefix in `TestCompare` to build a set of keys; for each key, build a block with that key, call `IsLowerBound` against every key and cross-check against `Compare`.

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5622)* status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`cockroachkvs/cockroachkvs.go` line 773 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5622#-OfaIfgr8D6k3NfrdNXy:-Ofe4COPCvx1zqOtLsfd:b3pf841) ([raw file](https://github.com/cockroachdb/pebble/blob/2aabb29ad29228e234d26d5f2fab5f8732e908b8/cockroachkvs/cockroachkvs.go#L773)):*

Is this only for point keys?
If for both point and range keys, I'll go read all the stuff about why we call `normalizeEngineSuffixForCompare` only in some cases.
</blockquote></details>

yes only for point keys
___
*[`cockroachkvs/cockroachkvs.go` line 780 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5622#-OfaIvvY8_QPs4oKNH75:-Ofe958A8DZzZnmbG5VD:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/2aabb29ad29228e234d26d5f2fab5f8732e908b8/cockroachkvs/cockroachkvs.go#L780)):*

It would be clearer if we had something like
// Reminder, DecodeEngineKey excludes the length byte at the end, and roachKey excludes the sentinel byte.
// The suffix below would be the result of calling k[Split(k):], now that we know it is a valid engine key.
suffix := k[len(roachKey)+1:]

And we should document what `DecodeEngineKey` returns in its declaration.
</blockquote></details>

done
___
*[`cockroachkvs/cockroachkvs.go` line 784 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5622#-OfaJm0g8SKSMlvqxyQe:-Ofe94iU1EhUBmhxxlIp:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/2aabb29ad29228e234d26d5f2fab5f8732e908b8/cockroachkvs/cockroachkvs.go#L784)):*

and then changed this to use `suffix`.
</blockquote></details>

done
___
*[`cockroachkvs/cockroachkvs.go` line 791 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5622#-OfaK41V7NlnDMKyyvdo:-Ofg8W1JA00ToRk0ps7V:b-7rl0jq) ([raw file](https://github.com/cockroachdb/pebble/blob/2aabb29ad29228e234d26d5f2fab5f8732e908b8/cockroachkvs/cockroachkvs.go#L791)):*

Something like this, feel free to take this commit and put it in this PR: https://github.com/cockroachdb/pebble/compare/master...RaduBerinde:cockroachkvs-lower-bound-test?expand=1
</blockquote></details>

done!

sumeerbhola (approved):
:lgtm: 

> @sumeerbhola reviewed 1 of 2 files at r4, 1 of 1 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5622#-:-OfkjTes9vcTFm4VN1kE:bg7igkv)* status: all files reviewed, 6 unresolved discussions (waiting on @annrpom)

___
*[`internal/base/comparer.go` line 516 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5622#-Ofkj8Kn-ycPoKqsCkNd:-Ofkj8Kn-ycPoKqsCkNe:blx8nb6) ([raw file](https://github.com/cockroachdb/pebble/blob/b545302d11f5d09645b787ac9ae266d8fe1bf4e4/internal/base/comparer.go#L516)):*
> ```Go
> 	for _, p := range prefixes {
> 		if len(p) > 1 {
> 			prefixes = append(prefixes, p[1+rand.IntN(len(p)-1):])
> ```

is this an unrelated change that tests a random prefix instead of exhaustively checking all?

RaduBerinde:
*[`internal/base/comparer.go` line 516 at r6](https://reviewable.io/reviews/cockroachdb/pebble/5622#-Ofkj8Kn-ycPoKqsCkNd:-OfkkDJ189fKOmbxeU1v:bufmekp) ([raw file](https://github.com/cockroachdb/pebble/blob/b545302d11f5d09645b787ac9ae266d8fe1bf4e4/internal/base/comparer.go#L516)):*

is this an unrelated change that tests a random prefix instead of exhaustively checking all?
</blockquote></details>

Yeah, it was way too slow once I added more test prefixes

annrpom:
TFTRs! ('-')7

PR #5623, opened 2025-12-03
tool: add db analyze-metadata command

RaduBerinde:
#### compressionanalyzer: move welford implementation

Move to metricsutil so it can be reused.

#### ascii: support unicode

Use runes instead of bytes in ascii.Board.

#### tool: add db analyze-metadata command

Add a command that prints out various metadata stats from sstables,
separated by level. This is useful to get an idea of block counts and
sizes, bytes per KV, etc.

### DIFF
--- a/internal/manifest/version.go
+++ b/internal/manifest/version.go
@@ -171,6 +171,10 @@ type Version struct {
 	prev, next *Version
 }
 
+func (v *Version) Comparer() *base.Comparer {
+	return v.cmp
+}
+
 // String implements fmt.Stringer, printing the TableMetadata for each level in
 // the Version.
 func (v *Version) String() string {
--- a/tool/db.go
+++ b/tool/db.go
@@ -49,6 +49,7 @@ type dbT struct {
 	IOBench         *cobra.Command
 	Excise          *cobra.Command
 	AnalyzeData     *cobra.Command
+	AnalyzeMetadata *cobra.Command
 
 	// Configuration.
 	opts            *pebble.Options
@@ -80,6 +81,10 @@ type dbT struct {
 		timeout       time.Duration
 		readMBPerSec  int
 		outputCSVFile string
+	}
+	analyzeMetadata struct {
+		samplePercent int
+		timeout       time.Duration
 	}
 }
 
@@ -243,6 +248,18 @@ experiments and produce a CSV file of the results.
 		Args: cobra.ExactArgs(1),
 		Run:  d.runAnalyzeData,
 	}
+	d.AnalyzeMetadata = &cobra.Command{
+		Use:   "analyze-metadata <dir>",
+		Short: "analyze sstable metadata statistics",
+		Long: `
+Sample sstables from the database in round-robin fashion across levels and
+compute running statistics (mean and standard deviation) for various metadata
+properties including file sizes, KV counts, index block characteristics, and
+filter sizes.
+`,
+		Args: cobra.ExactArgs(1),
+		Run:  d.runAnalyzeMetadata,
+	}
 
 	d.Inspect.AddCommand(&cobra.Command{
 		Use:   "manifest <dir>",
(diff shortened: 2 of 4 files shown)


RaduBerinde (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5623#-:-Ofkxisg1_wDJlZTzYj3:b-814g8k)* status: 4 of 10 files reviewed, 6 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`internal/ascii/whiteboard.go` line 14 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5623#-Ofk9NtZ3-SbTgxS6_V3:-OfkwigMDIFzy9F_PT-j:by960dk) ([raw file](https://github.com/cockroachdb/pebble/blob/2049bd69162264334f2991f9e5d7d48ee1c72c71/internal/ascii/whiteboard.go#L14)):*

is this no longer limited to ASCII, hence the runes?
</blockquote></details>

Updated the comment.
___
*[`internal/ascii/whiteboard.go` line 73 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5623#-Ofk9SIa5crE3CdiKiRh:-Ofkw4RAESeLPxFRne9R:b-yrd9b2) ([raw file](https://github.com/cockroachdb/pebble/blob/2049bd69162264334f2991f9e5d7d48ee1c72c71/internal/ascii/whiteboard.go#L73)):*

wouldn't  `utf8.RuneCountInString` be cheaper?

or given the `for _, ch := range s` below, which needs to convert to runes again, do it once here?
</blockquote></details>

The compiler recognizes `len([]rune(str))` and doesn't actually make the conversion (https://go-review.googlesource.com/c/go/+/108985). Added a comment.

We have to first grow the width, then we can write out the runes, I don't see how we can avoid going over the string twice. Anyway, this is not performance-sensitive code.
___
*[`tool/db_analyze_metadata.go` line 33 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5623#-OfkCdGPCzdUIa1qrgxi:-OfkxR6E1zFs-Xbv333A:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b88f0ed1ed2fa4304952f33ea2670854986bba8d/tool/db_analyze_metadata.go#L33)):*

this could use a comment
</blockquote></details>

Done.
___
*[`tool/db_analyze_metadata.go` line 47 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5623#-OfkBZ-JCw72c57Lpb8j:-OfkvbCw032SRExGOl9Z:bss4a2b) ([raw file](https://github.com/cockroachdb/pebble/blob/b88f0ed1ed2fa4304952f33ea2670854986bba8d/tool/db_analyze_metadata.go#L47)):*

I am pattern matching on "num", "size" or "bytes" to figure out the units of each name, with "size" also meaning "bytes". Is that reasonable?
`lowerLevelIndexBlocks` does not match this pattern. Should this `numLowerLevelIndexBlocks`?
</blockquote></details>

Removed.
___
*[`tool/db_analyze_metadata.go` line 234 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5623#-OfkDeJ-8xQtJcEVlm3r:-Ofkve7pBPnQh_2LamFs:btgbyde) ([raw file](https://github.com/cockroachdb/pebble/blob/b88f0ed1ed2fa4304952f33ea2670854986bba8d/tool/db_analyze_metadata.go#L234)):*

`numIndexBlocks` is the number of lower-level index blocks (or 1 in the case single level index). So it excludes the top-level index block, yes?
But `IndexSize` includes the top-level index block, so `ls.indexBlockSize.Add(float64(properties.IndexSize) / float64(numIndexBlocks))` doesn't seem quite right. Fine to leave a TODO.
</blockquote></details>

Switched to just tracking the total index size.
___
*[`tool/db_analyze_metadata.go` line 236 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5623#-OfkCro8ChtvfCbuBkbQ:-Ofkvgmh3dAhui_E4r2s:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/b88f0ed1ed2fa4304952f33ea2670854986bba8d/tool/db_analyze_metadata.go#L236)):*

If `numIndexBlocks` is 5, should this be adding 5 samples with this mean size, so at least the mean of the result is correctly weighted?
</blockquote></details>

Done.

sumeerbhola (approved):
:lgtm: 

> @sumeerbhola reviewed 2 of 2 files at r5, 4 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5623#-:-OflUNYxE31FUhHrooo8:bg7igkv)* status: all files reviewed (commit messages unreviewed), 4 unresolved discussions (waiting on @jbowens)

Issue #5466, opened 2025-10-20
db: add new columns to datablock


annrpom:
As a part of tiered storage, we should add tiering (spanId, attribute) columns to our DataBlock. See: https://github.com/cockroachdb/pebble/commit/f8876c6d9e184584df6a51cab8c1d3773833d3f1, https://github.com/cockroachdb/pebble/commit/2a099c206139ac2f67844a447ab1f5b8eca1c1c6

Jira issue: PEBBLE-1227

Epic CRDB-48982

PR #5626, opened 2025-12-05
sstable: add (spanID, tiering attr) cols to data block

annrpom:
Add tiering metadata columns to columnar data blocks to support tiered storage.
The new columns store a spanID and tiering attribute for each key, enabling
Pebble to track tiering policies and determine when data should transition
between hot and cold storage tiers.

Extend the iterator interfaces to propagate KVMeta through the compaction
pipeline, preserving tiering metadata when sstables are rewritten.

Fixes: #5466


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5626)* status: 10 of 52 files reviewed, 5 unresolved discussions (waiting on @annrpom, @jbowens, and @sumeerbhola)

___
*[`sstable/block/block.go` line 247 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5626#-OfzlBhR8Ws_c-Dyqk4I:-Og05PZwBPg5N8K2QqPe:b831m3r) ([raw file](https://github.com/cockroachdb/pebble/blob/abd7b93bf17a0a5ef3a6ff75fa5b5f507128f27e/sstable/block/block.go#L247)):*

> This increase is unfortunate. Especially since it will cause a lot of internal fragmentation when reading blocks written by existing versions, which were aligned to size classes according to the old metadata size.

Good catch.
This fragility is unfortunate, since we increased by only 48 bytes which is 0.14% of the 32KiB size class our data blocks will typically use. We could have padded the original value by 0.5%, which is ~163 bytes, for future growth.
Of course that is now in the past and we are stuck with this value since the next size class is 40KiB and we can't tolerate 20% internal fragmentation.
</blockquote></details>

Should we add a bit of extra slack in the [flush governor](https://github.com/cockroachdb/pebble/blob/501eb1672ddb689f3a4f6a28f330b4712c6a5b50/sstable/block/flush_governor.go#L45), for the future? The tradeoff is that we would waste that extra memory.

Or perhaps we should consider some other design for the metadata (e.g. a separate slab allocator).
___
*[`sstable/colblk/data_block.go` line 516 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5626#-Ofzkzwb-qpF1sQgsUTM:-Og04k1i9EZE_NHJrVBs:b-qhxwn6) ([raw file](https://github.com/cockroachdb/pebble/blob/abd7b93bf17a0a5ef3a6ff75fa5b5f507128f27e/sstable/colblk/data_block.go#L516)):*

At a high-level that sounds reasonable.
But I am unsure given our current practice.

- We pick one TableFormat to use for each FormatMajorVersion and the TableFormat has one layout for each block.
- NB: This ColumnFormat is not written to the sstable, and is simply implied by the TableFormat.
- The prototype introduced the ColumnFormat in case in the future we had to make changes that were not simply additions of columns. Also, we don't have a proto like encoding that can tell us from within the block which columns are present or absent (at decoding time), so having an externally provided ColumnFormat can make it simple to guide that decoding decision.

I think we should still have a single latest table format which has one "column format". The "column format" is a representation of all the features that it may need to support in this latest table format. Then the question becomes whether supporting all the features requires all the columns to be present, or whether some columns can be absent with the caller specifying what is present/absent. Something like `DataBlock{Encoder,Decoder}.Init(format ColumnarFormat, hasColumFoo bool, hasColumnBar bool, ...)`.

The prototype chose to make the two tiering columns in `ColumnFormatv2` always present since encoding the zero value was trivially cheap. That may not be the case for future changes.
</blockquote></details>

Yes, you are right. In the encoded block, the extra columns will be just one byte if unused. I agree, we can have one format.

annrpom (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5626)* status: 6 of 53 files reviewed, 2 unresolved discussions (waiting on @jbowens, @RaduBerinde, and @sumeerbhola)

___
*[`sstable/colblk/data_block.go` line 512 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5626#-OgAVTcL14su1WGugZ6x:-OgCy2nL2V8zhzx4JzPX:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/955984ae6d7d25b9d0dac1d429897631713f9eb6/sstable/colblk/data_block.go#L512)):*

[nit] I'd move the V1 and V2 at the end and set them to the specific column, and let the rest be iota
</blockquote></details>

done
___
*[`sstable/colblk/data_block.go` line 524 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5626#-OgAVEuDCYf3v1ua96H8:-OgDFW_62PVHAXtlXKJh:b-8c6xex) ([raw file](https://github.com/cockroachdb/pebble/blob/955984ae6d7d25b9d0dac1d429897631713f9eb6/sstable/colblk/data_block.go#L524)):*

I think we should either have two formats and determine the column set from the format, or don't add the `ColumnarFormat` concept at all.
</blockquote></details>

done (i removed the concept of the `ColumnarFormat` since we don't need it currently)
___
*[`sstable/colblk/data_block.go` line 617 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5626#-OgAUHXV1kVHCI4eZhZU:-OgD-dVb60oLkwGd07CO:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/955984ae6d7d25b9d0dac1d429897631713f9eb6/sstable/colblk/data_block.go#L617)):*

[nit] rename to `optionalColConfig` (maybe we'll use it for more than tiering)
</blockquote></details>

done
___
*[`sstable/colblk/data_block.go` line 627 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5626#-OgATiTRBVDX5nnfujSW:-OgD-lyt9igp3zjSK-op:byqoqzz) ([raw file](https://github.com/cockroachdb/pebble/blob/955984ae6d7d25b9d0dac1d429897631713f9eb6/sstable/colblk/data_block.go#L627)):*

We should store a copy of the config in the encoder instead of allocating a copy here.
</blockquote></details>

oops; done
___
*[`sstable/colblk/data_block.go` line 1523 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5626#-OgATZM5ByFmWhtYRGTe:-OgD0dPi97giKFUPD8QR:b2gbf6k) ([raw file](https://github.com/cockroachdb/pebble/blob/955984ae6d7d25b9d0dac1d429897631713f9eb6/sstable/colblk/data_block.go#L1523)):*

[nit] this could return bool so the caller can check directly
</blockquote></details>

good idea; done

annrpom:
TFTRs! ('-')7 !

PR #5628, opened 2025-12-06
crl-release-25.4: cockroachkvs: use bytes.Compare when comparing key version 

annrpom:
### cockroachkvs: use bytes.Compare when comparing key version

This patch uses bytes.Compare to compare key versions instead of
ComparePointSuffixes; the latter calls on
`normalizeEngineSuffixForCompare` - which could end up strip additional
bytes from the `suffix` (comparing only a prefix of the suffix) if the
last byte of the version happens to match one of the sentinel lengths
defined.

Informs: https://github.com/cockroachdb/cockroach/issues/158463

### cockroachkvs: add randomized test for IsLowerBound

### DIFF
--- a/cockroachkvs/cockroachkvs_test.go
+++ b/cockroachkvs/cockroachkvs_test.go
@@ -30,17 +30,34 @@ import (
 	testify "github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


annrpom:
TFTRs! ('-')7

PR #5631, opened 2025-12-06
manual: use sync.Pool to help the GC when CGo is off

elliotnunn:
When CGO_ENABLED=0, the manual cache memory management described at https://github.com/cockroachdb/pebble/blob/master/docs/memory.md is essentially stubbed out, because there is no access to C.free().

This was causing very large heap growth and GC churn in my application.

We can still get some of the benefit of manual memory management just by using a sync.Pool for each size class of cache allocation. This significantly reduces my application's heap usage.

### DIFF
--- a/internal/manual/manual_nocgo.go
+++ b/internal/manual/manual_nocgo.go
@@ -7,6 +7,8 @@
 package manual
 
 import (
+	"math/bits"
+	"sync"
 	"unsafe"
 
 	"github.com/cockroachdb/pebble/internal/invariants"
@@ -21,9 +23,8 @@ func New(purpose Purpose, n uintptr) Buf {
 		return Buf{}
 	}
 	recordAlloc(purpose, n)
-	slice := make([]byte, n)
 	return Buf{
-		data: unsafe.Pointer(unsafe.SliceData(slice)),
+		data: pools[sizeClass(n)].Get().(unsafe.Pointer),
 		n:    n,
 	}
 }
@@ -32,5 +33,33 @@ func New(purpose Purpose, n uintptr) Buf {
 // returned by New.
 func Free(purpose Purpose, b Buf) {
 	invariants.MaybeMangle(b.Slice())
+	if b.data == nil {
+		return
+	}
+
+	// Clear the block for consistency with the cgo implementation (which uses calloc)
+	clear(unsafe.Slice((*byte)(b.data), b.n))
+
 	recordFree(purpose, b.n)
+	pools[sizeClass(b.n)].Put(b.data)
+}
+
+var pools [bits.UintSize]sync.Pool // pools[n] is for allocs of size 1 << n
+
+func init() {
+	for i := range pools {
+		allocSize := 1 << i
+		pools[i].New = func() any {
+			// Boxing an unsafe.Pointer into an interface does not require an allocation
+			return unsafe.Pointer(unsafe.SliceData(make([]byte, allocSize)))
+		}
+	}
+}
+
+// sizeClass determines the smallest n such that 1 << n >= size
+func sizeClass(size uintptr) int {
+	if invariants.Enabled && size == 0 {
+		panic("zero size should never get through New")
+	}
+	return bits.Len(uint(size - 1))
 }


RaduBerinde (approved):
:lgtm:  Just some nits.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5631#-:-OfoRQ39-xLPOW-KRyba:b-c4i251)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @elliotnunn)

___
*[`internal/manual/manual_nocgo.go` line 48 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5631#-OfoQxyh3xkL42R5rAB5:-OfoQxyh3xkL42R5rAB6:b-irde2h) ([raw file](https://github.com/cockroachdb/pebble/blob/9681a49f74ad099ced4fcd1284d30a200c70d812/internal/manual/manual_nocgo.go#L48)):*
> ```Go
> 	var pools [bits.UintSize]sync.Pool
> 	for i := range pools {
> 		i := i // watch out for capture bugs
> ```

[nit] Could do `allocSize := 1 << i` instead
___
*[`internal/manual/manual_nocgo.go` line 50 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5631#-OfoSdovDO0wvImmpNbS:-OfoSdovDO0wvImmpNbT:bv45zof) ([raw file](https://github.com/cockroachdb/pebble/blob/9681a49f74ad099ced4fcd1284d30a200c70d812/internal/manual/manual_nocgo.go#L50)):*
> ```Go
> 		i := i // watch out for capture bugs
> 		pools[i].New = func() any {
> 			return unsafe.Pointer(unsafe.SliceData(make([]byte, 1<<i)))
> ```

[nit] Add a note `// Boxing an unsafe.Pointer into an interface does not require an allocation.` (maybe it's obvious but I for one had to check).
___
*[`internal/manual/manual_nocgo.go` line 58 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5631#-OfoQ_f55Db9tm3E0EK3:-OfoQ_f55Db9tm3E0EK4:bm4egf1) ([raw file](https://github.com/cockroachdb/pebble/blob/9681a49f74ad099ced4fcd1284d30a200c70d812/internal/manual/manual_nocgo.go#L58)):*
> ```Go
> // sizeClass determines the smallest n such that 1 << n >= size
> func sizeClass(size uintptr) int {
> 	return bits.UintSize - bits.LeadingZeros(uint(size-1))
> ```

[nit] This can just be `bits.Len(size-1)`. I'd also add an `if invariants.Enabled && size == 0 { panic() }`

RaduBerinde (commented):
By the way, you can populate `AllocatorSizeClasses` which will make the data blocks (plus in-memory metadata) fit better.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5631#-:-OfpxEiXCVifN_HEd5Dk:b-hkemln)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @elliotnunn)

___
*[`internal/manual/manual_nocgo.go` line 45 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5631#-OfpwyW52HDdg_gf4esG:-OfpwyW69SG2BbpvvlMb:be5mo2w) ([raw file](https://github.com/cockroachdb/pebble/blob/9681a49f74ad099ced4fcd1284d30a200c70d812/internal/manual/manual_nocgo.go#L45)):*
> ```Go
> var pools = mkPools() // pools[n] is for allocs of size 1 << n
> 
> func mkPools() [bits.UintSize]sync.Pool {
> ```

linter is complaining about copying the `sync.Pool`, we'll have to do `var pools [bits.UintSize]` and then `func init() { .. }`.  Hopefully the other test failures are related.

elliotnunn:
Thank you for the feedback. I have incorporated most of it, but I'm not sure how to set AllocatorSizeClasses.

A full-tree test ("go test ./...") was failing because stale pointers were found in freed blocks. I have pushed a commit to clear() these blocks. Should this be conditional on invariants.Enabled?

elliotnunn:
Hi, bump. Will this be merged?

RaduBerinde:
Sorry, missed your push.

PR #5632, opened 2025-12-08
bloom: use optimal number of probes

RaduBerinde:
The bloom filter code (derived from RocksDB) uses a variation: the
probes for a key are not fully random, but are generated using double
hashing to stay within a single cacheline.

This scheme behaves similarly (in terms of false positive rate) to
regular bloom filters for bits-per-key values up around 10. Beyond
that, it yields significantly lower FPR as a larger fraction of cache
lines "fill up". In addition, the optimal number of probes is smaller
than the vanilla bloom filter formula.

This change adds a simulation used to determine the best number of
probes in practice and updates the number of probes to the optimal
values. Note that currently we (in CRDB) only use 10 bits per key, for
which the optimal number of probes is unchanged.

We also optimize the code to avoid bound checks inside the inner loop.
```
name            old time/op  new time/op  delta
BloomFilter-10  44.9µs ± 1%  43.6µs ± 0%  -2.92%  (p=0.000 n=10+9)
```


RaduBerinde:
CC @petermattis - you may find this interesting.

RaduBerinde:
Friendly ping :)

RaduBerinde:
*[`bloom/simulation.txt` line 64 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5632#-OgDN7pVAeJYLMYheQRL:-OgDPW7B-XXXLIzRAniZ:bncj5nv) ([raw file](https://github.com/cockroachdb/pebble/blob/4c3e734e97b09763211fc381935f465e5de3be35/bloom/simulation.txt#L64)):*

4 and 5 probes seem to yield the same FPR. Is there a reason the probes table is using 5 and not 4?
</blockquote></details>

Good catch. It's probably a tiny bit better, I'll update the program to prefer fewer probes when the difference is trivial.

PR #5634, opened 2025-12-09
db: new FMV for blob file ingest

xinhaoz:
This FMV will gate the ingestiong of blob files in local ingestion and as flushableIngest.


xinhaoz:
Addressed nits, tftr!

Issue #5635, opened 2025-12-09
db: decide on the correct shape for large LSMs


RaduBerinde:
We currently use the code below to calculate a smoothed level multiplier:

```go
	bottomLevelSize := dbSize - dbSize/uint64(p.opts.Experimental.LevelMultiplier)
	...
	smoothedLevelMultiplier := 1.0
	if p.baseLevel < numLevels-1 {
		smoothedLevelMultiplier = math.Pow(
			float64(bottomLevelSize)/float64(baseBytesMax),
			1.0/float64(numLevels-p.baseLevel-1))
	}

	levelSize := float64(baseBytesMax)
	for level := p.baseLevel; level < numLevels; level++ {
		if level > p.baseLevel && levelSize > 0 {
			levelSize *= smoothedLevelMultiplier
		}
		...
	}
```

At the CRDB settings (baseBytesMax=64MiB, LevelMultiplier=10), once the LSM size goes over 7TiB, the smothedLevelMultiplier goes [above 10](https://www.google.com/search?q=%28%287*0.9%29+TiB+%2F+64+MiB%29%5E%281%2F5%29). At the current advisory limit of 10TiB, it is [10.8](https://www.google.com/search?q=%28%2810*0.9%29+TiB+%2F+64+MiB%29%5E%281%2F5%29). At a hypothetical 40TiB, it is [14.3](https://www.google.com/search?q=%28%2840*0.9%29+TiB+%2F+64+MiB%29%5E%281%2F5%29).

Instead of letting the multiplier go above the `LevelMultiplier`, we should increase the target for L1.

It could be something like this:
```
	if smoothedLevelMultiplier > float64(p.opts.Experimental.LevelMultiplier) {
		levelSize = float64(curLevelSize)
		smoothedLevelMultiplier = float64(p.opts.Experimental.LevelMultiplier)
	}
```

Jira issue: PEBBLE-1296

PR #5637, opened 2025-12-10
db: limit level size multiplier

RaduBerinde:
#### db: limit level size multiplier

For db sizes above ~7TB, the level size multiplier grows beyond the
configured one. This is not what we want, as the LSM would suffer from
high write amp.

Instead, we cap the multiplier to the configured one (10 by default)
and let L1 grow as needed. In the future, we should increase memtable
sizes proportionally with L1.

Informs #5635

### DIFF
--- a/compaction_picker.go
+++ b/compaction_picker.go
@@ -824,9 +824,9 @@ func (p *compactionPickerByScore) initLevelMaxBytes(inProgressCompactions []comp
 	// Determine the first non-empty level and the total DB size.
 	firstNonEmptyLevel := numLevels - 1
 	var dbSize uint64
-	for level := 1; level < numLevels; level++ {
+	for level := numLevels - 1; level > 0; level-- {
 		if p.vers.Levels[level].AggregateSize() > 0 {
-			firstNonEmptyLevel = min(firstNonEmptyLevel, level)
+			firstNonEmptyLevel = level
 			dbSize += p.vers.Levels[level].AggregateSize()
 		}
 	}
@@ -846,7 +846,7 @@ func (p *compactionPickerByScore) initLevelMaxBytes(inProgressCompactions []comp
 		// No levels for L1 and up contain any data. Target L0 compactions for the
 		// last level or to the level to which there is an ongoing L0 compaction.
 		p.baseLevel = firstNonEmptyLevel
-		for level := 0; level < numLevels; level++ {
+		for level := range p.levelMaxBytes {
 			p.levelMaxBytes[level] = math.MaxInt64
 		}
 		return
@@ -900,6 +900,17 @@ func calculateLevelSizes(
 	}
 
 	levelSize := float64(baseBytesMax)
+	if smoothedLevelMultiplier > float64(opts.Experimental.LevelMultiplier) {
+		// Don't let the multiplier grow beyond the configured level multiplier.
+		// Instead, grow all levels proportionally by allowing a larger L1.
+		//
+		// TODO(radu): revisit this decision, as it could result in much more
+		// expensive L0->L1 compactions. Consider increasing the memtable size
+		// proportionally.
+		smoothedLevelMultiplier = float64(opts.Experimental.LevelMultiplier)
+		levelSize = float64(curLevelSize)
+	}
+
 	for level := baseLevel; level < numLevels; level++ {
 		// Round the result since test cases use small target level sizes, which
 		// can be impacted by floating-point imprecision + integer truncation.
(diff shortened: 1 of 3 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5637)* status: all files reviewed, 4 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`compaction_picker.go` line 829 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5637#-OgAj2A51kdfR_FR_eS2:-OgENyhV1RMRtKD8ue4f:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/fe24b8f9ff83524c5502d35dbb727d0d2557c640/compaction_picker.go#L829)):*

nit: it is a bit odd to see a min here. Iterating with `for level := numLevels-1; level >= 1; level--` would avoid it.
Not a strong opinion.
</blockquote></details>

Done.
___
*[`testdata/compaction_picker_level_max_bytes` line 18 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5637#-OgAhozUErn1sz35mqFZ:-OgEO_xQ3KjTuTA3ONXO:bt2vr1b) ([raw file](https://github.com/cockroachdb/pebble/blob/5ee9851f9339d8d13f44e8c2fa08e841dc82ba1b/testdata/compaction_picker_level_max_bytes#L18)):*

is this a duplicate of the previous test case?
</blockquote></details>

Fixed.
___
*[`testdata/compaction_picker_level_max_bytes` line 196 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5637#-OgAmxxa8if4MmevoqJ4:-OgEOhcmDzvFgIAw2a0w:b-dz7alm) ([raw file](https://github.com/cockroachdb/pebble/blob/9a17d504fb07c12faa4f8f2f1aeba0578634af5b/testdata/compaction_picker_level_max_bytes#L196)):*

> Or when the LSM becomes this large, we could increase the size of our memtables so L0=>L1 compactions are just bigger but not necessarily more wasteful in terms of write amp? Conceptually we should be able to scale everything up proportionally.

Agreed. Our memtables are on the smaller side compared to some other system configurations I have seen.

Some of this will also implicitly improve when the memtables don't have the raft log entries, since when those are getting dropped during the flush (because of raft log truncation) the L0 files become even smaller which increases the L0 => L1 ratio for a compaction.
</blockquote></details>

Added a TODO and a note in the commit message.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5637)* status: 1 of 4 files reviewed, 4 unresolved discussions (waiting on @jbowens and @sumeerbhola)

___
*[`testdata/compaction_picker_level_max_bytes` line 196 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5637#-OgAmxxa8if4MmevoqJ4:-OgEQCG13hTPULEvNGsg:b-5oto2b) ([raw file](https://github.com/cockroachdb/pebble/blob/9a17d504fb07c12faa4f8f2f1aeba0578634af5b/testdata/compaction_picker_level_max_bytes#L196)):*

Added a TODO and a note in the commit message.
</blockquote></details>

I'll also repurpose the issue and keep it open until we do the memtable thing too, or something else

jbowens (approved):
:lgtm:

> @jbowens reviewed 1 of 3 files at r1, 3 of 3 files at r4, 1 of 1 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5637#-:-OgINhy31ena9Te8jq8g:bnfp4nl)* status: all files reviewed, 5 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`compaction_picker.go` line 849 at r5](https://reviewable.io/reviews/cockroachdb/pebble/5637#-OgINQDK2rEQqZn8tdm6:-OgINQDK2rEQqZn8tdm7:brn8zc6) ([raw file](https://github.com/cockroachdb/pebble/blob/0a5eeb20f2f00877f4f585fbaa47c5bd5349765e/compaction_picker.go#L849)):*
> ```Go
> 		// last level or to the level to which there is an ongoing L0 compaction.
> 		p.baseLevel = firstNonEmptyLevel
> 		for level := 0; level < numLevels; level++ {
> ```

nit: could be `for level := range numLevels {`, if you like

RaduBerinde:
TFTR!

PR #5640, opened 2025-12-10
db: remove unnecessary filtering during deletion hint collection

jbowens:
During table stats collection, we collect 'delete compaction hints' describing range tombstones and sstables they delete. Originally, there was a subtle edge case at the intersection of sequence number zeroing and these hints. In f729ccd we refactored this interaction to make it less subtle, but inadvertently left behind some of the previous logic.

Specifically, once we've collected hints, we previously would look up these hints' tombstone files in the current LSM version, discarding the hint if the file is not found in its original level. With f729ccd4's introduction of LargestSeqNumAbsolute, this extra lookup is no longer necessary. Additionally, removing this filter step ensures that we'll preserve the hint even if the sstable has been moved into a lower level.

### DIFF
--- a/table_stats.go
+++ b/table_stats.go
@@ -148,26 +148,8 @@ func (d *DB) collectTableStats() bool {
 
 	d.mu.tableStats.cond.Broadcast()
 	d.maybeCollectTableStatsLocked()
-	if len(hints) > 0 && !d.opts.private.disableDeleteOnlyCompactions {
-		// Verify that all of the hint tombstones' files still exist in the
-		// current version. Otherwise, the tombstone itself may have been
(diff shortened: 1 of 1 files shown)


sumeerbhola (approved):
:lgtm: 

> @sumeerbhola reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5640#-:-OgEBsoL9v77h1Ckk3MJ:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jbowens)

jbowens:
TFTR!

PR #5642, opened 2025-12-11
db: move delete-only compaction code into compaction_delete.go

jbowens:
Move code specific to delete-only compactions into compaction_delete.go.


jbowens:
TFTR!

PR #5643, opened 2025-12-11
bloom: add adaptive filter policy

RaduBerinde:
#### pebble: improve table filter infrastructure

This commit separates the two aspects of `TablePolicy`:
 - `TableFilterPolicy` now deals with creation of filters;
 - `TableFilterDecoder` now deals with using/reading filters.

The old table policy name is now a filter *family* name. Each family
is associated with a `TableFilterDecoder`.

The separation will allow smarter filter policies, like automatically
reducing the number of bits to keep the filter size under a threshold,
or even dynamically choosing between different families.

#### bloom: add adaptive filter policy

Add an adaptive bloom filter which is configured with a max size. The
bits per key are reduced as necessary so the filter fits within the
given size. The filter format does not change and remains readable by
older versions.

We plan to use this to generate smaller filters in L6 (in a future
change).

### DIFF
--- a/bloom/adaptive_policy.go
+++ b/bloom/adaptive_policy.go
@@ -1,0 +1,69 @@
+// Copyright 2025 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package bloom
+
+import (
+	"fmt"
+
+	"github.com/cockroachdb/pebble/internal/base"
+)
+
+// AdaptivePolicy implements base.TableFilterPolicy for Bloom filters. It
+// automatically reduces the number of bits per key so the filter size stays
+// under a given limit.
+func AdaptivePolicy(targetBitsPerKey uint32, maxFilterSize uint64) base.TableFilterPolicy {
+	return adaptivePolicyImpl{TargetBitsPerKey: targetBitsPerKey, MaxSize: maxFilterSize}
+}
+
+type adaptivePolicyImpl struct {
+	TargetBitsPerKey uint32
+	MaxSize          uint64
+}
+
+var _ base.TableFilterPolicy = adaptivePolicyImpl{}
+
+func (p adaptivePolicyImpl) Name() string {
+	return fmt.Sprintf("adaptive_bloom(%d,%d)", p.TargetBitsPerKey, p.MaxSize)
+}
+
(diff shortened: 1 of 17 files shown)


jbowens (approved):
:lgtm:

> @jbowens reviewed 38 of 38 files at r1, 15 of 15 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5643#-:-OgIyTzC1TbEh68rLNpo:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @RaduBerinde and @sumeerbhola)

___
*[`bloom/bloom.go` line 281 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5643#-OgIPlTO39vIuRU-UnWW:-OgIPlTO39vIuRU-UnWX:b-h7fr7g) ([raw file](https://github.com/cockroachdb/pebble/blob/a969150d229920aeec89e80d994444ace667d278/bloom/bloom.go#L281)):*
> ```Go
> 		return string(Family)
> 	}
> 	return fmt.Sprintf("bloom(%d)", p.BitsPerKey)
> ```

what codepaths attempt to parse this? Do we need to gate this on a format major version?
___
*[`bloom/adaptive_policy.go` line 15 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5643#-OgIxIdfEkffnCQ24JSg:-OgIxIdfEkffnCQ24JSh:b-cm0t0q) ([raw file](https://github.com/cockroachdb/pebble/blob/79897db787b66976d1b636a7b402253faa11ba33/bloom/adaptive_policy.go#L15)):*
> ```Go
> // AdaptivePolicy implements base.TableFilterPolicy for Bloom filters. It
> // automatically reduces the number of bits per key so the filter size stays
> // under a given limit.
> ```

nit: should this comment be on the `AdaptivePolicy` func?
___
*[`options.go` line 447 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5643#-OgIOA_z0x-UxNx236Ex:-OgIOA_z0x-UxNx236Ey:bns7rs8) ([raw file](https://github.com/cockroachdb/pebble/blob/a969150d229920aeec89e80d994444ace667d278/options.go#L447)):*
> ```Go
> 
> 	// TableFilterPolicy defines a filter algorithm (such as a Bloom filter) that can
> 	// reduce disk reads for Get calls.
> ```

nit: we could update this to say Get and SeekPrefixGE calls

PR #5647, opened 2025-12-12
db: clean up collection of delete-compaction hints

jbowens:
**db: simplify delete-only compaction snapshot threshold** 

Previously, a delete-only compaction hint was considered ready to be resolved
once the hint's largest tombstone sequence number fell in the same snapshot
stripe as the lowest sequence number of a table that was originally observed as
eligible for a delete-only compaction at the time the hint was created. Then,
during resolution of the delete-only compaction hint, the compaction picker
retrieved the set of eligble tables in the current LSM version (which may be
different from the version at which the hint was created), and it scheduled
deletion of a table if that table also fell in the same snapshot stripe as the
tombstone's largest sequence number.

This logic could allow a delete-only compaction to delete a table even if a
snapshot older than the table's data remained. This logic could also result in
premature discarding of a hint if compactions had reorganized data such that a
table that would otherwise be eligble for deletion now contains a key with a
sequence number less than the oldest snapshot. This logic is a tradeoff, and
it's unclear whether the tradeoff is beneficial.

Additionally, it's exceedingly unlikely for a snapshot to exist beneath a
eligble table's sequence number, especially with applications like CockroachDB
that use eventually file-only snapshots over ordinary LSM snapshots.

This commit simplifies these conditions to instead require that there exist no
snapshots beneath the largest tombstone sequence number whatsoever.

**db: remove deleteCompactionHintType**

Remove the deleteCompactionHintType enum, replacing usages with
manifest.KeyType.

**base: take UserKeyBounds by value in ContainsBounds, Overlaps** 

Update UserKeyBounds.ContainsBounds and UserKeyBounds.Overlaps to both take
UserKeyBounds, not *UserKeyBounds. This helps avoid allocations, especially
when working with TableMetadata.UserKeyBounds().

**db: clean up collection of delete-compaction hints**

This commit performs some pure hygenic cleanup of the table stats collection
code path that collection delete-compaction hints. We improve some of the
documentation and naming, and use UserKeyBounds to handle containment checking.

### DIFF
--- a/compaction_delete.go
+++ b/compaction_delete.go
@@ -115,7 +115,7 @@ func tombstoneKeyTypeFromKeys(keys []keyspan.Key) manifest.KeyType {
 		case base.InternalKeyKindRangeKeyDelete:
 			rangeKeys = true
 		default:
-			panic(fmt.Sprintf("unsupported key kind: %s", k.Kind()))
+			panic(errors.AssertionFailedf("unsupported key kind: %s", k.Kind()))
 		}
 	}
 	switch {
@@ -131,18 +131,18 @@ func tombstoneKeyTypeFromKeys(keys []keyspan.Key) manifest.KeyType {
 }
 
 // A deleteCompactionHint records a user key and sequence number span that has
-// been deleted by tombstones. A hint is recorded if at least one sstable falls
-// completely within both the user key and sequence number spans.  Once the
-// tombstones fall into the last snapshot stripe, a delete-only compaction may
-// delete any sstables within the range.
+// been deleted by tombstones. A hint is recorded if at least one sstable
+// containing keys older than the tombstones and either (a) completely falls
+// within both the user key, or (b) has a boundary that overlaps with the span
+// [which could be shorted with an excise]. Once the tombstones fall into the
+// last snapshot stripe, a delete-only compaction may delete or excise any
+// applicable sstables within the range.
 type deleteCompactionHint struct {
 	// The type of key span that generated this hint (point key, range key, or
 	// both).
 	keyType manifest.KeyType
-	// start and end are user keys specifying a key range [start, end) of
-	// deleted keys.
-	start []byte
-	end   []byte
+	// bounds are the bounds of the tombstone(s).
+	bounds base.UserKeyBounds
 	// The level of the file containing the range tombstone(s) when the hint
 	// was created. Only lower levels need to be searched for files that may
 	// be deleted.
@@ -172,11 +172,9 @@ const (
 
 func (h deleteCompactionHint) String() string {
 	return fmt.Sprintf(
-		"L%d.%s %s-%s seqnums(tombstone=%d-%d, type=%s)",
-		h.tombstoneLevel, h.tombstoneFile.TableNum, h.start, h.end,
-		h.tombstoneSmallestSeqNum, h.tombstoneLargestSeqNum,
-		h.keyType,
-	)
+		"L%d.%s %s seqnums(tombstone=%d-%d, type=%s)",
+		h.tombstoneLevel, h.tombstoneFile.TableNum, h.bounds,
+		h.tombstoneSmallestSeqNum, h.tombstoneLargestSeqNum, h.keyType)
 }
 
 func (h *deleteCompactionHint) canDeleteOrExcise(
@@ -216,8 +214,7 @@ func (h *deleteCompactionHint) canDeleteOrExcise(
 	default:
 		panic(errors.AssertionFailedf("pebble: unknown delete compaction key type: %s", h.keyType))
 	}
-	if cmp(h.start, m.Smallest().UserKey) <= 0 &&
-		base.UserKeyExclusive(h.end).CompareUpperBounds(cmp, m.UserKeyBounds().End) >= 0 {
+	if h.bounds.ContainsBounds(cmp, m.UserKeyBounds()) {
 		return hintDeletesFile
 	}
 	if !exciseEnabled {
@@ -227,8 +224,7 @@ func (h *deleteCompactionHint) canDeleteOrExcise(
 	}
 	// Check for any overlap. In cases of partial overlap, we can excise the part of the file
 	// that overlaps with the deletion hint.
-	if cmp(h.end, m.Smallest().UserKey) > 0 &&
-		(m.UserKeyBounds().End.CompareUpperBounds(cmp, base.UserKeyInclusive(h.start)) >= 0) {
+	if h.bounds.Overlaps(cmp, m.UserKeyBounds()) {
 		return hintExcisesFile
 	}
 	return hintDoesNotApply
@@ -319,7 +315,7 @@ func checkDeleteCompactionHints(
 		filesDeletedByCurrentHint := 0
 		var filesDeletedByLevel [7][]*manifest.TableMetadata
 		for l := h.tombstoneLevel + 1; l < numLevels; l++ {
-			for m := range v.Overlaps(l, base.UserKeyBoundsEndExclusive(h.start, h.end)).All() {
+			for m := range v.Overlaps(l, h.bounds).All() {
 				doesHintApply := h.canDeleteOrExcise(cmp, m, exciseEnabled)
 				if m.IsCompacting() || doesHintApply == hintDoesNotApply || files[m] {
 					continue
@@ -334,10 +330,10 @@ func checkDeleteCompactionHints(
 					// leaves a fragment of the file on the left, decrement
 					// the counter once. If the hint leaves a fragment of the
 					// file on the right, decrement the counter once.
-					if cmp(h.start, m.Smallest().UserKey) > 0 {
+					if cmp(h.bounds.Start, m.Smallest().UserKey) > 0 {
 						filesDeletedByCurrentHint--
 					}
-					if m.UserKeyBounds().End.IsUpperBoundFor(cmp, h.end) {
+					if m.UserKeyBounds().End.IsUpperBoundFor(cmp, h.bounds.End.Key) {
 						filesDeletedByCurrentHint--
 					}
 				}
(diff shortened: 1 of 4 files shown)


RaduBerinde:
*[`-- commits` line 15 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5647#-OgIz3weBrbOCjombyKr:-OgIz3weBrbOCjombyKs:b-euv3rm):*
> This logic could allow a delete-only compaction to delete a table even if a
>  snapshot older than the table's data remained.

Isn't that a correctness issue? Is yes, how come the metamorphic test didn't find it and do we need to backport a fix?

RaduBerinde:
*[`compaction_delete.go` line 206 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5647#-OgJ090I7HEphId9NsBh:-OgJ090I7HEphId9NsBi:b-v4sql6) ([raw file](https://github.com/cockroachdb/pebble/blob/7acdb655dcd5ac4333b64751dd9b555ab93ae485/compaction_delete.go#L206)):*
> ```Go
> 	cmp Compare, m *manifest.TableMetadata, exciseEnabled bool,
> ) deletionHintOverlap {
> 	// The file can only be deleted if all of its keys are older than the
> ```

There's something I don't quite get w.r.t rangedels and zeroing sequence numbers. If we don't do a delete compaction and just do a regular compaction, wouldn't the rangedel delete all keys with zero sequence number? How is it ever ok to have a key with zero seq num and a rangedel in a level above it that shouldn't delete the key?

And if it's always ok for a rangedel to delete zero sequence keys, why do we need ` LargestSeqNumAbsolute`?

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5647)* status: 0 of 18 files reviewed, 2 unresolved discussions (waiting on @jbowens and @xinhaoz)

___
*[`-- commits` line 15 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5647#-OgIz3weBrbOCjombyKr:-OgJ2pFqD108g3oAfoTd:bqubtxj):*

Naw, this was a somewhat desirable property in that it could allow us to delete a table earlier. When we apply tombstones, we allow them to delete keys within the same snapshot stripe. That's what we used to do here. This change makes the logic stricter, now requiring the tombstone (and the data being deleted) to all be within the last snapshot stripe.

I think in practice it's extremely rare for a table and the tombstone to be in the same snapshot stripe and _not_ be within the last snapshot stripe. Changing the behavior makes it simpler. When this case _does_ happen, we lose the promptness of the deletion that we had before, but we also avoid the premature discarding of the hint like the next sentence discusses.
</blockquote></details>

Oh, I see, thanks.
___
*[`compaction_delete.go` line 206 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5647#-OgJ090I7HEphId9NsBh:-OgJ2qdQETd74J2_8Dxl:bn1ulpd) ([raw file](https://github.com/cockroachdb/pebble/blob/7acdb655dcd5ac4333b64751dd9b555ab93ae485/compaction_delete.go#L206)):*

There's something I don't quite get w.r.t rangedels and zeroing sequence numbers. If we don't do a delete compaction and just do a regular compaction, wouldn't the rangedel delete all keys with zero sequence number? How is it ever ok to have a key with zero seq num and a rangedel in a level above it that shouldn't delete the key?

And if it's always ok for a rangedel to delete zero sequence keys, why do we need ` LargestSeqNumAbsolute`?
</blockquote></details>

Is it because a hint might be based off a tombstone that is later compacted away, and the hint still exists?

Would all this become simpler if instead of these hints, we would maintain a list of (say) top 100 "widest" (in terms of covered bytes) tombstones (together with their seqnum and level) in the current version?

RaduBerinde:
*[`compaction_delete.go` line 206 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5647#-OgJ090I7HEphId9NsBh:-OgJJ_of6s6qXCLUrkK1:b-7v58ns) ([raw file](https://github.com/cockroachdb/pebble/blob/7acdb655dcd5ac4333b64751dd9b555ab93ae485/compaction_delete.go#L206)):*

> Is it because a hint might be based off a tombstone that is later compacted away, and the hint still exists?

Yeah. If we ignore delete compactions, normal range deletions can always delete keys that are beneath them in the LSM, including if the keys' sequence numbers are zeroed, due to the LSM invariant. The LSM invariant guarantees that the keys lower in the LSM have lower sequence numbers than the range deletion higher in the LSM.

Things get weird only once we consider deletion compactions because we asynchronously discover the range deletions in the background table stats collector. Then we wait for any conflicting snapshots to close. Then we go to evaluate the hint. At that point, we don't know where the originating tombstones are, if they exist at all.
</blockquote></details>

This is just for my understanding, not necessarily a suggestion for implementation - but conceptually, if we kept a mapping from `TableMetadata` to relevant deletion hints and during compactions we would update the deletion hints to reflect the new tombstone level (or remove the hint if the tombstone goes away), we wouldn't need `LargestSeqNumAbsolute`?

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5647#-:-OgJLqOtE2oDRSbsFIGE:bnfp4nl)* status: 0 of 18 files reviewed, 4 unresolved discussions (waiting on @jbowens and @xinhaoz)

___
*[`compaction_delete.go` line 176 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5647#-OgJKvwH0bb5ScW_jQzG:-OgJKvwH0bb5ScW_jQzH:b-cpp8ns) ([raw file](https://github.com/cockroachdb/pebble/blob/3558f9389bf2580fc24514ed0de8dd78b91619a9/compaction_delete.go#L176)):*
> ```Go
> 	return fmt.Sprintf(
> 		"L%d.%s %s-%s seqnums(tombstone=%d-%d, type=%s)",
> 		h.tombstoneLevel, h.tombstoneFile.TableNum, h.bounds.Start, h.bounds.End.Key,
> ```

[nit] `h.bounds` has a better String()
___
*[`compaction_delete.go` line 228 at r4](https://reviewable.io/reviews/cockroachdb/pebble/5647#-OgJLm2kBHx4HItTR0sk:-OgJLm2kBHx4HItTR0sl:bxfzgr9) ([raw file](https://github.com/cockroachdb/pebble/blob/3558f9389bf2580fc24514ed0de8dd78b91619a9/compaction_delete.go#L228)):*
> ```Go
> 	// Check for any overlap. In cases of partial overlap, we can excise the part of the file
> 	// that overlaps with the deletion hint.
> 	if h.bounds.Overlaps(cmp, m.UserKeyBounds()) {
> ```

Great cleanup!
___
*[`excise.go` line 85 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5647#-OgJKeEwB5OGAdYSf9sv:-OgJKeEwB5OGAdYSf9sw:b1zii0f) ([raw file](https://github.com/cockroachdb/pebble/blob/3c8ad12f6feecf330706f86a54a935f00e9cae59/excise.go#L85)):*
> ```Go
> ) (leftTable, rightTable *manifest.TableMetadata, _ error) {
> 	// Check if there's actually an overlap between m and exciseSpan.
> 	if !exciseBounds.Overlaps(d.cmp, m.UserKeyBounds()) {
> ```

Did this really cause an allocation? The pointer shouldn't escape..

RaduBerinde:
*[`excise.go` line 85 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5647#-OgJKeEwB5OGAdYSf9sv:-OgXtuueBQj-JHsV_0SZ:b-15e4by) ([raw file](https://github.com/cockroachdb/pebble/blob/3c8ad12f6feecf330706f86a54a935f00e9cae59/excise.go#L85)):*

Looks like in this case `mBounds` doesn't, but `exciseBounds` does because `Overlaps` is defined on a pointer receiver :(

```
./excise.go:79:2: parameter exciseBounds leaks to {heap} for (*DB).exciseTable with derefs=0:
./excise.go:79:2:   flow: {heap} ← exciseBounds:
./excise.go:79:2:     from exciseBounds (address-of) at ./excise.go:86:18
./excise.go:79:2:     from (*base.UserKeyBounds).Overlaps(exciseBounds, d.cmp, &mBounds) (call parameter) at ./excise.go:86:27
```
</blockquote></details>

So frustrating, why would the receiver escape in `Overlap`..  Should we change `Overlap` to use a value receiver as well? And maybe all other `Contains` methods and such?

RaduBerinde:
*[`excise.go` line 85 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5647#-OgJKeEwB5OGAdYSf9sv:-OgXvdRiBWnHi9n9LacX:b-9rmh9v) ([raw file](https://github.com/cockroachdb/pebble/blob/3c8ad12f6feecf330706f86a54a935f00e9cae59/excise.go#L85)):*

So frustrating, why would the receiver escape in `Overlap`..  Should we change `Overlap` to use a value receiver as well? And maybe all other `Contains` methods and such?
</blockquote></details>

Are you sure that "leaks to" means it's moved to the heap? AFAICT "leaks" means that information derived from that arg makes it to the heap. There'd need to be a `"moved to heap"` message for an allocation to be involved.

jbowens (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5647)* status: 0 of 18 files reviewed, 3 unresolved discussions

___
*[`excise.go` line 85 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5647#-OgJKeEwB5OGAdYSf9sv:-OgYNWqHAgJFfrzH8q22:b1g3biw) ([raw file](https://github.com/cockroachdb/pebble/blob/3c8ad12f6feecf330706f86a54a935f00e9cae59/excise.go#L85)):*

Are you sure that "leaks to" means it's moved to the heap? AFAICT "leaks" means that information derived from that arg makes it to the heap. There'd need to be a `"moved to heap"` message for an allocation to be involved.
</blockquote></details>

Oh hrm yeah maybe it's not actually going to the heap then

PR #5650, opened 2025-12-12
sstable/tieredmeta: add tiering histogram using t-digest

annrpom:
Add StatsHistogram, a quantile sketch backed by t-digest, to track byte distribution by TieringAttribute. The histogram supports CDF and quantile queries to estimate bytes above/below thresholds for tiering decisions.

Add TieringHistogramBlockWriter to encode multiple histograms per sstable or blob file, keyed by (KindAndTier, SpanID). The block will support histogram types for inline values and blob references (hot/cold).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ require (
 	github.com/DataDog/zstd v1.5.7
 	github.com/HdrHistogram/hdrhistogram-go v1.1.2
 	github.com/RaduBerinde/axisds v0.0.0-20260105221726-1be486564c85
+	github.com/RaduBerinde/tdigest v0.0.0-20251022152254-90e030c3a314
 	github.com/cespare/xxhash/v2 v2.2.0
 	github.com/cockroachdb/crlib v0.0.0-20251122031428-fe658a2dbda1
 	github.com/cockroachdb/datadriven v1.0.3-0.20251123150250-ddff6747b112
--- a/go.sum
+++ b/go.sum
@@ -11,6 +11,8 @@ github.com/RaduBerinde/axisds v0.0.0-20260105221726-1be486564c85 h1:OjUilXBeTEMa
 github.com/RaduBerinde/axisds v0.0.0-20260105221726-1be486564c85/go.mod h1:gAhfbTwNsV064J6aon3TDJLzeh5IxfptXalQVnjjnWQ=
 github.com/RaduBerinde/btreemap v0.0.0-20260105202824-d3184786f603 h1:fSdiBlO4Bad28mJOPlAynvfgdDC9v+yRlzSFHvvjKYI=
 github.com/RaduBerinde/btreemap v0.0.0-20260105202824-d3184786f603/go.mod h1:0tr7FllbE9gJkHq7CVeeDDFAFKQVy5RnCSSNBOvdqbc=
+github.com/RaduBerinde/tdigest v0.0.0-20251022152254-90e030c3a314 h1:RcSNrxDZ1ZyEfpGwrpWqd3YWfMjQbKOtT+p3Wht6XN0=
+github.com/RaduBerinde/tdigest v0.0.0-20251022152254-90e030c3a314/go.mod h1:RClmoWh7JPAzCuExyOvKDCRVYGz3D11DcpzEkq4N3RA=
 github.com/aclements/go-moremath v0.0.0-20210112150236-f10218a38794 h1:xlwdaKcTNVW4PtpQb8aKA4Pjy0CdJHEqvFbAnvR5m2g=
 github.com/aclements/go-moremath v0.0.0-20210112150236-f10218a38794/go.mod h1:7e+I0LQFUI9AXWxOfsQROs9xPhoJtbsyWcjJqDd4KPY=
 github.com/aclements/go-perfevent v0.0.0-20240301234650-f7843625020f h1:JjxwchlOepwsUWcQwD2mLUAGE9aCp0/ehy6yCHFBOvo=
--- a/internal/base/internal.go
+++ b/internal/base/internal.go
@@ -672,6 +672,11 @@ type InternalKV struct {
 // The zero value is reserved to mean "no attribute" or "unknown".
 type TieringAttribute uint64
 
+// TieringSpanID is an identifier that provides a context for interpreting a
+// TieringAttribute value. At any point in time, the span policies determine
+// a set of non-overlapping key regions with distinct TieringSpanIDs.
+type TieringSpanID uint64
+
 // KVMeta describes optional metadata associated with an `InternalKV`.
 // It's currently produced only by sstable-backed iterators and is not embedded
 // within `InternalKV` to avoid overhead on the common iteration path.
@@ -683,7 +688,7 @@ type TieringAttribute uint64
 // These methods exist to support compaction-only logic (eg, `compaction.Iter`).
 // Regular iteration should use the standard methods that do not surface metadata.
 type KVMeta struct {
-	TieringSpanID    uint64
+	TieringSpanID    TieringSpanID
 	TieringAttribute TieringAttribute
 }
 
--- a/internal/testkeys/testkeys.go
+++ b/internal/testkeys/testkeys.go
@@ -555,7 +555,7 @@ func ExtractKVMeta(value []byte) base.KVMeta {
 	if e != nil {
 		panic(fmt.Sprintf("invalid tiering span in KV metadata in %q", value))
 	}
-	res.TieringSpanID = v
+	res.TieringSpanID = base.TieringSpanID(v)
 	v, e = strconv.ParseUint(m[2], 10, 64)
 	if e != nil {
 		panic(fmt.Sprintf("invalid tiering attr in KV metadata in %q", value))
--- a/sstable/colblk/data_block.go
+++ b/sstable/colblk/data_block.go
@@ -708,7 +708,7 @@ func (w *DataBlockEncoder) Add(
 		w.values.Put(value)
 	}
 	if w.columnConfig.SupportsTiering() && meta != (base.KVMeta{}) {
-		w.tieringSpanIDs.Set(w.rows, meta.TieringSpanID)
+		w.tieringSpanIDs.Set(w.rows, uint64(meta.TieringSpanID))
 		w.tieringAttributes.Set(w.rows, uint64(meta.TieringAttribute))
 	}
 	if len(ikey.UserKey) > int(w.maximumKeyLength) {
@@ -1518,7 +1518,7 @@ func (i *DataBlockIter) decodeMeta() base.KVMeta {
 		return base.KVMeta{}
 	}
 	return base.KVMeta{
-		TieringSpanID:    i.tieringSpanIDs.At(i.row),
+		TieringSpanID:    base.TieringSpanID(i.tieringSpanIDs.At(i.row)),
 		TieringAttribute: base.TieringAttribute(i.tieringAttributes.At(i.row)),
 	}
 }
--- a/sstable/tieredmeta/histogram.go
+++ b/sstable/tieredmeta/histogram.go
@@ -1,0 +1,200 @@
+// Copyright 2025 The LevelDB-Go and Pebble Authors. All rights reserved. Use
+// of this source code is governed by a BSD-style license that can be found in
+// the LICENSE file.
+
+package tieredmeta
+
+import (
+	"encoding/binary"
+
+	"github.com/RaduBerinde/tdigest"
+	"github.com/cockroachdb/pebble/internal/base"
+	"github.com/pkg/errors"
+)
+
+// digestDelta is the compression argument for t-digest, used to specify the
+// tradeoff between accuracy and memory consumption: meaning higher values give
+// more accuracy, but use more memory/space. Specifically, the digest has at most
+// 2*compression centroids, with ~1.3*compression in practice.
+//
+// A value of 100 is a common default for t-digest implementations, which should
+// provide at least ±1% quantile accuracy (and much better at the tails).
+const digestDelta = 100
+
+// StatsHistogram is a quantile sketch using t-digest to track the distribution
+// of a bytes value (with a caller-determined meaning) by TieringAttribute. It
+// can efficiently answer questions like:
+//   - What fraction of bytes have attribute <= threshold? (CDF)
+//   - What attribute value covers q% of bytes? (Quantile)
+//   - How many bytes are "cold" vs "hot"?
+//
+// The sketch is mergeable, making it easy to combine statistics from
+// multiple files without bucket alignment issues.
+type StatsHistogram struct {
+	// TotalBytes is the sum of all bytes recorded (including BytesNoAttr).
+	TotalBytes uint64
+	// TotalCount is the number of records added.
+	TotalCount uint64
+	// BytesNoAttr tracks bytes with attribute=0 separately, as this typically
+	// indicates unset/special values that shouldn't be tiered.
+	BytesNoAttr uint64
+	// digest is the t-digest for non-zero attributes (read-only).
+	digest tdigest.TDigest
+}
+
+// CDF returns the fraction of non-zero bytes with attribute <= threshold.
+// Returns a value in [0, 1].
+// Example: CDF(coldThreshold) tells you what fraction of bytes are "cold".
+func (s *StatsHistogram) CDF(threshold base.TieringAttribute) float64 {
+	if s.BytesWithAttr() == 0 {
+		return 0
+	}
+	return s.digest.CDF(float64(threshold))
+}
+
+// Quantile returns the attribute value at quantile ~q (where q is in [0, 1]) of
+// bytes with a tiering attribute set.
+// Example: Quantile(0.5) returns a value close to the median value.
+func (s *StatsHistogram) Quantile(q float64) base.TieringAttribute {
+	return base.TieringAttribute(s.digest.Quantile(q))
+}
+
+// BytesWithAttr returns bytes that have a tiering attribute set, excluding
+// bytes with attr=0 (which typically indicates unset/special values).
+func (s *StatsHistogram) BytesWithAttr() uint64 {
+	return s.TotalBytes - s.BytesNoAttr
+}
+
+// BytesBelowThreshold estimates how many bytes (excluding NoAttrBytes) have
+// attribute <= threshold. This is useful for tiering decisions.
+func (s *StatsHistogram) BytesBelowThreshold(threshold base.TieringAttribute) uint64 {
+	return uint64(s.CDF(threshold) * float64(s.BytesWithAttr()))
+}
+
+// BytesAboveThreshold estimates how many non-zero bytes have attribute > threshold.
+// This is useful for tiering decisions.
+func (s *StatsHistogram) BytesAboveThreshold(threshold base.TieringAttribute) uint64 {
+	return s.BytesWithAttr() - s.BytesBelowThreshold(threshold)
+}
+
+// encode serializes the histogram to bytes. The encoding format is:
+//
+//	<total bytes> <total count> <zero count> <digest size> <digest data>
+func (s *StatsHistogram) encode() []byte {
+	// Calculate the size needed for the t-digest.
+	digestSize := s.digest.SerializedSize()
+
+	// Pre-allocate with estimated capacity (4 uvarints + digest).
+	// Each uvarint is at most 10 bytes for uint64.
+	buf := make([]byte, 0, 4*binary.MaxVarintLen64+digestSize)
+
+	buf = binary.AppendUvarint(buf, s.TotalBytes)
+	buf = binary.AppendUvarint(buf, s.TotalCount)
+	buf = binary.AppendUvarint(buf, s.BytesNoAttr)
+	buf = binary.AppendUvarint(buf, uint64(digestSize))
+	buf = s.digest.Serialize(buf)
+
+	return buf
+}
+
+// DecodeStatsHistogram decodes a StatsHistogram from the provided buffer.
+// The encoding format is:
+//
+//	<total bytes> <total count> <zero bytes> <digest size> <digest data>
+func DecodeStatsHistogram(buf []byte) (StatsHistogram, error) {
+	var h StatsHistogram
+	var n int
+
+	h.TotalBytes, n = binary.Uvarint(buf)
+	if n <= 0 {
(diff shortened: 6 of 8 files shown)


annrpom:
I'll write the block in a separate PR

RaduBerinde (commented):
Nice!!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5650#-:-OgYiuYa19PpuGlgs91Y:bwuv9uq)* status: 0 of 8 files reviewed, 14 unresolved discussions (waiting on @annrpom, @jbowens, and @sumeerbhola)

___
*[`go.sum` line 14 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYj0f77nKz6gNE0bA5:-OgYj0f77nKz6gNE0bA6:byjdwed) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/go.sum#L14)):*
> ```
> github.com/RaduBerinde/btreemap v0.0.0-20250419174037-3d62b7205d54 h1:bsU8Tzxr/PNz75ayvCnxKZWEYdLMPDkUgticP4a4Bvk=
> github.com/RaduBerinde/btreemap v0.0.0-20250419174037-3d62b7205d54/go.mod h1:0tr7FllbE9gJkHq7CVeeDDFAFKQVy5RnCSSNBOvdqbc=
> github.com/RaduBerinde/tdigest v0.0.0-20251022152254-90e030c3a314 h1:RcSNrxDZ1ZyEfpGwrpWqd3YWfMjQbKOtT+p3Wht6XN0=
> ```

We should apply the same due dilligence as any external library since it's a forked library and my changes to it weren't reviewed. Maybe @sumeerbhola or @jbowens can take a brief look over it?
___
*[`sstable/tieredmeta/histogram.go` line 20 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYi-j83xMOuzSGu9Z0:-OgYi-j83xMOuzSGu9Z1:bjoq91h) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram.go#L20)):*
> ```Go
> // 2*compression centroids, with ~1.3*compression in practice.
> //
> // A value of 100 is a common default for t-digest implementations.
> ```

, which should provide at least ±1% quantile accuracy (and much better at the tails).
___
*[`sstable/tieredmeta/histogram.go` line 21 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYhw754yLTqodh__jw:-OgYhw754yLTqodh__jx:b-gygw36) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram.go#L21)):*
> ```Go
> //
> // A value of 100 is a common default for t-digest implementations.
> const CompressionLevel = 100
> ```

[nit] digestDelta
___
*[`sstable/tieredmeta/histogram.go` line 24 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYfnbM9L6aAvoK9kjT:-OgYfnbM9L6aAvoK9kjU:b-7499yx) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram.go#L24)):*
> ```Go
> 
> // StatsHistogram is a quantile sketch using t-digest to track the distribution
> // of bytes by TieringAttribute. It can efficiently answer questions like:
> ```

[nit] for bytes we could say "a bytes value (with a caller-determined meaning)"
___
*[`sstable/tieredmeta/histogram.go` line 38 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYfxBhASS7vDjluq8N:-OgYfxBhASS7vDjluq8O:b-wqbnex) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram.go#L38)):*
> ```Go
> 	// ZeroBytes tracks bytes with attribute=0 separately, as this typically
> 	// indicates unset/special values that shouldn't be tiered.
> 	ZeroBytes uint64
> ```

[nit] BytesNoAttr
___
*[`sstable/tieredmeta/histogram.go` line 54 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYg3aZ-OfFtZf86ZJY:-OgYg3aZ-OfFtZf86ZJZ:bwpgehr) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram.go#L54)):*
> ```Go
> }
> 
> // Quantile returns the attribute value at quantile q (where q is in [0, 1]).
> ```

[nit] `at quantile ~q`
___
*[`sstable/tieredmeta/histogram.go` line 55 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYg6_f7o5cR-vwDemV:-OgYg6_f7o5cR-vwDemW:b-8hjdl6) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram.go#L55)):*
> ```Go
> 
> // Quantile returns the attribute value at quantile q (where q is in [0, 1]).
> // Example: Quantile(0.5) returns the median attribute value by bytes.
> ```

[nit] returns a value close to the median value
___
*[`sstable/tieredmeta/histogram.go` line 64 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYg-B3-G1UvxB7CJYe:-OgYg-B3-G1UvxB7CJYf:b-1gu36r) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram.go#L64)):*
> ```Go
> 
> // NonZeroBytes returns the total bytes with non-zero attributes.
> func (s *StatsHistogram) NonZeroBytes() uint64 {
> ```

BytesWithAttr
___
*[`sstable/tieredmeta/histogram.go` line 68 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYgBDLEtQ7yQe8GzbJ:-OgYgBDLEtQ7yQe8GzbK:b-o62in4) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram.go#L68)):*
> ```Go
> }
> 
> // BytesBelowThreshold estimates how many non-zero bytes have attribute <= threshold.
> ```

how many bytes (excluding NoAttrBytes)
___
*[`sstable/tieredmeta/histogram_block.go` line 49 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYkQOSBKYxaGjLvEG0:-OgYkQOSBKYxaGjLvEG1:b-71x3gj) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram_block.go#L49)):*
> ```Go
> // Below is the block structure:
> //
> //	SSTable's TieringHistogramBlock:
> ```

IMO in sstables we should be writing only one tiering histogram, the one that matches the TieringID in the current policy. We can consider any KVs with other TierIDs as having "no attribute" when we populate this histogram. I don't think we would realistically see different TierIDs in the same keyspace anyway. WDYT @sumeerbhola?
___
*[`sstable/tieredmeta/histogram_block.go` line 58 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYjh5C6aQnKe8O2OZ6:-OgYjh5C6aQnKe8O2OZ7:bvzn82d) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram_block.go#L58)):*
> ```Go
> //	└── ...
> //
> //	Blob file's TieringHistogramBlock:
> ```

@sumeerbhola remind me, when do we expect to use a blob's histogram block? I was thinking everything would be done from the sstable's perspective.
___
*[`sstable/tieredmeta/histogram_block.go` line 92 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYhTin1bCf0ZgMOp0s:-OgYhTin1bCf0ZgMOp0t:b-d8i67d) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram_block.go#L92)):*
> ```Go
> 	SSTableKeyBytes KindAndTier = iota
> 	// SSTableValueBytes is the histogram for values stored inside the
> 	// sstable. The tier is implicit based on whether the sstable is in hot or
> ```

[nit] I'd remove the last sentence since all sstables are in hot storage currently.
___
*[`sstable/tieredmeta/histogram_block.go` line 94 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYhdxs6fOAuOOH4TrB:-OgYhdxt6Ntupx_hKeH9:b-xlfkq5) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram_block.go#L94)):*
> ```Go
> 	// sstable. The tier is implicit based on whether the sstable is in hot or
> 	// cold storage.
> 	SSTableValueBytes
> ```

SSTableInPlaceValueBytes
___
*[`sstable/tieredmeta/histogram_block.go` line 97 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYh2h6-fyxlOMqBisV:-OgYh2h6-fyxlOMqBisW:bqy79ee) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram_block.go#L97)):*
> ```Go
> 	// SSTableBlobReferenceHotBytes is the histogram for blob references from the
> 	// sstable to hot tier blob files. The size is the size of the value.
> 	SSTableBlobReferenceHotBytes
> ```

I don't think we need `SSTableBlobReferenceHotBytes` and `SSTableBlobReferenceColdBytes`. We should just have `SSTableBlobValueBytes` which tracks the blob value length.

Separately (in the sstable properties and/or TableMetadata), we will store the cutoff that was used, so we'll be able to use `BytesBelowThreshold` and `BytesAboveThreshold` to figure out cold vs hot.

sumeerbhola (changes requested):
> @sumeerbhola reviewed 8 files and all commit messages, and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5650)* status: all files reviewed, 8 unresolved discussions (waiting on @annrpom, @jbowens, and @RaduBerinde).

___
*[`sstable/tieredmeta/histogram_block.go` line 90 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYyA-I774WLM9JLkSp:-OgldQE5AkfpjdZuPtMX:b-s38gl1) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram_block.go#L90)):*

would we ever be way off if we were to track just the KV count? in practice, would we have some large variance of value bytes across an sstable that would cause us to severely overestimate or underestimate? maybe like: if we wrote some data, some time passes so that data turns cold, and then we write mvcc tombstones that have no value - won't we then underestimate cold data, since count-based tracking weights tombstones (no value, hot) equally with real KVs (large values, cold)?
</blockquote></details>

I'm also confused because the sstable physical size includes inline values.
Say we tracked the original non prefix compressed bytes in the histogram. And we also computed the sum of the key columns in all the data blocks in the sstable (say B). Then the estimate for cold key bytes with a cold threshold T could be (BytesBelowThreshold(T)/TotalBytes)*B.
___
*[`sstable/tieredmeta/histogram_block.go` line 72 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5650#-Oglf5U56ONz2IMk04jV:-Oglf5U56ONz2IMk04jW:b-g2s4yz) ([raw file](https://github.com/cockroachdb/pebble/blob/1cd44c222f79d41a1cbb48d8b9435f414cddf5f9/sstable/tieredmeta/histogram_block.go#L72)):*
> ```Go
> 	// SSTableBlobValueBytes is the histogram for values stored in blob files
> 	// referenced by the sstable.
> 	SSTableBlobValueBytes
> ```

Now that some files will have hot references also have cold references, I was trying to think through what we need here and for what purpose. I was imagining the following categories:
- Value that has both cold reference and hot (in place in sstable, or reference): Useful for observability, for figuring out how much duplication we have. Not used for deciding whether to write cold blob files in a compaction.
- Value that has only cold reference: Again, only for observability
- Value that is only hot: Needed for deciding whether to write cold references before starting the compaction. The prototype further distinguished the ones that were a reference versus in place since the former definitely met the value size threshold to be a reference, while the latter could be very small. So for the latter, there was an estimate used in the condition `k.KindAndTier == base.SSTableValueBytes && h.MeanSize() < uint64(policy.MinimumSize)`

The list in this PR is different than the above.
___
*[`sstable/tieredmeta/histogram.go` line 34 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgiKUy8Ba02iHDK0_Qo:-OgiKUy8Ba02iHDK0_Qp:b7uw5md) ([raw file](https://github.com/cockroachdb/pebble/blob/1cd44c222f79d41a1cbb48d8b9435f414cddf5f9/sstable/tieredmeta/histogram.go#L34)):*
> ```Go
> // multiple files without bucket alignment issues.
> type StatsHistogram struct {
> 	// TotalBytes is the sum of all bytes recorded (including ZeroBytes).
> ```

should this say BytesNoAttr instead of ZeroBytes?
___
*[`sstable/tieredmeta/histogram.go` line 56 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgiL91d-OT8nlk4h4XI:-OgiL91d-OT8nlk4h4XJ:b-gizatj) ([raw file](https://github.com/cockroachdb/pebble/blob/1cd44c222f79d41a1cbb48d8b9435f414cddf5f9/sstable/tieredmeta/histogram.go#L56)):*
> ```Go
> }
> 
> // Quantile returns the attribute value at quantile ~q (where q is in [0, 1]).
> ```

worth saying the quantile is of the non-zero bytes.
___
*[`sstable/tieredmeta/histogram.go` line 146 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgiLqQKE35ol8lOF96t:-OgiLqQKE35ol8lOF96u:b-gg7zag) ([raw file](https://github.com/cockroachdb/pebble/blob/1cd44c222f79d41a1cbb48d8b9435f414cddf5f9/sstable/tieredmeta/histogram.go#L146)):*
> ```Go
> 			digestSize, len(buf))
> 	}
> ```

is it ok to have `digestSize < len(buf)` or should that also be an error?

RaduBerinde (commented):
> @RaduBerinde made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5650)* status: all files reviewed, 8 unresolved discussions (waiting on @annrpom, @jbowens, and @sumeerbhola).

___
*[`sstable/tieredmeta/histogram_block.go` line 90 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgYyA-I774WLM9JLkSp:-OgmGe6QD2QooQJLripp:b-pvnbwz) ([raw file](https://github.com/cockroachdb/pebble/blob/9fce22e498b41033954dfc211c2744eac9b11bd8/sstable/tieredmeta/histogram_block.go#L90)):*
I forgot about the in-place values. Indeed we would need a size that excludes that.

> And we also computed the sum of the key columns in all the data blocks in the sstable (say B)

This would be the size of the encoded column as it appears in the block? That does make sense. Though I'm not sure if scaling according to the sum of key sizes is necessarily better than just by key count.

With the key count, we could also estimate the amount of other per-key data in the block (the tiering attribute, the blob reference).

I am suspecting that in many cases, the extra key storage itself might only be a few bytes per key (e.g. last few bytes of an ID), much less than the other per-key data in the sstable.
___
*[`sstable/tieredmeta/histogram_block.go` line 72 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5650#-Oglf5U56ONz2IMk04jV:-OgmIKGA5JNXwz4rOn8h:b-2cvnpf) ([raw file](https://github.com/cockroachdb/pebble/blob/1cd44c222f79d41a1cbb48d8b9435f414cddf5f9/sstable/tieredmeta/histogram_block.go#L72)):*

Now that some files will have hot references also have cold references, I was trying to think through what we need here and for what purpose. I was imagining the following categories:
- Value that has both cold reference and hot (in place in sstable, or reference): Useful for observability, for figuring out how much duplication we have. Not used for deciding whether to write cold blob files in a compaction.
- Value that has only cold reference: Again, only for observability
- Value that is only hot: Needed for deciding whether to write cold references before starting the compaction. The prototype further distinguished the ones that were a reference versus in place since the former definitely met the value size threshold to be a reference, while the latter could be very small. So for the latter, there was an estimate used in the condition `k.KindAndTier == base.SSTableValueBytes && h.MeanSize() < uint64(policy.MinimumSize)`

The list in this PR is different than the above.
</blockquote></details>

I think we should choose either a design where (for the levels where we actually do tiering) all hot values are both in hot and cold, or one where all hot values are only in hot. Mixing them will lead to a lot of complications.

With that in mind, these can all be deduced from `BlobValueBytes` and the threshold used for this file.

IMO upper level files that aren't tiered wouldn't have this data at all. The cost-benefit is not there - these are normally `~1%` of total size so statistics are not very valuable, but they get written much more frequently. At the same time, a single upper-level file can have keys from many different tables and we'd have to look up all each table's span config in every compaction.

annrpom (commented):
> @annrpom made 6 comments and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5650)* status: 3 of 8 files reviewed, 8 unresolved discussions (waiting on @jbowens, @RaduBerinde, and @sumeerbhola).

___
*[`sstable/tieredmeta/histogram.go` line 43 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgnvXTa4vyhhloWKvqR:-OiDsYPgAZrlVqQlz843:bjwsqz3) ([raw file](https://github.com/cockroachdb/pebble/blob/1cd44c222f79d41a1cbb48d8b9435f414cddf5f9/sstable/tieredmeta/histogram.go#L43)):*

Are there any situations where we would want this to be nil? If not, we should make it a non-pointer
</blockquote></details>

made it a non-pointer
___
*[`sstable/tieredmeta/histogram_block.go` line 72 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5650#-Oglf5U56ONz2IMk04jV:-OiEhxJx5brb9Gx2EYfn:b-6v7dre) ([raw file](https://github.com/cockroachdb/pebble/blob/1cd44c222f79d41a1cbb48d8b9435f414cddf5f9/sstable/tieredmeta/histogram_block.go#L72)):*

Hm, yes. I'm worried about overhead of updating too many histograms with every KV.

1. We could remove `SSTableKeyBytes` and replace it with just a per-sstable percentage of data bytes that are due to keys that have tiering attribute below the threshold at the time of compaction? I don't see what the full histogram is buying us.

2. Do we need to distinguish in-place values vs hot blob references? If not, we could have `SSTableHotOnlyValueBytes`, `SSTableColdOnlyValueBytes` and `SSTableHotAndColdValueBytes` where each KV updates only one of the three.
</blockquote></details>

for (1), done

for (2), we do have a comment from the prototype that states that distinguishing in-place values (vs hot blob references) allows us to make a more informed decision on whether or not to rewrite blob references:

```
// Additionally, we want to
// know which values are inside the sstable and which are blob references, since
// if most of the hot => cold transitions are for values inside the sstable, we
// may choose to not rewrite the blob references since the gain is small.
```
so i'm thinking something like `SSTableInPlaceValueBytes`,  `SSTableBlobReferenceHotBytes`, `SSTableBlobReferenceColdBytes` + have some per-sstable data bytes that are due to keys that have tiering + some per-sstable data bytes that are due to values being `HotAndCold`. i can't think of a reason why we need `HotAndCold` to be a histogram. i could be missing something
___
*[`sstable/tieredmeta/histogram_block.go` line 76 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgnuMqOB8ed7NdVeYTb:-OiE0C_C1H0LEgK2c1XF:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/1cd44c222f79d41a1cbb48d8b9435f414cddf5f9/sstable/tieredmeta/histogram_block.go#L76)):*

[nit] let's just say is instantiated for each output file that requires tiering histograms.
</blockquote></details>

done
___
*[`sstable/tieredmeta/histogram_block.go` line 111 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgnvklCDsgiUtofevN-:-OiE0cP9DW3h6u6z-s-3:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/1cd44c222f79d41a1cbb48d8b9435f414cddf5f9/sstable/tieredmeta/histogram_block.go#L111)):*

[nit] `MakeTieringHistogramBlockWriter() TieringHistogramBlockWriter` (too many pointers)
</blockquote></details>

done
___
*[`sstable/tieredmeta/histogram_block.go` line 130 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgnuW3C2HFsPKQTXXe9:-OiEfkwq2CjmNioieVaQ:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/1cd44c222f79d41a1cbb48d8b9435f414cddf5f9/sstable/tieredmeta/histogram_block.go#L130)):*

[nit] `Finish()`. Should mention if the writer can be reused after that.
</blockquote></details>

done
___
*[`sstable/tieredmeta/histogram_block.go` line 150 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5650#-OgnuseP65urDz1H70b7:-OiEh7-y4UaU-R40MDxg:bpze368) ([raw file](https://github.com/cockroachdb/pebble/blob/1cd44c222f79d41a1cbb48d8b9435f414cddf5f9/sstable/tieredmeta/histogram_block.go#L150)):*

[nit] the comment is likely to become stale, not sure if we'll ever load it into TableMetadata. I'd remove the second sentence. We could also just return `map[Key]StatsHistogram` since it's pretty self-explanatory.
</blockquote></details>

changed it to just return the map

annrpom:
TFTRs! ('-')7

merging for now; we can always amend as our thinking evolves

PR #5651, opened 2025-12-12
block: allow more slack when fitting blocks to size classes

RaduBerinde:
We currently take into account the in-memory metadata when trying to
fit blocks into size classes (when we create new blocks).

Growing the size of the metadata is problematic: all blocks written by
the previous release(s) will cause significant fragmentation, as they
will require an allocation a few bytes larger than a size class.
Typical data blocks of 32KiB will take up 40KiB (20% fragmentation).

This change adds some more slack into this fitting, leaving out
384-312=72 extra bytes for future growth of the metadata. This is only
an extra 0.2% of the typical data block size of 32KiB.

### DIFF
--- a/sstable/block/block.go
+++ b/sstable/block/block.go
@@ -250,6 +250,14 @@ const MetadataSize = 280
 // block data buffer aligned.
 const _ uint = -(MetadataSize % 8)
 
+// TODO(radu): increase MetadataSize so that the total overhead is a multiple of
+// 64 bytes (the typical cache line size). Some blocks, like bloom filters,
+// benefit from 64-byte alignment of data.
+//
+// We can only do this a few releases after introducing some slack into the
+// AllocationOverheadAllowance.
+// const _ uint = -((cache.ValueMetadataSize + MetadataSize) % 64)
+
 // NoReadEnv is the empty ReadEnv which reports no stats and does not use a
 // buffer pool.
 var NoReadEnv = ReadEnv{}
--- a/sstable/block/flush_governor.go
+++ b/sstable/block/flush_governor.go
@@ -34,21 +34,28 @@ type FlushGovernor struct {
 	targetBoundary int
 }
 
-// This value is the amount of extra bytes we allocate together with the block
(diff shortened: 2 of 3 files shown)


RaduBerinde:
I think we should backport this to 26.1 ASAP, and then later backport to other releases once it's been merged to master for at least a few weeks. Let me know if that sounds good.

RaduBerinde:
*[`sstable/block/block.go` line 255 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5651#-OgNz_VbEcQCFRDlGX64:-OgPHUx_8g_J38aegS9b:b-3y6i7g) ([raw file](https://github.com/cockroachdb/pebble/blob/d213d7af2fe183930d2dfbb77ba891498e638d08/sstable/block/block.go#L255)):*

We want the sum `cache.ValueMetadataSize + block.MetadataSize` to be a multiple of 64. It is currently 312. So the increase would only be 8 bytes (plus the 8 bytes would be usable for the metadata structures).
</blockquote></details>

That said, 448 sounds good to me if you prefer it.

sumeerbhola (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5651#-:-OgYKhZT8h0u9xXi35S6:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @jbowens and @RaduBerinde)

___
*[`sstable/block/block.go` line 255 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5651#-OgNz_VbEcQCFRDlGX64:-OgYKPvYBW6G4ii4WBdF:b-lewzqu) ([raw file](https://github.com/cockroachdb/pebble/blob/d213d7af2fe183930d2dfbb77ba891498e638d08/sstable/block/block.go#L255)):*

That said, 448 sounds good to me if you prefer it.
</blockquote></details>

Oh, ok. I misunderstood that we would still have 64 bytes of slack. That seems sufficient, so 384 is fine.

RaduBerinde:
TFTR!

PR #5652, opened 2025-12-12
metamorphic: fix level options randomization

RaduBerinde:
The code that randomizes the options for metamorphic tests assumes
that a later `EnsureDefaults` will copy the L0 level options to all
other levels. However, most of the time, the levels are populated are
already populated because of an earlier `parseOptions` call and we end
up modifying only the L0 options.

Fix this by copying the options to all levels. We also change the
compression randomization so it always happens through
`ApplyCompressionSettings`.


RaduBerinde:
TFTR!

PR #5656, opened 2025-12-15
db: don't split tables in delete compactions

jbowens:
Previously a delete-only compaction was allowed to excise a sstable into two parts, as long as the deleteCompactionHint dropped an entire sstable elsewhere in the LSM (i.e., the hint's overall effect on the number of tables needed to be net zero). This commit alters the logic to never split a sstable in two and to only pursue an excise during a delete compaction if the excise would simply shorten one of the table's two bounds. This logic is simpler and avoids performing an excise just to remove a potentially negligible portion of a table's data.

### DIFF
--- a/compaction_delete.go
+++ b/compaction_delete.go
@@ -213,20 +213,26 @@ func (h *deleteCompactionHint) canDeleteOrExcise(
 	default:
 		panic(errors.AssertionFailedf("pebble: unknown delete compaction key type: %s", h.keyType))
 	}
-	if h.bounds.ContainsBounds(cmp, m.UserKeyBounds()) {
+	tableBounds := m.UserKeyBounds()
+	if h.bounds.ContainsBounds(cmp, tableBounds) {
 		return hintDeletesFile
 	}
 	if !exciseEnabled {
 		// The file's keys must be completely contained within the hint range; excises
 		// aren't allowed.
 		return hintDoesNotApply
+	} else if !h.bounds.Overlaps(cmp, tableBounds) {
+		// Disjoint.
+		return hintDoesNotApply
+	} else if cmp(tableBounds.Start, h.bounds.Start) < 0 &&
+		tableBounds.End.CompareUpperBounds(cmp, h.bounds.End) > 0 {
+		// If the table's bounds completely contain the hint, applying the hint
+		// would cut the table into two virtual sstables, increasing the number
+		// of tables in the level. We only pursue excises if they shorten an
+		// existing table's bounds.
+		return hintDoesNotApply
 	}
-	// Check for any overlap. In cases of partial overlap, we can excise the part of the file
-	// that overlaps with the deletion hint.
-	if h.bounds.Overlaps(cmp, m.UserKeyBounds()) {
-		return hintExcisesFile
-	}
-	return hintDoesNotApply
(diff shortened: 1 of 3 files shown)


RaduBerinde:
*[`compaction_delete.go` line 225 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5656#-OgY-eHsBO16EUixAPY-:-OgY1bR-59fnb7oCROtU:b-mzjz2d) ([raw file](https://github.com/cockroachdb/pebble/blob/4405e8773c6141483b098d2db4b96f06968c5fb5/compaction_delete.go#L225)):*

Hrm, I'm realizing ContainsBounds isn't exactly what I want. If the bounds are exactly equal, it'll still return true. But we would want to excise in that case.
</blockquote></details>

That was already handled by `if h.bounds.ContainsBounds(cmp, tableBounds)` above.

RaduBerinde:
*[`compaction_delete.go` line 225 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5656#-OgY-eHsBO16EUixAPY-:-OgY1o6QDnMlQdj7b33y:b-do3t2w) ([raw file](https://github.com/cockroachdb/pebble/blob/4405e8773c6141483b098d2db4b96f06968c5fb5/compaction_delete.go#L225)):*

That was already handled by `if h.bounds.ContainsBounds(cmp, tableBounds)` above.
</blockquote></details>

Oh, wait, I understand what you mean. We should check that `h.bounds` contains either the start or the end endpoint.

RaduBerinde:
*[`compaction_delete.go` line 225 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5656#-OgY-eHsBO16EUixAPY-:-OgY278868vHQZp0vyzc:b-k34htc) ([raw file](https://github.com/cockroachdb/pebble/blob/4405e8773c6141483b098d2db4b96f06968c5fb5/compaction_delete.go#L225)):*

Oh, wait, I understand what you mean. We should check that `h.bounds` contains either the start or the end endpoint.
</blockquote></details>

We can split the `Overlaps` call into:
```
if h.bounds.ContainsUserKey(tableBounds.Start)
```
and
```
if h.bounds.End.CompareUpperBounds(tableBounds.End) >= 0 && tableBounds.End.IsUpperBoundFor(h.bounds.Start)
```

RaduBerinde:
*[`compaction_delete.go` line 225 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5656#-OgY-eHsBO16EUixAPY-:-OgY512L18u4XFREiMnW:be2rxsk) ([raw file](https://github.com/cockroachdb/pebble/blob/4405e8773c6141483b098d2db4b96f06968c5fb5/compaction_delete.go#L225)):*

We can split the `Overlaps` call into:
```
if h.bounds.ContainsUserKey(tableBounds.Start)
```
and
```
if h.bounds.End.CompareUpperBounds(tableBounds.End) >= 0 && tableBounds.End.IsUpperBoundFor(h.bounds.Start)
```
</blockquote></details>

I'm trying to think what low-level API would we need to make this clearer/simpler.
Maybe `a.Intersection(b) (_ UserKeyBounds, startChanged, endChanged, ok bool)`?

PR #5658, opened 2025-12-15
db: provide blob files as objstorage.Readables to ingestLoad1

xinhaoz:
Necessary to support blobs for WAL replay.


xinhaoz:
TFTR!

Issue #4883, opened 2025-06-11
db: hints from cancelled delete-only compactions lost


jbowens:
When a delete-only compaction is cancelled due to a conflicting excise, the hints that created the compaction are lost. This could cause us to fail to pursue an efficient delete-only compaction.

https://cockroachlabs.slack.com/archives/CAC6K3SLU/p1749674202146049?thread_ts=1749672645.831109&cid=CAC6K3SLU

PR #5667, opened 2025-12-19
db: refactor delete-only compactions

jbowens:
This commit refactors the implementation of delete-only compactions to simplify
logic and resolve a bug (https://github.com/cockroachdb/pebble/issues/4883) where an unfortunately timed excise could
cancel a delete-only compaction, preventing the delete-only compaction from
ever executing.

Previously, delete-only compaction picking scheduled large compactions deleting
and excising as many tables as possible from a set of resolved 'hints.' This
could result in large delete-only compactions with wide bounds, making
compaction cancellation more likely during bulk removal like a DROP TABLE
command in CockroachDB. Because the 'hints' were resolved during picking, the
knowledge of the wide tombstones that originated the compaction was lost and no
replacement delete-only compaction was scheduled.

This commit extracts the externalized, in-memory tracking of "wide tombstones"
into a new internal/tombspan package. The table stats collector propagates
knowledge of these wide tombstones to a [tombspan.Set]. When a wide tombstone
falls into the final snapshot stripe, knowledge of the tombstone is
incorporated into a region tree. Compaction picking consults the tombspan.Set.
Rather than large, multi-level and multi-file delete-only compactions,
compaction-picking now picks compactions that delete or excise 1 table at a
time.

These smaller delete-only compactions are less likely to be cancelled, complete
more quickly, and are more understandable within the event log. The
[tombspan.Set] only forgets a tombstoned span once it observes that there are
no tables eligible for a delete-only compaction within the span. If eligible
tables are currently compacting, the span is preserved, ensuring knowledge of
the tombstone survives compaction cancellation.

Fix https://github.com/cockroachdb/pebble/issues/4883.

### DIFF
--- a/compaction.go
+++ b/compaction.go
@@ -25,6 +25,7 @@ import (
 	"github.com/cockroachdb/pebble/internal/manifest"
 	"github.com/cockroachdb/pebble/internal/problemspans"
 	"github.com/cockroachdb/pebble/internal/sstableinternal"
+	"github.com/cockroachdb/pebble/internal/tombspan"
 	"github.com/cockroachdb/pebble/objstorage"
 	"github.com/cockroachdb/pebble/objstorage/objstorageprovider/objiotracing"
 	"github.com/cockroachdb/pebble/sstable"
@@ -317,15 +318,7 @@ type tableCompaction struct {
 	// compaction that does not merge or write sstables. Instead, it only
 	// performs deletions either through removing whole sstables from the LSM or
 	// virtualizing them into virtual sstables.
-	deleteOnly struct {
-		// hints are collected by the table stats collector and describe range
-		// deletions and the files containing keys deleted by them.
-		hints []deleteCompactionHint
-		// exciseEnabled is set to true if this compaction is allowed to excise
-		// files. If false, the compaction will only remove whole sstables that
-		// are wholly contained within the bounds of range deletions.
-		exciseEnabled bool
-	}
+	deleteOnly tombspan.DeleteOnlyCompaction
 	// flush contains information specific to flushes (compactionKindFlush and
 	// compactionKindIngestedFlushable). A flush is modeled by a compaction
 	// because it has similar mechanics to a default compaction.
@@ -1988,7 +1981,7 @@ func (d *DB) maybeScheduleCompaction() {
 	// Delete-only compactions are expected to be cheap and reduce future
 	// compaction work, so schedule them directly instead of using the
 	// CompactionScheduler.
-	if d.tryScheduleDeleteOnlyCompaction() {
+	for d.tryScheduleDeleteOnlyCompaction() {
 		env.inProgressCompactions = d.getInProgressCompactionInfoLocked(nil)
 		d.mu.versions.pickedCompactionCache.invalidate()
 	}
@@ -2686,7 +2679,7 @@ func (d *DB) runCompaction(
 	}
 	switch c.kind {
 	case compactionKindDeleteOnly:
-		return d.runDeleteOnlyCompaction(c, d.mu.snapshots.toSlice())
+		return d.runDeleteOnlyCompaction(c)
 	case compactionKindMove:
 		return d.runMoveCompaction(jobID, c)
 	case compactionKindCopy:
--- a/compaction_delete.go
+++ b/compaction_delete.go
@@ -5,13 +5,7 @@
 package pebble
 
 import (
-	"bytes"
-	stdcmp "cmp"
 	"context"
-	"fmt"
-	"maps"
-	"slices"
-	"sort"
 	"time"
 
 	"github.com/cockroachdb/errors"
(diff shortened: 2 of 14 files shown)


jbowens:
I've got some more file number nondeterminism to fight in the tests 😧

jbowens (commented):
> @jbowens made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5667)* status: 0 of 14 files reviewed, 5 unresolved discussions (waiting on @RaduBerinde).

___
*[`table_stats.go` line 493 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5667#-OgxqZbiBBPrBsHEN-bN:-OiDrVrk0EbreeEnwSB2:bc8fjf6) ([raw file](https://github.com/cockroachdb/pebble/blob/3de793ab9303a22fb364294666d5bb21d61de512/table_stats.go#L493)):*

It's kind of strange that we are considering each fragment separately. For example, a `[a,z)#4` tombstone by itself would delete a `[b, v]` table, but then if we also have a `[d,f)#5` tombstone, we get fragments `[a, d) [d, f) [f z)` neither of which covers the table. Maybe we should call out this limitation in a comment.
</blockquote></details>

The spans used for the deletion candidates are defragmented _within the context of the sstable itself_. It's true that if the tombstone fragments get split into separate tables, we won't notice that the defragmented union of the fragments makes a table eligible for a delete-only compaction.

Within a table, we defragment the spans during iteration (in newCombinedDeletionKeyspanIter), but we only defragment spans that have the same key-types deleted. This still is lossy (eg, you might have rangedels deleting [a, b) [b, c) [c, d) and a rangekeydel deleting [b,c). We will consider each of these fragments separately, which is insufficient to notice that a table [a,d) that only contains point keys is wholly deleted by the tombstones.).

There's an inherent tension between being as expansive as possible in determining deletion candidates and space and time complexity. I think we want to keep the work performed here and during compaction picking relatively lightweight, but be able to capture most cases during the broad deletions we see during table drops.
___
*[`table_stats.go` line 495 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5667#-OgxpaENBsNQmYDYg0q7:-OiDtJjS9AkTURiOz1c-:bei0ygh) ([raw file](https://github.com/cockroachdb/pebble/blob/3de793ab9303a22fb364294666d5bb21d61de512/table_stats.go#L495)):*

Given that "To simplify logic, we only act on a WideTombstone once there are no snapshots at any sequence number less than the WideTombstone's highest sequence number.", why do we ever care about anything other than the highest sequence number? What difference does having another lower-sequence tombstone for the same span make?
</blockquote></details>

The upper bound determines _when_ we can delete data, and the lower bound determines _what_ can be deleted. 

Expanded the comment on the `WideTombstone` type a bit.
___
*[`internal/tombspan/tombspan.go` line 131 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5667#-Ogxfnm9AjvHuWci8k_Y:-OiDu0lwChwUpzo1IEDF:bxsi2yg) ([raw file](https://github.com/cockroachdb/pebble/blob/3de793ab9303a22fb364294666d5bb21d61de512/internal/tombspan/tombspan.go#L131)):*

I am thoroughly confused here - the text starts out by saying that we construct a single wide tombstone from the two fragments, but then here we still treat a fragment separately? From the code it looks like the two fragments would be two separate wide tombstones?
</blockquote></details>

We construct a single wide tombstone if the tombstones are contained within the same sstable. (We defragment within a sstable, but not across sstables).
___
*[`internal/tombspan/tombspan.go` line 362 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5667#-OgxhgBt8kieo9IGo1rz:-OiDuRGr82kbMwNUr6LW:b-ii4hmk) ([raw file](https://github.com/cockroachdb/pebble/blob/3de793ab9303a22fb364294666d5bb21d61de512/internal/tombspan/tombspan.go#L362)):*

How large can `pending` become? Should we set a safety limit (with a test-only assertion)? If it can be large, maybe it should be a btree to avoid this resorting?
</blockquote></details>

Especially now that we have eventually-file only compactions (and that's all we use in Cockroach), it should typically remain very small and only grow large momentarily (for the interim between the writing of the tombstones to a sstable and the flush that converts the EFOS to file-only).

It _can_ be large momentarily, so an assertion might be tricky.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5667)* status: 0 of 14 files reviewed, 5 unresolved discussions (waiting on @jbowens).

___
*[`table_stats.go` line 495 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5667#-OgxpaENBsNQmYDYg0q7:-OiFFBqx-15Kzj7x-Au8:bhhhhg2) ([raw file](https://github.com/cockroachdb/pebble/blob/3de793ab9303a22fb364294666d5bb21d61de512/table_stats.go#L495)):*
> The upper bound determines _when_ we can delete data, and the lower bound determines _what_ can be deleted.

Wow, that is such a great way to put it, it really made things click for me.

jbowens (commented):
TFTR!

> @jbowens made 4 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5667#-:-OiYbFXwAgFe89PA5xV9:b-814g8k)* status: 0 of 14 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde).

___
*[`table_stats.go` line 495 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5667#-OgxpaENBsNQmYDYg0q7:-OiTXNgr17ejrer6HZiJ:b-j1we20) ([raw file](https://github.com/cockroachdb/pebble/blob/3de793ab9303a22fb364294666d5bb21d61de512/table_stats.go#L495)):*

> The upper bound determines _when_ we can delete data, and the lower bound determines _what_ can be deleted.

Wow, that is such a great way to put it, it really made things click for me.
</blockquote></details>

:raised_hands: 
___
*[`internal/tombspan/tombspan.go` line 455 at r8](https://reviewable.io/reviews/cockroachdb/pebble/5667#-OiIabeGEDUKIvZRbC1m:-OiYSlqT8jECZVbshCcj:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/42647ea9ee6186f24ab208964c7e654833a3c531/internal/tombspan/tombspan.go#L455)):*

[nit] The code here up to the end of the block could be moved to a separate function that returns `(_ DeleteOnlyCompaction, ok bool)`. It would make the higher-level logic flow more readable.
</blockquote></details>

Done.
___
*[`internal/tombspan/tombspan.go` line 527 at r8](https://reviewable.io/reviews/cockroachdb/pebble/5667#-OiIawSLA_3JBD9A8cxN:-OiTxO_U7ULtG8agjvsb:b-896fix) ([raw file](https://github.com/cockroachdb/pebble/blob/42647ea9ee6186f24ab208964c7e654833a3c531/internal/tombspan/tombspan.go#L527)):*

[nit] I'd consolidate these comments and do:
```
if !skippedDueToCompacting {
  clearSpans = append(clearSpans, tombBounds)
}
```
</blockquote></details>

Done.

Issue #5690, opened 2026-01-07
sstable/tieredmeta: write tiering histogram block to sstables


annrpom:
Add writing and reading support for histogram blocks in sstables.

Jira issue: PEBBLE-1321

PR #5691, opened 2026-01-07
sstable: write tiering histogram block to sstables

annrpom:
Previously, the tiering histogram structure and encoding were defined but not integrated into the sstable writing and layout. This commit adds writing and reading support for histogram blocks in columnar sstables.

Fixes: #5690

### DIFF
--- a/internal/base/internal.go
+++ b/internal/base/internal.go
@@ -687,16 +687,24 @@ type TieringSpanID uint64
 //
 // These methods exist to support compaction-only logic (eg, `compaction.Iter`).
 // Regular iteration should use the standard methods that do not surface metadata.
+//
+// The zero-value TieringAttribute indicates no metadata is available.
 type KVMeta struct {
 	TieringSpanID    TieringSpanID
 	TieringAttribute TieringAttribute
 }
 
 func (m KVMeta) String() string {
-	if m == (KVMeta{}) {
+	if !m.IsSet() {
 		return "<no meta>"
 	}
 	return fmt.Sprintf("tiering:span=%d,attr=%d", m.TieringSpanID, m.TieringAttribute)
+}
+
+// IsSet returns true if valid tiering metadata is present. Returns false if
+// TieringAttribute is 0.
+func (m KVMeta) IsSet() bool {
+	return m.TieringAttribute != 0
 }
 
 // Kind returns the KV's internal key kind.
--- a/metrics.go
+++ b/metrics.go
@@ -1256,7 +1256,7 @@ func (m *Metrics) StringForTests() string {
 
 	// We recalculate the file cache size using the 64-bit sizes, and we ignore
 	// the genericcache metadata size which is harder to adjust.
-	const sstableReaderSize64bit = 288
+	const sstableReaderSize64bit = 304
 	const blobFileReaderSize64bit = 120
 	mCopy.FileCache.Size = mCopy.FileCache.TableCount*sstableReaderSize64bit + mCopy.FileCache.BlobFileCount*blobFileReaderSize64bit
 	if math.MaxInt == math.MaxInt64 {
--- a/sstable/block/blockkind/kind.go
+++ b/sstable/block/blockkind/kind.go
@@ -16,6 +16,7 @@ const (
 	SSTableValue
 	BlobValue
 	BlobReferenceValueLivenessIndex
+	TieringHistogram
 	Filter
 	RangeDel
 	RangeKey
@@ -31,6 +32,7 @@ var kindString = [...]string{
 	SSTableValue:                    "sstval",
 	BlobValue:                       "blobval",
 	BlobReferenceValueLivenessIndex: "blobrefval",
+	TieringHistogram:                "tieringhist",
 	Filter:                          "filter",
 	RangeDel:                        "rangedel",
 	RangeKey:                        "rangekey",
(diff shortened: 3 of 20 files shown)


RaduBerinde (commented):
> @RaduBerinde made 7 comments.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5691)* status: 0 of 19 files reviewed, 7 unresolved discussions (waiting on @annrpom and @xxmplus).

___
*[`sstable/tieredmeta/histogram_block.go` line 115 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5691#-OioJ1YV2tNxHa-Cr6z3:-OioJ1YV2tNxHa-Cr6z4:bm9xxq1) ([raw file](https://github.com/cockroachdb/pebble/blob/89c96c40be7ff42b139322d39c1c33db7ba04125/sstable/tieredmeta/histogram_block.go#L115)):*
> ```Go
> func MakeTieringHistogramBlockWriter() TieringHistogramBlockWriter {
> 	return TieringHistogramBlockWriter{
> 		writers: make(map[Key]*histogramWriter),
> ```

[nit] we could initialize this lazily in Add so we avoid it in the non-tiered case
___
*[`sstable/writer_test.go` line 651 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5691#-OioK8KSAn0tLqEytdpq:-OioK8KSAn0tLqEytdpr:b-5nqlew) ([raw file](https://github.com/cockroachdb/pebble/blob/89c96c40be7ff42b139322d39c1c33db7ba04125/sstable/writer_test.go#L651)):*
> ```Go
> 
> 			tieringAttributeExtractor := func(key []byte, keyPrefixLen int, value []byte) (base.TieringAttribute, error) {
> 				if kv, ok := tieringMetadata[string(key)]; ok {
> ```

I believe this can just return 0 once we use the `meta` passed to `Add`.
___
*[`sstable/colblk_writer.go` line 420 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5691#-Oio7_QWEnAjfXZ2i0hi:-Oio7_QWEnAjfXZ2i0hj:bwxj5yl) ([raw file](https://github.com/cockroachdb/pebble/blob/89c96c40be7ff42b139322d39c1c33db7ba04125/sstable/colblk_writer.go#L420)):*
> ```Go
> 	// Track in-place value bytes in tiering histogram.
> 	if w.opts.TieringSpanIDGetter != nil && w.opts.TieringAttributeExtractor != nil {
> 		spanID := w.opts.TieringSpanIDGetter(key.UserKey)
> ```

We want to obtain these through callbacks only if they are not already passed in `meta`
___
*[`sstable/colblk_writer.go` line 425 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5691#-Oio6ua-3_ZW_pigdT-t:-Oio6ua-3_ZW_pigdT-u:bnft30u) ([raw file](https://github.com/cockroachdb/pebble/blob/89c96c40be7ff42b139322d39c1c33db7ba04125/sstable/colblk_writer.go#L425)):*
> ```Go
> 			return err
> 		}
> 		w.tieringHistogramBlock.Add(tieredmeta.SSTableInPlaceValueBytes, spanID,
> ```

I think if attribute == 0 we want to not insert it?
___
*[`sstable/colblk_writer.go` line 498 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5691#-OioHpaV5QV9gYNihFYA:-OioHpaV5QV9gYNihFYB:bixiym1) ([raw file](https://github.com/cockroachdb/pebble/blob/89c96c40be7ff42b139322d39c1c33db7ba04125/sstable/colblk_writer.go#L498)):*
> ```Go
> 
> 		// Track hot-and-cold blob reference bytes if value exists in both tiers.
> 		if w.opts.BlobValueExistsInBothTiers != nil && w.opts.BlobValueExistsInBothTiers(h.ReferenceID) {
> ```

I think we'll need to figure out some things about this hot-and-cold path (we'll need an extra column, and we'll need to somehow pass the blob handle as an argument both here and in `Add` (or maybe we'll have a new function). Feel free to remove `BlobValueExistsInBothTiers` for now and leave a TODO
___
*[`sstable/test_utils.go` line 113 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5691#-OioJhHM6Ye9ul71zJcH:-OioJhHM6Ye9ul71zJcI:b9d96eh) ([raw file](https://github.com/cockroachdb/pebble/blob/89c96c40be7ff42b139322d39c1c33db7ba04125/sstable/test_utils.go#L113)):*
> ```Go
> 	Attr       base.ShortAttribute
> 	Meta       base.KVMeta
> 	// Tiering metadata (optional, for tiering histogram tests).
> ```

This stuff already exists as the Meta field..
___
*[`sstable/test_utils.go` line 210 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5691#-OioJjgX2KDu0xPPQVMO:-OioJjgX2KDu0xPPQVMP:bc9y8h7) ([raw file](https://github.com/cockroachdb/pebble/blob/89c96c40be7ff42b139322d39c1c33db7ba04125/sstable/test_utils.go#L210)):*
> ```Go
> 
> 		// Check for tiering metadata suffix: tiering={spanID=X attr=Y}.
> 		if idx := strings.Index(line, " tiering={"); idx != -1 {
> ```

We shouldn't need this, `ExtractKVMeta` does it for us, it just needs to be part of the value string: `"<actual value>;tiering:span=1,attr=100"`

annrpom:
TFTR! ('-')7

PR #5692, opened 2026-01-07
vfs/errorfs,internal/dsl: add CallStackIncludes

jbowens:
Add a new CallStackIncludes predicate that evaluates to true if at evaluation time the call stack includes a function containing the configured substring. I'm intending to use this within some CockroachDB randomized error injection tests to limit error injection to CockroachDB iterators (as opposed to compactions, table stats), so that an error injection is guaranteed to translate to an iterator error.

This technique is a bit fragile because it's sensitive to function renames, but there's limited risk if we use functions within the exported Pebble interface. In the future if we extend the VFS interface by propagating read categories, we can replace usages with a predicate on the read category.


RaduBerinde (commented):
Cool!

xinhaoz (approved):

jbowens:
TFTRs!

PR #5696, opened 2026-01-09
options: enable value separation by default

annrpom:


RaduBerinde (approved):

RaduBerinde:
Can you run the compaction tests under stress to make sure they're not flaky now?

annrpom:
seems to be stable; i did:
```
go test --tags invariants --exec 'stress -p 2 --maxtime 40m --timeout 2m' -v -run "^(TestCompaction)$" "./"
```

and we're at 31m in with 850 successful runs so far

TFTR! ('-')7

PR #5698, opened 2026-01-09
vfs/errorfs: reexport dsl.CallStackIncludes

jbowens:
Re-export dsl.CallStackIncludes from the errorfs package so that users of Pebble, including Cockroach, can programmatically constructed the predicate.


jbowens:
TFTR!

PR #5699, opened 2026-01-09
metamorphic: add testing for blob file ingestion

xinhaoz:
### db: add CombinedBlobFileMapping to level checker
The level checker needs to read blob files from flushable ingest entries.

### db: use fmv instead of opts.FormatMajorVersion in blob validation
Use FMV from db version instead of the one from options in
constructBlobFileMetadataForIngestedTable.

### db: protect blob files in ingestedFlushable from being obsolete
Add reference counting for blob files in ingestedFlushable entries. When
an ingestedFlushable is created, its blob files are ref'd to prevent them
from becoming obsolete while the flushable is held by any read state.
When the flushable entry is released, the blob files are unref'd.

### metamorphic: add testing for blob file ingestion
Add support for ingesting SSTs with blob files in metamorphic testing.
When value separation is enabled and the format major version supports
blob file ingestion, buildForIngest now uses valsep.SSTBlobWriter to
potentially create blob files alongside the SST.

### DIFF
--- a/metamorphic/build.go
+++ b/metamorphic/build.go
@@ -18,6 +18,7 @@ import (
 	"github.com/cockroachdb/pebble/objstorage"
 	"github.com/cockroachdb/pebble/objstorage/objstorageprovider"
 	"github.com/cockroachdb/pebble/sstable"
+	"github.com/cockroachdb/pebble/valsep"
 	"github.com/cockroachdb/pebble/vfs"
 )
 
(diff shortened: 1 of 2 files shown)


jbowens (approved):
:lgtm:

> @jbowens reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5699#-:-OioFhya7c-pDQJGeYb-:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @annrpom).

PR #5713, opened 2026-01-15
sstable: add support for dual-tier blob handles

annrpom:
Previously, values existing in both hot and cold tiers had no mechanism to store both blob handles.

This adds a coldBlobHandles column to the columnar data block format for storing cold blob handles when a value exists in both tiers. The new AddWithDualTierBlobHandles method writes entries with the hot handle in the value column and the cold handle in the dedicated column. The tiering histogram now tracks hot-and-cold blob reference bytes separately.

Row-based writers return an error for dual-tier handles since they don't support the required columns


annrpom (commented):
> @annrpom made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5713)* status: 11 of 14 files reviewed, 3 unresolved discussions (waiting on @sumeerbhola and @xinhaoz).

___
*[`-- commits` line 11 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5713#-OjwB_Y_1iijM7_gnRZE:-Ok_2brP7udsHJVV92Sb:b-zcw49p):*

Is there a unit test that tries to retrieve the value using the secondary handle and validates that we get the same result, or will that come later when the read path is more ready?
</blockquote></details>

the latter! i added a todo so i do not forget
___
*[`sstable/colblk_writer.go` line 537 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5713#-OjwDBmh5URtptE12xvn:-Ok_dm5x3qfDdgBnBb3i:bf6vmg7) ([raw file](https://github.com/cockroachdb/pebble/blob/e2111f208de6c1be5c488998dfd906a7f73d233b/sstable/colblk_writer.go#L537)):*

can we avoid duplicating a bunch of this code from `AddWithBlobHandle` and have both delegate to an internal helper, which does nothing with the `secondaryHandle` if it is empty?
</blockquote></details>

yes; done! we now do all of this
___
*[`sstable/colblk/data_block.go` line 494 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5713#-Ojw8YCmAoLY8yzw-dN8:-Ok_mTEj3mWbUgM4durO:bb7f2mf) ([raw file](https://github.com/cockroachdb/pebble/blob/e2111f208de6c1be5c488998dfd906a7f73d233b/sstable/colblk/data_block.go#L494)):*

Consider also saying that for dual-tier blob values the cold tier handle is placed here, so that normal accesses should use the cheaper access path of the primary blob handle.
</blockquote></details>

done

sumeerbhola (approved):
:lgtm: 

> @sumeerbhola reviewed 3 files and all commit messages, made 2 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5713#-:-Okua8qGByviQu59z8us:bg7igkv)* status: all files reviewed, 3 unresolved discussions (waiting on @annrpom and @xinhaoz).

___
*[`sstable/colblk_writer.go` line 470 at r3](https://reviewable.io/reviews/cockroachdb/pebble/5713#-Oku_eVw1YpEnghiNl1A:-Oku_eVw1YpEnghiNl1B:b-h99cdj) ([raw file](https://github.com/cockroachdb/pebble/blob/23596d3a292c5c6f82740d68b1e95a66d3703d59/sstable/colblk_writer.go#L470)):*
> ```Go
> 		return w.err
> 	}
> 	return w.addWithBlobHandleInternal(key, hotHandle, &secondaryHandle, attr, forceObsolete, meta)
> ```

(just being paranoid) can you check that `&secondaryHandle` doesn't cause a heap allocation.

annrpom:
TFTR! ('-')7

it doesn't look like it allocates to the heap; i also did the following to double-check:

```
● λ go build -gcflags='-m' ./sstable 2>&1 | grep -i secondaryHandle
sstable/colblk_writer.go:580:2: secondaryHandle does not escape
sstable/colblk_writer.go:479:2: secondaryHandle does not escape
```

PR #5714, opened 2026-01-15
db: fix TestCompactionCorruption flake

RaduBerinde:
The test checks that we can issue a manual compaction if the file
comes back, even if it is marked as a problem span.

There is a subtle race with background loading of stats which may try
to open the same file right before we rename it. Because all file
opens are serialized through the file cache, the manual compaction
might see the same (now stale) error. The fix is to simply retry the
manual compaction.

Informs #5372

### DIFF
--- a/compaction_test.go
+++ b/compaction_test.go
@@ -3169,6 +3169,7 @@ func TestCompactionCorruption(t *testing.T) {
 		}
 		// wait until fn() returns true.
 		wait := func(what string, fn func() bool) {
+			// Decrease to 5 * time.Second to repro flakes.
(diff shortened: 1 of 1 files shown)


RaduBerinde (commented):
TFTR!

> @RaduBerinde made 1 comment and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5714#-:-Oj6dBlK1YDHs-6foTNb:b-814g8k)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @jbowens and @xinhaoz).

PR #5716, opened 2026-01-16
db: avoid slice header allocations in constructPointIter

xinhaoz:
This commit fixes a recent allocation regression in constructPointIter.
Previously, when attempting to resolve the blob file mapping to use in
constructPointIter, we would always convert the memtables slice into an
interface, causing the slice header to escape to the heap. We now always use
the combinedBlobFileMapping, which avoids converting memtables to an
interface.


RaduBerinde:
*[`flushable.go` line 578 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5716#-Oj60ejm5Dy3Cn32Z81f:-Oj60ejm5Dy3Cn32Z81g:bl8ppr) ([raw file](https://github.com/cockroachdb/pebble/blob/7b3628c9bf2f896f5f36f01ca16afa9bd086b001/flushable.go#L578)):*
> ```Go
> 		return primary
> 	case secondaryHasBlobFiles:
> 		return secondary
> ```

Can't we always return c here so we never have to allocate? (maybe rename to Init and havw no return)

Btw I'd consider making `*flushableList` implement the interface, to male allocations more explicit.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5716)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @jbowens and @xinhaoz).

___
*[`flushable.go` line 578 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5716#-Oj60ejm5Dy3Cn32Z81f:-Oj6I-0-1FYJ0fj5606f:b-o21e8e) ([raw file](https://github.com/cockroachdb/pebble/blob/7b3628c9bf2f896f5f36f01ca16afa9bd086b001/flushable.go#L578)):*

In fact, instead of having `flushableList` implement `BlobFileMapping`, we could have a constructor that takes a `flushableList` and returns the merged `blobFileMap`s, e.g.
```
func (l flushableList) BlobFileMap() map[base.BlobFileID]*manifest.PhysicalBlobFile {
	n := 0
	for _, e := range l {
		if ingested, ok := e.flushable.(*ingestedFlushable); ok {
			n += len(ingested.blobFileMap)
		}
	}
	if n == 0 {
		return nil
	}
	m := make(map[base.BlobFileID]*manifest.PhysicalBlobFile, n)
	for _, e := range l {
		if ingested, ok := e.flushable.(*ingestedFlushable); ok {
			for id, phys := range ingested.blobFileMap {
				m[id] = phys
			}
		}
	}
	return m
}
```

Then we don't need a separate `hasBlobFiles`
</blockquote></details>

Hm, actually, that is not better. As long as we always use `combinedBlobFileMapping` which uses the `flushableList` directly, we shouldn't allocate at all.. In that case, do we still need `flushableList` to implement `BlobFileMapping` directly? If not, I'd change to `lookup()` so it doesn't.

xinhaoz (commented):
> @xinhaoz made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5716)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @jbowens).

___
*[`flushable.go` line 578 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5716#-Oj60ejm5Dy3Cn32Z81f:-Oj6YJhn8JGvJE58yXem:b-37ipoj) ([raw file](https://github.com/cockroachdb/pebble/blob/7b3628c9bf2f896f5f36f01ca16afa9bd086b001/flushable.go#L578)):*

Hm, actually, that is not better. As long as we always use `combinedBlobFileMapping` which uses the `flushableList` directly, we shouldn't allocate at all.. In that case, do we still need `flushableList` to implement `BlobFileMapping` directly? If not, I'd change to `lookup()` so it doesn't.
</blockquote></details>

Done - flushableList no longer implements `BlobFileMapping` and instead of resolving into one of combinedBlobFileMapping, BlobFileSet or flushableList, we always use combinedBlobMapping.

RaduBerinde (approved):
:lgtm:

> @RaduBerinde made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5716#-:-Oj6d1jQ9ICoEfRs9RRC:bnfp4nl)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @jbowens and @xinhaoz).

___
*[`flushable.go` line 165 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5716#-Oj6cZHO0Y9D0rGxN4We:-Oj6cZHO0Y9D0rGxN4Wf:byi4mmv) ([raw file](https://github.com/cockroachdb/pebble/blob/05103b1ec9e840d249f6a84cfd9fa817aa3d8438/flushable.go#L165)):*
> ```Go
> 
> // lookup searches all flushable entries for the given blob file ID.
> func (l flushableList) lookup(fileID base.BlobFileID) (base.ObjectInfo, bool) {
> ```

[nit] consider inlining into `combinedBlobFileMapping` since we only use it there
___
*[`flushable.go` line 179 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5716#-Oj6cXa62dSq8IYwTY0V:-Oj6cXa62dSq8IYwTY0W:bzf0raw) ([raw file](https://github.com/cockroachdb/pebble/blob/05103b1ec9e840d249f6a84cfd9fa817aa3d8438/flushable.go#L179)):*
> ```Go
> 
> // hasBlobFiles returns true if any flushable entry contains blob files.
> func (l flushableList) hasBlobFiles() bool {
> ```

Do we need this anymore?
___
*[`flushable.go` line 541 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5716#-Oj6cAQK3nQaD7JK185h:-Oj6cAQK3nQaD7JK185i:b4ljhar) ([raw file](https://github.com/cockroachdb/pebble/blob/05103b1ec9e840d249f6a84cfd9fa817aa3d8438/flushable.go#L541)):*
> ```Go
> }
> 
> // combinedBlobFileMapping chains multiple BlobFileMapping implementations,
> ```

[nit] can make this more specific now - "implements BlobFileMapping, looking up blob files in a BlobFileSet and/or flushableList"
___
*[`flushable.go` line 551 at r2](https://reviewable.io/reviews/cockroachdb/pebble/5716#-Oj6bkFK7pgWod1fAXXf:-Oj6bkFK7pgWod1fAXXg:bhumxty) ([raw file](https://github.com/cockroachdb/pebble/blob/05103b1ec9e840d249f6a84cfd9fa817aa3d8438/flushable.go#L551)):*
> ```Go
> type combinedBlobFileMapping struct {
> 	// primary is the main blob file mapping, typically from the current version.
> 	primary base.BlobFileMapping
> ```

[nit] `*manifest.BlobFileSet`

xinhaoz:
TFTR, addressed all comments!

PR #5721, opened 2026-01-16
wal: validate recovery directory through a stable identifier

annrpom:
We now persist a stable identifier to both the OPTIONS file and a file within the secondary directory (stable_identifier). When a WAL recovery directory is also the current secondary location, we validate that its identifier matches what's in the OPTIONS file. If the identifiers don't match, we abort recovery indicating that the secondary seems incorrect / corrupt.

Recovery directories that are NOT the current secondary (i.e., previous secondaries specified via prev_path) are not validated, as they have their own identifiers from when they were the active secondary.

Fixes: #4416

### DIFF
--- a/open.go
+++ b/open.go
@@ -303,6 +303,12 @@ func Open(dirname string, opts *Options) (db *DB, err error) {
 
 	// Configure failover-specific histograms and options
 	if !opts.ReadOnly && opts.WALFailover != nil {
+		if walOpts.Secondary.ID == "" {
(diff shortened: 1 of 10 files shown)


annrpom:
ok verified this works on the `TestWALFailover` test in the cockroach repo with:

```
./dev test pkg/storage -f=TestWALFailover -v -- --override_repository=com_github_cockroachdb_pebble=/Users/anniepompa/go/src/github.com/cockroachdb/pebble
```

annrpom:
TFTR! ('-')7

PR #5724, opened 2026-01-20
tool: db analyze-metadata: show count of tables with two-level index

RaduBerinde:
Show the count, not just the percentage. Useful when the percentage is
very low and shown as `~0%`.


xinhaoz (approved):

RaduBerinde:
TFTR!

PR #5729, opened 2026-01-20
binaryfuse: add 10 bit support

RaduBerinde:
The jump from 8 bits for L5 to 12 bits for L4 is significant and could
add some non-trivial filter size without much benefit (especially when
the LSM is not ideally shaped). Add support for 10
bits-per-fingerprint and use that for L4.

### DIFF
--- a/metamorphic/options.go
+++ b/metamorphic/options.go
@@ -798,7 +798,9 @@ func RandomOptions(rng *rand.Rand, kf KeyFormat, cfg RandomOptionsCfg) *TestOpti
 	fpList := []pebble.DBTableFilterPolicy{
 		pebble.UniformDBTableFilterPolicy(pebble.NoFilterPolicy),
 		pebble.UniformDBTableFilterPolicy(bloom.FilterPolicy(1 + rng.Uint32N(20))),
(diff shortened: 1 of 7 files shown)


annrpom (approved):
:lgtm:

> @annrpom reviewed 7 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5729#-:-OjWGnx-2a6wnmd7GEjG:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @xinhaoz).

___
*[`sstable/tablefilters/binaryfuse/bitpacking/bitpacking.go` line 171 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5729#-OjWFo7G1Dw6YflggB5o:-OjWFo7G1Dw6YflggB5p:bahkjqy) ([raw file](https://github.com/cockroachdb/pebble/blob/5a14a5ec8308e0678d006aaa1aa0664cc653286b/sstable/tablefilters/binaryfuse/bitpacking/bitpacking.go#L171)):*
> ```Go
> 	// Clear the tail portion of the output buffer first.
> 	tailBytes := (len(input)*10 + 7) / 8
> 	for i := 0; i <= tailBytes; i++ {
> ```

does this clear one extra byte?

PR #5736, opened 2026-01-20
sstable, db: document SeekPrefixGE for sstable and level iters

RaduBerinde:
This is an attempt to document the current semantics as they are currently implemented.

Also see the discussions in https://github.com/cockroachdb/pebble/pull/3845

I hope to be looking at a way to resolve some of these complexities, perhaps by adding more tombstone-related logic in the merging iterator (postponing seeks instead of seeking beyond the prefix just to enact a levelIter positioning suitable for a subsequent try-seek-using-next).

### DIFF
--- a/level_iter.go
+++ b/level_iter.go
@@ -703,6 +703,71 @@ func (l *levelIter) internalSeekGE(
 	return l.verify(kv), kvMeta
 }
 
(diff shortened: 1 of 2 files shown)


RaduBerinde:
TFTRs!

PR #5738, opened 2026-01-21
db: change SpanPolicyFunc to take UserKeyBounds

RaduBerinde:
#### db: change SpanPolicyFunc to take UserKeyBounds

If the span policy function is non-trivial, it might have to do a lot
of work to find the end key at which e.g. the default policy doesn't
apply anymore.

We switch to passing `UserKeyBounds`, since we only care about the
policy up to the compaction end bound.

We also add `KeyRange` to `SpanPolicy`.


RaduBerinde:
TFTR!

Issue #5029, opened 2025-07-03
cache: investigate using xsync.RBMutex


petermattis:
[`cache.shard`](https://github.com/cockroachdb/pebble/blob/master/internal/cache/clockpro.go#L79) currently provides mutual exclusion using a `sync.RWMutex`. We should investigate whether [`xsync.RBMutex`](https://github.com/puzpuzpuz/xsync?tab=readme-ov-file#rbmutex) provides better scalability under concurrent access. `xsync.RBMutex` is optimized for read-lock acquisition and for cache-like workloads where readers are much more frequent than writers.

RaduBerinde:
The cache is already sharded, so it's unlikely any one RWMutex sees a lot of concurrency (except perhaps a lot of parallel accesses to the same block? perhaps a filter block or index block?)

petermattis:
> The cache is already sharded, so it's unlikely any one RWMutex sees a lot of concurrency (except perhaps a lot of parallel accesses to the same block? perhaps a filter block or index block?)

True, though I recall seeing `shard.get` on profiles. Perhaps unrealistic workloads. It's possible there is nothing here.

PR #5740, opened 2026-01-21
sstable: push filter check inside cache

RaduBerinde:
Previously, every bloom filter check in `SeekPrefixGE` required:
1. Getting the filter block from cache (atomic refcount increment)
2. Calling `MayContain` on the filter data (a very fast operation)
3. Releasing the block (atomic refcount decrement)
4. Updating hit counter (atomic increment)

For a simple bloom filter check (a few hash operations), the atomic
refcount operations represent significant overhead (especially when
there are a small number of very hot filters in the upper LSM levels).

Add a new `cache.Handle.TableFilterMayContain` method that performs
the filter check while holding the cache's read lock, without
incrementing/decrementing the refcount. The value is safe to access
while holding the read lock since eviction requires the write lock.

When the filter block is not in cache, the code falls back to the
existing `readFilterBlock` path which handles reading from disk and
populating the cache.

We no longer record these accesses as cache hits for statistics
purposes. We still record misses, and since we now separate statistics by
block type, we'll still know what percentage of overall misses are due
to filter blocks.

### DIFF
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -271,6 +271,28 @@ func (c *Handle) Get(
 	return c.cache.getShard(k).get(k, level, category, false /* peekOnly */)
 }
 
+// TableFilterMayContain looks up a cached filter block and checks if it may contain the
+// given key, while holding the cache's read lock. This avoids the refcount
+// overhead of Get/Release for the common case of bloom filter checks.
+//
+// TableFilterMayContain does not update hit counters.
+//
+// The dataOffset specifies how many bytes to skip at the start of the cached
+// buffer (typically block.MetadataSize to skip block metadata).
+//
+// Returns (true, mayContain) if found; (false, false) if not in cache.
+// When not found, the caller should fall back to reading from disk.
+func (c *Handle) TableFilterMayContain(
+	fileNum base.DiskFileNum,
+	offset uint64,
+	dataOffset int,
+	filter base.TableFilterDecoder,
+	filterKey []byte,
+) (found bool, mayContain bool) {
+	k := makeKey(c.id, fileNum, offset)
+	return c.cache.getShard(k).tableFilterMayContain(k, dataOffset, filter, filterKey)
+}
+
 // GetWithReadHandle retrieves the cache value for the specified handleID, fileNum
 // and offset. If found, a valid Handle is returned (with cacheHit set to
 // true), else a valid ReadHandle is returned.
--- a/internal/cache/clockpro.go
+++ b/internal/cache/clockpro.go
@@ -184,6 +184,34 @@ func (c *shard) getWithReadEntry(k key, level base.Level, category Category) (*V
 	return nil, re
 }
 
+// tableFilterMayContain looks up a cache entry and checks if the filter may
+// contain the given key, while holding the read lock. This avoids refcount
+// overhead for bloom filter checks.
+//
+// The dataOffset specifies how many bytes to skip at the start of the cached
+// buffer (to skip block metadata).
+//
+// tableFilterMayContain does not update hit counters.
+//
+// Returns (true, mayContain) if found; (false, false) otherwise.
+func (c *shard) tableFilterMayContain(
+	k key, dataOffset int, filter base.TableFilterDecoder, filterKey []byte,
+) (found bool, mayContain bool) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	if e, _ := c.blocks.Get(k); e != nil {
+		if v := e.val; v != nil && len(v.buf) > dataOffset {
+			// Update referenced flag for CLOCK-Pro (same as regular get).
+			if !e.referenced.Load() {
+				e.referenced.Store(true)
+			}
+			return true, filter.MayContain(v.buf[dataOffset:], filterKey)
+		}
+	}
+	return false, false
+}
+
 func (c *shard) set(k key, value *Value, markAccessed bool) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
--- a/sstable/block/block.go
+++ b/sstable/block/block.go
@@ -375,6 +375,11 @@ func (r *Reader) ChecksumType() ChecksumType {
 	return r.checksumType
 }
 
+// CacheHandle returns the cache Handle, or nil if caching is disabled.
+func (r *Reader) CacheHandle() *cache.Handle {
+	return r.opts.CacheOpts.CacheHandle
+}
+
 var kindToCacheCategory = [blockkind.NumKinds]cache.Category{
 	blockkind.Unknown:                         cache.CategoryBackground,
 	blockkind.SSTableData:                     cache.CategorySSTableData,
--- a/sstable/filter.go
+++ b/sstable/filter.go
@@ -65,3 +65,8 @@ func (f *tableFilterReader) mayContain(data, key []byte) bool {
 	}
 	return mayContain
 }
+
+// Decoder returns the underlying filter decoder.
+func (f *tableFilterReader) Decoder() base.TableFilterDecoder {
+	return f.decoder
+}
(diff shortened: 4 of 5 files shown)


petermattis (commented):
> @petermattis made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5740)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola).

___
*[`internal/cache/clockpro.go` line 209 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5740#-OjXIDz7Cc3tfSzgvbMH:-OjXIDz7Cc3tfSzgvbMI:b-kzdlhw) ([raw file](https://github.com/cockroachdb/pebble/blob/9a25ee793e1631af618e68c4546463662215eea2/internal/cache/clockpro.go#L209)):*
> ```Go
> 				e.referenced.Store(true)
> 			}
> 			return true, filter.MayContain(v.buf[dataOffset:], filterKey)
> ```

It's unfortunate that we're pushing knowledge of this `MayContains` interface into the cache. Is there a structure where we pass in a closure instead? Something like "fn func(data []byte)". The closure would have to capture the `TableFilterDecoder` and `filterKey` variables which would be problematic allocations. Perhaps there could be a small struct that contains the state needed by the closure and we can use a `sync.Pool` for that struct.

RaduBerinde:
*[`internal/cache/clockpro.go` line 209 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5740#-OjXIDz7Cc3tfSzgvbMH:-OjXQSeb7kWf37kl6vFh:b-ggpeb) ([raw file](https://github.com/cockroachdb/pebble/blob/9a25ee793e1631af618e68c4546463662215eea2/internal/cache/clockpro.go#L209)):*

It's unfortunate that we're pushing knowledge of this `MayContains` interface into the cache. Is there a structure where we pass in a closure instead? Something like "fn func(data []byte)". The closure would have to capture the `TableFilterDecoder` and `filterKey` variables which would be problematic allocations. Perhaps there could be a small struct that contains the state needed by the closure and we can use a `sync.Pool` for that struct.
</blockquote></details>

I wouldn't want a pool operation for this.. I'll try to think of something (perhaps cache the callback in the Reader). But I don't want to encourage arbitrary function execution inside the lock either, so I'm not sure it's much better in terms of API.  Another option would be an Acquire/Releaee type API, though that's prone to be abused too.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5740)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @petermattis and @sumeerbhola).

___
*[`internal/cache/clockpro.go` line 209 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5740#-OjXIDz7Cc3tfSzgvbMH:-OjYZ2ev0xjH86_-zaPD:b-9p66df) ([raw file](https://github.com/cockroachdb/pebble/blob/9a25ee793e1631af618e68c4546463662215eea2/internal/cache/clockpro.go#L209)):*

I wouldn't want a pool operation for this.. I'll try to think of something (perhaps cache the callback in the Reader). But I don't want to encourage arbitrary function execution inside the lock either, so I'm not sure it's much better in terms of API.  Another option would be an Acquire/Releaee type API, though that's prone to be abused too.
</blockquote></details>

I will move this to draft and come back to it later. I think it can be done nicely with something along the lines of
```
func QuickAccess<Arg any, Res any> (c *Cache, fileNum base.DiskFileNum, offset uint64, fn func(Arg, []byte) Res) (_ Res, found bool)
```

We can pass whatever args we need to the function via `Arg` without having to capture them.

I found some other things along the way that I will work on at the same time:
 - filter metrics add another per-access atomic ops and they are not particularly useful as they are (I'd rather have sharded counters separated by level)
 - TableFilterDecoder could be a struct that has a `mayContain` `func` fields since the decoder is a stateless singleton
 - we could be initializing some state in the block metadata once instead of "decoding" stuff from the filter each time (though it's minor)
 - we could be using `xsync.RBLock` instead of `RWMutex` (which shards the reader counter)

This is from a benchmark I just wrote: it's a bit pessimistic as 8 workers all access the same filter, but it's pretty crazy how small the `probe` part actually is:

petermattis (commented):
> @petermattis made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5740)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde and @sumeerbhola).

___
*[`internal/cache/clockpro.go` line 209 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5740#-OjXIDz7Cc3tfSzgvbMH:-Oja8I5731_q4GZlax1S:b-y0y4pm) ([raw file](https://github.com/cockroachdb/pebble/blob/9a25ee793e1631af618e68c4546463662215eea2/internal/cache/clockpro.go#L209)):*

I will move this to draft and come back to it later. I think it can be done nicely with something along the lines of
```
func QuickAccess<Arg any, Res any> (c *Cache, fileNum base.DiskFileNum, offset uint64, fn func(Arg, []byte) Res) (_ Res, found bool)
```

We can pass whatever args we need to the function via `Arg` without having to capture them.

I found some other things along the way that I will work on at the same time:
 - filter metrics add another per-access atomic ops and they are not particularly useful as they are (I'd rather have sharded counters separated by level)
 - TableFilterDecoder could be a struct that has a `mayContain` `func` fields since the decoder is a stateless singleton
 - we could be initializing some state in the block metadata once instead of "decoding" stuff from the filter each time (though it's minor)
 - we could be using `xsync.RBLock` instead of `RWMutex` (which shards the reader counter)

This is from a benchmark I just wrote: it's a bit pessimistic as 8 workers all access the same filter, but it's pretty crazy how small the `probe` part actually is:

</blockquote></details>

Ack that using a sync.Pool may be problematic (too slow) for this usage. `QuickAccess` looks good.

Last year you speculated that `xsync.RBMutex` might be useful for contended blocks like the filter block or index block: https://github.com/cockroachdb/pebble/issues/5029.

RaduBerinde:
*[`internal/cache/clockpro.go` line 209 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5740#-OjXIDz7Cc3tfSzgvbMH:-OjaJEgOC5Xs_gAecYEn:b-wa1bb4) ([raw file](https://github.com/cockroachdb/pebble/blob/9a25ee793e1631af618e68c4546463662215eea2/internal/cache/clockpro.go#L209)):*

Ack that using a sync.Pool may be problematic (too slow) for this usage. `QuickAccess` looks good.

Last year you speculated that `xsync.RBMutex` might be useful for contended blocks like the filter block or index block: https://github.com/cockroachdb/pebble/issues/5029.
</blockquote></details>

Yeah, I did run some experiments with it a while back and I didn't notice an obvious gain (I think it was TPCC). It might bring in a more noticeable benefit once we remove the other overhead.

sumeerbhola (commented):
> @sumeerbhola reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5740)* status: all files reviewed, 1 unresolved discussion (waiting on @petermattis).

___
*[`internal/cache/clockpro.go` line 209 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5740#-OjXIDz7Cc3tfSzgvbMH:-OjlLjIA1aEyZ_7B9lBB:b-9xcuyj) ([raw file](https://github.com/cockroachdb/pebble/blob/9a25ee793e1631af618e68c4546463662215eea2/internal/cache/clockpro.go#L209)):*

Yeah, I did run some experiments with it a while back and I didn't notice an obvious gain (I think it was TPCC). It might bring in a more noticeable benefit once we remove the other overhead.
</blockquote></details>

This profile was very revealing in terms of the cost of all the "not real work".
The generic function approach sounds good.

PR #5742, opened 2026-01-21
db: fix TestIterHistories probe tests

jbowens:
This commit fixes a few issues with TestIterHistories and its use of iterator probes. Specifically, over the lifetime of a database calls to combined-iter would nest overloaded newIters closures, resulting in probes from previous test cases leaking into the current test case. Additionally, some of the tests introduced in #5256 to test SeekPrefixGE lazy positioning used nonsensical probe prefixes; this commit fixes them to prefix the probe's log lines with the relevant file number.

### DIFF
--- a/iterator_histories_test.go
+++ b/iterator_histories_test.go
@@ -105,6 +105,32 @@ func TestIterHistories(t *testing.T) {
 			}
 			return err
 		}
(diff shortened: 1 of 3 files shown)


RaduBerinde (approved):

jbowens:
tftr!

PR #5743, opened 2026-01-21
internal/keyspan: fix interleaving iterator edge case

jbowens:
Fix an edge case in the interleaving iterator when seeking to, or beyond, an
iterator bound. Previously, an interleaving iterator configured with an upper
bound 'b' that executed a SeekGE('b') would leave the iterator in such a state
that a subsequent Prev could omit a span entirely.

This commit fixes the bug by detecting these problematic seeks upfront, and
eliding the seek altogether if necessary. Relative positioning operations
detect that the iterator is still unpositioned and perform seeks to position
the iterator at the appropriate bound.

### DIFF
--- a/internal/keyspan/interleaving_iter.go
+++ b/internal/keyspan/interleaving_iter.go
@@ -182,6 +182,8 @@ type interleavePos int8
 
 const (
 	posUninitialized interleavePos = iota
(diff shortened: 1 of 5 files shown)


jbowens:
TFTR!

PR #5745, opened 2026-01-22
internal/manifest: add tiering info to blob references 

annrpom:
Previously, blob references did not store which tier (hot or cold) the blob
file was written to.

This patch adds a Tier field to both BlobReference and PhysicalBlobFile.
The tier is now persisted in version edits using new tags (tagNewBlobFile2 and
customTagBlobReferences3) that include tier information in the encoding.
Decoding of old formats defaults to HotTier for backward compatibility.

WriterOptions.BlobReferenceTierGetter is replaced with BlobReferenceTiers,
a slice indexed by reference ID.


xxmplus:
Close to due to age

PR #5760, opened 2026-01-31
sstable: Expose block‑cache configuration for sstable.ReaderOptions

ankur-anand:
## What
This PR adds a small public shim to configure the block cache used by sstable.ReaderOptions.

## Why

* The block cache (cache.Cache) is already public, but callers cannot use it with sstable.ReaderOptions because
    CacheOpts is internal (sstableinternal.CacheOptions).
* This prevents external users who call sstable.NewReader directly from enabling block caching or evicting per‑file
    blocks.
    
## Changes:
* Added CacheHandle alias.
* Added sstable.SetCacheOptions and CacheHandleEvictFile helpers.
* No behavioral changes; fully backward compatible.

### Shim Example Usage.

```
  c := pebble.NewCache(128 << 20)
  defer c.Unref()
  h := c.NewHandle()
  defer h.Close()

  // Configure ReaderOptions to use the cache through the shim API
  opts := sstable.ReaderOptions{}
  sstable.SetCacheOptions(&opts, h, fileNum)

  r, err := sstable.NewReader(ctx, readable, opts)
  if err != nil {
  }
  defer r.Close()
```

### DIFF
--- a/sstable/external_reader.go
+++ b/sstable/external_reader.go
@@ -43,9 +43,6 @@ type CacheHandle = cache.Handle
 // cache handle and file number. The file number is caller-assigned and must be
 // unique within the cache handle's namespace to avoid collisions.
 func SetCacheOptions(opts *ReaderOptions, handle *CacheHandle, fileNum uint64) {
-	if opts == nil || handle == nil {
-		return
-	}
 	opts.CacheOpts = sstableinternal.CacheOptions{
 		CacheHandle: handle,
 		FileNum:     base.DiskFileNum(fileNum),
@@ -56,8 +53,5 @@ func SetCacheOptions(opts *ReaderOptions, handle *CacheHandle, fileNum uint64) {
 // from the cache handle. Call this when an sstable is no longer needed to free
 // cache space.
 func CacheHandleEvictFile(handle *CacheHandle, fileNum uint64) {
-	if handle == nil {
-		return
-	}
 	handle.EvictFile(base.DiskFileNum(fileNum))
 }


ankur-anand:
@sumeerbhola Can you please review? Seems this will need approval for workflows run as well.

sumeerbhola (changes requested):
> @sumeerbhola made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/pebble/5760)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @ankur-anand).

___
*[`cache_handle.go` line 11 at r1](https://reviewable.io/reviews/cockroachdb/pebble/5760#-OkUr5dt5seDexV1JeJ9:-OkUr5dt5seDexV1JeJA:bahig1c) ([raw file](https://github.com/cockroachdb/pebble/blob/c86e5e258af2db03c4facc35f6d2e4ef36bf2bf7/cache_handle.go#L11)):*
> ```Go
> 	"github.com/cockroachdb/pebble/internal/cache"
> )
> ```

The reason we did not expose `cache.Handle` being used by a Pebble instance is that someone can collide with the filenums in that Pebble instance. This PR of course does not have that risk since a new `cache.Handle` is being constructed and then used with caller assigned filenums.

I would prefer that these changes be localized to one place with the usage comment above them saying what these are for. Otherwise, someone can be confused that they need to call `SetCacheOptions`or `CacheHandleEvictFile` when dealing with Pebble internal sstables. Say in a external_reader.go file in the sstable package.

